Issue #5966, opened 2016-05-13
Replace 'Starter POM' by 'Starter' in the documentation


snicoll:
I find the "POM" part a bit confusing. Our starters are jar artifacts and all that's required is to add a groupId and artifactId. 

POM may lead the user to think this is a BOM or that they should add the `<type>pom</type>` to the declaration to make it work.

Thoughts?

snicoll:
I've also updated [the release notes for 1.4](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-1.4-Release-Notes)

PR #10314, opened 2017-09-16
Fixing an incorrect link

luchob:
This PR just fixes a link to the official documentation. Very trivial but offers some comfort for newcomers :)

### DIFF
--- a/spring-boot-starters/README.adoc
+++ b/spring-boot-starters/README.adoc
@@ -8,14 +8,14 @@ JPA for database access just include the `spring-boot-starter-data-jpa` dependen
 your project, and you are good to go.
 
 For complete details see the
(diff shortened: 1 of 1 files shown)


snicoll:
See #5966 - I thought adding two links there would not have broken the existing one. I'd rather fix that instead. I don't know yet how though. Thanks for the PR!

snicoll:
I did try to restore the anchors with two values (using [this SO answer](https://stackoverflow.com/questions/35060449/how-can-i-add-mutiple-anchors-to-the-same-block) but that didn't really work as expected. Given the link is broken for quite some time anyway I went ahead and remove the outdated duplicate. Thanks for the PR!

PR #10316, opened 2017-09-16
Fix names of new metrics properties

michael-simons:
The names of the configuration properties were not up to date in the docs.


snicoll:
Merged, thanks

PR #10350, opened 2017-09-19
Remove usage of `HttpStatus` in Web Endpoints

### LINKED ISSUE #10348: Remove usage of HttpStatus in WebEndpointExtensions

vpavic:
This resolves #10348.

### DIFF
--- a/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/audit/AuditEventsWebEndpointExtension.java
+++ b/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/audit/AuditEventsWebEndpointExtension.java
@@ -32,8 +32,6 @@ import org.springframework.boot.actuate.endpoint.web.annotation.WebEndpointExten
 @WebEndpointExtension(endpoint = AuditEventsEndpoint.class)
 public class AuditEventsWebEndpointExtension {
 
-	private static final int BAD_REQUEST_STATUS = 400;
-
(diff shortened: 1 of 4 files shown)


philwebb:
It might be worth pulling some of the more common status codes up into `WebEndpointResponse`.

wilkinsona:
That's a nice idea, @philwebb. I also wonder if we should add a Checkstyle import exclusion to stop `HttpStatus` from creeping in again.

vpavic:
Both ideas (common status codes and Checkstyle) sound very nice to me. Once consensus is reached on those two topics I can easily update the PR to address them.

philwebb:
Consider a consensus reached :)

vpavic:
I've added a commit that defines common status codes in `WebEndpointResponse`.

Checkstyle rule is not so trivial though, as there are other `HttpStatus` usages across Actuator, for example:
- `org.springframework.boot.actuate.env.EnvironmentEndpoint.NoSuchPropertyException`
- `org.springframework.boot.actuate.metrics.web.reactive.server.WebFluxTags#status`
- `org.springframework.boot.actuate.trace.WebRequestTraceFilter#doFilterInternal`

snicoll:
`NoSuchPropertyException` is unused and it looks like something is missing compared to what we had in `1.5.x`. The two others are legit so I guess we'd need to work with certain packages.

snicoll:
I found some more issues so let's handle the checkstyle rule in #10358

PR #10379, opened 2017-09-21
Update pom.xml to make spring-boot 1.5.x buildable on jdk-9

JurrianFahner:
The maven-jar-plugin 2.6 is not compatible with jdk-9. In order to make 1.5.x spring boot projects buildable under jdk-9 this dependency needs an upgrade. I discovered that 3.0.2 works very nice. 

A workaround could be for users to override the version in the plugin section:

```
      <plugin>
        <artifactId>maven-jar-plugin</artifactId>
        <version>3.0.2</version>
      </plugin>
```

### DIFF
--- a/spring-boot-dependencies/pom.xml
+++ b/spring-boot-dependencies/pom.xml
@@ -207,7 +207,7 @@
 		<maven-install-plugin.version>2.5.2</maven-install-plugin.version>
 		<maven-invoker-plugin.version>1.10</maven-invoker-plugin.version>
 		<maven-help-plugin.version>2.2</maven-help-plugin.version>
-		<maven-jar-plugin.version>2.6</maven-jar-plugin.version>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the pull request. We can upgrade the plugin from 2.6 to 3.0.2 in the manner proposed as it will force the same change on Boot's users and that's not appropriate in a 1.5.x release. Boot 2.0 is the release where we're targeted support for Java 9. I think we'll probably have to close this but, before I do, why do you want to be able to build 1.5.x with Java 9?

JurrianFahner:
If this change can give problems for some users, then we shouldn't proceed. I thought if this is only hurdle to support JDK-9, if it isn't the case than we can override the plugin ourselves.

I'm fine with the closing this pull request.

wilkinsona:
Thanks for following up. There's plenty more to do to support Java 9. Please see the issues labelled with `theme: java9`.

PR #10387, opened 2017-09-23
Fix check vendor location in FlywayAutoConfiguration

eddumelendez:
`flyway.locations` supports vendor resolution. But, if
`flyway.check-location=true` is added and there is just one location
with vendor location the current implementation fails. This commit adds
validation for vendor locations.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
@@ -16,12 +16,12 @@
 
 package org.springframework.boot.autoconfigure.flyway;
 
+import java.util.Arrays;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.List;
 import java.util.Set;
 
-import javax.annotation.PostConstruct;
 import javax.persistence.EntityManagerFactory;
 import javax.sql.DataSource;
 
@@ -104,6 +104,8 @@ public class FlywayAutoConfiguration {
 
 		private List<FlywayCallback> flywayCallbacks;
 
+		private static final String VENDOR_PLACEHOLDER = "{vendor}";
+
 		public FlywayConfiguration(FlywayProperties properties,
 				ResourceLoader resourceLoader, ObjectProvider<DataSource> dataSource,
 				@FlywayDataSource ObjectProvider<DataSource> flywayDataSource,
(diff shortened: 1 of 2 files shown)


eddumelendez:
> I don't understand why the code was moved to a separate auto-configuration. Can you expand on that please?

At the beginning I thought that previous validation (`FlywayValidatorAutoConfiguration`) would be right but in the process I just realized that if I am using the almost the same datasource configuration for `Flyway` I should throw a `BeanCreationException` maybe in the same class `FlywayAutoConfiguration`. I can rework this if we agreed to merge it. 

> I am not sure checking vendor location's location is a good thing.

Of course this is not required, we can use the default configuration o a custom one without need for vendor location. But, since this feature about resolve `{vendor}` was added, if users want to handle scripts just by vendors I think a proper validation for scripts location would be useful.

snicoll:
> if users want to handle scripts just by vendors I think a proper validation for scripts location would be useful.

Yes but the problem is that you're not providing a path in your configuration, you are providing a placeholder that's going to be replaced with the database that you are using. If you are using two systems and one of them do not need specific scripts, the startup will fail because the directory does not exist. It means that the use case of this feature is quite narrowed IMO: all the databases that you support _must_ require database specific scripts.

Does that make sense?

vpavic:
Thanks for the ping @snicoll.

> If you are using two databases and one of them does not need specific scripts, the startup will fail because the directory does not exist.

Unless I'm missing something from the scenario you're describing, there shouldn't be a problem like this since the location validation is still based on finding at least one match in provided locations. The only difference is that the location under validation are now having `{vendor}` placeholder replaced with an actual value. So having something like `flyway.locations=classpath:db/migration/common,classpath:db/migration/{vendor}` shouldn't run into problems when running for vendor that doesn't have a specific migration, right? OTOH with only `flyway.locations=classpath:db/migration/{vendor}` this provides validation capabilities that are currently not possible.

vpavic:
If one uses only vendor specific locations things are currently broken by default in `2.0.0.M6` due to #10807 being merged. So IMO either this needs to be addressed or #10807 need to be reverted.

I've just faced this problem while upgrading from `2.0.0.M5` to `2.0.0.M6` with `spring.flyway.locations` set to `db/migration/{vendor}`. Initialization fails with:

```
java.lang.IllegalStateException: Cannot find migrations location in: [db/migration/{vendor}] (please add migrations or check your Flyway configuration)
```

Setting `spring.flyway.check-location` to `false` works around the problem but it's not the greatest user experience.

wilkinsona:
@eddumelendez It looks like me like the code may be simpler if the validation was performed during the creation of the `Flyway` bean. That would provide a single place for both replacing the `{vendor}` placeholder and for validating that at least one of the locations exists. It would also remove the need for the `SpringBootFlyway` subclass. What do you think?

eddumelendez:
@wilkinsona I will rework this later today :wink:

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
```diff
@@ -117,32 +119,25 @@ public FlywayConfiguration(FlywayProperties properties,
...
+			String[] locations = resolveLocations(flyway.getDataSource());
+			checkLocationExists(locations);
+			flyway.setLocations(locations);
+			return flyway;
+		}
+
+		private Flyway createFlyway() {
+			Flyway flyway = new Flyway() {
+				@Override
+				public void setLocations(String... locations) {
```
eddumelendez: @wilkinsona I'm doing so because `@ ConfigurationProperties ` is overriding the value. May the `SpringBootFlyway` class approach looks better? WDYT?

wilkinsona: Ah, yes. Of course. I'd forgotten that we still bind directly to the `Flyway` instance. You're right. Let's keep the `SpringBootFlyway` class please.

philwebb:
Thanks for updates. I've extracted the resolve logic into its own class and pushed this to master.

PR #10440, opened 2017-09-28
Add auto-config for Spring Session cookie serializer

### LINKED ISSUE #10371: Support configuring Spring Session's session cookie using `ServerProperties`
Spring Session has support for configuring session cookie using Servlet API's `SessionCookieConfig`, so one would expect it would be possible to configure session cookie using Boot's `ServerProperties`. This however does not work due to following reasons.

The problem is that Spring Session's `CookieSerializer` is configured in `SpringHttpSessionConfiguration` while Boot's customization of `SessionCookieConfig` with `ServerProperties` is done in `DefaultServletWebServerFactoryCustomizer.SessionConfiguringInitializer` and happens after Spring Session has been initializing.

The problem can be observed using `spring-boot-sample-session` with configuration property `server.session.cookie.name=sid`, and breakpoints set in `org.springframework.boot.autoconfigure.web.servlet.DefaultServletWebServerFactoryCustomizer.SessionConfiguringInitializer#configureSessionCookie` and `org.springframework.session.config.annotation.web.http.SpringHttpSessionConfiguration#createDefaultCookieSerializer`.

Related to #7141.

vpavic:
This PR adds support for auto-configuring Spring Session's `DefaultCookieSerializer` using `ServerProperties`.

The auto-configuration of `DefaultCookieSerializer` will back-off is users provide their own `CookieSerializer` or `HttpSessionStrategy` bean.

This resolves #10371.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
@@ -38,9 +38,11 @@ import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.boot.autoconfigure.jdbc.JdbcTemplateAutoConfiguration;
 import org.springframework.boot.autoconfigure.mongo.MongoAutoConfiguration;
 import org.springframework.boot.autoconfigure.mongo.MongoReactiveAutoConfiguration;
+import org.springframework.boot.autoconfigure.web.ServerProperties;
 import org.springframework.boot.autoconfigure.web.reactive.HttpHandlerAutoConfiguration;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.ApplicationContext;
+import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.ImportSelector;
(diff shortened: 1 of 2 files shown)


wilkinsona:
>  or `HttpSessionStrategy` bean.

Is there a possibility that a user will want to use the auto-configured `DefaultCookieSerializer` to create a custom `CookieHttpSessionStrategy`?

vpavic:
Thanks for the feedback @wilkinsona.

> Is there a possibility that a user will want to use the auto-configured `DefaultCookieSerializer` to create a custom `CookieHttpSessionStrategy`?

Yes, that indeed is a possibility. We can only discard that option if user configured `HttpSessionStrategy` is an instance of `HeaderHttpSessionStrategy`.

wilkinsona:
Thanks, @vpavic. Sounds like an update to the conditions may be in order. However, I think the preferred solution may be to fix https://github.com/spring-projects/spring-boot/issues/10699 although I'm not yet sure how to do so without adding something like `ServletContextConfigurer` that's subtly different to `ServletContextInitializer`.

vpavic:
I agree that fixing #10699 would be the preferred solution.

Nevertheless, I'll update the discussed conditions to have this PR available as a backup plan.

vpavic:
> I'll update the discussed conditions to have this PR available as a backup plan.

Done.

vpavic:
With #10699 resolved, this PR isn't needed any more.

I've tested the latest `2.0.0.BUILD-SNAPSHOT` and auto-configuring Spring Session's `DefaultCookieSerializer` using `ServerProperties` now works as expected.

wilkinsona:
Thanks for testing the snapshot, @vpavic. Much appreciated.

PR #10450, opened 2017-09-28
embeddedLaunchScript: Add support for FreeBSD.

zaunerc:
The following things had to be changed:

* Shebang:
  * The default location of the Bash binary on FreeBSD is `/usr/local/bin/bash`.
    Usually one installs Bash on FreeBSD using the `pkg install bash` command.
  * On Linux the Bash binary is normally located at `/bin/bash`. Therefore we use `env` 
    to allow the interpreter to be looked up via the `PATH`.
  * Note: `env` is located at `/usr/bin/env` on both Linux and FreeBSD.
  * Also, creating a symlink to point from `/bin/bash` to `/usr/local/bin/bash` when
    using FreeBSD might not be feasable in cases where the `/bin/` directory is a
    shared directory between multiple FreeBSD jails.

* System V init script dir:
  * On FreeBSD the System V init scripts are stored
    under `/etc/rc.d/` instead of `/etc/init.d`.

### DIFF
--- a/spring-boot-tools/spring-boot-loader-tools/src/main/resources/org/springframework/boot/loader/tools/launch.script
+++ b/spring-boot-tools/spring-boot-loader-tools/src/main/resources/org/springframework/boot/loader/tools/launch.script
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 #
 #    .   ____          _            __ _ _
 #   /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
@@ -28,11 +28,21 @@ WORKING_DIR="$(pwd)"
 [[ -n "$JARFILE" ]] && jarfile="$JARFILE"
 [[ -n "$APP_NAME" ]] && identity="$APP_NAME"
 
+if [[ -d /etc/rc\.d ]]; then
+  # System V init scripts folder on FreeBSD
+  sysv_dir="rc\.d"
+else
+  sysv_dir="init\.d"
+fi
+
 # Follow symlinks to find the real jar and detect init.d script
 cd "$(dirname "$0")" || exit 1
 [[ -z "$jarfile" ]] && jarfile=$(pwd)/$(basename "$0")
 while [[ -L "$jarfile" ]]; do
-  if [[ "$jarfile" =~ init\.d ]]; then
+  # Quoting the string argument to the [[ command's =~ operator forces
+  # string matching (as with the other pattern-matching operators).
+  # Since Bash v3.2.
+  if [[ "$jarfile" =~ $sysv_dir ]]; then
     init_script=$(basename "$jarfile")
   else
     configfile="${jarfile%.*}.conf"


pivotal-issuemaster:
@zaunerc Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10450)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10450) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@zaunerc Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10450)!

wilkinsona:
Thanks for the pull request. While the changes are small, I'm a little wary of merging this as we currently have no way to test the launch script on FreeBSD. Our integration tests for the launch script use Docker and run on our Linux CI agents and my understanding is that a FreeBSD-based Docker container can only run on a FreeBSD-based host. Let's see what the rest of the team thinks.

zaunerc:
I see. So in the end this might boil down to if you require / want the script to be integration tested on each platform you support...

As far as I know FreeBSD-based Docker containers can only run on a FreeBSD-based host. See e.g. [here](https://store.docker.com/community/images/auchida/freebsd) and [here](https://stackoverflow.com/questions/33864142/can-freebsd-be-run-inside-docker#comment55492842_33864293). Furthermore Docker on FreeBSD seems to be experimental (see e.g. [here](https://wiki.freebsd.org/Docker) and [here](https://stackoverflow.com/questions/33864142/can-freebsd-be-run-inside-docker#comment72574399_33864293)).

So while I don't know your CI setup in detail it really might be impossible to run your current integration tests for this script on FreeBSD without any additional efforts (that is, e.g. a special CI setup for FreeBSD).

wilkinsona:
Correct, and I think it's unlikely that the effort of setting up and maintaining a FreeBSD CI environment will be worth the effort. Also, please note that we [already document](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#deployment-install-supported-operating-systems) that a custom script will be necessary for FreeBSD. If this is to be merged, and I think it's a fairly big if, that documentation should be updated too.

zaunerc:
If you decide to merge this in I will update the PR to also reflect the changes in the documentation. I have just seen that the Spring Boot reference guide happens to be maintained in this repo too...

zaunerc:
While the changes in this PR enable you on FreeBSD to execute the Spring Boot app like an executable and you are able to use the `start`, `stop`, etc. commands if you invoke the symlink from `/etc/rc.d/` manually, I have discovered that it lacks proper integration into the FreeBSD init system.

I had the assumption that FreeBSD, like most Linux distributions, also uses / supports SysV init scripts. But that's not the case:

> Many Linux® distributions use the SysV init system, whereas FreeBSD uses the traditional BSD-style init(8).  - [source](https://www.freebsd.org/doc/en/articles/linux-users/startup.html)

Therefore I have come to the conclusion that the best approach to deploy a Spring Boot app onto your FreeBSD server is to provide a custom init script. Like you already state in your [docs](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#deployment-install-supported-operating-systems) ;-)

Examples of FreeBSD init scripts dealing with programs which are not able to fork to background by themselves can be found [here](https://svnweb.freebsd.org/ports/head/devel/viewvc/files/viewvc.in?view=markup) or [here](https://svnweb.freebsd.org/ports/head/net/socat/files/socat.in?view=markup).

PR #10451, opened 2017-09-28
Replace some String.length() checks with String.isEmpty()

dreis2211:
Hey,

this PR replaces usages of some String.length() checks with the arguably more expressive String.isEmpty() checks.

Cheers,
Christoph


snicoll:
Thanks @dreis2211!

PR #10457, opened 2017-09-29
Polish

izeye:

### DIFF
--- a/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/ShutdownEndpoint.java
+++ b/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/ShutdownEndpoint.java
@@ -38,18 +38,18 @@ import org.springframework.context.ConfigurableApplicationContext;
 @Endpoint(id = "shutdown", defaultEnablement = DefaultEnablement.DISABLED)
 public class ShutdownEndpoint implements ApplicationContextAware {
 
-	private static final Map<String, Object> NO_CONTEXT_MESSAGE = Collections
-			.unmodifiableMap(Collections.<String, Object>singletonMap("message",
+	private static final Map<String, String> NO_CONTEXT_MESSAGE = Collections
+			.unmodifiableMap(Collections.<String, String>singletonMap("message",
 					"No context to shutdown."));
 
-	private static final Map<String, Object> SHUTDOWN_MESSAGE = Collections
-			.unmodifiableMap(Collections.<String, Object>singletonMap("message",
+	private static final Map<String, String> SHUTDOWN_MESSAGE = Collections
+			.unmodifiableMap(Collections.<String, String>singletonMap("message",
 					"Shutting down, bye..."));
 
 	private ConfigurableApplicationContext context;
 
 	@WriteOperation
-	public Map<String, Object> shutdown() {
+	public Map<String, String> shutdown() {
 		if (this.context == null) {
 			return NO_CONTEXT_MESSAGE;
 		}
--- a/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsPropertyDefaultsPostProcessor.java
+++ b/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsPropertyDefaultsPostProcessor.java
@@ -42,23 +42,23 @@ import org.springframework.core.env.PropertySource;
 @Order(Ordered.LOWEST_PRECEDENCE)
 public class DevToolsPropertyDefaultsPostProcessor implements EnvironmentPostProcessor {
 
-	private static final Map<String, Object> PROPERTIES;
+	private static final Map<String, Object> properties;
 
 	static {
-		Map<String, Object> properties = new HashMap<>();
-		properties.put("spring.thymeleaf.cache", "false");
-		properties.put("spring.freemarker.cache", "false");
-		properties.put("spring.groovy.template.cache", "false");
-		properties.put("spring.mustache.cache", "false");
-		properties.put("server.session.persistent", "true");
-		properties.put("spring.h2.console.enabled", "true");
-		properties.put("spring.resources.cache-period", "0");
-		properties.put("spring.resources.chain.cache", "false");
-		properties.put("spring.template.provider.cache", "false");
-		properties.put("spring.mvc.log-resolved-exception", "true");
-		properties.put("server.servlet.jsp.init-parameters.development", "true");
-		properties.put("spring.reactor.stacktrace-mode.enabled", "true");
-		PROPERTIES = Collections.unmodifiableMap(properties);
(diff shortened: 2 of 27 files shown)


spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/jersey/JerseyEndpointResourceFactory.java
```diff
@@ -111,7 +111,7 @@ private Resource createEndpointLinksResource(String endpointPath,
 	private static final class EndpointInvokingInflector
 			implements Inflector<ContainerRequestContext, Object> {
 
-		private static final List<Function<Object, Object>> bodyConverters;
+		private static final List<Function<Object, Object>> BODY_CONVERTERS;
```
wilkinsona: We don't consider this to be a constant so the previous capitalisation was correct. Can you revert this part of the change please?

izeye: @wilkinsona Sure but I thought having both `static` and `final` is generally considered as a constant and  I don't get the difference from `ServletComponentRegisteringPostProcessor.HANDLERS`: https://github.com/spring-projects/spring-boot/blob/master/spring-boot/src/main/java/org/springframework/boot/web/servlet/ServletComponentRegisteringPostProcessor.java#L46

Could you explain a little bit more?

wilkinsona: While the list of body converters is unmodifiable, each of its entries may not be. Specifically, you could add something that implemented `Function<Object, Object>` to the list that could also be mutated.

Applying that reasoning, `ServletComponentRegisteringPostProcessor` is wrong and the field should be named `handlers` not `HANDLERS`.

izeye: @wilkinsona You mean a collection can't be a constant unless its element class is `final` and all fields in its element class are immutable like `String`. Do I understand correctly?

wilkinsona: You do indeed

izeye: @wilkinsona Thanks for the clarification 👍 

I can see a few places which are broken by the rule. Is it okay to handle them in this PR?

wilkinsona: Of course! That'd would be much appreciated

izeye:
@wilkinsona I applied the rule and squashed in ccd9afb but I'm not sure for the change because there are some places which look constants semantically although they are syntactically mutable.

Please let me know the change is okay for you.

spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/ShutdownEndpoint.java
```diff
@@ -38,11 +38,11 @@
 @Endpoint(id = "shutdown", defaultEnablement = DefaultEnablement.DISABLED)
 public class ShutdownEndpoint implements ApplicationContextAware {
 
-	private static final Map<String, Object> NO_CONTEXT_MESSAGE = Collections
+	private static final Map<String, Object> noContextMessage = Collections
```
wilkinsona: This could become a `Map<String, String>` and remain as `NO_CONTEXT_MESSAGE`.

spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/ShutdownEndpoint.java
```diff
@@ -38,11 +38,11 @@
...
 			.unmodifiableMap(Collections.<String, Object>singletonMap("message",
 					"No context to shutdown."));
 
-	private static final Map<String, Object> SHUTDOWN_MESSAGE = Collections
+	private static final Map<String, Object> shutdownMessage = Collections
```
wilkinsona: This could become a `Map<String, String>` and remain as `SHUTDOWN_MESSAGE`

spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsPropertyDefaultsPostProcessor.java
```diff
@@ -42,31 +42,31 @@
 @Order(Ordered.LOWEST_PRECEDENCE)
 public class DevToolsPropertyDefaultsPostProcessor implements EnvironmentPostProcessor {
 
-	private static final Map<String, Object> PROPERTIES;
+	private static final Map<String, Object> properties;
```
wilkinsona: This could become a `Map<String, String>` and remain as `PROPERTIES`.

izeye: @wilkinsona If the type is changed, it's not matched with the type being used: `MapPropertySource(String name, Map<String, Object> source)`

I guess you don't want to copy for it.

izeye:
@wilkinsona `NO_CONTEXT_MESSAGE` and `SHUTDOWN_MESSAGE` are restored with type change in b8b4d95

spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsPropertyDefaultsPostProcessor.java
wilkinsona: Sorry, you're right. It'll need to be changed to `properties` as you've proposed.

wilkinsona:
Some of this polish applies to 1.5.x

wilkinsona:
Thanks once again, @izeye. I've merged some of this into 1.5.x and merged that forwards adding the master-specific stuff as I did so. I took a little bit of work to separate it like that so I hope I haven't lost anything during the process.

philwebb:
I think we may have gone a little too far renaming the static finals. I think we should use the uppercase format to indicate that _we_ consider the item to be a constant. For example, we have `DevToolsPropertyDefaultsPostProcessor.properties` which is totally immutable, so even if the ultimate signature is `Map<String, Object>` I don't think that should drive the naming.

`ServletComponentRegisteringPostProcessor.handlers` is another example. Whilst you could ague that `ServletComponentHandler` could be mutable, in practice they are not. I'm going to revert a few of these renames.

PR #10494, opened 2017-10-02
Remove explicit type arguments

izeye:
This PR removes explicit type arguments which can be inferred.


wilkinsona:
Thanks, @izeye. Out of interest, how did you find these? It'd be nice to rely on inference where possible in 1.5.x as well and cherry picking this back might not be straightforward.

izeye:
@wilkinsona I just filtered them using `\.<[^/a-z]` regex first and checked them manually 😅

wilkinsona:
The equivalent changes don't work will on 1.5.x, I think due to different type inferencing in javac when it's targeting 1.6-compatible byte code. Let's just make these changes in master.

wilkinsona:
Thanks once again, @izeye. This change is now in master. Just FYI, I dropped the changes to the JOOQ-generated source as they'll get clobbered the next time the source is regenerated.

PR #10508, opened 2017-10-04
Upgrade to HikariCP 2.7.2

shakuzen:
2.7.2 was recently released with [these changes](https://github.com/brettwooldridge/HikariCP/blob/dev/CHANGES). Being close to the release target for the next milestone, I figured the semi-automated dependency upgrade process may have already been ran before this new version was released.


snicoll:
>  I figured the semi-automated dependency upgrade process may have already been ran 

I am not sure, we run it quite close from the release date (2 days from now). Anyway, let's use that PR for this one.

Issue #8511, opened 2017-03-06
Provide support for Kotlin KClass parameter in SpringApplication.run()


sdeleuze:
It would be nice to support natively Kotlin `KClass` parameter to allow writing `SpringApplication.run(FooApplication::class, *args)` in Spring Boot 2.0.

I could be implemented using classpath detection, would avoid `java.lang.IllegalArgumentException: Invalid source type class kotlin.reflect.jvm.internal.KClassImpl` error that Kotlin users get by default and would provide more idomatic Kotlin code than the current `SpringApplication.run(FooApplication::class.java, *args)`.

philwebb:
Kotlin support would align nicely with Spring 5.0

sdeleuze:
I have started to work on this.

sdeleuze:
[Pull request](https://github.com/spring-projects/spring-boot/pull/10511) submitted.

snicoll:
Closing in favour of PR #10511

PR #10511, opened 2017-10-04
Add runApplication() Kotlin top level function

sdeleuze:
Since Kotlin extensions do not apply to static methods, this commit introduces a `runApplication()` Kotlin top level function that acts as a Kotlin shortcut for `SpringApplication.run()`.

This shortcut avoids to require using non-idiomatic code like `SpringApplication.run(FooApplication::class.java)` and provides a `runApplication<FooApplication>()` alternative (as well as an array of `KClass` based alternative when multiple classes need to be passed as parameter).

This commit also adds Kotlin main artifacts to Spring Boot dependency management and will be replaced by Kotlin BOM when it will be available (see [KT-18398](https://youtrack.jetbrains.com/issue/KT-18398)).

`runApplication<FooApplication>()` is a form usually preferred in Kotlin to `runApplication(FooApplication::class)` and consistent with what we do on Spring Framework side. 

The custom Maven configuration is the [typical way to deal with Java + Kotlin project](https://kotlinlang.org/docs/reference/using-maven.html#compiling-kotlin-and-java-sources), a similar configuration has been applied to [Spring Data Kay](https://github.com/spring-projects/spring-data-build/blob/beb9d5cfdfdd18b0055db7648f859b92b0a168e5/parent/pom.xml#L775).

Closes gh-8511

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -3025,6 +3025,11 @@
 					<version>${flyway.version}</version>
 				</plugin>
 				<plugin>
+					<groupId>kotlin-maven-plugin</groupId>
+					<artifactId>org.jetbrains.kotlin</artifactId>
+					<version>${kotlin.version}</version>
+				</plugin>
+				<plugin>
 					<groupId>pl.project13.maven</groupId>
 					<artifactId>git-commit-id-plugin</artifactId>
 					<version>${git-commit-id-plugin.version}</version>
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -107,6 +107,7 @@
 		<junit.version>4.12</junit.version>
 		<junit-jupiter.version>5.0.1</junit-jupiter.version>
 		<junit-platform.version>1.0.1</junit-platform.version>
+		<kotlin.version>1.1.51</kotlin.version>
 		<lettuce.version>5.0.0.RELEASE</lettuce.version>
 		<liquibase.version>3.5.3</liquibase.version>
 		<log4j2.version>2.9.1</log4j2.version>
@@ -2263,6 +2264,26 @@
 				<groupId>org.jooq</groupId>
 				<artifactId>jooq-codegen</artifactId>
 				<version>${jooq.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-reflect</artifactId>
+				<version>${kotlin.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-stdlib</artifactId>
+				<version>${kotlin.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-stdlib-jre7</artifactId>
+				<version>${kotlin.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-stdlib-jre8</artifactId>
+				<version>${kotlin.version}</version>
 			</dependency>
 			<dependency>
 				<groupId>org.liquibase</groupId>
--- a/spring-boot-project/spring-boot/pom.xml
+++ b/spring-boot-project/spring-boot/pom.xml
@@ -13,6 +13,79 @@
 	<properties>
 		<main.basedir>${basedir}/../..</main.basedir>
 	</properties>
+	<build>
+		<plugins>
+			<plugin>
+				<artifactId>kotlin-maven-plugin</artifactId>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<version>${kotlin.version}</version>
+				<configuration>
+					<jvmTarget>${java.version}</jvmTarget>
+				</configuration>
+				<executions>
+					<execution>
+						<id>compile</id>
+						<phase>compile</phase>
+						<goals>
+							<goal>compile</goal>
+						</goals>
+						<configuration>
+							<sourceDirs>
+								<sourceDir>${project.basedir}/src/main/kotlin</sourceDir>
+								<sourceDir>${project.basedir}/src/main/java</sourceDir>
+							</sourceDirs>
+						</configuration>
+					</execution>
+					<execution>
+						<id>test-compile</id>
+						<phase>test-compile</phase>
+						<goals>
+							<goal>test-compile</goal>
+						</goals>
+						<configuration>
+							<sourceDirs>
+								<sourceDir>${project.basedir}/src/test/kotlin</sourceDir>
+								<sourceDir>${project.basedir}/src/test/java</sourceDir>
+							</sourceDirs>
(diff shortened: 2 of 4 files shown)


sdeleuze:
The build failed for an unrelated flacky WebFlux test, but `./mvnw clean install` works fine locally and I have validated the top level function is usable in my MiXiT sample application.

spring-boot-dependencies/pom.xml
```diff
@@ -2285,6 +2286,26 @@
 				<version>${jooq.version}</version>
 			</dependency>
 			<dependency>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-reflect</artifactId>
+				<version>${kotlin.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-stdlib</artifactId>
```
sureshg: Doesn't `kotlin-stdlib-jre8` transitively pull `kotlin-stdlib` and `kotlin-stdlib-jre7`?

spring-boot-dependencies/pom.xml
```diff
@@ -2285,6 +2286,26 @@
...
+				<version>${kotlin.version}</version>
+			</dependency>
+			<dependency>
+				<groupId>org.jetbrains.kotlin</groupId>
+				<artifactId>kotlin-stdlib-jre7</artifactId>
```
snicoll: We don't support Java 7 so that should go away.

spring-boot/pom.xml
```diff
@@ -18,6 +18,79 @@
...
+			<plugin>
+				<groupId>org.apache.maven.plugins</groupId>
+				<artifactId>maven-compiler-plugin</artifactId>
+				<configuration>
+					<source>${java.version}</source>
+					<target>${java.version}</target>
+					<parameters>true</parameters>
+				</configuration>
+				<executions>
+					<execution>
```
snicoll: What are those? Can you explain?

spring-boot/src/test/kotlin/org/springframework/boot/SpringApplicationExtensionsTests.kt
```diff
@@ -0,0 +1,73 @@
...
+
+	@Configuration
+	internal open class ExampleConfig
+
+	@Configuration
+	internal open class ExampleWebConfig {
+
+		@Bean
+		open fun webServer(): TomcatServletWebServerFactory {
+			return TomcatServletWebServerFactory(0)
```
snicoll: Is there a reason why you need to create a web server for this test? That looks a bit overkill to me.

spring-boot/pom.xml
snicoll: Scratch that. I didn't read your description properly, sorry.

spring-boot-dependencies/pom.xml
wilkinsona: I'm not sure I agree with that. I'd rather that the dependency management was complete.

snicoll: I am not following. When are we going to use that dependency if we don't support Java 7 anymore?

sdeleuze: We support Java 8 so kotlin-stdlib, kotlin-stdlib-jre7 and kotlin-stdlib-jre8 should be provided for consistent versioning. kotlin-stdlib-jre7 contains Java 7 specific methods which are needed for Java 8 support, same for kotlin-stdlib-jre8 and Java 8.

spring-boot/src/test/kotlin/org/springframework/boot/SpringApplicationExtensionsTests.kt
sdeleuze: Yes, because if not the test fails, I just copied what you do in SpringApplicationTests on Java side.

sdeleuze:
For Maven plugin configuration, it is the standard way to compile Java + Kotlin projects, I have on purpose only applied it to the module where we need that, we do the same in Spring Data without issues reported, and many other project do that. If you face issue, I can ask Kotlin team to fixit, if not I would recommand go that way since I have no alternative to propose.

sdeleuze:
Hum, there is already a `kotlin.run` top level function, so reusing that name for Boot may cause confusion. I will update the PR with another function name.

sdeleuze:
I have splitted this PR in 2 commits, one for dependency management, the other for the extension itself. I have also renamed the top level function to `runApplication()` which seems to me a good trade off between `run()` which will cause confusion with `kotlin.run()` and looks reasonably short (`runSpringApplication()` looks ugly for such widely used function).

sdeleuze:
I have also rebased this PR on top of the recent module structure changes.

GreyTeardrop:
Hi! Awesome to see even more Kotlin niceties in Boot!

@sdeleuze,
Do you think it could make sense to also add something like this:

```kotlin
inline fun <reified T : Any> runApplication(vararg args: String,
                                            block: SpringApplication.() -> Unit): ConfigurableApplicationContext =
    SpringApplication(T::class.java).apply(block).run(*args)
```

for cases when some `SpringApplication` tuning is required?

sdeleuze:
@GreyTeardrop Great idea, this makes this feature much more flexible.

@snicoll I have updated the PR with this proposal and added related tests. I have also created the related https://github.com/spring-projects/spring-boot/issues/10712 issue.

sdeleuze:
I have added plugin management for Maven plugin, after discussion with @wilkinsona sadly we can't do similar things with Gradle plugin. Unfortunate but IMO we can live with that, it is possible to have different minor versions between dependencies and plugin.

snicoll:
@wilkinsona this would require a Spring Boot version to Kotlin version mapping in start.spring.io for Gradle projecs and I am really not keen to add that. 

Would it be possible to extrac the `kotlin.version` from the BOM and use that as a property to define the version of the plugin to use? Or do something of the kind in our own dependency management plugin?

wilkinsona:
I'd have to investigate, but right now I don't know how to do it, particularly with the `plugins` DSL. We can only extract the `kotlin.version` from the bom once the Spring Boot and Dependency Management plugins have been applied. By that point, the Kotlin plugin may already have been applied too. I think there's a chicken and egg problem here.

wilkinsona:
We avoid the problem for the Dependency Management plugin by making the Spring Boot plugin pull it in transitively. That lets us control the version of the Dependency Management plugin and allows a user to just apply the plugin without worrying about its version. We could do the same for the Kotlin plugin, but then everyone building a Gradle project would download the Kotlin plugin even if they weren't using it. That's similar to what we did for Jackson's Kotlin module so we could decide that it's a reasonable thing to do, probably depending on the weight of the Kotlin plugin and its dependencies.

philwebb:
I'm guessing 90% of applications are going to do something like this:

```kotlin
fun main(args : Array<String>) { 
    runApplication<ExampleApp>(args)
}
```

I'm wondering if we shouldn't drop the multiple config version? Doing this:

```kotlin
runSpringApplication(arrayOf(ExampleConfig::class, ExampleWebConfig::class), args)
```

Doesn't seem much better than:

```kotlin
SpringApplication(ExampleConfig::class, ExampleWebConfig::class).run(args)
```

@sdeleuze Are there any benefits to the `arrayOf` version that I'm missing?

sdeleuze:
@philwebb Good point, indeed I think we can remove the `Array` variant.

philwebb:
I'll take care of that

Issue #10436, opened 2017-09-27
Very generic bean name "clock" in 2.0.0.M4


fprochazka:
Hi,
I've upgraded our app from `2.0.0.M3` to `2.0.0.M4` and it started reporting that bean of type `java.time.Clock` was not found and that I should define it. But I have it defined as shown in the following snippet and it worked correctly in `2.0.0.M3`

```
import java.time.Clock;

@Configuration
public class ServicesConfiguration
{
    @Bean
    public Clock clock()
    {
        return Clock.systemDefaultZone();
    }
}
```

It stopped working, because `org.springframework.boot.actuate.autoconfigure.metrics.export.simple.SimpleExportConfiguration` in `2.0.0.M4` defines bean with very generic name `clock` that replaced my definition. Took me a while to figure it out, but enabling debug logs of `org.springframework` helped.

I've already renamed my service, so it's not an issue for me anymore. But `clock` is a very generic name and I think it should be changed to something more specific.

wilkinsona:
There are several other Micrometer configuration classes that declare a `Clock` bean named `clock`.

fprochazka:
Thank you! :)

PR #10519, opened 2017-10-05
Updates for Micrometer 1.0.0-rc.2

jkschneider:
Fixes and improvements made for rc.2 are listed [here](https://github.com/micrometer-metrics/micrometer/milestone/13?closed=1). Significantly, the gap on cache metrics has been closed.

The impact on Boot 2 is:

1. A new autoconfig for the StatsD registry.
2. A new protocol property for Graphite.
3. A reworked metrics actuator endpoint.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdExportConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdExportConfiguration.java
@@ -1,4 +1,24 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
 package org.springframework.boot.actuate.autoconfigure.metrics.export.statsd;
+
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.statsd.StatsdConfig;
+import io.micrometer.statsd.StatsdMeterRegistry;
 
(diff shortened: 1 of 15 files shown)


spring-boot-project/spring-boot-dependencies/pom.xml
```diff
@@ -113,7 +113,7 @@
 		<logback.version>1.2.3</logback.version>
 		<lombok.version>1.16.18</lombok.version>
 		<mariadb.version>2.1.2</mariadb.version>
-		<micrometer.version>1.0.0-rc.1</micrometer.version>
+		<micrometer.version>1.0.0-SNAPSHOT</micrometer.version>
```
jkschneider: If you are satisfied with what's been done in the rest of this PR, I can go ahead and release 1.0.0-rc.2

jkschneider:
cc/ @philwebb @wilkinsona

jkschneider:
Sample of the updated metrics endpoint:

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdExportConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+
+	@Bean
+	@ConditionalOnProperty(value = "spring.metrics.statsd.enabled", matchIfMissing = true)
+	public MetricsExporter statsdExporter(StatsdConfig statsdConfig, Clock clock) {
+		return () -> new StatsdMeterRegistry(statsdConfig, clock);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public Clock clock() {
```
shakuzen: See #10436

wilkinsona: 👍 I'll take care of that one as part of merging this

wilkinsona:
This looks good to me.

I've split it up into separate commits for each change here: https://github.com/wilkinsona/spring-boot/tree/gh-10519. The change the enables the metrics endpoint by default isn't there at the moment as I think it warrants separate consideration. I've opened #10525.

The only piece I'm not sure about is the removal of the requested tags from the available tags in the metrics endpoint:

```java
tags.forEach(t -> availableTags.remove(t.getKey()));
```

It feels a little odd to me to successfully use a particular tag as a query and then to not see that tag listed in the available tags for the metric. I wonder if this line should be removed and the tests updated accordingly?

wilkinsona:
Aah, I see. Thanks for the explanation. That makes sense. Let's leave it as it is.

PR #10533, opened 2017-10-06
fix build status image

WilderPereira:
The image of the Build Status wasn't being displayed in the README file, so I opened up the other repositories that has the build status image correctly displayed and followed the url pattern.


snicoll:
Thank you! This allows me to notice the badge is outdated as we've move to Concourse 2 days ago. I've created #10536 that will completely change the badge definition.

PR #10535, opened 2017-10-06
Expose metrics hiding recursively inside a composite registry

jkschneider:
This supports an unusual case such as the one found in `spring-cloud-stream` where registries that are members of a composite registry have metrics that are not found in the composite.

For `spring-cloud-stream`, an application may be ran with a parent and several child application contexts. The intent is to inject a registry into each child application context which will be configured with a common tag representing the "app" that it represents. 

cc / @viniciusccarvalho 
cc / @wilkinsona


philwebb:
@jkschneider I've merged this and also [polished up](https://github.com/spring-projects/spring-boot/commit/22a6ee03ebe9326364c3c79318942dc0694189f3) `MetricsEndpoint` a little. Let me know if think I've broken anything.

Issue #8690, opened 2017-03-22
Missing "url" property for DataSourceProperties will cause NullPointerException if embedded DB is not used


simonwg:
The issued failed with Spring Boot v1.5.2.

In my case, I am trying to create 2 `DataSource` (not any Embedded Database), Some configuration settings are common (e.g. Connection Pool type,  Driver name) and specific configuration (such as url, username, password etc) will be located in different configuration keys.

e.g.
```
application.yml

app:
    datasource:
        type: com.zaxxer.hikari.HikariDataSource
        driverClassName: com.mysql.jdbc.Driver
        db1:
            url: jdbc:mysql://localhost:3306/myDB1
            username: user1
            password: password1
        db2:
            url: jdbc:mysql://localhost:3306/myDB2
            username: user2
            password: password2
```

```
@Configuration
public class RepositoryConfig {

    @Bean
    @Primary
    @ConfigurationProperties("app.datasource")
    public DataSourceProperties dataSourceProperties() {
        return new DataSourceProperties();
    }

    @Bean
    @Primary
    @ConfigurationProperties("app.datasource.db1")
    public HikariDataSource db1DataSource(DataSourceProperties properties) {
        return (HikariDataSource) dataSourceProperties().initializeDataSourceBuilder()
                .type(HikariDataSource.class).build();
    }

    @Bean
    @ConfigurationProperties("app.datasource.db2")
    public HikariDataSource db2DataSource(DataSourceProperties properties) {
        return (HikariDataSource) dataSourceProperties().initializeDataSourceBuilder()
                .type(HikariDataSource.class).build();
    }
}
```

When I start with the above configuration, `NullPointerException` will be thrown.

```
Caused by: java.lang.NullPointerException: null
	at java.util.regex.Matcher.getTextLength(Matcher.java:1283) ~[na:1.8.0_121]
	at java.util.regex.Matcher.reset(Matcher.java:309) ~[na:1.8.0_121]
	at java.util.regex.Matcher.<init>(Matcher.java:229) ~[na:1.8.0_121]
	at java.util.regex.Pattern.matcher(Pattern.java:1093) ~[na:1.8.0_121]
	at java.util.Formatter.parse(Formatter.java:2547) ~[na:1.8.0_121]
	at java.util.Formatter.format(Formatter.java:2501) ~[na:1.8.0_121]
	at java.util.Formatter.format(Formatter.java:2455) ~[na:1.8.0_121]
[...]

PR #10546, opened 2017-10-06
Fix NPE if no database url is given

tinexw:
Fixes gh-8690


tinexw:
Will reopen after travis is green again

tinexw:
Closed because I wanted to squash the commits, will create new one

PR #10548, opened 2017-10-07
Samples cleanup

isopov:
- Modifying dependencies to starter-web with tomcat exclusion plus alternative
servlet container instead of manual dependency on spring-webmvc as it is the
preferrable way to use alternative servlet container
- Previously RestTemplate with ssl was configured manually in tests - now
it rellies on autoconfiguration - changed this for multi-connector test
and added test to ensure that ssl autoconfiguration is working
- Most samples with alterntative servlet containers used
some kind of service reading property and returning default since it
wasn't configured - removed it, since it is not specific to using alternative
servlet containers.
- Also `@DirtiesContext` removed from tests since it seems to be not needed
and to promote that its use is discouraged.


pivotal-issuemaster:
@isopov Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10548)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10548) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@isopov Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10548)!

isopov:
I removed rest of DirtiesContext from samples for the same reason and placed tham removal in separate commit in case there will be some problems (but 2 or 3 DirtiesContext removals are still in first commit)

snicoll:
@isopov thank you very much!

wilkinsona:
Unfortunately, this has broken the Actuator sample. The failure appears to depend on the order in which the tests classes are run. It can be reproduced with the following suite:

```java
@RunWith(Suite.class)
@SuiteClasses({ sample.actuator.EndpointsPropertiesSampleActuatorApplicationTests.class,
		sample.actuator.ShutdownSampleActuatorApplicationTests.class,
		sample.actuator.ServletPathSampleActuatorApplicationTests.class,
		sample.actuator.ManagementPortSampleActuatorApplicationTests.class,
		sample.actuator.SampleActuatorApplicationTests.class,
		sample.actuator.NoManagementSampleActuatorApplicationTests.class,
		sample.actuator.ManagementPathSampleActuatorApplicationTests.class,
		sample.actuator.ManagementPortAndPathSampleActuatorApplicationTests.class,
		sample.actuator.ManagementAddressActuatorApplicationTests.class })
public class AdhocSuite {

}
```

snicoll:
Thanks @wilkinsona - I saw the build failure but couldn't reproduce it locally.

PR #10549, opened 2017-10-07
marking UndertowServletWebServer as started before start since failed…

### LINKED ISSUE #10528: Undertow doesn't shut down properly if there is a startup error (e.g., port already in use)
Working with Spring Boot 1.5.6 (but glancing at the code I think it might still be a problem even on master).
If I start an app that uses Undertow, and startup fails due to an error such as port in use, then the app is left running. The reason is that Undertow.start is called, and it starts up non-daemon XNIO threads before throwing the exception due to port in use. This exception is caught in UndertowEmbeddedServletContainer.start, however the member variable started has not been set to true since startup failed. Then, UndertowEmbeddedServletContainer.stop is called later in response to the startup failure, but it does nothing since the startup variable is false. It seems calling Undertow.stop() in the exception handler would cause the XNIO threads to be stopped and therefore allow the application to exit.

isopov:
… exceptional start may still leave it started

Fixes #10528

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/undertow/UndertowServletWebServer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/undertow/UndertowServletWebServer.java
@@ -87,7 +87,8 @@ public class UndertowServletWebServer implements WebServer {
 
 	private Undertow undertow;
 
-	private volatile boolean started = false;
+	// visible for testing
+	volatile boolean started = false;
 
 	/**
 	 * Create a new {@link UndertowServletWebServer} instance.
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR, but I'd prefer to fix this by stopping Undertow in `start()` if starting it fails. This would match what we currently do in `TomcatWebServer`.

ajinkya-vaze:
I do not think anything needs to be done in spring boot code to fix this. The threads are left open by Undertow and it was a defect in undertow core. I opened an issue https://issues.jboss.org/browse/UNDERTOW-1180 for this and it has been fixed. Upgrading to Undertow 1.4.21.Final should automatically fix this.

Issue #7927, opened 2017-01-09
Auto configure SLF4J MDC with a servlet filter


bbrouwer:
I find it useful to automatically configure the Mapped Diagnostic Context in slf4j to include
- source of the web request
- user id
- requested URI

There could be plenty of other things worth capturing, but these are the ones I like to capture by default. What I've done is made a servlet filter that is registered by an `@Configuration` class registered in `META-INF/spring.factories` as a FilterRegistrationBean. 

What the filter does is use SpEL from the ServletRequest to retrieve these values:
- src=`getHeader('X-Forwarded-For') ?: getRemoteHost()`
- uid=`userPrincipal?.name`
- uri=`requestURI`

Then, all you need to do is have a logback pattern that includes the MDC and you easily get these added to the log. Is there interest in adding this to Spring Boot? I can submit the code if there is interest. Of course it would be nice to have some properties to alternatively configure it in `application.properties`. Maybe like this:

```
logging.mdc.src=getHeader('X-Forwarded-For') ?: getRemoteHost()
logging.mdc.uid=userPrincipal?.name
logging.mdc.uri=requestURI
# to show custom user values
logging.mdc.locale=locale.displayName
```

philwebb:
Thanks for the suggestion but we need to be a bit careful here that we don't end up implementing programming by properties. We've generally tried to avoid adding SpEL expression support in properties files whenever possible.

I get the feeling that setting MDC attributes is something that tends to be quite application specific and not something that we should necessarily add to Boot itself. Mostly applications should be able to add MDC attributes easily by doing something like this:

```java
@Component 
public static class MdcFilter extends OncePerRequestFilter() {

	@Override
	protected abstract void doFilterInternal(HttpServletRequest request,
			HttpServletResponse response, FilterChain filterChain)
					throws ServletException, IOException {
		MDC.put("src", request.getRemoteHost());
		// ...
	}

}
```

bbrouwer:
I'm fine if SpEL is not the way to go, but do you see value in injecting those 3 attributes into the MDC automatically for web apps? They have proven to be very valuable to me, and honestly nobody has ever gone away from those defaults in my organization. Plus, you can configure the logback log pattern to only include the MDC variables that you choose, or include them all, which is what I do. It can be a nice way to easily get useful information into your logs without really thinking much about it.

philwebb:
[...]

PR #10562, opened 2017-10-09
Add source, user and uri to slf4j MDC

bbrouwer:
This is the beginning of my solution for #7927. It involves 2 filters: one registered as early as possible to get the src and uri. One is registered after the spring security filter to grab the uid. 

The first filter also can grab the uid if one exists. I figured that way if this is deployed to an app server where user authentication is taken care of by the app server, this will grab the uid right away and make it available to the MDC. I'm actually having some difficulty writing an integration test for this because I can't seem to get the embedded Tomcat instance to setup security through Tomcat, rather than Spring Security.

I still need to write the automated tests for this. I also don't have a property yet to disable these filters. I'm looking for some feedback on where to go from here.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/SecurityProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/SecurityProperties.java
@@ -17,8 +17,10 @@
 package org.springframework.boot.autoconfigure.security;
 
 import java.util.Arrays;
+import java.util.EnumSet;
 import java.util.HashSet;
 import java.util.Set;
+import java.util.stream.Collectors;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.web.servlet.DispatcherType;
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/SecurityProperties.java
```diff
@@ -35,26 +37,27 @@
 public class SecurityProperties implements SecurityPrerequisite {
 
 	/**
-	 * Order applied to the WebSecurityConfigurerAdapter that is used to configure basic
```
bbrouwer: I'll revert these unnecessary changes.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/RequestSlf4jMdcAutoConfiguration.java
```diff
@@ -0,0 +1,125 @@
...
+		public void init(final FilterConfig filterConfig) throws ServletException {
+			// nothing to configure
+		}
+
+		@Override
+		public void doFilter(final ServletRequest request, final ServletResponse response, final FilterChain chain)
+				throws IOException, ServletException {
+			try (Closeable src = putMDC("src", RequestSlf4jMdcFilter::src, request)) {
+				try (Closeable uri = putMDC("uri", RequestSlf4jMdcFilter::uri, request)) {
+					try (Closeable uid = putMDC("uid", RequestSlf4jMdcFilter::uid, request)) {
```
bbrouwer: The uid is really only useful if the container is providing authentication. For most Spring Boot apps, I'm guessing the uid will not yet be set, and therefore this will be wasted effort.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/RequestSlf4jMdcAutoConfiguration.java
```diff
@@ -0,0 +1,125 @@
...
+						chain.doFilter(request, response);
+					}
+				}
+			}
+		}
+
+		private static String src(HttpServletRequest request) {
+			final String xff = request.getHeader("X-Forwarded-For");
+			return StringUtils.hasLength(xff) ? xff : request.getRemoteHost();
```
bbrouwer: If X-Forwarded-For is set, should I also append the remote host? If you put your application behind a load balancer, the remote host is going the be the address of the load balancer, which is not terribly useful. This can easily be mitigated in the load balancer by stripping out X-Forwaded-For headers from clients it does not trust. 

But if your application is not behind a load balancer that strips out malicious X-Forwarded-For headers, an attacker might be able to cover their tracks by injecting a fake X-Forwarded-For header. 

One solution would be this:
```
return StringUtils.hasLength(xff) ? xff + ", " + request.getRemoteHost() : request.getRemoteHost();
```
Alternatively, we could just put both in the MDC: both the X-Forwarded-For header and the remote host. Then apps could decide which of the MDC fields they want to add to their logging simply by how they configure logback (for example).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/RequestSlf4jMdcAutoConfiguration.java
```diff
@@ -0,0 +1,125 @@
...
+
+		private static Closeable putMDC(String key, Function<HttpServletRequest, String> value,
+				ServletRequest request) {
+			if (request instanceof HttpServletRequest) {
+				try {
+					final String val = value.apply((HttpServletRequest) request);
+					if (StringUtils.hasLength(val)) {
+						return MDC.putCloseable(key, val);
+					}
+				} catch (Exception e) {
```
bbrouwer: This try/catch is there to make sure this filter has as little impact as possible to a request. If, for example, we can't retrieve the X-Forwarded-For header and an exception is thrown, I don't want the request to fail.

philwebb:
After some consideration we've decided it would be best not add this into the core of Boot. I think it might make a nice 3rd-party starter, and we'll happily add a link to that if you decide to spin out a separate project. Thanks!

PR #10586, opened 2017-10-10
Replace equals() and hashCode() calls on URL objects

dreis2211:
Hey,

this PR replaces calls to `equals()` and `hashCode()` on URL objects. While this mostly means accessing local networks they can be still more unreliable than ones without netwerk access at all.

What do you think? I'm a bit conceirned about the additional allocations in  the `Handler` class when converting the URL to a URI. Is that a critical path?

Cheers,
Christoph


wilkinsona:
Thanks for the PR. I'm not sure that these changes are warranted, particularly as, in the vast majority of cases, the URLs involved will point to a resource on the local file system. Can you please describe in a bit more detail what prompted you to look into this and propose the changes?

dreis2211:
Hey,
I just noticed this in `PropertiesLauncher` and looked for other occurences. Maybe it's just me, but I always try to avoid calling equals() or hashCode() on URLs, because it still does host lookups. Even though it's localhost in those cases, I personally tend to avoid this.

```

wilkinsona:
Thanks for the explanation. Given that we haven't seen `equals()` or `hashCode()` on URL instances appear as hot spots in any profiling that we've done, I don't think this change can be justified. Thanks anyway.

Issue #10568, opened 2017-10-09
Include error body on Bad HTTP error code on wrong logging level


philwebb:
See issue #8798 and PR #8813 and [this comment](https://github.com/spring-projects/spring-boot/issues/8798#issuecomment-335147964):

>... the HTTP 400 response does not contain a body anymore which makes it impossible to understand the reason why the request has failed.
>
> IMO, the error response should contain a reason for the error, similar to the message of the initial HTTP 500 error.

eddumelendez:
@philwebb should it be based on branch `1.5.x` or `master`?

philwebb:
Closing in favor of PR #10588. Thanks again @eddumelendez !!

wilkinsona:
I've opened https://github.com/spring-projects/spring-boot/issues/10618 to do something more general purpose in the web endpoint infrastructure for 2.0.

PR #10588, opened 2017-10-10
Add proper message when wrong log level is sent

eddumelendez:
Previously, bad request with no reason was included in the response.
This commit introduces the reason when invalid log level is sent in the
request.

See gh-10568


wilkinsona:
I've opened #10618 to do something more general purpose in the web endpoint infrastructure for 2.0.

UtsaVakani:
Hello,
I am using 2.1.3.RELEASE version and still getting 500 on giving improper configuredLevel in POST payload for /loggers endpoint ...

wilkinsona:
@UtsaVakani You should get a 400 for an unrecognised `configuredLevel`. We have an integration test that verifies that this is the case:

https://github.com/spring-projects/spring-boot/blob/f15e04b3926f20c01e9ae986a56eac43b8cdebf5/spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/logging/LoggersEndpointWebIntegrationTests.java#L122-L129

If you are seeing different behaviour, please open a new issue with a small sample project that reproduces the 500 response.

PR #10612, opened 2017-10-12
Polish JobLauncherCommandLineRunner

mdeinum:
The JobLauncherCommandLineRunner reassigned method arguments inside the
merge method. Instead of reassigning the method argument just return
the new instance directly.

Also some minor cleanup (unnesseccary null check and continue keyword).


mminella:
We'll have that moved in the next milestone of batch (we're going to do the work next week).

PR #10615, opened 2017-10-12
Fix Security OAuth2 Client documentation

vpavic:
This PR fixes incorrect section level and sample configuration properties.

I've also included a second commit that changes the format of documented configuration example from YAML to properties as that seems both more readable and consistent with the rest of the documentation.


snicoll:
@vpavic thanks!

PR #10616, opened 2017-10-12
Fix broken links

izeye:
This PR fixes more broken links missed in https://github.com/spring-projects/spring-boot/pull/10601.


snicoll:
Thanks again @izeye

PR #10626, opened 2017-10-12
Fix NPE if no database url is given and no embedded database is available

tinexw:
Fixes gh-8690

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
@@ -54,7 +54,7 @@ import org.springframework.util.StringUtils;
  */
 @ConfigurationProperties(prefix = "spring.datasource")
 public class DataSourceProperties
-	implements BeanClassLoaderAware, EnvironmentAware, InitializingBean {
+		implements BeanClassLoaderAware, EnvironmentAware, InitializingBean {
 
 	private static final Log logger = LogFactory.getLog(DataSourceProperties.class);
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
@@ -24,6 +24,9 @@ import java.util.UUID;
 
 import javax.sql.DataSource;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.beans.factory.BeanClassLoaderAware;
 import org.springframework.beans.factory.BeanCreationException;
 import org.springframework.beans.factory.InitializingBean;
@@ -46,11 +49,14 @@ import org.springframework.util.StringUtils;
  * @author Stephane Nicoll
  * @author Benedikt Ritter
  * @author Eddú Meléndez
+ * @author Kristine Jetzke
  * @since 1.1.0
  */
 @ConfigurationProperties(prefix = "spring.datasource")
 public class DataSourceProperties
-		implements BeanClassLoaderAware, EnvironmentAware, InitializingBean {
+	implements BeanClassLoaderAware, EnvironmentAware, InitializingBean {
+
+	private static final Log logger = LogFactory.getLog(DataSourceProperties.class);
 
 	private ClassLoader classLoader;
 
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
```diff
@@ -288,6 +294,8 @@ public String determineUrl() {
 		if (StringUtils.hasText(this.url)) {
 			return this.url;
 		}
+
+		logger.debug("No url defined, looking for embedded database.");
```
snicoll: I don't think adding a logging statement here is part of what is discussed in the issue you've referenced.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
```diff
@@ -46,11 +49,14 @@
  * @author Stephane Nicoll
  * @author Benedikt Ritter
  * @author Eddú Meléndez
+ * @author Kristine Jetzke
  * @since 1.1.0
  */
 @ConfigurationProperties(prefix = "spring.datasource")
 public class DataSourceProperties
-		implements BeanClassLoaderAware, EnvironmentAware, InitializingBean {
+	implements BeanClassLoaderAware, EnvironmentAware, InitializingBean {
```
snicoll: Can you please fix this formatting change?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/EmbeddedDatabaseConnection.java
```diff
@@ -106,7 +107,7 @@ public String getUrl() {
 	 */
 	public String getUrl(String databaseName) {
 		Assert.hasText(databaseName, "DatabaseName must not be null.");
-		return String.format(this.url, databaseName);
+		return this.url != null ? String.format(this.url, databaseName) : null;
```
snicoll: That wasn't the intention as raised on #8690 - The idea is to avoid a NPE by an abuse of `DataSourceProperties` (as explained in the comments). I'd rather have an `IllegalStateException` here that states the url should not be null or something.

Thoughts?

tinexw: I think the result for the user would basically the same. 
With this implementation they get an ``DataSourceBeanCreationException `` with the message _Cannot determine embedded database url_
If I change it to your proposal they get an ``IllegalStateException``with a message that states the url should not be null or something.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
tinexw: My idea here was to give more insight into what's happening in the case described in the issue. As far as I understood it is hard to find out why there is this error because of the "magic" happening with with embedded database (i.e. no url means: try to connect to an embedded database) 

But the original issue was about something different anyway, I was trying to do the 
> quote Let's fix the error message though

So I can remove the logging statement if you want.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
tinexw: Done

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/EmbeddedDatabaseConnection.java
snicoll: The difference is that you give a perfectly valid exception message for a use case that isn't. I don't know if you've picked up that issue or if you faced it yourself but the issue as described doesn't result from a legit use case. I think we should then treat it as such.

tinexw: But it is a legit use case to define DataSourceProperties without a url - at least in the current implementation. Because then this fallback to the embedded database takes place (in my opinion it would be better to not have this magic at all but it’s there already so I don’t think it can/should be changed and is definitely not in scope of this issue.)
The error case from the issue is that no url is defined *and* no embedded database is available. 
I just think the exception should be thrown from the DataSourceProperties and not from the EmbeddedDatabaseConnection because if you don’t even add an embedded database, why should you get an error that originates in an EmbeddedDatabaseConnection class?
But maybe I misunderstand this completely...?

snicoll: > But maybe I misunderstand this completely...?
 
Or I do. Let me run that thing concretely, thanks for the feedback.

Issue #6733, opened 2016-08-23
Provide access to the root URI from TestRestTemplate


wilkinsona:
It'd be useful for asserting that `Location` headers are correct, for example. As things stand you have to assume that it'll be `http://localhost` and then use `@LocalServerPort` to inject the port.

wilkinsona:
Duplicate of #10641

PR #10641, opened 2017-10-14
Provide access to root URI from TestRestTemplate

tinexw:
Fixes gh-6733

Wasn't sure what to do in case the URI handler is not a ``RootUriTemplateHandler``. Decided to return an empty string but maybe it's better to throw an exception? Let me know what you think.


wilkinsona:
Thanks for the PR. Returning an empty String sounds good to me.

wilkinsona:
Thanks again, @tinexw. These changes have been merged into master.

Issue #10435, opened 2017-09-27
Drop auto-configured security for H2 console and DevTools remote start


wilkinsona:
Knowing that the H2 console isn't enabled by default and 
following on from the simplification of security auto-configuration, trying to auto-configure its security no longer makes sense. We should drop it and users who choose to enable the console in an environment where security is required should secure it appropriately.

mbhave:
`RemoteRestartWebSecurityConfigurer` in `RemoteDevToolsAutoConfiguration` can also go.

heroInCommunity:
I still need to add header "X-Frame-Options: SAMEORIGIN" when using Spring Boot v2.0.0.RELEASE to make frames working (I see message "Sorry, Lynx not supported yet" as it is described in #7198).
Could you please confirm that fix was included in version 2.0.0.M5 ?

wilkinsona:
@heroInCommunity Please see the [description of this issue](https://github.com/spring-projects/spring-boot/issues/10435#issue-261011295). What you have described is the expected behaviour and, if you really need to use the H2 console and have security enabled, you should configure it to meet your needs.

heroInCommunity:
@wilkinsona , my fault, sorry for disturbing and thank you.

Issue #10264, opened 2017-09-12
Decide how to deal with Jolokia and security


philwebb:
It's not an endpoint so the `EndpointRequest` builder won't cut it. Perhaps just manually defining the pattern is best?

wilkinsona:
We should give some thought to securing third-party servlets that are registered in the management context as well. Spring Cloud is likely to have to move to this model for part of their Hystrix support (see #10257).

mbhave:
maybe something for the h2 console too? there is a `spring.h2.console` path property so manually defining the pattern would duplicate it.

snicoll:
Yes, H2 console and Jolokia are on the same boat (although one is an actuator concern and one is not there atm).

philwebb:
@joshiste That's an interesting question. I think they should be. The problem is that both of these endpoints are MVC specific so it won't make sense to show them in a Jersey/WebFlux application.

snicoll:
Perhaps we want to focus only on H2 console (that truly isn't an endpoint and shouldn't be)? We can migrate jolokia back to a controller once #10257 is implemented.

philwebb:
The new `@EndpointController` is now an option for Jolokia but it isn't a very good one in my opinion since it won't work with Jersey. I think we need another type of endpoint that can be used to adapt Servlets and will work with both Spring MVC and Jersey.

philwebb:
I'm not sure that I agree that the H2 console isn't/shouldn't be an endpoint. I suppose it's slightly unusual in that it provides a rich HTML page rather than JSON.

snicoll:
I think adding the H2 console in this list is moving towards that anything  can be an endpoint and I don't think that's a good idea.

philwebb:
Yeah, I think that's probably true in retrospect. Let's make this issue about Jolokia and create a new one for H2.

philwebb:
Created #11704 for H2.

PR #10657, opened 2017-10-16
Remove "Securing the H2 console" section

izeye:
Auto-configured security for H2 console has been dropped in #10435 but its documentation still remains. This PR removes it.


snicoll:
Thanks @izeye!

See also #10264

PR #10663, opened 2017-10-17
Change HibernateJpaConfiguration to public to be able to extend it

damienbeaufils:
Hello,

Since [this commit](https://github.com/spring-projects/spring-boot/commit/bd02edf2ce749b70dcc45563bbeec2ab503d875c), we can't extend anymore `HibernateJpaAutoConfiguration` class to specify hibernate mapping files [like this](https://stackoverflow.com/questions/32062828/spring-boot-load-orm-xml/40060677#40060677), but we can’t either extend `HibernateJpaConfiguration` class because it is package-protected.
This PR changes `HibernateJpaConfiguration` class modifier to `public` to be able to extend it. 
It could be a solution for [this question on SO](https://stackoverflow.com/questions/46784760/how-to-specifiy-hibernate-mappings-with-spring-boot-2-0).

Thanks!

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaConfiguration.java
@@ -50,7 +50,7 @@ import org.springframework.util.ClassUtils;
  */
 @Configuration
 @ConditionalOnSingleCandidate(DataSource.class)
-class HibernateJpaConfiguration extends JpaBaseConfiguration {
+public class HibernateJpaConfiguration extends JpaBaseConfiguration {
 
 	private static final Log logger = LogFactory.getLog(HibernateJpaConfiguration.class);
 


pivotal-issuemaster:
@damienbeaufils Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10663)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10663) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@damienbeaufils Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=10663)!

snicoll:
@damienbeaufils extending from an auto-configuration isn't really a great idea. Can we take a step back and discuss why you need to do this? Perhaps we can adapt the auto-configuration to take that use case into account.

damienbeaufils:
@snicoll I agree extending from an autoconfiguration is not great.
My use case is quite simple I think: I would like to configure the Hibernate JPA datasource provided by Spring Boot with specific XML mapping files. 
The way people found is to override the `LocalContainerEntityManagerFactoryBean` bean from the old `HibernateJpaAutoConfiguration` and set mapping resources.

snicoll:
Can you show an example of such override please?

damienbeaufils:
Sure. An example:
```
@Configuration
public class MappingsConfiguration extends HibernateJpaAutoConfiguration {

    public MappingsConfiguration(DataSource dataSource, JpaProperties jpaProperties, ObjectProvider<JtaTransactionManager> jtaTransactionManager, ObjectProvider<TransactionManagerCustomizers> transactionManagerCustomizers, ObjectProvider<List<SchemaManagementProvider>> providers) {
        super(dataSource, jpaProperties, jtaTransactionManager, transactionManagerCustomizers, providers);
    }

    @Bean
    @Override
    public LocalContainerEntityManagerFactoryBean entityManagerFactory(EntityManagerFactoryBuilder factoryBuilder) {
        LocalContainerEntityManagerFactoryBean localContainerEntityManagerFactoryBean = getSuperEntityManagerFactory(factoryBuilder);
        localContainerEntityManagerFactoryBean.setMappingResources(
                "db/mappings/dummy.xml"
        );
        return localContainerEntityManagerFactoryBean;
    }

    LocalContainerEntityManagerFactoryBean getSuperEntityManagerFactory(EntityManagerFactoryBuilder factoryBuilder) {
        return super.entityManagerFactory(factoryBuilder);
    }
}
```
If you need a whole project using this code, see [this repository](https://github.com/damienbeaufils/clean-architecture-demo).

wilkinsona:
This feels like the sort of thing that we've previously solved with a customizer interface. If we consider this to be too much of an edge case to warrant that, a bean post-processor could be used instead.

damienbeaufils:
@wilkinsona Do you have an example of a customizer interface somewhere?

wilkinsona:
[`JacksonObjectMappingBuilderCustomizer`](https://github.com/spring-projects/spring-boot/blob/14081c8557fc3b8ee6958e070736ba49e01413d7/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/Jackson2ObjectMapperBuilderCustomizer.java`) is one example. It's used by `JacksonAutoConfiguration`.

snicoll:
perhaps `mappingResources` could be a `spring.jpa` property somehow. That would remove that need altogether (although I agree the customizer is typically what we do in such cases).

wilkinsona:
I like that idea, @snicoll. The method just takes `String…` so a `List<String>` `spring.jpa.mapping-resources` property would seem to fit nicely.

snicoll:
Cool. 

@damienbeaufils I am not keen to change the scope of that class (extending the `AutoConfiguration` is kind of a smell anyway). Let's add support for this in #10684. If you find other use cases where extending is necessary we can go the customizer route.

damienbeaufils:
@snicoll I'm fine with this solution. Thanks! If you need any feedback for #10684 before the next milestone release, feel free to ask.

PR #10672, opened 2017-10-17
UserNameAttribute for custom OAuth2 provider

thiagohirata:
The UserNameAttribute field for a custom OAuth2 provider was
ignored when building the ClientRegistration.


ericjturley:
But is there any workaround for this?
I kinda need it *now*. :)

mbhave:
@ericjturley This should be available in the latest 2.0.0 snapshots.

PR #10682, opened 2017-10-18
Remove @ConditionalOnEnabledEndpoint on EndpointExtensions

izeye:
EndpointExtension has `@ConditionalOnBean` on Endpoint which already has `@ConditionalOnEnabledEndpoint`, so they look redundant.

This PR removes them.


snicoll:
@izeye thanks for the PR but I don't think this is a good idea. You may have created the endpoint yourself and yet hope that the extension backs off if the endpoint is disabled via property. This is more about making sure the bean is not going to be created if the endpoint is disabled via property.

izeye:
@snicoll Thanks for the explanation!

PR #10714, opened 2017-10-20
cmdOptions configuration through application.properties for embedded mongodb autoconfigure

AdrienC:
Added feature to configure custom cmdOptions of embedded mongodb with spring boot.


wilkinsona:
Thanks for the PR, @AdrienC. Can you please describe your motivation for proposing this change? A number of these options feel like settings that may not be relevant for basic integration testing against an "embedded" instance of Mongo and are more applicable to a full-blown performance or acceptance test that would be performed with a multiple, separately installed instances of Mongo.

AdrienC:
It would be an easy way to enable oplog for testing oplog tailing with an embedded Mongo. (with spring.mongodb.embedded.cmdOptions.master=true)

philwebb:
Thanks for the PR but we're not keen to add so many additional properties for this. Perhaps a `Customizer` callback would work better, allowing programmatic configuration for these extra things that need customizing.

philwebb:
If you're interested in reworking the PR as a customizer callback, we'll certainly reconsider. Thanks!

Issue #10719, opened 2017-10-20
Upgrade Hibernate to 5.2.12


candrews:
http://in.relation.to/2017/10/19/hibernate-orm-5212-final-release/

snicoll:
Duplicate of #10720

PR #10720, opened 2017-10-20
Upgrade to Hibernate 5.2.12

candrews:
http://in.relation.to/2017/10/19/hibernate-orm-5212-final-release/

Closes gh-10719


snicoll:
Thanks for the PR but we have a semi-automated process that takes care of upgrading dependencies.

snicoll:
Also, no need to create an issue and a PR for the same thing.

candrews:
Thanks for the info, I didn't know :)

PR #10721, opened 2017-10-20
Upgrade hibernate-validator to 6.0.3.FINAL

candrews:
http://in.relation.to/2017/10/19/hibernate-validator-603-final-out/


snicoll:
Thanks for the PR but we have a semi-automated process that takes care of upgrading dependencies.

Issue #5361, opened 2016-03-08
.properties should be encoded in ISO-8859-1


jsiebahn:
By default Java .properties files are encoded in ISO-8859-1 (see https://en.wikipedia.org/wiki/.properties ) but since #4622 Spring-Boot reads them as UTF-8.

This causes incompatibility with Java and Spring itself.

The differences to Spring are very uncomfortable when it comes to unit tests, e.g. when the test case has its own Spring Context, .properties files are loaded by Spring and not by Spring Boot. They are parsed correctly as ISO-8859-1. But when the application is started, the same .properties are loaded by Spring-Boot and valid ISO-8859-1 characters are broken.

A JUnit test configured like this:

```
@RunWith(SpringJUnit4ClassRunner.class)
@ContextConfiguration(classes = MyTestConfig.class)
public class MyTest {
    // ...
}
```

with a test config importing a PropertySource:
```
@Configuration
@PropertySource(value = "/application.properties")
public class MyTestConfig {
}
```
will pass with ISO-8859-1 characters like "äöü" in the application.properties but the Spring-Boot application itself will show corrupted UTF-8 characters.

Furthermore the typical Java IDEs treat .properties files by default as ISO-8859-1 (or similar CP1252).

philwebb:
@dsyer can you provide any background on why #4622 changed the default?

dsyer:
We changed it because someone asked us to support extended character sets (8859 is pretty narrow). I don't think we should change it back. Surely `@PropertySource` is handled by Spring (not Boot) so it would always be the same.

philwebb:
It looks to me like `@PropertySource` uses `ResourcePropertySource` which uses `EncodedResource` but doesn't set `UTF-8` encoding. I think we might be unique with our, handling so I'd vote for reverting back to `ISO-8859-1`

snicoll:
Spring Framework recently supported an `encoding` attribute on `@PropertySource`. The default hasn't changed so I also think we should revert it. Having said that, maybe we could offer a configuration option (that will be hard to set) to customize this?

PR #10748, opened 2017-10-23
GitProperties, BuildProperties support unicode string.

hengyunabc:


wilkinsona:
I'm not sure about this one as the default for a `.properties` file is ISO-8859-1 and there have been problems in the past when we've deviated from that default (see https://github.com/spring-projects/spring-boot/issues/5361, for example). There's also the fact that Java 9 will move for UTF-8 by default to consider. This will make this change redundant while also providing consistent use of UTF, avoiding the problem reported in #5361.

Considering the above, I think we should leave this as it is. Any characters that don't fit in ISO-8859-1 should be escaped.

hengyunabc:
@wilkinsona 
Add a key like `spring.info.git.encoding=utf-8` to support unicode string?

A lots of applications will run under java8.

wilkinsona:
That's an interesting suggestion, thanks. I'm quite tempted to support this as you've proposed. I'll re-open this and let's see what the rest of the team thinks.

wilkinsona:
Oh, GitHub won't let me reopen it as "the master branch was force-pushed or recreated". That's a shame. Sorry for the hassle, but can you please open a new PR with your updated proposal?

PR #10765, opened 2017-10-24
Editing pass for using-spring-boot.adoc

Buzzardo:
I made the document more formal and corrected sentence errors.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
@@ -56,8 +56,8 @@ defaults. The parent project provides the following features:
 
 * Java 1.8 as the default compiler level.
 * UTF-8 source encoding.
-* A <<using-boot-dependency-management,Dependency Management section>> that lets you
-  omit `<version>` tags for common dependencies that are inherited from the
+* A <<using-boot-dependency-management,Dependency Management section>>, letting you
+  omit `<version>` tags for common dependencies, inherited from the
   `spring-boot-dependencies` POM.
 * Sensible https://maven.apache.org/plugins/maven-resources-plugin/examples/filter.html[resource filtering].
 * Sensible plugin configuration (http://www.mojohaus.org/exec-maven-plugin/[exec plugin],
@@ -66,9 +66,9 @@ defaults. The parent project provides the following features:
 * Sensible resource filtering for `application.properties` and `application.yml` including
   profile-specific files (for example, `application-foo.properties` and `application-foo.yml`)
 
-On the last point: Note that, since the default config files accept
+Note that, since the `application.properties` and `application.yml` files accept
 Spring style placeholders (`${...}`), the Maven filtering is changed to
-use `@..@` placeholders. (You can override that with a Maven property
+use `@..@` placeholders. (You can override that by setting a Maven property called
 `resource.delimiter`.)
 
 
@@ -137,7 +137,7 @@ The preceding sample setup does not let you override individual dependencies by 
 explained above. To achieve the same result, you need to add an entry in the
 `dependencyManagement` of your project **before** the `spring-boot-dependencies`
 entry. For instance, to upgrade to another Spring Data release train, you could add the
-following to your `pom.xml`, as follows:
+following element to your `pom.xml`:
 
 [source,xml,indent=0,subs="verbatim,quotes,attributes"]
 ----
@@ -186,7 +186,7 @@ section if you want to use it, as shown in the following example:
 ----
 
 NOTE: If you use the Spring Boot starter parent pom, you need to add only the plugin.
-There is no need for to configure it unless you want to change the settings defined in
+There is no need to configure it unless you want to change the settings defined in
 the parent.
 
 
@@ -290,8 +290,8 @@ installed, you can simply hit `ctrl-space` in the POM editor and type
 As explained in the <<spring-boot-features#boot-features-custom-starter,Creating Your Own Starter>>
 section, third party starters should not start with `spring-boot`, as it is reserved for
 official Spring Boot artifacts. Rather, a third-party starter typically starts with the
-name of the organization that made it. For example, a third-party starter from `acme`
-would typically be named `acme-spring-boot-starter`.
+name of the project. For example, a third-party starter project called `thirdpartyproject`
+would typically be named `thirdpartyproject-spring-boot-starter`.
 ****
 
 The following application starters are provided by Spring Boot under the
@@ -457,7 +457,7 @@ selection of core loggers and logs an auto-configuration report to the console.
 
 [[using-boot-disabling-specific-auto-configuration]]
 === Disabling Specific Auto-configuration Classes
-If you find that specific auto-configure classes that you do not want are being applied,
+If you find that specific auto-configuration classes that you do not want are being applied,
 you can use the exclude attribute of `@EnableAutoConfiguration` to disable them, as shown
 in the following example:
 
@@ -981,7 +981,7 @@ Then you need to set a `spring.devtools.remote.secret` property, as shown in the
 WARNING: Enabling `spring-boot-devtools` on a remote application is a security risk. You
 should never enable support on a production deployment.
 
-Remote devtools support is provided in two parts: a server_side endpoint that
+Remote devtools support is provided in two parts: a server-side endpoint that
 accepts connections and a client application that you run in your IDE. The server
 component is automatically enabled when the `spring.devtools.remote.secret` property
 is set. The client component must be launched manually.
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -56,27 +56,27 @@ defaults. The parent project provides the following features:
 
 * Java 1.8 as the default compiler level.
 * UTF-8 source encoding.
-* A <<using-boot-dependency-management,Dependency Management section>>, allowing you to
-  omit `<version>` tags for common dependencies, inherited from the
+* A <<using-boot-dependency-management,Dependency Management section>> that lets you
+  omit `<version>` tags for common dependencies that are inherited from the
```
wilkinsona: The meaning's wrong now. It's the dependency management that's inherited rather than the common dependencies

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -56,27 +56,27 @@ defaults. The parent project provides the following features:
...
   http://maven.apache.org/plugins/maven-shade-plugin/[shade]).
 * Sensible resource filtering for `application.properties` and `application.yml` including
-  profile-specific files (e.g. `application-foo.properties` and `application-foo.yml`)
+  profile-specific files (for example, `application-foo.properties` and `application-foo.yml`)
 
-On the last point: since the default config files accept
-Spring style placeholders (`${...}`) the Maven filtering is changed to
-use `@..@` placeholders (you can override that with a Maven property
-`resource.delimiter`).
+On the last point: Note that, since the default config files accept
```
wilkinsona: Should this be note rather than Note?

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -133,11 +133,11 @@ dependency:
...
 
-That setup does not allow you to override individual dependencies using a property as
-explained above. To achieve the same result, you'd need to add an entry in the
+The preceding sample setup does not let you override individual dependencies by using a property, as
+explained above. To achieve the same result, you need to add an entry in the
 `dependencyManagement` of your project **before** the `spring-boot-dependencies`
-entry. For instance, to upgrade to another Spring Data release train you'd add the
-following to your `pom.xml`.
+entry. For instance, to upgrade to another Spring Data release train, you could add the
+following to your `pom.xml`, as follows:
```
wilkinsona: IMO, following … as follows doesn't read very nicely

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -185,8 +185,8 @@ section if you want to use it:
 	</build>
 ----
 
-NOTE: If you use the Spring Boot starter parent pom, you only need to add the plugin,
-there is no need for to configure it unless you want to change the settings defined in
+NOTE: If you use the Spring Boot starter parent pom, you need to add only the plugin.
+There is no need for to configure it unless you want to change the settings defined in
```
wilkinsona: There's an extra "for" here. Not your doing, but thought I'd mention it as I noticed it during this review.

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -288,10 +287,11 @@ search dependencies by name. For example, with the appropriate Eclipse or STS pl
...
 "`spring-boot-starter`" for a complete list.
 
-As explained in the <<spring-boot-features#boot-features-custom-starter,Creating your own starter>>
-section, third party starters should not start with `spring-boot` as it is reserved for
-official Spring Boot artifacts. A third-party starter for `acme` will be typically named
-`acme-spring-boot-starter`.
+As explained in the <<spring-boot-features#boot-features-custom-starter,Creating Your Own Starter>>
+section, third party starters should not start with `spring-boot`, as it is reserved for
+official Spring Boot artifacts. Rather, a third-party starter typically starts with the
+name of the organization that made it. For example, a third-party starter from `acme`
```
wilkinsona: It's not the name of the organisation but the name of the project. Perhaps "acme" isn't an ideal example name as it does suggest that it's an organisation rather than a project.

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -444,21 +444,22 @@ recommend that you add it to your primary `@Configuration` class.
...
+selection of core loggers and logs an auto-configuration report to the console.
 
 
 
 [[using-boot-disabling-specific-auto-configuration]]
-=== Disabling specific auto-configuration
-If you find that specific auto-configure classes are being applied that you don't want,
-you can use the exclude attribute of `@EnableAutoConfiguration` to disable them.
+=== Disabling Specific Auto-configuration Classes
+If you find that specific auto-configure classes that you do not want are being applied,
```
wilkinsona: Not your doing but I think this should be "auto-configuration" rather than "auto-configure"

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -979,21 +981,21 @@ Then you need to set a `spring.devtools.remote.secret` property, for example:
 WARNING: Enabling `spring-boot-devtools` on a remote application is a security risk. You
 should never enable support on a production deployment.
 
-Remote devtools support is provided in two parts; there is a server side endpoint that
-accepts connections, and a client application that you run in your IDE. The server
+Remote devtools support is provided in two parts: a server_side endpoint that
```
wilkinsona: server-side?

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -979,21 +981,21 @@ Then you need to set a `spring.devtools.remote.secret` property, for example:
...
 component is automatically enabled when the `spring.devtools.remote.secret` property
 is set. The client component must be launched manually.
 
 
 
-==== Running the remote client application
+==== Running the Remote Client Application
 The remote client application is designed to be run from within your IDE. You need to run
-`org.springframework.boot.devtools.RemoteSpringApplication` using the same classpath as
-the remote project that you're connecting to. The _non-option_ argument passed to the
```
wilkinsona: I think non-option was right here. It's the difference between `whatever` and `--whatever`. Perhaps there's a better way to describe that?

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
Buzzardo: I'm not following this sentence. I suspect it needs to be two or perhaps three sentences, but I don't know what they should be. Can you rephrase them?

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
Buzzardo: Yup. Thanks.

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
Buzzardo: OK. I changed "acme" to "thirdpartyproject". Wile E. Coyote will be disappointed. ;)

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
Buzzardo: Good catch, but I think the better fix is to change "auto-configure" to "auto-configuration" in the first sentence. "auto-configure" should be a verb (that gets fuzzy in a hurry, though), while "auto-configuration should be the noun and adjective. So it's a smidgeon more correct to write, "auto-configuration."

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
Buzzardo: Yup. Thanks.

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
Buzzardo: I changed the sentence to describe the single required argument. It now reads, "The application's single required argument is the remote URL to which it connects."

wilkinsona: 👍

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
wilkinsona: I can try. How about something like this:

A dependency management section, inherited from the `spring-boot-dependencies` pom, that manages the versions of common dependencies. This dependency management allows you to omit `<version>` tags for those dependencies when used in your own pom.

Buzzardo: That works. Thanks, Andy.

PR #10771, opened 2017-10-25
GitProperties, BuildProperties support encoding.

hengyunabc:

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/info/ProjectInfoAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/info/ProjectInfoAutoConfiguration.java
@@ -36,8 +36,10 @@ import org.springframework.core.env.Environment;
 import org.springframework.core.io.DefaultResourceLoader;
 import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
+import org.springframework.core.io.support.EncodedResource;
 import org.springframework.core.io.support.PropertiesLoaderUtils;
 import org.springframework.core.type.AnnotatedTypeMetadata;
+import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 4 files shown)


hengyunabc:
@wilkinsona 

https://github.com/spring-projects/spring-boot/pull/10748

wilkinsona:
Please see #10748 for background. This revised proposal seems like a good compromise to me. Unlike the encoding for `application.properties`, we can be sure that any configuration setting for the encoding of Git properties or build properties has been read before trying to read the properties file.

snicoll:
Do the plugins control how the `properties` file is encoded? For sure `build.properties` is written with `ISO-9959-1` encoding. I haven't seen a property for the git plugin.

wilkinsona:
Good question, @snicoll. @hengyunabc How are you configuring your build to write UTF-8 characters to the properties files?

13567436138:
how to efficiently communicate in github? is there IM tools.

13567436138:
how guys communicate with each other?

hengyunabc:
@snicoll @wilkinsona 

https://github.com/ktoso/maven-git-commit-id-plugin/blob/v2.2.3/src/main/java/pl/project13/maven/git/GitCommitIdMojo.java#L319 

My project `project.build.sourceEncoding` is utf-8. 

The content in git.properties:

```
git.commit.user.name=中文
```

snicoll:
@hengyunabc thanks, that's quite interesting. I guess we should do the same thing. That doesn't help Gradle users though.

wilkinsona:
The equivalent Gradle plugin [always uses `UTF-8`](https://github.com/n0mer/gradle-git-properties/blob/master/src/main/groovy/com/gorylenko/GitPropertiesPlugin.groovy#L103).

snicoll:
@hengyunabc thanks for your contribution, I've merged it with a [polish commit](https://github.com/spring-projects/spring-boot/commit/c91d9bfda8f9c7203f6f45f08f55eebbf3e4b0cf)

PR #10806, opened 2017-10-29
Handle null early in Sanitizer.sanitize()

izeye:
This PR handles `null` value early in `Sanitizer.sanitize()` to avoid unnecessary `for` loop.


snicoll:
Good catch @izeye! Merged in `1.5.x` and `master`.

Issue #10464, opened 2017-09-30
Rationalize check location behaviour for Flyway and Liquibase


eddumelendez:
Currently, by default Flyway has `checkLocation` property equals to `false` and Liquibase has `checkChangeLogLocation` with `true`. Do you think they should share the same behaviour about checking locations?

The reason behind this, I have worked in a [feature](https://github.com/eddumelendez/spring-boot/tree/liquibase_vendor_location) to support resolution for `{vendor}` in `LiquibaseAutoConfiguration` and I realized that `checkChangeLogLocation` was `true` which makes me disable the check location in my test.

May related to #10387 about support checking vendor location.

wilkinsona:
Well spotted, @eddumelendez. I can see no reason for the default checking behaviour not to be consistent. It should probably be `true` in both cases. Flyway itself fails slightly later on if it doesn't find any migration scripts. @mbhave also suggested making the property names consistent (they could both be `checkLocation`).

izeye:
Duplicate of #10807 ?

eddumelendez:
Close because it was merged in https://github.com/spring-projects/spring-boot/commit/dcc6d8d6ffb257b454fc6b9986e09a8046808282

PR #10807, opened 2017-10-29
Make check-location property consistent in Flyway and Liquibase

eddumelendez:
See gh-10464


wilkinsona:
I'm thinking twice about renaming the properties. While things are now consistent between Flyway and Liquibase, the Liquibase properties for configuring the change log and checking it don't match up any more.

philwebb:
I'm tempted to keep the two properties distinct as well. I think the consistency within the class is more important that consistency between Flyway & Liquibase.

wilkinsona:
@eddumelendez Sorry for changing our mind on this one. Would you like to update your pull request so that it just switches the default value and leaves the property names unchanged?

wilkinsona:
Thank you once again, @eddumelendez

Issue #10817, opened 2017-10-30
Upgrade Spring Session to 2.0.0.RC1


vpavic:
It would be nice if we could get `master` to build against snapshots in the meanwhile.

wilkinsona:
I've just committed a change to build against snapshots in the run up to RC1.

PR #10818, opened 2017-10-30
Add support for configuring Spring Session cleanup cron

vpavic:
This PR adds support for configuring cron expression used for expired session cleanup job in Redis and JDBC session stores.

The two were recently introduced in Spring Session and will be available with the upcoming `2.0.0.RC1`.

~In order to be able to build this, I've also added a commit that updates Spring Session to current snapshot (see #10817).~


wilkinsona:
Thanks, @vpavic. This looks pretty complete. It is rather late in the day for M6, but let's see if we can squeeze this in.

snicoll:
Very complete indeed! thanks @vpavic

PR #10822, opened 2017-10-30
AbstractReactiveHealthIndicator does not handle doHealth exceptions

GreyTeardrop:
Exceptions inside `AbstractReactiveHealthIndicator.doHealthCheck()`
method, outside of `Mono` pipeline, could fail the whole endpoint
response instead of returning `DOWN` status from the indicator.

This issue could be reproduced with `RedisReactiveHealthIndicator`.
If Redis becomes unavailable,
```java
ReactiveRedisConnection connection = this.connectionFactory.getReactiveConnection();
```
throws `io.lettuce.core.RedisConnectionException` outside of reactive `Mono` pipeline
so `onErrorResume` does not handle it.


snicoll:
@GreyTeardrop thank you so much for your first contribution to Spring Boot. This is now merged in `master`.

PR #10831, opened 2017-10-30
Add RestTemplate customizer to log request/response bodies.

markhobson:
It's often useful to be able to log HTTP traffic when using `RestTemplate`. This `RestTemplateCustomizer` configures an interceptor that logs all request and response bodies, taking care of the associated buffering subtleties.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/LoggingCustomizer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/LoggingCustomizer.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks very much for the contribution. Although I can see this could be helpful, I'm worried that it's making some assumptions that might not hold true for all responses. Specifically, calling `StreamUtils.copyToString(response.getBody(), StandardCharsets.UTF_8)` means that any non-string responses would cause problems. There's also the potential that large responses would consume a lot of memory and overwhelm the logging system.

All things considered, I'm tempted to suggest that if users want this functionality they should probably add it directly themselves.

I've labeled this PR as `for: team-attention` so lets see what the rest of the team think.

markhobson:
Thanks for the feedback @philwebb. I agree that it should respect the content type encoding and not log non-string responses - I'm happy to update the PR for that.

I understand your concerns but it would be great to see this functionality in Spring Boot as it's a requirement that many of my colleagues have come up with time and time again. I see it as akin to [enabling wire logging in HttpClient](https://hc.apache.org/httpcomponents-client-ga/logging.html). I'm open to any alternative approaches that would satisfy the same wire debugging use cases.

philwebb:
@markhobson I think unfortunately that it's going to be impossible to work around the requirement that `BufferingClientHttpRequestFactory` is used, and we're very reluctant to promote that because of the memory that it can consume.

Whilst I think that generic logging using `RestTemplateCustomizer` can be useful, it very much depends on the application so I'm afraid it's not something that we want to add to Spring Boot itself because we're worried that people might not fully understand the trade-offs.

Thanks anyway for the PR!

markhobson:
@philwebb No worries, thanks for taking the time to review!

I've put the code here if anyone comes across this and wants to use it:
https://github.com/markhobson/rest-template-logger

PR #10834, opened 2017-10-30
Editing pass for spring-boot-cli.adoc

Buzzardo:
I made it more formal and corrected various errors. I also reduced the heading depth by making the two best practices into titled tips.


wilkinsona:
Thank you, @Buzzardo. These changes are now in master.

PR #10848, opened 2017-10-31
Remove System.out::println

izeye:
It looks accidentally added in ce77f48c3f9c8acdd12d3a629b9c8a393050c271.


philwebb:
Good spot. Thanks!

wilkinsona:
🤦‍♂️ Thank you, @izeye!

PR #10869, opened 2017-11-01
Add the enableSpringELCompiler on ThymeleafProperties

kazuki43zoo:
I suggest adding the `enableSpringElCompiler` property on the `ThymeleafProperties`.
By this change, we can be changed the `SpringTemplateEngine#enableSpringELCompiler` to `true` more easier.

WDYT?


philwebb:
LGTM but lets see what the rest of the team think.

kazuki43zoo:
I've updated.

* fie checkstyle error
* fix property name (enableSpringELCompiler -> enableSpringE**l**Compiler )
* update document

kazuki43zoo:
I've updated.

* fix setter argument variable name (enableSpringELCompiler -> enableSpringE**l**Compiler )

PR #10870, opened 2017-11-01
Editing pass for appendix-application-properties.adoc

Buzzardo:
I improved consistency and corrected sentence-level errors.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
@@ -2,16 +2,16 @@
 [appendix]
 [[common-application-properties]]
 == Common application properties
-Various properties can be specified inside your `application.properties`/`application.yml`
-file or as command line switches. This section provides a list of common Spring Boot
-properties and references to the underlying classes that consume them.
+Various properties can be specified inside your `application.properties` file, inside
+your  `application.yml` file, or as command line switches. This appendix provides a list
+of common Spring Boot properties and references to the underlying classes that consume
+them.
 
-NOTE: Property contributions can come from additional jar files on your classpath so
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
```diff
@@ -31,50 +31,50 @@ content into your application; rather pick only the properties that you need.
 	# BANNER
 	banner.charset=UTF-8 # Banner file encoding.
 	banner.location=classpath:banner.txt # Banner file location.
-	banner.image.location=classpath:banner.gif # Banner image file location (jpg/png can also be used).
+	banner.image.location=classpath:banner.gif # Banner image file location (jpg or png can also be used).
```
snicoll: @Buzzardo this description is actually used in the IDE (the metadata appendix you've also reviewed, thanks!). So if we change anything in here we have to change the source code. This content is actually a copy of the Javadoc of the related property. 

Given the number of changes here it will take some time to process.

wilkinsona: This is another unfortunate casualty of having the same information in two places. Can you bite the bullet on auto-generating the appendix?

philwebb:
We're going to need the code as the source of truth so we need to migrate these edits either way. I've made a start [here](https://github.com/philwebb/spring-boot/tree/gh-10870).

snicoll:
@Buzzardo @philwebb I am not a native English speaker so I certainly don't have an opinion about this change. However, I'd like to point out that the first sentence of each key is shown in an IDE where every character counts. Is there a way to take that into account?

wilkinsona:
FWIW, I think the updates are fine from a length perspective. Some of them are now shorter, and those that are longer are more accurate and consistent now. I can't see an easy way to shorten them without committing crimes against grammar.

PR #10887, opened 2017-11-02
Add auto-configuration for RabbitMQ metrics

acogoluegnes:
Automatically plug RabbitMQ Java client metrics collection to micrometer.


snicoll:
@jkschneider shouldn't the `RabbitMetrics` be part of Micrometer?

jkschneider:
@snicoll Micrometer support was added directly to the latest Rabbit java client. So we'll have to autoconfigure them in both Boot 2's metrics autoconfiguration and the 1.5 legacy module.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/amqp/RabbitMetrics.java
```diff
@@ -0,0 +1,211 @@
...
+
+	@Override
+	public void bindTo(MeterRegistry registry) {
+		this.connectionFactory.setMetricsCollector(new MicrometerMetricsCollector(registry, "rabbitmq", this.tags));
+	}
+
+	/**
+	 * Micrometer implementation of {@link com.rabbitmq.client.MetricsCollector}.
+	 */
+	public static class MicrometerMetricsCollector extends AbstractMetricsCollector {
```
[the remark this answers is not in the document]
acogoluegnes: Micrometer support isn't available in RabbitMQ Java Client 5.0.x (which is the version Spring Boot 2.0 depends on AFAIR). It was introduced in 4.3.0 and pushed upstream in 5.1.0. Spring Boot can choose to depend on 5.1.1 (latest) to re-use the existing `MicrometerMetricsCollector`, even though Spring AMQP depends on 5.0.0.
`MicrometerMetricsCollector` code is straightforward though, the parent class handle the heavy-lifting.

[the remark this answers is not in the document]
wilkinsona: @acogoluegnes Using 5.1.x seems like the best option purely from Boot's perspective. Is 5.1 backwards compatible with 5.0 or is it likely to cause problems for Spring AMQP do you know?

/cc @garyrussell @artembilan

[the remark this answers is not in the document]
garyrussell: I just tested master against 5.1.1 with no problems.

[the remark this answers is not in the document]
artembilan: I think we can't upgrade Spring AMQP 2.0.x to the AMQP Client 5.1.x, but Spring Boot 2.0 still can do that and therefore replace Spring AMQP's one.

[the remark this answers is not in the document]
acogoluegnes: @wilkinsona 5.1 is backwards compatible with 5.0, RabbitMQ Java Client uses semantic versioning as of 4.0.0. Thanks for testing out @garyrussell.

@wilkinsona As @artembilan and you suggested, you can use 5.1.1 in Spring Boot, it should cause any trouble to Spring AMQP.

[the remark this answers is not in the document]
jkschneider: @artembilan Could we push towards a Spring AMQP 2.1.x? Or is this too much of a support burden?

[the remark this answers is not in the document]
artembilan: There is nothing yet for Spring AMQP 2.1. 
More over we only talk about a compatible dependency. So, if you don't use Spring Boot but only Spring AMQP, you should override dependency manually. That's all.
Otherwise with the Spring Boot everything is transparent for you.
And yes: that is really too much to support...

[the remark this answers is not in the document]
garyrussell: If the version police will allow it, I wouldn't mind changing to 5.1.x in 2.0.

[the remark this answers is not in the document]
wilkinsona: I guess I'm (one of) the version police :). I have no objections to doing that.

snicoll:
@acogoluegnes can you rework your PR using the RabbitMQ Java Client 5.1?

acogoluegnes:
@snicoll Sure, I'll have a look this week hopefully.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfiguration.java
```diff
@@ -76,12 +77,13 @@
 @EnableConfigurationProperties(MetricsProperties.class)
 @Import({ MeterBindersConfiguration.class, WebMvcMetricsConfiguration.class,
 		WebFluxMetricsConfiguration.class, RestTemplateMetricsConfiguration.class,
-		DataSourcePoolMetricsConfiguration.class, AtlasExportConfiguration.class,
-		DatadogExportConfiguration.class, GangliaExportConfiguration.class,
-		GraphiteExportConfiguration.class, InfluxExportConfiguration.class,
-		JmxExportConfiguration.class, PrometheusExportConfiguration.class,
-		SimpleExportConfiguration.class, StatsdExportConfiguration.class })
-@AutoConfigureAfter({DataSourceAutoConfiguration.class, RabbitAutoConfiguration.class})
+		CacheMetricsConfiguration.class, DataSourcePoolMetricsConfiguration.class,
```
jkschneider: This adds `CacheMetricsConfiguration`, which probably should have been present already independently of RabbitMQ metrics?

acogoluegnes: I may have made something wrong when merging with master, but it seems `CacheMetricsConfiguration` was already there: https://github.com/spring-projects/spring-boot/blob/027e6baba53a83e385bc307925c654c55da346a8/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfiguration.java#L73. Am I missing something?

jkschneider: Ah, ok. Great.

acogoluegnes:
I can't re-use RabbitMQ Java Client's `MicrometerMetricsCollector` as it's based on Micrometer 1.0.0.rc2, and some APIs have changed since then. I can fix this and release 5.1.2 next week. Would you prefer to merge this PR as-is and have another PR to use the new `MicrometerMetricsCollector` or to wait for this PR until it uses 5.1.2?

snicoll:
@acogoluegnes I'd rather something that works with the 1.0.0 target. Do I understand that `RabbitMetrics` would go away and be replaced by the native feature in the driver?

acogoluegnes:
@snicoll Yes, `RabbitMetrics` would go away and be replaced by the native feature in the driver. Note what I've pushed yesterday works with the Micrometer version Spring Boot depends on.

acogoluegnes:
@snicoll @wilkinsona I updated the dependency to RabbitMQ Client 5.1.2 and `RabbitMetrics` now uses the native Micrometer support. This PR can now be merged hopefully.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfiguration.java
```diff
@@ -132,6 +139,19 @@ public SpringIntegrationMetrics springIntegrationMetrics(
...
 	}
 
+	@Configuration
+	@ConditionalOnClass({ RabbitTemplate.class, Channel.class })
+	@ConditionalOnBean(AbstractConnectionFactory.class)
+	@ConditionalOnProperty(prefix = "spring.rabbitmq", name = "metrics", matchIfMissing = true)
+	static class MetricsRabbitConfiguration {
+
+		@Bean
+		public RabbitMetrics rabbitMetrics(AbstractConnectionFactory connectionFactory) {
```
snicoll: that only works if there is one `AbstractConnectionFactory` in the context. What if there is more than one? Wouldn't  we want to monitor them all rather than failing?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/amqp/RabbitMetrics.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	public RabbitMetrics(ConnectionFactory connectionFactory, Iterable<Tag> tags) {
+		this.connectionFactory = connectionFactory;
+		this.tags = tags;
+	}
+
+	@Override
+	public void bindTo(MeterRegistry registry) {
+		this.connectionFactory.setMetricsCollector(new MicrometerMetricsCollector(registry, "rabbitmq", this.tags));
```
snicoll: Same idea here, if  we support multiple connection factories, there should be a way to identify  them.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -112,6 +113,11 @@
 
 	private List<Address> parsedAddresses;
 
+	/**
+	 * Enable metrics.
+	 */
+	private boolean metrics = true;
```
snicoll: That looks the wrong  place to  do this. Metrics related properties are in `management.metrics`. Check other implementations for inspiration (the JDBC one for  instance).

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfiguration.java
acogoluegnes: Makes sense. We could get all the `AbstractConnectionFactory` with their bean name as well (see below).

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/amqp/RabbitMetrics.java
acogoluegnes: Yes. For one connection factory, we could stick to the `rabbitmq` prefix and use `rabbitmq.<bean-name>` for multiple connection factories.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
acogoluegnes: Would `management.metrics.rabbitmq.instrument-connection-factory` work?

acogoluegnes:
@snicoll How do you want to proceed? Should I update this PR for the new changes?

snicoll:
@acogoluegnes nope we're good, I can take that  as part of the polish commit. Thank you for confirming this makes sense!

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
jkschneider: We don't need a specific enablement property for Rabbit, as we have property-bound `MeterFilter`s in Boot 2. For example, if all of the Rabbit metrics are prefixed with "rabbit", then `management.metrics.filter.rabbit.enabled=false` would disable them.

snicoll:
@acogoluegnes thank you so much for your first contribution to Spring Boot. This is now merged in `master` (see the [polish commit](beaa49a) if you're interested by what I've changed).

PR #10889, opened 2017-11-02
90-character formatting pass for documentation-overview.adoc

Buzzardo:
I removed extraneous white space and ensured that all the lines that can break at 90 characters do break as close to 90 characters as they can.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/documentation-overview.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/documentation-overview.adoc
@@ -10,7 +10,7 @@ as a map for the rest of the document.
 
 
 [[boot-documentation-about]]
-== About the documentation
+== About the Documentation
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/main/asciidoc/documentation-overview.adoc
```diff
@@ -7,8 +7,6 @@ This section provides a brief overview of Spring Boot reference documentation. I
 as a map for the rest of the document.
 --
 
-
-
```
snicoll: The three spaces between sections is intentional. Can you please restore them?

Buzzardo:
OK. I restored those. I also put the headings into title case (I missed that on the first pass).

I squashed my two commits, so that you only have to wade through one commit.

wilkinsona:
Thanks again, @Buzzardo. Merged into master.

PR #10901, opened 2017-11-03
Add support for vendor in Liquibase

eddumelendez:
This commit introduces resolution for `{vendor}` in
`spring.liquibase.change-log` configuration property. Also, it disables
check location which its true by default because `{vendor}` is resolve
afterwards.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
@@ -39,11 +39,14 @@ import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
 import org.springframework.boot.autoconfigure.orm.jpa.HibernateJpaAutoConfiguration;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.boot.jdbc.DataSourceBuilder;
+import org.springframework.boot.jdbc.DatabaseDriver;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thank you, @eddumelendez. I'm not sure that this makes sense for Liquibase in the same way that it did for Flyway. The key difference is that Liquibase requires a single location for the change log, whereas Flyway can look in multiple locations. With Flyway that means that you can combine general migrations and vendor-specific migrations spread across multiple files. There's still some value to the feature with Liquibase, but it becomes less compelling when only a single file can be targeted.

We're also introduced a new inconsistency here. Unless I've missed something, the Flyway support doesn't disable location checking when `{vendor}` is used. That may be an oversight and the Flyway support should be updated, or we might be happy to assume that there will always be one non-vendor-specific migration. Whatever we decide, I think we should consider if we want to be consistent.

If we want to keep the check even when all locations (or the only location when using Liquibase) use `{vendor}`, we could perhaps consider replacing `{vendor}` with a wildcard. It wouldn't be 100% accurate but may be better than disabling the check entirely.

vpavic:
With Liquibase being a higher level database migration tool I'm not sure adding this makes sense. You're using Liquibase's own DSL to define migrations and that abstracts the underlying RDBMS, so the support for multiple vendors is actually provided by Liquibase itself. Especially since their DSL offers the capabilities for making changesets conditional on RDBMS type and makes it possible to include SQL formatted changesets.

wilkinsona:
Thank you, @vpavic. For anyone who's curious, one way in which Liquibase supports this is via the [`dbms` attribute on a `<changeSet>`](http://www.liquibase.org/documentation/changeset.html). 

@eddumelendez In light of what @vpavic has said above, I think we should defer to Liquibase's built-in support for applying database-specific changes rather than partially reinventing the wheel.

eddumelendez:
forgot the `dbms` attribute. I totally understand the reason why is not merged thanks @wilkinsona

PR #10978, opened 2017-11-10
Make RabbitTemplate exchange and routingKey configurable

ArloL:
This adds three properties to `RabbitProperties.Template` and sets them on the auto-configured `RabbitTemplate` bean. These are the three properties our internal code sets the most and would love to be able to control them via external configuration.

Please let me know if there are any issues.

Thanks for the great work on Spring Boot!

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
@@ -188,6 +188,9 @@ public class RabbitAutoConfiguration {
 			if (templateProperties.getReplyTimeout() != null) {
 				rabbitTemplate.setReplyTimeout(templateProperties.getReplyTimeout());
 			}
+			rabbitTemplate.setExchange(templateProperties.getExchange());
+			rabbitTemplate.setRoutingKey(templateProperties.getRoutingKey());
+			rabbitTemplate.setChannelTransacted(templateProperties.isChannelTransacted());
 			return rabbitTemplate;
 		}
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
@@ -703,6 +703,21 @@ public class RabbitProperties {
 		 */
 		private Long replyTimeout;
 
+		/**
+		 * Name of the default exchange to use for send operations.
+		 */
+		private String exchange = "";
+
+		/**
+		 * Value of a default routing key to use for send operations.
+		 */
+		private String routingKey = "";
+
+		/**
+		 * Enable transactional channels.
+		 */
+		private boolean channelTransacted;
+
 		public Retry getRetry() {
 			return this.retry;
 		}
@@ -729,6 +744,30 @@ public class RabbitProperties {
 
 		public void setReplyTimeout(Long replyTimeout) {
 			this.replyTimeout = replyTimeout;
+		}
+
+		public String getExchange() {
+			return this.exchange;
+		}
+
+		public void setExchange(String exchange) {
+			this.exchange = exchange;
+		}
+
+		public String getRoutingKey() {
+			return this.routingKey;
+		}
+
+		public void setRoutingKey(String routingKey) {
+			this.routingKey = routingKey;
+		}
+
+		public boolean isChannelTransacted() {
+			return this.channelTransacted;
+		}
+
+		public void setChannelTransacted(boolean channelTransacted) {
+			this.channelTransacted = channelTransacted;
 		}
 
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfigurationTests.java
@@ -104,6 +104,21 @@ public class RabbitAutoConfigurationTests {
 	}
 
 	@Test
+	public void testDefaultRabbitTemplateConfiguration() {
+		this.contextRunner.withUserConfiguration(TestConfiguration.class)
+				.run((context) -> {
+					RabbitTemplate rabbitTemplate = context.getBean(RabbitTemplate.class);
+					RabbitTemplate defaultRabbitTemplate = new RabbitTemplate();
+					assertThat(rabbitTemplate.getRoutingKey())
+							.isEqualTo(defaultRabbitTemplate.getRoutingKey());
+					assertThat(rabbitTemplate.getExchange())
+							.isEqualTo(defaultRabbitTemplate.getExchange());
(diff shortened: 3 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -699,6 +699,21 @@ public void setConsumersPerQueue(Integer consumersPerQueue) {
 		 */
 		private Long replyTimeout;
 
+		/**
+		 * The name of the default exchange to use for send operations.
```
snicoll: Configuration properties description do not start with"The", "A", etc. Please look at the appendix to harmonize them.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -699,6 +699,21 @@ public void setConsumersPerQueue(Integer consumersPerQueue) {
...
+		 */
+		private String exchange;
```
snicoll: The default exchange is `""` so it would be nice if the property would say that rather than `null`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -699,6 +699,21 @@ public void setConsumersPerQueue(Integer consumersPerQueue) {
...
+
+		/**
+		 * The value of a default routing key to use for send operations.
+		 */
+		private String routingKey;
+
+		/**
+		 * Enable transactional channels.
+		 */
+		private boolean channelTransacted;
```
snicoll: This one looks totally unrelated to the the first two and makes me a bit neverous. 

@garyrussell, any advice here? Does that sound safe to change the default via properties?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
```diff
@@ -186,6 +186,13 @@ public RabbitTemplate rabbitTemplate(ConnectionFactory connectionFactory) {
 			if (templateProperties.getReplyTimeout() != null) {
 				rabbitTemplate.setReplyTimeout(templateProperties.getReplyTimeout());
 			}
+			if (templateProperties.getExchange() != null) {
+				rabbitTemplate.setExchange(templateProperties.getExchange());
+			}
+			if (templateProperties.getRoutingKey() != null) {
+				rabbitTemplate.setRoutingKey(templateProperties.getRoutingKey());
+			}
+			rabbitTemplate.setChannelTransacted(templateProperties.isChannelTransacted());
```
snicoll: A test should cover this. If we go with the route of flagging `""` a the default value. We should have a test that this value is consistent with what happens  by default.

ArloL:
Thanks for the feedback. I have reviewed your suggestions.

The default values now match those of RabbitTemplate. A test ensures this. I am not sure whether the test is the way you would want it though. Let me know if you want me to refactor it.

I have removed "The" from the descriptions to match the existing style.

If you have any further suggestions let me know and I'll be happy to review them.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
```diff
@@ -188,6 +188,9 @@ public RabbitTemplate rabbitTemplate(ConnectionFactory connectionFactory) {
 			if (templateProperties.getReplyTimeout() != null) {
 				rabbitTemplate.setReplyTimeout(templateProperties.getReplyTimeout());
 			}
+			rabbitTemplate.setExchange(templateProperties.getExchange());
+			rabbitTemplate.setRoutingKey(templateProperties.getRoutingKey());
+			rabbitTemplate.setChannelTransacted(templateProperties.isChannelTransacted());
```
garyrussell: I would prefer the use of null checks rather than "mirroring" the current defaults in the template - if we ever change the defaults in the template they will be undone here. See the properties above (reply timeout etc) for examples; so `channelTransacted` should be `Boolean`  so we can check whether the user has explicitly set the property.

While it is unlikely those defaults will change, this is the general pattern used by Boot's auto configuration.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
garyrussell: Routing key and exchange should default to `null` and make this a `Boolean` - see my other comment.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
snicoll: > this is the general pattern used by Boot's auto configuration.

Actually, that's not 100% accurate. When we can add a test  that validates that our default are the same as the underlying tech's default, we try as much as we can to put it back in the config. That way, users can know what the defaults are in the IDE.

garyrussell: OK; that makes sense.

As discussed offline; if we expose the `channelTransacted` flag on the template, we should do the same for the listener container factories so that any rabbit template operations performed on the container thread will participate in the same transaction.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
ArloL: I will leave this as `boolean` and `false` then since the value is tested.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
ArloL: I will leave these then since they are tested.

ArloL:
I have rebased the PR to trigger another CI build since the last one failed due to an unrelated module.

snicoll:
@ArloL thanks for the PR. I've polished your commit in 6a4a0e3 (in particular adding a test for the feature that was added). As discussed in the review, I have not added the transacted flag as it makes me nervous to add it without some sort of coordination with the message listener.

I think we need to discuss the various cases before adding this feature. Feel free to open a separate issue once you've a more precise idea.

habuma:
@snicoll : Just noticed this PR (and like it a lot). Was wondering if there's any reason why there's not a similar means of setting the default queue for the purposes of `RabbitTemplate.receive()` methods that rely on a default queue? Or am I overlooking something?

I'll be happy to submit a PR when I get a chance (I'm on the road ATM or else I'd just do it now) if this is something that's both missing and makes sense to you.

Issue #10470, opened 2017-10-02
More details logged in case of ApplicationFailedEvent


luchob:
Hi, guys!

I’ve made a trivial mistake in my `application.yml` file (a duplicate entry) and the app silently hanged in the air with the following message:

_“Application failed to start with classpath: …“_

That was everything. I traced the message to [ClasspathLoggingApplicationListener](https://github.com/spring-projects/spring-boot/blob/master/spring-boot/src/main/java/org/springframework/boot/context/logging/ClasspathLoggingApplicationListener.java) and down to this code: 

```
@Override
public void onApplicationEvent(ApplicationEvent event) {
	if (logger.isDebugEnabled()) {
		if (event instanceof ApplicationEnvironmentPreparedEvent) {
			logger.debug("Application started with classpath: " + getClasspath());
		}
		else if (event instanceof ApplicationFailedEvent) {
			logger.debug(
					"Application failed to start with classpath: " + getClasspath());
		}
	}
}
```

I debugged  and took the stack trace from the event - the reason became immediately apparent, there was an appropriate message within.

Question: Shouldn’t some more details be provided in case of ApplicationFailedEvent?  Is the log level really debug?

Shall I try to provide some more details?

Thanks,
L.

snicoll:
> Shall I try to provide some more details?

Yes please. A sample that reproduces the mistake would be ideal.

luchob:
Thanks you Stéphane! The sample is [here](https://github.com/luchob/boot-issue-10470). It is almost empty and contains an `application.yml` with duplicated keys:

```
a:
  b:
    c: xyz

a:
  b:
    c: xyz
```

The only dependency is:

```
dependencies {
  compile('org.springframework.boot:spring-boot-starter-web')
}
```
Spring Boot 1.5.7.RELEASE.

Running the app hangs with this DEBUG message and no more info is provided:

```
14:38:52.823 [main] DEBUG org.springframework.boot.logging.ClasspathLoggingApplicationListener - Application failed to start with classpath: [some jars here...]
```

Removing the duplicate key solves the problem.

Hope this helps.

snicoll:
Thanks for the sample, that is very useful!

wilkinsona:
This is a duplicate of #10362 which has been fixed in 1.5.8 snapshots. Running the sample against those snapshots yields, the app no longer hangs and it outputs the following:

```
20:21:04.666 [main] ERROR org.springframework.boot.SpringApplication - Application startup failed
[...]

PR #11017, opened 2017-11-14
Log throwable on ApplicationFailedEvent

kbecic:
When `ApplicationFailedEvent` happens currently only Classpath is logged by `ClasspathLoggingApplicationListener`. 
Added logging of `Throwable` that caused `ApplicationFailedEvent`.

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/RemoteSpringApplication.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/RemoteSpringApplication.java
@@ -16,10 +16,6 @@
 
 package org.springframework.boot.devtools;
 
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.List;
-
 import org.springframework.boot.Banner;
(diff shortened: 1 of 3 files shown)


philwebb:
There's some quite complex logic around when failure exceptions should be logged (see `SpringBootExceptionHandler`) and It should already be the case that exceptions do get reported.

Can you provide some background on what prompted you to submit the pull request? Are finding that exceptions aren't reported? Can you provide a sample that shows this?

snicoll:
@kbecic are you on `1.5.7` by any chance? Can you upgrade then because this might not be necessary, see #10470

kbecic:
@snicoll Correct, fixed in 1.5.8 issue #10470 

@philwebb What is the reasoning behind not always logging `ApplicationFailedEvent` on it's own?

snicoll:
Thanks for the feedback.

philwebb:
@kbecic There are a few reasons. The main ones are that an earlier `ApplicationFailedEvent` listener may itself throw an exception (which would stop the later ones from getting it) and it gives as an easy way to hook in the `SpringBootExceptionReporter` support (which may also suppress the stacktrace).

Issue #8062, opened 2017-01-23
Provide TestRestTemplateExtensions for Kotlin projects


sdeleuze:
As discussed in #5537, it would be nice for Spring Boot 2.0 to provide `TestRestTemplateExtensions` in order to bring the same kind of `RestTemplate` improvements Spring Framework 5.0 provides with [RestOperationsExtensions](https://github.com/spring-projects/spring-framework/blob/master/spring-web/src/main/kotlin/org/springframework/web/client/RestOperationsExtensions.kt) to `TestRestTemplate`.

philwebb:
Closing in favor of PR #11039

PR #11039, opened 2017-11-15
Introduce TestRestTemplate Kotlin extensions

sdeleuze:
This commit introduces Kotlin extensions similar to the `RestOperations` ones in order to be able to take advantage of Kotlin reified type parameters for example.

Fixes gh-8062

### DIFF
--- a/spring-boot-project/spring-boot-parent/pom.xml
+++ b/spring-boot-project/spring-boot-parent/pom.xml
@@ -54,6 +54,25 @@
 				<version>1.2.17</version>
 			</dependency>
 			<dependency>
+				<groupId>com.nhaarman</groupId>
+				<artifactId>mockito-kotlin</artifactId>
+				<version>1.5.0</version>
+				<exclusions>
+					<exclusion>
+						<groupId>org.jetbrains.kotlin</groupId>
+						<artifactId>kotlin-stdlib</artifactId>
+					</exclusion>
+					<exclusion>
+						<groupId>org.jetbrains.kotlin</groupId>
+						<artifactId>kotlin-reflect</artifactId>
+					</exclusion>
+					<exclusion>
+						<groupId>org.mockito</groupId>
+						<artifactId>mockito-core</artifactId>
+					</exclusion>
+				</exclusions>
+			</dependency>
+			<dependency>
 				<groupId>com.squareup.okhttp</groupId>
 				<artifactId>okhttp</artifactId>
 				<version>2.7.5</version>
--- a/spring-boot-project/spring-boot-test/pom.xml
+++ b/spring-boot-project/spring-boot-test/pom.xml
@@ -115,6 +115,16 @@
 			<artifactId>htmlunit</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>org.jetbrains.kotlin</groupId>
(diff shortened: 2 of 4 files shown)


spring-boot-project/spring-boot-parent/pom.xml
```diff
@@ -229,6 +229,25 @@
 				<artifactId>zt-zip</artifactId>
 				<version>1.7</version>
 			</dependency>
+			<dependency>
```
snicoll: Dependencies are (meant to be) ordered alphabetically. Can you please rework this?

snicoll: Dependencies are (meant to be) ordered alphabetically. Can you please rework this?

spring-boot-project/spring-boot-test/pom.xml
```diff
@@ -196,6 +211,38 @@
 					</execution>
 				</executions>
 			</plugin>
+			<plugin>
```
snicoll: I am confused. In the `spring-boot` module you've set the maven compiler phases to `none`. Why isn't it necessary here?

sdeleuze: Good catch, it is, I will update the PR accordingly.

sdeleuze:
@snicoll I have pushed an updated commit with your feedback taken in account.

snicoll:
Thanks @sdeleuze. One more thing: we have [a test that validates that our `TestRestTemplate` provides the same features as the original `RestTemplate`](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/web/client/TestRestTemplateTests.java#L126). Would it be possible to add a similar test to assert that this Kotlin extension is equivalent to the one for `RestTemplate`?

sdeleuze:
@snicoll In Java, the test uses reflection to list all `RestOperations`  methods and invoke their `TestRestTemplate` counterparts, but extensions are top level functions. Unless I miss something, I can see a reasonably complex way of doing the same kind of check with Kotlin extensions.

wilkinsona:
@sdeleuze Perhaps we could do a similar but less extensive check? It would at least be nice to have a test that verifies that we have a top-level function for each method, even if the test doesn't invoke the function. I am working on the perhaps naive assumption that Kotlin is compiled to byte code so we must be able to introspect that byte code…

sdeleuze:
@wilkinsona But we don't have 1 extension for each method, we only provide the relevant ones.

wilkinsona:
Sorry, "each method" was a little imprecise. I should have written "each method that's relevant". Assuming that we can identify a relevant method programatically, then my hope is that it would be possible to test that there's a corresponding top-level function for that method.

sdeleuze:
@snicoll @wilkinsona I have been able to create such test using a mix of Java and Kotlin reflection. It allowed to find 5 missing extensions, thanks for pushing for this ;-) I have added the missing extensions to this PR and to Spring Framework 5.0.2 as well via [SPR-16229](https://jira.spring.io/browse/SPR-16229). Can you please merge this commit in order to be included as part of `2.0.0.M7`?

snicoll:
Sébastien shared he'd like to use that feature during an upcoming workshop.

PR #11054, opened 2017-11-16
Fix `SessionAutoConfiguration` ordering

vpavic:
This PR updates `SessionAutoConfiguration` to ensure it is processed after `MongoDataAutoConfiguration` and `MongoReactiveDataAutoConfiguration`, instead of `MongoAutoConfiguration` and `MongoReactiveAutoConfiguration`. This is required in order for MongoDB Session auto-configuration to work, since `MongoSessionConfiguration` and `MongoReactiveSessionConfiguration` are conditional on `MongoOperations` and `ReactiveMongoOperations` beans, respectively.


snicoll:
Good catch @vpavic!

PR #11055, opened 2017-11-16
Add Spring Session WebFlux sample

vpavic:
This PR adds Spring Session WebFlux sample app to demonstrate the auto-configuration capabilities for reactive Spring Session.

The sample uses MongoDB profile by default and uses an embedded MongoDB to avoid dependencies on external services, similarly like `spring-boot-sample-session` uses JDBC profile by default.

I've also included an integration test however this has uncovered issues with Spring Session MongoDB auto-configuration that require fixes on multiple levels, see #11054 and spring-projects/spring-session-data-mongodb#19 - both of these are required to make this integration test to work.

### DIFF
--- a/spring-boot-samples/pom.xml
+++ b/spring-boot-samples/pom.xml
@@ -67,6 +67,7 @@
 		<module>spring-boot-sample-secure-webflux</module>
 		<module>spring-boot-sample-servlet</module>
 		<module>spring-boot-sample-session</module>
+		<module>spring-boot-sample-session-webflux</module>
 		<module>spring-boot-sample-simple</module>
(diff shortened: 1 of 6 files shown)


vpavic:
In order for this to make `2.0.0.M7`, Spring Session Data MongoDB would need to release `2.0.0.RC2` with  spring-projects/spring-session-data-mongodb#19 - ping @gregturn.

vpavic:
With `spring-session-data-mongodb` bumped to `2.0.0.BUILD-SNAPSHOT` in e4e635f this PR now builds successfully.

I also had to provide some security config due to recent change in Spring Security (spring-projects/spring-security#4825)

PR #11068, opened 2017-11-17
Remove explicit super() calls

izeye:
This PR removes explicit `super()` calls.


philwebb:
Thanks again for all your excellent contributions!

Issue #10176, opened 2017-09-06
Try to drop EndpointExposure


philwebb:
It might be possible to drop the `EndpointExposure` enum and replace it with a class. Currently it's somewhat limiting in that additional items can't be added easily.

We might be able to use `Class<? extends AnnotationEndpointDiscoverer>`.

philwebb:
The enum really reflects reality. For example, we need to generate `*.jmx` and `*.web` property meta-data. We probably shouldn't try to over complicate things.

philwebb:
Actually, now the package restructure is in I would like to reconsider our options. Something doesn't feel quite right about `EndpointExposure`, especially as it's really only used to limit a couple of endpoints.

The `jmx` and `web` meta-data might also be worth a revisit. Enabling on a per technology basis is adding complexity. Perhaps we can just get away with a single `enabled`?

wilkinsona:
> Perhaps we can just get away with a single enabled?

The need for per-technology enablement is driven by the shutdown endpoint. For an app that doesn't need HTTP security, I think it's useful to be able to only expose the shutdown endpoint over JMX otherwise anyone with HTTP access to the app could shut it down.

snicoll:
I second that and disagree with the complex argument. Things are located in a single class and the contract is isolated. I agree this particular class is a bit complex but that doesn't leak elsewhere.

philwebb:
I'll try to spike something then we can make a decision.

philwebb:
Still a way to go, but I've started experimenting [here](https://github.com/philwebb/spring-boot/tree/gh-10176).

Trying an annotation approach for both specializations and extensions:

```java
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@Documented
@EndpointExtension(filter = WebEndpointFilter.class) // Extension that only applies when the filter matches
public @interface EndpointWebExtension {

	@AliasFor(annotation = EndpointExtension.class)
	Class<?> endpoint();

}
```

```java
@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@Documented
@Endpoint
@FilteredEndpoint(WebEndpointFilter.class) // Endpoint only discovered when the filter matches
public @interface WebEndpoint {

	@AliasFor(annotation = Endpoint.class)
	String id();

	@AliasFor(annotation = Endpoint.class)
	boolean enableByDefault() default true;

}
```

PR #11091, opened 2017-11-21
Fix references to management related properties

vpavic:
There are quite a few places where references to `endpoints.*` properties haven't been updated to `management.endpoint[s].*` introduced in #10176.


spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/documentation/ShutdownEndpointDocumentationTests.java
```diff
@@ -37,7 +37,7 @@
  *
  * @author Andy Wilkinson
  */
-@TestPropertySource(properties = "endpoints.shutdown.enabled=true")
+@TestPropertySource(properties = "management.endpoint.shutdown.enabled=true")
```
wilkinsona: This one's redundant and can be removed entirely.

snicoll:
Thanks @vpavic, I went ahead and changed the sample to expose everything instead (that was the original intention with `management.security.enabled=false` in `1.5.x` anyway)

PR #11117, opened 2017-11-22
Extract social starters/autoconfig

habuma:
Remove Spring Social autoconfig and starters from Spring Boot, as autoconfig has been absorbed into Spring Social itself.


wilkinsona:
Thanks, @habuma. This looks pretty comprehensive. There are some properties listed in appendix-application-properties.adoc that should be removed too.

artembilan:
The Spring Integration PR (https://github.com/spring-projects/spring-integration/pull/2292) has been merged and we have started to fail against IO (https://build.spring.io/browse/INT-FATS5IC-338/). I'm pretty sure just because of still supporting Spring Social in Spring Boot.

@wilkinsona , I don't see any green lines in this PR - only removal. Why do you find it as "comprehensive".

@habuma , I see there is a conflict - would be better to rebase and resolve them.

After this I guess we have to add Spring Social into the IO Platform dependencies since Boot won't manage it any more.

If you are going to release today, would be better to pay attention to this PR - Spring Integration are not compatible otherwise.

Thanks

wilkinsona:
I wasn't expecting the SI PR to be merged until everyone had indicated that we were ready to proceed.

artembilan:
Hm...

We are in GA phase there and you have Milestone for Boot today. Therefore we have to release something for you today anyway. If we don't do that change in SI today, we have a problem with our GA, I guess.

On the other hand you didn't tell anything before, and @snicoll indicated that Initializr is going to be fine to stick with Spring Social `1.x`. That sounded like the decision has been landed.

What ma I missing and how to proceed then, @wilkinsona ?

Thanks

/CC @garyrussell

wilkinsona:
It was @habuma who asked us to wait till everyone was ready. However, given the need to get SI 5.0 release out of the door, I guess we don't have too much choice. I'll get this into Boot M7 and update the Platform to manage the older versions of Social that SI now uses.

artembilan:
Thanks, @wilkinsona , for flexible vision.

Well, it is not "older versions" - it is current. There is no `2.0` for Spring Social 😉

PR #11120, opened 2017-11-23
Improve extension of Neo4jHealthIndicator

espiegelberg:
I've gotten some feedback from colleagues that they would like to customize the Cypher validation query within the Neo4jHealthIndicator. I agree that this would be useful. This PR makes such customization possible through configuration. 

Neo4jHealthIndicator.java on line 44 could use a second set of eyes; please let me know how you prefer this to be coded. It previously was static and **final** but is now no longer final, to allow it to be changed. The way it is currently coded may not match your preferred style.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/Health.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/Health.java
@@ -230,6 +230,17 @@ public final class Health {
 		}
 
 		/**
+		 * Record details using the given {@link Map}.
+		 * @param details The details as a {@link Map}
+		 * @return this {@link Builder} instance
+		 */
+		public Builder withDetails(Map<String, Object> details) {
+			Assert.notNull(details, "Details must not be null");
+			this.details = details;
+			return this;
+		}
+
+		/**
 		 * Set status to {@link Status#UNKNOWN} status.
 		 * @return this {@link Builder} instance
 		 */
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/Health.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/Health.java
@@ -230,17 +230,6 @@ public final class Health {
 		}
 
 		/**
-		 * Record details using the given {@link Map}.
-		 * @param details The details as a {@link Map}
-		 * @return this {@link Builder} instance
-		 */
-		public Builder withDetails(Map<String, Object> details) {
-			Assert.notNull(details, "Details must not be null");
-			this.details = details;
-			return this;
-		}
-
-		/**
 		 * Set status to {@link Status#UNKNOWN} status.
 		 * @return this {@link Builder} instance
 		 */
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicator.java
@@ -38,21 +38,12 @@ import org.springframework.boot.actuate.health.HealthIndicator;
  */
 public class Neo4jHealthIndicator extends AbstractHealthIndicator {
 
-	/**
-	 * The Cypher statement used to verify Neo4j is up.
-	 */
-	static String CYPHER = "match (n) return count(n) as nodes";
-
 	private final SessionFactory sessionFactory;
 
 	/**
-	 * Create a new {@link Neo4jHealthIndicator} using the specified
-	 * {@link SessionFactory}.
-	 * @param sessionFactory the SessionFactory
+	 * The Cypher statement used to verify Neo4j is up.
 	 */
-	public Neo4jHealthIndicator(SessionFactory sessionFactory) {
-		this.sessionFactory = sessionFactory;
-	}
+	private final String cypher;
 
 	/**
 	 * Create a new {@link Neo4jHealthIndicator} using the specified
@@ -62,20 +53,20 @@ public class Neo4jHealthIndicator extends AbstractHealthIndicator {
 	 */
 	public Neo4jHealthIndicator(SessionFactory sessionFactory, String cypher) {
 		this.sessionFactory = sessionFactory;
-		CYPHER = cypher;
+		this.cypher = cypher;
 	}
 
 	@Override
 	protected void doHealthCheck(Health.Builder builder) throws Exception {
 		Session session = this.sessionFactory.openSession();
-		Result result = session.query(CYPHER, Collections.emptyMap());
+		Result result = session.query(this.cypher, Collections.emptyMap());
 		Map<String, Object> details = new HashMap<>();
 
 		for (Map<String, Object> queryResult : result.queryResults()) {
 			details.putAll(queryResult);
 		}
 
-		builder.up().withDetails(details);
+		builder.up().withDetail("results", details);
 	}
 
 }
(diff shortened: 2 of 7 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/Health.java
```diff
@@ -229,6 +229,17 @@ public Builder withDetail(String key, Object value) {
 			return this;
 		}
 
+		/**
+		 * Record details using the given {@link Map}.
+		 * @param details The details as a {@link Map}
+		 * @return this {@link Builder} instance
+		 */
+		public Builder withDetails(Map<String, Object> details) {
```
snicoll: That doesn't sound right. You can pass a `Map` to the regular method and we don't want to offer a feature that would wipe out all the details (that's why you  have to use a key)

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicator.java
```diff
@@ -39,7 +41,7 @@
 	/**
 	 * The Cypher statement used to verify Neo4j is up.
 	 */
-	static final String CYPHER = "match (n) return count(n) as nodes";
+	static String CYPHER = "match (n) return count(n) as nodes";
```
snicoll: If the `CYPER` is configurable this should stay final (`DEFAULT_CYPHER`) and you should have a regular `cypher` field that's initialized with the default

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicator.java
```diff
@@ -52,12 +54,28 @@ public Neo4jHealthIndicator(SessionFactory sessionFactory) {
...
 		Result result = session.query(CYPHER, Collections.emptyMap());
-		builder.up().withDetail("nodes",
-				result.queryResults().iterator().next().get("nodes"));
+		Map<String, Object> details = new HashMap<>();
+
+		for (Map<String, Object> queryResult : result.queryResults()) {
+			details.putAll(queryResult);
+		}
+
+		builder.up().withDetails(details);
```
snicoll: This should go away IMO, that's unrelated to making the cyper configurable.

spring-boot-project/spring-boot-docs/src/main/asciidoc/production-ready-features.adoc
```diff
@@ -381,6 +381,9 @@ The following `HealthIndicators` are auto-configured by Spring Boot when appropr
 |{sc-spring-boot-actuator}/mongo/MongoHealthIndicator.{sc-ext}[`MongoHealthIndicator`]
 |Checks that a Mongo database is up.
 
+|{sc-spring-boot-actuator}/neo4j/Neo4jHealthIndicator.{sc-ext}[`Neo4jHealthIndicator`]
+|Checks that a Neo4j server is up.
+
```
snicoll: This is already merged. Please rebase and squash your branch (you can push force on it to update this PR).

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicatorTests.java
```diff
@@ -86,4 +86,33 @@ public void neo4jDown() {
 		assertThat(health.getStatus()).isEqualTo(Status.DOWN);
 	}
 
+	@Test
+	public void customCypher() {
```
snicoll: What does this test tests exactly? If we offer a way to specify a custom cyper, I'd rather have a way for the user to configure that via a key then (`management.health.neo4j.cypher` or something like that).

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicator.java
espiegelberg: The previous code relies on a "nodes" value to be available in the results returned from Neo4j. As this PR makes the Cypher validation statement configurable, the code can no longer rely on this value to be present. 

The proposed code change instead iterates over the returned results, putting them into a Map, and stores this Map as the details. This way whatever results are present will be made available as details.

Do you have suggestions on how this can be improved?

espiegelberg:
Thanks @snicoll for the feedback.

Your suggestion to provide a way for the user to configure the cypher via a key (management.health.neo4j.cypher) was an improvement I hadn’t thought of. I have decided to go that route.

Can you please review these changes/updates and let me know any way they or this PR can be improved? I expect there will be some desired changes but am happy to iterate until the PR is ready to be merged.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicator.java
snicoll: I am afraid I don't know much Neo4j but adding "whatever results are present" doesn't sound good to me. There must be some sort of minimal set of data (we can document what the cypher can look like).

Regardless, there is absolutely no reason to change the API of the builder for that.

snicoll:
Thanks for the update @espiegelberg! 

This looks better but I am feeling uneasy with the "add whatever the cypher returns" part. Is there a way to  constraint that somehow? Perhaps you can expand why someone would want to customize the cypher?

espiegelberg:
I think it's important that spring-boot provide a reasonable default but also provide end-users with a configurable way to customize the cypher. Why? Even within my own team there is some conversation/debate as to what the preferred cypher validation statement is or should be. This leads me to believe that different teams and different applications will benefit from making the customization of the cypher easy to do.

What is your concern about the  "add whatever the cypher returns" part? While that description makes it sound very open ended, in my view what is put into the details *is* constrained... to the values returned from the cypher statement specified by the user.

If your concern is that the current coding could return too much data, such as if the user provided a query that returned numerous rows of query results, we could go the route of truncating the results, such as only adding the first query result row to the result object. While I am of course open to that, this may artificially restrict the user's ability to return the details they desire (such as if they truly wanted to return 2 or more rows to details) and is the same thing as if the user specified a cypher statement that included a "limit 1" clause. So the restriction we put into code can easily be accomplished within the user's cypher statement.

Please give this some thought and let's collaborate on a solution.

snicoll:
>  in my view what is put into the details is constrained

That's one to look at it. The way I look at it is that we'd offer a configuration option that has a major impact on the output format of the health indicator. We try to avoid that (for instance, that's why we don't have spEL support in `application.properties`).

To be honest, I don't get why this is so important so perhaps someone who actually knows neo4j could chime in and share his opinion?

ping @mp911de, @christophstrobl @olivergierke

meistermeier:
From our point of view the idea behind customizable queries sounds reasonable, especially for more complex ones e.g. calling a procedure that returns the health of a cluster. But in my opinion this is nothing I would expect to be supported by a "framework" like spring boot. The idea of this health checks is a simple 'cypher ping' like the current query that works with every instance (version and edition).
It would be a better option to implement the health indicator in the application itself because this generic way of returning the result can lead to output that could be very wordy (e.g. lots of data) and even unforeseen (e.g. sensitive data).

snicoll:
> It would be a better option to implement the health indicator in the application itself 

Fully agree with that. So perhaps we should restructure the current indicator so that it can easily be extended. And defining a bean of that type should be enough to swap the one Spring  Boot provides by default. @espiegelberg would you be willing to rework the PR in that direction?

espiegelberg:
Yes, if you are willing to outline the changes I am willing to rework the PR in that direction.

snicoll:
@espiegelberg thanks. Looking at summarizing the changes necessary, it turns out it was faster to actually implement them so I've just done that.

Issue #9882, opened 2017-07-27
 Livereload - spring.devtools.livereload.exclude


dsu:
`File Watcher` thread takes about of 98% of entire application memory.Memory consumption is still growing after application start. 
It is also impossible to exclude log files from third party libraries that need to write to files under resources directory. 
It would be nice to be able to exclude files by location or by file extension. Something similar to `restart.include/restart.exclude` properties.

wilkinsona:
Sorry, I don't understand your description of the problem.

> File Watcher thread takes about of 98% of entire application memory

Roughly speaking, a thread consumes a fixed amount of memory that's determined by the JVM's configured thread stack size. A thread itself shouldn't consume 98% of the available memory. Can you please clarify what you mean here?

> It is also impossible to exclude log files from third party libraries that need to write to files under resources directory.

Why would a library need to write its output to a location alongside your application's resources?

Before we can assess the need for an enhancement, we'll need to have a better understand of the problem that you're facing and why it's occurring. To help with that, can you please provide a small sample that we can clone or unzip that reproduces the problem?

dsu:
It just takes more memory than it should - [heapdump](https://ufile.io/g5d3i). I would like to reduce the number of files that are watched. 
Probably a few thousands of files under resources directory can help to reproduce the problem.
 
`Why would a library need to write its output to a location alongside your application's resources?`
All file paths rely on some configuration property. Library stores all the files and logs under a common root directory. Some of this files need to be accessible as a resources. This is just an example.  What does it matter?

wilkinsona:
Thanks.

> What does it matter?

It matters because we can make a change or enhancement without first fully understanding the problem that we're trying to solve. On that note, as I said above, can you please provide a small sample that we can clone or unzip that reproduces the problem?

dsu:
Imagine an app that needs to generate thousands of images that needs to be accessible by a reverse proxy..

`package com;

import java.io.BufferedWriter;
import java.io.File;
import java.io.IOException;
import java.io.RandomAccessFile;
import java.nio.file.Files;
import java.nio.file.Path;

import javax.annotation.PostConstruct;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.stereotype.Component;

@SpringBootApplication
public class WatcherKillApp {

	public static void main(String[] args) {
		System.out.println("WatcherKillApp");
		SpringApplication.run(WatcherKillApp.class, args);
[...]

PR #11125, opened 2017-11-23
FileSystemWatcher alternative - performance improvement

dsu:
Introduces new interface and a new class FileSystemNIOWatcher. It doesn't require any file system snapshots and it can wait for file change - it doesn't need to check every file every second. It would be a great impact on CPU and RAM usage when you have a lot of files in your project.
Property `spring.devtools.restart.nio=true` needs to be set on application.properties.
I can see that there is a FileFilter in FileSystemWatcher that would probably allow for filtering files that can trigger restart or reload. I have copied it to FileSystemNIOWatcher. There is an interface FileWatcher. I would be nice if FileSystemWatcher implements it.
Fixes gh-9882


wilkinsona:
@dsu Many thanks for the PR. While a NIO-based watcher is definitely a worthwhile enhancement, I think we'd like to implement it in a different way to what you have proposed here. The current approach of sub-classing `FileSystemWatcher` is problematic as it's required you to duplicate a number of fields and to override a number of methods without calling `super`. You're also using two new monitor objects and method-level synchronisation in addition to the existing monitor in the super-class. Unfortunately, that threading model is too complex for us to merge and then maintain. Thanks anyway.

PR #11128, opened 2017-11-23
replaced explicit null-checks by @NonNull for target method parameters

denis-zhdanov:
It's rather tedious to write explicit null-checks (Assert.notNull()) for target method parameters.  
An alternative is to mark those parameters by *@NotNull* annotation and generate the checks during compilation. Benefits:  
* less code (it's not necessary to write Assert.notNull())
* better contract definition - it's immediately clear that particular method parameter should not be *null*

The magic is done by the [Traute](http://traute.oss.harmonysoft.tech/) javac plugin, it's configured in the *spring-boot-parent*'s *pom.xml*.  

Please note that all tests pass from the command line (*mvn package*) but tests which ensure *null*-checks processing fail from the *IDE* at the moment (at least IntelliJ), e.g. [AbstractFilterRegistrationBeanTests.setServletRegistrationBeanMustNotBeNull()](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/servlet/AbstractFilterRegistrationBeanTests.java#L131). We can use at least two approaches to fix that:
1. Remove all tests which check *@NonNull* parameters validation
2. Modify those *null*-check tests in a way to stop processing unless particular system property (like *'ci'*) is defined, i.e. run them on CI server but skip during *IDE* runs
I recommend the second approach, along with the official recommendation to avoid creating such tests in future. Also, I'm fine with preparing corresponding pull request when we agree on what is the desired approach.

TESTED: mvn clean package

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveCloudFoundrySecurityService.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveCloudFoundrySecurityService.java
@@ -27,7 +27,7 @@ import org.springframework.boot.actuate.autoconfigure.cloudfoundry.CloudFoundryA
 import org.springframework.boot.actuate.autoconfigure.cloudfoundry.CloudFoundryAuthorizationException.Reason;
 import org.springframework.core.ParameterizedTypeReference;
 import org.springframework.http.HttpStatus;
-import org.springframework.util.Assert;
+import org.springframework.lang.NonNull;
 import org.springframework.web.reactive.function.client.WebClient;
 import org.springframework.web.reactive.function.client.WebClient.RequestHeadersSpec;
 import org.springframework.web.reactive.function.client.WebClientResponseException;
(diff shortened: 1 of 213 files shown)


philwebb:
Thanks very much for the suggestion. This certainly looks like an interesting approach that could reduce some duplication (especially as we want to add those annotations anyway to help with Kotlin support).

Unfortunately, at least for now, I don't think we should adopt this approach in Spring Boot itself. I'm keen to keep us aligned with what Spring Framework currently does, and they have kept the direct `Assert` calls.

I've also found some issue with the way that IDEs handle annotation processors. Our current approach has been to only use them to generate additional data. This means if the IDE fails, contributors can still work on the code and run tests in the IDE.

In our case, I think the explicitness of the `Assert` calls, combined with the fact they work well in all IDEs, means we should suffer the cost of a little duplication.

philwebb:
I just realized you're the author of Traute. I sorry we couldn't use it ourselves, but I'm sure for many projects will find it a very nice way to reduce boilerplate. Good luck with project!

denis-zhdanov:
Hi @philwebb,

Thanks for the feedback!

Let me make a couple of clarifications to ensure we're on the same page here - as far as I see, you're saying that there are two obstacles for applying *Traute* to *Spring-Boot* at the moment:
* *Traute* is not applied to the *Core Spring*
* there are existing tests which ensure *null*-checks and they would fail when ran from *IDE*

Regarding the first point - I can try to persuade *Core Spring* team to apply *Traute*. If that is done, we still need to work on the second point. My vision is that we don't need *null*-checks tests at all as the solution is fully declarative and automatic, i.e. we only need to decide what to do with existing tests. My suggestion is to modify them in a way like below:  
```java
@Test
public void myNullCheckTest() {
    if (!Boolean.parseBoolean("CI") {
        return;
    }
    // Do the check
}
```
That way all tests pass in *IDE* and we still run them on the *CI* server to ensure that *Traute* produces the expected result.  

So, the question is if we can return to the 'apply *Traute* to *Spring Boot*' question if *Spring Core* accepts it? (I'm fine with expanding current pull request in a way to modify existing *null*-tests in the way mentioned above, i.e. this wouldn't require any effort from the *Spring Boot* team).

P.S. *Traute* is not an annotation processor, it's a *javac* plugin, i.e. it doesn't produce additional content but enhances resulting byte code.

philwebb:
I'm happy to see what the Framework Team think, but honestly I think we'd be trading one set of problems for another if we switch. The explicitness of the `Assert` calls makes them easy to understand and to debug and we know they work absolutely everywhere. I'm just not sure that needing to change the tests and add an extra build dependency would be sensible.

denis-zhdanov:
Thanks for the honesty 👍 
But don't you think that one compile-time dependency worth removing all those explicit *null*-checks? ;)

PR #11143, opened 2017-11-24
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTest.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTest.java
@@ -43,7 +43,7 @@ import org.springframework.test.web.reactive.server.WebTestClient;
  * <p>
  * Using this annotation will disable full auto-configuration and instead apply only
  * configuration relevant to WebFlux tests (i.e. {@code @Controller},
- * {@code @ControllerAdvice}, {@code @JsonComponent}, {@code Converter}, and
+ * {@code @ControllerAdvice}, {@code @JsonComponent}, {@code Converter}/{@code GenericConverter}, and
  * {@code WebFluxConfigurer} beans but not {@code @Component}, {@code @Service} or
  * {@code @Repository} beans).
  * <p>
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/servlet/mockmvc/ExampleController2.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/servlet/mockmvc/ExampleController2.java
@@ -32,13 +32,13 @@ public class ExampleController2 {
 
 	@GetMapping("/two")
 	@ResponseBody
-	public String one(ExampleArgument argument) {
+	public String two(ExampleArgument argument) {
 		return argument + "two";
 	}
 
 	@GetMapping("/two/{id}")
 	@ResponseBody
-	public String one(@PathVariable ExampleId id) {
+	public String two(@PathVariable ExampleId id) {
 		return id.getId() + "two";
 	}
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveRepositoriesAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveRepositoriesAutoConfigurationTests.java
@@ -80,7 +80,7 @@ public class CouchbaseReactiveRepositoriesAutoConfigurationTests {
 
 	@Test
 	public void doesNotTriggerDefaultRepositoryDetectionIfCustomized() {
-		load(CouchbaseReactiveRepositoriesAutoConfigurationTests.CustomizedConfiguration.class);
+		load(CustomizedConfiguration.class);
 		assertThat(this.context.getBeansOfType(ReactiveCityCouchbaseRepository.class))
 				.isEmpty();
 	}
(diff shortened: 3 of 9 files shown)


spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/reactive/webclient/ExampleId.java
```diff
@@ -18,10 +18,8 @@
 
 import java.util.UUID;
 
-import org.springframework.core.convert.converter.Converter;
-
 /**
- * An example attribute that requires a {@link Converter}.
+ * An example attribute that requires a {@link org.springframework.core.convert.converter.Converter}.
```
snicoll: Can you expand why you think this is a good idea?

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTest.java
```diff
@@ -43,7 +43,7 @@
  * <p>
  * Using this annotation will disable full auto-configuration and instead apply only
  * configuration relevant to WebFlux tests (i.e. {@code @Controller},
- * {@code @ControllerAdvice}, {@code @JsonComponent}, {@code Converter}, and
+ * {@code @ControllerAdvice}, {@code @JsonComponent}, {@code Converter}, {@code GenericConverter}, and
```
snicoll: I initially wrote that and I removed it as I felt it was a bit of a noise. So yeah if we scan a "converter" that includes the `GenericConverter` companion interface. Perhaps there is a better way to phrase that?

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/servlet/mockmvc/ExampleController2.java
```diff
@@ -32,13 +32,13 @@
 
 	@GetMapping("/two")
 	@ResponseBody
-	public String one(ExampleArgument argument) {
+	public String two(ExampleArgument argument) {
```
snicoll: 😓

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTest.java
izeye: @snicoll The reason why I added `{@code GenericConverter}` was that `{@code Converter}` is decorated with `@code` and as a code it didn't look to include `GenericConverter`. If it was "converter", it would be okay. I'm not sure how it can be improved ATM but maybe `{@code Converter}/{@code GenericConverter}`?

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/reactive/webclient/ExampleId.java
izeye: @snicoll I'm not sure this is generally acceptable or just my preference but the rationale behind the change is that when an imported class is referenced only in Javadoc, to prevent it from cluttering `import` section, I usually prefer to use FQN in Javadoc instead of `import`.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTest.java
snicoll: Yeah I like that

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/reactive/webclient/ExampleId.java
snicoll: That sounds totally reasonable but that's not what we do here.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTest.java
izeye: @snicoll Sure, will do.

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/reactive/webclient/ExampleId.java
izeye: @snicoll Sorry but I don't get it. Could you give me some more pointer for the last comment?

izeye: @snicoll I guess you meant you want to revert the change, right?

snicoll: I do. Here [is an example that we use quite a lot](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfiguration.java#L56).

Having said that, I've seen that is not consistent, see [AopAutoConfiguration](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/aop/AopAutoConfiguration.java#L29) for instance. I suggest to revert for now and I'll raise that at the next team call to check with the rest of the team.

izeye:
@snicoll Thanks for reviewing this. I changed as discussed.

Issue #8606, opened 2017-03-14
Add @DataCouchbaseTest support


philwebb:
For completeness it would be nice if we had a complete set of test annotations for Spring Data. See #8544

alexsderkach:
The are 2 ways to achieve this:
- CouchbaseMock (Doesn't support N1QL)
- Launch Couchbase TestContainers (Requires Docker Machine & pulls testcontainers dependency)

Approve which way to go (I prefer first one), and I will implement this task.

philwebb:
I don't have detailed knowledge of Couchbase but the first option certainly seems like it would have fewer moving parts.

wilkinsona:
I don't either, but I too prefer the first option as requiring Docker Machine feels a bit too much to me.

snicoll:
That seems incompatible with the concept of slicing/unit test as well. So +1 for the first option.

alexsderkach:
Blocked, until https://github.com/spring-projects/spring-data-couchbase/pull/153 is merged.

mp911de:
spring-projects/spring-data-couchbase#153 is fixed now.

wilkinsona:
Closing in favour of the PR (#11146).

PR #11146, opened 2017-11-25
Add @DataCouchbaseTest support

alexsderkach:
This PR resolves https://github.com/spring-projects/spring-boot/issues/8606

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
@@ -17,7 +17,11 @@
 package org.springframework.boot.autoconfigure.couchbase;
 
 import java.time.Duration;
+import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
+
+import com.couchbase.client.core.env.DefaultCoreEnvironment;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.util.StringUtils;
@@ -27,6 +31,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Eddú Meléndez
  * @author Stephane Nicoll
+ * @author Alex Derkach
  * @since 1.4.0
  */
 @ConfigurationProperties(prefix = "spring.couchbase")
@@ -40,6 +45,26 @@ public class CouchbaseProperties {
 	private final Bucket bucket = new Bucket();
 
 	private final Env env = new Env();
+
+	private int bootstrapHttpDirectPort = DefaultCoreEnvironment.BOOTSTRAP_HTTP_DIRECT_PORT;
+
+	private int bootstrapCarrierDirectPort = DefaultCoreEnvironment.BOOTSTRAP_CARRIER_DIRECT_PORT;
+
+	public int getBootstrapHttpDirectPort() {
+		return this.bootstrapHttpDirectPort;
+	}
+
+	public void setBootstrapHttpDirectPort(int bootstrapHttpDirectPort) {
+		this.bootstrapHttpDirectPort = bootstrapHttpDirectPort;
+	}
+
+	public int getBootstrapCarrierDirectPort() {
+		return this.bootstrapCarrierDirectPort;
+	}
+
+	public void setBootstrapCarrierDirectPort(int bootstrapCarrierDirectPort) {
+		this.bootstrapCarrierDirectPort = bootstrapCarrierDirectPort;
+	}
 
 	public List<String> getBootstrapHosts() {
 		return this.bootstrapHosts;
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.couchbase.embedded;
+
+import java.io.IOException;
+import java.util.Collections;
+
+import com.couchbase.client.java.CouchbaseBucket;
+import com.couchbase.mock.Bucket;
+import com.couchbase.mock.BucketConfiguration;
+import com.couchbase.mock.CouchbaseMock;
+
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.couchbase.CouchbaseAutoConfiguration;
+import org.springframework.boot.autoconfigure.couchbase.CouchbaseProperties;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for Embedded Couchbase.
+ *
+ * @author Alex Derkach
+ * @since 2.0.0
+ */
+@Configuration
+@EnableConfigurationProperties({CouchbaseProperties.class})
+@AutoConfigureBefore(CouchbaseAutoConfiguration.class)
+@ConditionalOnClass({CouchbaseBucket.class, CouchbaseMock.class})
+public class EmbeddedCouchbaseAutoConfiguration {
+
+	@Bean(destroyMethod = "stop")
+	@ConditionalOnMissingBean
+	public CouchbaseMock couchbaseMock(CouchbaseProperties properties) throws IOException, InterruptedException {
+
+		CouchbaseProperties.Bucket bucketProperties = properties.getBucket();
+
+		BucketConfiguration bucketConfiguration = new BucketConfiguration();
+		bucketConfiguration.name = bucketProperties.getName();
+		bucketConfiguration.password = bucketProperties.getPassword();
+		bucketConfiguration.type = Bucket.BucketType.COUCHBASE;
+		bucketConfiguration.numReplicas = 1;
+		bucketConfiguration.numNodes = 1;
+
+		CouchbaseMock couchbaseMock = new CouchbaseMock(0, Collections.singletonList(bucketConfiguration));
+		couchbaseMock.start();
+		couchbaseMock.waitForStartup();
+
+		String host = couchbaseMock.getHttpHost();
+		int bootstrapHttpDirectPort = couchbaseMock.getHttpPort();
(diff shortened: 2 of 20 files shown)


alexsderkach:
@Aloren sounds good, updated doc

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
```diff
@@ -41,8 +48,28 @@
 
 	private final Env env = new Env();
 
+	private int bootstrapHttpDirectPort = DefaultCoreEnvironment.BOOTSTRAP_HTTP_DIRECT_PORT;
```
snicoll: What are those two properties? Why are they necessary?

The current keys are structure and those two "boostrap"  related keys aren't. They aren't documented either.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
```diff
@@ -41,8 +48,28 @@
...
+		return this.bootstrapCarrierDirectPort;
+	}
+
+	public void setBootstrapCarrierDirectPort(int bootstrapCarrierDirectPort) {
+		this.bootstrapCarrierDirectPort = bootstrapCarrierDirectPort;
+	}
+
 	public List<String> getBootstrapHosts() {
-		return this.bootstrapHosts;
+		return Objects.nonNull(this.bootstrapHosts) ? this.bootstrapHosts : DEFAULT_HOSTS;
```
snicoll: I don't think we should change that there. If the boostrap hosts aren't set, it should return `null` as before. Adding a logic of a "default" host while the property has no default value is wrong.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+		setEnvironmentProperties(this.applicationContext, host, bootstrapHttpDirectPort, bootstrapCarrierDirectPort);
+
+		return couchbaseMock;
+	}
+
+	private void setEnvironmentProperties(ApplicationContext currentContext,
+			String host,
+			int bootstrapHttpDirectPort,
+			int bootstrapCarrierDirectPort) {
+		if (currentContext instanceof ConfigurableApplicationContext) {
```
snicoll: This is fragile and should be avoided. It looks like that this auto-configuration should expose a `CouchbaseConfiguration` rather than augmenting the Environment so late in the game.

If we opened `CouchbaseConfiguration` a bit (make that its own class with the necessary parameters that you need), then we could expose one here and leave the original auto-configuration unchanged (it would simply backoff as it does already prior to this change).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
```diff
@@ -0,0 +1,151 @@
...
+	 * Additional configuration to ensure that {@link CouchbaseEnvironment} beans depend on the
+	 * {@code couchbaseMock} bean.
+	 */
+	@Configuration
+	@ConditionalOnClass({CouchbaseEnvironment.class})
+	protected static class EmbeddedCouchbaseDependencyConfiguration
+			implements BeanFactoryPostProcessor {
+
+		@Override
+		public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
```
snicoll: That's quite fragile IMO. There is no guarantee that `CouchbaseProperties` won't be processed differently in the future.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
Aloren: When CouchbaseMock is initialized on a random available port, these 2 ports are also changed from defaults to available ones. For a couchbase-client to be able to connect to CouchbaseMock we need to set these properties to the ports on which CouchbaseMock is actually listening.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
Aloren: When CouchbaseMock is initialized on a random available port, these 2 ports are also changed from defaults to available ones. For a couchbase-client to be able to connect to CouchbaseMock we need to set these properties to the ports on which CouchbaseMock is actually listening.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
alexsderkach: Correct, forgot to remove it.
The idea was, to create the same behavior as for mongo:
https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoProperties.java#L46

So that you can start couchbase-based application without any properties when CouchbaseMock is on classpath

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
alexsderkach: fixed

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
alexsderkach: removed this part

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+		couchbaseMock.start();
+		couchbaseMock.waitForStartup();
+
+		String host = couchbaseMock.getHttpHost();
+		int bootstrapHttpDirectPort = couchbaseMock.getHttpPort();
+		int bootstrapCarrierDirectPort = couchbaseMock.getBuckets().get(properties.getBucket().getName()).getServers()[0].getPort();
+
+		couchbaseMock.startHarakiriMonitor(host + ":" + bootstrapHttpDirectPort, false);
+
+		properties.setBootstrapHosts(Collections.singletonList(host));
```
snicoll: Thanks it's much better but that's not what I have in mind with "make that its own class with the necessary parameters that you need". This change is better as you're not relying on the order of processing of `CouchbaseProperties` but you're still changing its state and it is a singleton.

What I have more in mind is to apply those customizations to your extension of `SpringBootConfiguration`, either via a protected method of some sort or any other suitable means.

snicoll: Thanks it's much better but that's not what I have in mind with "make that its own class with the necessary parameters that you need". This change is better as you're not relying on the order of processing of `CouchbaseProperties` but you're still changing its state and it is a singleton.

What I have more in mind is to apply those customizations to your extension of `SpringBootConfiguration`, either via a protected method of some sort or any other suitable means.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
snicoll: So those properties must be set on the builder when `CouchbaseMock` is active but they don't need  to be exposed for the general use case? Again, that's calling for making those changes internally rather than exposing them.

snicoll:
It looks like that a callback with `DefaultCouchbaseEnvironment.Builder` should work. I am giving that a try now.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseProperties.java
Aloren: you mean `DefaultCouchbaseEnvironment.Builder` ? yes, that would be much better. thank you

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+		bucketConfiguration.numReplicas = 1;
+		bucketConfiguration.numNodes = 1;
+
+		CouchbaseMock couchbaseMock = new CouchbaseMock(0, Collections.singletonList(bucketConfiguration));
+		couchbaseMock.start();
+		couchbaseMock.waitForStartup();
+
+		String host = couchbaseMock.getHttpHost();
+		int bootstrapHttpDirectPort = couchbaseMock.getHttpPort();
+		int bootstrapCarrierDirectPort = couchbaseMock.getBuckets().get(properties.getBucket().getName()).getServers()[0].getPort();
```
snicoll: Wow,  that invocation chain is deep. Isn't there a better way to get this? Is it 100% safe?

alexsderkach: >> The direct side effect is that Couchbase mock shadows anything now (if it's available we'll use that no matter what).
The approach is the same as for Mongo
https://github.com/spring-projects/spring-boot/blob/890c28dfdcb335d8551180f58e24439e7684d0cd/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/embedded/EmbeddedMongoAutoConfiguration.java

alexsderkach:
@snicoll 
> The direct side effect is that Couchbase mock shadows anything now (if it's available we'll use that no matter what).

The approach is the same as for Mongo
https://github.com/spring-projects/spring-boot/blob/890c28dfdcb335d8551180f58e24439e7684d0cd/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/embedded/EmbeddedMongoAutoConfiguration.java

snicoll:
I am aware of that.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/embedded/EmbeddedCouchbaseAutoConfiguration.java
alexsderkach: This chain is required to start CouchbaseMock at random port. It is possible to remove a few lines, but we will lose ability to start at random port.
It is safe, because CouchbaseMock is simply a HTTP server with 1 thread. All data is stored is ConcurrentHashMap.

snicoll:
So, after one more polish I just realized that `CouchbaseMock` is repackaging gson (and, of course, a different version of the one we use). 

This is a blocker as there's no way we'd add dependency management for such lib in Spring Boot.

snicoll:
I've opened https://github.com/couchbase/CouchbaseMock/issues/33

snicoll:
Given that CouchbaseMock is not officially supported by Couchbase and that we are not supposed to embed it for testing purpose, there is nothing we can do here I am afraid. 

I did improve our auto-configuration so that custom code can add support for `CouchbaseMock`. The state of my polish in my fork defines those changes. 

Thanks for the pr @alexsderkach !

PR #11149, opened 2017-11-25
Copy userNameAttributeName to ClientRegistration#Builder from configuration property

kazuki43zoo:
In current implementation, property value does not copy(or override) to `ClientRegistration#Builder.userNameAttributeName`.
I've fixed to copy(or override) the `userNameAttributeName` to `ClientRegistration#Builder` from configuration property (e.g. `spring.security.oauth2.client.provider.google.userNameAttribute=name`)

### Related Issue
* spring-projects/spring-security#4861


mbhave:
Thanks for the PR. Closing because it looks like a duplicate of #10672

PR #11151, opened 2017-11-26
Fix wrong annotation in CouchbaseReactiveRepositoriesAutoConfiguratio…

eddumelendez:
…nTests

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveRepositoriesAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveRepositoriesAutoConfigurationTests.java
@@ -23,7 +23,6 @@ import org.springframework.boot.autoconfigure.TestAutoConfigurationPackage;
 import org.springframework.boot.autoconfigure.context.PropertyPlaceholderAutoConfiguration;
 import org.springframework.boot.autoconfigure.couchbase.CouchbaseAutoConfiguration;
 import org.springframework.boot.autoconfigure.couchbase.CouchbaseTestConfigurer;
-import org.springframework.boot.autoconfigure.data.alt.couchbase.CityCouchbaseRepository;
 import org.springframework.boot.autoconfigure.data.alt.couchbase.ReactiveCityCouchbaseRepository;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveRepositoriesAutoConfigurationTests.java
```diff
@@ -82,7 +81,7 @@ public void noRepositoryAvailable() throws Exception {
 	public void doesNotTriggerDefaultRepositoryDetectionIfCustomized() {
 		load(CouchbaseReactiveRepositoriesAutoConfigurationTests.CustomizedConfiguration.class);
 		assertThat(this.context.getBeansOfType(ReactiveCityCouchbaseRepository.class))
-				.isEmpty();
+				.isNotNull();
```
snicoll: That assertion on something that returns a `Map` is wrong (it will always be non null). The purpose of this test is to check that the auto-configuration backs off as soon as there is a custom directive for the repositories.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveRepositoriesAutoConfigurationTests.java
```diff
@@ -116,7 +115,8 @@ private void load(Class<?> config, String... environment) {
 
 	@Configuration
 	@TestAutoConfigurationPackage(CouchbaseReactiveRepositoriesAutoConfigurationTests.class)
-	@EnableMongoRepositories(basePackageClasses = CityCouchbaseRepository.class)
+	@EnableReactiveCouchbaseRepositories(basePackageClasses = ReactiveCityCouchbaseRepository.class)
```
snicoll: While the copy/pasted annotation there was wrong, I think the intend was to `@EnableCouchbaseRepositories` and check that no reactive repositories were enabled. Changing that  breaks that test so we need to investigate a bit more what is going on.

snicoll:
So @mp911de helped me to track down the problem. It's an issue in Spring Data Couchbase that can mix imperative and reactive repositories. I've upgraded to the snapshot and the test passes now.

PR #11155, opened 2017-11-26
Rename ReactiveWebServerFactory.addContextCustomizers

izeye:
This PR renames `ReactiveWebServerFactory.addContextCustomizers` to align with its doing.


philwebb:
Good spot. I'd say we can probably call this `addServerCustomizer` since the class name already includes `Netty`. We should fix the parameter name used in the exception as well.

izeye:
@philwebb If so, `getNettyServerCustomizers()` and `setNettyServerCustomizers(Collection<? extends NettyServerCustomizer> nettyServerCustomizers)` also needs to be aligned with it (i.e. drop `Netty`)?

snicoll:
@izeye thanks for the PR, I've taken care of that suggestion in 55f7b3a

Issue #9502, opened 2017-06-13
InfluxDb. Add support for SSL in InfluxDbAutoconfiguration


Dichotomia:
We could use `InfluxDBFactory.connect(url, userName, password, builder)` instead of `InfluxDBFactory.connect(url, userName, password)`.

snicoll:
If you're using it with SSL, please expand the description to explain what you're doing currently.

Dichotomia:
Currently (with spring boot 1.5.4) we use https to secure the data sent to our influxDB database.

The code :

final SSLSocketFactory sslSocketFactory = ....
final okhttp3.OkHttpClient.Builder builder = new okhttp3.OkHttpClient.Builder().sslSocketFactory(sslSocketFactory);
....
final InfluxDB influxDB = InfluxDBFactory.connect(url, userName, password, builder)....
For spring boot 2.0, we plan to use influxDbAutoconfiguration if it enables to use https.

Dichotomia:
Suggestion in pseudo code : 
If (Bean okhttp3.OkHttpClient.Builder is present){ 
final InfluxDB influxDB = InfluxDBFactory.connect(url, userName, password, builder)
}else{
final InfluxDB influxDB = InfluxDBFactory.connect(url, userName, password)
}

wilkinsona:
It looks to me like @eddumelendez has taken care of this in #9669.

PR #9669, opened 2017-07-03
Add configuration for OkHttpClient.Builder in InfluxDbAutoConfiguration

eddumelendez:
This commit enables to auto-configure OkHttpClient.Builder in
InfluxDbAutoConfiguration if it is available.

See gh-9502

PR #11157, opened 2017-11-26
Add support for SSL in `InfluxDbAutoconfiguration`

alimate:
`spring.influx.ssl.enabled` property determines whether or not we're going to use SSL to interact with InfluxDB. If enabled, then the `spring.influx.ssl.certificate` represents the path to X509 formatted public key.

Fixes gh-9502


pivotal-issuemaster:
@alimate Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=11157)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=11157) to manually synchronize the status of this Pull Request.

wilkinsona:
I'm not sure that we should merge this, at least not in its current form.

#9502 was about allowing the client to use HTTPS to talk to the InfluxDB server. #9669 has already given the auto-configuration for support for using a custom `OkHttpClient.Builder` which can be used to enable that.

The proposal here seems to be specifically about connecting to an InfluxDB server that's using a certificate that's self-signed or otherwise not trusted by default. In this proposal, if you enable SSL it requires that you also provide a certificate to populate the client's trust store. For cases where the server's certificate is already trusted, that's unnecessary.

Also, the way that the certificate is provided is different to how we do things for configuring the web server's SSL. In this proposal, a bare certificate is provided and it's the wrapped up in a `KeyStore`. In the web server case, a key store is provided which simplifies the implementation.

My inclination is to decline this. The current custom `OkHttpClient.Builder` approach seems to be sufficient for most cases. If we want to do something to simplify the SSL case, I think it should be more closely modelled on what we've done for configuring web server SSL.

snicoll:
> My inclination is to decline this. The current custom OkHttpClient.Builder approach seems to be sufficient for most cases.

+1

wilkinsona:
Ok, given the reactions from @alimate and @snicoll's agreement, I'm going to close this one. Thank you anyway for the contribution.

PR #11161, opened 2017-11-27
Add CouchbaseHealthIndicatorTests

eddumelendez:


snicoll:
thanks @eddumelendez - weird that the feature was added without a test in the first place :(

Issue #11201, opened 2017-11-29
NullPointer exception in sessions actuator endpoint


ptahchiev:
Testing the sessions actuator endpoint with the latest 2.0.0.BUILD-SNAPSHOT, I do the following request:
````
curl -k -v --user admin:nimda -X GET "https://solar.local:8112/storefront/platform/sessions/blah"
````
where `blah` is invalid session id. Then I get the following exception:
````
2017-11-29 12:36:26,352 org.apache.catalina.core.ContainerBase.[Tomcat].[localhost].[/storefront].[dispatcherServlet] [https-jsse-nio-127.0.0.1-8112-exec-1] ERROR: Servlet.service() for servlet [dispatcherServlet] in context with path [/storefront] threw exception [Request processing failed; nested exception is java.lang.NullPointerException] with root cause
java.lang.NullPointerException
	at org.springframework.boot.actuate.session.SessionsEndpoint$SessionDescriptor.<init>(SessionsEndpoint.java:110)
	at org.springframework.boot.actuate.session.SessionsEndpoint.getSession(SessionsEndpoint.java:63)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:223)
	at org.springframework.boot.actuate.endpoint.reflect.ReflectiveOperationInvoker.invoke(ReflectiveOperationInvoker.java:69)
	at org.springframework.boot.actuate.endpoint.web.servlet.WebMvcEndpointHandlerMapping$OperationHandler.handle(WebMvcEndpointHandlerMapping.java:133)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)

````
I can provide a pull-requested if needed :D

snicoll:
Please do.

snicoll:
Closing in favour of #11202

PR #11202, opened 2017-11-29
Fix NullPointer when requesting a session that does not exist

ptahchiev:
Add a null-pointer check to avoid null-pointer exception in the `SessionEndpoint` actuator in case
there is no session for the given id.

Close: #11201


ptahchiev:
Done.

PR #11210, opened 2017-11-29
Support multiple ObjectMappers

gregturn:
If more than one ObjectMapper is registered in the application context, handle registering codecs for them all in JacksonCodecConfiguration.


wilkinsona:
Thanks for the PR, but this isn’t how default codecs work. There’s a single default Jackson codec. The proposed change will just use the last `ObjectMapper` in the list, rather than providing a default codec for each of them.

Issue #11113, opened 2017-11-22
Drop /status endpoint


philwebb:
We've discussed the `/health` and `/status` endpoints quite a bit and we were planning to switch them. After some more thought I'm thinking a simple property to enabled/disable the full details on `/health` might be enough.

The status and health endpoint are so similar in code and concept that I think having them both might actually be confusing. I think offering a single endpoint with a switch might be enough for most users.

If a user is running behind a firewall they can set the property to expose more details. If they are not running behind a firewall they can either stick with hidden details, or configure securirty.

The only use-case we've not covered very well is someone getting an alert from an insecure `/health` and then wanting to get more details about which specific aspect caused the problem. That seems like a fairly edge case so I'm tempted to see of people complain first.

It's pretty easy for us to re-introduce a new endpoint later. It's much harder for us to take an existing one away.

pkostrzewa:
@philwebb could I try to switch these endpoints?

philwebb:
@pkostrzewa Thanks for the offer, but I've already got this one in hand.

snicoll:
> That seems like a fairly edge case so I'm tempted to see of people complain first.

This is a feature of `1.5.x` that our users are using (given the number of complains we got that the configuration part was confusing). I don't think dropping this is a good idea.

bclozel:
Agree with @snicoll, I quite liked the separate endpoint as it made it easier to think about security and addressing two different use cases: automated health checks and incident response.

Now about the edge case side of this, I'd argue the other way around; unless you're deploying your app in an on-premise infrastructure and you can configure actuator on a separate port - any Spring Boot app serving public traffic would miss the health details.

Comparing this with our other issue (re: renaming the health endpoint), I think it's easier to reconfigure the path your health check client sends requests to (like AWS ELB health checks) than configuring it to send authenticated requests.

philwebb:
> I quite liked the separate endpoint as it made it easier to think about security and addressing two different use cases: automated health checks and incident response.

I guess I'm still not convinced that hitting a secured version of `/health` to get the additional information is something that actually happens during incident response. I don't know, perhaps it is, I'm hoping that this change will cause people that care to shout.

If you think about a monitoring system like Pingdom, the chances are that you want the full details payload to be returned. You want some history to be stored in Pingdom so that if the status changes back to healthy you still know what caused the problem.

I really don't know which way to jump with this one. I still think having both `/health` and `/status` feels wrong.

snicoll:
[...]

PR #11213, opened 2017-11-30
Update documentation references to `/status` endpoint

vpavic:
While migrating to `2.0.0.M7` I've discovered a few references to the removed `/status` endpoint. This was apparently missed in #11113.


snicoll:
Thanks a lot @vpavic!

PR #11233, opened 2017-12-01
Use volatile for ReactiveTokenValidator.cacheTokenKeys

izeye:
In `cacheTokenKeys()`, `clear()` and then `putAll()` are not atomic.

This PR changes to use `volatile` for `cacheTokenKeys` and also changes its type to `HashMap` because with `volatile` there's no modification now.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidator.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidator.java
@@ -25,6 +25,7 @@ import java.security.spec.InvalidKeySpecException;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.TimeUnit;
 
 import reactor.core.publisher.Mono;
@@ -43,7 +44,7 @@ class ReactiveTokenValidator {
 
 	private final ReactiveCloudFoundrySecurityService securityService;
 
-	private Map<String, String> cachedTokenKeys = new ConcurrentHashMap<>();
+	private volatile ConcurrentMap<String, String> cachedTokenKeys = new ConcurrentHashMap<>();
 
 	ReactiveTokenValidator(ReactiveCloudFoundrySecurityService securityService) {
 		this.securityService = securityService;
@@ -91,8 +92,7 @@ class ReactiveTokenValidator {
 	}
 
 	private void cacheTokenKeys(Map<String, String> tokenKeys) {
-		this.cachedTokenKeys.clear();
-		this.cachedTokenKeys.putAll(tokenKeys);
+		this.cachedTokenKeys = new ConcurrentHashMap<>(tokenKeys);
 	}
 
 	private boolean hasValidSignature(Token token, String key) {
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidatorTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidatorTests.java
@@ -26,8 +26,8 @@ import java.security.Signature;
 import java.security.spec.InvalidKeySpecException;
 import java.security.spec.PKCS8EncodedKeySpec;
 import java.util.Collections;
-import java.util.LinkedHashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.codec.binary.Base64;
 import org.junit.Before;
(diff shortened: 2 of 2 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidator.java
```diff
@@ -91,8 +92,7 @@
 	}
 
 	private void cacheTokenKeys(Map<String, String> tokenKeys) {
-		this.cachedTokenKeys.clear();
-		this.cachedTokenKeys.putAll(tokenKeys);
+		this.cachedTokenKeys = new HashMap<>(tokenKeys);
```
wilkinsona: I think this may need to be a `ConcurrentHashMap`. AFAIK, the field being volatile doesn't provide a happens-before relationship between the entries being added to the new map in the constructor and any subsequent reads or writes. That aside, I think it would be safer for it to be a `ConcurrentHashMap` anyway. If the code evolves such that the map is mutated, there would be a happens-before relationship between that mutation and any subsequent reads or writes.

WDYT, @izeye?

izeye: @wilkinsona If I understand it correctly, as the current status, it's (effectively) immutable (could be decorated with unmodifiable wrapper to be explicit), so I think there's no visibility issue but if you think there's any chance to be mutable, I don't have a strong opinion `HashMap`. Let me know your final call.

wilkinsona: `new HashMap<>(tokenKeys)` mutates the map prior to returning it. There's no locking in `HashMap` so I don't think there's any guarantee that the mutation that's done in the constructor will be visible to another thread. I think we should use `ConcurrentHashMap`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidator.java
```diff
@@ -43,7 +44,7 @@
 
 	private final ReactiveCloudFoundrySecurityService securityService;
 
-	private Map<String, String> cachedTokenKeys = new ConcurrentHashMap<>();
+	private volatile Map<String, String> cachedTokenKeys = Collections.emptyMap();
```
wilkinsona: Based on [this comment](https://github.com/spring-projects/spring-boot/pull/11233#discussion_r154333769), I think the field should be declared as a `ConcurrentMap` to avoid accidental use of a plain `Map`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidator.java
izeye: @wilkinsona Sorry, I didn't understand your previous comment fully. With the additional comment, now I got your point 👍 I'll update this PR as you suggested.

izeye:
@wilkinsona Thanks for the feedback! I updated this PR as you advised.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidatorTests.java
```diff
@@ -97,7 +98,7 @@ public void validateTokenWhenKidValidationFailsTwiceShouldThrowException()
 			throws Exception {
 		PublisherProbe<Map<String, String>> fetchTokenKeys = PublisherProbe
 				.of(Mono.just(VALID_KEYS));
-		ReflectionTestUtils.setField(this.tokenValidator, "cachedTokenKeys", VALID_KEYS);
+		ReflectionTestUtils.setField(this.tokenValidator, "cachedTokenKeys", new ConcurrentHashMap<>(VALID_KEYS));
```
wilkinsona: Couldn't `VALID_KEYS` be a `ConcurrentHashMap` instead?

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidatorTests.java
```diff
@@ -124,7 +125,7 @@ public void validateTokenWhenKidValidationSucceedsInTheSecondAttempt()
 		PublisherProbe<Map<String, String>> fetchTokenKeys = PublisherProbe
 				.of(Mono.just(VALID_KEYS));
 		ReflectionTestUtils.setField(this.tokenValidator, "cachedTokenKeys",
-				INVALID_KEYS);
+				new ConcurrentHashMap<>(INVALID_KEYS));
```
wilkinsona: Couldn't `INVALID_KEYS` be a `ConcurrentHashMap` instead?

wilkinsona:
Thanks for the update, @izeye

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/ReactiveTokenValidatorTests.java
izeye: @wilkinsona I just changed there because `VALID_KEYS ` is also shared by `this.securityService.fetchTokenKeys()` in the tests. Is it better to make `VALID_KEYS` to `ConcurrentHashMap` for you?

wilkinsona: I think so, yes. It feels less noisy to not be wrapping `INVALID_KEYS` each time.

izeye:
@wilkinsona Updated the tests as you suggested.

wilkinsona:
Thank you, @izeye!

PR #11235, opened 2017-12-01
Upgrade to Kotlin 1.2

sdeleuze:
With this pull request, Spring Boot 2 now builds against Kotlin 1.2 while still being Kotlin 1.1 compatible thanks to the usage of `apiVersion = 1.1` and `languageVersion = 1.1` configuration options, as confirmed with Kotlin team.

See [SPR-16239](https://jira.spring.io/browse/SPR-16239) related Spring Framework issue.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -105,7 +105,7 @@
 		<junit.version>4.12</junit.version>
 		<junit-jupiter.version>5.0.2</junit-jupiter.version>
 		<junit-platform.version>1.0.2</junit-platform.version>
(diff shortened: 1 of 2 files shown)


snicoll:
We need to find out how to handle that properly on start.spring.io. If we change this, a snapshot  of 2.0 would be on 1.2 with Maven and for Gradle (as long as we upgrade the version on the gradle side). But M7 would get inconsistent versions (1.2 for Gradle and whatever Spring Boot M7 manages for Maven).

snicoll:
Unfortunately, we need https://github.com/spring-io/initializr/issues/556

snicoll:
I've just upgraded to Kotlin 1.2.10 as a follow up of this issue.

PR #11240, opened 2017-12-01
Use LoggingSystemProperties constants

izeye:
This PR uses constants in `LoggingSystemProperties` whenever possible.


philwebb:
I'd like to run a structure-101 analysis of this just to make sure it's not introduced cycles.

wilkinsona:
There are no leaf package tangles introduced by the proposed changes.

wilkinsona:
Thank you, @izeye. I've [polished this a little](https://github.com/spring-projects/spring-boot/commit/54369a72caab34936755ffdd07439ff5edfcefbd), primarily to avoid using a logging constant in the `org.springframework.boot` package. While it didn't create a cycle, it didn't feel right conceptually for the logging system to be relied upon to get the PID. I've replaced the use of the system property with `ApplicationPid`.

Issue #6587, opened 2016-08-08
Ability to sanitize all keys in /env from a particular property source


spencergibb:
https://github.com/spring-cloud-incubator/spring-cloud-vault-config/issues/27

Spring Cloud Vault Config uses vault to store encrypted values and load them into the spring environment. Since all values are encrypted, it would be useful to blank the values of every key in an entire property source. The current workaround is to use the key matching.

joshiste:
I did a similar request in  #2922

wilkinsona:
@joshiste Thanks for the pointer.

@spencergibb What about https://github.com/spring-cloud/spring-cloud-commons/issues/45?

spencergibb:
@wilkinsona, @dsyer said to open an issue here since there's not really a way for an external property source to tell the `/env` actuator to sanitize all keys.

wilkinsona:
@spencergibb So we're circling back around to Phil's suggestion in #2922?

> Perhaps we should offer some strategy so that Spring Cloud can add it to change the sanitize rules

And you'll then make use of whatever we provide for any property sources where sanitisation for the whole source makes sense?

spencergibb:
Yes, sounds like a plan.

philwebb:
@spencergibb We're going de-scope this for now since we want to focus on 2.0 issues. If you've got something specific in mind and you can put together a pull-request we can try to get it back into 1.5.

spencergibb:
OK

aantono:
@here Can someone comment as to why not using EnumerablePropertySource implementation and instead extend regular PropertySource is not a good solution for this use case?  To my understanding both Spring and Boot are very good and checking for the type of property source and only list the keys if they are from an EnumerablePropertySource.  The Environment functionality is not really impacted, because it uses direct getProperty calls to fetch the values.  Am I missing something?

wilkinsona:
@aantono Being able to enumerate the keys in a property source is very useful for other purposes

aantono:
@wilkinsona In general, totally agree that being able to enumerate properties is a helpful thing to have.  When it comes to looking at a secret sensitive data set, what could be the possible good reason for being able to do so, besides viewing those in the `/env` endpoint?

wilkinsona:
@aantono Configuration binding via `@ConfigurationProperties` relies on property sources being enumerable. See `PropertySourcesPropertyValues`, for example.

pbetkier:
[...]

PR #11264, opened 2017-12-05
Allow customizing sanitization rules for /env

pbetkier:
This PR introduces more flexible sanitization rules for /env actuator endpoint. 

Sanitization rules can be customized by registering one or more beans of type `EnvironmentEndpoint.SanitizeFilter`. Such filters decide whether to sanitize a given value basing on its property name and the property source it originates from. By default there is only one filter, `KeyBlacklistSanitizeFilter`, which contains the already existing logic of sanitizing based on the matching key names.

I made a couple of backwards-incompatible changes which I assumed are OK at this release stage. I pointed them in the PR comments. I also haven't wrote the docs for this feature assuming it won't be commonly used.

@philwebb @spencergibb you may be interested in this :)

Fixes gh-6587

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/env/EnvironmentEndpointAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/env/EnvironmentEndpointAutoConfiguration.java
@@ -16,7 +16,13 @@
 
 package org.springframework.boot.actuate.autoconfigure.env;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Optional;
+
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.actuate.autoconfigure.endpoint.condition.ConditionalOnEnabledEndpoint;
+import org.springframework.boot.actuate.endpoint.sanitize.KeyBlacklistSanitizeFilter;
 import org.springframework.boot.actuate.env.EnvironmentEndpoint;
 import org.springframework.boot.actuate.env.EnvironmentEndpointWebExtension;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
@@ -32,6 +38,7 @@ import org.springframework.core.env.Environment;
  *
  * @author Phillip Webb
  * @author Stephane Nicoll
+ * @author Piotr Betkier
  * @since 2.0.0
  */
 @Configuration
@@ -48,13 +55,19 @@ public class EnvironmentEndpointAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean
 	@ConditionalOnEnabledEndpoint
-	public EnvironmentEndpoint environmentEndpoint(Environment environment) {
-		EnvironmentEndpoint endpoint = new EnvironmentEndpoint(environment);
+	public EnvironmentEndpoint environmentEndpoint(Environment environment,
+			ObjectProvider<Collection<EnvironmentEndpoint.SanitizeFilter>> customFilters) {
+
 		String[] keysToSanitize = this.properties.getKeysToSanitize();
-		if (keysToSanitize != null) {
-			endpoint.setKeysToSanitize(keysToSanitize);
-		}
-		return endpoint;
+		KeyBlacklistSanitizeFilter keyMatchFilter = (keysToSanitize != null)
+				? KeyBlacklistSanitizeFilter.matchingKeys(keysToSanitize)
+				: KeyBlacklistSanitizeFilter.matchingDefaultKeys();
+
+		Collection<EnvironmentEndpoint.SanitizeFilter> filters = new ArrayList<>();
+		filters.add(EnvironmentEndpoint.SanitizeFilter.from(keyMatchFilter));
+		Optional.ofNullable(customFilters.getIfAvailable()).ifPresent(filters::addAll);
+
+		return new EnvironmentEndpoint(environment, filters);
 	}
 
 	@Bean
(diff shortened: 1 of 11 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/sanitize/KeyBlacklistSanitizeFilter.java
```diff
@@ -14,58 +14,60 @@
...
 
 	/**
-	 * Keys that should be sanitized. Keys can be simple strings that the property ends
-	 * with or regular expressions.
-	 * @param keysToSanitize the keys to sanitize
+	 * Creates the filter based on the given blacklist.
+	 * @param keysToSanitize keys to match
+	 * @return a new filter instance
 	 */
-	public void setKeysToSanitize(String... keysToSanitize) {
```
pbetkier: I made this class configurable by constructor only, couldn't find out why it used setters. Point out if it's a problem.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/env/EnvironmentEndpoint.java
```diff
@@ -195,8 +206,8 @@ private void extract(String root, Map<String, PropertySource<?>> map,
 		}
 	}
 
-	public Object sanitize(String name, Object object) {
-		return this.sanitizer.sanitize(name, object);
+	private Object sanitize(PropertyNameAndSource source, Object object) {
```
pbetkier: Couldn't find out why this method was public, made it private since I changed the signature anyway. If we make this public again then `PropertyNameAndSource` must have a public constructor which it now doesn't.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/PropertySourcesPlaceholdersResolver.java
```diff
@@ -63,18 +64,22 @@ public Object resolvePlaceholders(Object value) {
...
 				if (value != null) {
-					return String.valueOf(value);
+					return convertResolvedValueToString(placeholder, value, source);
 				}
 			}
 		}
 		return null;
 	}
 
+	protected String convertResolvedValueToString(String placeholder, Object value, PropertySource<?> source) {
```
pbetkier: I changed the method that we allow to override, making the previous one private. I thought it's not a problem since `resolvePlaceholder` was made protected only recently (commit eb045f1da13083cf2cd6f4d45184c252969be76c).

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/env/EnvironmentEndpoint.java
```diff
@@ -55,21 +59,27 @@
...
 
 	private final Environment environment;
 
 	public EnvironmentEndpoint(Environment environment) {
-		this.environment = environment;
+		this(environment, Collections.singleton(SanitizeFilter
+				.from(KeyBlacklistSanitizeFilter.matchingDefaultKeys())));
 	}
 
-	public void setKeysToSanitize(String... keysToSanitize) {
```
pbetkier: Again, dropped the setter in favour of constructor.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/env/EnvironmentEndpointAutoConfiguration.java
```diff
@@ -48,13 +55,19 @@ public EnvironmentEndpointAutoConfiguration(
...
+			ObjectProvider<Collection<EnvironmentEndpoint.SanitizeFilter>> customFilters) {
+
 		String[] keysToSanitize = this.properties.getKeysToSanitize();
-		if (keysToSanitize != null) {
-			endpoint.setKeysToSanitize(keysToSanitize);
-		}
-		return endpoint;
+		KeyBlacklistSanitizeFilter keyMatchFilter = (keysToSanitize != null)
+				? KeyBlacklistSanitizeFilter.matchingKeys(keysToSanitize)
+				: KeyBlacklistSanitizeFilter.matchingDefaultKeys();
```
pbetkier: I wanted to create the instance of `KeyBlacklistSanitizeFilter` as a bean that we would autowire here, but couldn't express the condition "register the filter bean only if environment endpoint is enabled". Is there a way for expressing that?

pbetkier:
@philwebb can you have a look at this?

pbetkier:
@philwebb @spencergibb it's been 3 months, are you still interested in this feature? I won't bother resolving conflicts if you're not.

philwebb:
Hi @pbetkier, 

Thanks for the PR and sorry I didn't get back to you earlier. We're planning to start the planning process for Spring Boot 2.1 soon and I'd like to review this at that point. There's no need to resolve conflicts now, we can deal with that when we merge.

FWIW I really like the idea of supporting more flexible sanitization rules, we just ran out of time to get this into 2.0.

mbhave:
@pbetkier Sorry for the delay in getting to this. Given that the code has moved on quite a bit since the original PR, we decided to approach this in a different manner. [This commit](https://github.com/spring-projects/spring-boot/commit/253f98c3e7ae577618ca7d16d9116205f1fb4215) adds a new method to the `Sanitizer` and leaves the existing API as is. Users can plug in a custom `SanitizingFunction` which can make use of the `propertySource` or `key`. Thanks again and sorry for the wasted time.

PR #11290, opened 2017-12-07
Set tarLongFileMod=posix in maven-assembly-plugin for spring-boot-cli

JorgenRingen:
Building spring-boot with maven is failing on my system.
- MacOs Sierra 10.12.6
- Maven: 3.5
- JDK: 1.8.0_144

Error when doing `mvn clean install` from parent project:
```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-assembly-plugin:3.1.0:single (bin-package) on project spring-boot-cli: Execution bin-package of goal org.apache.maven.plugins:maven-assembly-plugin:3.1.0:single failed: group id '562225435' is too big ( > 2097151 ). Use STAR or POSIX extensions to overcome this limit ->
```

Possible solution accodring to [Apache Maven Assembly Plugin official documentation](https://maven.apache.org/plugins/maven-assembly-plugin/faq.html): 
```
Tar complains about groupid value being too big
GNU tar has restrictions on max value of UID/GUID in tar files. Consider using the more well defined POSIX tar format, which should support larger values. Use tarLongFileMode=posix in the assembly:single goal.
```

### DIFF
--- a/spring-boot-project/spring-boot-cli/pom.xml
+++ b/spring-boot-project/spring-boot-cli/pom.xml
@@ -267,6 +267,9 @@
 						</configuration>
 					</execution>
 				</executions>
+				<configuration>
+					<tarLongFileMode>posix</tarLongFileMode>
+				</configuration>
 			</plugin>
 			<plugin>
 				<groupId>org.apache.maven.plugins</groupId>


wilkinsona:
@JorgenRingen Thanks for the PR. Before we make any changes, I'd like to understand the failure a bit more. Is 562225435 the id of the group that contains the logged in user when you're building the project?

JorgenRingen:
@wilkinsona That's right. I'm not able to change this as I'm currently working from a "corporate-policy-configured-computer" without sufficient admin :-)

wilkinsona:
Thanks. It's a bit undesirable that the building user's UID and GUID are baked into the tar, but Maven's assembly plugin doesn't appear to provide a way to do anything else. AIUI, they're dropped during extraction unless that's done as root so it makes little difference.

My concern with switching to posix is that it will cause problems for anyone who doesn't have a compatible version of `tar` installed. I don't know how many people that may be, but it is a risk. I'm not sure that risk is worth taking for something that's specific to your environment. It may be better if you make the change locally. Let's see what the rest of the Boot team thinks.

philwebb:
I'm tempted to say we should try this change and see if anyone grumbles. The CLI is primarily installed on developer machines so I expect a modern version of `tar` would be available. We also ship a `.zip` file, so even if `tar` fails, there's always that as an option.

snicoll:
> I'm tempted to say we should try this change and see if anyone grumbles.

Considering the scenario and the impact, I am not sure I like that. I've +1 @wilkinsona initial proposal and it has my vote still.

wilkinsona:
Thanks for the PR, but we've decided that we don't want to make a change for an unusual build environment that, potentially, affects everyone who consumes the CLI's tar file. Our recommendation is that you either patch this locally or that you use a different user to run the build.

Issue #9917, opened 2017-07-31
Executable jar results in error when using relative imports in XSD


hoeckxer:
While upgrading an application to the latest version of Spring Boot (1.5.6.RELEASE), I got an error when running the application as the executable jar. After analysis, it seems like there is an issue introduced in the spring-boot-maven-plugin from 1.4.2.RELEASE on.

The application is hosting a webservice, and as such creates an `org.springframework.xml.xsd.commons.CommonsXsdSchemaCollection`, with a `ClasspathResources` pointing to an xsd. This XSD has an import of another XSD using a relative path.

I have attached a sample application that reproduces the problem. The only change needed to (not) show the issue is change the `spring-boot-maven-plugin` to `1.4.1.RELEASE` (working) or a new version (not working).

Please be aware that the issue doesn't occur when running from an IDE/spring-boot:run; it only happens when running an executable JAR.

I looked in the `JarURLConnection` commit history for the 1.4.2.RELEASE, but I couldn't find anything obvious standing out that could cause this.

[relative_import.zip](https://github.com/spring-projects/spring-boot/files/1187716/relative_import.zip)

wilkinsona:
I believe this duplicates #7581. See [this comment](https://github.com/spring-projects/spring-boot/issues/7581#issuecomment-265515992) for an explanation of the change.

Can you please try with a shaded jar (rather than a Spring Boot fat jar) and see if it fails then as well?

hoeckxer:
Thanks for the reply.

I modified the pom of the sample application as per the comment you referred to, and updated the application to directly create the CommonsXsdSchemaCollection:

`public class App
{
	public static void main( String[] args )
	{
//		SpringApplication.run(App.class, args);
		new App().schemaCollection();
		System.out.println("Done");
	}

	@Bean
	public CommonsXsdSchemaCollection schemaCollection() {
		CommonsXsdSchemaCollection schemaCollection = new CommonsXsdSchemaCollection(new ClassPathResource("xsd/folderA/a.xsd"));
		schemaCollection.setUriResolver(new DefaultURIResolver());
		schemaCollection.setInline(true);
		return schemaCollection;
	}
}`

When I run this application after packaging it as a jar, it prints Done without throwing an exception. I would've expected it to throw an exception?

wilkinsona:
> and updated the application to directly create the CommonsXsdSchemaCollection

Why did you make this update? To be a useful comparison the only change should be in the jar's packaging.

[...]

PR #11312, opened 2017-12-11
Precompile current dir pattern in Handler

dreis2211:
Hi,

while looking into some of the recent startup performance issues in 1.5.x, I noticed that #9917 introduced quite an allocation heavy normalization step (in my case I'd say roughly ~4GB heap pressure only for char[] in 30 seconds) in the startup-phase.

Though I've not taken a too deep look into it (e.g. if it's possible to avoid the additional char[] allocations in some cases), I found one possible quick-win optimization that gets rid of the overhead of compiling the pattern for the current directory, which also accounts for ~600MB (170MB for the pattern itself + 470MB of int[] array).

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/Handler.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/Handler.java
@@ -29,6 +29,7 @@ import java.util.Map;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.logging.Level;
 import java.util.logging.Logger;
+import java.util.regex.Pattern;
 
 /**
  * {@link URLStreamHandler} for Spring Boot loader {@link JarFile}s.
@@ -49,6 +50,8 @@ public class Handler extends URLStreamHandler {
 	private static final String SEPARATOR = "!/";
 
 	private static final String CURRENT_DIR = "/./";
+
+	private static final Pattern CURRENT_DIR_PATTERN = Pattern.compile(CURRENT_DIR);
 
 	private static final String PARENT_DIR = "/../";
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks very much, @dreis2211. With what version of Boot did you gather those numbers? I'm wondering if the benefits you've seen are with out without the [recent improvements](https://github.com/spring-projects/spring-boot/commit/a491727b7c9610bf1ea86d30ce3b7ffefe5f3a60) in this area.

I'd also be interested in learning how you measured the allocations/heap pressure. We generally use YourKit for this sort of thing, and I'm sometimes left feeling that there's information that should be available but isn't, or that isn't as readily available as it could be.

dreis2211:
@wilkinsona They are indeed without the recent improvements as I was testing specifically with 1.5.9.RELEASE and to my shame overlooked these. Thanks for pointing them out! I hope the PR still gets accepted to enhance the cases were we can't early out. 

I generally use JProfiler as a first step to fire-up my JAR for those profilings and take a look in the `Live-Memory->Allocation Hot Spots`. If I can't draw an early picture from that, I additionally use JMC locally with startup flight recordings and take a look into the `Memory->Allocations in New TLAB->Allocations by Class` section. Though JMC is not strictly needed, it generally gives me a better picture of recording something over time. 
 
`java -XX:+UnlockCommercialFeatures -XX:+FlightRecorder -XX:StartFlightRecording=duration=30s,name=startup,settings=profile,filename=startup-boot.jfr -jar path\to\my.jar`

wilkinsona:
Thanks again. I see no reason why we can't merge this into 1.5.x. As you say, it's still a useful enhancement for when the replacement logic cannot be short-circuited.

dreis2211:
Thanks @wilkinsona - I appreciate it.

wilkinsona:
Closed by https://github.com/spring-projects/spring-boot/commit/d5370e6852cb61e333369127a517474de2a7ec58 (in which I typo'd the issue number).

wilkinsona:
Thanks again, @dreis2211. This change is now in 1.5.x and master.

PR #11313, opened 2017-12-11
Add @Configuration to ScheduledTasksEndpointAutoConfiguration

izeye:
I guess it's working without it but this PR adds it for consistency with the other auto-configuration classes.


snicoll:
@izeye good catch. There are other cases in nested classes that I intended to review, thanks for the reminder :)

PR #11314, opened 2017-12-11
Avoid allocations from lower-casing spec in Handler

dreis2211:
Hi,

while looking into the recent startup performance issues in 1.5.x I found a possible way to prevent allocations of `char[]` due to lower-casing the spec in `Handler`. In my case they are contributing to the heap pressure with several hundred MBs (roughly ~600MB) in 30 seconds.

Instead of lowercasing I do what `String.equalsIgnoreCase` internally does and call `String.regionMatches` with its case-insensitive variant.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/Handler.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/Handler.java
@@ -159,7 +159,7 @@ public class Handler extends URLStreamHandler {
 
 	@Override
 	protected void parseURL(URL context, String spec, int start, int limit) {
-		if (spec.toLowerCase().startsWith(JAR_PROTOCOL)) {
+		if (spec.regionMatches(true, 0, JAR_PROTOCOL, 0, JAR_PROTOCOL.length())) {
 			setFile(context, getFileFromSpec(spec.substring(start, limit)));
 		}
 		else {


dreis2211:
I'm unable to make the connection between my changes and the failing tests. They seem unrelated - could you confirm?

bclozel:
@dreis2211 I don't think this failure are related to your changes; this is a recurrent failure we see in our tests and we're currently investigating. Could you share a bit more about your use case and benchmark numbers so that we can see what kind of performance improvement we can expect for a typical Spring Boot app?

Thanks!

dreis2211:
@bclozel I can't share my application unfortunately. Judging from experience I would say it's medium to big, though....We have roughly 1500 classes, but nothing too fancy in terms of dependencies. Web, Web-Security, JPA etc. - the usual suspects. I think the importance of my change scales with the size of the application and the dependencies (and JARs) it has to process.

I can share JMH benchmark results for now that test the specific change of `startsWith` vs. `regionMatches` in an isolated fashion.

**1. Case: There is something to do (e.g. actually changing the case):**
```
Benchmark.testNew                                   thrpt   10   95863160,596 ± 4351789,502   ops/s
Benchmark.testNew:·gc.alloc.rate                    thrpt   10         ? 10??                MB/sec
Benchmark.testNew:·gc.alloc.rate.norm               thrpt   10         ? 10??                  B/op
Benchmark.testNew:·gc.count                         thrpt   10            ? 0                counts
Benchmark.testOld                                   thrpt   10   14003288,801 ±  468735,566   ops/s
Benchmark.testOld:·gc.alloc.rate                    thrpt   10       1637,205 ±      54,850  MB/sec
Benchmark.testOld:·gc.alloc.rate.norm               thrpt   10        184,000 ±       0,001    B/op
Benchmark.testOld:·gc.churn.PS_Eden_Space           thrpt   10       1623,952 ±     188,439  MB/sec
Benchmark.testOld:·gc.churn.PS_Eden_Space.norm      thrpt   10        182,442 ±      18,477    B/op
Benchmark.testOld:·gc.churn.PS_Survivor_Space       thrpt   10          0,142 ±       0,072  MB/sec
Benchmark.testOld:·gc.churn.PS_Survivor_Space.norm  thrpt   10          0,016 ±       0,008    B/op
Benchmark.testOld:·gc.count                         thrpt   10        110,000                counts
Benchmark.testOld:·gc.time                          thrpt   10         69,000                    ms
```

**2. Case: There is nothing to do (e.g. everything is already lower-case):**
```
Benchmark.testNew                         thrpt   10   98241712,472 ± 2764742,209   ops/s
Benchmark.testNew:·gc.alloc.rate          thrpt   10         ? 10??                MB/sec
Benchmark.testNew:·gc.alloc.rate.norm     thrpt   10         ? 10??                  B/op
Benchmark.testNew:·gc.count               thrpt   10            ? 0                counts
Benchmark.testOld                         thrpt   10   22829153,232 ±  567287,823   ops/s
Benchmark.testOld:·gc.alloc.rate          thrpt   10         ? 10??                MB/sec
Benchmark.testOld:·gc.alloc.rate.norm     thrpt   10         ? 10??                  B/op
Benchmark.testOld:·gc.count               thrpt   10            ? 0                counts
```

As you can see, both scenarios reasonably improve the old version and scenario 1 is actually allocation-free with my change.

I hope this justifies the change enough in order to approve the PR. I don't have too much free capacities at the moment to give more proof.

Cheers,
Christoph

bclozel:
I was just curious about the type of improvement we could expect from this. We'll run the startup benchmarks again anyway.

Note that I didn't merge your PR directly as I also wanted to apply that on the 1.5.x branch - [we tend to merge forward in this repository](https://github.com/spring-projects/spring-boot/wiki/Working-with-Git-branches).

Thanks for your contribution @dreis2211, keep them coming !

Issue #11241, opened 2017-12-01
 AuditEvent: Principal must not be null when Authentication.getName() returns null


lukas-krecan-lt:
If Spring security authentication fails (most likely due to null principal) I am getting this exception which in turn results in 500 HTTP error

```
Caused by: java.lang.IllegalArgumentException: Principal must not be null
	at org.springframework.util.Assert.notNull(Assert.java:134)
	at org.springframework.boot.actuate.audit.AuditEvent.<init>(AuditEvent.java:88)
	at org.springframework.boot.actuate.audit.AuditEvent.<init>(AuditEvent.java:64)
	at org.springframework.boot.actuate.security.AuthorizationAuditListener.onAuthorizationFailureEvent(AuthorizationAuditListener.java:67)
	at org.springframework.boot.actuate.security.AuthorizationAuditListener.onApplicationEvent(AuthorizationAuditListener.java:47)
	at org.springframework.boot.actuate.security.AuthorizationAuditListener.onApplicationEvent(AuthorizationAuditListener.java:33)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:172)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:165)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:139)
	at org.springframework.context.support.AbstractApplicationContext.publishEvent(AbstractApplicationContext.java:393)
	at org.springframework.context.support.AbstractApplicationContext.publishEvent(AbstractApplicationContext.java:347)
	at org.springframework.security.access.intercept.AbstractSecurityInterceptor.publishEvent(AbstractSecurityInterceptor.java:504)
	at org.springframework.security.access.intercept.AbstractSecurityInterceptor.beforeInvocation(AbstractSecurityInterceptor.java:236)
	at org.springframework.security.access.intercept.aopalliance.MethodSecurityInterceptor.invoke(MethodSecurityInterceptor.java:65)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
	at org.springframework.aop.framework.CglibAopProxy$DynamicAdvisedInterceptor.intercept(CglibAopProxy.java:673)
```

wilkinsona:
> most likely due to null principal

[...]

PR #11320, opened 2017-12-11
AuditEvent: Principal must not be null when Authentication.getName() returns null

nklmish:
See #11241 for more details

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/audit/AuditEvent.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/audit/AuditEvent.java
@@ -26,6 +26,7 @@ import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
 
+import org.apache.commons.lang3.StringUtils;
 import org.springframework.context.ApplicationEventPublisher;
 import org.springframework.context.ApplicationEventPublisherAware;
 import org.springframework.util.Assert;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/audit/AuditEvent.java
```diff
@@ -25,6 +25,7 @@
 import com.fasterxml.jackson.annotation.JsonFormat;
 import com.fasterxml.jackson.annotation.JsonInclude;
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
+import org.apache.commons.lang3.StringUtils;
```
snicoll: That import  doesn't look right.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/audit/AuditEvent.java
```diff
@@ -85,10 +87,9 @@ public AuditEvent(String principal, String type, String... data) {
 	public AuditEvent(Date timestamp, String principal, String type,
 			Map<String, Object> data) {
 		Assert.notNull(timestamp, "Timestamp must not be null");
-		Assert.notNull(principal, "Principal must not be null");
 		Assert.notNull(type, "Type must not be null");
 		this.timestamp = timestamp;
-		this.principal = principal;
+		this.principal = principal == null ? StringUtils.EMPTY : principal;
```
snicoll: That check should be reversed: we assert that it's not null, the `else` part dealing with the case when it is.

snicoll:
Thanks for the quick feedback!

PR #11330, opened 2017-12-12
Use String.equalsIgnoreCase() where possible

dreis2211:
Hey,

this PR replaces two occurences of the pattern `someString.equals(otherString.toLowerCase())` with `someString.equalsIgnoreCase(otherString)`.

This avoid unnecessary allocations coming from lowercasing.

Cheers,
Christoph


philwebb:
Thanks!

Issue #7467, opened 2016-11-23
Align AbstractFilterRegistrationBean dispatcher types with servlet spec


philwebb:
The default dispatcher types defined in `AbstractFilterRegistrationBean` were originally copied from `AbstractDispatcherServletInitializer.getDispatcherTypes()` however they [continue to cause problems](https://github.com/spring-projects/spring-boot/issues/5737).

Rather than trying to pick our own set we should align with [the servlet spec](http://docs.oracle.com/javaee/6/api/javax/servlet/FilterRegistration.html#addMappingForServletNames(java.util.EnumSet,%20boolean,%20java.lang.String...)) and use `DispatcherType.REQUEST` as the default.

We will then need to check our own configuration of filters to see if specified dispatcher types are required for them.

cemo:
It would be cool to be set default values from configuration files.

wilkinsona:
This change will mean that `FORWARD`, `INCLUDE`, and `ASYNC` are removed from the default dispatcher types. The loss of `FORWARD` and `INCLUDE` appears to be inconsequential. The loss of `ASYNC` does have an effect: it means that the security context is now null within an async dispatch. I think we should change the default there.

cemo:
@wilkinsona see https://github.com/spring-projects/spring-boot/issues/4505#issuecomment-157072267 please. I still believe that `ERROR`, `ASYNC` and `REQUEST` are suitable for default. These 3 dispatcher type are usually are the beginning of request chains. However `FORWARD` and `INCLUDE` are not used at the beginning of request, instead they are used after.

wilkinsona:
Sorry, @cemo, but we're not going to change our decision to use `REQUEST` as the general purpose default as we want to align with the Servlet spec. Just to be clear, my comment above was specifically with regards to the default for the security filter.

cemo:
Thanks @wilkinsona, really makes sense to align with spec.

wilkinsona:
See #8288 and #8289 for exceptions to the new default of `REQUEST`.

PR #11348, opened 2017-12-14
Fix WebMvcMetrics to meter async endpoints

GreyTeardrop:
At the moment, async WebMvc requests are not metered. This PR attempts to fix that:

* After gh-7467 got fixed, `WebMvcMetricsFilter` wasn't running on `ASYNC` dispatch; it is explicitly enabled now.
* As `WebMvcMetricsFilter` is executed twice for async requests, `WebMvcMetrics` should
  not overwrite `TIMING_REQUEST_ATTRIBUTE` on second execution.
* To simplify testing and make measurements consistent with the ones
  coming from Micrometer code, nano time is taken from MeterRegistry's clock
  instead of directly calling `System.nanoTime()`.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/servlet/WebMvcMetricsConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/servlet/WebMvcMetricsConfiguration.java
@@ -16,6 +16,8 @@
 
 package org.springframework.boot.actuate.autoconfigure.metrics.web.servlet;
 
+import javax.servlet.DispatcherType;
+
 import io.micrometer.core.instrument.MeterRegistry;
 
 import org.springframework.boot.actuate.autoconfigure.metrics.MetricsProperties;
@@ -27,6 +29,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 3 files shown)


GreyTeardrop:
I've included a fix for a couple of tests that started failing after rebasing on top of latest master. They are technically not related to this PR, but needed to make CI green for it.

GreyTeardrop:
I've rebased on top of the latest `master` and verified that build runs fine locally.

@snicoll could you please take a look? Would be great to make this change into RC1 if possible, as I'm not sure of a good workaround for the issue in application code.

snicoll:
ping @jkschneider - I'd like to upgrade to micrometer `1.0.0-SNAPSHOT` first.

GreyTeardrop:
Rebased on top of the latest `master` and accommodated to Micrometer 1.0.0.rc6 update.

That update included the most important fix from this PR (`WebMvcMetrics` no longer overwrites `TIMING_REQUEST_ATTRIBUTE` on second execution), and rest of changes could be worked around in application code by declaring `FilterRegistrationBean<WebMvcMetricsFilter>` manually, but it still would be great to have WebMVC metrics for async methods work out of the box.

GreyTeardrop:
@snicoll Is there anything else that could be done for this PR to be considered into Boot RC1 milestone? It should now be up-to-date with Micrometer 1.0.0-rc.7. I don't see any significant changes in [1.0.0-SNAPSHOT over 1.0.0-rc.7](https://github.com/micrometer-metrics/micrometer/compare/v1.0.0-rc.7...master) so it should work with 1.0.0-SNAPSHOT as well.

snicoll:
Nothing, thank you.

jkschneider:
Testing this on the 1.5.x backport currently. Seems like there are subtle differences in the way WebMVC treated `Callable` vs. `CompletableFuture` return types with respect to the servlet filter. In 1.5.x, the filter code was called twice for a `Callable` but only once for a `CompletableFuture`.

jkschneider:
Don't merge as is. I have added a bit more to this in the legacy module that we will port over. Specifically:

1. A fix and test to ensure that we don't double count active tasks on long task timings of async dispatches.
2. Support for multiple long task timings of a single mapping.
3. Consolidation of `WebMvcMetricsFilter` and `WebMvcMetrics` for efficiency (less map lookups per operation).
4. Elimination of `recordAsPercentiles` from the props for request instrumentation (since there is a generalization of this feature in `MeterFilter` now).

snicoll:
> Don't merge as is.

What does that mean concretely? Are you going to submit a PR of your own? Perhaps we could guide @GreyTeardrop so that they can update their pull request?

GreyTeardrop:
@jkschneider Could you please clarify what you'd like to be done in the scope of this PR?

jkschneider:
@GreyTeardrop I've dropped your commit on top of #11776.

PR #11364, opened 2017-12-16
Fix package name for InfluxDB

izeye:
This PR fixes package name for InfluxDB.


philwebb:
Oh my :D

snicoll:
How the heck did I miss that? 😭

PR #11365, opened 2017-12-16
Fix property key to enable influxdb health endpoint

eddumelendez:


snicoll:
Oh my, good catch @eddumelendez!

PR #11369, opened 2017-12-17
Fixed the command for a full build in the contribution guide.

thombergs:
The old command doesn't work anymore, since folders were moved / renamed.
Replaced the old command with the current command from the CI script.


philwebb:
Thanks for the PR. Those instructions are indeed wrong, but I have a feeling that we can now get rid of the step that builds `settings.xml` entirely. I've raised #11374 to look into it.

PR #11373, opened 2017-12-18
Replace contains() with indexOf()

izeye:
This PR replaces `contains()` with `indexOf()` to save a potential subsequent `indexOf()` invocation.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/audit/AuditEvent.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/audit/AuditEvent.java
@@ -95,8 +95,8 @@ public class AuditEvent implements Serializable {
 	private static Map<String, Object> convert(String[] data) {
 		Map<String, Object> result = new HashMap<>();
 		for (String entry : data) {
-			if (entry.contains("=")) {
-				int index = entry.indexOf("=");
+			int index = entry.indexOf("=");
+			if (index != -1) {
 				result.put(entry.substring(0, index), entry.substring(index + 1));
 			}
 			else {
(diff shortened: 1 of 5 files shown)


dreis2211:
@izeye If I may, I'd suggest to use the **char** variants of `indexOf` while being on it. What do you think?

izeye:
@dreis2211 Thanks for the suggestion!

Is it for cosmetic improvement or for performance improvement?

In performance perspective, based on the following StackOverflow answer (although I didn't double-check the answer is correct) it looks worse: https://stackoverflow.com/questions/33646781/java-performance-string-indexofchar-vs-string-indexofsingle-string

izeye:
I did the similar benchmark with this: https://github.com/izeye/sample-jmh-maven/blob/master/src/main/java/com/izeye/samples/IndexOfBenchmark.java

but got the opposite result: https://github.com/izeye/sample-jmh-maven/blob/master/src/main/java/com/izeye/samples/IndexOfBenchmark.md

I'm not sure the StackOverflow answer is correct.

@dreis2211 Any idea?

dreis2211:
I see the same results as you - hence the suggestion. I always try to use single-character variants of String related methods as they "generally" yield better performance. So I'm a bit surprised about the StackOverflow results. I assume something must have changed between 8u45 (version on SO thread) and 8u151 (my version and presumably more or less yours).

But there is indeed a counter-intuitive change in JDK 9 that will switch our results again. See
- http://mail.openjdk.java.net/pipermail/jdk9-dev/2017-January/005539.html
- https://bugs.openjdk.java.net/browse/JDK-8173585

Aleksey Shipilev states the following in the mail:
> Think about it this way: you have the improvement for both cases in JDK 9 due to
Compact Strings, but it somewhat twisted the performance model to be less
intuitive. Continue using indexOf(char) in the code, and wait for JDK to catch up.

To cut a long story short, I'd say: your call. ;)

Cheers

izeye:
@dreis2211 Thanks for the details!

Based on your comment, it looks arguable. I don't have a strong opinion on it, so if the Spring Boot team agrees with it, I will apply the suggested change.

snicoll:
@izeye I don't have a strong opinion on it either and I can see that it's used inconsistently in the code right now anyway so I've created #11416 to discuss the possibility to harmonize that.

PR #11376, opened 2017-12-18
Apply customize() to createCredentialNetworkMongoClient()

izeye:
This PR applies `customize()` to `createCredentialNetworkMongoClient()` and adds a test for it.


snicoll:
Good catch @izeye, thank you again!

Issue #8011, opened 2017-01-17
SpringProfileDocumentMatcher does not work as expected


michael-simons:
The following issue appears with Spring Boot 1.4.3:

Given a YAML configuration like

```
someValue: 4711

example:
    greeting: Hello, World!
---
spring:
    profiles: demo1
example:
    greeting: Hello, Demo1!

---
spring:
    profiles: demo2
example:
    greeting: Hello, Demo2!

---
spring:
    profiles: "!demo2"
someValue: 2109
```

(Full configuration and properties class in [example project extconfig](https://github.com/springbootbuch/extconfig) see: [application.yml](https://github.com/springbootbuch/extconfig/blob/19350c9bad0f4c6aeecb51724f44e5613d3238f4/src/main/resources/application.yml) and [ExampleConfiguration](https://github.com/springbootbuch/extconfig/blob/19350c9bad0f4c6aeecb51724f44e5613d3238f4/src/main/java/de/springbootbuch/extconfig/ExampleConfiguration.java))

and run with `java -jar target/extconfig.jar` I expect `someValue` to be 2109: The negated profile !demo2 overwriting the value of 4711.

run with `java -jar target/extconfig.jar --spring.profiles.active=demo2` I expect `someValue` to be 4711 as it is set in the default, not overwritten in demo2 (not set) and not overwritten bei !demo2. The same is expected for the other parts of the configuration.

In contrast to my expectations, I get in both cases 2109.

The project linked above contains a test class [NegatedProfilesTest](https://github.com/springbootbuch/extconfig/blob/b9d47497af37de63d1b213fbd3894b307cde3a52/src/test/java/de/springbootbuch/extconfig/NegatedProfilesTest.java) that fails.

wilkinsona:
@michael-simons and I discussed this a bit already. The problem appears to be that `SpringProfileDocumentMatcjer` isn't correctly configured with the active profiles. Instead, it only considers a profile to be active if it's processing a configuration file that's specific to that profile.

mbenson:
I've attacked this a couple of ways and it seems a pretty tough nut to crack within the confines of the existing `PropertySourcesLoader`/`PropertySourceLoader` API. I don't have anything helpful yet, but anything that can get folks thinking about the problem. The `SpringProfileDocumentMatcher` works pretty well in a vacuum. :|

mbenson:
[...]

PR #4953, opened 2016-01-15
support profile negation in YAML subdocuments

mbenson:

PR #11378, opened 2017-12-19
Remove mention of broken YAML subdocument profile negation

shakuzen:
Documenting this feature misleads users into thinking it should work. The documentation can be added back once gh-8011 is fixed.

### DIFF
--- a/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
@@ -784,11 +784,6 @@ profile, and it would have to be explicitly reset in all other profiles as neces
 	    password: weak
 ----
 
-Spring profiles designated using the "spring.profiles" element may optionally be negated
-using the `!` character. If both negated and non-negated profiles are specified for
-a single document, at least one non-negated profile must match and no negated profiles
-may match.
-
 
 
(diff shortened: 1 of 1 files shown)


snicoll:
@shakuzen I think I lost you as why that section of the doc is linked to #8011

shakuzen:
The removed section of the doc is the only description of negated profiles. Negated profiles are what is reported as not working in #8011 (with repro project and test linked in the issue description). I'm not sure if negated profiles ever worked. The unit tests for `SpringProfileDocumentMatcher` pass but I haven't been able to actually get negated profiles to work as described in any project or integration test.

wilkinsona:
Support for profile negation was [introduced in 1.4](https://github.com/spring-projects/spring-boot/pull/4953). That would suggest one of the following:

1. It's really broken and no one's using it.
2. It largely works other than the problem described in 8011.

I think that 2 is the more likely, particularly as it was a contribution and, presumably, it works for @mbenson's original use case at least. Rather than just removing the documentation I'd prefer to either fix 8011 or to remove the functionality altogether.

mbenson:
I'm still ready to participate in a solution for #8011. Recall that I had some code working there last year that hopefully would at least be a starting point.

wilkinsona:
We've agreed that just removing the documentation isn't the right approach. We'll try to get to #8011 before 2.0 GAs.

PR #11387, opened 2017-12-20
Fix epub generation for reference document

gonwan:
Cannot open epub reference document in iOS/iBooks and some other epub readers, since 2.0.0 M5. Works fine with 2.0.0 M4. Can be caused by this commit: https://github.com/spring-projects/spring-boot/commit/4df84c53bca75ff9eb81fadded325628dbcd4fc7

Downgrade docbkx-maven-plugin to 2.0.16/2.0.15 fix the problem.

### DIFF
--- a/spring-boot-project/spring-boot-docs/pom.xml
+++ b/spring-boot-project/spring-boot-docs/pom.xml
@@ -1154,7 +1154,7 @@
 					<plugin>
 						<groupId>com.agilejava.docbkx</groupId>
 						<artifactId>docbkx-maven-plugin</artifactId>
-						<version>2.0.17</version>
+						<version>2.0.16</version>
 						<configuration>
 							<sourceDirectory>${basedir}/target/generated-docs</sourceDirectory>
 							<includes>index.xml</includes>


pivotal-issuemaster:
@gonwan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=11387)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=11387) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@gonwan Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=11387)!

wilkinsona:
Thanks for the PR, @gonwan. Validating the built .epub file points to a problem with container.xml. Specifically, I think we're being affected by [this issue](https://github.com/mimil/docbkx-tools/issues/130). Assuming that it works, I'd prefer to implement the workaround described in the issue rather than downgrading the plugin. I am trying the workaround now.

wilkinsona:
@gonwan Thanks for bringing the problem to our attention. The workaround (plus another correction) seems to fix it so we'll take that approach rather than downgrading the plugin.

PR #11396, opened 2017-12-22
Remove org.flywaydb.core.internal.util.StringUtils import

izeye:
This PR removes `org.flywaydb.core.internal.util.StringUtils` `import` as it looks accidentally imported.

It happened before and was removed in 4235df180eac7be7c69e9cd1e93f1aa627074be3, so this PR also adds a Checkstyle rule to prevent it from sneaking into codebase again.


snicoll:
Note to self: merge the checkstyle change in `1.5.x`

PR #11418, opened 2017-12-23
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/CloudFoundryReactiveHealthEndpointWebExtension.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/CloudFoundryReactiveHealthEndpointWebExtension.java
@@ -22,13 +22,12 @@ import org.springframework.boot.actuate.autoconfigure.cloudfoundry.CloudFoundryE
 import org.springframework.boot.actuate.endpoint.annotation.EndpointExtension;
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
 import org.springframework.boot.actuate.endpoint.web.WebEndpointResponse;
-import org.springframework.boot.actuate.endpoint.web.annotation.EndpointWebExtension;
 import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.actuate.health.HealthEndpoint;
 import org.springframework.boot.actuate.health.ReactiveHealthEndpointWebExtension;
 
 /**
- * Reactive {@link EndpointWebExtension} for the {@link HealthEndpoint} that always
+ * Reactive {@link EndpointExtension} for the {@link HealthEndpoint} that always
  * exposes full health details.
  *
  * @author Madhura Bhave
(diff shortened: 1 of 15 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/CloudFoundryReactiveHealthEndpointWebExtension.java
```diff
@@ -22,13 +22,12 @@
...
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
 import org.springframework.boot.actuate.endpoint.web.WebEndpointResponse;
-import org.springframework.boot.actuate.endpoint.web.annotation.EndpointWebExtension;
 import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.actuate.health.HealthEndpoint;
 import org.springframework.boot.actuate.health.ReactiveHealthEndpointWebExtension;
 
 /**
- * Reactive {@link EndpointWebExtension} for the {@link HealthEndpoint} that always
+ * Reactive {@link EndpointExtension} for the {@link HealthEndpoint} that always
```
snicoll: Actually I think the javadoc is right but the annotation on the class is wrong. Can you please polish that instead?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/servlet/CloudFoundryHealthEndpointWebExtension.java
```diff
@@ -20,13 +20,12 @@
...
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
 import org.springframework.boot.actuate.endpoint.web.WebEndpointResponse;
-import org.springframework.boot.actuate.endpoint.web.annotation.EndpointWebExtension;
 import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.actuate.health.HealthEndpoint;
 import org.springframework.boot.actuate.health.HealthEndpointWebExtension;
 
 /**
- * {@link EndpointWebExtension} for the {@link HealthEndpoint} that always exposes full
+ * {@link EndpointExtension} for the {@link HealthEndpoint} that always exposes full
```
snicoll: Ditto

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/reactive/CloudFoundryReactiveHealthEndpointWebExtension.java
izeye: @snicoll AFAICT, it can't be `@EndpointWebExtension` as it doesn't expose the `filter` element. WDYT?

snicoll: I am confused, it returns `Mono<WebEndpointResponse<Health>>` which definitely makes it a web specific concern. I haven't properly looked at the new naming since the recent refactoring but that doesn't seem right to me.

izeye: @snicoll How about changing `@EndpointWebExtension` to expose `filter` element and make it have `WebEndpointFilter` as a default value? With the change, `CloudFoundryReactiveHealthEndpointWebExtension` and `CloudFoundryHealthEndpointWebExtension` can be annotated with `@EndpointWebExtension` instead of `@EndpointExtension`.

snicoll: The filter is really what makes the extension an extension atm so exposing it is a bit weird. You could have a `EndpointWebExtension` with the jmx filter :)

I've created #11428 to discuss this further. With the current state of affair, you change is definitely legit so I'll merge that.

Issue #11271, opened 2017-12-06
Configure Checkstyle to tighten up import restrictions


wilkinsona:
We should restrict the following:

- Commons Lang
- Commons Lang 3
- Guava

dreis2211:
@wilkinsona Out of interest (and with the risk of asking a stupid question): What's the reasoning behind those restrictions?

wilkinsona:
They are transitive dependencies of various libraries that Boot supports, but we don’t want to accidentally use them in Boot’s own code.

dreis2211:
I see - thanks for the heads-up. Good to know for contributors.

Issue #11267, opened 2017-12-06
Test support uses commons lang3


snicoll:
`AssertProviderApplicationContextInvocationHandler` has a commons lang3 import.

wilkinsona:
Let’s update Checkstyle as well to stop it happening again

wilkinsona:
Actually, let's use a separate issue and police the imports in 1.5.x too.

PR #11432, opened 2017-12-26
Add a Checkstyle rule to prevent Commons Lang/Lang 3 imports

izeye:
This looks missed in https://github.com/spring-projects/spring-boot/issues/11271 which was created from https://github.com/spring-projects/spring-boot/issues/11267.


snicoll:
Good catch _again_ @izeye! I noticed it too when I merged the Liquibase package in `1.5.x` and forgot to revisit after 😅

PR #11439, opened 2017-12-27
Move CorsEndpointProperties to the upper package

izeye:
`CorsEndpointProperties` lives in `org.springframework.boot.actuate.autoconfigure.endpoint.web.servlet` but is also used in `org.springframework.boot.actuate.autoconfigure.endpoint.web.reactive`, so this PR moves it to its common parent package.

This PR also extracts `CorsConfiguration` creation logic duplicated in `WebMvcEndpointManagementContextConfiguration` and `WebFluxEndpointManagementContextConfiguration` into `CorsEndpointProperties`.


snicoll:
@izeye thanks! I've polished the PR to use the new `PropertyMapper`.

PR #9355, opened 2017-05-30
Auto-detect jOOQ dialect

michael-simons:
This PR improves the auto configuration for jOOQ. It uses the same idea as auto configuration for `JPA` for determining the database platform.

It introduces a new class `SQLDialectLookup` that determines the dialect for databases supported by the open source edition of jOOQ.

This dialect or if not found `DEFAULT` as before is than used to configure the JOOQ Configuration.

I didn't have to change much in the test: The dialect has already been checked and there is also a test that sets the configuration property, so that case was already covered.

The PR also clarifies the documentation in regards of the dialects enum constants, which are not the same as used in JPA. Another idea would be "translating" `spring.datasource.platform` to jOOQs specific enums, but I'm not sure if that isn't too much heuristics.

This change here is required to make #9343 successfully build.

PR #11466, opened 2018-01-02
Detect jOOQ SQLDialect through jOOQ's JDBCUtils

lukaseder:
This change amends the previous PR #9355 by @michael-simons, passing the JDBC URL from the `DataSource` directly to jOOQ's `JDBCUtils` rather than creating an additional mapping between `org.springframework.boot.jdbc.DatabaseDriver` and `org.jooq.SQLDialect`.

This has the advantages that:

1. jOOQ's `SQLDialect` to URL mappings are already maintained by jOOQ, so no additional changes will be necessary to Spring Boot, in the future
2. The suggested PR also works for the commercial jOOQ distributions, e.g. when working with DB2, Oracle, SQL Server, etc., as the `JDBCUtils` of the commercial distribution also contains the relevant logic to map to e.g. `SQLDialect.DB2`, `SQLDialect.ORACLE`, `SQLDialect.SQLSERVER` (which are not available from the open source distribution linked by Spring Boot by default).


michael-simons:
Nice change, i like it.

PR #11489, opened 2018-01-03
Full editing pass plus linking

Buzzardo:
I made a complete editing pass plus another pass to see where I could add more links, both within the document and to other parts of Spring. The result was a thorough edit (though I'm sure I missed things, purely due to the size of the thing).


bclozel:
Merged with 0e7dbd59b2e54b419ddc9464eabd337a8217cada

Thanks @Buzzardo !

Issue #1946, opened 2014-11-18
Allow Gson to be configured declaratively


wilkinsona:
`GsonBuilder` allows a custom `Gson` instance to be created. `GsonAutoConfiguration` could be updated to use `GsonBuilder` and configure it from `spring.gson.*` properties in much the same way as `JacksonAutoConfiguration` does with `spring.jackson.*` properties

geoff-wasilwa:
Hi @wilkinsona is this still being worked on?

wilkinsona:
Not to my knowledge. It's just a possible enhancement at the moment. Pull requests welcome.

ioanngolovko:
Hi @wilkinsona. I want to implement this as well as jackson.  I have an only one question.

In which branch to make a pull request?

wilkinsona:
@ioanngolovko Against the master branch, please.

izeye:
This looks resolved via https://github.com/spring-projects/spring-boot/pull/11498.

snicoll:
Indeed @izeye, thanks!

Duplicate of #11498

PR #11498, opened 2018-01-04
Allow Gson to be configured declaratively (and programmatically via c…

ioanngolovko:
…ustomizers)

GsonAutoconfiguration was implemented in accordance with the formulation (#1946). 
Customizers are also supported.


philwebb:
@ioanngolovko Thanks very much for this contribution. I've merged it into master and it should be part of RC1. I also took the liberty of changing the `StandardGsonBuilderCustomizer` to use the new `PropertyMapper` class that we recently introduced. See commit 286a3bb10aa6884c4c93245d1c928ec7d3db69fb for the differences.

Thanks again!

Issue #11257, opened 2017-12-04
Consider additional kafka properties


garyrussell:
There is at least one property `....consumer.idle-event-interval` (and likely more, e.g. `....consumer.no-poll-interval`) that would be useful to expose as Boot configuration properties.

Add these, and examine the current state of auto config against any new properties that may have been added since we last revised boot auto config.

__feel free to assign to me (I don't have permissions to assign)__

snicoll:
@garyrussell if you intended a submit a  PR for this, there is no need for an issue. The PR is enough. Are you taking care of this?

garyrussell:
Eventually, yes, before the RC.

I don't get the "we don't need an issue if there's going to be a PR" approach.

How do you guys track your TODO lists? Without a github assigned issue, I'd have to track this TODO, someplace else.

Wouldn't having an open issue with a target milestone give you a picture of what's needed to be completed before a milestone can be built?

garyrussell:
Also, with an open issue, perhaps a community member might want to contribute a commit.

snicoll:
I am not denying any of that. I was reacting to "feel free to assign to me". That lead me to think you will take care of it in which case there is no need for an issue. 

If you want a community member to pick this up, we can flag it ideal for contribution but we should expand the description a bit more first.

garyrussell:
Yeah - it needs some research so it's probably easier if I do it. However, it's not high on my priority list (but I do want to get it in before the RC); hence this placeholder so I don't forget.

snicoll:
@garyrussell do you think you'll be able to work on a PR in the coming weeks for this?

garyrussell:
@snicoll Yes; my plan is to do so before Boot's RC1 - maybe even this week.

snicoll:
Closing in favour of PR #11502

PR #11502, opened 2018-01-04
Add support for additional Kafka listener properties

garyrussell:
Resolves: https://github.com/spring-projects/spring-boot/issues/11257

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


snicoll:
@garyrussell thanks for the PR, I've polished it a bit (see https://github.com/spring-projects/spring-boot/commit/c4db22007c2051284f1f40eacb4ab0285dc2ee1e). In particular:

* We use now `Duration` for every time-based property. If the default unit is not ms we have `DefaultDurationUnit` to precise what the default is
* We've polished the configuration key descriptions recently and I've adapted those keys to it. It would be great if you had a quick look to the polish commit or the appendix to review that

Also, we are now running against Kafka `2.1.1.BUILD-SNAPSHOT`. Could you please set a release date for it so that we can confirm we're good for Spring Boot `2.0.0.RC1` (see #11524).

Thanks!

garyrussell:
Thanks; done; I scheduled it for next Friday.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
```diff
@@ -101,6 +101,11 @@ private void configureContainer(ContainerProperties container) {
 				.to(container::setAckTime);
 		map.from(properties::getPollTimeout).whenNonNull().as(Duration::toMillis)
 				.to(container::setPollTimeout);
+		map.from(properties::getClientId).whenNonNull().to(container::setClientId);
```
idolice: Hi, I could not find the method setClientId for my ContainerProperties, I think it caused some issues

snicoll: If you do, you are overriding the managed version for Spring Kafka in your application. Please remove the override or upgrade to the snapshot for the time being. Spring Kafka `2.1.1.RELEASE` will have this property.

Issue #11517, opened 2018-01-05
KafkaAdmin client as part of KafkaAutoConfiguration


Jcamilorada:
[AdminClient](https://kafka.apache.org/10/javadoc/index.html?org/apache/kafka/clients/admin/AdminClient.html) provides useful admin operations over kafka. Consider adding as part of kafka auto-configuration beans. 

- One can argue that this functionality need to be wrapped by [KafkaAdmin](https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaAdmin.java)

- created as part discussion [11435](https://github.com/spring-projects/spring-boot/issues/11435) pull request discussion

Jcamilorada:
created as part discussion [11435](https://github.com/spring-projects/spring-boot/issues/11435) pull request discussion

snicoll:
paging @garyrussell to get more feedback on this.

garyrussell:
A `KafkaAdmin` `@Bean` is [already autoconfigured](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java#L134).

garyrussell:
If an application needs access to an `AdminClient` directly, e.g. 
to use features not exposed by the KafkaAdmin, it can use `AdminClient.create(kafkaAdmin.getConfig())`.

But I don't have a strong objection to adding the bean if it's felt necessary.

snicoll:
Unless this is really necessary, I'd rather avoid exposing it. @Jcamilorada perhaps you could rework your original PR with that suggestion?

garyrussell:
I looked at the health indicator PR.

One benefit of it being a `@Bean` is that it implements `AutoCloseable`, so Spring will close the connection, when the context is `stop()` ped, if an application forgets to do so.

With my suggestion above, the app is responsible for doing his own close.

On the other hand, I wonder if we really should keep open a long-lived admin connection for health checks; perhaps it would be better to use a new `AdminClient` each time (and close the resources).

Jcamilorada:
@garyrussell @snicoll  thanks, I will update pull request to use new `AdminClient` each time, probably test will require `KafkaEmbedded` as well.

snicoll:
Alright, thanks both!

massivespace:
[...]

PR #11518, opened 2018-01-05
Add auto-configuration support for KafkaAdmin client

Jcamilorada:
solves [11517](https://github.com/spring-projects/spring-boot/issues/11517)
- added Admin client bean


snicoll:
As discussed in #11517, we've decided not to do this.

@Jcamilorada fyi, there is no need to open an issue and a PR. If you intend to open a PR, that is enough.

Issue #8115, opened 2017-01-26
Expose the functional bean registration API via SpringApplication


wilkinsona:
`SpringApplication` makes it hard to get hold of the application context before it's refreshed. @michael-simons came up with this:

```java
SpringApplication springApplication = new SpringApplication(Application.class);
springApplication.addInitializers((GenericApplicationContext ctx) -> {
    ctx.registerBean(Greeter.class, Greeter::new);
});
springApplication.run(args);
```

It would be nice if the initialiser wasn't needed and the API provided an easier way to access the context before it's refreshed.

michael-simons:
Background was (again) creating an example for my book how to use Spring 5 enhancements out of the box with Spring Boot. I assume that this question will arise, especially with the functional router.

dsyer:
> way to access the context before it's refreshed

Isn't that the definition of an `ApplicationContextInitializer`? How could the API be any simpler without duplicating?

michael-simons:
I - as a user - would have been looking for something like

```
springApplication
    .registerBean(Greeter.class, Greeter::new);
    .registerBean(Foo.class, () -> new WhateverFoo());
```

and having Boot delegate to an Initializer for me.

dsyer:
As a user I'd expect to have something that worked in integration tests with `@SpringBootTest` as well as when I run my main method (changing the API of `SpringApplication` doesn't help there).

wilkinsona:
> How could the API be any simpler without duplicating?

I'm not sure, however the pure Spring example in the [Kotlin blog post](https://spring.io/blog/2017/01/04/introducing-kotlin-support-in-spring-framework-5-0) involves less ceremony:

```
GenericApplicationContext context = new GenericApplicationContext();
context.registerBean(Foo.class);
context.registerBean(Bar.class, () -> new 
	Bar(context.getBean(Foo.class))
);
```

It's a shame that `SpringApplication` gives you an extra hoop to jump through. Perhaps we'll have to live with that, but I think it's worth some thought.

michael-simons:
Doesn't seem to take integration tests with `@SpringBootTest` into account as well… It's a good point!

sdeleuze:
[...]

PR #11527, opened 2018-01-06
Adding Kotlin SpringApplicationBuilder Extensions

juan-medina:
This is an small PR to add some Kotlin Extensions Functions for SpringApplicationBuilder.

The overall idea is that allow to use as:

```Kotlin
buildApplication<MySpringApplication>().run()

buildApplication<MySpringApplication> {
            beans {
                bean {
                    SomeBean()
                }
            }
        }.run()

SpringApplicationBuilder().sources<MySpringApplication>().run()
```

### DIFF
--- a/spring-boot-project/spring-boot/src/main/kotlin/org/springframework/boot/SpringApplicationBuilderExtensions.kt
+++ b/spring-boot-project/spring-boot/src/main/kotlin/org/springframework/boot/SpringApplicationBuilderExtensions.kt
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot
+
+import org.springframework.boot.builder.SpringApplicationBuilder
+import org.springframework.context.ApplicationContextInitializer
+
+/**
+ * Top level function acting as a Kotlin shortcut allowing to write
+ * `buildApplication<FooApplication>()` instead of
+ * `SpringApplicationBuilder(FooApplication::class.java)`.
+ *
+ * @author Juan Medina
+ * @since 2.0.0
+ * @see SpringApplicationBuilder
+ */
+inline fun <reified T : Any> buildApplication(): SpringApplicationBuilder =
+        SpringApplicationBuilder(T::class.java)
+
+/**
+ * Top level function acting as a Kotlin shortcut allowing to write
+ * `buildApplication<FooApplication>() { // ApplicationContextInitializer ... }`
+ * instead of creating the builder and invoking `initializers()`.
+ *
+ * @author Juan Medina
+ * @since 2.0.0
+ * @see SpringApplicationBuilder.initializers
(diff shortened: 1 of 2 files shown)


philwebb:
@sdeleuze Any chance you could review this? Does it align with the extension that's been added for `SpringApplication`?

sdeleuze:
@juan-medina Thanks for this PR, but I am not sure about adding such extension since `runApplication()` already [provides a way to customize the application via function literals with receiver](https://github.com/spring-projects/spring-boot/blob/e28b98c59d43190b5ca8ea3ae4deac50c9737766/spring-boot-project/spring-boot/src/main/kotlin/org/springframework/boot/SpringApplicationExtensions.kt#L43), see for example [this test](https://github.com/spring-projects/spring-boot/blob/d2ee817d8cc7e914830d9a0551f9f9f2ac21bbe9/spring-boot-project/spring-boot/src/test/kotlin/org/springframework/boot/SpringApplicationExtensionsTests.kt#L44-L51) that use it. I tend to think this is more idiomatic in Kotlin than using a builder. Any thoughts?

juan-medina:
Hi this is not for receiver this is mainly for `sources` and `initializers` from `SpringApplicationBuilder`, 

I was reading Josh Long Kotlin Example of Dynamic registration on gist https://gist.github.com/joshlong/0d1082c0b4c84cad65982fbe89db71ea

and I thought could be more simple with just this extension:

```
@SpringBootApplication
class FeedApplication

fun main(args: Array<String>) {

  fun processSyndEntry(syndEntry: SyndEntry) {
    // todo
  }

  buildApplication<FeedApplication >{
  	beans {
  		val feeds = mapOf("atom" to listOf("https://spring.io/blog.atom"),
  				"rss" to listOf("https://cloudfoundry.org/feed/"))
  		feeds.entries.forEach { e ->
  			e.value.forEach { url ->
  				val urlAsKey = url.filter { !it.isLetterOrDigit() }
  				bean(name = "feedFlorFor${urlAsKey}") {
  					IntegrationFlows
  							.from(Feed.inboundAdapter(UrlResource(url), urlAsKey), { it.poller({ it.fixedDelay(1000) }) })
  							.handle(GenericHandler<SyndEntry> { syndEntry, headers ->
  								processSyndEntry(syndEntry)
  							})
  							.get()
  				}
  			}
  		}
  	}
    .run(*args)
}
```

sdeleuze:
Isn't that already possible to write:

```kotlin
@SpringBootApplication
class FeedApplication

fun main(args: Array<String>) {
	runApplication<FeedApplication>(*args) {
		addInitializers(
			beans {
				val feeds = mapOf("atom" to listOf("https://spring.io/blog.atom"),
						"rss" to listOf("https://cloudfoundry.org/feed/"))
				feeds.entries.forEach { e ->
					e.value.forEach { url ->
						val urlAsKey = url.filter { !it.isLetterOrDigit() }
						bean(name = "feedFlorFor$urlAsKey") {
							IntegrationFlows
								.from(Feed.inboundAdapter(UrlResource(url), urlAsKey), { it.poller({ it.fixedDelay(1000) }) })
								.handle(GenericHandler<SyndEntry> { syndEntry, headers ->
									processSyndEntry(syndEntry)
								})
								.get()
						}
					}
				}
			}
		)
	}
}
``` 
?

juan-medina:
Absolutely, but does not fill more fluent the other? 

I think bean able to use the beans DSL is wonderful for Kotlin I was trying to make it more straightforward to anyone.

Of course if is make sense

sdeleuze:
I totally understand the intent, but the very small difference between both (no need for `addInitializers { }`) is not worth adding this variant IMO, especially given the fact that `runApplication<FeedApplication>(*args) { }` is for Kotlin developers the idiomatic way to do what Java developers do with `SpringApplicationBuilder`.

Another reason for not integrating this and discussed in #8115 is that this syntax currently only apply to running the application but not tests. I really care about Boot + Functional but we need to take the time to provide a comprehensive first class support, and that will not happen for 2.0, so I think we are fine with current `runApplication` capabilities.

PR #11534, opened 2018-01-07
Use instance equality for Class

izeye:
This PR changes to use instance equality for `Class` to be consistent with the change via https://github.com/spring-projects/spring-boot/commit/2efa21c57074cfc5527266ebaa790aa3b3f208d3.


snicoll:
@izeye I am not sure that this change was a call to harmonization across the codebase (though I guess that shouldn't hurt but, again, we should write that down somewhere then). Flagging for team attention.

philwebb:
The change in 2efa21c57074cfc5527266ebaa790aa3b3f208d3 was a specific suggestion from JitWatch. It can't hurt to align the others, although they're probably not on such hot code-paths so it's certainly less critical.

PR #11591, opened 2018-01-10
Make GsonAutoConfiguration align with JacksonAutoConfiguration

izeye:
This PR makes `GsonAutoConfiguration` align with `JacksonAutoConfiguration` by:

- Add `@ConditionalOnMissingBean` on `GsonBuilder` `@Bean` method similar to `Jackson2ObjectMapperBuilder`.
- Add `@Primary` to `Gson` `@Bean` method similar to `ObjectMapper`.


snicoll:
@izeye I am not sure about `@Primary`. We basically had to do this because other components down the chain will expose an `ObjectMapper`. I'd rather not add `@Primary` in auto-configuration unless I am forced to. Can you please rework this part please?

izeye:
@snicoll Thanks for the explanation! I changed it in 7cb9a451fdc34b417fbb372ac17afaecca2851fa

Issue #11613, opened 2018-01-12
spring-boot-sample-quartz does not have any tests


wilkinsona:
We should at least have one that verifies that the context loads.

snicoll:
Closing in favour of PR #1621

PR #11621, opened 2018-01-13
Add test to spring-boot-sample-quartz

eddumelendez:
See gh-11613

### DIFF
--- a/spring-boot-samples/spring-boot-sample-quartz/pom.xml
+++ b/spring-boot-samples/spring-boot-sample-quartz/pom.xml
@@ -19,6 +19,11 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-quartz</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-test</artifactId>
+			<scope>test</scope>
+		</dependency>
 	</dependencies>
 	<build>
 		<plugins>
--- a/spring-boot-samples/spring-boot-sample-quartz/src/test/java/sample/quartz/SampleQuartzApplicationTests.java
+++ b/spring-boot-samples/spring-boot-sample-quartz/src/test/java/sample/quartz/SampleQuartzApplicationTests.java
@@ -1,0 +1,43 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
+ *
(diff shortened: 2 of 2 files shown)


spring-boot-samples/spring-boot-sample-quartz/pom.xml
```diff
@@ -19,6 +19,10 @@
 			<groupId>org.springframework.boot</groupId>
 			<artifactId>spring-boot-starter-quartz</artifactId>
 		</dependency>
+		<dependency>
+			<groupId>org.springframework.boot</groupId>
+			<artifactId>spring-boot-starter-test</artifactId>
```
snicoll: This is missing a `test`scope.

spring-boot-samples/spring-boot-sample-quartz/src/test/java/sample/quartz/SampleQuartzApplicationTests.java
```diff
@@ -0,0 +1,43 @@
...
+ * @author Eddú Meléndez
+ */
+public class SampleQuartzApplicationTests {
+
+	@Rule
+	public OutputCapture outputCapture = new OutputCapture();
+
+	@Test
+	public void test() {
+		SampleQuartzApplication.main(new String[0]);
```
snicoll: Can you please get the context back there and close it at the end? It is probably useless now but it's better to make sure we actually trigger a shutdown.

spring-boot-samples/spring-boot-sample-quartz/pom.xml
eddumelendez: 😱

spring-boot-samples/spring-boot-sample-quartz/src/test/java/sample/quartz/SampleQuartzApplicationTests.java
eddumelendez: should I use `ApplicationContextRunner`?

snicoll: I  don't think so. Running the app is ok, we have other samples that do that.

spring-boot-samples/spring-boot-sample-quartz/pom.xml
eddumelendez: fixed

Issue #11608, opened 2018-01-11
NoClassDefFoundError when running with Liquibase 2.x


tony-pelton:
i'm trying to get a spring 4.x application onto spring-boot. a problem i've run into, is that you can't have Liquibase 2.x libraries on the classpath or spring-boot intialization fails.

`java.lang.NoClassDefFoundError: liquibase/servicelocator/CustomResolverServiceLocator

	at org.springframework.boot.liquibase.LiquibaseServiceLocatorApplicationListener.onApplicationEvent(LiquibaseServiceLocatorApplicationListener.java:44)
	at org.springframework.boot.liquibase.LiquibaseServiceLocatorApplicationListener.onApplicationEvent(LiquibaseServiceLocatorApplicationListener.java:35)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:167)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:139)
	at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:122)
	at org.springframework.boot.context.event.EventPublishingRunListener.starting(EventPublishingRunListener.java:68)
	at org.springframework.boot.SpringApplicationRunListeners.starting(SpringApplicationRunListeners.java:48)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:303)`

the application listener does this :

`		if (ClassUtils.isPresent("liquibase.servicelocator.ServiceLocator", null)) {
			new LiquibasePresent().replaceServiceLocator();
		}
`
the problem is, the class it is looking for exists in 2.x Liquibase, but the classes the "replacer" looks for don't.

and there is no switch to stop this from happening, since it is an application listener.

this is a problem for me, as i'd like to upgrade to 3.x, but i'm having trouble, and wanted to be able to do some testing with 2.x vs 3.x ... but this listener prevents me from having any control at all over whether i can have Liquibase 2.x on the classpath.

wilkinsona:
Thanks for the analysis. We can tighten things up by checking for `liquibase.servicelocator.CustomResolverServiceLocator`. In the meantime, you can work around the problem with the following, slightly clunk approach:

```java
SpringApplication application = new SpringApplication(SampleLiquibaseApplication.class);
application.setListeners(application.getListeners().stream()
[...]

PR #11625, opened 2018-01-13
Add check for missing Liquibase's class `CustomResolverServiceLocator`

eddumelendez:
`CustomResolverServiceLocator` was introduced in Liquibase 2.0.4, this
check prevents failures during initialization when previous versions are
used.

See gh-11608

### DIFF
--- a/spring-boot/src/main/java/org/springframework/boot/liquibase/LiquibaseServiceLocatorApplicationListener.java
+++ b/spring-boot/src/main/java/org/springframework/boot/liquibase/LiquibaseServiceLocatorApplicationListener.java
@@ -31,6 +31,7 @@ import org.springframework.util.ClassUtils;
  *
  * @author Phillip Webb
  * @author Dave Syer
+ * @author Eddú Meléndez
  */
 public class LiquibaseServiceLocatorApplicationListener
 		implements ApplicationListener<ApplicationStartingEvent> {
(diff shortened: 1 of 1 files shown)


spring-boot/src/main/java/org/springframework/boot/liquibase/LiquibaseServiceLocatorApplicationListener.java
```diff
@@ -51,12 +52,14 @@ public void onApplicationEvent(ApplicationStartingEvent event) {
 	private static class LiquibasePresent {
 
 		public void replaceServiceLocator() {
-			CustomResolverServiceLocator customResolverServiceLocator = new CustomResolverServiceLocator(
-					new SpringPackageScanClassResolver(logger));
-			customResolverServiceLocator.addPackageToScan(
-					CommonsLoggingLiquibaseLogger.class.getPackage().getName());
-			ServiceLocator.setInstance(customResolverServiceLocator);
-			liquibase.logging.LogFactory.reset();
+			if (ClassUtils.isPresent("liquibase.servicelocator.CustomResolverServiceLocator", null)) {
```
snicoll: Can we please have a test that exercises this behaviour. You can use the `FilteredClassLoader` for that.

wilkinsona: I think the new check is in the wrong place. I’d expect it to replace the existing check above.

wilkinsona: Also, given that we use Liquibase 3.5.x in 1.5, it would be worth looking to see if there’s a class that’s new in 3.5 that we can check for. We should review the auto-configuration conditions too.

snicoll:
@eddumelendez I've applied those suggestions to  08ec3d2.  Thanks for the PR

eddumelendez:
sorry guys, I couldn't apply changes on time :( Thanks @snicoll for the hint to test this scenario

Issue #11607, opened 2018-01-11
Ordering of `WebMvcMetricsFilter` breaks character encoding


vpavic:
Actuator's `WebMvcMetricsFilter` has a hard-coded order of `Ordered.HIGHEST_PRECEDENCE`, which in practice puts it ahead of `CharacterEncodingFilter`. Since `WebMvcMetricsFilter` performs the mapping introspection, under some circumstances it will touch the POST data hence breaking the `CharacterEncodingFilter`.

I've managed to put together a [sample app](https://github.com/vpavic/boot-char-encoding) that exhibits this behavior with a specific set of request mapping paths. After starting the app, issue the following HTTP POST request (samples use HTTPie):

```shell
$ http -f POST :8080/pv/post name='Vedran Pavić' Content-Type:'application/x-www-form-urlencoded'
HTTP/1.1 200 
Content-Length: 21
Content-Type: text/plain;charset=UTF-8
Date: Thu, 11 Jan 2018 17:58:44 GMT

Hello Vedran PaviÄ
```

When `spring-boot-starter-actuator` is removed from the dependencies, which gets `WebMvcMetricsFilter` out of the picture, request is processed as expected:

```shell
$ http -f POST :8080/pv/post name='Vedran Pavić' Content-Type:'application/x-www-form-urlencoded'
HTTP/1.1 200 
Content-Length: 19
Content-Type: text/plain;charset=UTF-8
Date: Thu, 11 Jan 2018 17:59:32 GMT

Hello Vedran Pavić
```

vpavic:
On top of addressing the ordering problem that leads to this issue, would it makes sense to introduce some sort of validation that checks whether `CharacterEncodingFilter` is the first filter in the chain, and log a warning if it isn't?

Boot could have a test that verifies this, since it has happened a few time that other filters got in front of `CharacterEncodingFilter`, and also the users could be warned if they incidentally add a filter that gets ordered in front.

snicoll:
`ErrorPageFilter` is also flagged `Ordered.HIGHEST_PRECEDENCE`

sriki77:
Should the `CharacterEncodingFilter` be made to implement `PriorityOrdered` with highest precedence?

sriki77:
Making `OrderedCharacterEncodingFilter` implement `PriorityOrdered` makes it work. I will try and prepare a pull request, if this is acceptable solution.

snicoll:
[...]

PR #11626, opened 2018-01-13
CharacterEncodingFilter has higher precedence than WebMvcMetricsFilter

sriki77:
Made OrderedCharacterEncodingFilter to have higher precedence than WebMvcMetricsFilter

Made the OrderedCharacterEncodingFilter implement PriorityOrdered
Changed the WebMvcMetricsFilter to implement Ordered and provided
an option to change the order.

Fixes gh-11607


sriki77:
The build needs to retriggered, seems like an infra issue.

snicoll:
Thanks for the PR but as discussed in #11607 we took another approach.

PR #11643, opened 2018-01-16
Make it explicit that the private keys are only for testing

dsyer:
Calling them "fakeKey" is enough to tell the cred-alert-cli that
it can ignore the private keys we use for testing.


wilkinsona:
Historically, we've rejected this sort of environment-specific change as it's all to easy for someone to regress things as there's nothing enforcing the local variable naming convention. I'd rather leave things as they are. Apparently, anyone using the hooks can [whitelist a repository](https://github.com/pivotal-cf/git-hooks-core#whitelists). That's just me, though. Let's see what the rest of the team thinks.

philwebb:
I'm confused. Commit 1459d9a7963bc99abb9f732dcfcf2a291c79885c is just changing the name of some local variables. Was there another commit or some other change relating to this PR?

dsyer:
No, that’s the only change. If you used the cred alert cli it would barf on master and not in this PR.

wilkinsona:
Our preference is that the tool's whitelist support is used. Thanks anyway.

PR #11649, opened 2018-01-17
 Upgrade to JUnit Jupiter 5.0.3

eddumelendez:


snicoll:
Thanks @eddumelendez - FYI, we have a semi-automated process that takes care of upgrading dependencies in due course so we don't need PR for this.

eddumelendez:
I've got it :)

PR #11665, opened 2018-01-18
Use System.lineSeparator()

izeye:
This PR changes to use `System.lineSeparator()`.

### DIFF
--- a/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/ServiceCapabilitiesReportGenerator.java
+++ b/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/ServiceCapabilitiesReportGenerator.java
@@ -36,8 +36,6 @@ import java.util.TreeSet;
  */
 class ServiceCapabilitiesReportGenerator {
 
-	private static final String NEW_LINE = System.getProperty("line.separator");
-
 	private final InitializrService initializrService;
 
 	/**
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/ServiceCapabilitiesReportGenerator.java
```diff
@@ -36,8 +36,6 @@
  */
 class ServiceCapabilitiesReportGenerator {
 
-	private static final String NEW_LINE = System.getProperty("line.separator");
```
snicoll: Can we please keep the `NEW_LINE` assignement there so that the PR change is smaller?

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/logging/log4j2/ExtendedWhitespaceThrowablePatternConverterTests.java
```diff
@@ -32,8 +32,6 @@
  */
 public class ExtendedWhitespaceThrowablePatternConverterTests {
 
-	private static final String LINE_SEPARATOR = System.getProperty("line.separator");
```
snicoll: Can  be harmonized to `NEW_LINE`.

spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/ServiceCapabilitiesReportGenerator.java
wilkinsona: FWIW, I like the change as @izeye has proposed. I see no need for the constant now that there's a method we can call instead.

izeye: @snicoll Thanks for the quick feedback! I intentionally removed the constants as I considered `System.lineSeparator()` effectively a constant. What do you think?

snicoll: Thanks for the feedback. Yes, that makes sense.

PR #11674, opened 2018-01-18
Use the platform-specific line separator

izeye:
This PR changes to use the platform-specific line separator.


snicoll:
@philwebb I was looking at this one and you've polished the  original code back to `\n` while I used what @izeye just polished. I don't think that's the first time I see that.

I don't think there is anything wrong using `String.format` with only `%n`. If you disagree, it would be nice that we take a decision once for all rather than this polish un-polish scenario.

snicoll:
@philwebb sorry I didn't wait but I see no reason to as using `\n` is not right IMO. It's not really about what you've described but more about formatting  the data. And when I need that (like carriage  return support, I use `String.format`. Hope that makes sense.

philwebb:
@snicoll Yeah totally, I messed up this one. My comment about `String.format` was more about why I replace them now and again in polish commits. I also do the same with `StringBuffer.append(...).append(...).append(...)`. I know there can be a minor performance hit, but it's often negligible.

PR #11679, opened 2018-01-18
Make all metrics exporters conditional on missing bean.

ITman1:
I have a use case where I need to customize implementation of `JmxMeterRegistry` and export it by my own, but this is currently not possible because Spring Boot tries to autoconfigure `JmxMeterRegistry` exporter every time. There should be simple fix just to annotate metrics exporter of `JmxMeterRegistry` with `@ConditionalOnMissingBean`. I needed it only for JMX, but I have added it to all metrics exporters.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/atlas/AtlasExportConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/atlas/AtlasExportConfiguration.java
@@ -40,13 +40,16 @@ import org.springframework.context.annotation.Configuration;
 @EnableConfigurationProperties(AtlasProperties.class)
 public class AtlasExportConfiguration {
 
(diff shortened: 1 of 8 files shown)


pivotal-issuemaster:
@ITman1 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=11679)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=11679) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ITman1 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=11679)!

philwebb:
I'm not too sure about this one. We've generally avoided special bean names in the past and tried to keep to type based matches. Can you use the `management.metrics.export.atlas.enabled` property to disable them?

BTW. Some of that meter registry code might yet get an overhaul before we go GA.

wilkinsona:
Thanks for the PR, but this has been superseded by the changes proposed in #11776. `MetricsExporter` is no more, and the equivalent bean definitions are more specifically typed.

PR #11683, opened 2018-01-19
RabbitMQ - RetryTemplate configuration hook

garyrussell:
Provide a mechanism so that users can further configure `RetryTemplate`s
used in `RabbitTemplate` and listeners.

An example would be to add a retry listener so the user can monitor/log
retry attempts.

https://stackoverflow.com/questions/48331502/logging-exceptions-thrown-by-message-listeners-for-spring-amqp/48332001#48332001

Or, to set a custom `RetryContextCache`.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractRabbitListenerContainerFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractRabbitListenerContainerFactoryConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import org.springframework.amqp.rabbit.retry.MessageRecoverer;
 import org.springframework.amqp.rabbit.retry.RejectAndDontRequeueRecoverer;
 import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.boot.autoconfigure.amqp.RabbitProperties.ListenerRetry;
+import org.springframework.retry.support.RetryTemplate;
 import org.springframework.util.Assert;
 
 /**
@@ -42,6 +43,8 @@ public abstract class AbstractRabbitListenerContainerFactoryConfigurer<T extends
 
 	private RabbitProperties rabbitProperties;
 
+	private RabbitListenerRetryTemplateConfigurer retryTemplateConfigurer;
+
 	/**
 	 * Set the {@link MessageConverter} to use or {@code null} if the out-of-the-box
 	 * converter should be used.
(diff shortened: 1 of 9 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/retry/RetryTemplateConfigurer.java
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * A configurer for a {@link RetryTemplate}, for example, to add a RetryListener.
+ *
+ * @param <P> the properties type.
+ *
+ * @author Gary Russell
+ * @since 2.0
+ *
+ */
+public interface RetryTemplateConfigurer<P> {
```
snicoll: We usually use the term `Customizer` for this.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitTemplateRetryTemplateConfigurer.java
```diff
@@ -0,0 +1,30 @@
...
+import org.springframework.boot.autoconfigure.retry.RetryTemplateConfigurer;
+
+/**
+ * A {@link RetryTemplateConfigurer} for RabbitMQ listeners.
+ *
+ * @author Gary Russell
+ * @since 2.0
+ *
+ */
+public class RabbitTemplateRetryTemplateConfigurer extends RabbitRetryTemplateConfigurer {
```
snicoll: This empty class seems suspicious.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/retry/RetryTemplateConfigurer.java
```diff
@@ -0,0 +1,39 @@
...
+
+	/**
+	 * Configure the template.
+	 * @param template the template.
+	 * @param properties the properties.
+	 */
+	void configure(RetryTemplate template, P properties);
```
snicoll: There is a mixed of concept here. Usually a customizer is something that  applies something extra. It doesn't  get the configuration properties object in argument (as it is an auto-configuration concern).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractRabbitListenerContainerFactoryConfigurer.java
```diff
@@ -107,9 +118,13 @@ protected void configure(T factory, ConnectionFactory connectionFactory,
...
-			builder.maxAttempts(retryConfig.getMaxAttempts());
-			builder.backOffOptions(retryConfig.getInitialInterval(),
-					retryConfig.getMultiplier(), retryConfig.getMaxInterval());
+			RabbitListenerRetryTemplateConfigurer retryConfigurer = this.retryTemplateConfigurer;
+			if (retryConfigurer == null) {
+				retryConfigurer = new RabbitListenerRetryTemplateConfigurer();
+			}
+			RetryTemplate template = new RetryTemplate();
+			retryConfigurer.configure(template, retryConfig);
+			builder.retryOperations(template);
```
snicoll: This code doesn't is not equivalent right? `RabbitListenerRetryTemplateConfigurer()`  does more than what you removed as far as I see.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitTemplateRetryTemplateConfigurer.java
garyrussell: It's an empty class because there are two `RetryTemplate`s - one for the consumers and one for the template. 

How else can I determine that the user has provided a customizer for one Vs. the other?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractRabbitListenerContainerFactoryConfigurer.java
garyrussell: It's exactly equivalent; just that the old code used a shortcut on the builder. In any case; I'll refactor to a customizer and this will be moot.

PR #11697, opened 2018-01-19
RabbitMQ - RetryTemplate customization hook

garyrussell:
Provide a mechanism so that users can customize `RetryTemplate`s
used in `RabbitTemplate` and listeners.

An example would be to add a retry listener so the user can monitor/log
retry attempts.

https://stackoverflow.com/questions/48331502/logging-exceptions-thrown-by-message-listeners-for-spring-amqp/48332001#48332001

Or, to set a custom `RetryContextCache`.

temp

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractRabbitListenerContainerFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractRabbitListenerContainerFactoryConfigurer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -24,6 +24,7 @@ import org.springframework.amqp.rabbit.retry.MessageRecoverer;
 import org.springframework.amqp.rabbit.retry.RejectAndDontRequeueRecoverer;
 import org.springframework.amqp.support.converter.MessageConverter;
 import org.springframework.boot.autoconfigure.amqp.RabbitProperties.ListenerRetry;
+import org.springframework.retry.support.RetryTemplate;
 import org.springframework.util.Assert;
 
 /**
@@ -42,6 +43,8 @@ public abstract class AbstractRabbitListenerContainerFactoryConfigurer<T extends
 
 	private RabbitProperties rabbitProperties;
 
+	private RabbitListenerRetryTemplateCustomizer retryTemplateCustomizer;
+
 	/**
 	 * Set the {@link MessageConverter} to use or {@code null} if the out-of-the-box
 	 * converter should be used.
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractRabbitListenerContainerFactoryConfigurer.java
```diff
@@ -59,6 +62,14 @@ protected void setMessageRecoverer(MessageRecoverer messageRecoverer) {
 		this.messageRecoverer = messageRecoverer;
 	}
 
+	/**
+	 * Set the {@link RabbitListenerRetryTemplateCustomizer} to use.
+	 * @param retryTemplateCustomizer the customizer
+	 */
+	public void setRetryTemplateCustomizer(RabbitListenerRetryTemplateCustomizer retryTemplateCustomizer) {
```
snicoll: I've started a separate discussion (#13530) but I wonder if a `Consumer<RetryTemplate>` wouldn't be enough here.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfigurationTests.java
```diff
@@ -332,11 +336,30 @@ public void testRabbitTemplateBackOff() {
 	@Test
 	public void testRabbitMessagingTemplateBackOff() {
 		this.contextRunner.withUserConfiguration(TestConfiguration4.class)
+				.withPropertyValues("spring.rabbitmq.listener.simple.retry.enabled=true",
```
snicoll: I am not sure I understand the purpose of that test. We're creating a messaging template to show that it's taken into account. Why is retry processing added there?

garyrussell: I am testing that the customizer injected a listener into the listener container's retry template; in order to test that, we need to enable retry.

garyrussell:
I concur with the general comment about using `Consumer<?>` - the problem here is there are two retry templates (for consumers and publishers).

Perhaps another alternative would be for the `RabbitTemplate` to add a getter for its retry operations; then would have `Consumer<RabbitTemplate>` and the user could then use `template.getRetryOperations().registerListener(...)`.

And for the listener, we could have `Consumer<RetryTemplate>` (it's not so easy to add a getter on the listener containers since the retry template is injected into an interceptor and added to the container's generic advice chain.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfigurationTests.java
snicoll: I got that. But why are you testing that as part of `testRabbitMessagingTemplateBackOff` ?

snicoll:
> Perhaps another alternative would be for the RabbitTemplate to add a getter for its retry operations

I thought of that but I think we should do that only if that makes sense generally speaking and not to ease that particular case. I don't know if exposing the `retryTemplate` is a good idea (probably not).

My proposal of using `Consumer` was declined so we're back to the two functional interfaces. Another idea would be to find out the cases that require to customize the `RetryTemplate`. Perhaps we can have some sort of explicit support?

Or perhaps having two interfaces for this is fine? It just doesn't seem right to me but I can't put my fingers on why exactly. Flagging for team attention to get more feedback.

snicoll:
We've discussed this at the call and we're now leaning towards a unique `RabbitRetryTemplateCustomizer` that takes the `RetryTemplate` and the "source" (or the target). We don't want to expose a generic customizer and we are not keen on having two interfaces for this either. I'll try to spare some cycle to hack on this idea.

@garyrussell if you want to give it a try, please let me know.

garyrussell:
ok; will do

garyrussell:
@snicoll Sorry; mis-read; it's probably best for you to hack at it, to avoid a bunch of round trips 😄

snicoll:
@garyrussell alright I hacked something and I've opened #13793 to track it. Thanks for the PR and the help on this issue.

Issue #11650, opened 2018-01-17
Investigate CI integration with sonarcloud.io


philwebb:
#11624 as shown there could be some value.

wilkinsona:
I'm not opposed to adding extra code quality checks to the build as long as they don't slow things down too much and can be run locally prior to pushing a change.

snicoll:
I am ok to integrate it as long as it is an integral part of our development process and we take the time to craft a profile that eliminates what we consider false positive. The direct side effect of that is that we need a quality gate with concrete actions when it doesn't pass.

igor-suhorukov:
@wilkinsona it is possible to create maven profile to run locally on local sonarqube instance,
 also possible run IntelliJ Idea SonarLint plugin with rules and settings from http://sonarcloud.io.  SonarLint can analyse only on VCS changed files. We need measure sonar plugin overhead on typical project build.

wilkinsona:
Thanks for the info. Running a local instance is what I’ve done in the past and I’ve found it cumbersome and awkward. I’d like a way to analyse code locally but using the rules defined in the server. At a minimum I think that analysis needs to happen in command line builds (as we do today with Checkstyle). Being able to run them in IntelliJ and Eclipse would be a nice bonus.

igor-suhorukov:
>  I’d like a way to analyse code locally but using the rules defined in the server

I know only local IDE approach with SonarLint with remote rule server configuration
![sonarling-screenshot](https://user-images.githubusercontent.com/10332206/35187911-8b44b2c6-fe3d-11e7-8431-320154d5ee7c.png)

philwebb:
Perhaps running PMD and FindBugs as part of the build would be better. That might potentially find the same issues without requiring the server. [This article](https://www.sparkred.com/blog/open-source-java-static-code-analyzers/) has some relevant background.

igor-suhorukov:
@philwebb FindBugs is unsupported and outdated, [HuntBugs](https://habrahabr.ru/company/pvs-studio/blog/344734/#comment_10569860) also dead. Only Sonar complex rules analyser is alive. It is possible to run [IntelliJ Idea analyser](https://www.jetbrains.com/help/idea/running-inspections-offline.html) from scripts, but difficult to enforce rules and generate buid reports from output.

wilkinsona:
[...]

PR #11705, opened 2018-01-20
fix SonarCube issues  "Local variables should not shadow class fields"(squid:HiddenFieldCheck)

igor-suhorukov:
To get rid confusing to know whether the field or the local variable is being used. This pull request fix [171 issues](https://sonarcloud.io/project/issues?id=org.springframework.boot%3Aspring-boot-build&resolved=false&rules=squid%3AHiddenFieldCheck) with "local variables overriding class field names"( [squid:HiddenFieldCheck](https://sonarcloud.io/organizations/igor-suhorukov-github/rules#rule_key=squid%3AHiddenFieldCheck))

This pull request is related issues described in ticket #11650

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/condition/ConditionsReportEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/condition/ConditionsReportEndpoint.java
@@ -211,9 +211,9 @@ public class ConditionsReportEndpoint {
 		private final String message;
 
 		public MessageAndCondition(ConditionAndOutcome conditionAndOutcome) {
-			Condition condition = conditionAndOutcome.getCondition();
+			Condition outcomeCondition = conditionAndOutcome.getCondition();
 			ConditionOutcome outcome = conditionAndOutcome.getOutcome();
-			this.condition = ClassUtils.getShortName(condition.getClass());
+			this.condition = ClassUtils.getShortName(outcomeCondition.getClass());
 			if (StringUtils.hasLength(outcome.getMessage())) {
 				this.message = outcome.getMessage();
 			}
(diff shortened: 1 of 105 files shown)


wilkinsona:
Thanks for the PR but I disagree with this one. We prefix all field access with `this` so, IMO, there’s no confusion.

igor-suhorukov:
@wilkinsona ok, it is my mistake maybe

philwebb:
I'm with Andy on this one I'm afraid. We have a checkstyle rule to enforce the `this.` prefix so I think there's limited value in enforcing different local variable names. In fact, I often prefer the simpler names you get when you aren't forced to choose something different.

philwebb:
> what do you think about Utility classes should not have public constructors

Generally agree and we [have a checkstyle rule](https://github.com/philwebb/spring-boot/blob/0ca3a8d7ade7b319e48d46ad51ae32f36f4f747d/spring-boot-parent/src/checkstyle/checkstyle.xml#L40) for this. The problem with [the Sonar results](https://sonarcloud.io/project/issues?id=org.springframework.boot%3Aspring-boot-build&resolved=false&rules=squid%3AS1118&types=CODE_SMELL) here is that is' picking things like `CacheMeterBinderProvidersConfiguration` that I don't consider utility classes. In fact, we've generally tried to avoid util classes altogether in Boot.

> Generic wildcard types should not be used in return parameters

This one is very subjective as well. I generally try not to return them for public APIs, but it can be difficult to always do this without causing a lot of problems. Take [this one](https://sonarcloud.io/component?id=org.springframework.boot%3Aspring-boot-actuator%3Asrc%2Fmain%2Fjava%2Forg%2Fspringframework%2Fboot%2Factuate%2Finfo%2FInfoPropertiesInfoContributor.java&line=75) for example. We could return a `PropertySource<Object>` I guess, but really what's the point? We're really saying you can return a property source of any generic here and we'll deal with it. I'm not sure always enforcing the rule makes for better code, although it may well be worth a review of some APIs.

PR #11709, opened 2018-01-21
fix generic type constraint issues and try to avoid issues covered by SonarQube squid:S1452 rule

igor-suhorukov:
fix generic type constraint issues and try to avoid ["Generic wildcard types should not be used in return parameters"](https://sonarcloud.io/project/issues?id=org.springframework.boot%3Aspring-boot-build&resolved=false&rules=squid%3AS1452&types=CODE_SMELL) SonarQube [squid:S1452 rule](https://sonarcloud.io/organizations/igor-suhorukov-github/rules#rule_key=squid%3AS1452)

related to #11650

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/ConditionMessage.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/ConditionMessage.java
@@ -420,7 +420,7 @@ public final class ConditionMessage {
 			}
 		};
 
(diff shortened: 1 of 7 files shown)


philwebb:
Thanks for the suggestions, but overall I'm not sure that these edits really improve the codebase.

The change in `ConditionMessage` that returns a `Collection` rather than `Collection<?>` seems wrong, and most of the others are covered implicitly. For example `DataSourceProperties` currently returns a `DataSourceBuilder<?>`, the change to `DataSourceBuilder<? extends DataSource>` feels noisy given that the builder is already typed as `DataSourceBuilder<T extends DataSource>`.

Thanks anyway.

PR #11716, opened 2018-01-22
Manage dependencies for kotlin-stdlib-jdk7/8

sdeleuze:
As explained in the "Split package compatibility" section of
https://blog.jetbrains.com/kotlin/2017/09/kotlin-1-2-beta-is-out/,
`kotlin-stdlib-jdk7` and `kotlin-stdlib-jdk8` are the recommended
dependencies to use with Kotlin 1.2 for Java 9+ compatibility.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -1940,6 +1940,16 @@
 			</dependency>
 			<dependency>
 				<groupId>org.jetbrains.kotlin</groupId>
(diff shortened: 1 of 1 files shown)


snicoll:
> which replace the old kotlin-stdlib-jre7 and kotlin-stdlib-jre8.

Should we remove those then?

sdeleuze:
No because `kotlin-stdlib-jre7` and `kotlin-stdlib-jre8` can still be used by some projects (these upgrading Kotlin version from 1.1 for example), so I think we should define their version as long as the Kotlin version we use publish such artifacts.

PR #11724, opened 2018-01-22
Polishing

dreis2211:
Hi,

this tiny PR polishes a `String.length() == 0` occurence with `String.isEmpty()`.

Cheers,
Christoph


philwebb:
Ta!

PR #11742, opened 2018-01-22
Polish code: Avoid concatenating non-literals in a StringBuffer constructor or append()

igor-suhorukov:
Avoid concatenating non-literals in a StringBuffer constructor or append() since intermediate buffers will need to be be created and destroyed by the JVM.

Rule [InefficientStringBuffering](http://pmd.sourceforge.net/pmd-5.1.1/rules/java/strings.html)
Since: PMD 3.4


philwebb:
Sorry but I very much dislike this one. I prefer the readability of "string" + thing + "string" and I think we should only use `.append()` if it's in a hot area of code and we can prove it's improving performance.

Issue #11722, opened 2018-01-22
Support anonymousReadOnly property for LDAP


philwebb:
See https://github.com/spring-projects/spring-boot/issues/11691#issuecomment-359415667

snicoll:
Closing in favour of PR #11744

PR #11744, opened 2018-01-23
Add support for anonymousReadOnly in LdapProperties

eddumelendez:
See gh-11722

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/LdapProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/LdapProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -60,6 +60,11 @@ public class LdapProperties {
 	 */
 	private Map<String, String> baseEnvironment = new HashMap<>();
 
+	/**
+	 * Whether read-only operations should use an anonymous environment.
+	 */
+	private boolean anonymousReadOnly;
+
 	public String[] getUrls() {
 		return this.urls;
 	}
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/LdapProperties.java
```diff
@@ -60,6 +60,11 @@
 	 */
 	private Map<String, String> baseEnvironment = new HashMap<>();
 
+	/**
+	 * Enable anonymous read only.
```
snicoll: This description could use some love, something like

> Whether read-only operations should use an anonymous environment.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/ldap/LdapAutoConfigurationTests.java
```diff
@@ -16,13 +16,12 @@
...
 package org.springframework.boot.autoconfigure.ldap;
 
-import org.junit.After;
 import org.junit.Test;
 
-import org.springframework.boot.autoconfigure.context.PropertyPlaceholderAutoConfiguration;
-import org.springframework.boot.test.util.TestPropertyValues;
-import org.springframework.context.annotation.AnnotationConfigApplicationContext;
+import org.springframework.boot.autoconfigure.AutoConfigurations;
+import org.springframework.boot.test.context.runner.ApplicationContextRunner;
```
snicoll: Can the refactoring of the test be done in a separate commit please? First that with the test unchanged and then an extra commit on top with the feature.

snicoll:
Thanks @eddumelendez - regarding the test refactoring, that's not what I had in mind by "separate commit". What I meant here is first a commit that upgrades the existing tests and then one commit  on  top that implements the feature (with the relevant tests). I've restructured your PR with that and I've also [polished it](https://github.com/spring-projects/spring-boot/commit/960989cfe60e447229af954c31846ecb3b1f0461)

eddumelendez:
@snicoll I'll take it into account next time. Thanks! :+1:

PR #11751, opened 2018-01-23
Create Liquibase/Flyway DataSources's if user or url are specified.

### LINKED ISSUE #11747: Liquibase Configuration prone to misuse.
Hey all.

Quick question regarding the current implementation of [LiquibaseAutoConfiguration](
https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java). Though it is documented, it seems a little haphazard that failing to specify `liquibase.url` will cause a fallback of using the primary Datasource bean. 

As a user I would prefer if provided with a `liquibase.user` and `liquibase.password`, that those would be used against the default `datasource.url`, instead of Liquibase running with the primary `datasource.username` and `datasource.password` which can cause all sorts of issues in role based schemas.

Is there any strong feelings about the current implementation, or any information on why it was implemented this way? A quick Github search for `liquibase.user` shows tonnes of users, using this without implementing `liquibase.url`, and those that do are simply using the default `datasource.url`.

Interested to hear your thoughts.

DominicGunn:
This PR enables a more flexible Liquibase/Flyway configuration by allowing for a combination of both of the providers, and the primary DataSource's configurations. 

Allowing developers the flexibility to specify only a user or a url and having Liquibase/Flyway adapt to
existing properties, instead of ignoring them and falling back to the default data source.

Fixes #11747

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
@@ -39,6 +39,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.data.jpa.EntityManagerFactoryDependsOnPostProcessor;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfiguration;
+import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties;
 import org.springframework.boot.autoconfigure.orm.jpa.HibernateJpaAutoConfiguration;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.ConfigurationPropertiesBinding;
@@ -65,6 +66,7 @@ import org.springframework.util.ObjectUtils;
  * @author Stephane Nicoll
  * @author Jacques-Etienne Beaudet
  * @author Eddú Meléndez
+ * @author Dominic Gunn
  * @since 1.1.0
  */
 @Configuration
@@ -95,6 +97,8 @@ public class FlywayAutoConfiguration {
 
 		private final FlywayProperties properties;
 
+		private final DataSourceProperties dataSourceProperties;
+
 		private final ResourceLoader resourceLoader;
 
 		private final DataSource dataSource;
@@ -106,11 +110,13 @@ public class FlywayAutoConfiguration {
 		private List<FlywayCallback> flywayCallbacks;
 
 		public FlywayConfiguration(FlywayProperties properties,
-				ResourceLoader resourceLoader, ObjectProvider<DataSource> dataSource,
+				DataSourceProperties dataSourceProperties, ResourceLoader resourceLoader,
+				ObjectProvider<DataSource> dataSource,
 				@FlywayDataSource ObjectProvider<DataSource> flywayDataSource,
 				ObjectProvider<FlywayMigrationStrategy> migrationStrategy,
 				ObjectProvider<List<FlywayCallback>> flywayCallbacks) {
 			this.properties = properties;
+			this.dataSourceProperties = dataSourceProperties;
 			this.resourceLoader = resourceLoader;
 			this.dataSource = dataSource.getIfUnique();
 			this.flywayDataSource = flywayDataSource.getIfAvailable();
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
```diff
@@ -106,11 +110,13 @@ public FlywaySchemaManagementProvider flywayDefaultDdlModeProvider(
 		private List<FlywayCallback> flywayCallbacks;
 
 		public FlywayConfiguration(FlywayProperties properties,
```
johnou: Shouldn't we be retaining the original constructor for binary compatibility?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
```diff
@@ -83,16 +85,20 @@ public LiquibaseSchemaManagementProvider liquibaseDefaultDdlModeProvider(
...
 
+		private final DataSourceProperties dataSourceProperties;
+
 		private final ResourceLoader resourceLoader;
 
 		private final DataSource dataSource;
 
 		private final DataSource liquibaseDataSource;
 
 		public LiquibaseConfiguration(LiquibaseProperties properties,
```
johnou: And here

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
```diff
@@ -147,9 +164,8 @@ private void checkLocationExists(String... locations) {
 				Assert.state(locations.length != 0,
 						"Migration script locations not configured");
 				boolean exists = hasAtLeastOneLocation(locations);
-				Assert.state(exists,
-						() -> "Cannot find migrations location in: " + Arrays.asList(
-								locations)
+				Assert.state(exists, () -> "Cannot find migrations location in: "
```
johnou: Revert?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayProperties.java
```diff
@@ -133,7 +133,7 @@ public void setInitSqls(List<String> initSqls) {
 	}
 
 	public boolean isCreateDataSource() {
-		return this.url != null && this.user != null;
+		return this.url != null || this.user != null;
```
johnou: Support for datasource with just user?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
philwebb: This will be a 2.0 enhancement so this is fine.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
philwebb: Not sure what you mean by that?

johnou: Unnecessary formatting.

philwebb: That's fine. We deal with formatting issues all the time when we merge things. No need to resubmit the PR just for that.

DominicGunn: I'm pretty sure this actually changed after I ran the [eclipse formatter](https://github.com/spring-projects/spring-boot/blob/master/eclipse/eclipse-code-formatter.xml) over it.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayProperties.java
DominicGunn: Sure, we want to be able to support being able to set a flyway user without having to explicitly set a flyway url. We should fallback to the default datasource url when provided with just a flyway user.

PR #11759, opened 2018-01-24
Remove usage of Vector collection

dreis2211:
Hi,

this PR removes the usage of `Vector` in two tests and therefore from the codebase. As they simply create an empty Vector in order to have an Enumeration available, we can replace it with `Collections.emptyEnumeration()`.

Cheers,
Christoph


philwebb:
Thanks!

PR #11760, opened 2018-01-24
use EnumMap instead of other Map in case of enumeration key 

igor-suhorukov:
This pull request fix [issues ](https://sonarcloud.io/project/issues?id=org.springframework.boot%3Aspring-boot-build&resolved=false&rules=squid%3AS1640) related to sonar inspection [squid:S1640](https://sonarcloud.io/organizations/igor-suhorukov-github/rules#rule_key=squid%3AS1640): change map type to EnumMap in case of enumeration key


philwebb:
Thanks. Most of these look sensible changes but I'm a little worried about `MetricsEndpoint` since it's replacing a `LinkedHashMap` and insert order might be important (an `EnumMap` iterates in natural order)

igor-suhorukov:
@philwebb thank you for double check and your help!

Issue #11752, opened 2018-01-24
Use of Embedded LDAP with Spring Boot does not support multiple base DN


jatinvaidya:
**Environment:** 
Spring Boot 2.0.0.M7 with "com.unboundid:unboundid-ldapsdk:3.2.1"

**Background:**
A single LDAP directory server instance can have multiple "suffixes" (base) and usually does. For example, a single person can have TWO user entries in LDAP directory with each entry under a different suffix (base).

For example:

1. DN of user's entry in one suffix:
cn=john01,ou=users,**o=company,dc=com** 

2. DN of the same user's entry in another suffix:
uid=445566,ou=users,**dc=security**

_Note: The bases or suffixes are marked bold above._

**Problem:**
At this time, there is no way to load multiple suffixes (base) with the embedded ldap plugin with spring boot. This is even though the plugin itself (unboundid) **does support multiple suffixes** (base). 

[unboundid documentation](https://docs.ldap.com/ldap-sdk/docs/in-memory-directory-server.html)
> --baseDN {baseDN} -- Specifies the base DN to use for data in the server. This argument must be provided at least once, **but may be provided multiple times if more than one base DN should be used**. 

**Defect:**
The property **spring.ldap.embedded.base-dn** currently supports only one base though. Due to this it is not possible to work with the scenario described above.

eddumelendez:
Currently working on this and I need help 😅 

If the following property is set `spring.ldap.embedded.base-dns:dc=spring,dc=org` then there are 2 elements in the array first `dc=spring` and second `dc=org`. But, the first baseDn is `dc=spring,dc=org`

are there any way to avoid that? 🤔 

Thanks in advance

eddumelendez:
forgot I can do this `spring.ldap.embedded.base-dns[0]:dc=spring,dc=org`. I think it needs to be documented.

philwebb:
Closing in favor of PR #11764

PR #11764, opened 2018-01-25
Add support for multi baseDn

eddumelendez:
`spring.ldap.embedded.base-dn` property supports multiple baseDn now. It
should follows the following format:

```
spring.ldap.embedded.base-dn[0]=dc=spring,dc=org
spring.ldap.embedded.base-dn[1]=dc=pivotal,dc=io
```

See gh-11752

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
@@ -20,6 +20,7 @@ import java.io.InputStream;
 import java.util.HashMap;
 import java.util.Map;
 
+import javax.annotation.PostConstruct;
 import javax.annotation.PreDestroy;
 
 import com.unboundid.ldap.listener.InMemoryDirectoryServer;
@@ -33,7 +34,6 @@ import org.springframework.boot.autoconfigure.AutoConfigureBefore;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.ldap.LdapAutoConfiguration;
 import org.springframework.boot.autoconfigure.ldap.LdapProperties;
 import org.springframework.boot.autoconfigure.ldap.embedded.EmbeddedLdapProperties.Credential;
@@ -50,6 +50,7 @@ import org.springframework.core.env.PropertySource;
 import org.springframework.core.io.Resource;
 import org.springframework.ldap.core.ContextSource;
 import org.springframework.ldap.core.support.LdapContextSource;
+import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks, @eddumelendez. The need for `[n]` when using a property (rather than YAML) to configure this is rather unpleasant. I'm not sure what the answer is, but requiring someone who only wants to use a single base dn to deal with the `[n]` syntax to avoid splitting on the `,` feels like a step backwards to me.

I don't know much about LDAP. Does a base dn always contain exactly one comma? I wonder if we'd be better consuming a `String` and then splitting it on every second comma rather than relying on the binding to perform the splitting.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
```diff
@@ -64,7 +65,6 @@
 @EnableConfigurationProperties({ LdapProperties.class, EmbeddedLdapProperties.class })
 @AutoConfigureBefore(LdapAutoConfiguration.class)
 @ConditionalOnClass(InMemoryDirectoryServer.class)
-@ConditionalOnProperty(prefix = "spring.ldap.embedded", name = "base-dn")
```
snicoll: I think that's pretty bad. We should let the condition infra does its job  rather than failing as you've done instead.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
```diff
@@ -88,6 +88,11 @@ public EmbeddedLdapAutoConfiguration(EmbeddedLdapProperties embeddedProperties,
 		this.environment = environment;
 	}
 
+	@PostConstruct
+	public void validateBaseDns() {
+		Assert.notEmpty(this.embeddedProperties.getBaseDn(), "No baseDn found.");
+	}
```
snicoll: That's a regression isn't it? Previously we'd just backoff.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
eddumelendez: Yes, it is :cry: Using the following format `spring.ldap.embedded.base-dn[0]:dc=spring,dc=org` doesn't match with `@ConditionalOnProperty(prefix = "spring.ldap.embedded", name = "base-dn")`

wilkinsona:
Answering my own question above about the structure of a base DN, it is [not guaranteed to contain any particular number of commas](https://msdn.microsoft.com/en-us/library/aa366101(v=vs.85).aspx) so parsing multiple DNs from a single, comma-separated string does not appear to be possible.

Right now, I can't think of a good way to do this with property-based configuration.

snicoll:
@wilkinsona perhaps it should be mapped-based then. I'd hate to do that but that seems the only sensible way I could think of. If we had other properties per base DN that would be ideal of course.

wilkinsona:
Unfortunately, I don't think a map-based approach makes sense here. There's nothing that needs to be configured on a per-base-DN basis so I think the property really should be separate with nothing nested beneath it.

What I'd really like is for it to be possible for the following to be treated as a single entry:

```
spring.ldap.embedded.base-dn=o=company,dc=com
```

And for the following to be treated as two entries:

```
spring.ldap.embedded.base-dn[0]=o=company,dc=com
spring.ldap.embedded.base-dn[1]=dc=security
```

My goal is to only inflict the `[n]` syntax on users who want to configure more than on base DN.

philwebb:
Thanks @eddumelendez, I've merged this into master with a change to use the recently added `@Delimiter` support. The format suggested by @wilkinsona above now works!

eddumelendez:
Cool! Thanks @philwebb

Issue #11338, opened 2017-12-13
2.0.0.M7 “spring-boot-starter-actuator”  caused OOM


chenjinkai:
maybe this jar will caused the OOM problem

<dependency>
    <groupId>org.latencyutils</groupId>
    <artifactId>LatencyUtils</artifactId>
    <version>2.0.3</version>
</dependency>

chenjinkai:
after my application runned 24 hours, it will be shown.

chenjinkai:
dump the memory, amost all objects is instances of org/LatencyUtils/LatencyStats$PauseTracker .

bclozel:
Did you report that issue to https://github.com/LatencyUtils/LatencyUtils ?

philwebb:
Possibly related to #11333

jkschneider:
> Possibly related to #11333

Only vaguely in that they are related to the same mechanism.

jkschneider:
@chenjinkai Are you super-positive that this is running on Boot 2 M7 and that you have a version of micrometer >= 1.0.0-rc.4 on your runtime classpath?

We had a memory leak in 1.0.0-rc.3 (Boot 2 M6) that was fixed in rc4. It's not impossible that there is still a leak, but I'm having a difficult time reproducing it.

jkschneider:
One idea: perhaps you have a custom metric that is running up the cardinality on a particular tag? In effect, you create a separate `Timer` for every unique combination of name and tags. Since there are by default 5 histograms kept in a ring buffer for every timer with 1 `LatencyStats` and therefore 1 `PauseDetector` per histogram, in such a scenario it might appear that `PauseDetector` is the dominator. If you have 1/5 as many `Timer` instances as `PauseDetector`'s when the OOM occurs, we can be confident this is a tag cardinality problem and not a general leak in `Timer`.

mweirauch:
Here's some plotting of a test app running with micrometer-core:1.0.0-rc.4 (~4 weeks) and rc.5 (~11 hours). No problems observable. Note, this is with a steady amount of meters.

![evam-micrometer_1 0 0-rc 4-sb_1 5 8-4w](https://user-images.githubusercontent.com/1134031/33981321-a4e454f6-e0ac-11e7-957d-2749db470ca0.png)

[...]

PR #11776, opened 2018-01-25
Upgrade to Micrometer 1.0.0-SNAPSHOT

jkschneider:
* We now have [Jersey 2 support](https://github.com/micrometer-metrics/micrometer/pull/228) with the module `micrometer-jersey2`. It's not clear to me how we could register `MetricsApplicationEventListener` for Jersey 2 based Boot apps.
* I think the servlet filter is generic enough now to call it `MetricsFilter` rather than `WebMvcMetricsFilter`. At any rate, the goal is to make it _not_ WebMvc specific.

The following issues are addressed by this PR:
* #11348
* #11338


jkschneider:
I'll delete this comment when I'm done pushing changes.

philwebb:
I'm having a bit of a difficult time getting this one into a state that we can merge. I've so jar just been working on the first commit in the set. I've split it apart and tried to create smaller commits with related concerns. I've rename the MVC classes back to their old names (they are still very tied to MVC so I don't like the very generic `ServletMetricsFilter` name.

My branch is [here](https://github.com/philwebb/spring-boot/commits/gh-11776).

Things go relatively OK until the last commit. The changes to `WebMvcMetricsFilter` are a big concern to me. I've added some FIXME comments in the code but the main worry is around the `asyncTimingContext` and the fact that it's a synchronized map.

jkschneider:
@philwebb Good idea on the request attribute. Tested it against `micrometer-spring-legacy` successfully and committed there: https://github.com/micrometer-metrics/micrometer/commit/44a5d8834cd92a03c5b92adc1f37bb2f877f8c36#diff-974dd55a85a31feb0d4a829577d72046d

ITman1:
I suggest to add `@ConditionalOnMissingBean` on every `XxxxxxxMeterRegistry` bean auto-configuration. I know there is currently `@ConditionalOnProperty("management.metrics.export.xxxxxxx.enabled")`, which might be enough for application using spring-boot directly, but not for libraries which needs some customizations without affecting properties.

philwebb:
Sorry but there is just way too much going on in the PR for me to be able to merge it before RC1. I've split out the following issues which we can prioritize and deal with on a case-by-case basis:

#11797
#11798
#11799
#11800
#11801
#11802
#11803

PR #11782, opened 2018-01-25
Fix Optional Check

rajadilipkolli:
when Optional is in use, there should never be a question of returning or receiving null from a call. Hence checking with isPresent


philwebb:
Thanks for suggestion, but in this case the `Optional` itself can be `null`. This area of code was to safe re-evaluating the same `ENABLED_BY_DEFAULT_KEY` property multiple times.

The states are:
`null` - We've not get cached the result
`Optional` - We've cached the result which may contain `Boolean.TRUE`, `Boolean.FALSE` or `null`.

PR #11789, opened 2018-01-26
Align missed bits from trace endpoint changes

izeye:
This PR aligns missed bits from trace endpoint changes.


wilkinsona:
Thank you, @izeye!

Issue #7936, opened 2017-01-10
Server info leak after corrupted request


dreambrother:
I want to prevent any server info leak in error responses. I disabled whitelabel and stacktrace printing
`server.error.include-stacktrace=never`
`server.error.whitelabel.enabled=false`

I also tried to add my own error controller and custom error pages, but when I send corrupted request, embedded Tomcat responses with server info and stacktrace.

Example of corrupted request:
```
curl 'http://localhost:8080/foo' -H 'content-type: multipart/form-data; boundary=----WebKitFormBoundaryHBic2VjAi3E215bk' --data-binary $'------WebKitFormBoundaryHBic2VjAi3E215bk\r\nContent-Disposition: form-data; name="file"; filename="doc-xlsx.xlsx"\r\nContent-Type: application/vnd.openxmlformats-officedocument.spreadsheetml.sheet\r\n\r\n\r\n------WebKitFormBoundaryHBic2VjAi3E215bk\r\nContent-Disposition:  '
```

I tried to turn off report and server info printing manually, but exception message still presents in responses (it contains tomcat's and spring's classes names).

```java
@Bean
public EmbeddedServletContainerFactory servletContainer() {
    // Disable tomcat version and exception logging by ErrorReportValve
    TomcatEmbeddedServletContainerFactory tomcat = new TomcatEmbeddedServletContainerFactory();

    ErrorReportValve errorReportValve = new ErrorReportValve();
    errorReportValve.setShowReport(false);
    errorReportValve.setShowServerInfo(false);

    tomcat.addContextValves(errorReportValve);

    return tomcat;
}
```

I think it's possible to overwrite `TomcatEmbeddedServletContainerFactory#getEmbeddedServletContainer` to return `Tomcat` instance with different `host` property, `StandardHost#errorReportValveClass` should be changed for another class in that case. I didn't try this solution, because it creates too much coupling to underlying implementations. Looks like spring-boot should be able to do this.

wilkinsona:
Could you please provide a small sample that illustrates the problem you have described? I want to be sure that we're investigating exactly what you're describing.

dreambrother:
I've created a project https://github.com/dreambrother/spring-boot-server-info-leak-example illustrating issue described above. I've added public/error/500.html page to handle 500 errors, and it works as expected for exceptions thrown in controller 
```java
@RequestMapping("/exception")
public String error() {
    throw new RuntimeException("Error test");
}
```
[...]

Issue #6231, opened 2016-06-26
Register Tomcat valves at the engine container level


qerub:
In Tomcat, a valve can be active at a {engine, host, context} container level.

Spring Boot allows activation of {`AccessLogValve`, `RemoteIpValve`} through properties and activates them at a context container level.

Given that Spring Boot can only activate a single {`AccessLogValve`, `RemoteIpValve`}, wouldn't it be better to add them at the top of the container hierarchy (i.e. the engine)?

For `RemoteIpValve`, the Tomcat documentation states:

> Normally, this Valve would be used at the Engine level.

For `AccessLogValve`, the Tomcat documentation states:

> Where Tomcat cannot identify the Context that would have handled the request, e.g. in cases where the URL is invalid, Tomcat will look first in the Engine, […] for an AccessLog implementation.

Why bother? Well, I noticed that Spring Web MVC's `/error` action is invoked outside the scope of the `RemoteIpValve` (as configured by default), making my application capture the wrong IP addresses. Moving the `RemoteIpValve` to the engine resolves that issue.

I'd be happy to provide a patch if you think this is a change worth making.

philwebb:
On the surface it sounds sensible, we'll discuss it at our next meeting. If the code isn't too involved a patch would be most welcome, if it's going to eat a lot of your time then it's best to wait until we've talked about it.

wilkinsona:
When a failure occurs Tomcat forwards the request to our error page, `/error`. This is done in `StandardHostValve`. As its name would suggest, this `Valve` is configured on the `Host`.

`RemoteIpValve` follows this basic pattern:

```
Store request's original settings
Update request based on X-Forwarded-* headers
try {
    Call the next valve
} 
finally {
    Restore the request's original state
}
```

When it's operating at the context level, this means that the finally block runs before `StandardHostValve` catches the exception and forwards the request to `/error` so the error controller is called with the request in its original state so it doesn't see the changes that resulted from processing the `X-Forwarded-*` headers.

wilkinsona:
Duplicates #6311

PR #11790, opened 2018-01-26
configure ErrorReportValve for tomcat

panchenko:
In my application I have disabled the spring-boot error handling via 

````java
@SpringBootApplication(exclude = ErrorMvcAutoConfiguration.class)
````

in this case if I use a command from #7936 then a stack trace with server version is printed by tomcat.

In tomcat it's done by `org.apache.catalina.valves.ErrorReportValve`, which is added by  `org.apache.catalina.core.StandardHost.startInternal()`, if not already there.
This class can be configured appropriately, but by default it prints everything.

In this patch if `include-stacktrace=never` then a properly configured `ErrorReportValve` is added to tomcat Host, avoiding the server info leaks.

### DIFF
--- a/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -36,7 +36,9 @@ import io.undertow.UndertowOptions;
 import org.apache.catalina.Context;
 import org.apache.catalina.connector.Connector;
 import org.apache.catalina.valves.AccessLogValve;
+import org.apache.catalina.valves.ErrorReportValve;
 import org.apache.catalina.valves.RemoteIpValve;
+import org.apache.commons.logging.LogFactory;
 import org.apache.coyote.AbstractProtocol;
 import org.apache.coyote.ProtocolHandler;
 import org.apache.coyote.http11.AbstractHttp11Protocol;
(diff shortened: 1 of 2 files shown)


philwebb:
I consider this a bug. I wonder if we can use one of the Customizer callback to save needing to add the extra methods to the Tomcat Factory?

panchenko:
It seems like there are no existing customers for this purpose (only for Connector and Context).
We can introduce a new interface `TomcatCustomezer` and that would be more flexible, as currently the documented way for that is overriding `getTomcatEmbeddedServletContainer(Tomcat tomcat)`.
WDYT?

wilkinsona:
There’s some background on the current support for configuring valves in https://github.com/spring-projects/spring-boot/issues/6231.

wilkinsona:
Sorry for the confusion. My comment was intended to provide some information for the discussion about how to configure things: a customiser vs methods on the factory. We’ve used the latter thus far.

panchenko:
So, theoretically it's possible in 3 different ways:
- public API for host valves
- public API for tomcat customizers
- some simpler protected API for internal use in this package

Please let me know if I should change the implemenation  :-)

wilkinsona:
I share Phil's desire to do this without adding public API. If we need to add public API, then it would make sense to add an API for host valves. This would align with the existing API for engine valves. Thankfully, it appears that won't be necessary.

I think this can be achieved using a `TomcatContextCustomizer`, however we'd need a small change to `TomcatEmbeddedServletContainerFactory` first. It currently adds the context to the host after it's configured the context using any customisers but before its postProcessContext extension point. If we change it to add the context to the host before configuring the context then a customiser and the extension point will see the context in the same state and a customiser will be able to call `getParent` to access the context's host.

This would allow a customiser to add an appropriately configured `ErrorReportValve`:

```java
private void customizeErrorReportValve(
        TomcatEmbeddedServletContainerFactory factory) {
    factory.addContextCustomizers((new TomcatContextCustomizer() {

        @Override
        public void customize(Context context) {
            ErrorReportValve valve = new ErrorReportValve();
            valve.setShowServerInfo(false);
            valve.setShowReport(false);
            ((Host) context.getParent()).getPipeline().addValve(valve);
        }

    }));
}

@panchenko would you like to update your PR along these lines?

panchenko:
@wilkinsona Please take a look

wilkinsona:
Looks good. Thanks.

wilkinsona:
Thank you, @panchenko. I've merged this into 1.5.x and forwards into master along with [a little bit of polish and a few tests](https://github.com/spring-projects/spring-boot/commit/5a74f63f7ca02359511ce2ade76a000513575ea5).

PR #11809, opened 2018-01-27
Fix to assert parameter names in EndpointServlet.withInitParameters()

izeye:
This PR fixes to assert parameter names in `EndpointServlet.withInitParameters()`.


snicoll:
Good catch  @izeye! There was missing tests for that particular use case so I've added a couple in  6569ed8

PR #11811, opened 2018-01-27
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/ServletEndpointRegistrar.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/ServletEndpointRegistrar.java
@@ -29,8 +29,8 @@ import org.springframework.boot.web.servlet.ServletContextInitializer;
 import org.springframework.util.Assert;
 
 /**
- * {@link ServletContextInitializer} to register {@link ExposableServletEndpoint} servlet
- * endpoints.
+ * {@link ServletContextInitializer} to register {@link ExposableServletEndpoint servlet
+ * endpoints}.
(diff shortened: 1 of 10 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/ServletEndpointRegistrar.java
```diff
@@ -29,8 +29,8 @@
 import org.springframework.util.Assert;
 
 /**
- * {@link ServletContextInitializer} to register {@link ExposableServletEndpoint} servlet
- * endpoints.
```
snicoll: The intent here is to replace the link with the text "servlet endpoints". I prefer that than your proposal.

izeye: @snicoll Thanks for the feedback! I just tried to make it align with https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/annotation/ServletEndpointDiscoverer.java#L35 and https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/annotation/ServletEndpointsSupplier.java#L23. What do you think?

snicoll: Bah, I am an idiot. I read your change backwards. You did what the intent was, not the other way around. Sorry :(

PR #11812, opened 2018-01-27
polish/fix issue "Unused method parameters should be removed"

igor-suhorukov:
Remove unused method parameters to get rid [static code inspection issue](https://sonarcloud.io/project/issues?id=org.springframework.boot%3Aspring-boot-build&resolved=false&rules=squid%3AS1172) sonarcloud [squid:S1172](https://sonarcloud.io/organizations/igor-suhorukov-github/rules#rule_key=squid%3AS1172)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationImportSelector.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationImportSelector.java
@@ -84,7 +84,7 @@ public class AutoConfigurationImportSelector
 
 	@Override
 	public String[] selectImports(AnnotationMetadata annotationMetadata) {
-		if (!isEnabled(annotationMetadata)) {
+		if (!isEnabled()) {
 			return NO_IMPORTS;
 		}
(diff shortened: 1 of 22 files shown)


igor-suhorukov:
all invocation of ServletComponentHandler#extractUrlPatterns with attribute parameter = "urlPatterns"
SpringIntegrationMetrics#registerGauge looks like logical mistake with parameter/field before refactoring

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationImportSelector.java
```diff
@@ -107,7 +107,7 @@
 		}
 	}
 
-	protected boolean isEnabled(AnnotationMetadata metadata) {
```
snicoll: That's a protected method and an extension might use that parameter

wilkinsona:
The work-in-progress amendments are in [this branch](https://github.com/wilkinsona/spring-boot/tree/gh-11812). They:

- Remove methods/classes that were left unused as a result of the proposed changes
- Correct some cases where the change should have been to use a variable rather than remove it

PR #11817, opened 2018-01-28
Replace casted Mockito.any() calls

dreis2211:
Hi,

this PR replaces casted `any()` calls with `any(Class)`. So `(Servlet) any()` will become `any(Servlet.class)` for example. Please note, that a few of those replacements were unnecessary casts as the type could be inferred. I kept the class around on purpose to not lose the (intended) expressiveness, but I have no strong feelings against using only `any()` for them.

Let me know what you think.

Cheers,
Christoph


dreis2211:
Test failure seems to be unrelated.

PR #11818, opened 2018-01-28
Use InvocationOnMock.getArgument() where possible

dreis2211:
Hi,

while working on #11817 I noticed some calls to `InvocationOnMock.getArguments()[int i]` that could be replaced with the single argument variant. This has the nice side-effect of getting rid of some casts.

NOTE: The copyright changes assume that PR #11817 is merged as well in order to avoid eventual conflicts when both are accepted.

Cheers,
Christoph


snicoll:
Thanks again Christoph!

Issue #11822, opened 2018-01-29
ClassCastException with Redis cache and devtools


javamonkey:
<parent>
		<groupId>org.springframework.boot</groupId>
		<artifactId>spring-boot-starter-parent</artifactId>
		<version>2.0.0.M7</version>
    </parent>

when i first access the method getMenu,everything is fine. menu object  is cached to redis server. but when i call it again,i get the following exception:

 java.lang.ClassCastException: com.bee.sample.ch14.entity.Menu cannot be cast to com.bee.sample.ch14.entity.Menu
	at com.sun.proxy.$Proxy68.getMenu(Unknown Source) ~[na:na]
	at com.bee.sample.ch14.controller.AdminCrontroller.getMenu(AdminCrontroller.java:28)

**but 2.0.0.M2 is work well.**

i guess there is something wrong  with  JdkSerializationRedisSerializer has null classloader . If i  construct JdkSerializationRedisSerializer  whit Thread.currentThread().getContextClassLoader(),It work!!!

	@Bean
	public RedisCacheManager getRedisCacheManager(RedisConnectionFactory connectionFactory){
		RedisCacheWriter cacheWriter = RedisCacheWriter.lockingRedisCacheWriter(connectionFactory);
		ClassLoader loader = Thread.currentThread().getContextClassLoader();
		JdkSerializationRedisSerializer jdkSerializer = new JdkSerializationRedisSerializer(loader);
		SerializationPair<Object> pair = SerializationPair.fromSerializer(jdkSerializer);
		
		RedisCacheConfiguration cacheConfig = RedisCacheConfiguration.defaultCacheConfig().serializeValuesWith(pair);
		..... //ignore
		return cacheManager;
	}

snicoll:
Looks related to #9734 and the new `RedisCacheManager`.

@mp911de have you changed the way the default serializer is built on your side?

@javamonkey this probably happens because you are using devtools.

javamonkey:
https://github.com/javamonkey/rediscachetest  
this simple project can reproduce this issues

mp911de:
Yes, it's caused by/related to #9734. We're configuring a JDK serializer/deserializer without passing a class-loader. `RedisCacheConfiguration` should get configured with a `JdkSerializationRedisSerializer` that has the class loader set.

davidljuba15031979:
https://github.com/spring-projects/spring-boot/issues/9444

PR #11825, opened 2018-01-29
Configure default RedisSerializer

mp911de:
We now configure a default `RedisSerializer` bean using JDK serialization with the appropriate ClassLoader. RedisSerializers are used in imperative/reactive Redis templates and for Redis Cache. This bean centralizes configuration and allows easier override of the default serialization mechanism.

---

Related to #11822.


snicoll:
@mp911de and I have discussed and I am not keen to expose a default serializer (this is a new feature and not strictly required to fix the bug). What I'll do is adapt the cache configuration a bit and Spring Data REST will add a `RedisCacheConfiguration.defaultCacheConfig(ClassLoader)` that we can reuse later.

PR #11845, opened 2018-01-30
Reduce StringBuilder creation in TypeExtractor.visitDeclared()

izeye:
This PR reduces `StringBuilder` creation in `TypeExtractor.visitDeclared()`.


snicoll:
Thanks @izeye - I wanted to fix that in `1.5.x`  as well and cherry picking that change there turned out to be quite annoying so I've rolled out my own.

PR #11848, opened 2018-01-30
Add spring-boot-cli support for Windows MINGW environments

### LINKED ISSUE #11846: Support spring-boot-cli in MinGW Windows environments
## Problem Description
Currently the spring-boot-cli (1.5.9.RELEASE) bash script can deal with Cygwin and Darwin environments but not with MinGW (http://www.mingw.org/) environments. 
MinGW is minimalistic Unix Environment for Windows and is e.g. the default bash environment in the Cmder (http://cmder.net/) console emulator.

## Request
It would be great if the spring-boot-cli script would also run in MinGW environements.

## Current Behaviour
Current behaviour when trying to execute it is (sorry, it is a german environment):
```
$ spring version
Fehler: Hauptklasse org.springframework.boot.loader.JarLauncher konnte nicht gefunden oder geladen werden
$ uname
MINGW32_NT-6.1-WOW
```

## Expected Behaviour
```
$ spring version
Spring CLI v1.5.9.RELEASE
```

Note that in a Windows/cmd.exe environment it works fine.

stadler:
Fixes #11846 

As MINGW actually contains a light version of cygwin it is compatible with cygwin. Therefore only the match for MINGW is necessary.
Tested in my environment and it works fine.
Unfortunatelly I don't see a way to easily test this with an integration test.

### DIFF
--- a/spring-boot-project/spring-boot-cli/src/main/executablecontent/bin/spring
+++ b/spring-boot-project/spring-boot-cli/src/main/executablecontent/bin/spring
@@ -8,6 +8,10 @@ case "`uname`" in
 		cygwin=true
 		;;
 
+	MINGW*)
+		cygwin=true
+		;;
+
 	Darwin*)
 		darwin=true
 		;;


philwebb:
Thanks very much for the PR. I've applied this to 1.5.x since it seems like such a low risk but useful change.

PR #11856, opened 2018-01-31
CompositeMeterRegistryPostProcessor now injects a non-null clock

jkschneider:
The new `CompositeMeterRegistryPostProcessor` passes `null` as the first argument to `CompositeMeterRegistry`. It should pass the autoconfigured `Clock` coming from `MetricsAutoConfiguration`.

This manifests itself as an NPE when attempting later to use a timer/long task timer.

See https://github.com/micrometer-metrics/micrometer/issues/384.


philwebb:
Unfortunately we can't use the fix as it is because it will cause early bean initialization.

jkschneider:
At least as a stopgap to avoid NPEs, can we pass `Clock.SYSTEM`?

PR #11867, opened 2018-01-31
Fix default HttpMessageConverters not being replaced

diesieben07:
In case a `HttpMessageConverter` is a replacement for a default, we need to skip that default converter instead of adding both. Otherwise this results in duplicate converters, where one is completely unconfigured.

In the simplest case (empty project with just `org.springframework.boot:spring-boot-starter-web`) this causes two `MappingJackson2HttpMessageConverter` to be created and added to the web mvc systems. One of these converters is the default converter using a plain, unconfigured `ObjectMapper`, the other is the properly configured converter.


wilkinsona:
Thanks for the PR. The current behaviour is intentional and is described on the constructors that take additional converters where it says the following:

> Items are added just before any default converter of the same type (or at the front of the list if no
> default converter is found)

I'm going to close this PR as we can't just change the behaviour. If the behaviour is causing you a problem, please open an issue describing that problem in as much detail as possible and we'll see what can be done.

diesieben07:
After further investigation it turns out that me thinking this was a bug was caused by flawed debugging on my part. Apologies.

Issue #11879, opened 2018-02-01
WildFly deployment fails with `2.0.0.RC1`


vpavic:
Upgrading from `2.0.0.M7` to `2.0.0.RC1` I've encountered a problem with deployment to WildFly 10.1. The problem appears to be related to Micrometer:

```
21:26:57,081 INFO  [stdout] (ServerService Thread Pool -- 67) 2018-02-01 21:26:57.080 ERROR 4015 --- [read Pool -- 67] o.s.boot.SpringApplication               : Application run failed
21:26:57,081 INFO  [stdout] (ServerService Thread Pool -- 67) 
21:26:57,082 INFO  [stdout] (ServerService Thread Pool -- 67) org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'meterRegistryPostProcessor' defined in class path resource [org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfiguration.class]: Bean instantiation via factory method failed; nested exception is org.springframework.beans.BeanInstantiationException: Failed to instantiate [org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryPostProcessor]: Factory method 'meterRegistryPostProcessor' threw exception; nested exception is org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'jvmGcMetrics' defined in class path resource [org/springframework/boot/actuate/autoconfigure/metrics/MeterBindersConfiguration.class]: Post-processing of merged bean definition failed; nested exception is java.lang.IllegalStateException: Failed to introspect Class [io.micrometer.core.instrument.binder.jvm.JvmGcMetrics] from ClassLoader [ModuleClassLoader for Module "deployment.demo.war:main" from Service Module Loader]
21:26:57,083 INFO  [stdout] (ServerService Thread Pool -- 67) 	at org.springframework.beans.factory.support.ConstructorResolver.instantiateUsingFactoryMethod(ConstructorResolver.java:587) ~[spring-beans-5.0.3.RELEASE.jar:5.0.3.RELEASE]
21:26:57,083 INFO  [stdout] (ServerService Thread Pool -- 67) 	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.instantiateUsingFactoryMethod(AbstractAutowireCapableBeanFactory.java:1250) ~[spring-beans-5.0.3.RELEASE.jar:5.0.3.RELEASE]
[...]

PR #11886, opened 2018-02-02
Update deployment tests to include Actuator

vpavic:
This PR updates the existing deployment test to include Actuator. With this change, deployment tests can be used to reproduce the issue reported in #11879.

Note that the WildFly test app also requires exclusion of `org.slf4j` module to work with Actuator. This can be confirmed by adding `management.metrics.binders.jvm.enabled=false` to avoid problem reported in #11879.

### DIFF
--- a/spring-boot-tests/spring-boot-deployment-tests/spring-boot-deployment-test-glassfish/pom.xml
+++ b/spring-boot-tests/spring-boot-deployment-tests/spring-boot-deployment-test-glassfish/pom.xml
@@ -32,6 +32,10 @@
 			</exclusions>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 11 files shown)


vpavic:
I've updated the PR with tests to verify Actuator and a workaround in WildFly deployment configuration (see [this comment](https://github.com/spring-projects/spring-boot/issues/11879#issuecomment-362887846)).

vpavic:
The problem that was a motivation for this (originally reported as #11879) was fixed in micrometer-metrics/micrometer#401. With the last update to this PR WildFly configuration was updated to make `com.sun.management` available so this shouldn't be blocked either way now.

wilkinsona:
Awesome stuff, @vpavic! Thank you.

Issue #11800, opened 2018-01-26
Support property based MeterFilters


philwebb:
https://github.com/micrometer-metrics/micrometer/issues/366 / https://github.com/spring-projects/spring-boot/pull/11776/commits/d034a7f2b50863b5447bb3733f39c3bbbfb8e9c3

jkschneider:
To help you better imagine the use case of using filters to control distribution statistics, consider timers that are instrumented in low-level core libraries (like the instrumentation that now exists in HikariCP and Rabbit). These authors can't make a general-purpose decision about whether it is useful for users of their library to see certain summary stats like client-side percentiles. For example:

1. Billy is using InfluxDB, which does not have a rich-enough query language to compute aggregable percentiles from histograms. Billy might choose to ship a Micrometer-computed P95, knowing that the usefulness of this is limited to the tag cardinality of the underlying metric. For example, if the timer is tagged with "host" and "result" and his organization only runs 10 service instances where there are 2 possible results, he could plausibly plot a non-aggregable P95 by simply plotting all the individual time series independently and monitoring for outliers. If Billy's org ever grows, Influx's limitations will eventually them away from it to something more capable.

2. Sue is using Prometheus and never wants to ship Micrometer-computed percentiles because she has access to `histogram_quantile` and so the usefulness of her plots doesn't need to be constrained by tag cardinality. She chooses to ship the histogram buckets.

3. Jane is using Wavefront and doesn't care about percentile distributions at all. She is comfortable monitoring the range between the distribution average (a measure of centrality that is neither better or worse than P50 in the general case) and max. Wavefront publishes a blog about their new aggregable percentiles feature and Jane gets P99 on key latency metrics via an additional property.

4. (This one is real) Clint and DOMO have adopted a "instrument all the things" approach but are worried about the scalability of their on-prem Prometheus cluster. If their Prometheus cluster ever gets stressed, they want to be able to drop summary stats on some less-critical latency metrics and perhaps disable whole swaths of other metrics. They want to accomplish both in a property-driven way such that they could ease pressure on their monitoring system across their whole stack with a combination of something like Archaius/Config Server.

There is no way for the Rabbit client author to satisfy all of these needs without shipping wasteful extra time series to each of our imaginary users. Our recommendation for library authors is always just instrument with plain timers and let application developers decide at the last minute which summary statistics they want via filters.

snicoll:
[...]

PR #11888, opened 2018-02-02
recordRequestPercentiles has no effect on auto-timed RequestMappings

shakuzen:
Using Spring Boot 2.0.0.RC1 with the Actuator starter and setting `management.metrics.web.server.record-request-percentiles` to `true` does not have any effect. It should cause percentile histograms to be recorded for `RequestMapping` endpoints' timing.

The value of `recordRequestPercentiles` from the `MetricProperties` was not being considered to determine whether a histogram should be recorded for percentile usage. Now, setting `recordRequestPercentiles` has the expected effect of configuring histograms on auto timed requests.

/cc @jkschneider

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/servlet/WebMvcMetricsConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/servlet/WebMvcMetricsConfiguration.java
@@ -60,7 +60,8 @@ public class WebMvcMetricsConfiguration {
 		Server serverProperties = properties.getWeb().getServer();
 		WebMvcMetricsFilter filter = new WebMvcMetricsFilter(context, registry,
 				tagsProvider, serverProperties.getRequestsMetricName(),
-				serverProperties.isAutoTimeRequests());
+				serverProperties.isAutoTimeRequests(),
+				serverProperties.isRecordRequestPercentiles());
 		FilterRegistrationBean<WebMvcMetricsFilter> registration = new FilterRegistrationBean<>(
 				filter);
 		registration.setDispatcherTypes(DispatcherType.REQUEST, DispatcherType.ASYNC);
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for catching this, @shakuzen. And for such a thorough PR. The proposed changes look good to me. @jkschneider can you cast an eye over them before we merge please?

shakuzen:
Looking at this a little bit closer, something stood out to me. The JavaDoc on `recordRequestPercentiles` (emphasis is mine):

> Whether or not instrumented requests record percentiles histogram **by default**. Can be overridden by adding '@​Timed' to a request endpoint setting 'percentiles' to true.

Reading this leads me to believe the intention was to apply `recordRequestPercentiles` as a default to more than just the auto timed requests - that is, even `@RequestMapping`s with a `@Timed` should take the value of `recordRequestPercentiles` unless overridden like `@Timed(histogram = false)`. Currently, this pull request only applies `recordRequestPercentiles` to auto timed requests - those without an explicit `@Timed`.

However, I don't see how we could apply the default to `@Timed` requests unless the `histogram` attribute of `@Timed` changes from a `boolean` to a `Boolean`, allowing us to easily discern when the value is unset (`null`) and the configurable default should apply.

@jkschneider can you please shed some light on the intended behavior?

jkschneider:
@shakuzen @wilkinsona Sorry for the confusion. Ultimately, `recordRequestPercentiles` is destined for elimination once we get https://github.com/spring-projects/spring-boot/issues/11800 complete, because we'll have a property-driven mechanism to enable percentile histograms for _all_ timers and summaries, so won't need these targeted ones.

You can accomplish the same today by wiring a `MeterFilter` programmatically:

```
@Bean
public MeterFilter percentilesOnRequests() {
    return new MeterFilter() {
        @Override
        public HistogramConfig configure(Meter.Id id, HistogramConfig config) {
            if(id.getName().equals("http.server.requests"))
                return HistogramConfig.builder()
                    .percentilesHistogram(true)
                    .build()
                    .merge(config);
            return config;
        }
    };
}
```

jkschneider:
The overlap with `@Timed` hadn't even occurred to me, so there isn't a well-thought-out intended behavior. Honestly, I'm not sure how this should work. It would be great if we could change `histogram` on `@Timed` to a `Boolean`, but of course Java doesn't allow that.

jkschneider:
Percentiles and histogram defined on `@Timed` are now overridden by `MeterFilter`s currently, and I think that makes sense?

Issue #11860, opened 2018-01-31
server.address with an IP address can bind jetty to different IP address


henrjk:
We want to start a springboot/gradle rest server application configured to use jetty, so that it binds to a particular IP address using the `-Dserver.address` java parameter.

We are using spring boot 1.5.4  which pulls in jetty-server version 9.4.5.v20170502.

The spring boot integration code does a reverse lookup on the IP address configured with `server.address`. With the IP address 10.10.10.10 which this resolves in our context to the hostname.

When jetty server socket is bound the hostname is resolved to an IP address.
However if multiple network adapters are available this can be a different IP
address then what was specified with `server.address`. As a consequence the server is then not available on the specified IP address.

The fix seems straight forward. In `JettyEmbeddedServletContainerFactory.java` calls to getHostName() can be replaced with getHostString().
A side benefit is that this could increase startup time as reverse name resolution can be quite slow.

Do you see problems with this change?

I could contribute a PR for the 1.5.x branch. 
I have not yet investigated whether there is a similar issue in master.

wilkinsona:
I can't see any problems with what you've proposed. A pull request would be most welcome. Thanks.

henrjk:
In my fork I made the described change and running all unit test from `/spring-boot/src/test/java`  in Eclipse remains green.  Are there any other tests I should be running? 

A unit test which which validates that the host name from -Dserver.address will remain an IP string in the server connectors with a JettyServerCustomizer associated with the factory might be worthwhile. Do you have a suggestion on whether such a test should be provided and if so where to add this?  

Would this change require signing a CLA?

wilkinsona:
Thanks for the update.

> Do you have a suggestion on whether such a test should be provided and if so where to add this?

Yes, a test would be good please. I'd add it to `JettyEmbeddedServletContainerFactoryTests` where you can use the `setAddress(InetAddress)` method on the factory. You should then be able to start the container, and retrieve the connector from the server to verify its host.

> Would this change require signing a CLA?

The change is non-trivial so, yes, signing the CLA will be required.

henrjk:
To test the currently broken state my test code needs an IP address of the local machine which reverse name resolves without an UnknownHostException. 
The code for this involves looking at the network interfaces. However for better testing I would need to do this work outside of a VM where I currently working. This will probably not happen until Monday unless you would have code that does this already somewhere that I could use.

wilkinsona:
[...]

PR #11889, opened 2018-02-02
Fix jetty potentially binding to wrong IP address

henrjk:
When jetty is configured to an IP address whose reverse name resolution
followed by a subsequent name resolution does not lead to the same IP
address, jetty binds to that different IP address.

This is configured with the server.address parameter.

Changed code to not use reveres name resolution.

Fixes gh-11860


wilkinsona:
Thanks for the PR, @henrjk. I've merged this into 1.5.x and forwards into master. I also applied a [little bit of polish](https://github.com/spring-projects/spring-boot/commit/a1b823fc4320f13506066322c6763e09806bcfd9), largely to simplify the new test.

PR #11893, opened 2018-02-02
Integration Micrometer Autoconfiguration

garyrussell:
- also remove placeholder `SpringIntegrationMetrics`


philwebb:
Unfortunately I can't merge this one in its current form. Having `MicrometerMetricsFactory` require a `MeterRegistry` bean is problematic because we need to ensure correct auto-configuration ordering. We can't to that in `spring-boot-autoconfigure` because it can't know about actuator concerns. I've opened #11985 to look at a different solution.

PR #11917, opened 2018-02-06
Add support of git and build info variables in banner

szantopeter:
This PR will give support of git.properties and build info variables in the banner.txt. This is useful so even command line applications can display the informations that would be otherwise only available through actuator.

### DIFF
--- a/spring-boot/src/main/java/org/springframework/boot/ResourceBanner.java
+++ b/spring-boot/src/main/java/org/springframework/boot/ResourceBanner.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot;
 
+import java.io.IOException;
 import java.io.PrintStream;
 import java.nio.charset.Charset;
 import java.util.ArrayList;
@@ -23,6 +24,7 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Properties;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -31,9 +33,11 @@ import org.springframework.boot.ansi.AnsiPropertySource;
 import org.springframework.core.env.Environment;
 import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.MutablePropertySources;
+import org.springframework.core.env.PropertiesPropertySource;
 import org.springframework.core.env.PropertyResolver;
 import org.springframework.core.env.PropertySourcesPropertyResolver;
 import org.springframework.core.io.Resource;
+import org.springframework.core.io.support.PropertiesLoaderUtils;
 import org.springframework.util.Assert;
 import org.springframework.util.StreamUtils;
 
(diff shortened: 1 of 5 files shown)


spring-boot/src/main/java/org/springframework/boot/ResourceBanner.java
```diff
@@ -48,21 +52,25 @@
 
 	private static final Log logger = LogFactory.getLog(ResourceBanner.class);
 
-	private Resource resource;
+	private Resource bannerResource;
+	private Resource gitPropertiesResource;
+	private Resource buildInfoPropertiesResource;
+
+	public ResourceBanner(Resource bannerResource, Resource gitPropertiesResource, Resource buildInfoPropertiesResource) {
+		Assert.notNull(bannerResource, "Resource must not be null");
```
shakuzen: The message should probably specify that the **banner** resource must not be null, since the other two resources can be null.

spring-boot/src/main/java/org/springframework/boot/ResourceBanner.java
```diff
@@ -48,21 +52,25 @@
...
+		Assert.isTrue(bannerResource.exists(), "Resource must exist");
```
shakuzen: Again, best to clarify the resource in the message since there are multiple resources now.

spring-boot/src/main/java/org/springframework/boot/ResourceBanner.java
```diff
@@ -48,21 +52,25 @@
 
 	private static final Log logger = LogFactory.getLog(ResourceBanner.class);
 
-	private Resource resource;
+	private Resource bannerResource;
+	private Resource gitPropertiesResource;
+	private Resource buildInfoPropertiesResource;
+
+	public ResourceBanner(Resource bannerResource, Resource gitPropertiesResource, Resource buildInfoPropertiesResource) {
```
spencergibb: Since it's a public class with new optional fields, keep the original Constructor and delegate to this one.

szantopeter:
I applied the review comments. I am not an ascii doctor expert and couldn’t create a link that points to a section within an another page. If you know let me know.

wilkinsona:
Thanks for the PR. This feels like quite a bit of extra complexity for what is probably a niche benefit. Taking the extra `Resource` instances as constructor arguments is also problematic. Having three arguments all with the same type makes it easy to mix them up. It's also going to get unwieldy if we ever wanted to add any more.

I think this would be better tackled as a custom banner in your own project. That banner could be implemented as a `ResourceBanner` subclass. If `ResourceBanner` currently lacks the necessary extension points, I'd be more in favour of adding those than the changes proposed here.

Let's see what the rest of the team think.

szantopeter:
@wilkinsona Obviously it is not my decision, but short living tasks without REST (or any other) interface is not a niche, e.g.: Spring Cloud Data Flow. What other ways have these applications to log the git and build information? If the only problem is the constructor complexity then it can be easily solved with a builder style solution like this :

    ResourceBanner.builder().withBanner(b).withGitInfo(g).withBuildInfo(b).build()

What do you think?

wilkinsona:
I don't think that logging the information via the banner is the best way to make it available. Personally, I'd use the Actuator's JMX support and the existing info endpoint. I still believe that applications where the Actuator isn't available in any form are a niche.

A builder is an improvement from an API perspective, but it doesn't do much to reduce the additional internal complexity that I am still not convinced is justified.

szantopeter:
I hope Spring Batch, Spring Cloud Data Flow, and writing Big Data jobs with Spring Boot are not niche ;)

wilkinsona:
> I hope Spring Batch, Spring Cloud Data Flow, and writing Big Data jobs with Spring Boot are not niche

That's not what I said. None of the things that you have listed preclude you from using the Actuator and accessing it via JMX.

If you're looking for the information to be logged so that it's pushed rather than being pulled or so that it's available even when the process is no longer running, that's potentially of benefit to any type of application. However, I still don't think that the banner is the right place for it. I'd rather provide an option in the Actuator's info support for it to log all of the known contributions.

szantopeter:
I see your point, the question is how can we reuse Actuator code in an application that doesn't have web, nor actuator dependency. Or should we make actuator work even for non web applications?

wilkinsona:
> Or should we make actuator work even for non web applications?

It already works in non-web applications. If you add a dependency on `spring-boot-starter-actuator` its endpoints are [automatically available via JMX](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-jmx).

szantopeter:
ok, so what would be the ideal solution for short living applications to display git and build information? Add an option to Actuator to print information to the log?

wilkinsona:
> Add an option to Actuator to print information to the log?

Possibly, yes. Please open an issue describing your end goal and we can then figure out the best way to meet it.

philwebb:
Thanks for the PR, but I agree with the Andy that the added complexity to support this in the banner isn't ideal.

szantopeter:
ok I will open an issue, I am happy to add this feature to actuator, unless you want to do it yourself

szantopeter:
I opened this issue https://github.com/spring-projects/spring-boot/issues/11950

PR #11918, opened 2018-02-06
Standardized basic cache metrics

jkschneider:
In https://github.com/micrometer-metrics/micrometer/issues/393, a `CacheMetricsBinder` abstract class was introduced and cache binder implementations were made to extend it. Cache metrics are now uniformly prefixed with `cache.` to allow property-based configuration of cache metrics independent of the cache implementation.

The purpose is to standardize basic cache metrics like gets (hit and miss), puts, evictions, and size so from the perspective of monitoring there is some cache implementation independence.

This resulted in a handful of constructor changes which are reflected in this PR.


wilkinsona:
This PR seems to be incomplete. Two tests are failing and the `management.metrics.cache.metric-name` property now affects the value of a tag rather than the name of a metric. I suspect it's actually now redundant and can be removed.

Issue #11939, opened 2018-02-07
Conditionally auto-configure a ResourceConfig bean for use by the Jersey-based endpoints


wilkinsona:
As things stand, the Jersey-based endpoints are only available once the user has created their own `ResourceConfig` bean. We could auto-configure a ResourceConfig bean when the user hasn't provided their own so that the Jersey endpoints are available before the user's started setting up their JAX-RS resources.

/cc @michael-simons

wilkinsona:
Closing in favour of the PR (#11948).

PR #11948, opened 2018-02-07
Auto-configure a ResourceConfig for Jersey endpoints if needed

michael-simons:
Closes gh-11939


flangknecht:
Why not go a step further and auto-register spring-managed resources (and by that I mean any component annotated with javax.ws.rs.Path) by class?
This would allow us to stop having to manually enumerate JAX-RS resources and further eliminate boiler plate code from Spring-Boot projects using Jersey.

wilkinsona:
@flangknecht That's largely unrelated to this pull request. Please open a separate issue and we can consider what you've proposed in due course.

Issue #8029, opened 2017-01-19
Improve FailureAnalyzer for embedded datasource


snicoll:
The `FailureAnalyzer` for the embedded datasource is too focused on the embedded use case (that is the fallback to a proper configuration to a production database). 

Say that you start an app with JPA and MySQL, this is what you get

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Cannot determine embedded database driver class for database type NONE

Action:

If you want an embedded database please put a supported one on the classpath. If you have database settings to be loaded from a particular profile you may need to active it (no profiles are currently active).
```

The link between MySQL and the embedded database isn't obvious at all actually. The action mentions "database settings" but the description doesn't. Perhaps we could be smarter and look in the environment? 

If that exception is thrown, it means we've tried to auto-configure the `DataSource`. We can therefore safely assume that no `spring.datasource.url` was specified. Perhaps we should mention that in the description? 

Actually, describing that we've tried to create an embedded database because no settings to configuration the database were found would be a good first start IMO.

snicoll:
Closing in favour of PR #11953

ff-yzheng:
I am very new about Spring, but I got the following error:
***************************
APPLICATION FAILED TO START
***************************

Description:

Failed to auto-configure a DataSource: 'spring.datasource.url' is not specified and no embedded datasource could be auto-configured.

Reason: Failed to determine a suitable driver class


is there any workaround here , like put configuration in application.properieties? Thank you for your help.

snicoll:
@ff-yzheng this issue is closed and we don't use the tracker for questions. Feel free to ask on StackOverflow or come chat with us [on Gitter](https://gitter.im/spring-projects/spring-boot)

Issue #11569, opened 2018-01-08
FailureAnalyzer should be able to inject the environment


snicoll:
If a `FailureAnalyzer` needs to inspect the `Environment` it would be nice to be able to inject it.

This is a prerequisite of #10794

PR #11953, opened 2018-02-07
Improve FailureAnalyzer for embedded datasource

pkostrzewa:
Fixes gh-8029.

* Moved and refactored logic related to failure analysis from exception to `DataSourceBeanCreationFailureAnalyzer`.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceBeanCreationFailureAnalyzer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceBeanCreationFailureAnalyzer.java
@@ -21,6 +21,8 @@ import java.util.Objects;
 import org.springframework.boot.autoconfigure.jdbc.DataSourceProperties.DataSourceBeanCreationException;
 import org.springframework.boot.diagnostics.AbstractFailureAnalyzer;
 import org.springframework.boot.diagnostics.FailureAnalysis;
+import org.springframework.boot.jdbc.EmbeddedDatabaseConnection;
+import org.springframework.context.EnvironmentAware;
 import org.springframework.core.env.Environment;
 import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
```diff
@@ -520,39 +520,39 @@ public void setProperties(Map<String, String> properties) {
 
 	}
 
-	static class DataSourceBeanCreationException extends BeanCreationException {
+	class DataSourceBeanCreationException extends BeanCreationException {
```
pkostrzewa: I'm not sure whether I should remove `static` modifier. Maybe it would be even better to move this class outside of `DataSourceProperties`?

pkostrzewa:
@snicoll any feedback would be appreciated :)

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceBeanCreationFailureAnalyzer.java
```diff
@@ -16,26 +16,68 @@
...
 	protected FailureAnalysis analyze(Throwable rootFailure,
 			DataSourceBeanCreationException cause) {
-		String message = cause.getMessage();
-		String description = message.substring(0, message.indexOf('.')).trim();
-		String action = message.substring(message.indexOf('.') + 1).trim();
-		return new FailureAnalysis(description, action, cause);
+		return getFailureAnalysis(createDescription(cause), cause);
 	}
 
+	private String createDescription(DataSourceBeanCreationException cause) {
```
snicoll: I don't think that's what I had in mind. What we need to do here is check if "NONE" is active and change the description accordingly.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
```diff
@@ -522,37 +522,31 @@ public void setProperties(Map<String, String> properties) {
...
-				}
-			}
-			message.append(".");
-			return message.toString();
+		public EmbeddedDatabaseConnection getConnection() {
+			return this.connection;
 		}
 
-	}
+		public Environment getEnvironment() {
```
snicoll: You can't expose such thing in an exception.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
```diff
@@ -522,37 +522,31 @@ public void setProperties(Map<String, String> properties) {
...
-					message.append(" (the profiles \""
-							+ StringUtils.arrayToCommaDelimitedString(
-									environment.getActiveProfiles())
-							+ "\" are currently active)");
-
-				}
-			}
-			message.append(".");
-			return message.toString();
+		public EmbeddedDatabaseConnection getConnection() {
```
snicoll: Idem

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceProperties.java
```diff
@@ -522,37 +522,31 @@ public void setProperties(Map<String, String> properties) {
...
+
+		private final Environment environment;
+
+		private final String property;
+
 		DataSourceBeanCreationException(EmbeddedDatabaseConnection connection,
 				Environment environment, String property) {
-			super(getMessage(connection, environment, property));
+
+			super("Cannot auto-configure DataSource. ");
```
snicoll: If you want to pursue this, the message should be provided by the caller, you  can't hard-code it like that.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceBeanCreationFailureAnalyzer.java
pkostrzewa: You mean that such detailed description should be provided only in case when connection stands for "NONE"?

snicoll: No that's not what I meant. I think we need to take a step back. Can you  rework your PR to change the exception to expose the `EmbeddedDatabaseConnection`? Then you can adapt the message when it is NONE. The change to fix that issue must  not be that large. Let me know if that works for you.

pkostrzewa: It's ok for me. Actually I've reworked it similarly .

pkostrzewa:
There are two options:
- expose `connection` from `DataSourceBeanCreationException` (https://github.com/spring-projects/spring-boot/pull/11953/files/ec9cfa405dd716e97e5c54d0adb23e7b78f7c72c#r166909465 - you pointed out that might not be good idea to expose such information) and then build proper messages in failure analyzer depending on whether `connection` is `NONE` i
- expose only `description` and `action` already created inside of exception and don't change them in failure analyzer (hides details about exception)

Which one do you prefer @snicoll ?

snicoll:
@pkostrzewa Tanks. I think that  we need  to revert the logic and build the message in the failure analyzer as you've done initially. But I'd like the changes to be a bit more focused. Here is what you can do:

* You can inject the `Environment` in a `FailureAnalyzer` if you implement `EnvironmentAware` (#11569)
* We can expose the `EmbeddedDatabaseConnection` in  the exception as  a getter
* We can simplify the message to use the "property" part. We can use that  then in the failure analyzer as a support to build the description message, the two message can be "Failed to determine a suitable driver class" and "Failed  to determine a suitable jdbc url". We  can build those message as is in the  caller rather than computing them in the exception

So our  exception takes now a message (see above) and an `EmbeddedDatabaseConnection`. Once you have that, we can craft a better error message. And we can have a special case for NONE to indicate that no embedded database was found on the classpath.

pkostrzewa:
* `EmbeddedDatabaseConnection` exposed as a getter in exception
* `DataSourceBeanCreationFailureAnalyzer` implements `EnvironmentAware` so that current environment can be accessed
* `DataSourceBeanCreationException` takes an `connection` and a `message`
* Available actions are computed based on whether the connection type is `EmbeddedDatabaseConnection.NONE`

pkostrzewa:
@snicoll any feedback?

snicoll:
@pkostrzewa this is better but not quite there yet. Given the back and forth, I've polished your last proposal, see 2f13449

PR #11967, opened 2018-02-08
StatsD configuration option to publish unchanged pollable meters

jkschneider:
See https://github.com/micrometer-metrics/micrometer/pull/406.

Etsy statsd and Datadog's dogstatsd both drop unchanged gauges after a period of time. This option publishes an "update" to gauges on every poll, even if the gauge is unchanged. It is on by default.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdProperties.java
@@ -59,7 +59,8 @@ public class StatsdProperties {
 
 	/**
 	 * How often gauges will be polled. When a gauge is polled, its value is recalculated
-	 * and if the value has changed, it is sent to the StatsD server.
+	 * and if the value has changed (or publishUnchangedMeters is true),
+	 * it is sent to the StatsD server.
 	 */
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdProperties.java
```diff
@@ -59,7 +59,8 @@
 
 	/**
 	 * How often gauges will be polled. When a gauge is polled, its value is recalculated
-	 * and if the value has changed, it is sent to the StatsD server.
+	 * and if the value has changed (or {@link #publishUnchangedMeters} is true),
```
snicoll: You can't use javadoc tags in here. This is used as the description in IDEs and other metadata clients.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdProperties.java
```diff
@@ -68,6 +69,11 @@
 	 */
 	private Integer queueSize = Integer.MAX_VALUE;
 
+	/**
+	 * Enables or disables sending of unchanged meters to StatsD server
```
snicoll: We're trying to make sure descriptions are harmonized. Please look at other boolean flag descriptions in the same area.

Issue #11958, opened 2018-02-08
ValidationQuery for MySQL connection should start with /* ping */


chjlang:
I found that the default value of validationQuery of MySQL driver is ```SELECT 1```, see [DatabaseDriver](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java#L70).

But according to the [documentation](https://dev.mysql.com/doc/connector-j/5.1/en/connector-j-usagenotes-j2ee-concepts-connection-pooling.html) of MySQL Connector/J , the validation query should start with ```/* ping */```, which makes a lightweight ping to the server. In addition to efficiency, it also ensures each underlying physical is valid when using a ReplicationConnection or LoadBalancedConnection since the ping will be sent across all active connections. Otherwise, the query is executed on just one active connection and the other non-active physical connections are in a stale state. When re-balance happens, it might select a dead connection, resulting in an exception.

philwebb:
Interesting find. Thanks for reporting it.

eiselems:
Selected wrong target branch at first - sorry.
Think it should be also done for 2.x.x - should I create another PR for master referring to this issue?

philwebb:
@eiselems It looks like #11981 is against 1.5.x so that's perfect. We'll merge to 1.5.x then it'll also get forward merged to master.

snicoll:
Closing in favour of PR #11981

chjlang:
Thanks for the rapid response and PR

PR #11981, opened 2018-02-09
Update MySQL validation query to use lightweight ping

eiselems:
Fixes gh-11958

Checked it also for Maria since often they are quite similiar.
But was rejected there https://jira.mariadb.org/browse/CONJ-51 .


snicoll:
@eiselems thank you so much for you first contribution to Spring Boot. This is now merged in `1.5.x`  and `master`.

PR #12000, opened 2018-02-12
Replace Narayana JDBC with Commons DBCP2

gytis:
Replace Narayana JDBC driver with commons-dbcp2 and commons-pool2 to take advantage of a connection pooling support when using Narayana.
We would like to introduce this change to 1.5.x branch too.

cc @zhfeng and @tomjenkinson from Narayana project.


wilkinsona:
Thanks for the PR.

> We would like to introduce this change to 1.5.x branch too.

Unfortunately, I think this change is too big to go into 1.5. Previously, we'd talked about the Narayana auto-configuration moving out of Boot. Perhaps this change is further motivation for that move? There's probably just about time to do that in 2.0.

gytis:
Thanks for your feedback @wilkinsona. Since we need this feature to work with both versions, I'll go ahead and start creating a separate project for Narayana starter.

PR #12008, opened 2018-02-12
Upgrade to liquibase 3.5.4

candrews:
https://github.com/liquibase/liquibase/releases/tag/liquibase-parent-3.5.4


philwebb:
Thanks for the PR, but we have an automated process to pick up dependency upgrades that should do this one for us.

PR #12018, opened 2018-02-13
Allow customization of ConcurrentKafkaListenerContainerFactory Fixes gh-12003

### LINKED ISSUE #12003: Allow extension of ConcurrentKafkaListenerContainerFactoryConfigurer
I'm trying to extend the Spring Boot Kafka's auto-configuration (e.g. handling `spring.kafka.*` properties) by some custom settings, such as:
* my own `RetryTemplate`
* my own `ErrorHandler`

Ideally, such snippet would be possible:

```java
public class CustomKafkaConfigurer extends ConcurrentKafkaListenerContainerFactoryConfigurer {

    private final ErrorHandler errorHandler;

    private final RetryTemplate retryTemplate;

    public CustomKafkaConfigurer(KafkaProperties properties, ErrorHandler errorHandler, RetryTemplate retryTemplate) {
        setKafkaProperties(properties);
        this.errorHandler = errorHandler;
        this.retryTemplate = retryTemplate;
    }

    public void configure(ConcurrentKafkaListenerContainerFactory<Object, Object> listenerContainerFactory, ConsumerFactory<Object, Object> consumerFactory) {
        super.configure(listenerContainerFactory, consumerFactory);
        listenerContainerFactory.setRetryTemplate(retryTemplate);
        listenerContainerFactory.getContainerProperties().setErrorHandler(errorHandler);
    }

}
```

However, the `setKafkaProperties` method is package-private, which blocks the scenario above (unless one wants to make some dirty hacks).

My suggestions (choose one):
* change the setter method scope to `public`,
* create a `ConcurrentKafkaListenerContainerFactoryConfigurer` interface, the current class being one of its implementations (as suggested here: https://stackoverflow.com/questions/47862458/how-to-set-a-custom-consumer-task-executor-for-spring-kafka-when-using-auto-con#comment82773807_47870184),
* add some other plug-in point so that one can extend the auto-configuration (no necessarily modifying it).

pkostrzewa:
Simple implementation of `Customizer` that takes an `ConcurrentKafkaListenerContainerFactory` and allows customization whilst retaining default auto-configuration.

Fixes #12003

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
@@ -72,7 +72,7 @@ public class ConcurrentKafkaListenerContainerFactoryConfigurer {
 
 	/**
 	 * Add the {@link ConcurrentKafkaListenerContainerFactoryCustomizer}
-	 * customizer to use.`
+	 * customizer to use.
 	 * @param customizer the customizer
 	 */
 	public void addCustomizer(


pkostrzewa:
I'm not sure about multiple `customizers`. Should there be possibility to add more than one?

philwebb:
@pkostrzewa You, usually the pattern is that you can have multiple customizers.

pkostrzewa:
@philwebb I've no idea why build has failed again - I fixed checkstyle errors. Could you check please?

pkostrzewa:
Any feedback?

philwebb:
@pkostrzewa The CI issue was probably just a glitch, we can double check when we merge this.

snicoll:
I had a look to this PR previously and forgot to comment my thoughts on it, sorry about that. I don't think we can merge this PR (ignoring the lack of tests) because the customizer is applied regardless of the `ConcurrentKafkaListenerContainerFactory` that you are configuring. 

Spring Boot will configure a default `kafkaListenerContainerFactory` that you can use as a default for `@KafkaListener`. If, down the road, you need to configure another instance and flag it in an endpoint, e.g. `@KafkaListener(containerFactory = "myFactory")`, the customizer will apply as well. 

You may or may not want such customizer to be applied for those cases so I think we should take that into account. Getting back to the use case, this is what happens when a new factory is created with the configurer:

```java
@Bean
ublic ConcurrentKafkaListenerContainerFactory<?, ?> myKafkaListenerContainerFactory(
		ConcurrentKafkaListenerContainerFactoryConfigurer configurer,
		ConsumerFactory<Object, Object> kafkaConsumerFactory) {
	ConcurrentKafkaListenerContainerFactory<Object, Object> factory = new ConcurrentKafkaListenerContainerFactory<>();
	configurer.configure(factory, kafkaConsumerFactory);
	return factory;
}
```

So I am wondering why a customizer is needed in the first place as you can do:

```java
@Bean
ublic ConcurrentKafkaListenerContainerFactory<?, ?> myKafkaListenerContainerFactory(
		ConcurrentKafkaListenerContainerFactoryConfigurer configurer,
		ConsumerFactory<Object, Object> kafkaConsumerFactory) {
	ConcurrentKafkaListenerContainerFactory<Object, Object> factory = new ConcurrentKafkaListenerContainerFactory<>();
	configurer.configure(factory, kafkaConsumerFactory);
	factory.setReplyTemplate(myRetryTemplate);
	factory.getContainerProperties().setErrorHandler(myErrorHandler);
	return factory;
}
```

If you want to override the default, you can do so by specifying its bean name, e.g.

```java
@Bean(name = "kafkaListenerContainerFactory")
public ConcurrentKafkaListenerContainerFactory<?, ?> kafkaListenerContainerFactory(...) { ... }
```

@maciekpapiez what am I missing here? IMO, that flexibility is already there and doesn't require to introduce yet another concept.

maciekpapiez:
@snicoll We've just tried adopting your approach and it seems to be perfectly fine. I find it less obvious than an exposed `Customizer` interface, but it satisfies our requirements :) Thank you for the hint!

snicoll:
Thanks for feedback @maciejwalkowiak, that confirms that we shouldn't merge this.

PR #12021, opened 2018-02-13
Server compression breaks webflux + actuator

alecharp:
**Disclamer**
This is only a demonstration of a problem. 

---

When `server.compression.enable=true` is configured, even when request the response as `application/json`, the actuator responses end up in timeout. 

In Chrome, it is possible to see the answer, but the request is not closed by the application.

The code here is a additional test in the smallest sample I could find with webflux and actuator to see if I could reproduce the problem, and I could.

### DIFF
--- a/spring-boot-samples/spring-boot-sample-webflux/src/test/java/sample/webflux/CompressedWebFluxApplicationTests.java
+++ b/spring-boot-samples/spring-boot-sample-webflux/src/test/java/sample/webflux/CompressedWebFluxApplicationTests.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@alecharp Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12021)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12021) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@alecharp Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12021)!

bclozel:
Thanks for the report!
Closing in favor of #12025, since this isn't something we can fix in Spring Boot itself.

alecharp:
Thanks for the investigation and your reply. Surprise to be the first one to hit this.

PR #12032, opened 2018-02-13
Polish: "public static" fields should be constant, to get rid string duplication

igor-suhorukov:


philwebb:
Thanks!

igor-suhorukov:
@philwebb thank you

PR #12066, opened 2018-02-14
Add SignalFxConfig source property

jkschneider:
See https://github.com/micrometer-metrics/micrometer/issues/428

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxProperties.java
@@ -39,6 +39,11 @@ public class SignalFxProperties extends StepRegistryProperties {
 	 */
 	private String uri;
 
+	/**
+	 * Uniquely identifies the app instance that is publishing metrics to SignalFx. Defaults to the local host name.
+	 */
+	private String source;
+
 	public String getAccessToken() {
 		return this.accessToken;
 	}
@@ -53,6 +58,14 @@ public class SignalFxProperties extends StepRegistryProperties {
 
 	public void setUri(String uri) {
 		this.uri = uri;
+	}
+
+	public String getSource() {
+		return source;
+	}
+
+	public void setSource(String source) {
+		this.source = source;
 	}
 
 }
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxPropertiesConfigAdapter.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxPropertiesConfigAdapter.java
@@ -44,4 +44,9 @@ public class SignalFxPropertiesConfigAdapter
 		return get(SignalFxProperties::getUri, SignalFxConfig.super::uri);
 	}
 
+	@Override
+	public String source() {
+		return get(SignalFxProperties::getSource, SignalFxConfig.super::source);
+	}
+
 }


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxProperties.java
```diff
@@ -39,6 +39,11 @@
 	 */
 	private String uri;
 
+	/**
+	 * Uniquely identifies the app instance that is publishing metrics to SignalFx. Defaults to the local host name.
```
snicoll: I don't like those unpredictable defaults (we have a `default` concept in the metadata that is meant for that and I'd rather use it than putting that information in the description).

Is there anyway we can relax this?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxPropertiesConfigAdapter.java
```diff
@@ -44,4 +44,8 @@ public String uri() {
 		return get(SignalFxProperties::getUri, SignalFxConfig.super::uri);
 	}
 
+	@Override
+	public String source() {
+		return get(SignalFxProperties::getSource, SignalFxConfig.super::source);
+	}
```
snicoll: missing new line

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxProperties.java
```diff
@@ -55,4 +60,11 @@ public void setUri(String uri) {
 		this.uri = uri;
 	}
 
+	public String getSource() {
+		return source;
+	}
+
+	public void setSource(String source) {
+		this.source = source;
+	}
```
snicoll: missing newline

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxProperties.java
wilkinsona: There are other pieces of the portfolio that have a similar requirement. It'd be good to consider what they currently do and align if possible. Spring Cloud Bus is the project that comes to mind as it uses a unique ID for message routing. Can you give us some pointers please, @dsyer?

jkschneider: I asked @spencergibb about this as well, and I believe we use host name in discovery.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxPropertiesConfigAdapter.java
jkschneider: Fixed in 2592cb8

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxProperties.java
jkschneider: Fixed in 2592cb8

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxProperties.java
wilkinsona: Spring Cloud Bus uses the following:

```
"${vcap.application.name:${spring.application.name:${spring.application.name:application}}}:${vcap.application.instance_index:${spring.application.index:${local.server.port:${server.port:0}}}}:${vcap.application.instance_id:${random.value}}"
```

Note the use of `random.value` at the end. I think that makes it unsuitable for this purpose as the id will potentially change across restarts.

dsyer: The Bus id is structured as `{app/host}:{port/index}:{guid}`. If Spring Boot was supplying the first two we could add the “random” bit. Or maybe the whole thing would be useful for micrometer (don’t know).

jkschneider: @wilkinsona I wouldn't worry about change across restart. Every scaling event/deployment in AWS results in a new instance id, and we tagged metrics with those instance ids.

wilkinsona: It's too late in the day to be doing anything sophisticated here. We'll stick with Micrometer's default for 2.0 rather than trying to be clever and override it with something else and probably getting it wrong.

jkschneider: Agreed. It also doesn't really matter that much. The "source" bit that some of these monitoring systems require almost seem to be a cherry pick from the days of hierarchical metrics. Like the Bus id reveals: identification of a unique app instance is a multi-variate problem best solved with tags.

PR #12068, opened 2018-02-15
Add Wavefront support

jkschneider:
See https://github.com/micrometer-metrics/micrometer/issues/365.

Adds support for VMWare's dimensional monitoring system, Wavefront.

cc / @howardyoo

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.autoconfigure.metrics.export.wavefront;
+
+import io.micrometer.core.instrument.Clock;
+import io.micrometer.wavefront.WavefrontConfig;
+import io.micrometer.wavefront.WavefrontMeterRegistry;
+
+import org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration;
+import org.springframework.boot.actuate.autoconfigure.metrics.export.simple.SimpleMetricsExportAutoConfiguration;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for exporting metrics to Wavefront.
+ *
+ * @author Jon Schneider
+ * @since 2.0.0
+ */
+@Configuration
+@AutoConfigureBefore(SimpleMetricsExportAutoConfiguration.class)
+@AutoConfigureAfter(MetricsAutoConfiguration.class)
+@ConditionalOnBean(Clock.class)
+@ConditionalOnClass(WavefrontMeterRegistry.class)
+@ConditionalOnProperty(prefix = "management.metrics.export.wavefront", name = "enabled", havingValue = "true", matchIfMissing = true)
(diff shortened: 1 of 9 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/signalfx/SignalFxMetricsExportAutoConfigurationTests.java
```diff
@@ -50,7 +49,7 @@
 	@Test
 	public void backsOffWithoutAClock() {
 		this.runner.run((context) -> assertThat(context)
-				.doesNotHaveBean(InfluxMeterRegistry.class));
+				.doesNotHaveBean(SignalFxMeterRegistry.class));
```
jkschneider: Nit on similar SignalFx test that was referring to the wrong type.

wilkinsona: Oops. Thanks for the catch.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontProperties.java
```diff
@@ -0,0 +1,88 @@
...
+public class WavefrontProperties extends StepRegistryProperties {
+	/**
+	 * The URI to publish metrics to. The URI could represent a Wavefront sidecar or the
+	 * Wavefront API host. This host could also represent an internal proxy set up in your environment
+	 * that forwards metrics data to the Wavefront API host.
+	 *
+	 * If publishing metrics to a Wavefront proxy (as described in https://docs.wavefront.com/proxies_installing.html),
+	 * the host must be in the proxy://HOST:PORT format.
+	 */
+	private String uri;
```
wilkinsona: Could this be a `URI`? The same question applies to Atlas, Influx, New Relic and SignalFX but I missed those previously.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
```diff
@@ -0,0 +1,63 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for exporting metrics to StatsD.
```
wilkinsona: `StatsD` -> `Wavefront`

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontProperties.java
```diff
@@ -0,0 +1,88 @@
...
+
+	/**
+	 * Uniquely identifies the app instance that is publishing metrics to Wavefront. Defaults to the local host name.
+	 */
+	private String source;
```
wilkinsona: Whatever we decide to do with SignalFX's source, we should do the same thing here.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontProperties.java
```diff
@@ -0,0 +1,88 @@
...
+	 * Required when publishing directly to the Wavefront API host, otherwise does nothing.
+	 */
+	private String apiToken;
+
+	/**
+	 * Wavefront metrics are grouped hierarchically by name in the UI. Setting a global prefix separates
+	 * metrics originating from this app's whitebox instrumentation from those originating from other Wavefront
+	 * integrations.
+	 */
+	private String globalPrefix;
```
wilkinsona: The first sentence in the javadoc is what's used in the metadata. To improve what's shown in an IDE, I think I'd reword this to something like the following:

> Global prefix to separate metrics originating from this app's white box instrumentation from those originating from other Wavefront integrations when viewed in the Wavefront UI.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontProperties.java
jkschneider: I tested this, and yes, `URI` would work fine. I could change the type on the underlying interfaces (`InfluxConfig`, `NewRelicConfig`, etc.) for all _but_ `AtlasConfig`, which is a type that we don't control. So I think for consistency I'll leave them as String in Micrometer, but there should be no problem mapping a URI in Spring props to a String on the corresponding Micrometer interface.

I could change the PR to URI here or we can merge this as a String and I can follow up with a single PR that changes all of them to URI. Whichever you prefer.

wilkinsona: Let's go with URI for this one, and then make all the others use `URI` in a separate commit/PR.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontProperties.java
jkschneider: We decided to leave it as is I think.

Issue #12089, opened 2018-02-16
Inferior user experience with Micrometer-related properties


snicoll:
A bunch of micrometer properties have a sensitive default but they don't show up in the metadata. There should be an easy way to check that our hardcoded default matches the one in micrometer and fail the build when that's no longer the case.

snicoll:
@jkschneider @philwebb I've started to restore defaults from Micrometer in our infrastructure so that the IDEs show default values. I have an infra in place to make sure our defaults are consistent with the ones of Micrometer.

Doing so I've noticed that Spring Boot [disables the use of the `SimpleRegistry`](https://github.com/snicoll/spring-boot/commit/3a53c1d232a42779c3a3f1c022ea4051df27ede9#diff-7882fbd3a277da740b07019183dd2a16R40). Is that intended or is it an oversight? If we want to deviate from Micrometer's default there, that's fine but I'd like to make sure it is on purpose.

jkschneider:
> Doing so I've noticed that Spring Boot disables the use of the SimpleRegistry. Is that intended or is it an oversight?

Boot disables simple registry in the presence of a real one, which is the intended behavior. Since the simple registry doesn't publish anywhere, enablement doesn't even mean anything in plain Micrometer.

snicoll:
@jkschneider I am confused. If Spring Boot disables it if a real one exists, then the default value shouldn't be false isn't it? What does "real one" means? What is the user experience of using the simple one? 

>  enablement doesn't even mean anything in plain Micrometer.

Why is there a property then? Why is it true by default in micrometer?

wilkinsona:
I suspect that’s a side-effect of SimpleConfig extending StepRegistryConfig. One way or another, something’s not quite right as our SimpleProperties does not extend StepRegistryProperties so there appears to be some asymmetry there.

jkschneider:
At some point when all the individual implementation configurations were changed to auto-configurations in Boot 2, the conditionalization of the config based on `managament.metrics.export.simple.enabled` was dropped. It existed previously so that users could flip off even in-memory metrics collection, leaving an empty `CompositeMeterRegistry` that was essentially a NOOP.

> What does "real one" means?

Any implementation that actually ships metrics to a monitoring system - Datadog, Prometheus, Atlas, etc.

> What is the user experience of using the simple one?

[...]

PR #12106, opened 2018-02-18
Restore behavior of management.metrics.export.simple.enabled

jkschneider:
At some point when all the individual implementation configurations were changed to auto-configurations in Boot 2, the conditionalization of the config based on `managament.metrics.export.simple.enabled` was dropped. It existed previously so that users could flip off even in-memory metrics collection, leaving an empty CompositeMeterRegistry that was essentially a NOOP (see `CompositeMeterRegistryPostProcessorTests#registerWhenHasNoMeterRegistryShouldRegisterEmptyComposite`).

See https://github.com/spring-projects/spring-boot/issues/12089#issuecomment-366524703

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/simple/SimpleProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/simple/SimpleProperties.java
@@ -34,7 +34,8 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
 public class SimpleProperties {
 
 	/**
-	 * Enable publishing to the backend.
+	 * Enable in-memory metrics that aren't published anywhere (allows you to see
+	 * what metrics are collected in the metrics actuator endpoint).
 	 */
 	private boolean enabled;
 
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/simple/SimpleProperties.java
```diff
@@ -34,7 +34,8 @@
 public class SimpleProperties {
 
 	/**
-	 * Enable publishing to the backend.
+	 * Enable in-memory metrics that aren't published anywhere (allows you to see
+	 * what metrics are collected in the metrics actuator endpoint).
 	 */
 	private boolean enabled;
```
snicoll: The default value here should be true. And we shouldn't have a property at all since it's not used anywhere in the code.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/simple/SimpleProperties.java
```diff
@@ -34,7 +34,8 @@
 public class SimpleProperties {
 
 	/**
-	 * Enable publishing to the backend.
+	 * Enable in-memory metrics that aren't published anywhere (allows you to see
+	 * what metrics are collected in the metrics actuator endpoint).
```
snicoll: I think that's a bit misleading and we should rephrase that.

PR #12117, opened 2018-02-19
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/servlet/PathRequest.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/servlet/PathRequest.java
@@ -38,7 +38,7 @@ public final class PathRequest {
 
 	/**
 	 * Returns a {@link StaticResourceRequest} that can be used to create a matcher for
-	 * {@link StaticResourceLocation Locations}.
+	 * {@link StaticResourceLocation locations}.
 	 * @return a {@link StaticResourceRequest}
 	 */
 	public static StaticResourceRequest toStaticResources() {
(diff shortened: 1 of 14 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/servlet/PathRequest.java
```diff
@@ -38,7 +38,7 @@ private PathRequest() {
 
 	/**
 	 * Returns a {@link StaticResourceRequest} that can be used to create a matcher for
-	 * {@link StaticResourceLocation Locations}.
+	 * {@link StaticResourceLocation}s.
```
snicoll: There are many more places where that pattern is used.  In `StaticResourceRequest` for instance. I don't like that `s` being out of the link. I think this is a bit odd when you look at the generated Javadoc.

Thoughts?

izeye: @snicoll I thought "StaticResourceLocations" could mislead to think there is `StaticResourceLocations` class but I admit "`StaticResourceLocation`s" doesn't look good, either.

I don't have a strong opinion on this, so if you prefer `{@link StaticResourceLocation StaticResourceLocations}`, I can change them. Just let me know.

snicoll: Yes please. Using that rather than
locations looks sensible to me.

izeye: @snicoll Thanks for the feedback! I pushed the change as suggested.

philwebb:
Fixed in 88b184ca23549b87de282ee10bf6b43d5fa98b06. Thanks.

PR #12127, opened 2018-02-20
Consistent prefix for datasource pool metrics

jkschneider:
This yields a similar naming scheme to that employed by `FileDescriptorMetrics`

See https://github.com/micrometer-metrics/micrometer/issues/449


snicoll:
Approved, well, except the two tests failing 😕

Issue #12090, opened 2018-02-16
Auto-configure JettyMetrics


snicoll:
We have metrics support for Tomcat  and micrometer has [support for Jetty](https://github.com/micrometer-metrics/micrometer/blob/553f20679b163d7d6da3603a1d1e6a4486e06bcd/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/JettyStatisticsMetrics.java) so we should configure that  as well for consistency.

wilkinsona:
I looked at this briefly when I was adding support for Tomcat's metrics. It's not completely straightforward as it relies on Jetty being configured with a `StatisticsHandler` which is isn't by default. I wondered if the Jetty metrics auto-configuration should customise Jetty with a `StatisticsHandler` but didn't manage to convince myself it was the right thing to do.

wilkinsona:
Given the potential complications with `StatisticsHandler`, we decided that this can wait until 2.1.

evenh:
In my mind this sounds pretty useful. If I were to have a look at this, where would one start?

snicoll:
@evenh we've decided to postpone this to 2.1 (it's not a matter of a lack of time working on it). Thanks anyway.

davidkarlsen:
Should this be picked up again? https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/jetty/TimedHandler.java

wilkinsona:
Thanks, @davidkarlsen. Unfortunately, `TimedHandler` makes things more complicated rather than less. We now use Jetty's `StatisticsHandler` when graceful shutdown is enabled and we would not want to use both when Micrometer's on the classpath as well.

@jkschneider @shakuzen What was the motivation for creating `TimedHandler` rather than using Jetty's `StatisticsHandler`?

jkschneider:
@wilkinsona `TimedHandler` is like `StatisticsHandler`, but much higher resolution. Rather than relying on `StatisticsHandler`'s pre-aggregation of request throughput, etc. which are lossy it uses Micrometer timers.

`StatisticsHandler` is like a poor version of WebMvc and Webflux metrics (`http.server.requests`).

`TimedHandler` was designed for projects that are using the Jetty API more directly, like Javalin to get a request-level metrics experience similar to Spring Boot's WebMvc and WebFlux metrics.

[...]

PR #12128, opened 2018-02-20
Add autoconfiguration of Jetty micrometer statistics

evenh:
As suggested in #12090, auto-configuration of Jetty metrics would be an improvement. I've used the existing `TomcatMetricsAutoConfiguration`  as a starting point. I've also extended the `ServerProperties` configuration properties with `server.jetty.publish-statistics` in order for the user to be able toggle the functionality on or off (isolated in a separate commit).

This is my first contribution to the Spring (Boot) codebase, so feedback is very welcome!

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -209,6 +209,16 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.eclipse.jetty</groupId>
+			<artifactId>jetty-server</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>org.eclipse.jetty</groupId>
+			<artifactId>jetty-webapp</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>org.elasticsearch</groupId>
 			<artifactId>elasticsearch</artifactId>
 			<optional>true</optional>
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfiguration.java
@@ -1,0 +1,105 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.autoconfigure.metrics.web.jetty;
+
+
+import java.util.Collections;
+import java.util.List;
+import java.util.stream.Collectors;
+import java.util.stream.Stream;
+
+import io.micrometer.core.instrument.binder.jetty.JettyStatisticsMetrics;
+import org.eclipse.jetty.server.Handler;
+import org.eclipse.jetty.server.Server;
+import org.eclipse.jetty.server.handler.AbstractHandlerContainer;
+import org.eclipse.jetty.server.handler.StatisticsHandler;
+
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication.Type;
+import org.springframework.boot.web.embedded.jetty.ConfigurableJettyWebServerFactory;
+import org.springframework.boot.web.embedded.jetty.JettyReactiveWebServerFactory;
+import org.springframework.boot.web.embedded.jetty.JettyServletWebServerFactory;
+import org.springframework.boot.web.server.WebServerFactoryCustomizer;
+import org.springframework.context.annotation.Bean;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for {@link JettyStatisticsMetrics}.
+ *
+ * @author Even Holthe
+ * @since 2.1.0
+ */
+@ConditionalOnWebApplication
+@ConditionalOnClass({ JettyStatisticsMetrics.class, Server.class })
+public class JettyMetricsAutoConfiguration {
+	private volatile StatisticsHandler statsHandler;
+
+	@Bean
+	@ConditionalOnMissingBean(JettyStatisticsMetrics.class)
+	public JettyStatisticsMetrics jettyMetrics() {
+		return new JettyStatisticsMetrics(this.statsHandler, Collections.emptyList());
+	}
+
+	@Bean
+	@ConditionalOnWebApplication(type = Type.SERVLET)
+	public WebServerFactoryCustomizer<JettyServletWebServerFactory> serverCapturingServletJettyCustomizer() {
+		return this::statisticsCustomizer;
+	}
+
+	@Bean
+	@ConditionalOnWebApplication(type = Type.REACTIVE)
+	public WebServerFactoryCustomizer<JettyReactiveWebServerFactory> serverCapturingReactiveJettyCustomizer() {
+		return this::statisticsCustomizer;
+	}
+
+	private void extractExistingOrCreateStatisticsHandler(Server server) {
+		final Handler existingHandler = server.getHandler();
+		StatisticsHandler activeStatsHandler = null;
+
+		if (existingHandler instanceof StatisticsHandler) {
+			activeStatsHandler = (StatisticsHandler) existingHandler;
+		}
+
+		else if (existingHandler instanceof AbstractHandlerContainer) {
+			AbstractHandlerContainer abstractHandler = (AbstractHandlerContainer) existingHandler;
+			Handler[] existingHandlers = abstractHandler.getChildHandlers();
+
+			final List<StatisticsHandler> existingStatsHandlers = Stream.of(existingHandlers)
+					.filter(StatisticsHandler.class::isInstance)
+					.map(StatisticsHandler.class::cast)
+					.collect(Collectors.toList());
+
+			if (!existingStatsHandlers.isEmpty()) {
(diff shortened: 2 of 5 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
```diff
@@ -188,6 +188,16 @@
 			<artifactId>commons-dbcp2</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
```
wilkinsona: We try to keep dependencies in alphabetical order (by groupId and then by artifactId)

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for {@link JettyStatisticsMetrics}.
+ *
+ * @author Even Holthe
+ * @since 2.1.0
+ */
+@ConditionalOnWebApplication
+@ConditionalOnClass({ JettyStatisticsMetrics.class, Server.class })
+@ConditionalOnProperty(prefix = "server.jetty", name = "publish-statistics", matchIfMissing = true)
```
wilkinsona: I'm not sure that this needs to be conditional on a property. The auto-configuration can be disabled to remove the functionality completely. Alternatively, the property-based filtering can be used.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+	public WebServerFactoryCustomizer<JettyReactiveWebServerFactory> serverCapturingReactiveJettyCustomizer() {
+		return this::statisticsCustomizer;
+	}
+
+	/**
+	 * Add a statistics handler to all Jetty's {@link Connector}s.
+	 *
+	 * @param server An already configured Jetty {@link Server}.
+	 */
+	private void configureStatistics(Server server) {
```
wilkinsona: I don't think we need this method comment. The method could be renamed to `addStatisticsHandlerToConnectors` if the intent isn't clear without it.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -769,6 +769,11 @@ public void setCacheTtl(Duration cacheTtl) {
 		 */
 		private Integer selectors;
 
+		/**
+		 * Whether to publish Jetty connector statistics.
+		 */
+		private boolean publishStatistics = true;
```
wilkinsona: I think we can drop this field and the associated getter and setter.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/ServerPropertiesTests.java
```diff
@@ -165,6 +165,12 @@ public void testCustomizeJettySelectors() {
 		assertThat(this.properties.getJetty().getSelectors()).isEqualTo(10);
 	}
 
+	@Test
+	public void testCustomizeJettyStatistics() {
```
wilkinsona: If we drop the property, this test can be removed too.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,87 @@
...
+		final StatisticsHandler statsHandler = new StatisticsHandler();
+
+		for (Connector connector : server.getConnectors()) {
+			connector.addBean(statsHandler);
```
wilkinsona: The user may have already manually configured the connector with a statistics handler. We need to be sure that we don't end up with two such handlers. I think checking for an existing handler using `getBean(StatisticsHandler.class)` will do the trick.

evenh:
Thanks for taking the time to review this :-) I've pushed an updated commit and removed the one that introduced `server.jetty.publish-statistics`.

wilkinsona:
Thanks again. Having recalled that `StatisticsHandler` is a `HandlerWrapper` and looked a bit more closely at [Jetty's documentation](https://www.eclipse.org/jetty/documentation/current/statistics-handler.html), I don't think this PR is taking the right approach to configuring statistics. It's not completely clear from the XML, but I believe you need to create a `StatisticsHandler` and use it to wrap the server's existing `Handler`. You can see an example of this wrapping here: https://github.com/spring-projects/spring-boot/blob/ae45b6730b99025e457feda72a49bef78a1727f2/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyServletWebServerFactory.java#L179-L189.

evenh:
I've updated my approach to wrap the existing handlers.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,81 @@
...
+	@ConditionalOnWebApplication(type = Type.REACTIVE)
+	public WebServerFactoryCustomizer<JettyReactiveWebServerFactory> serverCapturingReactiveJettyCustomizer() {
+		return this::statisticsCustomizer;
+	}
+
+	private void wrapExistingHandlersInStatisticsHandler(Server server) {
+		final Handler[] existingHandlers = server.getHandlers();
+		final StatisticsHandler statsHandler = new StatisticsHandler();
+
+		statsHandler.setHandler(new HandlerCollection(existingHandlers));
```
wilkinsona: There's no need to use a `HandlerCollection` here. A `Server` only ever has a single `Handler` and `getHandlers()` just wraps that single handler in an array.

wilkinsona: We need to avoid adding a second `StatisticsHandler` to the server. I think it should be possible to get the `Handler` from the server and then walk down through all the wrappers by checking for instances of `HandlerWrapper`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfigurationTests.java
```diff
@@ -0,0 +1,176 @@
...
+
+						assertThat(registry.find("jetty.requests").meter())
+								.isNotNull();
+						assertThat(registry.find("jetty.requests.active").gauge())
+								.isNotNull();
+					});
+	}
+
+	@Test
+	public void sessionMetricsAreAvailableWhenEarlyMeterBinderInitializationOccurs() {
```
wilkinsona: jettyMetrics rather than sessionMetrics?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfiguration.java
evenh: Okay, I've tried to analyze the situation.

1. Get existing `Handler` for the `Server`
2. Check if instance of HandlerWrapper
    1. Get nested child handlers by casting the instance to HandlerWrapper and invoking [`#getChildHandlers()`](https://www.eclipse.org/jetty/javadoc/current/org/eclipse/jetty/server/handler/AbstractHandlerContainer.html#getChildHandlers--) (or maybe [`#getChildHandlersByClass(HandlerWrapper.class)`](https://www.eclipse.org/jetty/javadoc/current/org/eclipse/jetty/server/handler/AbstractHandlerContainer.html#getChildHandlersByClass-java.lang.Class-))
    2. Store all child handlers in a list
    3. Check list of child handlers for any existing instance of `StatisticsHandler`
3. Configure `server.setHandler` and field `statsHandler` to the following:
    - **If no existing `StatisticsHandler`:** Use the existing declared `statsHandler`
    - **If 1 existing `StatisticsHandler`:** Use this instance and remove it from the child handler list
    - **If >1 existing `StatisticsHandler`:** TBD?

Regardless of the outcome of **3**, the `StatisticsHandler#setHandler` must be invoked to wrap the child handlers that was extracted to a list. Here I would guess a `HandlerCollection` would be appropriate?

wilkinsona: 2 is where things get tricky and why we decided not to try to squeeze this into RC2. We basically need to navigate down from the server's handler so that we check to see if any of the wrapped handlers are a `StatisticsHandler`. I don't know exactly how to do this so some experimentation and accompanying tests would be the way to go.

If there happens to be multiple `StatisticHandler`s, I think we can just use the first one that we find. If there's more than one then that's a misconfiguration and I don't think we should add complexity to deal with it.

For 3, if there's no existing `StatisticsHandler` then we need to create one, use it to wrap the server's existing handler, set it on the server, and use it to create the `JettyStatisticsMetrics`. If we've found an existing `StatisticsHandler` then we should leave it in place and use it to create the `JettyStatisticsMetrics`.

evenh:
I can't quite see why the CI is failing :(

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/jetty/JettyMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,105 @@
...
+		return this::statisticsCustomizer;
+	}
+
+	@Bean
+	@ConditionalOnWebApplication(type = Type.REACTIVE)
+	public WebServerFactoryCustomizer<JettyReactiveWebServerFactory> serverCapturingReactiveJettyCustomizer() {
+		return this::statisticsCustomizer;
+	}
+
+	private void extractExistingOrCreateStatisticsHandler(Server server) {
```
wilkinsona: This still doesn't look quite right to me. A handler in Jetty can be wrapped any number of times. The code that looks for an existing statistics handler needs to work its way down through all of the wrappers to make sure that it doesn't add a duplicate statistics handler.

wilkinsona:
Closing due to the requested amendments not being made.

Issue #12060, opened 2018-02-14
Switch from webjars-locator to webjars-locator-core


wilkinsona:
[`webjars-locator`](https://github.com/webjars/webjars-locator) is "poorly named" and is "really just the RequireJS support that wraps the webjars-locator-core project". We should switch from `webjars-locator` to `webjars-locator-core` as Spring Framework [has recently done](https://github.com/spring-projects/spring-framework/commit/8d42888476e892dfa1b099684c9a381bbd26f916#diff-6b23f872b29c97049a56354504169781L29).

PR #12135, opened 2018-02-20
Update documentaion to webjars-locator-core

nkolosnjaji:
Update documentation about new webjars locator (core) dependency introduced here c587549a687e280bc002d3b58b7ac3efc2409112


rajadilipkolli:
good catch

snicoll:
God catch indeed @nkolosnjaji and thanks a lot for your contribution.

Relates to #12060

PR #12153, opened 2018-02-21
Use sized array for List.toArray()

izeye:
This PR changes to use sized array for `List.toArray()`.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveAndImperativeRepositoriesAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveAndImperativeRepositoriesAutoConfigurationTests.java
@@ -92,7 +92,7 @@ public class CouchbaseReactiveAndImperativeRepositoriesAutoConfigurationTests {
 					CouchbaseReactiveRepositoriesAutoConfiguration.class }) {
 				names.add(type.getName());
 			}
-			return names.toArray(new String[0]);
+			return names.toArray(new String[names.size()]);
 		}
 
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
@@ -137,7 +137,7 @@ public class FlywayAutoConfiguration {
 				String password = getProperty(this.properties::getPassword,
 						this.dataSourceProperties::getPassword);
 				flyway.setDataSource(url, user, password,
-						this.properties.getInitSqls().toArray(new String[0]));
+						this.properties.getInitSqls().toArray(new String[this.properties.getInitSqls().size()]));
 			}
 			else if (this.flywayDataSource != null) {
 				flyway.setDataSource(this.flywayDataSource);
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
@@ -113,7 +113,7 @@ public class EmbeddedLdapAutoConfiguration {
 
 	@Bean
 	public InMemoryDirectoryServer directoryServer() throws LDAPException {
-		String[] baseDn = this.embeddedProperties.getBaseDn().toArray(new String[0]);
+		String[] baseDn = this.embeddedProperties.getBaseDn().toArray(new String[this.embeddedProperties.getBaseDn().size()]);
 		InMemoryDirectoryServerConfig config = new InMemoryDirectoryServerConfig(baseDn);
 		if (hasCredentials(this.embeddedProperties.getCredential())) {
 			config.addAdditionalBindCredentials(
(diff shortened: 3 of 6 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveAndImperativeRepositoriesAutoConfigurationTests.java
```diff
@@ -92,7 +92,7 @@ public void shouldCreateInstancesForReactiveAndImperativeRepositories() {
 					CouchbaseReactiveRepositoriesAutoConfiguration.class }) {
 				names.add(type.getName());
 			}
-			return names.toArray(new String[0]);
+			return names.toArray(new String[names.size()]);
```
rajadilipkolli: Can't we write it using method reference as return names.toArray(String[]::new) ?

wilkinsona: No, I don't think so. AFAIK, you can't create an array using the syntax. Even if you could, it would presumably give you a 0-length array which would defeat the purpose of the change that @izeye is proposing.

joshiste:
Afaik a sized array is not always the fastest solution. When using a sized array the VM needs to initialize it with all zeroes, so it will be inited twice. When using a unsized array the VM can omit the second init as it knows that all positions are already empty. So it depends on the array size if the double init or the additional instantiation is faster...

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseReactiveAndImperativeRepositoriesAutoConfigurationTests.java
rajadilipkolli: @wilkinsona  You are correct, I mislook it.

izeye:
@joshiste Thanks for the feedback! I didn't intend to gain better performance via this change but tried to align them with the other parts in codebase and simply avoid unnecessary empty array creation. By the way I'm curious on the double initialization part from your comment. Could you explain more on it as I don't understand where the double initialization is coming from?

wilkinsona:
FWIW, I don't understand either. If the passed in array is too small, a new array is initialized reflectively.  Once armed with an appropriate array, `System.arraycopy` is used to populate it. The only difference when the passed in array is used is that, if its length is greater than the list's size, array[size] is nulled out. I'd be surprised if the extra if a setting a single element to null is slower than reflectively creating a new array.

joshiste:
I've just done a simple JMH on this. https://gist.github.com/joshiste/9c49661be53b752b36042b05581dbf20

Results:
image
(please note the logarithmic scale)

```
Benchmark               (size)   Mode  Cnt         Score          Error  Units
ArrayInit.doubledSize        1  thrpt    5  38215435.866 ±  1737630.022  ops/s
ArrayInit.doubledSize       10  thrpt    5  28907110.655 ±  1303972.196  ops/s
ArrayInit.doubledSize      100  thrpt    5   5771810.166 ±   739455.196  ops/s
ArrayInit.doubledSize     1000  thrpt    5    495665.297 ±    52844.547  ops/s
ArrayInit.doubledSize    10000  thrpt    5     45936.938 ±     7075.577  ops/s
ArrayInit.doubledSize   100000  thrpt    5      2526.681 ±      647.546  ops/s
ArrayInit.doubledSize  1000000  thrpt    5       191.916 ±        4.620  ops/s
ArrayInit.halfSize           1  thrpt    5  26223177.007 ±  2025324.645  ops/s
ArrayInit.halfSize          10  thrpt    5  27480147.367 ±  2661236.901  ops/s
ArrayInit.halfSize         100  thrpt    5   6406770.803 ±   102468.320  ops/s
ArrayInit.halfSize        1000  thrpt    5    591167.895 ±    24876.196  ops/s
ArrayInit.halfSize       10000  thrpt    5     53910.150 ±     3115.852  ops/s
ArrayInit.halfSize      100000  thrpt    5      2013.225 ±     1074.521  ops/s
ArrayInit.halfSize     1000000  thrpt    5       175.144 ±       47.438  ops/s
ArrayInit.perfectFit         1  thrpt    5  37938686.374 ±  3671828.175  ops/s
ArrayInit.perfectFit        10  thrpt    5  29661197.318 ±  1763519.988  ops/s
ArrayInit.perfectFit       100  thrpt    5   6324526.258 ±   154731.448  ops/s
ArrayInit.perfectFit      1000  thrpt    5    511508.907 ±    80152.869  ops/s
ArrayInit.perfectFit     10000  thrpt    5     49879.763 ±     2140.531  ops/s
ArrayInit.perfectFit    100000  thrpt    5      2458.360 ±     1117.743  ops/s
ArrayInit.perfectFit   1000000  thrpt    5       177.814 ±       28.586  ops/s
ArrayInit.zeroSize           1  thrpt    5  77624333.568 ± 14868114.381  ops/s
ArrayInit.zeroSize          10  thrpt    5  46873656.635 ±  6826839.885  ops/s
ArrayInit.zeroSize         100  thrpt    5   6975644.034 ±    95271.748  ops/s
ArrayInit.zeroSize        1000  thrpt    5    617080.671 ±    12099.464  ops/s
ArrayInit.zeroSize       10000  thrpt    5     56308.250 ±     4184.184  ops/s
ArrayInit.zeroSize      100000  thrpt    5      3173.536 ±      652.068  ops/s
ArrayInit.zeroSize     1000000  thrpt    5       185.522 ±       44.172  ops/s
```

joshiste:
Imho this proves (for smaller array sizes) that calling toArray() yields the highest throughput.

izeye:
@joshiste Thanks for the feedback! I found a few articles backing your feedback like this one: https://shipilev.net/blog/2016/arrays-wisdom-ancients/

Although I didn't read them thoroughly yet but based on the articles the opposite direction of this PR looks right. I'll look into it soon.

wilkinsona:
Very interesting (and surprising). Thanks, @joshiste. I just stumbled across the article that @izeye linked to as well.

philwebb:
I always preferred the syntax of the zero-array better. I've raised https://github.com/spring-projects/spring-boot/issues/12160 to switch to it.

joshiste:
My explanation wasn't quite correct...

If the toArray() method instantiates the array itself it can completely omit the initialisation of the array as it we will do the initialisation with the content.
If you instantiate the array beforehand the JVM will zero init the array, because it doesn't know that all indices will be set afterwards... So using a correct sized array results in a initialisation which won't happen when using zero sized arrays

izeye:
I did read the above mentioned article closely and I think I understand the benchmark result better now which looks counter-intuitive at first glance. I'd recommend anyone interested in this topic to read the article.

@joshiste Thanks for enlightening me on this!

PR #12200, opened 2018-02-23
Persistent Kafka AdminClient for health checks

### LINKED ISSUE #12198: KafkaHealthIndicator and older brokers
Currently, when the broker doesn't support obtaining configuration (e.g. 0.10.2.0), the health indicator reports `Down`.

Consider skipping the `replicationFactor` check if the broker cannot provide that information.

See https://github.com/spring-projects/spring-boot/pull/11515#issuecomment-368044582

garyrussell:
Use a single `AdminClient` in `KafkaHealthIndicator` instead of opening a new connection
each time.

Also fixes #12198

If the replication factor can't be determined (older broker), don't perform the replication
factor health check.


garyrussell:
Also see https://github.com/spring-projects/spring-boot/pull/11515#issuecomment-368252264

snicoll:
Thanks for the PR. We've been looking at this issue most of today and came to the conclusion it isn't mature enough for us to provide a health indicator out of the box. We'll revert it and reconsider this feature for a latter release.

Issue #12088, opened 2018-02-16
Wrong ClassLoader set within JMX Endpoint


Dav1dde:
Simple Scenario: Spring Boot 2.0.0.RC1 Application with Spring-Cloud Config is built as WAR-File and deployed into a Tomcat 8.5. Trying to refresh the context using the JMX Refresh endpoint results in an Exception:

```
java.lang.IllegalStateException: Cannot load configuration class: org.springframework.cloud.context.refresh.ContextRefresher$Empty
        at org.springframework.context.annotation.ConfigurationClassPostProcessor.enhanceConfigurationClasses(ConfigurationClassPostProcessor.java:416) ~[spring-context-5.0.3.RELEASE.jar:5.0.3.RELEASE]
        at org.springframework.context.annotation.ConfigurationClassPostProcessor.postProcessBeanFactory(ConfigurationClassPostProcessor.java:254) ~[spring-context-5.0.3.RELEASE.jar:5.0.3.RELEASE]
        at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors(PostProcessorRegistrationDelegate.java:284) ~[spring-context-5.0.3.RELEASE.jar:2.5.5]
        at org.springframework.context.support.PostProcessorRegistrationDelegate.invokeBeanFactoryPostProcessors(PostProcessorRegistrationDelegate.java:128) ~[spring-context-5.0.3.RELEASE.jar:2.5.5]
        at org.springframework.context.support.AbstractApplicationContext.invokeBeanFactoryPostProcessors(AbstractApplicationContext.java:693) ~[spring-context-5.0.3.RELEASE.jar:2.5.5]
        at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:531) ~[spring-context-5.0.3.RELEASE.jar:2.5.5]
        at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:752) ~[spring-boot-2.0.0.RC1.jar:2.0.0.RC1]
        at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:388) ~[spring-boot-2.0.0.RC1.jar:2.0.0.RC1]
        at org.springframework.boot.SpringApplication.run(SpringApplication.java:327) ~[spring-boot-2.0.0.RC1.jar:2.0.0.RC1]
[...]

PR #12209, opened 2018-02-24
Set classloader for JMX endpoints to application classloader

Dav1dde:
Related to and fixes:  #12088 

Sets the Thread's classloader for JMX endpoints, before invoking the endpoint, to the classloader which was initially used to load the context.

Before:
```
--- Class Loaders
getClass().getClassLoader(): java.net.URLClassLoader@4ac20dd2
currentThread().getContextClassLoader(): sun.misc.Launcher$AppClassLoader@18b4aac2
ClassUtils.getDefaultClassLoader(): sun.misc.Launcher$AppClassLoader@18b4aac2
---
```

After:
```
--- Class Loaders
getClass().getClassLoader(): java.net.URLClassLoader@4ac20dd2
currentThread().getContextClassLoader(): java.net.URLClassLoader@4ac20dd2
ClassUtils.getDefaultClassLoader(): java.net.URLClassLoader@4ac20dd2
---
```

Test-Code (using sample Application from start.spring.io including web and actuator):

```java
@Component
@JmxEndpoint(id = "classloader")
public class JmxClassloaderEndpoint {
    @WriteOperation
    public void someOperation() {
        System.out.println("--- Class Loaders");
        System.out.println("getClass().getClassLoader(): " + getClass().getClassLoader());
        System.out.println("currentThread().getContextClassLoader(): " + Thread.currentThread().getContextClassLoader());
        System.out.println("ClassUtils.getDefaultClassLoader(): " + ClassUtils.getDefaultClassLoader());
        System.out.println("---");
    }
}
```

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/jmx/EndpointMBean.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/jmx/EndpointMBean.java
@@ -29,8 +29,11 @@ import javax.management.MBeanException;
 import javax.management.MBeanInfo;
 import javax.management.ReflectionException;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
 import reactor.core.publisher.Mono;
 
+import org.springframework.beans.factory.BeanClassLoaderAware;
 import org.springframework.boot.actuate.endpoint.InvalidEndpointRequestException;
 import org.springframework.boot.actuate.endpoint.InvocationContext;
 import org.springframework.boot.actuate.endpoint.SecurityContext;
@@ -46,10 +49,14 @@ import org.springframework.util.ClassUtils;
  * @author Phillip Webb
  * @since 2.0.0
  */
-public class EndpointMBean implements DynamicMBean {
+public class EndpointMBean implements DynamicMBean, BeanClassLoaderAware {
 
 	private static final boolean REACTOR_PRESENT = ClassUtils.isPresent(
 			"reactor.core.publisher.Mono", EndpointMBean.class.getClassLoader());
+
+	private static final Log logger = LogFactory.getLog(EndpointMBean.class);
+
+	private ClassLoader classLoader;
 
 	private final JmxOperationResponseMapper responseMapper;
 
@@ -67,6 +74,11 @@ public class EndpointMBean implements DynamicMBean {
 		this.endpoint = endpoint;
 		this.info = new MBeanInfoFactory(responseMapper).getMBeanInfo(endpoint);
 		this.operations = getOperations(endpoint);
+	}
+
+	@Override
+	public void setBeanClassLoader(ClassLoader classLoader) {
+		this.classLoader = classLoader;
 	}
 
 	private Map<String, JmxOperation> getOperations(ExposableJmxEndpoint endpoint) {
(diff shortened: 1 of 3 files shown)


Dav1dde:
Build failure seems to be random and not related to this change, can someone re-trigger the build?

wilkinsona:
Closing and re-opening to trigger a new build.

wilkinsona:
I think that used to work with Travis. Doesn’t appear to have worked with Concourse.

wilkinsona:
I’ve manually triggered a new build on Concourse but I’m not sure it’s building the right change. There are some git-related errors in the logs.

Dav1dde:
Any news or interest in this?

Dav1dde:
@snicoll thank you!

You're right, it feels hack-ish. Want me to update it so the classloader comes from `BeanClassLoaderAware` instead of `getClass().getClassLoader()`?

snicoll:
That won't change much I am afraid. Switching the classloader of the current thread (without restoring the previous instance once the method has been invoked sounds definitely wrong to me. I would only do this as a last resort option.

Dav1dde:
Oh, that's not a problem, actually had a version that restored it. (that's also what we do at work right now). I'll update this tonight and let's see if you like that more.

snicoll:
Sorry, I meant that this proposal doesn’t look great to me (and at the very least the classloader should be restored).

Dav1dde:
Updated and restoring class loader now (no-op if bean class loader is not set, e.g. in tests). I personally don't like the solution either but I can't think of anything better, if you can think of something let me know and I can change the PR accordingly.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/jmx/EndpointMBean.java
```diff
@@ -90,7 +98,13 @@ public Object invoke(String actionName, Object[] params, String[] signature)
 					+ "' has no operation named " + actionName;
 			throw new ReflectionException(new IllegalArgumentException(message), message);
 		}
-		return invoke(operation, params);
+		ClassLoader previousClassLoader = ClassUtils.overrideThreadContextClassLoader(this.classLoader);
```
wilkinsona: This may throw a `SecurityException`. I think we should set the TCCL on a best-effort basis so any `SecurityException` that's thrown should be caught (and logged) and we should then proceed with the `invoke`.

wilkinsona:
FWIW, other than the `SecurityException` that I've just commented about, I think we should make this change.

Dav1dde:
I'll update it later tonight.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/jmx/JmxEndpointExporter.java
```diff
@@ -89,6 +97,7 @@ private ObjectName register(ExposableJmxEndpoint endpoint) {
 		try {
 			ObjectName name = this.objectNameFactory.getObjectName(endpoint);
 			EndpointMBean mbean = new EndpointMBean(this.responseMapper, endpoint);
+			mbean.setBeanClassLoader(this.classLoader);
```
Dav1dde: I considered having the class loader as a constructor argument, but for me the class loader is optional for the `EndpointMBean`.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/jmx/EndpointMBean.java
```diff
@@ -90,7 +102,28 @@ public Object invoke(String actionName, Object[] params, String[] signature)
...
+		if (classLoader == null) {
+			return null;
+		}
+
+		try {
+			return ClassUtils.overrideThreadContextClassLoader(classLoader);
+		}
+		catch (SecurityException exc) {
+			// can't set class loader, ignore it and proceed
+			logger.warn("Unable to override class loader for JMX endpoint.");
```
Dav1dde: Not sure if a info log would be better here.

Dav1dde:
Branch updated, I added a few comments where I am not sure if things are okay this way and fit into the spring codebase.

snicoll:
@Dav1dde thank you for your first contribution to Spring Boot. This has been merged in `2.0.x` and `master`.

PR #12216, opened 2018-02-25
Add cache actuator endpoint for listing and clearing caches

### LINKED ISSUE #2625: Create a cache actuator endpoint.
Hi,

I'd like to have a `/cache` actuator endpoint that will list all the cache regions will all the statistics for each cache region. More specifically:

```
GET /cache
[{
  "id" : "simpleCMSWidgetData",
  "stats" : {
    "AverageGetTime" : 0.0,
    "CacheHitPercentage" : 77.27273,
    "CacheEvictions" : 0,
    "AverageRemoveTime" : 0.0,
    "CacheMisses" : 30,
    "AveragePutTime" : 35.89833,
    "CacheMissPercentage" : 22.727274,
    "CacheHits" : 102,
    "CacheRemovals" : 0,
    "CachePuts" : 30,
    "CacheGets" : 132
  }
}, {
  "id" : "productData",
  "stats" : {
    "AverageGetTime" : 0.0,
    "CacheHitPercentage" : 50.0,
    "CacheEvictions" : 0,
    "AverageRemoveTime" : 0.0,
    "CacheMisses" : 1,
    "AveragePutTime" : 33.119,
    "CacheMissPercentage" : 50.0,
    "CacheHits" : 1,
    "CacheRemovals" : 0,
    "CachePuts" : 1,
    "CacheGets" : 2
  }
} ]
```

and 

```
GET /cache/productData
{
  "id" : "productData",
  "stats" : {
    "AverageGetTime" : 0.0,
    "CacheHitPercentage" : 50.0,
    "CacheEvictions" : 0,
    "AverageRemoveTime" : 0.0,
    "CacheMisses" : 1,
    "AveragePutTime" : 33.119,
    "CacheMissPercentage" : 50.0,
    "CacheHits" : 1,
    "CacheRemovals" : 0,
    "CachePuts" : 1,
    "CacheGets" : 2
  }
}
```

and 

```
DELETE /cache
```

would clear all the cache

and 

```
DELETE /cache/productData
```

would clear the given cache region.

joshiste:
This commits adds an actuator endpoint which lists the caches per
context and cacheManager and provides a delete operation to clear the
caches. As the statistics are exposed via the metrics endpoint they are
not included

closes #2625

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CacheEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CacheEndpoint.java
@@ -1,0 +1,156 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.cache;
+
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.HashMap;
+import java.util.List;
+import java.util.Map;
+
+import org.springframework.boot.actuate.endpoint.annotation.DeleteOperation;
+import org.springframework.boot.actuate.endpoint.annotation.Endpoint;
+import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
+import org.springframework.cache.CacheManager;
+import org.springframework.context.ApplicationContext;
+import org.springframework.lang.Nullable;
+
+/**
+ * {@link Endpoint} to expose cache operations.
+ *
+ * @author Johannes Edmeuer
+ * @since 2.0.0
+ */
+@Endpoint(id = "caches")
+public class CacheEndpoint {
+	private final ApplicationContext context;
+
+	public CacheEndpoint(ApplicationContext context) {
+		this.context = context;
+	}
+
+	@ReadOperation
+	public ApplicationCacheManagerBeans cacheManagerBeans() {
+		ApplicationContext target = this.context;
+		Map<String, ContextCacheManagerBeans> contextCacheManagerBeans = new HashMap<>();
+		while (target != null) {
+			Map<String, CacheManagerDescriptor> cacheManagerBeans = new HashMap<>();
+			target.getBeansOfType(CacheManager.class)
+					.forEach((name, cacheManager) -> cacheManagerBeans.put(name,
+							new CacheManagerDescriptor(cacheManager.getCacheNames())));
+			ApplicationContext parent = target.getParent();
+			contextCacheManagerBeans.put(target.getId(), new ContextCacheManagerBeans(
+					cacheManagerBeans, parent == null ? null : parent.getId()));
+			target = parent;
+		}
+		return new ApplicationCacheManagerBeans(contextCacheManagerBeans);
+	}
+
+	@DeleteOperation
+	public void clearCaches(@Nullable String contextId, @Nullable String cacheManagerName,
+			@Nullable String cacheName) {
+		ApplicationContext target = this.context;
+		while (target != null) {
+			if (contextId == null || contextId.equals(target.getId())) {
+				clearCaches(target, cacheManagerName, cacheName);
+			}
+			target = target.getParent();
+		}
+	}
+
+	private void clearCaches(ApplicationContext context, String cacheManagerName,
+			String cacheName) {
+		if (cacheManagerName == null) {
+			context.getBeansOfType(CacheManager.class)
+					.forEach((name, manager) -> this.clearCaches(manager, cacheName));
+		}
+		else {
+			this.clearCaches(context.getBean(cacheManagerName, CacheManager.class),
+					cacheName);
+		}
+	}
+
+	private void clearCaches(CacheManager cacheManager, String cacheName) {
+		if (cacheName == null) {
+			cacheManager.getCacheNames().forEach(cn -> cacheManager.getCache(cn).clear());
+		}
+		else {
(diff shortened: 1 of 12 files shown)


philwebb:
Nice, thanks.

snicoll:
See #2625

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CacheEndpoint.java
```diff
@@ -0,0 +1,156 @@
...
+@Endpoint(id = "caches")
+public class CacheEndpoint {
+	private final ApplicationContext context;
+
+	public CacheEndpoint(ApplicationContext context) {
+		this.context = context;
+	}
+
+	@ReadOperation
+	public ApplicationCacheManagerBeans cacheManagerBeans() {
```
snicoll: `cacheManagers` sounds more sensible to me.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CacheEndpoint.java
```diff
@@ -0,0 +1,156 @@
...
+			ApplicationContext parent = target.getParent();
+			contextCacheManagerBeans.put(target.getId(), new ContextCacheManagerBeans(
+					cacheManagerBeans, parent == null ? null : parent.getId()));
+			target = parent;
+		}
+		return new ApplicationCacheManagerBeans(contextCacheManagerBeans);
+	}
+
+	@DeleteOperation
+	public void clearCaches(@Nullable String contextId, @Nullable String cacheManagerName,
```
snicoll: I wonder if we need to bring the complexity of `BeansEndpoint` here. If you are working with an application, I don't see the use case of having two cache managers with the same name in different context. 

As a user, having to specify the context is very annoying isn't it?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CacheEndpoint.java
```diff
@@ -0,0 +1,156 @@
...
+		ApplicationContext target = this.context;
+		while (target != null) {
+			if (contextId == null || contextId.equals(target.getId())) {
+				clearCaches(target, cacheManagerName, cacheName);
+			}
+			target = target.getParent();
+		}
+	}
+
+	private void clearCaches(ApplicationContext context, String cacheManagerName,
```
snicoll: Naturally I would go with clearing a cache by name and just throw an exception if there are more than one `CacheManager` defining a cache with that name. Again, I don't think we should bring that complexity to the user for the very few that chose a complicated setup. The name of the cache manager could be used as a selector perhaps?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CacheEndpoint.java
```diff
@@ -0,0 +1,156 @@
...
+		else {
+			cacheManager.getCache(cacheName).clear();
+		}
+	}
+
+	/**
+	 * Description of an application's {@link CacheManager} beans, primarily intended for
+	 * serialization to JSON.
+	 */
+	public static final class ApplicationCacheManagerBeans {
```
snicoll: I have a feeling those names comes from the `BeansEndpoint`. I don't think this applies here.

joshiste:
@snicoll Thanks for your comments. I've copied some stuff from the liquibaseEndpoint hence the handling of multiple contexts.

joshiste:
I've renamed the endpoint to `CachesEndpoint`.
I've removed the handling of complex contexts.
I've removed the superfluous config metadata.

I didn't make the cacheManager a `@Selector` since having it as queryParam allows to delete all caches over all cacheManagers.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cache/CachesEndpoint.java
```diff
@@ -0,0 +1,109 @@
...
+					cacheName);
+		}
+	}
+
+	private void clearCaches(CacheManager cacheManager, String cacheName) {
+		if (cacheName == null) {
+			cacheManager.getCacheNames().forEach(cn -> cacheManager.getCache(cn).clear());
+		}
+		else {
+			cacheManager.getCache(cacheName).clear();
```
ptahchiev: I believe  a null-pointer check is necessary here.

ptahchiev:
Maybe it's a good idea to add also a method for removing a single entry from the `Cache` region? Would be nice to also list all the entries in a given cache region, but the current spring cache API does not allow it (jcache does it allow it).

joshiste:
Added null check and rebased to master

joshiste:
>  I don't like the idea of using a bean name there either.
If we need to qualify the caches in case of duplicates names in two CacheManagers, I guess there is nothing else left to distinct the caches

How about rendering links for cache eviction into the response and just prefixing the cache name in case of ambiguity?
The downside is: it would need a mvc extension for the endpoint and doesn't clarify the usage in case of using the MXBean

```json
{
  "caches": [
    {
      "name": "manager-foo/ambiguousName",
      "evict": "/actuator/caches/manager-foo/ambiguousName"
    },
    {
      "name": "manager-bar/ambiguousName",
      "evict": "/actuator/caches/manager-bar/ambiguousName"
    },
    {
      "name": "uniqueName",
      "evict": "/actuator/caches/manager-bar/uniqueName"
    }
  ]
}
```
... if a cache contains a `/` it need's to be encoded...

snicoll:
@joshiste what I had in mind was more an attribute that would qualify the cache

```json
{
  "caches": [
    {
      "name": "ambiguousName",
      "cacheManager": "manager-foo",
      "_links": {
        "evict": "/actuator/caches/ambiguousName?cacheManager=manager-foo"
      }
    },
    {
      "name": "ambiguousName",
      "cacheManager": "manager-foo",
      "_links": {
        "evict": "/actuator/caches/ambiguousName?cacheManager=manager-bar"
      }
    },
    {
      "name": "uniqueName",
      "cacheManager": "manager-foo",
      "_links": {
        "evict": "/actuator/caches/manager-bar/uniqueName"
      }
    }
  ]
}
```

joshiste:
I've modified the output of the read operation as you suggested.

snicoll:
@joshiste thanks for the PR. While reviewing, I've reworked and significantly changed the initial proposal, see fb8a5a9 for more details.

joshiste:
Yay! Thanks for merging!

PR #12217, opened 2018-02-25
Polish: move DEFAULT_CLASSPATH to a class that used it and lower its visibility

igor-suhorukov:


philwebb:
CLI extensions might be using that constant so I'm not sure we should make it private.

PR #12269, opened 2018-02-28
Add Netty Compression support

smaldini:
Reactor Netty 0.7.5 now supports all Compression options.
Unlike previous Reactor Netty versions, the minimum compression threshold is now based on the content-length instead of accumulating for a while the data.


bclozel:
Merged with 7866b946f771a6497a4211b137df3d87f9bd45bf
Thanks!

PR #12272, opened 2018-02-28
fix #12271 UnsatisfiedDependencyException with a datasource proxy and…

### LINKED ISSUE #12271: App does not load with a proxied DataSource using Hikari
With Spring-boot 2.0.0.RC2, when using HikariCP and a datasource proxy or any wrapper around a datasource, there is UnsatisfiedDependencyException:

```
Exception encountered during context initialization - cancelling refresh attempt: org.springframework.beans.factory.UnsatisfiedDependencyException: Error creating bean with name 'org.springframework.boot.autoconfigure.jdbc.DataSourceJmxConfiguration$Hikari': Unsatisfied dependency expressed through constructor parameter 0; nested exception is org.springframework.beans.factory.BeanNotOfRequiredTypeException: Bean named 'dataSource' is expected to be of type 'com.zaxxer.hikari.HikariDataSource' but was actually of type 'com.sun.proxy.$Proxy49'
```

Instead of a blocking exception I expect this to succeed like it was in Spring-boot 1.x.

Test case:
https://github.com/evernat/datasource-proxy-hikaricp
Use mvn spring-boot:run and see UnsatisfiedDependencyException instead of success.

The test case uses a DataSourceBeanPostProcessor to create a proxy as just a "a wrapper around the original" as suggested by the BeanPostProcessor's javadoc and it does not use ProxyFactory.setProxyTargetClass(true) to create a CGLIB proxy.

I think that it should still work in Spring-boot 2.0 and there are several really valid use cases for setProxyTargetClass(true) not being used in my opinion:
1. Calling setProxyTargetClass in this case is not needed in Spring-boot 1.x. And JMX enabled by default is not a good reason to change that in Spring-boot 2.x.
2. Or the datasource proxy may not be created by the application's code and may not be created with Spring either.
3. Or CGLIB may not be available in this application.
4. And the BeanPostProcessor's javadoc only suggest to create here a "wrapper around the original" as a standard practice ; it does not suggest to create a CGLIB proxy.

so why Spring-boot 2.0.0.RC2 requires to use CGLIB here to create a proxy of a standard DataSource ?
(The fact that it is a HikariDataSource and that Spring-boot registers the HikariDataSource into JMX is unknown in the DataSourceBeanPostProcessor by separation of concerns and should also be transparent for the application as much as possible in my opinion.)

I will submit a PR in a minute.

evernat:
… HikariCP

fix #12271

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceJmxConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceJmxConfiguration.java
@@ -55,8 +55,16 @@ class DataSourceJmxConfiguration {
 
 		private final ObjectProvider<MBeanExporter> mBeanExporter;
 
-		Hikari(HikariDataSource dataSource, ObjectProvider<MBeanExporter> mBeanExporter) {
-			this.dataSource = dataSource;
+		Hikari(DataSource dataSource, ObjectProvider<MBeanExporter> mBeanExporter) throws SQLException {
+			if (dataSource instanceof HikariDataSource) {
+				this.dataSource = (HikariDataSource) dataSource;
+			}
+			else if (dataSource.isWrapperFor(HikariDataSource.class)) {
+				this.dataSource = (HikariDataSource) dataSource.unwrap(HikariDataSource.class);
+			}
+			else {
+				throw new IllegalArgumentException("DataSource is expected to be of type '" + HikariDataSource.class.getName() + "' but was actually of type '" + dataSource.getClass().getName() + "'");
+			}
 			this.mBeanExporter = mBeanExporter;
 		}
 


pivotal-issuemaster:
@evernat Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12272)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12272) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@evernat Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12272)!

snicoll:
@evernat thanks a lot for the PR. I've decided to take a slightly different direction (see [this commit](https://github.com/snicoll/spring-boot/commit/b3b80d34098d99d2424573b5e89b2d4c9dcdc1ce) pending review).

I would have loved to give you some direction to adapt your PR and add a test to it but unfortunately we're in a hurry here with 2.0 GA about to be released. I really want to get this fix in (this should happen later today).

Thanks again.

PR #12276, opened 2018-02-28
Use forEach() directly without map() in LambdaSafe.Callbacks.invoke()

izeye:
This PR changes to use `forEach()` directly without `map()` in `LambdaSafe.Callbacks.invoke()`.


snicoll:
Thanks for the PR @izeye but @philwebb already polished this in bbf94c22da1549693eb357d3acc5098c385e67bc

Issue #12286, opened 2018-02-28
ClassCastException on LogbackMetrics when logback is present but not the SLF4J binding


jkschneider:
We need some additional conditional guard around the autoconfiguration of `LogbackMetrics` to ensure that `LoggerFactory.getILoggerFactory()` is of type `ch.qos.logback.classic.LoggerContext` before proceeding to wire the binder.

PR #12288, opened 2018-02-28
Only wire LogbackMetrics when logback is the selected SLF4J binding

jkschneider:
This is an attempt at #12286. Testing is tricky since the condition's failure requires multiple SLF4J bindings to be present and for SLF4J to deterministically select the _non-logback_ binding.


wilkinsona:
Sorry, @jkschneider. I didn't see this till after pushing https://github.com/spring-projects/spring-boot/commit/c4c50b7c49f0e240a0e5cc65cffd2b609e686dab. You might be interested in that commit as it shows a way to test this using our `ModifiedClassPathRunner`.

PR #12295, opened 2018-03-01
Make UserDetailsServiceAutoConfiguration.getOrDeducePassword() private

izeye:
It looks unnecessary to be `public`.


snicoll:
> It looks unnecessary to be public.

I'd be tempted to agree but I think there is more to it. @philwebb made that public in ba86b684847e6ecfc8d77d1c4d293099882aecfb so let's see if he can give us more context.

philwebb:
🤷‍♂️

PR #12304, opened 2018-03-01
Polish: replace class "Stack" by more modern "Deque"

igor-suhorukov:


philwebb:
These are all great enhancements. Thanks for submitting them!

igor-suhorukov:
@philwebb thank you

PR #12324, opened 2018-03-02
Improve docs on custom metrics

jkschneider:
Indicate that simple injection is preferred unless you have a reusable suite of metrics for which `MeterBinder` encapsulation is helpful.

cc / @snicoll This is the improved verbiage as promised.


wilkinsona:
Thank you, @jkschneider!

PR #12331, opened 2018-03-03
Expose Spring Integration graph using an actuator

TYsewyn:
Based on the idea of @michael-simons (read this [post](http://info.michael-simons.eu/2018/02/13/revised-actuators-in-spring-boot-2/)), this PR will add the functionality of exposing the Spring Integration graph via JMX, Jersey, Spring MVC & WebFlux.

NB: Currently says `@since 2.1.0` in the Javadoc.


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/integration/IntegrationGraphEndpointAutoConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for the {@link IntegrationGraphEndpoint}.
+ *
+ * @author Tim Ysewyn
+ * @since 2.1.0
+ */
+@Configuration
+@ConditionalOnClass(IntegrationGraphServer.class)
+public class IntegrationGraphEndpointAutoConfiguration {
```
michael-simons: You have to add this to `spring.factories` as well, see `/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/spring.factories`.

This mechanism facilitates the fact that the autoconfiguration classes can be looked up instead are discovered via a package scan.

TYsewyn: Good catch! For some reason it didn't get committed & pushed, but I do remember making the change 🤔
Anyway, will do that right away! Thanks!

wilkinsona:
@garyrussell @artembilan what do you think please?

garyrussell:
Fine with me; SI currently only supports MVC and WebFlux. Thanks Tim!

My only question is whether we should auto configure a `IntegrationGraphServer` by default when actuators are enabled.

This seems to be dependent on the presence of a bean; there is little overhead if it's never referenced.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/asciidoc/endpoints/integrationgraph.adoc
```diff
@@ -0,0 +1,40 @@
...
+
+include::{snippets}integrationgraph/graph/http-response.adoc[]
+
+
+
+[[integrationgraph-retrieving-response-structure]]
+=== Response Structure
+
+The response contains all Spring Integration components used within the application, as well as the links between them.
+More info about the structure can be found in the https://docs.spring.io/spring-integration/reference/htmlsingle/#integration-graph[subchapter] for the integration graph in the reference documentation of Spring Integration.
```
artembilan: I wouldn't use `htmlsingle` here as a cross-link. Consider to switch to this one: https://docs.spring.io/spring-integration/reference/html/system-management-chapter.html#integration-graph.

Also I don't think that `info` is formal word. The `information` doesn't pull any questions from me :smile:

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/integration/IntegrationGraphEndpointAutoConfiguration.java
```diff
@@ -0,0 +1,49 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
```
artembilan: Is it a rule in Spring Boot to include `2012` as well to new classes?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/integration/IntegrationGraphEndpointAutoConfiguration.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.integration.support.management.graph.IntegrationGraphServer;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for the {@link IntegrationGraphEndpoint}.
```
artembilan: I don't think that it's good idea to use imports only for JavaDocs.
Not sure how it is for Spring Boot, but in SI we disallow such a code via particular Checkstyle rule.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/integration/package-info.java
```diff
@@ -0,0 +1,20 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
```
artembilan: Is it really OK to have Copyright in the `package-info.java` files?

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/documentation/IntegrationGraphEndpointDocumentationTests.java
```diff
@@ -0,0 +1,69 @@
...
+import org.springframework.restdocs.mockmvc.MockMvcRestDocumentation;
+
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.get;
+import static org.springframework.test.web.servlet.request.MockMvcRequestBuilders.post;
+import static org.springframework.test.web.servlet.result.MockMvcResultMatchers.status;
+
+/**
+ * Tests for generating documentation describing the {@link IntegrationGraphEndpoint}.
+ *
+ * @author Tim Ysewyn
```
artembilan: I think `@since 2.1.0` is applied to test classes as well.

philwebb: Actually, we don't `@since` tests.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/integration/package-info.java
philwebb: Yes please, that's consistent with the others that we have.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/integration/IntegrationGraphEndpointAutoConfiguration.java
philwebb: We don't have a similar restriction, but now I'm wondering if we should. It's fine for now.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/integration/IntegrationGraphEndpointAutoConfiguration.java
philwebb: yes

TYsewyn:
@garyrussell @artembilan I was thinking the same. Should I just put following under `IntegrationAutoConfiguration.IntegrationConfiguration`?

```
@ConditionalOnMissingBean
@Bean
public IntegrationGraphServer integrationGraphServer() {
	return new IntegrationGraphServer();
}
```
And is this PR the right one to do this? I would assume that you would like to backport this to 1.x if possible?

artembilan:
IMO that's the right place to do that.
And I don't think that we are going to backport it. There is no bug in the subject and there is always a workaround to do the same task using existing tools and hooks.

TYsewyn:
@artembilan @garyrussell last addition has been pushed.
If everything is fine for you I think this PR can be marked for 2.1.0.M1?
Please let me know if you need me to change some things!

snicoll:
@TYsewyn we'll review and assign a milestone to this one in due course. Thanks!

snicoll:
I took an extensive look at this one and polished in my own fork. I didn't like the fact an `IntegrationGraphServer` is created as part of the main auto-configuration. That's a management concern so it shouldn't be there at all.

I've tried to make it conditional to the fact that the endpoint is not enabled but there is no easy way to do that as `@ConditionalOnEnabledEndpoint` only works on specific types. Perhaps we should revisit this a bit.

Also I didn't get why the endpoint is disabled by default. Is that really what we want?

artembilan:
I had the same thought originally, Stephane. Right, the `IntegrationGraphServer` is a management concern, but from the other hand it exists in the `core` module and it is a simple as possible passive component. So, in the end I decided that it won't hurt to keep one more bean from the main auto-configuration.

If that is still a concern, let's consider to instantiate it from some actuator auto-configuration then!
The problem that it is an independent component and can be used even if a mentioned endpoint is disabled.

+1 do not make the endpoint disabled by default. Missed that part somehow during review...
Any explanations on the matter, @TYsewyn ?

Thanks

snicoll:
Thanks for the feedback.

> and it is a simple as possible passive component

I am not sure I agree with that. There is an event listener that will build the graph once the context has refreshed. I see no reason to have that running if you don't use the actuator. If you need this feature, you can define the bean (as you'd probably do now).

> The problem that it is an independent component and can be used even if a mentioned endpoint is disabled.

That's fine but IMO there is a difference between providing this (that only the endpoint would be using) and providing this when the endpoint is required. My preference is for the latter but unfortunately, that's not easy to implement.

> +1 do not make the endpoint disabled by default.

I've already changed that in my [polish commit](https://github.com/snicoll/spring-boot/commit/c2e903fed7c8f6227e30a048c63dce9bf81233a9#diff-4556a22e9c024f917dd4d448c30f5330R31)

artembilan:
Stephane, 

I think we're good to go with your polishing: you have addressed all the concerns! :+1:

TYsewyn:
I disabled the endpoint because there wasn't an `IntegrationGraphServer` bean present in the auto-configuration, so it didn't make any sense to bootstrap that endpoint by default.
We added one later and I forgot to switch it on by default when adding the bean.

spring-boot-project/spring-boot-docs/src/main/asciidoc/production-ready-features.adoc
```diff
@@ -105,6 +105,10 @@ exchanges).
 |Displays arbitrary application info.
 |Yes
 
+|`integrationgraph`
+|Exposes the Spring Integration graph.
+|No
```
TYsewyn: @snicoll Can you update this in your polishing commit to indicate that it will be enabled by default? Thanks!

snicoll: Yup, thanks for the ping.

snicoll:
>  you have addressed all the concerns

I did not I think. I wish that the `IntegrationGraphServer` wasn't created at all if the endpoint is disabled or when Spring Integration is not enabled.

@artembilan how can I know that SI is enabled on a project? Is there a typical bean I could look for?

artembilan:
@snicoll ,

I think you can create `IntegrationGraphServer` conditionally when an endpoint is enabled.
On the other hand our main `IntegrationAutoConfiguration` relies on the `@ConditionalOnClass(EnableIntegration.class)`. Maybe that would be enough here as well?
If the story is still about an excluded `IntegrationAutoConfiguration`, then we can consider a condition against the `IntegrationContextUtils.INTEGRATION_HEADER_CHANNEL_REGISTRY_BEAN_NAME` bean which is a type of `DefaultHeaderChannelRegistry`. That is one of those which is registered by that `@EnableIntegration`.

Hope that helps.

snicoll:
Thanks for the feedback @artembilan 

> I think you can create IntegrationGraphServer conditionally when an endpoint is enabled.

Actually I can't at the moment, so we'll have to change the condition if you want to make that happen.

snicoll:
I've created #12945 to improve the condition. If this is merged, I can prevent `IntegrationGraphServer` to be created when then endpoint is disabled.

snicoll:
@TYsewyn that's now merged in `master`. Thanks a lot!

TYsewyn:
Thanks everyone for the valuable feedback!

PR #12337, opened 2018-03-04
Remove workaround for JDK-8023130 in RunProcess

dreis2211:
Hi,

with JDK 8 being the baseline and JDK 7 not being supported anymore we could get rid of the workaround for a bug in `ProcessBuilder.inheritIO()` on windows machines.

Let me know what you think.

Cheers,
Christoph


wilkinsona:
Thanks very much. This change is now in master.

PR #12339, opened 2018-03-04
RabbitMQ: Add connectionNamePrefix property

garyrussell:
Useful to identify which application owns a connection on the RabbitMQ management console.

See: https://stackoverflow.com/questions/49089915/how-to-set-custom-name-for-rabbitmq-connection/49090085#49090085

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
@@ -17,6 +17,7 @@
 package org.springframework.boot.autoconfigure.amqp;
 
 import java.time.Duration;
+import java.util.concurrent.atomic.AtomicInteger;
 
 import com.rabbitmq.client.Channel;
 
@@ -42,6 +43,7 @@ import org.springframework.context.annotation.Import;
 import org.springframework.retry.backoff.ExponentialBackOffPolicy;
 import org.springframework.retry.policy.SimpleRetryPolicy;
 import org.springframework.retry.support.RetryTemplate;
+import org.springframework.util.StringUtils;
 
 /**
  * {@link EnableAutoConfiguration Auto-configuration} for {@link RabbitTemplate}.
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
```diff
@@ -109,6 +111,11 @@ public CachingConnectionFactory rabbitConnectionFactory(
 			map.from(connection::getMode).whenNonNull().to(factory::setCacheMode);
 			map.from(connection::getSize).whenNonNull()
 					.to(factory::setConnectionCacheSize);
+			String connectionNamePrefix = properties.getConnectionNamePrefix();
+			if (StringUtils.hasText(connectionNamePrefix)) {
+				final AtomicInteger connectionNumber = new AtomicInteger();
+				factory.setConnectionNameStrategy(f -> connectionNamePrefix + "#" + connectionNumber.getAndIncrement());
```
snicoll: Rather than having an hardcoded strategy with the auto-incremented id, I'd very much prefer looking up for a `ConnectionNameStrategy` bean in the context. Does that sound sensible?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfigurationTests.java
```diff
@@ -66,7 +72,6 @@
  * @author Greg Turnquist
  * @author Stephane Nicoll
  * @author Gary Russell
- * @author Stephane Nicoll
```
snicoll: Why are you doing this to me? 😭

shakuzen: Looks like you are already listed 2 lines up. Double Stephane 😄

snicoll: Oh 😀

snicoll: I've removed my clone in 6f7501f

snicoll:
After discussing with @garyrussell we decided to take the route of detecting a `ConnectionNameStrategy` bean in the context.

snicoll:
Duplicate of #12367

PR #12405, opened 2018-03-08
Polish: Declarations should use Java collection interfaces such as "Deque" rather than specific implementation classes such as "LinkedList"

igor-suhorukov:


wilkinsona:
@philwebb Given that `RunArguments` is package-private, could this not got in 2.0.1?

philwebb:
Ah yes, I didn't notice that when reviewing in GitHub. Move to 2.0.1

PR #12411, opened 2018-03-08
Polish: to get rid of "magic" time constants

igor-suhorukov:


jhoeller:
FWIW, I like the idea of using `TimeUnit` that way... We had https://github.com/spring-projects/spring-framework/pull/1728 for the core framework along the same lines.

igor-suhorukov:
@jhoeller thank you for help :)

philwebb:
Yeah I like it as well, we already use the technique in a few places. For example [look at `Repackage`](https://github.com/spring-projects/spring-boot/blob/8109267de8a7a850ae8ab4f113eeee61c98c91a6/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java#L62)

PR #12422, opened 2018-03-09
Polish: private fields only used as local variables in methods should become local variables

igor-suhorukov:


wilkinsona:
This one applies to 1.5.x as well.

PR #12423, opened 2018-03-09
Polish: deprecated elements should have both the annotation and the Javadoc tag. 

igor-suhorukov:
Deprecated elements should have both the annotation and the Javadoc tag. Expose alternative to deprecated method.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyServletWebServerFactory.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyServletWebServerFactory.java
@@ -21,6 +21,7 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
+import java.net.URI;
 import java.net.URL;
 import java.nio.channels.ReadableByteChannel;
 import java.nio.charset.Charset;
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyServletWebServerFactory.java
```diff
@@ -577,12 +578,26 @@ public long length() {
...
 			return this.delegate.getURL();
 		}
 
+		/**
+		 * URI representing the resource.
+		 * @return an URI representing the given resource
+		 */
+		@Override
+		public URI getURI() {
+			return this.delegate.getURI();
```
wilkinsona: I don't think this is necessary. The overriden method does this:

```java
    public URI getURI()
    {
        try
        {
            return getURL().toURI();
        }
        catch(Exception e)
        {
            throw new RuntimeException(e);
        }
    }
```

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyServletWebServerFactory.java
```diff
@@ -577,12 +578,26 @@ public long length() {
 			return this.delegate.length();
 		}
 
+		/**
```
wilkinsona: This is a private class so there's no need for javadoc.

wilkinsona:
Thanks for the PR, @igor-suhorukov, but, for the reasons mentioned in my comments, I don't think we should merge this one.

igor-suhorukov:
@wilkinsona thank you for review and comments!

PR #12424, opened 2018-03-09
Polish: reorder the modifiers to comply with the Java Language Specification

igor-suhorukov:

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/util/LambdaSafe.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/util/LambdaSafe.java
@@ -100,7 +100,7 @@ public final class LambdaSafe {
 	/**
 	 * Abstract base class for lambda safe callbacks.
 	 */
-	private static abstract class LambdaSafeCallback<C, A, SELF extends LambdaSafeCallback<C, A, SELF>> {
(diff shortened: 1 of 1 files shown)


igor-suhorukov:
@snicoll Thank you for review and comments! In spring boot "main" classes it is only single occurrence of "static abstract"

wilkinsona:
We seem to be rather inconsistent:

```
$ find . -name "*.java" | xargs grep "static abstract" | wc -l
      12
$ find . -name "*.java" | xargs grep "abstract static" | wc -l
      16
```

Irrespective of whether it's test or main code, it'd be nice to consistently use one or the other. My vote is to align main and test code with the JLS as proposed (for main code only) here. I'm also in favour of enabling Checkstyle's modifier order check so that things remain consistent.

igor-suhorukov:
@wilkinsona this sounds reasonable with checkstyle rule and align rule with whole codebase. I have only one issue: my fork for this pull request is deleted. What I need to do?

wilkinsona:
> I can create another PR

I think that might be what you have to do. I don't know how to resurrect a pull request once the fork that it comes from has been deleted.

igor-suhorukov:
@wilkinsona @snicoll thank you for help! I fixed issues from this PR  in [PR 12432](https://github.com/spring-projects/spring-boot/pull/12432)

PR #12428, opened 2018-03-10
Refactor spring-boot JSON parsers

atelechev:
_This is my first attempt to submit a pull request in spring-boot project._

The changes in this pull request represent a minor refactoring, mainly aiming to eliminate a few code duplications.

While browsing the source code of `org.springframework.boot.json` package, I noticed that the implementations of `JsonParser` interface contain repetitions of code portions resembling the following:

    if (json != null) {
        json = json.trim();
        if (json.startsWith("{")) {
            return {JSON parser call};
        }
    }
    throw new IllegalArgumentException("Cannot parse JSON");

This concerned `BasicJsonParser`, `GsonJsonParser` and `YamlJsonParser` classes, so totally six repetitive blocks.

I created the `AbstractJsonParser` class in the same package, which provides common routines for the three classes above and uses Java 8 features.

The duplicated calls throwing `IllegalArgumentException` in case of parse failure are now also redirected to 
`AbstractJsonParser`.

This refactoring did not require to introduce unit tests. Previously existing tests have not been changed and continue to pass.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/AbstractJsonParser.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/AbstractJsonParser.java
@@ -1,0 +1,91 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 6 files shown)


philwebb:
@atelechev Thanks very much for your first contribution to Spring Boot. This has now been merged into master.

I've made a few refinements in 90949669be923b3b9b7692b663b49e9dee6702ff, specifically:
* We don't use `assert` so I've removed those.
* I changed the trim method so it calls the function directly. This just removes the need for `Optional`.

I also decided to create a new exception to represent parsing errors and remove a little more duplication by adding a `tryParse` method.

I nice enhancement overall. Cheers!

atelechev:
Thank you very much for the review and the feedback!

PR #12436, opened 2018-03-12
Add notes how to install Spring Boot CLI on Windows using Scoop

andreysaksonov:
App manifest for Spring Boot CLI is available in scoop-extras now (https://github.com/lukesampson/scoop-extras/pull/838)

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/getting-started.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/getting-started.adoc
@@ -387,6 +387,24 @@ completion scripts are automatically registered with your shell.
 
 
 
+[[getting-started-scoop-cli-installation]]
+==== Windows Scoop Installation
+If you are on a Windows and use http://scoop.sh/[Scoop], you can install the Spring Boot
+CLI by using the following commands:
+
+[indent=0]
+----
+	> scoop bucket add extras
+	> scoop install springboot
+----
+
+Scoop installs `spring` to `~/scoop/apps/springboot/current/bin`.
+
+NOTE: If you do not see the app manifest, your installation of scoop might be out-of-date. In
+that case, run `scoop update` and try again.
+
+
+
 [[getting-started-cli-example]]
 ==== Quick-start Spring CLI Example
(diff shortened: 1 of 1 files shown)


philwebb:
Thanks for the PR. I'm not a Windows user so I've not used scoop before. Is it similar to homebrew on the mac?

My only concern with merging this is that we ideally need a way to keep things updated when we release a new version.

We currently have our [own homebrew tap repository](https://github.com/pivotal/homebrew-tap) which means we can immediately update the version on a release. We also have [SDKMan](http://sdkman.io/) which provides an API for us to call when we release a new version.

Perhaps we just need to update our release process to generate the `springboot.json` file used by scoop so that we can easily submit a PR.

With homebrew we have a [template file](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-cli/src/main/homebrew/springboot.rb) that gets up [updated with each release](https://github.com/spring-projects/spring-boot/blob/b88e3cb27e48c6443cfe376e5bcbc16ae09a02aa/spring-boot-project/spring-boot-cli/pom.xml#L302-L340).

andreysaksonov:
@philwebb yes, scoop is like homebrew for Windows, "taps" there called "buckets" (example: https://github.com/andreysaksonov/scoop-bucket I used prior to merge of springboot.json to well-known "extras" bucket).

Scoop has autoupdate feature (check "checkver" and "autoupdate" fields here: https://github.com/lukesampson/scoop-extras/pull/838/files#diff-4837d1242bcd34c44b8e505f5c8ca524R15), so until version pattern of spring-boot-cli on GitHub will change (unlikely) or repository layout will change (unlikely) this manifest should not require any maintenance to get up-to-date spring-boot-cli.

I will have a look to template (it is just resource processed by gradle, right?) and try to deliver the same for scoop.

philwebb:
@andreysaksonov The `checkver` and `autoupdate` fields look interesting thanks. 

> I will have a look to template (it is just resource processed by gradle, right?) and try to deliver the same for scoop.

Yes, except it's processed by Maven not Gradle.

andreysaksonov:
Replaced with https://github.com/spring-projects/spring-boot/pull/12541

PR #12448, opened 2018-03-12
Polish: method names should comply with a naming convention

igor-suhorukov:


philwebb:
Errgh. How did we miss that one :)

igor-suhorukov:
@philwebb sonar found it

PR #12449, opened 2018-03-12
Polish: by contract "Iterator.next()" methods should throw NoSuchElementException

igor-suhorukov:


philwebb:
I don't think this one is necessary because the underlying `Enumeration` will throw the `NoSuchElementException`.

PR #12460, opened 2018-03-13
Use DEFAULT_MESSAGE in AbstractHealthIndicator()

izeye:
This PR changes to use `DEFAULT_MESSAGE` in `AbstractHealthIndicator()` as based on its Javadoc it looks intended to use `DEFAULT_MESSAGE`.


philwebb:
I think this actually fine as it is. It seems like `null` is a legitimate indication that the default message should be used and we check for that when we [log it](https://github.com/spring-projects/spring-boot/blob/ca7b2305218ca1790d9c1ba59256e8abf3369b88/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/AbstractHealthIndicator.java#L89).

Thanks anyway.

PR #12472, opened 2018-03-14
Polish: avoid check class by string name

igor-suhorukov:


philwebb:
Looks like the String based compare was added in 03a6f97e76e7b82deb6275cb6185a5b6ee1ad79d. @bclozel can you review? We hare changing the behavior if we merge this because the previous check was an exact instance match, not a `isAssignableFrom`.

bclozel:
Hello @igor-suhorukov 
I'd like to merge this but I can't checkout this change since the commit is not present on your fork anymore.

bclozel:
Never mind!
This was merged with 4c61d8abfe.
Thanks!

PR #12475, opened 2018-03-14
Replace Mockito.times(0) with Mockito.never()

dreis2211:
Hi,

this PR replaces occurences of `Mockito.times(0)` with `Mockito.never()`.

Let me know what you think.

Cheers,
Christoph


philwebb:
Nice one. I've merged this to master.

Issue #12447, opened 2018-03-12
URI tag of http.server.requests meter does NOT contain templated URI for jersey


jkschneider:
See https://github.com/micrometer-metrics/micrometer/issues/486 for a reproducible example.

wilkinsona:
This is an age-old problem that [I (overly optimistically) hoped would be taken care of by the move to Micrometer](https://github.com/spring-projects/spring-boot/issues/5875#issuecomment-321829008).

The meter's being registered by `WebMvcMetricsFilter`. Given the class's name and its focus on Spring MVC, that doesn't really make sense for Jersey. I think we should update the filter so that, if possible, it just ignores requests that aren't handled by Spring MVC. Perhaps we can no-op the filter if `getHandler(HttpServletRequest)` returns `null`. If we want to provide metrics for Jersey, that should be done in a separate filter.

mikksoone:
Thanks, the Bean provided in https://github.com/spring-projects/spring-boot/issues/5875#issuecomment-263294954 solves the issue, but yeah does not look so nice at all.

jkschneider:
@wilkinsona Note that we have complete Jersey 2 support in Micrometer with the `micrometer-jersey2` module, including a parameterized URI tag.

We just need to clean up shunting the `WebMvcMetricsFilter` if Jersey and autoconfiguring the Jersey support when relevant.

jkschneider:
@mikksoone If your app is wholly Jersey 2 and not a mixture of Jersey and WebMVC, I think a sufficient workaround for now is to disable the webmvc filter with `management.metrics.web.server.auto-time-requests=false`, and then configure the Jersey integration with something like [this](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/jersey2/server/JerseyServerMetricsConfiguration.java).

mikksoone:
@jkschneider disabling ```management.metrics.web.server.auto-time-requests=false``` is indeed disabling it, but configuring the Jersey according to the micrometer-spring-legacy did not work out of the box (no ```http.server.requests``` anymore). I'm using Spring Boot 2.0, mentioning it just in case.

mweirauch:
The referenced PR auto-configures the Jersey2 instrumentation.

[...]

PR #12482, opened 2018-03-14
Auto-configure Jersey2 server instrumentation

mweirauch:
This PR auto-configures the Micrometer Jersey2 server instrumentation. It's adopted from the `micrometer-spring-legacy` auto-configuration.

Relates to gh-12447 and micrometer-metrics/micrometer#486

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -94,6 +94,11 @@
 		</dependency>
 		<dependency>
 			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-jersey2</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>io.micrometer</groupId>
 			<artifactId>micrometer-registry-atlas</artifactId>
 			<optional>true</optional>
 		</dependency>
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/jersey2/server/JerseyServerMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/jersey2/server/JerseyServerMetricsAutoConfiguration.java
@@ -29,6 +29,7 @@ import org.glassfish.jersey.server.ResourceConfig;
 import org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration;
 import org.springframework.boot.actuate.autoconfigure.metrics.export.simple.SimpleMetricsExportAutoConfiguration;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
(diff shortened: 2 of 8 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/test/MetricsRun.java
```diff
@@ -76,7 +77,8 @@
 			RabbitMetricsAutoConfiguration.class, CacheMetricsAutoConfiguration.class,
 			DataSourcePoolMetricsAutoConfiguration.class,
 			RestTemplateMetricsAutoConfiguration.class,
-			WebFluxMetricsAutoConfiguration.class, WebMvcMetricsAutoConfiguration.class);
+			WebFluxMetricsAutoConfiguration.class, WebMvcMetricsAutoConfiguration.class,
+			JerseyServerMetricsAutoConfiguration.class);
```
mweirauch: This is just for completeness sake in case somebody is going to use the `MetricsRun` tooling in other tests. It's not used in the Jersey auto-configuration tests.

MarkusKull:
Did a quick test by copying this configuration into a jersey application with prometheus: Some uri-tags are now indeed templated. But raw uris are additionally recorded, probably because both MVC- and Jersey-RequestFilters apply to a http request. As quick workaround the jersey-meters were renamed and MVC-meters were suppressed by a MeterFilter. 
A better fix could simply disable the MVC-Filter when encountering Jersey. Or MVC should omit recording when seeing a RequestAttribute set by Jersey-Filter.

mweirauch:
Yap. WebMVC instrumentation kicks in. One could disable `WebMvcMetricsAutoConfiguration` in that case. Alternatively configure Jersey to work in servlet-filter mode and set it's filter order to `-2147483648`. That at least work in Spring-Boot 1.5. (Haven't tested Jersey pass-through of requests it doesn't handle, though.)

philwebb:
We might rework this depending on how the implementation of #13064 goes.

snicoll:
There is also another implementation for the same feature in #12905 that we should review as part of merging this.

mweirauch:
After some time has passed, what's your stance on this PR? #13064 hasn't seen any progress as of yet.

We are slowly moving to Spring Boot 2 services and I remembered this being still open. It's not hard to come up with a custom/proprietary setup for the time being, but it would be cool to not have to care about Jersey instrumentation in the future. (Even if it might get reworked with #13064.)

mweirauch:
Just in case, I went ahead and updated the PR against master/2.1.x.

Contains parts of #12905 and credited @michael-simons where appropriate. (Hope this is ok dear "Namensvetter" ;))

wilkinsona:
I think we should merge this one rather than #12905, with a single squashed commit amended to list @michael-simons [as a co-author](https://help.github.com/articles/creating-a-commit-with-multiple-authors/).

wilkinsona:
There are some differences in Micrometer's Jersey tag support and what it and Boot offer for Web MVC. I've opened https://github.com/micrometer-metrics/micrometer/issues/900 to try and align things. I don't think it needs to block merging this.

wilkinsona:
@mweirauch, thank you for making your first contribution to Spring Boot. @michael-simons, thanks for your contribution here too. You are a co-author of dd126fa. The proposed changes are now in master. I polished things a little in https://github.com/spring-projects/spring-boot/commit/72e2313fe84eed695e1e525722f0e05796c51cbb. Other than some repackaging, the main changes were to a meter filter and to reuse the `management.metrics.web.server` properties rather than adding Jersey-specific properties.

mweirauch:
Thanks @wilkinsona for taking care of this!

PR #12487, opened 2018-03-15
Polish: instance methods should not write to "static" fields

igor-suhorukov:

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/ConditionEvaluationDeltaLoggingListener.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/ConditionEvaluationDeltaLoggingListener.java
@@ -33,12 +33,16 @@ import org.springframework.context.ApplicationListener;
 class ConditionEvaluationDeltaLoggingListener
 		implements ApplicationListener<ApplicationReadyEvent> {
 
-	private final Log logger = LogFactory.getLog(getClass());
+	private static final Log logger = LogFactory.getLog(ConditionEvaluationDeltaLoggingListener.class);
 
 	private static ConditionEvaluationReport previousReport;
 
 	@Override
 	public void onApplicationEvent(ApplicationReadyEvent event) {
+		logConditionEvaluationDelta(event);
+	}
+
+	private static synchronized void logConditionEvaluationDelta(ApplicationReadyEvent event) {
 		ConditionEvaluationReport report = event.getApplicationContext()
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/ConditionEvaluationDeltaLoggingListener.java
```diff
@@ -33,27 +33,31 @@
...
+	private static final Log logger = LogFactory.getLog(ConditionEvaluationDeltaLoggingListener.class);
 
 	private static ConditionEvaluationReport previousReport;
 
 	@Override
 	public void onApplicationEvent(ApplicationReadyEvent event) {
+		logConditionEvaluationDelta(event);
+	}
+
+	private static synchronized void logConditionEvaluationDelta(ApplicationReadyEvent event) {
```
wilkinsona: Why is this now `synchronized`?

spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/HateoasObjenesisCacheDisabler.java
```diff
@@ -46,26 +46,26 @@ public void afterPropertiesSet() {
 		disableCaching();
 	}
 
-	private void disableCaching() {
+	private static synchronized void disableCaching() {
```
wilkinsona: Why is this now `synchronized`?

igor-suhorukov:
@wilkinsona thanks for review. I fix it

philwebb:
I'm not sure I totally understand the logic behind this one. Looking at the `HateoasObjenesisCacheDisabler` change for example, the static field is still being updated by an instance method. It's just now we've made a few more methods static to do it.

igor-suhorukov:
@philwebb thank you for feedback! You are right. Main idea was separate static methods from instance methods. Do you know more correct approach to save state and compare it between context restart?

wilkinsona:
I’m not sure that there’s anything incorrect about the current approach. Being static just affects the scope of the variable and that ClassLoader-wide scope is exactly what we want. There’s no harm in instance methods writing to that scope, particularly in this case where we know there’s only a single instance at a time.

philwebb:
I'm inclined to agree with Andy. I think for these two example we're best off with the code as it is.

philwebb:
Thanks anyway for the suggestion.

igor-suhorukov:
@philwebb @wilkinsona thank you for your time

PR #12492, opened 2018-03-15
Fix javadoc

tsachev:


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ServiceLevelAgreementBoundary.java
```diff
@@ -26,7 +26,7 @@
 
 /**
  * A service level agreement boundary for use when configuring micrometer. Can be
- * specified as either a {@link Long} (applicable to timers and distribution summaries) or
+ * specified as either a {@link Duration} (applicable to timers and distribution summaries) or
```
snicoll: Looking at the code, you fix the wrong `Long`. `Duration` only works with timers.

PR #12495, opened 2018-03-15
Use ApplicationContextRunner in ConditionalOnNotWebApplicationTests

dreis2211:


bclozel:
Thanks again @dreis2211 !

PR #12527, opened 2018-03-16
Polish: follow project convention for lambda expression

igor-suhorukov:


wilkinsona:
Thanks once again, @igor-suhorukov.

PR #12528, opened 2018-03-16
Polish: iterate map by using lambda function

igor-suhorukov:


philwebb:
Another good one!! Thanks @igor-suhorukov

PR #12535, opened 2018-03-18
Fix ModifiedClassPathRunner tests if run via IDEA

dreis2211:
Hi,

while looking into some possible performance improvements for `OnBeanCondition` I noticed `OnBeanConditionTypeDeductionFailureTests` to fail when run via IntelliJ. After a bit of investigation it turned out to be a short-coming of `ModifiedClassPathRunner` instead of a broken change of mine.

IntelliJ can shorten the classpath to a single classpath.jar in order to circumvent errors originating from a too long classpath. This breaks the filtering inside ModifiedClassPathRunner as the classpath contains only a single jar to match against. This can be fixed by applying a similar mechanism already provided for Surefire manifest-only booter JARs, which extracts the real classpath from the JAR's Manifest file.

Let me know what you think.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/runner/classpath/ModifiedClassPathRunner.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/runner/classpath/ModifiedClassPathRunner.java
@@ -28,6 +28,7 @@ import java.util.Collections;
 import java.util.List;
 import java.util.jar.Attributes;
 import java.util.jar.JarFile;
+import java.util.regex.Pattern;
 import java.util.stream.Stream;
 
 import org.apache.maven.repository.internal.MavenRepositorySystemUtils;
@@ -65,6 +66,9 @@ import org.springframework.util.StringUtils;
  * @author Andy Wilkinson
  */
 public class ModifiedClassPathRunner extends BlockJUnit4ClassRunner {
+
+	private static final Pattern INTELLIJ_CLASSPATH_JAR_PATTERN = Pattern.compile(
+			".*classpath(\\d+)?.jar");
 
 	public ModifiedClassPathRunner(Class<?> testClass) throws InitializationError {
 		super(testClass);
@@ -134,7 +138,19 @@ public class ModifiedClassPathRunner extends BlockJUnit4ClassRunner {
 	}
 
 	private boolean isShortenedIntelliJJar(URL url) {
-		return url.getPath().endsWith("classpath.jar");
+		String urlPath = url.getPath();
+		boolean isCandidate = INTELLIJ_CLASSPATH_JAR_PATTERN.matcher(urlPath).matches();
+		if (isCandidate) {
+			try {
+				Attributes attributes = getManifestMainAttributesFromUrl(url);
+				String createdBy = attributes.getValue("Created-By");
+				return createdBy != null && createdBy.contains("IntelliJ");
+			}
+			catch (Exception e) {
+				return false;
+			}
+		}
+		return false;
 	}
 
 	private List<URL> extractUrlsFromManifestClassPath(URL booterJar) {
@@ -151,9 +167,14 @@ public class ModifiedClassPathRunner extends BlockJUnit4ClassRunner {
 	}
 
 	private String[] getClassPath(URL booterJar) throws Exception {
-		try (JarFile jarFile = new JarFile(new File(booterJar.toURI()))) {
-			return StringUtils.delimitedListToStringArray(jarFile.getManifest()
-					.getMainAttributes().getValue(Attributes.Name.CLASS_PATH), " ");
+		Attributes attributes = getManifestMainAttributesFromUrl(booterJar);
+		return StringUtils.delimitedListToStringArray(attributes
+				.getValue(Attributes.Name.CLASS_PATH), " ");
+	}
+
+	private Attributes getManifestMainAttributesFromUrl(URL url) throws Exception {
+		try (JarFile jarFile = new JarFile(new File(url.toURI()))) {
+			return jarFile.getManifest().getMainAttributes();
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


philwebb:
It seems like a sensible change to me. It's a shame that `classpath.jar` is such a generic name but I doubt there's a better way to detect it.

dreis2211:
Indeed, it's sadly very generic. Another option would be to check for properties of the Manifest file which seems to contain an entry `Created-By:IntelliJ IDEA` which would involve a second access to the JAR file (if done somewhat "clean"). I tried this already locally and it looked roughly like this

```
	private boolean isShortenedIntelliJJar(URL url) {
		boolean isCandidate = url.getPath().endsWith("classpath.jar");
		if (isCandidate) {
			try {
				Attributes attributes = getManifestMainAttributesFromUrl(url);
				String createdBy = attributes.getValue("Created-By");
				return createdBy != null && createdBy.contains("IntelliJ");
			} catch (Exception e) {
				return false;
			}
		}
		return false;
	}
```
This involves the following change 
```diff
        private String[] getClassPath(URL booterJar) throws Exception {
-               try (JarFile jarFile = new JarFile(new File(booterJar.toURI()))) {
-                       return StringUtils.delimitedListToStringArray(jarFile.getManifest()
-                                       .getMainAttributes().getValue(Attributes.Name.CLASS_PATH), " ");
+               Attributes attributes = getManifestMainAttributesFromUrl(booterJar);
+               return StringUtils.delimitedListToStringArray(attributes
+                               .getValue(Attributes.Name.CLASS_PATH), " ");
+       }
+
+       private Attributes getManifestMainAttributesFromUrl(URL url) throws Exception {
+               try (JarFile jarFile = new JarFile(new File(url.toURI()))) {
+                       return jarFile.getManifest().getMainAttributes();
                }
        }
```

I guess there won't be too much - if any - `classpath.jar` files in the wild that would cause a performance drop here, so I don't mind adding this to the changeset if the general idea is accepted by the team.

Cheers,
Christoph

dreis2211:
It turns out that IntelliJ 2018.x seems to include a random number in the filename. See https://github.com/JetBrains/intellij-community/blob/bba6edf05bbf6d7dbac61fdd0f79d8e1bb3f021c/platform/util/src/com/intellij/execution/CommandLineWrapperUtil.java#L51

In order to account for that I'd change `isShortenedIntelliJJar` to make use of a regex:
```
	private static final Pattern INTELLIJ_CLASSPATH_JAR_PATTERN = Pattern.compile(
			".*classpath(\\d+)?.jar");

... 

	private boolean isShortenedIntelliJJar(URL url) {
		String urlPath = url.getPath();
		boolean isCandidate = INTELLIJ_CLASSPATH_JAR_PATTERN.matcher(urlPath).matches();
		if (isCandidate) {
			try {
				Attributes attributes = getManifestMainAttributesFromUrl(url);
				String createdBy = attributes.getValue("Created-By");
				return createdBy != null && createdBy.contains("IntelliJ");
			}
			catch (Exception e) {
				return false;
			}
		}
		return false;
	}
```

akozlova:
Hi guys, 

IntelliJ always added numbers to the created temp files to avoid conflicts between create concurrently processes. Please, don't match on the temp file name, it's an internal information and we never thought that we would break something if would rename these files. I can add custom manifest attribute if it would help. Thanks

dreis2211:
@akozlova The following [commit ](https://github.com/JetBrains/intellij-community/commit/0b4dd7ca6eb6b4bb204a006f1259619a5466b2ff#diff-af5a65366579cb739d066ccfbc8e5203) is from November 2017. For me this is fairly new. Could you elaborate what you mean?

I'd also like to add that the `classpath.jar` we are talking about is displayed in the menus as such and actually described in IntelliJs very own [blogs](https://blog.jetbrains.com/idea/2017/10/intellij-idea-2017-3-eap-configurable-command-line-shortener-and-more/). For me this is not an implementation detail at least. Love to hear your opinion on that. 

As the tests in the real CI pipeline are run differently anyways, my conceirns are not too high. My proposal just enhances the experience when contributing to Spring-Boot and running quick checks in IntelliJ. So I'd argue that a rename of the classpath.jar from IntelliJ side would only break this best-effort and nothing mission-critical (at least from an outside point-of-view).

What do you think, @philwebb or @snicoll - should I proceed by making the mentioned adjustments to my initial patch?

akozlova:
IDEA adds pseudo random prefix/suffixes in the com.intellij.openapi.util.io.FileUtilRt#doCreateTempFile based on files in the temp directory.  The latest commit made it predictable (and always there) but generally it was always there.

Actually, there are no mentions that there would be the jar with this name, it's the format of the classpath as -help would output the expected parameters of some process.

dreis2211:
> JAR Manifest: The IDE passes the long classpath via a temporary classpath.jar. The Original classpath is defined inside MANIFEST.MF as a Class-Path attribute in classpath.jar.

And again. The menu shows it as well.
screen-shot-2017-10-09-at-08 58 33

I get your point though, that this is not a perfect solution. I agree actually. Yet, I only plan to take the classpath(some-optional-suffix).jar name matching in order to **pre-filter** the candidates for the manifest lookup, which already contains a reference to IntelliJ (see my comment about the `Created-By` attribute).

I'd like to stretch again that this is just a best-effort to enhance the experience for people that run Spring-Boot tests via IntelliJ once in a while, so not a general problem for Spring-Boot users, but rather for contributors. If the manifest lookup fails or no classpath.jar candidate is found, it doesn't break but simply walks over it, so it's behaving as it does without my proposal.

One could argue that an additional custom attribute in the manifest is an implementation detail as well, by the way. So is the lookup for the Created-By attribute. And so is the pre-existing check for surefirebooter JARs. There are no guarantees that this is never renamed either. If something changes, one can always adjust to the changes.

Let me know what you think.

dreis2211:
@snicoll & @philwebb I included the changes of my comments. Let me know if/when I should squash

snicoll:
Fixed by 345b2c5

PR #12550, opened 2018-03-19
Add auto-configuration for Hibernate metrics

ruifigueira:
Hibernate statistics can be gathered by metrics actuator by exposing a micrometer `HibernateMetrics` bean, but it is not auto-configured.

This PR provides an auto-configuration class that binds `HibernateMetrics` for each entity manager factory bean available, in the same way as `DataSourcePoolMetricsAutoConfiguration` does for data sources (most of the code is based on it, btw).

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -173,6 +173,11 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.hibernate</groupId>
+			<artifactId>hibernate-core</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>net.sf.ehcache</groupId>
 			<artifactId>ehcache</artifactId>
 			<optional>true</optional>
@@ -438,6 +443,11 @@
 		<dependency>
 			<groupId>org.skyscreamer</groupId>
 			<artifactId>jsonassert</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.springframework</groupId>
+			<artifactId>spring-orm</artifactId>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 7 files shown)


pivotal-issuemaster:
@ruifigueira Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12550)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12550) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ruifigueira Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12550)!

snicoll:
@ruifigueira thanks for the PR. I saw your check for Hibernate in the PR and I believe something in micrometer can be done for this (or the constructor to be changed).

I've raised https://github.com/micrometer-metrics/micrometer/issues/518

ruifigueira:
@snicoll, I think the micrometer team is following the [specs](https://docs.oracle.com/javaee/7/api/javax/persistence/EntityManagerFactory.html#unwrap-java.lang.Class-), because they are catching the `PersistenceException` that is thrown, as stated in the javadocs.

My point there was more to prevent that badly implemented `EntityManagerFactory` (or just mocks) could make it fail badly, if `unwrap` method returns a `null` instead of throwing a `PersistenceException`.

snicoll:
Thanks for the feedback. I still think that check should go in micrometer.

ruifigueira:
@snicoll, I saw the update in https://github.com/micrometer-metrics/micrometer/issues/518, I basically removed that unnecessary check and updated the corresponding test.

ruifigueira:
I ended up doing a squash of those last two commits, basically the second one reverted changes in `HibernateMetricsAutoConfiguration` from the first commit, so this squashed commit only has test improvements.

snicoll:
@ruifigueira thanks, this is now merged in `master`.

PR #12553, opened 2018-03-19
Polish: diamond operator and StringBuilder

igor-suhorukov:
simplify code with diamond operator, use StringBuilder in single thread logic


snicoll:
Thanks again @igor-suhorukov!

PR #12554, opened 2018-03-20
Use this.out in SummaryProgressReporter.finished()

izeye:
This PR changes to use `this.out` in `SummaryProgressReporter.finished()` as it looks intended to be it, not `System.out`.


wilkinsona:
Well spotted, @izeye. `SummaryProgressReporter` is only ever created with `System.out` so this change won't make any functional difference, but you're right that `this.out` should be used there.

wilkinsona:
Thanks once again, @izeye.

PR #12555, opened 2018-03-20
Rename expose to include

izeye:
It looks missed in 7473642f5852900166d4dfa754b9362612e2db32.


snicoll:
good catch again, @izeye. Thanks!

PR #12559, opened 2018-03-20
Update 'propdeps-plugin' link to new Github repo

fabriziocucci:


snicoll:
@wilkinsona @rwinch why do we have that plugin in two places exactly?

wilkinsona:
Good question, @snicoll. I don't think we should. Thanks for the PR, @fabriziocucci, but I don't think we should be recommending the use of the propdeps plugin any more. I've opened https://github.com/spring-projects/spring-boot/issues/12573.

fabriziocucci:
@wilkinsona at least the PR wasn't for nothing! :)

PR #12574, opened 2018-03-21
use simple evaluation context in errorMvcAutoConfig

### LINKED ISSUE #12507: Switch ErrorMvcAutoConfiguration to SimpleEvaluationContext
`SimpleEvaluationContext` has better performance so we should switch to it.

htynkn:
Close #12507

SimpleEvaluationContext supports only a subset of the SpEL language syntax which has better performance. In SpELView, SimpleEvaluationContext is enough to support binding expressions.

SimpleEvaluationContext doesn't support a default root object. so we need provide object when using Expression.


philwebb:
I'd like to apply this to 1.5.x and unfortunately I think it's going to be easier to re-implement than deal with the merge conflicts. Thanks anyway.

PR #12588, opened 2018-03-22
Return the metric names in sorted order

mveitas:
It's helpful to have the metric names in a sorted order when visually inspecting them


wilkinsona:
Thanks for the suggestion, but I don’t think serverside sorting of the names is a good idea. If there are a large number of names, sorting them may be reasonably costly and when done on the serverside that cost is paid by everyone. I’d recommend that you sort the names on the client side in the cases where such sorting is useful.

Issue #11929, opened 2018-02-06
ApplicationContextRunner javadoc refers to a package protected class


snicoll:
This leads to a [404 in generated javadoc](https://docs.spring.io/spring-boot/docs/2.0.0.RC1/api/org/springframework/boot/test/context/runner/ApplicationContextRunner.html)

wilkinsona:
It's a bit worse than just being a javadoc problem. `ApplicationContextRunner` extends `AbstractApplicationContextRunner` and the latter is package-private. IMO, a public class shouldn't have any package-private classes in its type hierarchy. I think we should make `AbstractApplicationContextRunner` public.

PR #12590, opened 2018-03-22
Make intermediate classes in type hierarchy of public class public

izeye:
Based on https://github.com/spring-projects/spring-boot/issues/11929#issuecomment-363548571, there are a few classes which don't align with it. This PR tries to align them with it.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/AbstractNestedCondition.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/AbstractNestedCondition.java
@@ -40,8 +40,9 @@ import org.springframework.util.MultiValueMap;
  * Abstract base class for nested conditions.
  *
  * @author Phillip Webb
+ * @since 2.0.1
  */
-abstract class AbstractNestedCondition extends SpringBootCondition
+public abstract class AbstractNestedCondition extends SpringBootCondition
 		implements ConfigurationCondition {
 
 	private final ConfigurationPhase configurationPhase;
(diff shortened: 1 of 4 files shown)


philwebb:
Not sure I totally agree that a package private class shouldn't be part of the hierarchy. I've flagged this one for team discussion.

wilkinsona:
My reasoning is that if you indirectly extend a package-private class then you are inheriting from and are exposed to Boot's internals. I don't think we should inflict that exposure on people. Gradle's API has a number of classes where internals leak out in this manner and it's been the main source of compatibility problems with both Boot's plugin and the dependency management plugin.

izeye:
For example, `AbstractNestedCondition` is package-private but has a protected method and a protected nested class, so it's open for its children's inheritance unless its all children are final: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/AbstractNestedCondition.java#L68-L102

And also the protected nested class, `AbstractNestedCondition.MemberMatchOutcomes` is exposed in "Nested Class Summary" section in Javadoc as a broken link: https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/api/org/springframework/boot/autoconfigure/condition/AllNestedConditions.html

snicoll:
I felt the same way as @philwebb initially but @wilkinsona shared that compelling argument previously. I think it would be more consistent to make them public.

philwebb:
We discussed this on our call today and I've been convinced. Will merge this in 2.0.x.

philwebb:
We need to check the `@since` tags when we merge.

snicoll:
@philwebb `@since` tags look good. Thanks again @izeye!

PR #12594, opened 2018-03-22
Fix @EnableIntegrationManagement

garyrussell:
Incorrect property set to `true` - `countsEnabled` is an array of bean name patterns.
`defaultCountsEnabled` is the default property.

Inadvertently changed when removing `defaultStatsEnabled` for Micrometer support.


snicoll:
That has changed behaviour so I think it qualifies as a regression. Let's make sure we fix this for 2.0.1 (we need a test for that stuff as well).

Issue #12580, opened 2018-03-21
Cookie http-only setting has no effect when using embedded Tomcat


maku:
I am using spring boot 2.0.0.RELEASE

When I set the http-only flag to false it has no effect on the browser. I am able to inspect the session cookie in the browser -> I see the http flag is enabled. And therefore I am not able to access the cookie via javascript...

Here is the definition part in my application.yaml

```
server:
  servlet:
    session:
      cookie:
        max-age: 30m
        name: my_app_session
        http-only: false
        secure: false
      timeout: 30m
```

The reason why I do this is: I have an SPA where I would like to check the session cookie periodically to 
figure out when the session is expired (to present a certain view to the user)

wilkinsona:
Are you using Tomcat? It appears that it requires the setting to be applied to its `Context` as well as to the `SessionCookieConfig`. Boot does the latter but does not currently do the former.

maku:
@wilkinsona I am using the default (which means tomcat). But I have **NO** idea how to set it via the Context.
I implemented a ServletContextInitializer. Here the SessionCookieConfig is set correctly.

What do I have to do additionally?

BTW, when spring boot provides cookie config properties I would assume it handles **ALL** required steps. Now I fight already many hours to solve this problem!

maku:
@wilkinsona Is there a way to handle this problem in springboot 2.0 ?? 

how can I access the underlying tomcat instance to fix the behaviour? (maybe switching to jetty can help?)

wilkinsona:
You can use a customiser to configure the context:

```java
@Bean
public WebServerFactoryCustomizer<TomcatServletWebServerFactory> tomcatCustomizer() {
	return (tomcat) -> tomcat
			.addContextCustomizers((context) -> context.setUseHttpOnly(false));
}
```

> I would assume it handles ALL required steps.

There's a fair assumption and is what we need to fix so that the customiser above is not necessary.

maku:
@wilkinsona great, thank you with the customiser for tomcat it works....

hellowenqi:
Hi, Cookie http-only setting has no effect when using spring boot 2.1.0.RELEASE.

I also tried the method of @wilkinsona !

PR #12597, opened 2018-03-22
Setting http-only to false has no effect when using embedded Tomcat

kedar-joshi:
Fixes gh-12580


wilkinsona:
Thanks for the PR. I know that it's necessary to configure HTTP only on the context as well as on the session cookie config, but I'm not sure that's the case for the other properties. Have you verified Tomcat's behaviour? We also need some tests to verify that the problem's been fixed and isn't regressed in the future.

kedar-joshi:
Yes, I have verified the behaviour and adding the tests for the same, as we speak.

kedar-joshi:
> but I'm not sure that's the case for the other properties

Yes, only `HttpOnly` needs to be set. Updated the PR for the same.

rfshelley:
What are the chances of getting this merged into a snapshot? Is there a good work around in the mean time? The WebServerFactoryCustomizer in 12580 doesn't seem to work for me as it throws an error "No ServletContext set" so I'm hoping to get a version with the preferred patch. Thanks!

philwebb:
Turns out applying this to 1.5.x is a little more involved so I'm going to take a slightly different approach. Thanks anyway.

PR #12611, opened 2018-03-23
Issue 7447

Buzzardo:
Per issue-7447, I have removed Docbook from the documet-production tool chain.

I also added a collapsible Table of Contents for all the HTML files, changed the styles to match Spring Framework, added an index page for the multi-file HTML version, and fixed a couple of casing issues (significant words starting with lower case in headings) that I saw along the way. A side effect of moving to Asciidoctor was a substantial reduction in the number of HTML files, because we no longer chunk each file. The collapsible ToC makes the larger HTML files more usable, so it should work out well for our readers.


philwebb:
I don't know why I added this to 2.0.x, it should be in the backlog and something we consider for 2.1

bclozel:
The [shared asciidoctor theme has been started](https://github.com/spring-projects/spring-doc-resources), we need to first submit a few PRs there and then publish it as an artifact.

Buzzardo:
I fixed the merge conflict (which I didn't realize I had, or I would have fixed it long ago). Sorry about that. This should be ready for prime time now.

Buzzardo:
> The [shared asciidoctor theme has been started](https://github.com/spring-projects/spring-doc-resources), we need to first submit a few PRs there and then publish it as an artifact.

I made a project for our shared doc resources: https://github.com/spring-projects/spring-doc-resources, so that the whole Spring portfolio would have a single place for the doc resources (so that we can change things like stylesheets in only one place).

The changes I made to Spring Boot get the doc resources from that project.

bclozel:
@Buzzardo Thanks! Let me review this PR. I’ll probably send a PR myself to the docs resources project because we need it to generate the css files from the original theme rather than storing them.

I’ll let you know if I’ve got questions about it!

spring-boot-project/spring-boot-docs/pom.xml
```diff
@@ -960,6 +969,31 @@
 	<profiles>
 		<profile>
 			<id>full</id>
+			<dependencies>
+				<dependency>
+					<groupId>org.springframework.doc-resources</groupId>
```
wilkinsona: This is quite an unconventional group id due to the `-` in it. I would have raised an issue for it in https://github.com/spring-projects/spring-doc-resources but issues appear to be disabled and there's no link to an alternative issue tracker in the README. Could it be changed pleased?

Buzzardo: Should it be org.springframework.doc.resources?

wilkinsona: I think `docresources` would be more conventional. It would also match other projects like REST Docs (restdocs) and Web Flow (webflow) too.

There’s also a case to be made for it to be `io.spring` rather than `org.springframework`. The latter is typically reserved for projects that are used by the community when developing an application, have a programmatic API, etc.

Buzzardo: I'll change it to io.spring.docresources once Brian has finished his review. We have four projects (and growing) that use this set of resources now, so I'll have some changes to make.

bclozel:
I've sent a work in progress PR on the doc resources project, see: spring-projects/spring-doc-resources#1

I'll update it with stylesheet changes to reflect the Spring Asciidoctor theme, as it's currently just a CSS reset. I'm working with @oodamien on that.

PR #12626, opened 2018-03-24
Polish: iterate map by using lambda function

igor-suhorukov:


philwebb:
Thanks for this one, it must have taken a while! I've merged to master but rolled back a couple where I though that the for loops were easier to read.

PR #12627, opened 2018-03-24
Distinguish between read and write operations in MVC actuator endpoints

rlippolis:
While upgrading from Spring Boot 1.5.10.RELEASE to 2.0.0.RELEASE, we noticed that the generated Swagger documentation for the Actuator endpoints contained some errors. Some of our users are using Swagger to easily call our REST endpoints, including some of the Actuator URLs. We discovered 2 issues in the generated API documentation:

1. The GET requests contained a request body
2. The path parameters in endpoint URLs were not recognized.

After some debugging, we found the cause for these issues, for more details, see [this sample project](https://github.com/rlippolis/spring-boot2-actuator-swagger).

This PR fixes the first issue: we have split the `org.springframework.boot.actuate.endpoint.web.servlet.AbstractWebMvcEndpointHandlerMapping.OperationHandler` into a `ReadOperationHandler` and a `WriteOperationHandler`. The read variant doesn't have the `@RequestBody` annotated method parameter, and thus fixes the first mentioned issue.

This approach is actually similar to the Reactive implementation in `org.springframework.boot.actuate.endpoint.web.reactive.AbstractWebFluxEndpointHandlerMapping`, which already distinguishes between a `ReadOperationHandler` and `WriteOperationHandler`.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/servlet/AbstractWebMvcEndpointHandlerMapping.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/servlet/AbstractWebMvcEndpointHandlerMapping.java
@@ -30,6 +30,7 @@ import javax.servlet.http.HttpServletResponse;
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.boot.actuate.endpoint.InvalidEndpointRequestException;
 import org.springframework.boot.actuate.endpoint.InvocationContext;
+import org.springframework.boot.actuate.endpoint.OperationType;
 import org.springframework.boot.actuate.endpoint.SecurityContext;
 import org.springframework.boot.actuate.endpoint.invoke.OperationInvoker;
 import org.springframework.boot.actuate.endpoint.web.EndpointMapping;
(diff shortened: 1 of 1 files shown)


wilkinsona:
I assume that when you refer to Swagger, you're using SpringFox to automatically generate the Swagger definition?

As the lead of REST Docs I'm biased on this one, but I'm not convinced that this is a good idea for a few reasons:

1. The WebFlux code is only structured that way to, IIRC, work around a limitation. It would be the same as the MVC code if that had been possible when it was written.
2. It adds complexity to our codebase to deal with the limitations of a third-party tool.
3. Path parameters still won't work as `@Selector` isn't understood.
4. There are other ways to produce a Swagger definition from which you can generate clients or an API playground.

rlippolis:
You are correct, I am using SpringFox to generate the Swagger definition.

Regarding the mentioned reasons:
1. I did not know that, but I agree that a single `OperationHandler` is the preferred way if it had been possible.
2. You are right about the fact that it feels like adding unnecessary complexity, but I want to argue that it also feels incorrect to have a GET request mapped to a handler method which includes an `@RequestBody` annotated method argument. But if you feel that this is acceptable (to be able to write less code), I will just have to accept that... :)
3. True, I am still trying to figure out whether this would have to be corrected in `springfox-spring-web`, in Actuator, or if this isn't a bug at all...
4. Of course, but I liked the way it worked automatically prior to Spring Boot 2.0 :)

Anyway, thanks for considering it!

wilkinsona:
We've discussed this today and, for the reasons [described above], we're going to decline this change. Thanks anyway.

PR #12628, opened 2018-03-24
Sort meter names in actuator metrics endpoint

AlexFalappa:
This PR makes the actuator `metrics` endpoint return meter names alphabetically sorted.


snicoll:
We've very recently declined a similar PR, see https://github.com/spring-projects/spring-boot/pull/12588#issuecomment-375267311

PR #12629, opened 2018-03-24
Use secure connection to repositories

oflebbe:
Spring Boot should use secure connections to its repositories.


wilkinsona:
Indeed it should. Thanks for the PR, @oflebbe.

snicoll:
There are more usage of http for repositories. I am reviewing them now.

PR #12633, opened 2018-03-26
Fix Kotlin @ConfigurationProperties example

sdeleuze:
`lateinit` should not be used on `val` properties initialized with a default value, see https://github.com/mixitconf/mixit/blob/master/src/main/kotlin/mixit/MixitProperties.kt for a concrete example.


snicoll:
Thanks @sdeleuze, I've noticed a couple of new `foo`, `bar` were introduced in the docs, I'll fix that in a separate commit.

PR #12636, opened 2018-03-26
Docs: persistence.xml - use clearer English

jmkgreen:
Old version references a "feature", new version references what Boot doesn't do, may be easier for humans (like me) to parse first time.


snicoll:
@jmkgreen thanks for the PR. The change is a bit too final IMO: if you define an `orm.xml` for provider-specific customizations, Spring Boot will pick it up because the Framework's default infrastructure we configure will pick it up. Can you rephrase your PR to target `META-INF/persistence.xml` then please?

jmkgreen:
@snicoll I'm hoping I have interpreted your request correctly :-)

snicoll:
@jmkgreen you did. I am not entirely sure that the note on `orm.xml` is required there but we'll have one more pass before merging. Thanks for the quick feedback!

Issue #12518, opened 2018-03-16
Move away from Couchbase's deprecated endpoint configuration methods


wilkinsona:
We're using the following methods that are now deprecated:

- `com.couchbase.client.core.env.DefaultCoreEnvironment.Builder.kvEndpoints(int)`
- `com.couchbase.client.core.env.DefaultCoreEnvironment.Builder.queryEndpoints(int)`
- `com.couchbase.client.core.env.DefaultCoreEnvironment.Builder.viewEndpoints(int)`

Each has a recommend replacement. We should move to them, making any necessary adjustments to our configuration properties when we do so. We can probably do this in 2.0.1 if we keep the current properties, deprecating where necessary, and add any additional properties that the replacements in Couchbase need.

qinnnyul:
@wilkinsona hey, I have one question about these replacements, some of them support more method parameters, e.g: QueryServiceConfig, it accepts minEndpoints and maxEndpoints. do we need to make adjustments to "CouchbaseProperties" to reflect this change?

wilkinsona:
Yes. That’s why I said “making any necessary adjustments to our configuration properties when we do so” above.

philwebb:
Closing in favor of PR #12655 (thanks @qinnnyul!)

PR #12655, opened 2018-03-28
Move away from Couchbase's deprecated endpoint configuration methods 

qinnnyul:
Replaced Couchbase's deprecated methods with alternative ones, meanwhile, makes them backwards compatibility #12518


snicoll:
That's right and, yes, the same for `view`. What's important here is that the `query` and `view` parameter still works (and be flagged as deprecated) so we need dedicated tests for that.

qinnnyul:
@snicoll could you please review these changes when you are free? thx

snicoll:
Thanks @qinnnyul, Your contribution is now merged in `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/6692301d51d22e1a427180c8fc7754d6ea8e6df8).

Issue #12229, opened 2018-02-26
Bean Validation fails when Hibernate Validator not in classpath


mikebell90:
Here's the issue. Up through Spring Boot 2.0.0.M7/ Spring 5.0.3 (our last stack), things worked fine.

Now upgrading to Spring Boot 2.0.0.RC/Spring 5.0.4 we get this:

```
Caused by: javax.validation.NoProviderFoundException: Unable to create a Configuration, because no Bean Validation provider could be found. Add a provider like Hibernate Validator (RI) to your classpath.
	at javax.validation.Validation$GenericBootstrapImpl.configure(Validation.java:291)
	at org.springframework.boot.validation.MessageInterpolatorFactory.getObject(MessageInterpolatorFactory.java:53)
	at org.springframework.boot.context.properties.ConfigurationPropertiesJsr303Validator$Delegate.<init>(ConfigurationPropertiesJsr303Validator.java:71)
	at org.springframework.boot.context.properties.ConfigurationPropertiesJsr303Validator.<init>(ConfigurationPropertiesJsr303Validator.java:43)
	at org.springframework.boot.context.properties.ConfigurationPropertiesJsr303Validator.getIfJsr303Present(ConfigurationPropertiesJsr303Validator.java:64)
	at org.springframework.boot.context.properties.ConfigurationPropertiesBinder.<init>(ConfigurationPropertiesBinder.java:69)
	at org.springframework.boot.context.properties.ConfigurationPropertiesBindingPostProcessor.afterPropertiesSet(ConfigurationPropertiesBindingPostProcessor.java:78)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.invokeInitMethods(AbstractAutowireCapableBeanFactory.java:1769)
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1706)
	... 41 more
```

Now, in theory it's TRUE - I do have only the validation api (2.0) on the class, and no implementation. This was however DELIBERATE - sometimes I have no interest in actually enforcing the validation, but I choose to use the API for annotations to conveniently document CONTRACT.

Is there a way to bypass this check? I understand the reason for it, but I think it's wrong headed forcing this.

wilkinsona:
Can you please provide a sample that works with M7 but fails with RC2? M7 also fails for me:

```
  .   ____          _            __ _ _
[...]

PR #12669, opened 2018-03-29
Allow validation api in the classpath without an implementation

geo-m:
Allows using validation api annotations for documentation / contract purposes. The original exception is now thrown only when actually requesting validation with `@Validated`.

Fixes Issue #12229


philwebb:
@geo-m Thanks very much for your first contribution to Spring Boot. This is in master and will be part of 2.0.1. I've made some minor tweaks in commit 0c98d0e7d3a8d092c409799092d3252f2e71d98c

PR #12672, opened 2018-03-29
Unicode parsing in OriginTrackedPropertiesLoader does not work with all code points

### LINKED ISSUE #12671: Parsing unicode escapes in OriginTrackedPropertiesLoader is broken
`OriginTrackedPropertiesLoader` tries to parse `\uXXXX`-unicode escapes, but the parsing is broken, resulting garbled data:

https://github.com/spring-projects/spring-boot/blob/9dea19d83c07abfec16935759008b4559148676b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedPropertiesLoader.java#L226-L235

Clearly `digit > -'0'` should be `digit >= '0'` and similarly for the other two branches.

komu:
Unicode escape parsing in `OriginTrackedPropertiesLoader.CharacterReader` was broken and produced garbled data. There was a test for functionality, but selected escape `\u0026` happened to work correctly. Fixed the parsing and added three new code points to test value.

Closes #12671

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedPropertiesLoader.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedPropertiesLoader.java
@@ -224,13 +224,13 @@ class OriginTrackedPropertiesLoader {
 			this.character = 0;
 			for (int i = 0; i < 4; i++) {
 				int digit = this.reader.read();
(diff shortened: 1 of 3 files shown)


philwebb:
Nice spot. Thanks!

philwebb:
We managed to get someone else reporting this issue (#12716), and I forgot we had this PR. Sorry about that. It's now fixed in a657a28f58e0eb8487eede9c60062186d850408d

Issue #12676, opened 2018-03-29
Reconsider auto-configuration of HierarchicalNameMapper


snicoll:
See https://github.com/spring-projects/spring-boot/issues/12557#issuecomment-376874690

`HierarchicalNameMapper` is not a component that can be shared amongst multiple monitoring systems. Right now the auto-configuration for Ganglia, Graphite, JMX and Statsd all create an instance if none exists in the context. 

Rather that adapting to the presence of such a bean in the context, we should probably go with whatever makes sense and offer another way to customize the mapper. The simplest would be to require the user to override the registry.

ping @jkschneider

snicoll:
Closing in favour of PR #12683

PR #12683, opened 2018-03-29
Remove HierarchicalNameMapper as a configured @Bean

jkschneider:
Proposed solution for #12676.


snicoll:
Thanks @jkschneider, this is what I had in mind. It's a breaking change in the sense that a custom `HierarchicalNameMapper` bean wouldn't be taken into account anymore.

Given that doing that is wrong, I am in favour of merging this. Let's see what the rest of the team thinks.

philwebb:
I think we should merge this, probably better to make a small breaking change than limp on with a bad assumption.

PR #12697, opened 2018-03-30
Fix example to avoid wrong copy&paste

graben:
Chapter 57.1 Process Monitoring - Extending Configuration actually has a wrong example which can lead to wrong copy&paste error.


philwebb:
Nice! We moved those in 2.0 and must have missed the docs.

PR #12707, opened 2018-03-31
WebServiceTemplateBuilder 

nosan:
Added the **WebServiceTemplateBuilder** (similar to the RestTemplateBuilder) helper class for building **WebServiceTemplate**.        

```
        WebServiceTemplateBuilder webServiceTemplateBuilder = new WebServiceTemplateBuilder();
        WebServiceTemplate webServiceTemplate = webServiceTemplateBuilder
                .setConnectionTimeout(connectTimeout)
                .setReadTimeout(readTimeout)
                .addCustomizers(customizers())
                .addInterceptors(interceptors())
                .addWebServiceMessageSenders(webServiceMessageSenders())
                .setCheckConnectionForFault(true)
                .setCheckConnectionForError(true)
                .setMarshaller(marshaller)
                .setUnmarshaller(unmarshaller)
                .setDefaultUri(defaultUri)
                ...
                .build();
```
In a typical auto-configured Spring Boot application this builder will be available as bean and can be injected whenever a WebServiceTemplate is needed.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
@@ -1,0 +1,592 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.webservices.client;
+
+import java.net.URI;
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.LinkedHashSet;
+import java.util.Set;
+
+import javax.xml.transform.TransformerFactory;
+
+import org.springframework.beans.BeanUtils;
+import org.springframework.oxm.Marshaller;
+import org.springframework.oxm.Unmarshaller;
+import org.springframework.util.Assert;
+import org.springframework.util.CollectionUtils;
+import org.springframework.ws.WebServiceMessageFactory;
+import org.springframework.ws.client.core.FaultMessageResolver;
+import org.springframework.ws.client.core.WebServiceTemplate;
+import org.springframework.ws.client.support.destination.DestinationProvider;
+import org.springframework.ws.client.support.interceptor.ClientInterceptor;
+import org.springframework.ws.transport.WebServiceMessageSender;
+
+/**
+ * Builder that can be used to configure and create a {@link WebServiceTemplate}. Provides
+ * convenience methods to register {@link #messageSenders(WebServiceMessageSender...)
+ * message senders}, {@link #interceptors(ClientInterceptor...) client interceptors} and
+ * {@link #customizers(Collection) customizers}.
+ * <p>
+ * In a typical auto-configured Spring Boot application this builder is available as a
+ * bean and can be injected whenever a {@link WebServiceTemplate} is needed.
+ *
+ * @author Dmytro Nosan
+ * @author Stephane Nicoll
+ * @since 2.1.0
+ */
+public class WebServiceTemplateBuilder {
+
+	private final Set<ClientInterceptor> interceptors;
+
+	private final Set<WebServiceTemplateCustomizer> internalCustomizers;
+
+	private final Set<WebServiceTemplateCustomizer> customizers;
+
+	private final Set<WebServiceMessageSender> webServiceMessageSenders;
+
+	private final Marshaller marshaller;
+
+	private final Unmarshaller unmarshaller;
+
+	private final DestinationProvider destinationProvider;
+
+	private final Class<? extends TransformerFactory> transformerFactoryClass;
+
+	private final WebServiceMessageFactory messageFactory;
+
+	public WebServiceTemplateBuilder(WebServiceTemplateCustomizer... customizers) {
+		this(Collections.emptySet(), Collections.emptySet(),
+				append(Collections.<WebServiceTemplateCustomizer>emptySet(), customizers),
+				Collections.emptySet(), null, null, null, null, null);
+	}
+
+	private WebServiceTemplateBuilder(Set<ClientInterceptor> interceptors,
+			Set<WebServiceTemplateCustomizer> internalCustomizers,
+			Set<WebServiceTemplateCustomizer> customizers,
+			Set<WebServiceMessageSender> webServiceMessageSenders, Marshaller marshaller,
+			Unmarshaller unmarshaller, DestinationProvider destinationProvider,
+			Class<? extends TransformerFactory> transformerFactoryClass,
+			WebServiceMessageFactory messageFactory) {
+		this.interceptors = interceptors;
+		this.internalCustomizers = internalCustomizers;
+		this.customizers = customizers;
+		this.webServiceMessageSenders = webServiceMessageSenders;
+		this.marshaller = marshaller;
+		this.unmarshaller = unmarshaller;
+		this.destinationProvider = destinationProvider;
+		this.transformerFactoryClass = transformerFactoryClass;
+		this.messageFactory = messageFactory;
+	}
+
+	/**
+	 * Set the {@link ClientInterceptor ClientInterceptors} that should be used with the
+	 * {@link WebServiceTemplate}. Setting this value will replace any previously defined
+	 * interceptors.
+	 * @param interceptors the interceptors to set
+	 * @return a new builder instance
+	 * @see #additionalInterceptors(ClientInterceptor...)
+	 */
+	public WebServiceTemplateBuilder interceptors(ClientInterceptor... interceptors) {
+		Assert.notNull(interceptors, "Interceptors must not be null");
+		return interceptors(Arrays.asList(interceptors));
+	}
+
+	/**
+	 * Set the {@link ClientInterceptor ClientInterceptors} that should be used with the
+	 * {@link WebServiceTemplate}. Setting this value will replace any previously defined
+	 * interceptors.
+	 * @param interceptors the interceptors to set
+	 * @return a new builder instance
+	 * @see #additionalInterceptors(Collection)
+	 */
+	public WebServiceTemplateBuilder interceptors(
+			Collection<? extends ClientInterceptor> interceptors) {
+		Assert.notNull(interceptors, "Interceptors must not be null");
+		return new WebServiceTemplateBuilder(
+				append(Collections.<ClientInterceptor>emptySet(), interceptors),
+				this.internalCustomizers, this.customizers, this.webServiceMessageSenders,
+				this.marshaller, this.unmarshaller, this.destinationProvider,
+				this.transformerFactoryClass, this.messageFactory);
+	}
+
+	/**
+	 * Add additional {@link ClientInterceptor ClientInterceptors} that should be used
+	 * with the {@link WebServiceTemplate}.
+	 * @param interceptors the interceptors to add
+	 * @return a new builder instance
+	 * @see #interceptors(ClientInterceptor...)
+	 */
+	public WebServiceTemplateBuilder additionalInterceptors(
+			ClientInterceptor... interceptors) {
+		Assert.notNull(interceptors, "Interceptors must not be null");
+		return additionalInterceptors(Arrays.asList(interceptors));
+	}
+
+	/**
+	 * Add additional {@link ClientInterceptor ClientInterceptors} that should be used
+	 * with the {@link WebServiceTemplate}.
+	 * @param interceptors the interceptors to add
+	 * @return a new builder instance
+	 * @see #interceptors(Collection)
+	 */
+	public WebServiceTemplateBuilder additionalInterceptors(
+			Collection<? extends ClientInterceptor> interceptors) {
+		Assert.notNull(interceptors, "Interceptors must not be null");
+		return new WebServiceTemplateBuilder(append(this.interceptors, interceptors),
+				this.internalCustomizers, this.customizers, this.webServiceMessageSenders,
+				this.marshaller, this.unmarshaller, this.destinationProvider,
+				this.transformerFactoryClass, this.messageFactory);
+	}
+
+	/**
+	 * Set {@link WebServiceTemplateCustomizer WebServiceTemplateCustomizers} that should
+	 * be applied to the {@link WebServiceTemplate}. Customizers are applied in the order
+	 * that they were added after builder configuration has been applied. Setting this
+	 * value will replace any previously configured customizers.
+	 * @param customizers the customizers to set
+	 * @return a new builder instance
+	 * @see #additionalCustomizers(WebServiceTemplateCustomizer...)
+	 */
+	public WebServiceTemplateBuilder customizers(
+			WebServiceTemplateCustomizer... customizers) {
+		Assert.notNull(customizers, "Customizers must not be null");
+		return customizers(Arrays.asList(customizers));
+	}
+
+	/**
+	 * Set {@link WebServiceTemplateCustomizer WebServiceTemplateCustomizers} that should
+	 * be applied to the {@link WebServiceTemplate}. Customizers are applied in the order
+	 * that they were added after builder configuration has been applied. Setting this
+	 * value will replace any previously configured customizers.
+	 * @param customizers the customizers to set
+	 * @return a new builder instance
+	 * @see #additionalCustomizers(Collection)
+	 */
+	public WebServiceTemplateBuilder customizers(
+			Collection<? extends WebServiceTemplateCustomizer> customizers) {
+		Assert.notNull(customizers, "Customizers must not be null");
+		return new WebServiceTemplateBuilder(this.interceptors, this.internalCustomizers,
+				append(Collections.<WebServiceTemplateCustomizer>emptySet(), customizers),
+				this.webServiceMessageSenders, this.marshaller, this.unmarshaller,
+				this.destinationProvider, this.transformerFactoryClass,
+				this.messageFactory);
+	}
+
+	/**
+	 * Add additional {@link WebServiceTemplateCustomizer WebServiceTemplateCustomizers}
+	 * that should be applied to the {@link WebServiceTemplate}. Customizers are applied
+	 * in the order that they were added after builder configuration has been applied.
+	 * @param customizers the customizers to add
+	 * @return a new builder instance
+	 * @see #customizers(WebServiceTemplateCustomizer...)
+	 */
+	public WebServiceTemplateBuilder additionalCustomizers(
+			WebServiceTemplateCustomizer... customizers) {
+		Assert.notNull(customizers, "Customizers must not be null");
+		return additionalCustomizers(Arrays.asList(customizers));
+	}
+
+	/**
+	 * Add additional {@link WebServiceTemplateCustomizer WebServiceTemplateCustomizers}
+	 * that should be applied to the {@link WebServiceTemplate}. Customizers are applied
+	 * in the order that they were added after builder configuration has been applied.
+	 * @param customizers the customizers to add
+	 * @return a new builder instance
+	 * @see #customizers(Collection)
+	 */
+	public WebServiceTemplateBuilder additionalCustomizers(
+			Collection<? extends WebServiceTemplateCustomizer> customizers) {
+		Assert.notNull(customizers, "Customizers must not be null");
+		return new WebServiceTemplateBuilder(this.interceptors, this.internalCustomizers,
+				append(this.customizers, customizers), this.webServiceMessageSenders,
+				this.marshaller, this.unmarshaller, this.destinationProvider,
+				this.transformerFactoryClass, this.messageFactory);
+	}
+
+	/**
+	 * Sets the {@link WebServiceMessageSender WebServiceMessageSenders} that should be
+	 * used with the {@link WebServiceTemplate}. Setting this value will replace any
+	 * previously defined message senders.
+	 * @param messageSenders the message senders to set
+	 * @return a new builder instance.
+	 * @see #additionalMessageSenders(WebServiceMessageSender...)
+	 */
+	public WebServiceTemplateBuilder messageSenders(
+			WebServiceMessageSender... messageSenders) {
+		Assert.notNull(messageSenders, "MessageSenders must not be null");
+		return messageSenders(Arrays.asList(messageSenders));
+	}
+
+	/**
+	 * Sets the {@link WebServiceMessageSender WebServiceMessageSenders} that should be
+	 * used with the {@link WebServiceTemplate}. Setting this value will replace any
+	 * previously defined message senders.
+	 * @param messageSenders the message senders to set
+	 * @return a new builder instance.
+	 * @see #additionalMessageSenders(Collection)
+	 */
+	public WebServiceTemplateBuilder messageSenders(
+			Collection<? extends WebServiceMessageSender> messageSenders) {
+		Assert.notNull(messageSenders, "MessageSenders must not be null");
+		return new WebServiceTemplateBuilder(this.interceptors, this.internalCustomizers,
+				this.customizers,
+				append(Collections.<WebServiceMessageSender>emptySet(), messageSenders),
+				this.marshaller, this.unmarshaller, this.destinationProvider,
+				this.transformerFactoryClass, this.messageFactory);
+	}
+
+	/**
+	 * Add additional {@link WebServiceMessageSender WebServiceMessageSenders} that should
+	 * be used with the {@link WebServiceTemplate}.
+	 * @param messageSenders the message senders to add
+	 * @return a new builder instance.
+	 * @see #messageSenders(WebServiceMessageSender...)
+	 */
+	public WebServiceTemplateBuilder additionalMessageSenders(
+			WebServiceMessageSender... messageSenders) {
+		Assert.notNull(messageSenders, "MessageSenders must not be null");
+		return additionalMessageSenders(Arrays.asList(messageSenders));
+	}
+
+	/**
+	 * Add additional {@link WebServiceMessageSender WebServiceMessageSenders} that should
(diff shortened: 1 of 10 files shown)


nosan:
remove Jaxb2MarshallerBuilder from this PR.
I think it's much better to have it in a separate PR if we still needed it.

nosan:
@philwebb  
I was hoping to hear some news from you regarding my pull request.
Would you be so kind to check that, if you are not checking it by now? Maybe you have some questions/concerns regarding some code? I will be more than happy to explain all the misunderstanding moments.

philwebb:
@nosan I've been primarily focusing on bugs and regressions in the 2.0.x line so feature PRs haven't had much of my attention lately.

The code looks very complete but I guess my primary concern is with how many people are currently using `WebServiceTemplate`, and if the code would be best as part of Spring Boot or if some of it should live in `spring-ws`. @gregturn might have some thoughts about that.

I'm probably not going to have time to give much feedback on this in the near-term, but as a team we'll be looking again at PRs once work on 2.1 starts to ramp up.

nosan:
@philwebb Thank you for your reply. Looking forward to hearing from you soon.

rfelgent:
Out of curiosity: Aren't there other libraries already providing the feature to create webserivce client stubs like CXF?

nosan:
@snicoll  I saw that you had requested a review, thank you for this. FYI I have done some small changes on this one (I removed reflections where it is possible). I hope I didn't break your review.

nosan:
@rfelgent `spring-ws` has a [WebServiceTemplate](https://docs.spring.io/spring-ws/docs/current/api/org/springframework/ws/client/core/WebServiceTemplate.html) class. PR makes the creation of `WebServiceTemplate` much easier, also this builder could be shared between beans.

In future it would be great to have metrics for it as `spring` has for `RestTemplate`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+	 * Set {@link ClientInterceptor ClientInterceptors} that should be used with the
+	 * {@link WebServiceTemplate}. Interceptors are applied in the order that they were
+	 * added after builder configuration has been applied.
+	 *
+	 * <b>Note!</b> Override existing interceptors
+	 * @param interceptors the interceptors to set
+	 * @return a new builder instance
+	 * @see WebServiceTemplate#setInterceptors(ClientInterceptor[])
+	 */
+	public WebServiceTemplateBuilder setInterceptors(ClientInterceptor... interceptors) {
```
snicoll: I prefer if we keep the same style as what we're doing with `RestTemplate` (i.e. `interceptors` and `additionalInterceptors`). I've already started a polish on my fork so need to act on it.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+	 * each time when {@link #configure(WebServiceTemplate)} method is called.
+	 *
+	 * <b>Note!</b> Override existing WebServiceMessageSender {@code suppliers}
+	 * @param webServiceMessageSenderSuppliers Suppliers for the messageSenders
+	 * @return a new builder instance.
+	 * @see WebServiceTemplate#setMessageSenders(WebServiceMessageSender[])
+	 */
+
+	public WebServiceTemplateBuilder setWebServiceMessageSenders(
+			Collection<? extends Supplier<? extends WebServiceMessageSender>> webServiceMessageSenderSuppliers) {
```
snicoll: I've removed the Supplier in my polish commit. I didn't found a single use of it and would argue that it is something you can determine when you build the template

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+	 * Set {@link WebServiceTemplate#setCheckConnectionForFault(boolean)
+	 * setCheckConnectionForFault} on the underlying.
+	 * @param checkConnectionForFault Specify whether checkConnectionForFault should be
+	 * enabled or not.
+	 * @return a new builder instance.
+	 **/
+	public WebServiceTemplateBuilder setCheckConnectionForFault(
+			boolean checkConnectionForFault) {
+		return new WebServiceTemplateBuilder(this.interceptors,
+				append(this.internalCustomizers,
```
snicoll: I don't like this pattern. If you call the setter several times, you'll append this customizer and rely on the fact the last one called will set the expected value. I guess you've done this that way to avoid adding too much parameters to the builder? I think we need to find a different option for this.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+	 * <b>Note!</b> Override existing WebServiceMessageSender {@code suppliers}
+	 * @param webServiceMessageSenderClass {@code Class} of
+	 * {@link WebServiceMessageSender}
+	 * @return a new builder instance.
+	 * @see WebServiceTemplate#setMessageSenders(WebServiceMessageSender[])
+	 * @see #setWebServiceMessageSender(Supplier)
+	 * @see BeanUtils#instantiateClass(Class)
+	 */
+
+	public WebServiceTemplateBuilder setWebServiceMessageSender(
```
snicoll: there are too many flavours to set a `WebServiceMessageSender`. Specifying a vararg of instance and a `Class` looks wrong to me. I've removed that in my fork.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+					int timeout) {
+				ClassLoader classLoader = getClass().getClassLoader();
+				customize(CUSTOMIZERS, getRequestFactory(source), Timeout.CONNECTION,
+						timeout, classLoader);
+			}
+
+			private ClientHttpRequestFactory getRequestFactory(
+					ClientHttpRequestMessageSender source) {
+				ClientHttpRequestFactory requestFactory = source.getRequestFactory();
+				if (!(requestFactory instanceof AbstractClientHttpRequestFactoryWrapper)) {
```
snicoll: I haven't looked in detail but I am wondering why you need to resort to reflection for a builder pattern that should "just" do what the regular template can do. Perhaps this feature should be removed in benefit from something more advanced? I don't like the idea to use reflection in production code.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+	}
+
+	/**
+	 * Sets the connection timeout in milliseconds on the underlying.
+	 * @param connectionTimeout the connection timeout in milliseconds
+	 * @return a new builder instance.
+	 * @throws java.lang.IllegalStateException if the underlying source doesn't support a
+	 * connection timeout.
+	 */
+	public WebServiceTemplateBuilder setConnectionTimeout(int connectionTimeout) {
```
snicoll: Considering that we may have multiple `WebServiceMessageSender`, I don't think we should expose this method here. You can provide a configured `WebServiceMessageSender`, this feels weird to me that all of them are reconfigured behind the scenes.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+	}
+
+	/**
+	 * Sets the read timeout in milliseconds on the underlying.
+	 * @param readTimeout the read timeout in milliseconds
+	 * @return a new builder instance.
+	 * @throws java.lang.IllegalStateException if the underlying source doesn't support a
+	 * read timeout.
+	 */
+	public WebServiceTemplateBuilder setReadTimeout(int readTimeout) {
```
snicoll: This should also go away.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+
+		if (this.transformerFactoryClass != null) {
+			webServiceTemplate.setTransformerFactoryClass(this.transformerFactoryClass);
+		}
+
+		if (this.messageFactory != null) {
+			webServiceTemplate.setMessageFactory(this.messageFactory);
+		}
+
+		if (!CollectionUtils.isEmpty(this.customizers)) {
```
snicoll: This should be the last step

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -0,0 +1,1106 @@
...
+			webServiceTemplate.setMessageSenders(
+					webServiceMessageSenders.toArray(new WebServiceMessageSender[0]));
+		}
+		else if (this.detectWebServiceMessageSender) {
+			webServiceTemplate.setMessageSenders(
+					new WebServiceMessageSender[] { detectMessageSender() });
+		}
+
+		if (!CollectionUtils.isEmpty(this.webServiceMessageSenderCustomizers)) {
+			if (!ObjectUtils.isEmpty(webServiceTemplate.getMessageSenders())) {
```
snicoll: This looks weird to me. Why do we check if the template has existing messageSenders?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
nosan: You can't do this because of 'setReadTimeout' and 'setConnectionTimeout'. This test will not work.
```
HttpUrlConnectionMessageSender sender = new HttpUrlConnectionMessageSender();

		 WebServiceTemplate ws = this.builder
				 .setConnectionTimeout(5000)
				 .setReadTimeout(2000)
				 .messageSenders(sender)
				 .build();

		WebServiceTemplate ws1 = this.builder
				.setConnectionTimeout(4000)
				.setReadTimeout(3000)
				.build();

		assertThat(ReflectionTestUtils.getField(ws.getMessageSenders()[0], "connectionTimeout"))
				.isEqualTo(Duration.ofMillis(5000));
		assertThat(ReflectionTestUtils.getField(ws.getMessageSenders()[0], "readTimeout"))
				.isEqualTo(Duration.ofMillis(2000));
		//

		assertThat(ReflectionTestUtils.getField(ws1.getMessageSenders()[0], "connectionTimeout")).isEqualTo(Duration.ofMillis(4000));
		assertThat(ReflectionTestUtils.getField(ws1.getMessageSenders()[0], "readTimeout"))
				.isEqualTo(Duration.ofMillis(3000));

```

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
nosan: it could be handled by `Boolean`  type, but you still need a custom customizer for `FaultMessageResolver`. 

Yes, I did this to avoid adding too many parameters.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
nosan: the was an idea to have a class instead of a supplier.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
nosan: Completely agree, this should be the last step, I did this in favor of `RestTemplateBuilder` algorithm.
```
public <T extends RestTemplate> T configure(T restTemplate) {
		configureRequestFactory(restTemplate);
		if (!CollectionUtils.isEmpty(this.messageConverters)) {
			restTemplate.setMessageConverters(new ArrayList<>(this.messageConverters));
		}
		if (this.uriTemplateHandler != null) {
			restTemplate.setUriTemplateHandler(this.uriTemplateHandler);
		}
		if (this.errorHandler != null) {
			restTemplate.setErrorHandler(this.errorHandler);
		}
		if (this.rootUri != null) {
			RootUriTemplateHandler.addTo(restTemplate, this.rootUri);
		}
		if (this.basicAuthorization != null) {
			restTemplate.getInterceptors().add(this.basicAuthorization);
		}
		if (!CollectionUtils.isEmpty(this.restTemplateCustomizers)) {
			for (RestTemplateCustomizer customizer : this.restTemplateCustomizers) {
				customizer.customize(restTemplate);
			}
		}
		restTemplate.getInterceptors().addAll(this.interceptors);
		return restTemplate;
	}
```

That was weird to me...

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
nosan: You can easily override a `WebServiceTemplate` and override `default strategies`.

```
    protected void initDefaultStrategies() {
        DefaultStrategiesHelper strategiesHelper = new DefaultStrategiesHelper(WebServiceTemplate.class);
        if (this.getMessageFactory() == null) {
            this.initMessageFactory(strategiesHelper);
        }

        if (ObjectUtils.isEmpty(this.getMessageSenders())) {
            this.initMessageSenders(strategiesHelper);
        }

        if (this.getFaultMessageResolver() == null) {
            this.initFaultMessageResolver(strategiesHelper);
        }

    }

```

nosan:
@snicoll  Thanks a lot.
With a high probability, you will use `setReadTimeout` and `setConnectionTimeout` and  I agree that logic of setting up these timeouts are really ugly and I don't like it. 
`Spring Boot` helps people make their life easier, so why we should force people to set up these timeouts through the `WebServiceMessageSender` instead of doing this logic by ourselves?  We can remove this logic from the `WebServiceTemplate` and create a utility class `WebServiceTemplateCustomizers` and have controversial features there.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
snicoll: IMO that's one more argument to remove that feature.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
snicoll: In the meantime I've seen we're doing that with `RestTemplateBuilder` and I see you're using a separate collection. Retrospectively, it's not that bad at all.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
snicoll: Got it but I think the reason why a `Supplier` was added is wrong.

snicoll:
@nosan thanks for the feedback. Adding compromise in the public API to implement this feature is a strong sign that we shouldn't do it. I am afraid my knowledge of `spring-ws` is seriously lacking so I don't really know if that's something you'd often do. 

I don't think we have to be at the level of `RestTemplateBuilder` on step 1. It's important to make sure is consistent but maybe things aren't applied the same way. The fact RestTemplate has one factory and WebServiceTemplate has n message sender is a big difference IMO.

I can have another look if you want but I am wondering what you thoughts on that were. Thanks again!

nosan:
@snicoll Thanks.
 Maybe you are right and we should not add these features, but we will have another issue with `detectWebServiceMessageSender`. People will not have any way to configure these timeouts except `Customizers` and as can you see the `detectWebServiceMessageSender` is based on the classpath and people will not know exactly what type of the `WebServiceTemplate` they have received in the customize method. They will use `instanceof` and I am not a fan of that, thus `detectWebServiceMessageSender` a bit doubtful feature and I'm not sure whether we need it or not.  Any thoughts? 
Thanks in advance.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
nosan: Only for custom `WebServiceMessagaSenders`

nosan:
@snicoll  Please take a [look ](https://github.com/spring-projects/spring-boot/pull/12707/commits/3bd6597f411a81e5ab821aba3178fd51046acfca)

snicoll:
Alright so I took another round at this one and I have something in the `spring-boot` module that looks in a good state to be merged. I haven't looked at the auto-configuration and the doc but those are fairly minimal and obvious so it won't be a problem.

@gregturn @poutsma I could use an additional review on the high-level principles if you don't mind.

The biggest problem so far is the ability to detect a valid HTTP-based `WebServiceMessageSender` while retaining the ability to configure the read and connection timeout. The original PR offered a `setConnectionTimeout` and `setReadTimeout` on the builder which looks wrong to me as you may associate more than one message sender and some may simply not support this at all. 

To reduce the confusion, I've added a `detectHttpMessageSender(Duration connectionTimeout, Duration readTimeout)` that allows you to detect a valid HTTP-based sender with those settings. Such message sender is _always_ added unless `detectHttpMessageSender` is false. The only problem remaining is that the Javadoc of `messageSender` is a bit awkward as it states that it will override any existing message senders configured on the builder. If you call `messageSender(foo).build()` you'll get two message senders: `foo` and the HTTP one we've detected. The reason behind that is that if you decide to have a custom `MessageSender` then you lose the ability to easily configure an HTTP-based message sender with timeout values.

I've also extracted all the business of finding the proper HTTP library to a dedicated class that both `RestTemplateBuilder` and `WebServiceTemplateBuilder` use. 

I am still not sure about those detect methods and the fact the timeout is passed in a single call (forcing you to pass `null` if you only care about one of them).

@nosan let me know what you think of my polish, it's available [on my fork](https://github.com/snicoll/spring-boot/commits/pr/12707)

nosan:
@snicoll  Thank you.

Wouldn't it be a better idea to use `HttpComponentsMessageSender`, `HttpUrlConnectionMessageSender` implementations which support timeouts instead of `ClientHttpRequestMessageSender`.  **BTW**, the current implementation doesn't support `HttpComponentsMessageSender`, `HttpUrlConnectionMessageSender` and it looks weird for me. In my opinion we should avoid `ClientHttpRequestMessageSender` because it doesn't support this functionallity or we will have to use `Reflection`.


> The original PR offered a setConnectionTimeout and setReadTimeout on the builder which looks wrong to me as you may associate more than one message sender and some may simply not support this at all.



I still don't quite understand why it looks wrong. Yeah, it doesn't support a custom `WebServiceMessageSender`, but `RestTemplateBuilder` doesn't support a custom `ClientHttpRequestFactory` as well. Currently, we know that` HttpComponentsMessageSender`, `HttpUrlConnectionMessageSender` supports them, so we can configure them.

If someone needs a custom `WebServiceMessageSender` then these methods should be avoided.

**P.S** I think this builder becomes more complicated than easier.  If you don't mind I will update my PR based on your changes.

snicoll:
> Wouldn't it be a better idea to use HttpComponentsMessageSender, HttpUrlConnectionMessageSender implementations which support timeouts instead of ClientHttpRequestMessageSender. 

ClientHttpRequestMessageSender` does support timeout the exact same way `RestTemplateBuilder` uses it so it's much more consistent. I certainly prefer than than configuring something specific when we have an abstraction (i.e. `ClientHttpRequestFactory`).

> the current implementation doesn't support HttpComponentsMessageSender, HttpUrlConnectionMessageSender and it looks weird for me. 

It doesn't have to. The former is supported via `ClientHttpRequestFactory` and the latter is a simple fallback for which there is a `ClientHttpRequestFactory` as well (please look at the Javadoc of `HttpUrlConnectionMessageSender`).

 > I still don't quite understand why it looks wrong

It looks wrong because a) There isn't a concept of a `WebMessageServiceSender` but a collection and  you may add an JMS or another implementation that has no notion of read/connect timeout. What the builder does effectively is detecting an HTTP-based `WebServiceMessageSender` so the API should be very clear about that.

>  I think this builder becomes more complicated than easier. If you don't mind I will update my PR based on your changes.

Perhaps you could share why you feel it's more complicated? Please keep in mind that the behaviour also has to be consistent and explicit.

nosan:
@snicoll  Thanks.

> ClientHttpRequestMessageSender does support timeout the exact same wayRestTemplateBuilderuses it so it's much more consistent. I certainly prefer than than configuring something specific when we have an abstraction (i.e.ClientHttpRequestFactory`

No, it doesn't. It requires to have `ClientHttpRequestFactory` with methods `setReadTimeout(int)` and `setConnectTimeout(int)` and call them through `Reflection`. 

> It doesn't have to. The former is supported via ClientHttpRequestFactory and the latter is a simple fallback for which there is a ClientHttpRequestFactory as well (please look at the Javadoc of HttpUrlConnectionMessageSender).

Yes, you are right,  but If I don't want to have a dependency on ` org.springframework:spring-web:$VERSION`? 


> Perhaps you could share why you feel it's more complicated? Please keep in mind that the behavior also has to be consistent and explicit.

We are trying to support everything, but in the end we support nothing. So, perhaps you were right and we should avoid `setConnectionTimeout` ,`setReadTimeout` , `detectWebServiceMessageSender` from the `WebServiceTemplateBuilder`.

nosan:
@snicoll But still, If you think this current implementation is within its scope, then I will not interfere.

I would be great to reconsider this method. 
`public WebServiceTemplateBuilder detectHttpMessageSender(Duration connectionTimeout,
			Duration readTimeout)`

gregturn:
@nosan Another aspect to consider is that you have you wrote a custom `BeanPostPostProcessor`, can you apply the other 10% needed to the bean created by Boot and thus not require Boot's code too specific to one problem?

nosan:
@gregturn
It is not needed to have a `BeanPostProccesor` because we have already had `WebServiceTemplateCustomizer` and you can use it.

As I mentioned before it would be better to remove aspecific functionality from `WebServiceTemplateBuilder`

nosan:
@snicoll  I have updated my PR and I have removed all the controversial things in my opinion.

snicoll:
> No, it doesn't. It requires to have ClientHttpRequestFactory with methods setReadTimeout(int) and setConnectTimeout(int) and call them through Reflection.

That's exactly what `RestTemplateBuilder` is doing. 

I've seen you've edited your comment in the meantime as you've probably realized that `spring-websocket` has a [compile dependency on `spring-web`](https://github.com/spring-projects/spring-framework/blob/b71d0eeec9c9ed59f8b7f4e3acaec6f4919b46b0/spring-websocket/spring-websocket.gradle#L22). So we're good on that front.

We'll continue based on my original polish as I've already spent a significant amount of time on it. I am considering moving the detection of a suitable HTTP-based `WebServiceMessageSender` to a separate component.

nosan:
@snicoll  Thank you.

> I've seen you've edited your comment in the meantime as you've probably realized that spring-websocket has a compile dependency on spring-web. So we're good on that front.

`spring-websocket`? The PR was about `spring-ws`. 

> Spring Web Services (Spring-WS) is a product of the Spring community focused on creating document-driven Web services. Spring Web Services aims to facilitate contract-first SOAP service development, allowing for the creation of flexible web services using one of the many ways to manipulate XML payloads. The product is based on Spring itself, which means you can use the Spring concepts such as dependency injection as an integral part of your Web service.

[more...](https://spring.io/projects/spring-ws)

> It looks wrong because a) There isn't a concept of a WebMessageServiceSender but a collection so I find it weird that a setter impacts a collection rather than a well targeted object and b) you may add a JMS or another implementation that has no notion of read/connect timeout. What the builder does effectively is detecting an HTTP-based WebServiceMessageSender so the API should be very clear about that.


I have never seen SOAP JMS :)

nosan:
> I've seen you've edited your comment in the meantime as you've probably realized that spring-websocket has a compile dependency on spring-web. So we're good on that front.

Yes, I have checked `spring-ws` and it has a compile-time dependency on `spring-web`.
My apologies.

snicoll:
> spring-websocket? The PR was about spring-ws. I think it was just a typo.

Sorry about that, my brain mixed the two concepts, here is [the dependency section](https://github.com/spring-projects/spring-ws/blob/0d814c03520da0cb35977a9195cbae68b1d7e6d7/spring-ws-core/pom.xml#L44)  I intended initially.

> I have never seen SOAP which is based on JMS :)

The [reference documentation](https://docs.spring.io/spring-ws/site/reference/html/client.html) describes transports for HTTP, JMS, email and XMPP so we can't have a `detectMessageSender` method that assumes it's HTTP, even if is the dominant transport. I've given it some more thoughts and a separate builder for the `WebServiceMessageSender` looks a good compromise to me. I'll spike on that next week.

nosan:
> The reference documentation describes transports for HTTP, JMS, email and XMPP so we can't have a detectMessageSender method that assumes it's HTTP, even if is the dominant transport.

 I see, thank you, so I  was wrong on the beginning. Sorry about that.

> I've given it some more thoughts and a separate builder for the WebServiceMessageSender looks a good compromise to me. I'll spike on that next week.

Do you need some help with that?

gregturn:
@snicoll While it's true that SOAP supports all these various protocols, since we're talking Spring Boot which is based on autoconfiguring for the 80% case, perhaps we could assume HTTP (or trigger off some web-based criteria) and supply that.

HTTP is probably 80% or more of what people are doing with Spring WS, let alone the SOAP community.

snicoll:
That wasn’t my point. My point is that we can’t assume message sender is HTTP based as both its contract and the doc states otherwise. See the discussion above for more details.

snicoll:
Closed by c973488

nosan:
@snicoll Thank you a lot for your time.

PR #12711, opened 2018-04-01
Replace EnvironmentTestUtils documentation with TestPropertyValues

ruifigueira:
[Documentation ](https://docs.spring.io/spring-boot/docs/current/reference/html/boot-features-testing.html#boot-features-environment-test-utilities) still refers to the now deprecated `EnvironmentTestUtils` class instead of `TestPropertyValues` for setting environment values in tests.

This commit replaces it with `TestPropertyValues` and provides an equivalent code example.


snicoll:
@ruifigueira good catch, thank you!

Issue #12714, opened 2018-04-02
Multiple @ServletComponentScan annotated class may cause UnsupportedOperationException


vivimice:
# Description

When a project contains multiple `@ServletComponentScan` annotated class in classpath, and at least one annotation don't explicitly specify `basePackages` and `basePackageClass` attribute, the application MAY
fail on startup, caused by UnsupportedOperationException:

<pre>
java.lang.UnsupportedOperationException: null
	at java.util.AbstractCollection.add(AbstractCollection.java:262)
	at java.util.AbstractCollection.addAll(AbstractCollection.java:344)
	at org.springframework.boot.web.servlet.ServletComponentScanRegistrar.updatePostProcessor(ServletComponentScanRegistrar.java:62)
	at org.springframework.boot.web.servlet.ServletComponentScanRegistrar.registerBeanDefinitions(ServletComponentScanRegistrar.java:48)
	...
</pre>

# Reproduction

There is a single class application to reproduce this bug just for demostration.

<pre>
@SpringBootApplication
public class SpringBootTest1 {
    public static void main(String[] args) {
        SpringApplication.run(SpringBootTest1.class, args);
    }
    
    @ServletComponentScan
    public static class Config1 {
        
    }
    
    @ServletComponentScan
    public static class Config2 {
        
    }
}
</pre>

philwebb:
Closing in favor of PR #12715

PR #12715, opened 2018-04-02
Multiple  @ServletComponentScan with no base packages may cause startup to fail with an UnsupportedOperationException 

vivimice:
#12714

# Description

When a project contains multiple `@ServletComponentScan` annotated class in classpath, and at least one annotation don't explicitly specify `basePackages` and `basePackageClass` attribute, the application MAY
fail on startup, caused by UnsupportedOperationException:

<pre>
java.lang.UnsupportedOperationException: null
	at java.util.AbstractCollection.add(AbstractCollection.java:262)
	at java.util.AbstractCollection.addAll(AbstractCollection.java:344)
	at org.springframework.boot.web.servlet.ServletComponentScanRegistrar.updatePostProcessor(ServletComponentScanRegistrar.java:62)
	at org.springframework.boot.web.servlet.ServletComponentScanRegistrar.registerBeanDefinitions(ServletComponentScanRegistrar.java:48)
	...
</pre>

# Reproduction

There is a single class application to reproduce this bug just for demostration.

<pre>
@SpringBootApplication
public class SpringBootTest1 {
        SpringApplication.run(SpringBootTest1.class, args);
    }
    
    @ServletComponentScan
    public static class Config1 {
        
    }
    
    @ServletComponentScan
    public static class Config2 {
        
    }
}
</pre>

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/ServletComponentScanRegistrar.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/ServletComponentScanRegistrar.java
@@ -17,7 +17,6 @@
 package org.springframework.boot.web.servlet;
 
 import java.util.Arrays;
-import java.util.Collections;
 import java.util.LinkedHashSet;
 import java.util.Set;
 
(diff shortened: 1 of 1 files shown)


philwebb:
We should ideally add a unit test for this.

wilkinsona:
@vivimice Thank you very much for making your first contribution to Spring Boot. This fix that you proposed is [now in 1.5.x](https://github.com/spring-projects/spring-boot/commit/7bec780281fd613a9ef6b0bedccf8f66284aceb9) and master. I added some tests in [this polishing commit](https://github.com/spring-projects/spring-boot/commit/6078fdaed82cf96bc427fd9107d14d2ba5e8eece).

vivimice:
@wilkinsona My pleasure

PR #12725, opened 2018-04-03
Remove destroyMethod attribute in @Bean for MeterRegistry

izeye:
As `MeterRegistry` is `AutoCloseable` and `stop()` is called in `close()`, `destroyMethod` attributes can be removed.


snicoll:
Thanks again @izeye!

PR #12732, opened 2018-04-03
Fix Document cache in ConfigFileApplicationListener

dreis2211:
Hi,

I just noticed that the `loadDocumentsCache` inside `ConfigFileApplicationListener` was never updated and as such never used as a cache. While being on it, I removed a - in my opinion superfluous - call to `PropertySourceLoader.load` and a typo in its doc.

Let me know what you think.
Cheers,
Christoph


dreis2211:
Test failures seem to be unrelated

philwebb:
Thanks. Both look valid but we'll split this commit up when we merge it.

Issue #12743, opened 2018-04-03
Metrics from binders missing in CompositeMeterRegistry


pbetkier:
The following code behaves differently depending on whether you register a single or multiple instances of `MeterRegistry` in Spring context.

```java
@Component
class Test {
    Test(MeterRegistry registry) {
        if (registry.find("jvm.classes.loaded").gauge() != null) {
            // one MeterRegistry in context
        } else {
            // two or more MeterRegistry in context, forming a CompositeMeterRegistry
        }
    }
}
```

That's because meter binders, one of which registers `jvm.classes.loaded` gauge, bind to all meter registries except to `CompositeMeterRegistry`. So if you register multiple `MeterRegistry` beans, then each one will have the gauge, but the composite that gets injected does not.

Application code should not depend on such configuration detail as how many meter registries are configured. I also want to avoid writing `instanceof` conditions like in [collecting meter names in MetricsEndpoint](https://github.com/spring-projects/spring-boot/blob/6c61cbf5f76fda234c5c466cc1f6631b16860ecc/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java#L67).

There used to be always a `CompositeMeterRegistry` in context, I understand that it was important to remove the overhead and create such registry only when necessary. But then I suggest `CompositeMeterRegistry` methods to delegate and merge results from all the underlying registries, e.g. `find` to search in all the registries. Having to use `instanceof` and manually iterate over underlying registries kind of kills the whole idea of the composite pattern.

What do you think?

philwebb:
That feels like a bug `CompositeMeterRegistry` to me. I'd expect the composite not to have the gauge directly registered, but I would think that `find` should search the others and return the first. Hopefully @jkschneider can answer.

jkschneider:
I think this is the [problem](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryConfigurer.java#L62-L64).

philwebb:
I think the initial reason for discarding `CompositeMeterRegistry` was the assumption that the composite itself wouldn't have state but would just be a view of other registries. A simple example of the type pattern from the framework is `org.springframework.util.CompositeIterator` (an `Iterator` composed of other `Iterators`).

[...]

PR #12762, opened 2018-04-04
Apply MeterRegistryCustomizer to composites

jkschneider:
Fix for https://github.com/spring-projects/spring-boot/issues/12743.


spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryCustomizerTests.java
```diff
@@ -55,9 +58,21 @@ public void commonTagsAreAppliedBeforeRegistryIsInjectableElsewhere() {
 				.withUserConfiguration(MeterRegistryCustomizerConfiguration.class)
 				.run((context) -> {
 					MeterRegistry registry = context.getBean(MeterRegistry.class);
-					assertThat(
```
jkschneider: The assertion is superfluous in this case. `registry.get` throws an exception if no meter matches.

PR #12810, opened 2018-04-10
Fix broken links

izeye:
This PR fixes broken links.


snicoll:
Thanks @izeye, I am not sure I understand why the `current` symlink got removed. I've created https://github.com/spring-projects/spring-amqp/issues/740

philwebb:
It's been fixed upstream now. Thanks for raising it @izeye.

Issue #12805, opened 2018-04-09
Custom error.jsp page is not used when app is run with mvn spring-boot:run


wilkinsona:
The problem can be reproduced using `spring-boot-sample-web-jsp` with a `src/main/webapp/WEB-INF/jsp/error.jsp` with the following contents:

```
<!DOCTYPE html>
<html lang="en">
<body>
Something went wrong: ${status} ${error}
</body>
</html>
```

1. `$ mvn spring-boot:run`
2. `$ curl :8080/foo -H "Accept:text/html"`

```
<html><body><h1>Whitelabel Error Page</h1><p>This application has no explicit mapping for /error, so you are seeing this as a fallback.</p><div id='created'>Mon Apr 09 14:27:21 BST 2018</div><div>There was an unexpected error (type=Internal Server Error, status=500).</div><div>Foo</div></body></html>
```

By contrast, it works when packaged as a war:

1. `$ mvn clean package`
2. `$ java -jar target/spring-boot-sample-web-jsp-2.0.2.BUILD-SNAPSHOT.war`
3. ` curl :8080/foo -H "Accept:text/html"`

```
<!DOCTYPE html>
<html lang="en">
<body>
Something went wrong: 500 Internal Server Error
</body>
</html>
```

The problem occurs in both 1.5 and 2.0.

ralenmdp:
I think the problem is JspTemplateAvailabilityProvider is trying to look for the /WEB-INF/jsp/error.jsp in the classloader and failed to do so since /src/main/webapp is not included in the classpath.

When you add /src/main/webapp as resource in eclipse or maven this will fix the issue.

What do you think?

wilkinsona:
Closing in favour of the PR (#12859).

PR #12859, opened 2018-04-13
Custom error.jsp page is not used when app is run as anything other than a packaged war

ralenmdp:
https://github.com/spring-projects/spring-boot/issues/12805

Try to look for the webapps folder so that [WhitelabelErrorViewConfiguration ](https://github.com/spring-projects/spring-boot/blob/1.5.x/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ErrorMvcAutoConfiguration.java) will not create another error bean view and let InternalResourceViewResolver resolve the error page in embedded container.

### DIFF
--- a/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/JspTemplateAvailabilityProvider.java
+++ b/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/JspTemplateAvailabilityProvider.java
@@ -38,7 +38,8 @@ public class JspTemplateAvailabilityProvider implements TemplateAvailabilityProv
 			ClassLoader classLoader, ResourceLoader resourceLoader) {
 		if (ClassUtils.isPresent("org.apache.jasper.compiler.JspConfig", classLoader)) {
 			String resourceName = getResourceName(view, environment);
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ralenmdp Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12859)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12859) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ralenmdp Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12859)!

philwebb:
Thanks for the PR but I don't think this is the right way to fix the problem. I've not keen for `JspTemplateAvailabilityProvider` to have any special knowledge of the `src/main/webapp` folder structure that's used by the build system.

wilkinsona:
On reflection, we can't think of a better way to do this. The embedded containers already have knowledge or `src/main/webapp` and there's no easy way to pass that knowledge from the container to the availability provider. Hard-coding it seems like the best option.

PR #12867, opened 2018-04-15
Improve Joinfaces community starter entry

larsgrefer:
- Use more https URL's
- Added Adminfaces and Rewrite
- Fixed Mojarra URL


snicoll:
@larsgrefer thanks a lot for the PR but rather than taking care of that long list of libraries, I've decided to simplify our entry instead. Your PR is a good indication that we should rely on the target web site rather than copy/pasting a summary of what it does.

PR #12877, opened 2018-04-16
add javamelody spring-boot-starter

evernat:
add link to javamelody-spring-boot-starter in third-party starters


snicoll:
@evernat thanks for the PR. I'd like to make sure first that support for Spring Boot 2 is on its way as we can't really add a reference now if that's not the case.

evernat:
build and tested with Spring-boot 2, https://github.com/javamelody/javamelody/commit/9afe9860eff052ead3f9a9383e320b07fcdf58d0 for the next release (1.73)

PR #12882, opened 2018-04-17
Modify Status Response when Using Custom ErrorAttributes

kwyi:
I modified ErrorAttributes to change `javax.servlet.error.status_code`, but the response status value does not change.

https://github.com/spring-projects/spring-boot/blob/3cb2246e7a6f4094b9dda8ff767117e5f78f2dea/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/AbstractErrorController.java#L82-L94


kwyi:
@snicoll That's right. I'm using `BasicErrorController.java` and trying to respond to status values by changing only `ErrorAttributes.java`
https://github.com/spring-projects/spring-boot/blob/3cb2246e7a6f4094b9dda8ff767117e5f78f2dea/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/BasicErrorController.java#L99-L104
can be changed by checking the Status later

wilkinsona:
Thanks for the additional details, @kwyi. What's your usecase for changing the response status in this manner? The error attributes and error controller are intended to report an error, not to change it in some way so, while your proposed change is small, I'm not sure the functionality the it enables is something that we want to encourage.

wilkinsona:
Closing due to lack of requested feedback.

Issue #12904, opened 2018-04-18
Autoconfigure Jersey Server Metrics in a JAX-RS environment


michael-simons:
Micrometer metrics for Jersey Server should be collected when Spring Boot Actuator, Micrometer Core and Micrometer Jersey 2 is on the classpath of a JAX-RS / Jersey 2 Spring Boot application as they are collected for Web MVC and Web Flux.

snicoll:
Duplicate of #12904 - @michael-simons thanks but we don't need an issue and a PR.

PR #12905, opened 2018-04-18
Autoconfigure Jersey Server Metrics in a JAX-RS environment

michael-simons:
This PR implements #12904 by 

* Adding micrometer-jersey2 to Spring Boots dependency management
* Extracting the necessary autoconfiguration code from the "legacy"-module (i.e. https://github.com/micrometer-metrics/micrometer/tree/master/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/jersey2/server)
* Adding some tests and documentation for the new feature.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -94,6 +94,11 @@
 		</dependency>
 		<dependency>
 			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-jersey2</artifactId>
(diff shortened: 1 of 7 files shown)


shakuzen:
Looks like this might duplicate the efforts in #12482

michael-simons:
Yep. My search foo wasn’t good.

michael-simons:
It might be a good idea to copy some aspects of each PR after a quick check. I wouldn’t duplicate the properties for example and you have more tests.

wilkinsona:
Thank you, @michael-simons. We'll go ahead with #12482, modified to list you as a co-author.

Issue #5541, opened 2016-03-31
Using multiple embedded databases requires a lot of manual configuration exclusions


olivergierke:
If multiple embedded `DataSources` are declared, one needs to explicitly disable the `DataSourceAutoConfiguration` and `DataSourceTransactionManagerAutoConfiguration` as can be seen [here](https://github.com/spring-projects/spring-data-examples/blob/master/jpa/multiple-datasources/src/main/java/example/springdata/jpa/multipleds/Application.java#L41). They fail as they rely on unique `DataSource` instances to be present.

Can we let the simply opt out of being applied if multiple `DataSource`s are present?

wilkinsona:
> on needs to explicitly disable the DataSourceAutoConfiguration and DataSourceTransactionManagerAutoConfiguration as can be seen here

AFAIK, that isn't strictly accurate. You can use multiple datasources as long as one of them is marked as `@Primary` without excluding either `DataSourceAutoConfiguration` or `DataSourceTransactionManagerAutoConfiguration`. You can see this in action in @snicoll's [demon application](https://github.com/snicoll-demos/demo-multi-entity-managers).

That said, I expected this to be further improved by #2784 so that you could choose not to have a primary datasource and everything that worked with a single datasource would back off. That's not the case, though. @snicoll did you discover something that meant that using `@ConditionalOnSingleCandidate` wouldn't work in `DataSourceTransactionManagerAutoConfiguration`?

olivergierke:
I see. The point here is that making one `DataSource` the primary one can be a source of errors as you could `@Transactional` (without an explicit qualifier) by accident and thus run transactions on the "wrong" one. In the scenario I have here, both `DataSources` should be treated equally and not referring to one explicitly is rather considered an error.

PR #12914, opened 2018-04-18
Only apply autoconfiguration that injects DataSource by type, if there is a single or primary DataSource

rupert654:
Currently, an error is thrown if I add multiple `DataSource`s to my context.

There are two workarounds for this issue:

1. Mark one of the `DataSource`s as `@Primary`
2. Exclude all autoconfiguration that wants to inject `DataSource` by type

It is easy to do 1. but this can be arbitrary decision unless the user knows exactly what auto-configuration is being applied. It can lead to subtle differences in behaviour which can trip you up, especially if you don't understand what you are doing by choosing this option.

It is hard to do 2. without a full knowledge of which auto-configurations. A piece of auto-configuration knows if it wants to inject by type and can foresee that injecting a `DataSource` by type is likely to trip people up. Therefore, they should make use of `ConditionalOnSingleCandidate` so they don't trip anybody up. Afterall, if somebody has already added multiple DataSources without into their application, then they probably know what they want to do with them.

This was proposed in #5541 but the commit that closed that ticket did not actually add this annotation to either `DataSourceAutoConfiguration` nor `DataSourceTransactionManagerAutoConfiguration` so this PR fixes that and also updates the documentation to reflect this behaviour.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceAutoConfiguration.java
@@ -26,6 +26,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionOutcome;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnSingleCandidate;
 import org.springframework.boot.autoconfigure.condition.SpringBootCondition;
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceAutoConfiguration.java
```diff
@@ -49,6 +50,7 @@
  */
 @Configuration
 @ConditionalOnClass({ DataSource.class, EmbeddedDatabaseType.class })
+@ConditionalOnSingleCandidate(DataSource.class)
```
spencergibb: Not sure this makes sense as a few lines later there is `@ConditionalOnMissingBean({ DataSource.class, XADataSource.class })` which will now never happen.

snicoll:
@rupert654 I disagree with the analysis and we can't accept such change without additional tests that demonstrate and exercise the conditions you've added.

Please open a separate issue with a sample that we can run that demonstrates the problem you're experiencing.

Issue #10362, opened 2017-09-20
BackgroundPreinitializer can cause hang when context startup fails


ksperling:
I just observed a case of `BackgroundPreinitializer` causing a a test where the context failed to start to hang indefinitely.

Stack dump: https://gist.github.com/ksperling/b19382a5b9eef6fa0913b45cb9318367

From looking at the code it seems this might happen if a `ApplicationFailedEvent` is delivered without ever having received a `ApplicationEnvironmentPreparedEvent`. `preinitializationStarted` should be checked before calling `preinitializationComplete.await()`.

I am using Spring Boot 1.5.7.RELEASE

ksperling:
In my specific case this is cased by a `EnvironmentPostProcessor` that fails to load due to a `NoClassDefFoundError`. `ApplicationEnvironmentPreparedEvent` is never fired.

wilkinsona:
Oops. Thanks for the analysis.

PR #12946, opened 2018-04-23
Add to ApplicationFailedEvent.Exception information to debug log.

LiYing2010:
This commit provides exception information that can be used as diagnostics to a user when their application fails to start.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/logging/ClasspathLoggingApplicationListener.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/logging/ClasspathLoggingApplicationListener.java
@@ -54,7 +54,8 @@ public final class ClasspathLoggingApplicationListener
 			}
 			else if (event instanceof ApplicationFailedEvent) {
 				logger.debug(
-						"Application failed to start with classpath: " + getClasspath());
+						"Application failed to start with classpath: " + getClasspath(),
+						((ApplicationFailedEvent) event).getException());
 			}
 		}
 	}


LiYing2010:
@snicoll 
Sorry, this issue happened to me like a month ago, I got a `"Application failed to start with classpath: "` error log, but no Exception was output.
I don't remember the details now, I will try if I can create a sample app to reproduce this issue.

snicoll:
@LiYing2010 it is a debug log and the actual exception is logged after that unless there is a failure analzer than can handle it (you can still set the exception when the debug mode is on). The reason I am asking for a sample is that I don't think we should make that change.

LiYing2010:
@snicoll

Hi, It took me some time, but I finally reproduced this issue.

We have a web application using `Spring-boot` v1.5.7 + `Spring Framework` v4.3.11 .
(I am sorry, it happened in `Spring-boot` **v1.5.7**, but I did not test it in `Spring-boot` **v2.x**)

And someday, when I was developing a new feature, I modified our `application.yml` configuration file, but made some mistake by accident, and then:
1. Application failed to start, since YAML parser could not parse this incorrect yml file.
2. `SimpleApplicationEventMulticaster#multicastEvent` got executed, to notify all listeners about this `ApplicationFailedEvent`
3. `ClasspathLoggingApplicationListener#onApplicationEvent` got executed, and output `"Application failed to start with classpath: blah blah blah"` debug log.
4. `ClasspathLoggingApplicationListener#onApplicationEvent` try to call next listener, which is a `BackgroundPreinitializer`, and `BackgroundPreinitializer#onApplicationEvent` got executed
5. in method `BackgroundPreinitializer#onApplicationEvent`, it called `preinitializationComplete.await()`, and this call is blocked like forever
6. the `logger.warn("Unable to close ApplicationContext", ex)` error log never got a chance to execute, since `BackgroundPreinitializer#onApplicationEvent` caused a deadlock

So, the result is, we got a `"Application failed to start with classpath: blah blah blah"` debug log, but not exception information for diagnostics.

I understand the reason is something was going wrong at step 5, but it blocked step 6,
but in my opinion, is it better if we output error information ASAP?
which means, add exception info to the debug log in step 3

wilkinsona:
Thanks for the additional details. This is a duplicate of #10362.

LiYing2010:
Thanks, understood.

PR #12955, opened 2018-04-24
Update version of gradle-git-properties in doc

rajadilipkolli:
upgrading to latest version solves the issue with newer gradle version.

This should be done only in 2.0.x+ branch as this plugin now needs[ java 8+](https://github.com/n0mer/gradle-git-properties)

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
@@ -2564,7 +2564,7 @@ plugin, as shown in the following example:
 [source,groovy,indent=0]
 ----
 	plugins {
-		id "com.gorylenko.gradle-git-properties" version "1.4.17"
+		id "com.gorylenko.gradle-git-properties" version "1.4.21"
 	}
 ----
 
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
@@ -2564,7 +2564,7 @@ plugin, as shown in the following example:
 [source,groovy,indent=0]
 ----
 	plugins {
-		id "com.gorylenko.gradle-git-properties" version "1.4.21"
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -2564,7 +2564,7 @@ plugin, as shown in the following example:
 [source,groovy,indent=0]
 ----
 	plugins {
-		id "com.gorylenko.gradle-git-properties" version "1.4.17"
+		id "com.gorylenko.gradle-git-properties" version "1.5.0"
```
snicoll: The latest release seems to be `1.4.21`. The [Gradle portal](https://plugins.gradle.org/plugin/com.gorylenko.gradle-git-properties) provides the same info. Where is that `1.5.0` coming from?

rajadilipkolli: @snicoll  When I check in gradle portal it is 1.4.21 but from the original developer, it is 1.5.0 . hence changed to 1.5.0
Reference : https://github.com/n0mer/gradle-git-properties

wilkinsona: As far as I can tell, the README in that repository is incorrect. Version 1.5.0 has yet to be released. @rajadileepkolli Can you please update this PR accordingly? We'd also greatly appreciate it if you did some due diligence and checked that your proposed changes will work before submitting them.

snicoll: It's a version hardcoded in a README. it's not on the gradle portal, nor anywhere else as far as I can see. I've updated the doc to `1.4.21`.

rajadilipkolli: @wilkinsona , My bad it's my mistake, Initially I have tested locally with 1.4.21 and raised pull request with it only when providing the link I have misread and changed the value to 1.5.0.

I will take care that this won't happens again.

PR #12958, opened 2018-04-24
SpringBootExceptionHandler does not identify a log configuration message if its exception is wrapped in an InvocationTargetException

henri-tremblay:
The exception thrown by logback can be wrapped in an `InvocationTargetException`. In this case, it isn't forwarded to the parent and thus defeats the purpose.

This patch is solving that in a style similar to what is done in `isRegistered`.


wilkinsona:
Thank you very much for making your first contribution to Spring Boot, @henri-tremblay. I have merged your changes in 1.5.x and forwards into 2.0.x and master. I also applied a [little bit of polish](https://github.com/spring-projects/spring-boot/commit/678e12572093f5a0e04a4cc626eb4ab8980c3191) to align the changes with the project's coding style and conventions. Thanks again.

PR #12964, opened 2018-04-25
Add testRollbackOnUpdate property. Update Liquibase to version 3.6

vkiriushkin:
New liquibase property that enables testing the rollback before executing update added to LiquibaseProperties. This functionality was added in liquibase-core in version 3.6 (https://github.com/liquibase/liquibase/pull/571).


pivotal-issuemaster:
@vkiriushkin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12964)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12964) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@vkiriushkin Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=12964)!

spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -812,6 +812,16 @@
         "level": "error"
       }
     },
+    {
+      "name": "liquibase.test-rollback-on-update",
```
snicoll: No need to provide deprecation metadata for a property that was introduced after we move the namespace. This entry should be removed.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/liquibase/CommonsLoggingLiquibaseLogger.java
```diff
@@ -32,107 +32,109 @@
  */
 public class CommonsLoggingLiquibaseLogger extends AbstractLogger {
```
snicoll: Looks like that file was touched for no good reason. Can you please revert that part?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/liquibase/CommonsLoggingLoggerFactory.java
```diff
@@ -0,0 +1,52 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+/**
+ * Liquibase {@link liquibase.logging.LoggerFactory} that delegates to an Apache Commons {@link Log}.
+ *
+ * @author Vladyslav Kiriushkin
+ * @since 2.1.0
+ */
+public class CommonsLoggingLoggerFactory extends AbstractLoggerFactory {
```
snicoll: Ditto?

spring-boot-samples/spring-boot-sample-liquibase/pom.xml
```diff
@@ -31,6 +31,7 @@
 		<dependency>
 			<groupId>org.liquibase</groupId>
 			<artifactId>liquibase-core</artifactId>
+			<version>3.6.0</version>
```
snicoll: The version should not be specified here.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/liquibase/CommonsLoggingLiquibaseLogger.java
vkiriushkin: The functionality about rollback was introduced in liquibase-core 3.6 and at the same time logging was changed in that version. In order to compile the project after switching to that version I had to change logging implementation. You suggest that this part shouldn't be included in this PR as it will be done separately when upgrading to version 3.6. Did I get it right?

snicoll: Thanks for the feedback. Can you please submit a PR to upgrade to 3.6 only? You can then have this PR cleanly on top with only the feature that is targeted here.

vkiriushkin:
@snicoll I've left only changes required to update the version of liquibase to 3.6 first.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/liquibase/CommonsLoggingLoggerFactory.java
wilkinsona: @vkiriushkin Does Liquibase require this class to be public? If it doesn't, I'd prefer that it were package-private.

snicoll:
I’ve also noticed that the MDC implementation uses slf4j. I wonder if that’s a sensible thing to do

vkiriushkin:
Can rebuild be triggered without new commit? Test failed due to netty port already in use exception.

snicoll:
@mbhave I thought that closing and reopening the PR would trigger a build again but that doesn't seem to be the case. Is there any way to trigger a build again? Thanks!

vpavic:
Updating the (last) commit to generate new checksum and force pushing should work, i.e. `git commit --amend --no-edit && git push --force` .

mbhave:
@snicoll I don't think there's a way to do that atm and the only way to re-trigger the PR build is to update the last commit. See [this](https://github.com/jtarchie/github-pullrequest-resource/issues/46)

vkiriushkin:
Updating last commit triggered rebuild, thanks! 

But every time NettyReactiveWebServerFactoryTests fails with 
```
Stacktrace was: org.springframework.boot.web.server.WebServerException: Unable to start Netty
...
Caused by: reactor.core.Exceptions$ReactiveException: io.netty.channel.unix.Errors$NativeIoException: bind(..) failed: Address already in use
```

Can you suggest what is the reason for that? As locally everything builds successfully

snicoll:
We are aware of those ci only glitches and are working on it. Don’t worry about that in the context of this PR

vkiriushkin:
Thanks for the feedback. Waiting for approve then.

vkiriushkin:
Rebased on top of latest master, now build have passed. 
Please merge when possible, so I can create PR that adds new liquibase functionality.

Issue #11379, opened 2017-12-19
Support for Request timeout in spring rest template


ChristianFinckler:
I use spring rest template for performing a head request with a proxy server.
The target host does not exist and the proxy just does not finish the request. The problem is, that neither the connect timeout nor the read timeout stop the processing.

My code:
```
final ResponseEntity<String> sourceRes = restTemplate.exchange(head(new URI(post.getSourceUrl())).build(), String.class);
        return sourceRes.getStatusCode().is2xxSuccessful();
```
Here is my code for setting up the resttemplate:
 ```
final HttpClientBuilder clientBuilder = HttpClientBuilder.create();
    clientBuilder.useSystemProperties();
    clientBuilder.setDefaultRequestConfig(custom()
      .setCookieSpec(STANDARD)
      .setConnectTimeout(Duration.ofSeconds(1).toMillis())
      .setSocketTimeout(Duration.ofSeconds(1).toMillis())
      .build());
    final HttpClient httpClient = clientBuilder.build();
    final HttpComponentsClientHttpRequestFactory requestFactory = new HttpComponentsClientHttpRequestFactory(httpClient);
    requestFactory.setConnectTimeout(Duration.ofSeconds(1).toMillis());
    requestFactory.setReadTimeout(Duration.ofSeconds(1).toMillis());

    final RestTemplate restTemplate = new RestTemplate(requestFactory);
```

Here is the request performed with curl:
```
 curl -X HEAD --proxy "proxyuser:proxypassword@proxyhost:1234t"  "https://iAmNotExisting.com/dummy.jpg" --verbose
* About to connect() to proxy proxyhost port (#0)
*   Trying 15.112.12.32...
* Connected to proxyhost (15.112.12.32) port 1234 (#0)
* Establish HTTP proxy tunnel to iAmNotExisting.com:443
* Proxy auth using Basic with user 'proxyuser'
> CONNECT iAmNotExisting.com:443 HTTP/1.1
> Host: iAmNotExisting.com:443
> Proxy-Authorization: Basic blablub=
> User-Agent: curl/7.29.0
> Proxy-Connection: Keep-Alive
> 
* Operation timed out after 300314 milliseconds with 0 out of 0 bytes received
* Closing connection 0
curl: (28) Operation timed out after 300314 milliseconds with 0 out of 0 bytes received
```

As you can see, curl sends the request headers, but does not receive any response.

[...]

PR #12972, opened 2018-04-25
Added http HealthIndicator

HarryEMartland:
I have added a new Health Indicator and Auto Configuration for it.
This indicator makes a get request to a url reporting if the response is ok.
I keep writing and seeing people write similar things to this over and over again at work.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/http/HttpHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/http/HttpHealthIndicator.java
@@ -1,0 +1,78 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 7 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/http/HttpHealthIndicator.java
```diff
@@ -0,0 +1,78 @@
...
+	 * @param url the url to be checked
+	 */
+	public HttpHealthIndicator(HttpClient httpClient, String url) {
+		this.httpClient = httpClient;
+		this.url = url;
+	}
+
+	@Override
+	protected void doHealthCheck(Health.Builder builder) {
+		HttpGet httpGet = new HttpGet(this.url);
```
spencergibb: HttpClient is optional, might be better to use RestTemplate.

snicoll:
Thanks for the PR. A `HealthIndicator` is focused on a component and not a "technical mean to check if a service is up". As such, I am not keen to introduce a generic http health indicator as it does not fit with the general idea of an health indicator.

bclozel:
I'm not in favor of doing this for the reasons @snicoll mentioned. In this case, the application should use the client library or the actual HTTP client (with credentials?) being used to check that the component is working - which is much more than just checking that the remote HTTP endpoint is up.

There are a lot of subtle limitations and configuration aspects to it - which explains why it should be a custom indicator in the first place.

@HarryEMartland also as a side note, if you're using this indicator in your own project - the configured timeout won't work if the remote endpoint is HTTPS or if you need to go through a proxy (see #11379).

wilkinsona:
I agree with @snicoll and @bclozel.

snicoll:
Thanks all.

@HarryEMartland, we won't merge the PR for the reasons described above. Thanks anyway!

PR #12977, opened 2018-04-26
Optimize deducing main class in SpringApplication

dreis2211:
Hi,

I found a small optimization opportunity in `SpringApplication.deduceMainApplicationClass()` that iterates over a stacktrace to find the `main` method. Those methods are usually at the bottom of the stack, so we can start iterating from there.

This should save a few cycles - especially in tests where the `main` method is usually originating from some sort of test runner that creates some more stack elements.

Cheers,
Christoph


dreis2211:
Given this some additional thought, this change - although not reproducable in tests - might change the outcome of the method if multiple main methods are found. Although unlikely, feel free to decline the PR in case of concerns.

snicoll:
I'd prefer the way it is now as it makes us independent of whatever is launching the process (IMO, it's much easier to spot the method name clash the other way around).

Thanks for the PR anyway!

dreis2211:
As said - fair enough :-) Thanks for getting back to me!

PR #12981, opened 2018-04-27
Polish

izeye:


snicoll:
Thanks again @izeye - I've split the PR in two commits as most (but not all) of it was applicable to `2.0.x`.

PR #12984, opened 2018-04-27
Add a missing '=' condition to RandomAccessDataFile.read()

izeye:
This PR adds an `=` condition to `RandomAccessDataFile.read()` as it looks missed.


wilkinsona:
Thanks, @izeye. Reviewing this has made me realise that we need to specify the behaviour of `RandomAccessData.read(long, long)` more precisely. I've opened #12986. I'm going to close this PR as I suspect that we'll end up doing something different to the change that's proposed here.

PR #13003, opened 2018-04-30
Only display config delta on DEBUG log level when using DevTools

alek-sys:
To make DevTools behaviour consistent with [documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/howto-spring-boot-application.html#howto-troubleshoot-auto-configuration) and `ConditionEvaluationReportLoggingListener` class, only print configuration delta on debug level. Otherwise just report about the difference without acutally listing auto-configurations changed.


alek-sys:
Thanks for clarification, Andy! I was trying to silent the delta via changing the log level, but `spring.devtools.restart.log-condition-evaluation-delta=false` seems to be a much better option.

Issue #8607, opened 2017-03-14
Add @DataCassandraTest support


philwebb:
For completeness it would be nice if we had a complete set of test annotations for Spring Data. See #8544

Razi007:
I want to work on it.I am going through code base to understand for now.Also can i get some advice or mentorship.Or is there any irc channel or mailing list where i can collaborate with spring-boot community.

snicoll:
@Razi007 we have [a gitter channel](https://gitter.im/spring-projects/spring-boot).

Razi007:
can i pick this task as beginner in spring-boot code base

snicoll:
@Razi007 please use the Gitter channel if you have questions.

izeye:
Based on https://github.com/spring-projects/spring-boot/pull/13033#issuecomment-400430466 , it looks this can be closed.

wilkinsona:
Thanks, @izeye. I'll label this one as blocked for now. We'd still like to add support, but cannot do so without a project for Cassandra that's similar to [Embedded Mongo](https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo).

wilkinsona:
On second thoughts (prompted by @snicoll), let's close this one as an Embedded Mongo-like project becoming sufficiently established for us to depend upon it is out of our control. We can re-open this issue should it happen.

vpavic:
Couldn't you simply use Testcontainers here? IMO that's a good for _embedding_ pretty much anything that's available as a Docker container.

PR #13033, opened 2018-05-02
@DataCassandraTest(gh-8607)

nosan:
[see gh-8607](https://github.com/spring-projects/spring-boot/issues/8607).

This pull request includes:
 - `@DataCassandraTest` 
  Annotation that can be used in combination with `@RunWith(SpringRunner.class)` for a typical Cassandra test.
 - `EmbeddedCassandraServer` 
  A simple implementation of embedded Cassandra server. **The server will be run under a new Java process.**

```
@RunWith(SpringRunner.class)
@DataCassandraTest
public class DataCassandraTestIntegrationTests {
	@Autowired
	private CassandraTemplate cassandraTemplate;
	@Autowired
	private ExampleRepository exampleRepository;

	@Test
	public void testRepository() {
		City city = new City();
		city.setId(Long.MAX_VALUE);
		city.setName("Lviv");
		city = this.exampleRepository.save(city);
		assertThat(city.getId()).isNotNull();
		assertThat(this.cassandraTemplate.exists(city.getId(), City.class))
				.isTrue();
	}
}
```

p.s. I also left some comments related to maven exclusion. Please pay attention :)


nosan:
@philwebb @snicoll 
Could you please check it if you are not checking by now. Thanks in advanced.

snicoll:
@nosan I did already. I am not sure about embedding Cassandra that way but I have nothing better to offer at the moment. I am not sure Spring Boot should do this and if there isn't a way to officially embed the store, we shouldn't offer a dedicated slice annotation for it.

Let's see what the rest of the team thinks. If we decide to pursue, I'll have a deeper look (it looks very complete, thank you!).

nosan:
I added `@PropertyMapping` instead of the customizer, but am not sure about this.
Maybe we should use a custom customizer which registers `PropertySources` with the lowest order, so then `application.properties` or `@TestPropertySource` will be more prioritized.
```
private static final class AutoConfigureCassandraSessionContextCustomizer
			implements ContextCustomizer {
		private final AutoConfigureCassandraSession annotation;

		private AutoConfigureCassandraSessionContextCustomizer(AutoConfigureCassandraSession annotation) {
			this.annotation = annotation;
		}

		@Override
		public void customizeContext(ConfigurableApplicationContext context, MergedContextConfiguration mergedConfig) {
			Map<String, Object> props = new LinkedHashMap<>();
			ConfigurableEnvironment environment = context.getEnvironment();
			props.put("spring.data.cassandra.port", this.annotation.port());
			props.put("spring.data.cassandra.cluster-name", this.annotation.clusterName());
			props.put("spring.data.cassandra.keyspace-name", this.annotation.keyspaceName());
			props.put("spring.data.cassandra.contact-points", this.annotation.contactPoints());
			props.put("spring.data.cassandra.schema-action", this.annotation.schemaAction());
			props.put("spring.data.cassandra.test.keyspace-action", this.annotation.keyspaceAction());
			environment.getPropertySources()
					.addLast(new MapPropertySource("@AutoConfigureCassandraSession", props));
		}

		@Override
		public int hashCode() {
			return getClass().hashCode();
		}

		@Override
		public boolean equals(Object obj) {
			return (obj != null && obj.getClass() == getClass());
		}

	}
```

nosan:
@snicoll 
Did you have time to discuss whether this approach is suitable for `Spring Boot Team` or not?

snicoll:
@nosan we did and @philwebb wanted to look at the code a bit more before getting back to you. Thanks for bearing with us on this one.

philwebb:
Sorry @nosan, my daughter has been sick so I've not had the chance to review.

nosan:
@philwebb everything is ok. I hope your daughter gets well soon.

nosan:
@philwebb @snicoll What do you think, if I will create an embedded server separately as a library and will change this PR?  I'd like to use the same approach as `Embedded Mongo`.

philwebb:
@nosan Thanks for the PR and sorry it's taken me so long to get back to you. After some consideration, I think it would be better for the test support to live outside of Spring Boot for now. 

Having a separate library definitely seems like the right approach but I think the test support should also be separate, at least until the library becomes established. We can certainly reconsider things at a later date if needed. We can also make any minor changes that might be necessary if additional hook points are needed (for example in `CassandraAutoConfiguration`).

Issue #13026, opened 2018-05-02
It is unclear that configuring a JNDI name for the auto-configured mail sender's session causes property-based configuration to back off


iyerha:
The following bean definition in MailSenderAutoConfiguration does not set any mail properties when configuring the JavaMailSenderImpl using the session, especially the default encoding. This causes any message created via the MimeMessageHelper to generate the wrong content type.

```
@Bean
public JavaMailSenderImpl mailSender() {
	JavaMailSenderImpl sender = new JavaMailSenderImpl();
	if (this.session != null) {
		sender.setSession(this.session);
	}
	else {
		applyProperties(sender);
	}
	return sender;
}
```

wilkinsona:
Thanks for the report. Looks like this is an oversight from when the JNDI support was [added](https://github.com/spring-projects/spring-boot/commit/28cb13c31d3a33c5c728d8db0aa525f16e9330f8). As far as I can tell, the only properties that shouldn't be applied when using the `Session` from JNDI are the JavaMail properties (calling `setJavaMailProperties(Properties)` sets the season to null).

snicoll:
Closing in favour of PR #13050

iyerha:
AFAIK the only way to set an encoding on the session is to set the 'mail.mime.charset' property. However, this is never retrieved by the JavaMailSenderImpl (in contrast, the 'mail.transport.protocol' is processed by getTransport( ) from the session). So this doesn't work.
I have worked around it by post processing the bean and setting the default encoding in code. It would be nicer to have this work by configuration, though.

iyerha:
I guess in case of JNDI you would have to set the 'mail.mime.charset' system property as the default encoding.

snicoll:
@iyerha That's correct, the encoding is the only thing that is specific to `JavaMailSender` that we can't get from the `Session` so we should use that information and clarify what properties get ignored.

PR #13050, opened 2018-05-04
Allow to set MailProperties when JNDI is used

eddumelendez:
See gh-13026

### DIFF
--- a/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mail/MailSenderAutoConfiguration.java
+++ b/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mail/MailSenderAutoConfiguration.java
@@ -69,9 +69,10 @@ public class MailSenderAutoConfiguration {
 		JavaMailSenderImpl sender = new JavaMailSenderImpl();
 		if (this.session != null) {
 			sender.setSession(this.session);
+			applyProperties(sender);
 		}
 		else {
-			applyProperties(sender);
+			applyJavaMailProperties(sender);
 		}
 		return sender;
 	}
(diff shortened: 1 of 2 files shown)


spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mail/MailSenderAutoConfiguration.java
```diff
@@ -69,9 +69,10 @@ public JavaMailSenderImpl mailSender() {
 		JavaMailSenderImpl sender = new JavaMailSenderImpl();
 		if (this.session != null) {
 			sender.setSession(this.session);
+			applyProperties(sender);
```
snicoll: On a second thought, that doesn't sound right at all that we set username/password/host when we already have a `Session` from JNDI that should give us that stuff

spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/mail/MailSenderAutoConfigurationTests.java
```diff
@@ -149,9 +149,10 @@ public void jndiSessionAvailable() throws NamingException {
 		Session session = configureJndiSession("foo");
 		load(EmptyConfig.class, "spring.mail.jndi-name:foo");
 		Session sessionBean = this.context.getBean(Session.class);
+		JavaMailSenderImpl mailSender = this.context.getBean(JavaMailSenderImpl.class);
 		assertThat(sessionBean).isEqualTo(session);
-		assertThat(this.context.getBean(JavaMailSenderImpl.class).getSession())
-				.isEqualTo(sessionBean);
+		assertThat(mailSender.getSession()).isEqualTo(sessionBean);
+		assertThat(mailSender.getDefaultEncoding()).isEqualTo("UTF-8");
```
snicoll: That test is testing by side effect. Perhaps it should test what the expectations are instead.

eddumelendez:
no problem @wilkinsona I had my thoughts about this but my fault to not let you know

Issue #9683, opened 2017-07-05
Better detection of Spring Session store


snicoll:
With the modules split in Spring Session 2, there is a high chance that the project's classpath contains only one store implementation.

We should improve our auto-configuration structure to switch to that store in that case (and that case only) so that setting the property is not required.

There are two failures scenario to handle:

* More than one implementation is available: report that the store-type parameter is not set and which implementations are current available
* No implementation is available (that could happen for those migrating and having only `spring-session-core` on their classpath that used to contain the core implementations). Perhaps reusing the `NoSuchBeanDefinition` failure analyzer could be a good idea as it would tell why the stores were not configured. If Spring Session provides a way to discover the available implementations (as Spring Data does with `spring.factories` ) we could build a more dedicated error message.

/cc @rwinch @vpavic

snicoll:
@rwinch `MapSessionRepositry` is always available. Do I understand that's what you'd like Spring Boot to auto-configure then? It's quite consistent with what we do with caching but I'd like to make sure this is well understood.

In the description above, 2 is invalid so Spring Session would use the fallback automatically if no other store is available. I am not entirely convinced this is an improvement as we ruled out Redis to be used automatically in 1.5.

vpavic:
@snicoll We've made one significant change to `MapSessionRepository` recently in spring-projects/spring-session#866. As a consequence now users are now required to explicitly supply the `Map` used to store the sessions (i.e. there's no default constructor any more).

So our preference would actually be to remove the auto-config support for `MapSessionRepository` as Spring Session isn't providing any configuration facilities for it either.

snicoll:
Looks like the stars are aligning then. I've created #10253

snicoll:
Failure analyzer for the first case

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Multiple Spring Session store implementations are available on the classpath:
	-org.springframework.session.jdbc.JdbcOperationsSessionRepository
	-org.springframework.session.hazelcast.HazelcastSessionRepository


Action:

Consider any of the following:
	- Define the `spring.session.store-type` property to the store you want to use
	- Review your classpath and remove the unwanted store implementation(s)
```

snicoll:
Closed by https://github.com/spring-projects/spring-boot/commit/4670cc779548677c5286fa5f1a5beee9b08a61eb

PR #13061, opened 2018-05-04
Improve Spring Session docs

vpavic:
This PR improves Spring Session docs to state that `spring.session.store-type` property is not required when single Spring Session module is present on the classpath - see #9683.


snicoll:
Thanks @vpavic, I feel bad for overlooking that one considering the amount of discussion we had on the subject !

PR #13066, opened 2018-05-04
PageableHandlerMethodArgumentResolver not registered for WebMvcTest 

### LINKED ISSUE #12938: PageableHandlerMethodArgumentResolver not registered for WebMvcTest
While testing a controller that receives a `Pageable` parameter the test fails with this exception: `java.lang.NoSuchMethodException: org.springframework.data.domain.Pageable.<init>()`

If I register manually `PageableHandlerMethodArgumentResolver` then the test passes.

Spring version is 2.0.1.RELEASE

You can reproduce this behavior with this [project](https://github.com/ideaplugins/springbootissue1)

`FailingTest` reproduces the problem.

`PassingTest` shows the workaround.

ideaplugins:
Adds `SpringDataWebAutoConfiguration` when running `WebMvcTest`s. By doing this argument resolvers for `Sort` and `Pageable` get registered.
Fixes #12938


snicoll:
Thanks for the PR @ideaplugins - I've decided to roll a different fix with a smaller surface area, see  360f4e1.

PR #13117, opened 2018-05-09
Jersey auto-configuration does not find @ApplicationPath on a ResourceConfig created from an Application

### LINKED ISSUE #13116: Support for using WrappingResourceConfig with Jersey autoconfiguration
I tried using WrappingResourceConfig with Jersey like the following:

I defined a standard jax-rs application:
``` java
package my.wonderful.package;
@Named
@ApplicationPath("/api")
public MyApp extends Application {}
```

``` java
package my.wonderful.package;

@Path("/test")
public TestResource {
}
```

I then tried to wrap this application into a Jersey ResourceConfig:
``` java
@Configuration
public class JerseyConfiguration {
    @Bean
    public ResourceConfig conf(MyApp app) {
        config = ResourceConfig.forApplication(app);
        config.packages(app.getClass().getPackage().getName());
        return config;
    }
}
```

I expected my resource to be exposed behind /api/test, but that doesn't work:
``` bash
$ curl -s -o /dev/null -w %{http_code} localhost:8080/api/test
404
```
It is however exposed behind /test:
``` bash
$ curl -s -o /dev/null -w %{http_code} localhost:8080/test
200
```

I looked around the code, and the issue seems to be in the autoconfiguration not supporting WrappingResourceConfig:

https://github.com/spring-projects/spring-boot/blob/073849b9e5c802b85e36484694bbf11cd5f530db/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jersey/JerseyAutoConfiguration.java#L119-L120

That part should be using this instead:
``` java
this.config.getApplication().getClass()
```
Then the annotation is always picked from the right place:
If it's a regular ResourceConfig it's 'this', if its a WrappingResourceConfig, then it's the class of the wrapped Application class.

Although best would be to support also not just application instances but classes as well.

reegnz:
Fixes #13116

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jersey/JerseyAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jersey/JerseyAutoConfiguration.java
@@ -117,7 +117,7 @@ public class JerseyAutoConfiguration implements ServletContextAware {
 		}
 		else {
 			this.path = findApplicationPath(AnnotationUtils
-					.findAnnotation(this.config.getClass(), ApplicationPath.class));
+					.findAnnotation(this.config.getApplication().getClass(), ApplicationPath.class));
 		}
 	}
 


pivotal-issuemaster:
@reegnz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13117)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13117) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@reegnz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13117)!

snicoll:
@reegnz thank you for your first contribution to Spring Boot. I've added a test in https://github.com/spring-projects/spring-boot/commit/b48f34117ef637ca98eedc57f37c0588d4216efe

PR #13124, opened 2018-05-09
ServletEndpoints do not consider server.servlet.path 

gmcouto:
### Summary
I have changed ServletEndpointManagementContextConfiguration and documentation to support ServletEndpoints to have same path than web management endpoints.

### The problem
**If your server has no management port set** and there is a "server.servlet.path" set, the ServletEndpoint registration path would not consider "server.servlet.path" as all other web management endpoints. So the resulting path for servlet endpoints will look different from other web management paths.

In this case, for example, with jolokia:
How the jolokia endpoint is **currently**:
http://host:port/context-path/actuator/jolokia

Examples of other endpoints that I am trying to mimic:
http://host:port/context-path/servlet-path/actuator/health
http://host:port/context-path/servlet-path/actuator/auditevents
http://host:port/context-path/servlet-path/actuator/beans

How the jolokia endpoint **should** be:
http://host:port/context-path/servlet-path/actuator/jolokia

### The fix
This pull requests ensures the desired behaviour, and add to the documentation the information that it does consider "server.servlet.path" whenever management port is not set.

The way it was implemented is that ServletEndpointManagementContextConfiguration will have two configurations, one for different context (different management port), and another for same context.

The configuration for different context will be the same as before.
The configuration for same context will pull "server.servlet.path" from environment and add it to the servlet registration base path.


mbhave:
Thanks for the PR @gmcouto but I have a slightly different [fix](https://github.com/spring-projects/spring-boot/commit/c937bb68a03de950a1fd91f58b1a84c2e4cb4ad2) for it.

PR #13130, opened 2018-05-09
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
@@ -20,7 +20,7 @@ content into your application. Rather, pick only the properties that you need.
 	# COMMON SPRING BOOT PROPERTIES
 	#
 	# This sample file is provided as a guideline. Do NOT copy it in its
-	# entirety to your own application.			   ^^^
+	# entirety to your own application.               ^^^
 	# ===================================================================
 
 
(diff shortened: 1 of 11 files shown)


spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
```diff
@@ -20,7 +20,7 @@ content into your application. Rather, pick only the properties that you need.
 	# COMMON SPRING BOOT PROPERTIES
 	#
 	# This sample file is provided as a guideline. Do NOT copy it in its
-	# entirety to your own application.			   ^^^
+	# entirety to your own application.
```
snicoll: That has moved but the purpose is to emphasis the `NOT` part. Can you please rework the PR to restore it (under `NOT`)?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/SpringConfigurationPropertySources.java
```diff
@@ -91,10 +91,10 @@ public boolean hasNext() {
 		@Override
 		public ConfigurationPropertySource next() {
 			ConfigurationPropertySource next = fetchNext();
-			this.next = null;
 			if (next == null) {
```
snicoll: Oops. Looks like we have a test missing as well there.

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
izeye: @snicoll I suspected so but just removed it as "NOT" is already emphasized by making it upper case and adding "^^^" to the next line of the word to emphasize looks unfamiliar to me. The "unfamiliar" part might be only for me, so if you confirm once more I'll restore and relocate it in the indented position.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/SpringConfigurationPropertySources.java
izeye: @snicoll This change is just for avoiding a `null` assignment to `this.next` when `next` is `null` as `this.next` is already `null`, so it should not change practically nothing. Am I missing something?

snicoll: You don't but I certainly did 😂

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
snicoll: A double emphasis doesn't hurt as we really don't want users to do that. I agree with the unfamiliar part but that might actually allows us to reach our goal in the end :)

izeye:
@snicoll Thanks for the quick feedback!

I restored and relocated "^^^" in the intended position as you suggested.

snicoll:
Thanks again @izeye!

PR #13137, opened 2018-05-10
Add .gitignore for Gradle Plugin module

vpavic:
If one attempts to compile Spring Boot project in IntelliJ, there are errors in `spring-boot-gradle-plugin` module unless that module is imported as Gradle project. Doing so (by default) generates `out/` directory with compilation output and these files are not ignored by Git ATM.

This PR ~updates `.gitignore` to add `out/` pattern~ adds .gitignore for Gradle Plugin module.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/.gitignore
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/.gitignore
@@ -1,0 +1,3 @@
+/bin/
+/build/
+/out/


.gitignore
```diff
@@ -38,3 +38,4 @@ transaction-logs
 secrets.yml
 .gradletasknamecache
 .sts4-cache
+out/
```
snicoll: I am not super keen to add such a global pattern. Perhaps we would make that a bit more specific? See also https://github.com/spring-io/initializr/pull/657/files

vpavic: That one's a quite common `.gitignore` entry for IntelliJ. For instance, see GitHub's own [Jetbrains `.gitignore` template](https://github.com/github/gitignore/blob/master/Global/JetBrains.gitignore#L33). That being said, I'm fine with it being updated to something more specific if you prefer.

wilkinsona:
I agree on making this more specific. In fact, I'd do the same for `bin` too. It's Eclipse's equivalent of `out` and also only applies to the Gradle plugin. While we're at it, we could do something for `build` too.

There's more than one way to make the ignore more specific, but I think I'd create a separate `.gitignore` file in the Gradle plugin's directory.  That'd mean the new file had the following contents:

```
/bin/
/build/
/out/
```

vpavic:
Thanks for the feedback - I've updated the PR according to separate `.gitignore` file suggestion.

PR #13136, opened 2018-05-10
Improve LDAP auto-configuration

vpavic:
Auto-configuration of LDAP's `LdapTemplate` is currently a part of `LdapDataAutoConfiguration` which is conditional of presence of `LdapRepository` (i.e. Spring Data LDAP). This arrangement isn't ideal (and doesn't make much sense) since the `LdapTemplate` is a part of Spring LDAP project, and therefore should not be tied to Spring Data LDAP.

This PR improves and simplifies LDAP auto-configuration by moving `LdapTemplate` configuration to `LdapAutoConfiguration`. Consequently, `LdapDataAutoConfiguration` is not needed anymore and is removed.

PR #13143, opened 2018-05-11
Improve LDAP auto-configuration conditions

vpavic:
At present, auto-configuration of `LdapContextSource` is conditional on presence of `ContextSource` bean. However, there are valid use cases which require multiple `ContextSource` bean, for instance `PooledContextSource`. With the current arrangement, the auto-configuration of `LdapContextSource` will back off if user provides `PooledContextSource` bean, while it would still be reasonable to reuse auto-configured `LdapContextSource`.

This PR improves `LdapContextSource` factory method return value and condition to back off only if users actually provide `LdapContextSource` bean themselves.

With this change in place, users can provide `PooledContextSource` bean and reuse auto-configured `LdapContextSource`. `PooledContextSource` bean should then be marked as `@Primary` to ensure it is injected in auto-configured `LdapTemplate`.


vpavic:
> I've polished the PR in 8e9a873.

Thanks for that - I rebased on top of changes from #13136 in rush and messed up that test.

PR #13159, opened 2018-05-14
Add liquibase test rollback on update property

vkiriushkin:
New liquibase property that enables testing the rollback before executing update added to LiquibaseProperties. This functionality was added in liquibase-core in version 3.6 (liquibase/liquibase#571).


spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -818,6 +818,12 @@
         "level": "error"
       }
     },
+    {
+      "name": "liquibase.test-rollback-on-update",
+      "type": "java.lang.Boolean",
+      "description": "Enable test rollback before starting migration.",
+      "defaultValue": false
+    },
```
snicoll: That section shouldn't be there. The keys around this one are there when we moved `liquibase` to `spring.liquibase` (look at the derepcation). Keys introduced after that migration do not need any warning.

snicoll:
@vkiriushkin thank you for your first contribution to Spring Boot. This is now merged in `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/c7eec5c45e4f44bdd5913e67c2683b0281b9b6e5).

PR #13175, opened 2018-05-15
Respect lombok.AccessLevel attributes on creating property metadata

### LINKED ISSUE #13172: ConfigurationProperties metadata contains fields with non-public setter/getter from Lombok
Property metadata contains properties for fields which have a non-public (or none) setter or getter when using lombok. The bug occures if there are lombok annotations used, which define the `lombok.AccessLevel` as annotation attribute.

Example:

```java 
@Data
@ConfigurationProperties("test.lombok")
public class LombokProperties {

	@Getter(AccessLevel.PRIVATE)
	@Setter(AccessLevel.PRIVATE)
	private String shouldNotAppear;

	private String shouldAppear;

}
```

Property `test.lombok.should-not-appear` is contained in the metadata, but it should not.

Behaviour in spring-boot-configuration-processor: `ConfigurationMetadataAnnotationProcessor` only checks if lomboks `@Getter` or `@Setter` are used, not if any attributes were set.

See my example project to verify the bug: https://github.com/kesslerj/spring-boot-properties-metadata-lombok-bug

I started fixing the issue and will offer a pull request soon.

kesslerj:
## Description
Property metadata contains properties for fields which have a non-public (or none) setter or getter when using lombok. The bug occures if there are lombok annotations used, which define the `lombok.AccessLevel` as annotation attribute.

Example:

```java 
@Data
@ConfigurationProperties("test.lombok")
public class LombokProperties {

	@Getter(AccessLevel.PRIVATE)
	@Setter(AccessLevel.PRIVATE)
	private String shouldNotAppear;

	private String shouldAppear;

}
```

Property `test.lombok.should-not-appear` is contained in the metadata, but it should not.

Behaviour in spring-boot-configuration-processor: `ConfigurationMetadataAnnotationProcessor` only checks if lomboks `@Getter` or `@Setter` are used, not if any attributes were set.

See my example project to verify the bug: https://github.com/kesslerj/spring-boot-properties-metadata-lombok-bug

## Discussion
My first PR here, so let me know if I missed any contribution rule.

The part I really don't like in my solution is this one:

```java
private boolean isLombokPublic(AnnotationMirror lombokAnnotation) {
		return lombokAnnotation.getElementValues().values().stream()
				.noneMatch(e -> e.toString().startsWith(LOMBOK_ACCESS_LEVEL)
						&& !e.toString().equals(LOMBOK_ACCESS_LEVEL_PUBLIC));
}
```

I did not find a solution without using `toString`. Any suggestions are welcome.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessor.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessor.java
@@ -63,6 +63,7 @@ import org.springframework.boot.configurationprocessor.metadata.ItemMetadata;
  * @author Stephane Nicoll
  * @author Phillip Webb
  * @author Kris De Volder
+ * @author Jonas Keßler
  * @since 1.2.0
  */
 @SupportedAnnotationTypes({ "*" })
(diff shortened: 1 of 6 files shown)


snicoll:
@kesslerj thank you very much for your first contribution to Spring Boot. This [polished commit](https://github.com/spring-projects/spring-boot/commit/563d7d70a23a318959cd162b6480b8f31befd531) should answer your concern about the `toString` issue. Thanks again!

kesslerj:
@snicoll it was a pleasure to contribute to a project which I am using nearly daily. Thanks for the fast responses.

PR #13192, opened 2018-05-16
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/AbstractHealthAggregator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/AbstractHealthAggregator.java
@@ -16,10 +16,10 @@
 
 package org.springframework.boot.actuate.health;
 
-import java.util.ArrayList;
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ErrorProperties.java
```diff
@@ -74,10 +74,6 @@ public Whitelabel getWhitelabel() {
 		return this.whitelabel;
 	}
 
-	public void setWhitelabel(Whitelabel whitelabel) {
```
snicoll: I'd like to merge this in `2.0.x`. This is correct but I prefer to make such a breaking change on `master` only. Can you please remove it? (I can apply that part manually or feel free to submit a separate PR for that).

izeye: @snicoll Thanks for the quick feedback!

I reverted it in this PR and created another one for it: https://github.com/spring-projects/spring-boot/pull/13197

Issue #13186, opened 2018-05-15
Include DispatcherServlets only exposed as a ServletRegistrationBean in the mappings endpoint


filiphr:
When a `DispatcherServlet` is registered via:

```
@Bean
public ServletRegistrationBean<DispatcherServlet> idmServlet(MyProperties properties) {
    return registerServlet(properties.getServlet(), IdmEngineRestConfiguration.class);
}
```

Then when the `mappings` endpoint is invoked then the 
`context.application.mappings.servlet` correctly contains that servlet information. However, under `context.application.mappings.dispatcherServlets` only the `dispatcherServlet` and not my custom servlet mappings are displayed.

It would be good if `ServletRegistrationBean`(s) are also considered as part of the `MappingsEndpoint`

philwebb:
Closing in favor of PR #13211

filiphr:
@philwebb just out of my curiosity (want to understand how your labels are working). You closed this in favor of the PR (this part I get). You then marked it as duplicate, I assume a duplicate of the PR, but the PR is actually a fix for this issue.

snicoll:
github doesn’t make a difference between an issue and a PR. We’ll use the PR to track the issue described here so this issue becomes a duplicate of it.

PR #13211, opened 2018-05-19
Dispatcher servlets with a custom servlet name are not found by the mappings endpoint

filiphr:
Iterates all ServletRegistrationBeans and adds their mappings in case their Servlet is a DispatcherServlet.
Makes sure that a servlet registered through ServletRegistrationBean and exposed as a Bean is not
described twice.

Fixes gh-13186

I did this against the 2.0.x branch as I think that it can be included in a maintenance version. In case you decide that it can't, it would be unfortunate (as we would need to wait for 2.1 😄), but I can modify the PR to target the master branch

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/servlet/DispatcherServletsMappingDescriptionProvider.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/servlet/DispatcherServletsMappingDescriptionProvider.java
@@ -19,14 +19,19 @@ package org.springframework.boot.actuate.web.mappings.servlet;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
+import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Map.Entry;
+import java.util.Set;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
+import javax.servlet.Servlet;
+
 import org.springframework.boot.actuate.web.mappings.HandlerMethodDescription;
 import org.springframework.boot.actuate.web.mappings.MappingDescriptionProvider;
+import org.springframework.boot.web.servlet.ServletRegistrationBean;
 import org.springframework.context.ApplicationContext;
 import org.springframework.data.rest.webmvc.support.DelegatingHandlerMapping;
 import org.springframework.util.ClassUtils;
@@ -44,6 +49,7 @@ import org.springframework.web.servlet.mvc.method.RequestMappingInfoHandlerMappi
  * DispatcherServlets}.
  *
  * @author Andy Wilkinson
+ * @author Filip Hrisafov
  * @since 2.0.0
  */
 public class DispatcherServletsMappingDescriptionProvider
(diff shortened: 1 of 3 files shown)


snicoll:
@filiphr #13186 is assigned in the backlog (that is 2.1).

filiphr:
I thought that the reason for that was more on how you are planning to implement stuff and that you might consider accepting a PR with an implementation in an earlier version.

Is there a specific reason why 2.0.x is not acceptable? Is there something breaking in this PR?

The thing is that currently in 2.0.2 even if I expose the `DispatcherServlet` as a `Bean` Boot will only initialize the mappings properly when the bean name of the `DispatcherServlet` is the same as the servlet name in the `ServletRegistrationBean`

snicoll:
Please review the description of the Backlog milestone.

filiphr:
I saw, a guy can only hope 😄. I've changed the base of the PR from `2.0.x` to `master`.

If you can reconsider that I would be really grateful 😄

philwebb:
It might be worth using the generic rather than making `getServlet()` public.

philwebb:
Flagged for team attention to see if we'd be happy with 2.0. I think I'd be fine with that.

wilkinsona:
@filiphr made a good point [on Gitter](https://gitter.im/spring-projects/spring-boot?at=5afd9acf9ad18967b9f2ce03):

> The name is wrong. I have a custom name for the servlet in the ServletRegistrationBean and the name of the servlet is does not match that. It can't find the child. The problem is [here](https://github.com/spring-projects/spring-boot/blob/8364d121b1e39c23e872dd7ce587681e30675703/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/servlet/DispatcherServletsMappingDescriptionProvider.java#L85). The bean name is used for the name when it is searched in Tomcat

So, if you use a custom servlet name, even if you expose both the `Servlet` and the `ServletRegistrationBean` as beans, we'll fail to find it as we use the bean name rather than the servlet name. I think that should be considered to be a bug. Fixing it would require us to look for servlet registration beans as proposed here. In short, in light of that, I'm happy with 2.0.x for this too.

filiphr:
Thanks for targeting this to 2.0.x. Should I change the target branch of the PR back to 2.0.x?

Btw @philwebb you said:

>It might be worth using the generic rather than making getServlet() public.

Which generic? The [`ServletRegistrationBean#servlet`](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/ServletRegistrationBean.java#L61) is located into the the `ServletRegistrationBean` not in the parent classes. I couldn't find another way for this. Maybe I've missed something?

philwebb:
> Thanks for targeting this to 2.0.x. Should I change the target branch of the PR back to 2.0.x?

No need, we'll just cherry-pick the commit

> It might be worth using the generic rather than making getServlet() public.

I was thinking it might be possible to look at `<T>` from `ServletRegistrationBean<T>`. In hindsight, I'm not sure that's better than making `getServlet()` public. Don't bother changing that for now, we'll see what the rest of the team think is best.

filiphr:
>I was thinking it might be possible to look at `<T>` from `ServletRegistrationBean<T>`

If you were thinking about reflection and just getting the type, that won't be enough as we need the actual servlet (so the trigger can happen).

Just for info I changed it after looking at [`AbstractFilterRegistrationBean#getFilter()`](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/AbstractFilterRegistrationBean.java#L270) and it being `public`.

In any case if there is anything that needs changing from my side let me know, I have no problem with any modification, for me the functionality of it is important 😄.

filiphr:
You all probably have a lot of work on other issues and the next feature release. 

Just wanted to ask what the status for this one is? Is there something more that I can do so we can get it in the next patch for 2.0.4? Or is there something else blocking this?

snicoll:
Thanks for the PR @filiphr, I've fixed the issue with a slightly different approach, see 3423c5d for more details.

filiphr:
Thanks for fixing this @snicoll. I am a bit sad that it PR didn't get merged, but I learned something new, which is always good.

Looking forward to 2.0.4

snicoll:
If you get a chance to test the snapshot, it'd be very much appreciated.

filiphr:
@snicoll I just tried it out and it works great (as expected). Thanks for doing this improvement. It even works nicely in the IntelliJ Endpoint mappings.

There is only one small thing and that is that the Actuator endpoints are mapped in each servlet. However, this is not linked with this issue and I think you are not interested in doing something for this (which is fine for me)

Issue #13176, opened 2018-05-15
Add to JackonProperties and JacksonAutoconfiguration support to set visibility


Alexuy:
Hi!

When Spring Framework 5.1 will be released and when you upgrade spring-boot to it, please, add possibility to set visibility properties to default Jackson ObjectMapper.

See Spring Framework issue:

https://jira.spring.io/browse/SPR-16411

snicoll:
Closing in favour of PR #13214

PR #13214, opened 2018-05-19
Add support to set visibility via JacksonProperties

eddumelendez:
In spring 5.1, new method `visibility` was added to `Jackson2ObjectMapperBuilder`.
This commit introduce the support to set visibility via 
`JacksonProperties`.

See gh-13176


snicoll:
@eddumelendez thanks for the PR, looks good. We want to wait a bit before effectively switching `master` to `5.1.0.BUILD-SNAPSHOT`. I'll put this one on hold for the time being.

PR #13243, opened 2018-05-23
Polish

izeye:


snicoll:
Thanks again @izeye! FYI, we have a new way to format code that was added after you did submit this PR, see #13255 - You can run `mvn validate` as before to make sure the formatting is correct.

izeye:
@snicoll Thanks for the info! I'll check 😄

PR #13262, opened 2018-05-25
Reference the latest version of the jOOQ manual

lukaseder:
The current spring boot manual references an outdated version 3.6 of the jOOQ manual.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
@@ -3678,7 +3678,7 @@ commercial and open source editions can be used with Spring Boot.
 ==== Code Generation
 In order to use jOOQ type-safe queries, you need to generate Java classes from your
 database schema. You can follow the instructions in the
-http://www.jooq.org/doc/3.6/manual-single-page/#jooq-in-7-steps-step3[jOOQ user manual].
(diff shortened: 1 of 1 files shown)


philwebb:
We should probably link to the actual version that we specify in the BOM.

snicoll:
@philwebb yup and that's how I've found out about JOOq's error page :) 

There isn't a valid link that matches the version, i.e. https://www.jooq.org/doc/3.10.7/manual-single-page/#jooq-in-7-steps-step3

@lukaseder is there a way to make that happen?

lukaseder:
Something must've gone wrong when deploying things. All versions have a URL... will look into it on Monday

lukaseder:
My bad. The manual doesn't have 3-digit version numbers like the Javadoc. So the correct URL would be:

https://www.jooq.org/doc/3.10/manual-single-page/#jooq-in-7-steps-step3

snicoll:
@lukaseder that's what I understood which means we need to post-process the version of the bom to infer the two digits value.

lukaseder:
Meh, that sucks for you. I've added an .htaccess rule to ignore the third digit. 🤘

snicoll:
Sweet, thanks!

lukaseder:
> For the record, this would set a separate property with only the two first digits

Sure, I won't keep you from fixing it on your end, but if someone decides to paste the full version number into the URL, it would work for them, too, so still a useful feature for everyone else :-)

PR #13271, opened 2018-05-26
add enforce-rule to ensure maven version >= 3.5.0

hengyunabc:
Ensure maven version >= 3.5.0 to support `${revision}`.

https://maven.apache.org/docs/3.5.0/release-notes.html

When using maven 3.3.9, `${revision}` will not be replaced.


snicoll:
@hengyunabc Thank you for the PR.

PR #13280, opened 2018-05-28
Add assertion in IntegrationAutoConfigurationTests

dreis2211:
Hey,

this PR adds an assertion in IntegrationAutoConfigurationTests which seems to be missing.

Cheers,
Christoph


snicoll:
Good catch Christoph, thanks again!

PR #13286, opened 2018-05-29
Use more Tag constants

izeye:
This PR changes to use more `Tag` constants where possible.

This PR also aligns "exception" tag value in `WebFluxTags` with one in `WebMvcTags`, hence from "none" to "None".


izeye:
I also created a PR for the same change but targeted to Micrometer Spring Boot 1.5.x support: https://github.com/micrometer-metrics/micrometer/pull/629

snicoll:
Thanks again for the PR @izeye! I hesitated to transform some of those expressions to ternary operators but I left your change as is in the end.

Issue #13287, opened 2018-05-29
Use beanFactory instead of context to consist with codes


nkjackzhang:
Because `beanFactory instanceof BeanDefinitionRegistry` is used former, so `beanFactory` should be used latter.

PR #13288, opened 2018-05-29
Fix wrong reference in cast operation

nkjackzhang:
Because `beanFactory instanceof BeanDefinitionRegistry` is used former, so `beanFactory` should be used latter.

#13287


snicoll:
@nkjackzhang  thanks a lot for your first contribution to Spring Boot. This is now merged in `2.0.x` and `master`.

PR #13308, opened 2018-05-31
I see no reference to EnableReactor anywhere

jmcshane:
This appears to be a deprecated property from project reactor and should be removed from the docs.


snicoll:
@jmcshane good catch, that's the reactor 2 API.

Issue #9823, opened 2017-07-21
Add property 'spring-boot.version' to dependency managements


stepio:
Dear Spring Boot team,

Thank you very much for this project - I use it much and it's really great.
I'd like to propose a minor improvement.

Use case:
1. I'd like to use this set of modules:
```pom.xml
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-actuator</artifactId>
        </dependency>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-log4j2</artifactId>
        </dependency>
        <dependency>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-thymeleaf</artifactId>
        </dependency>
```
Although logback is nice, I found that log4j2 is a little bit better in terms of configurations' flexibility.

To prevent error with "multiple SLF4J bindings" I need to exclude "spring-boot-starter-logging" from both "spring-boot-starter-actuator" and "spring-boot-starter-thymeleaf".

The cleanest way to achieve this (correct me if I'm wrong) is to add next block to my dependency management block ("version" tag is required by maven and cannot be omitted):
```pom.xml
            <dependency>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-starter</artifactId>
                <version>${spring-boot.version}</version>
                <exclusions>
                    <exclusion>
                        <groupId>org.springframework.boot</groupId>
                        <artifactId>spring-boot-starter-logging</artifactId>
                    </exclusion>
                </exclusions>
            </dependency>
```

So if I had "spring-boot.version" variable defined in "spring-boot-dependencies" module, that would be a little bit cleaner in terms of future support. Otherwise, I have to maintain this variable myself and update both this value and parent's version (I extend "spring-boot-starter-parent") - not a big issue, obviously, but I'd prefer to avoid this.

What do you think? Does this make sense?

wilkinsona:
There's no need for the `<version>` tag if you're using Spring Boot's dependency management (either by importing the `spring-boot-dependencies` bom or by using `spring-boot-starter-parent` as your project's parent.

stepio:
Unfortunately you're wrong. I do use Boot's dependency management, but I also define my own - for the additional dependencies. And in this case I add:
```pom.xml
    <dependencyManagement>
        <dependencies>
            ...
            <dependency>
                <groupId>org.springframework.boot</groupId>
[...]

Issue #8493, opened 2017-03-04
[spring-boot-dependencies] Provide e.g. spring-boot.version for Spring Cloud


juergenzimmermann:
The bom file `spring-cloud-starter-parent` is using `spring-boot-starter-parent` as ist parent. The latest release train of _Spring Cloud_ is `Dalston.M1` which uses version `1.5.0.RC1` of `spring-boot-starter-parent`: https://github.com/spring-cloud/spring-cloud-release/blob/vDalston.M1/spring-cloud-starter-parent/pom.xml#L8
When `spring-boot-starter-parent` has a property like `spring-boot.version` and all `<dependency>` declarations are using this property instead of a hard-coded version, then it would be possible to easily upgrade _Spring Boot_ inside _Spring Cloud_.

spencergibb:
If you use spring-cloud-dependencies rather than the parent, you can upgrade the version of Spring Boot independently.

philwebb:
As mentioned by @spencergibb, you should use the BOM. I think if you generate your project from start.spring.io then this is what you'll get.

philwebb:
FWIW we've discussed using a property in the past but it causes some other issues.

wilkinsona:
See #7048

Issue #7048, opened 2016-09-28
Unable to Override Spring Boot Version Efficiently in Maven while using Spring IO Platform Pom (with possible resolution)


miller79:
Currently we have created a parent pom that we use in for our environment to manage versions of our internal dependencies and third party dependencies.  This parent pom has a parent of the Spring IO Platform parent.  The current platform pom uses 1.3.8.RELEASE of Spring Boot and there have been questions on how we can use the 1.4 version of Spring Boot while still using the parent for plugin management and dependency management.

After doing some analysis on the current structure of Spring Boot, I'm wondering if it makes sense to create a Spring-Boot Bom that would contain only Spring-Boot dependencies and have it as import scope inside of the spring-boot-dependencies pom.  This should allow it to have a similar structure as the Spring Framework Bom and allow it to be flexible enough as new versions come along.  Placing a property on this should allow users to override the version as necessary and also allow for flexibility of custom dependency management without having to worry about ordering them in a specific order as specified in the documentation:

[http://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-build-systems.html](http://docs.spring.io/spring-boot/docs/current/reference/html/using-boot-build-systems.html) 

The only issue with this is with the plugin management piece of Spring Boot as there isn't really a good way to override the versions in there while guaranteeing all the functionality required for the plugin.  I know in the 1.3 version of Spring Boot, the spring-boot.version property controlled the plugin versions and I'm not sure if the plugin management will look the same in future versions.  I know the property was removed in 1.4 due to confusion but with this structure maybe it could be reconsidered.

Let me know what you think and if there is another way that I may be missing.

wilkinsona:
Sorry, but the version of Spring Boot is very deliberately constrained. This applies both when using Boot directly and using the IO Platform.

wilkinsona:
I should add that if you want to use Spring Boot 1.4 via the IO Platform, you should upgrade to IO Platform Athens-RELEASE.

Issue #5014, opened 2016-01-23
Remove spring-boot.version property from spring-boot-dependencies


wilkinsona:
It, for good reason, doesn't control the version of any Spring Boot dependencies but people assume that it will. This is confusing people. Assuming it doesn't break the release process, we should get rid of it.

See #3992 and http://stackoverflow.com/questions/34961808/maven-and-spring-boot-dependencies/34961950#34961950

snicoll:
:+1:

klieber:
@wilkinsona I think there is a use case that was not considered before removing this property.  I have a multi-module project using spring boot.  In my parent pom I have a dependencies management section that excludes `spring-boot-starter-logging` from `spring-boot-starter` so that Log4j 2 can be used.  Prior to upgrading to `1.4.0` this is what that looked like:

```pom.xml
<parent>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-starter-parent</artifactId>
  <version>1.3.3.RELEASE</version>
</parent>
<dependencyManagement>
  <dependencies>
    <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-starter</artifactId>
        <version>${spring-boot.version}</version>
        <exclusions>
            <exclusion>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-starter-logging</artifactId>
            </exclusion>
        </exclusions>
    </dependency>
  </dependencies>
</dependencyManagement>
```

After upgrading I have to either hard code the version in the dependency management section or create my own property and hard code it there like this:

```pom.xml
<parent>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-starter-parent</artifactId>
  <version>1.4.0.RELEASE</version>
</parent>
<properties>
  <spring-boot.version>1.4.0.RELEASE</spring-boot.version>
</properties>
<dependencyManagement>
  <dependencies>
    <dependency>
        <groupId>org.springframework.boot</groupId>
        <artifactId>spring-boot-starter</artifactId>
        <version>${spring-boot.version}</version>
        <exclusions>
            <exclusion>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-starter-logging</artifactId>
            </exclusion>
        </exclusions>
    </dependency>
  </dependencies>
</dependencyManagement>
```

So now I have two places where I have to maintain the version of spring boot and run the risk of having mismatched versions.

[...]

PR #13311, opened 2018-05-31
Introduce spring-boot.version property

zbstof:
spring-boot.versioncould be used by used by child poms to resolve cases like the following:
https://stackoverflow.com/questions/50620293/maven-dependencymanagement-inherit-dependency-version-from-parent (5th approach)

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -31,6 +31,7 @@
 		</developer>
 	</developers>
 	<properties>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for opening your first contribution to Spring Boot. Unfortunately, we won't merge this as we've already considered and rejected such a change a few times. Please see the following issues:

- https://github.com/spring-projects/spring-boot/issues/5014
- https://github.com/spring-projects/spring-boot/issues/7048
- https://github.com/spring-projects/spring-boot/issues/8493
- https://github.com/spring-projects/spring-boot/issues/9823

Issue #12026, opened 2018-02-13
Remove our Narayana auto-configuration in favor of their own support


wilkinsona:
We currently have auto-configuration and starters for three JTA implementations:

- Atomikos
- Bitronix
- Narayana

Looking at Maven Central download numbers for the starters and project creation on start.spring.io, I don't think their usage justifies the ongoing maintenance cost. This is particularly true for Narayana where the maintenance cost has been higher than for Atomikos and Bitronix.

In addition to the usage being low, relatively speaking it's dropping. In February 2017, 0.76% of projects generated on start.spring.io used one of the three JTA options. In January 2018, that'd dropped to 0.41%.

There has been some discussion in the past about Narayana's auto-configuration being maintained externally. That's largely been driven by different release cadences. The request by @gytis to make [these changes](https://github.com/spring-projects/spring-boot/pull/12000) in 1.5.x is the latest example of this.

2.0 is the ideal opportunity to drop the starters and auto-configuration. I think we should take it.

snicoll:
Dropping JTA support from the core would remove quite a lot of complexity so that's a good thing but  I am afraid we're a bit  late to do this. How about acking the drop in 2.1? implementations that are willing to keep Spring Boot support can take the auto-configuration code (and starter) on their side.

philwebb:
I agree with @snicoll that it's a bit late to drop support entirely in 2.0, especially as we'd really need to find a new home for the existing code.

gytis:
I've moved the Narayana integration code to a separate project: https://github.com/snowdrop/spring-boot-narayana. I've split the starter into two as the JMS and JDBC wrappers were moved to a different package for Spring Boot 2.0. 
It is configured before the JTA configuration located in Spring Boot code base (https://github.com/snowdrop/spring-boot-narayana/blob/master/spring-boot-1-narayana-starter/src/main/java/me/snowdrop/boot/narayana/SpringBoot1NarayanaConfiguration.java#L69) so it can work without removing original transaction integrations.
If you guys agree, we could initially replace it in start.spring.io and eventually when it is convenient to you, you could remove the code too.

snicoll:
@gytis thanks for sharing. I've opened [3 issues in that repo](https://github.com/snowdrop/spring-boot-narayana/issues), could you please have a look?

gytis:
[...]

PR #13317, opened 2018-05-31
Autoconfiguration support for Hazelcast IMDG & Jet

### LINKED ISSUE #8863: Autoconfiguration support for Hazelcast IMDG & Jet
Hazelcast consists of four options - an In-Memory Data Grid client and server, and a Jet client and server.

IMDG server works ideally. If there is an `hazelcast.xml` file then a *server* `HazelcastInstance` bean is constructed.

IMDG client is being improved by https://github.com/spring-projects/spring-boot/pull/7469. Ideally there is a `hazelcast-client.xml` file then a *client* `HazelcastInstance` bean is constructed.

This could be extended to include Jet

If there is a `hazelcast.xml` file and Jet classes, then a *server* `JetInstance` bean can be constructed and a *server* `HazelcastInstance` bean taken from inside Jet.

If there is a `hazelcast-client.xml` file and Jet classes, then a *client* `JetInstance` bean can be constructed and a *client* `HazelcastInstance` bean taken from inside Jet.

Sample project https://github.com/neilstevenson/spring-boot-autoconfigure-test shows how they are currently built using Spring Boot 1.5.2.
Ideally, all the `MyConfiguration` classes can be removed in favour of autoconfiguration.

neilstevenson:
Fixes https://github.com/spring-projects/spring-boot/issues/8863

Provides auto-configuration support for Hazelcast Jet, in a manner that doesn't perturb existing auto-configuration for Hazelcast IMDG and Client.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-autoconfigure/pom.xml
@@ -82,6 +82,11 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>com.hazelcast.jet</groupId>
+			<artifactId>hazelcast-jet</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>com.h2database</groupId>
 			<artifactId>h2</artifactId>
 			<optional>true</optional>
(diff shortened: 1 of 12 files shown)


neilstevenson:
`InetAddressFormatterTests` breaks for me on 3 different Macs (10.13 and 10.11), will need to fix that first

snicoll:
What is the full Stacktrace for the error?

philwebb:
@neilstevenson Thanks very much for the PR.

Looking at the changes required to support Hazelcast Jet, my current feeling is that it would be better to create a separate project for the auto-configuration. The main reason is that since Hazelcast Jet is still relatively new, it will be easier to evolve the code if it has a separate release schedule.

We're trying to the same kind of thing with a few of our other auto-configurations (for example Narayana [#12026]). Our aim is allow others to develop their integration in the best way they seem fit, and at the same time reduce the amount of code that we need to maintenance.

Thanks again for the work, and please let us know if we can help review the new module if you decide to proceed in that direction.

PR #13324, opened 2018-06-01
Make MetricsProperties fields final if possible

izeye:
This PR makes `MetricsProperties` fields `final` if possible.


snicoll:
Thanks again @izeye!

PR #13345, opened 2018-06-03
Remove usage of deprecated Matchers class

dreis2211:
Hi,

just noticed this deprecation warning. I guess it's safe to remove the call to `same()` alltogether and hence the need for the alternative `ArgumentMatchers` import as well. Default behaviour should be checking the parameters with `equals()` which is sufficient here in my opinion.

Let me know what you think.
Cheers,
Christoph


snicoll:
Good catch @dreis2211, thanks again!

Issue #12846, opened 2018-04-12
Upgrade spring boot 2.0.x to Tomcat 8.5.30


mrcasablr:
Tomcat 8.5.30 is now available - http://mail-archives.apache.org/mod_mbox/www-announce/201804.mbox/%3Cd80fbc83-a1a5-953f-3a0d-32b712705700%40apache.org%3E

snicoll:
@mrcasablr thanks but there is no need to raise dependencies upgrade issue, we have a semi-automated process that takes care of this.

PR #13351, opened 2018-06-04
Add "Managed dependency upgrade" section

izeye:
This PR adds "Managed dependency upgrade" section to the issue template as I've seen issues for a minor version of a dependency upgrade constantly like https://github.com/spring-projects/spring-boot/issues/12846. This section hopefully reduces this kind of issues to be raised.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -18,4 +18,7 @@ STOP!! Please don't raise security vulnerabilities here. Head over to https://pi
 🎁 Enhancement
 Please start by describing the problem that you are trying to solve. There may already
 be a solution, or there may be a way to solve it that you hadn't considered.
+
+🙅 Managed dependency upgrade
+You DO NOT need to raise an issue for a managed dependency version upgrade as there's a semi-automatic process for checking managed dependencies for new versions before a release. But pull requests for upgrades that are more involved than just a version property change are still most welcome.
 -->


.github/ISSUE_TEMPLATE.md
```diff
@@ -18,4 +18,7 @@ STOP!! Please don't raise security vulnerabilities here. Head over to https://pi
 🎁 Enhancement
 Please start by describing the problem that you are trying to solve. There may already
 be a solution, or there may be a way to solve it that you hadn't considered.
+
+- Managed dependency upgrade
```
snicoll: Where is the emoji? 🤣

.github/ISSUE_TEMPLATE.md
```diff
@@ -18,4 +18,7 @@ STOP!! Please don't raise security vulnerabilities here. Head over to https://pi
...
+You DO NOT need to raise a minor version upgrade issue for a managed dependency as there's a semi-automatic process for checking managed dependencies for new minor versions before a release.
```
snicoll: I wonder if insisting on "minor" is the right call here. bomr is also able to let us know about major upgrades, right @wilkinsona ?

.github/ISSUE_TEMPLATE.md
izeye: @snicoll That was the hardest part, so I intentionally avoided choosing one by using "-" 😅 

Maybe 🙅?

.github/ISSUE_TEMPLATE.md
wilkinsona: Right. If a change just modifies the version property in `spring-boot-dependencies` it's quicker and easier for Bomr to take care of it, irrespective of it's a new major, minor, or patch release. Pull requests for upgrades that are more involved (such as [this one](https://github.com/spring-projects/spring-boot/pull/13322)) are still most welcome, irrespective of whether it's a new major, minor or, hopefully not, patch release.

izeye: Looking at @wilkinsona 's comment I realized I used the wrong term. "minor" meant to be "patch" as I thought it could handle only "patch" versions.

Based on the @wilkinsona 's comment I'd like to change the section to:

> You DO NOT need to raise an issue for a managed dependency version upgrade as there's a semi-automatic process for checking managed dependencies for new versions before a release. But pull requests for upgrades that are more involved than just a version property change are still most welcome.

WDYT?

snicoll: That sounds good to me Johnny, thank you!

izeye: @snicoll Thanks for the quick feedback! I updated as suggested.

PR #13355, opened 2018-06-04
Add duration support for setConnectTimeout and setReadTimeout

nosan:
this PR provides new methods `setConnectTimeout(Duration)` and `setReadTimeout(Duration)` for `RestTemplateBuilder` .

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
@@ -19,6 +19,7 @@ package org.springframework.boot.web.client;
 import java.lang.reflect.Constructor;
 import java.lang.reflect.Field;
 import java.lang.reflect.Method;
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
@@ -56,6 +57,7 @@ import org.springframework.web.util.UriTemplateHandler;
  * @author Phillip Webb
  * @author Andy Wilkinson
  * @author Brian Clozel
+ * @author Dmytro Nosan
  * @since 1.4.0
  */
 public class RestTemplateBuilder {
@@ -63,6 +65,10 @@ public class RestTemplateBuilder {
 	private final boolean detectRequestFactory;
 
 	private final String rootUri;
+
+	private final Duration readTimeout;
+
+	private final Duration connectTimeout;
 
 	private final Set<HttpMessageConverter<?>> messageConverters;
 
@@ -76,8 +82,6 @@ public class RestTemplateBuilder {
 
 	private final Set<RestTemplateCustomizer> restTemplateCustomizers;
 
-	private final Set<RequestFactoryCustomizer> requestFactoryCustomizers;
-
 	private final Set<ClientHttpRequestInterceptor> interceptors;
 
 	/**
@@ -88,6 +92,8 @@ public class RestTemplateBuilder {
 	public RestTemplateBuilder(RestTemplateCustomizer... customizers) {
 		Assert.notNull(customizers, "Customizers must not be null");
 		this.detectRequestFactory = true;
+		this.readTimeout = null;
+		this.connectTimeout = null;
 		this.rootUri = null;
 		this.messageConverters = null;
 		this.requestFactorySupplier = null;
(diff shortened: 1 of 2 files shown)


nosan:
```
	if (!CollectionUtils.isEmpty(this.restTemplateCustomizers)) {
			for (RestTemplateCustomizer customizer : this.restTemplateCustomizers) {
				customizer.customize(restTemplate);
			}
		}
		restTemplate.getInterceptors().addAll(this.interceptors);
```

Maybe `Customizers` should be moved in the end of the method in `2.1.0`?

snicoll:
@nosan please do not use this issue to discuss something totally unrelated. We're on gitter if you want to chat. Or you can create another issue.

nosan:
@snicoll Got it, sorry about that.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
```diff
@@ -64,6 +66,10 @@
 
 	private final String rootUri;
 
+	private final Duration readTimeout;
```
snicoll: That's not what I have in mind, please keep your change to a minimum. I expect two new methods passing a `Duration` to the existing customizers with the deprecated methods delegating to them.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
```diff
@@ -444,42 +447,61 @@ public RestTemplateBuilder additionalCustomizers(
...
+				this.basicAuthorization,
+				append(this.restTemplateCustomizers, customizers), this.interceptors);
 	}
 
 	/**
 	 * Sets the connect timeout in milliseconds on the underlying
 	 * {@link ClientHttpRequestFactory}.
 	 * @param connectTimeout the connect timeout in milliseconds
 	 * @return a new builder instance.
+	 * @deprecated Deprecated in favor of {@link #setConnectTimeout(Duration)}
```
snicoll: The format we use is `@deprecated since 2.1.0 in favour of {@link #setConnectTimeout(Duration)}`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
```diff
@@ -444,42 +447,61 @@ public RestTemplateBuilder additionalCustomizers(
...
 	 */
+	@Deprecated
 	public RestTemplateBuilder setConnectTimeout(int connectTimeout) {
+		return setConnectTimeout(Duration.ofMillis(connectTimeout));
+
+	}
+
+	/**
+	 * Sets the connection timeout on the underlying {@link ClientHttpRequestFactory}.
```
snicoll: I'd put the method taking `Duration` above the deprecated one, not below.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/HttpWebServiceMessageSenderBuilder.java
```diff
@@ -35,29 +36,28 @@
  */
 public class HttpWebServiceMessageSenderBuilder {
 
-	private Integer connectionTimeout;
+	private Duration connectTimeout;
```
snicoll: Good catch. I've polished the wrong name in https://github.com/spring-projects/spring-boot/commit/93396aef51ac32cdeb1545f2b7ede923fd9f0efb please rebase and push force an update.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/client/RestTemplateBuilderTests.java
```diff
@@ -497,6 +499,47 @@ public void unwrappingDoesNotAffectRequestFactoryThatIsSetOnTheBuiltTemplate() {
...
+		ClientHttpRequestFactory requestFactory = this.builder
+				.requestFactory(SimpleClientHttpRequestFactory.class)
+				.setConnectTimeout(1234).build().getRequestFactory();
+		assertThat(ReflectionTestUtils.getField(requestFactory, "connectTimeout"))
+				.isEqualTo(1234);
+	}
+
+	@Test
+	@SuppressWarnings("deprecation")
+	public void nullReadTimeoutShouldNotAppliedIgnorePreviousSettings() {
```
snicoll: Those two tests are new right? You could do the same with one that takes a `Duration` and one that takes `null` afteR. No need for the deprecation. Besides, I'd appreciate a separate PR for this as they are unrelated.

nosan:
@snicoll  Thank you.
Adding only `setReadTimeout(Duration)` and `setConnectTimeout(Duration)` without changing `RestTemplateBuilder` is not a correct way.
Currently, `RestTemplateBuilder` uses `Set<RestTemplateCustomizer> restTemplateCustomizers>` for configuring these timeouts, so every time when a user is trying to set a new timeout, `RestTemplateBuilder` will add a new `Customizer` in the `Set<RestTemplateCustomizer> restTemplateCustomizers>` .
It works perfectly for the current implementation because `RestTemplateBuilder` has only `setReadTimeout(int)` and `setConnectTimeout(int)` and a user is not able to set `null` timeout.
With the new methods user is able to do this and in that case, it means a user doesn't want to set a timeout on the underlying, thus I've provided new fields for reach this purpose.




**P.S.** With the current implementation this test doesn't work.
```
	ClientHttpRequestFactory requestFactory = this.builder
				.requestFactory(SimpleClientHttpRequestFactory.class)
				.setConnectTimeout(Duration.ofSeconds(5))
				.setConnectTimeout(null).build()
				.getRequestFactory();
		assertThat(ReflectionTestUtils.getField(requestFactory, "connectTimeout"))
				.isEqualTo(-1);
```

nosan:
@snicoll  I've updated PR.
I will create a separate PR for `HttpWebServiceMessageSenderBuilder`.

snicoll:
@nosan thanks for the PR. I found the changes in `RestTemplateBuilder` too invasive so I've pre-polished your PR, see 8691d01.

Issue #6701, opened 2016-08-20
Make it easier to add ClientHttpRequestInterceptors using RestTemplateBuilder


philwebb:
To add an interceptor you currently need to use the Customizer. We could probably make that easier.

PR #13358, opened 2018-06-04
RestTemplateCustomizers should be applied at the end of the `build` method

nosan:
Expected behavior is `Customizers` will be applied at the end of the `RestTemplateBuilder#build`, but currently  `Interceptors` at the end of the phase. 
This PR fixes this.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
@@ -531,12 +531,13 @@ public class RestTemplateBuilder {
 		if (this.basicAuthorization != null) {
 			restTemplate.getInterceptors().add(this.basicAuthorization);
 		}
+		restTemplate.getInterceptors().addAll(this.interceptors);
+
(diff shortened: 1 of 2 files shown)


snicoll:
This was added in #6701 by @wilkinsona - I am wondering if he had something in mind when he added the interceptors _after_ the customizers had ran.

wilkinsona:
If there was a reason for it, I can't remember what it was. Sorry. If it were important, it's the sort of thing that I would typically mention in the commit message or capture in a test. I suspect it was just an oversight on my part and the customisers should have stayed in last place.

nosan:
Apparently, it could be `RestTemplate.getRequestFactory()`. 'RestTemplate' returns `InterceptingClientHttpRequestFactory` instead of the real `Factory` if it has `Interceptors`. However `Interceptors` at the end doesn't help every time, because previously `BasicAuthInterceptor` might be added.

bclozel:
For consistency I think we should merge this. Spring Boot 2.1 is the right fix version since this would change the ordering of the interceptors or even the list itself for users calling `getInterceptors().set()` or `getInterceptors().replace*()` (even if it's not really advised).

PR #13364, opened 2018-06-05
`HttpWebServiceMessageSenderBuilder` use `Duration` instead of `int`.

nosan:


snicoll:
Thank you for the PR @nosan. This is now merged in `master`.

PR #13373, opened 2018-06-05
Metrics for `WebServiceTemplate`.

nosan:
This PR includes `WebServiceTemplate` metrics.  
PR contains:

- `MetricsClientInterceptor`  - counts metrics. 
- `MetricsWebServiceTemplateCustomizer`  -  Customizer for registration `MetricsClientInterceptor`
- `WebServiceTemplateMetricsAutoConfiguration` - Defines  `MetricsWebServiceTemplateCustomizer`  and `DefaultWebServiceTemplateExchangeTagsProvider` beans.
- `DefaultWebServiceTemplateExchangeTagsProvider`  - generates following `tags`: 
    -  `uri`,  the request's URI
    -  `status`, the response's code (for example, `OK`, `IO_ERROR`, `CLIENT_ERROR`, `SERVER_ERROR`).
    -  `action`, the `SOAP` action.


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/MetricsClientInterceptor.java
```diff
@@ -0,0 +1,82 @@
...
+			WebServiceTemplateExchangeTagsProvider tagProvider, String metricName) {
+		this.tagProvider = tagProvider;
+		this.meterRegistry = meterRegistry;
+		this.metricName = metricName;
+	}
+
+	@Override
+	public boolean handleRequest(MessageContext messageContext)
+			throws WebServiceClientException {
+		TIMERS.set(System.nanoTime());
```
jkschneider: Should use `Timer.start(meterRegistry)` so that you are using the registry's clock for this sample.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
```diff
@@ -0,0 +1,131 @@
...
+			return IO_ERROR;
+		}
+		if (exception instanceof WebServiceClientException) {
+			return CLIENT_ERROR;
+		}
+
+		return SERVER_ERROR;
+	}
+
+	private static String getUri() {
```
jkschneider: Does this return a parameterized uri. For example `/foo/{id}` instead of `/foo/1`?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
```diff
@@ -0,0 +1,131 @@
...
+		String uri = getUri();
+		return (!StringUtils.hasText(uri) ? NO_URI : Tag.of("uri", uri));
+	}
+
+	/**
+	 * Creates a {@code Tag(action)} using {@link SoapMessage#getSoapAction()}.
+	 * @param messageContext Context of the SOAP message.
+	 * @return the action tag
+	 */
+	public static Tag action(MessageContext messageContext) {
```
jkschneider: From the spec, it appears that action could potentially be a high cardinality tag:

image

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/MetricsClientInterceptor.java
jkschneider: If `MessageContext` is the same instance in both `handleRequest` and `afterCompletion`, we may be able to employ a `Map` rather than `ThreadLocal`. Not sure if there is any advantage to this.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
nosan: `WebServiceTemplate` doesn't have `uriTemplates`.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/MetricsClientInterceptor.java
nosan: Thank you. Good catch, I will change this.

nosan:
@jkschneider Thank you a lot 👍

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
nosan: I'm not sure about `SOAPAction: "myapp.sdl" -> myapp.sdl`
@jkschneider Should we do something with dots?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
jkschneider: What can we do then to limit tag cardinality?

snicoll:
Thanks for the PR @nosan 

We discussed this one on the call and we would like to see an official metrics support in Spring WS before moving forward with this issue (i.e moving the interceptor). @gregturn would you be interested to review and integrate that feature in Spring WS?

jkschneider:
I wholeheartedly agree @snicoll. It will be great to have the meat of this support in Spring WS so folks can still use these metrics outside of a Spring Boot context (they'd just have to configure them manually).

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
jkschneider: Don't concern yourself with dots. If they aren't an acceptable character for a particular monitoring system, it is the responsibility of Micrometer's `NamingConvention` for that registry to strip or escape them.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
nosan: Should we do?  With `RestTemplate` it could be an issue with `uriTemplates`, here we don't have it.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
nosan: that was my question, ty :)

gregturn:
On the master branch on Spring WS, I would certainly support efforts to add metrics. I'm sure the community would appreciate this.

To track the conversation, I've opened https://jira.spring.io/browse/SWS-1024.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/webservices/client/WebServiceTemplateExchangeTags.java
jkschneider: This speaks to my ignorance of SOAP I think. Are you saying SOAP URIs are always fixed and never have path variables in them?

nosan: There are 2 often cases.
 -  Fixed `URI` -  different `SoapAction`
- Different `URI` , Fixed `SoapAction` (e.g. "")

for the `RestTemplate` it could be a case when people use `URI + variables` instead of `uriVariables` map, thus instead of one `tag {id}` will be N `tags`.

**P.S.** We can easily add `MeterFilter` just to make sure everything is ok, as `Spring Boot` did for `RestTemplate`.

snicoll:
@nosan there hasn't been much progress on the WS front of things and this PR will have to be heavily restructured to build against the support in WS when it is available. Rather than having a PR here with no way for us to handle it, I suggest that the change in WS is the trigger for us to reconsider this support. Thanks for the PR in any case.

Issue #3039, opened 2015-05-25
MessageSourceAutoConfiguration: consider replacing ResourceBundleMessageSource by ReloadableResourceBundleMessageSource


php-coder:
Why `MessageSourceAutoConfiguration` creates `ResourceBundleMessageSource` instead of `ReloadableResourceBundleMessageSource`? Javadoc of `ResourceBundleMessageSource` says: "Consider ReloadableResourceBundleMessageSource for an alternative that is capable of refreshing the underlying bundle files."

So, why don't use `ReloadableResourceBundleMessageSource`?

dsyer:
`ReloadableResourceBundleMessageSource` has some cost at runtime, so I assume it isn't used by default because you wouldn't necessarily want it in production.

philwebb:
The recommended approach since 1.3 is to use Devtools.

PR #13377, opened 2018-06-05
Add support for ReloadableResourceBundleMessageSource

ruifigueira:
This PR adds support for `ReloadableResourceBundleMessageSource`, by adding a `spring.messages.reloadable` configuration property, which can be set to `true` to force the `messageSource` bean to be an instance of that class.

I'm currently working on a legacy Spring project that is progressively being migrated into Spring Boot, and we use a `ReloadableResourceBundleMessageSource` there, and we would like to replace it with the `MessageSourceAutoConfiguration`.

I'm well aware of #3039, and that:

> The recommended approach since 1.3 is to use Devtools.

However, Devtools is not an option in this project, and even if it was, reloading all the application takes much more time than the `ReloadableResourceBundleMessageSource` takes to refresh.

I think that changes to enable `ReloadableResourceBundleMessageSource` support are minimal, and the new introduced configuration, `spring.messages.reloadable` is set to `false` by default, so it's backaward compatible.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/MessageSourceProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/MessageSourceProperties.java
@@ -22,6 +22,8 @@ import java.time.Duration;
 import java.time.temporal.ChronoUnit;
 
 import org.springframework.boot.convert.DurationUnit;
+import org.springframework.context.support.ReloadableResourceBundleMessageSource;
+import org.springframework.context.support.ResourceBundleMessageSource;
 
 /**
  * Configuration properties for Message Source.
@@ -71,6 +73,12 @@ public class MessageSourceProperties {
 	 */
 	private boolean useCodeAsDefaultMessage = false;
 
+	/**
+	 * Whether to use a {@link ReloadableResourceBundleMessageSource} instead of the
+	 * default {@link ResourceBundleMessageSource}. Recommended during development only.
+	 */
+	private boolean reloadable = false;
+
 	public String getBasename() {
 		return this.basename;
 	}
@@ -117,6 +125,14 @@ public class MessageSourceProperties {
 
 	public void setUseCodeAsDefaultMessage(boolean useCodeAsDefaultMessage) {
 		this.useCodeAsDefaultMessage = useCodeAsDefaultMessage;
+	}
+
+	public boolean isReloadable() {
+		return this.reloadable;
+	}
+
+	public void setReloadable(boolean reloadable) {
+		this.reloadable = reloadable;
 	}
 
 }
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/context/MessageSourceAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/context/MessageSourceAutoConfigurationTests.java
@@ -21,6 +21,7 @@ import java.util.Locale;
 import org.junit.Ignore;
 import org.junit.Test;
 
+import org.springframework.beans.BeansException;
 import org.springframework.beans.DirectFieldAccessor;
 import org.springframework.boot.autoconfigure.AutoConfigurations;
 import org.springframework.boot.test.context.assertj.AssertableApplicationContext;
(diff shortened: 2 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/MessageSourceProperties.java
```diff
@@ -71,6 +73,12 @@
 	 */
 	private boolean useCodeAsDefaultMessage = false;
 
+	/**
+	 * Whether to use a {@link ReloadableResourceBundleMessageSource} instead of the
```
snicoll: Configuration key descriptions can't have javadoc tag (see documentation). I think we need to improve the description as high-level descriptions shouldn't contain implementation details (rather it should describe what the feature does).

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/context/MessageSourceAutoConfigurationTests.java
```diff
@@ -220,6 +224,45 @@ public void existingMessageSourceInParentIsIgnored() {
 								.isEqualTo("bar")));
 	}
 
+	@Test
+	public void testDefaultReloadableValueMessageSource() {
+		this.contextRunner.withPropertyValues("spring.messages.basename:test/messages")
```
snicoll: This assertion can be added to the existing basic test

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/context/MessageSourceAutoConfigurationTests.java
```diff
@@ -220,6 +224,45 @@ public void existingMessageSourceInParentIsIgnored() {
...
+				});
+	}
+
+	@Test
+	public void testReloadableMessageSource() {
+		this.contextRunner.withPropertyValues("spring.messages.basename:test/messages",
+				"spring.messages.reloadable:true").run((context) -> {
+					assertThat(getDeclaredMessageSource(context))
+							.isInstanceOf(ReloadableResourceBundleMessageSource.class);
+					assertThat(context.getMessage("foo", null, "Foo message", Locale.UK))
```
snicoll: What is the purpose of this assertion?

philwebb:
We discussed this a little and we're not too sure on the best way to continue. It would certainly be nice to offer an easy way to plug-in a custom `ResourceBundleMessageSource` but we're not sure if we should do that with a property.

There is some precedent for controlling caching with properties. For example with `spring.thymeleaf.cache` which does a similar thing for Thymeleaf templates. We need a little more time to consider the options. If we do add property support we should also change DevTools to make use of it so a full restart isn't required.

ruifigueira:
Thanks for the input. I understand that having a configuration property that "decides" which implementation to use is not ideal at all, and I wouldn't mind if I had to explicitly declare a bean for that:

```java
@Bean
public MessageSource messageSource() {
  return ReloadableResourceBundleMessageSource();
}
```

But then, I would have to configure it myself (even `MessageSourceProperties` bean is not available once a `MessageSource` bean is explicitly declared).

One possible idea would be to configure that declared `MessageSource` bean in the `MessageSourceAutoConfiguration` using the `MessageSourceProperties`.

candrews:
Should dev tools set `spring.messages.reloadable` to true when it's active?

snicoll:
@candrews probably, yes. Can you please create a separate issue?

wilkinsona:
With some regret, we've decided to revert this change. With hindsight, we don't think that the approach taken here is quite right. In particular, setting `reloadable: true` breaks the default approach of placing messages files in `src/main/resources` which isn't ideal. We're not yet sure what the right approach should be, or if it should have a boolean reloadable property. Rather than backing ourselves into a corner by shipping this in 2.1, we'd like to give ourselves the time and freedom to consider all alternatives without having to fit the configuration model that was proposed here.

wilkinsona:
I've opened https://github.com/spring-projects/spring-boot/issues/15007.

ruifigueira:
I understand, I honestly didn't expect this to be merged, because we were still discussing if my approach was the best one. I was already convinced it wasn't, and your point regarding files placement is a very good point against this approach.

php-coder:
> With some regret, we've decided to revert this change.

BTW it seems like it wasn't removed from the changelogs:
- https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1.0-RC1-Release-Notes#reloadable-messagesource
- https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1.0-RC1-Configuration-Changelog

snicoll:
@php-coder this change was reverted after RC1 was released so the current changelog is accurate.

PR #13378, opened 2018-06-05
Replace commons-codec Base64 usage

dreis2211:
Hi,

just noticed two usages of `org.apache.commons.codec.binary.Base64` in tests that could be removed in my opinion. I went for using the utility class from Spring-Framework (mainly because it was already imported as well). If you prefer using the `java.util.Base64` directly, I don't mind adjusting the PR. Just let me know.

Cheers,
Christoph

P.S.: It seems that Boot itself doesn't use any commons-codec code directly after this PR is merged.  I wonder if we could remove the dependency alltogether in a separate PR?


snicoll:
> P.S.: It seems that Boot itself doesn't use any commons-codec code directly after this PR is merged. I wonder if we could remove the dependency alltogether in a separate PR?

Thanks @dreis2211 - There are several dependencies that uses commons codec internally so I think it would be nice to keep dependency management for it. Having said that, I've updated our checkstyle rules to make sure we don't add imports on this package again (see 68380b68effc)

dreis2211:
Fair enough. Thanks for the heads up @snicoll

PR #13388, opened 2018-06-06
Improved the error message, when no main class was found using the gradle-plugin.  #12712

### LINKED ISSUE #12712: Gradle build error is unclear when no main class is specified or the main class could not be resolved
This is, admittedly, a very low-priority issue.

If, for some reason (I just answered a StackOverflow question about that problem), there is no main class configured, and the Spring Boot gradle plugin can't resolve any main class, the bootJar task fails with the following error:

> The value of a manifest attribute must not be null (Key=Start-Class)

This error message could be clearer, and clearly state that the specified main class is null, or that no main class could be resolved.

maximAtanasov:
Fixes #12712 - When building with the spring-boot-gradle-plugin the error mesage is now clear.

Previously the error message would say: `The value of a manifest attribute must not be null (Key=Start-Class)` .

It now says: `No main class could be resolved`

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootArchiveSupport.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootArchiveSupport.java
@@ -35,6 +35,7 @@ import org.gradle.api.specs.Specs;
 import org.gradle.api.tasks.WorkResult;
 import org.gradle.api.tasks.bundling.Jar;
 import org.gradle.api.tasks.util.PatternSet;
+import org.gradle.api.java.archives.ManifestException;
 
 /**
  * Support class for implementations of {@link BootArchive}.
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootArchiveSupport.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootArchiveSupport.java
@@ -76,7 +76,12 @@ class BootArchiveSupport {
 	void configureManifest(Jar jar, String mainClassName) {
 		Attributes attributes = jar.getManifest().getAttributes();
 		attributes.putIfAbsent("Main-Class", this.loaderMainClass);
-		attributes.putIfAbsent("Start-Class", mainClassName);
+		if (mainClassName == null && attributes.get("Start-Class") == null) {
+			throw new ManifestException("No main class could be resolved");
+		}
+		else {
+			attributes.putIfAbsent("Start-Class", mainClassName);
+		}
 	}
 
 	CopyAction createCopyAction(Jar jar) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@maximAtanasov Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13388)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13388) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@maximAtanasov Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13388)!

wilkinsona:
Thanks very much for opening your first Spring Boot pull request, @maximAtanasov.

spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootArchiveSupport.java
```diff
@@ -76,7 +76,12 @@
 	void configureManifest(Jar jar, String mainClassName) {
 		Attributes attributes = jar.getManifest().getAttributes();
 		attributes.putIfAbsent("Main-Class", this.loaderMainClass);
-		attributes.putIfAbsent("Start-Class", mainClassName);
+		if (mainClassName == null && attributes.get("Start-Class") == null) {
+			throw new ManifestException("No main class could be resolved");
```
wilkinsona: I'm not sure that this is the right message for the exception. `BootArchiveSupport` doesn't know anything about the main class being resolved. In fact, neither `BootJar` nor `BootWar` do either. It's only `JavaPluginAction` and `WarPluginAction` that set up `MainClassConvention` for the `mainClassName` property of the default `BootJar` and `BootWar` archives.

It would be interesting to know what the effect would be of changing `MainClassConvention` so that it throws an exception rather than returning `null` when it can't resolve the main class. It would also be interesting to know what the effect would be of removing `@Optional` from `BootArchive.getMainClassName()`.

In summary, I'm not 100% sure of the best way to improve the error message and some experimentation is needed. Would you like to do that and comment with your findings?

maximAtanasov: Yes, I will look into it and try to fix it the way you suggested.

maximAtanasov: @wilkinsona 
Removing `@Optional` causes the bootJarManifestMainClass test to fail with the following exception:

```org.springframework.boot.gradle.docs.PackagingDocumentationTests > bootJarManifestMainClass FAILED
    java.lang.RuntimeException at PackagingDocumentationTests.java:68
        Caused by: org.gradle.testkit.runner.UnexpectedBuildFailure at PackagingDocumentationTests.java:68
```
So it seems it is not finding the main class.

Also, having `MainClassConvention` throw an exception if `resolveMainClass()` fails makes the test fail with the same exception as above. Although, running `gradle build` on a test project without a main method will still produce the correct error message. I will try to get around that and push a solution.

maximAtanasov: Unfortunately, the `Start-Class` property in that particular is added later in the creation of the jar and there is no way  to check if it exists inside `MainClassConvention`, so throwing an exception there causes the jar not to be built, when it should. The only solution I could find that passes all tests and does what it should is the one I suggested, but it is true that `BootArchiveSupport` is not aware of the main class.

wilkinsona:
Thanks for the PR, but I've taken a slightly different approach to fixing this. If you are interested, please take a look at https://github.com/spring-projects/spring-boot/commit/6f2171c28222bcd7f208e0f72ab7bfeb86e795f3.

PR #13409, opened 2018-06-07
use default `timeouts` for detected `HttpWebServiceMessageSender`.

nosan:
Currently, `WebServiceTemplateBuilder` doesn't use any default timeouts for detected `HttpWebServiceMessageSender`, however  `HttpUrlConnectionMessageSender` and `HttpComponentsMessageSender` has default timeouts `60 seconds`. 

This PR sets default timeouts for the detected sender as well.


snicoll:
We don't configure any of that and you are free to configure them if you want to use them. The current default aligns to what `RestTemplateBuilder` does and is more consistent that way. I'd also argue that if you want to set default, these should be set in the builder rather than the component calling it.

I'd like to point out one more time that PR has unrelated changes that makes it quite hard to read, please refrain from doing polishing changes.

Issue #12576, opened 2018-03-21
Consider removing lombok from dependency management


philwebb:
We've seen issues (such as #12568) and I'm not sure we really recommend it. It's also confusing when picked on start.spring.io in combination with Kotlin.

olivergierke:
To be fair, #12568 is an issue in Jackson that apparently has all sorts of weird issues (see [here](https://github.com/FasterXML/jackson-databind/issues/1239) and [here](https://github.com/FasterXML/jackson-databind/issues/1810)) when it comes to reliably determine constructor properties. The change in Lombok shouldn't actually have caused that issues as the code generated stays the same, it's just not containing duplicate information on Java 8.

wilkinsona:
Yes.

olivergierke:
Thanks, Andy!

PR #13413, opened 2018-06-07
Update Lombok to 1.18.0

larsgrefer:
Update Lombok from 1.16.22 to 1.18.0

See also: https://twitter.com/project_lombok/status/1003778962911358976

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -120,7 +120,7 @@
 		<liquibase.version>3.5.5</liquibase.version>
 		<log4j2.version>2.10.0</log4j2.version>
 		<logback.version>1.2.3</logback.version>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR but we can't upgrade from 1.16 to 1.18 in 2.0.x. We will upgrade in 2.1 but we have a semi-automated process that will handle it for us.

rajadilipkolli:
@wilkinsona Java10 needs this version of Lombok so if we are maintaining documentation about using spring-boot 2.0.x with java10 then we should add a note about it to override version.

snicoll:
Unfortunately, that's against our upgrade policy for maintenance releases. That's one more reason for [removing lombok from managed dependencies](https://github.com/spring-projects/spring-boot/issues/12576). 

I've added [a note on the wiki](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-with-Java-9-and-above#lombok)

PR #13416, opened 2018-06-07
Delete Tomcat temp directories on exit

candrews:
Running a number of times results in a bunch of directories being created in the temp directory. Over time, that adds up to decent amount of wasted disk space. It would be nice to try to clean up those temp files when done we're done with them.

If any temp directories are created, use java.io.File.deleteOnExit() to ask the JVM to delete them when it exits.

Deletion will not occur if JVM termination is abnormal, but at least now they'll be automatically cleaned up under normal circumstances.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatServletWebServerFactory.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatServletWebServerFactory.java
@@ -70,6 +70,7 @@ import org.springframework.context.ResourceLoaderAware;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
+import org.springframework.util.FileSystemUtils;
 import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR. I’m not sure I understand the rationale here. To be causing a “decent amount of wasted disk space” the directories must, presumably, have a decent amount of content. However, the code as proposed will, as I understand it, only delete the directory if it is empty. Can you please clarify?

candrews:
Ugh, the javadoc isn't clear; I thought deleteOnExit would delete a directory even if it wasn't empty. Which is not true.

I've updated this PR to use a shutdown hook instead.

wilkinsona:
Thanks for the update, but I'm not too keen on using a shutdown hook for this. The cost of the extra thread feels too high given the problem that it solves, particularly given that we already have a shutdown hook for closing the application context on JVM exit.

If we really want to do this (I'm still not sure that we do), it feels like something that should be handled when the web server is stopped. I'm not exactly sure how to do that yet as it'll require passing some information from the factory to the web server that it creates.

candrews:
According to the javadoc https://docs.oracle.com/javase/8/docs/api/java/lang/Runtime.html#addShutdownHook-java.lang.Thread- the thread is not actually running so it wouldn't be costly. It sounds like it's treated more like a Runnable than a Thread (although I haven't read the JDK source code to know exactly how it works).

philwebb:
We think we'll need a different approach, probably one that hooks into the `WebServer` close method. @candrews I'll close this one but could you please open a new issue do it doesn't get forgotten.

PR #13421, opened 2018-06-08
Add binding test with a map key prefixed with "/"

### LINKED ISSUE #13404: Problem with yml configuration file parsing map
Spring boot version： 2.0.2 (1.5.13 is normal)

Problem Description：

If the map key is with “/”, it cannot be parsed normally. Others are normal.

Example：

```
application.yml 

shiro:
  testMap:
    /test1: test1
    /test2: test2
```

Configuration class 

```java
@Data
@ConfigurationProperties(prefix="shiro")
public class ShiroProperties {
	private Map<String,String> testMap;
}
```

Test class

```java
@Transactional
@RunWith(SpringRunner.class)
@SpringBootTest(classes=DataManage.class)
public class ShiroTest extends BaseTransactionalTest{

	@Autowired
	private ShiroProperties shiroConfig;
	
	@Test
	public void read() {
		System.out.println(shiroConfig.toString());
	}
	
}
```

console info：

```
ShiroProperties(testMap={test1=test1, test2=test2})
```

kflau:
This PR adds a test case as a user guide showing how to configure keys like URL.
The proper way to configure URL alike keys in yaml should look like:

```yaml
application.yml 

shiro:
  testMap:
    "[/test1]": test1
    "[/test2]": test2
    "[/test3/test31]": test31
```

Fixes #13404

### DIFF
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java
@@ -323,12 +323,11 @@ public class ConfigurationPropertiesTests {
 
 	@Test
 	public void loadWhenBindingToMapKeyWithForwardSlashShouldBind() {
-		load(MapProperties.class, "mymap[/test1]:test1", "mymap[/test2]:test2", "mymap[/test3/test31]:test31");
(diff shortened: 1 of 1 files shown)


snicoll:
@kflau thanks for the PR but I don't think that's actually fixing 13404. We've decided to tackle this issue and try to implement something that would be more user-friendly.

Issue #5875, opened 2016-05-05
MetricsFilter may create an unbounded number of metrics for requests with a templated URI that are not handled by Spring MVC


heroicefforts:
The MetricsFilter is unsuitable for REST microservices that are backing a large number of entities.  In our case, it was trying to generate gauges and counters for /widget/1 to /widget/60000000.  It eventually soaks the heap and causes the server to slow to a crawl as it spends most of its CPU cycles in GC.  There should be a big red "caveat" bubble in the documentation that details this scenario.  

Ideally, a Spring MVC controller metrics controller should be implemented that works similar to the DropWizard Jersey 2 implementation (plus some RequestMapping introspection), then the MetricsFilter could be disabled by default.

wilkinsona:
The behaviour you've described isn't intended, and it doesn't match what's tested.

In [one of the tests](https://github.com/spring-projects/spring-boot/blob/v1.3.3.RELEASE/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/autoconfigure/MetricFilterAutoConfigurationTests.java#L108-L121) there's a request mapping like this: `@RequestMapping("templateVarTest/{someVariable}")`. It's called with `/templateVarTest/foo` which results in `status.200.templateVarTest.someVariable` and `response.templateVarTest.someVariable` being updated. As you can see, it's the name of the variable in the request mapping rather than the value in the request URI that's used.

Can you please provide a sample or testcase that illustrates the behaviour you have described?

heroicefforts:
Then, perhaps, it is a consequence of using Jersey with SpringBoot (don't get me started on why...):

```
@GET
@Path("/parent/{pid}/child/{cid}")
@Produces("application/json")
@Timed(name="child.reads")
public Response readChild(@Context UriInfo uriInfo, @PathParam("pid") String parentId,
        @PathParam("cid") String childId) { ... } 
```

wilkinsona:
Thanks. That'll be it. We need to use Jersey's equivalent of Spring MVC's best matching pattern attribute (assuming it has such a thing).

beckje01:
So does that mean https://github.com/spring-projects/spring-boot/blob/master/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/autoconfigure/MetricsFilter.java#L104 needs to become pluggable?

wilkinsona:
[...]

PR #13429, opened 2018-06-08
MetricsFilter doesn't consider path variables when using Jersey

isopov:
#5875 - fix for this problem that will be fixed in other ways in 2.0.x and further

### DIFF
--- a/spring-boot-actuator/pom.xml
+++ b/spring-boot-actuator/pom.xml
@@ -284,6 +284,11 @@
 			<artifactId>spring-boot-configuration-processor</artifactId>
 			<optional>true</optional>
 		</dependency>
+		<dependency>
+			<groupId>org.glassfish.jersey.core</groupId>
+			<artifactId>jersey-server</artifactId>
+			<optional>true</optional>
+		</dependency>
 		<!-- Test -->
 		<dependency>
 			<groupId>org.springframework.boot</groupId>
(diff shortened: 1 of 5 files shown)


isopov:
@wilkinsona we discussed this is gitter and I don't want you to miss it (especially if this PR is not good enough), so tagging - it is really similar to #5875 and the main good part here is test.

wilkinsona:
Thanks for the PR, @isopov. My feeling is that we shouldn't try to patch this for Jersey. As pointed out in #5875, it would still leave us with a problem for web frameworks other than Spring MVC and Jersey.

As things stand, my preference is to fall back to `unmapped` for a request-response that does not have the best matching pattern attribute. I've implemented that in [this branch](https://github.com/wilkinsona/spring-boot/tree/gh-5875). They are the 1.5.x equivalent of the changes that have been made in 2.0.3 for https://github.com/spring-projects/spring-boot/issues/12447.  The longer term approach (2.1) is to make the metrics filter general purpose and call out to pluggable strategies that provide the necessary information. This would allow us to solve it for Jersey while also providing a documented, supported, plug point for other Frameworks and use cases to be plugged in as well.

wilkinsona:
Having just discussed this with @philwebb, we're going to align 1.5 with 2.0's current behaviour and fall back to `unmapped` when we can get a pattern from Spring MVC. Thanks anyway, @isopov.

isopov:
So this means that for Jersey in 1.5.x all urls will be one and the same "unmapped" even if there are not path veriables in them? Than we'll need to use some workaround...

PR #13438, opened 2018-06-11
Remove compile dependency on JUnit

vtintillier:
With this exclusion, spring-boot users no longer need to exclude
the junit dependency in each of their projects.

### DIFF
--- a/spring-boot-dependencies/pom.xml
+++ b/spring-boot-dependencies/pom.xml
@@ -648,6 +648,13 @@
 				<groupId>com.googlecode.json-simple</groupId>
 				<artifactId>json-simple</artifactId>
 				<version>${simple-json.version}</version>
+				<exclusions>
+					<!-- Remove compile time dependency on junit, see https://github.com/fangyidong/json-simple/issues/91 -->
+					<exclusion>
+						<groupId>junit</groupId>
+						<artifactId>junit</artifactId>
+					</exclusion>
+				</exclusions>
 			</dependency>
 			<dependency>
 				<groupId>com.h2database</groupId>


pivotal-issuemaster:
@vtintillier Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13438)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13438) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@vtintillier Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13438)!

snicoll:
Thanks for the PR

>  spring-boot users no longer need to exclude the junit dependency in each of their projects.

I am not sure what you mean by "each". `json-simple` is  [an optional dependency in Spring Boot](https://github.com/spring-projects/spring-boot/blob/5bd9a445ae231e7d689842dc17281090598a45f7/spring-boot-project/spring-boot/pom.xml#L62) so it will only be included if you include it yourself. I agree that it's annoying but that's more of a problem with json-simple than Spring Boot.

We can decide to add the exclude for the time being but I'd like to research first if we could have detected that automatically.

vtintillier:
The issue is that json-simple is abandonned, and so the fix [1] will never be released, despite it has been requested in [2].
This artifact has even be forked due to its lack of activity [3]

So even if not used by default, as Spring Boot already manages its version it would be convenient if it could exclude the junit dependency.
Otherwise any project that uses Spring Boot dependency management will have to repeat the managed version like, which somehow defeats the advantage of having the version managed in the first place:

```
        <dependency>
            <groupId>com.googlecode.json-simple</groupId>
            <artifactId>json-simple</artifactId>
            <version>1.1.1</version>
            <exclusions>
                <exclusion>
                    <groupId>junit</groupId>
                    <artifactId>junit</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

[1] https://github.com/fangyidong/json-simple/commit/7ae9400664f8b2b43cbd704cf15af52d89b4c773
[2] https://github.com/fangyidong/json-simple/issues/91
[3] https://cliftonlabs.github.io/json-simple/

snicoll:
Thanks for the additional details @vtintillier - I've noticed that looking at the related link and what the Mockito team ended up doing. We'll have to make up our own mind about it. 

The issue is flagged for team attention so we'll get to it ASAP.

PR #13453, opened 2018-06-12
Use @SpringBootTest.properties in tests and samples

izeye:
This PR changes to use `@SpringBootTest.properties` in tests and samples instead of `@TestPropertySource.properties` wherever appropriate.


snicoll:
Thanks again @izeye!

PR #13473, opened 2018-06-13
Reduce object creation in ConfigurationMetadata.findMatchingItemMetadata()

izeye:
This PR changes to use `Collections.emptyList()` in `ConfigurationMetadata.getCandidates()` to avoid object creation.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/ConfigurationMetadata.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/ConfigurationMetadata.java
@@ -145,10 +145,11 @@ public class ConfigurationMetadata {
 	}
 
 	private ItemMetadata findMatchingItemMetadata(ItemMetadata metadata) {
-		List<ItemMetadata> candidates = getCandidates(metadata.getName());
-		if (candidates.isEmpty()) {
+		List<ItemMetadata> candidates = this.items.get(metadata.getName());
+		if (candidates == null || candidates.isEmpty()) {
 			return null;
 		}
+		candidates = new ArrayList<>(candidates);
 		candidates.removeIf((itemMetadata) -> !itemMetadata.hasSameType(metadata));
 		if (candidates.size() > 1 && metadata.getType() != null) {
 			candidates.removeIf(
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/ConfigurationMetadata.java
```diff
@@ -167,7 +167,8 @@ private ItemMetadata findMatchingItemMetadata(ItemMetadata metadata) {
 
 	private List<ItemMetadata> getCandidates(String name) {
 		List<ItemMetadata> candidates = this.items.get(name);
-		return (candidates != null ? new ArrayList<>(candidates) : new ArrayList<>());
+		return (candidates != null ? new ArrayList<>(candidates)
```
snicoll: I am not sure I am keen to merge this. This would return an immutable collection in one case and a mutable in the other. Thoughts?

izeye: @snicoll I thought the same thing but I just chose this as it's a smaller change. How about inlining it in `findMatchingItemMetadata()` as follows?

```
		List<ItemMetadata> candidates = this.items.get(metadata.getName());
		if (candidates == null || candidates.isEmpty()) {
			return null;
		}
		candidates = new ArrayList<>(candidates);
```

snicoll: That works, I guess (using `ObjectUtils.isEmpty`)

izeye: @snicoll `spring-boot-configuration-processor` doesn't depend on Spring Framework, so the above form is okay, right?

izeye: @snicoll I just updated as the above. Please let me know if I miss something.

PR #13480, opened 2018-06-14
ZipHeaderPeekInputStream may advance its position too far when reading into a byte array

izeye:
This PR fixes `this.position` calculation in `JarWriter.read()` as the first `read` value looks accidentally added twice to `this.position`.


wilkinsona:
Thanks once again, @izeye. Could you rebase this on 1.5.x please?

When this is merged, we should also add a test to `ZipHeaderPeekInputStreamTests`. I must have missed a scenario in https://github.com/izeye/spring-boot/commit/34af02380c0a1b94dfa2987cce6375554b04eb79#diff-9cb09c369dfca2dd923576d778bb33e4.

izeye:
@wilkinsona Thanks for the quick feedback! I changed the target branch to `1.5.x` and rebased onto `1.5.x`.

wilkinsona:
#13525 has revealed that the change proposed here isn't quite right. Given the need to make a modified change and add a test, I'm going to close this one in favour of the issue.

PR #13483, opened 2018-06-14
Create a different embedded database for each Datasource when using annotation AutoConfigureTestDatabase

ebussieres:
Currently, when using annotation @ AutoConfigureTestDatabase, there's only the primary datasource that trigger the creation of an embedded database.

This PR create a different embedded database for each Datasource when using annotation @ AutoConfigureTestDatabase

### DIFF
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
@@ -42,6 +42,7 @@ import org.springframework.boot.jdbc.EmbeddedDatabaseConnection;
 import org.springframework.context.EnvironmentAware;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
 import org.springframework.core.env.Environment;
 import org.springframework.jdbc.datasource.embedded.EmbeddedDatabase;
@@ -80,7 +81,7 @@ public class TestDatabaseAutoConfiguration {
 		return new EmbeddedDataSourceBeanFactoryPostProcessor();
 	}
 
-	@Order
+	@Order(Ordered.LOWEST_PRECEDENCE)
 	private static class EmbeddedDataSourceBeanFactoryPostProcessor
 			implements BeanDefinitionRegistryPostProcessor {
 
@@ -161,12 +162,12 @@ public class TestDatabaseAutoConfiguration {
 		}
 
 		@Override
-		public void afterPropertiesSet() {
+		public void afterPropertiesSet() throws Exception {
 			this.embeddedDatabase = this.factory.getEmbeddedDatabase();
 		}
 
 		@Override
-		public DataSource getObject() {
+		public DataSource getObject() throws Exception {
 			return this.embeddedDatabase;
 		}
 
@@ -190,7 +191,7 @@ public class TestDatabaseAutoConfiguration {
 			this.environment = environment;
 		}
 
-		EmbeddedDatabase getEmbeddedDatabase() {
+		public EmbeddedDatabase getEmbeddedDatabase() {
 			EmbeddedDatabaseConnection connection = this.environment.getProperty(
 					"spring.test.database.connection", EmbeddedDatabaseConnection.class,
 					EmbeddedDatabaseConnection.NONE);
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
@@ -16,6 +16,10 @@
 
 package org.springframework.boot.test.autoconfigure.jdbc;
 
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+
 import javax.sql.DataSource;
 
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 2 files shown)


wilkinsona:
@ebussieres Thanks for opening your first contribution to Spring Boot. I see that you've used 1.5.x as the base branch for your pull request, but, given the nature of the change that you are proposing, the earliest release that could include it is 2.1. Could you please update it to use master as the base branch before we review it?

ebussieres:
Sure i'll rebase it

ebussieres:
@wilkinsona Did you have time to review this PR ? It's still with the waiting-for-triage label.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
```diff
@@ -76,7 +80,7 @@ public static EmbeddedDataSourceBeanFactoryPostProcessor embeddedDataSourceBeanF
 		return new EmbeddedDataSourceBeanFactoryPostProcessor();
 	}
 
-	@Order(Ordered.LOWEST_PRECEDENCE)
+	@Order
```
wilkinsona: This change appears to be unrelated.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
```diff
@@ -99,46 +103,46 @@ public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory)
...
 				BeanDefinition beanDefinition = beanFactory.getBeanDefinition(beanName);
-				if (beanDefinition.isPrimary()) {
-					return new BeanDefinitionHolder(beanDefinition, beanName);
-				}
+				beanDefinitionHolders
+						.add(new BeanDefinitionHolder(beanDefinition, beanName));
 			}
-			logger.warn("No primary DataSource found, "
-					+ "embedded version will not be used");
-			return null;
```
wilkinsona: We'll need to think carefully about this. The proposed change could break some tests that had multiple DataSources. Previously, if there was no primary DataSource, no DataSources would be replaced and if there was a primary DataSource, only that DataSource would be replaced. This change will result in all DataSources being replaced which may not be what someone wants. Perhaps we need another mode for `replace`, I'm not sure yet.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
```diff
@@ -156,12 +160,12 @@ public void setEnvironment(Environment environment) {
 		}
 
 		@Override
-		public void afterPropertiesSet() throws Exception {
+		public void afterPropertiesSet() {
```
wilkinsona: This change appears to be unrelated.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
```diff
@@ -156,12 +160,12 @@ public void setEnvironment(Environment environment) {
...
 			this.embeddedDatabase = this.factory.getEmbeddedDatabase();
 		}
 
 		@Override
-		public DataSource getObject() throws Exception {
+		public DataSource getObject() {
```
wilkinsona: This change appears to be unrelated.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
```diff
@@ -185,7 +189,7 @@ public boolean isSingleton() {
 			this.environment = environment;
 		}
 
-		public EmbeddedDatabase getEmbeddedDatabase() {
+		EmbeddedDatabase getEmbeddedDatabase() {
```
wilkinsona: This change appears to be unrelated.

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/jdbc/AutoConfigureTestDatabaseWithMultipleDatasourcesIntegrationTests.java
```diff
@@ -37,21 +38,40 @@
  * datasources.
  *
  * @author Greg Potter
+ * @author Eric Bussieres
  */
 @RunWith(SpringRunner.class)
 @JdbcTest
-@AutoConfigureTestDatabase
```
wilkinsona: This change appears to be unrelated.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfiguration.java
ebussieres: You're right on this point.  But if we don't automatically change it, I think that we need some mechanism to identify which datasource needs to be replaced. Presently, there's no option to define a non-primary datasource to be replaced with an H2 instance for example in a multidatasource environment.

ebussieres: You're right on this point.  But if we don't automatically change it, I think that we need some mechanism to identify which datasource needs to be replaced. Presently, there's no option to define a non-primary datasource to be replaced with an H2 instance for example in a multidatasource environment.

snicoll:
As we don't have any support for multiple datasources in the core at this point, I wonder if we should add such feature at all. I am not denying it is useful but that seems to be a bit on the side with the rest of that domain. I shared the reservation described in https://github.com/spring-projects/spring-boot/pull/13483#discussion_r200628438

wilkinsona:
Thanks again for the pull request. We discussed this today and decided that we don't want to introduce the complexity of dealing with multiple data sources to `@AutoConfigureTestDatabase`. The feature was implemented based on a single, typically auto-configured DataSources and we'd like to keep that simplicity. Thanks anyway.

PR #13484, opened 2018-06-14
Polish HibernateProperties and its test

izeye:
This PR polishes `HibernateProperties` and its test.


snicoll:

PR #13495, opened 2018-06-15
Support the X-Forwarded-* headers in the actuator endpoint listing

chrisgahlert:
I am running a Spring Boot server behind a Nginx reverse proxy. Both are running inside a Docker conntainer. I had the problem, that when I opened the URL `http://proxy-host:8080/my-api/actuator`, the links contained inside where using the internal hostname `http://my-api-service:80/actuator`. For this reason, they were wrong and not clickable.

The current implementation just takes the request URL from the servlet and constructs the links from it.

By using the `ServletUriComponentsBuilder#fromRequest` method, this also works, if we are behind a reverse proxy, if it provides the correct headers(`X-Forwared-Host`, `X-Forwarded-Port`, `X-Forwarded-Prefix`).

Here is an example Nginx-Config, that this will work with:
```
server {
    ...

    location /my-api/ {
        proxy_pass http://my-api-service/;
        proxy_set_header X-Forwarded-Proto $scheme;
        proxy_set_header X-Forwarded-Host $http_host;
        proxy_set_header X-Forwarded-Prefix /my-api;
    }
}
```

Therefore the URL shown will be correct in both cases.

Thanks & kind regards

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/servlet/WebMvcEndpointHandlerMapping.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/servlet/WebMvcEndpointHandlerMapping.java
@@ -67,8 +67,11 @@ public class WebMvcEndpointHandlerMapping extends AbstractWebMvcEndpointHandlerM
 	@ResponseBody
 	protected Map<String, Map<String, Link>> links(HttpServletRequest request,
 			HttpServletResponse response) {
-		String requestUrl = ServletUriComponentsBuilder.fromRequest(request).build().toString();
-		return Collections.singletonMap("_links", this.linksResolver.resolveLinks(requestUrl));
+        String requestUrl = ServletUriComponentsBuilder.fromRequest(request)
+                .build()
+                .toString();
+        return Collections.singletonMap("_links", 
+                this.linksResolver.resolveLinks(requestUrl));
 	}
 
 }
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@chrisgahlert Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13495)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13495) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@chrisgahlert Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=13495)!

bclozel:
I'm not sure using [ServletUriComponentsBuilder](https://github.com/spring-projects/spring-framework/blob/master/spring-webmvc/src/main/java/org/springframework/web/servlet/support/ServletUriComponentsBuilder.java), since the parsing of `X-Forwarded-*` headers won't happen anymore as of Spring 5.1.

chrisgahlert:
What would be a better choice then?

bclozel:
I'm not sure, but configuring a `ForwardedHeaderFilter` would work.

chrisgahlert:
If I understand that correctly, that is something every user would have to do manually? Otherwise a user would still see the "wrong" URLs... I'd prefer a solution that is enabled by default...

bclozel:
Yes, this is a valid approach.
Spring Boot also supports Forwarded headers through server-specific options like `server.use-forward-headers`.

The Spring Framework team chose [to centralize that support as of Spring 5.1](https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-5.x#forwarded-headers).

We can't really enable something like that by default, because:
* you can achieve this with the configuration property, or a filter
* you can choose to limit that support to a specific part of your application
* you can configure a front-end proxy like nginx to deal with this

In any case, this should be a conscious choice by the developers since it can also be abused if the application is not meant to support those headers. With that in mind, I'm closing this PR.

Thanks for your contribution!

PR #13501, opened 2018-06-17
Add reference to dozer-spring-boot-starter

kazuki43zoo:
The [Dozer](https://github.com/DozerMapper/dozer) is the Java Bean mapper. Since the Dozer 6.2.0(current latest version), the dozer-spring-boot-starter module was added. Hence; I will introduce the it as community starter.

### DIFF
--- a/spring-boot-project/spring-boot-starters/README.adoc
+++ b/spring-boot-project/spring-boot-starters/README.adoc
@@ -67,6 +67,9 @@ do as they were designed before this was clarified.
 | https://github.com/docker-java/docker-java/[Docker Java] and https://github.com/spotify/docker-client/[Docker Client]
 | https://github.com/jliu666/docker-api-spring-boot
 
+| https://github.com/DozerMapper/dozer[Dozer]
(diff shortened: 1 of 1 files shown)


bclozel:
Looking at the Dozer starter, it seems it is only compatible with Spring Boot 1.5 - is that the case?
In the meantime, we're thinking about pruning the current list from starters incompatible with Spring Boot 2.x. This should be done on the master/2.0.x branches.

Depending on the outcome of that discussion in #13510, this PR could be declined.

kazuki43zoo:
@bclozel @snicoll Thanks for your replying!
I will consult it with a maintainer of dozer.

kazuki43zoo:
@bclozel 
I've added the compatibility checker project on my account. See https://github.com/kazuki43zoo/dozer-spring-boot-compatibility-checker. The dozer-spring-boot-starter 6.2.0 depends on Spring Boot 1.5.x, but it works fine on Spring Boot 1.4.x - 2.1.x.

@snicoll 
I've fixed your comment.

Issue #11693, opened 2018-01-19
EmbeddedLdapAutoConfiguration not setting the base information into the LdapContextSource


vdubus:
Currently (as of version **2.0.0.M7**), `EmbeddedLdapAutoConfiguration` isn't properly setting the `base` information from `LdapContextSource` using the value from `spring.ldap.embedded.base-dn`.
Because of that, we can't test our `@repository` classes using an embedded LDAP server without overriding this bean unless if we want to set the whole path for each `@Entry`.

I would suggest the following change inside of `org.springframework.boot.autoconfigure.ldap.embedded.EmbeddedLdapAutoConfiguration#ldapContextSource`:

```java
	@Bean
	@DependsOn("directoryServer")
	@ConditionalOnMissingBean
	public ContextSource ldapContextSource() {
		LdapContextSource source = new LdapContextSource();
		if (hasCredentials(this.embeddedProperties.getCredential())) {
			source.setUserDn(this.embeddedProperties.getCredential().getUsername());
			source.setPassword(this.embeddedProperties.getCredential().getPassword());
		}
		source.setBase(this.embeddedProperties.getBaseDn()); // Add this line.
		source.setUrls(this.properties.determineUrls(this.environment));
		return source;
	}
```

This issue was already reported in the issue #10443 which was closed without fixing the problem.

snicoll:
> This issue was already reported in the ticket #10443 which was closed without fixing the problem.

That's inaccurate. #10443 has been closed as a duplicate of #10444 and the latter has been declined.

vdubus:
Ok, so because PR #10444 was made to fix Issue #10443, you closed the issue as a duplicate of the PR.
Then, you closed PR #10444 which isn't progressing since then.
And now you closed this ticket which report back the problem as a bug.
Meaning that there isn't even an opened ticket to signal the bug.

Would you mind reopening at least one ticket so that someone could find it and try working on it?
Unless there is another opened ticket of which I am unaware?

snicoll:
> Then, you closed PR #10444 which isn't progressing since then.

It hasn't progressed because the proposal was unsatisfactory. But you're right, looking back at the history, we could use a report for this. Let's use this one then.

tgeens:
The original change proposed here no longer cleanly applies after the changes in https://github.com/spring-projects/spring-boot/pull/11764/files - EmbeddedLdapProperties.getBaseDn() now returns a list, while LdapContextSource.setBase(base) expects a string

[...]

PR #10444, opened 2017-09-28
Ldap contextSource base property populating in EmbeddedLdapAutoConfig

pavelda2:
Fix issue spring-projects/spring-boot#10443

PR #13502, opened 2018-06-17
Add support to set base in EmbeddedLdapAutoConfiguration

eddumelendez:
Currently, full path should be present when queries are  performed.
This commit introduces support to set `base` in `EmbeddedLdapAutoConfiguration`.

By default, first entry in `spring.ldap.embedded.base-dn` will be
resolved.

See gh-11693

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfiguration.java
@@ -106,6 +106,7 @@ public class EmbeddedLdapAutoConfiguration {
 			source.setUserDn(this.embeddedProperties.getCredential().getUsername());
 			source.setPassword(this.embeddedProperties.getCredential().getPassword());
 		}
+		source.setBase(this.embeddedProperties.getBaseDn().get(0));
 		source.setUrls(this.properties.determineUrls(this.environment));
 		return source;
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/ldap/embedded/EmbeddedLdapAutoConfigurationTests.java
@@ -133,7 +133,7 @@ public class EmbeddedLdapAutoConfigurationTests {
 					assertThat(context.getBeanNamesForType(LdapTemplate.class).length)
 							.isEqualTo(1);
 					LdapTemplate ldapTemplate = context.getBean(LdapTemplate.class);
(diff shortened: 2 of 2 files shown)


eddumelendez:
AFAICS spring-ldap just support one `base`. So, the documentation should be pretty similar to handling multiple datasources (I can add some documentation if we agreed to proceed with this change). In this case, building `LdapContextSource` for every `dn` and its `LdapTemplate`. But, thinking about that, it would look cumbersome.

Just another idea, would be create a `MultiLdapTemplate` which can be query all the `ldapTemplate` with the right `base` set in `LdapContextSource`. I think, this can be implemented in `spring-ldap` project and reused by `spring-boot`.

On the other hand, I also considering this such a breaking change not only for current integration test also for current projects using embedded ldap support.

@snicoll WDYT?

snicoll:
To be completely honest? Not much I am afraid and I didn't get the analogy with our `DataSource` support.

It's not so much about the "one base" bit. It is about "the first on the list" _and_ there is no way to disable that behaviour. If we want to move forward, we have to take care of the Spring Data integration tests situation: either we believe these are wrong (and then we should do something about that) or they are legit and then we shouldn't break them. Applying this PR will last time I checked.

eddumelendez:
> It is about "the first on the list" and there is no way to disable that behaviour.

Agree with this and I understand your worries. 

Another option can be adding a customizer.

snicoll:
Or a `boolean` flag that sets the `base`, possibly `true` by default so that the behaviour we want is enabled. Or with no default value so that embedded and non embedded mode have different defaults. 

That said, I have no idea, I need to better understand the feature and the impact.

snicoll:
@eddumelendez gentle ping. As you know Spring LDAP way better than I do, perhaps you have some feedback on my (potentially dumb) proposal.

eddumelendez:
@snicoll sorry I miss the first notification and thanks for the gentle reminder about this topic.

Set `base` in `LdapContextSource` will work fine when there is just one `base` to work with if we are using just `LdapTemplate` provided by `spring-ldap` but will raise an issue if the project is using `Object Directory Mapping` provided by `spring-data-ldap` with the `base` attribute set in `@Entry`, that's the reason why integration tests were reported broken [here](https://github.com/spring-projects/spring-boot/pull/10444#issuecomment-333467536). e.g `spring.ldap.embedded.base-dn:dc=spring,dc=org`

If we set more than one

```
spring.ldap.embedded.base-dn[0]:dc=spring,dc=org
spring.ldap.embedded.base-dn[1]:dc=pivotal,dc=io
```

we will not be able to resolve a default `base` to set in `LdapContextSource` because `spring-ldap` supports just one `base` per `LdapContextSource`. But, in 270dc2cd7281ae80f595133ea6cf547e45856d08 and eab0b84a80e4e70bbf84289e1318e99c70ff899e  multi `baseDn` was introduced for `EmbeddedLdapAutoConfiguration`. Thinking more about it, even a flag enabling set the `base` would be weird. Otherwise, we can evaluate if there is just one `baseDn` and set the value, if there are more we will not set any `base`, we should also document this behaviour and take into account what I said before

> will raise issue if the project is using `Object Directory Mapping` with the `base` attribute set in `@Entry`

IMHO, the current state is good and we shouldn't modify it. This PR doesn't make sense to me now.

PR #13512, opened 2018-06-18
Add log level guards in ConfigFileApplicationListener

izeye:
This PR adds log level guards in `ConfigFileApplicationListener`.


snicoll:
Thanks again @izeye!

PR #13545, opened 2018-06-21
Fix query string encoding in TraceableHttpServletRequest

izeye:
This PR fixes query string encoding in `TraceableHttpServletRequest` as encoding "=" doesn't seem to be right.


snicoll:
Good catch again @izeye, thanks!

PR #13554, opened 2018-06-22
Remove CompoundEnumeration in RestartClassLoader

dreis2211:
Hi,

this PR removes the need for a special CompoundEnumeration in RestartClassLoader and replaces it with standard JDK functionality and therefore removes a bit complexity imho.

Let me know what you think.

Cheers,
Christoph


dreis2211:
Test failure seems to be unrelated.

philwebb:
I think the enumerator approach is probably more efficient since it isn't creating a new array for all the elements. Thanks for the suggestion, but I think we should leave this one as it is.

dreis2211:
Fair enough from a performance point of view. Thanks for taking the time @philwebb

Issue #11725, opened 2018-01-22
Support Hibernate 5.3.0 / JPA 2.2 (JDK8 Streams for pagination)


Blarne:
Based on functionality supporting streams JDK8 over JPA (automatic pagination) described in https://vladmihalcea.com/whats-new-in-jpa-2-2-stream-the-result-of-a-query-execution/ included JPA 2.2 with Hibernate 5.3.0.Beta1 - I recommend upgrade dependencies on Spring Boot Dependencies to:

```
<!-- https://mvnrepository.com/artifact/org.hibernate/hibernate-core -->
<dependency>
    <groupId>org.hibernate</groupId>
    <artifactId>hibernate-core</artifactId>
    <version>5.3.0.Beta1</version>
</dependency>

<!-- https://mvnrepository.com/artifact/javax.persistence/javax.persistence-api -->
<dependency>
    <groupId>javax.persistence</groupId>
    <artifactId>javax.persistence-api</artifactId>
    <version>2.2</version>
</dependency>
```
I can contribute on this issue, if possible.

wilkinsona:
Thanks, but things aren't ready for this yet. We're very close to shipping Boot 2.0 and we cannot use a beta version of Hibernate in a Boot GA. We may also need to wait for Spring Framework and Spring Data JPA to support JPA 2.2 and Hibernate 5.3. Once they do, Boot's upgrade to Hibernate 5.3 will follow.

Blarne:
Yes, undestand, I agree with stable version of HBM too.

jhoeller:
FYI, we're tracking Hibernate ORM 5.3 support in [SPR-16303](https://jira.spring.io/browse/SPR-16303), aiming for the Spring Framework 5.1 timeframe in Q2 2018 (when Hibernate ORM 5.3 itself will hopefully be GA). For the time being, the 5.3 Beta 1 release is known to work fine with the upcoming Spring Framework 5.0.3, just with no dedicated support for any Hibernate 5.3 specific features there yet.

As for the JPA 2.2 API, that's explicitly supported in Spring Framework 5.0 already, including proper handling of result streams in non-transactional scenarios. Since JPA 2.2 doesn't make much difference for Spring's management at runtime, in particular in regular transactional setups, providers such as EclipseLink 2.7 or Hibernate ORM 5.3 should even work fine with Spring Framework 4.3.x as well.

ask4gilles:
[...]

PR #13570, opened 2018-06-25
Infinispan 9.3.0 update

galderz:
* Update to Infinispan 9.3.0.Final.
* Remove Infinispan uber jar dependencies.
* Update to Hibernate to use Infinispan cache provider compatible with that Hibernate version.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -1869,11 +1869,6 @@
 			</dependency>
 			<dependency>
 				<groupId>org.hibernate</groupId>
-				<artifactId>hibernate-infinispan</artifactId>
-				<version>${hibernate.version}</version>
-			</dependency>
-			<dependency>
-				<groupId>org.hibernate</groupId>
 				<artifactId>hibernate-java8</artifactId>
 				<version>${hibernate.version}</version>
 			</dependency>
@@ -2009,7 +2004,7 @@
 			</dependency>
 			<dependency>
 				<groupId>org.infinispan</groupId>
-				<artifactId>infinispan-hibernate-cache</artifactId>
+				<artifactId>infinispan-hibernate-cache-v53</artifactId>
 				<version>${infinispan.version}</version>
 			</dependency>
 			<dependency>
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -76,7 +76,7 @@
 		<httpasyncclient.version>4.1.3</httpasyncclient.version>
 		<httpclient.version>4.5.5</httpclient.version>
 		<httpcore.version>4.4.10</httpcore.version>
-		<infinispan.version>9.1.7.Final</infinispan.version>
+		<infinispan.version>9.3.0.Final</infinispan.version>
 		<influxdb-java.version>2.9</influxdb-java.version>
 		<jackson.version>2.9.6</jackson.version>
 		<janino.version>3.0.8</janino.version>


wilkinsona:
Thanks for the PR. We generally try to keep changes has focussed as possible and the proposal here is mixing a few different changes. Some of those changes, such as removing dependency management for modules that still exist, shouldn't be made at all. We already have an issue for upgrading to Hibernate 5.3.x (#11725). Can you please re-work this pull request so that it focuses solely on the changes that are required for the Infinispan upgrade? If the Hibernate upgrade is a prerequisite then we can take care of that separately.

galderz:
Actually, Hibernate 5.3 has already been updated. I'll just fix the cache provider then.

galderz:
Should be ready now :)

wilkinsona:
Thanks for the updates. Do the modules that you have removed from dependency management no longer exist in Infinispan 9.3?

We have a policy of providing dependency management for all of a library's modules. This makes it easier to keep a library's dependencies in sync even for those modules that Boot itself doesn't use. If the modules still exist, we should keep the dependency management for them and just add entries for any new modules.

galderz:
* `org.hibernate:hibernate-infinispan` - the hibernate/infinispan integration is no longer in this location, the cache provider is now in Infinispan source tree and the module names have changed to adjust to individual Hibernate versions. Hence, the Hibernate 5.3 compatible Infinispan cache provider is located in `org.infinispan:infinispan-hibernate-cache-v53`

* `infinispan-embedded-query`, `infinispan-embedded`, `infinispan-cli` - these are Infinispan uber jar dependencies which cause issues in Maven environments. They should never have been there in the first place. These uber jars are for environments with no dependency management capabilities, e.g. ant wo/ ivy.

wilkinsona:
>  these are Infinispan uber jar dependencies which cause issues in Maven environments. They should never have been there in the first place.

I don't agree with that. We don't presume to know why someone may be using a dependency. It could be for inclusion in an assembly, for example, rather than something that's used for compilation. Either way, the changes are unrelated to an upgrade to 9.3 so they do not belong here.

galderz:
> I don't agree with that. We don't presume to know why someone may be using a dependency. It could be for inclusion in an assembly, for example, rather than something that's used for compilation.

We're trying to avoid people using them at all since most of the time people are using them for compilation and have had issues like [this](https://issues.jboss.org/browse/ISPN-6865). So, if they're not there, less chances of making the mistake. 

Besides, in recent Infinispan versions we don't publish them to Maven any more. For example, `infinispan-embedded` no longer published to Maven since 9.1.x: http://repo2.maven.org/maven2/org/infinispan/infinispan-embedded 

> Either way, the changes are unrelated to an upgrade to 9.3 so they do not belong here.

Fair point. I'll remove that and open a new PR with the other changes.

wilkinsona:
Thanks.

> Besides, in recent Infinispan versions we don't publish them to Maven any more.

The modules that are no longer published in 9.3 should be removed as part of the upgrade. Any other removals should be considered separately.

galderz:
I have updated the PR removing the uber jar removals

galderz:
Rebased PR. @wilkinsona What is going to be in the end? I've left the uber jars for now and then I'll send another PR to remove them.

wilkinsona:
Thanks for the PR, @galderz. The proposed changes are now in master, along with [another commit](https://github.com/spring-projects/spring-boot/commit/bc65c4cb2c733ec88837eacbc618d6c11e5642c9) that aligned the dependency management with modules that have been added or removed between Infinispan 9.1.7 and 9.3.0.

For the reasons explained above, we want to keep the dependency management for any remaining uber jars so I don't think any further changes are necessary here.

galderz:
Thanks @wilkinsona :)

PR #13572, opened 2018-06-25
Fix typo

dreis2211:


dreis2211:
Test failure is unrelated...Seems to be because of missing parameter names, which breaks actuator endpoints that make use `@Selector`.

dreis2211:
Thanks @wilkinsona . I haven't included `-parameters` yet on the IDE after a fresh install, sorry. Was mislead by the sort of tests that were failing on my side because of that.

mbhave:
Thanks for the PR @dreis2211. This is now on master.

PR #13582, opened 2018-06-26
Allow to configure Quartz's "overwriteExistingJobs" property

tdanylchuk:
As for now only way to change "overwriteExistingJobs" property is to create custom `SchedulerFactoryBeanCustomizer` with only purpose - to set one parameter.

If we create trigger via spring configuration and get e.g. cron expression from properties, in case of change cron expression property we need to re-scheduler job on start-up of the service.
I think it's pretty much common use-case to be simplified with help of `QuartzAutoConfiguration`.


snicoll:
@tdanylchuk thank you very much for your first contribution to Spring Boot. This is now merged in `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/66cb4ce3c7f9ba1091786d23737a4c0436862cf0). It turns out there were other properties that should be exposed, I've done that as well (see #13782).

tdanylchuk:
Yeah, thank you, wanted to propagate them also, but didn't want to overextend the PR.

PR #13619, opened 2018-06-30
Improve description of Endpoint exclude

72MiguelGomes:
After reading the code notice that the `*` was also valid for exclusion and that was not documented as in included fields.

Example:

**ExposeExcludePropertyEndpointFilter.class**
```java
private boolean isExposed(ExposableEndpoint<?> endpoint) {
	if (this.include.isEmpty()) {
		return this.exposeDefaults.contains("*")
				|| contains(this.exposeDefaults, endpoint);
	}
	return this.include.contains("*") || contains(this.include, endpoint);
}

private boolean isExcluded(ExposableEndpoint<?> endpoint) {
	if (this.exclude.isEmpty()) {
		return false;
	}
	return this.exclude.contains("*") || contains(this.exclude, endpoint);
}
```


philwebb:
Good spot. Thanks for the PR!

snicoll:
@72MiguelGomes thank you for your first contribution to Spring Boot. This is now merged in `2.0.x` and `master`. I've polished your contribution in 44b9aa1

PR #13621, opened 2018-06-30
Remove mock from Final Class.

72MiguelGomes:
After a PR I noticed the build was broken and after some investigation noticed that we were mocking a class that become final in SpringFramework(https://github.com/spring-projects/spring-framework/commit/eeebd51#diff-c3fe703036ed412c6797be9c6099f3b7). 

Currently Mockito supports mocking final classes but just as an experimental feature (https://github.com/mockito/mockito/wiki/What%27s-new-in-Mockito-2#unmockable), so I decided remove the mock for that class as we don't need that for that specific test.


72MiguelGomes:
Seems this problem was already addressed.

Closing the PR

Issue #13550, opened 2018-06-22
Add support for new SpringLiquibase properties from 3.6.2


isopov:
In liquibase/liquibase#774 support for liquibase properties
- liquibaseSchema
- databaseChangeLogTable
- databaseChangeLogLockTable
- liquibaseTablespace

was added to SpringLiquibase integration and it should be available in 3.6.2 release. (Yet to be released) Once updated to this version of  liquibase, please add support for these properties in LiquibaseAutoConfiguration and LiquibaseProperties

72MiguelGomes:
@isopov I already created a PR with the support (https://github.com/spring-projects/spring-boot/pull/13625), this way part of the work is already done.

philwebb:
Closing in favor of PR #13625

Issue #8237, opened 2017-02-08
Auto-generate the "Common application properties" appendix


garyrussell:
Given the size of the properties appendix, consider adding anchors between property sets, to make it easier to navigate to technology-based properties... 

- from the ToC
- as a hyperlink when answering StackOverflow questions etc

Something like...

```
--- a/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
+++ b/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
@@ -13,7 +13,10 @@ your own properties.
 WARNING: This sample file is meant as a guide only. Do **not** copy/paste the entire
 content into your application; rather pick only the properties that you need.
 
+=== Properties
 
+[[appendix-props-core]]
+==== Core Properties
 [source,properties,indent=0,subs="verbatim,attributes,macros"]
 ----
        # ===================================================================

... 

@@ -923,7 +926,12 @@ content into your application; rather pick only the properties that you need.
        spring.kafka.ssl.truststore-location= # Location of the trust store file.
        spring.kafka.ssl.truststore-password= # Store password for the trust store file.
        spring.kafka.template.default-topic= # Default topic to which messages will be sent.
+----
 
+[[appendix-props-rabbit]]
+==== RabbitMQ Properties
+[source,properties,indent=0,subs="verbatim,attributes,macros"]
+----
        # RABBIT ({sc-spring-boot-autoconfigure}/amqp/RabbitProperties.{sc-ext}[RabbitProperties])
        spring.rabbitmq.addresses= # Comma-separated list of addresses to which the client should connect.
        spring.rabbitmq.cache.channel.checkout-timeout= # Number of milliseconds to wait to obtain a channel if the cache size has

...
```

We then get a ToC entry, as well as an anchor which can be copied to the clipboard.

<img width="332" alt="screen shot 2017-02-08 at 12 28 24 pm" src="https://cloud.githubusercontent.com/assets/483832/22749251/a1b04632-edfa-11e6-8904-4a4985816c65.png">

garyrussell:
I am willing to submit a PR if this passes triage.

http://stackoverflow.com/questions/42118570/set-heartbeat-property-on-rabbitmq-autoconfig-connectionfactory-bean/42118917#42118917

wilkinsona:
[...]

PR #13625, opened 2018-06-30
Upgrade to Liquibase 3.6.2

72MiguelGomes:
The new version of SpringLiquibase was not released yet, but will add this PR here, and as soon as the new version is available we just need to update it and merge.

Issue: https://github.com/spring-projects/spring-boot/issues/13550.

This PR should keep on hold for a little. @isopov as soon as the new version was released please update us.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
@@ -54,6 +54,26 @@ public class LiquibaseProperties {
 	private String defaultSchema;
 
 	/**
+	 * Liquibase database schema.
+	 */
+	private String liquibaseSchema;
+
+	/**
+	 * Liquibase tablespace.
+	 */
+	private String liquibaseTablespace;
+
+	/**
+	 * Database changelog table.
+	 */
+	private String databaseChangeLogTable;
+
+	/**
+	 * Database changelog lock table.
+	 */
+	private String databaseChangeLogLockTable;
+
+	/**
 	 * Whether to first drop the database schema.
 	 */
 	private boolean dropFirst;
@@ -130,6 +150,38 @@ public class LiquibaseProperties {
 
 	public void setDefaultSchema(String defaultSchema) {
 		this.defaultSchema = defaultSchema;
+	}
+
+	public String getLiquibaseSchema() {
+		return this.liquibaseSchema;
+	}
+
+	public void setLiquibaseSchema(String liquibaseSchema) {
+		this.liquibaseSchema = liquibaseSchema;
+	}
+
+	public String getLiquibaseTablespace() {
+		return this.liquibaseTablespace;
+	}
+
+	public void setLiquibaseTablespace(String liquibaseTablespace) {
+		this.liquibaseTablespace = liquibaseTablespace;
+	}
+
+	public String getDatabaseChangeLogTable() {
+		return this.databaseChangeLogTable;
+	}
+
+	public void setDatabaseChangeLogTable(String databaseChangeLogTable) {
+		this.databaseChangeLogTable = databaseChangeLogTable;
+	}
+
+	public String getDatabaseChangeLogLockTable() {
+		return this.databaseChangeLogLockTable;
+	}
+
+	public void setDatabaseChangeLogLockTable(String databaseChangeLogLockTable) {
+		this.databaseChangeLogLockTable = databaseChangeLogLockTable;
 	}
 
 	public boolean isDropFirst() {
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-dependencies/pom.xml
```diff
@@ -118,7 +118,7 @@
 		<kafka.version>1.1.0</kafka.version>
 		<kotlin.version>1.2.41</kotlin.version>
 		<lettuce.version>5.1.0.M1</lettuce.version>
-		<liquibase.version>3.6.1</liquibase.version>
+		<liquibase.version>3.6.2-SNAPSHOT</liquibase.version>
```
72MiguelGomes: Please notice this was used just to test, we must remove the `SNAPSHOT`as soon as the new versions is available

isopov:
@72MiguelGomes 3.6.2 liqubiase version has been released

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
```diff
@@ -53,6 +53,26 @@
...
+	 */
+	private String liquibaseSchema;
+
+	/**
+	 * Liquibase tablespace.
+	 */
+	private String liquibaseTablespace;
+
+	/**
+	 * Database changelog table.
```
snicoll: table name?

72MiguelGomes:
@snicoll thanks for the feedback, will update it asap.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
72MiguelGomes: What do you mean? I am using the same that used in liquibase project

72MiguelGomes:
@snicoll Those values does not have any default value define, so there is no problem.

https://github.com/liquibase/liquibase/blob/master/liquibase-core/src/main/java/liquibase/integration/spring/SpringLiquibase.java#L81

Regarding de adoc, do you know if is possible generate that automatically during the build or is it always a manual process?

@philwebb The PR is ready to merge, please let me know if you find something that is missing.

@isopov Thanks for letting me know about liquibase release.

snicoll:
> @snicoll Those values does not have any default value define, so there is no problem.

uh? Surely there must be some sort of default somewhere right?

> Regarding de adoc, do you know if is possible generate that automatically during the build or is it always a manual process?

That's how it is right now, you can subscribe to #8237.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
snicoll: I mean that the description can be more informative.

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
```diff
@@ -566,6 +566,10 @@ content into your application. Rather, pick only the properties that you need.
 	spring.liquibase.check-change-log-location=true # Whether to check that the change log location exists.
 	spring.liquibase.contexts= # Comma-separated list of runtime contexts to use.
 	spring.liquibase.default-schema= # Default database schema.
+	spring.liquibase.liquibase-schema= # Liquibase database schema.
+  spring.liquibase.liquibase-tablespace= # Liquibase tablespace.
```
snicoll: There is an indentation problem here.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
snicoll: In `GlobalConfiguration` the description is 

> Name of table to use for tracking change history

which is way more informative than the Javadoc. I prefer to use that

snicoll:
@72MiguelGomes please don't act on the default value as I'd like to do this globally (https://github.com/spring-projects/spring-boot/issues/13765). 

Can you please fix the indentation and use a better description for those new keys?

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
72MiguelGomes: Thanks

snicoll:
@72MiguelGomes thank you, this is now merged in `master`.

DanielFran:
@snicoll Is it possible to include this to next 2.0.4 release too?

wilkinsona:
Unfortunately not. 2.0.x uses Liquibase 3.5.x and we won't upgrade to a new minor release of Liquibase in a maintenance release of Spring Boot.

PR #13629, opened 2018-07-01
Upgrade to Maven War Plugin 3.2.2

khmarbaise:
Upgrade to most recent version of maven-war-plugin


wilkinsona:
Thank you for you contribution. This change is now in master.

PR #13630, opened 2018-07-01
Upgrade maven plugins:

khmarbaise:
o maven-clean-plugin 3.1.0
 o maven-dependency-plugin 3.1.1
 o maven-help-plugin 3.1.0
 o maven-resources-plugin 3.1.0


wilkinsona:
Thanks for the pull request. We prefer to handle one upgrade per issue or pull request. This makes it easier to see what’s changed when browsing a milestone. Also, for straightforward one-like changes like these, it’s easier if we handle them ourselves. We have a semi-automated process that makes dependency upgrades very quick to do. Thanks anyway.

PR #13631, opened 2018-07-01
Remove maven-eclipse-plugin as the project has been retired

khmarbaise:
The maven-eclipse-plugin is retired for a long time and it's recommend not to use it anymore cause M2E in Eclipse handles everything better.


wilkinsona:
@khmarbaise Thank you very much for your first contribution to Spring Boot. This change is now in master.

PR #13707, opened 2018-07-05
Fix parameter order for RequestMatcherAssert.doesNotMatch()

izeye:
Callers for `RequestMatcherAssert.doesNotMatch()` have been updated to align with `RequestMatcherAssert.matches()` in fddc9e9c7e7300d93c0ba44ef702eb205326b3e2 but its signature itself hasn't been updated. This PR updates its parameter order to align with its callers.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/security/servlet/EndpointRequestTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/security/servlet/EndpointRequestTests.java
@@ -78,14 +78,15 @@ public class EndpointRequestTests {
 	@Test
 	public void toAnyEndpointWhenServletPathNotEmptyShouldMatch() {
 		RequestMatcher matcher = EndpointRequest.toAnyEndpoint();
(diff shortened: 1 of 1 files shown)


wilkinsona:
I think the current order (`servletPath` then `pathInfo`) makes sense. Isn't it only reversed in the matches case to accommodate the use of varags?

If anything's to change, I think I'd prefer to see matches become (List<String> servletPaths, String pathInfo) instead, but I'll defer to @mbhave as she is more familiar with this code than I am.

izeye:
@wilkinsona If the current order needs to be preserved, I think the changes on `doesNotMatch()` calls in fddc9e9c7e7300d93c0ba44ef702eb205326b3e2 need to be reverted.

izeye:
@wilkinsona Sorry, I misread your comment a bit. I agree with the suggestion and can change this that way if @mbhave agrees.

mbhave:
Good catch, @izeye! I agree with @wilkinsona's suggestion too.

bclozel:
This CI issue due to the Reactor Netty changes should be fixed now.

izeye:
@wilkinsona @mbhave @bclozel Thanks! I changed as suggested.

mbhave:
Thanks @izeye. The change is now on master.

PR #13710, opened 2018-07-05
Separate logging from getUrlMappings()

izeye:
Logging in "get*" methods doesn't seem usual, so this PR separates it from `getUrlMappings()` and places it to a similar place as before fddc9e9c7e7300d93c0ba44ef702eb205326b3e2.


izeye:
I rebased this onto the current `master` to resolve conflicts.

snicoll:
> I rebased this onto the current master to resolve conflicts.

Thanks but I think it would be worth harmonizing `2.0.x` as well (especially since the log output would actually show what was registered). Can you please change the base branch of this one to `2.0.x`?

izeye:
@snicoll Sure, I changed the base to `2.0.x` and force-pushed.

PR #13711, opened 2018-07-05
Split compound if statement in two so that Graal's native image builder is happy

dsyer:
Some static analysis tools will barf on BeanDefinitionLoader when Groovy is not on the classpath (e.g. Graal VM native image builder). The problem can be avoided just by nesting the "if" clause containing the Groovy reference inside a defensive outer "if".

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/BeanDefinitionLoader.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/BeanDefinitionLoader.java
@@ -147,12 +147,13 @@ class BeanDefinitionLoader {
 	}
 
 	private int load(Class<?> source) {
-		if (isGroovyPresent()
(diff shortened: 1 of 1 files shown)


wilkinsona:
Equally, some static analysis tools will complain that the two if statements should be combined into one. I'm surprised that Checkstyle hasn't done so.

We can merge this if really necessary, but I'd prefer that it was fixed in Graal rather than it forcing lots of projects to make their code a little bit worse.

philwebb:
Perhaps we can extract a method to make it look a little less odd.

@dsyer Is there an open Graal issue for this?

dsyer:
There is a [graal issue](https://github.com/oracle/graal/issues/523) and they marked it as a bug, so maybe we do not need the do this here.

wilkinsona:
Yeah, looks like this hopefully won't be needed. I've subscribed to the Graal issue. I'll close this one for now and we can re-open it if the Graal issue doesn't help.

PR #13727, opened 2018-07-08
Polish some Collectors

dreis2211:
Hi,

this trivial PR replaces usages like `Collectors.toCollection(ArrayList::new)` with `Collectors.toList()`.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/endpoint/ExposeExcludePropertyEndpointFilter.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/endpoint/ExposeExcludePropertyEndpointFilter.java
@@ -20,7 +20,6 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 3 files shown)


philwebb:
Although internally `Collectors.toList()` returns an `ArrayList` the javadoc made me nervous enough to not use it:

> here are no guarantees on the type, mutability, serializability, or thread-safety of the List returned

It's probably very unlikely to change in future versions, but you never know. I'll flag this one and see what the rest of the team think.

snicoll:
FTR, I've seen `Collectos.toList()` being used in many places so if we don't like that, we should probably change those as well.

wilkinsona:
It all depends on how the list is used. If it's only used internally then it's probably fine. Problems are more likely to arise when it's passed to another component that relies on certain behaviour.

dreis2211:
I guess this PR is safe in that regards. The lists/sets/collections we're talking about do not rely on anything special imho, but please check for yourself.

philwebb:
Yeah, double checking those changes and I think they are all safe. The only one that's a bit iffy is the result of `AutoConfigurations.sort` which I guess someone might expect to be always mutable.

snicoll:
Agreed, I've reverted that last change as I think we should enforce the fact we expect an order for the list there.

PR #13757, opened 2018-07-12
Remove @since tag from getServlet()

izeye:
> If a member changes from protected to public in a later release, the `@since` tag would not change, even though it is now usable by any caller, not just subclassers.

Based on the above description, it looks `@since` tag could be removed.

See http://www.oracle.com/technetwork/java/javase/documentation/index-137868.html#@since


wilkinsona:
Given that `getServlet()` appears in the [javadoc for 2.0.3](https://docs.spring.io/spring-boot/docs/2.0.3.RELEASE/api/org/springframework/boot/web/servlet/ServletRegistrationBean.html#getServlet--), I agree with removing `@since`.

snicoll:
Sorry about that, I mixed up with package private class becoming public.

PR #13776, opened 2018-07-15
Polish

izeye:


snicoll:
Thanks again @izeye - I've split the polish commit in two as some of it affected `2.0.x` as well.

PR #13779, opened 2018-07-16
Remove carriage returns in TypeUtils.getJavaDoc()

izeye:
This PR changes to use system-dependent line separator in `TypeUtils.getJavaDoc()`.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeUtils.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeUtils.java
@@ -142,8 +142,7 @@ class TypeUtils {
 		String javadoc = (element != null
 				? this.env.getElementUtils().getDocComment(element) : null);
 		if (javadoc != null) {
-			javadoc = javadoc.replaceAll("\\n", "");
-			javadoc = javadoc.trim();
+			javadoc = javadoc.replaceAll("[\r\n]+", "").trim();
 		}
 		return ("".equals(javadoc) ? null : javadoc);
 	}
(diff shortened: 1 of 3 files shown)


izeye:
On a second thought both line feed ('\n') and carriage return ('\r') should be removed, not system-dependent line separator. I'll update shortly.

izeye:
I've updated to remove carriage returns in addition to line feeds.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeUtils.java
```diff
@@ -142,8 +142,7 @@ public String getJavaDoc(Element element) {
 		String javadoc = (element != null
 				? this.env.getElementUtils().getDocComment(element) : null);
 		if (javadoc != null) {
-			javadoc = javadoc.replaceAll("\\n", "");
-			javadoc = javadoc.trim();
+			javadoc = javadoc.replaceAll("[\r\n]+", "").trim();
```
snicoll: Can we do two passes instead (`\r\n` then `\n`)? Technically speaking that would replace `\r`

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeUtils.java
```diff
@@ -142,8 +142,7 @@ public String getJavaDoc(Element element) {
 		String javadoc = (element != null
 				? this.env.getElementUtils().getDocComment(element) : null);
 		if (javadoc != null) {
-			javadoc = javadoc.replaceAll("\\n", "");
-			javadoc = javadoc.trim();
```
snicoll: I still would like surrounding spaces to be removed. Do you disagree with that?

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeUtils.java
izeye: @snicoll I thought removing "\r" in Javadoc, especially as description for application properties is safe. Am I missing something?

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeUtils.java
izeye: @snicoll No, I just collapsed the two statements into one. I thought it's less verbose than before.

snicoll: I missed that 😓

snicoll:
Thanks a lot for the feedback @izeye

PR #13790, opened 2018-07-16
Polish

izeye:


snicoll:
Thanks a lot again @izeye. I've split the polish commit in two and found some `depend-on` leftover that I fixed in a polish polish commit 😄

PR #13792, opened 2018-07-16
Add missing @Test in RestarterTests

dreis2211:
Hi,

I just stumbled upon this missing @Test annotation.

Cheers,
Christoph


snicoll:
Good catch @dreis2211, thanks again!

PR #13813, opened 2018-07-18
Base unit of metric values is unspecified when using the metrics endpoint

### LINKED ISSUE #13791: Include a meter id's base unit and description in the metrics endpoint's output
Spring Boot Admin takes the various time gauges from the metrics endpoint (e.g. `process.uptime`) . Unfortunately the time unit is  not given in the actuator output and can vary depending on the underlying micrometer registry.
Therefore it would be great if either the time unit is fixed in the metrics endpoint or included in the output. Currently the time unit can only be guessed.

joshiste:
closes #13791


joshiste:
@wilkinsona I did the requested changes.

PR #13817, opened 2018-07-18
Anonymous exception leads to empty exception tags in WebMvc/WebFlux metrics

jkschneider:
Empty tag values are rejected by some monitoring systems.

See https://github.com/micrometer-metrics/micrometer/issues/680.


snicoll:
Closed by  ee37dc1 (with the wrong issue id, sorry)

PR #13818, opened 2018-07-19
Add null guards for getBeanDefinition() in BeanTypeRegistry

izeye:
This PR adds `null` guards for `getBeanDefinition()` in `BeanTypeRegistry` as it could cause NPEs. It looks missed in 6dc14af92dbfa7b5281072d9fdde241ce0da3679.


wilkinsona:
Thanks very much, @izeye!

PR #13819, opened 2018-07-19
Upgrade to Micrometer 1.0.6

### LINKED ISSUE #13746: Upgrade to Micrometer 1.0.6
Target release date of July 12.

jkschneider:
Resolves #13746.

A couple notes:

1. Authorization key access has shifted to be lazily evaluated on each publish internally in push-based registries that require them. The purpose of this was to allow users to rotate keys at runtime.

2. This PR contains two commits, the second of which adds 3 new properties to the configuration of InfluxDB. While I'd typically reserve additions like this for a minor release, I think that these properties were an "error of omission" earlier. We've gotten feedback that retention policy is a critical thing to configure on any production Influx database. If you decide not to allow this in 2.0.4, please cherry pick to the master branch or I can open a separate PR.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/influx/InfluxProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/influx/InfluxProperties.java
@@ -58,6 +58,21 @@ public class InfluxProperties extends StepRegistryProperties {
 	private String retentionPolicy;
 
 	/**
+	 * Time period for which influx should retain data in the current database (e.g. 2h, 52w)
+	 */
+	private String retentionDuration;
+
+	/**
+	 * How many copies of the data are stored in the cluster. Must be 1 for a single node instance.
+	 */
+	private Integer retentionReplicationFactor;
+
+	/**
+	 * The time range covered by a shard group (e.g. 2h, 52w).
+	 */
+	private String retentionShardDuration;
+
+	/**
 	 * URI of the Influx server.
 	 */
 	private String uri = "http://localhost:8086";
@@ -137,4 +152,27 @@ public class InfluxProperties extends StepRegistryProperties {
 		this.autoCreateDb = autoCreateDb;
 	}
 
+	public String getRetentionDuration() {
+		return retentionDuration;
+	}
+
+	public void setRetentionDuration(String retentionDuration) {
+		this.retentionDuration = retentionDuration;
+	}
+
+	public Integer getRetentionReplicationFactor() {
+		return retentionReplicationFactor;
+	}
+
+	public void setRetentionReplicationFactor(Integer retentionReplicationFactor) {
+		this.retentionReplicationFactor = retentionReplicationFactor;
+	}
+
+	public String getRetentionShardDuration() {
+		return retentionShardDuration;
+	}
+
+	public void setRetentionShardDuration(String retentionShardDuration) {
+		this.retentionShardDuration = retentionShardDuration;
+	}
 }
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/influx/InfluxPropertiesConfigAdapter.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/influx/InfluxPropertiesConfigAdapter.java
@@ -61,6 +61,24 @@ class InfluxPropertiesConfigAdapter extends
 	}
 
 	@Override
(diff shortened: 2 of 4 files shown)


wilkinsona:
Thanks, Jon. I guess it’s too late now, but the new duration properties would have been more consistent with the rest of Boot if they were a `Duration`. It would ensure consistent handling of mapping a `String` to a `Duration` across all of Boot’s properties. Could they perhaps be overloaded?

snicoll:
> (e.g. 2h, 52w)

@jkschneider do I understand that you've replicated Spring Boot's `Duration` format? It is important if we want to expose `Duration` on our side as we need to give a `String` that would work. Can you explain a bit more how that format works?

izeye:
@snicoll Looking at [the code](https://github.com/micrometer-metrics/micrometer/blob/v1.0.6/implementations/micrometer-registry-influx/src/main/java/io/micrometer/influx/InfluxMeterRegistry.java#L106), it's for [the Influx durations](https://docs.influxdata.com/influxdb/v1.6/query_language/spec/#durations) which are different from [the Spring Boot ones](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#boot-features-external-config-conversion-duration).

snicoll:
Thanks Johnny!

jkschneider:
@snicoll Like @izeye said, this is a value that we pass to influx raw, not converting it to a `java.time.Duration` in the intervening period. We didn't choose to validate the input at this time. While Spring has `Duration` deserialization in the form of a `Converter`, I'm not aware of a serialization mechanism for the same.

jkschneider:
@wilkinsona In `InfluxProperties` we can make them `Duration` if you'd like, but then we'll have to write serialization logic to stuff them back into strings in `InfluxPropertiesConfigAdapter`. Ultimately, we have to ship a string on to Influx.

wilkinsona:
Thanks, @jkschneider. I'm torn on this now. The benefit of a `String` -> `Duration` -> `String` round-trip would be that we'd get an early failure if the user provides a value that's malformed. If we just pass the `String` straight through the failure won't occur till the call to Influx is made. The downside is that if a user if familiar with Influx, they may instinctively expect to be able to use Influx's duration format and, as @izeye notes above, it is subtly different to Boot's. On balance, I'm leaning towards passing a `String` straight through.

izeye:
How about relaxing `String` to `Duration` conversion rules in Spring Boot to be able to accommodate the Influx durations if they are reasonable?

snicoll:
Isn't that breaking the harmonization we've made for `Duration`? I know this one is subtle as it's reproducing a similar yet different format but I think I'd prefer us to be consistent. I expect us to drop the `duration` part of the property name as we've done in other places.

wilkinsona:
> Isn't that breaking the harmonization we've made for Duration?

No, I don't think it is.

In all the other places where we use `Duration`, the value that's passed into the component we're configuring is, I believe, a number. Using `Duration` provides flexibility around the number's unit. In this case, the value that's passed into the component we're configuring is a `String`.

> I expect us to drop the duration part of the property name as we've done in other places.

That would leave us with properties named `retention` and `retentionShard`. I don't think they make sense. `DURATION` and `SHARD DURATION` are clauses in the [`CREATE DATABASE`](https://docs.influxdata.com/influxdb/v1.6/query_language/database_management/#create-database) command. Combined with some other clauses they create a retention policy for the database. I think the current property names reflect what they do accurately and concisely.

wilkinsona:
> How about relaxing String to Duration conversion rules in Spring Boot to be able to accommodate the Influx durations if they are reasonable?

Unfortunately, I don't think `u or µ` is reasonable as it just means micro with no unit. They have `ns` for nanoseconds and `ms` for milliseconds which makes `u or µ` all the more confusing as it looks like it's millionths of anything.

snicoll:
One more argument to put on the balance for this is tooling support (`Duration` is or will be supported whereas this particular String needs a parser on its own which probably means no support at all). I guess it's no surprise if that's the main reason I am pushing for this.

Perhaps I am also expecting Spring Boot users to be more at ease with what they know in Spring Boot vs. the specificities of influx. If I am right, that should feel weird that there is a well-supported and documented format for durations in general but this one particular property behaves differently.

wilkinsona:
I think the key thing here is that this isn't a `java.time.Duration` but an [InfluxDB duration](
https://docs.influxdata.com/influxdb/v1.6/query_language/spec/#durations). It's unfortunate that the term is overloaded, but they are separate things and I now think we should treat them as such.

Given that we have no control over the format of an InfluxDB duration, and there's no guarantee that it won't diverge further from a `java.time.Duration`, I've come to the conclusion that it would be a mistake to use `java.time.Duration` as the type of the property. Doing so would back us into a corner and may make it impossible for a user to specify a value that's perfectly legal in InfluxDB but that we would not want to or would not be able to support in our `String` to `java.time.Duration` converter.

snicoll:
Thanks for the feedback. I agree this is the proper course of action.

PR #13835, opened 2018-07-20
polish sample-data-cassandra

nosan:


wilkinsona:
Thanks for the PR, but I don't think your Embedded Cassandra project is sufficiently well-established for us to start using it in one of our samples. Good luck with the project, though. We can reconsider this when it's more widely adopted by people using Cassandra on the JVM.

Issue #13862, opened 2018-07-22
FlywayAutoConfiguration filesystem: prefix fails


andyrbell:
Flyway locations can be prefixed with "classpath: or "filesystem:".

FlywayAutoConfiguration by default checks at least one location exists.

That check fails when using the "filesystem:" prefix due to:

    java.net.MalformedURLException: unknown protocol: filesystem

which results in the following:

    Caused by: java.lang.IllegalStateException: Cannot find migrations location in: [filesystem:src/main/resources/db/migration] (please add migrations or check your Flyway configuration)
        at org.springframework.util.Assert.state(Assert.java:94)
        at org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration$FlywayConfiguration.checkLocationExists(FlywayAutoConfiguration.java:168)
        at org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration$FlywayConfiguration.flyway(FlywayAutoConfiguration.java:152)
        at org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration$FlywayConfiguration$$EnhancerBySpringCGLIB$$65b82c95.CGLIB$flyway$1(<generated>)
        at org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration$FlywayConfiguration$$EnhancerBySpringCGLIB$$65b82c95$$FastClassBySpringCGLIB$$bf4a67a0.invoke(<generated>)
        at org.springframework.cglib.proxy.MethodProxy.invokeSuper(MethodProxy.java:228)
        at org.springframework.context.annotation.ConfigurationClassEnhancer$BeanMethodInterceptor.intercept(ConfigurationClassEnhancer.java:361)
        at org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration$FlywayConfiguration$$EnhancerBySpringCGLIB$$65b82c95.flyway(<generated>)
        at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
        at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
        at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
        at java.lang.reflect.Method.invoke(Method.java:498)
        at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:154)
        ... 70 more

[...]

PR #13863, opened 2018-07-22
Fix Flyway filesystem prefix location check

andyrbell:
Allows a Flyway location with a filesystem prefix to be validated.

Fixes gh-13862

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfigurationTests.java
@@ -189,7 +189,7 @@ public class FlywayAutoConfigurationTests {
 	@Test
 	public void changeLogDoesNotExist() {
 		this.contextRunner.withUserConfiguration(EmbeddedDataSourceConfiguration.class)
-				.withPropertyValues("spring.flyway.locations:file:no-such-dir")
+				.withPropertyValues("spring.flyway.locations:filesystem:no-such-dir")
 				.run((context) -> {
 					assertThat(context).hasFailed();
 					assertThat(context).getFailure()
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfigurationTests.java
```diff
@@ -189,7 +189,7 @@ public void overrideSchemas() {
 	@Test
 	public void changeLogDoesNotExist() {
 		this.contextRunner.withUserConfiguration(EmbeddedDataSourceConfiguration.class)
-				.withPropertyValues("spring.flyway.locations:file:no-such-dir")
+				.withPropertyValues("spring.flyway.locations:filesystem:no-such-dir")
```
snicoll: Can we keep a test that uses `file`?

andyrbell: I guess I didn't feel like there was any value in testing a `file:` location, since it would just fail in Flyway anyway because of the unknown prefix, regardless of whether it existed or not.

Happy to add it back in if you want though..?

andyrbell:
@snicoll, I have updated the documentation with an example showing the valid prefixes. Is that ok? Anything else you would like me to add?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfigurationTests.java
snicoll: Spring users know about `file` and `classpath` so it would be nice to offer that as a convenience. But I guess the swapping from `file` to `filesystem` starts to be a bit funky so let's keep it as is for now. Thanks.

andyrbell: Agreed. I did consider swapping `file` to `filesystem` but it felt a bit dirty. And I think the error from Flyway in this case is pretty clear anyway:

    Unknown prefix for location (should be either filesystem: or classpath:): file:src/main/resources/db/migration

snicoll:
@andyrbell thank you very much for your first contribution to Spring Boot. I had to rewrite your proposal in `1.5.x` and tried to [give you the contribution](https://help.github.com/articles/creating-a-commit-with-multiple-authors/) but I managed to screw it up, sorry about that.

andyrbell:
Never mind, maybe next time. :smile:

emeentag:
This path below is worked for me:
```    check-location: true
    locations:
    - classpath:./db/migration```
`./db` dot is needed in here.

Issue #13860, opened 2018-07-20
Support authenticationMethod property for oauth2-client ClientRegistration


jgrandja:
A new property has been added to `ClientRegistration.ProviderDetails.UserInfoEndpoint.authenticationMethod`, which is the authentication method used when requesting the `UserInfo` resource. The valid values are `header`, `form`, and `query` - see `AuthenticationMethod`.

The recommended Boot property -> `spring.security.oauth2.client.provider.[providerId].user-info-authentication-method`

mbhave:
Closing in favor of PR #13865.

PR #13865, opened 2018-07-23
Add user-info-authentication-method property for OAuth2-client ClientRegistration

mhyeon-lee:
gh-13860


philwebb:
Thanks for the PR, we're reconsidering the design of those properties so we might end up using different names depending on that outcome. I'll leave this one open, but we might need to revisit it before we can merge.

mbhave:
We are not going to restructure properties under `provider` so this PR can be merged.

mbhave:
Thanks @mhyeon-lee. This change is now on master.

PR #13868, opened 2018-07-23
Upgrade classmate version to 1.4.0

Hanope:
The classmate dependency version was not reflected for a long time (1.3.4 -> 1.4.0) so I updated it.

I tried contribution for the first time, but please point out any mistakes.


wilkinsona:
Thanks for the PR, but the version of Classmate is deliberately kept in alignment with the version used by Hibernate and Hibernate Validator. Both currently use 1.3.4.

PR #13901, opened 2018-07-25
upgrade to Spring Session Apple-SR4

ayudovin:
As far as I understand just version should be updated 
this [dependency-upgrade](https://github.com/spring-projects/spring-boot/issues/13888)


mbhave:
@ayudovin Thanks for the PR but we have a semi-automated process for upgrading dependencies so we don't need a PR for it.

Issue #13348, opened 2018-06-04
Perform failure analysis when a unique CacheManager bean cannot be found


iNikem:
My Spring Boot application failed today:

```
2018-06-04 08:38:16.310  INFO 21232 --- [           main] ConditionEvaluationReportLoggingListener : 

Error starting ApplicationContext. To display the conditions report re-run your application with 'debug' enabled.
2018-06-04 08:38:16.321 ERROR 21232 --- [           main] o.s.boot.SpringApplication               : Application run failed

java.lang.IllegalStateException: No CacheResolver specified, and no unique bean of type CacheManager found. Mark one as primary (or give it the name 'cacheManager') or declare a specific CacheManager to use, that serves as the default one.
	at org.springframework.cache.interceptor.CacheAspectSupport.afterSingletonsInstantiated(CacheAspectSupport.java:194)
	at org.springframework.beans.factory.support.DefaultListableBeanFactory.preInstantiateSingletons(DefaultListableBeanFactory.java:777)
	at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:869)
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:550)
	at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:759)
	at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:395)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:327)
	at org.springframework.boot.test.context.SpringBootContextLoader.loadContext(SpringBootContextLoader.java:139)
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContextInternal(DefaultCacheAwareContextLoaderDelegate.java:99)
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:117)
	at org.springframework.test.context.support.DefaultTestContext.getApplicationContext(DefaultTestContext.java:108)
	at org.springframework.test.context.web.ServletTestExecutionListener.setUpRequestContextIfNecessary(ServletTestExecutionListener.java:190)
	at org.springframework.test.context.web.ServletTestExecutionListener.prepareTestInstance(ServletTestExecutionListener.java:132)
	at org.springframework.test.context.TestContextManager.prepareTestInstance(TestContextManager.java:246)
[...]

PR #13916, opened 2018-07-26
NoSuchCacheManagerFailureAnalyzer

nosan:
see [gh-13348](https://github.com/spring-projects/spring-boot/issues/13348)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfiguration.java
```diff
@@ -52,8 +57,8 @@
  * Cache store can be auto-detected or specified explicitly via configuration.
  *
  * @author Stephane Nicoll
- * @since 1.3.0
  * @see EnableCaching
+ * @since 1.3.0
```
snicoll: please refrain from doing unrelated changes.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfiguration.java
```diff
@@ -75,8 +80,8 @@ public CacheManagerCustomizers cacheManagerCustomizers(
 
 	@Bean
 	public CacheManagerValidator cacheAutoConfigurationValidator(
-			CacheProperties cacheProperties, ObjectProvider<CacheManager> cacheManager) {
```
snicoll: You can do `ObjectProvider<Map<String,CacheManager>>` instead of injecting the context

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfiguration.java
```diff
@@ -134,4 +142,45 @@ public void afterPropertiesSet() {
 
 	}
 
+	/**
+	 * Exception thrown when {@link org.springframework.cache.CacheManager} implementation
+	 * are not specified.
+	 */
+	static class NoCacheManagerException extends RuntimeException {
```
snicoll: To mirror `NoSuchBeanDefinitionException`, you could call that `NoSuchCacheManagerException`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfiguration.java
```diff
@@ -134,4 +142,45 @@ public void afterPropertiesSet() {
...
+		}
+
+	}
+
+	/**
+	 * Exception thrown when multiple {@link org.springframework.cache.CacheManager}
+	 * implementations are available with no way to know which implementation should be
+	 * used.
+	 */
+	static class NonUniqueCacheManagerException extends RuntimeException {
```
snicoll: And this one could be `NoUniqueCacheManagerException` extending from the other exception (again to mirror `NoUniqueBeanDefinitionException`)

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/NoCacheManagerFailureAnalyzer.java
```diff
@@ -0,0 +1,50 @@
...
+import org.springframework.boot.diagnostics.FailureAnalysis;
+
+/**
+ * An {@link AbstractFailureAnalyzer} for
+ * {@link CacheAutoConfiguration.NoCacheManagerException}.
+ *
+ * @author Dmytro Nosan
+ * @since 2.1.0
+ */
+class NoCacheManagerFailureAnalyzer
```
snicoll: It would be nice to merge the `FailureAnalyzer` so that we only have one class (and an instance check to determine the type). This is my main motivation for having a hierarchy in the two exceptions at hand.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfigurationTests.java
```diff
@@ -1007,6 +1018,16 @@ public CacheResolver cacheResolver() {
 
 	}
 
+	@Configuration
+	static class AdditionalCacheManagerAutoConfiguration {
```
snicoll: That's not an auto-configuration and shouldn't be named like that. It shouldn't be added as an auto-config in the test either.

nosan:
@snicoll Frankly speaking, I really like these  two analyzers  `NoSuchBeanDefinitionFailureAnalyzer` and `NoUniqueBeanDefinitionFailureAnalyzer` and I'd like to reuse them as much as possible, that's why I'd prefer to extend `NoUniqueBeanDefinitionException` and `NoUniqueBeanDefinitionException` accordingly and provide `NoSuchCacheManagerFailureAnalyzer` only for a `spring.cache.type=generic`. 

Thoughts ?

nosan:
I need to revise this PR.

nosan:
@snicoll I've updated this PR. Please take a look :)
Thanks in advance.

nosan:
@snicoll 

Sorry for the long delay, I was a bit busy :(

I've updated this PR and fixed all conflicts.

snicoll:
@nosan no need to apologize, I am the one who didn't get a chance to have a deeper look at it yet.

snicoll:
Thanks for your patience @nosan. I am not keen to bring this amount of complexity for the task at hand. I've created [an issue in Framework](https://github.com/spring-projects/spring-framework/issues/22442) to improve the exception that is throw in such cases so that we can reuse the existing analyzers or very easily tune a new one for our purposes. Thanks for the PR, in any case.

nosan:
@snicoll fair enough, this PR contained a lot of changes.

PR #13939, opened 2018-07-29
Expose [a few] devtools classes as public classes

mmoayyed:
I am working on a custom configuration of devtools, supplying my own configuration and beans to experiment with restart behavior. In doing so, I learned that a number of classes, needed for the custom configuration of devtools, presently employed by `LocalDevToolsAutoConfiguration` are kept as package-private. This pull request proposes to open up these classes so as to improve code-reuse.

Spring Boot 2.0.3.RELEASE, JDK 10, embedded Tomcat 9.0.10.

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/ConditionEvaluationDeltaLoggingListener.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/ConditionEvaluationDeltaLoggingListener.java
@@ -30,7 +30,7 @@ import org.springframework.context.ApplicationListener;
  *
  * @author Andy Wilkinson
  */
-class ConditionEvaluationDeltaLoggingListener
+public class ConditionEvaluationDeltaLoggingListener
 		implements ApplicationListener<ApplicationReadyEvent> {
 
 	private final Log logger = LogFactory.getLog(getClass());
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/FileWatchingFailureHandler.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/FileWatchingFailureHandler.java
@@ -32,7 +32,7 @@ import org.springframework.boot.devtools.restart.Restarter;
  *
  * @author Phillip Webb
  */
-class FileWatchingFailureHandler implements FailureHandler {
+public class FileWatchingFailureHandler implements FailureHandler {
 
 	private final FileSystemWatcherFactory fileSystemWatcherFactory;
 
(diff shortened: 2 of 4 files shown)


philwebb:
Most of those classes aren't really designed to be public and changing the visibility won't necessarily mean that you can replace them. For example `FileWatchingFailureHandler` is created directly in `RestartConfiguration.onClassPathChanged`.

Can you provide some more background about the types of customization that you're trying to make?

mmoayyed:
Sure. Here is the long story:

I am working with a Spring Boot 2.0.3 web application, that is deployed by users and adopters as a Maven WAR Overlay. The core web application is bootiful, and can perfectly work with the likes of `bootRun` and such and it is published as a non-bootiful web application. Maven WAR overlays grab a copy of this web application and through the Spring Boot plugin for Maven or Gradle, repackage and put it to good use.

The trouble is with the `bootRun` task that is simply dysfunctional when it comes to working with Maven WAR overlays (or Gradle for that matter). Having had conversations with others in the past, I realize bootRun is simply not intended to work under such terms and furthermore, I understand that support for WAR Overlays is simply unlikely to happen or happen any time soon. However, `bootRun`'s ability to `addResources` is simply extremely advantageous in the context of working with an Overlay, specially for template modifications such as Thymeleaf. You have a web application that presents 100+ views, and using the Overlay you're given the chance to only modify the 2 views you actually care about for your deployment and it would be very helpful if you simply could modify these views *in place* having after repackaged the WAR; specially for rapid UI development this is very advantageous. 

I should note that we *somehow* through various types of sorcery had this setup *working* with Boot 1.5.x, and we simply got lucky for a while. Given Boot v2 and the advent of JDK 11, this same setup no longer works and I was trying to figure out alternatives and experiments that might prove more attractive. One playground idea was to try and see if I can make devtools to work with the application is run as an executable war. My experiments led to me design my own `DevToolsConfiguration` type of component, one thing led to another and here we are with this pull request :) The experiment did of course fail, as it was probably folly to begin with and in the grand scheme of things, this PR is likely m00t. 

What might work in the end would be to simply get the app deployed to an external container like tomcat, and set up a watch between the source directory artifacts and those deployed, and copy views, css, etc on demand when a change is detected.

wilkinsona:
Thanks for the explanation, @mmoayyed, but I'm not sure that I understand the need for the changes that you've proposed. You can use DevTools in your IDE and make updates to a remote application that's running as an executable war or jar. We have [tests](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-tests/spring-boot-integration-tests/spring-boot-devtools-tests/src/test/java/org/springframework/boot/devtools/tests/DevToolsIntegrationTests.java) that cover the jar case and, AFAIK, the war case should work fine too. If you've found a case where it doesn't work with a packaged war then that's probably a bug. If it is, we should fix the bug, but I don't think that making some classes public should be part of it.

mmoayyed:
Thank you Andy. I had in mind that perhaps a remote update may not be functional given this particular statement in the docs:

> Developer tools are automatically disabled when running a fully packaged application. If your application is launched from java -jar or if it is started from a special classloader, then it is considered a “production application”.

In fairness, I didn't spend much time reviewing the config to make this option work. It's more likely a mistake on my part rather than a bug, etc. I'll give it another shot. Feel free to close this PR and thank you for your time, all.

snicoll:
Thanks for the feedback.

PR #13941, opened 2018-07-29
Avoid annotation attribute lookup in OnBeanCondition

dreis2211:
Hi,

this PR avoids an annotation attribute lookup in `OnBeanCondition` in order to retrieve the `SearchStrategy` that seems superfluous - given that all attributes are already fetched a few lines earlier and the search strategy isn't affected by the class-to-string conversion parameter. I wonder if I miss something, so please feel free to close this PR if it turns out to be bogus.

Let me know what you think.
Cheers,
Christoph


philwebb:
Good spot, I can't see any logical reason why we didn't do that in the first place.

PR #13943, opened 2018-07-30
Polish AutoConfigurationGroup.selectImports()

izeye:
This PR polishes `AutoConfigurationGroup.selectImports()` by removing `processedExclusions` which has been built but never used later. It looks like a leftover which has been replaced with `allExclusions`.


snicoll:
> It looks like a leftover which has been replaced with allExclusions.

Not really, I just didn't read my own code properly 😓- Thanks again @izeye !

Issue #12348, opened 2018-03-05
Expose InfoContributor information via the metrics endpoint


jondavidnd1:
This is a request for an enhancement to the Spring Boot 2.0 actuator metrics endpoint.

It is common within prometheus to capture application metadata (e.g. version) as labels on a single metric (see https://www.robustperception.io/exposing-the-software-version-to-prometheus/).

I would propose creating (and exposing by default) an ApplicationInfoMetrics implementation of MeterBinder which converts all of the registered InfoContributor data to tags on a Guage named 'application.info' with a value of 1.0.  The info JSON would be flattened for the tag names:

```json
{
    "git": {
        "commit": {
            "id": "4616b7b"
        },
        "branch": "master"
    },
    "build": {
        "version": "0.0.1-SNAPSHOT",
        "artifact": "myapp",
        "name": "My Application",
        "group": "mygroup"
    }
}
```

would convert to the following prometheus metric:

```
# HELP application_info Application info
# TYPE application_info gauge
application_info{build_artifact="myapp",build_group="mygroup",build_name="My Application",build_version="0.0.1-SNAPSHOT",git_branch="master",git_commit_id="4616b7b"} 1.0
```

It would also be nice to have a configuration option for excluding/including InfoContributor fields for conversion to tags.

wilkinsona:
What you think, @jkschneider?

jkschneider:
Interesting. WDYT @checketts?

I don't imagine any of this would be autoconfigurable, do you? We provide the binder, you provide the inputs?

checketts:
I like it. It would be autoconfigurable in the same way the `ApplicationInfoMetrics` works. I believe the user has to provide the property values themselves. 

@jondavidnd1 I think initially excluding fields for use as tags won't be a problem in regards to metric cardinality. But probably should be included just incase users run into a value they don't want exposed or is problematic and needs to be stripped.  It isn't too urgent in regards to prometheus since it allow you to drop labels on ingestion via configuration.

jondavidnd1:
Glad to hear you think the idea has merit.

I would think this would be completely autoconfigurable even if not enabled by default.  Spring boot provides several InfoContributors (Git, Build, Environment) out of the box, and users can register more as beans.  Those InfoContributors can be read at startup time to configure the application_info metric.  That is how I have it implemented in my project.

[...]

PR #13946, opened 2018-07-30
Expose InfoContributor information via the metrics endpoint

tbilou:
Hi,

this PR exposes InfoContributor information via the metrics endpoint (see https://github.com/spring-projects/spring-boot/issues/12348)

Need some feedback regarding the flattenKeys recursive implementation. This is a problem that has been solved many times, but I was unable to find an existing method I could re-purpose for this case. Maybe someone can point out an existing implementation in spring boot


snicoll:
@tbilou thanks for the PR but exposing the whole content of the `info` endpoint as metrics is not something we want to do. Looking back at the original request, the focus was on git and build information and while the reporter [mentioned "all information", they really only used git and build](https://github.com/spring-projects/spring-boot/issues/12348#issuecomment-370782463).

I think we need to revisit this use case and figure out what we want to do. Perhaps an easy way to expose a `MeterFilter` with such information would be interesting. But `InfoContributor` shouldn't be the source. `BuildProperties` and `GitProperties` might. Thanks anyway

PR #13963, opened 2018-07-31
Polish

izeye:


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationImportSelector.java
```diff
@@ -412,8 +412,8 @@ public void process(AnnotationMetadata annotationMetadata,
 				DeferredImportSelector deferredImportSelector) {
 			Assert.state(
 					deferredImportSelector instanceof AutoConfigurationImportSelector,
-					String.format(
-							"AutoConfigurationImportSelector only supports %s implementations, got %s",
```
snicoll: This isn't a typo but a conscious decision. `AutoConfigurationGroup` is an internal concept while the one in the exception is the public concept. 

What do you think?

izeye: @snicoll Thanks for the clarification! When I saw it, "AutoConfigurationImportSelector only supports AutoConfigurationImportSelector implementations" sounds a bit weird and it's in `AutoConfigurationGroup.process()`, so I was trying to change it to `AutoConfigurationGroup` which sounds more natural but noticed that the class is `private` scope as you mentioned. I wasn't sure with this change at that time but with your comment, it'd be better to be reverted for now. I'll revert it soon.

izeye: @snicoll I reverted it and force-pushed.

snicoll: You got a valid point in any case. I'll reword that exception message

snicoll:
Thanks again @izeye - I've hopefully improved the message in f7032bd

Issue #13945, opened 2018-07-30
Document the need for compilation with -parameters when using @SpyBean and @Cacheable with a key expression that uses a parameter name


mdeinum:
Tested with both Spring Boot 1.5.14 and Spring Boot 2.0.3. 

When using `@SpyBean` to investigate method calls on a bean and a method is annotated with `@Cacheable` this breaks. 

```
@Service
public class OrderService {

    private final Logger logger = LoggerFactory.getLogger(getClass());

    @Cacheable(value="orders", key = "#id + #key.field1")
    public Order find(String id, PartialKey key) {

        double amount = ThreadLocalRandom.current().nextDouble(1000.00);
        Order order = new Order(id, BigDecimal.valueOf(amount));
        logger.info("Created: {}", order);
        return order;
    }
}
```

The issue is that when using a custom key, the extension/proxy generated by `@SpyBean` apparently doesn't have the needed debug information anymore. Switching from named parameters to indexed parameters makes it work again.

Exception:

```
java.lang.IllegalStateException: Failed to load ApplicationContext
	at org.springframework.test.context.cache.DefaultCacheAwareContextLoaderDelegate.loadContext(DefaultCacheAwareContextLoaderDelegate.java:124) ~[spring-test-4.3.18.RELEASE.jar:4.3.18.RELEASE]
	at org.springframework.test.context.support.DefaultTestContext.getApplicationContext(DefaultTestContext.java:83) ~[spring-test-4.3.18.RELEASE.jar:4.3.18.RELEASE]
	at org.springframework.boot.test.mock.mockito.MockitoTestExecutionListener.postProcessFields(MockitoTestExecutionListener.java:110) ~[spring-boot-test-1.5.14.RELEASE.jar:1.5.14.RELEASE]
	at org.springframework.boot.test.mock.mockito.MockitoTestExecutionListener.injectFields(MockitoTestExecutionListener.java:78) ~[spring-boot-test-1.5.14.RELEASE.jar:1.5.14.RELEASE]
	at org.springframework.boot.test.mock.mockito.MockitoTestExecutionListener.prepareTestInstance(MockitoTestExecutionListener.java:48) ~[spring-boot-test-1.5.14.RELEASE.jar:1.5.14.RELEASE]
	at org.springframework.test.context.TestContextManager.prepareTestInstance(TestContextManager.java:230) ~[spring-test-4.3.18.RELEASE.jar:4.3.18.RELEASE]
[...]

PR #13969, opened 2018-08-01
Failing test for gh-13945

mdeinum:
Failing testcase to show the issue. When Spring AOP need method
arguments to build a cache key (but probably also on other annotations
that require method arguments) the AOP part will fail.

Creating a Mockito mock results in the loss of the debug information
needed to read the correct method parameter.

issue: gh-13945

### DIFF
--- a/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/mock/mockito/SpyBeanWithAopProxyTests.java
+++ b/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/mock/mockito/SpyBeanWithAopProxyTests.java
@@ -18,6 +18,7 @@ package org.springframework.boot.test.mock.mockito;
 
 import java.lang.reflect.Method;
 import java.util.Arrays;
+import java.util.concurrent.ThreadLocalRandom;
 
 import org.junit.Test;
 import org.junit.runner.RunWith;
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/mock/mockito/SpyBeanWithAopProxyTests.java
```diff
@@ -64,6 +65,18 @@ public void verifyShouldUseProxyTarget() throws Exception {
 		verify(this.dateService, times(1)).getDate(matchesAnyBoolean());
 	}
 
+	@Test
+	public void verifyShouldUseProxyTargetWithMethodArguments() throws Exception {
+		Long d1 = this.dateService.getRandomDate(1234L);
+		Thread.sleep(200);
+		Long d2 = this.dateService.getRandomDate(1234L);
+		assertThat(d1).isEqualTo(d2);
```
wilkinsona: This assertion passes which means that the caching using `#seed` is working

spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/mock/mockito/SpyBeanWithAopProxyTests.java
```diff
@@ -64,6 +65,18 @@ public void verifyShouldUseProxyTarget() throws Exception {
...
+		verify(this.dateService, times(1)).getDate(false);
+		verify(this.dateService, times(1)).getDate(matchesFalse());
+		verify(this.dateService, times(1)).getDate(matchesAnyBoolean());
```
wilkinsona: These three verifications are verifying the wrong method. They should be checking that `getRandomDate(long)` has only been called once:

```java
verify(this.dateService, times(1)).getRandomDate(1234L);
```

With this change in place, the test passes.

wilkinsona:
Thanks for the PR, @mdeinum. Unfortunately the provided test does not appear to reproduce the problem. I'll take a look at 
https://github.com/spring-projects/spring-boot-issues/pull/79 instead.

spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/mock/mockito/SpyBeanWithAopProxyTests.java
mdeinum: Interestingly enough this failed at calling the actual method (so I didn't even get to the assertions). Will fix the assertions and see what happens.

Issue #13981, opened 2018-08-02
Spring boot doesn't include micrometer-registry-cloudwatch


cmakinen:
`micrometer-registry-cloudwatch` isn't part of spring-boot dependencyManagement section. As of now, version has to be explicitly provided if it is to be used.

Here's the dependency management section where it is missing. 
https://github.com/spring-projects/spring-boot/blob/435c47925ec80f4f98e488400d79889eeedab8e3/spring-boot-project/spring-boot-dependencies/pom.xml#L868-L921

Spring-Boot versions tested: 2.0.4.RELEASE, 2.0.3.RELEASE

snicoll:
Thanks for the report @cmakinen. FYI we don't need an issue and a PR as one duplicates the other.

Duplicate of #13981

PR #11276, opened 2017-12-06
CloudWatch metrics autoconfiguration

dkublik:
Hi,
Two weeks ago I added support for CloudWatch to micrometer.
https://github.com/micrometer-metrics/micrometer/commit/b2b13fb9fcca5ef5170cb9b0bb9ce91f93a2fe27
I thought I will PR for autoconfiguration as well :)

issue: #11275

PR #13982, opened 2018-08-02
Dependency management for micrometer-registry-cloudwatch is missing

cmakinen:
Added `micrometer-registry-cloudwatch` to spring-boot dependencyManagement section.

Refers to issue: #13981

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -99,6 +99,11 @@
 		</dependency>
 		<dependency>
 			<groupId>io.micrometer</groupId>
+			<artifactId>micrometer-registry-cloudwatch</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
+			<groupId>io.micrometer</groupId>
 			<artifactId>micrometer-registry-datadog</artifactId>
 			<optional>true</optional>
 		</dependency>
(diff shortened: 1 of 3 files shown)


snicoll:
@cmakinen thanks for the PR but before we add this to the dependency management, I'd like to understand why we don't have auto-configuration at all for this (in the current state adding the version in the two other modules is useless).

paging @jkschneider

spencergibb:
It is in spring cloud aws

snicoll:
Thanks Spencer, some [more info in this comment](https://github.com/spring-projects/spring-boot/pull/11276#issuecomment-364199576).

@cmakinen can you please rework your PR to remove the inclusion of that library in the `autoconfigure` and `docs` modules?

snicoll:
@cmakinen thank you very much for your first contribution to Spring Boot. This has been merged in `2.0.x` and `master`.

cmakinen:
Thanks @snicoll Really appreciate it!! Happy to contribute :)

jkschneider:
If we dependency manage `micrometer-registry-cloudwatch` in Spring Boot, we are effectively dependency managing the AWS SDK which iterates quite a bit faster than Spring Boot.

That’s why we left this to `spring-cloud-aws`.

snicoll:
Reopening to double check we didn’t do this on purpose

wilkinsona:
> If we dependency manage micrometer-registry-cloudwatch in Spring Boot, we are effectively dependency managing the AWS SDK

I don't think we are. The dependency management of `micrometer-registry-cloudwatch` will have no direct effect on the version of the AWS SDK. The version of the AWS SDK would only be affected if the user adds a dependency on `micrometer-registry-cloudwatch` and then, in the absence of any other dependency on the AWS SDK, they'll get whatever version of the AWS SDK is pulled in transitively. In terms of the version of the AWS SDK, this is no different to the user manually specifying the version of `micrometer-registry-cloudwatch`.

Issue #13974, opened 2018-08-01
Provide a rich type and converter to be used by properties that configure a number of bytes


wilkinsona:
We have a few places where you can use things like 1MB or 8KB and we'll parse that into the right number of bytes. We'd like to formalise this in a similar way to what we did for `Duration` so it's not just specific to one or two properties.

snicoll:
We're considering move the data type itself in Spring Framework as it could be reused in other places, see [SPR-17154](https://jira.spring.io/browse/SPR-17154) for more details.

PR #13984, opened 2018-08-02
DigitalAmount, DigitalUnit,  DigitalAmountStyle (gh-13974)

nosan:
see [gh-13974](https://github.com/spring-projects/spring-boot/issues/13974)
`DigitalAmount` -  class models a quantity or amount of digital information in terms of bytes. This class is immutable and thread-safe.
`DigitalUnit` - represents digital amount at a given unit of granularity and provides utility methods to convert across units.
`DigitalAmountStyle` -  format styles

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmount.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmount.java
@@ -22,6 +22,7 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 
 import org.springframework.lang.NonNull;
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 
 /**
@@ -35,10 +36,13 @@ import org.springframework.util.Assert;
  */
 public final class DigitalAmount implements Comparable<DigitalAmount>, Serializable {
 
-	private final long bytes;
+	/**
+	 * amount in terms of bytes.
+	 */
+	private final long amount;
 
-	private DigitalAmount(long bytes) {
-		this.bytes = bytes;
+	private DigitalAmount(long amount) {
+		this.amount = amount;
 	}
 
 	/**
@@ -46,7 +50,7 @@ public final class DigitalAmount implements Comparable<DigitalAmount>, Serializa
 	 * @return the total amount of the bytes.
 	 */
 	public long toBytes() {
-		return this.bytes;
+		return this.amount;
 	}
 
 	/**
@@ -54,7 +58,7 @@ public final class DigitalAmount implements Comparable<DigitalAmount>, Serializa
 	 * @return the total amount of the kilobytes.
 	 */
 	public long toKilobytes() {
-		return DigitalUnit.BYTES.toKilobytes(this.bytes);
+		return DigitalUnit.BYTES.toKilobytes(this.amount);
 	}
 
 	/**
@@ -62,7 +66,7 @@ public final class DigitalAmount implements Comparable<DigitalAmount>, Serializa
 	 * @return the total amount of the megabytes.
 	 */
 	public long toMegabytes() {
-		return DigitalUnit.BYTES.toMegabytes(this.bytes);
+		return DigitalUnit.BYTES.toMegabytes(this.amount);
 	}
 
 	/**
@@ -70,7 +74,7 @@ public final class DigitalAmount implements Comparable<DigitalAmount>, Serializa
 	 * @return the total amount of the gigabytes.
 	 */
 	public long toGigabytes() {
-		return DigitalUnit.BYTES.toGigabytes(this.bytes);
+		return DigitalUnit.BYTES.toGigabytes(this.amount);
 	}
 
 	/**
@@ -78,7 +82,7 @@ public final class DigitalAmount implements Comparable<DigitalAmount>, Serializa
 	 * @return the total amount of the terabytes.
 	 */
 	public long toTerabytes() {
-		return DigitalUnit.BYTES.toTerabytes(this.bytes);
+		return DigitalUnit.BYTES.toTerabytes(this.amount);
 	}
 
 	/**
@@ -86,7 +90,7 @@ public final class DigitalAmount implements Comparable<DigitalAmount>, Serializa
 	 * @return true if this {@code DigitalAmount} has a total amount less than zero
 	 */
 	public boolean isNegative() {
-		return this.bytes < 0;
+		return this.amount < 0;
 	}
 
 	/**
@@ -94,7 +98,7 @@ public final class DigitalAmount implements Comparable<DigitalAmount>, Serializa
 	 * @return true if this {@code DigitalAmount} has a total amount greater than zero
 	 */
 	public boolean isPositive() {
-		return this.bytes > 0;
+		return this.amount > 0;
 	}
 
 	/**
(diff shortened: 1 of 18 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmount.java
```diff
@@ -0,0 +1,307 @@
...
+/**
+ * This class models a quantity or amount of digital information in terms of bytes. This
+ * class is immutable and thread-safe.
+ *
+ * @author Dmytro Nosan
+ * @since 2.1.0
+ * @see DigitalAmountStyle
+ * @see DigitalUnit
+ */
+public final class DigitalAmount implements Comparable<DigitalAmount>, Serializable {
```
snicoll: Let's call that `DataSize` and move that to `org.springframework.boot.unit`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmount.java
```diff
@@ -0,0 +1,307 @@
...
+
+	/**
+	 * amount in terms of bytes.
+	 */
+	private final long amount;
```
snicoll: I'd call this `bytes`. It's not a `@ConfigurationProperties` so we don't need javadoc on that field.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmount.java
```diff
@@ -0,0 +1,307 @@
...
+	 */
+	public long toGigabytes() {
+		return DigitalUnit.BYTES.toGigabytes(this.amount);
+	}
+
+	/**
+	 * Converts this {@code DigitalAmount} to the terabytes.
+	 * @return the total amount of the terabytes.
+	 */
+	public long toTerabytes() {
```
snicoll: All those `to` can go away in benefit of a generic `public long to(Unit unit)` (more on that later).

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmount.java
```diff
@@ -0,0 +1,307 @@
...
+		return new DigitalAmount(Math.addExact(this.amount, other.amount));
+	}
+
+	/**
+	 * Returns a new {@code DigitalAmount} whose value is {@code (this -
+	 * other)}.
+	 * @param other value to be subtracted from this {@code DigitalAmount}.
+	 * @return {@code this - other}
+	 */
+	public DigitalAmount subtract(@NonNull DigitalAmount other) {
```
snicoll: All the methods above should go away please. There is really no need to start with something that complex/complete. The purpose of the enhancement is to have a data type to represent a "data size" for binding purposes.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmount.java
```diff
@@ -0,0 +1,307 @@
...
+	}
+
+	/**
+	 * Print this {@code DigitalAmount} using the given unit.
+	 * @param unit the {@code DigitalUnit} to use if the value doesn't specify,
+	 * {@link DigitalUnit#BYTES} will be used
+	 * @return the printed result
+	 * @see DigitalAmountStyle
+	 */
+	public String print(@Nullable DigitalUnit unit) {
```
snicoll: We don't use `@Nullable` in Spring Boot yet and I am not keen to introduce it at this point. 

This `print` method should go away in benefit of `toString`

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmount.java
```diff
@@ -0,0 +1,307 @@
...
+			@Nullable DigitalUnit unit) {
+		return DigitalAmountStyle.SIMPLE.parse(value, unit);
+	}
+
+	/**
+	 * Creates a new {@code DigitalAmount} from the bytes.
+	 * @param amount amount of the bytes
+	 * @return a {@code DigitalAmount}
+	 */
+	public static DigitalAmount fromBytes(long amount) {
```
snicoll: I'd use `of` to be consistent with `Duration` and have a value with a unit rather than those dedicated methods. 

The surface area of this object has to remain small given where it is actually going to be used.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalAmountStyle.java
```diff
@@ -0,0 +1,144 @@
...
+import org.springframework.util.NumberUtils;
+import org.springframework.util.StringUtils;
+
+/**
+ * {@code DigitalAmount} format styles.
+ *
+ * @author Dmytro Nosan
+ * @since 2.1.0
+ */
+public enum DigitalAmountStyle {
```
snicoll: Please move the logic of that class in a simplified version of `parse` on DurationSize` itself. There is no need to expose such open option at this point.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/DigitalAmountUnit.java
```diff
@@ -0,0 +1,45 @@
...
+ * Annotation that can be used to change the default unit used when converting a
+ * {@link org.springframework.boot.DigitalAmount}.
+ *
+ * @author Dmytro Nosan
+ * @since 2.1.0
+ */
+@Target(ElementType.FIELD)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+public @interface DigitalAmountUnit {
```
snicoll: `DataSizeUnit`

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/DigitalUnit.java
```diff
@@ -0,0 +1,498 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * A {@code DigitalUnit} represents digital amount at a given unit of granularity and
+ * provides utility methods to convert across units.
+ *
+ * @author Dmytro Nosan
+ * @since 2.1.0
+ */
+public enum DigitalUnit {
```
snicoll: This should be moved to a nested `Unit` of `DataSize`. It doesn't have to be that complicated either. We don't need to expose all the unit methods there

I had crafted something like this locally

```
public enum Unit {

        /**
         * KiloByte.
         */
        KILOBYTES((size) -> size * BYTES_PER_KB),

        /**
         * MegaByte.
         */
        MEGABYTE((size) -> size * BYTES_PER_MB);

        private final Function<Long, Long> converter;

        Unit(Function<Long, Long> converter) {
            this.converter = converter;
        }

        public long toBytes(long size) {
            return this.converter.apply(size);
        }

    }
```

nosan:
@snicoll Thanks, I am on vacation on next 2 weeks, so unfortunately I can't take a  look into it right now, but I definitely take a look.

snicoll:
@nosan thanks for your effort. We've decided to move the value object to Spring Framework (there is now a `DataSize`). This PR also has a scope that's way too large to what we envision initially. I've had a look and have now an alternative that I'll merge soon. 

Thanks anyway.

nosan:
@snicoll np, I think it's much better to have it in a `Spring Framework`.


P. S.  Typo in the javadoc :) 

```
/**
	 " (where a minute is 60 seconds)
	 * "20"   -- parses as "20 bytes"
	 * </pre>
	 * @param text the text to parse
	 * @return the parsed {@link DataSize}
	 * @see #parse(CharSequence, DataUnit)
	 */
	public static DataSize parse(CharSequence text) {
		return parse(text, null);
	}

```

PR #13989, opened 2018-08-02
Update assertion message in AnnotatedClassFinder.scanPackage()

izeye:
This PR updates assertion message in `AnnotatedClassFinder.scanPackage()` to align with the change made in df6feb3e2a796e09997cc6660a787b741d719c70.


snicoll:
Good catch again @izeye, that's a polish of the polish of the PR :)

Issue #13935, opened 2018-07-28
ZipInflaterInputStream  native memory leak


zhujibing:
`ZipInflaterInputStream` constructor function will create Inflater object where use native memory，but Memory is not freed。

wilkinsona:
We can't set `usesDefaultInflater` as it's a package-private field. We also aren't using the default `Inflater`; we're creating one with `new Inflater(true)`.

> wo should not depend on gc. native memory is hard to release , because Inflater enter old generation.

The GC generation of the object should not make any difference. AFAIK, its memory (and any associated native memory) will be freed when it is eligible for garbage collection.

Rather than describing a theoretical problem, perhaps you can provide a small sample that shows a failure due to memory being leaked?

izeye:
I guess @zhujibing tried to say a general rule that releasing resources explicitly whenever possible is better than resorting to GC, so I just created #14001 to resolve the concern.

wilkinsona:
Closing in favour of #14001.

PR #14001, opened 2018-08-06
Invoke Inflater.end() in ZipInflaterInputStream.close()

izeye:
This PR changes to invoke `Inflater.end()` in `ZipInflaterInputStream.close()` as it looks reasonable to do so likewise `InflaterInputStream.close()` as follows:

```java
    public void close() throws IOException {
        if (!closed) {
            if (usesDefaultInflater)
                inf.end();
            in.close();
            closed = true;
        }
    }
```

Closes gh-13935

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/ZipInflaterInputStream.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/ZipInflaterInputStream.java
@@ -30,12 +30,19 @@ import java.util.zip.InflaterInputStream;
  */
 class ZipInflaterInputStream extends InflaterInputStream {
 
+	private final Inflater inflater;
+
 	private boolean extraBytesWritten;
 
 	private int available;
 
 	ZipInflaterInputStream(InputStream inputStream, int size) {
-		super(inputStream, new Inflater(true), getInflaterBufferSize(size));
+		this(inputStream, new Inflater(true), size);
+	}
+
+	private ZipInflaterInputStream(InputStream inputStream, Inflater inflater, int size) {
+		super(inputStream, inflater, getInflaterBufferSize(size));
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/JarFileEntries.java
```diff
@@ -188,7 +189,8 @@ public InputStream getInputStream(FileHeader entry) throws IOException {
 		}
 		InputStream inputStream = getEntryData(entry).getInputStream();
 		if (entry.getMethod() == ZipEntry.DEFLATED) {
-			inputStream = new ZipInflaterInputStream(inputStream, (int) entry.getSize());
+			inputStream = new ZipInflaterInputStream(inputStream, new Inflater(true),
```
wilkinsona: I don't think we want to pass in an `Inflater` and than have `ZipInflaterInputStream` call `end()` on it.

spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/ZipInflaterInputStream.java
```diff
@@ -30,12 +30,15 @@
  */
 class ZipInflaterInputStream extends InflaterInputStream {
 
+	private final Inflater inflater;
+
 	private boolean extraBytesWritten;
 
 	private int available;
 
-	ZipInflaterInputStream(InputStream inputStream, int size) {
```
wilkinsona: I think we should keep this constructor and then call the new constructor proposed below with `this(inputStream, new Inflater(true), size)`;

izeye:
@wilkinsona Thanks for the advice! I changed as requested.

wilkinsona:
Thanks very much, @izeye.

PR #14002, opened 2018-08-06
Add table for supported build tools

### LINKED ISSUE #12097: Need compatibility chart for Spring Boot Gradle plugin with Gradle version
Gradle 4 introduces a completely new model of dependency management, as documented [here](https://docs.gradle.org/current/userguide/java_library_plugin.html#sec:java_library_configurations_graph). It appears that Spring Boot Gradle plugin 1.5.3 doesn't correctly work with builds using the new Gradle configurations. Even though a dependency is specified in `implementation`, it needs to be duplicated in `runtime`, otherwise we get a `ClassNotFoundException` at runtime when running the uber jar. Also, using `runtimeOnly` instead of `runtime` causes similar `ClassNotFoundException`.

This ticket is to have a compatibility chart in the documentation showing what version of the Boot plugin is supposed to work with what versions of Gradle. Without that, it seems like a c*apshoot, where the plugin may or may not work depending on the project configuration.

qct:


snicoll:
@qct thank you so much for your first contribution to Spring Boot. This is now merged in `1.5.x`, `2.0.x` and `master`.

PR #14003, opened 2018-08-06
Sync requiment of maven version with enforce rule, add a table

qct:
it's a correction to b75e58b70f416bd5e7cea4f26349bd2e43f2d20b, sync docs with enforce rule, and add a table.


wilkinsona:
Thanks for the PR, but it's only Spring Boot's own build that requires Maven 3.5+. Apps using Spring Boot are free to use Maven 3.2+. If #14002 is accepted, we'll pick up the proposed table when it is merged forwards.

PR #14015, opened 2018-08-07
Polish OnWebApplicationCondition

dreis2211:
Hi,

this PR avoids a superfluous ConditionMessage being created when the type in ConditionalOnWebApplication is explicitly set to REACTIVE or SERVLET. That made it also possible to further polish the respective method.

Let me know what you think.
Cheers,
Christoph


snicoll:
Thanks again @dreis2211 !

PR #14021, opened 2018-08-08
Kafka Streams AutoConfiguration

garyrussell:
If `kafka-streams` is on the classpath, auto configure Kafka Streams
`Properties`, set `@EnableKafkaStreams` and disable the factory
bean's `autoStartup` property.


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
```diff
@@ -138,4 +145,38 @@ public KafkaAdmin kafkaAdmin() {
 		return kafkaAdmin;
 	}
 
+	@Configuration
+	@ConditionalOnClass(StreamsBuilder.class)
+	public static class KafkaStreamsConfiguration {
+
+		@Bean(KafkaStreamsDefaultConfiguration.DEFAULT_STREAMS_CONFIG_BEAN_NAME)
+		public Properties defaultKafkaStreamsConfig(KafkaProperties properties) {
```
snicoll: I am not keen to expose a `@Bean` of type `Properties`, perhaps we could have a dedicated object for this or does the underlying configuration actually require this?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
```diff
@@ -138,4 +145,38 @@ public KafkaAdmin kafkaAdmin() {
...
+					// trim [...] - revert to comma-delimited list
+					value = value.substring(1, value.length() - 1);
+				}
+				streamsConfig.put(k, value);
+			});
+			return streamsConfig;
+		}
+
+		@Bean
+		public Object kafkaStreamsFactoryBeanConfigurer(
```
snicoll: `Object` ?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -31,6 +31,7 @@
 import org.apache.kafka.common.config.SslConfigs;
 import org.apache.kafka.common.serialization.StringDeserializer;
 import org.apache.kafka.common.serialization.StringSerializer;
+import org.apache.kafka.streams.StreamsConfig;
```
snicoll: That requires Kafka streams to be on the classpath and, as I understand, this is an optional dependency on top of Kafka

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -631,6 +651,137 @@ public void setFailFast(boolean failFast) {
...
+	 * High (and some medium) priority Streams properties and a general properties bucket.
+	 */
+	public static class Streams {
+
+		private final Ssl ssl = new Ssl();
+
+		/**
+		 * Kafka streams application.id property; default spring.application.name.
+		 */
+		private String applicationId = "${spring.application.name}";
```
snicoll: Is that supposed to be SpEL? I'd prefer if we perform that detection elsewhere (in the caller with access to the `Environment`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -631,6 +651,137 @@ public void setFailFast(boolean failFast) {
...
+
+		/**
+		 * Whether or not to auto-start the streams factory bean; default false.
```
snicoll: We don't add default values to descriptions.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -631,6 +651,137 @@ public void setFailFast(boolean failFast) {
...
+		 */
+		private boolean autoStartup;
+
+		/**
+		 * Comma-delimited list of host:port pairs to use for establishing the initial
+		 * connection to the Kafka cluster.
+		 */
+		private List<String> bootstrapServers;
```
snicoll: I am confused. Is that streams specific? Yet this property doesn't indicate that's the case.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -631,6 +651,137 @@ public void setFailFast(boolean failFast) {
...
+		 */
+		private Integer cacheMaxBytesBuffering;
+
+		/**
+		 * ID to pass to the server when making requests. Used for server-side logging.
+		 */
+		private String clientId;
+
+		/**
+		 * The replication factor for change log topics and repartition topics created by
```
snicoll: Key descriptions do not start with "The", "A", etc.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -5587,6 +5587,39 @@ The following component creates a listener endpoint on the `someTopic` topic:
 	}
 ----
 
+[[boot-deatures-kafka-streams]]
+==== Kafka Streams
+
+Spring for Apache Kafka provides a factory bean to create a `StreamsBuilder` object and
+manage the lifecycle of its streams; the factory bean is created when
+`@EnableKafkaStreams` is present on a `@Configuration` class.
```
snicoll: I can see the auto-configuration enables that so something is not consistent. Perhaps the auto-configuration should check for the presence of `@EnableKafkaStreams` if that makes sense? (We don't do that for Kafka so perhaps that's the wrong idea).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
garyrussell: Kafka deprecated the use of an immutable `StreamsConfig` in favor of a `Properties` object to configure the `StreamsBuilder`. I can add a wrapper class in spring-kafka if you prefer. 

https://jira.apache.org/jira/browse/KAFKA-6386

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
garyrussell: I used a `NullBean` (return null) to avoid creating an explicit `...Configurer` class just for this one property; I can do that if you prefer.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
garyrussell: I used a property placeholder which gets resolved as expected in the test case; I can move it to the caller.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
garyrussell: Since autoStartup is usually true by default; I thought it was important to add it here. The reason it's false is I don't want to start the factory bean for an existing Boot app that uses streams but doesn't use the auto configured infrastructure (start connects to Kafka).

I did mention it in the docs, however, so I can remove it here.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
garyrussell: This is similar to producer/consumer - we can add a global property and/or override at the component level - I will enhance these docs in all 3 places.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
garyrussell: OK; I simply copied the Kafka descriptions 😄

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
garyrussell: Oops

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
garyrussell: My mistake; I originally did not add `@EnableKafkaStreams` in the auto config because of the problem with auto-starting the factory bean.

I later added the autoStartup auto-config to prevent the FB from starting unless the user explicitly sets it, but failed to revisit the docs.

Which do you prefer - requiring a user-specified `@EnableKafkaStreams` or this autoStartup dance? I think I would prefer that the user adds the annotation, but that seems inconsistent with other practices.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
snicoll: Yes please, `Properties` is way to generic to be exposed as a bean.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
snicoll: I am not I understand what that means but I'd rather have something that backs off with a property type if we should do something.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
snicoll: We don't do property placeholder in properties. Yes please.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
snicoll: The point I was trying to make is that the metadata will already state the default is `false` (as inferred by the default value of the field). So your IDE will tell you it's false and having that in text is a duplicate that could potentially get out of sync.

Does that make sense?

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
snicoll: I don't know really. I'd like this to be consistent. On the other hand, I don't know if adding `kafka-streams` is enough to trigger that behaviour. Since adding `spring-kafka` is enough for the kafka use case, I'd be tempted to say that's the same here (you need `spring-kafka` and `kafka-streams` anyway).

Any potential clash with Spring Cloud Stream and the kafka streams story there?

snicoll:
> kafka-streams is already in the pom:

Wow I totally missed that. It means we can remove our hard-coded version in initializr, sweet! 

Thanks for the feedback, let me know if you need help with any of this.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
garyrussell: I'll stick with the presence of the jar for consistency (and fix the configurer).

I will reach out to the SCSt team to make them aware of this.

garyrussell:
I pushed a PR for spring-kafka to avoid the `Properties` bean; I'll address your other comments when that is merged.

Thanks.

garyrussell:
@snicoll I addressed your concerns.

Please note that I am on PTO this week; if this needs more polishing, please reach out to @artembilan ; I will make him a collaborator on my fork so he can push to this PR.

Thanks, Gary

snicoll:
@garyrussell @artembilan I've changed my mind looking at the current proposal, please see 
1d113de for more details.

Given that `application.id` is mandatory, I feel that the current proposal is missing some context: let's assume you upgrade and you happen to have `kafka-streams` on the classpath and no application id (or `spring.application.name`). We auto-configure things automatically, configure a bean without `application.id` (invalid) and not start anything so "all is fine".

Now if you enable auto-startup, you have an exception deep down that tells you application.id is not set. The link to Spring Boot is not super obvious. I am not a huge fan of overriding the default value of `auto-startup`.

I've modified the initial proposal so that it requires `@EnableKafkaStreams` and we fail hard if the application id is not set (using a failure analyzer that tells you exactly what happened). The auto-startup flag has also been restored to its default value.

Concretely, nothing happens until `@EnableKafkaStreams` is set. When it's set, the configuration is validated upfront.

What do you think about this proposal? Does that make sense? Thanks!

garyrussell:
@snicoll LGTM - that was my original approach (user opt-in with `@EnableKafkaStreams`).

Since I thought that was inconsistent with other auto-config (adding `@Enable...` automatically), I went to the auto-startup dance.

I definitely prefer this solution, though.

Thanks!

snicoll:
Sweet, thanks for the feedback @garyrussell!

PR #14024, opened 2018-08-09
Polish ItemMetadata.buildName()

izeye:
This PR polishes `ItemMetadata.buildName()`.

Along the way, I noticed the current `prefix` handling which removes all trailing dots looks unusual, so I changed it to remove the last one only. I just judged it's not intentional by passing all tests but if it's intentional, please let me know. I'll revert it.


snicoll:
> Along the way, I noticed the current prefix handling which removes all trailing dots looks unusual

It is a bit silly. I think the intent was to be as lenient as possible in case of trailing dots but I guess we can give that a try in 2.1 and see if someone grumbles.

PR #13966, opened 2018-07-31
Make max swallow size easily configurable

ayudovin:
Now "maxSwallowSize" could be configurable via follow property: `spring.tomcat.max-swallow-size` 
this [enhancement](https://github.com/spring-projects/spring-boot/issues/6669)

Issue #6669, opened 2016-08-17
Make "MaxSwallowSize" more easily configurable


ghillert:
This is relevant for larger file-uploads and is hard to trace and may cause errors in your browser like `The connection was reset.`

The following StackOverflow post provides a solution:

http://stackoverflow.com/questions/35748022/multipart-file-maximum-size-exception-spring-boot-embbeded-tomcat

It would be good if Spring Boot would support this out of the box using:

`spring.http.multipart.max-swallow-size`

Ideally, the Boot `MultipartProperties` should also be smart enough so that a `max-file-size` or `max-request-size` that is larger than the `max-swallow-size` will either throw an error/warning or implicitly increase the `max-swallow-size` (Not sure if there are edge-cases against it, though.)

Also,  document the issue in the reference guide.

wilkinsona:
The swallow size isn't specific to file uploads. It's for any request with a large body. From the Tomcat docs:

> The maximum number of request body bytes (excluding transfer encoding overhead) that will be swallowed by Tomcat for an aborted upload. An aborted upload is when Tomcat knows that the request body is going to be ignored but the client still sends it. If Tomcat does not swallow the body the client is unlikely to see the response. If not specified the default of 2097152 (2 megabytes) will be used. A value of less than zero indicates that no limit should be enforced.

As such, it doesn't make sense for it to be a multipart property

Note that it's also only for an aborted upload so I'm not sure that it necessarily makes sense to enforce any relationship between the swallow size and the max file/request sizes for multipart uploads.

philwebb:
We could add this under `server.tomcat.` but it's a balance of how many of those we should add. You can do this with a `TomcatConnectorCustomizer` already.

snicoll:
Closing in favour of PR #13966

PR #14025, opened 2018-08-09
Align server.tomcat.max-swallow-size with Tomcat's default value

izeye:
This PR makes `server.tomcat.max-swallow-size` align with Tomcat's default value as I couldn't find any good reason to diverge from Tomcat's default value in the related issue (#6669), PR (#13966) and commits (0d40c5aeccacb3123d3d5b4eb4391031cddb4adf, a8b9718073ec997e3cd4f0a696ed65a4af33364a). If there's any good reason to diverge which I missed, feel free to close this PR.

See https://tomcat.apache.org/tomcat-9.0-doc/config/http.html


snicoll:
> If there's any good reason to diverge which I missed, feel free to close this PR.

Oh there is a perfectly good reason, my maths is foobar 😓-  I read `2 * 1024 * 1024` and thought I was very smart to use `4096` 🙄

Thank you again Johnny !

PR #14029, opened 2018-08-09
fixed javadoc

hengyunabc:
fixed javadoc.

The links do not take effect: https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/web/servlet/ServletComponentScan.html

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/ServletComponentScan.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/ServletComponentScan.java
@@ -22,17 +22,14 @@ import java.lang.annotation.Retention;
 import java.lang.annotation.RetentionPolicy;
 import java.lang.annotation.Target;
 
-import javax.servlet.annotation.WebFilter;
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR. I've built your changes locally and, unfortunately, I don't think they've fixed the problem. AFAIK, the use of `import …` vs referring to the fully-qualified classname in the javadoc is just a convenience and should result in the same javadoc. That was certainly the case when I built your change.

We've configured javadoc with [a link to the Java EE API](https://github.com/spring-projects/spring-boot/blob/295995829f41fae715b7a2f2195e4d14f2f96b6d/spring-boot-project/spring-boot-docs/pom.xml#L951) which is what I thought should make this work. We'll need to investigate why that's not the case.

wilkinsona:
The problem doesn't just affect `ServletComponentScan`. For example, `ServletRegistrationBean` is affected too. I've opened #14031 to track the problem. Thank you for making us aware of it. I'll close this one as it doesn't appear to fix the problem with `ServletComponentScan` and we need to find a broader solution anyway.

PR #14032, opened 2018-08-09
URL's retryWrites query parameter is ignored when auto-configuring a reactive Mongo client

GreyTeardrop:
Add `retryWrites` flag to set of flags propagated from `connectionString`
to `ReactiveMongoClient`.

Regular `MongoClientFactory` relies of `MongoClientURI` class that
propagates all flags in its `getOptions()` method.


GreyTeardrop:
The issue this PR attempts to fix affects Boot 2.0 and master branches. Should I target `2.0.x` branch instead?

wilkinsona:
Yes, please. It looks like a bug to me so should be fixed in 2.0.x.

GreyTeardrop:
I've rebased commit on top of 2.0.x. Not sure how to trigger CI again though.

wilkinsona:
Thanks very much, @GreyTeardrop. The proposed changes are now in both 2.0.x and master.

PR #14035, opened 2018-08-10
Add "With an Immediate Pull Request"

izeye:
I have seen many "an issue with an immediate pull request" patterns like [this one](https://github.com/spring-projects/spring-boot/issues/13981#issuecomment-409913552) but the pattern is considered as a duplicate in the Spring Boot repository, so I added "With an Immediate Pull Request" note to the issue template.

This PR also reorganises with "Please DO NOT Raise an Issue" Cases.

### DIFF
--- a/.github/ISSUE_TEMPLATE.md
+++ b/.github/ISSUE_TEMPLATE.md
@@ -2,9 +2,16 @@
 Thanks for raising a Spring Boot issue. Please take the time to review the following
 categories as some of them do not apply here.
 
-❓Question
+🙅 "Please DO NOT Raise an Issue" Cases
+- Question
 STOP!! Please ask questions about how to use something, or to understand why something isn't
 working as you expect it to, on Stack Overflow using the spring-boot tag.
+- Security Vulnerability
+STOP!! Please don't raise security vulnerabilities here. Head over to https://pivotal.io/security to learn how to disclose them responsibly.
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE.md
```diff
@@ -12,13 +12,15 @@ are using. If possible, please provide a test case or sample application that re
...
-🚨 Security Vulnerability
-STOP!! Please don't raise security vulnerabilities here. Head over to https://pivotal.io/security to learn how to disclose them responsibly.
-
 🎁 Enhancement
 Please start by describing the problem that you are trying to solve. There may already
 be a solution, or there may be a way to solve it that you hadn't considered.
 
-🙅 Managed dependency upgrade
-You DO NOT need to raise an issue for a managed dependency version upgrade as there's a semi-automatic process for checking managed dependencies for new versions before a release. But pull requests for upgrades that are more involved than just a version property change are still most welcome.
+🙅 "Please DO NOT Raise an Issue" Cases
```
snicoll: Shouldn't question be in this category? If you agree, I wonder if that shouldn't be on top (let's increase our change that users actually read that).

izeye:
@snicoll Thanks for the quick feedback! I agree all your concerns including "TL;DR". I updated as you suggested.

PR #14059, opened 2018-08-14
Use the first "?" for a query part in Handler.setFile()

izeye:
This PR changes to use the first "?" for a query part in `Handler.setFile()` as the first "?" is the starting character for a query part in a URL.


wilkinsona:
Thanks, @izeye, but the goal was to align as closely as possible with the JDK and [it uses `lastIndexOf`](http://hg.openjdk.java.net/jdk8u/jdk8u/jdk/file/6bfaecb8ff77/src/share/classes/java/net/URLStreamHandler.java#l586).

Issue #14060, opened 2018-08-14
Prevent mysql-connector-java from pulling in an old version of protobuf-java


meltsufin:
Spring Boot 2.1 [upgraded to MySQL Connector/J version 8,](https://github.com/spring-projects/spring-boot/issues/13688) which [now includes `protobuf-java` 2.6.0](http://central.maven.org/maven2/mysql/mysql-connector-java/8.0.11/mysql-connector-java-8.0.11.pom).
This older version is conflicting with Google Cloud Java APIs which depend on `protobuf-java` 3.6.0, but since GCJ doesn't manage the version of `protobuf-java`, `mysql-connector-java` with its `protobuf-java:2.6` being closer to the root of the dependency tree, wins and breaks GCJ.

**Related issues**: https://github.com/spring-cloud/spring-cloud-gcp/issues/951, https://bugs.mysql.com/bug.php?id=91999

**Proposed solution**: Exclude `protobuf-java` from `mysql-connector-java` dependency import.
```
        <dependency>
            <groupId>mysql</groupId>
            <artifactId>mysql-connector-java</artifactId>
            <optional>true</optional>
            <!-- protobuf-java:2.6.0 included by MySQL Connector J is not compatible with
                 Google Cloud Java libraries. -->
            <exclusions>
                <exclusion>
                    <groupId>com.google.protobuf</groupId>
                    <artifactId>protobuf-java</artifactId>
                </exclusion>
            </exclusions>
        </dependency>
```

wilkinsona:
From the referenced MySQL bug:

> This dependency is needed only when using new X DevAPI as mentioned on https://dev.mysql.com/doc/connector-j/8.0/en/connector-j-installing-classpath.html or when building the c/J 8.0 from sources. Actually, for the last case you can use protobuf 3.6 version.
>
> So if you use only JDBC API then just don't put protobuf-java-2.6.0.jar into classpath.

Given this, I agree with @saturnism that the Protobuf dependency should probably be optional. Until that's the case, excluding it in our dependency management seems reasonable to me.

wilkinsona:
Closing in favour of PR #14062.

PR #14062, opened 2018-08-14
Prevent mysql-connector-java from pulling in an old version of protobuf-java

spencergibb:
fixes gh-14060


wilkinsona:
Thank you, @spencergibb. The proposed changes are now in master.

PR #14065, opened 2018-08-14
OIDC Provider Configuration Placement

jzheaux:
The location changed for the class that provides a ClientRegistration
via an issuer URL, and this updates Spring Boot accordingly.


mbhave:
Thanks @jzheaux. This is now on master.

PR #14072, opened 2018-08-15
Remove @UsesUnsafeJava annotations

dreis2211:
Hi,

this PR removes the `@UsesUnsafeJava` annotations that seem to be redundant given that Animal Sniffer was removed and they were introduced for it.

Let me know what you think.
Cheers,
Christoph


philwebb:
Reopening because I think we should do this only on master just in case someone happens to be using those annotations.

snicoll:
@philwebb good point, I thought of that as well and decided to go the oversight route and nuke it. Fair enough, I'll restore it in `2.0.x`.

Issue #14076, opened 2018-08-15
Remove redundant spring-boot-starter dependency from starters


mbhave:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before, or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Background
Spring Boot Starters are a set of convenient dependency descriptors that you can include in your application. You get a one-stop-shop for all the Spring and related technology that you need without having to hunt through sample code and copy paste loads of dependency descriptors. For example, if you want to get started using Spring and JPA for database access just include the spring-boot-starter-data-jpa dependency in your project, and you are good to go. 

## Problem
Including `spring-boot-starter` in some starter poms is redundant as this starter gets pulled in by other dependencies. Here is a list of starters that do not explicitly need to include `spring-boot-starter`:

`spring-boot-starter-batch`
`spring-boot-starter-data-jpa`
`spring-boot-starter-data-rest`
`spring-boot-starter-groovy-templates`
`spring-boot-starter-integration`
`spring-boot-starter-jersey`
`spring-boot-starter-jooq`
`spring-boot-starter-web`
`spring-boot-starter-web-services`
`spring-boot-starter-webflux`
`spring-boot-starter-websocket`

## Solution 
Remove the `spring-boot-starter` dependency from the `pom.xml` files for these modules. The modules can be found [here](https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-starters).

## Steps to Fix 
 
- [ ] Claim this issue with a comment here, below, and ask any clarifying questions you need
- [ ] Set up a repository locally following the [Contributing Guidelines](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc)
- [ ] Try to fix the issue following the steps above
- [ ] Commit your changes and start a pull request.

JanGroot:
I'll give it a whirl please.

wilkinsona:
Thank you, @JanGroot. Feel free to ask any questions you may have and we'll do our best to guide you. We'll look forward to your pull request.

JanGroot:
As mentioned in the pull request, I promise to resolve something more substantial soon.
[...]

PR #14078, opened 2018-08-16
Remove redundant spring-boot-starter dependencies

JanGroot:
fixes gh-14076

This is a "fix" for the first timers issue 14076, I really appreciate the fact that it was way more work to create the issue than to actually resolve it. So I hereby pinky swear to pick up something substantial soon.


wilkinsona:
Thanks very much for making your first contribution to Spring Boot. The proposed changes are now in master.

Issue #14083, opened 2018-08-16
Deprecate org.springframework.boot.orm.jpa.EntityManagerFactoryBuilder.EntityManagerFactoryBeanCallback


wilkinsona:
It was added in #894 but, as far as I can tell, it isn't used any more. We should deprecate it in 2.0.x and then remove it in 2.1.

leumas95:
Hi, I was using this functionality to change the CONNECTION_PROVIDER in the jpaPropertyMap.
Is there an alternate solution now that this code has been removed?

philwebb:
@leumas95 Does `HibernatePropertiesCustomizer` meet your needs?

leumas95:
@philwebb yes, thank you that works perfectly.

PR #14085, opened 2018-08-16
deprecate EntityManagerFactoryBuilder.EntityManagerFactoryBeanCallback

ayudovin:
Deprecate
org.springframework.boot.orm.jpa.EntityManagerFactoryBuilder.EntityManagerFactoryBeanCallback
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14083)


snicoll:
Thanks for the PR @ayudovin but we prefer to handle tasks like this ourselves as we need to remove the deprecated code in a merge forward commit. There was a bit more to it anyway, see https://github.com/spring-projects/spring-boot/commit/e03f01564aa7ddb5e7635f160d1a0fb28bdea966

PR #14130, opened 2018-08-18
Remove unused method in TypeUtils

dreis2211:
Hi,

this trivial PR removes an unused method from TypeUtils.

Cheers,
Christoph


snicoll:
Thanks again Christoph!

PR #14139, opened 2018-08-19
Support more Micrometer histogram properties

alexanderabramov:
Adds configuration properties for Micrometer histogram settings:
"minimumExpectedValue", "maximumExpectedValue"


snicoll:
paging @jkschneider for a review.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
```diff
@@ -184,6 +185,22 @@ public void setRequestsMetricName(String requestsMetricName) {
 		 */
 		private final Map<String, ServiceLevelAgreementBoundary[]> sla = new LinkedHashMap<>();
 
+		/**
+		 * The minimum value that this distribution summary is expected to observe.
+		 * Controls the number of buckets shipped by percentilesHistogram. Can be
+		 * specified as a long or as a Duration value (for timer meters, defaulting to ms
+		 * if no unit specified).
+		 */
+		private final Map<String, ServiceLevelAgreementBoundary> minimumExpectedValue = new LinkedHashMap<>();
```
jkschneider: While `ServiceLevelAgreementBoundary` is a convenient type to deserialize this value, `minimumExpectedValue` is unconnected from SLAs conceptually.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilterTests.java
```diff
@@ -296,6 +296,110 @@ public void configureWhenSlaLongShouldOnlyApplyToTimerAndDistributionSummary() {
...
+				.getMinimumExpectedValue()).isEqualTo(1L);
+	}
+
+	@Test
+	public void configureWhenHasMaximumExpectedValueShouldSetMaximumExpectedToValue() {
+		PropertiesMeterFilter filter = new PropertiesMeterFilter(createProperties(
+				"distribution.maximum-expected-value.[spring.boot]=5000"));
+		assertThat(filter.configure(createMeterId("spring.boot"),
+				DistributionStatisticConfig.DEFAULT).getMaximumExpectedValue())
+						.isEqualTo(5000000000L);
```
jkschneider: For those of us that aren't used to seeing the difference in significant digits between nanoseconds and milliseconds, it might be more clear to write `Duration.ofMillis(5000).toNanos()`

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
```diff
@@ -1365,6 +1365,8 @@ content into your application. Rather, pick only the properties that you need.
 	management.metrics.binders.processor.enabled=true # Whether to enable processor metrics.
 	management.metrics.binders.uptime.enabled=true # Whether to enable uptime metrics.
 	management.metrics.distribution.percentiles-histogram.*= # Whether meter IDs starting with the specified name should publish percentile histograms.
+	management.metrics.distribution.minimum-expected-value.*= # Minimum limit on the histogram buckets for IDs starting with the specified name. The longest match wins, the key `all` can also be used to configure all meters.
```
jkschneider: May or may not be worth mentioning that minimum/maximum can affect the accuracy of client-side percentiles if not set to realistic values.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
jkschneider: WDYT @snicoll?

alexanderabramov: Agree. Which alternative do you guys recommend:
- copy ServiceLevelAgreementBoundary to a better named class with exactly the same code
- rename ServiceLevelAgreementBoundary
- other?

alexanderabramov:
- rebased to latest master
- made PropertiesMeterFilterTests more readable by using Duration.ofMillis().toNanos()

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
alexanderabramov: I'm not sure how to phrase it in a useful way, and the docs already invite the reader to check Micrometer documentation for more details.

alexanderabramov:
@snicoll @jkschneider So what are the next steps on this PR?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
jkschneider: ping @snicoll

snicoll: I frankly have no opinion as I am not sure I've grasped those concepts fully just yet. I've flagged for team attention to get this one more visibility. We have a call today so we should provide a smarter feedback soon.

Thanks for your patience.

wilkinsona: It's unfortunate that the ideal type is either a `long` or a `Duration`. Separate properties for distribution summaries (`long`) and timers (`Duration`) would offer more type-safety in the configuration and could also simplify the property descriptions. I'm not sure what those properties would be called though.

Failing that, I'd type the value as `String` and then convert that to a `long` or `Duration` using the conversion service. That should ensure the same conversion behaviour as if the properties themselves had been more strongly typed.

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
wilkinsona: I don't think we can support `all` here. A `Duration` can specify a unit, for example `30s`. How would that be converted to a plain `long` for those meters that require that rather than a `Duration`?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
snicoll: I guess that ship has sailed with the `sla` property (or we can make this compromise only there). The problem by having two target types depending on the meter type is that we can't really get proper IDE support for this.

I am not sure how we could find a proper name for those and the convenience of doing it on a per meter id is quite appealing. So a `String` to long or `Duration` looks like the way to go.

@alexanderabramov are you interested to revisit the PR short term? RC1 is due soon.

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
snicoll: We've decided to drop support for `sla` so we shouldn't add support for all here either.

jkschneider: Dropping `sla` is a usability concern, as this is going be a frequently set property. What is the reasoning behind this decision?

wilkinsona: We’re not dropping support for `sla` as a whole, just `all` as it doesn’t make sense given the mixture of long and duration values depending on the meter type.

jkschneider: Oh, I see. Many thanks.

snicoll:
@alexanderabramov there is some additional discussion in the comments above. It would be nice to know if you intend to look at them soon as RC1 is due next week.

snicoll:
I've started to review and [polish the PR](https://github.com/snicoll/spring-boot/commit/a507738c5e57836b6f36e190b5cf894d4b150cdc) since RC1 is around the corner. 

I was trying to find a common type for the meter value and renamed `ServiceLevelAgreementBoundary` to `MeterValue`. I am not sure that type makes sense (as it's very generic). 

I may have found a side effect in the binder while doing that so I am investigating that. Perhaps we'll have to postpone this one I am afraid.

wilkinsona:
> I was trying to find a common type for the meter value and renamed `ServiceLevelAgreementBoundary` to `MeterValue`. I am not sure that type makes sense (as it's very generic).

As we just discussed on Slack, I think it's worth having two separate types here as, conceptually, the two things are quite different. Given that we have `ServiceLevelAgreementBoundary`, I'd call it something like `DistributionSummaryBoundary`.

> I may have found a side effect in the binder while doing that so I am investigating that

Unless of course this forces us to use a `String` anyway…

Issue #14151, opened 2018-08-20
ClientSecret is required for Spring Security 5.1 Resource Server and it shouldn't be


mraible:
I tried using Spring Boot 2.1.0.M1 to create an app with Spring Security 5.1 and its resource server support. The first issue I ran into was from Spring Security:

```
Caused by: org.springframework.beans.factory.BeanCreationException: Error creating bean 
with name 'spring.security.oauth2.client-org.springframework.boot.autoconfigure.security.oauth2.client.OAuth2ClientProperties': 
Invocation of init method failed; nested exception is java.lang.IllegalStateException: 
Client secret must not be empty.
```

This was [fixed in Spring Security 5.1.0.RC1](https://github.com/spring-projects/spring-security/issues/5652), so I upgraded to it. Now [I'm still getting the same error](https://gist.github.com/mraible/5d02cd07b34b57ac04925a7e15af4e1e), but it's from Spring Boot now. 

I discussed this issue with @rwinch and he asked me to enter this issue.

mbhave:
@mraible I'm confused because you mentioned resource server support. Did you mean the client support?

mbhave:
Looking at the Spring Security issue, it seems like this issue is about the validation of the client secret in [oauth2 client properties](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientProperties.java#L64).

mraible:
> Did you mean the client support?

It's possible it's called that as well. I call it resource server because there's currently an `@EnableResourceServer` annotation that allows you to create a resource server. When there's only a `clientId`, it just validates the JWT AFAIK.

Here's an example from one of [my blog posts](https://developer.okta.com/blog/2017/12/04/basic-crud-angular-and-spring-boot#spring-security--oauth-20):

```yaml
security:
    oauth2:
        client:
            access-token-uri: https://dev-737523.oktapreview.com/oauth2/default/v1/token
            user-authorization-uri: https://dev-737523.oktapreview.com/oauth2/default/v1/authorize
            client-id: {clientId}
            scope: openid profile email
        resource:
            user-info-uri: https://dev-737523.oktapreview.com/oauth2/default/v1/userinfo
            token-info-uri: https://dev-737523.oktapreview.com/oauth2/default/v1/introspect
            prefer-token-info: false
```

```java
[...]

PR #14156, opened 2018-08-21
clientSecret property is not required

ayudovin:
Remove the validation from Boot for the client secret.
this [bug](https://github.com/spring-projects/spring-boot/issues/14151)


mraible:
@mbhave Was this fix included in Spring Boot 2.1.0 M2?

mbhave:
@mraible No. It will be in M3. 

Thanks, @ayudovin. This commit is now on master.

ayudovin:
@mbhave, thank you!

PR #14162, opened 2018-08-21
Do not create a messageSource bean if one is already defined

mmoayyed:


spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/MessageSourceAutoConfiguration.java
```diff
@@ -86,6 +86,7 @@
 	 */
 	private boolean alwaysUseMessageFormat = false;
 
+	@ConditionalOnMissingBean
```
snicoll: Thanks for the PR but there is already a `@ConditionalOnMissingBean` at class level so something else must be going on in your project. 

There is no test that exercises the problem that you're trying to fix (nor a description). Could you please clarify?

snicoll:
I am closing this one (see review comment). We can reopen if additional clarification is provided that makes this change legit.

PR #14168, opened 2018-08-22
Be more defensive about unchecked exceptions in ApplicationHome

dsyer:
ProtectionDomain in particular can throw a non-Exception. It's
pretty safe to just catch Throwable here and it won't change the
flow.


snicoll:
Closing this one for the same reason expressed in https://github.com/spring-projects/spring-boot/pull/14186#issuecomment-416283703

PR #14172, opened 2018-08-22
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LocalDevToolsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LocalDevToolsAutoConfiguration.java
@@ -191,13 +191,10 @@ public class LocalDevToolsAutoConfiguration {
 
 		@Override
 		public void onApplicationEvent(ApplicationEvent event) {
-			if (event instanceof ContextRefreshedEvent) {
+			if (event instanceof ContextRefreshedEvent
+					|| (event instanceof ClassPathChangedEvent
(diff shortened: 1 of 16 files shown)


spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LocalDevToolsAutoConfiguration.java
```diff
@@ -191,14 +191,11 @@ public boolean supportsSourceType(@Nullable Class<?> sourceType) {
 
 		@Override
 		public void onApplicationEvent(ApplicationEvent event) {
-			if (event instanceof ContextRefreshedEvent) {
+			if (event instanceof ContextRefreshedEvent
```
snicoll: Can we please keep the two ifs please? (and combine the two conditions for the second if). It reads better IMO

izeye: @snicoll I thought that the two `if` statements doing the same thing are unusual as it looks to me as follows:

```
if (condition1) {
  doSomething();
}
if (condition2) {
  doSomething();
}
```

WDYT? If you still prefer as-is, please let me know. Then I'll revert it.

snicoll: Thanks for the feedback. There are but the proposal condition is a big fat IMO. I am not really leaning one way or the other actually, let me checkout the code in an IDE.

snicoll:
Thanks again @izeye!

PR #14176, opened 2018-08-23
Update MongoAutoConfiguration and MongoDataAutoConfiguration to support com.mongodb.client.MongoClient.

christophstrobl:
Next to `com.mongodb.MongoClient` the MongoDB Java drivers offers the `com.mongodb.client.MongoClient` as entry point for database and collection operations. Spring Data MongoDB supports `c.m.client.MongoClient` via its `MongoDbFactory` using `SimpleMongoClientDbFactory`.

The `MongoAutoConfiguration` now backs off if any of those two clients is already defined in the `ApplicationContext` allowing `MongoDataAutoConfiguration` to pick up the users driver implementation of choice.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/mongo/MongoDataAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/mongo/MongoDataAutoConfiguration.java
@@ -16,27 +16,35 @@
 
 package org.springframework.boot.autoconfigure.data.mongo;
 
+import java.util.Arrays;
+import java.util.List;
+
 import com.mongodb.ClientSessionOptions;
 import com.mongodb.DB;
 import com.mongodb.MongoClient;
 import com.mongodb.client.ClientSession;
 import com.mongodb.client.MongoDatabase;
 
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.AnyNestedCondition;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.data.mongo.MongoDataAutoConfiguration.AnySyncMongoClientAvailable;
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/ConditionalOnAnyBean.java
```diff
@@ -0,0 +1,99 @@
...
+ * another auto-configuration, make sure that the one using this condition runs after.
+ *
+ * @author Christoph Strobl
+ * @since 2.1
+ */
+@Target({ ElementType.TYPE, ElementType.METHOD })
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Conditional(OnBeanCondition.class)
+public @interface ConditionalOnAnyBean {
```
snicoll: We have `AnyNestedCondition` for that. Have you considered using that instead? 

I am not keen to expose yet another top level annotations for this feature if we can avoid it.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/mongo/MongoDataAutoConfiguration.java
```diff
@@ -113,6 +128,38 @@ public GridFsTemplate gridFsTemplate(MongoDbFactory mongoDbFactory,
 				mongoTemplate.getConverter());
 	}
 
+	@Nullable
```
snicoll: We don't use `@Nullable` in the code base (yet).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/mongo/MongoDataAutoConfiguration.java
```diff
@@ -113,6 +128,38 @@ public GridFsTemplate gridFsTemplate(MongoDbFactory mongoDbFactory,
...
+			return new SimpleMongoClientDbFactory(client, database);
+		}
+		return null;
+	}
+
+	@Nullable
+	private <T> T mongoClientOrNull(Class<T> mongoClientType) {
+
+		try {
+			return this.beanFactory.getBean(mongoClientType);
```
snicoll: Rather than doing that, could you please inject an `ObjectProvider` and use its API. That would prevent from calling the `BeanFactory` directly.

christophstrobl:
thanks @snicoll will modify the PR according to your input.

christophstrobl:
@snicoll changes are in.

snicoll:
Thanks Christoph, this is now merged with a polish commit (see 56329e6)

PR #14182, opened 2018-08-23
Use null check on groovyReader instead of class loading

dsyer:
To detect Groovy we can use the null value of the reader as a test
equivalently outside the constructor.

Alternative to #13711


dsyer:
Nevermind. It doesn't work with AOT still.

PR #14186, opened 2018-08-23
More defensive about logging system home

dsyer:
We can afford to catch Throwable here. It doesn't affect the flow
and it makes it work with AOT.


snicoll:
What exception do we have with GraalVM? Do we have reported an issue on their tracker?

dsyer:
I think it's the same as this one: https://github.com/oracle/graal/issues/564. Anyway, `ProtectionDomain` doesn't make much sense in a native image, and it's completely optional in Boot as far as I can tell, so it shouldn't really matter much.

dsyer:
See also #14168

snicoll:
Thanks but we prefer the issue on the GraalVM to be taken care of. It it turns out it can't, we can reopen this issue and investigate how this can be tested.

philwebb:
I've raised #14216 to see if we can setup a CI in the longer term

Issue #14150, opened 2018-08-20
Auto-configuration for spring-security-oauth2-resource-server by oidc issuer location


jzheaux:
Spring Boot should add auto configuration support by the oidc issuer location for spring-security-oauth2-resource-server.

For example, a user could configure a resource server in the following way:

```yaml
spring:
  security:
    oauth2:
      resource-server:
        jwt:
          oidc-issuer-location: https://example.org
```

Then, if spring-security-oauth2-resource-server is present on the classpath, as attested by the presence of `org.springframework.security.oauth2.server.resource.authentication.JwtAuthenticationToken`, Spring Boot's auto configuration support could configure a resource server in the following way:

```java
http
    .authorizeRequests()
        .anyRequest().authenticated()
        .and()
    .oauth2()
        .resourceServer()
            .jwt();
```

And then could expose an instance of `JwtDecoder` by invoking:

```java
JwtDecoders.fromOidcIssuerLocation(oidcIssuerLocation);
```

The main value of this specific name is that is comes directly from [the spec language](https://openid.net/specs/openid-connect-discovery-1_0.html#ProviderConfig):

> Using the Issuer location discovered as described in Section 2 or by other means, the OpenID Provider's configuration information can be retrieved.

If the user is curious, she can google "oidc issuer location", and the above spec is the first  link, which is also helpful.

It also clarifies the semantics of the value since there is more than one OAuth spec for discovering configuration, e.g. https://tools.ietf.org/html/rfc8414

Also, while of marginal benefit, it is nice that it matches the method name.

jgrandja:
The updated DSL should be:

```java
http.oauth2ResourceServer().jwt();
```

See related issue [#5715](https://github.com/spring-projects/spring-security/issues/5715)

ayudovin:
I begin working on this enhancement and now I'm blocked by adding unit test because this method `JwtDecoders.fromOidcIssuerLocation(oidcIssuerLocation)` sends request via `restTemplate` :
`return rest.getForObject(oidcIssuerLocation + "/.well-known/openid-configuration", String.class);`.
Could you give some advice how to mock it ?

mbhave:
[...]

PR #14190, opened 2018-08-23
auto-configuration for spring-security-oauth2-resource-server by oidc issuer location

ayudovin:
adding auto-configuration for spring-security-oauth2-resource-server by oidc issuer location
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14150)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwkConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwkConfiguration.java
@@ -48,7 +48,7 @@ class OAuth2ResourceServerJwkConfiguration {
 	}
 
 	@Bean
-	@ConditionalOnProperty(name = "spring.security.oauth2.resource.jwt.oidc-issuer-location")
+	@ConditionalOnProperty(name = "spring.security.oauth2.resourceserver.jwt.oidc-issuer-location")
 	@ConditionalOnMissingBean
 	public JwtDecoder jwtDecoderByOidcIssuerLocation() {
 		return JwtDecoders
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwkConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwkConfiguration.java
@@ -21,12 +21,15 @@ import org.springframework.boot.autoconfigure.security.oauth2.resource.OAuth2Res
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.security.oauth2.jwt.JwtDecoder;
+import org.springframework.security.oauth2.jwt.JwtDecoders;
 import org.springframework.security.oauth2.jwt.NimbusJwtDecoderJwkSupport;
 
 /**
- * Configures a {@link JwtDecoder} when a JWK Set URI is available.
+ * Configures a {@link JwtDecoder} when a JWK Set URI is available or Oidc Issuer
+ * Location.
  *
  * @author Madhura Bhave
+ * @author Artsiom Yudovin
  */
 @Configuration
 class OAuth2ResourceServerJwkConfiguration {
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwkConfiguration.java
```diff
@@ -40,9 +43,17 @@
...
 	@ConditionalOnProperty(name = "spring.security.oauth2.resource.jwt.jwk.set-uri")
 	@ConditionalOnMissingBean
-	public JwtDecoder jwtDecoder() {
+	public JwtDecoder jwtDecoderByUri() {
 		return new NimbusJwtDecoderJwkSupport(
 				this.properties.getJwt().getJwk().getSetUri());
 	}
 
+	@Bean
+	@ConditionalOnProperty(name = "spring.security.oauth2.resource.jwt.jwk.oidc-issuer-location")
```
jzheaux: The property should not use the `jwk` group as this configuration isn't related to it. The property should be more like:

`spring.security.oauth2.resource.jwt.oidc-issuer-location`

ayudovin: yeah, my mistake, I've fixed it

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerAutoConfigurationTests.java
```diff
@@ -86,6 +134,15 @@ public void autoConfigurationShouldBeConditionalOnJwtAuthenticationTokenClass()
...
 
+	@Test
+	public void autoConfigurationShouldBeConditionalOnJwtAuthenticationTokenClassOidcIssuerLocation() {
+		this.contextRunner.withPropertyValues(
+				"spring.security.oauth2.resource.jwt.oidc-issuer-location=http://jwk-oidc-issuer-location.com")
+				.withUserConfiguration(JwtDecoderConfig.class)
+				.withClassLoader(new FilteredClassLoader(JwtAuthenticationToken.class))
+				.run((context) -> assertThat(getBearerTokenFilter(context)).isNull());
+	}
+
```
mbhave: @ayudovin We should probably add a test for what happens when both `oidc-issuer-location` and `jwk-set-uri` are present. My guess is `jwk-set-uri` should win but let's see if @jzheaux thinks otherwise.

mbhave: The other option is to fail hard if both are specified.

ayudovin: From my point of view the second option will be better.

rwinch: I think that jwk-set-uri should win because it is more specific.

jzheaux: Agreed that `jwk-set-uri` is more specific and should take precedence because of that.

ayudovin: ok, now `jwk-set-uri` has precedence.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwkConfiguration.java
```diff
@@ -44,4 +47,12 @@ public JwtDecoder jwtDecoder() {
 		return new NimbusJwtDecoderJwkSupport(this.properties.getJwt().getJwkSetUri());
 	}
 
+	@Bean
+	@ConditionalOnProperty(name = "spring.security.oauth2.resourceserver.jwt.oidc-issuer-location")
```
mbhave: I think we should make it explicit that `jwk-set-uri` should take precedence by adding a custom condition that checks that `spring.security.oauth2.resourceserver.jwt.oidc-issuer-location` is present and `spring.security.oauth2.resourceserver.jwt.jwk-set-uri` is not present.

ayudovin: ok, I'll add custom condition

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwkConfiguration.java
```diff
@@ -44,4 +49,12 @@ public JwtDecoder jwtDecoder() {
 		return new NimbusJwtDecoderJwkSupport(this.properties.getJwt().getJwkSetUri());
 	}
 
+	@Bean
+	@Conditional(OidcIssuerLocationCondition.class)
```
ayudovin: @mbhave I've added custom condition.

mbhave:
Thanks again, @ayudovin! This change is now on master along with [this polish commit](https://github.com/spring-projects/spring-boot/commit/cf31325e5d969ae5ac38e814a9137030b650468f). I changed `oidcIssuerLocation` to `issuerUri` to make it consistent with the other properties in `OAuth2ClientProperties` and `OAuth2ResourceServerProperties`.

ayudovin:
@mbhave , it makes sense. Thank you!

Issue #14196, opened 2018-08-24
Allow the ability to set shareNativeConnection to false for LettuceConnectionFactory


csavory:
By default, the `LettuceConnection` will use the `sharedConnection` for almost all non-tx or non-blocking operations. 

I have found the exclusive use of the shared connection to not allow very high throughput of many concurrent Redis calls. My reads connections are constantly being blocked by `LettuceFutures.awaitOrCancel`.  Therefore I would like to set `shareNativeConnection` to false in the `LettuceConnectionFactory` so it always uses the pooled connections.  

Spring Boot `LettuceConnectionConfiguration` does not currently allow for `shareNativeConnection` to be set.

https://docs.spring.io/spring-data/redis/docs/2.1.0.RC2/reference/html/#redis:connectors:lettuce

philwebb:
@csavory For now you'll need to define your own `LettuceConnectionFactory` `@Bean`.

philwebb:
Closing in favor of PR #14217

mp911de:
We should clarify one thing first before extending configuration: Redis itself is single-threaded so increasing the connection count for Redis Standalone does not improve throughput performance. 

@csavory if you recognize a single connection as an obstruction for your application, then we should continue the discussion along with some profiling data in https://github.com/lettuce-io/lettuce-core.

snicoll:
@csavory switching to declined based on the comment above. Please follow up on the lettuce project and we can reopen if it turns out something needs to be done in Spring Boot.

PR #14217, opened 2018-08-27
Allow the ability to set shareNativeConnection to false for LettuceConnectionFactory

ayudovin:
Allow the ability to set shareNativeConnection for LettuceConnectionFactory
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14196)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.boot.autoconfigure.data.redis;
 import java.net.UnknownHostException;
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 
 import io.lettuce.core.RedisClient;
 import io.lettuce.core.resource.ClientResources;
@@ -45,6 +46,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Mark Paluch
  * @author Andy Wilkinson
+ * @author Artsiom Yudovin
  */
 @Configuration
 @ConditionalOnClass(RedisClient.class)
@@ -76,7 +78,15 @@ class LettuceConnectionConfiguration extends RedisConnectionConfiguration {
 			ClientResources clientResources) throws UnknownHostException {
 		LettuceClientConfiguration clientConfig = getLettuceClientConfiguration(
 				clientResources, this.properties.getLettuce().getPool());
-		return createLettuceConnectionFactory(clientConfig);
+		LettuceConnectionFactory lettuceConnectionFactory = createLettuceConnectionFactory(
+				clientConfig);
+
+		if (Objects.nonNull(this.properties.getLettuce().getShareNativeConnection())) {
+			lettuceConnectionFactory.setShareNativeConnection(
+					this.properties.getLettuce().getShareNativeConnection());
+		}
+
+		return lettuceConnectionFactory;
 	}
 
 	private LettuceConnectionFactory createLettuceConnectionFactory(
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
```diff
@@ -76,7 +78,15 @@ public LettuceConnectionFactory redisConnectionFactory(
 			ClientResources clientResources) throws UnknownHostException {
 		LettuceClientConfiguration clientConfig = getLettuceClientConfiguration(
 				clientResources, this.properties.getLettuce().getPool());
-		return createLettuceConnectionFactory(clientConfig);
+		LettuceConnectionFactory lettuceConnectionFactory = createLettuceConnectionFactory(
+				clientConfig);
+
+		if (Objects.nonNull(this.properties.getLettuce().getShareNativeConnection())) {
```
snicoll: This looks odd that this property alone has to be handled separately.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -325,6 +326,14 @@ public void setPool(Pool pool) {
 		 */
 		private Pool pool;
 
+		/**
+		 * If shareNativeConnection is true, the pool will be used to select a connection
+		 * for blocking and tx operations only, which should not share a connection. If
+		 * native connection sharing is false, the selected connection will be used for
+		 * all operations
+		 */
+		private Boolean shareNativeConnection;
```
snicoll: We should make up our mind about a default if that makes sense. If it does, it should be reflected here (and tested).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -325,6 +326,14 @@ public void setPool(Pool pool) {
 		 */
 		private Pool pool;
 
+		/**
+		 * If shareNativeConnection is true, the pool will be used to select a connection
```
snicoll: That's a bit unusual for the description of a key (see other descriptions around for inspiration). The first sentence must be a concise summary of the purpose of the key as it's the one showing up in auto-completion popup in an IDE.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfigurationTests.java
```diff
@@ -150,12 +151,15 @@ public void testPasswordInUrlStartsWithColon() {
...
 	@Test
 	public void testRedisConfigurationWithPool() {
-		this.contextRunner.withPropertyValues("spring.redis.host:foo",
-				"spring.redis.lettuce.pool.min-idle:1",
-				"spring.redis.lettuce.pool.max-idle:4",
-				"spring.redis.lettuce.pool.max-active:16",
-				"spring.redis.lettuce.pool.max-wait:2000",
-				"spring.redis.lettuce.shutdown-timeout:1000").run((context) -> {
+		this.contextRunner
+				.withPropertyValues("spring.redis.host:foo",
```
snicoll: I am not sure that reformat is necessary for adding a single key

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
ayudovin: the default value of  `shareNativeConnection` is true.  This default value is located in `LettuceConnectionFactory` but we also could the same default value in `RedisProperties`

mp911de:
`LettuceClientConfiguration` encapsulates driver-specifics. Connection sharing and other properties like connection validation are specific to `LettuceConnectionFactory`.

Changing connection sharing settings is a somewhat rare use-case and I'd rather propose to use a `BeanPostProcessor` on the application side to tweak settings where necessary. I left a comment in #14196 why I think using multiple connections won't really help.

snicoll:
Thanks for the feedback @mp911de 

@ayudovin thanks for the PR anyway.

PR #14221, opened 2018-08-28
Polish some joining collectors

dreis2211:
Hi,

this PR polishes two occurences where the result of `Collectors.joining()` is concatenated with another String. These could be rewritten to include the specific String as a prefix in the joining Collector, which saves some allocations.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/annotation/DiscoveredWebOperation.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/annotation/DiscoveredWebOperation.java
@@ -61,8 +61,8 @@ class DiscoveredWebOperation extends AbstractDiscoveredOperation implements WebO
 	}
 
 	private String getId(String endpointId, Method method) {
-		return endpointId + Stream.of(method.getParameters()).filter(this::hasSelector)
-				.map(this::dashName).collect(Collectors.joining());
+		return Stream.of(method.getParameters()).filter(this::hasSelector)
+				.map(this::dashName).collect(Collectors.joining("", endpointId, ""));
 	}
 
 	private boolean hasSelector(Parameter parameter) {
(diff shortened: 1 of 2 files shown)


snicoll:
Thanks again Christoph!

philwebb:
I'm not a big fan of this one, I think the three argument `joining` call is harder to read.

dreis2211:
I kind of understand that, because of the somehow useless empty string parameters. In the end a matter of taste, I'd say. Making it part of the stream is for me more readable because it feels more fluent. Yet, I have no strong feelings against reverting it if the team doesn't like it.

Yet, for completeness reasons here is the benchmark that compares the old vs. the new approach.
```
Benchmark                                              Mode  Cnt        Score        Error   Units
MyBenchmark.testNew                                   thrpt   10  6201587,136 ± 267064,494   ops/s
MyBenchmark.testNew:·gc.alloc.rate                    thrpt   10     3263,741 ±    140,629  MB/sec
MyBenchmark.testNew:·gc.alloc.rate.norm               thrpt   10      552,000 ±      0,001    B/op
MyBenchmark.testNew:·gc.count                         thrpt   10      135,000               counts
MyBenchmark.testNew:·gc.time                          thrpt   10       93,000                   ms
MyBenchmark.testOld                                   thrpt   10  5793186,832 ± 163477,518   ops/s
MyBenchmark.testOld:·gc.alloc.rate                    thrpt   10     3623,066 ±    102,184  MB/sec
MyBenchmark.testOld:·gc.alloc.rate.norm               thrpt   10      656,000 ±      0,001    B/op
MyBenchmark.testOld:·gc.count                         thrpt   10      129,000               counts
MyBenchmark.testOld:·gc.time                          thrpt   10       93,000                   ms
```

snicoll:
@dreis2211 thanks for the PR but we had our team meeting and decided to revert it to favour readability considering those two aren't hotspots. Thanks anyway

dreis2211:
Fair enough. Thanks for coming back to me, @snicoll

PR #14008, opened 2018-08-07
introduce HiddenHttpMethodFilter configuration in WebFlux

ayudovin:
Adding  `HiddenHttpMethodFilter` configuration in `WebFluxAutoConfiguration`.
this [enhancement](https://github.com/spring-projects/spring-boot/issues/9167)

Issue #3613, opened 2015-07-28
Add Order Constant for Filters that wrap the HttpServletRequest


rwinch:
The idea would be that any Filter that wraps the HttpServletRequest would be before this constant. Any Filter that relies on the wrapped HttpServletRequest would be after the constant. This would help in situations like https://github.com/grails/grails-core/issues/9137

dsyer:
Can we start by making a list of Filters that we know might wrap the request? Spring Security's main filter is one. Also `HiddenHttpMethodFilter`. There are various filters in Spring (e.g. `*RequestLoggingFilter`) that we don't use routinely.

The `ErrorPageFilter` wraps the response, does that count? Spring Cloud wraps the request inside `Zuul`, does that count? Similar question for the multipart support in Spring - it wraps the request, but not in a `Filter`.

dsyer:
OK, there's nothing we can do about the downstream components (like `DispatcherServlet` wrapping a request). All the filters added explicitly by Spring Boot now have order <= `FilterRegistrationBean.REQUEST_WRAPPER_FILTER_MAX_ORDER` (value 0).

PR #14228, opened 2018-08-29
Drop a constant from Servlet package in OrderedHiddenHttpMethodFilter

izeye:
The constant comes from Servlet package and has a value of `0`, so this PR simply drops the constant.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/filter/OrderedHiddenHttpMethodFilter.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/filter/OrderedHiddenHttpMethodFilter.java
@@ -16,7 +16,6 @@
 
 package org.springframework.boot.web.reactive.filter;
 
-import org.springframework.boot.web.servlet.FilterRegistrationBean;
 import org.springframework.core.Ordered;
 import org.springframework.web.filter.reactive.HiddenHttpMethodFilter;
 
@@ -32,8 +31,7 @@ public class OrderedHiddenHttpMethodFilter extends HiddenHttpMethodFilter
 	/**
 	 * The default order is high to ensure the filter is applied before Spring Security.
 	 */
-	public static final int DEFAULT_ORDER = FilterRegistrationBean.REQUEST_WRAPPER_FILTER_MAX_ORDER
-			- 10000;
+	public static final int DEFAULT_ORDER = -10000;
 
 	private int order = DEFAULT_ORDER;
 


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/filter/OrderedHiddenHttpMethodFilter.java
```diff
@@ -32,8 +31,7 @@
 	/**
 	 * The default order is high to ensure the filter is applied before Spring Security.
 	 */
-	public static final int DEFAULT_ORDER = FilterRegistrationBean.REQUEST_WRAPPER_FILTER_MAX_ORDER
-			- 10000;
+	public static final int DEFAULT_ORDER = -10000;
```
snicoll: I can understand why you'd want to drop a reference to this class but I am not sure copy/pasting the default (I mean `0`) is something we want to do. Perhaps we need a shared context for this?

WDYT?

izeye: @snicoll Thanks for the quick feedback! I thought the same thing but didn't find a good place to move them. So I considered adding a constant container class like `FilterOrders` but with only one member, I wasn't sure it's okay. Any suggestion?

snicoll: @bclozel that was added as part of #14008 - Does the constant has to be the same for reactive? If it's the same then we should probably need some concept that gathers them. What do you think?

snicoll:
This should be applied to `2.0.x` as well.

philwebb:
I think for 2.0.x we should just drop the reference and hard code 0 but it feels like we're missing something from the code. The constant was originally added in 1.3 (see #3613) but looking again I wonder if we should move it out of `FilterRegistrationBean`. It really feels quite specific to be in there.

I wonder if adding an `OrderedFilter` interface might be useful? We could add one in `org.springframework.boot.web.servlet.filter` which extended `Ordered` and `Filter` and another in `org.springframework.boot.web.reactive.filter` that extends `Ordered` and `WebFilter`. I don't think there's currently a need for a common constant. I think I'd rather copy/paste than introduce another package.

izeye:
This looks resolved via 3ff20b2440e3b268207a63b23b531e75cf1c9bd6, so we can close this PR unless there needs to discuss more on this.

philwebb:
It might be worth still considering the `OrderedFilter` idea.

philwebb:
Closing in favor of #14793

Issue #13831, opened 2018-07-20
Align max HTTP header size configuration across all supported containers


wilkinsona:
Max HTTP header size is currently configured using `server.max-http-header-size`. It is an `int` in `ServerProperties` with a default value of 0. The actual default value for Tomcat and Jetty is 8KB and for Undertow is 1MB. The property doesn't seem to have any effect on Netty. These different defaults make it impossible for a property that's shared across containers to have an accurate default in the metadata. An added complication is that Jetty allows different max headers sizes for requests and responses.

A couple of options:

- Pick a default in Boot and use this across all the containers.
- Make the property container specific so that it can accurately reflect the container's default and so that we can offer separate properties for requests and responses with Jetty if we wish.

snicoll:
Closing in favour of PR #13831

PR #14232, opened 2018-08-29
Add netty request decoder customizer

CinnamonBagels:
When making the jump from Tomcat to Netty, I noticed that my `-Dserver.max-http-header-size` configuration was not being honored. This PR allows users to configure the netty `HttpRequestDecoderSpec` using ServerProperties.

#### Notes on implementation
* (`serverProperties` max-http-header-size of <= 0 will defer to `serverProperties.netty` maxHeaderSize)


snicoll:
@CinnamonBagels thanks a lot for the PR but we already have an issue for this (#13831) and a PR (#14234). We'd rather fix the `server.max-http-header-size` situation first.

The other part of this PR look unrelated to the task at hand. We usually expose first class properties based on a concrete scenario. Can you please share why you believe those properties would be useful vs. a programmatic customizer?

CinnamonBagels:
I added all of the configuration options that HttpRequestDecoderSpec allowed on a nice-to-have basis.  Now  I understand the reasons why properties are exposed, so my changes don't make sense yet. Thanks for pointing me to the issue and PR and giving feedback.

snicoll:
Alright, thanks for the PR anyway.

Issue #13821, opened 2018-07-19
Documented defaults for some server properties do not match runtime defaults


wilkinsona:
For a number of properties, the default in the code and, therefore, in the generated metadata is zero or null. At runtime, such values are interpreted as meaning that the component's default should be used. An unwanted side-effect of this is that the default from the metadata that is shown in your IDE does not match the default that you'll get at runtime. We should align the defaults in the code and metadata with the runtime defaults and add tests to ensure that the two are kept in sync.

wilkinsona:
This problem is broader than just server properties.

snicoll:
I guess all the "bytes" type properties are covered by this one. When merging forward to `master` we should switch those to `DataSize`. Given we're approaching RC1 it would be nice to do this sooner than later.

wilkinsona:
To keep the scope management, let's narrow this back down to focus on server properties. We can open separate issues for other areas of the codebase.

snicoll:
This change unfortunately breaks the Tomcat 7 support in a very subtle way. `redirectContextRoot` has now a default value, which means that we're going to invoke the customizer no matter what and that method is not available on Tomcat 7.

PR #14234, opened 2018-08-29
Align max HTTP header size configuration across all supported containers 

qct:
Align max HTTP header size configuration across all supported containers

Fix: #13831

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -53,6 +53,7 @@ import org.springframework.util.unit.DataSize;
  * @author Aurélien Leboulanger
  * @author Brian Clozel
  * @author Olivier Lamy
+ * @author Chentao Qu
  */
 @ConfigurationProperties(prefix = "server", ignoreUnknownFields = true)
 public class ServerProperties {
@@ -81,9 +82,9 @@ public class ServerProperties {
 	private String serverHeader;
 
 	/**
-	 * Maximum size, in bytes, of the HTTP message header.
+	 * Maximum size of the HTTP message header.
 	 */
-	private int maxHttpHeaderSize = 8192;
+	private DataSize maxHttpHeaderSize = DataSize.ofKiloBytes(8L);
 
 	/**
 	 * Time that connectors wait for another HTTP request before closing the connection.
@@ -141,11 +142,11 @@ public class ServerProperties {
 		this.serverHeader = serverHeader;
 	}
 
-	public int getMaxHttpHeaderSize() {
+	public DataSize getMaxHttpHeaderSize() {
 		return this.maxHttpHeaderSize;
 	}
 
-	public void setMaxHttpHeaderSize(int maxHttpHeaderSize) {
+	public void setMaxHttpHeaderSize(DataSize maxHttpHeaderSize) {
 		this.maxHttpHeaderSize = maxHttpHeaderSize;
 	}
 
@@ -327,7 +328,9 @@ public class ServerProperties {
 
 		/**
 		 * Maximum size, in bytes, of the HTTP message header.
+		 * @deprecated since 2.1.0 in favor of {@link ServerProperties#maxHttpHeaderSize}
 		 */
+		@Deprecated
 		private int maxHttpHeaderSize = 0;
 
 		/**
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -729,11 +729,6 @@ public class ServerProperties {
 		 */
 		private Integer selectors;
 
-		/**
-		 * Maximum size, in bytes, of the HTTP response header.
-		 */
-		private int maxHttpResponseHeaderSize;
-
 		public Accesslog getAccesslog() {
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -83,7 +83,7 @@
 	/**
 	 * Maximum size, in bytes, of the HTTP message header.
 	 */
-	private int maxHttpHeaderSize = 0; // bytes
+	private int maxHttpHeaderSize = 8192;
```
snicoll: Please use `DataSize` and `DataSize.ofKiloBytes(8)` for the default value. The description should no longer mention "in bytes" since that's inferred by the type. The updated description and the value for that property should also be aligned to `8KB` in the appendix.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -729,6 +729,11 @@ public void setCacheTtl(Duration cacheTtl) {
 		 */
 		private Integer selectors;
 
+		/**
+		 * Maximum size, in bytes, of the HTTP response header.
+		 */
+		private int maxHttpResponseHeaderSize;
```
snicoll: I don't think that matches what [our intent was](https://github.com/spring-projects/spring-boot/issues/13831#issuecomment-409617001). By customizer, we mean that a user should create a customizer if they want to customize the response size. Please remove this new property and its handling.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
qct: @snicoll  thanks for review, two questions:
Q1: Do you mean change the type of maxHttpHeaderSize from int to DataSize ? That will need to do some refactor beccause there are some reference to getMaxHttpHeaderSize.  
Q2: How about other "in bytes" properties, do we also need to change their types to DataSize ? 
eg: https://github.com/spring-projects/spring-boot/blob/0590c4de3210a1cf41e63b928d46f99f6c0c5c6b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java#L326
https://github.com/spring-projects/spring-boot/blob/0590c4de3210a1cf41e63b928d46f99f6c0c5c6b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java#L331
https://github.com/spring-projects/spring-boot/blob/0590c4de3210a1cf41e63b928d46f99f6c0c5c6b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java#L720
https://github.com/spring-projects/spring-boot/blob/0590c4de3210a1cf41e63b928d46f99f6c0c5c6b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java#L933
https://github.com/spring-projects/spring-boot/blob/0590c4de3210a1cf41e63b928d46f99f6c0c5c6b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java#L938

snicoll: > Do you mean change the type of maxHttpHeaderSize from int to DataSize ?

Yes

> How about other "in bytes" properties, do we also need to change their types to DataSize ?

Yes and we have a separate issue for that (we need to align their default value first and when we'll merge forward we'll do the change. I've added a note in #13821 (in any case this PR is about a specific task and should not be expanded to cover unrelated changes). Thanks!

qct: Thank you for explain, still have a question, if we make default value of maxHttpHeaderSize  to 8M, these code seem to make no sense.
https://github.com/spring-projects/spring-boot/blob/0590c4de3210a1cf41e63b928d46f99f6c0c5c6b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/embedded/TomcatWebServerFactoryCustomizer.java#L117
I'm wondering why do you prefer `serverProperties.getMaxHttpHeaderSize()` than `serverProperties.getTomcat().getMaxHttpHeaderSize()` at the beginning?

snicoll: `8KB`, not `8MB`. The property must be deprecated and that method as well to offer a transition. If you don't know how to do this, I can take care of it in a polish commit.

Na3blis:
@snicoll This should definitely be added to the changelog! We just spent 3 weeks tracking down  random 400 errors we were receiving. We tracked it down to upgrading from Spring Boot 2.0 to 2.1, but after combing through the changelogs we didn't see anything that would cause our problems. Finally we came across this, and figured out the issue, since we use Undertow.

wilkinsona:
Sorry for the trouble, @Na3blis, and thanks for taking the time to make us aware of the problem caused. I've added [an entry](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#consistent-max-http-header-size-across-all-embedded-web-servers) to the release notes for 2.1 so that others hopefully have an easier time of it.

Issue #14069, opened 2018-08-15
Improve Spring WebSockets documentation for Spring WebFlux


mdeinum:
I was trying to piece together a simple chat sample using Spring Boot, WebFlux and WebSockets. However that isn't as straight forward (initially) as one might think. 

Including `spring-boot-starter-websocket` brings in `spring-boot-starter-web` and suddenly bootstraps Tomcat instead of Netty and rigs up Spring MVC. Adding exclusions won't be of much help either unless you do very detailed exclusions (excluding `spring-boot-starter-web` will exclude too much again). 

I made it work by just including `javax.websocket-api` (version 1.1) as a dependency (maybe this needs to be added to the `spring-boot-dependencies` as a managed dependency?).

For testing I include some WebSocket client library so I could write an integration test. However it took me  a while to figure out that `spring-boot-starter-websocket` is meant for the classic Servlet based approach and not the reactive approach. 

It might be worthwhile to include this in the WebSockets section of the Spring Boot documentation or split that into 2 sections for a Servlet and Reactive approach. 

Added bonus would be to have a sample web socket / reactive included in the spring boot samples.

alexmandt:
Hi. I'm a first time contributor and found this issue. I wanted to ask, if this issue was open for contribution?

snicoll:
@alexmandt thanks for offering to help. Most issues are opened to contribution unless they're already assigned. That's not the case here so please go ahead. If you have questions or need guidance, we're [on Gitter](https://gitter.im/spring-projects/spring-boot).

alexmandt:
Thanks

PR #14238, opened 2018-08-29
Improve WebSocket documentation for Spring WebFlux

alexmandt:
As part of Issue #14069, a passage about Reactive WebSockets is added to the WebSocket section of the Spring Boot Reference Documentation.

Also, as part of Issue #14069, the sample project `spring-boot-sample-websocket-reactive` is added to the 
Spring Boot Samples while also being mentioned in the Spring Boot Reference Documentation in the WebSocket Section.


snicoll:
@alexmandt thank you very much for the PR but there are several problems with it that led to me roll a separate fix:  2fc557a

We usually don't refer to a sample to explain how something must be done. If it's too complicated to phrase in the ref guide, then something else is wrong. In the case your sample, you import a dependency for which we have no dependency management for (and brings additional libraries that we don't need).

Also the support for reactive web applications is not at the same level as MVC (much more basic, no STOMP support) so I'd prefer to document it is possible rather than giving it more visibility with a sample.

Thanks anyway.

alexmandt:
Thanks for the feedback.

Issue #14246, opened 2018-08-30
doc-root not found


JoeKerouac:
doc-root not found when running in IDEA.

method:org.springframework.boot.web.servlet.server.DocumentRoot.getValidDirectory().

snicoll:
@JoeKerouac as indicated in the issue template, we do not need an issue and PR. Closing this one for now.

PR #14247, opened 2018-08-30
Fix doc root detection when running in IntelliJ IDEA

JoeKerouac:
Fix doc-root not found when running in IDEA.
#14246

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/server/DocumentRoot.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/server/DocumentRoot.java
@@ -100,7 +100,8 @@ class DocumentRoot {
 		}
 		else if (DEFAULT_DOC_ROOT.exists()) {
 			if (this.logger.isDebugEnabled()) {
-				this.logger.debug("Currently running in the IDEA, did not find the workspace, but found the doc-root under the classpath");
+				this.logger.debug("Currently running in the IDEA, did not find the workspace, but found the doc-root" +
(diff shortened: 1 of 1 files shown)


snicoll:
@JoeKerouac in the future could you please run the build locally and check rather than pushing those individual commits? That generates quite some unnecessary notification noise. Thanks.

JoeKerouac:
@snicoll OK.

wilkinsona:
Thanks for the PR, @JoeKerouac. Unfortunately, I'm not too keen on having the amount of IDE-specific code in `DocumentRoot`, particularly as that class isn't only used during development. Can you please describe the problem that you're facing and why these changes are necessary to solve it?

JoeKerouac:
When I run the Spring-boot project in Idea, my project contains JSP, there may be a problem with my usage, spring-boot does not find my JSP page correctly, I found the problem by debug Doc-root, When I try to add the above code, he can work normally. In addition, I was in the construction of the Spring-boot project itself has been an error, the environment is CENTOS+OPENJDK8, with the project mvnw.
@wilkinsona

JoeKerouac:
Sorry, this problem appeared in our project. I thought it was a spring-boot problem. I didn’t confirm it afterwards. I just found a simple demo and found that the problem has not been reproduced. It may be that our project may also have problems. It is an environmental issue, I will confirm it again. Finally, thank you for your patience. @wilkinsona

wilkinsona:
Thanks for following up. I'll close this one for now. If we need to revisit this, it's probably best to use #14246 to precisely identify the problem and determine a solution.

Issue #14122, opened 2018-08-17
Extend RabbitProperties with SimpleMessageListenerContainer.setMissingQueuesFatal


ramato-procon:
Please add to [RabbitProperties-SimpleContainer](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java#L623) the [SimpleMessageListenerContainer.setMissingQueuesFatal](https://docs.spring.io/spring-amqp/api/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.html#setMissingQueuesFatal-boolean-).
Especially for clustered rabbitmq configuration the default option (true) may cause the ListenerContainer to stop forever until the SpringBoot instance was restarted. E.g. exclusive (Anonymous) queues may not be recreated during a network partition.
It should also be discussed if Autoconfiguration should set the default value to "false".

garyrussell:
>It should also be discussed if Autoconfiguration should set the default value to "false".

The default should remain `true` IMHO.

mbhave:
Closing in favor of PR #14252.

PR #14252, opened 2018-08-30
Add support for configuring missingQueuesFatal property

nosan:
see [gh-14122](https://github.com/spring-projects/spring-boot/issues/14122)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/DirectRabbitListenerContainerFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/DirectRabbitListenerContainerFactoryConfigurer.java
@@ -40,6 +40,8 @@ public final class DirectRabbitListenerContainerFactoryConfigurer extends
 		configure(factory, connectionFactory, config);
 		map.from(config::getConsumersPerQueue).whenNonNull()
 				.to(factory::setConsumersPerQueue);
+		map.from(config::getMissingQueuesFatal).whenNonNull()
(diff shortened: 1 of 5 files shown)


snicoll:
Thanks again @nosan, please review the [polish commit](https://github.com/spring-projects/spring-boot/commit/c1322a543ce426f1daa71c9a9f30d051347b6e46) if you can.

nosan:
@snicoll I wanted to do the same way, but I haven't done because of the javadoc meanings. Could you please check one more time the difference between simple/direct  `missing-queues-fatal`, just to make everything is ok :)

[Simple ](https://docs.spring.io/spring-amqp/api/org/springframework/amqp/rabbit/listener/SimpleMessageListenerContainer.html#setMissingQueuesFatal-boolean-) , [Direct](https://docs.spring.io/spring-amqp/api/org/springframework/amqp/rabbit/listener/DirectMessageListenerContainer.html#setMissingQueuesFatal-boolean-)


Thanks in advance.

garyrussell:
>When true, if the queues are removed while the container is running, the containeris stopped.

This difference is because of the internal design of the containers.

With the SMLC, each consumer instance can listen to multiple queues; if any one of those queues is deleted from the broker; the consumer is killed by the broker.
With the new container (Direct), there is a consumer for each queue, so deleting one queue doesn’t affect the other consumer(s).

snicoll:
Alright, let's document that as two keys. I'll push another polish later today

snicoll:
@nosan thank you very much for drawing my attention to this. I thought that the properties were common since it was calling the abstract factory but it turns out that there is a subtle semantic difference the we have to explain. I took the opportunity to revisit this a bit further by exposing the difference in default values (rather than keeping `Boolean).

PR #14254, opened 2018-08-30
Fix String concatenation

Hanope:
I found that ErrorMvcAutoConfiguration uses String Concat as a "+" operation.
So I Convert String concatenation used the "+" operator to StringBuilder.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/ErrorMvcAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/ErrorMvcAutoConfiguration.java
@@ -253,13 +253,13 @@ public class ErrorMvcAutoConfiguration {
 
 		private String getMessage(Map<String, ?> model) {
 			Object path = model.get("path");
-			String message = "Cannot render error page for request [" + path + "]";
+			StringBuilder message = new StringBuilder(String.format("Cannot render error page for request [%s]", path));
 			if (model.get("message") != null) {
-				message += " and exception [" + model.get("message") + "]";
+				message.append(" and exception [" + model.get("message") + "]");
 			}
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/ErrorMvcAutoConfiguration.java
```diff
@@ -253,13 +253,13 @@ public void render(Map<String, ?> model, HttpServletRequest request,
 
 		private String getMessage(Map<String, ?> model) {
 			Object path = model.get("path");
-			String message = "Cannot render error page for request [" + path + "]";
+			StringBuilder message = new StringBuilder(String.format("Cannot render error page for request [%s]", path));
```
bclozel: Why introduce here a `String.format` call when the spirit of this PR is to improve performance?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/ErrorMvcAutoConfiguration.java
```diff
@@ -253,13 +253,13 @@ public void render(Map<String, ?> model, HttpServletRequest request,
...
 			if (model.get("message") != null) {
-				message += " and exception [" + model.get("message") + "]";
+				message.append(" and exception [" + model.get("message") + "]");
```
bclozel: Why did you keep concatenation operations here?

philwebb:
Thanks for the pull request, but I'm not a big fan of changing string concatenation to`StringBuilder` unless performance is absolutely an issue. In this case, the error controller isn't called that often and it isn't doing a great deal so I don't think the change is warranted.

I'm happy to be proved wrong if you find that it does make a practical difference, but for now I'm going to close this one.

PR #14255, opened 2018-08-30
MetricsRun could be run not only with ApplicationContextRunner

nosan:
`MetricsRun` could be run not only with `ApplicationContextRunner` but with any of `AbstractApplicationContextRunner's` implementation.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/amqp/RabbitMetricsAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/amqp/RabbitMetricsAutoConfigurationTests.java
@@ -34,8 +34,8 @@ import static org.assertj.core.api.Assertions.assertThat;
 public class RabbitMetricsAutoConfigurationTests {
 
 	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
-			.with(MetricsRun.simple())
-			.withConfiguration(AutoConfigurations.of(RabbitAutoConfiguration.class));
+			.with(MetricsRun.simple()).withConfiguration(AutoConfigurations.of(
+					RabbitAutoConfiguration.class, RabbitMetricsAutoConfiguration.class));
 
 	@Test
 	public void autoConfiguredConnectionFactoryIsInstrumented() {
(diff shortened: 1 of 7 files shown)


snicoll:
@nosan I am not sure I like `MetricsRun` in the first place. IMO, our test should only load what is strictly necessary and that doesn't go in that direction. If we configure all metrics test to run this, we no longer test what happens when a certain component is not there.

philwebb:
Metrics setup for tests is a bit of pain so I quite like `MetricsRun`. From a quick glace, I like the changes in this PR as well (other than the already flagged generics stuff).

snicoll:
I beg to differ. Most of what `MetricsRun` does is not required by those tests. Why would we load all these auto-configurations?

```
MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class,
			RabbitMetricsAutoConfiguration.class, CacheMetricsAutoConfiguration.class,
			DataSourcePoolMetricsAutoConfiguration.class,
			HibernateMetricsAutoConfiguration.class,
			RestTemplateMetricsAutoConfiguration.class,
			WebClientMetricsAutoConfiguration.class,
			WebFluxMetricsAutoConfiguration.class, WebMvcMetricsAutoConfiguration.class
```

The tests that were changed in this PR tests an isolated part of our infrastructure so I prefer that each of those tests describe an environment that is specific for the task at hand.

For instance, if you create a runner for a reactive web application, why would you attempt to load `WebMvcMetricsAutoConfiguration`?

Another example `RestTemplateMetricsAutoConfiguration` currently provides the filter for uri-tag that the `WebClient` is using. @nosan promised a PR to fix that but if we switch to this infrastructure, the fact the WebClient tests are loading a `RestTemplateMetricsAutoConfiguration` class would be completely hidden. 

I have no recollection that this pattern is used anywhere else. I see a value in "give me an in-memory `MeterRegistry`" so perhaps this class can be improved to do only that?

nosan:
first of all thank you for your feedbacks @snicoll , @philwebb .
I agree with @snicoll  that `MetricsRun` has a lot of unnecessary classes instead of two. 
I've updated this PR.


P.S. @philwebb Please pay attention to `WebClientMetricsAutoConfigurationTests` there is an issue, @snicoll was talking about.

P.S. @snicoll 

>  @nosan promised a PR to fix

https://github.com/spring-projects/spring-boot/pull/14267

snicoll:
Thanks for the PR @nosan. I've reviewed our current situation and decided to refactor `MetricsRun` a bit further. See 1ceb076 for more details.

PR #14267, opened 2018-08-31
WebClient and RestTemplate 'uri' tag filter.

nosan:
Currently, `WebClient` using `URI` tag filter which was provided by `ResTemplateMetricsAutoConfiguration`, hence if `RestTemplateMetricsAutoConfiguration` is excluded or `RestTemplate` class is not available in the classpath then `URI` tag filter will not be registered (this affects on` WebClient`).

This commit provides a new class `WebMetricsClientFiltersAutoConfiguration` for register `URI` tag filter if either `RestTemplate` or `WebClient` is available.


snicoll:
Thanks for the PR @nosan but I've decided to fix this in a different manner (see 7bee9df and #14269).

PR #14278, opened 2018-09-01
Documentation update for default compressible MIME types

kedar-joshi:
Currently Spring Boot documentation states that -

> By default, responses are compressed only if their content type is one of the following:
> 
> * `text/html`
> * `text/xml`
> * `text/plain`
> * `text/css`

Commits fff2804 and 70dabdb added few more MIME types to the list. This PR aligns the documentation with the changes in those commits.


bclozel:
Thanks @kedar-joshi !

PR #14284, opened 2018-09-02
Fix indent of Jolokia maven dependency declaration in doc

PascalSchumacher:
(obvious fix)

A superfluous blank corrupts the formatting, see:
 
https://docs.spring.io/spring-boot/docs/current/reference/html/production-ready-jmx.html#production-ready-jolokia


snicoll:
Good catch @PascalSchumacher, thank you!

PR #14291, opened 2018-09-03
Use InvalidEndpointRequestException for MetricsEndpoint.metric()

izeye:
This PR changes to use `InvalidEndpointRequestException` for `MetricsEndpoint.metric()` instead of `Assert` as the `Assert` will end up with responding with 500 for invalid requests, not 400.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
@@ -34,11 +34,11 @@ import io.micrometer.core.instrument.Statistic;
 import io.micrometer.core.instrument.Tag;
 import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
 
+import org.springframework.boot.actuate.endpoint.InvalidEndpointRequestException;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
```diff
@@ -80,8 +80,12 @@ private String getName(Meter meter) {
 	@ReadOperation
 	public MetricResponse metric(@Selector String requiredMetricName,
 			@Nullable List<String> tag) {
-		Assert.isTrue(tag == null || tag.stream().allMatch((t) -> t.contains(":")),
-				"Each tag parameter must be in the form key:value");
+		if (tag != null && tag.stream().anyMatch((t) -> !t.contains(":"))) {
```
snicoll: Perhaps we should move that to the `parseTag` method and add a test for it? the existing code that checks there is `:` in the String is a bit weak as well.

izeye: @snicoll Thanks for the quick feedback! I updated as you suggested.

snicoll:
Looks good @izeye, thanks again!

Issue #13524, opened 2018-06-19
Make it possible to disable the BackgroundPreinitializer


dsyer:
E.g. if you know you are running in a single CPU, there's no point trying to do it in the background. It's just cost. It's more common than you might expect because container management platforms all try to restrict CPU access, and even hypervisors, potentially.

If you know that you are running an app in that kind of environment, it's pointless creating extra threads on startup, and there is a cost, so it would be useful to switch it off.

Also just to be able to measure the effect of the initializer accurately it would help to be able to disable it.

philwebb:
It would have to be quite low level since that thing starts so early. Would an environment variable be enough?

dsyer:
Yeah that's what I was thinking (environment variable or system property).

ayudovin:
@philwebb, As far as I understand we should check if property disable we don't need to start extra thread.
This property should located in Environment variable but on this stage environment variable isn't available yet. Does it mean `system.getProperty()` should use in this case ?
Correct me if I'm wrong.

snicoll:
@ayudovin you are correct. We can't use the `Environment` at this point so `System.getProperty` is a good idea (and relaxed binding can't really apply there). I think we should probably also discuss the name of this property as it's slightly unusual.

ayudovin:
@snicoll, what do you think about `spring.background-preinitializer.extra-thread.enable` ?

philwebb:
Take a look at `CachedIntrospectionResults.IGNORE_BEANINFO_PROPERTY_NAME` from Spring Framework. That's a pretty similar idea.

ayudovin:
@snicoll `spring.backgroundpreinitializer.ignore` ?

philwebb:
@ayudovin Sounds good to me, we can easily change it later if necessary.

snicoll:
Ditto, thanks @ayudovin!

snicoll:
Closing in favour of PR #14298

PR #14298, opened 2018-09-04
Make it possible to disable the BackgroundPreinitializer

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/13524)

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
@@ -1164,6 +1164,7 @@ content into your application. Rather, pick only the properties that you need.
 	spring.rabbitmq.publisher-confirms=false # Whether to enable publisher confirms.
 	spring.rabbitmq.publisher-returns=false # Whether to enable publisher returns.
 	spring.rabbitmq.requested-heartbeat= # Requested heartbeat timeout; zero for none. If a duration suffix is not specified, seconds will be used.
+	spring.rabbitmq.ssl.algorithm= # SSL algorithm to use. By default, configured by the Rabbit client library.
 	spring.rabbitmq.ssl.enabled=false # Whether to enable SSL support.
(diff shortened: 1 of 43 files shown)


philwebb:
Thanks for the PR. I think we can simplify this by bypassing all initialization if the property is set. The initialization code doesn't actually need to run at all. The reason we currently do it is to kick off in the background so that something called later runs quickly.

Probably just returning early from [this method](https://github.com/spring-projects/spring-boot/blob/2af8a9617b4a67b41fd1fcdfc98a86548af01596/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/BackgroundPreinitializer.java#L58) if the property is set will be enough.

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
```diff
@@ -69,6 +69,7 @@ content into your application. Rather, pick only the properties that you need.
 
 	# SPRING CORE
 	spring.beaninfo.ignore=true # Whether to skip search of BeanInfo classes.
+    spring.backgroundpreinitializer.ignore=true # Whether background preinitializer should be ignored.
```
snicoll: We shouldn't add this here as it's not managed the same way as a regular property. Let's document the property and its usage on the class Javadoc instead.

snicoll:
Thanks @ayudovin, this is now merged in `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/304c6a02ba9a526f9a2bbbd944ba47a4003f874c).

ayudovin:
@snicoll, Thank you!

gdespres:
Hi, 

Boolean.getBoolean(IGNORE_BACKGROUNDPREINITIALIZER_PROPERTY_NAME) is always false.
The call to the method System.getProperty is missing. You can confirm it ?

dsyer:
Have a look at the java docs for `Boolean.getBoolean()`.

gdespres:
Thanks @dsyer, it's a misunderstanding of this method. 

I didn't known this feature.

Issue #14299, opened 2018-09-04
Allow adding of details using maps with Health.Builder


michael-pratt:
In some cases I would like to add a map of details when using `Health.Builder`. Currently the builder class only allows for adding details one at a time after object construction, and it would be helpful if I could add an entire map of details, some thing like:

    Health.Builder health = Health.unknown();
    Map<String, Object> details = ...; // grab a map of details from somewhere
    health.withDetails(details);

I assume the new `withDetails(Map<String, Object>)` method could just delegate to `Map.putAll(Map<String, Object>)` under the hood, and the semantics of overwriting existing keys is the same as if I called `withDetail(String, Object)` multiple times.

A lot of my health indicators pull multiple maps from other sources, so I'd rather not have to combine them into a single map that is known prior to the creation of the `Health` object.

I'm happy to submit a PR for this if it helps.

philwebb:
Seems like a fine addition to me and a pull-request would be most welcome. The incoming map should probably be `Map<String, ? extends Object>` so that you can pass in things like `Map<String, String>` if needed.

In the meantime, if you're on Java 8 you can do the following:

```java
map.forEach(healthBuilder::withDetail);
```

michael-pratt:
Thanks Phil, I'll code it up and submit a PR when I have a moment.

michael-pratt:
@philwebb PR has been submitted: #14305

One note: I made the new method signature:

    public Builder withDetails(Map<String, ?> details)

The `Map<String, ?>` parameter matches the existing constructor:

    public Builder(Status status, Map<String, ?> details)

wilkinsona:
Closing in favour of PR #14305.

PR #14305, opened 2018-09-05
Add Health details using maps

michael-pratt:
Added `Health.Builder.withDetails(Map<String, ?>)` to allow adding entire maps at once instead of calling `withDetails(String, Object)` for each entry. Also includes new unit tests.

See issue #14299


snicoll:
@michael-pratt thank you for making your first contribution to Spring Boot. This is now merged on `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/ca8be3f6bd40ca3353ba3fc7d2e634eb853e2941).

PR #14353, opened 2018-09-08
Add auto-config for Prometheus Pushgateway. Fixes #14346

### LINKED ISSUE #14346: Support Prometheus Pushgateway
It would be nice to bring the Prometheus push-gateway functionality that was implemented in micrometer-legacy into spring boot 2.x. See
* https://github.com/micrometer-metrics/micrometer/pull/394
* https://github.com/micrometer-metrics/micrometer/issues/830
* https://stackoverflow.com/questions/49779684/how-to-export-metrics-to-prometheus-via-pushgateway-using-spring-boot-2-0

The push-gateway is handy for batch-programs (e.g. non-longrunning processes)

davidkarlsen:
Fixes #14346 - adds support for Prometheus Pushgateway.
Basically a backport of master of https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-spring-legacy/src/main/java/io/micrometer/spring/autoconfigure/export/prometheus/PrometheusMetricsExportAutoConfiguration.java

Take note of the added dependency on the pushgateway library, care needs to be taken to make it harmonize with the transitive dependencies on other Prometheus libraries through io.micrometer:micrometer-registry-prometheus .

@snicoll 
Can you take a quick look and give me feedback on anything that needs fixing?

---
This change is [Reviewable](https://reviewable.io/reviews/spring-projects/spring-boot/14353)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
```diff
@@ -86,4 +97,109 @@ public PrometheusScrapeEndpoint prometheusEndpoint(
...
 
+	/**
+	 * Configuration for <a href="https://github.com/prometheus/pushgateway">Prometheus
+	 * Pushgateway</a>.
+	 *
+	 * @author David J. M. Karlsen
+	 */
+	@Configuration
+	@ConditionalOnClass(PushGateway.class)
+	@ConditionalOnProperty(prefix = "management.metrics.export.prometheus.pushgateway", name = "enabled", havingValue = "true", matchIfMissing = false)
```
philwebb: I think you can drop `havingValue` and `matchIfMissing` from this line. The defaults should suffice.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
```diff
@@ -86,4 +97,109 @@ public PrometheusScrapeEndpoint prometheusEndpoint(
...
+
+			public PushGatewayHandler(CollectorRegistry collectorRegistry,
+					PrometheusProperties prometheusProperties, Environment environment) {
+				this.collectorRegistry = collectorRegistry;
+				this.pushgatewayProperties = prometheusProperties.getPushgateway();
+				this.pushGateway = new PushGateway(
+						this.pushgatewayProperties.getBaseUrl());
+				this.environment = environment;
+				this.executorService = Executors.newSingleThreadScheduledExecutor((r) -> {
+					final Thread thread = new Thread(r);
```
philwebb: You can drop `final` from here, we generally don't use it for local variables.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
```diff
@@ -86,4 +97,109 @@ public PrometheusScrapeEndpoint prometheusEndpoint(
...
+	public static class PrometheusPushGatewayConfiguration {
+
+		@Bean
+		public PushGatewayHandler pushGatewayHandler(CollectorRegistry collectorRegistry,
+				PrometheusProperties prometheusProperties, Environment environment) {
+			return new PushGatewayHandler(collectorRegistry, prometheusProperties,
+					environment);
+		}
+
+		public static class PushGatewayHandler {
```
philwebb: I think this should be a private or package private class. It seems like we wouldn't expect people to use it directly.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
```diff
@@ -86,4 +97,109 @@ public PrometheusScrapeEndpoint prometheusEndpoint(
...
+			}
+
+			@PreDestroy
+			void shutdown() {
+				this.executorService.shutdown();
+				if (this.pushgatewayProperties.isPushOnShutdown()) {
+					push();
+				}
+				if (this.pushgatewayProperties.isDeleteOnShutdown()) {
+					try {
```
philwebb: Perhaps extract a `delete()` method here?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
```diff
@@ -86,4 +97,109 @@ public PrometheusScrapeEndpoint prometheusEndpoint(
...
+					}
+					catch (Throwable throwable) {
+						this.logger.error(
+								"Unable to delete metrics from Prometheus Pushgateway",
+								throwable);
+					}
+				}
+			}
+
+			private String job() {
```
philwebb: I'd be tempted to rename this `getJobName()`

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
```diff
@@ -128,6 +129,18 @@ public void allowsCustomScrapeEndpointToBeUsed() {
 						.hasSingleBean(PrometheusScrapeEndpoint.class));
 	}
```
philwebb: It would be nice if we could test more of the logic in `PushGatewayHandler` somehow. There's logic in the `push` and `shutdown` methods that we don't really have covered. I'm not totally sure how we can do that, perhaps splitting the class so the logic is separate from the scheduling and then testing the logic part with mocks?

Don't worry too much about that if you can't find a way to do it, we can look to it when we merge.

spring-boot-project/spring-boot-dependencies/pom.xml
```diff
@@ -140,6 +140,8 @@
 		<nio-multipart-parser.version>1.1.0</nio-multipart-parser.version>
 		<pooled-jms-version>1.0.3</pooled-jms-version>
 		<postgresql.version>42.2.5</postgresql.version>
+		<!-- need to take care that this version harmonizes with micrometer ones -->
```
philwebb: I think we have a maven plugin enabled that takes care of that. We should drop the comment in the POM regardless because we generally avoid them in the POM files.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
davidkarlsen: Fixed all comments except this one

philwebb: No worries. Thanks for your efforts, we'll try to find a way to add some additional tests when we merge it.

davidkarlsen: Great - I hope it will go into 2.1.0?

davidkarlsen: @philwebb WDYT?

davidkarlsen: ah - I see the added milestones now!

philwebb:
@davidkarlsen Thanks very much for contributing this. I've merged it into 2.1 with [some additional changes](https://github.com/spring-projects/spring-boot/commit/20ecf73cd156e0f6e6cd75c1979ae8925d25a94a). I'm not that familiar with the Pushgateway, so if anything looks wrong please let me know.

One specific change I want to point out is replacing the boolean flags with a `shutdown-operation` property. I assume that a user would never want to do both a push and a delete on shutdown. If this assumption is wrong, let me know.

Thanks again!

davidkarlsen:
@philwebb Thanks a lot for merging! It looks good.
Some background on why there is a delete at all: https://prometheus.io/docs/practices/pushing/ (bullet no. 3).

If it would make sense to do push and then delete depends on how this is implemented in pushgateway. All I know is that the delete is async and does not happen immediately: https://www.robustperception.io/common-pitfalls-when-using-the-pushgateway - if it is implemented so that it would allow for a last scrape to happen, THEN delete - it would make sense. If that would ever be the case then PUSH_AND_DELETE could be added to handle that. It would nevertheless be a corner-case - let's leave it easy and clean for now.

Thanks again for your effort - appreciated!

danielsuter:
Is there an example of how to use this functionality?

philwebb:
@danielsuter I don't think we every updated the docs. I've raised #16853.

Issue #11013, opened 2017-11-14
ConversionService used for Actuator endpoints should be configurable


shabino:
Because the `EndpointAutoConfiguration` `endpointOperationParameterMapper()` method is not annotated by `@ConditionalOnMissingBean(ParameterMapper.class)`, I had to resort to redundantly creating my own `ConversionServiceParameterMapper` (with my own, customized conversion service) with the `@Primary` annotation. This solved my problem, but now my code includes this line from `ConversionServiceParameterMapper`: `IsoOffsetDateTimeConverter.registerConverter(conversionService);` This isn't great because I now must remember to check the code in `ConversionServiceParameterMapper` when upgrading to newer versions of Spring Boot to make sure I don't need to configure my own `ConversionService` differently due to internal changes.  And, it's slightly sloppy because there's an extra `ParameterMapper` created for no reason.

I'd like for there to be a clean way to customize the Actuator-specific `ConversionService` used as the `ParameterMapper`.

snicoll:
@shabino thanks for the report. Can you share why you need a specific `ConversionService` for this? What is the use case?

shabino:
We use interned value object wrappers for identifiers. They wrap the raw String representations. By adding my String -> ID converter (works via reflection for all of our identifier classes of this style), I get more strongly-typed method signatures in my Endpoint classes. Not a huge benefit -- just a nice-to-have.

snicoll:
@shabino okay cool. I propose that we look for `Converter` and `GenerciConverter`  beans in the context and just add them to the `ConversionService`. Would that work for you?

shabino:
This would solve my problem.

What would happen if someone has his own String -> Date converter in his context? I presume that IsoOffsetDateTimeConverter would win.

arnokoehler:
Is there a setup for this available, for if I want to look in to this issue?

shabino:
My issue was partially fixed in this commit which added the `@ConditionalOnMissingBean` annotation: https://github.com/spring-projects/spring-boot/commit/1d39feffeabf4b94e3f900b941b58e0443d5887a#diff-5b71e8f38ba31c03d127203b2f8c0172

I look forward to the PR from @chang-chao nonetheless :)

philwebb:
Closing in favor of PR #16449

PR #14446, opened 2018-09-12
conversionService used for Actuator endpoints should be configurable

ayudovin:
I'm not sure that I absolutely understand goal of this [enhancement](https://github.com/spring-projects/spring-boot/issues/11013), but maybe such solution will be satisfactory.


wilkinsona:
Thanks for the PR, @ayudovin. Judging by [this comment](
https://github.com/spring-projects/spring-boot/issues/11013#issuecomment-344173992), I'm not sure this is quite what @snicoll had in mind. He's only holiday at the moment so we'll have to wait a week or so to get his input.

snicoll:
Thanks for the PR. Andy is right, that's not really what I have in mind. I remember that looking for `Converter` beans in the context would have been an effective way to make sure this works properly but it's been a while and maybe I am wrong. Perhaps you could start with a test that fails the use case described in the related issue and try to associate converters from the context? I'd rather do this in a generic manner rather than focusing on a particular converter.

ayudovin:
@snicoll, i'm not sure I understand how i can help you. Do you what I begin to work on fails tests in this pull request ?

ayudovin:
@snicoll, ok, i just close this pull request on current moment if you don't mind

PR #14460, opened 2018-09-13
Remove duplicate -l option for init command

### LINKED ISSUE #13940: Offer --list in the tab completion for spring init
`spring --version` gives me: Spring CLI v2.0.3.RELEASE

I have installed the bash completion script and if I do `spring init --<tab><tab>` I get:
```
--artifactId    
--boot-version
--build
--dependencies
--description 
--extract
--force
--format
--groupId
--java-version
--language
--name
--package-name
--packaging
--target
--type
--version
```
But the `--list` option is missing.

ruslanys:
`--list` option should be displayed by default out-of-the-box. But `InitCommand` had 2 options with `-l` key: `-l` for `--language` and `-l` for `--list`. In that case, when `OptionHelpFormatter.format(...)` pushing options into TreeSet, the `--list` option disappears.

### DIFF
--- a/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
+++ b/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
@@ -127,7 +127,7 @@ public class InitCommand extends OptionParsingCommand {
 			this.target = option(Arrays.asList("target"), "URL of the service to use")
 					.withRequiredArg()
 					.defaultsTo(ProjectGenerationRequest.DEFAULT_SERVICE_URL);
-			this.listCapabilities = option(Arrays.asList("list", "l"),
+			this.listCapabilities = option(Arrays.asList("list"),
 					"List the capabilities of the service. Use it to discover the "
 							+ "dependencies and the types that are available");
 			projectGenerationOptions();


pivotal-issuemaster:
@ruslanys Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14460)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14460) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ruslanys Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14460)!

philwebb:
Very interesting that the duplicate causes it to just disappear. I wonder if we can force the app to fail hard when that happens?

ruslanys:
@philwebb I agree this is strange behavior. 
Unfortunately, in the CLI context, it's not clear when can we check options uniqueness.
In the current implementation, Spring Boot CLI is initializing command options when a command choose.
That means that we will be able to throw an exception only when user types `spring init`.

Moreover, for options recognizing Spring Boot CLI is using `joptsimple` library, especially `OptionParser`.

So, one of the ideas is to update the `getParser()` method and put validation under [`options()`](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/options/OptionHandler.java#L68) invocation.
Like:
```java
private void checkOptionsUniqueness() {
	List<String> keys = parser.recognizedOptions().values().stream()
			.distinct()
			.flatMap(optionSpec -> optionSpec.options().stream())
			.collect(Collectors.toList());

	Set<String> hashSet = new HashSet<>();
	for (String key : keys) {
		if (!hashSet.add(key)) {
			throw new IllegalStateException("Option key [" + key + "] is duplicated.");
		}
	}
}
```

I can provide a separate PR for this.

ruslanys:
Unfortunately, the example above won't fix all problems. 
If somebody put several exactly the same keys, for instance, "list" for several options, then `OptionsParser` won't provide us with the actual list of options and `parser.recognizedOptions()` will return only one of the options.

ruslanys:
The second approach is to validate option name and aliases at `OptionHandler.option(...)` method, but it will be possible to get parser directly and register option out of the `OptionHandler` class.

ruslanys:
@philwebb Please take a look at #14466.

snicoll:
Closed by 6e6c22c

snicoll:
@ruslanys thank you very much for making your first contribution to Spring Boot.

PR #14466, opened 2018-09-13
Option key validation on uniqueness

ruslanys:
This PR is adding validation to prevent cases described in #14460.


ruslanys:
So, when a user executes `spring init` from the #14460 case, he will receive: 
`Option key -l duplicated.` message error.

`InitCommandTests` will fail as well.

ruslanys:
If this PR is accepted, it must be merged after # 14460, otherwise, the build will fail.

PR #14471, opened 2018-09-14
Auto-configure jOOQ to use an ExecutorProvider bean if available

davinkevin:
This improvement allow to define a custom Executor for Jooq. To provide it, you have to wrap it under a `org.jooq.ExecutorProvider`. This is useful in multiple case, especially for testing in my case when I want to run my `@Sql()` test inside a transaction with `fetchAsync` which wasn't compatible. Thanks to this, we can in our test use the `SyncTaskExecutor` to use the `main` thread for all our tests.

BTW, I've made some code modification in another commit (not in this PR) to simplify (from my point of view) the configuration system (using the `ObjectProvider` instead of relying on null). Do you think I can make a PR of this (https://github.com/davinkevin/spring-boot/commit/bf4e36c0d077664d6ae8832ff06974ad91ab559e) ?

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jooq/JooqAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jooq/JooqAutoConfiguration.java
@@ -21,6 +21,7 @@ import javax.sql.DataSource;
 import org.jooq.ConnectionProvider;
 import org.jooq.DSLContext;
 import org.jooq.ExecuteListenerProvider;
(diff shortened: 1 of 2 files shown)


philwebb:
Thanks for the PR. I'm not totally sold on https://github.com/davinkevin/spring-boot/commit/bf4e36c0d077664d6ae8832ff06974ad91ab559e, I wonder if using `PropertyMapper` might be better.

wilkinsona:
@davinkevin Thank you very much for your contribution. The proposed changes have now been merged into master. I also applied a small [polishing commit](
https://github.com/spring-projects/spring-boot/commit/a1d1a7392fae31ad6cf6b408274e5c3f9decb6ad), the main part of which was a small update to the documentation.

Issue #14467, opened 2018-09-13
Use ObjectProvider<T> with ordered list access instead of ObjectProvider<List<T>>


sdeleuze:
As discussed in [SPR-17272](https://jira.spring.io/browse/SPR-17272) (which is now fixed on Spring Framework `master`), `ObjectProvider` now provides ordered list access, superseding `ObjectProvider<List>`.

In addition to follow best practices, updating Boot auto-configuration classes to use `ObjectProvider<T>` parameters with ordered list access instead of `ObjectProvider<List<T>>` would make it usable with `BeanFactory.getBeanProvider()`, required with functional bean registration like in [Kofu configuration](https://github.com/spring-projects/spring-fu#kofu-configuration) or @dsyer work in [spring-boot-micro-apps](https://github.com/dsyer/spring-boot-micro-apps).

wilkinsona:
I've got an attempt at this in [this branch](https://github.com/wilkinsona/spring-boot/tree/object-provider). Unfortunately, `CustomHibernateJpaAutoConfigurationTests.hibernatePropertiesCustomizersAreAppliedInOrder()` is currently failing due to the customisers being applied in the wrong order. The customizers are being retrieved using the following code:

```
hibernatePropertiesCustomizers.orderedStream().collect(Collectors.toList())
```

`hibernatePropertiesCustomizers` is an `ObjectProvider<HibernatePropertiesCustomizer>`. Injecting `ObjectProvider<List<HibernatePropertiesCustomizer>>` into the same place and calling `getIfAvailable()` produces a list with the same two customisers but in the reverse order. This looks like a Framework bug but I've yet to reproduce it in a standalone test. /cc @jhoeller for awareness just now. I'll dig some more tomorrow.

jhoeller:
Let's try to sort this out tomorrow morning if we can. This is important enough for 5.1 GA still.

wilkinsona:
The problem's occurring because the ordering information isn't available. It isn't available as the order source provider doesn't have the right information. In `resolveMultipleBeans`, the matching beans returned by `findAutowireCandidates(String, Class<?>, DependencyDescriptor)` is a `Map<String, Class>` with the following contents:

```
{sampleCustomizer=interface org.springframework.boot.autoconfigure.orm.jpa.HibernatePropertiesCustomizer, anotherCustomizer=interface org.springframework.boot.autoconfigure.orm.jpa.HibernatePropertiesCustomizer}
```

[...]

PR #14472, opened 2018-09-14
Using ObjectProvider<T> with ordered list access instead of ObjectProvider<List<T>>

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14467)


wilkinsona:
Thanks for the PR, but `ObjectProvider` has be re-worked in the latest Framework snapshots (which is what prompted the opening of #14467. We also want to look at each current usage in turn and decide the best method to use in each case. Sorry for the wasted effort on your part. I'd recommend commenting on an issue before you start working on it so that we can let you know about this sort of thing in advance.

ayudovin:
@wilkinsona, ok, it's my mistake, sorry.
Thank you for advice.

Issue #14469, opened 2018-09-14
Add ApplicationContextInitializedEvent to correspond to contextPrepared in SpringApplicationRunListener


dsyer:
`EventPublishingRunListener` has a `contextPrepared()` method that is fired after the context is prepared but before any bean definitions are loaded. But there is no corresponding `AplicationEvent`. To listen for that callback the user can implement the whole of `SpringApplicationRunListener`, but it would be natural to expect there would be an `ApplicationEvent`.

dsyer:
Actually it's not even possible to do this with a `SpringApplicationRunListener` unless I miss something (at least not from `spring.factories`), because the `SpringApplication` instance is not available to the vanilla interface.

wilkinsona:
Any suggestions for a name for the new event. Here's its position among the others we already have:

1. `ApplicationStartingEvent`
2. `ApplicationEnvironmentPreparedEvent`
3. 
4. `ApplicationPreparedEvent`
5. `ApplicationStartedEvent`
6. `ApplicationReadyEvent`

`ApplicationContextPreparedEvent` is the best I've thought of thus far.

dsyer:
Or just `ContextPreparedEvent`? Maybe too close to Spring Framework naming convention.

I think `ApplicationContextPreparedEvent` is too hard to distinguish from `ApplicationPreparedEvent`.

Maybe `ApplicationContextRefreshingEvent`? It is called just before the context is refreshed, I think.

wilkinsona:
> It is called just before the context is refreshed, I think.

Nope, that's when `ApplicationPreparedEvent` is published as a result of `contextLoaded()` being called.

`contextPrepared()` is called once the application context has been created but its sources haven't been loaded.

dsyer:
OK, how about `ApplicationContextInitializedEvent` (it happens just after the `ApplicationContextInitializers` are applied?

philwebb:
I quite like that name

mbhave:
Closing in favor of #14478.

PR #14478, opened 2018-09-14
Add ApplicationContextInitializedEvent to correspond to contextPrepared in SpringApplicationRunListener

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14469)

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/event/ApplicationContextInitializedEvent.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/event/ApplicationContextInitializedEvent.java
@@ -1,0 +1,54 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.context.event;
+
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/event/ApplicationContextInitializedEvent.java
```diff
@@ -0,0 +1,54 @@
+/*
+ * Copyright 2012-2017 the original author or authors.
```
wilkinsona: This should be 2012-2018.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/event/ApplicationContextInitializedEvent.java
```diff
@@ -0,0 +1,54 @@
...
+ */
+
+package org.springframework.boot.context.event;
+
+import org.springframework.boot.SpringApplication;
+import org.springframework.context.ApplicationContext;
+import org.springframework.context.ConfigurableApplicationContext;
+
+/**
+ * Event published as when a {@link SpringApplication} is starting up and the
```
wilkinsona: I think this typo is in some of the other events' javadoc, but the "as" in "Event published as when a" is not needed here.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/event/ApplicationContextInitializedEvent.java
```diff
@@ -0,0 +1,54 @@
...
+ * {@link ApplicationContext} is prepared but before any bean definitions are loaded.
```
wilkinsona: I think it'd be good to mention that application context initialisers will have been called, something like:

ApplicationContext is prepared and ApplicationContextInitializers have been called but before any bean definitions are loaded.

ayudovin:
@wilkinsona, Thank for your comment, I've added changes by your suggestion.

wilkinsona:
Thanks once again, @ayudovin. The proposed changes are now in master.

ayudovin:
@wilkinsona, Thank you.

Issue #13801, opened 2018-07-17
New Tag on WebMvc Metrics To Distinguish Status Code Group


michaelmcfadyen:
_This effects spring boot actuator and specifically it's metrics classes._

**Situation**
I want to graph a count of calls made to an api that returned a 4xx status code.

Some monitoring systems ie. Datadog do not allow wildcards when selecting tag values.

It would be useful to have a tag on those metrics that would group the response codes appropriately. For example these groupings could be:
- 2xx -> success
- 4xx -> client_error
- 5xx -> server_error

So all api calls that returned a status code in the 4xx range would all have a tag "outcome: client_error"

wilkinsona:
@michaelmcfadyen Just in case you aren't already aware, you can do this in your own code by providing a custom `WebMvcTagsProvider` or `WebFluxTagsProvider` bean.

Let's get some feedback from @jkschneider and we can then decide if this is something that we want Boot to offer out of the box.

jkschneider:
I think we should. It's difficult on Prometheus and Atlas too. In fact, I don't know yet of a monitoring system for which we can select tags on `response >= 400`. It's always something like `response ~= [45]\d{2}`, which is just kinda ugly.

I was struggling to think of a tag key, but "outcome" is a clever choice.

mveitas:
One other thing to consider is the option to filter out certain codes from being included in this metric. For example in an API it might be ok to return back 404 as a "valid" status code and including these in the "client_error" metric might not be desirable.

I might suggest that the above is the default implementation, but it's left open to allow a developer to override and provide their own "outcome" bucketing strategy

jkschneider:
Good point @mveitas. I imagine this would come with a default implementation, much like `WebMvcTags#status` [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcTags.java#L71).

You'd be free to do your own thing if you provide a custom implementation of `WebMvcTagsProvider/WebFluxTagsProvider`, possibly delegating to the default implementations for everything other than what you want to change.

wilkinsona:
Closing in favour of #14486.

PR #14486, opened 2018-09-16
Add outcome tag to web mvc servlet and reactive metrics. Fixes gh-13801.

michaelmcfadyen:
Add a new tag "outcome" to web mvc metrics for both servlet and reactive models. The outcome tag can have 4 possible values based on the http status code of the response.

1xx, 2xx, 3xx -> SUCCESS
4xx -> CLIENT_ERROR
5xx -> SERVER_ERROR
no status code -> UNKNOWN

Related Issue: #13801

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/reactive/server/WebFluxTags.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/reactive/server/WebFluxTags.java
@@ -30,6 +30,7 @@ import org.springframework.web.util.pattern.PathPattern;
  *
  * @author Jon Schneider
  * @author Andy Wilkinson
+ * @author Michael McFadyen
  * @since 2.0.0
  */
 public final class WebFluxTags {
@@ -43,6 +44,14 @@ public final class WebFluxTags {
 	private static final Tag URI_UNKNOWN = Tag.of("uri", "UNKNOWN");
 
 	private static final Tag EXCEPTION_NONE = Tag.of("exception", "None");
+
+	private static final Tag OUTCOME_UNKNOWN = Tag.of("outcome", "UNKNOWN");
+
+	private static final Tag OUTCOME_SUCCESS = Tag.of("outcome", "SUCCESS");
+
+	private static final Tag OUTCOME_CLIENT_ERROR = Tag.of("outcome", "CLIENT_ERROR");
+
+	private static final Tag OUTCOME_SERVER_ERROR = Tag.of("outcome", "SERVER_ERROR");
 
 	private WebFluxTags() {
 	}
@@ -60,7 +69,7 @@ public final class WebFluxTags {
 	}
 
 	/**
-	 * Creates a {@code method} tag based on the response status of the given
+	 * Creates a {@code status} tag based on the response status of the given
 	 * {@code exchange}.
 	 * @param exchange the exchange
 	 * @return the "status" tag derived from the response status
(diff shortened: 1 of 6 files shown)


mveitas:
It would be great to give control to developer to be able to override the default bucketing strategy in someway. In some of the APIs we have, 404 is considered to be a successful outcome.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/reactive/server/WebFluxTags.java
```diff
@@ -117,4 +120,31 @@ public static Tag exception(Throwable exception) {
...
+	 * @param exchange the exchange
+	 * @return the "outcome" tag derived from the response status
+	 */
+	public static Tag outcome(ServerWebExchange exchange) {
+		if (exchange != null && exchange.getResponse().getStatusCode() != null) {
+			HttpStatus status = exchange.getResponse().getStatusCode();
+			String outcome;
+			if (status.is1xxInformational() || status.is2xxSuccessful()
+					|| status.is3xxRedirection()) {
+				outcome = "SUCCESS";
```
mveitas: These should be constant values or an enum since they are used in multiple places

mveitas: Should these outcome values be treated the same as `OUTCOME_UNKNOWN` and made into a constant value to avoid the creation for each request?

wilkinsona: > These should be constant values or an enum since they are used in multiple places

Perhaps. In terms of main code, they're used once on the MVC side and once on the WebFlux side. Sometimes duplication is preferable to creating an artificial construct that would allow them to be shared.

wilkinsona: > Should these outcome values be treated the same as OUTCOME_UNKNOWN and made into a constant value to avoid the creation for each request?

Yes, I think so. While the strings themselves aren't a problem, it would be better to avoid creating a new `Tag` each time.

@michaelmcfadyen Thank you for the pull request. Would you like to make this amendment?

michaelmcfadyen:
> It would be great to give control to developer to be able to override the default bucketing strategy in someway. In some of the APIs we have, 404 is considered to be a successful outcome.

You can override how the outcome tag is created by providing your own implementation of `WebMvcTagsProvider`.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/reactive/server/WebFluxTags.java
michaelmcfadyen: sure, will update the PR later today

michaelmcfadyen: updated PR to use constants for each outcome tag

wilkinsona:
@michaelmcfadyen I am a bit confused by them too. I agree that they're unrelated to these changes. Please just ignore them. I think we have all we need now to be able to merge this. Thanks again for the PR.

wilkinsona:
@michaelmcfadyen Thank you very much for making your first contribution to Spring Boot. The proposed changes have now been merged into master. I also applied [a little bit of polish](
https://github.com/spring-projects/spring-boot/commit/c974192497e2ebc9594756d163131bed51925122), the main parts of which were to update the documentation and to divide the success outcome up to separate out informational and redirection.

Issue #14066, opened 2018-08-14
Metric endpoint is summing up same metric


pulkitmehra:
We are using Spring Boot 2.0.3 actuators.

We are using `MetricAutoConfiguration` to use the Micrometer global `CompositeMeterRegistry` and we have `StatsdMeterRegistry` and  `JmxMeterRegistry` both in the classpath. We are seeing a count error in the Spring Boot actuator `MetricsEndpoint` where the value for a single meter counter is being doubled at: https://github.com/spring-projects/spring-boot/blob/80da9cf5ebdc256c0b743fd0c34c451a368530aa/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java#L135 

For example, we have a metric for keeping track of `200` HTTP responses and the metric is being added to both `StepMeterRegisty` objects. We would expect this as the value has been added to the global `CompositeMeterRegistry` which would be sent to each `StepMeterRegistry` for publishing. But why is the default behavior in the `MetricsEndpoint` to sum these values and essentially double the count metric when the metric is the same?

wilkinsona:
The summing was introduced in https://github.com/spring-projects/spring-boot/commit/bc05352290dccdbd069c9df2a31d24a4ca3b7b18, and is intentional judging by the following portion of the commit message:

> Selecting a metric by name (and optionally a set of tags) reports statistics that are the sum of the statistics on all time series containing the name (and tags).

I'll defer to @jkschneider now as he knows way more about metrics than I do.

jkschneider:
@wilkinsona That is correct!

ryanpconnors:
Is it a common use case for applications to have multiple `StepMeterRegistries`? This seems to be the source of the issue.

pulkitmehra:
For example, using `CompositeMeterRegistry` we are synching the same metric into two different measurement platform systems say, Datadog and JMX. So essentially, these `StepMeterRegistry` will have the same metric and statistics. 
 
So, what `/metric` endpoint is doing is getting all the internal registries from `CompositeMeterRegistry` and querying the same metric name in each internal registry (here its Dtatadog and JMX), and get the same stats from them and summing them up. 

The way we understand, the `CompositeMeterRegistry` aggregates other meter registry and provides a single interface to collect metrics. So, most of the time we will find the same metric and stats in an internal registry. 

[...]

PR #14497, opened 2018-09-17
Metric endpoint is summing up same metric

pulkitmehra:
This PR fixed issue https://github.com/spring-projects/spring-boot/issues/14066
- Recursively looking for matching metrics in nested MeterRegistries and return the first found
- Added test cases.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
@@ -83,7 +83,7 @@ public class MetricsEndpoint {
 			@Nullable List<String> tag) {
 		List<Tag> tags = parseTags(tag);
 		List<Meter> meters = new ArrayList<>();
-		if (!findFirstMatchingMeter(meters, this.registry, requiredMetricName, tags)) {
+		if (!findFirstMatchingMeters(meters, this.registry, requiredMetricName, tags)) {
 			return null;
 		}
 		Map<Statistic, Double> samples = getSamples(meters);
@@ -112,11 +112,11 @@ public class MetricsEndpoint {
 		return Tag.of(parts[0], parts[1]);
 	}
 
-	private boolean findFirstMatchingMeter(List<Meter> meters, MeterRegistry registry,
+	private boolean findFirstMatchingMeters(List<Meter> meters, MeterRegistry registry,
 			String name, Iterable<Tag> tags) {
 		if (registry instanceof CompositeMeterRegistry) {
 			return ((CompositeMeterRegistry) registry).getRegistries().stream()
-					.anyMatch((r) -> findFirstMatchingMeter(meters, r, name, tags));
+					.anyMatch((r) -> findFirstMatchingMeters(meters, r, name, tags));
 		}
 		else {
 			Collection<Meter> meterFound = registry.find(name).tags(tags).meters();
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
@@ -16,7 +16,6 @@
 
 package org.springframework.boot.actuate.metrics;
 
-import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
```diff
@@ -112,15 +112,20 @@ private Tag parseTag(String tag) {
...
 		if (registry instanceof CompositeMeterRegistry) {
-			((CompositeMeterRegistry) registry).getRegistries()
-					.forEach((member) -> collectMeters(meters, member, name, tags));
+			return ((CompositeMeterRegistry) registry).getRegistries().stream()
+					.anyMatch((r) -> findFirstMatchingMeters(meters, r, name, tags));
 		}
 		else {
-			meters.addAll(registry.find(name).tags(tags).meters());
+			Collection<Meter> meterFound = registry.find(name).tags(tags).meters();
+			if (meterFound != null && meterFound.size() > 0) {
```
jkschneider: Micrometer guarantees that `meterFound` is non-null. Could use `isEmpty()` on second condition.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
```diff
@@ -112,15 +112,20 @@ private Tag parseTag(String tag) {
 		return Tag.of(parts[0], parts[1]);
 	}
 
-	private void collectMeters(List<Meter> meters, MeterRegistry registry, String name,
-			Iterable<Tag> tags) {
+	private boolean findFirstMatchingMeters(List<Meter> meters, MeterRegistry registry,
```
jkschneider: Rather than modifying a parameter, this method could simply return a `Collection`.

pulkitmehra:
@jkschneider , please review.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java
```diff
@@ -112,15 +113,22 @@ private Tag parseTag(String tag) {
...
+					.map((r) -> findFirstMatchingMeters(r, name, tags))
+					.filter((match) -> !match.isEmpty()).findFirst()
+					.orElse(Collections.emptyList());
+
 		}
 		else {
-			meters.addAll(registry.find(name).tags(tags).meters());
+			Collection<Meter> metersFound = registry.find(name).tags(tags).meters();
+			if (!metersFound.isEmpty()) {
+				return new ArrayList<>(metersFound);
```
jkschneider: No need to wrap this. The list returned is already an immutable copy.

pulkitmehra: Thanks @jkschneider , the reason I am wrapping it into `List` is to avoid returning `Collection` from this method. 

`Collection<Meter> metersFound = registry.find(name).tags(tags).meters();`

If we return `Collection`, the `MetricEndpoint` methods should be refactored so that `Collection` could be pass around instead of `List` interface **OR** the return type should be downcasted to `List` (which is dangerous as we dont have control on `MeterRegistry#find`)

Changing `MetricEndpoint` methods to pass `Collection` is minor refactoring, with following highlight.

https://github.com/spring-projects/spring-boot/blob/13fef7d6cdaf5242c333f927dc0c2713823731ba/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/MetricsEndpoint.java#L93 

will be change to 

`meters.iterator().next().getId();`

Please confirm should I go ahead and do this minor change?

jkschneider: The minor refactoring seems preferable to me to generating extra garbage in this case, as minor as that will be.

pulkitmehra: Thanks @jkschneider, pls review.

pulkitmehra:
@philwebb , @jkschneider , the issue is closed but PR is still open. Is there anything l can do to help you merge it?

snicoll:
The issue has been closed because this PR supersedes it. All is good, we just need some time to review it.

philwebb:
@pulkitmehra Thanks for the fix. This is now in 2.0.x and master.

PR #14505, opened 2018-09-18
Dependency management for Caffeine is incomplete

marceloverdijk:
Please consider adding Caffeine jcache dependency


wilkinsona:
If we're going to expand our dependency management, we should add all of its modules. In addition to the `jcache` addition currently proposed here, the other missing modules are `guava` and `simulator`. @marceloverdijk Would you like to update this PR accordingly?

wilkinsona:
Thanks!

PR #14512, opened 2018-09-19
Use meaningful name for MockServerConfigurer bean

izeye:
It looks accidentally named as `get()`, so this PR renames it to `mockServerConfigurer()`.


snicoll:
Good catch again, thanks @izeye!

Issue #14510, opened 2018-09-19
Upgrade to Spring Data Lovelace-RELEASE


mp911de:

mp911de:
The following driver dependencies should be upgraded:

* Lettuce to 5.1.0.RELEASE
* MongoDB driver (sync) to 3.8.2
* Neo4j OGM to 3.1.3

PR #14515, opened 2018-09-19
Upgrade to Neo4j Ogm 3.1.3

michael-simons:
Neo4j OGM 3.1.3 will be the version of OGM used by Spring Data Neo4j Lovelace. See #14510.


michael-simons:
Thanks for letting us know.

PR #14518, opened 2018-09-19
Allow for additional customization in spring-boot-loader and spring-boot-loader-tools

handcraftedbits:
I'm working on an executable JAR format that utilizes spring-boot-loader-tools for creation and spring-boot-loader for initialization.  The built-in customization has allowed me to move move around libraries, create a custom runtime classloader, etc. but there are still two pieces of customization that I'd like to propose:

1. Allow for customization of the manifest before it is written.  There are some attributes I need to remove/add/etc. and today I have to let `Repackager.repackage` run and then crack open the JAR and rewrite the manifest to accomplish this.  The first commit in this PR simply adds a method called `Repackager.customizeManifest` which is run before the manifest is written and defaults to a no-op.
2. When unpacking nested archives, `JarFileArchive` creates a unique folder name of the form `<originalName>-spring-boot-libs-<UUID>`.  I'd like to have a way to customize this folder name, and to that end I've created a method called `JarFileArchive.getUniqueUnpackFolderFileName` that defaults to the current behavior.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java
@@ -314,7 +314,11 @@ public class Repackager {
 		if (StringUtils.hasLength(lib)) {
 			manifest.getMainAttributes().putValue(BOOT_LIB_ATTRIBUTE, lib);
 		}
+		customizeManifest(manifest);
 		return manifest;
+	}
+
+	protected void customizeManifest(Manifest manifest) {
 	}
 
 	private String findMainMethodWithTimeoutWarning(JarFile source) throws IOException {
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/test/java/org/springframework/boot/loader/tools/RepackagerTests.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/test/java/org/springframework/boot/loader/tools/RepackagerTests.java
@@ -664,6 +664,23 @@ public class RepackagerTests {
 		repackager.repackage(dest, NO_LIBRARIES);
 	}
 
+	@Test
+	public void customizedManifest() throws Exception {
+		this.testJarFile.addClass("a/b/C.class", ClassWithoutMainMethod.class);
+		File file = this.testJarFile.getFile();
+		Repackager repackager = new ManifestCustomizingRepackager(file);
+		repackager.setMainClass("a.b.C");
+		repackager.repackage(NO_LIBRARIES);
+		Manifest actualManifest = getManifest(file);
+		assertThat(actualManifest.getMainAttributes().getValue("Main-Class"))
+				.isEqualTo("org.springframework.boot.loader.JarLauncher");
+		assertThat(actualManifest.getMainAttributes().getValue("Start-Class"))
+				.isEqualTo("a.b.C");
+		assertThat(actualManifest.getMainAttributes().getValue("Test-Value"))
+				.isEqualTo("xyz");
+		assertThat(hasLauncherClasses(file)).isTrue();
+	}
+
 	private File createLibrary() throws IOException {
 		TestJarFile library = new TestJarFile(this.temporaryFolder);
 		library.addClass("com/example/library/Library.class",
@@ -732,6 +749,19 @@ public class RepackagerTests {
 		@Override
 		public void writeLoadedClasses(LoaderClassesWriter writer) throws IOException {
 			writer.writeEntry("test", new ByteArrayInputStream("test".getBytes()));
+		}
+
+	}
+
+	private static final class ManifestCustomizingRepackager extends Repackager {
+
+		private ManifestCustomizingRepackager(File source) {
+			super(source);
+		}
+
+		@Override
+		protected void customizeManifest(Manifest manifest) {
+			manifest.getMainAttributes().putValue("Test-Value", "xyz");
 		}
 
 	}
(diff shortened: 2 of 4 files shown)


pivotal-issuemaster:
@handcraftedbits Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14518)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14518) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@handcraftedbits Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14518)!

wilkinsona:
Thanks for the PR, but I'm not keen on increasing the surface are of the public API to add this level of flexibility. Let's see what the rest of the team thinks.

snicoll:
+1 for not extending the surface area.

wilkinsona:
The manifest customisation change is quite similar to the `JarFileArchive` change. To be able to use a custom `Repackager` subclass, you'll also have to provide a custom Maven plugin as `RepackageMojo` hasn't been designed to allow a custom `Repackager` subclass to be reused. This also the problem of the CLI's `jar` and `war` which use `Repackager` and equivalent support for Gradle (our plugin doesn't use `Repackager`). Given the apparently limited benefit, I'm not convinced that the increased surface area of the public API is warranted.

Can you please take a step back and describe your jar format, the benefits it has over Boot's default format, and why the existing layout customisation does not meet your needs. We've jumped straight to a solution here and it's hard to justify it with no knowledge of the problem it solves.

handcraftedbits:
> Can you please take a step back and describe your jar format, the benefits it has over Boot's default format, and why the existing layout customisation does not meet your needs.

Sure, fair enough.  I'm making something akin to Spring Boot with a strict JavaEE focus.  I'm aware of WildFly, Open Liberty, etc. but those solutions entail cramming an entire heavyweight (relatively speaking) appserver and WAR into a single JAR.  I'm looking to do something along the lines of Boot wherein the JAR only contains application code and the barest of essentials -- just the runtime libraries used to provide a JAX-RS+CDI+etc. environment.  I'm also aware of Dropwizard but that uses a shaded JAR approach, which has its own issues.

The main issue with Boot's layout is that it lumps application and "server" (for lack of a better term) libraries together in a single classloader.  This can lead to clashes when the "server" portion of the application and user code depend on the same things (the commons-lang, guava, and Jacksons of the world).  This can be handled with Maven dependency management but that has implications -- will the user code be able to handle the (most likely) older version of said dependency that the server portion uses, or will the server portion be able to handle an up-versioned dependency if the user code depends on it?  If not, you're left with shading individual artifacts to prevent clashes.

I'd rather avoid that by using multiple classloaders, as you normally see in appservers -- there's a classloader for the "server" and the application (WAR) has its own classloader with limited visibility into its parent.  At runtime you have to decide which libraries belong to which portion of the application (user versus server), and the easiest way to do that is to either:

* Package the libraries separately, e.g. `BOOT-INF/lib` for application code and `BOOT-INF/server-lib` for "server" code
* Package everything in `BOOT-INF/lib` and denote which libraries are which in the manifest

Then at runtime a custom `Launcher` can construct the classloader hierarchy.  So, that's where a custom layout and/or manifest customization come in.

Also, Boot likes to repackage the classes of the main project artifact into `BOOT-INF/classes`.  This presents problems for bean scanning in CDI.  Weld (but I'm sure this is probably true of other implementations) will look for all `META-INF/beans.xml` files on the classpath and determine which JARs they belong to.  The JARs are scanned for `.class` files and then bytecode-scanned for annotations, etc.  For performance reasons, the classloader is bypassed -- meaning the scanner isn't aware of the custom layout.  So this means, for example, that it will scan a class called `BOOT-INF/classes/x/y/Z.class` and expect the class name to be `BOOT-INF.classes.x.y.Z`, which is obviously not what's intended.  The solution is to not package classes in this manner and to instead package the project artifact.  This is something I've had no problem handling with the existing functionality in `Repackager`.

Ultimately all these changes do require making a custom `RepackageMojo`, which is perfectly reasonable -- I also handle dependency packaging in a very different way than Boot (user projects depend only on API JARs, and the corresponding implementation/"server" dependencies are mapped from them and packaged).  That brings up the last reason for the change, the resulting format and tooling is all different and there's no need for the "Spring Boot"-isms to be present (manifest entries, temp file names, etc.).

The fundamental issue, and honestly I'm fine with a "no" answer here is -- are spring-boot-loader and spring-boot-loader-tools meant to be generic enough for others to consume or is the intention really that they shouldn't be used outside of a Spring Boot context?  I may have been my misreading of these projects, there's essentially no dependencies and there's _tons_ of customization points, I guess I had pictured that what I was doing wasn't so far out there, despite being very different from how Boot does things.  These two projects really are the best implementation of the uber-JAR pattern, and I think the fact that I was able to get this far without any modifications shows how flexible they are.  Just looking for a tad bit more customization out of the box.

wilkinsona:
> The fundamental issue, and honestly I'm fine with a "no" answer here is -- are spring-boot-loader and spring-boot-loader-tools meant to be generic enough for others to consume or is the intention really that they shouldn't be used outside of a Spring Boot context?

They're part of Spring Boot and they have only been designed for use within a Spring Boot context. If they're useful outside of that context, then that's great, but it would be a secondary benefit.

Our solution for custom archive layout with Maven is to provide a `LayoutFactory` registered via an entry in `spring.factories`. That will give you complete control over the location of each dependency in the resulting jar file and sounds like it will go a long way to meet your needs.

Thanks for the proposal, but our recommendation is that you use `LayoutFactory`. If you explore this route and it doesn't offer everything you need, please feel free to raise issues for any additional capabilities that you'd like it to offer and we'll be happy to consider them. For example, we'd be happy to consider support for manifest customisation via the `Layout` and `LayoutFactory` approach.

PR #14519, opened 2018-09-19
add a warning message if PID/LOG location uses fallback

agebhar1:
If the content of `PID_FOLDER` is not valid it falls back to `/tmp` without any
notification. This holds for `LOG_FOLDER` too. Added a warning message to notify
the user in such a case.


wilkinsona:
Thanks for the PR. I'm not sure a warning is warranted in either case, at least not in the default launch script. Can you please explain why you think the warnings are warranted?

agebhar1:
Sure.

In a customer setup each service has the same directory layout with a directory for logging and pid file. The service name is also identical across all services. Only the base directory is the name of the service itself.

Unfortunately for two new services the directories for the pid folder didn't exists such that the PID file was located at ` /tmp`. Each time one service was started it overrides the file of the other one. I took some time to figure it out. A warning message about the fallback/reconfiguration (it's not the standard case) to `/tmp` had saved time.

The warning message only appears if the value of `PID_FOLDER` (`LOG_FOLDER`) exists but is not valid (executable).

agebhar1:
@wilkinsona is the feedback ok for you?

wilkinsona:
Yes, thank you. We plan to include the proposed changes in 2.1.

agebhar1:
Thank you @wilkinsona :smiley:

PR #14548, opened 2018-09-20
Fix and improve Neo4j related documentation

michael-simons:
This fixes several issues with the documentation of the Neo4j starter.

* `Neo4jOperations` and `Neo4Template` have been removed with Spring Data Neo4j 5 already and should not be mentioned neither in 2.1 or 2.0 documentation
* Fixes the paragraph where disabling auto configuration is described. It's done via an own `SessionFactory`
* Improves and clarifies the requirements for embedded mode
* Improves overall wording

There's one little unrelated change in there: The `City` object used in the Spring Data JPA example is "broken", it has no `country` attribute. So if someone would copy over the repository examples, Spring Boot won't start up due to a derived query validation error, in JPA, Mongo as Neo4j alike.

Thanks for considering this pull request.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
@@ -3726,7 +3726,7 @@ considered. A typical entity class resembles the following example:
 
 		public City(String name, String state) {
 			this.name = name;
-			this.country = country;
+			this.state = state;
 		}
 
 		public String getName() {
@@ -3778,7 +3778,7 @@ The following example shows a typical Spring Data repository interface definitio
 
 		Page<City> findAll(Pageable pageable);
 
-		City findByNameAndCountryAllIgnoringCase(String name, String country);
+		City findByNameAndStateAllIgnoringCase(String name, String state);
 
 	}
 ----
@@ -4189,7 +4189,7 @@ in the following example:
 
 		Page<City> findAll(Pageable pageable);
 
-		City findByNameAndCountryAllIgnoringCase(String name, String country);
+		City findByNameAndStateAllIgnoringCase(String name, String state);
 
 	}
 ----
@@ -4228,29 +4228,28 @@ the Mongo instance's configuration and logging routing.
 [[boot-features-neo4j]]
 === Neo4j
 http://neo4j.com/[Neo4j] is an open-source NoSQL graph database that uses a rich data
-model of nodes related by first class relationships, which is better suited for connected
-big data than traditional rdbms approaches. Spring Boot offers several conveniences for
+model of nodes connected by first class relationships, which is better suited for connected
+big data than traditional RDBMS approaches. Spring Boot offers several conveniences for
 working with Neo4j, including the `spring-boot-starter-data-neo4j` "`Starter`".
 
 
 
 [[boot-features-connecting-to-neo4j]]
 ==== Connecting to a Neo4j Database
-You can inject an auto-configured `Neo4jSession`, `Session`, or `Neo4jOperations`
-instance as you would any other Spring Bean. By default, the instance tries to connect to
-a Neo4j server at `localhost:7474`. The following example shows how to inject a Neo4j
-bean:
+To access a Neo4j server, you can inject an auto-configured `org.neo4j.ogm.session.Session`. 
+By default, the instance tries to connect to a Neo4j server at `localhost:7687` via the Bolt 
+protocol. The following example shows how to inject a Neo4j session:
 
 [source,java,indent=0]
 ----
 	@Component
 	public class MyBean {
 
-		private final Neo4jTemplate neo4jTemplate;
+		private final Session session;
 
 		@Autowired
-		public MyBean(Neo4jTemplate neo4jTemplate) {
-			this.neo4jTemplate = neo4jTemplate;
+		public MyBean(Session session) {
+			this.session = session;
 		}
 
 		// ...
@@ -4258,19 +4257,19 @@ bean:
 	}
 ----
 
-You can take full control of the configuration by adding a
-`org.neo4j.ogm.config.Configuration` `@Bean` of your own. Also, adding a `@Bean` of type
-`Neo4jOperations` disables the auto-configuration.
-
-You can configure the user and credentials to use by setting the `spring.data.neo4j.*`
+You can configure the uri and credentials to use by setting the `spring.data.neo4j.*`
 properties, as shown in the following example:
 
 [source,properties,indent=0]
 ----
-	spring.data.neo4j.uri=http://my-server:7474
+	spring.data.neo4j.uri=bolt://my-server:7687
 	spring.data.neo4j.username=neo4j
 	spring.data.neo4j.password=secret
 ----
+
+You can take full control over the session creation by adding a 
+`org.neo4j.ogm.config.Configuration` `@Bean`. To completely disable the auto-configuration
+provided by the "`Starter`" , add a `@Bean` of type `org.neo4j.ogm.session.SessionFactory`.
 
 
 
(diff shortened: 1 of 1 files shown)


michael-simons:
Sorry, I seem to have created my branch before you fixed the build.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4228,72 +4228,90 @@ the Mongo instance's configuration and logging routing.
...
 === Neo4j
 http://neo4j.com/[Neo4j] is an open-source NoSQL graph database that uses a rich data
-model of nodes related by first class relationships, which is better suited for connected
-big data than traditional rdbms approaches. Spring Boot offers several conveniences for
+model of nodes connected by first class relationships, which is better suited for connected
+big data than traditional RDBMS approaches. Spring Boot offers several conveniences for
 working with Neo4j, including the `spring-boot-starter-data-neo4j` "`Starter`".
 
 
-
```
snicoll: Can you please restore this line? We have 3 lines between sections.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4228,72 +4228,90 @@ the Mongo instance's configuration and logging routing.
...
-	spring.data.neo4j.uri=http://my-server:7474
+	spring.data.neo4j.uri=bolt://my-server:7687
 	spring.data.neo4j.username=neo4j
 	spring.data.neo4j.password=secret
 ----
 
+You can take full control over the session creation by adding a 
+`org.neo4j.ogm.config.Configuration` `@Bean`. To disable the auto-configuration
+provided by the "`Starter`" completely, add a `@Bean` of type 
+`org.neo4j.ogm.session.SessionFactory`. Take note that you have to provide a
```
snicoll: Is the `Neo4jTransactionManager` mandatory in case of custom setup? I am not sure that note should be added here: if users are taking full control, they are responsible to know what needs to be done and refer to the documentation of the components they are configuring themselves.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4228,72 +4228,90 @@ the Mongo instance's configuration and logging routing.
...
 ==== Using the Embedded Mode
 
 If you add `org.neo4j:neo4j-ogm-embedded-driver` to the dependencies of your application,
 Spring Boot automatically configures an in-process embedded instance of Neo4j that does
-not persist any data when your application shuts down. You can explicitly disable that
-mode by setting `spring.data.neo4j.embedded.enabled=false`. You can also enable
-persistence for the embedded mode by providing a path to a database file, as shown in the
-following example:
+not persist any data when your application shuts down. As the embedded Neo4j OGM  driver 
+does not provide the Neo4j kernel itself, you have to provide this dependency manually.
```
snicoll: What is the reason for moving the note here? I actually like the fact it's a note as it renders in a more eye-catching way than a sentence within a paragraph.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4228,72 +4228,90 @@ the Mongo instance's configuration and logging routing.
...
+
+Your dependencies for using embedded Neo4j should look as follows:
 
+[source,xml,indent=0]
 ----
-	spring.data.neo4j.uri=file://var/tmp/graph.db
+	<dependency>
+		<groupId>org.neo4j</groupId>
+		<artifactId>neo4j-ogm-embedded-driver</artifactId>
+		<version>${neo4j-ogm.version}</version>
```
snicoll: This shouldn't need a version.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4228,72 +4228,90 @@ the Mongo instance's configuration and logging routing.
...
+	</dependency>
+	<dependency>
+		<groupId>org.neo4j</groupId>
+		<artifactId>neo4j</artifactId>
+		<version>3.4.7</version>
```
snicoll: Sorry but I am not keen to hard-code versions in our doc. I think the link that is referenced above should mention this (if it doesn't already).

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4228,72 +4228,90 @@ the Mongo instance's configuration and logging routing.
...
+kernel are on the classpath as described above.
+
 [NOTE]
 ====
-The Neo4j OGM embedded driver does not provide the Neo4j kernel. Users are expected to
-provide this dependency manually. See
-http://neo4j.com/docs/ogm-manual/current/reference/#reference:getting-started[the
-documentation] for more details.
+You can enable persistence for the embedded mode by providing a path to a database file 
+through `spring.data.neo4j.uri`, for example `spring.data.neo4j.uri=file://var/tmp/graph.db`.
```
snicoll: I wouldn't mention the property twice. The example alone is fine.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4311,48 +4329,38 @@ pattern). If you do not want this behavior, add the following line to your
...
 
-TIP: For complete details of Spring Data Neo4j, including its rich object mapping
-technologies, refer to the https://projects.spring.io/spring-data-neo4j/[reference
+The `spring-boot-starter-data-neo4j` "`Starter`" enables the repository support as well 
+as transaction management. You can customize the locations to look for repositories and
+entities by using `@EnableNeo4jRepositories` and `@EntityScan` respectively on a 
+`@Configuration`-bean.
+
+TIP: For complete details of Spring Data Neo4j, including its object mapping
+technologies, refer to the https://docs.spring.io/spring-data/neo4j/docs/current/reference/html/[reference
```
snicoll: We're not pointing to `current` on purpose as it won't be accurate. This link is incorrect in the Spring Boot 2.0.x documentation for instance as we're not using the "current" version. This is the main reason we use the projects page link (see others Spring Data section for consistency).

I am not against reconsidering this but I'd prefer to do so consistently and in such a way that the version is correct.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
michael-simons: Sadly, it does. It's not a managed dependency. I've created a separate PR (#14560) for that. I'd rather not see a version here myself.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
michael-simons: Yes it is, we like to have it mentioned here. As much as I agree with you regarding responsibility, it helps to keep confusion away.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
michael-simons: I'll solve this together with the hardcoded version. You're right.

michael-simons:
Rebased and added most of the requested changes. Thanks for that feedback.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
snicoll: I understand but that's not very consistent. If we had to document things that should be done when our auto-config backs off, there are a gazillion of things that should be added to the current ref guide.

Can we take a step back and discuss the source of the confusion? If it's mandatory, perhaps the Neo4j integration could fail with an explicit error message?

snicoll:
@michael-simons thanks a lot for the PR and the feedback on my comments. This is now merged in `2.0.x` and `master`.

Issue #14345, opened 2018-09-07
Migrate size properties to DataSize


snicoll:
We have some properties that represents a size (usually in bytes). Now #13821 is done we should migrate them to `DataSize` so that we get proper IDE support.

qct:
Do you need any help?

snicoll:
@qct if you have time today then please go ahead but I've assigned myself as I really want this to be fix for M4 Monday.

qct:
@snicoll  please take a look, build error because of previous commits, seems I can't trigger a CI rebuild

snicoll:
Closing in favour of PR #14549

PR #14549, opened 2018-09-20
Migrate size properties to DataSize

qct:
see https://github.com/spring-projects/spring-boot/issues/14345


snicoll:
@qct thanks for the PR. They were actually a bunch of other properties to migrate. I've done that in  cbae22f with a polish of your contribution.

Issue #7094, opened 2016-10-04
Added Neo4j Embedded Driver to Spring Boot.


xenoterracide:
In addition to testing on #7012, I believe that the embedded driver is licensed under the Apache 2.0 License see: https://repo1.maven.org/maven2/org/neo4j/neo4j-ogm-embedded-driver/2.0.5/neo4j-ogm-embedded-driver-2.0.5.pom. 

If there is a component that is AGPL, [my](http://opensource.stackexchange.com/a/2789/1060) 
 [understanding](http://programmers.stackexchange.com/q/107883/2331) of [that](https://www.gnu.org/licenses/gpl-faq.en.html#MereAggregation) is that 
> A separable portion of the object code, whose source code is excluded from the Corresponding Source as a System Library, need not be included in conveying the object code work.

because this is a driver, so long as you are not extending any classes (or directly using the drivers internal apis (meaning all you're doing is providing it with the path), and you could easily substitute the http or bolt drivers, that it can't be considered a derived work.

The purpose here is 2 fold, determine if we can provide dependency management for the driver/why or why not. If not improve the documentation to clarify and document the reason (I'll be happy to write that patch), or add it to the BOM, which I would also be happy to do.

I'm going to be reaching out the the FSF for clarification on the AGPL given what I've just said, and will request that they comment here. I'll also reach out to Neo4J

xenoterracide:
https://neo4j.com/licensing/
> >> I’m using an Apache-licensed driver to access Neo4j. What does this mean for my application?
That’s great. Apache-licensed drivers allow you to redistribute your own code. Of course you still need to abide by the same Neo4j license terms as you normally would.

snicoll:
@xenoterracide don't take this the wrong way but I want someone from Neo4j to confirm the license term here. There's a reason why we didn't ship the embedded driver after all.

xenoterracide:
@snicoll I have emailed them, pending response

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

xenoterracide:
I did put in issues with both Neo4j Licensing, and FSF, but haven't heard from them obviously. Would it be possible to put a doc patch into the reference documentation explaining what the licensing concern/issue is?

snicoll:
[...]

PR #14560, opened 2018-09-20
Add dependency management for Neo4J OGM embedded driver

michael-simons:
At the moment one has to specify the version of the embedded driver for OGM in contrast to Bolt and HTTP driver. This PR adds the embedded driver with the same version as Bolt and Http as a managed dependency.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -2397,6 +2397,11 @@
 				<version>${neo4j-ogm.version}</version>
 			</dependency>
 			<dependency>
+				<groupId>org.neo4j</groupId>
+				<artifactId>neo4j-ogm-embedded-driver</artifactId>
+				<version>${neo4j-ogm.version}</version>
+			</dependency>
+			<dependency>
 				<groupId>org.postgresql</groupId>
 				<artifactId>postgresql</artifactId>
 				<version>${postgresql.version}</version>


snicoll:
I remember we didn’t add this one because of a different license. Is this artifact published with asl v2?

michael-simons:
The Neo4j OGM embedded driver is not to be confused with the embedded version of Neo4j itself. The driver is licensed under ASL v2 and it doesn't bring in the Neo4j kernel (as mentioned in the docs). 

But I like @meistermeier or @digitalstain to confirm this.

meistermeier:
I am not a license expert but I can say that the embedded-driver itself is ASLv2. The database it will then as a dependency pull in is GPLv3 (community edition).
I do not know if this is seen as linking.

michael-simons:
We just checked double checked regarding

* `org.neo4j:neo4j`
* `org.neo4j:neo4j-cluster`
* `org.neo4j:neo4j-ha`
* `org.neo4j:neo4j-enterprise`

They are all in the provided scope of Neo4j-OGM and thus not transitive dependencies. All of them have to be added explicitly and manual by users of embedded.

To sum this up: The Neo4j OGM embedded driver is licensed ASLv2 and doesn't bring in non-ASL Neo dependencies and therefor it's ok to add them to Spring Boots dependency management analogue to Neo4-OGM Bolt and HTTP drivers.

snicoll:
Thanks for the feedback. Looking at it a bit more, I realize now that declining #7094 was a mistake.

snicoll:
Thanks @michael-simons, this is now merged in `2.0.x` and `master`.

PR #14565, opened 2018-09-21
Add log messages to lines only when the destination isn't set

izeye:
This PR adds log messages to `this.lines` in `DeferredLog.log()` only when the `this.destination` isn't set as it doesn't seem to be intentional unless I'm missing something.


philwebb:
Good spot yet again! Thanks.

PR #14569, opened 2018-09-21
Polish RabbitMQ properties doc

garyrussell:
Fix description for `prefetch`; remove reference to transactions since
the direct container does not support transactionSize.

Also fix `transactionSize` to just reference acks; autoConfiguration doesn't
support enabling transactions.


snicoll:
Thanks for the update @garyrussell!

PR #14570, opened 2018-09-21
Added paragraph for setting up Hibernate second-level caching

marceloverdijk:
As discussed with @snicoll hereby a PR for documenting Hibernate second-level caching.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
@@ -2000,6 +2000,66 @@ for more details.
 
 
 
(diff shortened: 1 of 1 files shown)


snicoll:
I've decided to take a different approach in #14734. If prefer we do not duplicate too much of the documentation of Hibernate in our ref guide and we rather refer to it. 

Also, none of that configuration you've mentioned should be necessary when you've set the `CACHE_MANAGER` property (as the purpose is to provide an already initialized `javax.cache.CacheManager`) and the example wrongly set a `org.springframework.cache.CacheManager`. 

Please give a try to what is now documented and let us know if something should be polished.

Thanks for the for PR in any case.

Issue #14477, opened 2018-09-14
Cache Buster doesn't work with error templates


martinei:
### tl:dr 
`spring.resources.chain.strategy.content.enabled` doesn't work for error templates served by `BasicErrorController`

### Long version:

I have reproduced this with 2.0.5.RELEASE, `spring-boot-starter-thymeleaf` and `spring-boot-starter-web`.
Into a skeleton generated with Initializr I added a  _src/main/resources/templates/error/404.html_ which includes:

```
   <link rel="stylesheet" th:href="@{/style.css}">
```

I also added the  the corresponding style.css.

In `application.propeties` I enabled the cache buster:

```
spring.resources.chain.strategy.content.enabled=true
spring.resources.chain.strategy.content.paths=/**
```

**Expected**: When opening a unknown resource in the browser ("localhost:8080/foo") the error page should be shown and include a link with fingerprinted css file name: 

```html
<link rel="stylesheet" href="/style-cd91a83d51c65638e2a3cb401620bc8e.css">
```

**Is**:

```
<link rel="stylesheet" href="/style.css">
```

A Repository to reproduce: https://github.com/martinei/spring-boot-bug-14477/tree/master

philwebb:
@bclozel Do you know if this is intentional? Does framework back off the resource transformers if they are part of the error response?

bclozel:
I think this is due to a change introduced in Spring Boot 2. By default now servlet filters are dispatched only on DipatcherType.REQUEST. If we want this to work on error dispatch as well we need to change the auto-configuration of the ResourceUrlEncodingFilter into a FilterRegistragionBean with both request and error dispatch. See https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.0-Migration-Guide#servlet-filters

Would that be a good candidate for a first contribution issue?

wilkinsona:
@bclozel It sounds like it might be. Probably depends on how complex a new test would be.

philwebb:
Closing in favor of PR #14583

PR #14580, opened 2018-09-22
Cache Buster doesn't work with error templates

ayudovin:
this [bug](https://github.com/spring-projects/spring-boot/issues/14477)


snicoll:
Duplicate of #14583

PR #14592, opened 2018-09-24
Polish

mdeinum:
The SpringBootContextLoader still checked if tests were
annotated with @IntegrationTest or @WebIntegrationTest.
Those classes have been removed in Spring Boot 1.5 and
been deprecated since 1.4.

Removed the check on the annotations as well as the
static list of names.


philwebb:
Thanks! I think it's probably worth applying this to 2.0.x

PR #14593, opened 2018-09-24
Avoid using deprecated configurations in Gradle plugin reference documentation

jnizet:
See https://github.com/spring-projects/spring-boot/pull/14585#issuecomment-423804560


wilkinsona:
Thank you very much, @jnizet. The proposed changes have been merged into 2.0.x and master.

Issue #14599, opened 2018-09-25
Section header in Actuator API docs for query parameters of retrieving caches by name is incorrect


wilkinsona:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before, or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Problem
The documentation for the Actuator's API contains a section named "Request Structure" but it should be named "Query Parameters". The ID assigned to the section is also incorrect which has a knock-on effect on the anchors in the generated HTML.

## Solution

Both the title of the section and its ID need to be corrected.

The ID can be found here: https://github.com/spring-projects/spring-boot/blob/fd2080e7cc38af2db7c7a669e650fcce9a6194b3/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/asciidoc/endpoints/caches.adoc#L45

The ID should be `[[caches-named-query-parameters]]`

The name can be found here: https://github.com/spring-projects/spring-boot/blob/fd2080e7cc38af2db7c7a669e650fcce9a6194b3/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/asciidoc/endpoints/caches.adoc#L46

The name should be `=== Query Parameters`

## Steps to Fix 
 
- [x] Claim this issue with a comment below and ask any clarifying questions you need
- [x] Set up a repository locally following the [Contributing Guidelines](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc)
- [x] Try to fix the issue following the steps above
- [x] Commit your changes and start a pull request.

wilkinsona:
Looks like @hanope has taken care of this. Thank you. Closing in favour of #14600.

PR #14600, opened 2018-09-25
Section header in Actuator API docs for query parameters of retrieving caches by name is incorrect

Hanope:
I fix the section header in Actuator API docs. #14599


snicoll:
@Hanope thank you for making your first contribution to Spring Boot. This is now merged in `master`.

Issue #14555, opened 2018-09-20
Validation of Client Registration Ids


rwinch:
With the changes in https://github.com/spring-projects/spring-boot/commit/f5deebf0cb2fb802bafb5ff85652ebe89e9da98d#diff-b716836901d5fea39e485e35997f2f62 we should probably validate the client registration ids are not duplicated. For example, with the old configuration this:
```yml
spring:
  security:
    oauth2:
      client:
        provider:
          idp:
            issuer-uri: https://example.com
        registration:
          idp:
            client-id: client-id
            client-secret: client-secret
          idp:
            client-id: client-id
            client-secret: client-secret
```

 produced an error:

```
Caused by: org.yaml.snakeyaml.constructor.DuplicateKeyException: while constructing a mapping
 in 'reader', line 18, column 11:
              idp:
              ^
found duplicate key idp
 in 'reader', line 21, column 11:
              idp:
              ^

	at org.yaml.snakeyaml.constructor.SafeConstructor.processDuplicateKeys(SafeConst
```

However, the new configuration does not produce an error:

```yml
spring:
  security:
    oauth2:
      client:
        provider:
          idp:
            issuer-uri: https://example.com
        registration:
          login:
            idp:
              client-id: client-id
              client-secret: client-secret
            authorizationcode:
              idp:
                client-id: client-id
                client-secret: client-secret
```

Related: #14554

mbhave:
Closing in favor of #14609.

Issue #14554, opened 2018-09-20
Consider Alternative to in Grant Type Hiearchy


rwinch:
The changes in https://github.com/spring-projects/spring-boot/commit/f5deebf0cb2fb802bafb5ff85652ebe89e9da98d#diff-b716836901d5fea39e485e35997f2f62 make it so that the grant type is in the hierarchy. I'm not sure if this is a good long term solution since there will likely come a time (soon) where we support multiple grant types in a registration (with one being "preferred"). When a user requests a token they would then specify which grant type if they wanted something different than the default.

With these changes in mind it is going to make configuring the code challenging because the user would need to duplicate the configuration. For example:

```
spring:
  security:
    oauth2:
      client:
        provider:
          idp:
            issuer-uri: https://example.com/
        registration:
          authorizationcode:
            idp-ac:
              provider: idp
              client-id: client-id
              client-secret: client-secret
          clientcredentials:
            idp-cc:
              provider: idp
              client-id: client-id
              client-secret: client-secret
```

One option would be to allow additional grant types property on the registration. However, this becomes ambiguous how the user should configure it. Should they place it under authorizationcode or clientcredentials? If they place it under a grant type that is missing a property for the additional grant type then, the would get an error too. For example:

```yml
spring:
  security:
    oauth2:
      client:
        provider:
          idp:
            issuer-uri: https://example.com/
        registration:
          clientcredentials:
            idp-cc:
              client-id: client-id
              client-secret: client-secret
              additional_grant_types: authorization_code
              redirect-url: /foo/bar # redirect-url is not needed for client-credentials
```

I'd like to consider going back to the original way. If the authorization_grant_type is provided, the default redirect-url is then mapped accordingly.

```yml
spring:
  security:
    oauth2:
      client:
        provider:
          idp:
            issuer-uri: https://example.com/
        registration:
          idp:
            client-id: client-id
            client-secret: client-secret
            authorization_grant_type: authorization_code
```

mbhave:
@rwinch If there is going to be support for a "preferred" grant type, I don't see why additional_grant_types would be misleading. The default one would be the that's under the key and the additional ones would be supported if the user specifies that in the request. 

> If they place it under a grant type that is missing a property for the additional grant type then, the would get an error too

We don't have this sort of validation and when/if we do add validation we can take such factors into account.

rwinch:
> We don't have this sort of validation and when/if we do add validation we can take such factors into account.

[...]

PR #14609, opened 2018-09-26
validation of Client Registration Ids

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14555)


mbhave:
@ayudovin thank you for the PR. This one is also dependent on whether we keep the grant type hierarchy or not. I'm going to place this PR on hold till we decide what to do in #14554.

PR #14614, opened 2018-09-26
Expose preferProxiedForAddress Jetty access log property.

### LINKED ISSUE #14608: Expose preferProxiedForAddress of Jetty NCSARequestLog in ServerProperties
For applications behind an HTTP proxy, it is more correct to report the IP address of the `X-Forwarded-For` header in HTTP access logs instead of the socket remote address. Fortunately this behavior is easily enabled in Jetty by setting `NCSARequestLog#setPreferProxiedForAddress(true)`. It would be helpful to expose this setter as a Spring property.

serac:
Provides the Spring boot property `server.jetty.accesslog.prefer-proxied-for-address` to control Jetty access log behavior to prefer `X-Forwarded-For` header if present for logging IP address in HTTP access logs. Closes #14608.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -56,6 +56,7 @@ import org.springframework.util.unit.DataSize;
  * @author Brian Clozel
  * @author Olivier Lamy
  * @author Chentao Qu
+ * @author Marvin S. Addison
  */
 @ConfigurationProperties(prefix = "server", ignoreUnknownFields = true)
 public class ServerProperties {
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -831,6 +832,11 @@ public void setSelectors(Integer selectors) {
 			 */
 			private boolean logLatency;
 
+			/**
+			 * Prefer IP address specified in X-Forwarded-For header.
+			 */
+			private boolean preferProxiedForAddress;
```
snicoll: I'd rather research if we couldn't apply this behaviour when `server.use-forward-headers` is set rather than requiring a new property.

wilkinsona:
Thanks for the PR but as per my comments on #14608 starting [here](
https://github.com/spring-projects/spring-boot/issues/14608#issuecomment-426643445), I don't think we need to do anything here. The existing `server.use-forward-headers` property should already be sufficient to get the desired access logging behaviour with Jetty, Tomcat, and Undertow.

Issue #14552, opened 2018-09-20
Add Support for client_credentials Auto Configuration


rwinch:
With the changes in https://github.com/spring-projects/spring-boot/commit/f5deebf0cb2fb802bafb5ff85652ebe89e9da98d#diff-b716836901d5fea39e485e35997f2f62 the configuration below works but longer makes sense:

```yml
spring:
  security:
    oauth2:
      client:
        provider:
          idp:
            issuer-uri: https://example.com/
        registration:
          authorizationcode:
            idp:
              client-id: client-id
              client-secret: client-secret
              authorization_grant_type: client_credentials
```

Notice that a client_credentials grant type is being configured under the authorizationcode parent. 

Related #14554

mbhave:
Closing in favor of #14625.

PR #14625, opened 2018-09-27
adding support for client_credentials auto configuration

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14552)


mbhave:
@ayudovin thank you for the PR. We're still deciding whether we want to go with the client hierarchy based on the grant_type. I'm going to place this PR on hold till we decide what to do in #14554.

PR #14637, opened 2018-09-29
Use Commons Logging for OnlyOnceLoggingDenyMeterFilter

izeye:
This PR changes to use Commons Logging for `OnlyOnceLoggingDenyMeterFilter` as it seems that SLF4J is used accidentally.


izeye:
I missed dropping `this` keywords for `logger` as I changed it to `static`. I updated it now.

Issue #14022, opened 2018-08-08
Add support for health indicator groups


ngaya-ll:
My organization uses Spring Boot services running in a Kubernetes environment.  Kubernetes has a concept of two types of health probes: [liveness and readiness probes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-probes/).  Liveness probes are used to detect if the service needs to be restarted, while readiness probes detect if the service is ready to be added to load balancers.

It would be great if there were a way to perform different checks for readiness and liveness.  For example, if an application temporarily loses connectivity to its data source it should be taken out of load balancing, but doesn't need to be restarted.  So we would like to include the `DataSourceHealthIndicator` in the readiness probe but not the liveness probe.

Note that https://github.com/spring-projects/spring-boot/issues/8685 already allows querying a single health indicator, but not multiple indicators.  Perhaps Spring Boot could provide a way to configure named groups of indicators that could be accessed as `/actuator/health/{group}`.

eddumelendez:
I am working in this issue. I will be sending a PR this weekend.

wilkinsona:
Thank you, @eddumelendez. We had a few thoughts when we discussed this earlier this week. They are only initial thoughts so they aren't prescriptive, but hopefully they'll be of some use.

We envisaged a property-based mechanism for creating a group of health indicators. In `application.properties`, the value of the property would probably be a comma-separate list of the ids of the health indicators that are to be included in the group. The key would be the name of the group.

We then thought that `CompositeHealthIndicator` could be reused and created with all the health indicators in the group. One gotcha that @philwebb identified is that these composite health indicators that represent a group will need to be hidden from the main health check otherwise any grouped indicators would be called twice for a single check.

eddumelendez:
Thanks for letting me know @wilkinsona I'll be working on this taking into account these thoughts.

mbhave:
@eddumelendez Are you still interested in working on this? If not, we'll go ahead and pick it up.

eddumelendez:
@mbhave I have an initial draft which I can share it later today.

mbhave:
great, thanks!

eddumelendez:
sorry for the delay. I have submitted #16252

snicoll:
I gave this some thoughts and I am wondering if we shouldn't have a concept of "out-of-the-box groups" the same way we have logging groups. The two obvious ones are those defined in this very request.

[...]

PR #14639, opened 2018-09-30
Exclude Beans from the Health Check

MahmoudAGawad:


philwebb:
@MahmoudAGawad It looks like something has gone wrong with your pull request and all the commits from master are being merged into 2.0.x. Could you please try again and just submit a single commit that can be applied to current `master` branch. You can do a force push to update this PR.

philwebb:
Related, we also have issue #14022 which might be related.

Issue #14492, opened 2018-09-17
Make it easier to tell that DevTools restart has been disabled due to HotSwapAgent being present


danielpgross:
Testing using a simple spring-boot starter generated by initializr: [Download starter](https://start.spring.io/starter.zip?type=maven-project&language=java&bootVersion=1.5.16.RELEASE&baseDir=demo&groupId=com.example&artifactId=demo&name=demo&description=Demo+project+for+Spring+Boot&packageName=com.example.demo&packaging=jar&javaVersion=1.8&autocomplete=&generate-project=&style=devtools)

With identical JDKs and maven versions, devtools works fine on Ubuntu 16.04 but does not work on macOS 10.13.6. The symptoms are:
1. The main thread is called `main` instead of `restartedMain`.
2. The application does not restart automatically (even though it should) upon changes to files in the classpath.

I have tested this going back to older versions and determined that the first version in which this broke is 1.5.3.

To reproduce:
1. [Download starter](https://start.spring.io/starter.zip?type=maven-project&language=java&bootVersion=1.5.16.RELEASE&baseDir=demo&groupId=com.example&artifactId=demo&name=demo&description=Demo+project+for+Spring+Boot&packageName=com.example.demo&packaging=jar&javaVersion=1.8&autocomplete=&generate-project=&style=devtools)
2. Unzip the starter and cd into the unzipped directory.
3. Run `mvn spring-boot:run`

Testcases below:
**macOS**
```
$ mvn -v
Apache Maven 3.5.4 (1edded0938998edf8bf061f1ceb3cfdeccf443fe; 2018-06-17T14:33:14-04:00)
Maven home: /usr/local/Cellar/maven/3.5.4/libexec
Java version: 1.8.0_181, vendor: Oracle Corporation, runtime: /Library/Java/JavaVirtualMachines/jdk1.8.0_181.jdk/Contents/Home/jre
Default locale: en_CA, platform encoding: UTF-8
OS name: "mac os x", version: "10.13.6", arch: "x86_64", family: "mac"
```

```
[INFO] --- spring-boot-maven-plugin:1.5.16.RELEASE:run (default-cli) @ demo ---
[INFO] Attaching agents: []
09:24:58.767 [main] DEBUG org.springframework.boot.devtools.settings.DevToolsSettings - Included patterns for restart : []
[...]

PR #14648, opened 2018-10-01
Add Log Output when DevTools restart is disabled

eiselems:
Also did a small refactoring to get rid of the unused getter and inline
the highest precedence reference.

See gh-14492

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/RestartApplicationListener.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/RestartApplicationListener.java
@@ -16,6 +16,9 @@
 
 package org.springframework.boot.devtools.restart;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
 import org.springframework.boot.context.event.ApplicationFailedEvent;
 import org.springframework.boot.context.event.ApplicationPreparedEvent;
 import org.springframework.boot.context.event.ApplicationReadyEvent;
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/RestartApplicationListener.java
```diff
@@ -35,7 +38,7 @@
 public class RestartApplicationListener
 		implements ApplicationListener<ApplicationEvent>, Ordered {
 
-	private int order = HIGHEST_PRECEDENCE;
+	private static final Log logger = LogFactory.getLog(RestartApplicationListener.class);
```
snicoll: Please refrain from doing unrelated changes. This change to `order`  is unrelated to the task at hand.

spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/RestartApplicationListener.java
```diff
@@ -81,15 +85,6 @@ private void onApplicationFailedEvent(ApplicationFailedEvent event) {
...
-		return this.order;
-	}
-
-	/**
-	 * Set the order of the listener.
-	 * @param order the order of the listener
-	 */
-	public void setOrder(int order) {
-		this.order = order;
+		return HIGHEST_PRECEDENCE;
```
snicoll: I have no opinion has whether this should be changed but we should have this discussion in a separate PR.

eiselems: changed it back

spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/RestartApplicationListener.java
```diff
@@ -67,6 +72,7 @@ private void onApplicationStartingEvent(ApplicationStartingEvent event) {
 			Restarter.initialize(args, false, initializer, restartOnInitialize);
 		}
 		else {
+			logger.info("RestartApplicationListener is disabled.");
```
wilkinsona: Logging here could be useful, particularly if it indicated when the restarter hasn't been initialised, but it doesn't meet the requirement described in #14492. That issue is specifically interested in the case where `restartOnInitialize` is false because an agent reloader is active.

eiselems: Mhm maybe I gonna close this one and bring it together if I find a resolution to satisfy #14492 then.

PR #14653, opened 2018-10-01
Add factory method to RestTemplateBuilder

BrianDeacon:
Changed builder methods in `RestTemplateBuilder` to use an overridable factory method to facilitate creation of subclasses.
Previously, a subclass of `RestTemplateBuilder` would no longer be used if a non-overriden builder method was used.

i.e. a `SubclassedBuilder.defaultMessageConverters().build()` would not call `build()` on the subclass instance. Now that subclass
could get that behavior by overriding the new "newInstance" method.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RequestFactoryCustomizer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RequestFactoryCustomizer.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2012-2018 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.web.client;
+
+import java.time.Duration;
+import java.util.function.Consumer;
+
+import org.springframework.http.client.ClientHttpRequestFactory;
+
+/**
+ * A customizer to set properties on passed-in {@link ClientHttpRequestFactory} instances
+ * as used by {@link RestTemplateBuilder}.
+ *
+ * @author Brian Deacon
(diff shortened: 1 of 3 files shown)


philwebb:
`RestTemplateBuilder` wan't really designed to be subclassed and I'm not keen to promote it as a pattern.

Can we take a setup back and could you describe the problem you faced that made you want to subclass `RestTemplateBuilder` in the first place?

BrianDeacon:
Digging into this more, it looks like we were doing this poorly, so this has turned into a more philosophical issue than a practical one. We were overriding the build method to add our own RequestInterceptor to the built template, and I see now that RestTemplateBuilder::additionalInterceptors would be the way to solve that, and the fact that a new instance was being called every time wouldn't be an issue.

We have a shared library intended to be "drop in" for other developers in the organization that can just have particular configurations of a RestTemplateBuilder component immediately available for Autowiring (given a few boiler plate environment variables are set). We have three different flavors of these subclasses that need three different kinds of Interceptors added to built templates. Currently our use cases are around setting headers, but we weren't intending that to be the only thing we'd do. e.g., we have RestTemplateBuilders for "these templates will propagate auth headers" or "these other ones will look up service-level headers and use those". One of these subclasses might be used in a few places to then be further customized as needed. The situation where we ran into the problem was when an injected instance of our subclass did `theCustomInstance.errorHandler(handler).build()`. The build invocation was then not `theCustomInstance.build()` but on the new RestTemplateBuilder instantiated inside errorHandler

So that specific issue would be solved just by changing where we add the interceptors.

But it frankly didn't occur to us that the builder methods would return a new instance rather than just `this`. An immutable builder strikes me as odd, and I imagine was the reason our developers hit this unexpected behavior. Our short-term workaround for clients using our current version of the shared library is that we just make the errorHandler() call on the built template. In the medium-term, we'll update our shared library and have our subclasses call additionalInterceptors instead.

It also seems a little unexpected that if we inject a RestTemplateBuilder, then further customize it with various calls to the builder methods, that we are no longer dealing with the singleton instance of the injected Bean. (Let alone the oddness that we have an immutable builder building mutable RestTemplates)

philwebb:
The `RestTemplateBuilder` is usually a singleton bean in the context, but we wanted to allow users to provide further customization at injection time.

For example, you might have something like this:

```java
@Bean
public ServiceA serviceA(RestTemplateBuilder builder) {
    return new ServiceA(builder.basicAuthorization("sa", "sa"));
}

@Bean
public ServiceB serviceB(RestTemplateBuilder builder) {
    return new ServiceB(builder.defaultMessageConverters());
}
```

If the builder is mutable then it's possible that `serviceA` is called before `serviceB` and `serviceB` gets unexpected basic authorization configured.

PR #14657, opened 2018-10-02
Use non-reflective APIs to inject converters

dsyer:
The `@Qualifier` APIs are a bit limited in Spring still, so we might
want to use this to pursue additional convenience methods before
merging. See [SPR-8891](https://jira.spring.io/browse/SPR-8891).


wilkinsona:
I share your dislike of having to force the factory method to initialise itself. Hopefully you'll get some traction on SPR-8891.

dsyer:
Those 4 commits can be squashed together. SPR-8891 is now resolved and this looks like it works,

PR #14667, opened 2018-10-03
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/reactive/server/WebFluxTags.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/reactive/server/WebFluxTags.java
@@ -128,10 +128,11 @@ public final class WebFluxTags {
 	}
 
 	/**
-	 * Creates a {@code outcome} tag based on the response status of the given
+	 * Creates an {@code outcome} tag based on the response status of the given
 	 * {@code exchange}.
 	 * @param exchange the exchange
 	 * @return the outcome tag derived from the response status
+	 * @since 2.1.0
 	 */
 	public static Tag outcome(ServerWebExchange exchange) {
 		HttpStatus status = exchange.getResponse().getStatusCode();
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -29,12 +29,6 @@
         "level": "error"
       }
     },
-    {
-      "name": "server.compression.min-response-size",
-      "description": "Minimum \"Content-Length\" value that is required for compression to be performed.",
-      "type": "org.springframework.util.unit.DataSize",
-      "defaultValue": "2KB"
-    },
```
snicoll: Can you share a bit more context why this got removed? I am afraid I had to add it because `Compression` is in `spring-boot` and the annotation processor can't read the source description on it. 

If you've missed that, you're raising a red flag anyway as we should probably document all properties consistently.

izeye: @snicoll Thanks for the feedback! I just expected that [comments on fields in `Compression`](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/Compression.java#L30-L50) will be harvested for the configuration metadata by the annotation processor. Do I have a wrong expectation?

snicoll: > Do I have a wrong expectation?

You don't but, unfortunately, it doesn't work that way at the moment. 

`Compression` is in the `spring-boot` module and it is being used in the `spring-boot-autoconfigure` module. When the annotation processor runs, it doesn't have access to the source model of `Compression` (as it's not being compiled in that round). I can see one way to fix that by adding the sources of `spring-boot` when compiling `spring-boot-auto-configure`. 

In any case, let's raise a separate issue for this and we can take it from there.

izeye: @snicoll Thanks for the quick feedback! I reverted the change in 27b987c. I guess the other three fields need to be documented there as they are absent until the issue is resolved.

snicoll: Thanks!

Yes, let's figure this out in #14669 (as manually documenting them may be the outcome of that issue).

PR #14670, opened 2018-10-03
Polish docs

jvalkeal:
- Replace tabs to spaces in a yaml example for
  configuring endpoint web exposure include.


snicoll:
Good catch, thanks @jvalkeal!

PR #14673, opened 2018-10-03
Use proxied result return from BeanFactory

mdeinum:
Prior to this commit the initially created instance was always returned.
However when calling initializeBean on the BeanFactory this could
lead to a proxy being returned. Ignoring the return value of this method
could lead to proxies being ignored and AOP not being applied.

Now it always returns the result of the call to initializeBean which is
either a proxied/wrapped instance or the original instance.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/quartz/AutowireCapableBeanJobFactory.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/quartz/AutowireCapableBeanJobFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2018 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


snicoll:
Thanks for the PR. Can you please add a test that exercises the use case you've described?

mdeinum:
As there wasn't a test I didn't bother to create one, but will add one. 

This was actually based on an PR I did a long time ago for Spring Framework in general (see https://github.com/spring-projects/spring-framework/pull/1116) which also takes into account `DelegatingJob` and injecting into the delegate instead of the actual job. Would that be something to add as well?

mdeinum:
I had a problem getting to the internals of Quartz. So I created a new test for the `AutowireCapableBeanJobFactory`

mdeinum:
Has been implemented with d9e72535328ff56e447a9ec69b95263da8a8495b.

wilkinsona:
Master still seems to ignore the return value: 
https://github.com/spring-projects/spring-boot/blob/4ede33e09bef68b54893a9898caac889de0bbfec/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/quartz/AutowireCapableBeanJobFactory.java#L45

@mdeinum the commit you’ve referenced above doesn’t seem to exist and, as far as I can tell, this would still be a good change to make. What have I missed?

snicoll:
@wilkinsona I am about to create an issue to replace this code by the one of framework.

wilkinsona:
Great, thanks. FWIW, I’m still confused about the referenced commit. I guess it’s 
https://github.com/spring-projects/spring-framework/commit/d9e72535328ff56e447a9ec69b95263da8a8495b but that’s a Hibernate change.

PR #14676, opened 2018-10-03
Introduce support for passing env props in fully-executable jars

### LINKED ISSUE #14662: Allow to specify environment variables via conf file for fully-executable jars
With many APM tools it is useful to provide additional configuration in the environment of Java processes. Unfortunately, fully-executable jars do not currently have an easy way to add properties to the environment.

Ideally, one would have a `JAVA_ENV` property in the `<executable-file-minus-jar>.conf` file and have the default embedded script put the values in the environment of the Java process.

With multiple variables, the `JAVA_ENV` variable should probably be quotes within single quotes.

mmanciop:
This commit patches the default launch script for fully executable jars to allow the specification of a configuration property called `environmentProperties` and its matching environment property
`ENVIRONMENT_PROPERTIES` that allows to specify environment variables (as in `java.lang.System#getenv()`) to be passed to the process spawned by running the fully-executable jar.

The environment variables to be passed to the process need to be specified as a list of <key>=<value> pairs, e.g.:

environmentProperties=A=B C=D
ENVIRONMENT_PROPERTIES='A=B C=D'

Fixes #14662

I developed and tested this on a Devuan based on Debian Jesse (I am told that is the pickiest `init.d` system around). Please try it on other distros, especially more recent ones.

P.S. I almost lost my sanity over the intricacies of `start-stop-daemon`. I am fairly certain that at some point, in the throws of desperation, I started chanting `Ph'nglui mglw'nafh Cthulhu R'lyeh wgah'nagl fhtagn`.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/resources/org/springframework/boot/loader/tools/launch.script
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/resources/org/springframework/boot/loader/tools/launch.script
@@ -57,6 +57,9 @@ configfile="$(basename "${jarfile%.*}.conf")"
 # shellcheck source=/dev/null
 [[ -r "${CONF_FOLDER}/${configfile}" ]] && source "${CONF_FOLDER}/${configfile}"
 
+# Initialize ENVIRONMENT_PROPERTIES value if not provided by the config file
+[[ -z "$ENVIRONMENT_PROPERTIES" ]] && ENVIRONMENT_PROPERTIES="{{environmentProperties:}}"
+
 # Initialize PID/LOG locations if they weren't provided by the config file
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR, but I am not keen on the additional complexity that this introduces. There are standard mechanisms in Linux for a child process inheriting the environment from its parent. I’d much we prefer that we use those (passing `--preserve-environment` to `su` for example) than rolling our own.

mmanciop:
@wilkinsona I agree it is not a nice patch (see remarks about Cthulhu). But I truly could not find a more decent way of working with `start-stop-daemon` (only recent distros seem to ship the `--env` flag, and that allows to set just one environment property), and since a clean environment is a good thing for a process, I went for consistency in the `su` case too.

mmanciop:
@philwebb yep, have been doing that for a little while too. But it is much nicer to push it upstream and not having to maintain it on my own ;-)

PR #14680, opened 2018-10-03
refactor: use the gradle runner to set the plugin classpath

jnizet:
This allows simplifying the test build scripts by avoiding to set the classpath in each of them.
It also allows (and requires, as a matter of fact) to use the plugins block to apply the boot plugin
under test.

Unfortunately, this doesn't work for the tests for the reaction to the Kotlin plugin.
See the comments in the GradleBuild class and in each KotlingPluginActionIntegrationTests build script.

See https://github.com/spring-projects/spring-boot/pull/14585#issuecomment-426379856

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/gradle/getting-started/typical-plugins.gradle
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/gradle/getting-started/typical-plugins.gradle
@@ -1,7 +1,5 @@
-buildscript {
-	dependencies {
-		classpath files(pluginClasspath.split(','))
-	}
+plugins {
+	id 'org.springframework.boot' version '{version}'
 }
 
 // tag::apply[]
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/gradle/integrating-with-actuator/build-info-additional.gradle
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/gradle/integrating-with-actuator/build-info-additional.gradle
@@ -1,11 +1,7 @@
-buildscript {
-	dependencies {
-		classpath files(pluginClasspath.split(','))
-	}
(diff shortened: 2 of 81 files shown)


jnizet:
Thanks @wilkinsona!

I've asked a question on the Gradle Slack plugin (in the plugin-development channel) this morning, but I'm still waiting for an answer.

As far as I understand, reacting to the built-in pugins isn't a problem because they're in the "built-in" classpath (not sure about the terminology, but hopefully you get the idea). 

Reacting to the dependency management plugin is fine because it's bundled with the boot plugin and is thus part of the plugin classpath set on the gradle runner.

The Kotlin plugin is the only one that is part of the compile classpath of the boot plugin, but is not bundled with it, and is loaded dynamically by gradle at runtime. If you look at the implementation of the various PluginApplicationAction subclasses, the Kotlin one is the only one where Class.forName() is used to dynamically load the plugin class (and return null if loading the class fails). For all the other ones, a class literal (e.g. `WarPlugin.class`)  is returned directly.

I guess it would be possible to make the tests pass by using reflection to mutate the plugin and its tasks, but I don't think this is a good idea.

If the Gradle guys provide another workaround or a solution, I'll be happy to amend this PR or open a new one.

wilkinsona:
Thanks for the explanation, @jnizet, and once again for the PR. I've merged the changes into master. I've also tweaked things a bit in https://github.com/spring-projects/spring-boot/commit/8c6910cf58955f57078b21217ff98b8946dac92e to allow the runner's plugin classpath to be used for the Kotlin plugin tests too. It's using the same approach as was already used for the Dependency Management plugin.

jnizet:
Thanks for letting me know. I was reluctant to do that because it makes the tests run in conditions that are different from the actual runtime conditions, where the Kotlin plugin is not bundled with the Boot plugin.

wilkinsona:
The runtime conditions seem to be different to the test conditions with either approach. Given that, I thought it was better to take the approach that keep things consistent in the tests.

Issue #14678, opened 2018-10-03
Drop support for "all" from management.metrics.distribution.sla


wilkinsona:
All can apply to different types of meters, some which require a Duration and some which require a long. This means that a value like 30s will be converted to a Duration. If the meter requires a long, they'll end up with the nanos value of the duration. This doesn't make much sense. We should drop support for all so that the value for individual property doesn't have to deal with two different types.

wilkinsona:
Closing in favour of #14684.

PR #14684, opened 2018-10-04
Drop support for "all" from management.metrics.distribution.sla

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14678)

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilter.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilter.java
@@ -19,6 +19,7 @@ package org.springframework.boot.actuate.autoconfigure.metrics;
 import java.util.Arrays;
 import java.util.Map;
 import java.util.Objects;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import io.micrometer.core.instrument.Meter;
@@ -67,7 +68,8 @@ public class PropertiesMeterFilter implements MeterFilter {
 
 	@Override
 	public MeterFilterReply accept(Meter.Id id) {
-		boolean enabled = lookup(this.properties.getEnable(), id, true);
+		boolean enabled = lookup(this.properties.getEnable(), id, true,
+				(all) -> this.properties.getEnable().getOrDefault(all, null));
 		return enabled ? MeterFilterReply.NEUTRAL : MeterFilterReply.DENY;
 	}
 
@@ -80,11 +82,13 @@ public class PropertiesMeterFilter implements MeterFilter {
 	public DistributionStatisticConfig configure(Meter.Id id,
 			DistributionStatisticConfig config) {
 		Distribution distribution = this.properties.getDistribution();
-		return DistributionStatisticConfig.builder()
-				.percentilesHistogram(
-						lookup(distribution.getPercentilesHistogram(), id, null))
-				.percentiles(lookup(distribution.getPercentiles(), id, null))
-				.sla(convertSla(id.getType(), lookup(distribution.getSla(), id, null)))
+		return DistributionStatisticConfig.builder().percentilesHistogram(lookup(
+				distribution.getPercentilesHistogram(), id, null,
+				(all) -> distribution.getPercentilesHistogram().getOrDefault(all, null)))
+				.percentiles(lookup(distribution.getPercentiles(), id, null,
+						(all) -> distribution.getPercentiles().getOrDefault(all, null)))
+				.sla(convertSla(id.getType(),
+						lookup(distribution.getSla(), id, null, (all) -> null)))
 				.build().merge(config);
 	}
 
@@ -98,7 +102,8 @@ public class PropertiesMeterFilter implements MeterFilter {
 		return (converted.length != 0) ? converted : null;
 	}
 
-	private <T> T lookup(Map<String, T> values, Id id, T defaultValue) {
+	private <T> T lookup(Map<String, T> values, Id id, T defaultValue,
+			Function<String, T> all) {
 		if (values.isEmpty()) {
 			return defaultValue;
 		}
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilterTests.java
```diff
@@ -115,14 +116,6 @@ public void acceptWhenHasHigherEnableTrueExactEnableFalseShouldReturnDeny() {
 				.isEqualTo(MeterFilterReply.DENY);
 	}
 
-	@Test
-	public void acceptWhenHasAllEnableFalseShouldReturnDeny() {
```
wilkinsona: This test should be kept and should continue to pass. We only want to drop support for `all` when configuring the SLA. It should be kept for controlling which meters are enabled.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilterTests.java
```diff
@@ -181,14 +174,6 @@ public void configureWhenHasHigherHistogramFalseAndLowerTrueShouldSetPercentiles
 				DistributionStatisticConfig.DEFAULT).isPercentileHistogram()).isTrue();
 	}
 
-	@Test
-	public void configureWhenAllHistogramTrueSetPercentilesHistogramToTrue() {
```
wilkinsona: This test should be kept and should continue to pass.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilterTests.java
```diff
@@ -217,15 +202,6 @@ public void configureWhenHasHigherPercentilesAndLowerShouldSetPercentilesToHighe
 						3.5, 4);
 	}
 
-	@Test
-	public void configureWhenAllPercentilesSetShouldSetPercentilesToValue() {
```
wilkinsona: This test should be kept and should continue to pass.

ayudovin:
@wilkinsona, thank you for your comments. I have updated follow your advice.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilter.java
```diff
@@ -97,7 +102,8 @@ public DistributionStatisticConfig configure(Meter.Id id,
 		return (converted.length != 0) ? converted : null;
 	}
 
-	private <T> T lookup(Map<String, T> values, Id id, T defaultValue) {
+	private <T> T lookup(Map<String, T> values, Id id, T defaultValue,
```
wilkinsona: This doesn't look quite right as `defaultValue` is never used.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilter.java
```diff
@@ -79,11 +82,13 @@ public Id map(Id id) {
...
 			DistributionStatisticConfig config) {
 		Distribution distribution = this.properties.getDistribution();
-		return DistributionStatisticConfig.builder()
-				.percentilesHistogram(
-						lookup(distribution.getPercentilesHistogram(), id, null))
-				.percentiles(lookup(distribution.getPercentiles(), id, null))
-				.sla(convertSla(id.getType(), lookup(distribution.getSla(), id, null)))
+		return DistributionStatisticConfig.builder().percentilesHistogram(lookup(
+				distribution.getPercentilesHistogram(), id, null,
+				(all) -> distribution.getPercentilesHistogram().getOrDefault(all, null)))
```
wilkinsona: I'm not too keen on the repetition of `distribution.getPercentilesHistogram()`.

I think it would be better to have a `lookup` method and another method that's something like `lookupWithFallbackToAll`. Both could delegate to another method to do the actual lookup and could then either just apply the default value, or could fall back to looking for `all` and only then applying the default value if needed.

That would change this code to look something like this:

```java
return DistributionStatisticConfig.builder()
		.percentilesHistogram(lookupWithFallbackToAll(
				distribution.getPercentilesHistogram(), id, null))
		.percentiles(
				lookupWithFallbackToAll(distribution.getPercentiles(), id, null))
		.sla(convertSla(id.getType(), lookup(distribution.getSla(), id, null)))
		.build().merge(config);
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/PropertiesMeterFilter.java
```diff
@@ -66,7 +68,8 @@ private static MeterFilter createMapFilter(Map<String, String> tags) {
 
 	@Override
 	public MeterFilterReply accept(Meter.Id id) {
-		boolean enabled = lookup(this.properties.getEnable(), id, true);
+		boolean enabled = lookup(this.properties.getEnable(), id, true,
+				(all) -> this.properties.getEnable().getOrDefault(all, null));
```
wilkinsona: The defaults are out of sync here. One it `true` and the other is `null`. This means we're missing a test as, with these changes, the following test will fail with a `NullPointerException`:

```java
@Test
public void acceptWhenHasNoMatchingEnabledPropertyShouldReturnNeutral() {
	PropertiesMeterFilter filter = new PropertiesMeterFilter(
			createProperties("enable.something.else=false"));
	assertThat(filter.accept(createMeterId("spring.boot")))
			.isEqualTo(MeterFilterReply.NEUTRAL);
}
```

ayudovin:
@wilkinsona,  I took a look and added changes by your comments

PR #14688, opened 2018-10-05
Fix BeansEndpoint typo in readme

dreis2211:
Hi,

this trivial PR changes `BeansEndpoints` to be `BeansEndpoint` in the docs as the former doesn't seem to exist.

Cheers,
Christoph


snicoll:
Good catch, thanks again Christoph!

PR #14708, opened 2018-10-06
Fix broken Asciidoctor syntax in section on sanitizing values

izeye:
This PR fixes broken Asciidoctor syntax.


wilkinsona:
Thanks once again, @izeye.

PR #14712, opened 2018-10-08
Fix actuator endpoint tables in docs

dreis2211:
Hi,

this PR fixes a missing entry for the `caches` endpoint in the [production-ready-endpoints-exposing-endpoints](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#production-ready-endpoints-exposing-endpoints) section.

While doing that, I noticed that the `integrationgraph` endpoint is documented to be available via Web by default, which doesn't seem to be the case. Maybe I'm missing something here though (in which case I'm happy to revert that particular change).

Let me know what you think.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/production-ready-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/production-ready-features.adoc
@@ -228,6 +228,10 @@ endpoints:
 |Yes
 |No
 
+|`caches`
+|Yes
+|No
+
 |`conditions`
 |Yes
 |No
@@ -262,7 +266,7 @@ endpoints:
 
 |`integrationgraph`
 |Yes
-|Yes
+|No
 
 |`jolokia`
 |N/A


spring-boot-project/spring-boot-docs/src/main/asciidoc/production-ready-features.adoc
```diff
@@ -262,7 +266,7 @@ endpoints:
 
 |`integrationgraph`
 |Yes
-|Yes
+|No
```
snicoll: What makes you think this is not enabled by default? There is a graph only if you're actually using Spring Integration. If you have no graph, there's not much to show so the endpoint is conditional on that. 

So the endpoint is enabled by default but only shows up if you're using SI. Is that confusing?

wilkinsona: This table is for exposure rather than enablement. The endpoint is enabled by default, but I don't believe it is exposed by default. In other words, I agree with @dreis2211 that the table is currently incorrect.

mbhave: I think this is a bit different from exposure or enablement. I think adding the integration graph as `No` in the exposure table, might lead to the misconception that it will be available if `management.endpoints.web.exposure` is set `to *`. But it is conditionally available on SI and setting that flag won't make it available.

dreis2211: The flyway endpoint is very similar here. It requires Flyway of course, but also states a `No`. I don't care too much in the end and do what the team prefers, but setting `integrationgraph` to `No` seems more consistent at the moment.

Setting it to `No` also backs the code handling default exposure. E.g. `WebEndpointAutoConfiguration`
```
	@Bean
	public ExposeExcludePropertyEndpointFilter<ExposableWebEndpoint> webExposeExcludePropertyEndpointFilter() {
		WebEndpointProperties.Exposure exposure = this.properties.getExposure();
		return new ExposeExcludePropertyEndpointFilter<>(ExposableWebEndpoint.class,
				exposure.getInclude(), exposure.getExclude(), "info", "health");
	}
```

Or `ManagementWebSecurityConfigurerAdapter`
```
	@Override
	protected void configure(HttpSecurity http) throws Exception {
		http.authorizeRequests()
				.requestMatchers(
						EndpointRequest.to(HealthEndpoint.class, InfoEndpoint.class))
				.permitAll().anyRequest().authenticated().and().formLogin().and()
				.httpBasic();
	}
```

Both only expose the health and the info endpoint.

Having that said, I think we could do the following:
1. Accept the PR as it is to be more consistent
2. Clarify both the enablement and exposure tables documentation with an additional hint, that some endpoints can be conditional on something. Flyway having Flyway available, `integrationgraph` having SI on the classpath, etc

mbhave: @dreis2211 Ignore my comment. I misread the change. Thanks for the clarification.

snicoll: > This table is for exposure rather than enablement.

I misread that, sorry.

PR #14725, opened 2018-10-09
Allow JSON deserialization of HttpTrace

ferstl:
We need to use a custom `HttpTraceRepository` to persist all HTTP requests and responses for auditing reasons.

Persisting `HttpTrace` objects is not a problem. However, recreating them (`HttpTraceRepository.findAll()`) is not possible in custom `HttpTraceRepository` implementations due to the package-private constructors in `HttpTrace` and its inner classes.

This PR does:
- add a public no-arg constructor to `HttpTrace`
- add public setters for `timestamp`, `request` and `principal`
- add public constructors for the inner classes of `HttpTrace`
- add `@JsonCreator` and `@JsonProperty` annotations for proper deserialization
- add a test for JSON serialization and deserialization

### DIFF
--- a/spring-boot-project/spring-boot-actuator/pom.xml
+++ b/spring-boot-project/spring-boot-actuator/pom.xml
@@ -284,6 +284,11 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
+			<groupId>com.fasterxml.jackson.datatype</groupId>
+			<artifactId>jackson-datatype-jsr310</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks for the proposal. I’m not keen on opening up the `HttpTrace` API like this. We don’t intend `HttpTrace` instances to be created from JSON only serialised to JSON. Making things like the timestamp mutable is also problematic.

Can we please take a step back? Can you open an issue that describes what you’d like to do? I think you want to be able to persist and recreate `HttpTrace` instances on the server-side, although you haven’t said that in so many words. Once we fully understand the problem we can hopefully on some changes that solve that problem.

PR #14744, opened 2018-10-09
@ResponseStatus does not work when using WebFlux

### LINKED ISSUE #14742: WebFlux: @ResponseStatus does not work out of the box, but ResponseStatusException does
As documented in https://github.com/mmanciop/response-status-repro, `@ResponseStatus` does not work out of the box with WebFlux applications, but `ResponseStatusException` does.

I think `ResponseStatusException` works and not `@ResponseStatus` due to `org.springframework.boot.web.reactive.error.DefaultErrorAttributes#determineHttpStatus` and other methods.

I truly cannot tell if this is by design or by oversight. Would it make sense for me to open a PR and add `@ResponseStatus` support to `org.springframework.boot.web.reactive.error.DefaultErrorAttributes`?

mmanciop:
This commit adds support for @ResponseStatus in DefaultErrorAttributes
mimicking the semantics of @ResponseStatus in SpringMVC.

Throwables annotated with @ResponseStatus handled by
DefaultErrorAttributes will result in the following error attributes:
* 'status' set as the return value of the HttpStatus#value()
  defined as @ResponseStatus#value()
* 'error' set to the default reason phrase of the HttpStatus
  defined as @ResponseStatus#value()
* 'message' defined as the value of @ResponseStatus#reason(),
  or the default HttpStatus's reason phrase if left unspecified

Fixes #14742

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/error/DefaultErrorAttributes.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/error/DefaultErrorAttributes.java
@@ -22,9 +22,11 @@ import java.util.Date;
 import java.util.LinkedHashMap;
 import java.util.Map;
 
+import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.http.HttpStatus;
 import org.springframework.validation.BindingResult;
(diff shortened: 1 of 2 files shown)


snicoll:
@mmanciop thank you for making your first contribution to Spring Boot. This is merged in `2.0.x` and `master`.

PR #14774, opened 2018-10-10
Allow ClassPathResources to be filtered by FilteredClassLoader

RoyJacobs:
Hi! Based on a brief discussion on Gitter with @snicoll I've added support for filtering resources in 
`FilteredClassLoader`. This can be used when testing things like `@ConditionalOnResource`.

The current class is a bit tricky to augment since it relies on varargs lists of `Predicate<String>`. This is why I've chosen to follow the same approach and allow filtering `ClassPathResources`, but this means it's not possible to filter both classes *and* resources simultaneously. If this is desired the loader should probably be refactored to use a builder or something, but I didn't want to make this a breaking change.

I've based the pull request on the `2.0.x` branch because the `master` branch doesn't build right now.

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/FilteredClassLoader.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/FilteredClassLoader.java
@@ -20,12 +20,15 @@ import java.net.URL;
 import java.net.URLClassLoader;
 import java.util.function.Predicate;
 
+import org.springframework.core.io.ClassPathResource;
+
 /**
- * Test {@link URLClassLoader} that can filter the classes it can load.
+ * Test {@link URLClassLoader} that can filter the classes and resources it can load.
  *
  * @author Andy Wilkinson
  * @author Stephane Nicoll
  * @author Phillip Webb
+ * @author Roy Jacobs
  * @since 2.0.0
  */
 public class FilteredClassLoader extends URLClassLoader {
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/FilteredClassLoader.java
```diff
@@ -70,6 +82,16 @@ public FilteredClassLoader(Predicate<String>... filters) {
 		return super.loadClass(name, resolve);
 	}
 
+	@Override
+	public URL getResource(String name) {
+		for (Predicate<String> filter : this.filters) {
```
snicoll: This may be overkill but classes and resources predicates are being applied to both use cases with no way for an implementation to know what type of "resource" needs to be filtered.

spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/context/FilteredClassLoaderTests.java
```diff
@@ -16,19 +16,27 @@
...
 /**
  * Tests for {@link FilteredClassLoader}.
  *
  * @author Phillip Webb
+ * @author Roy Jacobs
  */
 public class FilteredClassLoaderTests {
 
+	private static ClassPathResource TEST_RESOURCE = new ClassPathResource(
+			"org/springframework/boot/test/context/FilteredClassLoaderTestResource.txt");
```
snicoll: There's a `s` missing there (`Tests`).

RoyJacobs: Hmm, that's odd. I'd have to revisit the test to make sure that's it actually working as expected then. Will fix.

spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/FilteredClassLoader.java
RoyJacobs: No exactly, this is why I found the current design a bit unfortunate. If there's no objection to changing FilteredClassLoader a bit more (and adjusting the tests currently using it), I could take a stab at refactoring into a builder idea, similar to `ApplicationContextRunner`.

spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/context/FilteredClassLoaderTests.java
snicoll: It's missing a s in the resource as well. All is fine, it's just a naming issue :)

snicoll:
@RoyJacobs thank you for making your first contribution to Spring Boot. I've merged it with a [polish commit](https://github.com/spring-projects/spring-boot/commit/baf83aec0d0c7d72326fe248843af4412a5f3fed). I've decided in the end to split the internals with two filter lists so that only the case where raw filters are specified are affected by that side effect we've discussed.

RoyJacobs:
Fair enough :)

PR #14801, opened 2018-10-12
Customize compression only when it is enabled

wenjiezhang2013:
When the user defines the compression strategy in EmbeddedServletContainerFactory, the compression strategy should only be overridden when there is any enabled compression configuration in the application.properties

### DIFF
--- a/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -211,7 +211,7 @@ public class ServerProperties
 		if (getJspServlet() != null) {
 			container.setJspServlet(getJspServlet());
 		}
-		if (getCompression() != null) {
+		if (getCompression() != null && getCompression().getEnabled()) {
 			container.setCompression(getCompression());
 		}
 		container.setServerHeader(getServerHeader());
--- a/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -92,6 +92,7 @@ import org.springframework.util.StringUtils;
  * @author Quinten De Swaef
  * @author Venil Noronha
  * @author Aurélien Leboulanger
+ * @author Wenjie Zhang
  */
 @ConfigurationProperties(prefix = "server", ignoreUnknownFields = true)
 public class ServerProperties
--- a/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/ServerPropertiesTests.java
+++ b/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/ServerPropertiesTests.java
@@ -57,6 +57,7 @@ import org.mockito.MockitoAnnotations;
 import org.springframework.beans.MutablePropertyValues;
 import org.springframework.boot.bind.RelaxedDataBinder;
 import org.springframework.boot.context.embedded.AbstractEmbeddedServletContainerFactory;
+import org.springframework.boot.context.embedded.Compression;
 import org.springframework.boot.context.embedded.ConfigurableEmbeddedServletContainer;
 import org.springframework.boot.context.embedded.EmbeddedServletContainer;
 import org.springframework.boot.context.embedded.jetty.JettyEmbeddedServletContainer;
@@ -76,6 +77,8 @@ import org.springframework.web.client.ResponseErrorHandler;
 import org.springframework.web.client.RestTemplate;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.junit.Assert.assertEquals;
+import static org.junit.Assert.assertNotEquals;
 import static org.mockito.Matchers.anyBoolean;
 import static org.mockito.Mockito.mock;
 import static org.mockito.Mockito.never;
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@wenjiezhang2013 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14801)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14801) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@wenjiezhang2013 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14801)!

philwebb:
@ wenjiezhang2013 Thanks for your PR but this should already be handled by the servlet container implementations. There's code for [Tomcat](https://github.com/spring-projects/spring-boot/blob/1.5.x/spring-boot/src/main/java/org/springframework/boot/context/embedded/tomcat/TomcatEmbeddedServletContainerFactory.java#L326), [Jetty](https://github.com/spring-projects/spring-boot/blob/8124ac2b79251736aae8997dae78d7f0e0671623/spring-boot/src/main/java/org/springframework/boot/context/embedded/jetty/JettyEmbeddedServletContainerFactory.java#L216) and [Undertow](https://github.com/spring-projects/spring-boot/blob/8124ac2b79251736aae8997dae78d7f0e0671623/spring-boot/src/main/java/org/springframework/boot/context/embedded/undertow/UndertowEmbeddedServletContainer.java#L230).

Have you found a problem with this logic? If so can you please submit a new issue with a small demo project that replicates it. Thanks!

wenjiezhang2013:
Hi Phil,

I have noticed that this issue has been fixed in 2.x, but 1.5.x still has this issue, is there any plan to fix this issue in 1.5.x?

Thanks,
Jack

wenjiezhang2013:
Actually, never mind, I did not noticed you meant this has been fixed in 1.5.x TomcatEmbeddedServletContainerFactory.

wenjiezhang2013:
Actually, this issue still exists.

I have defined a EmbeddedServletContainerFactory like this without defining anything in the application.properties for `server.compression`.

```
@Bean
	public EmbeddedServletContainerFactory tomcatContainer(@Value("${serverBindAddress:0.0.0.0}") final String serverBindAddress,
									@Value("${httpPort:8080}") final String httpPort, @Value("${httpsPort:8443}") final String httpsPort,
									@Value("${sslKeystoreFilePath}") final String sslKeystoreFilePath) {
		logger.debug("Using serverBindAddress --> {}", serverBindAddress);
		logger.debug("Using HTTP Port --> {}", httpPort);
		logger.debug("Using HTTPS Port --> {}", httpsPort);
		logger.debug("Using SSL keystore file path --> {}", sslKeystoreFilePath);

		final TomcatEmbeddedServletContainerFactory tomcatContainerFactory = new TomcatEmbeddedServletContainerFactory();
		tomcatContainerFactory.setContextPath(SalesPlatformConstants.WEB_CONTEXT_PATH);
		tomcatContainerFactory.setPort(Integer.valueOf(httpPort).intValue());

        final Compression compression = new Compression();
        compression.setEnabled(true);
        final String[] mimeTypes = {"application/vnd.apple.ist.platform.resource.territory-detail+json; version=1;charset=utf-8"};
        compression.setMimeTypes(mimeTypes);
        compression.setMinResponseSize(10);

        tomcatContainerFactory.setCompression(compression);
		final Connector sslConnector = this.createSSLConnector(serverBindAddress, httpsPort, sslKeystoreFilePath);

		if (sslConnector != null) {
			tomcatContainerFactory.addAdditionalTomcatConnectors(sslConnector);
		}

		return tomcatContainerFactory;
	}
```


And the problem is in `ServerProperties.java`, 

```

	@NestedConfigurationProperty
	private Compression compression = new Compression();

```

This `compression` will never be null, and if it always overrides my compression configuration in `EmbeddedServletContainerFactory`.

So this change is still needed. Or just change the configuration loading to not create the default 
`Compression` object.

Thanks,
Jack

philwebb:
I see what you mean, we unfortunately don't have a way of good way disabling the `ServerProperties` when you are trying to configure things directly. Can you try and add a test for your specific use-case to this PR?

wenjiezhang2013:
I will add the test case over the weekend, and I have created an issue https://github.com/spring-projects/spring-boot/issues/14803

philwebb:
Thanks, I'll close the issue and we'll just use this PR for discussion.

snicoll:
Can we take a step back and could you please describe the problem you are actually facing? This change prevents the `Compression` customization to be set if it's disabled. It isn't a bad move at all as the three implementations (Tomcat, Jetty, Undertow) have to check whether the compression object is non null and whether compression is enabled. 

They all do so the tests you've added is artificial and only cause an issue if you have a custom `ConfigurableEmbeddedServletContainer` implementation that doesn't check for the `enabled` flag. Is that the case? If not, can you share a sample that shows the problem?

wenjiezhang2013:
The problem is I want to programmatically configure the mime types for compression, so I had to customize `EmbeddedServletContainerFactory ` to achieve that, and the default compression object in `ServerProperties.java` overrode my change. 

I agree with your concern, and I do think the best way to fix this issue is to not create this default objects for any configurations, but I do not know how many places in spring boot has implemented the same, and I think this issue exists in spring boot 2.x as well.

wenjiezhang2013:
Actually, there is another way to achieve what I want to do is to get the `ServerProperties ` instance and modify it, but I am not sure if this is something I should do or not.

snicoll:
>  the default compression object in ServerProperties.java overrode my change.

It does so because your customizer is not ordered. Ours is ordered at 0 so if you want to run after that you need to implement `Ordered`  and provide an order higher than 0. Alternatively you can flag your customizer with `@Order(1)` (or something like that).

I am not sure yet it is worth changing the current customizer. Can you please try that and let us know?

wenjiezhang2013:
No, the `@Order(Ordered.HIGHEST_PRECEDENCE)` did not do the trick.

wilkinsona:
> Ours is ordered at 0 so if you want to run after that you need to implement Ordered and provide an order higher than 0. Alternatively you can flag your customizer with @Order(1) (or something like that).

The default order is `Ordered.LOWEST_PRECEDENCE` so any customiser without an explicitly-configured order should run after ours.

wilkinsona:
There doesn't appear to be a user-configured customiser involved here anyway. There's a custom `EmbeddedServletContainerFactory` factory bean provided by @wenjiezhang2013. Boot's `ServerProperties` bean, by virtue of bean a `EmbeddedServletContainerCustomizer` is the overriding some of the customisation done in the custom `EmbeddedServletContainerFactory` bean definition.

wenjiezhang2013:
Ops, I think misunderstood @snicoll 's point, sorry about that.

@wilkinsona , this approach works, now the question is should we close this PR?

philwebb:
Since you're using 1.5 and we're trying to keep changes there to a minimum let's just close this one. I'm happy you have a work-around.

wenjiezhang2013:
I agree that we do not have to fix this in 1.5.x, but what about 2.x, I reviewed the code in 2.x, i think it has the same issue.

snicoll:
@wenjiezhang2013 registering a customizer in the right order is not a workaround but the recommended approach. It is valid in `2.0.x` and `master` as well.

PR #14815, opened 2018-10-14
add a sample for tiles

tblsoft:
This pull requests contains a new sample for using spring boot with tiles.


wilkinsona:
Thanks for the pull request, but we only provide samples for things that we auto-configure and have dependency management for. Tiles doesn’t fall into that category and we do not have any plans to add support for it.

PR #14816, opened 2018-10-14
Fix copyright end years

izeye:
This PR fixes copyright end years.


wilkinsona:
Thanks once again, @izeye!

Issue #14799, opened 2018-10-12
Rework CouchbaseReactiveHealthIndicator to use AsyncCluster#diagnostics


wilkinsona:
In our investigation of https://github.com/spring-projects/spring-boot/issues/14685 we have concluded that there's no API that we can use to determine Couchbase's health in a timely manner. `CouchbaseReactiveHealthIndicator` was added earlier in 2.1. We should remove it before GA.

wilkinsona:
The Couchbase team have advised us to use Couchbase's `DiagnosticsReport`. In a reactive context, that means that we need to use `AsyncCluster#diagnostics`.

wilkinsona:
When we move over, we need to remove `CouchbaseHealthIndicatorProperties` as they're only used by the reactive indicator.

wilkinsona:
Ok. I'll take care of that as part of merging the fix for #14685 forwards.

wilkinsona:
I don't think we'd want to configure a `Cluster` and an `AsyncCluster` as they won't share resources.

A `CouchbaseCluster` (implementation of `Cluster`) is a blocking wrapper around `CouchbaseAsyncCluster` (implementation of `AsyncCluster`). Unfortunately, there's no public API for creating a `CouchbaseCluster` from ` CouchbaseAsyncCluster` or for getting the `CouchbaseAsyncCluster` from a `CouchbaseCluster`. Short of resorting to reflection, I'm not sure what we can do here. We might be back to dropping the reactive health indicator after all.

simonbasle:
I'm actually surprised not to see an `async()` method on the `Cluster` like there is on the `Bucket` 🤔
I guess using the `Cluster#diagnostics` _could_ be good enough _iff_ the underlying core response doesn't trigger IO every time, but I'm not sure of that. We need @daschl expertise there again 😄

daschl:
@simonbasle I think we can add the `async()` in the next bugfix release if we need to - it just hasn't come up yet (usually folks use the sync cluster and bucket to bootstrap since it's convenient and then drop into `async()` every time they need to do something with rx).

On another note, `diagnostics` right now never touches the network since it just aggregates local state - we could even wrap that into a `Single` and call it a day.

@wilkinsona which approach do you prefer?

daschl:
tracked via https://issues.couchbase.com/browse/JCBC-1259

PR #14817, opened 2018-10-14
Drop CouchbaseReactiveHealthIndicator

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14799)


wilkinsona:
Thanks for the PR. It looks like we may be able to use `DiagnosticsReport` after all. I'll have to explore how that may work in a reactive context or if it's only available as a blocking API.

wilkinsona:
Thanks again for the PR, but we can use `AsyncCluster#diagnostics` in a reactive health indicator for Couchbase.

PR #14822, opened 2018-10-15
Fix Throws typos

Hanope:


philwebb:
What was quick! I only pushed it a few hours ago :) Thanks.

PR #14823, opened 2018-10-15
Support expressing application `args` in `@SpringBootTest`

jgriff:
Add `args` property to the `@SpringBootTest` annotation so tests
can easily supply application arguments to pass to their app under test.

For example,
```
@SpringBootTest(args = {"--some.option=some-value", "other.arg=other-value"})
public class SpringBootTestArgsTest {
	@Autowired
	private ApplicationArguments args;

	@Test
	public void applicationArgumentsPopulated() {
		assertThat(args.getOptionNames()).contains("some.option");
		assertThat(args.getNonOptionArgs()).contains("other.arg=other-value");
	}
}
```

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
@@ -6660,6 +6660,17 @@ NOTE: If you directly use `@ComponentScan` (that is, not through
 `@SpringBootApplication`) you need to register the `TypeExcludeFilter` with it. See
 {dc-spring-boot}/context/TypeExcludeFilter.{dc-ext}[the Javadoc] for details.
 
+[[boot-features-testing-spring-boot-applications-arguments]]
+==== Testing with Application Arguments
+
+If your application expects <<boot-features-application-arguments,arguments>>, you can
+have `@SpringBootTest` inject them using the `args` field.
+
+[source,java,indent=0]
+----
+include::{test-examples}/context/ApplicationArgumentsExampleTests.java[tag=args]
+}
+----
 
 [[boot-features-testing-spring-boot-applications-testing-with-mock-environment]]
 ==== Testing with a mock environment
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for the PR. I think there’s a risk that this will give a false sense of security if a mistake has been made in the application’s main method and the args aren’t passed to `SpringApplication`. That said, I can see the benefit in terms of `ApplicationArguments`.

My feeling at the moment is that it may be better to unit test anything that uses `ApplicationArguments`. Let’s see what the rest of the team thinks.

jgriff:
Yes, I see what you mean on the potential to overlook the responsibility of the app's `main()` method to forward the args to `SpringApplication`.  Also agree on the strategy to unit test consumers of `ApplicationArguments` in isolation.

The primary motivation for this is to support integration tests driven by `@SpringBootTest`.  In particular, those with anonymous `ApplicationRunner` beans.

```
@Bean
public ApplicationRunner doOnStartup() {
        return args -> ... 
    }
}
```

wilkinsona:
We've discussed this today and we think you've convinced us. Tentatively assigning to 2.x.

jgriff:
:)

rfelgent:
Hello people,

may I ask how you @jgriff make the application arguments currently available to your business code / beans ?

Sorry for the off topic, but I am wondering why not using "properties" field of "SpringBootTest" annotation. In my opinion this is suitable for having an easy way of integration tests configuration. 2ndly, this option already exists.

snicoll:
@rfelgent this PR, as the title indicates, is to test [application arguments](https://github.com/spring-projects/spring-boot/blob/d8b487f1277b9117947f07e07023cee6bfc72f81/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ApplicationArguments.java). Those can't be set via the `properties` attribute you mentioned.

snicoll:
Thanks to the environment abstraction, System properties are no different to any other property source so they don’t warrant special treatment. If you’d like to continue discussing these, please come and chat on Gitter or ask a question on Stack Overflow, as we don’t want to derail this PR any further.

snicoll:
@jgriff thank you for making your first contribution to Spring Boot. This is now merged in `master` with  [a polish commit](https://github.com/spring-projects/spring-boot/commit/7413584b0088ce7c9f6227febe50764ef3048a38).

PR #14824, opened 2018-10-15
Make tags providers extend TagsProvider

izeye:
This PR makes tags providers extend `TagsProvider`.

See https://github.com/micrometer-metrics/micrometer/issues/542

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/client/RestTemplateExchangeTagsProvider.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/client/RestTemplateExchangeTagsProvider.java
@@ -17,6 +17,7 @@
 package org.springframework.boot.actuate.metrics.web.client;
 
 import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.TagsProvider;
 
 import org.springframework.http.HttpRequest;
 import org.springframework.http.client.ClientHttpResponse;
@@ -30,7 +31,7 @@ import org.springframework.web.client.RestTemplate;
  * @since 2.0.0
  */
 @FunctionalInterface
-public interface RestTemplateExchangeTagsProvider {
+public interface RestTemplateExchangeTagsProvider extends TagsProvider {
 
 	/**
 	 * Provides the tags to be associated with metrics that are recorded for the given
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks for the PR, @izeye. Generally speaking, I am against marker interfaces. There's no details I the Micrometer issue for why it's needed so it's hard to justify merging this. Perhaps @twicksell or @jkschneider can fill in the blanks.

izeye:
@wilkinsona Thanks for the feedback! When I read the issue, I just imagined it's for discoverability in IDE and simply added it although I wasn't sure it's a usual pattern. I might misunderstand its purpose.

snicoll:
>  I just imagined it's for discoverability

Me too, but that doesn't explain why that's necessary. I don't like the idea of having to implement a marker interface (namely because it is very easy to miss so it's a very loose contract in the end).

jkschneider:
@twicksell Has a definite use case. Could you elaborate?

twicksell:
As @izeye said, purely marker interface to ease discovery of child interfaces and their implementations. There are now 5 or so fo these. I expect many more will be added for metrics around caching, Spring Data repositories, messaging abstractions, etc. So it seemed reasonable to try and tie them together, even if it were purely by convention. 

If people definitely don't like marker interfaces I can see a possibility for TagsProvider becoming a proper interface. Something like:

```
public interface TagsProvider<T> {
   Iterable<Tag> getTags(T context);
   Iterable<Tag> getTags(T context, Throwable ex);
}
```
Where `T` would be `WebMVCRequestContext`, `WebFluxRequestContext`, `JerseyRequestContext` for each implementation. However, that would be a redesign beyond the scope of the original ask.

wilkinsona:
Thanks, @twicksell. I'd argue that the current naming convention of `…TagsProvider` is as good as a marker interface. Both are as likely to be remembered and the naming convention has the advantage of cluttering up the API. I can find all current `…TagsProvider` classes in Eclipse by searching for types named `*TagsProvider`.

I like the idea of a proper interface for `TagsProvider`. It's worth noting that the current marker interface is a step away from that as it would then be a breaking change to update it to specify a proper contract.

For the above reasons, I don't think we should adopt the implementation of `TagsProvider` in Boot while it's a marker interface. FWIW, I'm also in favour of removing it from Micrometer until there's sufficient time to design a proper contract.

twicksell:
> I can find all current …TagsProvider classes in Eclipse by searching for types named *TagsProvider.

That search will also find the [TagsProvider from SpringFox](https://github.com/springfox/springfox/blob/master/springfox-spring-web/src/main/java/springfox/documentation/spring/web/readers/operation/DefaultTagsProvider.java) and however other many unrelated classes happened to use that naming convention. 

But I'm fine with this not being merged as is. I mostly wanted to point out that there are a growing number of these interfaces which all do nearly identical things while having no unifying contract between them. +1 for rethinking the TagsProvider interface and creating a good contract sometime in the future.

wilkinsona:
Ok. Thanks for the feedback, @twicksell, and for the PR, @izeye. I'm going to close this one as it feels like we'll be backing ourselves into a corner for the future introduction of a `TagsProvider` interface with a proper contract.

PR #14830, opened 2018-10-15
Remove usages of BasicAuthorizationInterceptor

### LINKED ISSUE #14798: Remove usages of BasicAuthorizationInterceptor
Hi,

due to the work on [SPR-17326](https://jira.spring.io/projects/SPR/issues/SPR-17326) there are quite some deprecation warnings now caused by BasicAuthorizationInterceptor. I would be willing to tackle that, but I want to sync first on the approach.

1. Deprecate `RestTemplateBuilder.basicAuthorization(String username, String password)`
2. Remove `RestTemplateBuilder.basicAuthorization(String username, String password)` all together
3. Stick to `RestTemplateBuilder.basicAuthorization(String username, String password)` and use `BasicAuthenticationInterceptor` underneath.

For options 1&2, I would include a new method `RestTemplateBuilder.basicAuthentication(String username, String password)` that uses the new `BasicAuthenticationInterceptor`. In order to stay compatible to the current behaviour, I'd like to pass UTF-8 as a charset here, but I'm also fine with sticking to ISO_8859_1 that is used on `HttpHeaders.setBasicAuth` underneath.

Option 1 would be a "softer" variant, but needs work in a later release to eventually remove the leftovers.
Option 2 would be the cleanest in terms of code, but also the hardest from a user perspective.
Option 3 would be probably the simplest, but naming-wise it's not consistent to the new class (authentication vs. authorization).

We could also go for Option 1&3 in order to have consistent naming, and to give users a soft way of migrating.

What do you think?

Cheers,
Christoph

dreis2211:
Hi,

this PR fixes #14798 as discussed.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
@@ -376,7 +376,8 @@ public class RestTemplateBuilder {
 	 * @param username the user name
 	 * @param password the password
 	 * @return a new builder instance
-	 * @deprecated since 2.1.0 in favor of {@link #basicAuthentication(String username, String password)}
+	 * @deprecated since 2.1.0 in favor of
+	 * {@link #basicAuthentication(String username, String password)}
 	 */
 	public RestTemplateBuilder basicAuthorization(String username, String password) {
 		return basicAuthentication(username, password);
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/client/RestTemplateBuilderTests.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/client/RestTemplateBuilderTests.java
@@ -322,7 +322,8 @@ public class RestTemplateBuilderTests {
 
 	@Test
 	public void basicAuthenticationShouldApply() {
-		RestTemplate template = this.builder.basicAuthentication("spring", "boot").build();
+		RestTemplate template = this.builder.basicAuthentication("spring", "boot")
+				.build();
 		ClientHttpRequestInterceptor interceptor = template.getInterceptors().get(0);
 		assertThat(interceptor).isInstanceOf(BasicAuthenticationInterceptor.class);
 		assertThat(interceptor).extracting("username").containsExactly("spring");


dreis2211:
Let me fix the formatting issues. Sorry.

marcosscriven:
@snicoll @dreis2211 - we have an automated system that bumps minor and patch dependencies, on the assumption they follow semver.

As `basicAuthorization` was renamed `basicAuthentication`, and the former was dropped rather than deprecated - this causes a compilation error.

Just wanted to leave this comment here to raise awareness - and for feedback from you on how you approach such changes.

wilkinsona:
Thanks for the feedback.

> As basicAuthorization was renamed basicAuthentication, and the former was dropped rather than deprecated - this causes a compilation error.

I don't think that's what was done here. `basicAuthorization` was deprecated and `basicAuthentication` was introduced alongside it. The deprecated method is still there today in 2.1.x:

https://github.com/spring-projects/spring-boot/blob/2.1.x/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java#L345-L357

> we have an automated system that bumps minor and patch dependencies, on the assumption they follow semver.

I suspect that you have jumped from 2.0.x to 2.2.0. Spring Boot doesn't strictly follow semantic versioning. One way in which we deviate is that we will remove deprecated methods in the next minor release. So, in this case, the method was deprecated in 2.1.x making it eligible for removal in 2.2. This is [mentioned](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.2-Release-Notes#deprecations-from-spring-boot-21) in the release notes.

marcosscriven:
@wilkinsona - thanks for the swift response.

We bumped from 2.1.9 to 2.2.0 - but we only just hit the issue as automated bumps can't do anything about the APIs changes.

Would you be willing to consider supporting such tools by maintaining deprecated method throughout a major version? We do this automatically as we have such a huge number of dependencies across many projects - this enables picking up security patches automatically, while also encouraging devs to use added functionality.

In a perhaps counter-intuitive way, I'd argue this helps move away from deprecated methods *faster*, because the next time a dev visits consumer code, they see the line through it, and can fix it.

wilkinsona:
> Would you be willing to consider supporting such tools by maintaining deprecated methods throughout a major version?

Sorry, but we don't want to do that. We feel that our current approach strikes a good balance between allowing Spring Boot to evolve while also ensuring that minor upgrades are relatively painless.

> In a perhaps counter-intuitive way, I'd argue this helps move away from deprecated methods faster, because the next time a dev visits consumer code, they see the line through it, and can fix it.

There should have been a line through the call to method in question since 2.1.0. Unfortunately, if that warning has been ignored for every 2.1.x release, I don't think that it still being present in 2.2 would have made it any more likely for the necessary change to have been made.

PR #14836, opened 2018-10-15
Set hostname property for JMX in spring-boot-maven-plugin

goughy000:
This fixes an issue seen on some machines where the local hosts file has none-standard entries.  In this case, the build hangs with the following message written to debug until eventually giving up and failing
`Spring application is not ready yet, waiting XXms (attempt Y)`

In `SpringApplicationAdminClient`, the connection URL `service:jmx:rmi:///jndi/rmi://127.0.0.1` is used, however it's possible that the forked JVM has bound to a different address.

Adding this extra system property helps to ensure that the JVM will bind rmi to the correct interface.

We have previously been adding this as a custom jvmarg in the plugin configuration, but I think it makes sense for it to be part of the plugin


snicoll:
@goughy000 thank you for making your first contribution to Spring Boot. This is now merged in `2.0.x` and `master`.

goughy000:
Thanks!!

PR #14849, opened 2018-10-16
Thymeleaf 3.0.10 configuration

danielfernandez:
Thymeleaf 3.0.10 added two new configuration parameters that could be configured in Spring Boot:

   * `spring.thymeleaf.render-hidden-markers-before-checkboxes` will determine whether the hidden elements rendered in forms to mark checkboxes should be rendered before or after (default) the checkbox elements themselves. This is for better integration with some CSS frameworks.
   * `spring.thymeleaf.servlet.produce-partial-output-while-processing` (only for Spring MVC environments) will determine whether partial output should be sent to the server's output buffers as soon as it becomes available (default) or instead the engine should wait until the whole template is processed, keeping results in memory until processing is finished, and then send everything to the output buffers in one call.

This PR adds support for configuring both new flags, and their corresponding test cases.


bclozel:
Right on time - thanks @danielfernandez !

danielfernandez:
Also note that if Thymeleaf 3.0.10 was adopted in 2.0.x too, this might also be useful for that version of Spring Boot.

Issue #14855, opened 2018-10-16
Handle StepRegistry connect/read timeout deprecation


snicoll:
The `read-timeout` and `connect-timout` of `StepRegistry` is deprecated in favour of configuring the client yourself.

Any registry that is a `StepRegistry` is affected. We expose two related properties for each implementation. We should figure out what we want to do with this.

snicoll:
IMO, we have two choices here: 

* Either we cal the deprecated mapping method to the config object and those two properties should be deprecated
* Or, we don't call that mapping any longer and we do whatever is the recommendation in the auto-config

snicoll:
Thanks Jon, Andy told me about that discussion and I believe that this is the same thing as the second option in my previous comment.

wilkinsona:
I'm not sure that we can stop performing the mapping in 2.1 as it will be a breaking change for anyone who is using the auto-configured `…PropertiesConfigAdapter` `…Config` implementation to retrieve the connect or read timeout.

wilkinsona:
We've decided to remove the mappings in the config adapters.

wilkinsona:
Closing in favour of #14886.

PR #14886, opened 2018-10-18
Handle StepRegistry connect/read timeout deprecation

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14855)


wilkinsona:
Thanks for the PR. It looks like you've implemented the second of the [two options that @snicoll listed](https://github.com/spring-projects/spring-boot/issues/14855#issuecomment-430282929) but haven't addressed the mapping of the deprecated properties. I'm [not yet sure](
https://github.com/spring-projects/spring-boot/issues/14855#issuecomment-431038614) what to do about that. Will need to figure that out before we can proceed here.

ayudovin:
@wilkinsona , hi, do you have any updates about this pull request ?

wilkinsona:
@ayudovin We've decided to remove the mappings of the deprecated properties from Boot's various config adapters. Time is tight with 2.1.0.RELEASE due tomorrow. Do you have time to update your PR today? No problem at all if not, we can take care of the additional changes as part of merging this.

ayudovin:
@wilkinsona,  I updated pull request, I hope that I understood your suggestion correct.

wilkinsona:
Thank you for your work on this, @ayudovin. I've merged the changes into master with a [polish commit](https://github.com/spring-projects/spring-boot/commit/290e8886c92b75c50cfd986ef79ffb512164d5e7) that changes the auto-configurations so that their properties are injected once into the class rather than twice into two separate `@Bean` methods.

ayudovin:
@wilkinsona, thank you,I absolutely agree with your polish commit.

PR #14894, opened 2018-10-18
Support '/' in endpoint names

Ivan97:
Update the `EndpointId` constraints to allow '/' in names.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/EndpointId.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/EndpointId.java
@@ -23,7 +23,7 @@ import org.springframework.util.Assert;
 
 /**
  * An identifier for an actuator endpoint. Endpoint IDs may contain only letters, numbers
- * {@code '.'} and {@code '-'}. They must begin with a lower-case letter. Case and syntax
+ * {@code '.'},{@code '-'} and {@code '/'}. They must begin with a lower-case letter. Case and syntax
  * characters are ignored when comparing endpoint IDs.
  *
  * @author Phillip Webb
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@Ivan97 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14894)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14894) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Ivan97 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14894)!

wilkinsona:
Thanks for the PR. Given the ID's use in the path of the endpoint and also in the prefix of its configuration properties, I'm quite reluctant to support `/`. Can you please explain why you need it?

Ivan97:
Thanks for your reply.Here is the reason.In my project,I have this configuration `@Endpoint(id = "verticles/deploy")` to set endpoint url as `/actuator/verticles/deploy`.This configuration works just ok before the version 2.0.6.RELEASE.When I update to the latest version,something went wrong.I think the endpoint url contains '/' may be a common sense.Or maybe I’m using this annotation improperly?

wilkinsona:
We never intended `/` to be used in an endpoint id and 2.0.6 has formalised the rules with the introduction of a new `EndpointId` class.

Mapping an endpoint to `/actuator/verticles/deploy` feels a little strange to me. I would expect that to be a `verticles` endpoint that then uses a `@Selector` on a read operation to handle `deploy` in the same way the Boot's `HealthEndpoint` allows you to [request the health of an individual component](https://github.com/spring-projects/spring-boot/blob/22e722d9f9f885d7fc67d51d16d68d55db6d6efc/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/HealthEndpoint.java#L60).

wilkinsona:
If you really want your endpoint to be mapped to `verticles/deploy` you could use `verticles-deploy` as its ID and then customise its path mapping:

```
management.endpoints.web.path-mapping.verticles-deploy=verticles/deploy
```

Ivan97:
Thank you for your patient.I think I understand what you mean.😃😃😃

wilkinsona:
Great, I'll close this one then as it sounds like you have a way forward and we really don't want to allow `/` in the ID of an endpoint.

PR #14910, opened 2018-10-20
Use Commons Logging for PrometheusPushGatewayManager

izeye:
This PR changes to use Commons Logging for `PrometheusPushGatewayManager` as it seems that SLF4J is used accidentally.


mbhave:
Thanks again @izeye. This is now on master.

Issue #14352, opened 2018-09-07
Autoconfiguration for WebClient and OAuth2 


rwinch:
It would be nice if we could simplify the configuration for WebClient + OAuth2 + Servlet. For example, a user must do this right now:

```java
@Bean
WebClient webClient(ClientRegistrationRepository clientRegistrationRepository, OAuth2AuthorizedClientRepository authorizedClientRepository) {
	ServletOAuth2AuthorizedClientExchangeFilterFunction oauth2 = new ServletOAuth2AuthorizedClientExchangeFilterFunction(clientRegistrationRepository, authorizedClientRepository);
	return WebClient.builder()
			.apply(oauth2.oauth2Configuration())
			.build();
}
```

It would be nice if Spring Boot could auto configure this if there is a ClientRegistrationRepository and OAuth2AuthorizedClientRepository bean available and it is a servlet environment

-----

Description from #14351:

It would be nice if we could simplify the configuration for WebClient + OAuth2 + WebFlux. For example, a user must do this right now:

```
@Bean
WebClient webClient(ReactiveClientRegistrationRepository clientRegistrationRepository,
		ServerOAuth2AuthorizedClientRepository authorizedClientRepository) {
	ServerOAuth2AuthorizedClientExchangeFilterFunction oauth =
			new ServerOAuth2AuthorizedClientExchangeFilterFunction(clientRegistrationRepository, authorizedClientRepository);
	return WebClient.builder()
			.filter(oauth)
			.build();
}
```

It would be nice if Spring Boot could auto configure this if there is a `ReactiveClientRegistrationRepository` and `ServerOAuth2AuthorizedClientRepository` bean available and it is NOT a servlet environment (this includes web applications that are of type none since CLI tools would use this).

snicoll:
> could auto configure this 

Do you mean exposing a `WebClient` bean? We've done that in the past with `RestTemplate` and it turned out to be quite a bad idea (that's where the builder idea comes from). 

Or perhaps you have some kind of dedicated builder in mind?

rwinch:
I can see how you wouldn't want to necessarily expose a `WebClient` bean. Would you consider exposing `ServerOAuth2AuthorizedClientExchangeFilterFunction` as a `Bean` so that users can automatically get all `ExchangeFilterFunction` when defining their `WebClient`?

mbhave:
We could provide a `WebClientCustomizer` that adds the `ExchangeFilterFunction` to the builder if OAuth is present.

[...]

PR #14913, opened 2018-10-20
adding autoconfiguration for WebClient and OAuth2

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14352)


ayudovin:
@philwebb, do you have any updates ?

snicoll:
@ayudovin thanks for your patience. This PR is flagged `for: team meeting` which is happening next week so we'll cover it at that time.

ayudovin:
@snicoll, no problem, thank you for your answer.

bclozel:
#14352 has been declined so I'm closing this issue as a result.

Note that this PR needed more work:

* there was no `@ConditionalOnClass` that protects the auto-configuration; in its current form, applications without Spring Security would break with CNFEs.
* `WebClientOAuth2Customizer` is exposed publicly, it's not needed
* the `setDefaultOAuth2AuthorizedClient` option was enabled, whereas @rwinch said it shouldn't be

Thanks,

PR #14915, opened 2018-10-20
Skip restarter not only for JUnit4 but also for JUnit5

michaldo:
JUnit 5 test fails when starts application and devtools on classpath.

JUnit 4 pass because is treated as exception by DefaultRestartInitializer. It should also apply to JUnit 5


mbhave:
@michaldo Thank you for your first pull request to Spring Boot. This is now on master along with [this](https://github.com/spring-projects/spring-boot/commit/5f24af6c3c27a02ac6c20394875994e95f5fcb62) test.

PR #14916, opened 2018-10-21
Avoid creation of unnecessary collections

dreis2211:
Hi,

this PR replaces occurences of `new HashSet<>(Arrays.asList(element))` with `Collections.singleton(element)` where applicable. (E.g. where we don't change the collection afterwards or where we simply iterate over the collection).

Let me know what you think.
Cheers,
Christoph


mbhave:
Thanks @dreis2211. This is now on master.

Issue #14642, opened 2018-09-30
Ambiguity in Data Access chapter of online documentation


gmazza:
Website:  https://docs.spring.io/spring-boot/docs/current/reference/html/howto-data-access.html#howto-use-two-entity-managers

GitHub page: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc

Current text:  

**80.8 Use Two EntityManagers
Even if the default EntityManagerFactory works fine, you need to define a new one. Otherwise, the presence of the second bean of that type switches off the default. To make it easy to do, you can use the convenient EntityManagerBuilder provided by Spring Boot. Alternatively, you can just the LocalContainerEntityManagerFactoryBean directly from Spring ORM, as shown in the following example:**

Problem with above text:
1.) The title is not clear "Use Two EntityManagers", we're talking about EntityManager*Factories* in this section and "two" seems tied to the specific example in this section but is not generalized for why someone might want to read this section, perhaps "Using Multiple EntityManagerFactories" would be clearer about the point of this section.

2.) This entry paragraph is not making it clear the target audience for it.  The first sentence is written as if it holds for every one using Spring Data, period, because there is no filtering sentence allow the reader to know he can skip reading this section.  If it said something like _"Sometimes you may wish to work with multiple Entity Manager Factories because...  In those cases, ..."_ it's much clearer to the reader that he may skip this section if it doesn't match his need.  

3.) The first sentence is saying that we always need to "define" a new EntityManagerFactory.  "Define" here is ambiguous over whether it means we need to override the default EMF or create a second EntityManagerFactory to work alongside it, nice to have it clarified.

4.) The second sentence is giving a non-sensical reason for why we need to create this second EntityManagerFactory (whether sibling or override), namely that the second's existence would otherwise shut off the default EMF which is working fine for us per the first sentence.  But if we didn't create this second EMF to begin with, how could it shut off the default EMF?

In other words, the second sentence is like saying "If you want to check box A, you must also check box B, because checking box B negates the checking of box A" (something that we don't want given that we decided to check box A to begin with.)

5.) The last sentence "Alternatively, you can just ??? the LCEMFB..." is missing a verb at ???, I'm not sure if you're trying to say "override" or "use" or...?  The missing verb further adds confusion about what we're trying to accomplish in this section.

nicce:
[...]

PR #14928, opened 2018-10-23
Enhance multiple entity manager factories online documentation

nicce:
… gh-14642

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
@@ -2037,12 +2037,12 @@ entity manager in the presence of a bean of that type.
 
 
 
-[[howto-use-two-entity-managers]]
-=== Use Two EntityManagers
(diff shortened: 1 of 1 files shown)


snicoll:
@nicce thank you very much for the PR but I am afraid we can't merge this as is for several reasons. Please do not act on this just yet as we need to make up our mind first.

Team, the section on `EntityManagerBuilder` is wrong. I guess we meant to refer to `EntityManagerFactoryBuilder` and such bean isn't created as soon as we have a custom JPA setup so that hard-coded example simply doesn't work as far as I know. The JPA setup for two (or more) data sources is much more involved, see https://github.com/snicoll-demos/demo-multi-entity-managers/ for a concrete example.

This looks like a lot of code to be put in the documentation. Maybe we could retrofit that project as a sample and link it in the doc. Or we could link to this repo from our doc.

philwebb:
With GA so close we'll have to hold off on this one until 2.2.

snicoll:
@nicce thank you for making your first contribution to Spring Boot.

Issue #14840, opened 2018-10-15
Tighten EndpointId naming rules


philwebb:
Ideally we'd like naming rules for `EndpointId` to be a bit tighter and restrict `-` and `.` chars. We couldn't do this in 2.0.x because of some Spring Cloud endpoints.

Crow-EH:
Hi @philwebb,
Here is my contribution.
Please tell me if something is missing.

philwebb:
@Crow-EH Thanks for your contribution, I've added a note to the PR. Unfortunately this issue description wasn't clear enough. For 2.1 we'd still like to support `-` and `.` but just log a warning.

philwebb:
@spencergibb This might affect Spring Cloud. Will it be possible to migrate those endpoints if we make this change?

Crow-EH:
No problem, thank you for your time @philwebb

spencergibb:
@philwebb it will affect spring cloud. If we restrict the endpoint id can we add an attribute to the annotations that is similar to `management.endpoints.web.path-mapping`. Users will always want to hit /hystrix.stream.

philwebb:
@spencergibb Yeah, we were aware of that one so we introduced #14841 (take a look at [this test](https://github.com/spring-projects/spring-boot/commit/6b37d874975cec8c903360074f9c9342d9b5deb9#diff-b31df53d77834cd947b93db300e70fafR84). Hopefully you can rename the endpoint to `hystrixStream` and register a `PathMapper` bean to supply the default path. The user can still replace the path with properties if they want to.

philwebb:
@spencergibb If it's too late in the day for this change we can defer it until 2.2

spencergibb:
`PathMapper` looks just fine. I think the team would appreciate deferring until 2.2 while logging warnings in 2.1.

philwebb:
Sorry, I meant should we defer the warning logs as well since we're so close to GA?

spencergibb:
I don't mind the warnings.

puce77:
What's the reason for restricting the "-" character? It's pretty common in URLs and good for separating words, IMHO.

wilkinsona:
Every endpoint has a number of configuration properties associated with it and configuration properties should avoid using `-` characters in the prefix, reserving their use for the property's name. Prohibiting the use of `-` in the name of the endpoint ID prevents problems with the configuration properties for the endpoint.

puce77:
@wilkinsona : Thanks for the explanation!

PR #14932, opened 2018-10-23
Tighten EndpointId naming rules

Crow-EH:
The tests have been updated accordingly.
Please do not hesitate to tell me if something is missing.

Fixes gh-14840


philwebb:
Hi @Crow-EH, Thanks for the pull-request. This looks like the ideal end goal that we're trying to achieve but unfortunately I don't think we jump straight there. I'm afraid I didn't put enough detail in the original issue, but for Spring Boot 2.1 we'll want to still allow `-` and `.` in the `EndpointId`, but log a warning if they are used. This will hopefully give users enough time to migrate before we start throwing exceptions in Spring Boot 2.2.

philwebb:
I'll close this one because even if we kept it for 2.2, it will need a fair bit rebasing. Thanks anyway @Crow-EH and sorry the original issue didn't have a better description.

Issue #14484, opened 2018-09-16
JobLauncherCommandLineRunner picks a wrong job instance.


josemalonsom:
In a project in which I am working now we are using an specific parameter as identifier and without  incrementer in a restartable job. The problem that we are facing now is that when we try to launch a new job with a new identifier argument a previous job instance is retrieved, let me explain it:

- we have a job restartable with a specific identifier, let’s say for example a date.
- we launch the job for the date 2018-09-15 and it ends with a failed status.
- we launch the job for a new date, for example 2018-09-16, and we expect that a new job instance for 2018-09-16 would be created, but instead the job instance identified by 2018-09-15 is picked up and the identifier argument is changed to 2018-09-15.

I am not sure if I am missing something about the expected behavior but I would expect that the failed instance 2018-09-15 would be picked up only if we launch the batch with that date as argument but not that by default if a failed job instance exists that job instance is the one that is to be running in subsequent executions independently of the new arguments passed to the job.

I have tracked the problem to here: [JobLauncherCommandLineRunner](https://github.com/spring-projects/spring-boot/blob/ca0de4385c01cbbd120e96ec9f83427fb48c9a15/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherCommandLineRunner.java#L161-L162), it seems that the `JobLauncherCommandLineRunner` uses the `JobParametersBuilder` from Spring Batch to pick the job parameters, but the problem is that `JobParametersBuilder.getNextJobParameters()` doesn’t care about the identifier arguments of the job that it’s being launched, instead, in case of a previous failed job it picks this job instance assuming that you want to continue the previous job instance and without taking into account that you may want to launch a different instance, you can see it here: [JobParametersBuilder](https://github.com/spring-projects/spring-batch/blob/7a3248cabf5eb59def5a217e6d6113c5617cdf70/spring-batch-core/src/main/java/org/springframework/batch/core/JobParametersBuilder.java#L272-L276).

I am not sure if the problem is in Spring Batch’s `JobParametersBuilder` or Spring Boot’s  `JobLauncherCommandLineRunner`, but I think that the behavior of `JobParametersBuilder.getNextJobParameters()` makes sense if you want to act given the previous job state but maybe `JobLauncherCommandLineRunner` shouldn’t assume this because it is not always the case.

As I said maybe I am misunderstanding something but I think there is a bug here.

[...]

PR #14933, opened 2018-10-23
Fix parameters handling on a restart of a Spring Batch job

benas:
This PR resolves gh-14484.

Before this commit, Spring Boot was calling the `JobParametersBuilder#getNextJobParameters`
method from batch. This method was getting the previous parameters of the wrong job instance in a restart scenario.

This commit fixes the issue by first getting the right job instance with the provided parameters, then restarting it if necessary.

To correctly fix the issue reported in gh-14484, a fix needs to be applied on both sides (batch and boot). This PR is coupled with the one opened against batch: https://github.com/spring-projects/spring-batch/pull/660.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
@@ -56,6 +56,7 @@ import org.springframework.util.StringUtils;
  * @author Dave Syer
  * @author Eddú Meléndez
  * @author Kazuki Shimizu
+ * @author Mahmoud Ben Hassine
  */
 @Configuration
 @ConditionalOnClass({ JobLauncher.class, DataSource.class, JdbcOperations.class })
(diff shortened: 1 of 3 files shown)


dturanski:
@benas - I'm seeing the same behavior on a COMPLETED job. Just trying to rerun with a different parameter `localFilePath`.

benas:
@dturanski Does your job has a job parameters incrementer?

dturanski:
@benas - yes RunIdIncrementer. I will try against this PR

philwebb:
The PR to batch is going to be backported so we'll rebase this PR onto 2.0.x

benas:
I see the milestone for this PR changed to 2.0.x. Will this be merged to boot 2.1 too? The PR on batch has been merged in both the master branch for batch 4.1 (for boot 2.1) and the 4.0.x branch (for boot 2.0.x)

philwebb:
@benas Yeah, we always forward merge from the maintenance branches.

aritzbastida:
Hi @benas! Thanks for the PR! Just one small detail, which I think could be simplified a bit:

In a restart scenario, we calculate the job parameters as:
```
	JobParameters previousParameters = lastJobExecution.getJobParameters();
	JobParameters previousIdentifyingParameters = removeNonIdentifying(previousParameters);
	parameters = merge(previousIdentifyingParameters, jobParameters);
```
but, afaik, the result should be exactly the same jobParameters specified as input!! So we could use the jobParameters variable with no further calculation, right?

benas:
@aritzbastida Please let's discuss this point on the gitter channel of Spring Batch as this PR got merged and I will explain in details the code snippet above.

aritzbastida:
Ok, you're right. Where is that? Can you post the link?

benas:
It is here: https://gitter.im/spring-batch/Lobby

PR #14937, opened 2018-10-23
Replace lambdas with method reference or call

dreis2211:
Hi,

this PR replaces some lambdas with method references or calls where possible.

Cheers,
Christoph


snicoll:
Thanks again Christoph!

Issue #14882, opened 2018-10-17
Upgrade to 2.1.0.RC1 breaks message source with DevTools


vpavic:
A Spring Boot app that has a message source in default location (i.e. `messages.properties` in `src/main/resources`) suddenly stopped resolving messages after upgrade to `2.1.0.RC1` when started from IDE.

I've tracked this to enabling of `spring.messages.reloadable` for DevTools that was done in #14699 - I believe this shouldn't have been done with the existing default for `spring.messages.basename` (i.e. `messages`), since [`ReloadableResourceBundleMessageSource`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/ReloadableResourceBundleMessageSource.html) won't resolve the supplied basename from the classpath unless explicitly prefixed with `classpath:`.

Having said that I've tried setting `spring.messages.basename` to `classpath:messages` but that didn't work either.

wilkinsona:
Thanks, @vpavic.

> Having said that I've tried setting spring.messages.basename to classpath:messages but that didn't work either.

This doesn't work because of the custom `ResourceBundleCondition` that causes the auto-configuration to back off as it ends up looking for `classpath*:classpath:messages.properties` which doesn't exist.

wilkinsona:
There are a few problems here:

1. Setting `spring.messages.reloadable: true` makes the default basename of `messages` useless
2. Setting the basename to `classpath:messages` should work with `ReloadableResourceBundleMessageSource` but the custom condition causes the auto-configuration to back off
3. DevTools enables the broken behaviour by default.

wilkinsona:
We've decided to revert #14699 and #13377. With the benefit of hindsight, the approach take in #13377 isn't quite right as it causes problems 1 and 2 described above. We'd like to give ourselves time to consider alternative approaches and, with 2.1.0.RELEASE due tomorrow, the enhancement will have to wait for a future release.

PR #14947, opened 2018-10-24
Use message source in Web UI sample

vpavic:
This makes it possible to reproduce #14882 using a sample app.

I also haven't found any of the sample apps to be using message source, and found `spring-boot-sample-web-ui` to be most fitting for such thing.


philwebb:
Thanks, this should help when we try to fix #14882

PR #14950, opened 2018-10-24
Add starter for OAuth2 resource server

vpavic:
With OAuth2 resource server being available starting with Spring Security 5.1, it would be nice if Spring Boot 2.1 provided starter for this.


philwebb:
Flagged for team attention because I wonder how common the resource server use-case is vs the client.

mbhave:
Quite a common use-case I would say. Adding the starter makes sense to me.

philwebb:
Thanks once again @vpavic! This is now in master.

PR #14971, opened 2018-10-26
Polish mocks with default answer

dreis2211:
Hi,

this PR polishes some `mock()` calls to avoid some unnecessary `withSettings()` calls/imports.

For example the following:
```
BindHandler handler = mock(BindHandler.class,
             withSettings().defaultAnswer(Answers.CALLS_REAL_METHODS));
```
should be equivalent to:
```
BindHandler handler = mock(BindHandler.class, Answers.CALLS_REAL_METHODS);
```

Let me know what you think.
Cheers,
Christoph


philwebb:
Thanks, that certainly reads better.

PR #14972, opened 2018-10-26
Fix MetricsAutoConfigurationTests

dreis2211:
Hi,

while working on another PR I stumbled upon a test failure in MetricsAutoConfigurationTests caused by changes in [micrometer](https://github.com/micrometer-metrics/micrometer/commit/b7bbea3d6935aaf794d538cf5f6da5155c9bb93f#diff-f4ae48de52fffce2dfe4b359fb28ea39R691).

In theory, this should work without the removal of the spy() call (although it's not needed here anyway if you ask me), but it doesn't. With the spy() call the array is empty, without it it's filled. I wonder why, but maybe you can enlighten me. Maybe because the field is volatile now?

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfigurationTests.java
@@ -95,8 +95,7 @@ public class MetricsAutoConfigurationTests {
 
 		@Bean
 		MeterRegistry meterRegistry() {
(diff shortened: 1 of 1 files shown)


philwebb:
I'm surprised the field access worked at all with the `spy` in place. I've no idea what change caused it to be an empty array. Thanks for the PR!

mbhave:
Probably has something to do with the `filters` getting set in an inner class. The spy must've worked before because it was modifying the [same list](https://github.com/micrometer-metrics/micrometer/blob/8a2a0861ed66709e610534bdc6332f0b6395f7a6/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L689) but now points to a [new array](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/MeterRegistry.java#L695).

PR #14975, opened 2018-10-27
Use MeterRegistry.isClosed() for tests

dreis2211:
Hi,

while working on #14972 I noticed that in quite some tests in this package we have a method called `spyOnDisposableBean` that gets duplicated and seems rather complicated - at least for the purpose of just being able to verify that `stop()` is called on MeterRegistry instances.

My initial thought here was to extract this into some utility class to at least get rid of the duplication. The problem here was to find an appropriate place and name for the class. And of course, the complexity of the method remained.

So after giving this a bit more thought, I found a solution that doesn't need reflection and spying at all. We could simply check for `MeterRegistry.isClosed()`, which is slightly different to `stop()`, but still gets the job done if you ask me. The destroyMethod of the underlying bean seems to point to `close()` anyway, so we should be safe here.

Let me know what you think.
Cheers,
Christoph


dreis2211:
Concourse build seems to have stopped simply - I can't see if the failure is related. Local tests run through.

philwebb:
Probably just a CI wobble so don't worry about it. I like the change a lot!

wilkinsona:
Thanks very much, @dreis2211. I've merged to relevant parts of this into 2.0.x, adding in the parts that are specific to Boot 2.1/Micrometer 1.1 in the merge forwards to master.

Issue #14979, opened 2018-10-27
spring-cloud application shuts down silently after exception during startup


jorgefr:
Hi, I found a situation in which an exception interrupts the startup of the application, the SpringApplication class tries to log the error, but it doesn't get written, since the SLF4JBridgeHandler handler was already removed from the root logger.

After discovering it, I found another issue reported where the cause seems to be the same: https://github.com/spring-projects/spring-boot/issues/9457. But in the case I'm describing, the solution seems to be simple, just moving the log instruction some lines above.

I found at least 2 cases when this is happening for me:
I have a spring cloud application that tries to decrypt secrets from the yaml files by using aws-kms. If the KMS client is not able to contact the service, it throws an exception:

```
Cannot decrypt: key=oauth2.clientSecret
Caused by com.amazonaws.services.kms.model.AWSKMSException: The security token included in the request is expired (Service: AWSKMS; Status Code: 400; Error Code: ExpiredTokenException; Request ID: 5033cffe-34a0-4118-986c-8e11aba3b5d3)
```

A similar thing happens when the secret starts with `{cipher}` but the value is not wrapped with quotes.

 In both cases, the exception is captured by SpringApplication, which tries to [log it](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L858), but some lines before, the ApplicationFailedEvent [was triggered](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L828). And one of the listeners triggers a call to `org.springframework.cloud.bootstrap.LoggingSystemShutdownListener`, that removes the `SLF4JBridgeHandler` from the root Logger.

I am using version 2.0.3 of spring-boot.

jorgefr:
After all the investigation, it looks like excluding logging-commons from spring-cloud-stream-binder-kinesis solved the issue

PR #14980, opened 2018-10-27
Log error before emitting application failed event

jorgefr:
This is simplest possible solution.

With the previous implementation, the error never gets logged, because the `SLF4JBridgeHandler` is removed from the root logger as a response to the `ApplicationFailedEvent`

Some details about it:
spring-projects#14979

Obvious Fix

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
@@ -821,6 +821,9 @@ public class SpringApplication {
 			Throwable exception,
 			Collection<SpringBootExceptionReporter> exceptionReporters,
 			SpringApplicationRunListeners listeners) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@jorgefr Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14980)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14980) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jorgefr This Pull Request contains an obvious fix. Signing the Contributor License Agreement is not necessary.

philwebb:
Unfortunately we can't switch the logging in exactly the way that you've suggested. The existing `reportFailure` method only performs logging if an `SpringBootExceptionReporter` doesn't deal with the failure. I'll close this one and we can consider alternatives in #14979

PR #14985, opened 2018-10-28
Allow to disable SSL client authentication on the management port

alonbl:
When server and management are at different ports, and when server requires
TLS client authentication, then there is no simple method to disable TLS
client authentication for management port.

Adding ssl.client-auth=none enables this.

Example:

    server.port=8080
    server.ssl.enabled=true
    server.ssl.client-auth=need
    management.server.port=8081
    management.server.ssl.enabled=true
    management.server.ssl.client-auth=none

Signed-off-by: Alon Bar-Lev <alon.barlev@gmail.com>

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -303,7 +303,7 @@
     },
     {
       "name": "management.server.ssl.client-auth",
-      "description": "Whether client authentication is wanted (\"want\") or needed (\"need\"). Requires a trust store."
+      "description": "Whether client authentication is not wanted (\"none\"), wanted (\"want\") or needed (\"need\"). Requires a trust store."
     },
     {
       "name": "management.server.ssl.enabled",
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -136,7 +136,7 @@
     },
     {
       "name": "server.ssl.client-auth",
-      "description": "Whether client authentication is wanted (\"want\") or needed (\"need\"). Requires a trust store."
+      "description": "Whether client authentication is not wanted (\"none\"), wanted (\"want\") or needed (\"need\"). Requires a trust store."
     },
     {
       "name": "server.ssl.enabled",
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
@@ -234,7 +234,7 @@ content into your application. Rather, pick only the properties that you need.
 	server.servlet.session.timeout=30m # Session timeout. If a duration suffix is not specified, seconds will be used.
 	server.servlet.session.tracking-modes= # Session tracking modes.
 	server.ssl.ciphers= # Supported SSL ciphers.
-	server.ssl.client-auth= # Whether client authentication is wanted ("want") or needed ("need"). Requires a trust store.
+	server.ssl.client-auth= # Whether client authentication is not wanted ("none"), wanted ("want") or needed ("need"). Requires a trust store.
 	server.ssl.enabled=true # Whether to enable SSL support.
 	server.ssl.enabled-protocols= # Enabled SSL protocols.
 	server.ssl.key-alias= # Alias that identifies the key in the key store.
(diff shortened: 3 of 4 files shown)


pivotal-issuemaster:
@alonbl Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14985)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14985) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@alonbl Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=14985)!

snicoll:
@alonbl it's not very clear what this PR does. If you specify a SSL for `server.ssl.client-auth` it won't have an impact on the management config AFAIK. Not having a value already translates to `SslClientAuthMode.NOT_REQUESTED`. 

It would have helped to add a test that exercises the intended change. Am I missing something here?

alonbl:
@snicoll: thanks!

Here is the problem... maybe I am doing wrong solution as I am not deep into spring-boot but more security. And I see now that this is insufficient.

I have the following configuration:

```yaml
server:
  ssl:
    enabled: true
    key-store: file:///tmp/server1.p12
    key-store-password: changeit
management:
  server:
    port: 8081
```
When I try to access management port, I get:
```
$ curl -H -k https://localhost:8081/actuator/info
{}
```
Now I set:
```yaml
server:
  ssl:
    enabled: true
    client-auth: need
    key-store: file:///tmp/server1.p12
    key-store-password: changeit
management:
  server:
    port: 8081
```
And try again:
```
$ curl -H -k https://localhost:8081/actuator/info
curl: (35) error:14094412:SSL routines:ssl3_read_bytes:sslv3 alert bad certificate
```
This means that the SSL setting of server is copied to the management.
Am I right so far?
If so, then I try to disable the setting, using:
```yaml
server:
  ssl:
    enabled: true
    client-auth: need
    key-store: file:///tmp/server1.p12
    key-store-password: changeit
management:
  server:
    port: 8081
    ssl:
      client-auth:
```
Same result, so having an empty is not the same as reverting to null.

I expect to be able to write something like:
```yaml
server:
  ssl:
    enabled: true
    client-auth: need
    key-store: file:///tmp/server1.p12
    key-store-password: changeit
management:
  server:
    port: 8081
    ssl:
      client-auth: none
```
However, I see that every variable at `management.server.ssl` forces to specify all settings, which is not what I expect, as I do like the default behaviour of having a baseline intact.

Based on that, I thought I will write:

```yaml
server:
  ssl:
    enabled: true
    client-auth: need
    key-store: file:///tmp/server1.p12
    key-store-password: changeit
management:
  server:
    port: 8081
    ssl:
      ciphers: ${server.ssl.ciphers:}
      #client-auth: ${server.ssl.client-auth:}
      enabled: ${server.ssl.enabled:}
      enabled-protocols: ${server.ssl.enabled-protocols:}
      key-alias: ${server.ssl.key-alias:}
      key-password: ${server.ssl.key-password:}
      key-store: ${server.ssl.key-store:}
      key-store-password: ${server.ssl.key-store-password:}
      key-store-provider: ${server.ssl.key-store-provider:}
      key-store-type: ${server.ssl.key-store-type:}
      protocol: ${server.ssl.protocol:}
      trust-store: ${server.ssl.trust-store:}
      trust-store-password: ${server.ssl.trust-store-password:}
      trust-store-provider: ${server.ssl.trust-store-provider:}
      trust-store-type: ${server.ssl.trust-store-type:}
```

However, the full environment of `server.ssl` is not available for these variables that are not explicitly set, so substitution fails.

Summary: be able to inherit all settings of `server.ssl` into `management.ssl` and allow override individual keys, mainly the `enabled` and `client-auth`, maybe these should not be in the `server.ssl` and move to `server.{ssl-enabled,ssl-client-auth}`, but this will break existing applications.

Can you think of another solution in which I can configure `client-auth: need` for server and not for management without copying existing properties and/or be immune when someone adds additional properties to server, as I probably want these to be applied to management as well.

Thanks!

bclozel:
In that space, the current behavior is:

* if you don't specify **anything** for `management.server.ssl.*`, then the server customizer for `server.ssl.*` is applied and the management server replicates the main server SSL configuration.
* if you specified **anything** for `management.server.ssl.*`, then this configuration overrides the `server.ssl.*` one.

So I believe this PR can actually be useful, since apparently there is no way to only disable that part of the SSL configuration for the management server.

snicoll:
@alonbl thank you very much for making your first contribution to Spring Boot. I've merged this to `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/33000b6e8e01b1701d10d150a67e7b218b2c6a84)

alonbl:
@snicoll : Thanks!

Now we should think of how to allow setting individual setting in management instead of all-or-nothing copy from the server. Should I open a PR for this discussion?

bclozel:
@alonbl An issue probably, as I'm not sure how this could be achieved without breaking backwards compatibility or making things more complex than necessary.

Are you suggesting that `management.server.ssl.*` should inherit `server.ssl.*` by default and keys should only override what's inherited?

alonbl:
@bclozel: I suggest to have:

`server.ssl-client-auth` as an override of `server.ssl.client-auth` 
`management.server.ssl-client-auth` as an override of `management.server.ssl.client-auth`

Maybe bad naming, but the principal is to move this setting out of the ssl, while maintaining backward compatibility, leaving ssl to only contain the settings for the crypto of the ssl.

Any other idea?

PR #15003, opened 2018-10-29
changes in the readme

pradeepanchan:
mention differences between actuator in spring boot 1.5 and 2.0

### DIFF
--- a/spring-boot-project/spring-boot-actuator/README.adoc
+++ b/spring-boot-project/spring-boot-actuator/README.adoc
@@ -6,6 +6,7 @@ monitor your application using HTTP or JMX endpoints. Auditing, health and metri
 gathering can be automatically applied to your application. The
 https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready[user guide]
 covers the features in more detail.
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@pradeepanchan Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15003)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15003) to manually synchronize the status of this Pull Request.

philwebb:
Thanks for the suggestion, but we generally prefer to keep the documentation in the repo relevant only to the current release.

pradeepanchan:
The problem is that apart from some stackoverflow questions there is no hint that certain configurations are done in a different way in older versions. I hope that there is an easy way for people to know this difference. BTW, we are using spring-boot 1.5 in our company. Only the new micro-services are using 2.0.

philwebb:
@pradeepanchan I understand the concerns, but we also need to keep the documentation relevant and not make new users read too much of the project history. We do document the items that have changes with each release [on the wiki](https://github.com/spring-projects/spring-boot/wiki) and that feels like the best place. Spring Boot 1.x will also be [EOL in August next year](https://spring.io/blog/2018/07/30/spring-boot-1-x-eol-aug-1st-2019) which is another reason not to update the readme.

PR #15009, opened 2018-10-29
Fix deprecation warning

dreis2211:
Hi,

this PR fixes a deprecation warning caused by `HttpClient.wiretap()` that is replaced by `HttpClient.wiretap(true)`.

Cheers,
Christoph


mbhave:
This looks like it was done as part of [this](https://github.com/spring-projects/spring-boot/commit/f3fa20b2d127fd680fc266ca26efdb326c98b501) commit. Thanks anyway, @dreis2211!

PR #15049, opened 2018-10-31
Minor javadoc fix in EnableConfigurationPropertiesImportSelector

mmoayyed:
Removes an extra `it`.


snicoll:
Good catch, thank you.

PR #15051, opened 2018-10-31
Make StaticResourceRequest methods static. Fixes gh-15050

### LINKED ISSUE #15050: StaticResourceRequest.at*() have incorrect javadoc
Looks like `at*()` methods of [org.springframework.boot.autoconfigure.security.servlet.StaticResourceRequest](
https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/security/servlet/StaticResourceRequest.html) are not static, although it's examples in javadoc are shown as if they are:
```java
/**
 * Returns a matcher that includes all commonly used {@link StaticResourceLocation
 * Locations}. The
 * {@link StaticResourceRequestMatcher#excluding(StaticResourceLocation, StaticResourceLocation...)
 * excluding} method can be used to remove specific locations if required. For
 * example: <pre class="code">
 * StaticResourceRequest.atCommonLocations().excluding(StaticResourceLocation.CSS)
 * </pre>
 * @return the configured {@link RequestMatcher}
 */
public StaticResourceRequestMatcher atCommonLocations() {
    return at(EnumSet.allOf(StaticResourceLocation.class));
}
```

Having in mind the fact that its constructor is private, there is no way to use these methods.

I assume those `at*()` methods should be made static to follow the javadoc, unless, of course, I'm missing something...

Thank you in advance!

ykartsev:
Fixes #15050, where StaticResourceRequest.at*() methods were unusable
because of not being static, according to the javadoc.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/reactive/StaticResourceRequest.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/reactive/StaticResourceRequest.java
@@ -57,7 +57,7 @@ public final class StaticResourceRequest {
 	 * </pre>
 	 * @return the configured {@link ServerWebExchangeMatcher}
 	 */
-	public StaticResourceServerWebExchange atCommonLocations() {
+	public static StaticResourceServerWebExchange atCommonLocations() {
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks for the PR but I'm [afraid it's the Javadoc that's wrong](https://github.com/spring-projects/spring-boot/issues/15050#issuecomment-434928003).

Issue #14991, opened 2018-10-29
Provide a configuration property for configuring Tomcat's processor cache


rimal:
Can we please have a way to set a custom value for `processorCache` for embedded tomcat.

wilkinsona:
We can consider adding a configuration property, but you can already configure the size of the processor cache using a `TomcatConnectorCustomizer`. Something like the following:

```java
@Bean
public WebServerFactoryCustomizer<TomcatServletWebServerFactory> processorCacheCustomizer() {
	return (tomcat) -> {
		tomcat.addConnectorCustomizers(
			(connector) -> ((AbstractHttp11Protocol<?>) connector
				.getProtocolHandler()).setProcessorCache(250));
	};
}
```

rimal:
Ok. For now, this will help for sure! Thanks 👍

snicoll:
Closing in favour of PR #15054

philwebb:
We discussed this a little on the call and there's and decided we should expose this as a property. The main reason is the documentation for `processorCache` states:

> The default is 200. A value of `-1` means unlimited. In the unlimited case, the theoretical maximum number of cached Processor objects is`getMaxConnections()` although it will usually be closer to `getMaxThreads()`

This means it's quite likely you're going to want to change this value if you've set `server.tomcat.max-threads` or `server.tomcat.max-connections`.

PR #15054, opened 2018-11-01
Provide a configuration property for configuring Tomcat's processor cache

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/14991)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -381,6 +382,19 @@ public Session getSession() {
 		 */
 		private final Resource resource = new Resource();
 
+		/**
+		 * Configuring processor cache.
```
snicoll: Can you please polish the property description?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -381,6 +382,19 @@ public Session getSession() {
...
+		 */
+		private int processorCache = 200;
```
snicoll: We try to order things a bit more naturally. This property could be moved below `acceptCount`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -381,6 +382,19 @@ public Session getSession() {
...
+
+		public int getProcessorCache() {
```
snicoll: The getter/setter pair should be added in the right order (after the one for `acceptCount`) rather than adding them where they are now.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/embedded/TomcatWebServerFactoryCustomizerTests.java
```diff
@@ -280,6 +280,12 @@ public void testCustomizeMinSpareThreads() {
 		assertThat(this.serverProperties.getTomcat().getMinSpareThreads()).isEqualTo(10);
 	}
 
+	@Test
```
snicoll: Ditto. Please add this test after  `customAcceptCount`

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
```diff
@@ -284,6 +284,7 @@ content into your application. Rather, pick only the properties that you need.
 	server.tomcat.resource.cache-ttl= # Time-to-live of the static resource cache.
 	server.tomcat.uri-encoding=UTF-8 # Character encoding to use to decode the URI.
 	server.tomcat.use-relative-redirects= # Whether HTTP 1.1 and later location headers generated by a call to sendRedirect will use relative or absolute redirects.
+	server.tomcat.processor-cache= # Custom value for Tomcat's processor cache.
```
snicoll: Properties are ordered in natural order. This property must be moved. 

Also, the field description must match the description here.

ayudovin:
@snicoll, thank you for your suggestion, I have updated pull request this suggestions.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -369,6 +370,15 @@ public Session getSession() {
 		 */
 		private int acceptCount = 100;
 
+		/**
+		 * The maximum number of idle processors that will be retained in the cache and
```
snicoll: Thank you but we're not there yet. Could you please look at other keys in `ServerProperties` for inspiration? Key description does not start with `a` or `the` and can't contain javadoc tags as these are not processed by the annotation processed. We don't document default value either as the value is already set there and made available in the metadata.

Looking at your update, this looks ideal to me:

> Maximum number of idle processors that will be retained in the cache and reused with a subsequent request.

Please note that key descriptions are not meant to copy/paste the official documentation of the thing it's configuring.

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
```diff
@@ -276,6 +276,7 @@ content into your application. Rather, pick only the properties that you need.
 	server.tomcat.max-threads=200 # Maximum amount of worker threads.
 	server.tomcat.min-spare-threads=10 # Minimum amount of worker threads.
 	server.tomcat.port-header=X-Forwarded-Port # Name of the HTTP header used to override the original port value.
+	server.tomcat.processor-cache= # The maximum number of idle processors that will be retained in the cache and re-used with a subsequent request. The default is 200. A value of -1 means unlimited.
```
snicoll: This is still a different description than the field javadoc. As indicated previously, they must be identical.

(When a description is a bit more lengthy than usual, we take the first sentence only but that's not the case here).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
ayudovin: ok, thx for advice

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
ayudovin: ok, thx for advice

snicoll:
Thanks @ayudovin, I've merged your contribution with [a polish commit](https://github.com/spring-projects/spring-boot/commit/1b40b0edf112b764db773b671a27414ddcedf3fe).

PR #15055, opened 2018-11-01
Add validation query to DataSourceHealthIndicator details

Aerlant:
When the default validation Query is overwritten it would be nice to show the query in the actuator health endpoint.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/jdbc/DataSourceHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/jdbc/DataSourceHealthIndicator.java
@@ -115,6 +115,7 @@ public class DataSourceHealthIndicator extends AbstractHealthIndicator
 					new SingleColumnRowMapper());
 			Object result = DataAccessUtils.requiredSingleResult(results);
 			builder.withDetail("hello", result);
+			builder.withDetail("validationQuery", validationQuery);
 		}
 	}
 
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/jdbc/DataSourceHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/jdbc/DataSourceHealthIndicator.java
@@ -114,8 +114,9 @@ public class DataSourceHealthIndicator extends AbstractHealthIndicator
 			List<Object> results = this.jdbcTemplate.query(validationQuery,
 					new SingleColumnRowMapper());
 			Object result = DataAccessUtils.requiredSingleResult(results);
-			builder.withDetail("hello", result);
-			builder.withDetail("validationQuery", validationQuery);
+            builder.withDetail("hello", result);
+            builder.withDetail("result", result);
+            builder.withDetail("validationQuery", validationQuery);
 		}
 	}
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@Aerlant Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15055)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15055) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Aerlant Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15055)!

snicoll:
@Aerlant thanks for the PR but I am not sure I understand what you're after. Details of the health endpoint is about detail of the health status and the validation query is a static information for a given app. 

You haven't really shared why that's important. I'll also flag for team attention to see what the rest of the team thinks.

Aerlant:
@snicoll The feature I am after is the ability to see in the health endpoint if I have written my own validation sql. And in this specific case, what SQL ran to return the result. The only indication now is that the default returns HELLO. In my scenario I am writing a query that checks table and column names and returns the count of said query. In an environment with a significant number of applications, it would be nice to have a feature where you could check the health endpoints for non-default query's and what they are.

snicoll:
@Aerlant that doesn't really explain why that's important. The original db health indicator is meant to check that the database is available, the purpose is not to run some kind of "check" on the database itself which seems is what you're doing.

If I got this right, and that's the reason you need to have access to the validation query, have you considered implementing your own indicator? You could then return a custom output and have more flexibility in whatever that check is doing.

Aerlant:
@snicoll The healthcheck I am referring to is currently in its own implementation. I guess it depends on the definition of a healthcheck for the database. Why allow a custom validationQuery/connection-test-query if its always static?

I guess the feature I am asking for is adding a field in the default setup that allows me to setup the db test in my config file instead of creating my own implementation, when all I want to do is see the validation query used when it is not the default.

snicoll:
We've decided on the team call to do this and also created #15076  as a companion.

@Aerlant please complete your PR with a test that exercises the behaviour you've changed.

Aerlant:
@snicoll thank you. I will finish the PR asap.

Aerlant:
@snicoll I have added the test now and proposed an additional "result" line. My reasoning for this is that "hello" no longer gives the right context to what the result it is showing is. I would propose that "hello" should be deprecated for scenarios where a custom query is running. To not break compatability in case anyone is checking on hello, I have added "result" to give proper context to what it is actually showing. Let me know if this is outside of the scope of this PR and if you want me to remove it.

wilkinsona:
FWIW, irrespective of the change here, I'm in favour of migrating away from `hello` to something more descriptive like `result`.

philwebb:
I've raised #15114 to change the "hello" to "result". We'll do that one as a separate issue and commit I think before we merge this one.

snicoll:
@Aerlant thank you for making your first contribution to Spring Boot. I've [polished your contribution](https://github.com/spring-projects/spring-boot/commit/9e84abc46b8ac40dd6a5f5e88c677ac330448b3a) and realized after it would be nice to also have the validation query when the indicator fails. I've added that in 9483c91

Issue #15039, opened 2018-10-31
Consider disabling management.trace.http by default


philwebb:
Currently we only offer `InMemoryHttpTraceRepository` and if actuator is present then we trace every request. This seems a little wasteful so perhaps we should disable support by default.

philwebb:
Devtools could switch it back on perhaps.

philwebb:
Whoops, wrong milestone

philwebb:
See also #8679

mbhave:
We have decided to make the auto-configuration conditional on the presence of a `HttpTraceRepository` bean instead of auto-configuring an `InMemoryTraceRepository`. The `management.trace.http` flag can be used as an opt-out if tracing needs to be disabled even if the bean is present.

PR #15059, opened 2018-11-01
Consider disabling management.trace.http by default

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/15039)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/trace/http/HttpTraceAutoConfiguration.java
```diff
@@ -34,11 +34,12 @@
  * {@link EnableAutoConfiguration Auto-configuration} for HTTP tracing.
  *
  * @author Dave Syer
+ * @author Artsiom Yudovin
  * @since 2.0.0
  */
 @Configuration
 @ConditionalOnWebApplication
-@ConditionalOnProperty(prefix = "management.trace.http", name = "enabled", matchIfMissing = true)
+@ConditionalOnProperty(prefix = "management.trace.http", name = "enabled")
```
snicoll: That effectively means the default value of this property is now `false`. This should be reflected in the metadata, the doc and the appendix

snicoll:
@ayudovin thanks fo the PR but that issue wasn't really ready for contributions. As I've indicated in my initial comment, there are more things to consider than just flipping the value of that property. To make that more clear, we've created a new status "pending-design-work".

Issue #14075, opened 2018-08-15
Batch apps with failed jobs should be able to return a non zero exit code 


cppwfs:
Currently if a Spring Batch Job's `ExitStatus.ExitCode` is FAILED the Boot application ends with a 0 exit code.  However we have had several requests from users that would like for the application to return a non zero exit code if the `ExitStatus.ExitCode` of a Spring Batch Job returns a FAILED.  
This feature should be configurable where the default behavior should remain as is, but a user can configure it such that Boot app returns a non zero exit code if `ExitStatus.ExitCode` is FAILED.

toonborgers:
@cppwfs I'd like to take a look at this. From some quick looking around in the code, I think the solution should be something similar to `org.springframework.boot.autoconfigure.batch.JobExecutionExitCodeGenerator`, but then with the configurable property as result code instead of using the ordinal of the `BatchStatus`, correct?

akoserwal:
@cppwfs: 'getExitCode' is defined as abstract method in the functional interface `ExitCodeGenerator` and updating `getExitCode` type would change the default behaviour. We can add a new method in /batch/JobExecutionExitCodeGenerator such as `String getExitCodeStatus` with option to autoconfigure return code? I can submit a PR. if this seems fine?

nishantraut:
Hi @philwebb Can you please guide me on this, would like to wok if no one else is working on this. Thanks.

philwebb:
@nishantraut I'm afraid I'm not that familiar with the internals of Spring Batch so I'm not too sure what's involved here. Probably something changes are required to `JobExecutionExitCodeGenerator` but I don't know exactly how the Spring Batch `ExitStatus` is obtained. Perhaps @mminella has some advice?

dimitrisli:
PR submitted for this issue: https://github.com/spring-projects/spring-boot/pull/15066

Also requested review from @mminella @benas

benas:
> Currently if a Spring Batch Job's ExitStatus.ExitCode is FAILED the Boot application ends with a 0 exit code.

This is probably because the app is executed with:

```java
public static void main(String[] args) {
    SpringApplication.run(DemoApplication.class, args);
}
```

rather than:

```java
public static void main(String[] args) {
   System.exit(
      SpringApplication.exit(
         SpringApplication.run(DemoApplication.class, args)
      )
   );
}
```

[...]

PR #15066, opened 2018-11-01
#14075 Custom Exit Code on Job Execution Exit Status FAILED

dimitrisli:
1. Adding new property to enable the functionality to provide custom
Exit Code on Job Execution Exit Status FAILED. By default it's false
and no change to the previous functionality which was an Exit Code
according to the Status ordinal.

2. Adding new property with user specific value of Exit Code in case
the functionality via flag (1) is enabled.

3. Loading both these properties into the JobExecutionExitCodeGenerator
and amending the logic there regarding the Exit Code returned.

4. Adding tests.

5. Adding properties documentation.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
@@ -57,6 +57,7 @@ import org.springframework.util.StringUtils;
  * @author Eddú Meléndez
  * @author Kazuki Shimizu
  * @author Mahmoud Ben Hassine
+ * @author Dimitrios Liapis
  */
 @Configuration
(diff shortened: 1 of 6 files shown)


dimitrisli:
@mminella @benas  - may I kindly request review on this please

dimitrisli:
Issue related to this PR: https://github.com/spring-projects/spring-boot/issues/14075

benas:
@dimitrisli As explained in [#14075](https://github.com/spring-projects/spring-boot/issues/14075#issuecomment-435340599), I'm not sure we need a new feature to provide a custom exit code. All we need is a custom `ExitCodeGenerator`  bean, which is actually *the* feature we are looking for. Do you agree?

snicoll:
Thanks for the PR but I don't think we should offer such configurable options when all that it takes is to expose an `ExitCodeGenerator`. We may reconsider the configurable option route later once the scope is narrowed though.

dimitrisli:
@benas agreed yes. We won't need any change here since user can register as per needed in which case this issue/PR can be closed.

Issue #13510, opened 2018-06-18
Prune community starters list from non-2.x compatible ones


bclozel:
Our README file in `spring-boot-project/spring-boot-starters` lists Spring Boot starters maintained by the community.

I think we should take the 2.x generation as an opportunity to remove from that lists starters that are no longer maintained or not compatible with Spring Boot 2.x. We should first discuss what should be the requirements for starters to stay on that list.

snicoll:
Half of the starters on our list have already Spring Boot 2 support ✅(that's awesome!). Some are working on it 🍃  and for others I couldn't figure it out ❓ so I've created an issue.

There are a few starters that do no look supported anymore 🚫 - We should probably review them one more time when we close this and decide what we want to do

:leaves: Apache Camel 2.22 (not yet released)
:white_check_mark: Apache CXF [3.2.0](https://issues.apache.org/jira/browse/CXF-7460?jql=project%20%3D%20CXF%20AND%20text%20~%20%22spring%20boot%202%22)
:question: Apache Qpid
:white_check_mark: Apache Wicket
:leaves: [Axon](https://github.com/AxonFramework/AxonFramework/issues/516)
:white_check_mark: Azure
:white_check_mark: Bucket4j
:white_check_mark: Camunda BPM
:question: Central Authentication Service (CAS) client
:white_check_mark: Charon reverse proxy
:question: Citrus simulator
:question: Cloudant
:question: Couchbase HTTP session
:white_check_mark: DataSource decorating (P6Spy, datasource-proxy, FlexyPool)
:white_check_mark: Docker Java and Docker Client
:white_check_mark: ErroREST exception handler
:leaves: [Feature Flipping For Java (FF4J)](https://github.com/ff4j/ff4j/issues/305)
:white_check_mark: Flowable
:white_check_mark: Google’s reCAPTCHA
:white_check_mark: GraphQL and GraphiQL with GraphQL Java
:white_check_mark: gRPC
:question: HA JDBC
:white_check_mark: Handlebars
:white_check_mark: HDIV
:question: Hiatus for Spring Boot
:question: Infinispan
:question: Jade Templates (Jade4J)
:white_check_mark: JavaMelody monitoring
:white_check_mark: JODConverter
:white_check_mark: JSF integration for various libraries
:leaves: [Liquigraph](https://github.com/liquigraph/liquigraph/issues/211#issuecomment-399063767)
:no_entry_sign: Log4jdbc (doesn't look maintained anymore)
[...]

PR #15082, opened 2018-11-03
Add reference to Hiatus starter

jihor:
Add reference to Hiatus starter.
See https://github.com/spring-projects/spring-boot/issues/13510


snicoll:
@jihor thank you for making your first contribution to Spring Boot.

PR #15089, opened 2018-11-04
Rename logging.file to logging.file.name

### LINKED ISSUE #12510: Ensure all properties map nicely to yaml
```
logging.file=logs/app.log
logging.file.max-history=20 
logging.file.max-size=50MB 
```

how to config these properties in yml file

```
#log 
logging:
  file:  // how to config
    max-history: 30
    max-size: 50MB
```

TwinProduction:
As discussed in #12510 with @philwebb:

-------------------
Using `properties` configuration we can use this:
```properties
logging.file=logs/app.log
logging.file.max-history=20
logging.file.max-size=50MB
```
However, using YAML, we cannot set those at the same time without using the following workaround:  
```yaml
logging.file: logs/app.log
logging:
  file:
    max-history: 20
    max-size: 50MB 
```
----------------------

Thus, to prevent this specific issue from happening with `logging.file`, I suggest `logging.file` should be renamed to `logging.file.name`.

Furthermore, I modified `additional-spring-configuration-metadata.json` accordingly:

The new property entry:
```json
    {
      "name": "logging.file.name",
      "type": "java.lang.String",
      "description": "Log file name (for instance, `myapp.log`). Names can be an exact location or relative to the current directory.",
      "sourceType": "org.springframework.boot.context.logging.LoggingApplicationListener"
    }
```

And to notify users of the breaking change:
```json
    {
      "name": "logging.file",
      "type": "java.lang.String",
      "description": "Log file name (for instance, `myapp.log`). Names can be an exact location or relative to the current directory.",
      "sourceType": "org.springframework.boot.context.logging.LoggingApplicationListener",
      "deprecation": {
        "replacement": "logging.file.name",
        "level": "error"
      }
    }
```

This (partially) fixes #12510

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LogFileWebEndpointAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LogFileWebEndpointAutoConfiguration.java
@@ -63,24 +63,25 @@ public class LogFileWebEndpointAutoConfiguration {
 				AnnotatedTypeMetadata metadata) {
 			Environment environment = context.getEnvironment();
 			String config = environment
-					.resolvePlaceholders("${" + LogFile.FILE_PROPERTY + ":}");
+					.resolvePlaceholders("${" + LogFile.FILE_NAME_PROPERTY + ":}");
 			if (!StringUtils.hasText(config)) {
-				config = environment.resolvePlaceholders(
-						"${" + LogFile.DEPRECATED_FILE_PROPERTY + ":}");
+				config = environment
+						.resolvePlaceholders("${" + LogFile.FILE_PROPERTY + ":}");
 			}
 			ConditionMessage.Builder message = ConditionMessage.forCondition("Log File");
 			if (StringUtils.hasText(config)) {
 				return ConditionOutcome
-						.match(message.found(LogFile.FILE_PROPERTY).items(config));
+						.match(message.found(LogFile.FILE_NAME_PROPERTY).items(config));
 			}
-			config = environment.resolvePlaceholders("${" + LogFile.PATH_PROPERTY + ":}");
+			config = environment
+					.resolvePlaceholders("${" + LogFile.FILE_PATH_PROPERTY + ":}");
 			if (!StringUtils.hasText(config)) {
-				config = environment.resolvePlaceholders(
-						"${" + LogFile.DEPRECATED_PATH_PROPERTY + ":}");
+				config = environment
+						.resolvePlaceholders("${" + LogFile.PATH_PROPERTY + ":}");
 			}
 			if (StringUtils.hasText(config)) {
 				return ConditionOutcome
-						.match(message.found(LogFile.PATH_PROPERTY).items(config));
+						.match(message.found(LogFile.FILE_PATH_PROPERTY).items(config));
 			}
 			config = environment.getProperty("management.endpoint.logfile.external-file");
 			if (StringUtils.hasText(config)) {
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LogFileWebEndpointAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LogFileWebEndpointAutoConfiguration.java
@@ -60,11 +60,11 @@ public class LogFileWebEndpointAutoConfiguration {
 		public ConditionOutcome getMatchOutcome(ConditionContext context,
 				AnnotatedTypeMetadata metadata) {
 			Environment environment = context.getEnvironment();
-			String config = environment.resolvePlaceholders("${logging.file:}");
+			String config = environment.resolvePlaceholders("${logging.file.name:}");
 			ConditionMessage.Builder message = ConditionMessage.forCondition("Log File");
 			if (StringUtils.hasText(config)) {
 				return ConditionOutcome
-						.match(message.found("logging.file").items(config));
+						.match(message.found("logging.file.name").items(config));
 			}
 			config = environment.resolvePlaceholders("${logging.path:}");
 			if (StringUtils.hasText(config)) {
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LogFileWebEndpointAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LogFileWebEndpointAutoConfiguration.java
@@ -66,10 +66,10 @@ public class LogFileWebEndpointAutoConfiguration {
 				return ConditionOutcome
 						.match(message.found("logging.file.name").items(config));
 			}
-			config = environment.resolvePlaceholders("${logging.path:}");
+			config = environment.resolvePlaceholders("${logging.file.path:}");
 			if (StringUtils.hasText(config)) {
 				return ConditionOutcome
-						.match(message.found("logging.path").items(config));
+						.match(message.found("logging.file.path").items(config));
 			}
 			config = environment.getProperty("management.endpoint.logfile.external-file");
 			if (StringUtils.hasText(config)) {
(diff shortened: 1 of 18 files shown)


spring-boot-project/spring-boot/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -68,11 +68,21 @@
...
       "sourceType": "org.springframework.boot.context.logging.LoggingApplicationListener"
     },
+    {
+      "name": "logging.file",
+      "type": "java.lang.String",
+      "description": "Log file name (for instance, `myapp.log`). Names can be an exact location or relative to the current directory.",
+      "sourceType": "org.springframework.boot.context.logging.LoggingApplicationListener",
+      "deprecation": {
+        "replacement": "logging.file.name",
+        "level": "error"
```
snicoll: Sweet 👌

TwinProduction:
@wilkinsona Good point, when I get home, I'll make the change.

spring-boot-project/spring-boot/src/main/resources/META-INF/additional-spring-configuration-metadata.json
TwinProduction: I removed `"level": "error"`. This should do as you suggested, right? Or is there more steps that need to be taken for the value of the old `logging.file` property to be copied over to the new `logging.file.name` property?

TwinProduction:
Fixed @wilkinsona

spring-boot-project/spring-boot/src/main/resources/META-INF/additional-spring-configuration-metadata.json
snicoll: That's only flagging the kind of deprecation of the property. If you switch from error to warning, the code has to read the old property has a fallback. Check the place in your PR where you've changed the property name. 

That said, we haven't make up our mind about that so please hold on until we figure this out.

TwinProduction: Alright, no problem. When the decision has been made, let me know and I'll make the change accordingly.

philwebb:
> I am just wondering if we want to have a deprecation phase

+1

philwebb:
Looks like we have agreement that a deprecation phase would be useful. @TwinProduction Are you interested in also trying to add that to the existing PR?

TwinProduction:
> 
> 
> Looks like we have agreement that a deprecation phase would be useful. @TwinProduction Are you interested in also trying to add that to the existing PR?

Sure, I'll give it a shot.

TwinProduction:
@philwebb I've looked around for a bit, and while I could certainly keeping looking for a while, I think it'll be more time efficient if I just ask you directly instead. 

At a glance, it seems like there's already a property migration implementation that should take care of every properties in the `META-INF/additional-spring-configuration-metadata.json` files with the following:

```json
"deprecation": {
        "replacement": "..."
}
```

If this is not the case, then can you give me an example of a property that went through a deprecation phase? The only ones I could find after a quick search didn't go through a deprecation phase, such as `banner.charset` which was renamed to `spring.banner.charset`, though I might be wrong.

Alternatively, do you simply want me to just hard code it somewhere? I mean it's certainly easy to add an hard coded block that reads the properties `logging.file` as well as `logging.path` and copy them to `logging.file.name` and `logging.file.path` respectively, but I have a feeling there's already a deprecation system in place, in which case, can you point me to the related classes?

Thanks.

snicoll:
@TwinProduction the properties migration is yet another thing that we do to help users but we've decided to deprecate this property in 2.2 and remove it in the feature release. 

Concretely, it means that every place we are looking for  `logging.file.name` in your current PR, we should look for `logging.file` if the former is not set. The property should be flagged as deprecated (with the default error level). 

We should have tests that set the deprecated property and assert the feature still work as expected. Those tests should be flagged as `@Deprecated` so that we get a reminder when the next feature release opens. 

If you have more questions, please head to Gitter and we can chat. Thanks!

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LogFileWebEndpointAutoConfiguration.java
```diff
@@ -60,16 +62,25 @@ public LogFileWebEndpoint logFileWebEndpoint(Environment environment) {
 		public ConditionOutcome getMatchOutcome(ConditionContext context,
 				AnnotatedTypeMetadata metadata) {
 			Environment environment = context.getEnvironment();
-			String config = environment.resolvePlaceholders("${logging.file.name:}");
+			String config = environment
+					.resolvePlaceholders("${" + LogFile.FILE_PROPERTY + ":}");
+			if (!StringUtils.hasText(config)) {
+				config = environment.resolvePlaceholders(
+						"${" + LogFile.DEPRECATED_FILE_PROPERTY + ":}");
+			}
```
TwinProduction: Here I changed `logging.file.name` for `LogFile.DEPRECATED_FILE_PROPERTY` because that field is annotated with `@Deprecated`, which will make it easier to find and remove the deprecated code caused by this PR when the deprecation phase is over.

TwinProduction:
@philwebb @wilkinsona @snicoll What do you think?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LogFile.java
```diff
@@ -39,13 +40,27 @@
 	 * The name of the Spring property that contains the name of the log file. Names can
 	 * be an exact location or relative to the current directory.
 	 */
-	public static final String FILE_PROPERTY = "logging.file";
+	public static final String FILE_PROPERTY = "logging.file.name";
```
wilkinsona: I think this property should be deprecated in favour of `FILE_NAME_PROPERTY` (see below) using `@deprecated` in the javadoc and linking to the new constant. Its value should be left as `logging.file`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LogFile.java
```diff
@@ -39,13 +40,27 @@
...
 
 	/**
 	 * The name of the Spring property that contains the directory where log files are
 	 * written.
 	 */
-	public static final String PATH_PROPERTY = "logging.path";
+	public static final String PATH_PROPERTY = "logging.file.path";
```
wilkinsona: I think this property should be deprecated in favour of `FILE_PATH_PROPERTY` (see below) using `@deprecated` in the javadoc and linking to the new constant. Its value should be left as `logging.path`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LogFile.java
```diff
@@ -39,13 +40,27 @@
...
+
+	/**
+	 * The name of the Spring property that contains the name of the log file. Names can
+	 * be an exact location or relative to the current directory.
+	 */
+	@Deprecated
+	public static final String DEPRECATED_FILE_PROPERTY = "logging.file";
```
wilkinsona: I think this should be named `FILE_NAME_PROPERTY`, should not be deprecated, and its value should be `logging.file.name`. It should be documented as `@since 2.2.0`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LogFile.java
```diff
@@ -39,13 +40,27 @@
...
+
+	/**
+	 * The name of the Spring property that contains the directory where log files are
+	 * written.
+	 */
+	@Deprecated
+	public static final String DEPRECATED_PATH_PROPERTY = "logging.path";
```
wilkinsona: I think this should be named FILE_PATH_PROPERTY, should not be deprecated, and its value should be `logging.file.path`. It should be documented as `@since 2.2.0`.

TwinProduction: Done

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LogFile.java
TwinProduction: Done

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LogFile.java
TwinProduction: Done

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LogFile.java
TwinProduction: Done

snicoll:
@TwinProduction thank you very much for making your first contribution to Spring Boot. I've squashed and [polished your contribution](https://github.com/spring-projects/spring-boot/commit/7939b8b4da247fc3984cb253079372a8fbc49edb) on `master`.

PR #15124, opened 2018-11-07
HANA validation query does not use fully-qualified dummy table name

breglerj:
The validation query should use the fully qualified dummy table name (SYS.DUMMY) to avoid unexpected results if there is a local table named DUMMY


snicoll:
@breglerj thank you for making your first contribution to Spring Boot.

PR #15125, opened 2018-11-07
Add HANA to the lookup of well-known databases

breglerj:
HANA has been added to the `DatabaseDriver` enum with PR #14513, but it is still missing in the `DatabaseLookup` utility class.


snicoll:
Good catch, `HANA` was added in Spring Framework 5.1 and we forgot to update our mapping here. I've merged this one [with a polish commit](https://github.com/spring-projects/spring-boot/commit/089ef9085a472fbb740ee462775420204d93eda7).

PR #15130, opened 2018-11-08
Replace magic number by HttpStatus

holy12345:
Tiny change use `HttpStatus` replace magic number. thanks

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcTags.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcTags.java
@@ -181,16 +181,16 @@ public final class WebMvcTags {
 	public static Tag outcome(HttpServletResponse response) {
 		if (response != null) {
 			int status = response.getStatus();
-			if (status < HttpStatus.OK.value()) {
+			if (status < 200) {
 				return OUTCOME_INFORMATIONAL;
 			}
-			if (status < HttpStatus.MULTIPLE_CHOICES.value()) {
+			if (status < 300) {
 				return OUTCOME_SUCCESS;
 			}
-			if (status < HttpStatus.BAD_REQUEST.value()) {
+			if (status < 400) {
 				return OUTCOME_REDIRECTION;
 			}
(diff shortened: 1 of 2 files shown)


spencergibb:
Isn't there an is1xx, is2xx series of methods?

wilkinsona:
There are indeed, @spencergibb. Thanks for the suggestion. I have no objections to using those. Would you like to do that @holy12345?

holy12345:
Hi First of all thanks very much help me review the code even though this is a very tiny change : )

Thanks @spencergibb give the suggestion(in `HttpStatus.java` there is indeed such a method). but I think @wilkinsona said its make sense so I restored the file(Directly with `200`, `300`, `400`, the code is more readable).

Thanks again patiently explained.

wilkinsona:
Thank you for trying to help, @sepe81, but there’s no need for the commits to be squashed. It isn’t something that we ask our contributors to do as we’re more than happy to do that as part of merging the change.

@holy12345 No changes are needed. We’ll squash and merge this once we’ve created the 2.1.x branch and master has become 2.2.

holy12345:
@wilkinsona Great thanks for quick feedback and patiently explained  : )

sepe81:
@wilkinsona Thank you for the clarification. IMHO it would be helpful to see this explanation under https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc

For the two commits of this PR, I thought squashing would be a good idea b/c most of the changes cancel each other out. Of course this can be done as part of merging but means a bit more work for the reviewer. ;-)

snicoll:
@sepe81 thanks but I don't see the point of adding additional content to that page to mention it's fine if there are several commits. It's better if there is a single commit as you mentioned but we don't want tu burden contributors with too many details and the page is fine as it is from the perspective. Let's stop this conversation here please. If you have more questions, join the community on Gitter.

snicoll:
@holy12345 thank you for making your first contribution to Spring Boot. I've merged your contribution with a [polish commit](https://github.com/spring-projects/spring-boot/commit/f2ca2533d51281f4c27da8231e8ab036e8f88310). I've also updated `WebMvcTags` as discussed in this issue in [a separate commit](https://github.com/spring-projects/spring-boot/commit/a85e27c7cd391d4cbe1297857a8b4cbd0000cc69).

PR #15155, opened 2018-11-12
AnsiOutput support CygW/MingW

hengyunabc:


hengyunabc:
Currently, run spring boot application under `git bash`/`CygW`/`MingW`, AnsiOutput is disabled:

image

After :

image

hengyunabc:
CygW before: 

image

After:

image

wilkinsona:
Thanks for the PR, @hengyunabc. Unfortunately, I don't think we should merge this as the risk of false positives is too high, particularly when running an application on Windows in an IDE. When a false positive occurs console output appears corrupted as the ANSI control codes are output as text.

As an alternative, you can set the `spring.output.ansi.enabled` property to `always` in environments where you are certain that ANSI output is supported but Spring Boot is unable to detect it. If that's broadly applicable, you could set it as an environment variable using `SPRING_OUTPUT_ANSI_ENABLED=always`

PR #15163, opened 2018-11-13
Cookie http-only setting has no effect when using Spring Session

### LINKED ISSUE #15139: Cookie http-only setting has no effect when using Spring Session
Hi, Cookie http-only setting has no effect when using spring boot 2.1.0.RELEASE.
I also tried the method of @wilkinsona in https://github.com/spring-projects/spring-boot/issues/12580!

_Originally posted by @hellowenqi in https://github.com/spring-projects/spring-boot/issues/12580#issuecomment-437268881_

vpavic:
This PR enhances the existing Spring Session auto-configuration with `DefaultCookieSerializer` configuration. This `DefaultCookieSerializer` is configurable using `ServerProperties` i.e. `server.servlet.session.cookie.*` configuration properties.

By doing this, users are able to fully configure `DefaultCookieSerializer` off `server.servlet.session.cookie.*` which removes the limitation of Spring Session's native configuration that is unable to reliably set `httpOnly` and `secure` attributes.

This resolves #15139 - note that even though that issue is assigned to `2.0.x` I've targeted this at `master` branch as it feels like a big change to have so late in the 2.0 lifecycle. The changes here take basically the same approach as in #10440.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
@@ -28,6 +28,8 @@ import org.springframework.boot.WebApplicationType;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.AnyNestedCondition;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
@@ -46,6 +48,7 @@ import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.boot.web.servlet.server.Session.Cookie;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.context.annotation.ImportSelector;
@@ -53,9 +56,10 @@ import org.springframework.core.type.AnnotationMetadata;
 import org.springframework.session.ReactiveSessionRepository;
 import org.springframework.session.Session;
 import org.springframework.session.SessionRepository;
+import org.springframework.session.web.http.CookieHttpSessionIdResolver;
 import org.springframework.session.web.http.CookieSerializer;
 import org.springframework.session.web.http.DefaultCookieSerializer;
-import org.springframework.session.web.http.HeaderHttpSessionIdResolver;
+import org.springframework.session.web.http.HttpSessionIdResolver;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
```diff
@@ -78,6 +84,39 @@
...
+			this.serverProperties = serverProperties;
+		}
+
+		@Bean
+		@ConditionalOnMissingBean({ CookieSerializer.class,
+				HeaderHttpSessionIdResolver.class })
+		public DefaultCookieSerializer cookieSerializer() {
+			Cookie cookie = this.serverProperties.getServlet().getSession().getCookie();
+			DefaultCookieSerializer cookieSerializer = new DefaultCookieSerializer();
+			if (cookie.getName() != null) {
```
philwebb: We could probably use the `PropertyMapper` here:

Something like:

```java
map.from(cookie::getName).to(cookie::setName);
...
```

vpavic: That's very nice - will get to that shortly.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
```diff
@@ -78,6 +85,29 @@
...
 
+		private final ServerProperties serverProperties;
+
+		ServletSessionConfiguration(ServerProperties serverProperties) {
+			this.serverProperties = serverProperties;
+		}
+
+		@Bean
+		@ConditionalOnMissingBean({ CookieSerializer.class,
+				HeaderHttpSessionIdResolver.class })
```
vpavic: After taking a second look I think this would be better off being conditional on missing `CookieSerializer` and conditional on `CookieHttpSessionIdResolver`, because that's the component using the cookie serializer we're configuring here. Otherwise this would switch off when `HeaderHttpSessionIdResolver` bean is present, but not when users provide their custom `HttpSessionIdResolver` implementation.

Thoughts?

wilkinsona: Those are the gotchas that I was referring to in #15139 and that we started to discuss in #10440 before we went with the `SessionCookieConfig`-based approach. I'll need to spend some time refreshing my memory.

vpavic: After giving this some more thought, I've added a commit with custom condition that kicks in if either of the following is true:

* no `HttpSessionIdResolver` and `CookieSerializer` beans are registered - this is the default state with Spring Session
* `CookieHttpSessionIdResolver` bean is registered but `CookieSerializer` is not - this is when users configure a custom `CookieHttpSessionIdResolver` but they don't register the `CookieSerializer`

This covers the case (a quite valid one) when users provide their custom `HttpSessionIdResolver` implementation. That shouldn't result in `CookieSerializer` being configured. There's a test in second commit that covers this scenario.

IMO the options are either to have this composite condition approach, or abandon the case when users provide their own `CookieHttpSessionIdResolver` but not `CookieSerializer` - one might argue that since `CookieHttpSessionIdResolver` itself is `final` and `CookieSerializer` is the only thing can be configure on it that it's an edge case.

Do you have any thoughts on this @wilkinsona?

wilkinsona: Thanks, @vpavic. The scenario that you've described is the one that I needed to refresh my memory about. I like the custom condition approach that you've implemented. This looks ready to merge to me. Thanks for the PR.

wilkinsona:
Thanks again, @vpavic. I've merged the changes in 2.0.x and forwards into master.

PR #15169, opened 2018-11-14
Remove unnecessary echo in release notes script

izeye:
This PR removes an `echo` in the release notes generating script as it doesn’t seem to be necessary.


mbhave:
Good catch @izeye! This is now on master.

PR #15181, opened 2018-11-15
Avoid instantiation of BeanPropertyBinder in Binder

dreis2211:
Hi,

this PR moves an instantiation of `BeanPropertyBinder` in `Binder.bindBean` and therefore avoids unnecessary calls and allocations in case of early returns.

I'm not sure yet if this is side-effect free, though. Tests seem to run, but maybe there is a reason for it being done even though the BeanPropertyBinder is not used eventually. (In which case you can happily decline this PR).

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/Binder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/Binder.java
@@ -333,12 +333,12 @@ public class Binder {
 				|| isUnbindableBean(name, target, context)) {
 			return null;
 		}
-		BeanPropertyBinder propertyBinder = (propertyName, propertyTarget) -> bind(
-				name.append(propertyName), propertyTarget, handler, context, false);
 		Class<?> type = target.getType().resolve(Object.class);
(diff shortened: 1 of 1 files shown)


philwebb:
LGTM, but I'll target to 2.2 just in case.

dreis2211:
Sounds fair. Thanks @philwebb

dreis2211:
@snicoll May I ask about the reasoning of the "on-hold" label?

snicoll:
@dreis2211 sure thing (you can also join us on Gitter if you want to chat). The `2.2.x` line is not opened so we can't do anything with this PR until `master` switches to `2.2.0.BUILD-SNAPSHOT`. Generally speaking, flagging issues properly avoids us to revisit things to see if there's something we have to do.

dreis2211:
@snicoll All fine then. I thought targeting the 2.2.x milestone is enough "labeling" and was confused by the additional one.

snicoll:
@dreis2211 we've had a chat internally following your question and we agree this is confusing. We'll stop doing this in the future and I am also going to remove the ones I've added to make this more clear. Thanks for the feedback!

snicoll:
Thanks again Christoph!

PR #15206, opened 2018-11-18
Find DataSourceProxy even when it is wrapped or proxied 

ttddyy:
In `DataSourceJmxConfiguration` with tomcat `DataSource`, it can only find `DataSourceProxy` if given `DataSource` is a direct child of it.

The check logic uses `instanceof`. When target `DataSource` is wrapped(delegated) or proxied, it cannot find `DataSourceProxy`.

Since `DataSourceProxy#unwrap()` always returns null, here cannot use this method to directly obtain `DataSourceProxy`.

In this PR, updated the check logic to perform the best effort to retrieve `DataSourceProxy`. If given `DataSource` is wrapped or proxied *by spring*, tries to unwrap or get target datasource recursively.


snicoll:
Thanks for the PR @ttddyy. I've merged it with a [polish commit](https://github.com/spring-projects/spring-boot/commit/e424dfbe15570b013f05600389ce73b4f7e6cbf9).

Issue #15199, opened 2018-11-16
Remove redundant dependency management for dom4j:dom4j


wilkinsona:
We were managing it for Hibernate's benefit and it has moved to org.dom4j:dom4j.

charanbir:
This is my first contribution to your codebase. I signed the CLA. Please let me know if I missed anything. Thanks in advance.

$ mvn dependency:tree | grep dom4j
[INFO] |  +- org.dom4j:dom4j:jar:2.1.1:compile (optional)
[INFO] |  +- org.dom4j:dom4j:jar:2.1.1:compile (optional)
[INFO] |  +- org.dom4j:dom4j:jar:2.1.1:compile (optional)
[INFO] |  +- org.dom4j:dom4j:jar:2.1.1:compile (optional)
[INFO] |  +- org.dom4j:dom4j:jar:2.1.1:compile (optional)
[INFO] |  +- org.dom4j:dom4j:jar:2.1.1:compile
[INFO] |  +- org.dom4j:dom4j:jar:2.1.1:compile (optional)

PR #15207, opened 2018-11-18
Remove redundant dependency management for dom4j:dom4j #15199

charanbir:
@wilkinsona


philwebb:
Hi @charanbir, thanks for the efforts here, unfortunately this isn't what we're ultimately looking to do. We want to remove our dependency, not exclude the one from Hibernate. Thanks anyway!

PR #15212, opened 2018-11-19
Configure MessageSource if no "messageSource" bean defined

cac03:
Hello.

This PR fixes an issue with `MessageSourceAutoConfiguration` when a `MessageSource` bean with non standard name defined.

In particular `ApplicationContext` expects a `MessageSource` bean defined only with `"messageSource"` name, but `MessageSourceAutoConfiguration` does not convey the same policy.

So when a `MessageSource` is defined with non-standard name `MessageSourceAutoConfiguration` does not expose a `MessageSource` named `"messageSource"` and at the same time `ApplicationContext` does not use defined `MessageSource` and message resolution process fails.

See [`AbstractApplicationContext.initMessageSource()`](https://github.com/spring-projects/spring-framework/blob/afa38f5f97b0101f4b0f11768ca0c512c0dab45a/spring-context/src/main/java/org/springframework/context/support/AbstractApplicationContext.java#L714).
It is also stated in the [reference documentation](https://docs.spring.io/spring/docs/5.1.2.RELEASE/spring-framework-reference/core.html#context-functionality-messagesource):

> When an ApplicationContext is loaded, it automatically searches for a MessageSource bean defined in the context. **The bean must have the name messageSource.** If such a bean is found, all calls to the preceding methods are delegated to the message source. If no message source is found, the ApplicationContext attempts to find a parent containing a bean with the same name. If it does, it uses that bean as the MessageSource. If the ApplicationContext cannot find any source for messages, an empty DelegatingMessageSource is instantiated in order to be able to accept calls to the methods defined above.

[Project](https://github.com/cac03/spring-boot-messagesource-issue) reproducing the issue.
See created tests. In particular: 
* [MessageSourceWithNonStandardNameTests](https://github.com/cac03/spring-boot-messagesource-issue/blob/master/src/test/java/com/caco3/messagesource/MessageSourceWithNonStandardNameTests.java) demonstrates expected behaviour.
* [MessageSourceWithStandardNameTests](https://github.com/cac03/spring-boot-messagesource-issue/blob/master/src/test/java/com/caco3/messagesource/MessageSourceWithStandardNameTests.java) demonstrates behaviour  when `MessageSource` is defined with `"messagesSource"` name.
* [NotMessageSourceMessageSourceTests](https://github.com/cac03/spring-boot-messagesource-issue/blob/master/src/test/java/com/caco3/messagesource/NotMessageSourceMessageSourceTests.java) demonstrates that a bean named `"messageSource"` must actually be `MessageSource`

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/MessageSourceAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/MessageSourceAutoConfiguration.java
@@ -33,6 +33,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ConditionContext;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.support.AbstractApplicationContext;
 import org.springframework.context.support.ResourceBundleMessageSource;
 import org.springframework.core.Ordered;
 import org.springframework.core.io.Resource;
@@ -49,7 +50,7 @@ import org.springframework.util.StringUtils;
  * @author Eddú Meléndez
  */
 @Configuration
-@ConditionalOnMissingBean(value = MessageSource.class, search = SearchStrategy.CURRENT)
+@ConditionalOnMissingBean(name = AbstractApplicationContext.MESSAGE_SOURCE_BEAN_NAME, search = SearchStrategy.CURRENT)
 @AutoConfigureOrder(Ordered.HIGHEST_PRECEDENCE)
 @Conditional(ResourceBundleCondition.class)
 @EnableConfigurationProperties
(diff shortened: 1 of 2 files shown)


philwebb:
Thanks for the PR. We think that this change makes sense but we'll target it to 2.2.x just in case someone has overridden `AbstractApplicationContext.initMessageSource()` to use different logic and is relying on the existing behavior.

cac03:
@philwebb thank you.
I'm curious to know if you discussed the idea to drop the requirement about `MessageSource` bean name. 
If yes, then did you decide to leave it as is because of other applications relying on this requirement?

philwebb:
@cac03 We didn't discuss removing that requirement. The `AbstractApplicationContext` is part of Spring Framework and I don't think they'd be able to make such a change because of back-compatibility concerns.

snicoll:
@cac03 thank you for making your first contribution to Spring Boot. This is now merged with a [polish commit](https://github.com/spring-projects/spring-boot/commit/ec678eaa3b74d64c28bdcaba6ef5e011033678b6).

PR #15223, opened 2018-11-20
Upgrade to Maven Surefire Plugin 3.0.0-M1

candrews:
Fixes broken manifest jar classloading on latest Debian/Ubuntu Java

See https://issues.apache.org/jira/browse/SUREFIRE-1588


snicoll:
Thanks but we have a semi automated process to upgrade dependencies (including plugins). We'll take this one in due course.

candrews:
I figured that since it was such a prominent issue I'd report it.

I also wasn't sure that your semi automated process would suggest this upgrade because it's a milestone version (not a final/release one).

PR #15225, opened 2018-11-21
Update lombok version

mmoayyed:
Updates lombok version to 1.8.4.


snicoll:
Thanks for the PR but the template for the PR states the following:

> Dependency Upgrades
> Please do not open a pull request for a straightforward dependency upgrade (one that
only updates the version property). We have a semi-automated process for such upgrades
that we prefer to use. However, if the upgrade is more involved (such as requiring
changes for removed or deprecated API) your pull request is most welcome.

Issue #14981, opened 2018-10-27
Document that test slices should not be combined


G-Lem:
Could you have a look at this issue ?
https://stackoverflow.com/questions/52527394/in-spring-boot-2-1-many-test-slices-are-not-allowed-anymore-due-to-multiple-boo?noredirect=1#

philwebb:
To summarize the stackoverflow question, the following test:

```java
@RunWith(SpringRunner.class)
@JdbcTest
@JsonTest
public class MyTest {

	@Test
	public void test() {
		System.out.println("Hello, World !");
	}

}
```

Fails with Spring Boot 2.1 due to:

```
java.lang.IllegalStateException: Configuration error: found multiple declarations of @BootstrapWith for test class [scratch.MyTest]: [@org.springframework.test.context.BootstrapWith(value=class org.springframework.boot.test.autoconfigure.jdbc.JdbcTestContextBootstrapper), @org.springframework.test.context.BootstrapWith(value=class org.springframework.boot.test.autoconfigure.json.JsonTestContextBootstrapper)]
	at org.springframework.test.context.BootstrapUtils.resolveExplicitTestContextBootstrapper(BootstrapUtils.java:166)
	at org.springframework.test.context.BootstrapUtils.resolveTestContextBootstrapper(BootstrapUtils.java:127)
	at org.springframework.test.context.TestContextManager.<init>(TestContextManager.java:124)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.createTestContextManager(SpringJUnit4ClassRunner.java:151)
	at org.springframework.test.context.junit4.SpringJUnit4ClassRunner.<init>(SpringJUnit4ClassRunner.java:142)
	at org.springframework.test.context.junit4.SpringRunner.<init>(SpringRunner.java:49)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance0(Native Method)
	at sun.reflect.NativeConstructorAccessorImpl.newInstance(NativeConstructorAccessorImpl.java:62)
	at sun.reflect.DelegatingConstructorAccessorImpl.newInstance(DelegatingConstructorAccessorImpl.java:45)
	at java.lang.reflect.Constructor.newInstance(Constructor.java:423)
	at org.junit.internal.builders.AnnotatedBuilder.buildRunner(AnnotatedBuilder.java:104)
	at org.junit.internal.builders.AnnotatedBuilder.runnerForClass(AnnotatedBuilder.java:86)
[...]

PR #15310, opened 2018-11-27
Document that test slices should not be combined

dreis2211:
Fixes gh-14981


wilkinsona:

Issue #5550, opened 2016-04-01
When run as a war, spring boot loader .class files can be accessed through http. 


hengyunabc:
Start a spring boot war, then 

``` bash
wget http://localhost:8080/org/springframework/boot/loader/archive/Archive.class
```

When run as a war, the Tomcat context's docBase is set to `.war` path, so the `ServletContext` can getResource from tomcat's `WebResourceRoot`.

``` java
//AbstractEmbeddedServletContainerFactory.java
    /**
     * Returns the absolute document root when it points to a valid directory, logging a
     * warning and returning {@code null} otherwise.
     * @return the valid document root
     */
    protected final File getValidDocumentRoot() {
        File file = getDocumentRoot();
        // If document root not explicitly set see if we are running from a war archive
        file = file != null ? file : getWarFileDocumentRoot();
        // If not a war archive maybe it is an exploded war
        file = file != null ? file : getExplodedWarFileDocumentRoot();
        // Or maybe there is a document root in a well-known location
        file = file != null ? file : getCommonDocumentRoot();
        if (file == null && this.logger.isDebugEnabled()) {
            this.logger
                    .debug("None of the document roots " + Arrays.asList(COMMON_DOC_ROOTS)
                            + " point to a directory and will be ignored.");
        }
        else if (this.logger.isDebugEnabled()) {
            this.logger.debug("Document root: " + file);
        }
        return file;
    }

    private File getWarFileDocumentRoot() {
        return getArchiveFileDocumentRoot(".war");
    }
```

In a traditional war file, all classes/jars are under `/WEB-INF`, can not bd accessed through http. But spring boot maven plugin put spring boot loader classes files into  war `/` path.

philwebb:
Unfortunately the loader classes need to be accessible in order for `java -jar` to work. If you need a fully executable jar you can set `executable` to `false` in Spring Boot's Maven or Gradle plugin.

hengyunabc:
It may cause security problems? 

When service run as `java -jar ./demo.war`, people can access spring boot loader .class files through http. So they can know the service is a spring boot service, and the spring boot version.

philwebb:
@hengyunabc

wilkinsona:
I think you could add a Valve that 404s requests for the loader classes

philwebb:
I wonder if we can do that automatically?

wilkinsona:
[...]

PR #15421, opened 2018-12-07
Fix NPE when tomcat customizer accesses resources

ilya-ershov:
Absence of validDocumentRoot prevents setting resources that leads to NPE in customizer, that tries to load additional resources.
validDocumentRoot can't be resolved if webapp runs as executable Jar without webapp/staic/public directories.
That leads to resources being null. And TomcatContextCustomizer can't load additional resources.
TomcatContextCustomizer can set resources to new StandardRoot instance, but that will lead to disclosure of spring files, fixed in [gh 5550](https://github.com/spring-projects/spring-boot/issues/5550)

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatServletWebServerFactory.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatServletWebServerFactory.java
@@ -184,9 +184,7 @@ public class TomcatServletWebServerFactory extends AbstractServletWebServerFacto
 	protected void prepareContext(Host host, ServletContextInitializer[] initializers) {
 		File documentRoot = getValidDocumentRoot();
 		TomcatEmbeddedContext context = new TomcatEmbeddedContext();
-		if (documentRoot != null) {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ilya-ershov Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15421)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15421) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ilya-ershov Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15421)!

wilkinsona:
Thanks is for the pull request. I think I understand the problem but I am not sure that this is the right way to fix it.

Always setting the context’s resources just in case something wants to add to them will be unnecessary in many cases. I believe there’s a per-request performance cost in having the resources configured so it should be avoided when it is not necessary.

Can you please open an issue describing the problem that you are facing so that we can consider alternative solutions?

ilya-ershov:
Sure, thanks!

PR #15422, opened 2018-12-08
Remove Mockito 1 leftover

dreis2211:
Hi,

while looking through some code that uses reflection, I noticed that in `SpyBeanWithAopProxyTests` there is still a reference to Mockito 1, which isn't supported anymore. This PR removes that leftover.

Have yourself a nice weekend,
Christoph


snicoll:
Thanks again @dreis2211

PR #15424, opened 2018-12-08
Replace explicit generic type with diamond operator

igor-suhorukov:


snicoll:
@igor-suhorukov thanks, I've merged your contribution with a [polish commit](https://github.com/spring-projects/spring-boot/commit/ba29788482b5ce8f62eec659349affc86e281835). If you run the maven build, those formatting failures will show up with an instruction on how to fix them.

Issue #15164, opened 2018-11-13
Automate JDK CI upgrades using AdoptOpenJDK API


philwebb:
We should automate JDK upgrades by using https://api.adoptopenjdk.net/

philwebb:
Niiiiiiice !

PR #15430, opened 2018-12-10
Automate JDK CI upgrades using AdoptOpenJDK API

pkostrzewa:
Updated `setup.sh` script to retrieve binary link to the latest release of JDK in specified version.


wilkinsona:
Thanks for the PR, but this isn’t what is intended by #15164. The changes proposed here will stop builds from being repeatable as a fixed commit will use a changing version of the JDK. The intent of #15164 is to automate the process of detecting a new version of the JDK, updating the relevant Dockerfile, and committing that change. Thanks anyway and sorry for the wasted effort.

PR #15464, opened 2018-12-13
Make HierarchicalNameMapper injectable for GraphiteMeterRegistry

izeye:
This PR changes to make `HierarchicalNameMapper` injectable for `GraphiteMeterRegistry`.

See https://github.com/micrometer-metrics/micrometer/issues/638

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java
@@ -59,8 +59,9 @@ public class GraphiteMetricsExportAutoConfiguration {
 	}
 
 	@Bean
-	@ConditionalOnMissingBean
-	public HierarchicalNameMapper graphiteHierarchicalNameMapper(GraphiteConfig config) {
+	@ConditionalOnMissingBean(HierarchicalNameMapper.class)
+	public GraphiteHierarchicalNameMapper graphiteHierarchicalNameMapper(
+			GraphiteConfig config) {
 		return new GraphiteHierarchicalNameMapper(config.tagsAsPrefix());
 	}
 
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java
@@ -17,7 +17,9 @@
 package org.springframework.boot.actuate.autoconfigure.metrics.export.graphite;
 
 import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.util.HierarchicalNameMapper;
 import io.micrometer.graphite.GraphiteConfig;
+import io.micrometer.graphite.GraphiteHierarchicalNameMapper;
 import io.micrometer.graphite.GraphiteMeterRegistry;
 
 import org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegistryAutoConfiguration;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java
```diff
@@ -58,9 +60,15 @@ public GraphiteConfig graphiteConfig(GraphiteProperties graphiteProperties) {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public GraphiteMeterRegistry graphiteMeterRegistry(GraphiteConfig graphiteConfig,
-			Clock clock) {
-		return new GraphiteMeterRegistry(graphiteConfig, clock);
+	public HierarchicalNameMapper graphiteHierarchicalNameMapper(GraphiteConfig config) {
```
wilkinsona: This should declare that `GraphiteHierarchicalNameMapper` is returned to provide as much type information to the bean factory as possible.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java
```diff
@@ -58,9 +60,15 @@ public GraphiteConfig graphiteConfig(GraphiteProperties graphiteProperties) {
...
+		return new GraphiteHierarchicalNameMapper(config.tagsAsPrefix());
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public GraphiteMeterRegistry graphiteMeterRegistry(GraphiteConfig config, Clock clock,
+			HierarchicalNameMapper nameMapper) {
+		return new GraphiteMeterRegistry(config, clock, nameMapper);
```
wilkinsona: If we're going to make this change for Graphite, should it not also be made for all the other meter registries that take a `HierarchicalNameMapper`?

izeye: @wilkinsona It sounds right to me.

@jkschneider WDYT?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java
izeye: @wilkinsona I applied this requested change in fee2b05c9f85648b76eccf5fcfbb810cac7f83ba.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/graphite/GraphiteMetricsExportAutoConfiguration.java
jkschneider: I believe all other hierarchical implementations don't have a specialized type `GraphiteHierarchicalNameMapper`. At one point, we had this in the configurations but removed it because injecting one name mapper affects them all and it was easy enough to wire your own registry including a custom name mapper.

wilkinsona: Thanks, @jkschneider. So we could type the injection into the `@Bean` method more strongly by using `GraphiteHierarchicalNameMapper` or we could just not do anything and let people configure their own registry. Either way, you're declaring a single bean (your own name mapper or your own registry). I'm tempted to say that it's not worth making a change here and recommending that people just configure their own registry.

wilkinsona: I'd prefer to keep things consistent across registries rather than doing something that's Graphite-specific. A change needs to be made in the Spring Boot legacy support regardless as it's currently too weakly typed.

jkschneider: Removed in the legacy module for consistency in https://github.com/micrometer-metrics/micrometer/commit/659ac962cdb920d0c1f50533075360c1b2646765, which will be included in micrometer-spring-legacy 1.2.0

izeye: @wilkinsona @jkschneider Thanks for the feedback! Closing as discussed.

PR #15487, opened 2018-12-18
Feature/jackson properties initializer

mixaaaa:
Hey,

I already started a request in stackoverflow but it seems to be forgotten a little bit.
In my opinion it would be pretty cool to auto generate object mapper from `JacksonProperties`.

I also made a [small example project](https://github.com/mixaaaa/jackson-demo) using the changes implemented here. Small feedback would be pretty cool :)

### Changes
Moved customization from `JacksonAutoConfiguration.StandardJackson2ObjectMapperBuilderCustomizer` to `JacksonPropeties`

Thanks,
Uwe

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java
@@ -16,23 +16,14 @@
 
 package org.springframework.boot.autoconfigure.jackson;
 
-import java.lang.reflect.Field;
-import java.text.DateFormat;
-import java.text.SimpleDateFormat;
-import java.util.Collection;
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@mixaaaa Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15487)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15487) to manually synchronize the status of this Pull Request.

wilkinsona:
Thanks for the pull request. Our preference is for `@ConfigurationProperties` classes to contain as little logic as possible. Where they have contained logic in the past, we've been slowly moving that logic elsewhere. The proposal here is changing things in the opposite direction which we don't want to do.

If you'd like to pursue this, can you please open an issue that describes the problem that you're facing? Perhaps you could link to the question on Stack Overflow as well.

mixaaaa:
Thanks for having a look

PR #15549, opened 2018-12-21
Fix docs about authorization rules in description of Security features.

qwazer:
HTTP Basic security filter applied first (before Form Based Filter) and checks existence of HTTP header `Authorization`

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
@@ -3283,7 +3283,7 @@ The basic features you get by default in a web application are:
 bean with in-memory store and a single user with a generated password (see
 {dc-spring-boot}/autoconfigure/security/SecurityProperties.User.html[`SecurityProperties.User`]
 for the properties of the user).
-* Form-based login or HTTP Basic security (depending on Content-Type) for the entire
+* Form-based login or HTTP Basic security (if HTTP header `Authorization` present in a HTTP request) for the entire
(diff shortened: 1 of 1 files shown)


qwazer:
@mbhave HI take a look

snicoll:
@qwazer please do not mention team members, especially minutes after having opened a PR.

mbhave:
@qwazer I don't think the proposed change is accurate. The documentation states `depending on Content-Type` because Spring Security uses a `ContentNegotiationStrategy` to determine whether to use form-based login or HTTP basic.  

We can improve the documentation to be more explicit about that. Let us know if you'd like to modify this PR to make that change.

qwazer:
Maybe my PR is not accurate and need to be improved. I allow edits from maintainers. 
I'm trying to improve docs for users without familiarity with Spring Security.

Consider the next use case. "A very busy developer"™ want to set up monitoring with Prometheus for a Spring Boot 2.1 app. He knows:
   
  * Spring Boot app authorize with form-based auth (because it works in the browser)
  * Prometheus cannot use form-based auth but can use HTTP Basic auth.

He searches more info in Spring Boot docs and finds:

> Form-based login or HTTP Basic security (depending on Content-Type)

What this?!! What HTTP Content-Type header he needs to pass to enable Basic Auth. It's crazy!
Current documentation is really confusing!

In my case, I drilling into debug session, spend 1 hour and send this PR to help future users of Spring Boot 2.1 `/actuator` endpoint.

mbhave:
@qwazer I understand that it's a bit vague but I asked if you'd like to modify the existing PR to make things more explicit. I'll close this PR and clarify things as part of #16033.

Issue #15538, opened 2018-12-20
Stop JtaAutoConfiguration from enabling JtaProperties when they are never used


wilkinsona:
`JtaAutoConfiguration` currently enables `JtaProperties`. This means that if its conditions match, the properties will be enabled even if they're never used. Its conditions will match in any app that uses Hibernate as it pulls in the JTA API. The properties are only used if Bitronix or Atomikos are on the classpath which, compared to usage of Hibernate, is relatively unusual. The properties are already enabled by `BitronixJtaConfiguration` and `AtomikosJtaConfiguration` where they're consumed so there's no need for `JtaAutoConfiguration` to do so.

wilkinsona:
Closing in favour of #15550.

PR #15550, opened 2018-12-21
Stop JtaAutoConfiguration from enabling JtaProperties when they are never used #15538

nishantraut:


wilkinsona:
Thanks very much for the PR, @nishantraut. I've merged the changes into 2.1.x and forwards into master.

I had to do a [little bit of polishising](
https://github.com/spring-projects/spring-boot/commit/bfd63ab085eca84596b7c0122a2427678128f436) so that the code was correctly formatted and would therefore build successfully. It's not a big thing, but it would be much appreciated if you could build your changes and correct any such problems before submitting the pull request.

wilkinsona:
Thanks, no problem. As I said, it's not a big thing. FYI, if you want to just build an individual module, you can do something like this:

```
./mvnw -f spring-boot-projects/spring-boot-autoconfigure clean verify
```

PR #15554, opened 2018-12-21
Replace redundant call to class isInstance() with instanceof

igor-suhorukov:


snicoll:
Thanks again @igor-suhorukov

PR #15559, opened 2018-12-24
Replace indexOf() with a call to the contains()

igor-suhorukov:


snicoll:
@igor-suhorukov thanks again!

igor-suhorukov:
@snicoll you are welcome!

PR #15563, opened 2018-12-24
Simplify code by using for-each loop

igor-suhorukov:


snicoll:
@igor-suhorukov thanks again for the PR. I've polished it and removed the change to our shaded version of the json library as I am not keen to modify it.

Issue #15321, opened 2018-11-28
Disable Hibernate entity scanning for default JPA setup


odrotbohm:
As apparent from Hibernate's `MetadataBuildingProcess`, it insists on re-scanning a persistence unit's root URL, even if that already has managed classes defined and configured. In case of that scan picking up an `AttributeConverter` implementation, this is currently directly instantiated (see `ScanningCoordinator.applyScanResultsToManagedResources(…)` and thus bypasses the managed bean integration Spring provides and Boot configures.

In cases in which the reflection based instantiation of the `AttributeConverter` fails, this will break the startup process and the user is left wondering why `SpringBeanContainer` does not handle that instance. Even worse, in case the reflection based instantiation works, the user is left with an instance that didn't get any dependencies injected.

There's a [Hibernate property](https://hibernate.atlassian.net/browse/HHH-12505) that can be set to disable Hibernate's own entity scanning, which – independent of whether Hibernate moves to a managed bean aware processing of the scan result – we should be able to use safely, as our default setup configures a fully prepared persistence unit, so that the (re)scanning is not actually needed. This also improves Hibernate bootstrap performance in the context of a Boot app.

Here's [an example](https://github.com/st-tu-dresden/salespoint/commit/f3ba25cbdc0cf4095138f4ef1a05d1503e0d19af) of how I worked around this issue in a concrete project.

snicoll:
Closing in favour of PR #15565

PR #15565, opened 2018-12-25
Disable Hibernate entity scanning for default JPA setup

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/15321)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
@@ -34,6 +34,7 @@ import org.springframework.util.StringUtils;
  * Configuration properties for Hibernate.
  *
  * @author Stephane Nicoll
+ * @author Artsiom Yudovin
  * @since 2.1.0
  * @see JpaProperties
  */
@@ -56,6 +57,8 @@ public class HibernateProperties {
 	 */
 	private Boolean useNewIdGeneratorMappings;
 
+	private String archiveScanner;
+
 	public String getDdlAuto() {
 		return this.ddlAuto;
 	}
@@ -70,6 +73,14 @@ public class HibernateProperties {
 
 	public void setUseNewIdGeneratorMappings(Boolean useNewIdGeneratorMappings) {
 		this.useNewIdGeneratorMappings = useNewIdGeneratorMappings;
+	}
+
+	public String getArchiveScanner() {
+		return this.archiveScanner;
+	}
+
+	public void setArchiveScanner(String archiveScanner) {
+		this.archiveScanner = archiveScanner;
 	}
 
 	public Naming getNaming() {
@@ -95,6 +106,7 @@ public class HibernateProperties {
 			HibernateSettings settings) {
 		Map<String, Object> result = new HashMap<>(existing);
 		applyNewIdGeneratorMappings(result);
+		applyArchiveScanner(result);
 		getNaming().applyNamingStrategies(result);
 		String ddlAuto = determineDdlAuto(existing, settings::getDdlAuto);
 		if (StringUtils.hasText(ddlAuto) && !"none".equals(ddlAuto)) {
@@ -118,6 +130,12 @@ public class HibernateProperties {
 		}
 		else if (!result.containsKey(AvailableSettings.USE_NEW_ID_GENERATOR_MAPPINGS)) {
 			result.put(AvailableSettings.USE_NEW_ID_GENERATOR_MAPPINGS, "true");
+		}
+	}
+
+	private void applyArchiveScanner(Map<String, Object> result) {
+		if (!StringUtils.isEmpty(this.archiveScanner)) {
+			result.put(AvailableSettings.SCANNER, this.archiveScanner);
 		}
 	}
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
@@ -57,7 +57,12 @@ public class HibernateProperties {
 	 */
 	private Boolean useNewIdGeneratorMappings;
 
-	private String archiveScanner;
+	/**
+	 * Archive Scanner. This is actually a shortcut for the "hibernate.archive.scanner"
+	 * property. When not specified will default to
+	 * "org.hibernate.boot.archive.scan.internal.DisabledScanner".
+	 */
+	private String archiveScanner = "org.hibernate.boot.archive.scan.internal.DisabledScanner";
 
 	public String getDdlAuto() {
 		return this.ddlAuto;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -56,6 +57,8 @@
 	 */
 	private Boolean useNewIdGeneratorMappings;
 
+	private String archiveScanner;
```
snicoll: That's not really what I mind for this. I think that it is safer to disable that behaviour if no property is set so we shouldn't expose a property for this at all.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -121,6 +133,12 @@ else if (!result.containsKey(AvailableSettings.USE_NEW_ID_GENERATOR_MAPPINGS)) {
 		}
 	}
 
+	private void applyArchiveScanner(Map<String, Object> result) {
+		if (!StringUtils.isEmpty(this.archiveScanner)) {
+			result.put(AvailableSettings.SCANNER, this.archiveScanner);
```
snicoll: If the property is not set by the user, we should set it with the disabled scanner implementation.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/orm/jpa/HibernatePropertiesTests.java
```diff
@@ -123,6 +123,15 @@ public void useNewIdGeneratorMappingsFalse() {
 										"false")));
 	}
 
+	@Test
+	public void useArchiveScanner() {
+		this.contextRunner.withPropertyValues(
+				"spring.jpa.hibernate.archive-scanner:org.hibernate.boot.archive.scan.internal.DisabledScanner")
+				.run(assertHibernateProperties((hibernateProperties) -> assertThat(
+						hibernateProperties).containsEntry(AvailableSettings.SCANNER,
```
snicoll: We need two tests here. One where we show this is set by default and one where we customize the property (for instance setting the standard implementation explicitly and then asserting we're using that.

ayudovin: Added test for default implementation

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
ayudovin: set default value for property as disable scanning

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -56,6 +57,13 @@
 	 */
 	private Boolean useNewIdGeneratorMappings;
 
+	/**
+	 * Archive Scanner. This is actually a shortcut for the "hibernate.archive.scanner"
+	 * property. When not specified will default to
+	 * "org.hibernate.boot.archive.scan.internal.DisabledScanner".
+	 */
+	private String archiveScanner = "org.hibernate.boot.archive.scan.internal.DisabledScanner";
```
snicoll: Thanks for the update but that's not what I meant. There shouldn't be any property at all.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -121,6 +138,10 @@ else if (!result.containsKey(AvailableSettings.USE_NEW_ID_GENERATOR_MAPPINGS)) {
 		}
 	}
 
+	private void applyArchiveScanner(Map<String, Object> result) {
+		result.put(AvailableSettings.SCANNER, this.archiveScanner);
```
snicoll: That's not checking if a user has set the property

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
ayudovin: property was removed

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
ayudovin: added checking

snicoll:
@ayudovin that change broke Hibernate 5.2 support as indicated by the build failure. Have you noticed it?

ayudovin:
@snicoll, yes, I try to find solution. Hibernate 2 doesn't have DisabledScanner and I haven't found a solution yet

snicoll:
Thanks @ayudovin. I've [polished your contribution](https://github.com/spring-projects/spring-boot/commit/d0811b48df6ff2d860b96f6540d45f9de53aecd6) and fixed the Hibernate 5.2 problem as well.

odrotbohm:
I wonder if it makes sense to provide a simpler boolean flag as user facing property to avoid Hibernate internals leaking into user application configuration?

snicoll:
@odrotbohm there is not configuration exposed at all so I am not sure what you mean there. If you don't specify anything, we use the disabled scaner impl (which, I believe, matches your original report).

odrotbohm:
Oh, and does it make sense to use the disabled scanner by default. I guess an average Boot user doesn’t expect any Hibernate-driven scanning to run in the first place anyway?

snicoll:
> Oh, and does it make sense to use the disabled scanner by default. 

Is that a question? If that is, you lost me as you created an issue to request that scanning to be disabled by default. Unless I misunderstood it. If disabling the scanner is not 100% safe, we should revert this and clarify the scope of the issue you've created.

odrotbohm:
Okay, maybe the changes view is messed up then. I see tests that set a pretty complicated property that directly configured the scanner. Given our experience with Hibernate those things change quite frequently so that I wondered whether something like `.…disable-hibernate-scanning` to either be `true` or `false` allow us to mitigate potential changes.

odrotbohm:
I was just looking at what the changes tab provided on mobile view and that doesn’t seem to contain anything hinting at what default you chose.

snicoll:
There is nothing in your request that indicated this should be configurable. We can introduce a property but that doesn't match the description of #15321. I've already indicated that if disabling the scanner in the context of a Spring Boot auto-configured app wasn't safe, I think the scope of the issue should be clarified (with a potential reopen).

As for the "pretty complicated" property, it's just the standard Hibernate property with a prefix to help us grab it.

Issue #9776, opened 2017-07-17
Make JsonContentAssert filter stack traces


philwebb:
If we use `AbstractAssert.throwAssertionError` then stack trace elements will get filtered.

snicoll:
Closing in favour of PR #15569

PR #15569, opened 2018-12-26
Use assertJ to generate AssertionError and filter its stacktrace

cvienot:
Closes gh-9776

Rely on AssertJ to generate and throw AssertionError, taking advantage of stacktrace filtering.


snicoll:
@cvienot thank you for making your first contribution to Spring Boot.

PR #15584, opened 2018-12-29
Add unit tests for TomcatMetricsBinder and prevent potential NPE

OleksiiBondar:
Add unit tests for TomcatMetricsBinder and prevent potential NPE when method `findContext` returns null.


wilkinsona:
IIRC, I considered the possibility of an NPE when I wrote this code and concluded that we shouldn't be in a situation where there's no `Context`. Perhaps checking for `null` and throwing an `IllegalStateException` would be better, but I'm not sure that we should limp along.

If you have seen a situation where there was no `Context` can you please provide some more details? Knowing the details would also allow us to test the scenario without mocking everything.

wilkinsona:
The `Context` can never be `null` as it's enforced by `TomcatWebServer`:

https://github.com/spring-projects/spring-boot/blob/48fcaca15524e4e46805580ce3a28ada2186cfe0/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java#L130-L137

I am still not in favour of silently tolerating a `null` `Context` given that it should never happen. I think it's better to fail hard and as soon as possible when an assumption does not hold true. Thanks anyway for the PR.

PR #15587, opened 2018-12-30
Add unit tests for JettyServerThreadPoolMetricsBinder

OleksiiBondar:
Add unit tests for JettyServerThreadPoolMetricsBinder


wilkinsona:
Thanks for the PR, but due to the nature of the code under test and the extensive use of mocks, I don't think the tests do much to verify the implementation and reduce the risk that it's wrong. I'm not opposed to having tests for this, but given that the code under test doesn't contain much logic and is largely just caller getters and casting, I think the tests need to use real Jetty types rather than mocks to be worthwhile.

OleksiiBondar:
@wilkinsona As a main contributor to this repo, that's your call to take whether having tests against mocks is not any better than absolute absense of any tests. Feel free to decline/close this PR than.

snicoll:
@OleksiiBondar for the record I agree with Andy on this one. Thanks for the PR, in any case.

Issue #15455, opened 2018-12-13
Spring Boot Maven Plugin 'agent' parameter does not explain how to use multiple agents


kriegaex:
From the plugin documentation, it is unclear how to actually add multiple Java agents to the command line, even though `agent` permits an array of file names, see also my [answer](https://stackoverflow.com/a/53754498/1082681) to a corresponding StackOverflow question for more details. The strangest part is that `agent` is a singular word and not the expected plural `agents`, which would imply that inside you could list agents using `agent`. But in this case the singular is the list element. I found out that I can just invent any subtag name in order to list multiple agents, but this is neither documented nor intuitive. I know it would be a breaking change to rename the parameter to its plural form, but maybe you could add an alias and then deprecate the singular name and also explain how to use multiple agents in the plugin documentation and/or in an example. The way it is now just keeps people guessing. I also did not find any integration or unit test case for it in the code base.

philwebb:
Thanks for raising the issue. I don't think the naming is intentional and the plural would seem to make sense to me.

kriegaex:
This is nice, thanks for the swift reaction. In the commit I do not see much update with regard to comprehensive documentation and/or a usage example, though. Has this been done somewhere else or have you decided not to do it?

snicoll:
Your feedback on the documentation is more related to a general Maven usage than anything specific to this plugin . The attributes are [already documented](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/maven-plugin/run-mojo.html) and I don't think this deserves a dedicated example.

PR #15588, opened 2018-12-30
Spring Boot Maven Plugin 'agent' parameter does not explain how to use multiple agents #15455

nishantraut:


nishantraut:
Applied spring-javaformat:apply but still build is failing

snicoll:
@nishantraut thanks but that's not what we had in mind for this. See https://github.com/spring-projects/spring-boot/commit/51936e1a5cd032c9e622c30ac18ec9b38d5a294e

PR #15589, opened 2018-12-30
Avoid string concatenation inside StringBuilder append()

igor-suhorukov:

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/jmx/annotation/DiscoveredJmxOperation.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/jmx/annotation/DiscoveredJmxOperation.java
@@ -180,9 +180,9 @@ class DiscoveredJmxOperation extends AbstractDiscoveredOperation implements JmxO
 		public String toString() {
 			StringBuilder result = new StringBuilder(this.name);
 			if (this.description != null) {
-				result.append(" (" + this.description + ")");
+				result.append(" (").append(this.description).append(")");
 			}
-			result.append(":" + this.type);
+			result.append(":").append(this.type);
 			return result.toString();
 		}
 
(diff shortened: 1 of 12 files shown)


snicoll:
@igor-suhorukov thanks but I am going to hold on this one (see https://github.com/spring-projects/spring-boot/pull/14254#issuecomment-417435441).

dreis2211:
@snicoll If I may provide a test benchmark:

```
    @Benchmark
    public String testNew(TestState testState) {
		StringBuilder stringBuilder = new StringBuilder();
		stringBuilder.append("This is a test (");
		stringBuilder.append(testState);
		stringBuilder.append(")");
		return stringBuilder.toString();
    }

    @Benchmark
    public String testOld(TestState testState) {
		StringBuilder stringBuilder = new StringBuilder();
		stringBuilder.append("This is a test (" + testState + ")");
		return stringBuilder.toString();
    }
```
```
Benchmark             Mode  Cnt        Score       Error  Units
MyBenchmark.testNew  thrpt   10  3376165,616 ± 31221,896  ops/s
MyBenchmark.testOld  thrpt   10  2555779,678 ± 59353,058  ops/s
```

The comment in #14254 is about changing concatenations with + to a StringBuilder. Here we already have a StringBuilder available, so it should be quite natural to use it. In fact, IDEA is producing a warning about such places :
grafik

Cheers,
Christoph

snicoll:
@dreis2211 thanks for the feedback. I know about the potential speed difference and I am using IntelliJ IDEA myself so I know about the warning as well. We've declined somewhat related changes in the past and I'd like to double check with the team before merging.

snicoll:
We've had a look to this one during our last team call and decided to go for it. Thanks for the PR!

wilkinsona:
> I wonder is there a reason why these kinds of changes were denied in the past

Yes. We've denied them in the past when, in our opinion, they make the code harder to read and when the changes are not in a performance-criticial code path.

PR #15590, opened 2018-12-30
Allow using jacoco with Spring Boot modules, inherited from "spring-boot-parent"

stepio:
Hello,

I have a "[pet project](https://github.com/stepio/spring-actuator-kafka)", which inherits `spring-boot-parent` module, mostly to get your great dependency management and follow most of the building approach.

Recently I decided to configure a couple of code analyzers for it, including [SonarCloud](https://sonarcloud.io/dashboard?id=stepio_spring-actuator-kafka) check.

Everything was fine there except of test code coverage - although I had a fair number of tests, I could not make [jacoco](https://www.jacoco.org) work properly - file `jacoco.exec` simply was not generated. After studying the relevant documentation, stackoverflow and experimenting with my other projects I found that root cause was usage of `spring-boot-parent`.

The problem is that `maven-surefire-plugin` overrides value of `argLine` property, so jacoco cannot pass location of agent for surefire. That's a well-known issue and possible solutions are explained [in their docs](https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html).

I'm not sure if you aware of this issue or care about this particular case, but I decided to share my finding with you just to save your time if you encounter the same case.

P.S.: For some reason I cannot build latest `spring-boot` on my mac - same error before my changes and after them, so hope I did not mess it up.

### DIFF
--- a/spring-boot-project/spring-boot-parent/pom.xml
+++ b/spring-boot-project/spring-boot-parent/pom.xml
@@ -23,6 +23,7 @@
 		<java.version>1.8</java.version>
 		<project.build.sourceEncoding>UTF-8</project.build.sourceEncoding>
 		<project.reporting.outputEncoding>UTF-8</project.reporting.outputEncoding>
+		<argLine>-Xmx1024m</argLine>
 		<maven.version>3.5.4</maven.version>
 		<maven-resolver.version>1.1.1</maven-resolver.version>
 		<spock.version>1.0-groovy-2.4</spock.version>
@@ -631,7 +632,6 @@
 						<java.security.egd>file:/dev/./urandom</java.security.egd>
 						<java.awt.headless>true</java.awt.headless>
 					</systemPropertyVariables>
-					<argLine>-Xmx1024m</argLine>
 					<trimStackTrace>false</trimStackTrace>
 					<redirectTestOutputToFile>true</redirectTestOutputToFile>
 					<runOrder>alphabetical</runOrder>


snicoll:
> The problem is that maven-surefire-plugin overrides value of argLine property, so jacoco cannot pass location of agent for surefire. That's a well-known issue and possible solutions are explained in their docs.

I am not sure I understand what you mean by that or how your PR is fixing anything. You could override the `argLine` attribute in your project. If you cannot, can you point us to a sample or something that explains why you can't?

stepio:
I do can redefine the whole `maven-surefire-plugin` section and my own issue is already fixed.

The issue is reproduced when next maven command is executed:
`clean org.jacoco:jacoco-maven-plugin:prepare-agent test`
I created two branches with minimal changes in the same project.

When I execute the above given command, I get `jacoco.exec` generated for [spring-actuator-kafka/jacoco-success](https://github.com/stepio/spring-actuator-kafka/tree/feature/jacoco-success), but not for [spring-actuator-kafka/jacoco-fail](https://github.com/stepio/spring-actuator-kafka/tree/feature/jacoco-fail).
Here is the difference between them (same as my PR):
https://github.com/stepio/spring-actuator-kafka/pull/10/files

snicoll:
Thanks for the sample. I don't think that we should change anything considering there is an explicit documentation [that covers this arrangement](https://www.eclemma.org/jacoco/trunk/doc/prepare-agent-mojo.html).

I've changed your failed project to use ```<argLine>@{argLine} -Xmx1024m</argLine>``` and that works just fine. The default in the parent really is meant to provide a sensible out-of-the-box experience. There are many use cases where the plugin needs to be customized (like in your use case) and that's perfectly fine. I am not keen to introduce a non-scoped property considering this case is already taken into account in the documentation.

stepio:
Your parent pom causes hidden unexpected issue, which is do covered in jacoco docs, but I have not thought about it for a long time as I have not defined `maven-surefire-plugin` this way myself.

Fixing this could save some time for future maintainers. But... as you wish :)

Issue #15420, opened 2018-12-07
Add an outcome tag to RestTemplate/WebClient metrics similar to that provided for WebMVC and WebFlux


michaelmcfadyen:
Related to #13801

_This effects spring boot actuator and specifically it's metrics classes._

**Situation**
I want to graph a count of calls made to a downstream api that returned a 4xx status code.

It would be useful to have a tag on those metrics that would group the response codes appropriately. For example these groupings could be:
- 1xx -> informational
- 2xx -> success
- 3xx -> redirection
- 4xx -> client_error
- 5xx -> server_error

This tag has already been added to the web mvc metrics.

wilkinsona:
Sounds sensible to me, thanks for the suggestion. In the meantime, you can implement this yourself by configuring a custom `RestTemplateExchangeTagsProvider` bean.

snicoll:
We should make sure this gets added to `WebClient` as well for consistency.

nishantraut:
Updated PR for WebClient as well

snicoll:
Closing in favour of PR #15594

PR #15593, opened 2018-12-31
Add an outcome tag to RestTemplate metrics similar to that provided for WebMVC and WebFlux #15420

nishantraut:


nishantraut:
Can someone from spring boot team help me to understand why build is failing, i modified files ran mvn spring-javaformat:apply and then mvn clean verify and commited these files using tortoise git.

PR #15595, opened 2019-01-01
Use JSONObject.putOpt in JsonConverter

dreis2211:
Hi,

this PR removes a method from JsonConverter in favor of using `JSONObject.putOpt()` that does (almost) exactly the same.

Happy new year to the complete team!
Cheers,
Christoph


snicoll:
That didn't take long, thanks again Christoph. Hppy new year!

PR #15602, opened 2019-01-03
Introduce TransactionTemplateBuilder

ttddyy:
When programmatically controls transactional operations(especially with different settings such as readOnly or different propagation/isolation), it is helpful to have easier way of composing `TransactionTemplate`. 

Currently, spring/spring-boot doesn't offer builder style construction for `TransactionTemplate`.
Since `TransactionTemplate` itself is a `TransactionDefinition`, different "readOnly/propagation/isolation" needs different `TransactionTemplate` instances.

In this PR, I created a `TransactionTemplateBuilder` similar to `RestTemplateBuilder`.

Another way might be a different variant of `TransactionTemplate` that provides a method to allow passing `TransactionDefinition` at execution time.

```
public class SimpleTransactionTemplate {
   // with default
   public <T> T execute(TransactionCallback<T> action) throws TransactionException {
     ...
   }
   // with different definition
   public <T> T execute(TransactionDefinition definition, TransactionCallback<T> action) throws TransactionException {
     ...
   }
}
```

Thanks,

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/transaction/TransactionTemplateBuilder.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/transaction/TransactionTemplateBuilder.java
@@ -1,0 +1,298 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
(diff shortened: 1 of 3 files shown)


snicoll:
@ttddyy we've discussed this proposal on the team call and we don't see this use case compelling enough to merge that feature in Spring Boot itself. Thanks for the super complete PR, in any case!

PR #15637, opened 2019-01-07
Provide an option to use Neo4j-OGM native types.

michael-simons:
In the upcoming next release of Neo4j-OGM we provide means to use Neo4j native types (Spatial, Time ec.) in Neo4j-OGM. This native types allow the user to use `java.time.*`and others in entities without OGM converting them to Strings or longs.

These types are available for Bolt and Embedded modes and a feature requested by our users.

This PR contains the following:
* Added the modules containing the native type system for Bolt and Embedded Neo4j-OGM to the dependency management
* A new configuration property to enable them (They are an opt-in feature)
* A failure analyser when they are enabled but the dependencies aren't on the class path

The dependency of OGM has been set to a released alpha version to be able to discuss this PR.

Neo4j-OGM 3.2 will be the version Spring Data Moore (5.2) requires. It is however not compatible with Spring Data Lovelace (5.1) and as such, this PR should not make it into a Spring Boot version that uses Spring Data Lovelace or before.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicatorTests.java
+++ b/spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicatorTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -42,6 +42,7 @@ import static org.mockito.Mockito.mock;
  *
  * @author Eric Spiegelberg
  * @author Stephane Nicoll
+ * @author Michael Simons
  */
 public class Neo4jHealthIndicatorTests {
 
@@ -77,9 +78,8 @@ public class Neo4jHealthIndicatorTests {
 
 	@Test
 	public void neo4jDown() {
-		CypherException cypherException = new CypherException("Error executing Cypher",
-				"Neo.ClientError.Statement.SyntaxError",
-				"Unable to execute invalid Cypher");
+		CypherException cypherException = new CypherException(
+				"Neo.ClientError.Statement.SyntaxError", "Error executing Cypher");
 		given(this.session.query(Neo4jHealthIndicator.CYPHER, Collections.emptyMap()))
 				.willThrow(cypherException);
 		Health health = this.neo4jHealthIndicator.health();
--- a/spring-boot-project/spring-boot-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-autoconfigure/pom.xml
@@ -782,12 +782,17 @@
 		</dependency>
 		<dependency>
 			<groupId>org.neo4j</groupId>
-			<artifactId>neo4j-ogm-http-driver</artifactId>
+			<artifactId>neo4j-ogm-bolt-native-types</artifactId>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
 			<groupId>org.neo4j</groupId>
 			<artifactId>neo4j-ogm-embedded-driver</artifactId>
+			<scope>test</scope>
+		</dependency>
+		<dependency>
+			<groupId>org.neo4j</groupId>
+			<artifactId>neo4j-ogm-http-driver</artifactId>
 			<scope>test</scope>
 		</dependency>
 		<dependency>
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -33,10 +33,13 @@ import org.springframework.util.ClassUtils;
  * @author Michael Hunger
  * @author Vince Bickers
  * @author Aurélien Leboulanger
+ * @author Michael Simons
  * @since 1.4.0
  */
-@ConfigurationProperties(prefix = "spring.data.neo4j")
+@ConfigurationProperties(prefix = Neo4jProperties.CONFIGURATION_PREFIX)
 public class Neo4jProperties implements ApplicationContextAware {
+
+	static final String CONFIGURATION_PREFIX = "spring.data.neo4j";
 
 	static final String EMBEDDED_DRIVER = "org.neo4j.ogm.drivers.embedded.driver.EmbeddedDriver";
 
@@ -71,6 +74,12 @@ public class Neo4jProperties implements ApplicationContextAware {
 	 * entire processing of the request.",
 	 */
 	private Boolean openInView;
+
+	/**
+	 * Disables the conversion of java.time.* and spatial types in Neo4j-OGM entities and
+	 * uses Neo4j native types wherever possible.
+	 */
+	private boolean useNativeTypes = false;
 
 	private final Embedded embedded = new Embedded();
 
@@ -116,6 +125,14 @@ public class Neo4jProperties implements ApplicationContextAware {
 		this.openInView = openInView;
 	}
 
+	public boolean isUseNativeTypes() {
+		return this.useNativeTypes;
+	}
+
+	public void setUseNativeTypes(boolean useNativeTypes) {
+		this.useNativeTypes = useNativeTypes;
+	}
+
 	public Embedded getEmbedded() {
 		return this.embedded;
 	}
@@ -146,6 +163,9 @@ public class Neo4jProperties implements ApplicationContextAware {
 			builder.credentials(this.username, this.password);
 		}
 		builder.autoIndex(this.getAutoIndex().getName());
+		if (this.useNativeTypes) {
+			builder.useNativeTypes();
+		}
 	}
 
 	private void configureUriWithDefaults(Builder builder) {
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -138,7 +138,7 @@
 		<mssql-jdbc.version>6.4.0.jre8</mssql-jdbc.version>
 		<mysql.version>8.0.15</mysql.version>
 		<nekohtml.version>1.9.22</nekohtml.version>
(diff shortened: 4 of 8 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jProperties.java
```diff
@@ -72,6 +73,16 @@
 	 */
 	private Boolean openInView;
 
+	/**
+	 * Enables the direct usage of Neo4j native types in domain classes without applying
+	 * any converter. This includes all native date, time and duration of Neo4j 3.4+
+	 * (those are mapped to {@code java.time} types as well as the Neo4j spatial datatypes
```
snicoll: Configuration key description cannot use Javadoc tags ,see [the documentation](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#configuration-metadata-annotation-processor).

Also, this documentation is quite long and unusual (compare this text with any other property in the codebase right now). Please consider rephrasing it to be more succinct.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jProperties.java
```diff
@@ -136,16 +155,23 @@ public Configuration createConfiguration() {
 	}
 
 	private void configure(Builder builder) {
+
```
snicoll: those additional lines aren't necessary

spring-boot-project/spring-boot-dependencies/pom.xml
```diff
@@ -138,7 +138,7 @@
 		<mssql-jdbc.version>6.4.0.jre8</mssql-jdbc.version>
 		<mysql.version>8.0.13</mysql.version>
 		<nekohtml.version>1.9.22</nekohtml.version>
-		<neo4j-ogm.version>3.1.5</neo4j-ogm.version>
+		<neo4j-ogm.version>3.2.0-alpha02</neo4j-ogm.version>
```
snicoll: We cannot really move to an `alpha` version at this point. Can you clarify how this is scoped with the next Spring Data release train?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jNativeTypesFailureAnalyzer.java
```diff
@@ -0,0 +1,59 @@
...
+import org.springframework.boot.diagnostics.AbstractFailureAnalyzer;
+import org.springframework.boot.diagnostics.FailureAnalysis;
+
+/**
+ * A {@link org.springframework.boot.diagnostics.FailureAnalyzer} that performs analysis
+ * of a Neo4j-OGM configuration failure caused by a missing native types module.
+ *
+ * @author Michael Simons
+ */
+class Neo4jNativeTypesFailureAnalyzer
```
snicoll: I am not entirely convinced that's worth adding a dedicated `FailureAnalyzer`. You should be able to phrase that in the exception message itself, can't you?

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4377,6 +4377,29 @@ in your configuration, e.g. `spring.data.neo4j.uri=file://var/tmp/graph.db`.
 
 
 
+[[boot-features-neo4j-ogm-native-types]]
+==== Using native types
+Neo4j-OGM can map some types, like `java.time.*` to `String` based properties or
+use one of the various native types that Neo4j provide
+(Please refer to the https://neo4j.com/docs/developer-manual/3.4/drivers/cypher-values/#driver-neo4j-type-system[Cypher type system] for a full list).
```
snicoll: There is a hard-coded version there and we'd prefer not to have to chase those links down. There are other versions used in the doc already, see `index.adoc` for examples.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4377,6 +4377,29 @@ in your configuration, e.g. `spring.data.neo4j.uri=file://var/tmp/graph.db`.
...
+----
+	spring.data.neo4j.use-native-types=true
+----
+
+With this setting you can use Java 8 temporal types without any conversion necessary in your domain classes.
+
+[NOTE]
+====
+This feature is only available if Neo4j-OGM uses Bolt or embedded transport mode.
+====
```
snicoll: I feel that part of the description here would be better suited in the Neo4j documentation. We could have a link that explains this feature rather than explaining it here again.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicatorTests.java
```diff
@@ -77,9 +78,8 @@ public void neo4jUp() {
 
 	@Test
 	public void neo4jDown() {
-		CypherException cypherException = new CypherException("Error executing Cypher",
-				"Neo.ClientError.Statement.SyntaxError",
-				"Unable to execute invalid Cypher");
+		CypherException cypherException = new CypherException(
```
snicoll: I welcome such polish but I'd prefer to see them in a separate commit.

spring-boot-project/spring-boot-autoconfigure/pom.xml
```diff
@@ -790,6 +790,16 @@
 			<artifactId>neo4j-ogm-embedded-driver</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
snicoll: dependencies are added in natural order.

spring-boot-project/spring-boot-autoconfigure/pom.xml
```diff
@@ -790,6 +790,16 @@
...
+			<groupId>org.neo4j</groupId>
+			<artifactId>neo4j-ogm-bolt-driver</artifactId>
```
snicoll: Why do you need this explicitly?

spring-boot-project/spring-boot-dependencies/pom.xml
michael-simons: The final version of 3.2 is due to be released with / for Spring Data Moore. It has several new features (this here being one of it) but more important, we removed stuff we deprecated in 3.1.5 and 3.1.6. 

The reason for us using it in this PR is that we wanted to get at least the proposed changes into a discussion.

Neo4j releases those alphas into central btw, so no need to add additional repos (if this is good or not is another question).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jNativeTypesFailureAnalyzer.java
michael-simons: I could but than people would have to find it in the stack trace. I like the value provided by a `FailureAnalyzer` for providing clear instructions which can be seen much better than in a stack trace.

spring-boot-project/spring-boot-dependencies/pom.xml
snicoll: Thanks and good to hear the release will be aligned with the `Moore` release train.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jNativeTypesFailureAnalyzer.java
snicoll: I am not denying any of that. But I don't see why those two particular case should have a dedicated `FailureAnalyzer`. Have you considered using `InvalidConfigurationPropertyValueException`?

michael-simons: Oh nice, didn't know that. I'll have a look.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
michael-simons: I'm gonna change that including the hardcoded version number. There's a chapter on that on the 3.2 docs, it's just not uploaded to our docs server. Good point.

spring-boot-project/spring-boot-autoconfigure/pom.xml
michael-simons: I don't understand that one. What is the natural order of the dependencies?

snicoll: alphabetical order (`groupId` then `artifactId`).

michael-simons:
I followed your suggestion and replaced the analyzer by some logic throwing an `InvalidConfigurationPropertyValueException`. I looked through the mechanism handling those and I think it's appropriate for that use case here. Thanks for the hint.

One of the test will fail now as I discovered a bug in our OGM prerelease. So again, thanks for the feedback and ideas.

All other suggestions applied.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicatorTests.java
snicoll: Was this resolved really? I can still see the change

snicoll: There are seven commits right now so it's hard for me to figure out what is supposed to be squashed and what isn't.

Generally speaking, if you have something that's good to go, I very much prefer a quick PR that I can merge right away. If you don't have time to create one, that's fine as well but please squash the current commits to make that a bit more explicit. Thanks!

snicoll:
@michael-simons can you please squash what needs to be and give us an update about this one?

michael-simons:
I have extracted the version upgrade into #15937 and squashed the remaining commits here.

Thanks for your feedback and time on this.

wilkinsona:
Thanks very much for the PR, @michael-simons. I've merged the proposed changes into master along with some [polish](https://github.com/spring-projects/spring-boot/commit/18174a8e9ef0a01d91b0c05a3e7a38c05d35ef13). The biggest change there was to remove the exception translation. I think there may be some room for improvement in the exception message that could be done in OGM. If we then still want to do something in Boot, I think it would be better to add some dedicated failure analysers. As @snicoll pointed out to me, that would have the advantage of analysing the failure even when the auto-configuration has backed off.

PR #15653, opened 2019-01-08
Remove need for zt-zip dependency

dreis2211:
Hi,

while looking through the code I noticed that we have a test-dependency on `zt-zip` that we could get rid of by relatively little code.
There are no strong reasons though to remove the dependency to be entirely honest. It might be just the security issue that was still in my brain and forced me into the PR (I think it was a directory traversal fixed in 1.13). Feel free to close this PR in case you want to keep the dependency.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-parent/pom.xml
+++ b/spring-boot-project/spring-boot-parent/pom.xml
@@ -242,11 +242,6 @@
 				<artifactId>spock-spring</artifactId>
 				<version>${spock.version}</version>
 			</dependency>
-			<dependency>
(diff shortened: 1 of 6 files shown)


wilkinsona:
Thanks for this, @dreis2211. You've added some methods to public API so that the replacement code can work as required. While I'm all for minimising dependencies, this is, as you've noticed, only a test dependency. On balance, I don't think the benefit of dropping the dependency manages to outweigh the cost of the additional methods. Thanks anyway.

dreis2211:
Fair enough. Thanks for the update @wilkinsona

PR #15655, opened 2019-01-09
DecodingException should be wrapped into HTTP 400 as per its doc

yuranos:
As stated in DecodingException javadoc:

> For example in server web application, a DecodingException would
> translate to a response with a 400 (bad input) status while
>  CodecException would translate to 500 (server error) status.

Which was not the case for reactive. As a result, all Jackson exceptions, which are not covered by @ResponseStatus handlers, were mapped to HTTP 500. It's definitely undesirable for most services, because, apart from being confusing, it can draw a misleading picture for analytics, etc., with regard to your service's health/behaviour.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/error/DefaultErrorAttributes.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/error/DefaultErrorAttributes.java
@@ -23,6 +23,7 @@ import java.util.LinkedHashMap;
 import java.util.Map;
 
 import org.springframework.core.annotation.AnnotatedElementUtils;
+import org.springframework.core.codec.DecodingException;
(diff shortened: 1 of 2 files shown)


yuranos:
@bclozel , so, what do you think?

bclozel:
Hi @yuranos 

Thanks for reaching out and your contribution.

I couldn't reproduce this behavior with WebFlux annotations, since `DecodingException` is automatically wrapped in a `ServerWebInputException` and this one is handled as an HTTP 400 exception by the error handling infrastructure.

I've tested this with the functional variant and reproduced this problem. I don't think we should consider the `DecodingException` in the error handling parts of WebFlux, as it's a low-level contract and we've got dedicated types for that. I've created spring-projects/spring-framework#22290 to fix that at the Framework level, superseding this PR.

Thanks!

PR #15666, opened 2019-01-10
Remove an unused interface in RestarterTests

izeye:
This PR removes an unused interface in `RestarterTests`.


izeye:
@snicoll Thanks for merging and happy new year, too 😄

PR #15708, opened 2019-01-15
Auto-configure BatchMessagingMessageConverter with Listener.Type.Batch

leochans:
when use batch type listener, simply declare a bean of RecordMessageConverter will not effect, should warp the RecordMessageConverter to a BatchMessagingMessageConverter when use batch type listener

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
@@ -125,7 +125,8 @@ public class ConcurrentKafkaListenerContainerFactoryConfigurer {
 		map.from(this.replyTemplate).to(factory::setReplyTemplate);
 		map.from(properties::getType).whenEqualTo(Listener.Type.BATCH).toCall(() -> {
 			factory.setBatchListener(true);
-			factory.setMessageConverter(new BatchMessagingMessageConverter(this.messageConverter));
+			factory.setMessageConverter(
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@leochans Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15708)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15708) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@leochans Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15708)!

garyrussell:
@snicoll Functionally LGTM but there are code formatting violations.

It also needs a couple of tests - one with, and one without a configured `RecordMessageConverter` `@Bean` - it can be `null`, in which case the listener receives a `List<Foo>` where Foo is the unconverted `ConsumerRecord.value()` (`List<String>` with default Boot configuration); with a converter it gets `List<Bar>` where the `Bar`s are converted from `ConsumerRecord.value()`.

That said, I think a better solution would be to auto-configure a `BatchMessageConverter` (alongside the `RecordMessageConverter`) for use with batch listeners.

That would give the user more control over the converter, rather than hard-wiring a `BatchMessagingMessageConverter` implementation that wraps the configured `RecordMessageConverter`.

garyrussell:
I understand that (I wrote the code); it can be null without problems, but test cases are needed to ensure the configuration works as expected.

However, I think my suggestion to auto-configure a `BatchMessageConverter` is a better solution (but it could fall back to your code if none is provided, but a `RecordConverter` is).

snicoll:
Thanks for you for the PR @leochans but I've decided to take a different approach based on @garyrussell feedback, see #15942 for more details.

PR #15709, opened 2019-01-15
Unlike Jetty and Tomcat, Reactor Netty and Undertow do not cause a startup failure when SSL is enabled but no key store is configured

Raheela1024:
…artup failure when SSL is enabled but no key store is configured issue fixed.


wilkinsona:
Thanks very much for the pull request, @Raheela1024. It's great to see you making another contribution after taking care of a first-timers only issue for us. I did [a little bit of polishing](
https://github.com/spring-projects/spring-boot/commit/a172e146de8ce83f835eb3bde83161e7d8c71c63) that you might be interested in. I tweaked some of the exceptions that are thrown to make things even more consistent across containers and also added a few tests to hopefully verify that the behaviour is consistent.

Issue #15705, opened 2019-01-14
Rename ElasticSearchJestHealthIndicatorAutoConfiguration


philwebb:
It should be `ElasticsearchJestHealthIndicatorAutoConfiguration`

snicoll:
Closing in favour of PR #15723

snicoll:
While reviewing the related PR, we've decided not to pursue this, see https://github.com/spring-projects/spring-boot/pull/15723#issuecomment-464116701 for more details

Issue #14881, opened 2018-10-17
Move JerseyServerMetricsAutoConfiguration to web package


mbhave:
It looks like it should live in `org.springframework.boot.actuate.autoconfigure.metrics.web`.

mbhave:
Placing on hold till #14860 is sorted.

nishantraut:
Hi Madhura, This looks simple moving class from old package to new mentioned above. Please let me know once this issue is removed from hold state. Will be happy to work on this. Thanks.

wilkinsona:
Flagging for team attention so that we can consider doing this in 3.0 without #14860.

wilkinsona:
This is blocked until we've [reinstated support for Jersey](https://github.com/spring-projects/spring-boot/issues/28637) which should be possible once 3.1 has been released.

The main Jersey auto-configuration is in `org.springframework.boot.autoconfigure.jersey` so the current location of `JerseyServerMetricsAutoConfiguration` – `org.springframework.boot.actuate.autoconfigure.metrics.jersey` – feels right to me. If we moved the metrics auto-configuration to `org.springframework.boot.actuate.autoconfigure.metrics.web` then I wonder if `JerseyAutoConfiguration` should move too.

wilkinsona:
Thanks, @mbhave.

`org.springframework.boot.actuate.autoconfigure.web.servlet` is named after `org.springframework.web.servlet` which is the root package for Spring MVC. As such, I don't think Jersey-related code belongs in a `web.servlet` package.

`org.springframework.boot.actuate.autoconfigure.web.jersey` is the auto-configuration for the actuator's web endpoint infrastructure when running on top of Jersey. There are `.servlet` and `.reactive` packages here too for the MVC and WebFlux support. We also have `org.springframework.boot.actuate.autoconfigure.endpoint.web.jersey` (and `.servlet` and `.reactive` variants as well). We only have one class in each of these packages. The distinction between `.endpoint.web.<framework>` and `.web.<framework>` isn't very clear to me. It's all code for running Actuator on top of a particular framework. I wonder if they could be combined.

The location of `JerseyApplicationPath` and `DefaultJerseyApplicationPath` does seem inconsistent. I think they should both be in the same package as the rest of the Jersey auto-configuration. That's `org.springframework.boot.autoconfigure.jersey` at the moment.

It looks to me like that are two, somewhat separate packaging questions here:

[...]

Issue #14860, opened 2018-10-17
Provide support for deprecated auto-configuration classes


philwebb:
#14412 shows that we need a way to offer deprecated auto-configuration classes. Such classes would not need to actually contain configuration, but would still be valid when used as an `exclude` or and `@AutoConfigureBefore`/`@AutoConfigureAfter` value.

philwebb:
Prototype work https://github.com/philwebb/spring-boot/tree/gh-14860

mbhave:
This one looks similar: #6684

wilkinsona:
We have no need for this at this time. Closing for now.

PR #15723, opened 2019-01-16
Rename ElasticSearchJestHealthIndicatorAutoConfiguration

ayudovin:
Rename `ElasticSearchJestHealthIndicatorAutoConfiguration` to `ElasticsearchJestHealthIndicatorAutoConfiguration`
this [task](https://github.com/spring-projects/spring-boot/issues/15705)


mbhave:
#14881 is a similar issue which is currently on hold for the deprecated auto-config support which will enable it to be implemented in a non-breaking manner. It seems inconsistent to do this before #14860 given that moving a class is less of a breaking change than renaming it since the IDE will be able to find it in the former case.

ayudovin:
@mbhave, do you want that I add changes for this [#14881](https://github.com/spring-projects/spring-boot/issues/14881) in this pull request?

ayudovin:
@mbhave, does it mean that this pull request should be on hold until #14860 is done?

mbhave:
@ayudovin Sorry, I should've been a bit more clear about what I meant. The comment was for the team to consider that there could be such inconsistencies before merging the PR. The milestone would be decided based on what the team decides.

ayudovin:
@mbhave, ok, no problem

snicoll:
Thank you for the PR but there are actually 3 auto-configuration classes with the same problem and we don't believe adding a breaking change in 2.2 for this is worthwhile. We can reconsider when #14860  is implemented and we can group all three in a single auto-configuration. Right now they are split to benefit from the optimization of skipping the auto-configuration if the client library is not on the classpath.

Issue #15712, opened 2019-01-15
HttpMessageConvertersAutoConfiguration should be conditional on not being a reactive web application


wilkinsona:
Http message converters shouldn't be used in a reactive web application, but may be used in a servlet-based web application or a non-web application. We should update the conditions on HttpMessageConvertersAutoConfiguration accordingly to avoid their auto-configuration when running a reactive web application.

dsyer:
They are used in `RestTemplate` though, right, and there is autoconfiguration for that (probably needs to be picked apart and offered separately to `WebClient`)?

wilkinsona:
We talked about that, and decided that no one should be using `RestTemplate` in a reactive web application as it's a blocking API. We probably need an issue similar to this one to disable `RestTemplate` auto-configuration in the same situation.

daslan31:
What if I'm using FeignClient for invoking api's outside of the app? FeignClientsConfiguration requires these converters, in order to workaround this I'm creating these beans by myself

wilkinsona:
You should use https://github.com/Playtika/feign-reactive or similar in a reactive application.

PR #15724, opened 2019-01-16
HttpMessageConvertersAutoConfiguration should be conditional on not being a reactive web application

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/15712)


wilkinsona:
Thanks very much for the PR. Unfortunately, this doesn't quite take the right approach. We don't want to add a new condition to the public API and the condition can be implemented as a negation of a single conditional on reactive web application. Thanks anyway and sorry for the wasted effort. If you'd like to avoid that in the future, please feel free to comment on an issue to let us know you'd like to work on it. We can then make sure that we're all on the same page as to how it should be implemented.

ayudovin:
@wilkinsona, ok, no problem

PR #15594, opened 2018-12-31
Add an outcome tag to RestTemplate metrics similar to that provided for WebMVC and WebFlux

nishantraut:
See #15420

PR #15725, opened 2019-01-16
Update 'HTTP Client Metrics' section of docs to include "outcome" tag description

michaelmcfadyen:
Update `HTTP Client Metrics` section of the documentation to contain information on the `outcome` tag introduced as part of #15594. Also, update the layout of the tags section to match other similar sections


michaelmcfadyen:
@bclozel noticed that we missed the docs as part of #15594. Do you mind taking a look over this?

Issue #15722, opened 2019-01-16
Disable background preinitializer when only one processor is available


wilkinsona:
When a single processor is available, `BackgroundPreinitializer` can actually slow things down as it increases contention for the single CPU during startup. We should disable it when `Runtime.availableProcessors()` is 1.

PR #15726, opened 2019-01-17
Disable background preinitializer when only one processor is available

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/15722)


wilkinsona:
Thanks for the PR, but the changes proposed here do not implement the required logic. The changes here will cause background pre-initialization to always be performed when there's a single processor, irrespective of the event that's received or the system property's value. Thanks anyway. If you are interested, I have implemented the desired changes in 
https://github.com/spring-projects/spring-boot/commit/7dd65074e872c5541a303f75862917d9e4377abf.

ayudovin:
@wilkinsona, ok, I have got it

PR #15734, opened 2019-01-17
(update) upgrade commons-fileupload: 1.3.3 -> 1.4

chtompki:
From @chtompki on Apache Commons...Announced it yesterday here:

http://mail-archives.us.apache.org/mod_mbox/www-announce/201901.mbox/%3c77F27C1B-CBF9-432A-8401-08C0AFE3E7F7@apache.org%3e

All the best,
-Rob


snicoll:
Thanks for the PR but we have a semi automated process to upgrade dependencies so we don't accept such PRs for that reason.

PR #15735, opened 2019-01-18
Configure coordinated JSON provider in json-path. Fixes gh-15727

### LINKED ISSUE #15727: Use of json-path's default JsonSmartJsonProvider introduces asymmetric handling of special characters
By default json-path uses JsonSmartJsonProvider (https://github.com/json-path/JsonPath#jsonprovider-spi) to parse JSON documents. There are subtle differences in behaviour between that parser and Jackson, which is used to deserialise objects when using `org.springframework.boot.test.json.JacksonTester`, in particular an asymmetric treatment of some special characters. Consider the following test:

```
	private JacksonTester<String> stringJson;

	@Test
	public void parseSpecialCharactersTest() throws Exception {
		String stringWithSpecialCharacters = "\u0006\u007F";
		assertThat(this.stringJson.write(stringWithSpecialCharacters))
				.extractingJsonPathStringValue("@")
				.isEqualTo(stringWithSpecialCharacters);
	}
```

The test should pass, but it does not because Jackson is used to serialise the String to JSON, and Json-Smart is used to parse the JSON and evaluate the path expression. The result of evaluating the path expression does not contain the same Unicode code points as the original String:

```
org.opentest4j.AssertionFailedError: 
Expecting:
 <"">
to be equal to:
 <"">
but was not.
```

I'm using Spring Boot 2.1.0.

Suggested fix: json-path can be configured to use a different JSON library to do the parsing, including Jackson and Gson. Spring Boot should make sure that the correct JSON parser is configured in json-path, for example `JacksonJsonProvider` when using `JacksonTester`, so the symmetry is preserved.

dberrueta-atlassian:
This fixes #15727 by passing weaving the json-path configuration from the `JacksonTester` and `GsonTester` classes down to the point where json-path is actually called in `JsonContentAssert.JsonPathValue.getValue`. It also adds an integration test that demonstrates the fix.

I can think of other ways to address this issue. For example, a simpler solution would be to just hardcode json-path to always use Jackson. That would work great if you are using `JacksonTester`, and has a good chance to work fine if you are using `GsonTester`. Please let me know what you think.


pivotal-issuemaster:
@dberrueta-atlassian Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15735)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15735) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dberrueta-atlassian Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15735)!

berrueta:
I missed an integration test. I'll look into that.

berrueta:
I've pushed a simpler solution based on a new an abstract method in `AbstractJsonMarshalTester`.

wilkinsona:
Thanks for this, @berrueta. While I think it's unlikely that anyone will have written their own implementation of `AbstractJsonMarshalTester`, it is public API which makes the additional abstract method a breaking change.

It would be great if we could make this change in a backwards-compatible manner. I wonder if that should be achieve with a slightly modified version of your first attempt by further overloading the `AbstractJsonMarshTester` constructors with one that takes a `Configuration` instance.

snicoll:
@berrueta can you please let us know if you intend to review your proposal? Thank you.

berrueta:
Sorry for the delay and thank you for the feedback. I plan to update the PR in the next days.

berrueta:
I've returned to my first attempt and added some overloads to preserve the interface of `AbstractJsonMarshalTester`, as suggested. Let me know what you think.

wilkinsona:
Thanks again for the PR, @berrueta. I've been looking at testing the GSON side of the change and I believe that, as currently proposed, it causes a regression. The following test fails:

```java
public class GsonTesterIntegrationTests {

	private GsonTester<String> stringJson;

	private Gson gson;

	@Before
	public void setup() {
		this.gson = new Gson();
		GsonTester.initFields(this, this.gson);
	}

	@Test
	public void stringLiteral() throws Exception {
		String stringWithSpecialCharacters = "myString";
		assertThat(this.stringJson.write(stringWithSpecialCharacters))
				.extractingJsonPathStringValue("@")
				.isEqualTo(stringWithSpecialCharacters);
	}

}
```

It fails because when JsonPath reads `"myString"` it returns a `JsonPrimitive` rather than a `String`. I'm not yet sure what to do about that.

Have you observed the asymmetrical behaviour with Gson or has is only been a problem with Jackson? I wonder if we should just limit the fix to Jackson.

dberrueta-atlassian:
Replaced by #16629 .

Issue #15729, opened 2019-01-17
Async configuration ignores `applicationTaskExecutor` when `EnableScheduling` is enabled


snicoll:
If you enable both `@EnableScheduling` and `@EnableAsync` it looks like the async processing is detecting the `taskScheduler` as a candidate for async processing.

This leads to 

```
2019-01-17 07:53:16.174  INFO 48078 --- [   scheduling-1] .s.a.AnnotationAsyncExecutionInterceptor : More than one TaskExecutor bean found within the context, and none is named 'taskExecutor'. Mark one of them as primary or name it 'taskExecutor' (possibly as an alias) in order to use it for async processing: [applicationTaskExecutor, taskScheduler]
```

and the auto-configured `applicationTaskExecutor` is ignored and replaced by the simple `AsyncTaskExecutor`.

snicoll:
Closing in favour of PR #15748

PR #15747, opened 2019-01-21
Flag autoconfigured TaskExecutor as primary

cvienot:
Fixes gh-15729

Flag autoconfigured TaskExecutor as primary to prevent backoff when taskScheduler is also injected in context


snicoll:
@cvienot thank you but flagging the task executor as `@Primary` is not what I had in mind for this. I'd rather not go down that route unless I absolutely have to. 

My current plan is that add an alias to `taskExecutor` and see if that fixes the issue. Thanks for the PR, in any case.

cvienot:
@snicoll ok i undestand, i created another pull request, with the same test. Hope this is what you had in mind

PR #15748, opened 2019-01-21
Async configuration ignores `applicationTaskExecutor` when `EnableScheduling` is enabled 

cvienot:
Use alias for autoconfigured task executor, enabling async configuration to pick it, even though multiple task executors are injected

Fixes gh-15729


snicoll:
@cvienot thanks a lot, this is now merged in `2.1.x` and `master`.

PR #15754, opened 2019-01-22
Polish SessionConfigurationImportSelector.selectImports()

izeye:
This PR polishes `SessionConfigurationImportSelector.selectImports()`.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
@@ -60,7 +60,6 @@ import org.springframework.session.web.http.CookieHttpSessionIdResolver;
 import org.springframework.session.web.http.CookieSerializer;
 import org.springframework.session.web.http.DefaultCookieSerializer;
 import org.springframework.session.web.http.HttpSessionIdResolver;
-import org.springframework.util.StringUtils;
 
 /**
  * {@link EnableAutoConfiguration Auto-configuration} for Spring Session.
@@ -162,13 +161,10 @@ public class SessionAutoConfiguration {
 	abstract static class SessionConfigurationImportSelector implements ImportSelector {
 
 		protected final String[] selectImports(WebApplicationType webApplicationType) {
-			List<String> imports = new ArrayList<>();
-			StoreType[] types = StoreType.values();
-			for (StoreType type : types) {
-				imports.add(SessionStoreMappings.getConfigurationClass(webApplicationType,
-						type));
-			}
-			return StringUtils.toStringArray(imports);
+			return Arrays.stream(StoreType.values())
+					.map((type) -> SessionStoreMappings
+							.getConfigurationClass(webApplicationType, type))
+					.toArray(String[]::new);
 		}
 
 	}


wilkinsona:
Thanks, @izeye.

I think we should be careful with this sort of change. Streams like this are quite a bit slower than a for-loop and we want import selection to be fast. I've been wondering it we should remove much of our use of streams. For example, `org.springframework.boot.autoconfigure.condition.OnBeanCondition.BeanSearchSpec.hasAtLeastOne(List<?>...)` is ~10x faster as a for-loop. It's a tiny amount of time in isolation but it will all add up.

izeye:
@wilkinsona Thanks for the quick feedback! I didn't know streams underperform than for-loops. Feel free to close this if that's the case.

dreis2211:
@wilkinsona As you mentioned it: `BeanSearchSpec.hasAtLeastOne(List<?>...)` will be even faster by directly calling `isEmpty()` on the few collections. By that you can avoid the creation of the iterator (although it might be optimized via escape analysis anyway)

wilkinsona:
Good point, @dreis2211. I'm not sure if the iterator was optimised away, but the difference for the for-loop vs the stream was in the region of 0.001ms vs 0.01ms so it didn't feel worth pursuing (in isolation anyway) for code that's only called 10s or 100s of times in a typical app.

Thinking some more, the same argument could be made to an even greater extent for the change that @izeye is proposing here as it will, I believe, only be called once. We should measure the performance difference, but I'd be very surprised if it is even 1ms.

dreis2211:
```
Benchmark                                              Mode  Cnt         Score         Error   Units
MyBenchmark.testNew                                   thrpt   10   7652559,211 ±  672694,389   ops/s
MyBenchmark.testNew:·gc.alloc.rate                    thrpt   10      2159,600 ±     189,952  MB/sec
MyBenchmark.testNew:·gc.alloc.rate.norm               thrpt   10       296,000 ±       0,001    B/op
MyBenchmark.testNew:·gc.churn.PS_Eden_Space           thrpt   10      2187,497 ±     181,017  MB/sec
MyBenchmark.testNew:·gc.churn.PS_Eden_Space.norm      thrpt   10       300,045 ±      15,037    B/op
MyBenchmark.testNew:·gc.churn.PS_Survivor_Space       thrpt   10         0,203 ±       0,068  MB/sec
MyBenchmark.testNew:·gc.churn.PS_Survivor_Space.norm  thrpt   10         0,028 ±       0,009    B/op
MyBenchmark.testNew:·gc.count                         thrpt   10       133,000                counts
MyBenchmark.testNew:·gc.time                          thrpt   10        88,000                    ms
MyBenchmark.testOld                                   thrpt   10  13130675,742 ± 1095275,364   ops/s
MyBenchmark.testOld:·gc.alloc.rate                    thrpt   10      2007,129 ±     307,253  MB/sec
MyBenchmark.testOld:·gc.alloc.rate.norm               thrpt   10       160,000 ±      12,749    B/op
MyBenchmark.testOld:·gc.churn.PS_Eden_Space           thrpt   10      1997,456 ±     332,586  MB/sec
MyBenchmark.testOld:·gc.churn.PS_Eden_Space.norm      thrpt   10       159,249 ±      17,263    B/op
MyBenchmark.testOld:·gc.churn.PS_Survivor_Space       thrpt   10         0,172 ±       0,103  MB/sec
MyBenchmark.testOld:·gc.churn.PS_Survivor_Space.norm  thrpt   10         0,014 ±       0,008    B/op
MyBenchmark.testOld:·gc.count                         thrpt   10       128,000                counts
MyBenchmark.testOld:·gc.time                          thrpt   10        80,000                    ms
```

wilkinsona:
Thank you, @dreis2211. To avoid any confusion, can you please confirm which code is `testOld` and which is `testNew` in that benchmark?

dreis2211:
Sure thing. **testOld** is the classic loop. **testNew** is the proposed change with streams.

Issue #15566, opened 2018-12-26
Add more MustacheAutoConfiguration compiler options


fjcanyue:
Currently, MustacheAutoConfiguration compiler only support two options, loader and collector. Would you please add more options for convenience, e.g. defaultValue/nullValue.

```
public Mustache.Compiler mustacheCompiler(TemplateLoader mustacheTemplateLoader) {
		return Mustache.compiler().withLoader(mustacheTemplateLoader)
				.withCollector(collector());
	}
```

snicoll:
>  e.g. defaultValue/nullValue.

Thanks for the suggestion. We can improve the auto-configuration but I'd rather do that with concrete use cases rather than based on a couple suggestions. Do you have some concrete use case or some customizations you've applied in your own project you'd like to share?

fjcanyue:
In my own project, if I use default compiler,  an exception will be thrown when template variable value is null. (e.g. one variable is an empty object, {{object.field}} will throw an exception)

So I have to create a customized configuration to avoid that, but not use MustacheEnvironmentCollector.

```
@Configuration
public class MustacheAutoConfiguration
		extends org.springframework.boot.autoconfigure.mustache.MustacheAutoConfiguration {

	public MustacheAutoConfiguration(MustacheProperties mustache, Environment environment,
			ApplicationContext applicationContext) {
		super(mustache, environment, applicationContext);
	}

	@Override
	public Compiler mustacheCompiler(TemplateLoader mustacheTemplateLoader) {
		return Mustache.compiler().defaultValue("").withLoader(mustacheTemplateLoader);
	}
}
```

wilkinsona:
I think it's worth doing something here. Recreating the existing `Compiler` configuration is a little bit involved doe to the `MustacheEnvironmentCollector`. Perhaps it should be a bean? I think there's a case for some configuration properties (`standardsMode`, `defaultValue`, etc), some additional optional beans (`Escaper`, `Formatter`), and perhaps a customizer too. I'm not yet sure which combination of the three would be best.

PR #15756, opened 2019-01-22
Add more mustache auto configuration

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/15566)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mustache/MustacheAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mustache/MustacheAutoConfiguration.java
@@ -27,8 +27,10 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.template.TemplateLocation;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for the PR. As indicated in [this comment](
https://github.com/spring-projects/spring-boot/issues/15566#issuecomment-453065571), we're not exactly sure how to tackle this. We suspect we need some combination of new properties and a customizer so what it proposed here addresses only part of the likely solution.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mustache/MustacheAutoConfiguration.java
```diff
@@ -79,8 +81,31 @@ public void checkTemplateLocationExists() {
...
+		compiler = compiler.withLoader(mustacheTemplateLoader).withCollector(collector());
+
+		return compiler;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty("spring.mustache.formatter.value")
+	public Mustache.Formatter mustacheFormatter() {
+		return (value) -> this.mustache.getFormatter().getValue();
```
wilkinsona: This ignores the input `value` and always returns the value of the `spring.mustache.formatter.value` property.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mustache/MustacheAutoConfiguration.java
```diff
@@ -79,8 +81,31 @@ public void checkTemplateLocationExists() {
...
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	@ConditionalOnProperty("spring.mustache.escaper.value")
+	public Mustache.Escaper mustacheEscaper() {
+		return (value) -> this.mustache.getEscaper().getValue();
```
wilkinsona: This ignores the input `value` and always returns the value of the `spring.mustache.escaper.value` property.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mustache/MustacheProperties.java
```diff
@@ -42,6 +43,26 @@
...
+
+	/**
+	 * Value for any variable that resolves to null.
+	 */
+	private String nullValue;
+
+	/**
+	 * Handles converting objects to strings.
+	 */
+	private final Formatter formatter = new Formatter();
```
wilkinsona: This needs to be an implementation of the `Mustache.Formatter` interface. Unless we require that implementation to have a default constructor and treat the property as the fully-qualified name of a class that we instantiate, I don't think it can be configured using properties.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mustache/MustacheProperties.java
```diff
@@ -42,6 +43,26 @@
...
+
+	/**
+	 * Handles escaping characters in substituted text.
+	 */
+	private final Escaper escaper = new Escaper();
```
wilkinsona: This needs to be an implementation of the `Mustache.Escaper` interface. Unless we require that implementation to have a default constructor and treat the property as the fully-qualified name of a class that we instantiate, I don't think it can be configured using properties.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/PropertyMapper.java
```diff
@@ -333,6 +334,22 @@ public void to(Consumer<T> consumer) {
...
 		}
 
+		/**
+		 * Complete the mapping by creating a new instance from the non-filtered value.
+		 * @param <R> the resulting type
+		 * @param factory the factory used to create the instance
+		 * @param defaultValue the default value used in case of predicate is false.
+		 * @return the instance
+		 */
+		public <R> R toInstance(Function<T, R> factory, R defaultValue) {
```
wilkinsona: I don't think this method is needed. The existing `toInstance(Function<T, R> factory)` method should do what we need when given a method reference to one of `Mustache.Compiler`'s configuration methods that return a new `Compiler` instance.

ayudovin:
@wilkinsona , I would like to work on this issue

wilkinsona:
Thanks for the PR, @ayudovin. I'm going to close this one as it doesn't quite take the right approach. We can revisit this once the necessary design work has been done for #15566.

Issue #2652, opened 2015-03-14
composite HealthIndicator that runs them in parallel


snicoll:
Right now, invoking the health endpoint lead to a synchronous invocations of all underlying `HealthIndicator` that have been configured. It would be nice to offer an option to run these in parallel instead.

philwebb:
Definitely in the "nice to have" bucket. I imaging that the `health` endpoint isn't hit all that often so performance probably isn't a big deal.

snicoll:
@philwebb I had the idea when I saw #2630

fatroom:
I would also like to have support for this. My project got a bunch of health checks and some of them can run up to few seconds. So in total performance of /health endpoint is terrible with sequential approach.

fjcano:
:+1: 
Some of my projects have 20+ health checks.
For now we use our homebrew solution to do them in parallel.

Would love to see support for parallel health checks in Spring Boot out of the box.
Or at least a way to inject our own `CompositeHealthIndicator`.

prismec:
And it would be nice to include a standard detail information how long a certain health indicator took to execute.

erikgreif-acc:
Real-world example here:

Some services have requirements to respond within a given time limit. This time limit may be set by an external organization, not always under the control of the devs. This kind of timeout gets interpreted as a DOWN state, wakes devs up in the middle of the night, reboots service instances, and all the while provides no context as to which component was experiencing high latency (because there was no HTTP response). 

I have successfully added timeouts per-actuator, but with 20 checks that require timeouts of around 5 seconds (any less would cause alerts for minor network blips), we are asking our caller to wait up to 1:40 for a response. At my org, this is absurd, and nobody wants to permit this behavior.

rmie:
I second everything @erikgreif-acc said.

Sometimes I run into a situation in which a time out occurs but it is unclear which of the components caused it. So I'd like to propose a timeout for the parallel execution after which every unresponsive component is considered/reported as down. To keep the existing behavior the default for this timeout should be infinite.

And while working on it, exporting metrics (number of total, successful and failed executions, sum of execution time for successful executions for each component) would make alerting (and inhibiting) on health related issues easier and would help with dissecting problems faster.

jim-mclean:
+1 for this feature, or for some sort of async checking...

[...]

PR #15770, opened 2019-01-23
Run CompositeHealthIndicator in parallel

RikEnde:
See https://github.com/spring-projects/spring-boot/issues/2652

**Problem:** 
The CompositeHealthIndicator polls all the HealthIndicators in the registry sequentially. I have a number of HealtIndicator components that poll remote systems and report on latency and status, and the latency really adds up, especially if there are time-outs involved. 

**Solution:** 
This PR polls the HealthIndicators concurrently. It preserves the original ordering, and outputs the same as the old implementation, except the runtime is proportional to that of the slowest component, rather than the aggregate of all components. The impact of this change should be very small, unless someone's code absolutely depends on HealthIndicators being polled sequentially but in unknown order.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/CompositeHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/CompositeHealthIndicator.java
@@ -91,17 +91,14 @@ public class CompositeHealthIndicator implements HealthIndicator {
 		return this.registry;
 	}
 
-	// TODO Guarantee preserved ordering
-	// TODO it doesn't seem important but don't make assumptions on other people's behalf
 	@Override
 	public Health health() {
 		/* Gather healths in parallel */
 		Map<String, Health> healths = this.registry.getAll().entrySet().parallelStream()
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@RikEnde Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15770)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15770) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@RikEnde Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=15770)!

RikEnde:
I'm pretty sure my commit didn't break the build or cause that timeout...

snicoll:
@RikEnde you are right, your change did not break the build.

We have a task execution infrastructure now in place and we'd rather use that if it's available rather than using `parallelStream`. Thanks for the PR, in any case.

RikEnde:
Fair enough. Are you referring to the ConcurrentTaskExecutor, or something new?

Issue #15679, opened 2019-01-11
JspTemplateAvailabilityProvider causes java.security.AccessControlException


dezmond78:
We are running a Spring Boot powered application in form of a WAR file in a dedicated Tomcat server due to customer requirements. We're also using the Java Security Manager in order to fullfill our security requirements.

Yesterday we finally upgraded our stack to Spring Boot 2.1.1 (from 2.0.2). Since then our application is unable to start because `JspTemplateAvailabilityProvider` is causing a `java.security.AccessControlException`.

```
...
Caused by: java.security.AccessControlException: access denied ("java.io.FilePermission" "src/main/webapp/index" "read")
        at java.security.AccessControlContext.checkPermission(AccessControlContext.java:472)
        at java.security.AccessController.checkPermission(AccessController.java:884)
        at java.lang.SecurityManager.checkPermission(SecurityManager.java:549)
        at java.lang.SecurityManager.checkRead(SecurityManager.java:888)
        at java.io.File.exists(File.java:814)
        at org.springframework.boot.autoconfigure.web.servlet.JspTemplateAvailabilityProvider.isTemplateAvailable(JspTemplateAvailabilityProvider.java:45)
        at org.springframework.boot.autoconfigure.template.TemplateAvailabilityProviders.findProvider(TemplateAvailabilityProviders.java:158)
        at org.springframework.boot.autoconfigure.template.TemplateAvailabilityProviders.getProvider(TemplateAvailabilityProviders.java:145)
        at org.springframework.boot.autoconfigure.template.TemplateAvailabilityProviders.getProvider(TemplateAvailabilityProviders.java:119)
        at org.springframework.boot.autoconfigure.web.servlet.WelcomePageHandlerMapping.welcomeTemplateExists(WelcomePageHandlerMapping.java:69)
        at org.springframework.boot.autoconfigure.web.servlet.WelcomePageHandlerMapping.<init>(WelcomePageHandlerMapping.java:59)
        at org.springframework.boot.autoconfigure.web.servlet.WebMvcAutoConfiguration$WebMvcAutoConfigurationAdapter.welcomePageHandlerMapping(WebMvcAutoConfiguration.java:362)
        at org.springframework.boot.autoconfigure.web.servlet.WebMvcAutoConfiguration$WebMvcAutoConfigurationAdapter$$EnhancerBySpringCGLIB$$84cc0ba.CGLIB$welcomePageHandlerMapping$10(<generated>)
[...]

PR #15778, opened 2019-01-24
JspTemplateAvailabilityProvider causes java.security.AccessControlException

ayudovin:
this [bug](https://github.com/spring-projects/spring-boot/issues/15679)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/JspTemplateAvailabilityProvider.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/JspTemplateAvailabilityProvider.java
@@ -17,6 +17,7 @@
 package org.springframework.boot.autoconfigure.web.servlet;
 
 import java.io.File;
+import java.security.AccessControlException;
 
 import org.springframework.boot.autoconfigure.template.TemplateAvailabilityProvider;
 import org.springframework.core.env.Environment;
@@ -30,6 +31,7 @@ import org.springframework.util.ClassUtils;
  * @author Andy Wilkinson
  * @author Stephane Nicoll
  * @author Madhura Bhave
+ * @author Artsiom Yudovin
  * @since 1.1.0
  */
 public class JspTemplateAvailabilityProvider implements TemplateAvailabilityProvider {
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/JspTemplateAvailabilityProvider.java
```diff
@@ -38,11 +40,17 @@
...
-			String resourceName = getResourceName(view, environment);
-			if (resourceLoader.getResource(resourceName).exists()) {
-				return true;
+
+			try {
+				String resourceName = getResourceName(view, environment);
+				if (resourceLoader.getResource(resourceName).exists()) {
+					return true;
+				}
+				return new File("src/main/webapp", resourceName).exists();
```
wilkinsona: The scope of this `try` block is broader than necessary. It only needs to contain this line.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/JspTemplateAvailabilityProvider.java
```diff
@@ -38,11 +40,17 @@
...
+			}
+			catch (AccessControlException ex) {
+				return false;
```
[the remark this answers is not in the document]
ayudovin: in this case, Does it make sense to add `private final Log logger = LogFactory.getLog(JspTemplateAvailabilityProvider.class);` and log exception in catch ?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/JspTemplateAvailabilityProvider.java
ayudovin: I updated by your suggestion.

ayudovin: I updated by your suggestion.

Issue #15760, opened 2019-01-22
Slow binding of @ConfigurationProperties with large object graph


dsyer:
@snicoll and I were looking at startup times in Initializr - there's a 2 second delay while the `InitializrProperties` is being bound. Here are some benchmarks:

```
class                method  prof       size     median    mean      range
PropertiesBenchmarkIT  auto  small      1        3803.788  3946.267  96.863
PropertiesBenchmarkIT  auto  medium     134      62.431    64.536    1.551
PropertiesBenchmarkIT  auto  large      1093     0.641     0.657     0.014
```

UPDATE: I edited the table above to make the "size" metric actually match the size
of the properties bound (I don't know where JMH gets its counter metrics from sometimes).

The "large" sample above is using the same YAML as production app, and
0.641 ops/sec is getting on for 2 seconds per op (1.5 anyway). So it
sucks.

Looking in Flight Controller I'm seeing GC pressure from `ConfigurationPropertyName.buildString()`. 

List of property names passing through `ConfigurationPropertyName.buildString()` on application start up: [prop.txt](https://github.com/spring-projects/spring-boot/files/2783054/prop.txt). We can see that sometimes the method is called multiple times for the same property name (but only for indexed properties), and also that there is a lot of duplication in the indexed properties. Maybe that can be optimized?

philwebb:
Perhaps some cases we fixed in #13414 were missed. @dsyer is that `PropertiesBenchmarkIT` part of https://github.com/dsyer/spring-boot-startup-bench ? I can't seem to find it.

dsyer:
There's a branch of initializr with the benchmark tests here: https://github.com/dsyer/initializr/tree/bench

dsyer:
Here's another (different) measurement:

```
Benchmark                 (prof)  Mode  Cnt  Score   Error  Units
LauncherBenchmarkIT.auto  medium  avgt   10  0.323 ± 0.008   s/op
LauncherBenchmarkIT.auto   small  avgt   10  0.271 ± 0.006   s/op
LauncherBenchmarkIT.auto   large  avgt   10  1.936 ± 0.047   s/op
```

Where "small", "medium" and "large" mean the same as above. The benchmark this time is just the time to start an empty Spring Boot app with just `@EnableConfigurationProperties(InitializrProperties.class)`.

[...]

PR #15782, opened 2019-01-24
Optimize ConfigurationPropertyName

dreis2211:
Hi,

while looking into some possible improvements for #15760 I noticed a possible optimization that effectively turns around the iteration order when checking for element equality of two `ConfigurationPropertyName` instances via `elementEquals`.

This is largely based on the educated guess that child elements will likely differ while parent elements stay the same. (Almost everything is prefixed with a common `spring` or `initializr`, but the following elements are often different).

The reason for this improvement came apparent when I profiled the `BinderApplication` provided by @dsyer. There are a lot of calls to `elementEquals` and the corresponding `charAt` methods called in it.

I also ran the provided benchmarks of #15760  and got these results:

**Before**
grafik

```
Benchmark                        (prof)   Mode  Cnt      Score      Error  Units
PropertiesBenchmarkIT.auto       medium  thrpt   10    101,083 ±   47,304  ops/s
PropertiesBenchmarkIT.auto:size  medium  thrpt   10   1340,000                 #
PropertiesBenchmarkIT.auto        small  thrpt   10  11800,645 ± 3892,176  ops/s
PropertiesBenchmarkIT.auto:size   small  thrpt   10     10,000                 #
PropertiesBenchmarkIT.auto        large  thrpt   10      1,292 ±    0,261  ops/s
PropertiesBenchmarkIT.auto:size   large  thrpt   10  10930,000                 #

Benchmark                 (prof)  Mode  Cnt  Score   Error  Units
LauncherBenchmarkIT.auto  medium  avgt   10  1,106 ± 0,159   s/op
LauncherBenchmarkIT.auto   small  avgt   10  0,777 ± 0,092   s/op
LauncherBenchmarkIT.auto   large  avgt   10  2,562 ± 0,315   s/op
```

**After**
grafik

```
Benchmark                        (prof)   Mode  Cnt      Score      Error  Units
PropertiesBenchmarkIT.auto       medium  thrpt   10    181,640 ±   94,044  ops/s
PropertiesBenchmarkIT.auto:size  medium  thrpt   10   1340,000                 #
PropertiesBenchmarkIT.auto        small  thrpt   10  13601,079 ± 5695,056  ops/s
PropertiesBenchmarkIT.auto:size   small  thrpt   10     10,000                 #
PropertiesBenchmarkIT.auto        large  thrpt   10      2,505 ±    0,650  ops/s
PropertiesBenchmarkIT.auto:size   large  thrpt   10  10930,000                 #

Benchmark                 (prof)  Mode  Cnt  Score   Error  Units
LauncherBenchmarkIT.auto  medium  avgt   10  1,078 ± 0,109   s/op
LauncherBenchmarkIT.auto   small  avgt   10  0,783 ± 0,093   s/op
LauncherBenchmarkIT.auto   large  avgt   10  2,309 ± 0,223   s/op
```

I think this shows that we save quite a bit of cycles in the large example, while not loosing too much in the small one. But in the end it is an educated guess, so feel free to decline this PR.

Let me know what you think.
Cheers,
Christoph

P.S.: Don't pay too much attention to the timings in the screenshots. Due to the profiling overhead they are quite a bit off. The overall distribution and total calls are more important here.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/ConfigurationPropertyName.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/ConfigurationPropertyName.java
@@ -237,7 +237,7 @@ public final class ConfigurationPropertyName
 		if (this.getNumberOfElements() >= name.getNumberOfElements()) {
 			return false;
 		}
-		for (int i = 0; i < this.elements.getSize(); i++) {
+		for (int i = this.elements.getSize() - 1; i >= 0; i--) {
 			if (!elementEquals(this.elements, name.elements, i)) {
 				return false;
 			}
@@ -309,7 +309,7 @@ public final class ConfigurationPropertyName
 				&& other.elements.canShortcutWithSource(ElementType.UNIFORM)) {
 			return toString().equals(other.toString());
 		}
-		for (int i = 0; i < this.elements.getSize(); i++) {
+		for (int i = this.elements.getSize() - 1; i >= 0; i--) {
 			if (!elementEquals(this.elements, other.elements, i)) {
 				return false;
 			}


wilkinsona:
#### Main method

| | Baseline | New |
| - | - | - |
| 1. | 1 | 1 |
| 1. | 1 | 1 |
| 1. | 1 | 1 |
| 1. | 1 | 1 |
| 1. | 1 | 1 |

wilkinsona:
Wow! This makes a significant difference to the time it takes to bind `InitializrProperties` with the large YAML file. I used the following app:

```java
@EnableConfigurationProperties(InitializrProperties.class)
public class BinderPerformanceApplication {

		SpringApplication.run(BinderPerformanceApplication.class, args);
	}

}
```

The times in the following table are the JVM running time that's logged on startup:

| | Baseline | New |
| - | -: | -: |
| | 2.279 | 1.54 |
| | 2.177 | 1.42 |
| | 2.151 | 1.555 |
| | 2.703 | 1.459 |
| | 2.445 | 1.452 |
| | 2.358 | 1.613 |
| | 2.238 | 1.745 |
| | 2.445 | 1.763 |
| | 2.296 | 1.422 |
| | 2.486 | 1.456 |
| Mean | 2.358 | 1.543 |
| Range | 0.552 | 0.343 |

It's 800ms faster, reducing startup time by 34%.

dreis2211:
Wow. that's even more than on my machine. I'm glad that this seems to be a worthwhile improvement so far.

wilkinsona:
And here are equivalent numbers for the same app with a much smaller YAML file:

|       | Baseline |  New  |
| ----- | -------: | ----: |
|       |    0.780 | 0.744 |
|       |    0.742 | 0.729 |
|       |    0.759 | 0.727 |
|       |    0.756 | 0.735 |
|       |    0.767 | 0.748 |
|       |    0.747 | 0.747 |
|       |    0.757 | 0.738 |
|       |    0.755 | 0.730 |
|       |    0.770 | 0.735 |
|       |    0.767 | 0.740 |
|  Mean |    0.760 | 0.737 |
| Range |    0.038 | 0.021 |

New's faster, but by nowhere near as much. That's to be expected with far fewer properties to compare.

This change will be slower in scenarios where there are a lot of properties with the same number of elements that differ towards the beginning rather than at the end. I think configuration of that kind will be the exception rather than the rule, and even when the configuration has that structure, the performance will be no worse than it is today for what I believe is the more common case.

dreis2211:
@wilkinsona Thanks for checking the PR btw and running further benchmarks.

wilkinsona:
Thanks very much for the PR, @dreis2211. This one is now in master and is responsible for the lion's share of an [almost 40% reduction in startup time](
https://github.com/spring-projects/spring-boot/issues/15760#issuecomment-458102857).

dreis2211:
Thanks for reviewing it, @wilkinsona . On an unrelated note - just noticed that the merge says "Christoph Dreiss" instead of "Christoph Dreis". Is that automatically generated from somewhere - or something you typed manually?

wilkinsona:
Argh. Sorry, that's a typo on my part.

dreis2211:
Don't worry. Just thought I configured something wrong myself.

PR #15784, opened 2019-01-25
Remove duplicate code from HealthWebEndpointResponseMapper

vpavic:
While debugging I noticed there's some duplicate code in `HealthWebEndpointResponseMapper` that could easily be removed.


vpavic:
This should be moved to [`2.1.3` milestone](https://github.com/spring-projects/spring-boot/milestone/133), right?

snicoll:
Yep. I forgot, thanks.

PR #15793, opened 2019-01-26
Use Assertions.contentOf() where possible

dreis2211:
Hi,

this PR makes use of `Assertions.contentOf()` more often and includes some general polishing.

I just noticed after pushing that a change in `RestarterTests` is included, that is not using `Assertions.contentOf()` but cleans the assertion on an `InputStream` via `hasContent`. I hope that's okay - otherwise I'll revert this for the PR.

Cheers,
Christoph


snicoll:
Thanks again Christoph!

PR #15799, opened 2019-01-28
Polish

izeye:

Note this PR also changes not to handle `null` for `HttpStatus` in `RestTemplateExchangeTags.outcome()` as it isn't covered by any test and it doesn't seem to be possible to be `null` based on the Javadoc for `ClientHttpResponse. getStatusCode()` as follows:

```
	/**
	 * Return the HTTP status code of the response.
	 * @return the HTTP status as an HttpStatus enum value
	 * @throws IOException in case of I/O errors
	 * @throws IllegalArgumentException in case of an unknown HTTP status code
	 * @see HttpStatus#valueOf(int)
	 */
	HttpStatus getStatusCode() throws IOException;
```


wilkinsona:
Thanks once again, @izeye!

Issue #15797, opened 2019-01-28
Undertow connection timeout is set incorrectly


jan-krakora:
Spring Boot version: 2.1.1
Undertow version: 2.0.16 (Spring Boot's managed dependency)

Setting
```
server
    connection-timeout: whatever
```
then the autoconfiguration class UndertowWebServerFactoryCustomizer sets this property as **socket option**

`builder.setSocketOption(UndertowOptions.NO_REQUEST_TIMEOUT, (int) connectionTimeout.toMillis())`

This is ignored by Undertow since this property is expected to be set in **server options**. Therefore it should be

`builder.setServerOption(UndertowOptions.NO_REQUEST_TIMEOUT, (int) connectionTimeout.toMillis())`

see Undertow#start() where is merged a default value for this option with all options set in the builder.
```
OptionMap serverOptions = OptionMap.builder()
                    .set(UndertowOptions.NO_REQUEST_TIMEOUT, 60 * 1000)
                    .addAll(this.serverOptions)
                    .getMap();
```

bbzg:
Yes! I just found this as well! Can confirm!

wilkinsona:
Closing in favour of #15822.

MarsYoung:
it doesn't seem  to work. and my spring-boot version is 1.5.4,and my code is 
`	final int connectTimeOut=5000;
	@Bean
	public EmbeddedServletContainerFactory undertow() {
		UndertowEmbeddedServletContainerFactory factory = new  UndertowEmbeddedServletContainerFactory();
		factory.addBuilderCustomizers(builder ->
				{
				builder.setSocketOption(UndertowOptions.NO_REQUEST_TIMEOUT,
						connectTimeOut);
				}
		);
		return factory;
	}`

philwebb:
@MarsYoung Spring Boot 1.5 is end of life. Please upgrade.

MarsYoung:
@philwebb It's hard to update. I have an old project that would take much time to upgrade because there are a lot of old projects using in my project such as spring-boot-starter-jersey, spring-data-influx etc. I have try ,but too much error occurs, finnally I revert all the code. Is there any other solution for this problem based on spring boot 1.5?

philwebb:
@MarsYoung You may be able to fork the code and backport the fix to your own branch but I imagine working though the upgrade pain will ultimately be easier. I'm afraid we can't support the older code indefinitely.

PR #15822, opened 2019-01-31
Undertow connection timeout is set incorrectly

ayudovin:
this [bug](https://github.com/spring-projects/spring-boot/issues/15797)


wilkinsona:
Thanks very much, @ayudovin. The proposed changes have been merged into 2.1.x and forwards into master.

PR #15836, opened 2019-02-02
Add a ConcurrentCompositeHealthIndicator

RikEnde:
See https://github.com/spring-projects/spring-boot/issues/2652

I submitted a PR for this issue before, but it wasn't very good. I hope this one is better. 

**Problem:** 
Right now, invoking the health endpoint lead to a synchronous invocation of all underlying HealthIndicator instance(s) that have been configured. It would be nice to offer an option to run these in parallel instead.

**Real world example:**
I have a number of HealtIndicator components that poll remote systems and report on latency and status, and the latency really adds up, especially if there are time-outs involved.

**Solution:**
This PR adds a ConcurrentCompositeHealthIndicator that can run the HealthIndicator instances on the  ThreadPoolTaskExecutor. 

Usage in code:
```
    @Bean
    public HealthEndpoint healthEndpoint(HealthAggregator healthAggregator,
                                         HealthIndicatorRegistry registry,
                                         ThreadPoolTaskExecutor executor
                                         ) {
        return new HealthEndpoint(
            new ConcurrentCompositeHealthIndicator(healthAggregator, registry, executor, Duration.ofMillis(1100)));
    }
```

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/CompositeHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/CompositeHealthIndicator.java
@@ -90,6 +90,14 @@ public class CompositeHealthIndicator implements HealthIndicator {
 		return this.registry;
 	}
 
+	/**
+	 * Return the {@link HealthAggregator} of this instance.
+	 * @return the aggregator for {@link HealthIndicator health indicators}
+	 */
+	public HealthAggregator getAggregator() {
+		return this.aggregator;
+	}
+
 	@Override
 	public Health health() {
 		Map<String, Health> healths = new LinkedHashMap<>();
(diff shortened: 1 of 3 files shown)


snicoll:
@rikende did you mean to close this PR?

RikEnde:
While working on this, I found it could be solved at the application level once we move to spring boot 2.1.x

CLOUGH:
@RikEnde  could you share the workaround that you found.

RikEnde:
> @RikEnde could you share the workaround that you found.

Sure, but keep in mind this hasn't been touched in 2 years, it probably won't work for the latest version of spring boot https://github.com/RikEnde/concurrent-actuator-test.git

RikEnde:
> @RikEnde could you share the workaround that you found.



> @RikEnde could you share the workaround that you found.

The code from this PR could be implemented entirely in domain code, requiring no framework changes. However, this was 2 years old. It doesn't even compile against the current version.

Issue #15815, opened 2019-01-30
/actuator/scheduletasks endpoint fails with null pointer exception when there is custom SchedulingConfigurer


nguyenquoc:
If a custom SchedulingConfigurer is registered with spring boot, then the scheduledtasks actuator will fail with a null pointer exception.

Spring boot 2.1.2

Stack trace:
```
java.lang.NullPointerException: null
	at java.util.Objects.requireNonNull(Objects.java:203)
	at java.util.Optional.<init>(Optional.java:96)
	at java.util.Optional.of(Optional.java:108)
	at java.util.stream.FindOps$FindSink$OfRef.get(FindOps.java:193)
	at java.util.stream.FindOps$FindSink$OfRef.get(FindOps.java:190)
	at java.util.stream.FindOps$FindOp.evaluateSequential(FindOps.java:152)
	at java.util.stream.AbstractPipeline.evaluate(AbstractPipeline.java:234)
	at java.util.stream.ReferencePipeline.findFirst(ReferencePipeline.java:464)
	at org.springframework.boot.actuate.scheduling.ScheduledTasksEndpoint$TaskDescription.of(ScheduledTasksEndpoint.java:130)
	at org.springframework.boot.actuate.scheduling.ScheduledTasksEndpoint$TaskDescription.access$900(ScheduledTasksEndpoint.java:108)
	at org.springframework.boot.actuate.scheduling.ScheduledTasksEndpoint.lambda$scheduledTasks$1(ScheduledTasksEndpoint.java:63)
```

Sample code: https://stackoverflow.com/questions/14630539/scheduling-a-job-with-spring-programmatically-with-fixedrate-set-dynamically/14632758#14632758

The endpoint defined in ScheduledTasksEndpoint.java can only handle CronTrigger's and PeriodicTrigger's. Anything else will cause the endpoint to fail with a null pointer exception

```
// org.springframework.boot.actuate.scheduling.ScheduleTaskEndpoint.java

        private static ScheduledTasksEndpoint.TaskDescription describeTriggerTask(TriggerTask triggerTask) {
            Trigger trigger = triggerTask.getTrigger();
            if (trigger instanceof CronTrigger) {
                return new ScheduledTasksEndpoint.CronTaskDescription(triggerTask, (CronTrigger)trigger);
            } else if (trigger instanceof PeriodicTrigger) {
                PeriodicTrigger periodicTrigger = (PeriodicTrigger)trigger;
[...]

PR #15840, opened 2019-02-04
/actuator/scheduletasks endpoint fails with null pointer exception when there is custom SchedulingConfigurer

ayudovin:
this [bug](https://github.com/spring-projects/spring-boot/issues/15815)


wilkinsona:
Thanks for the PR, but I'd like to take a different approach with this one. I think we should see if we can include the tasks with custom triggers in the endpoint's output.

PR #15842, opened 2019-02-04
Make RestTemplateMetricsConfiguration conditional on a RestTemplateBuilder bean

Raheela1024:

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/client/HttpClientMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/web/client/HttpClientMetricsAutoConfiguration.java
@@ -27,6 +27,7 @@ import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.web.client.RestTemplateAutoConfiguration;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks very much for the PR, @Raheela1024. The first commit (
https://github.com/spring-projects/spring-boot/pull/15842/commits/7beca9e2a9f6b08dfe379671661ab11247ffcf3f) is great. I'm not 100% sure that we need the changes in the other two commits (other than adding you as an `@author`) but I could easily be missing something. Can you please fill in the blanks for me and explain why they're needed? If they're not strictly needed then I think it would be better to do without them (for now at least) as we try to keep changes to the minimum that's needed to fix and test a particular issue.

Raheela1024:
> Thanks very much for the PR, @Raheela1024. The first commit ([7beca9e](https://github.com/spring-projects/spring-boot/commit/7beca9e2a9f6b08dfe379671661ab11247ffcf3f)) is great. I'm not 100% sure that we need the changes in the other two commits (other than adding you as an `@author`) but I could easily be missing something. Can you please fill in the blanks for me and explain why they're needed? If they're not strictly needed then I think it would be better to do without them (for now at least) as we try to keep changes to the minimum that's needed to fix and test a particular issue.

Actually, In first commit i just added test case for added annotation but while running whole test case file getting below exception in all test cases instead of my added test cases.

```
io.micrometer.core.instrument.search.MeterNotFoundException: Unable to find a meter that matches all the requirements at once. Here's what was found:
   FAIL: No meter with name 'http.client.requests' was found.
   FAIL: No meters have the required tag 'uri'.
```
While debugging the issue i got to know in `InterceptingHttpAccessor`  `ClientHttpRequestInterceptor` is not initiated but when i remove `@ConditionalOnBean(RestTemplateBuilder.class)` from `RestTemplateMetricsConfiguration`   `ClientHttpRequestInterceptor` is initiated and test cases are working fine.

In test cases we have below code that is update the `http.client.requests` meter size into `MeterRegistry`.

```
for (int i = 0; i < 3; i++) {
			restTemplate.getForObject("/test/" + i, String.class);
		} 
```

May b i fixed in wrong way please let me know if there is any other way?

wilkinsona:
Thanks very much for the explanation, @Raheela1024.

I think the problem is that `HttpClientMetricsAutoConfiguration` is being processed before `RestTemplateAutoConfiguration`. As a result, when `RestTemplateMetricsConfiguration` is conditional on a `RestTemplateBuilder` bean, it backs off as there isn't one at that time. If I'm right, the problem can be fixed by adding `RestTemplateAutoConfiguration.class` to the existing `@AutoConfigureAfter` on `HttpClientMetricsAutoConfiguration`.

Raheela1024:
Thanks very much @wilkinsona  you solved my Problem. Thanks again kindly review and proceed.

wilkinsona:
Thanks very much, @Raheela1024. The proposed changes have now been merged into master. If you are interested, I polished things a little bit in https://github.com/spring-projects/spring-boot/commit/22192c2b5cd56c45802526ec7190e945143dc980. The biggest thing, and it was still really minor, was to move the new test method up next to the others rather than it being at the end of the class after the helper methods.

PR #15856, opened 2019-02-06
Add property support for Neo4j additional causal cluster URIs.

meistermeier:
Expose the setting of additional URIs for cluster members to the users by offering a property.
The API in Neo4j-OGM exists for a long time but was never made accessible through the auto configuration.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jProperties.java
@@ -34,6 +34,7 @@ import org.springframework.util.ClassUtils;
  * @author Vince Bickers
  * @author Aurélien Leboulanger
  * @author Michael Simons
+ * @author Gerrit Meier
  * @since 1.4.0
  */
 @ConfigurationProperties(prefix = "spring.data.neo4j")
@@ -51,6 +52,11 @@ public class Neo4jProperties implements ApplicationContextAware {
 	 * URI used by the driver. Auto-detected by default.
 	 */
 	private String uri;
+
+	/**
+	 * (Optional) additional core URIs used for causal cluster.
+	 */
+	private String[] uris;
 
 	/**
 	 * Login user of the server.
@@ -88,6 +94,14 @@ public class Neo4jProperties implements ApplicationContextAware {
 
 	public void setUri(String uri) {
 		this.uri = uri;
+	}
+
+	public String[] getUris() {
+		return this.uris;
+	}
+
+	public void setUris(String[] uris) {
+		this.uris = uris;
 	}
 
 	public String getUsername() {
@@ -153,7 +167,10 @@ public class Neo4jProperties implements ApplicationContextAware {
 		if (this.uri != null) {
 			builder.uri(this.uri);
 		}
-		else {
+		if (this.uris != null) {
+			builder.uris(this.uris);
+		}
+		if (this.uri == null && (this.uris == null || this.uris.length == 0)) {
 			configureUriWithDefaults(builder);
 		}
 		if (this.username != null && this.password != null) {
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the pull request.

While I realise that the OGM `Builder` class has both `uri` and `uris` methods and the proposal here is a reflection of that, I find it confusing and I think our users will too. In addition to having to use two separate properties to configure the URIs that will be used, it looks like setting the `uris` property requires the value of the `uri` property to change to be a `bolt+routing` URI.

I think a more Boot-like approach would be to support a single property for configuring one or more URIs. The property can then be configured as a single comma-separated value, as multiple `[n]` indexed properties, or as a list in YAML. Can we please take a step back and investigate and approach like this?

wilkinsona:
Let's place this one on hold until we know what's happening with the API in Neo4j-OGM.

snicoll:
@meistermeier do you have an update for us?

meistermeier:
@snicoll I just updated the PR with a updated, rebased and squashed force push. It is no allowed (in OGM) to only supply the `uris`. Because of backwards compatibility the old `url` is still valid but not mandatory anymore if `uris` get provided.

snicoll:
@meistermeier thanks for the feedback. I am confused by “It is no allowed … to only supply the uris” and then “the old url is still valid but not mandatory anymore”

Looking at the proposal, I'd like to deprecate `uri` and move everyone off to `uris`. If one uri is specified using the `uris` parameter then it should do what it did before and a user may add more than one to access the feature we're discussing here. It shouldn't change anything to you except Spring Boot would only configure `uris` going forward.

Does that make sense?

meistermeier:
The "It is no allowed..." should be a "It is now allowed...", sorry for the confusion.
Deprecating the `url` is fine for me because this is what will also happen in OGM in the future. So let's bring the future to Spring Boot first 😄

meistermeier:
We discussed this internally (again) and came to the conclusion that it might be better to not deprecated the `uri` field.
Even more the `uris` field should get deprecated right away because we want users to use a single entry point into the cluster in the future.
The current Neo4j driver usage forces us to make the `uris` attribute accessible to the user for providing multiple possible hosts.
In an upcoming version of the driver this will not be necessary anymore but the user will only provide one uri.
The problem is that before we have a Spring Data Neo4j version with a driver that supports it, we have to expose the `uris` property so that users can set up their cluster configuration.

Speaking about the future, we will create a configuration that is more closer to the driver, so the situation of unfitting property states like we have now -between SDN, Neo4j-OGM and the Java driver- will not occur anymore.

wilkinsona:
If `uris` should be deprecated right away then I am not sure that we should add it in the first place, particularly not without a replacement that could be used instead so that users can fix the deprecation warning. If such a replacement already exists then users could just use it, if no such replacement exists then we either need the property undeprecated or we should not add it at all.

meistermeier:
I did not add any changes / deprecations because I think we shouldn't deprecate it "on release". The current situation and roughly the next 9 months (at least) will not provide any possibilities for the users to supply additional routing / host information besides the `uris`.
There is no workaround except of creating a `Configuration` bean manually to achieve this through Spring Boot properties at the moment.
My comment was more about to give you an explanation why we cannot just deprecate the `uri` property with this change.

wilkinsona:
Thanks, @meistermeier. It sounds like we've circled back around to the situation that I commented on [above](https://github.com/spring-projects/spring-boot/pull/15856#issuecomment-461353904):

> While I realise that the OGM Builder class has both uri and uris methods and the proposal here is a reflection of that, I find it confusing and I think our users will too. In addition to having to use two separate properties to configure the URIs that will be used, it looks like setting the uris property requires the value of the uri property to change to be a bolt+routing URI.

Unfortunately, unless I've misunderstood the current situation, I think we should decline this proposal and wait for things to be straightened out a little in Neo4j. In the meantime, users can configure their own `Configuration` bean.

wilkinsona:
I've closed this one for now at least. If I have misunderstood please let us know and we can reconsider.

meistermeier:
After re-thinking the whole PR I think it is the best for Spring Boot to take the step back from the PR. Unfortunately the decision of removing or at least deprecating the `uris` happened while the PR was already opened. On the other hand we could improve Neo4j-OGM as a side-effect.
Thanks for your time and thoughts.

PR #15858, opened 2019-02-06
2.10.2 fixes a major bug in connection pool

gkorland:
Jedis 2.9.1 has a major bug in the connection pool which was fixed in 2.10.2 see - https://github.com/xetorthio/jedis/issues/1945


snicoll:
@gkorland thank you for the PR but we don't need those as we have a semi-automated process to upgrade dependencies and we'll upgrade this one in due course.

(There is a note about that in the template when you submit a PR).

Issue #11721, opened 2018-01-22
Schema generation properties not well documented


ghost:
Unfortunately my previous issue #11678 was closed as duplicate, though I don't think it is one.

In the issue I argued that the documentation is not entirely clear on which of several similar properties have precedence.

The referenced issue suggest other default values of certain properties in some contexts.

I would be glad if the documentation could be improved.

snicoll:
@FelixRiemann you were referring to several issues into one and the problem of not enabling `create-drop` with Flyway/Liquibase has been  addressed already (which means you don't have to set that `ddl.auto` either.

As for the documentation, it is not  clear what you're asking really. So please clarify and we can see what can be improved in the documentation.

In the future, please comment on the issue rather than creating another one. We can always refine and  reopen.

FelixRiemann:
There are four properties I am interested in:
* (1) `spring.jpa.generate-ddl` (boolean)
* (2) `spring.jpa.properties.javax.persistence.schema-generation.database.action` (one of `none`, `drop`, `create`, `drop-and-create`)
* (3) `spring.jpa.hibernate.ddl-auto` (one of `none`, `validate`, `update`, `create`, `create-drop`)
* (4) `spring.datasource.initialize` (boolean)

From the documentation, the following is unclear to me:
1. Both (1) and (2) exist to enable/disable schema generation. I would like to know which of both has precedence.
1. Property (4) governs whether the schema should be generated from DML scripts. What happens if this is enabled together with one of (1) or (2)?
1. Then there is (3), which is similar to (2). What happens if they contradict each other?
1. What happens if (1) and (3) contradict each other (and here by *contradict* I mean any combination other than (none, none) and (true, create))?
1. If I do not set (3) specifically, will its default value override (1)? In Spring Boot 1.5.9 this is the case, as I pointed out in #11678. But even if the default value of (3) will be `none` when Flyway/Liquibase is detected, as suggested in #9262, the question remains whether (3) overrides (1) if I do not use (3).

> In the future, please comment on a declined issue rather than creating another one. We can always refine and reopen.

Will keep that in mind.

matthiasstock:
Even if the questions are interesting, different approaches of schema creation are unnecessarily mixed here. A documentation of their correlation would be confusing and sometimes beyond the control of Spring Boot itself:

> Both (1) and (2) exist to enable/disable schema generation. I would like to know which of both has precedence.

[...]

PR #15859, opened 2019-02-06
Schema generation approach notes

matthiasstock:
Fixes gh-11721

Instead of trying to answer all specific questions, I'm trying to resolve the issue by giving a general advice in the documentation, see https://github.com/spring-projects/spring-boot/issues/11721#issuecomment-459536396


snicoll:
@matthiasstock thank you for your contribution. I don't believe that this addresses #11721 though so we'll have to give it another look.

Issue #10712, opened 2017-10-20
Add null-safety annotations on Spring Boot APIs


sdeleuze:
Like Spring Data did, it would be nice for both Java and Kotlin developers to leverage in Spring Boot 2 [null-safety annotations](https://docs.spring.io/spring-framework/docs/current/spring-framework-reference/core.html#null-safety) (`@Nullable` etc) in most common Spring Boot APIs like `SpringApplication` and `TestRestTemplate`.

These annotations are now flexible enough to just target the API of these 2 classes if we want (annotating all Spring Boot would be nice but we probably don't have the time for that).

I can contribute such feature via a PR, I would mainly need guidance on what are the most common APIs used in addition to the 2 proposed.

wilkinsona:
I think it may do more harm than good to only tackle this for the most common APIs as the developer experience would be a bit inconsistent. While it may well be a lot of work, my vote is that we try to tackle this in its entirety for 2.0 or defer it in its entirety until 2.1.

sdeleuze:
Given Framework 5.0 and Data Kay support, and the fact that this is considered by many as the most important feature of our Kotlin support, I can probably contribute a PR at `2.0.0.RC1` time that you could review and polish for integrating in `2.0.0.RC2`.

wilkinsona:
Thank you, @sdeleuze. I'd prefer that if we do this we get this into RC1 so that we have plenty of time for feedback and to get it right. Let's see what we can manage.

philwebb:
We think it's unlikely that we'll get to this in 2.2. In terms of effort vs reward we'd rather focus on immutable configuration properties for 2.2.

philwebb:
There are a few build tools that we should also look at:
- https://github.com/uber/NullAway
- https://checkerframework.org/
- https://fbinfer.com/docs/eradicate.html

wilkinsona:
This has now be superseded by #46587 and #46926.

PR #15876, opened 2019-02-08
Remove `Nullable` annotation

wonwoo:


wilkinsona:
Thanks for the PR. The intent of the proposed changes isn't clear to me. Can you please describe what you are aiming to harmonise? Note that we have an [issue](
https://github.com/spring-projects/spring-boot/issues/10712) open to add `@Nullable` across Spring Boot's API. It is something that we want to do, but we want to do it in one go across the entire API.

wonwoo:
nullable annotation is declared another convert method. in harmony with them..

But if you change the entire api at once, this pr is closed.

wilkinsona:
Thanks for the clarification. The existing usages of `@Nullable` on a `Converter` are only in some test code. I'm also not sure why it's there and think I could probably be removed.

@snicoll, can you recall why we have `@Nullable` here:

https://github.com/spring-projects/spring-boot/blob/e8fef9733520a7bd40813931c311a7cc46075f2d/spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/reactive/webclient/ExampleIdConverter.java#L42-L50

snicoll:
It is probably a mistake, IntelliJ IDEA can add the annotation automatically when you override a method.

wonwoo:
https://github.com/spring-projects/spring-boot/blob/ca0de4385c01cbbd120e96ec9f83427fb48c9a15/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/ArrayToDelimitedStringConverter.java#L53-L59

`ArrayToDelimitedStringConverter`, `DelimitedStringToArrayConverter`, `DelimitedStringToCollectionConverter` 

Nullable Annotations are declared in the class

wilkinsona:
Well spotted. They shouldn't be there in either (and are providing no benefit as it's a package-private class). Would you like to repurpose this PR to remove them? The only place where `@Nullable` should be used is to allow actuator endpoint operations to indicate that an input is optional.

wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @wonwoo. The proposed changes have been merged into 2.1.x and forwards into master.

PR #15883, opened 2019-02-09
Fix broken link to Appendix

tim-toady:
Fixed incorrect link.


snicoll:
@tim-toady thank you very much for making your first contribution to Spring Boot. There are actually a number of additional broken links so I've created a separate issue for it, see #15884

PR #15885, opened 2019-02-09
Clean up SpringPropertyAction.getValue()

dreis2211:
Hi,

while looking through the code of `SpringPropertyAction` I noticed a String concatenation like this: `source.substring(0, lastDot + 1) + source.substring(lastDot + 1)`. This should be effectively the same as using `source` directly. By doing so we can cleanup the code a bit.

Let me know what you think.
Cheers,
Christoph


wilkinsona:
Thanks once again, @dreis2211. The proposed changes are now in master.

Issue #6370, opened 2016-07-11
Move sample tests to real integration tests


snicoll:
This is a follow-up of  #6357 

Lots of samples are in between samples to demonstrate features to newcomers and integration tests. This has the unfortunate side effect to make some sample much more complex than they should be. The purpose of this issue is to revisit our samples and move such code that a dedicated integration tests infrastructure.

philwebb:
The following are candidates (edit this comment to add more)

- [ ] spring-boot-sample-test-nomockito

jmcshane:
I'd like to help with this, as I use spring boot fairly heavily and would like to help improve the integration tests.  I'm building up PR #8405 with some of the modules listed above by @philwebb.  Feel free to add comments on the PR

philwebb:
@jmcshane Cool. Thanks.

snicoll:
This one is superseded now that we've moved all of them to smoke tests, see #17197.

PR #15886, opened 2019-02-10
add webflux sample to text-stream post stream

kcwebapply:
Hello. 

I see the sample code of `spring-boot-sample-webflux` project and think that it 'd be better for adding post-stream api on sample code.

I changed get `echo` router to return application value every second by requesting like below. 
`curl -v localhost:8080/echo  -H 'Accept: text/event-stream'`

Also, added post  `echo` router for changing application value (after doing this, value from stream changed).  example are like below.
`curl -v localhost:8080/echo  -H 'Accept: text/event-stream' -X POST -d "ok" -H "Content-Type:application/json"` .

I think that  user can understand what `webflux`  is. 

This is only the suggestion and my opinion, but example code should include data-stream pattern because it is one of the most import function of `webflux`. 

Thank you, best regards.

### DIFF
--- a/spring-boot-samples/spring-boot-sample-webflux/src/main/java/sample/webflux/EchoHandler.java
+++ b/spring-boot-samples/spring-boot-sample-webflux/src/main/java/sample/webflux/EchoHandler.java
@@ -16,8 +16,13 @@
 
 package sample.webflux;
 
+import java.time.Duration;
+import java.util.stream.Stream;
+
+import reactor.core.publisher.Flux;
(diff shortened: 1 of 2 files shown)


bclozel:
We're in a strange situation where half of the samples are being used as integration tests, and others are really meant to show particular setups for Spring Boot applications.

I'm declining this PR, our goal is not to showcase possibilities nor common use cases for application types. We're already considering pruning the samples in #6370.

Thanks for your contribution!

PR #15888, opened 2019-02-10
spring-boot-starters for Embedded Cassandra and @DataCassandraTest.

nosan:
Hi there,
This PR adds links to:
 - [Cassandra Spring Boot Starter](https://github.com/nosan/embedded-cassandra-spring-boot) for [Embedded Cassandra Project](https://github.com/nosan/embedded-cassandra)
- [@DataCassandraTest Spring Boot Starter](https://github.com/nosan/cassandra-data-test-spring-boot)

P.S. as we discussed [here](https://github.com/spring-projects/spring-boot/pull/13033#issuecomment-400430466) with @philwebb, `@DataCassandraTest`, should be separate, at least until the library becomes established.

Please, let me know if you have willing to integrate `@DataCassandraTest`  in the future, I will do my best for helping.

Thanks in advance

### DIFF
--- a/spring-boot-project/spring-boot-starters/README.adoc
+++ b/spring-boot-project/spring-boot-starters/README.adoc
@@ -187,4 +187,10 @@ do as they were designed before this was clarified.
 | http://alexo.github.io/wro4j/[Wro4j]
 | https://github.com/michael-simons/wro4j-spring-boot-starter
 
+| https://github.com/nosan/embedded-cassandra[Embedded Cassandra]
+| https://github.com/nosan/embedded-cassandra-spring-boot
+
+| `@DataCassandraTest` annotation to test `Cassandra` components
+| https://github.com/nosan/cassandra-data-test-spring-boot
+
 |===


snicoll:
@nosan thank you and congrats on the release. The second link is a bit unusual and I am wondering why you wouldn't want to reverse things a bit? The whole point here (as far as I understood) is to test Cassandra-based applications using an embedded server.

Why don't you have a Cassandra test starter of some kind that brings the embedded module? That way you'd have only one starter to be added in test scope and it would bring `@DataCassandraTest` support. Something like `cassandra-test-spring-boot-starter`?

nosan:
@snicoll thank you, `@DataCassandraTest` can be used not only with [Embedded Cassandra Project](https://github.com/nosan/embedded-cassandra), but with `cassandra-unit`, `test containers` or with `Cassandra database` itself. 
Here is a quick teaser (_JUnit5_): 
```java
    import org.springframework.beans.factory.annotation.Autowired;
    import com.github.nosan.boot.test.autoconfigure.data.cassandra.DataCassandraTest;
    
    @DataCassandraTest
    public class ExampleDataCassandraTests {
    
        @Autowired
        private YourRepository repository;
    
    }
```
As you can mention, I did not use `@EmbeddedCassandra` annotation here.

**P.S**.
I had an idea to include `com.github.nosan:embedded-cassandra-test` into the `cassandra-data-test-spring-boot-starter`, but then,  I would force folks to use  [Embedded Cassandra Project](https://github.com/nosan/embedded-cassandra) instead of other projects.

snicoll:
@nosan that is fine. The only thing I am after really is one starter that brings `@DataCassandraTest` and the embedded server so that both projects are promoted. 

I agree those are separate but, in the context of Spring Boot, if you decide to use the embedded server, it makes sense to get `@DataCassandraTest` support as well.

In the context of this page (promoting Spring Boot starters), I don't think two entries are warranted.

nosan:
@snicoll, seems like you have not seen my message on the `Gitter`.

I think, I should provide more information about these projects.

There are 3 projects:

- [Embedded Cassandra Project](https://github.com/nosan/embedded-cassandra/wiki) this is a core project to start and stop  `Apache Cassandra`.

- [Embedded Cassandra Auto-Configuration Spring Boot Starter](https://github.com/nosan/embedded-cassandra-spring-boot) this project offers auto-configuration for `Embedded Cassandra`. 

- [Data Cassandra Spring Spring Boot Starter](https://github.com/nosan/cassandra-data-test-spring-boot) this project provides `@DataCassandraTest` annotation for testing **Cassandra Components**

`@DataCassandraTest` does not rely on **Embedded Cassandra Spring Boot Starter** and even more on **Embedded Cassandra Project** itself, so folks can use this annotation to run tests against `Cassandra container` or something else...

Summing up, I don't understand why `@DataCassandraTest` starter should have a compile dependency either on `com.github.nosan:embedded-cassandra-test` or `com.github.nosan:embedded-cassandra-spring-boot-starter`starters. 

Seems like I quite not get what you mean, could you please clarify?

Thanks in advance.

nosan:
Btw, May I use `local.cassandra.port`, `local.cassandra.address` properties or they are reserved?

nosan:
as we discussed with @snicoll on the `Gitter`, they should not be done as `starters`.

Issue #15879, opened 2019-02-08
Auto-configured OpenEntityManagerInViewInterceptor does not back off when an OpenEntityManagerInViewFilter is only defined as a registration bean


wilkinsona:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before, or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Background

Spring Boot auto-configures an `OpenEntityManagerInViewInterceptor` in `JpaWebConfiguration`. When the user provides their own equivalent bean, the auto-configured interceptor should back off.

## Problem

The auto-configuration is conditional on there being no `OpenEntityManagerInViewInterceptor` or `OpenEntityManagerInViewFilter`:

https://github.com/spring-projects/spring-boot/blob/99656b9dd8e26b0741d55e9937d8c1aa8aa751a9/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/JpaBaseConfiguration.java#L212-L213

This means that it correctly backs off when the user defines a bean of either type, however it does not back off if the user defines a `FilterRegistrationBean<OpenEntityManagerInViewFilter>` bean unless they also define an `OpenEntityManagerInViewFilter` bean. Defining only a `FilterRegistrationBean<OpenEntityManagerInViewFilter>` bean should be sufficient to cause the auto-configuration of the interceptor to back off.

## Solution

Rather than considering `OpenEntityManagerInViewFilter` in the `@ConditionalOnMissingBean`, `@ConditionalOnMissingFilterBean(OpenEntityManagerInViewFilter.class)` should be used instead. This will cause the auto-configured filter to back off when there is a user-defined `OpenEntityManagerInViewFilter` or `FilterRegistrationBean<OpenEntityManagerInViewFilter>` bean.

There is an existing test that checks the `OpenEntityManagerInViewFilter` case:

[...]

PR #15889, opened 2019-02-10
Auto-configured OpenEntityManagerInViewInterceptor does not back off when an OpenEntityManagerInViewFilter is only defined as a registration bean

acemoo:
Make auto-configured OpenEntityManagerInViewFilter backoff when defined as registration bean. Fixes GH-15879.

I also added a unit test for when a OpenEntityManagerInViewInterceptor has been manually created.


wilkinsona:
@acemoo Thank you very much for picking up a first-timers-only issue and then making your first contribution to Spring Boot. The proposed changes have been merged into 2.1.x and forwards into master. We hope to see you again soon with another contribution.

Issue #15808, opened 2019-01-30
Unable to exclude dependencies on repackaging war


rvijayarajan:
I'm using spring-boot-maven-plugin to repackage my archive into a bootable and deployable WAR file. Doing so, I had to exclude few dependencies (transitive in my case). I tried excluding using exclusion configuration as mentioned in plugin documentation (https://docs.spring.io/spring-boot/docs/2.1.2.RELEASE/maven-plugin/examples/exclude-dependency.html). But it doesn't works. The dependencies I configure to exclude appears in the generated WAR file.

Plugin version: 2.1.2.RELEASE

Unfortunately I don't have a code snippet to post. I'll try to add later. Is this a known issue ?

snicoll:
@rvijayarajan "it does not work" is not very helpful. As you've figured out yourself, the link describes what should be done to exclude a dependency. If that does not work for you we'll need a sample that we can run ourselves.

rvijayarajan:
@snicoll I've created a sample project with my use case and pushed the same into github. You can checkout this code and do a "mvn clean install" to experience the case I've mentioned.

**Github sample project URL:** https://github.com/rvijayarajan/spring-boot-maven-plugin-exclude-test

In this sample project I've excluded a dependency with group and artifactId namely org.webjars and bootstrap respectively. But I see that in the generated spring-boot.war, the excluded dependency is still present. 

Please let me know if I have to provide more details. Thanks.

snicoll:
Thank you for the sample. Have you noticed the repackage goal is invoked twice? Please [check the release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#maven-plugin).

Regardless of this, this works fine with `jar` packaging but does not for a war so we'll have to investigate what's wrong there.

andrejleitner:
Exclusions don't work for WAR packaging also in 2.0.8.RELEASE.

rvijayarajan:
Thanks @snicoll.

nosan:
@snicoll I have an idea how to fix this bug, don't you mind if I take this issue?

snicoll:
Thanks for the offer. I haven't had the time to look at it yet so I am happy to review a PR.

snicoll:
Closing in favour of PR #15947

logicatmidod:
Hi I am new to the contributors list trying to learn. Saw this is reopened again is there any reason why ?

snicoll:
[...]

PR #15947, opened 2019-02-13
Exclude dependencies during repackaging process for the war archive

nosan:
When `maven-war-plugin` builds a war file, it includes all artifact dependencies and put them into `WEB-INF/lib` folder, but `maven-jar-plugin` does not.
That is why `excluding` works perfectly for `jar`, but does not work for `war`.

This commit provides a new method `Layout.getLibrariesLocation()` that returns the location of the libraries within the archive.
The `Jar Entries` which are located in this `Libraries Location` should be filtered out, as the new libraries will be added at the end of the `repackaging` process.

fixes gh-15808


spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Layout.java
```diff
@@ -49,6 +49,12 @@
 	 */
 	String getClassesLocation();
 
+	/**
+	 * Returns the location of the libraries within the archive.
+	 * @return the libraries location
+	 */
+	String getLibrariesLocation();
```
wilkinsona: There's a lot of overlap with `getLibraryDestination(String, LibraryScope)`. Without the `LibraryScope`, this method will return the wrong location for a provided artifact in a war – it will return `WEB-INF/lib` when it should return `WEB-INF/lib-provided`. I wonder if the proposed solution can be reworked to use `getLibraryDestination(String, LibraryScope)` instead?

nosan:
Thanks, 
I spent a lot of time to find a way to rework this to use `getLibraryDestination(String, LibraryScope)`, 
and telling the truth I don't know how it can be done at the moment. 

> this method will return the wrong location for a provided artifact in a war – it will return WEB-INF/lib when it should return WEB-INF/lib-provided.

`maven-war-plugin` does not include `provided` dependencies.


Here is a small background why excluding is not working for WAR.

`RepackageMojo`: 
```java

	private void repackage() throws MojoExecutionException {
	   // (1)
		Artifact source = getSourceArtifact();
		File target = getTargetFile();
		Repackager repackager = getRepackager(source.getFile());
		// (2)
		Set<Artifact> artifacts = filterDependencies(this.project.getArtifacts(), getFilters(getAdditionalFilters()));
		//(3)
		Libraries libraries = new ArtifactsLibraries(artifacts, this.requiresUnpack, getLog()); 
		try {
			LaunchScript launchScript = getLaunchScript();
			repackager.repackage(target, libraries, launchScript);
		}
		catch (IOException ex) {
			throw new MojoExecutionException(ex.getMessage(), ex);
		}
		updateArtifact(source, target, repackager.getBackupFile());
	}

```

1. a source `war` has already had `WEB-INF/lib` folder which has been added by `maven-war-plugin` before.
2. `RepackageMojo` filters all dependencies using the filters. (such as `excludeGroupIds`, )
3. `RepackageMojo` creates `ArtifactsLibraries` which includes only filtered libraries.


then, `Repackager` tries to repackage the `sourceJar` using the given libraries and other parameters:

 `Repackager`:
 ```java
private void repackage(JarFile sourceJar, File destination, Libraries libraries, LaunchScript launchScript)
		throws IOException {
	
	
	// (1)
	WritableLibraries writeableLibraries = new WritableLibraries(libraries);
	try (JarWriter writer = new JarWriter(destination, launchScript)) {
		writer.writeManifest(buildManifest(sourceJar));
		writeLoaderClasses(writer);
		if (this.layout instanceof RepackagingLayout) {
		// (2)
			writer.writeEntries(sourceJar,
					new RenamingEntryTransformer(((RepackagingLayout) this.layout).getRepackagedClassesLocation()), writeableLibraries);
		}
		else {
		// (2)
			writer.writeEntries(sourceJar, writeableLibraries);
		}
		
		
		
		// (3) 
		
		writeableLibraries.write(writer);
	}
}
 
 ```
 
 1. Source `jar` has already had `WEB-INF/lib` folder with dependencies, as `maven-war-plugin` includes all `compile` dependencies and put them beneath `WEB-INF/lib` folder.
 2. `JarWriter` adds all jar entries from `sourceJar` into the `destination`. (including `WEB-INF/lib`)
 3. Write libraries
 
 So, summarizing the above I can not understand how `getLibraryDestination(String, LibraryScope)` can help here.
 In my opinion, `JarWriter` should avoid any libraries, because for this purpose `WritableLibraries` is responsible. Currently, there is no way to exclude/filter jar entries during the `JarWriter.writeEntries` process.

wilkinsona:
> `maven-war-plugin` does not include provided dependencies.

Unfortunately, I don't think that helps here. I really don't want to add a public API that only gives the right answer when used in one particular scenario.

It sounds like we need to find a different approach to solving this one. Thanks anyway for your efforts.

PR #15950, opened 2019-02-14
Stop using imports from shaded dependencies

michael-simons:
I noticed that some tests use dependencies that come in as shaded dependencies from test containers. I think you should use the direct dependencies.


snicoll:
Good catch, thanks @michael-simons!

snicoll:
I've updated our checkstyle rule in 10f58b3 so that it doesn't happen again.

PR #15951, opened 2019-02-14
Add support for setting the awaitTeminationSeconds and waitForTasksToCompleteOnShutdown

filiphr:
to the TaskExecutorBuilder and via the TaskExecutionProperties

I started the migration to Spring Boot 2.1 and I wanted to get rid of our custom executor configuration as we can reuse the one from Boot. Awesome job on adding that in 2.1. I noticed that there are 2 methods that we use and are not exposed. With this PR I am exposing those 2 properties.

Let me know if the target to 2.1.x and the PR in general is acceptable for you. If there is something that I need to change please do let me know.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
@@ -17,13 +17,16 @@
 package org.springframework.boot.autoconfigure.task;
 
 import java.time.Duration;
+import java.time.temporal.ChronoUnit;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.convert.DurationUnit;
 
 /**
  * Configuration properties for task execution.
  *
  * @author Stephane Nicoll
+ * @author Filip Hrisafov
  * @since 2.1.0
  */
 @ConfigurationProperties("spring.task.execution")
@@ -36,6 +39,22 @@ public class TaskExecutionProperties {
 	 */
 	private String threadNamePrefix = "task-";
 
+	/**
+	 * The maximum number of time that the executor is supposed to block on shutdown in
+	 * order to wait for remaining tasks to complete their execution before the rest of
+	 * the container continues to shut down. This is particularly useful if your remaining
+	 * tasks are likely to need access to other resources that are also managed by the
+	 * container. If a duration suffix is not specified, seconds will be used.
+	 */
+	@DurationUnit(ChronoUnit.SECONDS)
+	private Duration awaitTermination;
+
+	/**
+	 * Whether the executor should wait for scheduled tasks to complete on shutdown, not
+	 * interrupting running tasks and executing all tasks in the queue.
+	 */
+	private boolean waitForTasksToCompleteOnShutdown = false;
+
 	public Pool getPool() {
 		return this.pool;
 	}
@@ -46,6 +65,23 @@ public class TaskExecutionProperties {
 
 	public void setThreadNamePrefix(String threadNamePrefix) {
 		this.threadNamePrefix = threadNamePrefix;
+	}
+
+	public Duration getAwaitTermination() {
+		return this.awaitTermination;
+	}
+
+	public void setAwaitTermination(Duration awaitTermination) {
+		this.awaitTermination = awaitTermination;
+	}
+
+	public boolean isWaitForTasksToCompleteOnShutdown() {
+		return this.waitForTasksToCompleteOnShutdown;
+	}
+
+	public void setWaitForTasksToCompleteOnShutdown(
+			boolean waitForTasksToCompleteOnShutdown) {
+		this.waitForTasksToCompleteOnShutdown = waitForTasksToCompleteOnShutdown;
 	}
 
 	public static class Pool {
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
@@ -40,9 +40,8 @@ public class TaskExecutionProperties {
 	private String threadNamePrefix = "task-";
 
 	/**
-	 * The maximum number of time that the executor is supposed to block on shutdown in
-	 * order to wait for remaining tasks to complete their execution before the rest of
-	 * the container continues to shut down. This is particularly useful if your remaining
+	 * Maximum number of time that the executor is supposed to block on shutdown waiting
+	 * for remaining tasks to complete. This is particularly useful if your remaining
 	 * tasks are likely to need access to other resources that are also managed by the
 	 * container. If a duration suffix is not specified, seconds will be used.
 	 */
@@ -50,8 +49,7 @@ public class TaskExecutionProperties {
 	private Duration awaitTermination;
 
 	/**
-	 * Whether the executor should wait for scheduled tasks to complete on shutdown, not
-	 * interrupting running tasks and executing all tasks in the queue.
+	 * Whether the executor should wait for scheduled tasks to complete on shutdown.
 	 */
 	private boolean waitForTasksToCompleteOnShutdown = false;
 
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
```diff
@@ -36,6 +39,22 @@
 	 */
 	private String threadNamePrefix = "task-";
 
+	/**
+	 * The maximum number of time that the executor is supposed to block on shutdown in
```
snicoll: Configuration properties descriptions do not start with "The", "A", etc. The first sentence must be much shorter as it's used by IDEs on auto-completion.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
```diff
@@ -36,6 +39,22 @@
...
+	 * order to wait for remaining tasks to complete their execution before the rest of
+	 * the container continues to shut down. This is particularly useful if your remaining
+	 * tasks are likely to need access to other resources that are also managed by the
+	 * container. If a duration suffix is not specified, seconds will be used.
+	 */
+	@DurationUnit(ChronoUnit.SECONDS)
+	private Duration awaitTermination;
+
+	/**
+	 * Whether the executor should wait for scheduled tasks to complete on shutdown, not
```
snicoll: Ditto. The sentence should stop at the current comma IMO.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/task/TaskExecutionAutoConfigurationTests.java
```diff
@@ -63,13 +63,15 @@
...
-						"spring.task.execution.pool.keep-alive=5s",
-						"spring.task.execution.thread-name-prefix=mytest-")
+		this.contextRunner.withPropertyValues(
+				"spring.task.execution.pool.queue-capacity=10",
+				"spring.task.execution.pool.core-size=2",
+				"spring.task.execution.pool.max-size=4",
+				"spring.task.execution.pool.allow-core-thread-timeout=true",
+				"spring.task.execution.pool.keep-alive=5s",
+				"spring.task.execution.await-termination=30s",
+				"spring.task.execution.wait-for-tasks-to-complete-on-shutdown=true",
```
snicoll: Let's try to keep a consistent ordering in things. Adding those two properties at the end looks ok to me as they are logically grouped. Let's keep the same here and add them after the thread-name-prefix

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/task/TaskExecutionAutoConfigurationTests.java
```diff
@@ -79,6 +81,10 @@ public void taskExecutorBuilderShouldApplyCustomSettings() {
 							.hasFieldOrPropertyWithValue("allowCoreThreadTimeOut", true);
 					assertThat(taskExecutor.getKeepAliveSeconds()).isEqualTo(5);
 					assertThat(taskExecutor.getThreadNamePrefix()).isEqualTo("mytest-");
+					assertThat(ReflectionTestUtils.getField(taskExecutor,
```
snicoll: We don't use `ReflectionTestUtils` for such basic assertion but assertJ (`assertFieldOrPropertyWithValue`).

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-application-properties.adoc
```diff
@@ -175,6 +175,8 @@ content into your application. Rather, pick only the properties that you need.
 	spring.task.execution.pool.max-size= # Maximum allowed number of threads. If tasks are filling up the queue, the pool can expand up to that size to accommodate the load. Ignored if the queue is unbounded.
 	spring.task.execution.pool.queue-capacity= # Queue capacity. An unbounded capacity does not increase the pool and therefore ignores the "max-size" property.
 	spring.task.execution.thread-name-prefix=task- # Prefix to use for the names of newly created threads.
+	spring.task.execution.await-termination= # The maximum number of time that the executor is supposed to block on shutdown in order to wait for remaining tasks to complete their execution before the rest of the container continues to shut down. This is particularly useful if your remaining tasks are likely to need access to other resources that are also managed by the container. If a duration suffix is not specified, seconds will be used.
```
snicoll: Keys are ordered alphabetically and the description is way too long as well.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/task/TaskExecutorBuilder.java
```diff
@@ -158,7 +175,41 @@ public TaskExecutorBuilder keepAlive(Duration keepAlive) {
...
 		return new TaskExecutorBuilder(this.queueCapacity, this.corePoolSize,
 				this.maxPoolSize, this.allowCoreThreadTimeOut, this.keepAlive,
-				threadNamePrefix, this.taskDecorator, this.customizers);
+				threadNamePrefix, this.awaitTermination,
+				this.waitForTasksToCompleteOnShutdown, this.taskDecorator,
+				this.customizers);
+	}
+
+	/**
+	 * /** Set the maximum number of time that the executor is supposed to block on
```
snicoll: Formatting glitch here.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
filiphr: Wasn't aware about that. Will change it to

>The maximum number of time that the executor is supposed to block on shutdown waiting for remaining tasks to complete.

Is this text acceptable?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/task/TaskExecutionAutoConfigurationTests.java
filiphr: I initially wanted not to create a lot of unnecessary changes, but the format already did that 😄. I will change the ordering

filiphr:
>I've added a few suggestions and we'll figure out next where that should land. In theory, `master` is the rule.

Thanks for the review @snicoll. I've applied the fixes you mentioned. If the text is still not satisfiable I can adapt it. I more or less copied that from the `ExecutorConfigurationSupport`.

The reason why I did it against `2.1.x` was that I was hoping it might go in it so I can remove more from our code. I also saw that there were a lot of commits on `2.1.x` and which were then merged to `master`.

Feel free to squash everything in one commit, if you want me to do that let me know

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
snicoll: No it is isn't. It starts with `The` and I've indicated previously we're not documenting keys that way.

On a second thought those two keys are related to shutdown so it might be interesting to crete a sub group for them the same way we did for pool

snicoll:
@filiphr thanks but I believe the workaround for now is easy enough. You can inject the builder, apply it and then set the two properties you need. It doesn't remove your custom arrangement but it allows to configure the 2.1 properties at least.

I need to give a bit this a bit more thought. As I've indicated, those properties are related to shutdown so I am quite keen to group them, especially as the current form is too long IMO. 

I'll pick up from here.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionProperties.java
filiphr: Sorry, I actually removed `The` in my commit. Yes I'll move them into a shutdown sub group

filiphr:
Yes the current workaround works perfectly.

>I need to give a bit this a bit more thought. As I've indicated, those properties are related to shutdown so I am quite keen to group them, especially as the current form is too long IMO.
>
>I'll pick up from here.

No problem, if there is something that I can do please let me know.

snicoll:
@filiphr thanks for the contribution. I've polished it and also added support for TaskScheduling as they share a common feature there.

filiphr:
Thanks for merging the PR and for extending on it

Issue #15930, opened 2019-02-12
Consider switching from FileSize to DataSize for logback size-based properties


snicoll:
Follow-up of #15325

Our `logging.file` namespace has two properties that use `FileSize` behind the scenes. That object is very similar to our new `DataSize` support so it would be nice to expose `DataSize` based properties for consistency.

Unfortunately, [the format is not 100% the same](https://github.com/qos-ch/logback/blob/4dd08e68321548cd85251ec389f5a832505e1268/logback-core/src/main/java/ch/qos/logback/core/util/FileSize.java#L23-L27) so we could keep binding to `String` and do some magic in the code? Or simply move to `DataSize`.

nosan:
@snicoll,
I'm ready to start working on this issue, but, I have a few questions:
- Should I change the type of these properties? (`String` -> `DataSize`)
- Should I add support for both formats? 

Thanks in advance

snicoll:
The purpose is to support `DataSize` and pass a long bytes to Logback wherever we can. I can't tell you as I haven't investigated yet.

>Unfortunately, [the format is not 100% the same](https://github.com/qos-ch/logback/blob/4dd08e68321548cd85251ec389f5a832505e1268/logback-core/src/main/java/ch/qos/logback/core/util/FileSize.java#L23-L27) so we could keep binding to `String` and do some magic in the code? Or simply move to `DataSize`.

The preference is to only care about `DataSize` but we should investigate first what happens if we use the old (String) format with it

PR #15957, opened 2019-02-14
Switching from FileSize to DataSize for logback size-based properties.

nosan:
see gh-15930.
`DataSize` can contain a `negative` value, Should we handle this? If yes, should we throw an exception or use a default value?
Thanks

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/DefaultLogbackConfiguration.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/DefaultLogbackConfiguration.java
@@ -17,6 +17,9 @@
 package org.springframework.boot.logging.logback;
 
 import java.lang.reflect.Method;
+import java.util.Objects;
+import java.util.regex.Matcher;
+import java.util.regex.Pattern;
 
 import ch.qos.logback.classic.Level;
 import ch.qos.logback.classic.encoder.PatternLayoutEncoder;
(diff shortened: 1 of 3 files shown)


snicoll:
@nosan thanks for the PR. Did you check if the `FileSize` string format works transparently with `DataSize`?

nosan:
Thanks @snicoll,
I have added some tests to check this [here](https://github.com/spring-projects/spring-boot/pull/15957/files#diff-cc08d614c5d3cabca80135c77e110313R401) and [here](https://github.com/spring-projects/spring-boot/pull/15957/files#diff-cc08d614c5d3cabca80135c77e110313R491).

Current implementation parses the value via `DataSize` class, if something wrong, then it tries to parse a value using `FileSize` class.

**P.S.** one case that which is not covered is a `negative` value. I think the exception should be thrown as a value is not valid.

nosan:
FYI @snicoll,
I changed the implementation this morning and added more tests.  Instead of `stupid` parsing via both classes, I added a regex to handle both formats.

snicoll:
@nosan than you for the PR but I've decided to go with a different approach, see https://github.com/spring-projects/spring-boot/commit/eee07efee1bb6a9bdfd6e34000eb91b7cf347502 for more details.

nosan:
@snicoll sure, no problem.
BTW,  just to make sure, you did nothing with a negative value, is it ok?

snicoll:
I noticed your comment about that. Yes I think it is ok as the main purpose was to align tooling and size format

Issue #15835, opened 2019-02-02
Does not work 'classpath*:' for Resource[] property on configuration properties since 2.x


kazuki43zoo:
I encountered a un-expected behavior on configuration properties feature since Spring Boot 2.x. 
On 1.5.x, it work fine.

### Conditions

* Define a property of `Resource[]` on custom configuration properties class
* Specify a property value using `classpath*:` prefix such as `classpath*:files/*.txt`

### Expected result

* Binding multiple `Resource` instance that matches pattern string

### Actual result

* Binding a single `Resource` that holds a specify pattern string such as `classpath*:files/*.txt`

###  Repro test

[demo.zip](https://github.com/spring-projects/spring-boot/files/2823844/demo.zip)

```java
package com.example.demoresources;

import org.assertj.core.api.Assertions;
import org.junit.BeforeClass;
import org.junit.Test;
import org.junit.runner.RunWith;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.boot.context.properties.ConfigurationProperties;
import org.springframework.boot.context.properties.EnableConfigurationProperties;
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.core.io.Resource;
import org.springframework.test.context.junit4.SpringRunner;

import java.io.IOException;
import java.nio.file.Files;
import java.nio.file.Path;
import java.nio.file.Paths;
import java.util.Arrays;
import java.util.List;
import java.util.stream.Collectors;

@RunWith(SpringRunner.class)
@SpringBootTest(properties = "my.files=classpath*:files/*.txt")
@EnableConfigurationProperties({DemoResourcesApplicationTests.MyProperties.class})
public class DemoResourcesApplicationTests {

  @Autowired
  MyProperties myProperties;

  @BeforeClass
  public static void setup() throws IOException {
    Path dir = Paths.get("target", "test-classes", "files");
    Files.createDirectories(dir);
    createFileIfNotExist(dir.resolve("a.txt"));
    createFileIfNotExist(dir.resolve("b.txt"));
  }

  private static void createFileIfNotExist(Path path) throws IOException {
    if (!path.toFile().exists()) {
      Files.createFile(path);
    }
  }

  @Test
  public void contextLoads() {
    Resource[] files = myProperties.getFiles();
    List<String> fileNames = Arrays.stream(files).map(Resource::getFilename).collect(Collectors.toList());
    Assertions.assertThat(fileNames)
        .hasSize(2)
[...]

PR #15958, opened 2019-02-14
Binding multiple `Resource` instance that matches pattern string.

nosan:
see gh-15835


wilkinsona:
Thanks for the PR. That's the "obvious" fix (and the one I would have implemented based on my analysis if I was sure it was the right thing to do). I think we need some input from @mbhave and @philwebb here.

wilkinsona:
Thanks for the PR. We discussed this today and think we need something that is a little less specific to the `String` to `Resource[]` conversion as other Editor-based conversions won't work either. Thanks anyway.

Issue #15701, opened 2019-01-14
Add properties for setting the elasticsearch high level rest client timeouts


cbm-afoulfoin:
Hi, 
There are 2 ES clients supported by spring boot autoconfiguration: 
- Jest 
- The official ES RestHighLevelClient

For configuring Jest (`org.springframework.boot.autoconfigure.elasticsearch.jest.JestProperties`), we can set the properties `connectionTimeout` and `readTimeout`.
For configuring the ES RestHighLevelClient (`org.springframework.boot.autoconfigure.elasticsearch.rest.RestClientProperties`), we can only set the uris, username and password, but not the timeouts that can only be set by a `RestClientBuilderCustomizer`.

Since timeouts are common properties that often need to be customized, it would be great to be able to configure them the same way spring boot is doing for Jest, via properties only (under `spring.elasticsearch.rest.*`).

Thanks!

bclozel:
Superseded by #15965

PR #15965, opened 2019-02-15
connectionTimeout and readTimeout properties for configuring RestClient

nosan:
This PR adds `connectionTimeout` and `readTimeout` properties for configuring Elasticsearch `RestClient`.

gh-15701

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/rest/RestClientProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/rest/RestClientProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.autoconfigure.elasticsearch.rest;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.List;
@@ -47,6 +48,16 @@ public class RestClientProperties {
 	 */
 	private String password;
 
+	/**
+	 * Connection timeout.
+	 */
+	private Duration connectionTimeout = Duration.ofSeconds(1);
+
+	/**
+	 * Read timeout.
+	 */
+	private Duration readTimeout = Duration.ofSeconds(30);
+
 	public List<String> getUris() {
 		return this.uris;
 	}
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/rest/RestClientProperties.java
```diff
@@ -47,6 +48,16 @@
 	 */
 	private String password;
 
+	/**
+	 * Connection timeout.
+	 */
+	private Duration connectionTimeout = Duration.ofSeconds(1);
```
snicoll: Where is this coming from? The default for the client is `-1` ?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/rest/RestClientProperties.java
```diff
@@ -47,6 +48,16 @@
...
+
+	/**
+	 * Read timeout.
+	 */
+	private Duration readTimeout = Duration.ofSeconds(30);
```
snicoll: Idem. -1 states "undefined" so I'd be curious to understand where are those 30 coming from?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/elasticsearch/rest/RestClientAutoConfigurationTests.java
```diff
@@ -74,6 +76,31 @@ public void configureWhenBuilderCustomizerShouldApply() {
...
 	}
 
+	@Test
+	public void defaultTimeoutsShouldBeConfigured() {
+		this.contextRunner.run((context) -> {
+			assertThat(context).hasSingleBean(RestClient.class);
+			assertThat(context).hasSingleBean(RestClientProperties.class);
+			RestClient restClient = context.getBean(RestClient.class);
+			RestClientProperties properties = context.getBean(RestClientProperties.class);
+			assertTimeouts(restClient, properties.getReadTimeout(),
```
snicoll: That's not what that test is supposed to do. It should test that when no defaults are set, the defaults in the properties object are consistent. 

So you should configure a `RestClient` manually (i.e. without the auto-config) and query its timeouts and validate they are the same as the one defined in the configuration by default to ensure our defaults match the ones of ES if that's the plan.

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -4807,6 +4807,8 @@ You can further tune how `RestClient` is configured, as shown in the following e
 	spring.elasticsearch.rest.uris=https://search.example.com:9200
 	spring.elasticsearch.rest.username=user
 	spring.elasticsearch.rest.password=secret
+	spring.elasticsearch.rest.connection-timeout=10s
```
snicoll: I don't think that's necessary to list all properties. The `jest` doc has a reference to `read-timeout` so let's align to that only.

nosan:
@snicoll 
Thank you very much.
The default values I took from `org.elasticsearch.client.RestClientBuilder`
```java
    public static final int DEFAULT_CONNECT_TIMEOUT_MILLIS = 1000;
    public static final int DEFAULT_SOCKET_TIMEOUT_MILLIS = 30000;

```
I've updated the PR.

snicoll:
Thanks for the feedback @nosan. This is now merged with a [polish commit](https://github.com/spring-projects/spring-boot/commit/2cfcd2690eb22dfc59b979e3a74b8ea858e3b94e).

Issue #15480, opened 2018-12-17
Suggestion to support LazyConnectionDataSourceProxy


ttddyy:
Hi,

I would like to suggest a support for `LazyConnectionDataSourceProxy` in spring-boot.

Here is my opinion:

Transaction management in spring:
When a transactional method(`@Transactional`) is called, the configured impl of `PlatformTransactionManager` kicks in and starts/prepare transaction logic(call  `AbstractPlatformTransactionManager#doBegin`).

In the implementation of starting transaction(e.g: `JpaTransactionManager`, `HibernateTransactionManager`), when `@Transactional` definition has non default isolation level or readonly flag is set to true, it acquires a physical connection and set isolation level and/or readonly flag to the connection in its preparation phase. (call to `DataSourceUtils#prepareConnectionForTransaction`) 
Since this happens before invoking the actual target method, this consumes a connection from pool(underlying `DataSource`) regardless of the method really needs a connection or not.

It is problematic especially on high traffic application with:
- Many readonly transactional methods (or any non-default transactional methods)
  - This holds connection longer than it needs to do
- High hits for hibernate 2nd level cache
  - The method may not even need a connection

It is better to delay getting the physical connection as much as possible.

`LazyConnectionDataSourceProxy` in spring solves this problem. When isolation-level, read-only flag, etc methods are called, this proxy internally keeps those values. Physical connection acquisition is delayed until connection is really required.

Currently, spring-boot doesn't have any support for `LazyConnectionDataSourceProxy`.
My suggestion is to add a configuration of `LazyConnectionDataSourceProxy` in spring-boot.
Probably, application property with `spring.datasource.lazy-connection=true` would wrap the datasource with `LazyConnectionDataSourceProxy`. I also think it is beneficial to turn this on by default.
`LazyConnectionDataSourceProxy` should be recommended more, in general, IMO.

Thanks,

rpsandiford:
HI, all.

While it would be great to have this in place, I was able to write some code to wrap an underlying Data Source in a LazyConnectionDataSourceProxy.  This code specifically handles Hikari, but could be modified for the other data sources Spring Boot uses, or for some other data source all together if wanted.

It's conditional so that you can control it by a property setting ("my.datasource.lacy-connections").  (We have this in a shared library for our own Spring Boot starters, that we use across multiple applications)

And - it should be fairly easy to yank once Spring Boot provides this functionality natively.

[...]

PR #15968, opened 2019-02-15
Adds spring.datasource.lazy-connection property to enable Lazy DataSource support

nosan:
This commit adds `spring.datasource.lazy-connection` property into a `DataSourceProperties` class to enable a `LazyConnectionDataSourceBeanPostProcessor`. `BeanPostProcessor` wraps an auto-configured `DataSource` in a `LazyConnectionDataSourceProxy`. It will be added only if a
`spring.datasource.lazy-connection` property has `true` and a `LazyConnectionDataSourceProxy`
class is present. The default value of the `spring.datasource.lazy-connection` property is **false**, which means a `lazy-connection` feature is disabled by default.

see gh-15480

**Note!** `LazyConnectionDataSourceBeanPostProcessor` it is being invoked after `DataSourceInitializerPostProcessor` (I'm not sure that is the right behaviour)

Thanks in advance.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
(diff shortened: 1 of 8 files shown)


wilkinsona:
Thanks for the PR. I'm not sure that a bean post-processor is the best way to implement this. If there's a second auto-configured DataSource, it will also be wrapped in a `LazyConnectionDataSourceProxy` and I don't think it should be.

I think we'll need to do something that means we can be certain that it's only the auto-configured data source that will be wrapped. We also need to make a decision about whether or not the wrapping should be enabled by default. Let's use #15480 to do that. Thanks anyway.

Issue #15443, opened 2018-12-11
TestRestTemplate.withBasicAuth resets custom requestFactory


larsgrefer:
I'm currently using a `RestTemplateCustomizer` bean to set a custom request factory for all RestTemplate's:

https://github.com/freefair/okhttp-spring-boot/blob/f0f6dcc0ba685c40ebe15de725c1170307027173/autoconfigure/src/main/java/io/freefair/spring/okhttp/OkHttpRestTemplateAutoConfiguration.java#L22-L27

The `TestRestTemplate` I get injected into my unit tests also uses this custom request factory, but calling `withBasicAuth()` returns a new `TestRestTemplate`-instance which no longer contains the custom request factory but a default one.

Environment:
Spring Boot 2.1.1.RELEASE
OpenJDK 11

Similar issues:
see #7812
see #7441
see #7265

mbhave:
I think we should fix this but [this commit](https://github.com/spring-projects/spring-boot/commit/db7268b841b9b90d498822b4ab37981792303126) has some history around why a new request factory is selected when a `TestRestTemplate` with `withBasicAuth` is created.

mbhave:
Closing in favor of PR #15982.

PR #15982, opened 2019-02-18
Do not replace a request factory when creating a new `TestRestTemplate` with `withBasicAuth`

nosan:
See gh-15443

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/web/client/TestRestTemplate.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/web/client/TestRestTemplate.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,6 +26,7 @@ import java.util.HashSet;
 import java.util.List;
 import java.util.Map;
 import java.util.Set;
+import java.util.function.Supplier;
 
 import org.apache.http.client.HttpClient;
 import org.apache.http.client.config.CookieSpecs;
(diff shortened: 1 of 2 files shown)


mbhave:
@nosan Thanks for the PR but I don't think this is how we intended to fix it. We want to avoid using the same `reuqestFactory` instance when using `withBasicAuth`. For more details on why, take a look at the [commit message](https://github.com/spring-projects/spring-boot/commit/db7268b841b9b90d498822b4ab37981792303126) here.

I think what we'd want to do instead is use the supplier to create a new instance of the custom `requestFactory` class and if that fails, fallback to the current behavior. @bclozel WDYT?

bclozel:
I agree with your analysis @mbhave (and sorry about the late reply, that notification got lost in my inbox).

mbhave:
Thanks @bclozel. 

@nosan Do you want to update the PR to do what I've mentioned above?

nosan:
@mbhave 
Sorry for the long delay, sure, I can update the PR. I am not sure, that I understood you correctly, hence, just to make sure, could you please provide more information about your approach? 
Thanks in advance.

bclozel:
@nosan in the current PR, `withBasicAuth` is retrieving the current request factory instance from the embedded `RestTemplate` and reusing it. This line doesn't really change the behavior we're trying to fix.

Several HTTP client libraries are caching the authentication information at the connection pooling level. This means that reusing the same request factory will result in sending the Cookie request header even if the authentication was not meant to be included.

Creating a brand new instance of the configured request factory class (or as Madhura mentioned, falling back to the current behavior in case of problems) should avoid those issues. You can look at the `ClientHttpRequestFactorySupplier` to check how factories are instantiated.

nosan:
@mbhave @bclozel Thanks, 
I have updated the PR, please let me know.

mbhave:
Thanks again, @nosan. This is now on master along with [this additional test](https://github.com/spring-projects/spring-boot/blob/0611eb49c87f367d3e34a1a3bcf04eb70150d180/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/web/client/TestRestTemplateTests.java#L112).

nosan:
@mbhave Thanks!

PR #15993, opened 2019-02-19
Migrate Kotlin tests to Mockk

sdeleuze:
In order to be consistent with Spring Framework 5.2 tests and recommendation to use [Mockk](https://mockk.io/) for Kotlin tests, and in order to be consistent with the upcoming Coroutines pull requests for Spring Data MongoDB and R2DBC which are using Mockk as well, I have migrated Spring Boot tests.


snicoll:
Merged, thank you @sdeleuze !

Issue #16003, opened 2019-02-20
Repackage META-INF/<module-name>.kotlin_module files into BOOT-INF/classes


thorntonrp:
Issue #15890 was addressed to to repackage META-INF/<module-name>.kotlin-module (with a dash), but the the suffix is supposed to be .kotlin_module with an underscore instead of a dash. The dash instead of the underscore was my bad from when I reported the issue. I apologize for the confusion.

Running `jar tf` against a kotlin project reveals that the correct suffix is kotlin_module.

wilkinsona:
Gah, thanks for spotting this @thorntonrp and sorry that I didn't spot it when I implemented the fix. I'll close this one in favour of your PR.

PR #16004, opened 2019-02-20
Fat jar built with Maven does not repackage META-INF/<module-name>.kotlin_module files into BOOT-INF/classes

thorntonrp:
Fixes issue #16003 and adds a maven invoker test to verify that the fix works when there is a kotlin_module in the project.


wilkinsona:
Thanks very much for the PR, @thorntonrp. I've merged the proposed changes into 2.1.x and forwards into master.

PR #16005, opened 2019-02-20
Update instructions on how to use JAXB as XML renderer

larsgrefer:


spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -1167,7 +1167,7 @@ work. To use the Jackson XML renderer, add the following dependency to your proj
 	</dependency>
 ----
 
-If Jackson's XML extension is not available, JAXB (provided by default in the JDK) is
+If Jackson's XML extension is not available, JAXB (provided by default in the JDK 8) is
```
snicoll: Rather than focusing on JDK 8 and now that this is an optional feature, I'd prefer if we rephrase that section along the lines of "if available".

larsgrefer:
Should I rebase the PR against 2.1.x or will the change be cherry-picked into 2.1.x ?

snicoll:
Don't worry about that we can take care of it

snicoll:
@larsgrefer I did requested a change though, are you willing to update the PR?

larsgrefer:
@snicoll I've restructured the section and added more general instructions on how to use JAXB.

Issue #15986, opened 2019-02-18
Debug mode is not logging web and sql related loggers


cldsnchz:
In 2.1, the debug mode (`--debug`) is not logging in debug the `web` and `sql` related loggers, for example: `org.hibernate.SQL`

The issue seems to be in `LoggingApplicationListener`, which is setting in DEBUG the group itself (`web` and `sql`) instead of resolving the loggers in them, as it does for example when logging.level.sql=true is set.

wilkinsona:
Thanks for the report and for the initial analysis. I think you're right. `LoggingApplicationListener.initializeLogLevel(LoggingSystem, LogLevel)`, which is what is called when started with `--debug`, does not take the groups into account so `web` and `sql` are passed as-is rather than being expanded to the logger names that each represents.

wilkinsona:
Closing in favour of #16018.

PR #16018, opened 2019-02-22
Debug mode is not logging web and sql related loggers

nosan:
see gh-15986

Thanks!


snicoll:
Thank you again @nosan. I've merged your contribution to `2.1.x` and `master` with [a polish commit](https://github.com/spring-projects/spring-boot/commit/978f80122b9138f5606bd9e33597807b90129d61) as I didn't like the use of `Optional`.

nosan:
@snicoll  Thanks, you are right. `Optional` is not necessary here at all.

Issue #6719, opened 2016-08-23
Allow RabbitMQ ConnectionFactory to be customized (to set SaslConfig or enable NIO for example)


ellac-sportsbet:
With the auto configure for rabbitmq, currently it is not possible to set SaslConfig. Currently, it always uses the default one: Plain. Could this be enhanced?

wilkinsona:
What would you like to be able to set it to? One of the `DefaultSaslConfig` constants (`PLAIN` or `EXTERNAL`), your own implementation of `SaslConfig`, something else?

ellachen1234:
Thanks for your quick response. As of now, I am only looking for setting it as one of the DefaultSaslConfig(more precisely, in some cases, I would like to set it to EXTERNAL). But customized implementation might be beneficial to other people as well?

jimbo1007:
We would need to setSaslConfig to DefaultSaslConfig.EXTERNAL in order to be able to do client certificate authentication with RabittMQ. At the moment this is preventing us from using spring-boot

ellachen1234:
@jimbo1007  Not sure whether you have tried a work around like below in configuration class:

    @PostConstruct
    public void init() {
        if (rabbitProperties.getSsl().isEnabled() && rabbitProperties.getSsl().getKeyStore() != null) {
            cachingConnectionFactory.getRabbitConnectionFactory().setSaslConfig(DefaultSaslConfig.EXTERNAL);
        }
    }

jimbo1007:
I tried your workaround and it is working.
Thank you for the help

ayudovin:
@wilkinsona,  As far as I understand, you want to create `RabbitConnectionFactoryBeanCustomizer` interface. Implementations of this interface will be injected in `RabbitAutoConfiguration` and will customize `RabbitConnectionFactoryBean`, is it correct?

wilkinsona:
I'm not sure yet. This issue is assigned to 2.x so we've yet to look at it in detail. I think that some sort of customizer makes sense, but it's complicated a little here as the `RabbitConnectionFactoryBean` isn't exposed as a bean and is really an implementation detail. Therefore, I'm not sure that we want its use to become apparent via a customizer API.

There are a few options that I can see:

1. Customise the `RabbitConnectionFactoryBean`
2. Customise the `ConnectionFactory` that the factory bean creates
3. Customise the `CachingConnectionFactory` that wraps the `ConnectionFactory`

[...]

PR #16026, opened 2019-02-24
Allow RabbitConnectionFactoryBean to be customized (to set SaslConfig for example)

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/6719)


wilkinsona:
Thanks for the PR. Unfortunately, this isn’t really what we had in mind. As the issue says, being able to configure SASL config is just one example of what we’d like to make possible. The solution should make other customisations of the connection factory bean possible as well. Thanks anyway.

As I’ve said before, I think it would be worthwhile for you to comment on an issue before you work on a fix. That way we can make sure that we’re all in agreement about what needs to be done and can hopefully avoid sole wasted effort.

PR #16037, opened 2019-02-26
Fix deprecations

dreis2211:
Hi,

just noticed two deprecations that we can easily fix.

Cheers,
Christoph


snicoll:
Good catch, thank you @dreis2211!

Issue #16035, opened 2019-02-26
Expose properties to enable conditional logging for Tomcat Access Logs


amcghie:
The Tomcat `AccessLogValve` supports conditional logging via two properties:
* `conditionIf`
* `conditionUnless`

Unfortunately these are not yet supported via the `ServerProperties.Tomcat.AccessLog` configuration class.

wilkinsona:
Thus far, we've taken an incremental approach to configuring Tomcat's AccessLogValve. I think it would be good to use this issue to review the properties that we have, the properties that the valve offers, and decide which of them we want to support.

We currently have the following properties:

- `server.tomcat.accesslog.buffered`
- `server.tomcat.accesslog.directory`
- `server.tomcat.accesslog.enabled`
- `server.tomcat.accesslog.file-date-format`
- `server.tomcat.accesslog.max-days`
- `server.tomcat.accesslog.pattern`
- `server.tomcat.accesslog.prefix`
- `server.tomcat.accesslog.rename-on-rotate`
- `server.tomcat.accesslog.request-attributes-enabled`
- `server.tomcat.accesslog.rotate`
- `server.tomcat.accesslog.suffix`

`AccessLogValve` has the following properties that are not currently exposed as configuration properties:

- `checkExists`
- `condition` (the same as `conditionUnless` and provided for backwards compatibility)
- `conditionIf`
- `conditionUnless`
- `encoding`
- `ipv6Canonical`
- `locale`
- `renameOnRotate`

Tomcat's documentation also lists `maxLogMessageBufferSize` but there is no setter for it. It appears to have been removed in https://github.com/apache/tomcat/commit/92ce68ae96dfac3ea3b3cc87f5618207e13a4005. It's not clear to me if that was intentional and the docs are out of date or if the setter should still be there. /cc @markt-asf

amcghie:
I've got a PR ready to submit with the two properties `conditionIf` and `conditonUnless` I'm happy to include the other properties if you would like. Let me know.

Cheers

wilkinsona:
Thanks for the offer. I've added the team attention label so that we can discuss which properties we want to expose. I'm leaning towards all of them other than `condition`.

markt-asf:
Although part of me thinks if no-one has noticed if 5 years we can probably just hard-code it, that does looks like a bug to me. I'll add the getter/setter back.

wilkinsona:
Superseded by #16039.

PR #16039, opened 2019-02-26
Added support for additional Tomcat access log configuration properties

amcghie:
Added support for the following additional Tomcat Access Log configuration properties:

* `server.tomcat.accessLog.checkExists`
* `server.tomcat.accessLog.conditionIf`
* `server.tomcat.accessLog.conditionUnless`
* `server.tomcat.accessLog.encoding`
* `server.tomcat.accessLog.ipv6Canonical`
* `server.tomcat.accessLog.locale`

Proposed solution for #16035

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -57,6 +57,7 @@ import org.springframework.util.unit.DataSize;
  * @author Olivier Lamy
  * @author Chentao Qu
  * @author Artsiom Yudovin
+ * @author Andrew McGhie
  */
 @ConfigurationProperties(prefix = "server", ignoreUnknownFields = true)
 public class ServerProperties {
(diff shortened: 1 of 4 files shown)


wilkinsona:
This is great, thank you. Was there a reason for not adding a property for `renameOnRotate`?

amcghie:
No worries.  The `renameOnRotate` property was already present.

wilkinsona:
So it was. Sorry. I’ve edited my comment on #16035 to hopefully avoid further confusion.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -614,6 +615,42 @@ public Resource getResource() {
...
+			/**
+			 * Use IPv6 canonical representation format as defined by RFC 5952 in output.
+			 */
+			private boolean ipv6Canonical = false;
+
+			/**
+			 * Set the locale used to format timestamps in log entries and in log file
+			 * name suffix.
+			 */
+			private String locale = Locale.getDefault().toString();
```
snicoll: We can't really do that as the sole purpose of the field init is to express a default value that isn't dependant of the environment. I'll remove that as part of the polish.

snicoll:
@amcghie thank you very much for making your first contribution to Spring Boot. This is now merged in `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/076e327481f5442a00877ac2b140a3cfb9cf131d).

Issue #16044, opened 2019-02-27
NullPointerException with empty X-Forwarded-For header.


luiz290788:
spring-boot: 2.1.3
spring-boot-actuator: 2.1.3

A NullPointerException is thrown in [line 49](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/trace/reactive/ServerWebExchangeTraceableRequest.java#L49) of [ServerWebExchangeTraceableRequest](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/trace/reactive/ServerWebExchangeTraceableRequest.java) because `request.getRemoteAddress().getAddress()` is null when you make a request with an empty X-Forwarded-For header.

You can reproduce it with a request like this
```http
GET http://localhost:8080/some-path
X-Forwarded-For: 
```

The path doesn't matter because the request doesn't reach the routing system.

snicoll:
Closing in favour of PR #16046

PR #16046, opened 2019-02-27
Fix NullPointerException with empty X-Forwarded-For header

nosan:
gh-16044

@snicoll 
I have not found a way how to test this, if you have an idea how to test, please let me know.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/trace/reactive/ServerWebExchangeTraceableRequest.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/trace/reactive/ServerWebExchangeTraceableRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,8 @@
 
 package org.springframework.boot.actuate.web.trace.reactive;
 
+import java.net.InetAddress;
+import java.net.InetSocketAddress;
 import java.net.URI;
 import java.util.LinkedHashMap;
 import java.util.List;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/trace/reactive/ServerWebExchangeTraceableRequest.java
```diff
@@ -45,8 +48,9 @@
 		this.method = request.getMethodValue();
 		this.headers = request.getHeaders();
 		this.uri = request.getURI();
-		this.remoteAddress = (request.getRemoteAddress() != null)
-				? request.getRemoteAddress().getAddress().toString() : null;
+		this.remoteAddress = Optional.ofNullable(request.getRemoteAddress())
```
wilkinsona: We prefer not to use `Optional` for internal `null` checks such as this. If the readability of the ternary is a problem it could be pulled out into a method that's something like this instead:

```java
private static String asString(InetSocketAddress socketAddress) {
	return (socketAddress != null && socketAddress.getAddress() != null)
			? socketAddress.getAddress().toString() : null;
}
```

wilkinsona:
It should be possible to test it using an `InetSocketAddress` created with an unresolvable host.

nosan:
@wilkinsona  Thank you so much!
I have updated this PR.

wilkinsona:
Thanks for the updates, particularly the two new tests. Did you consider adding a new test class for `ServerWebExchangeTraceableRequest` rather than testing things at the level of the filter? I think that might be better as it would allow the test to focus specifically on the handling of an unresolved remote address.

nosan:
@wilkinsona sure, I need some time. I let you know.

nosan:
@wilkinsona 
updated

snicoll:
Thank you @nosan. Please review the [polish commit](https://github.com/spring-projects/spring-boot/commit/c224eebef3064e7b2e88e92fc85a4e9f41e7080c) when you get a chance. We prefer to use `BDDMockito` FYI.

nosan:
@snicoll I did not know, I will keep in mind. 
Thanks

Issue #16032, opened 2019-02-25
Database migrations may not have run before NamedParameterJdbcTemplate is used


xak2000:
This is a variation of already fixed issue with `JdbcTemplate` #13155. But this is for `NamedParameterJdbcTemplate`.

The fix (4881925) doesn't take into account another kind of `JdbcTemplate` - `NamedParameterJdbcTemplate`, which implements `NamedParameterJdbcOperations` (and doesn't implement `JdbcOperations`).

So, if an application doesn't create `JdbcTemplate` bean but only creates `NamedParameterJdbcTemplate` bean (using it's `NamedParameterJdbcTemplate(DataSource dataSource)` constructor), this bean still created before migration tool (e.g. Flyway) runs their migrations. So any `@PostConstruct`, that uses `NamedParameterJdbcTemplate`, runs earlier than migration. What is worse: there is not always the case. Sometimes migrations runs earlier. I think this is more or less "random" order because there is no explicit `@DependsOn` here.

So I think it's need to create a `NamedParameterJdbcOperationsDependsOnPostProcessor` in line with `JdbcOperationsDependsOnPostProcessor` and Flyway/Liqbase childs of it.

Spring Boot version tested: 2.0.5. But I don't see any `NamedParameterJdbcOperations` postprocessors in master, so I think this bug affects current version too.

clevertension:
@snicoll are you working to fix this bug? if not, i can provide a PR ASAP

snicoll:
Thanks for the offer. Please proceed and I am happy to review your contribution.

snicoll:
Closing in favour of PR #16047

PR #16047, opened 2019-02-27
Database migrations may not have run before NamedParameterJdbcTemplate is used

clevertension:
…se initialized, fix gh_16032

when there is no `JdcTemplate` in spring bean factory, `NamedParameterJdbcTemplate` should also be created after Flyway or Liquidba, so fix it #16032


snicoll:
@clevertension thank you very much for making your first contribution to Spring Boot. I've merged your contribution to `2.1.x` and `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/eebd906c8a24cb3170cc14fadb70b4c9f01829ef).

clevertension:
i am glad to contribute more code :smiley:

PR #16059, opened 2019-02-28
Remove unused constructor argument

nosan:
Removes `ObjectProvider<FlywayMigrationStrategy> migrationStrategy`, as is not being used.


snicoll:
Good catch, thanks Dmytro

PR #16070, opened 2019-03-02
Permit use of @JsonTest without Jackson

wonwoo:
I think @JsonTest is not only Jackson.

```xml
<dependencies>
    <dependency>
        <artifactId>spring-boot-starter</artifactId>
    </dependency>
    <dependency>
        <groupId>com.google.code.gson</groupId>
        <artifactId>gson</artifactId>
    </dependency>
</dependencies>
```

No dependency Jackson.

```java
@RunWith(SpringRunner.class)
@JsonTest
public class SpringBootJsontestApplicationTests {

    @Autowired
    private GsonTester<VehicleDetails> json;
    ...
}
```

Error Message 
`java.lang.NoClassDefFoundError: com/fasterxml/jackson/databind/Module`


snicoll:
@wonwoo good catch, thank you very much for the PR. This is now merged as for `1.5.x` all the way down to `master`.

PR #16080, opened 2019-03-04
Complete Jetty Access Log configuration properties support

### LINKED ISSUE #15677: Configure Jetty Access Logs - Ignore Paths
Could we enhance the jetty access log configuration to set uris to ignore from logging? 

The existing [configuration](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/autoconfigure/web/ServerProperties.Jetty.Accesslog.html) doesn't seem to allow any way. Looking at how the request log object is initialized in [the jetty customizer](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/embedded/JettyWebServerFactoryCustomizer.java), I think we could just add the property and set the array there.

My correct workaround is to create my own customizer and cast the RequestLog interface to the NCSARequestLog class. Unless this was by design, I'd like to avoid the blind cast in my configuration.

nhomble:
Exposing ignorePaths from ```NCSARequestLog``` into the jetty access log configuration so that users can choose to not log a URI in their request logs as supported out of the box by jetty.
This resolves #15677


nhomble:
I'll confess I am not sure how ssl related tests in the undertow test class are affected by my diff. I'll try to debug more and see if I am triggering the failure but this seems fishy.

wilkinsona:
@nhomble Thanks for the pull request. The problem with Undertow looks to be unrelated. Please feel free to ignore it.

snicoll:
@nhomble thank you for making your first contribution to Spring Boot. Your contribution is now merged on `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/0f605550317da95850b5b8bc588c9655e95ed139).

nhomble:
thanks a lot @snicoll! I hope to find more thing to contribute.

Issue #16073, opened 2019-03-02
Auto-configured PrometheusPushGatewayManager cannot be configured to use HTTPS to connect to the gateway


wilkinsona:
The String-based `PushGateway` constructor that we use hardcodes an `http` scheme to the `URL`. We should switch to the `URL`-based constructor instead to allow users to specify an `https` URL via `management.metrics.export.prometheus.pushgateway.base-url`. We'll have to take care to do so in a backwards-compatible manner, probably by preserving the existing assumption that the scheme should be `http` if not specified.

mbhave:
Closing in favor of PR #16084.

PR #16084, opened 2019-03-04
Auto-configured PrometheusPushGatewayManager cannot be configured to use HTTPS to connect to the gateway

nosan:
gh-16073

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
@@ -16,6 +16,8 @@
 
 package org.springframework.boot.actuate.autoconfigure.metrics.export.prometheus;
 
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.time.Duration;
 import java.util.Map;
 
@@ -119,13 +121,22 @@ public class PrometheusMetricsExportAutoConfiguration {
 				PrometheusProperties prometheusProperties, Environment environment) {
 			PrometheusProperties.Pushgateway properties = prometheusProperties
 					.getPushgateway();
-			PushGateway pushGateway = new PushGateway(properties.getBaseUrl());
+			PushGateway pushGateway = getPushGateway(properties.getBaseUrl());
 			Duration pushRate = properties.getPushRate();
 			String job = getJob(properties, environment);
 			Map<String, String> groupingKey = properties.getGroupingKey();
 			ShutdownOperation shutdownOperation = properties.getShutdownOperation();
 			return new PrometheusPushGatewayManager(pushGateway, collectorRegistry,
 					pushRate, job, groupingKey, shutdownOperation);
+		}
+
+		private PushGateway getPushGateway(String url) {
+			try {
+				return new PushGateway(new URL(url));
+			}
+			catch (MalformedURLException ex) {
+				return new PushGateway(url);
+			}
 		}
 
 		private String getJob(PrometheusProperties.Pushgateway properties,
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
```diff
@@ -117,13 +120,22 @@ public PrometheusPushGatewayManager prometheusPushGatewayManager(
...
-			return new PrometheusPushGatewayManager(pushGateway, collectorRegistry,
-					pushRate, job, groupingKey, shutdownOperation);
+			return new PrometheusPushGatewayManager(
+					getPushGateway(properties.getBaseUrl()), collectorRegistry, pushRate,
+					job, groupingKey, shutdownOperation);
+		}
+
+		private PushGateway getPushGateway(String url) {
+			try {
+				return new PushGateway(new URL(Objects.toString(url, "")));
```
mbhave: I don't think we need to do `Objects.toString(url, "")` since the property has a default value of `localhost:9091`.

nosan:
@mbhave  Thank you very much! 
I have updated this PR.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
```diff
@@ -117,13 +119,22 @@ public PrometheusPushGatewayManager prometheusPushGatewayManager(
...
 			PrometheusProperties.Pushgateway properties = prometheusProperties
 					.getPushgateway();
-			PushGateway pushGateway = new PushGateway(properties.getBaseUrl());
 			Duration pushRate = properties.getPushRate();
 			String job = getJob(properties, environment);
 			Map<String, String> groupingKey = properties.getGroupingKey();
 			ShutdownOperation shutdownOperation = properties.getShutdownOperation();
-			return new PrometheusPushGatewayManager(pushGateway, collectorRegistry,
-					pushRate, job, groupingKey, shutdownOperation);
+			return new PrometheusPushGatewayManager(
```
snicoll: I don't see a reason to change the flow of the code here. I also feel that the previous version was easier to read. Can you please revert that part?

nosan: `spring-java-format` :(
I've introduced a variable to avoid this.

snicoll: That's probably wasn't what I was talking about. I meant the fact you remove the `PushGateway` variable and therefore changed this constructor. I see no good reason for doing so.

nosan: There are 2 constructors.
 - `String`  creates `URL` as `"http://" + address` and delegates it to underlying `URL` constructor.
- `URL` uses `URL` as is.

So, that is why I've introduced a method to create a `PushGateway` instance.

snicoll: That's not what I am talking about and what you pushed force in the meantime is addressing it, thank you.

snicoll:
We've decided to try to parse the URL and log a warning if the URL is invalid, calling the deprecated constructor in that case. We'll also change the default value to include `http://` in it. I'll do that as part of the polish.

nosan:
@snicoll Thank you

Issue #15675, opened 2019-01-10
Application with spring.main.web-application-type=NONE doesn't shut down in conjunction with data mongodb reactive starter


jddipqd:
Non web application (spring.main.web-application-type=NONE) with the dependency to spring-boot-starter-data-mongodb-reactive will not shutdown when close() is called on it's context.
If spring-boot-starter-data-mongodb-reactive dependency is removed application stops regularly.

Spring Boot version is 2.1.1.RELEASE (but it was also occurring in 2.0.x version)

Here's the thread dump after calling close():
```
Full thread dump Java HotSpot(TM) 64-Bit Server VM (25.191-b12 mixed mode):

"nioEventLoopGroup-2-1" #18 prio=10 os_prio=2 tid=0x0000000021240800 nid=0x92d8 runnable [0x0000000027b9e000]
   java.lang.Thread.State: RUNNABLE
        at sun.nio.ch.WindowsSelectorImpl$SubSelector.poll0(Native Method)
        at sun.nio.ch.WindowsSelectorImpl$SubSelector.poll(WindowsSelectorImpl.java:296)
        at sun.nio.ch.WindowsSelectorImpl$SubSelector.access$400(WindowsSelectorImpl.java:278)
        at sun.nio.ch.WindowsSelectorImpl.doSelect(WindowsSelectorImpl.java:159)
        at sun.nio.ch.SelectorImpl.lockAndDoSelect(SelectorImpl.java:86)
        - locked <0x00000007713b5550> (a io.netty.channel.nio.SelectedSelectionKeySet)
        - locked <0x00000007713a1e98> (a java.util.Collections$UnmodifiableSet)
        - locked <0x00000007713a1d30> (a sun.nio.ch.WindowsSelectorImpl)
        at sun.nio.ch.SelectorImpl.select(SelectorImpl.java:97)
        at io.netty.channel.nio.SelectedSelectionKeySetSelector.select(SelectedSelectionKeySetSelector.java:62)
        at io.netty.channel.nio.NioEventLoop.select(NioEventLoop.java:757)
        at io.netty.channel.nio.NioEventLoop.run(NioEventLoop.java:412)
        at io.netty.util.concurrent.SingleThreadEventExecutor$5.run(SingleThreadEventExecutor.java:897)
        at io.netty.util.concurrent.FastThreadLocalRunnable.run(FastThreadLocalRunnable.java:30)
        at java.lang.Thread.run(Thread.java:748)

[...]

PR #16087, opened 2019-03-04
When MongoReactiveAutoConfiguration creates a MongoClient that uses Netty, EventLoopGroup threads prevent the JVM from exiting

nosan:
gh-15675


wilkinsona:
Thanks very much for the PR, @nosan. The proposed changes have been merged into 2.1.x and forward into master along with a [little bit of polish](https://github.com/spring-projects/spring-boot/commit/f20d9a62aeee672c399fd9c2e6fae29783f184db).

nosan:
@wilkinsona Thank you.

Issue #2895, opened 2015-04-29
Support zip64 format executable archives


pmdusso:
I first wrote this in the Gradle forum, and they recommended me transfer it to here.

 I wasn't able to execute the **war** generated from **bootRepackage** in a project derived from jHipster framework. The topic still opened in [Stackoverflow](https://stackoverflow.com/questions/29848088/cannot-run-the-war-generated-from-gradle-bootrepackage-in-a-jhipster-based-web-a).

I created a small application and put it on [Github](https://github.com/pmdusso/boot-repackage-test) to help you help me with this. Following these steps I believe you can reproduce the problem also. I'm first packing a simple Spring application and then running the generated war. This will create 66000 files in the resource folder. Packing the same app again will fail because in the second time there is more than 65535 files. We confirmed this by cleaning the resource folder and bootRepackaging it again successfully.

Then I set the **zip64** property to **true** (I had to do it in the _jar { }_ task also. In my original application however I do not have a jar task and set only in the war task). The bootRepackage task completes with success, but then when I run the war I get the **no Start-Class error**.

Running this for the first time will work, because total files < 65535 and zip64 = false

> ./gradlew clean bootRepackage; java -jar build/libs/app-0.1-SNAPSHOT.war

Running this for the second time will not work, because total files > 65535 and zip64 = false

> ./gradlew clean bootRepackage; java -jar build/libs/app-0.1-SNAPSHOT.war

Running this for the third time (equal to alternative one) will work, because total files < 65535 and zip64 = false

> rm src/main/resources/*; ./gradlew clean bootRepackage; java -jar build/libs/app-0.1-SNAPSHOT.war

At this point we have total files > 65535 and zip64 = false. We change zip64 property to true and try to bootPackage it again

> Set zip64 = true in build.gradle in jar and war tasks.
> ./gradlew clean bootRepackage; java -jar build/libs/app-0.1-SNAPSHOT.war

```
java.lang.IllegalStateException: No 'Start-Class' manifest entry specified in jar:file:../boot-repackage-test/build/libs/app-0.1-SNAPSHOT.war!/
        at org.springframework.boot.loader.archive.Archive.getMainClass(Archive.java:57)
        at org.springframework.boot.loader.ExecutableArchiveLauncher.getMainClass(ExecutableArchiveLauncher.java:69)
        at org.springframework.boot.loader.Launcher.launch(Launcher.java:61)
[...]

PR #16091, opened 2019-03-04
Support zip64 jars

cvienot:
Closes gh-2895

Add support for zip64 jars by parsing zip64 specific records, to read correct number of files, offsets and sizes


spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/CentralDirectoryEndRecord.java
```diff
@@ -109,21 +121,136 @@ public long getStartOfArchive(RandomAccessData data) {
...
 	 * @return the number of records in the zip
 	 */
 	public int getNumberOfRecords() {
-		long numberOfRecords = Bytes.littleEndianValue(this.block, this.offset + 10, 2);
-		if (numberOfRecords == 0xFFFF) {
-			throw new IllegalStateException("Zip64 archives are not supported");
+		if (isZip64()) {
+			return this.zip64End.get().getNumberOfRecords();
+		}
+		else {
```
johnhunt: Because of the return, you don't need an else {} statement

spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/CentralDirectoryEndRecord.java
```diff
@@ -109,21 +121,136 @@ public long getStartOfArchive(RandomAccessData data) {
 	 * @return the central directory data
 	 */
 	public RandomAccessData getCentralDirectory(RandomAccessData data) {
-		long offset = Bytes.littleEndianValue(this.block, this.offset + 16, 4);
-		long length = Bytes.littleEndianValue(this.block, this.offset + 12, 4);
-		return data.getSubsection(offset, length);
+		if (isZip64()) {
+			return this.zip64End.get().getCentratDirectory(data);
+		}
+		else {
```
johnhunt: Because of the return, you don't need an else {} statement

spring-boot-project/spring-boot-tools/spring-boot-loader/src/test/java/org/springframework/boot/loader/archive/JarFileArchiveTests.java
```diff
@@ -131,8 +132,12 @@ public void unpackedLocationsFromSameArchiveShareSameParent() throws Exception {
 	public void zip64ArchivesAreHandledGracefully() throws IOException {
```
ilya123s: This test name case should be renamed as it's not longer checking for an exception

spring-boot-project/spring-boot-tools/spring-boot-loader/src/test/java/org/springframework/boot/loader/archive/JarFileArchiveTests.java
```diff
@@ -131,8 +132,12 @@ public void unpackedLocationsFromSameArchiveShareSameParent() throws Exception {
...
-				.withMessageContaining("Zip64 archives are not supported");
+		JarFileArchive zip64Archive = new JarFileArchive(file);
+		Iterator<Entry> it = zip64Archive.iterator();
+		for (int i = 0; i < 65537; i++) {
+			assertThat(it.hasNext()).as(i + "nth file is present").isTrue();
+			it.next();
+		}
 	}
 
 	@Test
```
ilya123s: This test name case should be renamed as it's not longer checking for an exception

cvienot: i renamed it

wilkinsona:
Thanks very much for the PR, @cvienot. Zip64 was one of our oldest outstanding feature requests so it's great to finally have support for it. Thanks again!

PR #16099, opened 2019-03-05
Order the config location in javadoc by the search order

mhewedy:


snicoll:
@mhewedy thank you for the PR and spotting that the list was using a wrong order. However, this order [isn't the order in which properties are loaded](https://github.com/spring-projects/spring-boot/blob/09ad80b3d9c02c6d06ef46f19c11d704d88e4d84/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigFileApplicationListener.java#L111) but the precedence order. I've [polished your contribution](https://github.com/spring-projects/spring-boot/commit/372802a2038b7c180a3e65a4ef3c4ccff9401c9e) to [clarify that](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-external-config-application-property-files).

Issue #16102, opened 2019-03-05
Random management.server.port gets set to the same value as local.server.port


gkonupek:
A minimum project to reproduce this issue is available here: https://github.com/gkonupek/SpringBootActuatorIssue

The project is a basic Spring Boot web application with Actuator. A problem occurs in tests when using Spring Boot version 2.1.x. In this case, I'm using version 2.1.3.RELEASE.

In the mentioned project, on branch "master" there is the following situation:

- `DemoApplicationTests` I'm using `@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)` annotation to run the application on a random, free port.
- In `application.properties` file, I've specified `management.server.port=0` to make Actuator endpoint run on a random, free port as well.
- Furthermore, in `DemoApplicationTests`:
  - I'm injecting 2 values: `serverPort` (annotated with `@LocalServerPort`) and `managementPort` (annotated with `@LocalManagementPort`).
  - I've created a simple JUnit 5 test in which I'm checking, that both mentioned ports have a different value (that's my expectation).
  - Notice, that the test is also annotated with `@ActiveProfiles("test")` - it will be important later.
- The Spring Boot version used in this case is 2.0.5.RELEASE.

In this configuration I expect the test to pass. Indeed, the generated ports are different.

**Spring Boot 2.1.x**

Now, in the repository there is a second branch named "spring-boot-2.1.x". The only difference between master is that the Spring Boot version has been upgraded to 2.1.3.RELEASE.

Unfortunately, in this configuration the test is failing. Both `serverPort` and `managementPort` have the same value.

**Role of `@ActiveProfiles` annotation?**

Another interesting case is present on the branch named "without-active-profile". In this case, the Spring Boot version is still 2.1.3.RELEASE. However, I've removed the `@ActiveProfiles("test")` annotation from `DemoApplicationTests`. In this case, the test is passing again - the port numbers are different.

The last scenario is actually the workaround, which I've applied to make my application work. Instead of using the annotation I could just import my `application-test.properties` file manually, by using `@TestPropertySource` annotation. However, I'd really love to see this issue explained and eventually fixed.

Cheers,
Grzegorz

wilkinsona:
[...]

PR #16108, opened 2019-03-06
Random management.server.port should not set to the same value as local.server.port

nosan:
`management.server.port` should be **0** when `local.server.port` has the same value and `management.server.port` has been defined in the different `PropertySource`

gh-16102

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/web/SpringBootTestRandomPortEnvironmentPostProcessor.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/web/SpringBootTestRandomPortEnvironmentPostProcessor.java
@@ -52,7 +52,8 @@ class SpringBootTestRandomPortEnvironmentPostProcessor
 		}
 		Integer managementPort = getPropertyAsInteger(environment,
 				MANAGEMENT_PORT_PROPERTY, null);
-		if (managementPort == null || managementPort.equals(-1)) {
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/web/SpringBootTestRandomPortEnvironmentPostProcessor.java
```diff
@@ -57,7 +57,7 @@ public void postProcessEnvironment(ConfigurableEnvironment environment,
 		}
 		Integer serverPort = getPropertyAsInteger(environment, SERVER_PORT_PROPERTY,
 				8080);
-		if (!managementPort.equals(serverPort)) {
+		if (!managementPort.equals(serverPort) || managementPort == 0) {
```
mbhave: @nosan I think we'd be able to achieve this by adding a `|| managementPort.equals(0)` [here](https://github.com/spring-projects/spring-boot/blob/f5eb853f263ad96ee304749525cd7a853c8c18f2/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/web/SpringBootTestRandomPortEnvironmentPostProcessor.java#L55)

nosan: @mbhave thank you

mbhave:
Thanks @nosan. This is now on master.

PR #16115, opened 2019-03-06
Check for multiple processors in OnClassCondition

JWThewes:
This is my first commit so any help or guidance is welcome.
I've implemented the Runtime.getRuntime().availableProcessors() > 1 check and added a case when the processor count is equal 1.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -97,12 +97,6 @@
 			<groupId>io.micrometer</groupId>
 			<artifactId>micrometer-jersey2</artifactId>
 			<optional>true</optional>
-			<exclusions>
-				<exclusion>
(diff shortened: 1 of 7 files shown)


JWThewes:
So as this was my first PR I assume that it was ok? Would be great to get some sort of feedback I can apply when creating other PRs.

Kind regards
Jan

snicoll:
@JWThewes PR are reviewed and merged accordingly. We'll make sure to provide feedback in due course.

snicoll:
@JWThewes thank you for the contribution. They were several problems with it that I've fixed in  4d73211. 

In the future, please do not create a PR from `master` as it makes it harder to push additional changes in case the team is asking for changes. Make sure you're up to date and create a branch first. As you can see from the commit tab, this PR has two commits, one of which is unrelated. I've dropped it locally before merging.

As for the code, the single case was still starting a thread unfortunately (see `createOutcomesResolver`).

Issue #16101, opened 2019-03-05
@WebFluxTest does not enable auto-configuration for any reactive template engines


wilkinsona:
Mustache, FreeMarker, and Thymeleaf all support reactive web apps and should be included in `@WebFluxTest`

ayudovin:
@wilkinsona, I want to work on it. Please let me know if I can start working on it?

wilkinsona:
@ayudovin Thanks for asking, @ayudovin. Yes, you can start working on this. A PR would be much appreciated.

wilkinsona:
Closing in favour of #16170.

PR #16170, opened 2019-03-08
@WebFluxTest does not enable auto-configuration for any reactive template engines

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/16101)


wilkinsona:
Thanks for the PR. Unfortunately, this isn't the change that we need. Rather than creating separate annotations, the auto-configurations for the template engines with reactive support need to be included in `@WebFluxTest`. That can be achieved by adding them to the existing `org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebFlux` in `spring.factories`. Would you like to update your PR to do that?

ayudovin:
@wilkinsona, yes, I would like to update my PR.

PR #16178, opened 2019-03-08
Dependency management for micrometer-registry-stackdriver is missing

izeye:
This PR adds dependency management for `micrometer-registry-stackdriver`.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -980,6 +980,11 @@
 			</dependency>
 			<dependency>
 				<groupId>io.micrometer</groupId>
+				<artifactId>micrometer-registry-stackdriver</artifactId>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, @izeye. Does this mean we're missing some auto-configuration as well? On the dependency management side of it, I'd much rather fix this by being able to import a Micrometer bom. @jkschneider,  is there any chance of https://github.com/micrometer-metrics/micrometer/issues/469 being tackled soon?

izeye:
@wilkinsona Thanks for the feedback! I'm not sure about auto-configuration on it as [it's marked as incubating](https://github.com/micrometer-metrics/micrometer/blob/master/implementations/micrometer-registry-stackdriver/src/main/java/io/micrometer/stackdriver/StackdriverMeterRegistry.java#L60).

wilkinsona:
Thanks, @izeye. Let's hold off on any auto-configuration for now then.

PR #16180, opened 2019-03-09
Fail to bind ConfigurationProperties bean with no `BeanDefinition`

spencergibb:


snicoll:
Thanks @spencergibb. Unfortunately that test passes for me. I've tried to make it fail and was unable to do so. I also wanted to fix the issue slightly differently which is done in https://github.com/spring-projects/spring-boot/commit/d07a6ff0bb1c0b5e0f907ecb7508cb680ecd9e9f with a test. It would be nice to test what actually happens in you case so we're still interested by a test if you manage to get one failing without that commit. Thanks for the PR, in any case.

spencergibb:
Yeah, I could never run the test, so I want sure if it worked. Thanks.

PR #16221, opened 2019-03-13
Metrics may be updated twice when CompositeMeterRegistry exists

izeye:
This PR prevents adding `MeterBinder`s for `CompositeMeterRegistry`.

See https://github.com/micrometer-metrics/micrometer/issues/1282

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryConfigurer.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryConfigurer.java
@@ -22,6 +22,7 @@ import java.util.stream.Collectors;
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.Metrics;
 import io.micrometer.core.instrument.binder.MeterBinder;
+import io.micrometer.core.instrument.composite.CompositeMeterRegistry;
 import io.micrometer.core.instrument.config.MeterFilter;
 
 import org.springframework.beans.factory.ObjectProvider;
@@ -45,13 +46,16 @@ class MeterRegistryConfigurer {
 
 	private final boolean addToGlobalRegistry;
 
+	private final boolean hasCompositeMeterRegistry;
+
 	MeterRegistryConfigurer(ObjectProvider<MeterRegistryCustomizer<?>> customizers,
 			ObjectProvider<MeterFilter> filters, ObjectProvider<MeterBinder> binders,
-			boolean addToGlobalRegistry) {
+			boolean addToGlobalRegistry, boolean hasCompositeMeterRegistry) {
 		this.customizers = customizers;
 		this.filters = filters;
 		this.binders = binders;
 		this.addToGlobalRegistry = addToGlobalRegistry;
+		this.hasCompositeMeterRegistry = hasCompositeMeterRegistry;
 	}
 
 	void configure(MeterRegistry registry) {
@@ -59,7 +63,16 @@ class MeterRegistryConfigurer {
 		// tags or alter timer or summary configuration.
 		customize(registry);
 		addFilters(registry);
-		addBinders(registry);
+		if (this.hasCompositeMeterRegistry) {
+			if ((registry instanceof CompositeMeterRegistry)) {
+				addBinders(registry);
+			}
+		}
+		else {
+			if (!(registry instanceof CompositeMeterRegistry)) {
+				addBinders(registry);
+			}
+		}
 		if (this.addToGlobalRegistry && registry != Metrics.globalRegistry) {
 			Metrics.addRegistry(registry);
 		}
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for the PR. We'll need to tread carefully here as the proposed change will partially revert [this bug fix](https://github.com/spring-projects/spring-boot/commit/1fce46294437d7393ac8cf319816b857e4049cd2#diff-cff6f2c869efcb8c519f289e635f56b5).

izeye:
Sorry, this doesn't work. I judged it's working by seeing Spring Boot Actuator metrics endpoint but this change makes `CompositeMeterRegistry` have no bound meters. I think a possible fix is described in https://github.com/micrometer-metrics/micrometer/pull/1283#discussion_r265197044 but I'm not sure how to figure out there's any composite meter registry there at the moment.

izeye:
I tried the above mentioned approach and added some tests in 6413c5b.

izeye:
If there are more than one composite meter repositories, this fix won’t work. I know it is possible theoretically but am not sure such a setup is being used practically. If the root composite meter repository can be identified there, the case can be handled. Furthermore theoretically tree for meter registries could be more than one and I’m not sure this setup is valid.

checketts:
Thanks for the fix! @izeye. What is the `Metrics.globalRegistry`? Could that be the 'root' composite (if there were multiple?

izeye:
@checketts Thanks for the feedback! I considered it but it doesn't seem to be an option here as it's optional as follows:

```
management.metrics.use-global-registry=true # Whether auto-configured MeterRegistry implementations should be bound to the global static registry on Metrics.
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryConfigurer.java
```diff
@@ -45,21 +46,33 @@
...
+		this.hasCompositeMeterRegistry = hasCompositeMeterRegistry;
 	}
 
 	void configure(MeterRegistry registry) {
 		// Customizers must be applied before binders, as they may add custom
 		// tags or alter timer or summary configuration.
 		customize(registry);
 		addFilters(registry);
-		addBinders(registry);
+		if (this.hasCompositeMeterRegistry) {
```
shakuzen: Rather than `hasCompositeMeterRegistry` it might be better to be straightforward about what this controls. Something like `onlyBindToCompositeRegistry`.

izeye: @shakuzen Thanks for the feedback! I updated it as suggested in 6b3bd78.

Imaskar:
I have this problem on prod, we use prometheus+New Relic and our Hystrix metrics double. But we use Spring Boot 1.5, so I backported these changes to Micrometer's own bridge (micrometer-spring-legacy). After testing on QA we see that our metrics are no longer doubling. But Micrometer won't accept PR until this one is accepted.

So I would be grateful for an update on plans or concerns for this PR. Thank you.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryConfigurer.java
```diff
@@ -45,21 +46,33 @@
 
 	private final boolean addToGlobalRegistry;
 
+	private final boolean onlyBindToCompositeMeterRegistry;
+
 	MeterRegistryConfigurer(ObjectProvider<MeterRegistryCustomizer<?>> customizers,
 			ObjectProvider<MeterFilter> filters, ObjectProvider<MeterBinder> binders,
-			boolean addToGlobalRegistry) {
+			boolean addToGlobalRegistry, boolean onlyBindToCompositeMeterRegistry) {
```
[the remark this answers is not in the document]
izeye: @wilkinsona Thanks for the feedback! I see the point. I couldn't think of a good but short name describing what it's doing at the moment. How about dropping back into the original name, `hasCompositeMeterRegistry` describing its state?

izeye:
@wilkinsona Thanks for the feedback! I updated in 04c7553 as suggested.

Imaskar:
The other way to do it would be to always have a composite registry (a dispatcher) even when a single registry is in use or move globalRegistry from optional to required. But current variant seems to be fine too.

BTW, CI build is failing on unrelated test.

Imaskar:
@wilkinsona Are you fine with the current state or want to dig some more?

wilkinsona:
The current state looks good to me.

wilkinsona:
Thanks once again, @izeye. The proposed changes have been merged into 2.1.x and forwards into master.

izeye:
@wilkinsona Thanks! By the way, is there any reason not to merge into the 2.0.x branch?

mbhave:
@izeye 2.0.x is EOL now and [2.0.9](https://spring.io/blog/2019/04/03/spring-boot-2-0-9-released) was the last release in the 2.0.x line.

izeye:
@mbhave Thanks for the quick feedback 👍 I missed the news somehow 😅

PR #16224, opened 2019-03-13
Fix typo in logging.group property description

xak2000:
The description of `logging.group` property used wrong name in the example.


snicoll:
@xak2000 good catch, thank you!

Issue #16239, opened 2019-03-15
Upgrade to Spring Cloud Connectors 1.2.8


azizsonawalla:
New version released: https://github.com/spring-cloud/spring-cloud-connectors/releases/tag/v1.2.8.RELEASE

mbhave:
Thanks but we have a semi-automated process to upgrade dependencies. We'll process this in due course and therefore don't need an issue for it.

PR #16240, opened 2019-03-15
Upgrade spring cloud connectors 1.2.7->1.2.8

azizsonawalla:
Resolves https://github.com/spring-projects/spring-boot/issues/16239


mbhave:
Thanks for the PR but, as noted in the pull request template, please do not open a pull request for a straightforward dependency upgrade (one that only updates the version property). We have a semi-automated process for such upgrades that we prefer to use. However, if the upgrade is more involved (such as requiring changes for removed or deprecated API) your pull request is most welcome.

PR #16248, opened 2019-03-16
Use file: URL rather than jar:file: URL to speed up exploded application launch

hengyunabc:
Convert the jar file URL to file URL. 

For example, convert `jar:file:demo.jar` to `file:demo.jar`.

In URLClassLoader, file URL is faster than jar file URL.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/archive/ExplodedArchive.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/archive/ExplodedArchive.java
@@ -117,7 +117,7 @@ public class ExplodedArchive implements Archive {
 	protected Archive getNestedArchive(Entry entry) throws IOException {
 		File file = ((FileEntry) entry).getFile();
 		return (file.isDirectory() ? new ExplodedArchive(file)
-				: new JarFileArchive(file));
+				: new JarFileArchive(file, file.toURI().toURL()));
 	}
 
 	@Override
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/test/java/org/springframework/boot/loader/JarLauncherTests.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/test/java/org/springframework/boot/loader/JarLauncherTests.java
@@ -44,9 +44,7 @@ public class JarLauncherTests extends AbstractExecutableArchiveLauncherTests {
 		assertThat(archives).hasSize(2);
 		assertThat(getUrls(archives)).containsOnly(
 				new File(explodedRoot, "BOOT-INF/classes").toURI().toURL(),
-				new URL("jar:"
-						+ new File(explodedRoot, "BOOT-INF/lib/foo.jar").toURI().toURL()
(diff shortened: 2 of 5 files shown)


hengyunabc:
### Demo

Demo: https://start.spring.io/starter.zip?fakeusernameremembered=&fakepasswordremembered=&type=maven-project&language=java&bootVersion=2.1.3.RELEASE&baseDir=demo&groupId=com.example&artifactId=demo&name=demo&description=Demo+project+for+Spring+Boot&packageName=com.example.demo&packaging=jar&javaVersion=1.8&inputSearch=&style=web

On my mac:

* before: Started DemoApplication in 5.* seconds
* after: Started DemoApplication in 4.* seconds

### Large Application

In a large application which has 1000+ jars.

* before: Started Application in 360 seconds
* after: Started Application in 320 seconds

wilkinsona:
Thanks for the PR. Transformation of the URL as proposed here will lose the custom `URLStreamHandler` that is used by our loader. That may be ok, but it needs to be considered and tested very carefully.

If losing the handler is ok, then, given that we're in control of how the URLs are created, creating them and then transforming would seem to be more complicated than is necessary. It would be more efficient to create them differently in the first place.

Can you please review your changes with these points in mind?

hengyunabc:
## Why the URLs are `jar:file:` URLs

There are two kinds of Archive:

* ExplodedArchive
* JarFileArchive

In the `org.springframework.boot.loader.PropertiesLauncher#addNestedEntries(List<Archive>)` method, all the obtained nested arthives are `JarFileArchive`, so the URLs are `jar:file:` URLs.

```java
	private void addNestedEntries(List<Archive> lib) {
		// The parent archive might have "BOOT-INF/lib/" and "BOOT-INF/classes/"
		// directories, meaning we are running from an executable JAR. We add nested
		// entries from there with low priority (i.e. at end).
		try {
			lib.addAll(this.parent.getNestedArchives((entry) -> {
				if (entry.isDirectory()) {
					return entry.getName().equals(JarLauncher.BOOT_INF_CLASSES);
				}
				return entry.getName().startsWith(JarLauncher.BOOT_INF_LIB);
			}));
		}
		catch (IOException ex) {
			// Ignore
		}
	}
```

## URLStreamHandler

The `org.springframework.boot.loader.jar.Handler` only handle the `jar:file:` URL, so after converting to `file:` URL, it has no effect to the ClassLoader.

wilkinsona:
Thanks, but that's not really what I meant. I understand why they are currently `jar:file:` URLs. I was interested to know if you'd considered changing that at source rather than modifying the URLs once they've been returned. It would be cleaner to provide `file:` URLs directly when the jar is available directly on disk.

hengyunabc:
@wilkinsona I got what you mean, your suggestion is useful. 
Please check the new commit.

hengyunabc:
@wilkinsona Does the new commit match your meaning?

wilkinsona:
Yes, that's much more like what I had in mind. Thanks for the updates.

wilkinsona:
Thanks very much for the PR, @hengyunabc. The proposed changes have been merged into master along with a [polish commit](https://github.com/spring-projects/spring-boot/commit/5639685770c0c656e3190ff8c042aadfb311da0c).

PR #16252, opened 2019-03-18
Support resolve health groups

eddumelendez:

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/DefaultHealthIndicatorRegistry.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/DefaultHealthIndicatorRegistry.java
@@ -18,7 +18,9 @@ package org.springframework.boot.actuate.health;
 
 import java.util.Collections;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
+import java.util.stream.Collectors;
 
 import org.springframework.util.Assert;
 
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks for the PR, @eddumelendez. This seems to be taking a different approach to the one described in [this comment](
https://github.com/spring-projects/spring-boot/issues/14022#issuecomment-412039697). If you tried the approach described there, can you please let us know why it didn't work out and why you chose this approach instead?

eddumelendez:
sorry for the delay @wilkinsona. I had in mind the approach discussed in the issue which I can understand is not reflected into this PR. But, this is an initial draft indeed, this PR is missing the configuration through `application.properties`. Hope this part is ok so I can proceed with the missing parts.

wilkinsona:
Thanks, @eddumelendez. I am struggling to understand how the configuration of health indicator groups will map onto the changes that you have proposed thus far. For example, I had not expected the health endpoint to have an operation that takes a list of components added to it. Instead, we had anticipated that each configured group would result in a new composite health indicator being created:

> We then thought that `CompositeHealthIndicator` could be reused and created with all the health indicators in the group. One gotcha that @philwebb identified is that these composite health indicators that represent a group will need to be hidden from the main health check otherwise any grouped indicators would be called twice for a single check.

snicoll:
@eddumelendez I've made quite some progress (see https://github.com/spring-projects/spring-boot/issues/14022) and this draft hasn't moved in a while so I am closing it now. If you have time to look at the current proposal, your feedback is more than welcome. Thanks for the PR, in any case!

Issue #13627, opened 2018-07-01
Support WebExceptionHandler in @WebFluxTest


hantsy:
`@WebMvcTest` worked well with `@ExceptionHandler`. But `@WebFluxTest` ignores it at all and throw exceptions in the background, I created an post on stackoverflow before to describe it.

https://stackoverflow.com/questions/48162730/how-to-test-exception-with-webfluxtest

related codes: https://github.com/hantsy/angular-spring-reactive-sample/blob/master/server/src/test/java/com/example/demo/PostControllerTest.java#L112-L123

alimate:
@mbhave Does the `WebFluxTest` support the exception handlers?

PR #16266, opened 2019-03-19
Import ErrorWebFluxAutoConfiguration with @WebFluxTest

### LINKED ISSUE #16258: @WebFluxTest Doesn't Import the ErrorWebFluxAutoConfiguration
Synopsis
---
As I was experimenting with the `@WebFluxTest`s to test Reactive controllers, I noticed that the provided test _auto-configurations_ do not import the default `ErrorWebExceptionHandler`, so the following test would fail:
```java
@AutoConfigureErrors
@RunWith(SpringRunner.class)
@WebFluxTest(PersonController.class)
public class ReactiveWebApplicationTests {

    @Autowired private WebTestClient webClient;

    @Test
    public void personName_CanNotBeBlank() {
        webClient.post().uri("/persons")
                .contentType(MediaType.APPLICATION_JSON_UTF8)
                .syncBody("{}")
                .exchange()
                .expectStatus().isBadRequest()
                .expectBody()
                    .jsonPath("$.errors[0].code").isEqualTo("name.blank");
    }
}
```

This can be easily fixed just by importing the `ErrorWebFluxAutoConfiguration`:
```java
@AutoConfigureErrors
@RunWith(SpringRunner.class)
@WebFluxTest(PersonController.class)
@ImportAutoConfiguration(ErrorWebFluxAutoConfiguration.class)
public class ReactiveWebApplicationTests {

    @Autowired private WebTestClient webClient;

    @Test
    public void personName_CanNotBeBlank() {
        webClient.post().uri("/persons")
                .contentType(MediaType.APPLICATION_JSON_UTF8)
                .syncBody("{}")
                .exchange()
                .expectStatus().isBadRequest()
                .expectBody()
                    .jsonPath("$.errors[0].code").isEqualTo("name.blank");
    }
}
```

Suggestion
---
Although the problem is so easy to fix, that'd be great if we import the default auto-configuration automatically. For example by adding an `AutoConfigureErrorWebFlux` annotation.

Related Issues
---
Apparently supporting `WebExceptionHandler`s was added in #13627 but the default auto-configuration is missing from the `WebFluxTest`.

alimate:
It's worth mentioning that the imported auto-configuration will register a bean of type `ErrorWebExceptionHandler` if there is no such bean already in the application context. So in order to provide a custom handler, we need to register a bean of type `ErrorWebExceptionHandler`, not the `WebExceptionHandler`.

Closes #16258
Related Issue #13627


spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/web/reactive/webclient/ExampleWebExceptionHandler.java
```diff
@@ -27,9 +29,11 @@
  * Example {@link WebExceptionHandler} used with {@link WebFluxTest} tests.
  *
  * @author Madhura Bhave
+ * @author Ali Dehghani
  */
 @Component
-public class ExampleWebExceptionHandler implements WebExceptionHandler {
+@ConditionalOnProperty(name = "custom-error-handler.enable")
```
mbhave: @alimate Could you explain why this change is necessary?

alimate: @mbhave Apparently, As part of the #13627, We support the `WebExceptionHandler`s in `WebFluxTest`s. As far as I know, this behavior is inconsistent (Please correct me if I'm wrong) with when we run Spring Boot applications manually. In the latter scenario, we should register a bean of type `ErrorWebExxceptionHandler` to disable the default error handling and use a custom approach for error handling.

Basically, here I'm implementing the `ErrorWebExceptionHandler` interface to align error handling in automatic and manual testing scenarios.

 Also, I'm registering this new handler conditionally to test both of the following:
 - When there no `ErrorWebExceptionHandler` registered, the default provided by the `ErrorWebFluxAutoConfiguration` should be used.
 - Otherwise, the default one should be discarded and our custom `ErrorWebExceptionHandler` should be used.

mbhave: Thanks for the explanation. I think we'd still want to keep the `WebExceptionHandler` to make sure that those get [picked up](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTypeExcludeFilter.java#L53) by `@WebFluxTest`. For testing that the `ErrorWebFluxAutoConfiguration` is imported, the test that you've added in `WebFluxTestAutoConfigurationIntegrationTests` should suffice. 

Let's see what @bclozel thinks.

alimate: I've created a sample repo for this inconsistent behaviour
https://github.com/alimate/webflux-Inconsistency

bclozel: This is not a case of `ErrorWebExceptionHandler` vs. `WebExceptionHandler`, since `ErrorWebExceptionHandler` is extending `WebExceptionHandler` after all.

This is more about the ordering of the beans - by default Spring Boot configures the `ErrorWebExceptionHandler` instance at order `-1` to make sure to handle errors before the `WebExceptionHandler` provided by Spring WebFlux.

An application can still configure a `WebExceptionHandler` earlier to process errors before the handler provided by Spring Boot.

In short, I agree with @mbhave and the `@ConditionalOnProperty` change is not necessary.

alimate: @bclozel Thanks. So instead of registering the `ExampleWebExceptionHandler` as a `Component`, we're better off registering it manually using a `@Bean` with different `Order`s, right?

alimate: Anyway, I removed the `Component` and `ConditionalOnProperty` annotations from the `ExampleWebExceptionHandler`. 

Instead, register the exception handler using a `@Bean` with different priorities to test the fact that `WebExceptionHandler`s with appropriate `Order`s would be called before the default handler provided by the `ErrorWebFluxAutoConfiguration`.

What do you guys think? @bclozel @mbhave

mbhave: @alimate I don't think we need to change anything related to `ExampleWebExceptionHandler`. The fact that a `WebExceptionHandler` can be configured with a higher precedence than the one provided by Spring Boot doesn't need to be tested in a test for `@WebFluxTest`. This issue is about ensuring that `ErrorWebFluxAutoConfiguration` is imported by `@WebFluxTest` and I think the only tests you need to change for that are `WebFluxTestAutoConfigurationIntegrationTests`.

alimate: @mbhave OK. I'm reverting back some of the changes 👍

snicoll:
@alimate thank you for making your first contribution to Spring Boot. This is now merged in `2.1.x` and `master` with [a polish commit](https://github.com/spring-projects/spring-boot/commit/090f5f539710bd80995768a4a6decc8f6cb33550) as one test was still failing.

PR #16269, opened 2019-03-20
Use the latest DTD declaration for Checkstyle

izeye:
This PR changes to use the latest DTD declaration for Checkstyle as follows:

- https://github.com/checkstyle/checkstyle/blob/master/config/checkstyle_checks.xml#L2-L4
- https://github.com/checkstyle/checkstyle/blob/master/config/suppressions.xml#L3-L5


snicoll:
To the point as always. Thanks again @izeye!

PR #16275, opened 2019-03-20
Introduce Sticky health indicator

fvlad:
Hi,

We have lots of microservices running in K8s on production and and we use their /health endpoints to check whether everything is OK or not. Sometimes we have issues that that some health dependency gets broken (for example a temporal problem with a database) and the whole health of the service goes DOWN. This is the desired behaviour for most of the health indicators, but for some of them a DOWN health indicator should not stop traffic coming to the service, as the dependency might not be very important. But I would still like to have those health indicators, so that at least on startup I can be sure that some connection and credentials are defined correct.

The idea behind this PR is to be able to define a list of those special health indicators with some configuration like this: `management.health.status.sticky[0]=myindicator`. Sticky health indicator must get UP at least once (on startup), and then it will always stay UP. In case the original health indicator goes to DOWN or any other, the status is still shown UP, but original status and original details are provided inside the details.

I can add more tests and reactive health indicators support if needed.
WDYT?


wilkinsona:
Thanks for the proposal. I think we'd prefer to tackle this slightly differently by allowing health indicators to be grouped. This would allow, for example, a group for liveness checks and a group for readiness checks. This change is being tracked by #14022 and we hope to include it in 2.2.0. Thanks anyway.

Issue #16184, opened 2019-03-09
Skip lazy initialization of beans that explicitly set the lazy attribute to false


jhaeyaert:
Hi,

The suspected issue described below is related to the version 2.2.0.M1 of Spring Boot.

It appears that the management endpoints are not reachable when using the new lazy initialization feature with actuator endpoints exposed on another port than the default one.

Steps to reproduce :

- Create a new project from spring initializr with the new Spring Boot 2.2.0.M1 version
- Add Reactive Web and Actuator dependencies
- Set the properties `management.server.port=8081` and `spring.main.lazy-initialization=true`

Hit the endpoint `http://localhost:8081/actuator/health` and you'll get a connection refused.

Switch off the lazy initialization and the actuator endpoints will be back again. 
Note that actuator endpoints are working correctly if I switch on the lazy initialization and do not change the management port.

Is something I didn't understand or is there a little bug in the lazy initialization feature ?

By the way, thank you for this feature that I have been waiting for a long time!

wilkinsona:
Thanks for trying out the milestone. This sounds like a bug to me. We caught a similar one that switched off DevTools restart just before we released but this one slipped through.

wilkinsona:
Adding the following bean to your app should work around the problem for now:

```java
@Bean
static BeanFactoryPostProcessor ensureEagerManagementServerInitializationPostProcessor() {
	return (beanFactory) -> {
		String namePrefix = ManagementContextAutoConfiguration.class.getName() + "$";
		for (String definitionName : beanFactory.getBeanDefinitionNames()) {
			if (definitionName.startsWith(namePrefix)) {
				beanFactory.getBeanDefinition(definitionName).setLazyInit(false);
			}
		}
	};
}
```

This will ensure that the logic that creates the child context for the management server is not lazy and that the server is created and started.

There's a secondary problem here. The beans in the separate management context that's used when the management port does not match the server port are all initialised eagerly. They should be lazy like the main context so that the server is created and listening for requests but, for example, the endpoints are not created until a request is received.

tkvangorder:
Is it possible to skip any bean definitions that explicitly set their lazy attribute to false? `@Lazy(false)`

philwebb:
@tkvangorder That's a really nice idea, but I think we might need a framework change to be able to do it. Currently `AbstractBeanDefinition.lazyInit` is a `boolean` so we don't know if the value is `false` because that's the default, or if it's because `setLazyInit` has been called.

[...]

PR #16279, opened 2019-03-21
Allow eager init to be enforced

hcassus:
Enable eager init when spring.main.lazy-initialization is set to true.

When enabling spring.main.lazy-initialization, the ability to control
individual components to be eagerly initialized is lost.

This PR aims to allow granular configuration in both default init
strategies by doing the following:
    - **@Lazy** or **@Lazy(value=true)** annotated beans are always initialized lazily
    - **@Lazy(value=false)** annotated beans are always initialized eagerly
    - Non-annotated beans follow the default initialization strategy
      configured by spring.main.lazy-initialization

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
@@ -16,23 +16,8 @@
 
 package org.springframework.boot;
 
-import java.lang.reflect.Constructor;
-import java.security.AccessControlException;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Map;
-import java.util.Properties;
-import java.util.Set;
-
 import org.apache.commons.logging.Log;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@hcassus Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16279)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16279) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@hcassus Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16279)!

mbhave:
@hcassus thanks for the PR but I think we'd want to solve this slightly differently. There are other ways of setting the lazy attribute on the bean definition so just looking at the value of the annotation might be a bit limiting. There is some discussion about how we can make this a more general strategy [here](https://github.com/spring-projects/spring-boot/issues/16184#issuecomment-473063764).

hcassus:
Thanks for the feedback @mbhave :)
The intention was to have a starting point, and it's nice the discussion was already moving in the direction of selectively defining the init strategy. 
I'll keep following the discussion

mbhave:
cool, I'll go ahead and close this one then.

Issue #16211, opened 2019-03-12
Test the Gradle plugin against Gradle 5.3-rc-3


wilkinsona:

wilkinsona:
This is revealed what may be a regression in Gradle. I've opened https://github.com/gradle/gradle/issues/8748.

pkostrzewa:
Gradle [v5.3.0-RC3](https://github.com/gradle/gradle/releases/tag/v5.3.0-RC3) has been released. I can verify plugin with this newly released version.

PR #16294, opened 2019-03-22
Test against Gradle 5.3

dreis2211:
Hi,

as Gradle 5.3 was released a few days ago, we should probably test against it.

Cheers,
Christoph


pkostrzewa:
Tested against example provided in gh-16211 which resulted in `Main class name has not been configured and it could not be resolved` as earlier.

wilkinsona:
@pkostrzewa Thanks, but there’s no need to spend time on manual testing. This PR means that the test suite (that found the regression I reported to Gradle) will year against 5.3 automatically. The PR’s checks have passed and one of those checks is a CI build.

snicoll:
Thanks Christoph, this is now merged in `2.1.x` and `master`.

pkostrzewa:
@wilkinsona Thanks for information, actually it took no more than few minutes. But I wasn't aware of that fact.

PR #16299, opened 2019-03-25
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigFileApplicationListener.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigFileApplicationListener.java
@@ -75,7 +75,7 @@ import org.springframework.util.StringUtils;
  * properties from well known file locations. By default properties will be loaded from
  * 'application.properties' and/or 'application.yml' files in the following locations:
  * <ul>
- * <li>file:./config/:</li>
+ * <li>file:./config/</li>
  * <li>file:./</li>
  * <li>classpath:config/</li>
  * <li>classpath:</li>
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigFileApplicationListener.java
```diff
@@ -75,12 +75,12 @@
...
  * 'application.properties' and/or 'application.yml' files in the following locations:
  * <ul>
- * <li>file:./config/:</li>
+ * <li>file:./config/</li>
  * <li>file:./</li>
  * <li>classpath:config/</li>
  * <li>classpath:</li>
  * </ul>
- * The list is ordered by precedence (properties defined in locations higher in the list
+ * The list is ordered by precedence (properties defined in higher locations in the list
```
wilkinsona: I think this reads better before the proposed change

izeye: @wilkinsona I just tried to align with the following similar phrase, "defined in lower locations". If you still think it's preferred, just let me know. Then I'll revert it.

wilkinsona: Thanks, @izeye. I do prefer it as currently written rather than as proposed here.

izeye: @wilkinsona Thanks for the quick feedback! I reverted it.

Issue #15991, opened 2019-02-19
Add Spring-specific styling to the Actuator's API documentation


wilkinsona:
The Actuator's API documentation is using the default Asciidoctor styling. We should update it to use the Spring-specific styling that the reference documentation is now using.

simionAndrei:
@wilkinsona I would like to contribute in solving this issue, but I have a question about it. Where in the project structure can I find the Actuator API adoc files, since in spring-boot-project/spring-boot-actuator or in spring-boot-docs I wasn't able to identify it.  If this issue is not suitable for contribution, could you please guide me to another one?

wilkinsona:
Thanks for the offer.

The Actuator's API documentation can be found [here](https://github.com/spring-projects/spring-boot/tree/2bd3d19bee6f504b1a7e0742f6703908fa751784/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/asciidoc). This issue might be a little bit tricky for a first-time contribution, but if you're interested in having a go, the end result of [this commit](https://github.com/spring-projects/spring-boot/commit/ed3347f00da189346097faf4f948dca4b8c3f634#diff-f9e0e8080feb048de01369aec4cc234d) may help you figure out what's needed.

If you decide not to tackle this one, please let us know. You may want to look at an issue that's [labelled as ideal for contribution](https://github.com/spring-projects/spring-boot/labels/status%3A%20ideal-for-contribution) or keep an eye out for one that's specifically intended [for first-timers only](https://github.com/spring-projects/spring-boot/labels/status%3A%20first-timers-only).

DontSeeSharp:
I'll look into this, since this should be similar to #15990.

mbhave:
Closing in favor of PR #16326.

PR #16326, opened 2019-03-26
Add Spring-specific styling to Actuator's API documentation

simionAndrei:
This commit replaces the default Asciidoctor styling with
Spring specific styling.

First, we need to unzip the contents of the Spring Asciidoctor
documentation resources provided by the
`io.spring.docsresources:spring-docs-resources` distribution zip. This
is done in a `/target/refdocs` folder. We then copy all files from
`src/main/asciidoc` to the same location, and then launch the generation
process.

Closes https://github.com/spring-projects/spring-boot/issues/15991.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -14,6 +14,7 @@
 	<description>Spring Boot Actuator AutoConfigure</description>
 	<properties>
 		<main.basedir>${basedir}/../..</main.basedir>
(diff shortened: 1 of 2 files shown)


snicoll:
@simionAndrei thank you for making your first contribution to Spring Boot. This is now merged on `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/1bc47b7e749333b4fd90e0ea3ccf442072b9f2e4) as no `docs` zip was generated anymore. This also exposed a bug in asciidcotor that was fixed in 1.5.8 so I upgraded to it as well.

snicoll:
My polish broke deployment as there is now an extra `reference` directory in the generated zip: reopening to investigate and fix that.

PR #16330, opened 2019-03-26
(Obvious Fix) Add dependency management for JUnit Jupiter Aggregator

radistao:
Obvious Fix
## Enhancement

Additionally to [_JUnit Bom_](https://mvnrepository.com/artifact/org.junit/junit-bom) starting from version `5.4.0` the project provides [_JUnit Jupiter (Aggregator)_](https://mvnrepository.com/artifact/org.junit.jupiter/junit-jupiter) dependency, which simplifies dependency management for common cases, e.g. when one needs only `api+engine+params`, which seems to be the most common case. Gradle example:
  - now:
    ```
    dependencies {
         testImplementation 'org.junit.jupiter:junit-jupiter-api'
         testImplementation 'org.junit.jupiter:junit-jupiter-params'
         testRuntimeOnly 'org.junit.jupiter:junit-jupiter-engine'
    }
    ```
  - after enhancement:
    ```
    dependencies {
        testImplementation 'org.junit.jupiter:junit-jupiter'
    }
    ```
**Note:** in the PR for new added `org.junit.jupiter:junit-jupiter` dependency I reused the same _`${junit-jupiter.version}`_ as for `org.junit:junit-bom`

### More detailed information:
  
  - [_JUnit Jupiter (Aggregator)_ on Maven Central](https://mvnrepository.com/artifact/org.junit.jupiter/junit-jupiter)
    - [POM](http://central.maven.org/maven2/org/junit/jupiter/junit-jupiter/5.4.0/junit-jupiter-5.4.0.pom)
  - [JUnit subproject sources](https://github.com/junit-team/junit5/blob/master/junit-jupiter/junit-jupiter.gradle.kts)
  - [Blog post with usage example](https://sormuras.github.io/blog/2018-12-26-junit-jupiter-aggregator.html)

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -2504,6 +2504,12 @@
 				<type>pom</type>
 			</dependency>
 			<dependency>
+				<groupId>org.junit.jupiter</groupId>
+				<artifactId>junit-jupiter</artifactId>
+				<version>${junit-jupiter.version}</version>
+				<scope>test</scope>
+			</dependency>
+			<dependency>
 				<groupId>org.jvnet.mimepull</groupId>
 				<artifactId>mimepull</artifactId>
 				<version>${mimepull.version}</version>


pivotal-issuemaster:
@radistao Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16330)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16330) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@radistao This Pull Request contains an obvious fix. Signing the Contributor License Agreement is not necessary.

snicoll:
@radistao `junit-jupiter` is already provided by the BOM so there is no need to add it.

radistao:
> junit-jupiter is already provided by the BOM so there is no need to add it.

@snicoll that's exactly what i wrote in the beginning of the PR description:

> Additionally to JUnit Bom 

But the idea of this dependency is different: avoid dependencies "clogging" where it's possible. Obviously in almost all cases Jupiter is used with both `api` and `engine`, and in most cases with `params`. That's why JUnit developers introduces this ***aggregated*** dependency.

As an example from existent Spring-boot dependencies: 

  - [`spring-boot-starter`](https://github.com/spring-projects/spring-boot/blob/a2c5a7988370efd6e56f011b89e3337b3f383a62/spring-boot-project/spring-boot-dependencies/pom.xml#L309) aggregates _spring-boot-autoconfigure_, _spring-boot-starter-logging_, _snakeyaml_ etc, which already inherited by the BOM. But instead of including all these dependencies one-by-one developers may use aggregated dependency.

If _Spring-boot_ uses such approach, why _JUnit_ can't be used in such a way?

snicoll:
I know what the dependency is, we’ve actually asked for that feature.

What do you think your change does exactly? Looking at the diff it is redundant to what the bom import does. Perhaps we can take a step back and you can describe the problem you’re trying to fix and how that change helps.

radistao:
ah, i see now, my bad:
i was confused by versions, because i used spring boot `2.1.3.RELEASE`, which includes JUnit bom [`5.3.2`](https://github.com/spring-projects/spring-boot/blob/v2.1.3.RELEASE/spring-boot-project/spring-boot-dependencies/pom.xml#L124), [which didn't have `junit-jupiter`](http://central.maven.org/maven2/org/junit/junit-bom/5.3.2/junit-bom-5.3.2.pom).

But current spring-boot `2.2.0.M1` upgraded JUnit to [`5.4.0`](https://github.com/spring-projects/spring-boot/blob/v2.2.0.M1/spring-boot-project/spring-boot-dependencies/pom.xml#L139), and now it contains respective [`junit-jupiter` artifact](http://central.maven.org/maven2/org/junit/junit-bom/5.4.0/junit-bom-5.4.0.pom).

So yeah, sorry for disturbing - i mixed up the versions.

PR #16331, opened 2019-03-27
Ensure Qualifier with bean name is evaluated when using SpyBean

### LINKED ISSUE #16292: Using @SpyBean with @Qualifer does not work in Spring 1.5.x
Hello all
In the Spring boot docs [here](https://docs.spring.io/spring-boot/docs/1.5.19.RELEASE/api/org/springframework/boot/test/mock/mockito/SpyBean.html) it says

> If there is more than one bean of the requested type, qualifier metadata must be specified at field level: 
```
     @SpyBean
     @Qualifier("example")
     private ExampleService service;
```

This is not working for me on Spring boot 1.5.19.RELEASE. The simple example is 
```
@SpringBootApplication
public class QualifierspyApplication {
		SpringApplication.run(QualifierspyApplication.class, args);
	}

	@Primary
	@Bean("simpleService1")
	public SimpleService simpleService1() {
		return new SimpleService("hello 1");
	}

	@Bean("simpleService2")
	public SimpleService simpleService2() {
		return new SimpleService("hello 2");
	}
}
```
The test is 
```
@RunWith(SpringRunner.class)
@SpringBootTest
public class QualifierspyApplicationFailingTest {
	@SpyBean
	@Qualifier("simpleService2")
	private SimpleService simpleService2;

	@Test
	public void testSpy() {
		Assert.assertEquals("hello 2", simpleService2.method());
	}
}
```
The test only works if I put `@SpyBean(name = "simpleService2")`. 
If i remove `@Primary` it fails with
```
java.lang.IllegalStateException: No bean found for definition [SpyDefinition@209da20d name = '', typeToSpy = se.lolotron.qualifierspy.SimpleService, reset = AFTER]
```

You can find the example [here](https://github.com/dome313/Qualifierspy).

rhamedy:
Fixes #16292

### DIFF
--- a/spring-boot-test/src/main/java/org/springframework/boot/test/mock/mockito/MockitoPostProcessor.java
+++ b/spring-boot-test/src/main/java/org/springframework/boot/test/mock/mockito/MockitoPostProcessor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


philwebb:
Thanks for the PR, but given the limited amount of time before 1.5.x is EOL and the fact there's a work-around we're not keen to make changes in this area of the codebase. People should be considering an upgrade to 2.x at this point.

PR #16342, opened 2019-03-27
Simplify the configuration of the ProtocolHandler

### LINKED ISSUE #15035: Add Tomcat ProtocolHandlerCustomizer
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before, or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Background
Spring Boot provides a number of callback interfaces that can be used to customize the web server. A `TomcatConnectorCustomizer` can be used to customize a Tomcat `Connector`.

## Problem
Customizing the `ProtocolHandler` on the `Connector` requires creating a `TomcatConnectorCustomizer`. This is bit tedious as can be seen from the example below:

```java
@Bean
public TomcatConnectorCustomizer connectorCustomizer() {
       return (connector) -> { ((AbstractHttp11Protocol<?>) connector
				.getProtocolHandler()).setProcessorCache(250));
       };
}
```

Instead, if we added a `TomcatProtocolHandlerCustomizer` and applied those beans automatically, that would simplify the configuration to the following:

```java
@Bean
public TomcatProtocolHandlerCustomizer<AbstractHttp11Protocol<?>> processorCacheCustomizer() {
	return (handler) -> handler.setProcessorCache(250);
}
```

## Solution

Add a `TomcatProtocolHandlerCustomizer` type. Beans of  `TomcatProtocolHandlerCustomizer` type should be applied automatically to `TomcatServletWebServerFactory` and `TomcatReactiveWebServerFactory`. For `TomcatServletWebServerFactory`, this can be done [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/ServletWebServerFactoryConfiguration.java#L60) and the configuration for the reactive one can be found [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveWebServerFactoryConfiguration.java#L71). 

The `ObjectProvider` interface can be used for injecting a dependency. [Here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/artemis/ArtemisEmbeddedServerConfiguration.java#L59) is an example that can be used to inject customizers when there can be 0..n of them.

Tests for the servlet and reactive versions are [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/servlet/ServletWebServerFactoryAutoConfigurationTests.java) and [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveWebServerFactoryAutoConfigurationTests.java) respectively.

## Steps to Fix 
 
- [x] Claim this issue with a comment below and ask any clarifying questions you need
- [x] Set up a repository locally following the [Contributing Guidelines](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc)
- [x] Try to fix the issue following the steps above
- [x] Commit your changes and start a pull request.

pas2al:
Introduce the new callback interface TomcatProtocolHandlerCustomizer
that can be used to customize the ProtocolHandler on a Tomcat Connector.

Fixes #15035

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveWebServerFactoryConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveWebServerFactoryConfiguration.java
@@ -28,6 +28,7 @@ import org.springframework.boot.web.embedded.jetty.JettyReactiveWebServerFactory
 import org.springframework.boot.web.embedded.netty.NettyReactiveWebServerFactory;
 import org.springframework.boot.web.embedded.tomcat.TomcatConnectorCustomizer;
 import org.springframework.boot.web.embedded.tomcat.TomcatContextCustomizer;
+import org.springframework.boot.web.embedded.tomcat.TomcatProtocolHandlerCustomizer;
 import org.springframework.boot.web.embedded.tomcat.TomcatReactiveWebServerFactory;
 import org.springframework.boot.web.embedded.undertow.UndertowReactiveWebServerFactory;
 import org.springframework.boot.web.reactive.server.ReactiveWebServerFactory;
@@ -76,12 +77,16 @@ abstract class ReactiveWebServerFactoryConfiguration {
 		@Bean
 		public TomcatReactiveWebServerFactory tomcatReactiveWebServerFactory(
 				ObjectProvider<TomcatConnectorCustomizer> connectorCustomizers,
-				ObjectProvider<TomcatContextCustomizer> contextCustomizers) {
+				ObjectProvider<TomcatContextCustomizer> contextCustomizers,
+				ObjectProvider<TomcatProtocolHandlerCustomizer> protocolHandlerCustomizers) {
 			TomcatReactiveWebServerFactory factory = new TomcatReactiveWebServerFactory();
 			factory.getTomcatConnectorCustomizers().addAll(
 					connectorCustomizers.orderedStream().collect(Collectors.toList()));
 			factory.getTomcatContextCustomizers().addAll(
 					contextCustomizers.orderedStream().collect(Collectors.toList()));
+			factory.getTomcatProtocolHandlerCustomizers()
+					.addAll(protocolHandlerCustomizers.orderedStream()
+							.collect(Collectors.toList()));
 			return factory;
 		}
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/ServletWebServerFactoryConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/ServletWebServerFactoryConfiguration.java
@@ -35,6 +35,7 @@ import org.springframework.boot.autoconfigure.condition.SearchStrategy;
 import org.springframework.boot.web.embedded.jetty.JettyServletWebServerFactory;
 import org.springframework.boot.web.embedded.tomcat.TomcatConnectorCustomizer;
 import org.springframework.boot.web.embedded.tomcat.TomcatContextCustomizer;
+import org.springframework.boot.web.embedded.tomcat.TomcatProtocolHandlerCustomizer;
 import org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactory;
 import org.springframework.boot.web.embedded.undertow.UndertowServletWebServerFactory;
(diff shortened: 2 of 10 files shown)


mbhave:
@pas2al Thanks for the PR. I might be missing something but I don't see the new customizer being used anywhere in `TomcatServletWebServerFactory` and `TomcatReactiveWebServerFactory`. For example, the `TomcatConnectorCustomizer`s get used [here](https://github.com/spring-projects/spring-boot/blob/ea2eaa17372dd370f1e8e4148abaee6f30f39163/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatServletWebServerFactory.java#L306).

pas2al:
Can you elaborate on that?

mbhave:
Once the `tomcatProtocolHandlerCustomizers` have been set on the `TomcatServletWebServerFactory` and `TomcatReactiveWebServerFactory`, `customize` needs to be invoked for each one. This test will fail at the moment because the customize does not get called on the connector's protocol handler:

```java
TomcatServletWebServerFactory factory = getFactory();
factory.addProtocolHandlerCustomizers(
        (TomcatProtocolHandlerCustomizer) protocolHandler -> protocolHandler.setThreadPriority(23));
Tomcat tomcat = getTomcat(factory);
Connector connector = ((TomcatWebServer) this.webServer).getServiceConnectors()
        .get(tomcat.getService())[0];
AbstractProtocol protocolHandler = (AbstractProtocol) connector.getProtocolHandler();
assertThat(protocolHandler.getThreadPriority()).isEqualTo(23);
```

pas2al:
@mbhave Ok, that makes sense. I missed that, sorry! So should i call `customize` for each `ProtocolHandlerCustomizer` like this?

```
for (TomcatProtocolHandlerCustomizer customizer : this.tomcatProtocolHandlerCustomizers) {
	customizer.customize((AbstractProtocol<?>) connector.getProtocolHandler());
}
```

Would be the `customizeConnector` method a good place to do that (you pinned it in your comment above)?

mbhave:
@pas2al Yup, `customizeConnector` would be a good place to do that. I also don't think we need to force the customizer to take a `AbstractProtocol`. We can just have the `TomcatProtocolHandlerCustomizer` customize a `ProtocolHandler` instead. It would look like this:

```java
void customize(ProtocolHandler protocolHandler);
```

pas2al:
Ok, thanks for the feedback! Regarding the `ProtocolHandler`. In #15035, the description suggested to make it customizable like this:
```
@Bean
public TomcatProtocolHandlerCustomizer<AbstractHttp11Protocol<?>> processorCacheCustomizer() {
	return (handler) -> handler.setProcessorCache(250);
}
```

So passing the type to the `TomcatProtocolHandlerCustomizer` is still the way to go or do we pass just a `ProtocolHandler`? But then the Bean example above will not work.

mbhave:
Sorry, you're right. I meant we could change the generic to expect a `ProtocolHandler`. 

```java
public interface TomcatProtocolHandlerCustomizer<T extends ProtocolHandler> {
```

That way we don't need to cast it to an `AbstractProtocol` when calling `customize`.

pas2al:
Ok, should i type usages of the `TomcatProtocolHandlerCustomizer` like this `<TomcatProtocolHandlerCustomizer<ProtocolHandler>>`? For example for the usages in the `TomcatServletWebServerFactory` or do you suggest anything else?

mbhave:
I think if you could update the PR with the change mentioned [here](https://github.com/spring-projects/spring-boot/pull/16342#issuecomment-478757441) and loop over the customizers to call `customize` for each one, we can take care of anything that's missing when merging.

pas2al:
I updated the PR.

mbhave:
@pas2al Thank you for making your first contribution to Spring Boot. This is now merged into master along with this [polish commit](https://github.com/spring-projects/spring-boot/commit/ad767ca4e150bb25b2d73ce794950d51b7cb2502).

pas2al:
@mbhave Thank you very much for the opportunity and your support!

PR #16399, opened 2019-04-01
Enhance the maven plugin, add a feature that can exclude dependencies with specfic version while packaging.

JoshuaJeme:
As Maven can't exlude a dependency globally.
https://stackoverflow.com/questions/4716310/is-there-a-way-to-exclude-a-maven-dependency-globally

Setting the scope of the dependency to provided can avoid it from packaging to the fat jar, but as Spring Boot's Maven plugin include the provided dependencies by default, it doesn't work anymore.

An alternative solution is deploying an empty jar to private repo of company. Generally these jars are harmless and have a specfic version like ```999-empty```.

Using empty jars have a side effect that they were still packaged, but as they have the same version, one can exclude them easily if the maven plugin has this feature.


snicoll:
@JoshuaJeme thanks for the PR but I am not keen to introduce such a property. Excluding by version can be convenient in your specific scenario but it is too brittle IMO.

Note that we have [a separate PR](https://github.com/spring-projects/spring-boot/pull/15947) to make sure listed dependencies are excluded with war packaging (this is a bug of the plugin that we intend to fix).

JoshuaJeme:
> @JoshuaJeme thanks for the PR but I am not keen to introduce such a property. Excluding by version can be convenient in your specific scenario but it is too brittle IMO.
> 
> Note that we have [a separate PR](https://github.com/spring-projects/spring-boot/pull/15947) to make sure listed dependencies are excluded with war packaging (this is a bug of the plugin that we intend to fix).
> 
> Flagging for team attention to see what the rest of the team thinks.

Thanks for your reply. This feature is quite usefull while there are plenty of Jars to exclude, adding it to the exclude entry is a choice, but hard to maintain.

In my company there are thousands of Java applications using Spring Boot, it's a mission impossible to exclude jars one by one, but excluding some specfic version is ok.

zyhxq:
good job!

snicoll:
Excluding by version is too brittle and too specific to your use case to be added in the core plugin. That  said, we recognize that the current features is not great if you have a large amount of dependencies to exclude. IMO, restructuring your build so that it's not the case should be preferred but that might not always be possible.

We've decided to give it that feature another look, I've created #16563. Thanks for the PR, in any case.

Issue #16200, opened 2019-03-11
Rework o.s.b.a.couchbase.CouchbaseConfiguration to avoid the need for bean method proxying


wilkinsona:

ayudovin:
@wilkinsona, I want to work on it. Please let me know if I can start working on it?

wilkinsona:
@ayudovin You are more than welcome to start working on it, but please be aware that I don't think this is going to be an easy one to fix. I suspect that changes may be needed in Spring Data.

ayudovin:
@wilkinsona, I want to clarify this task. As far as I understand this class `CocuhbaseConfigurashion` has `@Bean` methods which called directly and these methods should be replaced on bean injection.
is it correct?

wilkinsona:
Yes, that’s right. To be suitable for having bean method producing disabled, calling `@@Bean` methods directly must be avoided. That may require some more complex restructuring than just moving to injection.

ayudovin:
From my point of view, We can create two extra configuration classes. The first one `CouchbaseEnvironmentComfig` for building `CouchbaseEnvironment` bean. The second one is `CouchbaseClusterConfig` for building `Cluster` bean (this config class will inject  `CocuhbaseEnvironment` bean from `CouchbaseEnvironmentComfig` using constructor injection). 
The last step is `CouchbaseConfiguration` will just inject `Cluster` via constructor injection.
@wilkinsona, what do you think about it?

wilkinsona:
@ayudovin It's hard to give accurate feedback without seeing some code. Can you please share a commit on a branch or a draft PR that I can look at?

ayudovin:
yes, certainly, I'll prepare a pull request.

snicoll:
This has been superseded by #20533

PR #16403, opened 2019-04-01
Rework o.s.b.a.couchbase.CouchbaseConfiguration to avoid the need for bean method proxying

ayudovin:
this [task](https://github.com/spring-projects/spring-boot/issues/16200)


snicoll:
@ayudovin can you please squash and push force rather than having merge commit here. Getting only your changes speed up the review process.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseConfiguration.java
```diff
@@ -44,97 +36,27 @@
 
 	private final CouchbaseProperties properties;
 
-	public CouchbaseConfiguration(CouchbaseProperties properties) {
```
wilkinsona: This, along with other changes to `public` and `protected` methods on this class, are breaking API changes that we'd prefer to avoid if at all possible. If it's not possible, we'll need to see if we can provide a migration path via deprecation.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/couchbase/CouchbaseConfigurerAdapterConfiguration.java
```diff
@@ -37,16 +41,21 @@
...
+		this.cluster = cluster;
 	}
 
 	@Bean
 	@ConditionalOnMissingBean
 	public CouchbaseConfigurer springBootCouchbaseConfigurer() throws Exception {
-		return new SpringBootCouchbaseConfigurer(
-				this.configuration.couchbaseEnvironment(),
-				this.configuration.couchbaseCluster(),
+		return new SpringBootCouchbaseConfigurer(this.couchbaseEnvironment, this.cluster,
```
wilkinsona: I'm worried that this may introduce some subtle bugs. Previously, the four components passed into `SpringBootCouchbaseConfigurer` all came came from the same source. That's no longer guaranteed to be the case. In particular using the cluster from any source and the cluster info from `CouchbaseConfiguration` doesn't feel quite right. I'm not sure what the answer is yet.

ayudovin:
@wilkinsona, in this case, I think we can close this pull request.

snicoll:
@ayudovin thank you for trying, we'll give it another go.

PR #16419, opened 2019-04-02
Avoid using deprecated @SolrDocument#solrCoreName

dreis2211:
Hi,

just noticed this deprecation warning that doesn't seem to be too hard to fix. ;-)

Cheers,
Christoph


philwebb:
Thanks!

PR #16446, opened 2019-04-03
Polish embedded tomcat setup

za-rudeboy:
Small code cleanups such as removing a redundant class cast, null check and making fields final that can be final

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatReactiveWebServerFactory.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatReactiveWebServerFactory.java
@@ -64,7 +64,7 @@ public class TomcatReactiveWebServerFactory extends AbstractReactiveWebServerFac
 
 	private File baseDirectory;
 
-	private List<Valve> engineValves = new ArrayList<>();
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatReactiveWebServerFactory.java
```diff
@@ -60,11 +60,11 @@
 	/**
 	 * The class name of default protocol used.
 	 */
-	public static final String DEFAULT_PROTOCOL = "org.apache.coyote.http11.Http11NioProtocol";
+	private static final String DEFAULT_PROTOCOL = "org.apache.coyote.http11.Http11NioProtocol";
```
mbhave: I don't think we can make this `private` in case someone else is relying on it.

za-rudeboy: You're right, sorry. I was thinking like someone working on their own final codebase as opposed to one that's used by others as a library :)

za-rudeboy:
Reverted the change

snicoll:
@za-rudeboy thank you for making your first contribution to Spring Boot. Your contribution is now merged on `master`.

Issue #16401, opened 2019-04-01
Investigate configuration property binding performance with extremely large input files


wilkinsona:
Courtesy of @martinvisser, we have [an example](https://github.com/martinvisser/config-properties-example) of the binder still being slow in 2.2 snapshots relative to 1.5's performance. The input YAML file is 7500 lines long so considerably bigger than the case we investigated with Initializr.

dreis2211:
Most of the time is spent in `ConfigurationPropertyName.elementEquals` still. Or `ConfigurationPropertyName.isAncestorOf` depending on how you see it. Here's a screenshot from an application that simply tries to bind the large file:

![grafik](https://user-images.githubusercontent.com/6304496/55491705-48233380-5636-11e9-82b8-8bc2de81f93e.png)
(This is a snapshot in between, so don't trust the overall invocations or timings. The share of CPU usage is more important here)

andresetevejob:
I want to investigate on this issue.can i do?

dreis2211:
I'm currently working already on some improvements as well, @wilkinsona . See the referenced issue #16447 for one of those. I'm currently working on another one already, but I'm not confident enough in the solution yet.

dreis2211:
Other than the already provided PRs I'm still trying around on this and found one interesting place that you might want to look at. When simply assigning the cache key in `SpringIterableConfigurationPropertySource.getCache()` instead of copying it, benchmarks yield the following results:

|       | Baseline |  New  |
| ----- | -------: | ----: |
|       |    13.386 | 8.062 |
|       |    12.658 | 8.121 |
|       |    12.530 | 8.365 |
|       |    12.147 | 8.813 |
|       |    12.016 | 8.401 |
|       |    12.316 | 9.393 |
|       |    11.928 | 8.388 |
|       |    11.754 | 8.242 |
|       |    11.663 | 8.674 |
|       |    11.704 | 8.622 |
|  Mean |    12.210 | 8.508 |
| Range |    1.723 | 1.331 |

Unfortunately, this is meant to fix #13344 and can thus not be removed at the moment. For the given example this means that AbstractSet.equals() is called on Sets with 5000 entries basically every time someone interacts with `SpringIterableConfigurationPropertySource.getCache()` - which is the majority of interaction.

[...]

PR #16447, opened 2019-04-03
Optimize JavaBeanBinder

dreis2211:
Hi,

I'm currently working on some improvements for #16401. While I have some more stuff that I want to try out, there is an optimization possibility in `JavaBeanBinder` that skips some costly checks, which I want to get out already. This shows the following results when I test it with @wilkinsona's benchmark harness:

|       | Baseline |  New  |
| ----- | -------: | ----: |
|       |    14.824 | 12.224 |
|       |    13.320 | 12.324 |
|       |    14.376 | 13.065 |
|       |    14.902 | 13.218 |
|       |    15.523 | 12.692 |
|       |    14.849 | 14.740 |
|       |    13.884 | 12.450 |
|       |    14.202 | 15.886 |
|       |    14.793 | 12.219 |
|       |    15.091 | 12.556 |
|  Mean |    14.576 | 13.137 |
| Range |    2.203 | 3.667 |

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/JavaBeanBinder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/JavaBeanBinder.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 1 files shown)


philwebb:
I still need to review this but I'll tentatively target to 2.2.x

mbhave:
This change makes sense to me. We use `hasKnownBindableProperties` to determine if the `getter` should be called to get the existing value and that only needs to be done if there is a supplier for the existing value.

wilkinsona:

Issue #16337, opened 2019-03-27
Add property for Mongo auto index creation


bclozel:
As of Spring Data MongoDB 2.2.x, using `@Indexed` on a `@Document` causes the following message to appear in the logs:

```
WARN 1471 --- [  restartedMain] eactiveMongoPersistentEntityIndexCreator : Automatic index creation will be disabled by default as of Spring Data MongoDB 3.x.
    Please use 'MongoMappingContext#setAutoIndexCreation(boolean)' or override 'MongoConfigurationSupport#autoIndexCreation()' to be explicit.
    However, we recommend setting up indices manually in an application ready block. You may use index derivation there as well.

    > -----------------------------------------------------------------------------------------
    > @EventListener(ApplicationReadyEvent.class)
    > public void initIndicesAfterStartup() {
    >
    >     IndexOperations indexOps = mongoTemplate.indexOps(DomainType.class);
    >
    >     IndexResolver resolver = new MongoPersistentEntityIndexResolver(mongoMappingContext);
    >     resolver.resolveIndexFor(DomainType.class).forEach(indexOps::ensureIndex);
    > }
    > -----------------------------------------------------------------------------------------
```

[More details on that in the reference documentation](https://docs.spring.io/spring-data/mongodb/docs/2.2.x-SNAPSHOT/reference/html/#mapping-usage).

We should create a Spring Boot configuration property to enable/disable this configuration to make the transition easier for developers.

rhamedy:
@bclozel Is this issue open for contribution?

philwebb:
@rhamedy The issue isn't assigned or pending any design work but it looks as if @ayudovin might have beaten you to it

bclozel:
Hey @mp911de @christophstrobl, does this issue sounds right to you?
What should the users do ideally? Remove `@Indexed` annotations from their model classes?
If we set that property to true, people will still get that WARN message at startup?
How should we ease the transition here?

mp911de:
Adding a property to control index creation sounds about right. We will disable auto-index creation in one of the next releases. The background is that users don't get much control over the actual index creation time if entity types are discovered during runtime and not known at startup time.  Therefore, we decided to deprecate the feature. We introduced a configuration flag `autoIndexCreation` on `MongoMappingContext`.

`@Indexed` is still useful and we also introduced utilities to derive and create indexes in a more controlled manner.

mbhave:
Thanks @mp911de. I'll close this in favor of PR #16454.

[...]

PR #16454, opened 2019-04-04
Add property for Mongo auto index creation

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/16337)


mbhave:
Thanks again, @ayudovin. This is now on master along with [this](https://github.com/spring-projects/spring-boot/commit/bf6f9b4c43355b92a3c5de4bec1af0e4deaf3ec5) polish commit. I changed the [test](https://github.com/spring-projects/spring-boot/blob/bf6f9b4c43355b92a3c5de4bec1af0e4deaf3ec5/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/mongo/MongoDataAutoConfigurationTests.java#L138) to set `false` because true is the default and the test would've passed even without setting the property to true explicitly.

ayudovin:
@mbhave, thank you!
I agree with you, it makes sense.

Issue #16393, opened 2019-03-31
AutoConfiguration support for Data Mongo ReactiveGridFsTemplate


hantsy:
The latest Spring Data Mongo has brought Reactive support for GridFs, but Spring Boot 2.2.0.M1
is lack of an `AutoConfiguration` class to activate it, I have to configure it manually at the moment.

https://github.com/hantsy/spring-reactive-sample/blob/master/boot-data-mongo-gridfs/src/main/java/com/example/demo/DemoApplication.java#L35-L43

ayudovin:
@wilkinsona, I would like to work on it. Please let me know if I can start working on it?

philwebb:
@ayudovin Much appreciated. The issue isn't assigned and it doesn't look like it should be too involved so please go head.

mbhave:
Closing in favor of PR #16467.

PR #16467, opened 2019-04-05
AutoConfiguration support for Data Mongo ReactiveGridFsTemplate

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/16393)


ayudovin:
Test failure seems unrelated

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/mongo/MongoReactiveDataAutoConfiguration.java
```diff
@@ -85,4 +87,12 @@ public MappingMongoConverter mappingMongoConverter(MongoMappingContext context,
 		return mappingConverter;
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
+	public ReactiveGridFsTemplate reactiveGridFsTemplate(
+			ReactiveMongoDatabaseFactory reactiveMongoDbFactory,
+			MappingMongoConverter mappingMongoConverter) {
+		return new ReactiveGridFsTemplate(reactiveMongoDbFactory, mappingMongoConverter);
```
mp911de: Ideally, we should configure `ReactiveGridFsTemplate ` with `DataBufferFactory` to share potentially resources.

ayudovin: `ReactiveGridFsTemplate` constructor has 4 parameters:

- `DataBufferFactory`

- `ReactiveMongoDatabaseFactory`

- `MappingMongoConverter`

- `bucket`

What value of bucket should be in this case?

mp911de: Defaults to `null`, see https://github.com/spring-projects/spring-data-mongodb/blob/master/spring-data-mongodb/src/main/java/org/springframework/data/mongodb/gridfs/ReactiveGridFsTemplate.java#L68.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/mongo/MongoReactiveDataAutoConfigurationTests.java
```diff
@@ -55,4 +56,10 @@ public void backsOffIfMongoClientBeanIsNotPresent() {
 				.doesNotHaveBean(MongoReactiveDataAutoConfiguration.class));
 	}
 
+	@Test
+	public void gridFsTemplateExists() {
+		this.contextRunner.run((context) -> assertThat(context)
+				.hasSingleBean(ReactiveMongoTemplate.class));
```
snicoll: This looks like you're checking for the wrong bean there.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/mongo/MongoReactiveDataAutoConfiguration.java
```diff
@@ -85,4 +89,20 @@ public MappingMongoConverter mappingMongoConverter(MongoMappingContext context,
...
+			ReactiveMongoDatabaseFactory reactiveMongoDbFactory,
+			MappingMongoConverter mappingMongoConverter,
+			DataBufferFactory dataBufferFactory) {
+		return new ReactiveGridFsTemplate(dataBufferFactory, reactiveMongoDbFactory,
+				mappingMongoConverter, null);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public DataBufferFactory dataBufferFactory() {
```
snicoll: Can we flip those two methods? Required dependencies are defined before so I'd like to keep it that way.

snicoll:
@ayudovin thanks for the feedback

PR #16476, opened 2019-04-06
Fix compilation error

dreis2211:
Hi,

this PR fixes a compilation error caused by a missing implementation in `AbstractWebFluxEndpointHandlerMapping`  of `Set<PathPattern> getMappingPathPatterns(RequestMappingInfo mapping)` that seems to have been introduced recently.

Cheers,
Christoph


dreis2211:
Test failure could be related, but I have the weird feeling this might be related to #16410 .

snicoll:
Me too even though we had green build after this was merged. Nothing jumps at me that might have changed in framework in the meantime.

Paging @sdeleuze.

snicoll:
@dreis2211 I might have fixed it in https://github.com/spring-projects/spring-boot/commit/84735f58f337975ef1b5dfb4bff9fdbf29e4287e - We've harmonized the context hierarchy recently and those tests were calling refresh twice (which is not something the regular hierarchy allows). Arguably, that was a problem in the test in the first place so I've moved them to use the `ApplicationContextRunner` utility.

dreis2211:
@snicoll Thanks for the heads up.

PR #16478, opened 2019-04-06
Fix Thymeleaf deprecations

dreis2211:
Hi,

just noticed these two deprecation warnings that could be avoided.

Cheers,
Christoph


snicoll:
Thanks again Christoph!

PR #16480, opened 2019-04-07
Separate compound properties in appendices

dreis2211:
Hi,

I just noticed that compound properties in the current [appendices docs](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/html/appendix.html#common-application-properties) look a bit weird as they seem to be missing newlines and therefore make them hard to read. See the following screenshot for a better explanation:

grafik

This PR is an attempt to fix this. To be honest I struggled a bit with asciidoc and the only way I found needs a change in https://github.com/spring-io/spring-doc-resources/pull/12 where I'm not sure about its side-effects.

Anyhow. If I apply the changes the table looks like this:
grafik

Let me know what you think. 
Cheers,
Christoph


dreis2211:
Thanks @bclozel for merging this. But I think we need spring-doc-resources in version 0.1.2 in order to include the mentioned style changes.

bclozel:
Yes, I haven’t released it yet.
We need to fix other things there as well. I’ve created #16690 to track that.
Thanks for your contributions!

Issue #16484, opened 2019-04-08
@WebMvcTest and @WebAppConfiguration behave differently with ServletContext resources


lorenzodee:
I'm porting a Spring app to Spring Boot 2.0.x and observed a different behavior with `ServletContext` resources.

Here's the non-boot test, and works as expected.
```java
@WebAppConfiguration
@RunWith(SpringRunner.class)
public class SampleNonBootTest {

	@EnableWebMvc
	@Configuration
	static class TilesConfig {
		@Bean
		public TilesConfigurer tilesConfigurer() {
			TilesConfigurer tilesConfigurer = new TilesConfigurer();
			tilesConfigurer.setDefinitions("/WEB-INF/views/tiles.xml");
			tilesConfigurer.setCheckRefresh(true);
			return tilesConfigurer;
		}

		// @Bean
		// tilesViewResolver not relevant for this issue
	}
	
	@Autowired
	private WebApplicationContext context;

	private MockMvc mockMvc;

	@Before
	public void setUp() throws Exception {
		mockMvc = MockMvcBuilders.webAppContextSetup(context).build();
	}

	@After
	public void tearDown() throws Exception {
	}

	@Test
	public void test() throws Exception {
		mockMvc.perform(get("/"))
			.andExpect(status().isNotFound());
	}

}
```

And here's the boot test, but fails to load `ApplicationContext` due to a missing resource `WEB-INF/views/tiles.xml`. But the non-boot test succeeds, and the file exists.
```java
@WebMvcTest
@RunWith(SpringRunner.class)
public class SampleBootTest {
	
	@Autowired
	private MockMvc mockMvc;

	@TestConfiguration
	static class TilesConfig {
		@Bean
		public TilesConfigurer tilesConfigurer() {
			TilesConfigurer tilesConfigurer = new TilesConfigurer();
			tilesConfigurer.setDefinitions("/WEB-INF/views/tiles.xml");
			tilesConfigurer.setCheckRefresh(true);
			return tilesConfigurer;
		}

		// @Bean
		// tilesViewResolver not relevant for this issue
	}

	@Before
	public void setUp() throws Exception {
		// Failed to load ApplicationContext
		// No URL for ServletContext resource [/WEB-INF/views/tiles.xml]
	}

	@After
	public void tearDown() throws Exception {
	}

	@Test
	public void test() throws Exception {
		mockMvc.perform(get("/"))
			.andExpect(status().isNotFound());
	}

}
```

[...]

PR #16485, opened 2019-04-08
@WebMvcTest ignores base path configured via @WebAppConfiguration

lorenzodee:
Set resource base path to make `@WebMvcTest` behave the same as `@WebAppConfiguration` and `@SpringBootTest`

This fixes issue #16484 .

Upon further investigation, when running tests that use `@WebAppConfiguration` or `@SpringBootTest`, the `resourceBasePath` (in `WebMergedContextConfiguration`) was being set to `src/main/webapp` for servlet-based environments. But when using `@WebMvcTest`, the `resourceBasePath` was being set to `""` (empty string). This causes the difference in behavior.

In the sample tests in #16484 , the empty string `resourceBasePath` causes the resource to be missing, since it was not looking under `src/main/webapp`.

### DIFF
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/WebMvcTestContextBootstrapper.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/WebMvcTestContextBootstrapper.java
@@ -17,8 +17,10 @@
 package org.springframework.boot.test.autoconfigure.web.servlet;
 
 import org.springframework.boot.test.context.SpringBootTestContextBootstrapper;
+import org.springframework.core.annotation.AnnotatedElementUtils;
 import org.springframework.test.context.MergedContextConfiguration;
 import org.springframework.test.context.TestContextBootstrapper;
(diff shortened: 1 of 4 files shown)


pivotal-issuemaster:
@lorenzodee Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16485)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16485) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@lorenzodee Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16485)!

philwebb:
Thanks, I'm tempted to look at extracting a method in `SpringBootTestContextBootstrapper` so we can share that identical logic. We should ideally also try to add a test for this.

lorenzodee:
Thanks Phil. I'd like to help, if I may. Would it be proper to add a test in `WebMvcTestAutoConfigurationIntegrationTests` that would check for a resource under `src/main/webapp`?

lorenzodee:
Ended up adding tests in separate test classes. Similar to what was done in `SpringBootMockServletContextTests`. Hope this helps.

PR #16507, opened 2019-04-10
Call MimeTypeUtils::parseMimeType earlier to improve performance

aftersss:
Consider that `MimeTypeUtils.parseMimeType`'s performance is not very good, `MimeTypeUtils::parseMimeType` should be called earlier to avoid the cost. See also: https://github.com/spring-projects/spring-framework/pull/22598


snicoll:
@aftersss thank you for making your first contribution to Spring Boot. This is merged in `2.1.x` and `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/2448efc0281e0356de7b1a5ba3e695ef1615d602).

PR #16513, opened 2019-04-10
Improve Spring Session DefaultCookieSerializer auto-configuration

vpavic:
Spring Session's own configuration support (i.e. `SpringHttpSessionConfiguration`) will configure the default `DefaultCookieSerializer` with `rememberMeRequestAttribute` if `SpringSessionRememberMeServices` bean has been detected in the application context.

In contrast, Spring Boot's auto-configured `DefaultCookieSerializer` does not do this which results in a different out-of-the-box experience for users that rely on Spring Session's remember-me integration.

This PR improves Spring Session `DefaultCookieSerializer` auto-configuration to match Spring Session's behavior and make the auto-configured `DefaultCookieSerializer` aware of `SpringSessionRememberMeServices` bean.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
@@ -56,6 +56,7 @@ import org.springframework.core.type.AnnotationMetadata;
 import org.springframework.session.ReactiveSessionRepository;
 import org.springframework.session.Session;
 import org.springframework.session.SessionRepository;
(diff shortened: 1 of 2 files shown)


mbhave:
Should we do this in 2.1.x?

vpavic:
I see no issues with that. It's a small change that affects only user the rely on Spring Session's remember-me integration, and for those it feels like a bug. But after all, auto-configuration for `DefaultCookieSerializer` itself was introduced in a maintenance release (`2.0.7.RELEASE`).

snicoll:
Thanks @vpavic. I've polished your contribution in https://github.com/spring-projects/spring-boot/commit/7b9471836b5f947a56189fafd01da3e203632094 - Please note we've reviewed the auto-configuration to avoid having a constructor when a given dependency is used in one place. Rather, we put that in the `@Bean` definition itself so that the context does not have to resolve it if the bean doesn't have to be created.

PR #16527, opened 2019-04-11
Fix package of java.time.Duration in documentation

adammichalik:
The discussed Duration class is actually in the `java.time` package.


snicoll:
@adammichalik spot on! Thank you very much for making your first contribution to Spring Boot. This is now merged in `2.1.x` and `master`.

PR #16539, opened 2019-04-11
Actuator /configprops shows Duration properties as { "units": [ "SECONDS", "NANOS" ] }

### LINKED ISSUE #16526: Actuator /configprops shows Duration properties as { "units": [ "SECONDS", "NANOS" ] }
Spring Boot 2.1.4.RELEASE.

Actuator `/configprops` endpoint shows this property class:

```java
@ConfigurationProperties("demo.props")
@Data
public class DemoProperties {
    private Duration duration = Duration.ofMinutes(10);
}
```

like so:

```js
{
  "contexts": {
    "application": {
      "beans": {
        "demo.props-com.example.demo.DemoProperties": {
          "prefix": "demo.props",
          "properties": {
            "duration": {
              "units": [
                "SECONDS",
                "NANOS"
              ]
            }
          }
        }
      }
    }
  }
}
```

The actual value is not show in any way. I would expect the actual value in the ISO-8601 format or the Spring Boot "human-readable" format.

dreis2211:
Hi,

this is an attempt to fix #16526 .

Feel free to decline the PR if you had a different solution in mind.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
@@ -57,7 +57,6 @@ import org.springframework.boot.context.properties.ConfigurationBeanFactoryMetad
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.ApplicationContextAware;
-import org.springframework.core.convert.ConversionService;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
 
@@ -86,8 +85,6 @@ public class ConfigurationPropertiesReportEndpoint implements ApplicationContext
 	private ApplicationContext context;
 
 	private ObjectMapper objectMapper;
-
-	private ConversionService conversionService;
 
 	@Override
 	public void setApplicationContext(ApplicationContext context) throws BeansException {
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2018 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.actuate.context.properties;
 
+import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
 import java.util.Collections;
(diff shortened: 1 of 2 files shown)


philwebb:
Wow that was quick! Thanks @dreis2211 !

dreis2211:
Let me fix the checkstyle issue. Sorry

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
```diff
@@ -171,6 +176,7 @@ private ConfigurationBeanFactoryMetadata getBeanFactoryMetadata(
 	 */
 	protected void configureObjectMapper(ObjectMapper mapper) {
 		mapper.configure(SerializationFeature.FAIL_ON_EMPTY_BEANS, false);
+		mapper.configure(SerializationFeature.WRITE_DATES_AS_TIMESTAMPS, false);
```
snicoll: Is that related to the support of `Duration`?

dreis2211: Yes and no - this depends on how you want the durations to be displayed. See https://github.com/FasterXML/jackson-datatype-jsr310/blob/master/src/main/java/com/fasterxml/jackson/datatype/jsr310/ser/DurationSerializer.java#L63

dreis2211: @snicoll I pushed a revised version that registers a module. I've tried this actually before, but it didn't work because I called it before the serializerFactory is overwritten again in `applySerializationModifier`. Anyway, this should be easier now. And btw: there would be a way of calling `ObjectMapper.findAndRegisterModules` but that might register more than you want and is sort of expensive.

snicoll:
While trying to backport this to `2.1.x`, I've noticed that `JavaTimeModule` is not available so I've changed something to include it only when the relevant module is on the classpath. It works on `master` because the new version of `neo4j-ogm-api` has a compile dependency on it. 

I do wonder though if that's what we want to do for the endpoint. Given the endpoint is focused on `@ConfigurationProperties` and we have a dedicated format for `Duration`, Maybe we should use our converter rather than the standard `Duration` format.

wilkinsona:
I don't like the idea of the actuator's output format changing depending on what's on the classpath. If we need the `JavaTimeModule` to always be available to produce predictable output then I think we should make it a runtime dependency of the module.

The format for a duration in `application.properties` is, from a user's perspective, intended only to be human readable and writable. The format in the Actuator's response is intended to both human and machine readable. As such, I think we should use a standard format for the duration, rather than our own. Assuming I've read the code correctly, this proposal will result in an ISO-8601 seconds based representation which sounds good to me.

snicoll:
Thanks for the feedback and good point on machine readable. Given the side effect of adding the module, I am moving this fix to `2.2.x`.

dreis2211:
Anything I can do to help here, @snicoll ?

snicoll:
@dreis2211 if you want to push force an update that adds the jackson module in runtime scope, feel free to do it but I am more than happy to do that as part of the polish later today.

dreis2211:
I would be only available this evening, so feel free to do it. Highly appreciated - as usual :)

snicoll:
Thanks again @dreis2211!

PR #16544, opened 2019-04-12
Add annotation to support registering Jackson key serializer/deserializers

### LINKED ISSUE #16469: Add annotation to support registering Jackson key serializer/deserializer
My use case for using modules in #16463 was to allow registration of a key serializer. 

It would be useful if there was an annotation similar to `@JsonComponent` that supported key serializers and deserializers.

maly7:
Fixes #16469 

Updates `@JsonComponent` so that serializers/deserializers can be registered as key based. Also allows for explicitly declaring the classes the component can handle.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jackson/JsonComponent.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jackson/JsonComponent.java
@@ -25,15 +25,16 @@ import java.lang.annotation.Target;
 import com.fasterxml.jackson.databind.JsonDeserializer;
 import com.fasterxml.jackson.databind.JsonSerializer;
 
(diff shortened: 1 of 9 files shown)


philwebb:
Nice. Thanks @maly7!

philwebb:
Thanks @maly7 for your first contribution to Spring Boot! This has now been squashed and merged to master along with a [polish commit](https://github.com/spring-projects/spring-boot/commit/bf633fc5754f455cd9317f095b9cda967b5ed31f). I went back and forth a few times trying different attribute names and ended up renaming them to `type` and `scope`. I quite like `type` because it aligns with the parameters that the `addSerializer`/`addDeserializer` methods take. I'm less happy with `scope` since it could be confused with Spring's `@Scope` annotation, but I couldn't come up with anything better.

Thanks again!

maly7:
Thanks, @philwebb!

Issue #12382, opened 2018-03-07
additionalProperty values should not be required when using the build-info goal of the spring-boot-maven-plugin


eztinkerdreams:
I would like to make an enhancement request. Currently when using the build-info goal of the spring-boot maven plugin, the plugin will generate an error if any of the 'additionalProperties' evaluate to null or empty strings. 

Can this behavior made optional, or just removed all together? I think it is perfectly acceptable to have a property in the file with no value. I've a use case for this in my build system.

wilkinsona:
> the plugin will generate an error if any of the 'additionalProperties' evaluate to null or empty strings.

I believe you'll only get an error if the value is `null`. `null` values are rejected as Java's `Properties` class does not support `null` values. If we didn't fail-fast the diagnostics would be poor (see https://github.com/spring-projects/spring-boot/issues/6724).

Removing the failure with a `null` value would be a breaking change for anyone who is relying on it catching a problem with their build configuration. This would have to wait for 3.0. The alternative would be to add a configuration option to control the behaviour. We could coerce `null` to an empty value or we could remove the property altogether. This would add complexity and that would need to be justified.

What's your usecase for wanting to tolerate `null` values?

eztinkerdreams:
@wilkinsona I'm using the [CI Friendly Versions Properties](https://maven.apache.org/maven-ci-friendly.html) of maven. I pass this in as an external argument in my CI pipeline. In the process of creating a 'release' I don't use changelist, so it gets passed in as empty. 

```
mvn spring-boot:build-info "-Dchangelist=${SOME_VAR}
```
I am able to work out a way to pass in an empty value. If I try to pass in escaped quotes, it doesn't error, but treats the value as literally two double quotes. So all other strings where I use this value get concatenated/formatted with two double quotes as the value.
For example, given my pom.xml is configured with this `<version>${revision}${sha1}${changelist}</version>`

```
mvn spring-boot:build-info -Dgit.commit.id.abbrev=abcde "-Dchangelist=\"\""
```
produces:
```
build.version=v1.1.8-abcde""
build.changelist=""
```

Obviously there are a variety of other ways for me to solve this, and I don't really need suggestions on how to code around this, there are probably dozens. 

What would be ideal is if I could configure the plugin to treat nulls as empty strings. i.e if a value is null, write it out as an empty property value, instead of throwing an error.

Pretty low priority, I can work around it.

```
<plugin>
                <groupId>org.springframework.boot</groupId>
[...]

PR #16583, opened 2019-04-16
Skip additional properties null values for spring-boot-maven-plugin

tdanylchuk:
In order to support null values configured in maven.
It solves case when maven property placeholder is used and corresponding property is missed during the build.

Example:
I have different profiles in by maven build configuration and some of them define particular maven parameter which I want to add to /info endpoint.
Problem is that for those profiles which are not define this property, build fails with ```NullAdditionalPropertyValueException```, in this case I think it's better just skip value instead of build fail.
The best solution is to have possibility to define default value, but maven does not support this feature.


philwebb:
The `NullAdditionalPropertyValueException` is far from ideal, but I'm also not convinced that we should skip unrecognized placeholders.

My concern is that someone could have a typeo in a property and have no indication that it was dropped.

@tdanylchuk Does defining a default property with an empty value and overriding it in profile solve your use-case? Or do you specifically need the property key to disappear from the info endpoint?

tdanylchuk:
@philwebb thx for the quick answer.

Actually I came to the same concern that you described. And originally I was thinking about 2 options:
- Not just skip null values, but log warn message to the build log regarding skipped values (didn't find logs in project, so I assumed that it's not an appropriate approach)
-  Just don't skip null values, return null to /info instead (In this case I need to change dependent project, and for some reason exception is thrown in such case, so I didn't propose that solution)

As for empty value: I have complicated case where this empty property value could override existent property, since I'm using maven mixin plugin for merging two poms (or more). 
But for simple case with one pom with profiles, it should work out (imho this solution requires adding redundant code to my pom.xml)

snicoll:
This looks like a duplicate of #12382 - This is standard behaviour of Maven and I am not keen to deviate from that by implementing something custom in our plugin.

tdanylchuk:
I consider this as not "custom thing", but as "compatibility" with standard Maven behavior, where we cannot define default values for placeholders, and missed values just should not fail the build.

In the end of the day "tools" should help developers, but not force them to add workarounds to their poms imho. It will help at least me and that developer from provided ticket so solve an obstacle and as for me it does not change plugin functionality that much either adds additional complexity.

philwebb:
> I consider this as not "custom thing", but as "compatibility" with standard Maven behavior, where we cannot define default values for placeholders, and missed values just should not fail the build.

Your specific requirement is that missing values don't fail the build, but another user may well prefer that the build fails if they make a typo.

I appreciate that this change makes things better for some users, but on balance I think it's best if we are consistent with our handling of maven properties. This really feels like a limitation of Maven and not something that we should attempt to work-around here.

Regardless, thanks for taking the time to put together the pull-request and I hope you can find a solution that works in your build. I think default value support would be nice addition to Maven, it's a shame they don't yet support it.

PR #16604, opened 2019-04-19
Auto-configure ReactiveStringRedisTemplate if necessary

dlmu-lq:
…AutoConfiguration

Like the @Bean-annotated method `stringRedisTemplate()` in @Configuration-annotated class `RedisAutoConfiguration,` add a @Bean-annotated method `reactiveStringRedisTemplate()` in class `RedisReactiveAutoConfiguration,` which avoids unnecessary repeated configuration when used in different applications.


snicoll:
@dlmu-lq thank you for making your first contribution to Spring Boot. This is now merged in `master`.

PR #16607, opened 2019-04-19
Instantiate default ReactiveStringRedisTemplate bean in RedisReactive…

dlmu-lq:
…AutoConfiguration

Like the @Bean-annotated method `stringRedisTemplate()` in @Configuration-annotated class `RedisAutoConfiguration`, add a @Bean-annotated method `reactiveRedisTemplate()` in class `RedisReactiveAutoConfiguration`, which avoids unnecessary repeated configuration when used in different applications.


snicoll:
@dlmu-lq please stop opening duplicate PRs for this. You don't need to base that on `2.1.x` in any case, we'll figure out where we decide to merge this change if necessary.

PR #16613, opened 2019-04-20
UndertowWebServer's logger is wrong

liaozan:
fix error logger


philwebb:
Good spot. Thanks!

snicoll:
@liaozan thank you for making your first contribution to Spring Boot. This is now merged in `2.1.x` and `master`.

PR #16615, opened 2019-04-20
Support programmatic lazy-int exclusion

tkvangorder:

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/LazyInitializationBeanFactoryPostProcessor.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/LazyInitializationBeanFactoryPostProcessor.java
@@ -16,6 +16,10 @@
 
 package org.springframework.boot;
 
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+
 import org.springframework.beans.BeansException;
 import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.BeanFactoryPostProcessor;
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the PR, @tkvangorder. We might explore the possibility of the callback interface extending `Predicate<Class<?>>` as part of merging this.

tkvangorder:
@wilkinsona A predicate is a good idea and gives downstream projects a little more control. I made some changes to support a predicate, is this what you had in mind?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/LazyInitializationBeanFactoryPostProcessor.java
```diff
@@ -36,8 +57,16 @@
...
-		for (String name : beanFactory.getBeanDefinitionNames()) {
-			BeanDefinition beanDefinition = beanFactory.getBeanDefinition(name);
+
+		List<EagerLoadingBeanDefinitionPredicate> eagerPredicateList = getEagerLoadingClassesFromContext(
+				beanFactory);
+
+		for (String beanName : beanFactory.getBeanDefinitionNames()) {
+			BeanDefinition beanDefinition = beanFactory.getBeanDefinition(beanName);
+			if (eagerPredicateList.stream().anyMatch(
+					(predicate) -> predicate.test(beanFactory.getType(beanName)))) {
```
wilkinsona: I believe this may lead to factory beans bean instantiated so we need to avoid it. We probably need something that's somewhat similar to `BeanTypeRegistry`. Making that class public doesn't feel like a good idea, but it's also complicated enough that copying it doesn't feel like a good idea either. In short, a change is needed here but I don't know what it should be. I wonder if Framework could provide an API for getting a bean's type that gives up and returns null rather than instantiating a factory bean that lacks a generic type.

tkvangorder: Or we can revert back to using a customizer that passes in a list of classes that should be ignored? This basically inverts the problem and allows the use of `beanFactory.isTypeMatch()`. I really like the idea of a Predicate (as you can build much smarter ignore logic) but I am wondering if there are any cases where a list of classes wouldn't suffice?

wilkinsona: Unfortunately, we can't safely use `isTypeMatch()` either. If a bean is a `FactoryBean`, Framework may instantiate it to determine the type of the object that it produces.

wilkinsona:
Flagged for team attention so we can discuss the best way to go from a bean definition to the type of the defined bean.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/LazyInitializationBeanFactoryPostProcessor.java
tkvangorder: One thing that could be done (at least for collecting the Predicates) would be to use `SpringFactoriesLoader` rather than scanning the list of bean definitions. Of course, once we have the list of predicates we are still back to the need of passing each bean definition's type to those predicates.

wilkinsona: We considered `spring.factories` (it has some notable benefits over beans being created very early) but preferred the `@Bean`-based approach that allows the predicate and the auto-configuration to which it relates to be defined together in the same class.

philwebb:
I've raised https://github.com/spring-projects/spring-framework/issues/23056 to see what (if any) of our `BeanTypeRegistry` can be pulled up into Spring Framework.

philwebb:
Although the `BeanTypeRegistry` code has now been pulled up, we're still blocked on https://github.com/spring-projects/spring-framework/issues/23374

tkvangorder:
@wilkinsona @philwebb 

Since we now have a method to extract all beans of a given type from the factory (without initializing the factory beans): 

How would you feel about not using a predicate, but rather reverting back to a "customizer" that returns a list of classes that should be excluded from the lazy-loading processor? I can then just iterate over the classes from the customizers and use `getBeansOfType(Class<?>, false, false);` without requiring any additional changes to the BeanFactory.

philwebb:
@tkvangorder I'm still hopeful that we can get a fix for https://github.com/spring-projects/spring-framework/issues/23374 into Framework before GA

tkvangorder:
@philwebb I have updated this PR to take advantage of the new `BeanFactory.getType(name, boolean)` method.

philwebb:
@tkvangorder Wow that was quick!! It only landed in Framework a few hours ago 👍

philwebb:
@tkvangorder I'm afraid we haven't got to it yet but this one is really high on the list and we need to merge it before RC1. Rest assured, we've not forgotten :)

philwebb:
Thanks very much @tkvangorder, this has now been merged to master.

I've refined things a little in commit 3ffc5f2a301186c67fa2d59cbdb83c2eaa7eaf48. I decided to rename `EagerLoadingBeanDefinitionPredicate` to `LazyInitializationExcludeFilter` to match some of the other callback interface names that we already have. I also extended the signature to pass in the `beanName` and `beanDefinition` as well as the `beanType`. Since that will complicate one of the more obvious implementations, I also added a static factory method so you can do `LazyInitializationExcludeFilter.forBeanTypes(IntegrationFlow.class)` to quickly create a type based filter.

Thanks again for all your work on this, and please let me know if you find any issues with my changes!

tkvangorder:
I like the changes. Thanks!

PR #16618, opened 2019-04-22
Polish Maven plugin tests

JoshuaJeme:
1. Delete useless test case for excludeArtifactIds
1. Correct jar location in test cases


philwebb:
Good spot. Thanks!

wilkinsona:
Thanks for the PR, @JoshuaJeme.

I am not sure that the deleted test case is useless, but it does appear to contain an assertion that's faulty in the same way as the others that you have fixed here. If I comment out the exclusion and fix the assertion the test then fails as expected.

Perhaps I've missed something. Can you please explain why you believe the test is useless?

JoshuaJeme:
> Thanks for the PR, @JoshuaJeme.
> 
> I am not sure that the deleted test case is useless, but it does appear to contain an assertion that's faulty in the same way as the others that you have fixed here. If I comment out the exclusion and fix the assertion the test then fails as expected.
> 
> Perhaps I've missed something. Can you please explain why you believe the test is useless?

The deleted case is for excludeArtifactIds feature, which was deprecated in [Deprecate excludeArtifactIds](https://github.com/spring-projects/spring-boot/commit/55ef9e6dc9c7af64db4f5b90ddd10ba7a7aa5c19#diff-9efac90a956489b90967a694478057eb) and removed in [Merge branch '2.0.x'](https://github.com/spring-projects/spring-boot/commit/5dd4a7e91ec30154fc6996460d6f29dd68993fa7#diff-9efac90a956489b90967a694478057eb)

```xml
			<plugin>
				<groupId>@project.groupId@</groupId>
				<artifactId>@project.artifactId@</artifactId>
				<version>@project.version@</version>
				<executions>
					<execution>
						<goals>
							<goal>repackage</goal>
						</goals>
						<configuration>
							<excludeArtifactIds>servlet-api</excludeArtifactIds>
						</configuration>
					</execution>
				</executions>
			</plugin>
```

The `servlet-api` can't be excluded and will be packaged in to `BOOT-INF/lib`

wilkinsona:
Thanks.

wilkinsona:
Thanks very much for making your first contribution to Spring Boot. The proposed changes have been merged into 2.1.x and forwards into master.

JoshuaJeme:
> Thanks very much for making your first contribution to Spring Boot. The proposed changes have been merged into 2.1.x and forwards into master.

My pleasure  : )

PR #16620, opened 2019-04-22
Optimize JarEntry construction

dreis2211:
Hi,

while profiling one of my applications I noticed that the respective inlining logs showed the following line:
```
@ 3   java.util.zip.ZipEntry::setExtra0 (360 bytes)   hot method too big
```

It seems that most of the time there is no extra information available anyway, so we could skip the call to the apparently expensive `setExtra()`. Before we go into the benchmarks let me say that this PR changes behaviour a bit as `getExtra()` will now return null instead of an empty byte array. But imho this is actually closer to the javadoc of `ZipEntry.getExtra()`:
```
@return the extra field data for the entry, or null if none
```

In an isolated benchmark (I'm just calling setExtra on a ZipEntry) I get the following results:
```
Benchmark             Mode  Cnt          Score         Error  Units
MyBenchmark.testNew  thrpt   10  308679323,969 ± 6449223,651  ops/s
MyBenchmark.testOld  thrpt   10  179973009,964 ± 2125668,778  ops/s
```

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/CentralDirectoryFileHeader.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/CentralDirectoryFileHeader.java
@@ -153,6 +153,10 @@ final class CentralDirectoryFileHeader implements FileHeader {
 		return this.extra;
 	}
 
+	public boolean hasExtra() {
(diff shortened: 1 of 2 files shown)


philwebb:
Nice. I've been trying to find time to have another round on the fat jar support. I think there are a few memory optimizations we could make.

wilkinsona:
Great stuff! Thanks once again, @dreis2211.

Issue #16612, opened 2019-04-20
@ConfigurationPropertiesScan not compatible with @Profile @ConfigurationProperties


romainmoreau:
With the new way of scanning `@ConfigurationProperties` with `@ConfigurationPropertiesScan`, a `@ConfigurationProperties` annotated class is not instanciated when `@Profile` is used in conjunction:

```java
@Profile("mail")
@ConfigurationProperties("mail")
public class MailProperties {
	private String address;

	public String getAddress() {
		return address;
	}

	public void setAddress(String address) {
		this.address = address;
	}
}
```

The full demo repo is available [here](https://github.com/romainmoreau/configuration-properties-scan-and-profile).

kedar-joshi:
Spring Boot 2.0.0.M2 documentation states that -

> Any @Component or @Configuration can be marked with @Profile to limit when it is loaded

So I don't think annotating `@ConfigurationProperties` classes with `@Profile` is a correct use. Although following implementation should work -

```java
@Configuration
@Profile("mail")
public class MailPropertiesConfiguration
{
	@ConfigurationProperties("mail")
	public static class MailProperties
	{
		private String address;

		@PostConstruct
		protected void initialize()
		{
			System.out.println("In MailProperties");
		}

		public String getAddress()
		{
			return address;
		}

		public void setAddress(String address)
		{
			this.address = address;
		}
	}
}
```

Reference : https://docs.spring.io/spring-boot/docs/2.2.0.M2/reference/html/spring-boot-features.html#boot-features-profiles

romainmoreau:
Yes, that's exactly my point, that `@ConfigurationProperties`/`@ConfigurationPropertiesScan` should work like for instance `@Repository`/`@EnableJpaRepositories` which supports `@Profile`.

Using a more complex code like you did doesn't make sense to me as this new feature is supposed to simplify `@ConfigurationProperties` classes (by removing `@Component`).

mbhave:
This is a bug and my guess is that the `ClassPathScanningCandidateComponentProvider` used for scanning `@ConfigurationProperties` doesn't have the `Environment` set and hence doesn't have access to the active profiles in the environment.

PR #16622, opened 2019-04-23
Document using @Profile with @ConfigurationProperties

kedar-joshi:
Documented use of `@Profile` with `@ConfigurationProperties`.

See gh-16612


snicoll:
Thanks for the PR but I am not 100% sure that I want to document that with such a broad scope as this one. While `@Component` and `@Configuration` are clearly stereotypes linked to  component scan, `@ConfigurationProperties` isn't.

wilkinsona:
I think this is worth documenting as proposed and/or by mentioning `@Profile` in the documentation for `@ConfigurationPropertiesScan`.

mbhave:
I've added `merge-with-amendments` so that we can add documentation regarding [this](https://github.com/spring-projects/spring-boot/issues/16645#issuecomment-487415257) as part of merging the PR.

mbhave:
Thanks @kedar-joshi. This is now merged into master.

PR #16629, opened 2019-04-24
JacksonTester handles some characters asymmetrically

### LINKED ISSUE #15727: Use of json-path's default JsonSmartJsonProvider introduces asymmetric handling of special characters

dberrueta-atlassian:
This fixes #15727 by customising the json-path configuration for each {Jackson|Gson|BasicJson}Tester. It also adds integration tests.

This PR replaces #15735 . It is essentially a squashed version of that one.

### DIFF
--- a/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/json/GsonTesterIntegrationTests.java
+++ b/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/json/GsonTesterIntegrationTests.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.test.json;
+
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+
+import com.google.gson.Gson;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration tests for {@link GsonTester}. Shows typical usage.
+ *
+ * @author Andy Wilkinson
+ * @author Diego Berrueta
+ */
+public class GsonTesterIntegrationTests {
+
+	private GsonTester<ExampleObject> simpleJson;
+
+	private GsonTester<List<ExampleObject>> listJson;
+
+	private GsonTester<Map<String, Integer>> mapJson;
+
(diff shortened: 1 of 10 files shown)


spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/json/GsonTesterIntegrationTests.java
```diff
@@ -0,0 +1,87 @@
...
+import com.google.gson.Gson;
+import org.junit.Before;
+import org.junit.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Integration tests for {@link GsonTester}. Shows typical usage.
+ *
+ * @author Andy Wilkinson
```
dberrueta-atlassian: @wilkinsona I hope you don't mind I attributed this file to you. You provided part of the code in https://github.com/spring-projects/spring-boot/pull/15735#issuecomment-484547242

spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/json/JacksonTesterIntegrationTests.java
```diff
@@ -81,6 +83,28 @@ public void typicalMapTest() throws Exception {
 				.isEqualTo(1);
 	}
 
+	@Test
+	public void stringLiteral() throws Exception {
```
dberrueta-atlassian: Same test than for Gson, just to confirm that there isn't a similar regression when using Jackson.

dberrueta-atlassian:
I took @wilkinsona 's suggestion from https://github.com/spring-projects/spring-boot/pull/15735#issuecomment-484547242 and in this new PR I limit the change just to `JacksonTester`, which was the only tester for which I had observed an asymmetric behaviour.

wilkinsona:
Given that this is a bug and the fix appears to be reasonably low-risk, I'm tempted to fix this in 2.1.x. Flagging for team attention to see what everyone else thinks.

philwebb:
I forgot to comment here yesterday. This is now merged into 2.1.x and master along with a [polish commit](https://github.com/spring-projects/spring-boot/commit/d4151e0b516cfbea1d9e533de41f727a83a91934). Thanks very much @dberrueta-atlassian for the PR!

PR #16632, opened 2019-04-24
Add support for Hazelcast YAML configuration

### LINKED ISSUE #16074: Support YAML for Hazelcast configuration
Hazelcast has been traditionally configured with an XML file, `hazelcast.xml` for servers and `hazelcast-client.xml` for clients, and this is supported by Spring Boot Autoconfiguration.

Hazelcast is enhancing this to allow the same to be done with YAML, see https://docs.hazelcast.org/docs/latest-dev/manual/html-single/#configuring-declaratively-yaml

As an enhancement, Spring Boot should auto-configure a Hazelcast instance based on the presence of this YAML in the same way as is done for the presence of XML

leszko:
Since Hazlecast 3.12, YAML configuration format is supported in
addition to XML. Therefore, this change makes Spring Boot automatically
discover not only `hazelcast.xml` (and `hazelcast-client.xml`), but also
`hazelcast.yaml` (and `hazelcast-client.yaml`).

fix #16074 

Signed-off-by: Rafal Leszko <rafal@hazelcast.com>

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastClientFactory.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastClientFactory.java
@@ -65,7 +65,7 @@ public class HazelcastClientFactory {
 		if (configFileName.endsWith(".xml")) {
 			return new XmlClientConfigBuilder(configUrl).build();
 		}
-		else { // yaml config available in the classpath
+		else {
 			return new YamlClientConfigBuilder(configUrl).build();
 		}
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastClientFactory.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastClientFactory.java
@@ -22,6 +22,7 @@ import java.net.URL;
 import com.hazelcast.client.HazelcastClient;
 import com.hazelcast.client.config.ClientConfig;
 import com.hazelcast.client.config.XmlClientConfigBuilder;
+import com.hazelcast.client.config.YamlClientConfigBuilder;
 import com.hazelcast.core.HazelcastInstance;
 
 import org.springframework.core.io.Resource;
@@ -59,7 +60,14 @@ public class HazelcastClientFactory {
 	private ClientConfig getClientConfig(Resource clientConfigLocation)
 			throws IOException {
 		URL configUrl = clientConfigLocation.getURL();
-		return new XmlClientConfigBuilder(configUrl).build();
+		String configFileName = configUrl.getFile();
+
+		if (configFileName.endsWith(".xml")) {
+			return new XmlClientConfigBuilder(configUrl).build();
+		}
+		else { // yaml config available in the classpath
+			return new YamlClientConfigBuilder(configUrl).build();
+		}
 	}
 
 	/**
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastInstanceFactory.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastInstanceFactory.java
@@ -21,6 +21,7 @@ import java.net.URL;
 
 import com.hazelcast.config.Config;
 import com.hazelcast.config.XmlConfigBuilder;
+import com.hazelcast.config.YamlConfigBuilder;
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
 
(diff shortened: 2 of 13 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastClientFactory.java
```diff
@@ -59,7 +60,14 @@ public HazelcastClientFactory(ClientConfig clientConfig) {
 	private ClientConfig getClientConfig(Resource clientConfigLocation)
 			throws IOException {
 		URL configUrl = clientConfigLocation.getURL();
-		return new XmlClientConfigBuilder(configUrl).build();
+		String configFileName = configUrl.getFile();
+
+		if (configFileName.endsWith(".xml")) {
+			return new XmlClientConfigBuilder(configUrl).build();
+		}
+		else { // yaml config available in the classpath
```
snicoll: What does this comment mean?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastInstanceFactory.java
```diff
@@ -61,7 +62,16 @@ public HazelcastInstanceFactory(Config config) {
 
 	private Config getConfig(Resource configLocation) throws IOException {
 		URL configUrl = configLocation.getURL();
-		Config config = new XmlConfigBuilder(configUrl).build();
+		String configFileName = configUrl.getFile();
+
+		Config config;
```
snicoll: Rather than having a field initialization, I'd create a private method that returns the `Config` instance.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
```diff
@@ -77,7 +77,8 @@ public HazelcastInstance hazelcastInstance(Config config) {
 
 		ConfigAvailableCondition() {
 			super(CONFIG_SYSTEM_PROPERTY, "file:./hazelcast.xml",
-					"classpath:/hazelcast.xml");
+					"classpath:/hazelcast.xml", "file:./hazelcast.yaml",
```
snicoll: Same ordering remark

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastAutoConfigurationTests.java
```diff
@@ -39,6 +39,8 @@
 	@Test
 	public void defaultConfigFile() {
 		// no hazelcast-client.xml and hazelcast.xml is present in root classpath
+		// this also asserts that XML has priority over YAML
```
snicoll: It seems wrong to me that XML on the classpath has priority over YAML on the file system. Is this really what Hazelcast does?

spring-boot-project/spring-boot-autoconfigure/src/test/resources/hazelcast.xml
```diff
@@ -5,7 +5,6 @@
 	<map name="defaultCache" />
 	<network>
 		<join>
-			<tcp-ip enabled="false" />
```
snicoll: This seems an unrelated change. I am happy to polish the support in a separate commit though. Can you explain why you've removed this line?

spring-boot-project/spring-boot-autoconfigure/src/test/resources/org/springframework/boot/autoconfigure/hazelcast/hazelcast-specific.xml
```diff
@@ -2,16 +2,13 @@
 		   xmlns="http://www.hazelcast.com/schema/config"
 		   xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance">
 
-	<queue name="foobar"/>
```
snicoll: Same remark. Is this related to the main purpose of this PR?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastInstanceFactory.java
leszko: Changed

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastAutoConfigurationTests.java
leszko: yeah, as explained in the other comment.

spring-boot-project/spring-boot-autoconfigure/src/test/resources/hazelcast.xml
leszko: Yeah, it's just not used. Reverted. Can be sent in a separate PR.

spring-boot-project/spring-boot-autoconfigure/src/test/resources/org/springframework/boot/autoconfigure/hazelcast/hazelcast-specific.xml
leszko: Reverted.

snicoll:
@leszko thank you for making your first contribution to Spring Boot and following up quickly on feedback. This is now merged in `master` with a [polish commit](https://github.com/spring-projects/spring-boot/commit/704da1750b62e192fa0ddb33ac831da40760002e). What I've changed in particular is the check on xml vs. yaml, with an explicit check on `yaml` and a fallback on xml rather than the other way around.

Also, tests assertions were all identical so that didn't really test that the right file was loaded. I couldn't fix it for system properties test of the server as I believe I've found an inconsistency in Hazelcast. I've raised https://github.com/hazelcast/hazelcast/issues/14960.

Thanks again.

leszko:
Thanks @snicoll

PR #16658, opened 2019-04-25
Optimize ConfigurationPropertyName.buildToString()

dreis2211:
Hi,

though minor this PR optimizes calls to StringBuilder.append with a single character in `ConfigurationPropertyName.buildToString()`. In the example provided in #16401 this is called 33k times at least and every little helps.

Cheers,
Christoph


wilkinsona:
It does indeed. Thanks for the PR.

PR #16665, opened 2019-04-26
Support spying on final classes

cvienot:
Closes gh-7033

We currently cannot spy on spring data repositories because the proxy are finals.
We can work around this limitation by using mockito inline-mocker-maker plugin.

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/resources/mockito-extensions/org.mockito.plugins.MockMaker
+++ b/spring-boot-project/spring-boot-test/src/main/resources/mockito-extensions/org.mockito.plugins.MockMaker
@@ -1,0 +1,1 @@
+mock-maker-inline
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the suggestion but I don't think we should opt everyone in to Mockito's support for mocking final classes. It's still marked as an incubating feature and the Mockito release notes have this to say about it:

> Mocking of final classes and methods is an incubating, opt-in feature. It uses a combination of Java agent instrumentation and subclassing in order to enable mockability of these types. As this works differently to our current mechanism and this one has different limitations and as we want to gather experience and user feedback, this feature had to be explicitly activated to be available

I believe that you can opt in to the new mechanism and its different limitations by adding the `org.mockito.plugins.MockMaker` metadata to your own application in `src/test/resources`.

PR #16668, opened 2019-04-26
Remove redundant info

Tetsuya3850:
Since the same information was repeated immediately, I suggest removing it.


philwebb:
I agree, it reads better without that duplication. Thanks for the PR!

snicoll:
@Tetsuya3850 thank you!

PR #16671, opened 2019-04-26
Polish ConfigurationPropertyName.dashIgnoringElementEquals()

izeye:
This PR polishes `ConfigurationPropertyName.dashIgnoringElementEquals()` by avoiding repeating `indexed2` `boolean` checks in the loop as checking it only once seems sufficient.


dreis2211:
True - actually the default element equals has the same "problem". That was still on my list of things to do.

philwebb:
Thanks @izeye and @dreis2211! I've pushed this to master and also aligned `defaultElementEquals`

PR #16678, opened 2019-04-28
Add jOOQ to SQL logging group

maciejwalkowiak:
Since Spring Boot comes with support for JOOQ it would be also nice if the `sql` logging group also include it.

After this change, setting `logging.level.sql=debug` and executing query through jOOQ will result entries like one below in logs:

```
Executing query          : select "public"."property_ad"."id", "public"."property_ad"."bathrooms", "public"."property_ad"."price", "public"."property_ad"."rooms", "public"."property_ad"."surface_area" from "public"."property_ad"
Fetched result           : +----+---------+-----+-----+------------+
                         : |  id|bathrooms|price|rooms|surface_area|
                         : +----+---------+-----+-----+------------+
                         : |   1|        1|   40|    3|         120|
                         : |   2|        2|   80|    5|         150|
                         : +----+---------+-----+-----+------------+
Fetched row(s)           : 2
```
Pinging @lukaseder in case I missed something.


lukaseder:
> Pinging @lukaseder in case I missed something.

That sounds very reasonable. The default `Settings.executeLogging` value is `true`, so this configuration should indeed suffice.

PR #16685, opened 2019-04-30
Polish

izeye:
This PR polishes `KafkaAutoConfigurationTests` by using AssertJ's method in favor of `KafkaTestUtils` for consistency.


snicoll:
Good catch Johnny, thanks again!

PR #16700, opened 2019-05-02
Polish RSocketServerBootstrap bean name

izeye:
This PR renames the bean name for `RSocketServerBootstrap` as it seems missed in 71362299aee1068d62f730b6f391602849f0774b.


snicoll:

PR #16702, opened 2019-05-02
Added Cassandra Cluster factory and Default implementation.

ssoerensen:
This allows the default cassandra builder to be overwritten. This is useful for example, when implementing jaeger/opentracing

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraAutoConfiguration.java
@@ -50,7 +50,7 @@ public class CassandraAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean
-	public CassandraClusterFactory cassandraClusterFactory(){
+	public CassandraClusterFactory cassandraClusterFactory() {
 		return new CassandraClusterFactoryImpl();
 	}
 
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@ssoerensen Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16702)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16702) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ssoerensen Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16702)!

snicoll:
> This is useful for example, when implementing jaeger/opentracing

Is that using a decorator pattern of some sort? I don't think increasing the surface area with an extra callback is actually worth it. Have you tried to use a `BeanPostProcessor` to wrap the `Cluster` bean instead?

ssoerensen:
It was my first thought to use a `BeanPostProcessor`. However the way the opentracing cassandra driver is made prevents that. Basically they have a TracingCluster class which extends the original Cluster class. it is instantiated like so:
```
// Instantiate Tracing Cluster
Cluster cluster = new TracingCluster(builder, tracer);
```

snicoll:
Thanks for the feedback. I am still not convinced we should create an interface for that specific use case. Perhaps we could expose the `Builder` and then let you inject it and create the cluster any way you like. 

wilkinsona:
@ssoerensen Thank you very much for making your first contribution to Spring Boot. The proposed changes have been merged into master, along with a [polish commit](https://github.com/spring-projects/spring-boot/commit/ece87cc4cc0772e334d7e9d185bc7d8639e59d38). The two main changes were to remove the default implementation of the new interface and to add a test.

PR #16703, opened 2019-05-02
Redis pooling minIdle property is ignored as timeBetweenEvictionRun defaults to -1 and cannot be configured

geminiKim:
Hi :D
I found the bug that `min-idle` does not work when configure Lettuce with pool config.

You can find out why the problem occurred by looking below link.
[https://commons.apache.org/GenericObjectPool.html#setMinIdle](https://commons.apache.org/proper/commons-pool/apidocs/org/apache/commons/pool2/impl/GenericObjectPool.html#setMinIdle-int-)
```
This setting only has an effect if it is positive and BaseGenericObjectPool.getTimeBetweenEvictionRunsMillis() is greater than zero.
```

`TimeBetweenEvictionRunsMillis` need to configured greater then zero for the `min-idle` to working correctly.

Some users can trust settings and use them incorrectly in an environment that does not work properly.
So, I added that property, but I don't know if this is a perfectly good solution.

Please review and let me know what you think about this issue.
Thank you :)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
@@ -147,7 +147,7 @@ class LettuceConnectionConfiguration extends RedisConnectionConfiguration {
 			config.setMaxIdle(properties.getMaxIdle());
 			config.setMinIdle(properties.getMinIdle());
 			config.setTimeBetweenEvictionRunsMillis(
-					properties.getIdleEvictPeriod().toMillis());
+					properties.getTimeBetweenEvictionRuns().toMillis());
 			if (properties.getMaxWait() != null) {
 				config.setMaxWaitMillis(properties.getMaxWait().toMillis());
 			}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
@@ -146,6 +146,8 @@ class LettuceConnectionConfiguration extends RedisConnectionConfiguration {
 			config.setMaxTotal(properties.getMaxActive());
 			config.setMaxIdle(properties.getMaxIdle());
 			config.setMinIdle(properties.getMinIdle());
+			config.setTimeBetweenEvictionRunsMillis(
+					properties.getIdleEvictPeriod().toMillis());
 			if (properties.getMaxWait() != null) {
 				config.setMaxWaitMillis(properties.getMaxWait().toMillis());
 			}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
@@ -146,8 +146,10 @@ class LettuceConnectionConfiguration extends RedisConnectionConfiguration {
 			config.setMaxTotal(properties.getMaxActive());
 			config.setMaxIdle(properties.getMaxIdle());
 			config.setMinIdle(properties.getMinIdle());
-			config.setTimeBetweenEvictionRunsMillis(
-					properties.getTimeBetweenEvictionRuns().toMillis());
+			if (properties.getTimeBetweenEvictionRuns() != null) {
+				config.setTimeBetweenEvictionRunsMillis(
+						properties.getTimeBetweenEvictionRuns().toMillis());
+			}
 			if (properties.getMaxWait() != null) {
 				config.setMaxWaitMillis(properties.getMaxWait().toMillis());
 			}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
@@ -173,8 +173,8 @@ public class RedisProperties {
 		 * Target for the minimum number of idle connections to maintain in the pool. This
 		 * setting only has an effect if it is positive.
 		 *
-		 * This setting only has an effect if it is positive and `idleEvictPeriod` is
-		 * greater than zero.
+		 * This setting only has an effect if it is positive and `timeBetweenEvictionRuns`
+		 * is greater than zero.
 		 */
 		private int minIdle = 0;
 
@@ -192,12 +192,12 @@ public class RedisProperties {
 		private Duration maxWait = Duration.ofMillis(-1);
 
 		/**
-		 * Time of milliseconds to sleep between runs of the idle object evictor thread.
+		 * Time to sleep between runs of the idle object evictor thread.
 		 *
 		 * When positive, the idle object evictor thread starts. When non-positive, no
 		 * idle object evictor thread runs.
 		 */
-		private Duration idleEvictPeriod = Duration.ofMillis(-1);
+		private Duration timeBetweenEvictionRuns = Duration.ofMillis(-1);
 
 		public int getMaxIdle() {
 			return this.maxIdle;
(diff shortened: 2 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
```diff
@@ -146,6 +146,8 @@ public LettuceClientConfigurationBuilder createBuilder(Pool properties) {
 			config.setMaxTotal(properties.getMaxActive());
 			config.setMaxIdle(properties.getMaxIdle());
 			config.setMinIdle(properties.getMinIdle());
+			config.setTimeBetweenEvictionRunsMillis(
```
wilkinsona: Jedis supports the same configuration property. Would you like to make the equivalent change there as well?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfigurationTests.java
```diff
@@ -165,6 +166,8 @@ public void testRedisConfigurationWithPool() {
 					assertThat(poolConfig.getMaxIdle()).isEqualTo(4);
 					assertThat(poolConfig.getMaxTotal()).isEqualTo(16);
 					assertThat(poolConfig.getMaxWaitMillis()).isEqualTo(2000);
+					assertThat(poolConfig.getTimeBetweenEvictionRunsMillis())
```
wilkinsona: Would you like to make an equivalent change to the tests for configuring Jedis as well?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -188,6 +191,14 @@ public Lettuce getLettuce() {
 		 */
 		private Duration maxWait = Duration.ofMillis(-1);
 
+		/**
+		 * Time of milliseconds to sleep between runs of the idle object evictor thread.
+		 *
+		 * When positive, the idle object evictor thread starts. When non-positive, no
+		 * idle object evictor thread runs.
+		 */
+		private Duration idleEvictPeriod = Duration.ofMillis(-1);
```
wilkinsona: I think this name should align more closely with the setting that it maps to. Something like `timeBetweenEvictionRuns`, perhaps.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -188,6 +191,14 @@ public Lettuce getLettuce() {
 		 */
 		private Duration maxWait = Duration.ofMillis(-1);
 
+		/**
+		 * Time of milliseconds to sleep between runs of the idle object evictor thread.
```
wilkinsona: As with the existing description for `maxWait`, the description shouldn't mention the unit (milliseconds in this case). Using `Duration` means that you can specify the time in any unit you want.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
gemini-kim: I think it does not the responsibility of Spring-Boot.
Spring-Boot configured the Jedis-Pool using the below code:

https://github.com/spring-projects/spring-boot/blob/36c1c051b8106371d37f8a747cea0e369a8e5c84/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/JedisConnectionConfiguration.java#L112

So, i think `JedisPoolConfig` is responsibility to `Jedis`.
`JedisPoolConfig` sets the default value of `TimeBetweenEvictionRunsMillis`

https://github.com/xetorthio/jedis/blob/ef4ab403f9d8fd88bd05092fea96de2e9db0bede/src/main/java/redis/clients/jedis/JedisPoolConfig.java#L10

So I'm not sure I should handle this equally.

I will wait for your opinion, Thank you for your fast feedback and thoughtful review :)

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
gemini-kim: I totally agree with your opinion. 👍

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
geminiKim: I think it does not the responsibility of Spring-Boot.
Spring-Boot configured the Jedis-Pool using the below code:

https://github.com/spring-projects/spring-boot/blob/36c1c051b8106371d37f8a747cea0e369a8e5c84/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/JedisConnectionConfiguration.java#L112

So, i think `JedisPoolConfig` is responsibility to `Jedis`.
`JedisPoolConfig` sets the default value of `TimeBetweenEvictionRunsMillis`

https://github.com/xetorthio/jedis/blob/ef4ab403f9d8fd88bd05092fea96de2e9db0bede/src/main/java/redis/clients/jedis/JedisPoolConfig.java#L10

So I'm not sure I should handle this equally.

I will wait for your opinion, Thank you for your fast feedback and thoughtful review :)

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
geminiKim: I totally agree with your opinion. 👍

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfigurationTests.java
geminiKim: Same discussion as [comments](https://github.com/spring-projects/spring-boot/pull/16703#discussion_r280812432)

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
wilkinsona: Thanks. I'd missed that `JedisPoolConfig` sets the property already and, therefore, has a different default to Lettuce.

Unfortunately, this creates a problem as the `Pool` class is shared between Jedis and Lettuce configuration properties. As currently proposed here, the changes will result in a `spring.redis.jedis.pool` property that has no effect. We need to figure out how to avoid that while also coping with the different defaults.

I think our best option is to declare the `timeBetweenEvictionRuns` property with a `null` default value and only map it onto the Lettuce or Jedis pool config if the value is non-null.

geminiKim: I agree that it seems to be the best way now, I modified it.
Thank you for your clear and good feedback. 😄

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfigurationTests.java
geminiKim: I made an equivalent change to the tests for configuring Jedis.

geminiKim:
@wilkinsona I apologize for the mention.
I think this task is done. Do I have more to do?

And failing tests seem to fail for other reasons, I only worked `Spring Boot AutoConfigure`, but failing test module is `Spring Boot Server Tests`.

wilkinsona:
No need to apologise. There's nothing more for you to do, I'm just a bit snowed under preparing for next week's Spring IO conference.

geminiKim:
Thank you for your quick feedback during the busy time. :D

wilkinsona:
@geminiKim Thank you very much for making your first contribution to Spring Boot. The proposed changes have been merged into 2.1.x and master along with a small [polishing commit](https://github.com/spring-projects/spring-boot/commit/96f3a482f0520d7d4c78deb6175a325a6fa0ef79).

geminiKim:
@wilkinsona Thank you for your kind review in busy times. I will continue to contribute.

And i will cheer your session on Spring I/O. :D

PR #16707, opened 2019-05-02
Document that JUnit 4 is the default

Tetsuya3850:
I assumed the latest spring boot starter test uses JUnit 5 and had to scratch my head over it. I suggest adding it to the documentation as other people may be confused as well.

Reference
http://central.maven.org/maven2/org/springframework/boot/spring-boot-starter-test/2.1.4.RELEASE/spring-boot-starter-test-2.1.4.RELEASE.pom

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
@@ -6875,7 +6875,8 @@ specific slices>> of an application.
 TIP: If you are using JUnit 4, don't forget to also add `@RunWith(SpringRunner.class)` to
 your test, otherwise the annotations will be ignored. If you are using JUnit 5, there's no
 need to add the equivalent `@RunWith(SpringExtension.class)` as `@SpringBootTest` and
-the other `@…Test` annotations are already annotated with it.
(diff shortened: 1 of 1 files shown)


Tetsuya3850:
Add Note

This part seemed to imply that the latest is JUnit 5.

If you are using JUnit 4, don’t forget to also add @RunWith(SpringRunner.class) to your test, otherwise the annotations will be ignored. If you are using JUnit 5, there’s no need to add the equivalent @ExtendWith(SpringExtension.class) as @SpringBootTest and the other @…Test annotations are already annotated with it.

wilkinsona:
Thanks for the suggestion. We [already mention](https://docs.spring.io/spring-boot/docs/2.1.4.RELEASE/reference/htmlsingle/#boot-features-kotlin-testing) that JUnit 4 is the default in the Kotlin testing section. I can't see any harm in mentioning it more generally, but I think it would be better done in [this section](https://docs.spring.io/spring-boot/docs/2.1.4.RELEASE/reference/htmlsingle/#boot-features-testing). If you agree, would you like to update your proposal?

Tetsuya3850:
Ah,,, I have missed it since I do not use Kotlin. Yes, I will update my proposal based on your suggestion!

Tetsuya3850:
I moved the doc to a more appropriate place!

snicoll:
@Tetsuya3850 thank you, this is now merged in `2.1.x` and `master`.

PR #16714, opened 2019-05-03
Add Tip that h2 needs to be in your class path for @DataJpaTest to work.

Tetsuya3850:
Docs "46.3.12 Auto-configured Data JPA Tests" doesn't specify the necessity to configure an in-memory database. It reads as if everything "just" works.
https://docs.spring.io/spring-boot/docs/2.1.4.RELEASE/reference/htmlsingle/#boot-features-testing-spring-boot-applications-testing-autoconfigured-jpa-test

As consequence, I experienced the same issue as indicated in this stack overflow QA.
https://stackoverflow.com/questions/40659912/spring-boot-starter-test-cannot-run-database-integration-test

So, I suggest clarifying that you need to add an embedded in-memory database for setup.


snicoll:
You've edited your first comment that quoted the error message we produce

> Failed to replace DataSource with an embedded database for tests. If you want an embedded database please put a supported one on the classpath or tune the replace attribute of @AutoConfigureTestDatabase.

I think that error message is quite explicit on its own. I agree the doc could be a bit more explicit that it does so if it's on the classpath, this is not a requirement at all actually. I've rephrased that in 680ac85

Thanks for the PR, in any case!

Tetsuya3850:
I agree with your rephrase! Thank you.

PR #16717, opened 2019-05-04
Optimize CacheKey handling in SpringIterableConfigurationPropertySource

dreis2211:
Hi,

as described in #16401 there is an optimization opportunity in `SpringIterableConfigurationPropertySource` when checking for CacheKey equality. Currently, for large Sets this takes a considerable amount of time as the two Sets are usually equal, but not the same instance. This is due to the fact that the internal key inside CacheKey is copied in order to fix #13344 and can thus not benefit from a `==` comparison.

The idea of this PR is to introduce a flag that effectively disables the copying of the internal key under certain circumstances. Imho, this could be done for `OriginTrackedMapPropertySource` instances which is used to load either `.yaml` or `.properties` files, which usually shouldn't be a subject of change (and even if require a refresh of some sort). I still implemented an additional check for the size of the internal key in a best effort to track key **additions** at least for `OriginTrackedMapPropertySource`s, too.

With the applied changes, I see major improvements compared to a M2 baseline.

|       | Baseline |  New  |
| ----- | -------: | ----: |
|       |    12.757 | 7.947 |
|       |    11.742 | 8.254 |
|       |    12.453 | 7.790 |
|       |    12.819 | 7.513 |
|       |    12.588 | 8.021 |
|       |    12.616 | 8.038 |
|       |    12.541 | 7.495 |
|       |    13.555 | 7.864 |
|       |    12.494 | 8.040 |
|       |    12.009 | 7.726 |
|  Mean |    12.557 | 7.869 |
| Range |    1.813 | 0.759 |

Let me know what you think.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/SpringIterableConfigurationPropertySource.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/SpringIterableConfigurationPropertySource.java
@@ -24,6 +24,7 @@ import java.util.List;
 import java.util.Set;
 import java.util.stream.Stream;
 
+import org.springframework.boot.env.OriginTrackedMapPropertySource;
 import org.springframework.core.env.EnumerablePropertySource;
 import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.PropertySource;
@@ -44,7 +45,7 @@ import org.springframework.util.ObjectUtils;
 class SpringIterableConfigurationPropertySource extends SpringConfigurationPropertySource
 		implements IterableConfigurationPropertySource {
 
-	private volatile Object cacheKey;
+	private volatile CacheKey cacheKey;
 
 	private volatile Cache cache;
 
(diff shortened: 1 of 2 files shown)


dreis2211:
As an addendum. This brings us pretty close to the performance of 1.5 for such big files.

|       | Baseline |  New  |
| ----- | -------: | ----: |
|       |    8.741 | 7.931 |
|       |    7.650 | 7.873 |
|       |    7.475 | 7.834 |
|       |    8.305 | 7.890 |
|       |    7.574 | 8.064 |
|       |    7.719 | 8.019 |
|       |    7.779 | 7.722 |
|       |    7.700 | 7.417 |
|       |    7.903 | 7.750 |
|       |    7.293 | 8.103 |
|  Mean |    7.814 | 7.860 |
| Range |    1.448 | 0.686 |

dreis2211:
Test failure seems unrelated

philwebb:
Thanks @dreis2211, sorry about the delay getting to these. I hope to review soon.

dreis2211:
Don't worry. There's release preparation and it's conference season etc.. Whenever you're ready :)

philwebb:
This is great!!! I've updated things a little in [this commit](https://github.com/spring-projects/spring-boot/commit/8160368c90bd2b21467102c528ba7f09101d168c) so that sources actually need to declare themselves as immutable. We can then use a shared cache key that always matches. This should save us from even creating the `CacheKey`. Fantastic work. Thanks once again @dreis2211.

Issue #7321, opened 2016-11-05
MockMvc doesn't use spring-boot's mvc exception handler


xak2000:
I'm trying to test my Controller using MockMvc.
The service used by Controller throws RuntimeException if something is wrong.
The spring-boot's default exception resolver catches this exception and responds with status 500.
But when I use MockMvc in my test, the test method just throws the exception instead of catch status 500.
If I add explicit exception resolver using `@ExceptionHandler`, then test method starts to work.

The simple example. My Controller just throws `new RuntimeException()` for simplicity.
```
@RunWith(SpringRunner.class)
@WebMvcTest(MyController.class)
public class ControllerTests {

	@Autowired
	private MockMvc mvc;

	@Test
	public void testInternalServerError() throws Exception {
		mvc.perform(get("/api/data"))
				.andExpect(status().isInternalServerError());
	}
}
```

Expected result:
Test catches status 500 and passes.

Actual result:
The exception is thrown and test fails.
```
org.springframework.web.util.NestedServletException: Request processing failed; nested exception is java.lang.RuntimeException
```

Maybe I'm wrong assuming that spring-boot installs some default mvc exception handler. But anyway the real application returns status 500 with properly formatted JSON of error details. So I think the MockMvc should do the same.

I can add the reproducible example if someone confirms that this is really not intended behaviour.

xak2000:
Also, even if I add explicit `@ExceptionHandler` like this:
```
	@ExceptionHandler(RuntimeException.class)
	public void hadnle(HttpServletResponse res) throws IOException {
		res.sendError(404, "bla bla");
	}
```

Then real response and MockMvc response is also different.

The real response contains JSON with error details, like this:
```
{
  "timestamp" : "2016-11-06T03:16:02.440+0000",
  "status" : 404,
  "error" : "Not Found",
  "exception" : "java.lang.RuntimeException",
  "message" : "bla bla",
  "path" : "/api/data"
}
```

But the MockMvc response body is empty:
```
MockHttpServletResponse:
           Status = 404
    Error message = bla bla
          Headers = {}
     Content type = null
             Body = 
    Forwarded URL = null
   Redirected URL = null
          Cookies = []
```

So we can't check, for example, the `message` field.

bclozel:
[...]

PR #16718, opened 2019-05-04
Doc the pros and cons of Out-of-Container and Integration Tests

Tetsuya3850:
The current doc does not explain why one should choose either the mockMVC or a full running server. 
https://docs.spring.io/spring-boot/docs/2.1.4.RELEASE/reference/htmlsingle/#boot-features-testing-spring-boot-applications-testing-with-mock-environment

I happened to choose mockMVC without knowing its’ shortcomings that error handling behaves differently than the real environment. This bumping point was indicated in this issue as well. (https://github.com/spring-projects/spring-boot/issues/7321) Therefore, I suggest explaining its pros and cons more clearly in the docs.


philwebb:
Thanks very much @ Tetsuya3850, this has now been merged with a few minor refinements.

PR #16728, opened 2019-05-06
Add kafka-streams-test-utils to dependency

garyrussell:
This jar is now a dependency of spring-kafka-test which is added in
test scope by initializr.

See https://github.com/spring-projects/spring-kafka/issues/1084


snicoll:
That feels a candidate to `2.1.x` as well as one might argue that our dependency management for Kafka is missing a library.

wilkinsona:
+1 for adding this in 2.1.x. I think we should review the other entries and see what else is missing too.

Issue #16488, opened 2019-04-08
spring.resources.cache.period doesn't work


OrangeDog:
Spring Boot 2.1.4.RELEASE

As far as I can tell, the property `spring.resources.cache.period` can't ever do anything.
`ResourceProperties.Cache.Cachecontrol.toHttpCacheControl` will always return an object, therefore [this else branch](https://github.com/spring-projects/spring-framework/blob/v5.1.6.RELEASE/spring-webmvc/src/main/java/org/springframework/web/servlet/config/annotation/ResourceHandlerRegistration.java#L172) is never taken.

A possible fix is to ensure `toHttpCacheControl` returns `null` if no properties are set.

snicoll:
The cache period is [configured as the MaxAge of the CacheControl](https://github.com/spring-projects/spring-boot/blob/cd914aaa4d7507b4e7fd66c65701335745a49303/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java#L169-L177). There is also a test that asserts it is set.

Rather than describing a technical solution, can you share a sample we can run that reproduces the problem you're experiencing?

OrangeDog:
Sorry, this is with WebMVC, not WebFlux.
It's reproduced with spring-boot-starter-web, no config, a single static file, and the `...period` property set to something valid.

snicoll:
And what are we supposed to reproduce with this setup? MVC is also covered by [a test](https://github.com/spring-projects/spring-boot/blob/cd914aaa4d7507b4e7fd66c65701335745a49303/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcAutoConfigurationTests.java#L796-L815) that's passing at the moment. Can we please ignore the technical analysis and focus on the feature?

OrangeDog:
I am focussing on the feature. Despite whatever tests might exist, it doesn't actually work.
Here's an explicit reproduction: https://gist.github.com/OrangeDog/8899a0b33561c3aa2100d9fe94a40f3f

```
$ curl -i http://localhost:8080/test.js
HTTP/1.1 200
Last-Modified: Mon, 08 Apr 2019 11:53:35 GMT
Accept-Ranges: bytes
Content-Type: application/javascript
Content-Length: 0
Date: Mon, 08 Apr 2019 12:15:09 GMT
```

[...]

PR #16730, opened 2019-05-06
'spring.resources.cache.period' is always overridden by 'spring.resources.cache.cachecontrol'

eiselems:
Before this change it got overwritten by forwarding an empty
CacheControl to Spring. Spring itself sets CacheSeconds already
correctly in absence (=null) of a CacheControl.

Also:
* Fixes bug in WebMvcAutoConfigurationTests.cachePeriod which prevented it to assert anything

See: #16488

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ResourceProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ResourceProperties.java
@@ -476,6 +476,10 @@ public class ResourceProperties {
 						.staleIfError(duration.getSeconds(), TimeUnit.SECONDS));
 				map.from(this::getSMaxAge).whenNonNull().to((duration) -> control
 						.sMaxAge(duration.getSeconds(), TimeUnit.SECONDS));
+				// check if cacheControl remained untouched
+				if (control.getHeaderValue() == null) {
+					return null;
+				}
 				return control;
 			}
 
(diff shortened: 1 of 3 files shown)


eiselems:
Concourse seems to have issues pulling the docker image :(
About the code, once it gets merged onto master the tests needs to be guarded by 
`Assertions.setExtractBareNamePropertyMethods(false);`

Similar to this PR where it was done to the WebFluxAutoconfigurationTests.class 		https://github.com/spring-projects/spring-boot/commit/a2c5a7988370efd6e56f011b89e3337b3f383a62

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcAutoConfiguration.java
```diff
@@ -332,14 +332,20 @@ public void addResourceHandlers(ResourceHandlerRegistry registry) {
 				return;
 			}
 			Duration cachePeriod = this.resourceProperties.getCache().getPeriod();
-			CacheControl cacheControl = this.resourceProperties.getCache()
+			ResourceProperties.Cache.Cachecontrol cacheControl = this.resourceProperties
+					.getCache().getCachecontrol();
+			if (cachePeriod != null && cacheControl.getMaxAge() == null) {
```
bclozel: `cachePeriod` is not only about max-age; if `cachePeriod=-1` nothing is done, `0` means a `no-store` configuration, a positive number results in a max-age configuration. But this should delegate to Spring Framework and not hardcode those assumptions here.
A short term solution would be to check whether `CacheControl.getHeaderValue()` is not `null` and only apply it in this case. A better solution IMO would be to create a Spring Framework issue and getting this solved at the Framework level (see first comment in #16488).

eiselems: Sounds really reasonable 👍 
I think we just need to keep in mind that Webflux has basically the same implementation and when resolved it should also getting fixed there. See: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java#L169-L177

eiselems:
@bclozel maybe you want to have another look. I documented my findings in #16488 
Basically it comes down to Spring already doing the right thing. It was just that boot was just never giving null as CacheControl

PR #16732, opened 2019-05-07
Rename "Devtools" to "DevTools" in class names

izeye:
I can see there are mixed uses between "Devtools" to "DevTools" but it seems to use "DevTools" at least for class names consistently. This PR renames "Devtools" to "DevTools" in class names.

This PR also adds missing `@since` tags.


wilkinsona:
Thanks as ever, @izeye. Much appreciated.

PR #16733, opened 2019-05-07
Remove unnecessary latches in tests

izeye:
This PR removes unnecessary latches in tests.


wilkinsona:
Thanks once again, @izeye.

PR #16737, opened 2019-05-07
Polish FlywayProperties

izeye:
This PR removes `null` initial values for fields in `FlywayProperties` for consistency.


wilkinsona:
Thanks very much, @izeye.

PR #16738, opened 2019-05-07
Align RestClientAutoConfiguration's use of PropertyMapper::asInt with the rest of the codebase

izeye:
This PR removes redundant `as()` calls in `RestClientAutoConfiguration.restClientBuilder()`.


wilkinsona:
The proposed changes will alter the behaviour of the mapping as a `long` value will now be converted to an `int` by `Number::intValue` rather than `Math::toIntExact`. As a result, if the value is too large to fit in an `int` it will now become negative rather than resulting in an exception. However, I still think we should merge this as it aligns `RestClientAutoConfiguration` with other usages of `PropertyMapper::asInt`.

wilkinsona:
And thanks yet again, @izeye. Please keep them coming!

PR #16739, opened 2019-05-07
Start building against Spring Kafka 2.3.0.M2 snapshots

garyrussell:
Kafka Streams properties are no longer converted to `String`.


wilkinsona:

Issue #16727, opened 2019-05-06
Add KafkaProperty for configuring ContainerProperties. missingQueuesFatal


mrusinak:
In spring-kafka 2.2, a new property was added to `ContainerProperties` as `missingQueuesFatal`; It would be nice to be able to configure this easily when using springboot:

- Have a property to be able to configure via `spring.kafka.***`
- Have `KafkaMessageDrivenChannelAdapter` do the `KafkaProperties` -> `ContainerProperties` configuration of the `missingQueuesFatal` setting automatically

Some background references:
https://stackoverflow.com/questions/55976911/spring-integration-kafka-3-0-1-3-1-2-tests-now-fail-to-start-if-no-broker-ava
https://docs.spring.io/spring-kafka/reference/html/#listener-container-changes

philwebb:
@garyrussell any thoughts on this? Is it a sensible property to surface?

garyrussell:
@philwebb Yes; it was a new feature, but it prevents apps from starting if the broker is down at that time.

It makes sense to expose it as a first class property.

I can issue a PR tomorrow if you want.

philwebb:
Sweet, thanks @garyrussell!

garyrussell:
It turns out that we need a change to the container factory first; stay tuned...

wilkinsona:
Closing in favour of #16740.

PR #16740, opened 2019-05-07
Provide a configuration property for Spring Kafka's missingTopicsFatal

garyrussell:
Resolves https://github.com/spring-projects/spring-boot/issues/16727

Requires https://github.com/spring-projects/spring-boot/pull/16739


snicoll:
@garyrussell thank you for the PR and sorry we didn't manage to merge that for M3. I've polished and changed the type to `boolean` as this property has an obvious default to Spring Kafka. I've added a test to check that our defaults match yours. We should probably review the list and swap other properties at some point, see https://github.com/spring-projects/spring-boot/commit/3c46b9e83de45f262f77fe6a59e3d04efff17c02

garyrussell:
Sounds good; thanks.

Issue #16661, opened 2019-04-25
Use MergedAnnotation.getRoot() method


philwebb:
We should replace the looped called to `getParent()` with `getRoot()` in  `ConfigurationPropertiesScanRegistrar.validateScanConfiguration` and `PropertyMappingCheckBeanPostProcessor` when https://github.com/spring-projects/spring-framework/issues/22818 is fixed.

PR #16745, opened 2019-05-08
Use MergedAnnotation.getRoot() in PropertyMappingCheckBeanPostProcessor

izeye:
This seems missed in #16661.


philwebb:
Thanks again @izeye !!

Issue #16705, opened 2019-05-02
RedisReactiveHealthIndicator blocks on error by calling close, which is not supported in thread reactor-http-epoll


AndreasKl:
```
Caused by: java.lang.IllegalStateException: block()/blockFirst()/blockLast() are blocking, which is not supported in thread reactor-http-epoll-3
	at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:77)
	at reactor.core.publisher.Mono.block(Mono.java:1494)
	at org.springframework.data.redis.connection.ReactiveRedisConnection.close(ReactiveRedisConnection.java:60)
	at org.springframework.boot.actuate.redis.RedisReactiveHealthIndicator.lambda$doHealthCheck$1(RedisReactiveHealthIndicator.java:50)
	at reactor.core.publisher.FluxDoFinally$DoFinallySubscriber.runFinally(FluxDoFinally.java:156)
	... 203 common frames omitted
```

While running a Spring Cloud Gateway application with Redis as session store, I observed the stack trace above. The cause is a blocking call in: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/redis/RedisReactiveHealthIndicator.java#L49

There is a `Mono<Void> closeLater();` https://github.com/spring-projects/spring-data-redis/blob/master/src/main/java/org/springframework/data/redis/connection/ReactiveRedisConnection.java#L68 which could be called instead.

mp911de:
Connection creation is blocking if connection sharing is disabled. From that perspective, either you allocate a connection on `subscribeOn(Schedulers.parallel())` or you obtain the connection in `afterPropertiesSet()`.

wilkinsona:
Closing in favour of #16756.

upwan111:
hi, i am using currently 2.1.3 release version of spring boot and facing this issue in my application. Can I get any information which release version this error will be fixed so that we can update accordingly?

wilkinsona:
@upwan111 The issue was fixed in Spring Boot 2.1.5. Please see #16756, linked to above, for further details.

Joan7:
Currently  I am using 2.1.5.RELEASE of spring-data-redis facing this issue in frequently.My usage on redis is call from redisratelimiter class. Can anyone help on it?

Joan7:
[...]

PR #16756, opened 2019-05-08
RedisReactiveHealthIndicator makes blocking call on error

ayudovin:
See #16705

---

```
Caused by: java.lang.IllegalStateException: block()/blockFirst()/blockLast() are blocking, which is not supported in thread reactor-http-epoll-3
	at reactor.core.publisher.BlockingSingleSubscriber.blockingGet(BlockingSingleSubscriber.java:77)
	at reactor.core.publisher.Mono.block(Mono.java:1494)
	at org.springframework.data.redis.connection.ReactiveRedisConnection.close(ReactiveRedisConnection.java:60)
	at org.springframework.boot.actuate.redis.RedisReactiveHealthIndicator.lambda$doHealthCheck$1(RedisReactiveHealthIndicator.java:50)
	at reactor.core.publisher.FluxDoFinally$DoFinallySubscriber.runFinally(FluxDoFinally.java:156)
	... 203 common frames omitted
```

While running a Spring Cloud Gateway application with Redis as session store, I observed the stack trace above. The cause is a blocking call in: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/redis/RedisReactiveHealthIndicator.java#L49

There is a `Mono<Void> closeLater();` https://github.com/spring-projects/spring-data-redis/blob/master/src/main/java/org/springframework/data/redis/connection/ReactiveRedisConnection.java#L68 which could be called instead.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/redis/RedisReactiveHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/redis/RedisReactiveHealthIndicator.java
@@ -19,6 +19,7 @@ package org.springframework.boot.actuate.redis;
 import java.util.Properties;
 
 import reactor.core.publisher.Mono;
+import reactor.core.scheduler.Schedulers;
 
 import org.springframework.boot.actuate.health.AbstractReactiveHealthIndicator;
 import org.springframework.boot.actuate.health.Health;
@@ -31,6 +32,7 @@ import org.springframework.data.redis.connection.ReactiveRedisConnectionFactory;
  *
  * @author Stephane Nicoll
  * @author Mark Paluch
+ * @author Artsiom Yudovin
  * @since 2.0.0
  */
 public class RedisReactiveHealthIndicator extends AbstractReactiveHealthIndicator {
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR, @ayudovin. As [mentioned](https://github.com/spring-projects/spring-boot/issues/16705#issuecomment-489006380) by @mp911de, we also need to change how connections are obtained. Of the two suggestions that Mark's made, I think the use of the parallel scheduler is the one to go for. Would you like to update your PR to that effect?

mp911de:
I filed [DATAREDIS-977](https://jira.spring.io/browse/DATAREDIS-977) to expose non-blocking connect methods. Right now, `Mono.fromSupplier().subscribeOn(Schedulers.parallel())` or similar should be a good compromise until we can provide `Mono<ReactiveRedisConnection>` methods.

ayudovin:
@wilkinsona, I would like to update PR

AndreasKl:
The `Mono` returned via `closeLater` is unhandled, if it would be lazy the close call would never happen.

philwebb:
Thanks once again @ayudovin!

PR #16758, opened 2019-05-09
Add negative tests to ConfigurationPropertiesScanTests

izeye:
This PR adds negative tests to `ConfigurationPropertiesScanTests` as they seem to be missing.


snicoll:

Issue #16849, opened 2019-05-14
Flyway and Liquibase auto configurations require DataSource bean to be declared


L00kian:
Hello,

`LiquibaseAutoConfiguration` and `FlywayAutoConfiguration` both require `DataSource` bean to be declared. However both configurations allow to explicitly define datasource properties which would create one anyway.

Context: I've been playing around with r2dbc and tried to assemble an application with it + flyway. was not able to apply migrations as there is no need for datasource when using r2dbc (also don't really fancy the idea that i need one living in context just for migrations spring-projects/spring-data-r2dbc#40 which is not directly related but explains why)

Regards,
Ilya

philwebb:
Thanks for the PR, I'll close this issue and we can continue the discussion on #16850

PR #16850, opened 2019-05-14
Allow Flyway or Liquibase to be auto-configured without a DataSource bean

L00kian:
`LiquibaseAutoConfiguration` and `FlywayAutoConfiguration` both require `DataSource` bean to be declared. However both configurations allow to explicitly define datasource properties which would create one anyway. #16849


philwebb:
We might refine the condition to check for Datasource or the URL property when we merge.

wilkinsona:
Thank you very much for making your first contribution to Spring Boot, @L00kian. The changes have now been merged into master with [some modifications](https://github.com/spring-projects/spring-boot/commit/880721557bf92e4b230f5f2cfff71e23bbfbfcc9). As suggested above by @philwebb, we now look for either a `DataSource` or a `spring.(flyway|liquibase).url` property. The auto-configuration backs off is neither is available. This reduces the scope of the changes as the custom exceptions and failure analyzers are no longer required. You can see combined effect of your changes and the modifications in the diff of the [merge commit](https://github.com/spring-projects/spring-boot/commit/960c03dee5e56e5371ed5f297118f090d7248ba7).

L00kian:
@wilkinsona Thank you for accepting it. I was glad to contribute

PR #16852, opened 2019-05-14
Do not start Tomcat webserver if the port is set to <0 value

JigarJoshi:
Like [Jetty](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyWebServer.java#L136) & [Undertow](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/undertow/UndertowWebServer.java#L96)  web server wrapper. Do not start Tomcat web server instance if port is negative.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
@@ -87,6 +87,9 @@ public class TomcatWebServer implements WebServer {
 	}
 
 	private void initialize() throws WebServerException {
+		if(!autoStart){
+			return;
+		}
 		logger.info("Tomcat initialized with port(s): " + getPortsDescription(false));
 		synchronized (this.monitor) {
 			try {
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
@@ -87,7 +87,7 @@ public class TomcatWebServer implements WebServer {
 	}
 
 	private void initialize() throws WebServerException {
-		if(!autoStart){
+		if (!this.autoStart) {
 			return;
 		}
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@JigarJoshi Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16852)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16852) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@JigarJoshi Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=16852)!

philwebb:
I'm pretty sure we intentionally start the server because we usually need to trigger the `ServletContextInitializer` passed to `ServletWebServerFactory`. I've flagged this for team attention to see if anyone else remembers the history of `autostart`.

wilkinsona:
Things are a bit inconsistent at the moment. Without the change proposed here, the behaviour is as follows when starting with `server.port=-1`:

- Jetty: app starts and stays running, but is not listening for HTTP connections
- Netty: app fails to start as -1 is an illegal port number
- Tomcat: app starts and stays running, but is not listening for HTTP connections 
- Undertow: app starts and then stops

With the change proposed here an app using Tomcat fails to start due to a missing `ServletContext`. Note that the proposed change returns early from `initialize()` whereas Jetty and Undertow return early from `start()`.

With Tomcat modifier to return early from `start()` we get into a strange situation where calling `getPort()` on `TomcatWebServer` causes Tomcat to create a `Connector` that listens on 8080 as it has no connectors at that point.

Further to the above, the way that `autoStart` is implemented at the moment is a little odd. When a web server is created with `autoStart` set to `false`, calling `start()` on that server will then have no effect and there's no way to set `autoStart` back to `true` so that the web server can be started. In other words, it's not really disabling automatic start, but disabling start altogether.

Thanks very much for the PR, but I think we need to take a step back, decide how we want things to behave across all 4 containers, and then figure out how and when we can get there. I've opened https://github.com/spring-projects/spring-boot/issues/17073 to track that effort.

Issue #16828, opened 2019-05-13
Maven plugin may repackage an archive twice, resulting in a StackOverflowError at runtime


snicoll:
It is possible that `Repackager` can run on an already repackaged archive. We have a check in place that skips repackaging if the source and the target File is identical and the source is a repackaged archive.

If the source does not match the target, repackaging will occur, reading the main class from the manifest of the source. Doing so means the target repackaged archive uses  `JarLauncher` as "main app" which will create a StackOverflow.

This is basically #16202 and what happens when you run repackaging twice with and without a `finalName`.

ielatif:
@snicoll I am interested to work on it.

> We have a check in place that skips repackaging if the source and the target File is identical and the source is a repackaged archive.

here is the check you referenced :
https://github.com/spring-projects/spring-boot/blob/36c1c051b8106371d37f8a747cea0e369a8e5c84/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java#L184-L186

I think keeping the check that source is already repackaged and removing the check for source and destination are identical will solve it.

The check for source and destination are identical will be still handled here :
https://github.com/spring-projects/spring-boot/blob/36c1c051b8106371d37f8a747cea0e369a8e5c84/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java#L187-L191

snicoll:
> @snicoll I am interested to work on it.

Thank you for offering to help. I am not sure yet how to proceed with this one but if you want to give it a try, feel free to do do so.

ielatif:
@snicoll I approached this issue with multiple ways and found that repackage detection works fine.

I think the issue here is to allow multiple executions even those that runs multiple times.

I ended up with this solution : 

```
if is already repackaged : 

    if backup file exists then repackage from backup file

    otherwise repackage from original jar
```

https://github.com/ielatif/spring-boot/blob/93640c5b71e26300370759f5f3a169a3fba45106/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java#L185-L196

snicoll:
We've been looking at this one and were unable to reproduce it. A double execution plus exploiting a bug in Maven makes it quite unlikely to happen in practice but we can reopen if we manage to reproduce it with a recent version.

Issue #15034, opened 2018-10-31
spring-boot-maven-plugin issue since 2.1.0: Source must refer to an existing file, got...


wouterv:
Since the release of 2.1.0.RELEASE, we get the following issue in our build:

````
build	31-Oct-2018 15:08:38	[ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.1.0.RELEASE:repackage (build-repackaged-jar) on project application: Execution build-repackaged-jar of goal org.springframework.boot:spring-boot-maven-plugin:2.1.0.RELEASE:repackage failed: Source must refer to an existing file, got /atlassian/bamboo/agent/5.15.3/xml-data/build-dir/BC-BGS-JOB1/repo/application/target/gear-exec-linux.jar.original -> [Help 1]
````

We fixed the plugin version to 2.0.6.RELEASE and all works fine again. 

This is our specific part in the pom, the special thing about is that it creates two artifacts:

```xml
 <plugin>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-maven-plugin</artifactId>
    <version>2.0.6.RELEASE</version> <!-- remove this and it will fail -->
    <executions>
        <execution>
            <id>build-linux-executable</id>
            <goals>
                <goal>repackage</goal>
            </goals>
            <configuration>
                <classifier>exec-linux</classifier>
                <executable>true</executable>
                <attach>false</attach>
            </configuration>
        </execution>    
        <execution>
            <id>build-repackaged-jar</id>
            <goals>
                <goal>repackage</goal>
            </goals>
            <configuration>
                <executable>false</executable>
                <attach>true</attach>
            </configuration>
        </execution>
    </executions>
</plugin>
```

wilkinsona:
Thanks for the pom snippet, but I've been unable to reproduce the problem with it. As far as I can tell, the Maven plugins from 2.0.6 and 2.1.0 behave in the same way with that configuration. I have been testing with Maven 3.5.4. If you'd like us to spend some more time on this, can you please spend the time to produce a complete, minimal example that reproduces the behaviour you have described.

snicoll:
@wouterv I am not 100% sure but this looks related to #1421 . Can you please double check? I've also updated [the release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.1-Release-Notes#maven-plugin) to mention the new `repackage` execution id.

wouterv:
[...]

PR #16873, opened 2019-05-15
make repackager thread-safe

ielatif:
This commit makes the repackager thread-safe when it is invoked concurrently in the same lifecycle. Closes gh-16828

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java
@@ -44,6 +44,7 @@ import org.springframework.util.StringUtils;
  * @author Phillip Webb
  * @author Andy Wilkinson
  * @author Stephane Nicoll
+ * @author Issam El-atif
  */
 public class Repackager {
 
@@ -168,7 +169,7 @@ public class Repackager {
 	 * @throws IOException if the file cannot be repackaged
 	 * @since 1.3.0
 	 */
-	public void repackage(File destination, Libraries libraries,
+	public synchronized void repackage(File destination, Libraries libraries,
 			LaunchScript launchScript) throws IOException {
 		if (destination == null || destination.isDirectory()) {
 			throw new IllegalArgumentException("Invalid destination");
@@ -181,7 +182,7 @@ public class Repackager {
 		}
 		destination = destination.getAbsoluteFile();
 		File workingSource = this.source;
-		if (alreadyRepackaged() && this.source.equals(destination)) {
+		if (alreadyRepackaged()) {
 			return;
 		}
 		if (this.source.equals(destination)) {
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/test/java/org/springframework/boot/loader/tools/RepackagerTests.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/test/java/org/springframework/boot/loader/tools/RepackagerTests.java
@@ -60,6 +60,7 @@ import static org.mockito.Mockito.mock;
  *
  * @author Phillip Webb
  * @author Andy Wilkinson
+ * @author Issam El-atif
  */
 public class RepackagerTests {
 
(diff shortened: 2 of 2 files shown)


spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java
```diff
@@ -168,7 +169,7 @@ public void repackage(File destination, Libraries libraries) throws IOException
 	 * @throws IOException if the file cannot be repackaged
 	 * @since 1.3.0
 	 */
-	public void repackage(File destination, Libraries libraries,
+	public synchronized void repackage(File destination, Libraries libraries,
```
snicoll: Can you expand how thread safety relates to the issue you are referring to?

spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java
```diff
@@ -181,7 +182,7 @@ public void repackage(File destination, Libraries libraries,
 		}
 		destination = destination.getAbsoluteFile();
 		File workingSource = this.source;
-		if (alreadyRepackaged() && this.source.equals(destination)) {
+		if (alreadyRepackaged()) {
```
snicoll: I think it's a bit more complicated than that. What that check was doing before your change was checking a case where a user tries to repackage an archive that was already repackaged with a replacement of the source jar.

If the source and the destination are different and the source is already repackaged, I think there's something to be done (like adding a warning to indicate we can't repackage as the source is already a fat jar.

ielatif:
I tested it with the demo app (demo.zip provided in https://github.com/spring-projects/spring-boot/issues/15034)  running :

> mvn clean package install 

 and it works

snicoll:
> and it works

Thanks for the feedback. I still don't understand how`synchronized` is relevant. Can you please clarify that?

spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java
ielatif: The check `&& this.source.equals(destination)` was added to fix #15034
I looked at the pom.xml provided in the sample app : 

```
<plugin>
	<version>2.1.0.RELEASE</version><!-- change to 2.0.6.RELEASE -->
	<executions>
		<execution>
			<id>build-linux-executable</id>
			<goals>
			        <goal>repackage</goal>
			</goals>
			<configuration>
				<classifier>exec-linux</classifier>
				<executable>true</executable>
				<attach>false</attach>
			</configuration>
		</execution>  
		<execution>
			<id>build-repackage-executable</id>
			<goals>
				<goal>repackage</goal>
			</goals>
			<configuration>
				<executable>false</executable>
				<attach>true</attach>
			</configuration>
		</execution>
	</executions>
</plugin>
```
and from the error message 

`[ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.1.0.RELEASE:repackage (build-repackage-executable) on project demo: Execution build-repackage-executable of goal org.springframework.boot:spring-boot-maven-plugin:2.1.0.RELEASE:repackage failed: Source must refer to an existing file, got C:\Users\issam\Downloads\demo\target\gear-exec-linux.jar.original`

I suspected that `build-repackage-executable` execution was running before `build-linux-executable`  execution ends.

Then I decided to give it a try and tested it with the provided sample

spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/Repackager.java
ielatif: I disabled the check 'source.equals(destination)' and created a test with identicals source and destination that runs repackage twice to reproduce the issue but it didn't fail. I said to myself may be it passed because my test is sequential!

ielatif:
@snicoll I was misleading It can not be a concurrency issue because `RepackageMojo` is thread-safe.
Sorry for the deflection.

snicoll:
No need to be sorry and thank you for contributing!

PR #16874, opened 2019-05-15
Improve notes for the import into IntelliJ IDEA

rstoyanchev:
I suffered the consequences of ignoring recommendation to increase memory, hoping that if there is an issue I will surely see it. The silent failure however makes it very confusing. Hence my suggestion here to use a stronger language.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -201,13 +201,14 @@ need to import the root `spring-boot` pom and the `spring-boot-samples` pom sepa
 
 
 === Importing into IntelliJ IDEA
-The separate Maven import process that IntelliJ IDEA uses to parse a project structure
-requires more memory. Before importing the project, we recommend increasing the memory
-allocation in `Preferences | Build, Execution, Deployment | Build Tools | Maven |
-Importing`. Set the `VM options for importer` option to `-Xmx2g`.
+**Please, do this first!**
+Go to `Preferences | Build, Execution, Deployment | Build Tools | Maven | Importing`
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.adoc
```diff
@@ -201,13 +201,14 @@ need to import the root `spring-boot` pom and the `spring-boot-samples` pom sepa
...
 
 === Importing into IntelliJ IDEA
-The separate Maven import process that IntelliJ IDEA uses to parse a project structure
-requires more memory. Before importing the project, we recommend increasing the memory
-allocation in `Preferences | Build, Execution, Deployment | Build Tools | Maven |
-Importing`. Set the `VM options for importer` option to `-Xmx2g`.
-
-To open the project in IntelliJ IDEA, select "`File`" -> "`Open`" and then click on the
-root `pom.xml`.
+**Please, do this first!!!**
```
snicoll: I am not a huge fan of the `!!!` part. Could we find a different way to emphasis this perhaps?

CONTRIBUTING.adoc
```diff
@@ -201,13 +201,14 @@ need to import the root `spring-boot` pom and the `spring-boot-samples` pom sepa
...
+Go to `Preferences | Build, Execution, Deployment | Build Tools | Maven | Importing`
+and set `VM options for importer` to `-Xmx2g` to allocate sufficient memory for IDEA's
```
snicoll: `IDEA` isn't a proper product name. We've actually been asked several times to align to "IntelliJ IDEA".

CONTRIBUTING.adoc
rstoyanchev: Updated to `!`

PR #16879, opened 2019-05-16
Correct expansion of jOOQ in the reference docs

lukaseder:
Fix jOOQ's trademark


lukaseder:
Possibly, this is an issue elsewhere as well - I'm not sure how to search all the sources in Spring Boot for such references...

philwebb:
Thanks, we can double check other areas as well but I think we mostly just refer to JOOQ and don't expand on what it means.

PR #16885, opened 2019-05-17
Restore HttpTraceAutoConfigurationTests.configuresServletFilter()

izeye:
This PR restores `HttpTraceAutoConfigurationTests.configuresServletFilter()` as it seems lost in de128fed20d7ee7f569e7219ba734ec0ea6de4cd accidentally.


snicoll:
Good catch once again. Thank you so much @izeye

PR #16894, opened 2019-05-17
Fix persistent session property name

gavvvr:


wilkinsona:
Well spotted! Thanks very much, @gavvvr. We have the same problem in the 1.5.x docs too.

wilkinsona:
Thanks again, @gavvvr. The proposed changes have been merged into 1.5.x and forwards into 2.0.x, 2.1.x, and master.

Issue #16822, opened 2019-05-10
@ConfigurationPropertiesScan should import ConfigurationPropertiesBindingPostProcessorRegistrar


mbhave:
Configuration properties are bound via the `ConfigurationPropertiesBindingPostProcessor` which are registered by `ConfigurationPropertiesBindingPostProcessorRegistrar`. This registrar is imported by `@EnableConfigurationProperties` which happens to get switched on if an auto-configuration that has the annotation is switched on or if the user themselves adds the `@EnableConfigurationProperties` annotation. 

In a situation where auto-configuration is disabled or no auto-configuration class with `@EnableConfigurationProperties` kicks in (unlikely), user-defined `@ConfigurationProperties` registered via configuration properties scanning won't get bound.

wilkinsona:
If we get [this change](https://github.com/spring-projects/spring-framework/issues/22995) in Framework, this will become unnecessary.

snicoll:
The related issue has been fixed differently and we still have `@ConfigurationPropertiesScan` and friends so revisiting this issue is in order.

PR #16898, opened 2019-05-17
@ConfigurationPropertiesScan should import ConfigurationPropertiesBindingPostProcessorRegistrar

ayudovin:
this [task](https://github.com/spring-projects/spring-boot/issues/16822)

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/ConfigurationPropertiesScan.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/ConfigurationPropertiesScan.java
@@ -37,7 +37,8 @@ import org.springframework.core.annotation.AliasFor;
 @Target(ElementType.TYPE)
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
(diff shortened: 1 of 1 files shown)


mbhave:
If there is a [framework change](https://github.com/spring-projects/spring-framework/issues/22995) that allows `@ConfigurationProperties` to be scanned by component scanning, we will not need the custom scanning feature. Placing this on hold till that's decided.

ayudovin:
@snicoll, I'm not sure what I need to check by a test. Could you please provide a little bit more details or something which can help me? I'd appreciate it.

snicoll:
Thanks for the feedback. What I had in mind is a test that showcases the situation that @mbhave has described in the related issue. So a simple context with no auto-configuration at all and just `@ConfigurationPropertiesScan` should not bind a scanned object without your fix. Then you should apply it and validate the test passes.

Let me know if you have another question.

snicoll:
Thank you for the PR @ayudovin but I've started to work on a test and will follow-up on the original issue.

ayudovin:
@snicoll, ok, no problem, sorry for it, I just could not find the time.

Issue #12538, opened 2018-03-19
/tmp/spring.log file is generating on spring boot service startup e


manjunadh:
Hi,
I am using spring boot 1.59 version.

I have `logback-spring.xml` with proper configurations still spring boot application is generating `/tmp/spring.log` file on application start up.

Below are my `logback-spring.xml` configurations.

```xml
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
	<include resource="org/springframework/boot/logging/logback/base.xml"/>
	<jmxConfigurator/>
	
	<property name="CONSOLE_LOG_PATTERN"
        value="%d{MM/dd/yyyy HH:mm:ss.SSS XXX} ${LOG_LEVEL_PATTERN:-%5p} ${PID:- } -- [%thread] %logger{60} : %msg%n${LOG_EXCEPTION_CONVERSION_WORD:-%wEx}"/>

	<!-- Appender to log to file -->
	<springProfile name="local,bamboo">​
	<appender name="file" class="ch.qos.logback.core.rolling.RollingFileAppender">
		<file>logs/service.log</file>
		<rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
		  <!-- daily rollover -->
		  <fileNamePattern>logs/service.%d{yyyy-MM-dd}_%i.gz</fileNamePattern>
		  <timeBasedFileNamingAndTriggeringPolicy
			  class="ch.qos.logback.core.rolling.SizeAndTimeBasedFNATP">
			  <!-- or whenever the file size reaches 10MB -->
			<maxFileSize>10MB</maxFileSize>
		  </timeBasedFileNamingAndTriggeringPolicy>
		  <!-- keep 7 days' worth of history -->
		  <maxHistory>7</maxHistory>
		</rollingPolicy>
		<encoder>
			<pattern>${CONSOLE_LOG_PATTERN}</pattern>
			<charset>utf8</charset>
		</encoder>
	</appender>
	<root >
		<appender-ref ref="file" level="INFO" />
	</root>
​	</springProfile>
	<springProfile name="development,qa1,qa4,pp,prod">​
	<appender name="file" class="ch.qos.logback.core.rolling.RollingFileAppender">
		<file>/apps/services/logs/inventory/service.log</file>
		<rollingPolicy class="ch.qos.logback.core.rolling.TimeBasedRollingPolicy">
		  <!-- daily rollover -->
[...]

PR #16901, opened 2019-05-19
Clarify Logback XML

michaldo:
Simple logging setup is configurable by application properties.

"To configure the more fine-grained settings of a logging system, you
need to use the native configuration format supported by the
LoggingSystem in question"

In other words Spring Boot provides default logging configuration +
application properties to modify. If application has to customize
something not covered by application properties, it must switch to XML
configuration.

Unfortunately, naive switching to XML cause lost default Spring Boot
configuration.

But there is an impression, that the lost can be avoided. It is possible
tweak low level logging detail by XML and keep default Spring Boot
configuration working. It is enough to import files 'base.xml' or
'default.xml'.

The impression is wrong.

'defaults.xml' does not setup console nor file logging, just setup
levels and properties.

'base.xml' by default logs to file '/tmp/spring.log', what is different
than default behavior that nothing is logged to a file.

Intention of my commit is to express that files 'base.xml' and
'defaults.xml' are not equivalent of default Spring Boot configuration

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
@@ -1543,14 +1543,14 @@ using the "logging.config" property.
 === Configure Logback for Logging
 If you put a `logback.xml` in the root of your classpath, it is picked up from there (or
 from `logback-spring.xml`, to take advantage of the templating features provided by
-Boot). Spring Boot provides a default base configuration that you can include if you
+Boot). Spring Boot provides a default level configuration that you can include if you
 want to set levels, as shown in the following example:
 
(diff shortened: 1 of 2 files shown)


michaldo:
See also:
https://github.com/spring-projects/spring-boot/issues/12538
https://stackoverflow.com/questions/50386520/spring-boot-spring-log-under-tmp

philwebb:
I agree that that section is a little confusing. I think we ideally need to go a little further than your suggested edits and describe clearly the difference between `base.xml` and `default.xml`.

philwebb:
Thanks very much for the PR @michaldo. This is now in 2.1.x and 2.2.x. I've also extended the documentation a little to try and clarify when the different include files should be used.

PR #16905, opened 2019-05-20
Exclude AWS Redshift from DatabaseDriverClassNameTests

dreis2211:
Hi,

apparently 146f35d broke the current master as the new driver is not excluded from DatabaseDriverClassNameTests. For simplicity reasons, I simply excluded it for now.

Cheers,
Christoph


snicoll:
Thank you @dreis2211 and @wilkinsona for catching up my faulty review!

wilkinsona:
Thank you, @dreis2211, but I'd already fixed this locally before I saw your PR. See https://github.com/spring-projects/spring-boot/commit/f8eb230dea9e9738d3f974d19ca884dd2689e0ad.

PR #16933, opened 2019-05-21
added sample app that demonstrates a basic use of spring-boot-starter-mail

Tetsuya3850:
I've written a sample app for spring-boot-starter-mail since it was lacking in the collection of samples.

Also added tests and checked that it works, but since I cannot include my own email and password, I’ve @Ignore the tests.


philwebb:
Thanks for the PR but I'm afraid the samples are somewhat poorly named and we primarily use them as a form of integration testing. We're not keen to add any additional ones at this point.

It might be better to instead consider contributing this to https://spring.io/guides (via https://github.com/spring-guides/getting-started-guides) where it will have more visibility.

PR #16946, opened 2019-05-22
Fix regular expression used in devools documentation

phillipuniverse:
In the docs for customizing the Spring Boot Devtools RestartClassLoader, the example regex captures jars named mycorp-common-abc.jar but does not capture normal Maven versioned jars that would appear in BOOT-INF, like mycorp-common-1.0.4-GA.jar or mycorp-common-1.0.5-SNAPSHOT.jar.

This updates the regex to have a more sensible default for users that copy/paste the example into their own project.


philwebb:
This is now in 2.1.x and master. Thanks very much @phillipuniverse!

PR #16971, opened 2019-05-25
Spring Data Cassandra auto-configuration should be conditional on Cluster bean

nosan:
`@ConditionalOnBean(Cluster.class)` is not present on class `CassandraDataAutoConfiguration`.


philwebb:
Thanks @nosan, this is now merged to 2.1.x and master.

PR #16974, opened 2019-05-27
Pass resolvedType into isOfDifferentType() in JavaBeanBinder.Bean.get()

izeye:
This PR changes to pass `resolvedType` into `isOfDifferentType()` in `JavaBeanBinder.Bean.get()` as [it could be reassigned](https://github.com/spring-projects/spring-boot/blob/15bdc12335f942c030757e028ee31adb00befd37/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/JavaBeanBinder.java#L211).


philwebb:
Thanks @izeye

Issue #16416, opened 2019-04-02
Migrate from NCSARequestLog to CustomRequestLog


wilkinsona:
In Jetty 9.4.15, `NSCARequestLog` has been deprecated and `CustomRequestLog` introduced as its replacement. We need to migrate to `CustomRequestLog`, making any necessary property updates when we do so. Assigning to 2.2.x for now but, depending on the updates that are necessary, we may want to do this in one or more of our current maintenance branches.

umeshrawat:
@wilkinsona Can I take this up?

wilkinsona:
Thanks for the offer. We're not yet sure what needs to be done. If you'd like to spend some time looking at the differences between `NSCARequestLog` and `CustomRequestLog` and seeing how well our current properties map to the latter, that would be much appreciated.

bclozel:
I've had a look and it seems that all properties related to the `RequestLog` itself are now gone in favor of the format string. Properties related to the writer are still available.

```
# Spring Boot support
server.jetty.accesslog.enabled=false # Enable access log.

# Log writer related, still there with RequestLogWriter
server.jetty.accesslog.append=false # Append to log.
server.jetty.accesslog.file-date-format= # Date format to place in log file name.
server.jetty.accesslog.filename= # Log filename. If not specified, logs redirect to "System.err".
server.jetty.accesslog.retention-period=31 # Number of days before rotated log files are deleted.

# All these keys are gone
server.jetty.accesslog.date-format=dd/MMM/yyyy:HH:mm:ss Z # Timestamp format of the request log.
server.jetty.accesslog.extended-format=false # Enable extended NCSA format.
server.jetty.accesslog.locale= # Locale of the request log.
server.jetty.accesslog.log-cookies=false # Enable logging of the request cookies.
server.jetty.accesslog.log-latency=false # Enable logging of request processing time.
server.jetty.accesslog.log-server=false # Enable logging of the request hostname.
server.jetty.accesslog.time-zone=GMT # Timezone of the request log.
```

I think we should deprecate all the removed keys in favor of new `format` and `customFormat` keys:

```
server.jetty.accesslog.format=NCSA # enum with predefined formats: NCSA, EXTENDED_NCSA
server.jetty.accesslog.custom-format= # String, custom format overriding the former if defined 
``` 

[...]

PR #16987, opened 2019-05-28
Avoid using deprecated NCSARequestLog

dreis2211:
Hi,

while #16416 migrated away from most of the usages of `NCSARequestLog`, it seems to have missed one occurrence in `ServletManagementChildContextConfiguration`.

Unfortunately, the parent-child context topic doesn't seem to be covered by a lot of tests and my time is limited at the moment to improve this (outside of this issue). But feel free to not merge this, because of the missing tests. 

Cheers,
Christoph


philwebb:
Thanks @dreis2211, the lack of tests isn't ideal but that's really our problem. I still think it's worth merging this. Cheers!

Issue #16206, opened 2019-03-12
Binder can fail when multiple overloaded setters exist


Walliee:
If a `ConfigurationProperties` class has a `String` property but two setters for that property, one that takes a `String` and sets it and another that takes in a `int` coverts value to a `String` and then sets it.

```
@ConfigurationProperties("app")
public class ApplicationSettings {
    private String property;

    public void setProperty(int property) {
        this.property = String.valueOf(property);
    }

    public void setProperty(String property) {
        this.property = property;
    }
}
```
and the properties files has a `String` value
```
app.property=string
```
At startup, the `Binder` sometimes uses the setter with the `int` param and fails with an error as follows:
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Failed to bind properties under 'spring.cloud.stream.kafka.binder.required-acks' to int:

    Property: spring.cloud.stream.kafka.binder.requiredacks
    Value: all
    Origin: class path resource [application.properties]:1:47
    Reason: failed to convert java.lang.String to int

Action:

Update your application's configuration
```

Turning on debug has the following stacktrace:
```
2019-03-12 02:49:07.915 DEBUG 30020 --- [           main] o.s.b.d.LoggingFailureAnalysisReporter   : Application failed to start due to an exception

org.springframework.boot.context.properties.bind.BindException: Failed to bind properties under 'app.property' to int
	at org.springframework.boot.context.properties.bind.Binder.handleBindError(Binder.java:249) ~[spring-boot-2.1.3.RELEASE.jar:2.1.3.RELEASE]
	at org.springframework.boot.context.properties.bind.Binder.bind(Binder.java:225) ~[spring-boot-2.1.3.RELEASE.jar:2.1.3.RELEASE]
	at org.springframework.boot.context.properties.bind.Binder.lambda$bindBean$3(Binder.java:336) ~[spring-boot-2.1.3.RELEASE.jar:2.1.3.RELEASE]
	at org.springframework.boot.context.properties.bind.JavaBeanBinder.bind(JavaBeanBinder.java:83) ~[spring-boot-2.1.3.RELEASE.jar:2.1.3.RELEASE]
	at org.springframework.boot.context.properties.bind.JavaBeanBinder.bind(JavaBeanBinder.java:72) ~[spring-boot-2.1.3.RELEASE.jar:2.1.3.RELEASE]
[...]

PR #17005, opened 2019-05-29
Handle all getters first in JavaBeanBinder.addProperties()

izeye:
This PR changes to handle all getters first in `JavaBeanBinder.addProperties()` to align with the description in https://github.com/spring-projects/spring-boot/issues/16206#issuecomment-492495443.


philwebb:
Oh wow! Another save from @izeye. Thanks once again!!

Issue #15078, opened 2018-11-03
RestTemplateBuilder.basicAuth causes the entire body to be read into memory


philwebb:
See https://github.com/spring-io/artifactory-resource/issues/27 for background. I'm not sure if we need to just document this as a warning or look for an alternative way to add the header.

wilkinsona:
This feels to me like it should be fixed/enhanced in Framework so that there’s a mechanism for setting request headers without buffering the whole body.

philwebb:
Ive raised [SPR-17470](https://jira.spring.io/browse/SPR-17470) to see what Framework can do. I'll leave this one open but blocked since we'll probably need changes regardless.

mbhave:
Given that the framework issue has been declined, we should decide if we just want to document this or do something else to prevent an OOM error.

philwebb:
I think I can probably port https://github.com/spring-io/artifactory-resource/commit/62215780e903bac4eb13c7b5f4ccf4aba8deb540

philwebb:
Closing in favor of PR #17010

PR #17010, opened 2019-05-29
Update `RestTemplateBuilder` and `TestRestTemplate` to use BasicAuthenticationClientHttpRequestFactory instead of Interceptor

nosan:
Update `RestTemplateBuilder` and `TestRestTemplate` to use a custom request factory to add
authentication headers.

Prior to this commit, the `RestTemplateBuilder` and `TestRestTemplate` used the
`BasicAuthenticationInterceptor` to add headers.

see gh-15078


philwebb:
Nice one! Thanks.

philwebb:
Thanks for another contribution @nosan !! This has now been merged to master. I've add a [polish commit](https://github.com/spring-projects/spring-boot/commit/76e075ddd02d011b1c02cb53ad98fb042c8259ea) that makes some of the new classes package private just to reduce the surface area of the public API.

PR #17014, opened 2019-05-29
Autoconfigure custom OAuth2RequestValidator bean

markhobson:
It is currently awkward to use a custom `OAuth2RequestValidator` when using the default `AuthorizationServerConfigurer`. Specifying a custom configurer to use the `AuthorizationServerEndpointsConfigurer` disables all other autoconfiguration.

This change automatically configures any registered `OAuth2RequestValidator` bean on the `AuthorizationServerEndpointsConfigurer`.


mbhave:
Thanks for the PR. This applies only to Spring Boot 1.5.x which is in maintenance mode and will EOL this August. I'm not too keen on adding this since the related code now lives in a [separate project](https://github.com/spring-projects/spring-security-oauth2-boot). Let's see what the rest of the team thinks.

markhobson:
Good point, I can resubmit the PR targetting `spring-security-oauth2-autoconfigure` instead if that'd be preferable?

mbhave:
@markhobson I'd suggest raising an issue in `spring-security-oauth2-autoconfigure` before creating a PR so that your efforts aren't wasted in case the Spring Security team has reasons for not adding this.

markhobson:
Sure, no probs. Thanks!

mbhave:
Thanks! I'll go ahead and close this one.

PR #16627, opened 2019-04-23
EmbeddedMongoAutoConfiguration should depend on bean by type

ielatif:
PR for https://github.com/spring-projects/spring-boot/issues/16540

PR #17020, opened 2019-05-30
Allow AbstractDependsOnBeanFactoryPostProcessor to identify depended on beans by type

nosan:
Add to `AbstractDependsOnBeanFactoryPostProcessor` new constructors to specify that bean and factory bean depends on particular types.

Prior to this commit, the `AbstractDependsOnBeanFactoryPostProcessor`
can be used only with bean names.

It could be useful in case of gh-16627


ielatif:
> It could be useful in case of [gh-16627](https://github.com/spring-projects/spring-boot/pull/16627)

👍

wilkinsona:
Thanks very much, @nosan. If you're interested, I polished things a little bit in https://github.com/spring-projects/spring-boot/commit/80650f485d31600d7c44acf934886c39fb154047 to avoid the logic branches for dealing with `String[]` vs `Class[]`.

nosan:
Thank you very much @wilkinsona

Issue #16943, opened 2019-05-22
Provide property to control ConfigureRedisAction


OrangeDog:
See [Spring Session docs](https://docs.spring.io/spring-session/docs/current/reference/html5/#api-redisoperationssessionrepository-sessiondestroyedevent) and https://github.com/spring-projects/spring-session/issues/124#issuecomment-71490616

Adding a switch for this in `RedisProperties` would be very useful.

philwebb:
I don't think we should add a property to `RedisProperties` since `ConfigureRedisAction` is specific to Spring Session. I think `SessionProperties` would be a better place, probably under a `redis` specific name. Something like `spring.session.redis.configure=` with `NOTIFY_KEYSPACE_EVENTS` and `NONE` as enum options.

We should also make sure that the user can register their own `ConfigureRedisAction` bean if they need to.

izeye:
This seems resolved via #17022.

wilkinsona:
Thank you, @izeye.

PR #17022, opened 2019-05-30
Add `configuration-strategy` property to RedisSessionProperties  to configure `ConfigureRedisAction`

nosan:
see gh-16943

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/RedisSessionProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/RedisSessionProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2017 the original author or authors.
+ * Copyright 2012-2019 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,6 +41,11 @@ public class RedisSessionProperties {
 	private RedisFlushMode flushMode = RedisFlushMode.ON_SAVE;
 
 	/**
+	 * Allows specifying a strategy for configuring and validating Redis.
+	 */
+	private ConfigurationStrategy configurationStrategy = ConfigurationStrategy.NOTIFY_KEYSPACE_EVENTS;
+
+	/**
 	 * Cron expression for expired session cleanup job.
 	 */
 	private String cleanupCron = DEFAULT_CLEANUP_CRON;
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/RedisSessionProperties.java
```diff
@@ -40,6 +40,11 @@
 	 */
 	private RedisFlushMode flushMode = RedisFlushMode.ON_SAVE;
 
+	/**
+	 * Allows specifying a strategy for configuring and validating Redis.
+	 */
+	private RedisAction action = RedisAction.NOTIFY_KEYSPACE_EVENTS;
```
snicoll: The action interface is quite generic and allows the user to do anything they want to configure and validate the connection. I don't think "action" is the right term for this as what we set is a "configuration and validation strategy". 

`configurationStrategy` is a bit mouthful but more explicit perhaps?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/RedisSessionProperties.java
```diff
@@ -69,4 +74,29 @@ public void setCleanupCron(String cleanupCron) {
...
+	}
+
+	public void setAction(RedisAction action) {
+		this.action = action;
+	}
+
+	/**
+	 * Allows specifying a strategy for configuring and validating Redis.
+	 */
+	public enum RedisAction {
```
snicoll: `ConfigurationStrategy` would be the logical name there but it fees a bit odd so a bit more brainstorming is required I think.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/RedisSessionProperties.java
vpavic: Having in mind the purpose of `ConfigureRedisAction`, naming this property `configurationStrategy` would be more appropriate.

nosan:
@snicoll @vpavic 
Thank you for your feedback.

PR has been updated.

philwebb:
Thanks for another contribution @nosan! 

I went back and forth on the property name a few times and ended up with `configure-action` in the end. It's a hard one to name because we ideally want some clues that it relates to configuring the Redis server, but we also want to indicate that it's us doing the configuration and to also somehow link back to the `ConfigureRedisAction` interface. I also managed to mess up the polish commit so there's an [additional one](https://github.com/spring-projects/spring-boot/commit/e0aa90cf32e2ae71885271c5645a8b468e46cbec) outside of the [merge commit](https://github.com/spring-projects/spring-boot/commit/1213654c7473db2458132aec3295cbb898f8ce2e).

@snicoll Let me know if you think the name might still be a problem. We still have time to refine it.

philwebb:
> I'd hold this off for a while because we at the Spring Session side should probably take a closer look at our support for configuring ConfigureRedisAction.

Sorry @vpavic, I missed this comment until it was too late. Hopefully whatever changes you've got in mind won't be too hard to apply to the auto-configuration once they're ready. Is there a Spring Session issue we should be tracking?

vpavic:
No problem @philwebb. I've opened spring-projects/spring-session#1458 to consider doing something about this on Spring Session side.

Issue #16886, opened 2019-05-17
Make use of @DynamicPropertySource in our integration tests


wilkinsona:
There's a bit of ceremony involved at the moment:

```java
@Testcontainers
@ContextConfiguration(initializers = DataRedisTestIntegrationTests.Initializer.class)
@DataRedisTest
public class DataRedisTestIntegrationTests {

    @Container
    public static RedisContainer redis = new RedisContainer();

    // …

    static class Initializer
            implements ApplicationContextInitializer<ConfigurableApplicationContext> {

        @Override
        public void initialize(
                ConfigurableApplicationContext configurableApplicationContext) {
            TestPropertyValues.of("spring.redis.port=" + redis.getMappedPort())
                    .applyTo(configurableApplicationContext.getEnvironment());
        }

    }
```

It would be nice to be able to do something like this instead:

```java
@Testcontainers
@DataRedisTest
public class DataRedisTestIntegrationTests {

    @Container
    @MapPortToProperty("spring.redis.port")
    public static RedisContainer redis = new RedisContainer();
```

A few thoughts:

- The examples above are JUnit 5. We'd need to consider JUnit 4 as well
- The lifecycle ordering will need to be right so that the container's port is available before the context refreshes
- Setting the port will affect context caching

/cc @bsideup

bsideup:
@wilkinsona 
Take a look at https://github.com/Playtika/testcontainers-spring-boot

Also, there are a few things to consider covering:

- it is very important to also map the host, not just port. It is "localhost" only in ~60% cases
- some containers provide convenient methods like `KafkaContainer#getBootstrapServers`, perhaps we can use a SpEL expression or something

BTW I have a bit crazy idea:
What if Spring Boot implements something like:
```java
@Testcontainers
@DataRedisTest
public class DataRedisTestIntegrationTests {

    @Container
    public static RedisContainer redis = new RedisContainer();

    @PropertyContributor
    static Callable<PropertySource> propertySource = () -> {
        return TestPropertyValues.of(
            "spring.redis.host=" + redis.getContainerIpAddress(),
            "spring.redis.port=" + redis.getMappedPort()
        );
    }
```

This way it can source properties from anywhere, including Testcontainers. WDYT?

antkorwin:
Hi, I think I solved this in my library (here: https://github.com/jupiter-tools/spring-test-redis), maybe it can help you:

```
@SpringBootTest
@RedisTestContainer
class RedisTestContainerTest {

[...]

PR #17030, opened 2019-05-31
Add an annotation for dynamic test properties processing

antkorwin:
I made an annotation for integration tests, which allows setting a value of properties in the spring context before test execution, but this value can be not constant. 

Something like this discussed in the issue #16886 
 
DynamicTestProperrty allows declaring a method which will provide values of properties for a test. It provides an ability to use a not constant value as the value of test properties. 

For example, you can set a value of the "spring.redis.port" property after start the Redis TestContainer:

```java
@Testcontainers
@DataRedisTest
class DataRedisTestIntegrationTests {

    @Container
    private static RedisContainer redis = new RedisContainer();

    @DynamicTestProperty
    private static TestPropertyValues setContainerProperties() {
        return TestPropertyValues.of("spring.redis.port=" + redis.getMappedPort());
    }
    
    @Test
    void testRedis(){
      ...
    }
}
```

Also, this solution correctly works with the spring context cache. If a value of the property changes then it reflects on the key of the context in cache and spring-test will run a new application context for test.


wilkinsona:
Thanks for the PR, @antkorwin. Unfortunately, I don't think we're ready to implement something for this yet. There are a few different approaches that we could take and we need to do some design work before we'll be in a position to proceed. Sorry for the wasted effort. I've labelled #16886 as pending design work to make its status more clear.

Issue #13300, opened 2018-05-30
WebMvcMetricsFilter records a 500 response status for any NestedServletException irrespective of response's actual status


wilkinsona:
The upgrade to Micrometer 1.0 RC 6 included [this change](https://github.com/spring-projects/spring-boot/commit/ccc820f723416c20f3163509d30af9597f8b3d49#diff-d3cdbe75bfa3f7df310dcfc8c9685021R106) that's derived from [this change](https://github.com/micrometer-metrics/micrometer/commit/a6a667418ff4c5ca205d4c49ab2b3de4338cfae4) for https://github.com/micrometer-metrics/micrometer/pull/264. The intent was to avoid recording a 200 response for a request that's resulted in an exception being thrown and a 500 response being sent. However, the effect is that a 500 response is now recorded irrespective of how the exception is handled and the response status that is actually sent.

A concrete example of the problem is when Spring Security's method security throws an `AccessDeniedException`. `WebMvcMetricsFilter` catches the resulting `NestedServletException` and records a 500 response. Spring Security's `ExceptionTranslationFilter` then catches the exception, finds the `AccessDeniedException`, and, via `org.springframework.security.web.access.AccessDeniedHandlerImpl`, sets the response status to 403. This results in the client receiving a 403 but Micrometer recording at 500.

/cc @jkschneider

jkschneider:
cc / @adriancole

wilkinsona:
Thanks, @jkschneider. I recalled you mentioning that in the past when I was opening this issue. An alternative might be to change the filter's order so that it goes first on the way in and last on the way back out. Another would be to split the filter into two, with one that has highest precedence and one that has lowest and both talking to a shared component that does the actual metric recording. Can you refresh my memory on the reasoning for the current ordering? One side-effect that it has beyond this issue is that no metrics are recorded at all for requests that are rejected by Spring Security's filter.

jkschneider:
@wilkinsona You answered your own question here at the end:

> Can you refresh my memory on the reasoning for the current ordering? One side-effect that it has beyond this issue is that no metrics are recorded at all for requests that are rejected by Spring Security's filter.

It was precisely to be able to see requests rejected by Spring security that we made metrics a higher priority filter.

> An alternative might be to change the filter's order so that it goes first on the way in and last on the way back out.

If this is possible, let's do this. It's always possible that something server-specific outside of the servlet framework is changing a response code, but probably very rarely. This would get us to near perfection.

[...]

PR #17033, opened 2019-05-31
Set `HttpStatus.INTERNAL_SERVER_ERROR` status in case if response status has not been set before.

nosan:
see gh-13300

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcMetricsFilter.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/servlet/WebMvcMetricsFilter.java
@@ -25,6 +25,7 @@ import javax.servlet.FilterChain;
 import javax.servlet.ServletException;
 import javax.servlet.http.HttpServletRequest;
 import javax.servlet.http.HttpServletResponse;
+import javax.servlet.http.HttpServletResponseWrapper;
 
 import io.micrometer.core.annotation.Timed;
 import io.micrometer.core.instrument.MeterRegistry;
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR, @nosan, but I don't think this will address #13300. The problem there is that an exception is thrown and caught by the `WebMvcMetricsFilter`. At this point, no error has been sent and nothing has set the status. The filter then sets the status to 500 and rethrows the exception. The exception is then caught by Spring Security's filter which maps it to a 4xx status code and sets that on the response.

As far as I can tell, the change proposed here won't help with the problem described above as the status has not been set when `WebMvcMetricsFilter` is called. Have I misunderstood what your change does?

nosan:
I've checked that `MetricsWebFilter` has `@Order(Ordered.HIGHEST_PRECEDENCE + 1)` and from the other end:
```java
/**
	 * Default order of Spring Security's Filter in the servlet container (i.e. amongst
	 * other filters registered with the container). There is no connection between this
	 * and the {@code @Order} on a WebSecurityConfigurer.
	 */
	public static final int DEFAULT_FILTER_ORDER = OrderedFilter.REQUEST_WRAPPER_FILTER_MAX_ORDER
			- 100;

```
So a chance when `spring-security-filters` replaces `HTTP status` is not as big as `MetricsWebFilter` could replace the actual status, that has already been sent. 

From my perspective, this MR tries to fix an issue when `HTTP status` has been already set but `MetricsWebFilter` overrides it.
Looks like, I'm trying to fix something else that could happen when a custom `HTTP status` is set and an unhandled exception has been thrown.

wilkinsona:
If something sets the status and then throws an exception, the status will be overwritten. That will either happen because another filter catches the exception and sets the status or because the exception reaches the servlet container. In the latter case the container will set the status to 500. 

The container's behaviour is what that the filter currently tries to anticipate. It sets the status to 500 in anticipation of the container doing so if the exception reaches it. If nothing else catches the exception and changes the status, this means that the status recording in the metric for the request will match the status of the response to the request.

If we merge your proposed change, I think we'll increase the likelihood of the status in the metric not matching the status in the response. If something calls setStatus(201), for example, and then an exception is thrown which reaches the container, the change proposed here will result in the metric having a status of 201 but the response having a status of 500.

nosan:
@wilkinsona 
fair enough, sorry for the inconvenience :(

PR #17040, opened 2019-06-01
Replace anonymous class with lambda

wonwoo:


philwebb:
Thanks @wonwoo this is now in master along with a polish commit to fix some checkstyle issues.

PR #17044, opened 2019-06-02
Fix typo in javadoc

wonwoo:


snicoll:
@wonwoo good catch, thanks!

PR #17048, opened 2019-06-03
AssertJ assertions that can be applied to a BeanDefinition.

nosan:
This PR adds `BeanDefinitionAssert` which can be obtained via `ApplicationContextAssert` using 
`assertThat(context).getBeanDefinition("<name>")` .

Consider the example:
```java
	@Test
	void fooBeansShouldDependOnBar() {
		this.contextRunner
				.run(((context) -> {
					assertThat(context).getBeanDefinition("foo")
							.dependsOn("bar", "barFactoryBean");
				}));
	}
```
Prior to this commit:
```java
	@Test
	void fooBeansShouldDependOnBar() {
		this.contextRunner.run((context) -> {
			ConfigurableListableBeanFactory beanFactory = context.getBeanFactory();
			assertThat(getBeanDefinition("foo", beanFactory).getDependsOn())
					.contains("bar", "barFactoryBean");
		});
	}

	private BeanDefinition getBeanDefinition(String beanName,
			ConfigurableListableBeanFactory beanFactory) {
		try {
			return beanFactory.getBeanDefinition(beanName);
		}
		catch (NoSuchBeanDefinitionException ex) {
			BeanFactory parentBeanFactory = beanFactory.getParentBeanFactory();
			if (parentBeanFactory instanceof ConfigurableListableBeanFactory) {
				return getBeanDefinition(beanName,
						(ConfigurableListableBeanFactory) parentBeanFactory);
			}
			throw ex;
		}
	}
```
`BeanDefinitionAssert` contains a lot of assertions that can be applied to the `BeanDefinition` such as `isSingleton`, `isLazy`, `isPrimary`, etc.

Please let me know what you think.

P.S. I left **TODO** in `ApplicationContextAssert.findBeanDefinition` method because I'm not sure which approach is better.

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/assertj/ApplicationContextAssert.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/assertj/ApplicationContextAssert.java
@@ -32,6 +32,8 @@ import org.assertj.core.error.BasicErrorMessageFactory;
 
 import org.springframework.beans.factory.BeanFactoryUtils;
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.beans.factory.config.AutowireCapableBeanFactory;
+import org.springframework.beans.factory.config.BeanDefinition;
 import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
 import org.springframework.boot.test.context.runner.ApplicationContextRunner;
 import org.springframework.context.ApplicationContext;
@@ -47,6 +49,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @param <C> the application context type
  * @author Phillip Webb
  * @author Andy Wilkinson
+ * @author Dmytro Nosan
  * @since 2.0.0
  * @see ApplicationContextRunner
  * @see AssertableApplicationContext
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks for the PR @nosan. We've discussed this one today as a team today and whilst we really like the way this has been written (the code looks excellent), we're worried that it might encourage bad practices when it comes to testing. For example, we might set a flag like `lazy` on a bean definition to solve a circular dependency problem. Although we could test by simply checking that the lazy flag was set, it's generally better if we recreate the actual problem in the test.

I'm curious if this PR was driven by a need you found in your own projects? Or was it just something you thought might make a nice addition? For substantial pieces of work like this, feel free to open an issue first or a smaller PR that just sketches some code ideas. We're always happy to discuss things at an earlier stage if that helps.

nosan:
Thanks for the feedback @philwebb.

I had come up with this idea when I was writing this [test](https://github.com/spring-projects/spring-boot/pull/17020/files#diff-36fd17768a8bf2cacfa34bbe682663dd)

I have faced with a situation when it was quite difficult to get a bean definition from a context and check that particular bean definition `depends on` other beans, thus I have decided to create this PR with a thought that it can be useful not only for me but for someone else who has the same or similar issue.


> For example, we might set a flag like lazy on a bean definition to solve a circular dependency problem. Although we could test by simply checking that the lazy flag was set, it's generally better if we recreate the actual problem in the test.


From my perspective, `BeanDefinitionAssert` rather is complement than replacement.  I totally agree it is much better to recreate the actual problem in the test than check the `isLazy`, but sometimes check `bd.depends(...)` is enough.

philwebb:
All things considered, I think we'll leave things as they are. Working with bean definitions is somewhat unusual and for the reasons discussed in [the comment above](https://github.com/spring-projects/spring-boot/pull/17048#issuecomment-499161681) we'd rather not encourage users to go that route.

Thanks anyway for the pull request.

nosan:
In any case, thank you very much for your time!

PR #17049, opened 2019-06-03
Polish OutputCapture and its JUnit Jupiter extension

sbrannen:
This PR contains a number of individual commits to improve the output capture extensions for JUnit 4 and JUnit Jupiter.

Commit 490e31570c9f856e57c470b9018906f5567bc051 is not essential, but the changes will likely be necessary to support parallel test execution. The subsequent commit (8ce12377d337b0d1cfcf7d25beb5acf9fc5e44b0) introduces a failing, disable test case which demonstrates that parallel execution is currently not supported. If the team decides not to support parallel test execution, I recommend that a note be added to the Javadoc and Reference Manual to document this known shortcoming.

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/system/OutputCapture.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/system/OutputCapture.java
@@ -39,6 +39,7 @@ import org.springframework.util.ClassUtils;
  * @author Madhura Bhave
  * @author Phillip Webb
  * @author Andy Wilkinson
+ * @author Sam Brannen
  * @since 2.2.0
  * @see OutputCaptureExtension
  * @see OutputCaptureRule
@@ -128,8 +129,11 @@ class OutputCapture implements CapturedOutput {
 
 	private String get(Predicate<Type> filter) {
 		Assert.state(!this.systemCaptures.isEmpty(),
-				"No system captures found. Check that you have used @RegisterExtension "
-						+ "or @ExtendWith and the fields are not private");
+				"No system captures found. When using JUnit 4, ensure that you have "
+						+ "registered the OutputCaptureRule via @ClassRule or @Rule "
+						+ "and that the field is public. "
+						+ "When using JUnit Jupiter, ensure that you have registered "
+						+ "the OutputCaptureExtension via @ExtendWith.");
 		StringBuilder builder = new StringBuilder();
 		for (SystemCapture systemCapture : this.systemCaptures) {
 			systemCapture.append(builder, filter);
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks, @sbrannen. This is quite timely as I had just come to the realisation that storing the `CapturedOutput` instance might be useful for similar reasons to those discussed in the JUnit issue about making `@TempDir` work in an extension. I'm not sure it'll work, but the changes here will give me an easy way to try it out at least.

sbrannen:
Sure. No problem.

FWIW, I already did a spike by storing `capturedStrings` in `SystemCapture` in a `ThreadLocal`. That gets you pretty far but not all the way. The issue is that different threads write to `System.err` and `System.out`. Of course, we expect that to be the "challenge".

Perhaps you have an idea of how to solve that. If not, you might consider the [implementation in the JUnit Platform for inspiration](https://github.com/junit-team/junit5/blob/master/junit-platform-launcher/src/main/java/org/junit/platform/launcher/core/StreamInterceptor.java).

wilkinsona:
It might be possible to support output capture during parallel test execution using `InheritableThreadLocal` but I'm not sure that it will be worth the effort. Asserting captured output is already something of a last resort so I don't think it would be too much of a problem to only support it for serial test execution.

spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/system/OutputCapture.java
```diff
@@ -128,8 +129,11 @@ void reset() {
 
 	private String get(Predicate<Type> filter) {
 		Assert.state(!this.systemCaptures.isEmpty(),
-				"No system captures found. Check that you have used @RegisterExtension "
-						+ "or @ExtendWith and the fields are not private");
+				"No system captures found. When using JUnit 4, ensure that you have "
```
philwebb: We should probably just make this a general comment about making sure it's registered correctly. We don't really need to tie `OutputCapture` to JUnit at all now.

philwebb:
I'm not sure we should support parallel execution for now. We probably ought to just wait until we hit the need for it ourselves.

spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/system/OutputCapture.java
sbrannen: Sounds reasonable to me.

Do you want me to update the PR, or will you just modify it yourself when merging?

sbrannen:
> I'm not sure we should support parallel execution for now.

I can certainly understand that. In my opinion, it's only a "nice to have" feature.

So, if the community doesn't really request/need it, then postponing any work on it is justified.

I still think it would be worth it to document that fact in the Javadoc and Boot reference docs.

wilkinsona:
Thanks very much, @sbrannen.

Issue #16171, opened 2019-03-08
Printer<T> are not autoconfigured when exposed as a bean


tloist:
## Status Quo

[The `WebMvcAutoConfiguration`](https://github.com/spring-projects/spring-boot/blob/48e781181e497a4ae7cd49373b9b79eebe7ab50d/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcAutoConfiguration.java#L309) auto detects beans of type:

* `Converter`
* `GenericConverter`
* `Formatter`

and automatically registers them for the `ConversionService`. Formatters are in fact only a combination of `Printer` and `Parser`. For each `Formatter`, the formatter is split into these two parts and registered as two serperate conversion services again.
My suggestion is:

* Scanning for `Parser` and `Printer` seperately, which means the above list changes to:
  * `Converter`
  * `GenericConverter`
  * `Printer`
  * `Parser`

  If this introduces performance problems, an alternative would be to:

* Scan for `Printer` instead, then for each `Printer` check if the `Printer` is also a `Parser` (and thus a `Formatter`) and register the second conversion service as well. Which would make that list from above into:
  * `Converter`
  * `GenericConverter`
  * `Printer`
    * `Parser`

## Intended advantages

What this would make possible is to have a `Printer` without a `Parser` auto registered in the conversion service if exposed as a bean. This is useful for those cases, where you want to represent an object on the UI (e.g. templating in thymeleaf with `${{MyClass}}`) but never read it from a `String`.

If you tend to my 2nd proposal, this leaves the question: Why only checking for a `Parser` in case it is already a `Printer`?

2 reasons:
* Compability: Currently `Formatter` also register conversions `String` -> `MyClass` and not doing it, would break these applications relying on this behaviour.
* On the other hand I can't really think of a lot of use cases where registering a `Parser` without a `Printer` is realistic. If you know a String representation for `MyClass` that you can _read_, what should be the reason, you can't also _write_ it?

## Workaround

Currently `Printer` which are not also `Formatter` are not registered in the conversion service. Either:

* configure them manually / programmatically with an `WebMvcConfigurer` or
* just convert them to be a `Formatter` and throw an Exception for reading them from a String. That turns this case into a runtime exception.

philwebb:
Closing in favor of PR #17064

PR #17064, opened 2019-06-05
Registers Printer<T> and Parser<T> as converters

nosan:
Registers `Printer<T>` and `Parser<T>` as converters if they have been exposed as beans.

see gh-16171


philwebb:
I raise https://github.com/spring-projects/spring-framework/issues/23110 to see if we can't improve Spring Framework first. It's a bit nasty that we need new `PrintConverter` and `ParseConverter` public classes.

philwebb:
Thanks for the PR @nosan! This is now in master. With the https://github.com/spring-projects/spring-framework/issues/23110 change I've been able to simplify it as we no longer need our own `PrinterConverter` and `ParserConverter` classes.

nosan:
Thank you very much @philwebb

PR #17069, opened 2019-06-05
Add missing @Test annotations

izeye:
This PR adds missing `@Test` annotations in `SpringIterableConfigurationPropertySourceTests`.

This PR also restores assertion removed in 44d832158a60ba518565869e868052f72d0a972e which doesn't seem to be intentional.


philwebb:
Thanks yet again!

PR #17100, opened 2019-06-10
Add JUnit 4 and the Vintage Engine sample

### LINKED ISSUE #17070: Add a sample to verify the use of JUnit 4 and the Vintage Engine

ielatif:
This commit adds a sample to verify the use of JUnit 4 and the Vintage Engine.
Closes #17070.

### DIFF
--- a/spring-boot-samples/pom.xml
+++ b/spring-boot-samples/pom.xml
@@ -57,6 +57,7 @@
 		<module>spring-boot-sample-jta-bitronix</module>
 		<module>spring-boot-sample-jta-jndi</module>
 		<module>spring-boot-sample-junit-jupiter</module>
+		<module>spring-boot-sample-junit-vintage</module>
 		<module>spring-boot-sample-kafka</module>
 		<module>spring-boot-sample-liquibase</module>
 		<module>spring-boot-sample-logback</module>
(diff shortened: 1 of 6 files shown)


ielatif:
@philwebb @mbhave JUnit 5 checkstyle rules introduced in #17093 are applied to junit-vintage-sample which cause it to fail.
Any idea on how to disable only JUnit 5 checkstyle rules?

ielatif:
May be add a suppression for SampleJunitVintageApplicationTests in checkstyle-suppressions.xml?

wilkinsona:
@ielatif Yes, I think that's the right approach.

ielatif:
@wilkinsona raised this [PR](https://github.com/spring-projects/spring-boot/pull/17106) to fix CI build failure

snicoll:
@ielatif thanks for the PR. I've polished in two main areas:

1. I think what we want is to show that adding the starter with no further modification allows us to run a Junit 4-based test and the exclusions could potentially hide a problem. I've removed them
2. There is no need to start an embedded web server for such a sample so I rewrote the test to use mock MVC.

ielatif:
@snicoll 
> 1. I think what we want is to show that adding the starter with no further modification allows us to run a Junit 4-based test and the exclusions could potentially hide a problem. I've removed them

Totaly agree 👍 


> 2\. There is no need to start an embedded web server for such a sample so I rewrote the test to use mock MVC.

I was thinking of the sample as a sample application (application running an embedded web server) and not as a smoke test like @philwebb mentioned in a newly raised feature.

PR #17107, opened 2019-06-11
Migrate LoggingSystemTests to JUnit 5

### LINKED ISSUE #17097: Migrate JavaLoggingSystemTests and Log4J2LoggingSystemTests to JUnit 5
Currently ignoring from the checkstyle rule since they need to be migrated

L00kian:
Closes #17097


spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/system/LocaleExtension.java
```diff
@@ -0,0 +1,51 @@
...
+import org.junit.jupiter.api.extension.AfterEachCallback;
+import org.junit.jupiter.api.extension.BeforeEachCallback;
+import org.junit.jupiter.api.extension.ExtensionContext;
+
+/**
+ * Internal JUnit 5 {@code @Extension} to override system locale for test execution.
+ *
+ * @author Ilya Lukyanovich
+ */
+public class LocaleExtension implements BeforeEachCallback, AfterEachCallback {
```
wilkinsona: As this is only used in one place (`JavaLoggingSystemTests`), I'd prefer to manage it in `@BeforeEach` and `@AfterEach` methods in the class rather the pulling out an extension in a separate module.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/logging/AbstractLoggingSystemTests.java
```diff
@@ -16,42 +16,39 @@
...
-import org.junit.rules.TemporaryFolder;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.io.TempDir;
 
 import org.springframework.util.StringUtils;
 
 /**
- * Base for {@link LoggingSystem} tests.
+ * Base for {@link LoggingSystem} junit 5 tests.
```
wilkinsona: This line can be reverted now that the JUnit 4/5 split is gone.

spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/system/LocaleExtension.java
wilkinsona: This was marked as resolved but it doesn't seem to have been. Do you disagree about there being no need for the extension, @L00kian?

L00kian:
@wilkinsona Yep, sorry I was absolutely certain that I did delete the class and its usage. Done now

wilkinsona:
Thanks very much, @L00kian.

PR #17111, opened 2019-06-12
Make dependency available for javadoc generation

dreis2211:
Hi,

since 95d0e26c35a0e492233075ffa440541038e3617b there is imho a need to include derby as a dependency to have it available for javadoc generation. That should hopefully fix the current failures in JDK 11 and 12 builds.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #17129, opened 2019-06-13
Use properties not method parameter but injected field

dlsrb6342:
Like other beans declared in `WebEndpointAutoConfiguration`, I modified `PathMappedEndpoints` bean to also use injected `WebEndpointProperties`.


snicoll:
@dlsrb6342 good catch and thank you for making your first contribution to Spring Boot.

Issue #17197, opened 2019-06-14
Rename samples to smoke tests


philwebb:
The samples confuse people, renaming to make it clear they are simply smoke tests might help

ielatif:
It confused me also before dicovering [‘How-to’ guides](https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html). I think README file should state that those applications are smoke tests and for samples consult [‘How-to’ guides](https://docs.spring.io/spring-boot/docs/current/reference/html/howto.html)

rugden:
smoke tests is more confusing for me. what does smoke tests mean in this context?

philwebb:
@rugden You can find some more information about the term [here](http://softwaretestingfundamentals.com/smoke-testing/)

rugden:
That's what i thought it meant. I generally looked to sample projects in the last to see how to use certain spring boot features. 

Thanks for clarifying!

PR #17198, opened 2019-06-15
Improve spring-boot-sample-jersey1

guoyiang:
- Upgrade to jersey 1.19.1
- Added jersey-spring dependency
- Use SpringServlet which will allow jersey resource to be spring managed beans
  - In previous example, jersey resource is not spring managed and dependency injection is not supported
- Use Servlet registration instead of filter


philwebb:
Thanks for the contribution, unfortunately our sample are somewhat misnamed and they're really just a form of smoke test (see #17197). As such we don't really want to add complexity to them, even if it might help other users.

If you're interesting in working on a more complete Jersey example, then https://github.com/spring-guides/getting-started-guides/issues would be a good place to do so.

Thanks anyway.

PR #17199, opened 2019-06-15
Use constants for HTTP headers

izeye:
This PR changes to use constants for HTTP headers.


philwebb:
Thanks @izeye. The Cloud Foundry ones look good but I'm not sure that we can do `WebEndpointAutoConfiguration` since I think it can be used without `spring-web`.

snicoll:
Indeed I had to revert the change in `WebEndpointAutoConfiguration` as the endpoint infra should not require our web stack. See `WebEndpointResponse` for what we did for HTTP statuses.

Thanks again!

izeye:
@philwebb @snicoll Thanks for the feedback! I can see now that they don't depend on `spring-web` intentionally but failed to understand why they shouldn't depend on `spring-web` as they are all in `web` package. Could you please explain the reason?

wilkinsona:
It you are using Actuator on top of Jersey, `spring-web` will not be on the classpath.

izeye:
@wilkinsona Thanks for the explanation! Somehow I had a faulty assumption that `spring-web` will always be present based on [the `pom.xml` for `spring-boot-starter-jersey`](https://github.com/spring-projects/spring-boot/blob/7405efa326773a69bfc71a06a31d51a9781eb98e/spring-boot-project/spring-boot-starters/spring-boot-starter-jersey/pom.xml#L38-L41).

PR #17216, opened 2019-06-17
Custom Tcp Netty Server for Spring WebFlux

eutkin:
Custom Netty TcpServer for Spring WebFlux


snicoll:
@eutkin thank you for the PR and your efforts thus far. Could you please run the build locally rather than relying on CI and pushing one commit at every trial as this triggers a notification for everyone watching the repo.

Formatting rules are checked by default when the project is built.

eutkin:
@snicoll  Sorry

bclozel:
Actually the first commit looked nice already and I don't see the point of extracting `TcpServer` from the server factory. No need to change anything, I'll just pick up and polish (if needed) your first commit.

eutkin:
@bclozel n the old commit, We inject TcpServer into the NettyReactiveWebServerFactory. NettyReactiveWebServerFactory modifies the Tcp server host, port, loopResources. As a result, we get two different TcpServer, since TcpServer is immutable.

In the new commit, we first modify the TcpServer and only then inject it into the NettyReactiveWebServerFactory (with the exception of the port)

eutkin:
@snicoll @bclozel Can you drop xml to configure the codestyle for an idea? The plugin does not work for me?

snicoll:
@eutkin run the build on the command line please. You shouldn't be relying on local IDE setup to validate the formatting.

bclozel:
@eutkin then I don't get the use case here. I thought this was about applying customizations to the `TcpServer` instance?
Spring Boot is already providing the `NettyServerCustomizer` interface - creating a bean implementing it allows you to customize the HTTP and TCP server configuration after the Spring Boot configuration has been applied.

Wouldn't `NettyServerCustomizer` be enough for your needs?
Could you explain the use case in more details?

eutkin:
@bclozel We can customize NettyServer, yep. And how do we get customized tcpServer as a bean? TcpServer is needed if we want to use another protocol (for example, RSocket over Netty Tcp Server as Transport) and at the same time want to use web flux as an external api. You can raise two Tcp Server on different ports, one for Spring WebFlux, the other for RSocket, but I would like to unify. It will be embarrassing if I'm wrong :)

bclozel:
@eutkin we're already supporting those use cases in Spring Boot (see [reference documentation on RSocket](https://docs.spring.io/spring-boot/docs/2.2.0.BUILD-SNAPSHOT/reference/html/#boot-features-rsocket-server-auto-configuration)).

As far as I understand, you can't have both RSocket TCP and a Reactor Netty HTTP server on the same port. If there's an officially supported way of doing that, I'd be interested! Reactor Netty does not currently support multiple connectors for the same server (see reactor/reactor-netty#67).

In Spring Boot, we support:

* creating a standalone RSocket TCP server
* plugging RSocket WebSocket on a existing Reactor Netty HTTP server

Both can be achieved already with the infrastructure in place: see `RSocketServerFactory` and `RSocketWebSocketNettyRouteProvider` in our codebase.

eutkin:
I correctly understood that you can not use one TcpServer simultaneously for HttpServer and directly TcpServer?

Example:
```java
TcpServer tcpServer;

RSocketFactory.transport(TcpClientTransport.create(tcpServer));

HttpServer.from(tcpServer);
```

bclozel:
Again, as far as I understand, it's a way for reactor netty to bootstrap the server and the netty pipeline (see `HttpServerBind`). Do you have a sample somewhere showing RSocket TCP and an HTTP handler on the same reactor netty `HttpServer`, over the same port?

eutkin:
nope

bclozel:
I think we need to take a step back here and really understand the use case and what's missing in our infrastructure. I'll close this PR for now.

Feel free to experiment with RSocket and Reactor Netty and then create an issue here if needed.
Thanks!

eutkin:
Thanks

PR #17226, opened 2019-06-17
Add "OAuth2TokenValidator<Jwt>" to "JwtDecoder" auto configuration

ttddyy:
Currently, auto configuration for OAuth2 Resource Server (`OAuth2ResourceServerJwtConfiguration` and `ReactiveOAuth2ResourceServerJwkConfiguration`) are not considering jwt validator beans. Therefore, if user has custom validators, user needs to set them to decoder explicitly.

This PR discovers jwt validator beans and set to jwt decoder in auto configuration.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwtConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwtConfiguration.java
@@ -19,7 +19,10 @@ import java.security.KeyFactory;
 import java.security.interfaces.RSAPublicKey;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.Base64;
+import java.util.List;
+import java.util.stream.Collectors;
 
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwtConfiguration.java
```diff
@@ -79,8 +91,19 @@ public JwtDecoder jwtDecoderByPublicKeyValue() throws Exception {
 
 		@Bean
 		@Conditional(IssuerUriCondition.class)
-		public JwtDecoder jwtDecoderByIssuerUri() {
-			return JwtDecoders.fromOidcIssuerLocation(this.properties.getIssuerUri());
+		public JwtDecoder jwtDecoderByIssuerUri(ObjectProvider<OAuth2TokenValidator<Jwt>> jwtValidators) {
+			NimbusJwtDecoder jwtDecoder = (NimbusJwtDecoder) JwtDecoders
```
mbhave: I'm not sure about the unchecked class cast here. If something were to change in Spring Security where it no longer returned a `NimbusJwtDecoder`, this would break.

ttddyy: Yes, and same for reactive one as well.
It can add `instanceof` check.
What is the behavior if it does not match the `instanceof` check?
Is it ok to silently ignore, or should it fail?

mbhave: Failing would potentially be a breaking change if the `JwtDecoder` was no longer a `NimbusJwtDecoder`. Let's see what the rest of the team thinks.

mbhave:
@ttddyy We discussed this on the team call today. We think that it would be better to configure your own `JwtDecoder` bean with the custom validator instead of applying the validator to the auto-configured `JwtDecoder` bean as we can't guarantee what the type is. I'm going to close the PR in favor of doing that. Thanks anyway!

PR #17271, opened 2019-06-20
Fix 'duplicate invoke method named getActiveProfiles'

chenqimiao:
The method named getActiveProfiles was duplicate invoked.
Now, fix it through define a variable named  ``evnActiveProfiles``.


philwebb:
Good spot. That initialize method is a hangover from an old bit of code and I think we can just delete it now.

chenqimiao:
@philwebb I agree with you if the old code is not used now.

wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @chenqimiao. The proposed changes have been merged into master.

PR #17272, opened 2019-06-20
Use "JUnit" instead of "Junit"

izeye:
This PR changes to use "JUnit" instead of "Junit" for consistency.


snicoll:
good catch, thanks again @izeye!

PR #17274, opened 2019-06-20
Add @WebServiceClientTest annotation that can be used when testing `SOAP` clients.

nosan:
Add a slice `@WebServiceClientTest` annotation that can be used when testing `SOAP` clients.

The implementation is quite similar to `@RestClientTest` but for `WebServiceTemplateBuilder`. 
PR is not ready yet, it is just a concept of how it could be.

Here is a small example of usage:

```java
@WebServiceClientTest(ExampleWebServiceClient.class)
class WebServiceClientIntegrationTests {

	@Autowired
	private MockWebServiceServers servers;

	@Autowired
	private ExampleWebServiceClient client;

	@Test
	void mockServerCall() {
		this.servers.expect(RequestMatchers.payload(new StringSource("<request/>"))).andRespond(
				ResponseCreators.withPayload(new StringSource("<response><status>200</status></response>")));
		assertThat(this.client.test()).extracting(Response::getStatus).isEqualTo(200);
	}

	@Test
	void mockServerCall1() {
		this.servers.expect(RequestMatchers.connectionTo("https://example1"))
				.andRespond(ResponseCreators.withPayload(new StringSource("<response/>")));
		assertThatExceptionOfType(SourceAssertionError.class).isThrownBy(this.client::test)
				.withMessageContaining("Unexpected connection expected");
	}
}
```
If you are interested in this feature I will add more tests and documentation.

Also, pay attention to https://github.com/spring-projects/spring-boot/compare/master...nosan:webservice-client-test?expand=1#diff-a0ae2b770ccb1b5cc0b45d9966df843aR66 it would be nice to have `reset()` functionality in `MockWebServiceServer` class. Is it possible to add?

Let me know what you think and thanks in advance.

P.S. **Work-in-Progress**

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1689,7 +1689,7 @@ bom {
 			]
 		}
 	}
-	library("Spring WS", "3.0.8.RELEASE") {
+	library("Spring WS", "3.0.9.BUILD-SNAPSHOT") {
 		group("org.springframework.ws") {
 			modules = [
 				"spring-ws-core",
--- a/spring-boot-project/spring-boot-test-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-test-autoconfigure/build.gradle
@@ -45,6 +45,8 @@ dependencies {
 	optional("org.springframework.restdocs:spring-restdocs-webtestclient")
 	optional("org.springframework.security:spring-security-config")
 	optional("org.springframework.security:spring-security-test")
+	optional("org.springframework.ws:spring-ws-core")
+	optional("org.springframework.ws:spring-ws-test")
 	optional("org.apache.tomcat.embed:tomcat-embed-core")
 	optional("org.mongodb:mongodb-driver-reactivestreams")
 	optional("org.mongodb:mongodb-driver-sync")
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/webservices/client/AutoConfigureMockWebServiceServer.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/webservices/client/AutoConfigureMockWebServiceServer.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.test.autoconfigure.webservices.client;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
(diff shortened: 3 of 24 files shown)


philwebb:
@nosan This looks like a really nice addition to the existing test slice support and would be great to add it in 2.2.x. The one things we're not totally sure about is the `MockWebServiceServers` class. I think we'd rather not introduce that in Boot and instead just support a single `MockWebServiceServer` (similar to the way we support `MockRestServiceServer`).

> it would be nice to have `reset()` functionality in `MockWebServiceServer` class. Is it possible to add?

@gregturn Can probably help answer that. I'd certainly like to keep a clean separation and have the general use Mock code in spring-ws and keep the Boot code specific to our slice concepts.

nosan:
Thank you @philwebb,
I have added `MockWebServiceServers` because I have not found a way to customize `MockWebServiceServer` with a custom `MockWebServiceMessageSender` to create a *deferred* `MockWebServiceServer` since `MockWebServiceMessageSender` has a **package private** modifier.

If adding support for a slice `@WebServiceClientTest` annotation is desired, then the possibility to reset `MockWebServiceServer` after each test should be added. Also, a  **public** modifier to `MockWebServiceMessageSender` class should be added as well.

nosan:
@philwebb I've updated PR  https://github.com/spring-projects/spring-boot/pull/17274/commits/fc53f37bb68bfed45a41ff9b78f9ac20cb7409a7 to support only a single `WebServiceTemplate`. 

This approach is quite messy with `Reflection` for me. I would rather not to add `@WebServiceClientTest` annotation at all than using this way.

I am hoping that `spring-ws` team could help here.

philwebb:
@nosan I totally agree, we'll really need some upstream changes I think.

wilkinsona:
@gregturn can you please review the discussion above? As currently proposed, there's quite a bit of reflection when interacting with Spring Web Services which we'd like to avoid.

gregturn:
I don't see why we can't alter the visibility of `MockWebServiceServer(MockWebServiceMessageSender)`. Looks like a valid use case to open it up for public usage. 

I opened https://jira.spring.io/browse/SWS-1085.

gregturn:
You can test out these changes using `2.4.7.BUILD-SNAPSHOT` or `3.0.9.BUILD-SNAPSHOT` (until a release is made).

gregturn:
If @nosan wishes to retool the PR using these modifications, so be it.

nosan:
Thanks for your feedback and changes, @gregturn

I checked `3.0.9.BUILD-SNAPSHOT` and unfortunately, those changes are not enough to avoid reflection at all. 

Could you please add the following changes?

- `MockWebServiceMessageSender` must be **public** as well.
-  Add _reset_ method to `MockWebServiceServer`, similar to _verify_ but for reset.

_'reset'_  just clears`expectedConnections` and sets `connectionIterator` to null.


Ideally,  this [class](https://github.com/spring-projects/spring-boot/pull/17274/files#diff-d269060f3f73330e5898aad3addbc304) should be removed.

 
Thanks in advance

gregturn:
I can extend the surface API to support a `rest`, but are you sure you can't just create a new instance for each test case?

wilkinsona:
@gregturn In the proposed changes, `MockWebServiceServer` is being [used as a bean](https://github.com/spring-projects/spring-boot/pull/17274/files#diff-fd38907a0519762d7b733a389905c34cR38-R41). Requiring a new instance per test would mean that we needed a new application context per test, making the test framework's context cache redundant and slowing down overall test execution.

gregturn:
@nosan Okay, the additional changes you requested have been applied. Again, check the latest snapshots and see if they are suitable.

nosan:
Thanks @gregturn


PR has been updated

gregturn:
`3.0.9.RELEASE` has been released.

wilkinsona:
Re-opening to address the CI failures on JDK 11 and 14.

nosan:
Thanks a lot! 👍

PR #17277, opened 2019-06-20
Make layout parameter available as a user property

marcrohlfs:
We need to change the packaging layout in some special cases (in CI builds) and don't want to clutter or production code adding overwritable properties to our POMs - furthermore we'd have to define a default that would disable the standard behavior of layout auto-detection.

With this PR the _layout_ parameter of the _repackage_ goal can be set from the
command line, using the system property `spring-boot.repackage.layout`.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/it/repackage-layout-default/invoker.properties
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/it/repackage-layout-default/invoker.properties
@@ -1,0 +1,1 @@
+invoker.goals=package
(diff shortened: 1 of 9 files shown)


snicoll:
> ) and don't want to clutter or production code adding overwritable properties to our POMs - furthermore we'd have to define a default that would disable the standard behavior of layout auto-detection.

@marcrohlfs regardless of the proposal here I'd like to understand what you mean by that. If you want to change the layout you can define a different plugin configuration in a profile and enable that profile on the command line with `-P`. What is wrong with that?

marcrohlfs:
@snicoll Normally You're right, but in our setup (with multiple applications not "attached" to the same Maven parent), it's barely possible to define (a profile with) a common plugin management. Furthermore we're solving an internal case on public code/setup with this, hence we rather don't want to have this extra configuration in the public POM.

snicoll:
Fair enough, thanks for the feedback @marcrohlfs

snicoll:
@marcrohlfs thank you for making your first contribution to Spring Boot. This is now merged in `master`.

marcrohlfs:
@snicoll You're welcome. And many thanks fo accepting and merging 👍

Issue #17283, opened 2019-06-21
HibernateJpaAutoConfiguration's condition uses deprecated classes


snicoll:
`HibernateEntityManagerCondition` checks for deprecated classes, we should review the condition and update it accordingly.

nosan:
@snicoll May I take this one?

snicoll:
I forgot to backport it, creating a separate issue for the `2.1.x` line.

PR #17285, opened 2019-06-21
Update HibernateJpaAutoConfiguration's condition to use non deprecated class

nosan:
see gh-17283


snicoll:
Thanks for the PR @nosan but I've decided to fix this differently, see the original issue for more details.

Issue #15960, opened 2019-02-14
Provide a way to customize RedisCacheManager


jntakpe:
Spring Boot uses CacheManagerCustomizer to customize cache managers. RedisCacheConfiguration calls a list of CacheManagerCustomizers but most of RedisCacheManager methods are private.
The only cache configuration access is the method RedisCacheManager#getCacheConfigurations() but it returns an UnmodifiableMap.

Maybe I'm missing something but the only way I found to customize RedisCacheConfiguration TTL per cache is to declared my own RedisCacheManager.

snicoll:
@jntakpe yes and that's pretty much the case for most cache implementations that do not rely on an external configuration file. If you want to start tuning things per cache, you'd better off adding the 3 lines of code that help you get started with the builder and use that. If we offered that flexibility then we'd be at odds with `spring.cache.cache-names` as they also allow you to setup caches on startup.

jntakpe:
@snicoll That not that simple to mix the default Spring cache configuration which works fine for most of our use cases and a tuning per cache. At least it's not "adding 3 lines of code" but more like 3 classes with a lot of copied code from Spring Boot configuration.
Anyway, I get your point but I still think allowing access to RedisCacheManager#initialCacheConfiguration or give an equivalent access to the old [RedisCacheManager#setExpires(Map<String,Long>)](https://docs.spring.io/spring-data/redis/docs/1.8.6.RELEASE/api/org/springframework/data/redis/cache/RedisCacheManager.html) should be considered and would be an improvement

snicoll:
Fair enough, I guess we can consider opening up things a bit now that `RedisCacheManager` is immutable.

jntakpe:
Great news ! Thank you for reconsidering

snicoll:
I've created [DATAREDIS-935](https://jira.spring.io/browse/DATAREDIS-935)

snicoll:
The plan here is to provide a `RedisCacheManagerBuilderCustomizer` of some sort since the `CacheManager` is immutable.

philwebb:
Closing in favor of PR #17287

uqix:
@snicoll `RedisCacheManagerBuilderCustomizer` usage could be mentioned in `reference` doc

PR #17287, opened 2019-06-21
Introduce a customizer callback for RedisCacheManagerBuilder

nosan:
see gh-15960


wilkinsona:
Thanks again, @nosan. The proposed changes are now in master.

PR #17289, opened 2019-06-21
Polish WebServiceTemplateBuilder

nosan:
This PR fixes `javadoc` and polishes trivial stuff.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
@@ -473,8 +473,8 @@ public class WebServiceTemplateBuilder {
 	private void applyCustomizers(WebServiceTemplate webServiceTemplate,
 			Set<WebServiceTemplateCustomizer> customizers) {
 		if (!CollectionUtils.isEmpty(customizers)) {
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/webservices/client/WebServiceTemplateBuilder.java
```diff
@@ -520,27 +520,28 @@ private WebServiceMessageSenders(boolean onlyAdditional, Set<WebServiceMessageSe
 			this.messageSenders = messageSenders;
 		}
 
-		public boolean isOnlyAdditional() {
+		boolean isOnlyAdditional() {
```
philwebb: I don't think we've yet defined a good code style for these kinds of methods. Obviously there's nothing wrong with dropping `public` but I don't think we're very consistent across the codebase. We should probably come up with a standard and start to enforce it with checkstyle.

philwebb: We discussed this today and like the style that you've adopted. Going forward, we'll not have unnecessary public methods.

nosan: Thanks

wilkinsona:
Thanks very much, @nosan.

Issue #15303, opened 2018-11-27
How can anonymous do a post when configured with springsecurity


chunzhenzyd:
this is my configure:

@Configuration
@EnableWebSecurity
public class MyWebSecurityConfigurer extends WebSecurityConfigurerAdapter {

	private static final Logger logger = LoggerFactory.getLogger(MyWebSecurityConfigurer.class);

	@Autowired
	private MySecurityInterceptor mySecurityInterceptor;

	@Autowired
	private MyAuthenticationSuccessHandler myAuthenticationSuccessHandler;
	@Autowired
	private MyAuthenticationFailureHandler myAuthenticationFailureHandler;

	@Bean
	UserDetailsService customUserService() { // 注册UserDetailsService 的bean
		logger.trace("customUserService()");
		return new MyUserDetailsService();
	}

	@Override
	protected void configure(AuthenticationManagerBuilder auth) throws Exception {
		logger.trace("configure(AuthenticationManagerBuilder)");
		// user Details Service验证
		auth.userDetailsService(customUserService()).passwordEncoder(new BCryptPasswordEncoder());
	}

	@Override
	protected void configure(HttpSecurity http) throws Exception {
		logger.trace("configure(HttpSecurity)");
		http
			.csrf()
				.csrfTokenRepository(new HttpSessionCsrfTokenRepository())
//				.csrfTokenRepository(new CookieCsrfTokenRepository())
//				.csrfTokenRepository(CookieCsrfTokenRepository.withHttpOnlyFalse())
				.ignoringAntMatchers("/login")
			.and()
			.addFilterBefore(mySecurityInterceptor, FilterSecurityInterceptor.class)
			.authorizeRequests()
				.anyRequest().authenticated()
			.and()
			.formLogin()
				.loginPage("/login")
				.successHandler(myAuthenticationSuccessHandler).failureHandler(myAuthenticationFailureHandler)
				.permitAll()
			.and()
			.logout()
				.permitAll();
	}

	/*** 设置不拦截规则 */
	@Override
	public void configure(WebSecurity web) throws Exception {
		web.ignoring().antMatchers("/actuator/**", "/css/**", "terminal/stb/**");
		web.ignoring().antMatchers(HttpMethod.GET);
	}

}

but when I do a post http://localhost:8080/terminal/stb/test, a 302 code returned.

PR #17291, opened 2019-06-21
Introduce dedicated interface to listen for ClassPathChangedEvents.

odrotbohm:
So far, `LocalDevToolsAutoConfiguration` had registered a Lambda-based event listener for `ClassPathChangedEvents`. As the generics information is not available for lookup from a Lambda defined generic interface, `SimpleApplicationEventMulticaster.doInvokeListener(…)` has no chance but to submit *all* events published to the listener, catching the `ClassCastExceptions` resulting from that to log them into the `TRACE` level.

For applications that produce a lot of events (esp. during startup) like ones that use Spring Data MongoDB which uses the events to trigger index creation etc., this means that a lot of exceptions are thrown and ultimately discarded most of the time.

This commit introduces a dedicated interface for that particular application listener so that it can be discarded for all events but the one it's really interested in, thus avoiding the superfluous exceptions being produced and ultimately leading to better (startup) performance.

This was tested using Spring Data Examples (rest/starbucks) and lead to a startup time improvement of ~500ms. The example uses Spring Data MongoDB whose object mapping is heavily making use of application events when reading and writing objects to the database. I assume the effect will be less visible in applications using application events less intensively.

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LocalDevToolsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LocalDevToolsAutoConfiguration.java
@@ -100,7 +100,7 @@ public class LocalDevToolsAutoConfiguration {
 		}
 
 		@Bean
-		public ApplicationListener<ClassPathChangedEvent> restartingClassPathChangedEventListener(
+		public ClassPathChangedEventListener restartingClassPathChangedEventListener(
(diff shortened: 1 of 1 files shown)


philwebb:
It's a shame we need to do this because the generics are available in the bytecode, it's just that they're on the factory method and not the bean instance.

philwebb:
I've raise https://github.com/spring-projects/spring-framework/issues/23178 to see if there's not a more general solution that Framework can apply. If they can't we'll do the change in Boot.

odrotbohm:
#15303 looks like a duplicate of this.

wilkinsona:
Looks like Framework has taken care of this for us.

PR #17322, opened 2019-06-26
Auto-configure Kafka listener container with user-provided RecordInterceptor

garyrussell:
spring-kafka.2.3.0.M3 introduced a new property `RecordInterceptor` to
the container factory.

Auto-configure the property if a single instance is present.


wilkinsona:
Thank you, @garyrussell.

PR #17326, opened 2019-06-26
Let parallel CI steps fail fast

dreis2211:
Hi,

I saw that you updated Concourse to 5.3 and that `in_parallel` is used now. This should give us the possibility to use the `fail_fast` option. Hopefully this will boost build times when parallel steps fail.

Let me know what you think. I didn't see a way of testing this other than providing the PR and seeing if it works there.

Cheers,
Christoph

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -150,6 +150,8 @@ jobs:
     trigger: true
   - get: spring-boot-ci-image
   - in_parallel:
+    fail_fast: true
+    steps:
     - task: detect-jdk8-update
       file: git-repo/ci/tasks/detect-jdk-updates.yml
       params:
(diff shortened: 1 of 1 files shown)


dreis2211:
I'm not exactly sure if the build is not triggered because of my changes or it simply takes longer than usual.

philwebb:
The CI build won't automatically apply pipeline chances, we'll need to do that manually.

dreis2211:
That's good to know. Thanks for the clarification :)

mbhave:
Thanks for the PR @dreis2211. I'm not sure if we necessarily want to fail fast though. It is helpful sometimes to get all failures so that they can be fixed together. Let's see what the team thinks.

dreis2211:
@mbhave As most steps fail fast themselves there's imho no way of actually fixing everything together that might be broken. E.g. you would need `--fail-at-end` for the respective Maven steps for that. Thanks for bringing up your concerns, though.  Happy to read what the team thinks.

philwebb:
Thanks for the suggestion, but we think it's probably more useful to have the complete set of results rather than failing fast.

dreis2211:
Too bad - I was hoping you decide otherwise. But fair enough :)

PR #17355, opened 2019-06-28
Use @EmbeddedKafka for Integration Tests

garyrussell:
`@EmbeddedKafka` now supports JUnit5 without the need for a
Spring test application context.

- remove the previous work-around that replaced the JUnit4 `@ClassRule`
- also use `TopicBuilder` to build the topic bean


wilkinsona:

Issue #6715, opened 2016-08-22
Allow Undertow's server and socket options to be configured via the environment


wilkinsona:
Something like:

```
server.undertow.options.socket.whatever: foo
server.undertow.options.server.whatever: bar
```

`whatever` is then coerced into one of the constants on `UndertowOptions`, ideally with relaxed binding. A concrete example:

```
server.undertow.options.server.enable-http2: true
```

See #3350 which this issue probably supersedes.

snicoll:
`Enum` has already some kind of relaxed bindings and you'll need to make a `Map<UndertowOptions, Object>` to make that work.

wilkinsona:
If only `UndertowOptions` were an `Enum`

ofbizbrazil:
I really missed this property for a long while. Please add this ASAP!!!

ofbizbrazil:
I really missed this property for a long while. Please add this ASAP!!!

mbhave:
Closing in favor of PR #17356.

PR #17356, opened 2019-06-28
Allow Undertow's server and socket options to be configured via the environment

htztomic:
This is a fix to GitHub issue #6715. Additional mapping of options and functionalities were added for configuring Undertow's server and socket options.


mbhave:
@htztomic I think you would need to apply the options to `UndertowReactiveWebServerFactory` as well.

htztomic:
@mbhave The `UndertowWebServerFactoryCustomizer` is compatible with both reactive and servlet.

PR #17357, opened 2019-06-29
Add a missing quote in DebugLogbackConfigurator

izeye:
This PR adds a missing quote in `DebugLogbackConfigurator.conversionRule()`.


wilkinsona:
Thanks once again, @izeye. I have no idea how you keep spotting these, but we are all very glad that you do.

Issue #17365, opened 2019-06-29
Springboot activemq supports SSL connections by default？


sh930108:
Springboot has no place to configure SSL related parameters and certificates for Activemq，Springboot activemq supports SSL connections by default？

wilkinsona:
There's no property based support for configuring a connection factory that uses SSL with a custom keystore or custom truststore. You can, however, provide your own `ActiveMQSslConnectionFactory` bean and keep the rest of the auto-configuration.

philwebb:
I've opened #17589 for us to consider more broadly how we should support SSL.

SaadiMelek:
We need too override the default connection using ActiveMQSslConnectionFactory bean with your own configuration properties :
`@Bean
public ActiveMQSslConnectionFactory activeMQSslConnectionFactory(
        @Value("${spring.activemq.broker-url}") String brokerUrl,
        @Value("${spring.activemq.ssl.trustStorePath}") String trustStorePath,
        @Value("${spring.activemq.ssl.trustStorePass}") String trustStorePass,
        @Value("${spring.activemq.ssl.keyStorePath}") String keyStorePath,
        @Value("${spring.activemq.ssl.keyStorePass}") String keyStorePass) throws Exception {
    ActiveMQSslConnectionFactory factory = new ActiveMQSslConnectionFactory(brokerUrl);
    factory.setTrustStore(trustStorePath);
    factory.setTrustStorePassword(trustStorePass);
    factory.setKeyStore(keyStorePath);
    factory.setKeyStorePassword(keyStorePass);
    return factory;
}`

scottfrederick:
Blocked by #34011

PR #17373, opened 2019-07-01
Provide configuration properties for configuring an ActiveMQ connection factory that uses SSL

nosan:
see gh-17365


philwebb:
Thanks for the PR @nosan. 

Our SSL support is becoming a bit disjointed and I think we need to take a step back and consider how we can rationalize it. I've raise #17589 for us to look more broadly at our options.

I'm afraid we've decided not to merge this one right now because we don't want to add any more complexity to the mix. Thanks anyway!

Issue #9114, opened 2017-05-05
Provide a `@BatchDataSource` qualifier to allow a Batch-specific DataSource to be identified


Ramblurr:
**Env**: Spring boot 1.5, Spring Batch 4.0.0.M1

In the `org.springframework.boot.autoconfigure.batch.BatchAutoConfiguration` there is no way to specify which DataSource to use.

Compare this to `org.springframework.cloud.task.configuration.SimpleTaskConfiguration`, which errors if multiple DataSources are detected, and provides a way to set the DataSource.

In non-trivial projects there can often be more than 1 DataSource. To specify a particular one for the batch auto config, you have to supply 3 beans:

* `BatchDatabaseInitializer`
* `JobExplorer`
* `basicBatchConfigurer` / `jpaBatchConfigurer`

Copy and pasting those into a different config class just to inject the right DataSource is quite painful.

It would be nice if there was a mechanism for providing the datasource.

snicoll:
Would a `@BatchDataSource` qualifier work for you? (similar to what we've done for Flyway and Liquibase)

philwebb:
Closing in favor of PR #17375

PR #17375, opened 2019-07-01
Provide a `@BatchDataSource` qualifier to allow a Batch-specific DataSource to be identified

nosan:
see gh-9114


wilkinsona:
Thanks once again, @nosan. Your addition of `@ConditionalOnBean` made me aware of [a bug](https://github.com/spring-projects/spring-boot/issues/17449). We'll tackle that separately so that it can be fixed in 2.1.x.

nosan:
Thank you very much, @wilkinsona

PR #17389, opened 2019-07-01
add isolation level to KafkaProperties

sparty02:
Add the "isolation.level" property as a well defined config.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
@@ -267,6 +267,13 @@ public class KafkaProperties {
 		private Duration heartbeatInterval;
 
 		/**
+		 * Controls how transactional messages are returned when polling the broker
+		 * (non-transactional messages will be unconditionally returned, regardless of
+		 * this setting).
+		 */
+		private String isolationLevel;
+
+		/**
 		 * Deserializer class for keys.
 		 */
 		private Class<?> keyDeserializer = StringDeserializer.class;
@@ -362,6 +369,14 @@ public class KafkaProperties {
 			this.heartbeatInterval = heartbeatInterval;
 		}
 
+		public String getIsolationLevel() {
+			return this.isolationLevel;
+		}
+
+		public void setIsolationLevel(String isolationLevel) {
+			this.isolationLevel = isolationLevel;
+		}
+
 		public Class<?> getKeyDeserializer() {
 			return this.keyDeserializer;
 		}
@@ -406,6 +421,7 @@ public class KafkaProperties {
 			map.from(this::getGroupId).to(properties.in(ConsumerConfig.GROUP_ID_CONFIG));
 			map.from(this::getHeartbeatInterval).asInt(Duration::toMillis)
 					.to(properties.in(ConsumerConfig.HEARTBEAT_INTERVAL_MS_CONFIG));
+			map.from(this::getIsolationLevel).to(properties.in(ConsumerConfig.ISOLATION_LEVEL_CONFIG));
 			map.from(this::getKeyDeserializer).to(properties.in(ConsumerConfig.KEY_DESERIALIZER_CLASS_CONFIG));
 			map.from(this::getValueDeserializer).to(properties.in(ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG));
 			map.from(this::getMaxPollRecords).to(properties.in(ConsumerConfig.MAX_POLL_RECORDS_CONFIG));
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -2229,6 +2229,24 @@
       ]
     },
     {
+      "name": "spring.kafka.consumer.isolation-level",
+      "values": [
+        {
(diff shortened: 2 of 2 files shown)


wilkinsona:
@garyrussell @artembilan Does making the isolation level configurable seem reasonable to you?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -266,6 +266,13 @@ public Template getTemplate() {
 		 */
 		private Duration heartbeatInterval;
 
+		/**
+		 * Controls how transactional messages are returned when polling the broker
+		 * (non-transactional messages will be unconditionally returned, regardless of
+		 * this setting).
+		 */
+		private String isolationLevel;
```
artembilan: It may make sense if it is mapped as an `org.apache.kafka.common.requests.IsolationLevel` enum type.
Otherwise this one indeed can be mapped as all other generic properties in the `spring.kafka.consumer.properties` container.

More over I think such a property description is too long. Essentially there is just enough to mention that it is mapped onto `ConsumerConfig.ISOLATION_LEVEL_CONFIG)`.

garyrussell:
As discussed [here](https://docs.spring.io/spring-boot/docs/2.1.6.RELEASE/reference/html/boot-features-messaging.html#boot-features-kafka-extra-props).

>Spring Boot auto-configuration supports all HIGH importance properties, some selected MEDIUM and LOW properties, and any properties that do not have a default value.

>Only a subset of the properties supported by Kafka are available directly through the KafkaProperties class. If you wish to configure the producer or consumer with additional properties that are not directly supported, use the following properties:

>spring.kafka.properties.prop.one=first
spring.kafka.admin.properties.prop.two=second
spring.kafka.consumer.properties.prop.three=third
spring.kafka.producer.properties.prop.four=fourth
spring.kafka.streams.properties.prop.five=fifth

So `spring.kafka.consumer.properties.isolation.level=read_committed` is all you need.

OTOH, it is nice to have content assist in IDEs so I have no objection (subject to Artem's suggestion of using the enum so that you get content-assist on valid values).

However, as we discussed when we added the initial support in boot, there are so many kafka properties, we can't make them all first class citizens.

wilkinsona:
Thank you, @garyrussell and @artembilan. Having had my memory refreshed about the myriad properties with which Kafka can be configured, I'm tempted to leave the line drawn where it is and decline this one. Let's see what the rest of the team thinks before making a decision.

sparty02:
To clarify, I'm aware of the ability to set any arbitrary properties through spring.kafka.consumer.* and friends.  But as mentioned in the [comment above](https://github.com/spring-projects/spring-boot/pull/17389#issuecomment-507660088), it's useful to have autocomplete in IDEs to nudge you in the right direction.  In an ideal world, the entire set of [consumer configs](https://kafka.apache.org/documentation/#consumerconfigs) (and others) would be available, but it seems that would only work if there was a process to generate that code/etc.  That being said, the current subset seems arbitrary and I'm not sure what the rationale would be for freezing it now vs letting it incrementally grow over time (until, if ever, a process was figured out to codegen based off the official config api).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
sparty02: Coincedentally, I had the same two thoughts!  On both accounts, I went the direction I did because of the precedent set by existing config options.  I found valid values being defined in metadata instead of enums in [this example](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L2184) and I added the description in the comment again based on all of the existing config.

Interestingly, the `isolation.level` property does have its [doc constant](https://github.com/apache/kafka/blob/2.3.0/clients/src/main/java/org/apache/kafka/clients/consumer/ConsumerConfig.java#L270) exposed as `public` even though most other config options are `private`

garyrussell: AFAIK, there is no enum for `ConsumerConfig.AUTO_OFFSET_RESET_CONFIG` (certainly not used there) which is why it's handled that way.

The isolation level enum is used in `ConsumerConfig` so it's best to use that.

garyrussell: Actually, they have to lower case the enum anyway, so maybe this is the best solution after all.

`IsolationLevel.READ_UNCOMMITTED.toString().toLowerCase(Locale.ROOT);`

artembilan: Still would be better to be based on enum. That way, when they introduce a new level, we are going to be covered automatically together with content assistant in the IDE.

wilkinsona:
We've discussed this today and come to the conclusion that we don't like the current halfway-house situation and would like to support all of Kafka's properties. That's a large piece of work that'll happen post-2.2 (if at all). In the meantime, given that planned direction, we're happy to add a property for the isolation level.

wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @sparty02. The changes have been merged into master with a [polish commit](https://github.com/spring-projects/spring-boot/commit/b39479bc07dd9359f453d9623a665b5c177ef15d) that makes use of the `IsolationLevel` enum, removes the now redundant additional metadata, and tests the new property.

PR #17391, opened 2019-07-01
Add extra validators to ConfigurationProperties

JurrianFahner:
Under spring-boot 2.1.x there are limited options to add validators to properties:

- implement the Validator interface
- add a `configurationPropertiesValidator` bean
- apply jsr303

For the use-case that somebody wants to have more validators on one configuration properties class, see also [an example implementation](https://github.com/JurrianFahner/demo-configuration-properties/blob/master/src/main/java/nl/ensignprojects/configurationpropertiesvalidation/AppConfigurationProperties.java) (see for even more info also the [readme](https://github.com/JurrianFahner/demo-configuration-properties/blob/master/Readme.md)), that is not so easy implemented now.

In the way how it is now set up, it is also not easy to reuse the code for validators. For example if somebody wants the validation logic in a library.   

The whole idea of this enhancement request is that it is trivial to add more than one validator. Also code reuse for the validation can be done more easily, especially if an interface is implemented on the properties class.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/ConfigurationPropertiesBinder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/ConfigurationPropertiesBinder.java
@@ -21,6 +21,7 @@ import java.util.List;
 import java.util.function.Consumer;
 import java.util.stream.Collectors;
 
+import org.springframework.beans.BeanUtils;
 import org.springframework.beans.PropertyEditorRegistry;
 import org.springframework.boot.context.properties.bind.BindHandler;
 import org.springframework.boot.context.properties.bind.Bindable;
@@ -47,6 +48,7 @@ import org.springframework.validation.annotation.Validated;
  *
  * @author Stephane Nicoll
  * @author Phillip Webb
+ * @author Jurrian Fahner
  */
 class ConfigurationPropertiesBinder {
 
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/ConfigurationPropertiesBinder.java
```diff
@@ -97,6 +98,14 @@ private Validator getConfigurationPropertiesValidator(ApplicationContext applica
 		if (target.getValue() != null && target.getValue().get() instanceof Validator) {
 			validators.add((Validator) target.getValue().get());
 		}
+		for (Class<? extends Validator> validator : annotation.validators()) {
+			try {
+				validators.add(validator.newInstance());
+			}
+			catch (InstantiationException | IllegalAccessException ex) {
+				throw new RuntimeException(ex);
```
JurrianFahner: Is there a better way to implement this? However, I don't expect it will create a lot of trouble.

philwebb: `BeanUtils.instantiateClass` will create the instance and throw a `BeanInstantiationException` on error.

JurrianFahner: Thanks! I'll refactor it later today.

philwebb:
I agree that plugging in additional validation is currently a bit cumbersome, but I'm not too keen to add an additional attribute to the annotation. There are a couple of aspects that concern me, namely:

* The validator is created using `BeanUtils.instantiateClass(validatorClass)` so can't use standard bean callback features.
* The `Validator.supports` method doesn't make much sense if the validator has been specifically chosen.

I wonder if there's some other way that we could improve support, or perhaps the existing `@Constraint` from `javax.validation` is enough to solve your issue?

snicoll:
Gentle ping @JurrianFahner

JurrianFahner:
@philwebb I agree that JSR303 is the best way to go for all other edge cases. Due to this PR I got a deep dive into JSR303. 
Thank you very much! :+1: 

I'll close this PR for now

For everyone else, if you want to do advanced JSR303 with configuration properties (something more than just min, max, etc), please checkout my repo https://github.com/JurrianFahner/demo-configuration-properties

snicoll:
Thanks for the feedback @JurrianFahner 👍

PR #17412, opened 2019-07-02
Switch Jersey starter to use Spring 5 module

dreis2211:
Hi,

I noticed that we upgraded to Jersey 2.29 recently. That should give us the possibility to switch from `jersey-spring4` to `jersey-spring5`. Unfortunately, the new module doesn't seem to be exposed through the BOM yet, so I added it explicitly for the time being and opened https://github.com/eclipse-ee4j/jersey/issues/4184 to include this in the BOM.

Let me know what you think.
Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #17440, opened 2019-07-05
Remove ArrayList in RestTemplateBuilder.listOf()

izeye:
This PR removes `ArrayList` creation  in `RestTemplateBuilder.listOf()` as it doesn’t seem to be necessary.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/RestTemplateBuilder.java
@@ -686,7 +686,7 @@ public class RestTemplateBuilder {
 	}
 
 	private static <T> List<T> listOf(T[] items) {
-		return Collections.unmodifiableList(new ArrayList<>(Arrays.asList(items)));
+		return Collections.unmodifiableList(Arrays.asList(items));
 	}
 
 	private static <T> Set<T> append(Collection<? extends T> collection, Collection<? extends T> additions) {


nosan:
@izeye 
If these changes are merged the test will fail:
```java

	@Test
	void builder() throws IOException {
		RestTemplateBuilder builder = new RestTemplateBuilder();
		AcceptHeader header = new AcceptHeader();
		header.add(MediaType.APPLICATION_JSON_VALUE);
		RestTemplate template = builder.defaultHeader(header.name, header.values).build();
		header.add(MediaType.APPLICATION_XML_VALUE);
		RestTemplate template1 = builder.defaultHeader(header.name, header.values).build();
		URI uri = URI.create("http://localhost:8080");

		ClientHttpRequest request = template.getRequestFactory().createRequest(uri, HttpMethod.POST);
		assertThat(request.getHeaders().getAccept()).containsExactly(MediaType.APPLICATION_JSON);

		ClientHttpRequest request1 = template1.getRequestFactory().createRequest(uri, HttpMethod.POST);
		assertThat(request1.getHeaders().getAccept()).containsExactly(MediaType.APPLICATION_JSON,
				MediaType.APPLICATION_XML);
	}

	static class AcceptHeader {

		private final String name = HttpHeaders.ACCEPT;

		private final String[] values = new String[10];

		private int index = 0;

		void add(String value) {
			this.values[this.index++] = value;
		}

	}

```

izeye:
@nosan Thanks for the feedback! I didn't expect the backing array to be modified but it doesn't seem to be the case. Closing.

wilkinsona:
Thanks for spotting that, @nosan, but I'm not sure that we want to preserve that behaviour. We don't, as far as I can see, have a test for it and I think users may be surprised by the behaviour, particularly given that it's a varargs method. We should either add a test that verifies the current behaviour and avoids a regression or we should make the change as was proposed here. I've opened #17441 to do that.

PR #17444, opened 2019-07-05
Use Assertions.contentOf() where possible

dreis2211:
Hi,

just found a couple of places where `Assertions.contentOf` could be used.

Let me know what you think.
Cheers,
Christoph


spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/test/java/org/springframework/boot/gradle/tasks/bundling/AbstractBootArchiveTests.java
```diff
@@ -248,7 +249,7 @@ void launchScriptCanBePrepended() throws IOException {
 		properties.put("initInfoProvides", this.task.getBaseName());
 		properties.put("initInfoShortDescription", this.project.getDescription());
 		properties.put("initInfoDescription", this.project.getDescription());
-		assertThat(Files.readAllBytes(this.task.getArchivePath().toPath()))
```
wilkinsona: I deliberately used byte-based checks when I wrote this as the archive is predominantly binary data. I think I'd prefer to keep it that way here and in the cases below as well.

dreis2211: Sounds fair, but I can't promise I will think of it in the future and might change it again. ;-)

PR #17471, opened 2019-07-09
Remove folder that only contains package-info

dreis2211:
Hi,

I just noticed that `spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/context` consists only of a package-info.java file. Imho, this could be cleaned up.

Let me know what you think.
Cheers,
Christoph


wilkinsona:
Thank you, @dreis2211.

Issue #17464, opened 2019-07-09
Spring Boot health check fails with a user-provided RestHighLevelClient as the health indicator continues to use the auto-configured RestClient


sudoapt-getclean:
In my `@SpringBootApplication` class I define this restclient

```java
    @Bean(destroyMethod = "close")
    public RestHighLevelClient client() {
        ClientConfiguration clientConfiguration = ClientConfiguration.builder()
            .connectedTo(elasticsearchHost)
            .usingSsl()
            .build();

        return RestClients.create(clientConfiguration).rest();
    }
```
This all works fine. however when the /actuator/health endpoint of the application is called. It fails and I get a stacktrace from elastic that shows connection refused.
Some debugging showed me that this is because I have two RestClient beans. The second bean is created by the code in RestClientAutoconfiguration below. This is done before my RestClient is created. 
This other Restclient is created with the default localhost:9200 configuration and is used for the health check (but not for the actual searching)

```java
/**
 * {@link EnableAutoConfiguration Auto-configuration} for Elasticsearch REST clients.
 *
 * @author Brian Clozel
 * @since 2.1.0
 */
@Configuration(proxyBeanMethods = false)
@ConditionalOnClass(RestClient.class)
@EnableConfigurationProperties(RestClientProperties.class)
public class RestClientAutoConfiguration {

	@Bean
	@ConditionalOnMissingBean
	public RestClient restClient(RestClientBuilder builder) {
		return builder.build();
	}
```
I happen to be running spring boot 2.2.0.M4 (Because of the support for using the highlevel restclient yay)
Keep up the good work and let me know If you need more information

--- Appendix ---
the original stacktrace is
```
2019-07-09 13:54:19.981 DEBUG 1 --- [io-8080-exec-10] o.s.web.servlet.DispatcherServlet        : GET "/actuator/health", parameters={}
2019-07-09 13:54:20.000  WARN 1 --- [io-8080-exec-10] s.b.a.e.ElasticsearchRestHealthIndicator : Elasticsearch health check failed

java.net.ConnectException: Connection refused
	at org.elasticsearch.client.RestClient$SyncResponseListener.get(RestClient.java:959) ~[elasticsearch-rest-client-6.7.2.jar!/:6.7.2]
	at org.elasticsearch.client.RestClient.performRequest(RestClient.java:233) ~[elasticsearch-rest-client-6.7.2.jar!/:6.7.2]
```

sudoapt-getclean:
[...]

PR #17488, opened 2019-07-11
Create RestClient bean from a RestHighLevelClient if such a bean has been contributed by the application.

nosan:
see gh-17464

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/rest/RestClientAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/rest/RestClientAutoConfiguration.java
@@ -30,6 +30,7 @@ import org.elasticsearch.client.RestHighLevelClient;
 
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/rest/RestClientAutoConfiguration.java
```diff
@@ -92,4 +93,17 @@ public RestHighLevelClient restHighLevelClient(RestClientBuilder restClientBuild
...
 	}
 
+	@Configuration(proxyBeanMethods = false)
+	@ConditionalOnClass(RestHighLevelClient.class)
+	@ConditionalOnBean(RestHighLevelClient.class)
+	public static class RestClientConfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean
+		public RestClient restClient(RestHighLevelClient restHighLevelClient) {
```
bclozel: There's already another bean method named for `RestClient` with the same name. Wouldn't that create issues with bean overriding?
Maybe we could change the existing bean method and inject `ObjectProvider<RestHighLevelClient>` and do everything there (return the client from the high level if it exists or create a new one from the builder). What fo you think (I've quickly looked at your PR, so I might haved missed something)?

nosan:
Thank you, @bclozel 

> There's already another bean method named for RestClient with the same name. Wouldn't that create issues with bean overriding?

Everything is ok, at least I added a test to check this.

> Maybe we could change the existing bean method and inject ObjectProvider<RestHighLevelClient>
 
We could not change the existing bean method as `RestHighLevelClient` and `RestClient` are present in different jars.
Use `ObjectProvider<RestHighLevelClient>` instead of `RestHighLevelClient` bean is a good idea 👍,
at least it is a guarantee that configuration will be applied only if there is a unique bean of `RestHighLevelClient`. (I added a test to check this as well)


PR has been updated.

snicoll:
Thanks again for the pr @nosan.

> Everything is ok, at least I added a test to check this.

It is ok indeed as they are conditions to avoid creating it. However `restClient` is quite generic and I think we should rename it. I've created #17725 

@nosan the first test in your PR has [an invalid assertion](https://github.com/spring-projects/spring-boot/pull/17488/files#diff-4010630ff478b9d04bdbb26d34a2765bR62). I fix it to assert that the rest client is the expected one and, of course, it failed since I've changed the assertion. The test failed as things were not created in the expected order.

Unrelated to your PR but there was a conceptual cycle as the inner class (processed first) required the `RestClientBuilder` that is provided by the auto-configuration class itself. I've polished the auto-configuration to avoid this cycle and provide an explicit order.

nosan:
Thank you so much, @snicoll 
> the first test in your PR has an invalid assertion. I fix it to assert that the rest client is the expected one and, of course, it failed since I've changed the assertion. The test failed as things were not created in the expected order.

 Sorry about that :confused:

Issue #16957, opened 2019-05-24
Provide a JUnit 5 equivalent of ModifiedClassPathRunner


wilkinsona:
We need this as part of [completing our migration to JUnit 5](/spring-projects/spring-boot/issues/16942). Unfortunately, it isn't possible at the moment. https://github.com/junit-team/junit5/issues/201 is required and there may be some other changes that are needed as well.

sbrannen:
@wilkinsona, the new `InvocationInterceptor` extension API coming in JUnit Jupiter 5.5 may actually be sufficient to port the functionality of `ModifiedClassPathRunner` to JUnit Jupiter.

Since JUnit Jupiter 5.5 will most likely be released in time for inclusion in Spring Boot 2.2, you may wish to give the porting a shot now with 5.5 snapshots or the upcoming 5.5 RC1.

If you need any assistance, just let me know.

/cc @marcphilipp

wilkinsona:
Thanks for the pointer, @sbrannen. Reading through [the commit](https://github.com/junit-team/junit5/commit/bb795075870ca1cd988d93fb359a500c72c0e67f) that introduced the new extension point, I'm struggling to see how it will help with implementing a `ModifiedClassPathRunner` replacement.

A replacement for `ModifiedClassPathRunner` requires the ability to load the test class using a custom class loader. I can't see how that can be achieved with the interception points that are offered and the requirement to always call `proceed` on the given `Invocation`. For example, even without the `proceed` requirement, the signature of `interceptTestClassConstructor` seems to prevents me from loading the test class using a new class loader and returning an instance of it.

What have I missed?

sbrannen:
> What have I missed?

I don't think you overlooked anything.

I overlooked the fact that `ModifiedClassPathTestClass` loads the actual test class by name via the custom `ClassLoader`:

https://github.com/spring-projects/spring-boot/blob/e06325e0296eb2c3d9a762fafa9dfa90117d47cf/spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/runner/classpath/ModifiedClassPathRunner.java#L287

At a first pass through the code (viewing it in the browser), it appeared that only the TCCL was being set and reset, which should be achieveable via the new interceptor extension API.

Sorry for the confusion.

You'll indeed have to wait for custom `ClassLoader` support in JUnit Jupiter in order to implement a `ModifiedClassPathRunner` replacement.

[...]

PR #17491, opened 2019-07-11
Provide a JUnit 5 equivalent of ModifiedClassPathRunner

dreis2211:
Hi,

the last two days I worked on an alternative for the `ModifiedClassPathRunner` in order to complete the migration to JUnit 5 and workaround the missing classloader support described in #16957 and https://github.com/junit-team/junit5/issues/201.

I've implemented an `ModifiedClassPathExtension` that works similar to the old `ModifiedClassPathRunner` and extracted the functionality that is shared between both. I did this because I deprecated `ModifiedClassPathRunner`  instead of directly removing it, but I have no strong feelings for keeping the old stuff personally.

The new extension makes use of JUnit 5's Launcher API and executes the tests with the modified classloader while swallowing the original test results. Yet, it still executes the original tests due to the fact that JUnit 5 unfortunately requires the underlying `Invocation` to be processed. This leads obviously to slightly higher test execution times and makes proper test setup and cleaning (e.g. through `@BeforeAll` and `@AfterAll`) more important, but I found this to be reasonable still.

With the new extension I was able to move all tests to JUnit 5. The extension is obviously not perfect and is a workaround until native classloader support in JUnit 5 arrives, but having no JUnit 4 tests anymore is hopefully a convincing argument for this PR.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/Log4J2MetricsWithLog4jLoggerContextAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/Log4J2MetricsWithLog4jLoggerContextAutoConfigurationTests.java
@@ -18,14 +18,14 @@ package org.springframework.boot.actuate.autoconfigure.metrics;
 
 import io.micrometer.core.instrument.binder.logging.Log4j2Metrics;
 import org.apache.logging.log4j.LogManager;
-import org.junit.Test;
-import org.junit.runner.RunWith;
+import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.extension.ExtendWith;
 
 import org.springframework.boot.actuate.autoconfigure.metrics.test.MetricsRun;
 import org.springframework.boot.autoconfigure.AutoConfigurations;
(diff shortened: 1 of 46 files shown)


dreis2211:
There is a test failing that doesn't seem to have anything to do with my changes as it's not using the new extension. Seems unrelated.

wilkinsona:
Thanks very much for taking a look at this, @dreis2211. While we certainly dislike having a mixture of JUnit 4 and JUnit 5-based tests, I'm afraid we dislike the tests being run twice a little bit more. We're concerned that it'll cause confusion in the future as log output will be generated twice and debugging will be harder as well due to everything being invoked twice.

That said, there's a lot of great stuff here even without the move to the new extension. Would you be interested in splitting this into two? We'd quite like to merge the introduction of the `ModifiedClassPathLoaderFactory` now and leave the move to a new `ModifiedClassPathExtension` till JUnit has the necessary extension point available.

dreis2211:
@wilkinsona That's unfortunate, but understandable. I have one more thing I'd like to try out in order to avoid the duplicated execution. Would this change your mind if that works?

dreis2211:
@wilkinsona I have found a way to avoid the duplicated test execution. Hope that changes your mind about the PR :)

philwebb:
@dreis2211 This was an absolute monster effort!! Thanks so much. It's now been merged to master but I've chopped the PR up into a number of smaller commits. The reflection calls are a little brittle, but hopefully those JUnit internals won't change too much.

Thanks again!

PR #17496, opened 2019-07-12
Test the Gradle Plugin against Gradle 5.5.1

dreis2211:
Hi,

Gradle 5.5.1 was released a few days ago, so I think we can test against it.

Cheers,
Christoph


dreis2211:
Test failure is unrelated.

wilkinsona:
Indeed. Thanks for tracking down that Spring Data change and sorry for the inconvenience. I've reached out to the Spring Data team for some guidance. It's not clear to me if it's a bug in Spring Data JDBC or if we're now missing something in Boot that should create a `JdbcConverter` that's then consumed to create another one.

wilkinsona:
@dreis2211, @odrotbohm is on it.

dreis2211:
@wilkinsona There is no inconvenience for me, so no need to be sorry. I could imagine it's rather an inconvenience for you as it makes reviewing harder when you don't get the feedback about test results. ;-) Thanks for reaching out to the Data team.

Issue #17487, opened 2019-07-03
Whitespace in provider name in the OAuth2 properties causes mismatch


wpuhl10775:
### Summary
Whitespace in provider name in the OAuth2 properties causes mismatch for provider ID as it is not able to match the ID to the other proprty parameters

### Actual Behavior

1. Define oauth2 properties in application.properties file

```
spring.security.oauth2.client.registration.xxx.provider=xxx     
spring.security.oauth2.client.provider.xxx.authorization-uri=http://localhost:8282/o/oauth2/v2/auth
```

2. Start Spring Boot
3. Spring Boot throws error: `Caused by: java.lang.IllegalStateException: Unknown provider ID 'xxx  
 '`

### Expected Behavior
Expect 'xxx' provider name to match provider properties

### Configuration
```
spring.security.oauth2.client.registration.fake_oauth.client-id=dummy-client-id
spring.security.oauth2.client.registration.fake_oauth.client-secret=dummy-client-secret
spring.security.oauth2.client.registration.fake_oauth.client-name=fake_oauth
spring.security.oauth2.client.registration.fake_oauth.scope=
spring.security.oauth2.client.registration.fake_oauth.provider=fake_oauth   
spring.security.oauth2.client.registration.fake_oauth.authorization-grant-type=authorization_code
spring.security.oauth2.client.registration.fake_oauth.client-authentication-scheme=basic
spring.security.oauth2.client.registration.fake_oauth.redirect-uri-template=http://localhost:8000/oseries-auth/login/oauth2/code/fake_oauth

spring.security.oauth2.client.provider.fake_oauth.authorization-uri=http://localhost:8282/o/oauth2/v2/auth
spring.security.oauth2.client.provider.fake_oauth.token-uri=http://localhost:8282/oauth2/v4/token
spring.security.oauth2.client.provider.fake_oauth.user-info-uri=http://localhost:8282/oauth2/v3/userinfo
spring.security.oauth2.client.provider.fake_oauth.user-info-authentication-method=basic
spring.security.oauth2.client.provider.fake_oauth.user-name-attribute=name
```

### Version

Spring boot 2.1.2.RELEASE
spring-security-oauth2-client-5.1.3.RELEASE

philwebb:
I've transferred the issue

philwebb:
See #4106 for a general discussion on whitespace. The last similar issue to this was #15972 where we trimmed the servlet context path.

philwebb:
[...]

PR #17497, opened 2019-07-12
Trimming the whitespace for the OAuth2 provider property.

nosan:
see gh-17487

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientPropertiesRegistrationAdapter.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientPropertiesRegistrationAdapter.java
@@ -56,9 +56,10 @@ public final class OAuth2ClientPropertiesRegistrationAdapter {
 
 	private static ClientRegistration getClientRegistration(String registrationId,
 			OAuth2ClientProperties.Registration properties, Map<String, Provider> providers) {
-		Builder builder = getBuilderFromIssuerIfPossible(registrationId, properties.getProvider(), providers);
+		String provider = StringUtils.trimWhitespace(properties.getProvider());
+		Builder builder = getBuilderFromIssuerIfPossible(registrationId, provider, providers);
 		if (builder == null) {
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientPropertiesRegistrationAdapter.java
```diff
@@ -75,7 +75,8 @@ private static ClientRegistration getClientRegistration(String registrationId,
 
 	private static Builder getBuilderFromIssuerIfPossible(String registrationId, String configuredProviderId,
 			Map<String, Provider> providers) {
-		String providerId = (configuredProviderId != null) ? configuredProviderId : registrationId;
+		String providerId = StringUtils
+				.trimWhitespace((configuredProviderId != null) ? configuredProviderId : registrationId);
```
mbhave: I think this logic can be moved a level up to avoid duplication of the trimming logic?

nosan: Do we need to trim `registrationId` as well?

nosan: updated

mbhave: I suppose it doesn't hurt to trim it but I don't think it's possible to have whitespace at the end of the `registrationId` since it's a key in the map.

nosan: <  I don't think it's possible to have whitespace at the end of the registrationId since it's a key in the map

I've just realized and removed my last comment.

nosan:
Thanks, @philwebb

Issue #17498, opened 2019-07-12
Add HealthIndicator for Hazelcast


Simon3:
I'd like to know if you plan on adding a Hazelcast HealthIndicator to Spring Boot.

philwebb:
Lets close this in favor of PR #17499 (thanks @nosan) and continue any discussion there.

PR #17499, opened 2019-07-12
Add HealthIndicator for Hazelcast

nosan:
see gh-17498

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicator.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.hazelcast;
+
+import java.util.LinkedHashMap;
+import java.util.Map;
+
+import com.hazelcast.core.Endpoint;
+import com.hazelcast.core.HazelcastInstance;
+import com.hazelcast.transaction.TransactionalTask;
+
+import org.springframework.boot.actuate.health.AbstractHealthIndicator;
+import org.springframework.boot.actuate.health.Health;
+import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.util.Assert;
+
+/**
+ * {@link HealthIndicator} for a Hazelcast.
+ *
+ * @author Dmytro Nosan
+ * @since 2.2.0
+ */
+public class HazelcastHealthIndicator extends AbstractHealthIndicator {
+
+	private static final TransactionalTask<?> TASK = (context) -> null;
+
+	private final HazelcastInstance hazelcast;
+
+	public HazelcastHealthIndicator(HazelcastInstance hazelcast) {
+		super("Hazelcast health check failed");
+		Assert.notNull(hazelcast, "HazelcastInstance must not be null");
+		this.hazelcast = hazelcast;
+	}
+
+	@Override
+	protected void doHealthCheck(Health.Builder builder) {
+		this.hazelcast.executeTransaction(TASK);
+		builder.up().withDetails(getDetails());
+	}
+
+	private Map<String, Object> getDetails() {
+		Map<String, Object> details = new LinkedHashMap<>();
+		Endpoint endpoint = this.hazelcast.getLocalEndpoint();
+		details.put("name", this.hazelcast.getName());
+		details.put("uuid", endpoint.getUuid());
+		return details;
+	}
+
+}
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicator.java
```diff
@@ -0,0 +1,100 @@
...
+	protected void doHealthCheck(Health.Builder builder) {
+		HazelcastInstance hazelcast = this.hazelcast;
+		hazelcast.executeTransaction((context) -> builder.up().withDetails(getDetails(hazelcast)));
+	}
+
+	private static Map<String, Object> getDetails(HazelcastInstance hazelcast) {
+		Map<String, Object> details = new LinkedHashMap<>();
+		details.put("name", hazelcast.getName());
+		Endpoint endpoint = hazelcast.getLocalEndpoint();
+		if (endpoint instanceof Member) {
```
snicoll: I would stick to `name` and `uuid` for now. Not denying that extra information might be useful but I'd start with a lighter surface area for a start.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicator.java
```diff
@@ -0,0 +1,100 @@
...
+	public HazelcastHealthIndicator(HazelcastInstance hazelcast) {
+		super("Hazelcast health check failed");
+		Assert.notNull(hazelcast, "HazelcastInstance must not be null");
+		this.hazelcast = hazelcast;
+	}
+
+	@Override
+	protected void doHealthCheck(Health.Builder builder) {
+		HazelcastInstance hazelcast = this.hazelcast;
+		hazelcast.executeTransaction((context) -> builder.up().withDetails(getDetails(hazelcast)));
```
snicoll: Curious as why this needs to run in a transaction.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicatorServerIntegrationTests.java
```diff
@@ -0,0 +1,59 @@
...
+/**
+ * Tests for {@link HazelcastHealthIndicator}.
+ *
+ * @author Dmytro Nosan
+ */
+class HazelcastHealthIndicatorServerIntegrationTests {
+
+	private final HazelcastInstance hazelcastServer = Hazelcast.newHazelcastInstance(new Config());
+
+	private final HazelcastHealthIndicator healthIndicator = new HazelcastHealthIndicator(this.hazelcastServer);
```
snicoll: If the plan is to write an integration test, why don't we use an `ApplicationContextRunner` with the regular auto-configuration? I wouldn't have two dedicated classes for that either.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicatorTests.java
```diff
@@ -0,0 +1,129 @@
...
+	private HazelcastInstance hazelcast;
+
+	@Mock
+	private TransactionalTaskContext context;
+
+	@Mock
+	private Client client;
+
+	@Mock
+	private Member member;
```
snicoll: As some of these mocks are test specific, I wouldn't initialize them at class level but rather in the test that uses them.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicatorServerIntegrationTests.java
```diff
@@ -0,0 +1,59 @@
...
+
+	@AfterEach
+	void stop() {
+		this.hazelcastServer.shutdown();
+	}
+
+	@Test
+	void hazelcastIsUp() {
+		Health health = this.healthIndicator.health();
+		assertThat(health.getStatus()).isEqualTo(Status.UP);
```
snicoll: As this is an integration test, perhaps we could test more than up if we're paying the cost of actually bootstrapping hazelcast.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicator.java
nosan: This is the only way to check `Hazelcast` painless which I have found. If you have any suggestions, please let me know.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicator.java
nosan: Should we still split the details between `client` and `member`?

nosan:
Thank you, @snicoll

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/hazelcast/HazelcastHealthIndicator.java
snicoll: IMO not at this time, no.

nosan:
Thanks once again, @snicoll 

PR has been updated.

snicoll:
Well done @nosan and thank you!

nosan:
Thank you very much, @snicoll

Simon3:
Hi, I see you decided to execute a null transaction on the hazelcastInstance to check its health, which is indeed good for this HealthIndicator. 

But a problem I might have with this is that it doesn't guarantee the current instance is synchronized with the cluster, i.e. all the partition migrations are done. 

I first thought I could get that behaviour by calling _isClusterSafe()_ in the HealthIndicator, but it's actually a bad idea because all instances will be considered unhealthy during any partition migration between any instances. The same problem will occur (to a lower extent) with _isMemberSafe()_. 

So I was wondering if there is any clean way to reach this behaviour, i.e. to _initially_ consider the instance unhealthy until it is in sync with the cluster.

snicoll:
@Simon3 please do not request new feature on a closed issue. If you think the health indicator can be improved or has a problem, feel free to create a separate issue. The health indicator is about whether or not the service can be used.

Issue #17472, opened 2019-07-09
Add idle count (number of idle connections in pool) to DataSourcePoolMetadata and DataSourcePoolMetrics


tgeens:
I would like to track the size of the jdbc connection pool. 

Spring Boot already expose a number of metrics (see `DataSourcePoolMetrics`):

* `jdbc.connections.active` - number of active connections
* `jdbc.connections.min` - minimum number of idle connections in the pool
* `jdbc.connections.max` - maximum number of active connections

What am I missing: the number of idle connections in the pool

I'm expecting the number of connections in the pool to be:
min <= active + idle <= max

Adding a new `jdbc.connections.idle` metric would cover this scenario.

Alternatively - and maybe more in line with micrometer philosophy - would be two gauges contributing to `jdbc.connections.count`, 1st with tag `state=idle` and 2nd with tag `state=active`.

This _idle_-count is currently not exposed by the `DataSourcePoolMetadata` abstraction. From a quick look in the javadocs, it seems to be available in most supported DataSources:

* DBCP2 - [getNumIdle](https://commons.apache.org/proper/commons-dbcp/apidocs/org/apache/commons/dbcp2/BasicDataSource.html#getNumIdle--)
* Tomcat - [getNumIdle](https://tomcat.apache.org/tomcat-9.0-doc/api/org/apache/tomcat/jdbc/pool/DataSourceProxy.html#getNumIdle--)
* Hikari - [getIdleConnections](https://static.javadoc.io/com.zaxxer/HikariCP/3.3.1/com/zaxxer/hikari/metrics/PoolStats.html#getIdleConnections--)

See https://github.com/spring-projects/spring-boot/issues/12788 for a similar discussion about wait-counts - but that stat was not available on all data sources

snicoll:
We'd like to harmonize with what Hikari does for consistency. It does expose this in flat more with an `idle` metric. This would be a breaking change so we should probably add a default method for `getIdle()` that returns `null`

philwebb:
Closing in favor of PR #17504

PR #17504, opened 2019-07-13
add idle count to DataSourcePoolMetadata and DataSourcePoolMetrics

ayudovin:
this [enhancement](https://github.com/spring-projects/spring-boot/issues/17472)

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfigurationTests.java
@@ -17,8 +17,10 @@
 package org.springframework.boot.actuate.autoconfigure.metrics.jdbc;
 
 import java.sql.SQLException;
+import java.util.List;
 import java.util.UUID;
 
+import javax.annotation.PostConstruct;
 import javax.sql.DataSource;
 
 import com.zaxxer.hikari.HikariDataSource;
@@ -29,6 +31,7 @@ import org.junit.jupiter.api.Test;
 
 import org.springframework.aop.framework.ProxyFactory;
 import org.springframework.beans.BeansException;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.BeanPostProcessor;
 import org.springframework.boot.actuate.autoconfigure.metrics.test.MetricsRun;
 import org.springframework.boot.autoconfigure.AutoConfigurations;
@@ -40,6 +43,8 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.Ordered;
 import org.springframework.core.PriorityOrdered;
+import org.springframework.jdbc.core.ConnectionCallback;
+import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.datasource.DelegatingDataSource;
 
 import static org.assertj.core.api.Assertions.assertThat;
@@ -50,6 +55,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * @author Stephane Nicoll
  * @author Andy Wilkinson
  * @author Tommy Ludwig
+ * @author Artsiom Yudovin
  */
 class DataSourcePoolMetricsAutoConfigurationTests {
 
@@ -60,8 +66,8 @@ class DataSourcePoolMetricsAutoConfigurationTests {
 
 	@Test
 	void autoConfiguredDataSourceIsInstrumented() {
-		this.contextRunner.withConfiguration(AutoConfigurations.of(DataSourceAutoConfiguration.class))
-				.run((context) -> {
+		this.contextRunner.withUserConfiguration(HikariPool.class)
+				.withConfiguration(AutoConfigurations.of(DataSourceAutoConfiguration.class)).run((context) -> {
 					context.getBean(DataSource.class).getConnection().getMetaData();
 					MeterRegistry registry = context.getBean(MeterRegistry.class);
 					registry.get("jdbc.connections.max").tags("name", "dataSource").meter();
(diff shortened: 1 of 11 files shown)


ayudovin:
@snicoll, sorry, you are right, I just forgot to update it, I will fix it

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfigurationTests.java
```diff
@@ -185,6 +193,27 @@ private static HikariDataSource createHikariDataSource(String poolName) {
 		return hikariDataSource;
 	}
 
+	@Configuration
```
snicoll: I don't understand why initializing the pool here is necessary. Can you please clarify why this is done as part of this PR?

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/metrics/jdbc/DataSourcePoolMetricsTests.java
```diff
@@ -51,6 +57,27 @@ void dataSourceIsInstrumented() {
 				});
 	}
 
+	@Configuration
+	static class HikariPool {
```
snicoll: Ditto.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/metadata/DataSourcePoolMetadata.java
```diff
@@ -79,4 +80,13 @@
 	 */
 	Boolean getDefaultAutoCommit();
 
+	/**
+	 * Return the number of established but idle connections. Can also return {@code null}
+	 * if that information is not available.
+	 * @return the number of established but idle connections or {@code null}
```
snicoll: This method should have a `@since 2.2.0` on it

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/metadata/HikariDataSourcePoolMetadata.java
```diff
@@ -69,4 +72,14 @@ public Boolean getDefaultAutoCommit() {
...
 	}
 
+	@Override
+	public Integer getIdle() {
+		HikariPool pool = getHikariPool();
+		if (Objects.nonNull(pool)) {
+			return pool.getIdleConnections();
+		}
+
+		return super.getIdle();
```
snicoll: I don't think calling super is what I'd do here

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/metadata/DataSourcePoolMetadata.java
```diff
@@ -79,4 +80,13 @@
...
+	 */
+	default Integer getIdle() {
```
snicoll: Rather than adding this method at the end I'd group it logically with a similar concept. Right below `getActive` sounds good. As a result, methods in implementations and tests should be reordered accordingly.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/metadata/HikariDataSourcePoolMetadata.java
```diff
@@ -69,4 +72,14 @@ public Boolean getDefaultAutoCommit() {
 		return getDataSource().isAutoCommit();
 	}
 
+	@Override
+	public Integer getIdle() {
+		HikariPool pool = getHikariPool();
```
snicoll: I'd do the same as what we do in `getActive()`

ayudovin:
@snicoll, thank you for suggestions, I pushed changes to follow by this suggestion. Regarding the tests, it was just my mistake and I have fixed it now.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/jdbc/metadata/HikariDataSourcePoolMetadataTests.java
```diff
@@ -57,4 +62,12 @@ private HikariDataSource createDataSource(int minSize, int maxSize) {
 		return dataSource;
 	}
 
+	@Test
+	void getIdle() {
+		// Make sure the pool is initialized
+		JdbcTemplate jdbcTemplate = new JdbcTemplate(getDataSourceMetadata().getDataSource());
+		jdbcTemplate.execute((ConnectionCallback<Void>) (connection) -> null);
+		assertThat(getDataSourceMetadata().getIdle()).isEqualTo(Integer.valueOf(1));
```
tgeens: This assumes that an Hikari connection pool with min=0; max=2 has 1 initial connection ?
The other connection pools seem to default to 0 - or is that because the other connection pools are _not initiliazed_ ?

ayudovin: yes, it's because the other connection pools are not initialized

ayudovin: I guess I need to add the same tests with initialization pool for other connection pools, what do you think about it?

snicoll: We should configure the pool explicitly so that the idle connection stays in the pool indeed.

ayudovin: sorry, I'm not sure that I catch you, Is this way of initializing pool good for the tests?

snicoll: We could use a `HikariDataSource dataSource = createDataSource(0, 2);` in the test itself. If there is a setting that we could set to increase the time an idle connection stays in the pool, that would be nice too as we ensure that our test is asserting the conditions as we expect them.

ayudovin: ok, I have got it.

snicoll:
Thanks again for the PR @ayudovin. That's not exactly what I had in mind for the test, I've updated your proposal with what I had in mind.

ayudovin:
@snicoll, ok, thank you!

PR #17508, opened 2019-07-14
Remove java-statsd-client dependency

dreis2211:
Hi,

I noticed that there is a dependency to `java-statsd-client` that seems superseded by the micrometer StatsD support. At least there is no direct usage of it anymore as far as I can tell, but maybe you know of any libs that need this dependency to be declared.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -70,11 +70,6 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
-			<groupId>com.timgroup</groupId>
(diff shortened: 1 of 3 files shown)


philwebb:
+1 for removing it from our own projects. I think we should also remove it from the managed dependencies but that's a breaking change.

wilkinsona:
I think we tidy up the `actuator-autoconfigure` dependency in 2.1 and remove the dependency management in 2.2. Micrometer's StatsD registry doesn't use `java-statsd-client` so I think it's quite unlikely that users will have a dependency upon it. If they do and they need to manage its version, it's quite straightforward and shouldn't cause many problems, particularly if we mention that change in 2.2's release notes.

wilkinsona:

Issue #17509, opened 2019-07-14
Tomcat crashes when requests containing certain (unencoded) characters.


dirkdeyne:
## Background
	
The HTTP/1.1 specification requires that certain characters are %nn encoded when used in URI paths. Unfortunately, many user agents including all the major browsers are not compliant with this specification and use these characters in unencoded form.

## Problem

Tomcat crashes when users make requests containing certain (unencoded) characters 

like: _http://localhost:8080/hello?message=[^_^]_

The user gets an ugly _HTTP Status 400 – Bad Request with a stack trace_

_console output:_
```
java.lang.IllegalArgumentException: Invalid character found in the request target. The valid characters are defined in RFC 7230 and RFC 3986
  at org.apache.coyote.http11.Http11InputBuffer.parseRequestLine(Http11InputBuffer.java:467) ~[tomcat-embed-core-9.0.19.jar:9.0.19]
  at org.apache.coyote.http11.Http11Processor.service(Http11Processor.java:294) ~[tomcat-embed-core-9.0.19.jar:9.0.19]
  at org.apache.coyote.AbstractProcessorLight.process(AbstractProcessorLight.java:66) [tomcat-embed-core-9.0.19.jar:9.0.19]
  at org.apache.coyote.AbstractProtocol$ConnectionHandler.process(AbstractProtocol.java:836) [tomcat-embed-core-9.0.19.jar:9.0.19]
  at org.apache.tomcat.util.net.NioEndpoint$SocketProcessor.doRun(NioEndpoint.java:1747) [tomcat-embed-core-9.0.19.jar:9.0.19]
  at org.apache.tomcat.util.net.SocketProcessorBase.run(SocketProcessorBase.java:49) [tomcat-embed-core-9.0.19.jar:9.0.19]
  at java.util.concurrent.ThreadPoolExecutor.runWorker(ThreadPoolExecutor.java:1149) [na:1.8.0_162]
  at java.util.concurrent.ThreadPoolExecutor$Worker.run(ThreadPoolExecutor.java:624) [na:1.8.0_162]
  at org.apache.tomcat.util.threads.TaskThread$WrappingRunnable.run(TaskThread.java:61) [tomcat-embed-core-9.0.19.jar:9.0.19]
  at java.lang.Thread.run(Thread.java:748) [na:1.8.0_162]
```

_references:_
[...]

PR #17510, opened 2019-07-14
Provide configuration properties for configuring Tomcat's relaxed path and query chars

dirkdeyne:
Make Tomcat's Context attributes relaxedPathChars and relaxedQueryChars configurable in application.properties

Fixes gh-17509

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -61,6 +61,7 @@ import org.springframework.util.unit.DataSize;
  * @author Artsiom Yudovin
  * @author Andrew McGhie
  * @author Rafiullah Hamedy
+ * @author Dirk Deyne
  * @since 1.0.0
  */
 @ConfigurationProperties(prefix = "server", ignoreUnknownFields = true)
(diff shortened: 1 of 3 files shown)


philwebb:
I wonder if we should consider enabling both of these my default? @markt-asf would there be any significant downsides? Did you consider changing the Tomcat defaults ever?

vpavic:
+1 for exposing these as configuration properties, but I wouldn't like to see Spring Boot configure non spec compliant behavior by default.

markt-asf:
Tomcat tightened up parsing to address security issues (request injection and such). There are some characters that are probably safe but my strong preference is that Tomcat is spec compliant by default.

dirkdeyne:
only ```" < > [ \ ] ^ ` { | }``` characters are allowed by tomcat, others are ignored

ref [doc](https://tomcat.apache.org/tomcat-9.0-doc/config/http.html)
> The HTTP/1.1 specification requires that certain characters are %nn encoded when used in URI query strings. Unfortunately, many user agents including all the major browsers are not compliant with this specification and use these characters in unencoded form. To prevent Tomcat rejecting such requests, this attribute may be used to specify the additional characters to allow. If not specified, no additional characters will be allowed. The value may be any combination of the following characters: " < > [ \ ] ^ ` { | } . Any other characters present in the value will be ignored.

philwebb:
Thanks for the input everyone. We're going to add the properties but keep the defaults as they are.

snicoll:
@dirkdeyne thank you for making your first contribution to Spring Boot. I've polished your proposal by using a list of characters rather than the raw string that Tomcat uses. This is also a chance for us to provide dedicated metadata for valid values,

image

qrqhuang:
How can i configured this properties with yml.

```
server:
  tomcat:
    relaxed-query-chars: []
``

If i do like this, `[]` will be translate as array mark.

**Comment by philwebb** [comment]
@qrqhuang You can also do the following:

```yaml
server:
  tomcat:
    relaxed-query-chars: "[,]"
```

et00448:
Hello, this new tomcat relaxed-query-chars works perfectly with spring boot 2.2.5
I have no more http 400 error produce by coyotte TOMCAT connector.
Nevertheless, I m facing an URLmalformedException in ServletHttpHandlerAdapter while receiving non ULencode "|" charcater. 
How can I bypass this exception in ServletHttpHandlerAdapter "Failed to get request  URL: Illegal character in query at index ..."
Thank's

wilkinsona:
@et00448 `ServletHttpHandlerAdapter` is part of Spring Framework. I don't believe it provides a mechanism for dealing with characters that are illegal in a URI. If you think that's something that should be possible, please [raise it with the Spring Framework team](https://github.com/spring-projects/spring-framework).

PR #17511, opened 2019-07-14
Add more Servlet filters to the documentation

larsgrefer:
This makes it easier to decide how custom filters should be ordered.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
@@ -3106,6 +3106,21 @@ precedence):
 |`ErrorPageFilter`
 |`Ordered.HIGHEST_PRECEDENCE + 1`
 
+|`SessionRepositoryFilter`
+|`Integer.MIN_VALUE + 50`
+
+|`OrderedHiddenHttpMethodFilter`
+|`OrderedFilter.REQUEST_WRAPPER_FILTER_MAX_ORDER - 10000`
+
+|`OrderedFormContentFilter`
+|`OrderedFilter.REQUEST_WRAPPER_FILTER_MAX_ORDER - 9900`
+
+|`OrderedRequestContextFilter`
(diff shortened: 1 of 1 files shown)


philwebb:
I'm a little worried this starts to make the documentation too overwhelming. Perhaps we can just add a NOTE about where to look in the source code?

wilkinsona:
I think we either need to ensure that the table is complete or remote it entirely and add a note about where to look. Rather than pointing people to the code, I wonder if we could use logging so that the ordering is shown at startup when enabled. If we could do it via `FilterRegistrationBean` it would work for first- and third-party filters and would always be up-to-date.

vpavic:
> I wonder if we could use logging so that the ordering is shown at startup when enabled.

This would be very helpful IMO.

wilkinsona:
Here's how it looks with a small change to `org.springframework.boot.web.servlet.AbstractFilterRegistrationBean.toString()` and 
debug web logging (`logging.level.web=debug`):

```
2019-07-15 09:19:51.080 DEBUG 29599 --- [           main] o.s.b.w.s.ServletContextInitializerBeans : Mapping filters: characterEncodingFilter urls=[/*] order=-2147483648, hiddenHttpMethodFilter urls=[/*] order=-10000, formContentFilter urls=[/*] order=-9900, requestContextFilter urls=[/*] order=-105
```

philwebb:
I prefer the logging I think.

wilkinsona:
I've opened https://github.com/spring-projects/spring-boot/issues/17520. Thanks anyway for the proposal, @larsgrefer, and the inspiration that it provided.

larsgrefer:
I agree that adding all filters to the documentation isn't the best idea. Keeping the documentation up-to-date would be too much work or the risk of having a out-of-date documentation is too high.

Pointing to the source code would be difficult, too. On the one hand, there is not "the one location" to look at (you'd have to find all usages of `FilterRegistationBean`) and on the other hand one would need a pretty good understanding of the spring boot internals in order to get this rather simple information.

I think improving the logging and maybe explaining in the docs how to activate this particular loggging is the best solution.

Issue #17397, opened 2019-07-02
Consider restructuring Spring Session configuration properties


vpavic:
This is a follow-up on #17278, which handled changes to flush mode related configuration properties. More changes in configuration support infrastructure are expected for Spring Session Corn-M3, and when that's in place it might be good idea to revisit the overall structure of Spring Session related configuration properties.

wilkinsona:
One change that we've already identified is to introduce `spring.session.servlet.<store>.*` properties where there's a store-specific configuration setting that applies to the Servlet web stack and not to the reactive web stack.

snicoll:
Unless I misunderstood your comment @wilkinsona, it feels inconsistent to me to introduce the `servlet` prefix only if necessary. Shouldn't we harmonize and have a `servlet` and `reactive` for everything, excluding `reactive` for store that may not work (yet?) with the reactive stack?

wilkinsona:
I'd envisaged three "sets" of properties:

- `spring.session.<store>.*` for store-specific configuration that applies to both web stacks
- `spring.session.servlet.<store>.*` for store-specific configuration that only applies to the Servlet web stack
- `spring.session.reactive.<store>.*` for store-specific configuration that only applies to the reactive web stack

snicoll:
Thanks for the feedback. Considering that we may add things in one stack or the other (or both) I think I prefer getting with the consistency to not have the first case at all and just duplicate configuration options in `servlet` and `reactive` when both are available.

vpavic:
I agree with what @snicoll outlined - that was also my original understanding from #17278 discussion. 

Omitting stack prefix seems risky in terms of breaking users, as it's not always easy to know whether a `SessionRepository` implementation will adopt some feature, and if it does at what pace compared to other implementations.

snicoll:
We've discussed this one this week and the plan is to go with a store specific configuration and keep the existing properties in 2.2 in a deprecated fashion. For stores that only support `servlet`, a replacement can be provided right the way. However, for stores that support both, we need some logic and the user will have to choose explicitly.

Looking at the [current package structure](https://github.com/spring-projects/spring-boot/tree/ea1dc85d50c1e2c86073b84b539515a4af511e9d/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session), introducing additional types do not spark joy so I am considering moving things in a `servlet` and `reactive` sub-packages.

snicoll:
[...]

PR #17514, opened 2019-07-15
Add support for configuring Spring Session SaveMode

vpavic:
This PR adds support for configuring Spring Session's newly introduced `SaveMode` for applicable `SessionRepository` implementations.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/session/SessionAutoConfigurationRedisTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/session/SessionAutoConfigurationRedisTests.java
@@ -33,6 +33,7 @@ import org.springframework.boot.testsupport.testcontainers.RedisContainer;
 import org.springframework.data.redis.connection.RedisConnection;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.session.FlushMode;
+import org.springframework.session.SaveMode;
 import org.springframework.session.data.mongo.MongoOperationsSessionRepository;
 import org.springframework.session.data.redis.RedisOperationsSessionRepository;
 import org.springframework.session.data.redis.config.ConfigureNotifyKeyspaceEventsAction;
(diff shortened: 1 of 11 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/session/SessionAutoConfigurationRedisTests.java
```diff
@@ -75,17 +76,19 @@ void defaultConfigWithUniqueStoreImplementation() {
...
+				.run(validateSpringSessionUsesRedis("spring:session:event:0:created:", FlushMode.ON_SAVE, "0 * * * * *",
+						SaveMode.ON_SET_ATTRIBUTE));
 	}
 
 	@Test
 	void redisSessionStoreWithCustomizations() {
 		this.contextRunner.withConfiguration(AutoConfigurations.of(RedisAutoConfiguration.class))
 				.withPropertyValues("spring.session.store-type=redis", "spring.session.redis.namespace=foo",
-						"spring.session.redis.flush-mode=immediate", "spring.session.redis.cleanup-cron=0 0 12 * * *",
+						"spring.session.redis.flush-mode=immediate", "spring.session.redis.save-mode=on-get-attribute",
```
snicoll: Can you please put the save-mode after the cron so that the order of properties are aligned with the assertion?

vpavic:
I've updated the PR with improved description of `saveMode` property and suggested change to `SessionAutoConfigurationRedisTests`.

snicoll:
Thanks for the update @vpavic. I've been back and forth with the change as I wanted to make sure somewhat related properties are grouped together rather than simply being added at the end. This made me revert part of the polish I suggested, sorry about that.

vpavic:
NP @snicoll - I've also contemplated that but in the end went for the simplest solution (add to end). 

However, we'll probably take a closer look at this on Spring Session side before `2.2.0.RELEASE` so I planned to polish up and harmonize things on the Boot side depending on the outcome of that effort. This could probably be done as a part of #17397.

snicoll:
Roger. I've just flagged that one for team attention so that we discuss the various namespace options at your next team call.

Issue #16680, opened 2019-04-28
Support configuring logging groups via loggers endpoint


mbhave:
I don't think it's possible to update or get the configuration of a group via the loggers endpoint at the moment. In fact, with logback, if you try to `GET` the configuration using the group's name, it creates a child logger with that name which is a bit confusing. Adding an issue to see what the rest of the team thinks about this.

wilkinsona:
I think we should support getting or updating the log level for a logging group via the endpoint. I don't think we should support creating groups or changing the loggers to which an existing group maps via the endpoint.

mbhave:
Yup, that's what I was thinking as well.

mbhave:
Closing in favor of PR #17515.

PR #17515, opened 2019-07-15
LoggerGroupsEndpoint created for modification of logger groups.

htztomic:
This is a fix to GitHub issue #16680 . The LoggerGroupsEndpoint gives support for GET configurations of logger groups or a single logger group as well configuring the LogLevel of a specific logger group.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LoggersEndpointAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LoggersEndpointAutoConfiguration.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.actuate.autoconfigure.logging;
 
+import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.actuate.autoconfigure.endpoint.condition.ConditionalOnAvailableEndpoint;
 import org.springframework.boot.actuate.logging.LoggersEndpoint;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
@@ -24,6 +25,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionOutcome;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.SpringBootCondition;
+import org.springframework.boot.logging.LoggingGroups;
 import org.springframework.boot.logging.LoggingSystem;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.ConditionContext;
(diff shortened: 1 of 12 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/logging/LoggersGroupEndpointAutoConfiguration.java
```diff
@@ -0,0 +1,50 @@
...
+ * {@link EnableAutoConfiguration Auto-configuration} for the
+ * {@link LoggerGroupsEndpoint}.
+ *
+ * @author HaiTao Zhang
+ * @since 2.2.0
+ */
+
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnAvailableEndpoint(endpoint = LoggerGroupsEndpoint.class)
+public class LoggersGroupEndpointAutoConfiguration {
```
snicoll: Without looking at the concrete details, I would have updated the existing `LoggersEndpoint` to handle both logger names and groups. While we're at the mercy of having a group having the same name as a logger, I wonder if handling them in a single place wouldn't be easier.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/logging/LoggersGroupEndpointAutoConfigurationTests.java
```diff
@@ -0,0 +1,87 @@
...
+	}
+
+	@Test
+	void runWithNoLoggingSystemShouldNotHaveEndpointBean() {
+		this.contextRunner.withPropertyValues("org.springframework.boot.logging.LoggingSystem=none")
+				.run((context) -> assertThat(context).doesNotHaveBean(LoggerGroupsEndpoint.class));
+	}
+
+	@Test
+	void runWithNoLoggerGroupsShouldNotHaveEndpointBean() {
```
snicoll: I am not entirely sure I get the rationale of this. IMO, If we go down the route of having a dedicated endpoint, I don't think it should 404 depending of the fact that at least a group is defined or not.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggerGroups.java
```diff
@@ -0,0 +1,65 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * Common abstraction for logger groups.
+ *
+ * @author HaiTao Zhang
+ * @since 2.2.0
+ */
+public class LoggerGroups {
```
snicoll: Have you considered updating `LoggingSystem` rather than registering another singleton in the context? Given that became a feature handled there, perhaps we should expose it rather than having to lookup another type.

snicoll:
@htztomic I missed the conversation of using a separate endpoint while I was on PTO, sorry about that. Please ignore that part of the review for now.

wilkinsona:
When we added support for logging groups we did so without making any, I think, changes to the logging systems. I wonder if we should strive to do the same here? Would it work if we pulled out a separate class, say `LoggerGroups`, that was responsible for managing the groups and passing the level changes for each logger in a group down to the `LoggingSystem`? A `LoggerGroups` instance could be made available as a bean in a similar manner to how the `LoggingSystem` is made available.

snicoll:
Gah, that's exactly what @htztomic had and I suggested in my review to [avoid registering yet another bean](https://github.com/spring-projects/spring-boot/pull/17515#discussion_r304298165).

wilkinsona:
IMO, registering another bean is preferable to adding another responsibility to `LoggingSystem` implementations. A middle-ground would be to introduce `LoggerGroups` as @htztomic originally proposed, but provide access to it via a `getLoggerGroups()` on `LoggingSystem` rather than registering it as a bean. I still prefer the bean registration approach, however. There's no technical need for `LoggingSystem` to know about logger groups and I'd prefer to keep that separation.

PR #17516, opened 2019-07-15
Include junit-platform-launcher in parent POM

dreis2211:
Hi,

after my PR #17491 got merged the full build seems to fail. This PR is an attempt to fix this. I'm very sorry for the inconveniences this causes you and others.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-parent/pom.xml
+++ b/spring-boot-project/spring-boot-parent/pom.xml
@@ -243,6 +243,11 @@
 			<scope>test</scope>
 		</dependency>
 		<dependency>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, @dreis2211. No apology is needed. We've tracked the problem down to a bug in Surefire. Your proposal here will, I believe, work around it, but we'd prefer not to have every module depend upon the launcher as they shouldn't need it. @philwebb is working on a solution that uses reflection instead. Please see https://github.com/spring-projects/spring-boot/issues/17517 for details. If that attempt doesn't work out as we hope it will, we can resurrect this as an alternative.

PR #17519, opened 2019-07-15
Fix formatting in Couchbase smoke tests

dreis2211:
Hi,

apart from the current LauncherFactory problem, the builds fail because of a tiny formatting issue in the Couchbase smoke tests.

Cheers,
Christoph


snicoll:
Stumbled upon that one when validating another change. Thanks for the PR Christoph!

Issue #3071, opened 2015-05-31
spring boot unable to interpolate message for jsr-303 validation


Zane-XY:
if I define the following validation messages,

``` java
@Size(min = 3, message = "{validation.account.name}")
private String name;
```

I expect spring boot would read from `messages*.properties` from classpath and find the `validation.account.name` key to interpolate the message, but it just shows `{validation.account.name}` on the page.

This https://github.com/Zane-XY/springboot-i18n-problem demos the problems.

snicoll:
We don't support that. Spring Framework does support it but only with the Hibernate validator implementation. That's interesting.

php-coder:
IMHO it should be easy to implement, like this:

```java
public class MvcConfig extends WebMvcConfigurerAdapter {

    @Inject
    private MessageSource messageSource;

	@Override
	public Validator getValidator() {
		LocalValidatorFactoryBean factory = new LocalValidatorFactoryBean();
		factory.setValidationMessageSource(messageSource);
		return factory;
	}
}
```

Zane-XY:
Why this isn't part of https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/MessageSourceAutoConfiguration.java? at least Springboot documentation/references should mention what `MessageSourceAutoConfiguration` has done, and how to enable reading from message sources.

pete-experimenter:
Good question. Is there an answer? And how do you make use of the MessageSourceAutoConfiguration?

bigfatsea:
I believe it's still an open issue https://jira.spring.io/browse/SPR-6380

philwebb:
Possibly relates to #7598

selimok:
It works.

To get it worked I have to move validation keys into the file ValidationMessages.properties while all oder messages reside in messages.properties.

And simple configured the validation beans as follows:

	@Bean
	public MethodValidationPostProcessor getMethodValidationPostProcessor() {
	    final MethodValidationPostProcessor methodValidationPostProcessor = new MethodValidationPostProcessor();
	    methodValidationPostProcessor.setValidator(getValidator());
	    return methodValidationPostProcessor;
	}
	
	@Bean
	public LocalValidatorFactoryBean getValidator() {
		LocalValidatorFactoryBean validator = new LocalValidatorFactoryBean();
		return validator;
	}

FYI.

snicoll:
[...]

PR #17530, opened 2019-07-16
Use MessageSource to interpolate bean validation messages

nosan:
see gh-3071

#### Limitation

`MessageSourceInterpolatorDelegate` does not support `\` escape character. The escape character is not supported because there is no way to configure`PropertyPlaceholderHelper` with it.
Is it possible to add escape support in `Spring Framework`? 

Thanks

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/validation/ValidationAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/validation/ValidationAutoConfiguration.java
@@ -25,6 +25,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnResource;
 import org.springframework.boot.validation.MessageInterpolatorFactory;
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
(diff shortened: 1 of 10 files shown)


philwebb:
I've raise https://github.com/spring-projects/spring-framework/issues/23348 to look at extending `PropertyPlaceholderHelper`

nosan:
Thanks, @philwebb  !

nosan:
Hi @philwebb,

I did some research related to this issue, and my idea to use `PropertyPlaceholderHelper`  was doomed initially.

I thought if `PropertyPlaceholderHelper` had escape support it would be possible to
resolve parameters via `MessageSource`, but it is not possible because [Bean Validation Specification](https://beanvalidation.org/2.0/spec/#validationapi-message)
has quite a specific behavior of interpolating messages and of course, `PropertyPlaceholderHelper` would not be able to support such behavior. 

I made a comparison between `Bean Validation` and `Apache StringSubstitutor`, and of course, they have different behavior and most probably `PropertyPlaceholder` would have something different.

| Message Template  | Bean Validation  |  Apache StringSubstitutor |
|---|---|---
| {foo} |{foo}   |  {foo} |
| \\\\{foo}  |\\{foo}   | \\{foo}  |
|  \\{foo} | \\{foo}  |  {foo} |
|  {foo\\} | {foo\\}  |  {foo\\} |
|  \\{foo\\} | {foo}  |  {foo\\} |

I think https://github.com/spring-projects/spring-framework/issues/23348 should be closed, as it would not help us here.

nosan:
Currently, I'm thinking of another variant of how it could be solved.

nosan:
PR has been updated and ready for the review.
Thanks in advance!

wilkinsona:
Thanks very much, @nosan, both for the PR and for your patience while we found time to process it.

nosan:
No problem at all @wilkinsona
Thanks a lot.

Issue #17483, opened 2019-07-10
ActiveMQConnectionFactory is not configured when CachingConnectionFactory is missing and enabled properties are false


nosan:
When both properties `spring.activemq.pool.enabled` and  `spring.jms.cache.enabled` set to `false` and `CachingConnectionFactory` is not on the classpath then `ActiveMQAutoConfiguration` will not register `ActiveMQConnectionFactory` bean.

```java
@RunWith(ModifiedClassPathRunner.class)
@ClassPathExclusions("spring-jms-*.jar")
public class ActiveMQAutoConfigurationTestsWithoutCachingConnectionFactoryTests {

	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
			.withConfiguration(AutoConfigurations.of(ActiveMQAutoConfiguration.class));

	@Test
	public void whenThereIsNoCachingConnectionFactoryOnTheClasspathThenSimpleConnectionFactoryAutoConfigured() {
		this.contextRunner.withPropertyValues("spring.activemq.pool.enabled=false", "spring.jms.cache.enabled=false")
				.run((context) -> assertThat(context).hasSingleBean(ActiveMQConnectionFactory.class));
	}

}
```

It happens because `@ConditionalOnClass(CachingConnectionFactory.class)` is present on `ActiveMQConnectionFactoryConfiguration$SimpleConnectionFactoryConfiguration` configuration class.

nosan:
Here is a possible fix:
<details>
 <summary>Click to expand</summary>

```java
@Configuration(proxyBeanMethods = false)
@ConditionalOnMissingBean(ConnectionFactory.class)
class ActiveMQConnectionFactoryConfiguration {

	private static ActiveMQConnectionFactory createConnectionFactory(ActiveMQProperties properties,
			List<ActiveMQConnectionFactoryCustomizer> connectionFactoryCustomizers) {
		return new ActiveMQConnectionFactoryFactory(properties, connectionFactoryCustomizers)
				.createConnectionFactory(ActiveMQConnectionFactory.class);
	}

	@Configuration(proxyBeanMethods = false)
	@ConditionalOnProperty(prefix = "spring.activemq.pool", name = "enabled", havingValue = "false",
			matchIfMissing = true)
	static class SimpleConnectionFactoryConfiguration {

		@ConditionalOnClass(CachingConnectionFactory.class)
		@ConditionalOnProperty(prefix = "spring.jms.cache", name = "enabled", havingValue = "true",
[...]

PR #17531, opened 2019-07-16
ActiveMQConnectionFactory is not configured when CachingConnectionFactory is missing and enabled properties are false 

nosan:
see gh-17483


nosan:
Thank you, @snicoll

Issue #17533, opened 2019-07-16
When NettyDriverMongoClientSettingsBuilderCustomizer is called multiple times, only the last EventLoopGroup that it creates is closed on shutdown


deripas:
Hi,

In spring boot I create two MongoDB reactive clients:
```java
List<MongoClientSettingsBuilderCustomizer> customizerList;   // autowired
MongoClientSettings.Builder builder = MongoClientSettings.builder().applyConnectionString(connectionString);
customizerList.forEach(customizer -> customizer.customize(builder));
MongoClient mongoClient = MongoClients.create(builder.build());
```
I see problem with `MongoReactiveAutoConfiguration`, rather `NettyDriverMongoClientSettingsBuilderCustomizer`.
In my case the method `NettyDriverMongoClientSettingsBuilderCustomizer#customize` is called twice
```java
                        private volatile EventLoopGroup eventLoopGroup;

			@Override
			public void customize(Builder builder) {
				if (!isStreamFactoryFactoryDefined(this.settings.getIfAvailable())) {
					NioEventLoopGroup eventLoopGroup = new NioEventLoopGroup();
					this.eventLoopGroup = eventLoopGroup;
					builder.streamFactoryFactory(NettyStreamFactoryFactory.builder()
							.eventLoopGroup(eventLoopGroup).build());
				}
			}
```
eventually `new NioEventLoopGroup()` is **created twice**, but reference **only last instance**.
`destroy` method clear only last `NioEventLoopGroup`
```java
			@Override
			public void destroy() {
				EventLoopGroup eventLoopGroup = this.eventLoopGroup;
				if (eventLoopGroup != null) {
					eventLoopGroup.shutdownGracefully().awaitUninterruptibly();
					this.eventLoopGroup = null;
				}
			}
```
And my application not shutdown correctly.

Why we must create `new NioEventLoopGroup()`  on each call `customize`?  Why not:
```java
                        private EventLoopGroup eventLoopGroup = new NioEventLoopGroup();

			@Override
			public void customize(Builder builder) {
				if (!isStreamFactoryFactoryDefined(this.settings.getIfAvailable())) {
					builder.streamFactoryFactory(NettyStreamFactoryFactory.builder()
							.eventLoopGroup(eventLoopGroup).build());
				}
			}
```

mp911de:
[...]

PR #17535, opened 2019-07-16
Use a single instance of `EventLoopGroup` to configure `NettyStreamFactory`

nosan:
Prior to this commit, `NettyDriverMongoClientSettingsBuilderCustomizer` could create more than one instance of `NioEventLoopGroup` in case if it has been invoked for several `MongoClientSettings`, but the reference was only to the last instance, hence shutdown was invoked only for the last `NioEventLoopGroup`.

see gh-17533

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoReactiveAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoReactiveAutoConfiguration.java
@@ -91,8 +91,7 @@ public class MongoReactiveAutoConfiguration {
 		@Override
 		public void customize(Builder builder) {
 			if (!isStreamFactoryFactoryDefined(this.settings.getIfAvailable())) {
-				NioEventLoopGroup eventLoopGroup = new NioEventLoopGroup();
-				this.eventLoopGroup = eventLoopGroup;
+				EventLoopGroup eventLoopGroup = eventLoopGroup();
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, @nosan. I'm not sure what direction we want to take with this one just yet. Two things in particular with your proposal that I'm not sure about:

1. The event loop group will now always be created even if it isn't used
2. The event loop group will be shared across multiple clients

@mp911de what's your take on #17533 and the fix proposed here?

nosan:
Thanks, @wilkinsona 
> The event loop group will now always be created even if it isn't used

Unfortunately yes ( by default, **0** threads)
> The event loop group will be shared across multiple clients

I'm not sure that could be an issue.

As an alternative approach,`List<NioEventLoopGroup` can be used.

mp911de:
> 1. The event loop group will now always be created even if it isn't used

We should do something about that.

>2. The event loop group will be shared across multiple clients

This is actually a good effect. We should ideally keep only a single `EventLoopGroup` around to not waste system resources. See my comment in #17533.

wilkinsona:
> This is actually a good effect. We should ideally keep only a single EventLoopGroup around to not waste system resources. See my comment in #17533.

Thank you, @mp911de. That's what I suspected but wanted some expert confirmation.

Let's see where the discussion on #17535 leads.

bclozel:
I'm closing this PR as I don't think that shared resources like this should be managed by the auto-configuration itself. Just like we have a `ReactorResourceFactory` now in Spring Framework, I believe they should be bound to the context lifecycle with a component in the first place.

I'll add more about that in a comment in #17533

Thanks @nosan for your help!

Issue #17262, opened 2019-06-19
getResourcePaths(String) allows navigation through the loader's directory structure when using Tomcat


wilkinsona:
While we prevent individual `ServletContext.getResource(String)` requests for the loader and do not allow it to be served static resources, when using Tomcat `getResourcePath(String)` will return paths that allow navigation of the loader's directory structure. The same is not possible when using Jetty. We should see if we can prevent it with Tomcat as well.

larsgrefer:
The fix for this should be trivial. Just filter out all paths starting with `org/springframework/boot/loader` in `org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactory.LoaderHidingWebResourceSet#listWebAppPaths`

this might be a candidate for a first-timers-only issue

adavid9:
@wilkinsona I hope this fix will help as was mentioned above I filtered out paths starting with the `org/springframework/boot/loader`

adavid9:
Closed above PR as I missed up to configure the `codeStyleConfig.xml` If you don't mind I will create new PR when configure all of those things.

snicoll:
@adavid9 there is no need to create another PR, please push additional commits on your branch and the existing PR will be updated automatically.

adavid9:
I had to close previously PR, because I did not create it on a separate branch I missed it somehow, sorry.

snicoll:
No problem @adavid9 and thanks for your efforts!

wilkinsona:
Closing in favour of #17538.

PR #17538, opened 2019-07-16
getResourcePaths(String) allows navigation through the loader's directory structure when using Tomcat

adavid9:
See gh-17262


wilkinsona:
Thank you very much for making your first contribution to Spring Boot, @adavid9. The proposed changes are now in both 2.1.x along with a [little bit of polish](https://github.com/spring-projects/spring-boot/commit/591250f75e04c47803a490f7c371291298e81d28). The polish primarily correct that path that was suggested to add a leading `/`. It also adds some tests to verify the behaviour.

PR #17549, opened 2019-07-17
Update JavaVersion class to use java.lang.Runtime$Version.

nosan:

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/condition/ConditionalOnJavaTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/condition/ConditionalOnJavaTests.java
@@ -80,14 +80,14 @@ class ConditionalOnJavaTests {
 	void equalOrNewerMessage() {
 		ConditionOutcome outcome = this.condition.getMatchOutcome(Range.EQUAL_OR_NEWER, JavaVersion.NINE,
 				JavaVersion.EIGHT);
-		assertThat(outcome.getMessage()).isEqualTo("@ConditionalOnJava (1.8 or newer) found 1.9");
+		assertThat(outcome.getMessage()).isEqualTo("@ConditionalOnJava (1.8 or newer) found 9");
 	}
 
 	@Test
 	void olderThanMessage() {
 		ConditionOutcome outcome = this.condition.getMatchOutcome(Range.OLDER_THAN, JavaVersion.NINE,
 				JavaVersion.EIGHT);
-		assertThat(outcome.getMessage()).isEqualTo("@ConditionalOnJava (older than 1.8) found 1.9");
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the proposal, @nosan. I see you've expanded the `JavaVersion` enum. That's a good thing to do but I think it should be tackled separately. I've opened https://github.com/spring-projects/spring-boot/issues/17565.

I'm not sure about the move to using reflection here. Our current approach feels cleaner to me, but let's see what the rest of the team thinks. Was there a reason for the change, perhaps because there isn't a class that can be used for the new versions that you've also proposed?

philwebb:
It looks like `Runtime.version()` was [added in Java 9](https://download.java.net/java/early_access/jdk13/docs/api/java.base/java/lang/Runtime.html#version()) so I guess the call isn't available to us without using reflection.

wilkinsona:
Yeah, we have to use reflection if we want to use it. I'm just not sure that we should. Furthermore, the `major()` method that this proposal calls reflectively was deprecated in Java 10. `feature()` is the recommended replacement there.

philwebb:
I was trying to find a new class that was added in 10 _and_ is in the `java.base` module so we know it'll always be there. I didn't have a lot of luck. I think using the `Runtime.version()` method is probably our only option going forward if we want to offer all versions in the enum.

nosan:
Thanks, @wilkinsona 

I have not found a new class that was added in java `10`. I believe that `JavaVersion` enum should be dependent on `java.lang.Runtime$Version` instead of `Class.isPresent(...)`.


PR has been updated to use either `major` or `feature` methods.

dreis2211:
@philwebb & @nosan If one can trust the following [JDK API Diff reports](https://github.com/AdoptOpenJDK/jdk-api-diff/blob/master/README.md#published-reports) `JDK 10` could be handled via `java.lang.invoke.ClassSpecializer$Factory` I guess.
11 (`java.lang.invoke.ConstantBootstraps`) and 12 (`java.lang.invoke.TypeDescriptor`) are also fine, but 13 is missing a new class so far you could check for.

Alternatively, one could check for new methods in very common `java.lang` classes. It would still involve the reflection ecosystem in the end but less obvious. E.g.:

| JDK | Method | Note |
| --- | --- | --- |
| JDK 8 | `ClassUtils.hasMethod(Optional.class, "empty")` | |
| JDK 9 | `ClassUtils.hasMethod(Optional.class, "stream")` | |
| JDK 10 | `ClassUtils.hasMethod(Optional.class, "orElseThrow")` | |
| JDK 11 | `ClassUtils.hasMethod(String.class, "strip")` | |
| JDK 12 | `ClassUtils.hasMethod(String.class, "describeConstable")` | |
| JDK 13 | `ClassUtils.hasMethod(String.class, "stripIndent")` | Deprecated but works |

nosan:
Thank you, @dreis2211 

looks pretty good  👍

dreis2211:
Let's see what the team thinks :)

philwebb:
Thanks for the PR @nosan. I'm going to close this one in favor of the method suggested by @dreis2211. We'll track this in #17565

Issue #17023, opened 2019-05-30
Replace deprecated MediaType.APPLICATION_JSON_UTF8 usage


philwebb:
Since it's deprecated in Spring Framework we should also stop using it.

bclozel:
Related to (but does not duplicate) #16979 - I see we have at least one of those in `DefaultErrorWebExceptionHandler`.

pluttrell:
Will a fix for this be in 2.2.0?

philwebb:
Closing in favor of PR #17596

philwebb:
> Will a fix for this be in 2.2.0? 

@pluttrell Yes we merged the PR and it will be in M5

PR #17596, opened 2019-07-20
Replace deprecated MediaType.APPLICATION_JSON_UTF8 usage

adavid9:
See gh-17023

I think I've searched the entire project and this was used only in this one class.


snicoll:
Thanks @adavid9!

PR #17617, opened 2019-07-24
Provide a way to run HealthIndicators concurrently

nosan:
see gh-2652

Please let me know if I need something to document.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/health/HealthEndpointConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/health/HealthEndpointConfiguration.java
@@ -21,6 +21,7 @@ import org.springframework.boot.actuate.health.CompositeHealthIndicator;
 import org.springframework.boot.actuate.health.HealthAggregator;
 import org.springframework.boot.actuate.health.HealthEndpoint;
 import org.springframework.boot.actuate.health.HealthIndicatorRegistry;
+import org.springframework.boot.actuate.health.HealthIndicatorStrategy;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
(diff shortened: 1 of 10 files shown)


nosan:
Thank you very much @snicoll 

PR has been updated. 

I believe it is what you are expecting.

nosan:
From another point of view, it does make sense to introduce `HealthIndicatorStrategy` interface.
For now, it could be two different implementations:
-  `DefaultHealthIndicatorStrategy` - returns health indications from all `HealthIndicators` sequentially
-  `ConcurrentHealthIndicatorStrategy` - returns health indications from all `HealthIndicators` concurrently

Update `CompositeHealthIndicator` to use `HealthIndicatorStrategy` to collect health results and finally expose `HealthIndicatorStrategy` as a bean.

nosan:
@snicoll 
There are 2 approaches:
- https://github.com/spring-projects/spring-boot/pull/17617/commits/9344765f2e27c115a05639d76c072f6aca379be8 
- https://github.com/spring-projects/spring-boot/pull/17617/commits/e59b517ca6371f6ca036c15699ee5ca73b9b03a1 

1. Update `CompositeHealthIndicator` with new methods `parallel` and `sequential` to configure a strategy that should be used.  Is not possible to define your own strategy at all. There is no way to configure strategy via bean. 
2. Update `CompositeHealthIndicator` to use `HealthIndicatorStrategy`. Currently, I've added two strategies `ConcurrentlyHealthIndicatorStrategy` and `DefaultHealthIndicatorStrategy`. Easy to define your own strategy.  Strategy can be registered as a bean.

snicoll:
@nosan there was a major refactoring of the health support this week. Would you be interested to revisit your original proposal against `master` ?

nosan:
@snicoll 
sure, but I need some time to check this out.

nosan:
@snicoll 
I'm closing this for now, as I need more time than I have expected. 
Sorry for the inconvenience and thank you for your time.

snicoll:
No need to apologize @nosan and thank you for the feedback.

Issue #17634, opened 2019-07-25
Reactive health indicators do not log health check failure


superlee9527:
Spring-boot 2.0.6 use lettuce as redis default client, and redis health check in actuator do not log the exception. If my redis host is wrong, the console log does not show exception message.
I can only get the cause by `/actuator/health`.
Why not log warn the exception in `AbstractReactiveHealthIndicator.java` in spring-boot-actuator like this:
```java
public abstract class AbstractReactiveHealthIndicator implements ReactiveHealthIndicator {
    private static final String DEFAULT_MESSAGE = "Health check failed";

    private final Log logger = LogFactory.getLog(getClass());

    private final Function<Exception, String> healthCheckFailedMessage;

    protected AbstractReactiveHealthIndicator(String healthCheckFailedMessage) {
	this.healthCheckFailedMessage = (ex) -> healthCheckFailedMessage;
    }
    @Override
    public final Mono<Health> health() {
	try {
	    return doHealthCheck(new Health.Builder()).onErrorResume(this::handleFailure);
	}
	catch (Exception ex) {
            //log the exception
             if (this.logger.isWarnEnabled()) {
		String message = this.healthCheckFailedMessage.apply(ex);
		this.logger.warn(StringUtils.hasText(message) ? message : DEFAULT_MESSAGE, ex);
	    }
	    return handleFailure(ex);
	}
    }
    ···········
}
```

snicoll:
Thanks, it's an oversight and inconsistent with what the imperative health indicators do.

snicoll:
Closing in favour of #17635

PR #17635, opened 2019-07-25
Reactive health indicators do not log health check failure

nosan:
see gh-17634


nosan:
oops, `@since` tag is wrong :(

snicoll:
Thanks once again, @nosan.

nosan:
Thanks, @snicoll

PR #17636, opened 2019-07-25
Escape "|" in javadoc to fix documentation issue

vivganes:
Fixes a documentation bug in the following URL.

URL : https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#common-application-properties

The table is broken and shows wrong data for the attribute `server.tomcat.relaxed-path-chars` due to a misplaced pipe symbol.


snicoll:
@vivganes thanks for reporting the issue and submitting a PR. I am afraid this should be fixed at the documentation level rather than escaping a valid character. I've created #17657 to track this effort.

PR #17648, opened 2019-07-26
Add missing assertion in dispatcherServletCustomConfig()

izeye:
This PR adds a missing assertion in `DispatcherServletAutoConfigurationTests.dispatcherServletCustomConfig()` as a property has been added in c7d2799f4e1b42daa96ea1c1190612eb0627039e but an assertion for it seems to be missing.


snicoll:

Issue #17610, opened 2019-07-23
Provide dedicated Neo4j driver auto-configuration


michael-simons:
@snicoll and I have discussed the following idea:

At the moment, there's auto configuration for Spring Data Neo4j+Object Graph Mapping (In the following SDN+OGM). The auto configuration for SDN+OGM does several things

* It configures the "OGM Driver". The OGM Driver is an additional abstraction over several ways to address a Neo4j Instance. It can be HTTP, Embedded or the Java Driver. 
* It configures the Spring Data part of SDN+OGM, that is enabling `Neo4jRepositories` and related infrastructure.

Spring Boot might contain two distinct aspects of configuring Neo4j and Spring Data Neo4j in the future:

* The Neo4j Java Driver (aka "Bolt driver") should be configured without going through Spring Data Neo4j, so that users can work with the plain driver object like they might do with a JDBC `DataSource` 
* The configuration would need new properties outside the `spring.data` namespace
* The auto configuration for SDN+OGM would need to be improved to act according to this:
  -- If there's a `Driver` bean, than use that to configure the OGM part accordingly 
  -- If there's a none, check the old properties whether SDN+OGM can configured as before

This change would require efforts in the current OGM as well.

The benefits of such a change would be 

* More choice for the users (direct Java Driver usage of our product or Spring Data abstraction)
* An migration path to upgrade to the new Neo4j driver (2.0/4.0, which will be not compatible with OGM).

We have written down about the current building blocks of an SDN+OGM application [here]( https://michael-simons.github.io/neo4j-sdn-ogm-tips/what_are_the_building_blocks_of_sdn_and_ogm.html).

Our goal is to move users to a server based solution (away from embedded) over the Bolt driver. Embedded is there to stay, but we rather see it as testing support only. That means, we might be touching `@DataNeo4jTest` eventually as well. It could be improved to use Neo4j's dedicated test harness to create the embedded database in contrast as of today, where it uses OGM.

We started to detangle things by providing an interim starter for the [new generation of the Java driver](https://github.com/neo4j/neo4j-java-driver-spring-boot-starter), [Neo4jDriverProperties](https://github.com/neo4j/neo4j-java-driver-spring-boot-starter/blob/4f85158747efda60905823302b9083d99a3d50ed/neo4j-java-driver-spring-boot-autoconfigure/src/main/java/org/neo4j/driver/springframework/boot/autoconfigure/Neo4jDriverProperties.java) especially shows what we like to configure in the future.

[...]

PR #17653, opened 2019-07-26
GH-17610 - Provide dedicated Neo4j driver auto-configuration.

michael-simons:
This is a proposal for #17610. 

It consists of three major changes:

1. Adding the native driver to the managed dependencies
1. Adding the actual autoconfiguration
1. Adapting and extending the health contributor

The ideas behind 2. are implemented via `Neo4jDriverAutoConfiguration` and `Neo4jDriverProperties`. 

The main configurational parts are taken from the next-gen driver configuration (see https://github.com/neo4j/neo4j-java-driver-spring-boot-starter). 
Important things to look at:
* A driver bean is only then created when `spring.neo4j.uri` is actually set, we don't default here, as that would change behaviour for all users immediate
* The driver bean for the 1.7 driver is `Lazy`, as that version eagerly tries to open a connection

While all this is probably "standard" auto config, a bit of more thought was needed to adapt `Neo4jDataAutoConfiguration`.

1. It still reads `spring.data.neo4j` into a properties class
1. It still creates an OGM configuration bean (`org.neo4j.ogm.config.Configuration`)
1. It now *conditionally* creates an OGM driver bean (`org.neo4j.ogm.driver.Driver`) as well when a native driver bean `org.neo4j.driver.v1.Driver` is in the context. 
1. Depending on the outcome of 3., the creation of the OGM `SessionFactory` now either uses the OGM driver or falls back to the previous behaviour.

## Supported scenarios

Now the following scenarios are supported

### User doesn't change anything

* No driver bean is present in the context
* No OGM driver is created
* Initialization of drivers happens solely in the `SessionFactory` of OGM
* User is free to use `bolt`, `http`, `embedded` as fit

### Users configures the driver via the new properties

* Driver bean is present in the context
* OGM driver is created
* Shared config is applied
* `SessionFactory` is preconfigured

### Users adds a custom native driver bean

See above

### Users adds a custom OGM driver bean

* Shared config is applied
* `SessionFactory` is preconfigured

All of this is a huge advantage for users, especially the ability to add their own native driver bean, either through the new configuration or as it is recognised by the OGM configuration.

## Health

The third commit adapts the health support to the new scenarios. If at least one native Driver bean is present in the context, a native Neo4j health indicator is used, that employs an optimised query.

This happens under the assumption that when a driver bean is present, it is used to configure the session factory as well.

## Outlook

This change would add official support for things our users are currently doing manual and even more important, allows for switching to the next generation bolt driver at some point in the future, together with our revamped Spring Data Neo4j version.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/pom.xml
@@ -363,6 +363,11 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.neo4j.driver</groupId>
+			<artifactId>neo4j-java-driver</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-jdbc</artifactId>
 			<optional>true</optional>
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/neo4j/Neo4jHealthIndicatorAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/neo4j/Neo4jHealthIndicatorAutoConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.boot.actuate.autoconfigure.neo4j;
 
 import java.util.Map;
 
+import org.neo4j.driver.v1.Driver;
 import org.neo4j.ogm.session.SessionFactory;
 
 import org.springframework.boot.actuate.autoconfigure.health.CompositeHealthIndicatorConfiguration;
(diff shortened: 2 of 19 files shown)


snicoll:
Michael and I had a chat about this proposal and this led me to be believe we should not introduce an auto-configuration for the current neo4j driver. We know a major new version is coming with significant changes and I am not keen to provide a feature in 2.2 that we'll have to somehow break in 2.3. Given the timeline of things, the new driver should be available around our 2.2 RC, which makes it hard for us to integrate with it.

Michael already started a 3rd party starter that works against an alpha of the new driver. I did suggest he provide an alternative version of that starter with the current driver (basically the content of this PR). This would give neo4j users a chance to use this feature right away, potentially testing the other version that uses the new driver and provide feedback. As [suggested on the related issue](https://github.com/spring-projects/spring-boot/issues/17610#issuecomment-514223635), I'd like us to resume work on this when work for 2.3 starts.

michael-simons:
I'm fine with our discussion and the outcome. As much as I'd like to see having our users the possibilities described in the PR out of the box, better opt-in to something and then consciously upgrade to the next generation driver instead of being suprised by the upgrade through Boot itself.

Whether we will configure the `SessionFactory` for the next generation version has to be decided then by us, depending on OGM's support for the it.

PR #17660, opened 2019-07-28
Fix deprecation warnings in spring-boot-actuator

dreis2211:
Hi,

this PR fixes two deprecation warnings in `spring-boot-actuator` by replacing `Mono.compose(Function)` with `Mono.transformDeferred(Function)`.

Cheers,
Christoph


snicoll:
Good catch, thanks once again @dreis2211!

PR #17662, opened 2019-07-29
Missing TransactionManager when user provides a custom Neo4j SessionFactory.

michael-simons:
Thise commit pushes `@ConditionalOnMissingBean(SessionFactory.class)` from the main Neo4j autoconfiguration class down to the bean declaration level. In addition, the OGM configuration bean won’t be created when there’s already a `SessionFactory`.

Thus, the `Neo4jWebConfiguration` will still be applied, even when user decides to run their own `SessionFactory`.

This change is needed to provide the mechanism proposed in #17610 in an external starter, which would provide the `SessionFactory`. As we like to provide the starter for both Boot 2.2.0 and 2.1.x, I'd like to see this change in both branches if possible.

Thanks for your ongoing support in this case.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java
@@ -53,28 +53,14 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
  * @author Vince Bickers
  * @author Stephane Nicoll
  * @author Kazuki Shimizu
+ * @author Michael Simons
  * @since 1.4.0
  */
 @Configuration
 @ConditionalOnClass({ SessionFactory.class, Neo4jTransactionManager.class, PlatformTransactionManager.class })
-@ConditionalOnMissingBean(SessionFactory.class)
 @EnableConfigurationProperties(Neo4jProperties.class)
 @Import(Neo4jBookmarkManagementConfiguration.class)
 public class Neo4jDataAutoConfiguration {
-
-	@Bean
-	@ConditionalOnMissingBean
-	public org.neo4j.ogm.config.Configuration configuration(Neo4jProperties properties) {
-		return properties.createConfiguration();
-	}
-
-	@Bean
-	public SessionFactory sessionFactory(org.neo4j.ogm.config.Configuration configuration,
-			ApplicationContext applicationContext, ObjectProvider<EventListener> eventListeners) {
-		SessionFactory sessionFactory = new SessionFactory(configuration, getPackagesToScan(applicationContext));
-		eventListeners.stream().forEach(sessionFactory::register);
-		return sessionFactory;
-	}
 
 	@Bean
 	@ConditionalOnMissingBean(PlatformTransactionManager.class)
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java
```diff
@@ -53,22 +53,23 @@
...
 
 	@Bean
-	@ConditionalOnMissingBean
+	@ConditionalOnMissingBean({org.neo4j.ogm.config.Configuration.class, SessionFactory.class})
 	public org.neo4j.ogm.config.Configuration configuration(Neo4jProperties properties) {
 		return properties.createConfiguration();
 	}
 
 	@Bean
+	@ConditionalOnMissingBean(SessionFactory.class)
```
snicoll: How about moving those two bean definitions to a dedicated class? Can be an inner class here. That way we could move the condition on the `SessionFactory` there and limit the change to a minimum.

The name of that inner class would also convey the use case we're trying to bring here.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfigurationTests.java
```diff
@@ -94,6 +94,7 @@ public void customSessionFactory() {
 		this.contextRunner.withUserConfiguration(CustomSessionFactory.class).run((context) -> {
 			assertThat(context).doesNotHaveBean(org.neo4j.ogm.config.Configuration.class);
 			assertThat(context).hasSingleBean(SessionFactory.class);
+			assertThat(context).hasSingleBean(OpenSessionInViewInterceptor.class);
```
snicoll: Did you intend to commit this?

michael-simons: Yes. I first had a dedicated test `customSessionShouldNotDisableOSiV`, my interpretation of 

> (i.e. a test that provides a custom SessionFactory and validates the necessary bits are still auto-configured.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java
michael-simons: Nice idea. I did this, but had to move the private helper method `getPackagesToScan`, too.

michael-simons:
To make things explicit: I think not providing a `TransactionManger` and also not a `OpenSessionInViewInterceptor` is a bug. My PR fixes that.

This is the actual test for the desired behaviour and included in the PR

```
@Test
	public void customSessionFactoryShouldNotDisableOtherDefaults() {
		this.contextRunner.withUserConfiguration(CustomSessionFactory.class).run((context) -> {
			assertThat(context).hasSingleBean(Neo4jTransactionManager.class);
			assertThat(context).hasSingleBean(OpenSessionInViewInterceptor.class);
		});
	}
```

michael-simons:
Fantastic. Much appreciated.

michael-simons:
You mentioned that you want me to test this out with our starter. Test is already there: 
https://github.com/neo4j/neo4j-java-driver-spring-boot-starter/blob/1.7.x/neo4j-java-driver-spring-boot-autoconfigure/src/test/java/org/neo4j/driver/springframework/boot/autoconfigure/Neo4jDriverAutoConfigurationTest.java#L87

PR #17663, opened 2019-07-29
Generated metadata documentation does not escape the | character

vivganes:


vivganes:
@snicoll is there a place where i can see the final output of the docs created by the build server?  That way i can do a final check myself.

snicoll:
Our README has a section on how you can [generate the reference documentation](https://github.com/spring-projects/spring-boot/blob/master/README.adoc#building-reference-documentation).

vivganes:
[spring-boot-reference.pdf](https://github.com/spring-projects/spring-boot/files/3448269/spring-boot-reference.pdf)
PDF file created locally after the changes - This escapes table properly 
To find the related table entry, search for string "Comma-separated list of additional unencoded" in the attached PDF file.

snicoll:
@vivganes thank you very much for making your first contribution to Spring Boot.

Issue #16555, opened 2019-04-13
Use the Kotlin Coroutines BOM


sdeleuze:
When available (should be before Boot 2.2 GA), see https://youtrack.jetbrains.com/issue/KT-30042.

snicoll:
Thanks for the nudge. If there is a dedicated bom for this, it must only include coroutines and no other artifacts (as per our policy of including BOMs here).

sdeleuze:
Yes that the plan

mbhave:
Blocked on [this](https://github.com/Kotlin/kotlinx.coroutines/issues/1110).

sdeleuze:
The Coroutines BOM [is now available](https://repo1.maven.org/maven2/org/jetbrains/kotlinx/kotlinx-coroutines-bom/1.3.0-RC/kotlinx-coroutines-bom-1.3.0-RC.pom), could you ship it with Spring Boot 2.2.0.M5 ?

snicoll:
@sdeleuze can you please upgrade the framework build to use it?

sdeleuze:
@snicoll Yes I am trying to do so, but I am currently facing some issues related to broken classpath while my tests with the preview version were green, I will comment here asap it is ok.

sdeleuze:
@snicoll Done https://github.com/spring-projects/spring-framework/commit/c006a83584052fdd48a5ec4722d327ad6356aa94

philwebb:
Closing in favor of PR #17701

PR #17701, opened 2019-07-30
Add dependency management for Kotlin Coroutines BOM

sdeleuze:
This pull request adds the Coroutines BOM to Spring Boot dependencies as discussed in gh-16555, as well as the related smoke test. 

The Coroutines BOM is not strictly needed for Spring Boot itself since I have managed to make the support transparent at Boot infra level (no new bean introduced) and fully handled at Spring Framework level.

That said for end users, it is critical to have this Coroutines BOM included at Spring Boot dependencies level in order to:
 - Be sure developers leverage Coroutines 1.3+ (we do not support previous ones)
 -  Avoid mixing versions of Coroutines dependencies which in practice occurs very frequently (for example when developer specify only `kotlinx-coroutines-reactor` and get an older incompatible `kotlinx-coroutines-reactive` dependency)
 - Provide a consistent first class Coroutines support at both Spring Framework 5.2 and Spring Boot 2.2 levels.

In term of context, it is also important to notice that with Spring Framework 5.2 / Spring Boot 2.2, Coroutines are likely going to be the most popular way to consume WebFlux for Kotlin developers.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -51,6 +51,7 @@
 		<commons-lang3.version>3.9</commons-lang3.version>
 		<commons-pool.version>1.6</commons-pool.version>
 		<commons-pool2.version>2.6.2</commons-pool2.version>
+		<coroutines.version>1.3.0-RC</coroutines.version>
 		<couchbase-client.version>2.7.7</couchbase-client.version>
 		<couchbase-cache-client.version>2.1.0</couchbase-cache-client.version>
 		<derby.version>10.14.2.0</derby.version>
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot-dependencies/pom.xml
```diff
@@ -51,6 +51,7 @@
 		<commons-lang3.version>3.9</commons-lang3.version>
 		<commons-pool.version>1.6</commons-pool.version>
 		<commons-pool2.version>2.6.2</commons-pool2.version>
+		<coroutines.version>1.3.0-RC</coroutines.version>
```
snicoll: I am not sure it should be called like that. I'd prefer if there was something that colocates it with Kotlin (i.e. `kotlin-coroutines.version`)

spring-boot-project/spring-boot-dependencies/pom.xml
```diff
@@ -2595,6 +2596,13 @@
 				<scope>import</scope>
 				<type>pom</type>
 			</dependency>
+			<dependency>
```
snicoll: If you agree, this should move below the kotlin bom.

spring-boot-project/spring-boot-dependencies/pom.xml
sdeleuze: Yeah I am open to both.

PR #17736, opened 2019-07-30
Harmonize settings.gradle documentation in Gradle getting started guide

dsyer:
This works better (copied from start.spring.io). Otherwise there might be plugins that cannot be resolved.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/gradle/getting-started/snapshot-settings.gradle
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/gradle/getting-started/snapshot-settings.gradle
@@ -1,13 +1,13 @@
 pluginManagement {
-	repositories {
-		maven {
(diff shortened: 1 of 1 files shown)


snicoll:
@dsyer this confused me more than once but `libs-snapshot` contains more than just snapshots. For instance, you can see snapshots, milestones and releases of [the Spring Boot Gradle plugin](https://repo.spring.io/libs-snapshot/org/springframework/boot/spring-boot-gradle-plugin/). I guess the `gradlePluginPortal` is what that matters here. 

On the other hand, we should probably align to whatever start.spring.io does but I just want to make sure I am not missing some context.

dsyer:
I just copied from start.spring.io. If you want to change it there as well, fine by me. I don't think you are missing anything.

snicoll:
There are four potentially impacted files in this directory. Milestone and snapshots as well as a Groovy DSL and a Kotlin DSL version. Let's align them to whatever start.spring.io does.

PR #17797, opened 2019-08-06
 Add support for configuring Spring Session JDBC flush mode

vpavic:
This PR adds support for configuring Spring Session JDBC flush mode that was added in Spring Session `Corn-M3`.

I've also added another commit that harmonizes flush mode properties javadoc, similarly to the javadoc for save mode properties that was added in #17514.


snicoll:
Thanks for the follow-up @vpavic!

Issue #17798, opened 2019-08-06
Unable to use off for enum values in yaml file


mbhave:
See [this comment](https://github.com/spring-projects/spring-boot/issues/17385#issuecomment-518825344). [This commit](https://github.com/spring-projects/spring-boot/commit/07acc4af08a8001487d5196a617a02441245ae4f) relies on the `StringToEnumIgnoringCaseConverterFactory` to coerce false to off (YAML 1.1 maps off to false). But since the `sourceType` is `Boolean`, the converter returned by the `StringToEnumIgnoringCaseConverterFactory` doesn't get called.

Issue #17385, opened 2019-07-01
Consistently coerce false to off where applicable


wilkinsona:
YAML 1.1 maps `off` to `false`. We deal with this in `LoggingApplicationListener` by coercing `false` back to `LogLevel.OFF`:

https://github.com/spring-projects/spring-boot/blob/35a178abab02217d64c3731736dc6921c03334f1/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/logging/LoggingApplicationListener.java#L363-L369

This allows users to configure logging using `logging.level.example.my.package: off` in YAML.

We do no perform any coercion when turning the banner off so `spring.main.banner-mode: off` will fail and `spring.main.banner-mode: 'off'` has to be used instead.

I think we should consistently allow `false` to be bound to an enum that would accept `off`. Perhaps this is something that could be done in the binder rather than as a special case for banner mode.

wilkinsona:
There's some advice in the docs about turning the banner off that we can remove now.

romainmoreau:
Since Spring Boot 2.2.0.M5, I have to use `logging.level.example.my.package: 'off'` instead of `logging.level.example.my.package: off` which worked with Spring Boot 2.2.0.M4 but is now silently ignored with Spring Boot 2.2.0.M5.
When I read this issue, I understand that it should not be the case, should it?

mbhave:
@romainmoreau Thanks for trying out the milestone. It does look like this isn't working anymore. I've created a [new issue](https://github.com/spring-projects/spring-boot/issues/17798) for it since this was already closed.

freekry:
> Since Spring Boot 2.2.0.M5, I have to use `logging.level.example.my.package: 'off'` instead of `logging.level.example.my.package: off` which worked with Spring Boot 2.2.0.M4 but is now silently ignored with Spring Boot 2.2.0.M5.
> When I read this issue, I understand that it should not be the case, should it?

I tried and they(logging.level.example.my.package: 'off' ,logging.level.example.my.package: off) were all worked with Spring Boot 2.2.0.M5,can you show you demo?

PR #17823, opened 2019-08-09
bug fixed for Unable to use off for enum values in yaml file #17798

freekry:
bug fixed for Unable to use off for enum values in yaml file #17798 and the [comment](https://github.com/spring-projects/spring-boot/issues/17385#issuecomment-518825344)

I found that `off` and `false` resolved to `Boolean false` in yaml, but it can't found mapping logging lever，so i added some codes to make Boolean to String.I am not sure if the location of the code is appropriate, but it works fine.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/YamlPropertySourceLoader.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/YamlPropertySourceLoader.java
@@ -19,9 +19,9 @@ package org.springframework.boot.env;
 import java.io.IOException;
 import java.util.ArrayList;
 import java.util.Collections;
+import java.util.Iterator;
 import java.util.List;
 import java.util.Map;
-import java.util.Iterator;
 
 import org.springframework.core.env.PropertySource;
 import org.springframework.core.io.Resource;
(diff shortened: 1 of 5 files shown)


freekry:
checkstyle and spring-javaformat maked me crazy （︶︿︶）

mbhave:
Thanks for creating a PR @freekry. I think we'd like to fix this slightly differently. Specifically, we don't want to add `ON` to `LogLevel` and `YamlPropertySourceLoader` shouldn't need to know about the `logging.level` property. Thanks anyway.

freekry:
@mbhave I got you code and run it. when i set `false` or `off` it worked, but when i set `true` or `on` it throw Exception:`Caused by: java.lang.IllegalArgumentException: No enum constant org.springframework.boot.logging.LogLevel.true`

philwebb:
@freekry "true" and "on" aren't valid logging levels so that exception is correct.

freekry:
@philwebb  I am sorry.I am just confused by [this](https://github.com/spring-projects/spring-boot/commit/835108e522ca4a1969d35a96ecf731e63f554ea2):`Support binding of YAML style true/false values to 'ON'/'OFF'.Thank you for your reply.


`

freekry:
@philwebb I am sorry.I am just confused by [this](https://github.com/spring-projects/spring-boot/commit/835108e522ca4a1969d35a96ecf731e63f554ea2):`Support binding of YAML style true/false values to 'ON'/'OFF'.`

Thank you for your reply.



.Thank you for your reply.

PR #17835, opened 2019-08-11
Fix syntax errors in docs

dreis2211:
Hi,

this PR fixes two syntax errors in code examples inside the docs.

Cheers,
Christoph


snicoll:
Let's review the `containsString` change as we should probably use something else now anyway.

dreis2211:
Which `containsString`?

snicoll:
Ah! I am the one introducing the problem when I backported your change in 2.1.x 😑

Issue #17828, opened 2019-08-10
Extend RabbitProperties to include simplePublisherConfirms


pcoates33:
Add new property to allow `simplePublisherConfirms` to be set on the default `connectionFactory`.

```
spring.rabbitmq.simple-publisher-confirms=false # Whether to enable simple publisher confirms.
```

We already have properties for

```
spring.rabbitmq.publisher-confirms=false # Whether to enable publisher confirms.
spring.rabbitmq.publisher-returns=false # Whether to enable publisher returns.
```

https://docs.spring.io/spring-boot/docs/current/reference/html/common-application-properties.html

philwebb:
See #3945

snicoll:
Closing in favour of PR #17848

PR #17848, opened 2019-08-13
Add confirm-type property to RabbitProperties

htztomic:
This is a fix to GitHub issue #17828. Additional property simplePublisherConfirms was added to RabbitProperties.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
@@ -23,6 +23,7 @@ import java.util.List;
 
 import org.springframework.amqp.core.AcknowledgeMode;
 import org.springframework.amqp.rabbit.connection.CachingConnectionFactory.CacheMode;
+import org.springframework.amqp.rabbit.connection.CachingConnectionFactory.ConfirmType;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.DeprecatedConfigurationProperty;
 import org.springframework.boot.convert.DurationUnit;
@@ -95,6 +96,11 @@ public class RabbitProperties {
 	 * Whether to enable publisher returns.
 	 */
 	private boolean publisherReturns;
+
+	/**
+	 * The type of publisher confirms to use.
+	 */
+	private ConfirmType publisherConfirmType;
 
 	/**
 	 * Connection timeout. Set it to zero to wait forever.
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -96,6 +96,11 @@
 	 */
 	private boolean publisherReturns;
 
+	/**
+	 * Whether to enable simple publisher confirms.
+	 */
+	private boolean simplePublisherConfirms;
```
snicoll: @garyrussell could you please educate me on the difference between `publisherConfirms` and `simplePublisherConfirms`. Looking at those two properties, this feels a bit odd to me.

Perhaps we should refactor the configuration?

garyrussell: @snicoll With standard publisher confirms, the caller gets a callback for each sent message, together with correlation data so he can figure out which sent message the confirmation is for.

Let's say we send 10 messages back-to-back; the broker might send us 10 discrete confirmations or one confirmation that says "the last 10 sends were good", depending on timing.

In the latter case, the framework still invokes the callback 10 times.

With simple confirms (added in spring-amqp 2.1), the caller is saying "I don't care to be told which messages succeeded or failed, I just want to wait until they all succeed"; the amap-client has a method `waitForConfirmsOrDie` which is called by the `RabbitTemplate` method with the same name.

See [Scoped Operations](https://docs.spring.io/spring-amqp/reference/html/#scoped-operations) in the reference guide.

You can still use simple confirms without it, but setting this property removes the additional scaffolding needed to support standard confirms.

snicoll: @garyrussell and I had a little chat offline and I shared having two properties for the same feature was a bit odd. Gary said he'd add an enum in Spring AMQP instead so we'll have to rework this PR to use that.

Wondering if switching a property's type in a feature release is acceptable so flagging for team attention.

snicoll:
Blocked by https://github.com/spring-projects/spring-amqp/issues/1067

garyrussell:
New enum is in 2.2.0.BUILD-SNAPSHOT https://github.com/spring-projects/spring-amqp/commit/acb56833d1af567c8d1a839ca92c97f5a8fe7d33

snicoll:
@htztomic do you have time to update your proposal. Based on AMQP snapshots, the idea is to introduce a new `confirm-type` property with the enum type and deprecate `publisher-return` in favour of it while still reading the value (translating a boolean value to the relevant enum value behind the scenes)?

We don't rely on snapshots for AMQP yet but the plan is to do that soon as part of #17898

garyrussell:
One correction

>deprecate `publisher-return`

should be deprecate `publisher-confirms` (returns is still a boolean).

snicoll:
Thanks Gary, I've edited my comment to reduce the confusion.

snicoll:
This is blocked again as upgrading to the snapshots broke our build. I think there is a glitch in the fix on the AMQP side, see https://github.com/spring-projects/spring-amqp/issues/1067#issuecomment-522916369

PR #17861, opened 2019-08-14
Use alwaysApplyingWhenNonNull PropertyMapper to set the values in the LdapContextSource

filiphr:
The `userDn` and `password` in `LdapContextSource` are not nullable.
The default values for `userDn` and `password` in `LdapProperties` are `null.

When the values are set to null there will eventually be a `NullPointerException`
during `AbstractContextSource#setupAuthenticatedEnvironment` since `HashTable` doesn't allow null for values

I am not sure whether this should be fixed here in spring-boot, or perhaps directly within `LdapContextSource`.


philwebb:
Thanks for spotting this one, it looks like a legitimate bug on our side but also something that could be improved in spring-ldap. I've raised https://github.com/spring-projects/spring-ldap/issues/538 to see if they can make things more robust.

filiphr:
Thanks for raising the issue in spring-ldap. Once I started writing the fix for Spring Boot I also realized that making it more robust in spring-ldap might be the best thing. Something like that is already done in `setBaseName`.

Thanks for taking the fix into consideration.

Issue #12414, opened 2018-03-08
AbstractErrorWebExceptionHandler is unfriendly to extend/use


Clockwork-Muse:
The [spring feature documentation](https://github.com/spring-projects/spring-boot/blob/42629cb8ae8e574760b81e30af6c70e955770e67/spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc#error-handling-1) has the following example for extending `AbstractErrorWebExceptionHandler`:

>To change the error handling behavior, you can implement ErrorWebExceptionHandler and register a bean definition of that type. Because a WebExceptionHandler is quite low-level, Spring Boot also provides a convenient AbstractErrorWebExceptionHandler to let you handle errors in a WebFlux functional way, as shown in the following example:
```Java
public class CustomErrorWebExceptionHandler extends AbstractErrorWebExceptionHandler {

	// Define constructor here

	@Override
	protected RouterFunction<ServerResponse> getRoutingFunction(ErrorAttributes errorAttributes) {

		return RouterFunctions
				.route(aPredicate, aHandler)
				.andRoute(anotherPredicate, anotherHandler);
	}

}
```

... see that `// Define constructor here`?  Yeah.  It has to _start_ like this (all properties are required):
```Java
public ExtendingClass(ErrorAttributes errorAttributes, ResourceProperties resourceProperties, ApplicationContext applicationContext) {
    super(errorAttributes, resourceProperties, applicationContext);
}
```

Annoyingly, `ErrorAttributes` is for some reason not provided when using `@WebFluxTest`, meaning you have to mock it.

...and when those beans are provided, there's a hidden dependency - it has an `afterPropertiesSet` override, which is looking for a `List<HttpMessageWriter<?>>`, which is normally provided via `ServerCodecConfigurer`.
<hr />
Can we at least get this documented better?  The constructor, at minimum, needs to look like this:
```Java
public ExtendingClass(ErrorAttributes errorAttributes, ResourceProperties resourceProperties,
         ApplicationContext applicationContext, ServerCodecConfigurer configurer) {
    super(errorAttributes, resourceProperties, applicationContext);
    this.setMessageWriters(codec.getWriters());
}
```
... but I'd really prefer a friendlier way to instantiate this thing.
And should `ErrorAttributes` be added as a dependency for `@WebFluxTest`?

bclozel:
Putting aside the improvements on the `ErrorWebExceptionHandler` (I'm taking a look at this).

[...]

PR #17863, opened 2019-08-14
Make logError() in AbstractErrorWebExceptionHandler overrideable

peterjurkovic:
A subclass extending of `AbstractErrorWebExceptionHandler` can't override `logError()`. We would like to change logging behavior in a subclass, but unfortunately is not quite easy, we would have to duplicate a lot of useful code in that class. 

Can we please make `AbstractErrorWebExceptionHandler.logError(..)` `protected`? Unless there is a serious reason to make it private.


wilkinsona:
This is related to https://github.com/spring-projects/spring-boot/issues/12414.

wilkinsona:
Thank you very much for making your first contribution to Spring Boot, @peterjurkovic. I also pushed a small [polishing commit](https://github.com/spring-projects/spring-boot/commit/6328f2e8a88850c8c79d353c823057b6c7b9ed4c) to add some javadoc and restore the original method order so that they're ordered by use rather than visibility.

Issue #5314, opened 2016-03-02
Allow Jetty's thread pool to be configured via the environment


dmwb:
Hello,

I'm wondering, if it is possible to have some basic configuration for embedded Jetty in application.properties. Right now to customize thread pool and idle timeout, we're creating bean, which is not convenient...

```
    @Bean
    public JettyEmbeddedServletContainerFactory jettyEmbeddedServletContainerFactory() {
        final JettyEmbeddedServletContainerFactory factory = new JettyEmbeddedServletContainerFactory(serverProperties.getPort());
        factory.addServerCustomizers((Server server) -> {
            final QueuedThreadPool threadPool = server.getBean(QueuedThreadPool.class);
            threadPool.setMinThreads(appProperties.getMinThreads());
            threadPool.setMaxThreads(appProperties.getMaxThreads());
            threadPool.setIdleTimeout(appProperties.getIdleTimeout());
        });
        return factory;
    }
```

Could we have something similar like for tomcat and undertow. F.e.

```
server.jetty.max-threads
server.jetty.min-threads
server.jetty.idle-timeout
```

Or make it more generic. I'm pretty sure each servlet container has ability to configure such properties. 

Thanks!

wilkinsona:
See also #5324

wilkinsona:
We already have the following properties:

```
server.tomcat.max-threads
server.undertow.io-threads
server.undertow.worker-threads
```

Tomcat's `AbstractProtocol` also supports configuring the minimum number of spare threads for which we do not have a property.

Jetty uses a `QueueThreadPool` by default. It has three simple properties:

```
maxThreads (default 200)
minThreads (default 8)
idleTimeout (default 60000)
```

It also allows its `BlockQueue` to be configured. It defaults to a `BlockingArrayQueue` which is unbounded with an initial capacity of the bigger of 8 and the minimum number of threads.

dmwb:
Thanks in advance! Jetty is widely used and having these common parameters configured via application properties etc. will make Boot more efficient...

btridip-in:
Is there a way to specify the max threads via the configuration (.yml) file to control how many concurrent requests can be serviced by the instance at any point in time?

wilkinsona:
@btridip-in No, that's what this issue is about and it's still open. We have made it easier to configure it programmatically. See #5324 that was included in 1.4.

mbhave:
Closing in favor of PR #17871.

PR #17871, opened 2019-08-14
Additional server properties added for Jetty

htztomic:
This is a fix to GitHub issue #5314 which allows for configuration of Jetty's default thread pool, `QueuedThreadPool`,  via the environment. Updated server properties are to include maxThreads, minThreads, and idleTimeout.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -62,6 +62,7 @@ import org.springframework.util.unit.DataSize;
  * @author Andrew McGhie
  * @author Rafiullah Hamedy
  * @author Dirk Deyne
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -905,6 +906,21 @@ public void setEnabled(boolean enabled) {
 		 */
 		private Integer selectors = -1;
 
+		/**
+		 * Maximum amount of threads.
+		 */
+		private Integer maxThreads = 200;
```
ehabqadah: would not be better to set the maximum number of threads based on the number of available cores?

snicoll: The values here are the default of the embedded servers. We don't deviate from that unless necessary. If you want to pursue this idea, I'd raise an issue against the Jetty project.

PR #17872, opened 2019-08-15
Support multiple drives in CLI's spring script

wanderrful:
Hi there,

This PR is to fix an edge case for MinGW and Cygwin users where projects and source files are stored on different drives, causing the Spring Boot CLI to fail on startup. I searched through the Issue log and unfortunately didn't find anything about this so I haven't attached an issue number.

I _originally_ noticed that the Spring Boot CLI wouldn't run on my home PC due to a space in the `SPRING_HOME` folder path (i.e. `C:\Program Files\...`) for the Windows Batch executable script (which unfortunately is not included in the current branch so I've omitted that fix from this PR).

The error message was `Error: Could not find or load main class org.springframework.boot.loader.JarLauncher` and after researching a bit I narrowed it the issue down to two separate problems:

I found the problem to be that the `$CLASSPATH` variable was being parsed incorrectly as my Spring project was stored on a different hard drive, `F:\`, as opposed to my `$SPRING_HOME` folder, `C:\`. Because of this, the part of the Shell script where it assembles the $CLASSPATH variable via `cygpath` would get confused about what the root drive actually is and so I'd end up with a bunch of `/f/` prefixes to each $CLASSPATH folder, which in turn led to the above error message.

To fix this, I added two ternary checks to modify what exactly we're feeding into the `$SPRING_HOME` and `$CLASSPATH` variables before we actually attempt to run the `java -cp` command. Lastly, I also noticed a minor cosmetic consistency thing where `${}` was used in one place and `$` was used in another so I adjusted that to stay consistent. Lastly, the space in the folder path (i.e. `Program Files`) didn't help the situation so I added some double quotes ("") to take care of that edge case as well.

Of course, please test my changes for yourselves and if you think this change is worthy of a merge then I look forward to living the dream of one day seeing that purple icon on my notifications page.

Well, that's my Pull Request! Thanks for reading.

### DIFF
--- a/spring-boot-project/spring-boot-cli/src/main/executablecontent/bin/spring
+++ b/spring-boot-project/spring-boot-cli/src/main/executablecontent/bin/spring
@@ -18,7 +18,7 @@ case "$(uname)" in
 esac
 
 # For Cygwin, ensure paths are in UNIX format before anything is touched.
-if ${cygwin} ; then
+if $cygwin ; then
 	[ -n "${JAVA_HOME}" ] && JAVA_HOME=$(cygpath --unix "${JAVA_HOME}")
 fi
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@wanderrful Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=17872)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=17872) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@wanderrful Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=17872)!

wilkinsona:
@wanderrful Thank you very much for making your first contribution to Spring Boot. The proposed changes are now in master along with a small [polish commit](https://github.com/spring-projects/spring-boot/commit/308f2db93099cafc21695e4384aad0900e5c89d5). The polishing is to address an issue in a "normal" bash shell to avoid the following output:

```
./bin/spring: line 102: cygpath: command not found
./bin/spring: line 108: cygpath: command not found
```

If you have a chance to try a snapshot once [this build](https://ci.spring.io/teams/spring-boot/pipelines/spring-boot/jobs/build/builds/3279) has completed and confirm that the changes work in your environment that would be much appreciated.

PR #17874, opened 2019-08-15
Use hasSize() assertion in favor of length checks

dreis2211:
Hi,

this PR polishes code snippets like the following:

`assertThat(something.length).isEqualTo(1);`

to make use of shorter and arguably cleaner assertions like `hasSize()` etc.:

`assertThat(something).hasSize(1);`

Cheers,
Christoph


snicoll:
Thanks again @dreis2211

PR #17876, opened 2019-08-15
Test the Gradle Plugin against Gradle 5.6

dreis2211:
Hi,

Gradle 5.6 was released yesterday and we can test against it.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #17884, opened 2019-08-16
Simplify if statements

thelproad:
Simplify if statements.


snicoll:
@thelproad thank you for making your first contribution to Spring Boot.

PR #17914, opened 2019-08-19
simplify useless code

kioyong:
i found the logic "loader != null" is always true, so simplify useless code


snicoll:
@kioyong please don't create PRs of `master` on your fork but rather a local branch. If you update your copy of `master` on opened PRs this makes it harder to manage. 

This PRs has four commits including two merge commits while it should have one really. Can you please clean that up?

kioyong:
sorry for the lack of experience, i have create a new pull request, close this

PR #17916, opened 2019-08-20
Remove unnecessary null check

kioyong:
"loader != null" is always true, can simplify useless code


snicoll:
Thanks again @kioyong

Issue #17923, opened 2019-08-20
enhancement - allow .spring-boot-devtools.properties file to exist in any location per project


scottmf:
Problem - I want local configuration for my spring projects outside of my repository.  Using spring devtools to achieve this is cumbersome since I projects have overlapping configurationProperties (i.e. spring.*, server.*, etc..).  I want different values for different projects.
Issue with spring devtools - the .spring-boot-devtools.properties file is only allowed in my home directory
Solution - allow for the .spring-boot-devtools.properties file exist outside the home directory.  Allow it to be driven by an environment variable or system property.

philwebb:
Closing in favor of PR #17924

tedyoung:
I see that this issue was closed, but the PR referenced #17924 was never merged. It'd be nice to have this kind of support for local-to-my-machine config on a per project basis.

snicoll:
@tedyoung this issue is now superseded by the PR and it still open.

PR #17924, opened 2019-08-20
Use a configurable location to load global Devtools settings

scottmf:
This allows separate projects to keep their own settings where common settings
such as spring.* or server.* don't conflict.

https://github.com/spring-projects/spring-boot/issues/17923

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsHomePropertiesPostProcessor.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsHomePropertiesPostProcessor.java
@@ -44,7 +44,7 @@ public class DevToolsHomePropertiesPostProcessor implements EnvironmentPostProce
 	@Override
 	public void postProcessEnvironment(ConfigurableEnvironment environment, SpringApplication application) {
 		if (DevToolsEnablementDeducer.shouldEnable(Thread.currentThread())) {
-			File home = getHomeFolder();
+			File home = getProjectRootFolder();
 			File propertyFile = (home != null) ? new File(home, FILE_NAME) : null;
 			if (propertyFile != null && propertyFile.exists() && propertyFile.isFile()) {
 				FileSystemResource resource = new FileSystemResource(propertyFile);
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@scottmf Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=17924)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=17924) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@scottmf Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=17924)!

spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsHomePropertiesPostProcessor.java
```diff
@@ -61,6 +61,17 @@ public void postProcessEnvironment(ConfigurableEnvironment environment, SpringAp
 		}
 	}
 
+	protected File getProjectRootFolder() {
```
chenqimiao: It seems a breaking change.
When `PropertyFile` of `HomeFolder` have a property which not exist in `ProjectRootFolder`, after this change the property will be invalid.

sebasanjose (approved):

spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsHomePropertiesPostProcessor.java
wilkinsona: I don't think this is a breaking change as you have to opt in to looking in the new location by setting that system property or environment variable.

wilkinsona:
While not a breaking change, [this comment](https://github.com/spring-projects/spring-boot/pull/17924#discussion_r356950946) raises the question of whether we should use the user's home directory to create property sources with lower precedence than those created for the custom home directory. A benefit of this is that it would allow project-specific settings to be set in the custom location and for common settings to be set in the user's home directory. A downside is that it would then be impossible for a property in the user's home directory to be unset. I'm leaning towards keeping things simple and only using one location as in [this branch](https://github.com/wilkinsona/spring-boot/tree/gh-17924). Flagging for team attention to set what everyone else thinks.

wilkinsona:
We discussed this today and decided not to use the default global location as a lower-precedence property source when a custom global location is configured.

wilkinsona:
@scottmf Thanks very much for your patience and for making your first contribution to Spring Boot.

PR #17962, opened 2019-08-26
Polish

OLPMO:
Simplify the code and improve execution efficiency of the code with Java8

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationSorter.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationSorter.java
@@ -136,8 +136,7 @@ class AutoConfigurationSorter {
 		}
 
 		Set<String> getClassesRequestedAfter(String className) {
-			Set<String> classesRequestedAfter = new LinkedHashSet<>();
-			classesRequestedAfter.addAll(get(className).getAfter());
+			Set<String> classesRequestedAfter = new LinkedHashSet<>(get(className).getAfter());
 			this.classes.forEach((name, autoConfigurationClass) -> {
 				if (autoConfigurationClass.getBefore().contains(className)) {
 					classesRequestedAfter.add(name);
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationSorter.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationSorter.java
@@ -136,7 +136,8 @@ class AutoConfigurationSorter {
 		}
 
 		Set<String> getClassesRequestedAfter(String className) {
-			Set<String> classesRequestedAfter = new LinkedHashSet<>(get(className).getAfter());
+			Set<String> setOfClassesRequestedAfter = get(className).getAfter();
+			Set<String> classesRequestedAfter = new LinkedHashSet<>(setOfClassesRequestedAfter);
 			this.classes.forEach((name, autoConfigurationClass) -> {
 				if (autoConfigurationClass.getBefore().contains(className)) {
 					classesRequestedAfter.add(name);
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ImportAutoConfigurationImportSelector.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ImportAutoConfigurationImportSelector.java
@@ -58,8 +58,7 @@ class ImportAutoConfigurationImportSelector extends AutoConfigurationImportSelec
 
 	@Override
 	public Set<Object> determineImports(AnnotationMetadata metadata) {
-		Set<String> result = new LinkedHashSet<>();
-		result.addAll(getCandidateConfigurations(metadata, null));
+		Set<String> result = new LinkedHashSet<>(getCandidateConfigurations(metadata, null));
 		result.removeAll(getExclusions(metadata, null));
 		return Collections.unmodifiableSet(result);
 	}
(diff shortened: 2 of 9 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/AutoConfigurationSorterTests.java
```diff
@@ -17,11 +17,7 @@
 package org.springframework.boot.autoconfigure;
 
 import java.io.IOException;
-import java.util.Arrays;
-import java.util.LinkedHashSet;
-import java.util.List;
-import java.util.Properties;
-import java.util.Set;
+import java.util.*;
```
snicoll: We don't use wildcard imports. Please revert this change

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java
```diff
@@ -125,10 +125,7 @@ private boolean isEnabled() {
 		if (mode == DataSourceInitializationMode.NEVER) {
 			return false;
 		}
-		if (mode == DataSourceInitializationMode.EMBEDDED && !isEmbedded()) {
-			return false;
-		}
-		return true;
+		return mode != DataSourceInitializationMode.EMBEDDED || isEmbedded();
```
snicoll: When there are multiple if statements like this, I find the current version more readable than collapsing the last if statement as you suggest.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationSorter.java
```diff
@@ -136,8 +136,7 @@ AutoConfigurationClass get(String className) {
 		}
 
 		Set<String> getClassesRequestedAfter(String className) {
-			Set<String> classesRequestedAfter = new LinkedHashSet<>();
-			classesRequestedAfter.addAll(get(className).getAfter());
+			Set<String> classesRequestedAfter = new LinkedHashSet<>(get(className).getAfter());
```
snicoll: I find personally the previous version more readable: nesting the method call in the constructor parameter makes it harder to read.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ImportAutoConfigurationImportSelector.java
```diff
@@ -58,8 +58,7 @@
 
 	@Override
 	public Set<Object> determineImports(AnnotationMetadata metadata) {
-		Set<String> result = new LinkedHashSet<>();
-		result.addAll(getCandidateConfigurations(metadata, null));
+		Set<String> result = new LinkedHashSet<>(getCandidateConfigurations(metadata, null));
```
snicoll: I find personally the previous version more readable: nesting the method call in the constructor parameter makes it harder to read.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jmx/ParentAwareNamingStrategy.java
```diff
@@ -57,8 +57,7 @@ public void setEnsureUniqueRuntimeObjectNames(boolean ensureUniqueRuntimeObjectN
 	@Override
 	public ObjectName getObjectName(Object managedBean, String beanKey) throws MalformedObjectNameException {
 		ObjectName name = super.getObjectName(managedBean, beanKey);
-		Hashtable<String, String> properties = new Hashtable<>();
-		properties.putAll(name.getKeyPropertyList());
+		Hashtable<String, String> properties = new Hashtable<>(name.getKeyPropertyList());
```
snicoll: I find personally the previous version more readable: nesting the method call in the constructor parameter makes it harder to read.

OLPMO: I think the previous version of the code will cause the resize of the hash table, so i commit these change.Is my thought something wrong?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/AutoConfigurationSorterTests.java
OLPMO: oh,sorry!It is my carelessness.I will correct it immediately.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ImportAutoConfigurationImportSelector.java
OLPMO: Create the hashset with default size ,and then put all the item in it.Would not it cause the hash table resize?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationSorter.java
OLPMO: Create the hashset with default size ,and then put all the item in it.Would not it cause the hash table resize?

OLPMO:
Is there anything else I need to improve on in my pr? @snicoll

snicoll:
@OLPMO thanks for the follow-up. I did asked for feedback and @mbhave shared an interesting idea. We could keep your change with the same readability by introducing a variable prior to creating the collection. Can you please update your PR in that direction?

OLPMO:
@snicoll I have submitted my code in order to increase the readability of the code.It seems like that my submitted code did not pass ci.However, my local code is able to build successfully.Is there anything wrong with my code?

snicoll:
@OLPMO thank you for making your first contribution to Spring Boot and following-up so quickly on the feedback to your proposal.

To answer your previous question, nothing wrong with the code, the CI failure is unrelated. Thanks again.

OLPMO:
@snicoll  You‘re welcome

PR #17968, opened 2019-08-27
Fix Undertow access log dir property name in doc

polarbear567:


snicoll:
@polarbear567 thank you for making your first contribution to Spring Boot.

PR #17970, opened 2019-08-27
Fix a broken Asciidoctor syntax

izeye:
This PR fixes a broken Asciidoctor syntax.


snicoll:
Thanks again @izeye. I used the opportunity to wrap content at 90 chars as the commit that introduced that broken syntax didn't.

PR #17976, opened 2019-08-27
Spring Boot CLI is unable to always capture ctrl-c

nebhale:
Previously, when the Spring Boot CLI ran the Java command line for an application it did a straight invocation which suppressed propagation of signals to the JVM in certain circumstances (e.g. within Docker containers).

This change prepends the command with `exec` which causes the Java process to replace the script process in a shell ensuring that signals are propagated to the Java process without suppression.


wilkinsona:
Thanks very much, @nebhale. The proposed change has been merged into 2.1.x and forwards into master.

Issue #17974, opened 2019-08-27
TypeUtils.process(TypeDescriptor, TypeMirror) swallows exception message and stack trace


wilkinsona:
Failures in `TypeUtils.process(TypeDescriptor, TypeMirror)` are very hard to diagnose as the exception message and stack trace is swallowed. See https://github.com/spring-projects/spring-boot/issues/17650 for an example of a problem that was hard to diagnose due to this.

Sauhardstark:
@wilkinsona, I am a novice so I might be completely wrong but why not add the exception message, name and the stacktrace to the already existing printMessage() method

`this.env.getMessager().printMessage(Kind.WARNING, "Failed to generated type descriptor for " + type + " The exception information could go here", this.types.asElement(type));`

Thank you

wilkinsona:
Thanks for the suggestion, @Sauhardstark.

Including more information in the printed message is certainly one option. The message from the caught exception could probably be included in a way that would work in all environments where the annotation processor is used but I doubt that the stack trace could.

As part of looking at this issue, I'd also like us to take a step back and consider if the exception should be caught at all. Catching an unchecked exception that's thrown by Eclipse's JDT due to (we believe) a bug may not be the right thing to do. It may be better to not catch the exception at all and allow the annotation processing to fail. That should result in the exception's stack trace being appearing in the compiler's output in Maven and Gradle and in the error log in Eclipse. This assumes that there aren't other exception that may be thrown by the same code that we do want to catch. I don't yet know if that's a reasonable assumption to make.

snicoll:
> As part of looking at this issue, I'd also like us to take a step back and consider if the exception should be caught at all.

If memory serves well, the catch was added as we decided to implement [a feature in a maintenance release](https://github.com/spring-projects/spring-boot/issues/15850). Playing with generics and different compilers lead to some surprising things for me in the past so I was felling nervous hoping for the best to be honest. 

Having said that, I am not against removing the catch in `2.2` and see how it goes if we think this is the right option moving forward.

PR #17977, opened 2019-08-28
add warn log to trace the TypeUtils.process(TypeDescriptor, TypeMirro…

polarbear567:
…r) exception stack for issue 17974


wilkinsona:
Thanks for the proposal. Unfortunately, an annotation processor is quite a constrained environment and I'm not sure that adding logging will improve its failure diagnostics. For example, I don't think there's any guarantee that what's logged will be shown to the user. I've labelled #17974 as pending design work to indicate that there's some design work to do before we can decide how it should be fixed. Sorry for not doing that sooner and the wasted effort here.

PR #17984, opened 2019-08-28
Test the Gradle Plugin against Gradle 5.6.1

dreis2211:
Hi,

Gradle 5.6.1 was released earlier today and we can test against it.

Cheers,
Christoph


dreis2211:
Tests seem to fail. I'm taking a look...

wilkinsona:
Thanks, @dreis2211. It looks to me like the failures are due to the upgrade to Kotlin 1.3.50 (https://github.com/spring-projects/spring-boot/commit/078d53a5f2504cc241d2e164ab9f8329ff250616).

dreis2211:
@wilkinsona Thought the same, but wanted to make sure it has nothing to do with my changes. It seems to be indeed the upgrade to 1.3.50, though not sure yet what exactly causes the error.

P.S.: The current builds and/or Concourse itself seems to be unstable/unreachable/unresponsive. I saw the other day that 5.4.1 was released, which fixes a memory leak. I didn't know where to address this - maybe it's worth a look.

Issue #17695, opened 2019-07-29
WebClientExchangeTags does not handle non-standard status codes


druppelt:
Currently when calling a service that returns non-standard status codes like 490 with the reactive webclient, the tagging for the metrics actuator will cause an exception.
I'm using version 2.1.5.RELEASE of spring-boot-starter-actuator and spring-boot-starter-webflux. If I find the time, I'll create a pull request. For which branch should I do that?

In case I don't find the time, here's what I found so far:
The issue is that `HttpStatus` is used instead of a raw int. `org.springframework.boot.actuate.metrics.web.reactive.client.WebClientExchangeTags#status(ClientResponse)` itself only needs the raw value, but calls `response.statusCode()` which calls `HttpStatus.valueOf(getRawStatusCode())` which throws an `IllegalArgumentException`.
At this point the solution is easy, just use `response.rawStatusCode()`.

But after that it fails in `WebClientExchangeTags#outcome(ClientResponse)`, and here I'm not sure what an elegant solution would be. Either directly use the rawStatusCode and ditch the comfort of the `HttpStatus` enum, or add an option to create `HttpStatus.UNKNOWN` values with the rawStatusCode as constructor argument I guess. I prefer the latter solution, but it may have side effects, as the Enum is part of spring-web and thus used a lot (681 usages according to my IDE).

snicoll:
@druppelt support of non-standard status code is a Spring Framework concern so I am wondering if we shouldn't move this one in `spring-framework`. Let's ask @poutsma first.

rstoyanchev:
@druppelt can you clarify what the 490 is for this service? One option is to add more enum values if they are common enough.

druppelt:
@rstoyanchev In my case 490 and 491 are API specific validation issues, so definitly not common. 

From my understanding everything matching ^[1-5]\d{2}$ should be allowed as status code, though unknown codes may be mapped to x00, see https://greenbytes.de/tech/webdav/draft-ietf-httpbis-p2-semantics-25.html#rfc.section.6.p.2

For me it's totally fine if spring doesn't support non standard codes ootb, but there should be hooks available to add custom status codes. And something like metrics imho should try it's best to not cause exceptions.

wilkinsona:
I don't think this should move to Framework as there are fixes to be made in Boot. We can work with the raw status from the `ClientResponse` and also use `HttpStatus.Series` when determining the outcome.

PR #17991, opened 2019-08-29
RestTemplateExchangeTags does not handle non-standard status codes

izeye:
This PR applies changes similar to gh-17695 but for `RestTemplate`.


wilkinsona:
I thought about the `RestTemplate` case when working on #17695 and then completely forgot about it. Thanks for stopping this half from falling through the cracks, @izeye.

wilkinsona:
Thanks again, @izeye. The proposed changes have been merged into master.

PR #18015, opened 2019-08-30
Remove reference to non-existent --edit option in CLI docs

dreis2211:
Hi,

I just noticed that the docs mention `--edit` as an option to the CLI command `spring run`. This option is long gone as it seems. See #611

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

Issue #6313, opened 2016-07-04
Use runnable code to generate documentation samples


philwebb:
We should look at using running sample code imported into our `.adoc` files to prevent issues like #6312. Spring Security has done this already.

snicoll:
That and actually testing them would be nice to ensure backward compatibility.

wilkinsona:
I've done it in REST Docs too. It's not too arduous.

wilkinsona:
I've made a (tiny) start on this. See in ccaa19d.

philwebb:
Marking this one as complete but we might need to do the same with Gradle/Maven docs.

PR #18017, opened 2019-08-30
Add kotlin code snippets to spring-boot refdoc

IEnoobong:
Add kotlin code snippets to spring-boot refdoc

See spring-projectsgh-21778


spring-boot-project/spring-boot-docs/pom.xml
```diff
@@ -1440,6 +1440,13 @@
 					<plugin>
 						<groupId>org.asciidoctor</groupId>
 						<artifactId>asciidoctor-maven-plugin</artifactId>
+						<dependencies>
+							<dependency>
+								<groupId>io.spring.asciidoctor</groupId>
+								<artifactId>spring-asciidoctor-extensions</artifactId>
+								<version>0.1.3.RELEASE</version>
```
snicoll: FTR this would break our build as this is not on Maven Central and we don't build against `repos.spring.io`.

sdeleuze:
FYI this PR is work in progress, more commits will come for other asciidoc files and I plan to review them.

spring-boot-project/spring-boot-docs/pom.xml
IEnoobong: Ok, I'll add`repos.spring.io` as a `<repository>`

IEnoobong: Done.

wilkinsona: It should be a `<pluginRepository>` as we've done in the Gradle plugin:

https://github.com/spring-projects/spring-boot/blob/9e640cf6efebf3efe662720da70981e93d670280/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/pom.xml#L184-L190

IEnoobong: I see, Thanks, Will revert

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -449,9 +477,25 @@ The following example shows a `@Service` Bean that uses constructor injection to
 	}
 ----
 
+[source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
+.Kotlin
+----
+    package com.example.service;
+
+	import org.springframework.beans.factory.annotation.Autowired;
```
sdeleuze: Wrong indent?

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
```diff
@@ -467,6 +511,15 @@ If a bean has one constructor, you can omit the `@Autowired`, as shown in the fo
 	}
 ----
 
+[source,kotlin,indent=0,subs="verbatim,quotes",role="secondary"]
+.Kotlin
+----
+    @Service
+	class DatabaseAccountService(private val riskAssessor: RiskAssessor): AccountService {
```
sdeleuze: Wrong indent?

wilkinsona:
Thanks for the PR, @IEnoobong. Before too much more time is invested in this, I think we should pause and consider https://github.com/spring-projects/spring-boot/issues/6313. That issue is already quite a big job and inlining Kotlin snippets in the documentation will make it bigger.

spring-boot-project/spring-boot-docs/src/main/asciidoc/using-spring-boot.adoc
IEnoobong: I doubt. Used `tab` tho

sdeleuze:
Hey @wilkinsona, would you be ok if we provide Kotlin code samples as runnable code directly?

I think providing a consistent experience across Boot and Framework reference documentation would bring a lot of value to developers  (see how it looks like on [here](https://docs.spring.io/spring-framework/docs/5.2.0.BUILD-SNAPSHOT/spring-framework-reference/index.html) where core, testing, WebFlux and WebMvc documentations are translated).

wilkinsona:
That sounds good to me, assuming that we don't run into Eclipse-related problems with the Kotlin source. We've been fine with the Kotlin code that we've got thus far so I don't anticipate any problems.

A nice side-effect would be that it might give us the necessary impetus to do the same for our Java examples.

spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/service/KotlinDatabaseAccountService.kt
```diff
@@ -0,0 +1,18 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired
+import org.springframework.stereotype.Service
+
+/**
+ * * Sample showing autowiring the [org.springframework.beans.factory.annotation.Autowired] annotation
+ */
+// tag::customizer[]
+@Service
+class KotlinDatabaseAccountService(@Autowired private val riskAssessor: RiskAssessor) : AccountService {
```
sdeleuze: The be equivalent to the Java version, `@Autowired` should be on the constructor:
`class KotlinDatabaseAccountService @Autowired contructor(private val riskAssessor: RiskAssessor) : AccountService {`

spring-boot-project/spring-boot-docs/pom.xml
```diff
@@ -14,6 +14,7 @@
 	<properties>
 		<main.basedir>${basedir}/../..</main.basedir>
 		<refdocs.build.directory>${project.build.directory}/refdocs/</refdocs.build.directory>
+		<kotlin.version>1.3.50</kotlin.version>
```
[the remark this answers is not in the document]
IEnoobong: True. Reverted

IEnoobong:
> Please use immutable classes with `val` properties since they are now supported by Spring Boot 2.2 like:
> 
> ```
> data class BlogProperties(val title: String, val banner: Banner) {
>   data class Banner(val title: String?, val content: String)
> }
> ```

Yes, there's an example on immutable binding on line 898

spring-boot-project/spring-boot-docs/src/main/asciidoc/spring-boot-features.adoc
```diff
@@ -859,18 +921,23 @@ The example in the previous section can be rewritten in an immutable fashion as
 
 		private final InetAddress remoteAddress;
 
+		private final int port;
+
 		private final Security security;
 
-		public AcmeProperties(boolean enabled, InetAddress remoteAddress, Security security) {
+		public AcmeProperties(boolean enabled, InetAddress remoteAddress, int port, Security security) {
```
IEnoobong: added port to enable instantiation of the server in https://github.com/spring-projects/spring-boot/pull/18017/files#diff-51b04ca54a1bd6f28cae0525b2b1a7fdR23

IEnoobong:
> For configuration properties I make the objects `val` but the properties `var`
> 
> ```kotlin
> class ImmutableAcmeProperties(
> 	var enabled: Boolean,
> 	var remoteAddress: InetAddress,
> 	var port: Int,
> 	val security = Security()
> ) {
> 
> 	class Security(
> 		var username: String?=null,
> 		var password: String?=null,
> 		@DefaultValue("USER")
> 		val roles: List<String> = mutableList()
> 	)
> }
> ```
> 
> Aren't there runnable tests for the Java equivalent?
 works with `val` too https://github.com/spring-projects/spring-boot/pull/18017#commitcomment-35075209

No there are not.

sdeleuze:
It will be too short for Spring Boot 2.2, I propose we target Spring Boot 2.3 for this PR instead (if possible beginning of the dev cycle in order to avoid too much merge conflicts). Maybe that will be also the opportunity to move forward on #6313 for Java as well in Spring Boot 2.3 timeframe if Boot team think it could be in the roadmap for next year.

snicoll:
A [pre-requisite of this PR](https://github.com/spring-projects/spring-boot/pull/18017#issuecomment-526593835) was that our code samples move first to code snippet so that we figure out what to do to provide a Kotlin counterpart. This PR has been sitting for a while and we'd like to to address the first part, see #6313. 

Thanks for the PR and sorry for the wasted time.

PR #18077, opened 2019-08-31
Simplify some code

freekry:
Simplify some code.


snicoll:
@freekry thanks again!

PR #18100, opened 2019-09-03
Fix typo in deprecation reason for liquibase.check-change-log-location property

izeye:
This PR fixes a typo.


wilkinsona:
Love changes like this. Thanks once again, @izeye.

PR #18122, opened 2019-09-04
Detect JDK 13 updates

dreis2211:
Hi,

I saw that JDK13 wasn't included yet in the `detect-jdk-updates` CI task. With JDK13 coming in 2-3 weeks every wednesday the task will propose to use the latest nightly - that should be manageable. Even if the created GitHub issue is not tackled, it sticks around and acts as a reminder. So overall, I think it's a good preparation step and as soon as JDK13 goes GA we can simply switch from `nightly` to `releases` in the API url.

It's kind of hard to test for me at the moment, so forgive me if it doesn't work.

Let me know what you think.
Cheers,
Christoph

P.S.: If this is considered to be worthwhile and we're fast enough, we might be included in this week's `detect-jdk-updates` build. 😉

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -213,6 +213,15 @@ jobs:
         GITHUB_USERNAME: ((github-username))
         JDK_VERSION: java12
       image: spring-boot-ci-image
+    - task: detect-jdk13-update
+      file: git-repo/ci/tasks/detect-jdk-updates.yml
+      params:
+        GITHUB_REPO: spring-boot
+        GITHUB_ORGANIZATION: spring-projects
+        GITHUB_PASSWORD: ((github-password))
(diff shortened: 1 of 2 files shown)


snicoll:
Sounds good, Christoph :)

dreis2211:
@snicoll Thanks! Any chance we could trigger the job to have this battle tested?

snicoll:
[You bet](https://ci.spring.io/teams/spring-boot/pipelines/spring-boot/jobs/detect-jdk-updates/builds/41)

snicoll:
I forgot to update the pipeline. Doing that now.

dreis2211:
I can only see it's green, but shouldn't there be a new issue created?

Issue #18140, opened 2019-09-05
logging: provide ability to customize the RollingFile pattern


jorgheymans:
In `file-appender.xml` I would like to disable just the gzip so that my log files stay more easily readable. It would be nice to have something like `${LOG_ROLLING_PATTERN}` where i could redefine just this `<fileNamePattern>${LOG_FILE}.%d{yyyy-MM-dd}.%i.gz</fileNamePattern>`

```
	<appender name="FILE"
		class="ch.qos.logback.core.rolling.RollingFileAppender">
		<encoder>
			<pattern>${FILE_LOG_PATTERN}</pattern>
		</encoder>
		<file>${LOG_FILE}</file>
		<rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
			<fileNamePattern>${LOG_FILE}.%d{yyyy-MM-dd}.%i.gz</fileNamePattern>
			<maxFileSize>${LOG_FILE_MAX_SIZE:-10MB}</maxFileSize>
			<maxHistory>${LOG_FILE_MAX_HISTORY:-0}</maxHistory>
		</rollingPolicy>
	</appender>
```

eddumelendez:
due to #18151 was merged I think this issue can be closed

wilkinsona:
Indeed it can. Thank you, @eddumelendez.

PR #18151, opened 2019-09-06
Add property ROLLING_FILE_LOG_PATTERN for File Appender

eddumelendez:
This new property allows to customize `fileNamePattern` if it is set.
Otherwise, a default pattern remains. Also, new property is supported
`logging.pattern.rolling-file`.

See gh-18140

### DIFF
--- a/spring-boot-project/spring-boot/src/main/resources/org/springframework/boot/logging/logback/file-appender.xml
+++ b/spring-boot-project/spring-boot/src/main/resources/org/springframework/boot/logging/logback/file-appender.xml
@@ -14,7 +14,7 @@ initialization performed by Boot
 		<file>${LOG_FILE}</file>
 		<rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
 			<cleanHistoryOnStart>${LOG_FILE_CLEAN_HISTORY_ON_START:-false}</cleanHistoryOnStart>
-			<fileNamePattern>${LOG_FILE}.%d{yyyy-MM-dd}.%i.gz</fileNamePattern>
+			<fileNamePattern>${ROLLING_FILE_NAME_PATTERN:-${LOG_FILE}.%d{yyyy-MM-dd}.%i.gz}</fileNamePattern>
 			<maxFileSize>${LOG_FILE_MAX_SIZE:-10MB}</maxFileSize>
 			<maxHistory>${LOG_FILE_MAX_HISTORY:-7}</maxHistory>
 			<totalSizeCap>${LOG_FILE_TOTAL_SIZE_CAP:-0}</totalSizeCap>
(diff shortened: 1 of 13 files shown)


spring-boot-project/spring-boot/src/main/resources/org/springframework/boot/logging/logback/file-appender.xml
```diff
@@ -14,7 +14,7 @@ initialization performed by Boot
 		<file>${LOG_FILE}</file>
 		<rollingPolicy class="ch.qos.logback.core.rolling.SizeAndTimeBasedRollingPolicy">
 			<cleanHistoryOnStart>${LOG_FILE_CLEAN_HISTORY_ON_START:-false}</cleanHistoryOnStart>
-			<fileNamePattern>${LOG_FILE}.%d{yyyy-MM-dd}.%i.gz</fileNamePattern>
+			<fileNamePattern>${LOG_FILENAME_PATTERN:-${LOG_FILE}.%d{yyyy-MM-dd}.%i.gz}</fileNamePattern>
```
jorgheymans: `LOG_FILENAME_PATTERN` is not a bad name but i wonder if it should somehow have the `ROLLING` notion in there since it's only used when rolling over log files according to the RollingPolicy ...  `ROLLING_FILENAME_PATTERN` perhaps?

@eddumelendez thanks for the quick response !

wilkinsona: We already have `FILE_LOG_PATTERN` that's mapped from `logging.pattern.file` so I agree that the system property needs to be more differentiated. How about `ROLLING_FILE_LOG_PATTERN`? We could then perhaps also introduce a new property named something like `logging.pattern.rolling-file` that maps to the new system property.

jorgheymans:
Should this also be added in the documentation somewhere ? (https://docs.spring.io/spring-boot/docs/current/reference/html/howto-logging.html)

spring-boot-project/spring-boot/src/main/resources/org/springframework/boot/logging/logback/file-appender.xml
eddumelendez: Thanks guys I haven taken this comments into account.

eddumelendez:
PR has been updated. Thanks @jorgheymans and @wilkinsona

snicoll:
I am wondering if the property shouldn't be `logging.pattern.rolling-file-name` as we have a `logging.pattern.file` that provides a pattern for the log content, not the name of the file.

Rather than `ROLLING_FILE_LOG_PATTERN` it could be `ROLLING_FILE_NAME_PATTERN`.

eddumelendez:
@snicoll should I apply the changes suggested?

snicoll:
If you have time shortly, feel free to do it. I intended to polish your PR sometime Monday. Either is fine, thanks for asking!

snicoll:
Thanks @eddumelendez

PR #18170, opened 2019-09-06
Remove management.health.status.order from docs

dreis2211:
Hi,

while working on another PR I noticed that `management.health.status.order` is still showing in the [Actuator Properties](https://docs.spring.io/spring-boot/docs/2.2.0.BUILD-SNAPSHOT/reference/htmlsingle/#actuator-properties) section of the docs, although it should be deprecated.

This PR also fixes `management.endpoint.health.status.http-mapping.*` not showing up at all as a side-effect.

Let me know what you think. I found it quite "unintuitive" - maybe that's the word I'm looking for - to work with the metadata stuff and I'm not sure if this should be solved differently or if there is a convention for the metadata.json files.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -29,6 +29,15 @@
       "defaultValue": "never"
     },
     {
+      "name": "management.endpoint.health.status.order",
+      "defaultValue": [
+        "DOWN",
+        "OUT_OF_SERVICE",
+        "UP",
+        "UNKNOWN"
+      ]
+    },
+    {
       "name": "management.endpoints.enabled-by-default",
       "type": "java.lang.Boolean",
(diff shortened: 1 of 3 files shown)


dreis2211:
@philwebb I just noticed another slightly related doc error related to this. Give me a minute to fix this.

philwebb:
Sure thing! I'm just triaging issues at the moment, not merging.

dreis2211:
Done.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -162,7 +171,10 @@
         "OUT_OF_SERVICE",
         "UP",
         "UNKNOWN"
-      ]
+      ],
```
snicoll: This shouldn't be required as the annotation processor is supposed to pick things up. I've opened #18184

snicoll: The issue is fixed so I'll rebase and polish this part.

PR #18166, opened 2019-09-06
Simplify code

gaohanghang:
1. Use `Math.max()` and lambda expression simplify some code.
2. Use `Long.parseLong()` replace ` Long.valueOf()`

PR #18175, opened 2019-09-07
Simplify code

gaohanghang:


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyReactiveWebServerFactory.java
```diff
@@ -167,7 +167,7 @@ protected JettyResourceFactory getResourceFactory() {
 	}
 
 	protected Server createJettyServer(JettyHttpHandlerAdapter servlet) {
-		int port = (getPort() >= 0) ? getPort() : 0;
+		int port = Math.max(getPort(), 0);
```
snicoll: I think the intention of the current code is clearer. We want to set the port to `0` if it's null or negative. My brain has to parse the `.max` so I am not keen to merge this change.

wilkinsona: I prefer the proposed change. I find it no harder to read than what we currently have, with the advantage that `getPort()` is only called once so it doesn't make me pause and wonder if the value could change between calls.

philwebb: I'm happy with the change as well

snicoll: Thanks both. I am not feeling strongly against it so I'll go ahead and process with the merge.

snicoll:
@gaohanghang thank you for making your first contribution to Spring Boot.

gaohanghang:
> I've seen you closed #18166. There is no need to create another PR when changes are requested, just push additional commits on your branch and this will update the PR.

👌, Thanks

PR #18196, opened 2019-09-10
Update Quartz smoke test to use JDBC job store

vpavic:
This PR updates Quartz smoke test to use JDBC job store backed by an embedded H2 database. This is more representative of a typical Quartz usage and therefore better suited to catch relevant issues.


snicoll:
Good idea, thanks again @vpavic

Issue #18194, opened 2019-09-10
HealthIndicatorRegistry bean is no longer available


krm1312:
The (standard I assume) solution to expose the health endpoint to micrometer does not seem to work in 2.2.0.M6.  A lot of the classes are deprecated (as indicated in release notes), but, it fails to start.

http://micrometer.io/docs/guide/healthAsGauge

```

***************************
APPLICATION FAILED TO START
***************************

Description:

Parameter 1 of constructor in com.acme.app.metrics.HealthMetricsConfig required a bean of type 'org.springframework.boot.actuate.health.HealthIndicatorRegistry' that could not be found.
```

Not sure if you want to make it work as is (with deprecated classes/beans) or update micrometer recipe.

krm1312:
Actually, it looks like maybe I changed it for a previous upgrade a bit.  This is what we currently have:

```
@Configuration
public class HealthMetricsConfig {

    private CompositeHealthIndicator healthIndicator;

    /**
     * Configure gauge representing spring boot health indicator.
     */
    public HealthMetricsConfig(HealthAggregator healthAggregator,
                               HealthIndicatorRegistry healthIndicatorRegistry,
                               List<HealthIndicator> healthIndicators,
                               MeterRegistry registry) {

        healthIndicator = new CompositeHealthIndicator(healthAggregator, healthIndicatorRegistry);

        for (int i = 0; i < healthIndicators.size(); i++) {
            healthIndicatorRegistry.register(Integer.toString(i), healthIndicators.get(i));
        }

        // presumes there is a common tag applied elsewhere that adds tags for app, etc.
        Gauge.builder("health", healthIndicator, health -> {
            Status status = health.health().getStatus();
            switch (status.getCode()) {
                case "UP":
                    return 3;
                case "OUT_OF_SERVICE":
                    return 2;
                case "DOWN":
                    return 1;
                case "UNKNOWN":
                default:
                    return 0;
            }
        }).description("Spring boot health indicator.  3=UP, 2=OUT_OF_SERVICE, 1=DOWN, 0=UNKNOWN")
            .register(registry);
    }

}```

philwebb:
Thanks, I think we're missing a `HealthIndicatorRegistry` bean in our auto-configuration.

snicoll:
@philwebb I am not sure why that is. This is deprecated in favour of `HealthContributorRegistry` and that bean is exposed so I don't understand why the regression might come from.

wilkinsona:
[...]

PR #18203, opened 2019-09-11
add HealthIndicatorRegistry bean -- 18194

polarbear567:
I noticed that this page(http://micrometer.io/docs/guide/healthAsGauge) uses a non-existent construct method, and I don't know if it's right.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/health/LegacyHealthEndpointCompatibilityConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/health/LegacyHealthEndpointCompatibilityConfiguration.java
@@ -16,7 +16,9 @@
 
 package org.springframework.boot.actuate.autoconfigure.health;
 
+import org.springframework.boot.actuate.health.DefaultHealthIndicatorRegistry;
 import org.springframework.boot.actuate.health.HealthAggregator;
+import org.springframework.boot.actuate.health.HealthIndicatorRegistry;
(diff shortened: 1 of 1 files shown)


chenqimiao:
https://github.com/spring-projects/spring-boot/blob/1f280e394bc4403ab5c8ac3bc7d5f55c0da386e4/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/DefaultHealthIndicatorRegistry.java#L33
This class have been deprecated.   @polarbear567

snicoll:
@polarbear567 thanks for the PR but I think the micrometer documentation has to change. Exposing this bean is not going to help as it's indeed deprecated.

polarbear567:
@snicoll , sure, I think it will be right to update the doc of micrometer and spring-boot. And the purpose of this PR is to derive from this issue, so I think that if someone else can use it, if you don't add this bean, it will report an error, although this class has been discarded. Thanks.

polarbear567:
@chenqimiao yes, I know the class has been discarded :), but someone may also use it, you can see this issue: https://github.com/spring-projects/spring-boot/issues/18194

snicoll:
@polarbear567 that's not really why I closed the PR. I am not convinced that a user needs the registry to upgrade their code, there is something else that needs to fundamentally change to let users get the same use case as before (and we're tracking that in #18194). Regardless of all of that, exposing an empty registry is not going to help as it should contain known health indicators.

polarbear567:
@snicoll yes, you're right, this pr is just able to make no mistakes, I think people should know that they shouldn't use it anymore:) I don't know if it's right what I said.

snicoll:
I don't think so I am afraid. Having a bean with an empty registry is more confusing than the error IMO.

Issue #6898, opened 2016-09-15
Update TestTypeExcludeFilter to support JUnit 5 types


philwebb:
https://github.com/spring-projects/spring-boot/issues/6768#issuecomment-243093363

PR #18214, opened 2019-09-12
TestTypeExcludeFilter does not detect JUnit 5's @Testable annotation

leonard84:
As a followup to #6898. The annotation `org.junit.platform.commons.annotation.Testable` can also be applied to the test class so it should also be added to the `CLASS_ANNOTATIONS` array. Spock-2.0 uses `@Testable` on the class instead of each test.


snicoll:
@leonard84 thank you for making your first contribution to Spring Boot.

Issue #18126, opened 2019-09-04
Allow forcing DataSourceClosingSpringLiquibase when manually defining a @LiquibaseDataSource


Tristan971:
Hello, this is an enhancement suggestion (which I'm happy to implement as well but wanted to make sure there's not a good reason that it was not done before prior to working on it).

As part of https://github.com/spring-projects/spring-boot/issues/9218, the `DataSourceClosingSpringLiquibase` class was added.
This allows closing the connection pool once the liquibase migrations are applied (in case of Liquibase being ran on application startup mostly) instead of hanging around for the whole application lifecycle.

Unfortunately it is only auto-configured (instead of the standard `SpringLiquibase`) if no `DataSource` has been configured with `@LiquibaseDataSource` annotated on it, as that datasource would be autowired here:
https://github.com/spring-projects/spring-boot/blob/a6f8351dd64ec288424ceb5bc6e4505e00c7d5a2/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java#L82-L90

and yield the "standard" branch here:

https://github.com/spring-projects/spring-boot/blob/a6f8351dd64ec288424ceb5bc6e4505e00c7d5a2/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java#L118-L129

The proposal is to allow forcing `DataSourceClosingSpringLiquibase` through a property of `LiquibaseProperties`.

Probably something like `closeDataSourceOnceMigrated` as it's already named in `DataSourceClosingSpringLiquibase` itself right now and overriden in `LiquibaseEndpointAutoConfiguration#preventDataSourceCloseBeanPostProcessor` if the Liquibase endpoint is enabled.

As for the motivation, it would allow more customization of the underlying `DataSource` used. So we could set hikari-level configuration that is specific to the Liquibase pool, (pool name, max pool size... and other such things as we already do for other `DataSource` instances) without having to share it with other "post-initialization" datasources.

Thank you

Tristan971:
Hello, 
Just chiming in to know if you have not had time to look at it yet or are discussing it internally?
Or should I make a PR to show how the general idea would work?
Thanks 🙂

wilkinsona:
[...]

PR #18240, opened 2019-09-16
Allow forcing self-closing Liquibase datasource

Tristan971:
Hello,

As per the comments on #18126, here's a PR that adds a property to `@LiquibaseDataSource` to be able to use a `DataSourceClosingSpringLiquibase` with minimal effort when defining your own datasource bean.

This is by opposition to now where you must rely on `LiquibaseProperties` or lose most advantages of the `LiquibaseAutoConfiguration`.

Note that as mentionned in the issue conversation, retrieving the `@LiquibaseDataSource` annotation on the bean declaration is not easy.
The solution here is less than ideal, but I could not find much better.

There are some nicer ways via the bean registry or some concrete context implementations, but I am not familiar enough with how these are picked to make safe assumptions by myself

Let me know if anything is not acceptable.

Thanks,
Tristan Deloche

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseDataSource.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseDataSource.java
@@ -34,7 +34,15 @@ import org.springframework.beans.factory.annotation.Qualifier;
 @Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@Qualifier
+@Qualifier("liquibaseDataSource")
 public @interface LiquibaseDataSource {
+
+	/**
+	 * Defines whether this datasource should automatically be closed once migrations have
+	 * been applied.
+	 * @return whether the {@code SpringLiquibase} created with the annotated datasource
+	 * should specifically be a {@code DataSourceClosingSpringLiquibase}.
+	 */
+	boolean closeDataSourceOnceMigrated() default false;
 
 }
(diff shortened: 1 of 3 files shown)


Tristan971:
Also I mistakenly reformatted assuming the Spring formatting plugin would ensure no more than exactly needed, but some of the imports have switched places and I'm not sure if that's expected or not.

Let me know if I should revert them.

Tristan971:
Unfortunately seems to not quite work as is in the project for which I wanted to use it. The object provider is empty at runtime while the beanfactory properly finds the datasource.

Will inquire a little more but the general idea should not change.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseDataSource.java
```diff
@@ -34,7 +34,15 @@
 @Target({ ElementType.FIELD, ElementType.METHOD, ElementType.PARAMETER, ElementType.TYPE, ElementType.ANNOTATION_TYPE })
 @Retention(RetentionPolicy.RUNTIME)
 @Documented
-@Qualifier
+@Qualifier("liquibaseDataSource")
```
Tristan971: This one was necessary because the custom qualifier's value is sensitive to the annotation properties.

i.e.:
`@LiquibaseDataSource` does not match `@LiquibaseDataSource(closeDataSourceOnceMigrated = false)`

Tristan971: Hopefully there's a nicer way to go about it...

Tristan971:
I've found an issue when the datasource that's auto closed being exposed as a bean means it's picked up by the datasource health auto indicator, which doesn't really like finding the pool closed.
Not sure if the good move is to filter it out once it closes or to on the contrary support a closed datasource pool as "healthy" in that case.
The second sounds really contrived to me so I would think of the first as better ; opinions?

wilkinsona:
@Tristan971 That's an interesting problem and we're not sure what the best solution is at the moment. Unfortunately that means we need to park this one for now and come back to it once 2.2. has been released and we've started planning the 2.3 release.

Tristan971:
Thanks for the feedback, I agree this sounds like it's a little more side-effects than expected initially.
Hopefully there's a solution that we just haven't thought of yet, but otherwise it just might be not worth it I suppose

philwebb:
Thanks for the PR @Tristan971. We're going to have another look at the original issue to see if we can implement a callback based approach. We're worried that closing an externally defined bean will cause other issues. Thanks anyway for the PR and sorry it has taken us so long to review.

Please follow #18126 for updates

Tristan971:
No worries :slightly_smiling_face:

Issue #18241, opened 2019-09-16
Allow shutdown endpoint to accept GET


philwebb:
Kubernetes can trigger a shutdown but only by issuing a `GET`. Our current shutdown endpoint only accept `POST` which makes it unusable.

snicoll:
See #18244

wilkinsona:
I really don't like this. `GET` requests should have no side-effects and be idempotent. Shutting down the app on `GET` violates both of those. My understanding is that Kubernetes will send `SIGTERM` to initiate shutdown. Can't we rely on that instead?

philwebb:
I raised this when watching @saturnism present Kubernetes best practices at a conference. Looking at the documentation, I'm not really sure why the HTTP method would be used over SIGTERM. Perhaps Ray can comment?

spring-issuemaster:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-issuemaster:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #18244, opened 2019-09-17
Allow shutdown endpoint to accept GET

polarbear567:


philwebb:
Thanks for the pull request. We're still having some discussion in #18241 about if we think this is a good idea or not.

polarbear567:
Sure, and do you mean trigger a shutdown by preStop hook in Kubernetes?

philwebb:
@polarbear567 Yes

wilkinsona:
We haven't heard a concrete need for this on #18241 and it has now been closed. Thanks anyway, @polarbear567.

PR #18248, opened 2019-09-17
Remove duplicated row from HealthIndicators table

AlexFalappa:
Removed a duplicate row in the table of auto-configured HealthIndicators.


wilkinsona:
Thanks for spotting and fixing this, @AlexFalappa.

PR #18270, opened 2019-09-18
Fix pre-defined logging group documentation

AlexFalappa:
In the documentation at https://docs.spring.io/spring-boot/docs/2.2.0.BUILD-SNAPSHOT/reference/html/spring-boot-features.html#boot-features-custom-log-groups the table describing loggers belonging to the pre-defined logging groups had a duplicated row for the _sql_ group.

I have also added missing loggers to the _web_ group (list taken from `org.springframework.boot.context.logging.LoggingApplicationListener` source)


snicoll:
In `2.1.x` we're missing the web ones, but the SQL one is new so I'll backport the former only

PR #18294, opened 2019-09-20
Add support for Diagnostic MBean in actuator

ldebello:
Hi,

Currently at work I have created a simple endpoint for exposing Diagnostic MBean from actuator, so I decided to contribute to it with this change.

Functionality:
- Expose operations from Diagnostic MBean.
- Get details for one specific operation.
- Execute an operation with or without parameters.
- Also this endpoint allows to get some resources (This is because some operations like JFR, generate a file for the result, anyway there is a mechanism to filter which kind of resources can be fetched).

Screenshots

Get Operations
image

Get Operation Details
image

Execute Operation
image

If you consider this can be useful but it requires more changes I can work on that but if you consider this is not a valid use case for actuator endpoints just close the pull request.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/diagnostics/VMDiagnostics.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/diagnostics/VMDiagnostics.java
@@ -24,6 +24,7 @@ import java.util.StringJoiner;
 import javax.management.Descriptor;
 import javax.management.InstanceNotFoundException;
 import javax.management.IntrospectionException;
+import javax.management.MBeanException;
 import javax.management.MBeanOperationInfo;
 import javax.management.MBeanServer;
 import javax.management.MalformedObjectNameException;
@@ -207,7 +208,7 @@ public final class VMDiagnostics {
 			result = (String) this.server.invoke(this.objectName, operation, params,
 					new String[] { String[].class.getName() });
 		}
-		catch (Exception ex) {
+		catch (ReflectionException | InstanceNotFoundException | MBeanException ex) {
 			logger.info(String.format("Unable to access operation: %s", operation), ex);
 			result = String.format("Unable to access operation: '%s' - %s", operation, ex.getMessage());
 		}
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/diagnostics/VMDiagnostics.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/diagnostics/VMDiagnostics.java
@@ -52,7 +52,7 @@ public final class VMDiagnostics {
 	 */
 	static final String NO_HELP_AVAILABLE_ERROR_MESSAGE = "No help available";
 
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks very much for the pull-request @ldebello. We plan to discuss this a little more as team when we meet at Spring One in a few weeks. Unfortunately we probably won't be able to merge this in time for 2.2, but we'll review it and consider the options for 2.3.

ldebello:
@philwebb thanks for the update, let me know if I can help in something

ldebello:
Hi, 

I would like to know if there is something from my side that I can do to help with this

wilkinsona:
@ldebello Thanks for the offer, but we don't need anything from you at the moment. As @philwebb indicated above, we'll review and consider this for 2.3 in due course.

ldebello:
Awesome, thanks :)

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/diagnostics/VMDiagnostics.java
```diff
@@ -0,0 +1,320 @@
...
+	}
+
+	/**
+	 * calculate parameter position.
+	 * @param parameterDescriptor descriptor.
+	 * @return the index based on the parameter descriptor.
+	 */
+	private int computeParameterPosition(Descriptor parameterDescriptor) {
+		Integer parameterPosition = (Integer) parameterDescriptor.getFieldValue(ARGUMENT_POSITION_DESCRIPTOR);
+		return (parameterPosition != -1) ? parameterPosition : 0;
```
yydzxz: `(parameterPosition == -1) ?  0 : parameterPosition` maybe more readable

ldebello:
Hi guys. Do I need to update my branch to allow you to review this or you can review and if this is ok do the update branch later?

wilkinsona:
Hi, @ldebello. Thanks again for the offer. We don't need anything from you at the moment.

The core team have discussed this a bit.  We're currently leaning towards not merging the changes but we haven't reached a final decision. The general problem of exposing MBeans over HTTP is something that we'd prefer to be solved by Jolokia for which Spring Boot already has auto-configuration. For those that want access specifically to the diagnostics MBean but do not want to use Jolokia, our current feeling is that support would be better maintained as a third-party starter rather than in Spring Boot itself.

I'll flag this for team attention so that we can discuss this again sometime soon and hopefully reach a final decision.

ldebello:
Awesome, thanks for the reply. Before sending the PR I thought the same about using jolokia, but I thought it could be more useful to expose this in a different way for diagnostics tools. I also thought about exposing it as starter project but I believed as contribution could be more valuable but I can create a starter for this :). If the team arrives to a conclusion please let me know and I can close the PR and create the starter-project.

wilkinsona:
We discussed this one today and I'm afraid that we concluded that we don't want to merge and support it ourselves. Thanks again for your PR and for your [understanding of our decision](https://github.com/spring-projects/spring-boot/pull/18294#issuecomment-593947267). If you do create a third-party starter with this functionality, we'd be happy to add it to the [list of community starters](https://github.com/spring-projects/spring-boot/tree/fa8a09d46d1bb579863a1e16564a72d90eba7bdd/spring-boot-project/spring-boot-starters#community-contributions).

PR #18333, opened 2019-09-24
Enforce ordering when ObjectProvider is used

ttddyy:
Fix to enforce ordering when `ObjectProvider` is used for retrieving multiple objects.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfiguration.java
@@ -68,6 +68,8 @@ public class DataSourcePoolMetricsAutoConfiguration {
 		@Autowired
 		void bindDataSourcesToRegistry(Map<String, DataSource> dataSources, MeterRegistry registry,
 				ObjectProvider<DataSourcePoolMetadataProvider> metadataProviders) {
+			// doesn't matter the ordering here since they are usually distinct and are
+			// used as first match wins basis.
 			List<DataSourcePoolMetadataProvider> metadataProvidersList = metadataProviders.stream()
 					.collect(Collectors.toList());
 			dataSources.forEach(
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/admin/SpringApplicationAdminJmxAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/admin/SpringApplicationAdminJmxAutoConfiguration.java
@@ -62,6 +62,7 @@ public class SpringApplicationAdminJmxAutoConfiguration {
 			ObjectProvider<MBeanExporter> mbeanExporters, Environment environment) throws MalformedObjectNameException {
 		String jmxName = environment.getProperty(JMX_NAME_PROPERTY, DEFAULT_JMX_NAME);
 		if (mbeanExporters != null) { // Make sure to not register that MBean twice
+			// ordering is not important for exporters, thus just iterate them
 			for (MBeanExporter mbeanExporter : mbeanExporters) {
 				mbeanExporter.addExcludedBean(jmxName);
 			}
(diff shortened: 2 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/admin/SpringApplicationAdminJmxAutoConfiguration.java
```diff
@@ -62,9 +62,9 @@ public SpringApplicationAdminMXBeanRegistrar springApplicationAdminRegistrar(
 			ObjectProvider<MBeanExporter> mbeanExporters, Environment environment) throws MalformedObjectNameException {
 		String jmxName = environment.getProperty(JMX_NAME_PROPERTY, DEFAULT_JMX_NAME);
 		if (mbeanExporters != null) { // Make sure to not register that MBean twice
-			for (MBeanExporter mbeanExporter : mbeanExporters) {
+			mbeanExporters.orderedStream().forEach(mbeanExporter -> {
```
wilkinsona: I don't think the ordering is important here.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
```diff
@@ -123,7 +123,7 @@ public Flyway flyway(FlywayProperties properties, DataSourceProperties dataSourc
 			configureCallbacks(configuration, orderedCallbacks);
 			fluentConfigurationCustomizers.orderedStream().forEach((customizer) -> customizer.customize(configuration));
 			configureFlywayCallbacks(configuration, orderedCallbacks);
-			List<JavaMigration> migrations = javaMigrations.stream().collect(Collectors.toList());
+			List<JavaMigration> migrations = javaMigrations.orderedStream().collect(Collectors.toList());
```
wilkinsona: The ordering doesn't matter as Flyway orders them itself based on `MigrationVersion` returned from `getVersion()`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfiguration.java
```diff
@@ -68,7 +68,7 @@
 		@Autowired
 		void bindDataSourcesToRegistry(Map<String, DataSource> dataSources, MeterRegistry registry,
 				ObjectProvider<DataSourcePoolMetadataProvider> metadataProviders) {
-			List<DataSourcePoolMetadataProvider> metadataProvidersList = metadataProviders.stream()
+			List<DataSourcePoolMetadataProvider> metadataProvidersList = metadataProviders.orderedStream()
```
wilkinsona: Does ordering matter here? The providers are used in a first wins manner. For the ordering to be important there would have to be overlap between providers.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionAutoConfiguration.java
```diff
@@ -67,7 +68,7 @@ public TaskExecutorBuilder taskExecutorBuilder(TaskExecutionProperties propertie
 		builder = builder.awaitTermination(shutdown.isAwaitTermination());
 		builder = builder.awaitTerminationPeriod(shutdown.getAwaitTerminationPeriod());
 		builder = builder.threadNamePrefix(properties.getThreadNamePrefix());
-		builder = builder.customizers(taskExecutorCustomizers);
+		builder = builder.customizers(taskExecutorCustomizers.orderedStream().collect(Collectors.toList()));
```
wilkinsona: 👍

ttddyy:
@wilkinsona 
Ok, updated the PR.
I have added comments on the usage of non ordering `ObjectProvider` based on your comments.

snicoll:
@ttddyy thanks for the update but I am not big fan of the comments you've added. We usually don't add those and I don't think it's warranted. I've flagged this one to see what the rest of the team thinks.

philwebb:
Yeah, we'll drop those when we merge it

philwebb:
Thanks for the fix @ttddyy, I've dropped the comments but and merged to 2.1 and 2.2. I also used a method reference trick to save needing to call a collector. The code now looks like this:

```java
builder = builder.customizers(this.taskExecutorCustomizers.orderedStream()::iterator);
```

Thanks again!

Issue #18018, opened 2019-08-30
When using Netty, an invalid content-type header value causes request to get stuck when compression is enabled


wilkinsona:
See https://github.com/spring-cloud/spring-cloud-gateway/issues/1111. We should make our predicate more robust by returning false for a malformed mime type in the `Content-Type` header rather than throwing an exception.

dreis2211:
This got me thinking if this should be done for Jetty, Undertow etc. as well, @wilkinsona ? I'm asking because out of interest I was working on a fix as well, but Jetty seems to hang indefinitely - for reasons that I can't explain yet. Can be reproduced by moving the new test from `NettyReactiveWebServerFactoryTests` to `AbstractReactiveWebServerFactoryTests`. Tomcat seems to work out of the box, Undertow seems to work despite having the same MimeType parsing as Netty in `UndertowCompressionConfigurer`. Might be worth a look?

dreis2211:
While I was working on the fix I also tried this for the servlet based tests and got them to run with similar changes as in 3d23277. Is it worth sharing a PR for the servlet based stuff at least?

dreis2211:
Just saw https://github.com/spring-projects/spring-framework/issues/23553 - seems like we had the same problem then ;)

PR #18337, opened 2019-09-24
Test compression with invalid content type for all webservers

dreis2211:
Hi,

as https://github.com/spring-projects/spring-framework/issues/23553 is fixed we can move the test introduced for #18018 so all webserver implementations are covered and produce the same result.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

wilkinsona:
This should have gone in 2.1.x as well. I've opened https://github.com/spring-projects/spring-boot/issues/18338 to take care of that.

dreis2211:
I wasn't sure. Thanks for handling that, @wilkinsona

PR #18339, opened 2019-09-24
Actuator endpoint time-to-live caching is not respected on WebFlux

### LINKED ISSUE #18324: TTL property is not working for health indicator when using webflux
The `management.endpoint.health.cache.time-to-live` doesn't seem  to work when using Webflux.

I created a dummy project and uploaded it here: https://github.com/worldtiki/healthchecks-ttl/blob/master/src/main/java/com/example/demo/MyHealthIndicator.java

The problem can be reproduced by calling `/actuator/health` multiple times.
Each time the custom health indicator is called, it prints "potato". It should only print potato once since we're asking Spring Boot to cache the responses for 60s, however the string is printed every time. 

This works fine with non reactive Boot but doesn't when using Webflux.
The `CachingOperationInvoker` is working as expected and it does return the cached response, but because it returns a `Mono` (`MonoMapFuseable`) and not the actual response body, the health indicator code is still executed.

dreis2211:
Hi,

this PR is an attempt to fix #18324 . I'm not super familiar with the reactive side of things, but the proposed solution seems to fix the provided sample in #18324. A proper review and feedback is therefore highly appreciated.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/invoker/cache/CachingOperationInvoker.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/invoker/cache/CachingOperationInvoker.java
@@ -16,8 +16,11 @@
 
 package org.springframework.boot.actuate.endpoint.invoker.cache;
 
+import java.time.Duration;
 import java.util.Map;
 import java.util.Objects;
+
+import reactor.core.publisher.Mono;
 
 import org.springframework.boot.actuate.endpoint.InvocationContext;
 import org.springframework.boot.actuate.endpoint.invoke.OperationInvoker;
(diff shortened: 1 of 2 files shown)


dreis2211:
Test failure seems unrelated

philwebb:
Thanks @dreis2211, this is now merged to 2.1.x and 2.2.x. I changing things slightly because I realized that we need to guard against reactor not being on the classpath. I also added `Flux` support whilst I was at it. Thanks again!

snicoll:
> I also added Flux support whilst I was at it.

I must be missing something but I don't understand why `Flux` support was added. I agree this is generally useful but, in the context of the invoker, you'd only get a `Mono<Health>` don't you?

philwebb:
@snicoll I'm not sure we've defined what reactive types you can use with an `@Endpoint`. It's certainly true that the `ReactiveHealthEndpointWebExtension` only returns `Mono`, but `OperationInvoker` is more general so I figured it would be wise to add it.

snicoll:
Sorry @philwebb, my brain read the "Health" endpoint for some reason and I was focused on `ReactiveHealthIndicator`. Lack of sleep I guess :)

PR #18341, opened 2019-09-24
Added bootstrap* properties files to resources with active filtering.

mgrzeszczak:
Added bootstrap* properties files to resources with filtering on. This allows to use maven property placeholders in side them.

This is useful for example when using spring-cloud-config. I wanted to use
```
spring:
    application:
        name: '@project.artifactId@'
```
in bootstrap.yaml but it was not replaced with the appropriate value.

### DIFF
--- a/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/pom.xml
+++ b/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/pom.xml
@@ -32,6 +32,9 @@
 					<include>**/application*.yml</include>
 					<include>**/application*.yaml</include>
 					<include>**/application*.properties</include>
+					<include>**/bootstrap*.yml</include>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, @mgrzeszczak.

I'm not sure that we should do this as it creates a conceptual cycle between Spring Boot and Spring Cloud. There's also the problem that enabling resource filtering when it isn't wanted can cause problems. If someone is using Spring Boot but is not using Spring Cloud they may be unpleasantly surprised by the consequences of Boot expressing an opinion about a file that has nothing to do with it.

mgrzeszczak:
Ok thanks for the reply, so should I just override this in my projects?

wilkinsona:
Yes, I think that's your best option. Thanks anyway for the suggestion.

PR #18360, opened 2019-09-26
Legacy OrderedHealthAggregator overrides default status list

polarbear567:
And some NPE issue


philwebb:
Thanks for the PR @polarbear567. I think that only the changes in `LegacyHealthEndpointCompatibilityConfiguration` are required. I wanted to add some failing tests to ensure that the issue didn't return and since the actual code changes were relatively small I just made them as part of a single commit (8f9fd9709518c48cd6867ec896dd1c7e85d5a0bd). Thanks anyway for the PR!

PR #18372, opened 2019-09-27
Typo in assertion error message produced by ApplicationContextAssert doesNotHaveBean

rweisleder:


snicoll:
@rweisleder good catch, thank you!

PR #18371, opened 2019-09-27
Add missing appendices to PDF reference documentation

dreis2211:
Hi,

this PR is an attempt to fix #18370. It seems that the change made to `index.htmlsingleadoc` in d5adbbb62675889606aed75eede4f8d40bb25358 was simply forgotten in the `.pdfadoc` file.

Cheers,
Christoph

PR #18374, opened 2019-09-27
Fix escaping of pipe character in reference tables

dreis2211:
Hi,

while working on #18371 I noticed that there are some issues with the handling of `|` inside tables in the reference docs. Here's a list of issues that I found, but there might be more:

- The default value of `server.tomcat.internal-proxies` shows `{vbar}` in both HTML and PDF reference instead of `|`
- The description of `server.tomcat.relaxed-path-chars` breaks the table formatting completely in the PDF version
- The description of `server.tomcat.relaxed-query-chars` breaks the table yet another time ;-)

It seems that neither of the escaping strategies work for both HTML and PDF. Luckily, the docs under https://asciidoctor.org/docs/user-manual/#escaping-the-cell-separator provide a third option.

> The cell separator is controlled using the separator attribute on the table block. You’ll want to select a character that will never be used for content. A good candidate is the broken bar, or ¦

Applying this seems to fix the issues mentioned above.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-docs/src/main/java/org/springframework/boot/configurationdocs/CompoundConfigurationTableEntry.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-docs/src/main/java/org/springframework/boot/configurationdocs/CompoundConfigurationTableEntry.java
@@ -45,9 +45,9 @@ class CompoundConfigurationTableEntry extends ConfigurationTableEntry {
 
 	@Override
 	void write(AsciidocBuilder builder) {
-		builder.append("¦");
+		builder.append("|");
 		this.configurationKeys.forEach(builder::appendKey);
-		builder.newLine().appendln("¦").appendln("¦+++", this.description, "+++");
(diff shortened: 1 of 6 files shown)


snicoll:
> breaks the table formatting completely in the PDF version

@dreis2211 just to make sure I understand this one. I've generated a PDF from `master` (with your latest fix to include the appendices) and I don't see an obvious problem for the two keys you've mentioned. I can see the `{vbar}` though.

What have I missed?

dreis2211:
Interesting. Now I can't reproduce this either anymore. I should have taken a screenshot. Maybe an error on my side by not (re-)compiling `spring-boot-configuration-docs` properly beforehand. Sorry for the confusion.

"At least" the `{vbar}` problem remains. But this gets me thinking if a simple escaping with `\|` might work after all. Give me a minute to try this.

dreis2211:
@snicoll The simpler escaping with `\|` seems to work indeed. Could you give it another try whenever you have a spare minute, please?

dreis2211:
Test failure seems unrelated

PR #18383, opened 2019-09-27
Mention Java 13 in system requirements docs

dreis2211:
Hi,

the System requirements section in the docs mention Java 12 as the upper limit, but I guess it's safe to mention 13 here.

Let me know what you think.
Cheers,
Christoph


philwebb:
Nice spot. Thanks!

PR #18397, opened 2019-09-28
Support template uri variable metric tags.

kingbbode:
This PR is the Uri variables Metric Tag feature support.

Metric Tag support of Uri variables will help us create more flexible monitoring metrics.

I'm curious about your thoughts.


wilkinsona:
Thanks for the suggestion, but the values of URI variables are deliberately not included in the created tags as the tag for each variable would have a potentially unbounded set of values.

Issue #16953, opened 2019-05-23
Disable auto-configuration of HiddenHttpMethodFilter by default


vpavic:
`WebMvcAutoConfiguration` currently configures a lot of components that are specific to use of Spring MVC with a templating engine to produce server-side generated `text/html` content. An example of this is `HiddenHttpMethodFilter`, which is enabled by default. Such components present an unnecessary overhead when Spring MVC is solely used to produce e.g. `application/json` based APIs.

This has caught my attention after we discovered that the aforementioned `HiddenHttpMethodFilter` has been taking up to 80% of processing time on certain endpoints. While it was simple enough to disable the filter using `spring.mvc.hiddenmethod.filter.enabled` property, you have to do it in each project and I believe configuration could be made more lightweight there. In the case of this particular filter, a case could be made that it's not really needed even in all projects that use a templating engine.

philwebb:
I'm not sure we've got a good signal that we won't need the `HiddenHttpMethodFilter`. Perhaps we could make it conditional on a `ViewResolver` bean. It feels a bit risky.

@vpavic I'm curious why `HiddenHttpMethodFilter` was taking so much time, it doesn't seem to do all that much. Did you ever discover the cause?

vpavic:
Before opening this issue I also explored options around `ViewResolver` being the condition but wasn't able to come up with anything since `defaultViewResolver` (which is `InternalResourceViewResolver`) will always be there.

The endpoints that were affected by this were accepting `MultipartFile` request parameter - we didn't look that deep in since disabling the filter basically fixed the problem and improved performance significantly but I assume this was related to size of the request as `HiddenHttpMethodFilter` does `ServletRequest#getParameter`. Here's an example of how it looked in New Relic:

![Screenshot from 2019-05-29 18-52-26](https://user-images.githubusercontent.com/1149230/58576332-5f505d00-8244-11e9-98d2-a24af7773d83.png)

Anyway, after taking a closer look at `WebMvcAutoConfiguration` I realized that significant part of it was not needed where you're using Spring MVC to simply serve API and not server-side generated HTML. Having in mind that nowadays a setup like this is quite common I'd like to see Spring Boot provide a way to have more suitable (i.e. reduced) set of MVC components registered.

vpavic:
[...]

PR #18435, opened 2019-09-29
Fix default values in metadata for HiddenHttpMethodFilter

dreis2211:
Hi,

this PR fixes the default values of `spring.mvc.hiddenmethod.filter.enabled` and `spring.webflux.hiddenmethod.filter.enabled` in the metadata and thus in the docs as reported in [this comment](https://github.com/spring-projects/spring-boot/issues/16953#issuecomment-536295963).

Cheers,
Christoph


snicoll:
Wow, I was about to fix it. You're fast @dreis2211! :)

dreis2211:
I can only give that back given that you're reacting that fast on a Sunday, @snicoll :)

PR #18436, opened 2019-09-29
Fix #16998 - improve logic for setting server.port to -1 only for web…

### LINKED ISSUE #16998: server.port should not be set to -1 if the test does not start an embedded web server
when i use spring-boot for no-web application, in the spring-boot-test ,"server.port" was be override by `org.springframework.boot.test.context.SpringBootContextLoader.getInlinedProperties()`
`properties.add("server.port=-1")` . therefore on my unit test, i cannot start server.

application.properties:

```
server.port=12345
server.name=myserver`
```

MyServer.java:

```java
@Component
public class MyServer implements Closeable {

    @Value("${server.port}")
    private int serverPort;

    @Value("${server.name}")
    private String serverName;

    private ServerSocket serverSocket;

    @PostConstruct
    public void init() {
        System.out.println("server.port:" + serverPort);
        System.out.println("server.name:" + serverName);
        // start my server on port "server.port"
        ....
    }
   
}
```

SpringBootDemoApplication.java:

```java
@SpringBootApplication
public class SpringBootDemoApplication {

		SpringApplication.run(SpringBootDemoApplication.class, args);
	}

}
```

SpringBootDemoApplicationTest.java:

```java
@RunWith(SpringRunner.class)
@SpringBootTest
public class SpringBootDemoApplicationTest {

    @Resource
    private MyServer myServer;

    @After
    public void tearDown() throws Exception {
        myServer.close();
    }

    @Test
    public void contextLoads() throws Exception {
        String message = MyClient.sendToServer();
        assertThat(message, startsWith("hello client!"));
    }


}
```

[spring-boot-demo.zip](https://github.com/spring-projects/spring-boot/files/3230669/spring-boot-demo.zip)

785172550:
Fix  #16998 - to set server.port to -1 only for web application and who doesn't have custom server port.
I have the same issue as it, so l'd like to sumbit a PR to fix it.


pivotal-issuemaster:
@785172550 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=18436)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=18436) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@785172550 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=18436)!

785172550:
When I fix the issue #16998, my IDEA inspected some weakness to improve so I do a second commit BTW

philwebb:
> When I fix the issue #16998, my IDEA inspected some weakness to improve so I do a second commit BTW

Could you please submit those as a second pull-request? Mixing the two things in one PR makes it harder to see the fix. Thanks!

785172550:
@philwebb ok will do that

785172550:
@philwebb could you take a review of this PR

snicoll:
@785172550 thanks for the PR but setting `server.port` to -1 for the embedded use case feels a bit odd since it either should be a defined port or a random port. Setting this to `-1` should have no effect as the chosen mode will override the port anyway. 

It looks like more we should not set it to -1 at all. Let's resume the conversation on the issue.

785172550:
@snicoll yes, I dont know why set it to -1 for all case. And one more thing, ``@TestPropertySource(properties = "server.port=12345")`` will overide the ``server.port`` while properties file not

PR #18442, opened 2019-09-30
Remove prohibited version of Maven Invoker Plugin

dreis2211:
Hi,

since Maven Invoker Plugin was updated to 3.2.1 lately, I guess the prohibited version for 3.2.0 can be removed from the `bomr.yaml`. I wasn't sure if it should stick around, but a downgrade that makes this relevant again seems unlikely.

Let me know what you think.
Cheers,
Christoph


snicoll:
Good catch again. Thanks @dreis2211!

PR #18443, opened 2019-09-30
Move the JUnit 4 support to Surefire and Failsafe

phillipuniverse:
The previous support required the JUnit Jupiter engine on the classpath. This further clarifies the default support for JUnit 4 to require an extra dependency within projects that require it, while the default support will be JUnit 5.

It was odd to me that brand new projects require an exclusion from `spring-boot-starter-test` to kick off the default support for JUnit 5. Since it is as simple as adding a user dependency to the vintage engine, why not go with that?


wilkinsona:
Thanks for the suggestion, but the starter needs to work for Gradle users and for Maven users who are not using `spring-boot-starter-parent`.

PR #18445, opened 2019-09-30
Fix comment position in ZipHeaderPeekInputStreamTests

dreis2211:
Hi,

I just found this comment that was not placed directly above the class but between package and import statements. I wonder if checkstyle/spring-javaformat could catch that.

Cheers,
Christoph


snicoll:
@dreis2211 thanks, weird oversight. I don't know if we could check for that but feel free to open an issue there if you think that would be worthwhile.

PR #18446, opened 2019-09-30
Upgrade to Testcontainers 1.12.2

dreis2211:
Hi,

this PR upgrades Testcontainers to version 1.12.2. Am I right that only the `pom.xml` under `spring-boot-dependencies` is covered by `bomr` and hence the (half-)automatic upgrade process? If not, please ignore this PR. ;-)

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-parent/pom.xml
+++ b/spring-boot-project/spring-boot-parent/pom.xml
@@ -26,7 +26,7 @@
 		<maven.version>3.5.4</maven.version>
 		<maven-resolver.version>1.1.1</maven-resolver.version>
 		<spock.version>1.0-groovy-2.4</spock.version>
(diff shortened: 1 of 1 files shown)


snicoll:
@dreis2211 you are right. `bomr` uses the content of `spring-boot-dependencies` but it could very well as much looks a different pom if instructed to do so. I thought one could override but it's driven by the yaml file now. They were interesting upgrades in there so I'll look at that a bit more and follow-up accordingly. 

Thanks for the nudge!

dreis2211:
Thanks @snicoll . That's what I thought as well. If you don't have time, I can take a look too. Just let me know.

snicoll:
I kind of already did and need to dig into the numbers, see #18448

PR #18464, opened 2019-10-01
Code improvement in `for` statement to use lambda expression

ze12augusto:
Code improvement in `for` statement to use lambda expression


wilkinsona:
Thanks for the proposal. We try to limit our use of Streams as they have quite a significant overhead in terms of thread local allocations that contributes to GC pressure. Some of these may be OK if they're not in a hot code path but any potential benefit is outweighed by the cost of measuring the impact. Thanks anyway.

PR #18477, opened 2019-10-02
Add some more notes on running exploded jar files

dsyer:
People building containers benefit from layer separation, and also can get a startup time boost by using the application "natural" main method. This change expands the notes on running an exploded jar to mention those two things briefly.

There is a tutorial that covers this in more detail (https://spring.io/guides/topicals/spring-boot-docker/) but I wasn't sure if we should link to there from the user guide (so I didn't).

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-executable-jar-format.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-executable-jar-format.adoc
@@ -164,18 +164,43 @@ The classpath is deduced from the nested jars.
 
 
 [[executable-jar-exploded-archives]]
-=== Exploded Archives
-Certain PaaS implementations may choose to unpack archives before they run.
+=== Containers and Exploded Archives
+If you are running your application from a container, you can use the unexploded jar file as above, but it is also often an advantage to explode it and run it in a different way.
+Certain PaaS implementations may also choose to unpack archives before they run.
 For example, Cloud Foundry operates this way.
-You can run an unpacked archive by starting the appropriate launcher, as follows:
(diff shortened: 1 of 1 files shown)


sdeleuze:
I see a lot of value in that update, that makes users aware of the most efficient way to run a Boot app (`java -cp BOOT-INF/classes:BOOT-INF/lib/* com.example.MyApplication`) still very easy to use since that's a one liner, and that's IMO important to provide guidance for such usage on containers in addition to PAAS like CF.

spring-boot-project/spring-boot-docs/src/main/asciidoc/appendix-executable-jar-format.adoc
```diff
@@ -164,18 +164,43 @@ The classpath is deduced from the nested jars.
...
+Once you have unpacked the jar file, you can also get an extra boost to startup time by running the app with its "natural" main method instead of the `JarLauncher`. For example:
+
+[indent=0]
+----
+	$ jar -xf myapp.jar
+	$ java -cp BOOT-INF/classes:BOOT-INF/lib/* com.example.MyApplication
+----
+
+More efficient container images can also be created by copying the dependencies to the image as a separate layer from the application classes and resources (which normally change more frequently).
+There is more than one way to achieve this layer separation.
```
larsgrefer: You could also mention https://github.com/GoogleContainerTools/jib here.

dsyer: We could. But I intentionally didn’t, to avoid having to enumerate a bunch of similar tools. There are other sources for information about build tools.

Issue #18392, opened 2019-09-28
Consider moving LocalServerPort


philwebb:
It feels like it should be in `org.springframework.boot.web.context` next to the `ServerPortInfoApplicationContextInitializer`

wilkinsona:
Superseded by #29589.

PR #18501, opened 2019-10-03
Move LocalServerPort to web-context. Fixes gh-18392

prashant171992:


philwebb:
Thanks for the PR but I'm not sure we'll get to this one for a while. Sine it's a relatively straight forward refactor, I think I'd prefer to close this one and handle it directly ourselves when we decide which milestone it should be in.

Thanks anyway!

PR #18503, opened 2019-10-03
Fix header x application context

ZhuoSiChen:
This PR is about management. server. add-application-context-header = true, and this configuration item does not work.
I saw this [problem](https://stackoverflow.com/questions/51755731/x-application-context-header-is-not-being-passed-to-htttpheaders-map-in-openshif) on stackoverflow, so I decided to solve it. This is my solution. What's wrong with it? Please explain it in detail.


wilkinsona:
Thanks for the pull request. Unfortunately, I can't see the connection between the changes that you have proposed and the problem described on Stack Overflow.

The problem on Stack Overflow describes different behaviour when deploying to a local Liberty installation and to Liberty running on OpenShift. Moving the bean definition for the filter into `WebMvcEndpointChildContextConfiguration` will mean that it's only available when the management server's been configured with a separate port to the main server. This is something that can only be done when using an embedded container so it isn't applicable to the question on Stack Overflow. Moving the definition of the filter in the child management context will also break the functionality for users using an embedded server with everything on the same port.

Without an example that reproduces the problem on Stack Overflow I think it's going to be quite hard to justify making a change here. When you were working on the proposed changes, did you manage to reproduce the problem that's described? If so, could you share that with us?

ZhuoSiChen:
I think the Filter is only work on  management.server.port=0 was set by the user and don't work on an embedded server with everything on the same port.Just like the history which the ApplicationContextHeaderFilter don't extract. Maybe we should Integate the management.server properties.

wilkinsona:
> and don't work on an embedded server with everything on the same port

Your fix will have the opposite effect. It removes the filter when everything is on the same port so it will only work when the app and the management server are running on different ports.

Thanks for your efforts, but I'm going to close this as I don't think we have a good enough understanding of the problem to be considering code changes. If you would like to pursue this, please open an issue with a small sample that reproduces the problem you are trying to fix. Once we have that we can discuss how best to fix it. At that point, a pull request that makes the necessary changes would be most welcome.

PR #18534, opened 2019-10-08
Optimize some code

OLPMO:


wilkinsona:
Thank you for the PR, @OLPMO. I have merged the changes into master, other than those to `JSONArray` which we want to keep in sync with the original source code.

OLPMO:
That's all right.

PR #18538, opened 2019-10-08
Use dedicated method to turn off Neo4jContainer's authentication

michael-simons:
Just a small thing. We added this method to Testcontainers to have a nicer API.
Thanks again for the hint some time ago.


snicoll:
Thanks for following-up. They were two extra usages that I've fixed in 32a1cdf

PR #18539, opened 2019-10-08
Remove accidental usage of UnsupportedOptionException

dreis2211:
Hi,

just noticed the usage of `org.xnio.channels.UnsupportedOptionException` that seems accidental. Probably Java's `UnsupportedOperationException` was wanted here. Let me know when I should add this to the illegalClasses check - I wasn't sure if I should do this already.

Cheers,
Christoph


dreis2211:
Build breaks because of checkstyle.

wilkinsona:
Thanks very much, @dreis2211.

PR #18548, opened 2019-10-09
Fix ImmutableConfigurationProperties import statement in documentation

polarbear567:
Fix ImmutableConfigurationProperties import statement in documentation


wilkinsona:
Thank you, @polarbear567.

PR #18557, opened 2019-10-10
Remove usage of @InjectMocks

dreis2211:
Hi,

this PR removes the usage of `@InjectMocks` inside tests. For some people - including me - it's similar to field injection and thus not desirable to have.

Cheers,
Christoph


philwebb:
I think we might even consider adding a checkstyle restriction for this.

wilkinsona:
I'd quite happily get rid of `@Mock` too so that `setUp()` isn't needed at all and the two fields can be final:

```java
private final IntegrationGraphServer server = mock(IntegrationGraphServer.class);

private final IntegrationGraphEndpoint endpoint = new IntegrationGraphEndpoint(this.server);
```

dreis2211:
Let me incorporate your comments. Hope SpringOne Platform was/is a blast.

snicoll:
Well done and thank you again @dreis2211. The conference went well!

Issue #18527, opened 2019-10-07
AOP Smoke test package reference is incorrect


philwebb:
the AOP example no longer works following the move as the advice needs to change to reflect the new package name.

vivganes:
@philwebb I can take and work on it. Can you please point me to the URL where i can find the incorrect README?

philwebb:
Thanks for the offer @vivganes. I actually made a mistake raising this issue, it's not the readme but [this](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-aop/src/main/java/smoketest/aop/monitor/ServiceMonitor.java#L29) expression that needs to change.

snicoll:
Closing in favour of PR #18558

PR #18558, opened 2019-10-10
Fix package in AOP pointcut

### LINKED ISSUE #17393: Move smoke tests under spring-boot-test
Along with that, the hibernate sample is more of a smoke test than an integration test. 

After the move, we still need to ensure that the smoke-tests run in parallel with the integration and deployment tests on CI.

jaferkhan:
Hello people,
This pull request modifies the existing pointcut expression to cater for the package name change that occurred earlier as part of the issue #17197

This fixes the issue #18527


snicoll:
@jaferkhan thank you for making your first contribution to Spring Boot.

jaferkhan:
@snicoll You're welcome :)

Issue #17983, opened 2019-08-28
Update getting started with start.spring.io instructions


philwebb:
Our getting started documentation has a lot of details about installing Gradle and Maven. It would be nice if we also pointed to start.spring.io and showed that the generated project has the wrapper.

rajputvishwas:
Hi phil,
I would like to work on this, if it is still open?

snicoll:
@rajputvishwas thanks for the offer. Yes the issue is still open.

codebuster007:
Hi! @wilkinsona ,

I'd like to know if this issue is still open and unassigned. I'd like to work on it.

My approach for this was to: 

Include a `TIP Section` just after [Installation Instructions for the Java Developer](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started.adoc#installation-instructions-for-the-java-developer) which will inform readers that the next two sections talks about [Maven ](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started.adoc#maven-installation) and [Gradle ](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started.adoc#gradle-installation) installation but they do not necessarily need to install them to work with SpringBoot as they could easily [Generate a SpringBoot Project](https://start.spring.io/) and the generated project will include Maven and Gradle wrappers as the case maybe which they can then use to build the project as required. 

Is this what is needed?

wilkinsona:
Thanks for the offer. We're not quite sure yet what is needed, hence the `pending-design-work` label. In addition to the section that you've linked to above, there's also [this section](https://docs.spring.io/spring-boot/docs/2.2.5.RELEASE/gradle-plugin/reference/html/#getting-started) in the Gradle plugin's documentation to consider. We need to spend some time figuring out exactly what we want to say and where. Unfortunately we have other, higher priority issues that we need to focus on at the moment so it may take some time before we're can give this issue the attention that it needs before some progress can be made.

snicoll:
Talking about the Gradle Wrapper, Maven has a wrapper as well now so the current doc feels a bit off.

PR #18560, opened 2019-10-11
files via upload

rajputvishwas:
With reference to issue #17983 sending update  getting started with start.spring.io instructions

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/getting-started.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/getting-started.adoc
@@ -65,8 +65,6 @@ Spring Boot supports the following embedded servlet containers:
 
 You can also deploy Spring Boot applications to any Servlet 3.1+ compatible container.
 
-
-
 [[getting-started-installing-spring-boot]]
 == Installing Spring Boot
 Spring Boot can be used with "`classic`" Java development tools or installed as a command line tool.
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal, @rajputvishwas.

I think this adds a bit too much detail about using start.spring.io and means that it's likely that things may become out-of-date quite quickly. I believe that what @philwebb had in mind was a smaller new section that encourages users to start a project using start.spring.io and to describe that the generated project will include Maven or Gradle's wrapper  without going into too much fine detail.

Let's circle back to #17983 and refine exactly what we want the updates to be. Sorry for the wasted effort here.

rajputvishwas:
Hi Andy,

I have reduced it as per your instructions. Let me know if it still needs some modification.

4. Spring Initializr

You can create Spring Boot application using “Spring Initializr”. It’s a web application which generates basic project structure for you with either Gradle or Maven based build specification.

Open Spring Initializr<https://start.spring.io/> in a new tab. You should see the form with sample fields and default values. You need to fill new values in these fields.

Once you are done with the form and dependencies selection, click on “Generate the Project” button to download project template. This action downloads a zip file, which contains empty Spring Boot application with wrapper, project directory structure and script file.

After you unzip the file you should get the deliverables as illustrated in the below table.
Deliverables
Description

mvnw and mvnw.cmd

This Maven wrapper is automatically generated, so you need not to install Maven to run the application.

src

This folder contains two sub folders “main” and “test”. Which in turn have the packages specified by you in the Spring Initializr form.

pom.xml

This file stores project metadata and build information.

When you select Gradle Project instead of Maven, your project folder contains a Gradle wrapper.

To get more details, visit Spring Initializr Reference<https://docs.spring.io/initializr/docs/current/reference/html/>.

PR #18562, opened 2019-10-11
Upgrade to rest-assured 4.1.2

clemstoquart:
Hi,

Since there are some breaking changes between rest-assured 3.3 and 4.0, I think it would be nice to upgrade to 4.x before Spring Boot 2.2 release.

Have a good day.


wilkinsona:
Thanks for the suggestion, but having reached RC1 it's too late to do this for Boot 2.2. We'd also need to be sure that the rest of the portfolio, such as Spring REST Docs (https://github.com/spring-projects/spring-restdocs/issues/635), is compatible with REST Assured 4. We can consider this again once we've started work on 2.3.

PR #18579, opened 2019-10-14
Rename tests to match Surefire expectations

dreis2211:
Hi,

just noticed two tests that don't match the include pattern for surefire and are thus not run during CI builds.

Cheers,
Christoph


dreis2211:
Test failure seems unrelated.

snicoll:
Thanks again Christoph!

PR #18580, opened 2019-10-14
Use LinkedHashSet for deterministic order in test assertions

contextshuffling:
Test `org.springframework.boot.context.properties.bind.CollectionBinderTests#bindToSetShouldNotAllowDuplicateValues` depends on the iteration order of `itemsSet` in `ExampleCollectionBean`. However, it is a `HashSet` and `HashSet` does not guarantee any specific order of entries. Thus, test can fail due to a different iteration order.

In this PR, we propose to use a `LinkedHashSet` instead to guarantee the iteration order (as the insertion orders).


snicoll:
@contextshuffling thank you for making your first contribution to Spring Boot. This is merged in `2.1.x` and  `master`.

PR #18581, opened 2019-10-15
Document spring.main.register-shutdown-hook

polarbear567:


snicoll:
@polarbear567 thank you very much for the update and crafting a PR for `2.1.x`.  I've [polished your contribution](https://github.com/spring-projects/spring-boot/commit/22f44917ca79d71878d0cd448110e719cf092298) as configuration properties are ordered both in the metadata file and the appendix.

polarbear567:
That's right. I didn't think about the order.

PR #18604, opened 2019-10-15
Optimize debug level logs

wycm:
Optimize debug level logs

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/livereload/Connection.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/livereload/Connection.java
@@ -108,8 +108,10 @@ class Connection {
 			else if (frame.getType() == Frame.Type.CLOSE) {
 				throw new ConnectionClosedException();
 			}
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/livereload/Connection.java
```diff
@@ -106,7 +108,7 @@ private void readWebSocketFrame() throws IOException {
 			else if (frame.getType() == Frame.Type.CLOSE) {
 				throw new ConnectionClosedException();
 			}
-			else if (frame.getType() == Frame.Type.TEXT) {
+			else if (frame.getType() == Frame.Type.TEXT && logger.isDebugEnabled()) {
```
wilkinsona: This will cause `TEXT` frames to be unexpected when debug logging isn't available.

wycm: I will modify it later, thank you for your reply.

snicoll:
@wycm thank you for making your first contribution to Spring Boot.

PR #18612, opened 2019-10-16
Use LinkedHashSet for deterministic order in test assertion

contextshuffling:
Test case `testSerialization` is trying to test the serialization of CompositeHealthIndicator. However, there is a subtle dependency that the serialization depends on, which is the iteration order of HashMap it uses. However, HashMap does not guarantee any specific order of entries. Therefore, the assertions will fail if the order is different.
Test cases should pass regardless of the HashMap iteration orders, otherwise it cannot serve as a regression test. 

This PR proposes to change HashMap to use LinkedHashMap that makes the iteration order deterministic, which is the insertion order. The test also assumes this order so no need to modify the test assertion.


omix222:
キャンセル
2019年10月16日(水) 10:36 contextshuffling <notifications@github.com>:

snicoll:
@contextshuffling thanks again!

PR #18637, opened 2019-10-17
Narrow type for Tomcat 'locale' and 'encoding' properties

AlexFalappa:
It might make sense to define `handle-as` providers for `server.tomcat.accesslog.locale` and `server.tomcat.accesslog.encoding` configuration properties.

Both are strings but represent a `java.util.Locale` and a `java.nio.charset.CharSet` respectively.

This way IDEs can provide completion for those properties.


snicoll:
Good stuff @AlexFalappa, thank you very much!

PR #18648, opened 2019-10-18
Test the Gradle Plugin against Gradle 5.6.3

dreis2211:
Hi,

Gradle 5.6.3 was released earlier today and we can test against it.

Cheers,
Christoph


wilkinsona:

PR #18659, opened 2019-10-18
Immutable configuration properties not shown by Actuator

polarbear567:
After testing, I found the problem was on the serializer modifier, so I made this change.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
@@ -233,7 +233,7 @@ public class ConfigurationPropertiesReportEndpoint implements ApplicationContext
 		@Override
 		public Object findFilterId(Annotated a) {
 			Object id = super.findFilterId(a);
-			return id == null ? CONFIGURATION_PROPERTIES_FILTER_ID : id;
+			return (id != null) ? id : CONFIGURATION_PROPERTIES_FILTER_ID;
 		}
 
 	}
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
@@ -367,8 +367,7 @@ public class ConfigurationPropertiesReportEndpoint implements ApplicationContext
 
 		private boolean isConstructorsContainParam(Constructor constructor, Class clazz, String name,
 				boolean isEnclosing) {
-			int size = constructor.getParameterCount();
-			for (int i = 0; i < size; i++) {
+			for (int i = 0; i < constructor.getParameterCount(); i++) {
 				if (!isEnclosing) {
 					if (constructor.getParameterTypes()[i].equals(clazz)
 							&& constructor.getParameters()[i].getName().equals(name)) {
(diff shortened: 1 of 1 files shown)


polarbear567:
@snicoll  Could you please spare time to look at it? I don't know why this kind of failure occurs. I tested all cases of jarcommandit locally, and they can pass.

snicoll:
@polarbear567 I am sure you have the best intention but there is no need to at me for a review, the PR is open and someone from the team will have a look to it when time permits.

polarbear567:
Hi, @snicoll I think you misunderstood me. It's not that this PR needs your review, but that there is a problem with the CI build of spring-boot(like this: https://ci.spring.io/builds/93431). I want to ask you to help me to have a look this problem. After that, I tried again and passed.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
```diff
@@ -231,10 +233,7 @@ else if (item instanceof List) {
 		@Override
 		public Object findFilterId(Annotated a) {
 			Object id = super.findFilterId(a);
-			if (id == null) {
-				id = CONFIGURATION_PROPERTIES_FILTER_ID;
-			}
-			return id;
+			return (id != null) ? id : CONFIGURATION_PROPERTIES_FILTER_ID;
```
snicoll: This is unrelated to the task at hand, please refrain from polishing code even if it's in the same class.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
```diff
@@ -312,6 +311,13 @@ public void serializeAsField(Object pojo, JsonGenerator jgen, SerializerProvider
 		}
 
 		private boolean isReadable(BeanDescription beanDesc, BeanPropertyWriter writer) {
+			// if the class has the @ConstructorBinding annotation or
+			// one constructor of this class has the @ConstructorBinding annotation,
+			// we can ignore the setter method.
```
snicoll: Unfortunately, that's not what the code should be doing. If constructor binding is to be used, only the parameters of the matching constructor should be used to determine the properties. Accessors (getters) shouldn't be checked at all.

polarbear567:
@snicoll Thank you for your opinion. I'd like to work this. It's complicated but interesting.

polarbear567:
Sorry to click the re-review button first. At present, it is judged according to the parameters in the constructor. There are many situations for the application of this annotation. I have considered all situations as much as possible, but I am not sure whether there will be any omissions.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
polarbear567: Yes, I will change this back, this is a temporary modification I made to re trigger CI builder. If you have any other methods that can be triggered repeatedly, please let me know.

snicoll:
@polarbear567 every commit you push sends a notification to a large number of watchers. In the interest of keeping those to an acceptable level, please refrain from pushing isolated commits and, in particular, commits to trigger a new build. The build should work for you locally before submitting changes and we can figure out on our side what led to the build failure when reviewing.

polarbear567:
@snicoll Sure, I will notice this. And I'm not sure how to rebuild cli build, because my local build was successful

snicoll:
@polarbear567 thanks for the efforts thus far but I think this is more involved than I thought initially and the lack of tests in this PR makes it hard to track what is covered and what isn't. I've started to roll-out a solution of my own and I've a number of tests that are covering the change. Given the complexity and potential change in other part of the codebase I am going to close this one. Thanks anyway.

PR #18668, opened 2019-10-20
Adapt Javadoc to reflect the removal of deprecated code.

cosminhume:
A trivial update of the documentation, removing references to methods that were first deprecated and then removed.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/MultipartConfigFactory.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/MultipartConfigFactory.java
@@ -22,8 +22,7 @@ import org.springframework.util.unit.DataSize;
 
 /**
  * Factory that can be used to create a {@link MultipartConfigElement}. Size values can be
- * set using traditional {@literal long} values which are set in bytes or using more
- * convenient {@link DataSize} variants.
+ * set using {@link DataSize} variants.
  *
  * @author Phillip Webb
  * @since 1.4.0


pivotal-issuemaster:
@cosminhume Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=18668)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=18668) to manually synchronize the status of this Pull Request.

cosminhume:
@pivotal-issuemaster This is an Obvious Fix

pivotal-issuemaster:
@cosminhume This Pull Request contains an obvious fix. Signing the Contributor License Agreement is not necessary.

wilkinsona:
Thanks for the suggestion but the javadoc is correct as it is. You can use a long value that is interpreted as data size in bytes or use something like 1MB which is a data size in megabytes.

snicoll:
@wilkinsona Actually we used to have methods in there that takes a `long` and `a `DataSize`. We removed the former so I think this update to the Javadoc in this context makes sense. Thoughts?

wilkinsona:
I don't think it makes sense as proposed as, IMO, too much context is lost. Specifically, it's not clear that you can specify a value without a unit and that it will be treated as bytes. If we feel it's necessary to describe how things can be set, I think the javadoc should remain unchanged. Alternatively, I'd remove the whole sentence about how things can be set as we don't provide such a description for any other class that I can recall.

snicoll:
Good point on losing the context and I like your idea of removing the last sentence altogether. @cosminhume would you be willing to rework your PR in that direction?

cosminhume:
@snicoll I removed the last sentence altogether, please see #18755

Issue #18666, opened 2019-10-20
LiquibaseEntityManagerFactoryDependsOnPostProcessor is not called since spring-boot 2.2.0


kflorian:
Since upgrading to spring-boot 2.2.0 (from 2.1.9), our Liquibase changelogs are no longer deployed before the Hibernate container starts.

My investigation indicates that the LiquibaseEntityManagerFactoryDependsOnPostProcessor is no longer called with spring-boot 2.2.0, causing the dependency from entityManagerFactory to liquibase to be missing.

I suspect that the cause is commit 795303d6676c:

It replaced the @Configuration annotation on the post-processor by an @Import annotation on the LiquibaseConfiguration. However, there is also a @ConditionalOnMissingBean(SpringLiquibase.class) annotation on that bean, and since we are providing our own SpringLiquibase bean definition, I suspect that the entire configuration including the import is skipped.

If this is intended behaviour, then it would at least be nice to include this breaking change in the release notes.
Preferably, the LiquibaseEntityManagerFactoryDependsOnPostProcessor should still be triggered, even if the application provides it's own SpringLiquibase bean definition.

Thanks!

Best regards,
Florian Kirchmeir

wilkinsona:
Closing in favour of #18673.

alexis-puska:
Did the liquibase:diff goal works for you ? did you use liquibase-hibernate5 maven plugins ? 
i have this issues https://github.com/liquibase/liquibase-hibernate/issues/222

kflorian:
Hi Alexis,

I didn't use liquibase:diff or the liquibase-hibernate5 maven plugin.
From reading your issue, I actually don't think the two are related. The issue reported by me was fixed in spring-boot 2.2.1, so it shouldn't affect you.

Best regards,
Florian

> Did the liquibase:diff goal works for you ? did you use liquibase-hibernate5 maven plugins ?
> i have this issues [liquibase/liquibase-hibernate#222](https://github.com/liquibase/liquibase-hibernate/issues/222)

PR #18673, opened 2019-10-21
Depends on semantic between Liquibase and JPA are no longer applied with custom SpringLiquibase

nosan:
`LiquibaseEntityManagerFactoryDependsOnPostProcessor` is no longer called if a user has registered `SpringLiquibase` bean.

This commit fixes this issue by changing the import place.

see gh-18666


snicoll:
Thanks again @nosan

Issue #18653, opened 2019-10-18
@TestConfiguration disables proxying of @Bean methods


agebhar1:
After upgrade to Spring Boot v2.2 the test configuration class which is annotated with `@TestConfiguration` is not proxied by CGLIB anymore, thus calling `@Bean` methods return new instances every time.

After quick search in Spring Framework cores [changes](https://github.com/spring-projects/spring-framework/wiki/Upgrading-to-Spring-Framework-5.x#core-container) gives a hint:

> [...]  or not finding annotations anymore where they have previously been found accidentally. 

The corresponding [commit](https://github.com/spring-projects/spring-framework/commit/fd8fa301a69e9ce586769c12b7572255d0f83f3c) in Spring Framework.

wilkinsona:
Thanks for the report. It may have been a mistake to change `@Configuration` to `@Configuration(proxyBeanMethods=false)` on `@TestConfiguration`. You can work around the current behaviour by replaced `@TestConfiguration` with `@Configuration` and `@TestComponent` instead.

> thus calling `@Bean` methods return new instances every time.

This is only a problem is your own code calls `@Bean` methods directly. If you leave it to the Framework to call them, it will ensure that they are only called once. We recommend not calling `@Bean` methods from your own code as it allows bean method proxying to be disabled which improves startup performance.

agebhar1:
Thank you @wilkinsona for your fast reply and workaround. The `@Bean` method calls are used only in test(s).

rsbeoriginal:
does this needs to be fixed ?

wilkinsona:
Closing in favour of #18675.

PR #18675, opened 2019-10-21
@TestConfiguration disables proxying of @Bean methods

nosan:
See gh-18653


snicoll:
Thanks @nosan. I've polished the tests and added the attribute again in our own use of the annotation.

nosan:
Thanks @snicoll,

> added the attribute again in our own use of the annotation.

I've completely forgotten about this :(

PR #18689, opened 2019-10-22
Fix Mockito deprecations

dreis2211:
Hi,

with the upgrade to Mockito 3 some deprecation warnings were introduced. This PR fixes them.

Cheers,
Christoph


snicoll:
@dreis2211 ironically enough I fixed those in initializr when I upgraded and totally forgot to check here. Glad we have you and thanks again!

PR #18702, opened 2019-10-23
Make optimizedLaunch property name more consistent

AlexFalappa:
Rename maven spring boot plugin `optimizedLaunch` property to `spring-boot.run.optimizedLaunch` to make it more consistent with the other spring-boot:run goal property names.


snicoll:
Gah, bummer. I am tempted to fix this in `2.2.1` even if that's potentially a breaking change. Flagging for team attention to see what the rest of the team thinks.

AlexFalappa:
@snicoll yeah I realised that myself, that's a bit unfortunate.

I would be in favour of renaming it now rather than later cause I think not many users will be affected and tooling has yet to catch up. If 2.2.1 is planned shortly no one will notice 😄.

The fact that it is enabled by default helps because turning it off is less frequently needed.

snicoll:
We've discussed this and agree with you @AlexFalappa. Thanks again for the PR.

PR #18710, opened 2019-10-23
Optimization log

wycm:
Optimization log


philwebb:
Thanks for the PR but I'm not sure we should apply all of these. Some of these log statements happen quite infrequently and as such don't need optimizations. I'll assign this to 2.2.x and we can take a look in more detail before we merge.

philwebb:
Thanks for finding these @wycm. I've polished things a bit in commit 597baf97745b304f75a07ab6eceaae77402a7af8 mainly by making use of Spring Frameworks new `LogMessage` class. I've also remove a few `if` statements where the log message was a simple String so relatively cheap.

Thanks again!

Issue #18686, opened 2019-10-22
Provide toolchains support for Spring Boot Maven plugin


juanpablo-santos:
As per https://maven.apache.org/guides/mini/guide-using-toolchains.html, 

> Maven Toolchains provide a way for plugins to discover what JDK (or other tools) are to be used during the build, without the need to configure them in each plugin nor in every `pom.xml`, or forcing a precise location among every machine building the project. [...] For example, you want to use a different JDK version to build your project than the version used to run Maven,

(my) concrete use case: at work, I've stumbled upon an Spring Boot 2 application, built with Maven + Java 8, which uses some javax.ws classes. When trying to run it having JDK 11 in path, the application fails with something along the lines of
```
[10/22 17:33:31.637][ERROR][o.springframework.boot.SpringApplication][Application run failed]
java.lang.IllegalArgumentException: error Type referred to is not an annotation type: javax$jws$WebService
 [...]
```

which is expected, as `WebService` is no longer present in JDK11, I've to use a lower version. However, I was able to build the application because the compiler plugin was making use of maven toolchains, so the compiler transparently built the application using a JDK8. 

I know I can switch between JDK 8 and 11 on path as needed when running the application through maven, but it would be nice if Spring Boot's maven plugin could provide toolchains support. It's a bit of an edge case, but maybe other people would find this nice to have too?

thx in advance

snicoll:
@juanpablo-santos just to make sure I understood, do you mean to use `mvn spring-boot:run` and use a different JDK for the execution of the application or something more than that?

juanpablo-santos:
Hi @snicoll, thanks for your prompt response :-) yeah I was only referring to use `mvn spring-boot:run` and use a different JDK for the execution of the application.

philwebb:
Closing in favor of PR #18732

PR #18732, opened 2019-10-24
Provide toolchains support for Spring Boot Maven Plugin

nosan:
see gh-18686


philwebb:
Nice, thanks @nosan! We'll look at merging this when we branch for 2.3.

nosan:
sure @philwebb,
The one thing that I want to mention there are no tests for windows os :(

PR #18750, opened 2019-10-25
Fix JsonParser deprecation

dreis2211:
Hi,

just noticed a tiny deprecation warning around GSONs `JsonParser`.

Cheers,
Christoph


dreis2211:
Concourse fails to create a volume at the moment...

mbhave:
Thanks @dreis2211. We're looking into the concourse issue.

PR #18754, opened 2019-10-26
Remove redundant explicit types

dreis2211:
Hi,

this PR removes some explicit type arguments in favor of using the diamond operator.

Cheers,
Christoph


snicoll:
Thanks again @dreis2211!

PR #18755, opened 2019-10-27
Delete Javadoc references to removed code

cosminhume:
A trivial update of the documentation, removing references to methods that were first deprecated and then removed.
@pivotal-issuemaster This is an Obvious Fix
Rework of #18668


snicoll:
@cosminhume thank you for the follow-up and for making your first contribution to Spring Boot.

PR #18769, opened 2019-10-28
Upgrade Kafka clients to 2.3.1

garyrussell:


snicoll:
Thanks for the PR but, as noted in the pull request template, please do not open a pull request for a straightforward dependency upgrade (one that only updates the version property). We have a semi-automated process for such upgrades that we prefer to use. However, if the upgrade is more involved (such as requiring changes for removed or deprecated API) your pull request is most welcome.

PR #18775, opened 2019-10-29
Update bomr config to suggest upgrades with same major version

dreis2211:
Hi,

commit b6e557c changed the config to only suggest versions with the same minor version in order to minimize risks during release creation for 2.2.0. As 2.2.0 is released we can go back to suggesting more upgrades, I guess.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/.bomr/bomr.yaml
+++ b/.bomr/bomr.yaml
@@ -6,7 +6,7 @@ bomr:
       repository: spring-boot
       issue-labels:
         - 'type: dependency-upgrade'
-    policy: same-minor-version
+    policy: same-major-version
     prohibited:
       - project: derby
         versions:


snicoll:
@dreis2211 this is actually the setup we should use for maintenance release. We haven't branched yet so our policy from now on is to upgrade to bugfixes only. 

Have I misunderstood the intention of this PR?

dreis2211:
Ah, in that case I misunderstood the referenced commit. I thought it was only for the 2.2.0 release. Feel free to close this then :)

snicoll:
What Andy meant is that RC is what we consider GA so, moving forward, we should only use a conservative upgrade policy. Here's the one we use in `2.1.x` for reference: https://github.com/spring-projects/spring-boot/blob/217c07a701d6213fc4e8a87d6403bd20baf17f63/.bomr/bomr.yaml#L9

dreis2211:
Thanks for the clarification. Can't wait to work on 2.3.x btw. :)

PR #18808, opened 2019-10-29
Support amqps:// URIs in spring.rabbitmq.addresses

### LINKED ISSUE #6401: spring.rabbitmq.addresses doesn't parse amqps:// uris
I can connect to a RabbitMQ over amqp+ssl by setting these properties:

```
spring.rabbitmq.host: 10.0.0.123
spring.rabbitmq.password: password
spring.rabbitmq.port: 5671
spring.rabbitmq.ssl.enabled: true
spring.rabbitmq.username: user
spring.rabbitmq.virtualHost: virtualhost
``` 

I would expect that this would work as well

```
spring.rabbitmq.addresses: amqps://user:password@10.0.0.123/virtualhost
```

But https://github.com/spring-projects/spring-boot/blob/v1.3.6.RELEASE/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java#L118 doesn't handle the amqps scheme. See https://www.rabbitmq.com/uri-spec.html for the spec on the amqps URI scheme.

markpollack:
If `ampqs` is used in the URI and no explicit port is declared, the port number 5671 will be used instead of 5672

Resolves #6401

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
@@ -960,6 +960,10 @@ public class RabbitProperties {
 
 		private static final int DEFAULT_PORT = 5672;
 
+		private static final String PREFIX_AMQP_SECURE = "amqps://";
+
+		private static final int DEFAULT_PORT_SECURE = 5671;
+
 		private String host;
 
 		private int port;
@@ -970,6 +974,8 @@ public class RabbitProperties {
 
 		private String virtualHost;
 
+		private boolean isSecureConnection;
+
 		private Address(String input) {
 			input = input.trim();
 			input = trimPrefix(input);
@@ -979,6 +985,10 @@ public class RabbitProperties {
 		}
 
 		private String trimPrefix(String input) {
+			if (input.startsWith(PREFIX_AMQP_SECURE)) {
+				this.isSecureConnection = true;
+				return input.substring(PREFIX_AMQP_SECURE.length());
+			}
 			if (input.startsWith(PREFIX_AMQP)) {
 				input = input.substring(PREFIX_AMQP.length());
 			}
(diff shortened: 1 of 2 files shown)


xyloman:
@markpollack thank you for performing this PR this is something that I have had on my list of things to contribute for some time.  I know other in the community will greatly benefit from this update.  :tada:

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -1015,7 +1025,11 @@ private void parseHostAndPort(String input) {
 			int portIndex = input.indexOf(':');
 			if (portIndex == -1) {
 				this.host = input;
-				this.port = DEFAULT_PORT;
+				if (this.isSecureConnection) {
+					this.port = DEFAULT_PORT_SECURE;
+				} else {
```
xyloman: @markpollack you likely already caught this but saw the ci failure and this should be 
```suggestion
				} 
				else {
```

Based upon the [checkstyle rules](https://ci.spring.io/builds/94734#L5d9fa795:3102).

```
[INFO] --- maven-checkstyle-plugin:3.0.0:check (checkstyle-validation) @ spring-boot-autoconfigure ---
[INFO] There is 1 error reported by Checkstyle 8.22 with src/checkstyle/checkstyle.xml ruleset.
[ERROR] src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java:[1030,33] (blocks) RightCurly: '}' at column 5 should be alone on a line.
```

markpollack: Actually I didn't notice, i'll look into it.

mbhave:
There's some room for improvement here where if an `amqps` address is provided, we can set `spring.rabbitmq.ssl.enabled` to `true`. This would remove the need for the user to set `spring.rabbitmq.ssl.enabled: true` in addition to setting an `amqps` address.

snicoll:
Good catch. I suppose we'd need to check that all adresses are either all using `amqps` or `amqp`.  If the net effect of this PR is that users must set the URI and the SSL flag, I think we should reconsider this change to avoid that.

wilkinsona:
I think we should determine whether or not to implicitly enable SSL based on the first address. This will align the behaviour with determining the host, port, etc that are set on the `RabbitConnectionFactoryBean`.

snicoll:
Thank you Mark. As suggested by @mbhave, I've polished your proposal to automatically enable ssl support when `amqps` is detected. That way, you can have SSL configuration and just use it (or not) based on the  actual `address`.

Issue #18812, opened 2019-10-30
Auto Configure ClientRSocketConnector and ServerRsocketConnect for Spring integration RSocket


hantsy:
When I created a sample project to taste Spring Integration RSocket, but I have to create `ServerRSocketConnector` and `ClientRSocketConnector` beans.  And the `ServerRSocketConnector` will create a new RSocket server that could cause a port binding error when running it as a standalone Spring Boot RSocket application if using the same port number.

See: https://stackoverflow.com/questions/58603425/error-running-spring-integration-rsocket-sample-with-spring-boot

It is better to provide auto-configuration to detect the existing RSokcet server and create Spring Integration RSocket facilities based on the existing.

artembilan:
I think we should consider to auto-configure a `ClientRSocketConnector` only in tests when we would like to verify RSocket requests against started RSocket server for test scope. In most cases with random port.

It really does not make sense to always instantiate a `ClientRSocketConnector` whenever we detect RSocket support.
We may consider to have something like `spring.integration.rsocket.client.uri` configuration property to indicate that we need to create a particular RSocket client though...
But at the same time we can simply live with a direct `ClientRSocketConnector` instantiation when we also can inject auto-configured `RSocketStrategies`.

For the server side I need to think more from Spring Integration perspective.
Right now there is no way to co-locate a regular Spring Messaging `@MassageMapping` approach and Spring Integration `IntegrationRSocketEndpoint` in the same RSocket server...
The `ServerRsocketConnect` just encapsulates the whole logic to start a new server - no way to bind to existing one what is done here in Spring Boot.

philwebb:
Closing in favor of PR #18834

hantsy:
Hi, I have refactored my samples using Spring Boot RSocket and Spring Integration RSocket, I have some issues about interaction between them.

I described it in [a new question on stackoverflow](https://stackoverflow.com/questions/60464223/spring-integraton-rsocket-merge-flux-stream-into-a-single-string).

The sample codes can be found [here](https://github.com/hantsy/rsocket-sample/tree/master/integration).

artembilan:
@hantsy ,

please, don't abuse a purse of GH issues and PRs.
We definitely are going take a look into your question on StackOverfow when our weekend is over.

Thanks for understanding.

PR #18834, opened 2019-11-01
Add auto-configuration for Spring Integration RSocket support

artembilan:
Fixes https://github.com/spring-projects/spring-boot/issues/18812

To reuse RSocket server auto-configuration in Spring Integration,
introduce an `IntegrationAutoConfiguration.IntegrationRSocketConfiguration`
based on the `ServerRSocketMessageHandler` from Spring Integration which
overrides an `RSocketMessagingAutoConfiguration` when SI is present on
classpath.
Configure a `ServerRSocketMessageHandler` from SI which is able to handle
SI RSocket channel adapters and `@MessageMapping` if configured via
new `IntegrationProperties.RSocket.Server.messageMappingEnabled`.
Then add a `ServerRSocketConnector` which is based on the provided
`ServerRSocketMessageHandler`.
Also auto-configure a `ClientRSocketConnector` if
`IntegrationProperties.RSocket.Client.port` or
`IntegrationProperties.RSocket.Client.uri` configuration properties are
present

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/pom.xml
+++ b/spring-boot-project/spring-boot-autoconfigure/pom.xml
@@ -489,6 +489,11 @@
 			<optional>true</optional>
 		</dependency>
 		<dependency>
+			<groupId>org.springframework.integration</groupId>
+			<artifactId>spring-integration-rsocket</artifactId>
+			<optional>true</optional>
+		</dependency>
+		<dependency>
 			<groupId>org.springframework</groupId>
 			<artifactId>spring-jms</artifactId>
 			<optional>true</optional>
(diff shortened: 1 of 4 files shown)


artembilan:
Depends on https://github.com/spring-projects/spring-integration/pull/3097

artembilan:
The upstream PR in SI has been merged, so this is ready for review and feedback.

While I'm here I have realized that such an integration test I have here with an explicit port for RSocket server could be is not what you would like to see in the auto-configuration module.
The sample might be the better place to show the feature and in this new test it would be better to check just bean auto-configured after apply all the conditions.

Let me know how to proceed!

Also: do you need some docs on the matter?

Thanks

wilkinsona:
Let's look at this for 2.3.0.M2.

Issue #18843, opened 2019-11-02
Drop JDK 12-based CI


wilkinsona:
Now that Java 13 is out and Java 12 is EOL, I think we should drop our JDK 12-based CI builds. That would leave 2.1.x building against 8 and 11 and 2.2.x building against 8, 11, and 13.

snicoll:
Closing in favour of PR #18848

PR #18848, opened 2019-11-04
Drop JDK 12-based CI

dreis2211:
Hi,

this PR removes the JDK12 build as discussed in #18843 . I targeted 2.1 already - hope that's okay.

Cheers,
Christoph


snicoll:
Perfect Christoph, thank you. I've updated the pipelines in both `2.1.x` and `master` to remove the JDK12 build.

PR #18855, opened 2019-11-04
Polish comment in pom.xml

TASure:

PR #18857, opened 2019-11-04
Update pom.xml

TASure:


wilkinsona:
This is a duplicate of #18855.

PR #18864, opened 2019-11-04
Remove non-existent FailureAnalyzer from spring.factories

dreis2211:
Hi,

I just noticed that `InvalidConfigurationPropertiesFailureAnalyzer` is still referenced in `spring.factories`. It has been removed in f033016364691e50877651116e22b352d275b937 as far as I can tell.

Cheers,
Christoph


snicoll:
Thanks again Christoph.

PR #18873, opened 2019-11-05
Let Liquibase smoke test pass on non-english systems

dreis2211:
Hi,

I noticed yesterday that the Liquibase smoke test failed on my machine. After having a quick look this is caused by a localized string that breaks the output assertion. On non-english systems it is therefore not possible to run the full smoke-test suite.

This PR simply sets `Locale.ENGLISH` for the test in question and resets it afterwards.

I didn't feel comfortable with changing the assertion, because it covers the whole lifespan of the liquibase changelog process currently. Cutting it down to check less stuff didn't seem like the appropriate thing to do, but I'm happy to do it if you think that's favorable.

Cheers,
Christoph


snicoll:
Thanks again Christhoph

PR #18883, opened 2019-11-05
Use try-with-resources blocks in JarFileArchiveTests

izeye:
This PR changes to use try-with-resources blocks in `JarFileArchiveTests`.


snicoll:
Thanks again @izeye

PR #18915, opened 2019-11-07
Handle JSON keys containing a dot from CF environment as a single path segment

schulzh:
This PR changes the handling of JSON keys in the CloudFoundry VCAP_SERVICES environment variable so that they represent only a single path segment at a time.

Before this change, if the JSON contained an object key with dots, the key would be treated as individual path segments, instead of a single one.
E.g. the VCAP_SERVICES variable
```json
{
  "user-provided": [
    {
      "name": "test",
      "label": "test-label",
      "credentials": {
        "key.with.dots": {
          "username": "foo",
          "password": "bar"
        }
      }
    }
  ]
}
```
Would lead to the spring property:
`vcap.services.test.credentials.key.with.dots.username=foo`
Instead of the escaped version:
`vcap.services.test.credentials[key.with.dots].username=foo`

This causes problems e.g. when trying to bind these properties to a map, as it would not be filled with any values:
```java
@ConfigurationProperties("vcap.services.test")
public class CfEnvCredentials {
    public final Map<String, BasicAuthentication> credentials;
    [...]
}
```

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudFoundryVcapEnvironmentPostProcessor.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudFoundryVcapEnvironmentPostProcessor.java
@@ -224,6 +224,9 @@ public class CloudFoundryVcapEnvironmentPostProcessor implements EnvironmentPost
 		if (key.startsWith("[")) {
 			return path + key;
 		}
+		if (key.contains(".")) {
+			return path + "[" + key + "]";
+		}
 		return path + "." + key;
 	}
 
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/cloud/cloudfoundry/CloudFoundryVcapEnvironmentPostProcessorTests.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/cloud/cloudfoundry/CloudFoundryVcapEnvironmentPostProcessorTests.java
@@ -116,6 +116,16 @@ class CloudFoundryVcapEnvironmentPostProcessorTests {
 		assertThat(getProperty("vcap.services.mysql.credentials.port")).isEqualTo("3306");
 	}
 
+	@Test
+	void testServicePropertiesContainingKeysWithDot() {
+		TestPropertySourceUtils.addInlinedPropertiesToEnvironment(this.context,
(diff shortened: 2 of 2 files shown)


pivotal-issuemaster:
@schulzh Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=18915)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=18915) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@schulzh Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=18915)!

philwebb:
@schulzh What sort of generic types do you have on your Map? We should automatically deal with this for `String` and scalar types (see [this test](https://github.com/spring-projects/spring-boot/blob/a66c4d3096877b33119366ea9ea165c892b462c8/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/bind/MapBinderTests.java#L279-L291)). Do you have a small sample that show the problem that lead to you creating this PR?

schulzh:
I have a `Map<String, BasicAuthentication>`, with this BasicAuthentication class:
```
public class BasicAuthentication extends Authentication {
    public final String username;
    public final String password;
}
```
This is meant to supply Authentication information for arbitrary hosts, which will be the keys of the map, therefore the dots in the keys.

I just tested a `Map<String, String>` example and indeed, it worked. Sorry, I should have tested the actual example before. I'll edit the PR with the actual example.

snicoll:
@schulzh would you have time to revisit the PR based on Madura's feedback?

mbhave:
@schulzh Thank you for making your first contribution to Spring Boot. This has now been merged along with [this addition](https://github.com/spring-projects/spring-boot/commit/544dca7f1b8a2adf7921f7c56b86f7abb8f37964).

wilkinsona:
Unfortunately, this change caused some unanticipated problems and we're going to have to revert it. Please see https://github.com/spring-projects/spring-boot/issues/20343 for details. We've opened https://github.com/spring-projects/spring-boot/issues/20400 to take another look at the problem. It's tentatively scheduled for 2.3.

PR #18955, opened 2019-11-11
Polish

thelproad:


snicoll:
@thelproad thanks for the PR but rather than removing the `adapter`, I prefer to fix the javadoc link. I did that in https://github.com/spring-projects/spring-boot/commit/cad236523718da427961d129f63a9fad2de09e3b

Issue #18959, opened 2019-11-11
Setting health endpoint cache time-to-live does not actually use CachingOperationInvoker


temesoft:
When setting `management.endpoint.health.cache.time-to-live=60s` configuration to specify caching of health endpoint response - it seems not to take an effect in 2.2.0 release version. When requesting health endpoint `ReflectiveOperationInvoker` is being called instead of `CachingOperationInvoker`.

Attached, is a sample spring-boot application with a JUnit test that shows this issue.

[my-app.zip](https://github.com/spring-projects/spring-boot/files/3833480/my-app.zip)

P.S. When switching to use an older version (for example 2.0.3) the health endpoint response is cached correctly.

Thank you
-Dmitriy

dreis2211:
Hi @temesoft . I think I'm on to something. Not a team member, but give me until this morning to try out my idea.

dreis2211:
So here's what I found so far. The problem comes from the fact that `ApiVersion` is treated as a mandatory parameter in `CachingOperationInvokerAdvisor` and thus breaks the caching. One could simply handle `ApiVersion` the same way `SecurityContext` is handled, but that would probably break return wrong results if request A has `ApiVersion.V3` specified and request B has `ApiVersion.V2`. 

This was likely introduced with #17929.

snicoll:
Thank you both. We'll have a look to handle `ApiVersion` so that it doesn't break caching.

dreis2211:
Made a proposal fix in #18961

snicoll:
Closing in favour or PR #18961

PR #18961, opened 2019-11-12
Handle ApiVersion in CachingOperationInvoker

dreis2211:
Hi,

this is a proposal for fixing #18959.

Prior to this commit, `ApiVersion` was treated as a mandatory parameter in `CachingOperationInvokerAdvisor` and thus prevented the `CachingOperationInvoker` to kick in. By skipping `ApiVersion` in the same way we're skipping `SecurityContext` we can avoid this.

In order to not return the same cached response, this commit also changes the cache handling in `CachingOperationInvoker` to account for different `ApiVersion`s being passed.

Let me know what you think.
Cheers,
Christoph


dreis2211:
Btw. I'm having trouble to see the concourse builds at the moment (it redirects me to a blank page). Might be related to the build-pipeline rename, but I thought I mention it at least ;-)

wilkinsona:
Thanks, @dreis2211. Could you try again now please?

dreis2211:
Works like a charm, @wilkinsona . Thanks 👍

snicoll:
Thanks again Christoph!

Issue #18681, opened 2019-10-21
Make Kubernetes detection slightly more efficient


dsyer:
In #15537 we implemented a kubernetes detector based on an env var pattern. But 99.99% of the time `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT` will be defined, and they can be looked up quickly without looping around the enumerable property source. The negative case, unfortunately, remains inefficient.

dsyer:
https://stackoverflow.com/a/54130803/1368236

I like this one:

```
if File.exists?('/var/run/secrets/kubernetes.io')
  puts "I'm also running in a Kubernetes pod"
end
```

That's cheap and efficient.
Probably. I guess if `/var/run/secrets` exists and is network attached it might be slow.

philwebb:
Thomas Risberg also suggested `/var/run/secrets/kubernetes.io/serviceaccount/namespace` that even gives you the namespace

https://kubernetes.io/docs/tasks/access-application-cluster/access-cluster/#accessing-the-api-from-a-pod

mbhave:
Closing in favor of PR #19002.

PR #19002, opened 2019-11-13
Make Kubernetes detection slightly more efficient

medamines1:
Hi,
This is a proposal for fixing #18681
A slite change of the way to detect Kubernetes environment.

Let me know what you think.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
@@ -16,6 +16,8 @@
 
 package org.springframework.boot.cloud;
 
+import java.io.File;
+
 import org.springframework.core.env.ConfigurableEnvironment;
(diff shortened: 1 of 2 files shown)


mbhave:
Thanks for the PR @medamines1. I don't think that's quite what Dave was suggesting [here](https://github.com/spring-projects/spring-boot/issues/18681#issuecomment-544701739). I think the order of checks would be

1. Check for the specific environment variables first: `KUBERNETES_SERVICE_HOST` and `KUBERNETES_SERVICE_PORT`.
2. If those aren't present, check for the secrets file
3. If the secret file isn't found, try the current check we have.

Do you want to update the PR to make those changes?

medamines1:
yes, I'll make the changes and I'll update the PR as soon as possible.

snicoll:
@medamines1 thank you for making your first contribution to Spring Boot. Given the potential side effect when testing an hardcoded file path, I've [removed this optimization and also made sure that only environment variables are queried](https://github.com/spring-projects/spring-boot/commit/ddb5cc3f9988bf7540de1a722f3022d865fa931d).

PR #19027, opened 2019-11-15
Use DeferredLog in CloudFoundryVcapEnvironmentPostProcessor

markbigler:
If I'm not wrong, currently nothing will be logged because the logging system has not yet been initialized when the EnvironmentPostProcessor's are invoked.

This PR makes use of the DeferredLog, so the log entries are no longer lost and become visible as soon as the logging system is ready.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudFoundryVcapEnvironmentPostProcessor.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudFoundryVcapEnvironmentPostProcessor.java
@@ -29,7 +29,6 @@ import org.springframework.boot.env.EnvironmentPostProcessor;
 import org.springframework.boot.json.JsonParser;
 import org.springframework.boot.json.JsonParserFactory;
 import org.springframework.boot.logging.DeferredLog;
-import org.springframework.context.ApplicationEvent;
 import org.springframework.context.ApplicationListener;
 import org.springframework.core.Ordered;
 import org.springframework.core.env.CommandLinePropertySource;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudFoundryVcapEnvironmentPostProcessor.java
```diff
@@ -89,9 +90,9 @@
  * @author Andy Wilkinson
  * @since 1.3.0
  */
-public class CloudFoundryVcapEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
+public class CloudFoundryVcapEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered, ApplicationListener<ApplicationEvent> {
```
snicoll: You could implement `ApplicationListener<ApplicationPreparedEvent>` instead and avoid the if check in the listener implementation.

markbigler: Thank you for pointing that out. Fixed with https://github.com/spring-projects/spring-boot/pull/19027/commits/a9c1e031e8c64e7807964bc2c6f0fdc9e62515f8

wilkinsona:
This seems to me like something that we should fix in 2.1.x. The fix doesn't look risky so I suspect I've missed something. What's making you unsure, @snicoll?

snicoll:
I can’t remember now. Thanks for confirming!

snicoll:
@markbigler thank you for making your first contribution to Spring Boot. This is merged in `2.1.x` as well as `2.2.x` and `master`.

PR #19028, opened 2019-11-15
Fix syntax error in configuration metadata sample in docs

dreis2211:
Hi,

this PR fixes a syntax error in a sample in the [Configuration Metadata section](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/html/appendix-configuration-metadata.html#configuration-metadata-annotation-processor) and does some related polishing.

Cheers,
Christoph


snicoll:
Good catch, thanks again!

PR #19029, opened 2019-11-15
Use CAPS for build args in Dockerfile example

dsyer:
Also provide a default value (for convenience)


snicoll:
Thanks @dsyer. I've updated the example that was still referring to the old variable. Also, we prefer to have 3 lines between sections.

PR #19039, opened 2019-11-18
Polish AbstractDependsOnBeanFactoryPostProcessor

nosan:
Polish AbstractDependsOnBeanFactoryPostProcessor to avoid using `try {} catch` block

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AbstractDependsOnBeanFactoryPostProcessor.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AbstractDependsOnBeanFactoryPostProcessor.java
@@ -134,16 +134,14 @@ public abstract class AbstractDependsOnBeanFactoryPostProcessor implements BeanF
 	}
 
 	private static BeanDefinition getBeanDefinition(String beanName, ConfigurableListableBeanFactory beanFactory) {
-		try {
+		if (beanFactory.containsBeanDefinition(beanName)) {
 			return beanFactory.getBeanDefinition(beanName);
 		}
-		catch (NoSuchBeanDefinitionException ex) {
-			BeanFactory parentBeanFactory = beanFactory.getParentBeanFactory();
-			if (parentBeanFactory instanceof ConfigurableListableBeanFactory) {
-				return getBeanDefinition(beanName, (ConfigurableListableBeanFactory) parentBeanFactory);
-			}
-			throw ex;
+		BeanFactory parentBeanFactory = beanFactory.getParentBeanFactory();
+		if (parentBeanFactory instanceof ConfigurableListableBeanFactory) {
+			return getBeanDefinition(beanName, ((ConfigurableListableBeanFactory) parentBeanFactory));
 		}
+		throw new NoSuchBeanDefinitionException(beanName);
 	}
 
 }
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/AbstractDependsOnBeanFactoryPostProcessorTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/AbstractDependsOnBeanFactoryPostProcessorTests.java
@@ -42,6 +42,14 @@ class AbstractDependsOnBeanFactoryPostProcessorTests {
 			.withUserConfiguration(FooBarConfiguration.class);
 
 	@Test
+	void throwNoSuchBeanDefinitionSingletonBean() {
+		this.contextRunner.withUserConfiguration(FooDependsOnBarTypePostProcessor.class)
+				.withInitializer((context) -> context.getBeanFactory().registerSingleton("fooSingleton", new Foo()))
+				.run((context) -> assertThat(context).hasFailed().getFailure()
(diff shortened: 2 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AbstractDependsOnBeanFactoryPostProcessor.java
```diff
@@ -134,16 +134,14 @@ public int getOrder() {
...
-			BeanFactory parentBeanFactory = beanFactory.getParentBeanFactory();
-			if (parentBeanFactory instanceof ConfigurableListableBeanFactory) {
-				return getBeanDefinition(beanName, (ConfigurableListableBeanFactory) parentBeanFactory);
-			}
-			throw ex;
+		BeanFactory parentBeanFactory = beanFactory.getParentBeanFactory();
+		if (parentBeanFactory instanceof ConfigurableListableBeanFactory) {
+			return getBeanDefinition(beanName, ((ConfigurableListableBeanFactory) parentBeanFactory));
 		}
+		throw new NoSuchBeanDefinitionException(beanName);
```
snicoll: I am not sure what you're trying to achieve here. The context may provide an exception with additional context it has when it throws the exception (besides the `beanName`) so I am not in favour of creating such low-level exception here.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/AbstractDependsOnBeanFactoryPostProcessorTests.java
```diff
@@ -41,6 +41,14 @@
 	private ApplicationContextRunner contextRunner = new ApplicationContextRunner()
 			.withUserConfiguration(FooBarConfiguration.class);
 
+	@Test
+	void throwNoSuchBeanDefinitionSingletonBean() {
```
snicoll: Is that test failing without this change?

nosan: Programmatically registered singleton does not have a `BeanDefinition`, hence test will fail independently of implementation.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AbstractDependsOnBeanFactoryPostProcessor.java
nosan: This exception is being thrown only in case when `BeanFactory` and its parent `BeanFactory` does not contain the provided `beanName`.

nosan:
Hi @snicoll,
Thank you very much for your feedback.
The idea of this PR is avoiding unnecessary `try { } catch` block and don't rely on exception throwing.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AbstractDependsOnBeanFactoryPostProcessor.java
snicoll: My point exactly. It's going to be the only case where it's thrown but the framework will not have created the instance, we would. And I am not keen to do that as explained in my first comment, makes sense?

nosan: I understood but in that case, this PR becomes meaningless and it should be closed 😄

nosan:
Maybe it makes sense to add support for `Programmatically registered singleton`.
Currently, it is not supported by `AbstractDependsOnBeanFactoryPostProcessor`.

snicoll:
@nosan you lost me I am afraid. We went from polishing to something that's not supported. I guess that's what I was trying to get at with https://github.com/spring-projects/spring-boot/pull/19039#discussion_r347849409

nosan:
You are right @snicoll,

Sorry about my last comment, it seems I have mixed two different things in one PR.

The initial request was about avoiding usage of the try-catch block and don't rely on `NoSuchBeanDefinitionException`. 
My concern that currently instead of use a `containsBeanDefinition + getBeanDefinition`, only `getBeanDefinition` is used. Of course, it is quite a rare case when BeanFactory has more than one parent in the hierarchy, and probably there is no performance issue here, but I feel that this PR makes this part more readable and understandable.

For instance:
```
#1
	private static String getText() {
		Map<String, Object> maps = new LinkedHashMap<>();
		try {
			return maps.get("key").toString();
		}
		catch (NullPointerException ex) {
			return "";
		}
	}

#2
	private static String getText() {
		Map<String, Object> maps = new LinkedHashMap<>();
		Object val = maps.get("key");
		return (val != null) ? val.toString() : "";
	}

```

I prefer the second approach, and I want to achieve something similar here.

snicoll:
Thanks for the PR @nosan. We've discussed this at the team meeting and while removing flow control in exception handling brings is the right thing to do, we're not keen to create the low-level framework exception ourselves as I've mentioned in the review.

There might be another way however, I've created #19164 to investigate that.

PR #19044, opened 2019-11-19
Support tc datasource prefix for test containers

polarbear567:

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
@@ -196,7 +196,12 @@ public enum DatabaseDriver {
 			return Arrays.asList("informix-sqli", "informix-direct");
 		}
 
-	};
+	},
+
+	/**
+	 * Testcontainers.
+	 */
+	TC("Testcontainers", "org.testcontainers.jdbc.ContainerDatabaseDriver");
 
 	private final String productName;
 
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
@@ -201,7 +201,14 @@ public enum DatabaseDriver {
 	/**
 	 * Testcontainers.
 	 */
-	TC("Testcontainers", "org.testcontainers.jdbc.ContainerDatabaseDriver");
+	TESTCONTAINERS("Testcontainers", "org.testcontainers.jdbc.ContainerDatabaseDriver") {
+
+		@Override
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
```diff
@@ -196,6 +196,18 @@ protected boolean matchProductName(String productName) {
 			return Arrays.asList("informix-sqli", "informix-direct");
 		}
 
+	},
+
+	/**
+	 * Testcontainers.
+	 */
+	TESTCONTAINERS("Testcontainers", "org.testcontainers.jdbc.ContainerDatabaseDriver") {
```
snicoll: is `Testcontainers` the product name of this driver? The way I understood it was that it is a wrapper for the actual underlying database container

bsideup: FYI the driver **does not** return a product name property from `getPropertyInfo`.

I would highly recommend to write a test that checks it with the TC's driver (no need to start a container, just the lookup part), especially given that it is already on classpath

polarbear567: Hi, @snicoll , what should I do?

snicoll: In the conversation above we've identified that Test Containers does not have such product name so the current PR is wrong. Also @bsideup was suggesting to write a test for the lookup part. Would you be willing to rework the PR to that effect please?

polarbear567: Sure, but I hava some questions.
Assuming that the current testcontainers is a pair of mysql packaging, if we dynamically match the produceName to mysql, will it cause confusion? Testcontainers itself does not seem to rename the driver for its packaging.

snicoll: Sorry but I don't understand the question. The bottom line here is that test containers does not have a product name and as such the first argument of that enum should be `null`. Tests should be updated accordingly.

polarbear567: I thought you mean that there must be a product name for Testcontainers. I've changed it to null now. For test, I'm not sure what @bsideup  wants to express, so the lookup part I wrote for test case is for "fromJdbcUrl" method. If there are other changes or suggestions about test, please let me know. Thanks!

snicoll:
Thanks @polarbear567, rather than this back and forth on the test, I've polished your proposal with the additional test, see https://github.com/spring-projects/spring-boot/commit/4c7896b936c5933e43f32208d7aa35595e9049ee

PR #19057, opened 2019-11-19
Incorrect Javadoc regarding default Tomcat async request timeout

### LINKED ISSUE #19056: Incorrect Javadoc regarding default Tomcat async request timeout
Follow-up on https://github.com/spring-projects/spring-framework/issues/24030

In `org.springframework.boot.autoconfigure.web.servlet.WebMvcProperties.Async#requestTimeout` the Javadoc says

```java
/**
 * Amount of time before asynchronous request handling times out. If this value is
 * not set, the default timeout of the underlying implementation is used, e.g. 10
 * seconds on Tomcat with Servlet 3.
 */
```

However, actually the default timeout in Tomcat is 30 seconds as noted in this comment: https://github.com/apache/tomcat/blob/master/java/org/apache/catalina/core/AsyncContextImpl.java#L74 and indeed set to the default 30000 ms here: https://github.com/apache/tomcat/blob/master/java/org/apache/catalina/connector/Connector.java#L134

The incorrect explicit values were removed from Spring WebMvc Javadocs and so should the Javadoc in Spring Boot.

adammichalik:
Fix #19056 - remove incorrect Javadoc part

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcProperties.java
@@ -220,8 +220,7 @@ public class WebMvcProperties {
 
 		/**
 		 * Amount of time before asynchronous request handling times out. If this value is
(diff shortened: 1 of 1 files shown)


snicoll:
@adammichalik thank you for the PR.

PR #19094, opened 2019-11-21
Fix StepVerifier usage in HealthIndicatorReactiveAdapterTests

dreis2211:
Hi,

just found a `StepVerifier` that doesn't call `verify()`. In the given test one could change to any exception being thrown and the verifier wouldn't notice.

Cheers,
Christoph


bclozel:
I believe IDEA is now warning us about some user errors with Reactor in general, that should help in the future.
Thanks for spotting this!

PR #19099, opened 2019-11-22
Configuration properties that use the builder pattern are not recognized when used with a generic

ChunMengLu:
I am writing a redisson starter. For the sake of simplicity, I have used the following this:

```java
@ConfigurationProperties(prefix = "redisson")
public class RedissonProperties {

	@NestedConfigurationProperty
	private ClusterServersConfig cluster = new ClusterServersConfig();

	public ClusterServersConfig getCluster() {
		return cluster;
	}

	public void setCluster(ClusterServersConfig cluster) {
		this.cluster = cluster;
	}
}
```

I found that the generated spring-configuration-metadata.json does not have [ClusterServersConfig](https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/config/ClusterServersConfig.java) parent [BaseMasterSlaveServersConfig](https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/config/BaseMasterSlaveServersConfig.java) and [BaseConfig](https://github.com/redisson/redisson/blob/master/redisson/src/main/java/org/redisson/config/BaseConfig.java) props!

```java
public class BaseMasterSlaveServersConfig<T extends BaseMasterSlaveServersConfig<T>> extends BaseConfig<T> {
    private ReadMode readMode = ReadMode.SLAVE;
	
    public T setReadMode(ReadMode readMode) {
        this.readMode = readMode;
        return (T) this;
    }
    public ReadMode getReadMode() {
        return readMode;
    }
}
```

setReadMode method returnType is generic type T.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeElementMembers.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeElementMembers.java
@@ -28,6 +28,7 @@ import javax.lang.model.element.ExecutableElement;
 import javax.lang.model.element.Modifier;
 import javax.lang.model.element.TypeElement;
 import javax.lang.model.element.VariableElement;
+import javax.lang.model.type.DeclaredType;
 import javax.lang.model.type.TypeKind;
 import javax.lang.model.type.TypeMirror;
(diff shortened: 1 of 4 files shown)


snicoll:
@ChunMengLu good catch and thank you for making your first contribution to Spring Boot. I've decided to take a different approach, reusing the infrastructure we already have to resolve generic types. Right now we do a type FQN name equality check which is something I would have liked to avoid but I don't want to invest into a too much sotipshicaated solution for this either.

PR #19106, opened 2019-11-23
Add configuration property for RabbitMQ client requestedChannelMax

fzilic:
According to changes in RabbitMQ Java client library 4.7.0 and 5.3.0 com.rabbitmq.client.ConnectionFactory sets its property requestedChannelMax to 2047 to align itself with changes to RabbitMQ server.

Spring Boot autoconfiguration for RabbitMQ does not expose any configuration properties that would allow developers to change this limit if required by their project.

Limit of 2047 channels per connection is reasonable, but in some cases changing this is required.

Logic for determining negotiated channel_max limit for RabbitMQ is as follows:
- if channel_max >0 (server) and channel_max > 0 (clinet) take smaller version
- if any channel_max is 0, and either server or client have it to non zero, take non zero value
- if all channel_max are 0, set to unlimited

This is negotiated on RabbitMQ server and would not impact existing users.

GitHub issue for RabbitMQ Java client
rabbitmq/rabbitmq-java-client#366

And RabbitMQ server GitHub issue
rabbitmq/rabbitmq-server#1593

Working on pull request.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
@@ -104,8 +104,9 @@ public class RabbitProperties {
 	private Duration connectionTimeout;
 
 	/**
-	 * Requested Channel Max. Set it to zero for unlimited. Default is 2047 from RabbitMQ
-	 * java client versions 4.7.0 and 5.3.0.
+	 * Requested Channel Max; zero for unlimited. Number of channels per connection client
+	 * will request from server, actual maximum will be negotiated between client and
+	 * server for lowest value (excluding zero as it represents unlimited).
 	 */
 	private int requestedChannelMax = 2047;
 
@@ -305,10 +306,6 @@ public class RabbitProperties {
 		return this.connectionTimeout;
 	}
 
-	public int getRequestedChannelMax() {
-		return this.requestedChannelMax;
-	}
-
 	public void setPublisherConfirmType(ConfirmType publisherConfirmType) {
 		this.publisherConfirmType = publisherConfirmType;
 	}
@@ -321,7 +318,11 @@ public class RabbitProperties {
 		this.connectionTimeout = connectionTimeout;
 	}
 
-	public void setConnectionTimeout(int requestedChannelMax) {
+	public int getRequestedChannelMax() {
+		return this.requestedChannelMax;
+	}
+
+	public void setRequestedChannelMax(int requestedChannelMax) {
 		this.requestedChannelMax = requestedChannelMax;
 	}
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
@@ -40,6 +40,7 @@ import org.springframework.util.StringUtils;
  * @author Josh Thornhill
  * @author Gary Russell
  * @author Artsiom Yudovin
+ * @author Franjo Zilic
  * @since 1.0.0
  */
 @ConfigurationProperties(prefix = "spring.rabbitmq")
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -310,6 +321,10 @@ public void setConnectionTimeout(Duration connectionTimeout) {
 		this.connectionTimeout = connectionTimeout;
 	}
 
+	public void setConnectionTimeout(int requestedChannelMax) {
```
snicoll: That doesn't sound right.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -102,6 +103,12 @@
 	 */
 	private Duration connectionTimeout;
 
+	/**
+	 * Requested Channel Max. Set it to zero for unlimited. Default is 2047 from RabbitMQ
+	 * java client versions 4.7.0 and 5.3.0.
+	 */
+	private int requestedChannelMax = 2047;
```
snicoll: The default value is already documented so it shouldn't be documented again. Can you please rephrase that to rather explain in one sentence what requestedChannelMax actually is?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -298,6 +305,10 @@ public Duration getConnectionTimeout() {
 		return this.connectionTimeout;
 	}
 
+	public int getRequestedChannelMax() {
```
snicoll: Getter/Setter should be co-located.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
fzilic: More detailed description is now in updated PR, thanks for suggestion.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
fzilic: Thank you, it was an honest mistake to make after several hours of late night troubleshooting why is 2047 still limit for number of channels on production even after changing server configuration to unlimited.

Fixed in updated PR, thanks for suggestion.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
fzilic: Attempt was made not to mess up existing ordering at the end of RabbitProperties, but I was not able to see this type of code convention, since getConnectionTimeout and setConnectionTimout are not co-located. Updated PR with changes

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
garyrussell: It would be better to change this to `Integer` (default `null`) and add `.whenNonNull()` to the mapping; that way, if they decide to change the default in the client library, we won't need a corresponding change here.

fzilic: Thanks for suggestion, when it is set to null in RabbitProperties object it does not impact actual default in com.rabbitmq.client.ConnectionFactory, added test to verify that and updated PR.

snicoll:
@fzilic thank you for making your first contribution to Spring Boot. 

@garyrussell I've decided to keep an `int`, testing the default value matches the expectation of the driver. That felt more consistent with other properties where we've also documented the default. If you feel that's not a good idea even with a test, we can reconsider.

garyrussell:
I have no strong objection, it just seems simpler to me that "undefined" means use the library's default. But, it's perhaps unlikely they will change this anyway.

fzilic:
Thanks for accepting PR. It is a product of 4 late night hours of troubleshooting during production deploy.

Issue #19102, opened 2019-11-22
ssl.enabled is ignored when Rabbit is configured with an address with no protocol


wilkinsona:

mbhave:
We should also determine the port based on `ssl.enabled` if no protocol and port is configured in the address. See [this](https://github.com/spring-projects/spring-boot/blob/6ac05539fc8495f7880f1c3dcb2564059a1c2297/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java#L1043).

snicoll:
Closing in favour of PR #19109

PR #19109, opened 2019-11-25
ssl.enabled is ignored when Rabbit is configured with an address with no protocol 

bono007:
Use ssl enabled flag when RabbitProperties specify address with no protocol.

Closes gh-19102


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -188,7 +188,7 @@ public void setAddresses(String addresses) {
 	private List<Address> parseAddresses(String addresses) {
 		List<Address> parsedAddresses = new ArrayList<>();
 		for (String address : StringUtils.commaDelimitedListToStringArray(addresses)) {
-			parsedAddresses.add(new Address(address));
+			parsedAddresses.add(new Address(address, getSsl().isEnabled()));
```
bono007: Not super keen on passing this info down into Address#init but doing so make all downstream processing in Address hinge off of `Address.secureConnection`.

bono007:
@wilkinsona I think this addresses the concerns.

snicoll:
@bono007 thank you for making your first contribution to Spring Boot. Agreed on that ssl flag, I've tried a few options and this one is the least bad option. I've moved its processing in a different place though, making `secureConnection` a `Boolean`.

Issue #19038, opened 2019-11-17
Allow to use Testcontainers with @AutoConfigureTestDatabase 


snicoll:
Follow-up of [this limitation reported on SO](https://stackoverflow.com/questions/58845698/can-you-set-the-compatibility-mode-on-the-autoconfigured-spring-boot-h2-test-db/58852607#58852607), it would be nice if users were able to configure the URL that `@AutoConfigureTestDatabase` uses. Right now if they want a specific dialect or any other extra parameter, they have to switch off our support to configure things manually.

snicoll:
We've spent some time discussing this feature and we're willing to investigate a support of testcontainers rather which would provide all the necessary flexibility.

snicoll:
Now that we have r2dbc support, we should also take that use case into account so that the `ConectionFactory` points to a database replaced via this mechanism.

wilkinsona:
I've been experimenting a little bit in this area. I have a test class like this:

```
@JdbcTest
@AutoConfigureTestDatabase(replace = Replace.NONE)
@Testcontainers(disabledWithoutDocker = true)
@Import(SomeRepository.class)
@Sql(scripts = "data.sql")
class SomeRepositoryTests {

	@Container
	private static final PostgreSQLContainer<?> postgresqlContainer = new PostgreSQLContainer<>("postgres:13.2")
			.withDatabaseName("example");

	@Autowired
	private SomeRepository someRepository;

        // @Test methods
```

I've then got a context customizer factory and a context customizer that turns each `JdbcDatabaseContainer` field into a `DataSource` bean:

```java
/**
 * A {@link ContextCustomizerFactory} that returns a
 * {@link JdbcDatabaseContainerContextCustomizer} when {@link JdbcDatabaseContainer} is on
 * the classpath and the test class has one or more static fields of that type.
 *
 * @author Andy Wilkinson
 */
class JdbcDatabaseContainerContextCustomizerFactory implements ContextCustomizerFactory {

	@Override
	public ContextCustomizer createContextCustomizer(Class<?> testClass,
			List<ContextConfigurationAttributes> configAttributes) {
		if (ClassUtils.isPresent("org.testcontainers.containers.JdbcDatabaseContainer", testClass.getClassLoader())) {
			Set<Field> jdbcDatabaseContainerFields = new HashSet<>();
			ReflectionUtils.doWithFields(testClass, (field) -> {
				if (Modifier.isStatic(field.getModifiers())
[...]

PR #19166, opened 2019-11-27
Allow custom url in @AutoConfigureTestDatabase

bono007:
This allows a custom URL to specified in `@AutoConfigureTestDatabase`.

### It was not straightforward:

We leverage `org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseBuilder` in `org.springframework.boot.test.autoconfigure.jdbc.TestDatabaseAutoConfiguration` which ultimately delegates to 
```java
org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseFactory 
-> 
org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseConfigurer#configureConnectionProperties
```
which in turn each configurer impl has the hardcoded URL it uses and sets the properties on the datasource. It is these configurers that have the last say in what the url is. 

The `EmbeddedDatabaseBuilder` does not expose a setter for its `EmbeddedDatabaseFactory` so getting a "custom url respecting" configurer in this call chain is not straightforward and may not be possible w/o modification to Spring Framework `EmbeddedDatabaseBuilder`.
Those configurers are all package protected as well so copying some code was not an option either. 

### Another note:
The hardcoding is in multiple places, the `EmbeddedDatabaseConnection` and in `EmbeddedDatabaseConfigurer` impls.  For the purposes of this change I believe its fine just modifying the latter as the former is not used by any test code AFAICT.

### DIFF
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/AutoConfigureTestDatabase.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/AutoConfigureTestDatabase.java
@@ -61,6 +61,13 @@ public @interface AutoConfigureTestDatabase {
 	EmbeddedDatabaseConnection connection() default EmbeddedDatabaseConnection.NONE;
 
 	/**
+	 * The custom connection URL options to use. If specified it will replace the options
+	 * on the default URL string in the embedded datasource.
+	 */
+	@PropertyMapping(skip = SkipPropertyMapping.ON_DEFAULT_VALUE)
+	String urlOptions() default "";
+
+	/**
 	 * What the test database can replace.
 	 */
 	enum Replace {
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfigurationTests.java
@@ -25,6 +25,7 @@ import org.springframework.boot.test.context.runner.ApplicationContextRunner;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.jdbc.core.JdbcTemplate;
+import org.springframework.jdbc.datasource.SimpleDriverDataSource;
 import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseBuilder;
 import org.springframework.jdbc.datasource.embedded.EmbeddedDatabaseType;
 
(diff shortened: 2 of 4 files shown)


spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/AutoConfigureTestDatabase.java
```diff
@@ -60,6 +60,15 @@
 	 */
 	EmbeddedDatabaseConnection connection() default EmbeddedDatabaseConnection.NONE;
 
+	/**
+	 * The custom connection URL to use. If specified it will override the default URL
+	 * string in the embedded datasource. It is expected to be fully specified and no
```
snicoll: I am not sure this brings us much to mention that. One might argue that using its own URL means that another feature is lost (unique names) so perhaps we should do something about it.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/AutoConfigureTestDatabase.java
```diff
@@ -60,6 +60,15 @@
...
+	 * string format replacements will be performed.
+	 * @return the custom connection url to use
+	 */
+	@PropertyMapping(skip = SkipPropertyMapping.ON_DEFAULT_VALUE)
+	String url() default "";
```
snicoll: I don't have any bright idea right now but this introduces a potential inconsistency combination where you could specify a connection type for HSQL and then an URL for H2. I wonder if there's a way to structure the config to avoid that.

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfigurationTests.java
```diff
@@ -60,6 +60,29 @@ void replaceWithUniqueDatabase() {
 		});
 	}
 
+	@Test
+	void replaceAnyWithCustomUrl() {
+		// There is a user configured HSQL we will replace with H2 using a custom url
```
snicoll: We don't put comment in our test but rather structure our test method name with additional context. Can you please remove those comments?

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/jdbc/AutoConfigureTestDatabase.java
bono007: I am going to explore adding a `urlOptions` instead of a full url. I think that would satisfy the original "ask" as well as limit the possible inconsistency, as well as not lose the other features like auto names.

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfigurationTests.java
bono007: I copied that from another existing test but i agree and will rename and remove.

bono007:
@snicoll 


Instead of replacing the url entirely, I renamed the attribute to `urlOptions` and if specified it will then replace the options on the connection url string. 

Example:
```java
jdbc:h2:mem:7a5cf237-9160-40ba-b633-fda0ede74341;DB_CLOSE_DELAY=-1;DB_CLOSE_ON_EXIT=false
```
with `urlOptions=MODE=MySQL` will result in 

```java
jdbc:h2:mem:7a5cf237-9160-40ba-b633-fda0ede74341;MODE=MySQL
```

This will allow the connection attribute to drive the base url and allow the user to override the connection options. I realize it may be nicer to override/append rather than replace but that got messy quickly. If you feel strongly about it I can add that back in. Otherwise, if the user is specifying any `urlOptions` they need to specify all of them. I think this is a fair tradeoff.

bono007:
Just pinging this one - I am guessing the team is just busy.

philwebb:
See #19038

snicoll:
Thanks for the PR @bono007. We've discussed this request and implementing it has its quirks and compromises. We've investigated several options and none of them (including this one) are satisfactory.

Taking a step back, we'd like to revisit `@AutoConfigureTestDabase` entirely, I'll comment in that issue shortly. Thanks again.

bono007:
No worries @snicoll . Every time that I dig into something, even if it does not get merged, I get more familiar w/ the codebase - win/win. Also,I am excited about #19038 and support for TestContainers.

PR #19168, opened 2019-11-27
Use AdoptOpenJDK API V3 when detecting JDK updates

dreis2211:
Hi,

I just saw that there is an [OpenJDK API V3](https://github.com/AdoptOpenJDK/openjdk-api-v3) that deprecates V2. This PR updates the `detect-jdk-updates` script to make use of the new API.

Let me know what you think.
Cheers,
Christoph


snicoll:
Nice one, thank you very much again @dreis2211!

PR #19170, opened 2019-11-28
Add missing indentations for pom.xml in doc

izeye:
This PR adds missing indentations for `pom.xml` in doc.


wilkinsona:
Thanks, @izeye. The indention looks ok in the rendered docs, but we should swap out the spaces for tabs regardless.

PR #19189, opened 2019-11-29
Remove code that is deprecated since 2.1.x

dreis2211:
Hi,

as discussed in #19180 here's a PR that removes code that was deprecated in 2.1.x.

Cheers,
Christoph


wilkinsona:
Thanks very much once again, @dreis2211.

Issue #19194, opened 2019-11-29
Non-public property in Saml2RelyingPartyProperties' Registration class


sbattistin:
I was trying to use the properties set in the YAML file through the `Registration` class inside `Saml2RelyingPartyProperties` class.

The problem is that the method has only package visibility and is not public.

```
		public Signing getSigning() {
			return this.signing;
		}

-->		Identityprovider getIdentityprovider() {
			return this.identityprovider;
		}

		public static class Signing {
```

P.S. I'm not sure if there is a reason to not make this public.

sbattistin:
I opened this PR if needed: https://github.com/spring-projects/spring-boot/pull/19195

wilkinsona:
As noted on #19195, `@ConfigurationProperties` classes are not considered to be public API. We should use this issue to address the inconsistent visibility of the getters but the result may be that `getSigning()` is made package-private rather than `getIdentityprovider` being made public.

woodforda:
Hi, just a comment regarding use cases for using the internal config properties classes.  I too have stumbled yesterday into this visibility issue.  I have a use case where the powers decreed that all config must be in TypeSafe format (sadly) not yml/properties.    So I have an EnvironmentPostProcessor to add the property sources for the TypeSafe config in the correct order.  So far so good.   My issue comes when I try to let Boot Auto config properties.  Simple primitive bindings work ok.  However, trying to bind an object (like a list of registrations) doesn't work (I end p with a list of a map and later a class cast exception).  I therefore need to transform the properties to the config class in my bean config class.  Sadly, this is not possible due to the visibility.  So now I'm a bit stuck...

wilkinsona:
@woodforda Thanks for the additional use case. That sounds like something isn’t quite right in your environment post-processor. If you map the keys in the same way as Boot does for YAML or properties files, the binding should work without the need for any further transformation. I’d recommend comparing the contents of the property sources when you have equivalent TypeSafe or YAML/properties configuration to help to diagnose the problem.

I’ve opened https://github.com/spring-projects/spring-boot/issues/19199 to straighten out the contradictions in the documentation. Thanks for pointing it out.

woodforda:
[...]

PR #19195, opened 2019-11-29
Update visibility for IdentityProvider property

sbattistin:
Update visibility for IdentityProvider property inside Registration class for Saml2 relying party properties

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
@@ -59,7 +59,7 @@ public class Saml2RelyingPartyProperties {
 			return this.signing;
 		}
 
-		Identityprovider getIdentityprovider() {
+		public Identityprovider getIdentityprovider() {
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@sbattistin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19195)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19195) to manually synchronize the status of this Pull Request.

wilkinsona:
Thanks for the PR. As [noted in the documentation](https://docs.spring.io/spring-boot/docs/2.2.1.RELEASE/reference/htmlsingle/#boot-features-external-config-java-bean-binding),  `@ConfigurationProperties` classes available in Spring Boot are for internal use only. The properties that map to the class, which are configured via properties files, YAML files, environment variables etc., are public API but the content of the class itself is not meant to be used directly.

Given that, I don't think we necessarily want to make the method public if it does not need to be. There's a mixture of public and package-private methods at the moment. Let's use #19194 to address that. Thanks anyway.

pivotal-issuemaster:
@sbattistin Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19195)!

Issue #17944, opened 2019-08-23
server.ssl.key-alias is ignored when configuring Netty


wilkinsona:

bono007:
I am looking into this issue and should have something together for review in 1-2 days

wilkinsona:
Thanks for the PR, @bono007. Closing in favour of #19197.

Issue #16168, opened 2019-03-08
No error message when server.ssl.keyAlias doesn't match an entry


numsu:
Hello!

This is a feature request to handle an error.

I ran into an issue when updating my project from version 1.x.x to 2.x.x. Seems like the property `server.ssl.keyAlias` was turned case sensitive and the application couldn't load the key from the keystore on startup. The application is still able to startup with no error or warn log about it but clients cannot complete a handshake.

**Example:**
Keystore `keystore.jks` contains a private key entry with an alias `myprivatekey`
Set the properties
```
server:
  ssl:
    enabled: true
    key-store: keystore.jks
    key-store-password: changeit
    keyAlias: MyPrivateKey
```

**The symptoms:**
Chrome gives the following error: `ERR_SSL_VERSION_OR_CIPHER_MISMATCH`
openssl printout:

```
CONNECTED(00000003)
140162214049696:error:14077410:SSL routines:SSL23_GET_SERVER_HELLO:sslv3 alert handshake failure:s23_clnt.c:744:
---
no peer certificate available
---
No client certificate CA names sent
---
SSL handshake has read 7 bytes and written 247 bytes
---
New, (NONE), Cipher is (NONE)
Secure Renegotiation IS NOT supported
Compression: NONE
Expansion: NONE
---
```

**The resolution:**
The application shouldn't be able to start with SSL turned on if the alias specified doesn't match anything in the keystore provided.

Used Spring Boot version: 2.1.2.RELEASE

philwebb:
We currently apply the `server.ssl.key-alias` property to the underlying servlet container (see `org.springframework.boot.web.embedded.tomcat.SslConnectorCustomizer` for one example) and let it deal with the alias. I'm not sure why things have changed between Boot 1.x and 2.x, but I imagine it's likely to be because of a dependency upgrade.

@numsu What servlet container are you using?

numsu:
I believe that the application uses the netty container

wilkinsona:
All supported containers appear to behave consistency across versions. Tomcat is the only container that fails to start when the key alias point to a non-existent key:

_ | Jetty | Netty | Tomcat | Undertow |
--- |-------|  ----- | ------- | ------
**1.5.22** | Starts | N/A | Fails  | Starts
**2.1.7** | Starts | Starts | Fails | Starts

wilkinsona:
[...]

PR #19197, opened 2019-11-29
server.ssl.key-alias is ignored when configuring Netty

bono007:
Closes gh-17944

Followed similar approach as in `org.springframework.boot.web.embedded.jetty.SslServerCustomizer` by wrapping the returned key managers with a version that supports looking up by alias. It was one level deeper as Netty passes the KeyManagerFactory, not the KeyManagers down into the bowels of SSL. With this change, the KeyManagerFactory actually returns the wrapped KeyManagers via its SPI. 

Verified that an invalid alias name does the same exception at runtime as reported by #16168. They are in parity w/ this change.


wilkinsona:
@bono007 Thank you very much for making your first contribution to Spring Boot. The proposed changes have been merged into 2.1.x, 2.2.x, and master.

PR #19202, opened 2019-12-01
Verify ssl key alias on server startup

bono007:
If there is an invalid `ssl.server.key-alias` specified (one that does not exist in the keystore) then the server will not startup. 

Closes gh-16168

**NOTE** Code was added to Netty, Tomcat, and Jetty but not to Tomcat as it already fails to startup in this case. I was going to add the code anyways so they all failed in the same manner but it became problematic getting a key store prior to the server starting up. I went ahead and left it alone.

**NOTE** Integration test cover each server in this scenario and I also verified all 4 servers manually in another consuming app.

### DIFF
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/embedded/netty/NettyReactiveWebServerFactoryTests.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/embedded/netty/NettyReactiveWebServerFactoryTests.java
@@ -19,8 +19,6 @@ package org.springframework.boot.web.embedded.netty;
 import java.time.Duration;
 import java.util.Arrays;
 
-import javax.net.ssl.SSLHandshakeException;
-
 import org.junit.Test;
 import org.mockito.InOrder;
 import reactor.core.publisher.Mono;
@@ -99,14 +97,6 @@ public class NettyReactiveWebServerFactoryTests extends AbstractReactiveWebServe
 		Mono<String> result = testSslWithAlias("test-alias");
 		StepVerifier.setDefaultTimeout(Duration.ofSeconds(30));
 		StepVerifier.create(result).expectNext("Hello World").verifyComplete();
-	}
-
-	@Test
-	public void whenSslIsConfiguredWithAnInvalidAliasTheSslHandshakeFails() {
-		Mono<String> result = testSslWithAlias("test-alias-bad");
-		StepVerifier.setDefaultTimeout(Duration.ofSeconds(30));
-		StepVerifier.create(result).expectErrorMatches((throwable) -> throwable instanceof SSLHandshakeException
-				&& throwable.getMessage().contains("HANDSHAKE_FAILURE")).verify();
 	}
 
 	protected Mono<String> testSslWithAlias(String alias) {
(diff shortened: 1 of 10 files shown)


spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/server/SslUtilsTest.java
```diff
@@ -0,0 +1,83 @@
...
+ * Tests for {@link SslUtils}.
+ *
+ * @author Chris Bono
+ */
+
+public class SslUtilsTest {
+
+	@Test
+	public void assertStoreContainsAliasPassesWhenAliasFound() throws KeyStoreException {
+		KeyStore keyStore = mock(KeyStore.class);
```
bono007: So I chose to go w/ mocks on the key store. One interesting point, `containsAlias` is final. I got around it easy w/ `mock-maker-inline`. If there are strong opinions against this approach I could just load the `test.jks` store instead. However, I would not be able to test the "throws exception during call to containsAlias" flow that way.

wilkinsona: `containsAlias` will throw an exception if you call it on a `KeyStore` that hasn't been initialised. I think you could test the exception handling with the following:

```
KeyStore keyStore = KeyStore.getInstance("JKS");
assertThatThrownBy(() -> SslUtils.assertStoreContainsAlias(keyStore, "alias"))
 				.isInstanceOf(IllegalStateException.class)
 				.hasMessage("Could not determine if keystore contains alias 'alias'").hasCause(keyStoreEx);
```

bono007: Ah good point @wilkinsona . What is your opinion/preference here - Mockito or load actual test KeyStore?

wilkinsona: As the alternative is to change the mock maker for the whole project, I'd prefer to load an actual `KeyStore`.

bono007: Yep this cross my mind too - a wide change for a teeny tiny test case - but figured I would get a 2nd opinion. I will adjust it sometime today.

bono007: This has ben removed @wilkinsona

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/embedded/netty/NettyReactiveWebServerFactoryTests.java
```diff
@@ -101,14 +99,6 @@ public void whenSslIsConfiguredWithAValidAliasARequestSucceeds() {
 		StepVerifier.create(result).expectNext("Hello World").verifyComplete();
 	}
 
-	@Test
```
bono007: This test is no longer valid as the server will not even startup if its configured w/ an invalid alias.

bono007:
Rebased to fix merge conflicts.

bono007:
Just pinging on this one as its been inactive for a bit.

snicoll:
@bono007 please don't do that. The issue has been triaged and has a milestone attached to it so it is effectively scheduled for resolution.

bono007:
Got it @snicoll . Then please excuse me in advance for the other issue I
did the same thing on. Won't happen again.

mbhave:
Thanks @bono007. This has now been merged into master.

PR #19211, opened 2019-12-02
Improve documentation coverage for Spring Batch

Buzzardo:
At Michael Minella's suggestion and with his input, I extended the Batch section of the How-to page to include more detail.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
@@ -2098,9 +2098,17 @@ The preceding example overrides the default factory, and it should be applied to
 
 [[howto-batch-applications]]
 == Batch Applications
-This section answers questions that arise from using Spring Batch with Spring Boot.
+A number of questions often arise when people use Spring Batch from within a Spring Boot application. This section addresses the following concerns:
 
-NOTE: By default, batch applications require a `DataSource` to store job details.
+* <<howto-spring-batch-specifying-a-data-source>>
+* <<howto-run-spring-batch-jobs-on-startup>>
+* <<howto-spring-batch-running-command-line>>
+* <<howto-spring-batch-storing-job-repository>>
+
+[[howto-spring-batch-specifying-a-data-source]]
+=== Specifying a Batch Data Source
+
+By default, batch applications require a `DataSource` to store job details.
 Batch autowires a single `DataSource` in your context and uses that for processing.
 To have Batch use a `DataSource` other than the application’s main `DataSource`, declare a `DataSource` bean, annotating its `@Bean` method with `@BatchDataSource`.
 If you do so and want two data sources, remember to create another one and mark it as `@Primary`.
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -2097,9 +2097,17 @@ The preceding example overrides the default factory, and it should be applied to
 
 [[howto-batch-applications]]
 == Batch Applications
-This section answers questions that arise from using Spring Batch with Spring Boot.
+A number of questions often arise when people use Spring Batch from within a Spring Boot application. This section addresses the following concerns:
 
-NOTE: By default, batch applications require a `DataSource` to store job details.
+* <<howto-spring-batch-specifying-a-data-source>>
```
snicoll: We usually don't list sections like that. Our toc on the left hand side offers a list of sub-sections anyway so I think this is redundant (and likely to be outdated if we forget to update the list).

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -2108,26 +2116,69 @@ See {spring-batch-api}/core/configuration/annotation/EnableBatchProcessing.html[
 
 For more about Spring Batch, see the {spring-batch}[Spring Batch project page].
 
-
-
-[[howto-execute-spring-batch-jobs-on-startup]]
-=== Execute Spring Batch Jobs on Startup
+[[howto-run-spring-batch-jobs-on-startup]]
```
snicoll: If the anchor is changed, I'd rather harmonize it to the same prefix as you've done for everything else.

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -2108,26 +2116,69 @@ See {spring-batch-api}/core/configuration/annotation/EnableBatchProcessing.html[
...
+[[howto-spring-batch-running-command-line]]
+=== Running from the Command Line
+
+Running Spring Batch with Spring Boot from the command line differs from running Spring
+Batch by itself from the command line. Spring Boot uses a class called
+`JobLaunchingCommandLineRunner`. Spring Batch uses a class called `CommandLineJobRunner`.
+They work a bit differently. However, for the most part, the differences do not cause
+trouble. However, they parse command line arguments differently, which can cause trouble,
+as described in the next section.
+// TODO What are the other differences between the two?
```
snicoll: Is that something we're supposed to handle?

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -2108,26 +2116,69 @@ See {spring-batch-api}/core/configuration/annotation/EnableBatchProcessing.html[
...
+
+==== Passing Command-line Arguments
+
+If you run Spring Batch with Spring Boot, Spring Boot strips the first `-` character (a
```
snicoll: This section should use one sentence per line as the rest of the doc please.

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -2108,26 +2116,69 @@ See {spring-batch-api}/core/configuration/annotation/EnableBatchProcessing.html[
...
+However, if you mean to not use the `jobParameters` value as the identifier for the
+`Job`, use two hyphens, as shown in the following example:
+
+[source]
+--jobParameters="someParameter"
+
+In the second example, Spring Boot passes the parameter to Spring Batch as
+`-jobParameters="someParameter"`, and "someParameter" does not become the name of the
+`Job`.
+// TODO In that case, what is the name of the job?
```
snicoll: I must admit I am bit confused by the current phrasing. Shouldn't we rephrase it around stating that Spring Boot uses `--` as a signal for application arguments and that this class with the way Spring Batch core works?

Buzzardo:
I've asked Mahmoud Ben Hassine (benas here on Github) to review the TODOs. Once that is done, I'll fix it up. Unfortunately, I can't add him as a reviewer.

snicoll:
Waiting for input from @benas then...

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -2108,26 +2116,69 @@ See {spring-batch-api}/core/configuration/annotation/EnableBatchProcessing.html[
...
+Batch without Spring Boot does not require a database. If you use Spring Batch without
+Spring Boot, you may provide a database (including an in-memory database), but it is not
+required. If you do not provide a database, Spring Batch uses a `Map` object as the `Job`
+repository.
+
+See also
+https://docs.spring.io/spring-batch/trunk/reference/html/configureJob.html#configuringJobRepository[Configuring
+a Job Repository]
+// TODO Is it best practice to use a database (rather than the default Map object) for
+// the Job Repository, even when Spring Boot isn't present?
```
benas: yes, the map based job repository is only intended for testing (as mentioned in its [javadoc](https://docs.spring.io/spring-batch/docs/4.2.x/api/org/springframework/batch/core/repository/support/MapJobRepositoryFactoryBean.html)).

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
benas: For this TODO, it is not the name of the job that is relevant, it is rather the job instance identity. When there is "-" in front of a parameter, the parameter is considered as non-identifying and will not contribute to the identity of a the job instance. So the sentence could be re-phrased to something like:

`In the second example, Spring Boot passes the parameter to Spring Batch as
 `-someParameter="someValue"`, and "someParameter" will be considered as a non-identifying job parameter.`

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
```diff
@@ -2108,26 +2116,69 @@ See {spring-batch-api}/core/configuration/annotation/EnableBatchProcessing.html[
...
+difference in Spring Boot. However, in Spring Batch, putting a single `-` character
+before the `jobParameters` parameter indicates that Spring Batch should not use the
+`jobParameters` value as the identifier for the `Job`. Best practice is to use the
+`jobParameters` value as the identifier for the `Job`, so this issue may cause problems.
+To avoid the issue, you should generally use no `-` characters for the command-line
+options that you pass to Spring Boot on behalf of Spring Batch, as shown in the following
+example:
+
+[source]
+jobParameters="someParameter"
```
benas: This is confusing, I would use `someParameter="someValue"` instead.

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
benas: The other difference is that the `CommandLineJobRunner` from batch can accepts options like `-next`, `-stop`, `-restart`, etc while the one from boot does not.

Buzzardo: I left out that detail, since I didn't think it would help a user of Spring Boot.

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
Buzzardo: I dropped most of this, because it doesn't apply under Spring Boot. (I originally wrote this document as a stand-alone document.)

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
Buzzardo: I wrote a paragraph at the top of this section to address Stephane's concern. Then I made the change Mahmoud suggested.

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
Buzzardo: I made that change.

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
Buzzardo: Got it.

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
Buzzardo: It's regarded as a best practice (part of information mapping) in the technical writing community. However, I removed it, because I have no idea how it works with the ToC being constantly present. Do readers refer to the ToC as they read? (I don't, but I'm not everyone.) I wish I had data around that, to have something other than intuition on which to base this kind of decision. Anyway, I removed it.

spring-boot-project/spring-boot-docs/src/main/asciidoc/howto.adoc
Buzzardo: Good plan. I kept `jobParameters` when it's clear we're talking about exactly that argument.

Issue #19282, opened 2019-12-05
Delete useless judgment 


lyyprean:
# 1.old  code

```java
if (StringUtils.hasText(validationQuery)) {
	try {
		// Avoid calling getObject as it breaks MySQL on Java 7
		List<Object> results = this.jdbcTemplate.query(validationQuery, new SingleColumnRowMapper());
		Object result = DataAccessUtils.requiredSingleResult(results);
		builder.withDetail("result", result);
	}
	finally {
		builder.withDetail("validationQuery", validationQuery);
	}
}
```

# 2. After updating the code

Delete useless judgment，Because `validationQuery` must not be null  or “” or “ ”

```java
try {
	// Avoid calling getObject as it breaks MySQL on Java 7
	List<Object> results = this.jdbcTemplate.query(validationQuery, new SingleColumnRowMapper());
	Object result = DataAccessUtils.requiredSingleResult(results);
	builder.withDetail("result", result);
}
finally {
	builder.withDetail("validationQuery", validationQuery);
}
```

philwebb:
Good spot. The code in question is in `DataSourceHealthIndicator` and isn't needed because `getValidationQuery()` always returns a string.

lyyprean:
Thank you:
&nbsp; &nbsp; &nbsp;&nbsp;May I ask if I can join the contributor list，My English is not very good, hope forgive me


------------------&nbsp;原始邮件&nbsp;------------------
发件人:&nbsp;"Phil Webb"<notifications@github.com&gt;;
发送时间:&nbsp;2019年12月5日(星期四) 中午12:46
收件人:&nbsp;"spring-projects/spring-boot"<spring-boot@noreply.github.com&gt;;
抄送:&nbsp;"满天花落"<869441600@qq.com&gt;;"Author"<author@noreply.github.com&gt;;
主题:&nbsp;Re: [spring-projects/spring-boot] Delete useless judgment type：enhancement (#19282)




Good spot. The code in question is in DataSourceHealthIndicator and isn't needed because getValidationQuery() always returns a string.

lyyprean:
thank you

philwebb:
You're most welcome to submit a PR if you'd like to become a contributor.

lyyprean:
What is PR ?   Sorry, my english is not very good

philwebb:
[...]

PR #19285, opened 2019-12-05
Delete useless judgment 

lyyprean:
# 1.old code
`
if (StringUtils.hasText(validationQuery)) { try { // Avoid calling getObject as it breaks MySQL on Java 7 List<Object> results = this.jdbcTemplate.query(validationQuery, new SingleColumnRowMapper()); Object result = DataAccessUtils.requiredSingleResult(results); builder.withDetail("result", result); } finally { builder.withDetail("validationQuery", validationQuery); } } }
`
# 2. After updating the code
Delete useless judgment，Because validationQuery must not be null or “” or “ ”
`
try { // Avoid calling getObject as it breaks MySQL on Java 7 List<Object> results = this.jdbcTemplate.query(validationQuery, new SingleColumnRowMapper()); Object result = DataAccessUtils.requiredSingleResult(results); builder.withDetail("result", result); } finally { builder.withDetail("validationQuery", validationQuery); }
`

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/deployment.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/deployment.adoc
@@ -41,8 +41,8 @@ For example, using a `Dockerfile` you could express it in this form:
 ----
 	FROM openjdk:8-jdk-alpine AS builder
 	WORKDIR target/dependency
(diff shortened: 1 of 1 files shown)


philwebb:
This PR seems unrelated to the comment so we'll carry on discussing the issue in #19282

PR #19290, opened 2019-12-05
Remove redundant hasTest check in DataSourceHealthIndicator

lyyprean:
Delete useless judgment，Because validationQuery must not be null or “” or “ ”


wilkinsona:
Thanks very much, @lyyprean. The proposed change is now in 2.1.x, 2.2.x, and master.

lyyprean:
May I be part of the contribution， thank you

Issue #19200, opened 2019-12-01
Add Liquibase "tag" property support


junaid155:
Please add a new property with name  **tag** in `org.springframework.boot.autoconfigure.liquibase.LiquibaseProperties`. Also update `LiquibaseAutoConfiguration.java` to set tag property in `SpringLiquibase` bean. 

This property will help in generating roll back scripts for particular tags only when user has specified `rollbackFile` property.

eddumelendez:
I can work on this issue :)

mbhave:
Thanks @eddumelendez! I've assigned it to `spring-contributor.`

junaid155:
@eddumelendez  it will be better if we have separate tag property for generateRollbackFile and performUpdate method in SpringLiquibase.java.

eddumelendez:
@junaid155 thanks, but `SpringLiquibase` only offers `tag` property for `generateRollbackFile` and `performUpdate`. The execution depends basically if property `rollbackFile` is set or not. So, I can not have separate properties for each one. 

Now, that you said that I just want to make sure if what you are expecting is a sql file as an output which will be in this case. That means the rollback is not executed automatically. I found this thread #1826 

I think we can set the use cases right here and evaluate them. Otherwise, some questions can raise for different flows. WDYT?

junaid155:
@eddumelendez  Thanks, My major concern is that i want to generate roll back scripts till particular tag only. just like liquibase command `liquibase   rollback=3.0.1 `

philwebb:
Closing in favor of PR #19316

snicoll:
@junaid155 I am looking at the PR and I am very confused how that feature actually works. I am looking at the liquibase forum and I can see some users confused as well. I am not keen to merge the PR as is as I'd like to understand the use case a bit better. 

Would you please rephrase the use cases you had in mind? It looks like to me this is something you'd like to apply once and not a property you'd hardcode in your configuration.

snicoll:
@junaid155 gentle ping for the question above.

PR #19316, opened 2019-12-06
Add property tag for LiquibaseProperties

eddumelendez:
See gh-19200

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
@@ -28,6 +28,7 @@ import org.springframework.util.Assert;
  * Configuration properties to configure {@link SpringLiquibase}.
  *
  * @author Marcel Overdijk
+ * @author Eddú Meléndez
  * @since 1.1.0
  */
 @ConfigurationProperties(prefix = "spring.liquibase", ignoreUnknownFields = false)
@@ -113,6 +114,11 @@ public class LiquibaseProperties {
 	 * Whether rollback should be tested before update is performed.
 	 */
 	private boolean testRollbackOnUpdate;
+
+	/**
+	 * Name of the tag.
+	 */
+	private String tag;
 
 	public String getChangeLog() {
 		return this.changeLog;
@@ -241,6 +247,14 @@ public class LiquibaseProperties {
 
 	public void setTestRollbackOnUpdate(boolean testRollbackOnUpdate) {
 		this.testRollbackOnUpdate = testRollbackOnUpdate;
+	}
+
+	public String getTag() {
+		return this.tag;
+	}
+
+	public void setTag(String tag) {
+		this.tag = tag;
 	}
 
 }
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
@@ -112,6 +112,7 @@ public class LiquibaseAutoConfiguration {
 			liquibase.setChangeLogParameters(this.properties.getParameters());
 			liquibase.setRollbackFile(this.properties.getRollbackFile());
 			liquibase.setTestRollbackOnUpdate(this.properties.isTestRollbackOnUpdate());
+			liquibase.setTag(this.properties.getTag());
 			return liquibase;
 		}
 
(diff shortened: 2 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseProperties.java
```diff
@@ -114,6 +115,11 @@
 	 */
 	private boolean testRollbackOnUpdate;
 
+	/**
+	 * Name of the tag.
+	 */
+	private String tag;
```
snicoll: While reviewing this PR, I was trying to improve the documentation of that property and couldn't really find my way in the liquibase documentation. The original issue states:

> This property will help in generating roll back scripts for particular tags only when user has specified rollbackFile property.

Do you know more about this property and how it's actually used? I've looked into the liquibase source code and I am not sure I understood the scope of that property. If that's used only for rollbacks, I think we should make that clearer in the description of the property.

eddumelendez: when I was looking into the code I found `SpringLiquibase` uses `tag` property at `performUpdate` and `generateRollbackFile` methods. It means is executed during the update of changelogs or if additionally `rollbackFile` is also set will generate the file.

snicoll: Thanks for the feedback. Same but I think "Name of the tag" isn't enough for users to figure out what to do. Perhaps this is an issue in the Liquibase documentation?

snicoll: I poked around and I am confused. I don't think we should just have a "tag" property, I've [commented on the original issue](https://github.com/spring-projects/spring-boot/issues/19200#issuecomment-568689549).

eddumelendez: sorry for the late reply Stéphane! I have been thinking about it and TBH I think the workflow is pretty much deploying changes using the CLI and generating the rollback file from the application. As we have seen, the same property is used for rolling out and rollback. Said that, we should be very careful about the docs as you are worried about.

Delivering this feature, people will be able to:
1. Use `tag` property, apply the tag value to the scripts

or

2. Use `tag` property and `rollbackFile`,  generate the file to perform a rollback

SteveDonie: Hi - I am the Liquibase Community Manager and can comment some on the use cases, which seem to be a bit muddy in this ticket. 

I'm still learning about how Liquibase works in the Spring Boot ecosystem, so please bear with me if I get something wrong.

In Liquibase, there is a notion of a tag that can be applied to the DATABASECHANGELOG table (the table that Liquibase uses to keep track of which changeSets have been applied to a particuloar database). A tag is represented by adding a non-null value to the TAG column of the DATABASECHANGELOG table.

A tag can be applied in one of two ways - it can be included in the changelog, or it can be applied 'manually' by running the tag command, which is available in the CLI and as a maven goal. 

Once a tag has been applied, it can then be used by the rollback command. When used by the rollback command, any changesets that are represented in the DATABASECHANGELOG table that are after the tagged changeset will be rolled back. 

When rolling back, the operation can either be done 'live', or a SQL file can be generated containing the SQL commands needed. 

This is not well documented in the Liquibase docs, so I'll be updating those soon.

snicoll: @SteveDonie thanks for the feedback and improving the docs.

We need a one sentence that would be a good summary for the property and we had a hard time finding something succinct.

Here is a proposal

> Tag name to add to applied changes. Can be used with "rollbackFile" to generate a rollback script for all existing changes with that tag. 

Thoughts?

SteveDonie: I think that is pretty good. Maybe even call out that there are two different uses:

> Tag name to apply when applying database changes. Tag name to use as a target when rolling back already applied changes.

SteveDonie: When rolling back, you don't have to generate a rollbackFile. I'm not sure how it works in spring boot, but from the command line there are two commands - rollback, and rollbackSQL. One does the actual rollback, and the other generates the SQL to perform the rollback and shows it on the console (or sands it to a file).

wilkinsona: Thanks very much for the pointers, @SteveDonie.

snicoll: Indeed, thank you @stevedonie. FTR, Spring Boot "merely" instruments  `SpringLiquibase` which is part of Liquibase itself.

Issue #19321, opened 2019-12-06
@DataJpaTest using H2 with schema.sql and spring.datasource.schema-username fails


silovmar:
Version: Spring Boot v2.2.1.RELEASE, Spring v5.2.1.RELEASE

Hi, I'm using `@DataJpaTest` with default embedded H2 DB for testing. Also I'm creating DB schema using `schema.sql` which contains `CREATE SCHEMA IF NOT EXISTS TEST_SCHEMA;`.

Unit tests annotated with the `@DataJpaTest` ran fine until I added to my production `application.properties` properties `spring.datasource.schema-username=${DB_USERNAME_SCH}` and `spring.datasource.schema-password=${DB_PASSWORD_SCH}`.

This causes that a datasource based on the application.properties is trying to be created instead of the default H2 database. The creation of DB from application.properties fails because the datasource configuration looks like this:

```
spring.datasource.driver-class-name= oracle.jdbc.OracleDriver
spring.datasource.url= ${DB_URL}
spring.datasource.username= ${DB_USERNAME}
spring.datasource.password= ${DB_PASSWORD}
spring.datasource.schema-username= ${DB_USERNAME_SCH}
spring.datasource.schema-password= ${DB_PASSWORD_SCH}
```

I would expect that the embedded H2 is used even if the schema-username and schema-password is used.

Source code which caueses the problem:
[org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java:197](https://github.com/spring-projects/spring-boot/blob/eb0eb492b4a12d12955e77f2d94492405ce4c076/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java#L197) which is called from [org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java:101](https://github.com/spring-projects/spring-boot/blob/eb0eb492b4a12d12955e77f2d94492405ce4c076/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java#L101)

philwebb:
This is an edge case that we've not considered. You might be able to work around it by overriding the `spring.datasource.schema-username` property to `""` in your test.

eddumelendez:
I have added my findings in https://github.com/spring-projects/spring-boot/pull/19327#issuecomment-562846662

wilkinsona:
[...]

PR #19327, opened 2019-12-07
@DataJpaTest using H2 with schema.sql and spring.datasource.schema-username fails

eddumelendez:
Currently, `jdbcUrl` is determined when `spring.datasource.schema-username`
and `spring.datasource.schema-password` configuration properties are set.
This behaviour causes an issue trying to perform sql scripts in the
wrong database. This commit introduces to get the right `url` from the
`dataSource` bean.

See gh-19321

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java
@@ -28,6 +28,7 @@ import org.apache.commons.logging.LogFactory;
 import org.springframework.boot.context.properties.source.InvalidConfigurationPropertyValueException;
 import org.springframework.boot.jdbc.DataSourceBuilder;
 import org.springframework.boot.jdbc.DataSourceInitializationMode;
+import org.springframework.boot.jdbc.DatabaseDriver;
 import org.springframework.boot.jdbc.EmbeddedDatabaseConnection;
 import org.springframework.core.io.DefaultResourceLoader;
 import org.springframework.core.io.Resource;
@@ -35,6 +36,8 @@ import org.springframework.core.io.ResourceLoader;
 import org.springframework.jdbc.config.SortedResourcesFactoryBean;
 import org.springframework.jdbc.datasource.init.DatabasePopulatorUtils;
 import org.springframework.jdbc.datasource.init.ResourceDatabasePopulator;
+import org.springframework.jdbc.support.JdbcUtils;
+import org.springframework.jdbc.support.MetaDataAccessException;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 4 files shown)


eddumelendez:
These are my finding regarding the issue: The first h2 database created is generated with a random uuid which is used in the `url`. When `spring.datasource.schema-username` and `spring.datasource.schema-password` are set there is a call to `determineUrl` which will return `testdb` in this case and this `schema.sql` is executed with there. This behaviour combined with `spring.datasource.data` will perform `data.sql` into the database with the random uuid which will cause issues because the schema doesn't exist due to in this case `determineUrl` is not call.

I'm missing the test in this PR but trying to figure it out how to do it. This is due to we need to set up the right `spring.datasource.*-username` and `spring.datasource.*-password` (both required right now) otherwise we will get an exception with a message `don't have right permissions`.

I have two thoughts here:
1. Add `!isEmbedded()` as a part of the condition in order to create the  new datasource with the right credentials because embedded datasources doesn't have password. And they way to do it for h2 at least which I was looking into it is a way more complicated to test it.
2. Relax the condition from `AND` to `OR`

I would like to discuss approaches in order to tackle this issue.

eddumelendez:
I have updated the PR, moving the password validation which make sense for embedded database and will work for other escenarios if is needed. Also, test was added.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java
```diff
@@ -193,13 +196,28 @@ private void runScripts(List<Resource> resources, String username, String passwo
...
+
+	private DataSource resolveDataSource(String username, String password) {
 		DataSource dataSource = this.dataSource;
-		if (StringUtils.hasText(username) && StringUtils.hasText(password)) {
-			dataSource = DataSourceBuilder.create(this.properties.getClassLoader())
-					.driverClassName(this.properties.determineDriverClassName()).url(this.properties.determineUrl())
-					.username(username).password(password).build();
+		if (StringUtils.hasText(username)) {
+			try {
+				String url = JdbcUtils.extractDatabaseMetaData(this.dataSource, "getURL");
```
snicoll: It feels to me that this is the wrong approach to the problem. Looking at the reported issue, setting `spring.datasource.schema-username` should not have any impact on an embedded database that we've created. 

Rather, we should detect such case and ignore the username and password.

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfigurationTests.java
```diff
@@ -60,6 +63,41 @@ void replaceWithUniqueDatabase() {
...
+	void testRunningSchemaSqlScriptEvenWithWrongCredentials() {
+		this.contextRunner.withConfiguration(AutoConfigurations.of(DataSourceAutoConfiguration.class))
+				.withPropertyValues("spring.datasource.schema-username=sa", "spring.datasource.schema-password=sa",
+						"spring.datasource.schema=" + getRelativeLocationFor("schema.sql"),
+						"spring.datasource.data=" + getRelativeLocationFor("data.sql"))
+				.run((context) -> {
+					assertThat(context).hasFailed();
+					assertThat(context).getFailure()
+							.hasMessageContaining("Failed to obtain JDBC Connection; nested exception is")
+							.hasMessageContaining("Wrong user name or password");
```
snicoll: That test looks wrong to me. If a test setup has replaced the production database by an embedded one, then the credentials of the production database should not impact the ability to initialize the embedded database.

These  settings have been introduced to perform the initialization of a "production" database with different credentials (and only for that step). That shouldn't apply to an embedded database IMO.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java
eddumelendez: I can add a validation that if `isEmbedded()` then `*-username` and `*-password` are ignored. Indeed, looking at the [comment](https://github.com/spring-projects/spring-boot/pull/19327#issuecomment-562846662) I left, that was one the approaches. WDYT?

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/jdbc/TestDatabaseAutoConfigurationTests.java
eddumelendez: Yes, you are right.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceInitializer.java
snicoll: Unfortunately, the fact the database is embedded is not exactly the same as us replacing the database with an embedded database. 

In theory, I'd very surprised if an embedded database would require different credentials for schema initialization so that approach would work but not really semantically correct IMO.

snicoll:
We've discussed this one and we'd like to try a different approach. I've reopened #19321, see [this comment](https://github.com/spring-projects/spring-boot/issues/19321#issuecomment-611025209). Thanks for the PR, in any case!

PR #19372, opened 2019-12-13
Clean Content-Encoding response header in WebFlux error handler

### LINKED ISSUE #19371: AbstractErrorWebExceptionHandler should not keep exchange response headers in an error response
I ran into a problem where one of my Gateway Filter which filters an exchange response throws an exception. 
There was a header `Content-Encoding: gzip` in the response and this header (along with the rest) was forwarded to error response headers. But the `ServerResponse` body was written as JSON (error attributes) and wasn't zipped therefore a server response is incorrect.

I suggest clearing all headers in an exchange's response in order to fix that and similar header problems.

Currently `AbstractErrorWebExceptionHandler` just overwrites `Content-Type` header of the response:
```
private Mono<? extends Void> write(ServerWebExchange exchange, ServerResponse response) {
	// force content-type since writeTo won't overwrite response header values
	exchange.getResponse().getHeaders().setContentType(response.headers().getContentType());
	return response.writeTo(exchange, new ResponseContext());
}
```
which can be replaced with
`exchange.getResponse().getHeaders().clear();`

timsazon:
Fix #19371

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/error/AbstractErrorWebExceptionHandler.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/error/AbstractErrorWebExceptionHandler.java
@@ -34,6 +34,7 @@ import org.springframework.boot.web.reactive.error.ErrorWebExceptionHandler;
 import org.springframework.context.ApplicationContext;
 import org.springframework.core.NestedExceptionUtils;
 import org.springframework.core.io.Resource;
+import org.springframework.http.HttpHeaders;
 import org.springframework.http.HttpLogging;
 import org.springframework.http.HttpStatus;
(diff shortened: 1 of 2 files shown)


bclozel:
@timsazon Could you amend this PR and:

* reinstate the previous code
* only remove the `Content-Encoding` header
* add a test to check that this is working as expected

I can assist you with the changes. If you don't have time for this, let me know and I'll handle it myself.
Thanks!

timsazon:
@bclozel Thanks for the answer!
I think it would be better to clear all content-related headers in that case. What do you think?

bclozel:
I think we can take it from here.
I don't think we should clear `Content-Disposition` as this header is used to protected against RFD exploits. Others seem sensible, maybe a bit overkill. We'll figure this out while processing your PR. Thanks!

bclozel:
After considering our options, we'll fix this in Spring Framework directly.
I've created spring-projects/spring-framework#24238.

Thanks for your contribution!

PR #19377, opened 2019-12-16
Remove duplicate entry for Neo4jHealthIndicator in doc

izeye:
This PR removes a duplicate entry for `Neo4jHealthIndicator` in the reference doc on the 2.1.x branch, which seems to be introduced in 87caa0a3316be05b4aac0900a8fd167fd2b2e3d8 somehow.


snicoll:
Thanks Johnny. This was backported to 2.1.x so maybe the error occurred when rebasing or something. If nobody beats me to it, I’ll double check those branches too.

bclozel:
Duplicate entry is only present on `2.1.x`, other branches look good.

PR #19397, opened 2019-12-18
Add Elide as 3rd party Spring Boot Starter

aklish:
Adding Elide Spring Boot starter to list of starters.

https://elide.io

### DIFF
--- a/spring-boot-project/spring-boot-starters/README.adoc
+++ b/spring-boot-project/spring-boot-starters/README.adoc
@@ -76,6 +76,9 @@ do as they were designed before this was clarified.
 | Elegant Error Handling for Spring Boot
 | https://github.com/alimate/errors-spring-boot-starter
 
+| https://elide.io/[Elide]
+| https://github.com/yahoo/elide/tree/master/elide-spring/elide-spring-boot-starter
+
 | ErroREST exception handler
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@aklish Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19397)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19397) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@aklish Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19397)!

bclozel:
Quick question: why isn't [the sample Elide+Spring Boot application](https://github.com/yahoo/elide-spring-boot-example/blob/master/pom.xml) using the Spring Boot parent / BOM?

aklish:
Good call.  I changed the example to use the spring-boot parent pom.  It
shaved down the length.

snicoll:
Good catch, I think also the Spring integration (at least) could use some harmonization with what's going on in practice. I've created https://github.com/yahoo/elide/issues/1116

Issue #19205, opened 2019-12-02
SSL support for the RSocketServer


joshiste:
Currently configuring the RSocketServer to use SSL is a quite tedious task:
- there are no properties to configure it
- there is no option to to customize the `HttpServer` for the websocket transport or the `TcpServer` for the tcp transport created by the NettyRSocketServerFactory

joshiste:
I've currently hacked use this hack reusing the `SslServerCustomizer`

```java
public class SecureNettyRSocketServerFactory implements RSocketServerFactory, ConfigurableRSocketServerFactory {
	private int port = 9898;
	private InetAddress address;
	private RSocketServer.Transport transport = RSocketServer.Transport.TCP;
	private ReactorResourceFactory resourceFactory;
	private Duration lifecycleTimeout;
	private Ssl ssl;
	private List<ServerRSocketFactoryProcessor> socketFactoryProcessors = new ArrayList<>();

	@Override
	public void setPort(int port) {
		this.port = port;
	}

	@Override
	public void setAddress(InetAddress address) {
		this.address = address;
	}

	@Override
	public void setTransport(RSocketServer.Transport transport) {
		this.transport = transport;
	}

	public void setSsl(Ssl ssl) {
		this.ssl = ssl;
	}

	/**
	 * Set the {@link ReactorResourceFactory} to get the shared resources from.
	 *
	 * @param resourceFactory the server resources
	 */
	public void setResourceFactory(ReactorResourceFactory resourceFactory) {
		this.resourceFactory = resourceFactory;
	}

	/**
	 * Set {@link ServerRSocketFactoryProcessor}s that should be called to process the
	 * {@link RSocketFactory.ServerRSocketFactory} while building the server. Calling this method will
	 * replace any existing processors.
	 *
	 * @param socketFactoryProcessors processors to apply before the server starts
	 */
	public void setSocketFactoryProcessors(
			Collection<? extends ServerRSocketFactoryProcessor> socketFactoryProcessors
	) {
		Assert.notNull(socketFactoryProcessors, "SocketFactoryProcessors must not be null");
		this.socketFactoryProcessors = new ArrayList<>(socketFactoryProcessors);
	}

	/**
	 * Add {@link ServerRSocketFactoryProcessor}s that should be called to process the
	 * {@link RSocketFactory.ServerRSocketFactory} while building the server.
	 *
	 * @param socketFactoryProcessors processors to apply before the server starts
	 */
[...]

PR #19399, opened 2019-12-19
Add SSL support to RSocketServer

bono007:
Closes gh-19205


eye2web:
Is it still the idea to integratie this feature? I'm currently in the need to create a SSL based Rsocket Spring server.

bclozel:
@eye2web it's rather late to include that in our 2.3.0 release, I've scheduled that for the next one.

Issue #19343, opened 2019-12-10
ApplicationContextRunner evaluates @Conditional on AutoConfiguration too early


ttddyy:
Hi there,

I faced another behavioral difference between `AutoConfigurationImportSelector` and `ApplicationContextRunner`.

When `@Conditional` exists on auto configuration class, with `ApplcicationContextRunner#withConfiguration` using `AutoConfigurations.of`, the evaluation of `Conditional` happens early(at import time, not at processing auto configurations).

Here is the usecase and sudo code:

I am trying to control autoconfiguration (enable/disable) based on the annotation on user config.

```java
// User Configuration
@Configuration
@EnableX  // this enables MyAutoConfiguration
public static class MyUserConfig {
}

@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@Documented
@Import(MyImportSelector.class)
public @interface EnableX {
}

public static class MyImportSelector implements ImportSelector {
  public static boolean enabled;  // flag to enable/disable ConditionalOnX on MyAutoConfiguration

  @Override
  public String[] selectImports(AnnotationMetadata metadata) {
    enabled = true;
    return new String[0];  // return empty
  }
}
```

```java
// Auto Configuration
@Configuration
@ConditionalOnX
static class MyAutoConfiguration {
  @Bean
  public String foo() {
    return "FOO";
  }
}

@Target(ElementType.TYPE)
@Retention(RetentionPolicy.RUNTIME)
@Documented
@Conditional(OnXCondition.class)
public @interface ConditionalOnX {
}

public static class OnXCondition extends SpringBootCondition {
  @Override
  public ConditionOutcome getMatchOutcome(ConditionContext context, AnnotatedTypeMetadata metadata) {
    return MyImportSelector.enabled ? ConditionOutcome.match("enabled") : ConditionOutcome.noMatch("disabled");
  }
}
```

```java
@Test
void contextRunnerWithConditionOnAutoConfiguration() {
  new ApplicationContextRunner()
      .withInitializer(new ConditionEvaluationReportLoggingListener(LogLevel.INFO))
      .withConfiguration(AutoConfigurations.of(MyAutoConfiguration.class))
      .withUserConfiguration(MyUserConfig.class)
      .run(context -> {
        assertThat(context)
            .hasNotFailed()
            .hasBean("foo")
        ;
      });
}
```

[...]

Issue #18228, opened 2019-09-13
Consider providing recommendation for override/replace beans usecase


ttddyy:
Since spring-boot 2.1, bean overriding is disabled by default.
I think this is very reasonable and protects from unknowns by preventing hidden bean overriding.

However, time to time, there is a need to override beans.
It would be great to have the best practice/guideline for handling such use cases.

### Background:

I write a library which is used by multiple service applications. The library is written on top of spring-boot and provides `@Enable...` annotations. 
Typically we create `@Enable...` annotations rather than creating auto configurations.
This is a decision that we want to *explicitly* turn on/off features in service rather than auto-magically happen via auto configuration.

The issue is that sometimes services want to override some beans defined in library.
If we provide auto configuration, we can use `@ConditionalOn...` annotations since it is guaranteed to be evaluated after all `@Configuration` classes are processed by `AutoConfigurationImportSelector`.

However, since we use `@EnableX`, there is no distinction for `@Configuration` classes whether they are from library's or from application's while processing configuration classes.
Therefore, due to the component scan is not deterministic to process configuration classes, we cannot reliably use `@ConditionalOnX` annotations because the configuration class processing order might change.

One approach is I can write a custom import-selector that makes sure all library configurations are  processed after application's configurations but before auto configurations. This could allow library configurations to use `@ConditionalOn` annotations. However, it is a bit of too much in custom framework and even writing a custom import-selector, we also need to `@EnableAutoConfiguration` equivalent to bring in the custom import-selector to be in effect on application.

We also want to avoid using `@Primary` defined in application configuration in order to override beans in library. This is because `@Primary` bean may have impact writing tests in application.

Instead, an approach I chose was using `BeanFactoryPostProcessor` and unregister bean definitions that I want to override.

```java
public class UnregisterBeanDefinitionBeanFactoryPostProcessor implements BeanFactoryPostProcessor, Ordered {
  private boolean lenient = false;
  private List<String> beanNames = new ArrayList<>();  // bean names to unregister

  @Override
  public void postProcessBeanFactory(ConfigurableListableBeanFactory beanFactory) throws BeansException {
    BeanDefinitionRegistry registry = (BeanDefinitionRegistry) beanFactory;
    for (String beanName : this.beanNames) {
        try {
            registry.removeBeanDefinition(beanName);
        } catch (NoSuchBeanDefinitionException e) {
            if (!this.lenient) {
                throw e;
            }
        }
    }
  }

  // ...
}
```

[...]

PR #19400, opened 2019-12-19
Add DeferredImportSelector that runs before/after auto configuration

ttddyy:
Hi,

I wrote custom `DeferredImportSelectors` and corresponding annotations that run before/after auto configurations for my library.
- _`Import[Before|After]AutoConfigurationDeferredImportSelector`:_
  `DeferedImportSelector`  implementation that runs before/after auto configuration
- _`@Import[Before|After]AutoConfiguration`:_
  Use the above deferred selector to specify which configurations to run before/after auto configuration

The background is we have shared library on top of spring-boot and provides shared configurations. Applications will pick some of the shared configurations to enable features.
Since those shared configurations are not autoconfigurations, because of our preference to explicitly enable each feature, I had some problems with configuration orders especially with `@ConditionalOn[Missing]Bean`.
By using these `DeferredImportSelectors`, I have better control for my library's configurations to run  after user's configurations, and then before/after auto configurations.

Sample usage are:

In library:
```java
// Make this configuration runs after user config but before autoconfig
class MyFeatureConfig {
  // @Bean to enable some feature
}

// annotation to enable my feature(config)
@Target(ElementType.TYPE)
@Documented
@ImportBeforeAutoConfiguration(MyFeatureConfig.class)  // <=== meta annotate
@interface EnableMyFeature {
}
```

In application:
```java
@EnableMyFeature
@SpringBootApplication
class Application {
}
```

I am going to apply these import selectors to our library.
At the same time, I think this is also beneficial if exists in spring-boot itself, hence the PR here.

Thanks,

Relates to: #18228, #19343

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AbstractImportBeforeAndAfterAutoConfigurationDeferredImportSelector.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AbstractImportBeforeAndAfterAutoConfigurationDeferredImportSelector.java
@@ -1,0 +1,127 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure;
+
+import java.util.ArrayList;
+import java.util.Comparator;
+import java.util.List;
+
+import org.springframework.context.annotation.DeferredImportSelector;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.AnnotationAttributes;
+import org.springframework.core.annotation.AnnotationAwareOrderComparator;
+import org.springframework.core.type.AnnotationMetadata;
+import org.springframework.util.Assert;
+import org.springframework.util.ClassUtils;
+
+/**
+ * {@link DeferredImportSelector} implementation that works with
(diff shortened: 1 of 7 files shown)


philwebb:
We've been discussing this issue along with #18228 and #19343 and ultimately feel like this is adding too much complexity into Spring Boot for a quite specialized use-case.

It sounds like a lot of the problems that have been raised against Spring Boot are coming from the desire to have something behave as both an `@Enable...` annotation, but also a form of auto-configuration. I think it's likely that such an approach is always going to be moving against the grain of the features we provide and I it's probably worth reconsidering to see if there's another design that might work better.

My personal opinion is that if you are using `@Enable` annotations then you shouldn't also try to combine them with `@Condition` annotations. I'd think it would be better to keep to a purely configuration based approach and model your code in a similar way to Spring MVC where configuration and bean overrides need to happen in a very specific way (e.g. the `WebMvcConfigurer` interface or `WebMvcConfigurationSupport` class).

If you don't want to do that, then I'd go all in on auto-configuration and design classes that work in the same way as other auto-configurations. If you really want them to be opt-in, you could write your own `@Condition` that checks against something. For example, you could have an annotation that uses an `ImportBeanDefinitionRegistrar` to records it was processed then use a condition that checks for a recorded result (`AutoConfigurationPackage` works a bit like that but it's not got a condition attached to it). If you go this route, it would be better to not use `@Enable...` as the annotation prefix since that tends to mean something that isn't auto-configuration. Perhaps you could use something like `@AllowCommonMetrics` instead?

I appreciate that this probably isn't the outcome you wanted for this pull-request. Unfortunately we really need to balance changes such as this against the future maintenance costs they might incur and the signal that they send. In this case, we just feel like there's going to be a lot of additional complexity and questions raised which won't be able to answer or support.

ttddyy:
@philwebb Thanks for the feedback.

First approach is a bit cumbersome for us to create such configuration extension points for each feature, and requires to consider rather many permutation of possibilities.

So, I'll probably go with the second suggestion since we need `@Conditional` and opt-in capability.
Putting our configurations in auto-configuration semantics is better to take advantage of auto-configuration infrastructure.
Probably I'll create custom annotation `@Allow...` or `@Use...` to enable feature which probably put some entry in in-memory map or a bean, and create a custom `@Condition` on our configurations to check the map to whether to activate the `@Configuration`. Basically, auto-configurations disabled by default and enabled by annotation.

I think writing library on top of spring-boot is one way of consuming spring-boot in order for micro-service applications in an organization to streamline the feature and adapting laid out infrastructure.
While evolving spring-boot, keeping in such aspect would be appreciated.

Thanks,

PR #19437, opened 2019-12-23
Polish LdapAutoConfigurationTests

izeye:
This PR polishes `LdapAutoConfigurationTests` by using AssertJ `extracting()`.


snicoll:
@izeye ah! I was trying to remember how I could avoid using `ReflectionTestUtils` there and forgot about `extracting`. Even better than a reminder, thank you so much again.

Issue #19348, opened 2019-12-11
Actuator Kafka consumer metrics turned off for Spring Boot 2.2.2 version


IvanIschenko:
Hi, Spring community!

Sorry if I created this issue in the wrong place. 

During the update process to spring-boot 2.2.2.RELEASE version, the Actuator Kafka consumer metrics were turned off, and they were not reachable with http://localhost:8080/actuator/metrics endpoint.
The metrics are working if I explicitly specify the versions of the Actuator and Spring Boot to 2.1.9.RELEASE.
I would appreciate if someone will have a look at the problem.

wilkinsona:
Kafka metrics should work the same with 2.2 as they did with 2.1. To help us to investigate why that is apparently not the case, can you please provide a small sample application that reproduces the problem?

IvanIschenko:
Of course, I am using the following dependencies:
- spring-boot-starter-parent 2.2.2.RELEASE
- spring-boot-starter-actuator 2.2.2.RELEASE (auto-configured)
- org.springframework.kafka.spring-kafka 2.3.4.RELEASE (auto-configured)
- org.apache.kafka.kafka-streams 2.3.1 (auto-configured)
- org.apache.kafka.kafka-clients 2.3.1 (auto-configured)

So I have just created a simple Kafka Stream API application. 
And when I am accessing  http://localhost:8080/actuator/metrics in the metrics list I can not see Kafka one. But when I used spring-boot-starter-parent 2.1.9.RELEASE (all other dependencies were auto-configured) in the metrics list were Kafka consumer metrics.

wilkinsona:
Thanks. Can you please turn that information into a small sample application that we can run? You can share it with us by zipping it up and attaching it to this issue or by pushing it to a separate repository on GitHub.

dreis2211:
Sorry to hijack this. Isn't that caused by JMX not being enabled by default anymore in 2.2? The `KafkaMetricsAutoConfiguration` is checking for an `MBeanServer` that isn't available when JMX is turned off. @IvanIschenko Can you try adding `spring.jmx.enabled=true` as a configuration property?

wilkinsona:
Ah, yes, thanks @dreis2211. That'll be what it is. I'd overlooked the side-effect of JMX now being disabled by default.

@IvanIschenko If this doesn't help, please provide the sample that I asked for above and we can re-open the issue and take another look.

IvanIschenko:
[...]

PR #19446, opened 2019-12-24
Add JMX note for Kafka consumer metrics

izeye:
Looking at #19348, it seems to be good to add a JMX note for Kafka consumer metrics similar to Tomcat metrics.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/asciidoc/production-ready-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/main/asciidoc/production-ready-features.adoc
@@ -1644,7 +1644,7 @@ Spring Boot registers the following core metrics when applicable:
 ** Number of classes loaded/unloaded
 * CPU metrics
 * File descriptor metrics
-* Kafka consumer metrics
+* Kafka consumer metrics (<<production-ready-jmx,JMX support>> should be enabled)
 * Log4j2 metrics: record the number of events logged to Log4j2 at each level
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/main/asciidoc/production-ready-features.adoc
```diff
@@ -1644,7 +1644,7 @@ Spring Boot registers the following core metrics when applicable:
 ** Number of classes loaded/unloaded
 * CPU metrics
 * File descriptor metrics
-* Kafka consumer metrics
+* Kafka consumer metrics (`spring.jmx.enabled` must be set to `true`)
```
snicoll: This is a good idea. I'd rather change the note to something like "JMX support should be enabled" or something like that with a link to the section that explains how to do it rather than the reference to the property.

izeye: @snicoll Thanks for the feedback! I changed it as suggested.

snicoll: Thanks for the quick feedback. I meant that the text on "JMX support" should be a link to the relevant section. Have you missed it or do you disagree with that approach?

izeye: @snicoll Sorry. I overlooked your comment. I changed it again. Please let me know if it isn't what you meant.

PR #19458, opened 2019-12-24
Adapt file rotation documentation to logging's maxHistory default change to 7 days

gintellect79:
Documentation is corrected to reflect the changes made to change the default `logging.file.max-history` to 7 instead of indefinite.


snicoll:
@gintellect79 thank you for making your first contribution to Spring Boot. I've polished a bit your proposal to strengthen it is about 7 days and not 7 archived files.

Issue #19381, opened 2019-12-16
Issue in Micrometer interceptor if intercept calls another endpoint


grahammkelly:
As of Spring boot 2.2.2 (and 2.1.9), there is an issue in the micrometer integration when you have a `RestTemplate` call being intercepted and redirected to another URL.

## Use case

The main use case I have, being affected by the issue, is where I intercept a call to an external API to provide authentication on the call.

For instance, we call an external API, but intercept to check if already authenticated. If authenticated, add the auth token to the call. If NOT authenticated, place a call to another endpoint to perform the authentication, then add the auth token to the original call and complete the original call.

In the case where the auth call is being made, the metrics details for the original call is being lost. TBF, this mainly affects calls with path parameters.

This seems to be an issue in how the `org.springframework.boot.actuate.metrics.web.client.MetricsClientHttpRequestInterceptor` stores the `urlTemplate`, as a `ThreadLocal`. In the example supplied, the `intercept` method will wipe the `urlTemplate` after completion of the auth call.

```java
    @Override
	public ClientHttpResponse intercept(HttpRequest request, byte[] body, ClientHttpRequestExecution execution)
			throws IOException {
		...
		try {
			response = execution.execute(request, body);
			return response;
		}
		finally {
			getTimeBuilder(request, response).register(this.meterRegistry).record(System.nanoTime() - startTime,
					TimeUnit.NANOSECONDS);
			urlTemplate.remove();   // <-- This is the offending command
		}
	}
```

Ideally, the `urlTemplate` needs to be a thread local stack of some type, OR if the performance hit on this is too large, there needs to be some way to override the default functionality for anyone needing this sort of functionality.

PS - If a fix is created for this, can it be backported to 2.1.X?

## Example project
Project demonstrating the issue is available here: https://github.com/grahammkelly/sb-issue

bclozel:
I've reproduced this issue with a simpler repro:

A sample controller exposing two endpoints:

```
@RestController
public class SampleController {

	private final RestTemplate restTemplate;

	public SampleController(RestTemplate restTemplate) {
		this.restTemplate = restTemplate;
	}

	@GetMapping("/test")
	public String test() {
		return restTemplate.getForObject("http://localhost:8080/{test}", String.class, "resource");
	}

	@GetMapping("/resource")
	public String resource() {
		return "resource";
	}
}
```

An interceptor nesting requests:

```
public class MyInterceptor implements ClientHttpRequestInterceptor {

[...]

PR #19464, opened 2019-12-26
Add support for nested requests in MetricsClientHttpRequestInterceptor

nosan:
See gh-19381

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/client/MetricsClientHttpRequestInterceptor.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/client/MetricsClientHttpRequestInterceptor.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/client/MetricsClientHttpRequestInterceptor.java
```diff
@@ -96,7 +98,9 @@ public ClientHttpResponse intercept(HttpRequest request, byte[] body, ClientHttp
 		finally {
 			getTimeBuilder(request, response).register(this.meterRegistry).record(System.nanoTime() - startTime,
 					TimeUnit.NANOSECONDS);
-			urlTemplate.remove();
+			if (urlTemplate.get().isEmpty()) {
+				urlTemplate.remove();
+			}
```
grahammkelly: How does the previous value get popped? Should this not have 

```java
... 
} else {
   urlTemplate.get.pop()
}
```

grahammkelly:
Looks good and it addresses the issue.

grahammkelly:
Will/can this be backported to the 2.1 branch?

wilkinsona:
We can certainly consider it. IIRC, we only made the change in 2.3.x as we were concerned about the risk of a regression.

Issue #18620, opened 2019-10-16
Regroup thread-related server properties under groups


bclozel:
There are already several thread-related configuration properties for Undertow, Tomcat and Jetty. We should regroup those under dedicated `threads` group to better organize the configuration namespace.

We might also consider other related issues like #18489 at the same time.

bono007:
@bclozel I have a PR almost ready to submit for this. Where can I find out more about Spring Boot property deprecation policy? Specifically, do I need to mark the older properties with `@DeprecatedConfigurationProperty(replacement = "server.tomcat.threads.max")` for example?

snicoll:
@bono007 no specific other than the usual deprecation: the existing property should still be bound and flagged as deprecated. Tests should adapt to the new scheme but you we need to keep a `@Deprecated` test that exercises the deprecated properties to validate they still work. Here is [a concrete example](https://github.com/spring-projects/spring-boot/blob/83e594ad6a581df62284125970b2090abd677991/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java#L113-L133).

snicoll:
Closing in favour of PR #19475

PR #19475, opened 2019-12-28
Add 'threads' configuration group for embedded containers

bono007:
Moves all thread related server properties under a "threads" group for each server implementation that has thread related properties (`tomcat|jetty|undertow`).

The existing property accessors have been marked as deprecated. The old impl just delegates to the new one which allows us to test that fact in a deprecated test and then all subsequent downstream tests can rely on the new api (ie have deprecate test in a single place for each property).

@snicoll do the Boot docs get auto-generated from the `@DeprecatedConfigurationProperty`s? 

Closes gh-18620

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -65,6 +65,7 @@ import org.springframework.util.unit.DataSize;
  * @author Dirk Deyne
  * @author HaiTao Zhang
  * @author Victor Mandujano
+ * @author Chris Bono
  * @since 1.0.0
  */
 @ConfigurationProperties(prefix = "server", ignoreUnknownFields = true)
@@ -309,6 +310,11 @@ public class ServerProperties {
 		private final Accesslog accesslog = new Accesslog();
 
 		/**
+		 * Thread related configuration.
+		 */
+		private final Threads threads = new Threads();
+
+		/**
 		 * Tomcat base directory. If not specified, a temporary directory is used.
 		 */
 		private File basedir;
@@ -319,16 +325,6 @@ public class ServerProperties {
 		 */
 		@DurationUnit(ChronoUnit.SECONDS)
 		private Duration backgroundProcessorDelay = Duration.ofSeconds(10);
-
-		/**
-		 * Maximum amount of worker threads.
-		 */
-		private int maxThreads = 200;
-
-		/**
-		 * Minimum amount of worker threads.
-		 */
-		private int minSpareThreads = 10;
 
 		/**
 		 * Maximum size of the form content in any HTTP post request.
@@ -417,20 +413,26 @@ public class ServerProperties {
 		 */
 		private final Remoteip remoteip = new Remoteip();
 
+		@Deprecated
+		@DeprecatedConfigurationProperty(replacement = "server.tomcat.threads.max")
 		public int getMaxThreads() {
-			return this.maxThreads;
+			return this.getThreads().getMax();
 		}
 
+		@Deprecated
 		public void setMaxThreads(int maxThreads) {
-			this.maxThreads = maxThreads;
+			this.getThreads().setMax(maxThreads);
 		}
 
+		@Deprecated
+		@DeprecatedConfigurationProperty(replacement = "server.tomcat.threads.min-spare")
 		public int getMinSpareThreads() {
-			return this.minSpareThreads;
+			return this.getThreads().getMinSpare();
 		}
 
+		@Deprecated
 		public void setMinSpareThreads(int minSpareThreads) {
-			this.minSpareThreads = minSpareThreads;
+			this.getThreads().setMinSpare(minSpareThreads);
 		}
 
 		@Deprecated
@@ -454,6 +456,10 @@ public class ServerProperties {
 
 		public Accesslog getAccesslog() {
 			return this.accesslog;
+		}
+
+		public Threads getThreads() {
+			return this.threads;
 		}
 
 		public Duration getBackgroundProcessorDelay() {
@@ -863,6 +869,39 @@ public class ServerProperties {
 		}
 
 		/**
+		 * Tomcat thread properties.
+		 */
+		public static class Threads {
+
+			/**
+			 * Maximum amount of worker threads.
+			 */
+			private int max = 200;
+
+			/**
+			 * Minimum amount of worker threads.
+			 */
+			private int minSpare = 10;
+
+			public int getMax() {
+				return this.max;
+			}
+
+			public void setMax(int max) {
+				this.max = max;
+			}
+
+			public int getMinSpare() {
+				return this.minSpare;
+			}
+
+			public void setMinSpare(int minSpare) {
+				this.minSpare = minSpare;
+			}
+
+		}
+
+		/**
 		 * Tomcat static resource properties.
 		 */
 		public static class Resource {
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -1066,44 +1087,57 @@ public void setMaxHttpFormPostSize(DataSize maxHttpFormPostSize) {
 			this.maxHttpFormPostSize = maxHttpFormPostSize;
 		}
 
+		@Deprecated
+		@DeprecatedConfigurationProperty(replacement = "server.jetty.acceptors")
 		public Integer getAcceptors() {
-			return this.acceptors;
+			return this.getThreads().getAcceptors();
```
bono007: A bit of inconsistency on naming and plurality between `Jetty#acceptors/selectors` and `Undertow#io/worker`. Should we 
a) move `acceptors -> acceptor`
b) move `io -> ios` 
c) include `num-` prefix in all of them?
d) leave alone - not a big deal

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -300,6 +301,11 @@ public Session getSession() {
 		 */
 		private final Accesslog accesslog = new Accesslog();
 
+		/**
+		 * Thread related configuration.
+		 */
+		private final Threads threads = new Threads();
```
snicoll: Looking at the code I am wondering now if `pool` wouldn't be better and more consistent with other properties of the same nature.

@bono007 no need to act on that now, I'd like to get some more feedback before asking you to change anything.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -1066,44 +1087,57 @@ public void setMaxHttpFormPostSize(DataSize maxHttpFormPostSize) {
 			this.maxHttpFormPostSize = maxHttpFormPostSize;
 		}
 
+		@Deprecated
+		@DeprecatedConfigurationProperty(replacement = "server.jetty.acceptors")
```
snicoll: That replacement value is wrong.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -1066,44 +1087,57 @@ public void setMaxHttpFormPostSize(DataSize maxHttpFormPostSize) {
...
+			return this.getThreads().getAcceptors();
 		}
 
 		public void setAcceptors(Integer acceptors) {
-			this.acceptors = acceptors;
+			this.getThreads().setAcceptors(acceptors);
 		}
 
+		@Deprecated
+		@DeprecatedConfigurationProperty(replacement = "server.jetty.selectors")
```
snicoll: That replacement value is wrong.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -1563,6 +1675,40 @@ public void setRotate(boolean rotate) {
...
+		/**
+		 * Undertow thread properties.
+		 */
+		public static class Threads {
+
+			/**
+			 * Number of I/O threads to create for the worker. The default is derived from
+			 * the number of available processors.
+			 */
+			private Integer io;
```
snicoll: `io` is a bit odd on its own. I don't have bright ideas but perhaps we should add something in the property name that better conveys it's a number of threads for a certain "type" of threads.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
bono007: Its interesting because both concepts are related but not exclusive of one another. Some servers may not have pool properties, some may have thread properties that are unrelated to pool, etc. 

I think `threads` is more generic than `pool` and most developers will associate that w/ the current thread related properties they are used to.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
bono007: Yep I struggled w/ this balance of brevity and better conveyed meaning. For this one I was thinking `io-per-worker` as that is actually what it is.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
bono007: Good catch. Updated.

@snicoll Boot MR question... Should I rebase and squash subsequent commits to MR based on feedback (like this case) or just add more commits and let Github squash at end of MR?

snicoll: Not a single answer here. When it's about fixing a PR review (as you did), we prefer a single squashed commit rebased on top of the latest state of the target branch (here `master`). When there is a discussion and you'd like to suggest something, a separate commit is better in case we don't pursue with the suggestion.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
bono007: After thinking about this some more, suppose I were to add another `maxQueueDepth` property to Jetty, this would not make sense under `threads` but would make sense under `pool`.

snicoll:
@bono007 now that your other PRs has been merged, would you be willing to rebase your proposal and include the `max-queue-capacity` property that has been introduced for Jetty?

bono007:
Absolutely. I will do that this evening.  MR ready sometime in the next 24
hours.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/ServerPropertiesTests.java
```diff
@@ -207,40 +209,132 @@ void testCustomizeHeaderSizeUseBytesByDefault() {
...
+		assertThat(this.properties.getJetty().getThreads().getIdleTimeout()).hasSeconds(10);
 	}
 
 	@Test
 	void testCustomizeJettyMaxQueueCapacity() {
+		bind("server.jetty.threads.max-queue-capacity", "5150");
+		assertThat(this.properties.getJetty().getThreads().getMaxQueueCapacity()).isEqualTo(5150);
+	}
+
+	@Deprecated
```
bono007: Has this been released anywhere yet? If not, I can remove it - just let me know.

bono007:
@snicoll there were a few comments in the original proposal that we never resolved. I am ok with the proposal as is - if you want anything changed just let me know.

snicoll:
Thanks again @bono007

PR #19476, opened 2019-12-28
Add reference to Codnect Validator

burakkoken:
Codnect Validator is open source library that provides Bean Validation using Spring Expression Language (SpEL). Especially, It’s very useful for cross-field validations in your classes. If you would like to check usage of it and get more information, check https://github.com/Codnect/codnect-validator/blob/master/README.md

### DIFF
--- a/spring-boot-project/spring-boot-starters/README.adoc
+++ b/spring-boot-project/spring-boot-starters/README.adoc
@@ -61,6 +61,9 @@ do as they were designed before this was clarified.
 | https://cloudant.com/[Cloudant]
 | https://github.com/icha024/cloudant-spring-boot-starter
 
+| Codnect Validator for Spring Boot
+| https://github.com/Codnect/codnect-validator
+
 | https://www.couchbase.com/[Couchbase] HTTP session
 | https://github.com/mkopylec/session-couchbase-spring-boot-starter
 


pivotal-issuemaster:
@burakkoken Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19476)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19476) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@burakkoken Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19476)!

snicoll:
Thanks for the PR but that list is about 3rd party spring boot starters only.

I haven’t seen any auto-configuration in this project. If I’ve missed that please let me know and we can reconsider.

burakkoken:
First of all thanks for your comment,
This library only provides Bean Validation (JSR 303/349) constraint, I mean it includes an annotation named Assert and its contraint validator. The constraint I have mentioned is already configured by hibernate validator. Can you please check again? If you like, you can check the code
https://github.com/Codnect/codnect-validator/blob/master/src/main/java/org/codnect/validator/annotation/Assert.java
https://github.com/Codnect/codnect-validator/blob/master/src/main/java/org/codnect/validator/AssertValidator.java

snicoll:
Thanks for the feedback. That’s not a starter per definition and therefore does not belong to this list. Thanks anyway.

burakkoken:
I see okay, So what should I do to add this library code to spring project? I mean the contraint. Because I think it's very useful for validation and it's going to be good for developers. In addititon, I would like to contribute. Thanks.

snicoll:
If you want to make it a starter, you can create one of your own that brings the library and let a Spring Boot application use constraint validations. The way I understood the project though, there is no need for auto-configuration and hence no need for a starter. 

This library is rather new so we can't consider providing dependency management or add it to the core at this point. And that's okay, there are tons of third party stuff we don't manage ourselves.

Issue #19493, opened 2019-12-31
Add max capacity configuration for embedded Jetty QueuedThreadPool.


bono007:
We are replacing our legacy in-house Java application framework with Spring Boot. The last missing feature parity is the ability to constrain Jetty's backing queue. 

Issue #8662 was the same issue from a couple years back. It landed on [this comment](https://github.com/spring-projects/spring-boot/issues/8662#issuecomment-287706410) from @wilkonsa

>  A pull request that adds this configuration option for Jetty, Tomcat, and Undertow would be welcome.

However, the issue ended up closing due to inactivity. I intend to submit that PR.

snicoll:
@bono007 thanks but there is no need to create an issue and a PR as one supersedes the other. 

Closing in favour of PR #19494

bono007:
Got it. 
I was curious about that @snicoll . In the case where I go straight to a PR I just skip the "Closes gh-abc" text in the commit message I am assuming,

snicoll:
The commit message of a PR we process is almost always rewritten. We rewrite the `closes` if pressent as the issue is closed (like here) and we use the PR issue id as the reference in the commit message.

Issue #8662, opened 2017-03-18
Embedded Jetty uses unbounded QueuedThreadPool by default.


jzampieron:
The default behavior for jetty when used in spring boot is for the
JettyEmbeddedServletContainerFactory to create an Jetty Server with an *unbounded* 
QueuedThreadPool.

This means that under any kind of load where the input request rates exceeds the servers ability to process that the JVM will _eventually_ OOM no matter what.

Essentially, this is a guaranteed breakage by default. 

Instead it would be better to replace the default BlockingQueue inside the QueuedThreadPool with a bounded queue, as is advised in the Jetty documentation. 

This changes the default behavior to load shed (fail fast) and is far more production ready out of the box.

(See https://wiki.eclipse.org/Jetty/Howto/High_Load#Jetty_Tuning - dated by still reflected in the code.)

wilkinsona:
> The default behavior for jetty when used in spring boot

I believe that it's the default behaviour with Jetty with or without Spring Boot.

> Instead it would be better to replace the default BlockingQueue inside the QueuedThreadPool with a bounded queue, as is advised in the Jetty documentation.

The problem with specifying an upper-bound for the queue is that we have no way of knowing what it should be. From the same documentation that you linked to above:

> The capability (maximum queue length) should be calculated according to the "no-response" time tolerable. For example, if the webapp can handle 100 requests per second, and if you can allow it one minute to recover from excessive high load, you can set the queue capability to 60*100=6000

That equation has two variables and both are unknown to us. If we resort to a guess we artificially constrain performance for some while not solving the problem for others.

How do you propose that we constrain the size of the queue in a way that's generally useful?

jzampieron:
I propose that we take the same opinionated view that Spring Boot typically takes. 

Assume that most applications are RESTful microservices. A fast REST API call takes ~10ms (database round trip, json ser/des, etc). Therefore your upper bound will be a 100Hz request rate.

Thymeleaf rendering and other MVC might take longer, so that's a reasonable starting point.

Allow 30 seconds for back log and go with `30*100 = 3000` for the default. 

Expose a `server.jetty.maxQueueLength` property and document in the spring boot property appendix with a link to the Jetty eclipse documentation.

jzampieron:
Also, specifically, it _is_ the default behavior for Jetty with or without Spring boot.

[...]

PR #19494, opened 2019-12-31
Adds configuration for embedded Jetty max queue capacity

bono007:
Closes gh-19493

This is [that](https://github.com/spring-projects/spring-boot/issues/8662#issuecomment-287706410) pull request for Jetty. I think this property is Jetty specific and does not have a direct equivalent or current need in Tomcat or Undertow.

### Approach
Create a `JettyThreadPoolFactory` interface that is used to set the backing thread pool on the `Jetty[Servlet|Reactive]WebServerFactory`. 

This was needed because the thread pool's backing queue (which is where the max capacity is set) can not be safely modified in a customizer as by the time the customizer is called, the server has already had its `QueuedThreadPool` and its backing queue implementation set during construction - its immutable after that.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/UndertowServletWebServerFactoryCustomizer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/UndertowServletWebServerFactoryCustomizer.java
@@ -16,6 +16,11 @@
 
 package org.springframework.boot.autoconfigure.web.servlet;
 
+import io.undertow.Handlers;
+import io.undertow.server.HandlerWrapper;
+import io.undertow.server.HttpHandler;
+import io.undertow.server.handlers.RequestLimitingHandler;
+
 import org.springframework.boot.autoconfigure.web.ServerProperties;
 import org.springframework.boot.web.embedded.undertow.UndertowServletWebServerFactory;
 import org.springframework.boot.web.server.WebServerFactoryCustomizer;
@@ -25,6 +30,7 @@ import org.springframework.boot.web.server.WebServerFactoryCustomizer;
  * Servlet web servers.
  *
  * @author Andy Wilkinson
+ * @author Chris Bono
  * @since 2.1.7
  */
 public class UndertowServletWebServerFactoryCustomizer
@@ -36,10 +42,55 @@ public class UndertowServletWebServerFactoryCustomizer
 		this.serverProperties = serverProperties;
 	}
 
+	// TODO why was this not being used before now?
+
 	@Override
 	public void customize(UndertowServletWebServerFactory factory) {
+
 		factory.addDeploymentInfoCustomizers((deploymentInfo) -> deploymentInfo
 				.setEagerFilterInit(this.serverProperties.getUndertow().isEagerFilterInit()));
+
+		addRateLimiterIfMaxRequestsSet(factory);
+	}
+
+	/**
+	 * Installs a {@link RequestLimitingHandler} in the initial handler chain if
+	 * {@link ServerProperties.Undertow#maxRequests} is set. If installed, it will use
+	 * {@link ServerProperties.Undertow#maxQueueCapacity} to control the size of the queue
+	 * in the rate limting handler.
+	 * @param factory the factory to add the deployment info customizer on
+	 */
+	private void addRateLimiterIfMaxRequestsSet(UndertowServletWebServerFactory factory) {
+		Integer maxRequests = this.serverProperties.getUndertow().getMaxRequests();
+		if (maxRequests == null) {
+			return;
+		}
+		Integer maxQueueCapacity = this.serverProperties.getUndertow().getMaxQueueCapacity();
+		factory.addDeploymentInfoCustomizers((deploymentInfo) -> deploymentInfo.addInitialHandlerChainWrapper(
+				new RequestLimiterHandlerWrapper(maxRequests, (maxQueueCapacity != null) ? maxQueueCapacity : -1)));
+	}
+
+	/**
+	 * Explicit implementation of HandlerWrapper rather than Lambda to make it possible to
+	 * verify the handler wrappers at runtime during tests. This could be instead written
+	 * as a Lambda but then tests can not see what type of handler wrapper it is.
+	 */
+	static class RequestLimiterHandlerWrapper implements HandlerWrapper {
+
+		Integer maxRequests;
+
+		Integer maxQueueCapacity;
+
+		RequestLimiterHandlerWrapper(Integer maxRequests, Integer maxQueueCapacity) {
+			this.maxRequests = maxRequests;
+			this.maxQueueCapacity = maxQueueCapacity;
+		}
+
+		@Override
+		public HttpHandler wrap(HttpHandler handler) {
+			return Handlers.requestLimitingHandler(this.maxRequests, this.maxQueueCapacity, handler);
+		}
+
 	}
 
 }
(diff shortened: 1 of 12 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveWebServerFactoryAutoConfigurationTests.java
```diff
@@ -63,6 +68,7 @@
  * @author Brian Clozel
  * @author Raheela Aslam
  * @author Madhura Bhave
+ * @author Chris Bono
```
bono007: At what level of update to a file are we supposed to add our name as `@author`? I am really not trying to splat my name on every single file - just not sure what the recommended theshold is.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyThreadPoolFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Factory to create a thread pool for use by a Jetty {@link Server}.
+ *
+ * @author Chris Bono
+ * @since 2.3.0
+ * @param <T> type of ThreadPool that the factory returns
+ */
+@FunctionalInterface
+public interface JettyThreadPoolFactory<T extends ThreadPool> {
```
bono007: What is the team's stance on generics in general in Boot? I know I had added some previously in a different issue/pr but they were removed in a polish commit. I am not too strongly opinionated in this particular instance as it really only cleans up some casting from tests. The main area thats is used (`EmbeddedJetty`s) does not take advantage of it and simply uses `@ConditionalOnMissingBean(JettyThreadPoolFactory.class)`.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveWebServerFactoryAutoConfigurationTests.java
snicoll: Thanks for asking. Please add your name only if you've significantly changed the class. Adding a property doesn't qualify as such. Polishing either. This one is legit as you're adding a number of tests.

snicoll:
@bono007 the issue you've referenced has [the following](https://github.com/spring-projects/spring-boot/issues/8662#issuecomment-287706410):

> A pull request that adds this configuration option for Jetty, Tomcat, and Undertow would be welcome.

What makes you think there isn't an equivalent for Tomcat or Undertow? I think the whole discussion on that thread is that whatever mechanism we introduce should be introduced consistently.

bono007:
@snicoll 

Sorry for the lack of explanation. Here was my thinking:

1) For Tomcat, the property already exists and is `server.tomcat.accept-count`. 

2) For Undertow, I clearly did not understand the architecture last night at first look and thought it was bounded. I do a little more now and here is what I have found:

Undertow's worker thread pool is unbounded but it is **NOT** currently configurable in Nio/XnioWorker.
```
org.xnio.XnioWorker#XnioWorker

this.taskQueue = new LinkedBlockingQueue();  // If only we could do a new LinkedBlockingQueue(SomeWorkerOptionForMaxCapacityHerePlease)
this.taskPool = new XnioWorker.TaskPool(threadCount, threadCount, (long)optionMap.get(Options.WORKER_TASK_KEEPALIVE, 60000), TimeUnit.MILLISECONDS, 							this.taskQueue, new XnioWorker.WorkerThreadFactory(threadGroup, optionMap, markThreadAsDaemon), new AbortPolicy());
```
 There is an option on `io.undertow.Undertow.Builder` to set the worker but it accepts an `XnioWorker` - again w/o the ability to configure its internal task queue. They do provide a [RequestLimitingHandler](http://undertow.io/undertow-docs/undertow-docs-1.4.0/index.html#built-in-handlers) which could be used to make sure that no more than N concurrent requests can be handled at once - which would roughly translate to the same as bounding the internal queue. 

### Summary
Mechanisms to set the "max queue capacity" (which will need to be renamed something besides "queue capacity" at this point):

Jetty: use our custom thread pool (this PR)
Tomcat: use the existing property
Undertow: set the request limiting handler (TBD)

Do we want to consolidate this functionality into this common property or handle the Tomcat and Undertow w/ documentation? For Tomcat its simple enough to set the other property. For Undertow though I could see doing a convenience default like I did for Jetty.

The underlying mechanisms vary enough that trying to abstract something on top of them  beyond this common property does not seem like a big win. Tomcat has this accept count. Jetty has this custom `JettyThreadPool`. Undertow has handlers to simulate the max capacity. If they all has the concept of an exposed "thread pool" then I could see adding a "thread pool customizer" etc.. 

### Proposal
* Solve Jetty as I have done in this PR.
* Tomcat - already done w/ its `server.tomcat.accept-count`
* Add code to translate a `sever.undertow.max-concurrent-requests` into a RateLimitingHandler setup

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/UndertowServletWebServerFactoryCustomizer.java
```diff
@@ -36,10 +42,55 @@ public UndertowServletWebServerFactoryCustomizer(ServerProperties serverProperti
 		this.serverProperties = serverProperties;
 	}
 
+	// TODO why was this not being used before now?
```
bono007: I noticed this Undertow specific customizer exists but was not being called anywhere. Will this be a problem?

bono007:
I have created a 2nd [commit](https://github.com/spring-projects/spring-boot/pull/19494/commits/f17a44230bb684b0d9defa9b2d68474702e2e71e) that does the same type of resource limiting for Undertow. However, it is a completely different mechanism in Undertow than in Jetty and Tomcat. In those servers, its a thread pool with min/max thread count and backing queue that we limit the capacity on. However, in Undertow we don't have access to the queue in the blocking thread pool that it uses. The recommended method is to install a [RequestLimitingHandler](http://undertow.io/undertow-docs/undertow-docs-2.0.0/#request-limiting-handler) to constrain the requests/queued up callers. Additionally I think it only is available in Servlet environment (no DeploymentInfoCustomizer in UndertowReactiveWebServerFactory). I will look into this a bit more but wanted to get this code in for Undertow so reviewers could see the differences in implementation for each server.

The main takeaway for me when comparing the 3 servers and the 3 mechanism to do this is that they are all different enough that I don't think it makes sense to try to unify them all under a single property. It would be artificial at best for Undertow specifically.

### Summary 
Jetty is solved w/ the previous commit and allows the following "dials":
```
server.jetty.max-threads=200
server.jetty.min-threads=8
server.jetty.max-queue-capacity=
```

Tomcat is already solved and allows the following "dials":
```
server.tomcat.max-threads=200
server.tomcat.min-spare-threads=8
server.tomcat.accept-count=100
```

Undertow is the 2nd commit and allows the following "dials":

```
server.undertow.io-threads
server.undertow.worker-threads
server.undertow.max-requests (NEW)
server.undertow.max-queue-capacity (NEW)
```

wilkinsona:
Thanks for the analysis, @bono007. In light of what you've said, and the differences in how this works in Undertow and Jetty and Tomcat, I think we should just tackle Jetty (for now at least) using `server.jetty.*` configuration rather than trying to do something across all three containers.

snicoll:
Thanks for the pr @bono007. I've reduced the amount of change by making sure a `ThreadPool` can be set without having to resort to a Servlet or Reactive implementation. That way, the existing customizer could be used to perform that job. It was already so now it's streamline in one place.

PR #19521, opened 2020-01-03
Spring MVC auto-config doc update

rstoyanchev:
The documentation refers to `WebMvcRegistrationsAdapter` which no longer exists. I've also taken the opportunity to improve readability.


snicoll:
Thanks @rstoyanchev! Brian, I took the liberty of backporting since I think this is generally useful for the `2.2.x` doc as well.

Issue #19524, opened 2020-01-03
Non-compatible bean instance of type [org.neo4j.ogm.session.SessionFactory]  being injected for [org.springframework.data.cassandra.SessionFactory]


aquasika:
I am using both spring-boot-starter-data-cassandra and  spring-boot-starter-data-neo4j.
Also, I am using multi keyspaces for cassandra so had to overrride AbstractCassandraConfiguration.
However, when I am starting application, I get error resolving SessionFactory.

It appears there is a clash of SessionFactory  and org.neo4j.ogm.session.SessionFactory is being injected into cassandra.SessionFactory. 

```
Caused by: java.lang.IllegalStateException: @Bean method AbstractCassandraConfiguration.sessionFactory called as bean reference for type [org.springframework.data.cassandra.SessionFactory] but overridden by non-compatible bean instance of type [org.neo4j.ogm.session.SessionFactory]. Overriding bean of same name declared in: class path resource [org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration$Neo4jOgmSessionFactoryConfiguration.class]```

**Comment by nosan** [comment]
It is expected behavior, `Neo4jDataAutoConfiguration` creates a bean named sessionFactory and `AbstractCassandraConfiguration` does the same.  
Even more, this behavior is forbidden in Spring Boot and you have to set (`spring.main.allow-bean-definition-overriding=true`) 

Currently, you can override the `sessionFactory` method in `AbstractCassandraConfiguration` and rename the bean name with `@Bean` annotation.


```java
@Configuration
public class CassandraConfiguration extends AbstractCassandraConfiguration {


	@Bean("cassandraSessionFactory")
	@Override
	public SessionFactory sessionFactory() {
		return super.sessionFactory();
	}
 //other methods

}
```
The other way is to create your own Neo4j `SessionFactory` with a different name.

nosan:
Maybe it makes sense to rename `sessionFactory` bean in `Neo4jDataAutoConfiguration` to **`neo4jSessionFactory`** to avoid such collisions.

snicoll:
Indeed, I don't think it is expected behaviour that such beans collide with each other. This unfortunately happens with such generic names and also a reason why we forbid bean overriding by default as it's easier for folks to report problems like this to us.

nosan:
> I don't think it is expected behaviour that such beans collide with each other

I meant bean-overriding.

Don't you mind if I create a PR?

[...]

PR #19544, opened 2020-01-06
Rename Neo4j bean names

nosan:
see gh-19524

Here is the second approach: https://github.com/nosan/spring-boot/commit/9c666b5a06a644a46ce35ff420e534070fbf2d73

I am not sure which way is better. Both variants support back-compatibility.


nosan:
I think it makes sense to rename`sessionFactory` to **`neo4jSessionFactory`** for `2.3.X` version
with no back-compatibility.

snicoll:
@nosan thanks for the PR. We've discussed it and we'd like to rename the names in `2.2.x` (with no back-compatibility as you indicated). Do you have time to review your PR in that direction?

nosan:
Sure @snicoll,
I let you know

nosan:
@snicoll 
PR has been updated

snicoll:
Thanks for the PR @nosan but looking at what's going on behind the scenes, a change in Spring Data is required. Once we have that in, we'll have to reconsider this from scratch.

PR #19556, opened 2020-01-07
Profile with comma in name does not work in tests

### LINKED ISSUE #19537: @SpringBootTest does not work when @ActiveProfiles contains a comma
When I define the active profiles like this:

    @ActiveProfiles("a,b")

on a Spring Boot Test (noticed on a `@WebMvcTest`), the correct profiles are still applied. This is because of https://github.com/spring-projects/spring-boot/blob/e250cce2744ea226761a20bcf2694e6b4244fc0c/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootContextLoader.java#L159

I'm reporting this because it actually confused the hell out of me while I was introducing a newbie to Spring (Boot) and testing support etc. I'm actually not sure whether this is supposed to work like this or not, seems to be an unknown side effect, and I'm not sure whether this is really an issue. This would not work, for instance, if there were a profile with a comma in its name - but no idea if this is allowed either.

scottfrederick:
Prior to this commit, active profiles were being added to the Spring Boot application environment by setting the `spring.profiles.active` property. This could result in profiles getting parsed differently than other uses of `@ActiveProfiles`. Setting the profiles directly in the `Environment` prevents this parsing.

Fixes #19537

### DIFF
--- a/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/context/SpringBootContextLoaderTests.java
+++ b/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/context/SpringBootContextLoaderTests.java
@@ -22,6 +22,8 @@ import org.junit.Ignore;
 import org.junit.Test;
 
 import org.springframework.context.annotation.Configuration;
+import org.springframework.core.env.Environment;
+import org.springframework.test.context.ActiveProfiles;
 import org.springframework.test.context.ContextConfiguration;
 import org.springframework.test.context.MergedContextConfiguration;
 import org.springframework.test.context.TestContext;
@@ -35,6 +37,7 @@ import static org.assertj.core.api.Assertions.assertThat;
  * Tests for {@link SpringBootContextLoader}
  *
  * @author Stephane Nicoll
+ * @author Scott Frederick
  */
 public class SpringBootContextLoaderTests {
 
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/context/SpringBootContextLoaderTests.java
```diff
@@ -88,18 +91,49 @@ public void environmentPropertiesNewLineInValue() {
...
+	@Test
+	public void noActiveProfiles() {
+		Environment environment = getApplicationEnvironment(SimpleConfig.class);
+		assertThat(environment.getActiveProfiles()).isEmpty();
+	}
+
+	@Test
+	public void multipleActiveProfiles() {
+		Environment environment = getApplicationEnvironment(MultipleActiveProfiles.class);
+		assertProfilesContains(environment, "profile1", "profile2");
```
wilkinsona: I'd find this easier to read as `assertThat(environment.getActiveProfiles()).containsExactly("profile1,2");` rather than calling the method, but that might just be me and I have no idea what the rest of the team might prefer.

spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/context/SpringBootContextLoaderTests.java
```diff
@@ -88,18 +91,49 @@ public void environmentPropertiesNewLineInValue() {
...
+	}
+
+	@Test
+	public void activeProfileWithComma() {
+		Environment environment = getApplicationEnvironment(ActiveProfileWithComma.class);
+		assertProfilesContains(environment, "profile1,2");
```
wilkinsona: As [above](https://github.com/spring-projects/spring-boot/pull/19556/files#r363847081).

spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/context/SpringBootContextLoaderTests.java
snicoll: Same

scottfrederick: made it so

wilkinsona:
This change has been reverted in 2.1.13 as it introduced a regression. See #19788 for details.

Issue #19192, opened 2019-11-29
DataSource url property is ignored when there is no connection pool


filiphr:
When there are multiple drivers on the classpath (e.g. h2 and postgres) and no connection pool is present. Then `spring.datasource.url` is not used, instead the embedded database is used.

I would expect that the datasource that is created is connected with the provided datasource URL. 

I was debugging this a bit and from the `DataSourceAutoConfiguration` the `PooledDataSourceConfiguration` is not active since there is no connection pool. This means that `EmbeddedDatabaseConfiguration` would be active leading to `EmbeddedDataSourceConfiguration` to create the datasource and that one then creates `EmbeddedDatabase`. 

This has clearly been an error on our side (forgetting to add the connection pool). 

I'd suggest that the `EmbeddedDataSourceConfiguration` is actually renamed to `NonPooledDataSourceConfiguration` and then based on whether a datasource url is defined or not a `SimpleDriverDataSource` is used or the current behaviour is kept.

wilkinsona:
Closing in favour of #19576.

filiphr:
Thanks for fixing this @snicoll.

PR #19576, opened 2020-01-08
DataSource url property is ignored when there is no connection pool

nosan:
see gh-19192

Btw, I found an issue in `DataSourceBuilder`, the builder does not support `SimpleDriverDataSource` without alias `aliases.addAliases("driver-class-name", "driver-class")`

Here is a test to reproduce:
```java

	@Test
	void simpleDriverDataSource() {
		this.dataSource = DataSourceBuilder.create().url("jdbc:h2:test").type(SimpleDriverDataSource.class).build();
		assertThat(this.dataSource).isInstanceOf(SimpleDriverDataSource.class);
		assertThat(this.dataSource).extracting("driver").isNotNull();
	}
```
`

**UPDATE**

My first approach was incorrect. Sorry for the inconvenience.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,13 +20,17 @@ import javax.sql.DataSource;
 
 import com.zaxxer.hikari.HikariDataSource;
 
+import org.springframework.boot.autoconfigure.condition.AnyNestedCondition;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.jdbc.DataSourceBuilder;
 import org.springframework.boot.jdbc.DatabaseDriver;
 import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.jdbc.datasource.SimpleDriverDataSource;
 import org.springframework.util.StringUtils;
 
 /**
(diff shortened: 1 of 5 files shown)


snicoll:
Thanks for the PR @nosan.

> I found an issue in DataSourceBuilder, the builder does not support SimpleDriverDataSource

It's not meant to at the moment. The supported connections pool are commons-dbcp2, tomcat and hikari (default).

nosan:
Thanks for the feedback, @snicoll 
>  It's not meant to at the moment. The supported connection pools are commons-dbcp2, tomcat and hikari (default).

Yes, I saw that when I was working on this PR.
Consider this test in `DataSourceAutoConfigurationTests`:
```java
@Test
void explicitTypeSupportedDataSource() {
	this.contextRunner
			.withPropertyValues("spring.datasource.driverClassName:org.hsqldb.jdbcDriver",
					"spring.datasource.url:jdbc:hsqldb:mem:testdb",
					"spring.datasource.type:" + SimpleDriverDataSource.class.getName())
			.run(this::containsOnlySimpleDriverDataSource);
}
```

Actually, this test just checks that `SimpleDriverDataSource` will be created, and actually, it will, but still, if you try to `getConnection` from this `DataSource` you will get an exception:

```
java.lang.IllegalArgumentException: Driver must not be null

	at org.springframework.util.Assert.notNull(Assert.java:198)
	at org.springframework.jdbc.datasource.SimpleDriverDataSource.getConnectionFromDriver(SimpleDriverDataSource.java:139)
	at org.springframework.jdbc.datasource.AbstractDriverBasedDataSource.getConnectionFromDriver(AbstractDriverBasedDataSource.java:205)
	at org.springframework.jdbc.datasource.AbstractDriverBasedDataSource.getConnection(AbstractDriverBasedDataSource.java:169)
	at org.springframework.boot.autoconfigure.jdbc.DataSourceAutoConfigurationTests.containsOnlySimpleDriverDataSource(DataSourceAutoConfigurationTests.java:172)
```

This happens because `DataSourceBuilder` does not add alias:
 ```
aliases.addAliases("driver-class-name", "driver-class");
```


**P.S.** I think you are right and this is not related to the initial issue.

**Comment by nosan** [comment]
As a variant instead of:
```java
DataSourceBuilder<?> builder = properties.initializeDataSourceBuilder();
if (properties.getType() == null && DataSourceBuilder.findType(properties.getClassLoader()) == null) {
	builder.type(SimpleDriverDataSource.class);
}
return builder.build();
```

Create a new method in `DataSourceProperties`:

```java
/**
 * Determine the datasource type to use based on this configuration and the
 * environment.
 * @return the datasource type to use
 * @since 2.1.12
 */
public Class<? extends DataSource> determineType() {
	Class<? extends DataSource> type = getType();
	if (type == null) {
		type = DataSourceBuilder.findType(getClassLoader());
	}
	return (type != null) ? type : SimpleDriverDataSource.class;
}
```

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceConfiguration.java
```diff
@@ -114,12 +118,39 @@ HikariDataSource dataSource(DataSourceProperties properties) {
...
+			Class<? extends DataSource> type = properties.getType();
+			if (type == null) {
+				type = DataSourceBuilder.findType(properties.getClassLoader());
+			}
+			return (type != null) ? type : SimpleDriverDataSource.class;
+		}
+
+	}
+
+	static class GenericCondition extends AnyNestedCondition {
```
wilkinsona: We should rename this when merging. Something like `ExplicitDataSourceTypeOrUrlCondition`.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/jdbc/DataSourceAutoConfigurationTests.java
```diff
@@ -158,6 +158,20 @@ void explicitTypeNoSupportedDataSource() {
 				.run(this::containsOnlySimpleDriverDataSource);
 	}
 
+	/**
+	 * This test makes sure that if no supported pool data source is present, a datasource
```
wilkinsona: If possible, when merging, we should replace this comment with a descriptive test method name

snicoll:
I was looking at this PR and I am not keen to bring support for `SimpleDataSource` back in `2.1.x`. I'd rather let it fail which is what would happen if there wasn't an embedded driver on the claspath in the first place.

Right now, it would fail as follow:

```
Caused by: java.lang.IllegalStateException: No supported DataSource type found
 	at org.springframework.boot.jdbc.DataSourceBuilder.getType(DataSourceBuilder.java:140)
 	at org.springframework.boot.jdbc.DataSourceBuilder.build(DataSourceBuilder.java:72)
 	at org.springframework.boot.autoconfigure.jdbc.DataSourceConfiguration$Generic.dataSource(DataSourceConfiguration.java:125)
 ```

I suggest to improve that so that our  failure analyzer catch that and provide a dedicated error message.

We can then add support for `SimpleDriverDataSource` in `2.3.x` as a fallback for this scenario. Adding such support is required for r2dbc anyway.

filiphr:
@snicoll the problem is not about not having an embedded driver on the classpath, but rather having a wrong or missing a connection pool. Which leads to using a wrong database even though you have configured your datasource url.

Originally I opened issue #19192 which was closed in favour of this PR.

snicoll:
Thank you, I am aware of the problem and I read the related issue already.

snicoll:
@nosan thank you for your effort on this but I am not keen to bring `SimpleDriverDataSource`  in `2.1.x`. Looking at the issue, avoiding the embedded configuration to kick in if an url is set sounds like a safer and more isolated change to me. I've just done that and I've reopened the related issue. Thanks again.

PR #19584, opened 2020-01-08
Simplify validation pattern of EndpointId

matthiasstock:
The literals `.` (dot) and `-` (hyphen) are escaped within the character classes. While a dot is not a meta character within a character class, a hyphen does not need to be escaped unless it can be interpreted as a meta character (usually at the end).

These changes do not change the behavior, but improve readability and understandability. I have added an explicit test only to avoid uncertainties for possible future escape characters.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/EndpointId.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/EndpointId.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


philwebb:
Thanks, those are nice improvements. I've found [regex101.com] quite useful for validating patterns as well. [Here's](https://regex101.com/r/429pgC/1) the output for `[a-zA-Z0-9.-]+`:

```
Match a single character present in the list below [a-zA-Z0-9.-]+
+ Quantifier — Matches between one and unlimited times, as many times as possible, giving back as needed (greedy)
a-z a single character in the range between a (index 97) and z (index 122) (case sensitive)
A-Z a single character in the range between A (index 65) and Z (index 90) (case sensitive)
0-9 a single character in the range between 0 (index 48) and 9 (index 57) (case sensitive)
.- matches a single character in the list .- (case sensitive)
```

Issue #19518, opened 2020-01-03
When the system time to package compilation environment is date 1970,the jar will run error


BenGAO:
When the system time to package compilation environment is date 1970,the jar will run error：

```
          Exception in thread "main" java.time.DateTimeException: Invalid value for MonthOfYear (valid 
            values 1 - 12): 0
        at java.time.temporal.ValueRange.checkValidValue(ValueRange.java:311)
        at java.time.temporal.ChronoField.checkValidValue(ChronoField.java:703)
        at java.time.LocalDate.of(LocalDate.java:267)
        at java.time.LocalDateTime.of(LocalDateTime.java:336)
        at org.springframework.boot.loader.jar.CentralDirectoryFileHeader.decodeMsDosFormatDateTime(CentralDirectoryFileHeader.java:127)
        at org.springframework.boot.loader.jar.CentralDirectoryFileHeader.getTime(CentralDirectoryFileHeader.java:116)
        at org.springframework.boot.loader.jar.JarEntry.<init>(JarEntry.java:58)
        at org.springframework.boot.loader.jar.JarFileEntries.getEntry(JarFileEntries.java:316)
        at org.springframework.boot.loader.jar.JarFileEntries.access$400(JarFileEntries.java:48)
        at org.springframework.boot.loader.jar.JarFileEntries$EntryIterator.next(JarFileEntries.java:366)
        at org.springframework.boot.loader.jar.JarFileEntries$EntryIterator.next(JarFileEntries.java:350)
        at org.springframework.boot.loader.jar.JarFile$2.nextElement(JarFile.java:204)
        at org.springframework.boot.loader.jar.JarFile$2.nextElement(JarFile.java:195)
        at org.springframework.boot.loader.archive.JarFileArchive$EntryIterator.next(JarFileArchive.java:189)
        at org.springframework.boot.loader.archive.JarFileArchive$EntryIterator.next(JarFileArchive.java:174)
        at org.springframework.boot.loader.archive.JarFileArchive.getNestedArchives(JarFileArchive.java:85)
        at org.springframework.boot.loader.ExecutableArchiveLauncher.getClassPathArchives(ExecutableArchiveLauncher.java:69)
        at org.springframework.boot.loader.Launcher.launch(Launcher.java:50)
        at org.springframework.boot.loader.JarLauncher.main(JarLauncher.java:52)
```

[...]

PR #19595, opened 2020-01-09
Limit ChronoField values ​​to their range

nosan:
see gh-19518

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/CentralDirectoryFileHeader.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/CentralDirectoryFileHeader.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,8 +17,11 @@
 package org.springframework.boot.loader.jar;
 
 import java.io.IOException;
-import java.time.LocalDateTime;
 import java.time.ZoneId;
+import java.time.ZonedDateTime;
+import java.time.temporal.ChronoField;
+import java.time.temporal.ChronoUnit;
+import java.time.temporal.ValueRange;
 
 import org.springframework.boot.loader.data.RandomAccessData;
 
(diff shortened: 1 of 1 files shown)


philwebb:
We should add a test for the original issue as well if possible when we merge this.

nosan:
Sure, could you please point me out, where the best place to add tests?

philwebb:
Very good question. It doesn't look like we have a `CentralDirectoryFileHeaderTests` class yet. We'll have a look when we merge to see if we can find a good place to add them.

mbhave:
We might be able to add it to `JarFileTests`.

nosan:
Unfortunately,  I haven't found a way to test this :(
I would appreciate any ideas regarding this.

mbhave:
Thanks for letting us know @nosan. I'm currently trying to write a test for it and will let you know how far I get.

nosan:
Thanks @mbhave

mbhave:
Thank you for your contribution @nosan!

venkatasreekanth:
Is there a workaround for this?

wilkinsona:
@venkatasreekanth As far as we know, this issue no longer occurs with Spring Boot 2.1.12 and later. If you are using a [supported version of Spring Boot](https://spring.io/projects/spring-boot#support) (2.5.x or later for OSS support) and the problem is occurring, please open a new issue with a minimal sample that reproduces the problem and we can take a look.

venkatasreekanth:
> @venkatasreekanth As far as we know, this issue no longer occurs with Spring Boot 2.1.12 and later. If you are using a [supported version of Spring Boot](https://spring.io/projects/spring-boot#support) (2.5.x or later for OSS support) and the problem is occurring, please open a new issue with a minimal sample that reproduces the problem and we can take a look.

I making changes to a legacy project which runs spring boot 2.0.6 and use spring batch on SCDF 1.7. Looking for a workaround for now. We are upgrading to latest SCDF and all the batch projects in the near future.

wilkinsona:
The only thing that I can think of would be to identify the jar with the entries with a modification date of 0 and modifying it so that's no longer the case. Unzipping it and and zipping it be up again may be enough. Failing that, you'll either have to upgrade the version of Spring Boot that you're using or avoid depending on a jar that has entries with a modification date of 0. If you have any further questions, please follow up on Stack Overflow or [Gitter](https://gitter.im/spring-projects/spring-boot). As mentioned in [the guidelines for contributing](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc#using-github-issues), we prefer to use GitHub issues only for bugs and enhancements.

PR #19620, opened 2020-01-10
Reuse optional configuration name constant

vpavic:


wilkinsona:
Thanks again, @vpavic.

PR #19623, opened 2020-01-10
Add missing @since tag in DisabledIfDockerUnavailable

dreis2211:
Hi,

this PR fixes the current checkstyle issue in the builds.

Cheers,
Christoph


dreis2211:
Fixed by @bclozel with e0171a5da813f5e514937f6ee9d96d7f7400ffa1

wilkinsona:
FWIW, I don’t think we need one as the code isn’t public API.

PR #19632, opened 2020-01-11
Make Tomcat javadoc link version dependent

dreis2211:
Hi,

the move to Gradle allows us to have a bit more flexible Javadoc links. Tomcat is only using the MAJOR.MINOR part of the version and achieving this wasn't really possible with Maven without introducing another (somewhat duplicated) property.

This PR makes the Tomcat links dependent on the Tomcat version instead of hardcoding the link. If we upgrade eventually to a newer version of Tomcat there is no need to update the link anymore.

Let me know what you think.
Cheers,
Christoph


snicoll:
Sweet, thank you again @dreis2211

Issue #19636, opened 2020-01-11
PoolingDataSourceBean.getParentLogger() method Dead cycle


zxm812:
When I was browsing the source code, I found that PoolingDataSourceBean class of method getParentLogger there was a Dead cycle code。

wilkinsona:
Thank you for reporting this, @zxm812. Closing in favour of #19637.

PR #19637, opened 2020-01-11
PoolingDataSourceBean getParentLogger throws StackOverflowError

nosan:
see gh-19636


nosan:
@wilkinsona 
Actually it does not return a global logger 😄 , it always throws `StackOverflowError`

wilkinsona:
Thanks, @nosan. Because it's catching `Exception` rather than `Throwable`.

snicoll:
Thanks a lot @nosan

PR #19658, opened 2020-01-12
Pass CI branch parameter to build scan

### LINKED ISSUE #19614: git branch captured in the build scan is wrong
It's captured as [`HEAD`](https://ge.spring.io/s/b4ucqvh6zyjis/custom-values#L1). I suspect it may be a side-effect of how the Concourse git resource checks out the code.

dreis2211:
Hi,

this is an attempt to fix #19614 by passing the Concourse branch parameter to the build scan directly and only fallback to the current mechanism if it isn't set (e.g. in local builds).

As the Concourse stuff is hard to test for me, feel free to decline the PR. Let me know what you think.

Cheers,
Christoph

### DIFF
--- a/gradle/build-scan-user-data.gradle
+++ b/gradle/build-scan-user-data.gradle
@@ -27,7 +27,7 @@ void tagCiOrLocal() {
 void addGitMetadata() {
 	gradleEnterprise.buildScan.background {
 		def gitCommitId = execAndGetStdout('git', 'rev-parse', '--short=8', '--verify', 'HEAD')
-		def gitBranchName = execAndGetStdout('git', 'rev-parse', '--abbrev-ref', 'HEAD')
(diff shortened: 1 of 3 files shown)


gradle/build-scan-user-data.gradle
```diff
@@ -50,6 +50,14 @@ boolean isCi() {
 	System.getenv('CI')
 }
 
+String getBranch() {
+	def branch = System.getenv('BRANCH')
+	if (branch != null && !branch.isEmpty()) {
```
wilkinsona: Note for merging: I think this can be `if (branch)`.

snicoll:
I've applied those changes to CI. Merging and we can revisit if that breaks

dreis2211:
image

PR #19669, opened 2020-01-13
Add a way to customize Hazelcast's Client/Server configs

nosan:
This PR provides a way to customize Hazelcast's Client/Server configs before `HazelcastInstance` is created.  At the moment there is no way to do this except creating your own config.

Here is a small example of a configuration, that shows how to set `SpringManagedContext` using the `HazelcastConfigCustomizer`

```java

static class HazelcastConfiguration {

	@Bean
	HazelcastConfigCustomizer hazelcastConfigCustomizer(ApplicationContext applicationContext) {
		return (config) -> {
			SpringManagedContext managedContext = new SpringManagedContext();
			managedContext.setApplicationContext(applicationContext);
			config.setManagedContext(managedContext);
		};
	}

}

```

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastInstanceFactory.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastInstanceFactory.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,7 +21,9 @@ import java.net.URL;
 
 import com.hazelcast.config.Config;
 import com.hazelcast.config.XmlConfigBuilder;
+import com.hazelcast.config.XmlConfigLocator;
 import com.hazelcast.config.YamlConfigBuilder;
+import com.hazelcast.config.YamlConfigLocator;
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
 
@@ -42,13 +44,38 @@ public class HazelcastInstanceFactory {
 	private final Config config;
 
 	/**
+	 * Create a new {@link HazelcastInstanceFactory} instance.
+	 * @param configCustomizers any {@link HazelcastConfigCustomizer customizers} that
+	 * should be applied when the {@link Config} is built and before
+	 * {@link HazelcastInstance} is created.
+	 * @since 2.3.0
+	 */
+	public HazelcastInstanceFactory(HazelcastConfigCustomizer... configCustomizers) {
+		Assert.notNull(configCustomizers, "ConfigCustomizers must not be null");
+		Config config = getDefaultConfig();
+		for (HazelcastConfigCustomizer configCustomizer : configCustomizers) {
+			configCustomizer.customize(config);
+		}
+		this.config = config;
+	}
+
+	/**
 	 * Create a {@link HazelcastInstanceFactory} for the specified configuration location.
 	 * @param configLocation the location of the configuration file
+	 * @param configCustomizers any {@link HazelcastConfigCustomizer customizers} that
+	 * should be applied when the {@link Config} is built and before
+	 * {@link HazelcastInstance} is created.
 	 * @throws IOException if the configuration location could not be read
 	 */
-	public HazelcastInstanceFactory(Resource configLocation) throws IOException {
+	public HazelcastInstanceFactory(Resource configLocation, HazelcastConfigCustomizer... configCustomizers)
+			throws IOException {
 		Assert.notNull(configLocation, "ConfigLocation must not be null");
-		this.config = getConfig(configLocation);
+		Assert.notNull(configCustomizers, "ConfigCustomizers must not be null");
+		Config config = getConfig(configLocation);
+		for (HazelcastConfigCustomizer configCustomizer : configCustomizers) {
+			configCustomizer.customize(config);
+		}
+		this.config = config;
 	}
 
 	/**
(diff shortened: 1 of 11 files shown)


nosan:
Related issues
- https://github.com/hazelcast/hazelcast/pull/16386

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastClientConfigFactory.java
```diff
@@ -0,0 +1,118 @@
...
+import org.springframework.core.io.Resource;
+import org.springframework.util.Assert;
+
+/**
+ * Factory that can be used to create the Hazelcast {@link ClientConfig}.
+ *
+ * @author Dmytro Nosan
+ * @since 2.3.0
+ */
+public final class HazelcastClientConfigFactory {
```
snicoll: This partially duplicates what  `HazelcastClientFactory` is doing. If that behaviour can/should be improved I don't think we should do it as part of this particular issue.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastConfigFactory.java
```diff
@@ -0,0 +1,130 @@
...
+import org.springframework.util.Assert;
+import org.springframework.util.ResourceUtils;
+
+/**
+ * Factory that can be used to create the Hazelcast {@link Config}.
+ *
+ * @author Dmytro Nosan
+ * @since 2.3.0
+ */
+public final class HazelcastConfigFactory {
```
snicoll: Ditto. I also question the creation of such high-level concept for this. IMO this feels like Hazelcast itself should provide that.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastInstanceFactory.java
```diff
@@ -45,10 +41,11 @@
 	 * Create a {@link HazelcastInstanceFactory} for the specified configuration location.
 	 * @param configLocation the location of the configuration file
 	 * @throws IOException if the configuration location could not be read
+	 * @deprecated since 2.3.0 in favor of {@link HazelcastConfigFactory}
 	 */
+	@Deprecated
```
snicoll: Isn't that making `HazelcastInstanceFactory` effectively deprecated? For the reason above, I don't think we should do that and rather research how to improve the existing class to account for the customizers.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
```diff
@@ -5666,9 +5666,6 @@ If https://hazelcast.com/[Hazelcast] is on the classpath and a suitable configur
 If you define a `com.hazelcast.config.Config` bean, Spring Boot uses that.
 If your configuration defines an instance name, Spring Boot tries to locate an existing instance rather than creating a new one.
 
-If you define a `com.hazelcast.config.Config` bean, Spring Boot uses that.
-If your configuration defines an instance name, Spring Boot tries to locate an existing instance rather than creating a new one.
-
```
snicoll: This shouldn't be removed. If a user specifies an existing `Config` we should use that rather than trying to lookup one.

spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/hazelcast/HazelcastClientConfigCustomizerConfiguration.java
```diff
@@ -0,0 +1,45 @@
...
+// tag::configuration[]
+@Configuration(proxyBeanMethods = false)
+public class HazelcastClientConfigCustomizerConfiguration {
+
+	@Bean
+	public HazelcastClientConfigCustomizer hazelcastClientConfigCustomizer(ApplicationContext applicationContext) {
+		return (clientConfig) -> {
+			SpringManagedContext managedContext = new SpringManagedContext();
+			managedContext.setApplicationContext(applicationContext);
+			clientConfig.setManagedContext(managedContext);
```
snicoll: Does it make sense to enable the management context in a Hazelcast client?

nosan:
Thank you for your feedback @snicoll 

>  If you have some more concrete use cases that the management context, please let us know.

Unfortunately, I don't have any use cases. I've seen some snippets on _StackOverflow_, e.g. how to set a `ClassLoader`.
I definitely can not say whether this will be useful for folks or not. I'm considering this feature as _nice-to-have_.

Of course, I understand that you will have to maintain this code and if you feel it is not worthy then feel free to close this PR. 



**PR** has been updated. I have removed additional factories that were used for configs creation.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
nosan: These duplicates.

spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/hazelcast/HazelcastClientConfigCustomizerConfiguration.java
nosan: Generally speaking, I don't know.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
snicoll: Ah sorry, I missed that. Such polish must be done in a separate commit. I've addressed it in #19678

spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/hazelcast/HazelcastClientConfigCustomizerConfiguration.java
snicoll: @jerrinot can you please help here and let us know if it does?

snicoll:
Thanks @nosan. I feel that adding two new top-level interface is not warranted here. If it turns out the callback is required to implement the configuration of the `SpringManagementContext` we can reconsider.

nosan:
@snicoll np, thanks

PR #19677, opened 2020-01-13
Made setDefaultProperties(Map) duplicate the map passed in, rather th…

AaronTraas:
I recently had an issue in a project that was calling `app.setDefaultProperties(Map)` and passing in an immutable map. Worked great, until we integrated with a cloud config server, in which case, it was trying to `put()` in an immutable map, which didn't go so well. 

I copied the code from `app.setDefaultProperties(Properties)`, which created an internal HashMap, and populated it with the contents. This seems like the safer approach, rather than passing in the data structure that is to be used for the inner workings of the properties system.

With the old approach, I could pass in ANY implementation of `Map`, including ones that make get/put non-functional, and Spring will try to use them. The new approach fixes that by normalizing the internal data structure used for properties, and hiding that from the outside.


pivotal-issuemaster:
@AaronTraas Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19677)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19677) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@AaronTraas Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19677)!

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
```diff
@@ -1023,7 +1023,10 @@ public void setAddConversionService(boolean addConversionService) {
 	 * @param defaultProperties the additional properties to set
 	 */
 	public void setDefaultProperties(Map<String, Object> defaultProperties) {
-		this.defaultProperties = defaultProperties;
+		this.defaultProperties = new HashMap<>();
```
wilkinsona: I think this could be a one liner:

```java
this.defaultProperties = new HashMap<>(defaultProperties);
```

AaronTraas: Yup -- you're right. Was being too lazy copying from setDefaultProperties(Properties)

AaronTraas:
I can write tests. Just to clarify the scope, I'm thinking two tests:

1. Sets the properties with an immutable map, tries to set a property, and setting the property works
2. Sets a property with a new hashmap, then try to set the property of the original hashmap, and compare it with the property in the system, and assert that they should not be the same

Do you think that would cover the regression? I can hammer that out this evening if so.

wilkinsona:
I was thinking of a test that covers what Spring Cloud Config server is doing. It would fail without your changes and pass with them. The test would need to get the `defaultProperties` property source from the `Environment`, call `getSource()` and then try to put a property into the source.

AaronTraas:
Gotcha. Thanks for the clarification. 

I was able to make two tests, both of which would fail against the current code, but do not fail with my changes.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/SpringApplicationTests.java
```diff
@@ -36,6 +36,7 @@
 import org.mockito.ArgumentMatchers;
 import org.mockito.InOrder;
 import org.mockito.Mockito;
+import org.springframework.core.env.*;
```
wilkinsona: We prefer not to use wildcard imports. Could you please remove this and reinstate the individual imports below?

AaronTraas: Oh -- sorry. IntelliJ just did that automatically. I didn't notice. Stupid IDEs. Fixed now.

snicoll: @AaronTraas you can configure IJ not to do that in Preferences (Editor > Code Style > Java > Imports and increase the class/name count to something ridiculously high :) )

AaronTraas: @snicoll Yeah, I discovered that before committing. Set it to 9999. I think that should have me covered :)

AaronTraas:
One last question -- did I make this in the right branch? This really should be for everything going forward, and could also be back-ported to 2.1, etc.

wilkinsona:
No need to worry about the branch. We can cherry-pick the change back if needed and then merge it forwards.

wilkinsona:
We discussed one and I’m afraid we’ve decided not to merge it. In addition to the possibly breaking change I described above, the team also pointed out that users may want the current behaviour where an immutable map remains immutable. We can’t fix your specific problem without changing that behaviour for everyone. If you need the default properties to be mutable, we think you should pass in a mutable map. Thank you anyway for the proposal.

You may want to consider raising this with the Spring Cloud Config project. They may be able to do something to cope more elegantly with a situation where the default properties are immutable.

Issue #18777, opened 2019-10-29
Raise the minimum supported version of Gradle to 5.6


wilkinsona:
In Spring Boot 2.2 we retained support for Gradle 4 (specifically 4.10.x) in a deprecated form and documented that it would be dropped in a future release. I think we should do that in 2.3. At the same time we should raise the minimum support version of 5.x to something relatively modern to make it easier to support 5.x and 6.x in parallel.

wilkinsona:
I think we should be quite aggressive here in M1 and see how it goes. I'd like to try only supporting 5.6.x and 6.x in Spring Boot 2.3.

dreis2211:
This breaks the jdk-13-build. See #19680 .

PR #19680, opened 2020-01-13
Disable failing Gradle tests on Java 13

dreis2211:
Hi,

apparently raising the new minimum version for Gradle via #18777 breaks the build on JDK 13. I actually don't have a clue why this worked with 4.10 previously, but as 5.6 isn't supporting JDK 13 we can maybe disable the tests under Java 13?

Maybe you have more insights on why this worked previously, in which case feel free to decline the PR.

Cheers,
Christoph


wilkinsona:
Thanks, @dreis2211. I can't explain why that worked with 4.10. I can only guess that the usage of ASM that's causing the failure now is something that's new in 5.x.

PR #19684, opened 2020-01-13
Improve messages in case of DataSource health check failure.

rleroy:
My application is running on several data sources.
I have a monitoring that raise alerts based on the logs, but do not allow to view the related stack trace easily.
One of the database was failing from time to time, and I had to dig in my logs to find which one.
Adding those logs will allow to find out which data source have a problem and get the error message without having to dig into the stack trace.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.boot.actuate.autoconfigure.jdbc;
 
 import java.util.Collection;
 import java.util.Map;
+import java.util.Map.Entry;
 import java.util.stream.Collectors;
 
 import javax.sql.DataSource;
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the proposal. If this is something that we want to support, I wonder if it might be better to make some changes to the super-class so that the name is passed into the method for creating the indicator rather than having to find it again in the map. That might be quite hard to do in a backwards-compatible manner. Flagging for team attention to see if we think it's worth it.

snicoll:
IMO, if we decide to implement something like that, we should make sure the format of the log entry is "parseable" and maintained consistent unless we have to break it the same way we do for APIs. I am not aware of other places in Spring Boot where our log entries are meant to be parsed by monitoring tools.

rleroy:
Passing the name of the data source all along the chain would indeed avoid to reverse search it in the map again. But I wanted to keep the change as little as possible. Feel free to make it bigger.

As for the message of the exception, making AbstractHealthIndicator.healthCheckFailedMessage not final and set it inside DataSourceHealthIndicator constructor could also be an option.

And of course, feel free to improve the message itself, as long as the name of the data source and the original exception message are kept this is fine for me.

wilkinsona:
Thanks for the proposal. We discussed this today and concluded that we'd like to be able to offer some more information when the health check fails, but we'd like to take a step back and consider some different approaches for how that might be achieved. I've opened https://github.com/spring-projects/spring-boot/issues/19983 to track that effort. Thanks again for the proposal and for initiating the discussion. Please comment on #19983 with any ideas and suggestions that you may have.

PR #19711, opened 2020-01-14
Fix deprecations in spring-boot-smoke-test-hateoas

dreis2211:
Hi,

this PR fixes two deprecations in `spring-boot-smoke-test-hateoas` that were introduced by building against Spring HATEOAS 1.1.0 snapshots.

Cheers,
Christoph


dreis2211:
Test failure seems unrelated

Issue #19702, opened 2020-01-14
Restore TomcatReactiveWebServerFactoryTests


philwebb:
A recent Netty upgrade brings [a change](https://github.com/netty/netty/commit/8494b046ec7e4f28dbd44bc699cc4c4c92251729) has highlighted a [potential Tomcat bug](https://bz.apache.org/bugzilla/show_bug.cgi?id=64073).

`TomcatReactiveWebServerFactoryTests` now fails with the following:

```
java.lang.IllegalArgumentException: Both 'Content-Length: 3000' and 'Transfer-Encoding: chunked' found
	at io.netty.handler.codec.http.HttpObjectDecoder.readHeaders(HttpObjectDecoder.java:648) ~[netty-codec-http-4.1.44.Final.jar:4.1.44.Final]
	at io.netty.handler.codec.http.HttpObjectDecoder.decode(HttpObjectDecoder.java:218) ~[netty-codec-http-4.1.44.Final.jar:4.1.44.Final]
	at io.netty.handler.codec.http.HttpClientCodec$Decoder.decode(HttpClientCodec.java:202) ~[netty-codec-http-4.1.44.Final.jar:4.1.44.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.decodeRemovalReentryProtection(ByteToMessageDecoder.java:493) ~[netty-codec-4.1.44.Final.jar:4.1.44.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.callDecode(ByteToMessageDecoder.java:432) ~[netty-codec-4.1.44.Final.jar:4.1.44.Final]
	at io.netty.handler.codec.ByteToMessageDecoder.channelRead(ByteToMessageDecoder.java:271) ~[netty-codec-4.1.44.Final.jar:4.1.44.Final]
	at io.netty.channel.CombinedChannelDuplexHandler.channelRead(CombinedChannelDuplexHandler.java:251) [netty-transport-4.1.44.Final.jar:4.1.44.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:377) [netty-transport-4.1.44.Final.jar:4.1.44.Final]
	at io.netty.channel.AbstractChannelHandlerContext.invokeChannelRead(AbstractChannelHandlerContext.java:363) [netty-transport-4.1.44.Final.jar:4.1.44.Final]
[...]

PR #19718, opened 2020-01-15
Restore disabled tests in TomcatReactiveWebServerFactoryTests

dreis2211:
Hi,

given that the publishing of Gradle module metadata was disabled again for now and we're back on using enforced platforms #19702 I think we can restore the tests in `TomcatReactiveWebServerFactoryTests`. What do you think?

Cheers,
Christoph


wilkinsona:
Thanks for the suggestion, @dreis2211. We’ll be upgrading to Netty 4.1.44 (or even 4.1.45) prior to the next batch of releases so if we merged this we’d only have to revert it again.

dreis2211:
Fair enough

PR #19770, opened 2020-01-16
Test the Gradle Plugin against Gradle 6.1

dreis2211:
Hi,

Gradle 6.1 was released earlier this night, so I guess we can test against it.

Cheers,
Christoph

PR #19776, opened 2020-01-16
Test the Gradle Plugin against Gradle 6.1

dreis2211:
Hi,

second try, but based on 2.2.x. Sorry for the inconveniences.

Cheers,
Christoph


snicoll:
Note when merging, use `-s ours` strategy, see https://github.com/spring-projects/spring-boot/pull/19770#issuecomment-575064128

PR #19790, opened 2020-01-17
Upgrade to Infinispan 10.1.1.Final

scottfrederick:
This includes removing modules that were made obsolete (notably `infinispan-spring4-*`) and adding modules that are new in 10.1.0.


snicoll:
Superseded by #20163

PR #19797, opened 2020-01-17
Revisit documentation that mentions Gradle "compile" configuration

### LINKED ISSUE #19794: Revise documentation that mentions Gradle "compile" configuration
Hi,

afaik `compile` has been deprecated in favor of using `api` or `implementation`. In some places in the documentation the old approach is still mentioned: 
```
dependencies {
	compile("org.springframework.boot:spring-boot-starter-actuator")
}
```

I specifically didn't open a PR directly, because I wanted to discuss first if we should stick to `api` or `implementation` here. `api` would be closer to the old approach, but `implementation` is generally preferable. What do you think?

Cheers,
Christoph

dreis2211:
Hi,

this PR fixes #19794 as discussed + a little polishing for `exclude`s that were in the same example scope.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator/README.adoc
+++ b/spring-boot-project/spring-boot-actuator/README.adoc
@@ -27,7 +27,7 @@ For Gradle, use the following declaration:
 [indent=0]
 ----
 	dependencies {
-		compile("org.springframework.boot:spring-boot-starter-actuator")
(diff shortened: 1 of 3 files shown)


dreis2211:
Test failure seems unrelated

wilkinsona:
Unrelated and annoying. I've opened https://github.com/spring-projects/spring-boot/issues/19801.

PR #19804, opened 2020-01-18
Protect against available port actually being unavailable

### LINKED ISSUE #19801: NettyRSocketServerFactoryTests fails intermittently due to findAvailableTcpPort returning an unavailable port

dreis2211:
Hi,

this is an attempt to fix #19801. I remembered an issue in the past that tackled things in the same way (see #19355 ) that I applied to `NettyRSocketServerFactoryTests`. Hopefully that helps stabilizing the test-suite.

Cheers,
Christoph


dreis2211:
Seems like @wilkinsona had the same idea. Closing 😉

PR #19825, opened 2020-01-20
Fix Linkage error when using the build cache

wolfs:
The error should be fixed now by adding explicit taskdefs
instead of changing the classpath of the classloader which
did load org.apache.tools.ant.Project.
This classloader is owned by Gradle and is probably a parent
of the one which is used for `TarBuildCacheEntryPacker`.

See https://github.com/gradle/gradle/issues/11914

### DIFF
--- a/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-ant/build.gradle
+++ b/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-ant/build.gradle
@@ -53,8 +53,16 @@ ant.properties['projectDir'] = project.layout.projectDirectory
 antresolve {
 	dependsOn syncTestRepository, configurations.antDependencies
 	doFirst {
-		ClassLoader antClassLoader = org.apache.tools.ant.Project.class.classLoader
-		configurations.antDependencies.each { antClassLoader.addURL it.toURI().toURL() }
+		ant.taskdef(
+				resource: "org/apache/ivy/ant/antlib.xml",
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@wolfs Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19825)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19825) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@wolfs Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19825)!

wilkinsona:
Thank you, @wolfs. Unfortunately, we really need a way of adding to Ant's classpath without any additional taskdefs. A user's of Spring Boot's Ant integration shouldn't have to define any tasks and the smoke test is intended to verify that is the case. Defining tasks up front means that the smoke test is no longer covering everything that it is intended to cover. We really need an equivalent of Maven's AntRun plugin where dependencies can be added to the classpath and Ant picks them up as it would do when being invoked "normally".

It may be that what's proposed here is better than the occasional loader constraint violations and it's worth merging it, but something that does nothing more than adding to Ant's classpath would be preferable. I'd like to be certain that such an option does not exist first.

wilkinsona:
Superseded by https://github.com/spring-projects/spring-boot/issues/19839.

PR #19826, opened 2020-01-20
Reinstate running of OutputCaptureRuleTests

dreis2211:
Hi,

I just noticed that `OutputCaptureRuleTests` were not running due to the fact that they need to be run with JUnit4. This PR is an attempt to fix this.

Let me know if there's maybe a smarter way to achieve this in Gradle.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-test/build.gradle
+++ b/spring-boot-project/spring-boot-test/build.gradle
@@ -52,4 +52,10 @@ dependencies {
 	testImplementation 'org.testng:testng'
 
 	testRuntimeOnly 'org.junit.platform:junit-platform-launcher'
+	testRuntimeOnly 'org.junit.vintage:junit-vintage-engine'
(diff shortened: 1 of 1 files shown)


dreis2211:
Test failure seems to be our infamous loader constraint violation.

sbrannen:
I realize this PR has been closed, but unless I'm missing something you should be able to remove `useJUnit()` since `useJUnitPlatform()` will result in the Vintage test engine running on the JUnit Platform.

TBH, I'm not even sure what happens behind the scenes if you declare both `useJUnit()` and `useJUnitPlatform()` in the same `test` task -- maybe JUnit 4 tests are executed twice.

dreis2211:
@sbrannen Good catch. It's not needed, but it doesn't seem to execute it twice either. Anyway, I've opened #20035

PR #19863, opened 2020-01-23
Fix TestFailuresPlugin.TestFailure.compareTo()

izeye:
This PR fixes `TestFailuresPlugin.TestFailure.compareTo()` as it seems that `getName()` was meant to be used.

This PR also polishes around it a bit.


wilkinsona:
Thanks very much, @izeye.

PR #19864, opened 2020-01-23
Use Supplier variants of Assert

dreis2211:
Hi,

this PR changes some `Assert` calls to make use of their `Supplier` variants in order to save the overhead of String construction in case the assertion passes.

Cheers,
Christoph


dreis2211:
Couldn't create volume on CI. Do we have any indication of what this might be? It's happening relatively often these days 😞

snicoll:
Sorry about that @dreis2211. We're investigating that problem for quite some time now. There's been some progress lately.

PR #19898, opened 2020-01-24
Introduce appendix section with version properties

dreis2211:
Hi,

with the introduction of Gradle we "lost" the list of version properties that were previously in the spring-boot-dependencies POM and were also linked inside the documentation (which are currently leading to 404 errors because the pom.xml does not exist anymore obviously). This PR introduces an appendix section in the docs and links the appropriate places to the new section to restore discoverability.

Let me know what you think.
Cheers,
Christoph


wilkinsona:
Love it. Thanks for catching the problem and then also implementing a solution. Much appreciated.

snicoll:
Thanks a lot @dreis2211!

PR #19899, opened 2020-01-24
Dirty changes when using CqlSessionBuilder to create several sessions

nosan:
`CqlSessionBuilder` is a mutable builder, hence this may lead to unexpected situations.
 Here is a configuration that shows a problem:
```java
class CassandraConfiguration {

	@Bean
	@DependsOn("session2")
	public CqlSession session1(CqlSessionBuilder builder) {
		return builder.build();
	}

	@Bean
	public CqlSession session2(CqlSessionBuilder builder) {
		return builder.withKeyspace("test").build();
	}

}

```

This PR adds `@Scope(prototype)` to avoid such problems 😃

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraAutoConfiguration.java
@@ -44,6 +44,7 @@ import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.context.annotation.Scope;
 
 /**
  * {@link EnableAutoConfiguration Auto-configuration} for Cassandra.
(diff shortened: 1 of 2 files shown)


snicoll:
@nosan thanks for the PR and testing M1 so quickly. Can you share why you'd want to create two `CqlSession` beans? Last time I checked with @mp911de that didn't seem to be something users would do.

nosan:
Hi @snicoll,
The simple example is: 
I'd like to have multiple `CqlSession` that connected to different keyspaces.

mp911de:
Shouldn't that be rather handled in your bean definitions?

e.g.

```java
class CassandraConfiguration {

	@Bean
	@DependsOn("session2")
	public CqlSession session1(CqlSessionBuilder builder) {
		return builder.withKeyspace("primaryKeySpace").build();
	}

	@Bean
	public CqlSession session2(CqlSessionBuilder builder) {
		return builder.withKeyspace("secondaryKeySpace").build();
	}

}
```

nosan:
Thank you for your feedback @mp911de,

Of course, it could be handled on my side, but your example works fine for one property.

From my perspective, it would be nice to have an independent builder in my configuration, so I can do whatever I want with it during `CqlSession` creation, otherwise, I have to remember what I have changed in builder before.

snicoll:
That makes sense, thanks for the feedback @nosan

Issue #19846, opened 2020-01-21
Spring Webflux ignores message of custom exception when annotated with @ResponseStatus


xabgesagtx:
**Affected component:** spring-webflux
**Version:** 2.2.4.RELEASE

When returning a custom exception with a `@ResponseStatus` annotation in a controller method, spring mvc extracts the message of the exception into the error attribute `message`. This message will then be available in the default json response and can also be used in templates.

Spring webflux however ignores the message of the exception when it's annotated with `@ResponseStatus`. The only way to set a message then is via the `reason` parameter of the annotation.

I created a [sample application](https://github.com/xabgesagtx/spring-boot-error-message-example) with a (currently failing) WebFluxTest to demonstrate the behaviour of ignoring the message.

After some investigation in the code, I think that in [DefaultErrorAttributes](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/error/DefaultErrorAttributes.java), when the message is determined, it's not checked if the value for the `reason` parameter is an empty string, which is the default value for this parameter. Therefore this empty string is always preferred to the message of the message of the exception.

snicoll:
@xabgesagtx thanks for the report and the sample!

wilkinsona:
Closing in favour of #19901.

PR #19901, opened 2020-01-24
Spring Webflux ignores message of custom exception when annotated with @ResponseStatus

nosan:
see gh-19846


snicoll:
Thanks again @nosan

PR #19907, opened 2020-01-24
Upgrade to JUnit Jupiter 5.6.0

dreis2211:
Hi,

this PR upgrades JUnit Jupiter to version 5.6.0. There was a field renamed in the internals that broke the reflection calls in `ModifiedClassPathExtension`, but luckily there as a new `Invocation.skip()` that makes the reflection obsolete.

I also used the new `@DisabledForJreRange` annotation already in order to disable some tests from JDK 13 onwards. That should make an eventual upgrade to JDK 14 smoother.

Let me know what you think.
Cheers,
Christoph


dreis2211:
Seems like the new GitHub actions integration fails.

snicoll:
Yep @dreis2211, we're looking into it, considering rollbacking that one if we can't manage to get that fixed soon. Thanks for the PR!

PR #19936, opened 2020-01-26
Fix broken documentation links

dreis2211:
Hi,

this PR fixes a variety of issues related to documentation links on the current master:

- 404 errors (e.g. for the reworked Maven/Gradle plugin docs)
- Some wrong fragments (e.g. pointing to the "Dependency versions" appendix)
- Permanent redirects

Cheers,
Christoph


snicoll:
Thanks again @dreis2211. This affects mostly master but I've noticed a couple of things that might be backported. I'll double check before merging.

PR #19952, opened 2020-01-27
Fix links in Maven Plugin documentation

dreis2211:
Hi,

@snicoll's work on the Maven Plugin documentation sparked a thought to check for broken links inside the documentation like I did it for the normal Spring-Boot docs in #19936 and I indeed found some links that do not work. 

Cheers,
Christoph


snicoll:
Thanks again!

Issue #19785, opened 2020-01-17
Asciidoctor warnings due to javadoc conversion


philwebb:
```
> Task :spring-boot-project:spring-boot-docs:asciidoctorPdf
Failed to parse formatted text: Whether to disable the escaping of HTML characters such as '<', '>', etc.
Failed to parse formatted text: Whether to disable the escaping of HTML characters such as '<', '>', etc.
Failed to parse formatted text: JNDI location of the datasource. Class, url, username & password are ignored when set.
Failed to parse formatted text: JNDI location of the datasource. Class, url, username & password are ignored when set.
Failed to parse formatted text: Comma-separated list of additional unencoded characters that should be allowed in URI paths. Only "< > [ \ ] ^ ` { | }" are allowed.
Failed to parse formatted text: Comma-separated list of additional unencoded characters that should be allowed in URI query strings. Only "< > [ \ ] ^ ` { | }" are allowed.
Failed to parse formatted text: Comma-separated list of additional unencoded characters that should be allowed in URI paths. Only "< > [ \ ] ^ ` { | }" are allowed.
Failed to parse formatted text: Comma-separated list of additional unencoded characters that should be allowed in URI query strings. Only "< > [ \ ] ^ ` { | }" are allowed.
```

I think this is due to our use if `<` and `>` in some of the configuration properties javadoc. We might want to escape those then update the converter code to deal with them. We'll also need to update the annotation processor that generates JSON.

snicoll:
>  We'll also need to update the annotation processor that generates JSON.

I don't think these needs escaping as it's part of the description that's already a `String`.

philwebb:
What I meant was we have comments like this in `ServerProperties`:

```java
/**
 * Comma-separated list of additional unencoded characters that should be allowed
 * in URI paths. Only "< > [ \ ] ^ ` { | }" are allowed.
 */
```

Those cause javadoc warnings because `<` should be written as `&lt;`. We've generally tried to avoid javadoc specific markup in the configuration properties but I think these are different.

If we replaced those chars, I don't know what the impact would be on the JSON generation.

wilkinsona:
[...]

PR #19953, opened 2020-01-27
Upgrade to Asciidoctor Gradle JVM 3.0.0

dreis2211:
Hi,

this PR upgrades to `org.asciidoctor:asciidoctor-gradle-jvm:3.0.0` and fixes related issues that prevented building the docs on that new version. This seems to overlap with #19785 so I've split the necessary commits.

Let me know what you think.
Cheers,
Christoph


dreis2211:
The current errors should be fixed with #19952 already

wilkinsona:
We should check if we need to tackle https://github.com/spring-io/spring-asciidoctor-extensions/issues/21 before merging this. That issue was opened as a placeholder so it may be that the extension is already AsciidoctorJ 2.0 compatible.

dreis2211:
Is there anything I can help with, @wilkinsona ?

wilkinsona:
Thanks for the offer, @dreis2211. I think it's been taken care of thanks to @rwinch. I just need to double-check that I didn't break it when merging and then cut a release of the extensions.

wilkinsona:
The extensions have been released and Boot's now using them. I think we should now be good-to-go on this one.

wilkinsona:
The PDF generation [failed](https://ge.spring.io/s/vx4qqz2lkhajs/console-log?task=:spring-boot-project:spring-boot-docs:asciidoctorPdf) when built locally. I think that's because we're using the configuration property JSON to generate the Asciidoc source. This makes the changes to use, for example, `&lt;` rather than `<` have no effect. The `&lt;` has been replaced with `<` again by the time the Asciidoctor processing is performed. I think we may need to do some escaping of the description in `SingleConfigurationTableEntry` instead.

dreis2211:
Interesting....I wonder why this didn't fail for me locally?

dreis2211:
Pushed an updated version

PR #19955, opened 2020-01-28
StackdriverMetricsExportAutoConfiguration is not registered in `META-INF/spring.factories`

wonwoo:


snicoll:
Oh dear 🤦‍♂️

Good catch @wonwoo

PR #19956, opened 2020-01-28
Fix typo

natrem:
Fix non-existing documented type ``java.util.String``


snicoll:
@natrem good catch and thank you for making your first contribution to Spring Boot.

PR #19959, opened 2020-01-28
Add shared jar launcher to core launcher implementations

alero:
We use this in production as it helps with different service runners (mainly to support windows proc-run that send stop/retart commands via the same open jvm and its running classloader)

The shared jar launcher means its possible to send commands to the same open service loaded via spring-boot

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/SharedJarLauncher.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/SharedJarLauncher.java
@@ -1,0 +1,63 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@alero Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19959)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=19959) to manually synchronize the status of this Pull Request.

alero:
Will wait for feedback before I progress with CLA and/or tests, not sure if this is seen as a feature that would be needed by the spring-boot community

wilkinsona:
Thanks for the PR, @alero. Can you give us an example of how you typically use the shared launcher? What's the entry point to the JVM where the shared instance of the launcher exists?

alero:
Hi, yes we utilize proc-run to run the spring-boot services within windows servers, this binds the service to the windows service management and makes it possible to stop/start/restart. But this does not utilize the more common way of sending kill signals to the JVM but rather calls the same method "init" again but with a kill/stop word, to be able to actually shutdown the running JVM we must be able to reach the classloader that runs the spring-boot application

https://commons.apache.org/proper/commons-daemon/procrun.html

wilkinsona:
Thanks for the info, @alero. I've flagged this one for team attention so that we can discuss whether or not this is something that we want to support out of the box.

wilkinsona:
Thanks for the suggestion. We've discussed this today and the team have decided that we do not want to add shared launcher support to Spring Boot at this time. We haven't seen sufficient demand for the functionality to justify the ongoing maintenance cost. Thanks anyway.

Issue #19356, opened 2019-12-12
Upgrade to SpringData MongoDB 3.0 and MongoDB Java Driver 4.0


christophstrobl:
SpringData MongoDB intends to upgrade to the MongoDB Java Driver 4.0 (sync/reactive) with the Neumann (M1) release. This will be a major upgrade due to breaking changes in the public API of SpringData (see [upgrading from 2.x to 3.x](https://github.com/spring-projects/spring-data-mongodb/blob/3.x/src/main/asciidoc/upgrading.adoc) for details).

Both, the MongoDB 4.0 drivers as well as SpringData MongoDB 3.0 are currently only available as SNAPSHOT dependencies. 

---

SpringData MongoDB - 3.0
https://github.com/spring-projects/spring-data-mongodb/tree/3.x

Spring Boot - Mongo Drivers 4.0 (incl. Spring Data MongoDB 3.0)
https://github.com/christophstrobl/spring-boot/tree/mongo-driver-4_0

Spring Session Data MongoDB
https://github.com/spring-projects/spring-session-data-mongodb/tree/mongo-driver-4_0

wilkinsona:
Thanks, @christophstrobl. This will require some coordination across a few projects to get this into Boot 2.3 (due in April 2020) as it will pull a few projects into that schedule.

I think at least the following projects will be affected:

- Spring Batch (@benas @mminella)
- Spring Integration (@artembilan @garyrussell)
- Spring Session (@vpavic @rwinch)

There may be others that are affected directly or indirectly. Can you and @mp911de please coordinate with the other projects and make sure that everyone has the bandwidth to absorb the breaking changes coming in Neumann?

christophstrobl:
@vpavic @gregturn there's already a branch with the required changes ([spring-session-data-mongodb/mongo-driver-4_0](https://github.com/spring-projects/spring-session-data-mongodb/tree/mongo-driver-4_0)) for Spring Session MongoDB.

artembilan:
See https://github.com/spring-projects/spring-integration/issues/3121.

This is going to be one more argument to release Spring Integration 5.3 in time for Spring Boot 2.3.

Thanks for heads up!

christophstrobl:
SD-MongoDB 4.0 (beta1) driver changes are on master now. I'll update https://github.com/christophstrobl/spring-boot/tree/mongo-driver-4_0 and open a PR.

snicoll:
Closing in favour of PR #19960 

[...]

PR #19960, opened 2020-01-28
Upgrade to SpringData MongoDB 3.0 and MongoDB Java Driver 4.0 beta1

christophstrobl:
Upgrade to Spring Data MongoDB 3.0 (Neumann). The update requires MongoDB Driver 4.0 (beta1 at this time). 

Switch from `com.mongodb.MongoClient` and `com.mongodb.MongoClientOptions` to `com.mongodb.client.MongoClient` & `com.mongodb.MongoClientSettings` and share code paths for `MongoClientSettings` between reactive and imperative configuration.

Closes: #19356

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoAutoConfiguration.java
@@ -16,8 +16,10 @@
 
 package org.springframework.boot.autoconfigure.mongo;
 
-import com.mongodb.MongoClient;
-import com.mongodb.MongoClientOptions;
+import java.util.stream.Collectors;
+
+import com.mongodb.MongoClientSettings;
+import com.mongodb.client.MongoClient;
 
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
@@ -45,10 +47,13 @@ import org.springframework.core.env.Environment;
 public class MongoAutoConfiguration {
 
 	@Bean
-	@ConditionalOnMissingBean(type = { "com.mongodb.MongoClient", "com.mongodb.client.MongoClient" })
-	public MongoClient mongo(MongoProperties properties, ObjectProvider<MongoClientOptions> options,
-			Environment environment) {
-		return new MongoClientFactory(properties, environment).createMongoClient(options.getIfAvailable());
+	@ConditionalOnMissingBean(type = { "com.mongodb.client.MongoClient" })
+	public MongoClient mongo(MongoProperties properties, Environment environment,
+			ObjectProvider<MongoClientSettingsBuilderCustomizer> builderCustomizers,
+			ObjectProvider<MongoClientSettings> settings) {
+		return new MongoClientFactory(properties, environment,
+				builderCustomizers.orderedStream().collect(Collectors.toList()))
+						.createMongoClient(settings.getIfAvailable());
 	}
 
 }
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoClientFactory.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoClientFactory.java
@@ -16,15 +16,11 @@
 
 package org.springframework.boot.autoconfigure.mongo;
 
-import java.util.Collections;
 import java.util.List;
 
-import com.mongodb.MongoClient;
-import com.mongodb.MongoClientOptions;
-import com.mongodb.MongoClientOptions.Builder;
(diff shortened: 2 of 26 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoClientFactorySupport.java
```diff
@@ -0,0 +1,144 @@
...
+
+	private void applyHostAndPort(MongoClientSettings.Builder settings) {
+
+		if (isEmbedded()) {
+			settings.applyConnectionString(new ConnectionString("mongodb://localhost:" + getEmbeddedPort()));
+			return;
+		}
+		if (!this.properties.determineUri().equals(MongoProperties.DEFAULT_URI)) {
+			if (hasCustomAddress()) {
+				Assert.state(this.properties.getUri() == null,
```
snicoll: This is breaking an existing test and I fail to see how that's related to the migration to the new driver. Can you please clarify?

christophstrobl: It seems to be a mismatch that existed between the reactive and imperative implementation. If you remove that `if` statement `ReactiveMongoClientFactoryTests.uriCannotBeSetWithHostPort` will fail which contradicts `MongoPropertiesTests.uriOverridesHostAndPort`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoAutoConfiguration.java
```diff
@@ -45,10 +47,13 @@
 public class MongoAutoConfiguration {
 
 	@Bean
-	@ConditionalOnMissingBean(type = { "com.mongodb.MongoClient", "com.mongodb.client.MongoClient" })
-	public MongoClient mongo(MongoProperties properties, ObjectProvider<MongoClientOptions> options,
-			Environment environment) {
-		return new MongoClientFactory(properties, environment).createMongoClient(options.getIfAvailable());
+	@ConditionalOnMissingBean(type = { "com.mongodb.client.MongoClient" })
+	public MongoClient mongo(MongoProperties properties, Environment environment,
+			ObjectProvider<MongoClientSettingsBuilderCustomizer> builderCustomizers,
```
snicoll: Now that the settings are shared by the two API, we can apply the same customizer consistently which is great. I've removed that in the case of the upgrade though as I'd prefer adding this in an isolated commit.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoClientFactory.java
```diff
@@ -39,102 +35,23 @@
  * @author Stephane Nicoll
  * @author Nasko Vasilev
  * @author Mark Paluch
+ * @author Christoph Strobl
  * @since 2.0.0
  */
-public class MongoClientFactory {
+public class MongoClientFactory extends MongoClientFactorySupport<MongoClient> {
```
snicoll: While I agree having a shared factory that behaves consistently, I am not sure we are ready to do this just yet. There's quite some history in the original MongoDB auto-configuration and sharing the code with the reactive counter part will change the semantics. I think we should keep the two separated for the time being and investigate what bits are problematic.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/mongo/MongoPropertiesTests.java
```diff
@@ -55,48 +58,43 @@ void canBindCharArrayPassword() {
 	@Test
 	@SuppressWarnings("deprecation")
 	void allMongoClientOptionsCanBeSet() {
```
snicoll: By restoring the code as it was before in my polish, this test fails. The replicat set is erased by the connection URI. I don't know if that's a new behaviour of the driver or if I messed up the change but I'd appreciate some help there.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/mongo/ReactiveMongoClientFactoryTests.java
```diff
@@ -117,13 +118,13 @@ void retryWritesIsPropagatedFromUri() {
 	}
 
 	@Test
-	void uriCannotBeSetWithCredentials() {
+	void uriCanBeSetWithCredentials() {
```
snicoll: I don't think we should be making that change as part of the upgrade to the new driver unless this use case is no longer supported.

snicoll:
This broke a smoke test as it looks like we don't connect to the embedded broker properly anymore

PR #19961, opened 2020-01-28
Fix Gradle plugin documentation links

dreis2211:
Hi,

I saw also one broken link in the Gradle Plugin docs that this PR will fix.

Cheers,
Christoph


snicoll:
Good catch, thank you so much.

PR #19986, opened 2020-01-30
Webflux coroutines smoke test improvement

SokoMishaLov:


wilkinsona:
Thanks for the proposal. Our smoke tests are intended to verify that functionality that Spring Boot has configured is working as expected. We don't have any specific configuration for the functionality that's covered by the proposed improvements. As such, I think that Spring Framework's build is a more appropriate place for this testing to be performed as that's where the functionality is implemented. Thanks anyway.

PR #20032, opened 2020-02-04
Use Dropwizard metrics BOM to manage dependencies

dreis2211:
Hi,

I just noticed that we could make use of Dropwizard's `metrics-bom` to manage the dependencies. While doing so, I upgraded to 4.1.2 directly, because it adds dependency management for `metrics-collectd`.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -250,25 +250,10 @@ bom {
 			]
 		}
 	}
-	library("Dropwizard Metrics", "4.1.1") {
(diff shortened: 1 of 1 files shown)


snicoll:
> While doing so, I upgraded to 4.1.2 directly

@dreis2211 I forgot to mention that I've polished the PR and downgraded as we prefer that dependency upgrades are done in an isolated commit with a dedicated issue so that users can find those more easily. I'll upgrade to `4.1.2` shortly.

dreis2211:
Thanks for the update. I wasn't sure if that's okay - thanks for clarifying it :)

PR #20047, opened 2020-02-05
Set includeantruntime in Ant smoke tests

dreis2211:
Hi,

this PR is an attempt to get rid of a warning inside `spring-boot-smoke-test-ant`:

```
compile:
    [javac] /path/to/spring-boot-smoke-test-ant/build.xml:32: warning: 'includeantruntime' was not set, defaulting to build.sysclasspath=last; set to false for repeatable builds
```

Cheers,
Christoph


snicoll:
That makes sense and removed the warning for me too. I am not sure what you meant by "an attempt to get rid of a warning" though. 

Thanks again!

dreis2211:
I'm writing defensively. I've seen the "works on my machine" too often and then it doesn't work on the actual build machine. In this case probably not needed, though. Will be more clear in the future.

Issue #20024, opened 2020-02-03
Unlike all other Maven properties, spring-boot.run.arguments on the command line takes precedence over the pom


rawfg:
**Description:** the `spring-boot.run.jvmArguments` does not take precedence over the `< systemPropertyVariables>` map when there is a `<jvmArguments>` section defined for the `spring-boot-maven-plugin`.

**Documented behavior:**
According to the https://docs.spring.io/spring-boot/docs/current/maven-plugin/examples/run-system-properties.html when the `spring-boot.run.jvmArguments` parameter is provided for the `spring-boot:run` command the parameter takes precedence and overwrites the system properties defined within the `<systemPropertyVariables>` XML configuration. I would expect that the presence of the `jvmArguments` section does not influence the documented behavior.

**Steps to reproduce:**
There is a sample application which demonstrates non-documented (faulty?) behavior located at https://github.com/rawfg/maven-jvmargs.

1. The sample application prints the system properties starting with the `xyz` and by default there are [two system properties](https://github.com/rawfg/maven-jvmargs/blob/master/pom.xml#L45-L50) configured: one comes from the `<jvmArguments>` section and the other one comes from the `< systemPropertyVariables>` section.

2. When the application is run from the command line via the command:
```bash
mvn spring-boot:run -Dspring-boot.run.jvmArguments="-Dxyz.MavenSystemProperty='cmd-system-property'" 
``` 
the output is (system properties configured in the pom.xml):
```bash
xyz.MavenSystemProperty=maven-system-property
xyz.MavenJvmProperty=maven-jvm-property
```

3. When the [`<jvmArguments>` section](https://github.com/rawfg/maven-jvmargs/blob/master/pom.xml#L46) is commented out the application behaves as it was documented, so running the command:
```bash
mvn spring-boot:run -Dspring-boot.run.jvmArguments="-Dxyz.MavenSystemProperty='cmd-system-property'" 
```
results in the output:
```bash
xyz.MavenSystemProperty=cmd-system-property
```
4. I tested it with the Spring Boot version `2.2.4` but I've downgraded to the `2.1.7` version and I observed the same outcome.

**Expected behavior:**
- `-Dspring-boot.run.jvmArguments` always works as documented (step 3) or
- the exception is documented in the https://docs.spring.io/spring-boot/docs/current/maven-plugin/examples/run-system-properties.html

scottfrederick:
[...]

PR #20058, opened 2020-02-07
Include licence and notice files in shipped jars

### LINKED ISSUE #15643: Include licence files in shipped jars
@mp911de noticed that we don't include license/notice files inside our JARs whereas Spring Framework does. We do have copyright on each file and a license in our POM but we should consider if we also want to include it in the shipped JARs.

dreis2211:
Hi,

this PR includes the `LICENSE.txt` file in the `META-INF` directory when a jar is built. This should fix #15643 .

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.build;
 
+import java.io.File;
 import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
@@ -51,6 +52,7 @@ import org.gradle.api.publish.maven.MavenPomScm;
 import org.gradle.api.publish.maven.MavenPublication;
 import org.gradle.api.publish.maven.plugins.MavenPublishPlugin;
 import org.gradle.api.publish.tasks.GenerateModuleMetadata;
+import org.gradle.api.resources.TextResourceFactory;
 import org.gradle.api.tasks.bundling.Jar;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.api.tasks.javadoc.Javadoc;
@@ -163,8 +165,9 @@ public class ConventionsPlugin implements Plugin<Project> {
 			String noticeContent = FileCopyUtils.copyToString(new InputStreamReader(notice, StandardCharsets.UTF_8))
 					.replace("${version}", project.getVersion().toString())
 					.replace("${currentYear}", Integer.toString(Year.now().getValue()));
-			metaInf.from(project.getResources().getText().fromString(noticeContent),
-					(noticeFile) -> noticeFile.rename((name) -> "NOTICE.txt"));
+			TextResourceFactory resourceFactory = project.getResources().getText();
+			File file = createLegalFile(resourceFactory.fromString(noticeContent).asFile(), "NOTICE.txt");
+			metaInf.from(file);
 		}
 		catch (IOException ex) {
 			throw new GradleException("Failed to copy NOTICE.txt", ex);
@@ -174,12 +177,19 @@ public class ConventionsPlugin implements Plugin<Project> {
 	private void copyLicenseFile(Project project, CopySpec metaInf) {
 		URL license = getClass().getClassLoader().getResource("LICENSE.txt");
 		try {
-			metaInf.from(project.getResources().getText().fromUri(license.toURI()),
-					(licenseFile) -> licenseFile.rename((name) -> "LICENSE.txt"));
+			TextResourceFactory resourceFactory = project.getResources().getText();
+			File file = createLegalFile(resourceFactory.fromUri(license.toURI()).asFile(), "LICENSE.txt");
+			metaInf.from(file);
 		}
 		catch (URISyntaxException ex) {
 			throw new GradleException("Failed to copy LICENSE.txt", ex);
 		}
+	}
+
+	private File createLegalFile(File source, String filename) {
+		File legalFile = new File(source.getParentFile(), filename);
+		source.renameTo(legalFile);
+		return legalFile;
 	}
 
 	private void configureSpringJavaFormat(Project project) {
--- a/buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
@@ -16,6 +16,13 @@
 
 package org.springframework.boot.build;
 
+import java.io.IOException;
+import java.io.InputStream;
+import java.io.InputStreamReader;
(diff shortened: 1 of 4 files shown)


buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
```diff
@@ -127,6 +127,7 @@ private void applyJavaConventions(Project project) {
 			});
 			project.getTasks().withType(Jar.class, (jar) -> {
 				project.afterEvaluate((evaluated) -> {
+					jar.metaInf((metaInf) -> metaInf.from(project.getRootProject().file("LICENSE.txt")));
```
snicoll: I am not sure if we want to ship our license or something a bit more complete as the one [Spring Framework use](https://github.com/spring-projects/spring-framework/blob/master/src/docs/dist/license.txt).

dreis2211: Let me know if I can help with something. Seems to be outside of my control at the moment, though!?

dreis2211:
I've added a test, @snicoll . Deciding what LICENSE.txt and/or if a notice.txt should be added, seems to outside of my hands at the moment, but I'm happy to add anything after you made a decision.

snicoll:
Thanks Christoph. I've added this one for team attention so that we figure it out during our next meeting.

buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
wilkinsona: I don't think the Framework one matches our needs as it includes information about CGLib and ASM which Framework jarjars. The file in the root of the repo looks like a better fit to me for most of our modules at least. We may need to do something different for `spring-boot-configuration-processor` as it shades some JSON stuff.

snicoll: > I don't think the Framework one matches our needs as it includes information about CGLib and ASM which Framework jarjars. 

I didn't literally mean that file but more something with more details than just our license. I was actually referring [to the bottom of the file](https://github.com/spring-projects/spring-boot/issues/20024#issuecomment-585151961). Not feeling strongly about it though.

wilkinsona:
We discussed this today and concluded that:

- We should include a `LICENSE.txt` file in the deployed jars with the same contents as the file in the root of the repo
- Rather than using the file in the root of the repo, we'd like to use a file in `src/main/resources` of the `buildSrc` plugins.
- We should include a `NOTICE.txt` file in the deployed jars. The file should have the following contents:
```
Spring Boot ${version}
Copyright (c) 2012-${currentYear} Pivotal, Inc.

This product is licensed to you under the Apache License, Version 2.0
(the "License"). You may not use this product except in compliance with
the License.
```

@dreis2211 If the offer still stands, would you like to update this PR to reflect the above?

dreis2211:
Sure, @wilkinsona . Give me a minute or two.

dreis2211:
Okay it was an hour, but it should be done now. Let me know if this is what you had in mind.

BTW: I have a bit of trouble to run the tests in `buildSrc` from IDEA which makes it quite a bit harder to debug them. Is there any trick to it?

wilkinsona:
> BTW: I have a bit of trouble to run the tests in buildSrc from IDEA which makes it quite a bit harder to debug them. Is there any trick to it?

Sorry, I'm not sure how IDEA handle's `buildSrc`. In Eclipse, I've just imported it alongside all the other modules and can then run its tests as usual.

buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
```diff
@@ -141,6 +152,36 @@ private void applyJavaConventions(Project project) {
...
+		copyLicenseFile(project, metaInf);
+	}
+
+	private void copyNoticeFile(Project project, CopySpec metaInf) {
+		try {
+			InputStream notice = getClass().getClassLoader().getResourceAsStream("NOTICE.txt");
+			String noticeContent = FileCopyUtils.copyToString(new InputStreamReader(notice, StandardCharsets.UTF_8))
+					.replace("${version}", project.getVersion().toString())
+					.replace("${currentYear}", Integer.toString(Year.now().getValue()));
+			metaInf.from(project.getResources().getText().fromString(noticeContent),
```
wilkinsona: I think this'll break up-to-date checking for the jar. Creating a text resource from a string results in the creation of a temporary file with a random name. As that random name varies from build-to-build, the jar tasks will never be considered up-to-date as it'll look like its inputs have changed.

We had this problem when adding the classpath index support to `BootJar`. The solution was to get the text resource's file and rename it before adding it to the copy spec rather than as part of the spec.

snicoll:
No problem here either but I am on `2020.1 EAP`. As long as you delegate to Gradle, it should just work(tm).

buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
dreis2211: I will take a look how it was done there and report back

wilkinsona: Thanks. I should have linked to the relevant code earlier. It is the following:

https://github.com/spring-projects/spring-boot/blob/16111f126e707fb5064bab6150df7a77f54850ee/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootJar.java#L113-L119

This will result in a `NOTICE.txt` and `LICENSE.txt` file beneath each project's `build` directory. An alternative could be to access the files in `buildSrc/src/main/resources` directly from the filesystem.

dreis2211: I tried that and that didn't work. Will take a look again.

dreis2211: After the latest changes I get the following:
```
 ./gradlew :spring-boot-project:spring-boot-properties-migrator:jar

BUILD SUCCESSFUL in 1s
9 actionable tasks: 9 up-to-date
```

Previously, I got the following:
```
 ./gradlew :spring-boot-project:spring-boot-properties-migrator:jar

BUILD SUCCESSFUL in 1s
9 actionable tasks: 2 executed, 7 up-to-date
```

Thanks for the good catch, @wilkinsona

wilkinsona:
It's just occurred to me that the suggestion to use ${currentYear} was a bad one as it'll break the repeatability of the build. We should probably just hardcode it and update as needed. We can make that amendment as part of merging.

wilkinsona:
Thanks very much, @dreis2211.

dreis2211:
Thanks for the proper review, @wilkinsona

PR #20102, opened 2020-02-08
Reinstate Cassandra smoke tests

### LINKED ISSUE #10453: Upgrade to a Cassandra version that works with Java 9
See [CASSANDRA-9608](https://issues.apache.org/jira/browse/CASSANDRA-9608). Our sample is disabled on Java 9 at the moment.

See also https://github.com/jsevellec/cassandra-unit/issues/249

dreis2211:
Hi,

I just noticed that `spring-boot-smoke-test-data-cassandra` is doing nothing. I traced this back to ca1710ee56d327e46a76f028f7bbc12dc974ee5e, but wasn't sure if the removal was intentional.

Reinstating the tests required some changes as they didn't work anymore as they were. While working on it, I noticed that this should effectively fix #10453 as we're on a version that works with Java 9 now and my PR makes use of `testcontainers` instead of the old `cassandra-unit-spring` approach.

If you think, they shouldn't be brought back at all, I'm also happy to simply remove the complete folder.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-data-cassandra/src/test/java/smoketest/data/cassandra/SampleCassandraApplicationTests.java
+++ b/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-data-cassandra/src/test/java/smoketest/data/cassandra/SampleCassandraApplicationTests.java
@@ -1,0 +1,71 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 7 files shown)


spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-data-cassandra/src/test/java/smoketest/data/cassandra/SampleCassandraApplicationTests.java
```diff
@@ -0,0 +1,71 @@
...
+@ContextConfiguration(initializers = SampleCassandraApplicationTests.Initializer.class)
+@Testcontainers(disabledWithoutDocker = true)
+class SampleCassandraApplicationTests {
+
+	@Container
+	static final CassandraContainer<?> container = new CassandraContainer<>().withStartupAttempts(5)
+			.withStartupTimeout(Duration.ofMinutes(10)).withInitScript("setup.cql");
+
+	@Test
+	@DisabledOnOs(OS.WINDOWS)
```
bsideup: Just curious: why it doesn't work on Windows?

snicoll:
>  I traced this back to ca1710e, but wasn't sure if the removal was intentional.

It was, yes. I missed the removal of the build file and the directory and I've [just done that](https://github.com/spring-projects/spring-boot/commit/bde7bd0a1a310f48fb877b9a0d4a05b8d829d6c0). Thanks for letting us know.

This is an artifact of when smoke tests were also called "samples". Now they really are smoke tests (and no longer samples) we felt it wasn't warranted to keep it as it doesn't do much that the integration tests are already doing. Thanks anyway.

PR #20113, opened 2020-02-10
Remove redundant jar task configuration

dreis2211:
Hi,

I just noticed that we have a jar configuration in `spring-boot-gradle-plugin` that seems redundant given that we apply the conventions plugin, which should add this. Am I missing something?

Cheers,
Christoph


snicoll:
Nothing I can think of. The manifest is the same with your change so I went ahead and merged it. Thanks again!

PR #20140, opened 2020-02-12
Proposed Fix for issue 20135

### LINKED ISSUE #20135: BuildInfo task not run in Gradle plugin when version number changes
If I disable the time property, the Gradle BuildInfo task indicates that it is up to date even when the project version number changes. 

In our configuration, we have changed the build info configuration to exclude the time property as suggested in the documentation:
```
springBoot {
    buildInfo {
        properties {
             time = null
        }
    }
}
```
Now, when we alter the project version number, this task always indicates that it is up to date and so the incorrect version number is included in the build-info files.
Looking at the code, I suspect the problem is that the `BuildInfoProperties` class has a null value for the version number by default, and so each time the input is checked, Gradle only sees this null value. It never changes to reflect the changing version number.

Using version 2.2.2.RELEASE of the plugin.

ddebree:
Default the properties in the constructor
Fixes #20135

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/buildinfo/BuildInfoProperties.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/buildinfo/BuildInfoProperties.java
@@ -48,7 +48,7 @@ public class BuildInfoProperties implements Serializable {
 
 	BuildInfoProperties(Project project) {
 		this.project = project;
-		//Default the properties:
+		// Default the properties:
 		this.group = project.getGroup().toString();
 		this.name = project.getName();
 		this.version = project.getVersion().toString();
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@ddebree Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20140)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20140) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@ddebree Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20140)!

wilkinsona:
Thanks for the pull request, but I think this will create a potential ordering problem where a change to a value may not be noticed if it's made after they've been captured. I've pushed a different approach and amended the faulty test in https://github.com/spring-projects/spring-boot/commit/c8907d46b40348f47a60055ca4016a6cba95d90c. Thanks for bringing the problem to our attention.

ddebree:
No worries. Figured there was more to this issue than just what I proposed. Thanks for the fix.

PR #20141, opened 2020-02-12
Exclude jcl-over-slf4j also on Reactive Cassandra starter

dreis2211:
Hi,

the missing exclude for `jcl-over-slf4j` is currently failing the build.

Cheers,
Christoph


snicoll:
🤦‍♂️

PR #20052, opened 2020-02-06
Invalid variable reference in maven-shade-plugin configuration

scheerer:
When trying to use the `maven-shade-plugin` plugin in a project using the latest 2.3.0.M1 milestone release we encounter this error:

```
[ERROR] [ERROR] Some problems were encountered while processing the POMs:
[ERROR] 'build.plugins.plugin[org.apache.maven.plugins:maven-shade-plugin].dependencies.dependency.version' for org.springframework.boot:spring-boot-maven-plugin:jar must be a valid version but is '${revision}'. @ org.springframework.boot:spring-boot-starter-parent:2.3.0.M1, /Users/demo/.m2/repository/org/springframework/boot/spring-boot-starter-parent/2.3.0.M1/spring-boot-starter-parent-2.3.0.M1.pom, line 211, column 24
```

Build Settings:
```
<build>
    <plugins>
        <plugin>
            <groupId>org.apache.maven.plugins</groupId>
            <artifactId>maven-shade-plugin</artifactId>
            <configuration>
                <createDependencyReducedPom>false</createDependencyReducedPom>
            </configuration>
            <executions>
                <execution>
                    <phase>package</phase>
                    <goals>
                        <goal>shade</goal>
                    </goals>
                </execution>
            </executions>
        </plugin>
    </plugins>
</build>
```

I believe this PR fixes what seems to be a very slight oversight during the Gradle migration.

PR #20143, opened 2020-02-12
Invalid variable reference in maven-shade-plugin configuration

scheerer:
This should fix my prior PR #20052. I am pretty sure this is correct, but sadly I seem to be unable to fully build the project to verify for sure. Might need a couple eyes from build masters to ensure. I am so sorry for the bad fix in #20052.

History:
1) after the Gradle migration it was incorrectly left as `${revision}`
2) I blindly used Maven conventions to set it to `${project.version}`
3) After trying to `2.3.0.BUILD-SNAPSHOT` build with my change I discovered ${project.version} always resolves to `0.0.1-SNAPSHOT`.
4) Here is where I think I found the rest of the Gradle setup uses `${version}` for the current SpringBoot build version. So I used that. :)

In order to know for sure a simple child project inheriting from the spring-boot-starter-parent 2.3.0.BUILD-SNAPSHOT must use the `shade` plugin like so:

```
<build>
    <plugins>
        <plugin>
            <artifactId>maven-shade-plugin</artifactId>
            <configuration>
                <createDependencyReducedPom>false</createDependencyReducedPom>
            </configuration>
            <executions>
                <execution>
                    <phase>package</phase>
                    <goals>
                        <goal>shade</goal>
                    </goals>
                </execution>
            </executions>
        </plugin>
    </plugins>
</build>
```

Then simply execute `./mvnw package` to see the error.

### DIFF
--- a/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
+++ b/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
@@ -191,7 +191,7 @@ publishing.publications.withType(MavenPublication) {
 								dependency {
 									delegate.groupId('org.springframework.boot')
 									delegate.artifactId('spring-boot-maven-plugin')
-									delegate.version('${project.version}')
+									delegate.version('${version}')
 								}
 							}
 							executions {


spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
```diff
@@ -191,7 +191,7 @@ publishing.publications.withType(MavenPublication) {
 								dependency {
 									delegate.groupId('org.springframework.boot')
 									delegate.artifactId('spring-boot-maven-plugin')
-									delegate.version('${project.version}')
+									delegate.version('${version}')
```
snicoll: `gradle publishToMavenLocal` generates the following for me: 

```
<dependencies>
            <dependency>
              <version>${version}</version>
            </dependency>
          </dependencies>
```

I believe it should generate the following:

```
<dependencies>
            <dependency>
              <version>2.3.0.BUILD-SNAPSHOT</version>
            </dependency>
          </dependencies>
```

wilkinsona: The following will work:

```groovy
delegate.version("${project.version}")
```

This aligns with how the parent version is configured earlier in the same `build.gradle` file:

```groovy
parent {
	delegate.groupId("${project.group}")
	delegate.artifactId("spring-boot-dependencies")
	delegate.version("${project.version}")
}
```

The change can be verified with the following command:

```
./gradlew spring-boot-project:spring-boot-starters:spring-boot-starter-parent:generatePom
```

This will write the generated pom to `spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build/publications/maven/pom-default.xml`

scheerer: @wilkinsona  If I'm understanding you correctly does that mean this PR is simply not needed?

wilkinsona: No, I think it's still needed, but in a modified form. We currently have `'${project.version}'`. We need `"${project.version}"` instead. Note the use of `"` rather than `'`.

snicoll:
@scheerer thanks for letting us know about the issue. I took the liberty of applying Andy's suggestion myself as we are releasing M2 today and I'd like to make sure the starter has a valid POM.

Issue #20147, opened 2020-02-12
Add Java 14 CI


snicoll:
With Java 14 now in release candidate phase, we ned to add a compatibility build to 2.3. There are two things to consider when doing so:

* Gradle does not support Java 14 so we'll have to install two different JDKs (one to run Gradle and one to compile & tests the code). The following [section of the Gradle doc](https://docs.gradle.org/current/userguide/building_java_projects.html#sec:java_cross_compilation) is related
* An upgrade to Groovy 3 is required. This is transparent but the CLI (see #20119). The plan is to override the version of Groovy and ignore the CLI for the time being.

dreis2211:
This time you beat me to it :D

snicoll:
Ah ah. PR still most welcome :-)

dreis2211:
I can offer to provide some stuff for JDK 14 in general in smaller chunks, if that's okay for you? I already have a couple of things ( [see my jdk-14-wip branch](https://github.com/spring-projects/spring-boot/compare/master...dreis2211:jdk-14-wip) ) but to make one big PR at which end we have a (green) CI for JDK 14 I'm missing the time, at the moment. I could offer the following 3 PRs (or at least independent commits) for now based on the WIP branch:

- Implement possibility to specify custom java home (so you guys can play around with it too locally)
- Fix JarFileTests for multi release jars on JDK 14
- Fix NoSuchBeanDefinitionFailureAnalyzerTests on JDK 14 (see [JDK-8164819](https://bugs.openjdk.java.net/browse/JDK-8164819)  & related tickets for the background )

Let me know what you think.
Cheers,
Christoph

snicoll:
Thanks @dreis2211. The first sounds like a good start. As for the two others, a single PR  should work. The next step is to crate an image that allows CI to run the build with Java 14.

dreis2211:
It looks like we might get a Groovy 2.5.10 release with JDK 14 compatibility.

snicoll:
Closing in favour of PR #20413

Issue #17259, opened 2019-06-19
Don't build docs on JDK > 8


philwebb:
It's a waste of time really.

snicoll:
Do you include javadoc generation? If so, I think we should keep it as recent java versions have changed requirements in subtle ways where javadoc would run on Java 8 and broke on 11 or 12.

wilkinsona:
I think we could afford to only build the javadoc on Java 8. The problems with 11 and 12 are almost always just a missing dependency. I don't think we'd be storing up too much trouble by not being aware of those till Java 11 becomes the baseline. IMO, the cost of that trouble would be outweighed by the reduced build execution time.

dreis2211:
This is problematic on another level: As of [JDK-8180019](https://bugs.openjdk.java.net/browse/JDK-8180019) the javadoc generation fails if a link can't be accessed. This happened a handful of times this week. For example:

- https://ci.spring.io/teams/spring-boot/pipelines/spring-boot-2.3.x/jobs/jdk11-build/builds/331
- https://ci.spring.io/teams/spring-boot/pipelines/spring-boot-2.3.x/jobs/jdk13-build/builds/330

As the move to Gradle has reduced build times quite a lot I wonder if we can repurpose this one to ignore failures at least on JDK 11+ in an effort to make the build pipeline overall greener. E.g. I could imagine the following (yet untested) patch:
```patch
--- a/buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
@@ -24,6 +24,7 @@ import io.spring.javaformat.gradle.FormatTask;
 import io.spring.javaformat.gradle.SpringJavaFormatPlugin;
 import org.apache.maven.artifact.repository.MavenArtifactRepository;
 import org.asciidoctor.gradle.jvm.AsciidoctorJPlugin;
+import org.gradle.api.JavaVersion;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.artifacts.DependencySet;
@@ -119,8 +120,10 @@ public class ConventionsPlugin implements Plugin<Project> {
                                        args.add("-parameters");
                                }
                        });
-                       project.getTasks().withType(Javadoc.class,
-                                       (javadoc) -> javadoc.getOptions().source("1.8").encoding("UTF-8"));
[...]

PR #20179, opened 2020-02-14
Implement possibility to set custom java home for Gradle tasks

dreis2211:
Hi,

as discussed in #20147 this PR should give us a first insight into what might be broken with JDK 14.

The new functionality can be used like this:
```
./gradlew -PcustomJavaHome=/Users/christoph.dreis/.sdkman/candidates/java/14-ea build
```

I specifically implemented a check for emptiness of the new property in preparation for an eventual CI. By that we could simply set it for JDK 14 builds and leave it empty for the others. That should keep the build-project.sh relatively simple.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
@@ -115,8 +115,8 @@ public class ConventionsPlugin implements Plugin<Project> {
 			project.setProperty("sourceCompatibility", "1.8");
 			project.getTasks().withType(JavaCompile.class, (compile) -> {
 				compile.getOptions().setEncoding("UTF-8");
-				if (hasCustomJavaHome(project)) {
-					String javaExecutable = getCustomJavaExecutable(project, "/bin/java");
+				if (hasCustomBuildJavaHome(project)) {
+					String javaExecutable = getCustomBuildJavaExecutable(project, "/bin/java");
 					compile.getOptions().getForkOptions().setJavaHome(new File(javaExecutable));
 				}
 				List<String> args = compile.getOptions().getCompilerArgs();
@@ -126,14 +126,14 @@ public class ConventionsPlugin implements Plugin<Project> {
 			});
 			project.getTasks().withType(Javadoc.class, (javadoc) -> {
 				javadoc.getOptions().source("1.8").encoding("UTF-8");
-				if (hasCustomJavaHome(project)) {
-					String javaExecutable = getCustomJavaExecutable(project, "/bin/javadoc");
+				if (hasCustomBuildJavaHome(project)) {
+					String javaExecutable = getCustomBuildJavaExecutable(project, "/bin/javadoc");
 					javadoc.setExecutable(javaExecutable);
 				}
 			});
 			project.getTasks().withType(Test.class, (test) -> {
-				if (hasCustomJavaHome(project)) {
-					String javaExecutable = getCustomJavaExecutable(project, "/bin/java");
+				if (hasCustomBuildJavaHome(project)) {
+					String javaExecutable = getCustomBuildJavaExecutable(project, "/bin/java");
 					test.setExecutable(javaExecutable);
 				}
 				test.useJUnitPlatform();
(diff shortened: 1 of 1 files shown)


snicoll:
> the changes would then be focused on Test tasks alone

Don't we want to compile with JDK14 as well?

wilkinsona:
> Don't we want to compile with JDK14 as well?

Oops. I'd completely overlooked that part of the PR.

I'm not sure that we do. The binaries that we ship will be built with Java 8 so our CI will be closer to what users will be running if we compile with Java 8 and test with JDK 14. This applies to our JDK 11 and 13 builds too, I just hadn't thought about it before.

dreis2211:
Anything I can do to help? I'm a bit confused what you want me to do ;-)

dreis2211:
> This applies to our JDK 11 and 13 builds too

I have trouble to actually see where the JDK 11+ builds do compile with JDK 8. Or do you mean that they do exactly the same as proposed here? And you simply wonder if that's the right approach?

Anyhow, I do see some value in compiling with the newer JDK versions as that will show deprecation warnings etc. E.g. compiling with JDK 11 shows warnings about using `Class.newInstance()` instead of `Class.getDeclaredConstructor().newInstance()`.

For the javadocs, this overlaps with #17259 if you ask me. Again, I see some value in generating them with the respective versions as that will show eventual changes in behaviour better. Similar to what @snicoll mentioned there already.

In summary, being aware of changes on the JDK level - let it be on compiler, javadoc or test side - is imho beneficial. But I'm also fine with removing the javadoc & compile part if you really want me to.

wilkinsona:
Sorry for the confusion, @dreis2211.

> do you mean that they do exactly the same as proposed here? And you simply wonder if that's the right approach?

Yes, that's what I meant. I was trying to acknowledge that the problem I described with what's proposed here also already exists in our JDK 11 and 13 builds.

> Anyhow, I do see some value in compiling with the newer JDK versions as that will show deprecation warnings etc. E.g. compiling with JDK 11 shows warnings about using Class.newInstance() instead of Class.getDeclaredConstructor().newInstance().

There's definitely some value. However, if we're going to pick one approach over the other, I think that testing Java 8-compiled code against later JDKs is of greater value as that's what our users will be doing. Feeling this way is also influenced by the fact that there's often nothing we can do about the deprecation warnings while we continue to support Java 8, although that's not the case with the example above.

A downside of compiling with JDK 8 and testing with a later JDK is that it complicates our CI set up. We currently have everything in place to produce Docker images with a single JDK in a well-known location. Adopting the approach I've described above would require our JDK 11, 13, and 14 images to have two JDKs in place. That's unavoidable for JDK 14 with Gradle's current capabilities no matter which approach we take so maybe it's not so bad.

dreis2211:
> That's unavoidable for JDK 14 with Gradle's current capabilities no matter which approach we take so maybe it's not so bad.

Exactly. It feels like we're discussing the bigger picture (which I enjoy) while this PR seems needed anyway for the time being and is only a tiny piece of the complete JDK 14 CI process...

wilkinsona:
@dres2211 Thanks again for your work on this one. We discussed it and the broader topic today and concluded that we'd like two properties:

1. `buildJavaHome` for controlling the Java home that's used for the whole build. This should affect compilation, javadoc, and tests
2. `testJavaHome` for controlling the Java home that's used for tests.

We'll use `testJavaHome` on CI so that we can test Java 8-compiled byte code with later JDKs. This will cover what users are doing with our published binaries. We'll then use `buildJavaHome` on our own machines while exploring compatibility with JDKs that Gradle itself cannot yet run on.

This PR pretty much covers 1, other than renaming `customJavaHome` to `buildJavaHome`. If you have time to make that tweak, that'd be much appreciated, otherwise we can handle it as part of merging. We'll then add the other property as a separate task.

dreis2211:
Sure...will do so in a couple of minutes.

dreis2211:
Pushed

wilkinsona:
Thanks for the updates, @dreis2211. I've polished this a bit (https://github.com/wilkinsona/spring-boot/commit/563083b25f50b0c18f5aeaa702e8045b52b64527) and in doing so, I think I've noticed the `CompileJava` tasks aren't using the custom Java home or `javac` executable. I can run a build with `-PbuildJavaHome=/whatever` and test and javadoc tasks will fail, but `CompileJava` tasks do not. I don't think I've introduced the problem while polishing, but I need to dig a bit more to be certain.

wilkinsona:
And almost immediately after writing the above, I realised that we need to call `compile.getOptions().setFork(true)`. As expected, failures now occur when the configured Java home does not exist:

```
Execution failed for task ':spring-boot-project:spring-boot-tools:spring-boot-loader-tools:compileJava'.
> Supplied javaHome must be a valid directory. You supplied: /whatever
```

dreis2211:
@wilkinsona As I saw test failures I didn't question that the custom Java home was used. Good catch! Sorry that I didn't catch it. :/

wilkinsona:
Thanks once again, @dreis2211. The proposed changes are now on the master branch.

PR #20193, opened 2020-02-15
Explicitly set java home in Maven Plugin integration tests

dreis2211:
Hi,

I'm regularly switching between JDKs at the moment and noticed that `RunIntegrationTests` fails if IDEA is setup with JDK 13+ while the actual Java home (e.g. inside the terminal) is set to something lower. 

This PR sets the Java home explicitly on the `InvocationRequest` in `MavenBuild`, so the Java version for the maven invocation matches the one that actually runs the test.

Cheers,
Christoph


wilkinsona:
Thanks very much once again, @dreis2211!

PR #20205, opened 2020-02-16
Provide a configuration property for the maximum size of the in-memory buffer used by the reactive Elasticsearch REST client

wonwoo:


wilkinsona:
Thanks for the PR, @wonwoo. I think we may need to take a step back and consider if this is the right approach.

We already have a property, `spring.codec.max-in-memory-size` for configuring reactive codecs. Those used by the reactive Elasticsearch client fall into that category. I think we need to decided if we want the existing property to apply to the reactive Elasticsearch client or if we want to provide a separate property that is specific to reactive Elasticsearch. If we go with the latter option, we should probably review the description of the existing property to make its scope more clear.

wilkinsona:
We discussed this today and concluded that we should have separate properties as the limits should be separate. It may be reasonable for a request or response to Elasticsearch to be large while wanting HTTP request and responses to have a smaller limit.

wilkinsona:
As part of merging this, we should clarify the scope of `spring.codec.max-in-memory-size`.

PR #20225, opened 2020-02-18
Polish log4j2 configuration

mustafau:
Tags starting with lowercase characters are inconsistent with others.


snicoll:
@mustafau well spotted. Thank you for making your first contribution to Spring Boot.

PR #20230, opened 2020-02-19
reformatting namespace 'ext' from Liquibase databaseChangelog element to fix local test failures

mikesmithson:
this is a  change to db.changelog-override.xml to fix LiquibaseAutoConfigurationTests#changelogXml() test to fix a local test failure.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/resources/db/changelog/db.changelog-override.xml
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/resources/db/changelog/db.changelog-override.xml
@@ -1,20 +1,19 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <databaseChangeLog
-        xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
-        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-        xmlns:ext="http://www.liquibase.org/xml/ns/dbchangelog-ext"
-        xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd
-        http://www.liquibase.org/xml/ns/dbchangelog-ext https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-ext.xsd">
+			xmlns="http://www.liquibase.org/xml/ns/dbchangelog"
+		xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		xsi:schemaLocation="http://www.liquibase.org/xml/ns/dbchangelog
+		                      http://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd">
 
-    <changeSet id="1" author="marceloverdijk">
-        <createTable tableName="customer">
-            <column name="id" type="int" autoIncrement="true">
-                <constraints primaryKey="true" nullable="false"/>
-            </column>
-            <column name="name" type="varchar(50)">
-                <constraints nullable="false"/>
-            </column>
-        </createTable>
-    </changeSet>
+	<changeSet id="1" author="marceloverdijk">
+		<createTable tableName="customer">
+			<column name="id" type="int" autoIncrement="true">
+				<constraints primaryKey="true" nullable="false"/>
+			</column>
+			<column name="name" type="varchar(50)">
+				<constraints nullable="false"/>
+			</column>
+		</createTable>
+	</changeSet>
 
 </databaseChangeLog>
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the suggestion.

The Liquibase FAQ [includes the ext namespace declaration](https://www.liquibase.org/faq.html#what-is-all-that-stuff-at-the-beginning-of-my-xml-changelog) and does the [XML format example](https://www.liquibase.org/documentation/xml_format.html). While the XML file you've altered doesn't, I think, use the namespace, I'm reluctant to remove it as we'd no longer be verifying that a changelog with the recommended namespaces works with Spring Boot.

Have you been able to determine why it fails for you locally?

mikesmithson:
> Thanks for the suggestion.
> 
> The Liquibase FAQ [includes the ext namespace declaration](https://www.liquibase.org/faq.html#what-is-all-that-stuff-at-the-beginning-of-my-xml-changelog) and does the [XML format example](https://www.liquibase.org/documentation/xml_format.html). While the XML file you've altered doesn't, I think, use the namespace, I'm reluctant to remove it as we'd no longer be verifying that a changelog with the recommended namespaces works with Spring Boot.
> 
> Have you been able to determine why it fails for you locally?

I am seeing the following failure when running this test: ` org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'liquibase' defined in org.springframework.boot.autoconfigure.liquibase.LiquibaseAutoConfiguration$LiquibaseConfiguration: Invocation of init method failed; nested exception is liquibase.exception.ChangeLogParseException: Error parsing line 7 column 124 of classpath:db/changelog/db.changelog-override.xml: schema_reference.4: Failed to read schema document 'https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd', because 1) could not find the document; 2) the document could not be read; 3) the root element of the document is not <xsd:schema>.`

mikesmithson:
It looks like using http 

> Thanks for the suggestion.
> 
> The Liquibase FAQ [includes the ext namespace declaration](https://www.liquibase.org/faq.html#what-is-all-that-stuff-at-the-beginning-of-my-xml-changelog) and so does the [XML format example](https://www.liquibase.org/documentation/xml_format.html). While the XML file you've altered doesn't, I think, use the namespace, I'm reluctant to remove it as we'd no longer be verifying that a changelog with the recommended namespaces works with Spring Boot.
> 
> Have you been able to determine why it fails for you locally?

It looks like there's an issue with using https in pointing to https://www.liquibase.org/xml/ns/dbchangelog/dbchangelog-3.0.xsd in the xsi:schemaLocation attribute. When I change it from https to http , the failing changelogXml() test goes green.

wilkinsona:
> Is there a reason why the test Liquibase xml's are using https and not http in them?

Yes, using HTTP introduces the risk of a man-in-the-middle attack. We moved to HTTPS (and enforce it via [NoHttp](https://spring.io/blog/2019/06/10/announcing-nohttp)) to avoid the problem described in [this blog post](https://medium.com/@jonathan.leitschuh/want-to-take-over-the-java-ecosystem-all-you-need-is-a-mitm-1fc329d898fb).

I think the problem with HTTPS must be specific to your environment. It works fine on CI and on the local machines of everyone else that I'm aware of. Thanks for you efforts here, but we won't be able to make a switch from `https` to `http` so I'm going to close this pull request.

mikesmithson:
@wilkinsona  OK, sounds good. Is this the only time you've ever heard of an external contributor having this issue?  I guess my expectation is that I would simply clone and run './gradlew build' without having to configure a local machine to be able to run a full end-to end build.

wilkinsona:
> Is this the only time you've ever heard of an external contributor having this issue?

Yes, it is.

> I guess my expectation is that I would simply clone and run './gradlew build' without having to configure a local machine to be able to run a full end-to end build.

Barring any environmental issues, that's my expectation too.

snicoll:
> Is this the only time you've ever heard of an external contributor having this issue?

Yes.

>  I guess my expectation is that I would simply clone and run './gradlew build' without having to configure a local machine to be able to run a full end-to end build. 

That's the right expectations. If something works with `http` but doesn't for `https` while it works with `https` on every environment we have access to (quite a lot), this is very much likely to be specific to your environment.

scottfrederick:
@mikesmithson Is there a chance you have a proxy in between your local machine and `https://www.liquibase.org` that requires authentication or otherwise might be blocking the connection?

mikesmithson:
> @mikesmithson Is there a chance you have a proxy in between your local machine and `https://www.liquibase.org` that requires authentication or otherwise might be blocking the connection?
@scottfrederick  - no proxy configuration, which usually solves about 99% of my problems. I can do some network traffic inspection to see what is going on but this is a brand new personal MacBook pro running Catalina that is running out of the box and I configured it myself. I am sure something will reveal itself eventually.

PR #9758, opened 2017-07-14
Fix: additional-spring-configuration-metadata.json is not being merged when building with Gradle 4

mmoayyed:
Closes #9732 

As an additional check, try to locate the json metadata by taking into account the presence of a new 'java' directory in the path.

Issue #9732, opened 2017-07-11
additional-spring-configuration-metadata.json is not being merged when building with Gradle 4


dima767:
Set up the simplest possible app according to the documentation wrt generating spring configuration metadata along with `additional-spring-configuration-metadata.json` for those additional "hints" to be merged into the main metadata file. It does not seem to work.

https://github.com/dima767/additional-config-metadata-not-working

Try `./gradlew clean build` or `./gradlew clean bootRun`

This is Boot `1.5.4` and Gradle `4.0`

/cc @mmoayyed

snicoll:
For the record, this project does work with Gradle 3.x as far as I can see (and we don't support Gradle 4 officially in the 1.5.x line). I wonder if that's a regression in Gradle 4. Perhaps @wilkinsona knows?

dima767:
Ah, OK, let me try with Gradle 3
		

		
	





On Wed, Jul 12, 2017 at 11:49 AM -0400, "Stéphane Nicoll" <notifications@github.com> wrote:












For the record, this project does work with Gradle 3.x as far as I can see (and we don't support Gradle 4 officially in the 1.5.x line). I wonder if that's a regression in Gradle 4. Perhaps @wilkinsona knows?

wilkinsona:
The problem's caused by Gradle changing the output location of classes in Gradle 4. The output location of resources has not changed which means that we fail to find the additional metadata file relative to the classes location.

In Gradle < 4 the layout is like this:

```
app/build/
├── classes
│   └── main
│       └── com
│           └── example
│               └── SomeClass.class
├── resources
│   └── main
│       └── META-INF
│           └── additional-spring-configuration-metadata.json
└── tmp
    └── compileJava
```

With Gradle 4 it now looks like this:

```
app/build/
├── classes
│   └── java
│       └── main
│           └── com
│               └── example
│                   └── SomeClass.class
├── resources
│   └── main
│       └── META-INF
│           └── additional-spring-configuration-metadata.json
└── tmp
    └── compileJava
```

Note the new `java` directory beneath `build/classes`

dima767:
From Gradle `4.0.1` release notes:

```
In projects that use multiple JVM languages (Java and Scala, Groovy and other languages) in separate source directories (e.g., src/main/groovy and src/main/java), Gradle now uses separate output directories for each language.

To return to the old behavior, explicitly set the classes directory:

// Change the output directory for the main source set back to the old path
[...]

PR #20323, opened 2020-02-26
Recommend configuring compileJava's inputs to include processResources' output when generating configuration metadata

larsgrefer:
`compileJava.dependsOn(processResources)` is not enough to ensure the correct behavior.

Using `dependsOn` only affects the execution order (`processResources` before `compileJava`) but not the up-to-date check of `compileJava`: After modifying `META-INF/additional-spring-configuration-metadata.json` the `processResouces` task will considered out-of-date and will be re-executed, but after that `compileJava` will still be considered up-to-date which causes the changes not to be merged into `META-INF/spring-configuration-metadata.json`

With this change the up-do-date check of `compileJava` is affected, too. Therefore it will correctly re-execute the configuration-processor when `META-INF/additional-spring-configuration-metadata.json` was changed.

Related Issues:

#9758
#9732


wilkinsona:
Thanks very much, @larsgrefer. The proposed change is now in 2.2.x and master.

Issue #17589, opened 2019-07-19
Unify SSL support


philwebb:
We currently have SSL configuration available in:

- `ServerProperties`
- `CouchbaseProperties`
- `KafkaProperties`
- `RabbitProperties`

In addition [we've had a request](https://github.com/spring-projects/spring-boot/issues/17365) to add them to `ActiveMQProperties`.

It feels like there would be some benefit in trying to unify our support.

Related issues: 

- [x] #34814
- [x] #25602
- [x] #34811
- [x] #34815
- [ ] #17365
- [ ] #34817
- [x] #35042
- [x] #18556
- [x] #34810
- [ ] #34839
- [x] #35155 
- [x] #37833
- [x] #37629

mp911de:
There's probably more such as Lettuce and Jedis (Redis Drivers).

jliu666:
It's time to launch the unified SSL feature.

wilkinsona:
And [Cassandra too](https://github.com/spring-projects/spring-boot/issues/25602).

philwebb:
@asaikali might have some sample code that shows common things that need to be configured for SSL

patpatpat123:
Hello team,

Many thanks for this great improvement.
Would it be possible to add ElasticSearch as well?

wilkinsona:
@patpatpat123 we can certainly consider it. Please open a separate issue so that the suggestion doesn't get lost in this one.

philwebb:
@patpatpat123 I've opened #35155

wilkinsona:
Closing this now in favor of the handful of remaining issues.

PR #20344, opened 2020-02-28
Make ActiveMQConnectionFactoryFactory public

dekelpilli:
As SSL ActiveMQ is not currently supported for autoconfiguration (https://github.com/spring-projects/spring-boot/pull/17373, https://github.com/spring-projects/spring-boot/issues/17589), this helps provide a low-cost stop-gap measure for autoconfiguring with SSL. This could be used in a way similar to [ActiveMQConnectionFactoryConfiguration's usage of ActiveMQConnectionFactoryFactory](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/activemq/ActiveMQConnectionFactoryConfiguration.java#L64):

```java

@Bean
public ActiveMQProperties properties() {
    return new ActiveMQProperties();
}

@Bean
@ConfigurationProperties(prefix = "javax.net.ssl") //standard ssl prefix that also fits ActiveMQSslConnectionFactory method naming, allowing easy SSL property injection
public ActiveMqSslConnectionFactory connectionFactory(ActiveMQProperties properties,
				ObjectProvider<ActiveMQConnectionFactoryCustomizer> factoryCustomizers) {
    new ActiveMQConnectionFactoryFactory(properties,
					factoryCustomizers.orderedStream().collect(Collectors.toList()))
							.createConnectionFactory(ActiveMQSslConnectionFactory.class);
}

```

Right now, with this class being package-private, wiring spring.active.x and spring.jms.x properties into an ActiveMQSslConnectionFactory is a bit of a hassle.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/activemq/ActiveMQConnectionFactoryFactory.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/activemq/ActiveMQConnectionFactoryFactory.java
@@ -33,7 +33,7 @@ import org.springframework.util.StringUtils;
  * @author Phillip Webb
  * @author Venil Noronha
  */
-class ActiveMQConnectionFactoryFactory {
+public class ActiveMQConnectionFactoryFactory {
 
 	private static final String DEFAULT_EMBEDDED_BROKER_URL = "vm://localhost?broker.persistent=false";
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the suggestion. I'm a little reluctant to increase the surface area of Spring Boot public API for something that should only be a stop-gap measure. It would commit us to longer-term maintenance and backwards compatibility that we'd rather avoid unless necessary. 

`ActiveMQConnectionFactoryFactory` isn't a particularly large class. You may be better copying that class into your application. Flagging for team attention to see what the rest of the core team thinks.

snicoll:
FTR, I am also not keen to increase the surface area of the public API, especially as we are going to revisit it to support SSL at some point. Having a non public API gives us the flexibility to change the signature any way we like.

wilkinsona:
We discussed this one today and agreed that we do not want to make ActiveMQConnectionFactoryFactory public. Thanks anyway for the suggestion. Our recommendation is to copy its code into your application if it is useful to you to do so.

PR #20346, opened 2020-02-28
Remove unnecessary lines

YooInKeun:
I think there are unnecessary lines.
Thank you!


snicoll:
Thanks for the PR. 

Those aren't empty lines, there is a `// ...` to mention that this class [has more content](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-custom) but we're not showing it for brevity.

YooInKeun:
Thank you for your kind review.

PR #20356, opened 2020-03-01
Include version and edition of neo4j database in health details

bmscomp:
Actually heath indicator details for **Neo4j** returns the number of nodes witch makes one database hit  

```javascript
  neo4j: {
  status: "UP",
  details: {
      nodes: 0
    }
  }
```

Comparing to other database vendors, I found more useful to show technical details (version and edition)  about the database rather then returning the number of nodes in the database

With this pull request the **Neo4j** heath details will be as follow 

```javascript
  neo4j: {
  status: "UP",
  details: {
      edition: "community",
      version: "4.0.0"
    }
  }
```

Notice that the new request will make 0 hits to the database, and I guess less time consuming then counting the total number of nodes

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/neo4j/Neo4jHealthIndicator.java
@@ -40,7 +40,8 @@ public class Neo4jHealthIndicator extends AbstractHealthIndicator {
 	/**
 	 * The Cypher statement used to verify Neo4j is up.
 	 */
-	static final String CYPHER = "match (n) return count(n) as nodes";
+	static final String CYPHER = "CALL dbms.components() YIELD versions, edition"
+			+ " UNWIND versions as version return version, edition";
 
 	private final SessionFactory sessionFactory;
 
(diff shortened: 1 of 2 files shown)


snicoll:
Thanks for the PR

> Notice that the new request will make 0 hits to the database, 

I think that can be a problem as making a hit to the database is validating the connection is operating properly. Not doing so may lead to an UP indicator while the access to the database (or the database itself) is down. 

paging @michael-simons for insights.

michael-simons:
That’s a valuable change. I actually don’t like the original query at all as it goes to the node store. 
The query here goes over the wire to the database, so the connection is verified. No issues on that side. 

I’m unsure however since when Neo4j has that stored procedure and OGM still supports 3.2. Please make also sure it works on embedded. 

For the alternate starter that uses bolt only and doesn’t bring in OGM, I’ll probably add the same information. Like it a lot. 

For comparison here’s what we do over there 
https://github.com/neo4j/neo4j-java-driver-spring-boot-starter/blob/master/neo4j-java-driver-spring-boot-autoconfigure/src/main/java/org/neo4j/driver/springframework/boot/actuate/Neo4jHealthIndicator.java#L99

bmscomp:
Yes the updated query works for version 3.2 and on embedded version of neo4j and of course 4.0.0

michael-simons:
@snicoll Big thumbs up from my side. 

Much less impact on the database while actually verifying the connection _and_ better info.

snicoll:
@bmscomp thanks a lot for the PR.

PR #20367, opened 2020-03-02
Polish DataSourceHealthIndicator

izeye:
This PR polishes `DataSourceHealthIndicator` by:

- Removing a workaround as it seems to be outdated based on its comment.
- Removing try-finally clauses as just switching the two seems to be sufficient.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/jdbc/DataSourceHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/jdbc/DataSourceHealthIndicator.java
@@ -105,24 +105,16 @@ public class DataSourceHealthIndicator extends AbstractHealthIndicator implement
 		builder.up().withDetail("database", getProduct());
 		String validationQuery = this.query;
 		if (StringUtils.hasText(validationQuery)) {
-			try {
-				// Avoid calling getObject as it breaks MySQL on Java 7
-				List<Object> results = this.jdbcTemplate.query(validationQuery, new SingleColumnRowMapper());
-				Object result = DataAccessUtils.requiredSingleResult(results);
-				builder.withDetail("result", result);
-			}
-			finally {
-				builder.withDetail("validationQuery", validationQuery);
-			}
+			builder.withDetail("validationQuery", validationQuery);
+			// Avoid calling getObject as it breaks MySQL on Java 7 and later
+			List<Object> results = this.jdbcTemplate.query(validationQuery, new SingleColumnRowMapper());
+			Object result = DataAccessUtils.requiredSingleResult(results);
+			builder.withDetail("result", result);
 		}
 		else {
-			try {
-				boolean valid = isConnectionValid();
-				builder.status((valid) ? Status.UP : Status.DOWN);
-			}
-			finally {
-				builder.withDetail("validationQuery", "isValid()");
-			}
+			builder.withDetail("validationQuery", "isValid()");
+			boolean valid = isConnectionValid();
+			builder.status((valid) ? Status.UP : Status.DOWN);
 		}
 	}
 


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/jdbc/DataSourceHealthIndicator.java
```diff
@@ -101,28 +94,18 @@ protected void doHealthCheck(Health.Builder builder) throws Exception {
 		}
 	}
 
-	private void doDataSourceHealthCheck(Health.Builder builder) throws Exception {
+	private void doDataSourceHealthCheck(Health.Builder builder) {
 		builder.up().withDetail("database", getProduct());
 		String validationQuery = this.query;
 		if (StringUtils.hasText(validationQuery)) {
-			try {
-				// Avoid calling getObject as it breaks MySQL on Java 7
```
wilkinsona: We should double-check that it's safe to remove this. The [original change](https://github.com/spring-projects/spring-boot/commit/333bc3e8425b0cd36bc93224e6fb2ab810ffe141) suggests that the problem exists with Java 7 and later.

izeye: Oops. I read it as Java 7 only, so this change is based on the faulty assumption.

izeye: I checked that `JdbcTemplate.queryForObject()` works with [this sample code](https://github.com/izeye/spring-boot-throwaway-branches/blob/d6100d76148e1c3aa3457a666cbce43693d37ff3/src/main/java/com/izeye/throwaway/service/JdbcTemplateTestService.java#L24-L27).

MySQL server version was 8.0.16 as follows:

```
$ mysqld --version
/usr/local/Cellar/mysql/8.0.16/bin/mysqld  Ver 8.0.16 for osx10.13 on x86_64 (Homebrew)
$
```
Java version was 1.8.0_231 as follows:

```
$ java -version
java version "1.8.0_231"
Java(TM) SE Runtime Environment (build 1.8.0_231-b11)
Java HotSpot(TM) 64-Bit Server VM (build 25.231-b11, mixed mode)
$
```

snicoll: @izeye thanks. While it seems that limitation has been fixed in a more recent driver, I am a bit nervous we would break users with an older mysql driver. The code is stable and used for quite some time now so I'd be tempted to keep it rather than breaking someone.

Changing the `protected` method signature is also fair for a new feature release but I don't see a lot of benefit in removing the `throws Exception` clause.

Thoughts?

izeye: @snicoll Thanks for the feedback! I don't have a strong opinion on those changes as they have been started based on my faulty assumption as I already mentioned above. So just let me know if you want me to revert them.

snicoll: Alright, let's revert that part then please. Thanks!

izeye: @snicoll Thanks for the quick feedback! I reverted them, and I also updated the comment to be explicit that it includes Java 8.

snicoll:
Thanks a lot for the follow-up @izeye, much appreciated!

Issue #16020, opened 2019-02-22
Review support for Spring HATEOAS and its new WebFlux integration


wilkinsona:
A non-exhaustive list of things to consider:

 - Auto-configuration ordering
 - `spring-boot-starter-hateoas` is currently MVC specific (it depends on `spring-boot-starter-web`)

mikrethor:
Sorry if it's not the right place but before opening an issue. I want ti be sure. Is hateoas compatible with router function or only with controllers?

philwebb:
@mikrethor I think it might depend on which parts of Spring HATEOAS you are trying to use. The best place to ask is probably stackoverflow. Or you could raise a documentation issue at https://github.com/spring-projects/spring-hateoas/issues if you think things need to be clarified.

gregturn:
@wilkinsona when I do this, I have managed to get things working with WebFlux...

```xml
<properties>
    <java.version>1.8</java.version>
    <spring-hateoas.version>1.0.0.BUILD-SNAPSHOT</spring-hateoas.version>
</properties>
```

```xml
<dependency>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-webflux</artifactId>
</dependency>

<dependency>
    <groupId>org.springframework.boot</groupId>
    <artifactId>spring-boot-starter-hateoas</artifactId>
    <exclusions>
        <exclusion>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-starter-web</artifactId>
        </exclusion>
    </exclusions>
</dependency>
```

```java
@SpringBootApplication
@EnableHypermediaSupport(type = { HAL, HAL_FORMS })
public class HackingSpringBootApplication {

	public static void main(String[] args) {
		SpringApplication.run(HackingSpringBootApplication.class, args);
	}
}
```

Presumably, because this takes Spring MVC off the classpath and puts WebFlux on the classpath, Spring HATEOAS itself is able to pick the right set of beans to register itself properly.

wilkinsona:
This is still blocked: https://github.com/spring-projects/spring-hateoas/issues/1061.

wilkinsona:
In addition to the blocking issue above, a secondary issue is that I don't think we can support the existing `spring.hateoas.use-hal-as-default-json-media-type` with WebFlux as we can't manipulate the JSON codecs in the same way as we currently manipulate the HTTP message converter with MVC.

wilkinsona:
[...]

PR #20373, opened 2020-03-02
Add hypermedia autoconfiguration for RestTemplate.

gregturn:
* Provide RestTemplateCustomizer.
* Verify RestTemplateCustomizer is also picked up for TestRestTemplate.

Related: #16020.


gregturn:
Per our video chat, we plan to:

* Let Spring HATEOAS provide a programmatic way to register hypermedia with `RestTemplate`. That way users can leverage it directly. See https://github.com/spring-projects/spring-hateoas/issues/1223 to track that development.
* For Spring Boot 2.3, no autoconfiguration.

PR #20411, opened 2020-03-06
Upgrade to Groovy 2.5.10

dreis2211:
Hi,

this PR upgrades to Groovy 2.5.10. I know this sort of upgrade is usually handled via bomr, but this is relevant for the JDK 14 setup which I'm currently working on.

In a separate PR I'd eventually use the new BOM if that's okay for you.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -367,7 +367,7 @@ bom {
 			]
 		}
 	}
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks Christoph. Using the BOM sounds good to me.

snicoll:
FTR, while we don't accept PRs for simple dependency upgrades like this one, I've decided to merge it as it is a prerequisite for another PR (#20413).

PR #20413, opened 2020-03-06
Add Java 14 CI

### LINKED ISSUE #20147: Add Java 14 CI

dreis2211:
Hi,

this PR should add a CI job for Java 14 and therefore closes #20147 . Again - as the CI topics are extremely hard to test for me locally, feel free to decline the PR or forgive me if things are not working as they were intended.

Eventually, when there is a Gradle version that works with JDK 14, we can clean things up again. For the time being, the spring-boot-gradle-plugin tests are excluded from the build.

In order to have a green build, the following two prior PRs need to be merged:
- #20180 
- #20411 

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -94,6 +94,14 @@ resources:
     username: ((docker-hub-username))
     password: ((docker-hub-password))
     tag: 2.3.x
+- name: spring-boot-jdk14-ci-image
+  type: docker-image
+  icon: docker
+  source:
+    repository: ((docker-hub-organization))/spring-boot-jdk14-ci-image
+    username: ((docker-hub-username))
+    password: ((docker-hub-password))
+    tag: 2.3.x
 - name: artifactory-repo
   type: artifactory-resource
   icon: package-variant
@@ -126,6 +134,14 @@ resources:
     access_token: ((github-ci-status-token))
     branch: ((branch))
     context: jdk13-build
+- name: repo-status-jdk14-build
+  type: github-status-resource
+  icon: eye-check-outline
+  source:
+    repository: ((github-repo-name))
+    access_token: ((github-ci-status-token))
+    branch: ((branch))
+    context: jdk14-build
 - name: slack-alert
   type: slack-notification
   icon: slack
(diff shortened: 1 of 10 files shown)


ci/scripts/build-project.sh
```diff
@@ -5,5 +5,14 @@ source $(dirname $0)/common.sh
 repository=$(pwd)/distribution-repository
 
 pushd git-repo > /dev/null
-./gradlew -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false --no-daemon --max-workers=4 -PdeploymentRepository=${repository} build publishAllPublicationsToDeploymentRepository
+if [[ -d /opt/openjdk-secondary ]]; then
```
dreis2211: I'm not super happy with this block here, but as said: this can be cleaned up when there is a Gradle version that works with JDK 14

ci/pipeline.yml
```diff
@@ -375,6 +404,44 @@ jobs:
         silent: true
         icon_emoji: ":concourse:"
         username: concourse-ci
+- name: jdk14-build
+  serial: true
+  public: true
+  plan:
+    - get: spring-boot-jdk14-ci-image
+    - get: git-repo
+      trigger: false
```
dreis2211: This needs to be set to true - I was unsure if I should do that from the start. Let me know if I should change it already.

dreis2211:
Updated the PR to use the first GA release of JDK 14

snicoll:
@dreis2211 we've decided to go with Gradle 6.3 RC2 so the secondary JDK is no longer necessary. I have a feeling though that this infrastructure can be useful going forward. I am going to revert that piece of the PR to create an extra pipeline that's similar to the others.

dreis2211:
@snicoll Makes sense

snicoll:
@dreis2211 I've just reverted the part of the PR that was using the infrastructure. I think that `testJavaHome` and `buildJavaHome` are generally useful and don't need to be reverted. Let's see what happens on CI now :) 

Let me know if something is wrong in the polish. Thanks again!

PR #20416, opened 2020-03-07
Upgrade to Testcontainers 1.13.0

dreis2211:
Hi,

this PR upgrades Testcontainers to version 1.13.0.

Cheers,
Christoph


snicoll:
Thanks again Christoph. 

@wilkinsona I wonder if we should apply the upgrade to `2.2.x` as well, given it is an internally-managed dependency (i.e. with regards to our upgrade policy in maintenance releases).

snicoll:
FTR, this dependency upgrade is a bit unusual as it is an internal dependency that has no public dependency management. As such, `bomr` does not cover it so we have to do that check manually.

PR #20417, opened 2020-03-07
Add support for configuring Liquibase's clearCheckSums

ferengra:
Liquibase auto configuration is extended with clearCheckSums to allow to clear all checksums in the current changelog, so they will be recalculated next update. SpringLiquibase supports it since 3.7.0.

The clearCheckSums is a pretty useful feature of Liquibase in the development phase. For instance if you just added a comment or some cosmetic changes to your changeset, Liquibase will break the startup of your application based on the changed checksum values. So if your changeset is already deployed to QA / UAT, this flag makes it possible to make those small changes, without a new changeset. Sometimes even a new changeset doesn't solve your problem, like in my project.
In my case I want to make the existing Liquibase scripts run on H2 database for some integration tests, but those scripts contain some Oracle default values (CREATE SEQUENCE with NOORDER) - which could be removed without any effect -, and unfortunately those script are already deployed to QA. So either this flag or clear the checksums in the table DATABASECHANGELOG on QA database, which is more difficult.

It'd be nice if you cherry-pick this change to spring-boot 2.2.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
@@ -64,6 +64,7 @@ import org.springframework.orm.jpa.LocalContainerEntityManagerFactoryBean;
  * @author Dominic Gunn
  * @author Dan Zheng
  * @author András Deák
+ * @author Ferenc Gratzer
  * @since 1.1.0
  */
 @Configuration(proxyBeanMethods = false)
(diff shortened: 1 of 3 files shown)


ferengra:
Could someone trigger a rebuild of this pull request?
I wonder how adding a flag to LiquibaseProperties class could result in a failure in ReactiveSessionAutoConfigurationMongoTests... Locally all tests are passed.

snicoll:
@ferengra don't worry about that, that was a glitch in a Spring Data snapshot that has nothing to do with your change.

wilkinsona:
Liquibase's property is, IMO, capitalised incorrectly. The property is named `clearCheckSums` but, given that it is checksums that are being cleared, I think it should be named `clearChecksums`. This is somewhat important in Boot as it effects the canonical form of the property that will be recommended by IDE auto-completion. `clearCheckSums` will result in `clear-check-sums` and `clearChecksums` will result in `clear-checksums`. I think the latter is preferable, but it would require us to diverge from Liquibase's capitalisation. Flagging for team attention to see what the rest of the team thinks.

snicoll:
+1  for diverging

wilkinsona:
@ferengra Thank you very much for making your first contribution to Spring Boot.

PR #20419, opened 2020-03-08
Polish

wycm:
Polish


spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/CommandRunner.java
```diff
@@ -171,7 +171,7 @@ public int runAndHandleErrors(String... args) {
 			ExitStatus result = run(argsWithoutDebugFlags);
 			// The caller will hang up if it gets a non-zero status
 			if (result != null && result.isHangup()) {
-				return (result.getCode() > 0) ? result.getCode() : 0;
+				return Math.max(result.getCode(), 0);
```
snicoll: I find the curent version more readable.

snicoll:
Thanks for the follow-up.

PR #20420, opened 2020-03-08
Upgrade CI to Docker 19.03.7

dreis2211:
Hi,

this upgrades the Docker version for the CI environment to 19.03.7.

Cheers,
Christoph

P.S.: I could offer to create a similar automation process that we use for detecting jdk or ubuntu image upgrades. Let me know if I should invest a bit of time for that.

### DIFF
--- a/ci/images/setup.sh
+++ b/ci/images/setup.sh
@@ -29,7 +29,7 @@ test -f /opt/openjdk/bin/javac
 ###########################################################
 
 cd /
-curl -L https://download.docker.com/linux/static/stable/x86_64/docker-19.03.5.tgz | tar zx
+curl -L https://download.docker.com/linux/static/stable/x86_64/docker-19.03.7.tgz | tar zx
 mv /docker/* /bin/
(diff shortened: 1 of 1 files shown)


snicoll:
@dreis2211 thanks for the PR and the offer. Upgrading docker feels less important to me than making sure CI works with an up-to-date base image (or JDK). I'd be tempted to do upgrades on a ad-hoc basis. I've flagged this for team attention to get more feedback from the team.

wilkinsona:
@dreis2211 We discussed this today and we like the idea of automating the Docker upgrades. If you have the time to look at it, we'd be grateful for a pull request. Thanks very much for the offer. Looking at the change required to update, we wonder if it might be possible for the automation to open a pull request that does the upgrade rather than opening an issue and the change being made manually.

dreis2211:
I'll take a look - the automatic pull request will be a bit harder, but let's see. I guess there's no time pressure on this one, so this might be something I'll work on in between.

PR #20423, opened 2020-03-08
Upgrade to Elasticsearch 7.6.1

dreis2211:
Hi,

the current builds seem to fail, probably because of an upgrade to Elasticsearch 7.6.1 in https://github.com/spring-projects/spring-data-elasticsearch/commit/9258cdde8438845f8a4730394873aeb4199b2979 . When upgrading Spring-Boot accordingly the failing tests are green again.

Cheers,
Christoph


snicoll:
Good catch Christoph, thanks again.

snicoll:
I've created #20424 to see what we can do about the version duplication in our test.

PR #20428, opened 2020-03-09
Remove hardcoded version from VersionOverridingElasticsearchContainer

### LINKED ISSUE #20424: Revisit VersionOverridingElasticsearchContainer
It overrides the version of ES we should use and the version hardcoded there can easily be out-of-sync with what our dependency management uses.

dreis2211:
Hi,

this PR fixes #20424 by using `org.elasticsearch.Version` instead of hardcoding the version.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/elasticsearch/VersionOverridingElasticsearchContainer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/elasticsearch/VersionOverridingElasticsearchContainer.java
@@ -15,6 +15,7 @@
  */
 package org.springframework.boot.autoconfigure.data.elasticsearch;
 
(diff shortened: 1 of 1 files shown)


scottfrederick:
At the time we did the upgrade in https://github.com/spring-projects/spring-boot/commit/d282eb619f62165a36fa91f419ec98b9034538d2, TestContainers was behind the latest version. This hard-coded version was the minimum required to make the tests work. If everything is back in sync now, it would be better to just remove `VersionOverridingElasticsearchContainer` altogether and use the default container again.

dreis2211:
@scottfrederick `ElasticsearchContainer` still uses 6.4.1, so it's not really in sync (maybe I got you wrong though). Nonetheless, it would work with the standard `ElasticsearchContainer`.

scottfrederick:
No, I think you're right. If `org.elasticsearch.Version` is coming from the Elasticsearch client lib that we're getting via dependency management (not from TestContainers as I first thought), then this seems like the safer change.

PR #20472, opened 2020-03-10
Remove redundant exclusions

dreis2211:
Hi,

I just noticed some dependency exclusions in `spring-boot-dependencies` that seem redundant. E.g. compare [selenium-remote-driver:3.14.0](https://mvnrepository.com/artifact/org.seleniumhq.selenium/selenium-remote-driver/3.14.0) with [selenium-remote-driver:3.141.0](https://mvnrepository.com/artifact/org.seleniumhq.selenium/selenium-remote-driver/3.141.0). On the latter, `commons-logging` is not a dependency anymore, while it is one in the former.

If I understood `CheckBom` correctly, it misses to check for non-wildcard excludes at the moment. This PR removes the redundant exclusions and handles non-wildcard ones inside `CheckBom` now.

Let me know what you think.
Cheers,
Christoph


wilkinsona:

PR #20530, opened 2020-03-15
Automate Docker updates in CI

dreis2211:
Hi,

as discussed in #20420 this is an attempt to automate Docker upgrades for the CI environment. I say attempt, because I can't really test it in the real setup, so forgive me if things are not working. If they are, it should open a PR for version 19.03.8.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -27,6 +27,14 @@ resources:
     username: ((github-username))
     password: ((github-password))
     branch: ((branch))
+- name: git-repo-ci-docker
+  type: git
+  icon: github-circle
+  source:
+    uri: ((github-repo))
+    username: ((github-username))
+    password: ((github-password))
+    branch: ci-docker-((branch))
 - name: git-repo-windows
   type: git
   source:
@@ -212,6 +220,7 @@ jobs:
 - name: detect-docker-updates
   plan:
   - get: git-repo
+    resource: git-repo-ci-docker
   - get: every-wednesday
     trigger: true
   - get: spring-boot-ci-image
@@ -223,7 +232,22 @@ jobs:
         GITHUB_ORGANIZATION: spring-projects
         GITHUB_PASSWORD: ((github-password))
         GITHUB_USERNAME: ((github-username))
+        ISSUE_TITLE: ((docker-upgrade-issue-title))
       image: spring-boot-ci-image
+    - put: git-repo-ci-docker
+      params:
+        repository: docker-updates-git-repo
+        force: true
+    - task: create-pull-request
+      file: git-repo/ci/tasks/create-pull-request.yml
+      params:
+        BASE_BRANCH: ((branch))
+        BRANCH: ci-docker-((branch))
+        GITHUB_REPO: spring-boot
+        GITHUB_ORGANIZATION: spring-projects
+        GITHUB_PASSWORD: ((github-password))
+        GITHUB_USERNAME: ((github-username))
+        ISSUE_TITLE: ((docker-upgrade-issue-title))
 - name: build
   serial: true
   public: true
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -209,6 +209,21 @@ jobs:
         GITHUB_PASSWORD: ((github-password))
         GITHUB_USERNAME: ((github-username))
       image: spring-boot-ci-image
+- name: detect-docker-updates
+  plan:
+  - get: git-repo
+  - get: every-wednesday
+    trigger: true
+  - get: spring-boot-ci-image
+  - do:
+    - task: detect-docker-updates
+      file: git-repo/ci/tasks/detect-docker-updates.yml
+      params:
+        GITHUB_REPO: spring-boot
+        GITHUB_ORGANIZATION: spring-projects
+        GITHUB_PASSWORD: ((github-password))
+        GITHUB_USERNAME: ((github-username))
+      image: spring-boot-ci-image
 - name: build
   serial: true
   public: true
@@ -595,6 +610,6 @@ groups:
 - name: "Release"
   jobs: ["stage-milestone", "stage-rc", "stage-release", "promote-milestone", "promote-rc", "promote-release", "sync-to-maven-central"]
 - name: "CI Images"
-  jobs: ["build-spring-boot-ci-images", "detect-jdk-updates", "detect-ubuntu-image-updates"]
+  jobs: ["build-spring-boot-ci-images", "detect-docker-updates", "detect-jdk-updates", "detect-ubuntu-image-updates"]
 - name: "Build Pull Requests"
   jobs: ["build-pull-requests"]
--- a/ci/scripts/detect-docker-updates.sh
+++ b/ci/scripts/detect-docker-updates.sh
@@ -19,6 +19,10 @@ existing_tasks=$( curl -s https://api.github.com/repos/${GITHUB_ORGANIZATION}/${
 existing_upgrade_issues=$( echo "$existing_tasks" | jq -c --arg TITLE "$ISSUE_TITLE" '.[] | select(.title==$TITLE)' )
 
 if [[ ${existing_upgrade_issues} = "" ]]; then
+	pushd git-repo > /dev/null
+	popd > /dev/null
+	git clone git-repo docker-updates-git-repo > /dev/null
+	pushd docker-updates-git-repo > /dev/null
 	# Create changes in dedicated branch
 	branch="ci-docker-$latest_version"
 	git config user.name "Spring Buildmaster" > /dev/null
--- a/ci/scripts/detect-docker-updates.sh
+++ b/ci/scripts/detect-docker-updates.sh
@@ -1,0 +1,46 @@
+#!/bin/bash
+
+latest_version=$(curl -I -s https://github.com/docker/docker-ce/releases/latest | grep "location:" | awk '{n=split($0, parts, "/"); print substr(parts[n],2);}' | awk '{$1=$1;print}' | tr -d '\r' | tr -d '\n' )
+
+if [[ $latest_version =~ (beta|rc) ]]; then
+	echo "Skip pre-release versions"
+	exit 0;
+fi
+
+latest="https://download.docker.com/linux/static/stable/x86_64/docker-$latest_version.tgz"
+current=$( git-repo/ci/images/get-docker-url.sh )
+
+if [[ $current = $latest ]]; then
+	echo "Already up-to-date"
+	exit 0;
+fi
+
+
+issue_title="Upgrade to Docker $latest_version in CI"
+
+existing_tasks=$( curl -s https://api.github.com/repos/${GITHUB_ORGANIZATION}/${GITHUB_REPO}/pulls\?labels\=type:%20task\&state\=open\&creator\=spring-buildmaster )
+existing_upgrade_issues=$( echo "$existing_tasks" | jq -c --arg TITLE "$issue_title" '.[] | select(.title==$TITLE)' )
+
+if [[ ${existing_upgrade_issues} = "" ]]; then
+	# Create changes in dedicated branch
+	branch="ci-docker-$latest_version"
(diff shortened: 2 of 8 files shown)


ci/scripts/detect-docker-updates.sh
```diff
@@ -0,0 +1,46 @@
...
+
+existing_tasks=$( curl -s https://api.github.com/repos/${GITHUB_ORGANIZATION}/${GITHUB_REPO}/pulls\?labels\=type:%20task\&state\=open\&creator\=spring-buildmaster )
+existing_upgrade_issues=$( echo "$existing_tasks" | jq -c --arg TITLE "$issue_title" '.[] | select(.title==$TITLE)' )
+
+if [[ ${existing_upgrade_issues} = "" ]]; then
+	# Create changes in dedicated branch
+	branch="ci-docker-$latest_version"
+	git config user.name "Spring Buildmaster" > /dev/null
+	git config user.email "buildmaster@springframework.org" > /dev/null
+	git checkout -b "$branch" origin/master > /dev/null
```
dreis2211: I'm not sure if this will work in the Concourse setup as I don't know if we have a real repository there.

ci/scripts/detect-docker-updates.sh
```diff
@@ -0,0 +1,46 @@
...
+	sed -i "s/version=.*/version=\"$latest_version\"/" git-repo/ci/images/get-docker-url.sh
+	git add git-repo/ci/images/get-docker-url.sh > /dev/null
+	git commit -m "$issue_title" > /dev/null
+	git push origin "$branch" > /dev/null
```
dreis2211: Do we maybe need to switch to the master branch again afterwards?

ci/scripts/detect-docker-updates.sh
mbhave: It will work. `stage.sh` does [something similar](https://github.com/spring-projects/spring-boot/blob/b66ca0016a8e08f0b86ea93d54112ce9d0fa50a5/ci/scripts/stage.sh#L35).

ci/scripts/detect-docker-updates.sh
mbhave: I don't think we can push from here as it will ask for credentials. We should probably commit inside the task and push using the `git-resource` in `pipeline.yml`. If there was no commit, it should be a no-op. Once the git-resource is used to push, we can trigger another task that creates the pull-request. This task config can then be shared between all the automatic detection jobs that we have (jdk, ubuntu and docker). It can take the branch name and issue title as params. What do you think @dreis2211?

ci/scripts/detect-docker-updates.sh
```diff
@@ -0,0 +1,46 @@
...
+
+issue_title="Upgrade to Docker $latest_version in CI"
+
+existing_tasks=$( curl -s https://api.github.com/repos/${GITHUB_ORGANIZATION}/${GITHUB_REPO}/pulls\?labels\=type:%20task\&state\=open\&creator\=spring-buildmaster )
+existing_upgrade_issues=$( echo "$existing_tasks" | jq -c --arg TITLE "$issue_title" '.[] | select(.title==$TITLE)' )
+
+if [[ ${existing_upgrade_issues} = "" ]]; then
+	# Create changes in dedicated branch
+	branch="ci-docker-$latest_version"
+	git config user.name "Spring Buildmaster" > /dev/null
```
mbhave: We'd need to `cd` into the `git-repo` before doing that.

ci/scripts/detect-docker-updates.sh
dreis2211: I'm not quite sure if I understand what you want to share exactly? Everything inside this if (so line 24-46) or just the curl command in line 36-43?
I must admit the Concourse setup you're asking for is not completely clear to me (although I generally understand the plan), but my Concourse experience is limited. I'd probably need a little bit of help with that, I guess.

mbhave:
Here's what I was thinking. `detect-docker-updates.sh` would look a lot like what you currently have except for line 33 (which tries to push to the git repo) and lines 36-43. For line 33, the push could be done in the `pipeline.yml`. We do something similar when [staging a release](https://github.com/spring-projects/spring-boot/blob/7a64b3f7619f3f91c9a75b6dc0c0aedd05760d1b/ci/pipeline.yml#L429). In this case, `stage-git-repo` is an [output](https://github.com/spring-projects/spring-boot/blob/7a64b3f7619f3f91c9a75b6dc0c0aedd05760d1b/ci/tasks/stage.yml#L6) of the `stage` task which has the [extra commit](https://github.com/spring-projects/spring-boot/blob/7a64b3f7619f3f91c9a75b6dc0c0aedd05760d1b/ci/scripts/stage.sh#L45).

Following the `put` to the git repo in the `pipeline.yml`, we could have another task which creates the PR. This task would call a script that contains the `curl` command in lines 36-43. The creation of the PR would be conditional on the branch differing from `master` (or `2.2.x`/`2.1.x`).

Here's what the `pipeline.yml` might look like:

```yaml
    - task: detect-docker-updates
      file: git-repo/ci/tasks/detect-docker-updates.yml
      params:
        BRANCH: ((branch)) [1]
        GITHUB_REPO: spring-boot
        GITHUB_ORGANIZATION: spring-projects
        GITHUB_PASSWORD: ((github-password))
        GITHUB_USERNAME: ((github-username))
      image: spring-boot-ci-image
    - put: git-repo-ci-docker [2]
      params:
        repository: docker-updates-git-repo
        force: true
    - task: create-pull-request
      file: git-repo/ci/tasks/create-pull-request.yml
      params:
        BRANCH: ((branch))
        GITHUB_REPO: spring-boot
        GITHUB_ORGANIZATION: spring-projects
        GITHUB_PASSWORD: ((github-password))
        GITHUB_USERNAME: ((github-username))
```
[1] `git-repo-ci-docker` would be another `resource` in `pipeline.yml` which points to the branch we want to push to.

```yaml
- name: git-repo-ci-docker
  type: git
  icon: github-circle
  source:
    uri: ((github-repo))
    username: ((github-username))
    password: ((github-password))
    branch: ci-docker-((branch))
```
[2] branch would be the branch configured in [1].

If this is still confusing and I can see how it can be because it is hard to test, I can prepare a branch with your PR and the changes I mentioned on top of that. Let me know what works best for you.

dreis2211:
Yeah, it's kind of hard if it can't be tested really. But I will give it a try - I have a way better idea now of what you were thinking. Will do so over the weekend, I guess. Thanks, @mbhave

ci/pipeline.yml
```diff
@@ -209,6 +217,37 @@ jobs:
         GITHUB_PASSWORD: ((github-password))
         GITHUB_USERNAME: ((github-username))
       image: spring-boot-ci-image
+- name: detect-docker-updates
+  plan:
+  - get: git-repo
+    resource: git-repo-ci-docker
```
dreis2211: I was wondering if that should point to the new resource.

ci/pipeline.yml
```diff
@@ -27,6 +27,14 @@ resources:
...
     password: ((github-password))
     branch: ((branch))
+- name: git-repo-ci-docker
+  type: git
+  icon: github-circle
+  source:
+    uri: ((github-repo))
+    username: ((github-username))
+    password: ((github-password))
+    branch: ci-docker-((branch))
```
dreis2211: Is this working if the branch doesn't exist?

ci/tasks/create-pull-request.yml
```diff
@@ -0,0 +1,14 @@
+---
+platform: linux
+inputs:
+  - name: git-repo
```
dreis2211: I'm wondering if this need to be a different input - e.g. the docker git resource in this case. Which makes it less generic somehow.

dreis2211:
I pushed an updated version, but the Concourse git-repo configs still confuse me. Their naming is simply not intuitive to me, sorry. I'm not happy with this PR - I can't really test the setup and it feels like doing open heart surgery with a blindfold on.

ci/pipeline.yml
```diff
@@ -209,6 +217,37 @@ jobs:
...
+      params:
+        GITHUB_REPO: spring-boot
+        GITHUB_ORGANIZATION: spring-projects
+        GITHUB_PASSWORD: ((github-password))
+        GITHUB_USERNAME: ((github-username))
+        ISSUE_TITLE: ((docker-upgrade-issue-title))
+      image: spring-boot-ci-image
+    - put: git-repo-ci-docker
+      params:
+        repository: docker-updates-git-repo
```
dreis2211: Where is the `docker-updates-git-repo` coming from? I was assuming it needs to be an output in the `detect-docker-updates.yml`?

mbhave:
@dreis2211 Thanks so much for updating this and sorry this is such a pain to test :( I will take a look at the updated PR today.

dreis2211:
No need to be sorry.

ci/pipeline.yml
mbhave: No, we need to create a long-running branch beforehand. I don't know if there's a way to create branches on the fly in concourse.

ci/pipeline.yml
mbhave: I don't think it should point to the new resource. We will only use the new resource to `put` to it since we want to `put` to that branch. Since the commit is based off `master`, we need to use the actual `git-repo` resource.

ci/tasks/create-pull-request.yml
mbhave: For this task, it only uses the `git-repo` to run the `create-pull-request.sh`. It's not actually used anywhere in the script itself so we can stick to using the regular `git-repo` and not the docker-specific one.

ci/pipeline.yml
mbhave: Yes, that's correct. You do have it as an output in `detect-docker-updates.yml` if I'm not wrong.

mbhave:
@dreis2211 I made a few minor changes to your PR [here](https://github.com/mbhave/spring-boot/commit/e229aa9f7ef8c4e8fdf94ef874710c2615cc99cb). I think this will work now. If you're okay with the changes, I can squash your commits, add mine on top and merge. We'll find out if it does work :)

dreis2211:
@mbhave I'm okay with whatever you suggest

dreis2211:
I guess it didn't work

mbhave:
It [did](https://github.com/spring-projects/spring-boot/pull/20754) with a few minor changes. The create pull request task was missing the image configuration and I had missed a `cd` when I added something in my polish commit.

mbhave:
Thanks for the PR @dreis2211!

dreis2211:
@mbhave The PR was created but it seems to be empty. Thoughts?

mbhave:
@dreis2211 I just realized that too. It had the right commit the first time I ran it. When I ran it again, it detected that a [PR already exists](https://github.com/spring-projects/spring-boot/blob/0b8b986ba0c2002fd7885c23894f92fad893a05b/ci/scripts/detect-docker-updates.sh#L34) and force pushed the existing `master` branch to the `ci-docker-image` branch. Effectively, the PR branch and master then became the same and GitHub thought the PR was merged.

I think if we remove this [`if` condition](https://github.com/spring-projects/spring-boot/blob/0b8b986ba0c2002fd7885c23894f92fad893a05b/ci/scripts/detect-docker-updates.sh#L23) and commit to `ci-docker-image` regardless of if a PR exists, it will keep the `ci-docker-image` branch up-to-date with `master` along with that additional commit on top.

We can then move that `if` condition](https://github.com/spring-projects/spring-boot/blob/0b8b986ba0c2002fd7885c23894f92fad893a05b/ci/scripts/detect-docker-updates.sh#L23) to the `create-pull-request.sh` script.

dreis2211:
Seems like a plan. Sorry, that you have additional work with this PR.

mbhave:
No problem! I hadn't anticipated this case. Anyway, I ran the pipeline twice and it [seems to work now](https://github.com/spring-projects/spring-boot/pull/20756) even when a PR exists. Thanks again!

PR #20537, opened 2020-03-17
Fix formatting in CustomLayersProvider

dreis2211:
Hi,

this PR fixes a tiny formatting violation in the current builds.

Cheers,
Christoph


mbhave:
Thanks @dreis2211 but I was working on [removing some duplication from that class](https://github.com/spring-projects/spring-boot/commit/fb8b531bfd98e22287bc31505f383d11a88c99b1) so I think the checkstyle should be fixed as part of that. Thanks anyway.

PR #20547, opened 2020-03-17
simplifly a little code

wanghenan:
I used idea inspect and found that some code can be simplified, so submit .


snicoll:
@wanghenan thanks for the PR. In the future, please create a PR from a branch rather than `master` so that you have a dedicated space to provide more changes if needed.

wanghenan:
I used idea inspect and found that some code can be simplified, so submit

PR #20559, opened 2020-03-18
Only test default Gradle version with JDK 14

dreis2211:
Hi,

since we're not excluding the `spring-boot-gradle-plugin` but rather use the RC of Gradle 6.3 we need another adjustment for the compatibility suite on JDK 14 already.

/@snicoll 

Cheers,
Christoph


dreis2211:
You were faster than me 👍

snicoll:
> You were faster than me 👍

Not really, that code was shelved while I was reviewing your PR and I forgot to add it before the merge 🙄

PR #20561, opened 2020-03-18
Upgrade to Gradle 6.3 RC3

dreis2211:
Hi,

this PR updates to Gradle 6.3 RC3. It seems that the upgrade to RC2 missed to update the gradle-wrapper.jar already.

Cheers,
Christoph


snicoll:
>  It seems that the upgrade to RC2 missed to update the gradle-wrapper.jar already.

I absolutely did, thank you very much for catching that Christoph.

PR #20566, opened 2020-03-18
Upgrade CI images to Ubuntu Bionic 20200311

dreis2211:
Hi,

this PR fixes #20565 .

Cheers,
Christoph

PR #20577, opened 2020-03-19
Only suggest Ubuntu versions available in the Docker registry

dreis2211:
Hi,

unfortunately the current Ubuntu upgrade task suggested to upgrade to a version that wasn't available in the Docker registry yet. See #20566 . This PR should fix that.

Sorry for the inconveniences this has caused.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

Issue #18506, opened 2019-10-03
Upgrade to OkHttp 4


larsgrefer:
I propose to update to OkHttp 4 for Spring Boot 2.2.x.

Despite being a new major version, it's binary and java-source compatible with OkHttp 3.
It even uses the Maven-Coordinates and the package name of OkHttp 3.

More information about this can be found here: https://square.github.io/okhttp/upgrading_to_okhttp_4/

wilkinsona:
Thanks for the suggestion. I think the switch to OkHttp4 and the introduction of a dependency on Kotlin is too big a change to make in 2.2, particularly after RC. Furthermore, Spring Framework 5.2 compiles and tests against 3.x. Thanks anyway.

membersound:
I cannot see the point here. I mean, the `kotlin` dependency is just pulled into the test scope, not into the packaged project. So why not upgrading the library to 4.x?

wilkinsona:
OkHttp is more than just the mock web server. It is also primarily used as an HTTP client in main application code. This means that moving to 4.x would result in Kotlin being pulled into the packaged project.

membersound:
Oh ok, that's not what I expected, because as Spring is having `RestTemplate` and `WebClient` already, I thought okhttp is used for testing purpose only...

scottfrederick:
`RestTemplate` and `WebClient` don't implement the HTTP transport layer directly, but delegate to external libraries like Apache HttpComponents, OkHttp, Netty, or simply `java.net.HttpURLConnection` to handle the transport layer. 

See [`ClientHttpRequestFactory`](https://docs.spring.io/spring/docs/current/javadoc-api/org/springframework/http/client/ClientHttpRequestFactory.html) in Spring Framework and it's usages in [`RestTemplateBuilder`](https://docs.spring.io/spring-boot/docs/current/api/org/springframework/boot/web/client/RestTemplateBuilder.html#detectRequestFactory-boolean-) in Spring Boot for more details.

odin568:
@wilkinsona The issue is quite old and it was clear to have that rejected to that time.
Are there any plans in the meantime as quite some time passed?

wilkinsona:
We don't have any plans to upgrade at this time. Spring Framework 5.3 continues to compile against 3.x and we continue to feel that version 4's Kotlin dependency is too large for Spring Boot to use that version by default.

larsgrefer:
[...]

PR #20592, opened 2020-03-20
Upgrade group("com.squareup.okhttp3") to 4.4.1 and start using bom.

kazuki-ma:
First: I don't know if there are any reason to avoid upgrade OkHttp to 4.x. Sorry if I'm sending PR about already rejected topic.

----

* Upgrade OkHttp to 4.4.1

According to https://square.github.io/okhttp/upgrading_to_okhttp_4/, 4.x has strict compatibility for 3.x. So upgrade to 4.x by dependency management is safer than typical major upgrade.

> Binary compatibility is the ability to compile a program against OkHttp 3.x, and then to run it against OkHttp 4.x. We’re using the excellent japicmp library via its Gradle plugin to enforce binary compatibility.
> https://square.github.io/okhttp/upgrading_to_okhttp_4/

* 2 test cases are updated due to OkHttp's internal field name is changed. (RestTemplateBuilderTests.java)

* Also InfluxDbProperties is changed due to issue in influxdb-java.
  * Under OkHttp3, influxdb-java sends "Authorization: encode(null:null"
  * Under new version of OkHttp, null username & password not leads illegal BasicAuth configuration but throws NullPointer exception.
  * Anyway, if user don't use auth, nothing affected I think.

This issue fixed in https://github.com/influxdata/influxdb-java/pull/644/files#diff-23360835991a654afd56314b41590521L150-R155, but not released yet.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/influx/InfluxDbProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/influx/InfluxDbProperties.java
@@ -36,12 +36,16 @@ public class InfluxDbProperties {
 	/**
 	 * Login user.
 	 */
-	private String user;
+	// null value leads invalid basic auth setup and RuntimeException until
+	// https://github.com/influxdata/influxdb-java/pull/644 released.
+	private String user = "";
 
 	/**
 	 * Login password.
 	 */
-	private String password;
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the PR, but this has already been [considered and rejected](https://github.com/spring-projects/spring-boot/issues/18506).

kazuki-ma:
Oh, sorry i don't find it. I understand your concern. I seek forward to upgrade it (and spring boot 3.x!)

PR #20604, opened 2020-03-22
Polish BindableTests

dreis2211:
Hi,

this PR removes two inner classes from `BindableTests` that seem to be unused plus a `System.out.println` that seems redundant.

Cheers,
Christoph


snicoll:
@dreis2211 good catch, thanks again!

PR #20608, opened 2020-03-23
"Building Docker Images" is incorrectly nested beneath "Kotlin support" in the reference documentation

izeye:
This PR fixes "Building Docker Images" section level.


wilkinsona:
Thanks very much, @izeye!

PR #20617, opened 2020-03-23
@FlywayDataSource does not work in an application that has multiple datasources 

### LINKED ISSUE #20598: @FlywayDataSource  does not work in an application that has multiple datasources
I upgraded spring-boot version from 2.1.2 to 2.2.2 in my project.
After upgrading, `FlywayAutoConfiguration` has been failed by `NoUniqueBeanDefinitionException`.

My application has multiple datasources and uses flyway whch uses the datasource is annotated  with`@FlywayDataSource`.
They are defined such as the following configuration class.

```java
@Configuration
public class DataSourceConfig {
    @Bean
    @ConfigurationProperties(prefix = "datasource.main")
    public DataSource mainDataSource() {
        return DataSourceBuilder.create().build();
    }

    @Bean
    @ConfigurationProperties(prefix = "datasource.secondary")
    public DataSource secondaryDataSource() {
        return DataSourceBuilder.create().build();
    }

    @Bean
    @FlywayDataSource
    public DataSource flywayDataSource(DataSource mainDataSource) {
        return mainDataSource;
    }
}

```

This problem may have been caused by the change of this commit (https://github.com/spring-projects/spring-boot/commit/de0c065625d370cf6e1224d62111a764523b814e).

In version 2.1.2, `getIfUnique()` was called in `FlywayAutoConfiguration.class` to get a unique datasource.  
https://github.com/spring-projects/spring-boot/blob/22e722d9f9f885d7fc67d51d16d68d55db6d6efc/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java#L136

But, in version 2.2.2 `getIfAvailable()` is called in `FlywayAutoConfiguration.class` to get a unique datasource.  

https://github.com/spring-projects/spring-boot/blob/de0c065625d370cf6e1224d62111a764523b814e/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java#L119

Is this change the migration mistake ?
as it is now,  `@FlywayDataSource` does not work well for multiple datasources.

ta7uw:
Resolves #20598

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
@@ -125,7 +125,7 @@ public class FlywayAutoConfiguration {
 				ObjectProvider<JavaMigration> javaMigrations, ObjectProvider<Callback> callbacks) {
 			FluentConfiguration configuration = new FluentConfiguration(resourceLoader.getClassLoader());
 			DataSource dataSourceToMigrate = configureDataSource(configuration, properties, dataSourceProperties,
-					flywayDataSource.getIfAvailable(), dataSource.getIfAvailable());
+					flywayDataSource.getIfAvailable(), dataSource.getIfUnique());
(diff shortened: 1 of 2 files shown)


snicoll:
@ta7uw thank you for making your first contribution to Spring Boot.

PR #20629, opened 2020-03-24
Spring Boot 2.2 Tomcat 8.5.53 compatibility

berngp:
Context
=======

We have some applications that would like to move to Spring Boot 2.2 but their clients depend on some of the features that are available in Tomcat 8, specifically the response phrase. Ideally they should be moving the clients to a state where they can upgrade to Tomcat 9. In the mean time we could also make some additional changes to the _Embedded Tomcat_ integration such that it doesn't fail if used with Tomcat 8.5.53.

Proposed Changes
================

The `JreCompat.isGraalAvailable` as well as `Registry.disableRegistry` methods were added in Tomcat 9. Their usage on TomcatEmbeddedWebappClassLoader, TomcatReactiveWebServerFactory, and TomcatServletWebServerFactory prevents users from pinning their tomcat dependencies to 8.5.53 and still be able to use Spring Boot 2.2.x.

This commit adds the TomcatCompatibilityUtil class that isolates those APIs and handles the case where the methods are not available. It also adds a Maven Profile that can be used to build and test spring-boot with Tomcat 8.5.53.

e.g.
```
./mvnw clean package -P tomcat85
```

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/pom.xml
+++ b/spring-boot-project/spring-boot-dependencies/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 <project xmlns="http://maven.apache.org/POM/4.0.0"
-	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
+		 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+		 xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
 	<modelVersion>4.0.0</modelVersion>
 	<parent>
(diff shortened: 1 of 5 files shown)


spencergibb:
I think the minimum of tomcat 9 happen in boot 2.1.x

berngp:
I was able to pin applications running 2.1.x to Tomcat 8.5.53 but the problem, and understandably so, is that 2.2 is accessing APIs only available in Tomcat 9.

wilkinsona:
Thanks for the proposal. As @spencergibb indicated above, Spring Boot has required Tomcat 9 since 2.1, as early as 2.1.0.M1 in fact. If 2.1 works with 8.5 that is by accident rather than by design. It should not be relied upon as we do not test with that version combination.

I'm afraid we've already considered supporting both Tomcat 8.5 and Tomcat 9 in parallel and decided not to do so as the additional maintenance burden was too great. That was several months ago during which time the reasons to do so have only decreased. Thanks anyway for the proposal.

berngp:
Thanks @wilkinsona and @spencergibb . I understand completely and I appreciate your quick reply to this matter.

PR #20630, opened 2020-03-24
Fix typos in packaging docs

dreis2211:
Hi,

7444306d0b2dfe89fe4f33ea197e0c670aba81d4 introduced a couple of typos. While fixing them I noticed that the docs were still referring to `spring-boot-layertools` instead `spring-boot-jarmode-layertools` so I fixed that as well while being on it.

Cheers,
Christoph


snicoll:
Well spotted, thank you very again @dreis2211.

PR #20662, opened 2020-03-24
Migrate off deprecated Spring Data Cassandra API

mp911de:
Configure `UserTypeResolver` and `CodecRegistry` on `MappingCassandraConverter`. Configure on `CassandraMappingContext` only the simple type holder instead of custom conversions.

Applies only to Spring Data Cassandra 3.0 (Neumann Release Train).

Related ticket: https://jira.spring.io/browse/DATACASS-470


snicoll:
This change breaks other tests, @mp911de is looking at the change.

snicoll:
Thanks for the follow-up @mp911de!

PR #20690, opened 2020-03-26
Favor exec-maven-plugin shorthand property over redefinition

fbiville:
With this change, users can now rely on the standard
`exec.mainClass` property when executing the exec-maven-plugin.
Users of the `start-class` property are unaffected.

Thanks @hboutemy for the initial diagnosis and suggestion for the
fix.


fbiville:
For more context, I am writing a helper for projectriff [streaming processor](https://github.com/projectriff/streaming-processor/). While the project uses Spring Boot, the helper is completely standalone and it took me a long time to understand that Spring Boot parent POM actually rewrites the `exec.mainClass` property, hence forcing me to comply.

While this issue is far from new, it seems it surprised a few more people, judging by the votes on [StackOverflow](https://stackoverflow.com/questions/27323104/spring-boot-and-maven-exec-plugin-issue/27324967).

snicoll:
Thank you for the additional context. We've discussed this one and decided to remove plugin management for the exec plugin altogether. I've created #20883

PR #20704, opened 2020-03-27
Fix typos

dreis2211:
Hi,

this PR fixes a couple of typos. I should note that the particular sentence in `spring-boot-features` still doesn't make a lot of sense to me even after the fix, but as I'm not a native speaker I wasn't sure if it might be correct.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/runner/AbstractApplicationContextRunner.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/runner/AbstractApplicationContextRunner.java
@@ -179,8 +179,8 @@ public abstract class AbstractApplicationContextRunner<SELF extends AbstractAppl
 	}
 
 	/**
-	 * Add a {@link ApplicationContextInitializer} to be
-	 * calWebMvcEndpointChildContextConfigurationTestsled when the context is created.
+	 * Add an {@link ApplicationContextInitializer} to be called when the context is
+	 * created.
 	 * @param initializer the initializer to add
 	 * @return a new instance with the updated initializers
 	 */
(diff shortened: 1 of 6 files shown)


snicoll:
> I should note that the particular sentence in spring-boot-features still doesn't make a lot of sense to me even after the fix

I wrote that and I agree the phrasing is a bit odd. I'll try to improve it as part of the polish.

spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/runner/AbstractApplicationContextRunner.java
```diff
@@ -179,8 +179,8 @@ public SELF withAllowBeanDefinitionOverriding(boolean allowBeanDefinitionOverrid
 	}
 
 	/**
-	 * Add a {@link ApplicationContextInitializer} to be
-	 * calWebMvcEndpointChildContextConfigurationTestsled when the context is created.
```
snicoll: oh dear 🙈

snicoll:
Thanks again @dreis2211. I've polished that sentence in https://github.com/spring-projects/spring-boot/commit/8154e672ddc91e07e8cbd77bc8511b263d94a4ce. Let me know if that still reads odd.

spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/runner/AbstractApplicationContextRunner.java
dreis2211: I liked that one 🤣

dreis2211:
Isn't that a different sentence? You polished the one about `DurationUnits` and not `DataSizeUnits`. Nonetheless, it reads better ;-)

snicoll:
It is. Looks like that pattern was applied for `DataSize` as well. I've polished the other one in https://github.com/spring-projects/spring-boot/commit/49ab5de91c8ee867ea87591a4aa0137e90f1d551

PR #20709, opened 2020-03-28
CassandraHealthIndicator runs a query that fails on some Consistency Levels

adutra:
The system keyspace has a replication factor of 1 and is local to each node; it is therefore recommended to query system.local with a consistency level of ONE or LOCAL_ONE.

Stronger consistency levels may result in an Unavailable error, but this does not mean that the node is down.

This patch does not include additional tests because it is hard to test the change in a stubbed environment; obviously, the existing tests still pass. Let me know if I can contribute an integration test for that.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
@@ -35,8 +35,8 @@ import org.springframework.util.Assert;
  */
 public class CassandraHealthIndicator extends AbstractHealthIndicator {
 
-	private static final SimpleStatement SELECT = SimpleStatement.newInstance("SELECT release_version FROM system.local")
-			.setConsistencyLevel(ConsistencyLevel.LOCAL_ONE);
+	private static final SimpleStatement SELECT = SimpleStatement
+			.newInstance("SELECT release_version FROM system.local").setConsistencyLevel(ConsistencyLevel.LOCAL_ONE);
 
 	private CassandraOperations cassandraOperations;
 
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.actuate.cassandra;
 
+import com.datastax.oss.driver.api.core.ConsistencyLevel;
 import com.datastax.oss.driver.api.core.cql.ResultSet;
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@adutra Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20709)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20709) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@adutra Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20709)!

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
```diff
@@ -52,8 +56,7 @@ public CassandraHealthIndicator(CassandraOperations cassandraOperations) {
 
 	@Override
 	protected void doHealthCheck(Health.Builder builder) throws Exception {
-		SimpleStatement select = SimpleStatement.newInstance("SELECT release_version FROM system.local");
-		ResultSet results = this.cassandraOperations.getCqlOperations().queryForResultSet(select);
+		ResultSet results = this.cassandraOperations.getCqlOperations().queryForResultSet(SELECT);
 		if (results.isFullyFetched()) {
```
adutra: I personally think that this check is inaccurate here since this method will always return true for this query. As a consequence, the code beneath the if block is never reached and the health indicator never reports the release version. But since this is not related to the fix I left it for another PR.

snicoll:
@adutra thank you very much for making your first contribution to Spring Boot.

We had to fix this one in `2.1.x` and `2.2.x` so I've adapted your contribution to the previous Cassandra API, see [this commit for 2.1.x](https://github.com/spring-projects/spring-boot/commit/1f2a6551b3f7df4e4f4d0511495311b21eb267bc). Along the way, I've harmonized that check that you've just  said is faulty so please let me know what it should be for the new and the previous API.

adutra:
@snicoll I would simply remove the check, since the query can either fail or return one row exactly. The reactive version of the check does that btw.

Speaking of reactive, `CassandraReactiveHealthIndicator` should be amended as well. I just pushed another commit to my branch but too late :-) Would you mind cherry-picking https://github.com/adutra/spring-boot/commit/2f0df4f00735260f35e58cee0c34975b6ffe841a ? Thanks!

snicoll:
> I would simply remove the check, since the query can either fail or return one row exactly.

Does that apply to the previous version of the SDK as well?

adutra:
> > I would simply remove the check, since the query can either fail or return one row exactly.
> 
> Does that apply to the previous version of the SDK as well?

Yes.

snicoll:
@adutra I've harmonized the reactive health indicator. I've also created an extra bug for the fact we never expose the version attribute (see #20179). Thanks for the feedback!

PR #20720, opened 2020-03-30
Fix for issue #10886 "Incremental compilation does not keep metadata…"

### LINKED ISSUE #10886: Incremental compilation does not keep metadata for inner classes
When the annotation processor runs incrementally, it doesn't keep the properties that are in a nested group (i.e. `NestedConfigurationProperty`). 

When working in the module where the `@ConfigurationProperties` object is defined, such properties became all the sudden invalid.

borisfox73:
Hi!

Proposed changes introduce additional metadata property "sourceCanonicalType" holding the canonical name of the source type. Existing property "sourceType" remains unchanged for backward compatibility.

Fixes #10886.

Regards,
Boris

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessor.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessor.java
@@ -195,7 +195,8 @@ public class ConfigurationMetadataAnnotationProcessor extends AbstractProcessor 
 
 	private void processAnnotatedTypeElement(String prefix, TypeElement element, Stack<TypeElement> seen) {
 		String type = this.metadataEnv.getTypeUtils().getQualifiedName(element);
-		this.metadataCollector.add(ItemMetadata.newGroup(prefix, type, type, null));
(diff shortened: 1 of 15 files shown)


borisfox73:
Keeping canonical source type name in the stored metadata is needed to determine whether a previous metadata item has been deleted in the current [incremental] build to exclude it from merging with the new metadata. The Elements API used in the MetadataCollector#deletedInCurrentBuild for this check does not provide a method to lookup type elements by their class names. Elements#getTypeElement accepts canonical name (one with components separated by dots), not a class name (separated by $ sign).
I couldn't (yet) find a way to get a type element from processing environment by it's class name.

Test scenario for nested configuration properties has been implemented.

borisfox73:
Fallen too behind, and introduces a breaking change anyway.

snicoll:
Sorry about that @borisfox73. I couldn't manage to figure out how to fix this one and, priority-wise, we've been quite busy on other topics. If that resume, I'll ping on the related issue. Thanks for your contribution, in any case

borisfox73:
Don't worry, it's ok :)
I have another fix we're successfully using awhile now, with simple flat 
to canonical class name conversion replacing dollar signs with dots. It 
passess the test scenario and seems to work flawlessly on our projects, 
but obviously won't work for properties class names containing dollar 
signs themselves (e.g. "MyTestProperties$").
I can share it as a new pull request.

Regards,
Boris

09.07.2020 18:52, Stéphane Nicoll пишет:

Issue #20144, opened 2020-02-12
Integrate Gradle's flaky test mitigation into our build


wilkinsona:
We have some tests in our build that are flaky. We should consider making use of [Gradle's flaky test mitigation](https://blog.gradle.org/gradle-flaky-test-retry-plugin) to help us to identify these tests. It would allow us to decide whether or not we want flaky tests to fail the build, while also making it easier for us to identify them via the test reports and build scans. We could continue to make code changes to main code where the flakiness is our fault and to test code where the flakiness is caused by third-party code and we need to work around it.

wilkinsona:
We're going to add the plugin and configure it to fail the build when there is a flaky test.

mikesmithson:
Can a PR be submitted for this issue?  If so, do you want this plugin configured at the root for all modules or module-specific ? My thinking is at the root build.gradle file, configure the plugin like so:
```
plugins {
  id "org.gradle.test-retry" version "1.1.3"
}
...
...
test {
    retry {
        maxRetries = 1
        maxFailures = 1
        failOnPassedAfterRetry = true
    }
}
```

wilkinsona:
Thanks for the offer, @mikesmithson. A pull request would be most welcome.

Rather than changing the root `build.gradle`, I think this should be handled in `buildSrc`. I'd imagined a dependency on `org.gradle:test-retry-gradle-plugin:1.1.3` add then some code in our `ConventionsPlugin` that reacts to the Java plugin being applied by also applying and configuring the test retry plugin.

mikesmithson:
@wilkinsona  - made the requested changes and pushed to my forked branch. Thanks...

PR #20722, opened 2020-03-30
#20144 - Gradle Flaky Test Mitigation Plugin 

mikesmithson:
added configuration for Gradle Test-Retry plugin to ConventionsPlugin.

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -2,6 +2,7 @@ plugins {
 	id "java-gradle-plugin"
 	id "io.spring.javaformat" version "${javaFormatVersion}"
 	id "checkstyle"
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks for your efforts here, @mikesmithson. Unfortunately, the latest updates do not get things working. The test retry plugin isn't being applied so they have no effect. When it is applied, a failure occurs as the extension is on the `Test` task not the project. I've made the necessary changes in https://github.com/spring-projects/spring-boot/commit/9bdf4a75e8a1bbada77bc8128e6dc2321e6222e0. Thanks anyway.

mikesmithson:
@wilkinsona - thanks, I gave it a shot. Was my first attempt at doing something like this.

Issue #20688, opened 2020-03-26
Clarify that `spring.datasource.hikari.*` properties are bound to a `HikariDataSource` instance


wilkinsona:
The documentation lists all of the third-party properties that can be bound, but simply described them as being "Hikari specific settings". We could help users to understand what the properties are bound to by replacing this with something like "Hikari-specific settings bound to an instance of `HikariDataSource`".

wilkinsona:
Closing in favour of #20727.

PR #20727, opened 2020-03-30
Clarify the type to which connection-pool-specific `spring.datasource` properties are bound

parameswaranvv:
For issue: #20688, Minor change to description of Hikari Datasource properties in documentation

Fixes gh-20688

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/context/properties/DocumentConfigurationProperties.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/context/properties/DocumentConfigurationProperties.java
@@ -81,8 +81,9 @@ public class DocumentConfigurationProperties extends AbstractTask {
 						"spring.jooq", "spring.jdbc", "spring.jpa", "spring.r2dbc")
 				.addOverride("spring.datasource.dbcp2", "Commons DBCP2 specific settings")
 				.addOverride("spring.datasource.tomcat", "Tomcat datasource specific settings")
-				.addOverride("spring.datasource.hikari", "Hikari specific settings").addSection("transaction")
-				.withKeyPrefixes("spring.jta", "spring.transaction").addSection("integration")
+				.addOverride("spring.datasource.hikari",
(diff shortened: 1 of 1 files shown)


buildSrc/src/main/java/org/springframework/boot/build/context/properties/DocumentConfigurationProperties.java
```diff
@@ -81,8 +81,9 @@ void documentConfigurationProperties() throws IOException {
 						"spring.jooq", "spring.jdbc", "spring.jpa", "spring.r2dbc")
 				.addOverride("spring.datasource.dbcp2", "Commons DBCP2 specific settings")
 				.addOverride("spring.datasource.tomcat", "Tomcat datasource specific settings")
-				.addOverride("spring.datasource.hikari", "Hikari specific settings").addSection("transaction")
-				.withKeyPrefixes("spring.jta", "spring.transaction").addSection("integration")
+				.addOverride("spring.datasource.hikari",
+						"Hikari specific settings bound to an instance of Hikari datasource")
```
wilkinsona: Rather than `Hikari datasource`, could you please change this to `` `HikariDataSource` `` (including the back-ticks) so that it's clearer that we're referring to the class named HikariDataSource?

parameswaranvv: Just made the change. Thanks for the suggestion.

wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @parameswaranvv. In reviewing the changes, I realised that they were applicable to Commons DBCP2 and Tomcat JDBC as well. I made similar updates for both of those in a [polishing commit](https://github.com/spring-projects/spring-boot/commit/6d7c3da1cf5bf592fe62cc1df77d748516c82012). I also had to drop the use of backticks around the class name as the column in the table was rendering them as-is. Apologies for the bad suggestion there.

PR #20743, opened 2020-03-31
Add warning about spring.config.additional-location

andersthorbeck:
It is unclear from the documentation that the `spring.config.location` and `spring.config.additional-location` properties are mutually exclusive. This change explicitly documents it.

The wording is subject to change, but I think the fact that these properties are mutually exclusive should be documented explicitly.

It took me about a day (before eventually delving into the spring code in `ConfigFileApplicationListener`) to realize I could not use these in tandem, I would like to avoid others having to make the same mistake.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
@@ -625,6 +625,8 @@ This search ordering lets you specify default values in one configuration file a
 You can provide default values for your application in `application.properties` (or whatever other basename you choose with `spring.config.name`) in one of the default locations.
 These default values can then be overridden at runtime with a different file located in one of the custom locations.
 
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@andersthorbeck Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20743)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20743) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@andersthorbeck Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20743)!

wilkinsona:
Thanks for the proposal. I'm not sure that it's intentional for those two properties to be mutually exclusive. I've opened #20745 so that we can take a step back and consider how we think things should behave and, if we decide that a change in behaviour is required, in what branch it should be changed.

PR #20747, opened 2020-03-31
use the applications rest template for the autoconfigured jwks receiver

m-kay:
If an application has a configured rest template (e.g. with a proxy configuration) the JwtDecoder auto configuration should use the applications rest template.


wilkinsona:
Thanks for the proposal but I'm not sure that this is a good idea. Using an application-wide `RestTemplate` for communicating with the resource server may result in, for example, credentials being sent to the resource server that are only intended for another service.

Instead, I think it would be better to provide a customiser callback interface that can be used to fine-tune the auto-configured `NimbusJwtDecoder`. I've opened https://github.com/spring-projects/spring-boot/issues/20750 so that we can take a step back and consider that approach. Thanks anyway.

m-kay:
Can I propose a solution for #20750 or does this issue need discussion first?

wilkinsona:
I noticed via an event on #20750 that you've [prototyped something](https://github.com/m-kay/spring-boot/commit/da8b69e8fd7f3246750c451296c2a8e0506261be). Thanks for doing that. We'll hopefully discuss #20750 later today so if you can hold fire for a little while we should know what direction we want to go in pretty soon.

PR #20793, opened 2020-04-01
Harmonize issue title for auto-created Docker upgrade issues

### LINKED ISSUE #20760: Harmonize issue title for auto-created Docker upgrade issues
The current title/commit is currently `Upgrade Docker version in CI`. It would be more consistent if we changed it to something like `Upgrade CI to Docker X.Y.Z` where `X.Y.Z` is the new version.

dreis2211:
Hi,

this PR closes #20760 as discussed. To fix this I needed to fix some related quirks around fetching already existing issues:

- The GitHub `/pulls` API is not having feature-parity with the `/issues` API when it comes to filtering and caused the API to return quite a lot of issues (e.g. all open ones instead of just open ones by the specific creator). I switched to the `/issues` API again + a corresponding `jq` selection filter for the `pull_request` key.
- Apparently there are some control characters in the response that cause the intermediate echo to fail on my local machine. I think I haven't noticed this before because it seems to be related to the milestone description (I'm guessing newlines) that wasn't assigned previously all the time. Inlining the `curl` and `jq` commands fixes this. Or at least makes it more stable.

Cheers,
Christoph

### DIFF
--- a/ci/parameters.yml
+++ b/ci/parameters.yml
@@ -1,4 +1,3 @@
-docker-upgrade-issue-title: "Upgrade Docker version in CI"
 email-server: "smtp.svc.pivotal.io"
 email-from: "ci@spring.io"
 email-to: ["spring-boot-dev@pivotal.io"]
--- a/ci/pipeline.yml
+++ b/ci/pipeline.yml
@@ -230,7 +230,6 @@ jobs:
       params:
         GITHUB_REPO: spring-boot
         GITHUB_ORGANIZATION: spring-projects
-        ISSUE_TITLE: ((docker-upgrade-issue-title))
       image: spring-boot-ci-image
     - put: git-repo-ci-docker
       params:
@@ -245,7 +244,6 @@ jobs:
         GITHUB_ORGANIZATION: spring-projects
         GITHUB_PASSWORD: ((github-password))
         GITHUB_USERNAME: ((github-username))
-        ISSUE_TITLE: ((docker-upgrade-issue-title))
       image: spring-boot-ci-image
 - name: build
   serial: true
--- a/ci/scripts/create-pull-request.sh
+++ b/ci/scripts/create-pull-request.sh
@@ -2,6 +2,7 @@
 set -e
 
 if [[ -f commit-details/message ]]; then
+	ISSUE_TITLE="$(cat commit-details/message)"
 	curl \
 	  -s \
 		-u ${GITHUB_USERNAME}:${GITHUB_PASSWORD} \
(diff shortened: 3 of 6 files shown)


mbhave:
Thanks @dreis2211. From the change it looks like we could have two PRs open at the same time? For example, if there is an outstanding PR titled `Upgrade CI to Docker 19.03.8` and the pipeline detects another docker version, it would create a PR titled `Upgrade CI to Docker 19.03.9`. Since the `ci-docker-image` branch is shared between the PRs, it would make the first PR invalid and there's no obvious way to indicate what invalidated that PR. I'm not sure what it should do in such a scenario.

dreis2211:
As the job only runs once per week that seems unlikely, but it's possible. It seems unlikely enough that you could simply manually close the older one in favor of the new one, if that really happens.

Apart from that, I could think of introducing a matching that doesn't do a full match against the title but only if the title starts with `Upgrade CI to Docker`. Simply let me know if I should add that.

mbhave:
Yeah, it does seem unlikely and closing the old PR manually doesn't seem so bad. The case that I'm more concerned about is if the latest-version is a pre-release. We won't find an existing issue with that title, let's say `Upgrade CI to Docker 19.03.9-rc`. We then [force push master](https://github.com/spring-projects/spring-boot/blob/4dad56a4911d0b0a50f86335db9311ab0e371edf/ci/scripts/detect-docker-updates.sh#L7) to the `ci-docker-master` branch, potentially affecting an open PR titled `Upgrade CI to Docker 19.03.8`.  Given that possibility, I think it might be better to look for an issue title that starts with `Upgrade CI to Docker` like you said.

dreis2211:
Pre-release versions are anyhow skipped already:
```
if [[ $latest_version =~ (beta|rc) ]]; then
	echo "Skip pre-release versions"
	exit 0;
fi
```

mbhave:
Yeah, but the [`put` to the `ci-docker-branch`](https://github.com/spring-projects/spring-boot/blob/4dad56a4911d0b0a50f86335db9311ab0e371edf/ci/pipeline.yml#L235) happens no matter what the `detect-docker-updates.sh` does. That's the reason we do [this](https://github.com/spring-projects/spring-boot/blob/4dad56a4911d0b0a50f86335db9311ab0e371edf/ci/scripts/detect-docker-updates.sh#L8-L11) so that if there is an existing PR, the branch remains unchanged.

dreis2211:
Right. In that case give me a couple of minutes to implement the new title matching...

dreis2211:
Pushed

mbhave:
Thansk again, @dreis2211.

PR #20805, opened 2020-04-02
Refactor ConventionsPlugin class

mikesmithson:
This class was extracted from several private methods in the ConventionsPlugin class. While trying to make a PR for an issue, I had some difficulty following the flow of this class and noticed that it had a lot of private methods that probably should be extracted to a separate class. I found several similar methods related to Maven pom conventions. There are probably a few more, but this one stood out as one obvious candidate.


buildSrc/src/main/java/org/springframework/boot/build/MavenPomConventions.java
```diff
@@ -0,0 +1,119 @@
...
+ * Maven Central's requirements.
+ * <li>If the {@link JavaPlugin Java plugin} has also been applied, creation of Javadoc
+ * and source jars is enabled.
+ * </ul>
+ *
+ * <p/>
+ *
+ * @author Mike Smithson
+ */
+public class MavenPomConventions implements Plugin<Project> {
```
wilkinsona, line 54: This class doesn't need to be a `Plugin`. It can also be package-private. The existing `AsciidoctorConventions` class sets a good precedent here.

buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
```diff
@@ -123,53 +116,81 @@ public void apply(Project project) {
 
 	private void applyJavaConventions(Project project) {
 		project.getPlugins().withType(JavaBasePlugin.class, (java) -> {
-			project.getPlugins().apply(TestFailuresPlugin.class);
+			applyTestFailuresPlugin(project);
```
wilkinsona, line 119: Given that applying the plugin is a single line and wrapping it in a method doesn't make the intent any clearer, I'd prefer to keep it inlined here.

buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
```diff
@@ -123,53 +116,81 @@ public void apply(Project project) {
...
 			configureSpringJavaFormat(project);
-			project.setProperty("sourceCompatibility", "1.8");
-			project.getTasks().withType(JavaCompile.class, (compile) -> {
-				compile.getOptions().setEncoding("UTF-8");
-				withOptionalBuildJavaHome(project, (javaHome) -> {
-					compile.getOptions().setFork(true);
-					compile.getOptions().getForkOptions().setJavaHome(new File(javaHome));
-					compile.getOptions().getForkOptions().setExecutable(javaHome + "/bin/javac");
+			setSourceCompatibility(project);
```
wilkinsona, line 121: I'd prefer to set the property directly here rather than wrapping it in a method call.

buildSrc/src/main/java/org/springframework/boot/build/MavenPomConventions.java
wilkinsona: This class applies more than pom conventions. I think `MavenPublishingConventions` would be a better name.

buildSrc/src/main/java/org/springframework/boot/build/ConventionsPlugin.java
```diff
@@ -111,6 +103,7 @@
  *
  * @author Andy Wilkinson
  * @author Christoph Dreis
+ * @author Mike Smithson
```
wilkinsona, line 106: The javadoc above should be updated to replace the list of actions when the `MavenPublishPlugin` is applied and to link to the new Maven conventions class instead.

mikesmithson:
@wilkinsona  - sure, I would be happy to accommodate your requests. I was thinking of going further but wanted to put out this proposal to you to see if there was any interest in pursuing it.

mikesmithson:
@wilkinsona - made another push to address your requested changes.

buildSrc/src/main/java/org/springframework/boot/build/DocumentConventions.java
```diff
@@ -0,0 +1,81 @@
...
+
+/**
+ * Class to support the generation of NOTICE.txt and LICENSE.txt files as a result of the
+ * {@link Jar} task being configured in {@link JavaConventions} class.
+ *
+ * @author Andy Wilkinson
+ * @author Christoph Dreis
+ * @author Mike Smithson
+ */
+class DocumentConventions {
```
wilkinsona, line 42: I'd prefer to keep these in the same class as the other Java-related conventions.

mikesmithson: OK, I inlined all of the methods in DocumentConventions into JavaConventions.

mikesmithson:
latest requested changes are in commit 13e7406.

wilkinsona:
Thanks very much, @mikesmithson. The proposed changes are now in master along with a [small polishing commit](https://github.com/spring-projects/spring-boot/commit/ba1d4ab458c6d3534840b92fcf784eafb53fe40a) to make things a little more consistent across the three `…Conventions` classes.

mikesmithson:
@wilkinsona  - thanks!

PR #20824, opened 2020-04-03
Update StringSequence

Konrado85:
Additional check of the "source" parameter. Based on the problem reported by SpotBugs 4.0.1. After adding a fix, any wrong calls of the first constructor will also be visible as a bug and can be caught earlier.


pivotal-issuemaster:
@Konrado85 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20824)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20824) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Konrado85 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20824)!

wilkinsona:
Thanks for the suggestion but I would prefer to keep a single `requireNonNull` check in the constructor to which the other one delegates.

PR #20832, opened 2020-04-04
Fix WebappLoader constructor deprecations

dreis2211:
Hi,

I just noticed two deprecation warnings coming from `WebappLoader(Classloader)` constructors. They should be removed in favor of using `Context.setParentClassLoader()` - which was already the case for those.

Cheers,
Christoph


snicoll:
Thanks @dreis2211. It's interesting to see that in both cases `Context#setParentClassLoader`  is actually using a different value. I don't know enough that area of the code to be sure that there isn't another problem. Flagging for team attention to see if someone else on the team has an opinion.

dreis2211:
The PR itself should be unaffected by this, but I guess it's good to clarify this in any case.

snicoll:
It depends whether the deprecated value is the one we were supposed to use in the end or not.

snicoll:
Sorry Christoph, @wilkinsona helped me realize that we're building the `WebAppLoader` with the  value we set previously in the context. I totally missed that 🤦‍♂️

dreis2211:
No worries. I was wondering yesterday if I had missed something.

PR #20847, opened 2020-04-05
Polish ConditionMessage#because()

dreis2211:
Hi,

I just noticed a tiny optimization opportunity in `ConditionMessage#because` to avoid unnecessary string concatenations when the `condition` is empty.

Cheers,
Christoph


snicoll:
Thanks again!

PR #20851, opened 2020-04-06
Change call to createClassLoader(URL) from 'super' to 'this'

olegz:
ExecutableArchiveLauncher.createClassLoader(Iterator) calls createClassLoader(URL) method with 'super' effectively removing any chance the later method could be overriden and successfully called without overriding both methods.


mbhave:
Thanks @olegz. This has now been merged into master.

PR #20857, opened 2020-04-06
Harmonize naming of Elasticsearch auto-configuration classes

vpavic:
This PR harmonizes the naming of Elasticsearch auto-configuration components by including _Elasticsearch_ in class names.

I've found the current naming to be very inconvenient to work with, since searching for Elasticsearch related auto-configuration components by entering _ElasticsearchConfiguration_ or _ElasticsearchProperties_ in IDEA's doesn't return some of vital components like `RestClientAutoConfiguration` and `RestClientProperties` - those two are especially harder to find this way since they're not in the same package as components that currently do have _Elasticsearch_ in their name (`org.springframework.boot.autoconfigure.elasticsearch.rest` vs `org.springframework.boot.autoconfigure.data.elasticsearch`).

A good example of well-named (and therefore easily discoverable) auto-configuration components is [`org.springframework.boot.autoconfigure.session`](https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session), where all components contain _Session_ in their names.

Additionally, one further point of harmonization (that I didn't apply to this PR yet) could be to move components from `org.springframework.boot.autoconfigure.elasticsearch.rest` to simply `org.springframework.boot.autoconfigure.elasticsearch` - as reactive client components in `org.springframework.boot.autoconfigure.data.elasticsearch` aren't nested in `rest` subpackage.


philwebb:
Dropping the `rest` package at the same time seems like a good idea, we can do that when we merge this one.

vpavic:
Thanks for the feedback - I've applied the drop of `rest` package, the branch had conflicts from other things that landed to the `master` in the meantime so I was to update it anyway.

Issue #20839, opened 2020-04-04
Migrate NewRelic metrics to NewRelicClientProvider


snicoll:
The `httpClient` is now deprecated so we need to migrate to the new API.

As a result, we should double check that the `connect-timeout` and `read-timeout` properties are still honoured. If they aren't we need to make sure we add metadata to flag them as error.

snicoll:
There is a `clientProviderType` that we may have to expose as well.

shakuzen:
@neiljpowell has work on this prepared, I believe.

neiljpowell:
Hi. Yes, I have a branch and can submit a PR for the new API later today.  @shakuzen we'll need to finalize moving the connect/readTimeout into NewRelicConfig or other, per the other Micrometer PR I submitted.

snicoll:
Superseded by #20908

PR #20871, opened 2020-04-07
Migrate NewRelic metrics to NewRelicClientProvider

neiljpowell:
Regarding https://github.com/spring-projects/spring-boot/issues/20839., this for the updates to support the new Micrometer NewRelicMeterRegistry changes introduced in 1.4.0, for the benefits noted in https://github.com/micrometer-metrics/micrometer/pull/1647.

The registry now uses NewRelicClientProvider implementations to support two New Relic Insights integration options, that result in the same metrics published. The first, is the current REST API approach and a second that delegates to the New Relic APM Java Agent if present and not explicitly configured to use the REST API.  If the Agent is not present, the API implementation will be implicitly configured.

@shakuzen


snicoll:
Thanks for the PR. I prefer that we stick to configuring what Micrometer provides and only apply additional customizations if necessary. If a `NewRelicClientProvider` bean is defined, we're going to use that which should cover those advanced use cases.

I've migrated our support in #20908

neiljpowell:
@snicoll @shakuzen Hi, can you clarify which customizations/options aren't acceptable? I'm happy to remove the added API client provider proxy config option. Or are you referring to the optional Agent provider all together?  The Agent client provider is part of the NewRelicMeterRegistry.

PR #20872, opened 2020-04-07
Change PropertiesLauncher equals method to use URI rather than URL

Konrado85:
We seem to have a little problem here:

"The equals and hashCode methods of URL are blocking"

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/PropertiesLauncher.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/PropertiesLauncher.java
@@ -515,7 +515,7 @@ public class PropertiesLauncher extends Launcher {
 	private List<Archive> getNestedArchives(String path) throws Exception {
 		Archive parent = this.parent;
 		String root = path;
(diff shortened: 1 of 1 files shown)


bclozel:
I guess you're referring to the DNS resolution issue with the URL class?
In this case, we should be dealing only with files on the file system - did you find a specific problem with this statement? Was that identified while profiling a particular application? Could you provide more information about this?

Note that the extra space in your PR is breaking the build.
Thanks!

Konrado85:
I came across this problem while analyzing my own application, in which I
use Spring Boot version 2.2.6

For the analysis I used, among others, a program:

SopotBugs version 4.0.1

https://spotbugs.github.io/

Besides the problems in my projection, SopotBugs showed some problems in
Spring itself. I check these remarks, what hasn't been reported yet I try
to report if I think the problem actually exists.

"The equals and hashCode methods of URL are blocking" appears twice in the
report and concerns two Springs classes.

Tomorrow I will try to propose an amendment to the second, but first I want
to look into it.


wt., 7 kwi 2020 o 18:25 Brian Clozel <notifications@github.com> napisał(a):

philwebb:
This has been raised previously in #10586

snicoll:
@Konrado85 thank you for making your first contribution to Spring Boot.

PR #20882, opened 2020-04-08
Polish JarFile to extra anonymous inner class

Konrado85:
When I'm checking my project with a tool:

SpotBugs version 4.0.1

https://spotbugs.github.io/

I found a little problem in this JarFile class in line 195

Problem description from SpotBugs tool:

Could be refactored into a named static inner class
This class is an inner class, but does not use its embedded reference to the object which created it.  This reference makes the instances of the class larger, and may keep the reference to the creator object alive longer than necessary.  If possible, the class should be made into a static inner class. Since anonymous inner classes cannot be marked as static, doing this will require refactoring the inner class so that it is a named inner class.
 
Bug kind and pattern: SIC - SIC_INNER_SHOULD_BE_STATIC_ANON

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/JarFile.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/JarFile.java
@@ -408,11 +408,11 @@ public class JarFile extends java.util.jar.JarFile {
 		DIRECT, NESTED_DIRECTORY, NESTED_JAR
 
 	}
-	
+
 	/**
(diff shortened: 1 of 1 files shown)


snicoll:
@Konrado85 thanks for the PR. We can polish the formatting as part of reviewing this PR. Going forward, please build the project locally before submitting the PR.

Konrado85:
Sorry for that, now I have only Browser. It Is last commit for formatting. I hope it's okay now.

wilkinsona:
I think the named inner class should be `private`. It should probably also be named differently. Its name reads like a specific type of `JarEntry` to me at the moment. We can take care of these as part of merging the proposal.

Issue #11574, opened 2018-01-08
Review all data properties


mbhave:
Redis properties are under `spring.redis` currently. This seems to be inconsistent with other stores that have properties under `spring.data` (eg, couchbase, cassandra, neo4j).

philwebb:
For java packages we have a split for some technologies. For example, the `org.springframework.boot.autoconfigure.mongo` package contains no Spring Data bits those are only in `org.springframework.boot.autoconfigure.data.mongo`.

snicoll:
Redis is a different beast in the sense that even the low-level connection stuff relies on Spring Data. In that sense the package is ok (since there _is_ a dependency on Spring Data.

There is no reason however to bring the `data` part in the namespace since those keys aren't related to the Spring Data part. 

I am not saying the situation is perfectly fine (I am sure we can find inconsistencies easily) but this one looks ok to me.

philwebb:
There's a lot of inconsistency with all our data properties. We should expand the scope of this to look at them all.

snicoll:
Note that the package structure also affects the actuator. We don't have any `data` package at all there and a number of support classes are effectively Spring Data specific. The current [health indicator for Cassandra](https://github.com/spring-projects/spring-boot/blob/a6efe77a0c19f935af83abe18cd5171997b881a7/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java#L41) is a good example.

PR #20887, opened 2020-04-08
Add Cassandra health indicator that uses CqlSession

adutra:
Motivation:

It is possible to use CassandraAutoConfiguration to create CqlSession
beans in an application with just a dependency on the DataStax Java driver,
without the requirement to also depend on Spring Data Cassandra.

However, it is currently not possible to do so if the application also
needs to create Actuator components. Indeed, both CassandraHealthIndicator
and CassandraReactiveHealthContributorAutoConfiguration have a hard
dependency on Spring Data Cassandra and its CassandraOperations bean.

Modifications:

This commit removes the hard dependency on Spring Data Cassandra from
CassandraHealthIndicator and CassandraReactiveHealthContributorAutoConfiguration,
and replaces all usages of CassandraOperations with direct access to the
underlying CqlSession bean.

Result:

Users are now able to create Actuator health components in their applications,
even if they use just Spring Boot and the DataStax Java driver, without
Spring Data Cassandra.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
@@ -32,6 +32,7 @@ dependencies {
 
 	optional(platform(project(":spring-boot-project:spring-boot-dependencies")))
 	optional("ch.qos.logback:logback-classic")
+	optional("com.datastax.oss:java-driver-core")
 	optional("com.fasterxml.jackson.dataformat:jackson-dataformat-xml")
 	optional("com.github.ben-manes.caffeine:caffeine")
 	optional("com.hazelcast:hazelcast")
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cassandra/CassandraDriverHealthContributorAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cassandra/CassandraDriverHealthContributorAutoConfiguration.java
@@ -1,0 +1,59 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.autoconfigure.cassandra;
+
+import java.util.Map;
+
+import com.datastax.oss.driver.api.core.CqlSession;
+
+import org.springframework.boot.actuate.autoconfigure.health.CompositeHealthContributorConfiguration;
+import org.springframework.boot.actuate.autoconfigure.health.ConditionalOnEnabledHealthIndicator;
+import org.springframework.boot.actuate.cassandra.CassandraDriverHealthIndicator;
+import org.springframework.boot.actuate.health.HealthContributor;
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.cassandra.CassandraAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for
+ * {@link CassandraDriverHealthIndicator}.
+ *
+ * @author Alexandre Dutra
+ * @since 2.4.0
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnClass(CqlSession.class)
+@ConditionalOnBean(CqlSession.class)
+@ConditionalOnEnabledHealthIndicator("cassandra")
+@AutoConfigureAfter({ CassandraAutoConfiguration.class, CassandraReactiveHealthContributorAutoConfiguration.class,
+		CassandraHealthContributorAutoConfiguration.class,
+		CassandraDriverReactiveHealthContributorAutoConfiguration.class })
+public class CassandraDriverHealthContributorAutoConfiguration
+		extends CompositeHealthContributorConfiguration<CassandraDriverHealthIndicator, CqlSession> {
+
+	@Bean
+	@ConditionalOnMissingBean(name = { "cassandraHealthIndicator", "cassandraHealthContributor" })
+	public HealthContributor cassandraHealthContributor(Map<String, CqlSession> sessions) {
+		return createContributor(sessions);
+	}
+
+}
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cassandra/CassandraDriverReactiveHealthContributorAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/cassandra/CassandraDriverReactiveHealthContributorAutoConfiguration.java
@@ -1,0 +1,58 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.boot.actuate.autoconfigure.cassandra;
+
+import java.util.Map;
+
+import com.datastax.oss.driver.api.core.CqlSession;
+import reactor.core.publisher.Flux;
+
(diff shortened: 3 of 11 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
```diff
@@ -38,25 +39,24 @@
...
+		this.session = session;
 	}
 
 	@Override
 	protected void doHealthCheck(Health.Builder builder) throws Exception {
-		String version = this.cassandraOperations.getCqlOperations().queryForObject(SELECT, String.class);
+		Row row = this.session.execute(SELECT).one();
+		Assert.notNull(row, "system.local should always return one row");
+		String version = row.getString(0);
+		Assert.notNull(version, "release_version should never be null");
```
adutra, line 59: These assertions really will never fail as any C* node would report exactly one row with one non-null column for such a CQL query. But they make static analysis tools happy.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicator.java
```diff
@@ -15,43 +15,47 @@
...
+		Assert.notNull(session, "session must not be null");
+		this.session = session;
 	}
 
 	@Override
 	protected Mono<Health> doHealthCheck(Health.Builder builder) {
-		return this.reactiveCassandraOperations.getReactiveCqlOperations().queryForObject(SELECT, String.class)
-				.map((version) -> builder.up().withDetail("version", version).build()).single();
+		return Mono.from(this.session.executeReactive(SELECT))
+				.map((row) -> Objects.requireNonNull(row.getString(0), "release_version should never be null"))
```
adutra, line 57: Same here, `Objects.requireNonNull` is just making my IDE happy.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
```diff
@@ -38,25 +39,24 @@
 	private static final SimpleStatement SELECT = SimpleStatement
 			.newInstance("SELECT release_version FROM system.local").setConsistencyLevel(ConsistencyLevel.LOCAL_ONE);
 
-	private CassandraOperations cassandraOperations;
-
-	public CassandraHealthIndicator() {
```
adutra, line 43: Removed because it doesn't buy us anything, but this is a breaking change. Let me know if you prefer to keep it.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
snicoll: There's no need to do that, you could check if the row is non null (since the API forces  you to do that check), something like:

```java
Row row = this.session.execute(SELECT).one();
builder.up();
if (row != null) {
	builder.withDetail("version", row.getString(0));
}
```

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicator.java
snicoll: IntelliJ IDEA? I think that's a false positive. 

```java
return Mono.from(this.session.executeReactive(SELECT))
		.map((row) -> builder.up().withDetail("version", row.getString(0)).build());
```

this is more concise and doesn't yield any warning for me.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicatorTests.java
```diff
@@ -47,25 +48,28 @@ void createWhenCassandraOperationsIsNullShouldThrowException() {
...
 	}
 
 	@Test
 	void healthWithCassandraDown() {
-		CassandraOperations cassandraOperations = mock(CassandraOperations.class);
-		given(cassandraOperations.getCqlOperations()).willThrow(new CassandraInternalException("Connection failed"));
-		CassandraHealthIndicator healthIndicator = new CassandraHealthIndicator(cassandraOperations);
+		CqlSession session = mock(CqlSession.class);
+		given(session.execute(any(SimpleStatement.class)))
+				.willThrow(new DriverTimeoutException("Connection timed out (not really)"));
```
snicoll, line 67: Please change that to `Test Exception` if you want to make it clear it's a test exception.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicatorTests.java
```diff
@@ -59,19 +65,38 @@ void testCassandraIsUp() {
...
 		StepVerifier.create(health).consumeNextWith((h) -> {
 			assertThat(h.getStatus()).isEqualTo(Status.DOWN);
 			assertThat(h.getDetails()).containsOnlyKeys("error");
 			assertThat(h.getDetails().get("error"))
-					.isEqualTo(CassandraInternalException.class.getName() + ": Connection failed");
+					.isEqualTo(DriverTimeoutException.class.getName() + ": Connection timed out (not really)");
 		}).verifyComplete();
 	}
 
+	private Answer<Void> mockReactiveResultSetBehavior(ReactiveRow row) {
```
snicoll, line 83: Meh. I am not keen to start mocking a `Publisher` here. The fact that the return type is a `Publisher` + something else makes it very hard to use existing reactor types. WIthout that I could have replaced that with `Mono.just`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/cassandra/CassandraReactiveRepositoriesAutoConfiguration.java
```diff
@@ -39,7 +41,7 @@
  * @see EnableReactiveCassandraRepositories
  */
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnClass({ ReactiveSession.class, ReactiveCassandraRepository.class })
+@ConditionalOnClass({ ReactiveSession.class, ReactiveCassandraRepository.class, Flux.class })
```
snicoll, line 44: I am not against that change but it has nothing to do with the task at hand. Please revert that and submit it in a separate change if you think this is worth exploring.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
adutra: Isn't that slightly different? In the very unlikely case where the row would be null, in my version the  health indicator would be set to DOWN, whereas in your case, it would be UP. But I will do as you suggest, since a null row is totally unrealistic anyways.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicator.java
adutra: Actually `GettableByIndex#getString(int)` is marked `@Nullable` in the driver API, so it may trigger a warning depending on how you configure things. But again I don't mind really, I'll do as you suggest.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicatorTests.java
adutra: Yeah, I don't see how to simplify this either. I'll take a note to provide a dummy implementation of `ReactiveResultSet` for test use cases like this one, in a future driver release.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/cassandra/CassandraReactiveRepositoriesAutoConfiguration.java
adutra: True, this is a mistake, I will remove.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
snicoll: Yes, you are right, I overlooked that. Although the "These assertions really will never fail" makes me think it won't make a difference.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicator.java
snicoll: I think that's going a tad too far. The `row` will never be `null` given the contract of `Publisher` and `onNext`.

snicoll:
We don't want to break backward compatible for this one. This should be possible by keeping the existing health indicator, adding this one and updating the auto-configuration to try this one and then the existing one if none apply. Unfortunately, the existing health indicator is in the `cassandra` package so we can't really have an additional one without having to use a name we wouldn't like. We've discovered brainstorming on this one that #11574 could be extended to package space as well as the actuator has no notion of data vs. raw driver.

It is unfortunately too late to make a breaking change in `2.3.x` so I've scheduled this for the next feature release.

adutra:
I understand your concerns. I will rework this PR in the way you suggested, and should have something ready shortly. If you don't mind, since we are back to square zero anyway, I will rebase on master and squash the commits.

adutra:
What do you suggest for class names?

I'm currently planning on declaring all health indicators under `org.springframework.boot.actuate.cassandra` and naming the new ones:

* `CassandraDriverHealthIndicator`
* `CassandraDriverReactiveHealthIndicator`

Is that ok?

philwebb:
@adutra Those names sound sensible to me. We can always refine them if we think of something better when we merge.

snicoll:
Thanks @adutra. I've merged your contribution with [a polish commit](https://github.com/spring-projects/spring-boot/commit/35e069e2cd4b563590ffd5869f78e33324a43961).

snicoll:
For the record, we've decided to deprecate the health indicators that rely on Spring Data (see #23226).

PR #20912, opened 2020-04-10
Add Cookie's SameSite directive property

### LINKED ISSUE #15047: Configure SameSite attribute on session Cookies with Spring Session
Currently, there's no way from `application.properties` to configure the Spring Session session cookie's `SameSite` attribute. It would be nice to be able to do that.

For consistency with the existing `server.servlet.session.cookie` properties, I suggest:
`server.servlet.session.cookie.sameSite` with a default value of "Lax" (to match Spring Session 2.1's behavior defined in [DefaultCookieSerializer](https://github.com/spring-projects/spring-session/blob/2.1.1.RELEASE/spring-session-core/src/main/java/org/springframework/session/web/http/DefaultCookieSerializer.java)). A value of empty string would map to `null` (which results in `DefaultCookieSerializer` not setting the `SameSite` attribute on the cookie).

[SessionAutoConfiguration](https://github.com/spring-projects/spring-boot/blob/v2.1.0.RELEASE/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java) would implement this behavior.

This setting would have no effect when Spring Session is not in use as no servlet containers currently expose a means by which to set the `SameSite` attribute on their session cookies (support for that can be added as containers gain that ability).

Note that this is likely to be increasingly used as the default session cookie in Spring Session 2.1 has the attribute `SameSite=Lax` (see https://github.com/spring-projects/spring-session/issues/1005) which breaks SAML login, so anyone using SAML (such as via Spring Security SAML) is going to have to need to change this configuration:

https://groups.google.com/a/chromium.org/d/msg/security-dev/AxY6BpkkH9U/vgKbDm7rFgAJ

As a further enhancement, perhaps if Spring Security SAML is detected, `server.servlet.session` can be set to null by default instead of "Lax".

cleankod:
See #15047

Added the ability to configure the `DefaultCookieSerializer`'s `SameSite` directive with an application property.

Fixes #15047

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
@@ -103,6 +103,7 @@ public class SessionAutoConfiguration {
 			map.from(cookie::getHttpOnly).to(cookieSerializer::setUseHttpOnlyCookie);
 			map.from(cookie::getSecure).to(cookieSerializer::setUseSecureCookie);
 			map.from(cookie::getMaxAge).to((maxAge) -> cookieSerializer.setCookieMaxAge((int) maxAge.getSeconds()));
+			map.from(cookie::getSameSite).to((sameSite) -> cookieSerializer.setSameSite(sameSite.toString()));
 			if (ClassUtils.isPresent(REMEMBER_ME_SERVICES_CLASS, getClass().getClassLoader())) {
 				new RememberMeServicesCookieSerializerCustomizer().apply(cookieSerializer);
 			}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -113,6 +113,10 @@
       "description": "Whether to always mark the session cookie as secure."
     },
     {
+      "name": "server.servlet.session.cookie.same-site",
+      "description": "The \"SameSite\" directive for the cookie."
+    },
+    {
       "name": "server.servlet.session.persistent",
       "description": "Whether to persist session data between restarts.",
       "defaultValue": false
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -114,6 +114,7 @@
     },
     {
       "name": "server.servlet.session.cookie.same-site",
+      "type": "org.springframework.boot.web.servlet.server.Session$Cookie$SameSite",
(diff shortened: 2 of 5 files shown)


pivotal-issuemaster:
@cleankod Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20912)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20912) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@cleankod Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=20912)!

bclozel:
Thanks for your contribution @cleankod  - did you see @vpavic commenting on the related issue? It seems this change goes against the general opinion. See [this comment](https://github.com/spring-projects/spring-boot/issues/15047#issuecomment-435454462).

Do you have new elements supporting this in the meantime?

cleankod:
@bclozel, @vpavic 
Since Google Chrome introduced the requirement for this attribute to be present, I really need to configure the application to include it in the Cookie. I have like 10 spring-boot-based web applications exposed to the browser. 

I made this PR because when the time comes and the Servlet API releases the attribute as well, it would be completely transparent to me and all I'd need to do is upgrade Spring Boot.

Now, without the possibility to set this via property, it leaves me with the necessity to provide a `DefaultCookieSerializer` as a custom `@Bean` initialisation. So when the time comes with the Servlet API release, I'd need not only to upgrade Spring Boot, but also to remove all of the custom  `@Bean`s, checking of course that no one added anything to them after they were introduced.

I completely understand your reason, but in such cases I would consider leaning towards Spring Boot users since the update of Servlet API is surely coming because this directive became an official one and can be found in [the specification](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie#SameSite). Waiting for Servlet API requires us to walk our way around this problem.

I'd therefore like to ask you to reconsider. If you stand by your decision anyway, then please consider merging my PR after the prerequisite Servlet API is released, so my work doesn't go to waste.

vpavic:
I don't think it's reasonable to expect Spring Boot to introduce a property under a Servlet API specific configuration property namespace for a thing that currently isn't by Servlet API and that requires an additional library (Spring Session) to be able to use it.

First and foremost, this is wrong as it's not usable if one's using Servlet container's default session management capabilities. Additionally, to my knowledge at least, the Servlet API support isn't imminent by any means and won't happen before `5.1`.

I've proposed an alternative approach in #20961. This would add the capability to customize the SameSite setting of Spring Session's `DefaultCookieSerializer` by simply registering a bean like:

```java
@Bean
CookieSerializerCustomizer cookieSerializerCustomizer() {
	return cookieSerializer -> cookieSerializer.setSameSite("None");
}
```

adamklinkosz:
@vpavic Could you please consider changing the target branch of your #20961 to `2.1.x`?

snicoll:
@adamklinkosz we don't need that and we will make up our mind on where to backport this in due course (@vpavic please don't change the base branch).

bclozel:
I'm closing this PR in favor of #20961, as it is less tied to the Servlet spec - thanks @cleankod for your contribution!

Issue #20828, opened 2020-04-03
Default for spring.kafka.listener.missing-topics-fatal is no longer aligned with Spring Kafka's default


garyrussell:
(Also Boot 2.2).

We changed the default in spring-kafka to `false` in 2.3.4.

This property was added in S-K 2.2 with default `true` but this caused too many problems for the community.

https://docs.spring.io/spring-kafka/docs/2.4.5.RELEASE/reference/html/#configuration-changes

dhirenmathur:
Hi, I'm looking to start contributing to OSS and this looks like a good place to start, is this up for grabs?

snicoll:
Yes. We’d need a test to assert the default is consistent. Thanks for the offer!

dhirenmathur:
Thanks @snicoll , I'll start working on it. 👍

snicoll:
Closing in favour of PR #20917

PR #20917, opened 2020-04-11
Default for spring.kafka.listener.missing-topics-fatal is no longer aligned with Spring Kafka's default

dhirenmathur:
Related to: 
[ISSUE -20828](https://github.com/spring-projects/spring-boot/issues/20828)

Issue Explanation:
[Spring-Kafka configuration change](https://docs.spring.io/spring-kafka/docs/2.4.5.RELEASE/reference/html/#configuration-changes) - Starting with version 2.3.4, the missingTopicsFatal container property is false by default. (Previously) When this is true, the application fails to start if the broker is down; many users were affected by this change; 
After this S-K config change, default for spring.kafka.listener.missing-topics-fatal is no longer aligned with Spring Kafka's default. 

AIM: 
Update the `missingTopicsFatal` default value to be consistent with above referenced Spring-Kafka change

This commit contains:
* In `KafkaProperties.java` update `missingTopicsFatal` default value to `false`.
* Unit test to assert that new default is consistent

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
@@ -52,6 +52,7 @@ import org.springframework.util.unit.DataSize;
  * @author Stephane Nicoll
  * @author Artem Bilan
  * @author Nakul Mishra
+ * @author Dhiren Mathur
  * @since 1.5.0
  */
 @ConfigurationProperties(prefix = "spring.kafka")
@@ -906,7 +907,7 @@ public class KafkaProperties {
 		 * Whether the container should fail to start if at least one of the configured
 		 * topics are not present on the broker.
 		 */
-		private boolean missingTopicsFatal = true;
+		private boolean missingTopicsFatal = false;
 
 		public Type getType() {
 			return this.type;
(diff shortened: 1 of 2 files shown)


snicoll:
@dhirenmathur thank you for making your first contribution to Spring Boot. Well done writing a test for testing the default, although [we already had one](https://github.com/spring-projects/spring-boot/blob/856543c1c249361a9f4c06007bac41fd7d8ed6f9/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfigurationTests.java#L406-L414). It is interesting we had the same chain of thoughts! Turns out those two tests are identical and  wrong since they assert the outcome of the auto-configuration rather than the actual default in Spring Kafka.

I've moved [that test where it belongs](https://github.com/spring-projects/spring-boot/blob/73ca7030072a2e4c19d3ce3f02d4001cb718e471/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/kafka/KafkaPropertiesTests.java#L45-L50) now. Thanks again!

dhirenmathur:
Thanks @snicoll , don't know how I missed the existing test 😅Thanks for the feedback. 
I'm Glad to be contributing to something I use daily at work 👍 

So now that I have successfully dipped my toes in the water, do you have any other unassigned small issues you can recommend me to try out? Also, are there any channels where I can gain more knowledge about the project workflow?

snicoll:
There is [contributing page](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc) and we have [labelled issues](https://github.com/spring-projects/spring-boot/labels/status%3A%20ideal-for-contribution) for things we believe is suitable to new contributors. However, anything that's triaged and unassigned is fair game. My recommendation is to pick something you care or you use yourself.

If you have more questions, please join us on [Gitter](https://gitter.im/spring-projects/spring-boot)

PR #20956, opened 2020-04-14
Change Handler hashCode method to use URI rather than URL

Konrado85:


snicoll:
@Konrado85 We've decided to fix that in `2.3.x`. If you disagree with that, let the team know in the original issue and we can reconsider but we don't need a PR for it.

PR #20961, opened 2020-04-14
Add support for customizing Spring Session's cookie serializer

vpavic:
This PR introduces `CookieSerializerCustomizer` as callback that allows the customization of auto-configured `DefaultCookieSerializer` bean. This is particularly useful for configuring cookie serializer's capabilities that are not supported by Servlet API and therefore not exposed via `server.servlet.session.cookie.*` properties, such as SameSite.

The existing `RememberMeServicesCookieSerializerCustomizer` has also been refactored to leverage the new callback.


wilkinsona:
Thanks very much, @vpavic.

PR #20967, opened 2020-04-15
Prevent shaded imports from Datastax

dreis2211:
Hi,

I just noticed two imports of `com.datastax.oss.driver.shaded.guava.common.base.Functions` that could be replaced with Java's `java.util.function.Function`.

I also added the shaded package to the checkstyle rules, so accidental imports are prevented in the future.

Cheers,
Christoph


snicoll:
Thanks again Christoph!

PR #20972, opened 2020-04-15
Handler.java - exception handling

Konrado85:
Bug: Exception is caught when Exception is not thrown.

Problem is also probably in the method:  getRootJarFile(String). But I'm not sure how to fix it there


snicoll:
Thanks but those catch block are meant to ignore those specific code paths.

PR #20975, opened 2020-04-15
Simplify application file encoding detection

liaozan:
simplify application file encoding detection，return directly when the condition is not met


snicoll:
Thanks for the PR but I think we should try to restrict the use of `return` which the original code does. There is a case to be made that we check if the property is set and then we get its value which can be simplified. I've done that in https://github.com/spring-projects/spring-boot/commit/fb2f6bb1afb69ec9623e4776818114609d00d574. Thanks again for the PR in any case.

liaozan:
Well, aside from the use of `return`, why does the encoding attribute need to be judged again?
image
The non-null check has been done in the equalsIgnoreCase method, why do we need to repeat？

snicoll:
Because we don't want to run the check at all if the system property is not set.

Issue #9238, opened 2017-05-15
Add response time to health indicators


michaelkuechler:
I recently ran to a situation where my `org.springframework.boot.actuate.endpoint.HealthEndpoint` became slow. debugging the application, I found that just one of the many indicators (that are pulled in without much further do on my side!) actually had a problem. yet the exact instance was hard to identify because there is no easy way to take the timings of each individual indicator.

for such cases it might proof helpful if there were an easy pluggable (or even default) feature that allows me to see individual response times of each indicator. this can be as simple as following little addition to `org.springframework.boot.actuate.endpoint.HealthEndpoint.CompositeHealthIndicator#health` (though there might be cleaner options that won't pollute the scope of `CompositeHealthIndicator`)
```
@Override
public Health health() {
	long start = System.currentTimeMillis();
	Map<String, Health> healths = new LinkedHashMap<>();
	for (Map.Entry<String, HealthIndicator> entry : this.indicators.entrySet()) {
		long start1 = System.currentTimeMillis();
		Health health = entry.getValue().health();
		healths.put(entry.getKey(), withDetail(health, "responseTimeMs", System.currentTimeMillis() - start1));
	}
	Health aggregate = this.healthAggregator.aggregate(healths);
	return withDetail(aggregate,"responseTimeMs", System.currentTimeMillis() - start);
}

/** @see Health.Builder#withDetail(java.lang.String, java.lang.Object) */
private static Health withDetail(Health health, String key, Object value) {
	Health.Builder builder = new Health.Builder(health.getStatus(), health.getDetails());
	builder.withDetail(key, value);
	return builder.build();
}
```

philwebb:
After a little more consideration we've decided that directly recording the response time in the health indicator itself might not be the best approach. Probably a better solution would be to use Micrometer to directly monitor the underlying infrastructure that the heath endpoints are monitoring.

PR #20993, opened 2020-04-16
Proposal, give indication of timing for health check execution.

dugshnay:
I'm submitting this PR as a proposal to provide a mechanism for tracking the duration any given health check implementation took, providing the information in the `details` of the `Health` object. 

In reality, this tracks the nanoseconds between the `Health.Builder` construction time and when the builder's `build()` method is invoked. Though this may blur the execution time very slightly from that of the `doHealthCheck()` method, it would be negligible.

The reason that timing would be beneficial (at least for me) is that in some boot applications the time it takes for the overall health check to return (based on monitoring metrics on the actuator endpoint) continuously increases to a point at which kubernetes will begin to restart our services. I am attempting to identify which individual health check may be responsible, but also feel that this is nice information to have for anyone showing the details of the health endpoint.

This may not be a desired solution, so do please let me know your thoughts. I did attempt to catch these timings with an `@Around` aspect, but was unable to successfully get granular enough for items like the `DataSourceHealthIndicator` which was (for me) just 1 of 4 such health indicators wrapped in a `CompositeHealthIndicator` based on my particular app having 4 DataSources.

I look forward to any and all feedback. If this is accepted, I would back port this at least to 2.1.x, but would go farther back based on guidance and suggestions from contributing members.


snicoll:
@dugshnay thanks for the PR but this was already suggested and we've declined the request, see #9238 for some background.

dugshnay:
@snicoll thank you for the feedback! In reading the notes on the referenced issue, we do monitor the underlying resources, yet they report healthy despite the growing slowness of the health check from a node in the cluster. I'll keep searching for a way to pinpoint this. Have a great day!

PR #20994, opened 2020-04-16
Improve Elasticsearch RestClient customization capabilities

vpavic:
At present, `RestClientBuilderCustomizer` allows general customization of `RestClientBuilder`. This is troublesome for users that want to customize `HttpAsyncClientBuilder` and `RequestConfig.Builder` since those are set on the `RestClientBuilder`. By customizing those two builders users lose out on Spring Boot's support for binding `username`, `password`, `connection-timeout` and `read-timeout` properties from `spring.elasticsearch.rest` namespace.

To illustrate the problem, attempting to customize aspects of `HttpAsyncClientBuilder` like this:

```java
@Bean
public RestClientBuilderCustomizer restClientBuilderCustomizer() throws Exception {
    SSLContext sslContext; // configure custom sslContext
    SchemeIOSessionStrategy sessionStrategy; // configure custom sessionStrategy
    return builder -> builder.setHttpClientConfigCallback(
            httpClientBuilder -> httpClientBuilder.setSSLStrategy(sessionStrategy).setSSLContext(sslContext));
    });
}
```

Will disable the binding of `username` and `password` properties applied here:
https://github.com/spring-projects/spring-boot/blob/70d4994502c848b3db82845c97a033448356c938/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java#L57-L62

To work around this, users need to replicate Spring Boot's configuration logic using something like:

```java
@Bean
public RestClientBuilderCustomizer restClientBuilderCustomizer(RestClientProperties properties) throws Exception {
    SSLContext sslContext; // configure custom sslContext
    SchemeIOSessionStrategy sessionStrategy; // configure custom sessionStrategy
    return builder -> builder.setHttpClientConfigCallback(httpClientBuilder -> {
        httpClientBuilder.setSSLStrategy(sessionStrategy).setSSLContext(sslContext);
        PropertyMapper.get().from(properties::getUsername).whenHasText().to(username -> {
            CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
            Credentials credentials = new UsernamePasswordCredentials(properties.getUsername(),
                    properties.getPassword());
            credentialsProvider.setCredentials(AuthScope.ANY, credentials);
            httpClientBuilder.setDefaultCredentialsProvider(credentialsProvider);
        });
        return httpClientBuilder;
    });
}
```

This PR enhances the `RestClientBuilderCustomizer` with support for customizing `HttpAsyncClientBuilder` and `RequestConfig.Builder` by providing additional `#customize` methods that accept the aforementioned builders. Both new methods are optional as they have no-op default implementations.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java
@@ -23,7 +23,9 @@ import org.apache.http.auth.AuthScope;
 import org.apache.http.auth.Credentials;
 import org.apache.http.auth.UsernamePasswordCredentials;
 import org.apache.http.client.CredentialsProvider;
+import org.apache.http.client.config.RequestConfig;
 import org.apache.http.impl.client.BasicCredentialsProvider;
+import org.apache.http.impl.nio.client.HttpAsyncClientBuilder;
 import org.elasticsearch.client.RestClient;
 import org.elasticsearch.client.RestClientBuilder;
 import org.elasticsearch.client.RestHighLevelClient;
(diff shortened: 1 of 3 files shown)


vpavic:
Also, it might be a good idea to expand the javadoc of the original `RestClientBuilderCustomizer#customize` method (so, the one that takes `RestClientBuilder` argument) and state that using `#setHttpClientConfigCallback` and `#setRequestConfigCallback` on the builder will override the other customizer methods, respectively, as well as the default customizer provided by Spring Boot. WDYT @bclozel?

PR #21030, opened 2020-04-20
Links to CLI binaries are broken for non-snapshot builds

dreis2211:
Hi,

I just noticed that the link to the CLI-binary in the latest milestone builds is not working. Essentially it's always pointing to the `snapshot` repo in Artifactory instead of the appropriate one. See https://docs.spring.io/spring-boot/docs/2.3.0.M4/reference/htmlsingle/#getting-started-manual-cli-installation

As this is also affecting release builds & release candidates, it's probably good to have this one merged before the next release is done.

Cheers,
Christoph

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/AsciidoctorConventions.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/AsciidoctorConventions.java
@@ -156,7 +156,7 @@ class AsciidoctorConventions {
 
 	private String determineArtifactoryRepo(Project project) {
 		String version = project.getVersion().toString();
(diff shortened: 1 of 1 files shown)


dreis2211:
Build failure is due to volume-creation failures.

wilkinsona:
> due to volume-creation failures.

😱

dreis2211:
I think it's me, @wilkinsona . I need to stop sending PRs 😛

wilkinsona:
Thanks for spotting and fixing this, @dreis2211. We have the same problem elsewhere so I've opened https://github.com/spring-projects/spring-boot/issues/21034 as a follow-on.

PR #21037, opened 2020-04-20
ExplodedArchive - missing Serializable

Konrado85:
Comparator doesn't implement Serializable  in EntryComparator class.


wilkinsona:
Thanks for the pull request. I don't believe there's any need for the comparator to implement `Serializable`. Why do you think that it is missing?

Konrado85:
I thought it might be a mistake.

In that case, I'm closing.

PR #21040, opened 2020-04-20
Update link to new Spring Security docs location

dreis2211:
Hi,

I found yet some more links to be broken in the latest snapshot docs. Apparently, the Spring-Security docs only exist under `/html5/` instead of `/htmlsingle/` as of 5.3.x.

Cheers,
Christoph


philwebb:
See https://github.com/spring-projects/spring-security/issues/8420 for discussion about why `html5` is being used.

snicoll:
Unfortunately, it doesn't look like the Spring Security team wants to reconsider the use of `html5` but they've put a redirect in place so that previous milestones now work. Thanks for spotting that @dreis2211!

PR #21054, opened 2020-04-21
Fix Actuator API PDF name

dreis2211:
Hi,

I found another broken link in the docs pointing to the actuator api PDF. An alternative solution to what I've done in this PR is obviously renaming the link in the docs, but that is a bit inconsistent to the other PDFs we have. (And I kind of like to switch between versions in the docs, which is easier if it's the same file name, but that's just personal).

Slightly related, I've noticed that the actuator-api doesn't have a `reference` folder, but directly contains the `pdf` and `html` folder. I was wondering if this should be harmonized maybe. I didn't do it here because it's not part of the bug really.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
@@ -194,6 +194,7 @@ task zip(type: Zip) {
 	duplicatesStrategy "fail"
 	from(asciidoctorPdf.outputDir) {
 		into "pdf"
+		rename { "spring-boot-actuator-web-api.pdf" }
 	}
 	from(asciidoctor.outputDir) {
 		into "html"


dreis2211:
Build failure seems to be related to a missing entry in the micrometer BOM (see https://github.com/micrometer-metrics/micrometer/issues/2012)

dreis2211:
I should note btw that the link in the current snapshot docs work, because there is an old file still available with the correct naming. It is broken in the latest milestone build & will be missing in the upcoming release (candidate) builds.

snicoll:
Good catch @dreis2211 and thank you so much again. I'll remove `index.pdf` on the build has completed. 

To answer your question, I don't think an extra `reference` directory is warranted there. We do this when there is an extra content (typically javadoc).

dreis2211:
Thanks for updating me, @snicoll :)

snicoll:
> I should note btw that the link in the current snapshot docs work, because there is an old file still available with the correct naming. It is broken in the latest milestone build & will be missing in the upcoming release (candidate) builds.

I've removed the duplicate `index.pdf` from `2.3.0.BUILD-SNAPSHOT` and I've fixed the name of the PDF in previous milestones on the doc server so that the link works again.

dreis2211:
Thanks

Issue #21045, opened 2020-04-20
Deprecate environment fallback for Mustache variable resolution


dsyer:
`MustacheEnvironmentCollector` has some strange logic in it (which I probably wrote). It looks for the "native" collector via `super.createFetcher()` but then uses it exclusively if it is not null. But a not-null fetcher is not the same as a fetcher of not-null values. What it should do is inject the fetcher into the `PropertyVariableFetcher` and try to use it before falling back to the `Environment`.

There's a complication though, to do with the fact that JMustache natively resolves variables with period separators as `bean.property` (i.e. "bean" is an object). We wouldn't want to break that, but we don't want to start returning random stuff from the `Environment` that happens to match `bean.property`. It might turn out that this concern is already handled by JMustache (it depends on the order it presents the variable names to the fetcher - does it try `bean` before `bean.property` or the other way round?). Need a test.

philwebb:
Closing in favor of PR #21060

wilkinsona:
Unfortunately, it's flawed to assume that a fetcher that a `null` response from a fetcher indicates that it could not fetch a value. The model could have an accessor that explicitly returns `null`. No fallback attempt should be made at this point and the compiler's `nullValue`, if configured, should be used to determine the result. This faulty assumption resulted in [a regression](https://github.com/spring-projects/spring-boot/issues/22039). I'm going to revert #21060 and re-open this issue so that we can revisit the fix.

wilkinsona:
The more I think about this, the less I am sure we should change the behaviour here. If the context contains a `null` value for a given `name`, I'm not sure that it should fall back to the environment at all. Falling back prevents the context from overriding a value in the environment with `null` and from making use of the compiler's `nullValue` support.

dsyer:
Can we have a debate about this please? The (reasonable) expectation from users is that they can put placeholders in templates for values in the `Environment`. If `null` is not the right signal that this is what is intended, then maybe something else is needed? Like an explicit `env` value in the model or something.

wilkinsona:
> The (reasonable) expectation from users is that they can put placeholders in templates for values in the Environment

[...]

PR #21060, opened 2020-04-21
Fix MustacheEnvironmentCollector to not ignore native fetcher

dsyer:
When the context is a map (as it is in a web View for instance) you can't
assume a non-null fetcher actually contains the property you are searching
for. This change alters the logic so that the native fetcher is always
consulted if it exists, but there is always a fallback.

Fixes gh-21045

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,6 +8,7 @@
 .#*
 .*.md.html
 .DS_Store
+.attach_pid*
 .classpath
 .factorypath
 .gradle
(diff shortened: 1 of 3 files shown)


dsyer:
CI failures look unrelated to me?

wilkinsona:
> CI failures look unrelated to me?

Confirmed. The latest Micrometer 1.5 snapshots have broken our build.

dsyer:
UPDATE: this patch is not complete. It fails to resolve `_csrf.token` in a basic Spring Security app. I'll try and fix that.

philwebb:
I'm tempted to just apply this to 2.3.x rather than the older branches. Any objections @dsyer?

dsyer:
No, I don’t think so.

snicoll:
There is an unrelated change in `.gitignore` that should be removed.

dsyer:
I left it as a separate commit so you can leave it out easily if you want.

philwebb:
Thanks @dsyer! I pulled the ignore commit out and applied it to 2.1.x onwards. The Mustache fix is just on 2.3.x+.

PR #21069, opened 2020-04-21
Improve property name in thrown Micrometer ValidationException

jkschneider:
image

This change makes the message for `ValidationException` include the Spring config property prefix for each registry implementation.

cc / @wilkinsona


wilkinsona:
That's a much nicer approach. Thanks very much, @jkschneider.

wilkinsona:
Thanks very much, @jkschneider.

Issue #20920, opened 2020-04-11
Document that schema.sql and data.sql will still be run when using Flyway or Liquibase but discourage their use


hauer-io:
**Expected Behavior:**
- The sql file "src/main/resources/data.sql"  is run independent from the database during start up
- The sql file "src/test/resources/data.sql"  is run independent from the database on test start

**Actual Behavior:**
 - The sql file "src/main/resources/data.sql" is only run with h2 and not postgres during start up
 - The sql file "src/test/resources/data.sql"  is only run with h2 and not postgres on test start

**Failed approaches:**
  - Use property "spring.datasource.initialization-mode=always"
  - Use property "spring.liquibase.drop-first=true"
  - Use property "spring.jpa.hibernate.ddl-auto=create-drop"

**Initial analyse:**
- With postgres there is no "DataSourceSchemaCreatedEvent" so the DataSourceInitializer.initSchema() is never run.

 Example-project: https://github.com/hauer-io/liquibase

snicoll:
That's the wrong expectation I am afraid. When you run a database migration with Liquibase, this is taking control of the schema and the basic `schema.sql` & `data.sql` does not run at all.

Said differently, you can't combine, [Execute Liquibase Database Migrations on Startup
](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto-execute-liquibase-database-migrations-on-startup) and [Initialize a database](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto-initialize-a-database-using-spring-jdbc).

The name of the latter section has been renamed in `2.3.x` but we could certainly improve the documentation to make that more obvious.

hauer-io:
Okay. But then the bug is with H2. Since there both is executed.

snicoll:
Can you demonstrate that with a small sample?

hauer-io:
Example-project: https://github.com/hauer-io/liquibase
Default profile h2 and postgres profile with postgres. So easy to compare that they behave differently.

snicoll:
> Default profile h2 and postgres profile with postgres. So easy to compare that they behave differently.

They behave differently because one is creating the schema via Hibernate and one does not. [According to the documentation](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#boot-features-creating-and-dropping-jpa-databases), embedded databases are created automatically on startup if you're using JPA. That's a sensible default since the database and its content are destroyed at every run.

[...]

PR #21077, opened 2020-04-22
Enhance Database initialization how to guide

nikmanzotti:
As discussed in #20920, I have enhanced the Database initialization guide clarifying the expected behavior when using a higher-level Database Migration Tool.

I also wanted to add a description for the `spring.datasource.initialization-mode` property, but I wasn't able to find the `config-docs/data.adoc` file. Proposed description for that property:
> Type of datasource you want to initialize. Use `always` to initialize the DataSource regardless of its type.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto.adoc
@@ -1946,7 +1946,7 @@ For example, you might choose to set it to the vendor name of the database (`hsq
 
 [NOTE]
 ====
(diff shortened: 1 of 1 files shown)


snicoll:
That's excellent @nikmanzotti, thank you very much.

snicoll:
> I also wanted to add a description for the spring.datasource.initialization-mode property

It is generated from the code, the current description is "Initialize the datasource with available DDL and DML scripts." I agree that this can be improved and we can do that as part of a separate issue. Type of datasource can be misleading IMO.

snicoll:
@nikmanzotti thank you for making your first contribution to Spring Boot.

PR #21108, opened 2020-04-24
Fix duplicate slashes in URL links

jkatada:
Fix links which have duplicate slashes in reference documentation.

link examples
* https://docs.spring.io/spring-boot/docs/2.3.0.BUILD-SNAPSHOT/api//org/springframework/boot/SpringApplication.html
* https://docs.oracle.com/javase/8/docs/api//java/time/Duration.html#parse-java.lang.CharSequence-
* https://docs.spring.io/initializr/docs/current/reference/html//#user-guide
* ...

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/attributes.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/attributes.adoc
@@ -24,11 +24,11 @@
 :test-examples: ../test/java/org/springframework/boot/docs
 
 :spring-boot-code: https://github.com/{github-repo}/tree/{github-tag}
-:spring-boot-api: https://docs.spring.io/spring-boot/docs/{spring-boot-version}/api/
+:spring-boot-api: https://docs.spring.io/spring-boot/docs/{spring-boot-version}/api
 :spring-boot-docs: https://docs.spring.io/spring-boot/docs/{spring-boot-version}/reference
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/java/org/springframework/boot/build/autoconfigure/DocumentAutoConfigurationClasses.java
```diff
@@ -91,7 +91,7 @@ private void writeTable(AutoConfiguration autoConfigurationClasses) throws IOExc
 				writer.println();
 				writer.printf("| {spring-boot-code}/spring-boot-project/%s/src/main/java/%s.java[`%s`]%n",
 						autoConfigurationClasses.module, autoConfigurationClass.path, autoConfigurationClass.name);
-				writer.printf("| {spring-boot-api}/%s.html[javadoc]%n", autoConfigurationClass.path);
+				writer.printf("| {spring-boot-api}%s.html[javadoc]%n", autoConfigurationClass.path);
```
snicoll, line 94: I think we should rather stick to a `/` after the variable rather than no slash. This means that `{spring-boot-api}` would not end with a slash.

jkatada:
@snicoll Thanks for the feedback. I've updated the PR. Please review.

snicoll:
@jkatada thanks for the follow-up and for making your first contribution to Spring Boot.

PR #21133, opened 2020-04-26
Aggregate TestExecutionListener orders

andreisilviudragnea:
Aggregated TestExecutionListener orders into:
- `SpringBootTestExecutionListenerOrder` for `spring-boot-test` module;
- `SpringBootAutoConfigureTestExecutionListenerOrder` for `spring-boot-test-autoconfigure` module.

This aggregation helps third-party integrators when choosing the order for a new listener.

I have done two similar PRs in [spring-framework](https://github.com/spring-projects/spring-framework/pull/24977) and [spring-security](https://github.com/spring-projects/spring-security/pull/8436) projects.

### DIFF
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/SpringBootAutoConfigureTestExecutionListenerOrder.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/SpringBootAutoConfigureTestExecutionListenerOrder.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2002-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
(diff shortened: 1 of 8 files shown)


wilkinsona:
Thanks for the proposal. I'd like to see what @sbrannen thinks before we consider merging this into Spring Boot. FWIW, I'm not sure that this should be an enum as, to me, an enum implies a finite set and we may add other listeners in the future.

wilkinsona:
I've also just noticed that, as proposed, this change introduces a package tangle between `org.springframework.boot.test.context` and all the packages that define execution listeners. I don't see how that can be avoided as the central location for defining the orders will have to know about every listener and every listener will have to know about the central location to get its order. Unless I've overlooked something and there's a way to break that cycle, I think we'll have to decline this.

andreisilviudragnea:
@wilkinsona Yes, I saw too the problem with not being able to centralize the listener order of the two modules `spring-boot-test` and `spring-boot-test-autoconfigure`.

Maybe at least making the order of each listener statically accessible, through a constant on each listener would work too.

I am working on introducing Mockito Strict Stubbing support to `@MockBean`-based tests and I found myself digging a lot through the code in order to understand the order of the listeners both in
the Spring TestContext Framework and Spring Boot Test itself.

This order, as you know, is essential to the correct functioning of the framework, but is hidden to anyone who does not really understand how it works. I thought that maybe making it more visible would help other people, apart from me, when trying to extend the existing functionality.

andreisilviudragnea:
The main idea was to enable someone extending the framework to specify something like "this listener needs to get executed before the `ResetMocksTestExecutionListener` from Spring Boot Test" with an order relative to the one of `ResetMocksTestExecutionListener`.

Maybe I overthought this and only extracting each order to a constant on each listener would have been enough.

wilkinsona:
I"m going to close this one as I can't see a way to centralise the filter ordering without introducing a package tangle. Thanks anyway for the suggestion, @andreisilviudragnea.

Issue #11890, opened 2018-02-02
MetricsAutoConfiguration causes big chunk of early instantiations


Dav1dde:
The [`MeterRegistryPostProcessor`](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAutoConfiguration.java#L75) defined in `MetricsAutoConfiguration` causes a bunch of early bean instantiations.

```
2018-02-02 14:36:04.166  INFO 15124 --- [           main] f.a.AutowiredAnnotationBeanPostProcessor : JSR-330 'javax.inject.Inject' annotation found and supported for autowiring
2018-02-02 14:36:04.339  INFO 15124 --- [           main] trationDelegate$BeanPostProcessorChecker : Bean 'org.springframework.transaction.annotation.ProxyTransactionManagementConfiguration' of type [org.springframework.transaction.annotation.ProxyTransactionManagementConfiguration$$EnhancerBySpringCGLIB$$e97a5208] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
2018-02-02 14:36:04.456  INFO 15124 --- [           main] trationDelegate$BeanPostProcessorChecker : Bean 'org.springframework.boot.context.properties.ConfigurationPropertiesBinderBuilder$ConversionServiceFactory' of type [org.springframework.boot.context.properties.ConfigurationPropertiesBinderBuilder$ConversionServiceFactory] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
2018-02-02 14:36:04.463  INFO 15124 --- [           main] trationDelegate$BeanPostProcessorChecker : Bean 'management.metrics-org.springframework.boot.actuate.autoconfigure.metrics.MetricsProperties' of type [org.springframework.boot.actuate.autoconfigure.metrics.MetricsProperties] is not eligible for getting processed by all BeanPostProcessors (for example: not eligible for auto-proxying)
[...]

PR #21134, opened 2020-04-26
Proper ordering of metrics auto-configurations

bendiscz:
Metrics auto-configurations that depend on a `MeterRegistry` bean
should be configured AFTER configurations that create the
`MeterRegistry`. However, `@AutoConfigureAfter(MetricsAutoConfiguration.class)`
is not sufficient. `MetricsAutoConfiguration` does not export a `MeterRegistry`.

This commit changes the `@AutoConfigureAfter` annotations so that the
proper ordering of configurations in ensured.

#### The long story:
I ran into issues in one of my spring-boot services when trying to fix [a Micrometer bug](https://github.com/micrometer-metrics/micrometer/issues/2028). After adding a simple custom auto-configuration with dependency on `MeterRegistry` and `LogbackMetrics`, the application failed to start because of a missing `LogbackMetrics` dependency. Debug output showed that the `LogbackMetricsAutoConfiguration` is disabled because the condition `@ConditionalOnBean(MeterRegistry.class)` is not met. After more debugging I found out that `LogbackMetricsAutoConfiguration` was evaluated before `PrometheusMetricsExportAutoConfiguration`.

I created [a sample project](https://github.com/bendiscz/logback-metrics-bug) that demonstrates this issue.

A simple workaround that I currently use in my projects is a dummy auto-configuration that inserts itself between `LogbackMetricsAutoConfiguration` and configurations that create the `MeterRegistry`:
```java
@Configuration
@AutoConfigureAfter({ CompositeMeterRegistryAutoConfiguration.class, SimpleMetricsExportAutoConfiguration.class })
@AutoConfigureBefore({
        JvmMetricsAutoConfiguration.class,
        KafkaMetricsAutoConfiguration.class,
        Log4J2MetricsAutoConfiguration.class,
        LogbackMetricsAutoConfiguration.class,
        SystemMetricsAutoConfiguration.class
})
public class ProperMetricsOrderingAutoConfiguration {
    // empty
}
```

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
@@ -37,7 +37,7 @@ import org.springframework.context.annotation.Configuration;
  * @since 2.1.0
  */
 @Configuration(proxyBeanMethods = false)
-@AutoConfigureAfter(MetricsAutoConfiguration.class)
+@AutoConfigureAfter(CompositeMeterRegistryAutoConfiguration.class)
 @ConditionalOnClass(MeterRegistry.class)
 @ConditionalOnBean(MeterRegistry.class)
 public class JvmMetricsAutoConfiguration {
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/KafkaMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/KafkaMetricsAutoConfiguration.java
@@ -43,7 +43,7 @@ import org.springframework.kafka.core.ProducerFactory;
  */
 @Configuration(proxyBeanMethods = false)
 @AutoConfigureBefore(KafkaAutoConfiguration.class)
-@AutoConfigureAfter(MetricsAutoConfiguration.class)
+@AutoConfigureAfter(CompositeMeterRegistryAutoConfiguration.class)
 @ConditionalOnClass({ KafkaClientMetrics.class, ProducerFactory.class })
 @ConditionalOnBean(MeterRegistry.class)
 public class KafkaMetricsAutoConfiguration {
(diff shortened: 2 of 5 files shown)


pivotal-issuemaster:
@bendiscz Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21134)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21134) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@bendiscz Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21134)!

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
```diff
@@ -37,7 +38,7 @@
  * @since 2.1.0
  */
 @Configuration(proxyBeanMethods = false)
-@AutoConfigureAfter(MetricsAutoConfiguration.class)
+@AutoConfigureAfter({ CompositeMeterRegistryAutoConfiguration.class, SimpleMetricsExportAutoConfiguration.class })
```
snicoll, line 41: `SimpleMetricsExportAutoConfiguration runs before `CompositeMeterRegistryAutoConfiguration` so I don't understand why both of them have to be defined here. If you run after `CompositeMeterRegistryAutoConfiguration`, `SimpleMetricsExportAutoConfiguration` should have been processed too.

bendiscz: Yes, you are right, `CompositeMeterRegistryAutoConfiguration` should be sufficient. I added both because I saw this combination in other metrics auto-configurations, but –looking at it again – those were configurations for exporters (and the ordering was "before").

I'm going to change it.

bendiscz:
> Did you fix those 5 auto-configurations because you were affected by them or have you reviewed all of them and those particular 5 were a problem?

I was directly affected by `LogbackMetricsAutoConfiguration`. Then I checked all other metrics auto-configurations in the same package and saw that they have the same problem.

philwebb:
Possible related to #11890

philwebb:
Thanks very much for the PR @bendiscz. One of the reasons we've got away with this for so long is that `CompositeMeterRegistryAutoConfiguration` just happens to be alphabetically before the configurations that use it.

Unfortunately, despite my best efforts, I couldn't find an easy way to test the fix. Each test I tried ended up causing a order cycle. I'm pretty confident with the fix, so we'll have to just run with it.

PR #21219, opened 2020-04-29
Support additional R2DBC pool properties

rodolphocouto:
This PR adds some missing properties from R2DBC pool:

* max-life-time
* max-acquire-time
* max-create-connection-time
* validation-depth

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
@@ -68,7 +68,6 @@ abstract class ConnectionFactoryConfigurations {
 				ObjectProvider<ConnectionFactoryOptionsBuilderCustomizer> customizers) {
 			ConnectionFactory connectionFactory = createConnectionFactory(properties, resourceLoader.getClassLoader(),
 					customizers.orderedStream().collect(Collectors.toList()));
-
 			R2dbcProperties.Pool pool = properties.getPool();
 			ConnectionPoolConfiguration.Builder builder = ConnectionPoolConfiguration.builder(connectionFactory)
 					.initialSize(pool.getInitialSize()).maxSize(pool.getMaxSize()).maxIdleTime(pool.getMaxIdleTime())
@@ -79,7 +78,6 @@ abstract class ConnectionFactoryConfigurations {
 			if (StringUtils.hasText(pool.getValidationQuery())) {
 				builder.validationQuery(pool.getValidationQuery());
 			}
-
 			return new ConnectionPool(builder.build());
 		}
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
@@ -43,6 +43,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Mark Paluch
  * @author Stephane Nicoll
+ * @author Rodolpho S. Couto
  */
 abstract class ConnectionFactoryConfigurations {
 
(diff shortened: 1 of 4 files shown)


snicoll:
Everyone, thank you for all the reviews, I think we're good now.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
```diff
@@ -67,12 +68,18 @@ ConnectionPool connectionFactory(R2dbcProperties properties, ResourceLoader reso
 				ObjectProvider<ConnectionFactoryOptionsBuilderCustomizer> customizers) {
 			ConnectionFactory connectionFactory = createConnectionFactory(properties, resourceLoader.getClassLoader(),
 					customizers.orderedStream().collect(Collectors.toList()));
+
```
snicoll, line 71: This doesn't seem necessary.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
```diff
@@ -67,12 +68,18 @@ ConnectionPool connectionFactory(R2dbcProperties properties, ResourceLoader reso
...
-					.maxSize(pool.getMaxSize()).initialSize(pool.getInitialSize()).maxIdleTime(pool.getMaxIdleTime());
+					.initialSize(pool.getInitialSize()).maxSize(pool.getMaxSize()).maxIdleTime(pool.getMaxIdleTime())
+					.maxLifeTime(pool.getMaxLifeTime()).maxAcquireTime(pool.getMaxAcquireTime())
+					.maxCreateConnectionTime(pool.getMaxCreateConnectionTime())
+					.validationDepth(pool.getValidationDepth());
+
 			if (StringUtils.hasText(pool.getValidationQuery())) {
 				builder.validationQuery(pool.getValidationQuery());
 			}
+
```
snicoll, line 82: Ditto

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/R2dbcProperties.java
```diff
@@ -153,6 +171,11 @@ public String determineUniqueName() {
 		 */
 		private String validationQuery;
 
+		/**
+		 * Validation depth.
+		 */
+		private ValidationDepth validationDepth = ValidationDepth.LOCAL;
```
snicoll, line 177: Enum default value are not detected by the annotation processor so additional metadata has to be written for that unfortunately.

Here is [an example](https://github.com/spring-projects/spring-boot/blob/64aac04862cd129dea2f8e929fb977f1138806cb/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L498-L501).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/R2dbcProperties.java
```diff
@@ -138,6 +141,21 @@ public String determineUniqueName() {
 		 */
 		private Duration maxIdleTime = Duration.ofMinutes(30);
 
+		/**
+		 * Max lifetime.
+		 */
+		private Duration maxLifeTime = Duration.ZERO;
```
snicoll, line 147: Please use the relevant `of` with `0` instead so that we can extract the proper unit.

snicoll:
@rodolphocouto how is it going? Have you had a chance to look at the review?

rodolphocouto:
Hey @snicoll, thanks for your review. I started working on that today and I should finish it soon.

snicoll:
@rodolphocouto thanks for the follow-up. Are you done with d3d38d2? I am asking as I've requested to complete the existing tests to exercise the code you've added. Thanks!

rodolphocouto:
I just added some tests on `R2dbcProperties`. I couldn't find a way to test `ConnectionFactory` or `ConnectionPool` because they don't expose those properties. Does that make sense to you?

snicoll:
Reviewing the code, I've noticed 3 properties that have a `Duration` of zero to express the absence of a timeout. This looks really odd to me and I am not keen to bring such a default in auto-completion for our users.

We could fix that by having no default value for those 3 properties (the lack of a timeout value meaning no timeout is something we do elsewhere). And only map the value if it's non-null. I've also [created an issue](https://github.com/r2dbc/r2dbc-pool/issues/77) and I'd like some feedback before moving forward.

snicoll:
This is going to be changed in `r2dbc-pool` so let's go with the "no default" option for the time being. I'll add a test that will fail as soon as the default changes so that we remember we need to do something about it.

PR #21259, opened 2020-04-29
Optimize StringSequence.startsWith

dreis2211:
Hi,

I noticed that a vanilla Boot app spends some time in `StringSequence.startsWith` or respectively `StringSequence.charAt`:
image

This PR optimizes `startsWith` in cases where the length doesn't differ from the source length, which essentially means that StringSequence is "equal" to the source. In those cases, we can use `startsWith` on the actual source and by that avoid the repeated boundary checks inside charAt. A little benchmark of old vs. new approach shows the following results.

```
Benchmark                                Mode  Cnt   Score    Error   Units
MyBenchmark.testNew                      avgt   10   6,270 ±  0,255   ns/op
MyBenchmark.testNew:·gc.alloc.rate       avgt   10  ≈ 10⁻⁴           MB/sec
MyBenchmark.testNew:·gc.alloc.rate.norm  avgt   10  ≈ 10⁻⁶             B/op
MyBenchmark.testNew:·gc.count            avgt   10     ≈ 0           counts
MyBenchmark.testOld                      avgt   10   9,029 ±  0,330   ns/op
MyBenchmark.testOld:·gc.alloc.rate       avgt   10  ≈ 10⁻⁴           MB/sec
MyBenchmark.testOld:·gc.alloc.rate.norm  avgt   10  ≈ 10⁻⁶             B/op
MyBenchmark.testOld:·gc.count            avgt   10     ≈ 0           counts

```

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/StringSequence.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/StringSequence.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


philwebb:
Nice!

philwebb:
Moving the parameter type to `String` opens the door for even more optimizations I think. @dreis2211 and @wilkinsona I'd appreciate it if you could look at commit 4a8492d428be5c659acc007e3b6f1462ff9fe645 and see if it makes sense.

philwebb:
Reopening to ensure we get at least one other pair of eyes on this before we release.

dreis2211:
Looks good to me. In reality it won't change much though, because there are actually no other cases in the code where the length differs from the original string (apart from tests). So it's a good polishing commit, but no real performance improvement anymore ;)

wilkinsona:
Thanks, @dreis2211. FWIW, it looks good to me too.

philwebb:
Thanks for the reviews!

PR #21283, opened 2020-04-30
Allow for customization of RedisCacheWriter

josephtaylor:
Currently it's difficult to customize the `RedisCacheWriter` when using redis-based caching and still take advantage of the auto-configuration provided by boot. This is an attempt to make it possible to declare a `RedisCacheWriter` bean and have the auto-configuration use it.

If the user doesn't create their own `RedisCacheWriter` bean this should result in the same functionality as before.


josephtaylor:
Another approach to this would be to allow for setting the `RedisCacheWriter` via a `RedisCacheManagerBuilderCustomizer`.

From what I can tell that would require changes to `RedisCacheManagerBuilder` in `spring-data-redis`.

philwebb:
@josephtaylor Thanks very much for the pull-request. I think the version of Spring Data Redis that we'll be using in Spring Boot 2.3 offers a `RedisCacheManagerBuilder.cacheWriter(...)` method which would open up the option of using a `Customizer` (see [DATAREDIS-1118](https://jira.spring.io/browse/DATAREDIS-1118)).

Would you be interested in reworking your pull-request in that direction? It seems like it would be more flexible than the bean approach.

josephtaylor:
Ah sorry I didn't see that. I _think_ with the change they're making to the builder, the cache writer should be customizable without any code changes here because there's already a list of `RedisCacheManagerBuilderCustomizer` being injected in the bean method here.

PR #21289, opened 2020-05-01
Add properties to control exceptions ignored by LdapTemplate

filiphr:
Added properties:

* `ignorePartialResultException` - For ignoring `PartialResultException` when searching with the `LdapTemplate`
* `ignoreNameNotFoundException` - For ignoring `NameNotFoundException` when searching with the `LdapTemplate`
* `ignoreSizeLimitExceededException` - For ignoring `SizeLimitExceededException` when searching with the `LdapTemplate`

I was not sure whether you prefer setting the defaults in the properties as they are done in the target class or use `null` and only set them when they are explicitly set. I can change it if you prefer to have the defaults in the `LdapProperties`

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/LdapAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ldap/LdapAutoConfiguration.java
@@ -64,8 +64,16 @@ public class LdapAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean(LdapOperations.class)
-	public LdapTemplate ldapTemplate(ContextSource contextSource) {
-		return new LdapTemplate(contextSource);
+	public LdapTemplate ldapTemplate(LdapProperties properties, ContextSource contextSource) {
+		PropertyMapper propertyMapper = PropertyMapper.get().alwaysApplyingWhenNonNull();
+		LdapTemplate ldapTemplate = new LdapTemplate(contextSource);
+		propertyMapper.from(properties.isIgnorePartialResultException())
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks for the PR, @filiphr.

> I was not sure whether you prefer setting the defaults in the properties as they are done in the target class or use null and only set them when they are explicitly set. 

Where possible, we've been moving towards setting the defaults in the properties and then adding some tests to make sure that out defaults align with the target class's defaults. The default from our `@ConfigurationProperties` class is then available to users in their IDE when auto-completing the property.

filiphr:
>Where possible, we've been moving towards setting the defaults in the properties and then adding some tests to make sure that out defaults align with the target class's defaults.

Makes sense. Do you perhaps have some test that you can point me to so I can have a look at it and add it for this as well?

filiphr:
I found

https://github.com/spring-projects/spring-boot/blob/ff3b05a421213b360473fd1a379d80d20c527611/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/amqp/RabbitPropertiesTests.java#L299-L306

I'll do it like that for the `LdapProperties` as well.

filiphr:
Thanks for the hint @wilkinsona. I've amended my commit with and added `LdapPropertiesTests`

wilkinsona:
Thanks very much, @filiphr. I polished things a little bit in [this commit](https://github.com/spring-projects/spring-boot/commit/8c341df73b8253eaacbbc2af130c60731cf08bc5), primarily to group the new properties under `spring.ldap.template`, aligning them with properties for `JdbcTemplate`, `RabbitTemplate`, etc.

filiphr:
That's great @wilkinsona. Thanks a lot for merging it and for polishing it. I was not aware about the grouping for the other templates.

PR #21311, opened 2020-05-04
Health indicators based on Service Level Objectives

jkschneider:
This feature adds support for commonly requested functionality for an application to be able to aggregate some set of metrics key performance indicators down to a health indicator. 

Some indicators are known to be broadly applicable to a wide range of Java applications, and those could be autoconfigured. An example of a set of such indicators is defined [here](https://www.jetbrains.com/help/teamcity/teamcity-memory-monitor.html) and autoconfigured by this pull request (`JvmServiceLevelObjectives.MEMORY`).

In many cases, users would like to configure a load balancer to avoid instances that are failing a key performance indicator by configuring an HTTP health check on the load balancer. In fact, some applications may already be doing this for the health indicators Spring Boot or users already provide. Example platform load balancer configurations that can be pointed to `/actuator/health`:

* CloudFoundry [health-check-http-endpoint](https://docs.cloudfoundry.org/devguide/deploy-apps/manifest-attributes.html#health-check-http-endpoint)
* AWS ALB [health checking](https://docs.aws.amazon.com/elasticloadbalancing/latest/classic/elb-healthchecks.html)
* Kubernetes service health checking:

```yml
metadata:
  name: instance-reported-utilization
  annotations:
    service.beta.kubernetes.io/do-loadbalancer-healthcheck-port: "80"
    service.beta.kubernetes.io/do-loadbalancer-healthcheck-protocol: "http"
    service.beta.kubernetes.io/do-loadbalancer-healthcheck-path: "/actuator/health"
```

See https://github.com/micrometer-metrics/micrometer/issues/2055 for more detail.

## The `HealthMeterRegistry`

As of 1.6.0, Micrometer has a new implementation: `micrometer-registry-health`. A autoconfiguration was added to `spring-boot-actuator-autoconfigure` for this new implementation.

Any `@Bean ServiceLevelObjective` is configured onto the `HealthMeterRegistry` and bound as a Spring Boot `HealthIndicator`.

## What it looks like in `/actuator/health`

image

## About `ServiceLevelObjective`

Service level objectives broadly have the following capabilities:

* Are defined as a single or multi-indicator test against a set of time series registered to `HealthMeterRegistry`.
* Can define required `MeterBinder` that contain the measurements that they need to determine availability.
* Contains a filterable and transformable name and tag set that is mapped to the Spring Boot bean name and `Health#details` map, respectively.
* Optionally contains a readable base unit that is mapped to health details.
* Can pretty-print values and thresholds for human-readable interpretation of an SLO at some instant.
* Can be defined to look back and aggregate over a time window in different ways.


jkschneider:
## Open questions

We build health indicators with `AbstractHealthIndicator(slo.getFailedMessage())`. It's unclear to me when over if the failed message ever appears in `/actuator/health` response body output.

Some of the SLOs are a combination of two or more indicators. For example, in `jvmTotalMemory`, we set a relatively low threshold on GC overhead (20% of CPU time over the last 5 minutes) _if_ there is also 90% pool utilization as well. These composite SLOs are registered with the relatively new `CompositeHealthContributor.fromMap(..)` API. Unfortunately there is no way I can see to provide details and a failed message name on the composite. I'd like to add details and a failed message name for each contributing health indicator and potentially a different one for what it means for a set of such indicators to fail together.

philwebb:
Thanks @jkschneider! I'll target this for 2.4.x so we remember to take a look as soon the 2.3.0 release crunch is over.

bclozel:
We haven't had a chance to take a look at this change, nor upgrade to Micrometer 1.6.
We're already quite late in the Milestone cycle and we don't think we'll have time to address this change properly.
We need to take a look at this change and its implications (including the new concepts introduced and the Health endpoint format).

mbhave:
@snicoll and I discussed this today. There are few things that came up:

1. Since we decided that the diskspace health indicator should ideally be something that can be configured in the monitoring system, this feels very much along those lines. If we decide to surface the SLO's as a health indicator, we should align our strategy for diskspace accordingly. Even if the deprecation of the diskspace indicator, we could surface that information in health via the SLOs.
2. We are not sure if having a top-level component for every SLO is the best way to do this. Maybe having some sort of nested structure for the SLOs might be a better alternative.
3. From an API perspective, we could have an API to expose SLOs which we could use to create the composite rather than the current method which registers beans within a bean method. 

Flagging for team-meeting so that we can discuss this on the next team call.

wilkinsona:
We discussed this some more as a team today and our feeling is that we're not sure that we have a strong enough opinion to auto-configure SLOs has health indicators. We can see that it may make sense for some users but not for others. For example, in some cases, a proxy will already be aware of the error rate for requests that it routes to an application instance. In this case, exposing the information via a health endpoint that it will also be monitoring will be of minimal value, and may even be harmful depending on how things behave when the application's health changes. For users that do want to expose SLOs as health indicators, we could provide some classes that make it easier to do so. 

Since this proposal was made, we've also introduced the concept of application state. It may be that some users want to configure things such that an unmet objective results in a change to the application state to indicate that it's no longer ready, for example. We could provide some helper classes that a user can configure to connect SLOs to application state.

We discussed possibly auto-configuring the `HealthMeterRegistry`, automatically adding any `ServiceLevelObjective` beans to it. We could auto-configure some `ServiceLevelObjective` beans such as `JvmServiceLevelObjectives.MEMORY` and `OperatingSystemServiceLevelObjectives.DISK` rather than hard-coding them as proposed here. This would align with our auto-configuring of Micrometer's various `Jvm…Metrics` classes.

Overall, our feeling was that we would stop short of anything that exposes the SLOs externally, instead auto-configuring the `HealthMeterRegistry` and supporting beans and making it easier for a user to then plug the SLOs into health or application state in a way that meets their specific needs.

@shakuzen @jonatan-ivarenov Could we have your input here please? Are we right to be cautious and just give users the parts they need and leave them to join things together or is there some clearly established usage of `HealthMeterRegistry` and SLOs that means that we can proceed with confidence in a particular direction?

mjf1310 (approved):

mjf1310:
sorry,operation error

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/health/HealthMetricsExportAutoConfiguration.java
```diff
@@ -0,0 +1,150 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
```
ronodhirSoumik, line 2: 2012-2025?
same for all files?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/health/package-info.java
```diff
@@ -0,0 +1,20 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Support for building health indicators with service level objectives.
+ */
+package org.springframework.boot.actuate.autoconfigure.metrics.export.health;
```
ronodhirSoumik, line 20: just to know - use of this file?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/health/HealthMetricsExportAutoConfiguration.java
philwebb: @ronodhirSoumik, thanks, but there's no need to comment on PRs for copyright dates. We always fix things like that when we merge code.

snicoll:
This PR has been opened for so long that it is no longer actionable. I've created #48327 for us to investigate the topic again when time permits.

PR #21365, opened 2020-05-08
Fix JarFileTests.getInputStreamWhenClosed()

izeye:
This PR fixes `JarFileTests.getInputStreamWhenClosed()` as the assertion was passing due to `this.jarFile.getEntry()`, not `this.jarFile.getInputStream()` that was intended to assert in the test.


snicoll:
Good catch!

Issue #20707, opened 2020-03-28
Polish PR "Dependency Upgrades" section


snicoll:
Our PR template for dependency upgrades currently states:

```
Dependency Upgrades

Please do not open a pull request for a straightforward dependency upgrade (one that
only updates the version property). We have a semi-automated process for such upgrades
that we prefer to use. However, if the upgrade is more involved (such as requiring
changes for removed or deprecated API) your pull request is most welcome.
```

I think a key element is missing in this text. Rather than focusing on the fact it only updates one line with a version property, we should IMO focus on the fact it is covered by `bomr` and therefore will be taken care "automatically".

Versions that are handled internally (and aren't covered by bomr) won't be checked automatically and can be outdated unless someone looks at it. Any help we can get to stay current is most welcome. 

Maybe we could rephrase the text to mention the content of `spring-boot-dependencies` since that's what bomr uses?

snicoll:
We've decided to keep the note as it is and expand bomr coverage, see #21439. We will merge PRs that only update the version property on a case-by-case basis. For internal dependencies, we feel it is less important to upgrade right away unless there is a benefit in doing so.

PR #21378, opened 2020-05-10
Upgrade to Testcontainers 1.14.1

dreis2211:
Hi,

this PR upgrades to Testcontainers 1.14.1.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-parent/build.gradle
+++ b/spring-boot-project/spring-boot-parent/build.gradle
@@ -12,7 +12,7 @@ javaPlatform {
 
 dependencies {
 	api(enforcedPlatform(project(":spring-boot-project:spring-boot-dependencies")))
-	api(enforcedPlatform("org.testcontainers:testcontainers-bom:1.14.0"))
+	api(enforcedPlatform("org.testcontainers:testcontainers-bom:1.14.1"))
 
 	constraints {
 		api("com.vaadin.external.google:android-json:0.0.20131108.vaadin1")


snicoll:
We are still discussing whether we want to accept certain "only version" upgrades PR, see #20707. I am going to put this one on-hold until we get a chance to finalize this discussion.

dreis2211:
Sure thing. I kept myself quite for a bit as the version was released two weeks ago already, I think. If that contributes to the discussion a bit.

philwebb:
We should upgrade this anyway, regardless of #20707. I'm not sure if BOMR will pick it up so I'll drop it in 2.3.0 just so we don't forget.

dreis2211:
Bomr won't pick it up. That's part of the discussion: if one-liner upgrades should be rejected even though they're not picked up by bomr (which is essentially only stuff spring-boot-dependencies)

snicoll:
Thanks for that @dreis2211.

> I kept myself quite for a bit as the version was released two weeks ago already, I think.

As I've indicated [in this comment](https://github.com/spring-projects/spring-boot/issues/20707#issuecomment-628076147), it is less important/urgent to upgrade an internally managed dependency. We do so when we could benefit from a new feature or the dependency fixed an issue we're facing. Otherwise, such upgrade happens in due course as part of a general chore similar to how we apply bomr upgrades.

dreis2211:
I think the perfect solution is teaching Bomr to handle internal dependencies, too - as suggested in #21439 . That avoids manual work 👍

Issue #21291, opened 2020-05-01
Elasticsearch: Support REST URIs including userInfo


murdos:
Setting the `spring.elasticsearch.rest.uris` with value `http://login:password@host:port` generates `UnknownHostException: login:password@host: Name or service not known`

It would be nice if the userInfo part of the URL could also be used as credentials (that would be overridden if `spring.elasticsearch.rest.username` or `spring.elasticsearch.rest.password` are defined)

FYI I encountered this issue while trying to use [Bonsai with Heroku](https://devcenter.heroku.com/articles/bonsai) that sets a variable BONSAI_URL including login and token.

philwebb:
It looks like this should be possible, although it's a little involved since we support more than one URL in `ElasticsearchRestClientProperties`.

I think to implement it we'd need to change `ElasticsearchRestClientConfigurations`. I think we'd be able to parse the `URL` ourselves using `java.net.URI` to access the `userInfo`. The `DefaultRestClientBuilderCustomizer` would need to change to apply a custom `CredentialsProvider` that would be able to lookup the userinfo for a given host.

evgeniycheban:
Can I work on this issue?

wilkinsona:
While reviewing the PR from @evgeniycheban I've given some thought to the following:

> It would be nice if the userInfo part of the URL could also be used as credentials (that would be overridden if spring.elasticsearch.rest.username or spring.elasticsearch.rest.password are defined)

I don't think the credentials from the URL should be overridden by the properties. The properties should only be used for URLs with no user info. This will align things with, for example, Boot's support for the addresses used to connect to RabbitMQ.

philwebb:
Closing in favor of PR #21381

PR #21381, opened 2020-05-11
Support user info in Elasticsearch URIs

evgeniycheban:
Fixes gh-21291

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.autoconfigure.elasticsearch;
 
+import java.net.URI;
 import java.time.Duration;
 
 import org.apache.http.HttpHost;
@@ -43,6 +44,7 @@ import org.springframework.context.annotation.Configuration;
  * @author Brian Clozel
  * @author Stephane Nicoll
  * @author Vedran Pavic
+ * @author Evgeniy Cheban
  */
 class ElasticsearchRestClientConfigurations {
 
@@ -58,7 +60,7 @@ class ElasticsearchRestClientConfigurations {
 		@Bean
 		RestClientBuilder elasticsearchRestClientBuilder(ElasticsearchRestClientProperties properties,
 				ObjectProvider<RestClientBuilderCustomizer> builderCustomizers) {
-			HttpHost[] hosts = properties.getUris().stream().map(HttpHost::create).toArray(HttpHost[]::new);
+			HttpHost[] hosts = properties.getUris().stream().map(this::createHttpHost).toArray(HttpHost[]::new);
 			RestClientBuilder builder = RestClient.builder(hosts);
 			builder.setHttpClientConfigCallback((httpClientBuilder) -> {
 				builderCustomizers.orderedStream().forEach((customizer) -> customizer.customize(httpClientBuilder));
@@ -70,6 +72,12 @@ class ElasticsearchRestClientConfigurations {
 			});
 			builderCustomizers.orderedStream().forEach((customizer) -> customizer.customize(builder));
 			return builder;
+		}
+
+		private HttpHost createHttpHost(String uri) {
+			URI parsedUri = URI.create(uri);
+			String userInfo = parsedUri.getUserInfo();
+			return HttpHost.create((userInfo != null) ? uri.replace(userInfo + "@", "") : uri);
 		}
 
 	}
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks very much for the PR, @evgeniycheban 

Could you please take a look at adding some tests for the new functionality? I think they could be implemented as additions to `org.springframework.boot.autoconfigure.elasticsearch.ElasticsearchRestClientAutoConfigurationTests`. It looks like you can get the nodes from the `RestClient` to verify that the configuration has been applied correctly. As part of the tests, it would be good to verify the behaviour of URIs with and without any user info and how the separate username and password properties are used.

evgeniycheban:
> Thanks very much for the PR, @evgeniycheban
> 
> Could you please take a look at adding some tests for the new functionality? I think they could be implemented as additions to `org.springframework.boot.autoconfigure.elasticsearch.ElasticsearchRestClientAutoConfigurationTests`. It looks like you can get the nodes from the `RestClient` to verify that the configuration has been applied correctly. As part of the tests, it would be good to verify the behaviour of URIs with and without any user info and how the separate username and password properties are used.

Thanks for the review, @wilkinsona 

Sure, I will add tests.

evgeniycheban:
@wilkinsona I added tests.

When you have a moment, please take a look.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java
```diff
@@ -114,25 +123,48 @@ RestClient elasticsearchRestClient(RestClientBuilder builder) {
 
 		private final ElasticsearchRestClientProperties properties;
 
+		private CredentialsProvider credentialsProvider = new BasicCredentialsProvider();
```
wilkinsona, line 126: Creating a field for the `CredentialsProvider` and allowing it to be injected doesn't seem to be related to supported user info in the URI. Can you please revert this part of the changes?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientAutoConfigurationTests.java
```diff
@@ -156,6 +163,69 @@ void restClientCanQueryElasticsearchNode() {
...
+
+					CredentialsProvider credentialsProvider = context.getBean(CredentialsProvider.class);
+
+					Credentials uriCredentials = credentialsProvider.getCredentials(new AuthScope("localhost", 9200));
+					assertThat(uriCredentials.getUserPrincipal().getName()).isEqualTo("user");
+					assertThat(uriCredentials.getPassword()).isEqualTo("password");
+
+					Credentials defaultCredentials = credentialsProvider
+							.getCredentials(new AuthScope("localhost", 9201));
+					assertThat(defaultCredentials.getUserPrincipal().getName()).isEqualTo("admin");
```
wilkinsona, line 214: I think the username in the URI should win so the name of the principal should be `user`.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientAutoConfigurationTests.java
```diff
@@ -156,6 +163,69 @@ void restClientCanQueryElasticsearchNode() {
...
+					assertThat(defaultCredentials.getPassword()).isEqualTo("admin");
```
wilkinsona, line 215: Similarly, I think the password in the URI should win so it should be `password`.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientAutoConfigurationTests.java
evgeniycheban: Yes, that is right, in this assert we are checking that if the URI does not contain credentials then `spring.elasticsearch.rest.username` and `spring.elasticsearch.rest.password` will be used.

wilkinsona: Sorry, I'd missed that there are two URIs being configured, one with credentials and one without.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientAutoConfigurationTests.java
wilkinsona: Sorry, ignore this. I'd missed that there are two URIs being configured, one with credentials and one without.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java
evgeniycheban: I added this to make testing easer, because we can not access `credentialsProvider` field from `InternalHttpAsyncClient`. It can be also useful in some cases when the user needs to use a custom `CredentialsProvider`.

wilkinsona: I don't think it's safe to expose a `CredentialProvider` as a bean. In this case the credentials that it provides are specific to Elasticsearch, but there's nothing about the bean that indicates that's the case. It could be accidentally consumed and used anywhere that's working with an Apache HTTP client, potentially leaking credentials to another service.

For this sort of situation, we prefer to use AssertJ's support for extracting fields. In this case it would look something like this:

```java
assertThat(client).extracting("client")
		.extracting("credentialsProvider", InstanceOfAssertFactories.type(CredentialsProvider.class))
		.satisfies((credentialsProvider) -> {
			Credentials uriCredentials = credentialsProvider.getCredentials(new AuthScope("localhost", 9200));
			assertThat(uriCredentials.getUserPrincipal().getName()).isEqualTo("user");
			assertThat(uriCredentials.getPassword()).isEqualTo("password");
			Credentials defaultCredentials = credentialsProvider.getCredentials(new AuthScope("localhost", 9201));
			assertThat(defaultCredentials.getUserPrincipal().getName()).isEqualTo("admin");
			assertThat(defaultCredentials.getPassword()).isEqualTo("admin");
		});
```

evgeniycheban: Thank you for the AssertJ example, @wilkinsona 

I will update tests.

evgeniycheban: Fixed.

evgeniycheban:
@wilkinsona Could you please take a look at the latest changes when you have a moment?

murdos:
@wilkinsona : is there any chance to get this in 2.3.x ? 
Indeed this feature was available with Jest, but support has been removed in 2.3.

wilkinsona:
Yes. Given the situation with Jest, it seems reasonable to include this in 2.3.x.

philwebb:
Thanks very much for the PR @evgeniycheban, this is now in 2.3.x and master. I made some minor tweaks in f8982bdc14af90b590c8131c09b95b05edebdd79 (mainly to protect against URL parse exceptions).

Issue #20335, opened 2020-02-27
Improve test coverage of launch script


wilkinsona:
Fixing #18540 and #19716 is higher risk that we would like as the changes that appear to be necessary aren't covered particularly well by the launch script's integration tests. For example the tests focus on starting an app via init.d and the problem reported in #19716 happens when executing the jar directly.

aivinog1:
@wilkinsona Hello! I want to start to contribute and fix this issue. Do you mind if I start working on it?

wilkinsona:
Not at all. Thank you very much for the offer.

mikesmithson:
@wilkinsona -  I would also like to give this a shot. Without a ton of context, are you saying that issues #18540 and #19716 cannot be addressed until test coverage of launch script is improved. Is there a test that needs additional test cases? Is this RepackagerTests.java? What tests do you see as being missing?

wilkinsona:
Thanks for the offer, @mikesmithson, but @aivinog1 is already looking at this one.

aivinog1:
@wilkinsona Hello. I have a question. Do we need in this task implement tests that start an application as systemd service? I don't see such tests in the project right now. And I stuck in writing a test that verifies that `DEBUG` variable works correctly (because as I can see we can't pass a variable in init.d service, besides modifying launch.script). So I can write such a test but not for init.d service.

wilkinsona:
> Hello. I have a question. Do we need in this task implement tests that start an application as systemd service?

Testing as a systemd service would be nice in the longer term, but I don't think it's of most importance. I would focus on launching the jar directly as a root and non-root user.

> And I stuck in writing a test that verifies that `DEBUG` variable works correctly

The behaviour of `DEBUG` is pretty low level and I think we could make some big improvements with getting to that level of testing. Focusing on the kind of testing that's currently done with init.d but for direct execution will bring greater benefit, IMO

aivinog1:
@wilkinsona Hello, I've developed some tests and I want to understand, do I do them properly. How can you review them? I can send a Github link on my commit (by the way here it is: https://github.com/aivinog1/spring-boot/commit/a1a97bbb5f4fc5bb5c4595202a4e26e96461c093). Or I can open a new PR.
Can I open PR without full coverage and further will open PR on each test?

Thanks.

wilkinsona:
@aivinog1 They look good to me. Thank you very much. Please do open a PR when you have a moment and then we can go through a proper review.

[...]

PR #21388, opened 2020-05-11
 Improve test coverage of launch script

aivinog1:
Implementation of #20335. Splitting launch script's integration tests on two: init.d tests and shell-like. Add some cases to test.


wilkinsona:
@aivinog1 Thanks very much for making your first contribution to Spring Boot.

PR #21390, opened 2020-05-11
Fix ClientHttpConnectorConfigurationTests.shouldApplyHttpClientMapper()

izeye:
This PR fixes `ClientHttpConnectorConfigurationTests.shouldApplyHttpClientMapper()` as it doesn't test that the mapper is called as intended.


snicoll:

PR #21394, opened 2020-05-11
Calculate hashCode only once in ConfigurationPropertyName

dreis2211:
Hi,

I just noticed that the new hashCode logic in `ConfigurationPropertyName` doesn't set the new `hashCode` field. Seems to be an oversight that might lead to performance issues, (although I haven't benchmarked anything).

Cheers,
Christoph


philwebb:
Nice spot. Thanks!

PR #21480, opened 2020-05-17
Configure tests using TestContainer to use the actual container IP address

olamy:
At the Jetty  project we want to have daily/weekly of Opensource project using Jetty but with last SNAPSHOT. To be sure we are not breaking anything :)
As our CI environment is based on K8s pods, using localhost for docker images started with testcontainers doesn't work.
I just made the changes to get it working (see https://jenkins.webtide.net/job/external_oss/job/spring-boot-build-jetty-9.4.x-snapshot/) and using the testcontainer recommended usage https://www.testcontainers.org/features/networking/#getting-the-container-ip-address.

Another question I have (maybe can be in a separate issue If you prefer?).
I looked at implementing this SNAPSHOT build for spring-boot/spring-framework but with the gradle build. Unfortunately I cannot change the jetty version dynamically. I tried different way but as it's using mavenBom and not the usual gradle dependency management system (this doesn't work see script here: https://github.com/jetty-project/external-oss-builds/blob/master/init.gradle.spring-framework) 
If you have any idea/trick please let me know?

Signed-off-by: olivier lamy <olamy@webtide.com>

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseAutoConfigurationIntegrationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseAutoConfigurationIntegrationTests.java
@@ -74,7 +74,7 @@ class CouchbaseAutoConfigurationIntegrationTests {
 	void setUp() {
 		this.context = new AnnotationConfigApplicationContext();
 		this.context.register(CouchbaseAutoConfiguration.class);
-		TestPropertyValues.of("spring.couchbase.bootstrap-hosts=localhost",
+		TestPropertyValues.of("spring.couchbase.bootstrap-hosts=" + couchbase.getContainerIpAddress(),
 				"spring.couchbase.env.bootstrap.http-direct-port:" + couchbase.getMappedPort(8091),
 				"spring.couchbase.username:spring", "spring.couchbase.password:password",
 				"spring.couchbase.bucket.name:default").applyTo(this.context.getEnvironment());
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseAutoConfigurationIntegrationTests.java
```diff
@@ -74,7 +74,7 @@ static void createBucket() {
 	void setUp() {
 		this.context = new AnnotationConfigApplicationContext();
 		this.context.register(CouchbaseAutoConfiguration.class);
-		TestPropertyValues.of("spring.couchbase.bootstrap-hosts=localhost",
+		TestPropertyValues.of("spring.couchbase.bootstrap-hosts=" + couchbase.getContainerIpAddress(),
```
bsideup, line 77: FYI Testcontainers 1.14.2 adds `getHost()` (returns the same value as `getContainerIpAddress()` but less confusing), consider using it once Spring Boot updates TC to 1.14.2

olamy:
@wilkinsona as you are the main contributor of https://github.com/spring-gradle-plugins/dependency-management-plugin any idea regarding the question about dependencies? Do you prefer create a separate issue for that?

wilkinsona:
@olamy Hard to say right now as I'm not sure what the problem is. The link you shared above 404s for me. Private repo, perhaps?

olamy:
@wilkinsona oh sorry it's now public.

wilkinsona:
@olamy I've opened [an issue](https://github.com/jetty-project/external-oss-builds/issues/1) on that repository with a suggestion.

snicoll:
Thanks for the PR Olivier!

PR #21490, opened 2020-05-18
Add an example for setting a custom image name on the command line with Gradle

dsyer:
The Maven docs have an example, so it seems like a good idea to have one here too.


spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/docs/asciidoc/packaging-oci-image.adoc
```diff
@@ -151,4 +151,11 @@ include::../gradle/packaging/boot-build-image-name.gradle.kts[tags=image-name]
 ----
 
 Note that this configuration does not provide an explicit tag so `latest` is used.
-It is possible to specify a tag as well, either using `${project.version}`, any property available in the build or a hardcoded version.
\ No newline at end of file
+It is possible to specify a tag as well, either using `${project.version}`, any property available in the build or a hardcoded version.
+
+The builder can be specified on the command line as well, as shown in this example:
```
wilkinsona, line 156: `builder` -> `image name`?

PR #21523, opened 2020-05-20
Optimize SystemEnvironmentPropertyMapper

dreis2211:
Hi,

I just noticed a couple of minor performance tweaks in `SystemEnvironmentPropertyMapper`. The biggest one is probably to avoid creating `ConfigurationPropertyName` instances twice for every `isAncestorOf` check (once for the normal `of()` call and once for the `isValid()` check).

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/SystemEnvironmentPropertyMapper.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/SystemEnvironmentPropertyMapper.java
@@ -58,7 +58,7 @@ final class SystemEnvironmentPropertyMapper implements PropertyMapper {
 		StringBuilder result = new StringBuilder();
 		for (int i = 0; i < numberOfElements; i++) {
 			if (result.length() > 0) {
-				result.append("_");
+				result.append('_');
 			}
 			result.append(name.getElement(i, Form.UNIFORM).toUpperCase(Locale.ENGLISH));
 		}
@@ -69,7 +69,7 @@ final class SystemEnvironmentPropertyMapper implements PropertyMapper {
 		StringBuilder result = new StringBuilder();
 		for (int i = 0; i < name.getNumberOfElements(); i++) {
 			if (result.length() > 0) {
-				result.append("_");
+				result.append('_');
 			}
 			result.append(convertLegacyNameElement(name.getElement(i, Form.ORIGINAL)));
 		}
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/SystemEnvironmentPropertyMapper.java
```diff
@@ -116,13 +116,24 @@ private boolean isLegacyAncestorOf(ConfigurationPropertyName name, Configuration
 		if (!hasDashedEntries(name)) {
 			return false;
 		}
+		StringBuilder legacyCompatibleName = buildLegacyCompatibleName(name);
+		try {
+			return ConfigurationPropertyName.of(legacyCompatibleName).isAncestorOf(candidate);
+		}
+		catch (Exception ex) {
```
snicoll, line 123: Can you share what that extra catch block is supposed to cover?

dreis2211: The removal of `isValid` here might lead to exceptions if the name can't be constructed.

snicoll: I'd argue that `legacyCompatibleName` is a bit misleading then. It makes me wonder looking at the code why a catch block is necessary. Thoughts?

dreis2211: I didn't change anything in that regards. I think the name was chosen to reflect the "legacy" format and `isValid()` was added to be sure that no incompatible output breaks the code here. This PR just sticks to that behaviour of not breaking anything with a potentially invalid name.
Having that said, I too had a hard time seeing how an invalid name can be produced from a valid name, but didn't want to change the intention of the code.

snicoll: Alright, thanks for the feedback. Flagging for team attention to see what the rest of the team thinks.

philwebb:
Thanks once again @dreis2211! I split this up into a few smaller commits and also added a new method to `ConfigurationPropertyName` to save needing to throw/catch the exception.

Issue #20901, opened 2020-04-09
Potential NPE in MetadataGenerationEnvironment.java


HermioneSW:
Hi,

Some more potential NPEs, maybe you can take a look. They are a little complicated here I will describe the chain of calls.

1. In ```MetadataGenerationEnvironment.java``` (spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/MetadataGenerationEnvironment.java), function ```getAnnotation``` return null at line 198.

2. In ```ConfigurationMetadataAnnotationProcessor``` (spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessor.java), function ```processEndpoint```, ```annotation``` on line 243 will be assigned null.

3. Same file as step 2 line 245 ```processEndpoint``` will be called with the first argument ```annotation``` being null.

4. Still same file line 253 ```processEndpoint``` function the first line ```Map<String, Object> elementValues = this.metadataEnv.getAnnotationElementValues(annotation);``` is called with ```annotation``` being null.

5.  In ```MetadataGenerationEnvironment.java``` (spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/MetadataGenerationEnvironment.java), function ```getAnnotationElementValues``` at line 237. The expression ```annotation.getElementValues()``` wil throw a NPE at line 239.

We found some similar issues, but you can confirm this one before we report the others. Again, thanks so much for providing the feedback!!!

philwebb:
@HermioneSW Have you seen this NPE in the wild? Or did you find the potential problem using some kind of static analysis?

HermioneSW:
@philwebb we found it using our static analysis tool.

HermioneSW:
@philwebb could you please confirm this is indeed a bug? Thanks so much!

philwebb:
@HermioneSW I categorized it as a task rather than a bug since we've not actually seen it being triggered in the wild. It's targeted to 2.2.x and we'll change the code when we can.

aivinog1:
@philwebb Hello. Do you mind if I take this? I think I can fix these problems.

wilkinsona:
Looking at the code, I don't think this NPE can happen.

[...]

PR #21530, opened 2020-05-21
Switch from nullable method signature to Optional method signature

aivinog1:
This is a part of the fix to #20901. I've decided to move one method to `Optional` to make it null-safe.


philwebb:
@aivinog1 We've generally tried to keep out use of `Optional` to a minimum in the Spring Boot codebase. If can have some advantages, but it also has a habit of adding quite a lot of noise to the codebase.

There's an [interesting talk by Stuart Marks](https://www.youtube.com/watch?v=fBYhtvY19xA) that's worth a watch if you want some more background on where the JDK team recommend that it's used.

I think I'd prefer to keep the `Optional` change out of the fix for #20901. Thanks anyway.

PR #21559, opened 2020-05-25
Print the java version being used on startup

### LINKED ISSUE #21553: Print the java version being used on startup
The frequency of JDK releases would make this useful information to have.

dreis2211:
Hi,

this PR closes #21553 by using `System.getProperty("java.version")`. Alternatively, we could check for `Runtime.version()` being available (which is the case for JDK 9+) and use that.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/StartupInfoLogger.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/StartupInfoLogger.java
@@ -70,6 +70,7 @@ class StartupInfoLogger {
 		appendOn(message);
 		appendPid(message);
 		appendContext(message);
+		appendJavaVersion(message);
 		return message;
 	}
 
@@ -145,6 +146,10 @@ class StartupInfoLogger {
 			message.append(context);
 			message.append(")");
 		}
+	}
+
+	private void appendJavaVersion(StringBuilder message) {
+		append(message, "on Java ", () -> System.getProperty("java.version"));
 	}
 
 	private void append(StringBuilder message, String prefix, Callable<Object> call) {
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, @dreis2211. Here's what's logged by `StartupInfoLoggerTests#sourceClassIncluded` with the proposed changes:

```
Starting StartupInfoLoggerTests on fulcrum.local with PID 42978 (started by awilkinson in /Users/awilkinson/dev/spring-projects/spring-boot/master/spring-boot-project/spring-boot) on Java 1.8.0_252
```

I wonder if the Java version should appear earlier in the message. I also wonder about the two uses of `on`, one to refer to the host and the other to the JVM.

What do you think of something like the following:

```
Starting StartupInfoLoggerTests using Java 1.8.0_252 on fulcrum.local with PID 42978 (started by awilkinson in /Users/awilkinson/dev/spring-projects/spring-boot/master/spring-boot-project/spring-boot)
```

dreis2211:
Funny enough, I had a similar version before (with `with` instead of `using` but hey) and I too find the double `on` sub-optimal. I decided to append it at the end to less likely break it for anyone parsing this log statement - (I've seen people do this for startup statistics).

I'm totally fine with what you suggested. Let me know if I should change it.

wilkinsona:
@dreis2211 Thanks. Given that this will go into 2.4, I'm comfortable with the small risk of breaking someone's log message parsing. It's an easy one for them to correct and we don't make any guarantees about the structure of any of our messages. 

I have a slight preference for `using` rather than `with` to avoid duplicating the `with` in `with PID …`.

If you have a moment to make the updates, that'd be great. Thank you.

dreis2211:
@wilkinsona Sure. Done.

wilkinsona:
Thanks again, @dreis2211.

PR #21579, opened 2020-05-26
Polish duration support javadoc

OrangeDog:
~~A breaking change if someone was using the Period converters externally, but they should've been package-final from the start.~~


snicoll:
Thanks for the PR but those are two completely unrelated changes unless I am missing something. We're still discussing whether we want to move that one package private so we can't review the second commit.

Can you please drop 69b543d and we can merge this one as a Javadoc polish.

OrangeDog:
Ah, didn't see that issue.
What about making the `@See` consistent?

snicoll:
If you revert the change of visibility we can consider that as part of a javadoc polish indeed.

snicoll:
Thanks for the follow-up @OrangeDog and for making your first contribution to  Spring Boot.

PR #21581, opened 2020-05-26
Upgrade to Testcontainers 1.14.2

dreis2211:
Hi,

this PR upgrades Testcontainers to 1.14.2 and uses the new `getHost` method instead of using `getContainerIpAddress`.

Cheers,
Christoph


dreis2211:
Clearly it's hard to break habits. I just noticed that I shouldn't open this one at all. Sorry. Feel free to close.

snicoll:
> I just noticed that I shouldn't open this one at all. Sorry. Feel free to close.

Our template states: "Please do not open a pull request for a straightforward dependency upgrade (one that only updates the version property)."

This PR is more involved so it does not fall in that category. Thanks for your efforts @dreis2211.

dreis2211:
Nonetheless, I didn't think about it really and just executed things based on muscle memory. If it didn't do the getHost migration, I would have opened it as well, which I shouldn't have done.

snicoll:
Yeah well it's not big deal. If you had done that we'd have closed it. It's no the end of the World either :)

dreis2211:
On another note: test failures seem unrelated.

Issue #21604, opened 2020-05-28
Add Java 15 CI


wilkinsona:
It might be a bit early for us (we waited for the JDK's RC phase for Java 14), but Framework 5.3 is building against Java 15 already. We should consider when we want to start doing the same.

dreis2211:
There are simply no AdoptOpenJDK builds for JDK 15 yet. I wonder if Spring Framework uses OpenJDK EA builds!?

aahlenst:
We're doing JDK 15 for quite some time: https://github.com/AdoptOpenJDK/openjdk15-binaries/releases. But we don't surface them yet on the website.

dreis2211:
That's what I meant to say, sorry. Spring-Boot does a check against the API to check for JDK upgrades, which wouldn't work on 15 because it's not officially exposed to the outside. But that might be acceptable.

dreis2211:
Btw...I would volunteer to do this again for JDK 15.

aahlenst:
@dreis2211 If we can make anything easier, please open tickets in either https://github.com/AdoptOpenJDK/openjdk-website or https://github.com/AdoptOpenJDK/openjdk-api-v3.

dreis2211:
@aahlenst I think there is already https://github.com/AdoptOpenJDK/openjdk-api-v3/issues/212 that would be sufficient ;)

philwebb:
> Btw...I would volunteer to do this again for JDK 15.

That would be great if you have time. Thanks @dreis2211!

dreis2211:
If there is no big pressure here, I'd like to resolve it in a way so that we don't have to skip the update logic for JDK 15, which means a little bit of waiting time until the AdoptOpenJDK issue is resolved. But I can also simply skip the update logic for JDK 15. Just let me know.

snicoll:
I don't think getting automatic update right away is a blocker. If that changes the pipeline config too much, we can wait.

dreis2211:
It doesn't - from what I have in mind. Just wanted to have it clarified upfront. Will work on JDK 15 next week then ;)

snicoll:
Closing in favour of PR #21173

PR #21605, opened 2020-05-28
Fix multi-release JAR test on JDK 15

dreis2211:
Hi,

I just did a quick for #21604 to see if everything works on JDK 15 with the custom `buildJavaHome` property I introduced for JDK 14.

This PR fixes a test failure in `JarFileTests.multiReleaseEntry`.

I'm wondering if we should add some entries in advance already for JDK 16 & 17 ;-)

Cheers,
Christoph


snicoll:
> I'm wondering if we should some entries in advance already for JDK 16 & 17 ;-)

Ah! I don't think we should do that as we'll have to do a check every time we need to support a new JDK anyway.

dreis2211:
I agree - I just wanted to mention it at least once. ;)

Issue #21648, opened 2020-06-02
Change livenessPort to management port in liveness and readiness probe example


mbhave:
See https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-kubernetes-probes.

fastso:
Hi, can I take the task?

mbhave:
Closing in favor of PR #21650.

PR #21650, opened 2020-06-02
Change liveness-port to management-port

fastso:
Change liveness-port to management-port in liveness and readiness probe example. #21648

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
@@ -900,6 +900,8 @@ readinessProbe:
   periodSeconds: ...
 ----
 
+NOTE: `"management-port"` means the port that the actuator endpoints are available on. It could be the server port if a separate management port has not been configured or the `"management.server.port"` property if that has been explicitly configured.
+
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, @fastso. There are some commits included in it that shouldn't be there. I think that may be because you have used 2.3.x as the base branch for a change that you made on top of the master branch. Can you please try [changing the base branch of your PR](https://help.github.com/en/github/collaborating-with-issues-and-pull-requests/changing-the-base-branch-of-a-pull-request) from 2.3.x to master?

fastso:
@wilkinsona Thank you for your help, I changed the branch from 2.3.x to master.

mbhave:
Thanks for the PR @fastso. Since `management-port` can mean a few things things we should probably add a note below that example clarifying what it means. Essentially, it means the port that the actuator endpoints are available on. It could be the server port if a separate management port has not been configured or the `management.server.port` property if that has been explicitly configured. Would you be able to update the PR with the NOTE that clarifies this?

fastso:
Thanks for the advice @mbhave . I added the note for management-port, plz confirm.

philwebb:
Thanks very much for the PR @fastso. I've merged this into 2.3.x along with a [minor edit](https://github.com/spring-projects/spring-boot/commit/47729be4e37d19069c9eaaff26dace7be11ef4b6) to the naming. I decided to use the term `<actuator-port>` rather than `management-port`.

Issue #21616, opened 2020-05-29
Add support for easily disabling metrics export


snicoll:
Most monitoring systems we have support for are push based and some of them will prevent the context to start if an API token (or account id) is not configured. 

Each of them have a `.enabled` property that can be used to disable that check (and the export of metrics for it. It would be nice if we had a monitoring agnostic way of doing this, while still allowing the metrics to be asserted in memory.

`@AutoConfigureMetrics` comes to mind. I don't know if that would be similar to `@AutoConfigureCache` but something that makes sure metrics are only exported in memory would be quite convenient, I suspect.

bono007:
@snicoll I have this next week available for OSS and am interested in this one. I want to make sure you have not already headed out on this.

snicoll:
Thanks for the offer @bono007. I am not assigned, so no. 

I said I would write a summary of our discussion after the team meeting and I forgot. In the meantime, I might have forgot some of it. 

What I do remember is that we want in memory metrics to be the default in tests and no monitoring system specific auto-configurations should apply. This isn't possible at the moment as our metrics auto-configuration can adapt multiple `MeterRegistry` beans (in case you'd want to publish metrics to multiple systems). So some changes is required there. The purpose of `@AutoConfigureMetrics` would be to override that behaviour.

bono007:
Thanks @snicoll for adding that detail - its helpful.

> auto-configuration can adapt multiple MeterRegistry beans (in case you'd want to publish metrics to multiple systems).
I am very familiar w/ this as we do it at my current company. 

I will start digging in and will ping back here any questions I have.

snicoll:
Closing in favour of PR #21658

PR #21658, opened 2020-06-03
Add support for easily disabling metrics export

bono007:
Add global metrics exporter disable property; set during tests with ability to override via `@AutoConfigureMetrics`.

Closes gh-21616

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/appoptics/AppOpticsMetricsExportAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/appoptics/AppOpticsMetricsExportAutoConfigurationTests.java
@@ -52,7 +52,14 @@ class AppOpticsMetricsExportAutoConfigurationTests {
 	}
 
 	@Test
-	void autoConfigurationCanBeDisabled() {
+	void autoConfigurationCanBeDisabledWithGlobalEnabledProperty() {
+		this.contextRunner.withUserConfiguration(BaseConfiguration.class)
+				.withPropertyValues("management.metrics.export.enabled=false").run((context) -> assertThat(context)
+						.doesNotHaveBean(AppOpticsMeterRegistry.class).doesNotHaveBean(AppOpticsConfig.class));
+	}
+
+	@Test
+	void autoConfigurationCanBeDisabledWithSpecificEnabledProperty() {
 		this.contextRunner.withUserConfiguration(BaseConfiguration.class)
 				.withPropertyValues("management.metrics.export.appoptics.enabled=false")
 				.run((context) -> assertThat(context).doesNotHaveBean(AppOpticsMeterRegistry.class)
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/metrics/AutoConfigureMetrics.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/metrics/AutoConfigureMetrics.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.test.autoconfigure.metrics;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that can be applied to a test class to enable auto-configuration for metrics
+ * exporters. By default, all metrics exporters other than the in-memory
+ * {@code SimpleMetricsExportAutoConfiguration} are disabled.
+ *
+ * @author Chris Bono
+ * @since 2.4.0
+ */
+@Target(ElementType.TYPE)
+@Retention(RetentionPolicy.RUNTIME)
+@Documented
+@Inherited
+public @interface AutoConfigureMetrics {
+
+}
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/metrics/AutoConfigureMetrics.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/metrics/AutoConfigureMetrics.java
@@ -25,8 +25,7 @@ import java.lang.annotation.Target;
 
 /**
  * Annotation that can be applied to a test class to enable auto-configuration for metrics
- * exporters. By default, all metrics exporters other than the in-memory
- * {@code SimpleMetricsExportAutoConfiguration} are disabled.
+ * exporters.
  *
  * @author Chris Bono
  * @since 2.4.0
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/metrics/ExcludeMetricExportersContextCustomizerFactoryTests.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/metrics/ExcludeMetricExportersContextCustomizerFactoryTests.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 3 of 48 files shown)


spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/metrics/ExcludeMetricExportersContextCustomizerFactoryTests.java
```diff
@@ -0,0 +1,74 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link AutoConfigureMetrics} and
+ * {@link ExcludeMetricExportersContextCustomizerFactory} working together.
+ *
+ * @author Chris Bono
+ */
+class ExcludeMetricExportersContextCustomizerFactoryTests {
```
bono007, line 33: This test and the sibling ones `AutoConfigureMetricsMissing/PresentIntegrationTest` overlap on coverage. If I choose between them I think that the integration test ones are more valuable as they verify against `@SpringBootTest` as the annotated class.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/appoptics/AppOpticsMetricsExportAutoConfigurationTests.java
```diff
@@ -52,7 +52,14 @@ void autoConfiguresItsConfigAndMeterRegistry() {
 	}
 
 	@Test
-	void autoConfigurationCanBeDisabled() {
+	void autoConfigurationCanBeDisabledWithGlobalEnabledProperty() {
```
bono007, line 55: I will do this same mod to the other export config tests once I know we are good w/ the direction of this proposal.

bono007:
@snicoll here is what I captured along the way while trying different approaches.

## Background

*Metircs auto-configuration topology (wrt AutoConfigureBefore/After of each other)*
```
MetricsAutoConfiguration	
	WavefrontMetricsExportAutoConfiguration (enabled!=false -> WavefrontMeterRegistry)
	DatadogMetricsExportAutoConfiguration (enabled!=false -> DatadogMeterRegistry)
	..
		SimpleMetricsExportAutoConfiguration (COMB-MR, enabled!=false -> SimpleMeterRegistry)
			CompositeMeterRegistryAutoConfiguration	
					- NoOpMeterRegistryConfiguration (COMB-MR -> CompositeMeterRegistry empty)
					- CompositeMeterRegistryConfiguration (multi MRs -> CompositeMeterRegistry w/ MRs) 
				MetricsEndpointAutoConfiguration
```

## Goal
To disable everything between `MetricsAutoConfiguration` and `SimpleMetricsExportAutoConfiguration` by default in tests. With ability to override and re-enable all/individual exporters.

## Approaches
I tried several approaches leveraging `ImportAutoConfiguration` and built-in auto-config exclusion etc. but they all ended w/ non-success or non-starters shortcomings. 

*TLDR;* Jump to approach #5 to see the content of this code proposal.


The following are of the "Don't modify any of the existing exporter auto-configurations" variety:

*Approach #1:*
------------
- @AutoConfigureMetrics as an @ImportAutoConfiguration which maps to list of ACs to use from spring.factories
- AutoConfigurationImportFilter that exclude ACs ending in "MetricsExportAutoConfiguration"
*Status:* Does not work. The ACIF exclusions take precedence over the IAC list.

*Approach #2:* 
------------
- @AutoConfigureMetrics as an @ImportAutoConfiguration which maps to list of ACs to use from spring.factories
- ContextCustomizer that sets the 'spring.autoconfigure.exclude' property to list of ACs from spring.factories (above)
*Status:* Does not work. The exclusion always overrides the inclusions.

*Approach #3:*
------------
- @AutoConfigureMetrics simple marker interface, nothing more
- ContextCustomizer that sets 'spring.autoconfigure.metric-exporters.enabled' based on whether the test is marked as @AutoConfigureMetrics
- AutoConfigurationImportFilter that exclude ACs ending in "MetricsExportAutoConfiguration" unless above property is set (aka marked w/ @AutoConfigureMetrics)

*Status:* works 
- Pros:
  - simple to implement (no change to existing auto-configs)
  - users custom metric exporters will be respected IF named properly
- Cons: 
  - excludes by naming scheme ("*MetricExportAutoConfiguration")
  - binary enablement (all or nothing) - no way to override and add an inclusion/exclusion

*Approach #4:*
------------
- @AutoConfigureMetrics simple marker interface, nothing more
- ContextCustomizer that sets the 'spring.autoconfigure.exclude' property to hardcoded set of metric exporter configs unless the test is marked w/ @AutoConfigureMetrics 

*Status:* works 
- Pros:
  - simple to implement (no change to existing auto-configs)
- Cons: 
  - excludes by naming scheme ("*MetricExportAutoConfiguration") (NOTE: this can be alleviated by reading list from spring.factories)
  - binary enablement (all or nothing)
  - users custom metric exporters will not be respected in exclusion


The following is of the "Do what you need to do to make this work properly" variety:

*Approach #5:*
------------
- @AutoConfigureMetrics simple marker interface, nothing more
- ContextCustomizer that sets new "global" 'management.metrics.export.enabled' based on whether the test is marked as @AutoConfigureMetrics
- @ConditionalOnMetricExportsEnabled that inspects global and specific property to decide match or not

*Status:* works
- Pros:
  - allows all export to be disabled w/ single property but respects specific export overrides
  - can be used outside of test-land
- Cons: 
  - small change to existing auto-configs - but not functional breaking change
  - users custom metric exporters will be respected in exclusion if they copy pattern and use the @ConditionalOnMetricExportsEnabled


## Outcome
New integration test defaults is that only `SimpleMeterRegistry` will be enabled. The `management.metrics.export[.<name>].enabled` properties can be used to adjust that. 

## Next
I have submitted a handful of tests to sanity check the functionality and illustrate the usage of the new annotation. If this is a direction that we agree on then I will add the missing tests and update the docs accordingly in a subsequent commit.

spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/metrics/AutoConfigureMetrics.java
```diff
@@ -0,0 +1,40 @@
...
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+/**
+ * Annotation that can be applied to a test class to enable auto-configuration for metrics
+ * exporters. By default, all metrics exporters other than the in-memory
+ * {@code SimpleMetricsExportAutoConfiguration} are disabled.
```
snicoll, line 29: I think that could confuse users as it can be read what the default behaviour of adding the annotation does. I'd rather keep it simple and we can handle the rationale in the reference guide rather.

bono007:
@snicoll 
I should have to add the tests and update the comments based on the feedback w/in the next 12 hours.

bono007:
@snicoll - I added the auto-config tests for each exporter and simplified the java doc.

bono007:
Is the failure around `RestAssuredRestDocsAutoConfigurationAdvancedConfigurationIntegrationTests` expected? Looks like its getting a 401 and when i hit the endpoint that its trying to hit I see a form based login screen appear. I don't think anything in this merge request could affect that.

snicoll:
Same. Don't worry about it, I'll investigate as part of polishing this change.

snicoll:
> I don't think anything in this merge request could affect that.

Actually, it does indirectly as the actuator (and it's auto-configuration) is now added to the project. I don't know why that happens but on the other hand Actuator does not seem required. I'll investigate a bit more.

bono007:
Ahh interesting. Sounds like security is kicking in now with the added deps.

bono007:
@snicoll did you get a chance to follow up on the issue? If not, I can take a look in the next 24hrs.

snicoll:
I've removed actuator as it's not required although I might change my mind and write a test with an actual registry implementation to validate it's auto-configuration isn't enabled. I am on PTO the next 2 days and I've already started to polish things quite a bit in a branch. If you want to investigate how to prevent security to kick in the way it does, I am definitely interested. Thanks!

bono007:
Enjoy your PTO! I will look into how to prevent security from kicking in as
it will be good to know, regardless of final direction.

bono007:
I just got around to look into this now. 

1. Yes, you are correct, we do not need those dependencies in there. I believe I had them in there as I was thinking initially of a heavier integration test - however ended up just verifying the expected properties were set etc. I forgot to remove the deps once I switched gears on that. 

2. The security is kicking in due to `ManagementWebSecurityAutoConfiguration` being available on classpath now. If you look at the [RestDocsTestApplication](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/restdocs/RestDocsTestApplication.java#L27) used by the IT you will notice it was already excluding SecurityAutoConfiguration for this same reason. The test goes green when we include the `ManagementWebSecurityAutoConfiguration` in the exclude: 
```
@SpringBootApplication(exclude = { SecurityAutoConfiguration.class, ManagementWebSecurityAutoConfiguration.class })
```

snicoll:
@bono007 thank you very much for making your first contribution to Spring Boot. I've polished your contribution in 
3530ac9.

bono007:
Ha! Its not my first contribution to Spring Boot. But thank you for the warm welcome ;)

wilkinsona:
This is your third first contribution no less 😬. Sorry about that. IIRC, GitHub kept listing you as a first-time contributor for a while. That's no longer the case so hopefully this will be your last first contribution.

snicoll:
I know something was off when I saw the badge but I decided to go with the warm welcome anyway. Thank you for making your third first contribution, indeed 😂

bono007:
3rd times a charm - so they say. ;)

PR #21673, opened 2020-06-03
Polish Slack notification when no build scan was published

### LINKED ISSUE #21666: Polish Slack notification for a build failure that did not generate a build scan
It looks like this at the moment:

```
spring-boot-2.3.x jdk14-build failed! [build log] [<(build scan not generated)|build scan>]
```

It should look something like this:

```
spring-boot-2.3.x jdk14-build failed! [build log] [build scan not generated]
```

dreis2211:
Hi,

this PR should close #21666 by moving the link to the generated file that is later read when sending the Slack notification.

I have no way of testing this really, so forgive me if things are not working as expected.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211. We don't really have a good way of testing this either. FWIW, what you've done is what I had in mind. Let's merge this one and see what happens.

PR #21689, opened 2020-06-04
Add autoconfiguration for r2dbc-proxy

ttddyy:
Add autoconfiguration for r2dbc-proxy.

Create a proxy of `ConnectionFactory` when:
- r2dbc-proxy is in classpath
- property is enabled (enabled by default)
- r2dbc-proxy listener beans are defined


snicoll:
@ttddyy thanks for the PR, can you share a bit more context that lead you to contribute this?

Paging @mp911de for insights.

ttddyy:
Hi @snicoll 

[`r2dbc-proxy`](https://github.com/r2dbc/r2dbc-proxy) is a library that I wrote to provide callback mechanism to `r2dbc-spi` classes. It is part of the [r2dbc umbrella](https://github.com/r2dbc). One of the usage is observability story for r2dbc, agnostic to actual r2dbc-driver implementations.

`r2dbc-proxy` creates a proxy of `ConnectionFactory` with listeners.
There are two listener interfaces - [ProxyExecutionListener](https://github.com/r2dbc/r2dbc-proxy/blob/master/src/main/java/io/r2dbc/proxy/listener/ProxyExecutionListener.java) and [LifeCycleListener](https://github.com/r2dbc/r2dbc-proxy/blob/master/src/main/java/io/r2dbc/proxy/listener/LifeCycleListener.java).
Listeners are called back with contextual information when `Connection`, `Statement`, `Batch`, etc are invoked.

For example, `r2dbc-proxy` can be used for [getting metrics](https://github.com/ttddyy/r2dbc-proxy-examples/blob/master/listener-example/src/main/java/io/r2dbc/examples/MetricsExecutionListener.java), [creating traces](https://github.com/ttddyy/r2dbc-proxy-examples/blob/master/listener-example/src/main/java/io/r2dbc/examples/TracingExecutionListener.java), [logging query executions](https://github.com/r2dbc/r2dbc-proxy#query-logging-1), etc.

The PR is for when those listeners are defined, automatically create a `ConnectionFactory` proxy and register defined listener beans.
This will alleviate manual proxy creation and listener registration.
If, for example, `spring-cloud-sleuth` implements tracing for r2dbc (https://github.com/spring-cloud/spring-cloud-sleuth/issues/1524), they can create a listener bean and this autoconfiguration can pick up such listener and registers to the proxy.

mp911de:
@ttddyy from my perspective, the proxy integration would make a lot more sense within Spring Cloud Sleuth. 

There we could add tracing and metrics support. Likely, it doesn’t make sense to add integrations with the same tool in two places.

ttddyy:
@mp911de From my view, usage of proxy is more generic in general.
It can be used for metrics and tracing, but also used for different scenarios such as slow query detection, query logging, debugging, etc.

mp911de:
The issue we have right now with this PR is that there is too much abstraction without a clear use-case. R2DBC Proxy adds value to the setup, but we need to look also from a perspective of what the integration provides and how it is built. 

Providing metrics support out of the box is desirable and likely we should start with metrics as use-case for R2DBC Proxy in Spring Boot. Over time, we can consider additional ones such as general listener support.

Is there already a Micrometer integration for R2DBC Proxy available? On a related note, there's also a [ticket in Micrometer](https://github.com/micrometer-metrics/micrometer/issues/635) asking for SQL metrics.

ttddyy:
[This r2dbc-proxy example](https://github.com/ttddyy/r2dbc-proxy-examples/blob/master/listener-example/src/main/java/io/r2dbc/examples/MetricsExecutionListener.java) populates micrometer metrics a bit.

So, if we want to start with metrics around r2dbc calls, I can write one.
We can start what to measure based on the [ticket in Micrometer](https://github.com/micrometer-metrics/micrometer/issues/635), or let me know if you have list of things to measure.

ttddyy:
I wrote a sample listener that populate metrics for query execution time per query type(read/write) based on the [ticket in Micrometer](https://github.com/micrometer-metrics/micrometer/issues/635).

https://github.com/ttddyy/r2dbc-proxy-examples/blob/master/listener-example/src/main/java/io/r2dbc/examples/QueryTimeMetricsExecutionListener.java

marcingrzejszczak:
Hey @ttddyy , it's Marcin from Spring Cloud Sleuth. How about we move this PR (or part of it) to Spring Cloud Sleuth? We could add an instrumentation to `org.springframework.cloud.sleuth.instrument.r2dbc` in package scope. Once it turns out that it's working fine and people use it with success, then we could most likely consider pushing it to Boot. WDYT?

wilkinsona:
Thanks, @marcingrzejszczak. I'm not sure that this needs to go into Sleuth. The Boot team discussed this one recently and, IIRC, the conclusion was that we didn't have any objection to the underlying infrastructure going into Boot and @snicoll was going to have another chat with @mp911de to see if we'd missed something. He's on PTO at the moment so we'll need to wait a bit until he's back.

mp911de:
From a tracing perspective, it makes sense to have a listener implementation for `r2dbc-proxy` in Sleuth that can contribute spans based on R2DBC activity once this PR is merged.

ttddyy:
Hi @marcingrzejszczak

I wrote a sample listener for tracing [here](https://github.com/ttddyy/r2dbc-proxy-examples/blob/master/listener-example/src/main/java/io/r2dbc/examples/TracingExecutionListener.java) long time ago.
Looking at it now, I can cleanup a bit.  I'll find a time to update it as for a sample listener impl for tracing.

marcingrzejszczak:
From my perspective, as a Spring Cloud Sleuth user, the easiest way to work with this would be the following

- add an optional r2dbc proxy
- create a bean of Listener type
- assume that Spring Boot would add my listener to the proxy

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/R2dbcProxyAutoConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+ * @since 2.3.1
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnClass(ProxyConnectionFactory.class)
+@AutoConfigureBefore(R2dbcDataAutoConfiguration.class)
+@AutoConfigureAfter(R2dbcAutoConfiguration.class)
+public class R2dbcProxyAutoConfiguration {
+
+	@Bean
+	@Primary
```
snicoll, line 54: We can't do this. Having to inject the `ConnectionFactory` and then exporting again is a smell and should be avoided, in particular in auto-configurations.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/R2dbcProxyAutoConfiguration.java
```diff
@@ -0,0 +1,88 @@
...
+			ObjectProvider<ProxyExecutionListener> proxyExecutionListeners,
+			ObjectProvider<LifeCycleListener> lifeCycleListeners) {
+		ProxyConfig proxyConfig = proxyConfigProvider.getIfAvailable(ProxyConfig::new);
+		proxyExecutionListeners.orderedStream().forEach(proxyConfig::addListener);
+		lifeCycleListeners.orderedStream()
+				.forEach((listener) -> proxyConfig.addListener(LifeCycleExecutionListener.of(listener)));
+		return ProxyConnectionFactory.builder(connectionFactory, proxyConfig).build();
+	}
+
+	static final class ProxyConnectionFactoryCondition extends AnyNestedCondition {
```
snicoll, line 70: This condition essentially states that we want to create a proxy if at least a relevant listener is around. That sounds like something that we should do in `ConnectionFactoryBuilder`. 

The fact that we have to return a different instance is a bit unusual for a customizer but something that "post-processes" the `ConnectionFactory` to create the proxy is what I have in mind. We want to make sure we only post-process the bean that has been created by auto-configuration so using a `BeanPostProcessor` might not be the best option. Modifying `ConnectionFactoryConfigurations#createConnectionFactory` could be an option.

spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -1468,6 +1468,11 @@
       "type": "java.lang.Boolean",
       "description": "Whether pooling is enabled. Enabled automatically if \"r2dbc-pool\" is on the classpath."
     },
+    {
+      "name": "spring.r2dbc.proxy.enabled",
+      "type": "java.lang.Boolean",
+      "description": "Whether to proxy ConnectionFactory. Enabled automatically if \"r2dbc-proxy\" is on the classpath."
```
snicoll, line 1474: That description does not seem accurate. "Whether to proxy the ConnectionFactory if necessary"  with a `defaultValue: true` sounds a better option to me.

ttddyy:
@snicoll Ok, I'll explore the path for modifying `ConnectionFactoryConfigurations#createConnectionFactory` with `ConnectionFactoryOptions`.
`ProxyConnectionFactoryProvider`(`ConnectionFactoryProvider` impl in `r2dbc-proxy`) understands proxy-listener options.
So, if proxy-listener beans are defined, pass them as `ConnectionFactoryOptions`, then `ConnectionFactories#find()` would delegate to `ProxyConnectionFactoryProvider` for actual `ProxyConnectionFactory` creation with those proxy-listener beans. 
I think this way is better since it would make more use of r2dbc-spi infrastructure.

ttddyy:
Hi @snicoll 
I added a commit 9cccc61004fa4616a02d32b3572149675f3f329b.

This uses `ConnectionFactoryOptionsInitializer`(via `ConnectionFactoryConfigurations#createConnectionFactory`) to construct `ConnectionFactory` options used by `ProxyConnectionFactoryProvider`(`r2bc-proxy`).
Also, added `ProxyListenersProvider` interface to avoid direct reference to the classes in `r2dbc-proxy` to make it work when `r2dbc-proxy` is not available.

If this looks good, I'll cleanup the commits with tests.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryBuilder.java
```diff
@@ -16,17 +16,21 @@
 
 package org.springframework.boot.autoconfigure.r2dbc;
 
+import java.util.List;
 import java.util.function.Consumer;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
+import io.r2dbc.proxy.ProxyConnectionFactoryProvider;
+import io.r2dbc.proxy.listener.ProxyExecutionListener;
```
snicoll, line 25: We can't really do that, can we? `ConnectionFactoryBuilder` should only rely on the SPI.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryBuilder.java
```diff
@@ -58,8 +62,26 @@ private ConnectionFactoryBuilder(ConnectionFactoryOptions.Builder optionsBuilder
...
+	 * {@link EmbeddedDatabaseConnection}
+	 * @param proxyListenersProvider a supplier for list of {@link ProxyExecutionListener}
+	 * @return a new builder initialized with the settings defined in
+	 * {@link R2dbcProperties}
+	 */
+	public static ConnectionFactoryBuilder of(R2dbcProperties properties,
+			Supplier<EmbeddedDatabaseConnection> embeddedDatabaseConnection,
+			ProxyListenersProvider proxyListenersProvider) {
+		return new ConnectionFactoryBuilder(new ConnectionFactoryOptionsInitializer().initializeOptions(properties,
+				embeddedDatabaseConnection, proxyListenersProvider));
```
snicoll, line 84: I wouldn't surface this so high. If post-processing the `ConnectionFactory` and returning another instance is something that's required by `r2dbc-proxy`, perhaps the builder should expose such hook-point and offer a method on the builder to register such implementation of that hook-point.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryBuilder.java
```diff
@@ -222,6 +251,22 @@ private String determineEmbeddedUsername(R2dbcProperties properties) {
...
+			}
+			String driver = urlOptions.getRequiredValue(ConnectionFactoryOptions.DRIVER);
+			if (!ProxyConnectionFactoryProvider.PROXY_DRIVER.equals(driver)) {
+				optionsBuilder.option(ConnectionFactoryOptions.DRIVER, ProxyConnectionFactoryProvider.PROXY_DRIVER);
+				String protocol = urlOptions.getValue(ConnectionFactoryOptions.PROTOCOL);
+				String newProtocol = StringUtils.isEmpty(protocol) ? driver : driver + ":" + protocol;
+				optionsBuilder.option(ConnectionFactoryOptions.PROTOCOL, newProtocol);
+			}
+			optionsBuilder.option(ProxyConnectionFactoryProvider.PROXY_LISTENERS, proxyListeners);
+		}
```
snicoll, line 268: I am a bit confused what's going on there but it looks like the proxy can be created without having to wrap the `ConnectionFactory` and return it. In that case, I think the hook-point should provide you a way to customize the builder the way you're doing here. The fact the customization is actually registering proxy-related options should be an implementation detail.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
```diff
@@ -86,13 +97,43 @@ ConnectionPool connectionFactory(R2dbcProperties properties, ResourceLoader reso
...
+					proxyListenersProviders.getIfAvailable(() -> ProxyListenersProvider.EMPTY));
 		}
 
 	}
 
+	@Configuration(proxyBeanMethods = false)
+	@ConditionalOnClass(ProxyConnectionFactory.class)
+	@ConditionalOnProperty(prefix = "spring.r2dbc.proxy", value = "enabled", havingValue = "true",
+			matchIfMissing = true)
+	static class Proxy {
```
snicoll, line 113: It would be nice if it didn't have to surface that way. this should register a  customizer implementation that does the necessary thing. And if no listeners are present then the customizer does nothing.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryBuilder.java
ttddyy: What here is doing is reconstructing `DRIVER` and `PROTOCOL` options.

e.g: `r2dbc:h2:mem://...` to `r2dbc:proxy:h2:mem://...`

| - | DRIVER | PROTOCOL |
| --- | --- | --- |
| From | h2 | mem |
| To| proxy | h2:mem |

This way, the discovery mechanism will create `ConnectionFactory` for proxy and mysql.

Actually, I think same way may be used for `r2dbc-pool` via `PoolingConnectionFactoryProvider` as well.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryBuilder.java
ttddyy: ok, I made not to throw `NoClassDefFoundError` but I'll go some other way not to use them here.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
ttddyy: Currently, `ConnectionFactoryOptionsBuilderCustomizer` provides a way to customize `Option`s on `ConnectionFactoryBuilder`. So, I could use it as a hook point to add/replace some of the `Option`s.

I wasn't sure it is ok to use customizer facility within Spring Boot, but if that's ok, then here can define a customizer bean (probably with `@Order` to allow users to run own customizers before and after). Then, it will be pretty much all we need for `r2dbc-proxy`.

mp911de:
After discussing with @snicoll we found that ordering of pool/proxy/driver wrapping matters (speaking in URL syntax, `r2dbc:pool:proxy:h2` vs. `r2dbc:proxy:pool:h2`) if someone wants to e.g. measure how many real connections were created/closed by the pool vs. how long a connection is in use by the application.

ttddyy:
> After discussing with @snicoll we found that ordering of pool/proxy/driver wrapping matters (speaking in URL syntax, `r2dbc:pool:proxy:h2` vs. `r2dbc:proxy:pool:h2`) if someone wants to e.g. measure how many real connections were created/closed by the pool vs. how long a connection is in use by the application.

Yes, the ordering matters, and it can be even `r2dbc:proxy:pool:proxy:h2`.(though, there is a problem specifying listeners for which proxy to apply since there is only one `proxyListener` option)
I consider most of the case `proxy` is the outer most wrapper since you want to check the Application behavior.
If a user wants to proxy the middle `ConnectionFactory`, the user needs to wrap it manually, but I think that is more advanced use case. 
For connection-pool in Spring Boot, it may provide a flag in connection-pool settings which would be specific to proxying the connection-pool with some canned proxy listener such as metrics.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
snicoll: Yeah I think that's ok.

snicoll:
@ttddyy I can see a WIP in your last update and the tests have been commented out. Do you need something from us? Thanks a lot for your effort.

ttddyy:
> 
> 
> @ttddyy I can see a WIP in your last update and the tests have been commented out. Do you need something from us? Thanks a lot for your effort.

@snicoll Thanks for checking. I wanted to check whether the current impl with customizer is ok before I proceed. And since you said ok, I'll finish up the rest of the PR.

snicoll:
Thanks. Devil in the details as always but I feel we've iterated enough and I can take it from there once you've restored the tests. Thanks again!

ttddyy:
@snicoll  While writing tests, I found using `Option` has a problem wrapping `ConnectionPool` with proxy. This is because `@Bean` method for `ConnectionPool` wraps the **result** of `createConnectionFactory()`. Therefore, proxy cannot wrap the `ConnectionPool`.
So, I fallback to the original plan to make a hook on `createConnectionFactory()` to post process configured `ConnectionFactory`. (which I also applied to creating `ConnectionPool`)

jarias:
Hi all, I was able to use the proxy with spring-boot 2.3.5 just by adding the dep in the class path and the following props:

```
spring:
  r2dbc:
    properties:
      driver: proxy
      protocol: postgresql
      proxyListener: test.foo.bar.MyListener
```

ttddyy:
@jarias Looks good. It uses ConnectionFactory Discovery mechanism which is a configuration centric approach. The shortcoming of this approach is that the `proxyListener` is created outside of the spring and never be a part of the spring context.

In this PR, I am integrating more to spring-boot. So that, proxy listener can be a bean and allowed to be dependency injected.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/r2dbc/ConnectionFactoryConfigurations.java
ttddyy: I still use this `Proxy` configuration class but now it has only single processor(`Function`) bean that is used by `ConnectionFactoryConfigurations#createConnectionFactory`.

In the latest impl, the conditions to apply proxy are:
- `spring.r2dbc.proxy.enabled` is `true`
- `spring.r2dbc.url` does not contain `:proxy:`
- `ProxyExecutionListener` beans OR a `ProxyConfig` bean exist


I have explored the option to use `ConnectionFactoryOptionsBuilderCustomizer` but the connection factory `Option` is limited to the use of connection factory discovery mechanism.

Currently, similar to auto configuration for `r2dbc-pool`, auto config for `r2dbc-proxy` does not modify connection-url. When `spring.r2dbc.url` contains `proxy` prefix, it will backoff the auto configuration. (same as pool)
So options will not be picked up even if they are customized by auto configuration.

ttddyy (commented):
I have modified `ConnectionFactoryConfigurations#createConnectionFactory` to take `Function<ConnectionFactory, ConnectionFactory>`. Pool and proxy pass this function to wrap the original `ConnectionFactory` if condition matches.

@snicoll Can you review the change?

snicoll:
In the meantime, Spring Cloud Sleuth has added support for R2DBC instrumentation which was the primary driver for this PR on our side. We haven't really found a way to find an arrangement that would be generic enough to warrant support here so I am going to close this now. We can reconsider based on the experience in Spring Cloud Sleuth or other changes in `r2dbc-proxy`. Thanks again @ttddyy!

PR #21702, opened 2020-06-04
Fix BasicErrorController to use binding error cfg

Aurdo:
`server.error.include-binding-errors` had no effect on BasicErrorController.isIncludeBindingErrors(), due to wrong property being used

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/BasicErrorController.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/BasicErrorController.java
@@ -172,7 +172,7 @@ public class BasicErrorController extends AbstractErrorController {
 	 * @return if the errors attribute should be included
 	 */
 	protected boolean isIncludeBindingErrors(HttpServletRequest request, MediaType produces) {
-		switch (getErrorProperties().getIncludeMessage()) {
+		switch (getErrorProperties().getIncludeBindingErrors()) {
 		case ALWAYS:
 			return true;
 		case ON_PARAM:


pivotal-issuemaster:
@Aurdo Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21702)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21702) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Aurdo Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21702)!

scottfrederick:
Good catch @Aurdo, thanks. This was my copy-and-paste mistake that happened when the control of the `message` and `errors` fields were split from one property into two. 

The reason this wasn't caught earlier is that the tests to verify the behavior of `BasicErrorController` [always test the `server.error.include-message` and `server.error.include-binding-errors` fields together](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/servlet/error/BasicErrorControllerIntegrationTests.java#L207-L240). Those tests should be split up so that `include-message` and `include-binding-errors` are tested separately. 

If you'd like, you can also modify `BasicErrorControllerIntegrationTests` to split up those tests and add that change in another commit to this PR. If you'd prefer not to do that, I can modify the tests when the PR is merged.

Aurdo:
> Good catch @Aurdo, thanks. This was my copy-and-paste mistake that happened when the control of the `message` and `errors` fields were split from one property into two.
> 
> The reason this wasn't caught earlier is that the tests to verify the behavior of `BasicErrorController` [always test the `server.error.include-message` and `server.error.include-binding-errors` fields together](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/servlet/error/BasicErrorControllerIntegrationTests.java#L207-L240). Those tests should be split up so that `include-message` and `include-binding-errors` are tested separately.
> 
> If you'd like, you can also modify `BasicErrorControllerIntegrationTests` to split up those tests and add that change in another commit to this PR. If you'd prefer not to do that, I can modify the tests when the PR is merged.

@scottfrederick whatever works for you is fine with me. Just not sure how long would it take to clone and import project locally

scottfrederick:
@Aurdo If you haven't cloned, imported, and built the project already don't worry about it. I'll fix up the tests.

PR #21713, opened 2020-06-05
Add Java 15 CI

dreis2211:
Hi,

this PR tackles #21604 as discussed. Fortunately, even the automatic JDK upgrade should already work.

Before merging this, #21605 needs to be merged. Apart from that I haven't seen any test failures when I run with the custom `buildJavaHome` property being set to my local JDK 15 installation.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/reactive/HandlerFunctionDescription.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/reactive/HandlerFunctionDescription.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,11 +29,17 @@ public class HandlerFunctionDescription {
 	private final String className;
 
 	HandlerFunctionDescription(HandlerFunction<?> handlerFunction) {
-		this.className = handlerFunction.getClass().getName();
+		this.className = getHandlerFunctionClassName(handlerFunction);
 	}
 
 	public String getClassName() {
 		return this.className;
+	}
+
+	private String getHandlerFunctionClassName(HandlerFunction<?> handlerFunction) {
+		Class<?> functionClass = handlerFunction.getClass();
+		String canonicalName = functionClass.getCanonicalName();
+		return canonicalName != null ? canonicalName : functionClass.getName();
 	}
 
 }
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/reactive/HandlerFunctionDescription.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/reactive/HandlerFunctionDescription.java
@@ -39,7 +39,7 @@ public class HandlerFunctionDescription {
 	private String getHandlerFunctionClassName(HandlerFunction<?> handlerFunction) {
 		Class<?> functionClass = handlerFunction.getClass();
 		String canonicalName = functionClass.getCanonicalName();
-		return canonicalName != null ? canonicalName : functionClass.getName();
+		return (canonicalName != null) ? canonicalName : functionClass.getName();
 	}
 
 }
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/reactive/HandlerFunctionDescription.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/reactive/HandlerFunctionDescription.java
@@ -29,7 +29,7 @@ public class HandlerFunctionDescription {
 	private final String className;
 
 	HandlerFunctionDescription(HandlerFunction<?> handlerFunction) {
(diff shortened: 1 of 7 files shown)


snicoll:
Thanks for the PR @dreis2211 

> Apart from that I haven't seen any test failures when I run with the custom buildJavaHome property being set to my local JDK 15 installation.

This was required when Gradle does not support the target JDK which does not seem to be case with Java15 at this point. I don't see this PR using that feature either so I am a bit confused.

Switching my local JDK to Java 15ea and building `master` leads to some failures. Do you have time to look at those?

dreis2211:
@snicoll You're absolutely right. I was under the impression that I'm getting the same results with the custom home property being set.
Let me try with your approach.

dreis2211:
@snicoll It would have also worked with the custom java home being set, but my tasks were cached and thus not all executed.

I see one additional failure in `SampleAntApplicationIT.runJar` + the already mentioned multi-release JAR test that is covered by #21605 .

Does that match your local build?

snicoll:
No, it does not. I have a compilation failure:

```
> Task :spring-boot-project:spring-boot-tools:spring-boot-loader:compileJava FAILED
/Users/snicoll/workspace/work/spring-boot/master/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/StringSequence.java:75: error: isEmpty() in StringSequence cannot implement isEmpty() in CharSequence
        boolean isEmpty() {
```

and one test failure before I gave up:

```
MappingsEndpointReactiveDocumentationTests > mappings() FAILED
    org.springframework.restdocs.payload.FieldTypesDoNotMatchException at MappingsEndpointReactiveDocumentationTests.java:126
```

This in a branch of your first PR I am about to merge. Maybe something is cached and shouldn't be?


```
$ java -version
openjdk version "15-ea" 2020-09-15
OpenJDK Runtime Environment (build 15-ea+26-1287)
OpenJDK 64-Bit Server VM (build 15-ea+26-1287, mixed mode, sharing)
```

dreis2211:
```
openjdk version "15-ea" 2020-09-15
OpenJDK Runtime Environment (build 15-ea+17-717)
OpenJDK 64-Bit Server VM (build 15-ea+17-717, mixed mode, sharing)
```
I guess the first one can be explained by `CharSequence.isEmpty` being only available in later builds. I'll try the latest JDK 15 build then.

dreis2211:
I've pushed two additional commits that fix the syntax error and the failure in `MappingsEndpointReactiveDocumentationTests`

dreis2211:
I see some intermittent failures for tests that all seem to be related to timing. I wonder if https://bugs.openjdk.java.net/browse/JDK-8242504 is related as this got introduced recently as well.

dreis2211:
I've also seen some flaky tests that are annotated with `@WebEndpointTest` that prematurely close the connection.

dreis2211:
`DevToolsIntegrationTests > createAControllerAndThenAddARequestMapping` seems to also fail on JDK 8 occasionally.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/web/mappings/reactive/HandlerFunctionDescription.java
```diff
@@ -29,7 +29,7 @@
 	private final String className;
 
 	HandlerFunctionDescription(HandlerFunction<?> handlerFunction) {
-		this.className = handlerFunction.getClass().getCanonicalName();
+		this.className = handlerFunction.getClass().getName();
```
snicoll, line 32: I don't think we should do this. Perhaps as a fallback if `getCanonicalName` is empty? It's a change of format so it's a breaking change. 

Or perhaps we could adapt the test? I haven't looked in details yet so thinking out loud.

dreis2211: I wrote to the core-libs mailing list a week ago, I think, and got this reply.

> A hidden class has no canonical name [1] and it can't be nominally referenced [...] For diagnosiability, Class::getName or Class::toString can be used.

As `HandlerFunction` is often a lambda, it will probably often return null on JDK 15 at least. I  don't have a problem with the fallback idea. In reality, there will be only a change for inner classes though.  See the following example output from JDK 14 of canonical vs name:
```
com.example.demo.Example$$Lambda$14/0x0000000800b66c40
com.example.demo.Example$$Lambda$14/0x0000000800b66c40
com.example.demo.AFunction
com.example.demo.AFunction
com.example.demo.Example.InnerFunction
com.example.demo.Example$InnerFunction
```

Give me a minute to implement the fallback.

dreis2211: Done

dreis2211:
Trying with 15-ea-27 introduced a change for missing Javadocs: https://bugs.openjdk.java.net/browse/JDK-8242607

That produces quite a lot of warnings like the following:
```
/Users/christoph.dreis/projects/spring-boot/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/transport/DockerEngineException.java:35: warning: no comment
        private final String reasonPhrase;
```
It's not breaking the builds, so I guess it's fine for this PR, but probably needs to be tackled in another one.

snicoll:
Ah ah. I was building locally and I've already polished that checkstyle issue here Christoph.

dreis2211:
Sorry - I've noticed this one just now. :/

Issue #20892, opened 2020-04-09
Can't use convert annotations on Kotlin data class ConfigurationProperties


arustleund:
Consider the following Kotlin data class for holding ConfigurationProperties:

```kotlin
@ConstructorBinding
@ConfigurationProperties
data class MyProps(

  @DataSizeUnit(DataUnit.MEGABYTES)
  val dataLimit: DataSize

)
```

Since the convert annotations such as `@DataSizeUnit` are field level annotations, and the `ValueObjectBinder` class is only looking for constructor parameter annotations, these convert annotations will not be processed and configuration data will not be converted properly. No error or warning is issued in the logs and the user is left to debug through Spring code for hours to find the issue! 😄 
Possible solutions: 
1. Search field level annotations on Kotlin data classes as well as constructor annotations.
2. Allow convert annotations to be placed on constructor parameters

encircled:
Hi,

As of now, `@ConfigurationProperties` support injection via setter only, regardless of whether it is written in Kotlin or Java. Looks like a good improvement to me, I will prepare a PR.

snicoll:
> As of now, @ConfigurationProperties support injection via setter only, regardless of whether it is written in Kotlin or Java

@encircled that's not accurate. We have constructor injection for both Kotlin and Java.

encircled:
Right, I forgot about `ConstructorBinding`. Let me check it again

mbhave:
Closing in favor of PR #21746.

PR #21746, opened 2020-06-07
Allow specifying unit type of configuration property when injected via constructor

encircled:
Allow specifying unit type of configuration property when injected via constructor #20892

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/DataSizeUnit.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/DataSizeUnit.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


snicoll:
Before merging this we should double check that the metadata is generated with the correct unit.

wilkinsona:
The unit seems to be missing from the metadata. For example:

```json
{
  "name": "test.size",
  "type": "org.springframework.util.unit.DataSize",
  "sourceType": "com.example.demo.ConstructorParameterWithConversionProperties",
  "defaultValue": "3"
}
```

The above is from a properties class based on the one that's added to the tests in this PR. The constructor argument is declared as `@DefaultValue("3") @DataSizeUnit(DataUnit.MEGABYTES) DataSize size` so I believe the default value should be 3MB.

snicoll:
I would have expected that indeed. We need to update this support to lookup the unit annotation on others places.

snicoll:
Looking at the metadata and how values are extracted, this applies using the code that's written for the field initialization something like, `Duration someProperty = Duration.ofHours(40);`. We can extract `40h` from that. If we compare with the constructor based approach, we should provide the value as-is IMO as `@DefaultValue` is the text-based default value of the property. I think it should be `@DefaultValue("3MB")` rather than `@DefaultValue("3") @DataSizeUnit(DataUnit.MEGABYTES)`.

Allowing the unit on the constructor parameter makes such construct possible so it'd be legitimate for users to expect the other variant to work. It means that the annotation processor has to look for those 3 additional annotations too and handle it only if the default value is an integer. I wonder if we want that extra complexity.

snicoll:
We've discussed this and concluded that having `3` in the metadata is acceptable considering that using `3` manually would lead to the expected result (`3MB`). To provider better metadata, the unit can be specified again in the default value so that the generated default value in the metadata is a bit more explicit.

PR #21758, opened 2020-06-08
Use Supplier assertion variants in AvailabilityStateHealthIndicator

dreis2211:
Hi,

I just noticed that `AvailabilityStateHealthIndicator` does assertions with string concats. Specifically, the one in `doHealthCheck` could be changed, because it's executed on every health check.

Cheers,
Christoph


philwebb:
Nice spot!

PR #21912, opened 2020-06-14
Fix snapshot deps resolution

eddumelendez:
As part of 8e8548756e, snapshot dependency resolution change from
`BUILD-SNAPSHOT` to `-SNAPSHOT`. Currently, if we add a version
dependency which is store in snapshots it is not resolved. This
commit allow to resolves plugins and dependencies.


wilkinsona:
Thanks very much, @eddumelendez. I tweaked your proposal a little bit to look for `-SNAPSHOT` rather than just `SNAPSHOT`.

PR #21913, opened 2020-06-14
Use Class.getDeclaredConstructor().newInstance()

dreis2211:
Hi,

this PR replaces the deprecated `Class.newInstance()` with `Class.getDeclaredConstructor().newInstance()`. While only deprecated since 9, it is already possible to change things on JDK 8.

Let me know what you think.
Cheers,
Christoph


wilkinsona:

Issue #21890, opened 2020-06-11
Micrometer Metrics for KafkaStreams


garyrussell:
Similar to https://github.com/spring-projects/spring-boot/commit/d16074d8fdc457107177bca026635e4575fe0c08#diff-2f3136f1958043f3d7c37de45bf062fa - consider adding a `KafkaStreamsMicrometerListener` to the default `StreamsBuilderFactoryBean`.

(When spring-kafka 2.5.3 is available - PR here https://github.com/spring-projects/spring-kafka/pull/1502)

snicoll:
Closing in favor of PR #21921

PR #21921, opened 2020-06-15
Kafka Stream consumer metrics were lost in the move away from KafkaConsumerMetrics

eddumelendez:
See gh-21890

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/StreamsBuilderFactoryBeanCustomizer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/StreamsBuilderFactoryBeanCustomizer.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.kafka;
+
+import org.springframework.kafka.config.StreamsBuilderFactoryBean;
+
+/**
+ * Callback interface for customizing {@code StreamsBuilderFactoryBean} beans.
+ *
+ * @author Eddú Meléndez
+ * @since 2.3.2
+ */
+@FunctionalInterface
+public interface StreamsBuilderFactoryBeanCustomizer {
+
+	/**
+	 * Customize the {@link StreamsBuilderFactoryBean}.
+	 * @param factoryBean the factory bean to customize
+	 */
+	void customize(StreamsBuilderFactoryBean factoryBean);
+
+}
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
@@ -67,6 +67,7 @@ dependencies {
 	optional("org.apache.activemq:activemq-broker")
 	optional("org.apache.commons:commons-dbcp2")
 	optional("org.apache.kafka:kafka-clients")
+	optional("org.apache.kafka:kafka-streams")
 	optional("org.apache.tomcat.embed:tomcat-embed-core")
 	optional("org.apache.tomcat.embed:tomcat-embed-el")
 	optional("org.apache.tomcat:tomcat-jdbc")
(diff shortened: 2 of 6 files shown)


snicoll:
Thanks a lot @eddumelendez. Considering that this was added in Spring Kafka `2.5` and this would complement our native metrics story for Kafka added in `2.3.0`, I wonder if merging this in `2.3.x` would be the right thing to do. I've flagged this for team attention to get more feedback from the team.

wilkinsona:
I think we should wait for 2.4.

garyrussell:
The problem is that the previous (deprecated) Micrometer code scraped JMX MBeans so it would automatically find consumer and producer MBeans, including those registered by Kafka streams.

The issue was reported [here](https://github.com/spring-cloud/spring-cloud-stream-binder-kafka/issues/914) - because the binder doesn't use Boot's producer/consumer factories.

When fixing that, I realized the same issue would arise for any Boot app that uses Streams.

eddumelendez:
Should I update the PR and point to `2.3.x`?

wilkinsona:
I think I understand this now. There's a loss in functionality from 2.2 as the streams-related metrics are gone. What is proposed here will bring them back. We think this is worth fixing in 2.3.

@eddumelendez Thanks for the offer, but there's no need to update the PR unless you particularly want to. We can take care of it as part of merging the change.

snicoll:
@garyrussell can you please let us know when you plan to release Spring Kafka `2.5.3`? We can switch to snapshot on `master` as we are releasing `2.4.0-M1` next week.

garyrussell:
@snicoll I have scheduled it for next Wednesday.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/KafkaStreamsMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+ *
+ * @author Eddú Meléndez
+ * @since 2.4.0
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureBefore(KafkaMetricsAutoConfiguration.class)
+@AutoConfigureAfter(MetricsAutoConfiguration.class)
+@ConditionalOnClass({ KafkaStreamsMetrics.class, StreamsBuilderFactoryBean.class })
+@ConditionalOnBean(MeterRegistry.class)
+public class KafkaStreamsMetricsAutoConfiguration {
```
snicoll, line 43: I don't think a separate auto-configuration class is required here. Please move that code to the existing `KafkaMetricsAutoConfiguration`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/KafkaStreamsMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,54 @@
...
+
+	@Bean
+	public StreamsBuilderFactoryBeanCustomizer kafkaProducerMetrics(MeterRegistry meterRegistry) {
+		return (factoryBean) -> addListener(factoryBean, meterRegistry);
+	}
+
+	private void addListener(StreamsBuilderFactoryBean factory, MeterRegistry meterRegistry) {
```
snicoll, line 50: This method is not necessary. We had that shortcut because of generics but this object doesn't have any. Just invoke the registration in the main `@Bean` method.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/spring.factories
```diff
@@ -40,6 +40,7 @@ org.springframework.boot.actuate.autoconfigure.management.ThreadDumpEndpointAuto
 org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegistryAutoConfiguration,\
 org.springframework.boot.actuate.autoconfigure.metrics.JvmMetricsAutoConfiguration,\
 org.springframework.boot.actuate.autoconfigure.metrics.KafkaMetricsAutoConfiguration,\
+org.springframework.boot.actuate.autoconfigure.metrics.KafkaStreamsMetricsAutoConfiguration,\
```
snicoll, line 43: This should be removed once the code has moved to the main auto-configuration.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/StreamsBuilderFactoryBeanCustomizer.java
```diff
@@ -0,0 +1,36 @@
...
+
+package org.springframework.boot.autoconfigure.kafka;
+
+import org.springframework.kafka.config.StreamsBuilderFactoryBean;
+
+/**
+ * Callback interface for customizing {@code StreamsBuilderFactoryBean} beans.
+ *
+ * @author Eddú Meléndez
+ * @since 2.4.0
```
snicoll, line 25: This should be `2.3.2`

eddumelendez:
@snicoll changes were applied and now the target branch is `2.3.x`. Once the suggestions is applied in `spring-kafka` I can submit additional changes.

garyrussell:
@snicoll @eddumelendez https://github.com/spring-projects/spring-kafka/pull/1515 is merged.

eddumelendez:
thanks for the notification @garyrussell ! I have introduced new changes in the PR

wilkinsona:

snicoll:
An inner class with an `AutoComfiguration` prefix seems a bit misleading to me so we should probably rename that, see https://github.com/spring-projects/spring-boot/blob/ecbc8ea2dfa308d6a2860c6490c80f36b64936cf/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/KafkaMetricsAutoConfiguration.java#L76

wilkinsona:
Well spotted, @snicoll. I've polished that in dfea2f4.

PR #21985, opened 2020-06-16
Do not start Jetty if auto-start is set to false

mschneid:
Currently, JettyWebServer's constructor argument autoStart did not
prevent the server from being started automatically during
initialization.
If the server is started with autoStart=false, it can not be started
manually.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyWebServer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyWebServer.java
@@ -132,9 +132,10 @@ public class JettyWebServer implements WebServer {
 					}
 
 				});
-				// Start the server so that the ServletContext is available
-				this.server.start();
 				this.server.setStopAtShutdown(false);
+				if (this.autoStart) {
+					start();
+				}
 			}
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@mschneid Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21985)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21985) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@mschneid Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21985)!

wilkinsona:
Thanks for the PR, @mschneid.

This appears to be related to https://github.com/spring-projects/spring-boot/issues/17073.

wilkinsona:
Thanks again for the pull request, @mschneid. Unfortunately, I don't think we can make this change without completing the design work that is required for https://github.com/spring-projects/spring-boot/issues/17073. We want to make the behaviour of all four servers consistent. I don't think we should change the behaviour of only one server when we don't know exactly what we want the consistent behaviour to be. Thanks anyway.

Issue #21873, opened 2020-06-10
Use Awaitility to address flakiness of LiveReloadServerTests


wilkinsona:
We're currently using Thread.sleep to allow enough time for a message to be processed. This fails occasionally if the message processing thread isn't scheduled quickly enough. We should use Awaitility instead.

tszmytka:
If no one else is working on this I would be happy to take over.

wilkinsona:
Yes please, @tszmytka. That'd be great.

wilkinsona:
Closing in favour of https://github.com/spring-projects/spring-boot/pull/21988.

PR #21988, opened 2020-06-16
21873 flaky tests

tszmytka:
This substitutes `Thread.sleep()` with calls to appropriate `Awaitility` methods as described in #21873


spring-boot-project/spring-boot-devtools/src/test/java/org/springframework/boot/devtools/livereload/LiveReloadServerTests.java
```diff
@@ -88,19 +89,14 @@ void servesLivereloadJs() throws Exception {
 	void triggerReload() throws Exception {
 		LiveReloadWebSocketHandler handler = connect();
 		this.server.triggerReload();
-		Thread.sleep(200);
-		this.server.stop();
```
wilkinsona, line 92: I think we need to keep this call to `stop()`.

spring-boot-project/spring-boot-devtools/src/test/java/org/springframework/boot/devtools/livereload/LiveReloadServerTests.java
```diff
@@ -88,19 +89,14 @@ void servesLivereloadJs() throws Exception {
...
+		await().atMost(Duration.ofSeconds(1)).until(handler::getMessages, msgs -> msgs.get(0).contains("http://livereload.com/protocols/official-7") && msgs.get(1).contains("command\":\"reload\""));
 	}
 
 	@Test
 	void pingPong() throws Exception {
 		LiveReloadWebSocketHandler handler = connect();
 		handler.sendMessage(new PingMessage());
-		Thread.sleep(200);
-		assertThat(handler.getPongCount()).isEqualTo(1);
-		this.server.stop();
```
wilkinsona, line 103: I think we need to keep this call to `stop()`.

spring-boot-project/spring-boot-devtools/src/test/java/org/springframework/boot/devtools/livereload/LiveReloadServerTests.java
tszmytka: The test class already has an `@AfterEach` which calls `this.server.stop();`.
I assumed the original author had written those 2 test cases first and only after that decided to have a general `@AfterEach` method (and maybe forgot to take the individual `stop()`s out).

I made sure that the tests run fine without them.
Do you think `stop()` should be kept nevertheless?

spring-boot-project/spring-boot-devtools/src/test/java/org/springframework/boot/devtools/livereload/LiveReloadServerTests.java
tszmytka: The same situation as [above](https://github.com/spring-projects/spring-boot/pull/21988#discussion_r441769817).

spring-boot-project/spring-boot-devtools/src/test/java/org/springframework/boot/devtools/livereload/LiveReloadServerTests.java
wilkinsona: Sorry, I'd missed that when reviewing the diffs in isolation. I agree there's no need to keep the `stop()` calls. To keep this change clean, I think it makes sense to remove the `stop()` calls separately. No need for you to do anything though. We can take care of it prior to merging this.

wilkinsona:
Thanks very much, @tszmytka.

PR #21995, opened 2020-06-17
remove useless param when define function getCandidateConfigurations

niubilityCode:
Hi,

this tiny PR remove useless param when define function getCandidateConfigurations in class AutoConfigurationImportSelector & ImportAutoConfigurationImportSelector

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationImportSelector.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationImportSelector.java
@@ -174,7 +174,7 @@ public class AutoConfigurationImportSelector implements DeferredImportSelector, 
 	 * attributes}
 	 * @return a list of candidate configurations
 	 */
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@niubilityCode Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21995)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21995) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@niubilityCode Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=21995)!

snicoll:
Thanks for the PR. While technically the argument is unused, it is defined in a `protected` method so an extension of this class may actually use it. It is technically a breaking change, too. 

Flagging for team attention to see if someone else on the team remembers why the parameter was added.

philwebb:
I think it was added because obtaining the attributes might be expensive and we don't want subclasses to do it unnecessarily. Given it would be a breaking change, I don't think we should merge this PR.

Thanks anyway.

PR #22018, opened 2020-06-19
Remove Javadoc throw for ConfigurationPropertyName.ofIfValid()

izeye:
This PR removes Javadoc `@throw` for `ConfigurationPropertyName.ofIfValid()` as it doesn't seem to throw it.


snicoll:
Good catch, thanks @izeye

PR #22022, opened 2020-06-19
Management health group does not inherit show-details property

polarbear567:


philwebb:
Thanks very much for the PR @polarbear567, this has now been merged to 2.2.x, 2.3.x and master. I made one small tweak in commit e6eb02603c3dd26d127057ce203b5638c2273074 which should helpfully provide correct JSON metadata for IDEs.

PR #22040, opened 2020-06-20
FieldValues test class would cause infinite recursive loop

Konrado85:
We expect the initial field value for filed "unknownArray" to be null. At least that's what the test says:

AbstractFieldValuesProcessorTests

An attempt to create a FieldValues object, ends up calling an endless loop.


philwebb:
Thanks @Konrado85, I made a small change and merged this into 2.2.x

PR #22093, opened 2020-06-24
Disable XML bean definition loading when spring.xml.ignore is true

sdeleuze:
This commit allows to set the `XmlBeanDefinitionReader` field from `BeanDefinitionLoader` to `null` in a way that allows the GraalVM native compiler to remove it from the native image when the `spring.xml.ignore` flag introduced by spring-projects/spring-framework#25151 is set to `true`. The Groovy reader which relies on the XML one is disabled as well when the `spring.xml.ignore flag` is set to `true`.

The purpose of this change is to allow smaller footprint on native images without requiring to use GraalVM native substitutions which are unmaintainable by nature and also to increase the consistency between JVM and native images.

In order to effective, this optimization requires `BeanDefinitionLoader` class to be initialized at build time.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/BeanDefinitionLoader.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/BeanDefinitionLoader.java
@@ -32,6 +32,7 @@ import org.springframework.beans.factory.support.BeanNameGenerator;
 import org.springframework.beans.factory.xml.XmlBeanDefinitionReader;
 import org.springframework.context.annotation.AnnotatedBeanDefinitionReader;
 import org.springframework.context.annotation.ClassPathBeanDefinitionScanner;
+import org.springframework.core.SpringProperties;
 import org.springframework.core.env.ConfigurableEnvironment;
 import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.Resource;
@@ -53,9 +54,19 @@ import org.springframework.util.StringUtils;
  *
  * @author Phillip Webb
  * @author Vladislav Kisel
+ * @author Sebastien Deleuze
  * @see #setBeanNameGenerator(BeanNameGenerator)
  */
 class BeanDefinitionLoader {
+
+	/**
+	 * Boolean flag controlled by a {@code spring.xml.ignore} system property that
+	 * instructs Spring to ignore XML, i.e. to not initialize the XML-related
+	 * infrastructure.
+	 * <p>
+	 * By default XML support is enabled.
+	 */
+	private static final boolean IS_XML_ENABLED = !SpringProperties.getFlag("spring.xml.ignore");
 
 	private final Object[] sources;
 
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/BeanDefinitionLoader.java
```diff
@@ -80,9 +91,13 @@
 		Assert.notEmpty(sources, "Sources must not be empty");
 		this.sources = sources;
 		this.annotatedReader = new AnnotatedBeanDefinitionReader(registry);
-		this.xmlReader = new XmlBeanDefinitionReader(registry);
-		if (isGroovyPresent()) {
-			this.groovyReader = new GroovyBeanDefinitionReader(registry);
+		if (IS_XML_ENABLED) {
+			this.xmlReader = new XmlBeanDefinitionReader(registry);
+			this.groovyReader = isGroovyPresent() ? new GroovyBeanDefinitionReader(registry) : null;
```
wilkinsona, line 96: This seems too broad to me. I don't think that a `spring.xml.ignore` property should have an effect on whether or not `.groovy` bean definitions can be loaded.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/BeanDefinitionLoader.java
```diff
@@ -148,7 +169,7 @@ private int load(Object source) {
 	}
 
 	private int load(Class<?> source) {
-		if (isGroovyPresent() && GroovyBeanDefinitionSource.class.isAssignableFrom(source)) {
+		if (isGroovyPresent() && GroovyBeanDefinitionSource.class.isAssignableFrom(source) && IS_XML_ENABLED) {
```
dreis2211, line 172: I guess the check for `IS_XML_ENABLED` could be the first to avoid any class lookups inside `isGroovyPresent` here.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/BeanDefinitionLoader.java
sdeleuze: Indeed, I thought initially that the XML reader was a hard requirement for the Groovy one, but after a deeper look it seems to not be the case, I will update this PR accordingly.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/BeanDefinitionLoader.java
sdeleuze: Indeed, but based on @wilkinsona feedback, I am going to avoid mixing things between the XML and the Groovy readers, so this check will be removed.

sdeleuze:
I have updated this PR to focus only on disabling XML reader as suggested by @wilkinsona. I was initially confused by the fact that `private int load(GroovyBeanDefinitionSource source)` used both XML and Groovy readers, but since its return value seems to be not used I was able to remove this cross dependency. I have validated via `spring-graalvm-native` that the code removal is still effective. Please let me know if I missed something.

PR #22110, opened 2020-06-25
Improve error handling in JDK upgrade checks

### LINKED ISSUE #22098: Improve error handling in JDK upgrade checks

dreis2211:
Hi,

this is an attempt to fix #22098 for the time being with some more incremental steps (where possible) and `set -e` in order to step out as soon as we see an error to prevent unnecessary issues being opened.

In case we want to catch more errors in a more convenient way it might be a solution to craft a small java application that detects updates. Similar to the releasescript app. This could be also used for all the other detect jobs (docker, ubuntu...). Let me know if I should investigate this.

Cheers,
Christoph

### DIFF
--- a/ci/scripts/detect-jdk-updates.sh
+++ b/ci/scripts/detect-jdk-updates.sh
@@ -1,5 +1,11 @@
 #!/bin/bash
-set -e
+
+report_error() {
+	echo "Script exited with error $1 on line $2"
+	exit 1;
+}
+
+trap 'report_error $? $LINENO' ERR
 
 case "$JDK_VERSION" in
 	java8)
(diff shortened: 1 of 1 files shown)


ci/scripts/detect-jdk-updates.sh
```diff
@@ -1,4 +1,5 @@
 #!/bin/bash
+set -e
```
wilkinsona, line 2: This results in a silent failure if the `curl` command fails. I wonder if something like this might be better:

```bash
report_error() {
	echo "Error $1 on line $2"
	exit 1;
}

trap 'report_error $? $LINENO' ERR
```

dreis2211: I like it. Added it.

wilkinsona:
Thanks very much, @dreis2211.

PR #22112, opened 2020-06-25
Temporary files not cleaned up after Maven plugin execution

### LINKED ISSUE #22108: Temporary files not cleaned up after Maven plugin execution
Whenever executing a Maven project build (with or without tests - doesn't matter), exactly two temporary files are left in place without any content, e.g.:

```
-rw-rw-r-- 1 ancoron ancoron 0 Jun 25 14:23 /tmp/springboot-11563970946139969408-entrycontent
-rw-rw-r-- 1 ancoron ancoron 0 Jun 25 14:23 /tmp/springboot-5070898860288932403-entrycontent
```

The only reference I found to this pattern was from the [SizeCalculatingEntryWriter](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/SizeCalculatingEntryWriter.java#L102).

As everyone can see, using the temporary file is optional, although created every time and there is no safe-guard that would delete those files on JVM exit.

This means that over time, the temporary directory gets flooded with empty files.

dreis2211:
Hi,

this PR fixes #22108 by calling `deleteOnExit` on the temporary file and additionally prevents creating the file all together if we don't exceed the thresholds for which it was introduced.

Since `deleteOnExit` is only executed on "normal" VM termination, there is still a slight chance that temporary files will not be cleaned up when the VM dies for unknown reasons. But with the additional lazy init they probably won't be created in the first place for most apps.

Let me know what you think.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/SizeCalculatingEntryWriter.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/SizeCalculatingEntryWriter.java
@@ -98,8 +98,7 @@ final class SizeCalculatingEntryWriter implements EntryWriter {
 
 		private OutputStream outputStream;
 
-		SizeCalculatingOutputStream() throws IOException {
-			this.tempFile = File.createTempFile("springboot-", "-entrycontent");
+		SizeCalculatingOutputStream() {
 			this.outputStream = new ByteArrayOutputStream();
 		}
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, @dreis2211. Ideally, I'd like a solution that can be tested but that's not easy to do with `deleteOnExit()`. I made an attempt that makes `EntryWriter` `Closeable` but the knock-on effect of that was broader than I'd like. Having spent some time in the code trying to find a nice solution, I've found myself wondering why we're using a file at all. It's inconsistent with how we handle calculating the hash for unpack comments which is currently done entirely in memory. I'm going to put this one on hold for now while we take a step back and consider things a bit more broadly.

dreis2211:
@wilkinsona I actually share the testability concern, I had this too while developing.
My concern on the other hand is that - depending on how long this will be on-hold - we won't solve the bug for users in the upcoming 2.3.2 release. What about merging this for now to improve things for users, while opening a follow-up issue that tries to improve things on a broader level?

wilkinsona:
It would certainly be nice to get a fix in some form or other into 2.3.2. I've labelled #22108 for team attention so we'll hopefully discuss it later today. Let's see what comes out of that and then make a decision on the best way to proceed.

dreis2211:
Sounds reasonable. Thanks

philwebb:
We discussed this today and decided that we should merge this in 2.3.x and also bump up the threshold to reduce the number of files written. We'll perhaps try to look at refactoring that `EntryWriter` code sometime in the future to make them close aware.

wilkinsona:
Thanks again, @dreis2211!

PR #22113, opened 2020-06-25
Wording changes

Buzzardo:
Replaced potentially insensitive language with more neutral language.

Note that the code still has some of those terms, so the docs cannot
change those until they change in the code.


wilkinsona:
Thanks for the PR, @Buzzardo. I think we need to take a bit of a step back on this one so that we can make updates across all of the Spring projects that are consistent in terms of things like the replacements that we make, how we deal with broken links due to anchor's changing their names, etc. I'll put this on hold for now until we've had a chance to do that.

Adriansun:
Stop with this nonsense. Do no merge.

wilkinsona:
Looking through the changes proposed here, I think those that can be made have now been made as part of other commits.

I don't think we can make the `executable` to `runnable` change as the terms `execute` and `executable` are so ingrained. Unfortunately, trying to move away from that terminology is likely to introduce confusion and ambiguity. For example, a fully-executable jar is so called because it relates to executing a process and the execute file permission, before of which are standard terminology with which I think we need to align. This is particularly important for those who don't speak English as a first language where consistency with the terminology that they find in things like the man page for `chmod` will hopefully ease their understanding.

PR #22125, opened 2020-06-26
Add StatsD transport protocol configuration option

lndobryden:
This will expose the StatsD transport protocol configuration option as a configuration option for actuator. I have a need to use TCP for exporting StatsD to telegraf, the underlying micrometer library supports using TCP, but it is not yet exposed as a configuration property for Spring Boot.


spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
```diff
@@ -1764,12 +1764,13 @@ You can also change the interval at which metrics are sent to Stackdriver:
 ==== StatsD
 The StatsD registry pushes metrics over UDP to a StatsD agent eagerly.
 By default, metrics are exported to a {micrometer-registry-docs}/statsd[StatsD] agent running on your local machine.
-The StatsD agent host and port to use can be provided using:
+The StatsD agent host,port, and protocol to use can be provided using:
```
kedar-joshi, line 1767: There should be a space after `host,` i.e. `.. host, port, ..`.

wilkinsona:
@lndobryden Thanks very much for making your first contribution to Spring Boot.

PR #22138, opened 2020-06-27
Accept empty filter dispatcher types in auto-configurations

### LINKED ISSUE #22128: EnumSet.copyOf calls fail with empty collections
Both `SecurityFilterAutoConfiguration` and `SessionRepositoryFilterConfiguration` use `EnumSet::copyOf` in a stream. If the stream ends up being empty then the `copyOf` method will throw an exception.

dreis2211:
Hi,

this PR fixes #22128 by using

- `Collectors.toCollection(() -> EnumSet.noneOf(DispatcherType.class))` rather than
- `Collectors.collectingAndThen(Collectors.toSet(), EnumSet::copyOf)`

to prevent exceptions when the underlying collection is empty.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/security/servlet/SecurityAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/security/servlet/SecurityAutoConfigurationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/security/servlet/SecurityAutoConfigurationTests.java
```diff
@@ -164,6 +164,19 @@ void customFilterDispatcherTypes() {
...
 	}
 
+	@Test
+	void emptyFilterDispatcherTypesDoNotThrowException() {
+		this.contextRunner.withPropertyValues("spring.security.filter.dispatcher-types:")
+				.withConfiguration(AutoConfigurations.of(SecurityFilterAutoConfiguration.class)).run((context) -> {
+					DelegatingFilterProxyRegistrationBean bean = context.getBean("securityFilterChainRegistration",
+							DelegatingFilterProxyRegistrationBean.class);
+					@SuppressWarnings("unchecked")
+					EnumSet<DispatcherType> dispatcherTypes = (EnumSet<DispatcherType>) ReflectionTestUtils
```
wilkinsona, line 174: I think we could use AssertJ’s extracting here and we prefer to do so where possible.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/session/SessionAutoConfigurationTests.java
```diff
@@ -142,6 +142,17 @@ void filterDispatcherTypesCanBeCustomized() {
 				});
 	}
 
+	@SuppressWarnings("unchecked")
+	@Test
+	void emptyFilterDispatcherTypesDoNotThrowException() {
+		this.contextRunner.withUserConfiguration(SessionRepositoryConfiguration.class)
+				.withPropertyValues("spring.session.servlet.filter-dispatcher-types=").run((context) -> {
+					FilterRegistrationBean<?> registration = context.getBean(FilterRegistrationBean.class);
+					Object dispatcherTypes = ReflectionTestUtils.getField(registration, "dispatcherTypes");
```
wilkinsona, line 151: I think this could use AssertJ’s extracting too.

dreis2211:
I was undecided on that one actually. I wanted to stay consistent inside the specific tests, but I can change it of course.

wilkinsona:
The dangers of reviewing a diff in isolation. Ok, let's keep it consistent within the test classes for now. We can do a separate pass that replaces the use of `ReflectionTestUtils`. `hasFieldOrPropertyWithValue` would probably be a better fit than my earlier suggestion of `extracting` anyway.

dreis2211:
Yeah, I think doing this in a separate pass is probably better. On top of that I think that neither `hasFieldOrPropertyWithValue` or `extracting` work really great here. The former because it doesn't really play well with the empty case and the latter because it doesn't really work great with Sets (I only know of `extracting().asList()` at least and a quick test didn't reveal any cool solution).

wilkinsona:
Thanks very much once again, @dreis2211.

PR #22139, opened 2020-06-27
enable Dependabot v2

sullis:
https://github.blog/2020-06-01-keep-all-your-packages-up-to-date-with-dependabot/

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,10 @@
+version: 2
+updates:
+  - package-ecosystem: "gradle"
+    directory: "/"
+    schedule:
+      interval: "daily"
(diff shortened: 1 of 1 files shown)


spencergibb:
The boot team uses https://github.com/spring-io/bomr

wilkinsona:
Thanks for the proposal. We already have [our own tooling](https://github.com/spring-projects/spring-boot/tree/v2.3.1.RELEASE/buildSrc/src/main/java/org/springframework/boot/build/bom/bomr) for keeping dependencies up-to-date that is better suited to our needs. Among other things, it allows us to configure an upgrade policy for each branch. For example, we only want to pick up dependencies’ maintenance releases in our maintenance branches.

Issue #22025, opened 2020-06-19
Automatically exclude the annotation processors we know about


snicoll:
The Spring ecosystem has a number of annotation processors that are usually added as an optional dependency to the Maven build. While the `maven-compiler-plugin` has a way to declare them in plugin configuration for quite some time, this requires the version to be specified which is quite annoying. Not all IDEs pick up this new way of defining the annotation processor.

Maybe we could improve the Spring Boot Maven plugin to exclude such well-known annotation processors as they are useless at runtime. See #22000

wilkinsona:
We've taken care of this for Spring Boot's annotation processors in [this commit](https://github.com/spring-projects/spring-boot/commit/e743d5fe66302feca910f81a7285ca0caf3e1b84). Was that all that you had in mind, @snicoll, or did you want to cover annotation processors from other projects too?

snicoll:
@wilkinsona I am not sure. I meant to add the `context-indexer` in framework but we can't really declare that property in the manifest there.

wilkinsona:
I wonder if there any modules that are used as an annotation processor and at runtime? If not, we could exclude based on the presence of `META-INF/services/javax.annotation.processing.Processor`.

There’s some risk there and we’re really working around a limitation of Maven. I’m not sure that risk is worth it. We could offer a config option to turn off the exclusion but I’m not sure that’s worth it either. I’m tempted to close this one as I think what we’ve done this far articles a reasonable balance.

filiphr:
I've be following this issue for a while, since I was interested in this functionality and would like to give my 2 cents.

>I wonder if there any modules that are used as an annotation processor and at runtime? If not, we could exclude based on the presence of `META-INF/services/javax.annotation.processing.Processor`.

This is actually not a good idea. In an ideal world every annotation processor will provide 2 jars, one where the annotations are (runtime needed dependencies) and another one for the compile time dependencies. As an example Lombok is one such processor, where you only have one dependency for everything.

>There’s some risk there and we’re really working around a limitation of Maven. 

Why a limitation of Maven? The `maven-compiler-plugin` has the [annotationProcessorPaths](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#annotationProcessorPaths) which are similar to the Gradle `annotationProcessor`. Long time ago I raised #10645, which was meant to provide Project templates which are compatible with that. I just checked and start.spring.io does use `annotationProcessor` for annotation processors in the generated project.

[...]

PR #22203, opened 2020-07-02
Add type metadata to jars' manifests

### LINKED ISSUE #22036: Exclude empty starters jars and annotation processors when running or packaging with Maven and Gradle
@odrotbohm brought up the interesting idea that we could simply exclude empty starter JARs from our fat jars. This might help performance when searching for missing classes.

dreis2211:
Hi,

this PR fixes #22036 by marking the starter jars with an additional attribute in their manifest which is later checked to exclude the starter JAR during packaging.

I opened the PR now to discuss the taken approach better.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -74,6 +74,10 @@ gradlePlugin {
 			id = "org.springframework.boot.optional-dependencies"
 			implementationClass = "org.springframework.boot.build.optional.OptionalDependenciesPlugin"
 		}
+		packageExcludePlugin {
+			id = "org.springframework.boot.package-exclude"
+			implementationClass = "org.springframework.boot.build.PackageExcludePlugin"
+		}
 		starterPlugin {
 			id = "org.springframework.boot.starter"
 			implementationClass = "org.springframework.boot.build.starters.StarterPlugin"
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -38,6 +38,10 @@ checkstyle {
 
 gradlePlugin {
 	plugins {
+		annotationProcessorPlugin {
+			id = "org.springframework.boot.annotation-processor"
+			implementationClass = "org.springframework.boot.build.processors.AnnotationProcessorPlugin"
+		}
 		autoConfigurationPlugin {
 			id = "org.springframework.boot.auto-configuration"
 			implementationClass = "org.springframework.boot.build.autoconfigure.AutoConfigurationPlugin"
@@ -73,10 +77,6 @@ gradlePlugin {
 		optionalDependenciesPlugin {
 			id = "org.springframework.boot.optional-dependencies"
 			implementationClass = "org.springframework.boot.build.optional.OptionalDependenciesPlugin"
-		}
-		packageExcludePlugin {
-			id = "org.springframework.boot.package-exclude"
-			implementationClass = "org.springframework.boot.build.PackageExcludePlugin"
 		}
 		starterPlugin {
 			id = "org.springframework.boot.starter"
--- a/buildSrc/src/main/java/org/springframework/boot/build/PackageExcludePlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/PackageExcludePlugin.java
@@ -1,0 +1,45 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.build;
+
+import java.util.Map;
+import java.util.TreeMap;
+
+import org.gradle.api.Plugin;
+import org.gradle.api.Project;
+import org.gradle.api.tasks.bundling.Jar;
+
+/**
+ * Plugin to apply a custom marker attribute to the project's JAR manifest that can be
+ * used to filter the mentioned JARs from ending up in packages.
+ *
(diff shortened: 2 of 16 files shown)


snicoll:
I wonder if generalizing this slightly could help us fix https://github.com/spring-projects/spring-boot/issues/22025 as well.

dreis2211:
@snicoll You mean marking `spring-boot-autoconfigure-processor` & `spring-boot-configuration-processor` with a similar entry in its manifest (or with one that suits all)?

snicoll:
Yes. I should mention this was targeted to the team at large to see if we could come up with a manifest entry that makes sense for both this use case and the issue I've referenced.

wilkinsona:
I like that idea, @snicoll. A more general purpose manifest header that can be used to influence what's included sounds good to me. I'm struggling to think of a good name for it though. In the interests of consistency, I think it makes sense for the exclusion to apply more broadly than just building fat jars and wars. It should probably also affect the classpath used for `spring-boot:run`, `bootRun`, when using Gradle's application plugin, etc.

dreis2211:
What about `Exclude-From-Runtime` as a name for the manifest entry?

@wilkinsona I would need a pointer which classes are involved when this should be applied on a broader level beyond packaging. At the moment I'm looking at `AbstractRunMojo` for Maven and `BootRun` for Gradle. But especially for Gradle I find it a bit difficult to apply the chosen mechanism to be honest. (I only have looked for a couple of minutes, though)

wilkinsona:
For Gradle, I'd try to do it by filtering the classpath that the `BootRun`, `BootJar`, and `BootWar` plugins are configured with. I'd do that in `JavaPluginAction` and `WarPluginAction` when those tasks are being defined.

I've just reminded myself that the application distribution that Boot configures uses the fat jar or war so we don't need to worry about that one.

dreis2211:
For Gradle, this seems to conflict with `AdditionalMetadataLocationsConfigurer` in `JavaPluginAction` that seems to check for `spring-boot-configuration-processor` being on the classpath.

dreis2211:
@snicoll I just tried adding the following to a test project's `pom.xml`:
```
		<dependency>
			<artifactId>spring-boot-configuration-processor</artifactId>
			<optional>true</optional>
		</dependency>
```
When packaging this up I see no jar of the processor included in the build.

wilkinsona:
> For Gradle, this seems to conflict with `AdditionalMetadataLocationsConfigurer`

That's strange. That configurer checks the classpath of the `JavaCompile` classpath which should be unaffected by any filtering that's being done to the classpath of `BootJar` etc.

dreis2211:
When adding the following to a `build.gradle`
```
annotationProcessor "org.springframework.boot:spring-boot-autoconfigure-processor"
```
and packaging this up via `bootJar` I see no jar included. So both in Maven & Gradle the processors don't seem to end up already in the fat JARs. Am I missing something here?

wilkinsona:
That's the behaviour that I would expect with Gradle. With Maven, my understand was that the annotation processor would be included in the fat jar and #22025 was opened to improve that.

dreis2211:
Would that also render the classpath discussion obsolete? If it doesn't I would need help there, I'm afraid...

dreis2211:
Edit: I tried Maven again with a clean build again and it included the processors. Sorry for the noise. For Gradle, it seems to be not included as I wrote earlier.

wilkinsona:
> Would that also render the classpath discussion obsolete?

No, I don't think so. I'd like to try to use a classpath-filtering-based mechanism for the starters.

> If it doesn't I would need help there, I'm afraid...

Sure thing. Let us know when you're ready to pass the baton and we can take it from there.

dreis2211:
I've pushed a more general solution for the packaging aspect with a new `PackageExcludeFilter`. Let me know what you think of that.
Other than that, this would be the point were you probably need to take over to cover the classpath filtering.

dreis2211:
I just renamed `PackageExcludeFilter` to `FileExcludeFilter` as this is a more fitting name imho.

wilkinsona:
Thanks, @dreis2211.

FWIW, I think `Exclude-From-Runtime` that you mentioned above is preferable to `Exclude-From-Packaging` that is currently proposed. That said, I've been wondering if we should go with something that's more declarative, describing what the dependency is rather than how it should be treated. Something like `Spring-Boot-Dependency-Scope` or `Spring-Boot-Dependency-Nature` with values of `compileOnly` or `dependencyMetadataOnly` respectively. I'm not sure yet though. Let's see what you and the rest of the team think.

dreis2211:
I wasn't really happy with either `Exclude-From-Runtime` or `Exclude-From-Packaging`, so I like the idea of introducing `Spring-Boot-Dependency-Scope`, but I'm not sure about the values yet. The configuration processors would be marked with `compileOnly` and the starters with `dependencyMetadataOnly`, right? What about the other modules - maybe a simple `default` value? I'm also thinking how the Gradle plugin structure would look like for that and how we would like to apply this to all modules.

philwebb:
We had a bit of a brainstorm today and we quite like`Spring-Boot-Jar-Type` for the key with values `dependencies-starter` or `annotation-processor`.

wilkinsona:
Yeah, leave it empty for the others please.

dreis2211:
Okay, I will craft a new plugin then and let you know when I'm finished.

dreis2211:
I added a new `AnnotationProcessorPlugin` similar to the starter plugin that applies the new jar type to the manifest. That seemed more pragmatic than introducing an additional DSL extension for the new jar type.
I finished the work that I wanted to do. I'm not happy here and there with some duplication (more specifically with the magic strings `Spring-Boot-Jar-Type` and its values). Maybe you can find a good place for those constants somewhere.
Let me know if I should do something on this PR still.
Cheers

wilkinsona:
Thanks for the PR, @dreis2211. In reviewing this I came to the conclusion that we should take a different approach to filtering out the jars in the Maven and Gradle plugins so I reworked the changes here to add the `Spring-Boot-Jar-Type` manifest entry to Boot's starters and annotation processors. The changes to the filtering were made in https://github.com/spring-projects/spring-boot/commit/143d19754bd7489889029afc85e109bc7115c6c3 (Gradle) and https://github.com/spring-projects/spring-boot/commit/e743d5fe66302feca910f81a7285ca0caf3e1b84 (Maven) and apply to both packaging and running an app.

PR #22209, opened 2020-07-03
Use Integer.parseInt rather than valueOf where result is assigned to an int

XenoAmess:
because you use an int to receive it not an Integer.


wilkinsona:
Thanks very much, @XenoAmess.

PR #22210, opened 2020-07-03
Use new ArrayList(Collection) rather than separate instantiation and addAll(Collection)

XenoAmess:
Though slightly.


wilkinsona:
Thank you, @XenoAmess.

PR #22211, opened 2020-07-03
Use new ArrayList(Collection) rather than separate instantiation and addAll(Collection)

XenoAmess:
as title.


wilkinsona:
Thanks again, @XenoAmess.

PR #22213, opened 2020-07-03
Remove redundant close() call from JarResourceManagerTests

XenoAmess:
why close it two times?


wilkinsona:
Thanks again, @XenoAmess.

PR #22214, opened 2020-07-03
Fix unwrapping of captured PrintStream in test support's OutputCapture

XenoAmess:
this should not be like this. I guess you made a mistake here.
if you really get the original work flow, you should use `if` insteadof `while`.

Warning: this fix might be WRONG, so take good care about it.
If your original codes be a typo, and what I fix now is the function you want, that is OK.
But if the original codes works exactly what you want, then you should change `while` to `if`
either of the two fixes should happen.


wilkinsona:
Thanks very much, @XenoAmess.

PR #22227, opened 2020-07-06
Upgrade to Maven Shade Plugin 3.2.4

anshlykov:
This PR updates `maven-shade-plugin` to 3.2.4.

The resource transformer was updated here https://github.com/spring-projects/spring-boot/commit/5e7917e33a142ea41e7e787478182122466d13f1, but the plugin version in `spring-boot-dependencies` was not updated.


wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @anshlykov, and for taking care of removing the no-longer-needed prohibited version.

PR #22240, opened 2020-07-07
Polish ErrorPage.equals()

izeye:
This PR polishes `ErrorPage.equals()` by short-circuiting immediately without additional evaluation and assignments.


snicoll:
Thanks again Johnny!

PR #22256, opened 2020-07-07
Incorrect Configuration Validation for mongo AutoConfigurationFactory() 

devzer01:
### The following assertion is incorrect, the behavior of Assert.state is to throw an exception when the expression is false
```
private void validateConfiguration() {
	if (hasCustomAddress() || hasCustomCredentials() || hasReplicaSet()) {
		Assert.state(this.properties.getUri() == null,
		   "Invalid mongo configuration, either uri or host/port/credentials/replicaSet must be specified");
	}
}
```
The code above **throws an exception when the configuration includes a valid mongo-URI**

TLDR;
Assert.state() // tests for the expression to be false. when Assert.state(getURI() == null) is false then Assert.state throws an Exception. 
Also if the configuration hasCustomAddress() then the test for URI should not be considered as a validator option.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoClientFactorySupport.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoClientFactorySupport.java
@@ -76,8 +76,7 @@ public abstract class MongoClientFactorySupport<T> {
 
 	private void validateConfiguration() {
 		if (hasCustomAddress() || hasCustomCredentials() || hasReplicaSet()) {
-			Assert.state(this.properties.getUri() == null,
-					"Invalid mongo configuration, either uri or host/port/credentials/replicaSet must be specified");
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@devzer01 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=22256)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=22256) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@devzer01 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=22256)!

wilkinsona:
Thanks for the PR but I believe that the current behaviour is correct. The intent of the assertion is to enforce that either a URI is configured or the host/port/credentials/replicaSet are configured. In other words you should configure everything using the URI or everything using the individual properties rather than mixing the two. Thanks anyway.

devzer01:
yep, I realized the same later on :-) maybe my English isn't that good to understand the error message being produced.

PR #22261, opened 2020-07-08
Replace ReflectionUtils with AssertJ's extracting()

izeye:
This PR replaces `ReflectionUtils` with AssertJ's `extracting()`.


snicoll:
Thanks again @izeye

PR #22266, opened 2020-07-08
Measure with nanoseconds in HttpExchangeTracer

### LINKED ISSUE #21742: Measure with nanoseconds in HttpExchangeTracer
Hi,

I just noticed that `HttpExchangeTracer.sendingResponse` is working with milliseconds or more specifically with `System.currentTimeMillis` in order to track the taken time.

As `System.currentTimeMillis` can be subject of time drift, I wonder if `System.nanoTime` should be used. Unfortunately, it's not simply done with just switching to the latter because we can't work with nano-time and `Instant` like it's done at the moment. There need to be changes to `HttpTrace` as well, which potentially means migration work for people with a persistent store and a custom `HttpTraceRepository`.

In theory this would also obsolete `HttpTrace.getTimestamp`, but probably it's okay to keep it to have an actual time representation of when the request/response exchange happened.

I've done a [first draft](https://github.com/spring-projects/spring-boot/compare/master...dreis2211:httpexchangetracer-nanotime) before I open an actual PR.

Let me know what you think.

Cheers,
Christoph

dreis2211:
Hi,

this PR closes #21742 as discussed by providing a new transient field `nanoTime` in `HttpTrace` and a corresponding package-private accessor, that eventual persistent implementations of `HttpTraceRepository` don't need to care about.

Let me know what you think.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/trace/http/HttpExchangeTracer.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/trace/http/HttpExchangeTracer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks very much, @dreis2211.

PR #22274, opened 2020-07-08
Fix dead link url and name of the STS download link

fernandezseb:
In the CONTRIBUTING documentation the link to the Spring Tool Suite url is a dead link. The url is replaced by the new url where STS can be downloaded.
The name of this link is contains a typo: Tools Suite (with an s after the Tool) instead of Tool Suite.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -76,7 +76,7 @@ added after the original pull request but before a merge.
 
 == Working with the Code
 If you don't have an IDE preference we would recommend that you use
-https://spring.io/tools/sts[Spring Tools Suite] or
+https://spring.io/tools[Spring Tool Suite] or
 https://eclipse.org[Eclipse] when working with the code. We use the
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.adoc
```diff
@@ -76,7 +76,7 @@ added after the original pull request but before a merge.
 
 == Working with the Code
 If you don't have an IDE preference we would recommend that you use
-https://spring.io/tools/sts[Spring Tools Suite] or
+https://spring.io/tools[Spring Tool Suite] or
```
MrHeer, line 79: Maybe this is better?
```suggestion
https://spring.io/tools[Spring Tools] or
```

wilkinsona: I think that `Spring Tool Suite` is the better name here. It's the name of the Eclipse distribution that includes Spring Tools 4 and that's what we're referring to. I can take care of changing it back as part of the merge.

wilkinsona:
@fernandezseb Thank you very much for making your first contribution to Spring Boot.

Issue #22297, opened 2020-07-09
Support for HTTP response headers in Spring boot actuators


JigarJoshi:
Spring boot actuator offers a simple HTTP api to write web endpoints. It is intended to be simple. I have a usecase where I use embedded async-profiler and allow profiling of my spring boot application.

One of the file format I support is JFR and I want to be able to allow user to download the file from actuators. In current API setup download of file is possible however naming the downloaded file is not.

This is to support such usecase.

wilkinsona:
From what you have described, the heap dump endpoint sounds similar to your endpoint in that it produces binary data that is typically written to a file. In the [Actuator API documentation](https://docs.spring.io/spring-boot/docs/2.3.1.RELEASE/actuator-api/html/#heapdump-retrieving) we document the use of curl's `-O` to write the heap dump to disk. This results in a file named `heapdump` (derived from the URL) which isn't quite what you want.

With curl, you can use [`-o, --output <file>`](https://curl.haxx.se/docs/manpage.html#-o) to specify the name of the file to which the response body is written. Even if the server was able to suggest a default filename, I think it's likely that a user may want to customise it to meet their own needs and is likely to use `-o` or an equivalent option with a different client to do so. From what I know thus far, it feels like your use case may be solved by some documentation.

JigarJoshi:
Yes we can make the client output to different file while using such clients. However in my experience I have seen many developers using browser to hit certain actuator endpoints. 

While we can't attach response header `Content-Disposition: attachment; filename="${NAME}"` it defaults to the last part of uri (for example heapdump in case of `/heapdump`). Users will then have to rename file to add certain extension for the toolings (For example JMC).

This isn't a must have feature but as actuator grows I think it will be nice to add it in. I am happy to take your suggestion and introduce/reuse proper model for HttpHeader instead of MultiValuedMap.

Please let me know what do you think

philwebb:
We discussed this today as a team and decided that unfortunately we don't want to surface HTTP headers in the `WebEndpointResponse` class. If your code is just targeting one specific technology, you might want to look at `@RestControllerEndpoint` (and equivalents) which surfaces the existing MVC API, including support for HTTP headers.

PR #22298, opened 2020-07-10
 Support for HTTP response headers in Spring boot actuators #22297 

JigarJoshi:
#22297


wilkinsona:
Thanks for the proposal. If we're to support HTTP headers in endpoint web extensions, I think we'll need a proper abstraction for the headers (like Framework's `HttpHeaders` which we can't use as it isn't available when using Jersey) rather than using `MultiValueMap`. Before we explore that, I'd like to take a step back and have a think about whether or not this is a use case that we want to support. Let's do that on #22297.

philwebb:
Thanks for the PR, but as discussed in https://github.com/spring-projects/spring-boot/issues/22297#issuecomment-657625669 we feel it's better not to surface HTTP headers.

PR #22299, opened 2020-07-10
Upgrade to Spring Data Neo4j 6

meistermeier:
This PR adapts Spring Boot 2.4 to Spring Data Neo4j 6.

For some background:
Neo4j has developed two projects in the last year

* A starter dedicated for the Java Driver for Neo4j (dubbed the "Bolt Driver") https://github.com/neo4j/neo4j-java-driver-spring-boot-starter
* SDN/RX: https://github.com/neo4j/sdn-rx Which does imperative and reactive Spring Data Neo4j purely based on the Bolt Driver

SDN/RX has become SDN 6 and is now merged into Spring Data Neo4j.

That means Spring Boot depending on the Ockham release train will not build anymore.

This adapts Spring Boot 2.4 to the SDN 6

* It brings automatic configuration for the Bolt Driver under the new namespace `spring.neo4j`, similar to others (like mongo and redis)
* Adapts automatic configuration for Spring Data Neo4j

This removes the dependency on Neo4j-OGM as well.

SDN 6 won't have support for bringing up an embedded instance, which is in line with our strategy to recommend Testcontainers or the Neo4j Testharness to our customers and users.

We refrained from adding support for the Neo4j Testharness to the test slices to not broaden the topic more.
However, if we want to help users to bring something up running very fast, that would be an additional option.

The change https://github.com/spring-projects/spring-boot/compare/master...meistermeier:feature/neo4j-starter?expand=1#diff-76ff4fe38f0762e0fcb54986cae8e0b6R1608 is in there to enable SDN 6 for now because the dependency to Spring Data is on Ockham M1 where SDN 6 is not available.

We are looking forward to your feedback.

PR #22301, opened 2020-07-10
Add auto-configuration for Neo4j driver

meistermeier:
This commit adds the support for creating a managed instance of the Neo4j Java driver.
The low-level support for Neo4j is helpful in situations where the high-level abstraction of Spring Data Neo4j is not needed.

The PR is the foundation for the Spring Data Neo4j Starter update PR https://github.com/spring-projects/spring-boot/pull/22299 because it provides classes needed in the updated SDN starter.

Also there is the suggested related change for the initializer that enables the Neo4j driver dependency and allows the autoconfiguration to kick in: https://github.com/spring-io/start.spring.io/pull/483


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/DriverConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+ * @author Michael J. Simons
+ */
+@Configuration(proxyBeanMethods = false)
+final class DriverConfiguration {
+
+	@Bean
+	@ConditionalOnMissingBean(Driver.class)
+	Driver neo4jDriver(final Neo4jDriverProperties driverProperties) {
+
+		final AuthToken authToken = driverProperties.getAuthentication().asAuthToken();
```
snicoll, line 41: Style, our codebase doesn't use `final` (neither in attribute, nor in constructor argument).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverProperties.java
```diff
@@ -0,0 +1,497 @@
...
+import org.springframework.util.StringUtils;
+
+/**
+ * Used to configure an instance of the {@link org.neo4j.driver.Driver Neo4j-Java-Driver}.
+ *
+ * @author Michael J. Simons
+ * @since 2.4.0
+ */
+@ConfigurationProperties(prefix = "spring.neo4j")
+public class Neo4jDriverProperties {
```
snicoll, line 43: Given the `spring.neo4j` namespace, this should be named `Neo4jProperties`. Usage of this class should also be named consistently. If you think the driver part is important, then it should be reflected in the namespace.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverProperties.java
```diff
@@ -0,0 +1,497 @@
...
+
+		public String getKerberosTicket() {
+			return this.kerberosTicket;
+		}
+
+		public void setKerberosTicket(String kerberosTicket) {
+			this.kerberosTicket = kerberosTicket;
+		}
+
+		AuthToken asAuthToken() {
```
snicoll, line 180: Although a number of other properties have the same structure, we are trying to make sure `@ConfigurationProperties`-annotated type are as simple as possible and do not contain any logic. Please move this to the configuration if possible.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverProperties.java
```diff
@@ -0,0 +1,497 @@
...
+
+	public void setConfig(DriverSettings config) {
+		this.config = config;
+	}
+
+	public AuthToken getAuthToken() {
+		return this.authentication.asAuthToken();
+	}
+
+	public Config asDriverConfig() {
```
snicoll, line 102: Same remark than `asAuthToken`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverAutoConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+ *
+ * @author Michael J. Simons
+ * @since 2.4.0
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureBefore(Neo4jDataAutoConfiguration.class)
+@ConditionalOnClass(Driver.class)
+@EnableConfigurationProperties(Neo4jDriverProperties.class)
+@Import(DriverConfiguration.class)
+public class Neo4jDriverAutoConfiguration {
```
snicoll, line 42: For consistency with other driver auto-config, this should be `Neo4jAutoConfiguration`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverAutoConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+ * other instance has been manually configured.
+ *
+ * @author Michael J. Simons
+ * @since 2.4.0
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureBefore(Neo4jDataAutoConfiguration.class)
+@ConditionalOnClass(Driver.class)
+@EnableConfigurationProperties(Neo4jDriverProperties.class)
+@Import(DriverConfiguration.class)
```
snicoll, line 41: I don't see with the separate configuration class is necessary. Can't you define the bean in this auto-configuration?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverAutoConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+ * Automatic configuration of Neo4js Java Driver.
+ * <p>
+ * Provides an instance of {@link Driver} if the required library is available and no
+ * other instance has been manually configured.
+ *
+ * @author Michael J. Simons
+ * @since 2.4.0
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureBefore(Neo4jDataAutoConfiguration.class)
```
snicoll, line 38: This should be the other way around. Conceptually the Spring Data bits require the driver so they should define the relationship. This also makes sure the low-level `neo4j` package does not have any import in the `data` package.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverAutoConfiguration.java
```diff
@@ -0,0 +1,44 @@
...
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.data.neo4j.Neo4jDataAutoConfiguration;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+
+/**
+ * Automatic configuration of Neo4js Java Driver.
+ * <p>
+ * Provides an instance of {@link Driver} if the required library is available and no
```
snicoll, line 31: This Javadoc is not consistent. We usually don't explain in Javadoc what the conditions do.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverProperties.java
```diff
@@ -0,0 +1,497 @@
...
+ *
+ * @author Michael J. Simons
+ * @since 2.4.0
+ */
+@ConfigurationProperties(prefix = "spring.neo4j")
+public class Neo4jDriverProperties {
+
+	/**
+	 * The uri this driver should connect to. The driver supports bolt or neo4j as
+	 * schemes. Defaults to bolt://localhost:7687.
```
snicoll, line 47: We don't document default value in the description of the key.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverProperties.java
```diff
@@ -0,0 +1,497 @@
...
+ * Used to configure an instance of the {@link org.neo4j.driver.Driver Neo4j-Java-Driver}.
+ *
+ * @author Michael J. Simons
+ * @since 2.4.0
+ */
+@ConfigurationProperties(prefix = "spring.neo4j")
+public class Neo4jDriverProperties {
+
+	/**
+	 * The uri this driver should connect to. The driver supports bolt or neo4j as
```
snicoll, line 46: Configuration key descriptions do not start with `The`, `A`, etc.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/DriverConfiguration.java
```diff
@@ -0,0 +1,47 @@
...
+		final Config config = driverProperties.asDriverConfig();
+
+		return GraphDatabase.driver(driverProperties.getUri(), authToken, config);
```
snicoll, line 44: There must be a fallback for `spring.data.neo4j.uri` and a metadata with warning deprecation should indicate the replacement.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverProperties.java
```diff
@@ -0,0 +1,497 @@
...
+			return this.kerberosTicket;
+		}
+
+		public void setKerberosTicket(String kerberosTicket) {
+			this.kerberosTicket = kerberosTicket;
+		}
+
+		AuthToken asAuthToken() {
+
+			boolean hasUsername = StringUtils.hasText(this.username);
```
snicoll, line 182: The code might a bit messy here but I had hoped if that makes sense to provide a fallback for `spring.data.neo4j.usename` and `spring.data.neo4j.password`. Same, a metadata with warning deprecation should indicate the replacement.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jSpringJclLogging.java
```diff
@@ -0,0 +1,103 @@
...
+import org.neo4j.driver.Logger;
+import org.neo4j.driver.Logging;
+
+/**
+ * Shim to use Spring JCL implementation, delegating all the hard work of deciding the
+ * underlying system to Spring and Spring Boot.
+ *
+ * @author Michael J. Simons
+ */
+final class Neo4jSpringJclLogging implements Logging {
```
snicoll, line 30: Shouldn't this be part of Spring Data Neo4j?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverPropertiesTests.java
```diff
@@ -0,0 +1,416 @@
...
+	private static void assertDuration(Duration duration, long expectedValueInMillis) {
+		if (expectedValueInMillis == org.neo4j.driver.internal.async.pool.PoolSettings.NOT_CONFIGURED) {
+			assertThat(duration).isNull();
+		}
+		else {
+			assertThat(duration.toMillis()).isEqualTo(expectedValueInMillis);
+		}
+	}
+
+	@Nested
```
snicoll, line 84: As I've indicated on the other PR we don't use `@Nested` anywhere so I'd prefer to leave it out of this PR if possible.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/neo4j/Neo4jDriverPropertiesTests.java
```diff
@@ -0,0 +1,416 @@
...
+/**
+ * @author Michael J. Simons
+ */
+class Neo4jDriverPropertiesTests {
+
+	private AnnotationConfigApplicationContext context;
+
+	private Neo4jDriverProperties load(String... properties) {
+
+		AnnotationConfigApplicationContext ctx = new AnnotationConfigApplicationContext();
```
snicoll, line 60: Even such tests use the application context runner. It is a superior option compared to a simple handling of the context as you still need to handle the context otherwise.

Given that the logic for the properties object would be moved elsewhere, this might not be necessary after all.

spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -1432,6 +1439,7 @@ bom {
 				"spring-boot-starter-logging",
 				"spring-boot-starter-mail",
 				"spring-boot-starter-mustache",
+				"spring-boot-starter-neo4j",
```
snicoll, line 1442: I think this should be removed as there is no starter for the driver (and their shouldn't be any if all that's needed is the driver.

meistermeier:
(Again) thanks for the valuable feedback. ❤️ 
I would now focus on this PR to get the foundation for the SDN 6 changes (and compilable code). At least I see a benefit of having these independent changes merged first and base the other PR(s) on this.

snicoll:
Yes! Thank you again.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/Neo4jSpringJclLogging.java
wilkinsona: I don't think that would help as this is in the part that just auto-configures the driver when SDN may not be around.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/DriverConfiguration.java
meistermeier: There is no plan for a deprecation phase for the (soon) former SDN properties.
The SDN 6 replacement will be so disruptive that it just does not make any sense to respect those values anymore. The SDN starter will require this auto-configuration to reflect the truth for those properties: They belong into the driver and not into the higher level configuration.
In this hopefully short period of having this change merged and the pending SDN PR merge, I don't really consider to support the former properties.

snicoll: > There is no plan for a deprecation phase for the (soon) former SDN properties.

I get that but that property is in Spring Boot and part of its public API so I was hoping we could honour it still if that makes any sense. I am thinking of the use case of someone using Neo4j with a custom URI and/or username/password authentication. 

Is there any problem to honour the existing property in a deprecated fashion?

meistermeier:
I hope that I fixed or clarified the problems you have found.
Unfortunately I have overseen the formatting issues in the first run(s). There was also one in the `Neo4jProperties` but a `gradle format` did not touch it and in the next run it wasn't there anymore 🤷‍♀️

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/neo4j/DriverConfiguration.java
meistermeier: SDN will not know about those properties in the future because it does not configure the `Driver` anymore. The only place I see this is in the driver configuration but this would mix up the responsibilities for the `spring.data.neo4j` namespace.
If there is a away to alias properties, this would be helpful. So, I could say <legacy-property> is deprecated but gives it value to <new-property> Unfortunately I am not aware of any.
I am aware that a sudden drop without deprecation is not the finest thing to do but yet it seems like the cleanest solution. Otherwise we would have to deprecate everything (also in SDN 5.3.x) and the users only see striked through text everywhere.

snicoll: To be clear, I am not talking about SDN at all. This is purely a Spring Boot topic. This is 100% on the user side too. The use case I have in mind is someone having `spring.data.neo4j.uri=xxxx` in their configuration and upgrading to 2.4. Rather than trying to connect to a Neo4j instance on localhost, I would like the auto-configuration of the driver to pick up this property as a fallback. 

> If there is a away to alias properties

Yes, there is. This is what I am suggesting here. I just wanted to make sure this wasn't a bad idea before trying to give it a try. I'll do that when polishing this one but feel free to comment if, based on this additional context, you think we shouldn't.

meistermeier: We are definitely on the same page here. I just wanted to underline that SDN will not be as it was before from an API perspective. That was my reasoning for "just dropping" the old properties.
I think it is a welcoming transition help for <2.4 users going for 2.4 but should be deprecated on arrival 😺 
If you can give me a hint how to make the old properties an alias for the new ones, I could try this myself. My mind somehow blocks the idea of throwing (partial) ownership for `spring.data.neo4j` and `spring.neo4j` in the same place and the rest of the real new `spring.data.neo4j` properties will get handled by another auto-configurer.

snicoll: Alright. I can give that a try as part of polishing this next week.

meistermeier:
Is there anything left to do for me or do you need any additional information?

snicoll:
I am good for this one, thanks @meistermeier!

snicoll:
Thanks a lot @meistermeier (and @michael-simons). I've polished things quite a bit in https://github.com/spring-projects/spring-boot/commit/8c418adb9b24f703cc8241d8bf225ce06290be3e and we have enough runway to adapt things if necessary. 

I am going to review the actuator stuff that are relevant to the driver next. And finally move to Spring Data `2020.0.0-SNAPSHOT` to upgrade to SDN 6.x. If you have time to rebase #22299, that would be much appreciated.

meistermeier:
Thanks a lot @snicoll . Rebasing and ensuring everything still works as expect is on my list next.

PR #22302, opened 2020-07-10
Switch Neo4j health checks to the new Neo4j driver

meistermeier:
This commit replaces the Neo4j-OGM based health checks and metrics
with new ones based on the Neo4j Java driver.

This PR cannot get merged on its own because it depends on https://github.com/spring-projects/spring-boot/pull/22299


snicoll:
I don't think that `Neo4jMetrics` belong here. After a discussion with @meistermeier, the plan is to move `Neo4jMetrics` to the Micrometer project. We will limit this PR to health checks.

Assuming the contribution to micrometer is accepted, we can resume integration of metrics in a separate issue.

PR #22330, opened 2020-07-14
Avoid unnecessary allocations in ConditionMessage for NORMAL style

dreis2211:
Hi,

I noticed a small possible improvement in `ConditionMessage`. When the `NORMAL` style is used, nothing really happens to the applied items, so it is not necessary to create a new `ArrayList` to copy things. Instead the initial collection can be used.

Cheers,
Christoph

P.S.: While working on this, I noticed that the several `items` methods specify the following:
```@param items the source of the items (may be {@code null})```
Yet, the default `applyTo` method in `Style` would throw an NPE in case null is passed. Do you want me to open another PR to fix this or should I do another commit in this one to fix things?


snicoll:
Thanks again for the PR!

>  Do you want me to open another PR to fix this or should I do another commit in this one to fix things?

Another PR please. This one is an optimization and what you've described is quite different.

dreis2211:
Hence the suggestion and no second commit yet ;-) My gut feeling was in favor of a second PR, but wanted to clarify it first.

Issue #21504, opened 2020-05-19
Provide configuration properties for all of Reactor Netty's HttpDecoderSpec


deci260:
Spring Boot version: 2.2.5.RELEASE
Project: Spring Boot Autoconfigure

The `ServerProperties` class defines some properties such as `MaxHttpHeaderSize`.  The reason why I think the `MaxInitialLineLength` property should be added is because there's currently no way to define it when working with reactor netty. Even trying to define it by using `NettyWebServerFactoryCustomizer`s doesn't work because the `HttpRequestDecoder` customizers get overridden by the one used by spring boot to define the `MaxHttpHeaderSize` property.

This issue is semi-related to [this](https://github.com/spring-cloud/spring-cloud-gateway/pull/1554) Spring Cloud Gateway's PR that added the same property but for the http client.

wilkinsona:
> Even trying to define it by using NettyWebServerFactoryCustomizers doesn't work because the HttpRequestDecoder customizers get overridden by the one used by spring boot to define the MaxHttpHeaderSize property.

Spring Boot's customizer has an order of 0. If you order yours to run after Spring Boot's the configuration that it applies will not be overridden.

deci260:
> Spring Boot's customizer has an order of 0. If you order yours to run after Spring Boot's the configuration that it applies will not be overridden.

Thanks for the feedback!

wilkinsona:
Closing in favour of #22367.

PR #22367, opened 2020-07-16
Add properties for Netty HttpDecoderSpec

Julien-Eyraud:
Closes gh-21504


wilkinsona:
@Julien-Eyraud Thank you very much for making your first contribution to Spring Boot. If you are interested, I made [a few changes](https://github.com/spring-projects/spring-boot/commit/0e8bf94289d20ff4fe10ef2867a0e6706e9f413f) while merging. They were primarily to provide default values for the properties (that will appear in your IDE when editing application.properties or application.yaml) and testing that they align with Reactor Netty's own defaults.

PR #22392, opened 2020-07-17
Ignore BackgroundPreinitializer in native images

sdeleuze:
`BackgroundPreinitializer` is not relevant in native images so it is better to ignore it automatically in order to reduce the footprint in such environment.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/BackgroundPreinitializer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/BackgroundPreinitializer.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2020 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -45,6 +45,7 @@ import org.springframework.http.converter.support.AllEncompassingFormHttpMessage
  * @author Phillip Webb
  * @author Andy Wilkinson
  * @author Artsiom Yudovin
+ * @author Sebastien Deleuze
  * @since 1.3.0
  */
 @Order(LoggingApplicationListener.DEFAULT_ORDER + 1)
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/BackgroundPreinitializer.java
```diff
@@ -59,14 +60,21 @@
 	 */
 	public static final String IGNORE_BACKGROUNDPREINITIALIZER_PROPERTY_NAME = "spring.backgroundpreinitializer.ignore";
 
+	/**
+	 * Whether this environment lives within a native image.
+	 */
+	private static final boolean IN_NATIVE_IMAGE = (System.getProperty("org.graalvm.nativeimage.imagecode") != null);
```
wilkinsona, line 66: Would Graal's code elimination work if we had a single `ENABLED` field that was initialised using both the value of `spring.backgroundpreinitializer.ignore` and the presence of `org.graalvm.nativeimage.imagecode`?

wilkinsona: We could perhaps include the available processors check too:

```java
private static final boolean ENABLED;

static {
	ENABLED = System.getProperty("org.graalvm.nativeimage.imagecode") == null
		&& !Boolean.getBoolean(IGNORE_BACKGROUNDPREINITIALIZER_PROPERTY_NAME)
		&& Runtime.getRuntime().availableProcessors() > 1;
}
```

sdeleuze: Good point for the unique `ENABLED` flag, but I think we should not include the `Runtime.getRuntime().availableProcessors() > 1` check because the class will be initialized at build time on native image, so you will have the available processors for the build machine, not the runtime one.

sdeleuze: Oh but we don't care given the previous native image check so I guess we can do that.

wilkinsona:
@sdeleuze I pulled the introducing of the `ENABLED` static final field out into a separate commit (https://github.com/spring-projects/spring-boot/commit/85828953bd82ce2e0f1a61847338a1bc32e41241) and then rebased the changes here on top of that (https://github.com/spring-projects/spring-boot/commit/137ce1a9de6087f26357a6829fa6864b948b641a). If you have a moment, please check that I haven't messed anything up from a native image perspective in the process.

PR #22412, opened 2020-07-20
Allow EntityScanner to be configured to scan for interfaces and abstract classes

ge0ffrey:
The EntityScanner, used by Spring Boot starter implementations to correct discover domain objects with specific annotations (such as `@Entity`, `@PlanningEntity`, etc), **only scans for concrete classes**, not for abstract classes nor interfaces.
Some starter implementations (such as the optaplanner one) need to scan for abstract classes and interfaces too.

This change allows them to do that, by enabling it with a constructor parameter (it's still excluding them by default), so they don't have to duplicate the EntityScanner logic (so they remain compatible with future Spring Boot releases).

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/domain/EntityScanner.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/domain/EntityScanner.java
@@ -42,6 +42,8 @@ public class EntityScanner {
 
 	private final ApplicationContext context;
 
+	private final ClassPathScanningCandidateComponentProvider componentProvider;
+
 	/**
 	 * Create a new {@link EntityScanner} instance.
 	 * @param context the source application context
@@ -49,6 +51,21 @@ public class EntityScanner {
 	public EntityScanner(ApplicationContext context) {
 		Assert.notNull(context, "Context must not be null");
 		this.context = context;
+		this.componentProvider = new ClassPathScanningCandidateComponentProvider(false);
+		this.componentProvider.setEnvironment(this.context.getEnvironment());
+		this.componentProvider.setResourceLoader(this.context);
+	}
+
+	/**
+	 * Create a new {@link EntityScanner} instance.
+	 * @param context the source application context
+	 * @param componentProvider the scanner, it's environment and resource loader must
+	 * already be set
+	 */
+	public EntityScanner(ApplicationContext context, ClassPathScanningCandidateComponentProvider componentProvider) {
+		Assert.notNull(context, "Context must not be null");
+		this.context = context;
+		this.componentProvider = componentProvider;
 	}
 
 	/**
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/domain/EntityScanner.java
```diff
@@ -42,11 +44,24 @@
...
+		this(context, true);
+	}
+
+	/**
+	 * Create a new {@link EntityScanner} instance.
+	 * @param context the source application context
+	 * @param excludeAbstractClasses true if only concrete classes are scanned, false if
+	 * abstract classes and interfaces are scanned too
+	 */
+	public EntityScanner(ApplicationContext context, boolean excludeAbstractClasses) {
```
snicoll, line 63: Considering the use cases of scanning abstract classes is quite specific, I would not surface such an option.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/domain/EntityScanner.java
```diff
@@ -63,7 +78,19 @@ public EntityScanner(ApplicationContext context) {
 		if (packages.isEmpty()) {
 			return Collections.emptySet();
 		}
-		ClassPathScanningCandidateComponentProvider scanner = new ClassPathScanningCandidateComponentProvider(false);
+		ClassPathScanningCandidateComponentProvider scanner;
+		if (this.excludeAbstractClasses) {
```
snicoll, line 82: Rather than doing all of this, I would expose a `protected` method where the `ClassPathScanningCandidateComponentProvider` can be tuned. You'd then be able to change its behaviour without having to duplicate the logic.

snicoll:
@ge0ffrey how is it going? Did you get a chance to look at the review?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/domain/EntityScanner.java
ge0ffrey: That makes sense. I 'll make those changes the week of 17 of August. Thanks for the review!

ge0ffrey:
Sorry for the delay, I'll do the sugested changes the week of 17 Aug.

ge0ffrey:
@snicoll Ready for review I think

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/domain/EntityScanner.java
```diff
@@ -42,13 +42,30 @@
...
+		this.componentProvider.setResourceLoader(this.context);
+	}
+
+	/**
+	 * Create a new {@link EntityScanner} instance.
+	 * @param context the source application context
+	 * @param componentProvider the scanner, it's environment and resource loader must
+	 * already be set
+	 */
+	public EntityScanner(ApplicationContext context, ClassPathScanningCandidateComponentProvider componentProvider) {
```
snicoll, line 65: I mentioned in my review "I would expose a protected method where the ClassPathScanningCandidateComponentProvider can be tuned.". 

There is nothing wrong having to extend from `EntityScanner` for your use case. I don't think exposing the provider as a parameter is warranted for your use case either.

snicoll:
@ge0ffrey how is it doing? I'd like to get a resolution on this one and will take over next week unless you beat me to it. Thanks!

snicoll:
Thanks for the PR but I've decided to implement this differently as described in my review. See #23154

PR #22484, opened 2020-07-21
Actual method parameter is the same constant inspection

minny-park:
methods where a value being passed to a particular parameter appears to be always the same constant


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/info/InfoPropertiesInfoContributor.java
```diff
@@ -128,12 +128,11 @@ protected void copyIfSet(Properties target, String key) {
 	/**
 	 * Replace the {@code value} for the specified key if the value is not {@code null}.
 	 * @param content the content to expose
-	 * @param key the property to replace
 	 * @param value the new value
 	 */
-	protected void replaceValue(Map<String, Object> content, String key, Object value) {
-		if (content.containsKey(key) && value != null) {
-			content.put(key, value);
+	protected void replaceValue(Map<String, Object> content, Object value) {
```
philwebb, line 133: We cannot change the signature of a protected method since someone may be overriding it.

snicoll:
@minny-park how is it going? Did you get a chance to look at the review?

PR #22499, opened 2020-07-22
Add plugin management for liquibase-maven-plugin

### LINKED ISSUE #22486: Add plugin management for liquibase-maven-plugin
Hello

Would it be possible to add liquibase-maven-plugin to spring-boot-dependencies pom.xml ?

With something like this, project will inherit the plugin version, which is nice
```
        <pluginManagement>
            <plugins>
                <plugin>
                    <groupId>org.liquibase</groupId>
                    <artifactId>liquibase-maven-plugin</artifactId>
                    <version>${liquibase.version}</version>
                </plugin>
            </plugins>
        </pluginManagement>
```
(I know you have switched to gradle recently but I am not familiar with it)

Because using different liquibase versions while running the application and running liquibase via maven can cause issue about checksum validation when starting the app. The issue caused is a slow start because it's recalculating checksums.

> 2020-07-21 18:01:13.585  INFO 50937 --- [  restartedMain] l.lockservice.StandardLockService        : Successfully acquired change log lock

After this log appears, it tends to get stuck for a while (several minutes) instead of starting in few seconds.

In a matter of coherence, I think it will be useful too as it avoids developer to redeclare the version manually and have to check it when upgrading spring boot dependencies version

I actually thought it was already done, because I did not declare version of the plugin in my project and it was working, but in fine maven was automatically selecting a plugin version which were older than the one managed by spring boot

dreis2211:
Hi,

this PR fixes #22486 .

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -944,6 +944,9 @@ bom {
 					exclude group: "ch.qos.logback", module: "logback-classic"
 				}
 			]
+			plugins = [
+				"liquibase-maven-plugin"
+			]
 		}
 	}
 	library("Log4j2", "2.13.3") {


wilkinsona:
Thanks very much, @dreis2211.

PR #22506, opened 2020-07-22
Update conditional inclusion of milestone repository

royclarkson:
The new Spring version schema does not include a "RELEASE" version. This
commit modifies the Gradle conditional to include the milestone repository
for all milestone and snapshot versions.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -17,7 +17,7 @@ allprojects {
 
 	repositories {
 		mavenCentral()
-		if (!version.endsWith('RELEASE')) {
+		if (version.contains('-')) {
 			maven { url "https://repo.spring.io/milestone" }
 		}
 		if (version.endsWith('-SNAPSHOT')) {


build.gradle
```diff
@@ -17,7 +17,7 @@ allprojects {
 
 	repositories {
 		mavenCentral()
-		if (!version.endsWith('RELEASE')) {
+		if (version =~ /((-M|-RC)[0-9]+|-SNAPSHOT)$/) {
```
wilkinsona, line 20: Thanks, Roy. I think we could simplify this to just look for a `-` in the version:

```java
if (version.contains('-')) {
```

WDYT?

royclarkson: Indeed, that works as well. I suppose there's no real benefit to being more specific with the regex in my commit.

settings.gradle
```diff
@@ -5,6 +5,9 @@ pluginManagement {
 		maven {
 			url 'https://repo.spring.io/plugins-release'
 		}
+		if (version.contains('-')) {
+			maven { url "https://repo.spring.io/milestone" }
+		}
```
wilkinsona, line 10: What's the motivation for adding this? We've thus far never had a milestone plugin dependency and I don't foresee having one in the future. Never say never, but I'd prefer to leave it out until it's needed.

royclarkson: Ah, I see. It was meant for consistency, but I didn't realize that we never publish plugin milestones. I'll remove it.

wilkinsona:
Thanks, Roy!

PR #22555, opened 2020-07-24
Add generated-snippets as output to test goal

runningcode:
The generated snippets were not declared as an output to the `spring-boot-actuator-autoconfigure` test goal so when the test was pulled from the cache, it did not contain the `generated-snippets` directory. This directory is required as an input to the asciidoctor plugin. This also declares that directory as an input to the asciidoctor plugin.

Note: This adds the `generated-snippets` directory as an output to _all_ the `default-test` goals.


wilkinsona:
Thanks very much, @runningcode.

PR #22557, opened 2020-07-24
Add condition that checks for the presence of a configurable annotation

rajvimalc:
**Enhancement for Conditional Annotations:**
When creating Auto Configurations we need to rely on @ConditionalOnProperty. This new annotation @ConditionalOnAnnotation serves as utility for conditional configuration with a custom enable annotation or an already existing annotation and acts similar to @ConditionalOnProperty but with annotation.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/AbstractAnnotationCondition.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/AbstractAnnotationCondition.java
@@ -16,13 +16,13 @@
 
 package org.springframework.boot.autoconfigure.condition;
 
-import org.springframework.context.annotation.Condition;
-import org.springframework.context.annotation.ConditionContext;
-import org.springframework.core.type.AnnotatedTypeMetadata;
-
 import java.lang.annotation.Annotation;
 import java.util.Map;
 import java.util.Objects;
+
+import org.springframework.context.annotation.Condition;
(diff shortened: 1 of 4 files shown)


snicoll:
@rajvimalc thanks for the PR. Please build locally before submitting as each commit to fix checkstyle, format or any other build checks trigger a new notification to all watchers of this repository.

rajvimalc:
@snicoll Where can I get checkstyle configurations for spring-boot? My local does not show any error since its not configured for spring builds.

snicoll:
@rajvimalc please run the build on the command line and not only in your IDE. That will make sure all the checks defined by the build are passing upfront. More details [on the contributing page](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc#code-conventions-and-housekeeping).

rajvimalc:
@snicoll Thank you for the guidance. Fixed all the issues and pushed a commit.

philwebb:
Thanks very much for the PR @rajvimalc, but I'm afraid we're not keen to add this functionality to Spring Boot itself. We've generally found it better to try and test on the outcome of annotations, rather than the presence of the annotations themselves. For example, if you have an `@Enable[Feature]` annotation, we prefer to use `@ConditionalOnBean` to look for presence of some `[Feature]Bean` that the `@Enable` annotation created. We also generally don't add `@Condition` annotations that we don't directly use ourselves in the Spring Boot codebase.

Thanks anyway for the effort that you put into the PR and I'm sorry we didn't get to discuss this before you started.

rajvimalc:
@philwebb Okay... Thank you for looking into it.

Issue #22559, opened 2020-07-25
Allow Jedis to be selected when both Jedis and Lettuce are on the classpath


brucelwl:
If both ``Lettuce`` and ``Jedis`` exist on the classpath, ``LettuceConnectionConfiguration`` and ``JedisConnectionConfiguration`` will be created, but actually ``JedisConnectionConfiguration`` is not necessary. It is recommended to add redis client type configuration, for example：
``spring.redis.client.type=jedis``，
``spring.redis.client.type=lettuce``

```java
@Configuration(proxyBeanMethods = false)
@ConditionalOnClass({ GenericObjectPool.class, JedisConnection.class, Jedis.class })
@ConditionalOnProperty(name = "spring.redis.client.type", havingValue = "jedis", matchIfMissing = true)
class JedisConnectionConfiguration extends RedisConnectionConfiguration {
      //...
}
```
```java
@Configuration(proxyBeanMethods = false)
@ConditionalOnClass(RedisClient.class)
@ConditionalOnProperty(name = "spring.redis.client.type", havingValue = "lettuce", matchIfMissing = true)
class LettuceConnectionConfiguration extends RedisConnectionConfiguration {
     //...
}
```

wilkinsona:
If both Lettuce and Jedis are on the classpath, Lettuce should win. `LettuceConnectionConfiguration` will define its `redisConnectionFactory` bean and the equivalent bean in `JedisConnectionConfiguration` should then back off as it is `@ConditionalOnMissingBean(RedisConnectionFactory.class)`.

If you have both Lettuce and Jedis on the classpath, there's no easy way to use Jedis and a property could help with that (and would align with what we've done for JDBC datasources). I'm not sure what you mean by "`JedisConnectionConfiguration` is invalid" though. Could you expand on that a bit please?

brucelwl:
@wilkinsona 
1. If both Lettuce and Jedis are on the classpath, Lettuce should win. But now it seems that i can't specify the reids client type
2. if Lettuce win,``JedisConnectionConfiguration`` is not necessary to register in the spring container

wilkinsona:
> if Lettuce win,JedisConnectionConfiguration is not necessary to register in the spring container

This could be addressed by moving the `@ConditionalOnMissingBean(RedisConnectionFactory.class)` from the `@Bean` method to the class.

> If both Lettuce and Jedis are on the classpath, Lettuce should win. But now it seems that i can't specify the reids client type

[...]

PR #22569, opened 2020-07-26
Allow Jedis to be selected when both Jedis and Lettuce are on the classpath

bono007:
Adds ability to choose Redis client when both Lettuce and Jedis on classpath.

Fixes gh-22559

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
@@ -76,6 +76,11 @@ public class RedisProperties {
 	 */
 	private String clientName;
 
+	/**
+	 * Type of client to use.
+	 */
+	private ClientType clientType = ClientType.Lettuce;
+
 	private Sentinel sentinel;
 
 	private Cluster cluster;
@@ -148,6 +153,14 @@ public class RedisProperties {
 		this.clientName = clientName;
 	}
 
+	public ClientType getClientType() {
+		return this.clientType;
+	}
+
+	public void setClientType(ClientType clientType) {
+		this.clientType = clientType;
+	}
+
 	public Sentinel getSentinel() {
 		return this.sentinel;
 	}
@@ -170,6 +183,22 @@ public class RedisProperties {
 
 	public Lettuce getLettuce() {
 		return this.lettuce;
+	}
+
+	/**
+	 * Type of Redis client to use.
+	 */
+	public enum ClientType {
+
+		/**
+		 * Use the Lettuce client
+		 */
+		Lettuce,
+
+		/**
+		 * Use the Jedis client
+		 */
+		Jedis
 	}
 
 	/**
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -76,6 +76,11 @@
 	 */
 	private String clientName;
 
+	/**
+	 * Client type to use - 'jedis' or 'lettuce'.
+	 */
+	private String clientType;
```
snicoll, line 82: I'd use an enum with `Jedis` and `Lettuce` here rather than a raw string. This also improves the developer experience in the IDE.

bono007: I will wait to see what shakes out of https://github.com/spring-projects/spring-boot/issues/22559#issuecomment-664897053 before updating.

snicoll: Thanks @bono007. We've decide to go ahead with the property so if you have time to update your proposal to use an enum that would be much appreciated.

cbo-indeed: Sounds good @snicoll. The property approach will be cleaner than the "move the auto-configs to the connection configs" I was proposing as that ended up being a bit messier than I originally thought when I prototyped it. 

I will get to it this evening.

bono007: Sounds good @snicoll. The property approach will be cleaner than the "move the auto-configs to the connection configs" I was proposing as that ended up being a bit messier than I originally thought when I prototyped it.

I will get to it this evening.

bono007: Done. I squashed and rebased as well.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -172,6 +185,22 @@ public Lettuce getLettuce() {
 		return this.lettuce;
 	}
 
+	/**
+	 * Type of Redis client to use.
+	 */
+	public enum ClientType {
+
+		/**
+		 * Use the Lettuce client
```
bono007, line 194: Not super thrilled about this comment - please update if you have a preference.

snicoll:
Thanks @bono007. I've polished things up in https://github.com/spring-projects/spring-boot/commit/589669d0cc49f05381cdf947e69804aa4d720a0d. In particular, I've removed the default value for `clientType` as it didn't match the code (the real default is that we configure the client based on the classpath).

bono007:
@snicoll good catch on the default. Thx.

Issue #22398, opened 2020-07-18
Lettuce: Add configuration to disable Redis Cluster dynamic sources refresh


DamianFekete:
In https://github.com/spring-projects/spring-boot/issues/15630 some Lettuce specific options were added:
```
spring.redis.lettuce.cluster.refresh.adaptive
spring.redis.lettuce.cluster.refresh.period
```

Please add a similar configuration option to disable the [dynamicRefreshSources](https://lettuce.io/core/5.1.0.M1/api/io/lettuce/core/cluster/ClusterTopologyRefreshOptions.html#useDynamicRefreshSources--). Maybe `spring.redis.lettuce.cluster.refresh.dynamic-sources` ?

To set this option we have to currently implement a `LettuceClientConfigurationBuilderCustomizer`, which also needs to set the previously mentioned clientOptions.

The option is needed to be able to (re-)connect to Redis using DNS names (in case nodes are restarted and get new IP addresses) (see https://github.com/lettuce-io/lettuce-core/issues/1343)

wilkinsona:
This sounds like a good idea to me, @DamianFekete. Thanks for the suggestion. @mp911de do you agree?

mp911de:
Makes sense.

wilkinsona:
Thanks, @mp911de.

snicoll:
Closing in favour of PR #22571

PR #22571, opened 2020-07-26
Add configuration to disable Redis Cluster dynamic sources refresh

bono007:
This commit adds an option to enable/disable the DynamicRefreshSources setting on the Lettuce cluster toplogy refresh options.

Closes gh-22398

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
@@ -410,6 +410,13 @@ public class RedisProperties {
 				 */
 				private boolean adaptive;
 
+				/**
+				 * Whether discovered nodes should be used as the source for the cluster
+				 * topology. When set to {@literal false}, only the initial seed nodes
+				 * will be used as sources for topology discovery.
+				 */
+				private Boolean dynamicSources = null;
+
 				public Duration getPeriod() {
 					return this.period;
 				}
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -410,6 +410,13 @@ public Refresh getRefresh() {
 				 */
 				private boolean adaptive;
 
+				/**
+				 * Whether discovered nodes should be used as the source for the cluster
+				 * topology. When set to {@literal false}, only the initial seed nodes
```
snicoll, line 415: You can't use Javadoc tag here as the annotation processor don't handle them.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -410,6 +410,13 @@ public Refresh getRefresh() {
...
+				 * will be used as sources for topology discovery.
+				 */
+				private Boolean dynamicSources = null;
```
snicoll, line 418: The `null` initialization here is unnecessary.  The AP will figure out the right default for a wrapper type.

bono007: I thought so but I noticed it explicitly specified in places like [this](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoProperties.java#L59). I will remove the redundant initialization.

snicoll:
Thanks again Chris. I've decided to switch the `Boolean` property a primitive as we can easily test that our default does not deviate from the default in Lettuce. See https://github.com/spring-projects/spring-boot/commit/903dc93887cc797a25cddd388a04edd87214dc77 for more info.

PR #22604, opened 2020-07-27
TaskExecutorBuilder only uses second-precision when configuring the await termination period

ttddyy:
Prior to this change, `TaskExecutorBuilder` used *second* as its precision to
map `awaitTerminationPeriod` value. This caused a loss of millisecond information
in the period.
This commit fixes the loss by converting the period to millisecond and use the
`setAwaitTerminationMillis` on executor/scheduler.


snicoll:
@wilkinsona I am not sure I'd qualify this as a bug. This feature did not exist in Spring Framework before `5.2.4`, see https://github.com/spring-projects/spring-framework/pull/24496

wilkinsona:
Thanks very much, @ttddyy.

PR #22631, opened 2020-07-28
Document properties to enable probe health indicators

ttddyy:
Document property names to enable probe(readiness/liveness) health indicators.


wilkinsona:
Related to this, the livenessstate and readinessstate indicators are listed twice – once in the table of auto-configured indicators and once in the table of indicators that are not enabled by default. I suspect they should only appear once.

I also wonder if we should list the full property (or just the ID with an accompany explanation to determine the property) for every indicator so that the information to enable or disable any of the indicators is available in one place?

bclozel:
I'd be in favor of adding that information in this table directly - we can find that key in the full application properties appendix but this feels like an unnecessary burden for the reader to check that other documentation section.

ttddyy:
I was working on readiness/liveness, so I was only looking them especially they are not enabled by default. But it is even better if all properties are documented on the table there.
It is easy for readers to have information there how to enable healthindicators (though other than readiness/liveness are enabled by default). 
I don't think it is easy for readers to realize there is an appendices that have all properties listed for actuator unless there is an explicit link in the section.

PR #22637, opened 2020-07-29
Optimized the code using Optional

abhishek-abhi:


wilkinsona:
Thanks for the suggestion but our preferred code style is to avoid using `Optional` in place of if statements.

PR #22638, opened 2020-07-29
added private constructor to enum

abhishek-abhi:
The Enums are required to have exclusively private constructors, this is because the Enum should be the only one responsible for returning the predefined instances.


wilkinsona:
Thanks for the suggestion, but there's no need for an enum to declare a private constructor. Unlike a classes, enums do not have a zero-args public constructor by default.

PR #22639, opened 2020-07-29
added null safe in setter

abhishek-abhi:


snicoll:
@Abdullah8006 `BeanNameAware` is a callback of the framework that makes sure a valid bean name is provided. It's unclear what lead you to create this PR or why you think that `''` is a good default. Feel free to join us on gitter if you want to chat about a particular feature .

PR #22643, opened 2020-07-29
Replaced HashMap with ConcurrentHashMap

abhishek-abhi:


wilkinsona:
Thanks for the proposal, but there's no need for the map to be a `ConcurrentHashMap` as it is never mutated. This is guaranteed by `AuditEvent` creating an unmodifiable map.

This is the fourth pull request of this kind that you've created today and, unfortunately, we'd have to decline them all. I'm sure that's frustrating for you and it's also time-consuming for us and everyone watching the repository. As @snicoll said, if you'd like to chat about things, please come and do that [on Gitter](https://gitter.im/spring-projects/spring-boot) first. That way we can hopefully avoid the overhead of creating a pull request that we won't want to merge.

PR #22646, opened 2020-07-29
Allow DurationFormat and PeriodFormat to be used on parameters

### LINKED ISSUE #22644: Allow DurationFormat and PeriodFormat to be used on parameters
We updated `@DurationUnit` in 2.4 to allow it to work with constructor binding, but we missed the `*Format` annotations.

dreis2211:
Hi,

this PR should close #22644 .

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java
@@ -785,16 +785,13 @@ class ConfigurationPropertiesTests {
 		source.put("test.duration", "12");
 		source.put("test.size", "13");
 		source.put("test.period", "14");
-		source.put("test.formattedDuration", "15d");
-		source.put("test.formattedPeriod", "16y");
 		sources.addLast(new MapPropertySource("test", source));
 		load(ConstructorParameterWithUnitConfiguration.class);
 		ConstructorParameterWithUnitProperties bean = this.context
 				.getBean(ConstructorParameterWithUnitProperties.class);
 		assertThat(bean.getDuration()).isEqualTo(Duration.ofDays(12));
 		assertThat(bean.getSize()).isEqualTo(DataSize.ofMegabytes(13));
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java
```diff
@@ -781,13 +785,16 @@ void loadWhenBindingToConstructorParametersWithCustomDataUnitShouldBind() {
...
+		source.put("test.formattedPeriod", "16y");
 		sources.addLast(new MapPropertySource("test", source));
 		load(ConstructorParameterWithUnitConfiguration.class);
 		ConstructorParameterWithUnitProperties bean = this.context
 				.getBean(ConstructorParameterWithUnitProperties.class);
 		assertThat(bean.getDuration()).isEqualTo(Duration.ofDays(12));
 		assertThat(bean.getSize()).isEqualTo(DataSize.ofMegabytes(13));
-		assertThat(bean.getPeriod()).isEqualTo(Period.ofYears(14));
+		assertThat(bean.getFormattedDuration()).isEqualTo(Duration.ofDays(15));
+		assertThat(bean.getFormattedPeriod()).isEqualTo(Period.ofYears(16));
```
wilkinsona, line 797: These assertions succeed without `@DurationFormat` and `@PeriodFormat` on `ConstructorParameterWithUnitProperties`. I think that's because the converter successfully infers the style to use from the value that's being converted. To verify that the format annotations are working as expected, I think it would be necessary to trigger a failure by trying, for example, to bind an ISO-8601 value to a property that requires the simple style.

dreis2211: I extended/reworked the testing a bit. Let me know if this is what you had in mind, @wilkinsona

wilkinsona: Spot on. Thanks.

PR #22684, opened 2020-07-31
Re-use the already built JAR file if possible

dsyer:
The build-image Mojo doesn't attempt to re-use an existing
JAR file that it has already built. This is a small change
that just checks if it already exists and re-uses it.

A side effect is that other plugins can modify the JAR file
(e.g. to optimize the reflection usage for a native-image
build) and those changes are visible in the container
image.

Possibly something similar can be done in the Gradle
plugin?

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/BuildImageMojo.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/BuildImageMojo.java
@@ -164,7 +164,11 @@ public class BuildImageMojo extends AbstractPackagerMojo {
 	}
 
 	private TarArchive getApplicationContent(Owner owner, Libraries libraries) {
-		ImagePackager packager = getConfiguredPackager(() -> new ImagePackager(getJarFile()));
+		File jarFile = getJarFile();
+		if (jarFile.exists()) {
+			return TarArchive.fromZip(jarFile, owner);
+		}
+		ImagePackager packager = getConfiguredPackager(() -> new ImagePackager(jarFile));
 		return new PackagedTarArchive(owner, libraries, packager);
 	}
 


wilkinsona:
> Possibly something similar can be done in the Gradle plugin?

By default, the Gradle plugin already uses the output of the `BootJar` task when creating the image. It should be possible to use a [dependency substitution](https://docs.gradle.org/current/dsl/org.gradle.api.artifacts.DependencySubstitutions.html) to swap in the optimised Tomcat dependency.

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/BuildImageMojo.java
```diff
@@ -164,7 +164,11 @@ private BuildRequest getBuildRequest(Libraries libraries) {
 	}
 
 	private TarArchive getApplicationContent(Owner owner, Libraries libraries) {
-		ImagePackager packager = getConfiguredPackager(() -> new ImagePackager(getJarFile()));
+		File jarFile = getJarFile();
+		ImagePackager packager = getConfiguredPackager(() -> new ImagePackager(jarFile));
```
dreis2211, line 168: The `packager` can be moved below the `if` as it doesn't seem to be used when the jar exists already, I guess.

dsyer: I assume someone is going to hack this change around a bit more, so it's probably re-arranging the proverbial deck chairs. But done ayway.

philwebb:
Unfortunately, I'm not convinced that this change will always work. There's some logic in `ImagePackager.packageImage` that ensures the non-repackaged source is always used. With the suggested change, we lose control over what actually gets uploaded and if the user hasn't created a fat jar, the OCI image will break. We also faced problems with Gradle when uploading packaged jars that include a launch script in the front.

I think if other optimizations are needed, we should look at some kind of dedicated API hook.

PR #22699, opened 2020-08-02
Use Supplier variants of Assert methods

dreis2211:
Hi,

this PR uses the `Supplier` variants for `Assert.state()` and alike where possible to avoid unnecessary String allocations.

Cheers,
Christoph


snicoll:
Thanks again @dreis2211!

PR #22710, opened 2020-08-03
Introduce @ForkedClassPath for testing unmodified class path

dreis2211:
Hi,

#22696 made `ModifiedClassPathExtension` public. Rather than doing that, this PR introduces a new annotation `@ForkedClassPath` that can be used instead. That seems also more consistent to the already existing `@ClassPathExclusions` & `@ClassPathOverrides`.

Let me know what you think.
Cheers,
Christoph


dreis2211:
Test failure seems to be an unrelated classpath conflict in RSocket

PR #22736, opened 2020-08-04
Add no-pull option for image building

### LINKED ISSUE #22679: Add no-pull option for image building
When building an image, the Maven and Gradle plugins will always attempt to pull the builder and run images from an image repository to the local Docker daemon. It is sometimes desirable to use a builder and/or run image that already exists in the local daemon instead of pulling from a remote repository (for example, when developing and testing a builder locally or just to save the time required to access the remote repo). 

The `pack build` CLI command pulls these images from a remote image repo by default but provides a `--no-pull` option to prevent remote repo access. The Maven and Gradle plugins should replicate this behavior.

anshlykov:
Closes #22679

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/AbstractBuildLog.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/AbstractBuildLog.java
@@ -41,23 +41,13 @@ public abstract class AbstractBuildLog implements BuildLog {
 	}
 
 	@Override
-	public Consumer<TotalProgressEvent> pullingBuilder(BuildRequest request, ImageReference imageReference) {
-		return getProgressConsumer(" > Pulling builder image '" + imageReference + "'");
+	public Consumer<TotalProgressEvent> pullingImage(ImageReference imageReference, ImageType imageType) {
+		return getProgressConsumer(String.format(" > Pulling %s '%s'", imageType.getDescription(), imageReference));
 	}
 
 	@Override
-	public void pulledBuilder(BuildRequest request, Image image) {
-		log(" > Pulled builder image '" + getDigest(image) + "'");
-	}
-
-	@Override
-	public Consumer<TotalProgressEvent> pullingRunImage(BuildRequest request, ImageReference imageReference) {
-		return getProgressConsumer(" > Pulling run image '" + imageReference + "'");
-	}
-
-	@Override
-	public void pulledRunImage(BuildRequest request, Image image) {
-		log(" > Pulled run image '" + getDigest(image) + "'");
+	public void pulledImage(Image image, ImageType imageType) {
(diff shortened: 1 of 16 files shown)


scottfrederick:
@anshlykov Thank you for the contribution. The changes look very good. 

As I was looking at implementing this myself, I started thinking that we could do better than replicating the `--no-pull` option from `pack build`. The semantics of `noPull=false` configuration in Gradle or a `<noPull>false</noPull>` configuration in Maven could be confusing to users because of the double-negative. 

Instead of a boolean "pull or no pull" option, we could model the Kubernetes behavior of a `pullPolicy` option with the possible values `ALWAYS`, `NEVER`, and `IF_NOT_PRESENT` where: 

* `ALWAYS` has the same behavior as your changes with `noPull=false` (the default)
* `NEVER` has the same behavior as your changes with `noPull=true`
* `IF_NOT_PRESENT` checks to see if the image exists in the local daemon and uses it if it does exist, otherwise pulls the image from the remote repository if it does not exist locally

This has the benefits of being familiar to Kubernetes users, providing one additional option for increased control, and making the configuration easier to understand. 

Would you mind updating the PR to implement this option?

anshlykov:
@scottfrederick thanks you for the review.

> instead of a boolean "pull or no pull" option, we could model the Kubernetes behavior of a pullPolicy option

It is a good idea. I will implement this option.

anshlykov:
@scottfrederick I added the implementation of the `PullPolicy` option. I also refactored the `BuildLog` interface a bit to reduce code duplication. Failed tests in CI are not relevant to this PR as far as I understand

spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/AbstractBuildLog.java
```diff
@@ -41,23 +41,13 @@ public void start(BuildRequest request) {
 	}
 
 	@Override
-	public Consumer<TotalProgressEvent> pullingBuilder(BuildRequest request, ImageReference imageReference) {
-		return getProgressConsumer(" > Pulling builder image '" + imageReference + "'");
+	public Consumer<TotalProgressEvent> pullingImage(ImageReference imageReference, ImageType imageType) {
```
scottfrederick, line 44: These are public methods of a public class, so we must maintain backward compatibility. Your changes do clean up the interface, and the methods are unlikely to be used outside of the Boot build plugins, but in the interest of compatibility we should leave the existing method signatures in place.

anshlykov: fixed

scottfrederick:
@anshlykov Thank you very much for the contribution. This will be a very useful feature. I've merged your changes along with [a commit](https://github.com/spring-projects/spring-boot/commit/6b15822cb13ae9d31bb3b1d0fc8c68c28bb64c71) to polish a few things, update the documentation, and add an integration test.

PR #22839, opened 2020-08-08
Fix some doc anchors

dreis2211:
Hi,

I just noticed a couple of broken doc anchors.

Cheers,
Christoph


snicoll:
Thanks Christoph. I've backported the IntelliJ IDEA separately as both `2.2.x` and `2.3.x` were affected too, see https://github.com/spring-projects/spring-boot/commit/8962d6ca21b39b5274b86ee3e832c3574eefdbfe

Issue #22776, opened 2020-08-07
StartupStep are closed twice in SpringApplicationRunListeners


mdeinum:
With the introduction of #22600 I was eager to test this with Flight Recorder based events. For this purpose, I cloned the Petclinic, updated it to Spring Boot 2.4.0-SNAPSHOT and confined it to run using the `FlightRecorderApplicationStartup`. However, when doing so it results in an exception during startup.

```
xception in thread "restartedMain" java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.springframework.boot.devtools.restart.RestartLauncher.run(RestartLauncher.java:49)
Caused by: java.util.NoSuchElementException
	at java.base/java.util.ArrayDeque.getFirst(ArrayDeque.java:403)
	at org.springframework.core.metrics.jfr.FlightRecorderApplicationStartup.start(FlightRecorderApplicationStartup.java:52)
	at org.springframework.boot.SpringApplicationRunListeners.doWithListeners(SpringApplicationRunListeners.java:111)
	at org.springframework.boot.SpringApplicationRunListeners.doWithListeners(SpringApplicationRunListeners.java:106)
	at org.springframework.boot.SpringApplicationRunListeners.environmentPrepared(SpringApplicationRunListeners.java:57)
	at org.springframework.boot.SpringApplication.prepareEnvironment(SpringApplication.java:356)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:318)
	at org.springframework.samples.petclinic.PetClinicApplication.main(PetClinicApplication.java:39)
	... 5 more
```

The events that are logged are 

1. `spring.boot.application.starting`, `sequenceId` = 0
2. `spring.boot.application.starting` (?), `sequenceId` = 0
3. `spring.boot.application.environment-prepared`, `sequenceId` = 1
4. `spring.boot.application.failed` , `sequenceId` = 2 -> which leads to the error due to the `currentSteps` `Deque` is empty.

[...]

PR #22845, opened 2020-08-10
Allow SpringApplicationRunListeners step action to be optional

mdeinum:
Prior to this commit the StartupStep.end method was being
called from the default step action. However when overriding
the default step action this might lead to the StartupStep.end
method not being called. As in the case of a failure, as that
enriches the information being written.

This commit also introduces a test for the failure case showing that
there is a missed call to end with the initial solution.

See: gh-22776

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplicationRunListeners.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplicationRunListeners.java
@@ -103,7 +103,8 @@ class SpringApplicationRunListeners {
 	}
 
 	private void doWithListeners(String stepName, Consumer<SpringApplicationRunListener> listenerAction) {
-		doWithListeners(stepName, listenerAction, StartupStep::end);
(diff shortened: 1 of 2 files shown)


mdeinum:
@bclozel There appears to be still a missing call to a `StartupStep.end()` in case of failure during startup, which that is I couldn't determine (might be one in Spring itself. 

Another note I couldn't run the tests locally it was hanging on the `BuildImageTests` (or at least it appeared that way as it didn't progress).

mdeinum:
#25572 fixes the missing call to end. It was the `spring.beans.instantiate` that wasn't ended when an error occurred.

philwebb:
This one was my fault, I messed up in a polish commit.

mdeinum:
Just noticed 4 test failures on Concourse-ci, however, it is 4 times the same test, appears that tests are either counted or run multiple times?

The test is failing due to the linked Spring Framework issue.

philwebb:
Thanks @mdeinum! I've tweaked the test for now until the Framework issue is fixed.

PR #22858, opened 2020-08-11
Prefer non-reflective access to ConfigurationClassPostProcessor

dsyer:
If the `ConfigurationClassPostProcessor` has been provided via an instance supplier then we can use that to set the `MetadataReaderFactory` non-reflectively, instead of using a PropertyValue (and the `BeanWrapper` eventually). 

The old behavior is left as a fallback - probably without a change in Spring Framework it will continue to be the default.


philwebb:
@dsyer I was looking to merge this but I'm a bit worried that it might cause early initialization of the `MetadataReaderFactory`. I'm not sure if calling `this.context.getBean(BEAN_NAME, MetadataReaderFactory.class)` from a `BeanDefinitionRegistryPostProcessor` is a good idea.

philwebb:
Oh wait, ignore me. I see you do that in the `Supplier`.

PR #22885, opened 2020-08-11
Support lambda based converters by parsing bean method signature generics

viviel:
If I define a Converter with lambda or method reference, the project will fail to start due to the failure of parsing generics. I don't think that's reasonable, so after understanding the reasons, I modified it to support some of the above features

```java
@Bean
public Converter<String, Integer> myConverter1() {
	return Integer::valueOf;
}
```

```java
@Bean
public Converter<String, Integer> myConverter2() {
	return e -> Integer.valueOf(e);
}
```

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/ApplicationConversionService.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/ApplicationConversionService.java
@@ -17,12 +17,23 @@
 package org.springframework.boot.convert;
 
 import java.lang.annotation.Annotation;
+import java.util.Collections;
 import java.util.LinkedHashSet;
+import java.util.Map;
+import java.util.Optional;
 import java.util.Set;
 
 import org.springframework.beans.factory.ListableBeanFactory;
+import org.springframework.beans.factory.config.BeanDefinition;
+import org.springframework.beans.factory.config.ConfigurableListableBeanFactory;
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.i18n.LocaleContextHolder;
+import org.springframework.core.GenericTypeResolver;
+import org.springframework.core.ResolvableType;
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the pull request, @viviel. Could you please add some tests that illustrate the problem that you changes solve?

viviel:
OK, then I'll look at the CI problem

wilkinsona:
Thank you. Please don't worry too much about the CI failure. There's a known problem with the latest Spring Data snapshots that the Spring Data team are working on at the moment. A quick look at the failures suggests that many, if not all of them, are related to that.

viviel:
OK, I'll add a test case later

philwebb:
We might want to look at the logic in `ConfigurationPropertiesBean` when we merge this.

viviel:
@wilkinsona 
I'm very sorry that I didn't pay attention to GitHub's email recently due to work reasons, so I didn't notice your message. 
If necessary and the master branch does not solve this problem, I am happy to continue trying to solve it. Re sort out the cause of this problem and the code logic of PR. By the way, resolve the conflict.

wilkinsona:
Thanks, @viviel. We'll see if we can merge this in 2.7.

viviel:
I'd like to talk about the plan I expected. You can see if it can be implemented in this way. If so, I'll re implement it from the branch of 2.7 and submit it.

I've combed the code these two days and found that there are similar problems with four interfaces: Converter, Formatter, Printer and Parser. My idea is to uniformly modify the four interfaces. The specific scheme is to make an adaptation layer for each interface, put the generic information in it, and then uniformly use the `void addConverter(GenericConverter converter)` method to register.

I'll implement it according to this scheme, and then you can see the specific code logic. If not, adapt the current changes to the code of the 2.7 branch

philwebb:
Thanks for your patience @viviel. This has finally been merged for 3.5.x.

PR #22896, opened 2020-08-11
Customize RestTemplateBuilder

bedla:
Hi,
I found that it would be handy to customize `RestTemplateBuilder` created by auto-configuration. It has some useful method that sometimes I want to call. Currently I have to use bean post processor to customize it - I think that it is ugly and error prone.
What do you think about this proposal?
Thx
Ivos

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/client/RestTemplateAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/client/RestTemplateAutoConfiguration.java
@@ -33,12 +33,14 @@ import org.springframework.boot.autoconfigure.http.HttpMessageConverters;
 import org.springframework.boot.autoconfigure.http.HttpMessageConvertersAutoConfiguration;
 import org.springframework.boot.autoconfigure.web.client.RestTemplateAutoConfiguration.NotReactiveWebApplicationCondition;
 import org.springframework.boot.web.client.RestTemplateBuilder;
+import org.springframework.boot.web.client.RestTemplateBuilderCustomizer;
 import org.springframework.boot.web.client.RestTemplateCustomizer;
 import org.springframework.boot.web.client.RestTemplateRequestCustomizer;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Lazy;
+import org.springframework.util.Assert;
 import org.springframework.web.client.RestTemplate;
 
 /**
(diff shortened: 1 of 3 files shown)


bedla:
there is lots of `java.io.IOException: No space left on device` error. could somebody fix CI?

bedla:
@wilkinsona sorry for disturbing but I found that you mentioned at comment https://github.com/spring-projects/spring-boot/pull/22885#issuecomment-671964861 that CI failures are general problem. Mine PF fails at `No space left on device` as mentioned above... is it same problem and will be fixed?

philwebb:
@bedla Sorry about the CI failures. We had some issues yesterday.

philwebb:
Flagging for team attention to see what the rest of the team think about the proposal. I'm a little concerned that the immutable nature of `RestTemplateBuilder` makes the `Customizer` slightly unusual. I think most of our existing ones have a `void` return and expect the customizable thing to be mutable.

Perhaps we need a different name for it. Another option might be have an optional `RestTemplateBuilderFactory` that's used to create the initial instance.

@bedla out of interest, which  `RestTemplateBuilder` methods do you typically want to customize?

bedla:
Yep, that name is bit unusal as you mentioned, but I cannot imagine something more suitable. 

My usecase is that I wan to use all the stuff (otherwise I have to copy-paste it) from auto-configuration (see `RestTemplateAutoConfiguration#restTemplateBuilder` method) and mainly call methods `setBufferRequestBody`, `setConnectTimeout` and `setReadTimeout` (sometimes `defaultHeader`)

snicoll:
Thanks for the PR @bedla and the follow-up. I am personally not super keen on the idea of the customizer. We discussed this at our team meeting a couple of weeks ago and I mentioned a similar pattern we use to further tune message listener container factories. 

I've pushed some code that mimics that pattern for `RestTemplateBuilder`, see https://github.com/snicoll/spring-boot/commit/0d06a84b1b6ac82fb273d03f1355064945b6876c

One downside with this approach is that we expose an additional bean for the purpose of customization. I wish we could provide a more efficient option. Anyway, foo for thoughts @philwebb

bedla:
@snicoll no problem, I am happy to help. Anyways I have checked commit you mentioned and maybe I am missing something, but I still I am unable to customize that builder, right? 
Those `setBufferRequestBody`, `setConnectTimeout` and `setReadTimeout` (sometimes `defaultHeader`) are interesting for me.

snicoll:
This test showcases that exact use case unless I am missing something: https://github.com/snicoll/spring-boot/commit/0d06a84b1b6ac82fb273d03f1355064945b6876c#diff-cfe5e8d68c724e41762c869df2ddaec6R241

bedla:
ahh, I see. Thx, Ivos

snicoll:
@bedla thank you for the suggestion and the PR. We've discussed that at our team meeting and we've decided to give my proposal a go. I am going to close this PR now, please follow-up on #23389

PR #22898, opened 2020-08-12
Document how to discover the HTTP Port at Runtime for both Servlet and Reactive web apps

johncfranco:
The section about retrieving the bound port for an embedded web server had only the proper interface for servlet-based servers, not for reactive servers.  It took me a while to figure it out, so adding what I found here might save other people the time that it cost me.


snicoll:
@johncfranco thank you for making your first contribution to Spring Boot. It turns out that How-To was a little bit outdated anyway as we have a common abstraction between Servlet-based and Reactive web apps. I've [polished your contribution](https://github.com/spring-projects/spring-boot/commit/ee2363fca724677b077311f0e5a8700fc20784cf) to mention how you can listen to a common event that works in both arrangements.

PR #22918, opened 2020-08-12
Fix Gradle Plugin custom image name example and documentation

jnizet:
There is no `artifactId` property in a gradle Project, and the default value uses the project name.


snicoll:
Thanks again @jnizet!

Issue #22902, opened 2020-08-12
PushGateway documentation does not document enabled flag properly


snicoll:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before, or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Problem
Prometheus's PushGateway support can be enabled if an additional library is present and if the `management.metrics.export.prometheus.pushgateway.enabled` property is set to `true`. Unfortunately, that last condition is not documented in the reference guide.

## Solution
The "Prometheus" section of the reference guide currently contains:

> When the Prometheus Pushgateway dependency is present on the classpath, Spring Boot auto-configures a `PrometheusPushGatewayManager` bean.

It should rather be:

> When the Prometheus Pushgateway dependency is present on the classpath and the configprop:management.metrics.export.prometheus.pushgateway.enabled[] property is set to `true`, a `PrometheusPushGatewayManager` bean is auto-configured.

The documentation should be updated in `src/docs/asciidoc/production-ready-features.adoc` from the `spring-boot-docs` module.

Also, to make sure the documentation matches the expectation, an additional test should be added in `PrometheusMetricsExportAutoConfigurationTests`:

```java
@Test
void pushGatewayIsNotConfiguredWhenEnabledFlagIsNotSet() {
  this.contextRunner.withUserConfiguration(BaseConfiguration.class)
      .run((context) -> assertThat(context).doesNotHaveBean(PrometheusPushGatewayManager.class));
}
```

## Steps to Fix 
 
- [x] Claim this issue with a comment below and ask any clarifying questions you need
- [x] Set up a repository locally following the [Contributing Guidelines](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc)
- [x] Try to fix the issue following the steps above
- [x] Commit your changes and start a pull request.

ravening:
@snicoll I'm interested in working on this

snicoll:
@ravening thank you for the interest but your profile indicate that you've already contributed quite a lot so I'd prefer to keep this one for someone who's new to contributing and OSS in general. Feel free to look at other issues in our tracker and I am happy to assist if you need help.

[...]

PR #22919, opened 2020-08-12
PushGateway documentation does not document enabled flag properly

smlc:
Fix issue #22902


snicoll:
@smlc thank you for making your first contribution to Spring Boot.

Issue #22942, opened 2020-08-13
Mark `bootArchives` as nonresolvable configuration in Gradle plugin to better convey its purpose.


chali:
If my understanding is correct `bootArchives` configuration in spring-boot gradle plugin is supposed to be used as a place where artifacts for publishing are placed. If this is correct, I would suggest marking it as non-resolvable. It would help in interactions with other gradle plugins which might manipulate all resolvable configurations and ignoring `bootArchives` requires special handling.

wilkinsona:
That's a good suggestion. Thank you. For future reference, please feel free to jump straight to a PR unless you'd like to discuss a change before implementing it. Closing in favour of #22943.

PR #22943, opened 2020-08-13
Mark `bootArchives` as nonresolvable configuration in Gradle plugin to better convey its purpose

chali:
PR for #22942.


wilkinsona:
@chali Thanks very much for making your first contribution to Spring Boot.

PR #22948, opened 2020-08-14
Fix typo `Assert4J` -> `AssertJ`

maciejwalkowiak:
The mentioned library is called AssertJ (https://joel-costigliola.github.io/assertj/)


philwebb:
Good spot. Thanks!

Issue #22836, opened 2020-08-08
Support authentication from Docker daemon to private docker registry


wmz7year:
spring-boot-maven-plugin 2.3.x version support build layers docker image.

but use custom builder image or run image with private docker registry it can't works

like this:

```
			<plugin>
				<groupId>org.springframework.boot</groupId>
				<artifactId>spring-boot-maven-plugin</artifactId>
				<executions>
					<execution>
						<goals>
							<goal>build-info</goal>
						</goals>
					</execution>
				</executions>
				<configuration>
					<layers>
						<enabled>true</enabled>
					</layers>
					<image>
						<builder>xxxxx.dkr.ecr.cn-north-1.amazonaws.com.cn/paketo-buildpacks-builder:base-platform-api-0.3</builder>
					</image>
				</configuration>
			</plugin>
```

then run ./mvnw spring-boot:build-image will throw 

```
 Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.3.2.RELEASE:build-image (default-cli) on project athena-marketing-server-start: Execution default-cli of goal org.springframework.boot:spring-boot-maven-plugin:2.3.2.RELEASE:build-image failed: Docker API call to 'localhost/v1.24/images/create?fromImage=882440569488.dkr.ecr.cn-north-1.amazonaws.com.cn%2Fpaketo-buildpacks-builder%3Abase-platform-api-0.3' failed with status code 500 "Internal Server Error" and message "Get https://xxxxx.dkr.ecr.cn-north-1.amazonaws.com.cn/v2/paketo-buildpacks-builder/manifests/base-platform-api-0.3: no basic auth credentials" -> [Help 1]
```

the docker registry xxxxx.dkr.ecr.cn-north-1.amazonaws.com.cn is private and authentication to pull docker images.

spring boot maven plugin call docker api does't have any auth params

https://github.com/spring-projects/spring-boot/blob/4933b2688fd2114c75f8f4835d9f27bbd08165c9/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/transport/HttpClientTransport.java#L126

docker engine api provide ```X-Registry-Auth``` Header to authentication https://docs.docker.com/engine/api/v1.39/#section/Authentication

```
Authentication
[...]

Issue #21001, opened 2020-04-17
Add support for publishing docker images to a registry


philwebb:
Currently we don't have the option to publish a locally built docker image to a remote repository. You can do it using the docker CLI, but it would be nice to have some built in support.

jayanth-iyer:
Will there be any support to create K8s manifests automatically -skeletons - just using the mvn command - part of native build pack? Later the developer can just add the values and add and remove the keys from the yaml
Does it even make sense to add this support? - want to understand

jorgemoralespou:
I don't think this use case should be something the spring-boot maven plugin should do. If it did, shouldn't it also need to include goals to deploy the application/image into different platforms? 
I think that users should stick to using multiple plugins rather than have one single plugin that does it all (not saying can't be provided by spring-boot). I saw that pattern in fabric8 maven plugin, and they had to slowly rollback and break that plugin into smaller ones, more single purposed.
My 2 cents.

jorgemoralespou:
Then my bad, as there's no description in the issue, it might seem confusing for others as it's been for me. Maybe @philwebb can add the description so people understand better what they are voting/commenting on.

philwebb:
Sorry about that, I threw in the issue for my own reference and wasn't expecting any votes/comments. I've added a brief description.

jorgemoralespou:
I think then thy my comment is not outdated. https://github.com/spring-projects/spring-boot/issues/21001#issuecomment-686548109

philwebb:
@jorgemoralespou Can you elaborate a bit more on the problems we might face? I was originally thinking we might be able to shell out and call `docker push`. I think that's currently how jib deals with it.

jorgemoralespou:
Well, there's 2 sides:
- Technical: You'd need to be able to support private and insecure registries, and you'd need to interface with Docker on Windows/Mac/Linux in different ways, so better to not shell out to docker but rather use a library that can deal with the pushing direclty.
[...]

PR #22972, opened 2020-08-16
Support authentication from Docker daemon to private docker registry.

wmz7year:
see: #22836

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Builder.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Builder.java
@@ -24,6 +24,7 @@ import org.springframework.boot.buildpack.platform.docker.DockerApi;
 import org.springframework.boot.buildpack.platform.docker.TotalProgressEvent;
 import org.springframework.boot.buildpack.platform.docker.TotalProgressPullListener;
 import org.springframework.boot.buildpack.platform.docker.UpdateListener;
+import org.springframework.boot.buildpack.platform.docker.configuration.DockerConfiguration;
 import org.springframework.boot.buildpack.platform.docker.transport.DockerEngineException;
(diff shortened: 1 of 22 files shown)


crimsonvspurple:
If I'm understanding it right, this is needed to  `bootBuildImage` push to ECR?
Any idea, when it will be in next snapshot?

Thank you in advance.

trajano:
Is there any documentation on how to use this?

snicoll:
> Is there any documentation on how to use this?

@trajano this issue is open and unresolved so there's nothing to be documented yet I am afraid.

trajano:
I mean shouldn't this PR also include a change in the documentation on how to use it?

scottfrederick:
Merging the PR is in progress, with some changes and added documentation.

scottfrederick:
@wmz7year Thank you very much for the contribution. I made some changes in https://github.com/spring-projects/spring-boot/commit/86fa8144f57ecfa59d02400a4ad26cd6fb297fe1 to refactor the code to better match project conventions and to add documentation. 

It took a while to test this, because using the `identitytoken` authentication [described in the Docker Engine documentation](https://docs.docker.com/engine/api/v1.39/#section/Authentication) does not appear to be supported by the Docker registries I was testing against. Any additional testing and feedback anyone can provide with these authentication options would be appreciated.

scottfrederick:
> If I'm understanding it right, this is needed to `bootBuildImage` push to ECR?

@crimsonvspurple Apologies for missing this question earlier. The `bootBuildImage` task won't push the resulting image to a Docker registry, it only creates the image in the local Docker daemon. This feature is needed if custom CNB builder or run images are used and those custom images need to be pulled from a private Docker registry.

crimsonvspurple:
I could not use spring's image builder to push to ECR even after trying a lot of things. It builds as long as I don't try to push to ECR. The errors always came down to some auth header being missing. I wasn't using any private image but built image was being pushed to private ECR.

I had to switch to jib which worked nicely.

I will give it another shot later.

wilkinsona:
@crimsonvspurple Support for pushing to a registry was added in 2.4.0-M4. Please see https://github.com/spring-projects/spring-boot/issues/21001 for details.

Issue #22364, opened 2020-07-16
Auto-configure ConcurrentKafkaListenerContainerFactory with a user-provided RecordFilterStrategy bean if available


tjk5567764363:
Version: 2.3.0.RELEASE

I want to implement a RecordFilterStrategy, But did not take effect

```java
@Component
public class RecordFilter implements RecordFilterStrategy{

    @Override
    public boolean filter(ConsumerRecord consumerRecord) {
        // do something
        return false;
    }
}
```

Is it possible to modify org.springframework.boot.autoconfigure.kafka.KafkaAnnotationDrivenConfiguration#kafkaListenerContainerFactory

like this

```java
@Bean
@ConditionalOnMissingBean(name = "kafkaListenerContainerFactory")
ConcurrentKafkaListenerContainerFactory<?, ?> kafkaListenerContainerFactory(
        ConcurrentKafkaListenerContainerFactoryConfigurer configurer,
        ObjectProvider<ConsumerFactory<Object, Object>> kafkaConsumerFactory,
        ObjectProvider<RecordFilterStrategy<Object, Object>> recordFilterStrategy) { // new add
    ConcurrentKafkaListenerContainerFactory<Object, Object> factory = new ConcurrentKafkaListenerContainerFactory<>();
    recordFilterStrategy.ifAvailable(factory::setRecordFilterStrategy); // new add
    configurer.configure(factory, kafkaConsumerFactory
            .getIfAvailable(() -> new DefaultKafkaConsumerFactory<>(this.properties.buildConsumerProperties())));
    return factory;
}
```

other issues in spring-kafka project
https://github.com/spring-projects/spring-kafka/issues/1535#issue-658024709

wilkinsona:
Yes, I think that makes sense. Thanks for the suggestion.

snicoll:
Closing in favour of PR #22973

PR #22973, opened 2020-08-16
Auto-configure ConcurrentKafkaListenerContainerFactory with a user-provided RecordFilterStrategy bean if available

anshlykov:
Closes gh-22364


snicoll:
@anshlykov thank you for the PR. We have a configurer one can use to create a custom factory with the things we've auto-configured. This should have been applied there rather than directly on the connection factory. I've polished things up in https://github.com/spring-projects/spring-boot/commit/8c0318edc797b20eb95e30ae09772c33b4675ab2

tjk5567764363:
@snicoll  can support multi `RecordFilterStrategy`?  such as `CompositeRecordFilterStrategy`

snicoll:
@tjk5567764363 I am not sure what you asking or what `CompositeRecordFilterStrategy` is. The issue updates the documentation that mentions a bean of type `RecordFilterStrategy` is considered. If your implementation uses a composite pattern in its implementation that'll work fine as long as you have a primary bean that implements `RecordFilterStrategy`.

PR #23002, opened 2020-08-19
Mention correct JUnit 5 annotations in Kotlin testing section

dreis2211:
Hi,

I just noticed that the Kotlin testing docs mention the JUnit 4 classes `@BeforeClass` and `@AfterClass` which seems to be wrong given that it describes the benefits of JUnit 5.

Cheers,
Christoph


snicoll:
Good catch, thanks again @dreis2211

PR #21624, opened 2020-05-29
Add Cassandra driver Actuator metrics

emerkle826:
Motivation:

The DataStax Java driver for Cassandra can be configured to expose driver session operational metrics. Currently these metrics are not available through Spring Boot Actuator.

Modifications:

This commit adds support for Cassandra driver metrics through Spring Boot Actuator by binding the native driver metrics to Micrometer metrics.

Result:

Users are now able to access Cassandra driver enabled metrics through Actuator .

PR #23008, opened 2020-08-19
Add Cassandra driver Actuator metrics

emerkle826:
Motivation:

The DataStax Java driver for Cassandra can be configured to expose driver session operational metrics. Currently these metrics are not available through Spring Boot Actuator.

Modifications:

This commit adds support for Cassandra driver metrics through Spring Boot Actuator by enabling the Cassandra driver's Micrometer metrics module.

Result:

Users are now able to access Cassandra driver enabled metrics through Actuator.

**NOTE:** This PR uses the Cassandra driver's native Micrometer metrics as opposed to the previous approach [here](https://github.com/spring-projects/spring-boot/pull/21624).


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsPostProcessor.java
```diff
@@ -0,0 +1,116 @@
...
+import org.springframework.beans.factory.config.BeanPostProcessor;
+import org.springframework.boot.autoconfigure.cassandra.CassandraProperties;
+import org.springframework.boot.autoconfigure.cassandra.CqlSessionBuilderCustomizer;
+import org.springframework.context.ApplicationContext;
+import org.springframework.core.Ordered;
+
+/**
+ * {@link BeanPostProcessor} that configures Cassandra metrics. This is necessary as the
+ * CqlSessionBuilder bean provided by {@link CassandraAutoConfigure} must be overridden in
+ * order to enable driver metrics. The post-processing here will do just that.
```
snicoll, line 39: There is a `CqlSessionBuilderCustomizer` that should be used for this. Please adapt the contract in the Cassandra API so that an override of the builder is not necessary.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfigurationIntegrationTests.java
```diff
@@ -0,0 +1,149 @@
...
+					.isGreaterThan(1d);
+
+		});
+	}
+
+	@Configuration(proxyBeanMethods = false)
+	static class SimpleDriverConfigLoaderBuilderMetricsConfig {
+
+		@Bean
+		DriverConfigLoaderBuilderCustomizer customizer() {
```
snicoll, line 113: If that represents what users have to do in order to get metrics support, I think something is missing.

snicoll (commented):
Thanks for the PR. I had a quick first look and I am not keen to implement this via a `BeanPostProcessor` before investigating other options. I've added a few comments.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsPostProcessor.java
emerkle826: Thank you for the review. Currently, metrics binding in the driver happens a little deeper than the Session builder level (at the DriverContext level). I'll look into trying to modify the driver code to allow for using the Customizer.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfigurationIntegrationTests.java
emerkle826: For Driver 4.x, metrics are not enabled by default. To enable metrics in the driver, the driver must be configured with the list of enabled metrics. This can be done programmatically, with System properties or by providing a driver config file. For this test, it was easiest to just provide a Driver config customizer bean. I can change the test to enable the metrics under test via System properties or with an external driver configuration file if you prefer.

Are you concerned with how metrics are enabled in the driver? Or should I be enabling metrics automatically in the Actuator?

snicoll: > Are you concerned with how metrics are enabled in the driver? Or should I be enabling metrics automatically in the Actuator?

If we provide first-class support for metrics here, the user shouldn't have to write code to enable it ideally. Looking at other metrics we have, they are usually enabled by default, sometimes with a flag that indicates if metrics for that particular `CqlSession` is enabled or not. Concretely we should then have a `enabled` property somewhere in the `spring.cassandra` namespace that user can set in `application.properties` the usual way.

I don't have an opinion as whether the flag should be enabled by default or not but listing the metrics seem a bit tedious and inconsistent for an "out-of-the-box" scenario.

Paging @shakuzen to get some more feedback on this.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsPostProcessor.java
snicoll: Thank you, much appreciated.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfigurationIntegrationTests.java
shakuzen: I'm not very familiar with Cassandra driver specifics, but it's probably only metrics on specific queries which might have high cardinality or privacy concerns (if query parameters/values are tagged, for example) that we wouldn't want to enable by default, I think. Connection pool or node or session metrics seem generally useful and probably safe from the aforementioned concerns.

snicoll:
@emerkle826 how is it going? Do you have an update for this change? Thanks!

emerkle826:
@snicoll I should have an update to this PR in a day or 2. I was waiting for the 4.9.0 release of the DataStax driver, which now has the ability to bind in the metrics with a Customizer.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfigurationIntegrationTests.java
```diff
@@ -0,0 +1,127 @@
...
+			// assert Node metrics
+			String nodeMetricPrefix = sessionMetricPrefix + "nodes." + cassandra.getHost() + ":"
+					+ cassandra.getMappedPort(9042) + ".";
+			assertThat(registry.get(nodeMetricPrefix + DefaultNodeMetric.BYTES_SENT.getPath()).counter().count())
+					.isGreaterThan(1d);
+
+		});
+	}
+
+	static final class CassandraWaitStrategy extends AbstractWaitStrategy {
```
snicoll, line 102: Unrelated but rather than copy/pasting this, I wonder if it wouldn't be better if Testcontainers had some sort of support for this. Feel free to chime in [in the issue I've just created](https://github.com/testcontainers/testcontainers-java/issues/3224).

emerkle826: Testcontainers does heave a Cassandra wait strategy implementation [here](https://github.com/testcontainers/testcontainers-java/blob/master/modules/cassandra/src/main/java/org/testcontainers/containers/wait/CassandraQueryWaitStrategy.java) which is pretty similar to the implementation I copied. I could swap the implementations in Spring for the one Testcontainers provides.

That said, I have an issue to try to update Testcontainers from Driver 3.x to Driver 4.x [here](https://github.com/testcontainers/testcontainers-java/issues/2768). My [PR is under review](https://github.com/testcontainers/testcontainers-java/pull/2830), though it hasn't gotten a lot of traction. The reason I mention it is because I haven't yet addressed the Driver 3.x API tie in that is used in their wait strategy. From a Testcontainer perspective, they would prefer to remove the coupling to specific Cassandra Driver APIs (so driver version changes in the future don't impact Testcontainers). Providing a Driver API agnostic wait strategy would have to be included to achieve that.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+@Configuration
+@AutoConfigureAfter({ MetricsAutoConfiguration.class, SimpleMetricsExportAutoConfiguration.class })
+@AutoConfigureBefore({ CassandraAutoConfiguration.class })
+@ConditionalOnClass({ MeterRegistry.class, MicrometerMetricsFactory.class })
+@ConditionalOnBean({ MeterRegistry.class })
+public class CassandraMetricsAutoConfiguration {
+
+	@Bean
+	public DriverConfigLoaderBuilderCustomizer cassandraMetricsConfigCustomizer() {
+		return (builder) -> builder.withString(DefaultDriverOption.METRICS_FACTORY_CLASS, "MicrometerMetricsFactory")
```
snicoll, line 57: You can import and use the class here, `@ConditionalOnClass` is evaluated without loading the class itself.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,70 @@
...
+				.withStringList(DefaultDriverOption.METRICS_SESSION_ENABLED,
```
snicoll, line 58: I've tried to write some more test where someone would want to override those settings. Unfortunately that didn't work as calling `withString` with the same key in another customizer leads to:

```
Caused by: java.lang.IllegalArgumentException: Duplicate key datastax-java-driver.advanced.metrics.session.enabled
	at com.datastax.oss.protocol.internal.util.collection.NullAllowingImmutableMap$Builder.failIfDuplicateKeys(NullAllowingImmutableMap.java:226)
	at com.datastax.oss.protocol.internal.util.collection.NullAllowingImmutableMap$Builder.build(NullAllowingImmutableMap.java:198)
	at com.datastax.oss.driver.internal.core.config.typesafe.DefaultProgrammaticDriverConfigLoaderBuilder.buildConfig(DefaultProgrammaticDriverConfigLoaderBuilder.java:263)
	at com.datastax.oss.driver.internal.core.config.typesafe.DefaultProgrammaticDriverConfigLoaderBuilder.lambda$build$2(DefaultProgrammaticDriverConfigLoaderBuilder.java:248)
	at com.datastax.oss.driver.internal.core.config.typesafe.DefaultDriverConfigLoader.<init>(DefaultDriverConfigLoader.java:131)
	at com.datastax.oss.driver.internal.core.config.typesafe.DefaultDriverConfigLoader.<init>(DefaultDriverConfigLoader.java:117)
	at com.datastax.oss.driver.internal.core.config.typesafe.DefaultProgrammaticDriverConfigLoaderBuilder.build(DefaultProgrammaticDriverConfigLoaderBuilder.java:245)
	at org.springframework.boot.autoconfigure.cassandra.CassandraAutoConfiguration.cassandraDriverConfigLoader(CassandraAutoConfiguration.java:114)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:62)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:498)
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:154)
```

I think we need a way to support this use case, we can't blindly set those values without a way for users to customise them. Is there a way to know that a key has been registered already (and therefore not apply anything) or a way for users to override whatever value has been set? 

I'd prefer the former as it makes the auto-configuration backs off in a consistent way. If only the latter is possible, that would mean that the customizer should be ordered to run after the auto-configured one which is more work than it should IMO.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfigurationIntegrationTests.java
```diff
@@ -0,0 +1,96 @@
...
+			for (int i = 0; i < 10; ++i) {
+				assertThat(session.execute(statement).one()).isNotNull();
+			}
+			// assert Session metrics
+			String sessionMetricPrefix = session.getContext().getSessionName() + ".";
+			assertThat(
+					registry.get(sessionMetricPrefix + DefaultSessionMetric.CONNECTED_NODES.getPath()).gauge().value())
+							.isEqualTo(1d);
+			assertThat(registry.get(sessionMetricPrefix + DefaultSessionMetric.CQL_REQUESTS.getPath()).timer().count())
+					.isEqualTo(10L);
```
snicoll, line 81: I was a  bit confused  by that. The `waitStrategy` runs some CQL requests as well, doesn't it? That makes this test a bit fragile IMO.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfigurationIntegrationTests.java
```diff
@@ -0,0 +1,96 @@
...
+	@Test
+	void autoConfiguredCassandraIsInstrumented() {
+		this.contextRunner.run((context) -> {
+			MeterRegistry registry = context.getBean(MeterRegistry.class);
+			// execute queries to peg metrics
+			CqlSession session = context.getBean(CqlSession.class);
+			SimpleStatement statement = SimpleStatement.newInstance("SELECT release_version FROM system.local")
+					.setConsistencyLevel(ConsistencyLevel.LOCAL_ONE);
+			for (int i = 0; i < 10; ++i) {
+				assertThat(session.execute(statement).one()).isNotNull();
```
snicoll, line 73: Why do you need to loop 10 times here?

snicoll (commented):
Thanks for the update and the change in the API. This is much simpler now and more aligned with what I was expecting.

I've tried to write some more tests as I want to make sure that users can easily override whatever the auto-configuration configures by default but I hit a problem. Can you please take a look and guide me please?

snicoll:
I should also mentioned that I've started to polish the proposal already so there is no need to push additional commits. I'd like some feedback on what API I should be using so that users can override those settings if they want to.

snicoll:
@emerkle826 @adutra any feedback on the review for me? In particular, not being able to see if the user has configured something is very annoying so I could use your insights. Thank you!

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfiguration.java
emerkle826: I'm using the class name here because the driver configuration is looking for that, not the class itself. Are you suggesting that I do something like:
```java
return (builder) -> builder.withString(DefaultDriverOption.METRICS_FACTORY_CLASS, MicrometerMetricsFactory.class.getName())
```

I could do that if you prefer.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfigurationIntegrationTests.java
emerkle826: I loop here just to peg the request count more than once and verify in the assertion below. 10 was just some arbitrary value that was more than 1.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfiguration.java
snicoll: It was just a note for you. See [this comment](https://github.com/spring-projects/spring-boot/pull/23008#issuecomment-698264174). And yes, it's better to use the class than a "magic string". I've already polished this locally so no need to do anything.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfigurationIntegrationTests.java
emerkle826: Testcontainers wait strategy does execute a query, whereas the wait strategy implementation in the Spring Boot Cassandra Autoconfigure test does not. The query is basic (it just queries the database version) and should succeed when the Cassandra cluster is up and ready. If the query fails or times out, it's likely due to the cluster not being available, which would mean that this integration test likely won't succeed either.

The wait strategy implementation I copied and removed just waits for a Cassandra session builder to be built, which doesn't guarantee that the cluster is ready to respond to queries (although I haven't had that happen in practice locally). Maybe the ideal solution is to update the wait strategy implementation in Testcontainers so that it is more similar to the implementation in the Spring Boot test [here](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cassandra/CassandraAutoConfigurationWithPasswordAuthenticationIntegrationTests.java#L98-L121), and then just use that as necessary for Spring Boot Cassandra ITs?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfiguration.java
emerkle826: I agree, this is not ideal. Let me look into it a bit on the driver side. I added all of the metrics here, thinking that if a client application wants driver metrics enabled, there isn't a lot of harm in just turning them all on. And if driver metrics are not desired, the client would simply not enable the actuator.

I'll come up with something that allows the metrics list to be overridden.

snicoll: > And if driver metrics are not desired, the client would simply not enable the actuator.

Yeah but it's more to do with the fact that if the user has expressed an opinion, we must be able to see that and not apply any option. So it's not about overriding really, it's about knowing that something in that area was already configured and therefore the auto-configuration should not apply.

snicoll:
@emerkle826 @adutra how is it going? Sorry for the ping again but if we want this feature to land in 2.4 we really need to find a solution for the problem described in the review. Can I help?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/cassandra/CassandraMetricsAutoConfiguration.java
adutra: > Is there a way to know that a key has been registered already (and therefore not apply anything) or a way for users to override whatever value has been set?

Unfortunately the driver rejects duplicate configuration keys when building the final config. This was done mostly to mimic the behavior of preexisting Guava-backed maps. 

We could of course change or reimplement `ProgrammaticDriverConfigLoaderBuilder` to allow keys to be overridden. Just for completeness, I will open a Java driver ticket to explore that idea. But I'm wondering if it would do more harm than good: allowing keys to be overridden can lead to non-deterministic results. Can Spring Boot guarantee a 100% deterministic configuration order so that it would be safe to override a config key? 

But for now: it seems that we should stick with Erik's suggestion: if you want Cassandra metrics in, add the `java-driver-metrics-micrometer` dependency and let SB configure it for you, otherwise, simply exclude the `java-driver-metrics-micrometer` dependency.

And for the future: we plan to propose to change the driver configuration logic in SB to the more recent [`OptionsMap` API](https://docs.datastax.com/en/drivers/java/4.9/com/datastax/oss/driver/api/core/config/OptionsMap.html). It's 100% in-memory and type-safe. Note that this API accepts duplicate keys. We don't have an ETA for that though because this API has a few shortcomings that will be addressed in a future 4.10 driver release. (Also, I don't know if this would create a binary compatibility problem with SB 2.4 – but I digress).

adutra: Opened [JAVA-2893](https://datastax-oss.atlassian.net/browse/JAVA-2893) FYI.

snicoll: > We could of course change or reimplement ProgrammaticDriverConfigLoaderBuilder to allow keys to be overridden.

That wasn't my main request by the way. My main request was to be able to determine whether a key was already set or not.

> But for now: it seems that we should stick with Erik's suggestion

Sorry but I am not keen to do that. The auto-configuration must not be an on/off switch. We should let users overrides some decisions one way or the other. We can explore other arrangement where this is possible but given that it is configured  on the driver right now, we don't have a lot of options.

I think also that's a general problem with the current approach. How do you compose driver configuration then? If the only thing you could do is set a property with no way to know whether or not it has been registered, this is quite limitative IMO. Have I missed something?

adutra: > How do you compose driver configuration then?

With [`DriverConfigLoader.compose`](https://docs.datastax.com/en/drivers/java/4.9/com/datastax/oss/driver/api/core/config/DriverConfigLoader.html#compose-com.datastax.oss.driver.api.core.config.DriverConfigLoader-com.datastax.oss.driver.api.core.config.DriverConfigLoader-).

> this is quite limitative IMO.

Yes, the current programmatic config API wasn't our masterpiece, I admit. As I said, the plan is to evangelize around the new one (`OptionsMap`).

The concrete consequence of this imo is that this PR cannot be merged for 2.4. I suggest that we table it for now and revisit it later, after the 4.10 release of the driver.

snicoll:
@adutra thanks for the feedback. I've moved this issue out of the `2.4.x` accordingly. I think what we really miss is a way for us to know that the driver was already configured with regards to metrics so that we can back-off (the overriding part is not necessary as we can document that users have to order their customisers before ours if they want to take control over the configuration).

snicoll:
@emerkle826 @adutra any feedback for this issue? An alternative would be to close this issue and another proposal can be submitted once the underlying infrastructure is available.

adutra:
Hi, sorry for taking so long to chime in. 

First of all, driver 4.10 was released last week, and would really recommend that Spring Boot upgrades to that version.

With 4.10 out, I would recommend switching from `ProgrammaticDriverConfigLoaderBuilder` to the new `OptionsMap` API. But that would require a separate PR since the scope is wider.

Once this is done, I *think* we could finish the work here (I say "think" because I didn't quite understand the concrete implications of the reservations expressed [here](https://github.com/spring-projects/spring-boot/pull/23008#discussion_r498886389)).

All in all, I would say that it's probably wiser to close this PR and we'll get back to it when possible.

snicoll:
Alright, let's close it now. Let me know when you want to give this one another go.

As for the reservation, if we auto-configure something, the user should have a way to be able to override this one way or the other. As long as this is possible, there is no problem but what I've experienced so far didn't let me do this (there are some details in this very issue).

PR #23028, opened 2020-08-20
Provide examples of Actuator custom endpoints in the reference documentation

bedla:
Hi,
while creating custom actuator with write operation I found that it would be handy to show in documentation some code example to get idea at first glance.
What do you think?
Thx
Ivos


snicoll:
@bedla fair point although the same can be said for the most-popular `@ReadOperation`. We also try to not write code directly in the doc but move that to actual code we import. Thanks for reporting it and submitting a PR, I'll polish things up.

bedla:
cool, thx. I will do better next time :)

snicoll:
@bedla thank you for making your first contribution to Spring Boot.

PR #23041, opened 2020-08-21
Improve CassandraHealthIndicator with more robust mechanism

tomekl007:

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
@@ -16,10 +16,12 @@
 
 package org.springframework.boot.actuate.cassandra;
 
-import com.datastax.oss.driver.api.core.ConsistencyLevel;
+import java.util.Collection;
+import java.util.Objects;
+
 import com.datastax.oss.driver.api.core.CqlSession;
-import com.datastax.oss.driver.api.core.cql.Row;
-import com.datastax.oss.driver.api.core.cql.SimpleStatement;
+import com.datastax.oss.driver.api.core.metadata.Node;
+import com.datastax.oss.driver.api.core.metadata.NodeState;
 
 import org.springframework.boot.actuate.health.AbstractHealthIndicator;
 import org.springframework.boot.actuate.health.Health;
@@ -31,12 +33,10 @@ import org.springframework.util.Assert;
  * Cassandra data stores.
  *
  * @author Alexandre Dutra
+ * @author Tomasz Lelek
  * @since 2.4.0
  */
 public class CassandraDriverHealthIndicator extends AbstractHealthIndicator {
-
-	private static final SimpleStatement SELECT = SimpleStatement
-			.newInstance("SELECT release_version FROM system.local").setConsistencyLevel(ConsistencyLevel.LOCAL_ONE);
 
 	private final CqlSession session;
 
@@ -52,11 +52,18 @@ public class CassandraDriverHealthIndicator extends AbstractHealthIndicator {
 
 	@Override
 	protected void doHealthCheck(Health.Builder builder) throws Exception {
-		Row row = this.session.execute(SELECT).one();
-		builder.up();
-		if (row != null && !row.isNull(0)) {
-			builder.withDetail("version", row.getString(0));
+		Collection<Node> nodes = this.session.getMetadata().getNodes().values();
+		boolean atLeastOneUp = nodes.stream().map(Node::getState).anyMatch((state) -> state == NodeState.UP);
+		if (atLeastOneUp) {
+			builder.up();
 		}
+		else {
+			builder.down();
+		}
+
+		// fill details with version of the first node (if the version is not null)
+		nodes.stream().map(Node::getCassandraVersion).filter(Objects::nonNull).findFirst()
+				.ifPresent((version) -> builder.withDetail("version", version));
 	}
 
 }
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
@@ -26,6 +26,7 @@ import com.datastax.oss.driver.api.core.metadata.NodeState;
 import org.springframework.boot.actuate.health.AbstractHealthIndicator;
 import org.springframework.boot.actuate.health.Health;
 import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.boot.actuate.health.Status;
 import org.springframework.util.Assert;
 
 /**
@@ -54,12 +55,7 @@ public class CassandraDriverHealthIndicator extends AbstractHealthIndicator {
 	protected void doHealthCheck(Health.Builder builder) throws Exception {
 		Collection<Node> nodes = this.session.getMetadata().getNodes().values();
 		boolean atLeastOneUp = nodes.stream().map(Node::getState).anyMatch((state) -> state == NodeState.UP);
-		if (atLeastOneUp) {
-			builder.up();
-		}
-		else {
-			builder.down();
-		}
+		builder.status(atLeastOneUp ? Status.UP : Status.DOWN);
 
 		// fill details with version of the first node (if the version is not null)
 		nodes.stream().map(Node::getCassandraVersion).filter(Objects::nonNull).findFirst()
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
@@ -17,7 +17,7 @@
 package org.springframework.boot.actuate.cassandra;
 
 import java.util.Collection;
-import java.util.Objects;
+import java.util.Optional;
 
 import com.datastax.oss.driver.api.core.CqlSession;
 import com.datastax.oss.driver.api.core.metadata.Node;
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicator.java
```diff
@@ -50,8 +55,30 @@ public CassandraReactiveHealthIndicator(ReactiveCassandraOperations reactiveCass
...
+			builder.up();
+		}
+		else {
+			builder.down();
+		}
+
+		// fill details with version of the first node (if the version is not null)
+		nodes.stream().map(Node::getCassandraVersion).filter(Objects::nonNull).findFirst()
+				.ifPresent((version) -> builder.withDetail("version", version));
+		return builder.build();
```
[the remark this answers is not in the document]
tomekl007: Good suggestion, I was thinking about the same. I didn't extract it yet because it was duplicated for the current health check logic as well. So if the spring-boot team is ok with extracting this to a separate component, I am +1 on this.

tomekl007:
When should I expect a review for this PR?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
```diff
@@ -52,11 +52,18 @@ public CassandraDriverHealthIndicator(CqlSession session) {
...
+		if (atLeastOneUp) {
+			builder.up();
 		}
+		else {
+			builder.down();
+		}
+
+		// fill details with version of the first node (if the version is not null)
+		nodes.stream().map(Node::getCassandraVersion).filter(Objects::nonNull).findFirst()
+				.ifPresent((version) -> builder.withDetail("version", version));
```
snicoll, line 66: What are the reasons for which a node wouldn't report a version? I think the first map operation should rather return the first node that's up and we should then use the information provided by that node.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
```diff
@@ -52,11 +52,18 @@ public CassandraDriverHealthIndicator(CqlSession session) {
 
 	@Override
 	protected void doHealthCheck(Health.Builder builder) throws Exception {
-		Row row = this.session.execute(SELECT).one();
-		builder.up();
-		if (row != null && !row.isNull(0)) {
-			builder.withDetail("version", row.getString(0));
+		Collection<Node> nodes = this.session.getMetadata().getNodes().values();
+		boolean atLeastOneUp = nodes.stream().map(Node::getState).anyMatch((state) -> state == NodeState.UP);
+		if (atLeastOneUp) {
```
snicoll, line 57: This if/else branch should be simplified with a ternary operator using `builder.status`.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicator.java
```diff
@@ -31,13 +36,11 @@
  *
  * @author Julien Dubois
  * @author Alexandre Dutra
+ * @author Tomasz Lelek
```
snicoll, line 39: We've decided to deprecate this class. Please revert those changes.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicator.java
```diff
@@ -15,27 +15,32 @@
...
 import org.springframework.boot.actuate.health.ReactiveHealthIndicator;
 import org.springframework.data.cassandra.core.ReactiveCassandraOperations;
+import org.springframework.data.cassandra.core.cql.ReactiveSessionCallback;
 import org.springframework.util.Assert;
 
 /**
  * A {@link ReactiveHealthIndicator} for Cassandra.
  *
  * @author Artsiom Yudovin
+ * @author Tomasz Lelek
```
snicoll, line 39: We've decided to deprecate this class. Please revert those changes.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicatorTests.java
```diff
@@ -45,30 +55,86 @@ void createWhenCqlSessionIsNullShouldThrowException() {
 		assertThatIllegalArgumentException().isThrownBy(() -> new CassandraDriverHealthIndicator(null));
 	}
 
+	@ParameterizedTest
+	@MethodSource
+	void reportCassandraHealthCheck(Map<UUID, Node> nodes, Status expectedStatus) {
```
snicoll, line 60: I am not a big fan of `@ParameterizedTest` for this and this looks inconsistent with all the other test. We have rather dedicated test where the expectations are described, potentially sharing a private method that setup the infrastructure. 

Can you please give that a try? If you need inspiration, the other health indicator tests should help.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/cassandra/CassandraDriverReactiveHealthIndicatorTests.java
```diff
@@ -50,29 +56,79 @@ void createWhenCqlSessionIsNullShouldThrowException() {
 		assertThatIllegalArgumentException().isThrownBy(() -> new CassandraDriverReactiveHealthIndicator(null));
 	}
 
+	@ParameterizedTest
+	@MethodSource
+	void reportCassandraHealthCheck(Map<UUID, Node> nodes, Status expectedStatus) {
```
snicoll, line 61: Ditto

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/cassandra/CassandraHealthIndicatorTests.java
```diff
@@ -16,45 +16,119 @@
...
-import static org.mockito.Mockito.mock;
+import static org.mockito.BDDMockito.mock;
+import static org.mockito.BDDMockito.when;
 
 /**
  * Tests for {@link CassandraHealthIndicator}.
  *
  * @author Oleksii Bondar
  * @author Stephane Nicoll
+ * @author Tomasz Lelek
```
snicoll, line 52: This should be reverted

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/cassandra/CassandraReactiveHealthIndicatorTests.java
```diff
@@ -25,38 +37,113 @@
...
 import static org.mockito.BDDMockito.given;
-import static org.mockito.Mockito.mock;
+import static org.mockito.BDDMockito.mock;
+import static org.mockito.BDDMockito.when;
 
 /**
  * Tests for {@link CassandraReactiveHealthIndicator}.
  *
  * @author Artsiom Yudovin
+ * @author Tomasz Lelek
```
snicoll, line 53: This should be reverted.

snicoll (changes requested):
Thanks for the PR. I've made some suggestions. I'd also like we keep the Spring Data equivalent untouched as they are deprecated anyway.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
adutra: > What are the reasons for which a node wouldn't report a version?

A healthy Cassandra installation would always report its version. But we've seen some edge cases in the past: firstly, the `system.peers` table is subject to corruption (intentional or accidental); secondly, this may happen when a node is being upgraded. This is the reason why `com.datastax.oss.driver.api.core.metadata.Node#getCassandraVersion` is annotated with `@Nullable`.

snicoll: Thanks for the feedback @adutra. In the case we keep the node that states healthy, wouldn't that be correct to assume its version is set? Can you have a node reporting healthy and not providing a version at the same time?

adutra: > Can you have a node reporting UP and not providing a version at the same time?

It is technically possible, although very unlikely.

snicoll:
@tomekl007 do you have time to process feedback on this PR?

tomekl007:
> @tomekl007 do you have time to process feedback on this PR?

Hello, unfortunately currently I don't have time to adapt it. Someone from the team should be able to take a look at it in the next week (that starts 28.09)

adutra:
@snicoll sorry for the delay, our team had to shift to a few urgent projects as Cassandra 4.0 GA approaches.

I addressed all your comments and requests, let me know what you think.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/cassandra/CassandraDriverHealthIndicator.java
```diff
@@ -52,11 +53,10 @@ public CassandraDriverHealthIndicator(CqlSession session) {
 
 	@Override
 	protected void doHealthCheck(Health.Builder builder) throws Exception {
-		Row row = this.session.execute(SELECT).one();
-		builder.up();
-		if (row != null && !row.isNull(0)) {
-			builder.withDetail("version", row.getString(0));
-		}
+		Collection<Node> nodes = this.session.getMetadata().getNodes().values();
```
adutra, line 56: This logic is still duplicated twice. Do you think it's worth extracting it to a separate, shared component?

snicoll:
@tomekl007 thank you for making your first contribution to Spring Boot. @adutra thank you for following up!

Issue #22986, opened 2020-08-17
Saml2RelyingPartyAutoConfiguration should use RelyingPartyRegistrations


jzheaux:
Like `ClientRegistrations` in Spring Security's OAuth 2.0 support, Spring Security's SAML 2.0 support can now prepare a `RelyingPartyRegistration.Builder` from a given metadata endpoint as of `5.4.0-RC1`.

It would be nice for an application to be able to do the following:

```yaml
spring:
  security:
    saml2:
      relyingparty:
        registration:
          example:
            identityprovider:
              metadata-uri: https://simplesaml-for-spring-saml.cfapps.io/saml2/idp/metadata.php
```

in order to reduce configuration.

In the absence of a `RelyingPartyRegistrationRepository` in the application context, Spring Boot could do:

```java
RelyingPartyRegistration registration = RelyingPartyRegistrations
        .fromMetadataLocation(metadataUri)
        .registrationId(registrationId)
        .build()
```

Any SP configuration should still be provided to the builder. For example, if the application also specifies signing keys:

```yaml
spring:
  security:
    saml2:
      relyingparty:
        registration:
          example:
            signing.credentials:
              - private-key-location: "classpath:credentials/rp-private.key"
                certificate-location: "classpath:credentials/rp-certificate.crt"
            identityprovider:
              metadata-uri: https://simplesaml-for-spring-saml.cfapps.io/saml2/idp/metadata.php
```

then these should still be configured:

```java
        // ...
        .registrationId(registrationId)
        .signingX509Credentials(c -> c.add(privateKey, certificate))
        .build()
```

Note that Spring Boot's auto configuration currently errors if there are no signing keys specified by the application. From a validation standpoint, though, if `RelyingPartyRegistration#getAssertingPartyDetails#wantAuthnRequestsSigned` returns `false`, there is no need for the application to specify these keys.

wilkinsona:
Closing in favour of #23045.

PR #23045, opened 2020-08-21
Configure SAML 2.0 Service Provider via Metadata

jzheaux:
Closes gh-22986

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
@@ -143,7 +143,7 @@ public class Saml2RelyingPartyProperties {
 		/**
 		 * Endpoint for discovery-based configuration.
 		 */
-		private String metadataUrl;
+		private String metadataUri;
 
 		private final Singlesignon singlesignon = new Singlesignon();
 
@@ -157,12 +157,12 @@ public class Saml2RelyingPartyProperties {
 			this.entityId = entityId;
 		}
 
-		public String getMetadataUrl() {
-			return this.metadataUrl;
+		public String getMetadataUri() {
+			return this.metadataUri;
 		}
 
-		public void setMetadataUrl(String metadataUrl) {
-			this.metadataUrl = metadataUrl;
+		public void setMetadataUri(String metadataUri) {
+			this.metadataUri = metadataUri;
 		}
 
 		@Deprecated
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
@@ -140,6 +140,11 @@ public class Saml2RelyingPartyProperties {
 		 */
 		private String entityId;
 
+		/**
+		 * Endpoint for discovery-based configuration.
+		 */
+		private String metadataUrl;
+
 		private final Singlesignon singlesignon = new Singlesignon();
 
 		private final Verification verification = new Verification();
@@ -150,6 +155,14 @@ public class Saml2RelyingPartyProperties {
 
 		public void setEntityId(String entityId) {
 			this.entityId = entityId;
+		}
+
+		public String getMetadataUrl() {
+			return this.metadataUrl;
+		}
+
+		public void setMetadataUrl(String metadataUrl) {
+			this.metadataUrl = metadataUrl;
 		}
 
 		@Deprecated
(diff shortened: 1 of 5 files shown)


jzheaux:
The build appears to be failing due to something unrelated:

```
Found test failures in 1 test task:

:spring-boot-project:spring-boot-autoconfigure:test
    org.springframework.boot.autoconfigure.session.SessionAutoConfigurationMongoTests > defaultConfigWithUniqueStoreImplementation()
```

and

```
runc run: exit status 1: container_linux.go:346: starting container process caused "process_linux.go:449: container init caused \"rootfs_linux.go:58: mounting \\\"tmpfs\\\" to rootfs \\\"/var/vcap/data/worker/work/volumes/live/80eb908b-5ccf-4263-65e6-2f4c827cb6bf/volume/rootfs\\\" at \\\"/dev/shm\\\" caused \\\"mkdir /var/vcap/data/worker/work/volumes/live/80eb908b-5ccf-4263-65e6-2f4c827cb6bf/volume/rootfs/dev/shm: no space left on device\\\"\""
```

Is there something I should do to address the build failures?

snicoll:
> Is there something I should do to address the build failures?

Thanks for asking @jzheaux. We had some CI issue recently, we will take care of that as part of reviewing the PR.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
```diff
@@ -140,6 +140,11 @@ public void setCertificateLocation(Resource certificate) {
 		 */
 		private String entityId;
 
+		/**
+		 * Endpoint for discovery-based configuration.
+		 */
+		private String metadataUrl;
```
snicoll, line 146: Would that make sense to move that to a `metadata` sub-namespace the way we do for `singlesignon` and `verification`. Not sure if this could be extended in the future though.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyRegistrationConfiguration.java
```diff
@@ -28,6 +28,7 @@
 import org.springframework.boot.autoconfigure.security.saml2.Saml2RelyingPartyProperties.Identityprovider.Verification;
 import org.springframework.boot.autoconfigure.security.saml2.Saml2RelyingPartyProperties.Registration;
 import org.springframework.boot.autoconfigure.security.saml2.Saml2RelyingPartyProperties.Registration.Signing;
+import org.springframework.boot.context.properties.PropertyMapper;
```
snicoll, line 31: The use of `PropertyMapper` adds a lot of noise to what this PR actually changes. I'd prefer if we did that separately (if at all).

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyAutoConfigurationTests.java
```diff
@@ -55,6 +58,15 @@
 	private final WebApplicationContextRunner contextRunner = new WebApplicationContextRunner().withConfiguration(
 			AutoConfigurations.of(Saml2RelyingPartyAutoConfiguration.class, SecurityAutoConfiguration.class));
 
+	private MockWebServer server;
```
snicoll, line 61: There is only one test that uses this. An argument could be made that we could make this local to that test only.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyAutoConfigurationTests.java
```diff
@@ -176,6 +201,46 @@ private boolean hasFilter(AssertableWebApplicationContext context, Class<? exten
 		return filters.stream().anyMatch(filter::isInstance);
 	}
 
+	private void setupMockResponse() {
+		String metadataResponse = "<md:EntityDescriptor entityID=\"https://idp.example.com/idp/shibboleth\"\n"
```
snicoll, line 205: Move this XML file to a test resource perhaps?

snicoll (commented):
Thanks for the PR. I've made a few comments, would you have time to review?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyRegistrationConfiguration.java
jzheaux: If we leave [the associated code](https://github.com/spring-projects/spring-boot/pull/23045/commits/006f58d11e480c5739fa3e0ef229a5f88aea3f57#diff-0639742143d9b9f5e21afa9c54628892L73) as-is, then nulls will override what was specified through the metadata endpoint.

Would surrounding each of those lines with if statements be preferrable?

snicoll: Sorry I missed that. PropertyMapper it is then!

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
jzheaux: I don't know of a use case where there would be more to specify or configure regarding the identity provider's metadata. I'd recommend it stay at this level. In most cases, an application would either specify `metadata-url` or the rest of the properties.

That said, it occurs to me that there's metadata-based configuration support for OAuth 2.0 where the property is named `issuer-uri`. It may be better to call this `metadata-uri`.

snicoll:
thanks for the follow-up @jzheaux!

PR #23048, opened 2020-08-22
Provide a configuration property for Kafka listener container's idleBetweenPolls

s50600822:
In my experience people sometimes need to control their consumption rate.
Spring kafka support this: https://github.com/spring-projects/spring-kafka/blob/master/spring-kafka/src/main/java/org/springframework/kafka/listener/KafkaMessageListenerContainer.java

This simply add the property so it can be configure easily.


wilkinsona:
@s50600822 Thanks very much for making your first contribution to Spring Boot. If you are interested, I polished the proposed changes a little bit in e9ab269.

PR #22946, opened 2020-08-13
Configurable SSL Logback extensions can't be supported due to URL checks

rgoers:
Log4j 2 provides enhanced Spring Boot and Spring Cloud Config support with its [log4j-spring-boot](https://github.com/apache/logging-log4j2/tree/release-2.x/log4j-spring-boot) and [log4j-spring-cloud-config-client](https://github.com/apache/logging-log4j2/tree/release-2.x/log4j-spring-cloud-config/log4j-spring-cloud-config-client) modules.  The Spring Cloud Config Client allows Log4j configuration files to be managed in Spring Cloud Config allowing for dynamic updating of the configuration, configurable SSL support, and support for credentials to access Spring Cloud Config server.

Unfortunately, Spring Boot's LoggingApplicationListener makes it impossible to provide the configurable SSL and credentials support since it needlessly tries to access the configured URL ostensibly to determine if the URL is valid. This check really provides no value since the underlying LoggingSystem will throw an exception if the url cannot be accessed, and because it does not include credentials it results in a 401 response that prevents the Spring Boot application from starting if credentials are required to access the Spring Cloud Configuration server.

PR #23053, opened 2020-08-23
Alternate approach to addressing the url validation check

rgoers:
This is an alternate approach to address the issue noted in #22946. Instead of just removing the validation it moves the code to the LoggingSystem where it can be overridden if, in fact, authentication is supported. The benefit of this approach is that the stack traces the users sees will only change by the addition of an addition StackTraceElement for the LoggingSystem.  Thus this change shouldn't be noticeable by most users.

I also added a checkstyle suppression for the .idea directory since the urls within it were causing the build to fail. Spring shouldn't care about IntelliJ's configuration anyway.


philwebb:
Closing in favor of #22946

PR #23071, opened 2020-08-25
Disallow Apache Commons Lang & Lang3 imports

### LINKED ISSUE #23062: Checkstyle illegal package imports rule for Apache commons does not work
We have a rule to prevent imports on several Apache Commons project using the following regex `^org\.apache\.commons\.(?!compress|dbcp2|lang|lang3|logging|pool2).*`. 

We learned the hard way this rule does not work (see #23059). We also have legit imports on `dbcp2` that are legit.

dreis2211:
Hi,

this PR fixes #23062 as discussed.

Cheers,
Christoph


snicoll:
Thanks again, @dreis2211

PR #23080, opened 2020-08-25
Exclude hibernate-jpa-2.1-api from data-jpa

leonard84:
This dependency duplicates the classes from `jakarta.persistence:jakarta.persistence-api`


wilkinsona:
Thanks for the pull request, but I don't believe this change is necessary. Hibernate [switched away from its own API dependency in the 5.3 betas](https://github.com/hibernate/hibernate-orm/commit/dedc9ef8eb8832cf7820c0c826ac85e90c1df5ed). If you're downgrading Hibernate to 5.2 then you may have to declare some exclusions of your own, but as far as I can tell our starter doesn't need any.

PR #23091, opened 2020-08-26
Support addressShuffleMode to Spring Amqp Autoconfiguration

01045972746:
This support connecting to RabbitMQ clustered host addresses with shuffling.
https://docs.spring.io/spring-amqp/reference/html/#cluster

Because of `shuffle-addresses` mode is deprecated, i have added `addressShuffleMode` property which is `String`.
Please review and comment. Thanks.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
@@ -23,6 +23,7 @@ import java.util.List;
 import java.util.Optional;
 
 import org.springframework.amqp.core.AcknowledgeMode;
+import org.springframework.amqp.rabbit.connection.AbstractConnectionFactory;
 import org.springframework.amqp.rabbit.connection.CachingConnectionFactory.CacheMode;
 import org.springframework.amqp.rabbit.connection.CachingConnectionFactory.ConfirmType;
 import org.springframework.boot.context.properties.ConfigurationProperties;
@@ -41,6 +42,7 @@ import org.springframework.util.StringUtils;
  * @author Gary Russell
  * @author Artsiom Yudovin
  * @author Franjo Zilic
+ * @author Jonghan Kim
  * @since 1.0.0
  */
 @ConfigurationProperties(prefix = "spring.rabbitmq")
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -87,6 +89,11 @@
 	 */
 	private String addresses;
 
+	/**
+	 * Shuffling mode for connecting host. Default mode is NONE.
+	 */
+	private String addressShuffleMode = "NONE";
```
snicoll, line 95: The enum should be exposed rather as it provides better metadata

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -87,6 +89,11 @@
 	 */
 	private String addresses;
 
+	/**
+	 * Shuffling mode for connecting host. Default mode is NONE.
```
snicoll, line 93: The default value should not be documented as it's the role of the configuration metadata to do that (look at other keys for inspiration).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -282,6 +289,28 @@ public void setVirtualHost(String virtualHost) {
...
+	}
+
+	/**
+	 * If addresses have been set and address shuffle mode has been set it is returned.
+	 * Otherwise returns the result of calling {@code getAddressShuffleMode()}.
+	 * @return the address shuffle mode
+	 * @see #setAddressShuffleMode(String)
+	 * @see #getAddressShuffleMode()
+	 */
+	public AbstractConnectionFactory.AddressShuffleMode determineAddressShuffleMode() {
```
snicoll, line 303: I don't think we should do that. Rather we should set whatever has been set by the user.

snicoll (changes requested):
Thanks for the suggestion and the PR. I've made a few comments, can you please have a look to those and also add the necessary tests to exercise the code that you've changed?

01045972746:
@snicoll , i have edited what you mention. please re-review plz.

snicoll:
@01045972746 thank you for making your first contribution to Spring Boot. I've polished your contribution and added a test.

Issue #22151, opened 2020-06-29
Making optional the creation of health indicators for routing datasources


juliojgd:
https://github.com/spring-projects/spring-boot/blob/8436fa9159abcc0355f40fe42370caf4cb47f3d1/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java#L108

Hi, does it make sense to add a `HealthIndicator` with a fixed outcome? (always `UNKNOWN`).
 I mean, would'n it be better to allow the user to choose to not create a HealthIndicator for this type of Datasources (`AbstractRoutingDataSource`)? 
I feel the important information is if any of the routed datasources is Ok or not. If you configured the datasources as beans there will be HealthIndicator's for them, making useless a fixed status value from the routing one.

I am thinking of a new property allowing to exclude the creation of HeatlhIndicator's for `AbstractRoutingDataSource` ones.

Do you think it would be useful? Thanks in advance for reading.

wilkinsona:
Another option may be to reflectively examine the routing data source's `resolvedDataSources` or for Framework to provide an accessor that we can use.

juliojgd:
I agree, it's an option to examine the `resolvedDatasources` but in that case is important to avoid to check the resolved datasources twice in each check cycle (one on its own and another one through the routing one). If this is controlled, might be a good alternative.

wilkinsona:
We discussed this one today and decided to add a property to allow routing data sources to be ignored.

juliojgd:
I can contribute a PR for that

wilkinsona:
Yes please, @juliojgd. That'd be great. I think the property should start with `management.health.db`, something like `management.health.db.ignore-routing-datasources`. Please let us know if you have any questions while you're working on it and we'll do our best to help.

juliojgd:
I have a doubt.
There are several cases to cover divided by the number of routingDatasources in the context:
1.- No RoutingDatasource - OK
2.-Let "M" the number of non routing datsources and "N" the number of routing datasources. OK, I filter the datasources to avoid creating HealthIndicator's for the "N" routing ones.
3.- Only one RoutingDatasource in the context and no other datasource at all. (this case is depicted in the tests)

I think cases (1) and (2) are covered if I do something like this:
```java
        @Bean
[...]

PR #23104, opened 2020-08-26
Port this enhancement to 2.2.x

juliojgd:
Related to #22151 issue. 
Could it be of the general interest to backport this? We have project that cannot make the upgrade to 2.4.x as they are in production and use intesively the RoutigDatasource. These ones are based in 2.2.9.


philwebb:
We discussed this today on our team call and decided that this really does feel like an enhancement and something that we'd prefer not to backport. Since it's possible to work around the issue by defining your own DB health indicator in earlier versions, we don't want to risk any instability in our maintenance releases.

PR #23111, opened 2020-08-27
Simplify code

EruDev:
I found some code to simplify when using IDEA to inspect the code, so I submit a PR here.


snicoll:
@EruDev thank you for making your first contribution to Spring Boot. Automatic inspection from your IDE isn't always a clear cut for us. For instance, even if it can be reduced, we often prefer the readability of not collapsing the last if branch. It's a judgement call and I've reverted one of them.

Issue #23037, opened 2020-08-21
Improve sanitization for list of URI types


crmky:
**Problem**

I'm using Spring Boot 2.2.9. The change introduced by https://github.com/spring-projects/spring-boot/pull/19999 considers any keys contains "uri", "uris", "address" or "addresses" are "comma separated URLs". This is not always the right assumption. It will try to remove password from those URLs, however if it's not a URL format, it will **return the original content**. 

**Expected Behavior**

If that key is not a URL format, it should return as `******`. Or at least allow developers to configure whether they want sanitize URLs or not.

**Reproducer**

```java
Sanitizer sanitizer = new Sanitizer();
System.out.println(sanitizer.sanitize("uris", "[amqp://foo:bar@host/]"));
System.out.println(sanitizer.sanitize("uris", "amqp://foo:bar@host/"));
```

The output is:
```
[amqp://foo:bar@host/]
amqp://foo:******@host/
```

bclozel:
Thanks for raising this issue.
We agree that the behavior is not consistent with our documentation and the whole content should be sanitized if the property is not of `URI` type.

helloworldless:
This looks reasonable for my first contribution! Taking a look...

Davio:
In this case, the part inside the square brackets is still a normal URL which would be properly sanitized if parsed without the square brackets. Wouldn't it make more sense to make the regex more flexible so it can detect embedded URLs?

For instance, changing it to `.*[A-Za-z]+:/.+:(.*)@.+` ?

helloworldless:
That solution definitely makes sense as well. I just implemented the fix in this way per @bclozel's suggestion which I hope I've understood correctly:

> sanitize the password section only if the value matches the user info pattern, or completely sanitize the value and return ******

I was able to justify this solution to myself as follows...

There are probably other patterns out there too, like the URI surrounded by curly braces instead of square brackets, or some other format we can't even imagine. If we accept that it's unlikely that we would ever be able to account for every possible pattern, we can make things simple, and just sanitize anything that doesn't match the most common pattern. That will undoubtedly cause us to sanitize values which do not contain passwords, but it also will prevent us from accidentally exposing a password which is contained in an unhandled pattern.

Davio:
I certainly understand the route you have chosen. It's better to err on the side of caution especially when security is concerned.
Do you know the purpose behind the square brackets for AMQP addresses?

We also use AMQP and specify the username and password separately so it is not embedded in the URL.

[...]

PR #23116, opened 2020-08-28
Sanitize entire value if property is not of URI type; Fixes gh-23037

helloworldless:
Handles case where a property has a uri/address key but the value is not matched against the value regex. In this case, we now sanitize the value.


bclozel:
Closing in favor of the fix variant committed in gh-23037. Thanks!

PR #23161, opened 2020-09-01
Add reference to AOProfiling Spring Boot starter

darioseidl:
A Spring Boot starter to auto-configure aspects to log and measure the execution time of Spring service, controller and repository bean method calls.


snicoll:
@darioseidl sweet, thank you for the reference and your work on the starter.

PR #23173, opened 2020-09-03
Polish SpringBootPluginTests

izeye:
This PR removes `@ClassPathExclusions` in `SpringBootPluginTests` as it doesn't seem to be necessary.


wilkinsona:
Thanks, @izeye. Unfortunately, it is necessary. `kotlin-daemon-client` embeds an old version of `net.rubygrapefruit.platform` which is incompatible with the version that Gradle's generated API embeds. It isn't a problem in a TestKit-powered integration test where, I think, Gradle's class loader isolation keeps the two separate but is a problem in tests that use `ProjectBuilder` when `kotlin-daemon-client` appears on the classpath before Gradle's generated API jar.

PR #23184, opened 2020-09-07
Fix deprecation error in AbstractDataSourceInitializer

dreis2211:
Hi,

https://github.com/spring-projects/spring-framework/commit/670b9fd60b3b5ada69b060424d697270eeee01c2 introduced a deprecation that is causing the builds to fail.

Cheers,
Christoph


dreis2211:
Test failure seems unrelated

snicoll:
@dreis2211 thanks for the PR but I am in the process of backporting the fix, which is why I've created those framework upgrade issues. It's going to land soon. 

I've also spotted the deprecation for AssertJ that I missed because I ran the build with Java11 and  we only do that check with Java 8.

PR #23215, opened 2020-09-07
Fix typo

spencergibb:


snicoll:
Thanks @spencergibb!

wilkinsona:
I've switched this to a task as the typo that it corrects hadn't made it into a release.

Issue #17254, opened 2019-06-18
Add "application/wasm" to MimeMappings


ememisya:
https://github.com/spring-projects/spring-boot/blob/cc1807608e6569c1ceb49e558c99f622d6d788d4/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/MimeMappings.java#L206

Please prepend:

```
mappings.add("wasm", "application/wasm");
```

To the indicated line.  Currently I'm handling recognition of the wasm payload as such:

```
@Component
public class ServletCustomizer implements EmbeddedServletContainerCustomizer
{

  @Override
  public void customize(final ConfigurableEmbeddedServletContainer container)
  {
    final MimeMappings mappings = new MimeMappings(MimeMappings.DEFAULT);
    mappings.add("wasm", "application/wasm");
    container.setMimeMappings(mappings);
  }
}
```

Thank you for your time!

See (https://webassembly.org/docs/web/) for further details on the type.

> This compilation can be performed in the background and in a streaming manner. If the Response is not CORS-same-origin, does not represent an ok status, or does not match the `application/wasm` MIME type, the returned promise will be rejected with a TypeError;

wilkinsona:
I don't think we should make this addition, for the same reasons as #6022:

> We need to draw the line somewhere as each mapping that we configure has a small cost so we configure the three embedded containers that we support (Tomcat, Jetty, and Undertow) with the union of the mime mappings that each of those containers provides by default. Put another way, none of the supported containers provides a mapping for eot or ttf by default so Spring Boot doesn't either.

Further to this, [from Mark Thomas](https://github.com/spring-projects/spring-boot/issues/4161#issuecomment-149190171), Tomcat uses Apache HTTPD's mappings as the source for its mappings and, at the time of writing, HTTPD does not have a mapping for `application/wasm` either.

ememisya:
WebAssembly isn't going anywhere nor is the standard changing anytime soon.  As referenced the mime-type is required by the standard (see referenced part of documentation).  Soon enough everybody who implements wasm will require it and it will need to be added much like application/json as json wasn't going anywhere either.

philwebb:
I think we should continue to drive the contents of `MimeMappings` from the choices that the upstream servlet containers make. They are in a much better position than us to make these kinds of decisions. We also already have tests in place to ensure that our `MimeMappings` includes all of those supported by Tomcat, Undertow and Jetty.

[...]

PR #23234, opened 2020-09-09
Add mime mapping for mjs files

clemstoquart:
Hello,

`.mjs` is a file extension used by javascript modules. More details here : https://v8.dev/features/modules#mjs

It would be nice if Spring Boot supports them out of the box.

Have a nice day !


bclozel:
These days we're trying to only add media types as static entries only if Spring projects have a specific support for that. In the interest of efficiency, we won't add that one right now.

Thanks!

wilkinsona:
Also, please see https://github.com/spring-projects/spring-boot/issues/17254 for some background on the mappings that we include by default.

clemstoquart:
Ok 🙂

PR #23304, opened 2020-09-14
MainClassConvention: use JavaApplication.getMainClass() instead of deprecated JavaApplication.getMainClassName()

rpalcolea:
Hello folks,

While working on Gradle 6.7-rc-1 upgrades I found this Gradle warning:

```
The JavaApplication.getMainClassName() method has been deprecated. This is scheduled to be removed in Gradle 8.0. Use #getMainClass() instead. See https://docs.gradle.org/6.7-rc-1/dsl/org.gradle.api.plugins.JavaApplication.html#org.gradle.api.plugins.JavaApplication:mainClass for more details.
	at org.gradle.api.plugins.internal.DefaultJavaApplication.getMainClassName(DefaultJavaApplication.java:64)
	at org.gradle.api.plugins.internal.DefaultJavaApplication_Decorated.getMainClassName(Unknown Source)
	at org.springframework.boot.gradle.plugin.MainClassConvention.call(MainClassConvention.java:58)
	at org.gradle.util.GUtil.uncheckedCall(GUtil.java:442)

```

I went ahead an modify the `MainClassConvention` to use the `MainClass` lazy property instead

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/plugin/MainClassConvention.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/plugin/MainClassConvention.java
@@ -55,8 +55,8 @@ final class MainClassConvention implements Callable<Object> {
 			return springBootExtension.getMainClassName();
 		}
 		JavaApplication javaApplication = this.project.getConvention().findByType(JavaApplication.class);
-		if (javaApplication != null && javaApplication.getMainClassName() != null) {
-			return javaApplication.getMainClassName();
+		if (javaApplication != null && javaApplication.getMainClass().isPresent()) {
+			return javaApplication.getMainClass().get();
 		}
 		return resolveMainClass();
 	}


philwebb:
I suspect that we won't be able to do a simple switch to the new method since `getMainClass()` was introduced in Gradle 6.4 and we support earlier versions.

rpalcolea:
Hi @philwebb , thanks for looking at this

That makes complete sense. It is just a deprecation warning, do you think it should be better to tackle this in the future instead and keep it as an issue for now?

My understanding is that newer versions of spring boot require 6.3 now or specific version of 5.x with a patch.

I'm not a big fan of having a check for Gradle version and then use reflection to invoke that method.

Happy to follow any direction you think is correct

wilkinsona:
There's no nice way to deal with these deprecations unfortunately. The approach we've taken in the past is to use reflection to try and find the new method. We call it if it's there, falling back to the deprecated method if it's absent or it the attempt to call it reflectively fails. There's an example of this approach in `ApplicationPluginAction`:

https://github.com/spring-projects/spring-boot/blob/63268f33fe646339b706439a1d77cb09c27d23f8/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/plugin/ApplicationPluginAction.java#L95-L111

@rpalcolea Would you like to update your PR to take a similar approach with the main class name?

wilkinsona:
I wanted to test our plugin against the 6.7 release candidate so I've tackled this in https://github.com/spring-projects/spring-boot/commit/581190d7a01745e38779003dc480a47cfcbfe36e. Thanks very much anyway for the PR, @rpalcolea.

PR #23391, opened 2020-09-17
Upgrade Java 15 pipeline to use GA builds

dreis2211:
Hi,

this PR upgrades the Java 15 pipeline to make use of GA builds.

Cheers,
Christoph


snicoll:
This is strange. I ran that command with the same change as you this morning and didn’t get any result. Have you tried to run it manually?

dreis2211:
I called the API manually, yes. But `https://api.adoptopenjdk.net/v3/info/available_releases` for example listed 15 only a couple of hours ago. So I guess you were a little bit too early with your try.

aahlenst:
We published AdoptOpenJDK 15+36 around 09:30+00:00, so very late in the morning for western Europe.

Issue #22664, opened 2020-07-30
Note the Kt class name suffix when describing how to configure a Kotlin application's main class name


Inego:
This can probably be obvious to Kotlin developers, but values of the `mainClassName` property for Kotlin classes should include the `Kt` postfix, if the corresponding application is a Kotlin class.

For example, a Kotlin `ExampleApplication` like

```kotlin
package com.example

@SpringBootApplication
class ExampleApplication

fun main() {
    runApplication<ExampleApplication>()
}
```

would require specifying its entry point as

```kotlin
springBoot {
    mainClassName = "com.example.ExampleApplicationKt"
}
```

and not `"com.example.ExampleApplication"` (as shown in the Reference).

wilkinsona:
Thanks for the suggestion. The Kotlin examples in the Gradle plugin's documentation are for configuring Gradle using its Kotlin DSL. It's not intended to imply that the application itself has been written in Kotlin. That said, there's no harm in including a tip in the [section on configuring the application's main class](https://docs.spring.io/spring-boot/docs/2.3.2.RELEASE/gradle-plugin/reference/html/#packaging-executable-configuring-main-class).

fernandezseb:
Hey, I'd gladly make a PR to add a note.
I would add something along the lines:

> If the main class is written in Kotlin, the name of the generated Java class should be used,
because the Kotlin function `fun main(args: Array<String>)` is defined at the package-level.
By default this is the name of the Kotlin class with the `Kt` suffix added.
For example: `ExampleApplication` becomes `ExampleApplicationKt`.
(If another name is defined using the `@JvmName` annotation, then that name should be used.)

wilkinsona:
@fernandezseb Yes, please. That'd be great. FWIW, I'd omit "because the Kotlin function fun `main(args: Array<String>)` is defined at the package-level" purely in the interests of brevity.

wilkinsona:
Closing in favour of #23418.

PR #23418, opened 2020-09-20
Note the Kt class name suffix when describing how to configure a Kotlin application's main class name

fernandezseb:
The purpose of this PR is to make it clearer that when Kotlin is used, the name of the main class is not equal to the name of the Kotlin class. This is unrelated with what language the Gradle build script is written in.

In issue gh-22664 this confusion was brought up.


wilkinsona:
Thanks very much, @fernandezseb. I've merged this into 2.2.x and then forwards so it'll be part of the next 2.3.x and 2.4.x releases as well.

fernandezseb:
@wilkinsona Thanks for your support!

PR #23502, opened 2020-09-27
Dependency management for Netty tcNative does not include its netty-tcnative module

chenrujun:
Add netty-tcnative in spring-boot-dependencies.


snicoll:
@chenrujun good catch and thank you for making your first contribution to Spring Boot.

PR #23540, opened 2020-09-30
Provide a configuration property for setting the logging charset 

qxo:
ie:
```
logging.pattern.charset=UTF-8
```

```
logging.pattern.charset=GBK
```


wilkinsona:
Thanks for the pull request. I'm wondering how common it is to want to write log files using an encoding that differs from the JVM's default. What's your usecase for using a different encoding?

qxo:
> Thanks for the pull request. I'm wondering how common it is to want to write log files using an encoding that differs from the JVM's default. What's your usecase for using a different encoding?

Yes, for native language was not english.

Such as using os with GBK(chinese), but for some reason using UTF-8 as app default charset, in order to fix console display issue, we have to  set  right charset for console

philwebb:
Thanks for the pull-request @qxo. The recent changes made for #23609 have unfortunately made it impossible to merge this in its current form. I've opened #23827 to look at adding the feature in a different way.

PR #23554, opened 2020-10-01
Improve performance of DispatcherServletRegistrationCondition

dreis2211:
Hi,

this PR slightly improves the performance of `DispatcherServletRegistrationCondition` inside `DispatcherServletAutoConfiguration` by avoiding a relatively costly call to `beanFactory.getBeanNamesForType` if the beanFactory doesn't contain a default dispatcher servlet - which seems to be the case most of the time.

I didn't do any benchmarks on this one - it won't save much obviously. It's just a tiny improvement.

Cheers,
Christoph


snicoll:
Good catch, thanks again Christoph!

PR #23559, opened 2020-10-01
Batch JobParameters can now be configured via properties

### LINKED ISSUE #23411: Add a property to all for Batch JobParameters to enable configuration via environment variables
Currently the only way for a user to provide Spring Batch Job Parameters is via command line args. This requested enhancement is to create a Boot property that accepts JSON to be deserialized into `JobParameters`. This is needed for the Kubernetes functionality when you attempt to restart a job using the `shell` entry point on your image. In this use case, there is no way to pass command line arguments to the application and some elements of normal parameters are not friendly to being set as environment variable names (`run.id(long)=5` for example). I'd assume that the order of precedence for resolving duplicate `JobParameters` would follow the normal Boot conventions.

cppwfs:
A user can now configure batch JobParameters using the `spring.batch.job-parameters` property.
The expected format is json, for example: 
```
export spring_batch_jobParameters="{\"run.id(long)\":\"5\",\"foo.bat\":\"test\"}"
```

resolves #23411

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
@@ -77,6 +77,10 @@ public class BatchAutoConfiguration {
 		if (StringUtils.hasText(jobNames)) {
 			runner.setJobNames(jobNames);
 		}
+		String jobParameters = properties.getJobParameters();
+		if (StringUtils.hasText(jobParameters)) {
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks Glen, but I'm afraid this issue was reserved for a conference event. I'm going to close this one until the event is over just in case a participant wants to submit something. If nobody does, we'll reopen it.

snicoll:
The event is over and we didn't get a PR for this one so let's reopen.

snicoll:
@cppwfs thanks for the PR but we're not keen to paste raw json this way. I've started to [work on another proposal](https://github.com/snicoll/spring-boot/commit/cc9f02bfbc86b63969f23286d639504a7ae0efc4), please follow-up on the original issue.

PR #23564, opened 2020-10-01
Add configuration option to configure RabbitConnectionFactory's channelRpcTimeout

jkhoward:
Added the property channelRpcTimeout in RabbitProperties.java file to be
configurable in RabbitAutoConfiguration

Thank you for the OSD opportunity to learn! I hope you all made a lot of progress on this project today.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
@@ -119,6 +119,11 @@ public class RabbitProperties {
 	 * Connection timeout. Set it to zero to wait forever.
 	 */
 	private Duration connectionTimeout;
+	
+	/**
+	 * Channel RPC timeout.
+	 */
+	private Duration channelRpcTimeout;
 
 	/**
 	 * Cache configuration.
@@ -323,6 +328,10 @@ public class RabbitProperties {
 	public Duration getConnectionTimeout() {
 		return this.connectionTimeout;
 	}
+	
+	public Duration getChannelRpcTimeout() {
+		return this.channelRpcTimeout;
+	}
 
 	public void setPublisherConfirmType(ConfirmType publisherConfirmType) {
 		this.publisherConfirmType = publisherConfirmType;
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@jkhoward Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=23564)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=23564) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@jkhoward Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=23564)!

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -119,6 +119,11 @@
 	 * Connection timeout. Set it to zero to wait forever.
 	 */
 	private Duration connectionTimeout;
+	
+	/**
+	 * Channel RPC timeout.
+	 */
+	private Duration channelRpcTimeout;
```
pierre-sion, line 126: May be annotate with `@DurationUnit(ChronoUnit.SECONDS)` (see field `requestedHeartbeat`)
The default duration is infinite, which may lead to a freeze of the Spring Boot application startup in some situations
=> may be a sensible default of 1 min would be wise?

snicoll:
@jkhoward nicely done and thank you for making your first contribution to Spring Boot. The default for this property [is 10 minutes](https://github.com/rabbitmq/rabbitmq-java-client/blob/283902d3ed7e42098b60cbbf95154aa93a741dbd/src/main/java/com/rabbitmq/client/ConnectionFactory.java#L146) so I've polished that in c4e1b4f.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
snicoll: @pierre-sion the default is already ms so need to change the unit. In the original PR, no default value was set (`null`) and the mapping only applies if a non null value is provided. The default of 10 minutes is preserved but I've made that more explicit in a polish commit.

PR #23571, opened 2020-10-02
Format string should use %n rather than \n

mnhock:


snicoll:
@mnhock good catch! Thank you for making your first contribution to Spring Boot.

PR #23572, opened 2020-10-02
Use JdbcTransactionManager in favor of DatasourceTransactionManager

mdeinum:
Spring 5.3 introduced the JdbcTransactionManager with exception translation
on commit and rollback. With this commit, this transaction manager is going
to be used in favour of the DataSourceTransactionManager used previously.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceTransactionManagerAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceTransactionManagerAutoConfiguration.java
@@ -30,12 +30,11 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.Ordered;
 import org.springframework.jdbc.core.JdbcTemplate;
-import org.springframework.jdbc.datasource.DataSourceTransactionManager;
+import org.springframework.jdbc.support.JdbcTransactionManager;
 import org.springframework.transaction.TransactionManager;
 
 /**
- * {@link EnableAutoConfiguration Auto-configuration} for
- * {@link DataSourceTransactionManager}.
+ * {@link EnableAutoConfiguration Auto-configuration} for {@link JdbcTransactionManager}.
  *
  * @author Dave Syer
  * @author Stephane Nicoll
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/JdbcTransactionManagerAutoConfiguration.java
```diff
@@ -47,17 +46,17 @@
 @ConditionalOnClass({ JdbcTemplate.class, TransactionManager.class })
 @AutoConfigureOrder(Ordered.LOWEST_PRECEDENCE)
 @EnableConfigurationProperties(DataSourceProperties.class)
-public class DataSourceTransactionManagerAutoConfiguration {
+public class JdbcTransactionManagerAutoConfiguration {
```
snicoll, line 49: That's a breaking change as the name of the class is public API.

snicoll (commented):
Thanks for the PR. I need to get back to the framework doc to better understand this change in the hierarchy. I've added a note for myself in the review, no need to act on it now.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/JdbcTransactionManagerAutoConfiguration.java
mdeinum: I did consider changing only the configuration to return a `JdbcTransactionManager` instead of a `DataSourceTransactionManager`. I can still do that and revert this rename and only keep the implementation change?

mdeinum: Reverted the rename and only kept the config change.

snicoll: Thanks but I explicitly asked you to not act on it 

>  I've added a note for myself in the review, no need to act on it now.

I might have to ask you to undo that based on further analysis.

mdeinum: No problem. I assumed you meant the change from `DataSourceTransactionManager` to `JdbcTransactionManager` and not the change in the name of the configuration. If I need to change it again or add a migration path of sorts please let me know.

snicoll:
Unfortunately, I made a typo while copying the reference of the issue. This was fixed by 9293b2629ecfe5931c41638b2b0db2cfc8aebfe5

PR #23575, opened 2020-10-02
Avoid explicit initialization of Atomics with their default values

dreis2211:
Hi,

this PR changes some constructor calls like `new AtomicInteger(0)` that cause a volatile write which can be saved in cases where the constructor parameter is the default value.

Cheers,
Christoph


dreis2211:
Damn it - forgot to change the copyrights of the affected files. Let me push this real quick.

Issue #23480, opened 2020-09-24
DataSourceBuilder should only alias a property when the expected DataSource is configured


snicoll:
`DataSourceBuilder` has a number of aliases to accommodate for small changes in API. For instance, Hikari has a `jdbc-url` rather than the more common `url` property so we have a mapping for that. This mapping is useless if the `DataSource` we configure isn't using Hikari.

So far we got away with it, but adding support for Oracle UCP means we need to add yet another alias that's actually a valid property for DBCP2 (yet referring to something completely different).

snicoll:
Taking a step back, it's rather `ConfigurationPropertyNameAliases` that attempts to bind to any alias that is configured irrespective to the fact that the "main property" has been bound.

snicoll:
To reproduce quickly the issue, you need to patch `DataSourceBuilder` [as follows](https://github.com/snicoll/spring-boot/commit/84e41813cd7dd61d029e8950862e7f1d4a22f1b1#diff-497ec5b3d88628b2d03237ddffeb1e96R90) and the run `CommonsDbcp2DataSourcePoolMetadataTests`. Both `setDriverClassName` and `setConnectionFactoryClassName` will be invoked because those two properties are present in the environment.

philwebb:
Unfortunately we might not be able to fix this one by changing the alias logic. The `alias` feature is part of `ConfigurationPropertySource` and the `Binder` isn't aware of it. I think we'll need to go back to only adding aliases when we know the `DataSource` type supports them.

fabio-grassi-gbs:
I committed a change in pull request #23403 that doesn't use ConfigurationPropertySource in favour of plain BeanWrapper plus some custom logic local to DataSourceBuilder. It's not super-clean, but does the job.

snicoll:
@fabio-grassi-gbs I saw that, that's an interesting approach. I think we'd prefer something more declarative and I've been working on that Friday.

snicoll:
Unfortunately, we broke support for the `OracleDataSource` where an alias is applied with a type that does not implement `DataSource`. I'll reuse this issue to fix that as we haven't released this feature in a milestone yet.

snicoll:
Taking that back, that's rather a type safety issue as we have a test to cover that use case. Looking at the hierarchy, I'll settle for `OracleDataSource` anyway.

snicoll:
I've refined things in ef2fee2

PR #23584, opened 2020-10-05
Improved type safety in DataSourceBuilder

fabio-grassi-gbs:
Switched from oracle.jdbc.datasource.OracleCommonDataSource to oracle.jdbc.datasource.OracleDataSource to fix invalid cast of oracle.jdbc.datasource.OracleCommonDataSource.class to Class<? extends DataSource>.

Added some readability improvements.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
@@ -17,14 +17,12 @@
 package org.springframework.boot.jdbc;
 
 import java.util.ArrayList;
+import java.util.Collection;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Objects;
-import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Function;
-import java.util.stream.Stream;
 
 import javax.sql.DataSource;
 
@@ -36,8 +34,6 @@ import org.springframework.boot.context.properties.source.ConfigurationPropertyN
 import org.springframework.boot.context.properties.source.ConfigurationPropertySource;
 import org.springframework.boot.context.properties.source.MapConfigurationPropertySource;
 import org.springframework.util.ClassUtils;
-
-import static java.util.Optional.ofNullable;
 
 /**
  * Convenience class for building a {@link DataSource} with common implementations and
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the pull request.

I'm afraid I'm finding it rather difficult to identify the fix for the invalid class cast exception from the readability improvements. Also, personally speaking, and find a lot of the readability changes harder to read than the original code. For example, the one of streams and `Optional` while shorter is harder for me to read. Throughout the codebase we avoid using Optional in this way for this sort of reason.

Could you please update the proposal to focus on the fix for the invalid cast and illustrate the problem with a test that fails without the proposal?

snicoll:
> Added some readability improvements.

I second @wilkinsona here. We very much prefer that an issue report (and code change) focuses on the task at hand. Good catch on the fact that the current interface does not implement `DataSource`, I totally missed that. 

I personnel find the readability improvements harder to read that the current arrangement. If you feel differently, we'll have to discuss that in a separate issue.

snicoll:
@fabio-grassi-gbs I've seen the update but I am not sure that's actually addressing the concern. I can take over if you prefer as this is a regression we should fix before the next milestone.

fabio-grassi-gbs:
Ok @snicoll , please take over if time is short. Let me know if it is not clear what I meant to improve/fix, which by the way is a minor problem, type safety is only at risk if the source code is modified, not during regular use.

snicoll:
@fabio-grassi-gbs alright, [I'll take over](https://github.com/spring-projects/spring-boot/issues/23480#issuecomment-704826251). I think the most important here is to focus on one task at a time. Yes we pass along a type doesn't implement `DataSource` for Oracle so the type safety issue exists The issue with `OracleDataSource` is real but that doesn't mean we should change `DataSourceBuilder` the way you did. Thanks for the PR, in any case.

fabio-grassi-gbs:
Ok, I agree it's a minor issue and mostly cosmetic as method at risk is private. Furthermore I saw [ef2fee2](https://github.com/spring-projects/spring-boot/commit/ef2fee22cbdba5d209c8cd099d1c4c5a74c03ff4) now uses a real DataSource. Thanks.

fabio-grassi-gbs:
Just one final question if you will and for mere curiosity. I'm not sure I understood why you decided to introduce the subclass OracleCommonDataSourceSettings. The only reason I see is to override getType and than potentially preventing the attempt to instantiate an interface. Is there some deeper reason? TIA.

snicoll:
The builder can be used so that a suitable connection pool is detected from the classpath or the user specifies the type that they want to use. If someone sets the type to `OracleDataSource` and wants Spring Boot to configure the url, username and password, an alias is required [as described in the constructor](https://github.com/spring-projects/spring-boot/blob/ef2fee22cbdba5d209c8cd099d1c4c5a74c03ff4/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java#L173). Without this class we won't be able to call the `setUser` method.

That makes me think that this class should probably be renamed to `OracleDataSourceSettings` so thank you for that. Having said that, please consider using Gitter going forward if you have design questions like this.

PR #23591, opened 2020-10-05
Fix Undertow do not preserve path of forward request by default

cz111000:
Access requests forwarded to async results will cause thread race. By default, after the Undertow task thread preserve the path of forwarding requests, InvocableHandlerMethod of async requests will be forward methods, not async methods. I writed an example unit test.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/undertow/UndertowServletWebServerFactory.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/undertow/UndertowServletWebServerFactory.java
@@ -75,6 +75,7 @@ import org.springframework.util.Assert;
  * @author Andy Wilkinson
  * @author Marcos Barbero
  * @author Eddú Meléndez
+ * @author 橙 子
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the proposal.

I don't think we should change Undertow's default for everyone as it will be a breaking change for those who are relying on the current behaviour. I suspect this is why the Undertow team [made the setting configurable (originally via a system property) and left the default unchanged](https://github.com/undertow-io/undertow/commit/e5caf63083e8ba281f7f7524f41a75f128c32bf5).

You can configure it in Spring Boot using an `UndertowDeploymentInfoCustomizer` `@Bean`:

```java
@Bean
UndertowDeploymentInfoCustomizer forwardPathPreservation() {
	return (deploymentInfo) -> deploymentInfo.setPreservePathOnForward(false);
}
```

cz111000:
You're welcome, I just discovered this problem when I changed Tomcat to Undertow. It took me some time to solve it. I hope that Spring-boot can provide a way for developers to not care about the differences in the details of the Servlet container, and make the migration easier.

PR #23602, opened 2020-10-06
Fix typo

snicoll:


snicoll:
In case you're wondering what the heck I am doing, that was part of a live demo [for a webinar](https://www.meetup.com/Tech-Meetups-ING/events/273635212/). 😁

PR #23628, opened 2020-10-09
Update default builder image for image builds with buildpacks

ekcasey:
Now that the Cloud Native Buildpacks lifecycle [supports multiple platform API versions](https://github.com/buildpacks/rfcs/blob/main/text/0049-multi-api-lifecycle-descriptor.md) Spring Boot can safely consume the latest Paketo builder image. This image will contain the latest lifecycle and will support API versions 0.3+. The platform API version can be configured by setting CNB_PLATFORM_API in the lifecycle container; it defaults to 0.3.

Also, the Paketo team is now recommending Dockerhub instead of GCR as the default location from which to pull images.


scottfrederick:
Thanks for the contribution @ekcasey!

PR #23632, opened 2020-10-11
Fix XML tag in image publishing example

dreis2211:
Hi,

I just noticed this broken XML closing tag in the Maven-Plugin docs.

Cheers,
Christoph


snicoll (approved):

snicoll:
Good catch, thanks again Christoph.

PR #23636, opened 2020-10-12
Add KafkaStream's CleanupConfig properties

adrien-ben:
Add properties for Spring Kafka's cleanup configuration.


snicoll:
@adrien-ben thank you for making your first contribution to Spring Boot.

I had started a review with a number of things to change and it turned out I had written the code anyway so I went ahead and polished your contribution. Going forward, please make sure to run the build and add the necessary tests that exercise the new behaviour. As the properties had no setter, there was no way the properties could be set to customise the cleanup behaviour. See b960fa3 for more details.

adrien-ben:
@snicoll thanks for fixing and merging it. I pushed a bit early. I'll be more careful next time.

PR #23706, opened 2020-10-15
String to Pattern and String to Regex converters

sokomishalov:


philwebb:
Thanks very much for the PR. I feel like these converters would be better in the core Spring Framework code so that they can be used by any downstream project. There's already a pull-request open, so perhaps you can join the discussion at https://github.com/spring-projects/spring-framework/pull/24311.

sokomishalov:
Oh, haven't seen it, and that makes sense. Thank you!

Issue #23527, opened 2020-09-29
BootJar.getConfigurations() should not be public API


wilkinsona:
The `getConfigurations()` method was [introduced in this commit](https://github.com/spring-projects/spring-boot/commit/7bc7d86ad4a2aae902f032494f9756012cfe94b6).  It's only used for testing but it's `protected` and is therefore part of the public API. We should deprecate it in favour of `Project.getConfigurations()` so that we can remove it in the future.

Ideally, I'd like to do this in 2.3.x.

somayaj:
Can this be worked?

PR #23728, opened 2020-10-16
fix for "BootJar.getConfigurations() should not be public API" #23527

somayaj:
Changes: 
This a fix for BootJar.getConfigurations() should not be public API #23527
deprecated the method. All tests have passed and the method is no longer in the public API.


philwebb:
Thanks for the pull-request, but it's easier if we handle this one. We'll need to apply it to an older branch and also document the deprecation.

somayaj:
Sounds good. Thanks

PR #23737, opened 2020-10-18
Fix links to Spring Framework documentation

dreis2211:
Hi,

recent changes to Spring-Framework (https://github.com/spring-projects/spring-framework/commit/0db3f2b4de4bdd723b65e718c67635fdce8d13fd) have changed the docs location, that this PR addresses. (It seems there is a redirect planned, but I guess it's better to directly adjust to the new location)

Cheers,
Christoph


snicoll:
Thanks @dreis2211. Yes, a redirection is planned but it's better to avoid the redirect.

PR #23766, opened 2020-10-20
Add configuration properties for RabbitMQ's batch listener config

garyrussell:
New features in spring-amqp 2.3.0


snicoll:
Thanks @garyrussell. I am wondering if the opinion of enabling de batching can be moved to Spring AMQP itself? This https://github.com/spring-projects/spring-boot/pull/23766/files#diff-fab84a3c825b61b634548a074e83e5dad7e51987e0abf3c7b8eb51aa9e653e9fR43 looks a bit strange to me and if that's the right default, I'd rather have that done in the library consistently. What do you think?

garyrussell:
Good point - we currently have an assertion; I will change it to just enforce it instead; then we can remove the code here.

PR #23799, opened 2020-10-21
Remove RabbitMQ AutoConfig Boolean Coercion

garyrussell:
It has been moved to spring-amqp instead.


wilkinsona:
I think this should be a task. It's a refinement of https://github.com/spring-projects/spring-boot/pull/23766 which hasn't been release yet.

PR #23886, opened 2020-10-26
fix maven plugin versions in doc

hengyunabc:


hengyunabc:
The version `1.2.3` doesn't exist.

http://search.maven.org/classic/#search%7Cgav%7C1%7Cg%3A%22org.apache.maven.plugins%22%20AND%20a%3A%22maven-jar-plugin%22

http://search.maven.org/classic/#search%7Cgav%7C1%7Cg%3A%22org.apache.maven.plugins%22%20AND%20a%3A%22maven-failsafe-plugin%22

http://search.maven.org/classic/#search%7Cgav%7C1%7Cg%3A%22org.codehaus.mojo%22%20AND%20a%3A%22build-helper-maven-plugin%22

wilkinsona:
Thanks for the PR. The version's really intended to be a hypothetical placeholder rather than an actual version. The 2.2 docs have versions like this:

```
<version>@maven-jar-plugin.version@</version>
```

I wonder if we'd be better using something like that in 2.3 and later (where the docs are now Asciidoctor), rather than specifying a genuine version which we'd then probably want to try to keep up-to-date. Either way, I think it'd be nice to handle this consistently across all the currently maintained branches.

What would you find most useful, @hengyunabc? I'll also flag this for team attention to see what the rest of the team thinks.

philwebb:
Thanks for the pull-request. We've decided it's better for us to remove the `<version>` elements entirely. We'll do that in #23909

PR #23914, opened 2020-10-27
Fix spring.rabbitmq.listener.type in RabbitAutoConfigurationTests

izeye:
This PR fixes `spring.rabbitmq.listener.type` properties in `RabbitAutoConfigurationTests` to align with their tests.


snicoll:
Thanks @izeye. Rather than harmonizing, I've removed the property altogether as those tests were not relying on it at all, see https://github.com/spring-projects/spring-boot/commit/e6c63b5b4e3bc0a37079574094d1bbd258149220

PR #23926, opened 2020-10-28
Make sure ReactiveOAuth2ClientConfigurations actually creates a ReactiveClientRegistrationRepositoryConfig

dxmann73:
Trying out Spring Cloud Gateway 2.4.0-M4 with TokenRelayGatewayFilterFactory, which will not be autoconfigured correctly due to 

```
GatewayAutoConfiguration.TokenRelayConfiguration:
      Did not match:
         - @ConditionalOnBean (types: org.springframework.security.oauth2.client.registration.ReactiveClientRegistrationRepository; SearchStrategy: all) did not find any beans of type org.springframework.security.oauth2.client.registration.ReactiveClientRegistrationRepository (OnBeanCondition)

```

Traced this back to autoconfiguration, where the default config actually creates a bean of the specialized type `InMemoryReactiveClientRegistrationRepository`, rather than `ReactiveClientRegistrationRepository`.

Workaround for now is to configure the bean manually via 

```
@Component
public class ReactiveClientRegistrationRepositoryConfig {

    @Bean
    **ReactiveClientRegistrationRepository** clientRegistrationRepository(@Autowired OAuth2ClientProperties properties) {
        List<ClientRegistration> registrations = new ArrayList<>(
                OAuth2ClientPropertiesRegistrationAdapter.getClientRegistrations(properties).values());
        return new InMemoryReactiveClientRegistrationRepository(registrations);
    }
}

```

This PR should hopefully fix this, although I have no way of testing this really.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/reactive/ReactiveOAuth2ClientConfigurations.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/reactive/ReactiveOAuth2ClientConfigurations.java
@@ -51,7 +51,7 @@ class ReactiveOAuth2ClientConfigurations {
 	static class ReactiveClientRegistrationRepositoryConfiguration {
 
 		@Bean
-		InMemoryReactiveClientRegistrationRepository clientRegistrationRepository(OAuth2ClientProperties properties) {
+		ReactiveClientRegistrationRepository clientRegistrationRepository(OAuth2ClientProperties properties) {
 			List<ClientRegistration> registrations = new ArrayList<>(
 					OAuth2ClientPropertiesRegistrationAdapter.getClientRegistrations(properties).values());
 			return new InMemoryReactiveClientRegistrationRepository(registrations);


pivotal-issuemaster:
@dxmann73 Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=23926)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=23926) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@dxmann73 Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=23926)!

wilkinsona:
Thanks for the proposal. Unfortunately, I don't think this will fix the problem that you have observed. `InMemoryReactiveClientRegistrationRepository` implements `ReactiveClientRegistrationRepository`. This means that the auto-configured bean as currently defined can be injected anywhere that a `ReactiveClientRegistrationRepository`. Generally speaking, the more type information that's provided in a `@Bean` method's signature and the method declaring that it returns the most-specific type possible is recommended.

A `InMemoryReactiveClientRegistrationRepository ` bean will also be able to satisfy the condition on `GatewayAutoConfiguration.TokenRelayConfiguration`, but only if it has been defined at the time when that condition is evaluated. I suspect it's this timing that's causing the problem you're seeing and that there's an auto-configuration ordering problem here. Looking at the Spring Cloud Gateway code, I can't see anything in the auto-configuration ordering on `GatewayAutoConfiguration` that would ensure that it runs after `ReactiveOAuth2ClientAutoConfiguration`. In short, this looks like a Spring Cloud Gateway bug to me.

dxmann73:
Thanks, I wasn't aware of that, although I think I should have guessed. :-)

Will try to come up with a PR for  Spring Cloud Gateway :-)

PR #23935, opened 2020-10-28
Upgrade log4j to 2.13.3 in MavenBuild integration tests

nohn:
as discussed with security@pivotal.com we both agree that this is not a vulnerability in spring boot and can be handled through a PR


philwebb:
Thanks for the pull-request. This version is only used in our own integration tests, but it's worth upgrading nonetheless.

snicoll:
I've moved this to a task as it's not user facing and shouldn't show up in the "dependency upgrades" section of the changelog.

Issue #23810, opened 2020-10-22
Add driverClassName Property to LiquibaseProperties.java


gurpreetsingh85:
Spring Boot - 2.3.4

`LiquibaseProperties.java` doesn't contain `driverClassName` property.

During `LiquibaseAutoConfiguration` instantiation, `createNewDataSource` method takes only url, username, password properties while relying on default driver classname for a given database type. 

For our use case, we need to use MySQL 5.1 driver which doesn't have `com.mysql.cj.jdbc.Driver` class to load hence the only way to get around this issue is by creating your own `LiquibaseDataSource` bean.

It would be a nice to have feature to avoid creating this bean manually.

Thanks in advance

wilkinsona:
I wondered if we'd have the same problem with Flyway, but that doesn't appear to be the case. When Flyway is considered with a URL, username, and password it uses its own logic to figure out the driver class. With MySQL it'll try `com.mysql.cj.jdbc.Driver` first. If that class fails to load, it will fall back to `com.mysql.jdbc.Driver`. An alternative to providing a property would be for us to do something similar.

evgeniycheban:
@wilkinsona I'd like to work on this.

wilkinsona:
Thanks for the offer, @evgeniycheban. That'd be great. We've decided to go with @gurpreetsingh85's suggestion and add a `driverClassName` property to `LiquibaseProperties` and use it when creating the Liquibase-specific `DataSource`. Please let us know if you have any questions while working on the necessary changes.

wilkinsona:
Closing in favour of #23958.

PR #23958, opened 2020-10-29
Provide a configuration property for the driverClassName of the Liquibase DataSource

evgeniycheban:
Fixes gh-23810


wilkinsona:
Thanks very much for the PR, @evgeniycheban. I've squeezed this into 2.4. While I was checking to see if an update to the documentation was needed, I was reminded that we fallback to the equivalent `spring.datasource.*` property for the Liquibase-specific `url`, `username`, and `password` properties so I've polished the proposal to do the same for `driver-class-name`.

PR #23990, opened 2020-11-02
Auto-configure Mongo metrics

bono007:
The current mechanism to enable Micrometer metrics for Mongo is to add the [MongoMetricsCommandListener](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsCommandListener.java) to the [MongoClientSettings.Builder](https://github.com/mongodb/mongo-java-driver/blob/master/driver-core/src/main/com/mongodb/MongoClientSettings.java#L152). This code proposal simply auto-configures in this command listener when appropriate. 

:information_source: There is also a [MongoMetricsConnectionPoolListener](https://github.com/micrometer-metrics/micrometer/blob/master/micrometer-core/src/main/java/io/micrometer/core/instrument/binder/mongodb/MongoMetricsConnectionPoolListener.java#L18) that would add more Mongo (connection pool) metrics. However, I left it out because it has a dependency on legacy `com.mongodb.MongoClient` as well as some Mongo event classes that were removed in Mongo 4 Java driver. I am planning on submitting a fix to Micrometer to bring it up to speed w/ Mongo 4 driver. Once that is available, I will circle back here and auto-configure it in.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.autoconfigure.metrics;
+
+import com.mongodb.MongoClientSettings;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.mongodb.MongoMetricsCommandListener;
+
+import org.springframework.boot.autoconfigure.AutoConfigureAfter;
+import org.springframework.boot.autoconfigure.AutoConfigureBefore;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.autoconfigure.mongo.MongoAutoConfiguration;
+import org.springframework.boot.autoconfigure.mongo.MongoClientSettingsBuilderCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for Mongo metrics.
+ *
+ * @author Chris Bono
+ * @since 2.4.0
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureBefore(MongoAutoConfiguration.class)
+@AutoConfigureAfter({ MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class })
+@ConditionalOnClass(MongoClientSettings.class)
+@ConditionalOnBean(MeterRegistry.class)
+public class MongoMetricsAutoConfiguration {
+
+	@ConditionalOnClass(MongoMetricsCommandListener.class)
+	@ConditionalOnProperty(name = "management.metrics.mongo.command-listener.enabled", havingValue = "true",
+			matchIfMissing = true)
+	static class MongoMetricsCommandListenerConfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean
+		MongoMetricsCommandListener mongoMetricsCommandListener(MeterRegistry meterRegistry) {
+			return new MongoMetricsCommandListener(meterRegistry);
+		}
+
+		@Bean
+		MongoClientSettingsBuilderCustomizer mongoMetricsCommandListenerClientSettingsBuilderCustomizer(
+				MongoMetricsCommandListener mongoMetricsCommandListener) {
+			return (builder) -> builder.addCommandListener(mongoMetricsCommandListener);
+		}
+
+	}
+
+}
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
@@ -46,7 +46,7 @@ import org.springframework.context.annotation.Configuration;
 public class MongoMetricsAutoConfiguration {
 
 	@ConditionalOnClass(MongoMetricsCommandListener.class)
-	@ConditionalOnProperty(name = "management.metrics.mongo.command-listener.enabled", havingValue = "true",
+	@ConditionalOnProperty(name = "management.metrics.mongo.commandlistener.enabled", havingValue = "true",
 			matchIfMissing = true)
 	static class MongoMetricsCommandListenerConfiguration {
 
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
@@ -52,9 +52,9 @@ import org.springframework.context.annotation.Configuration;
 public class MongoMetricsAutoConfiguration {
 
 	@ConditionalOnClass(MongoMetricsCommandListener.class)
-	@ConditionalOnProperty(name = "management.metrics.mongo.commandlistener.enabled", havingValue = "true",
+	@ConditionalOnProperty(name = "management.metrics.mongo.command.enabled", havingValue = "true",
 			matchIfMissing = true)
-	static class MongoMetricsCommandListenerConfiguration {
+	static class MongoCommandMetricsConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
@@ -78,9 +78,9 @@ public class MongoMetricsAutoConfiguration {
 	}
 
 	@ConditionalOnClass(MongoMetricsConnectionPoolListener.class)
-	@ConditionalOnProperty(name = "management.metrics.mongo.connectionpoollistener.enabled", havingValue = "true",
+	@ConditionalOnProperty(name = "management.metrics.mongo.connectionpool.enabled", havingValue = "true",
 			matchIfMissing = true)
-	static class MongoMetricsConnectionPoolListenerConfiguration {
+	static class MongoConnectionPoolMetricsConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
@@ -18,7 +18,12 @@ package org.springframework.boot.actuate.autoconfigure.metrics;
 
 import com.mongodb.MongoClientSettings;
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.mongodb.DefaultMongoMetricsCommandTagsProvider;
+import io.micrometer.core.instrument.binder.mongodb.DefaultMongoMetricsConnectionPoolTagsProvider;
 import io.micrometer.core.instrument.binder.mongodb.MongoMetricsCommandListener;
+import io.micrometer.core.instrument.binder.mongodb.MongoMetricsCommandTagsProvider;
+import io.micrometer.core.instrument.binder.mongodb.MongoMetricsConnectionPoolListener;
+import io.micrometer.core.instrument.binder.mongodb.MongoMetricsConnectionPoolTagsProvider;
 
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.AutoConfigureBefore;
@@ -36,7 +41,8 @@ import org.springframework.context.annotation.Configuration;
  * {@link EnableAutoConfiguration Auto-configuration} for Mongo metrics.
  *
  * @author Chris Bono
- * @since 2.4.0
+ * @author Jonatan Ivanov
+ * @since 2.5.0
  */
 @Configuration(proxyBeanMethods = false)
 @AutoConfigureBefore(MongoAutoConfiguration.class)
@@ -52,14 +58,49 @@ public class MongoMetricsAutoConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
-		MongoMetricsCommandListener mongoMetricsCommandListener(MeterRegistry meterRegistry) {
-			return new MongoMetricsCommandListener(meterRegistry);
+		MongoMetricsCommandListener mongoMetricsCommandListener(MeterRegistry meterRegistry,
+				MongoMetricsCommandTagsProvider mongoMetricsCommandTagsProvider) {
+			return new MongoMetricsCommandListener(meterRegistry, mongoMetricsCommandTagsProvider);
+		}
+
+		@Bean
+		@ConditionalOnMissingBean
+		MongoMetricsCommandTagsProvider mongoMetricsCommandTagsProvider() {
+			return new DefaultMongoMetricsCommandTagsProvider();
 		}
 
 		@Bean
 		MongoClientSettingsBuilderCustomizer mongoMetricsCommandListenerClientSettingsBuilderCustomizer(
 				MongoMetricsCommandListener mongoMetricsCommandListener) {
-			return (builder) -> builder.addCommandListener(mongoMetricsCommandListener);
+			return (clientSettingsBuilder) -> clientSettingsBuilder.addCommandListener(mongoMetricsCommandListener);
+		}
+
+	}
+
+	@ConditionalOnClass(MongoMetricsConnectionPoolListener.class)
+	@ConditionalOnProperty(name = "management.metrics.mongo.connectionpoollistener.enabled", havingValue = "true",
+			matchIfMissing = true)
+	static class MongoMetricsConnectionPoolListenerConfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean
+		MongoMetricsConnectionPoolListener mongoMetricsConnectionPoolListener(MeterRegistry meterRegistry,
+				MongoMetricsConnectionPoolTagsProvider mongoMetricsConnectionPoolTagsProvider) {
+			return new MongoMetricsConnectionPoolListener(meterRegistry, mongoMetricsConnectionPoolTagsProvider);
+		}
+
+		@Bean
+		@ConditionalOnMissingBean
+		MongoMetricsConnectionPoolTagsProvider mongoMetricsConnectionPoolTagsProvider() {
+			return new DefaultMongoMetricsConnectionPoolTagsProvider();
+		}
+
+		@Bean
+		MongoClientSettingsBuilderCustomizer mongoMetricsConnectionPoolListenerClientSettingsBuilderCustomizer(
+				MongoMetricsConnectionPoolListener mongoMetricsConnectionPoolListener) {
+			return (clientSettingsBuilder) -> clientSettingsBuilder
+					.applyToConnectionPoolSettings((connectionPoolSettingsBuilder) -> connectionPoolSettingsBuilder
+							.addConnectionPoolListener(mongoMetricsConnectionPoolListener));
 		}
 
 	}
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfigurationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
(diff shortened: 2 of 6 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureBefore(MongoAutoConfiguration.class)
+@AutoConfigureAfter({ MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class })
+@ConditionalOnClass(MongoClientSettings.class)
+@ConditionalOnBean(MeterRegistry.class)
+public class MongoMetricsAutoConfiguration {
+
+	@ConditionalOnClass(MongoMetricsCommandListener.class)
+	@ConditionalOnProperty(name = "management.metrics.mongo.command-listener.enabled", havingValue = "true",
```
bono007, line 49: This is a bit "non-standard" wrt to the other metrics auto-configurations. There seems to be no other way to disable the auto-configured metrics besides disabling their auto-configuration class. The intention here is to make that more straightforward. 

That being said, if there is a reason to not do this I don't have a strong opinion on it.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+			matchIfMissing = true)
+	static class MongoMetricsCommandListenerConfiguration {
```
bono007, line 51: I split this out into a nested config in anticipation of adding the metrics connection pool listener I alluded to in the pull description description. They would be guarded by different classes and properties.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfigurationTests.java
```diff
@@ -0,0 +1,95 @@
...
+
+	private void assertThatMetricsListenerNotAdded(final AssertableApplicationContext context) {
+		assertThat(context).doesNotHaveBean(MongoMetricsCommandListener.class);
+		assertThat(getActualMongoClientSettingsUsedToConstructClient(context)).isNotNull()
+				.extracting(MongoClientSettings::getCommandListeners).asList().isEmpty();
+	}
+
+	private MongoClientSettings getActualMongoClientSettingsUsedToConstructClient(
+			final AssertableApplicationContext context) {
+		final MongoClient mongoClient = context.getBean(MongoClient.class);
```
bono007, line 91: Sad but true, we need to get at the actually used [MongoClientSettings](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoAutoConfiguration.java#L54) where the builder customizer is applied. Its not exposed so we reflect on it.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
```diff
@@ -2357,6 +2358,41 @@ For more details refer to {spring-kafka-docs}#micrometer-native[Micrometer Nativ
 
 
 
+[[production-ready-metrics-mongodb]]
+==== MongoDB Metrics
+Auto-configuration will register a `MongoMetricsCommandListener` for the auto-configured MongoClient.
+
+A timer metric with the name `mongodb.driver.commands` is created for each command issued to the underlying MongoDB driver.
```
bono007, line 2365: I would rather have put a line that says "For more information, see the Micrometer docs...." - however there are no MM docs for this fine-grained aspect. Their docs are pretty coarse-grained.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfigurationTests.java
```diff
@@ -0,0 +1,95 @@
...
+import org.springframework.test.util.ReflectionTestUtils;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link MongoMetricsAutoConfiguration}.
+ *
+ * @author Chris Bono
+ */
+class MongoMetricsAutoConfigurationTests {
```
wilkinsona, line 39: Thanks for taking the time to test this so comprehensively.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
```diff
@@ -2100,6 +2100,7 @@ Spring Boot registers the following core metrics when applicable:
 * Kafka consumer and producer metrics
 * Log4j2 metrics: record the number of events logged to Log4j2 at each level
 * Logback metrics: record the number of events logged to Logback at each level
+* MongoDB metrics for all commands issued to the client
```
wilkinsona, line 2103: Should this be from the client rather than to the client?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
wilkinsona: We try to avoid `-` in metrics prefixes so, assuming we keep the property, this should be `commandlistener`

wilkinsona (commented):
Thanks very much for the PR, @bono007. On first review, this looks excellent. I've left a few minor comments for your consideration when you have a moment.

Unfortunately, it's too late to add enhancements in 2.4, but we should hopefully be able to merge this for 2.5.

bono007:
> Thanks very much for the PR, @bono007. On first review, this looks excellent. I've left a few minor comments for your consideration when you have a moment.

Thanks @wilkinsona . I am just now getting back to this and will update w/ suggestions shortly.

> Unfortunately, it's too late to add enhancements in 2.4, but we should hopefully be able to merge this for 2.5.

No worries.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
bono007: Let me clarify a bit...

So the `mongo.command-listener` is not a metrics prefix but instead the "type" of metric. Where intent of the property is "configure the mongo command listener metrics to be enabled". 

This is similar to ["management.metrics.web.server.request.metric-name"](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#production-ready-metrics-spring-mvc) where the "web.server.request" is the type of metric where the intent of the property is "configure the web server request metrics to use a name of ...". 

If the above is ok, and you are just referring to the name of the property not having the `-` then I will rename. I just want to be sure that you are ok w/ the intention of the property before I move forward.

bono007: @wilkinsona I went ahead and renamed `command-listener` to `commandlistener`.

jonatan-ivanov:
@bono007 Would you be up to merging this and #25758 together so that both mongo metrics listeners and tag providers will be auto-configured?

bono007:
Perfect timing @jonatan-ivanov. My plan was to add the connection pool metrics listener auto-config this weekend. I will instead happily merge your proposal into this one in the next 12-36hrs. 

@wilkinsona are you ok w/ me combining #25758 into here?

wilkinsona:
The code in #25758 in its current form needs to be reworked as the additions have been made in `spring-boot-autoconfigure` rather than `spring-boot-actuator-autoconfigure`. I haven't looked at things in detail, but I suspect it may be quicker to redo the tag listener part on top of what's already proposed here rather than trying to merge things together.

bono007:
Yep I saw that as well. Sorry for the lack of clarification @wilkinsona - I did not mean literally "merge" - but rather bring the code into this MR. 

Sounds good. I will do this in the 24-36hrs.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+		@Bean
+		MongoClientSettingsBuilderCustomizer mongoMetricsCommandListenerClientSettingsBuilderCustomizer(
+				MongoMetricsCommandListener mongoMetricsCommandListener) {
+			return (clientSettingsBuilder) -> clientSettingsBuilder.addCommandListener(mongoMetricsCommandListener);
+		}
+
+	}
+
+	@ConditionalOnClass(MongoMetricsConnectionPoolListener.class)
+	@ConditionalOnProperty(name = "management.metrics.mongo.connectionpoollistener.enabled", havingValue = "true",
```
bono007, line 81: @wilkinsona it is a bit hard to **not** split `connectionpoolllistener` with hyphens in this case. I resisted the urge though.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
```diff
@@ -2387,6 +2388,97 @@ For more details refer to {spring-kafka-docs}#micrometer-native[Micrometer Nativ
...
+===== Connection Pool Metrics
+Auto-configuration will register a `MongoMetricsConnectionPoolListener` for the auto-configured MongoClient.
+
+The following gauge metrics are created for the connection pool:
+
+* `mongodb.driver.pool.size` that reports the current size of the connection pool, including idle and and in-use members
+* `mongodb.driver.pool.checkedout` that reports the count of connections that are currently in use
+* `mongodb.driver.pool.waitqueuesize` that reports the current size of the wait queue for a connection from the pool
+
+NOTE: The `waitqueuesize` metric was https://github.com/mongodb/specifications/blob/master/source/connection-monitoring-and-pooling/connection-monitoring-and-pooling.rst#why-are-waitqueuesize-and-waitqueuemultiple-deprecated[removed] in MongoDB 4.x
```
bono007, line 2445: @jonatan-ivanov I went ahead and added the info about the `waitqueuesize` deprecation here as well.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+			matchIfMissing = true)
+	static class MongoMetricsConnectionPoolListenerConfiguration {
```
bono007, line 83: Basically everything in this nested config class was shift/lift from @jonatan-ivanov MR. Jonatan, I did not carry over the [customizers concept](https://github.com/spring-projects/spring-boot/pull/25758/files#diff-930122d5dad13377b2fa067da3caca20c66490e4d255705117951b29b9266b32R30) as adding all listeners to the client settings is a more general Mongo auto-config feature and not needed specifically for the metric listeners. I could see that we may want to add that later though.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
wilkinsona: The documentation that's linked to here doesn't seem to be directly related to the metric. The documentation is talking about the removal of a configuration setting rather than a metric for observing the current wait queue size. Given that Micrometer's `MongoMetricsConnectionPoolListener` provides the metric without any reliance of Mongo Driver APIs, I don't think there's any need to mention the deprecation here.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
wilkinsona: Yeah, that's a pretty painful property name. I think we could improve the situation by keeping the fact that there's a listener involved as a lower-level detail and not exposing it in the property name. How about one of the following:

1. `management.metrics.mongo.connectionpool`
2. `management.metrics.mongo.connection`
3. `management.metrics.mongo.pool`

Given the naming used in Micrometer, I'm leaning towards 1.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureBefore(MongoAutoConfiguration.class)
+@AutoConfigureAfter({ MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class })
+@ConditionalOnClass(MongoClientSettings.class)
+@ConditionalOnBean(MeterRegistry.class)
+public class MongoMetricsAutoConfiguration {
+
+	@ConditionalOnClass(MongoMetricsCommandListener.class)
+	@ConditionalOnProperty(name = "management.metrics.mongo.commandlistener.enabled", havingValue = "true",
```
wilkinsona, line 55: I don't think we need `listener` in the name of this property. How about `management.metrics.mongo.command.enabled` instead?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
wilkinsona: Perhaps this should be called `MongoCommandMetricsConfiguration`?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
wilkinsona: Perhaps this should be called `MongoConnectionPoolMetricsConfiguration`?

wilkinsona (changes requested):
Thanks again, @bono007. I've left a few comments for you consideration.

I've also opened https://github.com/micrometer-metrics/micrometer/issues/2541 which may have an effect on this changes. This is just FYI at the moment. If the Micrometer team adopt my suggestion, we can make any updates in Boot at that time.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
bono007: **[APPLY_THIS_TO_ALL_NAMING_COMMENTS]** 

@wilkinsona I agree that dropping the "listener" from all the naming things makes it much better and will follow up w/ that sometime today.

bono007:
> Thanks again, @bono007. I've left a few comments for you consideration.
> 
> I've also opened [micrometer-metrics/micrometer#2541](https://github.com/micrometer-metrics/micrometer/issues/2541) which may have an effect on this changes. This is just FYI at the moment. If the Micrometer team adopt my suggestion, we can make any updates in Boot at that time.

Oh man, that was me. I created these and while I did see the other tags provider I failed to snap to the naming not containing "Metrics". Uggh, naming. :) I will submit a proposal for these changes in Micrometer.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
bono007: Fair enough. Ideally those docs would live in Micrometer anyways and at best this would just point to their docs. I will remove.

bono007:
I made the suggested naming changes and yanked the link to the Mongo docs @wilkinsona . I like it much better. Please take a scan when you can. Thanks.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
```
jonatan-ivanov, line 2:
```suggestion
 * Copyright 2020-2021 the original author or authors.
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+ * {@link EnableAutoConfiguration Auto-configuration} for Mongo metrics.
+ *
+ * @author Chris Bono
+ * @author Jonatan Ivanov
+ * @since 2.5.0
+ */
+@Configuration(proxyBeanMethods = false)
+@AutoConfigureBefore(MongoAutoConfiguration.class)
+@AutoConfigureAfter({ MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class })
+@ConditionalOnClass(MongoClientSettings.class)
```
jonatan-ivanov, line 50: Is the `MongoClient` available here, so that we could do `@ConditionalOnClass(MongoClient.class)` instead?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,108 @@
...
+		}
+
+		@Bean
+		@ConditionalOnMissingBean
+		MongoMetricsCommandTagsProvider mongoMetricsCommandTagsProvider() {
+			return new DefaultMongoMetricsCommandTagsProvider();
+		}
+
+		@Bean
+		MongoClientSettingsBuilderCustomizer mongoMetricsCommandListenerClientSettingsBuilderCustomizer(
```
jonatan-ivanov, line 73: What do you think about being able to add every `CommandListener` and `ConnectionPoolListener` beans?
That might need to go to `boot.autoconfigure.mongo` though.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfigurationTests.java
```diff
@@ -0,0 +1,200 @@
+/*
+ * Copyright 2012-2020 the original author or authors.
```
jonatan-ivanov, line 2:
```suggestion
 * Copyright 2020-2021 the original author or authors.
```

jonatan-ivanov (changes requested):

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
bono007: I commented about that [here](https://github.com/spring-projects/spring-boot/pull/23990/commits/a6b31a2cf4ccf1564fa8214f2b0516f8b2152dcf#r602722685) but I think it got lost in the context and resolved before you saw it. 

I opted to just wire these metrics related listeners in the scope of this PR as to avoid any more changes. We can easily add that in when needed.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
bono007: Since `MongoClientSettings` is shared by both the [reactive](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/ReactiveMongoClientFactory.java#L39) and [blocking](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoAutoConfiguration.java#L54) client this allows the listeners to be shared by both flavors.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
jonatan-ivanov: Oh, sorry, I missed it :)
I think it makes sense,

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfiguration.java
wilkinsona: `2012-2021`is correct here.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/MongoMetricsAutoConfigurationTests.java
wilkinsona: `2012-2021` is correct here.

wilkinsona (approved):

jonatan-ivanov:
fyi: the tags providers were renamed, they are available in the snapshot: https://github.com/micrometer-metrics/micrometer/pull/2548/files

jonatan-ivanov (approved):

wilkinsona:
Thanks again for the PR, @bono007. I've merged the proposed changes along with a [small polishing commit](https://github.com/spring-projects/spring-boot/commit/73e1dd8728525af6e81a216b5d923c7d5400ad85).

PR #24011, opened 2020-11-03
Align MustacheViewResolver with other similar resolvers

dsyer:
The `FreeMarkerViewResolver` (and other template-based resolvers) in Spring Framework avoid using reflection if they know how to instantiate the view class. This change aligns with them.


wilkinsona:
See https://github.com/spring-projects/spring-framework/issues/25847 for details of the related Framework change

PR #24147, opened 2020-11-13
WebFluxTags performs unnecessary Pattern compilation when ignoring trailing slashes

stokpop:
Fix minor cpu hotspot where String.replaceAll calls Pattern.compile for each call. Reuse a precompiled Pattern.

Found in cpu profiling session on spring boot version: 2.3.5.RELEASE


snicoll:
@stokpop thank you for making your first contribution to Spring Boot.

PR #24173, opened 2020-11-17
elasticsearch-rest-client-sniffer is missing from the dependency management for Elasticsearch's REST client modules

izeye:
This PR changes to provide dependency management for `elasticsearch-rest-client-sniffer` as it seems to be missing.


snicoll:
Thanks @izeye, I've backported the suggestion to Maven in `2.2.x` and added the exclusion to `commons-logging`.

PR #24204, opened 2020-11-18
Reduce the overhead of char[] creation

mdeinum:
There some locations which could benefit from not using a
toCharArray on a String, but rather use the charAt method from
the String itself. This to prevent an additional copy of the
char[] being created.


mdeinum:
Whilst optimizing the `SimpleHttpCodeStatusMapper` and `SimpleStatusAggregator` I noticed they both share the exact same `getUniformCode` method. Maybe that should be moved to a shared class or to the `Status` class as to avoid duplication?

snicoll:
@mdeinum thanks. As for the duplication, it's fine as it is.

PR #24208, opened 2020-11-19
Redis health indicator requests more information than it needs resulting in unnecessarily large responses from Redis

xJoeWoo:
Health check of Redis only uses `redis_version` field in `server` section, however, `info` command without any section is used so that all sections of information of Redis will be returned. The size of responded result can be up to 3.5KB for each query. If health check interval is set low like 2 seconds, decoding/processing the result may cost unnecessary CPU time.

image

This PR adds section argument `server` of `info` command for reducing sections to be queried.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/redis/RedisHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/redis/RedisHealthIndicator.java
@@ -60,7 +60,7 @@ public class RedisHealthIndicator extends AbstractHealthIndicator {
 			RedisHealth.up(builder, ((RedisClusterConnection) connection).clusterGetClusterInfo());
 		}
 		else {
-			RedisHealth.up(builder, connection.info());
+			RedisHealth.up(builder, connection.info("server"));
 		}
 	}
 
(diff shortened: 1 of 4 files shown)


snicoll:
Thanks for the PR.

Regardless of the outcome of this issue,  a call to `info` should not lead to this so please consider reporting that against the Redisson project.

snicoll:
I find a little odd that we have to hardcode `"server`" in that method call there. Checking with @mp911de, this API was built to give maximum flexibility if Redis decided to change something. Mark is open to an update of the API to take also a `Enum` with the well-known command. Before proceeding, I'd like to see what the rest of the team thinks about hardcoding "server" there.

wilkinsona:
Given that `server` is documented as an input for the `INFO` command, hardcoding it seems fine to me. An enum of String constant would be an improvement, but in their absence I see no reason not to use the existing `info(String)` API with a hard-coded `server` argument.

xJoeWoo:
> Thanks for the PR.
> 
> Regardless of the outcome of this issue, a call to `info` should not lead to this so please consider reporting that against the Redisson project.
> 
> See also https://redis.io/commands/info

Yes, I've made another PR for improving decoding performance in Redisson. Reducing queried result size from source would be a better solution and make benefit to all libraries.

philwebb:
I agree with @wilkinsona. I' wondering where we should target this one. Do we consider it a bug?

snicoll:
I was about to ask that, believe it or not. I think this can be considered as such and targeted to `2.2.x`.

snicoll:
@xJoeWoo thank you for making your first contribution to Spring Boot. It has been applied to `2.2.x`, `2.3.x` and `master`.

PR #24211, opened 2020-11-19
Correct external application properties load order

MikeN123:
It seems the documentation is showing the wrong order for configuration properties.

The correct order judging from `ConfigDataEnvironment` and the remainder of the documentation is:

```
		locations.add(ConfigDataLocation.of("optional:classpath:/"));
		locations.add(ConfigDataLocation.of("optional:classpath:/config/"));
		locations.add(ConfigDataLocation.of("optional:file:./"));
		locations.add(ConfigDataLocation.of("optional:file:./config/*/"));
		locations.add(ConfigDataLocation.of("optional:file:./config/"));
```

But the documentation states the subdirectories of `./config` are loaded last. I have corrected this and added 'in the current directory' to make sure everyone understands which subdirectories are meant.

I have not verified this is actually the correct loading order, but the code seems clear enough to me.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
@@ -591,8 +591,8 @@ Spring Boot will automatically find and load `application.properties` and `appli
 . The classpath root
 . The classpath `/config` package
 . The current directory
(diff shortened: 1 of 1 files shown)


philwebb:
Thanks for the PR @MikeN123. After some discussion we think it would be better to change the code rather than the docs. I've opened #24308 to deal with that.

PR #24215, opened 2020-11-19
Remove unused AutoConfigureMetricsSpringBootApplication

dreis2211:
Hi,

this PR removes a package private class in the test scope that seems to be unused.

Cheers,
Christoph


dreis2211:
Okay, maybe it's actually used given the tests are failing in that area. Let me investigate.

dreis2211:
I should have called it a day. It's indeed used - although not directly. Sorry for the noise on this one.

PR #24230, opened 2020-11-22
ErrorAttributeOptions.excluding throws IllegalArgumentException if includes is empty

wanderleisouza:

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/error/ErrorAttributeOptions.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/error/ErrorAttributeOptions.java
@@ -27,6 +27,7 @@ import java.util.Set;
  *
  * @author Scott Frederick
  * @author Phillip Webb
+ * @author Wanderlei Souza
  * @since 2.3.0
  */
 public final class ErrorAttributeOptions {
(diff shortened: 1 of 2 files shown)


spencergibb:
Did you target the wrong branch?

snicoll:
@wanderleisouza thanks for the PR.

Thanks @spencergibb, wrong branch indeed. I've fixed that and we can decide where to target this when we review/polish the PR.

wanderleisouza:
oh, thank you @snicoll and @spencergibb. I think it should be in 2.3.x, no? Let me know how to proceed (notice concourse-ci/status is broken due to boot:checkstyleTest). Should I submit a new PR?

snicoll:
Ideally you should have built the project locally before submitting the PR. 

As I've already indicated we can merge the PR in the appropriate branches as part of reviewing the PR. 

Please don't open a new PR as it'd create unnecessary noise. Pushing additional changes to your existing branch will update this PR.

snicoll:
@wanderleisouza thank you for making your first contribution to Spring Boot. I've polished it and applied to both `2.3.x` and `master`. 

I've reworked the tests in particular as the assertions were not running properly. It can be tricky with assertJ but the assertion should run "outside" of the main thing to assert. So `assertThat(errorAttributeOptions.getIncludes().isEmpty());` should have been `assertThat(errorAttributeOptions.getIncludes()).isEmpty();`. This assertion was wrong for the include use case so I've fixed that too.

PR #24239, opened 2020-11-24
Persistent and RelationshipProperties-annotated types are not scanned with Spring Data Neo4j

meistermeier:
To align with the other store modules, I added the `Persistent` type to the initial scan list. We already added this in SDN itself as a supported type.
Also the `RelationshipProperties` type got added to this list, because the loading of those entities would fail if we are running in Spring Data's strict mode (e.g. multi store modules) and they would get requested during runtime. In contrast to the startup phase when this method gets called.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java
@@ -36,6 +36,7 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.data.annotation.Persistent;
 import org.springframework.data.neo4j.config.Neo4jDefaultCallbacksRegistrar;
 import org.springframework.data.neo4j.core.DatabaseSelectionProvider;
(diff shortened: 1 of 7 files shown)


snicoll (changes requested):
@meistermeier thanks for the PR. While I can see adding `Persistent` to be generally useful and consistent, I think that adding  `RelationshipProperties` calls for adding a test that showcases why that is necessary.

meistermeier:
Thanks for the quick response @snicoll . 
I will have to check now first other changes that will come into SDN that might encapsulate this behaviour. So the additional tests will be placed in SDN and the auto configuration will just call its methods.
(Also I have noticed that I missed to add the changes for the reactive configuration to this commit)

meistermeier:
Unfortunately how I thought the change could be done better was not successful. So I stick now to the initial solution and added the tests to ensure that only those classes with annotations that mark explicitly entities (`@Node` and `@Persistent`) or are some kind of helper entity (`@RelationshipProperties`) will be part of the initial entity set.

snicoll:
Thanks for the follow-up. 

> Also the RelationshipProperties type got added to this list, because the loading of those entities would fail if we are running in Spring Data's strict mode (e.g. multi store modules) 

I am not sure I got that. Does adding this type fixes a bug? And if so, can we please have a test that showcases that? The tests that were added merely checked that scanning works as expect but that's not really explaining why adding that type is necessary.

meistermeier:
In a scenario that enforces the Spring Data infrastructure to go into _strict mode_, implicit registration of entities are not possible (throws exception on later registration). Besides being not a prominent entity in the broader API sense, technically a `RelationshipProperties` annotated class is used as one internally.
As a result classes with this annotation also needs to be one of the accepted types for the set of initial entities.

PR #24268, opened 2020-11-27
OriginTrackedPropertiesLoader#isNewDocument has wrong visibility

JimStockwell:
OriginTrackedPropertiesLoader#isNewDocument doesn't seem like
an intended part of the class' public interface,
and it is not called from anywhere outside the class.


snicoll:
@JimStockwell good catch, thank you!

JimStockwell:
Thank you!

Issue #5136, opened 2016-02-12
Add an option to the `run` task to copy resources as they change when devtools is present


snicoll:
We want devtools to work as smoothly as possible. Right now, devtools only monitor the classpath which means that you have to update it to trigger a change. This confuses a lot of people and in particular front-end dev that are used to _only_ change their files.

When a project is running in a development environment, its working directory is usually the root of the module holding the code. All IDEs have settings to make sure that it is the case. We could add an option that does the following:
- Watch for `${user.dir}/src/main/resources/templates` and trigger a LiveReload refresh when thing changes there (maybe we need to be smart to avoid refreshing the browser too often
- Add `${user.dir}/src/main/resources/static` to the list of resources Spring MVC uses to serve static assets

Of course, both of these can be reconfigured by the user. We should inspect the environment and react accordingly.

philwebb:
I think this issue is somewhat unique to IntelliJ since Eclipse automatically copies the assets and triggers a reload.

wilkinsona:
Not just, IntelliJ. It also applies if you're trying to work on the front end of an app using `mvn spring-boot:run`. The story with Gradle is a bit better as you can use `./gradlew bootRun` and `./gradlew build -t` to have your changes continuously built and the classpath auto-updated.

rstoyanchev:
Right and aside from not wanting to feel unique in this context :) ... another part of this is the change in behavior that comes with the upgrade 1.2 to 1.3.

snicoll:
> I think this issue is somewhat unique to IntelliJ since Eclipse automatically copies the assets and triggers a reload.

I don't think that's a fair statement. In Eclipse you have to save and you're used to that because you _need_ to do it (so it's part of the workflow). IJ doesn't need to do anything (inspections, errors, etc  are shown to you live). Devtools is distributive in this workflow because you have to update the classpath and that's not something an IJ user does while coding.

snicoll:
> Devtools is distributive in this workflow because you have to update the classpath

Actually, let me take that back. When you debug an app and you want to hot swap things you have to invoke "make project" as well so Devtools is no different in that area. I think the one thing that matters is resources that do not require a restart. It would be nice if those one were handled in such a way that we don't need to update the classpath.

snicoll:
[...]

Issue #3315, opened 2015-06-24
Devtools reload doesn't work with spring-boot-maven-plugin


donovanmuller:
After reading about the new dev tools features I decided to try it out.
I generally run my apps with `mvn spring-boot:run`. So after the app started up (with `mvn spring-boot:run`), changing some controller code and compiling the app, the app reload didn't happen as I expected. Running the app standalone, however, does reload as expected.

The [docs](spring-boot-maven-plugin) do not mention any constraints about how your run you app.
Should the Maven plugin also support devtools or will this not be supported?

Tried with 1.3.0.M1 and 1.3.0-BUILD-SNAPSHOT.

philwebb:
For the M1 release we've primarily been focusing on reload from the IDE, however, I would like to get things working from the build plugins.

How are you recompiling the application? Do you have `mvn spring-boot:run` in one window and you run `mvn compile` in another? Out of interest, what editor are you using?

> Running the app standalone, however, does reload as expected

How are you running the app standalone in this case? Are you launching it from an IDE or using a different `mvn` goal?

It would be helpful if you could provide some more details of how you usually work. I can then try a similar workflow and try to track down why things aren't working.

donovanmuller:
I'm using IntelliJ.
Below are gifcasts of the two workflows:

`mvn spring-boot:run` (not working)

![boot-mvn-devtools](https://cloud.githubusercontent.com/assets/1175891/8324624/ade14690-1a52-11e5-93fa-cbfa915c9e6d.gif)

So basically, make changes to controller and then `Make project`.
This compiles the changed classes to the `target` directory as  `mvn compile` would do.

Standalone (working)

![boot-app-devtools](https://cloud.githubusercontent.com/assets/1175891/8324625/b072e22e-1a52-11e5-8441-adfbfca9670c.gif)

philwebb:
Thanks! That's a neat way to show the problem. I'll check with our resident IntelliJ/Maven expert (hi @snicoll) to see if IntelliJ writes updated classes in the `/target` directory or not.

snicoll:
Hey. Yes it does. Let me try to run my sample app with `spring-boot:run` then.

@donovanmuller out of curiosity, why are you running your app via the plugin if you can do so in the IDE directly?

donovanmuller:
There's no specific reason.
Some members of the team just find it more familiar/convenient than running from within an IDE.

[...]

PR #24271, opened 2020-11-29
Document how to use Devtools without an IDE

CubeOfCheese:
Updates docs in order to avoid confusion for people who aren't using an IDE. The following NOTE paragraph alone made it seem like the build plugins should just work in the same way as devtools via an IDE.

This somewhat helps address the problem brought up in issue #5136

I discovered that this step was necessary only after hours of searching and testing when I noticed it here: https://github.com/spring-projects/spring-boot/issues/3315#issuecomment-114740127

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using-spring-boot.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using-spring-boot.adoc
@@ -592,6 +592,7 @@ As DevTools monitors classpath resources, the only way to trigger a restart is t
 The way in which you cause the classpath to be updated depends on the IDE that you are using.
 In Eclipse, saving a modified file causes the classpath to be updated and triggers a restart.
 In IntelliJ IDEA, building the project (`Build +->+ Build Project`) has the same effect.
+If using a build plugin, running `mvn compile` for maven or `gradle build` for gradle will trigger a restart.
 ****
 
 NOTE: As long as forking is enabled, you can also start your application by using the supported build plugins (Maven and Gradle), since DevTools needs an isolated application classloader to operate properly.


pivotal-issuemaster:
@CubeOfCheese Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24271)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24271) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@CubeOfCheese Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24271)!

snicoll:
> The following NOTE paragraph alone made it seem like the build plugins should just work in the same way as devtools via an IDE.

I am not sure I understand what you mean. The purpose of the note is to warn you that the classpath should be updated for devtools to show the change. A build system is not going to update the classpath for you unless instructed to do so. 

That being said, thank you for the PR, I think the additional note is a good idea and provide additional context for those who aren't using an IDE.

snicoll:
As part of this change, I'd like we explore the opportunity to reference the trick Andy shared previously with Gradle, see https://github.com/spring-projects/spring-boot/issues/5136#issuecomment-183380358

CubeOfCheese:
> > The following NOTE paragraph alone made it seem like the build plugins should just work in the same way as devtools via an IDE.
> 
> I am not sure I understand what you mean. The purpose of the note is to warn you that the classpath should be updated for devtools to show the change. A build system is not going to update the classpath for you unless instructed to do so.
> 
> That being said, thank you for the PR, I think the additional note is a good idea and provide additional context for those who aren't using an IDE.

When I said following Note, I was referring to this paragraph:
```
NOTE: As long as forking is enabled, you can also start your application by using the supported build plugins (Maven and Gradle), since DevTools needs an isolated application classloader to operate properly.
By default, the Gradle and Maven plugins fork the application process.
```
This paragraph implied to me that using a build plugin with devtools should work just as it does with an IDE (recompiling on save) as long as forking is enabled. Maybe that was just my own poor interpretation.

snicoll:
Thanks for the follow-up. I had missed that and I agree this particular note is confusing. I'll review it as part of polishing your contribution.

philwebb:
Thanks very much for the PR @CubeOfCheese

Issue #24299, opened 2020-11-30
Contributing and unit tests


JimStockwell:
The "Code Conventions and Housekeeping" section of CONTRIBUTING.adoc says,
"A few unit tests would help a lot as well -- someone has to do it."

Is there a standardized way of running such unit tests, for example, the ones in the org.springframework.boot.actuate.endpoint.EndpointIdTests class?

I think there would be value in addressing that question right there in CONTRIBUTING.adoc.

If I can wish, I wish for something as convenient as `./gradlew test --tests EndpointIdTests`.

Thank you for your thought and consideration.  Spring Boot is a great help to all the world.

snicoll:
Thanks for the suggestion. We can add something like that although I'd rather have it generalized so that it also runs other checks (such as check style and code format).

I have this on my zsh profile:

```
function check() {
  g check test --tests "$1*"
}
```

Running check and test makes sure the  code is formatted properly and it also runs tests starting with the first argument so `check Tomcat` would run checks in the current module and all tests starting with `Tomcat`.

> If I can wish, I wish for something as convenient as ./gradlew test --tests EndpointIdTests.

I am confused as this command does exactly what you've asked. Have I missed something?

JimStockwell:
Thank you @snicoll ,  that's just what I'm looking for.  But now I discover that although the right way is simple, it doesn't work for me.  (Or I'm missing something else.)

Sorry to turn this into a debugging session.  Can you help me here?  When I try the following:

`./gradlew check test --tests EndpointIdTests`

I get back:
```
Execution failed for task ':spring-boot-project:spring-boot:test'.
> No tests found for given includes: [EndpointIdTests](--tests filter)
```
I'm not familiar with Gradle, so I won't be surprised if that's a big part of my problem.  Is there an obvious thing I'm doing wrong?  EndpointIdTests seems like a valid unit test to me.  It's in `spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/endpoint/EndpointIdTests.java`.

Thanks again.

snicoll:
This issue is now superseded by #24301.

PR #24301, opened 2020-11-30
Document how to run unit tests

JimStockwell:
Explicitly show how to run unit tests and checkstyle.

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -54,7 +54,7 @@ added after the original pull request but before a merge.
   Plugin] or format the code from the Gradle build by running
   `./gradlew format`. Note that if you have format violations in `buildSrc`, you can fix
   them by running `./gradlew -p buildSrc format` from the project root directory.
-* The build includes checkstyle rules for many of our code conventions. Run
+* The build includes Checkstyle rules for many of our code conventions. Run
   `./gradlew checkstyleMain checkstyleTest` if you want to check your changes are
   compliant.
 * Make sure all new `.java` files have a Javadoc class comment with at least an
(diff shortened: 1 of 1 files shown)


CONTRIBUTING.adoc
```diff
@@ -96,7 +96,13 @@ The project can be built from the root directory using the standard Gradle comma
 	$ ./gradlew build
 ----
 
+=== Running Unit Tests
+Unit tests and checkstyle can be run from the relevant *module directory* using the standard Gradle command:
 
+[indent=0]
+----
+        $ ../../gradlew check test
```
wilkinsona, line 104: `check` depends on `test` so running both isn't necessary. `check` alone is sufficient.

I'm also not sure about running from a particular module directory. The style that I see more frequently is to run from the root using a task's full path, for example:

```
./gradlew spring-boot-project:spring-boot:check
```

wilkinsona (changes requested):

CONTRIBUTING.adoc
wilkinsona: Looking at https://github.com/spring-projects/spring-boot/issues/24299 reminded me that running both `check` and `test` is necessary when using `--tests`. IIRC, there's a Gradle issue about this but I can't find it right now. That's quite specialised usage so, IMO, we should be sticking to the single `check` task here.

JimStockwell: That makes a lot of sense to me @wilkinsona -- thank you.  I've incorporated those changes.  In light of the insight you brought, I made some other small changes to this pull request, which you will see.  If there's any concern with them, please let me know.  Thanks again.

CONTRIBUTING.adoc
```diff
@@ -66,6 +66,8 @@ added after the original pull request but before a merge.
   than cosmetic changes).
 * Add some Javadocs.
 * A few unit tests would help a lot as well -- someone has to do it.
+* Verification tasks, including tests and stylecheck, can be executed
```
wilkinsona, line 69: Typo? Should be Checkstyle rather than stylecheck, I think.

JimStockwell: Ouch!  Thanks! :)

philwebb:
Thanks very much for PR. I've merged this to master.

Issue #24237, opened 2020-11-24
Use HttpSessionOAuth2AuthorizedClientRepository instead of AuthenticatedPrincipalOAuth2AuthorizedClientRepository as bean.


chenrujun:
https://github.com/spring-projects/spring-boot/blob/v2.4.0/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/servlet/OAuth2WebSecurityConfiguration.java

IMO, Use `HttpSessionOAuth2AuthorizedClientRepository`  is more reasonable: the `authorizedClient` will cleared automatically after session expired.

philwebb:
The javadoc of `AuthenticatedPrincipalOAuth2AuthorizedClientRepository` suggests that it delegates to `HttpSessionOAuth2AuthorizedClientRepository` for unauthenticated requests. Could you provide a bit more background about your suggested change?

/cc @jgrandja

chenrujun:
Hi, @philwebb , Thank you for your response.

`InMemoryOAuth2AuthorizedClientService` will save all `authorizedClient`s in memory, it may cause OOM in some case.

I created a draft PR: https://github.com/spring-projects/spring-boot/pull/24303/files

philwebb:
Thanks @chenrujun.

One problem with the suggested change is that users will no longer be able to just provide a `OAuth2AuthorizedClientService` bean. They'd need to provide an `AuthenticatedPrincipalOAuth2AuthorizedClientRepository` with the `OAuth2AuthorizedClientService` already set.

I wonder if `InMemoryOAuth2AuthorizedClientService` should be constrained so that it accepts a maximum number of entries. @jgrandja Do you have any thoughts on this?

chenrujun:
@philwebb 

Current logic is like this:

If one user login, saved some `authorizedClient`s in `InMemoryOAuth2AuthorizedClientService`, then left.
The the `authorizedClient` will never be deleted automically.

If the web application have billions of users, it will cause OOM.

---
IMU, with `HttpSessionOAuth2AuthorizedClientRepository`, when session expired, the `authorizedClient`s will be cleaned automically.

chenrujun:
@philwebb 

How about create a new class `HttpSessionOAuth2AuthorizedClientService`, use it to replace `InMemoryOAuth2AuthorizedClientService`?

If you agree, I'm willing to create another PR.

philwebb:
[...]

PR #24303, opened 2020-12-01
Use HttpSessionOAuth2AuthorizedClientRepository instead of AuthenticatedPrincipalOAuth2AuthorizedClientRepository as bean.

chenrujun:


philwebb:
This will unfortunately break back-compatibility. Let's continue the discussion in #24237 to see if there are other options.

Issue #17766, opened 2019-08-02
Support mvn spring-boot:run with classpaths that exceeds Windows' length limits


pneuschwander:
Hi there, we use Spring Boot Version 2.1.5.RELEASE + maven 3.6.0 on a windows 10 machine to develop a spring webmvc application.

When invoking `mvn spring-boot:run` the following exception occurs:
```
Caused by: java.io.IOException: CreateProcess error=206, The filename or extension is too long
    at java.lang.ProcessImpl.create (Native Method)
    at java.lang.ProcessImpl.<init> (ProcessImpl.java:420)
    at java.lang.ProcessImpl.start (ProcessImpl.java:151)
    at java.lang.ProcessBuilder.start (ProcessBuilder.java:1107)
    at java.lang.ProcessBuilder.start (ProcessBuilder.java:1071)
    at org.springframework.boot.loader.tools.RunProcess.run (RunProcess.java:81)
    at org.springframework.boot.maven.RunMojo.forkJvm (RunMojo.java:103)
```
Problem: Our classpath is too long as we have many dependencies.

A feature/fix to shorten the command line in such cases (e.g. use of temp argFile) would be awesome.

philwebb:
Out of interest, have you tried this: https://www.howtogeek.com/266621/how-to-make-windows-10-accept-file-paths-over-260-characters/

regmebaby:
@philwebb Thank you for that hint. I tried it but sadly it does not work in this case. Still getting the same exception :-/

Maybe it is not about the filepath length limit. 
java.lang.ProcessImpl.create javadoc states "Create a process using the win32 function CreateProcess."
Found these information sources:
- https://devblogs.microsoft.com/oldnewthing/20031210-00/?p=41553
- https://docs.microsoft.com/en-us/windows/win32/api/processthreadsapi/nf-processthreadsapi-createprocessa

stating that the command line length is limited to 32767 characters. 
Our classpath exceeds this limit as there are too many entries...

philwebb:
Thanks for trying, we'll have to try and investigate what options we have.

wilkinsona:
I believe `./gradlew bootRun` will also be affected.

[...]

PR #24320, opened 2020-12-03
Issue #17766 - change classpath from command line to environment variable when forking

cquoss:
This should fix issue #17766.  Please take into account that this solution would append to any CLASSPATH env var already provided in configuration.
Another solution would be to simply leave the provided classpath env var as it is in such cases.
Then the caller would have to include all maven related classpath items.
In either case an amendment of the goal documentation should be taken into consideration.
Please guide me on how to do so if you find this useful.

Regards

Clemens Quoß


pivotal-issuemaster:
@cquoss Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24320)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24320) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@cquoss Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24320)!

philwebb:
We discussed this today as a team and we've decided that unfortunately there are too many downsides to used the environment variable. We think the pathing jar is the best long term solution as it will work with both forked and non-forked processes. We'll leave #17766 open to track that.

Thanks anyway for the idea and the PR.

PR #24327, opened 2020-12-03
Allow recursive profile group references

### LINKED ISSUE #24319: Spring Boot profile groups with recursive references end up in endless loop
While migrating conigurations to Spring Boot 2.4, I accidentally ended up with something like:

```yaml
spring.profiles.group.myprofile: myprofile, other
```

The JVM failed to start with an out of memory error, I assume that this ends up in a recursive loop. I suggest that the recursive definition should yield and error or be ignored.

dreis2211:
Hi,

this PR fixes #24319 by removing the current profile from group profiles in order to avoid an endless loop when expanding/flattening the profile list.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/Profiles.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/Profiles.java
@@ -118,7 +118,7 @@ public class Profiles implements Iterable<String> {
 			String current = stack.pop();
 			expandedProfiles.add(current);
 			List<String> groupProfiles = asReversedList(this.groups.get(current));
-			groupProfiles.remove(current);
+			groupProfiles.removeAll(expandedProfiles);
 			groupProfiles.forEach(stack::push);
 		}
 		return asUniqueItemList(StringUtils.toStringArray(expandedProfiles));
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/Profiles.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/Profiles.java
@@ -117,7 +117,9 @@ public class Profiles implements Iterable<String> {
 		while (!stack.isEmpty()) {
 			String current = stack.pop();
 			expandedProfiles.add(current);
-			asReversedList(this.groups.get(current)).forEach(stack::push);
+			List<String> groupProfiles = asReversedList(this.groups.get(current));
+			groupProfiles.remove(current);
+			groupProfiles.forEach(stack::push);
 		}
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/Profiles.java
```diff
@@ -128,7 +130,7 @@ private boolean hasExplicit(Supplier<String[]> supplier, String propertyValue, S
 		}
 		List<String> reversed = new ArrayList<>(list);
 		Collections.reverse(reversed);
-		return Collections.unmodifiableList(reversed);
+		return reversed;
```
dreis2211, line 133: Alternatively, we could do a stream/filter approach, but the unmodifiable list seemed unnecessary as the result is only used above and cannot be modified from the outside.

mbhave:
I'm not sure if just removing the current profile is enough. I think we'd end up with the same issue if we had a configuration like:  

```
spring.profiles.group.a=e,f
spring.profiles.group.e=a,x,y
```

Maybe we should throw an exception in this case since it is an invalid configuration.

dreis2211:
@mbhave Good catch - I covered the additional case now.

Which sort of exception do you want to be thrown here? `IllegalStateException` or a new one? Just let me know if I should add the exception and if so which type. The "maybe" suggests that you're not sure yet.

philwebb:
I think an `IllegalStateException` would be fine. We can always change it later if we need to.

dreis2211:
@philwebb I added the `IllegalStateException`. Let me know if this is what you had in mind.

philwebb:
Thanks once again @dreis2211! I've merged this into master with a minor update to make the message catch duplicates in the expanded set as well as the stack.

philwebb:
Reopening because I just realized that a user might want to have the same expansion in different groups:

```
	@Test
	void test() {
		MockEnvironment environment = new MockEnvironment();
		environment.setProperty("spring.profiles.active", "a,b,c");
		environment.setProperty("spring.profiles.group.a", "e,f");
		environment.setProperty("spring.profiles.group.e", "f,g");
		Binder binder = Binder.get(environment);
		assertThatIllegalStateException().isThrownBy(() -> new Profiles(environment, binder, null))
				.withMessageContaining("Profiles could not be resolved. Remove profile 'a' from group: 'e'");
	}
```

philwebb:
After some more thought, I think it's best if we tolerate duplicates rather than throwing an exception. I've changed things again in commit bef5fe29e3827f538ed95468fe6ea532f0b84867.

@mbhave @dreis2211 I'd appreciate a review if you have time.

dreis2211:
Looks good to me, @philwebb

Issue #18583, opened 2019-10-15
Add some guidance to the reference documentation about diagnosing unexpected property values


mdeinum:
Lately I have been answering question on stackoverflow for people using "reserved" property names. Like `env`, `environment`, `classpath` and the likes. I think it would be beneficial to add a note/section to the reference guide mentioning those properties. This both as to add clarity and as a reference for answering questions. 

See:  https://stackoverflow.com/questions/58387443/spring-boot-reading-different-value-for-one-key-from-application-properties

wilkinsona:
Thanks for the suggestion.

> Lately I have been answering question on stackoverflow for people using "reserved" property names. Like env, environment, classpath and the likes.

I'm not sure that any of those are reserved. As far as I can tell, none of them is set by default as either a system property or an environment variable. We already [document the property source ordering](https://docs.spring.io/spring-boot/docs/2.2.0.RELEASE/reference/html/spring-boot-features.html#boot-features-external-config) which should help people to diagnose why a property set in one source is being overridden. The `env` and `configprops` endpoints can also be useful in that regard. I think the best option here is probably to add a note to the docs about using those two endpoints to diagnose unexpected property values.

mdeinum:
If I remember correctly the `environment` issue arose due to use an SpEL expression, `#{environment}` instead of a value expression, `${environment}`. The SpEL expression resolves to the `Environment` instance configured, the value expression would indeed fail. So the first was due to a faulty use of SpEL to obtain a pre-configured property. 

The `classpath` isn't indeed default set but when set through env or system properties it might be confusing. 

Adding additional information on how to resolve/investigate this will certainly help. Another addition could be a best practice for naming properties, like to put them in a application/component namespace, so that they don't get mixed up with other properties.

wilkinsona:
Closing in favour of #24336.

PR #24336, opened 2020-12-04
Add some guidance to the reference documentation about diagnosing unexpected property values

wesuga:
Fixes gh-18583


philwebb:
Thanks for the PR @wesuga. This is now in 2.3.x, 2.4.x and master.

Issue #22632, opened 2020-07-28
Log failing calls to health indicators


ttddyy:
This is similar to #22509 to improve root cause analysis when probe endpoints returned non 200 response.

I am migrating k8s http probes to use readiness and liveness health group endpoints(`/actuator/health/[readiness|liveness]`).
When these endpoints return non `UP` status(other than 200 response), k8s stops traffic or shutdown the pod. When such event happens, k8s http probe only record the returned http status for the reason of its probe failure.
This makes hard to investigate WHY readiness/liveness probes returned non 200 response  when somebody needs to investigate the failure reason later. Even if k8s could record body of probe response, it would be nicer to have such information in application log.

I wrote this implementation to our services to log information when health endpoints returns non `UP` response.

```java
@Slf4j
public class LoggingHealthEndpointWebExtension extends HealthEndpointWebExtension {

	public LoggingHealthEndpointWebExtension(HealthContributorRegistry registry, HealthEndpointGroups groups) {
		super(registry, groups);
	}

	@Override
	public WebEndpointResponse<HealthComponent> health(ApiVersion apiVersion, SecurityContext securityContext,
			boolean showAll, String... path) {
		WebEndpointResponse<HealthComponent> response = super.health(apiVersion, securityContext, showAll, path);
		HealthComponent health = response.getBody();
		if (health == null) {
			return response;
		}

		Status status = health.getStatus();
		if (status != Status.UP) {
			Map<String, HealthComponent> components = new TreeMap<>();
			if (health instanceof CompositeHealth) {
				Map<String, HealthComponent> details = ((CompositeHealth) health).getComponents();
				if (details != null) {
					components.putAll(details);
				}
			}
			log.warn("Health endpoints {} returned {}. components={}", path, status, components);
		}

		return response;
	}

}
```

If `HealthEndpointSupport` could have logging capability (or `HealthEndpointWebExtension` and `ReactiveHealthEndpointWebExtension` for web only), then we don't need to have this custom implementation.

Something like:
```java
boolean enableLogging;

if(this.enableLogging && health.getStatus() != Status.UP) {
  log.warn(...);
}
```

wilkinsona:
Each indicator that subclasses `AbstractHealthIndicator` should already log a warning when a health check fails:

[...]

PR #24345, opened 2020-12-07
log health status (and description) if status != UP

jackhammer2k:
It is important to not only know **that** a service is unhealthy, but also **why** it is (which health check, description). Unfortunately
some external monitoring systems do not provide the output/body of the HTTP request or do not create a history of that. A simple log message solves the issue completely in these cases.
For this reason I implemented a log statement in case the health check is not `UP`. To not change the default behavior and annoy people I choosed the `DEBUG` level.

I think that can be a useful feature for some people and does not harm anyone else.

btw: Its a bit unclear for me how you manage your branches. If this is merged into master, how will it be integrated best into `2.2.x` and `2.3.x`?

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/HealthEndpointWebExtension.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/HealthEndpointWebExtension.java
@@ -20,6 +20,8 @@ import java.util.Arrays;
 import java.util.Map;
 import java.util.Set;
 
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 import org.springframework.boot.actuate.endpoint.SecurityContext;
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
 import org.springframework.boot.actuate.endpoint.annotation.Selector;
@@ -43,6 +45,8 @@ import org.springframework.boot.actuate.endpoint.web.annotation.EndpointWebExten
  */
 @EndpointWebExtension(endpoint = HealthEndpoint.class)
 public class HealthEndpointWebExtension extends HealthEndpointSupport<HealthContributor, HealthComponent> {
+
+	private static final Logger logger = LoggerFactory.getLogger(HealthEndpointWebExtension.class);
 
 	private static final String[] NO_PATH = {};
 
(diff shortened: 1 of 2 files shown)


philwebb:
See #22632

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/HealthEndpointWebExtension.java
```diff
@@ -20,6 +20,8 @@
 import java.util.Map;
 import java.util.Set;
 
+import org.slf4j.Logger;
```
philwebb, line 23: Could you please update the PR to use commons-logging rather than SLF4J

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/HealthEndpointWebExtension.java
```diff
@@ -76,6 +81,10 @@ public HealthEndpointWebExtension(HealthContributorRegistry registry, HealthEndp
 		}
 		HealthComponent health = result.getHealth();
 		HealthEndpointGroup group = result.getGroup();
+		if (!Status.UP.equals(health.getStatus())) {
+			logger.debug("health status=" + health.getStatus().getCode() + " with description="
```
ravening, line 85: Should this be debug or warn ? Since the app is not un up state, warn sounds more logical to me

jackhammer2k: I agree, that `warn` would be the preferred level, but I didn't want to change the default behavior of Spring Boot for all Spring Boot users. In addition the log level will be `warn` in #22632 I guess. Thus its just additional logging for those who want to have it (by setting the class-level to `debug`). I also hope it willl make it earlier to `master` than #22632.

snicoll: >  I also hope it willl make it earlier to master than #22632.

Sorry but it won't. This issue or the other one is going to be superseded and I already have some work on a branch when I was working on #22632.

mbhave:
@jackhammer2k We have decided to use the approach suggested by @snicoll [here](https://github.com/spring-projects/spring-boot/issues/22632#issuecomment-707261129) by exposing an accessor for the `exception` on the `builder`. Thanks anyway and sorry for the wasted effort.

Issue #24402, opened 2020-12-08
Add support for Java 16


dreis2211:
Hi,

I want to raise this issue to collect/track the needed things in order for JDK 16 to work.

- [x] Add JavaVersion.SIXTEEN
- [x] Adjust TestJarCreator multi-release jar tests
- [ ] Add build pipeline for JDK 16

**Related/Required** issues:

- Spring-Framework: https://github.com/spring-projects/spring-framework/issues/25264
- Gradle: https://github.com/gradle/gradle/issues/13481

I would volunteer do this again for JDK 16 after working on JDK 14 & JDK 15 in the past. I would start - probably tomorrow - with fixing the first two points on this list. The pipeline task requires Gradle support for JDK 16, so this would need to wait a bit.

Let me know what you think.
Cheers,
Christoph

snicoll:
Ah ah, @dreis2211 this is too funny. We've discussed that very topic on the team meeting yesterday :) Sure, that sounds great and thanks a lot for your efforts!

wilkinsona:
I'd like to see us use Gradle's toolchain support so that we don't have to wait for their Java 16 support. We can run Gradle itself on Java 8 and compile and test with Java 16. This'll require our Java 16 CI image to contain JDK 8 and JDK 16.

dreis2211:
I talked with @snicoll earlier. I'll see what I can come up with. My current plan is to essentially replace & adjust the functionality that I once introduced for a custom `buildJavaHome` property. E.g. introduce a new parameter `toolchainVersion`, which - if available - configures the toolchain stuff inside `JavaConventions`. I might need a little help in the end for the CI image stuff though.

snicoll:
With the pipeline in place I think we're good with this one. Thanks again @dreis2211!

wilkinsona:
I've moved this one out of 2.4.4 as I think it's been superseded by some other issues and PRs:

- https://github.com/spring-projects/spring-boot/pull/24453
- https://github.com/spring-projects/spring-boot/pull/25519
- https://github.com/spring-projects/spring-boot/issues/25527

The last one will make the support "official" as it'll update the documentation.

PR #24453, opened 2020-12-10
Add preliminary support for JDK 16

dreis2211:
Hi,

this PR is a first step into supporting JDK 16 (see #24402 ), which can be tested when setting the custom `buildJavaHome` property (at least build `16.ea.27` is needed to have `Stream.toList()` available). For everything else in #24402 we need Gradle with JDK 16 support.

Cheers,
Christoph


snicoll:
Thanks @dreis2211. I've backported the fix for the test in `2.3.x` and updated `JavaVersion` for `2.4.x` onwards. While I'd like us to be able to build Spring Boot 2.3.x with Java 16, an official support of it is [only available as of Spring Framework 5.3](https://github.com/spring-projects/spring-framework/wiki/Spring-Framework-Versions#jdk-version-range) so introducing the enum in that version sends the wrong message IMO.

dreis2211:
@snicoll Agree about the enum

PR #24470, opened 2020-12-11
Add a "force-restart" command to force restart.

lizhuoxuan:
When I use spring-boot-loader-tools to publish Spring Boot application as a Linux service, sometimes my jar failed to stop due to some personal code error and I have to force stop before starting,like this “service my-app force-stop && service my-app start”. How about adding a "force-restart" command to force restart?

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/resources/org/springframework/boot/loader/tools/launch.script
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/resources/org/springframework/boot/loader/tools/launch.script
@@ -254,6 +254,10 @@ restart() {
   stop && start
 }
 
+force_restart() {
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the suggestion but I'd prefer to keep the actions as close to [those that are required](https://refspecs.linuxfoundation.org/LSB_1.2.0/gLSB/iniscrptact.html) as possible. Combining `force-stop` with `start` as you have done with `service my-app force-stop && service my-app start` feels like a fine solution to me without adding extra complexity to the default launch script. Let's see what the rest of the team thinks.

philwebb:
I think I agree with @wilkinsona. Whilst the changes aren't to the launch script are minimal, I think we're better off keeping close to the standard actions. I'll close this one for now, but if we get people commenting here we can always reconsider.

PR #24477, opened 2020-12-12
Use BindResult.orElseGet() where beneficial

izeye:
This PR changes to use `BindResult.orElseGet()` where beneficial.


snicoll:
Thanks again @izeye!

PR #24490, opened 2020-12-14
Allow spring.config.location directories to end with system-dependent separators

mederel:
In Windows the file path separator character is \ not /. Currently when
parsing a config location ending with \ in windows as it should for a
folder, an exception is raised complaining it does not end with '/'.

Fix: Add File.separator as valid suffix for directory detection


mederel:
Trying to see what the build failure is, but my browser gets stuck at loading the data...

snicoll:
@mederel thanks for the follow-up but given the change is small, we can review that build error as part of reviewing if it is related.

snicoll:
There is also [a dedicated note](https://docs.spring.io/spring-boot/docs/current/reference/html/spring-boot-features.html#boot-features-external-config) in the reference guide we'd have to update if we decide to pursue with this change.

philwebb:
We should also update the reference docs

philwebb:
@mederel Thanks very much for making your first contribution to Spring Boot. This has now been merged to master and will be in Spring Boot 2.5.

PR #24532, opened 2020-12-16
Avoid duplicate classes in MergedContextConfiguration

dreis2211:
Hi,

I just noticed the following log while debugging some test context cache optimizations in one of my projects:

```
19:44:41.758 [Test worker] DEBUG o.s.t.c.s... - Before test class: context [DefaultTestContext@209b736 ... classes = '{class com.example.Application, class com.example.Application}' ...], class annotated with @DirtiesContext [false] with mode [null].
```
It's trimmed down, but as you can see `classes` contains the same class twice. This seems to be happen when I explicitly declare the `classes` on `@SpringBootTest` as follows:
```
@SpringBootTest(classes = {Application.class})
```

While not critical at all, I guess we can make an effort to avoid the duplicates. That should fix the logs and maybe saves a few cycles on the road (although that impact should be veeery small as the duplicates are filtered in `SpringApplication` the latest).

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootTestContextBootstrapper.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootTestContextBootstrapper.java
@@ -135,7 +135,7 @@ public class SpringBootTestContextBootstrapper extends DefaultTestContextBootstr
 	}
 
 	private void addConfigAttributesClasses(ContextConfigurationAttributes configAttributes, Class<?>[] classes) {
-		List<Class<?>> combined = new ArrayList<>(Arrays.asList(classes));
+		Set<Class<?>> combined = new HashSet<>(Arrays.asList(classes));
 		if (configAttributes.getClasses() != null) {
 			combined.addAll(Arrays.asList(configAttributes.getClasses()));
 		}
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootTestContextBootstrapper.java
```diff
@@ -135,7 +135,7 @@ protected ContextLoader resolveContextLoader(Class<?> testClass,
 	}
 
 	private void addConfigAttributesClasses(ContextConfigurationAttributes configAttributes, Class<?>[] classes) {
-		List<Class<?>> combined = new ArrayList<>(Arrays.asList(classes));
+		Set<Class<?>> combined = new HashSet<>(Arrays.asList(classes));
```
wilkinsona, line 138: I wonder if this should be a `LinkedHashSet` or similar? Assuming that the ordering is preserved elsewhere, losing it here by switching to a `HashSet` could, in theory, have an effect on the beans that are defined as it'll change the order in which they're processed during refresh.

dreis2211: Good point - I'll change this.

philwebb:
Thanks @dreis2211. I merged this into 2.3.x, 2.4.x and master.

dreis2211:
@philwebb Thanks. On that note: is there a 2.5.x pipeline on Concourse yet?

snicoll:
>  is there a 2.5.x / master pipeline on Concourse yet?

[Yes](https://ci.spring.io/teams/spring-boot/pipelines/spring-boot-2.5.x)

dreis2211:
@snicoll Thanks, but the page is blank for me. Also, the build status badge inside the README can't be displayed at the moment. Seems like a permission problem.

wilkinsona:
The pipeline should be visible now.

dreis2211:
Indeed. Thanks

Issue #24547, opened 2020-12-17
Not work for both private registry and public registry


WqyJh:
This is my base image built upon paketobuildpacks/run:base named docker.example.com/run:base which has been pushed to my private registry with basic auth.
```bash
FROM paketobuildpacks/run:base

# Install packages (ffmpeg)
RUN echo "debconf debconf/frontend select noninteractive" | debconf-set-selections && \
  export DEBIAN_FRONTEND=noninteractive && \
  apt-get -y update && \
  apt-get -y upgrade && \
  apt-get -y --no-install-recommends install ffmpeg && \
  find /usr/share/doc/*/* ! -name copyright | xargs rm -rf && \
  rm -rf \
    /usr/share/man/* /usr/share/info/* \
    /usr/share/groff/* /usr/share/lintian/* /usr/share/linda/* \
    /var/lib/apt/lists/* /tmp/*

ARG cnb_uid=1000
ARG cnb_gid=1000
ARG distro_name="Ubuntu"
ARG distro_version="18.04"
ARG homepage="https://github.com/paketo-buildpacks/stacks"
ARG maintainer="Paketo Buildpacks"
ARG stack_id="io.buildpacks.stacks.bionic"
ARG description="ubuntu:bionic + openssl + CA certs + ffmpeg"
ARG released

RUN groupadd cnb --gid ${cnb_gid} && \
  useradd --uid ${cnb_uid} --gid ${cnb_gid} -m -s /bin/bash cnb

USER ${cnb_uid}:${cnb_gid}
LABEL io.buildpacks.stack.description=${description}
LABEL io.buildpacks.stack.distro.name=${distro_name}
LABEL io.buildpacks.stack.distro.version=${distro_version}
LABEL io.buildpacks.stack.homepage=${homepage}
LABEL io.buildpacks.stack.id=${stack_id}
LABEL io.buildpacks.stack.maintainer=${maintainer}
LABEL io.buildpacks.stack.metadata=${fully_qualified_base_image}
LABEL io.buildpacks.stack.mixins=${mixins}
LABEL io.buildpacks.stack.released=${released}
```
This is how I config bootBuildImage.
```groovy
bootBuildImage {
    docker {
        builderRegistry {
            username = "username"
            password = "password"
            url = "https://docker.example.com/v1/"
        }
    }
    imageName = "docker.example.com/app"
    runImage = "docker.example.com/run:base"
}
```

But I've got the following problem when executing `bootBuildImage` task.
```bash
[...]

PR #24549, opened 2020-12-17
Enable both public and private repository for pulling image during bootBuildImage

WqyJh:
As described in #24547, if using runImage in private registry which was built upon images in public registry, image pulling would fail because apply a basic auth on no-auth-required public registry.

Take the following configuration as explanation, basic auth should only be applied when the image tag was prefixed with `docker.example.com/`, should never be applied when the parent images of this image was belong to `docker.io`.
```bash
bootBuildImage {
    docker {
        builderRegistry {
            username = "username"
            password = "password"
            url = "https://docker.example.com/v1/"
        }
    }
    imageName = "docker.example.com/app"
    runImage = "docker.example.com/run:base"
}
```

The solution is quite simple:

- Just add n field `registryUrl` in `DockerConfiguration`, and save the url into it during the construction.
- Check wether the image `domain` was contained in `registryUrl`, if so, apply the basic auth, otherwise do not apply.

This only support `username/password` authentication, not for `token` authentication beause it doesn't provide an url.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Builder.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Builder.java
@@ -150,7 +150,12 @@ public class Builder {
 	private Image pullImage(ImageReference reference, ImageType imageType) throws IOException {
 		Consumer<TotalProgressEvent> progressConsumer = this.log.pullingImage(reference, imageType);
 		TotalProgressPullListener listener = new TotalProgressPullListener(progressConsumer);
-		Image image = this.docker.image().pull(reference, listener, getBuilderAuthHeader());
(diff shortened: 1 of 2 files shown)


pivotal-issuemaster:
@WqyJh Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24549)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24549) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@WqyJh Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24549)!

scottfrederick:
Thanks for the suggestion, but this approach doesn't cover all the use cases we'd want to consider when the builder and run images are in different registries. We'll keep the issue open to consider other approaches.

PR #24550, opened 2020-12-17
Remove deprecated SpringBootContextLoader#getArgs()

dreis2211:
Hi,

I just noticed this (since 2.2.7) deprecated method, that seems to have slipped through the usual deprecation removals. As the comment is not really clear about the alternative I simply used the method's body directly.

Let me know what you think.
Cheers,
Christoph


snicoll:
Thanks again @dreis2211

Issue #24540, opened 2020-12-16
WebServer.getPort() returns 0 instead of -1 for closed connections


philwebb:
The javadoc for `WebServer.getPort()` states that it returns "the port (or -1 if none)". The implementations appear to return `0`.

philwebb:
Flagging to see if we should change the code or the javadoc.

mbhave:
When setting the port `0` is used for selecting a random port and `-1` to disable. That makes me think the javadoc is correct.

wilkinsona:
The implementations could be correct too as I think we need to distinguish between being disabled (-1) and not knowing what the port is because the connector's stopped and it's configured to use an ephemeral port (0).

mbhave:
Good point @wilkinsona.

philwebb:
We're going to change the implementation not the javadoc

spartusch:
Hi everyone, I've created a PR for this issue https://github.com/spring-projects/spring-boot/pull/24606. I hope that's ok. It's my first contribution so I'm not familiar with the processes. In general the PR changes the implementations to return -1 instead of 0.

philwebb:
Thanks @spartusch, I'll close this issue in favor of PR ##24606 and we can continue any discussion there when we get a chance to review it.

PR #24606, opened 2020-12-23
WebServer implementations should return -1 if not listening on a port

spartusch:
The interface `org.springframework.boot.web.server.WebServer` defines that `getPort()` returns `-1` if the server isn't listening on a port. But the current implementations of the interface return 0 instead. This changes the implementations to return -1.

Please note: Jetty itself does return -2 as the local port after the connection has been closed. Therefore in its implementation getPort() maps all local ports <= 0 to -1 to match the method's contract.

See #24540


Koellewe (approved):
Thorough work.

philwebb:
Thanks very much for the compressive PR @spartusch. This has been merged to 2.3.x, 2.4.x and master.

PR #24620, opened 2020-12-29
Update yaml shortcomings wrong example

friscoMad:
This is a small documentation fix, I was working today with multi-document profile-specific YAML files and after reading the example I thought that directly that setup does not work, but that's not the case it is just the example is not correct.

The current example does work as expected as the active profile is the same as the one in the profile-specific file so the nested document is not filtered out, the file name or the active profile should be changed to make the filter out the nested document.

Anyway, that whole part and example probably should be moved to section 2.3.8 as it is no longer exclusive of Yaml files, it is related to multi-document files.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
@@ -971,7 +971,8 @@ The `SpringApplication` class automatically supports YAML as an alternative to p
 
 NOTE: If you use "`Starters`", SnakeYAML is automatically provided by `spring-boot-starter`.
 
-
+WARNING: YAML files cannot be loaded by using the `@PropertySource` or `@TestPropertySource` annotations.
+So, in the case that you need to load values that way, you need to use a properties file.
 
 ==== Mapping YAML to Properties
 YAML documents need to be converted from their hierarchical format to a flat structure that can be used with the Spring `Environment`.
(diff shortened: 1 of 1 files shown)


pivotal-issuemaster:
@friscoMad Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24620)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24620) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@friscoMad Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24620)!

mbhave:
@friscoMad I don't think I understand the reason for changing the profile name to `prod`. That bit of documentation is trying to say that a profile-specific files (in this case, `application-dev.yml` that is loaded when the `dev` profile is active) cannot be used in conjunction with `spring.config.activate.on-profile`. 

However, I do agree that this section is true for any multi-document files and not just YAML. Would you like to update your PR to reflect that?

renhhh:
cnm

snicoll:
@friscoMad how is it going? Can you please let us know if you have the time to review the proposal as @mbhave indicated? Thanks!

friscoMad:
@snicoll @mbhave Sorry I had a week off and forgot about this PR.
I will try to update the PR to reflect that this common for yaml and properties but that will leave the yaml shortcomings section with just the @PropertySource warning.
But I need to be sure what was the intention of the example we are talking about, my understanding is that in the example we want to show a nested document with an activation condition that is true but the parent is a profile-specific file and can be fully skipped if the profile is not active.

In the current example the parent depends on `dev` and the child on `!test` but we are also saying that we run it with `--spring.profiles.active=dev`, if that's the case the file is processed and the nested document will be loaded as `test` is not enabled.
My proposal was changing the command line part to `prod` but we can just say that the `mypassword` will be set to "secret" if and only if the dev profile is active and `test` is not active. So that nested documents conditions in profile-specific config files have `& parent-profile` implicitly added to the activation condition.

mbhave:
@friscoMad The example is trying to show unexpected behavior when there is a nested document in a profile-specific file. It is saying that even if you have the `dev` profile activated and the `application-dev.yml` file is loaded, the nested document under `!test` will not be processed. This is due to the filtering of nested documents that occurs in profile-specific files. It is an example to show why you should not use `spring.config.activate.on-profile` with profile-specific files. Hope this answers your question.

mbhave:
@friscoMad My comment above is no longer valid since the behavior has changed in #24990. This change allows `spring.config.activate.on-profile` to be used in a profile-specific file. That makes that entire example invalid and we need to remove it. Sorry about the confusion. Since this change will go in `2.4.3`, we would like to update this bit before it is released (Feb 18th). Will you be able to make the change that removes this section before that? No worries if you can't, we can take care of it.

friscoMad:
Ok, after reviewing the mentioned ticket I think I understand my misunderstanding. I am working with 2.4.1 where that feature is working, it was on 2.4.2 when this stopped working, that's why I did not understand the example.
That said, this paragraph was there in the docs before 2.4, so it was not added with 2.4.2 regression, I don't know what was the behavior in older versions but I agree with you that if what you describe was the original intent of the documentation is better to just remove it.
I will modify it and remove the paragraph.

friscoMad:
@mbhave Done! Please confirm that this is what you expected.

mbhave:
Thanks @friscoMad. This has been merged into `2.4.x` and `master`. I decided to [move the warning](https://github.com/spring-projects/spring-boot/commit/52a15647a910b1e148be12205a795577b8e70215) to the bottom of the section.

PR #24628, opened 2021-01-01
Remove unnecessary semicolons

andrey4623:


snicoll:
@andrey4623 thanks again. I've found a few more that I've polished in https://github.com/spring-projects/spring-boot/commit/f33c7bc27cd51877cf6760c0716e494b3d6efafe

PR #24629, opened 2021-01-01
Make a field final

andrey4623:


snicoll:
@andrey4623 good catch. Thank you for making your first contribution to Spring Boot.

PR #24640, opened 2021-01-04
Add Revved up by Gradle Enterprise badge to Readme

runningcode:
Add Revved up by Gradle Enterprise badge to Readme


wilkinsona:
Thanks very much, @runningcode. I really like the badge.

Issue #20505, opened 2020-03-11
Disable error message on default error page unless explicitly enabled


mbhave:
We could display a generic message by default. We could reuse the `includeStackTrace` property or add a new property.

wilkinsona:
IIRC, when we discussed this recently we decided that we would use a separate property.

wilkinsona:
I'm reopening this as I'm not sure about the property being named `include-details`. I wonder if people may be confused by having a details property that control both the exception message and any binding error and a stack trace property that control only the stack trace. Arguably, all three could be considered detailed information. IIRC, the problem that prompted this issue being raised was only concerned with the exception message. While I think it's a good idea to control the inclusion of binding errors are well, I wonder if we'd be better with two new properties.

wilkinsona:
Looking at https://github.com/spring-projects/spring-boot/issues/21132 again, I also wonder if a generic message for all response statuses might be a bit misleading. For example, we're now returning an error response with the message "An error occurred while processing the request" as a result of a `ResponseStatusException` that maps to a 404. I'm not sure that the message is appropriate for a 404. Arguably, it's not really an error or, if it is, it's a client-side error so it didn't occur while the server was processing the request. I'm wondering if we should omit the message entirely or return a different message depending on the response status. Given that the property is about inclusion, omitting entirely is probably more inline with what people will expect.

scottfrederick:
> While I think it's a good idea to control the inclusion of binding errors are well, I wonder if we'd be better with two new properties.

When the error is caused by a binding exception, the handling of the `message` and `errors`  fields in the response becomes a bit tangled. Both fields will leak similar internal details of the application (e.g. object, class, and field names). It would be more clear to have separate `include-message` and `include-binding-errors` fields, but it's hard to imagine a case where a user would want to set those two flags to different values. For example, would a user want to include the `message` field that gives the name of the object that failed to bind but exclude the `errors` with more details? 

I agree that the name is vague and requires more information to understand what it controls. Splitting it would be easy enough to implement, as would changing the name to something more clear (maybe `include-message-errors`?). 

> I also wonder if a generic message for all response statuses might be a bit misleading. 

This was debated a bit during the implementation. Previously, the `message` field in a servlet error response was always present and never empty (it contained the value `No message available` if the exception contains an empty message). There was concern that omitting the field or leaving it empty would break the existing contract, but the idea of a contract w.r.t. fields in a `Map` is not clear to me. 

[...]

PR #24661, opened 2021-01-06
Remove redundant judgment conditions in DefaultErrorAttributes#getErr…

weixsun:
https://github.com/spring-projects/spring-boot/blob/b3278f459e9f69bac948e691d07d79330da7ff20/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/error/DefaultErrorAttributes.java#L121

https://github.com/spring-projects/spring-boot/blob/b3278f459e9f69bac948e691d07d79330da7ff20/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/error/DefaultErrorAttributes.java#L98

I carefully checked the DefaultErrorAttributes.java code context, I think the judgment statement `errorAttributes.get("message") != null` will always be true, and even if it is not true, this judgment will not affect the `execution of errorAttributes.put("message", "");`

So I think it is not necessary to have a conditional judgment

I understand the source code is `errorAttributes.put("message", "");` instead of `errorAttributes.remove("message");` because no matter whether `server.error.include-message` is `always` or `never`, the response body should always be Contains the `message` field, right? If it is, it should be indicated in the configuration file to distinguish it from other server.error.*, because the current explanation says `When to include "message" attribute`. It is easy for users to misunderstand.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ErrorProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ErrorProperties.java
@@ -46,7 +46,7 @@ public class ErrorProperties {
 	private IncludeAttribute includeStacktrace = IncludeAttribute.NEVER;
 
 	/**
-	 * When to include data in "message" attribute.
+	 * When to persist data in "message" attribute.
 	 */
 	private IncludeAttribute includeMessage = IncludeAttribute.NEVER;
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ErrorProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ErrorProperties.java
@@ -46,7 +46,7 @@ public class ErrorProperties {
 	private IncludeAttribute includeStacktrace = IncludeAttribute.NEVER;
 
 	/**
-	 * When to include "message" attribute.
+	 * When to include data in "message" attribute.
 	 */
 	private IncludeAttribute includeMessage = IncludeAttribute.NEVER;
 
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@weixsun Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24661)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24661) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@weixsun Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=24661)!

philwebb:
I can't remember why we add an empty message rather than removing the entry. It might have been for back-compatibility. I wonder if we should change things in 2.5.x so that the entry is removed entirely.

scottfrederick:
I also suspect that the `message` attribute is handled differently from the other attributes for back-compatibility reasons, although it does not appear we have tests that verify the existing behavior. If we make a change like the one suggested I agree it should be in a minor release not a patch.

weixsun:
https://github.com/spring-projects/spring-boot/commit/a30740f8d2e590c1f3adc20fbc4ea9c653c5a94b

weixsun:
> I can't remember why we add an empty message rather than removing the entry. It might have been for back-compatibility. I wonder if we should change things in 2.5.x so that the entry is removed entirely.

Add empty message from here : https://github.com/spring-projects/spring-boot/issues/20505#issuecomment-621295137

philwebb:
We discussed this today as a team and decided to try to remove the error attribute rather than blank it. We'll handle that as an enhancement (#24712, hopefully in 2.5).

Looking at the current PR, I think it's possible that a user may have overridden the `DefaultErrorAttributes.getErrorAttributes` that returns the `Map` and may also have removed the `message` entry. As such, I don't think it's safe to assume that the judgment statement can be removed.

Thanks for the PR and for prompting the discussion.

Issue #24666, opened 2021-01-06
Support get requests for /actuator/startup


philwebb:
Currently the `startup` actuator endpoint only supports `POST` because it clears the buffer. We should offer a `GET` version that leaves the buffer in place. This will be easier to call call because CSRF protection won't be needed.

nfrankel:
Subscribed 😅

jenshadlich:
I understand that `POST` makes sense because of the side effects. But it's totally unintuitive. So yes, please support `GET`.

bclozel:
Superseded by #24717

PR #24717, opened 2021-01-09
Support get requests for /actuator/startup

bono007:
Adds GET support for /actuator/startup

Fixes gh-24666


spring-boot-project/spring-boot-actuator-autoconfigure/src/docs/asciidoc/endpoints/startup.adoc
```diff
@@ -16,7 +16,9 @@ The resulting response is similar to the following:
 
 include::{snippets}/startup/http-response.adoc[]
 
-
+NOTE: The above call resets the application startup steps buffer - subsequent calls to the endpoint will
```
bono007, line 19: This may need some wordsmith polishing - it feels like I'm bleeding impl details to the doc.

snicoll:
Thanks again Chris! I've polished the documentation by adding a dedicated example with `GET` so that we give it a bit more importance in the Actuator API documentation.

Issue #24714, opened 2021-01-08
Add @Selector support to the configprops actuator endpoint


philwebb:
It would be nice if you could limit the output of `/actuator/configprops` to a specific value in the same way that you can with the environment endpoint.

bono007:
@philwebb I picked this up and wanted to run my thoughts on filtering by you before getting too much further. My plan is to:

**Filter by prefix:**  `/actuator/configprops?prefix=<config-props-prefix-goes-here>` 
**Filter by config props simple classname:** `/actuator/configprops/<config-props-simple-classname-goes-here>`

**Examples:**
- `/actuator/configprops?prefix=spring.kafka`
- `/actuator/configprops/ServerProperties`

snicoll:
I think it's worth discussing how to filter the data. I don't know what Phil had in mind but if we want to do something similar to `/env`, I think the selector should be the id of the entity in the result. Unfortunately, this id isn't super easy to remember. I don't think we should be using `SeverProperties`, in particular, as you may have two cases with the same name.

Given that the prefix is easier to remember (and something that's easier to resonate than a class name), my vote would be the first option in your examples.

bono007:
> Unfortunately, this id isn't super easy to remember. 

Yeh that is exactly what I ran into when I started w/ that as the filter. 

>  in particular, as you may have two cases with the same name.

Yes that is true. In that case they would both show up as well w/ what prefix they are mapped to. I added this filter by asking myself what would I want to filter by. Prefix for sure. The properties class is the next thing I think of - `KafkaProperties` is one I end up looking at multiple times a day for example. Because I seem to find my way into those configprops classes in my editor I am familiar w/ the class name so it seemed like a good candidate. For me, they are both valuable.

philwebb:
Thanks for looking at this @bono007. I added the issue as a note to myself so it's a bit light on details.

I think most users will probably want to filter on the property prefix, since that's the most user-facing value. I would suggest trying `/actuator/configprops/<prefix>` and then filtering the results so that only configuration properties starting with the given prefix are shown.

bono007:
@snicoll @philwebb I will get the updates in w/in the next 24-48 hrs

bono007:
@snicoll @philwebb I pivoted the merge request to filter only by prefix and also added the docs.

philwebb:
Thanks @bono007!

wilkinsona:
Closing in favour of #24718.

PR #24718, opened 2021-01-10
Allow the configprops endpoint to return properties with a particular prefix

bono007:
Adds ability to filter by prefix for /actuator/configprops

Fixes gh-24714

### Filter by prefix
(ℹ️ this was taken from v1 screenshot and url does not contain prefix param anymore)

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/docs/asciidoc/endpoints/configprops.adoc
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/docs/asciidoc/endpoints/configprops.adoc
@@ -6,15 +6,15 @@ The `configprops` endpoint provides information about the application's `@Config
 
 
 [[configprops-retrieving]]
-== Retrieving the @ConfigurationProperties Bean
+== Retrieving All @ConfigurationProperties Beans
 
-To retrieve the `@ConfigurationProperties` beans, make a `GET` request to `/actuator/configprops`, as shown in the following curl-based example:
+To retrieve all of the `@ConfigurationProperties` beans, make a `GET` request to `/actuator/configprops`, as shown in the following curl-based example:
 
-include::{snippets}/configprops/curl-request.adoc[]
+include::{snippets}/configprops/all/curl-request.adoc[]
 
 The resulting response is similar to the following:
 
-include::{snippets}/configprops/http-response.adoc[]
+include::{snippets}/configprops/all/http-response.adoc[]
 
 
 
@@ -26,3 +26,26 @@ The following table describes the structure of the response:
 
 [cols="2,1,3"]
 include::{snippets}/configprops/response-fields.adoc[]
+
+
+[[configprops-retrieving-by-prefix]]
+== Retrieving @ConfigurationProperties Beans By Prefix
+
+To retrieve the `@ConfigurationProperties` beans mapped under a certain prefix, make a `GET` request to `/actuator/configprops/{prefix}`, as shown in the following curl-based example:
+
+include::{snippets}/configprops/prefixed/curl-request.adoc[]
+
+The resulting response is similar to the following:
+
+include::{snippets}/configprops/prefixed/http-response.adoc[]
+
+NOTE: The `{prefix}` does not need to be exact, a more general prefix will return all beans mapped under that prefix stem.
+
+[[configprops-retrieving-by-prefix-response-structure]]
+=== Response Structure
+
+The response contains details of the application's `@ConfigurationProperties` beans.
+The following table describes the structure of the response:
+
+[cols="2,1,3"]
+include::{snippets}/configprops/prefixed/response-fields.adoc[]
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
@@ -25,6 +25,7 @@ import java.util.HashMap;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.function.Predicate;
 import java.util.stream.Collectors;
 
 import com.fasterxml.jackson.annotation.JsonInclude.Include;
@@ -54,6 +55,7 @@ import org.springframework.beans.BeansException;
 import org.springframework.boot.actuate.endpoint.Sanitizer;
 import org.springframework.boot.actuate.endpoint.annotation.Endpoint;
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
+import org.springframework.boot.actuate.endpoint.annotation.Selector;
 import org.springframework.boot.context.properties.BoundConfigurationProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.ConfigurationPropertiesBean;
@@ -85,6 +87,7 @@ import org.springframework.util.StringUtils;
  * @author Stephane Nicoll
  * @author Madhura Bhave
  * @author Andy Wilkinson
+ * @author Chris Bono
  * @since 2.0.0
  */
 @Endpoint(id = "configprops")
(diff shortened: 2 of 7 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
```diff
@@ -107,17 +111,36 @@ public void setKeysToSanitize(String... keysToSanitize) {
...
+	 * @return info for all registered configuration properties
+	 * @deprecated in favor of {@link #configurationProperties(String)}
+	 */
 	public ApplicationConfigurationProperties configurationProperties() {
-		return extract(this.context);
+		return configurationProperties(null);
+	}
+
+	@ReadOperation
+	public ApplicationConfigurationProperties configurationProperties(@Nullable String prefix) {
```
bono007, line 124: I chose to keep the filtering somewhat constrained and not implement a generic end-all-be-all regex pattern. This does allow matching a broader prefix as it does a `starts with` on it.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
```diff
@@ -107,17 +111,36 @@ public void setKeysToSanitize(String... keysToSanitize) {
...
+		Predicate<ConfigurationPropertiesBean> prefixPredicate = !StringUtils.hasText(prefix) ? (bean) -> true
+				: (bean) -> bean.getAnnotation().prefix().startsWith(prefix);
+		return extract(this.context, prefixPredicate);
 	}
 
-	private ApplicationConfigurationProperties extract(ApplicationContext context) {
+	@ReadOperation
+	public ApplicationConfigurationProperties configurationPropertiesByClassName(@Selector String simpleClassName) {
```
bono007, line 131: Also kept it constrained to a exact match on the simple class name of the backing ConfigurationProperties instance.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
```diff
@@ -164,10 +187,14 @@ private void applySerializationModifier(ObjectMapper mapper) {
...
 
-	private ContextConfigurationProperties describeBeans(ObjectMapper mapper, ApplicationContext context) {
+	private ContextConfigurationProperties describeBeans(ObjectMapper mapper, ApplicationContext context,
+			Predicate<ConfigurationPropertiesBean> beanFilterPredicate) {
 		Map<String, ConfigurationPropertiesBean> beans = ConfigurationPropertiesBean.getAll(context);
-		Map<String, ConfigurationPropertiesBeanDescriptor> descriptors = new HashMap<>();
-		beans.forEach((beanName, bean) -> descriptors.put(beanName, describeBean(mapper, bean)));
+
+		Map<String, ConfigurationPropertiesBeanDescriptor> descriptors = beans.values().stream()
+				.filter(beanFilterPredicate::test)
```
bono007, line 195: I took the approach of simply filtering out at the config props level and not returning a different payload for the individual api  case. This differs form the `/env` endpoint as it returns an `XYZEntry` object instead of an `XYZ`. This provided the filtering and introduced less ripple.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpointWebIntegrationTests.java
```diff
@@ -0,0 +1,132 @@
...
+import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.test.web.reactive.server.WebTestClient;
+
+import static org.hamcrest.Matchers.greaterThanOrEqualTo;
+import static org.hamcrest.Matchers.hasSize;
+
+/**
+ * Integration tests for {@link ConfigurationPropertiesReportEndpoint} exposed by Jersey,
```
bono007, line 34: I noticed that we did not have `@WebEndpointTest` coverage for `/actuator/configprops` - this adds it.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpointWebIntegrationTests.java
```diff
@@ -0,0 +1,132 @@
...
+	}
+
+	@WebEndpointTest
+	void filterByClassName() {
+		this.client.get().uri("/actuator/configprops/Foo").exchange().expectStatus().isOk().expectBody()
+				.jsonPath("$..beans[*]").value(hasSize(1)).jsonPath("$..beans['fooDotCom']").exists();
+	}
+
+	@WebEndpointTest
+	void filterByNonExistantClassName() {
```
bono007, line 76: Another thing to point out is that I did not create a `WebEndpointExtension` variant of this. We would need to do so if we want the mismatched config props classname in the path to result in 404. As you can see here, it does not do that yet.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
```diff
@@ -107,17 +111,36 @@ public void setKeysToSanitize(String... keysToSanitize) {
 		this.sanitizer.setKeysToSanitize(keysToSanitize);
 	}
 
-	@ReadOperation
+	/**
+	 * Retrieves info for all registered configuration properties.
+	 * @return info for all registered configuration properties
+	 * @deprecated in favor of {@link #configurationProperties(String)}
```
bono007, line 117: So at first I yanked this method out completely w/ the thought that the actuator endpoint path is replaced w/ the newer `@ReadOperation` in `configurationProperties(@Nullable String prefix)`. However, its possible that someone has coded directly to the endpoint (which means that someone has) so I marked deprecated. Lmk if there is anything else I need to do wrt docs for this. 

Remember - I have intentionally left docs out until I get a nod on the current code proposal.

spring-boot-project/spring-boot-actuator-autoconfigure/src/docs/asciidoc/endpoints/configprops.adoc
```diff
@@ -26,3 +26,26 @@ The following table describes the structure of the response:
 
 [cols="2,1,3"]
 include::{snippets}/configprops/response-fields.adoc[]
+
+
+[[configprops-retrieving-by-prefix]]
```
bono007, line 31: @wilkinsona I absolutely ❤️ what you did here w/ the endpoint doc generation - clever. It provides a consistent and accurate set of docs as well as provides some test coverage. Thank you.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpointWebIntegrationTests.java
```diff
@@ -0,0 +1,128 @@
...
+	}
+
+	@WebEndpointTest
+	void filterByGeneralPrefix() {
+		this.client.get().uri("/actuator/configprops/com.").exchange().expectStatus().isOk().expectBody()
+				.jsonPath("$..beans[*]").value(hasSize(2)).jsonPath("$..beans['fooDotCom']").exists()
+				.jsonPath("$..beans['barDotCom']").exists();
+	}
+
+	/*
```
bono007, line 69: I left this in intentionally to ask the question, should we add a Web Extension for this enpdoint and 404 when the prefix does not match anything?

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpointWebIntegrationTests.java
```diff
@@ -0,0 +1,126 @@
...
+
+	@WebEndpointTest
+	void filterByGeneralPrefix() {
+		this.client.get().uri("/actuator/configprops/com.").exchange().expectStatus().isOk().expectBody()
+				.jsonPath("$..beans[*]").value(hasSize(2)).jsonPath("$..beans['fooDotCom']").exists()
+				.jsonPath("$..beans['barDotCom']").exists();
+	}
+
+    @WebEndpointTest
+    void filterByNonExistentPrefix() {
```
bono007, line 70: Should we return 404 in case of no prefix match? If so I need to create a web endpoint extension. Just let me know what the preference is here. Because the prefix can match multiple entries it feels more of a "List findByPrefix" than a "Object findByPrimaryKey" (if you will).

wilkinsona: A 404 feels right to me. That's what we do in the `env` endpoint for a property that doesn't exist. It too can match multiple entries as the same property may be defined in multiple property sources.

spring-boot-project/spring-boot-actuator-autoconfigure/src/docs/asciidoc/endpoints/configprops.adoc
wilkinsona: That's lovely to hear. Thank you!

snicoll:
nicely done Chris. I've polished your contribution in b92bb93. The interesting bits is that we name the method differently when there is a selector so that the JMX operation (that uses the method name) can provide a bit more context as what it does compared to the general operation.

Thanks again!

PR #24789, opened 2021-01-12
Upgrade to Gradle Enterprise Maven Extension 1.8.1

runningcode:


wilkinsona:
Thanks again, @runningcode. I polished the change a little bit as the version's declared in two places.

PR #24849, opened 2021-01-15
Fix ConfigDataProperties blank config import NPE

mhyeon-lee:
Filter ConfigDataLocation imports null elements.

It could be NPE at [ConfigDataEnvironment#getMandatoryImports](https://github.com/spring-projects/spring-boot/blob/48e00fc7f4283f4a2b0277f1304386621d8bc5f0/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataEnvironment.java#L377)

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataProperties.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataProperties.java
@@ -18,7 +18,9 @@ package org.springframework.boot.context.config;
 
 import java.util.Collections;
 import java.util.List;
+import java.util.Objects;
 import java.util.function.Predicate;
+import java.util.stream.Collectors;
 
 import org.springframework.boot.cloud.CloudPlatform;
 import org.springframework.boot.context.properties.bind.BindContext;
@@ -35,6 +37,7 @@ import org.springframework.util.ObjectUtils;
  *
  * @author Phillip Webb
  * @author Madhura Bhave
+ * @author Myeonghyeon Lee
  */
 class ConfigDataProperties {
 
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for the pull request. I'm not sure that we should silently filter out null entries. If an import has been configured with a trailing comma, as shown in the tests (`spring.config.import: testimport.yml,`), this is likely to be due to a typo. I think the NPE is preferable to silently hiding the mistake, but something that fails and describes the cause of the null entry (if we can figure that out) would be preferable to a NPE. Let's see what the rest of the team thinks.

mhyeon-lee:
@wilkinsona 
Thank you for your review.
If Exception is a better choice than filtering, I will change it to throw NPE with a message explaining the reason.

prasoonanand:
The issue is with `delimitedStringToCollectionConverter.getElements(source, delimiter)`
which in turn calls `StringUtils.delimitedListToStringArray(source, Delimiter.NONE.equals(delimiter) ? null : delimiter)`...
For an example string "a,b," the method of `StringUtils` returns ["a", "b", ""]... 
whereas Java String class `String.split(delimiter)` returns ["a", "b"]...

Don't know if we call this a issue at spring bot end...

scottfrederick:
@mhyeon-lee We received issue #26342, which I fixed in https://github.com/spring-projects/spring-boot/commit/dbee0cf9f1c8ed07bda4609fa7552ffdfcef7e1d. The commit removes the empty `spring.config.import` entries in a different place than the changes in this PR, which results in empty entries also being removed for `spring.config.location` when a comma-separated list is provided for that property. 

Since the existing commit has broader coverage, I'll close this PR. I apologize for not realizing that we had an open PR that addressed the same problem and giving you a chance to modify your contribution. Thanks very much for noticing the issue first and for submitting the PR, and we look forward to more contributions in the future.

PR #24851, opened 2021-01-15
Configuration metadata for logging.charset.* has invalid reference for java.nio.charset.Charset

dreis2211:
Hi,

this PR fixes the configuration metadata for `logging.charset.console` and `logging.charset.file` by adjusting the wrong type information. 

I only noticed in IDEA that no warning was shown when I entered a wrong charset encoding.
image
I guess other places consuming the metadata could be affected as well.

Cheers,
Christoph


wilkinsona:
Thanks, @dreis2211. I've opened https://github.com/spring-projects/spring-boot/issues/24852 to see if we can catch this sort of mistake automatically.

Issue #24732, opened 2021-01-11
Add customizer for jOOQ's DefaultConfiguration


snicoll:
jOOQ has a number of "provider" interface one can implement to tune its behaviour. We've started to look for beans of such type to automatically customise the `DefaultConfiguration` jOOQ is using.

Over the years, an increasing number of callback interfaces have been added which makes our auto-configuration a bit verbose. Also, it is unclear if those implementations have to be bean in the first place. It feels like they are registered as beans for the sole purpose of being found by the auto-configuration.

Given the current arrangement, we've decided to provide a customizer and let users tune the configuration going forward.

snicoll:
This is noteworthy as checking for the presence of the following beans is now deprecated:

* `ConnectionProvider`
* `ExecutorProvider`
* `TransactionProvider`
* `RecordMapperProvider`
* `RecordUnmapperProvider`
* `Settings`
* `RecordListenerProvider`
* `ExecuteListenerProvider`
* `VisitListenerProvider`
* `TransactionListenerProvider`

Users relying on this behaviour can restore the configuration by registering a `DefaultConfigurationCustomizer` bean that customize the `DefaultConfiguration`. Note that the underlying provider(s) no longer need to be registered as a bean.

PR #24862, opened 2021-01-18
Polish

izeye:
This PR changes to configure jOOQ `ConnectionProvider` via customizer as it seems to be the original intention based on #24732.


snicoll:
Ah ha. Good catch there @izeye, I forgot that one.

snicoll:
Actually, let me take that back. I did that on purpose, but didn't do it properly anyway. 

The plan with the customizer was to only customise things that have been defined externally. I didn't configure `ConnectionProvider` this way as the plan is to deprecate the customizer eventually and that would stop configuring the `ConnectionProvider` that we auto-configure ourselves.

Re-reading the code, I noticed we also auto-configure a `DefaultExecuteListenerProvider`, so the processing of `ExecuteListenerProvider` should rather move to the main auto-configuration for the same reason.

Would you be willing to review your PR in that direction?

izeye:
@snicoll Thanks for the feedback! I missed it. I updated as you suggested.

snicoll:
Thanks for the quick feedback Johnny! I've polished by adding assertions outside of that deprecated test to make it more obvious those two providers will be honoured once we remove the customizer (see dcc0ca0).

PR #24865, opened 2021-01-18
Add possibility to disable event propagation

GreenRover:
Add possibility to disable event propagation,

this is required to fix the broken "readiness" event propagation of SpringCloudStream.
For further Information see:
https://github.com/spring-cloud/spring-cloud-stream/issues/2083


wilkinsona:
Thanks for the proposal, @GreenRover. Unfortunately, I think an option to disable all run listeners is too broad for a problem that is specific to Spring Cloud Stream. Let's continue the discussion on the SCS issue to determine what, if anything, needs to be done in Spring Boot.

PR #24866, opened 2021-01-18
Restore persistOfNullThrowsIllegalArgumentExceptionWithoutExceptionTranslation()

izeye:
`PersistenceExceptionTranslationAutoConfigurationTests.persistOfNullThrowsIllegalArgumentExceptionWithoutExceptionTranslation()` has been ignored in 59d2b0a3fb2e9f83345c23b32e5bbac47c74dbd2 somehow. I'm not sure if it was intentional, but it seems to pass for now. This PR simply restores it.


snicoll:
>  I'm not sure if it was intentional

Given how it has been commented out, I suspect this was a temporary measure that got pushed by accident. Thanks again for spotting that @izeye!

Issue #24321, opened 2020-12-03
Exceptions are still translated by JdbcTransactionManager when spring.dao.exceptiontranslation.enabled is false


Gama11:
We ran into an issue with the 2.4.0 release because it replaced `DataSourceTransactionManager` with `JdbcTransactionManager`. The latter does exception translation:

- https://github.com/spring-projects/spring-boot/pull/23572 / https://github.com/spring-projects/spring-boot/commit/9293b2629ecfe5931c41638b2b0db2cfc8aebfe5
- https://github.com/spring-projects/spring-framework/issues/24064#issuecomment-624619883

This led to a `DuplicateKeyException` exception being thrown rather than a `TransactionSystemException`. We didn't expect this because we had explicitly disabled exception translation:

```properties
spring.dao.exceptiontranslation.enabled=false
```

Perhaps `DataSourceTransactionManager` should continue to be used if this property is set to false?

snicoll:
Thanks for the report @Gama11. I overlooked that property when I did that change.

wilkinsona:
Please see https://github.com/spring-projects/spring-boot/issues/24867 for details of a refinement that will be made to the fix for this issue. It should be available in 2.4.3.

PR #24867, opened 2021-01-18
Auto-configured DataSourceTransactionManager uses spring.dao.exceptiontranslation.enable rather than spring.dao.exceptiontranslation.enabled to control exception translation

izeye:
This PR fixes property name for `spring.dao.exceptiontranslation.enabled`.


wilkinsona:
Thanks very much, @izeye. This is a refinement of the fix made for https://github.com/spring-projects/spring-boot/issues/24321.

philwebb:
Gahhh, how frustrating that I messed that up. Thanks for the fix @izeye !

Issue #22519, opened 2020-07-22
Document the need to use the launcher to be able to use application.* properties in a custom banner


orenw1:
Some of the banner properties not populated when running with unpacked jar and custom banner

When running Spring Boot application with custom banner and unpacked jar as documented in [1],
the following banner properties that described in [2] are not printed

*   application.version
*   application.formatted-version
*   application.title

[1] [Deploying to Containers](https://docs.spring.io/spring-boot/docs/current/reference/html/deployment.html#containers-deployment)

[2] [Customizing the Banner](https://docs.spring.io/spring-boot/docs/2.2.8.RELEASE/reference/html/spring-boot-features.html#boot-features-banner)

Spring Boot version: 2.3.1.RELEASE

i saw this #21705 that looks similar issue, but i tried with 2.3.1.RELEASE version and it still does not show the values.

[Example project that re-create the issue](https://github.com/orenw1/demo-banner)

wilkinsona:
Thanks for the sample. The problem is as @philwebb [described here](https://github.com/spring-projects/spring-boot/issues/21705#issuecomment-639278008). The change made in #21705 has not helped in your case as you are not using Spring Boot's launcher. This means that it doesn't get a chance to make it possible to load a manifest (which is the source of the application version) from a folder.

You can avoid the problem by unpacking the jar and then using `JarLauncher` to start the app:

```
$ java -cp . org.springframework.boot.loader.JarLauncher
DEMO BANNER START

application version           = 0.0.1-SNAPSHOT
application formatted version =  (v0.0.1-SNAPSHOT)
spring-boot version           = 2.3.1.RELEASE
spring-boot formatted version =  (v2.3.1.RELEASE)
application title             = demo-banner

DEMO BANNER END
2020-07-23 09:41:21.631  INFO 2004 --- [           main] c.e.demobanner.DemoBannerApplication     : Starting DemoBannerApplication v0.0.1-SNAPSHOT on fulcrum.local with PID 2004 (/Users/awilkinson/dev/temp/demo-banner/target/unpacked/BOOT-INF/classes started by awilkinson in /Users/awilkinson/dev/temp/demo-banner/target/unpacked)
2020-07-23 09:41:21.633  INFO 2004 --- [           main] c.e.demobanner.DemoBannerApplication     : No active profile set, falling back to default profiles: default
[...]

PR #24982, opened 2021-01-24
Document the need to use the launcher to be able to use application.* properties in a custom banner

robert-smith-911:
This is a documentation update to describe the need to use JarLauncher to use application.* properties in custom banner. This is in response to issue [22519](https://github.com/spring-projects/spring-boot/issues/22519)


philwebb:
Thanks @robert-smith-911. I've merged this to 2.3.x and forward with a slight tweak to the wording since you you don't need the `-cp .` option when using the `JarLauncher`.

robert-smith-911:
Thanks @philwebb !

PR #25005, opened 2021-01-27
Introduce support for configuring SeedNodes as an alternative to the connectionString

### LINKED ISSUE #24987: Support configuring Couchbase Seed Nodes
Currently it's not possible to use Spring Boot's Couchbase auto configuration when running Couchbase locally in docker. This is because the Couchbase SDK does not accept alternative service ports in the connection string, a different API is to be used instead.

See Documentation here:
https://docs.couchbase.com/java-sdk/current/howtos/managing-connections.html#alternate-addresses

I propose that new properties be added to `org.springframework.boot.autoconfigure.couchbase.CouchbaseProperties`

```java
    private List<SeedNode> seedNodes = new ArrayList<>();
....  
    public static class SeedNode {
        private String address;
        private Integer kvPort;
        private Integer clusterManagerPort;
    }
```
So the configuration looks something like this:

```
spring.couchbase.seed-nodes[0].address=localhost
spring.couchbase.seed-nodes[0].kv-port=12345
spring.couchbase.seed-nodes[0].cluster-manager-port=54321
```

And then `org.springframework.boot.autoconfigure.couchbase#couchbaseCluster` can be modified something like this:

```java
       ClusterOptions options = ClusterOptions.clusterOptions(properties.getUsername(), properties.getPassword())
                .environment(couchbaseClusterEnvironment);

        if (properties.getSeedNodes().isEmpty()) {
            return Cluster.connect(properties.getConnectionString(), options);
        } else {
            Set<SeedNode> seedNodes = properties.getSeedNodes()
                    .stream()
                    .map(s -> SeedNode.create(s.getAddress(), Optional.ofNullable(s.getKvPort()), Optional.ofNullable(s.getClusterManagerPort())))
                    .collect(Collectors.toSet());
            return Cluster.connect(seedNodes, options);
        }
```

aaronjwhiteside:
Fixes #24987

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/couchbase/CouchbaseAutoConfiguration.java
@@ -18,12 +18,16 @@ package org.springframework.boot.autoconfigure.couchbase;
 
 import java.net.URL;
 import java.security.KeyStore;
+import java.util.Optional;
+import java.util.Set;
+import java.util.stream.Collectors;
 
 import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.TrustManagerFactory;
 
 import com.couchbase.client.core.env.IoConfig;
 import com.couchbase.client.core.env.SecurityConfig;
+import com.couchbase.client.core.env.SeedNode;
 import com.couchbase.client.core.env.TimeoutConfig;
 import com.couchbase.client.java.Cluster;
 import com.couchbase.client.java.ClusterOptions;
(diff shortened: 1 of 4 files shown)


wilkinsona:
When merging this, we should rename `SeedNode` to `Seednode` to avoid a `-` in the property prefix.

snicoll:
We may also want to change the condition to avoid relying on `"spring.couchbase.seed-nodes[0].address"`. 

@daschl what do you think? `SeedNode` doesn't feel like a 100% replacement for `connectionString` to me and I am wondering if we may be missing something. Ideally, I'd like whatever we introduce here to be a "super set" of what connection string can do. Thoughts?

daschl:
The `SeedNode` ist just an alternate way of configuring the bootstrap host list, and is mainly only useful if you want to provide both a custom cluster manager port and a custom kv port. This mostly happens in testing if you run against custom docker setups for example.

Note that we actually internally from a connection string build up the seed nodes. Note that for the purpose of spring boot, I'm not sure if we actually need to expose it. The reason is, and maybe this is what @aaronjwhiteside  is missing, you CAN actually configure both ports in the connection string like: `couchbase://foo:1234=manager,foo:5678=kv`, and we'll internally coalesce the foo host to use 1234 for the cluster manager port and 5678 for the kv port.

aaronjwhiteside:
> The reason is, and maybe this is what @aaronjwhiteside is missing, you CAN actually configure both ports in the connection string like: couchbase://foo:1234=manager,foo:5678=kv, and we'll internally coalesce the foo host to use 1234 for the cluster manager port and 5678 for the kv port.

Oh I am missing this!

Might want to update the documentation at https://docs.couchbase.com/java-sdk/current/howtos/managing-connections.html#alternate-addresses
Where SeedNodes are the recommended way to specify alternative ports.

I also hope this isn't documented because it's not intended to be a support feature going forward?

snicoll:
Alright so I am going to close this now per @daschl comment. Thanks for the pr @aaronjwhiteside, in any case!

Issue #24688, opened 2021-01-07
Allow application properties to be contributed from module jars


philwebb:
Currently `application.properties` and `application.yaml` files are loaded from the classpath and the first match wins. This makes it hard to bundle common properties into jar files. It would be nice if we could find a way for jars to also contribute properties without needing to implement a `EnvironmentPostProcessor`.

There are a few things we need to be careful of:

* What order to we load the resources and how do we make sure a developer can override a contributed value
* We don't want to increase startup time (for example by doing a full classpath search)
* We can't break back-compatibility
* We need to be mindful of `spring.config.import=classpath:`

mbhave:
Ordering will be tricky with this one. If two jars happen to provide `application.properties` files with the same value, we need a way to determine which one will win.

alex-lx:
Would using an `order` property in application.properties be a simple way?

philwebb:
We should look into the use-case behind #25033 when we fix this one.

stategen:
@philwebb & @wilkinsona ,thanks for your great work!
--spring.config.intergration-location=classpath*:/sink//[application.yml] ,in my pull request #25082 ,similar #25080 
the dependency jar's configs just to reduce the duplicate or mistake works by the the project who use it. 
so the rule: lowest order and can be overrided by any, not care conflicts(use map value),not overrides higher/outside value, limit 2 wildcards ,endswith '*/' or filename and filtered by '/fold/' .
the rule will never break back-compatibility.
and  with  a new arg to ensure never break back-compatibility.
In my real project（240 jars） with the update i pulled, the startup time is not slow,just normally.
I perfer to use spring in xml(<resource import="classpth*:">)  ,but more and more cloud-framework only provide with spring-boot'style config,there is really no back-way to use spring-boot

stategen:
I perfer to use spring in xml() = I perfer to use spring in xml(&lt; resource import="classpath:*"&gt;)

wilkinsona:
We should also consider https://github.com/spring-projects/spring-boot/issues/25084 when looking at this.

jeffbswope:
I was mulling something along the lines of #25084 like making `spring.config.name` multi-value (or an additional property) and letting libraries contribute their own values to that.

Maybe all becomes absurd at some point and ordering/precedence remains difficult but for most of our cases anyway you wouldn't expect collisions/conflicts since the libraries own their own vertical pieces.

[...]

PR #25033, opened 2021-01-28
Introduce SPI for automatic bean definition config registration

kevinraddatz:
In our project we created a starter which contains most general stuff, like database, feign clients and so on.
Now we want to configure this starter with different properties per stage.
Currently we created a file `application-commons.yaml` in our starter and include this "profile" at spring.profiles.active, but this does not really scale well if we create more starters, or more applications. We would have to keep in mind to add all the profiles to all the services.

With this PR we are able to implement an `AdditionalConfigLoader`, which returns the name of the starters "name", and then add the class to the `spring.factories`: `org.springframework.boot.context.config.AdditionalConfigLoader=example.AdditionalCommonConfigLoader`

We could now create multiple files like `commons.yaml`, `commons-local.yaml` and so.

This PR relates to https://github.com/spring-projects/spring-framework/pull/26458 which adds the ability to load an additional file like `application-commons.yaml`.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/AdditionalConfigLoader.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/AdditionalConfigLoader.java
@@ -1,0 +1,57 @@
+/*
+ * Copyright 2012-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.context.config;
+
+/**
+ * Load additional configs using
+ * {@link org.springframework.core.io.support.SpringFactoriesLoader}
+ * <p>
+ * Example: In this special scenario, where you want to have a spring-boot-starter with
+ * its own configname-profile.properties with configured properties for multiple profiles,
+ * you can implement this interface, so spring will include the defined configName in its
+ * configs.
+ *
+ * If you want to order your configs you can annotate your class with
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for the proposal. This is one possible way of implementing https://github.com/spring-projects/spring-boot/issues/24688.

kevinraddatz:
I think this would resolve most issues of #24688. 

- If two modules provide both `application.properties` nothing would change, but both could also provide their own files like `module1.properties` and `module2.properties`. Each module would then need to implement the `AdditionalConfigLoader`
- The files will be ordered using the `AnnotationAwareOrderComparator` that orders all `AdditionalConfigLoader`first by an `@PriorityOrdered` and second by `@Ordered` with a precedence. All unannotated class will be put at the end in an arbitrary order. This might be a problem, I don't know if this order could change within multiple starts, it would be the developers task to make sure that there are no overlapping properties. Only way around I see is overloading `SpringFactoriesLoader.loadFactories` with an other comparator that orders all unannotated classes by name or something else.
- Startup time should not be impaced affected, but I didn't do any measuments.
- It should be still backward compatible.
- `spring.config.import` should also not be impacted.

kevinraddatz:
Why is this pull request marked as draft?

philwebb:
I'm afraid I don't think that we can accept this pull-request in its current form. I feel like the we need to look at a different approach for #24688. I'm particularly concerned that having additional configuration only apply when `spring.config.name` is not bound will be quite limiting.

If that's what you want for your current setup, it might be possible to do it with an `EnvonmentPostProcessor`. If you have one that runs before the `ConfigDataEnvironmentPostProcessor` then you'd be able to insert a `PropertySource` with a `spring.config.name` value populated via your `AdditionalConfigLoader` interface. You can also add a `spring.profiles.include` which will remove the need for the `AdditionalProfileLoader` changes in Spring Framework.

I'll add a note to #24688 so we can consider this use-case when we work on the issue.

Thanks anyway for the pull-request.

wilkinsona:
@kevinraddatz I believe Brian marked it as a draft as you force-pushed changes to it several times after opening it. Each force-push notifies everyone watching the repository which adds a lot of noise. It also suggested that the changes were still a work in progress.

kevinraddatz:
Thank you for your reply, I understand your concerns. The idea to put this at such an early stage was to be able to also set `spring.main` properties, but I didn't think about risks, as a module also could allow `allow-bean-definition-overriding`.

I will try your approach with the `EnvironmentPostProcessor`, this seems to be enough for me. 

Sorry for the noise I created, I didn't knew about this.

PR #25036, opened 2021-01-28
Fix formatting violation

dreis2211:
Hi,

just noticed that the builds were failing.

Cheers,
Christoph


dreis2211:
@mbhave was just a second faster.

Issue #24605, opened 2020-12-23
Improve usage description for BufferedApplicationStartup


ascheman:
I was trying the nice new Spring Boot startup tracking feature and used `BufferedApplicationStartup` to collect initialization data. I found that the collected data is drained from the buffer after each request. In particular, it took me a moment to recognize that the most important data is lost after the first request. As discussed with @bclozel this could be emphasized in the documentation, cf. https://stackoverflow.com/questions/65413579/spring-boot-bufferedapplicationstartup-drained-after-first-request

oleg-sukhov:
Hello, @snicoll. I would like to make my first contribution to spring boot project. Do you think that this issue is a good start point? If yes, what should I do to start working on it?

snicoll:
@oleg-sukhov thanks for asking. If you've used the startup feature and see what needs to be documented, please go ahead and submit a PR. The documentation is located in `spring-boot-project/spring-boot-docs/src/docs/asciidoc` in this repository and the [contributing page](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc) should provide you more details to get started.

mbhave:
Closing in favor of PR #25075.

PR #25037, opened 2021-01-29
Add extra documentation to clarify the draining of startup steps on retrieval

robert-smith-911:
Documentation fix to clarify the draining of startup steps on retrieval when using [BufferingApplicationStartup](https://docs.spring.io/spring-boot/docs/2.4.0/api/org/springframework/boot/context/metrics/buffering/BufferingApplicationStartup.html)

Clarifies Issue #24605


wilkinsona:
Thanks for the PR, @robert-smith-911. We'd like to wait for a contribution from @oleg-sukhov for this one as they have asked if they can do so.

PR #25048, opened 2021-01-29
Fix Asciidoc build step

dreis2211:
Hi,

just noticed that the Asciidoc generation is failing due to a duplicate anchor ID.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #25080, opened 2021-02-02
add 'spring.config.intergration-location' with 'classpath*:' support

stategen:
This new feature  add a lowest  arg spring.config.intergration-location with value  'classpath*:' support.
In some cases, we integrate jars with an uncertain number, such as message subscribers. This new feature can make message sending truly dynamic and zero configuration for providers

PR #25082, opened 2021-02-02
add 'spring.config.intergration-location' with 'classpath*:' support

stategen:
The feature is  adding a lowest  argrement '--spring.config.intergration-location' which value support 'classpath*:' .
In some cases, the project integrates jars with an uncertain number, such as message subscribers. This can make a truly dynamic spring-stream and zero configuration for providers


wilkinsona:
Thanks for the proposal, @stategen. We don't want to add support for loading configuration from dependencies' jars but this isn't quite how we'd like to go about it. Please see https://github.com/spring-projects/spring-boot/issues/24688 for details. If you'd like to use what you have proposed here rather than waiting for #24688 to be implemented, you should be able to plug it in something similar using an `EnvironmentPostProcessor`. Thanks again for the suggestion.

wilkinsona:
I see this was also proposed as #25080. @stategen sorry that we couldn't merge your changes. Please don't spam the repository with duplicate pull requests.

stategen:
@wilkinsona ,thks and sorry, I'm not familiar with the operation of GitHub. My first submission failed because of the formatting problem. I thought it was not a submission, so I submitted it again

PR #25093, opened 2021-02-03
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
@@ -1055,7 +1055,7 @@ Also, if an application is using Kubernetes https://kubernetes.io/docs/tasks/run
 [[production-ready-kubernetes-probes-lifecycle]]
 ==== Application lifecycle and Probes states
 An important aspect of the Kubernetes Probes support is its consistency with the application lifecycle.
-There is a significant different between the `AvailabilityState` which is the in-memory, internal state of the application -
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/CorsEndpointProperties.java
```diff
@@ -46,7 +46,7 @@
 
 	/**
 	 * Comma-separated list of origin patterns to allow. Unlike allowed origins which only
-	 * supports '*', origin patterns are more flexible (for example
+	 * support '*', origin patterns are more flexible (for example
```
wilkinsona, line 49: I think "supports" is correct here.

izeye: @wilkinsona "origins" is plural, so I thought it needs to be "support". Am I missing something?

wilkinsona: While "allowed origins" is plural, it's referring to the allowed origins property and there's one of those.

izeye: @wilkinsona Thanks for the explanation! I reverted it.

Issue #23955, opened 2020-10-28
Update Java conventions so that projects creating jars from jars don't have to exclude duplicate LICENSE and NOTICE files


wilkinsona:
The build creates a handful of jars from other jars. This results in duplicate `LICENSE.txt` and `NOTICE.txt` files as they come both from the original jar and our conventions that add them to each jar that's built.

wilkinsona:
The first attempt at this caused #24131 so it needs to be reverted for now.

dreis2211:
May I give this an attempt, @wilkinsona ?

wilkinsona:
@dreis2211 That'd be great, thank you. As you can see above I had one attempt that fixed the duplicates but introduced another problem so perhaps use https://github.com/spring-projects/spring-boot/commit/1e100677c722db4f9839d39d4b104307d4917cc9 as an example of what not to do! I haven't had a chance to take another look.

wilkinsona:
When merged, https://github.com/spring-projects/spring-boot/pull/25116 will fix this problem by adding some excludes in a handful of projects that create a jar from an existing jar. It'd be nice to fix the conventions so that they don't create the problem in the first place.

philwebb:
I wonder if we should close this one. I think in general we still want our conventions to apply to all jar tasks to make sure we have the manifest customization. We could try and be smart about applying `extractLegalResources`, but there's no guarantee that the jar we're including actually has a `NOTICE.txt` and `LICENSE.txt`. I think the fix in #25116 might be enough.

wilkinsona:
Sounds good to me. We can always re-open this if we change our minds for some reason.

PR #25115, opened 2021-02-05
Fix Gradle deprecation warnings for duplicates

dreis2211:
Hi,

I was trying to build Boot with the newest Gradle 7 milestone to see what we have to do. And while there are certain bigger things like the `MavenPluginAction` that need to be addressed, I found a couple of deprecation warnings around setting the `duplicatesStrategy` that could be fixed already without too much effort. They are already deprecation warnings now and with 7 it fails.

I went with the warning strategy as I found it quite useful to see what exactly is duplicated instead of just including (or excluding) things. Currently it's the legal files that are duplicated. But let me know if you have a favourite here.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-cli/build.gradle
+++ b/spring-boot-project/spring-boot-cli/build.gradle
@@ -117,6 +117,7 @@ task fullJar(type: Jar) {
 			"Start-Class": "org.springframework.boot.cli.SpringCli"
 		)
 	}
+	duplicatesStrategy = DuplicatesStrategy.WARN
 }
 
 def configureArchive(archive) {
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks, @dreis2211. We already have details of the warnings on ge.spring.io. See https://ge.spring.io/s/3sustfylfr6e4/deprecations?expanded=WyI0Il0 for example. Rather than replacing the deprecation warning with a normal warning, I'd prefer to fix the problem by removing the duplicates and then configuring the duplicates strategy to be fail. https://github.com/spring-projects/spring-boot/issues/23955 is tracking that.

dreis2211:
I missed this, sorry @wilkinsona

PR #25116, opened 2021-02-05
Fix build deprecation warnings about duplicate jar entries

dreis2211:
Hi,

this is an attempt at fixing #23955. I say attempt because I chose to manually exclude the duplicate legal files in the few problematic places instead of doing things via the JavaConventions plugin (like in https://github.com/spring-projects/spring-boot/commit/1e100677c722db4f9839d39d4b104307d4917cc9).

That seems acceptable given the very few places we have, but let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-cli/build.gradle
+++ b/spring-boot-project/spring-boot-cli/build.gradle
@@ -108,7 +108,10 @@ task fullJar(type: Jar) {
 		into "BOOT-INF/classes"
 	}
 	into("") {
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks, @dreis2211. Ideally, I'd like to avoid the conventions creating the problem in the first place. That said, what you've proposed here is considerably better than what we have now. We'll merge this and repurpose #23955 to maybe find an alternative solution in the future.

Issue #25054, opened 2021-01-29
javax.persistence.schema-generation.database.action may be ignored


wilkinsona:
Either `javax.persistence.schema-generation.database.action` or `hibernate.hbm2ddl.auto` can be used to control schema creation when using Hibernate, with the latter property taking precedence over the former.

When `HibernateDefaultDdlAutoProvider` determines that the default should be `create-drop`, `hibernate.hbm2ddl.auto` is set with that value. Due to the Hibernate property taking precedence, if the user has set `javax.persistence.schema-generation.database.action` it is now ignored.

When checking for a user-provided setting before configuring the default setting, I think we need to consider `javax.persistence.schema-generation.database.action` as well as `hibernate.hbm2ddl.auto`.

snicoll:
Closing in favor of PR #25129

PR #25129, opened 2021-02-07
Checks 'javax.persistence.schema-generation.database.action' when determining DDL auto default

bono007:
Checks 'javax.persistence.schema-generation.database.action' before returning default ddlAuto in HibernateProperties.

Fixes gh-25054

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/DataSourceInitializedPublisher.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/DataSourceInitializedPublisher.java
@@ -128,7 +128,8 @@ class DataSourceInitializedPublisher implements BeanPostProcessor {
 				: "none");
 		Map<String, Object> hibernate = this.hibernateProperties.determineHibernateProperties(
 				this.jpaProperties.getProperties(), new HibernateSettings().ddlAuto(defaultDdlAuto));
-		return hibernate.containsKey("hibernate.hbm2ddl.auto");
+		return hibernate.containsKey("hibernate.hbm2ddl.auto") || !hibernate
+				.getOrDefault("javax.persistence.schema-generation.database.action", "none").equals("none");
 	}
 
 	/**
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
@@ -136,7 +136,18 @@ public class HibernateProperties {
 		}
 		String ddlDatabaseAction = existing.get(AvailableSettings.HBM2DDL_DATABASE_ACTION);
 		if (ddlDatabaseAction != null) {
-			return ddlDatabaseAction;
+			if (ddlDatabaseAction.equals("none")) {
+				return "none";
+			}
+			if (ddlDatabaseAction.equals("create")) {
+				return "create-only";
+			}
+			if (ddlDatabaseAction.equals("drop")) {
+				return "drop";
+			}
+			if (ddlDatabaseAction.equals("drop-and-create")) {
+				return "create";
+			}
 		}
 		return (this.ddlAuto != null) ? this.ddlAuto : defaultDdlAuto.get();
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
@@ -35,6 +35,7 @@ import org.springframework.util.StringUtils;
  * Configuration properties for Hibernate.
  *
  * @author Stephane Nicoll
+ * @author Chris Bono
  * @since 2.1.0
  * @see JpaProperties
  */
@@ -132,6 +133,10 @@ public class HibernateProperties {
 		String ddlAuto = existing.get(AvailableSettings.HBM2DDL_AUTO);
 		if (ddlAuto != null) {
 			return ddlAuto;
+		}
+		String ddlDatabaseAction = existing.get(AvailableSettings.HBM2DDL_DATABASE_ACTION);
+		if (ddlDatabaseAction != null) {
+			return ddlDatabaseAction;
 		}
 		return (this.ddlAuto != null) ? this.ddlAuto : defaultDdlAuto.get();
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
@@ -130,24 +130,12 @@ public class HibernateProperties {
 	}
 
 	private String determineDdlAuto(Map<String, String> existing, Supplier<String> defaultDdlAuto) {
+		if (existing.get(AvailableSettings.HBM2DDL_DATABASE_ACTION) != null) {
+			return null;
+		}
 		String ddlAuto = existing.get(AvailableSettings.HBM2DDL_AUTO);
 		if (ddlAuto != null) {
 			return ddlAuto;
-		}
-		String ddlDatabaseAction = existing.get(AvailableSettings.HBM2DDL_DATABASE_ACTION);
-		if (ddlDatabaseAction != null) {
-			if (ddlDatabaseAction.equals("none")) {
-				return "none";
-			}
-			if (ddlDatabaseAction.equals("create")) {
-				return "create-only";
-			}
-			if (ddlDatabaseAction.equals("drop")) {
-				return "drop";
-			}
-			if (ddlDatabaseAction.equals("drop-and-create")) {
-				return "create";
-			}
 		}
 		return (this.ddlAuto != null) ? this.ddlAuto : defaultDdlAuto.get();
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/orm/jpa/HibernatePropertiesTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/orm/jpa/HibernatePropertiesTests.java
@@ -19,15 +19,10 @@ package org.springframework.boot.autoconfigure.orm.jpa;
 import java.util.Map;
 import java.util.function.Consumer;
 import java.util.function.Supplier;
-import java.util.stream.Stream;
 
 import org.hibernate.cfg.AvailableSettings;
 import org.junit.jupiter.api.BeforeEach;
 import org.junit.jupiter.api.Test;
-import org.junit.jupiter.params.ParameterizedTest;
-import org.junit.jupiter.params.provider.Arguments;
-import org.junit.jupiter.params.provider.MethodSource;
-import org.junit.jupiter.params.provider.ValueSource;
 import org.mockito.Mock;
 import org.mockito.MockitoAnnotations;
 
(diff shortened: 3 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -133,6 +134,10 @@ private String determineDdlAuto(Map<String, String> existing, Supplier<String> d
 		if (ddlAuto != null) {
 			return ddlAuto;
 		}
+		String ddlDatabaseAction = existing.get(AvailableSettings.HBM2DDL_DATABASE_ACTION);
```
bono007, line 137: I did not see any docs to update that talk about this property.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -133,6 +134,10 @@ private String determineDdlAuto(Map<String, String> existing, Supplier<String> d
...
+		if (ddlDatabaseAction != null) {
+			return ddlDatabaseAction;
```
snicoll, line 139: The problem here is that `ddlDatabaseAction` is not, strictly speaking, a `ddlAuto`. The values are close but not the same. Considering that, I don't think that change is correct.

Rather, I think we should review this method altogether and set a ddl-auto only if necessary, calling the supplier if need to be. This could prove tricky as we've had a few regressions in this area.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/orm/jpa/HibernatePropertiesTests.java
```diff
@@ -135,6 +136,13 @@ void defaultDdlAutoIsNotInvokedIfHibernateSpecificPropertyIsSet() {
 				.run(assertDefaultDdlAutoNotInvoked("create"));
 	}
 
+	@Test
+	void defaultDdlAutoIsNotInvokedIfJpaDatabaseActionPropertyIsSet() {
+		this.contextRunner
+				.withPropertyValues("spring.jpa.properties.javax.persistence.schema-generation.database.action=create")
+				.run(assertDefaultDdlAutoNotInvoked("create"));
```
snicoll, line 143: I don't think this test is correct. The [spec](https://docs.oracle.com/javaee/7/tutorial/persistence-intro005.htm) states that `drop-and-create` is a valid value. If you'd rewrite this test with this, you'd have to assert that ddl auto was set with it, which is not a valid ddlAuto.

Rather, we should check the ddl auto isn't set (but the spec property is).

snicoll (changes requested):
Thanks for the PR. I don't think handling the spec value as a ddl auto value is legit. See further comments for details.

bono007:
Thanks for that info @snicoll . I will make the adjustments sometime w/in the next 24 hrs.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
bono007: @snicoll I found the mapping from JPA (`javax.persistence.schema-generation.database.action`) to HBM (`hibernate.hbm2ddl.auto`) [here](https://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#configurations-hbmddl). 

There is a direct mapping from each possible value of the JPA property to its corresponding HBM property. 

| JPA | HBM | Description | 
| --- | --- | ----------- | 
| none | none | No action will be performed |
| create | create-only | Database creation will be generated |
| drop | drop | Database dropping will be generated |
| drop-and-create | create | Database dropping will be generated followed by database creation |

The code in [CREATE_ONLY](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/tool/schema/Action.java#L29) and [CREATE](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/tool/schema/Action.java#L42) confirms this as well.

Since this ticket is about setting the HBM default value from the JPA value, I think the above grid gives us what we want to use. What do you think?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -133,6 +134,21 @@ private String determineDdlAuto(Map<String, String> existing, Supplier<String> d
 		if (ddlAuto != null) {
 			return ddlAuto;
 		}
+		String ddlDatabaseAction = existing.get(AvailableSettings.HBM2DDL_DATABASE_ACTION);
+		if (ddlDatabaseAction != null) {
```
bono007, line 138: ## Strict Approach
I chose a strict approach that only considers valid JPA database action values ([none, create, drop, drop-and-create](https://docs.jboss.org/hibernate/orm/5.2/userguide/html_single/Hibernate_User_Guide.html#configurations-hbmddl)) - each of which has a well-known counterpart `hbm2ddl.auto` value. Any other values, including valid  HBM specific values are ignored and the defaultDdlAuto provider kicks in.

## More Relaxed Approach 
However, Hibernate seems to be verify forgiving and relaxed when it comes to these values. It seems to use the JPA/HBM properties [interchangeable](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/tool/schema/spi/SchemaManagementToolCoordinator.java#L485). It prefers the JPA properties but also allows HBM values to be set on the JPA property.

A more relaxed approach would be to allow whatever values are set on the JPA property to be used as the default ddl auto and simply handle the 2 JPA special cases: `create -> create-only` and `drop-and-create -> create`. 


Anyways, let's continue the discussion.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/orm/jpa/HibernatePropertiesTests.java
```diff
@@ -135,9 +141,37 @@ void defaultDdlAutoIsNotInvokedIfHibernateSpecificPropertyIsSet() {
 				.run(assertDefaultDdlAutoNotInvoked("create"));
 	}
 
+	@ParameterizedTest
+	@MethodSource("validJpaDatabaseActionsToExpectedHbmValues")
+	void defaultDdlAutoIsNotInvokedIfJpaDatabaseActionPropertyIsSetToValidJpaValue(String jpaDatabaseAction,
```
bono007, line 146: Of course, these tests all reflect the "Strict Approach" outlined [above](https://github.com/spring-projects/spring-boot/pull/25129/files#r575549829).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
snicoll: > Since this ticket is about setting the HBM default value from the JPA value, I think the above grid gives us what we want to use. What do you think?

Unfortunately, I don't think this statement is accurate. IMO, this issue is about not setting a default ddlAuto value if the JPA spec property is set. I think translating one to the other is also a completely different feature that I am not convinced we want to provide.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
bono007: ### Misconception
Ahh ok, I misunderstood the issue. I think my misconception was that the ddl-auto property had to be set to something even if the db-action was set.

> Rather, I think we should review this method altogether and set a ddl-auto only if necessary, 

I interpreted this to mean set the ddl-auto property if the db-action was set to a value that was a valid ddl-auto

> When checking for a user-provided setting before configuring the default setting, I think we need to consider javax.persistence.schema-generation.database.action as well as hibernate.hbm2ddl.auto.

I interpreted "configuring the default setting" as setting the ddl-auto property. Which is true but IFF the user did not already specify a ddl-auto or db-action. If the user set it already under `spring.jpa.` then there is no need for us to do anything - it will flow down to HB8 and all is well. 

### Silver lining
I learned quite a bit about the 2 properties and how we use them and how HB8 uses them.

### Third time's a charm
Pivoting here shortly and doing the following:

1. If db-action property is set, do not set a default ddl-auto (we do not require db-auto to be set if db-action is set)
2. Adjust tests 

**Question 1)** 
I feel like it would be helpful to briefly mention this property in the docs. It's not mentioned anywhere but we do support it and looking into the HB8 code they seem to prefer that property over the legacy HB8 specific one. Thoughts?

**Question 2)**
I noticed that currently we [remove the ddl-auto](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java#L107) if the default came back as "none" - which could also be from the user setting it to "none". I understand this has no functional impact because the property default is "none". I am curious as to why? Should we do the same for the db-action property? 

My **mild** concern on removing the property when user has explicitly set it to none is this: 
* User sets ddl-auto to "none"
* We remove it prior to passing through to HB8
* User steps through HB8 code in debugger
* User says "I set ddl-auto, why is it not here?"

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -129,6 +130,9 @@ private void applyScanner(Map<String, Object> result) {
 	}
 
 	private String determineDdlAuto(Map<String, String> existing, Supplier<String> defaultDdlAuto) {
+        if (existing.get(AvailableSettings.HBM2DDL_DATABASE_ACTION) != null) {
```
bono007, line 133: I made the db-action property take precedence over the hbm property to be consistent w/ [Hibernate](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/tool/schema/spi/SchemaManagementToolCoordinator.java#L488).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -129,6 +130,9 @@ private void applyScanner(Map<String, Object> result) {
...
+            return null;
```
bono007, line 134: If the user specified db-action, respect that, do not set any ddl-auto default.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
bono007, line 133: I made the db-action property take precedence over the hbm property to be consistent w/ [Hibernate](https://github.com/hibernate/hibernate-orm/blob/master/hibernate-core/src/main/java/org/hibernate/tool/schema/spi/SchemaManagementToolCoordinator.java#L488).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/DataSourceInitializedPublisher.java
```diff
@@ -128,7 +128,8 @@ private boolean isInitializingDatabase(DataSource dataSource) {
 				: "none");
 		Map<String, Object> hibernate = this.hibernateProperties.determineHibernateProperties(
 				this.jpaProperties.getProperties(), new HibernateSettings().ddlAuto(defaultDdlAuto));
-		return hibernate.containsKey("hibernate.hbm2ddl.auto");
+		return hibernate.containsKey("hibernate.hbm2ddl.auto") || !hibernate
+				.getOrDefault("javax.persistence.schema-generation.database.action", "none").equals("none");
```
bono007, line 132: > I am curious as to why? Should we do the same for the db-action property?

I found out [what relies](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/DataSourceInitializedPublisher.java#L132) on the removed key. I have added the db-action property to this check to ensure we have consistent behavior when the user is setting only the db-action property to drive db schema creation. I also added tests to verify the cases.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateProperties.java
```diff
@@ -35,6 +35,7 @@
  * Configuration properties for Hibernate.
  *
  * @author Stephane Nicoll
+ * @author Chris Bono
```
bono007, line 38: Do I need to add myself as author here when its 2 lines of code? I mean, it does add change in behavior so I can see either way. I can't remember the threshold when to add to author tag.

snicoll:
Thanks @bono007. I've polished the PR in c2f452a. One thing to note is that the Hibernate property should take precedence over the JPA one (as noted by @wilkinsona in the original issue) and the proposal did the opposite. I've also reworked the tests to focus on the vendor properties as it felt more focused and the event doesn't exist on `master` anymore.

PR #25150, opened 2021-02-09
initQueryTimeout and ildeTimeout defaults are not aligned with Cassandra defaults

bono007:
See #25130 for background.

@snicoll I went ahead and added the defaults test in this PR as I needed to update it to include the default and also use it to verify the changes. 

If however, you want the 2.3.x changes to merge into here (purist tracking standpoint) just lmk and I will remove the test from this PR.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraProperties.java
@@ -270,7 +270,7 @@ public class CassandraProperties {
 		 * Timeout to use for internal queries that run as part of the initialization
 		 * process, just after a connection is opened.
 		 */
-		private Duration initQueryTimeout = Duration.ofMillis(500);
+		private Duration initQueryTimeout = Duration.ofSeconds(5);
 
 		public Duration getConnectTimeout() {
 			return this.connectTimeout;
@@ -360,7 +360,7 @@ public class CassandraProperties {
 		/**
 		 * Idle timeout before an idle connection is removed.
 		 */
-		private Duration idleTimeout = Duration.ofSeconds(120);
+		private Duration idleTimeout = Duration.ofSeconds(5);
 
 		/**
 		 * Heartbeat interval after which a message is sent on an idle connection to make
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraProperties.java
```diff
@@ -270,7 +270,7 @@ public Request getRequest() {
 		 * Timeout to use for internal queries that run as part of the initialization
 		 * process, just after a connection is opened.
 		 */
-		private Duration initQueryTimeout = Duration.ofMillis(500);
+		private Duration initQueryTimeout = Duration.ofSeconds(5);
```
bono007, line 273: Taken from https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/api/core/config/OptionsMap.java#L266

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraProperties.java
```diff
@@ -360,7 +360,7 @@ public Throttler getThrottler() {
 		/**
 		 * Idle timeout before an idle connection is removed.
 		 */
-		private Duration idleTimeout = Duration.ofSeconds(120);
+		private Duration idleTimeout = Duration.ofSeconds(5);
```
bono007, line 363: Taken from https://github.com/datastax/java-driver/blob/4.x/core/src/main/java/com/datastax/oss/driver/api/core/config/OptionsMap.java#L339

snicoll:
> I went ahead and added the defaults test in this PR as I needed to update it to include the default and also use it to verify the changes.

That's not what I was expecting. If testing the defaults and fixing the defaults are two separate PRs, including the former in the latter makes the separate PR completely irrelevant. I have reverted.

snicoll:
Sigh. One of those two is also wrong in `2.3.x`. I've created #25165

bono007:
> Sigh. One of those two is also wrong in `2.3.x`. I've created #25165

Yeh the value changed in 4.8.1 (used by Boot 24) but was still 500ms in 4.6 driver (used by Boot 23).

bono007:
> > I went ahead and added the defaults test in this PR as I needed to update it to include the default and also use it to verify the changes.
> 
> That's not what I was expecting. If testing the defaults and fixing the defaults are two separate PRs, including the former in the latter makes the separate PR completely irrelevant. I have reverted.
> 
> My understanding is that those two properties are tested in the `2.3.x` version of the PR (and that passes because the values haven't been updated yet). By merging this PR ahead, we make sure the first PR you've raised can merge nicely.

Sorry for the confusion @snicoll .  It passed in 2.3 branch because the value changed in 4.8.1 (used by Boot 24) but was still 500ms in 4.6 driver (used by Boot 23). 

In hindsight, had I figured that out sooner it would have probably made the most sense to do all of the work in 2.4 rather than 2.3 and partially here. Thanks for your patience.

Issue #25127, opened 2021-02-07
Incorrect detection of MariaDB


lbruun:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before, or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Problem
Detection of MariaDB in [org.springframework.boot.jdbc.DataBaseDriver](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java#L73) is incorrect.  The MariaDB driver returns [`MariaDB` as the `productName`](https://github.com/mariadb-corporation/mariadb-connector-j/blob/9536ba7afbb29226be7f89724178f4e548ad01be/src/main/java/org/mariadb/jdbc/MariaDbDatabaseMetaData.java#L1221) but we are using `MySQL` as the `productName`.

## Solution
Update the [`DatabaseDriver`](https://github.com/spring-projects/spring-boot/blob/301e80be8c4da73dd734d93f915088de4997c88c/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java) enum with the right value for the `productName` for `MariaDB`.

Update [the tests](https://github.com/spring-projects/spring-boot/blob/301e80be8c4da73dd734d93f915088de4997c88c/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/jdbc/DatabaseDriverTests.java) to ensure that the right `DatabaseDriver.fromProductName()` works for `MariaDB`.

## Steps to Fix 
 
- [ ] Claim this issue with a comment below and ask any clarifying questions you need
- [ ] Set up a repository locally following the [Contributing Guidelines](https://github.com/spring-projects/spring-boot/blob/master/CONTRIBUTING.adoc)
- [ ] Try to fix the issue following the steps above
- [ ] Commit your changes and start a pull request.

anudeepg:
I am interested to contribute, Can I claim this issue ?

scottfrederick:
[...]

PR #25173, opened 2021-02-10
Fix MariaDB driver product name

anudeepg:
Fix MariaDB driver product name issue : https://github.com/spring-projects/spring-boot/issues/25127


snicoll:
@anudeepg thank you for making your first contribution to Spring Boot.

This issue is noteworthy as a MariaDB backend previously returned a `mysql`  platform id. Those who are using this identifier for database initialization purposes can adapt their scripts accordingly (or translate that back to mysql if need to be).

PR #25264, opened 2021-02-13
Disable Gradle's welcome message in gradle.properties

dreis2211:
Hi,

I'm currently preparing the JDK 16 pipeline and noticed that the `org.gradle.internal.launcher.welcomeMessageEnabled` property could be disabled via `gradle.properties`. That makes things a little easier to read in the build scripts if you ask me. This will especially help because I'm likely complicating them a bit with the pipeline adjustments.

Obviously, this will also disable the welcome message for local builds, but I think that's acceptable given that we should be probably aware of the new features before upgrading.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/ci/scripts/build-pr-project.sh
+++ b/ci/scripts/build-pr-project.sh
@@ -4,5 +4,5 @@ set -e
 source $(dirname $0)/common.sh
 
 pushd git-repo > /dev/null
-./gradlew -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false --no-daemon --max-workers=4 --continue build
+./gradlew --no-daemon --max-workers=4 --continue build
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks for the suggestion, @dreis2211. I'd prefer to keep the message enabled for local builds. I think it provides useful information for contributors who aren't familiar with the version of Gradle that we're using.

dreis2211:
That's fair, although from personal experience I didn't get much from the welcome message as a contributor. After all it only shows you the last features which most of the time are not directly adopted or irrelevant.

But I'm also highly biased here as a regular contributor and someone who's reading Gradle release notes anyhow, so don't worry ;-)

Issue #25055, opened 2021-01-29
Migrations performed by Flyway or Liquibase may not have completed before the database is accessed via jOOQ


wilkinsona:
If jOOQ is used to interact with a database during application context refresh, migrations performed by Flyway or Liquibase may not have completed. This can happen because there's no guarantee that the Flyway or Liquibase beans will have been created and initialised before any jOOQ beans are created and used. I think we need some `AbstractDependsOnBeanFactoryPostProcessor` beans so that any jOOQ `DSLContext` beans depend on the Flyway and Liquibase beans.

lukaseder:
Interesting, could this be a more general problem? Would a migration-aware `DataSource` (or connection pool) be interesting, with a mutex that prevents migrations and ordinary access from happening at the same time?

wilkinsona:
It is a general problem and we have a (fairly) general solution for it. There's a bug here as we've neglected to use that solution for jOOQ's `DSLContext`. I noticed this as part of a broader reworking of DataSource initialization for Boot 2.5. Part of that may result in a further generalisation of the current solution so that it's more extensible.

>  a mutex that prevents migrations and ordinary access from happening at the same time

Other than one scenario involved deferred bootstrapping of JPA, everything is single threaded. We need to ensure that things happen in the right order rather than preventing them from happening in parallel.

lukaseder:
I see, thanks for the explanation!

wilkinsona:
Closing in favour of #25279.

neiser:
[...]

PR #25279, opened 2021-02-15
Migrations performed by Flyway or Liquibase may not have completed before the database is accessed via jOOQ

eddumelendez:
In order to make sure that `Jooq` is used when `flyway` and
`liquibase` beans are created and migrations are done,
`DslContextDependsOnPostProcessor` was added.

See gh-25055


wilkinsona:

Issue #25121, opened 2021-02-05
Clarify precedence between yml and properties files on the same level


leonard84:
In [4.2. Externalized Configuration](https://docs.spring.io/spring-boot/docs/2.4.2/reference/htmlsingle/#boot-features-external-config) (Spring Boot 2.4.2) it states that 

> Config data files are considered in the following order:
> 1. Application properties packaged inside your jar (application.properties and YAML variants).
> 2. Profile-specific application properties packaged inside your jar (application-{profile}.properties and YAML variants).
> 3. Application properties outside of your packaged jar (application.properties and YAML variants).
> 4. Profile-specific application properties outside of your packaged jar (application-{profile}.properties and YAML variants).

However, what is unclear is which config values wins if you have both `application.yml` and `application.properties` on the same level. We'd like to use **Variant 1** (see below), which seems to work, but relies on undocumented behavior that could easily change, so it would be great if the precedence handling could be added to the docs and checked with tests, if that is not already the case.

### In SB <2.4 we used:
* `src/main/application.yml` to store profiles
* `src/main/application-default.properties` to set `spring.profiles.include=stage-workstation` to activate a profile if nothing was activated via ENV/systemproperty. (We used the properties file as this key is invalid in yml, but using this file makes it also easy to override for test)
*  `src/test/application-default.properties` to set `spring.profiles.include=stage-workstation, test` to activate the test profile
* `src/main/application-test.yml` with profile in the name to avoid overriding the `application.yml`

### For SB 2.4.x
We are evaluating two solutions

#### Variant 1
* `src/main/application.yml` to store profiles and use `spring.profiles.group.default=stage-workstation`
* `src/test/application.properties` to set `spring.profiles.group.default=stage-workstation,test` to activate the test profile and use properties instead of yml to avoid overriding the `application.yml`
* `src/main/application-test.yml` with profile in the name to avoid overriding the `application.yml`

#### Variant 2
* `src/main/application.yml` to store profiles and use `spring.profiles.group.default=stage-workstation`
* `src/test/application.yml` that uses `spring.config.import=file:${user.dir}/src/main/resources/application.yml` and overrides the `spring.profiles.group.default=stage-workstation,test` (IMHO it is rather brittle to rely on `${user.dir}` to be set to the correct location)

philwebb:
[...]

PR #25300, opened 2021-02-16
Add note concerning application yml and properties file precedence

robert-smith-911:


snicoll:
Thanks for the PR. It looks like a proposal for #25121

robert-smith-911:
@snicoll You're welcome, yes that's correct.

snicoll:
Thanks again @robert-smith-911

Issue #25053, opened 2021-01-29
Group JDBC-related batch properties beneath spring.batch.jdbc


benas:
As of v2.4.2, JDBC-related properties for batch are under `spring.batch`, unlike quartz, session and integration where those properties are under `spring.xxx.jdbc`.

After discussing this with @wilkinsona , I understand that this is because JDBC is currently the only store option for batch. However, I think it would be better to group those properties beneath `spring.batch.jdbc` for 1) naming consistency and 2) to pave the way for other storage options for batch in the future (We have indeed a feature request to support a non-jdbc job repository implementation, see https://github.com/spring-projects/spring-batch/issues/877 which we might consider in the next major release).

chaundhyan:
@benas, @wilkinsona 
If allowed, may I take this enhancement ahead, please?
If you guys agree, please assign this to me.

Thanks

wilkinsona:
Thanks for the offer, @chaundhyan. That would be great. We'll need to deprecate the existing properties (using `@DeprecatedConfigurationProperty` and then provide replacements in the new `spring.batch.jdbc.` location. Please let us know if you have any questions.

chaundhyan:
@benas @wilkinsona : I've had done the changes. Would you like to review it over my repo?
[My Change Comparison](https://github.com/spring-projects/spring-boot/compare/master...chaundhyan:master)

Shall i raise the PR, if yes then into which branch? 
I've had taken a fork and pushed changes into main branch over my forked repo.

If any suggestion, please do let me know.

Also there exists 3 test cases which are failing while doing `./gradlew build`. But they are passing individually.
JettyReactiveWebServerFactoryTests. whenServerIsShuttingDownGracefullyThenNewConnectionsCannotBeMade()
              org.springframework.web.reactive.function.client.WebClient was not fulfilled within 30 seconds.

Same way other are failing.
NettyReactiveWebServerFactoryTests. whenServerIsShuttingDownGracefullyThenNewConnectionsCannotBeMade()
TomcatReactiveWebServerFactoryTests. whenServerIsShuttingDownGracefullyThenNewConnectionsCannotBeMade()

Please suggest.

Thanks
Mukul Kumar Chaundhyan

wilkinsona:
[...]

PR #25316, opened 2021-02-16
Group JDBC-related batch properties beneath spring.batch.jdbc

chaundhyan:
This is an enhancement mentioned as #25053. Deprecating the existing batch jdbc properties and moving those under spring.batch.jdbc.


snicoll:
@chaundhyan thank you for making your first contribution to Spring Boot.

PR #25321, opened 2021-02-17
Refactor CacheConfigurations to avoid storing configuration classes

sdeleuze:
This commit refactors `CacheConfigurations` implementation to make it more native friendly by storing strings instead of classes in
order to avoid loading the configuration classes when `CacheConfigurations` is initialized at build time.


sdeleuze:
As discussed with @snicoll there are various pending design questions about caching support in Spring Native, so probably better to wait before merging this PR. Feel free to close it for now or maybe to tag it accordingly.

snicoll:
Thanks for the PR and the discussion!

Given the contract of the public API that change looks good to me but if you need this for `2.4.x` it is a bit unusual to backport something like this. Let's work together on a solution and make change in SB accordingly. If it turns out that this way is enough, we can reopen the PR and merge it.

snicoll:
Looking again on the code, I've changed my mind as this makes more sense to store the FQN since anyway the API is working on that and not the class.

PR #25353, opened 2021-02-18
Polish HTTP client metrics documentation wording

shakuzen:
The current wording sounded odd to me, so I'm proposing wording I thought was more natural.


snicoll:
Good catch, thanks Tommy!

Issue #1303, opened 2014-07-29
When using tomcat-jdbc in its default configuration, DataSourceHealthIndicator doesn't recover when MySQL is restarted


wilkinsona:
Start MySQL and check the health:

```
{"status":"UP","database":"MySQL","hello":1}
```

Stop MySQL and check health again:

```
{
"database": "MySQL", "error": "org.springframework.dao.RecoverableDataAccessException: StatementCallback; SQL [SELECT 1]; Communications link failure\n\nThe last packet successfully received from the server was 47,301 milliseconds ago.  The last packet sent successfully to the server was 6 milliseconds ago.; nested exception is com.mysql.jdbc.exceptions.jdbc4.CommunicationsException: Communications link failure\n\nThe last packet successfully received from the server was 47,301 milliseconds ago.  The last packet sent successfully to the server was 6 milliseconds ago.", "status": "DOWN"
}
```

Restart MySQL and check health:

```
{"status":"DOWN","error":"org.springframework.dao.DataAccessResourceFailureException: ConnectionCallback; SQL []; No operations allowed after connection closed.; nested exception is com.mysql.jdbc.exceptions.jdbc4.MySQLNonTransientConnectionException: No operations allowed after connection closed."}
```

It looks like it's not letting go of the broken connection. This was with MySQL 5.6.19 and using the `tomcat-jdbc` connection pool with minimal configuration, i.e. just `spring.datasource.url`.

philwebb:
I think we should probably push this back to 1.3. @wilkinsona feel free to move back to 1.2.x if you disagree.

wilkinsona:
Rather than focussing on the health indicator, we've decided that we should be more opinionated but the defaults for the auto-configured data source. `tomcat-jdbc` should, be default, be configured to call `Connection.isValid`. This will align its default behaviour with Hikari.

snicoll:
This one is a dup of #4906

snicoll:
I just tried with `1.4.0.BUILD-SNAPSHOT` and this is indeed fixed thanks to #4906

Issue #1111, opened 2014-06-17
Problems using 1.1.1 together with Hibernate Search 4.5.1.Final


michael-simons:
Hello,

i'm trying to upgrade another project from Boot 1.0.2 to 1.1.1.

I've manually configured datasources by extending TomcatDataSourceConfiguration (several reasons, one was that i have multiple to different databases). This is working fine with Boot 1.0.2.

This class is gone and i'm trying to use "DataSourceBuilder" as stated in the docs:

``` java
@Bean(name = "leiDatasource")     
@ConfigurationProperties(prefix = "eemc.lei.datasource") 
public DataSource dataSource() { 
    return DataSourceBuilder.create().build();
}
```

This works.

The JPA entities in this data source are index using hibernate search. This is on the class path together with all dependencies, especially solr-core.

Using the builder Hibernate Search isn't able to retrieve his dependencies:

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'leiEntityManagerFactory' defined in class path resource [de/enerko/eemc/config/LEIDatabasePersistenceConfig.class]: Invocation of init method failed; nested exception is java.lang.NoClassDefFoundError: org/apache/solr/common/ResourceLoader
    at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1553)
    at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:539)
    at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:475)
    at org.springframework.beans.factory.support.AbstractBeanFactory$1.getObject(AbstractBeanFactory.java:304)
    at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:228)
    at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:300)
    at org.springframework.beans.factory.support.AbstractBeanFactory.getBean(AbstractBeanFactory.java:195)
    at org.springframework.context.support.AbstractApplicationContext.getBean(AbstractApplicationContext.java:973)
    at org.springframework.context.support.AbstractApplicationContext.finishBeanFactoryInitialization(AbstractApplicationContext.java:750)
[...]

Issue #15732, opened 2019-01-17
Provide support for auto-configuring multiple datasources


wilkinsona:
Auto-Configuration of a single DataSource works well for the vast majority of users, but when a subsequent DataSource is required things get harder (see #7652 for one example) as all of the data sources then need to be manually configured. We'd like to make things easier by providing support for auto-configuring multiple DataSources. In terms of properties, this could look something like this:

```
spring.datasource.primary.url=…
spring.datasource.primary.username=…
spring.datasource.primary.password=…

spring.datasource.secondary.url=…
spring.datasource.secondary.username=…
spring.datasource.secondary.password=…
```

Some more design work is needed, but `primary` could be used as a "special" name that results in the auto-configured bean being marked as `@Primary`. We'd also need similar functionality for components that consume a `DataSource` such as JPA and transaction management.

mbhave:
We might also need `@AutoConfigureTestConfigure#replace` to support replacing all auto-configured databases.

OrangeDog:
With such a property design, I would like the current `spring.datasource` properties, including vendor-specific ones, to become the defaults for the other ones.

OrangeDog:
Re. transaction management. If you've got multiple sources then you'll probably want JTA. However, as I understand the default datasources that Boot constructs will not work because they use HikariCP, which isn't compatible.

membersound:
Please note that `spring.datasource.hikari.data-source-properties.*` should then also automatically be applied to any secondary auto-configured datasource. As it's likely the user wants the same config for all databases in the application.

wilkinsona:
We should keep `AbstractRoutingDataSource` in mind while working on this. It may be useful to group together two or more auto-configured DataSources into an `AbstractRoutingDataSource`, and perhaps not even expose the underlying DataSources as beans.

We should also ensure that whatever we comes up with isn't just applicable to DataSources. #25369 raised the possibility of configuring multiple RabbitMQ connection factories and we need something that's consistent across all sorts of different data stores, message brokers, etc.

muhmud:
Completely concur with that last point, I have just recently had the same requirement for kafka.

OrangeDog:
If `AbstractRoutingDataSource` is being looked at, is there any possibility of rearranging things so it can see the target transaction properties? I did have a look at routing based on read-only status, but the routing happens too early and searching the call stack for the `@Transactional` was too messy.

wilkinsona:
[...]

PR #25369, opened 2021-02-19
Introduced auto-configuration for MultiRabbit

rwanderc:
Hi Spring Team!

This feature introduces configuration-based initialization for multiple Rabbit brokers:
- introduces the AutoConfiguration for **MultiRabbit** (contributed to `spring-amqp` [#1111](https://github.com/spring-projects/spring-amqp/pull/1111) & [#1303](https://github.com/spring-projects/spring-amqp/pull/1303) and available at version 2.3.5);
- and provides documentation.

Javadocs are mentioning 2.4. I wonder if that's possible to be available in that version, but I will update it if eventually that's not the release version for this feature.

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/context/properties/DocumentConfigurationProperties.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/context/properties/DocumentConfigurationProperties.java
@@ -92,7 +92,8 @@ public class DocumentConfigurationProperties extends DefaultTask {
 						"Hikari specific settings bound to an instance of Hikari's HikariDataSource")
 				.addSection("transaction").withKeyPrefixes("spring.jta", "spring.transaction").addSection("integration")
 				.withKeyPrefixes("spring.activemq", "spring.artemis", "spring.batch", "spring.integration",
-						"spring.jms", "spring.kafka", "spring.rabbitmq", "spring.hazelcast", "spring.webservices")
+						"spring.jms", "spring.kafka", "spring.rabbitmq", "spring.multirabbitmq", "spring.hazelcast",
+						"spring.webservices")
 				.addSection("actuator").withKeyPrefixes("management").addSection("devtools")
 				.withKeyPrefixes("spring.devtools").addSection("testing").withKeyPrefixes("spring.test");
 		DocumentOptions options = builder.build();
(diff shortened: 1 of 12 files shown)


wilkinsona:
Thanks for the proposal. We have an [open issue](https://github.com/spring-projects/spring-boot/issues/15732) for auto-configuring multiple DataSources that is related to this. Unfortunately, I don't think we can add support for auto-configuring connections to multiple RabbitMQ brokers without considering how it will look for DataSources and other components.

You've taken a routing-based approach here that is akin to Spring Framework's `AbstractRoutingDataSource`. We also need to support the use case where users want multiple `DataSource` or `ConnectionFactory` beans in the context and things will need to be done in a way where one does not clash with the other.

On the DataSource side, we've seen more demand for auto-configuring multiple DataSources rather than a single AbstractRoutingDataSource and I suspect that will be the same for other backend services, including Rabbit brokers. In short, I think we'll need to do the design work for 15732 before we can consider this.

rwanderc:
Thank you @wilkinsona for the quick answer!
Regarding the intention of having a matching design between multiple DataSources and ConnectionFactories, I can clearly see the benefit.
But on the other hand, if that takes some more time (this issue #15732 is already 2 years old), blocking this feature might be less beneficial than updating the implementation later on.
Don't you think it's worth it to carry on with the routing-based approach and update it when the decision is made?

wilkinsona:
No, I'm afraid not.

If it were only a matter of updating implementation details then it wouldn't be a problem, but I think it's almost certain that things would change from a user's perspective, and probably change considerably. That would mean that we're unable to do anything until Spring Boot 3.0 as the breaking changes would be too great to include in a minor release.

The benefit of adding multi-Rabbit auto-configuration now is that it makes it easier to do something that's already possible. The cost of adding multi-Rabbit auto-configuration now is that it may make it impossible to make some other enhancements until 3.0. Unfortunately, that cost outweighs the benefit.

Thanks very much for the pull request, but I think it's best that we close this one as we won't be able to merge it in its current form. As part of tackling #15732 we can consider how things would look for other backends including RabbitMQ. Right now I'm imagining properties for configuring multiple DataSources, ConnectionFactories, etc and with an additional property to group two or more of those into a routing DataSource or ConnectionFactory. As such, everything would be configured using `spring.datasource.*` or `spring.rabbitmq` rather than introduction completely separate groups of properties.

Issue #23726, opened 2020-10-16
LiveReload doesn't work without restart enabled


cdalexndr:
Using `spring.devtools.restart.enabled=false` disables livereload even if `spring.devtools.livereload.enabled=true` .
Browser connects successfully to LiveReload, but changing files does not trigger reload.

Spring Boot 2.3.3.RELEASE

philwebb:
This is expected, the live reload feature need something to trigger it and means that restart needs to be enabled.

cdalexndr:
A warning should be added to [documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/using-spring-boot.html#using-boot-devtools-livereload)

wilkinsona:
We'd be happy to review a pull request if you'd like to propose something, @cdalexndr.

cdalexndr:
What project contains spring boot documentation? Is there any guide to update documentation?

wilkinsona:
The reference documentation is all in this repository in the `spring-boot-docs` project. You can find the source for the existing docs on LiveReload here: https://github.com/spring-projects/spring-boot/blob/3f1f830efee820d04460ce6a5ca88d3f130f3b5d/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using-spring-boot.adoc#livereload. GitHub's built-in editor for Asciidoc source is quite good so you may want to use the Edit button. (top-right of the page) to edit the documentation in your browser. Alternatively, you can clone the code and change it in an editor of your choice on your machine.

PR #25374, opened 2021-02-19
Docs contribution section

cdalexndr:

### DIFF
--- a/CONTRIBUTING.adoc
+++ b/CONTRIBUTING.adoc
@@ -31,7 +31,12 @@ Before we accept a non-trivial patch or pull request we will need you to https:/
 Signing the contributor's agreement does not grant anyone commit rights to the main repository, but it does mean that we can accept your contributions, and you will get an author credit if we do.
 Active contributors might be asked to join the core team, and given the ability to merge pull requests.
 
+== Documentation contributions
 
+The documentation is stored in this repository, in the link:https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-docs/src/docs/asciidoc[spring-boot-docs] project.
+
+You may use GitHub's built-in editor for Asciidoc by pressing the top-right Edit button to edit the documentation in browser.
+Alternatively, you can clone and change it in an editor of your choice on your machine.
 
 == Code Conventions and Housekeeping
 None of these is essential for a pull request, but they will all help.  They can also be


CONTRIBUTING.adoc
```diff
@@ -31,7 +31,12 @@ Before we accept a non-trivial patch or pull request we will need you to https:/
 Signing the contributor's agreement does not grant anyone commit rights to the main repository, but it does mean that we can accept your contributions, and you will get an author credit if we do.
 Active contributors might be asked to join the core team, and given the ability to merge pull requests.
 
+== Documentation contributions
 
+The documentation is stored in this repository, in the link:https://github.com/spring-projects/spring-boot/tree/master/spring-boot-project/spring-boot-docs/src/docs/asciidoc[spring-boot-docs] project.
+
+You may use GitHub's built-in editor for Asciidoc by pressing the top-right Edit button to edit the documentation in browser.
```
snicoll, line 38: I am not sure we should bother describing the various ways changes can be contributed on GitHub. As a matter of a fact, I prefer contributions that originate from "real branches" as it makes the follow-up easier. We expect contributors to build the documentation and check that the output is valid which you cant' do using the web editor.

CONTRIBUTING.adoc
```diff
@@ -31,7 +31,12 @@ Before we accept a non-trivial patch or pull request we will need you to https:/
...
+Alternatively, you can clone and change it in an editor of your choice on your machine.
```
snicoll, line 39: I am not sure what that sentence brings to this section.

snicoll (changes requested):
Thanks for the PR. Can you share what made you create a dedicated section for the documentation? I'd rather keep a general contribution section (perhaps improving it?) rather than creating a dedicated section like this.

cdalexndr:
Contribution section needs a guide to editing docs (https://github.com/spring-projects/spring-boot/issues/23726#issuecomment-712205289), to answer questions like:

1. Where are the docs stored?
2. What format do the docs use?
3. How can I edit docs?

CONTRIBUTING.adoc
cdalexndr: Aren't there automatic checks (doc build) on a PR request even when using the github's web editor?

snicoll:
I saw that but I am not sure why we should create a dedicated section for this. Flagging for team attention to get some more feedback.

CONTRIBUTING.adoc
snicoll: A contributor is expected to validate their changes before submitting a PR and I am not keen to let contributors use our CI infrastructure to validate their changes.

Part of that is because if you keep this trial and error until you push something that works, you'd trigger a notification for each commit attempt to the 3K people watching this repository.

wilkinsona:
To me, the documentation is part of the project's code. I think a subsection in the "Working with the Code" section, describing where the documentation lives and noting that it's written in Asciidoc, would be a good addition.

Issue #25366, opened 2021-02-18
Reset spies on separate test classes using same context


cdalexndr:
Using `@TestExecutionListeners(MockitoTestExecutionListener.class)`

Two separate test classes should be independent, even when reusing the same cached app context.
Stubbing inside a test class should not remain in the other. Same for invocation counts.

wilkinsona:
What are you using `MockitoTestExecutionListener` for? As mentioned in its javadoc its intended for use with `@MockBean`. `@MockBean` will automatically reset the mock after each test.

cdalexndr:
I'm using it for `@SpyBean`

wilkinsona:
`@SpyBean` should have the same automatic reset behaviour, tunable via its `reset` attribute. If your `@SpyBean` spies aren't being reset, then something's not behaving as expected. If you would like us to spend some more time investigating, please spend some time providing a complete yet minimal sample that reproduces the problem. You can share it with us by pushing it to a separate repository on GitHub or by zipping it up and attaching it to this issue.

cdalexndr:
Test sample repo: https://github.com/cdalexndr/spring-boot-issue-25366

wilkinsona:
Thanks. The default behaviour of `@TestExecutionListeners` is to replace the default execution listeners which will mean that you're losing `ResetMocksTestExecutionListener`. I think it'll work if you either specify `ResetMocksTestExecutionListener` or configure the merge mode to be `MERGE_WITH_DEFAULTS`.

cdalexndr:
Indeed, using `@TestExecutionListeners({MockitoTestExecutionListener.class, ResetMocksTestExecutionListener.class})` tests succeed.

But shouldn't `ResetMocksTestExecutionListener` be part of `MockitoTestExecutionListener`? Why are they separate?

This separation allows invalid behavior, because `@Spybean` is enabled by `MockitoTestExecutionListener`, but only **partially**, because it's `reset` field won't work unless using `ResetMocksTestExecutionListener`.
So it seems natural to bundle them together, and using only one instance lowers occurrences of bugs due to missing test listener.

Note that with `testng` I'm required to use `@TestExecutionLisetners` annotation. I guess on JUnit, the listeners are automatically added.

wilkinsona:
[...]

PR #25375, opened 2021-02-19
Include ResetMocksTestExecutionListener

cdalexndr:
ResetMocksTestExecutionListener is required to enable `@SpyBean/@MockBean`'s `reset()` attribute.

https://github.com/spring-projects/spring-boot/issues/25366

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
@@ -6656,7 +6656,7 @@ To use this feature with a different arrangement, a listener must be explicitly 
 
 [source,java,pending-extract=true,indent=0]
 ----
-	@TestExecutionListeners(MockitoTestExecutionListener.class)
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/spring-boot-features.adoc
```diff
@@ -6656,7 +6656,7 @@ To use this feature with a different arrangement, a listener must be explicitly
```
snicoll, line 6656: The section above states "a listener" so this should be updated accordingly.

snicoll (changes requested):
Thanks for the PR. I've added a a note, can you please look at it?

wilkinsona:
Thanks very much, @cdalexndr. In addition to your updates to the reference documentation, I've also [made some additions to the javadoc of `MockitoTestExecutionListener` and `ResetMocksTestExecutionListener`](https://github.com/spring-projects/spring-boot/commit/22b73f3020c5089f2653a4616f0709c59b52d9eb).

PR #25382, opened 2021-02-21
Fix wording error in build-info goal description

mustafau:


snicoll:
@mustafau good catch, thank you again for your contribution.

PR #25409, opened 2021-02-24
Enable stricter plugin validation for the Gradle plugin

wolfs:
The validate plugin task can be configured to do even stricter validation. The Spring boot plugin currently doesn't have any problems which make the stricter validation fail, and it makes sense to enable this validation so no problems are introduced.

This PR also simplifies the build script a little bit to avoid copying around classes only to exclude one class from plugin validation.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/build.gradle
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/build.gradle
@@ -65,6 +65,7 @@ task preparePluginValidationClasses(type: Copy) {
 
 validatePlugins {
 	classes.setFrom preparePluginValidationClasses
+	enableStricterValidation = true
 }
 
 task dependencyVersions(type: org.springframework.boot.build.constraints.ExtractVersionConstraints) {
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, @wolfs. Something seems to be going awry with the change to how the classes are filtered. The pull request failed to build on CI due to it not being able to find one of the plugin's classes:

```
Execution failed for task ':spring-boot-project:spring-boot-tools:spring-boot-gradle-plugin:validatePlugins'.
> There was a failure while executing work items
   > A failure occurred while executing org.gradle.plugin.devel.tasks.internal.ValidateAction
      > Could not load class: org.springframework.boot.gradle.plugin.SpringBootPlugin
```

I see a similar failure locally:

```
Execution failed for task ':spring-boot-project:spring-boot-tools:spring-boot-gradle-plugin:validatePlugins'.
> There was a failure while executing work items
   > A failure occurred while executing org.gradle.plugin.devel.tasks.internal.ValidateAction
      > Could not load class: org.springframework.boot.gradle.tasks.bundling.LayeredSpec$DependenciesSpec
```

I think this is happening because the change to the filtering is putting each individual class file on the classpath rather than a single directory that contains all of the classes in their packages. Is there a way to provide a single "virtual" root directory with filtered content or do we need to keep the `preparePluginValidationClasses` task?

wolfs:
> I think this is happening because the change to the filtering is putting each individual class file on the classpath

I think you are right :( I'll remove this commit from the PR.

wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @wolfs.

Issue #25410, opened 2021-02-24
Investigate flaky tests


dreis2211:
Hi,

both locally and on CI I encounter relatively frequent flaky tests. While I see most of the time that they're flaky and ignore them, it happens often enough that I spent/waste time on identifying if those failures are caused by my changes.

Probably not a complete list of things, but things I noticed lately:
```
ReactiveElasticsearchRepositoriesAutoConfigurationTests > doesNotTriggerDefaultRepositoryDetectionIfCustomized()
ReactiveElasticsearchRepositoriesAutoConfigurationTests > testDefaultRepositoryConfiguration()
DataCassandraTestIntegrationTests > didNotInjectExampleService()
Jetty10ServletWebServerFactoryTests > whenServerIsShuttingDownGracefullyThenResponseToRequestOnIdleConnectionWillHaveAConnectionCloseHeader()
CouchbaseAutoConfigurationIntegrationTests > defaultConfiguration()
```

Subjectively, the JDK 15 pipeline is a bit flakier, but that might be a false lead.

Anyhow - I wonder if we can do anything about those. I remember that you did an awesome job of increasing timeouts here and there already and tweaked the testcontainer startup attempts, but I think we're past the testcontainers stage in most of the cases mentioned above.

Cheers,
Christoph

dreis2211:
Maybe a stupid question, but is there the possibility to "grep" over the failed build scans on ge.spring.io to get a more complete list of flaky tests?

wilkinsona:
There is indeed and it's really useful. [Here](https://ge.spring.io/scans/tests?search.relativeStartTime=P7D&search.rootProjectNames=spring-boot-build&search.timeZoneId=Europe/London&tests.sortField=FLAKY&tests.unstableOnly=true) are all the test failures over the last 7 days sorted with flaky tests first.

I thought I'd stabilized the Cassandra tests with a timeout increase, but one [failed again today with a 10s timeout](https://ge.spring.io/s/53c6qrchhmhny/tests/:spring-boot-project:spring-boot-test-autoconfigure:test/org.springframework.boot.test.autoconfigure.data.cassandra.DataCassandraTestIntegrationTests/didNotInjectExampleService()#1).

dreis2211:
Oh, that is a lovely - I was hoping that Gradle Enterprise had such a feature. Thanks for sharing that view, @wilkinsona.

[...]

PR #25446, opened 2021-02-26
Improve deployment integration tests

dreis2211:
Hi,

this PR improves the `spring-boot-deployment-tests` build step by splitting them apart based on their underlying container (e.g. Wildfly, TomEE etc.).

By doing so we can make use of static per-class testcontainers instead of starting a container for every test. This reduces the amount of started containers from 12 to 4 and therefore obviously the test time as well. (On my machine it only takes a third of what it took before). My hope is that this reduces the pressure on CI as well and helps to minimize flaky tests reported in #25410.

Let me know what you think.
Cheers,
Christoph

P.S.: The separate renaming commit was kind of intentional in order to preserve history for `DeploymentIntegrationTests`.


wilkinsona:
Lovely stuff! Thanks very much once again, @dreis2211.

PR #25451, opened 2021-02-26
Polish

izeye:

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/KafkaMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/KafkaMetricsAutoConfiguration.java
@@ -76,7 +76,7 @@ public class KafkaMetricsAutoConfiguration {
 	static class KafkaStreamsMetricsConfiguration {
 
 		@Bean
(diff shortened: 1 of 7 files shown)


buildSrc/src/test/java/org/springframework/boot/build/context/properties/ConfigurationPropertiesTests.java
```diff
@@ -32,7 +32,7 @@
 class ConfigurationPropertiesTests {
 
 	@Test
-	void whenJsonHasAnIntegerDefaultValueThenItRemainsAnIntegerWhenRead() {
+	void whenJsonHasAnIntegerDefaultValueThenItRemainsAsIntegerWhenRead() {
```
wilkinsona, line 35: `An` is correct here, IMO.

wilkinsona (changes requested):
Thanks very much, @izeye. There's one change that I don't agree with. WDYT?

buildSrc/src/test/java/org/springframework/boot/build/context/properties/ConfigurationPropertiesTests.java
izeye: @wilkinsona Thanks for the quick feedback!

I didn't know that a noun could come after "remain". After searching it a bit, I found I was wrong. I reverted it in 2420f0c.

snicoll:
Thanks for the follow-up @izeye!

Issue #25433, opened 2021-02-25
Cannot enforce CloudPlatform.NONE via spring.main.cloud-platform


breun:
I'm using Spring Boot 2.3.9.RELEASE and have an auto-configuration class annotated with `@ConditionalOnCloudPlatform(CloudPlatform.KUBERNETES)`. In the tests for this auto-configuration class I want to verify that this auto-configuration is only executed when running on Kubernetes.

I have these two tests:

```
@Test
void should_load_config_when_on_kubernetes() {
    new ApplicationContextRunner()
        .withUserConfiguration(DefaultConfiguration.class)
        .withSystemProperties("spring.main.cloud-platform=kubernetes")
        .run(context -> assertThat(context).hasSingleBean(LocalPod.class));
}

@Test
void should_not_load_configuration_when_not_on_kubernetes() {
    new ApplicationContextRunner()
        .withUserConfiguration(DefaultConfiguration.class)
        .withSystemProperties("spring.main.cloud-platform=none")
        .run(context -> assertThat(context).doesNotHaveBean(LocalPod.class));
}
```

These tests pass locally on my machine, but on GitLab CI these tests are executed in a GitLab Runner running on Kubernetes. In that setup the second test fails. The reason for this seems to be that the logic for `@ConditionalOnCloudPlatform` in `OnCloudPlatformCondition` calls `CloudPlatform.isActive(Environment)`, which is implemented like this:

```
public boolean isActive(Environment environment) {
	return isEnforced(environment) || isDetected(environment);
}
```

This means that even when setting `spring.main.cloud-platform=none` to force the cloud platform to be `CloudPlatform.NONE` in a test, the actual cloud platform the code is running on will still be detected, so you can't simulate not running on a cloud platform in a test.

I would expect `CloudPlatform.isActive` to be implemented like this, so detection is only used if a cloud platform is not explicitly forced:

```
public boolean isActive(Environment environment) {
	if (isEnforced(environment)) return true;
	return isDetected(environment);
}
```

wilkinsona:
Thanks for the report. That sounds like a bug to me. Not only will the condition behave incorrectly but I think both `NONE` and `KUBERNETES` will say that they're active while `CloudPlatform.getActive(Environment)` with return `NONE`.

philwebb:
Closing in favor of PR #25455

PR #25455, opened 2021-02-27
CloudPlatform.isActive can return true when spring.main.cloud-platform is set to NONE

nguyensach:
Fixes https://github.com/spring-projects/spring-boot/issues/25433

add isAutoDetectionEnabled method to determine that whether it is enabled that the platform is detected by looking for platform-specific environment variables.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -29,6 +29,7 @@ import org.springframework.core.env.StandardEnvironment;
  *
  * @author Phillip Webb
  * @author Brian Clozel
+ * @author Nguyen Sach
  * @since 1.3.0
  */
 public enum CloudPlatform {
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the pull request, @nguyensach. An alternative approach would be to change the condition. It could get the active CloudPlatform and see if it matches the required platform rather than checking if the required platform is active. I think I prefer what you've proposed here but I want to check with the team to see if the current behaviour of `CloudPlatform` is intentional. The follow code is in `ConfigDataActivationContext`:

https://github.com/spring-projects/spring-boot/blob/59d8c976344e47d7fdb95ae96424b15b11358147/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataActivationContext.java#L58-L65

It's making me wonder if there are some subtleties here that I'm overlooking. By calling isEnforced on each CloudPlatform and then falling back to `getActive` it seems to be allowing a CloudPlatform to be enforced ahead of an earlier one that would have been detected as active.

nguyensach:
@wilkinsona Thanks for your response. I think that the code in `ConfigDataActivationContext` is right. When `deduceCloudPlatform` method is called, there is not acitved profile.
https://github.com/spring-projects/spring-boot/blob/59d8c976344e47d7fdb95ae96424b15b11358147/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataActivationContext.java#L37-L38

Therefore, All `CloudPlatform#isEnforced(environment)` will be false. Then only check `isDetected`.

philwebb:
Thanks very much @nguyensach for contributing to Spring Boot. This is now in 2.3.x, 2.4.x and 2.5.x

PR #25457, opened 2021-02-27
Speed up spring-boot-server-tests

dreis2211:
Hi,

this PR improves the build time of `spring-boot-server-tests` by applying some caching in certain places. Specifically:

- A cache for the `ApplicationBuilder`s, which avoids packaging the JAR (or WAR) for each test invocation. With the cache the packaging now only run once per container and packaging method (e.g. Tomcat & JAR or Undertow & WAR)
- A cache for the AbstractApplicationLauncher, which pretty much follows the same idea as the tests don't need to start the application for each test invocation, but rather only once per container, packaging & launcher combination.

The second cache is a bit more aggressive and has the implication that the started processes can only be killed after all tests are run. Unfortunately, I found no way to make this a little cleaner in JUnit 5 as the `TestTemplateInvocationContext(Provider)` mechanism only works on the single test invocation level. The same applies to the exposed additional extensions in those contexts, which the `AbstractApplicationLauncher` is.

On my local machine the new approach reduces the time for testing from almost 8 minutes to roughly 1:30-2 minutes. The bigger part of this comes from the launcher cache, because we don't have to pay the 3-4 second application startup time hit on every test invocation.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-tests/spring-boot-integration-tests/spring-boot-server-tests/src/test/java/org/springframework/boot/context/embedded/AbstractApplicationLauncher.java
+++ b/spring-boot-tests/spring-boot-integration-tests/spring-boot-server-tests/src/test/java/org/springframework/boot/context/embedded/AbstractApplicationLauncher.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -26,7 +26,6 @@ import java.util.ArrayList;
 import java.util.List;
 
 import org.awaitility.Awaitility;
-import org.junit.jupiter.api.extension.AfterEachCallback;
 import org.junit.jupiter.api.extension.BeforeEachCallback;
 import org.junit.jupiter.api.extension.ExtensionContext;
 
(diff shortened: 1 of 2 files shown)


wilkinsona:
This is beautiful, Christoph. Thanks very much.

wilkinsona:
Unfortunately, this has [made Windows grumpy](https://ge.spring.io/s/zjgiszbensdgc/tests). I think the problem, or at least part of the problem, is that the processes are still running when we attempt to clean up the jar files. I'm hopeful that destroying the processes and then clean up the temp dir will fix it.

dreis2211:
@wilkinsona I'm very sorry - I don't have a Windows machine anymore to test against :/

wilkinsona:
@dreis2211 A wise man once said to me not to worry about breaking some tests… Seriously, please don't worry about it. These things happen after all. I reviewed the changes and it didn't occur to me that it'd cause grief with Windows.

dreis2211:
I saw it's still failing...what about bypassing the cache on Windows as a last resort?

wilkinsona:
The Windows builds only run daily unless we trigger them manually so they hadn't run with my latest tweak. I've just triggered [a 2.3.x Windows build](https://ci.spring.io/teams/spring-boot/pipelines/spring-boot-2.3.x/jobs/windows-build/builds/525) 🤞 .

wilkinsona:
The build [passed](https://ge.spring.io/s/qbil5d2pjl7fy) with `spring-boot-server-tests:test` taking a little over 7 mins. They took [16 mins yesterday](https://ge.spring.io/s/lvjbcvp74bzqs/tests?collapse-all#:spring-boot-tests:spring-boot-integration-tests:spring-boot-server-tests).

Thanks again for the changes, @dreis2211. That's a very welcome reduction in the build time.

dreis2211:
Thank you for tackling the Windows stuff 🙏

PR #25486, opened 2021-03-03
Highlight link to Actuator API docs in the reference docs

cdalexndr:
Users that fast search this doc may miss that there is another document explaining endpoints api.
The api link is not accessible from main doc reference (https://docs.spring.io/spring-boot/docs/2.4.3/reference/html/), and it is scattered through this doc in 3 places.

This change helps highlight the API link into a header so that it is more visible.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
@@ -43,7 +43,7 @@ For Gradle, use the following declaration:
 
 
 [[production-ready-endpoints]]
-== Endpoints
+== Endpoints ({spring-boot-actuator-restapi}/html/[API])
 Actuator endpoints let you monitor and interact with your application.
 Spring Boot includes a number of built-in endpoints and lets you add your own.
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the pull request. I agree that we should make the link to the Actuator API documentation more prominent, but I don't think this is how we should do it. I'd rather make the following paragraph more prominent:

> To learn more about the Actuator's endpoints and their request and response formats, please refer to the separate API documentation ({spring-boot-actuator-restapi}/html/[HTML] or {spring-boot-actuator-restapi}/pdf/spring-boot-actuator-web-api.pdf[PDF]).

Perhaps it could be made a `TIP` and/or moved to earlier in the section so it doesn't get lost beneath the table. Would you like to update your PR to that effect?

cdalexndr:
@wilkinsona what about the other 2 links pointing to the same api doc?

wilkinsona:
I think the other two links are fine as they are.

cdalexndr:
@wilkinsona done

wilkinsona:
Thanks very much, @cdalexndr. I moved the tip down slightly (still before the tables) so that the reader has a bit more context before coming to it.

Issue #20732, opened 2020-03-30
Provide diagnostic information about the ordering of auto-configuration classes


ttddyy:
Currently, the order of auto configuration classes that being sorted and applied is not part of the evaluation report.
While debugging `@ConditionalOn...` on auto configurations, the actual order is very helpful information.
Can this be added as part of the auto config report?

The sorting of auto configuration is happening at `AutoConfigurationImportSelector.AutoConfigurationGroup#selectImports`.
So, here `ConditionEvaluationReport` can be retrieved from `beanFactory` and records the sorted auto config order information. Then, it can be used when writing the evaluation report.

wilkinsona:
See https://github.com/spring-projects/spring-boot/issues/1067 for some background. The classes were originally ordered by their fully-qualified class name. This then changed to ordering by their short name, matching how they are displayed.

I'm not sure how we can satisfy both requirements here. If you don't know the exact name of the auto-configuration class, alphabetical ordering is useful. If you want to know exactly what order the classes ran in, using the sorted ordering would be useful.

ttddyy:
I think current shortname alphabetical ordering is good to keep for conditions.

What I initially thought was rather the ordering of autoconfiguration classes. (ordered entries from `spring.factories`).

For example, having separate section in condition report:
```
Auto Configuration Processing Order
-----------------------------------
1. org.springframework.boot.autoconfigure.context.MessageSourceAutoConfiguration
2. org.springframework.boot.autoconfigure.context.PropertyPlaceholderAutoConfiguration
3. org.springframework.boot.autoconfigure.websocket.reactive.WebSocketReactiveAutoConfiguration
4. org.springframework.boot.autoconfigure.web.reactive.ReactiveWebServerFactoryAutoConfiguration
5. org.springframework.boot.autoconfigure.websocket.servlet.WebSocketServletAutoConfiguration
...
```

Or, may be using `ConditionEvaluationReport#recordConditionEvaluation` to record actual conditional order, but it may be too much detail if it shows all nested conditions?

Since each auto configuration classes may import other configurations, having actual import processed order may not be deterministic. But, showing order of auto configuration loading entries would give big picture of ordering and they are deterministic, so that developer can rely on.

Strictly speaking, this ordering info might not belong to "conditional evaluation report"  because it is one step before getting into the conditions.

If team thinks this should not be part of conditional evaluation report, at least I want them to be logged.

Thanks,

wilkinsona:
[...]

PR #25503, opened 2021-03-05
Log the order of auto-configuration classes

ttddyy:
There is no easy way to obtain the sorted order of the auto configuration classes unless putting the debugger to the application. It is a helpful information when developing/troubleshooting the auto configuration classes.

Currently, I use [a reflection hack](https://github.com/spring-projects/spring-boot/issues/20732#issuecomment-687544812
) to get this information, but it's better if this is available out of the box.

This PR adds the ordering to the TRACE level loging, but even better if it could be DEBUG level.

Relates to #20732

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationSorter.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationSorter.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal, @ttddyy. When we [originally discussed #20732](https://github.com/spring-projects/spring-boot/issues/20732#issuecomment-608560476), I think we felt that it would be best to somehow include the information in the original report and the proposal here doesn't do that. There's also the [actuator endpoint to consider](https://docs.spring.io/spring-boot/docs/2.4.2/actuator-api/htmlsingle/#conditions). Should the information be part of the existing endpoint, exposed via a new endpoint, or not exposed at all? Unfortunately, I don't think we can merge this and have people start to rely upon it when there's a strong chance that we'll want to change it in the future. As indicated on #20732 we need to do some design work before we can consider making any changes in this area. Thanks anyway.

PR #25505, opened 2021-03-05
Layertools may extract entries outside of the destination path

trungPa:
The check for subdirectory in ExtractCommand.java will always return true because `getAbsolutePath` does not resolve `..` and symlinks.
This PR uses `getCanonicalPath` and adds a file separator at the end of destination path to avoid treating directories with the same prefix as subdirectories (e.g /foo, /foobar).


wilkinsona:
Thanks for the proposal but I am not sure that this change is necessary due to this existing line:

https://github.com/spring-projects/spring-boot/blob/9d11781c0812abec1ead8b01a82986568e290939/spring-boot-project/spring-boot-tools/spring-boot-jarmode-layertools/src/main/java/org/springframework/boot/jarmode/layertools/ExtractCommand.java#L89

Cleaning the path should resolve any `..` entries. If you've seen a scenario where this isn't sufficient, perhaps you could add a test demonstrating the problem?

wilkinsona:
No, it is necessary. For example, the path for a `e/../../../../../e.jar` entry in the jar becomes `../../../../e.jar`.

wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @trungPa.

PR #25519, opened 2021-03-05
Add Java 16 CI pipeline

dreis2211:
Hi,

this is hopefully one of the last steps for the JDK 16 ticket #24402 (apart from the known issues like missing Gradle support). I say hopefully, because the Concourse stuff is impossible to test for me and I need your help with this.

As you probably want this in 2.4.x as well, I based this on the 2.4.x mainline. Hope that's okay.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/ci/images/ci-image-jdk16/Dockerfile
+++ b/ci/images/ci-image-jdk16/Dockerfile
@@ -3,7 +3,7 @@ FROM ubuntu:focal-20210119
 ADD setup.sh /setup.sh
 ADD get-jdk-url.sh /get-jdk-url.sh
 ADD get-docker-url.sh /get-docker-url.sh
-RUN ./setup.sh java11 java11
+RUN ./setup.sh java8 java16
 
 ENV JAVA_HOME /opt/openjdk
 ENV PATH $JAVA_HOME/bin:$PATH
--- a/ci/images/setup.sh
+++ b/ci/images/setup.sh
@@ -26,14 +26,14 @@ curl -L ${JDK_URL} | tar zx --strip-components=1
 test -f /opt/openjdk/bin/java
 test -f /opt/openjdk/bin/javac
 
-if [[ $1 != $2 ]]; then
-	SECONDARY_JDK_URL=$( ./get-jdk-url.sh $2 )
+if [[ $# -eq 2 ]]; then
+	TOOLCHAIN_JDK_URL=$( ./get-jdk-url.sh $2 )
 
-	mkdir -p /opt/openjdk-secondary
-	cd /opt/openjdk-secondary
-	curl -L ${SECONDARY_JDK_URL} | tar zx --strip-components=1
-	test -f /opt/openjdk-secondary/bin/java
-	test -f /opt/openjdk-secondary/bin/javac
+	mkdir -p /opt/openjdk-toolchain
+	cd /opt/openjdk-toolchain
+	curl -L ${TOOLCHAIN_JDK_URL} | tar zx --strip-components=1
+	test -f /opt/openjdk-toolchain/bin/java
+	test -f /opt/openjdk-toolchain/bin/javac
 fi
 
 ###########################################################
--- a/ci/images/setup.sh
+++ b/ci/images/setup.sh
@@ -26,6 +26,15 @@ curl -L ${JDK_URL} | tar zx --strip-components=1
 test -f /opt/openjdk/bin/java
 test -f /opt/openjdk/bin/javac
 
+if [[ $1 != $2 ]]; then
+	SECONDARY_JDK_URL=$( ./get-jdk-url.sh $2 )
+
+	mkdir -p /opt/openjdk-secondary
+	cd /opt/openjdk-secondary
+	curl -L ${SECONDARY_JDK_URL} | tar zx --strip-components=1
+	test -f /opt/openjdk-secondary/bin/java
+	test -f /opt/openjdk-secondary/bin/javac
+fi
 
 ###########################################################
 # DOCKER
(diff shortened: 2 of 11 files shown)


ci/images/ci-image-jdk16/Dockerfile
```diff
@@ -0,0 +1,12 @@
+FROM ubuntu:focal-20210119
+
+ADD setup.sh /setup.sh
+ADD get-jdk-url.sh /get-jdk-url.sh
+ADD get-docker-url.sh /get-docker-url.sh
+RUN ./setup.sh java8 java16
```
dreis2211, line 6: To elaborate on this. I think that executing things with the actual JDK is always better than working with the toolchain support. Hence specifying the same version everywhere else except for this JDK 16 image. As soon as everything works on JDK 16 (e.g. the Gradle Plugin) I would also suggest to specify `RUN ./setup.sh java16 java16`. 

This is based on my personal experience while working on the JDK upgrades in the past. The real deal always beats the toolchain/artificial setup.

ci/scripts/build-project.sh
```diff
@@ -5,5 +5,10 @@ source $(dirname $0)/common.sh
 repository=$(pwd)/distribution-repository
 
 pushd git-repo > /dev/null
-./gradlew -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false --no-daemon --max-workers=4 -PdeploymentRepository=${repository} build publishAllPublicationsToDeploymentRepository
+if [[ -d /opt/openjdk-secondary ]]; then
+  secondary_java_version=$( ./$(dirname $0)/get-secondary-java-version.sh )
+	./gradlew -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false --no-daemon --max-workers=4 -PdeploymentRepository=${repository} build publishAllPublicationsToDeploymentRepository -PtoolchainVersion=${secondary_java_version} -Porg.gradle.java.installations.auto-detect=false -Porg.gradle.java.installations.auto-download=false -Porg.gradle.java.installations.paths=/opt/openjdk-secondary/
```
dreis2211, line 10: I'm basically disabling auto-downloads and auto-detection here and specify the location for the toolchain to look for installed JDKs.
I hope this works on Concourse the same way as on my local machine.

ci/images/setup.sh
```diff
@@ -26,6 +26,15 @@ curl -L ${JDK_URL} | tar zx --strip-components=1
 test -f /opt/openjdk/bin/java
 test -f /opt/openjdk/bin/javac
 
+if [[ $1 != $2 ]]; then
```
wilkinsona, line 29: Could we change this to check to see if `$2` has been defined? That would remove the need to change the Dockerfiles for 8, 11, and 15.

ci/images/setup.sh
```diff
@@ -26,6 +26,15 @@ curl -L ${JDK_URL} | tar zx --strip-components=1
...
+	SECONDARY_JDK_URL=$( ./get-jdk-url.sh $2 )
```
wilkinsona, line 30: I think I'd prefer toolchain to secondary here. IMO, it ties it together a bit better with the Gradle terminology.

ci/scripts/build-project.sh
```diff
@@ -5,5 +5,10 @@ source $(dirname $0)/common.sh
 repository=$(pwd)/distribution-repository
 
 pushd git-repo > /dev/null
-./gradlew -Dorg.gradle.internal.launcher.welcomeMessageEnabled=false --no-daemon --max-workers=4 -PdeploymentRepository=${repository} build publishAllPublicationsToDeploymentRepository
+if [[ -d /opt/openjdk-secondary ]]; then
+  secondary_java_version=$( ./$(dirname $0)/get-secondary-java-version.sh )
```
wilkinsona, line 9: It'd be nice if we could pass in the toolchain java version rather than figuring it out using the script (clever though it is). I'm not 100% sure that's possible though due to how pipeline.yml is structured. I can experiment a bit when we come to merging this.

wilkinsona (commented):
Thanks very much, @dreis2211. This is looking really good. I've left a few comments but they're quite subjective/stylistic. Please let me know what you think.

ci/images/setup.sh
dreis2211: Good suggestion - at least when I can get rid of the `$1 != $2`. It felt natural to call this secondary in this case, but I like your suggestion more.

ci/scripts/build-project.sh
dreis2211: I wasn't sure either and that was the only way I could more or less play through (at least in my head). It would be great if I can leave this to you after I've tackled the other remarks.

ci/images/setup.sh
dreis2211: Yeah, I think I have a way already. Thanks for the suggestion

wilkinsona:
Thanks again, @dreis2211. I've pushed the changes with a few tweaks and updated the 2.4.x and 2.5.x pipelines. Let's see what Concourse makes of it :).

dreis2211:
@wilkinsona The good news - the pipeline works. The bad news - there seems to be an error that my local tests didn't show. I'll take a look

wilkinsona:
Thanks. It's looking pretty good I think. The launch script tests try to use the Gradle plugin which hasn't been built. I think we should just disable them for now when using a Java 16 toolchain.

dreis2211:
Yeah - that was my idea as well because it's only temporary anyhow (until Gradle 7 is out). Already working on a follow-up PR. Unfortunately, I had stuff in my local Gradle caches still and hence they ran locally. Sorry for that.

PR #25523, opened 2021-03-05
Exclude GradleBuild tasks on JDK 16 for now

dreis2211:
Hi,

unfortunately, the first [pipeline run](https://ci.spring.io/teams/spring-boot/pipelines/spring-boot-2.4.x/jobs/jdk16-build/builds/1) on JDK 16 failed due to the gradle-plugin not being built which the `spring-boot-launch-script-tests` and `spring-boot-loader-tests` depend upon. I've missed this on local tests, because previous runs populated my Gradle caches. I'm sorry for that.

As this is only a temporary solution until Gradle 7 is out, I chose to pragmatically disable things for now. There are other ways to fix this. E.g. by extending the toolchain DSL to enforce the compilation step, which would cause the `spring-boot-gradle-plugin` JAR to be built. Just let me know if you prefer that and I will play around with that idea.

Cheers,
Christoph

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/toolchain/ToolchainPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/toolchain/ToolchainPlugin.java
@@ -22,6 +22,7 @@ import java.util.List;
 import org.gradle.api.Plugin;
 import org.gradle.api.Project;
 import org.gradle.api.plugins.JavaPluginExtension;
+import org.gradle.api.tasks.GradleBuild;
 import org.gradle.api.tasks.compile.JavaCompile;
 import org.gradle.api.tasks.javadoc.Javadoc;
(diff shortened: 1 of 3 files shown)


dreis2211:
On a related note: Are you planning to eventually upgrade to Gradle 7 in 2.4.x at all?

snicoll:
> On a related note: Are you planning to eventually upgrade to Gradle 7 in 2.4.x at all?

We're running against the 7.0 milestones on `master` only at this point.

snicoll:
Thanks for the follow-up Christoph. Let's see how it goes and we can refine if needed.

dreis2211:
And we have a green [build](https://ci.spring.io/teams/spring-boot/pipelines/spring-boot-2.4.x/jobs/jdk16-build/builds/2) ladies and gentlemen.

PR #25533, opened 2021-03-07
Use only non-identifying parameters from last execution for batch with incrementer

hpoettker:
### The issue

When the `JobLauncherApplicationRunner` is used to start a Spring Batch job with a `JobParametersIncrementer`, parameter keys that have once been assigned a value cannot be unset without manipulating the history of previous job executions. If a parameter is not explicitly set to something, its value will be recovered from the last job execution (of the job, not the job instance).

This may be desirable for identifying parameters of the job instances but in my opinion not for non-identifying parameters.

### A minimal example

I've written a [minimal example](https://github.com/hpoettker/spring-cloud-task-gh-770/tree/without-task). Originally, I've set up the repository for an analogous issue with Spring Cloud Task: https://github.com/spring-cloud/spring-cloud-task/issues/770 The branch `without-task` uses only Spring Batch and Spring Boot.

First, I start a MySQL container and build the code:
```
$ docker run -p 3306:3306 --name mysql -e MYSQL_ROOT_PASSWORD=password -e MYSQL_DATABASE=task -d mysql:5.7.25
$ ./gradlew build
```
Then, I start the application twice:
```
$ java -jar build/libs/task-0.0.1-SNAPSHOT.jar foo=bar -yada=yada
$ java -jar build/libs/task-0.0.1-SNAPSHOT.jar foo=bar
```
From the logs, one can see that the second job execution is still started with the non-identifying parameter:
```
o.s.b.a.b.JobLauncherApplicationRunner   : Running default command line with: [foo=bar, -yada=yada]
o.s.b.c.l.support.SimpleJobLauncher      : Job: [SimpleJob: [name=job]] launched with the following parameters: [{run.id=1, yada=yada, foo=bar}]
...
o.s.b.a.b.JobLauncherApplicationRunner   : Running default command line with: [foo=bar]
o.s.b.c.l.support.SimpleJobLauncher      : Job: [SimpleJob: [name=job]] launched with the following parameters: [{run.id=2, yada=yada, foo=bar}]
```
It seems to me that there is no way to ever unset the `yada` parameter without manually removing the corresponding row in the database table for the last job execution which effectively means rewriting history.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
@@ -211,13 +211,13 @@ public class JobLauncherApplicationRunner implements ApplicationRunner, Ordered,
 		}
 		JobParameters nextParameters = new JobParametersBuilder(jobParameters, this.jobExplorer)
 				.getNextJobParameters(job).toJobParameters();
-		return merge(nextParameters, jobParameters);
+		return merge(getIdentifying(nextParameters), jobParameters);
 	}
 
 	private JobParameters getNextJobParametersForExisting(Job job, JobParameters jobParameters) {
 		JobExecution lastExecution = this.jobRepository.getLastJobExecution(job.getName(), jobParameters);
 		if (isStoppedOrFailed(lastExecution) && job.isRestartable()) {
-			JobParameters previousIdentifyingParameters = getGetIdentifying(lastExecution.getJobParameters());
+			JobParameters previousIdentifyingParameters = getIdentifying(lastExecution.getJobParameters());
 			return merge(previousIdentifyingParameters, jobParameters);
 		}
 		return jobParameters;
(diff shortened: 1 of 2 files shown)


snicoll:
Thanks for the PR. 

paging @benas for a review

benas:
Identifying and non-identifying job parameters are specific to Spring Batch, and Spring Boot should not be concerned with that. As far as Spring Boot is concerned, it should call the `JobParametersIncrementer` if one is provided by the user and that's it. It should not start to interpret parameters and remove non-identifying ones. This would go against the contract of a `JobParametersIncrementer` which is expected to get all previous parameters and not only identifying ones. Changing the user's input in any way would be a mistake in my opinion.

@hpoettker In your minimal example, if you remove Spring Boot and use the `CommandLineJobRunner` from Spring Batch, you will see the same behaviour when starting the next job instance with `-next`. So changing this in boot would be inconsistent with batch and a breaking change for those who are migrating to boot.

If you want to filter non-identifying job parameters, you can use a custom `JobParametersIncrementer`. You can find an example [here](https://github.com/benas/spring-batch-lab/tree/master/issues/boot25533).

> It seems to me that there is no way to ever unset the yada parameter without manually removing the corresponding row in the database table for the last job execution

As mentioned previously, this should be done in your custom incrementer, and the non-identifying parameter will be unset in the subsequent execution automatically, see [here](https://github.com/benas/spring-batch-lab/tree/master/issues/boot25533#check-job-parameters-in-db).

snicoll:
Thanks a lot  for the review @benas. I am going to close this PR now, thanks for your contribution in any case.

hpoettker:
@beans Thanks for the detailed answer. Makes total sense.

But it still seems a bit surprising to me that `JobParametersIncrementer` has this responsibility. It should be considered to extend the java doc of `JobParametersIncrementer` or `JobParametersBuilder::getNextJobParameters` accordingly.

PR #25537, opened 2021-03-08
Move DatabaseSelection to Neo4jClient

meistermeier:
There is an upcoming change in the main branch of Spring Data Neo4j that requires changes in the `Neo4j(Reactive)DataAutoConfiguration`.
Previously the database selection was only supported on the template level and upwards. With the change (https://github.com/spring-projects/spring-data-neo4j/pull/2161) the selection will be already on the `Neo4jClient` level and as a consequence cannot be defined on a higher level (template) anymore.


snicoll:
Thanks for the PR.

> There is an upcoming change in the main branch of Spring Data Neo4j

When is this change scheduled? I can see a Spring Data `2021.0.0-M5` scheduled before our next milestone. Is that change going to be applied?

meistermeier:
Yes, this will be part of the next milestone release (`2021.0.0-M5`).

snicoll:
Thanks. I've applied this (and more) as part of switching to SD snapshots, see https://github.com/spring-projects/spring-boot/commit/46baf462ac83efecd3bfcd481efb6b2a7840bd07

PR #25550, opened 2021-03-08
Run spring-boot-cli tasks on Java 16

dreis2211:
Hi,

this PR removes the `maximumCompatibleJavaVersion` setting from `spring-boot-cli` that was introduced in 2.4.x, because it only supports Groovy 2.x which is not supporting Java 16. On 2.5.x and Groovy 3.x things should be running just fine, though.

Cheers,
Christoph


wilkinsona:
I had intended to drop this as part of merging forwards the original changes but evidently I forgot. Thanks very much, @dreis2211.

PR #25562, opened 2021-03-09
Deprecate FlywayMigrationScriptMissingFailureAnalyzer

dreis2211:
Hi,

the current builds on JDK 8 seem to [fail](https://ci.spring.io/builds/147913)

This PR deprecates `FlywayMigrationScriptMissingFailureAnalyzer` as part of the now deprecated location checking for Flyway.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211. Sorry for the build break. Yet more evidence that I shouldn't be trying to do two things at once.

PR #25596, opened 2021-03-10
Add Javadoc since to Bootstrapper.initialize()

izeye:
This PR adds Javadoc `@since` tag to `Bootstrapper.initialize()`.

See gh-25400


wilkinsona:
Thanks very much, @izeye.

PR #25620, opened 2021-03-14
When multiple schema or data script locations are configured and no resources can be found at a location, you can't tell which location is faulty

### LINKED ISSUE #25613: When multiple schema or data script locations are configured and no resources can be found at a location, you can't tell which location is faulty
`application.properties`:

```
spring.datasource.data=dml-one.sql,dml-two.sql
```

Failure:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

Invalid value 'dml-one.sql,dml-two.sql' for configuration property 'spring.datasource.data' (originating from 'class path resource [application.properties] - 1:24'). Validation failed for the following reason:

The specified resource does not exist.

Action:

Review the value of the property with the provided reason.
```

Rather than saying "The specified resource does not exist", we should indicate which location (in this case `dml-one.sql` or `dml-two.sql`) resulted in no resources being found when it was resolved.

uc4w6c:
Hi

Resolves #25613 

Thanks.


wilkinsona:
Thank you very much for making your first contribution to Spring Boot, @uc4w6c.

PR #25690, opened 2021-03-16
Extendable JooqExceptionTranslator

cdalexndr:
`JooqExceptionTranslator` needs to allow extension for users to custom handle untranslated exceptions.
https://github.com/spring-projects/spring-boot/pull/25493

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jooq/JooqExceptionTranslator.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jooq/JooqExceptionTranslator.java
@@ -57,7 +57,7 @@ public class JooqExceptionTranslator extends DefaultExecuteListener {
 		}
 	}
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal but I don't think this is a change that we should be considering at this time. As I said in https://github.com/spring-projects/spring-boot/pull/25493#issuecomment-800314086, I think this should be pursued at the Spring Framework level, for the time being at least, as we don't want to increase the surface area of Spring Boot's public API unnecessarily. Depending on the outcome of things in Framework, we can then consider what if anything needs to be done in Spring Boot. In the meantime, `JooqExceptionTranslator` is less than 100 lines long so maintaining your own variant shouldn't be too cumbersome.

If we want to make the auto-configured exception translation easily configurable then I think we'll need to make more changes than are proposed here. Ideally, the auto-configured `jooqExceptionTranslatorExecuteListenerProvider` would back off in favour of one provided by the user.

cdalexndr:
Workaround: copy paste code from spring boot `JooqExceptionTranslator` in my class to do custom behavior...

Issue #22824, opened 2020-08-07
Provide health for an AbstractRoutingDataSource's resolved targets


wilkinsona:

philwebb:
Closing in favor of PR #25708

limkinZero:
Hi @bono007 @wilkinsona , I´m check this functionality in a spring boot (2.5.0) test app with one abstractRoutingDataSource and I have some dudes.

Is possible, show the health of resolve target datasources which have been added after the initialization of the datasource bean. In my case, I fill the map with info pickup from one table on my database to create the data sources.

Regards,

bono007:
@limkinZero do you mind asking the question on [Gitter](https://gitter.im/spring-projects/spring-boot#) instead of this closed issue?

PR #25708, opened 2021-03-17
Provide health for an AbstractRoutingDataSource's resolved targets

bono007:
Fixes gh-22824

Screen Shot 2021-03-16 at 9 32 53 PM

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java
@@ -17,19 +17,19 @@
 package org.springframework.boot.actuate.autoconfigure.jdbc;
 
 import java.util.Collection;
+import java.util.Iterator;
 import java.util.Map;
+import java.util.function.Function;
 import java.util.stream.Collectors;
 
 import javax.sql.DataSource;
 
 import org.springframework.beans.factory.InitializingBean;
 import org.springframework.beans.factory.ObjectProvider;
-import org.springframework.boot.actuate.autoconfigure.health.CompositeHealthContributorConfiguration;
 import org.springframework.boot.actuate.autoconfigure.health.ConditionalOnEnabledHealthIndicator;
-import org.springframework.boot.actuate.health.AbstractHealthIndicator;
-import org.springframework.boot.actuate.health.Health.Builder;
+import org.springframework.boot.actuate.health.CompositeHealthContributor;
 import org.springframework.boot.actuate.health.HealthContributor;
-import org.springframework.boot.actuate.health.HealthIndicator;
+import org.springframework.boot.actuate.health.NamedContributor;
 import org.springframework.boot.actuate.jdbc.DataSourceHealthIndicator;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java
```diff
@@ -64,8 +71,7 @@
 @ConditionalOnEnabledHealthIndicator("db")
 @AutoConfigureAfter(DataSourceAutoConfiguration.class)
 @EnableConfigurationProperties(DataSourceHealthIndicatorProperties.class)
-public class DataSourceHealthContributorAutoConfiguration extends
-		CompositeHealthContributorConfiguration<AbstractHealthIndicator, DataSource> implements InitializingBean {
```
bono007, line 68: I broke out of the CHCC hierarchy as this stretched that contract a bit too much. We need the `createIndicator` to return `HealthContributor` rather than `AbstractHealthIndicator` to allow for the case where one of the datasource beans is actual the AbstractRoutingDataSource which in turn is composite contributor.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java
```diff
@@ -108,14 +122,29 @@ private String getValidationQuery(DataSource source) {
...
-	static class RoutingDataSourceHealthIndicator extends AbstractHealthIndicator {
+	static class RoutingDataSourceHealthIndicator implements CompositeHealthContributor {
+
+		private CompositeHealthContributor delegate;
+
+		RoutingDataSourceHealthIndicator(AbstractRoutingDataSource routingDataSource,
+				Function<DataSource, HealthContributor> indicatorFunction) {
+			Map<String, DataSource> routedDataSources = routingDataSource.getResolvedDataSources().entrySet().stream()
+					.collect(toMap((e) -> e.getKey().toString(), Map.Entry::getValue));
+			delegate = CompositeHealthContributor.fromMap(routedDataSources, indicatorFunction);
```
bono007, line 137: We could have simply returned this `CompositeHealthContributor.fromMap(routedDataSources, indicatorFunction)` from [here](https://github.com/spring-projects/spring-boot/pull/25708/files#diff-3c55a3e9dc27c7e0efe132a5e0251634b7b86dfe54f65143f3233f5219f01601R114) and not created a 1st class type for it. However, its helpful for testing as well as for later extensions to the indicator if needed.

bono007:
@wilkinsona not sure if this is what you had in mind - let me know where we need to adjust. I have a couple of points/questions/caveats to discuss as well.

### Relies on metadataprovider
As it is written, it relies on the available metadata providers to generate the validation query. Will it ever be the case that one of the routed datasources is not of the types we know about? Is that even possible? 

### Duplicate indicators
If a routed datasource is also a registered `@Bean DataSource` then it will have a regular `DataSourceIndicator` created as well as be represented in the `RoutingDataSourceHealthIndicator`.  The `IsolationLevelDataSourceRouter` seems like an example where the DS are available to the app by themselves but the router is a convenience to choose the correct one based on desired isolation level. What should we do in this case?

wilkinsona:
This looks great. Thank you, @bono007. 

> I am inclined to go w/ option 1. I am curious to hear others thoughts as well.

Agreed. I think this is the better option.

> Having the RoutingDataSourceHealthIndicator is helpful for testing as well as for future extensions to the indicator when/if needed

Agreed again. When looking at the main code I was tempted to get rid of it, but having looked at the tests I think it's useful to have the distinction. It's also a package-private implementation detail so we can refine in the future as needed without worrying about backwards compatibility.

PR #25709, opened 2021-03-17
Upgrade to Java 16 GA builds in CI

dreis2211:
Hi,

with JDK 16 going GA we can switch to GA builds in the CI environment.

Cheers,
Christoph


snicoll:
Thanks again Christoph.

PR #25721, opened 2021-03-17
Support sending metrics to InfluxDB v2

shakuzen:
Micrometer 1.7.0-M1 adds support for InfluxDB v2 in addition to the existing support for InfluxDB v1. Some concepts have changed in v2. So we need additional properties for users to be able to configure where to write metrics and authenticate via token.

Maybe we want to consider also updating the description of existing properties to note which are specific to v1. What do you think?

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/influx/InfluxProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/influx/InfluxProperties.java
@@ -16,10 +16,12 @@
 
 package org.springframework.boot.actuate.autoconfigure.metrics.export.influx;
 
+import io.micrometer.influx.InfluxApiVersion;
 import io.micrometer.influx.InfluxConsistency;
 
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/influx/InfluxProperties.java
```diff
@@ -33,7 +35,7 @@
 public class InfluxProperties extends StepRegistryProperties {
 
 	/**
-	 * Tag that will be mapped to "host" when shipping metrics to Influx.
+	 * Destination to write metrics when using an InfluxDB 1.x database.
```
shakuzen, line 38: The previous description was just wrong and some version of the updated description should probably be backported.

snicoll: Thanks Tommy. I've created: https://github.com/spring-projects/spring-boot/issues/25723

snicoll:
> Maybe we want to consider also updating the description of existing properties to note which are specific to v1. What do you think?

That's a good idea. I've done that in 8808d5e. Thanks!

Issue #14685, opened 2018-10-04
Couchbase health check fails due to timeout


prafsoni:
Spring boot -  2.0.5.RELEASE

I see in release notes a similar issues was supposed to be resolved. However, I am seeing this quite frequently(every 30min-60min) in logs, service become unhealthy and gets back healthy in about 30sec or so.

```
2018-10-04 12:08:01.182  WARN 1 --- [io-8080-exec-15] o.s.b.a.c.CouchbaseHealthIndicator       : Couchbase health check failed
--
  | java.util.concurrent.TimeoutException: null
  | at com.couchbase.client.java.util.Blocking.blockForSingle(Blocking.java:77)
  | at com.couchbase.client.java.bucket.DefaultBucketManager.info(DefaultBucketManager.java:127)
  | at org.springframework.boot.actuate.couchbase.CouchbaseHealthIndicator.getBucketInfo(CouchbaseHealthIndicator.java:84)
  | at org.springframework.boot.actuate.couchbase.CouchbaseHealthIndicator.doHealthCheck(CouchbaseHealthIndicator.java:75)
  | at org.springframework.boot.actuate.health.AbstractHealthIndicator.health(AbstractHealthIndicator.java:84)
  | at org.springframework.boot.actuate.health.CompositeHealthIndicator.health(CompositeHealthIndicator.java:68)
  | at org.springframework.boot.actuate.health.HealthEndpointWebExtension.getHealth(HealthEndpointWebExtension.java:50)
  | at sun.reflect.GeneratedMethodAccessor149.invoke(Unknown Source)
  | at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
  | at java.lang.reflect.Method.invoke(Method.java:498)
  | at org.springframework.util.ReflectionUtils.invokeMethod(ReflectionUtils.java:223)
  | at org.springframework.boot.actuate.endpoint.invoke.reflect.ReflectiveOperationInvoker.invoke(ReflectiveOperationInvoker.java:76)
  | at org.springframework.boot.actuate.endpoint.annotation.AbstractDiscoveredOperation.invoke(AbstractDiscoveredOperation.java:61)
  | at org.springframework.boot.actuate.endpoint.web.servlet.AbstractWebMvcEndpointHandlerMapping$ServletWebOperationAdapter.handle(AbstractWebMvcEndpointHandlerMapping.java:274)
[...]

PR #25726, opened 2021-03-17
If at least one couchbase node is connected to the cluster, couchbase is in up the health indicator

Rohatsahin:
In the couchbase java sdk [documentation](https://docs.couchbase.com/java-sdk/current/howtos/health-check.html), the cluster state diagnostic is specified as degraded when at least one node is connected.
However, cluster state is expected online for couchbase health indicator in method isCouchbaseUp in class CouchbaseHealth and this makes the application look unhealthy while couchbase is at least operational.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/couchbase/CouchbaseHealth.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/couchbase/CouchbaseHealth.java
@@ -48,7 +48,7 @@ class CouchbaseHealth {
 	}
 
 	private boolean isCouchbaseUp(DiagnosticsResult diagnostics) {
-		return diagnostics.state() == ClusterState.ONLINE;
+		return diagnostics.state() != ClusterState.OFFLINE;
 	}
 
 	private Map<String, Object> describe(EndpointDiagnostics endpointHealth) {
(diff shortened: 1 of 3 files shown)


pivotal-issuemaster:
@Rohatsahin Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=25726)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=25726) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@Rohatsahin Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=25726)!

wilkinsona:
Thanks for the suggestion but it is contrary to [the recommendation](https://github.com/spring-projects/spring-boot/issues/14685#issuecomment-429707626) that we received from the Couchbase team. We could perhaps make the mapping of ClusterState.DEGRADED configurable, but I don't think we should change the default behaviour.

Rohatsahin:
> Thanks for the suggestion but it is contrary to [the recommendation](https://github.com/spring-projects/spring-boot/issues/14685#issuecomment-429707626) that we received from the Couchbase team. We could perhaps make the mapping of `ClusterState.DEGRADED` configurable, but I don't think we should change the default behaviour.

Rohatsahin:
thanks @wilkinsona, we encountered this situation when a node exits the cluster and this situation affects the readiness probe in Kubernetes. I think then we'll have to remove the couchbase health indicator

wilkinsona:
It's worth giving some thought to which external services belong in the readiness group as, if that service is shared, a single outage can lead to cascading failures as all application instances report that they aren't ready to handle traffic.

If you'd like us to consider making whether `DEGRADED` maps to `UP` or `DOWN` configurable, please open an issue. Thanks again for the pull request, but I think it's best that we close this one as we don't want to change the default behaviour.

PR #25731, opened 2021-03-18
Upgrade to Tomcat Tomcat for 2.3.x

Soontao:
Hello Spring Boot Maintainers

As the title, I know that you have `a semi-automated process` to upgrade dependencies, but in the 2.3.x branch, nothing has changed in these two weeks.
So I open the PR for upgrade the tomcat dependency to fix the vulnerability of tomcat. If you can perform these change, just upgrade in your way.

Thanks, have a nice day :)

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1807,7 +1807,7 @@ bom {
 			]
 		}
 	}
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, but our semi-automated process will pick up the latest version of Tomcat that’s available when we perform the next 2.3.x release in March.

Issue #22823, opened 2020-08-07
Provide metrics for an AbstractRoutingDataSource's resolved targets


wilkinsona:

wilkinsona:
Based on [this review comment](https://github.com/spring-projects/spring-boot/pull/25749/files#r611028656), I think the main thing that we need to consider here is how to name each target `DataSource`. There are a few scenarios to consider:

1. A target `DataSource` is a target of a single `AbstractRoutingDataSource`.
2. A target `DataSource` is a target of multiple `AbstractRoutingDataSource`s.
3. A `DataSource` is both a top-level bean and a target of a single `AbstractRoutingDataSource`.
4. A `DataSource` is both a top-level bean and a target of multiple `AbstractRoutingDataSource`s.

When a target `DataSource` is also a bean (scenarios 3 and 4), I think its `name` tag should continue to be its bean name. This will ensure backwards compatibility with 2.4 and earlier. This would then require a different tag to display its routing information. When the `DataSource` is a target of a single routing data source, the value of this new tag could be a combination of the router's bean name and the routing key. When the `DataSource` is a target of multiple routing data sources, the value would, ideally, convey that information. We're limited to a `String` so we'd have to delimit each value. In bean's name can, in theory, be anything you want so it's not immediately clear what a good delimiter would be.

When a target `DataSource` is only a target (scenarios 1 and 2), we need a unique name for it. The routing key is only guaranteed to be unique in the scope of the routing data source. We need something that will be unique across the whole application so we can't use the routing key alone. The bean name of the routing data source will be unique across the whole application so we could combine the bean name of the routing data source with the routing key. A downside of this would be that the name tag's value will have a different format depending on whether the target is or is not also exposed as a bean.

The four scenarios described above could be mapped onto the following beans:

- dsThree
- dsFour
- routerOne
  - key1 -> dsOne
  - key2 -> dsTwo
  - key3 -> dsThree
  - key4 -> dsFour
- routerTwo
  - key1 -> dsTwo
  - key2 -> dsFour

`dsThree` would be tagged with `name -> dsThree` and `routing -> routerOne:key3`.
`dsFour` would be tagged with `name -> dsFour` and `routing -> routerOne:key4,routerTwo:key2`
`dsOne` would be tagged with `name -> routerOne:key1` and `routing -> routerOne:key1`
`dsTwo` would be tagged with `name -> routerOne:key2,routerTwo:key1` and `routing -> routerOne:key2,routerTwo:key1`

[...]

PR #25749, opened 2021-03-20
Add metrics for AbstractRoutingDataSource.

bono007:
Fixes gh-22823

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfiguration.java
@@ -18,6 +18,7 @@ package org.springframework.boot.actuate.autoconfigure.metrics.jdbc;
 
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.stream.Collectors;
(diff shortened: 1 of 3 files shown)


bono007:
I will fix the conflicts w/in next 24 hrs - just noticed it.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfiguration.java
```diff
@@ -69,10 +74,31 @@
...
+					return;
+				}
+				if (properties.isIgnoreRoutingDataSources()) {
+					return;
+				}
+				AbstractRoutingDataSource routingDataSource = (AbstractRoutingDataSource) dataSource;
+				routingDataSource.getResolvedDataSources().forEach((routingKey, routedDataSource) -> {
+					boolean alreadyRegistered = dataSources.containsValue(routedDataSource);
+					if (!properties.isDeduplicateRoutingDataSources() || !alreadyRegistered) {
+						allDataSources.put(getRoutedDataSourceName(beanName, routingKey), routedDataSource);
```
wilkinsona, line 96: This is tricky.

If a `DataSource` is a target of multiple routing data sources, the name for its metrics will depend upon the ordering of the `dataSources` map. Similarly, if a `DataSource` is both a top-level bean and a target of an `AbstractRoutingDataSource` its name will also depend upon the ordering of the `dataSources` map.

I'm not sure exactly what the answer to this is at the moment, but I am sure that the names need to be deterministic. While the injected `dataSources` map is a `LinkedHashMap` (so it'll iterate in insertion order), I'm not sure that Framework guarantees that order. My reading of `DefaultListableBeanFactory.findAutowireCandidates` is that it does not.

Let's take a step back until we've completed the design work on #22823.

wilkinsona:
Thanks again for your efforts here, @bono007. I'm afraid that we've decided that we need to change tack on this one. Please see [this comment](https://github.com/spring-projects/spring-boot/issues/22823#issuecomment-817939467) for details.

bono007:
No worries @wilkinsona - sounds like a good plan.

PR #25778, opened 2021-03-24
DefaultErrorWebExceptionHandler does not remove MetaType.ALL when a quality values is present

framon:
The "match-all" media type is not properly ignored if it has a quality value, showing HTML content if accept header is, for example, `application/json, */*;q=0.9`.

The broken behaviour is rendering different content depending on the presence of a quality value and the fix ignores quality property while comparing media types.

```bash
$ curl -H "Accept: application/json, */*" http://localhost:8080/api
{"timestamp":"2021-03-24T03:38:37.888+00:00","path":"/api","status":404,
"error":"Not Found","message":null,"requestId":"c621637c-1"}⌅

$ curl -H "Accept: application/json, */*;q=0.9" http://localhost:8080/api
<html><body><h1>Whitelabel Error Page</h1><p>This application has no configured error view,
so you are seeing this as a fallback.</p><div id='created'>Wed Mar 24 00:38:41 BRT 2021</div>
<div>[14552d94-1] There was an unexpected error (type=Not Found, status=404).</div></body></html>⌅
```


philwebb:
Thanks very much for pull-request @framon. This has now been merged to 2.3.x, 2.4.x and master.

PR #25805, opened 2021-03-26
Tomcat keepalive customizations

### LINKED ISSUE #23539: Expose Tomcat keepAliveTimeout and maxKeepAliveRequests properties
Hi 👋

I was going through the [server properties for Tomcat](https://docs.spring.io/spring-boot/docs/current/reference/html/appendix-application-properties.html#server-properties) and I couldn't find the `keepAliveTimeout` and `maxKeepAliveRequests`.

At the moment we configure them programmatically using a customizer but I'm wondering if it'd be possible/useful for Spring Boot to expose them.

I also found [this comment](https://github.com/spring-projects/spring-boot/issues/15210#issuecomment-439834531).
In our case we have L7 load balancing and we set `maxKeepAliveRequests` to -1 (i.e. unlimited) with a `keepAliveTimeout` which is higher than the connection timeout.

mlubavin:
new properties:
server.tomcat.keep-alive-timeout
server.tomcat.max-keep-alive-requests

Fixes #23539


wilkinsona:
Thanks for the PR, but @parviz-93 is already working on this one.

PR #25809, opened 2021-03-26
Increase test coverage on Java 16

dreis2211:
Hi,

after a chat with @wilkinsona about Gradle 7 in 2.5.x it became apparent that it might be a little late to do this once it's out. As only Gradle 7 is supporting Java 16 at the moment, I went ahead and removed the remaining `maximumCompatibleJavaVersion` configurations from the build in favor of this PR, which mainly includes more `DisabledForJreRange` usages.

Let me know what you think.
Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211. That's considerably fewer `@DisabledForJreRange` annotations than I had feared we'd need.

PR #25829, opened 2021-03-30
Add detection of Azure App Service to CloudPlatform

jdubois:
__Add Azure platform support to Spring Boot.__

More specifically, this adds "Azure App Service" support to Spring Boot - as this is the Azure standard PaaS service, it makes sense to give it the "AZURE" enum name. If it provides any additional value, we can add later Azure Functions and Azure Spring Cloud, but those shouldn't be the default and would have their own enumeration name, as they work very differently.

This discovers Azure using [the "WEBSITE_SITE_NAME" environment properties](https://stackoverflow.com/questions/25678419/how-to-check-if-code-is-running-on-azure-websites), which is similar to what is done with Heroku (using the "DYNO" environment property).

I put Azure near the top of the file, trying to have some alphabetical sorting after "NONE", as I believe this is how people will read that class.

As a consequence, setting this value should put `server.forward-headers-strategy=native` by default, which will allow applications to receive sensitive security headers from the reverse proxy. More specifically, this will solve this error, when using Active Directory to secure App Service, which looks pretty common: https://stackoverflow.com/questions/60365017/redirect-url-for-spring-oauth2-app-on-azure-with-active-directory-invalid-redir

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
@@ -48,7 +48,7 @@ public enum CloudPlatform {
 	/**
 	 * Azure App Service platform.
 	 */
-	AZURE {
+	AZURE_APP_SERVICE {
 
 		@Override
 		public boolean isDetected(Environment environment) {
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
@@ -46,6 +46,18 @@ public enum CloudPlatform {
 	},
 
 	/**
+	 * Azure App Service platform.
+	 */
+	AZURE {
+
+		@Override
+		public boolean isDetected(Environment environment) {
+			return environment.containsProperty("WEBSITE_SITE_NAME");
+		}
+
+	},
+
+	/**
 	 * Cloud Foundry platform.
 	 */
 	CLOUD_FOUNDRY {
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
@@ -17,7 +17,11 @@
 package org.springframework.boot.cloud;
 
 import org.springframework.boot.context.properties.bind.Binder;
-import org.springframework.core.env.*;
+import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.EnumerablePropertySource;
+import org.springframework.core.env.Environment;
+import org.springframework.core.env.PropertySource;
+import org.springframework.core.env.StandardEnvironment;
 
 /**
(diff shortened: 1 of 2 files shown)


jdubois:
After internal discussion at Microsoft, I'm going to modify this PR to use "AZURE_APP_SERVICE" instead of "AZURE", in order to be clear on the service that is being used.

cuspymd:
There seems to be no problem with the operation, but I'm afraid I can't understand it. It's a very common variable name, could you be sure that only azure uses this variable name?

Is it difficult to add an environment variable representing azure? It may sound rude, but I ask because I'm really curious.

jdubois:
You are correct @cuspymd but unfortunately there isn't a good specific environment variable for Azure. Let me try to find a second one, and then I'll use the combinaison of them to reduce the risk.

jdubois:
I pushed my commit, but I know there is a formatting issue - it breaks when I "gradle build" on my machine, but it doesn't give me any indication of what the formatting rules are, or what the error is.... I'm hoping I'll have a better idea with the CI build.

cuspymd:
@jdubois Thank you for checking.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java
```diff
@@ -17,11 +17,7 @@
 package org.springframework.boot.cloud;
 
 import org.springframework.boot.context.properties.bind.Binder;
-import org.springframework.core.env.ConfigurableEnvironment;
-import org.springframework.core.env.EnumerablePropertySource;
-import org.springframework.core.env.Environment;
-import org.springframework.core.env.PropertySource;
-import org.springframework.core.env.StandardEnvironment;
+import org.springframework.core.env.*;
```
wilkinsona, line 20: Can you please revert this change to the imports?

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/cloud/CloudPlatformTests.java
```diff
@@ -50,7 +50,17 @@ void getActiveWhenNotInCloudShouldReturnNull() {
 		Environment environment = new MockEnvironment();
 		CloudPlatform platform = CloudPlatform.getActive(environment);
 		assertThat(platform).isNull();
+	}
 
+	@Test
+	void getActiveWhenHasWebsiteSiteNameShouldReturnAzureAppService() {
```
wilkinsona, line 56: It would be good to have a test or two that asserts that `AZURE_APP_SERVICE` isn't active when only one of the environment variables is set.

wilkinsona (changes requested):
Thanks for the PR, @jdubois. I've left a couple of comments for your consideration.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/cloud/CloudPlatformTests.java
jdubois: Thank you for your review, I have added 2 tests

jdubois (commented):
I fixed both issues, @wilkinsona can you check and approve the review?

wilkinsona (approved):

wilkinsona:
Thanks very much, @jdubois. The proposed changes have now been merged.

PR #25873, opened 2021-04-02
log4j2(xml) configuration enhancement, support <SpringProfile> tag

kongwu-:
Enhance the configuration of log4j2 (xml), support Profile-specific Configuration (`<SpringProfile>`), consistent with logback extension.

Spring Boot currently only enhances the Logback (XML) configuration to support the <springProfile> tag. This feature is very useful, but is not supported by Log4j2.

I copied the code in Log4j2 XML to parse the XML configuration and created a new SpringBootXmlConfiguration to support the <SpringProfile> tag, which is as simple and easy to use as Logback Extension.

**Compatibility issues with rewriting the Log4j2 parsing code:**

1. I just copied the `XmlConfiguration` code directly from Log4j2, adding very little code and making no big changes like formatting. If there is an update to Log4j2, it is easy to rewrite the parsing class and update it accordingly.
2. The `XmlConfiguration` class in Log4j2 was last updated in June 2019, with no updates between [2.12.0,2.14.1] and the default dependent version of Log4j2 in Springboot (master) is 2.14.1

**To sum up, there is no risk in this kind of enhancement**

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import ch.qos.logback.classic.joran.JoranConfigurator;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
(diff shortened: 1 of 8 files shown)


kongwu-:
Hot to execute recheck?

snicoll:
> Hot to execute recheck?

@kongwu please don't open duplicate PRs, it creates unnecessary noise for those watching this repository. Rather than relying on our CI to validate your changes does not fail the build, please run the build locally before submitting a contribution. 

Duplicate of #25874

Issue #22149, opened 2020-06-29
Investigate impact of switching default logging system to log4j2


philwebb:
We've had requests before to switch the default logging system from logback to log4j2. In order to prepare for Spring Boot 3.x, we should reassess what the impact of such a switch would be for our users.

(somewhat triggered by the request to support JSON logging [#5479])

rgoers:
@philwebb If there is anything the Apache Logging team can do to assist you in this please let us know.

wilkinsona:
@rgoers Thank you.

One area that's just come up is providing log4j2-based support for the [`<springProfile>`](https://docs.spring.io/spring-boot/docs/2.4.4/reference/htmlsingle/#profile-specific-configuration) and the [`<springProperty>`](https://docs.spring.io/spring-boot/docs/2.4.4/reference/htmlsingle/#environment-properties) elements that we currently support with Logback.

`<springProperty>` is rather like a lookup and is quite similar to the existing Spring Boot lookup. `<springProperty>` doesn't have a dependency on Spring Cloud whereas it appears the Log4j2's Spring Boot lookup does?

`<springProfile>` provides some basic conditional configuration support, allowing configuration to be applied based on the Spring profiles that are active. From what I've seen, there isn't something similar in Log4j2 at the moment.

Spring Boot 3 is by no means imminent (it won't be in 2021) so there's no urgency here at the moment, but it would be interesting to explore what plugging `<springProperty>` and `<springProfile>` support into Log4j2 might look like.

rgoers:
@wilkinsona The Spring Lookup was in log4j-spring-cloud-config when it was first released. However, it was broken out into log4j-spring-boot in 2.14.0. As you might imagine, that module only requires Spring Boot.

As for <springProperty>, I've never had a need for it so it never occurred to me to implement it. We simply configure the different config file locations for each profile in our Spring Cloud Config server in bootstrap.yml. That said I don't see any reason why we couldn't support "conditional" plugins. They would be a bit different than any other plugins as their action would take place while processing configuration file and would not appear in the resulting configuration object.  

I will add a Jira issue to Log4j 2 to add support for conditionals. In general, we want to stay away from introducing any kind of programming logic in the config file but having a component that simply decides whether its children should be included or not seems reasonable.

rgoers:
I have added support for <springProfile>. It should be in Log4j 2.15.0. I don't see a reason to add support for a SpringProperty plugin since 

[...]

PR #25874, opened 2021-04-02
log4j2(xml) configuration enhancement, support <SpringProfile> tag 

kongwu-:
Enhance the configuration of log4j2 (xml), support Profile-specific Configuration (`<SpringProfile>`), consistent with logback extension.

Spring Boot currently only enhances the Logback (XML) configuration to support the <springProfile> tag. This feature is very useful, but is not supported by Log4j2.

I copied the code in Log4j2 XML to parse the XML configuration and created a new SpringBootXmlConfiguration to support the <SpringProfile> tag, which is as simple and easy to use as Logback Extension.

**Compatibility issues with rewriting the Log4j2 parsing code:**

1. I just copied the `XmlConfiguration` code directly from Log4j2, adding very little code and making no big changes like formatting. If there is an update to Log4j2, it is easy to rewrite the parsing class and update it accordingly.
2. The `XmlConfiguration` class in Log4j2 was last updated in June 2019, with no updates between [2.12.0,2.14.1] and the default dependent version of Log4j2 in Springboot (master) is 2.14.1

**To sum up, there is no risk in this kind of enhancement**

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
@@ -26,7 +26,6 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 
-import ch.qos.logback.classic.joran.JoranConfigurator;
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
 import org.apache.logging.log4j.Marker;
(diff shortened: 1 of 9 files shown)


kongwu-:
I have built it locally through

wilkinsona:
Thanks for the proposal.

I'm afraid that I'm not keen on copying Log4j 2's `XmlConfiguration` into Spring Boot and then modifying it, regardless of how stable the code may be. Even if we were comfortable with the approach, we'd also need to do the same for `JsonConfiguration` and `YamlConfiguration` as Boot supports yaml and json configuration in addition to XML.

If you would like to see Spring Boot's Log4j2 integration support springProfile and springProperty as they do with Logback, I think it should be done via some officially supported plug points in Log4j2. https://github.com/spring-projects/spring-boot/issues/22149 is tracking possibly switching from Logback to Log4j2 and the Log4j team have offered their help. We can use that issue to explore things.

Issue #24225, opened 2020-11-20
Raise the minimum supported version of Gradle to 6.x


wilkinsona:
Support for Gradle 5.6 has been deprecated since 2.3 so we should drop it in 2.5. Until Gradle 7 is released we should continue to support 6.3+ of Gradle 6.x. If Gradle 7 is released during 2.5's development, we should consider supporting the last 6.x release and 7.x.

divyajnu08:
Hi Andy , can I take up this issue ?

snicoll:
@divyajnu08 thank you for the offer. We're not yet sure how we're going to approach this so I don't think it's a great candidate for an external contribution at this point I am afraid. That said, we might have a better idea during the course of 2.5.x (and when/if Gradle 7 is released) so that this issue can be clarified and worked on by an external contributor. We're eager to help you learn so please continue watching this repository for an existing issue (or contribute new ideas of your own).

divyajnu08:
Thank you Stephane and Andy !! I will continue watching this repository.

wilkinsona:
7.0 RC1 is currently due on 29 January so it looks supports that latest 6.x release and 7.x in 2.5 may work out.

wilkinsona:
We're going to drop support for 5.6.x now and then add official support for 7.0 when it GAs. We're currently testing against 7.0-milestone-3 so we're already in pretty good shape there. Also, we may yet raise the minimum version of Gradle 6 that we support.

PR #25981, opened 2021-04-09
Remove last mention of Gradle 5.6.x support from docs

dreis2211:
Hi,

the getting started section still mentions Gradle 5.6 support, but that was dropped a short while ago.

Cheers,
Christoph


wilkinsona:
Thanks, @dreis2211. I missed that one in https://github.com/spring-projects/spring-boot/issues/24225.

wilkinsona:
Thanks again, @dreis2211.

PR #25986, opened 2021-04-09
Remove unused method parameters

dreis2211:
Hi,

this PR removes a couple unused method parameters.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/EnvironmentPostProcessorApplicationListener.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/EnvironmentPostProcessorApplicationListener.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -86,10 +86,10 @@ public class EnvironmentPostProcessorApplicationListener implements SmartApplica
 			onApplicationEnvironmentPreparedEvent((ApplicationEnvironmentPreparedEvent) event);
 		}
 		if (event instanceof ApplicationPreparedEvent) {
-			onApplicationPreparedEvent((ApplicationPreparedEvent) event);
+			onApplicationPreparedEvent();
 		}
 		if (event instanceof ApplicationFailedEvent) {
-			onApplicationFailedEvent((ApplicationFailedEvent) event);
+			onApplicationFailedEvent();
 		}
 	}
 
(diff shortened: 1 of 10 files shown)


dreis2211:
Test failure seems unrelated.

philwebb:
Nice!

wilkinsona:
> Test failure seems unrelated

I've just [seen the same failure](https://ge.spring.io/s/gkqfsieykqe3c/tests/:spring-boot-tests:spring-boot-smoke-tests:spring-boot-smoke-test-saml2-service-provider:test/smoketest.saml2.serviceprovider.SampleSaml2RelyingPartyApplicationTests/loginShouldHaveAllIdentityProvidersToChooseFrom()#1). I believe it's due to [this change](https://github.com/spring-projects/spring-security/commit/d0d0a8d958f5e2b250c84319cf902556f4430a68#) in Spring Security. I've reached out to the Security team for some guidance as I'm not sure if the change is intentional and if we need to add `spring-security-web` to the smoke test's dependencies.

wilkinsona:
The `spring-security-web` dependency has been [reinstated](https://github.com/spring-projects/spring-security/commit/a8a7ab4ffafdd1942358c6feca92cb60ef0720cd).

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/EnvironmentPostProcessorApplicationListener.java
```diff
@@ -101,11 +101,11 @@ private void onApplicationEnvironmentPreparedEvent(ApplicationEnvironmentPrepare
...
 
-	private void onApplicationPreparedEvent(ApplicationPreparedEvent event) {
+	private void onApplicationPreparedEvent() {
 		finish();
 	}
 
-	private void onApplicationFailedEvent(ApplicationFailedEvent event) {
+	private void onApplicationFailedEvent() {
 		finish();
 	}
```
cuspymd, line 110: It seems a bit ambiguous to call it a handler function for the event by deleting the event parameter. 
Wouldn't it be better to get rid of the wrapper function and call finish() right away?

dreis2211: Thought about that actually, but I liked the clear separation of things here. For me they are still the containers for the logic that should happen in case of the respective event and just share the common `finish()` call - regardless of the fact that they don't need the parameter. But I have no strong feelings for either option. Let's see what the team thinks and I'll gladly change to whatever is preferred.

wilkinsona: I prefer it as currently proposed.

cuspymd: Thanks for the considering.

snicoll (approved):

Issue #25996, opened 2021-04-10
Auto-configure micrometer DiskSpaceMetrics


philwebb:
AFAICT we don't currently configure `DiskSpaceMetrics`.

izeye:
For your information, there was a PR against the `micrometer-spring-legacy` module that was pending for some reason.

See https://github.com/micrometer-metrics/micrometer/pull/703

/cc @jkschneider

snicoll:
Closing in favor of PR #26001

PR #26001, opened 2021-04-11
Add auto-configuration for Micrometer DiskSpaceMetrics

bono007:
@philwebb this one is a bit interesting as to where the "home" for the auto-configuration should go for "DiskSpaceMetrics". It is in the Micrometer JVM package but feels a bit more like "System" metrics. 

2 reasons I landed it in the JvmMetricsAutoConfiguration:
- I rooted it to `user.dir` which is the current working directory the JVM was launched from
- the metrics implementation uses methods on File that are from the viewpoint of the VM. For example, from [File.getUsableSpace](https://docs.oracle.com/javase/8/docs/api/java/io/File.html#getUsableSpace--) : 
    > Returns the number of bytes available to this virtual machine on the partition named by this abstract pathname

Fixes gh-25996

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
@@ -16,8 +16,11 @@
 
 package org.springframework.boot.actuate.autoconfigure.metrics;
 
+import java.io.File;
+
 import io.micrometer.core.instrument.MeterRegistry;
 import io.micrometer.core.instrument.binder.jvm.ClassLoaderMetrics;
+import io.micrometer.core.instrument.binder.jvm.DiskSpaceMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmGcMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmMemoryMetrics;
 import io.micrometer.core.instrument.binder.jvm.JvmThreadMetrics;
@@ -64,6 +67,12 @@ public class JvmMetricsAutoConfiguration {
 	@ConditionalOnMissingBean
 	public ClassLoaderMetrics classLoaderMetrics() {
 		return new ClassLoaderMetrics();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public DiskSpaceMetrics diskSpaceMetrics() {
+		return new DiskSpaceMetrics(new File(System.getProperty("user.dir")));
 	}
 
 }
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
@@ -72,7 +72,7 @@ public class JvmMetricsAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean
 	public DiskSpaceMetrics diskSpaceMetrics() {
-		return new DiskSpaceMetrics(new File(System.getProperty("user.dir")));
+		return new DiskSpaceMetrics(new File("."));
 	}
 
 }
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfigurationTests.java
@@ -54,43 +54,44 @@ class JvmMetricsAutoConfigurationTests {
 
 	@Test
 	void allowsCustomJvmGcMetricsToBeUsed() {
-		this.contextRunner.withUserConfiguration(CustomJvmGcMetricsConfiguration.class).run(
-		        validateBaseJvmMetricsBeansArePresentWithCustomBean("customJvmGcMetrics"));
+		this.contextRunner.withUserConfiguration(CustomJvmGcMetricsConfiguration.class)
+				.run(validateBaseJvmMetricsBeansArePresentWithCustomBean("customJvmGcMetrics"));
 	}
 
 	@Test
 	void allowsCustomJvmMemoryMetricsToBeUsed() {
-		this.contextRunner.withUserConfiguration(CustomJvmMemoryMetricsConfiguration.class).run(
-                validateBaseJvmMetricsBeansArePresentWithCustomBean("customJvmMemoryMetrics"));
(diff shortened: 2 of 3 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
```diff
@@ -66,4 +69,10 @@ public ClassLoaderMetrics classLoaderMetrics() {
 		return new ClassLoaderMetrics();
 	}
 
+	@Bean
+	@ConditionalOnMissingBean
+	public DiskSpaceMetrics diskSpaceMetrics() {
+		return new DiskSpaceMetrics(new File(System.getProperty("user.dir")));
```
bono007, line 75: Would we want to make this configurable? That could easily move it out of the "JVM" category if it was just some random path.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfigurationTests.java
```diff
@@ -43,41 +49,53 @@
...
+
+	@Test
+	void allowsCustomDiskSpaceMetricsToBeUsed() {
+		this.contextRunner.withUserConfiguration(CustomDiskSpaceMetricsConfiguration.class).run((context) -> {
+			validateBaseJvmMetricsBeansArePresent();
+			assertThat(context).hasBean("customDiskSpaceMetrics");
+		});
+	}
+
+	private ContextConsumer<AssertableApplicationContext> validateBaseJvmMetricsBeansArePresent() {
```
bono007, line 95: **[UNRELATED]**: Just refactored to make it a bit more clear as to what was changing/being verified in each test.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfigurationTests.java
```diff
@@ -43,41 +49,53 @@
 
 	@Test
 	void autoConfiguresJvmMetrics() {
-		this.contextRunner.run(
-				(context) -> assertThat(context).hasSingleBean(JvmGcMetrics.class).hasSingleBean(JvmMemoryMetrics.class)
-						.hasSingleBean(JvmThreadMetrics.class).hasSingleBean(ClassLoaderMetrics.class));
+		this.contextRunner.run((context) -> validateBaseJvmMetricsBeansArePresent());
```
cuspymd, line 52: Shouldn't the code be called like below for it to actually run?
```
this.contextRunner.run(validateBaseJvmMetricsBeansArePresent());
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfigurationTests.java
```diff
@@ -43,41 +49,53 @@
...
 	}
 
 	@Test
 	void allowsCustomJvmGcMetricsToBeUsed() {
-		this.contextRunner.withUserConfiguration(CustomJvmGcMetricsConfiguration.class)
-				.run((context) -> assertThat(context).hasSingleBean(JvmGcMetrics.class).hasBean("customJvmGcMetrics")
-						.hasSingleBean(JvmMemoryMetrics.class).hasSingleBean(JvmThreadMetrics.class)
-						.hasSingleBean(ClassLoaderMetrics.class));
+		this.contextRunner.withUserConfiguration(CustomJvmGcMetricsConfiguration.class).run((context) -> {
+			validateBaseJvmMetricsBeansArePresent();
```
cuspymd, line 58: Shouldn't the code be called like below for it to actually run?
```suggestion
			validateBaseJvmMetricsBeansArePresent().accept(context);
```

cuspymd (commented):
Since the `validateBaseJvmMetricsBeansArePresent()` function returns the `ContextConsumer` interface, it seems that the calling method needs to be changed.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfigurationTests.java
bono007: Ha! Yeh that would help if we actually ran the validation. Good catch. I will update.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfigurationTests.java
```diff
@@ -43,41 +49,49 @@
...
+	}
+
+	@Test
+	void allowsCustomDiskSpaceMetricsToBeUsed() {
+		this.contextRunner.withUserConfiguration(CustomDiskSpaceMetricsConfiguration.class)
+				.run(validateBaseJvmMetricsBeansArePresentWithCustomBean("customDiskSpaceMetrics"));
+	}
+
+	private ContextConsumer<AssertableApplicationContext> validateBaseJvmMetricsBeansArePresentWithCustomBean(
+			String customBean) {
```
snicoll, line 86: I wouldn't overuse `ContextConsumer` like this. The essence of the assertion is to show that a single bean of a given type is present, and the one with the custom name we've added there as well (hence it's used). By moving this to a method that takes "a bean name", the link between the two is less obvious so I don't think this abstraction improves readability.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
snicoll: Given how easy it is to expose it, I am not sure I would make it configurable (at least for a first version). However the disk space health indicator uses the current path (i.e. `new File(".")`) so I'd rather use that.

snicoll (commented):
Thanks for the PR @bono007. I've scheduled it for our next feature release.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfigurationTests.java
bono007: Fair enough. I will update sometime today.

snicoll (approved):

snicoll:
Thanks again @bono007.

bono007:
> Thanks again @bono007.

My pleasure @snicoll

PR #26078, opened 2021-04-13
Fix Spring Data Elasticsearch deprecations

dreis2211:
Hi,

this PR removes some deprecated `@Document` properties in favor of using the new `@Setting` annotation.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #26085, opened 2021-04-13
Mention special behavior of spring.config.additional-location in docs

rehevkor5:
- `spring.config.additional-location` doesn't work if you put it in a
config file.


snicoll:
@rehevkor5 good catch. Thank you for making your first contribution to Spring Boot.

PR #26086, opened 2021-04-13
Fix Docker update detection on CI

dreis2211:
Hi,

I just noticed that the Docker update detection job apparently didn't notice any new updates, while there are newer versions available. I traced this back to https://github.com/docker/docker-ce being deprecated in favor of moby or docker-cli. I chose moby here as the new url as this is basically the engine of Docker.

I decided against including the upgrade to 20.10.6 here as I want to see that the job actually creates tickets.

Cheers,
Christoph


philwebb:
Thanks again @dreis2211!

PR #26129, opened 2021-04-15
Avoid some redirects in docs

dreis2211:
Hi,

this PR polishes some links to avoid redirects (and one broken anchor).

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started.adoc
@@ -346,7 +346,7 @@ If you need to solve a specific problem, check there first.
 
 You can shortcut the steps below by going to https://start.spring.io and choosing the "Web" starter from the dependencies searcher.
 Doing so generates a new project structure so that you can <<getting-started-first-application-code,start coding right away>>.
-Check the {spring-initializr-docs}/#user-guide[Spring Initializr documentation] for more details.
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started.adoc
```diff
@@ -346,7 +346,7 @@ If you need to solve a specific problem, check there first.
 
 You can shortcut the steps below by going to https://start.spring.io and choosing the "Web" starter from the dependencies searcher.
 Doing so generates a new project structure so that you can <<getting-started-first-application-code,start coding right away>>.
-Check the {spring-initializr-docs}/#user-guide[Spring Initializr documentation] for more details.
+Check the {spring-initializr-docs}[Spring Initializr documentation] for more details.
```
dreis2211, line 349: I wasn't sure what more details the link to Initializr would give in that context, but I didn't want to remove it just yet. Just let me know and I'll remove it if you think it's not providing any benefit here.

snicoll: Yeah it's a good point. I wonder if https://github.com/spring-io/start.spring.io/blob/main/USING.adoc wouldn't be a better link now. What do you think?

dreis2211: Seems like a better link - I'll adjust my PR.

snicoll:
Thank you very much again, Christoph.

PR #26162, opened 2021-04-20
Fix broken assertions in tests

izeye:
This PR fixes broken assertions in tests.

This PR also restores the HTTP response body assertion that seems to have been dropped in 5873dddc1c9ff41a27eae9ea48f6bf681b859e7e accidentally.


snicoll:
Good catch. Thanks very much @izeye!

PR #26183, opened 2021-04-21
Make the program execution consistent with the default configuration indicated by the document

weixsun:
The current situation is as follows:

When integrating redis, whether you choose lettuce client, if you use it's pool, you must display and modify one or more pool configurations in the configuration file, otherwise the pool will not be used, as described in the document Inconsistent, it is easy to mislead users.

Not good at words, hope to explain it.😂

WechatIMG326


weixsun:
This pull is unfair to Jedis.

PR #26200, opened 2021-04-21
Test Maven plugin against latest Maven versions

slawekjaranowski:
Maven version 3.8.1 is only bug fixes for 3.6.3 so can be used in it tests instead of 3.6.2

https://maven.apache.org/docs/3.8.1/release-notes.html
https://maven.apache.org/docs/3.6.3/release-notes.html

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/build.gradle
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/build.gradle
@@ -115,7 +115,7 @@ task zip(type: Zip) {
 }
 
 prepareMavenBinaries {
-	versions "3.6.2", "3.5.4"
+	versions "3.8.1", "3.6.3", "3.5.4"
 }
 
 artifacts {


pivotal-issuemaster:
@slawekjaranowski Please sign the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=26200)!

[Click here](https://cla.pivotal.io/sync/spring?repositoryId=spring-projects/spring-boot&pullRequestId=26200) to manually synchronize the status of this Pull Request.

pivotal-issuemaster:
@slawekjaranowski Thank you for signing the [Contributor License Agreement](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=26200)!

snicoll:
Testing on 3.8.1 is a good idea but I wouldn't remove testing from the `3.6.x` line. I wonder if rather we shouldn't bump the system requirements. We're currently stating 3.3+. Flagging for team attention to see what the rest of the team thinks.

wilkinsona:
I agree that we should drop 3.6.x from the versions that we test against. I think we should add it along side the existing versions for now and also bump the system requirements as a separate issue.

slawekjaranowski:
please consider that it tests on **one** maven versions take about **6** minutes (2,6 GHz 6-Core Intel Core i7)

wilkinsona:
The length of time take to run the tests isn't ideal, but I don't think that's a good reason not to run them. We're in a similar situation with the Gradle plugin's tests where we've considered refining things a little to limit the number of tests that run against multiple versions. We could consider something similar with the Maven plugin's tests if it becomes problematic.

slawekjaranowski:
documentation fix: #26209

snicoll:
@slawekjaranowski can you please update your PR with the review above?

slawekjaranowski:
fixed.

by the way `3.8.1` could be as next build for `3.6.x` - it contains only fix for CVEs  security, more explanation at: https://maven.apache.org/docs/3.8.1/release-notes.html#why-does-this-version-have-the-value-3-8-1

snicoll:
We are aware of what `3.8.1` contains and it is not the same thing as `3.6.x`.

PR #26202, opened 2021-04-22
Enable Redis connection pool if commons-pool2 is on the classpath

weixsun:
See #26183  for some more background. jedis turns on the pool by default but lettuce is not.


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
```diff
@@ -69,13 +70,29 @@ DefaultClientResources lettuceClientResources() {
 		return DefaultClientResources.create();
 	}
 
+	@Bean
+	@ConditionalOnMissingBean(LettuceClientConfigurationBuilder.class)
+	@ConditionalOnProperty(prefix = "spring.redis.lettuce.pool", value = "enabled", havingValue = "false",
+			matchIfMissing = true)
+	LettuceClientConfigurationBuilder lettuceClientConfigurationBuilder() {
```
snicoll, line 77: Rather then exposing `LettuceClientConfigurationBuilder` as a bean, we should rather apply a `LettuceClientConfigurationBuilderCustomizer` to it when the pool library is on the classpath (and that new `enabled` property is not `false`).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -421,7 +421,7 @@ public void setPool(Pool pool) {
 		/**
 		 * Lettuce pool configuration.
 		 */
-		private Pool pool;
+		private final Pool pool = new Pool();
```
snicoll, line 424: Jedis should be updated as well I think.

snicoll (commented):
@weixsun thanks for the suggestion and the PR! I think Jedis is impacted the same way and pooling should be set automatically if `commons-pool2` is on the classpath.  Would you have time to look at that? If you do, please do not open another PR. This PR can be amended by pushing more commits on your `pool-redis` branch. If you don't have the time, no worries, we can review as part of merging this one.

weixsun (commented):
recode.

snicoll:
@weixsun there is no need to rebase your PR against `main` as we will do it as part of reviewing the PR. We prefer to rebase and squash all your commits into one anyway. Please note that every commit you push sends a notification to all watchers of this repository.

snicoll:
Duplicate of #26326

Issue #26151, opened 2021-04-18
Profiles programmatically set on the Envionment are no longer merged with active profile properties


patpatpat123:
Hello Spring Boot Team,

Many thanks for the release of 2.5.0-RC1.

I was using 2.5.0-M3 fine with `spring.profiles.active=mycustomeprofile,vault,kubernetes`, nothing wrong, very happy.

```
INFO [myservice,,] 10 --- [           main] Application  : The following profiles are active: mycustomeprofile,vault,kubernetes
```

Just upgraded to 2.5.0-RC1, no code change, just a pom bump.

now, I get from the log:

```
The following profiles are active: kubernetes
```

It seems there is an issue with the profile.

Thank you

wilkinsona:
This works for me with a basic project generated on start.spring.io and `spring.profiles.active=mycustomeprofile,vault,kubernetes` added to `application.properties`:

```

  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::            (v2.5.0-RC1)

2021-04-18 08:58:28.331  INFO 81906 --- [           main] com.example.demo.Gh26151Application      : Starting Gh26151Application using Java 1.8.0_282 on wilkinsona-a01.vmware.com with PID 81906 (…/gh-26151/target/classes started by awilkinson in …/gh-26151)
2021-04-18 08:58:28.333  INFO 81906 --- [           main] com.example.demo.Gh26151Application      : The following profiles are active: mycustomeprofile,vault,kubernetes
```

If you would like us to spend some more time investigating, please spend some time providing a complete yet minimal sample that reproduces the problem. You can share it with us by pushing it to a separate repository on GitHub or by zipping it up and attaching it to this issue.

patpatpat123:
Hello Andy,

Many thanks for your time looking at this.
While preparing the minimal example, I had some technical findings I would like to share with you.

I believe it might be related to a Spring Cloud Kubernetes Issue.
You are correct, on a local machine, it is behaving correctly, cannot reproduce the issue.
However, when the app in deployed in Kubernetes, I am observing the following:

[...]

PR #26213, opened 2021-04-22
Append active profiles of the properties into active profiles of the environment

nguyensach:
Fixes https://github.com/spring-projects/spring-boot/issues/26151


mbhave:
Thanks for the PR @nguyensach. There were few edge cases with this code so we decided to take a [slightly different approach](cfa26735d26bd94c885787aa0bba3599b233e846).

PR #26221, opened 2021-04-23
Reproducible build for spring-boot-maven-plugin:build-info

slawekjaranowski:
In order to make maven build will be reproducible we use property `project.build.outputTimestamp` in maven project.

So `spring-boot-maven-plugin:build-info` should take this value as default to not break reproducible build.

Most of maven plugins takes into account this property.

https://maven.apache.org/guides/mini/guide-reproducible-builds.html

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/BuildInfoIntegrationTests.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/BuildInfoIntegrationTests.java
@@ -64,6 +64,15 @@ public class BuildInfoIntegrationTests {
 	}
 
 	@TestTemplate
+	void generatedBuildInfoReproducible(MavenBuild mavenBuild) {
+		mavenBuild.project("build-info-reproducible")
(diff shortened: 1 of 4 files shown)


wilkinsona:
Both build info's `time` and `project.build.outputTimestamp` support ISO-8601 formatted input so this is already supported by setting `time` using the value of `project.build.outputTimestamp`:

```xml
<plugin>
    <executions>
        <execution>
            <goals>
                <goal>build-info</goal>
            </goals>
            <configuration>
                <time>${project.build.outputTimestamp}</time>
            </configuration>
        </execution>
    </executions>
</plugin>
```

The risk of changing the default behaviour is pretty small, but, given that it's already possible with minimal additional configuration, I do wonder if we should consider this to be an enhancement and implement it in 2.6.

slawekjaranowski:
It is ok for me, that you scheduled change for 2.6 version.
Can I do something before merge?

wilkinsona:
@slawekjaranowski No, nothing for you to do, thank you. We'll aim to merge this for inclusion in 2.6.0-M1 once we've released 2.5.0 and created the 2.5.x maintenance branch.

Arianepd:
testando

snicoll (approved):

PR #26230, opened 2021-04-25
Fixing ordering for multiple classes that share same configuration property

### LINKED ISSUE #26217: Not fully consistent ordering in spring-configuration-metadata.json
[Current comparator implementation](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java#L161-L189) produces not consistent order for special case, when a codebase contains multiple classes that shares same "namespace" and (optionally) same properties.

Consider this example:
```java
package com.example;

@ConfigurationProperties("app.section")
public class Foo {
  private int val;
}

@ConfigurationProperties("app.section")
public class Bar {
  private boolean flag;
  private int val;
}
```
As both classes share same section (and it works in runtime just fine), this leads to multiple outcomes in `groups` sections, as well as in `properties` section:
```json5
  "groups": [
    {
      "name": "app.section",                          // order can be swapped!
      "type": "com.example.Foo",
      "sourceType": "com.example.Foo"
    },
    {
      "name": "app.section",
      "type": "com.example.Bar",
      "sourceType": "com.example.Bar""
    }
  ],
```

Same is applied for `val` in this exampe, as name is same, but type/sourceType is different.

I understand that this is extreme example, but it is valid in terms of spring support, so I think that adding sort stage for type can solve this problem:
```java
private final Comparator<ItemMetadata> propertyComparator = Comparator.<ItemMetadata, Boolean>comparing(ItemMetadata::isDeprecated)
                .thenComparing(ItemMetadata::getName)
                .thenComparing(ItemMetadata::getType);
private final Comparator<ItemMetadata> groupComparator = Comparator.<ItemMetadata, String>comparing(ItemMetadata::getName).thenComparing(ItemMetadata::getType);
```

zeldigas:
Fixes #26217

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java
@@ -170,9 +170,9 @@ class JsonConverter {
 		@Override
 		public int compare(ItemMetadata o1, ItemMetadata o2) {
 			if (o1.isOfItemType(ItemType.GROUP)) {
-				return groupComparator.compare(o1, o2);
+				return this.groupComparator.compare(o1, o2);
 			}
-			return itemComparator.compare(o1, o2);
+			return this.itemComparator.compare(o1, o2);
 		}
 
 		private boolean isDeprecated(ItemMetadata item) {
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java
@@ -161,8 +161,7 @@ class JsonConverter {
 	private static class ItemMetadataComparator implements Comparator<ItemMetadata> {
 
 		private final Comparator<ItemMetadata> itemComparator = Comparator.comparing(this::isDeprecated)
-				.thenComparing(ItemMetadata::getName)
-				.thenComparing(ItemMetadata::getSourceType);
+				.thenComparing(ItemMetadata::getName).thenComparing(ItemMetadata::getSourceType);
 
 		private final Comparator<ItemMetadata> groupComparator = Comparator.comparing(ItemMetadata::getName)
 				.thenComparing(ItemMetadata::getSourceType);
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java
@@ -160,26 +160,19 @@ class JsonConverter {
 
 	private static class ItemMetadataComparator implements Comparator<ItemMetadata> {
 
+		private final Comparator<ItemMetadata> itemComparator = Comparator.comparing(this::isDeprecated)
+				.thenComparing(ItemMetadata::getName)
+				.thenComparing(ItemMetadata::getSourceType);
+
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/test/java/org/springframework/boot/configurationprocessor/metadata/JsonMarshallerTests.java
```diff
@@ -82,7 +82,7 @@ void marshallOrderItems() throws IOException {
 		ByteArrayOutputStream outputStream = new ByteArrayOutputStream();
 		JsonMarshaller marshaller = new JsonMarshaller();
 		marshaller.write(metadata, outputStream);
-		String json = new String(outputStream.toByteArray());
+		String json = outputStream.toString();
```
zeldigas, line 85: ide suggestion to simplify statement

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java
```diff
@@ -160,26 +160,18 @@ private Object extractItemValue(Object value) {
...
-		private int compareProperty(ItemMetadata o1, ItemMetadata o2) {
-			if (isDeprecated(o1) && !isDeprecated(o2)) {
-				return 1;
-			}
-			if (isDeprecated(o2) && !isDeprecated(o1)) {
-				return -1;
+				return this.groupComparator.compare(o1, o2);
 			}
-			return o1.getName().compareTo(o2.getName());
+			return this.itemComparator.compare(o1, o2);
```
cuspymd, line 174: If `o1` is of `GROUP` type and `o2` is of `PROPERTY` type, comparison is made except for `this::isDeprecated`, but `o1` is of `PROPERTY` type and `o2` is of `GROUP` type, it includes `this::isDeprecated` to compare. Isn't it inconsistent?

Wouldn't it be better to always compare `this::isDeprecated` regardless of type?

zeldigas: This is not possible. Sort is applied after filtering of items to have just one type: https://github.com/spring-projects/spring-boot/blob/400b98700db7e5075888cba7bc50277b0d9a085b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java#L42-L43

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/metadata/JsonConverter.java
```diff
@@ -160,26 +160,18 @@ private Object extractItemValue(Object value) {
...
-			return o1.getName().compareTo(o2.getName());
-		}
-
-		private int compareProperty(ItemMetadata o1, ItemMetadata o2) {
-			if (isDeprecated(o1) && !isDeprecated(o2)) {
-				return 1;
-			}
-			if (isDeprecated(o2) && !isDeprecated(o1)) {
-				return -1;
+				return this.groupComparator.compare(o1, o2);
```
cuspymd, line 172: > This is not possible. Sort is applied after filtering of items to have just one type:

Sorry to keep bothering you. In the case of `GROUP` type, `this::is Deprecated` will always be false. Isn't it okay to use `itemComparator` regardless of type?

zeldigas: I don't use it for items, only for groups - take a look at `compare(ItemMetadata o1, ItemMetadata 02)`, diff can be misleading

snicoll:
@zeldigas thank you for making your first contribution to Spring Boot. I've polished things in d610f3c . In particular, the `sourceType` is not mandatory and that was breaking with a NPE in that case.

zeldigas:
Great, thank you!

PR #26253, opened 2021-04-27
Adjust to main branch in docs

dreis2211:
Hi,

this PR adjust two more places where the new `main` branch naming of Spring projects could be used. It doesn't break anything because GitHub redirects properly, but it shows a banner at the top which is slightly annoying (at least for me).

image

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211. I've taken care of the change to the dependency management plugin's docs location as part of [some changes](https://github.com/spring-projects/spring-boot/issues/26255) that I wanted to make across all three maintained branches.

Issue #26037, opened 2021-04-13
Add spring.webflux.multipart configuration properties


bclozel:
See spring-projects/spring-framework#26790 for background information.

We could provide configuration properties to make the multipart configuration easier with WebFlux. Currently this is supported by `org.springframework.http.codec.multipart.DefaultPartHttpMessageReader`. Note that we need to carefully consider the configuration keys, as several of them don't apply if the "streaming" mode is selected.

bclozel:
Closing in favor of #26254

PR #26254, opened 2021-04-27
Add spring.webflux.multipart configuration properties

bono007:
@bclozel I put together a WIP and wanted to get a nod from you (or not) on my current direction before investing in adding tests. 

Fixes gh-26037

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfiguration.java
@@ -1,0 +1,80 @@
+package org.springframework.boot.autoconfigure.web.reactive;
+
+import java.io.IOException;
+import java.io.UncheckedIOException;
+import java.nio.file.Paths;
+
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication.Type;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.context.properties.PropertyMapper;
+import org.springframework.boot.web.codec.CodecCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.annotation.Order;
+import org.springframework.http.codec.CodecConfigurer;
+import org.springframework.http.codec.multipart.DefaultPartHttpMessageReader;
+import org.springframework.util.unit.DataSize;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for multipart support in Spring
+ * Webflux.
+ * <p>
+ * Configures the {@link DefaultPartHttpMessageReader} via a {@link CodecCustomizer}.
+ *
+ * @author Chris Bono
+ * @since 2.6.0
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnClass({ CodecConfigurer.class, DefaultPartHttpMessageReader.class })
+@ConditionalOnWebApplication(type = Type.REACTIVE)
+@EnableConfigurationProperties(ReactiveMultipartProperties.class)
+class ReactiveMultipartAutoConfiguration {
+
+	@Bean
+	@Order(1)
+	CodecCustomizer defaultPartHttpMessageReaderCustomizer(ReactiveMultipartProperties multipartProperties) {
+		return (configurer) -> configurer.defaultCodecs().configureDefaultCodec((codec) -> {
+			if (!DefaultPartHttpMessageReader.class.isInstance(codec)) {
+				return;
+			}
+			DefaultPartHttpMessageReader defaultPartHttpMessageReader = (DefaultPartHttpMessageReader) codec;
+			boolean streaming = multipartProperties.getStreaming() != null && multipartProperties.getStreaming();
+			boolean unlimitedMaxMemorySize = multipartProperties.getMaxInMemorySize() != null
+					&& multipartProperties.getMaxInMemorySize().toBytes() == -1;
+			PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
+			map.from(multipartProperties::getStreaming).to(defaultPartHttpMessageReader::setStreaming);
+			if (!streaming) {
+				map.from(multipartProperties::getMaxInMemorySize).asInt(this::convertToBytes)
+						.to(defaultPartHttpMessageReader::setMaxInMemorySize);
+			}
+			if (!streaming && !unlimitedMaxMemorySize) {
+				map.from(multipartProperties::getMaxDiskUsagePerPart).as(this::convertToBytes)
+						.to(defaultPartHttpMessageReader::setMaxDiskUsagePerPart);
+				map.from(multipartProperties::getFileStorageDirectory)
+						.to(dir -> setFileStorageDirectory(dir, defaultPartHttpMessageReader));
+			}
+			map.from(multipartProperties::getMaxParts).to(defaultPartHttpMessageReader::setMaxParts);
+			map.from(multipartProperties::getMaxHeadersSize).asInt(this::convertToBytes)
+					.to(defaultPartHttpMessageReader::setMaxHeadersSize);
+			map.from(multipartProperties::getHeadersCharset).to(defaultPartHttpMessageReader::setHeadersCharset);
+		});
+	}
+
+	private void setFileStorageDirectory(String fileStorageDirectory,
+			DefaultPartHttpMessageReader defaultPartHttpMessageReader) {
+		try {
+			defaultPartHttpMessageReader.setFileStorageDirectory(Paths.get(fileStorageDirectory));
+		}
+		catch (IOException ex) {
+			throw new UncheckedIOException(ex);
+		}
+	}
+
+	private Long convertToBytes(DataSize size) {
+		return (size != null) ? size.toBytes() : null;
+	}
+
+}
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfigurationTests.java
@@ -1,0 +1,251 @@
+/*
+ * Copyright 2012-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.web.reactive;
+
+import java.lang.reflect.Method;
+import java.nio.charset.Charset;
+import java.nio.charset.StandardCharsets;
+import java.nio.file.Path;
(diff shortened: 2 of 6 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/DefaultPartHttpMessageReaderCustomizerConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+import org.springframework.boot.web.codec.CodecCustomizer;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.core.annotation.Order;
+import org.springframework.http.codec.multipart.DefaultPartHttpMessageReader;
+import org.springframework.util.unit.DataSize;
+
+@Configuration(proxyBeanMethods = false)
+@EnableConfigurationProperties(MultipartProperties.class)
+class DefaultPartHttpMessageReaderCustomizerConfiguration {
```
bono007, line 19: Naming... 

So, part of me (no pun intended - terrible, terrible) wants to name this more generically like `MultipartConfiguration` but on the other hand it is currently very specific to what the target of configuration is - `DefaultPartHttpMessageReader`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/DefaultPartHttpMessageReaderCustomizerConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+
+	@Bean
+	@Order(10)
```
bono007, line 22: The other built-in customizers run at `@Order(0)` - I just ran this one before those and I think it really does not matter much as the setting of most fields on the Codecs cause them all to be re-inited.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/DefaultPartHttpMessageReaderCustomizerConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+	CodecCustomizer defaultPartHttpMessageReaderCustomizer(MultipartProperties multipartProperties) {
+		return (configurer) -> configurer.defaultCodecs().configureDefaultCodec((codec) -> {
```
bono007, line 24: Not super happy about our options to "get at" the `DefaultPartHttpMessageReader`. 

I chose to use an auto-configured customizer rather than insert a specific `Consumer<Object>` in [WebFluxAutoConfiguration.java](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java#L182) such as :
```java
		@Override
		public void configureHttpMessageCodecs(ServerCodecConfigurer configurer) {
             configurer.defaultCodecs().configureDefaultCodec((codec) -> { ... });
			this.codecCustomizers.orderedStream().forEach((customizer) -> customizer.customize(configurer));
		}
```
Seemed less intrusive and more auto-configed-like.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/MultipartProperties.java
```diff
@@ -0,0 +1,135 @@
...
+ */
+@ConfigurationProperties(prefix = "spring.webflux.multipart", ignoreUnknownFields = false)
+public class MultipartProperties {
+
+	/**
+	 * Maximum amount of memory allowed per part in bytes. Leave blank to use default of
+	 * 256K or set to -1 to store all contents in memory. Ignored when streaming is
+	 * enabled.
+	 */
+	private DataSize maxInMemorySize;
```
bono007, line 40: What is the recommended approach here when we are configuring a target that already has defaults that we want to use? One one hand specifying  `private DataSize maxInMemorySize = DataSize.ofKilobytes(256);` the default here insulates us against any changes to the underlying target. However, it insulates us against any changes to the underlying target. Assuming we want to lock in-step w/ the underlying target this approach is fine.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/DefaultPartHttpMessageReaderCustomizerConfiguration.java
```diff
@@ -0,0 +1,65 @@
...
+			if (!DefaultPartHttpMessageReader.class.isInstance(codec)) {
+				return;
+			}
+			DefaultPartHttpMessageReader defaultPartHttpMessageReader = (DefaultPartHttpMessageReader) codec;
+			boolean streaming = multipartProperties.getStreaming() != null && multipartProperties.getStreaming();
```
bono007, line 29: This is all a bit messy when the underlying target already handles the ignores logic. Very similar concern here as w/ the [default values](https://github.com/spring-projects/spring-boot/pull/26254/files#r621253129).

aoudiamoncef:
Hi @bono007,

I'm interested on this PR, if you wanna a help, don't hesitate.

bono007:
Hi @aoudiamoncef - I almost forgot about this proposal, thanks for the reminder. I was waiting for nod from @bclozel before I continued w/ adding tests. I will go ahead and add them in the next 1-2 days.

bclozel:
Thanks @bono007 , please go ahead with tests and what you had in mind. I'll do a proper review of that one - I need to review what we have in Framework and probably ask around to make sure we're on the right track!

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/DefaultPartHttpMessageReaderCustomizerConfiguration.java
bclozel: Ha! The only other option I'm seeing is `ReactiveMultipartConfiguration` which doesn't leak implementation details and is specific enough - it's also in line with the `Properties` class. Once we're done with the overall implementation we can always look at the Servlet equivalent and check for inconsistencies.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/MultipartProperties.java
bclozel: What we can do is set the default value in the properties class but write a test that checks that both defaults are aligned.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/DefaultPartHttpMessageReaderCustomizerConfiguration.java
bclozel: Looks good to me!

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/MultipartProperties.java
bono007: I should have known this as I did that same thing when we aligned the CassandraProperties in another proposal a while back. :)

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+ * {@link EnableAutoConfiguration Auto-configuration} for multipart support in Spring
+ * Webflux.
+ * <p>
+ * Configures the {@link DefaultPartHttpMessageReader} via a {@link CodecCustomizer}.
+ *
+ * @author Chris Bono
+ * @since 2.6.0
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnClass({ CodecConfigurer.class, DefaultPartHttpMessageReader.class })
```
bono007, line 31: There may be a more tactical class(es) to use for this - I used what the auto-config is directly operating on.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfiguration.java
```diff
@@ -0,0 +1,80 @@
...
+@ConditionalOnWebApplication(type = Type.REACTIVE)
+@EnableConfigurationProperties(ReactiveMultipartProperties.class)
+class ReactiveMultipartAutoConfiguration {
+
+	@Bean
+	@Order(1)
```
bono007, line 37: Order is 1 so that it runs after the more generic CodecCustomizer which is at order 0.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfigurationTests.java
```diff
@@ -0,0 +1,251 @@
...
+ * Tests for {@link ReactiveMultipartAutoConfiguration}.
+ *
+ * @author Chris Bono
+ */
+class ReactiveMultipartAutoConfigurationTests {
+
+	private final ReactiveWebApplicationContextRunner contextRunner = new ReactiveWebApplicationContextRunner()
+			.withConfiguration(AutoConfigurations.of(ReactiveMultipartAutoConfiguration.class));
+
+	private static final boolean DEFAULT_STREAMING = false;
```
bono007, line 53: I chose to pull these defaults out to well-named vars to make the tests clear when the reader was using a default vs. when it had its value changed by the customizer.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfigurationTests.java
```diff
@@ -0,0 +1,251 @@
...
+	@Test
+	void shouldNotProvideCustomizerWhenCodecConfigurerNotAvailable() {
+		this.contextRunner.withClassLoader(new FilteredClassLoader(CodecConfigurer.class))
+				.run((context) -> assertThat(context).doesNotHaveBean(CodecCustomizer.class));
+	}
+
+	@Test
+	void customizerSetsAppropriatePropsWhenStreamingEnabled() {
+		this.contextRunner.withPropertyValues("spring.webflux.multipart.streaming:true",
+				"spring.webflux.multipart.max-in-memory-size=1GB", "spring.webflux.multipart.max-headers-size=512MB",
```
bono007, line 90: My DRY alarm is going off... however, I think the repetition is just at the cusp of acceptable. I made a pass at factoring out some of the common things between the tests but their slight differences are just enough that it was not an improvement (it was less readable).

bono007:
I believe this one is ready for re-review @bclozel . Thx

bclozel:
Thanks @bono007 - this is now merged.
I've simplified a bit your contribution: the `DefaultPartHttpMessageReader` itself doesn't validate its properties against conflicting values, so let's not do that on our side. Also, not mapping properties might confuse developers more than help them. The properties documentation looks good enough for that.

Closed with 19318bf5618020c3e

bono007:
Thx @bclozel . Yeh, I went back/forth on that. What drove my direction was mostly the [comment](https://github.com/spring-projects/spring-boot/issues/26037#issue-856733005) in the issue around: 

> Note that we need to carefully consider the configuration keys, as several of them don't apply if the "streaming" mode is selected.

I figured if we put the logic in the customizer it would insulate us from any changes in the DPHMR around setting/not setting. 

But I agree, this simplifies things.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/ReactiveMultipartAutoConfiguration.java
bono007: I see it in your merge commit (`WebConfigurer.class`)

PR #26258, opened 2021-04-27
Add support for Dynatrace metrics API v2

### LINKED ISSUE #24978: Add auto-configuration for Micrometer's Dynatrace v2 meter registry
See related Issue & PR :  
     https://github.com/micrometer-metrics/micrometer/issues/2295
     https://github.com/micrometer-metrics/micrometer/pull/2406

With the hopeful addition of the new Micrometer Dynatrace v2 Meter Registry, it will be necessary to add an additional Auto Configuration for it to this project in order to utilize it. This will allow users to add in the necessary application properties needed for dynatrace2 to send metrics to Dynatrace. 

I do have a working example for this, but cannot submit a PR as the micrometer PR is still open (as of 2021/22/1) and the required dependency would cause the build to fail.

pirgeo:
This PR contains new additions to the properties that allow Micrometer to export to two different Dynatrace APIs. The code in the two Micrometer PRs (see below) depends on this, and will hopefully be included in the 1.8.0 release of Micrometer.

Resolves #24978

- Micrometer issue: https://github.com/micrometer-metrics/micrometer/issues/2295
- Move Dynatrace exporter v1: https://github.com/micrometer-metrics/micrometer/pull/2553
- Add Dynatrace exporter v2: https://github.com/micrometer-metrics/micrometer/pull/2581

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
@@ -20,6 +20,7 @@ import io.micrometer.dynatrace.DynatraceApiVersion;
 import org.springframework.boot.actuate.autoconfigure.metrics.export.properties.StepRegistryProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
+import java.util.Collections;
 import java.util.Map;
 
 /**
@@ -27,6 +28,7 @@ import java.util.Map;
  * metrics export.
  *
  * @author Andy Wilkinson
+ * @author Georg Pirklbauer
  * @since 2.1.0
  */
 @ConfigurationProperties(prefix = "management.metrics.export.dynatrace")
@@ -156,7 +158,7 @@ public class DynatraceProperties extends StepRegistryProperties {
 	}
 
 	public Map<String, String> getDefaultDimensions() {
-		return defaultDimensions;
+		return defaultDimensions != null ? Collections.unmodifiableMap(defaultDimensions) : null;
 	}
 
 	public void setDefaultDimensions(Map<String, String> defaultDimensions) {
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
@@ -86,12 +86,12 @@ public class DynatraceProperties extends StepRegistryProperties {
 	private String metricKeyPrefix;
 
 	/**
-	 * An optional Boolean that allows enabling of the OneAgent metadata export. On by
+	 * An optional Boolean that allows enabling of the Dynatrace metadata export. On by
 	 * default.
 	 *
 	 * API v1: not applicable (ignored), API v2: optional
 	 */
-	private Boolean enrichWithOneAgentMetadata = true;
+	private Boolean enrichWithDynatraceMetadata = true;
 
 	/**
 	 * Optional default dimensions that are added to all metrics in the form of key-value
@@ -149,12 +149,12 @@ public class DynatraceProperties extends StepRegistryProperties {
 		this.metricKeyPrefix = metricKeyPrefix;
 	}
 
-	public Boolean getEnrichWithOneAgentMetadata() {
-		return this.enrichWithOneAgentMetadata;
+	public Boolean getEnrichWithDynatraceMetadata() {
+		return this.enrichWithDynatraceMetadata;
 	}
 
-	public void setEnrichWithOneAgentMetadata(Boolean enrichWithOneAgentMetadata) {
-		this.enrichWithOneAgentMetadata = enrichWithOneAgentMetadata;
+	public void setEnrichWithDynatraceMetadata(Boolean enrichWithDynatraceMetadata) {
+		this.enrichWithDynatraceMetadata = enrichWithDynatraceMetadata;
 	}
 
 	public Map<String, String> getDefaultDimensions() {
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
@@ -84,12 +84,12 @@ public class DynatraceProperties extends StepRegistryProperties {
 	private String metricKeyPrefix;
 
 	/**
-	 * An optional Boolean that allows enabling of the OneAgent metadata export. Off by
+	 * An optional Boolean that allows enabling of the OneAgent metadata export. On by
 	 * default.
 	 *
 	 * API v1: not applicable (ignored), API v2: optional
 	 */
-	private Boolean enrichWithOneAgentMetadata = false;
+	private Boolean enrichWithOneAgentMetadata = true;
 
 	/**
 	 * Optional default dimensions that are added to all metrics in the form of key-value
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
@@ -18,10 +18,9 @@ package org.springframework.boot.actuate.autoconfigure.metrics.export.dynatrace;
 
 import java.util.Map;
 
-import io.micrometer.dynatrace.DynatraceApiVersion;
-
 import org.springframework.boot.actuate.autoconfigure.metrics.export.properties.StepRegistryProperties;
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.DeprecatedConfigurationProperty;
 
 /**
  * {@link ConfigurationProperties @ConfigurationProperties} for configuring Dynatrace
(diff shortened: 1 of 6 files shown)


snicoll:
@pirgeo thank you for the PR but I am going to close this now as it isn't actionable. There is no way for us to review the  code as the changes in Micrometer are not available yet. When micrometer has a build with the necessary changes of this PR, we can obviously reconsider this.

jonatan-ivanov:
@snicoll The changes were merged to micrometer: https://github.com/micrometer-metrics/micrometer/pull/2619

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
```diff
@@ -97,4 +139,36 @@ public void setGroup(String group) {
...
+
+	public Boolean getEnrichWithOneAgentMetadata() {
+		return enrichWithOneAgentMetadata;
+	}
+
+	public void setEnrichWithOneAgentMetadata(Boolean enrichWithOneAgentMetadata) {
+		this.enrichWithOneAgentMetadata = enrichWithOneAgentMetadata;
+	}
+
+	public Map<String, String> getDefaultDimensions() {
```
jonatan-ivanov, line 158: I'm not sure if it is the preferred way in spring boot, but I might make this immutable: `Collections.unmodifiableMap(defaultDimensions);` (I don't really remember if configprops will like it or not).

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatracePropertiesConfigAdapter.java
```diff
@@ -16,14 +16,17 @@
...
-
 import org.springframework.boot.actuate.autoconfigure.metrics.export.properties.StepRegistryPropertiesConfigAdapter;
 
+import java.util.Map;
+
 /**
  * Adapter to convert {@link DynatraceProperties} to a {@link DynatraceConfig}.
  *
  * @author Andy Wilkinson
+ * @author Georg Pirklbauer
```
jonatan-ivanov, line 29: I think you can add yourself to the previous file and to the test class too.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
```diff
@@ -1723,25 +1723,87 @@ You can also change the interval at which metrics are sent to Datadog:
...
+Please ensure that the endpoint URI contains the path (e.g. `/api/v2/metrics/ingest`).
+
+[source,yaml,indent=0,configprops,configblocks]
+----
+	management:
+	  metrics:
+	    export:
+	      dynatrace:
+	        api-version: "v2"
+	        api-token: "YOUR_TOKEN" # should be read from environment variables or properties
```
jonatan-ivanov, line 1758: I would just say from a secure property source (there are other occurrences in this file).

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
wilkinsona: It should be modifiable for consistency with other configuration properties classes.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
```diff
@@ -1723,25 +1723,87 @@ You can also change the interval at which metrics are sent to Datadog:
...
-	        api-token: "YOUR_TOKEN"
 	        device-id: "YOUR_DEVICE_ID"
-	        uri: "YOUR_URI"
+	        # uri: "https://{your-domain}/e/{your-environment-id}" on managed deployments.
+	        uri: "https://{your-environment-id}.live.dynatrace.com"
+	        api-token: "YOUR_TOKEN" # should be read from a secure property source
 ----
 
-You can also change the interval at which metrics are sent to Dynatrace:
+You can also change the interval at which metrics are sent to Dynatrace (works for both API versions).
```
wilkinsona, line 1805: As this applies to both API versions, I don't think it belongs in the API v1 section.

pirgeo: You are right, I added a subsection for version independent settings.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/dynatrace/DynatraceProperties.java
pirgeo: I reset it to be modifiable. Thanks for the clarification!

spring-boot-project/spring-boot-docs/src/docs/asciidoc/production-ready-features.adoc
pirgeo: Changed it everywhere.

wilkinsona:
Given that there are some configuration properties that are specific to v1 and some that are specific to v2, I wonder if we should introduce separate groups for those. Properties that are common to both version of the API would remain at `management.metrics.export.dynatrace` while v1-specific properties would move to `management.metrics.export.dynatrace.v1` and, similarly, v2-specific properties would move to `management.metrics.export.dynatrace.v2`. WDYT, @snicoll?

snicoll:
I think that's an interesting idea. That said our influx support has the same problem and we're currently mapping the same way Micrometer does. I don't know if there's a precedent where our config diverges quite a bit from the related Micrometer's `Config` class.

pirgeo:
Would that be a breaking change, or is there a way to do this without breaking existing `v1` configurations? 
@jonatan-ivanov Have you come across a similar situation and have input here?

jonatan-ivanov:
@pirgeo On Boot side it depends how the Boot Team wants to handle this.
If we can come up with a worthy breaking change for Micrometer, that can be the subject of 2.x.
As Stéphane mentioned the Influx config is similar where the v1-v2 is handled similarly to this PR, though I really like Andy's proposal.

pirgeo:
Thanks @jonatan-ivanov. I also think the proposal makes sense. We would also prefer to introduce this as a non-breaking, additive change. This would allow us to provide a version compatible with Micrometer 1.x without having to wait for a 2.x release.
@snicoll, @wilkinsona Would the current proposal in this PR be fine with you?

I have also moved the documentation to the new location. The CI is currently breaking, since this code relies on the `1.8.0-SNAPSHOT` build of Micrometer. How can we go ahead here?

jonatan-ivanov:
@snicoll @wilkinsona Would it make sense to you to depend on Micrometer 1.8.0 in `main` early on so that Boot support for new features in Micrometer could be merged in?

wilkinsona:
Yes, I think so. We can do that once we have some dates for the Micrometer 1.8 milestones and we know that the schedules line up.

jonatan-ivanov:
@wilkinsona We are planning to release M1/M2/M3/RC1/GA 8 days before Boot releases these milestones, e.g.: `1.8.0-M1` is planned on 14th July, `2.6.0-M1` is on 22nd July. Can we include this in `2.6.0-M1`?

pirgeo:
I updated the PR to reflect the renaming of the config option introduced in https://github.com/micrometer-metrics/micrometer/pull/2692

wilkinsona:
I've pushed [a branch](https://github.com/wilkinsona/spring-boot/tree/gh-26258) which hopefully illustrates what I had in mind. The approach is to move v1- and v2-specific properties beneath `management.metrics.export.dynatrace.v1` and `management.metrics.export.dynatrace.v2` respectively. For backwards compatibility, the old v1-specific properties remain at `management.metrics.export.dynatrace` in a deprecated form.

Rather than having a property for the API version, it is now inferred based on whether or not the device ID has set. A downside of this approach is that a user who is attempting to use the V1 API will not get an error message is they forget to set the device ID. Instead, an attempt will be made to use the V2 API. The upside of the approach is that the IDE experience is better as auto-completion guides you towards the right properties and there's no chance of, for example, setting the API version to v1 and then trying to configure v2-specific properties.

pirgeo:
Hi @wilkinsona, Thanks for the suggestion, this is great. We have a small change in mind: would it be possible to rename `v1` and `v2` to `api-v1`  and `api-v2`, respectively? `v1`/`v2` sounds like registry-versions similar to `@since` where features can be mixed. When using `api-v1` or `api-v2` it's more clear that we mean the Dynatrace API version, which should clarify that only one of the two (the latest one by default) is used, and features cannot be mixed between versions. What do you think? I have enabled "Allow maintainers to modify" on this PR, so you should be able to push your commit to the open branch if you'd like, otherwise I can take that over. I'll put some finishing touches the docs after we've settled for one of the alternatives
Thanks!

wilkinsona:
Thanks for taking a look, @pirgeo. We don't use hyphens in configuration property prefixes so we try to keep each `.`-separated section of the prefix to a single word. We could use `apiv1` and `apiv2` or `v1api` and `v2api` but not `api-v1` and `api-v2`. Personally I prefer the concision of `v1` and `v2` and don't find it confusing, but you know your user base far better than I do. For `v1api` vs `apiv1` I prefer the former as the latter could be read as apiv 1.

pirgeo:
@wilkinsona Oh, I see. `apiv1` and `apiv2` does indeed look a bit odd. Lets go ahead with `v1` and `v2` then, and I will update the documentation to make it very clear for everyone reading it. Would you push your commit to this branch, or do you want me to do that? Thanks again for the support!

wilkinsona:
GitHub's giving me a 403 when I try to push to the `add-dynatrace-v2-properties` branch of `dynatrace-oss-contrib/spring-boot`. Please feel free to push my commit onto the branch yourself.

pirgeo:
@wilkinsona  I have pushed your commit and refined the documentation, it looks good from our end now. Thanks for your help!
I think the build will still break due to this branch pointing to Micrometer 1.7.0, which doesn't have our changes. How should we proceed here?

wilkinsona:
I can take care of that. I'll rebase it before merging it and that'll pick up the upgrade to 1.8.0-M1.

wilkinsona:
@pirgeo Thanks very much for making your first contribution to Spring Boot. This has now been merged into `main`.

pirgeo:
@wilkinsona Thanks a lot!

PR #26292, opened 2021-04-28
When a single bean causes a dependency cycle the failure analysis does not clearly illustrate the cycle

### LINKED ISSUE #26269: When a single bean causes a dependency cycle the failure analysis does not clearly illustrate the cycle
I've encountered a case where I've left a bean requiring itself to be autowired and it took me a long time to understand what was wrong since log didn't indicate the cycle properly (if there's one bean in cycle, there are no arrows or whatsoever).

How it looks now:
```
2021-04-28 13:43:34,929 [main] ERROR o.s.b.d.LoggingFailureAnalysisReporter:40 - 

***************************
APPLICATION FAILED TO START
***************************

Description:

The dependencies of some of the beans in the application context form a cycle:

┌─────┐
|  beanName defined in file [xxx/yyy.class]
└─────┘
```

This one would look better in this case:
```
┌──->──┐
|  beanName defined in file [xxx/yyy.class]
└──<-──┘
```

How it looks when there are more than one bean (it's clear to understand):
```
┌─────┐
|  beanName1 defined in file [xxx/yyy.class]
↑     ↓
|  beanName2 defined in file [xxx/zzz.class]
└─────┘
```

I'm willing to submit a PR for this issue if you think this change is an appropriate one.

Haarolean:
From now on, if a bean depends on itself we'll get such a message:
```
The dependencies of some of the beans in the application context form a cycle:

┌──->──┐
|  bean defined in org.springframework.boot.diagnostics.analyzer.BeanCurrentlyInCreationFailureAnalyzerTests$SelfReferenceBeanConfiguration
└──<-──┘
```

Closes #26269.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/diagnostics/analyzer/BeanCurrentlyInCreationFailureAnalyzer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/diagnostics/analyzer/BeanCurrentlyInCreationFailureAnalyzer.java
@@ -74,7 +74,7 @@ class BeanCurrentlyInCreationFailureAnalyzer extends AbstractFailureAnalyzer<Bea
 		for (int i = 0; i < beansInCycle.size(); i++) {
 			BeanInCycle beanInCycle = beansInCycle.get(i);
 			if (i == cycleStart) {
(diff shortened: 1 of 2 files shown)


snicoll:
@Haarolean thank you for making your first contribution to Spring Boot.

PR #26315, opened 2021-04-29
Link all three document formats

Buzzardo:
Per issue 26314, add links to each document format (multi-page HTML, single-page
HTML, and PDF)  such that each document format links to the other two.


wilkinsona:
Thanks for the PR, @buzzardo, but we already cross-link all three formats. See https://docs.spring.io/spring-boot/docs/2.4.x/reference/htmlsingle/#boot-documentation-about, for example.

PR #26321, opened 2021-04-30
Fix typo in NamedContributors Javadoc

matthiasstock:
Compared to the existing implementation [NamedContributorsMapAdapter](https://github.com/spring-projects/spring-boot/blob/4febe6485c226a8e5372bd66a68607150abec6cc/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/NamedContributorsMapAdapter.java) of the affected interface, only a typo needs to be corrected.


snicoll:
Good catch, thanks again @matthiasstock.

PR #26376, opened 2021-05-06
Optimize cache storage and solve memory release

wuwen5:
When the DataSource has been destroyed, the instance cannot be recycled.

In a dynamic refresh DataSource scenario,  the DataSource instances cannot be garbage collected after data source has destroyed.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/jdbc/DataSourcePoolMetrics.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/jdbc/DataSourcePoolMetrics.java
@@ -33,6 +33,7 @@ import org.springframework.boot.jdbc.metadata.DataSourcePoolMetadata;
 import org.springframework.boot.jdbc.metadata.DataSourcePoolMetadataProvider;
 import org.springframework.util.Assert;
 import org.springframework.util.ConcurrentReferenceHashMap;
+import org.springframework.util.ObjectUtils;
 
 /**
  * A {@link MeterBinder} for a {@link DataSource}.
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/jdbc/DataSourcePoolMetrics.java
```diff
@@ -108,10 +108,11 @@ public void bindTo(MeterRegistry registry) {
 
 		@Override
 		public DataSourcePoolMetadata getDataSourcePoolMetadata(DataSource dataSource) {
-			DataSourcePoolMetadata metadata = cache.get(dataSource);
+			int dataSourceKey = System.identityHashCode(dataSource);
```
weixsun, line 111: It might be better to use `ObjectUtils.identityToString(Object obj)`.

wilkinsona, line 111: I don't think this is safe. Over the lifetime of the JVM, there's no guarantee that every `identityHashCode` will be unique.

wilkinsona (commented):
Thanks for the pull request, @wuwen5. Unfortunately, I don't think we can make the change you have proposed due to the potential for identity hash codes not being unique. I think we'll need to take a step back and find another way to solve the problem. Perhaps you could open an issue that describes the problem you're facing in more detail, ideally illustrating it with a sample application or failing test. We can then consider how we may be able to fix it.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/jdbc/DataSourcePoolMetrics.java
```diff
@@ -108,10 +109,11 @@ public void bindTo(MeterRegistry registry) {
 
 		@Override
 		public DataSourcePoolMetadata getDataSourcePoolMetadata(DataSource dataSource) {
-			DataSourcePoolMetadata metadata = cache.get(dataSource);
+			String dataSourceId = ObjectUtils.identityToString(dataSource);
```
wilkinsona, line 112: This doesn't address my earlier concern. `ObjectUtils.identityToString` uses the object's class name and its identity hash code. That's still insufficient to guarantee uniqueness for the lifetime of the JVM.

wilkinsona:
Thanks for opening #26378, @wuwen5. Let's continue things over there.

PR #26399, opened 2021-05-06
Fix wrong reference to Cassandra in DockerImageNames#couchbase

nguyensach:
Fix typo in DockerImageNames. Fix Cassandra to Couchbase.


snicoll:
Good catch, thanks again @nguyensach.

PR #26416, opened 2021-05-08
Polish SpringApplicationAdminJmxAutoConfigurationTests

nguyensach:
Refactor SpringApplicationAdminJmxAutoConfiguration
- Because the default value of `matchIfMissing` is false in `ConditionalOnProperty`, delete unnecessary `matchIfMissing=false`.
- Add new `WhenThereAreNotAnyMBeanExporters` test case proving that `mbeanExporters` can not be null. So, there is no need to check if `mbeanExporters` is null.


nguyensach:
Ci failed because of a bug in Spring. I fixed this bug by https://github.com/spring-projects/spring-framework/pull/26910

snicoll:
@nguyensach thanks again for the PR. I've kept only the test as the change for `mbeanExporters` is a behavior change we can't apply so late in the cycle. As for `matchIfMissing` we had other places in a similar situation so I've applied the change consistently, see #26432.

PR #26422, opened 2021-05-11
Allow to configure StartMojo's wait and maxAttempts attributes from the command-line

marcus-nl:
This allows overriding the settings for the Maven StartMojo (`spring-boot:start`) from the command line, most notably `wait` and `maxAttempts`, but also the other missing ones and for StopMojo as well. It is useful to increase the timeout on an CI environment, where it can take a longer time to start than usual. I've also moved the default values to the `@Parameter` annotation, so that these show up in the documentation.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/StartMojo.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/StartMojo.java
@@ -61,13 +61,13 @@ public class StartMojo extends AbstractRunMojo {
 	 * The JMX name of the automatically deployed MBean managing the lifecycle of the
 	 * spring application.
 	 */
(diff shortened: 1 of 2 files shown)


marcus-nl:
> No doubt that exposing that via a property is useful but you can achieve that too by configuring the plugin with a property of your own and set that property in your CI environment.
True, but doesn't that apply to every configuration option? I think the timeout settings are particularly good candidates to give their own properties because their values may quite realistically differ in different circumstances. Not having to change the pom.xml (even to add your own property) makes it a lot more convenient.

> I am, in particular, not keen to expose separate properties for thing that must have the same value. Thoughts?
You are right. Alternative combined names could be something like "spring-boot.management.jmxName" (/Port) or "spring-boot.jmx.name" (port)? Or just leave them propertly-less, since these seem less likely to need customization. I would suggest keeping the default values in the annotation though, so they show up in the generated documentation.

snicoll (approved):

snicoll:
@marcus-nl thank you for making your first contribution to Spring Boot.

PR #26437, opened 2021-05-12
Allow the WebSessionManager's session ID resolver to be customized using a WebSessionIdResolver bean

weixsun:
When you want to set up `WebSessionIdResolver`, you don’t have to expose the `WebSessionManager` bean every time.

That should be the meaning, my word is poor.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
@@ -23,6 +23,7 @@ import org.apache.commons.logging.LogFactory;
 
 import org.springframework.beans.factory.ListableBeanFactory;
 import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.AutoConfigureOrder;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
```diff
@@ -307,12 +309,17 @@ public LocaleContextResolver localeContextResolver() {
 
 		@Bean
 		@ConditionalOnMissingBean(name = WebHttpHandlerBuilder.WEB_SESSION_MANAGER_BEAN_NAME)
-		public WebSessionManager webSessionManager() {
+		public WebSessionManager webSessionManager(
+				@Autowired(required = false) WebSessionIdResolver webSessionIdResolver) {
```
wilkinsona, line 313: For consistency, it would be good to use `ObjectProvider` rather than `@Autowired(required = false)` here. `getIfAvailable()` can then be used to get the provided ID resolver.

wilkinsona (commented):
Thanks very much for the PR, @weixsun. I've left one minor comment that we can take care of when we merge the changes, or you could do it if you prefer.

Would you be interested in updating the PR to add a test for the change? The test would go in [`WebFluxAutoConfigurationTests`](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfigurationTests.java) and use a small `@Configuration` class to define a `WebSessionIdResolver` bean and check that the `WebSessionManager` has been configured to use it.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
weixsun: Done it.

weixsun:
@wilkinsona Thank you for responding so quickly, reviewing the code, and writing simple test cases, but it seems to be enough proof

wilkinsona (approved):
Looks good to me now, @weixsun. Thank you for the updates.

snicoll (approved):

PR #26449, opened 2021-05-13
Use try-with-resources statement in test package

weixsun:
As described in the title. All test cases have been checked.


wilkinsona:
@weixsun Thank you very much for making your first contribution to Spring Boot.

PR #26465, opened 2021-05-14
Remove add-opens workaround in ToolchainPlugin

dreis2211:
Hi,

this PR removes the `--add-opens=jdk.compiler/com.sun.tools.javac.code=ALL-UNNAMED` workaround when using toolchains as https://github.com/gradle/gradle/issues/16126 was backported to Gradle 6.9 which we fortunately use.

Cheers,
Christoph


dreis2211:
For whatever reason the Gradle wrapper validation failed - can't tell why, but I'm pretty sure it's unrelated.

wilkinsona:
I can't tell why either. It passed when it ran a second time but just seemed to fail to run entirely first time round 🤷

wilkinsona:
Thanks very much, @dreis2211.

PR #26476, opened 2021-05-16
Polish SpringApplication

nguyensach:
Reduce nested try block.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
@@ -150,6 +150,7 @@ import org.springframework.util.StringUtils;
  * @author Madhura Bhave
  * @author Brian Clozel
  * @author Ethan Rubinson
+ * @author Nguyen Bao Sach
  * @since 1.0.0
  * @see #run(Class, String[])
  * @see #run(Class[], String[])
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
```diff
@@ -802,22 +803,20 @@ private void callRunner(CommandLineRunner runner, ApplicationArguments args) {
...
+				listeners.failed(context, exception);
 			}
 		}
 		catch (Exception ex) {
 			logger.warn("Unable to close ApplicationContext", ex);
 		}
+		finally {
+			reportFailure(getExceptionReporters(context), exception);
+			if (context != null) {
+				context.close();
```
wilkinsona, line 817: This changes behaviour as any exception from the `close()` call will no longer be caught and logged as a warning.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
```diff
@@ -1367,24 +1366,22 @@ public static int exit(ApplicationContext context, ExitCodeGenerator... exitCode
...
+			if (exitCode != 0) {
+				context.publishEvent(new ExitCodeEvent(context, exitCode));
 			}
 		}
 		catch (Exception ex) {
 			ex.printStackTrace();
 			exitCode = (exitCode != 0) ? exitCode : 1;
 		}
+		finally {
+			close(context);
```
wilkinsona, line 1383: This also changes behaviour as any exception thrown from `close(context)` will no longer be caught.

wilkinsona:
Thanks for the PR, @nguyensach, but both pieces of polish change behaviour and, therefore, should not be made.

PR #26586, opened 2021-05-18
Update documentation following the release of Spock Framework 2.0

dreis2211:
Hi,

yesterday Spock 2.0 was finally released in a non milestone version. This PR updates the dependency management & docs.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211. I tweaked the wording in the documentation a little in https://github.com/spring-projects/spring-boot/commit/88b383c6bb492d0db25bedf7ca088c0459acd53f as, now that Spock 2.0 is GA, I don't think we should encourage people to downgrade Groovy any more.

dreis2211:
I was not sure if I should remove the downgrade hint - thanks for the follow-up.

Issue #3558, opened 2015-07-20
Investigate HCON as a config format


philwebb:
See https://github.com/typesafehub/config/blob/master/HOCON.md

Driven by this conversation https://twitter.com/antonarhipov/status/622860399453577217

fkowal:
+1

philwebb:
We already hit some problems supporting yaml and properties so I'd rather not add another one to the mix. It's possible to define your own `PropertySourceLoader` if you want to support a different format outside of Spring Boot.

bluejaguar:
Phill, please provide a reference to Spring Cloud Config using a custom PropertySourceLoader.  This one here does not work with with Dalton.SR3 (that I can see from trying it) .   https://github.com/spring-projects/spring-boot/issues/4027#issuecomment-144649875

PR #26625, opened 2021-05-21
Add HOCON Property Source

desiderantes:
This is just for testing the idea from #3558 and to see if there's any interest in adopting HOCON as a property file format. HOCON is used a lot in the Scala ecosystem, it's really easy, and a bit more powerful than YAML. It also supports importing other HOCON files natively.

This implementation is taken from https://github.com/zeldigas/spring-hocon-property-source


wilkinsona:
Thanks for the proposal, @desiderantes. The CLA requires that a contribution be your original work so, unfortunately, we cannot accept this PR.

Issue #26638, opened 2021-05-22
MVC controller database connection pool starvation


cdalexndr:
By default, spring's `DefaultJpaDialect` doesn't close the connection, instead it delays until the entity manager is closed:
https://github.com/spring-projects/spring-framework/blob/ac0e71745b1de86e293382cce64e94f04e0ab6af/spring-orm/src/main/java/org/springframework/orm/jpa/DefaultJpaDialect.java#L112-L114

Issuing a web request to a controller, uses `OpenEntityManagerInViewInterceptor` who creates an entityManager that lives the whole request.

Thus, any connection opened inside a web request will remain open until the request is finished, potentially leading to starvation, unable to acquire new connections.

### A real life example:
A web request starts an async process that will compute some results, and waits for it to finish. 
The async process requires a db connection for computation.
All web requests on the same endpoint share and wait on the same async process to compute results.
If all requests are issued concurrently before the async process starts, the requests will acquire all the pool db connections, so the process will not be able to acquire a connection for the computation, and will timeout.

Sample repo: https://github.com/cdalexndr/spring-boot-issue-26638

Related: https://github.com/spring-projects/spring-boot/issues/7107

**Spring Boot 2.4.4**

cdalexndr:
Workaround: `spring.jpa.open-in-view=false`

snicoll:
Thanks for the report but that's essentially a duplicate of the discussion in #7107 where we've discussed the pros and cons of enabling this behavior by default or not. What you call a workaround is actually what you should be doing and the reason why the warning is issued.

cdalexndr:
@snicoll in #7107 the opinions are performance based:

@odrotbohm [says](https://github.com/spring-projects/spring-boot/issues/7107#issuecomment-260633493):
> So I'd love to learn what's wrong with the default setting?

also [this](https://github.com/spring-projects/spring-boot/issues/7107#issuecomment-260813410):
> Still, there's nothing that inherently breaks code when an O(S|E)IVF is in place.

This issue proves that enabling OSIV can actually break code.

With the fact that it's also an anti-pattern, I think it adds sufficient validation to tilt the scale balance in favor of removing the default enabled OSIV. It should only be used in dev environment, or experimental setups...

snicoll:
> The current issue proves that enabling OSIV can actually break code.

[...]

PR #26659, opened 2021-05-25
OSIV connection pool starvation warning

cdalexndr:
Issue https://github.com/spring-projects/spring-boot/issues/26638


philwebb:
Thanks for the pull-request but the `OpenEntityManagerInViewInterceptor` configuration is quite a contentious issue and there are many things to consider when enabling/disabling it. I'm not keen to add this specific warning since I don't want to over complicate the documentation with when/if things are a good idea.

PR #26664, opened 2021-05-25
Update Jersey to 2.34

DieBauer:
Address CVE-2021-28168

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -849,7 +849,7 @@ bom {
 			]
 		}
 	}
-	library("Jersey", "2.33") {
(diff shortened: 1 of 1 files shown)


DieBauer:
Thanks for the pointer. I created this PR because of the listed CVE, and the fact that jersey 2.34 is out since more than 5 weeks. https://github.com/eclipse-ee4j/jersey/releases/tag/2.34

At what interval is the project updating 3rd party dependencies?

philwebb:
@DieBauer We cover a bit of this on the wiki https://github.com/spring-projects/spring-boot/wiki/Supported-Versions#third-party-dependencies. We'd usually expect patch releases of third-party dependencies to be made available to solve CVEs, but it seems like Jersey aren't planning a `2.33.1` version.

Issue #26688, opened 2021-05-28
Add SpringApplicationBuilder constructor with ResourceLoader parameter


jaminh:
It would be helpful to have a consturctor for the SpringApplicationBuilder that allows passing a custom ResourceLoader when a different classloader needs to be used for the SpringApplication. There is a method for setting the ResourceLoader after construction but this means the spring factories instances have already been initialized with the default classloader rather than the classloader from the ResourceLoader.

The issue https://github.com/spring-projects/spring-boot/issues/26126 presents a likely scenario where this might be desired.

snicoll:
As I've indicated on the other issue, we don't need a PR and an issue for the same thing. Considering that you've submitted a proposal, let's continue the discussion there (#26690)

PR #26690, opened 2021-05-28
Add SpringApplicationBuilder constructor with ResourceLoader parameter

jaminh:
https://github.com/spring-projects/spring-boot/issues/26688

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java
@@ -97,9 +97,9 @@ public class SpringApplicationBuilder {
 		this.application = createSpringApplication(sources);
 	}
 
-    public SpringApplicationBuilder(ResourceLoader resourceLoader, Class<?>... sources) {
-            this.application = createSpringApplication(resourceLoader, sources);
-    }
+	public SpringApplicationBuilder(ResourceLoader resourceLoader, Class<?>... sources) {
+		this.application = createSpringApplication(resourceLoader, sources);
+	}
 
 	/**
 	 * Creates a new {@link org.springframework.boot.SpringApplication} instances from the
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java
```diff
@@ -109,6 +113,19 @@ protected SpringApplication createSpringApplication(Class<?>... sources) {
...
 	}
 
+	/**
+	 * Creates a new {@link org.springframework.boot.SpringApplication} instances from the
+	 * given sources. Subclasses may override in order to provide a custom subclass of
+	 * {@link org.springframework.boot.SpringApplication}
+	 * @param resourceLoader the resource loader
+	 * @param sources the sources
+	 * @return the {@link org.springframework.boot.SpringApplication} instance
+	 * @since 1.1.0
```
snicoll, line 123: That since tag is wrong.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java
```diff
@@ -109,6 +113,19 @@ protected SpringApplication createSpringApplication(Class<?>... sources) {
...
+	 */
+	protected SpringApplication createSpringApplication(ResourceLoader resourceLoader, Class<?>... sources) {
```
snicoll, line 125: If we decided to pursue this, we should deprecate the other protected `createSpringApplication` in favor of this one.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java
```diff
@@ -97,6 +97,10 @@ public SpringApplicationBuilder(Class<?>... sources) {
 		this.application = createSpringApplication(sources);
 	}
 
+	public SpringApplicationBuilder(ResourceLoader resourceLoader, Class<?>... sources) {
```
snicoll, line 100: The other constructor should call this constructor with a `null` `ResourceLoader` (to mirror what  we do in `SpringApplication`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java
```diff
@@ -109,6 +113,19 @@ protected SpringApplication createSpringApplication(Class<?>... sources) {
 		return new SpringApplication(sources);
 	}
 
+	/**
+	 * Creates a new {@link org.springframework.boot.SpringApplication} instances from the
+	 * given sources. Subclasses may override in order to provide a custom subclass of
+	 * {@link org.springframework.boot.SpringApplication}
+	 * @param resourceLoader the resource loader
```
snicoll, line 120: We should indicate that it could be `null`.

snicoll (commented):
Thanks for the PR. This looks sensible to me but there are a number of changes necessary and an update of the tests suite is required before we could consider merging this.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java
jaminh: Fixed

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/builder/SpringApplicationBuilder.java
jaminh: Fixed

snicoll (approved):

Issue #26126, opened 2021-04-15
ConfigFileApplicationListener#loadPostProcessors can not load EnvironmentPostProcessor in child classLoader


ChenKangQiang:
My spring-boot version is 2.3.3.RELEASE. I want use spring.factories to load EnvironmentPostProcessor.
In my project, I haven a parent-child classLoader structure. ConfigFileApplicationListener.class is loaded by parent classLoader, and my environmentPostProcessorImpl is in child classloader path. The source code show that ConfigFileApplicationListener use its classloader to load EnvironmentPostProcessor. So，it can not load.
![image](https://user-images.githubusercontent.com/14541768/114815729-64433900-9de9-11eb-890d-4555934ac338.png)
With Java SPI，it use ThreadContextClassLoader to load implementation class. Why Spring do not use ThreadContextClassLoader? Or is there any other solutions.

snicoll:
@ChenKangQiang I don't really understand what you're doing or why the child context is necessary. An application has a designated `ClassLoader` and loads its resources from it. If you create a child `ClassLoader`, the app classloader is not supposed to have access to that. These are general principles that are not linked to Spring Boot.

If you want support, you need to share more about your setup in the form of a small sample that shows the issue. You can do so by attaching a zip to this issue or sharing a GitHub repo url. Thanks.

ChenKangQiang:
@snicoll We design such a structure，be similar to Tomcat，but follow the parental delegation mechanism.
![image](https://user-images.githubusercontent.com/14541768/114836010-c6f5fe00-9e04-11eb-86d5-5fa5b3db830a.png)

We use a ``SharedClassLoader`` to load shared spring class in order to reduce ``jvm matespace``，and ``FunctionClassLoader`` to load function. We use different thread to load every function，and  every function has own ApplicationContext with different beans. so, every function has its own spring.factories.

wilkinsona:
`SpringApplication` loads all `ApplicationListener`s, including `ConfigFileApplicationListener`, via `spring.factories` using its resource loader's class loader. If it has no resource loader, it'll use `ClassUtils.getDefaultClassLoader()` which is typically the thread context class loader. `ConfigFileApplicationListener` then uses its own `ClassLoader` to load `EnvironmentPostProcessor`s. This feels a bit inconsistent to me, although using `getClass().getClassLoader()` is a common pattern with `SpringFactoriesLoader` elsewhere in Spring Boot and in other projects including Spring Framework.

[...]

Issue #26689, opened 2021-05-28
Use resource loader's class loader in config loading


jaminh:
If possible back patch the changes for https://github.com/spring-projects/spring-boot/issues/26126 to older versions of spring-boot (2.4.x, 2.3.x, ...).

snicoll:
@jaminh going forward please do not open a PR and an issue for the same thing. We don't need PRs for backports either. The fix in `2.5.x` only was intentional.

PR #26691, opened 2021-05-28
Use resource loader's class loader in config loading

jaminh:
Previously, classes involved in config loading used a variety of
potentially different class loaders when calling SpringFactoriesLoader.
Some classes would use their own class loader and others would use null
which results in SpringFactoriesLoader's class loader being used.

This commit updates the config loading classes to consistently use the
resource loader's class loader.

Fixes gh-26126
Back patch for https://github.com/spring-projects/spring-boot/issues/26689


snicoll:
Thanks for the PR but we intentionally decided to only fix this in `2.5.x`.

PR #26714, opened 2021-05-30
Add more session cookie config property for WebFlux

weixsun:
## Add rich session cookie config property for WebFlux

**when spring-session is not integrated**

- [x] Provide `spring.webflux.session.cookie.*` properties.
- [x] Provide `spring.webflux.session.timeout` properties.

**when spring-session is on classpath(On the way)**

- [ ] `spring.webflux.session.cookie.*` properties adaptation spring-session
- [ ] `spring.webflux.session.timeout` and `spring.session.timeout` are equivalent

  > However, spring.session.timeout has a higher priority.

If the core members of springboot feel that this pr is worth continuing, I will set it to draft status and complete the rest of the work, otherwise it will be a bankrupt pr.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
@@ -39,8 +39,11 @@ import org.springframework.boot.autoconfigure.web.WebProperties;
 import org.springframework.boot.autoconfigure.web.WebProperties.Resources;
 import org.springframework.boot.autoconfigure.web.format.DateTimeFormatters;
 import org.springframework.boot.autoconfigure.web.format.WebConversionService;
+import org.springframework.boot.autoconfigure.web.reactive.WebFluxProperties.Cookie;
 import org.springframework.boot.autoconfigure.web.reactive.WebFluxProperties.Format;
+import org.springframework.boot.autoconfigure.web.reactive.WebFluxProperties.SameSite;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.boot.convert.ApplicationConversionService;
 import org.springframework.boot.web.codec.CodecCustomizer;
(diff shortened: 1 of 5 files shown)


snicoll:
@weixsun thanks for the PR. I am curious, what lead you to submit this? Are you using this in your own project?

weixsun:
@snicoll Thank you for your attention.

> what lead you to submit this?

1. You should have noticed that this pr is closely related to #26437. It is not necessary to set the session separately and repeatedly in the code of each project every time (Of course, if you like, you can also set the session in an encoding way, which is the reason for the existence of #26437)
2. If you don't integrate spring-session and want to set the session expiration time, it is currently not supported and can only endure the [default value of 30m](https://github.com/spring-projects/spring-framework/blob/e0690335fcfb91f12e4412d8aef0ea669b7b6104/spring-web/src/main/java/org/springframework/web/server/session/InMemoryWebSessionStore.java#L180), unless the developer does work similar to this pr.

> Are you using this in your own project?

Yes, this pr is derived from the needs of a microservice module in our actual project. This microservice module does not need to integrate spring-session in its architectural design, so when you want to set the session expiration time and cookie-related attributes , That the way of providing configuration is better than the way of encoding.

The advantages of adding these configurations must outweigh the disadvantages. It is not good at words and can only describe the general meaning, but the final result is still determined by experienced spring-boot members.

snicoll:
I've rebased and squashed all the commits into one so @weixsun please do not push further update for now. 

I was looking at the current arrangement where we have `server.servlet.session.cookie.*` for MVC. This PR adds a number of items that can be honored using webflux under `spring.webflux.session.cookie.*`, including a `spring.webflux.session.timeout` that aligns with `server.servlet.session.timeout`. 

Before we start duplicating these properties, I'd like to reach out to the team again and verify this is what we want to do.

DidierLoiseau:
Am I right that the new properties are not documented? I can see that [`server.reactive.session.timeout`](https://docs.spring.io/spring-boot/docs/current/reference/html/application-properties.html#application-properties.server.server.reactive.session.timeout) is documented but not any of the `server.reactive.session.cookie` ones.

I could only find [a small mention of the new properties in the SB 2.6 Release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.6-Release-Notes#reactive-server-session-properties).

wilkinsona:
You are right, @DidierLoiseau. Thanks for spotting this. I've opened #31896.

PR #26737, opened 2021-06-02
Remove reference to surefirebooter JAR

dreis2211:
Hi,

I just noticed this remaining reference to the `surefirebooter` JAR that is effectively not needed since we're not using Maven Surefire anymore.

Cheers,
Christoph


snicoll:
Thanks again Christoph.

PR #26751, opened 2021-06-03
Fix typo in 'Reacting to the Maven plugin' title

abelsromero:
Fixes a type in the section markup of `Reacting to the Maven plugin`.
This broke the AsciiDoc syntax with the following result in the HTML published version.
Instead of a title and admonitions we get a paragraf with all together.

image


snicoll:
Good catch, thank you!

Issue #26767, opened 2021-06-04
Document support for Java 17


dreis2211:
Hi,

since JDK 17 is around the corner, I want to raise this issue to talk about the necessary steps for supporting JDK 17. I would again volunteer to help out here.

Given the impact of https://openjdk.java.net/jeps/403 we should probably start a little earlier than usual as we might need to reach out to libraries that Boot is integrating with.

Known issues so far from the top of my head - in no particular order:

1. `JsonbTesterTests.writeArrayShouldReturnJsonContent()` fails due to [JOHNZON-347](https://issues.apache.org/jira/browse/JOHNZON-347)
2. JavaVersion should get a new entry `SEVENTEEN` - see #26769
3. Multi-Release Jar tests fail - see #27229
4. Gradle Plugin Tests fail - as usual for newer JDKs. See https://github.com/gradle/gradle/issues/16857
5. CLI tests are failing because it's impossible to clear `URL.factory` via reflection - possible workaround would be to open `java.net` to the unnamed module. I don't see a cooler way here at the moment. Ideally `TomcatURLStreamHandlerFactory.disable()` would work, but my first tests didn't succeed. See #26930
6. Similar problem as 5. but in `AbstractServletWebServerFactoryTests.tearDown()` - Possibly similar solution with opening modules. See #27089
7. `JarFileTests.jarFileEntryWithEpochTimeOfZeroShouldNotFail()` fails because it tries to set `xdostime` on ZipEntry, which is internal. We could disable that test for JDK 17 and higher or again fiddle around with opening up modules. See #27100 
8. Two tests in `RabbitAutoConfigurationTests` fail because they are trying to access the `contextSpi` field on `SSLContext`. See #26954
9. Generally we can't use `--illegal-access=warn` anymore. The option will be ignored, so it needs to be removed. See #27089
10. `ConditionalOnSingleCandidateTests.singleCandidateOneScopedProxyCandidate()` fails because of the following. Haven't looked deep enough yet to see who should do what here.
```
 Caused by: java.lang.IncompatibleClassChangeError: class jdk.proxy3.$Proxy583 cannot implement sealed interface java.lang.constant.ConstantDesc
 	at java.base/java.lang.ClassLoader.defineClass1(Native Method)
 	at java.base/java.lang.System$2.defineClass(System.java:2259)
 	at java.base/java.lang.reflect.Proxy$ProxyBuilder.defineProxyClass(Proxy.java:560)
[...]

PR #26769, opened 2021-06-05
Add Java 17 to JavaVersion enum

dreis2211:
Hi,

this is the first tiny step for #26767 by adding Java 17 to the known versions.

Cheers,
Christoph


snicoll (commented):
I'd be in favor of adding this once we have a better view on what's required to have a green build with Java 17. WDYT?

dreis2211:
The class comment of `JavaVersion` says `Known Java versions.` and not `Supported Java Versions`. I always saw it similar to Gradle's JavaVersion enum that includes 17 on version 6.9 while not supporting it.

I don't have strong feelings in holding this back until we have a better picture for Java 17, but for me `JavaVersion` is not really connected to either running or supporting the respective version. Rather an acknowledgement of the version - nothing more nothing less. But in the end it's totally fine if this PR is staying open for a while - it's not really needed to get a green build for Java 17 anyway.

wilkinsona:

PR #26775, opened 2021-06-06
Upgrade to Json Path 2.6.0

strehle:
fix CVE-2021-27568


snicoll:
@strehle thanks for the PR but as indicated in the template, we prefer to do such dependency upgrades ourselves. In the meantime you can override the version of `json-path` using the `json-path.version`.

Issue #26811, opened 2021-06-08
parse http headers in case insensitive manner


akosyakov:
HTTP headers are case insensitive by the spec. Particularly, we are trying to run spring-boot behind caddy (golang based reverse proxy) and stumbled upon https://github.com/gitpod-io/gitpod/issues/4047#issuecomment-856566526

francislavoie:
In particular, this regexp needs to be changed to be case insensitive: https://github.com/spring-projects/spring-boot/blob/4c71e8124d99c12dfaa2e342b12355bd0954cb8b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/livereload/Connection.java#L44

wilkinsona:
Closing in favour of #26813.

PR #26813, opened 2021-06-08
DevTools' LiveReload support's HTTP header handling is case sensitive

francislavoie:
Attempted fix for https://github.com/spring-projects/spring-boot/issues/26811

The livereload functionality should not be relying on specific case sensitivity of header fields to work. HTTP header field names are case insensitive.

Disclaimer: I don't use this project, I just made these changes directly in GitHub because it seems like a simple fix, from a report we got on https://github.com/caddyserver/caddy.

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/livereload/Connection.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/livereload/Connection.java
@@ -41,7 +41,7 @@ class Connection {
 
 	private static final Log logger = LogFactory.getLog(Connection.class);
 
-	private static final Pattern WEBSOCKET_KEY_PATTERN = Pattern.compile("^Sec-WebSocket-Key:(.*)$", Pattern.MULTILINE);
+	private static final Pattern WEBSOCKET_KEY_PATTERN = Pattern.compile("^Sec-WebSocket-Key:(.*)$", Pattern.MULTILINE | Pattern.CASE_INSENSITIVE);
 
 	public static final String WEBSOCKET_GUID = "258EAFA5-E914-47DA-95CA-C5AB0DC85B11";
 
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/livereload/Connection.java
```diff
@@ -77,7 +77,7 @@
 	 * @throws Exception in case of errors
 	 */
 	void run() throws Exception {
-		if (this.header.contains("Upgrade: websocket") && this.header.contains("Sec-WebSocket-Version: 13")) {
+		if (this.header.contains("Upgrade: websocket") && this.header.toLowerCase().contains("sec-websocket-version: 13")) {
```
weixsun, line 80: Maybe we can maintain the minimum changes, so here we can go back to the original state.

francislavoie: I'm not sure I understand. Do you mean to do `"Sec-WebSocket-Version: 13".toLowerCase()` instead?

weixsun: I mean revert this line. It seems superfluous to me, doesn't it?

francislavoie: I don't think so. Proxies may pass the header with the casing `Sec-Websocket-Version` (lowercase `s` instead of uppercase `S`) so this comparison would fail.

weixsun: I agree with what you said, I get it.

philwebb:
Thanks very much for the fix @francislavoie! This has been merged to 2.4.x, 2.5.x and main.

PR #26861, opened 2021-06-11
Update some more references to 2.5.x following work starting on 2.6

dreis2211:
Hi,

I just noticed that the work on 2.6.x has started, but noticed that a few places were missed. Feel free to close this, if you have this already locally, but simply not pushed yet.

Cheers,
Christoph


git/hooks/prepare-forward-merge
```diff
@@ -4,7 +4,7 @@ require 'net/http'
 require 'yaml'
 require 'logger'
 
-$main_branch = "2.5.x"
+$main_branch = "2.6.x"
```
dreis2211, line 7: This was done in the past as well. I must admit, I'm not sure if this is correct.

wilkinsona: Thanks, @dreis2211. Your change is correct. Without it a forward merge into `main` would say `2.5.x` in its message rather than `2.6.x`. This would then cause the `forward-merge` hook to match against the wrong milestone on GitHub so the issue it creates would be in the wrong version.

dreis2211: Ah, it's matching the milestone - that makes sense. Thanks for clarifying.

Issue #26878, opened 2021-06-12
Spring-CLI: "packageName" parameter is not aligned with the docs and doesn't fit the naming convention of other parameters


skryvets:
### Steps to reproduce:
1. In the command line, run `spring init --list`. As a result a list of dependencies and parameters will appear in the console:
```
..... 
(omitted for brevity)
....

Parameters
+-------------+------------------------------------------+------------------------------+
| Id          | Description                              | Default value                |
+-------------+------------------------------------------+------------------------------+
| artifactId  | project coordinates (infer archive name) | demo                         |
| bootVersion | spring boot version                      | 2.5.1                        |
| description | project description                      | Demo project for Spring Boot |
| groupId     | project coordinates                      | com.example                  |
| javaVersion | language level                           | 11                           |
| language    | programming language                     | java                         |
| name        | project name (infer application name)    | demo                         |
| packageName | root package                             | com.example.demo             |
| packaging   | project packaging                        | jar                          |
| type        | project type                             | maven-project                |
| version     | project version                          | 0.0.1-SNAPSHOT               |
+-------------+------------------------------------------+------------------------------+
```

2. Using the suggested docs, try to run init a spring app with a custom package name, e.g. : 
```
spring init --packageName=com.skryvets.demo spring-demo
```

### Expected result:

Spring app should be created. The following (or similar output) should be produced in the console:

```
Using service at https://start.spring.io
Project extracted to '/yourfolder/spring-demo'
```

### Actual result:

Spring app creation fails with the following message: `packageName is not a recognized option`

### Notes:
The reason I put "doesn't fit the naming convention of other parameters" in the description as it seems like all other options are "camel cased" whereas this particular one is dash separated. There are 2 way of fixing - either changing the docs or parameter name. I felt like changing the parameter is more appropriate in this case.

skryvets:
This is my first issue as well as PR that I submitted here: https://github.com/spring-projects/spring-boot/pull/26879. Please correct me if I wrong.

snicoll:
> The reason I put "doesn't fit the naming convention of other parameters" in the description as it seems like all other options are "camel cased" whereas this particular one is dash separated.

`java-version`, `boot-version` are also camel case to follow the  convention of the CLI. If you run `spring help init` you'll see how you can build a command (either with a shorthand, i.e. `-p` for the package name or the full name). The thing that is a bit confusing here is that the server gives you back a number of "parameters" that it supports primarily to tell you about default values.

[...]

PR #26879, opened 2021-06-12
Rename "package name" cli parameter

skryvets:
Address #26878


snicoll:
Thanks for the PR but we can't just change the attribute this way. Let's continue the discussion on #26878

PR #26903, opened 2021-06-14
Move away from Gradle's ConfigureUtil in buildSrc

dreis2211:
Hi,

Gradle 7.1 deprecated or moved some classes in `org.gradle.util`. One of those being `ConfigureUtil` that is used in both `buildSrc` and the Gradle Plugin. This PR updates at least `buildSrc` already. I guess the Gradle Plugin usages need to be handled as well somewhen, but are less straight forward.

Cheers,
Christoph

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/bom/BomExtension.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/bom/BomExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -51,7 +51,6 @@ import org.gradle.api.plugins.JavaPlatformPlugin;
 import org.gradle.api.publish.maven.tasks.GenerateMavenPom;
 import org.gradle.api.tasks.Sync;
 import org.gradle.api.tasks.TaskExecutionException;
-import org.gradle.util.ConfigureUtil;
 import org.w3c.dom.Document;
 import org.w3c.dom.NodeList;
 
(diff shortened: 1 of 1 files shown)


ciscoo:
I think switching from `Closure` to `Action` would be a cleaner approach. Cleaner in the sense that it will avoid passing around the `Project` instance just to use its `configure` method.

```java
public void upgrade(Closure<?> closure) {
    ConfigureUtil.configure(closure, this.upgradeHandler);
}

// to

public void upgrade(Action<? super UpgradeHandler> action) {
    action.execute(this.upgradeHandler);
}
```

dreis2211:
@ciscoo Theoretically yes - in fact I've tried this already according to the following comment https://github.com/gradle/gradle/issues/17446#issuecomment-861121174. Practically though, this is the only place in BomExtension where this works out of the box and where I had the project available anyway. So we don't win much to be honest. Changing this throughout - e.g. in UpgradeHandler
```
public void gitHub(Action<GitHubHandler> action) {
	action.execute(this.gitHub);
}
```
fails with the following exception:
```
Caused by: groovy.lang.MissingPropertyException: Could not set unknown property 'issueLabels' for object of type org.springframework.boot.build.bom.BomExtension$UpgradeHandler.
```
I'm not a Groovy expert to be entirely honest, so I might be missing something very stupid.

If the team wants to have this changed I'm open to this, but the change at hand seemed to be more consistent (and actually working with a fairly simple approach).

Cheers

wilkinsona:
I believe that it doesn't work due to the nesting. `BomExtension` is known to Gradle which gives it a chance to augment the type with methods that take a `Closure` and configure its delegate. The `UpgradeHandler` instance is hidden from Gradle so that augmentation isn't performed which leaves it using Groovy's standard type coercion which doesn't set the delegate.

It may be possible to get Gradle to create the `UpgradeHandler` instance so that it can augment it, but `Action` isn't a drop-in replacement for `Closure` and the use of `ConfigureUtil`.

dreis2211:
@wilkinsona Do you want me to change `BomExtension.upgrade` or leave things as-is?

wilkinsona:
Thanks for the offer, @dreis2211, and for bringing this to our attention. After some experimentation, I have concluded that we should take a different approach. If we use Gradle's `ObjectFactory` to create instances of the classes with `Closure` methods, we can then change those methods to take an `Action` instead and still get the desired delegation behaviour when called with a `Closure`. I've implemented that in https://github.com/spring-projects/spring-boot/commit/0f52bbc5607ad91fcd3fa2791b2130c5c0d4ee96.

PR #26910, opened 2021-06-15
Add Azure info to the cloud deployment docs

aclement:
Let me know if I'm missing any resources that additionally should be editing as I'm not an expert on the docs structure.


snicoll:
Thanks @aclement. There's no need to patch the url rewrite for new sections, this is meant to provide backward compatible anchors for existing sections that used to be identified differently before 2.5.x.

PR #26937, opened 2021-06-17
Update reference documentation to reflect @SpringBootApplication's use of @SpringBootConfiguration

quaff:


snicoll:
Thanks for the PR but I think the use of `@Configuration` here is deliberate as we may not want to get into the details of what `@SpringBootConfiguration` does compared to the regular `@Configuration`. I've flagged for team attention to get more feedback.

wilkinsona:
To me, the current documentation reads as if you could replace `@SpringBootApplication` with `@EnableAutoConfiguration`, `@ComponentScan`, and `@Configuration` and get the same behaviour. That's not the case as using `@Configuration` rather than `@SpringBootConfiguration` may break some of your tests. With this in mind, I think the proposed changes are an improvement. We could perhaps go a small step further and cross-link to the testing section that discusses `@SpringBootConfiguration`.

snicoll:
I've re-read the section and we're already mentioning quite advanced details about `@ComponentScan`. That nullifies my previous argument.

wilkinsona:
Thanks very much, @quaff.

Issue #24174, opened 2020-11-17
Provide out-of-the-box support for Elasticsearch Sniffer


izeye:
It would be nice if Spring Boot provides out-of-the-box support for Elasticsearch `Sniffer`.

See https://www.elastic.co/guide/en/elasticsearch/client/java-rest/current/_usage.html

somayaj:
Hi, can this be worked on? Also, is this an integration of the Sniffer into the spring boot libs?

snicoll:
@somayaj yes, it's flagged ideal for contribution. The link shows some usage of the library and this issue is about providing auto-configuration for regular use cases, potentially coupled with configuration properties to tune things as we usually do.

scottfrederick:
Closing in favor of #24340

PR #26949, opened 2021-06-17
Add possibility to disable the Elasticsearch Sniffer functionality

filiphr:
With #24174 out-of-the-box support was added for Elasticsearch Sniffer.

After our upgrade to the latest Spring Boot 2.5.1, we noticed that using Elasticsearch Sniffer with Elasticsearch in Docker was not working correctly (at least not on Mac). The resolved nodes had some IP that could not be accessed from my host. I know that the functionality is there only if the Sniffer is on the classpath (it was for us by mistake). 

In my opinion it could be beneficial to enable / disable the functionality using a property as well. This can be interested for application build on top of Spring Boot that would like to offer an opt in to the Sniffer by enabling the flag.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientConfigurations.java
@@ -36,6 +36,7 @@ import org.elasticsearch.client.sniff.SnifferBuilder;
 import org.springframework.beans.factory.ObjectProvider;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnSingleCandidate;
 import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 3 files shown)


philwebb:
Looking at the auto-configuration it doesn’t seem like there’s currently an easy way to exclude the sniffer whilst keeping everything else. I think another way to disable it would be good. That could be a property, or we could extract a new auto-configuration class.

filiphr:
Using a different AutoConfiguration class would work as well. We already have a mechanism that uses `AutoConfigurationImportFilter` to enable / disable certain auto configurations based on properties.  Of course having a simple property directly from Spring Boot is the easiest. 

I created this PR since it was easy to do it and I thought I should just do it and we can discuss in the PR, without the need to go through issues.

philwebb:
We discussed this today as team and we'd currently prefer to leave things as they are. It feels like having and Elasticsearch Sniffer dependency and not enabling it shouldn't be that common. We've also been trying to reduce the number of `.enabled` properties.

If more people comment on this issue, or we find a good reason why the sniffer dependency needs to be present but not used, we can reconsider.

philwebb:
Regardless, thanks a lot for the PR @filiphr!

carofe82:
We are having the same issue after updating Spring. We use sniffer in our testing and production environments, however, our developers have a shared environment they connect to to be able to access a large amount of testing data. This environment is behind an ingress proxy, so developers used to have the sniffer off. After updating spring boot, we are having the issue described in the original post.

It would be great to be able to separate the sniffer mechanism from the policy of enabling it or not. The configuration proposed here would be the best approach. Preventing the sniffer from showing up in the classpath in one environment and not in others is not a feasible option.

filiphr:
@carofe82 there is a way to achieve what you are looking for without the need for a change in Spring Boot. It is a nasty and in my opinion a hacky approach but it it works. Have a look at https://github.com/spring-projects/spring-boot/pull/28496#issuecomment-965579391 for a hint for how to achieve it.

carofe82:
@filiphr ,
Thank you for the recommendation. At the end, we have decided to just remove sniffer all together from classpath, and instead use other techniques in Production.

PR #26954, opened 2021-06-17
Avoid illegal reflective access in RabbitAutoConfigurationTests

dreis2211:
Hi,

this is an attempt to avoid some illegal reflective access in `RabbitAutoConfigurationTests`. The idea is that rather than using reflection in order to get down to the used `TrustManager` we look for output that `TrustEverythingTrustManager` is producing when in use. It's not pretty, but it works and is reflection free. It's also the least invasive option without opening modules and/or changing the production code that I could find.

As this is part of the overall JDK 17 story under #26767, it should probably go into 2.5.x.

Let me know what you think.
Cheers,
Christoph


snicoll:
Thanks @dreis2211, I am not a big fan of asserting the console output but, looking at the previous version, I think it's an acceptable trade-off.

dreis2211:
Me neither, @snicoll . But thanks for accepting it.

PR #26955, opened 2021-06-18
Add missing space for asciidoc ordered list

quaff:
The list is not rendering before this commit


snicoll:
@quaff good catch! Thank you for making your first contribution to Spring Boot.

Issue #14338, opened 2018-09-06
Configure memory leak detection for Netty


bclozel:
We should add a property for configuring the level of leak detection with Netty.

This can be done statically, has obviously a performance cost, but is really useful if an application is dealing with `DataBuffer` instances or extensive customizations on the Netty engine.

This applies to both server and client use cases, and Netty offers an API for that:

```java
ResourceLeakDetector.setLevel(Level.PARANOID);
```

We could also consider enabling this by default when the devtools are activated.

bclozel:
Closing as there isn't much demand for this and the JVM argument might still be the best way to achieve that.

JamesChenX:
@bclozel  Is it possible to reopen this ticket? Our project wants to set the level to ADVANCED in the development environment and SIMPLE in the production environment according to the profile in application.yaml.

bclozel:
@JamesChenX Thanks for the comment, we'll reconsider this. This is indeed a good use case.

@violetagg I'm wondering if this is enough, or if we should make this feature more general. When dealing with memory leak issues with Netty, are there other options to enable? Typically, when working on a memory leak report - what would you ask developers to enable in their applications to collect useful information? Are specific logs enabled?

I'm trying to figure out if changing the Netty flag is enough or if we can do something more here. Thanks!

wilkinsona:
Should this be assigned to 2.5.x rather than 2.4.x? If is is an enhancement then I don't think it belongs in 2.4.x.

bclozel:
Yes, wrong milestone indeed!

PR #26958, opened 2021-06-18
Default value of spring.netty.leak-detection does not match Netty's default

lfz757077613:
I use netty to build a tcp server from springboot 2.4.x to 2.5.x and I found the project does not detect memory leak any more.
The NettyAutoConfiguration was added from 2.5.x, and the LeakDetection of NettyProperties  is disable.
I think it shoud keep same with netty default ResourceLeakDetector level, otherwise the netty debug log shows its ResourceLeakDetector level is simple, but it is disable actually.
In most situations, we need the default simple ResourceLeakDetector  and it works well
By the way, NettyAutoConfiguration seems useless for now, because it only set ResourceLeakDetector. Maybe it will provide more properties in the future？

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/netty/NettyAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/netty/NettyAutoConfiguration.java
@@ -37,8 +37,7 @@ public class NettyAutoConfiguration {
 
 	public NettyAutoConfiguration(NettyProperties properties) {
 		if (properties.getLeakDetection() != null) {
-			NettyProperties.LeakDetection leakDetection = properties.getLeakDetection();
-			ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.valueOf(leakDetection.name()));
+			ResourceLeakDetector.setLevel(properties.getLeakDetection());
 		}
 	}
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/netty/NettyProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/netty/NettyProperties.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.autoconfigure.netty;
 
+import io.netty.util.ResourceLeakDetector;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 
 /**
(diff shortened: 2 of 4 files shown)


snicoll:
Thanks for the suggestion. It looks like either the default has changed or the default we've set in #14338 was the wrong one.

bclozel:
We've set it to the wrong level originally. I'm scheduling this for 2.6.0.

lfz757077613:
I agree with it 100%. At least for me, I choose to exclude NettyAutoConfiguration unless it can provide more properties.
image

bclozel:
@lfz757077613 Thanks for raising this issue. I'm not sure I understand your concern here - besides working around this particular issue (to be fixed in 2.5.2), why would you want to exclude this auto-configuration?

lfz757077613:
> @lfz757077613 Thanks for raising this issue. I'm not sure I understand your concern here - besides working around this particular issue (to be fixed in 2.5.2), why would you want to exclude this auto-configuration?

Like you commented on 27 oct 2018, it effect "everything netty". The jvm argument maybe the best practice for the global setting such like it.
And NettyAutoConfiguration  make users who both use springboot and netty confused, such as me.
When I found my project did not detect memory leak any more, I checked the netty debug log  at the first time. It really make me confused when  I saw the log like below. Maybe it's better to add a log in NettyAutoConfiguration
```
- - 2021-06-17 23:44:33.928 DEBUG 17072 --- [           main] io.netty.util.ResourceLeakDetector       : -Dio.netty.leakDetection.level: simple
- - 2021-06-17 23:44:33.928 DEBUG 17072 --- [           main] io.netty.util.ResourceLeakDetector       : -Dio.netty.leakDetection.targetRecords: 4
- - 2021-06-17 23:44:33.933 DEBUG 17072 --- [           main] i.n.util.ResourceLeakDetectorFactory     : Loaded default ResourceLeakDetector: io.netty.util.ResourceLeakDetector@7e91ed74
```
Even through I unify the ResourceLeakDetector level setting by using NettyAutoConfiguration, I also need to consider the lifecircle between spring container and other component which also set ResourceLeakDetector level. So I prefer to control the ResourceLeakDetector level by myself.
But I'd like to use it  if NettyAutoConfiguration can provide other settings like BufAllocator, etc in the future
Thanks for reading my large text above 😅

bclozel:
Thanks for your comment @lfz757077613  - so you were previously setting that value yourself and you were surprised that Spring Boot is setting it as well.

Maybe we should change the current situation so that we don't override a non-default value? Or maybe we should set a `null` value as a default and only change it if the developer provides one?

Maybe we could add logging statements [using the same logger](https://github.com/netty/netty/blob/d34212439068091bcec29a8fad4df82f0a82c638/common/src/main/java/io/netty/util/ResourceLeakDetector.java#L128-L130) to show that the level has been changed by the application?

As for additional properties, we might add new ones if the community makes a good case for them.

lfz757077613:
> Thanks for your comment @lfz757077613 - so you were previously setting that value yourself and you were surprised that Spring Boot is setting it as well.
> 
> Maybe we should change the current situation so that we don't override a non-default value? Or maybe we should set a `null` value as a default and only change it if the developer provides one?
> 
> Maybe we could add logging statements [using the same logger](https://github.com/netty/netty/blob/d34212439068091bcec29a8fad4df82f0a82c638/common/src/main/java/io/netty/util/ResourceLeakDetector.java#L128-L130) to show that the level has been changed by the application?
> 
> As for additional properties, we might add new ones if the community makes a good case for them.

I think default "null" is a good idea, I have modify this pr and add test in NettyAutoConfigurationTests and fix additional-spring-configuration-metadata.json

bclozel:
Thanks for your contribution @lfz757077613 (both the PR and the discussion). We've discussed this matter and chose to go in a different direction. Because we'd like to ship this change with the next release tomorrow, we're closing this PR and addressing that in #27046.

A few things to consider:

1. In #27046, we'll align our configuration property with the default value in Netty and cover that with a unit test
2. We understand that the logging behavior can be confusing, but currently this is only logging changes done at the static initialization of the class at the Netty level, so really logging the system property set. If anyone is changing that by calling `ResourceLeakDetector.setLevel(level)` statically, nothing will be logged. If you think this is a problem, you can voice your opinion in [the Netty issue tracker](https://github.com/netty/netty/issues/) and discuss this matter. Personally, I think the behavior is consistent still, because it's only surfacing system properties.
3. We think that other options (null default value, additional logging) aren't helping here and would not align with the rest of Spring Boot's configuration properties behavior.

I'm closing this PR as a result. Thanks!

lfz757077613:
@bclozel Hi, I think additional-spring-configuration-metadata.json is forgot to change the defaultValue in 2.5.2

bclozel:
@lfz757077613 Thanks, I've opened #27104 to fix that.

PR #26961, opened 2021-06-18
Add resource labels to Stackdriver metrics configuration properties

mikouaj:
Using resource labels is mandatory for most Stackdriver resources other then `Global`, i.e. `k8s_pod`.
While setting resource type is supported in `StackdriverProperties`, it is not possible to set labels in the same, configurable way.
Configuring resource type only lead to issues as described in micrometer-metrics/micrometer#1907.

Using valid resource type along with required labels makes it possible to use given metric in a wider set of
GCP solutions, i.e. custom metric based GKE horizontal pod autoscaler.


wilkinsona:
@mikouaj Thank you very much for making your first contribution to Spring Boot.

PR #26975, opened 2021-06-18
Document uuid support in RandomValuePropertySource

hrybs:
Class `org.springframework.boot.env.RandomValuePropertySource` contains incorrect Java Doc. When the suffix is not `int` or `long` the value not be always `byte[]`. When the suffix is `uuid` then result will be UUID value.


snicoll:
@hrybs thank you for making your first contribution to Spring Boot.

PR #26982, opened 2021-06-20
Add Rabbit connection factory configurers

bono007:
These changes are then used by 
https://github.com/spring-cloud/spring-cloud-stream-binder-rabbit/pull/323

@wilkinsona

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractConnectionFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractConnectionFactoryConfigurer.java
@@ -1,0 +1,61 @@
+package org.springframework.boot.autoconfigure.amqp;
+
+import org.springframework.amqp.rabbit.connection.AbstractConnectionFactory;
+import org.springframework.amqp.rabbit.connection.ConnectionNameStrategy;
+import org.springframework.boot.context.properties.PropertyMapper;
+import org.springframework.util.Assert;
+
+/**
+ * Configures {@link AbstractConnectionFactory Rabbit ConnectionFactory} with sensible
+ * defaults.
+ *
+ * @param <T> the connection factory type.
+ * @author Chris Bono
+ * @since 2.6.0
+ */
+public abstract class AbstractConnectionFactoryConfigurer<T extends AbstractConnectionFactory> {
+
+	private RabbitProperties rabbitProperties;
+
+	private ConnectionNameStrategy connectionNameStrategy;
+
+	public RabbitProperties getRabbitProperties() {
+		return rabbitProperties;
+	}
+
+	public void setRabbitProperties(RabbitProperties rabbitProperties) {
+		this.rabbitProperties = rabbitProperties;
+	}
+
+	public ConnectionNameStrategy getConnectionNameStrategy() {
+		return connectionNameStrategy;
+	}
+
+	public void setConnectionNameStrategy(ConnectionNameStrategy connectionNameStrategy) {
+		this.connectionNameStrategy = connectionNameStrategy;
+	}
+
+	/**
+	 * Configure the specified Rabbit connection factory - delegating to
+	 * {@link #configureSpecific} for the connection factory implementation specific
+	 * settings. The factory can be further tuned and default settings can be overridden.
+	 * @param connectionFactory the connection factory instance to configure
+	 */
+	public void configure(T connectionFactory) {
+		Assert.notNull(connectionFactory, "ConnectionFactory must not be null");
+		PropertyMapper map = PropertyMapper.get();
+		map.from(this.rabbitProperties::determineAddresses).to(connectionFactory::setAddresses);
+		map.from(this.rabbitProperties::getAddressShuffleMode).whenNonNull()
+				.to(connectionFactory::setAddressShuffleMode);
+		map.from(connectionNameStrategy).whenNonNull().to(connectionFactory::setConnectionNameStrategy);
+		configureSpecific(connectionFactory);
+	}
+
+	/**
+	 * Configure the specified Rabbit connection factory with implementation specific
+	 * settings.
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
```diff
@@ -93,73 +92,48 @@
 public class RabbitAutoConfiguration {
 
 	@Configuration(proxyBeanMethods = false)
-	@ConditionalOnMissingBean(ConnectionFactory.class)
 	protected static class RabbitConnectionFactoryCreator {
 
 		@Bean
-		public CachingConnectionFactory rabbitConnectionFactory(RabbitProperties properties,
+		@ConditionalOnMissingBean
+		RabbitConnectionFactoryBeanConfigurer rabbitConnectionFactoryBeanConfigurer(RabbitProperties properties,
```
bono007, line 99: I know this is not `protected`, but rather `package` visibility. Is there a reason that in general SB will use public methods under protected classes which allow the subclass abuse listed in gh-187 to take place? I am guessing its to allow users to locally bug fix or workaround issues temporarily via the subclass as an immediate fix in odd cases?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
```diff
@@ -93,73 +92,48 @@
...
+				ObjectProvider<ConnectionNameStrategy> connectionNameStrategy) {
+			CachingConnectionFactoryConfigurer configurer = new CachingConnectionFactoryConfigurer();
+			configurer.setRabbitProperties(rabbitProperties);
+			configurer.setConnectionNameStrategy(connectionNameStrategy.getIfUnique());
+			return configurer;
+		}
+
+		@Bean
+		@ConditionalOnMissingBean(ConnectionFactory.class)
+		CachingConnectionFactory rabbitConnectionFactory(
```
bono007, line 122: This is a breaking change:
1. visibility public->package (talking about that in https://github.com/spring-projects/spring-boot/pull/26982/files#r655007725)
2. method signature change (params)

We could probably leave the old method and deprecate and then do this additively. I just wanted to put this out there before investing in it though.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractConnectionFactoryConfigurer.java
```diff
@@ -0,0 +1,61 @@
...
+
+/**
+ * Configures {@link AbstractConnectionFactory Rabbit ConnectionFactory} with sensible
+ * defaults.
+ *
+ * @param <T> the connection factory type.
+ * @author Chris Bono
+ * @since 2.6.0
+ */
+public abstract class AbstractConnectionFactoryConfigurer<T extends AbstractConnectionFactory> {
```
bono007, line 16: While SB only auto-configures AbstractConnectionFactory impl of CachingConnectionFactory currently, I went ahead and split it out into ABC + impl specific concrete class to set the specific properties. This would allow some other consumer to extend the ABC for say PooledConnectionFactory or ThreadConnectionFactory.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
wilkinsona: No, it should be package-private so your change here is a good one.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
wilkinsona: A breaking change here is fine as we don't consider the `@Bean` methods on an auto-configuration class to be public API.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractConnectionFactoryConfigurer.java
```diff
@@ -0,0 +1,61 @@
...
+
+	private RabbitProperties rabbitProperties;
+
+	private ConnectionNameStrategy connectionNameStrategy;
+
+	public RabbitProperties getRabbitProperties() {
+		return rabbitProperties;
+	}
```
wilkinsona, line 24: As the properties are required, I'd that they're a constructor argument. I'll make that change as part of merging.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/AbstractConnectionFactoryConfigurer.java
```diff
@@ -0,0 +1,61 @@
...
+		map.from(connectionNameStrategy).whenNonNull().to(connectionFactory::setConnectionNameStrategy);
+		configureSpecific(connectionFactory);
+	}
+
+	/**
+	 * Configure the specified Rabbit connection factory with implementation specific
+	 * settings.
+	 * @param connectionFactory the connection factory instance to configure
+	 */
+	protected abstract void configureSpecific(T connectionFactory);
```
wilkinsona, line 59: I think this could be an overload of `configure` that takes the `RabbitProperties` as well. That would remove the need for the getter. I'll make this change as part of merging.

wilkinsona:
Thanks very much, @bono007.

PR #27017, opened 2021-06-22
ConfigDataPropertiesTests contains a property name typo

limo520:
"spring.config.activate.on-profiles" should be "spring.config.activate.on-profile".


wilkinsona:
Thanks for the proposal. Given that the tests passes at the moment with the wrong property, I'm not sure that the test is exercising the intended code path. Flagging for team attention.

philwebb:
@mbhave did some analysis on this and discovered the test passed because `on-profiles` is simply ignored. I've added some additional tests in 93d316ddd17976ec95d721c79597ab2182ea5ed3 to test the negative case. If the typo is in those, they fail.

philwebb:
Thanks for spotting this @limo520

PR #27085, opened 2021-06-24
Avoid reflection in tests

dreis2211:
Hi,

this PR removes some reflection calls from the test code.

Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/websocket/servlet/WebSocketMessagingAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/websocket/servlet/WebSocketMessagingAutoConfigurationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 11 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/websocket/servlet/WebSocketMessagingAutoConfigurationTests.java
```diff
@@ -138,13 +137,10 @@ void customizedConverterTypesMatchDefaultConverterTypes() {
 		return customizedConverters;
 	}
 
-	@SuppressWarnings("unchecked")
 	private List<MessageConverter> getDefaultConverters() {
 		DelegatingWebSocketMessageBrokerConfiguration configuration = new DelegatingWebSocketMessageBrokerConfiguration();
-		// We shouldn't usually call this method directly since it's on a non-proxy config
```
dreis2211, line 144: I was unsure about this one. I didn't get why we shouldn't call the method directly inside tests.

ADarko22: looks safe to me

ADarko22 (approved):

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/websocket/servlet/WebSocketMessagingAutoConfigurationTests.java
wilkinsona: My first guess was that the method hasn’t always been public, but that’s not the case.

The change to using reflection was made in [this polishing commit](https://github.com/spring-projects/spring-boot/commit/68aec8b6ee5e3c51c12f7a17fe4b0cef1709f908) by @philwebb. My second guess is that the change was made to avoid [our tooling](https://github.com/wilkinsona/spring-boot-development-eclipse-plugin) reporting an error about a direct call to an unproxied `@Bean` method.

dreis2211: @wilkinsona Should I keep the proposed change or do you want me to revert this particular one?

PR #27100, opened 2021-06-25
Avoid illegal reflective access on ZipEntry.xdostime

dreis2211:
Hi,

this PR provides an alternative solution for the illegal reflective access on `ZipEntry.xdostime` inside `JarFileTests.jarFileEntryWithEpochTimeOfZeroShouldNotFail` and is part of the overall JDK 17 story at #26767.

As per [comment](https://github.com/spring-projects/spring-boot/pull/27089#issuecomment-868212766) from @AlanBateman I took another look and found that the JDK is testing things a bit differently for similar scenarios that doesn't require reflection or opening up modules. See [ZeroDate.java](https://github.com/openjdk/jdk/blob/master/test/jdk/java/util/zip/ZipFile/ZeroDate.java)

The idea is basically to change the necessary bytes in the actual file. Also not too pretty, I must say. But it's reflection free and doesn't access any internals anymore. 

Let me know what you think.
Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #27102, opened 2021-06-25
Fix typo in section title

cordin:
This small typo was causing that the section title wasn't rendered as
expected. It happens since boot 2.5.0, as can be checked in 
https://docs.spring.io/spring-boot/docs/2.5.0/reference/htmlsingle/#howto.batch.running-jobs-on-startup
where instead of the next section title, the following text is shown:

[[howto.batch.?running-from-the-command-line]] === Running from the Command Line Spring Boot converts any command line


ADarko22 (approved):

snicoll:
@ADarko22 thank you but we don't need a review for trivial changes like this.

PR #27103, opened 2021-06-25
Remove unnecessary unit for spring.datasource.tomcat.max-active

temnur:
There is mistakenly added measure millisecond(ms) for config definition `spring.datasource.tomcat.max-active=50`. Here 50 should be pure count


ADarko22 (approved):

snicoll:
@temnur thank you for making your first contribution to Spring Boot.

Issue #26220, opened 2021-04-23
Remove spring.mongodb.embedded.version default value


rajadilipkolli:
Currently, the default MongoDB version that comes with embedded MongoDB dependency in spring boot is 3.5.5,  the latest available version is 4.4.5. Request you to either upgrade it to a 4+ version or provide the documentation on how to change the default version.

philwebb:
You can change the version by setting [`spring.mongodb.embedded.version`](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#spring.mongodb.embedded.version). I think we should align with the managed dependency going forward.

snicoll:
See also #14476 for a similar request. @philwebb I wonder what you had in mind to align the two going forward considering it is now hardcoded in code.

wilkinsona:
When I updated the issue's title, I had naively assumed there would be some sort of mapping between the version of the Java Driver and the version of MongoDB that we should use by default. We could then have used `com.mongodb.internal.build.MongoDriverVersion.VERSION` in a test to check that the two were aligned. Unfortunately, there's no such mapping. The latest Java driver version is 4.2.3 while the latest MongoDB version is 4.4.5. The 4.2 driver is [compatible with](https://docs.mongodb.com/drivers/java/#mongodb-compatibility) MongoDB 4.4, 4.2, 4.0, 3.6, 3.4, 3.2, 3.0, and 2.6.

wilkinsona:
I wonder if we should require users to set `spring.mongodb.embedded.version` in order to use embedded MongoDB. The version really needs to align with whatever version of MongoDB is being used in production and I can't see how we can provide a sensible default value for that.

Testcontainers made some changes a while ago to deprecate all of the default constructors that cause a default tag to be used. This feels quite similar to me.

rajadilipkolli:
HI @philwebb, When I am playing with setting `spring.mongodb.embedded.version` I found that we can only set the versions that are present [here](https://dl.mongodb.org/dl/win32/x86_64) 

For example, if I want to use 4.4.1 it will give 403. So letting users choose the version number by default will break or we need to document it carefully.

snicoll:
Closing in favor or PR #27108

PR #27108, opened 2021-06-26
Remove default spring.mongodb.embedded.version

bono007:
Removes the previous default setting for `spring.mongodb.embedded.version` of `3.5.5`. 

Fixes gh-26220

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/nosql.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/nosql.adoc
@@ -151,6 +151,13 @@ TIP: For complete details of Spring Data MongoDB, including its rich object mapp
 Spring Boot offers auto-configuration for https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo[Embedded Mongo].
 To use it in your Spring Boot application, add a dependency on `de.flapdoodle.embed:de.flapdoodle.embed.mongo`.
 
+The version of the MongoDB server **must** be configured by setting the configprop:spring.data.mongodb.embedded.version property.
+
+NOTE: The default download configuration allows access to most of the versions listed in https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo/blob/de.flapdoodle.embed.mongo-3.0.0/src/main/java/de/flapdoodle/embed/mongo/distribution/Version.java[Version.java]
+as well as some others. Inaccessible versions result in error when attempting to download the server. The download configuration may be adjusted to handle the desired version.
+
+The download configuration can be customized by declaring a `DownloadConfigBuilderCustomizer` bean.
+
 The port that Mongo listens on can be configured by setting the configprop:spring.data.mongodb.port[] property.
 To use a randomly allocated free port, use a value of 0.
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/nosql.adoc
```diff
@@ -151,6 +151,13 @@ TIP: For complete details of Spring Data MongoDB, including its rich object mapp
 Spring Boot offers auto-configuration for https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo[Embedded Mongo].
 To use it in your Spring Boot application, add a dependency on `de.flapdoodle.embed:de.flapdoodle.embed.mongo`.
 
+The version of the MongoDB server **must** be configured by setting the configprop:spring.data.mongodb.embedded.version property.
+
+NOTE: The default download configuration allows access to most of the versions listed in https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo/blob/de.flapdoodle.embed.mongo-3.0.0/src/main/java/de/flapdoodle/embed/mongo/distribution/Version.java[Version.java]
+as well as some others. Inaccessible versions result in error when attempting to download the server. The download configuration may be adjusted to handle the desired version.
```
bono007, line 157: I am not sure what the recommended way is to point to the `Version` source file from the docs. 

I pointed to the 3.0.0 tag since that is what the main Spring Boot branch currently points to for flapdoodle. This makes the info in that particular file accurate but also makes a location that needs to be updated when the flapdoodle version changes. However, If we use link to the flapdoodle main branch then the info may not be accurate as the Version file may change over time.

wilkinsona:
The upgrade instructions in the releases notes should mention that `spring.mongodb.embedded.version` now needs to be set and explain why that's the case.

wilkinsona:
Thanks very much, @bono007.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/nosql.adoc
wilkinsona: Just in case you're interested, @bono007, I used an attribute for this with its value set to the version of the embedded Mongo dependency: https://github.com/spring-projects/spring-boot/commit/4d76660f1735f34cf095b44c983ce99fae269dd5#diff-ad4cb147eace80b1ffbafcd284f6c2c8479a07ecc150d71f49eb34ce6d37293dR223.

bono007: > Just in case you're interested, @bono007, I used an attribute for this with its value set to the version of the embedded Mongo dependency: [4d76660#diff-ad4cb147eace80b1ffbafcd284f6c2c8479a07ecc150d71f49eb34ce6d37293dR223](https://github.com/spring-projects/spring-boot/commit/4d76660f1735f34cf095b44c983ce99fae269dd5#diff-ad4cb147eace80b1ffbafcd284f6c2c8479a07ecc150d71f49eb34ce6d37293dR223).

TIL! Thx for pointing that out as I will use this in the future.

PR #27144, opened 2021-07-03
Initialize BatchDataSourceInitializer eagerly

hpoettker:
Spring Boot apps with autoconfiguration for Spring Batch currently do not initialize the data source with `BatchDataSourceInitializer` if `spring.main.lazy-initialization` is set to `true` independent of what the property `spring.batch.jdbc.initialize-schema` is set to.

The issue can be worked around by users with a `LazyInitializationExcludeFilter` but a consistent behavior for `spring.batch.jdbc.initialize-schema` independent of lazy or eager initialization would be nice.

PR #27145, opened 2021-07-03
Initialize FlywayMigrationInitializer eagerly

hpoettker:
Spring Boot apps with autoconfiguration for Flyway currently do not trigger the migration if `spring.main.lazy-initialization` is set to true independent of what the property `spring.flyway.enabled` is set to.

The issue can be worked around by users with a `LazyInitializationExcludeFilter` for `FlywayMigrationInitializer` but a consistent behavior for `spring.flyway.enabled` independent of lazy or eager initialization would be nice.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
@@ -55,6 +55,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
+import org.springframework.context.annotation.Lazy;
 import org.springframework.core.convert.TypeDescriptor;
 import org.springframework.core.convert.converter.GenericConverter;
 import org.springframework.core.io.ResourceLoader;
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the proposal here as well, but this change also doesn't feel quite right to me.

As with #27144, I think it should be possible for the initialiser to be lazy and for Flyway to then run prior to first use of the DataSource. If Flyway needs to be eagerly initialized for a reason that I have overlooked then I think it should be done with a LazyInitializationExcludeFilter that Spring Boot auto-configures. We should also add a test that replicates the scenario that requires eager initialization and verify that it now works.

wilkinsona:
I've tested this and the dependencies between other beans, for example JdbcTemplate, and Flyway seem to be sufficient to ensure that the migration initializer is called before the database is accessed.

@hpoettker I assume that you've seen a failure when using Flyway and lazy init is enabled. Can you please provide some details of that failure? I'd like to understand which bean dependencies are missing.

hpoettker:
@wilkinsona Thank you so much for your interest in getting to the bottom of why this PR was raised. This and all your other efforts to get lazy initialization working seamlessly are really appreciated.

I've created a minimal example with a failing test here: https://github.com/hpoettker/lazyflyway

Thanks to your feedback, I now see that if neither Spring JDBC, nor JPA, nor jOOQ is used, then there is no `DependsOnDatabaseInitializationDetector` to create a dependency on `FlywayMigrationInitializer`.

I guess there are two opposed ways to look at the minimal example:
- Spring Boot cannot know that I use the `DataSource` with a third-party library (i.e. jdbi). Thus, it is my responsibility to (for example) annotate `LazyflywayApplication::jdbi` with `@DependsOnDatabaseInitialization`.
- There should be a `DependsOnDatabaseInitializationDetector` that detects any `DataSource`. If Flyway is on the classpath and a `DataSource` is (auto-)configured, is there any doubt that I want the `DataSource` to be initialized with flyway?

In any case, this PR is not the solution. Thus, I'll close it.

Issue #27193, opened 2021-07-08
App fails to start when using Spring Batch with JDBC and lazy initialization is enabled


wilkinsona:
```
  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::                        

2021-07-08 18:27:56.391  INFO 73510 --- [           main] smoketest.batch.SampleBatchApplication   : Starting SampleBatchApplication using Java 1.8.0_252 on wilkinsona-a01.vmware.com with PID 73510 (/Users/awilkinson/dev/spring-projects/spring-boot/2.4.x/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-batch/bin/main started by awilkinson in /Users/awilkinson/dev/spring-projects/spring-boot/2.4.x/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-batch)
2021-07-08 18:27:56.393  INFO 73510 --- [           main] smoketest.batch.SampleBatchApplication   : No active profile set, falling back to default profiles: default
2021-07-08 18:27:57.124  INFO 73510 --- [           main] smoketest.batch.SampleBatchApplication   : Started SampleBatchApplication in 1.145 seconds (JVM running for 1.521)
2021-07-08 18:27:57.247  INFO 73510 --- [           main] o.s.b.a.b.JobLauncherApplicationRunner   : Running default command line with: []
2021-07-08 18:27:57.302  INFO 73510 --- [           main] com.zaxxer.hikari.HikariDataSource       : HikariPool-1 - Starting...
2021-07-08 18:27:57.560  INFO 73510 --- [           main] com.zaxxer.hikari.pool.PoolBase          : HikariPool-1 - Driver does not support get/set network timeout for connections. (feature not supported)
2021-07-08 18:27:57.563  INFO 73510 --- [           main] com.zaxxer.hikari.HikariDataSource       : HikariPool-1 - Start completed.
2021-07-08 18:27:57.571  INFO 73510 --- [           main] o.s.b.c.r.s.JobRepositoryFactoryBean     : No database type set, using meta data indicating: HSQL
[...]

Issue #27207, opened 2021-07-08
AbstractDataSourceInitializers are not detected as database initializers


wilkinsona:
When we fix this, we should be able to replace `QuartzSchedulerDependencyConfiguration` with a `DependsOnDatabaseInitializationDetector` for `Scheduler` and `SchedulerFactoryBean`. We may want to see if we can somehow make it specific to the JDBC store type.

wilkinsona:
Slight change of plan. Let's close this in favour of #27215 and then use some separate issues to make the relevant Batch, Session JDBC, and Quartz beans depend on data source initialization.

PR #27215, opened 2021-07-09
AbstractDataSourceInitializers are not detected as database initializers

hpoettker:
@wilkinsona Thanks for the feedback on #27144. I think this is the approach you suggested.

Issue #27193


wilkinsona:
Thank you, @hpoettker. This is what I had in mind for the Batch side of things in 2.5. For 2.4 things are quite different and I've fixed that in #27193 .

I've realised that the problem's a bit broader as Spring Session JDBC has a similar problem and we can refine so Quartz configuration as well once we've got the necessary detection in place. I'll merge (some of) the changes proposed here – which will take care of #27207 – and then make some changes that build on top of them in separate issues.

wilkinsona:
@hpoettker Thanks once again for making your first contribution to Spring Boot.

wilkinsona:
I've merged the Batch-specific parts of the originally-proposed changes in [this commit](https://github.com/spring-projects/spring-boot/commit/cf69785393137bf512eb6c9521f923cf4f7c0cd8).

PR #27229, opened 2021-07-09
Fix tests for multi-release JARs on JDK 17

dreis2211:
Hi,

this is part of the overall JDK story #26767 and fixes the multi-release JAR tests. Instead of just adding another entry for JDK 17 this PR provides an alternative approach that is a little more dynamic and forgiving when a new JDK arrives.

Let me know what you think.
Cheers,
Christoph


snicoll:
Very nice, thanks again Christoph.

PR #27294, opened 2021-07-12
Use server version from database in Neo4j health details

meistermeier:
The version are now fetched via the build in meta-information from the database itself. They were previously retrieved via the Neo4j driver's `ServerInfo#version` method.
Is the reactive return type nice? No, but pragmatic.


meistermeier:
Still see one issue with this. Will ping you again if it is solved.

meistermeier:
Ok, good to go..

wilkinsona:
Thanks very much, @meistermeier. I couldn't quite bring myself to live with the `Tuple2<Tuple2<String, String>, ResultSummary>` so I introduced a new type in https://github.com/spring-projects/spring-boot/commit/9f9c89a35777fc63bf672ff34c8f7897b79e2130 that encapsulates all of the details that's now used by both indicators.

meistermeier:
Hehe. I agree 100% but did not know if creating a class for a short living object would be too much overhead in Spring Boot terms.
But there is one thing: the result of the driver call should not get carried outside the session boundaries. Would be better to extract edition and version in place and pass both to the wrapper class.

wilkinsona:
Thanks, Gerrit. I've fixed the extraction of the edition and version in e1049a55ffda51cde86c7df9fdacd496d957086e.

PR #27300, opened 2021-07-13
Redis health indicators report that Redis is up when the cluster's state is fail

hoanvh:
## Given
a Redis cluster with 3 shards + 3 replicas as:
 - :7001 (master) ~> :7004 (slave)
 - :7002 (master) ~> :7005 (slave)
 - :7003 (master) ~> :7006 (slave)

Kill processes with ports 7003, 7004, 7005, 7006.

check `CLUSTER INFO` command:
```
cluster_state:fail
cluster_slots_assigned:16384
cluster_slots_ok:5462
cluster_slots_pfail:10922
cluster_slots_fail:0
cluster_known_nodes:6
cluster_size:3
...
```

## As is:
Although cluster's state is **fail** but RedisHealthIndicator still mark Health as **UP**

## To be:
If cluster's state is **fail**, then we will mark Redis Health as **DOWN**

## Solution:
Check `cluster_state` value, if it equals (ignore case) with `fail`, then Redis Health is DOWN.

## Reference
https://redis.io/commands/cluster-info

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/redis/RedisHealth.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/redis/RedisHealth.java
@@ -38,11 +38,16 @@ final class RedisHealth {
 		return builder.up();
 	}
 
-	static Builder up(Health.Builder builder, ClusterInfo clusterInfo) {
+	static Builder info(Health.Builder builder, ClusterInfo clusterInfo) {
 		builder.withDetail("cluster_size", clusterInfo.getClusterSize());
 		builder.withDetail("slots_up", clusterInfo.getSlotsOk());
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @hoanvh. If you're interested, I polished your changes to bring them into line with the project's conventions and code style in [this commit](https://github.com/spring-projects/spring-boot/commit/49baacbc1cb02a59efce1fe1698166de92d41d67).

hoanvh:
@wilkinsona Regarding to this patch, it seems to be available on SB version `2.4.9`? We're using SB `2.5.2`, so I wonder which version `2.5.x` includes it?

wilkinsona:
It will be in 2.5.3 as per the [forward port issue](https://github.com/spring-projects/spring-boot/issues/27304) that’s [shown above](https://github.com/spring-projects/spring-boot/pull/27300#ref-issue-943422373).

PR #27310, opened 2021-07-13
Fix JDK 16 pipeline

dreis2211:
Hi,

after #27089 was merged I noticed that we can't get rid of `--illegal-access=warn` until a JDK 16 pipeline exists, so this PR brings that back (for pragmatic reasons without any special check).

Sorry for not noticing earlier.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #27328, opened 2021-07-14
Disable failing Gradle plugin tests on JDK 17

dreis2211:
Hi,

this is the last open item before we can create a JDK 17 pipeline - see #26767.
I wanted to make an effort to not just exclude the whole module (or all of its tests) via CI, but test as much as we can. The added helper classes that we need to workaround the shortcomings of Gradle, will be likely removed again as soon as Gradle is ready for JDK 17. Since they're not published and only test relevant, I figured this to be acceptable. The fact that it's more fine-grained and explicit was appealing to me, but I could also exclude all tests with some hackery in CI (we did this in the past). Neither solution is beautiful if you ask me, but I went for the one that keeps more test coverage.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/test/java/org/springframework/boot/gradle/junit/GradleCompatibilityExtension.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/test/java/org/springframework/boot/gradle/junit/GradleCompatibilityExtension.java
@@ -57,6 +57,9 @@ final class GradleCompatibilityExtension implements TestTemplateInvocationContex
 
 	@Override
 	public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(ExtensionContext context) {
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/test/java/org/springframework/boot/gradle/junit/GradleCompatibilityExtension.java
```diff
@@ -57,6 +57,9 @@
 
 	@Override
 	public Stream<TestTemplateInvocationContext> provideTestTemplateInvocationContexts(ExtensionContext context) {
+		if (JavaVersion.current() == JavaVersion.VERSION_17) {
+			return Stream.of(new DisabledTemplateInvocationContext());
```
dreis2211, line 61: Unfortunately, we can't just return an empty stream here. JUnit is expecting at least one invocation context

wilkinsona:
LGTM! Thanks very much once again, @dreis2211.

PR #27371, opened 2021-07-16
 Add configuration property for Reactor Netty's idle timeout

### LINKED ISSUE #27312: Add property for reactor-netty server idleTimeout
Add a configuration property for reactor-netty server's idleTimeout.

Property name: `server.netty.idle-timeout`
Property type: Duration
Default value: not configured (infinite)

Reactor-netty added an idleTimeout configuration property in https://github.com/reactor/reactor-netty/pull/1377.

Currently, this can be set programmatically like this:

```java
@Bean
WebServerFactoryCustomizer<NettyReactiveWebServerFactory> nettyServerCustomizer() {
   return factory -> factory.addServerCustomizers(httpServer -> httpServer.idleTimeout(Duration.ofSeconds(20)));
}
```

Netty server's default idle timeout is infinite.   Spring-boot can keep the same default.

We have to set this value in all of our apps due to network "stuff" killing idle connections non-gracefully (sigh).  Therefore occasionally clients don't notice terminated connections in their connection pool, and attempt to reuse it, and receive response timeouts.  We've noticed that when we add the server-side idleTimeout, the clients can detect the closed connections, and don't try to reuse them.  Clients can also reduce their idleTimeout to fix this problem, but the server-side fix helps in general.

Having a configuration property would be really nice so we can avoid the programmatic config!

saraswathy-krish:
Through this change,Spring boot is setting its idleTimeout property for netty server to infinite.This is done by introducing a configuration property "server.netty.idle-timeout".

Fixes #27312

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -1300,6 +1300,11 @@ public class ServerProperties {
 		 */
 		private boolean validateHeaders = true;
 
+		/**
+		 * IdleTimeout for netty server .If an idletimeout is not specified, this indicates no timeout(infinite).
+		 */
+		private Duration idleTimeout;
+
 		public Duration getConnectionTimeout() {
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks very much, @saraswathy-krish.

Issue #22273, opened 2020-07-08
Support redis-socket:// URLs when configuring a Redis client


scottfrederick:
This is related to #21920. 

Spring Data Redis has a `RedisSocketConfiguration` that can be used to configure a Unix domain socket connection to Redis. The Lettuce client [supports domain socket configuration](https://lettuce.io/core/release/reference/#unix-domain-sockets) with URLs like `redis-socket://redis/redis.sock`. 

If we decide to support `redis-sentinel://` URLs in `spring.redis.url`, we might also want to support `redis-socket://` URLs.

snicoll:
Closing in favor of PR #27373

Issue #21920, opened 2020-06-15
Support redis-sentinel:// URLs when configuring a Redis client


conanca:
When "spring.redis.url" set with

redis-sentinel :// [[_username_ *:*] password@] host1[*:* _port1_] [, host2[*:* _port2_]] [, hostN[*:* _portN_]] [*/* _database_][?[timeout=timeout[_d|h|m|s|ms|us|ns_]] [&_sentinelMasterId=sentinelMasterId_] [&_database=database_]]

(refer to [redisuri.uri-syntax](https://lettuce.io/core/release/reference/#redisuri.uri-syntax) )

got an error with:

[...]

PR #27373, opened 2021-07-17
Add support for redis-sentinel in spring.redis.url (Lettuce only)

onobc:
Adds support for setting the following `spring.redis.uri` when Lettuce is used:

- `redis-sentnel`
- `rediss-sentinel`
- `redis-socket`

⚠️  This also implicitly adds support for `redis-socket://` (#22273) 

### Alt Schemes
ℹ️ : Because the RedisURI is being used to parse the url string, other url schemes are [actually supported](https://github.com/lettuce-io/lettuce-core/blob/b161915eefbc1ea92131c65086787230ea227253/src/main/java/io/lettuce/core/RedisURI.java#L143) - namely:
- `redis+ssl`
- `redis+tls`
- `redis+socket`

Should we block user's from using these alt schemes? It seems that if we don't block, we at least don't advertise the ability to use these alt schemes. @wilkinsona  @snicoll  wdyt?

### Config Validation 
I was tempted to add a validateConfiguration` that asserted that if the url was used, only the relevant other props could be used. This would of course, potentially break existing users (both Lettuce and Jedis). Just wanted to get your opinions here as well. 

```java
private void validateConfiguration() {
    if (!StringUtils.hasText(getProperties().getUrl())) {
        return;
    }

    Assert.state(getProperties().getSentinel() == null || !StringUtils.hasText(getProperties().getSentinel().getMaster()),
            "Invalid redis configuration, 'sentinel.master' not supported when 'url' is specified");

    Assert.state(getProperties().getSentinel() == null || ObjectUtils.isEmpty(getProperties().getSentinel().getNodes()),
            "Invalid redis configuration, 'sentinel.nodes' not supported when 'url' is specified");

    Assert.state(this.getProperties().getCluster() == null,
            "Invalid redis configuration, 'cluster' properties not supported when 'url' is specified");
}
```

### TODO 

- [x] Get resolution on "Alt Schemes" (above)
- [x] Get resolution on "Config Validation" and adjust (or not) accordingly
- [x] Update docs

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
@@ -136,18 +136,20 @@ class LettuceConnectionConfiguration extends RedisConnectionConfiguration {
 
 	private boolean urlConfiguredForSentinelScheme(String url) {
 		String scheme = getUrlScheme(url);
-		return scheme != null
-				&& RedisURI.URI_SCHEME_REDIS_SENTINEL.equals(scheme) || RedisURI.URI_SCHEME_REDIS_SENTINEL_SECURE.equals(scheme);
+		return scheme != null && RedisURI.URI_SCHEME_REDIS_SENTINEL.equals(scheme)
+				|| RedisURI.URI_SCHEME_REDIS_SENTINEL_SECURE.equals(scheme);
 	}
 
 	private boolean urlConfiguredForSocketScheme(String url) {
 		String scheme = getUrlScheme(url);
-		return scheme != null && (RedisURI.URI_SCHEME_REDIS_SOCKET.equals(scheme) || RedisURI.URI_SCHEME_REDIS_SOCKET_ALT.equals(scheme));
+		return scheme != null && (RedisURI.URI_SCHEME_REDIS_SOCKET.equals(scheme)
+				|| RedisURI.URI_SCHEME_REDIS_SOCKET_ALT.equals(scheme));
 	}
 
 	private boolean urlConfiguredForNonStandardStandaloneScheme(String url) {
 		String scheme = getUrlScheme(url);
-		return scheme != null && !RedisURI.URI_SCHEME_REDIS.equals(scheme) && !RedisURI.URI_SCHEME_REDIS_SECURE.equals(scheme);
+		return scheme != null && !RedisURI.URI_SCHEME_REDIS.equals(scheme)
+				&& !RedisURI.URI_SCHEME_REDIS_SECURE.equals(scheme);
 	}
 
 	private String getUrlScheme(String url) {
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
@@ -20,6 +20,7 @@ import java.time.Duration;
 
 import io.lettuce.core.ClientOptions;
 import io.lettuce.core.RedisClient;
+import io.lettuce.core.RedisURI;
 import io.lettuce.core.SocketOptions;
 import io.lettuce.core.TimeoutOptions;
 import io.lettuce.core.cluster.ClusterClientOptions;
@@ -38,6 +39,7 @@ import org.springframework.boot.autoconfigure.data.redis.RedisProperties.Pool;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.data.redis.connection.RedisClusterConfiguration;
+import org.springframework.data.redis.connection.RedisConfiguration;
 import org.springframework.data.redis.connection.RedisConnectionFactory;
 import org.springframework.data.redis.connection.RedisSentinelConfiguration;
 import org.springframework.data.redis.connection.RedisStandaloneConfiguration;
@@ -45,6 +47,7 @@ import org.springframework.data.redis.connection.lettuce.LettuceClientConfigurat
 import org.springframework.data.redis.connection.lettuce.LettuceClientConfiguration.LettuceClientConfigurationBuilder;
 import org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory;
 import org.springframework.data.redis.connection.lettuce.LettucePoolingClientConfiguration;
+import org.springframework.util.ObjectUtils;
 import org.springframework.util.StringUtils;
 
 /**
@@ -84,6 +87,11 @@ class LettuceConnectionConfiguration extends RedisConnectionConfiguration {
 	}
 
 	private LettuceConnectionFactory createLettuceConnectionFactory(LettuceClientConfiguration clientConfiguration) {
+		if (StringUtils.hasText(this.getProperties().getUrl())) {
+			RedisURI redisURI = createRedisUri(getProperties().getUrl());
+			RedisConfiguration redisConfiguration = LettuceConnectionFactory.createRedisConfiguration(redisURI);
+			return new LettuceConnectionFactory(redisConfiguration, clientConfiguration);
+		}
 		if (getSentinelConfig() != null) {
 			return new LettuceConnectionFactory(getSentinelConfig(), clientConfiguration);
 		}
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfigurationLettuceUrlTests.java
```diff
@@ -0,0 +1,238 @@
...
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests {@link RedisAutoConfiguration} when using {@code Lettuce} and configuring via the
+ * {@code spring.redis.url} property.
+ *
+ * @author Chris Bono
+ */
+class RedisAutoConfigurationLettuceUrlTests {
```
bono007, line 36: I chose to break these out into their own test as the existing test was already pretty unwieldy and adding this in would not help. If this is an issue we can of course consolidate back in.

snicoll: There isn't a relationship between the main test and the jedis related one. If you want to break things down, why not `LettuceConnectionConfigurationTests` ?

snicoll (commented):
Thanks for the PR. I don't think the "Alt Schemes" are really a problem if we expand the scope of the PR a bit. Support for  `RedisURI` should cover our bases pretty well.

As for the validation, I don't know if failing is really an option. We do something like this in Mongo (`MongoPropertiesClientSettingsBuilderCustomizer`) though. I've flagged for team attention to get more feedback.

bono007:
> Thanks for the PR. I don't think the "Alt Schemes" are really a problem if we expand the scope of the PR a bit. Support for `RedisURI` should cover our bases pretty well.
> 

Sounds good. I will probably add some more tests to cover the alt-schemes a bit. 

> As for the validation, I don't know if failing is really an option. We do something like this in Mongo (`MongoPropertiesClientSettingsBuilderCustomizer`) though. I've flagged for team attention to get more feedback.

Sounds good. I will see what you all come back w/ and adjust (or not) accordingly.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfigurationLettuceUrlTests.java
bono007: I guess naming this new one containing the words "AutoConfiguration" and "Lettuce" may make it seem like this new test is the one that tests the Lettuce variant of auto-config when really the main test `RedisAutoConfigurationTests` does that. 

Naming this new test as you recommend is a bit clearer and may produce less confusion. I will adjust.

wilkinsona:
> As for the validation, I don't know if failing is really an option.

I’m in two minds. On the one hand, given that we don’t have good support for clearing a property’s value, I think it’s safer not to perform the validation. You may have a profile where you’re setting the URL and want the sentinel nodes that are configured elsewhere to be ignored. On the other hand, silently ignoring some configuration may be confusing.

For consistently with (most of) the rest of Boot, I think I’m leaning towards not performing any validation.

bono007:
> > As for the validation, I don't know if failing is really an option.
> 
> I’m in two minds. On the one hand, given that we don’t have good support for clearing a property’s value, I think it’s safer not to perform the validation. You may have a profile where you’re setting the URL and want the sentinel nodes that are configured elsewhere to be ignored. On the other hand, silently ignoring some configuration may be confusing.
> 
> For consistently with (most of) the rest of Boot, I think I’m leaning towards not performing any validation.

Another option would be to log a warning rather than fail. One one hand, this could be helpful to a user when debugging a connection issue. On the other hand, a warn log may not be noticed and that is code we have to add/maintain. I will proceed as-is w/o validation. It can always be added later.

bono007:
@wilkinsona @snicoll 
I added the test coverage for the other alt schemes.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/redis/RedisAutoConfigurationLettuceUrlTests.java
bono007: Ok after going back into the code I think the current name makes sense. Let me summarize my thinking and you can tell me if I am off of what you are thinking about this @snicoll.


#### Before my changes: 

`RedisAutoConfiguration`: 
- tests focused on provider agnostic auto-config features (albeit w/ Lettuce being the default provider). 
- handles the URL configured cases (with default provider)

`RedisAutoConfigurationJedisTests`: 
- tests when Jedis is used as the provider
- not as exhaustive as the main `RedisAutoConfiguration` test and only focuses on URL configured cases and other Jedis specifics.

At this point the URL configured cases are handled for both providers from the 2 above tests.

#### My changes:

Rather than put the new Lettuce specific URL configured cases in `RedisAutoConfiguration` I split it out into its own Lettuce URL auto-config test and removed any URL configured cases from the main test.

Up to this point the `LettuceConnectionConfiguration` and `JedisConnectionConfiguration` are tested implicitly via the _RedisAutoConfiguration*Tests_. I think naming this Lettuce URL test `LettuceConnectionConfigurationTests` will leave contributors wondering where the Jedis version is and also it would not be testing all of the LettuceConnectionConfiguration. Therefore I think leaving it named as `RedisAutoConfigurationLettuceUrlTests` makes sense.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/data/nosql.adoc
```diff
@@ -41,6 +41,41 @@ The following listing shows an example of such a bean:
 include::{docs-java}/features/nosql/redis/connecting/MyBean.java[]
 ----
 
+You can set the configprop:spring.data.redis.url[] property to change the URL and configure
```
bono007, line 44: @Buzzardo would you happen to be available for some of your famously fussy feedback 😺 ?

ℹ️ I based this on the MongoDB URL in the section directly below it.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/data/nosql.adoc
```diff
@@ -41,6 +41,41 @@ The following listing shows an example of such a bean:
...
+	      host: "redis.example.com"
+	      port: 6399
+	      username: "user"
+	      password: "secret"
+----
+When using both the URL and the discrete properties, any values that are set in the URL will take precedence over the values set in the properties.
+
+The url property supports the `redis` and `rediss` schemes as well as the following additional schemes when using Lettuce:
+
+* `redis+ssl`
```
bono007, line 72: The list of schemes we support is only limited by what the Lettuce [RedisURI](https://github.com/lettuce-io/lettuce-core/blob/b161915eefbc1ea92131c65086787230ea227253/src/main/java/io/lettuce/core/RedisURI.java#L143). I was hesitant to link to that detail here though and instead list out the schemes directly.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
```diff
@@ -82,6 +85,11 @@ LettuceConnectionFactory redisConnectionFactory(
 	}
 
 	private LettuceConnectionFactory createLettuceConnectionFactory(LettuceClientConfiguration clientConfiguration) {
+		if (StringUtils.hasText(this.getProperties().getUrl())) {
```
pirateskipper, line 88: if (StringUtils.hasText(getProperties().getUrl())) {

bono007: `Checkstyle` nor `format` seem to mind but it is not needed and inconsistent w/ the rest of the `getProperties` invocations in here.

snicoll: I am not sure about this one. The behavior is that we're checking for specific properties for `.cluster` or `.sentinel` and configure that if present. This reverse the trend if an URI is set, only for Lettuce.

This makes it quite hard to resonate about. Also, any of those properties are ignored and fully replaced by the URI, which sounds like quite a big change.

snicoll (commented):
I am not so sure we should be doing this. It's quite impactful and I don't know how we could conciliate the current arrangement while supporting this as an addition. Flagging for team attention again.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
bono007: > This reverse the trend if an URI is set, only for Lettuce.

True. Now the order they get connection config is:
- Jedis: `sentinel` -> `cluster` -> `standalone (which then uses uri)`
- Lettuce: `url` -> `sentinel` -> `cluster` -> `standalone (which then uses uri)`

This does create divergence between the 2 providers. Is the difference between the 2 providers the main concern, or the change in Lettuce connection config behavior? 

> Also, any of those properties are ignored and fully replaced by the URI, which sounds like quite a big change

Yeh this is why I was initially thinking about adding similar validation to [MongoPropertiesClientSettingsBuilderCustomizer](https://github.com/spring-projects/spring-boot/blob/3e21ae12dc2653ab7ce7d561c29bc018a57a4103/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoPropertiesClientSettingsBuilderCustomizer.java#L59) - which would force the consumer to use either the url or the direct props.

snicoll: >  Is the difference between the 2 providers the main concern, or the change in Lettuce connection config behavior?

Both.

wilkinsona:
This has been sitting for several years now without much interest here or on the original issues (#21920 and #22273). As such, I'm going to close it for now. We can reconsider in the future should a strong need for it become apparent.

PR #27409, opened 2021-07-19
Remove redundant deprecation suppressions

dreis2211:
Hi,

just noticed a couple of redundant deprecation suppressions.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #27453, opened 2021-07-22
DataSourceBuilder throws an UnsupportedDataSourcePropertyException when trying to derive a DataSource from an unknown DataSource type

### LINKED ISSUE #27295: DataSourceBuilder throws an UnsupportedDataSourcePropertyException when trying to derive a DataSource from an unknown DataSource type
In our application (Spring Boot 2.5.2) we are trying to use `com.amazonaws.xray.sql.TracingDataSource` as the primary `DataSource`; it's essentially a wrapper around a regular data source to enable AWS X-Ray metrics.
During startup, the application tries to create a derived `DataSource` from it (`DataSourceBuilder.derivedFrom(dataSource)`, the Liquibase starter is doing this).
The `TracingDataSource` is not recognized as a source for `MappedDataSourceProperties`, so we end up with a `ReflectionDataSourceProperties`, trying to to find the getters for the properties by reflection. However, it is erroneously looking for getter methods with one `String` parameter `(DataSourceProperty:293-294)`. Even if such getters were to exist, they are later on (`ReflectionDataSourceProperties:553`) called with no parameters, resulting in an `IllegalArgumentException` anyway.

Example configuration to trigger the issue:
```
  @Bean(name = "dataSource")
  @Primary
  public DataSource tracingDataSource(HikariDataSource dataSource) {
    return new TracingDataSource(dataSource) {
      public String getUrl() {
        return ((HikariDataSource) delegate).getJdbcUrl();
      }
      ...
    };
  }

  @Bean
  public HikariDataSource hikariDataSource(DataSourceProperties properties) {
    HikariDataSource dataSource =
        properties.initializeDataSourceBuilder().type(HikariDataSource.class).build();
    if (StringUtils.hasText(properties.getName())) {
      dataSource.setPoolName(properties.getName());
    }
    return dataSource;
  }

  @Bean
  public DataSource derived(@Qualifier("dataSource") DataSource dataSource) {
    DataSourceBuilder<?> builder =
        DataSourceBuilder.derivedFrom(dataSource).type(SimpleDriverDataSource.class);
    return builder.build(); // UnsupportedDataSourcePropertyException: Unable to find suitable method for url
  }
```

saraswathy-krish:
This fix eliminates the unsupported datasource property exception thrown when trying to derive a datasource from an unknown datasource type  by calling ReflectionUtils on the getters and setters of the unknown datasource type.

Fixes #27295

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
@@ -281,17 +281,22 @@ public final class DataSourceBuilder<T extends DataSource> {
 		}
 
 		Method findSetter(Class<?> type) {
-			return extracted("set", type);
+			return extracted("set", type, true);
 		}
 
 		Method findGetter(Class<?> type) {
-			return extracted("get", type);
+			return extracted("get", type, false);
 		}
 
-		private Method extracted(String prefix, Class<?> type) {
+		private Method extracted(String prefix, Class<?> type, boolean hasParameter) {
 			for (String candidate : this.names) {
-				Method method = ReflectionUtils.findMethod(type, prefix + StringUtils.capitalize(candidate),
-						String.class);
+				Method method;
+				if (hasParameter) {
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks very much, @saraswathy-krish.

PR #27482, opened 2021-07-23
A URI with non-alpha characters in its scheme is not sanitized

billyto:
The Sanitizer.java class infers that the scheme portion of a URI can only contain letters, but as described in[ RFC2396](https://datatracker.ietf.org/doc/html/rfc2396#section-3.1) it can follow:

`scheme        = alpha *( alpha | digit | "+" | "-" | "." )`
 
We have a use case where a valid URI with the scheme `mongodb+sr`, won't get the password token sanitized due to the current regex pattern.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Sanitizer.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Sanitizer.java
@@ -50,7 +50,7 @@ public class Sanitizer {
 	private static final Set<String> URI_USERINFO_KEYS = new LinkedHashSet<>(
 			Arrays.asList("uri", "uris", "url", "urls", "address", "addresses"));
 
-	private static final Pattern URI_USERINFO_PATTERN = Pattern.compile("\\[?[A-Za-z]+://.+:(.*)@.+$");
+	private static final Pattern URI_USERINFO_PATTERN = Pattern.compile("^[A-Za-z][\\w\\+\\.\\-]+://.+:(.*)@.+$");
 
 	private Pattern[] keysToSanitize;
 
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Sanitizer.java
```diff
@@ -50,7 +50,7 @@
 	private static final Set<String> URI_USERINFO_KEYS = new LinkedHashSet<>(
 			Arrays.asList("uri", "uris", "url", "urls", "address", "addresses"));
 
-	private static final Pattern URI_USERINFO_PATTERN = Pattern.compile("\\[?[A-Za-z]+://.+:(.*)@.+$");
+	private static final Pattern URI_USERINFO_PATTERN = Pattern.compile("^[A-Za-z][\\w\\+\\.\\-]+://.+:(.*)@.+$");
```
wilkinsona, line 53: `\w` includes `_` which I don't think is correct here. Should it be `A-Za-z0-9` instead?

wilkinsona (changes requested):
Thanks very much for the pull request, @billyto. I left one comment regarding the updated regex. Could you please take a look when you have a moment?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/Sanitizer.java
billyto: You are totally correct, good catch! Let me fix it in a couple of hours when getting back home.

billyto: done.

wilkinsona (approved):

wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @billyto.

PR #27502, opened 2021-07-27
Update examples in the Dynatrace documentation

pirgeo:
This PR updates the documentation to include Spring Placeholder notation for the v2 exporter in order to make the setup process even easier.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
@@ -182,7 +182,7 @@ The URLs of the Metrics API v2 ingest endpoint are:
 	    export:
 	      dynatrace:
 	        uri: "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest"
-	        api-token: "YOUR_TOKEN" # should be read from a secure source and not hard-coded.
+	        api-token: "YOUR_TOKEN"
 ----
 
 Depending on your setup, the environment variables `+DT_METRICS_INGEST_URL+` and `+DT_METRICS_INGEST_API_TOKEN+` might be populated accordingly already and can be used via the Spring placeholder notation:
@@ -204,17 +204,16 @@ When using the Dynatrace v2 API, the following optional features are available:
 * Default dimensions: Specify key-value pairs that are added to all exported metrics.
   If tags with the same key are specified with Micrometer, they overwrite the default dimensions.
 
+In the example below URI and API token are not specified, therefore the local OneAgent endpoint is used.
+It is also possible to explicitly use the `uri` and `api-token` properties as shown in the examples above.
+
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
 	  metrics:
 	    export:
 	      dynatrace:
-	        # Specify API token and URI or leave blank for OneAgent export, e.g:
-	        # uri: "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest"
-	        # api-token: "YOUR_TOKEN" # should be read from a secure source and not hard-coded.
-
-	        # Configure v2-specific properties:
+	        # Specify uri and api-token here if not using the local OneAgent endpoint.
 	        v2:
 	          metric-key-prefix: "your.key.prefix"
 	          enrich-with-dynatrace-metadata: true
@@ -237,9 +236,7 @@ To export metrics to {micrometer-registry-docs}/dynatrace[Dynatrace], your API t
 	  metrics:
 	    export:
 	      dynatrace:
-	        # uri: "https://{your-domain}/e/{your-environment-id}" on managed deployments.
 	        uri: "https://{your-environment-id}.live.dynatrace.com"
-	        # should be read from a secure property source
 	        api-token: "YOUR_TOKEN"
 	        v1:
 	          device-id: "YOUR_DEVICE_ID"
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
@@ -176,9 +176,13 @@ You must ensure that the endpoint URI contains the path (for example, `/api/v2/m
 	  metrics:
 	    export:
 	      dynatrace:
-	        # uri: "https://{your-domain}/e/{your-environment-id}/api/v2/metrics/ingest" for managed deployments.
-	        uri: "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest"
-	        api-token: "YOUR_TOKEN" # should be read from a secure source and not hard-coded.
+	        # uri: "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest" # for SaaS
+	        # uri: "https://{your-domain}/e/{your-environment-id}/api/v2/metrics/ingest"    # for managed deployments
+	        # or use the Spring placeholder notation to read it from an environment variable:
+	        uri: ${DT_METRICS_INGEST_URL}
+
+	        # should be read from a secure source and not hard-coded.
+	        api-token: ${DT_METRICS_INGEST_API_TOKEN}
 ----
 
 When using the Dynatrace v2 API, the following optional features are available:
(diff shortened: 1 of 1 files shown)


jonatan-ivanov:
@wilkinsona This is slightly related to the recent Dynatrace changes in Boot and Micrometer (that we released today). Do you think we can merge these doc improvements before 2.6.0-M2?

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -175,9 +175,13 @@ Please ensure that the endpoint URI contains the path (e.g. `/api/v2/metrics/ing
 	  metrics:
 	    export:
 	      dynatrace:
-	        # uri: "https://{your-domain}/e/{your-environment-id}/api/v2/metrics/ingest" for managed deployments.
-	        uri: "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest"
-	        api-token: "YOUR_TOKEN" # should be read from a secure source and not hard-coded.
+	        # uri: "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest" # for SaaS
+	        # uri: "https://{your-domain}/e/{your-environment-id}/api/v2/metrics/ingest"    # for managed deployments
```
snicoll, line 179: I am not keen to provide multiple options in the same example. These are not rendered for properties anyway and it feels like this is something that should be described rather than being put in the example directly.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -175,9 +175,13 @@ Please ensure that the endpoint URI contains the path (e.g. `/api/v2/metrics/ing
...
+	        # or use the Spring placeholder notation to read it from an environment variable:
```
snicoll, line 180: I don't think this is warranted. Any property can be used with such a placeholder.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -194,7 +198,11 @@ When using the Dynatrace v2 API, the following optional features are available:
 	  metrics:
 	    export:
 	      dynatrace:
-	        # specify token and uri or leave blank for OneAgent export
+	        # Specify API token and URI or leave blank for OneAgent export, e.g.:
+	        # uri: ${DT_METRICS_INGEST_URL}
```
snicoll, line 202: Again, nothing particular about these properties.

snicoll (changes requested):
Thanks for the PR but I am not keen to add additional information in the form of comments in the example without a good reason to. 

If that's important, perhaps we should have two examples, each of them describing the differences in text rather than in the configuration itself. Please note that those comments are not rendered with the properties format.

The use of spring placeholder is not warranted IMO. It is a general feature that is applicable to any property.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
pirgeo: The idea here was that users could just copy the example and remove the # for the option they want to use.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
pirgeo: In some user's environments this particular property will already be populated. I didn't want to lay out in which cases this applies since that's subject to change and likely too detailed for these docs. How about this?
```
# If the DT_METRICS_INGEST_URL and DT_METRICS_INGEST_API_TOKEN are populated in your environment, you can use these with the Spring placeholder notation:
uri: ${DT_METRICS_INGEST_URL}
# should be read from a secure source and not hard-coded.
api-token: ${DT_METRICS_INGEST_API_TOKEN}
```

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
snicoll: OK but that doesn't address my previous comment.

pirgeo: Oh, I might have missed something here. Do I understand correctly that the comments within the code blocks will not show up in the documentation at all?

pirgeo: @snicoll, Is there something I missed here? Would you be able to provide some guidance on how to proceed?

snicoll: >  Do I understand correctly that the comments within the code blocks will not show up in the documentation at all?

There are not for properties. You can figure that out yourself by looking at the [current documentation](https://docs.spring.io/spring-boot/docs/2.6.0-M2/reference/html/actuator.html#actuator.metrics.export.dynatrace.v2-api). We render the snippet in both `yml` and `properties` format. The latter doesn't have the comment. Ideally I'd like to rework those samples irrespective of this PR to remove the comment so I am not keen to add more comments.

> Would you be able to provide some guidance on how to proceed?

From my perspective, it boils down to my first comment, specifically "it feels like this is something that should be described rather than being put in the example directly.". If we feel these are important enough examples to document, the different options should be described in text with some context, rather than in a comment of a snippet. I've flagged for team attention to get more feedback from the team.

pirgeo: In 57bd19f I moved the additional information to text paragraphs, either way is fine with me. The comments that still persist are optional, meaning that users that copy from the `properties` format still get the expected behavior, but in the `yml` format, they provide additional information.

wilkinsona: We should aim to treat users equally, irrespective of the format that they choose to use for configuration files. If the additional information is needed, I don't think it should be hidden in a comment that's only visible if you're using YAML.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
wilkinsona: > should be read from a secure source and not hard-coded

I don't think this is necessary. There are numerous properties mentioned elsewhere in the documentation where this is the case and we don't have such a comment for each of them.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
pirgeo: @wilkinsona I have removed all comments, except for one that is only relevant for users of the YAML format. Thanks for explaining this, I did not realize that they would not show up in the properties file.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -169,15 +169,31 @@ The {dynatrace-help}/dynatrace-api/basics/dynatrace-api-authentication/[API toke
...
 	  metrics:
 	    export:
 	      dynatrace:
-	        # uri: "https://{your-domain}/e/{your-environment-id}/api/v2/metrics/ingest" for managed deployments.
 	        uri: "https://{your-environment-id}.live.dynatrace.com/api/v2/metrics/ingest"
-	        api-token: "YOUR_TOKEN" # should be read from a secure source and not hard-coded.
+	        api-token: "YOUR_TOKEN"
+----
+
+Depending on your setup, the environment variables `+DT_METRICS_INGEST_URL+` and `+DT_METRICS_INGEST_API_TOKEN+` might be populated accordingly already and can be used via the Spring placeholder notation:
```
snicoll, line 187: What does "depending on your setup" mean? I think that if something is setting those, perhaps we should link to whatever is setting them rather and put that doc there? I am mentioning this as I would prefer that our documentation does not provide too much information for something that would be better covered elsewhere anyway.

snicoll (commented):
Thanks for the update. I am still not sure we should document the `DT_METRICS_*` properties. I don't know how and when it would apply to a user reading the doc...

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
pirgeo: @snicoll I have removed the example for now since I don't have any good reference to put into the docs. Thanks for pointing out that that might be unclear for readers!

PR #27539, opened 2021-08-01
Switch to Adoptium JDK builds in CI

dreis2211:
Hi,

this PR switches the AdoptOpenJDK builds over to the new Adoptium builds. The side-effect of this are updated versions as the previous equivalents are not available anymore.

For completeness reasons, I should also mention that the previous JDK 17 build didn't include JEP 403 yet, because apparently AdoptOpenJDK didn't update things for quite a while anymore. My local tests used always the latest OpenJDK builds (30 or 32) with JEP 403 included and were successful, so I'm hopeful CI will be as well. But you never know.

Cheers,
Christoph

### DIFF
--- a/ci/images/get-jdk-url.sh
+++ b/ci/images/get-jdk-url.sh
@@ -3,16 +3,16 @@ set -e
 
 case "$1" in
 	java8)
-		 echo "https://github.com/AdoptOpenJDK/openjdk8-binaries/releases/download/jdk8u292-b10/OpenJDK8U-jdk_x64_linux_hotspot_8u292b10.tar.gz"
+		 echo "https://github.com/adoptium/temurin8-binaries/releases/download/jdk8u302-b08/OpenJDK8U-jdk_x64_linux_hotspot_8u302b08.tar.gz"
 	;;
 	java11)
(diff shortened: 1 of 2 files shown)


dreis2211:
The AdoptOpenJDK will shortly redirect to the newest Adoptium builds. Hence the CI created all tickets for JDK 8 - 17 already in the meantime. I guess this leaves this PR only with the update to JDK 17 from Adoptium

snicoll:
It didn't create them for JDK 17, which surprised me a bit. I'll rebase what's left on `2.3.x` onward. Thanks again Christoph!

dreis2211:
@snicoll Thanks. I just noticed that I wrote JDK 8-17, but in fact it is JDK 8-16. The JDK 17 stuff is just not redirected to the Adoptium builds when accessing the AdoptOpenJDK API, so the update detection job didn't notice any change.

snicoll:
It doesn't work for JDK 15 either (whereas the old api does). It's not big deal for that one but we won't pick up new JDK 17 build if we merge this as far as I understand?

dreis2211:
Why would it not pick up new builds for JDK 17 once new builds are available? Well, apart from being on the latest available build after merging the PR. Or did I get your question wrong, @snicoll ?

snicoll:
Sorry, I was using the `ga` url against 17, that's not going to work very well. All good, sorry for the noise.

PR #27546, opened 2021-08-02
Fix checkstyle violation

dreis2211:


dreis2211:
Apparently, @bclozel fixed it already.

PR #27592, opened 2021-08-09
Add ErrorWebFluxAutoConfiguration to Actuator ManagementContext

### LINKED ISSUE #27504: When a WebFlux-based management server is running on a separate port, the error infrastructure is not used
With the main server on 8080 and the management server on 8081, the error responses are different:

```
$ http :8080/404
HTTP/1.1 404 Not Found
Content-Length: 132
Content-Type: application/json

{
    "error": "Not Found",
    "message": null,
    "path": "/404",
    "requestId": "e06c56d2-1",
    "status": 404,
    "timestamp": "2021-07-27T09:49:12.467+00:00"
}

```

```
$ http :8081/404
HTTP/1.1 404 Not Found
content-length: 0

```

yao-yuan-yy:
Fix #27504

The reason of this bug is managment ApplicationContext do not get order in `@Order(-1) DefaultErrorWebExceptionHandler` in parent ApplicationContext.
`@Order(0) ResponseStatusExceptionHandler` has higher level in ExceptionHandlers.
so let us config ErrorWebFluxAutoConfiguration in ManagementContextConfiguration, the DefaultErrorWebExceptionHandler will be the highest ExceptionHandler and the response body will not be null.


wilkinsona:
Thanks for the proposal, but I don't think this is the right way to fix this one. An entry listed beneath `org.springframework.boot.actuate.autoconfigure.web.ManagementContextConfiguration` should be annotated with `@ManagementContextConfiguration` rather than being a plain `@Configuration` class. The configuration should also only apply when a separate child management context is being used. Thanks anyway.

PR #27613, opened 2021-08-09
Add option to create tags for the built image

### LINKED ISSUE #20560: Apply multiple docker image tags when publishing images
As of  #19698 spring boot has native support for building docker images which is a great approach to further simplify the build process.

When publishing docker images a common approach is to provide multiple tags per image. Probably most common is to publish the actual version along with an additional `latest` tag. I'd love to see Spring Boot support multiple tags per image (that would also simplify migrating from the also great `jib` plugin which already supports multiple tags)

ceccone:
This commit adds configuration to the Maven and Gradle plugins to allow specifying multiple tag names to be created for the built image.

Resolves #20560

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/AbstractBuildLog.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/AbstractBuildLog.java
@@ -31,6 +31,7 @@ import org.springframework.boot.buildpack.platform.docker.type.VolumeName;
  * @author Phillip Webb
  * @author Scott Frederick
  * @author Andrey Shlykov
+ * @author Rafael Ceccone
  * @since 2.3.0
  */
 public abstract class AbstractBuildLog implements BuildLog {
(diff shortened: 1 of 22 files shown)


scottfrederick:
@ceccone Thank you very much for your contribution. It has been merged along with a [polishing commit](https://github.com/spring-projects/spring-boot/commit/64c49003aaf43caf937df605e47022f1e884a319) and some changes to `DockerApi` to make tests pass.

nithril:
Hello,

It does not seems possible to publish multiple tags by using a maven properties (like with jib `Use -Djib.to.tags=a,b,c`)

Could you please confirm if my understanding is correct?

scottfrederick:
@nithril Your understanding is correct. In the updated [snapshot documentation](https://docs.spring.io/spring-boot/docs/2.6.0-SNAPSHOT/maven-plugin/reference/htmlsingle/#build-image.customization), there is no `(User Property)` provided for tags, so any additional tags must be provided in the `pom.xml` build configuration.

VitaliyKulikov:
something wrong has been implemented for option: `tags`.
let's say we have the following:

```
  tags = listOf("bbb", "aaa")
```
instead of creating images, like: <image_name>:bbb, <image_name>:aaa
new images will be created bbb:latest, aaa:latest

 id("org.springframework.boot") version "2.7.5"

PR #27615, opened 2021-08-10
Add .gitattributes

Goooler:
See [Add .gitattributes](https://github.com/square/okhttp/pull/6773).

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,4 @@
+* text=auto eol=lf
+
+*.bat text eol=crlf
+*.jar binary


wilkinsona:
I think we need to tread carefully here. CI currently uses `core.autocrlf=true`:

https://github.com/spring-projects/spring-boot/blob/8966b1308b17f1a8c4e50b25508f4b17e10fa20f/ci/pipeline.yml#L127-L129

I'm not sure how this change will interact with that configuration.

Goooler:
> I think we need to tread carefully here. CI currently uses `core.autocrlf=true`:
> 
> https://github.com/spring-projects/spring-boot/blob/8966b1308b17f1a8c4e50b25508f4b17e10fa20f/ci/pipeline.yml#L127-L129
> 
> I'm not sure how this change will interact with that configuration.

.gitattributes should just effect files when commiting.

philwebb:
https://git-scm.com/docs/gitattributes#_checking_out_and_checking_in seems to suggest they also change checked out files.

Goooler:
If we should run CI for test once ?

PR #27626, opened 2021-08-10
feat: spring.config.exclude-file-extensions

qxo:
if custom project do not have some file extensions for the config, we use the property to exclude these execution(such as the file exists)
ie:
-Dspring.config.exclude-file-extensions=xml,json,yaml
means we do not have xml/json/yaml extensions for the spring boot config

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigFileApplicationListener.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigFileApplicationListener.java
@@ -296,6 +296,8 @@ public class ConfigFileApplicationListener implements EnvironmentPostProcessor, 
 
 		private final List<PropertySourceLoader> propertySourceLoaders;
 
+		 private final Set<String> excludeFileExtentions;
+
 		private Deque<Profile> profiles;
 
 		private List<Profile> processedProfiles;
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the suggestion, but I'm not sure that I understand the benefit of being able to exclude certain file extensions. Can you please expand a bit on what you would use it for and why it would be useful?

qxo:
> Thanks for the suggestion, but I'm not sure that I understand the benefit of being able to exclude certain file extensions. Can you please expand a bit on what you would use it for and why it would be useful?

ie: `spring.config.exclude-file-extensions=xml,json,yaml,properties`
means we skip the check file exist of  these file extensions, cause  the project do not have xml/json/yaml/properties as spring boot config file.

with the PR, switch logging (logging.level.org.springframework.boot.context.config=TRACE）
You will see the different: we do not have these file extensions, so there no need for check whether if the exists. -- reduce unnecessary invoke

qxo:
On test spring-boot project without  `-Dspring.config.exclude-file-extensions=xml,json,yaml,properties`
we have 95 lines of "o.s.b.c.c.ConfigFileApplicationListener  : Skipped missing config..."
with  `-Dspring.config.exclude-file-extensions=xml,json,yaml,properties` there only 15 line.

then with '-Dspring.config.location=classpath:/` we do not have “: Skipped missing config..."

wilkinsona:
Thanks for the additional details. Have you run any benchmarks to see if there's a noticeable improvement in startup time?

qxo:
> Thanks for the additional details. Have you run any benchmarks to see if there's a noticeable improvement in startup time?

Not yet. Maybe a few ms:)

wilkinsona:
Ok. Without some benchmarks that show a pretty significant improvement, I don't think we'll be in a position to consider the additional complexity of the new configuration option.

PR #27628, opened 2021-08-11
Add ErrorWebFluxAutoConfiguration to ReactiveManagementChildContextConfiguration

### LINKED ISSUE #27504: When a WebFlux-based management server is running on a separate port, the error infrastructure is not used

yao-yuan-yy:
use nested `@Configuration` instead of spring.factories.
now this configuration only apply when a separate child management context is being used.
If we use same management port, `ErrorWebFluxAutoConfiguration` will not be used

Fix #27504


wilkinsona:
Thanks for having another go, @yao-yuan-yy. I'm afraid that I don't think this is the right way to fix the problem either. Unfortunately, I don't know exactly how we should fix it – otherwise I would have suggested something more specific in https://github.com/spring-projects/spring-boot/pull/27592#issuecomment-895110773 – so we'll have to wait for the discussion on #27504 to reach a conclusion. Thanks again and sorry for the wasted time.

PR #27659, opened 2021-08-14
Added property to customise Jackson default leniency

### LINKED ISSUE #27547: Add application property to customize Jackson default leniency
Jackson provides a leniency setting for deserialization, which can be configured per property (using [`@JsonFormat(lenient=OptBoolean.FALSE`)](http://fasterxml.github.io/jackson-annotations/javadoc/2.12/com/fasterxml/jackson/annotation/JsonFormat.html#lenient--)) or globally (per [`ObjectMapper.setDefaultLeniency(false)`](https://fasterxml.github.io/jackson-databind/javadoc/2.12/com/fasterxml/jackson/databind/ObjectMapper.html#setDefaultLeniency-java.lang.Boolean-)).

This flag will become more useful as of (the not yet released) Jackson 2.13 per https://github.com/FasterXML/jackson-modules-java8/issues/212, as disabling it will cause Jackson to reject dates with time components (e.g. "2007-12-03T10:15:30", "2007-12-03T10:15:30Z") when deserializing a `LocalDate` value.  Disallowing these time components can be useful, as including them can lead to unintended issues truncating the time part results in the wrong date due to time zone discrepancies between the source & target systems.

In light of this, it would be handy if there were a way to set the default value of this lenient property through a Spring Boot application property.

## Workaround

Right now, this property can be configured using  a custom [`Jackson2ObjectMapperBuilderCustomizer`](https://docs.spring.io/spring-boot/docs/2.5.3/api/org/springframework/boot/autoconfigure/jackson/Jackson2ObjectMapperBuilderCustomizer.html) bean.  Since Spring's [`Jackson2ObjectMapperBuilder`](https://docs.spring.io/spring-framework/docs/5.3.9/javadoc-api/org/springframework/http/converter/json/Jackson2ObjectMapperBuilder.html?is-external=true) type does not directly expose this property, a post-configurer is probably the simplest way to set this:

```java
@Configuration
public class CustomizedJacksonConfiguration {
    @Bean
    public Jackson2ObjectMapperBuilderCustomizer nonLenientObjectMapperBuilderCustomizer() {
        return builder -> builder.postConfigurer(
                objectMapper -> objectMapper.setDefaultLeniency(false));
    }
}
```

lower-case:
Fixes #27547

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java
@@ -187,6 +187,7 @@ public class JacksonAutoConfiguration {
 				configurePropertyNamingStrategy(builder);
 				configureModules(builder);
 				configureLocale(builder);
+				configureLeniency(builder);
 			}
 
 			private void configureFeatures(Jackson2ObjectMapperBuilder builder, Map<?, Boolean> features) {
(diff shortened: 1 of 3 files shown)


snicoll:
@lower-case thank you for your contribution. I've polished it in 68a47a7.

Issue #27306, opened 2021-07-13
Provide a configuration property for setting the path used by auto-configured disk space metrics


wilkinsona:
The path for disk space metrics is hardcoded as `.` but we allow the path for the disk space health indicator to be auto-configured. I think the path for the metrics should be equally configurable. I wonder if it might be possible to use the same property for both as I suspect it's unlikely that you'd want the two to be different.

When looking at this, we should keep [the request to support multiple paths for the disk space health check](https://github.com/spring-projects/spring-boot/issues/18359) in mind. I suspect that the requirement to have a health check for multiple paths would apply equally to metrics.

bono007:
@wilkinsona 

**QUESTION**: Somewhat unrelated, but I have been curious about this before. Are `DiskSpaceMetrics` really JVM metrics, or more aligned w/ SystemMetrics? The `DiskSpaceHealthIndicator` seems to align more w/ the latter as it lives in a `system` sub-package. 

**ASSUMPTION**: I am assuming a breaking change is ok as this is going into 2.6.

#### THOUGHTS
So this is interesting as the thought of sharing the same property would be the first bridge of a metric and a health indicator properties that I am aware of. This is the typical layout of the properties if they are not shared: 

- `DiskSpaceHealthIndicatorProperties` -> "`management.health.diskspace.path`"
- `MetricsProperties` (typical metrics property home) -> "`management.metrics.diskspace.path`" 

I also took a read through the [request to support multiple paths for the disk space health check](https://github.com/spring-projects/spring-boot/issues/18359) and that would also complicate a sharing of properties as it requires additional settings for `threshold`. One possible unshared properties layout might look like:

```
management.metrics.diskspace.paths: a,b,c

management.health.diskspace.paths:
  - path: a
    threshold: .8
  - path: b 
    threshold: .7
  - path: c
    threshold: .5
```

I am thinking that it may be better to keep these separated. I would think the number of paths configured would be pretty small and therefore having to configure them twice is probably not that much of a pain. Do we have any metrics or idea on how heavily these 2 features are used?

wilkinsona:
[...]

PR #27660, opened 2021-08-15
Provide a configuration property for setting the path used by auto-configured disk space metrics

bono007:
@wilkinsona I was playing around w/ the options listed in the ticket and had this simple approach coded up so went ahead and submitted this proposal. I know I did not wait for feedback on the ticket and am more than happy to close this out if If you end up leaning another direction.

This goes w/ the simple approach of not sharing properties w/ the `DiskSpaceHealthIndicator` and only configures a single path for now (in the `MetricsProperties`). 

Fixed gh-27306

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
@@ -354,16 +354,16 @@ public class MetricsProperties {
 
 	public static class System {
 
-		private final DiskSpace diskspace = new DiskSpace();
+		private final Diskspace diskspace = new Diskspace();
 
-		public DiskSpace getDiskspace() {
+		public Diskspace getDiskspace() {
 			return this.diskspace;
 		}
 
-		public static class DiskSpace {
+		public static class Diskspace {
 
 			/**
-			 * Paths to report disk metrics for.
+			 * Comma-separated list of paths to report disk metrics for.
 			 */
 			private List<File> paths = Arrays.asList(new File("."));
 
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
@@ -16,8 +16,11 @@
 
 package org.springframework.boot.actuate.autoconfigure.metrics;
 
+import java.io.File;
 import java.time.Duration;
+import java.util.Arrays;
 import java.util.LinkedHashMap;
+import java.util.List;
 import java.util.Map;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
@@ -30,6 +33,7 @@ import org.springframework.boot.context.properties.NestedConfigurationProperty;
  * @author Jon Schneider
  * @author Alexander Abramov
  * @author Tadaya Tsuyukubo
+ * @author Chris Bono
  * @since 2.0.0
  */
 @ConfigurationProperties("management.metrics")
@@ -56,6 +60,8 @@ public class MetricsProperties {
 	private final Web web = new Web();
 
 	private final Data data = new Data();
+
+	private final System system = new System();
 
 	private final Distribution distribution = new Distribution();
 
@@ -85,6 +91,10 @@ public class MetricsProperties {
 
 	public Distribution getDistribution() {
 		return this.distribution;
+	}
+
+	public System getSystem() {
+		return this.system;
 	}
 
 	public static class Web {
@@ -338,6 +348,33 @@ public class MetricsProperties {
 
 		public Map<String, Integer> getBufferLength() {
 			return this.bufferLength;
+		}
+
+	}
+
+	public static class System {
+
+		private final DiskSpace diskspace = new DiskSpace();
+
+		public DiskSpace getDiskspace() {
+			return this.diskspace;
+		}
+
+		public static class DiskSpace {
+
+			/**
+			 * Paths to report disk metrics for.
+			 */
+			private List<File> paths = Arrays.asList(new File("."));
+
+			public List<File> getPaths() {
+				return this.paths;
+			}
+
+			public void setPaths(List<File> paths) {
+				this.paths = paths;
+			}
+
 		}
 
 	}
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/system/DiskSpaceMetricsBinder.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/system/DiskSpaceMetricsBinder.java
@@ -1,0 +1,52 @@
+/*
+ * Copyright 2012-2021 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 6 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
```diff
@@ -318,4 +326,31 @@ public AutoTimeProperties getAutotime() {
 
 	}
 
+	public static class Jvm {
```
bono007, line 329: So far all instrumented metrics seem to have their config props in here. It does feel like it is starting to get a bit overloaded, especially w/ this addition. My gut instinct is to create a separate `JvmMetricsConfigurationProperties` (still prefixed at `management.metrics.jvm`) but I was unsure if it is intentional to have all things metrics config props to be in here.

bono007:
@wilkinsona I rebased and moved from jvm->system

wilkinsona:
@bono007 Are you interested in reworking this to support multiple paths?  In light of https://github.com/spring-projects/spring-boot/issues/27306#issuecomment-903907394, that can be done without considering the disk space health indicator.  I think we'd probably end up with a `MetricBinder` implementation that consumes the paths. When called, it would create one `DiskSpaceMetrics` instance per path and binds it to the registry.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
```diff
@@ -342,4 +349,30 @@ public AutoTimeProperties getAutotime() {
 
 	}
 
+    public static class System {
+
+        private final DiskSpace diskSpace = new DiskSpace();
```
wilkinsona, line 354: This should be named `Diskspace` to avoid the `-` in the configuration property.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
```diff
@@ -342,4 +349,30 @@ public AutoTimeProperties getAutotime() {
...
+
+        public DiskSpace getDiskSpace() {
+            return this.diskSpace;
+        }
+
+        public static class DiskSpace {
```
wilkinsona, line 360: This should be named Diskspace to avoid the - in the configuration property.

bono007:
> @bono007 Are you interested in reworking this to support multiple paths? In light of [#27306 (comment)](https://github.com/spring-projects/spring-boot/issues/27306#issuecomment-903907394), that can be done without considering the disk space health indicator. I think we'd probably end up with a `MetricBinder` implementation that consumes the paths. When called, it would create one `DiskSpaceMetrics` instance per path and binds it to the registry.

You know I am 😺 

I like the MeterBinder -> N DiskSpaceMetrics approach as well. This also allows us to close https://github.com/micrometer-metrics/micrometer/pull/2747#issuecomment-903915522 . 

I will get to it in the next 1-2 days. That timeline work?

wilkinsona:
> You know I am 😺

😀 I didn’t want to assume you would be

> I will get to it in the next 1-2 days. That timeline work?

Absolutely. It’d be good to get the change in before RC1.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
```diff
@@ -342,4 +352,31 @@ public AutoTimeProperties getAutotime() {
...
+		public DiskSpace getDiskspace() {
+			return this.diskspace;
+		}
+
+		public static class DiskSpace {
+
+			/**
+			 * Paths to report disk metrics for.
+			 */
+			private List<File> paths = Arrays.asList(new File("."));
```
bono007, line 368: `management.metrics.diskspace.path` or  `management.metrics.diskspace.paths`? 

Of course, the singular reads better when there is a single path. 
```yaml
management.metrics.diskspace.path: path1
```

And the plural reads better when there are multiple paths.
```yaml
management.metrics.diskspace.paths:
  - path1
  - path2
```

snicoll: That should be `paths`,  I think with "Comma-separated list of paths..." for the description.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
```diff
@@ -342,4 +352,31 @@ public AutoTimeProperties getAutotime() {
...
 
+	public static class System {
+
+		private final DiskSpace diskspace = new DiskSpace();
+
+		public DiskSpace getDiskspace() {
+			return this.diskspace;
+		}
+
+		public static class DiskSpace {
```
wilkinsona, line 363: I still think it'd be good to rename this to `Diskspace`. That'd be consistent with what we've done elsewhere. See `org.springframework.boot.autoconfigure.web.WebProperties.Resources.Cache.Cachecontrol`, for example.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/system/DiskSpaceMetricsBinder.java
```diff
@@ -0,0 +1,52 @@
...
+
+import org.springframework.util.Assert;
+
+/**
+ * A {@link MeterBinder} that binds one or more {@link DiskSpaceMetrics}.
+ *
+ * @author Chris Bono
+ * @since 2.6.0
+ */
+public class DiskSpaceMetricsBinder implements MeterBinder {
```
wilkinsona, line 35: Just to be clear, I would not change to capitalization of this class's name as it aligns with Micrometer's `DiskSpaceMetrics` at the moment.

wilkinsona (changes requested):

snicoll:
Thanks Chris, that's excellent. FYI, you might be interested by the [polish commit](https://github.com/spring-projects/spring-boot/commit/b626adf1ee35133289bfca0d6e3d057b6de4803e), in particular:

* The `List` configuration property must be mutable if the binder needs to append to the list depending on the configuration style
* The annotation processor can't detect a default value expressed this way so I've added a manual hint.

bono007:
> Thanks Chris, that's excellent. FYI, you might be interested by the [polish commit](https://github.com/spring-projects/spring-boot/commit/b626adf1ee35133289bfca0d6e3d057b6de4803e), in particular:
> 
You're welcome Stéphane. And thank you for taking the time to point me to the polish commit and outline what changed - much appreciated. 

> * The `List` configuration property must be mutable if the binder needs to append to the list depending on the configuration style

Yep, I know about this one and was almost certain that `Arrays.asList` produced a mutable list as the code looks like this:
```java
    public static <T> List<T> asList(T... a) {
        return new ArrayList<>(a);
    }
````
⚠️ However, looking closer one will notice that is not a `java.util..ArrayList<T>` it returns but rather a `java.util.Arrays.ArrayList` private imutable impl. Wow, thanks for the clever naming JDK8. 


> * The annotation processor can't detect a default value expressed this way so I've added a manual hint.

Good catch. I think this crossed my mind at one point as something to follow up on as I was unsure if it would find that complex default (obviously I forgot to circle back :) )

PR #27661, opened 2021-08-15
Use Pattern for Groovy closure checks

dreis2211:
Hi,

this PR changes the checks for Groovy closures in `BeanDefinitionLoader` to a precompiled Pattern. It doesn't matter really from a performance perspective, because it's only called once in standard Boot applications. But in theory it could be called more often and then it saves the repeated pattern compilation overhead. Still very minor, so feel free to decline.

Cheers,
Christoph


snicoll:
Thanks again Christoph.

PR #27663, opened 2021-08-16
Add support for multiple paths in DiskSpaceHealthIndicator

bono007:
Fixed gh-18359

#### Assumption
I know there was a previous proposal that was closed due to concerns around breaking compatibility of the actuator output when configuring each path's threshold. This code proposal assumes that breaking the output is ok since this is going into 2.6. 

#### Etiquette
I am not sure what the etiquette is in the situation where there is a dated closed proposal. I am starting to work on the [counterpart of this for the DiskSpaceMetrics](https://github.com/spring-projects/spring-boot/issues/27306) and figured I would put this together. If instead, we should resurrect the elder proposal I am totally fine w/ that as well. This newer implementation does support configuring the threshold per each path. I am not trying to be rude or step on toes. 

#### Usage

The config props now look like the following:
```yaml
management.health.diskspace.paths:

  - path: /some/dir/path1
    threshold: 10GB

  - path: /some/other/path2
  
  - path: /some/extra/path3
    threshold: 100MB
```
* I am not super excited about the "paths.path" 
* With low effort we could also keep the existing "management.health.diskspace.path" for the singular case. Thoughts?

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/system/DiskSpaceHealthIndicatorProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/system/DiskSpaceHealthIndicatorProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,6 +17,8 @@
 package org.springframework.boot.actuate.autoconfigure.system;
 
 import java.io.File;
+import java.util.Arrays;
+import java.util.List;
 
 import org.springframework.boot.actuate.system.DiskSpaceHealthIndicator;
 import org.springframework.boot.context.properties.ConfigurationProperties;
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/system/DiskSpaceHealthIndicatorProperties.java
```diff
@@ -28,36 +30,54 @@
...
  */
 @ConfigurationProperties(prefix = "management.health.diskspace")
 public class DiskSpaceHealthIndicatorProperties {
 
 	/**
-	 * Path used to compute the available disk space.
+	 * Paths to consider for computing the available disk space.
 	 */
-	private File path = new File(".");
+	private List<PathInfo> paths = Arrays.asList(new PathInfo());
```
bono007, line 42: As stated in the overview comment, we could keep support of the `management.health.diskspace.path` (for singular case only) by mapping a `private PathInfo path = new PathInfo();` here as well.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/system/DiskSpaceHealthContributorAutoConfigurationTests.java
```diff
@@ -71,4 +99,13 @@ void runWhenDisabledShouldNotCreateIndicator() {
 				.run((context) -> assertThat(context).doesNotHaveBean(DiskSpaceHealthIndicator.class));
 	}
 
+	@SafeVarargs
```
bono007, line 102: Yuk....

Alternative is to use `List<Map.Entry<K, V> entries>` which then leads to the tests having to use something like `Arrays.asList(...)`.  I went w/ convenience of callers namely because this is test code.

wilkinsona:
> This code proposal assumes that breaking the output is ok since this is going into 2.6

The breaking changes we rejected previously, not because of their timing but because there hadn't been much demand for them. As such, it was hard to justify the cost of the breaking change. The situation has changed a little bit since then as the request for disk space metrics is related to this and I think it makes sense for the two to be in sync in terms of their configuration. I'm not sure if that warrants making a breaking change in 2.6, if we should defer everything till 3.0, or if we should ship disk space metrics with configuration that's out of sync with disk space health. Flagging for team meeting so that we can discuss our options once again.

wilkinsona:
Thanks for the proposal here, @bono007. In light of https://github.com/spring-projects/spring-boot/issues/27306#issuecomment-903907394, we no longer want to add support for configuring the disk space health indicator with multiple paths.

PR #27698, opened 2021-08-16
NullPointerException in RoutingDataSourceHealthContributor when a routing data source has a target with a null routing key

### LINKED ISSUE #27694: RoutingDataSourceHealthContributor should heed Maps with null values in keys
https://github.com/spring-projects/spring-boot/blob/b65cc4d62f4a818e36be8b99b98eba01a28f6f2a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java#L130

I just upgraded our project from springboot 2.4.1 to 2.5.3. I am super exited, that there is support for AbstractRoutingDataSource now. Thank' and kudos for that! 
Due to the update the health check now run's into an NPE which kills the whole application.
It happens because the entryset 'routingDataSource.getResolvedDataSources().entrySet()' contains null as a key.

We create our datasource like this:
```
@Bean
@Primary
public DataSource dataSource() {
  final RoutingDataSource routingDataSource = new RoutingDataSource();

  final DataSource primaryDataSource = buildDataSource("PrimaryHikariPool", primaryDataSourceProperties());
  final DataSource replicaDataSource = buildDataSource("ReplicaHikariPool", replicaDataSourceProperties());

  final Map<Object, Object> targetDataSources = new HashMap<>();
  // HERE WE PUT IN OUR EVIL null KEY
  targetDataSources.put(null, primaryDataSource);
  targetDataSources.put(RoutingDataSource.Route.REPLICA, replicaDataSource);

  routingDataSource.setTargetDataSources(targetDataSources);
  routingDataSource.setDefaultTargetDataSource(primaryDataSource);

  return routingDataSource;
}
```
It might be strange or even stupid to put 'null' as a key into a map, but the RoutingDataSourceHealthContributor shouldn't crash because of that anyway.

thegeekyasian:
The RoutingDataSourceHealthContributor constructor throws a NullPointerException (NPE) if the provided map has a null key. With this commit the NPE issue has been fixed by filtering the entries that are not-null. 

Alternatively, a constant value can be added when the key for a datasource in the ‘Map’ is null. 

This closes #27694

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/jdbc/DataSourceHealthContributorAutoConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.boot.actuate.autoconfigure.jdbc;
 import java.util.Collection;
 import java.util.Iterator;
 import java.util.Map;
+import java.util.Optional;
 import java.util.function.Function;
 import java.util.stream.Collectors;
 
@@ -125,11 +126,13 @@ public class DataSourceHealthContributorAutoConfiguration implements Initializin
 
 		private final CompositeHealthContributor delegate;
 
+		private static final String UNNAMED_DATASOURCE_KEY = "unnamed";
+
 		RoutingDataSourceHealthContributor(AbstractRoutingDataSource routingDataSource,
 				Function<DataSource, HealthContributor> contributorFunction) {
 			Map<String, DataSource> routedDataSources = routingDataSource.getResolvedDataSources().entrySet().stream()
-					.filter((e) -> e.getKey() != null)
-					.collect(Collectors.toMap((e) -> e.getKey().toString(), Map.Entry::getValue));
+					.collect(Collectors.toMap((e) -> Optional.ofNullable(e.getKey()).map(Object::toString)
+									.orElse(UNNAMED_DATASOURCE_KEY), Map.Entry::getValue));
 			this.delegate = CompositeHealthContributor.fromMap(routedDataSources, contributorFunction);
 		}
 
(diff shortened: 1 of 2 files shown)


philwebb:
Thanks for the pull-request @thegeekyasian. I'm wondering if we'd be better off changing `e.getKey().toString()` to something else so that `null` entries aren't removed? Perhaps `null` could be replaced with "unnamed" or something similar.

We should also add some tests to `DataSourceHealthContributorAutoConfigurationTests`.

Are you interested in updating your pull request in that direction?

thegeekyasian:
> Thanks for the pull-request @thegeekyasian. I'm wondering if we'd be better off changing `e.getKey().toString()` to something else so that `null` entries aren't removed? Perhaps `null` could be replaced with "unnamed" or something similar.
> 
> We should also add some tests to `DataSourceHealthContributorAutoConfigurationTests`.
> 
> Are you interested in updating your pull request in that direction?

Thanks @philwebb  Definitely, we can replace the null one with a constant string value 'unnamed'. I'd like to do the suggested changes and update the PR.

In addition to this, do you think a public method should be exposed to getUnnamedDataSource() in `RoutingDataSourceHealthContributor` ?

philwebb:
@thegeekyasian I don't think `null` keys will work because of [this assertion](https://github.com/spring-projects/spring-boot/blob/b65cc4d62f4a818e36be8b99b98eba01a28f6f2a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/NamedContributorsMapAdapter.java#L46).

thegeekyasian:
Ah, I see! 'unnamed' key is the way to go then.

Should a warning be logged in this case though, when setting the datasource for unnamed?

philwebb:
From @timmalich in https://github.com/spring-projects/spring-boot/issues/27694#issuecomment-899860315

> Thanks @thegeekyasian, I've added a comment to the pull-request.

> @philwebb, a PR has been created to fix the null pointer by filtering the entries with the non-null only. Can you please review and share your thoughts?

Thank you both very much. 

Wouldn't that filter just drop an otherwise valid configuration? Wouldn't it be better to check null within the next line? Like:
e.getKey()==null?null:e.getKey().toString()

Option 2) Would it be possible to at least log a warning?

philwebb:
I think the warning is unnecessary since the health info name is only for reference. The only problem might be if `Map` happens to contain `null` and `"unnamed"` which seem quite unlikely.

thegeekyasian:
@philwebb, the PR has been updated with the suggested changes and the tests have been update to assert the 'unnamed' datasource.

wilkinsona:
@thegeekyasian Thanks very much for making your first contribution to Spring Boot.

PR #27734, opened 2021-08-17
Use "2012" for copyright beginning year

izeye:
This PR changes to use "2012" for the copyright beginning year as it seems to be a typo of "2012" based on the copyright headers in other Java files.


snicoll:
What have we done 😂

wilkinsona:
Thanks very much, @izeye. There were several others where the beginning year was also wrong, but with a value other than 2021. I've fixed those in https://github.com/spring-projects/spring-boot/commit/0b7994a679c8d47d3b4f5a42615b0c0211232b51.

PR #27736, opened 2021-08-17
Polish access modifiers for test classes

izeye:
This PR polishes access modifiers for test classes.


philwebb:
I thought we has a checkstyle for this at one point. Perhaps it's not working or I'm misremembering.

dreis2211:
@philwebb I think this just checks for method visibility. See https://github.com/spring-io/spring-javaformat/blob/main/spring-javaformat/spring-javaformat-checkstyle/src/main/java/io/spring/javaformat/checkstyle/check/SpringJUnit5Check.java#L153

philwebb:
Thanks @dreis2211!

wilkinsona:
Thanks very much, @izeye.

PR #27737, opened 2021-08-17
Only link to the two alternative formats from each format of the reference documentation

Buzzardo:
And some small editing changes.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/index.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/index.adoc
@@ -4,8 +4,7 @@ include::authors.adoc[]
 v{spring-boot-version}
 include::attributes.adoc[]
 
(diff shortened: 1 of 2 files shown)


philwebb:
We should double check the condition works because I'm not sure if the new backend sets the same property name.

wilkinsona:
@Buzzardo For future reference, we try to keep individual commits focussed on a single change. This PR's mixing improvements to the links to the other forms of the documentation with some general editing of the multi-page documentation's index. I'll split things up while merging this and have happened https://github.com/spring-projects/spring-boot/issues/27805 to track the editorial changes.

wilkinsona:
> We should double check the condition works because I'm not sure if the new backend sets the same property name.

Good catch, Phil. As proposed, the changes remove the sentence beginning "This document is also available as …" from both the single-page documentation and from the PDF.

wilkinsona:
Thanks, @Buzzardo. I've merged these changes into main.

For future reference, it'd be greatly appreciated if you could build your changes (`./gradlew spring-boot-project:spring-boot-docs:build` should do the trick) before submitting them and double-check that they're rendered as expected.

PR #27759, opened 2021-08-18
Editing pass for the Actuator section

Buzzardo:

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/cloud-foundry.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/cloud-foundry.adoc
@@ -15,14 +15,12 @@ To use the endpoint, you must pass a valid UAA token with the request.
 === Disabling Extended Cloud Foundry Actuator Support
 If you want to fully disable the `/cloudfoundryapplication` endpoints, you can add the following setting to your `application.properties` file:
 
-====
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
 	  cloudfoundry:
 	    enabled: false
 ----
-====
 
 
 
@@ -31,14 +29,12 @@ If you want to fully disable the `/cloudfoundryapplication` endpoints, you can a
 By default, the security verification for `/cloudfoundryapplication` endpoints makes SSL calls to various Cloud Foundry services.
 If your Cloud Foundry UAA or Cloud Controller services use self-signed certificates, you need to set the following property:
 
-====
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
 	  cloudfoundry:
 	    skip-ssl-validation: true
 ----
-====
 
 
 
@@ -51,9 +47,7 @@ If you expect the Cloud Foundry endpoints to always be available at `/cloudfound
 The configuration differs, depending on the web server in use.
 For Tomcat, you can add the following configuration:
 
-====
 [source,java,indent=0,subs="verbatim"]
 ----
 include::{docs-java}/actuator/cloudfoundry/customcontextpath/MyCloudFoundryConfiguration.java[]
 ----
-====
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/cloud-foundry.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/cloud-foundry.adoc
@@ -4,10 +4,10 @@ Spring Boot's actuator module includes additional support that is activated when
 The `/cloudfoundryapplication` path provides an alternative secured route to all `@Endpoint` beans.
 
 The extended support lets Cloud Foundry management UIs (such as the web application that you can use to view deployed applications) be augmented with Spring Boot actuator information.
-For example, an application status page may include full health information instead of the typical "`running`" or "`stopped`" status.
+For example, an application status page can include full health information instead of the typical "`running`" or "`stopped`" status.
 
 NOTE: The `/cloudfoundryapplication` path is not directly accessible to regular users.
-In order to use the endpoint, a valid UAA token must be passed with the request.
+To use the endpoint, you must pass a valid UAA token with the request.
 
 
 
@@ -15,13 +15,14 @@ In order to use the endpoint, a valid UAA token must be passed with the request.
 === Disabling Extended Cloud Foundry Actuator Support
 If you want to fully disable the `/cloudfoundryapplication` endpoints, you can add the following setting to your `application.properties` file:
 
-
+====
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
 	  cloudfoundry:
 	    enabled: false
 ----
+====
 
 
 
@@ -30,25 +31,29 @@ If you want to fully disable the `/cloudfoundryapplication` endpoints, you can a
 By default, the security verification for `/cloudfoundryapplication` endpoints makes SSL calls to various Cloud Foundry services.
 If your Cloud Foundry UAA or Cloud Controller services use self-signed certificates, you need to set the following property:
 
+====
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
 	  cloudfoundry:
 	    skip-ssl-validation: true
 ----
+====
 
 
 
 [[actuator.cloud-foundry.custom-context-path]]
 === Custom Context Path
-If the server's context-path has been configured to anything other than `/`, the Cloud Foundry endpoints will not be available at the root of the application.
-For example, if `server.servlet.context-path=/app`, Cloud Foundry endpoints will be available at `/app/cloudfoundryapplication/*`.
+If the server's context-path has been configured to anything other than `/`, the Cloud Foundry endpoints are not available at the root of the application.
+For example, if `server.servlet.context-path=/app`, Cloud Foundry endpoints are available at `/app/cloudfoundryapplication/*`.
 
-If you expect the Cloud Foundry endpoints to always be available at `/cloudfoundryapplication/*`, regardless of the server's context-path, you will need to explicitly configure that in your application.
-The configuration will differ depending on the web server in use.
-For Tomcat, the following configuration can be added:
+If you expect the Cloud Foundry endpoints to always be available at `/cloudfoundryapplication/*`, regardless of the server's context-path, you need to explicitly configure that in your application.
+The configuration differs, depending on the web server in use.
+For Tomcat, you can add the following configuration:
 
+====
 [source,java,indent=0,subs="verbatim"]
 ----
 include::{docs-java}/actuator/cloudfoundry/customcontextpath/MyCloudFoundryConfiguration.java[]
 ----
+====
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
@@ -6,7 +6,7 @@ For example, the `health` endpoint provides basic application health information
 
 You can <<actuator#actuator.endpoints.enabling, enable or disable>> each individual endpoint and <<actuator#actuator.endpoints.exposing, expose them (make them remotely accessible) over HTTP or JMX>>.
 An endpoint is considered to be available when it is both enabled and exposed.
-You can auto-configure the built-in endpoints only when they are available.
+The built-in endpoints are auto-configured only when they are available.
 Most applications choose exposure over HTTP, where the ID of the endpoint and a prefix of `/actuator` is mapped to a URL.
 For example, by default, the `health` endpoint is mapped to `/actuator/health`.
 
@@ -76,7 +76,7 @@ The following technology-agnostic endpoints are available:
 
 | `sessions`
 | Allows retrieval and deletion of user sessions from a Spring Session-backed session store.
-  Requires a Servlet-based web application that uses Spring Session.
+  Requires a servlet-based web application that uses Spring Session.
 
 | `shutdown`
 | Lets the application be gracefully shutdown.
@@ -122,7 +122,6 @@ By default, all endpoints except for `shutdown` are enabled.
 To configure the enablement of an endpoint, use its `management.endpoint.<id>.enabled` property.
 The following example enables the `shutdown` endpoint:
 
-====
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
@@ -130,12 +129,10 @@ The following example enables the `shutdown` endpoint:
 	    shutdown:
 	      enabled: true
 ----
-====
 
 If you prefer endpoint enablement to be opt-in rather than opt-out, set the configprop:management.endpoints.enabled-by-default[] property to `false` and use individual endpoint `enabled` properties to opt back in.
 The following example enables the `info` endpoint and disables all other endpoints:
 
-====
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
@@ -145,7 +142,6 @@ The following example enables the `info` endpoint and disables all other endpoin
 	    info:
 	      enabled: true
 ----
-====
 
 NOTE: Disabled endpoints are removed entirely from the application context.
 If you want to change only the technologies over which an endpoint is exposed, use the <<actuator#actuator.endpoints.exposing, `include` and `exclude` properties>> instead.
@@ -288,7 +284,6 @@ You can configure both the `include` and the `exclude` properties with a list of
 
 For example, to stop exposing all endpoints over JMX and only expose the `health` and `info` endpoints, use the following property:
 
-====
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
@@ -297,12 +292,10 @@ For example, to stop exposing all endpoints over JMX and only expose the `health
 	      exposure:
 	        include: "health,info"
 ----
-====
 
 `*` can be used to select all endpoints.
 For example, to expose everything over HTTP except the `env` and `beans` endpoints, use the following properties:
 
-====
 [source,yaml,indent=0,subs="verbatim",configprops,configblocks]
 ----
 	management:
@@ -312,7 +305,6 @@ For example, to expose everything over HTTP except the `env` and `beans` endpoin
 	        include: "*"
 	        exclude: "env,beans"
 ----
-====
 
 NOTE: `*` has a special meaning in YAML, so be sure to add quotation marks if you want to include (or exclude) all endpoints.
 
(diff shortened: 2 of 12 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/cloud-foundry.adoc
```diff
@@ -30,25 +31,29 @@ If you want to fully disable the `/cloudfoundryapplication` endpoints, you can a
...
+For example, if `server.servlet.context-path=/app`, Cloud Foundry endpoints are available at `/app/cloudfoundryapplication/*`.
 
-If you expect the Cloud Foundry endpoints to always be available at `/cloudfoundryapplication/*`, regardless of the server's context-path, you will need to explicitly configure that in your application.
-The configuration will differ depending on the web server in use.
-For Tomcat, the following configuration can be added:
+If you expect the Cloud Foundry endpoints to always be available at `/cloudfoundryapplication/*`, regardless of the server's context-path, you need to explicitly configure that in your application.
+The configuration differs, depending on the web server in use.
+For Tomcat, you can add the following configuration:
 
+====
```
philwebb, line 54: What are the extra `====` lines for? I don't think we usually include them in source markup

wilkinsona:
Sorry, I typo'd the issue number in the message for https://github.com/spring-projects/spring-boot/commit/6be27e2212fc9b6d8bf3832804c21a5610782dda.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/cloud-foundry.adoc
Buzzardo: They add consistency. Some of our code listings have callouts to explain bits of code within the listing. When those are present, we want a box around both the listing and the callouts. We get that with `====` before the code listing and after the callouts. The trouble is that those code listings then look different than the other code listings, so I add the `====` around all the code listings, to make them all look the same.

wilkinsona:
@Buzzardo For future reference, we try to keep individual commits focussed on a single change. This PR's mixing improvements to the links to the other forms of the documentation with some general editing of the multi-page documentation's index. I'll split things up while merging this and have happened https://github.com/spring-projects/spring-boot/issues/27805 to track the editorial changes.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/cloud-foundry.adoc
philwebb: We've generally been avoiding callouts because they're a bit hard to manage with the extracted code that we use for all samples. The extra `====` add quite a lot of unnecessary markup, I think I'd rather we didn't add it to the Boot docs.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/jmx.adoc
```diff
@@ -1,17 +1,17 @@
 [[actuator.jmx]]
 == Monitoring and Management over JMX
 Java Management Extensions (JMX) provide a standard mechanism to monitor and manage applications.
-By default, this feature is not enabled and can be turned on by setting the configuration property configprop:spring.jmx.enabled[] to `true`.
+By default, this feature is not enabled. You can turn it on by setting the configprop:spring.jmx.enabled[] configuration property to `true`.
```
philwebb, line 4: We use [one sentence per line](https://asciidoctor.org/docs/asciidoc-recommended-practices/#one-sentence-per-line) so we'll need a CR here.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/cloud-foundry.adoc
Buzzardo: With beginning and ending tags, you can add callouts to the code samples that you include from source files. The individual callouts also need to be in a comment at the end of the line. Callouts can be a big help to readers, as the individual steps of an algorithm or other details of a block of code can be pinpointed. I'd like to see more use of callouts for that reason.

Getting your code from the actual repositories is the gold standard, and I appreciate that you do it. We're getting better across the portfolio about not using "pseudo code," but there's a ways to go yet.

philwebb: I realize we can use them if we need to, but I still think adding extra `====` blocks around source that doesn't use them add noise that we'd rather not have. I think we should only add the extra `====` blocks if we have a block that actually uses callouts. Those will need to be different subs as well since by default we use `subs="verbatim"` and for those blocks we'll need `subs="verbatim,callouts"`

Buzzardo: The trouble is that then the code blocks with callouts look different than the code blocks without callouts. It just occurred to me that the better fix for all this is to extend the box for the code to include the callouts, so there's only a single box. I bet that's tricky to do, though.

philwebb: I've opened https://github.com/spring-projects/spring-boot/issues/27806 do deal with the callout issue. It's going to be a general problem and if we break it out into its own issue we can solve it later and continue with the general edits. We don't have any callouts at the moment anyway so we don't need to solve it urgently.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -259,12 +287,14 @@ The location of the Elastic server to use can be provided using the following pr
 	      elastic:
 	        host: "https://elastic.example.com:8086"
 ----
+====
 
 [[actuator.metrics.export.ganglia]]
 ==== Ganglia
 By default, metrics are exported to {micrometer-registry-docs}/ganglia[Ganglia] running on your local machine.
-The http://ganglia.sourceforge.net[Ganglia server] host and port to use can be provided using:
+You can provide the http://ganglia.sourceforge.net[Ganglia server] host and port to use by using:
```
wilkinsona, line 295: "use by using" reads a little bit strangely to me. How about "You can configure the Ganglia server host and port to use, as the following example shows:" instead?

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -274,14 +304,15 @@ The http://ganglia.sourceforge.net[Ganglia server] host and port to use can be p
...
 ----
-
+====
 
 
 [[actuator.metrics.export.graphite]]
 ==== Graphite
 By default, metrics are exported to {micrometer-registry-docs}/graphite[Graphite] running on your local machine.
-The https://graphiteapp.org[Graphite server] host and port to use can be provided using:
+You can provide the https://graphiteapp.org[Graphite server] host and port to use by using:
```
wilkinsona, line 313: See my previous comment about "use by using"

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -115,14 +129,16 @@ The location of the https://github.com/Netflix/atlas[Atlas server] to use can be
...
+====
 
 
 
 [[actuator.metrics.export.datadog]]
 ==== Datadog
-Datadog registry pushes metrics to https://www.datadoghq.com[datadoghq] periodically.
-To export metrics to {micrometer-registry-docs}/datadog[Datadog], your API key must be provided:
+A Datadog registry periodically pushes metrics to https://www.datadoghq.com[datadoghq].
+To export metrics to {micrometer-registry-docs}/datadog[Datadog], you must provide API key:
```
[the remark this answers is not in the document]
wilkinsona: A subsequent edit uses "your".

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -429,20 +480,22 @@ By default, metrics are published via REST calls but it is also possible to use
...
 Finally, you can take full control by defining your own `NewRelicClientProvider` bean.
 
 
 
 [[actuator.metrics.export.prometheus]]
 ==== Prometheus
-{micrometer-registry-docs}/prometheus[Prometheus] expects to scrape or poll individual app instances for metrics.
-Spring Boot provides an actuator endpoint available at `/actuator/prometheus` to present a https://prometheus.io[Prometheus scrape] with the appropriate format.
+{micrometer-registry-docs}/prometheus[Prometheus] expects to scrape or poll individual application instances for metrics.
+Spring Boot provides an available actuator endpoint at `/actuator/prometheus` to present a https://prometheus.io[Prometheus scrape] with the appropriate format.
```
wilkinsona, line 492: It's not the endpoint that's available. "available" was being used to describe the location at which it can be accessed. Perhaps "Spring Boot provides an actuator endpoint, available at `/actuator/prometheus`, to present … " instead?

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -429,20 +480,22 @@ By default, metrics are published via REST calls but it is also possible to use
...
 
-TIP: The endpoint is not available by default and must be exposed, see <<actuator#actuator.endpoints.exposing,exposing endpoints>> for more details.
+TIP: By default, he endpoint is not available and must be exposed. See <<actuator#actuator.endpoints.exposing,exposing endpoints>> for more details.
```
wilkinsona, line 494: "he" -> "the"

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -611,165 +684,166 @@ You can also change the interval at which metrics are sent to Wavefront:
...
-| Response's HTTP status code (for example, `200` or `500`)
+| The response's HTTP status code (for example, `200` or `500`)
 
 | `uri`
-| Request's URI template prior to variable substitution, if possible (for example, `/api/person/\{id}`)
+| The request's URI template prior to variable substitution, if possible (for example, `/api/person/\{id}`)
 |===
 
-To add to the default tags, provide one or more ``@Bean``s that implement `WebFluxTagsContributor`.
+To add to the default tags, provide one or more `@Bean` instances that implement `WebFluxTagsContributor`.
```
wilkinsona, line 810: `@Bean` is an annotation so you can't really have instances of it. If you'd like to replace `@Bean`s, I think "beans" would be better. Either way, the next line should probably be changed too for consistency.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -914,61 +990,64 @@ To replace the default tags, provide a `@Bean` that implements `RepositoryTagsPr
...
 
 
 
 [[actuator.metrics.supported.kafka]]
 ==== Kafka Metrics
-Auto-configuration will register a `MicrometerConsumerListener` and `MicrometerProducerListener` for the auto-configured consumer factory and producer factory respectively.
-It will also register a `KafkaStreamsMicrometerListener` for `StreamsBuilderFactoryBean`.
-For more details refer to {spring-kafka-docs}#micrometer-native[Micrometer Native Metrics] section of the Spring Kafka documentation.
+Auto-configuration registers a `MicrometerConsumerListener` and `MicrometerProducerListener` for the auto-configured consumer factory and producer factory, respectively.
+It alsos register a `KafkaStreamsMicrometerListener` for `StreamsBuilderFactoryBean`.
```
wilkinsona, line 1007: "also registers"

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
```diff
@@ -914,61 +990,64 @@ To replace the default tags, provide a `@Bean` that implements `RepositoryTagsPr
...
-| Address of the server the command was sent to
+| The address of the server to which the command was sent.
 
 | `status`
-| Outcome of the command - one of (`SUCCESS`, `FAILED`)
+| The outcome of the command (`SUCCESS` or `FAILED`).
 |===
 
-To replace the default metric tags, define a `MongoCommandTagsProvider` bean, as shown in the following example:
+To replace the default metric tags, define a `MongoCommandTagsProvider` bean:
```
wilkinsona, line 1039: Has the style changed here? Previous editing added "as shown in the following example" in situations like this.

wilkinsona (changes requested):

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
```diff
@@ -459,49 +483,53 @@ Finally, if you need access to web-framework-specific functionality, you can imp
...
 ----
 	{
 		"name": "test",
 		"counter": 42
 	}
 ----
+====
 
-This can be used to invoke a write operation that takes `String name` and `int counter` parameters, as shown in the following example:
+You can use this to invoke a write operation that takes `String name` and `int counter` parameters, as the following example shows:
```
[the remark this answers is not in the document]
Buzzardo: The only rule that I try to enforce is to have some kind of lead-in to the code (or image or whatever). I've switched to this wording because it is a bit shorter, but it's OK to have both (though better to have just the one). In other words, it's not official. Rather, it's what I'm using at present (again, because it's a smidgeon shorter).

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
```diff
@@ -987,15 +1031,16 @@ readinessProbe:
...
-You can enable them in any environment using the configprop:management.endpoint.health.probes.enabled[] configuration property.
+These health groups are automatically enabled only if the application <<deployment#deployment.cloud.kubernetes,runs in a Kubernetes environment>>.
+You can enable them in any environment by using the configprop:management.endpoint.health.probes.enabled[] configuration property.
 
-NOTE: If an application takes longer to start than the configured liveness period, Kubernetes mention the `"startupProbe"` as a possible solution.
-The `"startupProbe"` is not necessarily needed here as the `"readinessProbe"` fails until all startup tasks are done, see <<actuator#actuator.endpoints.kubernetes-probes.lifecycle,how Probes behave during the application lifecycle>>.
+NOTE: If an application takes longer to start than the configured liveness period, Kubernetes mentions the `"startupProbe"` as a possible solution.
+The `"startupProbe"` is not necessarily needed here, as the `"readinessProbe"` fails until all startup tasks are done. See <<actuator#actuator.endpoints.kubernetes-probes.lifecycle,How Probes behave during the application lifecycle>>.
```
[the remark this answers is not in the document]
Buzzardo: I was only capitalizing the first word of a section title. It turns out that it is not a section title, so I have changed it back. However, I changed it to "See the section that desribes [link stuff] how to...."

Links to sections should usually be the title of the section. If not, then it needs to be clear that the section title doesn't match the words in the link, and that means adding this kind of language.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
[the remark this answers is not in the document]
Buzzardo: My bad. I created a grammatical error on that one. I need an editor, too, which is why I appreciate these reviews.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
Buzzardo: Yup. I created an awkwardness. Thanks.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
Buzzardo: Removing "available" altogether does the trick. I did that.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/metrics.adoc
Buzzardo: There's been discussion amongst the Tanzu writers about whether to have these "following" clauses. I favor them, but there's a case to be made for not using them (it's fewer words). I think about our English-as-a-second-language readers, though, who have largely been raised on a more formal kind of documentation. I'm a little conflicted about the whole issue, but I will stick to the more formal path going forward. Thanks for making me think about it again.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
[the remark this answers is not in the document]
wilkinsona: I still don't understand why "How" is capitalized here. How does doing so make it clear that the section title doesn't match the words in the link? I also don't understand why "Probes" is capitalised here but "servlet" elsewhere in the documentation is not.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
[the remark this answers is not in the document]
wilkinsona: Can we please try to avoid this sort of churn in the documentation? It makes consistency across all of the docs a moving target, and one that I suspect we'll never reach as a result.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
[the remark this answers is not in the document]
Buzzardo: My most recent commit makes it all lower case.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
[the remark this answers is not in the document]
Buzzardo: I have yet to come up with a lead-in to code that I think is really excellent. I changed it to "as the following example shows" because it's shorter and more direct (ditching the preposition). Honestly, though, if I think of or find something better in the future, I'll change it again, because it will be better.

philwebb:
Thanks @Buzzardo. I've rebased this PR and split it up into a number of smaller commits before pushing. I'm keen to try and improve the consistency across the board for the documentation so I've opened #27895 and tried to identify a bunch of changes that we can make (based on things observed in this PR). Please feel free to add any more that you think of.

I think if we can make those broad cross-cutting changes first (and one at a time) it should make the actual editing commits smaller and much easier to review.

PR #27768, opened 2021-08-19
Added details to failed SpringBootCondition

ascopes:
It would appear, at least under some logging configurations, that the
exception cause is not actually being logged in the shown stacktrace
when the failed condition exception occurs (at least, in Spring Boot 2.5.2).
This means when the condition throws an IllegalStateException, there is
no details as to what actually went wrong. This change amends the logged
message to include the cause exception name and the associated message
to assist in debugging.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/SpringBootCondition.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/SpringBootCondition.java
@@ -57,7 +57,8 @@ public abstract class SpringBootCondition implements Condition {
 					+ "put a @ComponentScan in the default package by mistake)", ex);
 		}
 		catch (RuntimeException ex) {
-			throw new IllegalStateException("Error processing condition on " + getName(metadata), ex);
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal. Unfortunately, I'm not too keen on making this change as it would be the only catch block, as far as I can recall, where we do this. I think there's a high chance that the change would be reverted in the future as it would not be clear that it's necessary. Before we can consider making this change, I think we need to understand two things:

1. In exactly which circumstances the chained exception is not sufficient
2. Why the change is only applicable to condition evaluation and not in other places where we chain exceptions.

ascopes:
The issue is more around the fact that in some circumstances, of which I am not yet able to get to the bottom of, the only exception being logged out by logback is the most recent one that is thrown. The issue is that then without the aid of a debugger, the cause of the exception is left unclear.

ascopes:
For now, looking back at this, I am not entirely certain that there is not another issue elsewhere within our codebase that is preventing the cause of the exception from being logged correctly. When I am back in the office Monday, I will have another dig through and reopen a new issue to address that if I find anything that suggests it is occurring within Spring rather than the applications themselves.

Thanks for the quick response, @wilkinsona.

PR #27811, opened 2021-08-25
Tune pre-defined logging group `sql`

quaff:
add org.hibernate.type.descriptor.sql.BasicBinder to output binding parameters


philwebb:
Thanks for the suggestion but we're not keen on having bind parameters appear in the log by default since they could contain sensitive information. We prefer the users specifically opt-in to `BasicBinder` logging.

PR #27823, opened 2021-08-27
Include WebMvcRegistrations in @WebMvcTest

quaff:
`WebMvcRegistrations` should be treat as well as `WebMvcConfigurer`


quaff:
@philwebb Would you take a look at this?

philwebb:
@quaff I think it looks sensible. @bclozel any concerns?

bclozel (approved):

Issue #18242, opened 2019-09-16
Add dependency management for Oracle JDBC driver


jsaneu:
It's available on Maven Central: [https://repo1.maven.org/maven2/com/oracle/ojdbc/](https://repo1.maven.org/maven2/com/oracle/ojdbc/)

Issue #26574, opened 2021-05-17
Drop dependency management for org.jboss:jboss-transaction-spi


wilkinsona:
None of Boot's modules depend upon `org.jboss:jboss-transaction-spi` so there's no longer any need for Boot to express an opinion about its version. We should remove the dependency management for it so that users are completely free to use whatever version meets their needs.

PR #27827, opened 2021-08-27
Drop dependency management for Oracle drivers with old groupId

gavvvr:
There is an 'ojdbc-bom' v21 already included with groupId 'com.oracle.database.jdbc' covering all jars listed explicitly for v19.3 here (excluding ojdbc10).

Check out the content of Oracle's Maven bom here:
https://repo1.maven.org/maven2/com/oracle/database/jdbc/ojdbc-bom/21.1.0.0/ojdbc-bom-21.1.0.0.pom

Also see Oracle's announcements:
https://medium.com/oracledevs/all-in-and-new-groupids-oracle-jdbc-drivers-on-maven-central-a76d545954c6
https://blogs.oracle.com/developers/post/oracle-database-client-libraries-for-java-now-on-maven-central

The `com.oracle.ojdbc` version management was originally introduced in: #18242

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1295,30 +1295,6 @@ bom {
 			]
 		}
 	}
-	library("OJDBC", "19.3.0.0") {
-		group("com.oracle.ojdbc") {
-			modules = [
-				"dms",
-				"ojdbc10",
-				"ojdbc10_g",
-				"ojdbc10dms",
-				"ojdbc10dms_g",
-				"ojdbc8",
-				"ojdbc8_g",
-				"ojdbc8dms",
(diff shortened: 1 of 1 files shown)


snicoll:
This can't really be a `task` as it is a user-facing change that should end up in the changelog.

wilkinsona:
I've used `enhancement` for this sort of issue in the past. See https://github.com/spring-projects/spring-boot/issues/26574, for example.

snicoll:
> There is an 'ojdbc-bom' v21 already included with groupId 'com.oracle.database.jdbc' covering all jars listed explicitly for v19.3 here

This is a noteworthy change as users relying on the old `groupId` won't get dependency management anymore and should upgrade to toe `com.oracle.database.jdbc` groupId.

Issue #20971, opened 2020-04-15
Support SameSite attributes on session Cookies for Servlet-based applications


bclozel:
This issue is about configuring the SameSite attribute in a consistent way for session cookies in Servlet-based applications (like Spring MVC).

Here's a link explaining the basics about SameSite Cookie attribute: https://web.dev/samesite-cookies-explained/

Google Chrome and other browsers are about to enforce new security policies for SameSite attributes in Cookies. See:
* Chromium timeline for SameSite updates: https://www.chromium.org/updates/same-site
* Initial announcement: https://blog.chromium.org/2019/10/developers-get-ready-for-new.html
* Rollout is delayed: https://blog.chromium.org/2020/04/temporarily-rolling-back-samesite.html

[As explained by Rossen here](https://github.com/spring-projects/spring-boot/issues/15047#issuecomment-613515647), the Servlet spec does not support this attribute yet, [support is scheduled for Servlet 5.1](https://github.com/eclipse-ee4j/servlet-api/issues/175).

Currently, Servlet containers are shipping native support, each in its own way.

Jetty [provides an API on its Cookie implementation directly](https://github.com/eclipse/jetty.project/issues/4247), this require to get access to Jetty's native `Request`. Developers can also set that value for Session cookies only [through a session config comment (in web.xml and probably the Servlet API](https://github.com/eclipse/jetty.project/issues/4247#issuecomment-555721734). As of eclipse/jetty.project#4512, one can also provide a default SameSite value for all Cookies.

As [explained in this comment](https://github.com/spring-projects/spring-boot/issues/15047#issuecomment-553734433), Tomcat provides a Cookie processor that changes the SameSite attribute for all outgoing Cookies.

Undertow has a [dedicated issue for comprehensive support](https://issues.redhat.com/browse/UNDERTOW-1600) (scheduled for 2.1.0 release currently). A `SameSiteCookieHandler` will allow to configure a default SameSite value for all Cookies (or a subset using a name pattern). Note that initial support was added in [UNDERTOW-1024](https://issues.redhat.com/browse/UNDERTOW-1024) directly on its Cookie implementation, but it does not support all SameSite values from the spec.

See #15047 for more information, in the context of the same improvement for Spring Session.

natevaughan:
Bump

philwebb:
@natevaughan Please don't add "bump" comments to this tracker, it causes notifications for everyone watching this repo and it won't make us get to the issue any quicker.

[...]

PR #27853, opened 2021-09-02
Support configure "domainNamePattern" and "sameSite" for spring session cookie

quaff:


philwebb:
I'm not sure that we should add additional properties here since we currently also use them in [AbstractServletWebServerFactory](https://github.com/spring-projects/spring-boot/blob/253f98c3e7ae577618ca7d16d9116205f1fb4215/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/server/AbstractServletWebServerFactory.java#L320-L330) where we can't easily add equivelent support.

We have an open issue (#20971) to deal with the `same-site` property, but we're currently blocked.

Flagging to see what the rest of the team think.

vpavic:
IMO this again really boils down to [#15047 (comment)](https://github.com/spring-projects/spring-boot/issues/15047#issuecomment-435454462).

philwebb:
Thanks @vpavic, I had a feeling there was a better issue to link to but I couldn't find it. I think we should keep the customizer approach for now.

PR #27855, opened 2021-09-02
Added goal [`fmc`], support to find main class and set to specified property

kerwin612:
* Added goal [`fmc`], support to find main class and set to specified property.

Sometimes we use spring-boot, but do not want to use repackage. We use spring-boot, but still use maven-jar-plugin for packaging, and do not want manually entering main-class. spring-boot-maven-plugin supports automatic location main-class Ability, I just open it up only.

The following is a rough example: 
```xml
<plugin>
	<configuration>
		<mainClassPropertyName>foundMainClass</mainClassPropertyName>
	</configuration>
</plugin>
<plugin>
	<artifactId>maven-jar-plugin</artifactId>
	<configuration>
		<archive>
			<manifest>
				<mainClass>${foundMainClass}</mainClass>
			</manifest>
		</archive>
	</configuration>
</plugin>
```


philwebb:
Thanks for the suggestion but Mojo doesn't feel quite right to me. It feels odd that `FindMainClassMojo` has all sorts of dependency filtering properties and also a little strange that `AbstractRunMojo` now extends a concrete mojo.

I'm not too keen to introduce a new goal just for this. I think the majority of our users are working with repackaged jars. I would suggest developing your own mojo that's specifically for this task. The `MainClassFinder` is part of `spring-boot-loader-tools` which you are free to depend on.

PR #27861, opened 2021-09-02
update jakarta.el CVE-2021-28170

strehle:
see
https://github.com/eclipse-ee4j/el-ri/issues/155


scottfrederick:
As mentioned in the pull request template: 

> Please do not open a pull request for a straightforward dependency upgrade (one that
only updates the version property). We have a semi-automated process for such upgrades
that we prefer to use. 

In the interim, you can override the Spring Boot managed version in your own build.

PR #27862, opened 2021-09-03
Add testing for AbstractFailureAnalyzer.findCause

smoothbear:
Added testing for ```findCause``` in```AbstractFailureAnalyzer```.

* ```findCauseExtendsThrowable``` and ```findCauseExtendsOtherException``` are to check if it is working properly when an exception contains generic class type.
* ```findCauseOtherException``` is to check if it returns null when an exception doesn't contain generic class type.
* ```findCauseNullObject``` is to check if it returns ```null``` when providing ```null``` as throwable parameter.

### DIFF
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/diagnostics/AbstractFailureAnalyzerTest.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/diagnostics/AbstractFailureAnalyzerTest.java
@@ -38,13 +38,24 @@ class AbstractFailureAnalyzerTest {
 	}
 
 	static class FailureAnalyzerConcrete extends AbstractFailureAnalyzer<Throwable> {
+
 		@Override
 		protected FailureAnalysis analyze(Throwable rootFailure, Throwable cause) {
 			return null;
 		}
+
 	}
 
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot/src/test/java/org/springframework/boot/diagnostics/AbstractFailureAnalyzerTest.java
```diff
@@ -0,0 +1,77 @@
...
+ */
+
+package org.springframework.boot.diagnostics;
+
+import org.junit.jupiter.api.BeforeEach;
+import org.junit.jupiter.api.Test;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+class AbstractFailureAnalyzerTest {
```
wilkinsona, line 24: This should be renamed to `…Tests` when we merge.

smoothbear: I renamed it to AbstractFailureAnalyzerTests

snicoll:
@smoothbear thank you for making your first contribution to Spring Boot.

PR #27867, opened 2021-09-03
Fix abstract class's constructor access modifier

leeseojune53:
Abstract class's constructor was public access modifier, so i changed it to protected because we must use the most restricted range available.


philwebb:
Thanks for the suggestion but we're not worried about using public constructors on abstract classes. If we did want such a change we'd ideally enforce it using a checkstyle rule.

PR #27868, opened 2021-09-03
Auto-configure JVM heap pressure metrics

vladimirfx:
Micrometer JVM heap pressure provides 2 very useful metrics that we are used to predict service outgage:
 - `jvm.memory.usage.after.gc` used for simple alert like >0.9 and trend alert that detects continius glow of non collectable heap (detects memory leaks)
 - `jvm.gc.overhead` used for simple alert (>0.3) that indicates likely incorrect heap memory settings

This PR enables these metrics by default.


snicoll (commented):
Thanks for the PR. Can you please update the test suite as well?

vladimirfx:
> Thanks for the PR. Can you please update the test suite as well?

Yes, sure. I've added a check in one unit test only. Please point me if it needed to update/add more tests.

snicoll:
@vladimirfx thank you for making your first contribution to Spring Boot. I've polished it in 39a1c85

PR #27870, opened 2021-09-04
Polish if-else to switch case

leeseojune53:
if-else statement to switch case


leeseojune53:
I don't know why the code style fails.

philwebb:
This is shaded code from another project so we don't want to make any changes to it.

PR #27871, opened 2021-09-04
feat: spring-boot version switchable without change pom.xml

qxo:
spring-boot-starter-parent should not using spring-boot-dependencies as parent.
we can using `-Dspring-boot.version=xxx` or maven profile way to change spring-boot version
without change pom.xml file


philwebb:
We prefer to have spring-boot-starter-parent inherit spring-boot-dependencies so that version properties can be used. If we use an import, it's harder to override things.

PR #27872, opened 2021-09-04
Polish Deprecated add @deprecated javadoc

leeseojune53:
Add @deprecated on @Deprecated class


philwebb:
We don't need to do this since these are package private classes that users can't interact with directly.

Issue #26729, opened 2021-06-01
Add MeterRegistry to SpringApplication and add in micrometer metrics for `StartupInfoLogger` to show overall application and jvm startup times


mbazos:
I wanted to monitor over time the full startup time of our spring boot applications. I can see that `ApplicationStartup` interface in spring includes detailed information about the start time of different steps but really what I think would be nice is an overall startup time as a metric.

I really like how `StartupInfoLogger` contains startup time for the JVM/application and having this as a metric would allow us to send it to external metric systems via micrometer and monitor over time.

After looking at the `ApplicationStartup` information I am not sure there is a good rollup to show what the `StartupInfoLogger` shows:
```
[06-01-21 16:00:13.878] [main] INFO   c.p.r.web.LocalApplicationRunner - Started LocalApplicationRunner in 16.605 seconds (JVM running for 17.318)
```

bono007:
@mbazos have you seen [ApplicationStartup](https://docs.spring.io/spring-boot/docs/2.5.0/actuator-api/htmlsingle/#startup.retrieving.snapshot) yet? I could see consuming this output from this endpoint to emit the metrics or alternatively creating a custom `ApplicationStartup` impl that emits the metrics.

mbazos:
@bono007 I originally looked at ApplicationStartup but looking at how the data is captured I think you would need to take the first event `startTime` and the last event `endTime` would that give you the full application startup time? Also I am not sure this captures JVM running time like the `StartupInfoLogger`

So I think it's a good idea maybe a custom ApplicationStartup that could emit metrics, but I was thinking more in the context of applications running containers where startup time can be critical metric.

I guess it's the difference between just having the overall aggregate time vs knowing all the individual details. I guess depending on the use-case people might want both but I think the overall aggregate time is maybe more valuable as it acts as a good starting point before someone would dig deeper to see why their application takes as long as it does.

mbazos:
@bono007 Also looking at the docs here https://docs.spring.io/spring-boot/docs/2.5.0/actuator-api/htmlsingle/#startup.retrieving.response-structure maybe we are just missing `timeline.endTime`? So then you could get the fully startup time taken and produce a metric from that?

Also I think that would cover the spring application but not the time since the JVM has been running

bono007:
> bonoo007: have you seen ApplicationStartup yet? 

@mbazos my bad, I see that you specified this in your original message. 

> maybe we are just missing timeline.endTime? 

[...]

PR #27878, opened 2021-09-05
Add startup time metrics

bono007:
Fixes gh-26729

@wilkinsona I picked your [draft changes](https://github.com/wilkinsona/spring-boot/tree/gh-26729) up and added the Actuator side of things. 

---
**TODO**
- [ ] Add docs
---

Here is what the metric looks like in Wavefront. 

Screen Shot 2021-09-05 at 9 25 44 PM


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/startup/StartupTimeMetrics.java
```diff
@@ -0,0 +1,71 @@
...
+		this.tags = (tags != null) ? tags : Collections.emptyList();
+	}
+
+	@Override
+	public void onApplicationEvent(ApplicationStartedEvent event) {
+		if (event.getStartupTime() == null) {
+			return;
+		}
+		TimeGauge
+				.builder("spring.boot.application.started", () -> event.getStartupTime().toMillis(),
```
bono007, line 57: @mbazos I know you had mentioned the following possible [metric candidates](https://github.com/spring-projects/spring-boot/issues/26729#issuecomment-878315046) for a StartupTimeMetrics.

- jvm start time
- application start time
- spring context start time


### spring context start time

Just to clarify that the `StartupInfoLogger` is actually logging ["spring context start time"](https://github.com/spring-projects/spring-boot/blob/6c48e989b6140f4cac77cfc0d1f8ab25a7f3ad65/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L306) which corresponds to the `ApplicationStartedEvent` (the app is live but not ready for requests yet - it still has to invoke any runners). This pull request logs this startup time.

### application start time 
The ["application start time"](https://github.com/spring-projects/spring-boot/blob/6c48e989b6140f4cac77cfc0d1f8ab25a7f3ad65/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java#L317) I am guessing would correspond to the `ApplicationReadyEvent` (the app is live and the app is ready for requests - all runners have been invoked). This pull request does not log this startup time. It could easily be added though and it would allow users to get an idea how much their runners are adding to startup time.

### jvm start time
The `StartupInfoLogger` is logging ["jvm uptime"](https://github.com/spring-projects/spring-boot/blob/6c48e989b6140f4cac77cfc0d1f8ab25a7f3ad65/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/StartupInfoLogger.java#L95) which corresponds to how many ms the JVM has been up. This could also give an idea of the contribution the JVM has on the startup time. This pull request does not log this startup time either but it could easily be added. 

@wilkinsona @mbazos 
Wdyt about me adding in the 2 above metrics?

bono007:
I added the startup time to the ApplicationReadyEvent as well. You can tell by the graph below which portion of the startup time is due to the runners (I think its helpful). I am waiting to push that 2nd commit though until I get a nod on adding the metric. 

Screen Shot 2021-09-06 at 5 57 21 PM

bono007:
I went ahead and pushes the subsequent commits that add the times for application.running and the jvm uptime as the point that the application is in ready state. 

Looks like this in WF...

Screen Shot 2021-09-06 at 10 26 31 PM

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
```diff
@@ -302,10 +304,11 @@ public ConfigurableApplicationContext run(String... args) {
 			refreshContext(context);
 			afterRefresh(context, applicationArguments);
 			stopWatch.stop();
+			stopWatch.start("applicationReady");
```
philwebb, line 307: I think this should be moved to past the `logStarted` call

philwebb:
Thanks very much for the PR @bono007.

We discussed some aspects of this today on our call and we think that the timer names should be changed from:

```
spring.boot.application.started
spring.boot.application.running
```

to:

```
application.started.time
application.ready.time
```

We're also wondering if there should be configuration options to change them.

Do you have any time to update the PR in that direction? If not, we can handle it when we merge.

bono007:
> Thanks very much for the PR @bono007.

My pleasure @philwebb 

> 
> We discussed some aspects of this today on our call and we think that the timer names should be changed from:
> 
> ```
> spring.boot.application.started
> spring.boot.application.running
> ```
> 
> to:
> 
> ```
> application.started.time
> application.ready.time
> ```

❤️ yeh the names were too verbose and in lockstep w/ the application event names. I went back and forth and just submitted these in hopes that I would get some feedback from team (I meant to call this out in comment). Ugghh naming... :)

One follow up question though, `spring.boot.application.running.jvm` change to default name of `jvm.read.time` or `application.ready.jvm.time`, or something else?

> We're also wondering if there should be configuration options to change them.

I have went back/forth on the ability to change metric names (in general, not just here) and one hand its great because it allows consumers to name the metrics what they want for their custom requirements. On the other hand, it allows consumers to name the metrics what they want. This can lead to a somewhat standard metric name such as "mongodb.driver.commands" now being called "acme.company.mongodb.driver.or.whatever".  I know the instrumentation in Micrometer tends to stay on the hardcoded side. I notice that SpringBoot has started to allow customization of some of the meters. I am ok w/ either approach but just wanted to bring this up. 

> Do you have any time to update the PR in that direction? If not, we can handle it when we merge.

Absolutely. On it now.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
bono007: So the `application.ready.time` is the total time from the app run()  to the end of the runners and not just the portion of time that the runners took. I will give an example to make sure we are on the same page: 

| hh:mm:ss | step |
| --- | --- |
|00:00:00|jvm is started|
|00:00:04|`run` is invoked|
|00:00:08|context has been refreshed|
|00:00:12|runners done executing|

The "base T0" time from the app perspective is `00:00:04`.  Which gives the following:

* `application.started.time`:  4s (00:00:04 -> 00:00:08)
* `application.ready.time`: 8s (00:00:04 -> 00:00:12)
* `application.ready.jvm.time`: 12s (00:00:00 -> 00:00:12)

Is that what you were thinking as well?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
```diff
@@ -59,6 +60,9 @@
 
 	private final Distribution distribution = new Distribution();
 
+	@NestedConfigurationProperty
```
bono007, line 63: I went w/ nested props merely to keep the startup details out of here. Its not being shared elsewhere - is this a proper use of NCP?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/startup/StartupTimeMetricsProperties.java
```diff
@@ -0,0 +1,66 @@
...
+ *
+ * @author Chris Bono
+ * @since 2.6.0
+ */
+public class StartupTimeMetricsProperties {
+
+	/**
+	 * Name of the metric for time taken for the application to start.
+	 */
+	private String applicationStartedTimeMetricName = "application.started.time";
```
bono007, line 30: I am really not in ❤️ w/ these uber long names (especially when you see them in use). Thoughts?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/startup/StartupTimeMetrics.java
```diff
@@ -0,0 +1,126 @@
...
+	private final String applicationStartedTimeMetricName;
+
+	private final String applicationReadyTimeMetricName;
+
+	private final String applicationReadyJvmTimeMetricName;
+
+	private final Iterable<Tag> tags;
+
+	public StartupTimeMetrics(MeterRegistry meterRegistry, RuntimeMXBean runtimeMXBean,
+			String applicationStartedTimeMetricName, String applicationReadyTimeMetricName,
```
bono007, line 56: I thought of introducing a builder, however all of these args are required (except tags) and the defaults are managed in the config properties so it would not buy us much.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/startup/StartupTimeMetrics.java
```diff
@@ -0,0 +1,126 @@
...
+		if (event instanceof ApplicationReadyEvent) {
+			onApplicationReady((ApplicationReadyEvent) event);
+		}
+	}
+
+	private void onApplicationStarted(ApplicationStartedEvent event) {
+		if (event.getStartupTime() == null) {
+			return;
+		}
+		TimeGauge
```
bono007, line 93: Wdyt about the auto-formatting on this one? It's not the easiest on the eyes.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/startup/StartupTimeMetricsProperties.java
bono007: On 2nd thought...

@philwebb wdyt about removing the metric names from config props but leaving the metric names in the `StartupTimeMetrics` which would allow configuration by consumers creating their own custom `StartupTimeMetrics` bean instead? I feel like this would probably be a rarely used config option - and one that I would feel better about not encouraging anwyays.

bono007: I ended up removing the config props. If we want to add them back in, just lmk and I will. This keeps things simpler IMO.

wilkinsona:
> One follow up question though, spring.boot.application.running.jvm change to default name of jvm.read.time or application.ready.jvm.time, or something else?

Ah, that reminds me of the other thing we discussed. We don't think this should be part of Spring Boot. Instead, it feels like a general JVM metric that could be provided my Micrometer.

bono007:
> > One follow up question though, spring.boot.application.running.jvm change to default name of jvm.read.time or application.ready.jvm.time, or something else?
> 
> Ah, that reminds me of the other thing we discussed. We don't think this should be part of Spring Boot. Instead, it feels like a general JVM metric that could be provided my Micrometer.

Yep, I agree w/ that. It was really added here as a convenience to see how much of the Spring Boot startup attributed to the overall JVM uptime (or vice-versa). The thing that Micrometer will not have is knowing that the application is ready (aka Spring event listener). 

There is already the `UptimeMetrics` which gives the JVM start time and uptime so that may be able to be used and correlated w/ these metrics. Or consumers could easily add a custom metric that listens for the app ready (like the JVM metric in this proposal). 

Either way, I will push another commit that removes the JVM metric.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/startup/StartupTimeMetrics.java
philwebb: I'm not too worried about it. We might try some "extract variable/method" refactoring when we merge.

snicoll:
Thanks again, Chris.

Issue #27826, opened 2021-08-27
Update documentation of spring.redis.jedis.pool.enabled to note that pooling is implicitly enabled in Sentinel mode


weixsun:
When using jedis and connecting redis in sentinel mode, `Jedis Sentinel cannot operate without a pool.` For details, please refer to [JedisConnectionFactory.java#L713](https://github.com/spring-projects/spring-data-redis/blob/f215f9dd31996063fce0c119f0c2b408449d23ca/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java#L713) and [JedisSentinelPool.java](https://github.com/redis/jedis/blob/master/src/main/java/redis/clients/jedis/JedisSentinelPool.java)

When using sentinel mode, spring-data-redis uses pooling by default. details is [here](https://github.com/spring-projects/spring-data-redis/blob/f215f9dd31996063fce0c119f0c2b408449d23ca/src/main/java/org/springframework/data/redis/connection/jedis/JedisConnectionFactory.java#L711-L719 )

As a result, when the user sets `spring.redis.jedis.pool.enabled=false` and uses `Jedis Sentinel Mode`operate Redis, the user will mistakenly think that spring-boot has made some efforts to make it closed under `Jedis Sentinel Mode Pooling` is implemented, but in fact it is not. The configuration in the default `JedisPoolConfig` will still be used internally.

wilkinsona:
Thanks for raising this. I think we should align with Spring Data Redis's `JedisClientConfiguration.JedisClientConfigurationBuilder.usePooling()` and update the javadoc of the `enabled` property to explain that it only applies when using single node Redis and that pooling is implicitly enabled in sentinel mode.

snicoll:
Closing in favor of PR #27891

PR #27891, opened 2021-09-08
Update documentation of spring.redis.jedis.pool.enabled to note that pooling is implicitly enabled in Sentinel mode

weixsun:
Closing #27826 

Clarify pooling is implicitly enabled in Sentinel mode.


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/RedisProperties.java
```diff
@@ -235,7 +235,8 @@ public Lettuce getLettuce() {
 
 		/**
 		 * Whether to enable the pool. Enabled automatically if "commons-pool2" is
-		 * available.
+		 * available.This setting only applies to Jedis when using non-sentinel mode and
```
mamadaliev, line 238: I would add a space before "This".

Issue #27921, opened 2021-09-09
Selenium generic autoconfiguration


cdalexndr:
Currently, only HtmlUnit driver is supported, and the following utility classes are not public so they cannot be used/extended by user with other drivers: [WebDriverScope](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/WebDriverScope.java), [WebDriverContextCustomizerFactory](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/WebDriverContextCustomizerFactory.java), [WebDriverTestExecutionListener](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/WebDriverTestExecutionListener.java)

The rhino engine used by HtmlUnit has limitations (https://github.com/HtmlUnit/htmlunit/issues/232 : https://github.com/mozilla/rhino/issues/652, https://github.com/mozilla/rhino/issues/678 ), and with the headless option already implemented by browsers,  using other drivers is a good alternative.

Autoconfigurations should be made for other drivers, maybe with the help of [WebDriveManager](https://github.com/bonigarcia/webdrivermanager/).

philwebb:
It looks like we currently apply the `WebDriverScope` to all `org.openqa.selenium.WebDriver` beans and to the `MockMvcHtmlUnitDriverBuilder` (see [this static final](https://github.com/spring-projects/spring-boot/blob/3bc5f7903b252764fb2f321082e85f662f8eec0e/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/WebDriverScope.java#L48-L49)). We also have `MockMvcWebDriverAutoConfiguration` which creates a `MockMvcHtmlUnitDriverBuilder` with a `WebConnectionHtmlUnitDriver`.

Looking at `MockMvcHtmlUnitDriverBuilder.withDelegate` it appears that we can't configure anything other than an `HtmlUnit` driver.

[...]

PR #7630, opened 2016-12-13
Add TestNG support in TestTypeExcludeFilter

eddumelendez:
Spring Framework has have support for testng for a long time. This
commit add checks for org.testng.annotations.Test annotation.

See gh-6901

PR #27947, opened 2021-09-12
Allow configuring WebDriver with TestNG

cdalexndr:
Accessible webdriver test execution listener for testng:
`@TestExecutionListeners({WebDriverTestExecutionListener.class})`

Access `WebDriverScope.NAME` outside package to be used by user webdriver implementations. Needs matching name to be picked up by autoconfiguration.
```
@Bean(name = WebDriverScope.NAME)
public WebDriver customWebDriver() {...
```

See https://github.com/spring-projects/spring-boot/issues/27921

### DIFF
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/WebDriverScope.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/servlet/WebDriverScope.java
@@ -38,6 +38,7 @@ import org.springframework.util.StringUtils;
  * @author Phillip Webb
  * @see WebDriverContextCustomizerFactory
  * @see WebDriverTestExecutionListener
+ * @since 2.6.0
  */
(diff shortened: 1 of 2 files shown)


cdalexndr:
Should I add `@since` when the feature was originally created, or current version when I made them public?

wilkinsona:
When they were made public, please.

wilkinsona:
@cdalexndr Thanks for the pull request. Unfortunately, we can't merge this until we understand why it's needed so that we can justify increasing the surface area of our public API. As [@philwebb requested](https://github.com/spring-projects/spring-boot/issues/27921#issuecomment-917536318), can you please provide a sample for #27921?

wilkinsona:
We [have a sample](https://github.com/spring-projects/spring-boot/issues/27921#issuecomment-931345119) so I don't think we're waiting for feedback.

wilkinsona:
We made changes in the past – such as [these](https://github.com/spring-projects/spring-boot/pull/7630) – in support of TestNG. 3.0 gives us an opportunity to decide if supporting TestNG is something with which we want to continue.

Issue #26657, opened 2021-05-25
Expose osbasi.SettingsCreator.createFrom


OrangeDog:
To facilitate configuration of multiple `DataSource`s, the missing piece of the convenience puzzle is `SettingsCreator.createFrom`.

I suggest adding a method to the properties, `SqlInitializationProperties.toDatabaseInitializationSettings()`, to match e.g. `Cachecontrol.toHttpCacheControl()`.

The required config to initialise a second `DataSource` would then be simply:
```java
@Bean
@ConfigurationProperties(prefix = "second.sql.init")
public SqlInitializationProperties secondInitializationProperties() {
    return new SqlInitializationProperties();
}

@Bean
public DataSourceScriptDatabaseInitializer secondDataSourceInitializer() {
    DatabaseInitializationSettings settings = secondInitializationProperties().toDatabaseInitializationSettings();
    return new DataSourceScriptDatabaseInitializer(secondDataSource(), settings);
}
```

radoslawdabrowski:
Is this issue already refined? Can I work on this?

snicoll:
Closing in favor of PR #27972

wilkinsona:
Unfortunately we missed that this is no longer necessary. The changes made in https://github.com/spring-projects/spring-boot/issues/27206 have superseded the need for this.

OrangeDog:
@wilkinsona sorry, I missed this comment before. How exactly is this no longer necessary? I still do not see any way to make a `DataSourceScriptDatabaseInitializer` from a `SqlInitializationProperties`, other than via this still package-private `createFrom` method.

wilkinsona:
There's `SqlDataSourceScriptDatabaseInitializer.SqlDataSourceScriptDatabaseInitializer(DataSource, SqlInitializationProperties)` and `SqlDataSourceScriptDatabaseInitializer.getSettings(SqlInitializationProperties)` that you can use.

Issue #27206, opened 2021-07-08
Deprecate AbstractDataSourceInitializer and its subclasses in favour of DataSourceScriptDatabaseInitializer-based equivalents


wilkinsona:
This really should have been done as part of the reworking of database initialization in 2.5 but unfortunately I missed it.

wilkinsona:
This is more complicated than I had expected.

We currently have `DataSourceInitializationConfiguration` annotated with `@ConditionalOnMissingBean(AbstractScriptDatabaseInitializer.class)`. To allow the initialisers for Batch, Integration, Quartz, and Session JDBC to use `DataSourceScriptDatabaseInitializer`-based equivalents, they would have to be auto-configured after `SqlInitializationAutoConfiguration` which imports `DataSourceInitializationConfiguration`. So far so good.

Things get complicated due to the current Batch, Integration, Quartz, and Session JDBC initializers also being `@ConditionalOnMissingBean`. If a user wants to take advantage of this and define their own initializer and that initializer is a `DataSourceScriptDatabaseInitializer`, it'll cause the main initializer to back off which we don't want.

A few ideas:

1. Drop the `@ConditionalOnMissingBean(AbstractScriptDatabaseInitializer.class)` from `DataSourceInitializationConfiguration`. It can still be disabled by setting `spring.sql.init.mode=never`. This feels rather inconsistent with the rest of Boot.
2. Don't use `DataSourceScriptDatabaseInitializer` beans to initialize the Batch, Integration, Quartz, and Session schemas. This is effectively what we have today.
3. Drop support for users defining their own Batch, Integration, Quartz, and Session JDBC initializers

At the moment, 2 feels the least bad to me. There is something of a conceptual difference with the Batch, Integration, Quartz, and Session JDBC initializers as they're supposed to be complementary to each other and to be performed in addition to the general script-based initialization.

PR #27972, opened 2021-09-13
Expose database initialization settings creator

radoslawdabrowski:
~~I've created a class to provide exposing of `SettingsCreator#createFrom`. As it is now public, name `SettingsCreator` is not very helpful, and that's why I created another class for exposing purpose (backward compatibility is kept)~~

Link: https://github.com/spring-projects/spring-boot/issues/26657

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/sql/init/SqlInitializationProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/sql/init/SqlInitializationProperties.java
@@ -21,12 +21,14 @@ import java.util.List;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.sql.init.DatabaseInitializationMode;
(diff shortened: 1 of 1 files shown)


radoslawdabrowski:
Yes, I could do that in `AbstractScriptDatabaseInitializer`.

radoslawdabrowski:
So, I am taking an advice and I implemented that in a way you suggested

snicoll:
I've noticed that `SqlDataSourceScriptDatabaseInitializer` has a public accessor via `getSettings` so I am wondering what we should do here after all.

Paging @wilkinsona

radoslawdabrowski:
Yes, I also noticed that (also there is `SqlR2dbcScriptDatabaseInitializer#getSettings`) and was wondering if this change is needed but in case mentioned by @OrangeDog it is not very useful I guess.

wilkinsona:
There are also the overloaded constructors on the `…DatabaseInitializer` classes that were introduced in https://github.com/spring-projects/spring-boot/issues/27206. I don't think we need to make any changes here now. Apologies for the wasted effort here, @radoslawdabrowski.

PR #27992, opened 2021-09-14
Add Spring Integration default poller auto-config

artembilan:
When polling consumers or source polling channel adapters
are used in Spring Integration applications, they require
some polling policy to be configured

* Add some sensitive default bean `PollerMetadata` auto-configured
bean which can be overridden by end-user or customized via
respective newly added `spring.integration.poller.*` configuration
properties

For reference a recent StackOverflow question:
https://stackoverflow.com/questions/69175808

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationAutoConfiguration.java
@@ -124,7 +124,7 @@ public class IntegrationAutoConfiguration {
 			map.from(poller::getMaxMessagesPerPoll).to(pollerMetadata::setMaxMessagesPerPoll);
 			map.from(poller::getReceiveTimeout).as(Duration::toMillis).to(pollerMetadata::setReceiveTimeout);
 			map.from(poller::getCron).whenHasText().as(CronTrigger::new).to(pollerMetadata::setTrigger);
-			map.from(poller.getFixedDelay() != null ? poller.getFixedDelay() : poller.getFixedRate())
+			map.from((poller.getFixedDelay() != null) ? poller.getFixedDelay() : poller.getFixedRate())
 					.as(Duration::toMillis).as(PeriodicTrigger::new).as((trigger) -> {
 						map.from(poller::getInitialDelay).as(Duration::toMillis).to(trigger::setInitialDelay);
 						trigger.setFixedRate(poller.getFixedRate() != null);
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationAutoConfiguration.java
@@ -16,6 +16,8 @@
 
 package org.springframework.boot.autoconfigure.integration;
 
+import java.time.Duration;
+
 import javax.management.MBeanServer;
 import javax.sql.DataSource;
 
@@ -56,10 +58,13 @@ import org.springframework.integration.rsocket.IntegrationRSocketEndpoint;
 import org.springframework.integration.rsocket.ServerRSocketConnector;
 import org.springframework.integration.rsocket.ServerRSocketMessageHandler;
 import org.springframework.integration.rsocket.outbound.RSocketOutboundGateway;
+import org.springframework.integration.scheduling.PollerMetadata;
 import org.springframework.messaging.rsocket.RSocketRequester;
 import org.springframework.messaging.rsocket.RSocketStrategies;
 import org.springframework.messaging.rsocket.annotation.support.RSocketMessageHandler;
 import org.springframework.scheduling.concurrent.ThreadPoolTaskScheduler;
+import org.springframework.scheduling.support.CronTrigger;
+import org.springframework.scheduling.support.PeriodicTrigger;
 import org.springframework.util.StringUtils;
 
 /**
@@ -109,6 +114,24 @@ public class IntegrationAutoConfiguration {
 	@Configuration(proxyBeanMethods = false)
 	@EnableIntegration
 	protected static class IntegrationConfiguration {
+
+		@Bean(PollerMetadata.DEFAULT_POLLER)
+		@ConditionalOnMissingBean(name = PollerMetadata.DEFAULT_POLLER)
+		public PollerMetadata defaultPoller(IntegrationProperties integrationProperties) {
+			IntegrationProperties.Poller poller = integrationProperties.getPoller();
+			PollerMetadata pollerMetadata = new PollerMetadata();
+			PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
+			map.from(poller::getMaxMessagesPerPoll).to(pollerMetadata::setMaxMessagesPerPoll);
+			map.from(poller::getReceiveTimeout).as(Duration::toMillis).to(pollerMetadata::setReceiveTimeout);
+			map.from(poller::getCron).whenHasText().as(CronTrigger::new).to(pollerMetadata::setTrigger);
+			map.from(poller.getFixedDelay() != null ? poller.getFixedDelay() : poller.getFixedRate())
+					.as(Duration::toMillis).as(PeriodicTrigger::new).as((trigger) -> {
+						map.from(poller::getInitialDelay).as(Duration::toMillis).to(trigger::setInitialDelay);
+						trigger.setFixedRate(poller.getFixedRate() != null);
+						return trigger;
+					}).to(pollerMetadata::setTrigger);
+			return pollerMetadata;
+		}
 
 	}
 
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
```diff
@@ -295,4 +302,86 @@ public void setMessageMappingEnabled(boolean messageMappingEnabled) {
...
+
+		/**
+		 * Maximum of messages to poll per polling cycle.
+		 */
+		private Long maxMessagesPerPoll;
+
+		/**
+		 * How long to wait for messages on poll.
+		 */
+		private Duration receiveTimeout;
```
wilkinsona, line 315: This field could have a default value that matches `PollerMetadata.DEFAULT_RECEIVE_TIMEOUT`. That would then make the default available in auto-complete when configuring the property.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
```diff
@@ -295,4 +302,86 @@ public void setMessageMappingEnabled(boolean messageMappingEnabled) {
 
 	}
 
+	public static class Poller {
+
+		/**
+		 * Maximum of messages to poll per polling cycle.
+		 */
+		private Long maxMessagesPerPoll;
```
wilkinsona, line 310: This field could have a default value that matches `PollerMetadata.MAX_MESSAGES_UNBOUNDED`. That would then make the default available in auto-complete when configuring the property.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
```diff
@@ -295,4 +302,86 @@ public void setMessageMappingEnabled(boolean messageMappingEnabled) {
...
+		 */
+		private Duration fixedRate;
+
+		/**
+		 * Polling initial delay.
+		 */
+		private Duration initialDelay;
+
+		/**
+		 * Cron expression for polling.
```
wilkinsona, line 333: It would be good to document that `cron`, `fixedRate`, and `fixedDelay` are mutually exclusive.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
```diff
@@ -295,4 +302,86 @@ public void setMessageMappingEnabled(boolean messageMappingEnabled) {
...
+		 */
+		private Duration fixedDelay;
+
+		/**
+		 * Polling rate period.
+		 */
+		private Duration fixedRate;
+
+		/**
+		 * Polling initial delay.
```
wilkinsona, line 328: It would be good to document that `cron`, `fixedRate`, and `fixedDelay` are mutually exclusive.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
```diff
@@ -295,4 +302,86 @@ public void setMessageMappingEnabled(boolean messageMappingEnabled) {
...
+		 */
+		private Duration receiveTimeout;
+
+		/**
+		 * Polling delay period.
+		 */
+		private Duration fixedDelay;
+
+		/**
+		 * Polling rate period.
```
wilkinsona, line 323: It would be good to document that `cron`, `fixedRate`, and `fixedDelay` are mutually exclusive.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationAutoConfiguration.java
```diff
@@ -110,6 +115,24 @@
...
+			PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
+			map.from(poller::getMaxMessagesPerPoll).to(pollerMetadata::setMaxMessagesPerPoll);
+			map.from(poller::getReceiveTimeout).as(Duration::toMillis).to(pollerMetadata::setReceiveTimeout);
+			map.from(poller::getCron).whenHasText().as(CronTrigger::new).to(pollerMetadata::setTrigger);
+			map.from(poller.getFixedDelay() != null ? poller.getFixedDelay() : poller.getFixedRate())
+					.as(Duration::toMillis).as(PeriodicTrigger::new).as((trigger) -> {
+						map.from(poller::getInitialDelay).as(Duration::toMillis).to(trigger::setInitialDelay);
+						trigger.setFixedRate(poller.getFixedRate() != null);
+						return trigger;
+					}).to(pollerMetadata::setTrigger);
```
wilkinsona, line 132: I wonder how we should handle the user setting more than one of `cron`, `fixedDelay`, and `fixedRate`. As things stand it looks like the order of precedence is `fixedDelay`, `fixedRate`, `cron`. Should we document this ordering, fail when more than one is configured, or something else?

wilkinsona (changes requested):
Thanks for the PR, @artembilan. I've left a few comments for your consideration when you have a moment.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
artembilan: Thank you Andy for feedback!
If I recall correctly, you or Stéphane pointed us in the past that it is better to keep configuration properties beans free from target libraries dependencies, so I decided to go with that sensitive defaults which comes with a `PollerMetadata` instance and what could be changed in the past if we would use values over here instead of those constant references. If that is still OK, I can change it as you suggest to point to those `PollerMetadata` constants. 
Or I can use their values directly for time being. That's in the case if we don't go with nulls as I have done now.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
artembilan: This is true. 
I Stream Applications projects we use `@Validated` feature to prevent ambiguity downstream and to avoid precedence as you pointed in the next comment: https://github.com/spring-projects/spring-boot/pull/27992#discussion_r709106055.
The approach is like this: https://github.com/spring-cloud/stream-applications/blob/main/functions/consumer/log-consumer/src/main/java/org/springframework/cloud/fn/consumer/log/LogConsumerProperties.java#L57.
Since I don't see anything like that in Spring Boot by itself, I forgot to ask such a question when opened PR.

Thank you for reminder!
So, now let's discuss what is the best way to go.
I simply can validate exclusiveness in target code where I apply those properties, although it is still feels more nature to validate them on the configuration properties phase.
If you point me to something similar, I'll fix it ASAP

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
wilkinsona: We typically duplicate the default without referencing anything in the library. We can then test that our default matches the library's default, as you have already done in this case.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
wilkinsona: I've labeled this for discussion at a team meeting so that we can consider our options.

wilkinsona: We're going to follow the precedent set by Framework's handling of `@Scheduled` and fail if more than one of `cron`, `fixed-rate`, and `fixed-delay` is set.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationAutoConfiguration.java
wilkinsona: We've decided to fail.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationProperties.java
wilkinsona: Doing it in the `PollerMetadata` `@Bean` method sounds good to me. Generally, we try to keep the properties classes themselves as free of logic as possible.

>Although it is not clear why you are talking about precedence

I was talking about the precedent set by Framework in how it handles similar properties on `@Scheduled`. Nothing to do with which property takes precedence over another. [That was an option](https://github.com/spring-projects/spring-boot/pull/27992#discussion_r709106055), but we've decided to fail now.

Issue #27990, opened 2021-09-14
Allow users to contribute a RedisStandaloneConfiguration bean


Peak-Song:
Hi, in my project, the redis password is gained from remote, so it cannot be configed in the `application.yaml`.
I use lettuce as the redis connector.

I have learned that in `org.springframework.boot.autoconfigure.data.redis.LettuceConnectionConfiguration` , its constructor contains `ObjectProvider<RedisSentinelConfiguration>` and `ObjectProvider<RedisClusterConfiguration>`, and more than that, in  `RedisConnectionConfiguration`, `getSentinelConfig()` and `getClusterConfiguration()` injects the corresponding bean gained from provider in priority if available, 
so in the scene of redis cluster or sentinel, I can provide my own `RedisSentinelConfiguration` or `RedisClusterConfiguration` in my own configuration class to set the password from remote.

However, in `getStandaloneConfig()`, I haven't seen the same action, then I'm pluzzled, is there an elegant way to gain  password from remote for standalone redis? Have I miss something? 

Thx.

### dependency
- springboot 2.5.4
- lettuce.core 6.1.4
- spring data redis 2.5.4

philwebb:
I think it would be nice if `RedisStandaloneConfiguration` could be provided as a bean in the same way as `RedisSentinelConfiguration` and `RedisClusterConfiguration`. We'll look into making that enhancement.

In the meantime, you might be able to use an `EnvironmentPostProcessor` to add a property source that contains the credentials that you want to use.

Peak-Song:
Ok，thx

snicoll:
Closing in favor of PR #28028

PR #28028, opened 2021-09-16
Allow users to contribute a RedisStandaloneConfiguration bean

eddumelendez:
Currently, `RedisStandaloneConfiguration` is only created via
`RedisProperties`. This change allows to use a custom `RedisStandaloneConfiguration`
bean.

See gh-27990


snicoll:
Thanks again Eddú.

PR #28060, opened 2021-09-17
Add auto-configuration for RabbitStreamTemplate

eddumelendez:
`RabbitStreamTemplate` is provided at spring-amqp 2.4.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
@@ -21,6 +21,7 @@ import java.util.stream.Collectors;
 import com.rabbitmq.client.Channel;
 import com.rabbitmq.client.impl.CredentialsProvider;
 import com.rabbitmq.client.impl.CredentialsRefreshService;
+import com.rabbitmq.stream.Environment;
 
 import org.springframework.amqp.core.AmqpAdmin;
 import org.springframework.amqp.rabbit.connection.CachingConnectionFactory;
@@ -43,6 +44,10 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
 import org.springframework.core.io.ResourceLoader;
+import org.springframework.rabbit.stream.producer.ProducerCustomizer;
+import org.springframework.rabbit.stream.producer.RabbitStreamOperations;
+import org.springframework.rabbit.stream.producer.RabbitStreamTemplate;
+import org.springframework.rabbit.stream.support.converter.StreamMessageConverter;
 
 /**
  * {@link EnableAutoConfiguration Auto-configuration} for {@link RabbitTemplate}.
@@ -83,6 +88,7 @@ import org.springframework.core.io.ResourceLoader;
  * @author Phillip Webb
  * @author Artsiom Yudovin
  * @author Chris Bono
+ * @author Eddú Meléndez
  * @since 1.0.0
  */
 @Configuration(proxyBeanMethods = false)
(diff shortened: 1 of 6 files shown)


snicoll:
paging @garyrussell for a review.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
```diff
@@ -186,4 +192,36 @@ public RabbitMessagingTemplate rabbitMessagingTemplate(RabbitTemplate rabbitTemp
 
 	}
 
+	@Configuration(proxyBeanMethods = false)
+	@ConditionalOnClass(RabbitStreamTemplate.class)
+	@ConditionalOnMissingBean(RabbitStreamTemplate.class)
+	@ConditionalOnProperty(prefix = "spring.rabbitmq.listener", name = "type", havingValue = "stream")
```
snicoll, line 198: I am not sure I got that. Isn't the listener unrelated to the creation of the template? The latter is for programmatic access, isn't it?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
```diff
@@ -186,4 +192,36 @@ public RabbitMessagingTemplate rabbitMessagingTemplate(RabbitTemplate rabbitTemp
...
+	protected static class RabbitStreamTemplateConfiguration {
+
+		@Bean
+		@ConditionalOnMissingBean(RabbitStreamOperations.class)
+		@ConditionalOnProperty(prefix = "spring.rabbitmq.stream", name = "name")
+		public RabbitStreamTemplate rabbitStreamTemplate(Environment rabbitStreamEnvironment,
+				RabbitProperties properties, ObjectProvider<MessageConverter> messageConverters,
+				ObjectProvider<StreamMessageConverter> streamMessageConverters,
+				ObjectProvider<ProducerCustomizer> producerCustomizers) {
+			RabbitStreamTemplate template = new RabbitStreamTemplate(rabbitStreamEnvironment,
```
snicoll, line 208: It sounds like it's calling for a `RabbitStreamTemplateConfigurer` to be consistent with what we're currently doing with `RabbitTemplate`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
```diff
@@ -1193,6 +1194,11 @@ private boolean determineSslEnabled(boolean sslEnabled) {
 		 */
 		private String password;
 
+		/**
+		 * Name of the stream.
+		 */
+		private String name;
```
snicoll, line 1200: I wonder if creating a `stream` namespace wouldn't be better. At the very least it should be `streamName` but perhaps we'd get more strem-related properties in the future?

snicoll (commented):
Thanks for the PR. I don't know this feature but I've added a few observations.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
garyrussell: Agreed - using a stream producer is independent of whether listeners are also stream, by default.

Also, for consistency, this should be moved to `RabbitStreamConfiguration` (which will need some rework because the whole class is excluded via this property - move the condition from the class to the  factory `@Bean`. 

The `Environment` `@Bean` is needed for both producers and consumers.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
garyrussell: Agreed.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
garyrussell: Yes, I think we should add a `stream` namespace.

garyrussell (changes requested):
Sorry for the delay - I was on PTO.

snicoll:
@eddumelendez would you have time to review the PR?

eddumelendez:
@snicoll I will work on the changes next week

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
snicoll: Sorry, I missed that reviewing the change. It's all good as it is.

snicoll:
@eddumelendez would you have time to look at the other comments in the review? I am not sure they were all addressed. If you don't, just let us know and we can take over. Thanks!

eddumelendez:
@snicoll I will be updating the PR in the next couple of days. sorry for the delay.

snicoll:
No worries, it is flagged for `2.7.x` so you have time.

Issue #28039, opened 2021-09-16
Fail or generate a warning when a property has a default value that the annotation processor cannot extract, and there's no additional metadata providing a default value


wilkinsona:
The hope is that we could avoid problems like https://github.com/spring-projects/spring-boot/issues/28037. In that case there's a `List<String>` property with a default value created with `new ArrayList<>("…")`. The annotation processor can't extract the default value so we rely on an entry in the manual metadata for it. For such a property that doesn't have an entry in the manual metadata, it'd be good if the annotation processor could fail or generate a warning.

bono007:
I too could have used this protection [here](https://github.com/spring-projects/spring-boot/pull/27660#issuecomment-920597725).

bono007:
I was sad when I dug in and discovered that the config annotation processor was not an actual magic lamp that created magic metadata for me but instead was just [some code](https://github.com/spring-projects/spring-boot/blob/30815cc0e0ace7c849fc58c61a079545c9805fd5/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/fieldvalues/javac/JavaCompilerFieldValuesParser.java#L60) written by humans that can handle only a limited set of default value types. ;)

But all kidding aside, I am interested in working this one @wilkinsona if its up for grabs. I have not yet done anything in the way of the annotation processors yet (other than figure out how to step through debugger during compile step.


If so, I have some questions before moving forward. I see the following options:

### Option 1  - fail

If there is an initializer expression but the [default turns out null](https://github.com/spring-projects/spring-boot/blob/30815cc0e0ace7c849fc58c61a079545c9805fd5/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/fieldvalues/javac/JavaCompilerFieldValuesParser.java#L168)(eg. `new ArrayList<>("…")`) then fail. This could be done directly in the `FieldValuesParser` implementation.

### Option 2 - warn 
If there is an initializer expression but the default turns out null (eg. `new ArrayList<>("…")`) then log warning. We could just catch the the exception thrown from option 1 and handle it by log warning rather than letting it bubble up.

### Option 3 - ability to configure fail or warn
Variation of the above where the "warn lever" of option 2 is used (or not) based on config prop. I am not sure if this is desirable NOR how to set configuration properties for annotation processors. 

Thoughts?

[...]

PR #28062, opened 2021-09-18
Log a fail/warn when config metadata default values are unable to be resolved

onobc:
### Example 
I created a simple app w/ the following config props:
```java
@ConfigurationProperties(prefix = "foo")
public class TestConfigProps {

    /**
     * Paths of things
     */
    private List<File> paths = new ArrayList<>(Collections.singletonList(new File(".")));

    public List<File> getPaths() {
        return paths;
    }

    public void setPaths(final List<File> paths) {
        this.paths = paths;
    }
}
```

#### Warn Mode
- I ran the config processor which output the following:

  Screen Shot 2021-09-18 at 10 42 33 AM
  
- I then added the following hint:
  ```json
  {
    "properties": [
      {
        "name": "foo.paths",
        "type": "java.util.List<java.io.File>",
        "defaultValue": [
          "."
        ]
      }
    ]
  }
  ```
- I ran the config processor again - problem fixed:

  Screen Shot 2021-09-18 at 10 43 19 AM

#### Fail Mode

When using "fail mode" its the same as above w/ the exception of the problem output looking like the following:

  Screen Shot 2021-09-18 at 10 44 21 AM

### TODOS
- [ ]  add ability to disable error mode??
- [ ]  add test for configprocessor validation checks
- [ ]  docs?

Fixes gh-28039


spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/fieldvalues/UnresolvableDefaultValue.java
```diff
@@ -0,0 +1,40 @@
...
+package org.springframework.boot.configurationprocessor.fieldvalues;
+
+/**
+ * A placeholder for a field default value used to indicate the field was unable to be
+ * resolved by the parser.
+ *
+ * @author Chris Bono
+ * @since 2.6.0
+ */
+public class UnresolvableDefaultValue {
```
bono007, line 26: I do not ❤️ using this marker interface to signal the condition but its deep enough down the chain that rippling something else to track this fact seemed less desirable than this approach. I do think we can add some default methods in `FieldResolver` to signal that a field was unresolvable and also inspect if a field was unresolvable - that would remove the config processor from having direct knowledge of this inner impl detail.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/fieldvalues/FieldValuesParser.java
```diff
@@ -29,6 +29,7 @@
  * @author Phillip Webb
  * @since 1.1.2
  * @see JavaCompilerFieldValuesParser
+ * @see UnresolvableDefaultValue
```
bono007, line 32: I only edited this to show the intention to add some default methods to set and inspect if a field was unresolvable to "hide" the `UnresolvableDefaultValue` impl detail.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/fieldvalues/javac/JavaCompilerFieldValuesParser.java
```diff
@@ -165,7 +167,11 @@ private Object getValue(VariableTree variable) throws Exception {
 			Class<?> wrapperType = WRAPPER_TYPES.get(variable.getType());
 			Object defaultValue = DEFAULT_TYPE_VALUES.get(wrapperType);
 			if (initializer != null) {
-				return getValue(initializer, defaultValue);
+			    Object initializerResolvedValue = getValue(initializer, defaultValue);
```
snicoll, line 170: It needs to be a bit more precise than that. I think the idea is that we want to fail if we have a `FieldValuesParser` available and ignore if we don't.  We we could do is changing the contract to return a `ValueWrapper` and just return that here. Our fallback implementation when we couldn't find an implementation should then return null. 

Note that we probably want to differentiate between `null` as "detected as null" and `null` as "couldn't figure out". This could be an attribute on `ValueWrapper` or something.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessor.java
```diff
@@ -353,6 +356,20 @@ private ConfigurationMetadata mergeAdditionalMetadata(ConfigurationMetadata meta
 		return metadata;
 	}
 
+	private void validateDefaultValuesResolved(ConfigurationMetadata metadata) {
+        // TODO do we want this? How do we configure annotation processor w/ 'levers'
```
snicoll, line 360: Ideally fail when we know that we're in an environment where default values can be discovered. Ignore otherwise.

snicoll (commented):
Thanks for the initial work on this. I've added a couple comments to hopefully give you directions.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/fieldvalues/javac/JavaCompilerFieldValuesParser.java
bono007: Thx for the direction @snicoll . It makes sense. I will get back to this in the next couple days.

snicoll:
@bono007 do you have an update  for us? If you don't, no worries, we can pick things up as they are as RC1 is our last chance to include this in 2.6.

bono007:
Hi @snicoll - things got busy on my end but I have carved out some time this morning to work on this. 

To help me gauge if this needs to be done in "a day" or "the next few days",  is [10/21](https://github.com/spring-projects/spring-boot/milestone/219) still accurate?

I will give an update (code or no code) by today @ "12:00:00 -5:00" (~5hrs from time of writing)

snicoll:
Thanks for the follow-up Chris. Yeah that's the RC1 release date so we still have plenty of time. Please do not rush it and let us know if you have any question.

bono007:
> Thanks for the follow-up Chris. Yeah that's the RC1 release date so we still have plenty of time. Please do not rush it and let us know if you have any question.

@snicoll my plans to carve out time this morning were foiled by a Kafka Streams production outage that I am putting out. It sounds though it is ok to have something in the next 1-2 days is ok though (correct me if that is not the case). I should have time later today to get back to it. 

Thanks

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/fieldvalues/javac/JavaCompilerFieldValuesParser.java
bono007: @snicoll 

> It needs to be a bit more precise than that. I think the idea is that we want to fail if we have a FieldValuesParser available and ignore if we don't. 

So, when no FVP is available (ie the env is not an `com.sun.tools.javac.processing.JavacProcessingEnvironment`) the [default FVP is NONE](https://github.com/spring-projects/spring-boot/blob/29637706a9978bd40b7a5593885ce279e168f17f/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/MetadataGenerationEnvironment.java#L125) which simply returns an empty map. Because this code is in the FVP it will only apply when the FVP (itself) is available. 

> We we could do is changing the contract to return a ValueWrapper and just return that here. Our fallback implementation when we couldn't find an implementation should then return null.

Again, the fallback impl currently is an empty map. 

I may be missing something here - please clarify.

snicoll: > Again, the fallback impl currently is an empty map.

I am aware of that and I am asking you to change the contract. You can't know what fields are available if you don't have an implementation, but you could change the contract so that, if you're being requested a field with name "foo" you can then return `null` or a `ValueWrapper` that holds additional information as I've suggested. So instead of a map a POJO that can tell you if the result contains a field with a given name and another method that gives you a `ValueWrapper` based on a field name.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessor.java
```diff
@@ -353,6 +356,22 @@ private ConfigurationMetadata mergeAdditionalMetadata(ConfigurationMetadata meta
 		return metadata;
 	}
 
+	private void validateDefaultValuesResolved(ConfigurationMetadata metadata) {
+        metadata.getItems().forEach((itemMetadata -> {
+			if (itemMetadata.getDefaultValue() instanceof ValueWrapper) {
```
bono007, line 361: I have not propagated the change from `Object -> ValueWrapper` up from FVP to `ItemMetadata` yet but will do that shortly. Unless you can think of a reason to keep it `Object`.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/fieldvalues/javac/JavaCompilerFieldValuesParser.java
bono007: I wanted to submit what I have so far before calling it a day as I know you 7hr ahead of me @snicoll . 

I have not moved from map to POJO and changed the contract yet as the map (w/ its ValueWrapper values) pretty much fulfills the above mentioned contract. I don't mind moving to POJO but am curious if you still think it will add value to the current proposal.

I am now marching through the JCFPV tests and updating it. I will continue that tomorrow.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessor.java
bono007: One reason to keep it Object is to prevent the ripple around all areas that are accessing ItemMetadata#defaultValue (JsonConverter, MetadataStore, etc..).

snicoll: Yeah. I didn't even think of changing `ItemMetadata` to be honest. I was hoping we could trigger the warning/error locally to field resolution by passing the necessary bits to it.

bono007: Unfortunately we have to wait until the additional metadata is merged to see if the user added a manual hint.

bono007:
@snicoll I was looking through the build failures and our update to the processor is suggesting that we have quite a few "hints" to add ourselves...

```java
> Task :spring-boot-project:spring-boot-autoconfigure:compileJava FAILED
error: Could not resolve default value for property 'server.tomcat.additional-tld-skip-patterns' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'server.tomcat.relaxed-path-chars' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'server.tomcat.relaxed-query-chars' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'server.undertow.accesslog.dir' using expression 'new File("logs")'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'server.undertow.max-headers' using expression 'UndertowOptions.DEFAULT_MAX_HEADERS'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'server.undertow.max-parameters' using expression 'UndertowOptions.DEFAULT_MAX_PARAMETERS'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'server.undertow.options.server' using expression 'new LinkedHashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'server.undertow.options.socket' using expression 'new LinkedHashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.activemq.packages.trusted' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.artemis.embedded.cluster-password' using expression 'UUID.randomUUID().toString()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.artemis.embedded.server-id' using expression 'serverIdCounter.getAndIncrement()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.cache.cache-names' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.data.elasticsearch.client.reactive.endpoints' using expression 'new ArrayList<>(Collections.singletonList("localhost:9200"))'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.datasource.xa.properties' using expression 'new LinkedHashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.flyway.init-sqls' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.flyway.jdbc-properties' using expression 'new HashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.flyway.placeholders' using expression 'new HashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.flyway.schemas' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.freemarker.settings' using expression 'new HashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.integration.channel.max-broadcast-subscribers' using expression 'Integer.MAX_VALUE'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.integration.channel.max-unicast-subscribers' using expression 'Integer.MAX_VALUE'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.integration.endpoint.no-auto-startup' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.integration.endpoint.read-only-headers' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.jersey.init' using expression 'new HashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.jpa.properties' using expression 'new HashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.kafka.bootstrap-servers' using expression 'new ArrayList<>(Collections.singletonList("localhost:9092"))'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.kafka.consumer.key-deserializer' using expression 'StringDeserializer.class'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.kafka.consumer.value-deserializer' using expression 'StringDeserializer.class'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.kafka.producer.key-serializer' using expression 'StringSerializer.class'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.kafka.producer.value-serializer' using expression 'StringSerializer.class'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.ldap.embedded.base-dn' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.mail.properties' using expression 'new HashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.mvc.contentnegotiation.media-types' using expression 'new LinkedHashMap<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.rabbitmq.stream.port' using expression 'DEFAULT_STREAM_PORT'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.security.user.password' using expression 'UUID.randomUUID().toString()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.security.user.roles' using expression 'new ArrayList<>()'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.session.servlet.filter-order' using expression 'SessionRepositoryFilter.DEFAULT_ORDER'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.task.execution.pool.max-size' using expression 'Integer.MAX_VALUE'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.task.execution.pool.queue-capacity' using expression 'Integer.MAX_VALUE'. Add hint in additional-spring-configuration.metadata.json
error: Could not resolve default value for property 'spring.webservices.servlet.init' using expression 'new HashMap<>()'. Add hint in additional-spring-configuration.metadata.json
40 errors
````
I will start marching through the above list and will add these to the additional metadata.

bono007:
When I encountered the above list I was a bit surprised - I did not expect to see so many entries in violation. I am thinking we should leave the default behavior as `Kind.ERROR` but provide the ability to switch to `Kind.WARN`.  In most cases I bet the list of updates needed is small but I am sure there are some cases where this will be a pain to absorb at once. Thoughts?

snicoll:
Thanks for the follow-up.

> I did not expect to see so many entries in violation.

There are a lot of false positive in this list. Creating an empty list or map shouldn't lead to a violation. `UUID.randomUUID().toString()` on the default security password is interesting. We'd probably need a way to escape that check.

bono007:
> Thanks for the follow-up.
Np.

> 
> There are a lot of false positive in this list. Creating an empty list or map shouldn't lead to a violation. `UUID.randomUUID().toString()` on the default security password is interesting. We'd probably need a way to escape that check.

I will add these exemptions and then add "hints" for the remaining entries.

wilkinsona:
Thanks, @onobc. Unfortunately, having seen what it takes to implement it we've realised that my initial idea wasn't a particularly good one. I'm afraid we're going to decline this and try some alternatives idea. Sorry for the wasted effort. We'll repurpose #28039 to do that.

onobc:
@wilkinsona no worries - we try some approaches and some work and some don't. :)

Issue #27895, opened 2021-09-09
Edit documentation for consistency


philwebb:
With #27759 @Buzzardo applied an editing pass to the actuator documentation. There are quite a few patterns that emerged from that set that we should try and adopt going forward. We can update this epic with items as we identify them and create distinct issues/commits for each one when we apply it.

- [ ] Unless a title or start of a sentence "Servlet" should be written "servlet".
- [ ] We should avoid "via" and use "over" or "though" instead.
- [ ] We should replace "refer" and "check out" with "see".
- [ ] "don't" should be expanded to "do not".
- [ ] "it's" should be expanded to "it is".
- [ ] "will enable" can be replaced with "enables".
- [ ] And `See "<<...` references should use lowercase contents.
- [ ] Table with descriptions should start with "The...". For example, "The color of the sky" rather than "Color of the sky".
- [ ] We should when possible change "deny" to "prevent".
- [ ] We need to check if we said "less" when we meant "fewer".
- [ ] We should expand "e.g." to "for example".
- [ ] We should expand "i.e." to "that is".
- [ ] Use a consistent quote style of ``"`thing`"``
- [ ] Use "as the following example shows" for example lead-in

Buzzardo:
In a couple of cases, you have identified specific examples of wider rules. Expanding "don't" to "do not" and "it's to "it is" are instances of the broader rule to not use contractions. That is, an apostrophe should always indicate a possessive. (That one comes from formal academic writing, which has had a large influence on technical writing over the years.)

"e.g." to "for example" and "i.e." to "that is" and "via" to "through" are all examples of avoiding Latin. More than half the population confuse "e.g." and "i.e.". Similarly, most people couldn't tell you what "via" actually means. (It's Latin for "through." I studied Latin in college.) Also, screen readers sometimes do awful things to them, most notably rendering "e.g." as "egg."

"Will enable" is an instance of the broader rule to never use "will" or otherwise use the future tense. The better reason to do that is consistency. Keeping everything in the simple present tense is more consistent. Doing so often makes sentences shorter, too. The worse reason is one of those funny-sad stories. When I worked at GE, we got sued because we had used "will" in a document. Some twit sued us because we made a promise we didn't keep. The case got tossed out, but GE had to pay lawyers to deal with it.

[...]

PR #28064, opened 2021-09-19
Remove english contractions from the documentation

ghostd:
(won't -> will not, it's -> it is, and so on)

This patch is a step forward for the issue #27895


mamadaliev (approved):

snicoll:
@ghostd thank you for making your first contribution to Spring Boot.

Issue #27437, opened 2021-07-21
AspectJ + devtools restart: java.lang.LinkageError attempted duplicate class definition


cdalexndr:
Using Spring Boot 2.4.4 with AspectJ 1.9.6 runtime weaving.

During triggered devtools restart, I get a bunch of errors like:
```
java.lang.reflect.InvocationTargetException
	at java.base/jdk.internal.reflect.GeneratedMethodAccessor13.invoke(Unknown Source)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:566)
	at org.aspectj.weaver.loadtime.ClassLoaderWeavingAdaptor.defineClass(ClassLoaderWeavingAdaptor.java:1103)
	at org.aspectj.weaver.loadtime.ClassLoaderWeavingAdaptor.access$300(ClassLoaderWeavingAdaptor.java:66)
	at org.aspectj.weaver.loadtime.ClassLoaderWeavingAdaptor$SimpleGeneratedClassHandler.acceptClass(ClassLoaderWeavingAdaptor.java:150)
	at org.aspectj.weaver.tools.WeavingAdaptor$WeavingClassFileProvider$1.acceptResult(WeavingAdaptor.java:920)
	at org.aspectj.weaver.bcel.BcelWeaver.weaveAndNotify(BcelWeaver.java:1434)
	at org.aspectj.weaver.bcel.BcelWeaver.weave(BcelWeaver.java:1195)
	at org.aspectj.weaver.tools.WeavingAdaptor.getWovenBytes(WeavingAdaptor.java:551)
	at org.aspectj.weaver.tools.WeavingAdaptor.weaveClass(WeavingAdaptor.java:387)
	at org.aspectj.weaver.loadtime.Aj.preProcess(Aj.java:116)
	at org.aspectj.weaver.loadtime.ClassPreProcessorAgentAdapter.transform(ClassPreProcessorAgentAdapter.java:51)
	at java.instrument/java.lang.instrument.ClassFileTransformer.transform(ClassFileTransformer.java:246)
	at java.instrument/sun.instrument.TransformerManager.transform(TransformerManager.java:188)
	at java.instrument/sun.instrument.InstrumentationImpl.transform(InstrumentationImpl.java:563)
[...]

PR #28071, opened 2021-09-20
Document that devtools restart doesn't work when using AspectJ weaving

cdalexndr:
https://github.com/spring-projects/spring-boot/issues/27437


philwebb:
Thanks very much @cdalexndr. I've merged this into 2.4.x, 2.5.x and main.

Issue #28121, opened 2021-09-23
Provide a specific exception and failure analysis for situations where more than one mutually exclusive configuration property has been configured


wilkinsona:
It'd be useful for https://github.com/spring-projects/spring-boot/pull/27992.

wilkinsona:
Here's an example of the failure analysis:

```
***************************
APPLICATION FAILED TO START
***************************

Description:

The following configuration properties are mutually exclusive:

	spring.integration.poller.cron
	spring.integration.poller.fixed-delay
	spring.integration.poller.fixed-rate

However, more than one of those properties has been configured at the same time:

	spring.integration.poller.cron (originating from 'class path resource [application.properties] - 6:34')
	spring.integration.poller.fixed-rate (originating from 'class path resource [application.properties] - 5:40')


Action:

Update your configuration so that only one of the mutually exclusive properties is configured.
```

PR #28123, opened 2021-09-23
Support PEM format for Kafka SSL certificates and private key (KIP-651)

timtebeek:
Before [KIP-651](https://cwiki.apache.org/confluence/display/KAFKA/KIP-651+-+Support+PEM+format+for+SSL+certificates+and+private+key) Kafka only supported file-based key and trust stores for SSL.
After KIP-651 it's now also possible to use PEM, which can be passed as environment variables rather than files.
This frees us from having to write files to the Docker images that run our application, and enables us to use build pack images.

Not sure if I've covered all corners in this initial implementation; please let me know if you would like me to make any changes. Tests had not covered Kafka SSL configuration before, and since we're merely passing values I'd not thought to add those now.
Had also thought these changes are not notable enough to include my name as author at the top of the file, but not sure if it would be breaking convention to leave it out.

One thing to note is that the referenced properties were only added in Spring Kafka 2.7, so any users that both downgrade to Kafka 2.6 _and_ use SSL might now pass unknown properties to Kafka, leading to a warning. (Provided the org.apache.kafka.common.config.SslConfigs constants are inlined by the compiler.)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
@@ -1047,6 +1047,16 @@ public class KafkaProperties {
 		private String keyPassword;
 
 		/**
+		 * Certificate chain in the format specified by 'ssl.keystore.type'.
+		 */
+		private String keyStoreCertificateChain;
+
+		/**
+		 * Private key in the format specified by 'ssl.keystore.type'.
+		 */
+		private String keyStoreKey;
+
+		/**
 		 * Location of the key store file.
 		 */
 		private Resource keyStoreLocation;
@@ -1060,6 +1070,11 @@ public class KafkaProperties {
 		 * Type of the key store.
 		 */
 		private String keyStoreType;
+
+		/**
+		 * Trusted certificates in the format specified by 'ssl.truststore.type'.
+		 */
+		private String trustStoreCertificates;
 
 		/**
 		 * Location of the trust store file.
@@ -1089,6 +1104,22 @@ public class KafkaProperties {
 			this.keyPassword = keyPassword;
 		}
 
+		public String getKeyStoreCertificateChain() {
+			return keyStoreCertificateChain;
+		}
+
+		public void setKeyStoreCertificateChain(String keyStoreCertificateChain) {
+			this.keyStoreCertificateChain = keyStoreCertificateChain;
+		}
+
+		public String getKeyStoreKey() {
+			return keyStoreKey;
+		}
+
+		public void setKeyStoreKey(String keyStoreKey) {
+			this.keyStoreKey = keyStoreKey;
+		}
+
 		public Resource getKeyStoreLocation() {
 			return this.keyStoreLocation;
 		}
@@ -1111,6 +1142,14 @@ public class KafkaProperties {
 
 		public void setKeyStoreType(String keyStoreType) {
 			this.keyStoreType = keyStoreType;
+		}
+
+		public String getTrustStoreCertificates() {
+			return trustStoreCertificates;
+		}
+
+		public void setTrustStoreCertificates(String trustStoreCertificates) {
+			this.trustStoreCertificates = trustStoreCertificates;
 		}
 
 		public Resource getTrustStoreLocation() {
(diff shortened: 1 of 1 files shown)


timtebeek:
@garyrussell would you mind taking a look at this change? Thought it might help people switch from JKS to PEM, with the latter passed in through environment variables rather than files.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -1046,6 +1046,16 @@ public void setMissingTopicsFatal(boolean missingTopicsFatal) {
 		 */
 		private String keyPassword;
 
+		/**
+		 * Certificate chain in the format specified by 'ssl.keystore.type'.
```
garyrussell, line 1050: My understanding from the KIP and Kafka docs is that only PEM is supported; I wonder if we should indicate that here and/or enforce it in `buildProperties()`?

OTOH, leaving it like this will make it future proof.

garyrussell (commented):
LGTM - just a comment for the wider Boot team's opinion.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
timtebeek: I had copied [the comments from the SslConfigs source](https://github.com/apache/kafka/blob/e8754ba7a01f33ba503ef79de66a5f644cc2ced8/clients/src/main/java/org/apache/kafka/common/config/SslConfigs.java#L76).
I had left out the lines immediately following it that detail (current) support for formats as a way of future proofing.

I'm fine with any which way: expanding the comments here, refer to SslConfigs in JavaDoc, or leaving it as is.
This comment is mostly to provide context; up to you how to best document this going forward.

wilkinsona: We could also do with some tests for the new support. That would help us to make an informed decision around what to do when the type is and is not set and when it's set to a contradictory value.

wilkinsona (changes requested):
Thanks for the PR, @timtebeek. I've left a few comments for your consideration when you have a minute.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -1046,6 +1046,16 @@ public void setMissingTopicsFatal(boolean missingTopicsFatal) {
 		 */
```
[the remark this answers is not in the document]
timtebeek: Hopefully updated as you had intended; if not feel free to suggest alternative wording which I'll gladly apply.

timtebeek:
> We could also do with some tests for the new support. That would help us to make an informed decision around what to do when the type is and is not set and when it's set to a contradictory value. -- @wilkinsona 

I'm struggling a bit with what type of test to add, as there's no conditionals yet in for instance `KafkaProperties.Ssl.buildProperties()`.
You're right to point out that the new properties can only be used when a corresponding value of `spring.kafka.ssl.key-store-type=PEM` and/or `spring.kafka.ssl.trust-store-type=PEM` is provided, as the default is `JKS`.
For now all properties are passed as is to the relevant KafkaAdmin / KafkaConsumerFactory / KafkaProducerFactory / StreamsBuilderFactoryBean, without additional validation of what combination of properties is set.

So these properties would work
```properties
spring.kafka.ssl.key-store-type=PEM
spring.kafka.ssl.key-store-key=-----BEGIN
spring.kafka.ssl.key-store-certificate-chain=-----BEGIN

spring.kafka.ssl.trust-store-type=PEM
spring.kafka.ssl.trust-store-certificates=-----BEGIN
```

As would these properties, as taken from existing: `@Test KafkaAutoConfigurationTests.consumerProperties()`
```properties
spring.kafka.ssl.key-store-location=classpath:ksLoc
spring.kafka.ssl.key-store-password=p2
spring.kafka.ssl.key-store-type=PKCS12 // Overrides Kafka default of JKS as per SslConfigs

spring.kafka.ssl.trust-store-location=classpath:tsLoc
spring.kafka.ssl.trust-store-password=p3
spring.kafka.ssl.trust-store-type=PKCS12 // Overrides Kafka default of JKS as per SslConfigs
```

But at present invalid combinations of store-type and properties could also be passed, with interpretation and validation up to Apache Kafka itself.

Would you like me to add conditionals that guarantee that the new properties are only passed with the proper corresponding key type?
And if so, would `KafkaProperties.Ssl.buildProperties()` be the best place for that?

Otherwise it becomes hard to write tests that will actually break when envisioned/needed.

timtebeek:
After thinking it over some more I'm still somewhat hesitant in adding validation that mimics the validation done by Kafka; since we'd always be trying to keep up with any validation changes there.

Perhaps we could as a minimum require that key-store-type and trust-store-type are non-null when the corresponding PEM properties are passed? That should decouple us from hardcoding Kafka knowledge, allow for future changes, while still preventing potentially common misconfiguration.

wilkinsona:
I was more thinking that writing some additional tests that use the new properties would help us to learn about how Kafka behaves. From what you've described above, @timtebeek, it sounds like you already knew or have already learned that.

We [recently added support for reporting an error when configuration properties that are mutually exclusive are used together](https://github.com/spring-projects/spring-boot/issues/28121). I wonder if that may be useful here as I suspect it doesn't make sense to configure `key-store-location` and `key-store-certificate-chain` at the same time, particularly if Kafka would just ignore one in favour of the other.

I'll flag this for team attention to see what everyone else thinks. I suspect we may have to do some experimentation to get a feel for how things work and what the best thing to do will be.

PR #28138, opened 2021-09-27
Support both kebab-case and camelCase as Spring init CLI Options

vignesh1992:
Spring init CLI to support both kebab-case and camelCase as project generation options. This PR addresses the issue #26878

### DIFF
--- a/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
+++ b/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
@@ -128,16 +128,20 @@ public class InitCommand extends OptionParsingCommand {
 			otherOptions();
 		}
 
+		/**
+		 * Supports both kebab-case and camelCase as project CLI Options camelCase to be
+		 * deprecated in future releases
+		 */
 		private void projectGenerationOptions() {
-			this.groupId = option(Arrays.asList("groupId", "g"), "Project coordinates (for example 'org.test')")
-					.withRequiredArg();
-			this.artifactId = option(Arrays.asList("artifactId", "a"),
+			this.groupId = option(Arrays.asList("groupId", "group-id", "g"),
+					"Project coordinates (for example 'org.test')").withRequiredArg();
+			this.artifactId = option(Arrays.asList("artifactId", "artifact-id", "a"),
 					"Project coordinates; infer archive name (for example 'test')").withRequiredArg();
 			this.version = option(Arrays.asList("version", "v"), "Project version (for example '0.0.1-SNAPSHOT')")
 					.withRequiredArg();
 			this.name = option(Arrays.asList("name", "n"), "Project name; infer application name").withRequiredArg();
 			this.description = option("description", "Project description").withRequiredArg();
-			this.packageName = option("package-name", "Package name").withRequiredArg();
+			this.packageName = option(Arrays.asList("packageName", "package-name"), "Package name").withRequiredArg();
 			this.type = option(Arrays.asList("type", "t"),
 					"Project type. Not normally needed if you use --build "
 							+ "and/or --format. Check the capabilities of the service (--list) for more details")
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
```diff
@@ -128,16 +129,20 @@ protected void options() {
...
+			this.groupId = option(Arrays.asList("groupId", "group-id", "g"),
+					"Project coordinates (for example 'org.test')").withRequiredArg();
+			this.artifactId = option(Arrays.asList("artifactId", "artifact-id", "a"),
 					"Project coordinates; infer archive name (for example 'test')").withRequiredArg();
 			this.version = option(Arrays.asList("version", "v"), "Project version (for example '0.0.1-SNAPSHOT')")
 					.withRequiredArg();
 			this.name = option(Arrays.asList("name", "n"), "Project name; infer application name").withRequiredArg();
 			this.description = option("description", "Project description").withRequiredArg();
-			this.packageName = option("package-name", "Package name").withRequiredArg();
+			this.packageName = option(Arrays.asList("packageName", "package-name"), "Package name").withRequiredArg();
```
snicoll, line 145: The plan is to use kebab case consistently, not introducing a mixed case for new properties.

spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
```diff
@@ -148,11 +153,11 @@ private void projectGenerationOptions() {
 					.defaultsTo("maven");
 			this.format = option("format", "Format of the generated content (for example 'build' for a build file, "
 					+ "'project' for a project archive)").withRequiredArg().defaultsTo("project");
-			this.javaVersion = option(Arrays.asList("java-version", "j"), "Language level (for example '1.8')")
-					.withRequiredArg();
+			this.javaVersion = option(Arrays.asList("javaVersion", "java-version", "j"),
```
snicoll, line 156: The plan is to use kebab case consistently, not introducing a mixed case for new properties.

spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
```diff
@@ -148,11 +153,11 @@ private void projectGenerationOptions() {
...
+					"Language level (for example '1.8')").withRequiredArg();
 			this.language = option(Arrays.asList("language", "l"), "Programming language  (for example 'java')")
 					.withRequiredArg();
-			this.bootVersion = option(Arrays.asList("boot-version", "b"),
+			this.bootVersion = option(Arrays.asList("bootVersion", "boot-version", "b"),
```
snicoll, line 160: The plan is to use kebab case consistently, not introducing a mixed case for new properties.

spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
```diff
@@ -128,16 +129,20 @@ protected void options() {
...
 
+		/**
+		 * Supports both kebab-case and camelCase as project CLI Options. camelCase to be
+		 * deprecated as part of future releases
+		 */
 		private void projectGenerationOptions() {
-			this.groupId = option(Arrays.asList("groupId", "g"), "Project coordinates (for example 'org.test')")
-					.withRequiredArg();
-			this.artifactId = option(Arrays.asList("artifactId", "a"),
+			this.groupId = option(Arrays.asList("groupId", "group-id", "g"),
```
snicoll, line 137: `groupId` is not deprecated.

spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
```diff
@@ -128,16 +129,20 @@ protected void options() {
...
+					"Project coordinates (for example 'org.test')").withRequiredArg();
+			this.artifactId = option(Arrays.asList("artifactId", "artifact-id", "a"),
```
snicoll, line 139: `artifactId` is not deprecated.

snicoll (changes requested):
Thanks for the PR but this adresses only a small part of the original issue. In particular, the issue was about showing consistent options in the CLI:

> The code has a mapping from CLI options (kebab-case) to query parameters (camelCase) when issuing a project creation request. I think it needs to map things the other way when displaying information for init --help

The other part was to use kebab case consistently:

> We should probably also fix the CLI options so that all multi-word options have a kebab-case variant. Specifically --artifactId and --groupId should be joined by --artifact-id and --group-id respectively. The camelCase variants of the CLI options should be deprecated.

If you intend to only provide a fix for the second part, please let us know and we can then reopen the other issue.

spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/command/init/InitCommand.java
wilkinsona: I thought we were going to support both camelCase and kebab-case until we've improved the output of `spring init --list` so that it converts the server's output from the camelCase that it expects to the kebab-case that the CLI prefers. That improvement to `spring init --list` was to be handled separately.

snicoll: Ah ok, right. I understand now. Yes, so we should have to consistent indeed but we need to deprecate them.

snicoll:
@vignesh1992 please ignore my comment about the table. I forgot this is something that should be addressed elsewhere.

VigneshThangavelIlangovan:
Hi @snicoll, Thank you so much for your feedback! Should there be any change which I have to perform or is this PR good to go?

PR #28141, opened 2021-09-27
Replace latin locutions in the documentation

ghostd:
Replace "i.e." by "that is", and "e.g." by "for example".

This patch is a step forward for the issue #27895


ghostd:
I did not update the Javadoc; do we want to update it?

PR #28162, opened 2021-10-01
Avoid explicit initialization of Atomics with their default values

dreis2211:
Hi,

this PR changes calls like `new AtomicBoolean(false)` to `new AtomicBoolean()` as they cause a volatile write which can be saved in cases where the constructor parameter is the default value.

Cheers,
Christoph


spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/metrics/web/client/MetricsRestTemplateCustomizerTests.java
```diff
@@ -157,7 +157,7 @@ void whenCustomizerAndLocalHostUriTemplateHandlerAreUsedTogetherThenRestTemplate
 
 	@Test
 	void whenAutoTimingIsDisabledUriTemplateHandlerDoesNotCaptureUris() {
-		AtomicBoolean enabled = new AtomicBoolean(false);
+		AtomicBoolean enabled = new AtomicBoolean();
```
analog-alex, line 160: Just thinking out loud, but I find it more expressive in the tests to include the explicit `false`, or at least a comment indicating that the default value is false.  
It indicates intention.

Issue #28122, opened 2021-09-23
Defer OIDC lookup until first use of the issuer location JwtDecoder


jzheaux:
Spring Security 5.6 is shipping with [a new `JwtDecoder`](https://github.com/spring-projects/spring-security/pull/10310) that defers the OIDC discovery lookups that normally happen during startup.

Users will be able to do, for example:

```java
@Bean 
JwtDecoder jwtDecoder() {
    Supplier<JwtDecoder> jwtDecoder = () -> JwtDecoders.fromIssuerLocation("https://issuer/endpoint");
    return new SupplierJwtDecoder(jwtDecoder);
}
```

And the startup configuration won't be invoked until the app first calls `JwtDecoder#decode`.

Today, Spring Boot does something like the following when only an `issuer-uri` is provided:

```java 
@Bean 
JwtDecoder jwtDecoder() {
    return JwtDecoders.fromIssuerLocation("https://issuer/endpoint");
}
```

It would be nice if Spring Boot published the `JwtDecoder` as a `SupplierJwtDecoder` to provide a quicker and more resilient startup experience.

If there is a need for users to restore the previous eager-loading behavior, they can publish the bean themselves; however, a property may also be worth considering.

mbhave:
Closing in favor of PR #28169.

PR #28169, opened 2021-10-01
#28122 - Defer OIDC lookup until first use of the issuer location JwtDecoder

Pooja199:


mbhave:
Spring Security made [a change](https://github.com/spring-projects/spring-security/blob/9b24f66f1cf8a1a53a0a7530109e2185aef4656b/oauth2/oauth2-jose/src/main/java/org/springframework/security/oauth2/jwt/ReactiveJwtDecoders.java#L111) that defers the call to get the signature algorithm once `decode` is actually called. This is causing the tests in `ReactiveOAuth2ResourceServerAutoConfigurationTests.java` to fail because the number of requests to the mock server is different. We will need to update the assertions as part of merging this PR.

mbhave:
@Pooja199 Thank you for making your first contribution to Spring Boot! This has now been merged into `main`.

PR #28170, opened 2021-10-01
Update docs to be explicit about dot notation being correctly mapped

dvonsegg:
Update to Docs. There was confusion on if dot notation was correct. Updated docs to explicitly state dot notation is allowed.


mbhave:
@dvonsegg Thank you for making your first pull request to Spring Boot! This has now been merged into 2.4.x, 2.5.x and main along with this [polish commit](https://github.com/spring-projects/spring-boot/commit/61736e9958755ac70d804599728cd8345ec458d3) which adds an example.

PR #28173, opened 2021-10-01
Issue 27793: Add smoke test with Spring Session and Hazelcast

kandulsh:
Added smoke test with spring session and Hazelcast


vpavic:
Fair enough - I didn't consider the user perspective to still be of primary interest here since these aren't considered _samples_ so I focused on the actual touching points of Spring Boot's integration for Spring Session.

PR #28204, opened 2021-10-06
Log URLs for all DataSources.

### LINKED ISSUE #27881: Log URLs for all DataSources when H2 Console is auto-configured
I'm using two datasources (Oracle + h2) in my spring boot project. I configure them manually in my configuration class, and the Oracle Datasource is annotated with Primary. When my project starts, the H2ConsoleAutoConfiguration prints out its log:
 ```o.s.b.a.h2.H2ConsoleAutoConfiguration    : H2 console available at '/h2-console'. Database available at 'jdbc:oracle:thin:@xxx'```
I replace my oracle ip and sid with xxx.

The expected behavior is H2ConsoleAutoConfiguration should print jdbc:h2:mem:xxx instead of jdbc:oracle:thin.

I delve into H2ConsoleAutoConfiguration.java and find the following code:
```
dataSource.ifAvailable((available) -> {
    try (Connection connection = available.getConnection()) {
        logger.info("H2 console available at '" + path + "'. Database available at '"
                + connection.getMetaData().getURL() + "'");
    }
    catch (Exception ex) {
        // Continue
    }
});
```
Is it better to check the connection URL starts with jdbc:h2 such that the log can print the correct h2 database connection?

Spring Boot version: 2.4.5
my sample project:
[multi-datasource.zip](https://github.com/spring-projects/spring-boot/files/7118572/multi-datasource.zip)

StitzL:
Always log path of H2 console, then log details for each data source that offers a connection, in order, primary first.

Resolves #27881


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/h2/H2ConsoleAutoConfiguration.java
```diff
@@ -63,9 +63,10 @@
 		String urlMapping = path + (path.endsWith("/") ? "*" : "/*");
 		ServletRegistrationBean<WebServlet> registration = new ServletRegistrationBean<>(new WebServlet(), urlMapping);
 		configureH2ConsoleSettings(registration, properties.getSettings());
-		dataSource.ifAvailable((available) -> {
+		logger.info("H2 console available at '" + path + "'.");
+		dataSource.orderedStream().forEachOrdered((available) -> {
 			try (Connection connection = available.getConnection()) {
-				logger.info("H2 console available at '" + path + "'. Database available at '"
+				logger.info("Database available at '"
```
snicoll, line 69: I'd rather create a single log entry with them. Ideally, something that shortcuts to the following log if there is only one and an optimized one for the multi-datasources scenario.

snicoll (changes requested):
Thanks for the PR.

We already have a test for the single datasource use case (see `dataSourceUrlIsLoggedWhenAvailable`). We'd need an additional test for the code that you've changed. Can you please look at that?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/h2/H2ConsoleAutoConfiguration.java
StitzL: Thank you for the kind suggestion, I'll implement that. I actually this in mind as well, but did not want to overcomplicate things, this being my first contribution.

snicoll: Yeah, let's see how that will look like and we can refine if necessary. Thanks for the contribution!

StitzL:
> We already have a test for the single datasource use case (see `dataSourceUrlIsLoggedWhenAvailable`). We'd need an additional test for the code that you've changed. Can you please look at that?

Sure, I just have to figure out how to create a second datasource in that test scenario.

StitzL:
Condensed the log to a single entry with 4b0edea, which complicated the code.

Added tests for boundary conditions with 8f78983.

StitzL:
My pleasure! Thanks for the feedback and your work on this as well.

PR #28211, opened 2021-10-06
add SmallRye Mutiny dependency

hantsy:
Aligned to Spring 5.3.10: https://github.com/spring-projects/spring-framework/pull/27331 and Spring Data Commons:
 https://github.com/spring-projects/spring-data-commons/pull/2471


snicoll:
@hantsy thanks but we generally don't add dependency management for a library that we don't provide auto-configuration for.

hantsy:
I just saw RxJava 1 and RxJava 2 in the dependencies, so tried to add these to manage them.

PR #28290, opened 2021-10-11
Add support for IdlePartitionEventInterval

pascal-ayotte:
IdlePartitionEventInterval was added, but the value is never retrieved from configuration. Adding support to set the value from configuration. No Issue found related to this.


snicoll:
@pascal-ayotte thank you for making your first contribution to Spring Boot.

PR #28292, opened 2021-10-11
Add option to customize build cache volume names

### LINKED ISSUE #25994: Allow build image cache volume names to be customized
When the Maven `spring-boot:build-image` goal or Gradle `bootBuildImage` task invokes the CNB lifecycle API, it creates and mounts volumes for the lifecycle to use as build and launch caches. These volume names are determined by the plugins, but in some cases it is useful to let the user override the default names with names of their choosing. 

See https://github.com/buildpacks/pack/issues/665 for one use case and the [Pack cache options RFC](https://github.com/buildpacks/rfcs/blob/pack-cache/text/0000-pack-cache-options.md) for the `pack` CLI implementation of a similar capability.

TSRBerry:
This PR adds the option `cacheVolumeNames` to the spring-boot maven/gradle plugins, which allows the specification of custom build cache volume names.

At the moment the name can be customized for the following volume types:
- `launch`
- `build`

In Gradle specifying cache volume names can be done like this:
```
bootBuildImage {
    cacheVolumeNames = [
        "build": "example-build-cachevol",
        "launch": "example-launch-cachevol"
    ]
}
```

This PR should resolve #25994.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/BuildRequest.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/BuildRequest.java
@@ -38,6 +38,7 @@ import org.springframework.util.Assert;
  * @author Andrey Shlykov
  * @author Jeroen Meijer
  * @author Rafael Ceccone
+ * @author Julian Liebig
  * @since 2.3.0
  */
 public class BuildRequest {
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Lifecycle.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Lifecycle.java
@@ -39,6 +39,7 @@ import org.springframework.util.Assert;
  * @author Phillip Webb
  * @author Scott Frederick
  * @author Jeroen Meijer
+ * @author Julian Liebig
  */
 class Lifecycle implements Closeable {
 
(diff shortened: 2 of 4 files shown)


scottfrederick:
@jeliebig Thank you very much for your contribution. Your commits have been merged with some changes that you can see in [this commit](https://github.com/spring-projects/spring-boot/commit/871468931fd6b600fdbe9d38b42d24fdf806a046). 

After reviewing the Cloud Native Buildpacks changes that this issue was meant to match, we decided that we needed an approach that provided a DSL for configuring the caching instead of a simpler map structure.  We also have a desire to support other cache options in the future to align with the CNB direction [documented in an RFC](https://github.com/buildpacks/rfcs/blob/main/text/0091-pack-cache-options.md). I've added #28386 and #28387 to track these future cache options in Spring Boot.

scottfrederick:
> It seems like you had to do a lot of work to merge my PR which I feel like is not optimal. Do you have any feedback for me like things that I could work on, so you don't have to spend so much time on my PRs in the future?

This was not a case of you doing anything wrong. The original issue #25994 was written to track work that was done in the `pack` CLI so we can try to maintain parity across CNB tooling, but the issue did not have enough detail and design documented in it. We will often tag issues like that with `pending-design` to indicate that more work needs to be done, but we did not do that in this case. After looking at your contribution and the [CNB RFC](https://github.com/buildpacks/rfcs/blob/main/text/0091-pack-cache-options.md) that discusses caching changes, it was apparent that we should design this feature with future work in mind. I was keen to get this change into the upcoming `2.6.0-RC1` release so I made the changes without further discussion. Given more time, we could have discussed the design and implementation in the PR. 

Even though further work was done, your contribution was very thorough and complete. Thanks again for that. 

> Is there any way for me to discuss the approach with [the team] beforehand?

You are welcome to add comments to an issue to get clarity on a proposed feature or to say that you'd like to work on a PR for the issue. Sometimes that will prompt us to look closer to see if there is additional design that should be taken into account, and discuss the implementation. 

> Which leads me to my question, are there any optimizations that I could do to improve [the caching] situation? 

I think everything else you've mentioned here is under the control of the CNB builder and buildpacks. Spring Boot can configure caches but only buildpacks can decide what to cache. If you're using the Paketo builder, you can raise issues in the appropriate Paketo GitHub project or [start a discussion](https://paketo.io/community/) with that community. Caching of JRE/JDK specifically has been [discussed with the Paketo team](https://github.com/paketo-buildpacks/java/issues/86) in the past.

jdupont22:
Hi @TSRBerry, @scottfrederick 

Thanks for the try (2021 i know), but buildpack and pack cli now support this king of use case.
Please see [pack --cache](https://buildpacks.io/docs/tools/pack/cli/pack_build/)

PR #28308, opened 2021-10-13
spring-boot-starter-oauth2-client has an unnecessary dependency on com.sun.mail:jakarta.mail

martin-v:
oauth2-client don't depend on any mail dependency, so it's just confusing and causing problems with `javax.mail.*` to `jakarta.mail.*` package name change.

Was introduced with d6a869fa98eaeb3c180a8a3d94dcce8e3f51e204, probably there are more similar cases.

### DIFF
--- a/spring-boot-project/spring-boot-starters/spring-boot-starter-oauth2-client/build.gradle
+++ b/spring-boot-project/spring-boot-starters/spring-boot-starter-oauth2-client/build.gradle
@@ -6,11 +6,8 @@ description = "Starter for using Spring Security's OAuth2/OpenID Connect client 
 
 dependencies {
 	api(project(":spring-boot-project:spring-boot-starters:spring-boot-starter"))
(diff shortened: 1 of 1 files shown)


martin-v:
Thanks for the fast response.

> Can you expand a bit on this, please? If there are other cases, it'd be good to address them as well but I don't understand why you think they're likely.

I just took a quick look at the commit d6a869fa98eaeb3c180a8a3d94dcce8e3f51e204 and saw that there are more exclude/add cases similar to one I reverted in the PR. So I guessed that it is possible that here are more cases where the exclude/add is no longer necessary. E.g. the upstream project uses the jakarta version by itself now. (would be just a code smell, because the dependency is still necessary) or as in this PR case the dependency is obsolete now. But I didn't check any other cases.

martin-v:
Thanks, it was a pleasure :)

PR #28344, opened 2021-10-14
spring-boot-starter-test has an unnecessary dependency on jakarta.xml.bind-api

martin-v:
Same reason as in gh-28308 introduced with d6a869fa98eaeb3c180a8a3d94dcce8e3f51e204

@wilkinsona exclusion was removed in 2fb8c8d27e277150c4f5bf1037f5c76e24292a42


martin-v:
I checked the complete 2fb8c8d27e277150c4f5bf1037f5c76e24292a42 but this was the only one I found.

philwebb:
I think we should leave this one in and fix #28332 to allow it.

PR #28361, opened 2021-10-15
Do not auto-configure a UserDetailsService when an AuthenticationManagerResolver bean is present

### LINKED ISSUE #26766: Make UserDetailsServiceAutoConfiguration conditional on missing bean of type AuthenticationManagerResolver
When setting up an OAuth2 resource server with multi-tenancy (e.g., the one from the example https://github.com/spring-projects/spring-security-samples/tree/main/servlet/spring-boot/java/oauth2/resource-server/multi-tenancy ), none of the conditions for disabling the `UserDetailsServiceAutoConfiguration` are met and therefore, an InMemoryUserDetailsManager is auto-configured:
```
2021-06-01 18:34:39.832  INFO 26857 --- [           main] .s.s.UserDetailsServiceAutoConfiguration : 

Using generated security password: f89a4c05-3b75-4c2c-89a0-60dceba7e4ec
```

From my point of view, in this use case, the generation of a default passowrd is not desirable (I think a similar topic was addressed in PR #17646).

As an enhancement, I would therefore propose to also deactivate the `UserDetailsServiceAutoConfiguration` for such cases.

One way I see to achieve this, would be to make `UserDetailsServiceAutoConfiguration`conditional on missing beans of type `org.springframework.security.authentication.AuthenticationManagerResolver` as well. As far as I can tell, this would be consistent with the fact that `UserDetailsServiceAutoConfiguration` is already conditional on missing `AuthenticationManager` beans and, in some sense, having an `AuthenticationManagerResolver` is equivalent to having an `AuthenticationManager`. However, I have not fully investigated the implications of this yet. There may be better ways to achieve this.

bartoszJaszczak:
Resolves #26766. As mentioned in the issue, AuthenticationManagerResolver was added to the ConditionalOnMissingBean in the UserDetailsServiceAutoConfiguration.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/servlet/UserDetailsServiceAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/servlet/UserDetailsServiceAutoConfiguration.java
@@ -32,6 +32,7 @@ import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Lazy;
 import org.springframework.security.authentication.AuthenticationManager;
+import org.springframework.security.authentication.AuthenticationManagerResolver;
 import org.springframework.security.authentication.AuthenticationProvider;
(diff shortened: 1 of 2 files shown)


wilkinsona:
@bartoszJaszczak Thanks very much for making your first contribution to Spring Boot.

PR #28495, opened 2021-11-01
Log a debug message with the reason for restarting devtools

Artur-:
This helps with debugging certain situations when Spring Boot keeps restarting and the reason is unclear


snicoll:
Thanks @Artur-. I've moved and rephrased the log message so that it fits better in b81c004.

Issue #22358, opened 2020-07-16
Remove Elasticsearch RestClient auto-configuration


latuszek:
spring-boot-starter-parent v2.3.1.RELEASE
elasticsearch-rest-high-level-client (version inherited from parent above)

According to spring-boot documentation:

> If you have the org.elasticsearch.client:elasticsearch-rest-high-level-client dependency on the classpath, Spring Boot will auto-configure a RestHighLevelClient, which wraps any existing RestClient bean, reusing its HTTP configuration.

In our project, we have a single `RestClient` bean registered in spring context:

```java
    @Bean
    public RestClient elasticsearchRestClient(ElasticsearchProperties elasticsearchProperties) {
        RestClientBuilder builder = RestClient.builder(
                new HttpHost(elasticsearchProperties.getHost(),
                        elasticsearchProperties.getPort(),
                        elasticsearchProperties.getScheme())
        );

        ElasticsearchCredentials esCredentials = new ElasticsearchCredentials(elasticsearchProperties);
        if (esCredentials.hasCredentials()) {
            builder.setHttpClientConfigCallback(httpClientBuilder -> httpClientBuilder.setDefaultCredentialsProvider(
                    esCredentials.getCredentialsProvider()));
        }

        return builder.build();
    }
```

Spring actuator health endpoint returns correct response regarding connection to Elasticsearch (uses our parameters etc.).
However when we're trying to use injected `RestHighLevelClient` to connect (e.g. using `client.cluster().health(new ClusterHealthRequest(), RequestOptions.DEFAULT);` it connects using default parameters from spring-boot.

snicoll:
Thanks for the report. I can see the documentation does not match what the auto-configuration does as far as I can see and I am not sure I understand why. I don't see a way to create a `RestHighLevelClient` based on an existing `RestClient` so perhaps something was changed in ElasticSearch and the documentation wasn't updated?

In any case, we should not create the high-level client if a `RestClient` exists I guess.

As for the sample above, You should not inject `ElasticSearchProperties` in your own code but rather use `RestClientBuilderCustomizer` for whatever customization that you need to do on the builder. This will fix this issue as a side-effect of it.

I've flagged this issue for team attention to see if someone else on the team can fill the blanks.

bclozel:
I guess this is mainly a documentation issue.

It should be instead:

[...]

PR #28496, opened 2021-11-01
Support Elasticsearch RestClientBuilder auto-configuration without RestHighLevelClient

filiphr:
Prior to this commit, Spring Boot would only auto-configure
the `RestHighLevelClient` and `RestClientBuilder` if the `RestHighLevelClient`
was present. This was done in 1d73d4eda75ef0d18c747c57aa2b3385674483c7 as part of https://github.com/spring-projects/spring-boot/issues/22358

This commit brings back the exposing of the `RestClient` bean in
Spring Boot when exposing the `RestHighLevelClient` or
when the `RestHighLevelClient` is not present.
It allows for using the Spring Boot auto configuration and its customizers
of the `RestClientBuilder` in a similar way as it is done for the `RestTeamplateBuilder`
and the `WebClient.Builder`.
Now the presence of the `org.elasticsearch.client:elasticsearch-rest-high-level-client` is optional.
This opens the door for potentially adding support to the new
[Elasticsearch Java Client](https://github.com/elastic/elasticsearch-java) that is based on the same `RestClient`

-----

The exposing of the `RestClient` as a bean is not the most important thing in this PR.
I would rather say that the fact that the `RestClientBuilder` is exposed as a bean is the most important part.
This allows users to rely on the Spring Boot provided mechanisms for configuring the `RestClientBuilder` and not depend on the now deprecated Elasticsearch Rest High Level Client.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/elasticsearch/ElasticSearchRestHealthContributorAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/elasticsearch/ElasticSearchRestHealthContributorAutoConfiguration.java
@@ -16,44 +16,31 @@
 
 package org.springframework.boot.actuate.autoconfigure.elasticsearch;
 
-import java.util.Map;
-
 import org.elasticsearch.client.RestClient;
 
-import org.springframework.boot.actuate.autoconfigure.health.CompositeHealthContributorConfiguration;
+import org.springframework.boot.actuate.autoconfigure.elasticsearch.ElasticSearchRestHealthContributorConfigurations.RestClientHealthContributorConfiguration;
 import org.springframework.boot.actuate.autoconfigure.health.ConditionalOnEnabledHealthIndicator;
-import org.springframework.boot.actuate.elasticsearch.ElasticsearchRestHealthIndicator;
-import org.springframework.boot.actuate.health.HealthContributor;
 import org.springframework.boot.autoconfigure.AutoConfigureAfter;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
-import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.elasticsearch.ElasticsearchRestClientAutoConfiguration;
-import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
 
 /**
- * {@link EnableAutoConfiguration Auto-configuration} for
- * {@link ElasticsearchRestHealthIndicator} using the {@link RestClient}.
+ * {@link EnableAutoConfiguration Auto-configuration} for Elasticsearch health indicator
+ * using REST clients.
  *
  * @author Artsiom Yudovin
  * @since 2.1.1
  */
 @SuppressWarnings("deprecation")
 @Configuration(proxyBeanMethods = false)
-@ConditionalOnClass(org.elasticsearch.client.RestHighLevelClient.class)
-@ConditionalOnBean(org.elasticsearch.client.RestHighLevelClient.class)
+@ConditionalOnClass(RestClient.class)
 @ConditionalOnEnabledHealthIndicator("elasticsearch")
 @AutoConfigureAfter(ElasticsearchRestClientAutoConfiguration.class)
-public class ElasticSearchRestHealthContributorAutoConfiguration extends
-		CompositeHealthContributorConfiguration<ElasticsearchRestHealthIndicator, org.elasticsearch.client.RestHighLevelClient> {
-
-	@Bean
-	@ConditionalOnMissingBean(name = { "elasticsearchHealthIndicator", "elasticsearchHealthContributor" })
-	public HealthContributor elasticsearchHealthContributor(
-			Map<String, org.elasticsearch.client.RestHighLevelClient> clients) {
-		return createContributor(clients);
-	}
+@Import({ ElasticSearchRestHealthContributorConfigurations.RestHighLevelClientHealthContributorConfiguration.class,
+		RestClientHealthContributorConfiguration.class })
+public class ElasticSearchRestHealthContributorAutoConfiguration {
 
 }
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/elasticsearch/ElasticSearchRestHealthContributorConfigurations.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/elasticsearch/ElasticSearchRestHealthContributorConfigurations.java
@@ -1,0 +1,69 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.autoconfigure.elasticsearch;
+
+import java.util.Map;
+
+import org.elasticsearch.client.RestClient;
+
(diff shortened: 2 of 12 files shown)


filiphr:
I didn't do any changes to the Gradle Wrapper. Seems like some problem in the Gradle Wrapper action itself

bclozel:
This PR is reversing the team's decision explained in https://github.com/spring-projects/spring-boot/issues/22358#issuecomment-679937047. I'm not sure I understand the rationale behind this. The presence of `elasticsearch-rest-high-level-client` has always been optional, but we've found that the low level client was of little use in itself at the application level.

If the main goal is to support the newly released client, maybe we should auto-configure as well and deprecate the high-level variant in 2.6.x?

filiphr:
The biggest reason why I created this PR was to get the `RestClientBuilder` as a bean in order to rely on the Spring properties / customizers. With the current setup of things you have to have `elasticsearch-rest-high-level-client` in order to create your own `RestClient`.

The reason for that is

https://github.com/spring-projects/spring-boot/blob/83e44305122a9bd45194cd5db649ad3a12b64690/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientAutoConfiguration.java#L41-L42

We have a platform that heavily uses Spring Boot and we are using the `RestClient` directly and not the `RestHighLevelClient`. We started removing the dependency entirely, since we want to completely not depend on it due to Licensing issues. The High Level Client is since 7.11 no longer Apache 2.0 licensed. I thought that everything will keep working as expected, but then the `RestClientBuilder` was not present anymore.

The addition of the `RestClient` as a bean is an addition to the original requirement that we had. I honestly, thought that it is not that complicated to also expose it as a bean since it can be exposed through the `RestHighLevelClient#getLowLevelClient` easily and if the `RestHighLevelClient` is not there then it can easily be created through the `RestClientBuilder`.  That's why I did it.

Even if you only partially integrate this (without the new beans) we can easily do

```

@Bean
public RestClient elasticsearchRestClient(RestTemplateBuilder builder) {
    return builder.build();
}
```

and have the `RestClient` bean in our own application without the need to depend on the `RestHighLevelClient` and without the need to copy things that are already done in Spring Boot.

----

In addition to all of this, there is also [OpenSearch](https://opensearch.org/) with its own things. Without going into too much detail, for our own simple purposes it might be possible to use one `RestClient` to communicate with both types of clusters.


-----

In a nutshell for us having to maintain support for our own things on both OpenSearch and Elasticsearch it makes a lot of sense to only use the Low Level Clients. I hope that you are going to take this into consideration and revise your decision about this. Thanks a lot for the understanding

bclozel:
Thanks for the background information @filiphr, that's super useful.

With that I think we should:
* bring back `RestClient` or its builder as a bean. This should allow the OpenSearch use case
* schedule the support for the new official client and deprecate the current one

filiphr:
Something else, that I forgot to add to this PR. The `ElasticSearchRestHealthContributorAutoConfiguration` needs to be adapted as well.

Ideally the health contributor will be adapted to only use `RestClient` or somehow a combination of things (if the `RestHighLevelClient` is there). e.g. All `RestClient` beans plus the `RestHighLevelClient#getRestLowLevelClient()` that are not exposed as beans. The `ElasticsearchRestHealthIndicator` needs to be adapted to only take the `RestClient` as a constructor in order to not have a mandatory dependency on the `RestHighLevelClient`. If this is done then the same class can be used both for the old `RestHighLevelClient` and new Elasticsearch Java Client.

I can of course provide PRs and / or adapt this PR if you decide that you are OK with my proposals.

filiphr:
One, I hope final, argument is also the fact that removing the dependency on the `RestHighLevelClient` just shaved off ~20MB from our final jar. Not sure if things like this are important from your perspective or not, but I wanted to share this as well

filiphr:
Thanks a lot for this decision @bclozel.

I am going to adapt this PR to take into consideration for 1) and 2).

What about exposing `RestClientBuilder` as a bean when the `RestHighLevelClient` is not present, but the `RestClientBuilder` is. Will you accept a separate PR for 2.6 that will do this?

I have managed to achieve what I need with 2.5 already, but the way I did it is not the best approach, since I am doing some nasty magic with `ImportSelector`, and I would rather get rid of that nasty magic we have.

bclozel:
> What about exposing RestClientBuilder as a bean when the RestHighLevelClient is not present, but the RestClientBuilder is. Will you accept a separate PR for 2.6 that will do this?

Exposing the `RestClientBuilder` might work for your case, but I'm wondering if it's the right choice here: I don't think we usually want to have several instances of `RestClient` in the same application? That would create several HTTP client instances, right? Even if the application itself only creates one, we'd need to reuse it for the health indicator?

It's a bit late in the 2.6 cycle unfortunately so if we want to introduce a change here, it has to be extra-safe.

filiphr:
@bclozel I think that I have adapted the PR as you asked. Please have a look at it and let me know if there is something more that needs to be done.

Also adapted the use of `ElasticsearchRestHealthIndicator`. I deprecated the constructor using `RestHighLevelClient`. Not sure what your policy for removal there is. Ideally I'd like to get rid of the constructor, since the `ElasticsearchRestHealthIndicator` in the current state is not usable when the `RestHighLevelClient` is not there. I think that the class will not load at all.

-----

>Exposing the `RestClientBuilder` might work for your case, but I'm wondering if it's the right choice here: I don't think we usually want to have several instances of `RestClient` in the same application? That would create several HTTP client instances, right? Even if the application itself only creates one, we'd need to reuse it for the health indicator?

Exposing the `RestClientBuilder` does not mean that there will be several instances of the `RestClient`. The build method needs to be invoked for a `RestClient` to be created. Which means that from a Spring Boot point of view for 2.6 when the `RestHighLevelClient` is not there then there will be a bean of type `RestClientBuilder`. The users can then choose and decide what to do with that `RestClientBuilder`. In order to reduce the amount of changes the health indicator will not use it. People will need to expose their own health indicator in 2.6. And as I mentioned in the previous paragraph the current `ElasticsearchRestHealthIndicator` is not capable to be used without the `RestHighLevelClient`.

bclozel:
@filiphr you're right, I forgot we were already exposing the builder as a bean. I need to take a deeper look.

filiphr:
@bclozel I'll try to do a small proposal as a separate PR for that. You can then decide what you want to do

filiphr:
@bclozel I did a proposal in PR #28655. Please have a look at it and let me know what you think. This PR and that one of course have some overlap. However, based on your decision, I can adapt this PR on top of that other one if needed.

filiphr:
@bclozel I finally found the time and adapted the changed in this PR to address point 1) and 2) from https://github.com/spring-projects/spring-boot/pull/28496#issuecomment-965569351. I've changed the base of the PR to be 2.7. Please have a look at it and let me know what you think about it.

snicoll:
For the record `2.7.x` has been upgraded to Elasticsearch `7.16.2`.

filiphr:
>For the record `2.7.x` has been upgraded to Elasticsearch `7.16.2`.

Damn, I first rebased on main, but then realized that main and 2.7.x are not the same, so I did it on 2.7.x. I'll try to do take the new changes from 2.7.x into consideration and fix the conflicts.

Basically, I need to make sure that there are no deprecation warnings, right?

wilkinsona:
Thanks very much, @filiphr. If you're interested, I reworked things a bit in https://github.com/spring-projects/spring-boot/commit/a7a71da9ef159e375ff0faa344e888d435a7052c. This was primarily to simplify the changes to the health indicator. You can see the net changes in https://github.com/spring-projects/spring-boot/commit/227c3164f1c0f08410d72666fec64bb989d061c5.

filiphr:
@wilkinsona I agree with you that the code is simplified with your change. However, your change means that you can only have Elasticsearch health checks if you have the `RestHighLevelClient` on the classpath, because the loading of `ElasticsearchRestHealthIndicator` will fail due to the `RestHighLevelClient` being used in one of its constructors.

I personally see no need to lose Elasticsearch auto configuration if you do not have `RestHighLevelClient` on the classpath. Due to the changes in `ElasticSearchRestHealthContributorAutoConfiguration` it also means that the application will not boot up properly when you only have the `RestClient` available.

If this is something that the Boot team wants to do, then I am all OK with it. However, I would ask you to reconsider and have health checks without the RestHighLevelClient.

wilkinsona:
That’s an unintentional over-simplification. Thanks for catching it, @filiphr. I’ll take another look. Requiring a `RestClient` bean for auto-configuration of the health indicator looks like it could be a good middle ground.

filiphr:
> Requiring a `RestClient` bean for auto-configuration of the health indicator looks like it could be a good middle ground.

It is not about the auto-configuration. The auto configuration works correctly and it auto configures if a `RestClient` bean is there. The problem is the fact that the loading of the `ElasticsearchRestHealthIndicator` will fail when the `RestHighLevelClient` class is not there.

filiphr:
Thanks a lot for bringing back the old changes @wilkinsona. It is really appreciated.

nightswimmings:
@wilkinsona  Are we opensearch users expected to keep using opensearch.RestHighLevelClient after 2.7.0?

wilkinsona:
Spring Boot does not have any auto-configuration for OpenSearch so we don't have any expectations about what you should do.

filiphr:
@nightswimmings as far as I know, you can keep using the Elasticsearch `RestClient` with OpenSearch. You can't use the `RestHighLevelClient` though.

PR #28497, opened 2021-11-01
Replace "etc" in reference documentation

Buzzardo:
as part of issue 27895.


philwebb:
Thanks Jay, this one is now in `main`.

PR #28501, opened 2021-11-02
Polish

polarbear567:
Polish in some modules.


snicoll:
@polarbear567 thanks for the PR. Would you please make sure your branch is up to date with `main` before contributing? This PR has 58 unrelated commits. I've polished them but if you'd be great if the PR only contained the intended changes.

polarbear567:
> @polarbear567 thanks for the PR. Would you please make sure your branch is up to date with `main` before contributing? This PR has 58 unrelated commits. I've polished them but if you'd be great if the PR only contained the intended changes.

Sure. In fact, I'll update the latest code every time. It seems that there is no problem manually before, but recently I have used the fetch upstream function provided by the GitHub page. It seems that this problem has occurred. I'm not sure, but next time I will switch back to manually updating the latest code. Finally, thanks for the reminder.

PR #28548, opened 2021-11-07
Fix typo in "Ant-style path matching"

slowjoe007:
When reading the documentation I stumbled upon this typo. I know ANT-style path matching... patch matching I never heard of.


snicoll:
@slowjoe007 thank you for making your first contribution to Spring Boot.

PR #28563, opened 2021-11-08
Ignore JUnit annotations when caching test contexts

ledoyen:
Hello spring-team :wave:

As JUnit Jupiter comes with many new features through annotations (such as `@TestInstance` or `@Order`), it makes sens that they be ignored by the *key* created by `ImportsContextCustomizer`.

This is a small patch, but one can wonder if the *key* should not consider only `org.springframework.*` annotations or supply an extension point for custom user-defined annotations to be ignored when creating the context caching key (part).

Let me know if you need more contexts.

Best regards !


philwebb:
I think it makes sense to ignore `org.junit` imports. We already do the same for Spock. Unfortunately I don't think we can just limit ourselves to `org.springframework.*` since users might have their own annotations. I do think an extension point is a good idea. I've raised https://github.com/spring-projects/spring-boot/issues/28564 for that.

PR #28655, opened 2021-11-12
Expose RestClientBuilder when RestHighLevelClient is not available

filiphr:
This commits exposes the RestClientBuilder as a bean even when the RestHighLevelClient is not available
It allows users to create their own RestClient beans using the Spring Boot configured RestClientBuilder
when they are not using the RestHighLevelClient.

In addition to that a new ElasticsearchRestClientHealthIndicator decoupled from the RestHighLevelClient has been added
and is used as an elasticsearch health contributor when there are no RestHighLevelClient beans.

This is similar to PR #28496, with the difference being that this is a potential proposal for 2.6.0. With this people can freely remove the dependency on the `RestHighLevelClient` and have all the benefits from Spring Boot in 2.6.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/elasticsearch/ElasticsearchRestHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/elasticsearch/ElasticsearchRestHealthIndicator.java
@@ -16,75 +16,31 @@
 
 package org.springframework.boot.actuate.elasticsearch;
 
-import java.io.InputStream;
-import java.nio.charset.StandardCharsets;
-import java.util.Map;
-
-import org.apache.http.HttpStatus;
-import org.apache.http.StatusLine;
-import org.elasticsearch.client.Request;
-import org.elasticsearch.client.Response;
 import org.elasticsearch.client.RestClient;
 import org.elasticsearch.client.RestHighLevelClient;
 
-import org.springframework.boot.actuate.health.AbstractHealthIndicator;
-import org.springframework.boot.actuate.health.Health;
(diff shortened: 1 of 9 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/elasticsearch/ElasticsearchRestHealthIndicator.java
```diff
@@ -16,75 +16,31 @@
...
  */
-public class ElasticsearchRestHealthIndicator extends AbstractHealthIndicator {
-
-	private static final String RED_STATUS = "red";
-
-	private final RestClient client;
-
-	private final JsonParser jsonParser;
+@Deprecated
+public class ElasticsearchRestHealthIndicator extends ElasticsearchRestClientHealthIndicator {
```
filiphr, line 36: I really wanted to avoid deprecating this. However, it is not possible to use this class when the `RestHighLevelClient` is not available.

bclozel:
@filiphr I've merged this PR, removing the bits about the health indicators.
While that part was totally valid, we're really late in the milestone phase and we repurposed this PR as a bug to only fix the `RestClientBuilder` auto-configuration part.

I know this won't solve everything on your side, but I hope it will make things a bit easier for the 2.6.x timeline.

filiphr:
Thanks a lot for bringing this into the 2.6.x timeline @bclozel. It helps us a lot to properly use the Spring Boot support here.

>removing the bits about the health indicators.

I completely understand this. I was not entirely sure about it left it there to leave it up to you. Adding our own health contributor is super trivial.

Does the change about the health contributors make sense for the 2.7 release? If yes I will bring those changes to PR #28496

Issue #28046, opened 2021-09-17
Allow ConfigurationProcessor to work with Kotlin Symbol Processor


kkocel:
The problem:
Currently, ConfigurationProcessor is backed by Annotation Processing. In Kotlin annotation processors can be used thanks to KAPT. Sadly KAPT does not work with JDK 16+ ( https://youtrack.jetbrains.com/issue/KT-45545 ). For Kotlin there are other processors like KSP ( https://github.com/google/ksp ), yet they do not rely on Java Annotation processing.

Proposed solution:
Provide an abstraction layer on ConfigurationMetadataAnnotationProcessor that will allow plugging in other processors. 
Additionally provide an adapter for KSP.

@sdeleuze

wilkinsona:
Thanks for raising this.

> Sadly KAPT does not work with JDK 16+ ( https://youtrack.jetbrains.com/issue/KT-45545). 

My reading of that Kotlin issue is that KAPT does with JDK 16+, but that it requires adding some JVM arguments to open up some of the JVM's internals to KAPT. Have I misunderstood the discussion?

kkocel:
@wilkinsona You understood it better than I :) Yes, it's possible to break `javac` encapsulation using `add-opens` flags, but ihmo it's temporary (this flag can be taken away as with `--illegal-access=permit` in JDK 17).

I know that adding an abstraction layer over processors can add in additional complexity, but it can also be a safer solution (also please note that KSP authors claim that it's up to 2x faster than java annotation processors). 

As with everything you should weigh your options as Spring maintainers.

bespaltovyj:
Hi, @wilkinsona. 
The new Kotlin [documentation ](https://kotlinlang.org/docs/kapt.html) says that `kapt` is in maintenance mode and new features are not planned. 
And recommend using `ksp`(it should work faster).
![image](https://user-images.githubusercontent.com/32296746/139596680-256890e3-57a8-436d-a564-1887763a4f6c.png)
Will you support ksp after this news?
If you don't mind, then I can start adding `ksp` support.

kkocel:
@wilkinsona I asked about relation of KSP and Kotlin: https://twitter.com/relizarov/status/1455452359669264384

So it looks like KSP is the way to go for Kotlin

stillya:
[...]

PR #28668, opened 2021-11-13
Add Kotlin Symbol Processor for auto configuration

bespaltovyj:
KSP issue - https://github.com/spring-projects/spring-boot/issues/28046
I made a draft in which I added support for ksp.
I have a few questions:
1. Сan I use kotlin?
2. Do I need to create a separate gradle module for ksp?
3. Is there a description of the `spring-autoconfigure-metadata.properies` file somewhere? The result of processing java annotations is slightly different from the results of ksp and it is unclear whether these differences affect something.

If you tell me what fixes are needed now, then after that I will start doing ksp support for `spring-boot-configuration-processor`

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-autoconfigure-processor/build.gradle
+++ b/spring-boot-project/spring-boot-tools/spring-boot-autoconfigure-processor/build.gradle
@@ -1,5 +1,6 @@
 plugins {
 	id "java-library"
+	id "org.jetbrains.kotlin.jvm"
 	id "org.springframework.boot.conventions"
 	id "org.springframework.boot.deployed"
 	id "org.springframework.boot.annotation-processor"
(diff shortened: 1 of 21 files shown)


wilkinsona:
Thanks for opening this, @bespaltovyj. I haven't yet looked closely at what will be required to support KSP, but I had imagined that we would refactor the existing logic so that it can be called from either a Kotlin symbol processor or a Java annotation processor. Compared to duplicating the logic, this should make it easier to achieve consistent behaviour across the two processors particularly in the case of the `@ConfigurationProperties` processing which is quite complex. I would also prefer that we try to use Java as much as possible.

bespaltovyj:
Ok, I'll try to combine KSP and annotation processing in Java.

bespaltovyj:
@cd-ff I don't understand what you mean. Can add more details?

wilkinsona:
@bespaltovyj I believe @cd-ff was referring to the [earlier request to sign the CLA](https://github.com/spring-projects/spring-boot/pull/28668#issuecomment-968130188).

bespaltovyj:
I resign CLA yesterday

wslaghekke:
What would be required to allow this feature to be implemented, are there changes that need to be made or just tests that need to be fixed / need to work?

wilkinsona:
We chatted about this today and I'm afraid that we decided that we don't think this is something that we'll be able to merge. Sorry. We need to take a step back, consider our options for #28046 and decide how we want to tackle KSP support in general.

PR #28670, opened 2021-11-14
Fix yaml sample format in reference doc

weixsun:
The original configuration block is not a valid yaml syntax format.


snicoll:
Thanks @weixsun. I've dropped the second  commit as we order things in alphabetical order. Turns out that the properties sample below was a bit unusual and I've fixed that instead in 6d79599.

PR #28671, opened 2021-11-14
Configuration sample in reference doc has wrong yaml formatting

weixsun:
This pr is just a modification of the code style of the yaml code block.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
@@ -780,13 +780,13 @@ This style of configuration works particularly well with the `SpringApplication`
 [source,yaml,indent=0,subs="verbatim"]
 ----
 	my:
-		service:
-			remote-address: 192.168.1.1
-			security:
-				username: admin
-				roles:
-				  - USER
-				  - ADMIN
+	  service:
+	    remote-address: "192.168.1.1"
+	    security:
+	      username: "admin"
+	      roles:
+	      - USER
+	      - ADMIN
 ----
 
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
```diff
@@ -780,13 +780,13 @@ This style of configuration works particularly well with the `SpringApplication`
...
 	my:
-		service:
-			remote-address: 192.168.1.1
-			security:
-				username: admin
-				roles:
-				  - USER
-				  - ADMIN
+	  service:
+	    remote-address: "192.168.1.1"
```
snicoll, line 784: This PR is doing two different things here. Are the quote around the value necessary?

spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
```diff
@@ -780,13 +780,13 @@ This style of configuration works particularly well with the `SpringApplication`
...
+	    security:
+	      username: "admin"
```
snicoll, line 786: Ditto

snicoll (commented):
Thanks for the PR, I've left a couple comments.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
weixsun: Ha, I think the String type here is more suitable, so I just did it.

snicoll:
Thanks again @weixsun. I have reverted the quote change as discussed.

philwebb:
I think we should probably quote Strings in YAML. I tried to add a review note but I didn't click "submit" 🤦. I've opened #28709.

Issue #28603, opened 2021-11-10
Fix @deprecated and @see in org.springframework.boot.loader.archive.Archive's javadoc


wilkinsona:
I mangled them in https://github.com/spring-projects/spring-boot/commit/9c9fbf400e11c40cd10b883cf986a9d4a65009e4.

wilkinsona:
Closing in favor of #28677.

PR #28677, opened 2021-11-15
Fix @deprecated and @see in org.springframework.boot.loader.archive.Archive's javadoc

polarbear567:
For [issue 28603](https://github.com/spring-projects/spring-boot/issues/28603).


wilkinsona:
Thanks very much, @polarbear567.

PR #28710, opened 2021-11-16
Replace "you're"

Buzzardo:
with some other structure, often simplifying the sentence in the process.

Also checked for words that end with "'ll" and "n't" but did not find any.


wilkinsona:
Thanks, @buzzardo.

PR #28712, opened 2021-11-16
Fix "Configure Two DataSources" example

vpavic:
With the present example, the `secondDataSource` bean factory method will use `firstDataSourceProperties` due to it being annotated with `@Primary`.

This commit adds the `@Qualifier` needed to ensure `secondDataSource` bean factory method uses `secondDataSourceProperties`.


snicoll:
Good catch, thanks Vedran!

Issue #22109, opened 2020-06-25
Make it easier to enable Hibernate bytecode enhancement


mauromol:
Using Spring Boot 2.3.1.

I'm setting up a Spring Boot web application using Tomcat+JPA+Hibernate. I then wanted to enable bytecode enhancement for Hibernate.
The first attempt was to just add the following properties to my `application.properties` (just mentioning a couple of them), to enable runtime bytecode enhancement:

    spring.jpa.properties.hibernate.enhancer.enableDirtyTracking=true
    spring.jpa.properties.hibernate.enhancer.enableLazyInitialization=true

This, however, produces this at runtime:

```
org.springframework.beans.factory.BeanCreationException: Error creating bean with name 'entityManagerFactory' defined in class path resource [org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaConfiguration.class]: Invocation of init method failed; nested exception is java.lang.IllegalStateException: Cannot apply class transformer without LoadTimeWeaver specified
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.initializeBean(AbstractAutowireCapableBeanFactory.java:1796) ~[spring-beans-5.2.7.RELEASE.jar:5.2.7.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.doCreateBean(AbstractAutowireCapableBeanFactory.java:595) ~[spring-beans-5.2.7.RELEASE.jar:5.2.7.RELEASE]
	at org.springframework.beans.factory.support.AbstractAutowireCapableBeanFactory.createBean(AbstractAutowireCapableBeanFactory.java:517) ~[spring-beans-5.2.7.RELEASE.jar:5.2.7.RELEASE]
	at org.springframework.beans.factory.support.AbstractBeanFactory.lambda$doGetBean$0(AbstractBeanFactory.java:323) ~[spring-beans-5.2.7.RELEASE.jar:5.2.7.RELEASE]
	at org.springframework.beans.factory.support.DefaultSingletonBeanRegistry.getSingleton(DefaultSingletonBeanRegistry.java:226) ~[spring-beans-5.2.7.RELEASE.jar:5.2.7.RELEASE]
	at org.springframework.beans.factory.support.AbstractBeanFactory.doGetBean(AbstractBeanFactory.java:321) ~[spring-beans-5.2.7.RELEASE.jar:5.2.7.RELEASE]
[...]

PR #28740, opened 2021-11-18
Dependency manage hibernate-gradle-plugin

candrews:
`org.hibernate:hibernate-gradle-plugin` is the Hibernate byte code enhancer for Gradle: https://docs.jboss.org/hibernate/orm/5.6/topical/html_single/bytecode/BytecodeEnhancement.html

As part of Hibernate, it would be nice if Spring Boot dependency management version managed this dependency like it does for all of the other Hibernate dependencies already.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -427,6 +427,7 @@ bom {
 				"hibernate-ehcache",
 				"hibernate-entitymanager",
 				"hibernate-envers",
+				"hibernate-gradle-plugin",
 				"hibernate-hikaricp",
 				"hibernate-java8",
 				"hibernate-jcache",


wilkinsona:
Thanks for the suggestion, @candrews. Can you explain why you'd like the Gradle plugin to be included in our dependency management? AIUI, it won't allow you to omit the version when configuring a Gradle project to use the plugin.

candrews:
> Can you explain why you'd like the Gradle plugin to be included in our dependency management? AIUI, it won't allow you to omit the version when configuring a Gradle project to use the plugin.

True, but I'm hoping that we (or I) can fix that. I'm hoping that having this version specified in the BOM will be a step towards resolving https://github.com/spring-projects/spring-boot/issues/22109

wilkinsona:
Until Gradle offers support for managing plugin versions using a bom, I don't think we should add anything to `spring-boot-dependencies`. I think it would be confusing for a Gradle plugin to be managed as a standard dependency. Instead, they should be managed separately (as Maven plugins currently are via `<pluginManagement>`) once that separate mechanism is available. Thanks anyway for the suggestion.

PR #28767, opened 2021-11-22
LettuceMetricsAutoConfiguration should not build ClientResources instance

quaff:
It is not the responsibility of `ClientResourcesBuilderCustomizer` to build `ClientResources` instance.

Fix:
```
i.l.c.r.DefaultClientResources WARN io.lettuce.core.resource.DefaultClientResources was not shut down properly, shutdown() was not called before it's garbage-collected. Call shutdown() or shutdown(long,long,TimeUnit)
```


snicoll:
@quaff good catch, thanks again.

PR #28778, opened 2021-11-23
fix redundant content in ResourceProperties

polarbear567:
Please have a check if here is a redundant content.


wilkinsona:
This was intentional so that the javadoc does not repeat `WebProperties` and reads "accessed through WebProperties and getResources()" and not "accessed through WebProperties and WebPropertries.getResources()".

Issue #28772, opened 2021-11-22
server.servlet.session.cookie.same-site isn't applied to Spring Session's SESSION cookie


wilkinsona:
As [reported by @OrangeDog on Gitter](https://gitter.im/spring-projects/spring-boot?at=619b7f21026dc63ca60d7ed3), there's an unfortunate mismatch between Servlet's default cookie name (`JSESSIONID`) and Spring Session's default cookie name (`SESSION`). This mismatch means that the `server.servlet.session.cookie.same-site` property has no effect when using Spring Session.  I think that setting `server.servlet.session.cookie.name=SESSION` will gets things working. We should confirm that this is the case and also see if there's something that we can do so that this works out of the box.

OrangeDog:
Setting `server.servlet.session.cookie.name=SESSION` does not appear to have any effect.

wilkinsona:
Thanks for trying that out, @OrangeDog. It doesn't work as I misdiagnosed the cause here. It's actually simpler than I thought and we just need to map the property onto Spring Session's `DefaultCookieSerializer`. You can do that yourself as a workaround in the meantime by defining the following bean in your application:

```java
@Bean
DefaultCookieSerializer cookieSerializer(ServerProperties serverProperties,
        ObjectProvider<DefaultCookieSerializerCustomizer> cookieSerializerCustomizers) {
    Cookie cookie = serverProperties.getServlet().getSession().getCookie();
    DefaultCookieSerializer cookieSerializer = new DefaultCookieSerializer();
    PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
    map.from(cookie::getName).to(cookieSerializer::setCookieName);
    map.from(cookie::getDomain).to(cookieSerializer::setDomainName);
    map.from(cookie::getPath).to(cookieSerializer::setCookiePath);
    map.from(cookie::getHttpOnly).to(cookieSerializer::setUseHttpOnlyCookie);
    map.from(cookie::getSecure).to(cookieSerializer::setUseSecureCookie);
    map.from(cookie::getMaxAge).asInt(Duration::getSeconds).to(cookieSerializer::setCookieMaxAge);
    map.from(cookie::getSameSite).as(SameSite::attributeValue).to(cookieSerializer::setSameSite);
    cookieSerializerCustomizers.orderedStream().forEach((customizer) -> customizer.customize(cookieSerializer));
    return cookieSerializer;
}
```

OrangeDog:
[...]

PR #28784, opened 2021-11-23
server.servlet.session.cookie.same-site isn't applied to Spring Session's SESSION cookie

vpavic:
This commit adds the mapping of `server.servlet.session.cookie.same-site` configuration property to `DefaultCookieSerializer` bean configured in the Spring Session auto-configuration.

See gh-28772


snicoll:
Thanks again @vpavic

PR #28787, opened 2021-11-23
WebFluxTest include user defined security configuration

### LINKED ISSUE #16088: Document how to structure configurations so that @Bean methods are included in slice tests
Hi!

I just updated several services to **SpringBoot 2.1.3** and observed a different behavior regarding integration tests in  **WebFlux** and **WebMVC** in combination with **spring-security**.

In both services a custom SecurityConfig exists. 

In the WebMVC service the config looks like:

```java
@Configuration
@EnableWebSecurity
public class WebSecurityConfig extends WebSecurityConfigurerAdapter {

    @Override
    protected void configure(HttpSecurity http) throws Exception {
      // custom config is here
    }
}
```
In the WebFlux service the config looks like:

```java
@Configuration
@EnableWebFluxSecurity
public class SecurityConfiguration {

    @Bean
    public SecurityWebFilterChain springWebFilterChain(ServerHttpSecurity http) {
      //custom config is here
    }
}
```

When implementing an integration test using `@WebMvcTest` the custom WebSecurityConfig is called (placing a breakpoint in the configure method works). But in order to be active it is necessary to call`apply(SecurityMockMvcConfigurers.springSecurity())` on the `MockMvcBuilder`.
I think this is in line with the documentation provided here:
https://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#test-mockmvc-setup

When implementing an integration test using `@WebFluxTest` the custom SecurityConfiguration is **not called**, but instead the standard `org.springframework.security.config.annotation.web.reactive.WebFluxSecurityConfiguration` is used. I have to use `@Import` on the test class to get the custom SecurityConfiguration to work.

I couldn't find anything in the documentation that describes this behavior of `@WebFluxTest`. I would expect that `@WebFluxTest` behaves like `@WebMvcTest` and uses the custom SecurityConfiguration.

RazorNd:
Fix #16088
In order to `WebFluxTypeExcludeFilter` include user defined security configuration, I check that the class annotated with `@ Configuration` has a method with` @ Bean` and return `SecurityWebFilterChain`. I understand that someone might call this solution a dirty trick. But it seems to me that this is the only possible solution that does not introduce a large number of architectural changes.

I also added a smoke test. But not sure if I did it in the right place. If you see fit, I will change or remove this test.

### DIFF
--- a/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTypeExcludeFilter.java
+++ b/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/web/reactive/WebFluxTypeExcludeFilter.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.test.autoconfigure.web.reactive;
 
+import java.io.IOException;
 import java.util.Arrays;
 import java.util.Collections;
 import java.util.LinkedHashSet;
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for the PR, @RazorNd. It's quite an ingenious solution, but unfortunately we don't think we can merge it as it may result in other, unwanted beans being defined in the application.

The unwanted beans could come from other `@Bean` methods on the same `@Configuration` class or from other `@Configuration` classes that are `@Import`ed directly or via an `@Enable…` annotation. We considered only including a `@Configuration` class when all it does is define a security web filter chain, but determining that is the case would require the filter to have too much knowledge of `@Configuration` class parsing and processing.

Thanks anyway.

Issue #27854, opened 2021-09-02
Properties Migrator does not detect properties of Map type that are marked as deprecated


jvillarb:
If we set a Map property as deprecated, because it is replaced or removed, the Spring Boot Properties Migrator module, when scanning the application environment and printing diagnostics, does not detect these types of properties.

For example, if we have the **custom.logging.log-level** property of type **java.util.Map** defined in the the metadata file as follows:

```json
{
  "name": "custom.logging.log-level",
  "type": "java.util.Map<java.lang.String,java.lang.String>",
  "description": "",
  "deprecation": {
	"level": "error",
	"replacement": "logging.level"
  }
}
```

and set in the application.yml:

```yaml
custom:
  logging:
      log-level:
          xx.xxxx.xxxx: INFO
```

When the **spring-boot-properties-migrator** module is included and the application is run, the generated report does not show that this property has been replaced by **"logging.level"**.

The same problem can also be encountered with the **management.health.status.http-mapping** property of the **spring-boot-actuator-autoconfigure** library:

```json
{
  "name": "management.health.status.http-mapping",
  "type": "java.util.Map<java.lang.String,java.lang.Integer>",
  "sourceType": "org.springframework.boot.actuate.autoconfigure.health.HealthIndicatorProperties",
  "deprecated": true,
  "deprecation": {
	"replacement": "management.endpoint.health.status.http-mapping"
  }
}
```
```yaml
spring:
  http:
    encoding:
      charset: UTF-8

management:
  health:
    status:
      http-mapping:
        down: 500
        out_of_service: 503
        warning: 500
```

The displayed result does not show the deprecated property **management.health.status.http-mapping**.

```
[WARN ] 2021-09-02 16:35:49.601 [main]     PropertiesMigrationListener - 
The use of configuration keys that have been renamed was found in the environment:

Property source 'applicationConfig: [classpath:/config/application.yml]':
	Key: spring.http.encoding.charset
		Line: 9
		Replacement: server.servlet.encoding.charset


Each configuration key has been temporarily mapped to its replacement for your convenience. To silence this warning, please update your configuration to use the new keys.
```

dominiccroce:
[...]

PR #28796, opened 2021-11-24
Spring Boot Properties Migrator - Detect properties of Map type that are marked as deprecated

dominiccroce:
Addresses https://github.com/spring-projects/spring-boot/issues/27854 with unit test demonstrating desired functionality

### DIFF
--- a/spring-boot-project/spring-boot-properties-migrator/src/main/java/org/springframework/boot/context/properties/migrator/PropertiesMigrationReporter.java
+++ b/spring-boot-project/spring-boot-properties-migrator/src/main/java/org/springframework/boot/context/properties/migrator/PropertiesMigrationReporter.java
@@ -25,6 +25,7 @@ import java.util.stream.Collectors;
 
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataProperty;
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataRepository;
+import org.springframework.boot.configurationmetadata.Deprecation;
 import org.springframework.boot.context.properties.source.ConfigurationProperty;
 import org.springframework.boot.context.properties.source.ConfigurationPropertyName;
 import org.springframework.boot.context.properties.source.ConfigurationPropertySource;
@@ -32,6 +33,7 @@ import org.springframework.boot.context.properties.source.ConfigurationPropertyS
 import org.springframework.boot.env.OriginTrackedMapPropertySource;
 import org.springframework.boot.origin.OriginTrackedValue;
 import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.MapPropertySource;
 import org.springframework.core.env.PropertySource;
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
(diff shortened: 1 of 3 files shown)


dominiccroce:
Running into  
```
What went wrong:
Execution failed for task ':spring-boot-project:spring-boot-properties-migrator:checkFormatTest'.
> Formatting violations found in the following files:
   * src/test/java/org/springframework/boot/context/properties/migrator/PropertiesMigrationReporterTests.java
  
  Run `format` to fix.
```
I've run ./gradlew build locally and get
```
 BUILD SUCCESSFUL in 41s
233 actionable tasks: 230 executed, 3 up-to-date

A build scan was not published as you have not authenticated with server 'ge.spring.io'.
```

Any ideas on how to determine what is wrong with my format?

spring-boot-project/spring-boot-properties-migrator/src/main/java/org/springframework/boot/context/properties/migrator/PropertiesMigrationReporter.java
```diff
@@ -105,10 +107,53 @@ PropertiesMigrationReport getReport() {
...
 						new PropertyMigration(configurationProperty, metadata, determineReplacementMetadata(metadata)));
 			}
+			getMatchingPropertiesForMapEntries(name, source, metadata, result);
 		}));
 		return result;
 	}
 
+	private void getMatchingPropertiesForMapEntries(String name, ConfigurationPropertySource source,
+			ConfigurationMetadataProperty metadata, MultiValueMap<String, PropertyMigration> result) {
+		if (source.getUnderlyingSource() instanceof MapPropertySource && name.equals(metadata.getName())) {
```
snicoll, line 117: I don't understand what this is doing. `MapPropertySource` is an implementation detail and you shouldn't be looking at it. It has nothing to do with a deprecated property that is of type `Map`.

spring-boot-project/spring-boot-properties-migrator/src/test/java/org/springframework/boot/context/properties/migrator/PropertiesMigrationReporterTests.java
```diff
@@ -144,6 +144,31 @@ void invalidReplacementHandled() throws IOException {
...
 	}
 
+	@Test
+	void detectDeprecatedPropertyInMap() {
+		MutablePropertySources propertySources = this.environment.getPropertySources();
+		Map<String, Object> content = new LinkedHashMap<>();
+		content.put("status.http-mapping.down", 500);
+		content.put("status.http-mapping.out-of-service", 503);
+		content.put("status.http-mapping.warning", 500);
+		propertySources.addFirst(new MapPropertySource("status.http-mapping", content));
```
snicoll, line 154: If I change the name of the property source to be something else than this, the test fails. The `PropertySource` implementation and its name are irrelevant for this, it could be anything.

snicoll (changes requested):
Thanks for the PR but it doesn't address the issue.

It looks like there is some confusion between `MapPropertySource` (an implementation detail, irrelevant to this issue) and a property that happens to be of type map.

I think one way of fixing this would be to patch `getMatchingProperties`. Right now, it looks for an exact match but we could also look for properties that would match the prefix if the property is deprecated.

Would you be willing to give it another try? No problem if you can't, I can take over.

spring-boot-project/spring-boot-properties-migrator/src/main/java/org/springframework/boot/context/properties/migrator/PropertiesMigrationReporter.java
dominiccroce: I see, I thought MapPropertySource indicated the Map deprecation. But if we don't have a way to filter for Map deprecations than agreed it must be done by prefix matching on the name itself.

spring-boot-project/spring-boot-properties-migrator/src/test/java/org/springframework/boot/context/properties/migrator/PropertiesMigrationReporterTests.java
dominiccroce: Nothing is name specific, but yes it is specific to MapPropertySource

snicoll: It is. Replace `new MapPropertySource("status.http-mapping", content)` by `new MapPropertySource("test", content)` and you'll see. The name of the property source has nothing to do whatsoever with the prefix used by the keys.

dominiccroce:
> Thanks for the PR but it doesn't address the issue.
> 
> It looks like there is some confusion between `MapPropertySource` (an implementation detail, irrelevant to this issue) and a property that happens to be of type map.
> 
> I think one way of fixing this would be to patch `getMatchingProperties`. Right now, it looks for an exact match but we could also look for properties that would match the prefix if the property is deprecated.
> 
> Would you be willing to give it another try? No problem if you can't, I can take over.

I have pushed a new commit to address the issues. The main problem I am facing is for any PropertyResource, I want to be able to iterate through the ConfigurationPropertyNames. Once we have those, I rely on existing code to handle prefix matching. From there I handle prefix matching. I am lacking some context on PropertySource but feels like there should be a way to get all PropertyNames without casting to IterableConfigurationPropertySource.

spring-boot-project/spring-boot-properties-migrator/src/main/java/org/springframework/boot/context/properties/migrator/PropertiesMigrationReporter.java
```diff
@@ -105,10 +107,50 @@ PropertiesMigrationReport getReport() {
...
+			getMatchingPropertiesForMapEntries(name, source, metadata, result);
 		}));
 		return result;
 	}
 
+	private void getMatchingPropertiesForMapEntries(String name, ConfigurationPropertySource source,
+			ConfigurationMetadataProperty metadata, MultiValueMap<String, PropertyMigration> result) {
+		if (source instanceof IterableConfigurationPropertySource)
+			((IterableConfigurationPropertySource) source).iterator().forEachRemaining(configurationPropertyName -> {
+				ConfigurationProperty configurationProperty = source.getConfigurationProperty(configurationPropertyName);
```
dominiccroce, line 119: This should be easier, something like source.getConfigurationProperties() would be nice API to add. Not sure if it belongs on PropertySource though

snicoll: How about using `filter`?

snicoll: @dominiccroce do you want to get another look at it? `filter` takes a predicate and should lets you filter the propertysource with only the keys that are relevant. I haven't tried but I think it's worth exploring.

snicoll:
@dominiccroce thanks but I think we will have to restart this one from scratch I am afraid. The last changes are breaking existing tests and we should really consider looking at `filter`.

PR #28801, opened 2021-11-24
Allow injection of Spring-managed beans into Hazelcast components

### LINKED ISSUE #19487: Create Hazelcast config with SpringManagedContext when available
Hazelcast instances created by Spring Boot do not have the SpringManagedContext set. 
[`SpringManagedContext`](https://docs.hazelcast.org/docs/3.12.5/manual/html-single/index.html#enabling-springaware-objects) is used to inject dependencies into for example [EntryProcessors](https://docs.hazelcast.org/docs/3.12.5/manual/html-single/index.html#entry-processor) executed by Hazelcast.

The `SpringManagedContext` is created when `HazelcastInstance` bean is created via [Hazelcast namespace support](https://docs.hazelcast.org/docs/3.12.5/manual/html-single/index.html#configuring-within-spring-context), but obviously the XML configuration is no longer the prefer way to configure Spring

I would be happy to send a pull request if there is an interest to add this to Spring Boot.

jerrinot:
Fixes #19487

When Spring Boot creates Hazelcast configuration and the Hazelcast-Spring
integration module is available on a classpath then inject SpringManagedContext
into Hazelcast configuration. 

This change makes it possible to inject Spring managed beans into objects
instantiated by Hazelcast. 

I'm not too familiar with Spring Boot so chances are the code is not idiomatic. It looks rather complex for what is basically an if-else condition. Any guidance is very much appreciated.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
@@ -105,6 +105,7 @@ class HazelcastServerConfiguration {
 
 	}
 
+	@FunctionalInterface
 	private interface ConfigurationCustomizer {
 		ConfigurationCustomizer EMPTY_CUSTOMIZER = c -> c;
 		Config customize(Config configuration);
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
@@ -25,8 +25,11 @@ import com.hazelcast.config.YamlConfigBuilder;
 import com.hazelcast.core.Hazelcast;
 import com.hazelcast.core.HazelcastInstance;
 
+import com.hazelcast.spring.context.SpringManagedContext;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnSingleCandidate;
+import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
 import org.springframework.context.annotation.Configuration;
@@ -60,10 +63,12 @@ class HazelcastServerConfiguration {
 	static class HazelcastServerConfigFileConfiguration {
 
 		@Bean
-		HazelcastInstance hazelcastInstance(HazelcastProperties properties, ResourceLoader resourceLoader)
+		HazelcastInstance hazelcastInstance(HazelcastProperties properties, ResourceLoader resourceLoader,
+											ConfigurationCustomizer customizer)
 				throws IOException {
 			Resource configLocation = properties.resolveConfigLocation();
 			Config config = (configLocation != null) ? loadConfig(configLocation) : Config.load();
+			config = customizer.customize(config);
 			config.setClassLoader(resourceLoader.getClassLoader());
 			return getHazelcastInstance(config);
 		}
@@ -93,12 +98,41 @@ class HazelcastServerConfiguration {
 	@Configuration(proxyBeanMethods = false)
 	@ConditionalOnSingleCandidate(Config.class)
 	static class HazelcastServerConfigConfiguration {
-
 		@Bean
 		HazelcastInstance hazelcastInstance(Config config) {
 			return getHazelcastInstance(config);
 		}
 
+	}
+
+	private interface ConfigurationCustomizer {
+		ConfigurationCustomizer EMPTY_CUSTOMIZER = c -> c;
+		Config customize(Config configuration);
+	}
+
+	@Configuration(proxyBeanMethods = false)
+	static class HazelcastConfigCustomizerConfiguration {
+		private final ApplicationContext applicationContext;
+
+		public HazelcastConfigCustomizerConfiguration(ApplicationContext applicationContext) {
+			this.applicationContext = applicationContext;
+		}
+
+		@Bean
+		@ConditionalOnClass(name = "com.hazelcast.spring.context.SpringManagedContext")
+		public ConfigurationCustomizer springManagedContextConfigurationCustomizer() {
+			return configuration -> {
+				SpringManagedContext springManagedContext = new SpringManagedContext(applicationContext);
+				configuration.setManagedContext(springManagedContext);
+				return configuration;
+			};
+		}
+
+		@Bean
+		@ConditionalOnMissingBean(ConfigurationCustomizer.class)
+		public ConfigurationCustomizer emptyConfigCustomizer() {
+			return ConfigurationCustomizer.EMPTY_CUSTOMIZER;
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


jerrinot:
the same change should be probably done in the client auto-configuration.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
```diff
@@ -93,14 +98,43 @@ private static Config loadConfig(URL configUrl) throws IOException {
...
+		public ConfigurationCustomizer springManagedContextConfigurationCustomizer() {
+			return configuration -> {
+				SpringManagedContext springManagedContext = new SpringManagedContext(applicationContext);
+				configuration.setManagedContext(springManagedContext);
+				return configuration;
+			};
+		}
+
+		@Bean
+		@ConditionalOnMissingBean(ConfigurationCustomizer.class)
```
jerrinot, line 132: I'm not sure if this is the right way to introduce a fallback. 

It appears it's dependant on the ordering of methods in a class. If I move this method above `springManagedContextConfigurationCustomizer()` then both beans are constructed. So I reckon there is a less fragile way, any hint is appreciated.

jerrinot:
I was also considering not to introduce the concept of `Customizers` and keep it simple. Something along this lines:
````java
try {
  var context = new SpringManagedContext(appContext);
  config.setManagedContext(context);
catch (ClassNotFoundError err) {
 // the Hazelcast-Spring integration JAR not available on a classpath, that's OK
}
```
but this felt dirty.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
```diff
@@ -60,10 +63,12 @@ private static HazelcastInstance getHazelcastInstance(Config config) {
 	static class HazelcastServerConfigFileConfiguration {
 
 		@Bean
-		HazelcastInstance hazelcastInstance(HazelcastProperties properties, ResourceLoader resourceLoader)
+		HazelcastInstance hazelcastInstance(HazelcastProperties properties, ResourceLoader resourceLoader,
+											ConfigurationCustomizer customizer)
```
ldziedziul, line 67: Couldn't we use `List<ConfigurationCustomizer> customizers` to avoid the need of `EMPTY_CUSTOMIZER` and at the same time support multiple customizers?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
```diff
@@ -93,14 +98,43 @@ private static Config loadConfig(URL configUrl) throws IOException {
...
 	static class HazelcastServerConfigConfiguration {
-
 		@Bean
 		HazelcastInstance hazelcastInstance(Config config) {
 			return getHazelcastInstance(config);
 		}
 
 	}
 
+	private interface ConfigurationCustomizer {
```
ldziedziul, line 108: Why `private`? We don't want to expose customizers functionality to users or it's just the first phase?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
ldziedziul: I guess the trick is to use a separate configuration class, take a look: https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceAutoConfiguration.java#L66-L72

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
ldziedziul, line 108:
```suggestion
	@FunctionalInterface
	private interface ConfigurationCustomizer {
```

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
wilkinsona: This is a good suggestion, thank you. I'm not sure at this point that we need to support multiple beans, but avoiding the need for `EMPTY_CUSTOMIZER` is a good goal. Our style is the achieve that via Spring Framework's `ObjectProvider` and its `getIfAvailable()` method.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
wilkinsona: Until we have a concrete need for other customisers, I'd prefer to minimise the surface area of the public API.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
jerrinot: Good suggestions, thanks to both of you! 

I pushed an update with ObjectProvider and indeed it simplified the code quite a bit. 

I also made the customizer to return `null`. It makes it less flexible, but that's not a concern here. It led to further code simplification and it's also aligned with most of the other customizers I saw in the Spring codebase.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
jerrinot: Thanks for this suggestion, I was considering it, but it looked like an overkill to me. Too much ceremony, for my taste anyway. 

Given the other changes `EMPTY_CUSTOMIZER` is not needed anymore the ordering of bean factory methods is not a concern anymore.

jerrinot:
@ldziedziul, @wilkinsona: thanks to both of you for the review. I pushed a change simplifying the code and I'm not quite happy with how it looks. Please let me know what you think.

ldziedziul (commented):
Looks good :+1:

jerrinot:
Hi @wilkinsona,

may I kindly ask you what are the next steps with this PR? Is there anything else I could do to help merge it? I see you added a flag "merge-with-amendments". What are these amendments? 

Thanks a lot!

wilkinsona:
Thanks for asking, @jerrinot. There's nothing more for you to do. The amendments are just cosmetic changes to align with the project's code style. We'll take care of those as part of merging the changes. Work on 2.7 is only just getting started (we're in a planning phase at the moment) with the first milestone not due until 20 January.

jerrinot:
@wilkinsona thank you, that's very good to know!

jerrinot:
@snicoll many thanks for taking this over the finish line. 
Feel free to reach out to [me](https://twitter.com/jerrinot) if you need any help with anything Hazelcast-related!

PR #28831, opened 2021-11-28
Fix output of "spring --version" in reference documentation

mikrethor:
Line 100 should be the same line 110. 
The result of spring --version is : Spring CLI v2.X.X and not Spring Boot v2.x.x.


snicoll:
@mikrethor good catch and thank you for making your first contribution to Spring Boot.

mikrethor:
I would have liked to make a bigger contribution but it's a start.

PR #28835, opened 2021-11-28
Polish README.adoc

vashisthabhinav:
I have corrected the sentences grammatically in the README.adoc

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -160,7 +160,7 @@ It includes many annotations that can automatically configure a slice of your ap
 
 === spring-boot-loader
 Spring Boot Loader provides the secret sauce that allows you to build a single jar file that can be launched using `java -jar`.
-Generally, you will not need to use `spring-boot-loader` directly but work with the link:spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin[Gradle] or link:spring-boot-project/spring-boot-tools/spring-boot-maven-plugin[Maven] plugin.
+Generally, you will not need to use `spring-boot-loader` directly but work with the link:spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin[Gradle] or link:spring-boot-project/spring-boot-tools/spring-boot-maven-plugin[Maven] plugin instead.
 
 
 
--- a/README.adoc
+++ b/README.adoc
@@ -160,7 +160,7 @@ It includes many annotations that can automatically configure a slice of your ap
 
 === spring-boot-loader
 Spring Boot Loader provides the secret sauce that allows you to build a single jar file that can be launched using `java -jar`.
-Generally, you will not need to use `spring-boot-loader` directly but instead, work with the link:spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin[Gradle] or link:spring-boot-project/spring-boot-tools/spring-boot-maven-plugin[Maven] plugin.
+Generally, you will not need to use `spring-boot-loader` directly but work with the link:spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin[Gradle] or link:spring-boot-project/spring-boot-tools/spring-boot-maven-plugin[Maven] plugin.
 
 
 
--- a/README.adoc
+++ b/README.adoc
@@ -13,7 +13,7 @@ Our primary goals are:
 * Provide a radically faster and widely accessible getting started experience for all Spring development.
 * Be opinionated, but get out of the way quickly as requirements start to diverge from the defaults.
 * Provide a range of non-functional features common to large classes of projects (for example, embedded servers, security, metrics, health checks, externalized configuration).
-* No code generation and no requirement for XML configuration.
+* Absolutely no code generation and no requirement for XML configuration.
 
 
 
--- a/README.adoc
+++ b/README.adoc
@@ -6,14 +6,14 @@ Spring Boot helps you to create Spring-powered, production-grade applications an
 It takes an opinionated view of the Spring platform so that new and existing users can quickly get to the bits they need.
 
 You can use Spring Boot to create stand-alone Java applications that can be started using `java -jar` or more traditional WAR deployments.
-We also provide a command line tool that runs Spring scripts.
+We also provide a command-line tool that runs Spring scripts.
 
 Our primary goals are:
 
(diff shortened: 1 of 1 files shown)


README.adoc
```diff
@@ -114,19 +114,19 @@ Spring Boot can configure large parts of typical applications based on the conte
...
 Auto-configuration attempts to deduce which beans a user might need. For example, if `HSQLDB` is on the classpath, and the user has not configured any database connections, then they probably want an in-memory database to be defined.
-Auto-configuration will always back away as the user starts to define their own beans.
+Auto-configuration will always back away as the user starts to define their beans.
 
 
 
 === spring-boot-starters
 Starters are a set of convenient dependency descriptors that you can include in your application.
-You get a one-stop-shop for all the Spring and related technology you need without having to hunt through sample code and copy paste loads of dependency descriptors.
+You get a one-stopn shop for all the Spring and related technology you need without having to hunt through sample code and copy-paste loads of dependency descriptors.
```
snicoll, line 123: Looks like that extra `n` should be removed.

README.adoc
```diff
@@ -160,7 +160,7 @@ It includes many annotations that can automatically configure a slice of your ap
 
 === spring-boot-loader
 Spring Boot Loader provides the secret sauce that allows you to build a single jar file that can be launched using `java -jar`.
-Generally you will not need to use `spring-boot-loader` directly, but instead work with the link:spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin[Gradle] or link:spring-boot-project/spring-boot-tools/spring-boot-maven-plugin[Maven] plugin.
+Generally, you will not need to use `spring-boot-loader` directly but instead, work with the link:spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin[Gradle] or link:spring-boot-project/spring-boot-tools/spring-boot-maven-plugin[Maven] plugin.
```
snicoll, line 163: Should we have this comma after "instead"?

snicoll (commented):
Thanks for the PR. I've added a couple of comments.

README.adoc
vashisthabhinav: Yes, the extra 'n' has been removed

README.adoc
vashisthabhinav: As "instead" is an introductory phrase, it is always followed by a comma. That's why I added a comma in this sentence.

wilkinsona: If instead were an introductory adverb it would be at the start of the sentence and, in that case, it should then be followed by a comma. As used here, I disagree that a comma is necessary. I think it would be better to reword the sentence to avoid using instead in its middle:

> Generally, you will not use `spring-boot-loader` directly but work with the Gradle or Maven plugin instead.

vashisthabhinav: Ok removing "instead" as directed.

vashisthabhinav: Placed "instead" at the end of the sentence.

README.adoc
```diff
@@ -99,12 +99,12 @@ There are several modules in Spring Boot. Here is a quick overview:
 
 
 === spring-boot
-The main library providing features that support the other parts of Spring Boot. These include:
+The main library provides features that support the other parts of Spring Boot. These include:
```
wilkinsona, line 102: For consistency, I don't think should be changed.

README.adoc
```diff
@@ -6,14 +6,14 @@ Spring Boot helps you to create Spring-powered, production-grade applications an
...
-We also provide a command line tool that runs Spring scripts.
+We also provide a command-line tool that runs Spring scripts.
 
 Our primary goals are:
 
 * Provide a radically faster and widely accessible getting started experience for all Spring development.
 * Be opinionated, but get out of the way quickly as requirements start to diverge from the defaults.
 * Provide a range of non-functional features common to large classes of projects (for example, embedded servers, security, metrics, health checks, externalized configuration).
-* Absolutely no code generation and no requirement for XML configuration.
+* No code generation and no requirement for XML configuration.
```
wilkinsona, line 16: I'm not sure why absolutely is being removed here.

README.adoc
```diff
@@ -114,19 +114,19 @@ Spring Boot can configure large parts of typical applications based on the conte
 A single `@EnableAutoConfiguration` annotation triggers auto-configuration of the Spring context.
 
 Auto-configuration attempts to deduce which beans a user might need. For example, if `HSQLDB` is on the classpath, and the user has not configured any database connections, then they probably want an in-memory database to be defined.
-Auto-configuration will always back away as the user starts to define their own beans.
```
wilkinsona, line 117: I'm not sure why own is being removed here.

README.adoc
```diff
@@ -143,7 +143,7 @@ This module provides many endpoints, including the `HealthEndpoint`, `Environmen
 This provides auto-configuration for actuator endpoints based on the content of the classpath and a set of properties.
 For instance, if Micrometer is on the classpath, it will auto-configure the `MetricsEndpoint`.
 It contains configuration to expose endpoints over HTTP or JMX.
-Just like Spring Boot AutoConfigure, this will back away as the user starts to define their own beans.
```
wilkinsona, line 146: I'm not sure why own is being removed here.

wilkinsona (changes requested):

README.adoc
vashisthabhinav: I changed here because it appears that this sentence is using an incorrect form of the verb "providing". But for consistency, I may change it as per your guidance.

README.adoc
vashisthabhinav: "Absolutely" seemed redundant here but I am changing it back accordingly.

README.adoc
vashisthabhinav: "Own" is redundant in this case as we are already saying users to  _define their beans_, so there is no need for "own" according to me. Rest the decision is up to you.

README.adoc
vashisthabhinav: "Own" is redundant in this case as we are already saying user   _starts to define their beans_, so there is no need for "own" according to me. Rest the decision is up to you.

snicoll:
@vashisthabhinav thank you for making your first contribution to Spring Boot.

PR #28852, opened 2021-11-30
Delete temp files - 22305

Tharik67:


philwebb:
Thanks for the suggestion, but I don't think we can delete the temp file in the constructor. The content needs to be available either as a file or byte array in the `write` method. I think we'll need some bigger refactoring before we can solve this issue.

PR #28859, opened 2021-11-30
Add option to allow Spring Batch custom isolation levels

### LINKED ISSUE #28802: Add option to allow Spring Batch custom isolation levels with JPA
Currently Spring Boot overrides Spring Batch default behaviour and specifies the connection-default isolation level for creating job metadata in the database (Spring Batch default is SERIALIZABLE with an option to override it).

This can cause issues where concurrent job executions can occur on the same database (our case).

So far the recommended way to workaround this was to override BasicBatchConfigurer (Spring Boot 1.x) or [JpaBatchConfigurer](https://github.com/spring-projects/spring-boot/blob/4eed637481186b0b85bc1cf0b1e77090139477dc/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JpaBatchConfigurer.java) (Spring Boot 2.x) and modify this method:

```java
	@Override
	protected String determineIsolationLevel() {
		logger.warn("JPA does not support custom isolation levels, so locks may not be taken when launching Jobs");
		return "ISOLATION_DEFAULT";
	}
```

I propose to have a configuration property instead which can be used to disable this behaviour and allow Spring Batch to use custom isolation levels. For example, `spring.batch.jpa.allow-custom-isolation-levels` with `false` as default value which means keep the current behaviour.

Rationale: 
This behaviour was introduced in 2014 by [this commit](https://github.com/spring-projects/spring-boot/commit/147968cf837da0398fdb186e6f80d201c668ffcb#diff-2a21e0bd0059a4525994b14571bd8e3961807450051d063571db9b7e8f1c0644). Since then Spring Boot adopted HikariCP as default connection pool, which enables safe usage of custom isolation levels on top of JPA, because connections returned to HikariCP are properly cleaned up. Also, most JPA vendors allow this nowadays (see various JpaDialect implementations), however, not in all cases.

Since the connection pool can be changed to a less well-behaving one, also, not all JPA vendors support this, the default setting for the new property would be `false`. This would also ensure backwards compatibility.

Happy to contribute a PR for this enhancement request.

tiborsulyan:
Fixes #28802 

Create `spring.batch.jdbc.isolation-level-for-create` property with the following behavior:

- Default value is `null`
- If not set, Spring Batch with JPA will use ISOLATION_DEFAULT level, and if no JPA is detected, then the Spring Batch default (ISOLATION_SERIALIZABLE) is used. This is the current behavior
- If the property is set, it will be passed to Spring Batch and used as isolation level for create operations.


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BasicBatchConfigurer.java
```diff
@@ -42,13 +42,13 @@
  */
 public class BasicBatchConfigurer implements BatchConfigurer, InitializingBean {
 
-	private final BatchProperties properties;
+	protected final BatchProperties properties;
```
philwebb, line 45: Generally we prefer to keep member variables `private`. Is there a reason you changed these?

tiborsulyan: From the design of this class it looks like it it's open for extension (see `protected` methods). However, subclasses cannot access the members those protected methods operate on, even though it's passed to their constructor. I'd suggest either extend the visibility to `protected` or provide protected getters to these. Since the members are `final` I think it's safe and would improve extensibility. What do you think?

For this concrete member, the new `JpaBatchConfigurer` implementation uses the `properties` too in `determineIsolationLevel`

wilkinsona: Given that the fields don't need to be made `protected` to implement the configurable isolation level, I think they should be considered via a separate issue please. Even `properties` could remain `private` as `JpaBatchConfigurer` could capture what it needs in its constructor. This would minimise the changes to the public API.

tiborsulyan: Sorry for the delay. Sure, I'll roll this change back and will consider a separate PR for this.

snicoll:
Thanks for the follow-up @tiborsulyan. Spring Batch has added an annotation in the next major release so we'd like to create a local enum to improve user experience in 2.7.x. No need to act on it, we'll do it as part of merging it.
.

snicoll:
@tiborsulyan thank you for making your first contribution to Spring Boot.

PR #28862, opened 2021-11-30
Simplify AssertJ usage

scordio:
I've noticed some potential simplifications related to AssertJ, so here is the PR 🙂


snicoll:
Thanks @scordio, the extracting with a path is a nice one.

Issue #22924, opened 2020-08-12
Spring Boot Actuator "manifest" endpoint


philsttr:
I'd like an actuator endpoint that shows details of the jars that are included in the spring boot application.

Perhaps the endpoint id could be "manifest".

Some examples of things that could be discovered and reported for each jar:
* jar filename
* Identifying details from `META-INF/MANIFEST.MF`  (subset of main attributes. no per-entry attributes)
  * Name
  * Automatic-Module-Name
  * Specification-Title/Version
  * Implementation-Title/Version
* Identifying details from `META-INF/**/pom.properties`
  * version
  * groupId
  * artifactId

I haven't fully thought through all the intricacies, so I'm happy to discuss further.

wilkinsona:
Thanks for the suggestion. This seems, somewhat at least, similar to https://github.com/spring-projects/spring-boot/issues/1465.

philwebb:
It's also somewhat related to the existing `InfoEndpoint`/`InfoContributor` areas. I've generally preferred git info for this kind of thing. The git sha doesn't give much away if it's accidentally exposed, but you can use it to discover exactly what was in your jar at the time it was created.

philsttr:
@philwebb, Yes, I agree that given a git sha and/or app version from the info endpoint, you can do additional detective work to figure out what was packaged within an app.  i.e. look at some report (if it exists) from the app build output for the specific version, or go grab the artifact and peek inside it.

I'm looking to short-circuit that detective work by just obtaining that info from the running app instance, since it has absolute knowledge about the jars in the app.

Currently, actuator exposes information on A) the beans that are running (beans endpoint), and B) the configuration (env endpoint).  The next logical extension is to add a third dimension... C) the version of those beans.  With those three dimensions, I can have a complete view of a running instance.  And yes, as mentioned, C can currently be derived from app version from the info endpoint, but it takes some additional detective work.  I'd love if it was a bit easier.  

Thanks for considering this feature!

wilkinsona:
Via some discussions with the Maven Central folks, I was recently reminded of various efforts to define a software bill-of-materials (sbom) standard. It's a bit like a Maven bom, but with much broader scope. An sbom can be used to describe all of the components that comprise an application, for example. https://cyclonedx.org is one example of an sbom format. https://www.ntia.gov/SBOM and https://spdx.dev have also been mentioned. AIUI, CycloneDX builds upon the latter.

[...]

PR #28884, opened 2021-12-01
Add SpringInfoContributor to expose Spring Framework/Boot version on info actuator

madorb:
This PR provides an opt-in InfoContributor that exposes the Spring and SpringBoot versions on the `info` actuator endpoint. 

This is controlled by the property `management.info.spring.enabled=true`

If true it will expose the data as:

```
"spring:" {
  "springBootVersion": "x.x.x",
  "springFrameworkVersion": "y.y.y"
}
```

though, there is likely a better naming convention, open to suggestions


wilkinsona:
We should consider https://github.com/spring-projects/spring-boot/issues/22924 when evaluating this.

madorb:
@wilkinsona ah, that's an interesting ticket. i actually just recently built a similar capability - as you mention there, in my app's gradle script itself. Thus it's not based on anything spring knows or the actual contents of the generated jar, but based on what gradle reports its runtime dependencies as. While familiar with the SBOM concept, I was unfamiliar with cyclonedx... it looks like it's also got a gradle plugin, so, could likely cover the vast majority of spring projects if it made sense to expose the generated SBOM.

philwebb:
Thanks for the PR, but I think we should roll this one into #22924. If we add a dedicated `SpringInfoContributor` now, we're likely going to want to change it when #22924 lands.

Issue #28858, opened 2021-11-30
Add configuration property for supported media types with reactive Mustache views


dsyer:
`MustacheProperties` has a `content-type` but this isn't a list and isn't used in the reactive case.

Spring Boot has support for Thymeleaf (reactive only?) configuration to set the support media types of the views as a list, so it seems like they could be more consistent.

wilkinsona:
Yes, `spring.thymeleaf.reactive.media-types` is specific to reactive apps.

To be consistent with Thymeleaf in this area, it looks like we're missing a `spring.mustache.servlet.content-type` property and a `spring.mustache.reactive.media-types` property.

wilkinsona:
There's already a `spring.mustache.content-type` property but it is Servlet-specific. To make that clear it should move to `spring.mustache.servlet.content-type`. This will mean that `MustacheProperties` can no longer extend `AbstractTemplateViewResolverProperties`. There are other properties `spring.mustache.*` properties that also appear to be specific to Servlet apps:

- `spring.mustache.allow-request-override`
- `spring.mustache.allow-session-override`
- `spring.mustache.cache`
- `spring.mustache.expose-request-attributes`
- `spring.mustache.expose-session-attributes`
- `spring.mustache.expose-spring-macro-helpers`

We should double-check each of these as part of separating out the reactive- and servlet-specific properties.

wilkinsona:
Rearranging the properties to make it clear which are Servlet-specific has been done in https://github.com/spring-projects/spring-boot/issues/30273.

PR #28887, opened 2021-12-02
Add configuration propertie media types (reactive) with Mustache views.

polarbear567:
For [issue 28858](https://github.com/spring-projects/spring-boot/issues/28858).
Now I just add media types, and here are some questions:
1. `content-type` is supported as `spring.mustache.content-type` now, do we need to migrate to servlet?
2. Does freemaker need the same change?
Because these problems are uncertain, the current changes may not be in the final code location.
Hi @wilkinsona , can you help to check?

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mustache/MustacheProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mustache/MustacheProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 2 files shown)


wilkinsona:
> content-type is supported as spring.mustache.content-type now, do we need to migrate to servlet?

Yes, it's servlet-specific so that should be made clear by the property name. This is quite a big change as it means that `MustacheProperties` can no longer be a sub-class of `AbstractTemplateViewResolverProperties`.

> Does freemaker need the same change?

No. FreeMarker is only supported with servlets so there's no need to distinguish between servlet-specific and reactive-specific properties.

wilkinsona:
Thanks for the PR here, @polarbear567. Looking more closely at this, it's become apparent that there's quite a bit more work to do to identify all of the properties that need to be moved into a reactive- or servlet-specific namespace. I think it's best if we do that before trying to make any good changes and I've labelled the issue as pending design work to reflect this. Thanks anyway.

polarbear567:
Sure, I'd like to do something after you finish the design.

PR #28911, opened 2021-12-06
Add consistent quotes in YAML samples of reference doc

### LINKED ISSUE #28709: Consistently use quotes in YAML samples
As raised in #28671 we have some YAML in the docs that doesn't quote strings. I think most of the samples do so we should probably align. Using quotes helps prevent the ["norway" problem](https://hitchdev.com/strictyaml/why/implicit-typing-removed/).

viktorardelean:
Fix #28709

- Added quotes in all YAML samples from the .adoc files where they were missing
- Replaced ' with " in one YAML sample for consistency


snicoll:
@viktorardelean thank you for making your first contribution to Spring Boot. I've found a number of additional samples that needed to be converted and I've done that in 5599e8d.

Issue #28924, opened 2021-12-08
JsonObjectSerializer.serialize - refactor condition in catch block.


Drezir:
JsonObjectSerializer.serialize - use own catch block for IOException.

snicoll:
Thanks but we don't need both a PR and an issue as the former supersedes the latter.

PR #28925, opened 2021-12-08
Replace if condition in catch block with catch for the specific excep…

Drezir:
https://github.com/spring-projects/spring-boot/issues/28924


snicoll:
Thanks foe the PR but I prefer the code as it currently is.

PR #28928, opened 2021-12-08
Upgrade Java version to 17 in spring-boot-starter-parent's pom

dramatist:
Although spring boot 3.0 has upgraded to java 17, but the parent maven pom in spring-boot-starter-parent has not been upgraded yet


wilkinsona:
@dramatist Thanks very much for making your first contribution to Spring Boot.

PR #28944, opened 2021-12-09
Add public to getter and setter for StackdriverProperties.useSemanticMetricTypes

izeye:
This PR adds `public` access modifier to getter and setter for `StackdriverProperties.useSemanticMetricTypes` as it seems to be missing accidentally.


snicoll:
Thanks again @izeye

Issue #28947, opened 2021-12-09
App fails to start when it depends on thymeleaf-extras-springsecurity5 but does not have Spring Security on the classpath


mdeinum:
Currently the auto configuration of the `SpringSecurityDialect` is solely dependend on its own existence on the classpath, however if this was accidentily added without Spring Security the application fails to start. 

There is now only an `@ConditionalOnClass({ SpringSecurityDialect.class })` on the configuration class, it might be beneficial to add an additional check for Spring Security on there. 

See https://stackoverflow.com/questions/70290743/how-to-temporary-disable-dependencies-in-spring-boots-pom-xml

wilkinsona:
Thanks, @mdeinum. Somewhat surprisingly, the Spring Security dependencies are declared as provided in `thymeleaf-extras-springsecurity5`'s pom even though they are required at runtime. Given that is the case, we should add a check for a Spring Security class too, backing off in its absence.

terminux:
Hi, I submitted a PR to fix this issue.#28967

snicoll:
Closing in favor of PR #28967

PR #28967, opened 2021-12-10
App fails to start when it depends on thymeleaf-extras-springsecurity5 but does not have Spring Security on the classpath

terminux:
Closes gh-28947

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/thymeleaf/ThymeleafAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/thymeleaf/ThymeleafAutoConfiguration.java
@@ -58,6 +58,9 @@ import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.core.Ordered;
+import org.springframework.security.core.context.ReactiveSecurityContextHolder;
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/thymeleaf/ThymeleafAutoConfiguration.java
```diff
@@ -279,7 +280,7 @@ DataAttributeDialect dialect() {
 	}
 
 	@Configuration(proxyBeanMethods = false)
-	@ConditionalOnClass({ SpringSecurityDialect.class })
+	@ConditionalOnClass({ SpringSecurityDialect.class, AbstractSecurityWebApplicationInitializer.class })
```
wilkinsona, line 283: I think the intent would be clearer if we checked for one of the classes that `SpringSecurityDialect` depends upon such as `org.springframework.security.web.server.csrf.CsrfToken`.

terminux: Done.

terminux:
Thanks @wilkinsona , nice suggestions. I updated the PR and it's ready for your review again.

snicoll:
@terminux thanks for making your first contribution to Spring Boot. There's no need to list all 3 classes so I've only kept what @wilkinsona suggested, see f6f5aec.

PR #28982, opened 2021-12-10
log4j 2.15.0 CVE-2021-44228

sullis:
https://logging.apache.org/log4j/2.x/security.html

https://cve.mitre.org/cgi-bin/cvename.cgi?name=CVE-2021-44228

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -757,7 +757,7 @@ bom {
 			]
 		}
 	}
-	library("Log4j2", "2.14.1") {
+	library("Log4j2", "2.15.0") {
 		group("org.apache.logging.log4j") {
 			modules = [
 				"log4j-to-slf4j"


snicoll:
Thanks for the PR but, as stated in the [PR template](https://raw.githubusercontent.com/spring-projects/spring-boot/main/.github/PULL_REQUEST_TEMPLATE.md), we don't accept dependency upgrades like this.

bclozel:
As noted in the issue template, we're taking care of dependencies upgrades.

jonathan-baker-disney:
@snicoll - Is there a plan to resolve this CVE quickly, or will you just change the version with the next release?  I know we can all override the property inside of our projects individually, but I am curious if Spring would release a quick patch just for this.

snicoll:
The CVE is not ours so I am not sure what you're asking us. Spring Boot is not affected by default, only users that are opting-in for log4j2 are. I don't think we would have released a new version but it turns out that we have a round of releases next Thursday that will include it.

jonathan-baker-disney:
@snicoll - Thank you for the quick reply.  I know the CVE wasn't yours, but I thought since you included the compromised version that you might have been putting out a version of spring-boot with the fixed log4j.  I did forget that logback was the default, so that makes sense that you aren't doing a quick fix.  I just wanted to get a feel for when a spring-boot fix would be released.  Thank you again.

rsousa:
As it can help others: https://spring.io/blog/2021/12/10/log4j2-vulnerability-and-spring-boot

PR #28999, opened 2021-12-12
Fix a typo in "Packaging Executable and Plain Archives" section

fml2:


snicoll:
@fml2 thanks for the PR but it became apparent that the whole sentence should be rephrased. I've done that in 496cd3d2bad.

Issue #28918, opened 2021-12-07
Include AbstractJdbcConfiguration beans in @DataJdbcTest


wilkinsona:
When using Spring Data JDBC, it's fairly common to provide a bean that's an `AbstractJdbcConfiguration` sub-class. For example, such a sub-class is used to register custom converters used during reading and writing of entities. At the moment, such beans are filtered out by `@DataJdbcTest`.

They can be included using a custom filter:

```java
@DataJdbcTest(includeFilters = @ComponentScan.Filter(type = FilterType.ASSIGNABLE_TYPE,
        classes = AbstractJdbcConfiguration.class))
```

It would be good if this include filter wasn't necessary and `AbstractJdbcConfiguration` beans were included automatically.

raqueru:
Hello! I'm new to this project and I'd like to understand better where can i find the solution for this issue

wilkinsona:
I'd start by creating a new test class named `DataJdbcTypeExcludeFilterTests`, and writing a test that asserts that 
 [`DataJdbcTypeExcludeFilter`](https://github.com/spring-projects/spring-boot/blob/2.7.x/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/data/jdbc/DataJdbcTypeExcludeFilter.java) does not exclude an `AbstractJdbcConfiguration` sub-class. It should fail to begin with.

With the test in place, `DataJdbcTypeExcludeFilter` can then be changed to include `AbstractJdbcConfiguration`. To do so, its `getDefaultIncludes()` method should be overridden to return a `Set` that contains `AbstractJdbcConfiguration.class`. With this change in place, the new test that was failing should now pass.

raviu:
@wilkinsona I've raised a PR for this enhancement per your instructions. Hoping to get my feet wet with the project and this seemed a good place to start! @raqueru apologies in case you're still working on this one.

snicoll:
@raqueru were you working on this issue or are you still willing to contribute something? 

@raviu thanks for the PR but I'd have preferred to check first before submitting. I agree it is a bit blurry whether @raqueru is working on it or not but we prefer to give contributors a chance to sort things out before someone else submits something.

Let's see what @raqueru says and we'll proceed accordingly.

raqueru:
> No problem! I was trying to solve it but if Raviu already did it, you can proceed

mbhave:
Closing in favor of PR #29003.

[...]

PR #29003, opened 2021-12-13
Include AbstractJdbcConfiguration beans in @DataJdbcTest

raviu:
Addresses enhancement issue https://github.com/spring-projects/spring-boot/issues/28918

Adds `AbstractJdbcConfiguration` to default includes.


snicoll:
@raviu thank you for making your first contribution to Spring Boot.

snicoll:
slice tests annotation can't bring a dependency when they're used. The issue addressed an inconsistency with regards to `@DataJdbcTest`. If you want to test the JDBC slice of your app and you don't want to use Spring Data, you have to use `@JdbcTest`.

PR #29007, opened 2021-12-14
Add tip for testing OutputCapture with gradle

quaff:


philwebb:
@quaff Do you have a sample project you could attach that shows a test failing when `showStandardStreams` is `false`. It would be nice if we could still capture output even if Gradle is capturing the streams.

quaff:
I forgot add `follow="true"` in `log4j2-test.xml`, close it.

PR #29019, opened 2021-12-14
Use log4j-bom to manage log4j-to-slf4j dependency

copbint:
log4j-to-slf4j is already in [log4j-bom](https://github.com/apache/logging-log4j2/blob/master/log4j-bom/pom.xml#L151) since this [commit](https://github.com/apache/logging-log4j2/commit/9b0345931e978633b80d24c37610cd021d719ed4), so we should remove it to keep clear。


snicoll:
@copbint thank you for making your first contribution to Spring Boot.

Issue #25945, opened 2021-04-09
A functional command line runner


xenoterracide:
So in order to do a command line runner with an exit code you have to do mutable state (at least from what I can find documented), here's [picocli's example](https://github.com/remkop/picocli/tree/master/picocli-spring-boot-starter)

```java
import picocli.CommandLine;
import picocli.CommandLine.IFactory;

import org.springframework.boot.CommandLineRunner;
import org.springframework.boot.ExitCodeGenerator;
import org.springframework.stereotype.Component;

@Component
public class MyApplicationRunner implements CommandLineRunner, ExitCodeGenerator {

	private final MyCommand myCommand;

	private final IFactory factory; // auto-configured to inject PicocliSpringFactory

	private int exitCode;

	public MyApplicationRunner(MyCommand myCommand, IFactory factory) {
		this.myCommand = myCommand;
		this.factory = factory;
	}

	@Override
	public void run(String... args) throws Exception {
		exitCode = new CommandLine(myCommand, factory).execute(args);
	}

	@Override
	public int getExitCode() {
		return exitCode;
	}
}
```

and from what little is in the reference documentation this s true, I think we need an interface that can work without mutability.

```java
interface ExitingCommandLineRunner {

   int run( String... args );
}
```

philwebb:
The exit code logic could certainly do with a refresh. We have another issue (#25100) related to the `SpringApplication.exit` method.

One trick that might work in the meantime is to put `ExitCodeGenerator` on an exception:

```java
@Component
public class MyApplicationRunner implements CommandLineRunner {

	private final MyCommand myCommand;

	private final IFactory factory; // auto-configured to inject PicocliSpringFactory

	public MyApplicationRunner(MyCommand myCommand, IFactory factory) {
		this.myCommand = myCommand;
		this.factory = factory;
	}

	@Override
	public void run(String... args) throws Exception {
		int exitCode = new CommandLine(myCommand, factory).execute(args);
		if (exitCode != 0) {
			throw new MyExitException(exitCode);
		}
	}

	static class ExitException implements ExitCodeGenerator {

		private final int code;

		ExitException(int code) {
			this.code = code;
		}

		@Override
		public int getExitCode() {
			return this.code;
		}

	}

}
```

raqueru:
Hello, i would like to solve this issue

raqueru:
Should this interface be implemented only by classes that use exit code or in general? I can't seem to find any example of run codes that use exit code in general

[...]

PR #29024, opened 2021-12-14
Created classes ExitingCommandLineRunner and ExitException that deals…

raqueru:
Addresses enhancement issue #25945

creates classes ExitingCommandLineRunner and ExitException

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ExitException.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ExitException.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


philwebb:
Thanks for the pull-request @raqueru, but I think we want to look at the exit code logic in general rather than adding new public APIs. Your code will be a handy reference for anyone looking into doing this themselves, but I don't think we want to include it into Spring Boot itself.

raqueru:
> Thanks for the pull-request @raqueru, but I think we want to look at the exit code logic in general rather than adding new public APIs. Your code will be a handy reference for anyone looking into doing this themselves, but I don't think we want to include it into Spring Boot itself.

Sure! No problem, do you have any tips to solve the main problem so i can try it again?

philwebb:
@raqueru I'm afraid I don't have any good suggestions just yet. It's likely to take a bit of time to investigate, and since there's a workaround it's not a particularly high priority.

Issue #25100, opened 2021-02-04
Provide SpringApplication.exit alternative


philwebb:
The `SpringApplication.exit` method seems quite hard to read. We currently document that it can be used as follows:

```java
System.exit(SpringApplication.exit(SpringApplication.run(MyExitCodeApplication.class, args)));
```

It feels like a `runAndExit` method might be helpful:

```java
SpringApplication.runAndExit(MyExitCodeApplication.class, args);
```

wilkinsona:
What should do about the exit code generators? Both run and exit have varargs variants:

```
ConfigurableApplicationContext run(Class<?> primarySource, String... args)
int exit(ApplicationContext context, ExitCodeGenerator... exitCodeGenerators)
```

I suspect having additional generators (beyond those that are beans) is unusual so we could have something like this:

```
int runAndExit(Class<?> primarySource, String... args)
```

Then perhaps we could have one or more of these as well:

```
int runAndExit(Class<?> primarySource, ExitCodeGenerators exitCodeGenerator, String... args)
int runAndExit(Class<?> primarySource, Collection<ExitCodeGenerator> exitCodeGenerators, String... args)
```

PR #29081, opened 2021-12-16
creation of function run and exit and adjusting all uses of it in the…

raqueru:
Addresses enhancement issue #25100

create function run and exit using a Inline Method refactoring.
Changed all uses from exit to the new function


snicoll:
@raqueru thanks for the PR but the related issue is flagged "on-hold" as we still need to figure out how we'd like this to be fixed.

Issue #22523, opened 2020-07-23
Remove spring.profiles support


philwebb:
The `spring.profiles` property was replaced in Spring Boot 2.4 with `spring.config.activate.on-profile`. We currently still support the old property and we should consider when to remove it.

shawnz:
What would be the intended way of setting the active profiles if this property were removed? My understanding is that `spring.config.activate.on-profile` doesn't actually activate the profile itself.

mbhave:
@shawnz that's correct. `spring.profiles.active` is not deprecated and can be used to activate profiles.

shawnz:
Understood, thanks!

raqueru:
In this case we should simply remove every mention of spring.profiles.include or change it for other thing?

philwebb:
This issue is only about removing the `spring.profiles` property. The `spring.profiles.include` and `spring.profiles.active` properties will remain.

raqueru:
> This issue is only about removing the `spring.profiles` property. The `spring.profiles.include` and `spring.profiles.active` properties will remain.

got it! thanks

philwebb:
Moving this to 3.0.x and flagging for team attention because I'm not sure we should drop `ConfigFileApplicationListener` before then.

wilkinsona:
> As far as I can see, the documentation for SpringBoot 3.0.0-M2 is still using this property

That is to be expected as this issue is still open and we have not yet removed support for `spring.profiles`.

PR #29082, opened 2021-12-16
 Remove spring.profiles support

raqueru:
Addresses issue #22523

removed all instances of spring profiles, changing it for spring.config.activate.on-profile


philwebb:
Thanks, but I'm afraid we probably need to take a different approach here. I think we'll want to remove `LEGACY_PROFILES_NAME` entirely rather than change it to `spring.config.activate.on-profile`.

raqueru:
> Thanks, but I'm afraid we probably need to take a different approach here. I think we'll want to remove `LEGACY_PROFILES_NAME` entirely rather than change it to `spring.config.activate.on-profile`.

Ah yes, sorry for the misunderstood, I may try changing it then

philwebb:
Thanks @raqueru. You might want to wait until the new year, I've flagged the original issue so we can discuss as a team exactly what we want to do.

PR #29084, opened 2021-12-16
Validate the aud claim in OAuth2 resource server

### LINKED ISSUE #28427: Validate the aud claim in OAuth2 resource server
# Background

According to [this doc], resource-server must validate `iss` and `aud`.

# Current situation

### iss
1. Congiguration: Now `iss` will be configured by `spring.security.oauth2.resourceserver.jwt.issuer-uri`, 
2. Validation logic:
https://github.com/spring-projects/spring-boot/blob/30c27274e228e1a8995a7c4750b730232fae9a58/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/servlet/OAuth2ResourceServerJwtConfiguration.java#L68-L76

### aud
No related logic

# Requirement
Just like `iss`, we should validate `aud`. And add a property like `spring.security.oauth2.resourceserver.jwt.audience`.

[this doc]: https://tools.ietf.org/id/draft-bertocci-oauth-access-token-jwt-00.html#rfc.section.4

ahmedmq:
Configure a new optional property for audience in resource server. Validate the audience claim in JWT if the audience property is provided.

Closes  #28427

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/OAuth2ResourceServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/OAuth2ResourceServerProperties.java
@@ -31,6 +31,7 @@ import org.springframework.util.StreamUtils;
  *
  * @author Madhura Bhave
  * @author Artsiom Yudovin
+ * @author Mushtaq Ahmed
  * @since 2.1.0
  */
 @ConfigurationProperties(prefix = "spring.security.oauth2.resourceserver")
@@ -71,6 +72,11 @@ public class OAuth2ResourceServerProperties {
 		 */
 		private Resource publicKeyLocation;
 
+		/**
+		 * Identifies the recipients that the JWT is intended for.
+		 */
+		private String audience;
+
 		public String getJwkSetUri() {
 			return this.jwkSetUri;
 		}
@@ -101,6 +107,14 @@ public class OAuth2ResourceServerProperties {
 
 		public void setPublicKeyLocation(Resource publicKeyLocation) {
 			this.publicKeyLocation = publicKeyLocation;
+		}
+
+		public String getAudience() {
+			return this.audience;
+		}
+
+		public void setAudience(String audience) {
+			this.audience = audience;
 		}
 
 		public String readPublicKey() throws IOException {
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/OAuth2ResourceServerProperties.java
```diff
@@ -71,6 +72,11 @@ public Opaquetoken getOpaquetoken() {
 		 */
 		private Resource publicKeyLocation;
 
+		/**
+		 * Identifies the recipients that the JWT is intended for.
+		 */
+		private String audience;
```
chenrujun, line 78: Consider the spring-cloud-gateway scenario, I suggest to Use `List<String>` instead of `String`.
```
client -> spring-cloud-gateway -> resource-server-1
                               -> resource-server-2
                               -> etc..
```
When spring-cloud-gateway validate the access-token, the valid audience is a list of string.

chenrujun: And rename variable from `audience` to `audiences`.

wilkinsona: Thanks, @chenrujun. This makes sense to me. The validation should then use a predicate that checks that at least one audience in the JWT matches one of the configured audiences. Does that makes sense to you, @mbhave?

fstorz:
The changes in this PR only apply the audience validation if decoder is configured via "jwk-set-uri". Would it make sense to also provide this feature when decoder is configured via "issuer-uri"?

snicoll:
@ahmedmq thank you for making your first contribution to Spring Boot.

mrodal:
This feature is not mentioned in the docs, in fact audience validation is shown as an example for custom validators:

https://docs.spring.io/spring-security/reference/servlet/oauth2/resource-server/jwt.html#oauth2resourceserver-jwt-validation-custom

This feature should be in the docs

snicoll:
@mrodal thanks for the nudge. Going forward, rather than commenting on a closed PR, please create an issue to request a change in the documentation. I've done that in #34848.

Issue #22305, opened 2020-07-10
Refactor EntryWriter logic so SizeCalculatingEntryWriter can be closed


philwebb:
It would be good to refactor the `EntryWriter` interface and the `SizeCalculatingEntryWriter` to that we can delete temporary files as soon as we're done with them. See #22108 for background.

PR #29085, opened 2021-12-16
Del temp file

Tharik67:
implementation of issue #22305


wilkinsona:
Thanks for the PR, @Tharik67. I'd prefer to try and solve this in a way that doesn't expose the use of a temporary file in the public API of `EntryWriter`.

Issue #28180, opened 2021-10-02
Use switch expressions where appropriate


wilkinsona:
I don't think we make much use of switch statements, but we should see if there are places where using switch expressions will make the code more concise.

philwebb:
I tried to use one today and got:

```
> Task :spring-boot-project:spring-boot-docs:aggregatedJavadoc FAILED
/Volumes/Data/projects/spring-boot/code30x/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/AotProcessor.java:137: error: switch expressions are not supported in -source 8
                return switch (kind) {
                       ^
  (use -source 14 or higher to enable switch expressions)
/Volumes/Data/projects/spring-boot/code30x/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/AotProcessor.java:138: error: switch rules are not supported in -source 8
                case SOURCE -> this.sourceOutput;
                            ^
  (use -source 14 or higher to enable switch rules)
2 errors

```

wilkinsona:
@philwebb I think I've fixed that in https://github.com/spring-projects/spring-boot/issues/31210. You should be able to use switch expressions on main now.

dreis2211:
I think there needs some work to be done on the formatting end:

Running `:format` turns this:
```
return switch (getErrorProperties().getIncludeBindingErrors()) {
     case ALWAYS -> true;
     case ON_PARAM -> getErrorsParameter(request);
     default -> false;
};
```
into this:
```
return switch (getErrorProperties().getIncludeBindingErrors()) {
case ALWAYS -> true;
case ON_PARAM -> getErrorsParameter(request);
default -> false;
};
```

Arguably, the missing indentation makes it harder to read.

Otherwise this is ready on my local branch.

wilkinsona:
Thanks, @dreis2211. I prefer the pre-formatted version too. We can change this in Spring Java Format with this setting:

```
org.eclipse.jdt.core.formatter.indent_switchstatements_compare_to_switch=true
```

It also affects a "traditional" switch. This:

```
protected boolean isIncludeBindingErrors(HttpServletRequest request, MediaType produces) {
    switch (getErrorProperties().getIncludeBindingErrors()) {
    case ALWAYS:
        return true;
    case ON_PARAM:
        return getErrorsParameter(request);
    default:
        return false;
    }
}
```

Will become this:

```
protected boolean isIncludeBindingErrors(HttpServletRequest request, MediaType produces) {
[...]

PR #29087, opened 2021-12-16
Ref switch

Tharik67:
refactor to use the switch expression when possible to make the code more clean about the issue: https://github.com/spring-projects/spring-boot/issues/28180


philwebb:
Thanks for the pull-request, but I think some of these changes will break things. Specifically, any case without a break could run multiple code paths. I think we indented to use the new `->` syntax with #28180.

wilkinsona:
I agree with Phil. I don't think we should merge this as it'll break things in a number of places due to fall-through. I also find the code harder to read in a few places. Thanks anyway, @Tharik67.

Issue #29137, opened 2021-12-20
Make the runScripts method on AbstractScriptDatabaseInitializer more extensible


snicoll:
As revealed by fixing #29095, the configuration of the settings are done by a protected method that takes individual arguments. Adding an extra setting (for instance, comment prefixes) would break the method in its current form.

We should revisit this arrangement so that perhaps the settings (or at least something inherited from it) is used to configure the populator.

terminux:
Hi, maybe we can replace a single parameter with the corresponding `Properties`(such as `QuartzProperties`), I submitted a PR.#29162

PR #29162, opened 2021-12-24
QuartzDataSourceScriptDatabaseInitializer support more extra settings

terminux:
Closes gh-29137


snicoll:
Thanks for the PR but it doesn't address the related issues. What we'd like to do is revisit the base class so that adding new settings doesn't break the current protected runScripts method.

terminux:
Thanks @snicoll , I get it. Maybe we can move the extra settings to `DatabaseInitializationSettings`, in this way, these settings can be obtained in the `runScripts` method and assigned to `ResourceDatabasePopulator`, and the `customize` method can be removed. Do you think this is feasible?

snicoll:
That would break the current method which is what we're trying to fix, actually.

Issue #27319, opened 2021-07-14
Mutate existing default HttpMessageConverter instead of adding duplication


quaff:
I found Spring Boot add two flavored HttpMessageConverters to default messageConverters
https://github.com/spring-projects/spring-boot/blob/47516b50c39bd6ea924a1f6720ce6d4a71088651/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/http/JacksonHttpMessageConvertersConfiguration.java#L47-L54

https://github.com/spring-projects/spring-boot/blob/47516b50c39bd6ea924a1f6720ce6d4a71088651/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/http/HttpMessageConvertersAutoConfiguration.java#L77-L84

Here is combined converters by `HttpMessageConverters`
1. org.springframework.http.converter.ByteArrayHttpMessageConverter@16b050d3
2. **org.springframework.http.converter.StringHttpMessageConverter@23c51bca (inserted by Spring Boot, defaultCharset customized)**
3. org.springframework.http.converter.StringHttpMessageConverter@27be770a
4. org.springframework.http.converter.ResourceHttpMessageConverter@7c7c3afd
5. org.springframework.http.converter.ResourceRegionHttpMessageConverter@65d583ec
6. org.springframework.http.converter.xml.SourceHttpMessageConverter@7177614d
7. org.springframework.http.converter.support.AllEncompassingFormHttpMessageConverter@25a54e4e
8. **org.springframework.http.converter.json.MappingJackson2HttpMessageConverter@305bd1b8 (inserted by Spring Boot, objectMapper customized)**
9. org.springframework.http.converter.json.MappingJackson2HttpMessageConverter@381323d5
10. org.springframework.http.converter.xml.Jaxb2RootElementHttpMessageConverter@71901dd8

there are two cons:
 * duplication
 * other `AbstractJackson2HttpMessageConverter`s such as `MappingJackson2CborHttpMessageConverter` and `MappingJackson2SmileHttpMessageConverter` are not considered

wilkinsona:
Duplicate of #21374 and the various other issues to which it links.

quaff:
[...]

PR #29167, opened 2021-12-24
fix: HttpMessageConverters.getCombinedConverters

qxo:
fix defaultConverter not replaced


quaff:
Seems duplicated with https://github.com/spring-projects/spring-boot/issues/27319#issuecomment-880497942

snicoll:
Indeed, thanks @quaff

PR #29168, opened 2021-12-24
fix: BackgroundPreinitializer only init Jackson2ObjectMapperBuilder once

qxo:
AllEncompassingFormHttpMessageConverter will invoke Jackson2ObjectMapperBuilder,
so there no need invoke Jackson2ObjectMapperBuilder again


snicoll:
@qxo Thanks for the PR, but things right now are working as expected. The idea of `BackgroundPreinitializer` is that it can do so early in a separate thread so that this costly call is applied (and cached) before `AllEncompassingFormHttpMessageConverter` calls it in the main thread that starts the context.

qxo:
> @qxo Thanks for the PR, but things right now are working as expected. The idea of `BackgroundPreinitializer` is that it can do so early in a separate thread so that this costly call is applied (and cached) before `AllEncompassingFormHttpMessageConverter` calls it in the main thread that starts the context.

It's not that case!
In BackgroundPreinitializer  we already create AllEncompassingFormHttpMessageConverter, so there not create Jackson2ObjectMapperBuilder again!

snicoll:
The whole point of the background pre-init is to pre-initialize potentially costly objects so that actually calls in the main thread are fast. We could surely revisit this but what `AllEncompassingFormHttpMessageConverter` does is an implementation detail so calling the jackson converter explicitly is intended.

PR #29178, opened 2021-12-28
Enforce using BDDMockito

quaff:
1. Replace `Mockito.verify*()` with `BDDMockito.then()`
2. Replace `Mockito.doReturn()` with `BDDMockito.willReturn()`
3. Adjust checkstyle rule

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/servlet/CloudFoundrySecurityInterceptorTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/cloudfoundry/servlet/CloudFoundrySecurityInterceptorTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2012-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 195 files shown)


snicoll (changes requested):
Thanks for the PR but I don't see a reason to change calls to `verify`.

quaff:
Mixing `given` and `verify` is not elegant, pure BDD-style is better for readability.
https://github.com/spring-projects/spring-boot/blob/47516b50c39bd6ea924a1f6720ce6d4a71088651/src/checkstyle/checkstyle.xml#L49-L55

>Please use BDD-style (given, when, then) using BDDMockito imports.

snicoll:
@quaff that's a team decision I am afraid and what you've quoted doesn't include `verify` on purpose, it states:

> Please use BDD-style (given, when, then) using BDDMockito imports.

We've been there before and we have no intention to change calls to `verify` at this point. If you revert that part, we can continue reviewing this PR.

quaff:
> @quaff that's a team decision I am afraid and what you've quoted doesn't include `verify` on purpose, it states:
> 
> > Please use BDD-style (given, when, then) using BDDMockito imports.
> 
> We've been there before and we have no intention to change calls to `verify` at this point. If you revert that part, we can continue reviewing this PR.

Does the team decide to keep weird `given` and `verify` usage? I will create another PR to 
remove unnecessary explicit `Mockito.times(1)` if the answer is YES.

vpavic:
@snicoll What @quaff is trying to say is that _Please use BDD-style (given, when, then) using BDDMockito imports_ message on the Checkstyle rule is actually in conflict with what you describe as the team's preference - you don't actually use `BDDMockito#then` (as message the suggests) but rather `Mockito#verify`.

FWIW I also find this mixed usage of BDD-style and traditional Mockito APIs to be rather confusing and unintuitive.

snicoll:
Thanks both. We've discussed this as a team and decided to harmonize and use `then` rather than `verify`.

Issue #29140, opened 2021-12-20
Add Spring GraphQL support


bclozel:
Now that the Spring GraphQL project has made significant project in its Milestone phase, we should move the experimental auto-configuration and starter in this project.

We should address the following tasks:

- [x] Add dependency management for Spring GraphQL and GraphQL Java
- [x] Auto-configure Spring GraphQL base infrastructure
- [x] Add a FailureAnalyzer for InvalidSchemaLocationsException
- [x] Auto-configure HTTP transport for MVC and WebFlux
- [x] Auto-configure schema printer and graphiql endpoints
- [x] Auto-configure CORS options for GraphQL web endpoints
- [x] Auto-configure GraphQL WebSocket endpoint
- [x] Auto-configure Query* support for GraphQL
- [x] Auto-configure security for Spring GraphQL
- [x] Add testing support with `GraphQlTester` and `@GraphQlTest`
- [x] Add Metrics support
- [x] Create Spring Boot starter

See spring-projects/spring-graphql#207

PR #29194, opened 2021-12-31
Polish GraphQL changes

izeye:
This PR polishes GraphQL changes a bit.

FYI, I created this against the `2.7.x` branch as the changes have been reverted in f61c87c3a7b9d978ea7c1f7e8acb709e2623bb72 on the `main` branch for some reason.

See gh-29140


philwebb:
Thanks @izeye. The reason that we don't have GraphQL on `main` is because there's not yet a Jakarta EE compatible release of it.

Issue #29034, opened 2021-12-15
H2 version 2.0.202 compatibility with Spring Boot


meier-th:
A vulnerability (CVE-2021-23463: https://nvd.nist.gov/vuln/detail/CVE-2021-23463) was discovered in h2 database and it is said to be fixed in version 2.0.202.
However, Spring Boot 2.4.13, 2.5.7 and 2.6.1 use 1.4.200 version of h2 by default. Since the new version seems to be a major release, the question arises - is h2:2.0.202 compatible with Spring Boot 2.4.13, 2.5.7 and 2.6.1?

nathan-wanono:
Hi, 

I just tested on my current project to force the H2 version t 2.0.202 using the  following property : `<h2.version>2.0.202</h2.version>`

Actually, Spring Batch does not work with this version of H2 database : the database is never initialized.
I tryed to force it using the `spring.batch.jdbc.initialize-schema=always` property (also tryed with `embedded`).
I also tryed to do a manual initilization using the classpath scripts but they are not compatible aswell.

Spring version: 2.6.1

andifalk:
This also has issues with using Flyway DB in spring boot with detecting H2 mode correctly.
Spring Boot would need to upgrade dependency to flywaydb at least to version 8.2.2
See https://flywaydb.org/documentation/learnmore/releaseNotes#8.2.2

philwebb:
@meier-th We plan to upgrade to the latest H2 release in Spring Boot 2.7. Generally we only upgrade dependencies to their latest patch releases in a Spring Boot patch release. There is [some discussion on the H2 issue tracker about the CVE](https://github.com/h2database/h2database/issues/3195#issuecomment-994683482) and if the fix can be backported. Have you tried upgrading manually?

@nathan-wanono I don;t know if you're facing an issue in Spring Boot or Spring Batch. Could you please open a new issue and if possible provide a sample that shows initialization failing?

@andifalk We'll be upgrading to Flyway 8.2 (or newer) in Spring Boot 2.7. Have you tried overriding the flyway version in your Maven or Gradle file? I'd be interested to know if they are compatible.

philwebb:
I'll close this one for now. I don't think we can upgrade our managed dependency, but if other folks find compatibility issues that we can solve in Spring Boot 2.5 or 2.6 please comment here.

subes:
[...]

PR #29200, opened 2022-01-02
Adjust test schema DDLs for compatibility with H2 2.x

hpoettker:
This PR is related to #29034 and shall help to support H2 v2.0.x with Spring Boot. It adjusts all DDLs for tests such that the syntax is compatible not only with HSQL and H2 v1.4.x but also H2 v2.0.x.

I renamed the schema file for the Spring Batch autoconfiguration tests as H2 and not HSQL is actually used during the tests.


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java
```diff
@@ -233,7 +233,7 @@ void testRenamePrefix() {
 				.withUserConfiguration(TestConfiguration.class, EmbeddedDataSourceConfiguration.class,
 						HibernateJpaAutoConfiguration.class)
 				.withPropertyValues("spring.datasource.generate-unique-name=true",
-						"spring.batch.jdbc.schema:classpath:batch/custom-schema-hsql.sql",
+						"spring.batch.jdbc.schema:classpath:batch/custom-schema-h2.sql",
```
snicoll, line 236: Can you share why you're switching the test to use H2 instead of HSQL?  If the purpose is to test that Spring Batch works with their H2 scripts, that's something that should be done as part of a Spring Batch release.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationWithoutJpaTests.java
```diff
@@ -75,7 +75,7 @@ void jdbcWithDefaultSettings() {
 	void jdbcWithCustomPrefix() {
 		this.contextRunner.withUserConfiguration(DefaultConfiguration.class, EmbeddedDataSourceConfiguration.class)
 				.withPropertyValues("spring.datasource.generate-unique-name=true",
-						"spring.batch.jdbc.schema:classpath:batch/custom-schema-hsql.sql",
+						"spring.batch.jdbc.schema:classpath:batch/custom-schema-h2.sql",
```
snicoll, line 78: Ditto.

snicoll (changes requested):
Thanks for the PR but I don't understand why you've switched some of the tests from HSQL to H2.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java
hpoettker: The scope of the test is that the property `spring.batch.jdbc.schema` works as expected. It doesn't actually matter which DB is used.

I changed the name because the test does not use HSQL but H2. I noticed that as it fails with upgraded H2 which it shouldn't if HSQL were used.

As the test is not really about the concrete DB and the new DDL works with both HSQL and H@, would it be okay to rename the script to `custom-schema.sql`?

hpoettker:
I now removed the suffix `-hsql` from the custome schema file for the Spring Batch autoconfiguration files. The actually used DB is not important for the tests that use the file and it currently suggests that `HSQL` is used although it is not.

snicoll (approved):

snicoll:
Thanks a lot @hpoettker

PR #29274, opened 2022-01-06
Use emptyList() instead of null

hjun-park:
For me, I used emptyList() instead of null
so I want to suggest this PR for this.


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/error/AbstractErrorController.java
```diff
@@ -62,7 +63,7 @@ public AbstractErrorController(ErrorAttributes errorAttributes, List<ErrorViewRe
 
 	private List<ErrorViewResolver> sortErrorViewResolvers(List<ErrorViewResolver> resolvers) {
 		List<ErrorViewResolver> sorted = new ArrayList<>();
-		if (resolvers != null) {
```
spencergibb, line 65: There's no guarantee that resolvers is not null since there is no assertion in the constructor.

snicoll:
Thanks for the PR but as Spencer pointed out, there is no enforcement in the other constructor.

PR #29276, opened 2022-01-06
Warn about the dangers of early bean initialization when using @ConditionalOnExpression

### LINKED ISSUE #28771: Warn about the dangers of early bean initialization when using @ConditionalOnExpression
Referencing a bean in `@ConditionalOnExpression` will cause the bean to be initialized very early in context refresh processing. This can cause problems such as [this one](https://github.com/spring-projects/spring-framework/issues/27718). We should add a warning about this to the annotation's javadoc and, perhaps, somewhere in the references docs as well.

mihailcornescu:
Fix #28771

- added a note on using beans in `@ConditionalOnExpression` in the Javadoc and .adoc


snicoll:
@mihailcornescu thank you for making your first contribution to Spring Boot.

PR #29295, opened 2022-01-09
Implement InputStream.available() for RandomAccessDataFile-backed input streams

Gnaily:
When someone calles new `ClassPathResource("a.file").getInputStream().available()` in application, they will get 0 if  "a.file"(eg:"a.xlsx") is a `ZipEntry.Store` entry in the executable jar of that application. See [JarFileEntries](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/JarFileEntries.java)#getInputStream(FileHeader entry) method.


philwebb:
We should ideally add a test for this when we merge

licunzhi:
> When someone calls `new ClassPathResource("a.file").getInputStream().available()` in their application, they will get 0 if "a.file"(eg:"a.xlsx") is a `ZipEntry.Store` entry in the executable jar of that application. See [JarFileEntries](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/jar/JarFileEntries.java)#getInputStream(FileHeader entry) method.

Excuse me, 
How can I fix this problem with the previous version?

Issue #27899, opened 2021-09-09
Make TaskExecutor of SimpleJobLauncher configurable via BasicBatchConfigurer


aahlenst:
The auto-configuration for Spring Batch uses the `SimpleJobLauncher` with `SyncTaskExecutor`, which is the default. Consequently, Batch jobs are run on the same thread they are launched on, even if it is a Tomcat worker thread. I find this surprising, considering that Spring Boot typically employs sensible defaults, and I have not seen it mentioned in the docs (or in any issue). Furthermore, I expected that Spring Boot would pick up the auto-configured `TaskExecutor`, but it does not.

https://github.com/spring-projects/spring-boot/blob/56595c00278f1284c4889915e73d279cc4b20fdd/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BasicBatchConfigurer.java#L119-L124

If it was a conscious decision to use the `SimpleJobLauncher` with its default task executor, it would be nice if it were documented. Ideally, the docs would also offer some guidance on properly configuring Spring Batch with a different task executor (I expect the answer is to extend `BasicBatchConfigurer` and to override `createJobLauncher()`) and whether it's appropriate to use the auto-configured `TaskExecutor`.

Otherwise, I think it would make sense to change `BasicBatchConfigurer` to either pick up the auto-configured `TaskExecutor`, possibly introducing an annotation modelled after `@BatchDataSource` that allows to configure a dedicated `TaskExecutor` for batch jobs easily.

wilkinsona:
@benas What do you consider to be a sensible default for Batch's `TaskExecutor`?

benas:
I think the default should be a synchronous task executor as it is now. Changing the default to an asynchronous task executor would probably make a non-web boot app running a batch job exit without waiting for the job to finish. Executing batch jobs in servlet containers running 24/7 is actually not the recommended way and was one of the main reasons for [deprecating Spring Batch Admin](https://github.com/spring-attic/spring-batch-admin/blob/master/MIGRATION.md#whats-different-between-spring-batch-admin-and-spring-cloud-data-flow) in favor of the single-job containerized boot apps orchestrated by Spring Cloud Dataflow. That's why I think the default should be in line with the recommended packaging/execution model.

[...]

Issue #22403, opened 2020-07-20
Make it easier to define a bean in addition to an auto-configured bean of the same type


hjohn:
Problem:

I want to define my own `ObjectMapper` as a bean (as it needs to be injected with other dependencies) without suddenly breaking the rest of my application.

      @Bean("my-object-mapper")
      @Qualifier("my-object-mapper") 
      public ObjectMapper flowObjectMapper(RequiredDependencies etc) {  ...  }

This unfortunately breaks the rest of the application.  Despite giving it a name that nobody else uses, a `Qualifier` that nobody else knows about, things like Feign, RabbitListeners, etc. suddenly start using this completely differently configured `ObjectMapper` as their own.  Why?  Because `JacksonAutoConfiguration` decides not to configure the default primary bean because `ConditionalOnMissingBean` thinks that a bean with the correct **type** (not name, not qualifier) should match.  This results in some completely randomly configured bean that was clearly not intended to be used for anything but a specific purpose (hence the name and qualifier) is suddenly used as the default globally for everything.

Would it not have made more sense to still create this `Primary` bean if there is no suitable candidate which is **unqualified** and/or has the correct bean name?

Work-arounds:

- Define your own `Primary` `ObjectMapper`.  Problem: how do I know how the Spring default one was configured?  I suppose I could copy the source code if that doesn't change too often... `builder.createXmlMapper(false).build()` -- poor solution IMHO.

- Subclass `ObjectMapper` so Spring's type matching doesn't mistake my custom mapper as something it can use... won't work most likely, as it is still an `ObjectMaper`... so wrap it then and delegate... ouch.

- Complicated stuff to register my bean as a non-primary bean using a `BeanFactoryPostProcessor`.  Seriously?

Possible solutions that would work for me:

- Allow a non-primary bean to be defined easily.  `@Bean(primary = false)` or `@Bean(cannotBeUsedUnqualified = true)`

- Introduce a `ConditionalOnMissingUnqualifiedBean` and use that in `JacksonAutoConfiguration`.  Probably not backwards compatible.

wilkinsona:
Thanks for the suggestions.

Generally speaking, if an instance of something that is widely used is intended to be used for a specific purpose, particularly if it's in a single place, it may be better not to define it as a bean. I can't tell if that applies here as three-line code snippet above doesn't provide us with enough information.

Alternatively if you do need to define your `ObjectMapper` as a bean, there are a couple of other options that you could consider:

[...]

PR #29296, opened 2022-01-09
Allow customization of Spring Batch TaskExecutor

aahlenst:
This PR addresses some concerns I raised in #27899, taking the discussion therein into account. My primary objective is to make it easier for newcomers to use Spring Batch properly.

Currently, the support for Spring Batch is geared towards single-job containerized Spring Boot apps orchestrated by a job scheduler. I pointed that out in the docs because it's not obvious. I also added a note that using Spring Batch as part of a web app is a viable option but that it requires some configuration, namely configuring an asynchronous `TaskExecutor`.  Doing that is awkward because it requires overriding either `BasicBatchConfigurer` or `JpaBatchConfigurer`. Picking the right one requires knowledge about the Spring Batch auto-configuration. To avoid that I introduced `@BatchTaskExecutor` (modeled after `@BatchDataSource`). Thanks to  `@BatchTaskExecutor`, swapping the `TaskExecutor` used by Spring Batch is now a matter of defining an additional bean.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BasicBatchConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BasicBatchConfigurer.java
@@ -29,6 +29,7 @@ import org.springframework.beans.factory.InitializingBean;
 import org.springframework.boot.autoconfigure.batch.BatchProperties.Isolation;
 import org.springframework.boot.autoconfigure.transaction.TransactionManagerCustomizers;
 import org.springframework.boot.context.properties.PropertyMapper;
+import org.springframework.core.task.TaskExecutor;
 import org.springframework.jdbc.datasource.DataSourceTransactionManager;
 import org.springframework.transaction.PlatformTransactionManager;
 
@@ -39,6 +40,7 @@ import org.springframework.transaction.PlatformTransactionManager;
  * @author Andy Wilkinson
  * @author Kazuki Shimizu
  * @author Stephane Nicoll
+ * @author Andreas Ahlenstorf
  * @since 1.0.0
  */
 public class BasicBatchConfigurer implements BatchConfigurer, InitializingBean {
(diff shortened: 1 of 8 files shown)


snicoll:
@aahlenst the ask sounds sensible, but I am not keen to introduce a dedicated annotation for this. The issue is already flagged for team attention so let's see what the rest of the team thinks.

wilkinsona:
I have no objections to adding `@BatchTaskExecutor` here. As already noted, it feels nicely aligned with `@BatchDataSource`. I think it also fits fairly well with `@FlywayDataSource`, `@LiquibaseDatasource`, etc.

snicoll:
My original comment was lacking context. In #27899, it was identified that the default sync task executor is preferred, which rules out associating "the" `TaskExecutor` that is available in the context (we auto-configure one by default and it's definitely not suitable as Spring Batch's default).

`@BatchDataSource` was used as a qualifier when multiple datasources are available in the context, and to be able to elect the one that we want to use for batch processing. This doesn't apply here as we are not setting a task executor at all. If we want to improve the scenario of specifying a custom task executor, perhaps a customizer of some sort could be used?

wilkinsona:
> Besides, if you create such a bean, then our regular auto-configuration for application task executions will back off

I'd overlooked this aspect of it. While we have the same problem with, for example, `@BatchDataSource` causing the main auto-configured `DataSource` to back off,  I still think it's worth looking at a different approach. If we ever tackle https://github.com/spring-projects/spring-boot/issues/22403 we could then reconsider something based on a  `TaskExecutor` bean.

aahlenst:
According to my plan, there would always be the auto-configured `TaskExecutor` (or its manually configured replacement) and, if enabled, the additional `@BatchTaskExecutor`. But as you correctly pointed out, the auto-configuration would back off if it encountered a `@BatchTaskExecutor`. Therefore, my proposal does not make any sense. I am also reluctant to introduce a customizer. A customizer would be another layer of indirection. To me, it does not feel worth the additional complexity and maintenance effort considering the little value it would provide. Thanks anyway!

Issue #27801, opened 2021-08-23
server.tomcat.use-relative-redirects=true not honored when server.forward-headers-strategy=framework


merusso-regions:
For a Spring Boot webmvc application, when the property `server.tomcat.use-relative-redirects` is set to `true` and the server returns a 302 redirect, the `Location` header should contain only the URI path, not a full URL ("/path" instead of "http://host/path").

However, if the property `server.forward-headers-strategy` is also set to `framework` *and* the HTTP request contains a "forwarded" header like "X-Forwarded-Proto: http", the `Location` header contains a full URL, instead of only the URI path as expected.

### Detected in versions

Spring Boot: 2.5.4, 2.4.5
Java: 11.0.6

### Code Example

Controller:

```java
import org.springframework.stereotype.Controller;
import org.springframework.web.bind.annotation.GetMapping;
import org.springframework.web.bind.annotation.ResponseBody;

@Controller
public class RedirectController {
    @GetMapping("/redirect-to-foo")
    public String getRedirect() {
        return "redirect:/foo";
    }

    @GetMapping("/foo")
    @ResponseBody
    public String getFoo() {
        return "This is a test";
    }
}
```

Controller Test:

```java
import org.junit.jupiter.api.Test;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.boot.test.autoconfigure.web.reactive.AutoConfigureWebTestClient;
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.test.context.TestPropertySource;
import org.springframework.test.web.reactive.server.WebTestClient;

@SpringBootTest(webEnvironment = SpringBootTest.WebEnvironment.RANDOM_PORT)
@AutoConfigureWebTestClient
@TestPropertySource(properties = {
    "server.tomcat.use-relative-redirects=true",
    "server.forward-headers-strategy=framework"
})
class RedirectControllerTest {
    @Test
    void testRedirect(@Autowired WebTestClient webClient) throws Exception {
        webClient.get()
            .uri("/redirect-to-foo")
            .exchange()
            .expectStatus().is3xxRedirection()
            .expectHeader().location("/foo");
        // Succeeds
    }

    @Test
    void testRedirectWithForwardedHeaders(@Autowired WebTestClient webClient) throws Exception {
        webClient.get()
            .uri("/redirect-to-foo")
            .header("X-Forwarded-Proto", "http")
            .exchange()
            .expectStatus().is3xxRedirection()
[...]

PR #29333, opened 2022-01-12
Relative redirects is automatically enabled when `server.forward-headers-strategy` is configured as `framework`

terminux:
Fixes gh-27801


wilkinsona:
Thanks for the proposal, @terminux, but I think it may be a little too broad. I believe we should only configure the filter to use relative redirects when Tomcat's in use and it has been configured to use relative redirects. If another container's in use or Tomcat's using absolute redirects, the filter's configuration should be unchanged.

Unfortunately, I haven't had time to verify that what I believe is required actually is required or to think about how best to implement it. If you have time to do so, that'd be great. Please don't worry if you don't though and thanks again for your efforts thus far.

terminux:
Thank you very much for your reply @wilkinsona , nice suggestions. I will reconsider how to implement it based on your suggestion.

terminux:
Hi @wilkinsona , my idea is to add a static config class specifically for `ForwardedHeaderFilter` and we don't automatically enable relative redirects for tomcat containers, only if `server.tomcat.use-relative-redirects` is configured to `true`, It might be better. I updated the PR and it's ready for your review again.

wilkinsona:
Thanks, @terminux. When initially thinking about this, I had rejected something along these lines as I didn't want the configuration of the `ForwardedHeaderFilter` to know about Tomcat directly. However, having seen your proposal it's quite an elegant and pragmatic solution. It does need to be refined a little as there's no guarantee about the processing order of the two `@Bean` methods for the filter. The ordering can be controlled by putting each method in a separate class and `@Import`ing those two classes in the desired order.

The alternative to coupling to `ForwardedHeaderFilter` configuration directly to Tomcat would be a general callback interface for customizing the `ForwardedHeaderFilter` and a Tomcat-specific implementation of it. That would remove the need for multiple `@Bean` methods, but would require the new callback interface.

I'll flag this for a team meeting so that we can discuss these options and any others that come to mind.

terminux:
Thanks @wilkinsona , I originally thought about using `@Order` to control the processing order of two `@Bean` methods, but it's less elegant. However, when I saw your comment, I found that `@Import` is indeed a good choice !

I also thought about using the callback interface way, for example:

```java
interface ForwardedHeaderFilterCustomizer {

	void customize(ForwardedHeaderFilter filter);

}
```

But actually `ForwardedHeaderFilter` has only two methods (`setRemoveOnly` and `setRelativeRedirects`) that may be called. I'm not sure if adding a new callback interface for this would be a bit wasteful, but it scales well.

wilkinsona:
Thanks very much, @terminux. I went for a customizer callback in the end, but as in implementation detail that isn't part of Boot's public API.

PR #29389, opened 2022-01-14
Avoid split java.xml packages in docs project that cause compilation errors in Eclipse with Java 9 and later

dsyer:
Without this change I get compiler errors from JDT in the IDE (VSCode). It seems to be harmless from the point of view of the command line build with the JDK.

### DIFF
--- a/spring-boot-project/spring-boot-docs/build.gradle
+++ b/spring-boot-project/spring-boot-docs/build.gradle
@@ -83,6 +83,7 @@ dependencies {
 	implementation("jakarta.validation:jakarta.validation-api")
 	implementation("net.sourceforge.htmlunit:htmlunit") {
 		exclude group: "commons-logging", module: "commons-logging"
+		exclude group: "xerces", module: "xercesImpl"
 	}
 	implementation("org.apache.commons:commons-dbcp2") {
 		exclude group: "commons-logging", module: "commons-logging"


wilkinsona:
That's a bit odd. It works fine with JDT in Eclipse. What are the compile errors you're getting?

dsyer:
```
The package javax.xml.transform is accessible from more than one module
```

wilkinsona:
`xercesImpl` only contains `org.apache.*` and `org.w3c.dom.html` packages. I guess that it's its dependency on `xml-apis` that's the problem and is probably what, if anything, should be excluded. It contains a superset of the packages that are in the JDK so excluding it may break something.

We have three other modules that transitively depend on `xml-apis`:

- `spring-boot-test`
- `spring-boot-test-autoconfigure`
- `spring-boot-smoke-test-test`

While `spring-boot-test` and `spring-boot-test-autoconfigure` are depended on by users, they won't pick up the `xml-apis` dependency as it's pulled in via an optional dependency on HTML Unit. I guess we can try excluding it across all four and see what happens.

sbrannen:
In https://github.com/spring-projects/spring-framework/commit/a247b83cd9c9aefd3c329d493c5ce7cd11d0cdfa, you can see what we had to do for Spring Framework as well as additional resources on the subject.

wilkinsona:
> It works fine with JDT in Eclipse

Looking more closely at my Boot 3.0 Eclipse workspace, this isn't the case. The only errors we used to get were a handful of erroneous "The target type of this expression must be a functional interface" errors but I'm now seeing a number of "accessible from more than one module" errors too. I'm still on Eclipse 2021-09.

wilkinsona:
Thanks, @sbrannen. It's useful to know that you were able to exclude `xml-apis` without any problems.

wilkinsona:
The problem only appears in Eclipse if you have `xml-apis` on the classpath and there's code in the project that uses code in a package that's then in both the unnamed module and the `java.xml` module. As a result of this, it's only `spring-boot-docs` that's affected as `spring-boot-test`, `spring-boot-test-autoconfigure`, and `spring-boot-smoke-test-test` don't use any `java.xml` APIs.

PR #29429, opened 2022-01-16
Provide a callback mechanism for customizing LocalValidatorFactoryBean's configuration

dangzhicairang:
Add a AddValueExtractorsLocalValidatorFactoryBean that expand LocalValidatorFactoryBean which based SpringFramework, it can support user add custom ValueExtractor(s). For example, user can define ValueExtractor like this in their application:
```
@Component
public class CustomValueExtractor implements ValueExtractor<CustomResult<@ExtractedValue ?>> {
 
     @Override
     public void extractValues(CustomResult<?> result, ValueReceiver valueReceiver) {
         valueReceiver.value(null, result.getData());
     }
}
```

And modify the class ValidationAutoConfiguration, auto-configure an instance of type AddValueExtractorsLocalValidatorFactoryBean instead of type LocalValidatorFactoryBean

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/validation/ValidationAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/validation/ValidationAutoConfiguration.java
@@ -60,7 +60,8 @@ public class ValidationAutoConfiguration {
 	@Bean
 	@Role(BeanDefinition.ROLE_INFRASTRUCTURE)
 	@ConditionalOnMissingBean(Validator.class)
-	public static LocalValidatorFactoryBean defaultValidator(ApplicationContext applicationContext, List<ValueExtractor> valueExtractors) {
+	public static LocalValidatorFactoryBean defaultValidator(ApplicationContext applicationContext,
+			List<ValueExtractor> valueExtractors) {
 		AddValueExtractorsLocalValidatorFactoryBean factoryBean = new AddValueExtractorsLocalValidatorFactoryBean();
 		factoryBean.setValueExtractors(valueExtractors);
 		MessageInterpolatorFactory interpolatorFactory = new MessageInterpolatorFactory(applicationContext);
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/validation/ValidationAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/validation/ValidationAutoConfiguration.java
@@ -16,8 +16,11 @@
 
 package org.springframework.boot.autoconfigure.validation;
 
+import java.util.List;
+
 import jakarta.validation.Validator;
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for the proposal, @dangzhicairang. Do you think it's likely that you will want to inject dependencies into a custom`ValueExtractor` implementation? If so, there would be a benefit to them being beans, but if not it may be better to provide a more general callback for customizing the validator `Configuration`.

dangzhicairang:
> Thanks for the proposal, @dangzhicairang. Do you think it's likely that you will want to inject dependencies into a custom`ValueExtractor` implementation? If so, there would be a benefit to them being beans, but if not it may be better to provide a more general callback for customizing the validator `Configuration`.

Thank you, it is a really nice suggestion, and  i provide a callback class  Customizer for the Configuration. And i retained the ability  to automatically add the ValueExtractor(s) bean, so user can choose whether these ValueExtractor(s) are bean or not. The ValueExtractor which is bean will automatically add but added by Customizer or AddValueExtractorCustomizer if it not.

wilkinsona:
Thanks for the updates, @dangzhicairang. Our feeling is that the general-purpose callback mechanism possibly belongs in Spring Framework rather than in Spring Boot, particularly as Framework already has its own `LocalValidatorFactoryBean` sub-class. I've raised https://github.com/spring-projects/spring-framework/issues/27956 so that the Framework team can consider things. I'll place this proposal on hold until we know what, if anything, the Framework team are going to do.

dangzhicairang:
> Thanks for the updates, @dangzhicairang. Our feeling is that the general-purpose callback mechanism possibly belongs in Spring Framework rather than in Spring Boot, particularly as Framework already has its own `LocalValidatorFactoryBean` sub-class. I've raised [spring-projects/spring-framework#27956](https://github.com/spring-projects/spring-framework/issues/27956) so that the Framework team can consider things. I'll place this proposal on hold until we know what, if anything, the Framework team are going to do.

Thank you for your reply @wilkinsona . It is right.

snicoll:
https://github.com/spring-projects/spring-framework/issues/27956 has been addressed.

wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @dangzhicairang.

I have updated your proposal a little in  https://github.com/spring-projects/spring-boot/commit/0e00fafe385256d5da1e5a26fd8fd97173ba8d32 to make use of the changes to Framework's `LocalValidatorFactoryBean`. In the interests of simplicity, I've removed support for `ValueExtractor` beans, preferring a more general callback mechanism that can be used to customise the configuration as needed, including adding value extractors.

dangzhicairang:
> Thanks very much for making your first contribution to Spring Boot, @dangzhicairang.
> 
> I have updated your proposal a little in [0e00faf](https://github.com/spring-projects/spring-boot/commit/0e00fafe385256d5da1e5a26fd8fd97173ba8d32) to make use of the changes to Framework's `LocalValidatorFactoryBean`. In the interests of simplicity, I've removed support for `ValueExtractor` beans, preferring a more general callback mechanism that can be used to customise the configuration as needed, including adding value extractors.

Thanks a lot for your update  @wilkinsona

Issue #29390, opened 2022-01-14
Enhancement: Add option for spring-boot-maven-plugin:repackage goal to always generate filenames from GAV


u3r:
When packaging two dependencies with same artifact name + version (but different groupID) the  spring-boot-maven-plugin:repackage goal already adds the groupID to the packaged filename (see [ArtifactsLibraries.java:doWithLibraries()](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/ArtifactsLibraries.java)

For better traceability within the produced fat-jar I'd like to have an option in the repackage configuration to force the inclusion of full GAV for all artifacts. Otherwise (badly named) artifacts with common names like `core` `common` are not immediately obvious.

dratler:
Hi @philwebb I have created a [PR](https://github.com/spring-projects/spring-boot/pull/29437) for this issue

wilkinsona:
Maven's war plugin provides an `outputFileNameMapping` for this purpose, the usage of which is [described here](https://maven.apache.org/plugins/maven-war-plugin/examples/file-name-mapping.html). It seems like a good source of inspiration and I think it makes sense to align with something that Maven users may already be familiar with.

 If we to make this possible with Maven, we should also consider what, if anything, we want to do for Gradle. It's already possible with Gradle without any changes to Boot with some configuration like this:
 
```
tasks.named('bootJar') {
    rootSpec.rename { fileInJar ->
        def artifact = project.configurations.runtimeClasspath.resolvedConfiguration.resolvedArtifacts.find { it.file.name == fileInJar }
        return artifact ? "${artifact.moduleVersion.id.group}-${fileInJar}" : fileInJar
    }
}
```

It assumes that everything in `BOOT-INF/lib` will have come from the `runtimeClasspath`. This assumption will hold true in most cases. It's not as declarative as something like Maven's `outputFileNameMapping` but it is far more flexible. Internally we have `LayerResolver` which provides some more sophisticated mapping of a file that's being copied into a jar to its dependency coordinates. We could, perhaps, build something like `outputFileNameMapping` on top of this if we wanted a simpler, more declarative configuration model that would look something like this:

tasks.named('bootJar') {
     outputFileNameMapping = "${groupId}-${artifactId}-${version}.${ext}"
}

Given the above, I don't think this issue is ideal for contribution. We need to make some design decisions before it can progress.

[...]

PR #29437, opened 2022-01-16
fix: add groupId before version number

dratler:


u3r:
Hi @dratler, this _always_ adds the groupid. I didn't provide **this** solution as it changes the status quo.

Adding a configuration option (as mentioned in the issue) would involve patching through the option from the plugin into the `ArtifactsLibararies` or even configure a naming strategy.

For me personally this solution is perfectly ok - but I didn't spend time to think of implications of very long group/artifact-name combinations or other possible clashes.

dratler:
Hi @u3r ,
I have added it and also on tests add ref to getClassifier() it's also part of the solution

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/test/java/org/springframework/boot/maven/ArtifactsLibrariesTests.java
```diff
@@ -118,19 +118,22 @@ void renamesDuplicates() throws Exception {
...
 		given(artifact2.getArtifactHandler()).willReturn(this.artifactHandler);
 		this.artifacts = new LinkedHashSet<>(Arrays.asList(artifact1, artifact2));
 		this.libs = new ArtifactsLibraries(this.artifacts, Collections.emptyList(), null, mock(Log.class));
 		this.libs.doWithLibraries(this.callback);
 		verify(this.callback, times(2)).library(this.libraryCaptor.capture());
-		assertThat(this.libraryCaptor.getAllValues().get(0).getName()).isEqualTo("g1-artifact-1.0.jar");
-		assertThat(this.libraryCaptor.getAllValues().get(1).getName()).isEqualTo("g2-artifact-1.0.jar");
+		assertThat(this.libraryCaptor.getAllValues().get(0).getName()).isEqualTo(this.libraryCaptor.getAllValues().get(1).getName());
+		assertThat(this.libraryCaptor.getAllValues().get(0).getName()).isEqualTo("g1-artifact-g1-1.0-cl-1.jar");
+		assertThat(this.libraryCaptor.getAllValues().get(1).getName()).isEqualTo("g1-artifact-g1-1.0-cl-1.jar");
```
u3r, line 136: Isn't this the thing to avoid? According to the test name?
So shouldn't 
a) the input to these tests have at least one differing field so the resulting filenames are different **or**
b) throw an error as packaging two artifacts with identical coordinates seem fishy **or**
c) (least likely, considering (b) ) Add something to discern the files?

Can anyone official and familiar with the specific semantic of the plugin chime in, what expected output for Artifacts with identical coordinates should be or where this is filtered/addressed?

wilkinsona: Maven doesn't allow multiple dependencies with identical coordinates.

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/ArtifactsLibraries.java
```diff
@@ -164,12 +164,15 @@ private boolean isLocal(Artifact artifact) {
 
 	private String getFileName(Artifact artifact) {
 		StringBuilder sb = new StringBuilder();
-		sb.append(artifact.getArtifactId()).append("-").append(artifact.getBaseVersion());
+		sb.append(artifact.getArtifactId()).append("-").append(artifact.getGroupId()).append("-")
```
wilkinsona, line 167: Rather than doing this here, I think it would be better to reuse [the existing logic that prepends the group ID when a duplicate is detected](https://github.com/spring-projects/spring-boot/blob/a1c8997f615a66da7cd471f87921830ea0408ba7/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/ArtifactsLibraries.java#L114-L118). A configuration option should be added so that, when enabled, the group ID always prepended irrespective of whether or not a duplicate has been detected.

wilkinsona (changes requested):

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/ArtifactsLibraries.java
dratler: Hi @wilkinsona ,
method `` doWithLibraries`` it is using  method `getDuplicates` and it's using `getFileName` and on loop for every file \ Artifacts we are using that method this is why I thought we can fix only at method `getFileName`
how can I inject that opinion ?

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/test/java/org/springframework/boot/maven/ArtifactsLibrariesTests.java
dratler: I will be working on :
1. Duplicate Entry 
2. Entry with diffrent Art id
3. Entry with diffrent Group Ip
4. Entry with diffrent Version Number
So all scenarios needs to be checked

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/ArtifactsLibraries.java
dratler: Hi @wilkinsona ,
do you think that ```getDuplicates``` should return Artifices and then it can be better compare ?

wilkinsona: I don't think there's anything to compare. The file name already contains the artifact ID, version, and any classifier. 
Maven guarantees that within a group, each combination of artifact ID, version, and classifier will be unique. Therefore, prepending the group ID is sufficient to avoid a duplicate entry name.

wilkinsona:
How's it going, @dratler?

wilkinsona:
Thanks for you efforts here, @dratler. Given the lack of progress in the last few weeks and [my comment on #29390](https://github.com/spring-projects/spring-boot/issues/29390#issuecomment-1039136555), I think we need to take a step back and think some more about exactly what we want to do here. Thanks anyway.

PR #29438, opened 2022-01-16
Provide some guidance on identifying and resolving Devtools classloading issues

aahlenst:
If you encounter the problem of the two classloaders in Devtools, there's no hint that there are two classloaders and that some configuration changes might help you. You only get an IllegalArgumentException. Googling does not help because the exception is not mentioned in the docs. Futhermore, the docs sound like the problem can only occur in IDEs, which is not the case.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/devtools.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/devtools.adoc
@@ -24,6 +24,8 @@ To include devtools support, add the module dependency to your build, as shown i
 	}
 ----
 
+CAUTION: Devtools might cause classloading issues, in particular in multi-module projects. <<using#using.devtools.diagnosing-classloading-issues>> explains how to diagnose and solve them.
+
 NOTE: Developer tools are automatically disabled when running a fully packaged application.
 If your application is launched from `java -jar` or if it is started from a special classloader, then it is considered a "`production application`".
 You can control this behavior by using the `spring.devtools.restart.enabled` system property.
@@ -37,6 +39,17 @@ TIP: Repackaged archives do not contain devtools by default.
 If you want to use a <<using#using.devtools.remote-applications,certain remote devtools feature>>, you need to include it.
 When using the Maven plugin, set the `excludeDevtools` property to `false`.
 When using the Gradle plugin, {spring-boot-gradle-plugin-docs}#packaging-executable-configuring-including-development-only-dependencies[configure the task's classpath to include the `developmentOnly` configuration].
+
+
+
+[[using.devtools.diagnosing-classloading-issues]]
+=== Diagnosing Classloading Issues
+
+As described in the <<using#using.devtools.restart.restart-vs-reload>> section, restart functionality is implemented by using two classloaders.
+For most applications, this approach works well.
+However, it can sometimes cause classloading issues, in particular in multi-module projects.
+
+To diagnose whether the classloading issues are indeed caused by devtools and its two classloaders, <<using#using.devtools.restart.disable,try disabling restart>>. If this solves your problems, <<using#using.devtools.restart.customizing-the-classload,customize the restart classloader>> to include your entire project.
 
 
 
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/devtools.adoc
```diff
@@ -217,10 +217,18 @@ For IntelliJ IDEA, you can follow the https://www.jetbrains.com/help/idea/spring
...
+
+[source,text]
+----
+Caused by: java.lang.IllegalArgumentException: non-public interface is not defined by the given loader
+----
 
 By default, any open project in your IDE is loaded with the "`restart`" classloader, and any regular `.jar` file is loaded with the "`base`" classloader.
 If you work on a multi-module project, and not every module is imported into your IDE, you may need to customize things.
-To do so, you can create a `META-INF/spring-devtools.properties` file.
+The same is true if you use `mvn spring-boot:run` or `gradle bootRun` in a multi-module project: The project containing your `@SpringBootApplication` is loaded with the "`restart`" classloader, and everything else with the "`base`" classloader.
```
wilkinsona, line 229: I think this is a good addition. Thank you.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/devtools.adoc
```diff
@@ -217,10 +217,18 @@ For IntelliJ IDEA, you can follow the https://www.jetbrains.com/help/idea/spring
...
+
+To instruct Spring Boot to load `.jar` files with a different classloader, you can create a `META-INF/spring-devtools.properties` file.
```
wilkinsona, line 231: I'm not sure that this is needed. If it is, I think it should be part of the next paragraph rather than in a paragraph of its own.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/devtools.adoc
```diff
@@ -217,10 +217,18 @@ For IntelliJ IDEA, you can follow the https://www.jetbrains.com/help/idea/spring
 As described earlier in the <<using#using.devtools.restart.restart-vs-reload>> section, restart functionality is implemented by using two classloaders.
 For most applications, this approach works well.
 However, it can sometimes cause classloading issues.
+These might manifest in an exception like:
+
+[source,text]
+----
+Caused by: java.lang.IllegalArgumentException: non-public interface is not defined by the given loader
+----
```
wilkinsona, line 225: I'm not in favour of specifically mentioning `IllegalArgumentException` here. Unfortunately, class loading issues can appear in many different ways. Mentioning one specifically is likely to do more harm than good.

wilkinsona (changes requested):
Thanks for the PR, @aahlenst. I've made some comments for your consideration when you have a moment.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/devtools.adoc
aahlenst: It's less about the `IllegalArgumentException` and more about the error message itself. Recently, I was called in to troubleshoot exactly this problem. At first I was stumped because the application ran fine in the IDE and when packaged. It took me a lot of time to figure out what the problem was. Feeding the error message to search engines didn't yield good results, either. 

Do you have an idea how to make this easier to discover without misleading people that have other classloading problems?

wilkinsona: I'm afraid not. It's a shame that the JDK uses "loader" rather than the arguably less ambiguous "class loader" as that may help to point people in the right direction.

The docs on DevTools currently mention "classloading issues" but don't really expand on how to  diagnose them. Rather than trying to list specific failures that may be a classloading issue, I think it would be better to point people towards [completely disabling restart](https://docs.spring.io/spring-boot/docs/2.5.x/reference/htmlsingle/#using.devtools.restart.disable) as a way of identifying if the split class loader is the cause.

aahlenst:
@wilkinsona I reworked my PR. There's now a separate section about diagnosing classloading issues incl. an admonition block that points out that devtools might cause them.

Issue #29384, opened 2022-01-13
Add option to pass configured Docker host to the image builder container


scottfrederick:
When the Docker host used for building images in the build plugins is configured using the `DOCKER_HOST` environment variable or the `docker.host` configuration property, this value should be bound to the builder container in order for the CNB lifecycle to access the host. The build plugins currently [always bind the default socket](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Phase.java#L135)  `/var/run/docker.sock:/var/run/docker.sock` instead of effectively binding `${DOCKER_HOST}:/var/run/docker.sock` when `DOCKER_HOST` is set.

terminux:
Hi, I submitted a PR to fix this issue. #29441

scottfrederick:
After further research and testing, there are cases where we want the socket to always be bound to `/var/run/docker.sock` as is currently done and other cases where we want the socket bound to the value of the configured docker host. The `/var/run/docker.sock` value will always work with Docker Engine and minikube because those daemons make that socket available in the container that uses the binding. Other container engines require a different value like `/run/user/1000/podman/podman.sock` for podman.

In order to support both cases we will need a build plugin configuration property that the user can enable to select the non-default behavior.

PR #29441, opened 2022-01-17
Fix Docker socket not binding effectively

terminux:
Fixes gh-29384

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Phase.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Phase.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2021 the original author or authors.
+ * Copyright 2012-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -20,6 +20,7 @@ import java.util.ArrayList;
 import java.util.Arrays;
 import java.util.LinkedHashMap;
 import java.util.List;
+import java.util.Locale;
 import java.util.Map;
 
 import org.springframework.boot.buildpack.platform.docker.type.Binding;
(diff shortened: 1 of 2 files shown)


scottfrederick:
@terminux Thanks for the contribution, but this change doesn't meet the requirements in the issue. With this change, a `docker.host` system property or `DOCKER_HOST` environment variable must be provided or the request to build an image will fail. The current code to access the Docker daemon accepts a `DOCKER_HOST` environment variable or a [host provided in plugin configuration](https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#build-image.docker-daemon) and defaults to a local socket otherwise. 

To properly implement this fix, more design will need to be done to pass the Docker configuration information [contained in the `Builder`](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Builder.java#L53), through the [created `Lifecycle`](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Builder.java#L150), to the [created `Phase`](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/build/Lifecycle.java#L148) so it can be properly bound. 

I'm going to close this PR for now. If you'd like to work on this more, just let us know and push more commits so we can review again.

terminux:
Thank you very much for your guidance @scottfrederick , I get it. I will reconsider how to implement it based on your suggestion.

terminux:
Hi @scottfrederick , my idea is to add a private constructor to `DockerApi` and the original constructor will call it.

For example:

```java
private final DockerHost host;

private DockerApi(HttpTransport http, DockerHost host) {
	......
	this.host = host;
}
```

or

```java
private final String dockerHost;

private DockerApi(HttpTransport http, DockerHost host) {
	......
	this.dockerHost = (host != null) ? host.getAddress() : null;
}
```


This way you can get docker host from the `DockerApi` when the `Phase` is created and then pass it. I compared a few ways I can think of, this way the change is probably the smallest. I updated this [commit](https://github.com/terminux/spring-boot/commit/fb5e43a3fe85c294e596841c5a65d29af9c570c3) in my repo and it's ready for your review again. Do I need to resubmit the PR ? Looking forward to your reply :)

Issue #29304, opened 2022-01-10
Improve project setup for IntelliJ IDEA


vpavic:
At present, the IntelliJ IDEA project setup has some manual steps - see [_Import additional code style_](https://github.com/spring-projects/spring-boot/wiki/Working-with-the-Code#import-additional-code-style) section of the _Working with the Code_ wiki page that refers to [`idea/codeStyleConfig.xml`](https://github.com/spring-projects/spring-boot/tree/main/idea).

This step could be a bit easier by checking in this file in version control under it's natural place (that is, under `.idea/codeStyles`). This is also suggested by Jetbrains - see the article on [_How to manage projects under Version Control Systems_](https://intellij-support.jetbrains.com/hc/en-us/articles/206544839-How-to-manage-projects-under-Version-Control-Systems).

The manual step on project setup isn't the only downside of the current setup:
- any updates to the file from the team require manual export (since it's natural location is git ignored)
- contributors need to re-import to pick up any updates

Additionally, on top of the code style config, some additional things like inspection profiles could also be shared. For example, the _Instance field access not qualified with 'this'_ inspection is helpful because IDEA will then qualify fields with `this` when it generates any code (for instance, field accessors). This helps because it prevents Checkstyle errors and therefore saves time.

Basically, I'm proposing to ditch the mentioned manual step in favor of checking in `.idea/codeStyles` and `.idea/inspectionProfiles` directories. In terms of `.gitignore` that boils down to something like this:

```.gitignore
!.idea/
.idea/*
!.idea/codeStyles
!.idea/inspectionProfiles
```

dreis2211:
Why not just use and extend the `.editorconfig` instead?

anilkumarkatta204:
Also, may be add pre-requiste details for PKIX Certificate Issues, when configuring the unsecured environments.

snicoll:
In the past, importing a project with an existing `.idea` directory led to a change of behavior in the way updates were applied so we'd have to test this first.

vpavic:
I've opened a draft PR to take a look at this - see #29446.

> ... led to a change of behavior in the way updates were applied ...

@snicoll, can you clarify this? I'm not sure I understood what _behavior_ and _updates_ exactly did you refer to.

> Why not just use and extend the `.editorconfig` instead?

[...]

PR #29446, opened 2022-01-17
Improve project setup for IntelliJ IDEA

vpavic:
This commit improves project setup for IntelliJ IDEA by adding code style and inspection profile project files to version control.

Closes gh-29304

---

Some notes:
- importing the existing `idea/codeStyleConfig.xml` as project code style config resulted in some diff - see the diff of this file vs `.idea/codeStyles/Project.xml`
- I've added inspections profile with one non-default option (_Instance field access not qualified with 'this'_), this likely should be expanded
- existing `idea/codeStyleConfig.xml` should be removed before merging

### DIFF
--- a/.idea/codeStyles/Project.xml
+++ b/.idea/codeStyles/Project.xml
@@ -1,0 +1,121 @@
+<component name="ProjectCodeStyleConfiguration">
+  <code_scheme name="Project" version="173">
+    <option name="AUTODETECT_INDENTS" value="false" />
+    <option name="OTHER_INDENT_OPTIONS">
+      <value>
+        <option name="USE_TAB_CHARACTER" value="true" />
+      </value>
+    </option>
+    <option name="CLASS_COUNT_TO_USE_IMPORT_ON_DEMAND" value="50" />
+    <option name="NAMES_COUNT_TO_USE_IMPORT_ON_DEMAND" value="500" />
+    <option name="IMPORT_LAYOUT_TABLE">
+      <value>
+        <package name="java" withSubpackages="true" static="false" />
+        <emptyLine />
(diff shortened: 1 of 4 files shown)


.idea/codeStyles/Project.xml
```diff
@@ -0,0 +1,121 @@
...
+        <option name="USE_TAB_CHARACTER" value="true" />
+      </indentOptions>
+    </codeStyleSettings>
+    <codeStyleSettings language="kotlin">
+      <indentOptions>
+        <option name="USE_TAB_CHARACTER" value="true" />
+      </indentOptions>
+    </codeStyleSettings>
+  </code_scheme>
+</component>
```
Schifty1988, line 121: add newline character to end of file

vpavic: This file is generated by IntelliJ, so it shouldn't really be edited manually.

Schifty1988: it was added here though:
https://github.com/spring-projects/spring-boot/blob/main/idea/codeStyleConfig.xml

vpavic: Exactly the point of #29304 and this PR is to avoid/reduce manual project setup. The files in this PR are both located exactly where IntelliJ places them and with IntelliJ generated contents, unlike the one you linked.

Either way, this discussion doesn't really add any value. Concerns like these are handled by the team if and when the PR gets merged.

snicoll:
Thanks @vpavic. I've played with it a little bit in a fresh project and this seems to work quite nicely. I've added the copyright headers with the relevant scope so that headers of changed files can be updated automatically.

Issue #28926, opened 2021-12-08
Perform failure analysis when a web application cannot be started due to a missing web server factory bean


CangYa2021:
When building the springboot project, I configured spring.main.web-application-type: reactive in application.yml, but forgot to add the corresponding starter:（spring-boot-starter-weblux）. I also added the web starter（spring-boot-starter-web）. The error reports given after the program starts are very confusing.

```
org.springframework.context.ApplicationContextException: Unable to start reactive web server; nested exception is org.springframework.context.ApplicationContextException: Unable to start ReactiveWebApplicationContext due to missing ReactiveWebServerFactory bean.
	at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext.onRefresh(ReactiveWebServerApplicationContext.java:82)
	at org.springframework.context.support.AbstractApplicationContext.refresh(AbstractApplicationContext.java:577)
	at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext.refresh(ReactiveWebServerApplicationContext.java:64)
	at org.springframework.boot.SpringApplication.refresh(SpringApplication.java:765)
	at org.springframework.boot.SpringApplication.refreshContext(SpringApplication.java:445)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:338)
	at com.springboot.SpringBoot.main(SpringBoot.java:34)
Caused by: org.springframework.context.ApplicationContextException: Unable to start ReactiveWebApplicationContext due to missing ReactiveWebServerFactory bean.
	at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext.getWebServerFactoryBeanName(ReactiveWebServerApplicationContext.java:108)
	at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext.createWebServer(ReactiveWebServerApplicationContext.java:90)
	at org.springframework.boot.web.reactive.context.ReactiveWebServerApplicationContext.onRefresh(ReactiveWebServerApplicationContext.java:79)
	... 6 common frames omitted
```

For my scenario, it is caused by the lack of the corresponding class library. I think it should be verified when the springapplication instance is initially constructed. The error exception should not be reported at the stage of refresh context , and the corresponding exception information means that the corresponding bean is missing, which can't remind me that I forgot to add the corresponding starter.

wilkinsona:
[...]

PR #29447, opened 2022-01-18
Custom applicationContextException failure analyzer for webflux 

zhaoxianming:
I have customized an applicationContextException failure analyzer for webflux, mainly for the project where the web-application-type is reactive but there is no exception message that the corresponding class file is missing.


zhaoxianming:
related issue #28926

wilkinsona:
Thanks for the proposal, @zhaoxianming. As described on #28926, I think it would be better to implement this by throwing a custom exception rather than looking for a `ApplicationContextException` with a specific message. It should make the failure analysis more robust. We also need to add analysis on the servlet side as well as the failure isn't specific to reactive web applications. Would you like to update this pull request along these lines?

zhaoxianming:
Thanks a lot, that's a great suggestion, I'll resubmit the PR as per this.

wilkinsona:
It's been three weeks now. Closing.

PR #29499, opened 2022-01-20
Add Kotlin alternatives to Java samples

p7nov:
In this PR, I'm adding the Kotlin alternatives to Java samples in the Spring Boot reference. They are mostly translated with the IntelliJ IDEA built-in J2K converter and later adjusted manually for a more Kotlin-idiomatic look.

Feel free to change the new samples as you see best or ask me to do this.

Regards, Pavel
Kotlin Technical Writer


buildSrc/src/main/java/org/springframework/boot/build/KotlinConventions.java
```diff
@@ -46,9 +46,9 @@ void apply(Project project) {
 		project.getPlugins().withId("org.jetbrains.kotlin.jvm", (plugin) -> {
 			project.getTasks().withType(KotlinCompile.class, (compile) -> {
 				KotlinJvmOptions kotlinOptions = compile.getKotlinOptions();
-				kotlinOptions.setApiVersion("1.3");
-				kotlinOptions.setLanguageVersion("1.3");
-				kotlinOptions.setAllWarningsAsErrors(true);
+				kotlinOptions.setApiVersion("1.6");
+				kotlinOptions.setLanguageVersion("1.6");
+				kotlinOptions.setAllWarningsAsErrors(false);
```
p7nov, line 51: I took the liberty to switch this to `false` because there are warnings that are meaningless in sample code (such as `parameter is never used` in functions with empty bodies) and they break the build.

spring-boot-project/spring-boot-docs/src/main/kotlin/org/springframework/boot/docs/features/testing/springbootapplications/springmvctests/MyControllerTests.kt
```diff
@@ -0,0 +1,46 @@
...
+import org.springframework.boot.test.mock.mockito.MockBean
+import org.springframework.http.MediaType
+import org.springframework.test.web.servlet.MockMvc
+import org.springframework.test.web.servlet.request.MockMvcRequestBuilders
+import org.springframework.test.web.servlet.result.MockMvcResultMatchers
+
+@WebMvcTest(UserVehicleController::class)
+class MyControllerTests(
+	@Autowired val mvc: MockMvc,
+	@MockBean val userVehicleService: UserVehicleService) {
```
jnizet, line 32: AFAIK, there is no ParameterResolver for parameters annotated with MockBean (which, BTW annotates the backing field, and not the parameter). So, unless I missed something, this will fail at runtime. Instead, `userVehicleService` should be a `lateinit var`.


Sorry for just leaving this comment: I stumbled on this almost by accident and felt the need to signal the issue, but I applaud your huge effort to translate all those examples in Kotlin. 👏

buildSrc/src/main/java/org/springframework/boot/build/KotlinConventions.java
wilkinsona: We should do this specifically for the docs project, leaving warnings as errors by convention.

spring-boot-project/spring-boot-docs/src/main/kotlin/org/springframework/boot/docs/features/testing/springbootapplications/springmvctests/MyControllerTests.kt
p7nov: Thanks, fixed this case in three examples.

buildSrc/src/main/java/org/springframework/boot/build/KotlinConventions.java
p7nov: Returned this setting to `true`; configured `false` in `build.gradle` of spring-boot-docs.

spring-boot-project/spring-boot-docs/src/main/kotlin/org/springframework/boot/docs/features/testing/springbootapplications/springmvctests/MyControllerTests.kt
jnizet: Oh sorry, I wasn't clear enough. The MockBean annotation is needed, but it needs to annotate the lateinit var:

```
@MockBean
lateinit var userVehicleService: UserVehicleService
```

p7nov: Got it, fixed.

snicoll:
Thank you very much for this contribution. There is some related work that needs to happen before we can merge this. Our asciidoctor backend automatically [removes imports, package statement and copyright headers](https://docs.spring.io/spring-boot/docs/current/reference/html/getting-started.html#getting-started.installing.cli.quick-start) but that doesn't work for Kotlin source code. I've created https://github.com/spring-io/spring-asciidoctor-backends/issues/55

philwebb:
Thanks very much for this amazing PR @p7nov! I've merged it to `2.7.x` and `main`.

I made a few changes to our asciidoctor backend to support Kotlin and to also automate the discovery of Kotlin source files. I've also tweaked a few things in [this](https://github.com/spring-projects/spring-boot/commit/d212243eef8c08ab76aa27aafd6500f5a2f259e6) polish commit. I hope the updates are OK.

The new docs should appear [here](https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/) once CI has completed a build.

Thanks again!

Issue #28845, opened 2021-11-29
Consider generics in @ConditionalOnMissingBean when possible


roookeee:
Consider the following scenario:

```
// loaded via spring factories
public class SomeAutoConfiguration {

    @ConditionalOnMissingBean
    @Bean
    public Converter<String, Integer> converter() {
         // ...
     }
}

// user configuration
@Configuration
public class SomeOtherConfiguration {
   @Bean
   public Converter<Integer, Float> someConverter() {
       // ...
   }
}
```

The `@CondtionalOnMissingBean` annotated factory method will not be invoked as the conditional disregrads generic type information.

`@ConditionalOnMissingBean`s JavaDoc states that `@Bean` annotated methods default to the return type of the method for missing bean checks. A bean factories methods return types reflection data contains generic information which Spring Boot could use to do a more thorough check of existing beans. Currently Spring Boots `OnBeanCondition` relies on `SimpleMethodMetadata` which does not expose the generic return type of the factory method, so spring-core probably needs adjustments as well.

I think the `@ConditionalOnMissingBean` condition should include generic type information in the future as Spring (Boot) supports autowring of generic types and has extended it's generic type support over the years which seems to be lacking in this area.

The only workaround is to use bean names instead of types, which feels off when building Spring Boot starters as users have to name their beans accordingly instead of just supplying a bean of the correct type.

Note: non-factory-method provided beans could also be supported via `getGenericSuperclass()` based checks for `@Component`s.

philwebb:
Closing in favor of PR #29500

PR #29500, opened 2022-01-20
Add ConditionalOnBean support for generic @Bean return types

VladislavSevruk:
Add generic bean support using `@ConditionalOnBean` and `@ConditionalOnMissingBean` annotations as described in #28845. Added two method to annotations: `typeArguments` and `typeArgumentNames` for classes and class names respectively.

Usage example:

```java
// loaded via spring factories
public class SomeAutoConfiguration {

    @ConditionalOnMissingBean(typeArguments = { String.class, Integer.class })
    @Bean
    public Converter<String, Integer> converter() {
         // ...
     }
}
```

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/ConditionalOnBean.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/ConditionalOnBean.java
@@ -56,7 +56,6 @@ import org.springframework.context.annotation.Conditional;
  * another auto-configuration, make sure that the one using this condition runs after.
  *
  * @author Phillip Webb
- * @author Uladzislau Seuruk
  * @since 1.0.0
  */
 @Target({ ElementType.TYPE, ElementType.METHOD })
@@ -110,21 +109,5 @@ public @interface ConditionalOnBean {
 	 * @since 2.1.0
 	 */
 	Class<?>[] parameterizedContainer() default {};
-
-	/**
-	 * The class types of generic bean type arguments that should be checked. The
-	 * condition matches when beans of all classes specified are contained in the
-	 * {@link BeanFactory}.
-	 * @return the class types of beans to check
-	 */
-	Class<?>[] typeArguments() default {};
-
-	/**
-	 * The class type names of generic bean type arguments that should be checked. The
-	 * condition matches when beans of all classes specified are contained in the
-	 * {@link BeanFactory}.
-	 * @return the class type names of beans to check
-	 */
-	String[] typeArgumentNames() default {};
 
 }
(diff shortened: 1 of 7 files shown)


philwebb:
Thanks very much for the pull-request @VladislavSevruk. We discussed this today on our team call and we'd like to try and remove the `typeArguments` and `typeArgumentNames` attributes and instead detect the generics from the return type.

If you're interested in updating your pull-request in that direction, that would be great but please be aware that we're currently focusing on some Spring Boot 3.0 issues and this one is in the "general backlog" so we might not be able to review and merge things all that quickly.

No worries if you don't want to spend any additional time on this. We'll leave this issue open and mark it for `merge-with-amendments`.

VladislavSevruk:
Thank you for feedback @philwebb. I've updated code to auto-detect type arguments for generics from method return type.
But I also left ``typeArguments`` and ``typeArgumentNames`` attributes at annotations for cases when user uses ``type`` or ``value`` attributes.
Please let me know if I need to remove annotation attributes anyway.

snicoll:
@VladislavSevruk yes please. We really want to support that mode only for the return type.

VladislavSevruk:
@philwebb @snicoll the ``typeArguments`` and ``typeArgumentNames`` attributes were removed

snicoll:
Thanks a lot for the follow-up @VladislavSevruk. When merging this, we probably want to restore the existing tests using `ExampleBean` and create a separate test class with a generic as moving everything to generic doesn't seem right IMO.

VladislavSevruk:
@snicoll tests with ``ExampleBean`` are reverted to initial state, tests for generic return type are extracted to separate classes

mitasov-ra:
Any updates on when this PR will be merged?

Will it affect older versions of Spring Boot (pre 3.x) or is it just for 3.x?

wilkinsona:
@mitasov-ra The issue remains in the general backlog so things haven't changed since [this comment](https://github.com/spring-projects/spring-boot/pull/29500#issuecomment-1028576603). We consider the change to be an enhancement so it will only be made in a new minor (or major) release and won't affect older versions of Boot.

Bennett-Lynch:
Is this possible at all today? E.g., taking the example:

```
@ConditionalOnMissingBean(typeArguments = { String.class, Integer.class })
@Bean
public Converter<String, Integer> converter() {
    // ...
}
```

Is it possible to achieve this with the current `parameterizedContainer` behavior? E.g.,

```
@ConditionalOnMissingBean(value = { String.class, Integer.class }, parameterizedContainer = { Converter.class })
@Bean
public Converter<String, Integer> converter() {
    // ...
}
```

sheiy:
How is this going?

wilkinsona:
@sheiy No change since the last time someone asked. Any updates will appear here when we have them.

philwebb:
Thanks very much @VladislavSevruk, this has now been merged for 3.5.x. I needed to make a few changed to the original PR since the code has diverged a little since it was written.

PR #29516, opened 2022-01-20
Fix indentation in README.adoc

itsandreramon:
This PR only changes the indents in the README.adoc file because the current rendering looks weird on GitHub. The following images show the difference. I am using macOS Monterey 12.1 and Google Chrome 97.

Edit: Added screenshots (before, after)
Screenshot 2022-01-20 at 18 18 29
Screenshot 2022-01-20 at 18 18 56

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -24,24 +24,23 @@ Here is a quick teaser of a complete Spring Boot application in Java:
 
 [source,java,indent=0]
 ----
-	import org.springframework.boot.*;
-	import org.springframework.boot.autoconfigure.*;
-	import org.springframework.web.bind.annotation.*;
+import org.springframework.boot.*;
+import org.springframework.boot.autoconfigure.*;
+import org.springframework.web.bind.annotation.*;
 
-	@RestController
-	@SpringBootApplication
-	public class Example {
+@RestController
+@SpringBootApplication
+public class Example {
 
-		@RequestMapping("/")
-		String home() {
-			return "Hello World!";
-		}
+    @RequestMapping("/")
+    String home() {
+        return "Hello World!";
+    }
 
-		public static void main(String[] args) {
-			SpringApplication.run(Example.class, args);
-		}
-
-	}
+    public static void main(String[] args) {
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR. Other than tabs vs spaces (we prefer tabs), the indentation looks the same to me as what we currently have. Can you share some more details about how it looks weird to you, perhaps a screenshot and details of the browser that you're using?

itsandreramon:
Hey Andy, thanks for the quick response. I added two images (before, after) to the initial PR comment.

philwebb:
@itsandreramon What OS and web browser are you using?

itsandreramon:
I added the requested information at the top!

philwebb:
Thanks for the additional info @itsandreramon 

I'd like to keep the README file using tabs since I think it's more accessible. The good news is that I discovered GitHub has a setting that you can use to tweak the tab width. Take a look at [this blog post](https://github.blog/changelog/2021-09-21-tab-size-rendering-preference/). Changing "Tab size preference" to 4 in my [appearance settings](https://github.com/settings/appearance) makes things look better.

vpavic:
FWIW @philwebb, I've had the _Tab size preference_ setting set to 4 ever since the GitHub introduced the setting but I don't think that applies to the code snippets inside files that are otherwise plain text.

This is what it looks like for me in Firefox on Ubuntu:
Screenshot from 2022-01-21 09-20-52

philwebb:
Ohh, that's really annoying. I still feel like we should keep tabs in the README files. I wonder if wonder if GitHub might fix that sometime in the future? Looks like there are some extensions that might also work around the issue (e.g. https://addons.mozilla.org/en-US/firefox/addon/github-tab-render-size)

mahozad:
screenshot

@philwebb

I wanted to create a PR and "fix" the weird indentation.
I found this PR. From two years ago...

Could you please change from tabs to spaces in this code snippet?

wilkinsona:
I'm afraid not. Accessibility is as important to use now as it was two years ago so we want to continue using tabs.

PR #29525, opened 2022-01-21
Add support for JDK 18 to 2.7.x

aahlenst:
I assume that Spring Boot 2.7.0 is released around the same time as JDK 18. So I thought it would be nice if there were official support. I did the minimum necessary to get the tests to pass locally with a toolchain build. This includes upgrading to Gradle 7.0.2. Gradle 6.9 does not know about newer Java versions than 17 and accesses internal parts of the JDK.


wilkinsona:
Thanks for the PR, @aahlenst. Unfortunately, we'll want to take quite a different approach to adding JDK 18 to CI. For example, we'll support JDK 18 with 2.6.x (and possibly 2.5.x as well). We'll achieve this by using Gradle's Toolchain support rather than upgrading to a new version of Gradle. Thanks anyway.

We'll be in a position to consider adding CI for JDK 18 once https://github.com/adoptium/temurin18-binaries/releases has a stable, i.e. non-nightly, build available.

Issue #29550, opened 2022-01-24
Allow custom RSocket WebsocketServerSpecs to be defined


philwebb:
See https://github.com/spring-projects/spring-framework/issues/27973 for background.

The `RSocketWebSocketNettyRouteProvider` auto-configuration calls `HttpServerRoutes.ws` to create the `HttpServerRoutes`. It would be nice if the `WebsocketServerSpec` argument could be specified easily.

snicoll:
Closing in favor of PR #29567

PR #29567, opened 2022-01-26
Expose configuration properties to configure the  WebsocketServerSpec to use

polarbear567:
For issue [29550](https://github.com/spring-projects/spring-boot/issues/29550).
Preliminary submission.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/rsocket/RSocketProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/rsocket/RSocketProperties.java
@@ -73,6 +73,8 @@ public class RSocketProperties {
 		@NestedConfigurationProperty
 		private Ssl ssl;
 
+		private Spec spec = new Spec();
+
 		public Integer getPort() {
 			return this.port;
 		}
@@ -119,6 +121,74 @@ public class RSocketProperties {
 
 		public void setSsl(Ssl ssl) {
 			this.ssl = ssl;
+		}
+
+		public Spec getSpec() {
+			return this.spec;
+		}
+
+		public void setSpec(Spec spec) {
+			this.spec = spec;
+		}
+
+		public static class Spec {
+
+			/**
+			 * Sub-protocol to use in websocket handshake signature. Null by default.
+			 */
+			private String protocols;
+
+			/**
+			 * Specifies a custom maximum allowable frame payload length. 65536 by
+			 * default.
+			 */
+			private int maxFramePayloadLength;
+
+			/**
+			 * Flag whether to proxy websocket ping frames or respond to them. False by
+			 * default.
+			 */
+			private boolean handlePing;
+
+			/**
+			 * Flag whether the websocket compression extension is enabled if the client
+			 * request presents websocket extensions headers. By default compression is
+			 * disabled.
+			 */
+			private boolean compress;
+
+			public String getProtocols() {
+				return this.protocols;
+			}
+
+			public void setProtocols(String protocols) {
+				this.protocols = protocols;
+			}
+
+			public int getMaxFramePayloadLength() {
+				return this.maxFramePayloadLength;
+			}
+
+			public void setMaxFramePayloadLength(int maxFramePayloadLength) {
+				this.maxFramePayloadLength = maxFramePayloadLength;
+			}
+
+			public boolean isHandlePing() {
+				return this.handlePing;
+			}
+
+			public void setHandlePing(boolean handlePing) {
+				this.handlePing = handlePing;
+			}
+
+			public boolean isCompress() {
+				return this.compress;
+			}
+
+			public void setCompress(boolean compress) {
+				this.compress = compress;
+			}
+
 		}
 
 	}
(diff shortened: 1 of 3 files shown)


philwebb:
Thanks for the pull-request @polarbear567. When raising #29550 I was originally thinking we'd offer a way to do this programmatically. The properties are an interesting alternative that didn't occur to me and I'm now not sure which is best. I'm going to flag this for team discussion to see what other think.

wilkinsona:
I quite like the property-based approach. They do feel quite low-level but not so low that I think they ought to only be available programatically.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/rsocket/RSocketProperties.java
```diff
@@ -121,6 +123,74 @@ public void setSsl(Ssl ssl) {
...
+		}
+
+		public void setSpec(Spec spec) {
+			this.spec = spec;
+		}
+
+		public static class Spec {
+
+			/**
+			 * Sub-protocol to use in websocket handshake signature. Null by default.
```
snicoll, line 137: There's no need to write "by default" in the description as we harvest defaults and process them separately.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/rsocket/RSocketProperties.java
```diff
@@ -121,6 +123,74 @@ public void setSsl(Ssl ssl) {
...
+			 */
+			private String protocols;
+
+			/**
+			 * Specifies a custom maximum allowable frame payload length. 65536 by
+			 * default.
+			 */
+			private int maxFramePayloadLength;
```
snicoll, line 145: This field should be initialized  with `65536` then. Currently the advertised default is `0`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/rsocket/RSocketProperties.java
```diff
@@ -121,6 +123,74 @@ public void setSsl(Ssl ssl) {
...
+
+			/**
+			 * Flag whether to proxy websocket ping frames or respond to them. False by
+			 * default.
+			 */
+			private boolean handlePing;
```
snicoll, line 151: The second sentence should be removed as discussed above.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/rsocket/RSocketProperties.java
```diff
@@ -121,6 +123,74 @@ public void setSsl(Ssl ssl) {
...
+
+			/**
+			 * Flag whether the websocket compression extension is enabled if the client
+			 * request presents websocket extensions headers. By default compression is
```
snicoll, line 155: Ditto.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/rsocket/RSocketWebSocketNettyRouteProvider.java
```diff
@@ -42,19 +45,35 @@
...
-	RSocketWebSocketNettyRouteProvider(String mappingPath, SocketAcceptor socketAcceptor,
+	private final String protocols;
+
+	private final int maxFramePayloadLength;
+
+	private final boolean handlePing;
+
+	private final boolean compress;
+
+	RSocketWebSocketNettyRouteProvider(String mappingPath, Spec spec, SocketAcceptor socketAcceptor,
```
snicoll, line 56: There's no need to assign the fieds here. keeping the `Spec` object should be enough.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/rsocket/RSocketWebSocketNettyRouteProvider.java
```diff
@@ -42,19 +45,35 @@
...
 			Stream<RSocketServerCustomizer> customizers) {
 		this.mappingPath = mappingPath;
 		this.socketAcceptor = socketAcceptor;
 		this.customizers = customizers.collect(Collectors.toList());
+		this.protocols = spec.getProtocols();
+		this.maxFramePayloadLength = (spec.getMaxFramePayloadLength() <= 0) ? 65536 : spec.getMaxFramePayloadLength();
```
snicoll, line 62: This handling of the default here doesn't feel right to me.

snicoll (changes requested):
Thanks for the PR @polarbear567. We need tests that exercise those properties and validate they are mapped properly please. Would you have time to look at that?

polarbear567:
Sure, I'll take a look at it when I'm free. I'll modify PR according to comments. I don't have any good ideas for testing. Do you have any suggestions? @snicoll

snicoll:
> I don't have any good ideas for testing. Do you have any suggestions?

Yeah good point, I didn't look far enough. The mapping is hidden behind the predicate so it's hard to test this way. We can look ourselves as part of merging this, thanks for the update!

PR #29572, opened 2022-01-26
configure SpringOpaqueTokenIntrospector

lukaskusterbi:
Since Spring Security implemented their own SpringOpaqueTokenIntrospector (see https://github.com/spring-projects/spring-security/issues/9354), Spring Boot should configure that by default instead of relying on the com.nimbusds:oauth2-oidc-sdk dependency.

This PR replaces the NimbusOpaqueTokenIntrospector with SpringOpaqueTokenIntrospector in the default configuration, allowing applications to remove the com.nimbusds:oauth2-oidc-sdk dependency.


lukaskusterbi:

snicoll:
@lukaskusterbi unfortunately I don't think that qualifies as an obvious fix. I don't know what we're going to do with this PR but you'll need to sign the CLA if we decide to merge it.

lukaskusterbi:
OK, thanks for the heads-up. I'll need to check with Legal since it's not on our list of approved CLAs.

mbhave:
@jzheaux Is this what the Spring Security team recommends for Spring Boot auto-configuration?

lukaskusterbi:
Shall I adapt the configuration for `SpringReactiveOpaqueTokenIntrospector` as well? Or should that be done in a separate PR?

snicoll:
@lukaskusterbi if you have time, please go ahead!

PR #29592, opened 2022-01-29
Use CachedIntrospectionResults.IGNORE_BEANINFO_PROPERTY_NAME instead of "spring.beaninfo.ignore"

MangKyu:
I think there is no reason to use hardcoded "spring.beaninfo.ignore" instead of CachedIntrospectionResults.IGNORE_BEANINFO_PROPERTY_NAME


snicoll:
@MangKyu thank you for making your first contribution to Spring Boot.

PR #29596, opened 2022-01-29
Add ability to match Endpoint requests by HTTP method

onobc:
### What
This proposal adds the ability to match Actuator endpoint requests by http method. 

### Why
This provides a finer grained matching of requests. One example use is allowing unrestricted access to an endpoint for GET requests but restricting access to the endpoint for POST requests. 

### How
- Support has been added to both servlet and reactive flavors of EndpointRequest

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/security/servlet/EndpointRequest.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/security/servlet/EndpointRequest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2021 the original author or authors.
+ * Copyright 2012-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -40,6 +40,8 @@ import org.springframework.boot.security.servlet.ApplicationContextRequestMatche
 import org.springframework.boot.web.context.WebServerApplicationContext;
 import org.springframework.core.annotation.MergedAnnotation;
 import org.springframework.core.annotation.MergedAnnotations;
+import org.springframework.http.HttpMethod;
+import org.springframework.lang.Nullable;
 import org.springframework.security.web.util.matcher.AntPathRequestMatcher;
 import org.springframework.security.web.util.matcher.OrRequestMatcher;
 import org.springframework.security.web.util.matcher.RequestMatcher;
@@ -53,6 +55,7 @@ import org.springframework.web.context.WebApplicationContext;
  *
  * @author Madhura Bhave
  * @author Phillip Webb
+ * @author Chris Bono
  * @since 2.0.0
  */
 public final class EndpointRequest {
@@ -164,8 +167,8 @@ public final class EndpointRequest {
 		protected List<RequestMatcher> getLinksMatchers(RequestMatcherFactory requestMatcherFactory,
 				RequestMatcherProvider matcherProvider, String basePath) {
 			List<RequestMatcher> linksMatchers = new ArrayList<>();
-			linksMatchers.add(requestMatcherFactory.antPath(matcherProvider, basePath));
-			linksMatchers.add(requestMatcherFactory.antPath(matcherProvider, basePath, "/"));
+			linksMatchers.add(requestMatcherFactory.antPath(matcherProvider, null, basePath));
+			linksMatchers.add(requestMatcherFactory.antPath(matcherProvider, null, basePath, "/"));
 			return linksMatchers;
 		}
 
@@ -174,7 +177,7 @@ public final class EndpointRequest {
 				return context.getBean(RequestMatcherProvider.class);
 			}
 			catch (NoSuchBeanDefinitionException ex) {
-				return AntPathRequestMatcher::new;
+				return (pattern, method) -> new AntPathRequestMatcher(pattern, (method != null) ? method.name() : null);
 			}
 		}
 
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/security/servlet/EndpointRequest.java
```diff
@@ -191,38 +194,53 @@ protected RequestMatcherProvider getRequestMatcherProvider(WebApplicationContext
 
 		private final boolean includeLinks;
 
+		@Nullable
```
onobc, line 197: Re: `@Nullable`... You will notice I did **not** use it in the reactive counterpart - on purpose. I wanted to see what the recommendation from the team is in this area. I do not see it widely used in the SB codebase other than mostly on a few params to `@Endpoint` methods (selectors, params).  

Whatever direction we land I will make both imperative and reactive impls consistent.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/security/reactive/EndpointRequestIntegrationTests.java
```diff
@@ -0,0 +1,208 @@
...
+import org.springframework.boot.web.embedded.tomcat.TomcatReactiveWebServerFactory;
+import org.springframework.boot.web.reactive.context.AnnotationConfigReactiveWebServerApplicationContext;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.http.HttpMethod;
+import org.springframework.security.config.web.server.ServerHttpSecurity;
+import org.springframework.security.web.server.SecurityWebFilterChain;
+import org.springframework.test.web.reactive.server.WebTestClient;
+
+/**
```
onobc, line 50: This is net-new and is based on its servlet counterpart. I then added the coverage for the new matching on http method.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/servlet/AntPathRequestMatcherProvider.java
```diff
@@ -36,8 +39,9 @@ public AntPathRequestMatcherProvider(Function<String, String> pathFactory) {
 	}
 
 	@Override
-	public RequestMatcher getRequestMatcher(String pattern) {
-		return new AntPathRequestMatcher(this.pathFactory.apply(pattern));
+	public RequestMatcher getRequestMatcher(String pattern, @Nullable HttpMethod httpMethod) {
```
onobc, line 42: Again, w/ the `@Nullable`...

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/servlet/RequestMatcherProvider.java
```diff
@@ -33,6 +35,17 @@
...
+	}
+
+	/**
+	 * Return the {@link RequestMatcher} to be used for the specified pattern and http
+	 * method.
+	 * @param pattern the request pattern
+	 * @param httpMethod the http method
+	 * @return a request matcher
+	 */
+	RequestMatcher getRequestMatcher(String pattern, HttpMethod httpMethod);
```
onobc, line 49: I am not super-excited about this change. However, this seemed the least intrusive from a breakage standpoint. This leaves the other API in tact but it would break any previous lambda usage as the functional interface method now takes `(pattern, method)` vs. the previous `(pattern)`.

onobc:
We would also like to backport this to 2.6.x branch but am unsure if the changes to [RequestMatcherProvider](https://github.com/spring-projects/spring-boot/pull/29596#discussion_r795071859) would prevent that.

snicoll:
Thanks for the PR  Chris.

> We would also like to backport this to 2.6.x branch

@onobc we don't backport enhancement and this feels like one to me.

onobc:
> Thanks for the PR Chris.
> 
> > We would also like to backport this to 2.6.x branch
> 
> We don't backport enhancement and this feels like one to me.

TIL. I respect that sane policy. 

And yep, this is an enhancement which is can easily be functionally worked around via composition w/ a guard on http method such as:

```java
/**
 * Returns a matcher that includes the specified {@link Endpoint actuator endpoints} and http method.
 * @param httpMethod the http method to include
 * @param endpoints the endpoints to include
 * @return the configured {@link RequestMatcher}
 */
RequestMatcher to(HttpMethod httpMethod, String... endpoints) {
    final EndpointRequest.EndpointRequestMatcher matcher = EndpointRequest.to(endpoints);
    return (request) -> {
        if (httpMethod != null && !httpMethod.name().equals(request.getMethod())) {
            return false;
        }
        return matcher.matches(request);
    };
}
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/security/servlet/EndpointRequest.java
philwebb: We currently don't use them in our own codebase so we may as well strip them out for now. We've got #10712 open to add them consistently.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/servlet/RequestMatcherProvider.java
philwebb: I've been playing with this locally and I'll try to push something soon. I think this API is really in wrong place so we could deprecate it here and move it to `org.springframework.boot.actuate.autoconfigure.security.servlet` where it's closer to the actual thing that uses it.

philwebb:
I've pushed a polish commit to https://github.com/philwebb/spring-boot/tree/gh-29596.

I'd like someone else from the team to review. I'm wondering if we should include `HttpMethod` in the API or not. It's in `spring-web` and I wonder if it will always be available. Interestingly, the `AntPathRequestMatcher` uses Strings in the API but internally does rely on `HttpMethod`. I suspect that means we're fine to have it in out API.

snicoll:
Given how the endpoint abstraction is designed at the moment, wouldn't it make more sense to provide the kind of operation that is requested on the endpoint? `EndpointRequest` we use an endpoint concept there so it feels natural to me to extend it to customize the "kind of operation" (read/write/delete).

wilkinsona:
I agree that's natural for standard endpoints but I think it may be confusing for `@ControllerEndpoint` where you deal with HTTP methods directly.

onobc:
> I'm wondering if we should include HttpMethod in the API or not. It's in spring-web and I wonder if it will always be available. Interestingly, the AntPathRequestMatcher uses Strings in the API but internally does rely on HttpMethod. I suspect that means we're fine to have it in out API.

Yep, I saw this as well @philwebb . I ended up keeping HttpMethod until we go down into the `AntPathRequestMatcher`.  It is also used in `org.springframework.security.config.annotation.web.builders.HttpSecurity.RequestMatcherConfigurer#mvcMatchers(org.springframework.http.HttpMethod, java.lang.String...)` which is where these matchers will be primarily used.

onobc:
> Given how the endpoint abstraction is designed at the moment, wouldn't it make more sense to provide the kind of operation that is requested on the endpoint? EndpointRequest we use an endpoint concept there so it feels natural to me to extend it to customize the "kind of operation" (read/write/delete).

I thought about this as well @snicoll . I ended up putting on the EndpointRequestMatcher directly to keep the functionality isolated to this particular case where the matchers are being used in the security mappings and where http methods are used.

onobc:
Strange failure in CI which I believe is completely unrelated 
```
spring-boot-project:spring-boot-autoconfigure:test
    org.springframework.boot.autoconfigure.liquibase.LiquibaseAutoConfigurationTests > overrideDataSourceAndDriverClassName()
```
I am rebased so not sure what could be causing it other than flaky test?

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/security/reactive/EndpointRequestIntegrationTests.java
mbhave: I think these tests could inherit from `AbstractEndpointRequestIntegrationTests` too. I think that's why there was an `abstract` class and looks like an oversight on our part that there wasn't a reactive implementation before.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/security/servlet/EndpointRequest.java
onobc: Will you handle that in your polish commit @philwebb ? I don't mind doing it - just let me know.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/security/reactive/EndpointRequestIntegrationTests.java
onobc: > I think these tests could inherit from AbstractEndpointRequestIntegrationTests too.

I believe AERIT has servlet specific things in it though. 

> I think that's why there was an abstract class

I believe it was there to [test](https://github.com/spring-projects/spring-boot/blob/2.5.x/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/security/servlet/JerseyEndpointRequestIntegrationTests.java) the previous Jersey-flavored `RequestMatcherProvider` impl that has since been removed in 2.6.3.

csterwa:
@philwebb was this merged in a different PR?

philwebb:
@csterwa No, I don't believe this has been merged in any form as yet.

philwebb:
Thanks for your patience @onobc, this has been merged for 3.5.

PR #29636, opened 2022-02-03
Do not set ConversionService on AnnotatedControllerConfigurer

rstoyanchev:
After https://github.com/spring-projects/spring-graphql/commit/0b449d89e1d9e8754d42ad5e5af7f780c705ce32, the `ConversionService` on `AnnotatedControllerConfigurer` is an internally managed instance that is customized with `FormatterRegistrar` callbacks rather than set.


bclozel:
To be merged before #29638 is addressed.

Issue #27003, opened 2021-06-21
Docs appendix is at wrong level


philwebb:
Appendix sections are all at the top level. They should probably move under the `Appendix` title

mhalbritter:
I tried to fix that by adjusting the `include::application-properties.adoc[leveloffset=+1]` to `include::application-properties.adoc[leveloffset=+2]`. This works, but the https://github.com/spring-io/spring-asciidoctor-extensions/tree/main/section-ids asciidoctor plugin, which checks that section ids in child sections adhere to the `parentId.childId` convention logs a warning.

The section id of the "Appendices" is `appendix`, but the section id of the application-properties.adoc isn't `appendix.application-properties`, but only `application-properties`. The application-properties.adoc includes more files like application-properties/core.adoc, which gets autogenerated and has the wrong section ids, too.

snicoll:
Closing in favor of PR #29667

PR #29667, opened 2022-02-07
Move appendix subsections under appendix section

mhalbritter:
This involved a small code change to the generated configuration properties snippets: the section id has to start with 'appendix.', otherwise the section-id asciidoctor extension complains.

See #27003


philwebb:
I like the `appending` prefix on the section titles, but we should update `anchor-rewrite.properties`[https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/docs/asciidoc/anchor-rewrite.properties] so that older links work.

the format of that file is pretty self explanatory, but it's also covered [here](https://github.com/spring-io/spring-asciidoctor-backends#anchor-rewriting)

mhalbritter:
Thanks for the hint, I will update that file.

mhalbritter:
Updated the anchors file, anchors are now backward compatible.

Issue #29307, opened 2022-01-11
build-image fails sometimes with ArrayIndexOutOfBoundsException


nils-christian:
Hi,

We are experiencing an _ArrayIndexOutOfBoundsException_ in some very rare cases while building our image with the _spring-boot-maven-plugin_. We are using Spring Boot 2.4.6. Unfortunately we cannot provide an example project, because the error cannot be reproduced. It happens in something like 1 of 50 builds. Can we somehow do something for a better analysis? Is there a possible root cause?

Best regards

  Nils

```
10:05:11  org.apache.maven.lifecycle.LifecycleExecutionException: Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.4.6:build-image (default) on project XYZ: Execution default of goal org.springframework.boot:spring-boot-maven-plugin:2.4.6:build-image failed: Index 3 out of bounds for length 3
10:05:11      at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:215)
10:05:11      at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:156)
10:05:11      at org.apache.maven.lifecycle.internal.MojoExecutor.execute (MojoExecutor.java:148)
10:05:11      at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:117)
10:05:11      at org.apache.maven.lifecycle.internal.LifecycleModuleBuilder.buildProject (LifecycleModuleBuilder.java:81)
10:05:11      at org.apache.maven.lifecycle.internal.builder.singlethreaded.SingleThreadedBuilder.build (SingleThreadedBuilder.java:56)
10:05:11      at org.apache.maven.lifecycle.internal.LifecycleStarter.execute (LifecycleStarter.java:128)
10:05:11      at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:305)
10:05:11      at org.apache.maven.DefaultMaven.doExecute (DefaultMaven.java:192)
10:05:11      at org.apache.maven.DefaultMaven.execute (DefaultMaven.java:105)
10:05:11      at org.apache.maven.cli.MavenCli.execute (MavenCli.java:957)
10:05:11      at org.apache.maven.cli.MavenCli.doMain (MavenCli.java:289)
10:05:11      at org.apache.maven.cli.MavenCli.main (MavenCli.java:193)
[...]

PR #29675, opened 2022-02-08
Ignore invalid stream types when reading log update events

mhalbritter:
See gh-29307


spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/LogUpdateEvent.java
```diff
@@ -90,7 +90,15 @@ private static LogUpdateEvent read(InputStream inputStream) throws IOException {
 		if (header == null) {
 			return null;
 		}
-		StreamType streamType = StreamType.values()[header[0]];
+
+		// First byte denotes stream type. 0 = stdin, 1 = stdout, 2 = stderr
+		byte streamTypeId = header[0];
+		if (streamTypeId < 0 || streamTypeId >= StreamType.values().length) {
+			// Stream type is out of bounds, ignore this event. See gh-29307
```
mhalbritter, line 97: I thought about logging a warning here, but I didn't found any logging infrastructure. Any thoughts on that?

philwebb: It's not particularly easy. There's a `BuildLog` interface that's used to do all the logging. I think Gradle outputs it to a `PrintStream` and maven has a `MojoBuildLog` class that send the text to the standard maven logger. Probably the easiest option here is to slightly abuse the `LogUpdateEvent`. I think you could do perhaps throw an exception from `read` and catch it in `readAll` then construct a `LogEvent` with a sensible message and pass that to the `consumer` before exiting.

mhalbritter: Will do that, thanks!

Issue #29676, opened 2022-02-08
Dependency management for Liquibase does not include its liquibase-cdi module


wilkinsona:
While using CDI in a Boot app is unusual, for completeness we should include `liquibase-cdi` in `spring-boot-dependencies`.

PR #29736, opened 2022-02-11
include liquibase-cdi in spring-boot-dependencies.

polarbear567:
fix issue [29676](https://github.com/spring-projects/spring-boot/issues/29676)


wilkinsona:
Thanks for the PR @polarbear567, but for an obvious single-line change such as this that we've already identified, we prefer to take care of it ourselves. The overhead of processing the PR, particularly when the change belongs on a different branch, isn't really worth it. Thanks anyway.

Issue #28181, opened 2021-10-02
Use pattern matching for instanceof where appropriate


wilkinsona:
```java
if (obj instanceof String) {
    String s = (String) obj;
    // use s
}
```

becomes

```java
if (obj instanceof String s) {
    // use s
}
```

wilkinsona:
Not at the moment. Thanks for the offer, though. We haven't started work on 3.0 yet and there are a number of things that'll have to be tackled before we can look at this one.

wilkinsona:
Closing in favor of #31475.

PR #29737, opened 2022-02-11
Apply "instanceof pattern matching"

xixingya:
use instanceof that new jdk feature


wilkinsona:
Thanks very much for the pull-request, unfortunately we're not yet ready to accept Java 17 updates. We're tracking this one in https://github.com/spring-projects/spring-boot/issues/28181 which is currently marked as blocked.

Duplicates #29253.

PR #29758, opened 2022-02-13
Link directly to the Integration Properties section of the appendix when cross-referencing Kafka properties

dreis2211:
Hi,

this PR fixes a link in the docs, which leads to a 404 at the moment.
I also took the freedom to polish this a bit more and align this with other sections where we link to the application.properties Appendix.

## Before
image

## After
image

Cheers,
Christoph


wilkinsona:
Thanks for spotting this, @dreis2211. While the link's only broken in 3.0 snapshots, I think it'd be nice to link to a more specific part of the appending in all our maintained branches. We can then fix the broken link as part of merging that change forwards.

PR #29762, opened 2022-02-14
Gradle Plugin triggers eager configuration of some tasks

### LINKED ISSUE #29681: Remove code that provides compatibility with unsupported versions of Gradle
We have some code in our Gradle plugin that provides compatibility with unsupported versions of Gradle. For example, in 2.5.x the minimum supported version of Gradle is 6.8 yet `ApplicationPluginAction` contains some logic that checks for Gradle 6.4. Similarly, `JavaPluginAction` contains some logic that checks for Gradle 6.7.

larsgrefer:
These changes should avoid the eager creation of some Gradle tasks, which were caused by the Plugin.
This also fixes #29681

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/plugin/ApplicationPluginAction.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/plugin/ApplicationPluginAction.java
@@ -31,6 +31,7 @@ import org.gradle.api.file.CopySpec;
 import org.gradle.api.file.FileCollection;
 import org.gradle.api.plugins.ApplicationPlugin;
 import org.gradle.api.plugins.ApplicationPluginConvention;
+import org.gradle.api.tasks.TaskProvider;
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/application/CreateBootStartScripts.java
```diff
@@ -1,37 +0,0 @@
...
-import org.gradle.jvm.application.tasks.CreateStartScripts;
-
-/**
- * Customization of {@link CreateStartScripts} that makes the {@link #getMainClassName()
- * main class name} optional.
- *
- * @author Andy Wilkinson
- * @since 2.0.0
- */
-public class CreateBootStartScripts extends CreateStartScripts {
```
wilkinsona, line 29: This is a breaking change to public API. It's also unrelated to task configuration avoidance. Can you please revert the changes for #29681 and we'll tackle those separately.

wilkinsona (changes requested):

wilkinsona:
Thanks very much, @larsgrefer.

PR #29812, opened 2022-02-15
Add auto-configuration for Kafka Retry Topics

### LINKED ISSUE #28450: Add AutoConfiguration to Spring Kafka's Retry Topic Feature
This is a proposal for enabling auto configuration for the delayed non-blocking retries using topics feature that we introduced earlier this year on Spring for Apache Kafka 2.7.0.

More details on the feature can be found in the documentation: https://docs.spring.io/spring-kafka/docs/current/reference/html/#retry-topic

I already have a working draft with tests and should open a PR shortly so we can discuss it in more details if necessary.

The minimal configuration would be as simple as adding `spring.kafka.retry-topic-enabled=true` to the application yaml or properties configurations file, which would apply a default global retry configuration to all topics.

```
spring:
  kafka:
    retry-topic-enabled: true
```

With that minimal configuration, the framework would set up topics and consumers with the following defaults:

- Retries for all topics
- 3 processing attempts before sending the message to the DLT
- FixedBackOffPolicy of 1 second between attempts
- Default topic suffixes of "-retry-1", "-retry-2", "-dlt"
- Create the retry topics and dlts automatically via NewTopic instances
- Retry on all exceptions
- Use a dedicated default ConcurrentKafkaListenerContainerFactory instance provided by the auto configuration specifically for creating the retry topics and dlt consumers
- Use the default Spring Kafka's KafkaTemplate from auto configuration for forwarding the messages to the topics

More specific configurations could also be provided in the configuration files, such as:

```
spring:
  kafka:
    retry-topic-enabled: true
    retry-topic:
      myGlobalConfiguration:
        attempts: 4
        back-off:
          delay: 300s
          multiplier: 2
          max-delay: 1200s
        topic-suffixing-strategy: suffix-with-index-value
        dlt-strategy: fail-on-error
```

Optionally, more than one configuration could be provided for more fine-grained control over the configuration:
```
spring:
  kafka:
    retry-topic-enabled: true
    retry-topic:
      myGlobalConfiguration:
        attempts: 4
        back-off:
          delay: 300s
          multiplier: 2
          max-delay: 1200s
        topic-suffixing-strategy: suffix-with-index-value
        dlt-strategy: fail-on-error
        exclude-topics: my-specific-topic
     mySpecificConfiguration:
        include-topics: my-specific-topic
        attempts: 5
        back-off:
          delay: 5s
        not-retry-on:
          - com.mycompany.myproject.mypackage.exceptions.MyException
          - com.mycompany.myproject.mypackage.exceptions.MyOtherException
        topic-suffixing-strategy: suffix-with-index-value
        dlt-strategy: fail-on-error
```

Note: If it gets too verbose for many topic configurations the user should be able to provide that in one or more separate configuration files using standard spring boot configuration features such as `spring.config.import`.

I'd appreciate any feedback you may have on this. Also please let me know if there are any questions or anything I can help with.

@garyrussell and @artembilan, I'd much appreciate your feedback too if and when possible.

Thank you all for the hard work you put into the awesome Spring projects!

tomazfernandes:
Closes #28450

This PR adds basic auto-configuration capabilities for Spring Kafka's non-blocking delayed retries feature (https://docs.spring.io/spring-kafka/reference/html/#retry-topic).

I chose not to add exception type classification to it just now as was previously suggested by @garyrussell so I can get feedback on the basic solution first. I can add exception type classification in a separate PR if and when this one is approved.

I wasn't sure whether to add an entry to the Messaging section of the documentation, since it seems to cover more basic use cases. I can do so if that's the case.

Seems like we should add the new properties to the A.8. Integration Properties section of the documentation, but I couldn't really find it in the source code. I can add it if you point me to it.

Please feel free to request any changes you see fit, or to make a polish commit on top of it.

As always, If @garyrussell and @artembilan can take a look into this it'd be great.

Thanks


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -93,6 +94,8 @@
 
 	private final Security security = new Security();
 
+	private final RetryTopic retryTopic = new RetryTopic();
```
snicoll, line 97: This would create a `retry-topic` group and we don't use dashes in group parts. Perhaps a `Retry` inner type with a `Topic` if that makes sense to separate them. This would give `spring.kafka.retry.topic`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfigurationUtils.java
```diff
@@ -0,0 +1,62 @@
...
+ * @author Tomaz Fernandes
+ * @since 2.7.0
+ */
+public final class KafkaAutoConfigurationUtils {
+
+	private KafkaAutoConfigurationUtils() {
+
+	}
+
+	static SleepingBackOffPolicy<?> createBackOffFrom(Long min, Long max, Double multiplier, Boolean isRandom) {
```
snicoll, line 37: Spring Boot doesn't do utils class like this. It looks like this is an internal detail anyway so I'd rather make that a package protected class?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
```diff
@@ -113,6 +118,20 @@ public KafkaAutoConfiguration(KafkaProperties properties) {
 		return new KafkaTransactionManager<>(producerFactory);
 	}
 
+	@Bean
+	@ConditionalOnProperty(name = "spring.kafka.retry-topic.enabled")
```
snicoll, line 122: This does not have any metadata at the moment and it should. See `META-INF/additional-spring-configuration-metadata.properties` for other enabled flags

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -1314,6 +1321,82 @@ public void setOptions(Map<String, String> options) {
 
 	}
 
+	public static class RetryTopic {
+
+		private Integer attempts;
```
snicoll, line 1326: This should have field Javadoc like other properties in this class. Please review the to phrase it consistently.

snicoll (commented):
Thanks for the PR. I've made some suggestions for your consideration.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
```diff
@@ -113,6 +118,20 @@ public KafkaAutoConfiguration(KafkaProperties properties) {
...
+	@ConditionalOnMissingBean
+	public RetryTopicConfiguration kafkaRetryTopicConfiguration(KafkaOperations<Object, Object> kafkaOperations) {
+		RetryTopicConfigurationBuilder builder = RetryTopicConfigurationBuilder.newInstance();
+		KafkaProperties.RetryTopic retryTopic = this.properties.getRetryTopic();
+		JavaUtils.INSTANCE.acceptIfNotNull(retryTopic.getAttempts(), builder::maxAttempts).acceptIfNotNull(
```
artembilan, line 127: I doubt this util from spring-kafka is acceptable in Spring Boot code.
See other similar auto-configuration classes which use something like this: `PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();`
The `IntegrationAutoConfiguration` is good one to look into.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfigurationUtils.java
artembilan: +1 from me.
This one is only for specific auto-configuration, so better to have it as a `private static` method over there.
See for example what we have in the `IntegrationAutoConfiguration` for `defaultPollerMetadata` bean definition.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
artembilan: On the other hand why this `enabled` is not a property in that `KafkaProperties.RetryTopic` ?

artembilan (changes requested):
> add the new properties to the A.8. Integration Properties section of the documentation

That part of the doc is auto-generated by source code. So, whatever you have added to `KafkaProperties` is going to be reflected over there as well.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
tomazfernandes: Sure, thanks. I didn't add it there since it's only used in the `@ConditionalOnProperty` annotation and not in the actual configuration code. I'll add it then. I understand that by doing so there's no need for the META-INF properties file entry.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
tomazfernandes: The `PropertyMapper` class is great, thanks!

tomazfernandes:
Thank you very much for the reviews. I've made the changes as suggested.

I removed the BackOff inner class since it'd have the same problem of creating a group with a dash in the name. Also, most of the BackOff creation logic and JavaDocs are from the Spring Retry project, not sure if we should add proper credit there somehow.

Let me know if there are any other suggestions or anything else to change.

Thanks.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
```diff
@@ -137,4 +146,48 @@ public KafkaAdmin kafkaAdmin() {
...
+		map.from(createBackOffFor(retryTopic)).to(builder::customBackoff);
+		return builder.create(kafkaOperations);
+	}
+
+	private SleepingBackOffPolicy<?> createBackOffFor(KafkaProperties.Retry.Topic properties) {
+		Long min = properties.getDelayMillis();
+		Long max = properties.getMaxDelayMillis();
+		Double multiplier = properties.getMultiplier();
+		Boolean isRandom = properties.isRandomBackOff();
+		if (min == null && max == null && multiplier == null && isRandom == null) {
```
wilkinsona, line 166: This doesn't look quite right to me. If random is non-null, it'll cause a `FixedBackOffPolicy` to be returned but the random setting will never be used. I think the `isRandom == null` check could just be removed here. WDYT, @tomazfernandes?

tomazfernandes: Sure, we can do that, thanks for pointing it out. In the end, the result will also be a `FixedBackOffPolicy`, but the way you're proposing makes the behavior clearer -  that we're in fact ignoring the isRandom flag.

Actually, I'd like to ask you a question about this, if that's ok. This logic was mainly brought from the [Spring Retry](https://github.com/spring-projects/spring-retry/blob/302f7999932421d39800b81b5fa9f6b1c557b957/src/main/java/org/springframework/retry/annotation/AnnotationAwareRetryOperationsInterceptor.java#L404) project, but it's private there as you can see in the link, so we have a slightly modified copy in the [Spring Kafka](https://github.com/spring-projects/spring-kafka/blob/43cb78127b3851c03037418533bf393c2fc7db3a/spring-kafka/src/main/java/org/springframework/kafka/annotation/RetryableTopicAnnotationProcessor.java#L171
) project, but it's not accessible there either, that's why I made another copy here.

I'm wondering if maybe it'd be better to open up this logic in Spring Kafka, and use it here, so we can keep the exact same behavior whether the user is configuring through auto configuration or through the `@RetryableTopic` annotation. I've been meaning to review a few things there too - maybe let the user set a `delay` of 0 and get a `NoBackOffPolicy` for example, besides a couple of issues I've noticed recently.

That might hold this feature back a while though, until SK with this change is released, and Boot is updated to build against that version. And since we won't make any breaking changes there too, maybe it's something we can change in a later version.

WDYT?

Thanks

wilkinsona: I think that's a great idea as it'll avoid any drift between the two projects with how the four settings are mapping to a policy. Let's wait for that API to appear in Spring Kafka and then we can update the proposal here to use it.

tomazfernandes: Ok, sounds good, I’ll open an issue for this change there. Thanks @wilkinsona.

tomazfernandes: Hi @wilkinsona, we addressed this issue [here](https://github.com/spring-projects/spring-retry/pull/277) in the `Spring Retry` project.

Not sure how the release schedule is there now, but I'll update this PR when the API appears.

Thanks.

wilkinsona: @tomazfernandes The changes in Spring Retry should now be available in 1.3.3-SNAPSHOT. Do you have time to have another look at this before next week's Spring Boot 2.7.0-RC1 release?

tomazfernandes: Hi @wilkinsona, that's good news, sure, thanks for letting me know. I should have these changes soon.

tomazfernandes: @wilkinsona it seems [build.gradle](https://github.com/spring-projects/spring-boot/blob/3485070aeef0b5ebff0cf4dfcc76381a82cf4f6a/spring-boot-project/spring-boot-dependencies/build.gradle#L1421-L1427) is still pointing to `1.3.2`.

Is it going to be changed to `1.3.3-SNAPSHOT`, or should I do that as part of this PR?

Thanks

wilkinsona: Please change it in the PR for now. We can straighten things out as part of the merge.

snicoll: @tomazfernandes this PR targets the `2.7.x` branch and we've switched to snapshot there https://github.com/spring-projects/spring-boot/blob/f96efa72abbc8ab4abd6f5378ebc9574a84c30b1/spring-boot-project/spring-boot-dependencies/build.gradle#L1783-L1787

tomazfernandes: Also, I don't seem to be able to cleanly either rebase my branch onto origin/2.7.x., or merge that branch into mine. Do you have any pointers on that? Thanks again

wilkinsona: We prefer forward merges to back-ports, however, if switching the base branch to 2.7.x is proving tricky, please leave it as main and switch to Spring Retry 1.3.3-SNAPSHOT in your PR. We can then take care of getting the changes based on top of 2.7.x as part of merging the contribution.

tomazfernandes: I've submitted the changes. I think this looks a lot better overall than the previous code, thanks mainly to the great guidance I've been receiving. Please let me know if there's anything you'd like me to change.

Besides using the `BackOffPolicyBuilder`, I've also changed a few things in the `sensible defaults`, more on par with what I've came to understand about the auto configuration role.

Now we have:

`useSingleTopicForFixedDelays()`
`suffixTopicsWithIndexValues()`
`doNotAutoCreateRetryTopics()`

With these settings, by default users will get 3 topics, independently of the number of `attempts` or `delay` value:
`myTopic`, `myTopic-retry` and `myTopic-dlt`

If a `multiplier` or `maxDelay` is set, yelding multiple delay values, users will get one topic per attempt plus the dlt:
`myTopic`, `myTopic-retry-0`, `myTopic-retry-1`, `myTopic-retry-2` ... `myTopic-dlt`

A Kafka topic can only handle a single delay value due to message ordering - if delays were different within a topic, we couldn't tell if a record after the one being processed was supposed to be processed earlier.

Note that although we're setting the feature not to create the topics, Kafka's `Consumer` can still auto create them if configured to do so.

@garyrussell and @artembilan, if possible, please let me know if you agree with these defaults and with the code changes.

We can add more features to the auto configuration as users request them, but if there's anything that comes to mind that should be in this first version I can add them.

Thanks, looking forward to seeing this in place!

tomazfernandes: @wilkinsona, there's a conflict on the `build.gradle` file regarding the `spring-retry` version. Should I try to resolve the conflict here on GitHub, or is it something you should do? It would be interesting so we can see the test results.

Thanks

tomazfernandes: Also, sorry for my not-so-great git skills - this is a complex project and I couldn't solve this in a proper way with the solutions I know. I appreciate the assistance.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
```diff
@@ -137,4 +146,22 @@ public KafkaAdmin kafkaAdmin() {
...
+				.customBackoff(getBackOffPolicy(retryTopic)).useSingleTopicForFixedDelays()
+				.suffixTopicsWithIndexValues().doNotAutoCreateRetryTopics().create(kafkaOperations);
+	}
+
+	private SleepingBackOffPolicy<?> getBackOffPolicy(Topic retryTopic) {
+		BackOffPolicy policy = BackOffPolicyBuilder.newBuilder().delay(retryTopic.getDelayMillis())
+				.maxDelay(retryTopic.getMaxDelayMillis()).multiplier(retryTopic.getMultiplier())
+				.random(retryTopic.isRandomBackOff()).build();
+		Assert.isInstanceOf(SleepingBackOffPolicy.class, policy,
+				() -> "BackOffPolicy must be an instance of SleepingBackOffPolicy. Provided: " + policy);
```
artembilan, line 163: I don't think it is OK to say this: the maximum what end-user can provide are exactly those props you use for `BackOffPolicyBuilder`.
So, we might need to ensure that only `SleepingBackOffPolicy` is created. Perhaps just ignore some props which are not sufficient...

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
```diff
@@ -137,4 +146,22 @@ public KafkaAdmin kafkaAdmin() {
...
+	@Bean
+	@ConditionalOnProperty(name = "spring.kafka.retry.topic.enabled")
+	public RetryTopicConfiguration kafkaRetryTopicConfiguration(KafkaOperations<Object, Object> kafkaOperations) {
+		KafkaProperties.Retry.Topic retryTopic = this.properties.getRetry().getTopic();
+		return RetryTopicConfigurationBuilder.newInstance().maxAttempts(retryTopic.getAttempts())
+				.customBackoff(getBackOffPolicy(retryTopic)).useSingleTopicForFixedDelays()
+				.suffixTopicsWithIndexValues().doNotAutoCreateRetryTopics().create(kafkaOperations);
+	}
+
+	private SleepingBackOffPolicy<?> getBackOffPolicy(Topic retryTopic) {
```
artembilan, line 158: `static`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -1314,6 +1321,119 @@ public void setOptions(Map<String, String> options) {
...
+			/**
+			 * The maximimum wait between retries. If less than the delay then the default
+			 * of 30 seconds is applied.
+			 */
+			private Duration maxDelay = Duration.ZERO;
+
+			/**
+			 * In the exponential case (multiplier() > 0) set this to true to have the
+			 * backoff delays randomized, so that the maximum delay is multiplier times
+			 * the previous delay and the distribution is uniform between the two values.
```
artembilan, line 1373: A description for these configuration properties must be as short as possible: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#appendix.configuration-metadata.format.property

artembilan (changes requested):

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
tomazfernandes: Thanks for looking into this @artembilan. Currently, the `BackOffPolicyBuilder` only returns instances of `SleepingBackOffPolicy`, but I didn't want to add this constraint to the builder's method signature.

So this assertion here is just for the case that, if we in the future update the `Builder` to for example return a `NoBackOffPolicy` if delay is set to `0`, we don't get a class cast exception, but get this error instead.

But I agree that this is not a really useful message to the user. Maybe we should just blindly cast? Not sure what a useful message for the users should be in this case, since it'd not really be their fault.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
tomazfernandes: Sure. This description was extracted from the `@BackOff` annotation, I'll try to come up with something more succinct. Thanks.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
artembilan: If we don't support a `delay == 0`, then it has to be rejected as a configuration properties validation, before calling this builder.
What I mean, if we don't support anything else in the builder, then we need to ensure the props provided for it are in the valid range.
The message you say now is fully misleading: end-user does not create instances of this, so the message (if any), must really point what is wrong with end-user config.

tomazfernandes: Well, currently if the provided delay < 1 it is set to 1 in the `FixedBackOffPolicy`, and most or all other `BackOffPolicy` implementations follow that pattern. In `RetryTopic`'s annotation processing and builder I chose to mimic `Spring Retry` behavior as much as possible, since we relied on the same `@BackOff` annotation.

I've been meaning to ask if maybe we should change this in `Spring Retry` to return the `NoBackOffPolicy` instead if `delay=0` is provided, but didn't want to mix things up. Also, that seems like a breaking change, so it should be a good fit only for `SR` 2.0.

Maybe it'd be worth it to validate user input here to make sure the inputs are > or >=  0. Although it might be confusing having different behaviors for auto configuration and the `@RetryableTopic` annotation and builder.

artembilan: So, why then does `RetryTopicConfigurationBuilder` support only a `SleepingBackOffPolicy`?
Why it cannot be that super `BackOffPolicy`?

artembilan (changes requested):

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
tomazfernandes: That's because the way we [generate the delay values](https://github.com/spring-projects/spring-kafka/blob/8ccdbd7a8a54815382645a8e9dd30cd82c0334dc/spring-kafka/src/main/java/org/springframework/kafka/retrytopic/BackOffValuesGenerator.java#L78-L92
) is we inject a `Sleeper` implementation in the `BackOffPolicy` that calls `backOff` and stores the values, so we need that any custom policy the user provides respect the `SleepingBackOffPolicy` contract.

The builder offers a `noBackOff()` method if that's what the user desires, and if that's the case we don't need the `Sleeper` to fetch the values.

artembilan: I see there `public RetryTopicConfigurationBuilder noBackoff() {`, so perhaps with some props config we can fall back to this instead of `customBackoff()` then?

tomazfernandes: You mean, if users set `delay=0` we could set `noBackOff()` for them there? That's an interesting thought. My concern would be having different behaviors in auto configuration and annotation processing - would that be confusing for users? I guess since these are kind of corner cases, maybe we can depart from that and also add properties validation.

WDYT?

artembilan: Yes, that would be reasonable.

Thanks

tomazfernandes: Sure, I’ll add that then. Thanks a lot for your input!

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
tomazfernandes: I solved the git issue by checking out the 2.7.x brach, cherry-picking my commits, and hard reseting this branch to that. No clue on why I couldn't do a clean merge or rebase though - have you rebased this branch at some point maybe?

Also, I don't seem to be able to do a clean build locally, there's always a failing test. I've tried different JDK versions / implementations, cleaned up my Gradle caches, and though the failing tests change, the build never gets through cleanly. Maybe my computer is not powerful enough to handle such intensive testing. If you have any pointers on that it'd be great, otherwise I'll keep trying.

The tests in the `kafka` package are ok when running locally though, so it should be ok here.

Thanks!

tomazfernandes:
Implemented @artembilan's suggestions. 

I looked into using `@Valid` to validate the input, but since a couple properties are `Duration` and there doesn't seem to be a way to properly validate since annotations don't allow such value types, I preferred to keep it simple and validate all properties together in the code.

Please let me know if there's a better way or anything else to change.

Thanks

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
tomazfernandes: > @tomazfernandes this PR targets the `2.7.x` branch and we've switched to snapshot there
> 
> https://github.com/spring-projects/spring-boot/blob/f96efa72abbc8ab4abd6f5378ebc9574a84c30b1/spring-boot-project/spring-boot-dependencies/build.gradle#L1783-L1787

I've changed the PR to target 2.7.x. - hope it's ok now.

Thanks

tomazfernandes:
I've cleaned up the validation logic and added tests for that. Please let me know if there's anything else. Thanks!

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -1447,6 +1454,23 @@ public void setRandomBackOff(Boolean randomBackOff) {
 				this.randomBackOff = randomBackOff;
 			}
 
+			private Topic validate() {
```
artembilan, line 1457: I won't mind, but I don't know Spring Boot development process in details, so I'll defer a decision for a proper validation and its place to respective Spring Boot team members.
Thanks

artembilan (changes requested):

tomazfernandes:
FYI, this unrelated test failed in the PR build:

```java
Found test failures in 1 test task:
[00:00:08](https://ci.spring.io/builds/185388#L62321ef2:5476)

[00:00:08](https://ci.spring.io/builds/185388#L62321ef2:5477)
:spring-boot-project:spring-boot-devtools:test
[00:00:08](https://ci.spring.io/builds/185388#L62321ef2:5478)
    org.springframework.boot.devtools.tunnel.server.HttpTunnelServerTests > disconnectTimeout()`
```

I'll force-push a no-ops commit to trigger the PR build again. Thanks.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
tomazfernandes: Sure, thanks. Just for reference, it seems we already have some validation in this `buildProperties` method of the `KafkaProperties` class, so I thought it seemed ok there. Of course, I'm ok with validating any other way or place.

https://github.com/spring-projects/spring-boot/blob/c58e24abd9a415f7811dfc7bc3730087df23dd15/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java#L1216-L1218

Thanks for looking into this.

tomazfernandes:
@garyrussell, if and when possible, I'd like your input on the chosen `sensible defaults`, since it's something not so easy to change later.

> Now we have:
>
> useSingleTopicForFixedDelays()
> suffixTopicsWithIndexValues()
> doNotAutoCreateRetryTopics()
>
> With these settings, by default users will get 3 topics, independently of the number of attempts or delay value:
myTopic, myTopic-retry and myTopic-dlt
>
> If a multiplier or maxDelay is set, yelding multiple delay values, users will get one topic per attempt plus the dlt:
> myTopic, myTopic-retry-0, myTopic-retry-1, myTopic-retry-2 ... myTopic-dlt

The other defaults are the same as for the rest of the feature, `maxAttempts = 3` and `FixedBackOff` with `delay=1s`.

Thanks.

tomazfernandes:
Hi @snicoll and @wilkinsona, looks like from the Spring Kafka team's perspective there are no further concerns. Is there anything you'd like to be changed implementation-wise? We can revise any part of this based on user feedback for 2.7.

Thanks

snicoll:
Thanks for the follow-up and for working with the team.

Time is running out with RC1 happening tomorrow. I will do my best to review it before that happens if possible.

tomazfernandes:
Sure, no worries. I should be available if there's anything to be changed prior to the release. Thanks.

snicoll:
@tomazfernandes thank you for making your first contribution to Spring Boot. If you're interested, I've polished your contribution in [b3e3581](https://github.com/spring-projects/spring-boot/commit/b3e3581271696682608e98255961610220e26712), mostly about using primitive types if possible. I also decided to remove the validation business as we don't really do such a thing in configuration properties usually. We'd rather let the underlying component we configure throw an exception if necessary.

tomazfernandes:
Thanks a lot @snicoll for finding the time to look into this in such short notice! The retryable topics feature has been gaining a lot of traction with the community and with auto configuration I'm sure more users will be able to benefit from it and in a simpler way.

I had missed that if users had more than one `KafkaTemplate` it'd throw an exception. If I understand correctly, now with `@ConditionalOnSingleCandidate(KafkaTemplate.class)` if users have more than one `KafkaTemplate` bean defined the retry topic configuration will back off. Is that correct?

I wonder if it would make sense to add a `kafkaTemplate` configuration property that defaults to "kafkaTemplate", and then fetch the template bean by name, so that users can still benefit of the feature by leveraging auto configuration's `KafkaTemplate` if that's the case, or set a different one to be used in case they have more than one defined. Is this something you usually do in auto configuration?

Or maybe that's expected behavior - if users choose to define multiple `KafkaTemplate` beans, they might just as well define `RetryTopicConfiguration` beans. We'd just need to properly document that.

What do you think?

Thanks again!

snicoll:
That's not what `@ConditionalOnSingleCandidate` does. A single `KafkaTemplate` should be defined (so either one in the context, or one flagged with `@Primary` in case of multiple instnaces). This is is consistent with what we do elsewhere and the recommendation is indeed to configure the feature if a "main"  `KafkaTemplate` cannot be determined. The auto-configuration report will mention why the feature wasn't configured (again, something users rely on elsewhere) so I don't think it needs an explicit mention in the doc.

tomazfernandes:
I see, makes sense. I've learned a lot about autoconfiguration and what it is and isn't supposed to do in this issue, so thanks a lot everyone involved.

Thanks again @snicoll for looking into this in time for the `RC` release!

PR #29884, opened 2022-02-16
Add Kafka immediateStop Property

garyrussell:


snicoll:
We should switch to the snapshot on `main` first so that we can merge this forward.

PR #8636, opened 2017-03-15
Add @DataElasticSearchTest support

mdeinum:
This commit introduces the @DataElasticSearchTest annotation.

See: gh-8544

PR #29911, opened 2022-02-20
Add DataElasticsearchTest annotation

eddumelendez:
Add parity to other test annotations for spring data projects.


wilkinsona:
Thanks, @eddumelendez. It looks like Testcontainers is required as a means of bootstrapping an ES node. Is that correct? It feels more viable than https://github.com/spring-projects/spring-boot/pull/8636 but it would be the only `@…DataTest` annotation where Testcontainers is required.

eddumelendez:
>  It looks like Testcontainers is required as a means of bootstrapping an ES node.

Yes, tescontainers is setting a single node

wilkinsona:
> it would be the only `@Data…Test` annotation where Testcontainers is required

I was wrong here. We already have `@DataRedisTest`, `@DataCassandraTest`, and `@DataNeo4jTest` where Testcontainers  will typically be used to manage the data store.

wilkinsona:

PR #29929, opened 2022-02-21
GraphQlService may not used contributed BatchLoaderRegistry

dugenkui03:


wilkinsona:
We should add a test to `GraphQlAutoConfigurationTests` that verifies that a user-provided `BatchLoaderRegistry` bean is used as expected.

snicoll:
@dugenkui03 thank you for making your first contribution to Spring Boot.

PR #29932, opened 2022-02-22
Document `WebSecurityCustomizer` for H2 Console

### LINKED ISSUE #28268: Document PathRequest.toH2Console and using the console with Spring Security
See https://github.com/spring-projects/spring-boot/issues/3734 for background. We added `PathRequest.toH2Console()` in 2.0 but it doesn't appear to be mentioned in the reference documentation. I think it'd be good to mention it in https://docs.spring.io/spring-boot/docs/2.5.x/reference/htmlsingle/#features.sql.h2-web-console.

hpoettker:
Resolves #28268.

The `WebSecurityCustomizer` is a bit blunt. But for a development setup, a proper `SecurityFilterChain` seems overly ceremonial, in my opinion. 😄

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/sql.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/sql.adoc
@@ -322,6 +322,21 @@ You can customize the console's path by using the configprop:spring.h2.console.p
 
 
 
+[[features.sql.h2-web-console.spring-security]]
+==== Configuring Spring Security for H2 Console
+H2 Console uses frames and, as it's intended for development only, does not implement CSRF protection measures. If your application uses Spring Security, you need to configure it accordingly.
+
+For example, Spring Security will ignore the console if the following `WebSecurityCustomizer` is exposed:
+
+[source,java,indent=0,subs="verbatim"]
+----
+include::{docs-java}/features/sql/h2webconsole/springsecurity/MySecurityConfiguration.java[]
+----
+
+TIP: `PathRequest.toH2Console()` returns the correct request matcher also when the console's path has been customized.
+
+
+
 [[features.sql.jooq]]
 === Using jOOQ
 jOOQ Object Oriented Querying (https://www.jooq.org/[jOOQ]) is a popular product from https://www.datageekery.com/[Data Geekery] which generates Java code from your database and lets you build type-safe SQL queries through its fluent API.
--- a/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/features/sql/h2webconsole/springsecurity/MySecurityConfiguration.java
+++ b/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/features/sql/h2webconsole/springsecurity/MySecurityConfiguration.java
@@ -1,0 +1,32 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.docs.features.sql.h2webconsole.springsecurity;
+
+import org.springframework.boot.autoconfigure.security.servlet.PathRequest;
+import org.springframework.context.annotation.Bean;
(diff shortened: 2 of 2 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/sql.adoc
```diff
@@ -322,6 +322,21 @@ You can customize the console's path by using the configprop:spring.h2.console.p
 
 
 
+[[features.sql.h2-web-console.spring-security]]
+==== Configuring Spring Security for H2 Console
```
snicoll, line 326: I am slightly worried that we had something in the reference guide that essentially gives access to everyone to a resource, especially with the current naming.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/sql.adoc
```diff
@@ -322,6 +322,21 @@ You can customize the console's path by using the configprop:spring.h2.console.p
...
+H2 Console uses frames and, as it's intended for development only, does not implement CSRF protection measures. If your application uses Spring Security, you need to configure it accordingly.
+
+For example, Spring Security will ignore the console if the following `WebSecurityCustomizer` is exposed:
```
snicoll, line 329: I think we need to be more explicit than "ignoring" and mention anybody could access it.

snicoll (commented):
Thanks for the PR. I've added a couple comments that are more directed to the team as I'd like to get more feedback on this. I've flagged the issue accordingly.

hpoettker:
@snicoll @wilkinsona Thanks for the feedback!

I'll rewrite the PR to show how a dedicated `SecurityFilterChain` can be exposed for the console.

wilkinsona:
I'm not sure that we should go that far. The security for the console may need to be quite complex, and that'll depend on the user's security configuration which is an unknown to us. I think we should avoid prescribing a particular approach and just point out any common pitfalls such as the situation with CSRF protection.

hpoettker:
I removed the code snippet and added some explanations on what needs to be done together with a warning not to do it in production.

wilkinsona:
@hpoettker Your code snippet in https://github.com/spring-projects/spring-boot/pull/29932/commits/1573e32e0011dcf8dae0e5076a0a1e4029985c23 has changed my mind. I was concerned about providing something that users would copy and paste and try to use as-is. I think your snippet is informative without making that too much of a risk. Sorry to be a pain, but would you mind reinstating it?

hpoettker:
@wilkinsona I reverted the last commit. Quick think and no pain at all. 😄 

I fully agree that one needs to be careful with publishing code snippets in the reference documentation that have the potential to create security risks if used lightheartedly. The code snippet makes it pretty clear that it is not supposed to be used in production code. However, if just the bean method is copied verbatim and the comment asking for authorization to be configured is ignored, then the H2 console is again unprotected as in the first draft with the `WebSecurityCustomizer`. I realized that only after pushing, and then wrote the version without any code snippet.

hpoettker:
The code snippet that I came up with that is the most safe to be copied and pasted is
```
@Component
@Order(Ordered.HIGHEST_PRECEDENCE)
public class H2ConsoleFilterBean extends HeaderWriterFilter {

  private final RequestMatcher h2ConsoleRequestMatcher = PathRequest.toH2Console();

  public H2ConsoleFilterBean() {
    super(Arrays.asList(new XFrameOptionsHeaderWriter(XFrameOptionsMode.SAMEORIGIN)));
  }

  @Override
  protected void doFilterInternal(HttpServletRequest request, HttpServletResponse response, FilterChain filterChain)
      throws ServletException, IOException {
    CsrfFilter.skipRequest(request);
    super.doFilterInternal(request, response, filterChain);
  }

  @Override
  protected boolean shouldNotFilter(HttpServletRequest request) {
    return !h2ConsoleRequestMatcher.matches(request);
  }

}
```
But this code snippet is quite dense and may confuse more people than it helps.

linghengqian:
@hpoettker 
- It seems that Spring Security encourages the adoption of the `Lambda DSL` instead of `and()`, in your opinion, would the following be better?
```java
@Bean
@Order(Ordered.HIGHEST_PRECEDENCE)
SecurityFilterChain h2ConsoleSecurityFilterChain(HttpSecurity http) throws Exception {
    return http.requestMatcher(PathRequest.toH2Console())
            .csrf(AbstractHttpConfigurer::disable)
            .headers(headers -> headers.frameOptions(HeadersConfigurer.FrameOptionsConfig::sameOrigin))
            .build();
}
```

hpoettker:
Hi @linghengqian, thanks for joining the discussion!
Please note that your question could be perceived as off-topic here and that I won't answer it. As an unaffiliated contributor, I really have no authority to do so.

wilkinsona:
The Spring Security docs seem to use the lambda-style API, such as in [their example for disabling CSRF](https://docs.spring.io/spring-security/site/docs/5.5.0/reference/html5/#servlet-csrf-configure). I think it makes sense to align with that. I've opened https://github.com/spring-projects/spring-boot/issues/30432.

PR #29976, opened 2022-02-23
Remove redundant `ResourceLoader`s from graphql configuration

dugenkui03:
Remove redundant `ResourceLoader`s from graphql configuration.


wilkinsona:
Thanks very much, @dugenkui03.

Issue #6177, opened 2016-06-16
Startup time 2.5 times slower with executable WAR than in Eclipse


stefanocke:
Spring Boot 1.3.5.Release
OS: Windows 7 (but no different results with Linux)

I have a Spring Boot Application with some REST Endpoints and Spring Data JPA for Persistence.
When running in Spring Tool Suite, it starts within about **11 seconds**. When running as executable war with "java -jar", it needs **29 seconds** to start.  
- The application is packaged as war. 
- Spring Boot plugin is called with goal "repackage". 
- Spring-Boot-starter-tomcat is included as provided dependency.

I did some sampling with JVisualVM for both cases. In the following, the diff of the hotspot mehtods is shown. It seems that accessing the WAR/JAR entries seems to be quite slow, but I am not sure.
Please ignore the difference for Thread.sleep(). (I have added this)

![image](https://cloud.githubusercontent.com/assets/7280950/16119175/a2474738-33da-11e6-94a8-4026791987b1.png)

Is there anything that can be done to improve war startup time? (Other packaging options etc)

I have the same experience with other spring boot apps packaged the same way. So I guess it does not depend too much on the details of the application...

wilkinsona:
An executable jar will always be a bit slower than running directly in your IDE, but I wouldn't expect the difference to be so large. For example, our Freemarker sample starts in 2.603 seconds in Eclipse and 2.818 seconds when using `java -jar`. The latter is 108.3% of the former compared with 263% in your case.

Can you please share your application or a representative sample? I'm interested to see exactly what it's doing and why there's such a dramatic difference.

stefanocke:
@snicoll : Thanks for correct title. I just tried to change from WAR to JAR and now its 18 seconds. This is better but still about 160% compared to Eclipse.

@wilkinsona : Thanks for feedback. Currently I cannot share, but maybe I can prepare something in the next days. Perhaps it halps to know that the JAR / WAR is quite big (50 MB) since Spring Cloud is used also  (Eureka Client and Config Client).

snicoll:
> When running in Spring Tool Suite,

@stefanocke what are you doing in STS? Are you running the main method or something else? 

11s to 18s is not a ~10% increase so something extra is going on. But the key thing here is that running as an executable war makes things way worse it seems.

stefanocke:
@snicoll In STS, I use the "Boot Dashboard" and there "(Re)start".  I guess that is quite the same as running the main method.

[...]

Issue #6216, opened 2016-06-24
Locations are scanned twice for TLDs when starting an executable war


wilkinsona:
Jasper scans both `WEB-INF/lib` and the class loader hierarchy for TLDs. `StandardJarScanner` does contains some logic that should avoid duplicate scanning:

```
if (processedURLs.contains(url)) {
    // Skip this URL it has already been processed
    continue;
}
```

However, the URLs from `WEB-INF/lib` and the classpath do not matching up. For example:

```
jar:file:/Users/awilkinson/dev/workspaces/spring/spring-boot/master/gh-6177/target/gh-6177-0.0.1-SNAPSHOT.war!/WEB-INF/lib/archaius-core-0.7.4.jar
jar:file:/Users/awilkinson/dev/workspaces/spring/spring-boot/master/gh-6177/target/gh-6177-0.0.1-SNAPSHOT.war!/WEB-INF/lib/archaius-core-0.7.4.jar!/
```

The first is returned from the `ServletContext`, the second comes from the class loader and is created by our `JarFile` implementation.

A couple of possibilities:

- Remove the `!/` suffix so that the URLs match
 - Disable scanning of the classpath when launching an executable war. I think this will work as the war is intended to be self-contained, i.e. all of the TLDs should be packaged in the jar in `WEB-INF/classes` or in a jar in `WEB-INF/lib`. Jasper will find them in either of these locations with also looking at the class path.

wilkinsona:
This is still a problem, although the duplicate URLs now have a slightly different form. For example:

```
war:file:/Users/awilkinson/dev/spring/spring-boot/master/spring-boot-samples/spring-boot-sample-web-jsp/target/spring-boot-sample-web-jsp-2.0.0.BUILD-SNAPSHOT.war*/WEB-INF/lib/spring-core-5.0.0.BUILD-SNAPSHOT.jar
jar:file:/Users/awilkinson/dev/spring/spring-boot/master/spring-boot-samples/spring-boot-sample-web-jsp/target/spring-boot-sample-web-jsp-2.0.0.BUILD-SNAPSHOT.war!/WEB-INF/lib/spring-core-5.0.0.BUILD-SNAPSHOT.jar!/
```

PR #29998, opened 2022-02-25
Set ExtractingRoot as context ressources to avoid first load slow

dlemaignent:
At fisrt load into browser response is very slow. Set ExtractingRoot as context ressources fix this issue by pre-extracting the resources.
Maybe this commit can be linked this issue #6177

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/TomcatServletWebServerFactoryCustomizer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/TomcatServletWebServerFactoryCustomizer.java
@@ -22,6 +22,7 @@ import org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactor
 import org.springframework.boot.web.server.WebServerFactoryCustomizer;
 import org.springframework.core.Ordered;
 import org.springframework.util.ObjectUtils;
+import org.apache.catalina.webresources.ExtractingRoot;
 
 /**
  * {@link WebServerFactoryCustomizer} to apply {@link ServerProperties} to Tomcat web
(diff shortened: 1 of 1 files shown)


dlemaignent:
Note : to reproduce issue you need to launch spring boot web application with a war package set as executable (./my_app.war)

wilkinsona:
Thanks for the suggestion.

> At fisrt load into browser response is very slow. Set ExtractingRoot as context ressources fix this issue

Can you please provide a minimal sample application that illustrates the difference?

> Maybe this commit can be linked this issue https://github.com/spring-projects/spring-boot/issues/6177

Our belief is that https://github.com/spring-projects/spring-boot/issues/6216 accounts for much of the difference when running a war. A sample application that illustrates the problem you're facing will help us to see if that's not the case.

> Note : to reproduce issue you need to launch spring boot web application with a war package set as executable (./my_app.war)

Looking at Tomcat's code, I cannot see why it would not also reproduce when running with `java -jar my_app.war`. Again, a sample application will help us to understand exactly what's happening.

dlemaignent:
Thanks for your quick response.

I've created this repo https://github.com/dlemaignent/first-war-start-slow to reproduce the issue.

I confirm that the application can be launched with java -jar my_app.war , the issue is the same.

wilkinsona:
Thanks, @dlemaignent. I'm seeing a significant difference in startup time with `./mvnw spring-boot:run` (3.078s) vs `java -jar` (8.561s):

```
2022-03-10 15:14:02.082  INFO 64384 --- [  restartedMain] c.e.f.FirstWarStartSlowApplication       : Started FirstWarStartSlowApplication in 3.078 seconds (JVM running for 3.544)
```

```
2022-03-10 15:15:24.511  INFO 64501 --- [           main] c.e.f.FirstWarStartSlowApplication       : Started FirstWarStartSlowApplication in 8.561 seconds (JVM running for 9.487)
```

Switching to Undertow, the difference is much less pronounced:

```
2022-03-10 15:25:40.502  INFO 68556 --- [  restartedMain] c.e.f.FirstWarStartSlowApplication       : Started FirstWarStartSlowApplication in 2.893 seconds (JVM running for 3.343)
```

```
2022-03-10 15:19:44.830  INFO 68316 --- [           main] c.e.f.FirstWarStartSlowApplication       : Started FirstWarStartSlowApplication in 4.61 seconds (JVM running for 5.209)
```

Switching back to Tomcat but with jar packaging, it is significantly quicker with performance that is similar to Undertow's:

```
2022-03-10 15:27:43.556  INFO 69563 --- [           main] c.e.f.FirstWarStartSlowApplication       : Started FirstWarStartSlowApplication in 4.739 seconds (JVM running for 5.265)
```

There's obviously work to do here so that Tomcat starts quickly when packaged as a war. However, I don't think that extracting by default should be the answer. It would mean that changing packaging alters the runtime behaviour in terms of how the application writes to the filesystem and I think that might catch a lot of people out. I'd prefer that we find a way to close the performance gap without making a functional change. In the meantime, you can opt in as you have done using a customizer. Thanks anyway for the suggestion.

PR #30000, opened 2022-02-26
Use Gradle's task configuration avoidance APIs in the main reference docs

larsgrefer:
see https://github.com/spring-projects/spring-boot/pull/29762
see https://docs.gradle.org/7.0/userguide/task_configuration_avoidance.html

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
@@ -34,7 +34,7 @@ If you are using an `additional-spring-configuration-metadata.json` file, the `c
 [source,gradle,indent=0,subs="verbatim"]
 ----
 	tasks.named('compileJava') {
-	    inputs.files(processResources)
+	    inputs.files(tasks.named('processResources'))
 	}
 ----
 
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
@@ -33,7 +33,9 @@ If you are using an `additional-spring-configuration-metadata.json` file, the `c
 
 [source,gradle,indent=0,subs="verbatim"]
 ----
-	compileJava.inputs.files(processResources)
+	tasks.named('compileJava') {
+	    inputs.files(processResources)
+	}
 ----
 
 This dependency ensures that the additional metadata is available when the annotation processor runs during compilation.
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/properties-and-configuration.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/properties-and-configuration.adoc
@@ -71,7 +71,7 @@ You can automatically expand properties from the Gradle project by configuring t
 
 [source,gradle,indent=0,subs="verbatim"]
 ----
-	processResources {
+	tasks.named('processResources', Copy) {
 		expand(project.properties)
 	}
 ----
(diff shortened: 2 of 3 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
```diff
@@ -33,7 +33,9 @@ If you are using an `additional-spring-configuration-metadata.json` file, the `c
 
 [source,gradle,indent=0,subs="verbatim"]
 ----
-	compileJava.inputs.files(processResources)
+	tasks.named('compileJava') {
+	    inputs.files(processResources)
```
wilkinsona, line 37: I think will eagerly create the `processResources` task.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/properties-and-configuration.adoc
```diff
@@ -71,7 +71,7 @@ You can automatically expand properties from the Gradle project by configuring t
 
 [source,gradle,indent=0,subs="verbatim"]
 ----
-	processResources {
+	tasks.named('processResources', Copy) {
```
wilkinsona, line 74: I don't think the task's type is needed here.

wilkinsona (commented):
Thanks, @larsgrefer. I've left a couple of comments for your consideration when you have a minute.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
larsgrefer: This will only create the `processResources` task when the closure is executed, which in turn only happens when `compileJava` is being created/configured. In this case, `processResources` would have to be created anyway.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/properties-and-configuration.adoc
larsgrefer: Yes, it would work without the type.

I've added the explicit type declaration to give the IDE a chance to resolve what `expand` refers to, but at least for IntelliJ IDEA this does not seem to be sufficient.

For proper auto-completion in IntelliJ IDEA we'd need this:
```gradle
tasks.named('processResources', Copy) {
    it.expand(project.properties)
}
```

If we don't care about IDE support, this would be enough:
```gradle
tasks.named('processResources') {
    expand(project.properties)
}
```

wilkinsona: Thanks. Given that it’s unrelated to configuration avoidance, please revert this part of the changes.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
wilkinsona: I wonder if that is that the case 100% of the time.

Regardless, this is an example of the increased complexity of task configuration avoidance — you need to be more aware of the relationships between tasks. According to Gradle’s own docs, you also need to [consider where the task was created](https://docs.gradle.org/current/userguide/task_configuration_avoidance.html#sec:task_configuration_avoidance_pitfalls).

In this case `processResources` is defined by another plugin, but so is `compileJava`. That makes Gradle’s guidance a little unclear to me.

My feeling is that consistently using tasks.named in a situation such as this reduces cognitive load. You don’t need to think about where a task has come from or their existing relationships and things will just work (I think).

larsgrefer: In my tests it did not make a difference if `processResources` is used lazily or directly (inside the lazily evaluated `compileJava` closure). The nearest match in the Gradle docs I could find is this:

```gradle
tasks.register("someTask")

tasks.register("anEagerTask") {
    dependsOn someTask
}
```
Here `someTask` is referenced directly inside the lazy configuration closure of `anEagerTask`.

That being said, I agree that just always using `tasks.named` makes things more consistent.

larsgrefer:
I've updated the PR based on your comments.

wilkinsona:
Thanks very much, @larsgrefer.

Issue #29298, opened 2022-01-09
Add support for declaring a classifier when configuring dependency management


wilkinsona:
We need this for https://github.com/spring-projects/spring-boot/issues/29023. It would also be useful in 3.0 where we may want to reintroduce dependency management for Apache Johnzon and its `-jakarta`-classified artifacts.

PR #30002, opened 2022-02-26
Restore Ehcache 3 Support 

chrisdennis:
This is a prototype of the restoration of Ehcache 3 support to Spring Boot. Ehcache uses Gradle feature variants (with appropriate capabilities) to model the split "Jakarta/Java EE" landscape. These are exposed in Maven as artifacts with 'jakarta' classifiers with appropriate optional dependencies. I've therefore implemented a solution to #29298 as a prerequisite here.

Currently this is depending on an alpha release of Ehcache 3.10, hence it's only a draft PR, but I wanted to get this out there for two reasons:
1. To confirm the acceptability of the classifier/feature variant approach for Jakarta support.
2. To understand the need for the `ehcache-transactions` module. The transactions module allows Ehcache instances to be exposed as XA resources and participate in 2PC managed by a JTA transaction manager. It's not clear to me how this relates to Springs caching support - I'm not sure how or why a user would need this dependency in a Spring (Boot) context.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -191,6 +191,19 @@ bom {
 			]
 		}
 	}
+	library("Ehcache3", "3.10.0") {
+		group("org.ehcache") {
+			modules = [
+					"ehcache" {
+						classifier = 'jakarta'
+					},
+					"ehcache-clustered",
+					"ehcache-transactions" {
+						classifier = 'jakarta'
+					}
+			]
+		}
+	}
 	library("Elasticsearch", "7.17.0") {
 		group("org.elasticsearch") {
 			modules = [
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/documentation/io.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/documentation/io.adoc
@@ -2,7 +2,7 @@
 == IO
 If your application needs IO capabilities, see one or more of the following sections:
 
-* *Caching:* <<io#io.caching, Caching support with Hazelcast and more>>
+* *Caching:* <<io#io.caching, Caching support with EhCache, Hazelcast and more>>
 * *Quartz:* <<io#io.quartz, Quartz Scheduling>>
 * *Mail:* <<io#io.email, Sending Email>>
 * *Validation:* <<io#io.validation, JSR-303 Validation>>
--- a/spring-boot-project/spring-boot-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-autoconfigure/build.gradle
@@ -88,6 +88,11 @@ dependencies {
 	optional("org.eclipse.jetty.websocket:websocket-jetty-server") {
 		exclude(group: "org.eclipse.jetty", module: "jetty-jndi")
 	}
+	optional("org.ehcache:ehcache") {
+		artifact {
+			classifier = 'jakarta'
+		}
+	}
 	optional("org.elasticsearch.client:elasticsearch-rest-client") {
 		exclude group: "commons-logging", module: "commons-logging"
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cache/EhCache3CacheAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cache/EhCache3CacheAutoConfigurationTests.java
@@ -1,0 +1,67 @@
+/*
+ * Copyright 2012-2019 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.cache;
+
+import org.ehcache.jsr107.EhcacheCachingProvider;
+import org.junit.jupiter.api.Test;
+
+import org.springframework.boot.autoconfigure.cache.CacheAutoConfigurationTests.DefaultCacheConfiguration;
+import org.springframework.boot.testsupport.classpath.ClassPathExclusions;
+import org.springframework.cache.jcache.JCacheCacheManager;
+import org.springframework.core.io.ClassPathResource;
+import org.springframework.core.io.Resource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+/**
+ * Tests for {@link CacheAutoConfiguration} with EhCache 3.
+ *
+ * @author Stephane Nicoll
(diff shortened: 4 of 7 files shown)


wilkinsona:
Thanks, @chrisdennis.

> I've therefore implemented a solution to https://github.com/spring-projects/spring-boot/issues/29298 as a prerequisite here.

Thanks for this. Unfortunately, it doesn't quite meet the requirements that we have for some of Kafka's modules. I [implemented something](https://github.com/wilkinsona/spring-boot/tree/gh-29298) earlier this year which should work for both cases, I just haven't had a chance to merge it yet.

> To confirm the acceptability of the classifier/feature variant approach for Jakarta support

I don't think we have any dependencies that have taken this approach just yet, but I don't think it'll be a problem.

> To understand the need for the ehcache-transactions module

Spring Boot supports JTA-based transaction coordination either by retrieving the `UserTransaction` and `TransactionManager` from JNDI when deployed as a war file in an application server, or via integration of a third-party transaction manager such as [Narayana](https://github.com/snowdrop/narayana-spring-boot). We need to position Spring Boot 3.0 such that this will work in both cases with Jakarta EE 9. That should include cases where a user wants to enlist their cache as a resource.

chrisdennis:

This reinstates previously existing code and introduce no new IP.

wilkinsona:
@chrisdennis You’ll need to sign the CLA, I’m afraid. The changes proposed here [are not considered to be obvious](https://cla.pivotal.io/about#obvious-fixes) as they will result in a change in functionality.

chrisdennis:
@wilkinsona no problem... I figured I would see what you guys thought before I went ahead and engaged with the bureaucracy on this end.

FYI: the alpha release referenced here is likely very close to what will become the 3.10.0 GA release. That should happen next week. Once I have that out I'll update the deps here and remove the draft status from this PR.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfigurationTests.java
```diff
@@ -81,7 +86,10 @@
  * @author Ryon Day
  */
 @ClassPathExclusions("hazelcast-client-*.jar")
-class CacheAutoConfigurationTests extends AbstractCacheAutoConfigurationTests {
```
wilkinsona, line 84: I can't see the connection between this change (and deleting AbstractCacheAutoConfigurationTests) and reinstating support for EhCache 3. I'd prefer to make the changes as minimal as possible and leave this as-is if we can.

spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -191,6 +191,19 @@ bom {
...
+	library("Ehcache3", "3.10.0-alpha1") {
+		group("org.ehcache") {
+			modules = [
+					"ehcache" {
+						classifier = 'jakarta'
+					},
+					"ehcache-transactions" {
+						classifier = 'jakarta'
+					},
+					"ehcache-clustered"
```
wilkinsona, line 203: We try to list modules in alphabetical order.

wilkinsona (changes requested):
Thanks for the updates, @chrisdennis. I've left a few comments for your consideration when you have a minute.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfigurationTests.java
chrisdennis: My understanding is that AbstractCacheAutoConfigurationTests was only introduced to allow the Ehcache tests to have a @ClassPathExclusions annotation on the Ehcache 2 JARs. The bug that caused that to be necessary was fixed in Ehcache 2 a while back... and regardless Ehcache 2 I assume isn't part of Spring Boot 3.x?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/orm/jpa/Hibernate2ndLevelCacheIntegrationTests.java
```diff
@@ -41,9 +41,9 @@
 			.withUserConfiguration(TestConfiguration.class);
 
 	@Test
-	void hibernate2ndLevelCacheWithJCacheAndHazelcast() {
-		String cachingProviderFqn = HazelcastServerCachingProvider.class.getName();
-		String configLocation = "classpath:hazelcast.xml";
+	void hibernate2ndLevelCacheWithJCacheAndEhCache3() {
+		String cachingProviderFqn = EhcacheCachingProvider.class.getName();
+		String configLocation = "ehcache3.xml";
```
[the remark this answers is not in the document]
chrisdennis: I started out here with a simple revert of the commit that removed Ehcache 3 support. That's why this came in. I can remove as you wish.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/documentation/io.adoc
```diff
@@ -2,7 +2,7 @@
 == IO
 If your application needs IO capabilities, see one or more of the following sections:
 
-* *Caching:* <<io#io.caching, Caching support with Hazelcast and more>>
+* *Caching:* <<io#io.caching, Caching support EhCache, Hazelcast and more>>
```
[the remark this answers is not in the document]
chrisdennis: Artifact of the initial revert.

- [ ] restore correct grammar in `io.adoc`

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfigurationTests.java
wilkinsona: Unfortunately, I can't recall the exact reason. Folding everything back into a single class is a good thing to do if we can, but it's the sort of thing that we prefer to do separately. If, as you suspect, a bug's been fixed in EhCache 2 such that the separate classes are no longer necessary we can make that change in all our currently maintained branches and not just in main.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/orm/jpa/Hibernate2ndLevelCacheIntegrationTests.java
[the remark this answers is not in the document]
wilkinsona: Yes, please.

wilkinsona:
Either's fine. Thank you. We'll multiple commits into one as part of merging so a force-push is probably slightly easier to process.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/cache/CacheAutoConfigurationTests.java
chrisdennis: If I recall correctly without digging through the issue-tracker we fixed this in Ehcache 3 by shading and relocating some library version clashes that meant the two couldn't co-exist. Regardless, I'll back these changes out of this PR.

wilkinsona (approved):
This looks great now, @chrisdennis. Thank you.

snicoll:
@chrisdennis thank you for the follow-up and for making your first contribution to Spring Boot.

Issue #8933, opened 2017-04-18
Slf4JLoggingSystem removes JUL root handlers via SLF4JBridgeHandler


ifigotin:
I'm part of Google Cloud engineering team, and I'm looking into making sure Spring-Boot apps could run seamlessly on Google App Engine Java 8 runtime.

Version:  1.5.2-RELEASE

Issue Description:  Currently, Spring-Boot logging does not work out of the box for a simple (spring-boot-starter-web) hello world app running on App Engine Java 8.

My understanding is that Spring-boot starter relies on Logback. See [LogbackLoggingSystem](
https://github.com/spring-projects/spring-boot/blob/1.5.x/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java).

Logback in turn relies on Slf4J, which during the initialization removes all previously configured handlers for Java Util's root logger. See [Slf4JLoggingSystem](
https://github.com/spring-projects/spring-boot/blob/1.5.x/spring-boot/src/main/java/org/springframework/boot/logging/Slf4JLoggingSystem.java#L78).

This is problematic for App Engine Standard Java runtime, which relies on root logger handler itself to properly route log messages.

`SLF4JBridgeHandler` javadoc states that the removal of root handlers is optional. See [the documentation](https://www.slf4j.org/api/org/slf4j/bridge/SLF4JBridgeHandler.html). 

Would it be possible not to call `SLF4JBridgeHandler.removeHandlersForRootLogger()` from `Slf4JLoggingSystem`?

To reproduce the issue, one needs an App Engine application that is white-listed for Java 8 runtime, and I'd be happy to help with this.

snicoll:
To get you going with `1.5.2.RELEASE` in the meantime there are a couple of options.

First you can remove logback and slf4j altogether if you add the following to your build

```xml
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter</artifactId>
	<exclusions>
		<exclusion>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-logging</artifactId>
		</exclusion>
	</exclusions>
</dependency>
<dependency>
	<groupId>commons-logging</groupId>
	<artifactId>commons-logging</artifactId>
	<version>1.2</version>
</dependency>
```

[...]

PR #30003, opened 2022-02-26
Use Log4jBridgeHandler to route JUL-based logging into Log4j 2

ppkarwasz:
Since version 2.15.0 `log4j-jul` contains a `Log4jBridgeHandler`, that allows to forward JUL to Log4j 2.x and synchronizes the logger levels of the two frameworks.

This PR adds support for the `Log4jBridgeHandler` and sets it as default bridge handler for the Log4j 2.x stack and fallback handler for Logback.

It also removes `jul-to-slf4j` from the `spring-boot-starter-log4j2`, hence leaving a pure Log4j 2.x stack.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/Slf4JLoggingSystem.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/Slf4JLoggingSystem.java
@@ -16,14 +16,11 @@
 
 package org.springframework.boot.logging;
 
-import java.util.Collections;
-import java.util.List;
 import java.util.logging.ConsoleHandler;
 import java.util.logging.Handler;
 import java.util.logging.LogManager;
 import java.util.logging.Logger;
 
-import org.apache.logging.log4j.jul.Log4jBridgeHandler;
 import org.slf4j.bridge.SLF4JBridgeHandler;
 
 import org.springframework.util.Assert;
@@ -37,21 +34,10 @@ import org.springframework.util.ClassUtils;
  */
 public abstract class Slf4JLoggingSystem extends AbstractLoggingSystem {
 
-	private static final String SLF4J_BRIDGE_HANDLER = "org.slf4j.bridge.SLF4JBridgeHandler";
-
-	private static final String LOG4J_BRIDGE_HANDLER = "org.apache.logging.log4j.jul.Log4jBridgeHandler";
-
-	private static final String LOG4J_LOG_MANAGER = "org.apache.logging.log4j.jul.LogManager";
-
-	private final List<BridgeHandler> handlers;
+	private static final String BRIDGE_HANDLER = "org.slf4j.bridge.SLF4JBridgeHandler";
 
 	public Slf4JLoggingSystem(ClassLoader classLoader) {
-		this(classLoader, Collections.singletonList(new Slf4jHandler(classLoader)));
-	}
-
-	protected Slf4JLoggingSystem(ClassLoader classLoader, List<BridgeHandler> handlers) {
 		super(classLoader);
-		this.handlers = handlers;
 	}
 
 	@Override
@@ -62,7 +48,9 @@ public abstract class Slf4JLoggingSystem extends AbstractLoggingSystem {
 
 	@Override
 	public void cleanUp() {
-		this.handlers.stream().filter(BridgeHandler::isAvailable).forEach(BridgeHandler::remove);
+		if (isBridgeHandlerAvailable()) {
+			removeJdkLoggingBridgeHandler();
+		}
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/Slf4JLoggingSystem.java
```diff
@@ -16,11 +16,14 @@
...
 package org.springframework.boot.logging;
 
+import java.util.Collections;
+import java.util.List;
 import java.util.logging.ConsoleHandler;
 import java.util.logging.Handler;
 import java.util.logging.LogManager;
 import java.util.logging.Logger;
 
+import org.apache.logging.log4j.jul.Log4jBridgeHandler;
```
wilkinsona, line 26: The SLF4J logging system should know anything that’s Log4j2- or Logback-specific. If there’s a need for the two logging system to use different bridge handlers then this class should be reworked and the handler logic pushed down into the two system-specific sub-classes.

wilkinsona (changes requested):
Thanks for the proposal. Please see my comments for your consideration when you have a minute. I’d also like to learn more about the motivation for this change.

Is it purely to simplify the routing from JUL into Log4j2 so that it no longer needs to go via SLF4J, or is there something more to it that I have overlooked?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java
```diff
@@ -100,7 +101,7 @@ public FilterReply decide(Marker marker, ch.qos.logback.classic.Logger logger, L
 	};
 
 	public LogbackLoggingSystem(ClassLoader classLoader) {
-		super(classLoader);
+		super(classLoader, Arrays.asList(new Slf4jHandler(classLoader), new Log4jHandler(classLoader)));
```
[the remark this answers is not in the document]
ppkarwasz: Maybe not, but I provided this for duality.

Until now JUL logs were delivered to the Log4j 2.x API through SLF4J. It seemed natural to provide a JUL -> Log4j 2.x API -> SLF4J forwarding for Logback if `jul-to-slf4j` is not available, but `log4j-jul` is.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/Slf4JLoggingSystem.java
ppkarwasz: I hesitated on changing this class name to `JulBridgingLoggingSystem`, but that would break API compatibility.

I'll refactor the code and push the changes down to the `Log4j2LoggingSystem`.

ppkarwasz:
@wilkinsona,

While simplification is my main motivation for this change proposal (a Log4j stack requires one less dependency), it also comes with a bonus feature: the `Log4jBridgeHandler` synchronizes the logging levels between JUL and Log4j 2.x and therefore enhances performance.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
```diff
@@ -155,10 +161,50 @@ public void beforeInitialize() {
...
+		final String logManagerClassName = java.util.logging.LogManager.getLogManager().getClass().getName();
+		return LOG4J_LOG_MANAGER.equals(logManagerClassName);
+	}
+
+	private boolean isLog4jBridgeHandlerAvailable() {
+		return ClassUtils.isPresent(LOG4J_BRIDGE_HANDLER, getClassLoader());
+	}
+
+	private void installLog4jBridgeHandler() {
+		Log4jBridgeHandler.install(true, null, true);
```
wilkinsona, line 195: I think the first `true` here will cause a regression of the changes for https://github.com/spring-projects/spring-boot/issues/8933. Logic similar to `removeDefaultRootHandler` on `Slf4JLogingSystem ` needs to be used instead.

That makes me wonder if we need to take a slightly different approach overall, possibly moving the bridge handler-related logic out into a separate class that can then be used by the different logging systems. I'm not yet sure quite how that would look so I'll need to experiment a bit before I can make a more concrete suggestion.

ppkarwasz: I added the logic used by the `Slf4JLoggingSystem`, which should solve the immediate problem.

The bridge logic should probably be refactor to separate classes. I started with this approach in commit [39ed0a1](https://github.com/spring-projects/spring-boot/pull/30003/commits/39ed0a12512402cf54dadbf7169a54af8febba8f), but then it looked as an overkill.

I think that especially the logic to decide whether to modify JUL's configuration or not should be expanded (a switch case on the name of `java.util.logging.LogManager` implementation?): all three logging systems must decide whether they can modify the root handler and how.

ppkarwasz:
@wilkinsona, thank you. Is there any chance this change will be backported to 2.6.x? Of course in 2.6.x the `SLFJLoggingSystem` should probably stay for backward compatibility?

wilkinsona:
No, sorry. The change isn't fixing a bug so we won't make it in a maintenance release.

Issue #28909, opened 2021-12-06
Provide compatibility or replace the need for the Log4j-Spring-Boot Module


rgoers:
Spring Boot 2.6.0 added support for Log4j2 Composite Configurations.  Log4j's log4j-spring-boot module also provides support for this but in a different manner since prior to 2.6.0 Spring Boot only allowed a single property to identify the logging configuration files. But log4j-spring-boot also allows these files to be stored in a Spring Cloud Configuration server instance with access via HTTP(S) even if the access is password protected. Spring Boot does not so anyone looking for that support will have to forego Spring Boot's support and use log4j-spring-boot instead.  Log4j also allows SSL configuration to be provided although this is probably not very important in a Spring Boot environment where the SSL/TLS configuration for the JVM is probably OK.

log4j-spring-boot also provides

### Spring Lookup
The Spring Lookup allows configuration files to reference properties defined in Spring configuration files from a Log4j configuration file. For example:

`<property name="applicationName">${spring:spring.application.name}</property>`
would set the Log4j applicationName property to the value of spring.application.name set in the Spring configuration. Users can then do things like:

 ```
 <Appenders>
    <RollingFile name="log4j" fileName="${LOG_DIR}/${applicationName}.log" filePattern="${LOG_DIR}/archive/${applicationName}.log.%d{yyyyMMdd_HH}-%i">
      <PatternLayout pattern="${log4jPattern}"/>
      <Policies>
        <SizeBasedTriggeringPolicy size="30 MB"/>
      </Policies>
      <!-- A max of 20 will allow 20 files per second with the date pattern specified on the RollingFile declaration.
          Hopefully that is a ridiculous value -->
      <DefaultRolloverStrategy min="1" max="20">
        <Delete basePath="${LOG_DIR}/archive">
          <!-- Nested conditions: the inner condition is only evaluated on files for which the outer conditions are true. -->
          <IfFileName glob="${applicationName}.log.*">
            <!-- Only allow 500 MB of files to accumulate -->
            <IfAccumulatedFileSize exceeds="500 MB"/>
          </IfFileName>
        </Delete>
      </DefaultRolloverStrategy>
    </RollingFile>
  </Appenders>  
```

### Spring Property Source
[...]

PR #30004, opened 2022-02-27
Adds support for Log4j SpringLookup

ppkarwasz:
Since Log4j 2.14.0, configuration files can resolve properties from Spring's `Environment`.

While this is already possible with the custom `Log4j2CloudConfigLoggingSystem`, this PR enables this feature in the
standard `Log4j2LoggingSystem`.

### DIFF
--- a/spring-boot-project/spring-boot/build.gradle
+++ b/spring-boot-project/spring-boot/build.gradle
@@ -10,9 +10,11 @@ plugins {
 description = "Spring Boot"
 
 def tomcatConfigProperties = "$buildDir/tomcat-config-properties"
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the proposal. Unfortunately, we’re not keen on the existing circular dependencies between Spring Boot and Log4j 2 and this will strengthen that cycle. We feel that is the wrong direction to take.

We’d prefer to break the cycle as proposed in #28909 by providing functionality that is equivalent to SpringLookup. This would also remove one of the differences between Spring Boot’s Logback and Log4j 2 integration.

Thanks again for the proposal.

Issue #29620, opened 2022-02-02
Support overriding the default SanitizingFunction


robin-carry:
Based on the session given by Phil Webb [1], in the project with simple demo he show-cased that the `SanitizingFunction` can be included for hiding for example **Command Line Arguments** from the actuator endpoint. 

Based on the code given [here](https://github.com/philwebb/whats-new-in-spring-boot-2-6/blob/main/src/main/java/com/example/myproject/MyApplication.java#L15) if they are two command line arguments passed to this code `--test.secret=password` and `--test.key=important`, then

- `password` will actually be displayed as `****` 
- `important` will be displayed as `you're never gonna get it`

Even though both the parameters are passed as command line argument, for `password` the newly define `SantizingFunction` was not applied. 

Enhancement: Either support some sort of order or override the default behavior?

[1] https://github.com/philwebb/whats-new-in-spring-boot-2-6

philwebb:
A few options spring to mind for this issue.

* We could add `@Order` support to the `Sanitizer` default sanitizing function. This might be a bit tricky because the function is added by the `Sanitizer` so we'd only be able to order on class annotations (not on `@Bean` ones).
* We could make `Sanitizer` a bean and allow a completely different implementation
* We could add a property to disable the default ones
* We could add something to `SanitizableData` to indicate that a value has been sanitized already so later functions can skip it.

In the meantime, I think you can set `management.endpoint.env.keys-to-sanitize` (or equivalent other properties) to an empty list so that the default function becomes a no-op.

philwebb:
Closing in favor of PR #30006. Thanks @terminux!

PR #30006, opened 2022-02-28
Support overriding the default SanitizingFunction

terminux:
Hi, I submitted this PR to solve #29620 based on @philwebb ’s [comment](https://github.com/spring-projects/spring-boot/issues/29620#issuecomment-1027545966).

When a value has been sanitized already, later functions can skip it.

Closes gh-29620


philwebb:
When we look at merging this one we should check that `orderedStream()` is used to provide the `SanitzerFunction` beans.

terminux:
Thank you very much for your reply @philwebb  , nice suggestions. I can also implement this check if I have time.

PR #30012, opened 2022-02-28
Improve property placeholder documentation to mention environment variables and default values

62mkv:
I've figured that way too often I try to find an example in official reference documentation, and invariably end up on this SO question: https://stackoverflow.com/questions/23027315/does-application-yml-support-environment-variables

given it's popularity, I think this deserves to be added in the documentation

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
@@ -440,8 +440,8 @@ For example, if a secret named `db.password` is mounted at location `/run/secret
 
 [[features.external-config.files.property-placeholders]]
 ==== Property Placeholders
-The values in `application.properties` and `application.yml` are filtered through the existing `Environment` when they are used, so you can refer back to previously defined values (for example, from System properties).
-The standard `$\{name}` property-placeholder syntax can be used anywhere within a value.
+The values in `application.properties` and `application.yml` are filtered through the existing `Environment` when they are used, so you can refer back to previously defined values (for example, from System properties or environment variables).
+The standard `$\{name:default}` property-placeholder syntax can be used anywhere within a value.
 
 For example, the following file will set `app.description` to "`MyApp is a Spring Boot application`":
 
@@ -449,7 +449,7 @@ For example, the following file will set `app.description` to "`MyApp is a Sprin
 ----
 	app:
 	  name: "MyApp"
-	  description: "${app.name} is a Spring Boot application"
+	  description: "${app.name} is a Spring Boot application written by ${USERNAME:john.doe}"
 ----
 
 TIP: You can also use this technique to create "`short`" variants of existing Spring Boot properties.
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
```diff
@@ -608,6 +608,19 @@ The preceding example would be transformed into these properties:
...
+==== Using environment variables and default values in YAML property definitions
+
+In order to use environment variable as a source for application property, one can use placeholders like this: 
+
+[source,yaml,indent=0,subs="verbatim"]
+----
+	environments:
+	  dev:
+	    url: ${DEV_ENV_URL}
+	    name: ${DEV_ENV_NAME:development-environment}
```
62mkv, line 621: by the way, would be great to see how to use values with spaces, or "in quotes" as default values

wilkinsona (changes requested):
Thanks for the suggestion. We already have a [section in the docs about property placeholders](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.external-config.files.property-placeholders):

> The values in `application.properties` and `application.yml` are filtered through the existing `Environment` when they are used, so you can refer back to previously defined values (for example, from System properties). The standard `${name}` property-placeholder syntax can be used anywhere within a value.

I think this would be better as an addition to this section. Note that there isn't really anything special about environment variables – they're all part of the `Environment` as far as Boot is concerned so perhaps just changing "for example, from System properties" to "for example, from System property and environment variables" would be specific for that side of things.

For the default value, it's a [Spring Framework feature](https://docs.spring.io/spring-framework/docs/5.3.15/reference/html/core.html#beans-value-annotations), but I think it would be worth mentioning it in Boot's docs too. Rather than just adding an example without any explanation, I think it would be better to also explain the `${name:default}` syntax.

62mkv:
@wilkinsona I've tried..

62mkv:
speaking about "explaining the name:default syntax" I would love to understand it in more details.. but can not seem to find where is it defined

wilkinsona:
It's mentioned, very briefly, in the [Spring Framework documentation](https://docs.spring.io/spring-framework/docs/5.3.15/reference/html/core.html#beans-value-annotations) that I linked to [above](https://github.com/spring-projects/spring-boot/pull/30012#pullrequestreview-895365222). In those docs it says:

> It is possible to provide a default value as following:
>
> ```java
> @Component
> public class MovieRecommender {
> 
>     private final String catalog;
> 
>     public MovieRecommender(@Value("${catalog.name:defaultCatalog}") String catalog) {
>         this.catalog = catalog;
>     }
> }

It's also described in the [javadoc of PlaceholderConfigurerSupport](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/config/PlaceholderConfigurerSupport.html) where it says the following:

> Default property values can be defined […] on a property-by-property basis using the value separator which is ":" by default and customizable via [setValueSeparator(String)](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/beans/factory/config/PlaceholderConfigurerSupport.html#setValueSeparator-java.lang.String-).

62mkv:
Indeed.. still am not sure I understand how could we provide more details in this section, other than mentioning the syntax and providing an example. any suggestions?

wilkinsona:
I think that’s all that we should do in Boot’s docs.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/external-config.adoc
wilkinsona: I don't think spaces in the default makes any difference. For example, you can do this:

```
property.example=${some.other.property:The default value}
```

Assuming that `some.other.property` hasn't been set the value of `property.example` will default to `The default value`.

wilkinsona:
@62mkv Thanks very much for making your first contribution to Spring Boot.

wilkinsona (approved):

62mkv:
thanks @wilkinsona !!

PR #30027, opened 2022-03-02
Polish web examples in reference doc

wonwoo:


snicoll:
@wonwoo this is a `Long` attribute so naming it `user` doesn't feel right. That really is the user identifier, hence the current code is better. Thanks for the PR, in any case.

wonwoo:
I think this example is wrong. 
Cannot resolve path variable `user`

```java
@GetMapping("/{user}") 
```
from
```java
@GetMapping("/{userId}") 
```

What do you think?

snicoll:
Thanks for the follow-up. You are right that the mapping is wrong, I overlooked that, sorry. 

I don't think that we should rename the variables as I find them a bit confusing. Rather I'd rename the path variable pattern to use `{userId}` rather than `{user}. Would you be willing to update your PR in that direction?

wonwoo:
please review again!
thank!

wilkinsona:
Thanks very much, @wonwoo.

Issue #30055, opened 2022-03-01
DataFetcherExceptionResolver beans are not considered in GraphQlTester slice tests


JimHosken:
I believe this is a feature request, as I did not see any facilities in the current spring-graphql-test library to accomplish this.

When testing the Controller layer of our application using GraphQlTester, it would be nice to be able to configure the inclusion of our custom DataFetcherExceptionResolver to validate the error handling capabilities we have configured and to assert the ErrorType / error maps in the response.

bclozel:
Thanks @JimHosken this is a bug in the Spring Boot ˋ@GraphQlTest` support. We are not including all the required types in this sliced test.

bclozel:
Hi, this is a `first-timers-only` issue. This means we've worked to make it more legible to folks who either haven't contributed to our codebase before, or even folks who haven't contributed to open source before.

If that's you, we're interested in helping you take the first step and can answer questions and help you out as you do. Note that we're especially interested in contributions from people from groups underrepresented in free and open source software!

If you have contributed before, consider leaving this one for someone new, and looking through our general `ideal-for-contribution` issues. Thanks!

## Problem
Spring Boot helps you to write integration tests without involving the entire application context (all beans, live servers, etc.). These ["test slices"](https://docs.spring.io/spring-boot/docs/2.7.x/reference/html/features.html#features.testing.spring-boot-applications.autoconfigured-tests) are supported with annotations that are specific to the "slice" we'd like to test.
For Spring GraphQL applications, we now provide a [`@GraphQlTest`](https://docs.spring.io/spring-boot/docs/2.7.x/reference/html/features.html#features.testing.spring-boot-applications.spring-graphql-tests) slice.

Part of the test slice support is about filtering what kind of beans we should consider when scanning - rejecting the types that are out of scope for this slice and only keeping the relevant ones for what we're testing. For that, we [a `GraphQlTypeExcludeFilter` imported by the test annotation itself](https://github.com/spring-projects/spring-boot/blob/d47b56a40ed2ec29a1b784a3295d960840ccd4fa/spring-boot-project/spring-boot-test-autoconfigure/src/main/java/org/springframework/boot/test/autoconfigure/graphql/GraphQlTest.java#L85).

[...]

PR #30078, opened 2022-03-07
DataFetcherExceptionResolver beans are not considered in GraphQlTester slice tests

GauthamM-official:
GraphQL slice test - Include all beans used to setup GraphQL source

GraphQlTypeExcludeFilter currently allows JsonComponent, RuntimeWiringConfigurer,
Converter and GenericConverter beans only. But DataFetcherExceptionResolver,
Instrumentation and GraphQlSourceBuilderCustomizer are used in setting up
the GraphQL source. Hence these beans should also be included by default.

See https://github.com/spring-projects/spring-boot/issues/30055


bclozel:
Thanks so much @GauthamM-official for your first contribution!

GauthamM-official:
@bclozel Thanks for the opportunity! Hoping to contribute more in the future.

PR #30107, opened 2022-03-08
Fix redirect for Spring HATEOAS

Buzzardo:
Change the redirect location to point to the Spring HATEOAS part
 of the documentation.


snicoll:
Thanks for the PR @Buzzardo but there is no such anchor.

Issue #30096, opened 2022-03-08
The default value of refresh policy for reactive ES template should be null or false


ToffeeLu:
Check https://github.com/spring-projects/spring-data-elasticsearch/issues/2110 for details

mhalbritter:
As @sothawo said, you're right, and this is a breaking change. We can change this in the next release. Adding it to the documentation should be easier, though. I've created ticket https://github.com/spring-projects/spring-boot/issues/30103 for this.

mhalbritter:
We should remove the `set`-calls in the [auto-configuration](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/elasticsearch/ElasticsearchDataConfiguration.java#L111) and rely on Spring Data's defaults.

wilkinsona:
We can remove this set call as a task in 2.5.x:

https://github.com/spring-projects/spring-boot/blob/44f71a46aa01c90ff0910a27627107e13a282841/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/elasticsearch/ElasticsearchDataConfiguration.java#L111

The default is already `RefreshPolicy.IMMEDIATE` so it is redundant. Removing it will allow us to automatically pick up any future change in Spring Data Elasticsearch.

This set call is a bit trickier:

https://github.com/spring-projects/spring-boot/blob/44f71a46aa01c90ff0910a27627107e13a282841/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/elasticsearch/ElasticsearchDataConfiguration.java#L110

Here we have a different default to Spring Data Elasticsearch. I think we should align but we need to decide when to do that.

We may also want to offer a configuration properties for the template’s settings to make the defaults easier to change without defining a new bean.

sothawo:
Another info, I still have to see how this default setting and a change affects this: When using the operations, i.e. the "low level" operation the default should be to not set anything and so use Elasticsearch 's default (which is not immediate). A user that saves an entity and directly needs it to be accessible would then have to either set the refresh policy or do an explicit refresh() call before doing a search()

 But when using the higher level repository functions the user expects the data to be available after a save. There used to be an explicit call to refresh after writing operations, I think I changed that to do the writing operations with immediate refresh when called from the repository implementation. 

[...]

PR #30117, opened 2022-03-09
Change default value of refresh policy to NONE.

polarbear567:
Fix [issue](https://github.com/spring-projects/spring-boot/issues/30096)


mhalbritter:
Hi, as [mentioned here](https://github.com/spring-projects/spring-boot/issues/30096#issuecomment-1061879608), we want to remove all `set` calls in the autoconfiguration and rely on Spring Boots defaults.

polarbear567:
Sure, will update when I'm free.

mhalbritter:
Please see also the discussion [in the issue](https://github.com/spring-projects/spring-boot/issues/30096#issuecomment-1063735160).

snicoll:
@polarbear567 thanks for the PR but we're still trying to figure out how to approach this. One liner-PR like this are making things harder to track and we can take care of this ourselves.

PR #30157, opened 2022-03-10
NullBean check in ApplicationContextAssert.doesNotHaveBean(String)

deadeuzesse:
getApplicationContext().getBean(name) may return a org.springframework.beans.factory.support.NullBean bean and does not throw the expected NoSuchBeanDefinitionException that validates the assertion. However, this NullBean is not meant to be, therefore it must be checked prior throwing the assertion error.

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/assertj/ApplicationContextAssert.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/assertj/ApplicationContextAssert.java
@@ -198,7 +198,8 @@ public class ApplicationContextAssert<C extends ApplicationContext>
 		}
 		try {
 			Object bean = getApplicationContext().getBean(name);
-			throwAssertionError(new BasicErrorMessageFactory(
+			if(!bean.equals(null))
+				throwAssertionError(new BasicErrorMessageFactory(
 					"%nExpecting:%n <%s>%nnot to have a bean of name:%n <%s>%nbut found:%n <%s>",
 					getApplicationContext(), name, bean));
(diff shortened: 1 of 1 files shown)


snicoll:
@deadeuzesse thanks for the PR but throwing `NoSuchBeanDefinitionException` would be wrong as the bean definition actually exists (the instance is `null`). Can you expand a bit more on the use case? The lack of additional tests in the PR doesn't help from that perspective.

deadeuzesse:
Indeed, point is not to throw a NoSuchBeanDefinitionException since the bean does exists. 
I don't know why the applicationContext contains such bean.
The point is the bean is of type [org.springframework.beans.factory.support.NullBean](https://github.com/spring-projects/spring-framework/blob/main/spring-beans/src/main/java/org/springframework/beans/factory/support/NullBean.java) is not meant to be available:
Initial commit : https://github.com/spring-projects/spring-framework/commit/b94302b5bde7475b9926dacb9fb69f3c8c894444 

According to that, when I assert my context, I expected they won't be available, that's why i considered the ApplicationContextAssert.doesNotHaveBean(String) should have managed such case.

For now i'm using this convolution, it's a bit sad but it works
`// NullBean convolution
  Object myBean= context.getBean("myBean");
  assertTrue(myBean.equals(null)); // that's a NullBean -_-'
`

wilkinsona:
I don't think we should make this change. The context does contain a bean with the requested name so asserting it does not should fail as it currently does. This behaviour is particularly important with `@ConditionalOnMissingBean` and `@ConditionalOnBean` because a bean definition with a `null` value will still affect the matching of those conditions. Thanks anyway.

Issue #29666, opened 2022-02-07
Add auto-configuration for Micrometer 2.0 Observation API


marcingrzejszczak:
# Micrometer 2.0.0 changes 

Micrometer 2.0.0 comes with the following changes

- [Moving part of Micrometer Core to Micrometer API](https://github.com/micrometer-metrics/micrometer/pull/2980)
- [Introduction of the Observation API](https://github.com/micrometer-metrics/micrometer/pull/2992)
- Tracing (see https://github.com/spring-projects/spring-boot/issues/30156)

# Observation API

The most notable change is adding the `ObservationRegistry` interface. `MeterRegistry` implements it so there wouldn't be any change there. The thing that does change is that `MeterRegistry` now has a `observationConfig` configuration that allows to plug in `ObservationHandlers`.

## Required by Boot AutoConfiguration

To create an `Observation` one has to call factory methods on `Observation` (`start` or `createNotStarted`). The `MeterRegistry#ObservationConfig` can have a list of `ObservationPredicate` injected to define whether for given Observation name and metadata an Observation should be created or a NoOp observation should be returned. Example use case is `AutoTimer` in Spring Boot. Now `AutoTimer` is an inbuilt feature of the Observation API. 

## `ObservationHandler`

An `ObservationHandler` allows plugging in into the lifecycle of an Observation. That means that we can react to events such as observation was started, stopped, there was an error etc. 

Micrometer comes with an interface `MeterObservationHandler` whose default implementation is `TimerObservationHandler`. That in turn will create a `Timer` when an Observation was started and will stop it when the Observation was finished.  `MeterRegistry` comes with a method `withTimerObservationHandler` that will by default create this handler. Theoretically there might be case that someone adds Micrometer but doesn't want to create a `Timer` when an Observation is created.

Micrometer Tracing comes with an interface `TracingObservationHandler` that contains additional logic for tracing related handlers.

Micrometer comes with 2 additional interfaces that group `ObservationHandler`s. First is `ObservationHandler.FirstMatchingCompositeObservationHandler` . It will group `ObservationHandler`s and pick the first one whose `supportsContext` method returns `true`.
Second is `ObservationHandler.AllMatchingCompositeObservationHandler`. It will group `ObservationHandler`s and pick all whose `supportsContext` method returns `true`.

After having grouped the handlers they have to be injected to the `ObservationRegistry` via the  `ObservationRegistry#observationConfig()#observationHandler` method.

### Required by Boot AutoConfiguration

[...]

PR #30186, opened 2022-03-11
Add auto-configuration for Micrometer 2.0 Observation API

mhalbritter:
For discussions on the implementation for https://github.com/spring-projects/spring-boot/issues/29666

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/ObservationAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/ObservationAutoConfiguration.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.autoconfigure.observation;
+
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.observation.Observation.GlobalTagsProvider;
+import io.micrometer.observation.ObservationHandler;
+import io.micrometer.observation.ObservationPredicate;
+import io.micrometer.observation.ObservationRegistry;
+import io.micrometer.tracing.handler.TracingObservationHandler;
+
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegistryAutoConfiguration;
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for the Micrometer Observation API.
+ *
+ * @author Moritz Halbritter
+ * @since 3.0.0
+ */
+@AutoConfiguration(after = CompositeMeterRegistryAutoConfiguration.class)
+@ConditionalOnClass(ObservationRegistry.class)
+public class ObservationAutoConfiguration {
+
+	@Bean
+	static ObservationRegistryPostProcessor observationRegistryPostProcessor(
+			ObjectProvider<ObservationRegistryCustomizer<?>> observationRegistryCustomizers,
+			ObjectProvider<ObservationPredicate> observationPredicates,
+			ObjectProvider<GlobalTagsProvider<?>> tagProviders,
+			ObjectProvider<ObservationHandler<?>> observationHandlers,
+			ObjectProvider<ObservationHandlerGrouping> observationHandlerGrouping) {
+		return new ObservationRegistryPostProcessor(observationRegistryCustomizers, observationPredicates, tagProviders,
+				observationHandlers, observationHandlerGrouping);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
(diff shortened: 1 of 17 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/TimerObservationHandlerMeterRegistryCustomizer.java
```diff
@@ -0,0 +1,43 @@
...
+import io.micrometer.observation.ObservationRegistry;
+
+import org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryCustomizer;
+
+/**
+ * Registers the {@link TimerObservationHandler} with a {@link ObservationRegistry}.
+ *
+ * @author Moritz Halbritter
+ */
+class TimerObservationHandlerMeterRegistryCustomizer implements MeterRegistryCustomizer<MeterRegistry> {
```
wilkinsona, line 30: Could this be an `ObservationRegistryCustomizer` that has the `MeterRegistry` passed into its constructor and then configures the `ObservationRegistry` when it's called?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/TracingObservationRegistryCustomizer.java
```diff
@@ -0,0 +1,72 @@
...
+ * {@link NoTracingObservationRegistryCustomizer}, only that
+ * {@link TracingObservationHandler} are grouped into a
+ * {@link FirstMatchingCompositeObservationHandler}.
+ *
+ * This class is used by {@link ObservationAutoConfiguration} when a
+ * {@link TracingObservationHandler} is found on the classpath.
+ *
+ * @author Moritz Halbritter
+ */
+class TracingObservationRegistryCustomizer extends NoTracingObservationRegistryCustomizer {
```
wilkinsona, line 40: Naming is hard, but it feels a little strange that `Tracing…` is a subclass of `NoTracing`. The naming suggests that one is the opposite of the other which, to me at least, means that the inheritance doesn't really fit.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/observation/ObservationAutoConfigurationTests.java
```diff
@@ -0,0 +1,383 @@
...
+import static org.assertj.core.api.Assertions.assertThatThrownBy;
+
+/**
+ * Tests for {@link ObservationAutoConfiguration}.
+ *
+ * @author Moritz Halbritter
+ */
+class ObservationAutoConfigurationTests {
+
+	@Nested
```
wilkinsona, line 61: We've avoided using `@Nested` thus far.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/TimerObservationHandlerMeterRegistryCustomizer.java
mhalbritter: Yup, it's more obvious your way 'round.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/observation/ObservationAutoConfigurationTests.java
mhalbritter: Okay, will flatten the structure or use two classes.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/TracingObservationRegistryCustomizer.java
mhalbritter: Yes, i abused OOP to get the grouping logic right. I refactored the code, it's now using a `ObservationHandlerGrouping` strategy interface. Let me know what you think!

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/HandlersObservationRegistryCustomizer.java
```diff
@@ -0,0 +1,53 @@
...
+package org.springframework.boot.actuate.autoconfigure.observation;
+
+import java.util.List;
+
+import io.micrometer.observation.ObservationHandler;
+import io.micrometer.observation.ObservationHandler.FirstMatchingCompositeObservationHandler;
+import io.micrometer.observation.ObservationRegistry;
+
+/**
+ * Installs {@link ObservationHandler} into an {@link ObservationRegistry}.
```
wilkinsona, line 26: Looks like this needs to be updated following the change to the grouping mechanism.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/ObservationHandlerGrouping.java
```diff
@@ -0,0 +1,40 @@
...
+import io.micrometer.observation.ObservationRegistry.ObservationConfig;
+
+/**
+ * Strategy to apply {@link ObservationHandler ObservationHandlers} to an
+ * {@link ObservationConfig}.
+ *
+ * @author Moritz Halbritter
+ * @since 3.0.0
+ */
+public interface ObservationHandlerGrouping {
```
wilkinsona, line 31: I think this could be package-private.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/ObservationAutoConfiguration.java
```diff
@@ -0,0 +1,105 @@
...
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	ObservationRegistry observationRegistry() {
+		return ObservationRegistry.create();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
```
wilkinsona, line 65: Given that `HandlersObservationRegistryCustomizer` is package-private, users won't be able to define a bean that exercises this condition.

We either need to make `HandlersObservationRegistryCustomizer` public (and that would require `ObservationHandlerGrouping` to remain public) or we should remove this condition. I am tempted to go with the latter for now and keep things locked down. We can relax things in a subsequent milestone if someone needs it and we'd then have a real-world usecase to help us to determine exactly how things should be opened up.

mhalbritter:
Merged in https://github.com/spring-projects/spring-boot/commit/5ab9112bbc639c817dd96aafe6e8a54616d9d18f

PR #30209, opened 2022-03-15
Refactor code to avoid internal LazyPublishArtifact

### LINKED ISSUE #29074: Stop using or fix usage of LazyPublishArtifact for upcoming changes
Spring Boot Gradle Plugin 2.6.1 uses the constructor `LazyPublishArtifact(Provider<?>)`, which is soon to be deprecated in Gradle 7.4. It would be best if this internal API could be avoided, or if not, changed to use `LazyPublishArtifact(Provider<?>, FileResolver)` instead, with a non-null `FileResolver`. The relevant PR for this change in Gradle is https://github.com/gradle/gradle/pull/18715.

octylFractal:
This fixes #29074 by allow `ArtifactHandler` to create the `PublishArtifact` objects, and instead taking in a `TaskProvider` since that's what the callees use.

This appears to pass tests, but I didn't test it on any projects myself as I am not really familiar with Spring Boot.

I targeted this to 2.3.x since that's the oldest still-supported release. Let me know if I should pick a different branch.


wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @octylFractal.

> I targeted this to 2.3.x since that's the oldest still-supported release. Let me know if I should pick a different branch.

2.5.x is our oldest supported branch. I've cherry-picked the changes onto that branch and they'll be in next week's 2.5.11 and 2.6.5 releases.

PR #30329, opened 2022-03-19
Fix JsonSerializer example in reference guide

PPakSang:
If you follow the docs writing custom json serializer, it throws  an exception "JsonGenerationException". Of course, if there is an existing Json Generator, it is applicable code, but users (including myself) who usually follow docs are more likely to write the code first. Therefore, I think that the text should be modified as an example with the addition of JsonGenerator.writeStartObject()/writeEndObject() or a separate notice should exist for that part.


snicoll:
@PPakSang you are correct and that example is wrong. You can see below that is rewritten to use `JsonObjectSerializer` which does exactly what you've mentioned.

snicoll:
@PPakSang thank you for making your first contribution to Spring Boot.

PPakSang:
I was worried that I said something inaccurate because it was my first PR but it works. Thank you for the explanation. I'm really interested in Spring-boot project!!

Issue #28618, opened 2021-11-11
Update links to Spring Security's reference documentation


rwinch:
The docs group has made some changes to how the URLs are going to be handled. Spring Boot should update spring-security attribute to point to  https://spring.io/projects/spring-security/reference

philwebb:
Is it live yet? I get a 404 currently.

wilkinsona:
As discussed previously, it'd also be good to have a versioned URL so that someone reading a certain version of Boot's docs is linked to the matching version of Security's docs.

rwinch:
> Is it live yet? I get a 404 currently.

Sorry it I pointed to the incorrect URL. It should be http://docs.spring.io/spring-security/reference/

> As discussed previously, it'd also be good to have a versioned URL so that someone reading a certain version of Boot's docs is linked to the matching version of Security's docs.

The URL for a specific version is `http://docs.spring.io/spring-security/reference/{version}`

For example, http://docs.spring.io/spring-security/reference/5.6.0-RC1 is the link for 5.6.0-RC1. For now 5.6.0-RC1 is the latest release (there are no GA releases either), so it redirects to the versionless URL.

wilkinsona:
Thanks, Rob.

We currently take the actual Security version and then remove the last `.` and anything that follows it:

https://github.com/spring-projects/spring-boot/blob/313470ac9dc8800263b67846c33c158f1b268dbb/spring-boot-project/spring-boot-docs/build.gradle#L259-L260

I think that'll continue to work although I wonder if it's exactly what we want. https://docs.spring.io/spring-security/reference/5.6.0-SNAPSHOT and https://docs.spring.io/spring-security/reference/5.6.1-SNAPSHOT are not found so it looks like we have to do something for `-SNAPSHOT` versions, but it also sounds like we could leave other versions (`5.6.0-RC1`, `5.6.0`, `5.6.1`, and so on) unchanged and they would work.

Issue #28407, opened 2021-10-21
Spring Security reference doc links not working


dreis2211:
Hi,

I noticed that the links to the Spring Security Reference are not working in the current snapshot docs. https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/htmlsingle/#web.security

I don't know if this is maybe a spring-security issue or if it's just a matter of adjusting to a new location, but maybe you could coordinate/forward this.

Cheers,
Christoph

wilkinsona:
Thanks, @dreis2211. https://spring.io/projects/spring-security#learn links to https://docs.spring.io/spring-security/ which works but it's unversioned so we'd prefer not to use it. The 5.6.0-RC1 javadoc link (https://docs.spring.io/spring-security/docs/5.6.0-RC1/api/) is versioned but it 404s. There are a few other broken links on Spring Security's project page. I'll ask the Security team to take a look.

PR #30349, opened 2022-03-21
Fix links to Spring Security Reference Guide in Accessing the H2 Console in a Secured Application

hpoettker:
Thanks for merging #29932! Configuring Spring Security can be a challenge for beginners, and I think the added paragraph will be helpful.

I checked the deployed snapshot versions of the documentation. Unfortunately, only the 2.5.x version has working links to the Spring Security reference guide. For Spring Boot 2.6.x, they need to be adjusted against the new structure of the Spring Security documentation that came with 5.6: https://docs.spring.io/spring-boot/docs/2.6.5-SNAPSHOT/reference/htmlsingle/#data.sql.h2-web-console.spring-security

This PR is only half the solution, as `{spring-security-docs}` resolves to `https://docs.spring.io/spring-security/reference/5.6.2`. But all links with this particular version do not work while they do e.g. for `5.6.1` or `5.6`. It seem to be the same problem as in #28407 but the fix for that has been basically reverted for #28618. As I don't know, what the best approach for this problem is, the PR doesn't address this.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/data/sql.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/data/sql.adoc
@@ -345,7 +345,7 @@ If your application uses Spring Security, you need to configure it to
 * disable CSRF protection for requests against the console,
 * set the header `X-Frame-Options` to `SAMEORIGIN` on responses from the console.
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, @hpoettker. The docs should be available at https://docs.spring.io/spring-security/reference/5.6.2. I've asked the Security team to take a look.

wilkinsona:
Thanks to @eleftherias, the 5.6.2 are now in place so we can merge this as-is I believe.

wilkinsona:
Thanks very much, @hpoettker.

PR #30356, opened 2022-03-22
Upgrading code using JDK 16's Pattern Matching for instanceof

pruidong:
I used JDK 16's JEP 394: Pattern Matching for instanceof (https://openjdk.java.net/jeps/394), and modified some of the code. If possible, I would like to upgrade the remaining code.


wilkinsona:
Thanks for the pull request. #28181 is tracking this and, as indicated by it being labelled as blocked, we’re not ready to accept contributions for it.

PR #30406, opened 2022-03-24
Prevent org.jetbrains.annotations imports

quaff:


wilkinsona:
Until we tackle https://github.com/spring-projects/spring-boot/issues/10712, I think we should prevent import of Framework's null-safety annotations too. We'll need to make some exceptions for Actuator if we do that as it supports the use of `@Nullable` on endpoint operation parameters.

wilkinsona:
Thanks, @quaff.

PR #30415, opened 2022-03-24
Switch to GA releases for Java 18 update checks

eddumelendez:


dreis2211:
It would make sense to update to https://github.com/spring-projects/spring-boot/blob/main/ci/scripts/detect-jdk-updates.sh#L24 `ga` here if proper GA builds are available now. The automatic JDK update detection jobs will create tickets for the actual updates then.

eddumelendez:
thanks @dreis2211 !

wilkinsona:
Thanks, @eddumelendez. I've only merged the change to the detection script. I'd like to check that detection is working and picks up the new release rather than upgrading manually.

PR #30457, opened 2022-03-29
order ExitCodeGenerator in ExitCodeGenerators by Ordered and @Order b…

dugenkui03:
### Describing Your Changes

Different order in `ExitCodeGenerators#generators` will lead to different exitCode. I suggest add ‘order’ support in order to sort `ExitCodeGenerators#generators` by `Ordered` and `@Order`.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/spring-application.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/spring-application.adoc
@@ -352,6 +352,7 @@ include::code:MyApplication[]
 Also, the `ExitCodeGenerator` interface may be implemented by exceptions.
 When such an exception is encountered, Spring Boot returns the exit code provided by the implemented `getExitCode()` method.
 
+If several `ExitCodeGenerator` are registered in a `ExitCodeGenerators` that must be called in a specific order, you can additionally implement the `org.springframework.core.Ordered` interface or use the `org.springframework.core.annotation.Order` annotation.
 
 
 [[features.spring-application.admin]]
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ExitCodeGenerators.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ExitCodeGenerators.java
@@ -21,6 +21,7 @@ import java.util.Arrays;
 import java.util.Iterator;
 import java.util.List;
 
+import org.springframework.core.annotation.AnnotationAwareOrderComparator;
 import org.springframework.util.Assert;
 
 /**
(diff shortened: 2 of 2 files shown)


wilkinsona:
Thanks for the proposal. Unfortunately, I think this will cause some confusion. `ExitCodeGenerators` currently returns the code that's furthest from zero:

https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ExitCodeGenerators.java#L85-L100

We have tests for this behaviour:

https://github.com/spring-projects/spring-boot/blob/6865f1f3d913a40192570d51857664edd2ed5a48/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/ExitCodeGeneratorsTests.java#L64-L80

Ordering these generators would make no difference to the exit code that's returned.

The only place where ordering would make a difference is if it changed the first generator to return a non-zero value. That non-zero value will decide if the returned code is positive or negative. The other generators will then be called and the value that's further from zero in the direction determined by the first non-zero value will be returned.

Given the above, I don't think we should add ordering support as you have proposed here. We could, perhaps, consider it in combination with a change to return the first non-zero value. That would be a breaking change so we'd have to make it with some care. I'll flag this for a forthcoming team meeting so that we can discuss our options.

dugenkui03:
Combine order with 'return the first non-zero value'  will be a greate idea, it is easy understand.

> That non-zero value will decide if the returned code is positive or negative. 

It seems that the  'positive or negative' of first generator value would not change the return value.


**base**: `-1, -3, 2`, output: 2
```
		generators.add(mockGenerator(-1));
		generators.add(mockGenerator(-3));
		generators.add(mockGenerator(2));
```

**the  'positive or negative' would not change:** `1, -3, 2`, also output: 2
```
		generators.add(mockGenerator(1));
		generators.add(mockGenerator(-3));
		generators.add(mockGenerator(2));
```

**order does**:`1, 2, -3`: output: -3, not 2;
```
		generators.add(mockGenerator(1));
		generators.add(mockGenerator(2));
		generators.add(mockGenerator(-3));
```

wilkinsona:
> It seems that the 'positive or negative' of first generator value would not change the return value.

You're right. I'd misread the code. Thanks for the correction.

philwebb:
We discussed this today and we'd like to order the `ExitCodeGenerator` beans then return the first non-zero result. This will be a breaking change so we'll do it in 2.7.

dugenkui03:
@philwebb This pr is updated, does this work for the new logic: ’order the ExitCodeGenerator beans then return the first non-zero result‘.

wilkinsona:
Thanks very much, @dugenkui03.

PR #30461, opened 2022-03-29
Update list of default internal proxies in Web Server howto

dalbani:
The documentation lags a bit behind the code in terms of default internal proxies.
I was using IPv6 and it took me some time to figure out what was going on.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/webserver.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/webserver.adoc
@@ -444,7 +444,7 @@ If you use Tomcat, you can additionally configure the names of the headers used 
 ----
 
 Tomcat is also configured with a default regular expression that matches internal proxies that are to be trusted.
-By default, IP addresses in `10/8`, `192.168/16`, `169.254/16` and `127/8` are trusted.
+By default, IP addresses in `10/8`, `192.168/16`, `169.254/16`, `127/8`, `172.16/12` and `::1` are trusted.
 You can customize the valve's configuration by adding an entry to `application.properties`, as shown in the following example:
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the suggestion. Rather than updating the list of internal proxies here, which may get out of sync again, I think it would be better to link to [the documentation for `server.tomcat.remoteip.internal-proxies`](https://docs.spring.io/spring-boot/docs/2.5.x/reference/htmlsingle/#application-properties.server.server.tomcat.remoteip.internal-proxies) which lists its default value. While the regular expression isn't quite as easy to read, I think that's outweighed by the fact that it'll always be accurate. If you agree, would you like to update your proposal to make that change?

wilkinsona:
Thanks for the suggestion. Rather than updating the list of internal proxies here, which may get out of sync again, I think it would be better to link to [the documentation for `server.tomcat.remoteip.internal-proxies`](https://docs.spring.io/spring-boot/docs/2.5.x/reference/htmlsingle/#application-properties.server.server.tomcat.remoteip.internal-proxies) which lists its default value. While the regular expression isn't quite as easy to read, I think that's outweighed by the fact that it'll always be accurate. If you agree, would you like to update your proposal to make that change?

dalbani:
Yes, sure, I will do as you suggested. I agree that it's a better to link to the reference documentation.

dalbani:
I have now provided a link to the "Server Properties" section and the name of the property to look for.
I'd have preferred providing a direct link to the property, but I couldn't find out how to do it.  
I tried a reference like `<<application-properties#appendix.application-properties.server.server.tomcat.remoteip.internal-proxies>>` but it couldn't be resolved.  
Probably related to the fact that AsciiDoc sections in `appendix.application-properties` are dynamically generated??

dalbani:
Good to see how you implemented it in the end, thanks for the tip! 👍

PR #30489, opened 2022-03-31
Make ApplicationAvailabilityBean threadsafe

dugenkui03:
`ApplicationAvailabilityBean#onApplicationEvent ` could be invoked concurrently, we should replace HashMap(event map) with ConcurrentHashMap. 

Details in [jdk - `HashMap.java` - L89](https://github.com/openjdk/jdk/blob/3e73a0b726a97df0a4e92f9cf917429346090f45/src/java.base/share/classes/java/util/HashMap.java#L89)

>Note that this implementation is not synchronized. If multiple threads access a hash map concurrently, and at least one of the threads modifies the map structurally, it must be synchronized externally.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/availability/ApplicationAvailabilityBean.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/availability/ApplicationAvailabilityBean.java
@@ -16,8 +16,8 @@
 
 package org.springframework.boot.availability;
 
-import java.util.HashMap;
 import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
@@ -38,7 +38,7 @@ import org.springframework.util.Assert;
 public class ApplicationAvailabilityBean
 		implements ApplicationAvailability, ApplicationListener<AvailabilityChangeEvent<?>> {
 
-	private final Map<Class<? extends AvailabilityState>, AvailabilityChangeEvent<?>> events = new HashMap<>();
+	private final Map<Class<? extends AvailabilityState>, AvailabilityChangeEvent<?>> events = new ConcurrentHashMap<>();
 
 	private final Log logger;
 


mhalbritter:
In which case would this method be invoked concurrently? The [JavaDoc from EventListener](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/event/EventListener.html) states that concurrency is used if the method is annotated with `@Async`, which isn't the case here.

dugenkui03:
> In which case would this method be invoked concurrently?

spring boot allow developer inject `ApplicationAvailabilityBean` in multiple custom beans and publish their custom AvailabilityState event. there is a chance that `ApplicationAvailabilityBean#onApplicationEvent` could be invoked concurrently.

mhalbritter:
It's not expected that users inject the bean and call `onApplicationEvent` directly. If you want to change the application availability status, [you should emit an AvailabilityChangeEvent](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.spring-application.application-availability.managing). The spring event system will then ensure that these listeners are called. And they are not called concurrently.

wilkinsona:
I think we should accept this. While `onApplicationEvent` is being called on one thread by the event multicaster and it mutates the map, another thread could call one of the `get…` methods that reads from the map.

mhalbritter:
Ah, yeah, you're right. I missed that case. Will merge the PR.

dugenkui03:
> ... they are not called concurrently.

@mhalbritter I trace the code from  [AvailabilityChangeEvent](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.spring-application.application-availability.managing), and find there is  nothing to ensure that listeners are not called concurrently —— I can't find any lock or things like `BlockingQueue`.
```java
// 1. multiple thread invoke this static method concurrently and publish different state.
 AvailabilityChangeEvent.publish(this.eventPublisher, ex, LivenessState.BROKEN);

// 2.  call multicastEvent in ‘AbstractApplicationContext#publishEvent(Object, ResolvableType)‘
//      still don't have any lock, 
getApplicationEventMulticaster().multicastEvent(applicationEvent, eventType)

// 3. multiple thread could invoke 'SimpleApplicationEventMulticaster#multicastEvent' at same time and publish different event concurrently.
protected void invokeListener(ApplicationListener<?> listener, ApplicationEvent event) {
                 ...
		doInvokeListener(listener, event);
}

private void doInvokeListener(ApplicationListener listener, ApplicationEvent event) {
	try {
		listener.onApplicationEvent(event);
	}
        ...
}
```

PR #30514, opened 2022-04-04
Make JSON comparisons in tests more robust

chessvivek:
Hi,

The test `customGsonBuilder()` is non deterministic or flaky. `Gson.toJson()` function does not promise any ordering of fields in the returned JSON string (at least not by itself). Therefore, the result of `gson.toJson(new DataObject())` could be `{data: 1, owner: null}` or `{owner: null, data: 1}`. Both of them are the same JSON objects, and ideally, application should not differentiate based on the ordering of JSON fields. These tests may start to fail in the near future and ring a false alarm. Here's a SO post for more details: https://stackoverflow.com/questions/6365851/how-to-keep-fields-sequence-in-gson-serialization.

I ran the test again after my changes to confirm it passes: `./gradlew :spring-boot-project:spring-boot-autoconfigure:test --tests org.springframework.boot.autoconfigure.gson.GsonAutoConfigurationTests.customGsonBuilder`.

Please let me know if you have any questions. There are some more flaky tests in this repository. If you like this fix, I can open more PRs for the other flaky tests.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-autoconfigure/build.gradle
@@ -213,6 +213,7 @@ dependencies {
 	testImplementation("org.junit.jupiter:junit-jupiter")
 	testImplementation("org.mockito:mockito-core")
 	testImplementation("org.mockito:mockito-junit-jupiter")
+	testImplementation("org.skyscreamer:jsonassert")
 	testImplementation("org.springframework:spring-test")
 	testImplementation("org.springframework.kafka:spring-kafka-test")
 	testImplementation("org.springframework.security:spring-security-test")
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/gson/GsonAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/gson/GsonAutoConfigurationTests.java
@@ -18,10 +18,8 @@ package org.springframework.boot.autoconfigure.gson;
 
 import java.time.ZoneId;
 import java.time.ZonedDateTime;
-import java.util.Arrays;
 import java.util.Date;
 import java.util.LinkedHashMap;
-import java.util.List;
 import java.util.Map;
 
 import com.google.gson.ExclusionStrategy;
@@ -31,6 +29,7 @@ import com.google.gson.Gson;
 import com.google.gson.GsonBuilder;
 import com.google.gson.LongSerializationPolicy;
 import org.junit.jupiter.api.Test;
+import org.skyscreamer.jsonassert.JSONAssert;
 
 import org.springframework.boot.autoconfigure.AutoConfigurations;
 import org.springframework.boot.test.context.runner.ApplicationContextRunner;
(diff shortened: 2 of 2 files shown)


snicoll:
Thanks for the PR.

> There are some more flaky tests in this repository.

Can you share more details about that?

chessvivek:
> Thanks for the PR.
> 
> > There are some more flaky tests in this repository.
> 
> Can you share more details about that?

These are some of the other flaky tests that I have found in this project:

org.springframework.boot.autoconfigure.websocket.servlet.WebSocketMessagingAutoConfigurationTests.basicMessagingWithJsonResponse
org.springframework.boot.context.properties.bind.CollectionBinderTests.bindToSetShouldNotAllowDuplicateValues
org.springframework.boot.diagnostics.analyzer.BeanCurrentlyInCreationFailureAnalyzerTests.cycleWithAutowiredFields

I have found these flaky tests using the NonDex tool (https://github.com/TestingResearchIllinois/NonDex), created by Prof. Darko Marinov at the University of Illinois. I have not yet looked into the source of flakiness for these tests, or how to fix them. But I can do that once I am done with this PR.

chessvivek:
There's also this nice library called JSONAssert - https://github.com/skyscreamer/JSONassert that allows us to do JSON to JSON comparison on two JSON strings. In the long run, I think using this library is the "right" way to prevent any non-determinism. I just did not use this library because it will add an extra dependency. But if you would like, I can do definitely do that.

wilkinsona:
Yes, please do use JSONAssert. It is already included in Boot’s dependency management.

chessvivek:
> Yes, please do use JSONAssert. It is already included in Boot’s dependency management.

Thanks for letting me know. I have revised the changes to use the JSONAssert library.

wilkinsona:
@chessvivek Thanks very much for making your first contribution to Spring Boot.

> WebSocketMessagingAutoConfigurationTests.basicMessagingWithJsonResponse

This is the same problem as we had in `GsonAutoConfigurationTests.customGsonBuilder`. I have [fixed it](https://github.com/spring-projects/spring-boot/commit/b88f062f0499fafe487b6e98e3cc9ff9bbbba1a2) as part of merging this PR. Thanks for bringing it to our attention.

> CollectionBinderTests.bindToSetShouldNotAllowDuplicateValues

This looks like a false positive. A `LinkedHashSet` is used so the ordering is deterministic.

> BeanCurrentlyInCreationFailureAnalyzerTests.cycleWithAutowiredFields

I suspect this may be a false positive as well, although it's hard to say without knowing why the tool believes it to be flaky. IIRC, we've never seen this test flake on CI or in local builds.

chessvivek:
> @chessvivek Thanks very much for making your first contribution to Spring Boot.
> 
> > WebSocketMessagingAutoConfigurationTests.basicMessagingWithJsonResponse
> 
> This is the same problem as we had in `GsonAutoConfigurationTests.customGsonBuilder`. I have [fixed it](https://github.com/spring-projects/spring-boot/commit/b88f062f0499fafe487b6e98e3cc9ff9bbbba1a2) as part of merging this PR. Thanks for bringing it to our attention.
> 
> > CollectionBinderTests.bindToSetShouldNotAllowDuplicateValues
> 
> This looks like a false positive. A `LinkedHashSet` is used so the ordering is deterministic.
> 
> > BeanCurrentlyInCreationFailureAnalyzerTests.cycleWithAutowiredFields
> 
> I suspect this may be a false positive as well, although it's hard to say without knowing why the tool believes it to be flaky. IIRC, we've never seen this test flake on CI or in local builds.

Thanks @wilkinsona for accepting my fix. I see that you enabled the strict mode by passing `true` as the third argument in `JSONAssert.assertEquals`. However, the order is only ignored when you disable the strict mode. In the documentation, it says - `When strict is set to false (recommended), it forgives reordering data and extending results (as long as all the expected elements are there), making tests less brittle.` 

If you just want to forgive reordering data and not extending results, you can also pass in `JSONCompareMode.NON_EXTENSIBLE` as the third argument. The third argument can also take a `JSONCompareMode`, the documenation for which you can see here - https://jsonassert.skyscreamer.org/apidocs/index.html

wilkinsona:
As there are no arrays involved, I believe strict mode meets our needs. It is described as follows:

> Strict tests require all of the elements requested to be returned, and only those elements (ie, the tests are non-extensible). Arrays of elements must be returned in the same order as expected.

I verified this be temporarily changing the order of the expected json and the test passed.

PR #30522, opened 2022-04-04
Move Jetty 9 specific exclusions to the correct dependency

izeye:
This PR fixes Jetty 9 specific dependency exclusions in doc.


wilkinsona:
Thanks very much, @izeye.

PR #30535, opened 2022-04-05
add Tomcat locale mapping for Japanese to preserve UTF-8 charset

chrisrhut:
Tomcat 9.0.59 [added](https://github.com/apache/tomcat/commit/ad8d837ea335928d74833509e4f96747b97b503b) a charset mapping for Japanese to default to `Shift_JIS` encoding.

This results in corrupted rendering if the client expects UTF-8 and can be easily seen by setting the `Accept-Language: ja` header (by changing the primary language in Chrome settings to Japanese).

The fix is to apply the same override mapping in `TomcatServletWebServerFactory.java` that already existed for `en` and `fr` (in those cases, to override `ISO-8559-1` to `UTF-8`):

**index.html:**
```html
<!DOCTYPE html>
<html lang="ja">
<head>
  <meta charset="UTF-8">
  <title>こんにちは世界</title> <!-- "Hello, World" -->
</head>
<body>
管理者画面 <!-- "Admin" -->
</body>
</html>
```


snicoll:
@chrisrhut thank you for the PR. I've removed the screenshots as I believe they're adding noise to the report.

mhalbritter:
Thanks for your contribution, i merged it.

Issue #30560, opened 2022-04-06
allow CassandraAutoConfiguration to be disabled with a property


wakingrufus:
In my organization's microservices, we do our own wiring for the datastax cassandra driver, so I would like to disable spring boot's auto configuration of the cassandra driver.
Because `CassandraAutoConfiguration` is only conditional on `CqlSession` existing on the classpath, the only way to do this is via `@SpringBootApplication(exclude = {CassandraAutoConfiguration.class})`.

However, we have a custom `ApplicationContextInitializer` that pre-sets some environment properties for our environment, so if there was a way to set a property to disable `CassandraAutoConfiguration` we would not need the boilerplate to disable in each and every microservice.

I propose a `@ConditionalOnProperty` for `CassandraAutoConfiguration` which defaults to true.
I will be opening a PR for this shortly.

wilkinsona:
Duplicates #30561.

PR #30561, opened 2022-04-06
Add an environment property toggle for CassandraAutoConfiguration

wakingrufus:
Add an environment property toggle via @ConditionalOnProperty for CassandraAutoConfiguration which defaults to true (#30560)


wilkinsona:
Generally speaking, we avoid having enable properties for auto-configuration as it duplicates what can be achieved by excluding the auto-configuration. Have you considering setting [`spring.autoconfigure.exclude`](https://docs.spring.io/spring-boot/docs/2.6.x/reference/htmlsingle/#application-properties.core.spring.autoconfigure.exclude) as [described in the documentation](https://docs.spring.io/spring-boot/docs/2.6.x/reference/htmlsingle/#using.auto-configuration.disabling-specific)?

wakingrufus:
oh, I was not aware of the `spring.autoconfigure.exclude` property. I will try this out, thanks!

wakingrufus:
This worked for me. Thanks @wilkinsona !

PR #30562, opened 2022-04-06
Add missing configuration metadata for "management.endpoint.health.probes.add-additional-paths"

luozhenyu:
According to the documentation of spring-boot-actuator, we can define "management.endpoint.health.probes.add-additional-paths=true" to enable `liveness` and `readiness`. However, the configuration metadata for this is missing.

> If your Actuator endpoints are deployed on a separate management context, the endpoints do not use the same web infrastructure (port, connection pools, framework components) as the main application. In this case, a probe check could be successful even if the main application does not work properly (for example, it cannot accept new connections). For this reason, is it a good idea to make the liveness and readiness health groups available on the main server port. This can be done by setting the following property:
> ```java
> management.endpoint.health.probes.add-additional-paths=true
> ```
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#actuator.endpoints.kubernetes-probes

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -49,6 +49,12 @@
       ]
     },
     {
+      "name": "management.endpoint.health.probes.add-additional-paths",
+      "type": "java.lang.Boolean",
+      "description": "Whether to make the liveness and readiness health groups available on the main server port.",
+      "defaultValue": false
+    },
+    {
       "name": "management.endpoint.health.probes.enabled",
       "type": "java.lang.Boolean",
       "description": "Whether to enable liveness and readiness probes.",


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -54,6 +54,12 @@
       "description": "Whether to enable liveness and readiness probes.",
       "defaultValue": false
     },
+    {
```
snicoll, line 57: This should come before the previous entries as we sort them by alphabetical order.

snicoll (commented):
Thanks for the PR. I've left a comment for your consideration.

I believe this change was introduced in `2.6.x` so the  current base branch doesn't look right.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
luozhenyu: Thanks, I have rearranged it.

luozhenyu:
> Thanks for the PR. I've left a comment for your consideration.
> 
> I believe this change was introduced in `2.6.x` so the current base branch doesn't look right.
@snicoll You are right. I changed the base branch to`2.6.x`.

snicoll:
@luozhenyu thank you for the quick follow-up and for making your first contribution to Spring Boot.

Issue #29080, opened 2021-12-16
Reactor Kafka support


thepaep:
As said in https://github.com/reactor/reactor-kafka/issues/100#issuecomment-995178795 native support for Reactor Kafka would be nice to lots of users. Spring Boot currently supports WebFlux, but auto-configuration and more capabilities aren't supported yet with Reactor Kafka. As others said in different issues, I also think the connection between Reactor Kafka and Spring Boot should be more documented and supported since it's the most basic usage of Reactor Kafka. There is currently no official way of using Reactor Kafka with Spring Boot, which is pretty odd.

almogtavor:
Can someone describe what features are currently missing on this subject and how should they be implemented (what functionalities should be added and where exactly), so it'll be easier for contributors to help with this?

lynch19:
+1

lynch19:
@snicoll @simonbasle @garyrussell any help with this?

roger751:
+1

snicoll:
@lynch19 @roger751 please use the reaction on the original description rather than +1 comments like this.

garyrussell:
@lynch19 reactor-kafka has two fundamental properties objects

https://github.com/reactor/reactor-kafka/blob/main/src/main/java/reactor/kafka/receiver/ReceiverOptions.java

https://github.com/reactor/reactor-kafka/blob/main/src/main/java/reactor/kafka/sender/SenderOptions.java

Each with a number of properties.

In addition, creation of these takes a `Map<String, Object>` of regular Kafka Properties (similar to the ...properties extension for the spring-kafka auto configuration).

There are several levels of Boot auto configuration that would be useful, with the MVP being the auto-configuration of these two beans using application properties.

AntonLGVS:
I'm using the reactive kafka in my applications. The Reactor kafka provides everything you need to work. Or does it means support in the form of annotations?
For example:
```java
@ReactiveKafkaListener(topic = "smth", autoCommit=true)
public Mono<Void> consumer(DTO dto) {
    log.info("received dto: " + dto);
    return service.call(dto).then();
}
```

garyrussell:
No; not at all; this is just about configuring the sender and receiver options via Boot properties.

Such a mechanism would belong in Spring for Apache Kafka (spring-kafka), but there are currently no plans to do so.

ipanimnim:
@garyrussell where in the Spring Boot project files hierarchy would you add the auto-configuration modules for Reactor Kafka?

[...]

Issue #18804, opened 2019-10-29
Provide auto-configuration for Spring for Apache Kafka's reactive support


rycee:
To produce a `ReactiveKafkaProducerTemplate` I have to explicitly include the `ProducerConfig.TRANSACTIONAL_ID_CONFIG` property as per the following example
```java
@Bean
public ReactiveKafkaProducerTemplate<String, byte[]> reactiveKafkaProducerTemplate(
    KafkaProperties properties) {
  Map<String, Object> props = properties.buildProducerProperties();
  props.put(
      ProducerConfig.TRANSACTIONAL_ID_CONFIG, properties.getProducer().getTransactionIdPrefix());
  return new ReactiveKafkaProducerTemplate<>(SenderOptions.create(props));
}
```
I would have expected this property to be included by default when calling `properties.buildProducerProperties()` making
```java
@Bean
public ReactiveKafkaProducerTemplate<String, byte[]> reactiveKafkaProducerTemplate(
    KafkaProperties properties) {
  Map<String, Object> props = properties.buildProducerProperties();
  return new ReactiveKafkaProducerTemplate<>(SenderOptions.create(props));
}
```
work.

The transactional ID prefix property was added in #11076.

wilkinsona:
When configured only at the property level, my understanding is that `TRANSACTION_ID_CONFIG` is useless without else setting `ENABLE_IDEMPOTENCE_CONFIG` to `true`. From the docs for the former:

> Note that enable.idempotence must be enabled if a TransactionalId is configured.

As its name suggests, the `transaction-id-prefix` configuration property was added to map to the transaction ID prefix of `DefaultKafkaProducerFactory`. The producer factory automatically enables idempotent behaviour when a transaction ID prefix is set.

Like all of Boot's own `@ConfigurationProperties` classes, we also consider `KafkaProperties` to be [internal](https://docs.spring.io/spring-boot/docs/2.2.0.RELEASE/reference/htmlsingle/#boot-features-external-config-java-bean-binding):

> Similar to auto-configuration classes, `@ConfigurationProperties` classes available in Spring Boot are for internal use only. The properties that map to the class, which are configured via properties files, YAML files, environment variables etc., are public API but the content of the class itself is not meant to be used directly.

As such, you shouldn't be using it for your own purposes and should use your own `@ConfigurationProperties` class instead.

[...]

Issue #18751, opened 2019-10-25
Provide auto-configuration for Reactor Kafka


jntakpe:
Spring Boot provides a great helper configuration [KafkaProperties](https://github.com/spring-projects/spring-boot/blob/master/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java) class for Kafka. 

This class includes two `spring-kafka` module classes : 

```java
import org.springframework.kafka.listener.ContainerProperties.AckMode;
import org.springframework.kafka.security.jaas.KafkaJaasLoginModuleInitializer;
```

Due to those imports, in order to leverage Spring Boot's Kafka auto configuration, it is required to import `spring-kafka` module which is largely increasing the application size (our projects are using Reactor Kafka).

IMHO, the configuration that depends on `spring-kafka` should be moved to `spring-kafka` module, allowing Spring Boot's projects to leverage automatic type safe Kafka configuration without importing `spring-kafka`.

wilkinsona:
Thanks for the suggestion. The whole auto-configuration, of which `KafkaProperties` is a part, depends on Spring for Apache Kafka. In other words, it's intentionally auto-configuration for Spring for Apache Kafka and not for Kafka itself. The auto-configuration will back off completely in the absence of Spring for Apache Kafka. Also, you shouldn't try to use `KafkaProperties` in your own code. As [noted in the documentation](https://docs.spring.io/spring-boot/docs/2.2.0.RELEASE/reference/htmlsingle/#boot-features-external-config-typesafe-configuration-properties), all configuration properties classes are for internal use only:

> Similar to auto-configuration classes, `@ConfigurationProperties` classes available in Spring Boot are for internal use only. The properties that map to the class, which are configured via properties files, YAML files, environment variables etc., are public API but the content of the class itself is not meant to be used directly.

I don't think it makes sense for Spring Boot to offer configuration for Kafka without using Spring for Apache Kafka. Our opinion is that it makes sense to use the latter when interacting with Kafka in a Spring application. If you have use cases where that's not true, please share them here and we can reconsider or pass them on to the Spring for Apache Kafka team for them to look at.

jntakpe:
[...]

PR #30567, opened 2022-04-06
Add auto-configuration for Reactor Kafka

almogtavor:
Some notes:

* This PR resolves https://github.com/spring-projects/spring-boot/issues/29080 and https://github.com/spring-projects/spring-boot/issues/18804.
* I've put the Reactor Kafka autoconfiguration under the `kafka` hierarchy and not `reactor.kafka` since there is no other component that its reactive equivalent has its folder. This can also be easier when a so-called "common kafka autoconfigurations superclass" will be implemented. This can also be easier to maintain when changes will be needed to get implemented since both of the autoconfigurations will sit at the same place.
* The common Kafka autoconfigurations superclass has not been implemented in this PR, and therefore `ReactiveKafkaProperties` has a static `Properties` class of its own.
* This PR is not resolving the need for Spring Kafka when using Reactor Kafka, i.e. [this issue](https://github.com/spring-projects/spring-boot/issues/18751). To resolve this we need to extract all of the common properties that both of the clients use and expose them to both of them.
* Since basic fields like `bootstrapServers` are defined in `KafkaProperties`, `ReactiveKafkaProperties` is annotated with `@ConditionalOnBean(KafkaProperties.class)`.
* The autoconfiguration functions use the `ReceiverOptions`'s & `SenderOptions`'s setters with null checks\ greater than zero checks, and not just building of a class like `ImmutableReceiverOptions` since this class is internal and not public.
* I've been creating customizer interfaces, but I wasn't sure about the classes that I should customize since the `DefaultKafkaReceiver` & `DefaultKafkaSender` are classified as internals.

There is no autoconfiguration for the `scheduler`, `assignListeners`, `revokeListeners` and `schedulerSupplier` fields. 
* Since `reactor.kafka.sender.ImmutableSenderOptions` is not public it is currently hard to pass fields like Scheduler since it is being read from the configuration file as a field of type `Class<?>`, which cannot be easily passed to `senderOptions.scheduler()`.
* `assignListeners` and `revokeListeners` seemed unnecessary to be configured from a YAML, and even if we'd wanted to, I think it'd be pretty hard to accomplish.
* `schedulerSupplier` also seemed unnecessary to me for autoconfigurations.
* I couldn't bind `assignTopicPartitions` due to problems with binding a `String` & `int` KV pair from the YAML file as a `TopicPartition` class.
* I thought of throwing an error when no topics have been configured, but I gave up on this since it can damage people who have the autoconfiguration on the classpath but don't use it.


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaProperties.java
```diff
@@ -0,0 +1,341 @@
...
+	}
+
+	public static class Receiver {
+
+		/**
+		 * Sets the timeout for each {@link KafkaConsumer#poll(Duration)} operation. Since
+		 * the underlying Kafka consumer is not thread-safe, long poll intervals may delay
+		 * commits and other operations invoked using
+		 * {@link KafkaReceiver#doOnConsumer(java.util.function.Function)}. Very short
+		 * timeouts may reduce batching and increase load on the broker.
```
artembilan, line 110: The description is these configuration properties must be short and simple.
And it cannot be like a method Javadoc: we must not use any Javadoc tag and style in this descriptions.
This property indeed doesn't set anything.
See docs for more info: https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.developing-auto-configuration.custom-starter.configuration-keys

artembilan (changes requested):
I see you auto-configure only options. Why there are no auto-configured beans for `KafkaReceiver` and `KafkaSender`? Or even their respective Spring for Apache Kafka wrappers: `ReactiveKafkaConsumerTemplate` and `ReactiveKafkaProducerTemplate` ?

Thanks for contribution anyway!

almogtavor:
@artembilan That's actually a great idea. I've auto configured options basically because that's what we agreed on the issue. That's can be pretty easy to migrate this for an auto configuration to the templates. Would you suggest to still keep the auto configuration for the options (in case people would like to customize stuff)?

artembilan:
Sure! I will leave that decision to @garyrussell since it looks like exactly he advised to auto-configure only options.
But the feature doesn't look complete without something what we already can use use for sending or received to/from Kafka.
I just had in mind over here that auto-configured `KafkaTemplate` and some infrastructure for consumer...

garyrussell:
I suggested configuring just the options because that is the biggest pain point - Boot already has mechanisms to configure the properties, which are identical to spring-kafka's; I don't see much benefit in creating the sender and receiver because they can't be reused and it only needs one more line for the user to write; so it doesn't save a lot of boilerplate.

In most cases

```java
KafkaReceiver.create(options)
    .receive()
    . ...
    .subscribe()
```
The idea here was to get the minimal auto config and have the community request more as time goes by.

Those `Reactive....Template` are very lightweight wrappers and don't currently add much value; I hope to work on tighter integration between reactor-kafka and spring-kafka later this year; and then the auto configuration can evolve further.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaAutoConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+			ReactiveKafkaProperties reactiveKafkaProperties) {
+		this.kafkaProperties = kafkaProperties;
+		this.reactiveKafkaProperties = reactiveKafkaProperties;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(ReceiverOptions.class)
+	public <K, V> ReceiverOptions<K, V> receiverOptions() {
+		Map<String, Object> properties = this.kafkaProperties.buildConsumerProperties();
+		properties.putAll(this.reactiveKafkaProperties.buildReceiverProperties());
```
garyrussell, line 65: What is the purpose of this? It is adding reactor-kafka properties to the Kafka receiver properties; Kafka knows nothing about these properties.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaAutoConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+			}
+		}
+		return receiverOptions;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(SenderOptions.class)
+	public <K, V> SenderOptions<K, V> senderOptions() {
+		Map<String, Object> properties = this.kafkaProperties.buildProducerProperties();
+		properties.putAll(this.reactiveKafkaProperties.buildSenderProperties());
```
garyrussell, line 114: Ditto - we shouldn't be adding these; they are reactor-kafka specific properties.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaProperties.java
```diff
@@ -0,0 +1,341 @@
...
+
+		public int getMaxDeferredCommits() {
+			return this.maxDeferredCommits;
+		}
+
+		public void setMaxDeferredCommits(int maxDeferredCommits) {
+			this.maxDeferredCommits = maxDeferredCommits;
+		}
+
+		public Map<String, Object> buildProperties() {
```
garyrussell, line 238: We don't need any of these `build*Properties()` methods; all the kafka specific properties are already handled by parts of `KafkaProperties`.

garyrussell (changes requested):
We really need to pull `Producer`, `Consumer` etc in KafkaProperties to their own class(es) and use them in both places; these represent native Kafka Properties.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaAutoConfiguration.java
```diff
@@ -0,0 +1,128 @@
...
+			receiverOptions = receiverOptions.commitInterval(commitInterval);
+		}
+		Collection<String> subscribeTopics = this.reactiveKafkaProperties.getReceiver().getSubscribeTopics();
+		if (subscribeTopics != null) {
+			receiverOptions = receiverOptions.subscription(subscribeTopics);
+		}
+		else {
+			Pattern subscribePattern = this.reactiveKafkaProperties.getReceiver().getSubscribePattern();
+			if (subscribePattern != null) {
+				receiverOptions = receiverOptions.subscription(subscribePattern);
```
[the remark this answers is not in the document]
almogtavor: Great idea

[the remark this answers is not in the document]
almogtavor: @artembilan what do you think?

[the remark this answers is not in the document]
artembilan: Sorry for delay.

I don't think it is OK to propagate them blindly.
I might not be interested in subscription, so I won't provide that `receiverProperties.getSubscribePattern()` and we will fail here with the `Objects.requireNonNull(pattern)`.
Therefore `PropertyMapper` is better way to accept those props which are expected and valid.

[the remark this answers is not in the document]
almogtavor: 👍

almogtavor:
> We really need to pull `Producer`, `Consumer` etc in KafkaProperties to their own class(es) and use them in both places; these represent native Kafka Properties.

You'd like to do this in this PR or on another one? I think it would be better to get done on another PR since it changes `KafkaProperties` and other classes that are unrelated to this PR.

Also, @artembilan  & @garyrussell, I've resolved your notes, you're welcome to review the PR once more.

artembilan:
> We really need to pull Producer, Consumer etc in KafkaProperties to their own class

Isn't that going to be a breaking change? Therefore such a fix cannot be applied to the current Spring Boot `2.7.x` and can be aimed only for the next major `3.0`.
Yes, for the number of changes it really might be better to do that in the separate PR.

Do I miss anything else, @garyrussell ?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaAutoConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for the Reactive client of Apache
+ * Kafka.
+ *
+ * @author Almog Tavor
+ * @since 2.7.0
+ */
+@AutoConfiguration
+@ConditionalOnClass({ KafkaReceiver.class, KafkaSender.class })
+@ConditionalOnBean(KafkaProperties.class)
```
artembilan, line 46: I guess something like `@AutoConfigureAfter(KafkaAutoConfiguration.class)` is better otherwise we cannot predict the order of auto-configurations.

artembilan (changes requested):
Otherwise this is OK with me.

Thanks

almogtavor:
@garyrussell the only change that there is to do is to move the `KafkaProperties.Consumer` and `KafkaProperties.Producer` into separate classes & adjust this right?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaAutoConfiguration.java
almogtavor: Good idea

garyrussell:
I would say so, yes.

But, I don't know enough about Boot's property documentation as to whether such a change would break things, so we need input from the Boot team.

almogtavor:
I think this would only break native usage of the `KafkaProperties` that shouldn't really happen. Anyway, would you prefer to do this on another PR?

almogtavor:
@garyrussell wdyt?

garyrussell:
As I said

>But, I don't know enough about Boot's property documentation as to whether such a change would break things, so we need input from the Boot team.

This has missed the 2.7.x train anyway so there is now plenty of time to resolve this.

almogtavor:
Ok. I'll add the commits for the classes split to this PR in the following days

wilkinsona:
Thanks for the PR, @almogtavor. A few general comments:

- I'm a bit confused by the apparent overlap or at least similarity between some of the existing properties and those that are being proposed here. For example `spring.kafka.consumer.auto-commit-interval` and `spring.reactor.kafka.receiver.commit-interval`, `spring.kafka.properties.*` and `spring.reactor.kafka.properties.*`
- Given that there is no `spring.reactor.kafka.consumer.properties.*`, there seems to be an imbalance between `spring.reactor.kafka.properties.*` and `spring.reactor.kafka.sender.properties.*`
- Given the reuse of the `spring.kafka.producer.*` and `spring.kafka.consumer.*` properties to configure Reactor Kafka, would `spring.kafka.reactor` be a better prefix as it brings the property names closer together?

I think things would benefit from taking a small step back and considering exactly how we want the receiver and sender options to be configured. Once that's been established, we can then map that into properties classes and auto-configuration.

onobc:
> I would say so, yes.
> 
> But, I don't know enough about Boot's property documentation as to whether such a change would break things, so we need input from the Boot team.

If we decided to "share" the consumer/producer properties they would be moved out to separate classes and then each instance var to the `Consumer/Producer` would be marked w/ [@NestedConfigurationProperty](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#appendix.configuration-metadata.annotation-processor.automatic-metadata-generation.nested-properties) and the annotation processor should handle it well. A good example is [ServerProperties.java](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java#L112).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaAutoConfiguration.java
```diff
@@ -0,0 +1,118 @@
...
+				receiverOptions);
+		if (Optional.ofNullable(receiverProperties.getSubscribeTopics()).isEmpty()) {
+			receiverOptions = setPropertyWhenNonNull(receiverProperties.getSubscribePattern(),
+					receiverOptions::subscription, receiverOptions);
+		}
+		return receiverOptions;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(SenderOptions.class)
```
onobc, line 92: The bean type is redundant here as the return type of the method is the default.

onobc:
> Thanks for the PR, @almogtavor. A few general comments:
> 
> * I'm a bit confused by the apparent overlap or at least similarity between some of the existing properties and those that are being proposed here. For example `spring.kafka.consumer.auto-commit-interval` and `spring.reactor.kafka.receiver.commit-interval`, `spring.kafka.properties.*` and `spring.reactor.kafka.properties.*`
> * Given that there is no `spring.reactor.kafka.consumer.properties.*`, there seems to be an imbalance between `spring.reactor.kafka.properties.*` and `spring.reactor.kafka.sender.properties.*`
> * Given the reuse of the `spring.kafka.producer.*` and `spring.kafka.consumer.*` properties to configure Reactor Kafka, would `spring.kafka.reactor` be a better prefix as it brings the property names closer together?
> 
> I think things would benefit from taking a small step back and considering exactly how we want the receiver and sender options to be configured. Once that's been established, we can then map that into properties classes and auto-configuration.

@wilkinsona I will reply w/ what I see in regards to the overlap of the properties and also stepping back and thinking about how we want to configure this thing....

> ℹ️  I talk mostly of Receiver but the Sender follows the same vein of thought...

So, the pattern is that all of the properties in the "ReceiverOptions.properties" map and the key and value serializer property
are passed into the KafkaConsumer. 

```java
public <K, V> Consumer<K, V> createConsumer(ReceiverOptions<K, V> config) {
    return new KafkaConsumer<>(config.consumerProperties(),
                               config.keyDeserializer(),
                               config.valueDeserializer());
}
```
These are the only "true" overlapping ones. All of the other options are used outside of the `KafkaConsumer` in the reactive control layer. You can see the usage in [ConsumerEventLoop](https://github.com/reactor/reactor-kafka/blob/main/src/main/java/reactor/kafka/receiver/internals/ConsumerEventLoop.java#L117).

I am guessing they chose not to keep the key/value consumer property map entry and the 1st class key/value serde properties in sync since they are both passed into the KafkaConsumer constructor. We could do the same, or sync them if we want. 

I remember when we I used `ReactorKafka` in the past, it is confusing as to what "levers" to configure (the native Kafka consumer ones or the ReactorKafka ones) to get the behavior you want. The ref docs for ReactorKafka do talk about this IIRC. I think we should stay out of that business and simply let the 1st class options properties exist where they do and not try to consolidate/reconcile for example w/ the 
`spring.kafka.consumer.auto-commit-interval and spring.reactor.kafka.receiver.commit-interval`.

### Possible direction

To summarize what I think I am hearing and also what I think would be a good direction: 

- Prefix the RKP as 'spring.kafka.reactor'
- Do not leverage the KP class in the RKP any longer
- Break out Consumer/Producer properties into own class (leverage `NestedConfigurationProperty`) in KP and RKP
- Add buildConsumer/ProducerProeprties in RKP to get common props such as bootstrap etc..
- Mirror the consumer/producer properties directly as done in the receiver/sender options (aka don't try to reconcile similiar properties)

Tradeoff of the split approach is that the consumer properties need to be mapped multiple times if someone is
using both normal and reactive Kafka. I would think the "levers" would be slightly different between them anyways though.

An example yaml would look like:
```yaml
spring:
  kafka:
    bootstrap-servers: localhost:9092   
    consumer:
      auto-offset-reset: earliest
      auto-commit-interval: 5s
    reactive:
      consumer:
        auto-offset=reset: latest
        auto-commit-interval: 3s
```

Thoughts?

onobc (changes requested):

wilkinsona:
Thanks for sharing your thoughts, @onobc.

> and not try to consolidate/reconcile for example w/ the
`spring.kafka.consumer.auto-commit-interval` and `spring.reactor.kafka.receiver.commit-interval`.

I'm not too comfortable with that. Once Boot surfaces the properties, IDEs will offer auto-completion for them which increases the chances of two similarly named properties causing confusion for users and being difficult to use as a result. If the two properties serve the same purpose, I think they should be consolidated into a single property. If the two properties serve different purposes, I think they should be renamed. In either case, I suspect some changes in Spring Kafka and/or Reactor Kafka will be necessary so that the property or properties map onto similar settings.

onobc:
> If the two properties serve the same purpose, I think they should be consolidated into a single property. If the two properties serve different purposes, I think they should be renamed

@wilkinsona that makes sense. So a deeper level of understanding of these "overlapping" properties is required to understand what to do for each one (consolidate, rename, leave-alone, etc.).  I will try to take a scan today and get a list of these properties so we can start discussing.  

Other than the few questionable properties, wdyt of the other points: 

> - Prefix the RKP as 'spring.kafka.reactor'
> - Do not leverage the KP class in the RKP any longer
> - Break out Consumer/Producer properties into own class (leverage `NestedConfigurationProperty`) in KP and RKP
> - Add buildConsumer/ProducerProeprties in RKP to get common props such as bootstrap etc..
> - ~Mirror the consumer/producer properties directly as done in the receiver/sender options (aka don't try to reconcile similiar properties)~

wilkinsona:
I think `spring.kafka.reactor` makes sense as the property prefix. For the others, I think it's too soon to say. I think we need to know what we want the auto-configuration to offer before we spend any more time thinking about the precise details. What's offered should be defined as the beans that will be auto-configured and the properties that will be available to control their configuration. Once we know that, we can figure out how to implement it.

Let's continue the discussion on #29080 rather than here please. Once it's reach a conclusion we can decide if this PR is roughly applicable or if a different approach is needed.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaProperties.java
```diff
@@ -0,0 +1,333 @@
...
+    }
+
+    public Receiver getReceiver() {
+        return this.receiver;
+    }
+
+    public Sender getSender() {
+        return this.sender;
+    }
+
```
adavid9: nit: Unnecessary new line

philwebb:
We're going to close this PR for now until we've had time to do the design work required in #29080. Thanks for the efforts so far, we can reopen if necessary when we're further along with the design.

Issue #17113, opened 2019-06-12
Support for yaml 1.2 specification?


KiranMohan:
Does spring boot 2.1.x support yaml 1.2 specification?

If not, is it possible to use yaml 1.2 + snakeyaml-engine for the spring boot application.yml support?

I have a requirement to add some json content within application.yml to use as is at runtime.

wilkinsona:
> Does spring boot 2.1.x support yaml 1.2 specification?

No. Spring Boot uses `org.yaml:snakeyaml` which does not support YAML 1.2. 

> If not, is it possible to use yaml 1.2 + snakeyaml-engine for the spring boot application.yml support?

Not at the moment, no. `snakeyaml-engine` has a completely different API so would have to be supported separately.

> I have a requirement to add some json content within application.yml to use as is at runtime.

This sounds like it would be YAML-specific and would not work well with application.properties, command line arguments, system properties, etc. Generally speaking we recommend that a configuration property is used to specify a file which contains the more complex data. This files then read at runtime and its content used as needed. This is what Boot itself does for configuring SSL trust and key stores, for example.

Does that not meet your needs?

KiranMohan:
> Generally speaking we recommend that a configuration property is used to specify a file which contains the more complex data. 

This is what I would do now until Spring supports yaml 1.2 for application.yml

> Does that not meet your needs?
Yes. I evaluating all available options.

asomov:
@wilkinsona feel free to ping me if you need support when you decided to go to YAML 1.2 with SnakeYAML Engine. I have already some preparations.

gy-mate:
> Not at the moment, no. snakeyaml-engine has a completely different API so would have to be supported separately.

@wilkinsona @philwebb I would recommend supporting it. Rookie mistake on my side, but I've spent several hours trying to figure out why a `@Value` injection assigned a wrong value (`1067` instead of `00002053`) to a `String` field. Turns out, it parsed `00002053` as an octal number, then casted that number in decimal (`1067`) to a `String`.

YAML `1.2` fixed this by requiring `Oo` at the beginning of octal integers.

I know, if I want to use the number in the YAML as a `String`, I should've put it between apostrophes, but this whole thing could've been avoided with YAML `1.2`.

Related questions:
https://stackoverflow.com/questions/42828311/value-from-application-yml-returns-wrong-value
https://stackoverflow.com/questions/56558270/does-spring-boot-2-1-x-support-yaml-1-2-specification

PR #30626, opened 2022-04-10
Remove unnecessary quotes from examples

michaldo:
`include: "readinessState,customCheck"` > `include: readinessState,customCheck`

I did not find Spring Boot recommendations for YAML style, then I assume default should be applied: https://yaml.org/spec/1.2.2/

I think readability should be promoted 

 > The plain (unquoted) style has no identifying indicators and provides no form of escaping. It is therefore the most readable

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/endpoints.adoc
@@ -736,7 +736,7 @@ Additional `HealthIndicators` are available but are not enabled by default:
 
 
 
-[[actuator.endpoints.health.writing-custom-health-indicators]]
+[[actuator.endpoints.health.writing-totom-health-indicators]]
 ==== Writing Custom HealthIndicators
 To provide custom health information, you can register Spring beans that implement the {spring-boot-actuator-module-code}/health/HealthIndicator.java[`HealthIndicator`] interface.
 You need to provide an implementation of the `health()` method and return a `Health` response.
@@ -1020,7 +1020,7 @@ You can, for example, configure additional Health Indicators:
 	    health:
 	      group:
 	        readiness:
-	          include: "readinessState,customCheck"
+	          include: readinessState,customCheck
 ----
 
 By default, Spring Boot does not add other health indicators to these groups.


wilkinsona:
Thanks for the proposal but always quoting strings is intentional. Please see #28911 and #28709.

michaldo:
@wilkinsona I understood reasons for quoting single string.
But it is a collection. When I see this example, I am not sure what is expected translation:
- [ "readinessState,customCheck" ]
- [ "readiness", "customCheck" ]

I have impression that someone tries force interpret value as single string
When I see `include: readinessState,customCheck` I'm sure it is  collection
That it works with CSV:

- "readinessState,customCheck"  : 1 column
- readinessState,customCheck      : 2 columns

wilkinsona:
Thanks for the clarification.

The string will be split automatically on each comma during property binding. Also, if we were to change this, it would have to be consistently across every example rather than in just one place.

michaldo:
I understood that due Norway problem you decided to quote all strings. However, there are drawbacks:

1. Yaml briefness is lost due unnecessary quotes
2. Quotes in YAML usually are used to force typing to String. Here target type is a List. Is is confusing for someone used to CSV. 

Is it obvious that
```
list: foo,bar,baz 
list: "foo, bar, baz"
```
maps to the same 3-elements list? 
Is it obvious if you rely only on Spring Boot documentation, without implementation background?

But there is a way to have configuration safe from Norway problem and have briefness YAML. Switch to Yaml 1.2 as requested in #17113 and forget about Norway problem.

BTW, I met Norway problem in my job - I did not know it is well-known. For me it is a real problem, more important than documentation readability, and should be solved by switch to snakeyaml-engine

wilkinsona:
> Yaml briefness is lost due unnecessary quotes

We prefer the consistency of always quoting strings to ensure that they're not accidentally converted into another type. It means the reader does not have to know YAML's conversion rules and understand exactly when string-like values will be converted into another type.

> Quotes in YAML usually are used to force typing to String. Here target type is a List.

From YAML's perspective, the target type is not a list. It's a single scalar value that's subject to YAML's type conversion rules. For it to work correctly and ultimately be converted into a list, you need it to be treated as a comma-separated string. When you write `list: foo,bar,baz` you are relying upon YAML not misinterpreting the value and turning it into something other than a string. This ambiguity is removed by using `list: "foo,bar,baz"`.

PR #30627, opened 2022-04-10
Expose Git and build information as metrics

mafr:
Introduce new metrics `git.info` and `build.info` that expose some information from `GitProperties` and `BuildProperties` respectively, if available. This is the same info that `/actuator/info` provides, but represented as metrics, which allows for convenient integration in Grafana dashboards.

Using tags on a pseudo gauge seems to be the standard way of doing this; see for example Prometheus' `go_info` metric or Micrometer's own `jvm.info` metric (not yet included in Spring Boot).


mafr:
There were some formatting issue that I'm going to fix on a separate PR:

snicoll:
@mafr Going forward, please don't do that. Pushing additional commits on `mafr:project-info-metrics` will update this PR and make sure we don't end up with several PRs for the same thing.

PR #30628, opened 2022-04-10
Expose Git and build information as metrics

mafr:
Introduce new metrics `git.info` and `build.info` that expose some information from `GitProperties` and `BuildProperties` respectively, if available. This is the same info that `/actuator/info` provides, but represented as metrics, which allows for convenient integration in Grafana dashboards.

Using tags on a pseudo gauge seems to be the standard way of doing this; see for example Prometheus' `go_info` metric or Micrometer's own `jvm.info` metric (not yet included in Spring Boot).

Build runs locally, auto formatter has been applied, Checkstyle has no objections.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
@@ -1,0 +1,83 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.autoconfigure.metrics;
+
+import java.util.function.Supplier;
+
+import io.micrometer.core.instrument.Gauge;
+import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.MeterBinder;
+
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.info.ProjectInfoAutoConfiguration;
+import org.springframework.boot.info.BuildProperties;
+import org.springframework.boot.info.GitProperties;
+import org.springframework.context.annotation.Bean;
+
+/**
+ * Auto-configuration that exposes Git and build information as metrics.
+ *
+ * @author Matthias Friedrich
+ * @since 3.0.0
+ */
+@AutoConfiguration(after = { MetricsAutoConfiguration.class, CompositeMeterRegistryAutoConfiguration.class,
+		ProjectInfoAutoConfiguration.class })
+@ConditionalOnClass(MeterRegistry.class)
+@ConditionalOnBean(MeterRegistry.class)
+public class ProjectMetricsAutoConfiguration {
+
+	@Bean
+	@ConditionalOnBean(GitProperties.class)
+	public MeterBinder gitInfoMetrics(GitProperties props) {
+		return (registry) -> {
+			Gauge.Builder<Supplier<Number>> builder = Gauge.builder("git.info", () -> 1L)
+					.description("Project Git information").strongReference(true);
+
+			addTag(builder, "branch", props.getBranch());
+			addTag(builder, "id", props.getShortCommitId());
+			addTag(builder, "time", props.getCommitTime());
+
+			builder.register(registry);
+		};
+	}
+
+	@Bean
+	@ConditionalOnBean(BuildProperties.class)
+	public MeterBinder buildInfoMetrics(BuildProperties props) {
+		return (registry) -> {
+			Gauge.Builder<Supplier<Number>> builder = Gauge.builder("build.info", () -> 1L)
+					.description("Project build information").strongReference(true);
+
+			addTag(builder, "name", props.getName());
+			addTag(builder, "artifact", props.getArtifact());
+			addTag(builder, "group", props.getGroup());
+			addTag(builder, "version", props.getVersion());
+			addTag(builder, "time", props.getTime());
+
+			builder.register(registry);
+		};
+	}
+
+	private static void addTag(Gauge.Builder<Supplier<Number>> builder, String name, Object value) {
+		if (value != null) {
+			builder.tag(name, String.valueOf(value));
+		}
+	}
+
+}
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfigurationTest.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfigurationTest.java
@@ -1,0 +1,118 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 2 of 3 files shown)


wilkinsona:
See #12348 for some history on this topic.

jonatan-ivanov:
I think this makes as much sense (if not more) as https://github.com/micrometer-metrics/micrometer/pull/2221. Also, what everyone thinks (/cc @shakuzen) about adding a `GitInfoMetrics` and a `BuildInfoMetrics` class or maybe just `BuildInfoMetrics` to Micrometer?
The users need to supply the data (e.g.: through ctor) but it would give a simple contract about what needs to be provided and what will be produced. The advantage of doing it is having the same naming convention across Micrometer users also for non spring boot apps.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+	@Bean
+	@ConditionalOnBean(GitProperties.class)
+	public MeterBinder gitInfoMetrics(GitProperties props) {
+		return (registry) -> {
+			Gauge.Builder<Supplier<Number>> builder = Gauge.builder("git.info", () -> 1L)
+					.description("Project Git information").strongReference(true);
+
+			addTag(builder, "branch", props.getBranch());
+			addTag(builder, "id", props.getShortCommitId());
+			addTag(builder, "time", props.getCommitTime());
```
jonatan-ivanov, line 54: How would you use this?
I'm not sure this is very usable with other metrics but it might be useful for setting up an alert to detect apps that were build against a commit before a certain date (e.g.: we made a company wide change that everyone needs to adopt and if your last change is before this date you did not).

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+public class ProjectMetricsAutoConfiguration {
+
+	@Bean
+	@ConditionalOnBean(GitProperties.class)
+	public MeterBinder gitInfoMetrics(GitProperties props) {
+		return (registry) -> {
+			Gauge.Builder<Supplier<Number>> builder = Gauge.builder("git.info", () -> 1L)
+					.description("Project Git information").strongReference(true);
+
+			addTag(builder, "branch", props.getBranch());
```
jonatan-ivanov, line 52: How would you use this?
I'm not sure this is very usable with other metrics but it might be useful for setting up an alert to detect apps that were build (or not) against certain branches (e.g.: everyone needs to build against `main`/`release` all the time).

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+	public MeterBinder buildInfoMetrics(BuildProperties props) {
+		return (registry) -> {
+			Gauge.Builder<Supplier<Number>> builder = Gauge.builder("build.info", () -> 1L)
+					.description("Project build information").strongReference(true);
+
+			addTag(builder, "name", props.getName());
+			addTag(builder, "artifact", props.getArtifact());
+			addTag(builder, "group", props.getGroup());
+			addTag(builder, "version", props.getVersion());
+			addTag(builder, "time", props.getTime());
```
jonatan-ivanov, line 71: I have the same thoughts as above for commit time.

wilkinsona:
> The advantage of doing it is having the same naming convention across Micrometer users also for non spring boot apps

That's quite a compelling advantage. +1 from me, FWIW.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
```diff
@@ -0,0 +1,83 @@
...
+
+			builder.register(registry);
+		};
+	}
+
+	private static void addTag(Gauge.Builder<Supplier<Number>> builder, String name, Object value) {
+		if (value != null) {
```
jonatan-ivanov, line 78: This can be a little dangerous. Let's say one app has the all the git info, the other does not. If that happens you will end-up with the same metric name but different set of tags. This can lead to all sorts of issues and some backends does not or does not like to support this (e.g.: [prometheus](https://github.com/prometheus/client_java/issues/696#issuecomment-920353692)). I would return a default value like `unknown` or `N/A` if the value is missing but I would always want to have the same set of keys.

Here's is an example (ServiceA has all the details, ServiceB only has the commit id)
```
git_info{service="serviceA",branch="main",id="cafecafe",time="2022-04-26T18:39:18"} 1.0
git_info{service="serviceB",id="deadbeef"} 1.0
```
This can be problematic (see above), my proposal instead is:
```
git_info{service="serviceA",branch="main",id="cafecafe",time="2022-04-26T18:39:18"} 1.0
git_info{service="serviceB",branch="unknown",id="deadbeef",time="unknown"} 1.0
```

I have a sad but real-life example for this. If you use heroku in a way that it will build your app when you deploy, heroku removes the .git folder so the git plugin will fail. You can [generate](https://github.com/jonatan-ivanov/resourceater/blob/bd0b061c8d70fd3b8330ed4430491cc3a439f2c4/build.gradle#L69) `git.properties` on your own but heroku will only tell you the commit id so the git info will be [incomplete](https://resourceater.herokuapp.com/actuator/info) (this page might take some time to load).

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfigurationTest.java
```diff
@@ -0,0 +1,118 @@
...
+		this.contextRunner.withBean(GitProperties.class, new Properties()).run(withRegistry((registry) -> {
+			Gauge gauge = registry.find("git.info").gauge();
+
+			assertThat(gauge).isNotNull();
+			assertThat(gauge.getId().getTags()).isEmpty();
+		}));
+	}
+
+	@Test
+	void gitPropertiesPresent() {
```
jonatan-ivanov, line 62: I would add a test for the case when one of the properties are not present.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
jonatan-ivanov: If you do this, you can also simplify things above.

```java
Gauge.builder("git.info", () -> 1L)
        .description("Project Git information").strongReference(true);
        .tag("branch", getOrDefault(props.getBranch()));
		.tag("id", getOrDefault(props.getShortCommitId()));
		.tag("time", getOrDefault(props.getCommitTime()));
		.register(registry);
```

jonatan-ivanov (changes requested):

mafr:
Thanks for your feedback, guys! If there's consensus on this, I'm happy to make a PR against Micrometer to add GitInfoMetrics and BuildInfoMetrics (with the suggested changes) and later adjust this PR to use them (maybe add JvmMetrics while I'm at it).

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
mafr: We found this very handy on Grafana dashboards in development systems. Product owners often asked us whether a specific branch (or yesterday's fix) has already been deployed. I haven't used the commit time it in any type of reporting query yet.

As for why these tags and not others: For consistency reasons I decided to expose the exact information that Actuator's Info endpoint provides.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/ProjectMetricsAutoConfiguration.java
mafr: Sounds good to me!

shakuzen:
If this is worth doing, I agree something in Micrometer standardizing it irrespective of framework would be a worthwhile goal. I'm not sure exactly what it would look like, since there isn't a generic git/build info class like Boot has.

> which allows for convenient integration in Grafana dashboards.

I'd like to better understand the use cases we're trying to enable to make sure we're doing it in the best way for most users. I talked to @jonatan-ivanov today a bit about my concern over how well these kinds of info metrics can be used across different metrics backends. Prometheus consistently seems to be the backend people requesting it are using, and I want to make sure it is useful in other backends as well. We can discuss that more in the Micrometer repo or Slack.

wilkinsona:
Thanks, @shakuzen. I'll place this PR on hold until the discussion in the Micrometer repo or Slack has reached a conclusion. Could you please link to the discussion once it's begun as I'd like to follow along if I can.

PR #30656, opened 2022-04-14
Update Hibernate to 5.6.8.Final

### LINKED ISSUE #30445: The referred version of Hibernate jpamodelgen did not work when upgrading to 3.0.0-M2
The reproduce example: https://github.com/hantsy/spring6-sandbox/blob/master/boot
* Spring Boot 3.0.0-M2
* Java 17
* Maven

I have configured annotation processors: Lombok, Hibernate JpaModelGen, and QueryDSL in this project, when upgrading Spring Boot to 3.0.0-M2, and compiling the project, all JPA metadata models are broke.

When I specify `hibenate-jpamdolegen-jakarta` to an older version, eg. 5.6.5.Final, check here: https://github.com/hantsy/spring6-sandbox/blob/master/boot/pom.xml#L71 it worked again.

Reproduce steps:
1. Remove the version in https://github.com/hantsy/spring6-sandbox/blob/master/boot/pom.xml#L71
2. Compile the project.

hantsy:
Fixes  #30445

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -355,7 +355,7 @@ bom {
 			]
 		}
 	}
-	library("Hibernate", "5.6.7.Final") {
+	library("Hibernate", "5.6.8.Final") {
 		group("org.hibernate") {
 			modules = [
 				"hibernate-c3p0",


snicoll:
Please review the template before submitting a PR, we don't need one for straightforward dependency upgrades like this.

Issue #30654, opened 2022-04-13
bootBuildInfo fails with a NullPointerException when an additional property has a null value


chriswhite199:
If you attempt to add additional properties to buildInfo which could be null at runtime, coersion to string fails with a NPE

### Sample build.gradle entry

```
springBoot {
    buildInfo {
        properties {
            def testProp = project.hasProperty("sdhjksad") ? project.getProperty("sdhjksad").toString() : null
            additional = [
                'nullable': testProp
            ]
        }
    }
}
```

### Stack Trace

```
Caused by: java.lang.NullPointerException
        at org.springframework.boot.gradle.tasks.buildinfo.BuildInfo.lambda$coerceToStringValues$0(BuildInfo.java:111)
        at org.springframework.boot.gradle.tasks.buildinfo.BuildInfo.coerceToStringValues(BuildInfo.java:111)
        at org.springframework.boot.gradle.tasks.buildinfo.BuildInfo.generateBuildProperties(BuildInfo.java:64)
```

### Offending code

* https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/buildinfo/BuildInfo.java#L111
```java
	private Map<String, String> coerceToStringValues(Map<String, Object> input) {
		Map<String, String> output = new HashMap<>();
		input.forEach((key, value) -> output.put(key, value.toString()));   // add null check, and filter when value is null
		return output;
	}
```

pruidong:
```
def testProp = project.hasProperty("sdhjksad") ? project.getProperty("sdhjksad").toString() : ""
```

I think you can replace null with "" to avoid this error.

wilkinsona:
See https://github.com/spring-projects/spring-boot/issues/6724, https://github.com/spring-projects/spring-boot/issues/12382, and https://github.com/spring-projects/spring-boot/pull/16583 for background.

I think we should avoid the NPE, but we should not do that by filtering out the values. As described on the related issues, a property being `null` could indicate a mistake that will then be silently ignored if we filter it out.

For consistency with Maven, I think we should leave the value as `null` so that it fails with a more informative error:

```
* What went wrong:
Execution failed for task ':bootBuildInfo'.
> Additional property 'nullable' is illegal as its value is null
```

PR #30669, opened 2022-04-14
fix: NPE with Gradle plugin's BuildInfo#coerceToStringValues

pruidong:
fix bug: https://github.com/spring-projects/spring-boot/issues/30654


wilkinsona:
Thanks for the PR, @pruidong. Unfortunately, as described in https://github.com/spring-projects/spring-boot/issues/30654#issuecomment-1099443180, I don't think this is the right way to fix the problem. Thanks anyway.

PR #30674, opened 2022-04-15
Fix incorrect link in kafka.adoc

chanhyeong:
I think the document about kafka might be moved from [`features`](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html) to [`messaging`](https://docs.spring.io/spring-boot/docs/current/reference/html/messaging.html#messaging.kafka.streams), but there is still an old link.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/kafka.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/kafka.adoc
@@ -61,7 +61,7 @@ The former can be configured using `spring.kafka.streams.application-id`, defaul
 The latter can be set globally or specifically overridden only for streams.
 
 Several additional properties are available using dedicated properties; other arbitrary Kafka properties can be set using the `spring.kafka.streams.properties` namespace.
-See also <<features#messaging.kafka.additional-properties>> for more information.
+See also <<kafka#messaging.kafka.additional-properties>> for more information.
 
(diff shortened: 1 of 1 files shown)


snicoll:
@chanhyeong tanks for the PR but I believe that the link is correct. The `features` at the beginning is an old artifact from a previous version but handled properly as we rewrite anchors.

Did you notice a version of the Spring Boot docs where the link isn't working? If that's the case, please share a link to it and we can investigate. 

Irrespective of this, I don't think this change is correct.

chanhyeong:
@snicoll 

I think it will be fixed after deploying https://github.com/spring-projects/spring-boot/commit/4c5cb772ffe526eed89dff92ecc75a2423c6a976.

I found this in current version (2.6.6). (https://docs.spring.io/spring-boot/docs/current/reference/html/index.html)
Currently, it works in [single HTML page](https://docs.spring.io/spring-boot/docs/2.6.6/reference/htmlsingle/#messaging.kafka.streams) but doesn't work in [messaging](https://docs.spring.io/spring-boot/docs/current/reference/html/messaging.html#messaging.kafka.streams).
There are difference sentences too.

#### single HTML page
image

#### messaging
image

snicoll:
@chanhyeong Sorry about that. I thought anchor rewrite was handling it, especially after having looked at the current doc. I looked for the single page version though. Going forward, please provide the link to the broken link please.

`kafka` is incorrect as I've mentioned but I'd like to give you the contribution. Thanks for getting back to us!

snicoll:
@chanhyeong thank you for making your first contribution to Spring Boot.

Issue #30543, opened 2022-04-05
Test that each entry in Tomcat's CharsetMapperDefault.properties has a corresponding override


wilkinsona:
As a follow-on to #30535, we could add a test that loads `org/apache/catalina/util/CharsetMapperDefault.properties` and asserts that we have an override in place for each entry.

scottfrederick:
Closing in favor of #30692

PR #30692, opened 2022-04-15
Test that each entry in Tomcat's CharsetMapperDefault.properties has a corresponding override

terminux:
Load `org/apache/catalina/util/CharsetMapperDefault.properties` and assert that each entry has an override.

Closes gh-30543


snicoll:
Thanks again @terminux

PR #30705, opened 2022-04-16
Event handling in JobExecutionExitCodeGenerator is not thread-safe

dugenkui03:
The listener could be invoked concurrently by multiple thread, we should make them threadsafe.

[SimpleApplicationEventMulticaster doc](https://github.com/spring-projects/spring-framework/blob/f40a391916ed6c1f9e1130638a0bf19479e514dd/spring-context/src/main/java/org/springframework/context/event/SimpleApplicationEventMulticaster.java#L40): 
>By default, all listeners are invoked in the calling thread.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobExecutionExitCodeGenerator.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobExecutionExitCodeGenerator.java
@@ -16,8 +16,8 @@
 
 package org.springframework.boot.autoconfigure.batch;
 
-import java.util.ArrayList;
 import java.util.List;
+import java.util.concurrent.CopyOnWriteArrayList;
 
 import org.springframework.batch.core.JobExecution;
 import org.springframework.boot.ExitCodeGenerator;
@@ -31,7 +31,7 @@ import org.springframework.context.ApplicationListener;
  */
 public class JobExecutionExitCodeGenerator implements ApplicationListener<JobExecutionEvent>, ExitCodeGenerator {
 
-	private final List<JobExecution> executions = new ArrayList<>();
+	private final List<JobExecution> executions = new CopyOnWriteArrayList<>();
 
 	@Override
 	public void onApplicationEvent(JobExecutionEvent event) {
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaAutoConfigurationTests.java
```diff
@@ -600,7 +600,7 @@ EventCapturingApplicationListener jpaUsingApplicationListener(EntityManagerFacto
 
 		static class EventCapturingApplicationListener implements ApplicationListener<ApplicationEvent> {
 
-			private final List<ApplicationEvent> events = new ArrayList<>();
```
wilkinsona, line 603: I don't think this change is necessary. Within the context of the two tests that use the listener, events are only ever delivered on the main thread.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobExecutionExitCodeGenerator.java
```diff
@@ -31,7 +31,7 @@
  */
 public class JobExecutionExitCodeGenerator implements ApplicationListener<JobExecutionEvent>, ExitCodeGenerator {
 
-	private final List<JobExecution> executions = new ArrayList<>();
```
wilkinsona, line 34: 👍 This could also break with a single-threaded event multi-caster if `getExitCode()` is called one one thread when another thread is delivering an event.

wilkinsona (changes requested):
Thanks very much for the PR, @dugenkui03. While I think the change to `ClassPathFileSystemWatcherTests` is worth making, I'd like to focus on `JobExecutionExitCodeGenerator` as that's a bug in main code.

dugenkui03:
@wilkinsona Thanks for your review. Code in `HibernateJpaAutoConfigurationTests` have been reverted , please help review again.

wilkinsona:
Thanks very much, @dugenkui03. I separated out into a separate issue (#30844) the changes to ClassPathFileSystemWatcherTests as, unlike the problem with JobExecutionExitCodeGenereator, the problem with the tests won't affect users.

PR #30706, opened 2022-04-16
Modified to method references

pruidong:
Modified to method references and simplified passing of null values.


snicoll:
Thanks for the PR but we prefer to handle stylistic changes like this ourselves. While passing `result` as the first call is going to be `null` indeed, I believe this is more consistent to pass it for all calls like we do now.

Issue #30631, opened 2022-04-11
SearchStrategy argument of MethodValidationExcludeFilter byAnnotation(Class, SearchStrategy) is not used


wilkinsona:
From @chanhyeong on Gitter:

> I have a question about the static method [byAnnotation(Class<? extends Annotation> annotationType, SearchStrategy searchStrategy)](https://github.com/spring-projects/spring-boot/blob/52d7424de8ed362381b5fb477a4cf2ba4d69711f/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/validation/beanvalidation/MethodValidationExcludeFilter.java#L58-L59) in MethodValidationExcludeFilter.
It has a searchStrategy argument that doesn't be used (has a fixed value SearchStrategy.SUPERCLASS)

This looks like a bug to me. The search strategy argument should be passed into the call to `MergedAnnotations.from`.

pruidong:
@wilkinsona Do you mean should remove SearchStrategy searchStrategy parameter, or change MergedAnnotations.from(type, SearchStrategy.SUPERCLASS) to MergedAnnotations.from(type, searchStrategy)?

wilkinsona:
Thanks for the offer. I have a fix and a test ready to push once I’m back in the office.

pruidong:
> Thanks for the offer. I have a fix and a test ready to push once I’m back in the office.

Ok, hope to have the opportunity to provide PR to Spring Boot in the future.

PR #30707, opened 2022-04-16
fix: SearchStrategy argument of MethodValidationExcludeFilter byAnnotation(Class, SearchStrategy) is not used

pruidong:
SearchStrategy argument of MethodValidationExcludeFilter byAnnotation(Class, SearchStrategy) is not used,fix: https://github.com/spring-projects/spring-boot/issues/30631


snicoll:
@pruidong thanks for the PR but that is issue is already assigned to someone and Andy[mentioned he had done the work already](https://github.com/spring-projects/spring-boot/issues/30631#issuecomment-1100074817).

Issue #30642, opened 2022-04-12
Change identityprovider to assertingparty in Saml2RelyingPartyProperties


marcusdacoregio:
Since [this commit](https://github.com/spring-projects/spring-security/commit/8e8a642e5a8a25138361fc5d0127c27b453da35d) in Spring Security, some properties in the `RelyingPartyRegistration` were deprecated in favor of others aligned SAML 2.0 Spec Language.

Consider changing `identityprovider` in favor of `assertingparty` in `Saml2RelyingPartyProperties` to better align with the naming in Spring Security.

Related:
 - https://github.com/spring-projects/spring-security/issues/8777

mhalbritter:
I've added a note to the release notes: https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-2.7.0-RC1-Release-Notes#deprecations-in-spring-boot-27

Issue #9945, opened 2017-08-03
Provide configuration property metadata for binding to Map<String, RichType> and List<RichType>


wilkinsona:
One example of such a `Map` is [in Spring Cloud stream](https://github.com/spring-cloud/spring-cloud-stream/blob/bd002e4aaf05f2f4e31cb09d706910b470f960a3/spring-cloud-stream/src/main/java/org/springframework/cloud/stream/config/BindingServiceProperties.java#L60-L61).  As things stand, there's no metadata for any of the properties on `BindingProperties`. I think it would be interesting to explore the possibility of providing metadata for those properties, perhaps using a wildcard for the key in the map:

```
spring.cloud.stream.bindings.*.content-type
spring.cloud.stream.bindings.*.destination
spring.cloud.stream.bindings.*.group
…
```

For this to be useful, we'd need a corresponding enhancement in each of the IDE plugins.

snicoll:
I've thought long and hard about this and I don't think we should do this. There are several reasons:

1. Scalar vs. Group: the only way for us to harvest nested object is by using `@NestedConfigurationProperty` (essentially, this is not a scalar value, please create a "sub-group" for all the properties defined by that type). If we support that for collections or maps now, how do we tell if the value is scalar or not?
2. Scope: if we find a way to fix this and we offer automatic harvesting of POJOs, where do we stop? The next step is to expect this to be somehow recursive (`BindingProperties` may hold another nested object).
3. Duplication: the use of the `*` above is what I ended up considering myself but that pojo can be used in multiple places, in multiples maps and/or in different objects. It means that the metadata is duplicated. If you don't want that, we need a third root objects (to harvest the pojo) and a way to reference to it somehow).

Let's see what the rest of the team thinks.

philwebb:
For "1)" the `@NestedConfigurationProperty` could be the indicator that the value isn't a scalar. So  `@NestedConfigurationProperty Map<String, Foo>` is a hint that `Foo` can be expanded with the `*` syntax, where as `@NestedConfigurationProperty Map<String, Foo>` cannot (and we expect to be able to convert a `String` to a `Foo` somehow.

For "2)" and "3)" I'd probably go with duplicate `*`. I don't see the duplication being all that much of a problem, as long as there isn't an infinite loop (e.g. `Foo` contains a `Map<String,Foo>`).

philwebb:
FWIW, I think we shouldn't invest too much time thinking about this until after 2.0.

[...]

PR #30742, opened 2022-04-19
Change identityprovider to asserting-party in Saml2RelyingPartyProperties

mhalbritter:
This was a bit of a hassle, because the properties-migrator can't be used for map-based types, which the `identityprovider` is (see https://github.com/spring-projects/spring-boot/issues/9945). I had to write codepaths for both the old and the new property names, and a warning is logged if the old one is used. I adjusted the tests to test with both the old and the new names.

I had to change `private boolean signRequest = true;` to `private Boolean signRequest;` to get the fallback working. This doesn't make a difference, because this only gets applied if `!= null`. And the default in Spring Security is true for that field (see `org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistration.AssertingPartyDetails.Builder#wantAuthnRequestsSigned`).

Closes gh-30642

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2021 the original author or authors.
+ * Copyright 2012-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -67,7 +67,14 @@ public class Saml2RelyingPartyProperties {
 		/**
 		 * Remote SAML Identity Provider.
 		 */
-		private final Identityprovider identityprovider = new Identityprovider();
+		private final AssertingParty assertingParty = new AssertingParty();
+
+		/**
+		 * Remote SAML Identity Provider.
+		 * @deprecated use {@link #assertingParty}
+		 */
+		@Deprecated
+		private final AssertingParty identityprovider = new AssertingParty();
 
 		public String getEntityId() {
 			return this.entityId;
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
```diff
@@ -300,11 +317,11 @@ public void setBinding(Saml2MessageBinding binding) {
 				this.binding = binding;
 			}
 
-			public boolean isSignRequest() {
+			public Boolean isSignRequest() {
```
snicoll, line 320: I thinktThis should be deprecated with the original signature and a getter style should be added (considering the wrapper type).

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyAutoConfigurationTests.java
```diff
@@ -61,18 +63,20 @@ class Saml2RelyingPartyAutoConfigurationTests {
...
+	void autoConfigurationShouldBeConditionalOnRelyingPartyRegistrationRepositoryClass(boolean useDeprecated) {
+		this.contextRunner.withPropertyValues(getPropertyValues(useDeprecated)).withClassLoader(new FilteredClassLoader(
 				"org.springframework.security.saml2.provider.service.registration.RelyingPartyRegistrationRepository"))
 				.run((context) -> assertThat(context).doesNotHaveBean(RelyingPartyRegistrationRepository.class));
 	}
 
-	@Test
-	void autoConfigurationShouldBeConditionalOnServletWebApplication() {
+	@ParameterizedTest
+	@ValueSource(booleans = { true, false })
```
snicoll, line 75: A bit odd. I'd rather have two separate tests (and the one using the deprecated form should be marked as deprecated so that we can clean things up once we remove the deprecated feature that it tests.

snicoll (commented):
I had a quick look and added a couple comments for your consideration.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/saml2/Saml2RelyingPartyProperties.java
snicoll: Actually maybe that without deprecating it since we're going to roll back to that once the support of the old format is removed.

Issue #30273, opened 2022-03-17
Rework spring.mustache properties to make it clear which are Servlet-specific


wilkinsona:
Split out from https://github.com/spring-projects/spring-boot/issues/28858.

The following properties are Servlet-specific but that's not apparent from their prefix:

spring.mustache.allow-request-override
spring.mustache.allow-session-override
spring.mustache.cache
spring.mustache.content-type 
spring.mustache.expose-request-attributes
spring.mustache.expose-session-attributes
spring.mustache.expose-spring-macro-helpers

Each should be deprecated and a replacement with a `spring.mustache.servlet.`prefix introduced. This will mean that `MustacheProperties` can no longer extend `AbstractTemplateViewResolverProperties`. Once these changes are in place, we can tackle #28858 by introducing a reactive-specific property that configures the media types supported by a `org.springframework.boot.web.reactive.result.view.MustacheViewResolver`.

PR #30774, opened 2022-04-22
DevTools sets deprecated spring.mustache.cache property

mmoayyed:
Spring Boot 2.7 RC1, on startup, issues the following warning:

```
The use of configuration keys that have been renamed was found in the environment:

Property source 'devtools':
        Key: spring.mustache.cache
                Replacement: spring.mustache.servlet.cache


Each configuration key has been temporarily mapped to its replacement for your convenience. To silence this 
warning, please update your configuration to use the new keys.
```

As the message suggests, this seems to be issued by the "DevTools" property source, and `spring.mustache.cache` should be replaced with `spring.mustache.servlet.cache`, to avoid configuration warnings on startup.


bclozel:
See #30273

wilkinsona:
Thanks very much, @mmoayyed.

PR #30776, opened 2022-04-22
Provide a configuration property for Kafka's async acks

TheCK:


wilkinsona:
@TheCK Thanks very much for making your first contribution to Spring Boot.

PR #30778, opened 2022-04-23
Fix property in CustomHibernateJpaAutoConfigurationTests.defaultDatabaseIsSet()

izeye:
This PR fixes a property in `CustomHibernateJpaAutoConfigurationTests.defaultDatabaseIsSet()`.


wilkinsona:
Thanks very much, @izeye. The property could be remove altogether so I've done that in 
7a98627.

Issue #30661, opened 2022-04-14
Add WebClient based sender for Zipkin


mhalbritter:
We already have a `RestTemplate` based sender for Zipkin, we should add a `WebClient` based one, too. We can [take the code from sleuth](https://github.com/spring-cloud/spring-cloud-sleuth/blob/22be601b89126bc2a4c212675522a3d021f4b32f/spring-cloud-sleuth-zipkin/src/main/java/org/springframework/cloud/sleuth/zipkin2/WebClientSender.java#L33).

StefanBratanov:
Hi, I am keen to try this issue as a first commit. 🙂 Is it possible to assign?

mhalbritter:
Hey, we have [issues labeled as "first-timers only"](https://github.com/spring-projects/spring-boot/issues?q=is%3Aopen+is%3Aissue+label%3A%22status%3A+first-timers-only%22) for exactly this purpose. The list is currently empty, but keep an eye out for issues with this label.

StefanBratanov:
Hi, thank you. I will keep an eye out. In the meantime, I was interested and did changes on a branch in my fork yesterday. I understand it could be throw away work, but it was fun looking at the code. Let me know if it looks good and it can be used.

https://github.com/StefanBratanov/spring-boot/commits/web-client-sender-zipkin

StefanBratanov:
Great, just created a PR :) https://github.com/spring-projects/spring-boot/pull/30792

mhalbritter:
Thanks, i will close this issue then.

PR #30792, opened 2022-04-25
Add WebClient based sender for Zipkin

StefanBratanov:
## PR Description

- Introduce abstract `HttpSender` class
- Introduce `ZipkinWebClientSender`
- Add `ZipkinWebClientSender` bean to configuration

related to #30661

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/HttpSender.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/HttpSender.java
@@ -1,0 +1,163 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.autoconfigure.tracing.zipkin;
+
+import java.io.ByteArrayOutputStream;
+import java.io.IOException;
+import java.util.List;
+import java.util.zip.GZIPOutputStream;
+
+import zipkin2.Call;
+import zipkin2.Callback;
+import zipkin2.CheckResult;
+import zipkin2.codec.Encoding;
+import zipkin2.reporter.BytesMessageEncoder;
+import zipkin2.reporter.ClosedSenderException;
+import zipkin2.reporter.Sender;
+
+import org.springframework.http.HttpHeaders;
+import org.springframework.util.unit.DataSize;
+
+/**
+ * A Zipkin {@link Sender} that uses an HTTP client to send JSON spans. Supports automatic
+ * compression with gzip.
+ *
+ * @author Moritz Halbritter
+ * @author Stefan Bratanov
+ */
+abstract class HttpSender extends Sender {
+
+	private static final DataSize MESSAGE_MAX_BYTES = DataSize.ofKilobytes(512);
+
+	volatile boolean closed;
+
+	@Override
+	public Encoding encoding() {
+		return Encoding.JSON;
+	}
+
+	@Override
+	public int messageMaxBytes() {
+		return (int) MESSAGE_MAX_BYTES.toBytes();
+	}
+
+	@Override
+	public int messageSizeInBytes(List<byte[]> encodedSpans) {
+		return encoding().listSizeInBytes(encodedSpans);
+	}
+
+	@Override
+	public int messageSizeInBytes(int encodedSizeInBytes) {
+		return encoding().listSizeInBytes(encodedSizeInBytes);
+	}
+
+	@Override
+	public CheckResult check() {
+		try {
+			sendSpans(List.of()).execute();
+			return CheckResult.OK;
+		}
+		catch (IOException | RuntimeException ex) {
+			return CheckResult.failed(ex);
+		}
+	}
+
+	@Override
+	public void close() throws IOException {
+		this.closed = true;
+	}
+
+	/**
+	 * The returned {@link HttpPostCall} will send span(s) as a POST to a zipkin endpoint
+	 * when executed.
+	 * @param batchedEncodedSpans list of encoded spans as a byte array
+	 * @return an instance of a Zipkin {@link Call} which can be executed
+	 */
+	protected abstract HttpPostCall sendSpans(byte[] batchedEncodedSpans);
+
+	@Override
+	public Call<Void> sendSpans(List<byte[]> encodedSpans) {
+		if (this.closed) {
+			throw new ClosedSenderException();
+		}
+		return sendSpans(BytesMessageEncoder.JSON.encode(encodedSpans));
+	}
+
+	abstract static class HttpPostCall extends Call.Base<Void> {
+
+		byte[] body;
+
+		protected HttpPostCall(byte[] body) {
+			this.body = body;
+		}
+
+		/**
+		 * Only use gzip compression on data which is bigger than this in bytes.
+		 */
+		private static final DataSize COMPRESSION_THRESHOLD = DataSize.ofKilobytes(1);
+
+		/**
+		 * Implementations should send a POST request.
+		 * @param body the body to send as bytes
+		 * @param defaultHeaders the headers for the POST request. They can be further
+		 * modified if necessary.
+		 */
+		abstract void post(byte[] body, HttpHeaders defaultHeaders);
+
+		@Override
+		protected Void doExecute() throws IOException {
+			HttpHeaders headers = new HttpHeaders();
+			headers.set("b3", "0");
+			headers.set("Content-Type", "application/json");
+			byte[] body;
+			if (needsCompression()) {
+				headers.set("Content-Encoding", "gzip");
+				body = compress(this.body);
+			}
+			else {
+				body = this.body;
+			}
+			post(body, headers);
+			return null;
+		}
+
+		@Override
+		protected void doEnqueue(Callback<Void> callback) {
+			try {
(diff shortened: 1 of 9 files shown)


wilkinsona:
Nothing at the moment, @StefanBratanov. Moritz is out for a few weeks on paternity leave so it may take us a little while to get to it. 3.0 doesn't GA until November so there's no rush at this stage.

StefanBratanov:
> Nothing at the moment, @StefanBratanov. Moritz is out for a few weeks on paternity leave so it may take us a little while to get to it. 3.0 doesn't GA until November so there's no rush at this stage.

Sure. No worries. Thank you for quick update. Congratulations to Moritz. 🙂

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -56,6 +62,20 @@ Sender urlConnectionSender(ZipkinProperties properties) {
...
+		@ConditionalOnMissingBean
+		@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.REACTIVE)
+		@ConditionalOnBean(WebClient.Builder.class)
+		@ConditionalOnMissingClass("zipkin2.reporter.urlconnection.URLConnectionSender")
+		Sender webClientSender(ZipkinProperties properties, WebClient.Builder webClientBuilder) {
+			HttpClient client = HttpClient.create()
+					.option(ChannelOption.CONNECT_TIMEOUT_MILLIS, (int) properties.getConnectTimeout().toMillis())
+					.doOnConnected((conn) -> conn
+							.addHandlerLast(new ReadTimeoutHandler((int) properties.getReadTimeout().toSeconds())));
+			WebClient webClient = webClientBuilder.clientConnector(new ReactorClientHttpConnector(client)).build();
```
mhalbritter, line 75: This introduces a hard dependency on Netty, "just" for the sake of setting a timeout.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/HttpSender.java
```diff
@@ -0,0 +1,163 @@
...
+	public Call<Void> sendSpans(List<byte[]> encodedSpans) {
+		if (this.closed) {
+			throw new ClosedSenderException();
+		}
+		return sendSpans(BytesMessageEncoder.JSON.encode(encodedSpans));
+	}
+
+	abstract static class HttpPostCall extends Call.Base<Void> {
+
+		byte[] body;
```
mhalbritter, line 102: Can this be private final?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/HttpSender.java
```diff
@@ -0,0 +1,163 @@
...
+ * compression with gzip.
+ *
+ * @author Moritz Halbritter
+ * @author Stefan Bratanov
+ */
+abstract class HttpSender extends Sender {
+
+	private static final DataSize MESSAGE_MAX_BYTES = DataSize.ofKilobytes(512);
+
+	volatile boolean closed;
```
mhalbritter, line 46: Can this be private?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinWebClientSender.java
```diff
@@ -0,0 +1,72 @@
...
+			super(body);
+			this.endpoint = endpoint;
+			this.webClient = webClient;
+		}
+
+		@Override
+		void post(byte[] body, HttpHeaders defaultHeaders) {
+			this.webClient.post().uri(this.endpoint).headers((headers) -> headers.addAll(defaultHeaders))
+					.bodyValue(body).retrieve().toBodilessEntity().subscribe((__) -> {
+					}, (__) -> {
```
mhalbritter, line 61: This looks wrong, the errors will be suppressed. The webclient and the Zipkin Sender interface both support callbacks, maybe we could refactor this to fix the error handling?

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinWebClientSenderTests.java
```diff
@@ -0,0 +1,151 @@
...
+
+	public static MockWebServer mockBackEnd;
+
+	private static String ZIPKIN_URL;
+
+	private ZipkinWebClientSender sut;
+
+	@BeforeAll
+	static void beforeAll() throws IOException {
+		mockBackEnd = new MockWebServer();
```
mhalbritter, line 56: Is this integration testing the only way to test webclients? RestTemplate has a built-in test infrastructure, I wonder if WebClient has the same?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -56,6 +62,20 @@ Sender urlConnectionSender(ZipkinProperties properties) {
 					.build();
 		}
 
+		@Bean
```
mhalbritter, line 65: I have fixed a bug in `main` regarding the auto-configuration, I guess you'll have to rebase.

mhalbritter (changes requested):

mhalbritter:
Hey, thanks for the contribution. I'm back and have some comments on the code.

StefanBratanov:
> Hey, thanks for the contribution. I'm back and have some comments on the code.

Welcome back! Thank you. I will have a look.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinRestTemplateSender.java
```diff
@@ -16,153 +16,71 @@
...
-		}
-
-		private byte[] compress(byte[] input) throws IOException {
-			ByteArrayOutputStream result = new ByteArrayOutputStream();
-			try (GZIPOutputStream gzip = new GZIPOutputStream(result)) {
-				gzip.write(input);
-			}
-			return result.toByteArray();
+		protected void doEnqueue(Callback<Void> callback) {
+			CompletableFuture.supplyAsync(this::doExecute).whenComplete((__, throwable) -> {
```
mhalbritter, line 76: We don't need the `CompletableFuture` in this case. The UrlConnectionSender implements the `enqueue` call without additional threading, so we should be fine here without introducing multi-threading.

mhalbritter:
Thanks @StefanBratanov for the changes, the PR looks quite nice now!

> Hi Moritz, sorry for the delay. I did make a refactor based on the comments, which i agree with. I wasn't sure about handling the `WebClient` async call, but now i think it makes sense and is cleaner. I also changed the `ZipkinRestTemplateSender` async implementation to use a `CompletableFuture` since it was blocking before. I also added tests for both sync and async in both implementations.

I don't think we need that. The `UrlConnectionSender` from zipkin implements the enqueue call without additional threading, so we should be fine here without introducing multi-threading.

> I am not sure about initializing the `WebClient` in `ZipkinConfigurations`. I was using the `ReactorClientHttpConnector` which is based on` Netty`, so that i can honor the read and connect timeouts set in `ZipkinProperties`. What do you think is a better approach for initializing the `WebClient`? 

I have to double check if there's really no way to set timeouts without tying us directly on the implementation.

> Also i couldn't find a mock server similar to `MockRestServiceServer` for `WebClient`, so kept my old implementation. If there is one, I am happy to change it.

Yes, I was wrong, there's no such support in the Spring Framework. The recommendation is to use the [OkHttp MockServer](https://docs.spring.io/spring-framework/docs/6.0.0-SNAPSHOT/reference/html/web-reactive.html#webflux-client-testing), like you did.

StefanBratanov:
Hi @mhalbritter thanks for review, I replaced the` CompletableFuture` implementation with a normal one. I will wait for you to check about the WebClient initialization. My logic was that `@ConditionalOnWebApplication(type = ConditionalOnWebApplication.Type.REACTIVE)` in _ZipkinConfigurations_ if true, should mean that Netty should be on the classpath. I could be wrong though.

mhalbritter:
Unfortunately, this is not the case. You can [use other servers for reactive applications](https://docs.spring.io/spring-boot/docs/3.0.0-SNAPSHOT/reference/htmlsingle/#web.reactive.reactive-server).

> Spring Boot includes support for the following embedded reactive web servers: Reactor Netty, Tomcat, Jetty, and Undertow

The best option would be to have timeout setter on the `WebClient.Builder`, but unfortunately this is not available [and won't be](https://github.com/spring-projects/spring-framework/issues/25115#issuecomment-677685984).

I would drop the timeout code in the configuration for now, and I will bring this up with the team if we can find a solution which works for all the connector implementations.

StefanBratanov:
> Unfortunately, this is not the case. You can [use other servers for reactive applications](https://docs.spring.io/spring-boot/docs/3.0.0-SNAPSHOT/reference/htmlsingle/#web.reactive.reactive-server).
> 
> > Spring Boot includes support for the following embedded reactive web servers: Reactor Netty, Tomcat, Jetty, and Undertow
> 
> The best option would be to have timeout setter on the `WebClient.Builder`, but unfortunately this is not available [and won't be](https://github.com/spring-projects/spring-framework/issues/25115#issuecomment-677685984).
> 
> I would drop the timeout code in the configuration for now, and I will bring this up with the team if we can find a solution which works for all the connector implementations.

Got it. Makes sense. Good to know! The default builder will take care of initialising the appropriate connector. I removed the Netty connector.

mhalbritter:
I've merged your PR. Thanks for your work and congratulations on your first contribution to Spring Boot!

PR #30839, opened 2022-05-02
Upgrade to Maven Dependency Plugin 3.3.0 for 2.6.x

ryenus:
Currently this has only been addressed for 2.7.x, which is not released yet.
It's better to also fix this in 2.6.x due to the severity of the issue.

See https://github.com/spring-projects/spring-boot/issues/30301
and https://issues.apache.org/jira/browse/MDEP-761

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1150,7 +1150,7 @@ bom {
 			]
 		}
 	}
-	library("Maven Dependency Plugin", "3.2.0") {
+	library("Maven Dependency Plugin", "3.3.0") {
 		group("org.apache.maven.plugins") {
 			plugins = [
 				"maven-dependency-plugin"


ryenus:
Hi, @scottfrederick I really want to agree with you, with the hope that the critical performance improvement achieved in https://issues.apache.org/jira/browse/MDEP-761 is fixed in a patch version of the maven-dependency-plugin, so we don't have to update to the next minor version.

Except that's not the case with maven-dependency-plugin, if you take a look at https://repo.maven.apache.org/maven2/org/apache/maven/plugins/maven-dependency-plugin/ you can see that only version with the much needed fix is `3.3.0`.

Also looking at the changes in `3.3.0` there's no backward incompatible changes:
https://blogs.apache.org/maven/entry/apache-maven-dependency-plugin-version3

I totally understand the importance of the [patch-only support policy](https://github.com/spring-projects/spring-boot/wiki/Supported-Versions#third-party-dependencies). But, if the issue is only fixed in the unreleased 2.7.x version, are you going to ask all the spring users to switch to an unreleased spring-boot version? Or manually override the maven-dependency-plugin version, and make sure to remove that at a later time once the next spring boot version is available?

As a result, the spring boot community users will end up with either upgrading maven-dependency-plugin, or upgrading spring-boot, just for this critical performance issue. Thinking about the cost and the benefit of the community, I still think having this change made in spring-boot itself is the most economical one.

With that said, could you please kindly reconsider your decision? We as the community would really thank you for that!

Issue #30832, opened 2022-04-28
Add initial-keys-to-sanitize


mrgrew:
**Affects:** \2.5.x+

---

### Motivation
The `org.springframework.boot.actuate.autoconfigure.context.properties.ConfigurationPropertiesReportEndpointProperties` and `org.springframework.boot.actuate.autoconfigure.env.EnvironmentEndpointProperties` classes provide `additionalKeysToSanitize`, which are sanitized after the defaults. There are cases where sanitizing keys BEFORE the defaults would be preferable.

### Example
The `database_uri` property below is a connection string that includes credentials. Because "database_uri" matches one of the `org.springframework.boot.actuate.endpoint.Sanitizer.URI_USERINFO_KEYS` but the value doesn't match `org.springframework.boot.actuate.endpoint.Sanitizer.URI_USERINFO_PATTERN`, the credentials in that property are exposed by the actuator. Attempting to have the whole property value sanitized using `additionalKeysToSanitize` doesn't work because the Sanitizer has already processed that property as part of the `org.springframework.boot.actuate.endpoint.Sanitizer.DEFAULT_KEYS_TO_SANITIZE`.
```properties
database_uri=jdbc:mariadb://dbhost/dbname?user=dbuser&password=dbpassword
```

A workaround is to redefine all keys to sanitize in application.properties, which is error prone and brittle to future changes. For example:
```properties
management.endpoint.configprops.keys-to-sanitize=database_uri,password,secret,key,token,.*credentials.*,vcap_services,^vcap\\.services.*$,sun.java.command,^spring[\\._]application[\\\\._]json$,uri,uris,url,urls,address,addresses
```

### Proposed Solution
Add an optional `initial-keys-to-sanitize` property to `org.springframework.boot.actuate.autoconfigure.context.properties.ConfigurationPropertiesReportEndpointProperties` and `org.springframework.boot.actuate.autoconfigure.env.EnvironmentEndpointProperties` which is processed _before_ the DEFAULT_KEYS_TO_SANITIZE. Then developers can still leverage the defaults and provide overrides if needed. With such a property, the very complex keys-to-sanitize above goes away and is replaced with
```properties
management.endpoint.configprops.initial-keys-to-sanitize=database_uri
```

wilkinsona:
Thanks for the suggestion.

[...]

PR #30843, opened 2022-05-03
Prepend additional keys to sanitize

mrgrew:
Per discussion in #30832, this change prepends additional keys to sanitize ahead of the defaults. This allows users to sanitize keys that would otherwise be handled by the defaults but still expose credentials.


mrgrew:
I'm looking forward to some feedback on this PR - it's my first submission so I'm not sure how patient to be.

wilkinsona:
Thanks for your patience thus far, @mrgrew. As indicated by the labels on #30832, we'd like to look at this topic as a team so that we can decide what to do. We have a number of competing priorities at the moment but we will get to it as soon as we can.

mbhave:
Closing in favor of  #32156. We've decide to move away from `keys-to-sanitize` because predicting the most secure default is tricky. We've instead decided to go for the most secure default which is to sanitize everything and allow that to be configured based on roles.

PR #30879, opened 2022-05-06
Extend documentation on Datadog metrics

denisw:
- Document that an application key must be set to publish metadata
  for the exported metrics.

- Point out that using a non-US Datadog site (e.g., EU) requires
  changing the `uri` property.


wilkinsona:
Thanks very much, @denisw.

Issue #30885, opened 2022-05-06
MimeMappings does not include application/wasm


ememisya:
https://github.com/spring-projects/spring-boot/blob/cc1807608e6569c1ceb49e558c99f622d6d788d4/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/MimeMappings.java#L206

Please prepend:

```
mappings.add("wasm", "application/wasm");
```

To the indicated line.  Currently I'm handling recognition of the wasm payload as such:

```
@Component
public class ServletCustomizer implements EmbeddedServletContainerCustomizer
{

  @Override
  public void customize(final ConfigurableEmbeddedServletContainer container)
  {
    final MimeMappings mappings = new MimeMappings(MimeMappings.DEFAULT);
    mappings.add("wasm", "application/wasm");
    container.setMimeMappings(mappings);
  }
}
```

Thank you for your time!

See (https://webassembly.org/docs/web/) for further details on the type.

> This compilation can be performed in the background and in a streaming manner. If the Response is not CORS-same-origin, does not represent an ok status, or does not match the `application/wasm` MIME type, the returned promise will be rejected with a TypeError; 

This was previously held up by IANA status which has now been resolved.

References:

https://github.com/WebAssembly/spec/issues/573#issuecomment-824715263

https://www.iana.org/assignments/media-types/application/wasm

https://github.com/spring-projects/spring-boot/issues/17254

bclozel:
Looking at https://github.com/spring-projects/spring-boot/issues/17254#issuecomment-503619255, it sounds like we were not waiting for IANA official registration but rather its adoption by Servlet containers. It looks like this MIME type got added in Tomcat with https://github.com/apache/tomcat/commit/54898ee4c0974fc3280abcb753e54d76d2149d2c .

Maybe we should consider it then?

wilkinsona:
Tomcat added it in 9.0.37 so I think we should add it. [Like Phil](https://github.com/spring-projects/spring-boot/issues/17254#issuecomment-503619255), I thought we had some tests that checked our mappings were in sync with Tomcat's but we haven't had a test failure so I'm not sure what's going on there.

wilkinsona:
I think [this](https://github.com/spring-projects/spring-boot/issues/4161#issuecomment-149202347) may explain it:

>  Both Jetty and Undertow have default mime mappings that Boot's mime mappings are then added to, whereas Tomcat just uses Boot's mime mappings.

[...]

PR #30897, opened 2022-05-08
Add missing Tomcat's default MIME mappings

StefanBratanov:
## PR Description

- Added missing Tomcat's default MIME mappings to `MimeMappings`. Used [MimeTypeMappings.properties](https://github.com/apache/tomcat/blob/main/java/org/apache/catalina/startup/MimeTypeMappings.properties) from Tomcat.
- Modified test to call `addDefaultMimeTypeMappings` from `Tomcat` class to ensure new Tomcat mappings are not missed in the future.

related to #30885


wilkinsona:
Closing per https://github.com/spring-projects/spring-boot/issues/30885#issuecomment-1120423709.

PR #30898, opened 2022-05-08
Configuration property for Statsd's buffered and step properties are missing

izeye:
This PR changes to expose `buffered` property for `StatsdConfig`.

See https://github.com/micrometer-metrics/micrometer/issues/1375

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdProperties.java
@@ -76,6 +76,11 @@ public class StatsdProperties {
 	 */
 	private boolean publishUnchangedMeters = true;
 
+	/**
+	 * Whether measurements should be buffered before sending to the StatsD server.
+	 */
+	private boolean buffered = true;
+
 	public boolean isEnabled() {
 		return this.enabled;
 	}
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks, @izeye. It's a shame that we missed this. I've opened https://github.com/spring-projects/spring-boot/issues/30901 to see if we can stop if from happening again.

There is a work around for the missing property, but it's not very elegant:

```
@Bean
StatsdConfig statsdConfig(StatsdProperties statsdProperties) {
    return new StatsdPropertiesConfigAdapter(statsdProperties) {

        @Override
        public boolean buffered() {
            return false;
        }

    };
}
```

As such, I am a bit tempted to treat this as a bug and fix it in 2.5.x. Flagging for team attention to see what everyone else thinks.

scottfrederick:
I think it's reasonable to consider this a bug of omission.

wilkinsona:
I've started looking at implementing something for #30901 and it looks like we're missing a configuration property for `step` as well. WDYT, @izeye? Does it make sense to have a configuration property for it?

izeye:
@wilkinsona I don't have much experience working with StatsD, but it seems to have been missed accidentally, too and it sounds reasonable to align configurability level with its corresponding meter registry configuration.

/cc @shakuzen

shakuzen:
Statsd's `step` makes sense to expose as a configuration property and I suspect we just missed it.

wilkinsona:
Thanks, @shakuzen. @izeye, do you have time to update this PR to add `step` as well?

izeye:
@wilkinsona Sure, I pushed ccbe758 for it.

snicoll:
Thanks again @izeye

PR #30825, opened 2022-04-28
Add `OtlpMeterRegistry` auto-configuration

eddumelendez:
micrometer 1.9 ships a new module for otlp and this commit add the
auto-configuration.

See gh-30798

PR #30984, opened 2022-05-12
Add OtlpProperties resourceAttributes

eddumelendez:


wilkinsona:
I've labelled this as a task as it's a follow-on to https://github.com/spring-projects/spring-boot/pull/30825 which has not yet been released.

PR #31027, opened 2022-05-13
Import OkHttp's bom rather than declaring modules individually

ssobue:
According to the OkHttp document, it provides bill of materials. 
And it have same modules Spring's and OkHttp BOM.

So, replaces modules definition with BOM.

- Refrence documents
  - OkHttp Document: https://github.com/square/okhttp#releases
  - Maven Repository Informtion: https://mvnrepository.com/artifact/com.squareup.okhttp3/okhttp-bom/4.9.3


wilkinsona:
@ssobue Thanks very much for making your first contribution to Spring Boot.

PR #31033, opened 2022-05-14
App fails to start when it depends on thymeleaf-extras-springsecurity6 but does not have Spring Security on the classpath

terminux:
Spring Security dependency check for `SpringSecurityDialect` seems to be lost when restoring support for Thymeleaf, see 12cd97a. So I restored it and added some unit tests.

Related: #28967


wilkinsona:
Thanks very much, @terminux. I pulled the test side of this out into [a separate issue](https://github.com/spring-projects/spring-boot/issues/31097) so it could be tackled in our maintenance branches too.

Issue #28139, opened 2021-09-27
OAuth2ClientPropertiesRegistrationAdapter improve issuerUri value check


bitxon:
**Expected Behavior**
`OAuth2ClientPropertiesRegistrationAdapter.getBuilderFromIssuerIfPossible` should validate that `issuer` is `not blank`
Because few lines after, code throws exception if this `issuer` is `null` or `emptyString` or `blankString`

**Current Behavior**
`OAuth2ClientPropertiesRegistrationAdapter.getBuilderFromIssuerIfPossible` validates that `issuer` is `not null`

**Context**
```java
class OAuth2ClientPropertiesRegistrationAdapter {
  ...
  private static Builder getBuilderFromIssuerIfPossible(...) {
      ...
      if (issuer != null) {  // this code performs only null check
        Builder builder = ClientRegistrations.fromIssuerLocation(issuer).registrationId(registrationId);
        return getBuilder(builder, provider);
      }
      ...
  }
  ...
}
```
```java
public static ClientRegistration.Builder fromIssuerLocation(String issuer) {
  Assert.hasText(issuer, "issuer cannot be empty"); // this code perform null check + String content check
  ...
}
```

wilkinsona:
Thanks for the suggestion, @bitxon. Can you provide a bit of background on this please? I'd like to understand how you came to have the property configured but with an empty value.

If we treat an empty string as we currently treat `null`, we'll silently ignore the issuer URI configuration and use the provider's issuer URI instead. It sounds like that's what you'd prefer to happen, but someone else may prefer the current failure. If they have been trying to set the issuer URI and have made a mistake that led to it having an empty value, silently ignoring it may make that mistake harder to spot.

bitxon:
If in `application.yml` we have following configuration 
```
spring.security.oauth2.client:
  provider:
    keycloak:
      issuerUri: ${oauthUrl}
      authorization-uri: ${oauthUrl}/protocol/openid-connect/auth
      token-uri: ${oauthUrl}/protocol/openid-connect/token
      jwk-set-uri: ${oauthUrl}/protocol/openid-connect/certs
      user-info-uri: ${oauthUrl}/protocol/openid-connect/userinfo
      user-name-attribute: preferred_username
```
So we will not be able to unset issuerUri in `application-test.yml` without tricks.  Because following configuration will be considered as value with empty string:
```
spring.security.oauth2.client:
  provider:
    keycloak:
      issuerUri:
```

[...]

Issue #24133, opened 2020-11-12
ConfigurationProperties Cannot be cleared at runtime in some situation


chengaofeng:
Application has a ben annotated with```@ConfigurationProperties("test")``` that has a ```List<String>``` attribute called ```users```. at the beginning, one item listed in the external configuration file

```
test:
  users:
  - Andy
```

as the application is running, change the configuration file to the following

```
test:
  users:
```

or 

```
test:
  users: null
```

what I expect is ```users=null``` or ```users =[]```,while the result is still ```users[0]=Andy```

<b>it‘s ok for the following pattern :</b>

```
test:
  users: ""

```

Here's a test that fails  with Spring Boot ```2.1.17.RELEASE``` 

```
@SpringBootTest(classes=ConfigPropertiesTest.class)
@RunWith(SpringRunner.class)
@EnableConfigurationProperties(ConfigPropertiesTest.TestProperties.class)
@TestPropertySource(properties = {"test.users[0]=Andy"})
public class ConfigPropertiesTest {
    @Autowired
    private TestProperties properties;
    @Autowired
    private ConfigurationPropertiesBindingPostProcessor processor;
    @Autowired
    private ConfigurableEnvironment environment;

    @Test
    public void liveRebindToNull() throws Exception {
        assertEquals(properties.getUsers().get(0),"Andy");
        setNullProperties("test.users");
	assertTrue(environment.getPropertySources().get("test").containsProperty("test.users"));;
        assertNull(environment.getPropertySources().get("test").getProperty("test.users"));

        processor.postProcessBeforeInitialization(properties, "testProperties");
        assertEquals(properties.getUsers().size(), 0); //**_TEST NG_**
    }

    /**
     * TestPropertyValues has no public method set null value, using reflect to do it
     * 
     * @param prefix
     * @throws Exception
     */
    private void setNullProperties(String prefix) throws Exception {
        TestPropertyValues testPropertyValues =  TestPropertyValues.empty();
        Method method = TestPropertyValues.class.getDeclaredMethod("and", Stream.class);
        method.setAccessible(true);

        TestPropertyValues.Pair pair = new TestPropertyValues.Pair(prefix,null);
        testPropertyValues = (TestPropertyValues) method.invoke(testPropertyValues, Stream.of(pair));

        testPropertyValues.applyTo(environment);
    }

    @ConfigurationProperties("test")
    public static class TestProperties {
        private List<String> users;

        public List<String> getUsers() {
            return users;
        }

[...]

PR #31037, opened 2022-05-16
Support empty issuer-uri in OAuth2 provider

natrem:
When using a custom OAuth2 provider then its discovery occurs during the application startup which complicate the use of HTTP mock servers.

If the application defines a default issuer-uri - in order to ease configuration management in multiple environments - then it is not possible to reset the value to null in unit tests.

```yaml
spring:
  security:
    oauth2:
      client:
        registration:
          example-client-api1:
            provider: my-company-oauth2-server
            client-id: xxx
            client-secret: xxx
            authorization-grant-type: client_credentials
            scope: scope1
        provider:
          my-company-oauth2-server:
            issuer-uri: https://my-company-oauth2-server/oauth2
```

Defining ``spring.security.oauth2.client.provider.my-company-oauth2-server.issuer-uri=`` results in the property being an empty string. The consequence is that the unit test will call the discovery endpoint even if we define a custom endpoint (access_token in this case). This is due to the behavior of [OAuth2ClientPropertiesRegistrationAdapter#getBuilderFromIssuerIfPossible](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientPropertiesRegistrationAdapter.java#L81)

I often use mock-server which allows an actual connection made to an HTTP server with dynamic port.
Since the OAuth2 discovery occurs during the application start, many mock server (in this case [mock-server](https://mock-server.com/)) make it difficult or complicated to use a dynamic port and set up expectations before the Spring application initialization.

This evolution make Spring Boot consider that an empty issuer-uri is similar to an unspecified one, thus making easy to configure a test with no issuer:

```java
	@MockServerTest({
		"spring.security.oauth2.client.provider.my-company-oauth2-server.issuer-uri=",
		"spring.security.oauth2.client.provider.my-company-oauth2-server.token-uri=http://localhost:${mockServerPort}/oauth2/access_token"
	})
```

An alternative with the current implementation is to define the issuer-uri in a profile, which is activated by default and deactivated in the tests.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientPropertiesRegistrationAdapter.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientPropertiesRegistrationAdapter.java
@@ -79,7 +79,7 @@ public final class OAuth2ClientPropertiesRegistrationAdapter {
 		if (providers.containsKey(providerId)) {
 			Provider provider = providers.get(providerId);
 			String issuer = provider.getIssuerUri();
-			if (issuer != null) {
+			if (StringUtils.hasText(issuer)) {
 				Builder builder = ClientRegistrations.fromIssuerLocation(issuer).registrationId(registrationId);
 				return getBuilder(builder, provider);
 			}


wilkinsona:
Thanks for the proposal. We have already considered making this change in https://github.com/spring-projects/spring-boot/issues/28139 and decided not to do so. Defining the issuer URI in a profile is our recommended approach at this time. https://github.com/spring-projects/spring-boot/issues/24133 is tracking a more general solution,

natrem:
Sorry for missing the issue in my search (not thorough enough) - and the lost time. Thank you for checking.

PR #31067, opened 2022-05-16
add equals and hashCode to MongoProperties

XenoAmess:
Add equals and hashCode to MongoProperties

Because in my company we want to use them as key in HashMap, thus build a map from properties to MongoClient, and make it a cache pool.

Yes we have a mechanism to config multiple MongoProperties, and some of them are same, others not.


snicoll:
Thanks for the suggestion but we don't want to do this ad-hoc and adding equals/hashCode on all `@ConfigurationProperties` type can be tricky.

PR #31104, opened 2022-05-18
Allow Prometheus PushGateway to be configured to perform a PUT on shutdown

cansi:
Metrics push to Prometheus Pushgateway can happen in two different ways:

- with PUT HTTP method, all existing metrics with the same grouping keys are replaced by new metrics
- with POST HTTP method, works exactly like the PUT method but only metrics with the same name as the newly pushed metrics are replaced (among those with the same grouping key)

from [Prometheus Pushgateway README](https://github.com/prometheus/pushgateway#post-method).

Right now, PrometheusPushGatewayManager is only configurable to use the POST method by specifying the `PUSH` shutdown operation (could be useful to rename it to `POST`, not done yet).

The present PR expands the `ShutdownOperation` enum with a new `PUT` value that will be used by PrometheusPushGatewayManager to use the correspondent method of the internal PushGateway instance.


wilkinsona:
@cansi Thank you very much for making your first contribution to Spring Boot

PR #31122, opened 2022-05-19
Use Spring Batch Bom

hpoettker:
Starting with the release of Spring Batch 5.0.0-M3, a bom is provided: https://github.com/spring-projects/spring-batch/issues/1097


wilkinsona:
Thanks very much, @hpoettker.

Issue #31124, opened 2022-05-19
Switch CI to Bellsoft JDK


philwebb:
CI could be switched to the Bellsoft JDK to align with buildpacks.

Details of the API are at https://api.bell-sw.com/api.html#/Binaries/get_liberica_releases

We'd need something like this in our existing scripts:

```
curl "https://api.bell-sw.com/v1/liberica/releases?version-feature=17&version-modifier=latest&os=linux&arch=x86&bitness=64&package-type=tar.gz&bundle-type=jdk"
```

wilkinsona:
Closing in favor of #31139.

PR #31139, opened 2022-05-21
Use Bellsoft JDK in CI

eddumelendez:
See gh-31124


wilkinsona:

Issue #31121, opened 2022-05-19
Consider adding `.sdkmanrc`


vpavic:
It would be nice if `.sdkmanrc` would be added to all supported branches in order to make project handling a bit easier for SDKMAN users. For example, Spring Framework already [does this](https://github.com/spring-projects/spring-framework/blob/main/.sdkmanrc).

philwebb:
I thought we has an issue for this already, but I can't find it. We have discussed it and want to add one. Probably also align JDK versions in CI.

wilkinsona:
Closing in favor of #31141.

PR #31141, opened 2022-05-21
Add .sdkmanrc with latest java version

eddumelendez:
Using java 8 as the lowest jdk supported by 2.7.x. For 3.x, `17.0.3-librca` can be used.

See gh-31121


wilkinsona:

PR #31143, opened 2022-05-21
Use Lambda-based API in Spring Security examples

### LINKED ISSUE #30432: For consistency with Spring Security's reference documentation, use Lambda-based APIs in Spring Security examples
This was raised on #29932 which added an example for securing H2's web console, but we should do a broader review as there may be other examples that should also be updated.

hpoettker:
This PR is intended to resolve #30432.

I've formatted the Spring Security examples consistently such that
- the last word of the respective method name is chosen as the variable name in the lambda,
- `withDefaults()` is used where no actual `Customizer` is used,
- chaining is not used as that is sometimes hard to style nicely if lines are continued.

If a different style is preferred, I'll be happy to change it.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/data/sql/h2webconsole/springsecurity/DevProfileSecurityConfiguration.java
+++ b/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/data/sql/h2webconsole/springsecurity/DevProfileSecurityConfiguration.java
@@ -22,6 +22,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Profile;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
+import org.springframework.security.config.Customizer;
 import org.springframework.security.config.annotation.web.builders.HttpSecurity;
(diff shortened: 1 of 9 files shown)


spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/data/sql/h2webconsole/springsecurity/DevProfileSecurityConfiguration.java
```diff
@@ -32,13 +33,18 @@ public class DevProfileSecurityConfiguration {
...
 	SecurityFilterChain h2ConsoleSecurityFilterChain(HttpSecurity http) throws Exception {
-		// @formatter:off
-		return http.requestMatcher(PathRequest.toH2Console())
-				// ... configuration for authorization
-				.csrf().disable()
-				.headers().frameOptions().sameOrigin().and()
-				.build();
-		// @formatter:on
+		http.requestMatcher(PathRequest.toH2Console());
+		http.authorizeRequests(yourCustomAuthorization());
```
hpoettker, line 37: To underline that the console should not be left unsecured without prescribing a security setup as discussed in #29932.

Issue #31133, opened 2022-05-20
Mimepull 1.10.0 dependency breaks compatibility with Java 8 for Spring Boot 2.7.0


symphony-youri:
In Spring Boot 2.7.0, [mimepull](https://github.com/eclipse-ee4j/metro-mimepull) has been upgraded to 1.10.0 (#30725).
It seems this version has been built with Java 11 (https://github.com/eclipse-ee4j/metro-mimepull/pull/61), making Spring Boot 2.7.0 incompatible with Java 8 if the library is loaded (unless you override the version of course).

For instance, here is the error during the startup of a Spring Boot application using Jersey:

```
java.lang.RuntimeException: java.util.concurrent.ExecutionException: java.lang.UnsupportedClassVersionError: org/jvnet/mimepull/MIMEConfig has been compiled by a more recent version of the Java Runtime (class file version 55.0), this version of the Java Runtime only recognizes class file versions up to 52.0
	at org.glassfish.hk2.utilities.cache.LRUHybridCache.compute(LRUHybridCache.java:292)
	at org.glassfish.hk2.utilities.reflection.internal.ClassReflectionHelperImpl.getAllMethods(ClassReflectionHelperImpl.java:85)
	at org.jvnet.hk2.internal.Utilities.findInitializerMethods(Utilities.java:1400)
	at org.jvnet.hk2.internal.DefaultClassAnalyzer.getInitializerMethods(DefaultClassAnalyzer.java:83)
	at org.glassfish.jersey.inject.hk2.JerseyClassAnalyzer.getInitializerMethods(JerseyClassAnalyzer.java:218)
	at org.jvnet.hk2.internal.Utilities.getInitMethods(Utilities.java:196)
	at org.jvnet.hk2.internal.ClazzCreator.initialize(ClazzCreator.java:121)
	at org.jvnet.hk2.internal.ClazzCreator.initialize(ClazzCreator.java:156)
	at org.jvnet.hk2.internal.SystemDescriptor.internalReify(SystemDescriptor.java:716)
....
```

philwebb:
We have a similar situation with Jetty where we pin to v9 because v10+ requires Java 11. Ideally I think we should be consistent and do the same with mimepull, but unfortunately we've already cut the release. I'm now not sure if we should downgrade or not.

Flagging for team attention to see what others think.

philwebb:
@symphony-youri in the meantime you can try setting the `mimepull.version` property in your build to downgrade to the previous version.

[...]

Issue #14924, opened 2018-10-22
Web Services starter depends on a missing mimepull version


Crow-EH:
Hi everyone,

# The context

We tried to migrate one of our projects from Spring Boot 2.1.0.M4 to 2.1.0.RC1 and encountered an error while building. The project depends on spring-boot-starter-web-services.

# The error

While building :
```
* What went wrong:
Could not resolve all files for configuration ':compileClasspath'.
> Could not find org.jvnet.mimepull:mimepull:1.9.8.
  Required by:
      project : > org.springframework.boot:spring-boot-starter-web-services:2.1.0.RC1 > com.sun.xml.messaging.saaj:saaj-impl:1.5.0
```

# Our analysis

It seems like while the version 1.5.0 on[ the archived Java EE ](https://github.com/eclipse-ee4j/metro-saaj) repository (the one used in the starter) [depends on mimepull 1.9.8](https://github.com/javaee/metro-saaj/blob/1.5.0/saaj-ri/pom.xml#L482), mimepull isn't available on repositories like Maven Central.

While migrating to the new EE4J repository, it seems like saaj [has been silently upgraded to 1.5.1 ](https://github.com/eclipse-ee4j/metro-saaj/commit/9acf1248858d042b0a7590005bc4a1c7830643ec) and  its dependency to mimepull [has been rolled back to 1.9.7](https://github.com/eclipse-ee4j/metro-saaj/commit/9acf1248858d042b0a7590005bc4a1c7830643ec#diff-600376dffeb79835ede4a0b285078036R445).

This should make the application build again.

philwebb:
@Crow-EH Thanks for the analysis. Are you using Java 11 by any chance?

Crow-EH:
Hi @philwebb,

Yes we are.
In fact we migrated to 2.1.0 for Java 11 and it has been a breeze until this issue.

But the problem persists even with Java 8.
A simple project from Spring Boot initializr + Starter Web Services dependency is enough to reproduce.

philwebb:
@Crow-EH I've added a managed dependency to mimepull 1.9.10 which is in Maven Central. This should mean the transitive dependency gets replaced with the version that can be downloaded.

patrickmann:
@philwebb How do I get the fixed version? 2.1.0.RELEASE still looks for mimepull 1.9.8.

snicoll:
[...]

PR #31145, opened 2022-05-23
Dependency management for mimepull is redundant and the managed version is incompatible with Java 8

youribonnaffe:
Mimepull dependency was introduced in 1c18fd8 for gh-14924 to force the
version as the one coming from saaj-impl was not on Maven Central.
This is no longer the case.

Version 1.10.0 of Mimepull has been built with Java 11, breaking
compatibility with Java 8 for Spring Boot.

Closes gh-31133


wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @youribonnaffe.

PR #31151, opened 2022-05-23
Auto-configure JvmCompilationMetrics

izeye:
This PR changes to auto-configure `JvmCompilationMetrics` as it seems to have been missed somehow.


wilkinsona:
Thanks once again, @izeye.

PR #31172, opened 2022-05-25
Fix Custom Layers Configuration section title in Maven plugin docs

dalbani:


wilkinsona:
Thanks very much for spotting and fixing this, @dalbani.

Issue #31146, opened 2022-05-23
Support for Remote Cloud Development with Gitpod & GitHub Codespaces


Siddhant-K-code:
👋🏼 Hello! Siddhant here from Gitpod

I am thinking to add the Support for Remote Cloud Dev. so that folks can easily contribute to the project without any extra effort of setting up the environment, and Contributing from an Device.

Creating this issue for discussing the approach, before creating an PR.

You Can Checkout the [[Blog] Gitpod for Open Source](https://www.gitpod.io/blog/gitpod-for-opensource)

---
<a href="https://gitpod.io/#https://github.com/spring-projects/spring-boot">
  <img src="https://img.shields.io/badge/Contribute%20with-Gitpod-908a85?logo=gitpod" alt="Contribute with Gitpod" />
</a>

## Requirements
- We would need a `.gitpod.yml` that would do all the setup and create a prebuild every time a contributor stars the environment in Gitpod.

wilkinsona:
@Siddhant-K-code Thanks for raising this. We're definitely interested in Gitpod integration but we're stretched pretty thin at the moment. What would it require to "do all the setup and create a prebuild"? We'd like to get a feel for how much work will be involved both initially and in terms of ongoing maintenance.

Siddhant-K-code:
Thanks for your interest in Gitpod integration.
I have created a PR #31185 for the same. And this would not affect the ongoing maintenance.

wilkinsona:
Closing in favor of #31185.

PR #31185, opened 2022-05-27
Add Support for Remote Cloud Development with Gitpod

Siddhant-K-code:
> Related to #31146

## About this PR
This PR Contains a `.gitpod.yml` which enables the prebuild and creates a ready to code workspace for contributors.

## Maintainability
You do not have to change anything, until or unless you change the build commands flow. In all cases it would work perfect. & This Will Not affect the ongoing maintenance 

<a href="https://gitpod.io/#https://github.com/gitpod-forks/spring-projects-spring-boot">
  Contribute with Gitpod
</a>

**_PS: Prebuilds are enabled - saving ~3 minutes of build/ workspace startup time from a cold state._**

Gitpod Prebuild Saved Building Time


.gitpod.yml
```diff
@@ -0,0 +1,139 @@
...
+#  - name: Web Browser
+#    openMode: split-right
+#    env:
+#      WEBSERVER_PORT: 8080
+#    command: |
+#     gp await-port $WEBSERVER_PORT
+#     lynx `gp url`
+
+tasks:
+  - init: yes | sdk install java && ./gradlew build
```
wilkinsona, line 84: Is it possible to tie this to the version that's configured in `.sdkmanrc`? That's particularly important for branches other than `main` where a Java 8 install is required.

.gitpod.yml
```diff
@@ -0,0 +1,139 @@
...
+## for GitHub repositories when the GitHub application
+## 'https://github.com/apps/gitpod-io' is installed in GitHub and granted
+## permissions to access the repository.
+##
+## Learn more at 'https://www.gitpod.io/docs/prebuilds'
+
+github: 
+  prebuilds:
+    # enable for the default branch
+    master: true
```
wilkinsona, line 127: The repository's default branch is named `main`. Should this be `main` rather than `master`?

.gitpod.yml
```diff
@@ -0,0 +1,139 @@
...
+    # enable for all branches in this repo
+    branches: true
+    # enable for pull requests coming from this repo
+    pullRequests: true
+    # enable for pull requests coming from forks
+    pullRequestsFromForks: true
+    # add a check to pull requests
+    addCheck: true
```
wilkinsona, line 135: What's the purpose of the `check`? What does Gitpod do as part of checking a pull request?

.gitpod.yml
```diff
@@ -0,0 +1,139 @@
...
+    # add a "Review in Gitpod" button as a comment to pull requests
+    addComment: false
+    # add a "Review in Gitpod" button to the pull request's description
+    addBadge: true
```
wilkinsona, line 139: To avoid the risk of editing conflicts, I think we might prefer to have this disabled too.

.gitpod.yml
ghuntley: 👀 that's tech-debt on Gitpod end. Thanks for pointing it out "master" represents the default branch (whatever it is called). We are tracking resolution over at https://github.com/gitpod-io/gitpod/issues/10358

.gitpod.yml
Siddhant-K-code: It is kind of CI system, it checks the status of gitpod prebuild on PR. 

> More detailed docs.- `prebuilds.addCheck`: https://www.gitpod.io/docs/references/gitpod-yml#prebuildsaddcheck

.gitpod.yml
Siddhant-K-code: It won't create any editing conflict. It would just add a button on PR, so that PR reviewer can easily open that PR in gitpod & test it out. 

> Related Docs.: [`prebuilds.addBadge`](https://www.gitpod.io/docs/references/gitpod-yml#prebuildsaddbadge)

.gitpod.yml
Siddhant-K-code: yeah, we can directly use the command 
```yml
sdk env install
``` 

to download the required Java Version.

Updated the init command.

.gitpod.yml
wilkinsona: Those docs, which is where I learned of the potential for a conflict, say that adding the badge "can also create a concurrent editing conflict when the bot and a user try to edit the description of a pull request at the same time".

Siddhant-K-code: Resolved in [`8785659`](https://github.com/spring-projects/spring-boot/commit/8785659a8909f8f1d252494542adaafbeeb85330)

wilkinsona (commented):
Thanks for the updates, @Siddhant-K-code. I wanted to try things out so I opened https://gitpod.io/#https://github.com/gitpod-forks/spring-projects-spring-boot to pick up the changes in your fork.

It took a while for the IDE to be ready as it downloaded JDK 17 and ran the build. Am I right in thinking that a pre-build will typically take care of this?

An error popped up during the project import that said the following:
> Cannot activate the 'Test Runner for Java' extension because it depends on the 'Language Support for Java(TM) by Red Hat' extension, which is not installed. Would you like to install the extension and reload the window?

Is it possible for this extension to be pre-installed so that users get an error-free experience when they first open the project on Gitpod?

Siddhant-K-code:
> Is it possible for this extension to be pre-installed so that users get an error-free experience when they first open the project on Gitpod?

Yeah, we can do that, will do the required changes.

_Related Doc: [vscode-extensions in `.gitpod.yml` file](https://www.gitpod.io/docs/ides-and-editors/vscode-extensions)_

wilkinsona:
> Thanks for pointing this out, it is a known issue we are tracking in https://github.com/gitpod-io/gitpod/issues/9839

Thanks for the link to the Gitpod issue. I'll mark this as blocked until it has been resolved.

Siddhant-K-code:
I have added the needed extension for java, that issue will not affect now, as it would install the extension while creating the dev. environment. We can merge this!

wilkinsona:
Unfortunately, it doesn't seem to work. With your latest changes, the first time I opened https://gitpod.io/#https://github.com/gitpod-forks/spring-projects-spring-boot, it reported an error that the Java extension wasn't installed. The second time, it seemed to install the Java extension automatically but then reported an error that said "https://gitpod.io/#https://github.com/gitpod-forks/spring-projects-spring-boot "redhat.java extension is not synced, but not added in .gitpod.yml". In both cases no Java projects were found in the workspace.

Siddhant-K-code:
I have installed the [required VsCode Extensions](https://marketplace.visualstudio.com/items?itemName=vscjava.vscode-java-pack) 

But, it is now showing this, is there something issue with dev. environment or some build command? can you pls help me on this. Getting this: 
image

wilkinsona:
The error in the first screenshot means that it's using a version of Java lower than 17 to run the build. The second looks to be unrelated but indicates that something else isn't quite right as it's using a nightly build of Gradle 7.5. It should be using the Gradle wrapper (`./gradlew`) which is Gradle 7.4.2 at this time.

wilkinsona:
Thanks for the updates.

I've opened https://gitpod.io/#https://github.com/gitpod-forks/spring-projects-spring-boot in my browser. 25 minutes later, the prebuild is still running. I can understand the benefits of it downloading all of the dependences  so that things work smoothly in the IDE, but it's running all of the tests which feels unnecessary to me. I suspect it might be a lot quicker if https://github.com/gitpod-forks/spring-projects-spring-boot wasn't 100s of commits being https://github.com/spring-projects/spring-boot as cache hits from Gradle Enterprise would be more likely. Regardless of that, the current prebuild configuration feels suboptimal to me.

I'll take another look once the prebuild has completed and I've had a chance to use the IDE.

Siddhant-K-code:
Sorry for kept you waiting @wilkinsona @snicoll 

I think, it should work now. I have put the build commands to `init` that means it would be build before a workspace will start.

image


To get the full benefit of [prebuilds](https://www.gitpod.io/docs/prebuilds) (which would save time) you need to install [the Gitpod app](https://github.com/apps/gitpod-io) to this repository.

This Saves Build time, ([very recently Node does which can save around ~41 minutes every time a person open a gitpod workspace](https://github.com/nodejs/node/pull/43698))

Also, I think we should skip the running tests while build. Any flag in build command to skip tests?

wilkinsona:
The prebuild failed after almost an hour with the following message

> "We are having trouble connecting to the server. Either you are offline or websocket connections are blocked."

Clicking `Skip Prebuild` then allowed the IDE to open in my browser but there's no Java extension installed. I can open the code, but it's treated as plain text.

Unfortunately, it's getting increasingly difficult to justify spending more time on this. Flagging for discussion at a team meeting so that we can decide what we want to do.

Siddhant-K-code:
> The prebuild failed after almost an hour with the following message
> > "We are having trouble connecting to the server. Either you are offline or websocket connections are blocked."

It was due to, Gitpod has 1 hour timeout for the prebuilds. I think with skipping the tests check while build will decrease this time. & It would work smoothly.


> Clicking Skip Prebuild then allowed the IDE to open in my browser but there's no Java extension installed. I can open the code, but it's treated as plain text.

It was in the previous commit (outdated workspace image). Now, this problem won't be there. I have updated the workspace image.

> Unfortunately, it's getting increasingly difficult to justify spending more time on this. Flagging for discussion at a team meeting so that we can decide what we want to do.

Sorry for the friction. Would be fix soon 🤞🏼

wilkinsona:
>  I think with skipping the tests check while build will decrease this time. & It would work smoothly

We can't skip the tests entirely, unfortunately. Some of our documentation is test-driven so skipping them will prevent the documentation from being built. Running `./gradlew assemble` may be a better option. That'll build everything and only run the handful of tests that are used to generate documentation. If it would be beneficial to also pre-download test dependencies, `./gradlew assemble compileTestJava` should still be relatively quick.

Build time should also be improved by our Gradle Enterprise cache that's hosted at https://ge.spring.io. Cache retention is relatively short so your branch being 600 commits behind our main branch will be negating its benefits. Perhaps you can rebase and force-push?

Siddhant-K-code:
> We can't skip the tests entirely, unfortunately. Some of our documentation is test-driven so skipping them will prevent the documentation from being built. Running ./gradlew assemble may be a better option. That'll build everything and only run the handful of tests that are used to generate documentation. If it would be beneficial to also pre-download test dependencies, ./gradlew assemble compileTestJava should still be relatively quick.

Thanks for the suggestions, I have updated the `.gitpod.yml` file. It should work now.

> Build time should also be improved by our Gradle Enterprise cache that's hosted at https://ge.spring.io/. Cache retention is relatively short so your branch being 600 commits behind our main branch will be negating its benefits. Perhaps you can rebase and force-push?

Updated that 👍🏼

bclozel:
The team discussed this PR and unfortunately we're going to decline it.
Having seen how difficult it is to set up, we're not confident in our ability to maintain this in the long run. We don't use a cloud setup on a day-to-day basis, so only newcomers are likely to find out about issues. On the other hand, we're maintaining our Contributing docs and have a dedicated first-time-contributor process that brought many new contributors to the project.

Thanks @Siddhant-K-code !

PR #31190, opened 2022-05-27
Document that when using Lombok it must be configured to run before spring-boot-configuration-processor

### LINKED ISSUE #30986: Document ordering of Lombok and spring-boot-configuration-processor annotation processors
I just made the same mistake as the reporter in #23405 that I had `spring-boot-configuration-processor` ordered before `lombok` in my pom.xml and thus combining `@lombok.Value` with `@ConstructorBinding` didn't work with the configuration-processor.

It would be great to have an info box in the [Configuration Metadata](https://docs.spring.io/spring-boot/docs/current/reference/html/configuration-metadata.html#appendix.configuration-metadata.annotation-processor) mentioning the importance of the order of the dependencies if using Lombok.

tudormarc:
Added note to use lombok dependency before spring-boot-configuration-processor dependency.

Closes #30986

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/configuration-metadata/annotation-processor.adoc
@@ -59,6 +59,29 @@ You could also let the AspectJ plugin run all the processing and disable annotat
 ----
 ====
 
+[NOTE]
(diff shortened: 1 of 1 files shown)


snicoll:
@tudormarc thank you for making your first contribution to Spring Boot.

tudormarc:
@snicoll thanks a lot. I'm really happy to be able to help :)

PR #31230, opened 2022-06-01
OAuth2 Resource Server Auto-Configuration can only configure a single JWS algorithm

rs017991:
Spring Security added support for specifying multiple JWS algorithms in https://github.com/spring-projects/spring-security/pull/7162

However, since the auto-configuration was not updated, it was only possible to leverage that enhancement via a custom bean.
_(a bit clunky and undesirable compared to setting it in yml)_

This PR fills the gap by allowing the existing property to contain a comma-delimited list of algorithms.
_(if only one algorithm is defined as before, then it will behave no differently)_

I'd like to get this backported to `2.7` as well as `2.6` if at all possible.
**Please let me know if/when you would like me to do this.**

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/reactive/ReactiveOAuth2ResourceServerJwkConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/reactive/ReactiveOAuth2ResourceServerJwkConfiguration.java
@@ -20,6 +20,7 @@ import java.security.KeyFactory;
 import java.security.interfaces.RSAPublicKey;
 import java.security.spec.X509EncodedKeySpec;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.Base64;
 import java.util.Collections;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


rs017991:
It's all green now; I was able to appease checkstyle, but I couldn't figure out how to get the formatter to stop mangling my code so I had to disable it.

wilkinsona:
Thanks for raising this, @rs017991. It's unfortunate that we missed this when the change was made in Spring Security 5.2. To consider making the proposed changes in 2.6.x or 2.7.x, we'd have to consider this to be a bug as we do not make enhancements in maintenance releases. It could be argued that it's a bug of omission. I'll label the issue for discussion at a team meeting so that we can consider our options.

philwebb:
We're going to look at adding a `jwsAlgorithms` `String[]` property and deprecate `jwsAlgorithm`. We'll treat this one as a bug and attempt to fix it in 2.7.

rs017991:
Thanks for the update @philwebb. I'll be afk for a few weeks, so feel free to rework my contribution as you see fit in my absence.

wilkinsona:
Thanks for the pull request, @rs017991. Unfortunately, the approach that I think we need to take is sufficiently different to what's proposed here that I don't think it's worth building on top of these changes as we'd revert them almost entirely. Thanks anyway for bringing the problem to our attention. I've opened gh-31321 to track the problem.

Issue #31025, opened 2022-05-13
CassandraAutoconfiguration leads to ignored parameters in db-config.conf


nilsvillwock:
We are using the cassandra driver specific configuration file `db-config.conf` via the property:

```yaml
spring:
  data:
    cassandra:
      config: file:/db-config.conf
```

The problem now is that even when I do not define further spring-boot `CassandraProperties` in the yaml, there are default values already set and these properties are ignored in the `db-config.conf`. file. The reason seem to be, because the `db-config.conf` properties are loaded as fallback after the default spring boot `CassandraProperties`. Spring Boot `CassandraProperties` should win, but only if they are actively set.

The following Cassandra properties always initialized and so always ignored in `db-config.conf`

![initCassandraProperties](https://user-images.githubusercontent.com/46843697/168272455-9e4a00d6-6809-4598-a65f-9cff202fda0d.png)

snicoll:
@philwebb I can see you've flagged this as a bug but the current behavior is the expected behavior, see https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#data.nosql.cassandra.connecting. Perhaps the documentation should be more explicit that a default value counts as a value being present.

@nilsvillwock the purpose of the config property is to let define properties that the Spring Boot namespace does not support.

nilsvillwock:
OK, please adjust the documentation from:

If a property is both present in spring.data.cassandra.* and the configuration file, the value in spring.data.cassandra.* takes precedence.

to:

If a property is both present **(even it is only by default)** in spring.data.cassandra.* and the configuration file, the value in spring.data.cassandra.* takes precedence.

snicoll:
We've discussed this at the team meeting and we come to realize that it was supposed to work as expected (see #24065). The problem is that some properties were overlooked, typically the ones that have a primitive type (and a default value of 0).

@nilsvillwock it would be good to list the properties that you tried to define in `db-config.conf` that did not work.

nilsvillwock:
@snicoll 
The properties which are ignored in the db-config file are the ones you can see at the screenshot  in my initial post

ittays:
Can you please look into my PR? 
Eventually, to simplify the port configuration, I preferred not to defer the `contact-points` config, and not to cross-use a host from one config layer and a port from another.

snicoll:
@ittays there's no need to ask us to do that. The whole team already received a notification when you created the pR.

philwebb:
[...]

PR #31238, opened 2022-06-02
Parameters in spring.data.cassandra.config file will override defaults

ittays:
This commit changes two things:

1. Most primitives on `CassandraProperties` are replaced with object values.    This allows distinguishing between default-values and no-values. Then   `CassandraAutoConfiguration.mapConfig()` can use `whenNonNull()` predicate    to ignore those.

2. `CassandraProperties` no longer populate default values that are overridable by `spring.data.cassandra.config` file. This let the defaults to be applied on top of the file   `spring.data.cassandra.config`; i.e. the config file have higher    precedence than the defaults, but lower that any `spring.data.cassandra.*`    property.

On the test side, the file `override-defaults.conf` is defined to override the values that were non-overridable with values which are different from the default.

---

In this proposal, the `spring.data.cassandra.port` configuration has a bit rough behaviour:

1. In case property `spring.data.cassandra.contact-points` is configured without a port - the value of `spring.data.cassandra.port` (which defaults to 9042) is used.

2. Otherwise, in case `contact-points` on `spring.data.cassandra.config` file are defined -- then the contact-points must be configured with a port (as the conf file provides no separate port property).

3. Otherwise, `contact-points` defaults to `127.0.0.1:9042` regardless of `spring.data.cassandra.port` configuration.

---

Closes gh-31025

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cassandra/CassandraAutoConfiguration.java
@@ -126,7 +126,7 @@ public class CassandraAutoConfiguration {
 
 	private Config applyDefaultFallback(Config config) {
 		ConfigFactory.invalidateCaches();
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks very much for this contribution @ittays, it has now been merged into 2.6.x and forwards.

PR #31258, opened 2022-06-07
WIP: ReactorKafka auto-configuration (approach 1)

onobc:
This is a WIP proposal for ReactorKafka auto-configuration. It is a modification of the original proposal in that it leverages the already existing producer/consumer from KafkaProperties as well as supports property hierarchy of KafkaProperties. 

There are still tests that need to be fleshed out but I wanted to get something for us to look at before proceeding. 

### Commit 1
Has nothing to do w/ ReactorKafka and is strictly a refactor to pull up common properties for re-use. This is leveraged by the 2nd commit but it useful on its own. 

### Commit 2
Add the ReactorKafka auto-config. 

**NOTE:** I know the "ship has sailed" on this going into 2.7x but I left it here (FOR NOW) as ReactorKafka has not yet made it into the latest Reactor BOM for SB3 (seems to have issues w/ Java17 from 1st glance). 

**NOTE:** This approach inserts the RK props in the KP hierarchy - the same approach used by Admin, Producer, Consumer, and Streams clients. 

(+) Behaves like the other Kafka clients do and should be easy for Spring Kafka users to grok
(+) Allows setting producer/consumer at higher level (or not) and sharing (or not) w/ non-reactor clients
(-) The extra layer of overrides (hierarchy) complicates things a bit

@garyrussell I know you advocated for a separate RKP and that would be easy to implement w/ this refactored base properties in commit 1. I do think this approach feels very much like the other clients config props though. 

cc: @wilkinsona


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaAutoConfiguration.java
```diff
@@ -0,0 +1,98 @@
...
+		SenderOptions<K, V> senderOptions = SenderOptions.create(this.kafkaProperties.buildReactorProducerProperties());
+		KafkaProperties.Reactor.Sender senderProperties = this.kafkaProperties.getReactor().getSender();
+		PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
+		senderOptions = toInstanceSafe(map.from(senderProperties::getCloseTimeout), senderOptions::closeTimeout, senderOptions);
+		senderOptions = toInstanceSafe(map.from(senderProperties::getMaxInFlight), senderOptions::maxInFlight, senderOptions);
+		senderOptions = toInstanceSafe(map.from(senderProperties::isStopOnError), senderOptions::stopOnError, senderOptions);
+		return senderOptions;
+	}
+
+	private <T, R> R toInstanceSafe(PropertyMapper.Source<T> source, Function<T, R> target, R defaultOptions) {
```
onobc, line 87: I am going to submit a separate PR to add an API to Source.toInstance that accepts a default instance when the value was filtered.

spring-boot-project/spring-boot-autoconfigure/src/test/resources/kafka/application-sample-reactor.yml
```diff
@@ -0,0 +1,92 @@
...
+      value-serializer: com.example.FooSerializer
+      ...
+      # common-props
+      bootstrap-servers: ...
+      client.id: ...
+      ssl: ...
+      security: ...
+      properties: ...
+
+    reactor:
```
onobc, line 43: **[TEMPORARY]**
This file only exists to illustrate the "full" yaml available.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaProperties.java
```diff
@@ -137,533 +80,107 @@ public Streams getStreams() {
...
-	 * @return the consumer properties initialized with the customizations defined on this
+	 * default kafkaAdmin bean.
+	 * @return the admin properties initialized with the customizations defined on this
 	 * instance
 	 */
-	public Map<String, Object> buildConsumerProperties() {
-		Map<String, Object> properties = buildCommonProperties();
-		properties.putAll(this.consumer.buildProperties());
+	public Map<String, Object> buildAdminProperties() {
+		// spring.kafka.<common-props>
```
onobc, line 108: ℹ️ will probably yank these prefix/path comments out later - just easily shows what the mapping is for this proposal review only. It may be helpful to move them up to the method javadoc though in some capacity.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaPropertiesBaseWithBootstrapServersAndProducerConsumer.java
```diff
@@ -0,0 +1,90 @@
...
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+
+/**
+ * Common Spring Kafka configuration properties - including bootstrap servers and producer and consumer
+ * Kafka client properties.
+ *
+ * @author Chris Bono
+ * @since 2.7.0
+ */
+class KafkaPropertiesBaseWithBootstrapServersAndProducerConsumer extends KafkaPropertiesBaseWithBootstrapServers {
```
onobc, line 31: ℹ️ This hierarchy is a bit forced and brittle by nature of the config properties method / naming convention. Using inheritance makes the get/set available. Otherwise I would have explored composition.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ReactiveKafkaAutoConfiguration.java
onobc: https://github.com/spring-projects/spring-boot/pull/31323

onobc:
> @onobc thank you for the PR but why can't we continue collaborating on #30567?

Hi @snicoll - 
The direction was unclear and there were enough questions that we wanted to stop the existing proposal to allow some exploration of approaches. The propsals are fairly divergent at this point so I think it makes sense to focus on this more recent one. I thought that we closed the other PR but I see it is open still which is a bit confusing.

Issue #31076, opened 2022-05-17
Project cannot be build under WSL due to NAME environment variable


vpavic:
I'm unable to build the project in WSL environment due to `ConfigurationPropertiesTests#loadWhenBindingWithDefaultsInXmlShouldBind` failing. More specifically, this assertion is failing:
https://github.com/spring-projects/spring-boot/blob/f17a00aad1d2d5753e5313268a2b40f1568bd22a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java#L258

It seems that this is due to WSL exposing the hostname (which is the actual value of `name` property I'm seeing when the test fails) in the `NAME` environment variable and thus overrides the value set in `spring-boot-project/spring-boot/src/test/resources/org/springframework/boot/context/properties/testProperties.xml`.

vpavic:
I can confirm that the project now builds successfully in WSL - thanks for the prompt fix.

LinkDownZ:
The value of the `NAME` variable is indeed the same as the Windows hostname, and the compilation can temporarily succeed by using `unset NAME`.

PR #31267, opened 2022-06-07
Update smoke tests to avoid conflicts with NAME environment variable

vpavic:
This commit updates several smoke tests in order to avoid conflicts with `NAME` environment variable that is present in WSL and causes project build to fail. Previous attempt to fix this in 7da42d71 (see #31076) was incomplete.


wilkinsona:
Thanks very much, @vpavic.

PR #31276, opened 2022-06-08
Remove "earlier in this chapter" from places where content is now elsewhere in the documentation

anthonyvdotbe:
Remove all occurrences of `, earlier in this chapter`, as this is no longer true.


snicoll:
@anthonyvdotbe thank you for making your first contribution to Spring Boot.

Issue #31301, opened 2022-06-08
Malformed json causes BasicJsonParser to throw a NullPointerException


philwebb:
Adding the following to `AbstractJsonParserTests` will cause all our parser implementations to fail, however, `BasicJsonParser` fails with an ugly `NullPointerException` rather than something that tells the user that the JSON is bad:

```java
@Test
void malformedJson() {
    String input = "[tru,erqett,{\"foo\":fatrue,true,true,true,tr''ue}]";
    List<Object> list = this.parser.parseList(input);
    assertThat(list).hasSize(3);
}
```

philwebb:
@wilkinsona pointed out that `{ "foo" }` is enough to trigger the NPE

wilkinsona:
`YamlJsonParser` is a bit funky at the moment. It parses `{ "foo" }` into a map of:

- `foo` `->` `null`

`[tru,erqett,{\"foo\":fatrue,true,true,true,tr''ue}]` becomes a list of:

- `tru`
- `erqett`
- A map of:
  - `foo` `->` `fatrue`
  - `true` `->` `null`
  - `tr''ue` `->` `null`

Flagging for a team meeting as I'd like to discuss if there's anything we can and should do about this.

I've also noticed that we aren't consistently using `tryParse`. Doing so would mean that each `JsonParser` implementation would consistently throw a `JsonParseException` when the underlying parser throws an exception.

My changes thus far are in [this branch](https://github.com/wilkinsona/spring-boot/tree/gh-31301).

philwebb:
We're going to merge this but ignore the YAML test failure. #31498 will remove the YAML parser in 3.0

PR #31304, opened 2022-06-09
Add null check on throws exception indicating malformed json if null

Egraveline:
This fixes this issue [#31301 ](https://github.com/spring-projects/spring-boot/issues/31301 ). 

This problem only applies to the BasicJsonParser since the other parser return exceptions that give details of the malformed JSON.

I choose the exception that the JacksonJsonParser returns on malformed JSON, JsonParseException. the GsonJsonParser returns a MalformedJsonException.

image

Not sure exactly if the project has a preferable way of handling null cases like this. 

Didn't add the proposed test from the issue because I suspect there will be another PR to write tests for malformed JSON that assert the exception on malformed JSON parsing.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/BasicJsonParser.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/BasicJsonParser.java
@@ -16,10 +16,7 @@
 
 package org.springframework.boot.json;
 
-import java.util.ArrayList;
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the pull request, @Egraveline, but there's a bit more to this one. Please see my [comment on the issue](https://github.com/spring-projects/spring-boot/issues/31301#issuecomment-1150721952) for details.

PR #31312, opened 2022-06-09
Circleci project setup

nexlesoft-devops:


snicoll:
Thanks but we already have a CI setup on Concourse.

PR #31313, opened 2022-06-09
Disable H2 Optimize Reuse Results

filiphr:
We recently updated to Spring Boot 2.7 and were hit by some strange issues with the new H2 2.x. After some analysis it turns out that there is some bug in how subqueries are evaluated and sometimes leading to wrong query results. 

The issue has been reported to H2 (https://github.com/h2database/h2database/issues/3534). The workaround currently is to disable optmize reuse of results by setting `;OPTIMIZE_REUSE_RESULTS=FALSE`. 

My proposal would be to automatically disable this for the Spring Boot auto generated in memory URL and remove it once the bug is fixed in H2 and Spring Boot upgrades to the new H2 version. 

I would also understand if you decide not to go down this path and reject this PR.


snicoll:
> My proposal would be to automatically disable this for the Spring Boot auto generated in memory URL and remove it once the bug is fixed in H2 and Spring Boot upgrades to the new H2 version.

Thanks for the suggestion but we're not keen of doing that. Users affected by the bug in a third-party library should apply this change themselves.

PR #31323, opened 2022-06-09
Add PropertyMapper toInstance that tolerates filtered values

onobc:
### Overview 
This proposal adds an API to `PropertyMapper.isInstance` that returns a user-specified default instance when the mapped value has been filtered. The current behavior is to throw an exception when the mapped value has been filtered.

###  Motivation
Some target objects are immutable and return a new instance w/ the mapped value rather than updating the value on the target instance. Consider the following example:

```java
class Target {

  private String foo;

  Target(String foo) {
     this.foo = foo;
  }
  
  Target updateFoo(String foo) {
    return new Target(foo);
  }
}
```
And suppose we have some config prop for "foo" and we want to update the target object's "foo" w/ that value. Typically we would use something like:
```java
PropertyMapper.get().from(someProps::foo).to(someTarget::updateFoo);
```
However, `Source.to()` does not work in the immutable case because `updateFoo` returns the new updated instance. Ok, lets instead use `Source.toInstance()` such as:
```java
someTarget =  PropertyMapper.get().from(someProps::foo).toInstance(someTarget::updateFoo);
```
This works fine when the value is not filtered. However, if we add some constraints on the mapper such as:
```java
someTarget =  PropertyMapper.get().from(someProps::foo).whenNonNull().toInstance(someTarget::updateFoo);
```
throws a `NoSuchElementException` when `someProps::foo` is null. This results in having to test the filter predicate manually prior to calling `toInstance`. 

### Proposed Solution
Allow the caller to pass in a default instance to use when the value is filtered (rather than throw exception). Something like this:
```java
someTarget =  PropertyMapper.get().from(someProps::foo).whenNonNull().toInstance(someTarget::updateFoo, someTarget);
```

You can see where I have worked around this in https://github.com/spring-projects/spring-boot/pull/31258/files#diff-7f2cc68fbde121f72bcff98a44dc29738be96efb911ffc54b80b6a6dd334b5ccR60

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/PropertyMapper.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/PropertyMapper.java
@@ -24,6 +24,7 @@ import java.util.function.Function;
 import java.util.function.Predicate;
 import java.util.function.Supplier;
 
+import org.springframework.lang.Nullable;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 import org.springframework.util.function.SingletonSupplier;
@@ -302,27 +303,31 @@ public final class PropertyMapper {
 		 * @throws NoSuchElementException if the value has been filtered
 		 */
 		public <R> R toInstance(Function<T, R> factory) {
-			Assert.notNull(factory, "Factory must not be null");
-			T value = this.supplier.get();
-			if (!this.predicate.test(value)) {
-				throw new NoSuchElementException("No value present");
-			}
-			return factory.apply(value);
+			return toInstance(factory, true);
 		}
 
 		/**
-		 * Complete the mapping by creating a new instance
+		 * Complete the mapping by creating a new instance from the non-filtered value.
 		 * @param <R> the resulting type
 		 * @param factory the factory used to create the instance
-		 * @return the created instance or empty when the value is filtered
+		 * @param failIfFiltered whether to throw exception or return {@code null} if the
+		 * value has been filtered
+		 * @return the instance or {@code null} if the value has been filtered and
+		 * {@code failIfFiltered} is {@code false}
+		 * @throws NoSuchElementException if the value has been filtered and
+		 * {@code failIfFiltered} is {@code true}
 		 */
-		public <R> Optional<R> toOptionalInstance(Function<T, R> factory) {
+		@Nullable
+		public <R> R toInstance(Function<T, R> factory, boolean failIfFiltered) {
 			Assert.notNull(factory, "Factory must not be null");
 			T value = this.supplier.get();
 			if (!this.predicate.test(value)) {
-				return Optional.empty();
+				if (failIfFiltered) {
+					throw new NoSuchElementException("No value present");
+				}
+				return null;
 			}
-			return Optional.ofNullable(factory.apply(value));
+			return factory.apply(value);
 		}
 
 		/**
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/PropertyMapper.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/PropertyMapper.java
@@ -18,6 +18,7 @@ package org.springframework.boot.context.properties;
 
 import java.util.NoSuchElementException;
 import java.util.Objects;
+import java.util.Optional;
 import java.util.function.Consumer;
 import java.util.function.Function;
 import java.util.function.Predicate;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/PropertyMapper.java
```diff
@@ -308,6 +309,23 @@ public <R> R toInstance(Function<T, R> factory) {
...
 
+		/**
+		 * Complete the mapping by creating a new instance from the non-filtered value or
+		 * using the specified default instance when the value is filtered.
+		 * @param <R> the resulting type
+		 * @param factory the factory used to create the instance
+		 * @param defaultInstance the instance to use if the value has been filtered
+		 * @return the instance
+		 */
+		public <R> R toInstance(Function<T, R> factory, R defaultInstance) {
```
onobc, line 320: I went w/ a direct value for the default rather than a supplier of the default. Mostly because that is the mode that works for my current need. 

However, I do see that for the `from` API we have both supplier and direct value version of the API. Should we do the same here, or keep one or the other?

philwebb:
I wonder if we should have something like `toOptionalInstance(factory)` that returns an `Optional` result? That way we get all the `orElse` methods for free:

```java
someTarget =  PropertyMapper.get().from(someProps::foo).whenNonNull().toOptionalInstance(someTarget::updateFoo).orElse(someTarget);
```

onobc:
> I wonder if we should have something like `toOptionalInstance(factory)` that returns an `Optional` result? That way we get all the `orElse` methods for free:
> 
> ```java
> someTarget =  PropertyMapper.get().from(someProps::foo).whenNonNull().toOptionalInstance(someTarget::updateFoo).orElse(someTarget);
> ```

Absolutely! This gets us out of the "picking what to do in the case it is filtered value" business. No need to pass in default etc..  I will amend  the PR shortly.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/PropertyMapper.java
```diff
@@ -308,6 +310,21 @@ public <R> R toInstance(Function<T, R> factory) {
 			return factory.apply(value);
 		}
 
+		/**
+		 * Complete the mapping by creating a new instance from the non-filtered value.
+		 * @param <R> the resulting type
+		 * @param factory the factory used to create the instance
+		 * @return the created instance or empty when the value is filtered
+		 */
+		public <R> Optional<R> toOptionalInstance(Function<T, R> factory) {
```
wilkinsona, line 319: I don't think this description is quite right, specifically "from the non-filtered value" seems to be contradicted by "or empty when the value is filtered". I think the opening sentence should end with "from the filtered value".

Given the use of the term `filtered` here and in the javadoc of various other methods, I think `toFilteredInstance` would be a better name for the method.

onobc: @wilkinsona I too thought the terms were off but in the PropertyMapper the term "filtered" means that it did not pass all of the predicates - unfiltered means it passed all of the predicates. Is that correct @philwebb ?

wilkinsona: On second reading, you're right I think. Perhaps just "Complete the mapping by creating a new instance" would be better?

That leaves me unsure about the method name. I'm still not keen on `toOptionalInstance`. Maybe `toFilteredInstance` is still ok. It's an instance created from a value that has gone through filtering. As a result of that filtering it may be `null` or it may be the value 🤷‍♂️

onobc: Sounds good on the doc update. 

> That leaves me unsure about the method name. I'm still not keen on toOptionalInstance

For consistency we would then need probably need to update the existing `toInstance`. Going a little further, would we also add "filtered" and "non-filtered" to the other APIs that only "do their thing" (call, consume, etc) when they are non-filtered?

snicoll: I don't like `toOptionalInstance` either. I wonder if we shouldn't keep `toInstance` in both cases but with a boolean flag that express what should happen if the predicate does not match.

I should say that, out-of-context, the "non-filtered" bit is a bit hard to grasp.

onobc: @snicoll I pushed a commit w/ the suggestion. I am ok w/ either approach. 

> I should say that, out-of-context, the "non-filtered" bit is a bit hard to grasp.

Agreed

philwebb:
Thanks for all the reviews. After a bit more discussion with @snicoll we decided to circle back to almost the original proposal. Rather than adding a `toInstance` method, I've added another `to` method and made it accept a `BiFunction`. This means you can write the immutable code like this:

```java
SenderOptions<K, V> senderOptions = SenderOptions.create(this.kafkaProperties.buildReactorProducerProperties());
KafkaProperties.Reactor.Sender senderProperties = this.kafkaProperties.getReactor().getSender();
senderOptions = map.from(senderProperties::getCloseTimeout).to(senderOptions, SenderOptions::closeTimeout);
senderOptions = map.from(senderProperties::getMaxInFlight).to(senderOptions, SenderOptions::maxInFlight);
senderOptions = map.from(senderProperties::isStopOnError).to(senderOptions, SenderOptions::stopOnError);
return senderOptions;
```

Issue #31232, opened 2022-06-01
Clarify that MVC components provided through WebMvcRegistrations are subject to subsequent processing and configuration by MVC


vpavic:
Previously, I was able to disable Spring MVC's trailing slash matching using something like this:

```java
@Configuration(proxyBeanMethods = false)
class WebMvcConfiguration {

    @Bean
    WebMvcRegistrations webMvcRegistrations() {
        return new WebMvcRegistrations() {

            @Override
            public RequestMappingHandlerMapping getRequestMappingHandlerMapping() {
                RequestMappingHandlerMapping handlerMapping = new RequestMappingHandlerMapping();
                handlerMapping.setUseTrailingSlashMatch(false);
                return handlerMapping;
            }

         };
    }

}
```

This doesn't work any more since `2.6.0` and the change of behavior appears to be related to introduction of `PathPatternParser` as the default path matching strategy.

I've put together a reproduced in https://github.com/vpavic/mcve-spring-mvc-trailing-slash.

By default, trailing slash matching is enabled and `curl -sv -u user:password http://localhost:8080/resources/` (note the trailing slash) returns `200` response. There are two profiles that disable trailing slash matching - `configurer` and `registrations`. The former uses `WebMvcConfigurer` based approach and this works (meaning, cURL request returns `404`), while the latter uses `WebMvcRegistrations` based approach (as in the snippet above) and this doesn't work (returns `200`).

Downgrading Spring Boot to `2.5.x` makes the `registrations` profile work as expected.

I'm reporting this here for starters because what's failing is a Spring Boot specific configuration mechanism (`WebMvcRegistrations`) even though it seems to me that the fix might very well be needed in the Framework.

wilkinsona:
Thanks for the report and, in particular, the MCVE. 

`WebMvcRegistrations` plugs into Spring MVC via `WebMvcConfigurationSupport.createRequestMappingHandlerMapping()`. Its javadoc states that it is a "protected method for plugging in a custom subclass of `RequestMappingHandlerMapping`. Based on this, I would argue that your `WebMvcRegistrations` instance isn't using the plug point as intended. The javadoc of `WebMvcRegistrations.getRequestMappingHandlerMapping` also alludes to the current behaviour as it says that the method should "return the custom `RequestMappingHandlerMapping` that should be used and processed by the MVC configuration". The key part here is "and processed by the MVC configuration" as it's this subsequent processing that is causing the trailing slash configuration to be lost.

[...]

PR #31327, opened 2022-06-10
Add property to allow configuring Spring MVC trailing slash match

vpavic:
See #31232.


vpavic:
With the outcome of https://github.com/spring-projects/spring-framework/issues/28552, this PR probably makes sense only for `2.x` where it could provide an easier option to align with the behavior (no trailing slash matching) that will become the default in Spring Framework 6.

wilkinsona:
It would have been nice to provide a migration path for this by flipping the default in Boot 2.x and providing a property to restore the old behaviour for those that really need it. Unfortunately, we've missed our window opportunity as we currently expect 2.7 to be the last minor in the 2.x line. As such, I think it makes sense to close this. Boot 3 users who want to re-enable the deprecated support for trailing slash matching can do so using a configurer bean.

PR #31400, opened 2022-06-16
Upgrade to Bouncycastle 1.71

larsgrefer:
BouncyCastle changed its artifact name suffix from `jdk15on` to `jdk18on` in the latest releases, as it now requires at least Java 1.8


wilkinsona:
Thank you, @larsgrefer.

PR #31444, opened 2022-06-17
Use Runtime.version() instead of reflection

dreis2211:
Hi,

this PR removes some reflection code in order to get the runtime version and replaces it with `Runtime.version()`.

Have a good weekend,
Christoph


snicoll:
Thanks again Christoph, enjoy the weekend as well.

PR #31449, opened 2022-06-17
Fix checkstyle violation

dreis2211:


wilkinsona:
Thanks very much, @dreis2211.

PR #31454, opened 2022-06-19
Add Javadoc since to AbstractApplicationContextRunner.prepare()

izeye:
This PR adds Javadoc `@since` tag to `AbstractApplicationContextRunner.prepare()`.


snicoll:
Thanks @izeye. I was about to mention that we use the `x.y` convention here and I polished the code that way. Only to realize after I pushed it that it was the other way around 🙄.

PR #31458, opened 2022-06-19
Fix kotlin example

ningenMe:
I found the example source code in the docs that has a difference between java and kotlin.

The docs is [here]( https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.spring-application.application-exit ) , the java source code is [here]( https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/features/springapplication/applicationexit/MyApplication.java#L33 ) , and 
the kotlin source code is [here](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/main/kotlin/org/springframework/boot/docs/features/springapplication/applicationexit/MyApplication.kt#L35).

Maybe, the kotlin example is wrong, so I fixed it.


snicoll:
@ningenMe thank you for making your first contribution to Spring Boot.

PR #31475, opened 2022-06-20
Use pattern matching for instanceof where appropriate

### LINKED ISSUE #28181: Use pattern matching for instanceof where appropriate

dreis2211:
Hi,

this PR uses pattern matching for `instanceof` where appropriate and closes #28181.

I explicitly didn't make use of it where its introduction would generate new rawtype warnings. And while I could add `@SuppressWarnings("rawtypes")` everywhere this didn't feel right. Typical example of this are Maps/Lists etc.:
```
if (item instanceof Map) {
	// do something (Map<String, Object>) item);
}
else if (item instanceof List) {
	// do something (List<Object>) item);
}
```
Let me know if I should make use of pattern matching in those cases as well and suppress the rawtype warnings.

Local tests were green, but I had no Docker running to not spend hours on waiting - so there might be some tests popping up in CI that I'll of course fix in case there is anything.

Cheers,
Christoph

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/AsciidoctorConventions.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/AsciidoctorConventions.java
@@ -120,10 +120,10 @@ class AsciidoctorConventions {
 		configureOptions(asciidoctorTask);
 		asciidoctorTask.baseDirFollowsSourceDir();
 		createSyncDocumentationSourceTask(project, asciidoctorTask);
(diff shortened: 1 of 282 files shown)


dreis2211:
Give me a minute @wilkinsona - I noticed a couple of missing opportunities to use pattern matching.

PR #31480, opened 2022-06-21
Add log level guards in org.springframework.boot.autoconfigure.graphql

izeye:
This PR adds log level guards in `org.springframework.boot.autoconfigure.graphql` package.


philwebb:
Thanks for the PR @izeye, but I think we can get away without most of these. The `LogMessage` class is designed to be lightweight enough that it can be used instead of guards. It only applies the formatting if the log message is evaluated. I've fixed the auto-configuration one in https://github.com/spring-projects/spring-boot/commit/4c0a4cb6692253432c3ce59840c13a569cfbe207.

Issue #28357, opened 2021-10-15
JarFileWrapper may cause many FinalReferences causing GC pressure


philwebb:
Forward port of issue #28356 to 2.6.x.

PR #31481, opened 2022-06-21
Normalize DataSize inputs via removing all whitespaces

kevin0x90:
Trim all whitespaces before parsing a given DataSize string so that the parsing is more robust

This relates to the following issue in Spring Framework: https://github.com/spring-projects/spring-framework/issues/28643 since one of the maintainers recommended a fix in StringToDataSizeConverter in SpringBoot.

Closes gh-28357


kevin0x90:
Thanks for the feedback what do you think about applying the fix in spring framework? In the original issue there was a discussion about a general solution that would include Converters.

kevin0x90:
@wilkinsona ok then i would apply the fix directly in the DataSize class at first and create a separate PR in the spring framework and leave the Converters untouched at the moment.

So from my perspective this PR could be closed then.

PR #31494, opened 2022-06-22
Use InputStream.transferTo() when possible

### LINKED ISSUE #28221: Use InputStream.transferTo when possible
Added in JDK 9, might be a nice alternative to `StreamUtils`. Possibly also worth raising with Framework.

dreis2211:
Hi,

this PR closes #28221 by using `InputStream.transferTo` where possible.

I'm not super convinced of that PR to be honest because for consistency reasons and to make sure it works as before, we need to call `flush` on the `OutputStream`. While this wouldn't be necessary for `FileOutputStream` strictly speaking, because its implementation does nothing, I added it everywhere.

Let me know what you think.
Cheers,
Christoph

### DIFF
--- a/spring-boot-project/spring-boot-cli/src/intTest/java/org/springframework/boot/cli/infrastructure/CommandLineInvoker.java
+++ b/spring-boot-project/spring-boot-cli/src/intTest/java/org/springframework/boot/cli/infrastructure/CommandLineInvoker.java
@@ -95,6 +95,7 @@ public final class CommandLineInvoker {
 						file.getParentFile().mkdirs();
 						try (FileOutputStream output = new FileOutputStream(file)) {
 							input.transferTo(output);
+							output.flush();
 							if (entry.getName().endsWith("/bin/spring")) {
 								file.setExecutable(true);
 							}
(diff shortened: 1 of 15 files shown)


mdeinum:
Wouldn't it make more sense to modify the [`StreamUtils.copy`](NoHandlerFoundException) method instead? And update the `BUFFER_SIZE` to the same default as used inside the JDK (which is `8192` double of what `StreamUtils` is using)?

philwebb:
Having looked at the changes I'm not sure we should proceed. I think when I raised #28221 I thought that it would be a 1-1 replacement, with the possibility that Framework might deprecate their version. The flush method makes things a bit more involved and I think I prefer keeping things as they are.

Flagging to see what others think.

scottfrederick:
I agree with Phil, the `.flush()` calls with every `.transferTo()` is a bad smell, and looks more cluttered than just using `StreamUtils.copy()`.

dreis2211:
@philwebb It wasn't wasted effort - the point was to demonstrate how it looks. Don't worry :)

PR #31520, opened 2022-06-24
Update StartupInfoLogger.java

shenjianeng:
change access level


bclozel:
Could you explain why this change is needed? What is the use case behind this?
We don't expose contracts unless necessary.

shenjianeng:
这是来自邮箱的自动回复邮件，我已收到邮件。-------------------------------------------------

shenjianeng:
it is a utility class.

I want to use it in my project.

thanks.

wilkinsona:
Thanks for the suggestion but we do not view it as a utility class. Making it public will limit our ability to change it in the future and we do not want that restriction. If you want similar functionality in your own application, you  are free to copy the code and adapt it to meet your needs.

PR #31527, opened 2022-06-24
Use switch expressions where appropriate

### LINKED ISSUE #28180: Use switch expressions where appropriate

dreis2211:
Hi,

this PR fixes #28180 by applying switch expressions where appropriate (there are a couple of places where they didn't make things more concise imho).

The nice side-benefit is that with switch expressions we get a compilation failure now if a new enum case is added instead of needing to rely on throwing exceptions during runtime.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

Issue #18286, opened 2019-09-19
Support multipart uploads with @RestControllerEndpoint when the management server is running on a separate port


sponomarev72:
Hello,

i'm trying to implement custom `@RestControllerEndpoint`  with a method that accepts `MultipartFile` as parameter. Everything works when actuator endpoints are in the same context as main application (port 8080.) After I set up dedicated context for actuator end-points with `management.server.port=9000` I've got this exception:

```
2019-09-19 13:32:49.196 ERROR 21724 --- [nio-9000-exec-6] o.a.c.c.C.[.[.[/].[dispatcherServlet]    : Servlet.service() for servlet [dispatcherServlet] in context with path [] threw exception [Request processing failed; nested exception is org.springframework.web.multipart.MultipartException: Failed to parse multipart servlet request; nested exception is java.lang.IllegalStateException: Unable to process parts as no multi-part configuration has been provided] with root cause

java.lang.IllegalStateException: Unable to process parts as no multi-part configuration has been provided
	at org.apache.catalina.connector.Request.parseParts(Request.java:2802) ~[tomcat-embed-core-9.0.24.jar:9.0.24]
	at org.apache.catalina.connector.Request.getParts(Request.java:2770) ~[tomcat-embed-core-9.0.24.jar:9.0.24]
	at org.apache.catalina.connector.RequestFacade.getParts(RequestFacade.java:1098) ~[tomcat-embed-core-9.0.24.jar:9.0.24]
	at org.springframework.web.multipart.support.StandardMultipartHttpServletRequest.parseRequest(StandardMultipartHttpServletRequest.java:94) ~[spring-web-5.1.9.RELEASE.jar:5.1.9.RELEASE]
	at org.springframework.web.multipart.support.StandardMultipartHttpServletRequest.<init>(StandardMultipartHttpServletRequest.java:87) ~[spring-web-5.1.9.RELEASE.jar:5.1.9.RELEASE]
	at org.springframework.web.multipart.support.StandardServletMultipartResolver.resolveMultipart(StandardServletMultipartResolver.java:87) ~[spring-web-5.1.9.RELEASE.jar:5.1.9.RELEASE]
	at org.springframework.web.servlet.DispatcherServlet.checkMultipart(DispatcherServlet.java:1176) ~[spring-webmvc-5.1.9.RELEASE.jar:5.1.9.RELEASE]
[...]

PR #31554, opened 2022-06-30
Support multipart uploads with @RestControllerEndpoint when the management server is running on a separate port

terminux:
Support multipart uploads when Spring MVC uses a different management port

Closes gh-18286


wxl88888888:
this is a good question

bclozel:
Sorry this got overlooked.
We are considering deprecating such annotations entirely in #31768 so we don't expect to expand the footprint there in all cases. Thanks for your contribution!

PR #31561, opened 2022-06-30
Use Objects.requireNonNullElse when suitable

matiasdsalerno:
Use Objects.requireNonNullElse instead of if statements.

### DIFF
--- a/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/app/SpringApplicationLauncher.java
+++ b/spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/app/SpringApplicationLauncher.java
@@ -20,6 +20,7 @@ import java.lang.reflect.Constructor;
 import java.lang.reflect.Method;
 import java.util.HashMap;
 import java.util.Map;
+import java.util.Objects;
 
 /**
  * A launcher for {@code SpringApplication} or a {@code SpringApplication} subclass. The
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-cli/src/main/java/org/springframework/boot/cli/app/SpringApplicationLauncher.java
```diff
@@ -69,14 +70,11 @@ public Object launch(Class<?>[] sources, String[] args) throws Exception {
...
 	private String getSpringApplicationClassName() {
-		String className = System.getProperty("spring.application.class.name");
-		if (className == null) {
-			className = getEnvironmentVariable("SPRING_APPLICATION_CLASS_NAME");
-		}
-		if (className == null) {
-			className = DEFAULT_SPRING_APPLICATION_CLASS;
-		}
-		return className;
+		return Objects.requireNonNullElse(amount,
```
maciejwalkowiak, line 73: I am not a maintainer so feel free to ignore my comment.

It's questionable if using `Objects.requireNonNullElse` improves anything here - personally I don't find it more readable, rather the opposite.

But even bigger problem is that the changed code does not compile (what is `amount`?) and does not do what the original code did.

wilkinsona: I agree. I actually find this harder to read.

wilkinsona:
Thanks for the suggestion, @matiasdsalerno, but I think the code's more readable as it is currently written.

PR #31566, opened 2022-07-03
Trailing whitespace in the value of a property is hard to identify in failure analysis descriptions

### LINKED ISSUE #31407: Trailing whitespace in the value of a property is hard to identify in failure analysis descriptions
I think we should quote or somehow delimit the value that's output in the failure analysis. It looks like this at the moment:

```
Property: spring.servlet.multipart.file-size-threshold
Value: 2KB 
Origin: class path resource [application.properties]:24:46
Reason: failed to convert java.lang.String to org.springframework.util.unit.DataSize
```

I think something like this would be better:

```
Property: spring.servlet.multipart.file-size-threshold
Value: "2KB "
Origin: class path resource [application.properties]:24:46
Reason: failed to convert java.lang.String to org.springframework.util.unit.DataSize
```

castanhocorreia:
Resolve #31407, by adding quotes in the value of a property, to avoid confusion with trailing whitespaces
Also, add `.tool-versions` from [asdf](https://github.com/asdf-vm/asdf) version manager to .gitgnore

Before:
```
Property: spring.servlet.multipart.file-size-threshold
Value: 2KB 
Origin: class path resource [application.properties]:24:46
Reason: failed to convert java.lang.String to org.springframework.util.unit.DataSize
```
After:
```
Property: spring.servlet.multipart.file-size-threshold
Value: "2KB "
Origin: class path resource [application.properties]:24:46
Reason: failed to convert java.lang.String to org.springframework.util.unit.DataSize
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,6 +18,7 @@
 .recommenders
 .settings
 .springBeans
+.tool-versions
 .vscode
 /code
 MANIFEST.MF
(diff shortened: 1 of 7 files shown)


wilkinsona:
Thanks for the PR, @castanhocorreia, but as indicated by the assignment of the issue, @TheoCaldas is already working on this and we don't want their time to be wasted.

castanhocorreia:
Oh, no problem! Sorry for that.

Issue #31573, opened 2022-07-05
Refactor PublishingCustomizer(GodClass)


MatheusPK:
My team used PMD Software on this project for our college project and we found some interesting refactoring points. One of them is the presence of a God Class. The class in the ocasion has multiple responsibilitys(whats infringe the S - Single Responsibility rule of SOLID principles) that can be shifted to another class and be reused in other contexs in the project.

The God Class is  "PublishingCustomizer " and our refactoring suggestion is moving this node methods to another class called NodeUtils

wilkinsona:
Thanks for the suggestion. I disagree with PMD here. `PublishingCustomizer` as a single responsibility which is to customize the `MavenPublication`, altering the contents of its pom. At under 300 lines of code I think it's size if pretty manageable.

PR #31574, opened 2022-07-05
Created NodeUtil helper class - refactored Publishing Customizer

MatheusPK:
#31573 Possible class refactoring


wilkinsona:
Thanks for the suggestion, but for the reasons stated in [this comment](https://github.com/spring-projects/spring-boot/issues/31573#issuecomment-1174717213) this isn't a change that we want to make. Furthermore, separating the code out like this without gaining any reuse arguably makes it harder to read for no benefit.

Issue #30438, opened 2022-03-25
Break build if entries in additional-spring-configuration-metadata.json are not sorted


mhalbritter:
I sorted the properties by name in #30434.

We should add something to the `ConfigurationPropertiesPlugin` which checks if they are sorted and if not, break the build.

MatheusPK:
Hi, everyone! My team would like to tackle this one. is it ok?

mhalbritter:
Hi! Yes, go ahead! I'll assign the issue to you.

wilkinsona:
Closed by https://github.com/spring-projects/spring-boot/commit/a1443d1cdc7faa85b47507452139c17940b42f30.

PR #31575, opened 2022-07-05
Update ConfigurationPropertiesPlugin.java

MatheusPK:
#30438

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/context/properties/ConfigurationPropertiesPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/context/properties/ConfigurationPropertiesPlugin.java
@@ -16,17 +16,28 @@
 
 package org.springframework.boot.build.context.properties;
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal. Unfortunately, I don't think this is the right way to approach this.

As proposed here, the ordering check will be performed once each project has been evaluated. In a Gradle build, work like this should not be performed until the build's [execution phase](https://docs.gradle.org/current/userguide/build_lifecycle.html#sec:build_phases). To occur in this phase, the check should be implemented as a task. It should perhaps be a [`SourceTask`](https://docs.gradle.org/current/javadoc/org/gradle/api/tasks/SourceTask.html) subclass that's configured to include the `META-INF/spring-configuration-metadata.json` file. Implemented as a task with appropriate inputs and outputs, the ordering check will then be performed in the correct phase and its execution can be avoided through Gradle's up-to-date checking when the file's contents haven't changed.

PR #31576, opened 2022-07-05
Add a configuration property for the trusted proxies of Tomcat's Remote IP valve

### LINKED ISSUE #31542: Add a configuration property for the trusted proxies of Tomcat's Remote IP valve
Tomcat's _RemoteIpValve_ supports both _internal proxies_ as well as _trusted proxies_.  _Internal proxies_ default to the private IP spaces, but _trusted proxies_ defaults to nothing.

_Internal proxies_ can be customized by setting _server.tomcat.remoteip.internal-proxies_ which is later honored by _TomcatWebServerFactoryCustomizer_, but _trusted proxies_ is not honored by it.

When using third party proxies with public IPs (e.g. Cloudflare Access), it is desirable to include their IPs to be ignored by the _RemoteIpValve_.  To do this through configuration, the only option is to _replace_ the _internal proxies_ pattern list since the _trusted proxies_ list is not exposed to be configurable.

My suggestion is to also expose _trusted proxies_ for configuration.

aooohan:
This PR fix #31542 .

See gh-31542

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -966,6 +966,11 @@ public class ServerProperties {
 					+ "0:0:0:0:0:0:0:1|::1";
 
 			/**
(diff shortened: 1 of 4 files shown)


snicoll:
@aooohan thank you for making your first contribution to Spring Boot.

PR #31582, opened 2022-07-07
Start building against Spring Data 2022.0.0-M5 snapshots

mp911de:
Upgrade to Spring Data `2022.0.0` snapshots and upgrade to Jedis 4.2.3 along with registering `ManagedTypes` bean to enable AOT post-processing hooks of entities and adopt to package changes in Spring Data Elasticsearch.


wilkinsona:
Thanks, @mp911de.

Issue #9263, opened 2017-05-18
Spring not throwing NoHandlerFoundException anymore


Harown:
After we upgraded from Spring Boot 1.4.0 to 1.5.3 our web application does not throw the NoHandlerFoundException anymore. Configuration looks like the following:

in application.properties:
```
spring.mvc.throw-exception-if-no-handler-found=true
spring.resources.add-mappings=false
```
the corresponding exception controller:

```
@ControllerAdvice
@EnableWebMvc
public class ExceptionController 
{
       // works
       @ExceptionHandler(AccessDeniedException.class)
	public String handleAccessDeniedException(AccessDeniedException ex, HttpServletRequest request) {
		return "403";
	}

	// doesn't work anymore
	@ExceptionHandler(NoHandlerFoundException.class)
	public String handleNotFoundError(NoHandlerFoundException ex, HttpServletRequest request) {
		return "404";
	}
}
```
We had no problems with Spring 1.4.0 but it stopped working in 1.5.3. Why is the exception not thrown anymore?

wilkinsona:
Did you have `@EnableWebMvc` with both 1.4 and 1.5? I would expect it to turn off the auto-configuration of Spring MVC making `spring.mvc.throw-exception-if-no-handler-found=true` have no effect

Harown:
Unfortunately, omitting `@EnableWebMvc` has no effect.

wilkinsona:
Something that is, I think, the same as you are doing works for me with both 1.4.6 and 1.5.3. Can you please provide a small sample that reproduces the problem you're seeing? Something that works with 1.4.0 but fails with 1.5.3 would be ideal.

Harown:
While preparing a sample application we've found the culprit:
```
@Configuration
public class WebMvcConfig extends WebMvcConfigurerAdapter {
	private static final String[] CLASSPATH_RESOURCE_LOCATIONS = {
	        "classpath:/META-INF/resources/", "classpath:/resources/",
	        "classpath:/static/", "classpath:/public/" };
	
	@Override
	public void addResourceHandlers(ResourceHandlerRegistry registry) {
	    registry.addResourceHandler("/**").addResourceLocations(CLASSPATH_RESOURCE_LOCATIONS);
	}
}
```

Replacing the URl pattern with the subfolder containing the static resources `spring.mvc.throw-exception-if-no-handler-found=true` have the desired effect. Thank you.

sreekanthsnair:
[...]

PR #31660, opened 2022-07-08
Document additional configuration that is required for spring.mvc.throw-exception-if-no-handler-found=true to be effective

michaldo:
`spring.mvc.throw-exception-if-no-handler-found=true` requires additional configuration to be effective because default `+spring.mvc.static-path-pattern=/**+` collides it.
This PR explains correlation, which is confusing.
The confusion was reported in https://github.com/spring-projects/spring-boot/issues/9263 and others


wilkinsona:
See https://gitter.im/spring-projects/spring-boot?at=62ba1378568c2c30d30790af for a little bit of background.

wilkinsona:
@michaldo Thanks for your patience and for making your first contribution to Spring Boot.

PR #31663, opened 2022-07-09
Upgrade to Hibernate 6

odrotbohm:
Initial draft of a Hibernate 6 upgrade for Spring Boot. The first commit upgrades to Spring Data JPA and Envers versions that use Hibernate 6 already. This commit can be removed from the PR once spring-projects/spring-data-jpa#2423 is merged. We didn't merge that yet, as it is likely to break Boot unless it also upgrades to H6.

The upgrade removes support for the `USE_NEW_ID_GENERATOR_MAPPINGS` settings as H6 had that removed, unfortunately without prior deprecation in 5.6. Does it make sense to deprecate the Boot setting triggering that setting in its 2.7 branch then?

I also had to adapt the sequence definitions for Liquibase and Flyway, as the names H6 expects are different (entity specific) now.


wilkinsona:
Whoops. I forgot that we had a PR for the changes. They've been committed on `main` as part of https://github.com/spring-projects/spring-boot/issues/31674.

PR #31671, opened 2022-07-11
AOT support for Actuator

mhalbritter:
This branch has some rough edges and is currently missing tests and I have to squash the commits.

I've tested the actuator with the endpoints on both WebMVC and Webflux.

* `/conditions` is completely empty. I think this must be because all the conditions are
  evaluated at build time. Double check!
* Couldn't test `/integrationgraph` - i've added hints, which should be enough as soon as spring-integration is AOT ready
* `/heapdump` and `/threaddump` don't work in native-image. Both don't throw exceptions, they return 503 Service Unavailable to the caller
* JMX is not supported on native-image

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/aot/ActuatorAnnotationsRuntimeHintsRegistrar.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/aot/ActuatorAnnotationsRuntimeHintsRegistrar.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.actuate.aot;
+
+import java.util.stream.Stream;
+
+import org.springframework.aot.hint.RuntimeHints;
+import org.springframework.aot.hint.RuntimeHintsRegistrar;
+import org.springframework.aot.hint.support.RuntimeHintsUtils;
+import org.springframework.boot.actuate.endpoint.annotation.DeleteOperation;
+import org.springframework.boot.actuate.endpoint.annotation.Endpoint;
+import org.springframework.boot.actuate.endpoint.annotation.EndpointExtension;
+import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
+import org.springframework.boot.actuate.endpoint.annotation.WriteOperation;
+import org.springframework.core.annotation.SynthesizedAnnotation;
+
+/**
+ * Registrar which registers the annotations needed for actuator support.
+ *
+ * @author Moritz Halbritter
+ * @since 3.0.0
+ */
+public class ActuatorAnnotationsRuntimeHintsRegistrar implements RuntimeHintsRegistrar {
+
+	@Override
+	public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
+		Stream.of(Endpoint.class, ReadOperation.class, WriteOperation.class, DeleteOperation.class,
+				EndpointExtension.class)
+				.forEach((annotationType) -> RuntimeHintsUtils.registerAnnotation(hints, annotationType));
+		// TODO: See https://github.com/spring-projects/spring-framework/issues/28767
+		Stream.of(Endpoint.class, EndpointExtension.class).forEach(
+				(annotationType) -> hints.proxies().registerJdkProxy(annotationType, SynthesizedAnnotation.class));
+	}
+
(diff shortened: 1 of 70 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/aot/ActuatorAnnotationsRuntimeHintsRegistrar.java
```diff
@@ -0,0 +1,49 @@
...
+import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
+import org.springframework.boot.actuate.endpoint.annotation.WriteOperation;
+import org.springframework.core.annotation.SynthesizedAnnotation;
+
+/**
+ * Registrar which registers the annotations needed for actuator support.
+ *
+ * @author Moritz Halbritter
+ */
+// TODO MH: Currently registered in aot.factories, is this the best way to do it?
```
mhalbritter, line 36: Currently registered in `aot.factories`, so it should only run if the JAR is on the classpath. Or is there a better way to hook into?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/info/BuildProperties.java
```diff
@@ -100,4 +106,14 @@ private static void coerceDate(Properties properties, String key) {
 		}
 	}
 
+	static class BuildPropertiesRuntimeHints implements RuntimeHintsRegistrar {
+
+		@Override
+		public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
+			// TODO MH: This path is configurable in ProjectInfoProperties!
+			hints.resources().registerPattern("META-INF/build-info.properties");
```
mhalbritter, line 114: How to make this dynamic?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/info/GitProperties.java
```diff
@@ -126,4 +132,14 @@ private static Long parseEpochSecond(String s) {
 		}
 	}
 
+	static class GitPropertiesRuntimeHints implements RuntimeHintsRegistrar {
+
+		@Override
+		public void registerHints(RuntimeHints hints, ClassLoader classLoader) {
+			// TODO MH: This path is configurable in ProjectInfoProperties!
```
mhalbritter, line 139: How to make this dynamic?

snicoll: Usually we register hints for the default location. This is the same for static resources & co.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/management/ThreadDumpEndpoint.java
```diff
@@ -48,7 +49,41 @@ public String textThreadDump() {
 	}
 
 	private <T> T getFormattedThreadDump(Function<ThreadInfo[], T> formatter) {
-		return formatter.apply(ManagementFactory.getThreadMXBean().dumpAllThreads(true, true));
+		ThreadDumper threadDumper = createThreadDumper();
+		return formatter.apply(threadDumper.dumpAllThreads());
+	}
+
+	private ThreadDumper createThreadDumper() {
+		if (AotDetector.useGeneratedArtifacts()) {
```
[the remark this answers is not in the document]
snicoll: That would disable the endpoint on the JVM as well with AOT optimizations. That doesn't sound right to me. `NativeDetector` maybe?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/aot/ActuatorAnnotationsRuntimeHintsRegistrar.java
snicoll: You can import the hints when the auto-configuration for the actuator kicks in, see `@ImportRuntimeHints`.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/management/ThreadDumpEndpoint.java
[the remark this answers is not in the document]
mhalbritter: Yeah, you're right!

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/info/GitProperties.java
mhalbritter: So there's no way to get the `Environment` or properties binding in the AOT phase, right? We just register the default locations and have to add something to the docs that users have to provide their own resource hints if they change that location?

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/aot/ActuatorAnnotationsRuntimeHintsRegistrar.java
mhalbritter: Thanks, i now import them in `EndpointAutoConfiguration`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/info/GitProperties.java
mhalbritter: After a discussion with Stéphane we decided to just register the static locations, so this is fine.

Issue #29640, opened 2022-02-03
HealthContributor beans managed by a CompositeHealthContributor are recreated on each call


schuch:
Spring Boot (tested with 2.6.3) aggregates multiple instances of the same health indicator

- Every aggregated health indicator gets recreated every time the health endpoint is pulled.
- Non aggregated health indicator instances are reused.

This lifecycle inconsistency seems unintended to me.

Demo: https://github.com/schuch/actuator-test

Fire up the application and call http://localhost:8080/actuator/health multiple times, then view the log.

This will be logged for every request:
```
created TestHealthIndicator for 1
created TestHealthIndicator for 2
```

This is logged only once at startup:
```
created SingleHealthIndicator
```

Sample Log:
```
2022-02-03 16:14:22.972  INFO 27776 --- [           main] w.s.c.ServletWebServerApplicationContext : Root WebApplicationContext: initialization completed in 1857 ms
2022-02-03 16:14:23.174  INFO 27776 --- [           main] com.example.demo.SingleHealthIndicator   : created SingleHealthIndicator
2022-02-03 16:14:23.197  INFO 27776 --- [           main] com.example.demo.TestHealthIndicator     : created TestHealthIndicator for 1
2022-02-03 16:14:23.199  INFO 27776 --- [           main] com.example.demo.TestHealthIndicator     : created TestHealthIndicator for 2
2022-02-03 16:14:23.801  INFO 27776 --- [           main] o.s.b.a.e.web.EndpointLinksResolver      : Exposing 1 endpoint(s) beneath base path '/actuator'
2022-02-03 16:14:23.849  INFO 27776 --- [           main] o.s.b.w.embedded.tomcat.TomcatWebServer  : Tomcat started on port(s): 8080 (http) with context path ''
2022-02-03 16:14:23.868  INFO 27776 --- [           main] com.example.demo.DemoApplication         : Started DemoApplication in 3.43 seconds (JVM running for 5.703)
2022-02-03 16:14:25.243  INFO 27776 --- [1)-10.180.42.16] com.example.demo.TestHealthIndicator     : created TestHealthIndicator for 1
2022-02-03 16:14:25.244  INFO 27776 --- [1)-10.180.42.16] com.example.demo.TestHealthIndicator     : created TestHealthIndicator for 2
[...]

PR #31676, opened 2022-07-12
HealthContributor beans managed by a CompositeHealthContributor are recreated on each call

terminux:
During the construction of the `NamedContributorsMapAdapter`, each value adapter is called only once.

Closes gh-29640


philwebb:
I think my original concern was that existing users might be relying on the dynamic adapter code. I've pushed the PR with a polish commit, but I think I'll change it a little more so that the adaption is done on first access rather than during construction.

philwebb:
Actually, on second thoughts, I think things are fine. We previously called the `valueAdapter` during construction to validate the result.

Thanks for the PR @terminux

PR #31679, opened 2022-07-12
Add config property for KafkaAdmin modifyTopicConfigs

m-kay:
Adds a new configuration property `spring.kafka.admin.modify-topic-configs` to facilitate enabling of this [recently added functionality](https://github.com/spring-projects/spring-kafka/issues/1482)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
@@ -142,6 +142,7 @@ public class KafkaAutoConfiguration {
 	public KafkaAdmin kafkaAdmin() {
 		KafkaAdmin kafkaAdmin = new KafkaAdmin(this.properties.buildAdminProperties());
 		kafkaAdmin.setFatalIfBrokerNotAvailable(this.properties.getAdmin().isFailFast());
+		kafkaAdmin.setModifyTopicConfigs(this.properties.getAdmin().isModifyTopicConfigs());
 		return kafkaAdmin;
 	}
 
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the proposal. We can consider this for 3.0.x but it's too late now for 2.7.x, particular given that it's only a few lines of code to define your own `KafkaAdmin` bean:

```java
public KafkaAdmin kafkaAdmin(KafkaProperties properties) {
	KafkaAdmin kafkaAdmin = new KafkaAdmin(properties.buildAdminProperties());
	kafkaAdmin.setFatalIfBrokerNotAvailable(properties.getAdmin().isFailFast());
	kafkaAdmin.setModifyTopicConfigs(true);
	return kafkaAdmin;
}
```

@garyrussell do you think it makes sense to surface a configuration property for this setting?

m-kay:
@wilkinsona The reason I would have liked it as configuration property is to make it easily configurable in a k8s environment.

wilkinsona:
@m-kay Until Boot has a property, you can externalise it using your own property. You could even name that property `spring.kafka.admin.modify-topic-configs` if you were prepared to deal with a potential clash in the future.

m-kay:
Yes sure I can do that. I just thought I could also help others with this small change here :wink:
Besides that I like to use spring boots auto configuration wherever possible, instead of creating the beans in my own config, to benefit from new configurations added in future releases.

garyrussell:
Yes, it makes sense to add it. Perhaps a simpler work around (without having to add your own `KafkaAdmin` bean) is to set the property in one of the `NewTopic` bean definitions.

```java
	@Bean
	public NewTopic topic(KafkaAdmin admin) {
		admin.setModifyTopicConfigs(true);
		return TopicBuilder.name("topic1")
				.partitions(1)
				.replicas(1)
				.config("retention.ms", "60000")
				.build();
	}
```

garyrussell:
I agree, in general, but in this case it's safe because `KafkaAdmin` is a `SmartInitializingSingleton` so it doesn't do any work until all the beans have been instantiated.

snicoll:
@m-kay thank you for making your first contribution to Spring Boot.

PR #31709, opened 2022-07-13
Add auto-configuration for JdkClientHttpConnector

### LINKED ISSUE #31479: Add auto-configuration for JdkClientHttpConnector
Spring Framework 6 introduces `JdkClientHttpConnector`, a `ClientHttpConnector` based on the JDK HttpClient. Would be nice if Spring Boot would support an auto-configuration for it. With this you could use the WebClient out of the box, without Netty, Apache HttpClient, etc.

aooohan:
This PR fix #31479 .

See gh-31479


wilkinsona:
Thanks very much, @aooohan.

PR #31796, opened 2022-07-19
Clean the code.

gavrilovegor519:


snicoll:
I personally find `+=` harder to read and I suspect we'd have more use cases across the codebase so merging this would be inconsistent. Same remark on switch.

mhalbritter:
I agree with @snicoll.

gavrilovegor519:
OK. I'd rather focus on my pet projects, as it's easier for me to focus on my own project than someone else's.

PR #31828, opened 2022-07-21
Take MethodType into account to deduce main application class

quaff:

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
@@ -17,6 +17,7 @@
 package org.springframework.boot;
 
 import java.lang.StackWalker.StackFrame;
+import java.lang.invoke.MethodType;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Arrays;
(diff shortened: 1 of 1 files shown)


philwebb:
See #31701

wilkinsona:
I don't think this is a bug. The code that was contributed is equivalent to what we had before. This suggestion improves things in a way that wasn't possible before.

quaff:
> I don't think this is a bug. The code that was contributed in #31701 is equivalent to what we had before. This suggestion improves things in a way that wasn't possible before.

Yes, It's just an improvement.

wilkinsona:
Thanks very much, @quaff.

snicoll:
This breaks running a Spring Boot application as a native image as follows:

```
Exception in thread "main" java.lang.UnsupportedOperationException
        at java.lang.StackWalker$StackFrame.getMethodType(StackWalker.java:148)
        at org.springframework.boot.SpringApplication.lambda$findMainClass$1(SpringApplication.java:290)
```

We should figure out what to do here. Maybe a fallback although that would change the behavior between JVM and native.

snicoll:
For the reason above, I've reverted this change. We can follow-up in https://github.com/spring-projects/spring-boot/issues/31858

Issue #29654, opened 2022-02-05
Use Kotlin 1.6 for API and language version in reference documentation


philwebb:
`KotlinConventions` currently uses 1.3, we should upgrade.

wilkinsona:
I think we should do this only in the docs, otherwise we’re raising our Kotlin baseline from 1.3 to 1.6. That feels a bit too much for 2.7.

philwebb:
@wilkinsona What's the difference between the version in the conventions and the version [in the `gradle.properties`](https://github.com/spring-projects/spring-boot/blob/1d153fde3782fe2ec2d9c34e6da2fdd84425fce3/gradle.properties#L7). I assumed we were already on 1.6 and had just missed it in the conventions.

wilkinsona:
Our Kotlin set up is a bit like compiling with Java 11 and using `-release 8` to retain compatibility with an earlier version. While we depend on Kotlin 1.6, setting the API and language version to 1.3 produces Kotlin 1.3 compatible byte code.

PR #31849, opened 2022-07-22
Align Kotlin API and language versions

dreis2211:
Hi,

this PR aligns the version handling of Kotlin API and language level by removing the pinning on 1.6 inside the docs project, which got introduced in #29654 to allow for more modern docs while still maintaining a 1.3 baseline overall.

Now that 1.7 is the baseline (see #31391) this imho doesn't serve a purpose anymore and would hold back any modernization of documentation snippets at the moment.

Let me know what you think.

Cheers,
Christoph


wilkinsona:
Well spotted, @dreis2211. Thanks very much.

PR #31863, opened 2022-07-26
Disable auto-detection in tests that start Hazelcast

### LINKED ISSUE #31715: Disable auto-detection in tests that start Hazelcast
Programmatically, it's something like:

```java
getAutoDetectionConfig().setEnabled(false)
```

In YAML:

```yaml
hazelcast:
  network:
    join:
      auto-detection:
        enabled: false
```

And in XML:

```xml
<hazelcast>
    ...
    <network>
        <join>
            <auto-detection enabled="false" />
            ...
        </join>
    </network>
    ...
</hazelcast>
```

When configured correctly, the node should log the following:

```
WARNING: [192.168.178.45]:5701 [dev] [4.2.5] No join method is enabled! Starting standalone.
```

We don't want to see this:

```
INFO: [192.168.178.45]:5701 [dev] [4.2.5] Using Multicast discovery
```

adrianbob:
Updated test configs to disable auto-detection for Hazelcast instances

Closes #31715

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/resources/hazelcast.xml
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/resources/hazelcast.xml
@@ -5,6 +5,7 @@
 	<map name="defaultCache" />
 	<network>
 		<join>
+			<auto-detection enabled="false"/>
 			<tcp-ip enabled="false"/>
 			<multicast enabled="false"/>
 		</join>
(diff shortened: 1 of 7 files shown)


snicoll:
Unfortunately, there's no such thing as `auto-detection` in the Hazelcast 4 xsd. There was one for Hazelcast 3 but it's gone in version 4. The programmatic API still has it so I am not sure what's the expected replacement.

snicoll:
Turns out it was an issue with an earlier XSD in the 4.x generation.

snicoll:
Thanks again, @adrianbob. Interesting that GitHub flagged this PR as a first contribution again...

adrianbob:
> Thanks again, @adrianbob. Interesting that GitHub flagged this PR as a first contribution again...

Thanks! The previous commit on SpringBoot was done with my work email so it did not match the one from my personal account.

PR #31903, opened 2022-07-28
Remove unused code in GitInfoContributor

dalbani:
As stated in the documentation at https://docs.spring.io/spring-boot/docs/current/actuator-api/htmlsingle/#info.retrieving.response-structure.git, the `info` Actuator endpoint doesn't include a `build.time` in the response.
Yet, in `GitInfoContributor::postProcessContent()`, the method looks for a `build.time` property that thus never exists, because it's not retained in the `toSimplePropertySource()` method above.
The line of code in question doesn't break anything but it has no effect either.
That's why I suggest removing it altogether for clarity.


snicoll:
@dalbani that's the output for the simple mode. When it outputs the full data, there is a `build.time` property exposed. The mode is [documented here](https://docs.spring.io/spring-boot/docs/current/reference/html/actuator.html#actuator.endpoints.info.git-commit-information).

dalbani:
Thanks a lot @snicoll for this super fast and helpful feedback 👍

PR #31941, opened 2022-08-01
External configuration documentation uses incorrect placeholder syntax

vilmos:
Property placeholder format is not `"{property}"` but `"${property}"`.


philwebb:
Thanks very much @vilmos for making your first contribution to Spring Boot!

PR #31958, opened 2022-08-02
Add WebServerBeforeStartEvent before web server start

weiwei121723:
Add WebServerBeforeStartEvent before web server start.
It is useful in some business scenarios, we need to do some perception and necessary initialization work before the web server starts.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/context/WebServerBeforeStartEvent.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/context/WebServerBeforeStartEvent.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2020 the original author or authors.
+ * Copyright 2002-2021 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


philwebb:
Thanks for the pull request. Before we consider merging this do any of the existing event provide what you need? For example `ApplicationContextInitializedEvent` or `ApplicationPreparedEvent`?

If we do add a `BeforeServletWebServerInitializedEvent`, I don't think that it should extends `ServletWebServerInitializedEvent` since that even is for when a web server is ready.

weiwei121723:
thanks @philwebb , 
In my project,  I try to use the `ApplicationPreparedEvent ` to do some initialization work, but it seems not to prevent the web server start

image


image


> it should extends ServletWebServerInitializedEvent since that even is for when a web server is ready.

This point is very correct, I will reconsider this question about the event.

Thanks for your replay

philwebb:
@weiwei121723 We discussed this today as a team and decided that we need a stronger use-case before we can consider adding a new event. There are several existing mechanisms that we feel should work and we'd like to prove that they won't before we offer another.

If you feel you have a real use-case that you can't get working with existing mechanisms then please open a new issue and attach a full application that we can run and debug. We're a little confused about how the `ApplicationPreparedEvent` is working on the snippet of code you shared so a full application is really needed.

Regardless, thanks for working on the pull-request.

PR #31997, opened 2022-08-07
Misleading error message when using JarMode Layertools and the source is not an archive

nilshartmann:
If the source file given to the layertools is not an archive file, the error message says that the file cannot be found. That might be confusing, if the file actually exists but "only" has the wrong suffix.
I think splitting the assertion into two separate ones gives more precise error messages.


snicoll:
@nilshartmann thank you for making your first contribution to Spring Boot.

Issue #32028, opened 2022-08-09
Add native-image support for Mustache on WebFlux


mhalbritter:
The `mustache-webflux` sample fails because there is a constructor hint on `MustacheView` missing:

```
ion: Failed to instantiate [org.springframework.boot.web.reactive.result.view.MustacheView]: No default constructor found
	at org.springframework.beans.BeanUtils.instantiateClass(BeanUtils.java:146) ~[na:na]
	Suppressed: reactor.core.publisher.FluxOnAssembly$OnAssemblyException: 
Error has been observed at the following site(s):
	*__checkpoint ⇢ Handler com.example.mustache.webflux.TestController#index(String, Model) [DispatcherHandler]
	*__checkpoint ⇢ HTTP GET "/greeting" [ExceptionHandlingWebHandler]
Original Stack Trace:
		at org.springframework.beans.BeanUtils.instantiateClass(BeanUtils.java:146) ~[na:na]
		at org.springframework.web.reactive.result.view.UrlBasedViewResolver.instantiateView(UrlBasedViewResolver.java:283) ~[mustache-webflux:6.0.0-SNAPSHOT]
		at org.springframework.web.reactive.result.view.UrlBasedViewResolver.createView(UrlBasedViewResolver.java:298) ~[mustache-webflux:6.0.0-SNAPSHOT]
		at org.springframework.boot.web.reactive.result.view.MustacheViewResolver.createView(MustacheViewResolver.java:72) ~[mustache-webflux:3.0.0-SNAPSHOT]
		at org.springframework.web.reactive.result.view.UrlBasedViewResolver.resolveViewName(UrlBasedViewResolver.java:237) ~[mustache-webflux:6.0.0-SNAPSHOT]
		at org.springframework.web.reactive.result.view.ViewResolutionResultHandler.lambda$resolveViews$2(ViewResolutionResultHandler.java:274) ~[mustache-webflux:6.0.0-SNAPSHOT]
		at reactor.core.publisher.FluxConcatMapNoPrefetch$FluxConcatMapNoPrefetchSubscriber.onNext(FluxConcatMapNoPrefetch.java:183) ~[mustache-webflux:3.5.0-M4]
		at reactor.core.publisher.FluxIterable$IterableSubscription.slowPath(FluxIterable.java:272) ~[mustache-webflux:3.5.0-M4]
[...]

PR #32030, opened 2022-08-09
Remove use of reflection in MustacheViewResolver

mdeinum:
Prior to this commit the MustacheViewResolver used reflection
to instantiate a MustacheView class, which fails when using AOT.
Creating the view without reflection (analogous to the
FreemarkerViewResolver) will fix this without the need for additional
constructor hints.

Fixes: #32028

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/result/view/MustacheViewResolver.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/result/view/MustacheViewResolver.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2022 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR.

> Fixes: https://github.com/spring-projects/spring-boot/issues/32028

Have you tried the smoke test against your change?

mdeinum:
There was a style error which failed the validation (apparently they are slightly different then the Spring Framework ones).

snicoll:
Sorry, that wasn't my question. You've mentioned that it fixes #32028 so I am asking if you've tested https://github.com/spring-projects/spring-aot-smoke-tests/tree/main/mustache-webmvc with that change.

mdeinum:
Tried to run it but it downloads the dependencies from the SNAPSHOT repository so it still has the old code and obviously it fails the tests. How can I run the test (it should be the `mustache-webflux` one) with my local artifacts? 

As this commit removes the reflection and replaces it with a regular `new MustacheView()` (just as with FreeMarker and the servlet `MustacheViewResolver`) I'm pretty confident it will work. Nonetheless if there is an easy way to run the test with a local artifact I'm going to try and run it.

snicoll:
Thanks for the reply and for trying out the sample. You can run the test with Gradle's `--include-build` and point to your checkout of Spring Boot with the change.

mdeinum:
Thanks for the hint. Just ran the tests and I have a successful build with the change.

Issue #10827, opened 2017-10-30
Provide an Actuator endpoint for non-indexed session repositories


wilkinsona:
Spring Session's reactive support doesn't appear to have an equivalent of `FindByIndexNameSessionRepository` so the functionality of the current endpoint can't be ported over to a reactive view of the world. We'll need to figure out what to do there.

/cc @rwinch @vpavic

rwinch:
@wilkinsona Thanks for reaching out. We are currently planning on adding support for finding by username for 2.1. For tracking I created https://github.com/spring-projects/spring-session/issues/914

wilkinsona:
Thank you @rwinch. I’ll assign this one to our general 2.1 bucket for now on the assumption that our 2.1 scheduled will align. We can adjust as needed if that’s not the case.

wilkinsona:
We've repurposed this to provide an endpoint for an unindexed session repository that will allow sessions to be retrieved and deleted by ID. The ID will have to be found by some other means as listing all sessions will not be possible.

vpavic:
Could you clarify whether this issue is still focused on reactive stack only or is it a general (meaning, targeting both Servlet and reactive stacks) concern now? If it's a general concern sounds like the session endpoint should probably be refactored in a way to separate functionality that depends on only `SessionRepository` from bits that depend on `FindByIndexNameSessionRepository`.

Either way, I'd gladly help out with this one.

Also, is there an example of actuator endpoint (outside of health, but that one's quite complex) that supports both stacks?

wilkinsona:
> Could you clarify whether this issue is still focused on reactive stack only or is it a general (meaning, targeting both Servlet and reactive stacks) concern now?

It's a general concern now.

> refactored in a way to separate functionality that depends on only SessionRepository from bits that depend on FindByIndexNameSessionRepository

There are a few different ways this could be implemented. It should be possible for the existing `SessionsEndpoint` to take a `SessionRepository` and then do an `instanceof` check in `sessionsForUsername`, returning `null` when the repository isn't a `FindByIndexNameSessionRepository`.

> Either way, I'd gladly help out with this one.

That would be great, thank you. I'll assign the issue to you.

> Also, is there an example of actuator endpoint (outside of health, but that one's quite complex) that supports both stacks?

[...]

Issue #32024, opened 2022-08-08
AuditEventRepository is not suitable for reactive


vpavic:
With its current API where all operations are of blocking nature, `AuditEventRepository` cannot be properly implemented using a non-blocking data store and is therefore not really suitable for use on a reactive runtime.

PR #32046, opened 2022-08-09
Provide an Actuator endpoint for non-indexed session repositories

vpavic:
WIP

Closes gh-10827


spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/session/ReactiveSessionsEndpointWebIntegrationTests.java
```diff
@@ -0,0 +1,75 @@
...
+ * @author Vedran Pavic
+ */
+class ReactiveSessionsEndpointWebIntegrationTests {
+
+	private static final Session session = new MapSession();
+
+	@SuppressWarnings("unchecked")
+	private static final ReactiveSessionRepository<Session> sessionRepository = mock(ReactiveSessionRepository.class);
+
+	@WebEndpointTest
```
vpavic, line 44: @wilkinsona is there a way to have `@WebEndpointTest` run only on some endpoint implementations, like WebFlux in this case? If not, what is the preferred way of writing actuator endpoint tests that target only a specific implementation?

wilkinsona: There isn't.

One option would be to copy-paste `WebEndpointTestInvocationContextProvider.WebFluxEndpointConfiguration` into these tests and modify it as needed. However, I think it might be better if `WebEndpointTest` had an attribute that allowed the tests to be run against particular web frameworks rather than always running against all three of them.

vpavic: Thanks for the early feedback - I've opened https://github.com/spring-projects/spring-boot/issues/32054 to consider enhancing `@WebEndpointTest` as suggested.

wilkinsona: Thanks. Could you please try the changes in https://github.com/wilkinsona/spring-boot/tree/gh-32054 and see if they meet your needs here?

vpavic: I did, and outside of Checkstyle bugging me about redundant `public` modifier [here](https://github.com/wilkinsona/spring-boot/blob/7d64e9496a425aed3af5c9e543fcf212b5d8deb9/spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/endpoint/web/test/WebEndpointTest.java#L50), everything was great. I've updated the PR and pushed those changes.

I hope to continue working on the PR tomorrow and expect to wrap up the changes by the end of the week.

vpavic:
This is now ready for review.

Note @wilkinsona that `:spring-boot-project:spring-boot-docs:asciidoctor*` tasks are failing due a problem that seems related with the two different `@Endpoint`s having the same id (as discussed in https://github.com/spring-projects/spring-boot/issues/10827#issuecomment-1209664435). You should see this failure in the CI as well.

wilkinsona:
As part of merging this, we'll need to figure out what to do about the two endpoints with the same ID. This may require some changes in the configuration properties annotation processor.

vpavic:
FWIW, I think sessions endpoint won't be the only endpoint that needs two different implementations which share the same id. https://github.com/spring-projects/spring-boot/issues/32024 has not been triaged yet, but I don't see any other way around the limitation expressed there outside of doing something similar like with sessions endpoint.

vpavic:
Any updates on this?

Status quo means that the default session repository for the most popular session data store (Redis) does not have actuator endpoint capability.

mhalbritter:
I've updated the PR, it's now using the `ReactiveFindByIndexNameSessionRepository`: https://github.com/mhalbritter/spring-boot/tree/pr/32046

I'll take a look at the problem with the two endpoint ids which are the same.

mhalbritter:
PR for ignoring the duplicate property: https://github.com/spring-io/spring-asciidoctor-extensions/pull/95

mhalbritter:
Thanks @vpavic!

PR #32051, opened 2022-08-10
Add socketKeepAlive configuration property for Elasticsearch

puppylpg:
Keepalive property for the connection between client and elasticsearch should be considered as configurable. Futher more, i do think it's good to keep it alive by default.

Here is the issue:

When read/write elasticsearch in high concurrency, threads have chances to stuck on obsolete connection between client and elasticsearch. The more threads involves, the quicker stucking on connection happens. The reason is described [here](https://github.com/elastic/elasticsearch/issues/65213#issue-745760090):
> Worse, if the connection is dropped silently while a request is in flight then the client may block indefinitely since it will send no more data on this connection and will therefore never find out that it's been closed.

And here is the situation i run into. Logs can also be found here:
- https://stackoverflow.com/questions/73217318/elasticsearch-java-response-stuck

similar issue:
- https://github.com/elastic/elasticsearch/issues/65909

However, spring boot can't config keepalive through current properties. We have to write some config codes like:
```
    @Bean
    public RestClientBuilderCustomizer keepalive() {
        return new RestClientBuilderCustomizer() {
            @Override
            public void customize(RestClientBuilder builder) {

            }

            @Override
            public void customize(HttpAsyncClientBuilder builder) {
                builder.setDefaultIOReactorConfig(IOReactorConfig.custom().setSoKeepAlive(true).build());
                log.info("es tcp keep-alive must be enabled!");
            }
        };
    }
```

So i add some support for `spring.elasticsearch.keepalive=true` to make it easier to config.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchProperties.java
@@ -63,9 +63,9 @@ public class ElasticsearchProperties {
 	private String pathPrefix;
 
 	/**
-	 * Whether to enable keepalive between client and Elasticsearch.
+	 * Whether to enable socket keep alive between client and Elasticsearch.
 	 */
-	private boolean keepalive = true;
+	private boolean socketKeepAlive = false;
 
 	private final Restclient restclient = new Restclient();
 
@@ -117,12 +117,12 @@ public class ElasticsearchProperties {
 		this.pathPrefix = pathPrefix;
 	}
 
-	public boolean isKeepalive() {
-		return this.keepalive;
+	public boolean isSocketKeepAlive() {
+		return this.socketKeepAlive;
 	}
 
-	public void setKeepalive(boolean keepalive) {
-		this.keepalive = keepalive;
+	public void setSocketKeepAlive(boolean socketKeepAlive) {
+		this.socketKeepAlive = socketKeepAlive;
 	}
 
 	public Restclient getRestclient() {
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchProperties.java
@@ -62,6 +62,11 @@ public class ElasticsearchProperties {
 	 */
 	private String pathPrefix;
 
+	/**
+	 * Whether to enable keepalive between client and Elasticsearch.
+	 */
+	private boolean keepalive = true;
+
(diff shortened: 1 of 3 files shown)


wilkinsona (changes requested):
Thanks for the proposal, @puppylpg. I've left a couple of comments for you consideration. In addition to those, it would be good to have a test that verifies that our default for the new property aligns with Elasticsearch's default. If the Elastic default is changed in the future, this will ensure that we notice and can update our default to match.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchProperties.java
```diff
@@ -62,6 +62,11 @@ public class ElasticsearchProperties {
 	 */
 	private String pathPrefix;
 
+	/**
+	 * Whether to enable keepalive between client and Elasticsearch.
+	 */
+	private boolean keepalive = true;
```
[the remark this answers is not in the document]
puppylpg: I do think keep alive is important enough to be enabled by default to avoid sneaky troubles, however I also approve what you said: leave it to Elasticsearch itself, and change the default behavior when Elasticsearch's default changes.

puppylpg (commented):
resolved

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientAutoConfigurationTests.java
```diff
@@ -183,6 +183,27 @@ void configureWithCustomPathPrefix() {
 		});
 	}
 
+	@Test
+	void configureWithNoSocketKeepAliveApplyDefaults() {
```
wilkinsona, line 187: This isn't quite what I meant about testing that the defaults match. This checks that Spring Boot's default is `false`, but it doesn't check that Elasticsearch has the same default. Rather than creating an application context and getting the `RestClient` from it, the test itself should use `RestClientBuilder` to create a `RestClient` in its default configuration. It should then check that this `RestClient` has `soKeepAlive` set to `false`.

wilkinsona (changes requested):
Thanks for the updates. I've left a comment about the testing of Elastic's default.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchRestClientAutoConfigurationTests.java
puppylpg: ok, that's resonable. Thanks for your patience~

snicoll:
@puppylpg thank you for making your first contribution to Spring Boot.

puppylpg:
ok, thanks for your reply (sad, for still using jdk8 :sweat_smile::sweat_smile::sweat_smile:)

Issue #32074, opened 2022-08-13
PropertiesPropertySourceLoader class is not convenient to customize the character set UTF-8


lizongbo:
### Versions
Spring Boot Version 2.7.2

### Overview
(I'm sorry ,The English translation comes from the search engine Bing.com )
I was migrating the previous Spring application to the current SpringBoot application, and I found that the encoding of application.properties was ISO_8859_1, so I customized the chaset to UTF-8, but because SpringBoot's class PropertiesPropertySourceLoader  is not convenient to inherit, I can only use the ugly inheritance implementation to set (Copy OriginTrackedPropertiesLoader.java to my customized java file). It is therefore expected that class plus an extended feature that allows the specified encoding to be used makes it easy to customize the character set using a concise inheritance implementation.

config in Spring application:

java code :

```java
@PropertySource(value = "classpath:application.properties", encoding = "UTF-8")
```

or xml code:

```xml
<context:property-placeholder location="classpath:application.properties" file-encoding="UTF-8"/>
```

### My custom extension steps are as follows:

1. go to https://start.spring.io/ , generate a gradle project with Java 8 and SpringBoot 2.7.2.

2. add chinese value to application.properties,just like follow:

```
  testCnVal = 测试中文
  testCnValUcode = \u6D4B\u8BD5\u6C49\u5B57
  testArrayVal[] = a,b,c
```

![image](https://user-images.githubusercontent.com/1670666/184503582-3bc0c795-d8af-41ce-8c37-83e5e9020aad.png)

3. write test code to get value,just like follow:

```
@Service
public class PropTestBean {

    @Value("${testCnVal}")
    private String testCnVal;
    @Value("${testCnValUcode}")
    private String testCnValUcode;
    @Value("${testArrayVal[1]}")
    private String testArrayVal1;

    @EventListener
    void onRefreshEvent(ContextRefreshedEvent event) {
        System.out.println("testCnVal==" + testCnVal);
        System.out.println("testCnValUcode==" + testCnValUcode);
        System.out.println("testArrayVal1==" + testArrayVal1);
        Environment env = event.getApplicationContext().getEnvironment();
        System.out.println("find key|testArrayVal=" + env.containsProperty("testArrayVal"));
        System.out.println("find key|testArrayVal[]=" + env.containsProperty("testArrayVal[]"));
        System.out.println("find key|testArrayVal[1]=" + env.containsProperty("testArrayVal[1]"));
[...]

Issue #28663, opened 2021-11-13
charset wrong for "spring.config.import"


lfz757077613:
It work well when I set my properties in application.yml directly
But when I set these properties in additional file imported by "spring.config.import" like below, its charset is wrong
Is there some options to set charset for "spring.config.import"? Otherwise I have to use @PropertySource to set charset, but it do not support yml file
![image](https://user-images.githubusercontent.com/16242100/141604551-f95ce101-8ec0-441f-bb1f-519773f4cf03.png)

wilkinsona:
What do you mean by wrong? Which charset did you expect to be used and which charset was actually used? Can you provide a minimal sample that reproduces the problem so that we can see exactly what you mean? You can share one with us by zipping it up and attaching it to this issue or by pushing it to a separate repository on GitHub.

lfz757077613:
> What do you mean by wrong? Which charset did you expect to be used and which charset was actually used? Can you provide a minimal sample that reproduces the problem so that we can see exactly what you mean? You can share one with us by zipping it up and attaching it to this issue or by pushing it to a separate repository on GitHub.

you can try this demo project, all of files are utf8
![image](https://user-images.githubusercontent.com/16242100/141613161-3d2b859f-2d86-418d-90e1-d0ac0a159056.png)

[springtest.zip](https://github.com/spring-projects/spring-boot/files/7531708/springtest.zip)

wilkinsona:
Java's default encoding for properties files is ISO-8859-1 and we deliberately align with this (see https://github.com/spring-projects/spring-boot/issues/5361 and https://github.com/spring-projects/spring-boot/pull/10565). As [recommended in Properties' javadoc](https://docs.oracle.com/en/java/javase/11/docs/api/java.base/java/util/Properties.html#load(java.io.InputStream)) you should use Unicode escapes for non-Latin1 characters. YAML uses UTF-8 by default which is why the value of `static-config.right` is correct.

We can consider offering a setting to load `.properties` files using `UTF-8` but it will be somewhat difficult to set as it creates something of chicken and egg problem.

lfz757077613:
Thanks! I understand, this case confused me a long time.
For now, I think "yml" is more convenient than "properties" in springboot project.
Perhaps "spring.config.import" should provide a prefix for encoding such like "optonal" prefix.
Thanks again.

philwebb:
[...]

PR #32077, opened 2022-08-14
Add field and method to allow easy custom application.properties file encoding

lizongbo:
add protected field and method,so can easy extends the PropertiesPropertySourceLoader class to custom default encoding,for example UTF-8

#32074


wilkinsona:
Thanks for the proposal. Unfortunately, we need to complete the design work that is being tracked by https://github.com/spring-projects/spring-boot/issues/28663 before we can consider any changes in this area.

PR #32083, opened 2022-08-15
Allow to disable JacksonFeature in JerseyAutoConfiguration

xhaggi:
The Spring Boot auto-configuration for Jersey registers the `JacksonFeature` by default if an `ObjectMapper` bean is present. This is always the case if Jackson is on the classpath, because `JacksonAutoConfiguration` kicks in. In some cases, however, you may want to use Jersey without preconfigured Jackson, e.g. if you use MOXy instead. Having both configured in `ResourceConfig` leads to problems with XML serialization.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jersey/JerseyAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jersey/JerseyAutoConfiguration.java
@@ -74,6 +74,7 @@ import org.springframework.web.filter.RequestContextFilter;
  * @author Andy Wilkinson
  * @author Eddú Meléndez
  * @author Stephane Nicoll
+ * @author Sascha Woo
  * @since 1.2.0
  */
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the proposal. Unfortunately, I don't think this is quite the right way to provide this capability as it doesn't feel extensible enough. In the future, we may have other features that we want to enable and disable using properties.

In the particular scenario that you have described, I wonder if excluding `jersey-media-json-jackson` from the classpath would be sufficient. If not, please open an issue with some more details about how you're using Jackson and MOXy with Jersey and we can consider what options are available to meet your needs.

xhaggi:
>In the particular scenario that you have described, I wonder if excluding jersey-media-json-jackson from the classpath would be sufficient.

It does not. We need `jersey-media-json-jackson` for a Jersey client in the same project. For now we exclude `JacksonAutoConfiguration` as a workaround, because the `JacksonResourceConfigCustomizer` requires an `ObjectMapper` that is created by `JacksonAutoConfiguration`. Not so nice, but it works ;). Thank you for your reply, I will open an issue for this.

Issue #31171, opened 2022-05-25
Consider alignment of MimeMappings with Tomcat's defaults


wilkinsona:
See https://github.com/spring-projects/spring-boot/issues/30885 for background. Tomcat now has many more default mappings. We need to consider if we want to align with those new defaults and, if we do, how to go about it. There may be a more memory efficient way to do it then simply updating the static map in `MimeMappings`.

philwebb:
Closing in favor of PR #32101. Thanks @terminux!

PR #32101, opened 2022-08-17
Alignment of MimeMappings with Tomcat's defaults

terminux:
Hi, this PR is used to solve #31171 .

* Migrate default mime mappings maintained in `MimeMappings` into properties file

* Load `org/apache/catalina/startup/MimeTypeMappings.properties` and assert that each entry has an override.

Closes gh-31171


snicoll:
We need to double check that `mime-mappings.properties` is covered by a resource hint or we won't be able to load the file. I wonder if it would be better to update what we have (list in code) and then have the loading of the file only in the test.

terminux:
Thanks for your feedback @snicoll . I've also considered updating the list directly in the code, but since there are so many mappings (1000+), maybe it's easier to maintain using a properties file.

bclozel:
Also did we consider Andy's comment in the linked issue?
> There may be a more memory efficient way to do it then simply updating the static map in MimeMappings.

philwebb:
> There may be a more memory efficient way to do it then simply updating the static map in MimeMappings.

I think that may have referred to the idea that loading properties from a file might be better than having a large static block in the class. One advantage of the file is we could load it on demand then let it get garbage collected.

Another, more complex idea would be to override `findMimeMapping` and `findMimeMappings` in `org.apache.catalina.Context`. We could keep the most common mappings in memory and only load the full list if a more exotic extension is requested.

Either way, I generally think having the mappings in a file opens up more options to us.

philwebb:
Thanks @terminux, this is now in main along with some refinements that should hopefully reduce memory usage for Tomcat users that aren't doing anything exotic.

PR #32123, opened 2022-08-20
Add JVM options to prevent warnings with asciidoctor

dsyer:
See https://github.com/asciidoctor/asciidoctor-gradle-plugin/issues/597 for details.


philwebb:
Thanks Dave!

Issue #32047, opened 2022-08-09
Provide property to switch to simplified logging format


philwebb:
See https://twitter.com/bsideup/status/1557116232146554882

We could offer a quick property to switch to simplified logging.

wilkinsona:
I'm not sure about this. IMO, demos are better when things with which the audience is likely to already be familiar look as they expect them to look. This then means that the audience can focus on the new stuff rather than being distracted by things that aren't as expected.

PR #32124, opened 2022-08-20
Provide logging.pattern.console.simple property

An1s9n:
Dear team,

this pull request is my attempt to help with #32047 issue. I kindly ask you to review it and tell me what should be fixed.

Thank you in advance for your time!


wilkinsona:
Thanks for the PR, @An1s9n, but I'm not yet convinced that this is a good idea. If we do eventually decide to implement it we'll have to do so differently to what you have proposed here. The existing `logging.pattern.console` property prevents us from adding a `logging.pattern.console.simple` property as we do not want property names to overlap. Thanks anyway.

Issue #32141, opened 2022-08-22
Hazelcast auto-configuration recognizes hazelcast.xml and hazelcast.yaml files but not hazelcast.yml


neilstevenson:
Hazelcast autoconfiguration recognises `hazelcast.xml` and `hazelcast.yaml` but not `hazelcast.yml`

Similarly for clients, autoconfiguration recognises `hazelcast-client.xml`, `hazelcast-client.yaml` but not `hazelcast-client.yml`

Hazelcast itself accepts `.yaml` and `.yml`, see https://github.com/hazelcast/hazelcast/blob/master/hazelcast/src/main/java/com/hazelcast/internal/config/DeclarativeConfigUtil.java#L67

Enhancement is to make Spring Boot recognise `yml` suffix config files

philwebb:
Closing in favor of PR #32142. Thanks @neilstevenson!

PR #32142, opened 2022-08-22
Hazelcast auto-configuration recognizes hazelcast.xml and hazelcast.yaml files but not hazelcast.yml

neilstevenson:
Fixes issue #32141

Adds support for "yml" suffix config files. "yaml" already supported


snicoll:
Thanks @neilstevenson

PR #32143, opened 2022-08-22
Add null check on a actuator endpoint

justinbaskaran:
Adding these changes, to prevent null / empty string data from entering into constructor method of the actuator


snicoll:
@justinbaskaran thanks for the PR but I think the code is fine as it is. It is better to assign the field rather than silently ignoring null values. `setDataSource` can be used to set the `DataSource` after construction and before `afterPropertiesSet` is invoked.

PR #32144, opened 2022-08-23
Netty 'spring.netty leak detection' default property value is always applied to resource leak detector

### LINKED ISSUE #32107: Netty 'spring.netty-leak-detection' default property value is always applied to ResourceLeakDetector
Spring will override the Netty leak detection level set using `-Dio.netty.leakDetection.level=advanced`

The code below want to set leak detection if property was not null. But the property has default value as SIMPLE.  This will get  leak detection overridden as SIMPLE always if you set it other value like 'advanced' outside of Spring 

```java
public class NettyProperties {
	private LeakDetection leakDetection = LeakDetection.SIMPLE;

...
public class NettyAutoConfiguration {
	public NettyAutoConfiguration(NettyProperties properties) {
		if (properties.getLeakDetection() != null) {
			NettyProperties.LeakDetection leakDetection = properties.getLeakDetection();
			ResourceLeakDetector.setLevel(ResourceLeakDetector.Level.valueOf(leakDetection.name()));
		}
	}
}
...
```

changed to 
```
if (properties.getLeakDetection() != LeakDetection.SIMPLE) {
```

will solve the problem as LeakDetection will be simple if not set. 

Happy to raise a PR if this is 

Version:  current main branch

russellyou:
Fix #32107


snicoll:
@russellyou thank you for making your first contribution to Spring Boot.

Issue #2814, opened 2015-04-13
spring-boot-starter-velocity incompatible with spring-boot-starter-data-jpa


stunaz:
Hello,
Using Spring boot 1.2.3 , spring-boot-starter-velocity is incompatible with spring-boot-starter-data-jpa due to the resolved version of antlr

wilkinsona:
I assume you're using Maven? If you order your dependencies with `spring-boot-starter-data-jpa` first:

```xml
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-data-jpa</artifactId>
</dependency>
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-velocity</artifactId>
</dependency>
```

Maven will choose to use Antlr 2.7.7. However, if you list `spring-boot-starter-velocity` first:

```xml
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-velocity</artifactId>
</dependency>
<dependency>
	<groupId>org.springframework.boot</groupId>
	<artifactId>spring-boot-starter-data-jpa</artifactId>
</dependency>
```

Maven will choose Antlr 2.7.2. This difference is because of [Maven's "dependency mediation"](https://maven.apache.org/guides/introduction/introduction-to-dependency-mechanism.html):

> -  _Dependency mediation_ - this determines what version of a dependency will be used when multiple versions of an artifact are encountered. Currently, Maven 2.0 only supports using the "nearest definition" which means that it will use the version of the closest dependency to your project in the tree of dependencies. You can always guarantee a version by declaring it explicitly in your project's POM. Note that if two dependency versions are at the same depth in the dependency tree, until Maven 2.0.8 it was not defined which one would win, but since Maven 2.0.9 it's the order in the declaration that counts: the first declaration wins.
 - "nearest definition" means that the version used will be the closest one to your project in the tree of dependencies, eg. if dependencies for A, B, and C are defined as A -> B -> C -> D 2.0 and A -> E -> D 1.0, then D 1.0 will be used when building A because the path from A to D through E is shorter. You could explicitly add a dependency to D 2.0 in A to force the use of D 2.0

If you were using Gradle you would always get 2.7.7 as, when there are multiple versions of a dependency in the graph, Gradle will choose the latest version.

There are a few ways you can fix this yourself:

 - List `spring-boot-starter-data-jpa` first in your pom
[...]

PR #32149, opened 2022-08-23
Remove ANTLR 2 dependency management

hak7a3:
Dependency management for ANTLR 2 is introduced in #2814 to solve conflicts of Hibernate and Velocity.
However, it is unnecessary in Spring Boot 3.0 because of following reasons:

- Velocity is not supported since Spring Boot 1.5.
- Hibernate 6.1.1.Final use ANTLR 4.


wilkinsona:
Thanks very much, @hak7a3.

PR #32162, opened 2022-08-24
Don't warn about missing Freemarker templates if disabled

chrylis:
The `FreeMarkerAutoConfiguration` logs a warning about missing templates even if the entire auto-configuration class is disabled via properties. I am using FreeMarker internally for non-Web reasons, but its presence triggers spurious warnings from the auto-configuration even after adding `spring.freemarker.enabled: false`.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/freemarker/FreeMarkerAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/freemarker/FreeMarkerAutoConfiguration.java
@@ -61,7 +61,7 @@ public class FreeMarkerAutoConfiguration {
 	}
 
 	public void checkTemplateLocationExists() {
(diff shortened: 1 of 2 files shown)


philwebb:
The `spring.freemarker.enabled` property is somewhat poorly named. It's actually documented as "Whether to enable MVC view resolution for this technology".

I don't think we can apply the fix that you suggest because we have `FreeMarkerNonWebConfiguration` which will apply even if `spring.freemarker.enabled` is false and calls `factory.setTemplateLoaderPaths(this.properties.getTemplateLoaderPath());` (via `AbstractFreeMarkerConfiguration`).

Setting `spring.freemarker.check-template-location` to `false` should stop the warning.

chrylis:
This behavior conflicts with general behavior of other `enabled` properties, which is to shut off the relevant auto-configuration bean registrations entirely. I'm aware that this is a backward-compatibility issue, but I suggest that for Boot 3 the property `spring.freemarker.enabled` should actually prevent Boot from registering any beans related to FreeMarker; as it is, I'd have to actively exclude the base auto-configuration class itself.

PR #32186, opened 2022-08-29
Refine `ConfigurationPropertiesReflectionHintsProcessor`

sdeleuze:
This commit refines `ConfigurationPropertiesReflectionHintsProcessor` Java bean properties handling in order to register reflection hints only for getters and setters, not for all methods.

It avoids including unconditionally method like `SpringApplication#load` which in turn avoids shipping `BeanDefinitionLoader` and related transitively used classes in the native image.

The gain is significant: it allows to remove up to [700 classes](https://gist.github.com/sdeleuze/3a4ebee08e0f86c3896378aab29d52b3#file-configuration-properties-diff) (when no XML parser is used elsewhere) and to reduce the memory footprint by 2M of RSS.


philwebb:
Thanks @sdeleuze. Is there any good way to unit test this? The changes look quite subtle and I'm worried we might accidentally revert back at some point and not notice that we've made things worse.

sdeleuze:
Sure, I will update the PR with `ConfigurationPropertiesReflectionHintsProcessor` unit tests.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/SpringApplicationTests.java
```diff
@@ -158,7 +157,7 @@
  * @author Marten Deinum
  * @author Nguyen Bao Sach
  * @author Chris Bono
- * @author Brian Clozel
```
sdeleuze, line 161: Duplicate

sdeleuze:
I have pushed a refined commit with proper tests.

sdeleuze:
More related to https://github.com/spring-projects/spring-aot-smoke-tests but potentially complementary to the tests provided in this PR : we could use `-H:ReportAnalysisForbiddenType=org.springframework.boot.BeanDefinitionLoader` native image parameter to   fails the native build of let say `command-line-runner` sample if this class is in the resulting native image (it should not). Once this PR merged, I can submit a PR on smoke tests side to add this.

Issue #28208, opened 2021-10-06
Example log output is out-of-date and inconsistent


wilkinsona:
https://docs.spring.io/spring-boot/docs/2.4.x/reference/htmlsingle/#boot-features-spring-application

```
  .   ____          _            __ _ _
 /\\ / ___'_ __ _ _(_)_ __  __ _ \ \ \ \
( ( )\___ | '_ | '_| | '_ \/ _` | \ \ \ \
 \\/  ___)| |_)| | | | | || (_| |  ) ) ) )
  '  |____| .__|_| |_|_| |_\__, | / / / /
 =========|_|==============|___/=/_/_/_/
 :: Spring Boot ::   v2.4.11

2019-04-31 13:09:54.117  INFO 56603 --- [           main] o.s.b.s.app.SampleApplication            : Starting SampleApplication v0.1.0 on mycomputer with PID 56603 (/apps/myapp.jar started by pwebb)
2019-04-31 13:09:54.166  INFO 56603 --- [           main] ationConfigServletWebServerApplicationContext : Refreshing org.springframework.boot.web.servlet.context.AnnotationConfigServletWebServerApplicationContext@6e5a8246: startup date [Wed Jul 31 00:08:16 PDT 2013]; root of context hierarchy
2019-04-01 13:09:56.912  INFO 41370 --- [           main] .t.TomcatServletWebServerFactory : Server initialized with port: 8080
2019-04-01 13:09:57.501  INFO 41370 --- [           main] o.s.b.s.app.SampleApplication            : Started SampleApplication in 2.992 seconds (JVM running for 3.658)
```

As we've updated the example over the years, things have become a bit inconsistent:

- The PID isn't the same in all the log lines
- The date goes backwards
- The logged startup date doesn't match the date at the start of the line

I wonder if we could automate capturing the example output to avoid the problem reoccurring.

wilkinsona:
This has got a bit worse in the 3.0 docs since we [changed the date format](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0.0-M3-Release-Notes#logging-date-format). I'd like to fix this in all branches so that we don't have to remember to update things manually any more.

PR #32196, opened 2022-08-31
Add example Conditions Report for reference

kirankashyap:
Right now there is no example for Conditions Report in the documentation. I believe it will help the reader to see an example of how it looks like.


wilkinsona:
Thanks for the suggestion. If we do this, and I'm not yet sure that we should, the example report should be generated at build time ([as we did in this issue for other console output examples](https://github.com/spring-projects/spring-boot/issues/28208)). Please open an issue so that we can consider it.

PR #32203, opened 2022-08-31
Update SnakeYAML to address CVE-2022-25857

LoganDev99:
Update SnakeYAML to address CVE-2022-25857

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1382,7 +1382,7 @@ bom {
 			]
 		}
 	}
-	library("SnakeYAML", "1.30") {
+	library("SnakeYAML", "1.31") {
 		group("org.yaml") {
 			modules = [
 				"snakeyaml"


bclozel:
Thanks but as stated in our issue template, we don't accept issues or PRs for trivial dependency upgrades. This will be taken care of in due time.

> 🙅 "Please DO NOT Raise an Issue" Cases
> - Managed Dependency Upgrade
> You DO NOT need to raise an issue for a managed dependency version upgrade as there's a semi-automatic process for checking managed dependencies for new versions before a release. BUT pull requests for upgrades that are more involved than just a version property change are still most welcome.

and0p:
> Thanks but as stated in our issue template, we don't accept issues or PRs for trivial dependency upgrades. This will be taken care of in due time.

@bclozel I appreciate that sentiment, but anecdotally: my company is having high-priority failures in [snyk](https://snyk.io/) over this for all of our Spring services, and I'd wager a lot of other companies are as well. Someone more knowledgable than I has determined this is a critical vulnerability, so it might be worth investigating how vulnerable Spring is. Or just considering how this could be affecting certain CI/CD pipelines, if that's a priority for contributors, but I'd understand perfectly if not.

Thanks either way! Just wanted to throw in my 🎩  / 🪙🪙

bclozel:
@and0p this doesn't mean we don't care about this upgrade, quite the opposite. With many active branches and lots of managed dependencies, we do need to ensure that we don't miss an upgrade and that we're always sticking to [our upgrade policy](https://github.com/spring-projects/spring-boot/wiki/Team-Practices#third-party-dependency-upgrades). This needs to be semi-automated or we'll risk missing important upgrades. You [can take a look at our issue tracker](https://github.com/spring-projects/spring-boot/issues?q=is%3Aissue+is%3Aclosed+label%3A%22type%3A+dependency-upgrade%22+) if you're not convinced, as your comment suggests.

For example, this PR has been sent against the main branch, so this would have upgraded the dependency for the upcoming 3.0.0-M5 only. I guess you're more interested in getting an upgrade [for the supported branches](https://spring.io/projects/spring-boot#support), with 2.6.12 and 2.7.4. Backporting commits is also prone to human errors!

It's often really hard for security tools to detect if an application is actually vulnerable, especially if they're just looking at build dependencies. In this case, it looks like the CVE is about possible DoS triggered by malicious input. Most Spring Boot applications only need SnakeYaml to parse their own `application.yml` configuration. I don't think we can consider this content as untrusted input.

If your application is using SnakeYaml to decode untrusted Yaml, for example from a web controller, you should [override the SnakeYAML version property](https://docs.spring.io/spring-boot/docs/current/reference/html/dependency-versions.html#appendix.dependency-versions.properties) (`snakeyaml.version`) as soon as possible in your [Gradle](https://docs.spring.io/spring-boot/docs/2.7.3/gradle-plugin/reference/htmlsingle/#managing-dependencies.dependency-management-plugin.customizing) or [Maven](https://docs.spring.io/spring-boot/docs/2.7.3/maven-plugin/reference/htmlsingle/#using.parent-pom) build. In this case, waiting for a new Spring Boot release to patch that version would not be safe.

and0p:
@bclozel makes perfect sense, thanks for the response.

PR #32205, opened 2022-08-31
Add support for configuring Spring Session Redis repository

### LINKED ISSUE #30673: Adapt to Spring Session Data Redis switching to an unindexed repository by default
See https://github.com/spring-projects/spring-session/issues/1711 for background.

Spring Session's Redis module has switched to a simpler default repository implementation. This broke Boot's build as the simpler repository does not require a cleanup cron. 4a401bfa16e8fc20810e52288a9e9b3fb97dd00b has adapted to the breaking change but means that Boot's still using an indexed repository by default.

We should probably align our default with Spring Session's and use the simpler, more performant repository by default. To do so, we'll need to do something with the `spring.session.redis.cleanup-cron` as it only applies to the indexed repository. We may also want to provide a configuration property to control the type of repository that's used, making it easy for users to switch back to the indexed repository if they wish to do so.

/cc @eleftherias @vpavic

vpavic:
With Spring Session moving to `RedisSessionRepository` as the preferred session repository, Spring Boot auto-configuration should make it possible to easily switch back to the previous default (`RedisIndexedSessionRepository`).

This commit introduces `spring.session.redis.repository` configuration property that allow selecting the desired Redis-backed session repository implementation.

Closes #30673

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -2758,6 +2758,20 @@
       "defaultValue": "on-save"
     },
     {
+      "name": "spring.session.redis.repository",
+      "description": "Redis session repository implementation to use.",
+      "values": [
+        {
+          "value": "default",
+          "description": "Use default Redis session repository."
+        },
+        {
+          "value": "indexed",
+          "description": "Use indexed Redis session repository."
+        }
+      ]
+    },
+    {
       "name": "spring.session.redis.save-mode",
       "defaultValue": "on-set-attribute"
     },
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/RedisSessionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/RedisSessionConfiguration.java
@@ -22,6 +22,7 @@ import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnBean;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.web.ServerProperties;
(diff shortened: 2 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -2757,6 +2757,20 @@
       "name": "spring.session.redis.flush-mode",
       "defaultValue": "on-save"
     },
+    {
+      "name": "spring.session.redis.repository",
```
vpavic, line 2761: I wasn't 100% sure how to name this property, so I initially went with the simple option. Alternatively, this could maybe be named `spring.session.redis.repository-type`.

spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -2757,6 +2757,20 @@
...
+      "description": "Redis session repository implementation to use.",
+      "values": [
+        {
+          "value": "default",
```
vpavic, line 2765: `RedisSessionRepository` by itself doesn't offer much inspiration on how to name this other than `default` or maybe `basic`.

spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -2757,6 +2757,20 @@
...
+          "description": "Use default Redis session repository."
+        },
+        {
+          "value": "indexed",
```
vpavic, line 2769: OTOH, `RedisIndexedSessionRepository` makes this quite straightforward. One might get tempted to throw mention of _events_ here as well (based on soon to be removed `EnableRedisHttpSession#enableIndexingAndEvents`) however I'd avoid that as support for session events is an implementation detail of a session repository (and not something expressed through a contract, like indexing capabilities).

wilkinsona:
@vpavic I see you just force-pushed. What did you change? I was just in the process of getting this ready to merge and I don't want to miss something.

vpavic:
Hey @wilkinsona, sorry I had intended to drop a note here after force pushing but forgot to do so in the end. Nothing changed, I just rebased the branch as it was few hundred commits back and also wanted to verify everything works as expected with the changes that were shipped with the latest Spring Session milestone yesterday.

vpavic:
> I've also tried to make it fail nicely if you try to configure the cleanup cron when using the default repository.

I just tried the snapshots and this works (fails?) very nicely indeed. Very useful.

Issue #32218, opened 2022-09-02
Detection of DurationStyle.ISO8601 does not support lower-case input


valentine-dev:
# Bug Report

### Version of Spring Boot

2.7.3

### Version of Java

11

## Expected (what is in the Reference Documentation)

>The standard ISO-8601 format [used by java.time.Duration](https://docs.oracle.com/javase/8/docs/api/java/time/Duration.html#parse-java.lang.CharSequence-)

from https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.typesafe-configuration-properties.conversion.durations

## Actual Behavior

Some format used by java.time.Duration is not working in Spring Boot app, like "pt25.234s" demonstrated in the example application at https://github.com/valentine-dev/spring-boot-duration-style-iso8601

## Suggestion Solution

1. Add `Pattern.CASE_INSENSITIVE` as the second argument to the `Pattern.compile` method at [Line 87 of DurationStyle.java](https://github.com/spring-projects/spring-boot/blob/01c68a3dc829e6c82a8f64841ae98c0bf09e6fa9/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/DurationStyle.java#L87)
2. Use the following pattern string in java.time.Duration source code (Line 151 ~ 152 in the file from JDK 11) at [Line 65 of DurationStyle.java](https://github.com/spring-projects/spring-boot/blob/01c68a3dc829e6c82a8f64841ae98c0bf09e6fa9/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/convert/DurationStyle.java#L65):

``"([-+]?)P(?:([-+]?[0-9]+)D)?" +``
        ``"(T(?:([-+]?[0-9]+)H)?(?:([-+]?[0-9]+)M)?(?:([-+]?[0-9]+)(?:[.,]([0-9]{0,9}))?S)?)?"``

mdeinum:
Not sure it is a bug. 

From the ISO-8601 spec (emphasize is mine)

>  The **capital** letters P, Y, M, W, D, T, H, M, and S are designators for each of the date and time elements and are not replaced

It explicitly mentions capital letters. So one could argue that Spring (Boot) might adhere to strictly to the standard as opposed the `java.time.Duration` class?

wilkinsona:
Given that we say we use the format that is used by java.time.Duration, I think it's reasonable to expect the detect and detectAndParse methods to align with Duration's own parsing.

[...]

PR #32223, opened 2022-09-02
Detection of DurationStyle.ISO8601 does not support lower-case input

valentine-dev:
Fix bug related to issue #32218 by adding lower-case P in the pattern string of DurationStyle.ISO8601.
Test cases also updated.


wilkinsona:
@valentine-dev Thanks very much for making your first contribution to Spring Boot.

valentine-dev:
@wilkinsona It is my pleasure. Looking forward to contribute more.

PR #32224, opened 2022-09-03
Clarify documentation of config sub-directory from which external application properties are read

obfischer:
The current documentation relates to the subdirectory for external configurations as `/config`. Read as UNIX path, the leading slash could be read as reference to the root directory `/`. Therefore I would suggest to use `./config` to make it a little bit more understandable.
This is of course not a major issue, but IMHO makes it a little bit more understandable.


wilkinsona:
Thanks for spotting this, @obfischer, and raising a PR. I agree that `/config` is misleading. As a replacement, I wonder if `config/` would be better than `./config`. The trailing slash is commonly used to indicate that a path is a directory (and it's what both bash and zsh tab-complete to).

obfischer:
I think @wilkinsona has a good point. I updated my pull request.

snicoll:
Thanks for the follow up and the contribution @obfischer

PR #32225, opened 2022-09-04
add example for conditions report

kirankashyap:
So far, the docs do not have an example for Conditions Report. This change generates the conditions report during the docs build and adds it to the generated docs.

Please refer #32196

### DIFF
--- a/spring-boot-project/spring-boot-docs/build.gradle
+++ b/spring-boot-project/spring-boot-docs/build.gradle
@@ -285,11 +285,21 @@ task runLoggingFormatExample(type: org.springframework.boot.build.docs.Applicati
 	normalizeTomcatPort()
 }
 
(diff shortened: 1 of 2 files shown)


vivganes:
I use this report to debug my auto config all the time. Didn't realise we didn't have an example in the docs. 

Thanks @kirankashyap for raising this PR

wilkinsona:
Thanks for the proposal. Unfortunately, having seen it in action, I don't think having an example of the conditions report is particularly useful. Unless you know exactly what application produced the report and have details of its dependencies, configuration properties, etc, seeing the auto-configurations that do and do not match doesn't feel very useful to me. Furthermore, I think the report is too long for the documentation, even with the collapsible section. It's particularly noticeable in the PDF where things can't be collapsible and the report is over 20 pages long. Thanks anyway for the PR.

PR #31998, opened 2022-08-08
Add disk space health indicator's path to its details and log message

rreich:
Directly seeing the path would make it easier for developers and administrators to find and fix a configuration issue.

PR #32234, opened 2022-09-05
Add disk space health indicator's path to its details and log message (2.6.x)

rreich:
Follow-up for https://github.com/spring-projects/spring-boot/pull/31998 on 2.6.x


bclozel:
Thanks @rreich, but we're not backporting features (especially changes in REST APIs) in maintenance versions. #31998 was scheduled for the next 3.0.0 milestone with that in mind.

rreich:
Ok, thank you.

PR #32235, opened 2022-09-06
Detection of PeriodStyle.ISO8601 does not support lower-case input

valentine-dev:
Support lower-case input in PeriodStyle  for ISO-8601 format used by java.time.Period.

This is a pull request similar to pull request #32223.

It will fix a bug similar to the one reported as #32218, found in Spring Boot version 2.7.3 with JDK 11.

The bug is demonstrated in the example application at https://github.com/valentine-dev/spring-boot-period-style-iso8601.


snicoll:
Thanks again @valentine-dev

valentine-dev:
@snicoll  My pleasure.

PR #32374, opened 2022-09-14
perf:performance code

Korov:


philwebb:
Thanks for the suggestion but I don't think the change will make any difference to performance.

PR #32419, opened 2022-09-19
Fix Spring Batch deprecations

dreis2211:
Good morning,

this PR fixes a couple of deprecations in `JobLauncherApplicationRunnerTests` and `spring-boot-smoke-test-batch`.

Cheers,
Christoph


snicoll:
Thanks again Christoph.

PR #32425, opened 2022-09-19
Integrate full Log4j 2 support into Spring Boot

rgoers:
This change merges the contents of the log4j-spring-boot module into Spring Boot, effectively making Log4j 2 on par with Logback and eliminating the need for the log4j-spring-boot module.

I was able to verify this with a sample project I have. While the project doesn't fully run (it hasn't completely converted to Spring 3) it did get far enough along to verify that I could successfully retrieve the logging configuration from Spring Cloud Config server using basic password authentication and resolve the application name using the SpringLookup.

Note: This requires Log4j 2.19.0. While building I discovered there were tests that were hard-coded to older versions of Log4j 2. These had to be modified to reference version 2.19.0 as they no longer work. Those older versions shouldn't be supported with Spring 3 in any case.


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/SpringProfileArbiter.java
```diff
@@ -0,0 +1,138 @@
...
+import org.apache.logging.log4j.core.config.plugins.PluginConfiguration;
+import org.apache.logging.log4j.core.config.plugins.PluginLoggerContext;
+import org.apache.logging.log4j.status.StatusLogger;
+
+import org.springframework.core.env.Environment;
+import org.springframework.core.env.Profiles;
+import org.springframework.util.StringUtils;
+
+/**
+ * An Aribter that uses the active Spring profile to determine if configuration should be
```
andrei-ivanov, line 36: typo

rgoers: fixed

rgoers:
Unfortunately, merge conflicts made it necessary to drop this repo and create a new PR.

PR #32464, opened 2022-09-21
Maintain independent YAML types

asomov:

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedYamlLoader.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedYamlLoader.java
@@ -165,16 +165,38 @@ class OriginTrackedYamlLoader extends YamlProcessor {
 	}
 
 	/**
-	 * {@link Resolver} that limits {@link Tag#TIMESTAMP} tags.
+	 * {@link Resolver} that limits {@link Tag#TIMESTAMP} tags, keeps only true and false
+	 * as booleans, drops octal integers, base 2 integers, NaN numbers.
 	 */
 	private static class NoTimestampResolver extends Resolver {
 
+		public static final Pattern SPRING_BOOL = Pattern.compile("^(?:true|True|TRUE|false|False|FALSE)$");
+
+		public static final Pattern SPRING_FLOAT = Pattern
+				.compile("^(" + "[-+]?(?:[0-9][0-9_]*)\\.[0-9_]*(?:[eE][-+]?[0-9]+)?" + // (base
+																						// 10)
+						"|[-+]?(?:[0-9][0-9_]*)(?:[eE][-+]?[0-9]+)" + // (base 10,
+																		// scientific
+																		// notation
+																		// without .)
+						"|[-+]?\\.[0-9_]+(?:[eE][-+]?[0-9]+)?" + // (base 10, starting
+																	// with .)
+						"|[-+]?\\.(?:inf|Inf|INF)" + ")$");
+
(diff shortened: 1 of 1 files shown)


asomov:
I find the code in PR a better solution that in this commit
https://github.com/spring-projects/spring-boot/commit/724f9ebf7107d99ac9bae4204b4f7c6444a49e03

asomov:
I wonder how can I see the build failure. Locally it is green.

wilkinsona:
You can see the failure [here](https://ci.spring.io/builds/202560#L62f3c732:4855). It's linked to from the failing check that GitHub shows below. Running `./gradlew build` locally should reproduce the failure, or you can just run `./gradlew format` to fix it.

asomov:
@wilkinsona  I did run `./gradlew format` locally (it formatted the source)
My `./gradlew build` local run is green
I use JDK 17. Should I try another JDK
I do not have access to your link - Access denied

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedYamlLoader.java
```diff
@@ -165,16 +165,39 @@ private static Node get(Node node) {
...
+																						// 10)
+						"|[-+]?(?:[0-9][0-9_]*)(?:[eE][-+]?[0-9]+)" + // (base 10,
+																		// scientific
+																		// notation
+																		// without .)
+						"|[-+]?\\.[0-9_]+(?:[eE][-+]?[0-9]+)?" + // (base 10, starting
+																	// with .)
+						"|[-+]?\\.(?:inf|Inf|INF)" + ")$");
+
+		public static final Pattern SPRING_INT = Pattern.compile("^(?:" + "|[-+]?(?:0|[1-9][0-9_]*)" + // (base
```
robert-gdv, line 186: Shouldn't this (and hex below) be limited in length?

wilkinsona:
Thanks for the proposal, @asomov. I'd like to make sure I understand why this is a better solution from your perspective. I'm guessing it's because you would prefer us to override `addImplicitResolvers` rather than `addImplicitResolver`? From our perspective, the first impression of what's proposed here is that it's a step backwards. There's quite a lot more code that we need to maintain and it feels like low-level details of YAML parsing which definitely isn't an area of expertise for us.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedYamlLoader.java
asomov: The spec does not put any restriction. The restriction comes from CVE which try to send invalid data and report an issue.

asomov:
@wilkinsona 

- it is robust. It does not depend on the SnakeYAML version
- it removes useless types (octals, YES as boolean etc)
- it makes Regular Expression smaller, which means quicker to apply
- it makes it possible to deviate from the general parser if needed (general parser is not for configuration).

philwebb:
Thanks very much for the PR @asomov. Although 724f9ebf7107d99ac9bae4204b4f7c6444a49e03 is not ideal, out current `main` branch no longer needs that check and [doesn't look too bad](https://github.com/spring-projects/spring-boot/blob/a8168515829c779f2436117b3c53f1315b1d6715/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedYamlLoader.java#L169).

I'm a little concerned that having Spring specific regular expressions might break back-compatibility. For example, users might be relying on "yes/no" for boolean values. I think we should stick with our current approach for now and see how things go. If we need to adapt to more `Resolver` changes in the future, we can always reconsider and re-open the PR.

Thanks again anyway for your efforts here and for maintaining snakeyaml!

PR #32476, opened 2022-09-22
Remove redundant @ExtendWith(SpringExtension.class) from MyJmxTests example

1993heqiang:
`@SpringBootTest` already contains `@ExtendWith(SpringExtension.class)`.


snicoll:
@1993heqiang thanks again.

PR #32480, opened 2022-09-22
MDC & Baggage AutoConfiguration for Micrometer Tracing with Brave and OpenTelemetry

marcingrzejszczak:
# Rationale

In order to provide the users with the basic tracing observability story we need the following, additional features

* [W3C](https://www.w3.org/TR/trace-context/) propagation type as the default propagation option (there would be 2 options, W3C and B3)
* Baggage support (remote fields only (these are passed over the wire) - no local field support, no tagged field support)
  * Correlation fields support (automatically ads to MDC given baggage entries)
* MDC updating for OTel and Brave (we will NOT modify the logging format but we WILL prepare the MDC so that the logging format would pick those entries)

## Conditionals

* If Micrometer Tracing IS NOT on the classpath
  * For Brave 
    * [Context-Propagation] We will setup Brave B3 Propagation as default (there's no W3C support out of the box in Brave)
  * For OTel
    * [Baggage] [Context-Propagation] We will setup W3C Propagation with Baggage as default
    * [Context-Propagation] We will setup W3C Propagation without Baggage when Baggage disabled via property
    * [Context-Propagation] We will setup B3 Propagation without Baggage when B3 propagation property turned on (there's no option to have baggage with B3 without Micrometer Tracing)
* If Micrometer Tracing IS on the classpath
  * For Brave
    * If baggage is not opt-out (management.tracing.baggage.enabled)      
      * [Context-Propagation] [Baggage] By default we setup W3C context propagation with Baggage
      * [Context-Propagation] [Baggage] We can switch to B3 context propagation (via a property) with Baggage
      * [Baggage] If provided up-front via properties, we will set remote baggage entries
      * [Baggage] [MDC] If MDC is on the classpath 
        * We will setup injecting correlated baggage entries to MDC (those need to be provided up-front)
    * If baggage is opt-out      
      * [Context-Propagation] By default we setup W3C context propagation without Baggage (that will be possible with Tracing snapshots - https://github.com/micrometer-metrics/tracing/pull/72)
      * [Context-Propagation] We can switch to B3 context propagation (via a property) 
    * [MDC] If MDC is on the classpath 
      * We will setup injecting spanId / traceId to MDC context 
  * For OTel:
    * [Baggage] [Context-Propagation] We will setup W3C Propagation with Baggage as default
    * [Context-Propagation] We will setup W3C Propagation without Baggage when Baggage disabled via property
    * [Baggage] [Context-Propagation] We will setup B3 Propagation with Baggage when B3 propagation property turned on
[...]

PR #32214, opened 2022-09-01
MDC auto-configuration for Brave and OpenTelemetry

marcingrzejszczak:
A sample with a test that proves that things are working fine [mdc-observability.zip](https://github.com/spring-projects/spring-boot/files/9471155/mdc-observability.zip)

PR #32487, opened 2022-09-23
Context Propagation via W3C and B3 for OTel and Brave 

marcingrzejszczak:
# Rationale

In order to provide the users with the basic tracing observability story we need the following, additional features

* [W3C](https://www.w3.org/TR/trace-context/) propagation type as the default propagation option (there would be 2 options, W3C and B3)

## Conditionals

* If Micrometer Tracing IS NOT on the classpath
  * For Brave 
    * [Context-Propagation] We will setup Brave B3 Propagation as default (there's no W3C support out of the box in Brave)
  * For OTel
    * [Context-Propagation] We will setup W3C Propagation without Baggage
    * [Context-Propagation] We will setup B3 Propagation without Baggage when B3 propagation property turned on
* If Micrometer Tracing IS on the classpath
  * For Brave
      * [Context-Propagation] By default we setup W3C context propagation without Baggage (that will be possible with Tracing snapshots - https://github.com/micrometer-metrics/tracing/pull/72)
      * [Context-Propagation] We can switch to B3 context propagation (via a property) 
  * For OTel:
    * [Context-Propagation] We will setup W3C Propagation without Baggage
    * [Context-Propagation] We will setup B3 Propagation without Baggage when B3 propagation property turned on
  

superseeds https://github.com/spring-projects/spring-boot/pull/32214
split from https://github.com/spring-projects/spring-boot/pull/32480


wilkinsona:
I got confused by the separation of the 3 PRs and processed this all in https://github.com/spring-projects/spring-boot/pull/32480 in the end. Sorry.

PR #32489, opened 2022-09-23
Polish JavaVersionTests.currentJavaVersionNineteen()

izeye:
This PR polishes `JavaVersionTests.currentJavaVersionNineteen()` by using the `JRE.JAVA_19` enum value.


snicoll:
Thanks Johnny. We won't be able to backport this as we're not using JUnit 5.9.1 in maintenance branches.

PR #32504, opened 2022-09-26
Avoid using JUnit 4 assertions

dreis2211:
Hi,

this PR removes usages of the old JUnit 4 assertions in two Kotlin tests. I guess the checkstyle rule to prevent that doesn't work on the Kotlin files.

Cheers,
Christoph


philwebb:
Thanks @dreis2211!

PR #32505, opened 2022-09-26
CouchbaseReactiveHealthIndicator uses blocking API to relieve the cluster diagnostics

micopiira:
Title pretty much says it all, but 
use non-blocking `cluster.reactive().diagnostics()`
instead of blocking `cluster.diagnostics()` call


snicoll:
@micopiira thank you for making your first contribution to Spring Boot.

Issue #32185, opened 2022-08-29
Multi-Document-Separators "#---" comment checks do not account for '!' prefixed comments


PaulKlumpp:
Using Spring Boot 2.7.3. The docs for Spring Boot 2.7.3 say on  https://docs.spring.io/spring-boot/docs/2.7.3/reference/htmlsingle/#features.external-config.files.multi-document

> The lines immediately before and after the separator must not be comments.

Using IDEs like IntelliJ, often, blank lines get removed. It is a setting for "Code Style" on "properties" files.

When the IDE removed a blank line, it happened. A comment from the line above got directly placed in front of the document separator line containing "#---". 

Like this, the multi document was not recognized anymore and led to problems like a misconfiguration of the application.

Sure, we will activate the Code Style Feature "Keep blank lines" for "properties", but still... Since comments should be perfectly fine in a property-file, it feels like a parser bug that it isn't allowed to have a comment line in front of _and_ after the "#---" line.

Please make it valid to have a multi-document like the following:

* When the "testing" profile is active, myProp=abcdef should be set.
* When the "testing" profile is not active, myProp should not be set.

```
#======= eye-sight-separator =======
#---
spring.config.activate.on-profile=testing
myProp=abcdef
```
With 2.7.3, when we did not activate the "testing" profile, myProp=abcdef was set. Ouch.

At the very moment, only the following works correctly:

```
#======= eye-sight-separator =======

#---
spring.config.activate.on-profile=testing
myProp=abcdef
```

To add a little more of an info..
I think the problem might be in here: https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/OriginTrackedPropertiesLoader.java

philwebb:
Some history about why we added this restriction can be found in #22963. If we remove it then we're likely to break folks that have comments such as:

```
#----
# Some comment
#----
```

In this example, the user doesn't want `#---` to be treated as a separator.

`!` is also a valid comment line so I wonder if we can extend our existing support to also include `!---` and change the rule so that the line before must not have the same comment prefix. In other words, we'd allow:

```
#======= eye-sight-separator =======
!---
spring.config.activate.on-profile=testing
myProp=abcdef
```

I think the existing code only considers `#` when checking the previous line, so you might already be able to do this:

```
!======= eye-sight-separator =======
#---
spring.config.activate.on-profile=testing
myProp=abcdef
```

[...]

PR #32521, opened 2022-09-27
Multi-Document-Separators "#---" comment checks do not account for '!' prefixed comments

terminux:
* Support `!` as prefix for document separator
* Allow lines before document separator to have different comment prefixes

Closes gh-32185


philwebb:
Thanks @terminux

PR #32534, opened 2022-09-29
Use ByteArrayOutputStream.toString where possible

dreis2211:
Hi,

the method was introduced in Java 10 and is slightly faster (because it avoids unnecessary array copies), but more importantly it is a little more concise ;-)

Cheers,
Christoph


philwebb:
Ohh nice! I like these little changes they've been making to the APIs

PR #32545, opened 2022-09-29
Removed @Primary from the default ObjectMapper bean

Erates:
The method I updated provides an ObjectMapper bean that can be injected elsewhere. But the method is annotated with `@Primary` and `@ConditionalOnMissingBean`. I think these two should never be used together.
I removed the `@Primary` annotation on the method.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonAutoConfiguration.java
@@ -51,7 +51,6 @@ import org.springframework.boot.jackson.JsonMixinModule;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
-import org.springframework.context.annotation.Primary;
 import org.springframework.context.annotation.Scope;
 import org.springframework.core.Ordered;
 import org.springframework.http.converter.json.Jackson2ObjectMapperBuilder;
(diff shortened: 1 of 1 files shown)


bclozel:
> @Primary and @ConditionalOnMissingBean. I think these two should never be used together.

Applications can still use `ObjectProvider` or [qualifiers](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-autowired-annotation-qualifiers) to get a specific bean, even if primary is being used.

I'm sorry but we need to decline this change as removing this now would break existing applications relying on this behavior.

Erates:
> > @primary and @ConditionalOnMissingBean. I think these two should never be used together.
> 
> Applications can still use `ObjectProvider` or [qualifiers](https://docs.spring.io/spring-framework/docs/current/reference/html/core.html#beans-autowired-annotation-qualifiers) to get a specific bean, even if primary is being used.
> 
> I'm sorry but we need to decline this change as removing this now would break existing applications relying on this behavior.

@bclozel 
I agree for Spring Boot 2.x, but can't this behaviour change be implemented in 3.x?

bclozel:
We're trying to make the upgrade experience as smooth as possible and right now I don't think this change is worth it given its potential disruption and the lack of value for developers.

If we remove that `@Primary` annotation, how would developers get the `ObjectMapper` they're customizing through configuration properties?

Erates:
In my opinion it would make it easier to the developers. Now when I need to change the behaviour of the `ObjectMapper` I always have to add `@Primary` or use the `@Qualifier` like you said.
I think it will be easier for the developer that he can inject an `ObjectMapper` and not configure it yourself if the default is fine, or create a bean yourself if the default implementation is not OK.

bclozel:
When the behavior of the `ObjectMapper` needs customizing, [Spring Boot is providing many ways to contribute and configure the behavior](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.json.jackson), hopefully this should cover quite a lot of cases.

Erates:
Ok I see.

Thanks for looking into it though!

PR #32554, opened 2022-09-30
Rework Spring Session auto-configuration to use customizers

vpavic:
This PR reworks Spring Session auto-configuration to avoid extending Spring Session's configuration classes. Instead, those configuration classes are now imported and customizations are applied using dedicated `(Reactive)SessionRepositoryCustomizer` beans.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/HazelcastSessionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/HazelcastSessionConfiguration.java
@@ -16,18 +16,19 @@
 
 package org.springframework.boot.autoconfigure.session;
 
-import java.time.Duration;
-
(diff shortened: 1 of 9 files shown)


wilkinsona:
On hold until we're ready to move to snapshots.

vpavic:
Is there a rough estimate when the move to snapshots is expected to be unblocked?

I'm asking because a lot of Spring Session's configuration infrastructure is set up the way it currently is because of reliance of Spring Boot's auto-configuration on inheritance, and the changes proposed here would allow Spring Session to evolve its configuration infrastructure. Having in mind the RCs are scheduled in just a little over 2 weeks from now, I'm trying to get a sense of the time we'll have to explore further improvements on Spring Session side.

wilkinsona:
We're trying to figure that out at the moment. Right now, it's looking like Boot will switch to snapshots of all the Spring projects on which is depends on Thursday (6 October).

vpavic:
Thanks. If the switch (and hopefully merge of this PR) happens by the end of this week, that would be great.

Issue #22405, opened 2020-07-20
Make it easier to apply customisations to the SpringApplication created by @SpringBootTest


wilkinsona:
When creating a `SpringApplication` in their application's main method, users can apply customizations to meet their needs. When these same customisations are also needed in their application's tests, reusing the same customizations isn't particularly easy. Without sub-classing `SpringBootContextLoader`, you can't get hold of the `SpringApplication` without jumping through some hoops. It's possible via a listener and the `ApplicationEnvironmentPreparedEvent` but that feels harder than it should.

Related: https://github.com/spring-projects/spring-boot/issues/22379, https://github.com/spring-projects/spring-boot/issues/15077

snicoll:
This looks like what we're doing for the war-based deployment stuff. What if your Spring Boot Application could implement an additional interface that provides a hook point with `SpringApplicationBuilder`? When this is present the callback would be invoked to customize the application. The added-benefit is that `SpringBootContextLoader` would do too when running an integration test. 

The downside is that an existing application class would have to be rearranged a bit to move the customization of the builder in a dedicated method. That sounds like reasonable for those who care about such customizations.

philwebb:
One possible idea we discussed was something like this:

```java
public interface SpringApplicationCustomizer {
    void customize(SpringApplication application)
}
```

```java
@SpringApplication(customizer=Foo.class)
public class MyApp {
}
```

OLibutzki:
As #24583 got closed as duplicate I would like to expand this issue's scope to support context hierarchies.

wilkinsona:
@OLibutzki Any customizations that are made to a `SpringApplication`, either directly or via `SpringApplicationBuilder`, are in scope for this issue. That includes context hierarchies.

PR #32568, opened 2022-10-02
Add Javadoc since for UseMainMethod

izeye:
This PR adds Javadoc `@since` tag for `SpringBootTest.UseMainMethod`.

See gh-22405

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootTest.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootTest.java
@@ -187,6 +187,7 @@ public @interface SpringBootTest {
 	 * Enumeration of how the main method of the
 	 * {@link SpringBootConfiguration @SpringBootConfiguration}-annotated class is used
 	 * when creating and running the {@link SpringApplication} under test.
+	 * @since 3.0.0
 	 */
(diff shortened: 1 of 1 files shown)


izeye:
I pushed a8eb298 as the Spring Java Format doesn't seem to support this case yet.

wilkinsona:
Thanks, @izeye. Without a8eb298, I see a failure like this:

```
> Task :spring-boot-project:spring-boot-test:checkstyleMain
[ant:checkstyle] [ERROR] /Users/awilkinson/dev/spring-projects/spring-boot/main/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootTest.java:191:19: Javadoc @since tag should not be used on private classes. [SpringJavadoc]
```

Do you see the same? Curiously, this should work due to the changes made for https://github.com/spring-io/spring-javaformat/issues/132.

wilkinsona:
And, of course, shortly after making that comment I realised what's going on. It's the difference between `interface` and `@interface`, that is to say an interface definition and an annotation definition.

philwebb:
See https://github.com/spring-io/spring-javaformat/issues/344

philwebb:
If we wait for a javaformat release we can amend the PR to not need the checkstyle exclusion.

wilkinsona:
Thank you, @izeye!

PR #32573, opened 2022-10-03
Allow JMX Registration Policy to be set

mdeinum:
Currently the `JmxAutoconfiguration` sets the `RegistrationPolicy` to `FAIL_ON_EXISTING`. However there can be situations where it might be useful to set it to something else like `IGNORE_EXISTING`. 

This could help to workaround the issue with both HikariCP and Spring Boot registering JMX beans. 

See https://stackoverflow.com/questions/73896176/spring-boot-hikari-config-not-loaded/73896695#73896695 (for my recommended solution). 

And [the issue from HikariCP](https://github.com/brettwooldridge/HikariCP/issues/342) with several other proposed solutions/workaround. 

The drawback of those solutions, imho, is that the disable (or at least partially) the auto-configuration from Spring Boot.


mhalbritter:
Thank you!

PR #32578, opened 2022-10-04
Provide extensions that remove the need for the log4j-spring-boot log4j2 module

rgoers:
This change merges the contents of the log4j-spring-boot module into Spring Boot, effectively making Log4j 2 on par with Logback and eliminating the need for the log4j-spring-boot module.

I was able to verify this with a sample project I have.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
@@ -16,14 +16,18 @@
 
 package org.springframework.boot.logging.log4j2;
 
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks very much @rgoers, and thanks for your patience whilst we go to this. I've split the PR into a number of distinct issues so that our automated changelog is correct.

I made a few amendments to some of the functionality, namely:

* I removed the ability in `StrLookup` to resolve `profiles.active` and `profiles.enabled`. This simplifies quite a bit of code and I can't think of a good reason why users should need to resolve the profiles.
* A few of the extensions now fail hard if the `Environment` isn't available when they're used
* URL based lookups must resolve a config and if they don't the application will fail to start

I've also added a new section in the reference guide to explain the features.

Thanks again for the contribution and for all your hard work on Log4j2!

Issue #32638, opened 2022-10-07
Documentation Inconsistency - Externalized Config, Using @ConfigurationProperties


eurythmia:
Live Documentation Section: [7.2.8. Type-safe Configuration Properties -> Using @ConfigurationProperties-annotated types](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.external-config.typesafe-configuration-properties.using-annotated-types)

### Details

Code snippet in the linked section uses data type `SomeProperties`. Other subsections of [7.2.8](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#features.external-config.typesafe-configuration-properties) (both before and after the mentioned subsection) define a properties class called `MyProperties` (first example snippet here):

```java
@ConfigurationProperties("my.service")
public class MyProperties {

    private boolean enabled;

    private InetAddress remoteAddress;

    private final Security security = new Security();

    // getters / setters...

    public static class Security {

        private String username;

        private String password;

        private List<String> roles = new ArrayList<>(Collections.singleton("USER"));

        // getters / setters...

    }

}
```

### Issue

The properties type given in the subsection code example (provided below) appears to exhibit the same structure as the above example, but is a data type not yet seen (`SomeProperties`). This reduces coherency of the section. When reading, I had to go back to check if I had missed something.

```java
import org.springframework.stereotype.Service;

@Service
public class MyService {

    private final SomeProperties properties;

    public MyService(SomeProperties properties) {
        this.properties = properties;
    }

    public void openConnection() {
        Server server = new Server(this.properties.getRemoteAddress());
        server.start();
        // ...
    }

    // ...

}
```

### Proposed Solution

```sh
git clone git@github.com:spring-projects/spring-boot.git
cd spring-boot/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/features/externalconfig/typesafeconfigurationproperties/usingannotatedtypes/

# replace `SomeProperties` with `MyProperties` in both java source files (and remove sed backups)
sed -i.bkp 's/SomeProperties /MyProperties /g' MyService.java SomeProperties.java && rm *.java.bkp

# rename SomeProperties.java to MyProperties.java
git mv {Some,My}Properties.java
```

[...]

PR #32644, opened 2022-10-09
Rename SomeProperties class to MyProperties

eurythmia:
Rename the SomeProperties class so that it is consistent with other examples given within the Type-safe Configuration Properties. This ensures that readers can intuitively draw the conclusion tha the MyProperties class is annotated with ConfigurationProperties.

Fixes gh-32638


snicoll:
Thanks for the PR @eurythmia. We can't merge this in `2.5.x` though I am afraid as it is [out of OSS support](https://spring.io/projects/spring-boot#support). When looking into this, there are further confusing reference to `SomeProperties` in the doc itself that should be revisited.

There's no need for you to do anything btw, we'll review this as part of merging.

eurythmia:
Thanks for taking a look and handling the admin work! I did my best to follow the wiki, but never thought to check what's currently in supported status.

I chose 2.5.x because it was the earliest branch exhibiting the specific issue I saw, forward merging or cherry-picking to 2.6.x and 2.7.x should be relatively straightforward (2.7.x is where I originally noticed this).

I can take a look at the section to update the *.adoc, and the kotlin example with this change if you'd like (perhaps targeting 2.6.x).
I also noticed a typo in my commit message after pushing and was reticent to rewrite the history on a PR (`'s/tha /that /'`) with a commit amendment, so that should definitely be a candidate for commit polish :-)

snicoll:
@eurythmia thank you for making your first contribution to Spring Boot.

Issue #8788, opened 2017-03-31
Welcome page forwards to the wrong location if static-path-pattern has been customized


wilkinsona:
We forward to `index.html` which assumes that the static path pattern is `/**`. If it's been customised to `/foo/**` them we'll still forward to `index.html` and a 404 will result. We should avoid making this faulty forward.

There's an argument to be made that we should forward to `foo/index.html`. However, that makes things complicated when relative URLs in the `index.html` page are considered. As it's a forward rather than a redirect, the browser will resolve those URLs using `/` whereas they should be resolved using `/foo/`. We don't want to change it to a redirect so the safest option is to not set up the welcome page when the static path pattern has been customised.

PR #32745, opened 2022-10-16
WelcomePageHandlerMapping have some bugs in the code, should add customization

chenjjiaa:
Hello SpringBoot's developers 😀

description :   
In the configuration file, if both 【`spring.mvc.static-path-pattern`】 and 【`spring.web.resources.static-locations`】 are configured, the Welcome Page ability is invalidated

cause : WelcomePageHandlerMapping's construction method of judging condition :
            `if (welcomePage != null && "/**".equals(staticPathPattern))`

-----------------------------------------------------
Among them :  `"/**".equals(staticPathPattern) `
I think can be customized by users
If this code is removed, the Welcome Page can be used
So I think this code can be changed by the user

-----------------------------------------------------
I think there are two ways.
1、Add `@ConditionalOnMissingBean` on WelcomePageHandlerMapping in EnableWebMvcConfiguration
2、Just remove 【 `&& "/**".equals(staticPathPattern)` 】in WelcomePageHandlerMapping's construction

Now I'm going to take approach one :)

I'm glad to participate in the contribution of the project ! 
Hope my code can be incorporated into the project ❤❤❤

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/WebMvcAutoConfiguration.java
@@ -430,6 +430,7 @@ public class WebMvcAutoConfiguration {
 		}
 
 		@Bean
+		@ConditionalOnMissingBean(name = "welcomePageHandlerMapping")
 		public WelcomePageHandlerMapping welcomePageHandlerMapping(ApplicationContext applicationContext,
 				FormattingConversionService mvcConversionService, ResourceUrlProvider mvcResourceUrlProvider) {
 			WelcomePageHandlerMapping welcomePageHandlerMapping = new WelcomePageHandlerMapping(


chenjjiaa (commented):
Check completed. 
Pending approval :)

chenjjiaa:
Hello, when will the classification be completed? 👋👋👋

wilkinsona:
Thanks for your patience, @chenjjiaa. The team is focused on our 3.0 release at the moment. We will triage this pull request as soon as we can

wilkinsona:
Thanks for the proposal, @chenjjiaa, and for your patience while we found time to review it.

`WelcomePageHandlerMapping` is package-private and `final`. That means that it can't be caused to back off by defining a custom bean of the same type. I guess this is why you've defined a name in your proposed condition. We try to avoid "magic" bean names so I'm not keen on doing this either.

The check for `staticPathPattern` being `/**` was added to fix https://github.com/spring-projects/spring-boot/issues/8788. We cannot remove it without regressing that fix.

Unfortunately, for the reasons above, I don't think we should make either of the changes proposed here. Thanks anyway for the PR.

PR #32812, opened 2022-10-20
Fix typos in logging.adoc

jamessoun93:
Something seems to be off with this part of the doc so I gave it a try.

According to the doc, I can create my own configuration file and include logback configurations provided by Spring Boot, not from my own configuration.

Please correct me if I'm wrong.


snicoll:
@jamessoun93 good catch.

snicoll:
@jamessoun93 thank you for making your first contribution to Spring Boot.

PR #32822, opened 2022-10-20
Provide a workaround to allow multiple @ImportRuntimeHints

sdeleuze:
This commit provides a workaround for spring-projects/spring-framework#29361 that should cover Actuator web end point use cases (exception Cloud Foundry one not possible due to package access issue).


philwebb:
I referenced the wrong commit, but merged with 7ddd1e2bed94caaabb2d09fff5d8d073148decd5

PR #32844, opened 2022-10-21
Wavefront application tags differ from those used in a Spring Boot 2.x application

oppegard:
This PR fleshes out the use of Wavefront [Application Tags](https://docs.wavefront.com/trace_data_details.html#application-tags), used in Wavefront to organize a service within a larger distributed application.

- **application** (required): this is the name of a distributed application comprised of multiple microservices (not to be confused with an individual service, below).
- **service** (required): this is the name of the current service that is part of the larger application. It will use `spring.application.name` if available.
- **cluster** (optional)
- **shard** (optional)

This PR add two new properties: 
- `management.wavefront.tracing.cluster-name`
- `management.wavefront.tracing.shard-name`

In wavefront-spring-boot [these properties](https://github.com/wavefrontHQ/wavefront-spring-boot/blob/master/wavefront-spring-boot/src/main/java/com/wavefront/spring/autoconfigure/ApplicationTagsFactory.java#L22) were `wavefront.application.cluster` and `wavefront.application.shard`. I'm not sure if there's a way to migrate users of the old properties (same goes for `wavefront.application.name` and `wavefront.application.service`).

This PR also adds Application Tags to metrics exported by instances of `WavefrontMeterRegistry`, replacing [functionality](https://github.com/wavefrontHQ/wavefront-spring-boot/blob/master/wavefront-spring-boot/src/main/java/com/wavefront/spring/autoconfigure/WavefrontMetricsConfiguration.java#L51-L56) from wavefront-spring-boot.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
@@ -77,7 +77,6 @@ public class WavefrontMetricsExportAutoConfiguration {
 
 	@Bean
 	@ConditionalOnBean(ApplicationTags.class)
-	@ConditionalOnClass(MeterRegistryCustomizer.class)
 	MeterRegistryCustomizer<WavefrontMeterRegistry> applicationTagsCustomizer(ApplicationTags applicationTags) {
 		Tags commonTags = Tags.of(applicationTags.toPointTags().entrySet().stream()
 				.map(WavefrontMetricsExportAutoConfiguration::asTag).toList());
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
@@ -16,12 +16,18 @@
 
 package org.springframework.boot.actuate.autoconfigure.metrics.export.wavefront;
 
+import java.util.Map;
+
 import com.wavefront.sdk.common.WavefrontSender;
+import com.wavefront.sdk.common.application.ApplicationTags;
 import io.micrometer.core.instrument.Clock;
+import io.micrometer.core.instrument.Tag;
+import io.micrometer.core.instrument.Tags;
 import io.micrometer.wavefront.WavefrontConfig;
 import io.micrometer.wavefront.WavefrontMeterRegistry;
 
 import org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegistryAutoConfiguration;
+import org.springframework.boot.actuate.autoconfigure.metrics.MeterRegistryCustomizer;
 import org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration;
 import org.springframework.boot.actuate.autoconfigure.metrics.export.ConditionalOnEnabledMetricsExport;
 import org.springframework.boot.actuate.autoconfigure.metrics.export.simple.SimpleMetricsExportAutoConfiguration;
@@ -42,6 +48,7 @@ import org.springframework.context.annotation.Import;
  * @author Jon Schneider
  * @author Artsiom Yudovin
  * @author Stephane Nicoll
+ * @author Glenn Oppegard
  * @since 2.0.0
  */
 @AutoConfiguration(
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
```diff
@@ -68,4 +75,17 @@ public WavefrontMeterRegistry wavefrontMeterRegistry(WavefrontConfig wavefrontCo
 		return WavefrontMeterRegistry.builder(wavefrontConfig).clock(clock).wavefrontSender(wavefrontSender).build();
 	}
 
+	@Bean
+	@ConditionalOnBean(ApplicationTags.class)
+	@ConditionalOnClass(MeterRegistryCustomizer.class)
+	MeterRegistryCustomizer<WavefrontMeterRegistry> applicationTagsCustomizer(ApplicationTags applicationTags) {
```
oppegard, line 81: Forgive my Spring ignorance, but will this allow for multiple `MeterRegistryCustomizer<WavefrontMeterRegistry>` beans in the event a user wants to provide customizations in addition to this one?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/wavefront/WavefrontProperties.java
```diff
@@ -261,15 +264,28 @@ public void setBatchSize(Integer batchSize) {
 	public static class Tracing {
 
 		/**
-		 * Application name. Defaults to 'spring.application.name'.
+		 * Application name used in {@link ApplicationTags}. Defaults to
```
jonatan-ivanov, line 267: Just like in the previous case, I would call out that this is the Wavefront application, service, etc. name.

jonatan-ivanov (changes requested):

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/wavefront/WavefrontMetricsExportAutoConfiguration.java
oppegard: I believe multiple Customizers are allowed. Here's where they're streamed and `.customize()` is invoked: https://github.com/spring-projects/spring-boot/blob/e0b67889a8f70228227e06732f9973a1e9bc1a86/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MeterRegistryConfigurer.java#L74-L75

However, since my code has an `@Bean` annotation, I figured that means there can only be a singleton of `MeterRegistryCustomizer<WavefrontMeterRegistry>`.

wilkinsona: Multiple customisers are allowed. The singleton scope applies to each specific bean, not to each type of bean. In other words, your `applicationTagsCustomizer` will be a singleton but there could be any number of other beans that implement `MeterRegistryCustomizer<WavefrontMeterRegistry>` and there will be a single instance of each of those as well.

oppegard: @wilkinsona thanks so much for correcting my understanding!

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/wavefront/WavefrontTracingAutoConfiguration.java
```diff
@@ -58,20 +59,29 @@
 @ConditionalOnEnabledTracing
 public class WavefrontTracingAutoConfiguration {
 
+	private static final String DEFAULT_APPLICATION_NAME = "unnamed_application";
+
 	/**
-	 * Default value for application name if {@code spring.application.name} is not set.
+	 * Default value for service name if {@code spring.application.name} is not set.
```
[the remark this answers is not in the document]
oppegard: Along with renaming variable names, in the docs I added links to https://docs.wavefront.com/trace_data_details.html#application-tags. Let me know if you think it's too much or needs tweaking.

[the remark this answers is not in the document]
jonatan-ivanov: 👍🏼 

I think this is out of scope for this PR, but I find those docs equally confusing, e.g.:
>application - Name that identifies the application, for example, beachshirts. All microservices in the same application should use the same application name.

To me a JVM process is an application. A microservice is also an app, a monolithic webservice too as well as bots, mobile-, and desktop apps, etc.
Wavefront makes the (I think wrong) assumption that every app is a webservice (actually microservices) and this makes the understanding of these terms hard.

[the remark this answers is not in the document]
oppegard: Thanks for the feedback. Changing Wavefront's logical hierarchy in distributed tracing would be a big effort, but perhaps we can clarify the docs with more examples.

Issue #32721, opened 2022-10-13
Dependency management for Selenium is incomplete


wilkinsona:
At least `selenium-chromium-driver` is missing. Others may be missing too.

philwebb:
Closing in favor of PR #32861. Thanks @izeye

wilkinsona:
Turns out we're missing some dependencies in 2.6.x as well.

wilkinsona:
I've used this issue for 2.6.x. #32861 and its forward merge will take care of 2.7.x and main.

PR #32861, opened 2022-10-24
 Dependency management for Selenium is incomplete

izeye:
This PR changes to provide dependency management for missing Selenium modules.

Closes gh-32721


wilkinsona:
Thanks very much, @izeye.

PR #32920, opened 2022-10-30
Upgrade CI image update detection to Ubuntu Jammy

dreis2211:
Hi,

this PR updates the `detect-ubuntu-image-updates.sh` script to check for updates on `Jammy` rather than `Focal`. The former is the LTS version since April 2022.

I didn't update the CI images on purpose since the script should pick this up, but let me know if I should directly do this.

Cheers,
Christoph


wilkinsona:
Thanks, @dreis2211.

PR #32931, opened 2022-10-31
[Bug][Document] Fix a minor bug in an example in oauth2 doc

EricGao888:
* Fix a minor bug in a code example in `OAuth2` doc. There is a typo in the code snippet shown in the screenshot below, `authorization-code` should be `authorization_code`. This minor bug may cause a developer to spend quite some time wondering why his / her `OAuth2` configuration does not work. 

image


wilkinsona:
@EricGao888 Thanks very much for making your first contribution to Spring Boot.

Issue #32950, opened 2022-11-01
Auto-configure BatchInterceptor on KafkaListenerContainerFactory


ThomasKasene:
I was hoping to re-use as much as possible of the default Kafka auto-configuration, but I need to use a `BatchInterceptor`, and the default `ConcurrentKafkaListenerContainerFactoryConfigurer` doesn't appear to respect the `BatchInterceptor` bean I've created; it only cares about a `RecordInterceptor` bean, if one exists.

I've gotten around it for now by calling the `setBatchInterceptor` method on the `AbstractKafkaListenerContainerFactory` beans directly, but I'd much rather avoid even that if I could.

If the team agrees this is a good idea, I'd be happy to look into submitting a PR for it. But in that case, please advise whether any tests need to be expanded, and also, which ones.

wilkinsona:
It sounds like a good idea to me. Thanks for the offer of a PR. Hopefully the changes in https://github.com/spring-projects/spring-boot/commit/eeda12bd58c2b49e6a053fc545eade7168c0ff1f, which added support for a `RecordInterceptor`, will show the sorts of changes that are needed to support `BatchInterceptor` as well.

wilkinsona:
Closing in favor of #32951. Thanks for the PR, @ThomasKasene.

PR #32951, opened 2022-11-01
Auto-configure BatchInterceptor on the default ConcurrentKafkaListenerContainerFactory

ThomasKasene:
As discussed in #32950, this auto-configures the default `ConcurrentKafkaListenerContainerFactory` with the single candidate `BatchInterceptor`, if one exists.


mhalbritter:
Thank you!

PR #32973, opened 2022-11-02
management.observations.http.server.requests.name is ignored when auto-configuring meter filters for maximum allowable uri tags

izeye:
This PR changes to handle custom observation name for HTTP server requests in max URI tag filters.


wilkinsona:
Thanks very much, @izeye.

PR #32997, opened 2022-11-04
Favor `xmlunit-assertj3` instead of `xmlunit-assertj`

scordio:
As an outcome of xmlunit/xmlunit#203, [`xmlunit-assertj3`](https://github.com/xmlunit/xmlunit/tree/main/xmlunit-assertj3) was created to better integrate with the AssertJ 3.x API (Java 8 based), while the existing [`xmlunit-assertj`](https://github.com/xmlunit/xmlunit/tree/main/xmlunit-assertj) remains compatible with AssertJ 2.x (Java 7 based and on maintenance mode).

Knowing that Spring Boot 2 was already requiring Java 8 as the minimum requirement, I think it would be a more sensible choice to include `xmlunit-assertj3`.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1620,7 +1620,7 @@ bom {
 	library("XmlUnit2", "2.9.0") {
 		group("org.xmlunit") {
 			modules = [
-				"xmlunit-assertj",
+				"xmlunit-assertj3",
 				"xmlunit-core",
 				"xmlunit-legacy",
 				"xmlunit-matchers",


wilkinsona:
Thanks, @scordio. Rather than replacing the dependency management for xmlunit-assertj, which will break people's builds when they upgrade, I think we should add dependency management for xmlunit-assertj3 alongside it. Looking at this also helped me to notice that dependency management for `xmlunit-jakarta-jaxb-impl` is also missing for Boot 3.0. I've taken care of things in #32998, #32999, and, #33000. Thanks again for brining this to our attention.

scordio:
Thanks for the detailed explanation, @wilkinsona. I approached this topic quite quickly and didn't give enough weight to the breaking change, let alone the possible backporting 🙂

PR #33013, opened 2022-11-05
Polish ImportCandidates

biergit:
As the class not only handles autoconfiguration imports I thought the code should reflect this more clearly.


biergit:
What do you think about providing this class in the Spring framework instead? Functionality previously provided by `SpringFactoriesLoader` is now only available by using Spring Boot. [Here](https://github.com/spring-io/initializr/blob/10d7e857c79e2d3e9bc436b1e5ab189a363c6e90/initializr-generator/src/main/java/io/spring/initializr/generator/project/ProjectGenerator.java#L125) is a case where the `ImportCandidates` functionality would be needed but there is no dependency on Spring Boot

philwebb:
I like it! Thanks @biergit

wilkinsona:
I think this change is worth making in 2.7.x as well.

PR #33080, opened 2022-11-09
Add reflection hints for Jackson's PropertyNamingStrategies

ueberfuhr:
In a Spring Boot app with MVC, we can configure Jackson's property naming strategy by the following application.yml:

```yaml
spring:
  jackson:
    property-naming-strategy: SNAKE_CASE
```

Unfortunately, this is read by Spring Boot Autoconfigure using reflection, and therefore, the static fields have to be registered in the case of a native build. Otherwise, we get an error message when running the app:

```bash
Caused by: java.lang.IllegalArgumentException: Constant named 'SNAKE_CASE' not found
  at org.springframework.util.Assert.notNull(Assert.java:219) ~[na:na]
  at org.springframework.boot.autoconfigure.jackson.JacksonAutoConfiguration$Jackson2ObjectMapperBuilderCustomizerConfiguration$StandardJackson2ObjectMapperBuilderCustomizer.configurePropertyNamingStrategyField(JacksonAutoConfiguration.java:287) ~[spring-features.exe:na]
```

So I added native hints for the constants, hopefully the correct way and in the correct project. (There is already a Hint Registrar existing in the `spring-boot` project, that registers serializer implementations. But I suggest to create this registrar in the `spring-boot-autoconfigure` project, because this is the location where the error occurs.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonRuntimeHints.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonRuntimeHints.java
@@ -1,0 +1,54 @@
+package org.springframework.boot.autoconfigure.jackson;
+
+import com.fasterxml.jackson.databind.PropertyNamingStrategies;
+import com.fasterxml.jackson.databind.PropertyNamingStrategy;
+import org.springframework.aot.hint.ReflectionHints;
+import org.springframework.aot.hint.RuntimeHints;
+import org.springframework.aot.hint.RuntimeHintsRegistrar;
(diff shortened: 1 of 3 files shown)


philwebb:
Good spot! Thanks very much for contributing the fix.

sbrannen:
> So I added native hints for the constants

Wouldn't you also need to register reflection hints for the constructors of those implementations?

See https://github.com/spring-projects/spring-framework/issues/29386#issuecomment-1308835391

wilkinsona:
No, I don't think so. We use reflection to get the value of a field declared on `PropertyNamingStrategies`. Those values are instances of a `PropertyNameStrategy` implementation. We don't instantiate the strategy as Jackson's already done so and it doesn't use reflection.

PR #33098, opened 2022-11-10
docfix: kafka streams auto start description typo

iamgd67:
remove "it creates"  from "By default, the streams managed by the `StreamBuilder` object it creates are started automatically."


philwebb:
Thanks @iamgd67!

PR #33099, opened 2022-11-10
Provide a configuration property for the observation patterns of Spring Integration components

artembilan:
Spring Integration can instrument its component with an `Observation` according to the `@EnableIntegrationManagement.observationPatterns` value

* Expose `spring.integration.management.observation-patterns` configuration property
* Propagate this property into an `@EnableIntegrationManagement` in the `IntegrationAutoConfiguration.IntegrationManagementConfiguration`
* Verify that property has an effect via `IntegrationAutoConfigurationTests.integrationManagementInstrumentedWithObservation()`

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/integration/IntegrationAutoConfiguration.java
@@ -209,7 +209,8 @@ public class IntegrationAutoConfiguration {
 
 		@Configuration(proxyBeanMethods = false)
 		@EnableIntegrationManagement(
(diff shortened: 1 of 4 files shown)


philwebb:
Considering this one a bug of omission, but we can't add it yet because we can't move to SNAPSHOTs before our next RC.

artembilan:
No. This is not a bug or omission. 
An observation instrumentation is a new feature in Spring Integration 6.0.
Yes, we can squeeze this for the next release for Spring Boot in two week.
Or you can just move it as a theme of next Spring Boot `3.1` (or `3.0.1` if any).

Thanks

philwebb:
By "bug of omission" I meant that it's a feature we should ideally have support for in the 3.0 release since the feature was added in Spring Integration 6. I guess you're right though, it's really an enhancement. I still think it might be worth adding to 3.0 if we can. It seems low risk.

artembilan:
Sounds good, Phil!
Sorry that I came up with this property a bit late.
Let's see if it can make it into `3.0`!

PR #33108, opened 2022-11-11
Polish

izeye:
This PR polishes a bit.


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/observation/web/reactive/ServerRequestObservationConventionAdapter.java
```diff
@@ -47,8 +47,7 @@ class ServerRequestObservationConventionAdapter implements ServerRequestObservat
 	}
 
 	ServerRequestObservationConventionAdapter(String name, List<WebFluxTagsContributor> contributors) {
-		this.name = name;
-		this.tagsProvider = new DefaultWebFluxTagsProvider(contributors);
+		this(name, new DefaultWebFluxTagsProvider(contributors));
```
RUBenGAMArrarodRiguEZ-ToMtOm, line 50: For me it's easier to read and trace if there are not nested calls in the arguments. Before each line was doing exactly one thing. Not sure what is recommended here for the code style, probably there are examples of both.

PR #33115, opened 2022-11-12
Use AntPathMatcher at AutoConfigurationImportSelector

meiblorn:
Spring Boot *AutoConfiguration-classes exclusion management is actually a hell. Thanks to the `spring-boot-autoconfigure` module which is suited to be a `global` module for everything.

The issue that has motivated me to do such pull request is all the things that have been pulled up when I added `spring-boot-starter-data-r2dbc` and `org.jooq:jooq` dependencies. Adding these dependencies automatically triggers creation of `ConnectionFactory`, `TransactionManager`, repository support released beans, `JOOQ`' related `DSL` and `Configuration` beans creation. **BUT I was needed only to have `ConnectionFactory`!**.

I don't want to manage transactions through Spring. And I don't want Spring to create JOOQ related beans — I am going to manage them right in the business logic code using `DSL.using(connectionFactory). ...` API.

So I started to exclude these auto configurations manually.
My `main method` with `@SpringAutoConfiguration` annotation started to getting bigger on the almost fresh project. At another project it already has 15 `exclusionName` rules !
```
@SpringBootApplication(
    excludeName = [
        "org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration",
        "org.springframework.boot.autoconfigure.r2dbc.R2dbcAutoConfiguration",
        "org.springframework.boot.autoconfigure.r2dbc.R2dbcTransactionManagerAutoConfiguration",
        "org.springframework.boot.autoconfigure.jooq.JooqAutoConfiguration",
    ]
)
class SpringAutoconfigurePracticeApplication

fun main(args: Array<String>) {
    runApplication<SpringAutoconfigurePracticeApplication>(*args)
}
```

Comparing to Spring, Micronaut doesn't have such problem because configuration classes ship with features inside the same jar. Hence, if you don't need some configuration to apply or even to exist — just remove the jar from the classpath. 

I suggest to allow glob patterns in `excludeName` parameter to make this block shorter:
```
@SpringBootApplication(
    excludeName = [
        "org.springframework.boot.autoconfigure.flyway", // valid because exact path
        "org.springframework.boot.autoconfigure.r2dbc.*", // valid glob
        "org.springframework.boot.autoconfigure.jooq.**", // valid glob
    ]
)
class SpringAutoconfigurePracticeApplication

fun main(args: Array<String>) {
    runApplication<SpringAutoconfigurePracticeApplication>(*args)
}
```

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationImportSelector.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurationImportSelector.java
@@ -55,9 +55,7 @@ import org.springframework.core.io.support.SpringFactoriesLoader;
 import org.springframework.core.type.AnnotationMetadata;
 import org.springframework.core.type.classreading.CachingMetadataReaderFactory;
 import org.springframework.core.type.classreading.MetadataReaderFactory;
-import org.springframework.util.Assert;
-import org.springframework.util.ClassUtils;
-import org.springframework.util.StringUtils;
+import org.springframework.util.*;
 
 /**
  * {@link DeferredImportSelector} to handle {@link EnableAutoConfiguration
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the proposal. I don't think we understand the problem well enough to evaluate a possible change. Let's use #33116 to gain that understanding.

Issue #33090, opened 2022-11-09
Default security filter dispatcher types are not aligned with Spring Security's defaults


marcusdacoregio:
https://github.com/spring-projects/spring-security/issues/11027 has changed the Spring Security default by applying authorization to every request.

Currently, to apply the `FilterChainProxy` to all dispatcher types using Spring Boot, one can do:
`spring.security.filter.dispatcher-types=request,async,error,forward,include`, where the default are just `request,async,error`.

Spring Boot should register the `FilterChainProxy` to all dispatcher types by default to better align with Spring Security defaults.

PR #33252, opened 2022-11-18
Update Spring Security filter dispatcher types docs to reflect change in default value

vpavic:
This aligns configuration metadata and documentation with new defaults for `spring.security.filter.dispatcher-types` property.

See #33090.


wilkinsona:
Thanks again, @vpavic.

PR #33270, opened 2022-11-18
Exclude IntComponentScanConfiguration from AOT

artembilan:
In AOT mode the component scan is done during build phase before we generate code for beans to be registered at runtime. 
Therefore, `IntegrationAutoConfiguration.IntegrationComponentScanConfiguration` has already done its job and its "void" bean is not needed to be exposed into runtime.

* The `IntegrationAutoConfiguration.IntegrationComponentScanConfiguration` implements now `BeanRegistrationAotProcessor` as the easiest way on the code to support bean exclusion from runtime


artembilan:
After discussion with @wilkinsona we came into a conclusion that this is not acceptable: it's making quite a big assumption about the behavior of the `IntegrationComponentScanRegistrar` which is really an implementation detail as far as Boot's concerned.

PR #33276, opened 2022-11-19
`ScheduledBeanLazyInitializationExcludeFilter` is auto-configured even when annotation-based scheduled has not been enabled

hpoettker:
This PR makes the exposure of `ScheduledBeanLazyInitializationExcludeFilter` conditional on the annotation `@EnableScheduling`.

When the property `spring.main.lazy-initialization` is set to `true`, the `LazyInitializationBeanFactoryPostProcessor` currently applies the `ScheduledBeanLazyInitializationExcludeFilter` to each bean definition. This is independent of whether the user has opted in to scheduling with `@EnableScheduling` or not.

The goal of the PR is to improve the start-up performance of applications that use lazy initialization but not scheduling. An example could be a short-lived Spring Batch app that contains multiple job beans but executes only a single job when started and exits afterwards. Such an app would not use scheduling and profits from not initializing all beans on start up.

I've assembled a toy application with a flame graph here: https://github.com/hpoettker/lazy-scheduler-demo

About 3% of the whole application run-time is spent on the `ScheduledBeanLazyInitializationExcludeFilter`.

The start-up time (as logged by the application) varies between runs. But on my machine it's about 0.65 seconds and reduces to 0.60 seconds when the `TaskSchedulingAutoConfiguration` is excluded.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskSchedulingAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskSchedulingAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2021 the original author or authors.
+ * Copyright 2012-2022 the original author or authors.
  *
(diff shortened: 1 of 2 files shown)


snicoll:
Thanks again @hpoettker

PR #33371, opened 2022-11-26
@SpringBootTest now throws a NullPointerException rather than a helpful IllegalStateException when @SpringBootConfiguration is not found

### LINKED ISSUE #33370: Spring Boot Test cannot  give a clear message but throw NPE now when @SpringBootConfiguration cannot be found
When we try spring boot test without ```@SpringBootConfiguration```,NPE exception will occur.
Before supporting AOT,it throws IllegalStateException with message:

```
Unable to find a @SpringBootConfiguration, you need to use "
				+ "@ContextConfiguration or @SpringBootTest(classes=...) with your test
```

but now it occurs exception like this:
```
java.lang.NullPointerException: Cannot invoke "java.lang.Class.getName()" because "found" is null

	at org.springframework.boot.test.context.SpringBootTestContextBootstrapper.findConfigurationClass(SpringBootTestContextBootstrapper.java:261)
```

I find that class is not checking null before setting aotTestAttributes :

```java
Class<?> found = new AnnotatedClassFinder(SpringBootConfiguration.class).findFromClass(testClass);
this.aotTestAttributes.setAttribute(propertyName, found.getName());
``` 

We can avoid this with a small change by checking whether class with ```@SpringBootConfiguration``` is null in advance.We can move checking code from ```getOrFindConfigurationClasses``` to ```findConfigurationClass``` before setting aotTestAttributes

lishangbu:
close #33370

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootTestContextBootstrapper.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/SpringBootTestContextBootstrapper.java
@@ -244,8 +244,6 @@ public class SpringBootTestContextBootstrapper extends DefaultTestContextBootstr
 			return classes;
 		}
 		Class<?> found = findConfigurationClass(mergedConfig.getTestClass());
-		Assert.state(found != null, "Unable to find a @SpringBootConfiguration, you need to use "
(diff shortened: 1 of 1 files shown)


mhalbritter:
Thank you and congratulations on your first contribution to Spring Boot!

Issue #33395, opened 2022-11-28
ehcache managed version in spring-boot-dependencies 3.0.x requires jakarta classifier


grafjo:
To use the managed version of ehcache with Spring Boot 3.0.x, it's required to use the classifier `jakarta`. Documentation says nothing about that classifier and it feels like the classifier can be removed and was fixing some issues with jakarta package structure in the past.

```xml
    <dependency>
      <groupId>org.ehcache</groupId>
      <artifactId>ehcache</artifactId>
      <classifier>jakarta</classifier>
    </dependency>
```

wilkinsona:
Thanks for pointing this out. I have updated the migration guide to [mention the need for the classifier](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.0-Migration-Guide#ehcache3). The list of third-party dependency upgrades now also links to the [Ehcache 3.10 release notes](https://github.com/ehcache/ehcache3/releases/tag/v3.10.0) which show how to use Ehcache 3.10 with Jakarta EE 9+.

grafjo:
@wilkinsona thanks for updating the migration guide!

PR #33396, opened 2022-11-28
remove jakarta classifier from ehcache managed version

grafjo:
related issue
https://github.com/spring-projects/spring-boot/issues/33395


wilkinsona:
Thanks for the proposal. The classifier is required for compatibility with Jakarta EE 9 and later so we cannot remove it.

PR #33418, opened 2022-11-30
update: output capture tests

currenjin:
Hi! 
The readability was enhanced by applying chaining of the Assert.

Test method list
- `getOutShouldOnlyReturnOutputCapturedFromSystemOut`
- `getErrShouldOnlyReturnOutputCapturedFromSystemErr`


wilkinsona:
Thanks for the proposal but I don't think this really improves the code. It makes no functional difference and I'm not sure that it reads better. Thanks anyway.

Issue #32886, opened 2022-10-26
bootBuildImage fails with combination  of environment and buildpacks


bitbaggi:
Hello,

i have the normal vanilla Spring Boot 3.0.0-RC1 with only org.graalvm.buildtools.native as extra plugin.
You can download it from here [Spring Initializr](https://start.spring.io/#!type=gradle-project&language=java&platformVersion=3.0.0-RC1&packaging=jar&jvmVersion=17&groupId=com.example&artifactId=demo&name=demo&description=Demo%20project%20for%20Spring%20Boot&packageName=com.example.demo&dependencies=native)

When i try to build with gradlew bootBuildImage with buildpacks and environment it gives me an error.
```
Running creator
    [creator]     Restoring data for SBOM from previous image 
    [creator]     ===> DETECTING                              
    [creator]     ERROR: No buildpack groups passed detection.
    [creator]     ERROR: Please check that you are running against the correct path.
    [creator]     ERROR: failed to detect: no buildpacks participating

```

My bootBuildImage task is configured in build.gradle:
```
tasks.named('bootBuildImage') {
    builder = 'docker.io/paketobuildpacks/builder:tiny'
    runImage = 'docker.io/paketobuildpacks/run:tiny-cnb'
    buildpacks = ['gcr.io/paketo-buildpacks/bellsoft-liberica:9.9.0-ea', 'gcr.io/paketo-buildpacks/java-native-image']
    environment = [
            'BP_NATIVE_IMAGE_BUILD_ARGUMENTS': '--verbose'
    ]
}
```

The --verbose is only a test, i try to add a reflection-config.json to native image.

When i remove the environment part from task configuration everything is fine.

```
plugins {
    id 'org.springframework.boot' version '3.0.0-RC1'
    id 'io.spring.dependency-management' version '1.1.0'
    id 'org.graalvm.buildtools.native' version '0.9.16'
    id 'java'
}

group = 'com.example'
version = '0.0.1-SNAPSHOT'
sourceCompatibility = '17'

repositories {
    mavenCentral()
    maven { url 'https://repo.spring.io/milestone' }
}

dependencies {
    implementation 'org.springframework.boot:spring-boot-starter'
    testImplementation 'org.springframework.boot:spring-boot-starter-test'
}

tasks.named('bootBuildImage') {
    builder = 'docker.io/paketobuildpacks/builder:tiny'
    runImage = 'docker.io/paketobuildpacks/run:tiny-cnb'
[...]

PR #33424, opened 2022-11-30
Update instead of replace environment in bootBuildImage documentation

candrews:
Updates the documentation for `bootBuildImage` to update `environment` instead of replacing it so that defaults / automatic values are preserved.

Closes gh-32886


mhalbritter:
Thank you!

PR #33460, opened 2022-12-05
Fix typo in tomcat accesslog checkExists doc

aksh1618:
Bumped into this typo today while working, thought to drop a quick fix, cheers!


aksh1618:
Hi @mhalbritter, no issues. Actually I went to do sign it, but the [instructions](https://cla.pivotal.io/about#obvious-fixes) said signing is not required for such changes. I've signed nonetheless, just notifying in case those instructions need to be updated.

mhalbritter:
Ah sorry, you're right. Thanks for signing it nonetheless!

mhalbritter:
Thank you and congratulations on your first contribution!

PR #33493, opened 2022-12-08
Adding Unleash Auto-Configuration

mxsb:
# Adding Unleash Auto-Configuration

Added an auto-configuration for [Unleash](https://www.getunleash.io/) feature toggles that works either with properties or a provided UnleashConfig bean. This works on the latest 7.0.0 [Unleash Java client](https://github.com/Unleash/unleash-client-java) release.

The Unleash client is under Apache 2.0 license, so I belief this should work together with the Spring Boot Apache 2.0 license.

Created tests to make sure everything works as intended.

I didn't really find a place for documentation. If I overlooked it, please advice and I'll add some documentation around this feature. In general it's based on the [clients config options](https://github.com/Unleash/unleash-client-java#configuration-options).

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-autoconfigure/build.gradle
@@ -37,6 +37,7 @@ dependencies {
 	optional("io.rsocket:rsocket-transport-netty")
 	optional("io.undertow:undertow-servlet")
 	optional("io.undertow:undertow-websockets-jsr")
+	optional("io.getunleash:unleash-client-java")
 	optional("jakarta.jms:jakarta.jms-api")
 	optional("jakarta.mail:jakarta.mail-api")
 	optional("jakarta.json.bind:jakarta.json.bind-api")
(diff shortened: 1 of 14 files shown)


scottfrederick:
@mxsb Have you considered creating a stand-alone Spring Boot starter project that contains this auto-configuration, instead of adding it to Spring Boot? As a separate project, it could be maintained and released independent of the Spring Boot release cycle and you could contribute to it without needing to submit PRs to this project. We maintain a list of stand-alone starters [here](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-starters/README.adoc).

mxsb:
Hey @scottfrederick 
I've not considered a starter so far. Would you recommend to create a starter instead of creating an auto config in the core project? I hope that this does not require a lot of maintenance, but let me know your thoughts.
Happy to create a starter if you don't want to include it into the core project.

wilkinsona:
I agree with Scott. I think it would be better to maintain this outside of the main Spring Boot repository, either within the Unleash organisation or as a separate third-party project. Thanks for being part of the Spring community, @mxsb.

PR #33505, opened 2022-12-09
Documented Java compiler level is wrong

peter-janssen:
Documentation is states the default compiler level is java 1.8 which has been changed to 17 since Spring Boot 3.0.0


peter-janssen:

mhalbritter:
Thanks and congratulations on your first contribution!

Issue #33482, opened 2022-12-07
Remove run example from CLI documentation


ljz572000:
The Spring Boot CLI version 3.0.0 hasn't the "run" command.  But the [spring-boot-reference](https://docs.spring.io/spring-boot/docs/3.0.0/reference/html/getting-started.html#getting-started.installing.cli.quick-start) document still has that. 

```shell
>spring run app.groovy  
'run' is not a valid command. See 'help'.

>
```

```shell
>spring help
usage: spring [--help] [--version] 
       <command> [<args>]

Available commands are:

  init [options] [location]
    Initialize a new project using Spring Initializr (start.spring.io)

  encodepassword [options] <password to encode>
    Encode a password for use with Spring Security

  shell
    Start a nested shell
  shell
    Start a nested shell

Common options:

  --debug Verbose mode
    Print additional status information for the command you are running


See 'spring help <command>' for more information on a specific command.
```

digitter:
What's the alternative command?

reasonpun:
What's the alternative command?

scottfrederick:
@reasonpun There is no replacement for the Spring CLI `run` command in Spring Boot 3.0. See #32263 for the justification.

sannanansari:
Thanks @scottfrederick for your reference, I have made pull request #33519.

reasonpun:
Thanks @scottfrederick

scottfrederick:
Closing in favor of #33519.

PR #33519, opened 2022-12-14
Remove the run example

sannanansari:
As mention in #33482 , I have remove the run example from documentation.


mhalbritter:
Thanks!

PR #33536, opened 2022-12-16
Update and rename LICENSE.txt to LICENSE

thomasturrell:
* Remove boilerplate from license.
* Rename to LICENSE, which is the current default name for a LICENSE in a GitHub repository.

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -176,18 +176,7 @@
 
    END OF TERMS AND CONDITIONS
 
-   APPENDIX: How to apply the Apache License to your work.
-
-      To apply the Apache License to your work, attach the following
-      boilerplate notice, with the fields enclosed by brackets "{}"
-      replaced with your own identifying information. (Don't include
-      the brackets!)  The text should be enclosed in the appropriate
-      comment syntax for the file format. We also recommend that a
-      file or class name and description of purpose be included on the
-      same "printed page" as the copyright notice for easier
-      identification within third-party archives.
-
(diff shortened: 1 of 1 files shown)


philwebb:
Thanks for the PR but most Spring projects use `LICENSE.txt` which is still recognized by GitHub. I think we should also leave the appendix section.

thomasturrell:
@philwebb thank you for taking the time to review my pull request.

An additional issue that I resolved in this pull request replacing the template: 

`Copyright {yyyy} {name of copyright owner}`

with:

`Copyright 2002-2022 Pivotal, Inc.`

Do you think it is worth me opening another pull request with just this (minor) change?

thomasturrell:
> `Copyright 2002-2022 Pivotal, Inc.`

Although I believe that this should now be

`Copyright 2002-2023 Pivotal, Inc.

philwebb:
@thomasturrell I think we usually leave that template information as it is in the `LICENCE.txt` file. For example, [here is the one from Spring Framework](https://github.com/spring-projects/spring-framework/blob/88cac6a4a3a84ff9891d226d7ad47d2707af6d96/LICENSE.txt#L190).

We do update the copyright in source files that we change. For example, [here's](https://github.com/spring-projects/spring-boot/commit/f588793445dee905335625ba946674ed3a771db0) a recent commit where I updated the date. We use "the original author or authors" rather than "Pivotal" or any specific company.

PR #33545, opened 2022-12-19
Updated installation file - Run example from CLI

shekharAggarwal:
As mentioned in https://github.com/spring-projects/spring-boot/issues/33482 , I have updated the run example from the documentation.


scottfrederick:
Thanks very much for the contribution. We have an existing PR #33519 that addresses this issue that we have not merged yet.

PR #33569, opened 2022-12-21
Fix typo in LocalDevToolsAutoConfiguration logging

Artur-:


philwebb:
Good spot. Thanks for the fix!

Issue #33528, opened 2022-12-15
spring-boot-maven-plugin tags option, multiple issues


DemianTinkiel:
when creating an image with spring-boot-maven-plugin I encountered the following issues:

1. Does not accept uppercase for tags, this doesn't allow me to add ${project.version}. Yet this is the default behaviour if tags is not specified. However, specifying tags overrides the behaviour, does not add to it (which is what I interpreted by the documentation  by "One or more **additional** tags to apply to the generated image".)
```
Execution artifacts-prepare of goal org.springframework.boot:spring-boot-maven-plugin:2.7.3:build-image failed: Unable to parse image reference "3.0.0-SNAPSHOT". Image reference must be in the form '[domainHost:port/][path/]name[:tag][@digest]', with 'path' and 'name' containing only [a-z0-9][.][_][-] -> [Help 1]

```
2. Setting <tags> as latest generates 2 images - myRepo:8090/group/myproject:latest and latest:latest, both with the same hash
```
docker images | grep 757bb1ccd2ff
latest                              latest              757bb1ccd2ff   42 years ago    264MB
myRepo:8090/group/myproject         latest              757bb1ccd2ff   42 years ago    264MB
```

for issue 1, my pom is
```
<plugin>
  <groupId>org.springframework.boot</groupId>
  <artifactId>spring-boot-maven-plugin</artifactId>
  <version>2.7.3</version>
  <configuration>
    <docker>
      <publishRegistry>
        <url>${image.repo.url}</url>
        <username>${image.repo.user}</username>
        <password>${image.repo.pass}</password>
      </publishRegistry>
    </docker>
    <image>
      <name>${image.prefix}/${project.artifactId}</name>
      <tags>latest,${project.version}</tags>
    </image>
  </configuration>
  <executions>
    <execution>
      <id>artifacts-prepare</id>
      <goals>
        <goal>build-image</goal>
      </goals>
    </execution>
  </executions>
</plugin>
```

for issue 2 the pom is the same but with `<tags>latest</tags>`

scottfrederick:
The values provided to the `tags` option should be full image references in the form of `[image name]:[tag]` or `[repository]/[image name]:[tag]`. It was done this way to provide users control over the full image reference. The name `tags` was chosen to be consistent with the [`pack` CLI](https://buildpacks.io/docs/tools/pack/#pack-cli), but it can be confusing. 

We should clarify this in the Maven and Gradle plugin documentation.

[...]

PR #33609, opened 2022-12-21
Improve maven plugin tags documentation

shekharAggarwal:
As mentioned in https://github.com/spring-projects/spring-boot/issues/33528, I have updated the description of `tags`.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/docs/asciidoc/packaging-oci-image.adoc
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/docs/asciidoc/packaging-oci-image.adoc
@@ -190,7 +190,7 @@ The value supplied will be passed unvalidated to Docker when creating the builde
 | `false`
 
 | `tags`
-|One or more additional tags to apply to the generated image. The values provided to the `tags` option should be full image references in the form of `[image name]:[tag]` or `[repository]/[image name]:[tag]`. The name `tags` was chosen to be consistent with the https://buildpacks.io/docs/tools/pack/#pack-cli[`pack CLI`].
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/docs/asciidoc/packaging-oci-image.adoc
```diff
@@ -190,7 +190,7 @@ The value supplied will be passed unvalidated to Docker when creating the builde
 | `false`
 
 | `tags`
-| One or more additional tags to apply to the generated image.
+|One or more additional tags to apply to the generated image. The values provided to the `tags` option should be full image references in the form of `[image name]:[tag]` or `[repository]/[image name]:[tag]`. The name `tags` was chosen to be consistent with the https://buildpacks.io/docs/tools/pack/#pack-cli[`pack CLI`].
```
scottfrederick, line 193: We can remove the last sentence `The name 'tags' was chosen to be consistent with the https://buildpacks.io/docs/tools/pack/#pack-cli['pack CLI']`. While accurate, the statement is true of many arguments here, and we don't need to call this one out specifically.

scottfrederick:
Thanks for the contribution @shekharAggarwal. We need to make the same change to the [Gradle plugin documentation](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/docs/asciidoc/packaging-oci-image.adoc). Are you interested in making that change also? 

If so, you can just push an additional commit to this branch and it will update the pull request.

shekharAggarwal:
Sure

shekharAggarwal (commented):
Removed unwanted statements.

philwebb:
Thanks!

PR #33614, opened 2022-12-22
Improve gradle plugin tags documentation

shekharAggarwal:
As mentioned in #33528, I have updated the description of the tags for the Gradle plugin documentation.


philwebb:
Thanks!

Issue #33474, opened 2022-12-06
Remove Jetty / Netty / Apache HttpClient checks for WebClient


mhalbritter:
With https://github.com/spring-projects/spring-framework/issues/29644 the `WebClient` doesn't need external dependencies anymore. If no netty / jetty / Apache HttpClient is found, it falls back to the `JdkClientHttpConnector` which is baked into the JDK.

At least in `org.springframework.boot.test.web.reactive.server.WebTestClientContextCustomizerFactory` we check for the presence of netty / jetty / apache httpclient. This check can be removed, and maybe such checks are somewhere else in the codebase.

wilkinsona:
I wonder if we should consider this a bug? Looking at the code, the checks are preventing the auto-configuration of a `WebTestClient` that uses the `JdkClientHttpConnector`.

scottfrederick:
Closing in favor of #33640.

PR #33640, opened 2022-12-27
Remove Jetty/Netty/ApacheHttpClient checks for WebClient

anugrahsinghal:
adds the functionality according to [#33474](https://github.com/spring-projects/spring-boot/issues/33474)
- removes additional checks for netty/jetty/Apache HttpClient dependencies


anugrahsinghal:
> At least in org.springframework.boot.test.web.reactive.server.WebTestClientContextCustomizerFactory we check for the presence of netty / jetty / apache httpclient. This check can be removed, and maybe such checks are somewhere else in the codebase.

@mhalbritter 
I have used [this sourcegraph query](https://sourcegraph.com/search?q=context:global+repo:%5Egithub%5C.com/spring-projects/.*+%22reactor.netty.http.client.HttpClient%22+or+%22org.eclipse.jetty.client.HttpClient%22+or+%22org.apache.hc.client5.http.impl.async.CloseableHttpAsyncClient%22+or+%22org.apache.hc.core5.reactive.ReactiveDataConsumer%22&patternType=standard&sm=1) to find the presence of similar dependencies in other places in the codebase, and could not find one that would need removal.

anugrahsinghal:
@mhalbritter a little bit unsure of how to handle [this test](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test/src/test/java/org/springframework/boot/test/web/reactive/server/WebTestClientContextCustomizerWithoutSupportedHttpClientTests.java#L35)


This test becomes invalid because there is always the `JdkClientHttpConnector` available
even after the `@ClassPathExclusions({ "reactor-netty*.jar", "jetty-client*.jar" })`

This means that the `WebTestClientContextCustomizerFactory` is always `non-null`

Do you have any suggestions on how to modify this test case?

scottfrederick:
@anugrahsinghal I think the `WebTestClientContextCustomizerWithoutSupportedHttpClientTests` can be removed, since there is no longer a case where no client is availble.

anugrahsinghal:
hey @scottfrederick 

I have removed the `WebTestClientContextCustomizerWithoutSupportedHttpClientTests` test.

mhalbritter:
Thanks and congratulations to your first contribution to Spring Boot!

Issue #24992, opened 2021-01-25
Allow addition of custom logback Actions 


a1dutch:
Hi,

I was looking at adding a custom `Action` via the `SpringBootJoranConfigurator` but this is not currently possible.

My use case is to enable or disable an appender based on a property via the spring `Environment` class.

I managed to add a custom action based on the `newRule` extension, but its not possible to get access to the `Environment` class.

A way to either configure the `SpringBootJoranConfigurator` with the `LoggingInitializationContext` or a new way to enable/disable based on a property would be handy, e.g. like a <conditionalOnProperty property="my.property.enabled" havingValue="true" matchOnMissing=true />

Cheers

Andy

philwebb:
I wonder if we could update `SpringBootJoranConfigurator` to add additional `JoranConfigurator` instances loaded via `spring.factories`.

> I managed to add a custom action based on the newRule extension

@a1dutch How did you do this? Can you share some code?

wilkinsona:
In Logback, [conditional configuration is typically handled via Janino](http://logback.qos.ch/manual/configuration.html#conditional). It would be less general purpose, but for this specific use case I wonder if we should looking at making the `Environment` available to Janino's expressions? By default they are "a Java expression in which only context properties or system properties are accessible".

a1dutch:
@philwebb - here is the docs for newRule: http://logback.qos.ch/manual/onJoran.html#newRule

here is my working example based on classes present, the following is added to logback-spring.xml

```
<newRule pattern="configuration/enabledWhenClassPresent" actionClass="****.logging.slf4j.EnabledWhenClassPresentAction"/>

<enabledWhenClassPresent class="some.classname.Here">
...
</enabledWhenClassPresent>
```

The `EnabledWhenClassPresentAction` is the same as `SpringPropertyAction` apart from its doing a `Class.forname` to check if the class exists.

i think two new actions supported by default (enabled based on a property & enabled based on class present) would be a good addition to spring boot and cover any possible use cases. 

@wilkinsona i initially looked at Janino expressions, i will take another look and see if its possible to use `springProperty` in conjunction with this.

philwebb:
I wonder if we can put the `org.springframework.boot.logging.LoggingInitializationContext` class into `ch.qos.logback.core.Context`? That way `Actions` could get it via the `InterpretationContext`. E.g.:

```java
[...]

Issue #25847, opened 2021-03-31
Provide a hook for configuring Logback programatically


celcius112:
At the moment we are considering migrating few applications to spring-native. Everything works fine excepts our Logback configuration. We use a `logback.xml`, and as stated in https://github.com/spring-projects-experimental/spring-native/issues/625 it will not work as intended, nor is it an easy problem to tackle. 

We were already considering moving to the programmatic Logback configuration, using a [Configurator](https://github.com/qos-ch/logback/blob/master/logback-classic/src/main/java/ch/qos/logback/classic/spi/Configurator.java), which would probably be easier to use with spring-native and would provide better application startup for regular applications. However, we cannot migrate all of our `logback.xml` configuration since we cannot import default spring configuration nor use `sprintProperty/springProfile` elements.

sdeleuze:
Status update:
 - Currently Spring Boot 3 only supports Logback configuration via `application.properties` or Yaml equivalent, `logback.xml` is unsupported.
 - After a deeper look, I confirm that build time generation of programmatic Logback configuration from `logback.xml` is the only identified way to support it, there are already discussions and experiments in the JVM/GraalVM communities about that.
 - Providing such hook to users would unlock those who need more advanced customization capabilities than what `application.properties` offers and could allow Spring community to experiment about build time `logback.xml` to programmatic version in an autonomous manner. To be evaluated against other priorities of course.

philwebb:
It looks like a `ServiceLoader` could be used to plug-in a custom `ch.qos.logback.classic.spi.Configurator`, however, I don't think our logging initialization code would detect that happened and we we'd probably end up applying our configuration. The `org.springframework.boot.logging.AbstractLoggingSystem` class has a `initializeWithConventions` which attempts to detect `logback.xml` and if it's found not apply our defaults.

We do already have a package-private `org.springframework.boot.logging.logback.DefaultLogbackConfiguration` class which applies our defaults in a programmatic way. I don't really want to make that class public, especially if the GraalVM community is working on a `logback.xml` -> `Configurator` processor. Presumably, such a processor would need to deal with `<include>` tags, at which point our existing XML files could just be included and used directly.

Adding hook points for logging is quite a challenge since they're normally needed before the Spring `ApplicationContext` is loaded. I think we should see how the community experiments progress before we do too much on our side.

jack-berg:
I'd like to see this issue solved to accommodate what I think is a similar use case:

[...]

PR #33643, opened 2022-12-28
Allow JoranConfigurators to be passed ahead of time in logback

libetl:
Hello Spring Boot team,

I'm quite new here and I don't know what kind of checklist I need to pass prior to submitting a pull request, sorry about that.

I am trying to understand what kind of changes I am expected to do to use spring-boot 3 in my company.

Logback team has deactivated xml declaration of rules `<newRule>` since logback 1.3 for security purposes, so we have to do it programmatically.

`SpringBootJoranConfigurator` declares a limited set of rules, so I thought it would be nice to let the user provide its own JoranConfigurators ahead of time.

I need that, because I don't want some critical information to be sent in the logs messages.
One of the rule we have in my company is `ObfuscationRule`, it is a requirement to have it in spring-boot 3.

Thank you for your help.
Related issue (not from me : #24992)
@philwebb would be the one-to-go developper for this item.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java
@@ -188,7 +188,7 @@ public class LogbackLoggingSystem extends AbstractLoggingSystem implements BeanF
 		if (isAlreadyInitialized(loggerContext)) {
 			return;
 		}
-		if (!initializeFromAotGeneratedArtifactsIfPossible(initializationContext, configurators, logFile)) {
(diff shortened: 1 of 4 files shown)


mhalbritter:
Thanks and congrats to your first contribution to Spring Boot!

wilkinsona:
Regrettably, upon subsequent review we've released that we made a mistake when accepting these changes. Sorry. The changes only apply during AOT processing which is too specialised and they will also make it harder for us to offer full support for programmatic configuration of Logback in the future (#24992 and #25847). #34361 is tracking the revert which will allow us the time that we need to determine the right way forward here without backing ourselves into a corner.

Frettman:
@libetl, have you found a workaround for this issue?

libetl:
@Frettman 
https://libetl.wordpress.com/2022/12/31/spring-boot-3-and-logback-without-hassle/

PR #33648, opened 2022-12-29
Upgrade to Logback 1.3.5

ibrusentsev:
I encountered the [LOGBACK-1162](https://jira.qos.ch/browse/LOGBACK-1162) bug using Spring Boot Framework. 
This bug was fixed in Logback 1.3.+ versions, so I decided to update the Spring Boot dependency accordingly. As Spring Boot 2.x has to support Java 8, the latest compatible Logback version is 1.3.5 - [Logback 1.4.x works only with Java 11](https://logback.qos.ch/dependencies.html).
The Spring Boot 3.x has been already upgraded to Logback 1.4.5. So this change contains cherry-picking from Spring Boot 3.0 with upgrading to Logback 1.4.5 and downgrading Logback to 1.3.5 with tuning necessary libraries versions.


philwebb:
Thanks for the pull-request but we're unable to upgrade to Logback 1.3 in a maintenance release of Spring Boot 2.7. We need to stick on Logback 1.2.x. You can read about our upgrade policy at https://github.com/spring-projects/spring-boot/wiki/Supported-Versions#third-party-dependencies.

PR #33653, opened 2022-12-29
Simplify AssertJ assertions and also make them more readable

krzyk:
Simplified AssertJ assertions (e.g. instead of doing `assertThat(list.size()).isEqualTo(2)` one can do `assertThat(list).hasSize(2)` which provides cleaner error message and makes the test slightly more readable).
Change was mostly automated using "Concise AssertJ Optimizing Nitpicker ​(Cajon)​", with some manual work where it made incorrect changes.


philwebb:
Nice work! Thanks.

PR #33663, opened 2022-12-30
Migrate from soon to be deprecated SnakeYAML constructor

asomov:

### DIFF
--- a/spring-boot-system-tests/spring-boot-image-tests/src/systemTest/java/org/springframework/boot/image/paketo/LayersIndex.java
+++ b/spring-boot-system-tests/spring-boot-image-tests/src/systemTest/java/org/springframework/boot/image/paketo/LayersIndex.java
@@ -25,6 +25,7 @@ import java.util.Map;
 import java.util.jar.JarFile;
 import java.util.zip.ZipEntry;
 
+import org.yaml.snakeyaml.LoaderOptions;
 import org.yaml.snakeyaml.Yaml;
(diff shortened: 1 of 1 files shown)


asomov:
Do not use deprecated method in SnakeYAML - it might be removed

wilkinsona:
Thanks for the PR, @asomov. The code that's changed here as part of Spring Boot's system tests. It's only used by Boot's tests and isn't published anywhere. As such, what's the motivation for these changes?

asomov:
@wilkinsona as I said - it uses deprecated method which will be removed in the coming major release of SnakeYAML. This project stops to compile. This will simplify the migration

philwebb:
Looks like [this commit](https://bitbucket.org/snakeyaml/snakeyaml/commits/2b8d47c8bcfd402e7a682b7b2674e8d0cb25e522) deprecated the constructor method we use and [this one](https://bitbucket.org/snakeyaml/snakeyaml/src/3e755d254aeaa902675053047fd53368a175565a/src/main/java/org/yaml/snakeyaml/constructor/Constructor.java?at=master) removed it. I don't think they'll be a release where users see the deprected warning.

asomov:
@philwebb yes, you are right. Because SnakeYAML will break the backwards compatibility in the coming release, we take the chance to remove a lot of old code. (Even though Spring itself is not affect by this hype about quasi-vulnerability, the Spring users are confused)

PR #33687, opened 2023-01-04
Hazelcast auto-configuration loads configuration from custom locations without overriding and from default locations with overriding

luben93:
use the hazelcast `Config.loadConfig()` which in turn calls `Config.applyEnvAndSystemVariableOverrides()` to override the loaded config with environment variables (which is the default behavior when starting a hazelcast without spring boot auto-configuration)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/hazelcast/HazelcastServerConfiguration.java
@@ -20,8 +20,6 @@ import java.io.IOException;
 import java.net.URL;
 
 import com.hazelcast.config.Config;
-import com.hazelcast.config.XmlConfigBuilder;
-import com.hazelcast.config.YamlConfigBuilder;
 import com.hazelcast.core.Hazelcast;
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal. I'm not sure that this should be considered a bug. I can't recall the details, but the code's been this was since the auto-configuration [was introduced](https://github.com/spring-projects/spring-boot/commit/721b5a2395322b7c13a1361296361d26118dd6d3) and it may well be intentional. Even if we did consider it to be a bug, changing it is not without risk as it may be a breaking change if things are now overridden that weren't previously. I'll flag this for team attention so that we can consider our options.

luben93:
Perhaps not a bug but it's very counter intuitive in regards to the hazelcast documentation that states that you should be able to override with environment variables, but when using autoconfigure with _spring.hazelcast.config_ set that just stops to work

luben93:
If the _spring.hazelcast.config_ is not set Config.load() is called instead and that does enable the overriding

philwebb:
We discussed this today and we think that this should be considered as a bug, but it's quite risky for us to fix it in a patch release. As such, we'll change the code in Spring Boot 3.1 and recommend that folks use the work-around if they are on earlier versions.

wilkinsona:
The workaround is to define a `Config` bean, loaded using the preferred mechanism.

mhalbritter:
Thanks!

PR #33711, opened 2023-01-06
Upgrade to AssertJ 3.24.2

scordio:
Aside from just upgrading the version, this also switches to the [AssertJ BOM](https://assertj.github.io/doc/#dependency-metadata-assertj-bom).


scordio:
Switching to draft due to assertj/assertj#2901 which will be fixed by 3.24.1 (release in progress).

wilkinsona:
Thanks very much, @scordio.

Issue #33685, opened 2023-01-04
Documentation should be there for better understanding in SpringApplicationAdminMXBeanRegistrar.java


deepak1067:

spring-projects-issues:
If you would like us to look at this issue, please provide the requested information. If the information is not provided within the next 7 days this issue will be closed.

spring-projects-issues:
Closing due to lack of requested feedback. If you would like us to look at this issue, please provide the requested information and we will re-open the issue.

PR #33770, opened 2023-01-12
Documentation issue solved for #33685

deepak1067:
Documentation issue solved for #33685


mhalbritter:
Hey, this PR has no changed files. Could you please check that on your end?

wilkinsona:
Also, until we have some feedback we cannot make any changes to the documentation as it's unclear what, if anything, is needed, as the code as an internal.

Issue #32527, opened 2022-09-28
Failing calls to reactive health indicators are not logged


jGleitz:
In https://github.com/spring-projects/spring-boot/issues/22632, it was decided to log `WARN` messages whenever a `HealthIndicator` reports a status of `DOWN` _and_ adds an exception to the builder. This was implemented in `AbstractHealthIndicator`, but not in `AbstractReactiveHealthIndicator`. `AbstractReactiveHealthIndicator` still only logs if `doHealthCheck` throws or returns an error Mono. 

For the same reasons as in https://github.com/spring-projects/spring-boot/issues/22632, and for reasons of consistency, `AbstractReactiveHealthIndicator` should also log a WARN message when `doHealthCheck` returns a Mono that resolves to a Health with status `DOWN` and an exception.

wilkinsona:
@snicoll and I have discussed this and we think it should be treated as a bug of omission. The logging should have been added in 2.6 as part of #22632.

mhalbritter:
I've created a PR #33774 for this.

mhalbritter:
Superseded by https://github.com/spring-projects/spring-boot/issues/32527

PR #33774, opened 2023-01-12
Failing calls to reactive health indicators are not logged

mhalbritter:
See https://github.com/spring-projects/spring-boot/issues/32527

As the `org.springframework.boot.actuate.health.AbstractReactiveHealthIndicator#doHealthCheck` method returns a `Mono<Health>`, i had to add the `exception` to the `Health` class. I marked it as `@JsonIgnore` so that it doesn't end up in the JSON.

Please let me know if this is fine.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/AbstractReactiveHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/AbstractReactiveHealthIndicator.java
@@ -31,6 +31,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Stephane Nicoll
  * @author Nikolay Rybak
+ * @author Moritz Halbritter
  * @since 2.0.0
  */
 public abstract class AbstractReactiveHealthIndicator implements ReactiveHealthIndicator {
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/health/AbstractReactiveHealthIndicator.java
```diff
@@ -75,19 +76,24 @@ protected AbstractReactiveHealthIndicator(Function<Throwable, String> healthChec
 
 	@Override
 	public final Mono<Health> health() {
+		Mono<Health> result;
 		try {
-			return doHealthCheck(new Health.Builder()).onErrorResume(this::handleFailure);
+			result = doHealthCheck(new Health.Builder()).onErrorResume(this::handleFailure);
```
linyonggui, line 81: why not return in line ?

mhalbritter: Take a look at line 86. `result` is used in a chain with `.doOnNext`.

wilkinsona:
> Please let me know if this is fine.

LGTM, but I'm not sure how much that means. WDYT, @bclozel?

bclozel:
I think it looks good @mhalbritter

mhalbritter:
Thanks, i'll merge it.

PR #33838, opened 2023-01-16
Update com.gorylenko.gradle-git-properties version to 2.4.1 in doc

izeye:
This PR updates the `com.gorylenko.gradle-git-properties` Gradle plugin version to the latest (2.4.1) in the reference doc.


mhalbritter:
Thanks!

Issue #19605, opened 2020-01-10
Prevent Servlet filter re-registration


lukas-krecan:
We are trying to register two servlet filters implemented by the same class on two different paths with different configuration

```kotlin
    @Bean
    fun authFilter1(): FilterRegistrationBean<*> {
        val filter = BasicAuthFilter("username1", "password1")
        val reg = FilterRegistrationBean(filter)
        reg.addUrlPatterns("/path1")
        return reg
    }

    @Bean
    fun authFilter2(): FilterRegistrationBean<*> {
        val filter = BasicAuthFilter("username2", "password2")
        val reg = FilterRegistrationBean(filter)
        reg.addUrlPatterns("/path2")
        return reg
    }
```

If we register the filters it like this, Spring Boot uses the filter class name to deduce the filter name and since both filters are implemented by the same class, only one filter is effectively registered which is surprising. Basically it's a silent failure. 

I would prefer Spring Boot to throw an exception in such case or event better, generate a unique name for each filter. 

Of course, once the user is aware of the issue, it can be simply fixed by explicitly specifying the filter name in FilterRegistrationBean.

mbhave:
I think it would be better to throw and exception and let the user explicitly provide a name.

philwebb:
The `setName(...)` method can be used to set a specific name. We could also look at making the registration class `BeanNameAware`. I looks like we already check for a `null` response but currently just log "<name> was not registered (possibly already registered?)"

philwebb:
@mhalbritter I like that, but I think we should consider it an enhancement and target it to 3.1. There's the chance that existing users might be depending on the class based convention name.

philwebb:
Looking at this again, we do log a warning if registration fails. https://github.com/spring-projects/spring-boot/blob/69d6aa418b68435fb2e14e4537d614b66dcc8734/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/DynamicRegistrationBean.java#L109.

Perhaps we should keep things as they are in 2.7 but change that to an exception in 3.1.

mhalbritter:
Sounds good. I'll refine the PR and close this issue in favor of the PR https://github.com/spring-projects/spring-boot/pull/33911

PR #33911, opened 2023-01-19
Prevent Servlet and Filter re-registration

mhalbritter:
Registration failures now throw an `IllegalStateException` instead of being logged. I had to build in an escape hatch named `setIgnoreRegistrationFailure` which is used in the `JerseyAutoConfiguration`. If this is set to `true`, it logs instead of throwing an exception.

If the name is set, it is used. If not, the bean name is used (by implementing `BeanAware`). If that's not set, the convention based name is taken.

See https://github.com/spring-projects/spring-boot/issues/19605


mhalbritter:
We currently log an INFO level message when a servlet / filter is registered but this has no effect as there was already an registration with the same name. We should investigate if this can be changed to an exception.

Issue #33916, opened 2023-01-20
update-homebrew-tap.sh uses repo.spring.io


philwebb:
It should use central

wilkinsona:
Closing in favor of #33918.

PR #33918, opened 2023-01-20
update-homebrew-tap.sh should retrieve formula from Maven Central

sannanansari:
#33916 update the url.


mhalbritter:
Thank you!

PR #33924, opened 2023-01-20
Remove hints for Spring Integration Graph

artembilan:
Related to: https://github.com/spring-projects/spring-integration/issues/3994

Spring Integration's `IntegrationGraphServer` can be used without actuator or even Spring Boot at all.
That is more natural to have respective reflection hints exposed by Spring Integration itself.
See related GH issue and its fix which just does a migration of the remove code from `IntegrationGraphEndpoint`


wilkinsona:
Switching to a task as this change combined with the Spring Integration change should have no noticeable effect for Boot users.

wilkinsona:
We need to move to Spring Integration snapshots (https://github.com/spring-projects/spring-boot/issues/34039) first. We should do that late next week in advance of the release train starting the following week.

Issue #33938, opened 2023-01-23
Consistently use repo.maven.apache.org to reference Maven Central


wilkinsona:
With thanks the @breun for [this comment](https://github.com/spring-projects/spring-boot/issues/33705#issuecomment-1398995167) on #33705:

> According to https://issues.apache.org/jira/browse/MNG-5151 Maven first used repo1.maven.org, but later decided on making repo.maven.apache.org the default hostname for Maven Central, so maybe that should be used?

wilkinsona:
According to [this blog post](https://central.sonatype.org/news/20190715_http_deprecation_update/#canonical-urls), Sonatype consider both https://repo.maven.apache.org and https://repo1.maven.org to be canonical URLs for Maven Central. We currently use a mixture of the two. In the interests of consistency, I think there's still some value in make a change here.

PR #33943, opened 2023-01-23
Updated the maven url as per default hostname for Maven Central

sannanansari:
#33938 updated url.


wilkinsona:
Thanks for the PR but we prefer to handle a change like this that may affect multiple places ourselves. By the time that we have checked that all usages have been replaced, it is less time consuming to just make the change.

Issue #33748, opened 2023-01-10
Spring Sesssion database initializer configuration - support initialization on different table names


alexanderankin:
I was reading #6649 and thought that since the fix got lost in the 2.1 merge (its [here][8f1be4cdedd425aa40830235b42e1b01b3facaa8] but [not here][aef92b9295f62d008faa9ab79905a474bf3496f3]) - why not support it fully. I can submit the original fix if that is preferable, but at work we use Mongo and i thought it was nice that you could customize the collection name, so thought it would work the same way on jdbc.

I don't mind submitting a PR to restore the intended functionality either, but making it fully functional seems preferable if possible.

[8f1be4cdedd425aa40830235b42e1b01b3facaa8]: https://github.com/spring-projects/spring-boot/commit/8f1be4cdedd425aa40830235b42e1b01b3facaa8
[aef92b9295f62d008faa9ab79905a474bf3496f3]:  https://github.com/spring-projects/spring-boot/commit/aef92b9295f62d008faa9ab79905a474bf3496f3

<details><summary>i figure the fix could look something like this</summary>

```diff
diff --git a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionDataSourceScriptDatabaseInitializer.java b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionDataSourceScriptDatabaseInitializer.java
index a5279ecb45..4669897a74 100644
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionDataSourceScriptDatabaseInitializer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionDataSourceScriptDatabaseInitializer.java
@@ -16,6 +16,9 @@
 
 package org.springframework.boot.autoconfigure.session;
 
+import java.io.IOException;
+import java.nio.charset.Charset;
+import java.util.ArrayList;
 import java.util.List;
 
 import javax.sql.DataSource;
@@ -24,6 +27,8 @@ import org.springframework.boot.jdbc.DatabaseDriver;
 import org.springframework.boot.jdbc.init.DataSourceScriptDatabaseInitializer;
 import org.springframework.boot.jdbc.init.PlatformPlaceholderDatabaseDriverResolver;
[...]

PR #9752, opened 2017-07-13
Harmonize database initializers

vpavic:
This commit updates database initializers configuration to enable them automatically only when an embedded `DataSource` is used. Related configuration properties have been updated to use a more expressive `DatabaseInitializerMode` enum rather than `Boolean` flag.

This resolves #8981.

PR #6649, opened 2016-08-15
Validate Spring Session database initializer configuration

vpavic:
This PR adds Spring Session JDBC configuration validation that disables database initializer in case custom table name is configured with default schema.

Currently, it is possible to set custom table name but leave default schema configuration property. This will either result in an error in runtime since expected table (with custom name) won't be found, or, if user has set up the correct table by some other means, with created unused table.

In both cases the correct approach would be to fail fast and indicate invalid configuration.

PR #33961, opened 2023-01-23
addresses #33748 validate default table on jdbc session init

alexanderankin:
Refuses to start when jdbc session is initialized and custom table is specified, but default script is specified. default scripts are not currently customizable for a custom table name, so this configuration will result in a schema being defined for one table name, but sessions being stored in another table. this is almost definitely not the intended the end result. so we will ask the user to specify the schema (or set the initialization to never).

this is the intermediate solution to restore functionality lost in a merge as discussed in #33748

implementation notes:

I'm not sure whether i should have used the "customize" protected method, which can be overridden, and occurs close enough in the general execution workflow, that an exception thrown there or in my overridden runScripts method. But it seems that customization logic should live there, not validation logic? I can override the other method instead of runScripts if that is at all preferable.

it seems there is already an architectural decision in place here - to deal with the configuring the different implementations - all the implementations of `DataSourceScriptDatabaseInitializer` need to call the super constructor taking `DataSource` and `DatabaseInitializationSettings` as the second argument. perhaps it makes sense to augment the database initialization settings to add the necessary information to perform this particular kind of validation... then this solution will more easily translate to other implementations. otherwise it is less than ideal to hook into the constructors and store the implementation-specific ConfigurationProperties object.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionDataSourceScriptDatabaseInitializer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionDataSourceScriptDatabaseInitializer.java
@@ -88,7 +88,7 @@ public class JdbcSessionDataSourceScriptDatabaseInitializer extends DataSourceSc
 		super.runScripts(scripts);
 	}
 
-	private void validateConfiguration() {
+	void validateConfiguration() {
 		if (properties == null) return; // cannot validate without this
 		JdbcSessionProperties defaults = new JdbcSessionProperties();
 		boolean willHappen = switch (properties.getInitializeSchema()) {
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionDataSourceScriptDatabaseInitializer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/JdbcSessionDataSourceScriptDatabaseInitializer.java
@@ -17,10 +17,12 @@
 package org.springframework.boot.autoconfigure.session;
 
 import java.util.List;
+import java.util.Objects;
 
 import javax.sql.DataSource;
 
 import org.springframework.boot.jdbc.DatabaseDriver;
+import org.springframework.boot.context.properties.source.InvalidConfigurationPropertyValueException;
 import org.springframework.boot.jdbc.init.DataSourceScriptDatabaseInitializer;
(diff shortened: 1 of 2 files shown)


alexanderankin:
I might be a little bit more liberal in what I'd try to implement all use cases (checks for jdbc, r2dbc, skips schema-less (mongo, hazel, etc)) in an application codebase, but tried to submit the smallest patch and start a discussion about the implementation in an established codebase that supports graalvm, all sorts of platforms some I might not be familiar with etc

wilkinsona:
I don't think that the functionality that was added in #6649 was lost in a merge. As far as I can tell, it was intentionally removed as part of https://github.com/spring-projects/spring-boot/pull/9752.

There's a [lengthy discussion](https://github.com/spring-projects/spring-boot/pull/6649/files#r74967361) about whether or not to fail fast and the conclusion was not to do so. We'll need to think carefully about whether or not failing fast is the right thing to do now. There's also other initialization to consider for Quartz, Spring Batch, etc. Things are quite nicely aligned and consistent at the moment. If we decide that failing fast is the right thing to do, we should then consider if the need to do so is specific to Spring Session JDBC or if it should be applied more broadly.

alexanderankin:
ah, thanks for that context, I wasn't aware of that and somehow couldn't find it from the commit history.

I agree that where this pr places the checking code is implementation specific and has the drawback of not covering related implementations.

wilkinsona:
> There's also other initialization to consider for Quartz, Spring Batch, etc.

It looks like Quartz is fine as there's no customization related to table names.

Spring Batch does appear to be affected due to the `spring.batch.jdbc.table-prefix` property. It defaults to `BATCH_` and all of the tables in the default schema files are named `BATCH_…`. Changing the table name prefix without providing a custom schema will, as far as I can tell, result in a failure.

alexanderankin:
I don't mind going and adding templating using the established pattern (`JdbcIndexedSessionRepository` queries being parametrized for the table name) for all other related projects, but i would hesitate to go touch large parts of the code without clarification that this is actually desired/acceptable.

wilkinsona:
In #6649, the auto-configuration for Spring Session JDBC was changed so that initialization of the database was enabled automatically if the default table name was set or a custom schema was configured. At this time, failing fast was considered but rejected as, among other reasons, there was no way to be certain that the configuration wouldn't work.

In #9752, this changed as part of harmonizing the behavior of all of the initializers (Batch, Integration, Quartz, and Session JDBC). There is no longer any automatic enablement based on other properties with the default behavior being that only an embedded database is initialized.

As was the case when #6649 was discussed, I don't think we can fail fast here without it being a breaking change. A user could have their own `org/springframework/session/jdbc/schema-*.sql` files in `src/main/resources` that override the files from the `spring-session-jdbc` jar and use the custom table name. Failing fast would break such an arrangement. Users would have to rename the files and configure a custom schema to get things working again. I think that cost outweighs the benefit of failing fast. Thanks anyway for the PR.

PR #33967, opened 2023-01-24
Replace Base64Utils with JDK's Base64

izeye:
This PR replaces Framework's `Base64Utils` with JDK's `Base64` as it has been deprecated in https://github.com/spring-projects/spring-framework/pull/28434.


mhalbritter:
Categorized as task, as this has no visible effect on Boot users.

wilkinsona:
We should do this in 3.0.x as our build will fail due to the deprecation warnings when we switch to Framework 6.0.5 snapshots.

mhalbritter:
Thank you!

PR #33970, opened 2023-01-25
fix(src): check if the session timeout is negative before setting the value

srivatsa-cfp:


wilkinsona:
Thanks for the suggestion. Each web server factory contains logic for handling `null` as well as negative and zero durations. Preventing the duration from being set to a negative value would result in the session timeout remaining at its default value. This would be a breaking change so I don't think it should be made.

PR #33973, opened 2023-01-25
Read property from file in environmet variable

SamueleFacenda:
Working with a docker image of a spring boot application I encoutered some issues: it is very difficult to set properties via docker secrets. Docker secrets are stored in the container as files. To simplify the configuration process some images(like [postgres](https://hub.docker.com/_/postgres)) let you define properties via environment variables in the `PROPERTY_NAME=value` style (like spring boot) but also in the `PROPERTY_NAME_FILE=/path/to/file/with/value`. 

I changed the code that read system properties to, after attempting to read the system property and the environment variable, search a environment variable with the `_FILE` appended and read the content of the file in the path thus defined. 

P.S.
This is my first time looking at this code, I have not understood things like the directory arrangement or the general working. I suppose that I have modified the correct file but I'm not sure(I've found it using the github code search feature querying for `System.getenv` and `System.getProperty`).

Plus, my code comments and function names are not very good, they should probably be modified.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/util/SystemPropertyUtils.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/util/SystemPropertyUtils.java
@@ -16,6 +16,9 @@
 
 package org.springframework.boot.loader.util;
 
+import java.io.File;
+import java.io.IOException;
+import java.nio.file.Files;
 import java.util.HashSet;
 import java.util.Locale;
 import java.util.Properties;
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal. This looks to overlap with Spring Boot's existing [support for "config trees"](https://docs.spring.io/spring-boot/docs/2.7.x/reference/html/features.html#features.external-config.files.configtree). If you were not already aware of that support, please read through the documentation to see if it meets your needs. If it does not, please open a new issue describing exactly what you want to do and why config trees do not meet your needs. I'm afraid we won't be able to merge this proposal as-is. Instead, we would want to consider expanding the existing config tree support.

PR #33987, opened 2023-01-26
Use new Java features

krzyk:
Slight refactorings to use new Java features like text blocks, pattern matching in `instanceof`, new JDK classes, (not so new) diamonds and try-with-resources, and few other java features that weren't used but hopefully increase readability.


philwebb:
Very nice. Thanks very much! I've taken the liberty of splitting the PR into a number of distinct commits and also backported a couple to 2.7.

PR #34001, opened 2023-01-27
liquibase-hibernate6 does not have version 4.17.2

lnthai2002:
springboot 3 uses spring-orm6 thus, hibernate6 and liquibase-hibernate6 must be used. However, the first version of liquibase-hibernate6 is 4.19.0 and since liquibase recommend matching the version of liquibase with all the liquibase extension , i think it is wise for spring boot dependencies to use liquibase 4.19.0

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -779,7 +779,7 @@ bom {
 			]
 		}
 	}
-	library("Liquibase", "4.17.2") {
+	library("Liquibase", "4.19.0") {
 		group("org.liquibase") {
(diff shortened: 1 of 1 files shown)


philwebb:
Thanks for the pull-request. Liquibase 4.19.0 wasn't available at the time that we release Spring Boot 3.0. Due to our [third-party upgrade policy](https://github.com/spring-projects/spring-boot/wiki/Supported-Versions#third-party-dependencies) we can't upgrade to Liquibase 4.19 in a patch release. We will, however, upgrade in Spring Boot 3.1.

We have a semi-automated process to deal with upgrades, so I'm going to close this one.

Thanks anyway.

Issue #31469, opened 2022-06-20
When using OTel, auto-configure CompositeSpanExporters with span exporting predicates, filters, and reporters


mhalbritter:
Micrometer Tracing uses a `SpanExportingPredicate` to determine if a span should be exported and a `SpanFilter` to change the span (like a `MeterFilter`) before it is reported. Both are used in the bridges for brave and Otel (see `CompositeSpanExporter` and `CompositeSpanHandler`). We should auto-configure this in boot.

wilkinsona:
I've opened https://github.com/spring-projects/spring-boot/issues/32753 for the Brave side of things. This issue is now specific to Otel.

wilkinsona:
This requires some changes in Micrometer Tracing. Pushing back to 3.1.x.

scottfrederick:
Closing in favor of #34002

PR #34002, opened 2023-01-27
Add Micrometer Tracing's CompositeSpanExporter

marcingrzejszczak:
Micrometer Tracing comes with 3 generic interfaces, `SpanExportingPredicate`, `SpanReporter` and `SpanFilter`, thanks to which you can decide whether you want to export a span, how you want to report it and how to mutate it regardless of which tracer library you are using.

Without this change usage of these 3 interfaces is not possible for OpenTelemetry.

With this change we're putting all `SpanExporters` to the `CompositeSpanExporter` and apply the exporters only after predicates, filtering and reporting took place.

fixes https://github.com/spring-projects/spring-boot/issues/31469


marcingrzejszczak:
fixes https://github.com/spring-projects/spring-boot/issues/31469

mhalbritter:
Thank you!

PR #34004, opened 2023-01-29
Add property `headers` in `OtlpProperties`

eddumelendez:
In Micrometer 1.11, headers property has been introduced.


mhalbritter:
Thank you!

PR #34012, opened 2023-01-31
Updated the docker url version

sannanansari:
#33983


wilkinsona:
Thanks for the PR but we prefer to process single-line version upgrades like this ourselves.

Issue #10963, opened 2017-11-09
Consider making the default user's username and password configurable


wilkinsona:
Please see the discussion that starts [here](https://github.com/spring-projects/spring-boot/issues/7958#issuecomment-343102936) for details.

`AuthenticationManagerConfiguration` could consume some properties for configuring the default user's username and password. I'm not sure how useful it would be as it'll quickly back off when the user sets up their own security. Perhaps it's still of sufficient value for demos and the out-of-the-box experience as users are finding their feet?

mbhave:
I agree that it's only useful for demos/OOTB since real-world use-cases wouldn't use a single user. I don't particularly see any harm in bringing the property back but we should be careful about naming it since `security.basic.user` is no longer applicable.

rwinch:
Perhaps we could allow more than one user to be created? Boot would just create a `Map<String,String>` and then delegate to a Spring Security's `UserDetailsMapFactoryBean` that creates a `Collection<UserDetails>` from that `Map<String,String>`.

That would mean users could provide something like this:

```
spring:
  security:
    users:
        user: {noop}password,ROLE_USER
        admin: {noop}password,ROLE_USER,ROLE_ADMIN
```

This has the advantage of being able to provide multiple users and configure any User property. The disadvantage is that it might be more difficult to remember the syntax and the username and roles must be provided.

mbhave:
There was a similar discussion [here](#10349)

snicoll:
Thanks @mbhave, that discussion summarizes our concerns with map-based binding and magic strings breaking IDE support. 

I know @michael-simons has built a starter to bring back the single user configuration. Perhaps he can chime in and share his use cases?

michael-simons:
Hi.
My starter is [here](https://github.com/michael-simons/configurable-single-user-spring-boot-starter), still on M5. 

My use case is actually the single user, in several applications. For those it's enough to configure username, password and roles.

@dasniko had another production use case he described to me… I think it was passing the user properties from an external service during startup or something like that.

philwebb:
I don't mind properties to allow a single user to be configured quickly. It's going to help with migration, give us a nicer learning experience and it seems like it could be useful for demos. I'm tempted to say we should steer clear of the full `Map` based approach since it complicates things quite a bit.

philwebb:
We'll add properties back for single user support.

wilkinsona:
[...]

PR #34025, opened 2023-01-31
fix InMemoryUserDetailsManager: encode password if encoder is available

power721:
The InMemoryUserDetailsManager should encode generated password if the password encoder is provided.


wilkinsona:
Thanks for the proposal but the current behaviour is intentional. Please see https://github.com/spring-projects/spring-boot/issues/10963 for details.

Issue #27513, opened 2021-07-27
TaskExecutionAutoConfiguration falls into a hole when an ExecutorService is defined


jdmichal:
`TaskExecutionAutoConfiguration.applicationTaskExecutor` is conditional on missing bean `java.util.concurrent.Executor`.  This results in a `java.util.concurrent.ThreadPoolExecutor` bean stopping instantiation of the `applicationTaskExecutor`.

https://github.com/spring-projects/spring-boot/blob/1bca2bbc859b0a40b6b5616c0b6433e6047c36ca/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/task/TaskExecutionAutoConfiguration.java#L75-L81

However, a `ThreadPoolExecutor` is *not* a Spring `TaskExecutor`.  So eventually (somewhere) a stop-gap `SimpleAsyncTaskExecutor` is created, and then warnings start getting logged:

```
!!!
An Executor is required to handle java.util.concurrent.Callable return values.
Please, configure a TaskExecutor in the MVC config under "async support".
The SimpleAsyncTaskExecutor currently in use is not suitable under load.
-------------------------------
Request URI: '/api/whatever'
!!!
```

I believe there is an implicit assumption here that any `Executor` bean defined would *naturally* be a Spring `TaskExecutor`, because it's a Spring program.  This seems to me to be a bad assumption.

I propose one of the following:

1. Make the `applicationTaskExecutor` bean conditional on `TaskExecutor` instead of just `Executor`.  That way, other `Executor` or `ExecutorService` beans which are not also `TaskExecutor` will not prevent the `applicationTaskExecutor` from being created.
2. If an `Executor` is defined, create a `TaskExecutor` wrapping the given executor.  If an `ExecutorService` is defined, an `AsyncTaskExecutor` wrapper could be created instead.  Based on the interfaces, these wrappers would be very simple pass-through implementations, as the Spring interfaces are basically subsets of the Java ones.

Happy to help with pull requests, but I thought this warranted some discussion first.

jdmichal:
Ends up the wrapper necessary for (2) already exists:  `org.springframework.core.task.support.TaskExecutorAdapter`.

https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/core/task/support/TaskExecutorAdapter.html

[...]

Issue #15447, opened 2018-12-12
Improve TaskExecution documentation with custom Executor and Spring MVC


StasKolodyuk:
Hi!

If you declare any `Executor` in Spring Context it is picked up by `TaskExecutionAutoConfiguration` due to:
```
@Lazy
@Bean(name = APPLICATION_TASK_EXECUTOR_BEAN_NAME)
@ConditionalOnMissingBean(Executor.class)
public ThreadPoolTaskExecutor applicationTaskExecutor(TaskExecutorBuilder builder) {
	return builder.build();
}
```
If using in conjunction with `spring-boot-starter-webflux`, this leads to usage of your custom executor by `WebMvcAutoConfiguration.configureAsyncSupport()` and then by `ReactiveTypeHandler` printing the error.

```
2018-12-11 19:05:10.174  WARN 1612 --- [nio-9015-exec-1] o.s.w.s.m.m.a.ReactiveTypeHandler        :
!!!
Streaming through a reactive type requires an Executor to write to the response.
Please, configure a TaskExecutor in the MVC config under "async support".
The SimpleAsyncTaskExecutor currently in use is not suitable under load.
-------------------------------
Controller:	com.example.executorissue.controller.EchoController
Method:		echo
Returning:	reactor.core.publisher.Flux<java.lang.String>
!!!
```

Maybe it makes sense to use `@ConditionalOnMissingBean(name = APPLICATION_TASK_EXECUTOR_BEAN_NAME)` to allow custom Executors be created in Spring context and not being picked up by `TaskExecutionAutoConfiguration`?

See the example [here](https://github.com/spring-projects/spring-boot-issues/pull/84)

Thanks

snicoll:
@StasKolodyuk your analysis is incorrect as far as I can see and your sample does not match the log that you've provided. A few things:

* Your sample does not demonstrate that your custom `Executor` is picked up by `TaskExecutionAutoConfiguration`. Actually it was designed to not do that and I have no evidence of the  contrary
* That log you gave is based on an MVC app which doesn't match your description
* I have looked at your sample and it is running locally. That log you claim in the `README` doesn't happen 

Please update your sample so that it matches the problem you are describing.

StasKolodyuk:
@snicoll I've updated the example. I've created two tests, please compare the log output.

Sorry for confusion about what's actually going on under the hood.

Let's imagine some custom Executor in context. 
[...]

PR #34049, opened 2023-02-02
TaskExecutionAutoConfiguration should import task executor when missi…

QuantumXiecao:
Hi, when scanning the spring issues, I have found one problem related to spring-boot which has been depicted in this website(https://github.com/spring-projects/spring-framework/issues/29725).
In my opinion, TaskExecutionAutoConfiguration should import task executor only when develops not set one executor with specific name. It is very likely that develops add Executor bean not used for task executor--This would indirectly lead to spring use SimpleAsyncTaskExecutor to handle async events(which would always start one new thread when events come).
My PR is to solve this problem. There might be other solutions like: Extending @ConditionalOnMissingBean to support And/Or logic(Current implementation use Or logic--types or names or annotation) which is really one break change. 
I prefer to minimize the impact only within the TaskExecutionAutoConfiguration which leads to this PR.
Welcome to discuss! Thanks for you time!


wilkinsona:
Thanks for the proposal. The current arrangement is intentional and I don't think we should change it. Please see https://github.com/spring-projects/spring-boot/issues/15447 and https://github.com/spring-projects/spring-boot/issues/27513 for details. There's also a [note in the documentation about the current behavior](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.task-execution-and-scheduling).

PR #34050, opened 2023-02-02
Add RabbitTemplateCustomizer to allow customization of RabbitTemplates

dangzhicairang:
If there has a more convenient way to configure `ConfirmCallback` and `ReturnsCallback` for `RabbitTemplate` by auto-configure , like declare a bean of `ConfirmCallback` or `ReturnsCallback`, so we need not to declare a new bean of `RabbitTemplate` in extra `ConfigClass`.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
@@ -154,10 +154,9 @@ public class RabbitAutoConfiguration {
 		@Bean
 		@ConditionalOnSingleCandidate(ConnectionFactory.class)
 		@ConditionalOnMissingBean(RabbitOperations.class)
-		public RabbitTemplate rabbitTemplate(RabbitTemplateConfigurer configurer
-				, ConnectionFactory connectionFactory
-				, ObjectProvider<RabbitTemplate.ConfirmCallback> confirmCallback
-				, ObjectProvider<RabbitTemplate.ReturnsCallback> returnsCallback) {
+		public RabbitTemplate rabbitTemplate(RabbitTemplateConfigurer configurer, ConnectionFactory connectionFactory,
+				ObjectProvider<RabbitTemplate.ConfirmCallback> confirmCallback,
+				ObjectProvider<RabbitTemplate.ReturnsCallback> returnsCallback) {
 			RabbitTemplate template = new RabbitTemplate();
 			configurer.configure(template, connectionFactory);
 			confirmCallback.ifAvailable(ccb -> template.setConfirmCallback(ccb));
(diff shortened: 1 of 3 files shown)


dangzhicairang:
I got an error when visiting ci.spring.io/builds/240492.

Error code: 1020

Ray ID: 7931a65cbcdc96a1

Country: CN

Data center: sjc06

IP: 171.221.64.103

Timestamp: 2023-02-02 08:45:18 UTC

wilkinsona:
Our CI environment is being upgraded at the moment so it's likely to be unstable for a while.

philwebb:
I wonder if we should instead offer a `RabbitTemplateCustomizer` callback which could be used to attach a `ConfirmCallback`, `ReturnsCallback` or perform any other customization.

wilkinsona:
We already have the `RabbitTemplateConfigurer` that may confuse things a bit. That said, I'm increasingly favour of the `…Customizer` based approach over defining numerous beans that are only beans so that they can be set on another bean.

dangzhicairang:
> We already have the `RabbitTemplateConfigurer` that may confuse things a bit. That said, I'm increasingly favour of the `…Customizer` based approach over defining numerous beans that are only beans so that they can be set on another bean.

It's right, i will supply a RabbitTemplateCustomizer like @philwebb said.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
```diff
@@ -154,9 +154,11 @@ public RabbitTemplateConfigurer rabbitTemplateConfigurer(RabbitProperties proper
 		@Bean
 		@ConditionalOnSingleCandidate(ConnectionFactory.class)
 		@ConditionalOnMissingBean(RabbitOperations.class)
-		public RabbitTemplate rabbitTemplate(RabbitTemplateConfigurer configurer, ConnectionFactory connectionFactory) {
+		public RabbitTemplate rabbitTemplate(RabbitTemplateConfigurer configurer, ConnectionFactory connectionFactory,
+				ObjectProvider<RabbitTemplateCustomizer> rabbitTemplateCustomizers) {
 			RabbitTemplate template = new RabbitTemplate();
 			configurer.configure(template, connectionFactory);
```
ppapou, line 160: Just a cosmetic typo. Assume it's better to apply the configurator instead of configurer.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitTemplateCustomizer.java
```diff
@@ -0,0 +1,36 @@
...
+ * Callback interface that can be used to customize a {@link RabbitTemplate}.
+ *
+ * @author dang zhicairang
+ * @since 2.3.0
+ */
+@FunctionalInterface
+public interface RabbitTemplateCustomizer {
+
+	/**
+	 * Customize the {@link RabbitTemplate}.
```
ppapou, line 31: Minor note:
  It would be nice to have a bit more info of the method. What excactly it does?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitAutoConfiguration.java
dangzhicairang: Most places in Spring project use configurer, so I think it's best to leave things as they are

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitTemplateCustomizer.java
dangzhicairang: I will add more info for it.

mhalbritter:
Thank you!

Issue #33990, opened 2023-01-26
Update Customizing Sanitization as built-in key-based sanitization is no longer performed


wilkinsona:
https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/index.html#howto.actuator.sanitize-sensitive-values.customizing-sanitization

> Sanitization can be customized in two different ways.
>
> To take more control over the sanitization, define a SanitizingFunction bean. The SanitizableData with which the function is called provides access to the key and value as well as the PropertySource from which they came. This allows you to, for example, sanitize every value that comes from a particular property source. Each SanitizingFunction is called in order until a function changes the value of the sanitizable data. If no function changes its value, the built-in key-based sanitization is performed.

There's only one way now so I think the first and last sentence should be removed.

philwebb:
Closing in favor of PR #34051. Thanks @izeye!

PR #34051, opened 2023-02-02
Update Customizing Sanitization documentation as built-in key-based sanitization is no longer performed

izeye:
This PR simply tries to do what #33990 says.


mhalbritter:
Thanks!

PR #34063, opened 2023-02-04
Maven Plugin's PropertiesMergingResourceTransformer closes InputStream when it should not do so

enimiste:
Hi,
This is a conformity fix, as the interface `ReproducibleResourceTransformer` requires that "An input stream for the resource, the implementation should *not* close this stream" :
```java
public interface ReproducibleResourceTransformer
    extends ResourceTransformer
{
    /**
     * Transform an individual resource
     * @param resource The resource name
     * @param is An input stream for the resource, the implementation should *not* close this stream
     * @param relocators  A list of relocators
     * @param time the time of the resource to process
     * @throws IOException When the IO blows up
     */
    void processResource( String resource, InputStream is, List<Relocator> relocators, long time )
        throws IOException;
}
```
I think that the implementation shouldn't close the input stream.
Thanks


mhalbritter:
Thank you and congrats on your first spring boot contribution!

PR #34064, opened 2023-02-05
Add RetryTopic topic creation to autoconfiguration

### LINKED ISSUE #34046: Add auto-creation of retry topics to Spring Kafka non-blocking retries auto-configuration
When auto-configuration for `Spring for Apache Kafka`'s `Non-Blocking Delayed Retries` feature was implemented, automatic creation of topics was disabled by default.

We should allow users to override that configuration for speeding up development.

My proposal is to add the following properties:

```properties
spring.kafka.retry.topic.create.retry-topics=true
spring.kafka.retry.topic.create.partitions=2
spring.kafka.retry.topic.create.replication-factor=3
```

Or, if we want less options:

```properties
spring.kafka.retry.topic.auto-create-retry-topics=true
```

How does that look? I can open a PR once we reach an agreement, and we can change any properties names.

Here are the [docs](https://docs.spring.io/spring-kafka/reference/html/#retry-topic) for the feature, and [here](https://docs.spring.io/spring-kafka/reference/html/#topics-autocreation) for topics auto-creation. Please let me know if you have any questions.

@garyrussell and @artembilan, feel free to pitch in if you have the time.

Thanks!

tomazfernandes:
Resolves #34046


snicoll:
@tomazfernandes thank you for the PR but I find this single property a bit odd considering the three options a user can configure. Let's continue the conversation on the related issue, and thanks for the PR, in any case.

PR #34122, opened 2023-02-08
Add metadata for default for spring.jmx.registration-policy

izeye:
This PR adds configuration metadata for default value for `spring.jmx.registration-policy` property.

See gh-32573


wilkinsona:
Thanks very much, @izeye.

PR #34125, opened 2023-02-08
Add default-headers in autoconfigure for elasticsearch

icerockq:
Adds the ability to auto-configure elasticsearch with default headers. Without create own restclient configurations in the code.


wilkinsona:
Thanks for the proposal, @icerockq. Have you considered using a `RestClientBuilderCustomizer` bean for this? I'm not sure that there's a huge amount of additional value in allowing the headers to be configured using properties.

icerockq:
> [wilkinsona](/wilkinsona)

Of course I know about such functionality. But it seemed to me that this method would simplify the setup of the restClient. But after reflection, I admit that my changes may be redundant.

wilkinsona:
Thanks again for the proposal but I think we should leave this as it is for now. If we see significant demand for property-based configuration of the headers in the future we can consider this one again.

PR #34135, opened 2023-02-09
Upgrade CI to Docker 23.0.0 for 2.7.x

hellozrh:
Upgrade CI to Docker 23.0.0  for 2.7.x


wilkinsona:
Thanks for the PR but we prefer to process single-line version upgrades like this ourselves.

Issue #34172, opened 2023-02-13
Document ConfigurationPropertyCaching


jizhuozhi:
spring-boot version: 3.0.2 (since 2.3.0)

In commit 7afd25fc9a30e4e439791504eb8105acfde1d0cc, a cache was introduced to optimize the overhead caused by copying and deduplication of `getPropertyNames` objects.

But currently we have encountered some problems related to memory pressure. In the process of following the call stack, we found that the `cache` created in `SpringIterableConfigurationPropertySource` only exists in two states:
1. Never expires
2. Refresh forever

This is because there is no configuration in the `SpringIterableConfigurationPropertySource`, and there is no way to configure it (programmatic API or configuration file).

philwebb:
Are you able to provide some more details about the problems your facing? Do you have very large property sources or is there some other reason why the memory pressure happens? Perhaps you can share a sample application that replicates the problem.

The cache states are currently set based on if the property source is immutable. The assumption being that if the property source can't change, it's not a problem to keep the cache.

> ...there is no way to configure it (programmatic API or configuration file

Have you seen the `ConfigurationPropertyCaching` interface? You should be able to do something like this to configure things:

```java
ConfigurationPropertyCaching.get(sources, propertySource).setTimeToLive(timeToLive);
```

jizhuozhi:
Hello, @philwebb . Thanks for your reply :)

The actual usage scenario is a rules engine that relies on several keys in dynamic configuration, here is an example

```java
        String[] keys = new String[]{ "foo", "bar", "foobar" /* other elements... */};
        String[] values = Arrays.stream(keys).map(environment::getProperty).toArray(String[]::new);
        /* do something with values... */
```

Different rules consume different keys. The complete data collection contains tens of thousands of keys and has a high QPS.

At the same time, unfortunately, we are using G1's immature JDK version, and due to enterprise restrictions, we cannot upgrade it. Therefore, the young generation garbage collection algorithm is still Parallel New.

> Have you seen the ConfigurationPropertyCaching interface? You should be able to do something like this to configure things:

I did browse a lot of classes before this, but I paid too much attention to the `SpringIterableConfigurationPropertySource` class itself, but did not find the content related to `ConfigurationPropertyCaching`, which really solved my problem, but due to the lack of corresponding documents ( Including `javadoc` and various search engines) to introduce this way, I think other users may also encounter the same problem.

Just a suggestion, would it be very valuable to provide a user-accessible documentation.

[...]

PR #34173, opened 2023-02-13
Optimizing the ConfigurationPropertyCaching strategy

jizhuozhi:
In this commit, first tried enabling caching by default, if `neverExpire` is false.

Unlimited caching is not a good idea for any dynamic application, especially dynamic configuration. So next tried turning the default duration when enabled, 30s is a reasonable default value. It will neither cause memory pressure nor cause long-term reading of expired configuration.

See [gh-34172](https://github.com/spring-projects/spring-boot/issues/34172)


philwebb:
I'd like to get a bit more information in #34172 before we consider changing any defaults.

Issue #34059, opened 2023-02-03
Manifests of jars built with Gradle do not have a Implementation entries


scottfrederick:
The Spring Boot Maven plugin configures the `maven-jar-plugin` to add the default `Implementation-*` entries to the built archive. Archives built with the Spring Boot Gradle plugin do not have these manifest attributes by default. The Gradle plugin should add the following manifest entries for consistency: 

* `Implementation-Title`
* `Implementation-Version`
* `Implementation-Vendor`

This is similar to #32829, which we categorized as a bug and fixed in 3.0.x. We should fix this in the same version.

wilkinsona:
Maven uses the following defaults:

```
Implementation-Title: ${project.name}
Implementation-Version: ${project.version}
Implementation-Vendor: ${project.organization.name}
```

`project.name` and `project.version` are straightforward with Gradle. `project.organization.name` is not. Maven will only add `Implementation-Vendor` if `<organization>/<name>` appears in the pom` which, in my experience, isn't very common. I think we should just focus on the title and version.

PR #34177, opened 2023-02-14
Fix name for Implementation-Title in Spring Boot Gradle plugin

izeye:
This PR fixes the name for the `Implementation-Title` manifest entry in the Spring Boot Gradle plugin as `Implementation-Name` seems to have been typed accidentally.

See gh-34059


wilkinsona:
🤦 Thanks, @izeye!

wilkinsona:
Thanks again, @izeye.

PR #34180, opened 2023-02-14
Update Gradle Enterprise plugin

jprinet:


wilkinsona:
Thanks, @jprinet, but we prefer to apply this sort of one-line version upgrade ourselves. If we've missed an upgrade that you feel is important, please feel free to ping us on Gitter or Slack to give us a nudge.

PR #34216, opened 2023-02-16
defaultTracingObservationHandler is not ordered as documented

ttddyy:
The javadoc on `MicrometerTracingAutoConfiguration.DEFAULT_TRACING_OBSERVATION_HANDLER_ORDER` says it is a `@Order` value for the `defaultTracingObservationHandler` bean.
However, in the code, this constant is not applied to the `defaultTracingObservationHandler` bean definition.

This PR specifies the missing `@Order` on the `defaultTracingObservationHandler` bean.


wilkinsona:
Thanks very much, @ttddyy.

PR #34220, opened 2023-02-16
org.springframework.boot.web.embedded.jetty.GracefulShutdown uses the wrong class to create its logger

1993heqiang:
Polishing: Update this logger to use "GracefulShutdown.class".


mhalbritter:
Thanks!

PR #34316, opened 2023-02-21
Upgrade to Spring Integration 6.1.0-M1 #34308

shekharAggarwal:


wilkinsona:
Thanks for the PR but, as [noted in the pull request template](https://github.com/spring-projects/spring-boot/blame/main/.github/PULL_REQUEST_TEMPLATE.md#L16-L19), we prefer to handle single-line upgrades ourselves. Also, please check that your changes build cleanly before submitting them. In this case they do not as Spring Integration 6.1.0-M1 has not yet been released.

PR #34327, opened 2023-02-21
Added getting support in README.md

esivakumar18:
Added getting support in README.md


scottfrederick:
Thanks for the suggestion, but this is already covered in the [Getting Help](https://github.com/spring-projects/spring-boot#getting-help) section of the project README with inks spring.io and stackoverflow.com.

PR #34365, opened 2023-02-24
Configure maven.compiler.release in spring-boot-starter-parent to prevent impossible JDK API usage

mtheiss:
When cross compiling with newer JDKs against an older JDK API, it might happen, that a JDK API is used, which is not available in the intended Java version. Using maven.compiler.release will migitate this.

- [Maven Compiler Plugin Documentation](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release)
- [Baeldung Explanation for the release flag](https://www.baeldung.com/java-compiler-release-option)


mhalbritter:
Thank you!

PR #34421, opened 2023-03-01
Polish name of JettyServletWebServerFactory bean method

1993heqiang:
This PR renames the method name to conform to the naming convention.


wilkinsona:
Thanks very much, @1993heqiang.

Issue #34424, opened 2023-03-01
Maven plugin uses timezone-local timestamps when outputTimestamp is used


nielsbasjes:
# Summary
The `spring-boot-maven-plugin` `repackage` goal does not set a reproducible timestamp.
The provided timestamp is used but offset with the local timezone of the developer.

A project wants to be reproducible and sets the `project.build.outputTimestamp` to a desired value with a clear timezone in it.

Then the `spring-boot-maven-plugin` is used to `repackage` the jar/war file.
The `outputTimestamp` by default uses the correct timestamp from `project.build.outputTimestamp`.

The problem is that the timestamp in final jar/war file has this timestamp expressed in the timezone of the local build system.
So someone in a different timezone can only reproduce the build if they are able to figure out the timezone of the original build system.

This makes the build needlessly hard to reproduce.

# Reproduce
This mini project demonstrates this problem by building the same demo project twice in a docker image with a different timezone.

https://github.com/nielsbasjes/BugreportSpringPluginTimezone

# Expected
I expect the output to be in the specified timezone in the `outputTimestamp`
This will make the end result always be the same regardless of the timezone of the build system.

# Actual output:

For different timezones the output file is different
  
    $ md5sum *jar
    51101377d3c14d4def5607adc7694eef  amsterdam.jar
    bfc83311fec00946edde56b0313e642f  eucla.jar
    1208e63002398ff0606ab62a0b7f06a1  hawaii.jar

Snippets from diffoscope:

The content (all jar files and such) are identical

    │┄ Archive contents identical but files differ, possibly due to different compression levels. Falling back to binary comparison.

In timezone Europe/Amsterdam

    │ +drwxr-xr-x  2.0 unx        0 bX defN 11-Nov-11 23:22 META-INF/
    │ +-rw-r--r--  2.0 unx      379 bl defN 11-Nov-11 23:22 META-INF/MANIFEST.MF
    │ +drwxr-xr-x  2.0 unx        0 bl defN 11-Nov-11 23:22 org/
    │ +drwxr-xr-x  2.0 unx        0 bl defN 11-Nov-11 23:22 org/springframework/
    │ +drwxr-xr-x  2.0 unx        0 bl defN 11-Nov-11 23:22 org/springframework/boot/
    │ +drwxr-xr-x  2.0 unx        0 bl defN 11-Nov-11 23:22 org/springframework/boot/loader/
[...]

PR #34428, opened 2023-03-01
Fix plugin timezone dependency

nielsbasjes:
This is a proposed fix for the reproducibility problem described here
https://github.com/spring-projects/spring-boot/issues/34424

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/RepackageMojo.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/RepackageMojo.java
@@ -22,6 +22,7 @@ import java.nio.file.attribute.FileTime;
 import java.time.OffsetDateTime;
 import java.util.List;
 import java.util.Properties;
+import java.util.TimeZone;
 import java.util.concurrent.TimeUnit;
 import java.util.regex.Pattern;
 
@@ -207,6 +208,8 @@ public class RepackageMojo extends AbstractPackagerMojo {
 	}
 
 	private void repackage() throws MojoExecutionException {
+		TimeZone originalDefaultTimeZone = TimeZone.getDefault();
+		TimeZone.setDefault(TimeZone.getTimeZone("UTC"));
 		Artifact source = getSourceArtifact(this.classifier);
 		File target = getTargetFile(this.finalName, this.classifier, this.outputDirectory);
 		Repackager repackager = getRepackager(source.getFile());
@@ -219,6 +222,7 @@ public class RepackageMojo extends AbstractPackagerMojo {
 			throw new MojoExecutionException(ex.getMessage(), ex);
 		}
 		updateArtifact(source, target, repackager.getBackupFile());
+		TimeZone.setDefault(originalDefaultTimeZone);
 	}
 
 	private FileTime parseOutputTimestamp() {


nielsbasjes:
I installed this version of the plugin locally and then when I use that version in my reproduction project I actually get the right timestamps and all files (using all timezones in docker) are binary identical and the timestamp looks exact like what was configured:  2011-11-11 22:22 .

https://github.com/nielsbasjes/BugreportSpringPluginTimezone

scottfrederick:
@nielsbasjes I don't think manipulating the time zone like this is the right way to approach the fix.

Looking at the `.jar.original` files in your reproducing sample, the same timestamps are applied to both jars, independent of the local timezone. So Maven is doing the right thing. The code that Maven uses to parse the `project.build.outputTimeStamp` is here: https://github.com/apache/maven-archiver/blob/maven-archiver-3.6.0/src/main/java/org/apache/maven/archiver/MavenArchiver.java#L886-L887. 

Unlike Maven, Spring Boot does not apply the `withOffsetSameInstant(ZoneOffset.UTC)` modifier after `OffsetDateTime.parse(outputTimeStamp)` when parsing the date here: https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/RepackageMojo.java#L238. If applying this modifier fixes the problem, it looks like a cleaner solution to me. 

Would you like to try this change instead?

nielsbasjes:
Also: It is possible to provide the desired timestamp as epoch which cannot have a timestamp and thus a timestamp must be chosen. The current choice is the local system timezone, my proposal is to force (or make configurable?) this to be the UTC timezone.

philwebb:
I wonder if we can change this logic in [JarWriter.writeToArchive(...)](https://github.com/spring-projects/spring-boot/blob/0b23ffd73a55b89ae5944d36c964114daea5d49c/spring-boot-project/spring-boot-tools/spring-boot-loader-tools/src/main/java/org/springframework/boot/loader/tools/JarWriter.java#L92) so that `jarEntry.setLastModifiedTime` changes `this.lastModifiedTime` based on `Calendar.getInstance()`. That way we write an offset time with the knowledge that `ZipUtil.toDosTime` will change it back again.

nielsbasjes:
@philwebb So you shift the epoch timestamp using the current timezone, because you expect the underlying code to shift it back again? I would vote against such code.

philwebb:
Yeah. The reason I prefer it is because `TimeZone.setDefault(...)` with change the default timezone for the entire JVM. If Maven is doing something else in another thread (now or in the future) then the default timezone could shift under its feet.

scottfrederick:
> `TimeZone.setDefault(...)` with change the default timezone for the entire JVM

That was my concern too, that this could have unpredictable side-effects that are difficult to diagnose.

nielsbasjes:
Yes that is true.
Perhaps marking this plugin as not threadsafe?

nielsbasjes:
I'll have a look tomorrow if I can make this nasty time shifting work reliable.
Speaking of timezones .. it is already tomorrow over here ...

philwebb:
I discovered the javadoc for `setTime` is much better than `setLastModifiedTime` and actually describes that the default TimeZone is used when writing values. I've switched to this method and gone with the shift approach. The fix is in commit 998d59b7ac1a75b26634e4fd2843a7833e554840.

Thanks for raising the issue and providing the analysis.

PR #34393, opened 2023-02-27
Add build-info skip support

didiez:
Allow skip `build-info` maven goal via `-Dspring-boot.build-info.skip`, as other spring-boot maven goals do, e.g. `-Dspring-boot.repackage.skip`

PR #34482, opened 2023-03-05
Add Javadoc since to BuildInfoMojo.skip

izeye:
This PR adds Javadoc `@since` tag to the `BuildInfoMojo.skip` parameter.

See gh-34393


wilkinsona:
Thanks very much, @izeye.

PR #34503, opened 2023-03-07
Remove final from ProblemDetailsExceptionHandler classes to allow proxying

vy:
Removes `final` modifier from `ProblemDetailsExceptionHandler` implementations to allow proxying these beans. See #34426 for the issue.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-autoconfigure/build.gradle
@@ -221,6 +221,7 @@ dependencies {
 	testImplementation("com.querydsl:querydsl-core")
 	testImplementation("com.squareup.okhttp3:mockwebserver")
 	testImplementation("com.sun.xml.messaging.saaj:saaj-impl")
+	testImplementation("com.tngtech.archunit:archunit-junit5:0.22.0")
 	testImplementation("io.projectreactor:reactor-test")
 	testImplementation("io.r2dbc:r2dbc-h2")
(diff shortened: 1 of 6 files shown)


philwebb:
I wonder if we should add some tests just to stop `final` sneaking back in the future?

philwebb:
Thanks @vy! Rather than a new dependency we might change this to use standard reflection when we merge it.

vy:
@philwebb, ArchUnit was already used in `TaskConfigurationAvoidanceTests`, hence was my preference for it. But I can replace it with reflection, if that is what you want.

If you prefer reflection, please also let me know if you want to target `ProblemDetailsExceptionHandler` _interfaces_ or `ResponseEntityExceptionHandler` _subclasses_, since the latter will require some stunt to pull.

wilkinsona:
Thanks very much for the PR, @vy. I decided to test the changes by verifying that the exception handlers could be proxied. If you're interested, please see https://github.com/spring-projects/spring-boot/commit/1eb5bbe3ea638e292463a5eca2c3adfc38c4450d for details.

vy:
Thanks so much for the prompt processing of the ticket and keeping me in the loop. :bow: I didn't know about `AopUtils.isCglibProxy`, which makes your tests precisely demonstrative of the bug reported in the first place. :100:

sbrannen:
@wilkinsona, at a glance it appears that those two `assertThat(AopUtils.isCglibProxy(context.getBean(ProblemDetailsExceptionHandler.class)));` assertions are missing `.isTrue()`.

PR #34509, opened 2023-03-08
Upgrade to spring-asciidoctor-backends 0.0.5

sannanansari:
#34432


wilkinsona:
Thanks, @sannanansari, but as with #34012 we prefer to process single-line version upgrades like this ourselves.

PR #34511, opened 2023-03-08
28798: created datetime no longer used in generating hash for filename. Also made default time now instead of windows epoch

plebcity:


wilkinsona:
Thanks for the proposal, @plebcity. Unfortunately, I don't think this is the right way to implement this. The most important thing is that the the behaviour needs to be configurable and the current behaviour must remain the default. As indicated by the pending design work label on 28978, we don't know exactly how this will look. I suspect we'll add something that's the  equivalent of `--creation-time` in the pack CLI but we won't know the answer or be able to proceed with the enhancement until we've done the required design work. Thank you for your patience in the meantime.

quaff:
> Thanks for the proposal, @plebcity. Unfortunately, I don't think this is the right way to implement this. The most important thing is that the the behaviour needs to be configurable and the current behaviour must remain the default. As indicated by the pending design work label on 28978, we don't know exactly how this will look. I suspect we'll add something that's the equivalent of `--creation-time` in the pack CLI but we won't know the answer or be able to proceed with the enhancement until we've done the required design work. Thank you for your patience in the meantime.

I don't think **current behaviour must remain the default**, nobody want creation time to be WINDOWS EPOCH.

vpavic:
> nobody want creation time to be WINDOWS EPOCH

This couldn't be any further from the truth. Many of us highly value reproducibility and prefer it as a default wherever possible.

I'm saying this as someone who doesn't use Spring Boot's support for building Docker images, but rather perfers Jib -  a build plugin that does [the same thing](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#reproducible-build-timestamps) by default.

quaff:
> > nobody want creation time to be WINDOWS EPOCH
> 
> This couldn't be any further from the truth. Many of us highly value reproducibility and prefer it as a default wherever possible.
> 
> I'm saying this as someone who doesn't use Spring Boot's support for building Docker images, but rather perfers Jib - a build plugin that does [the same thing](https://github.com/GoogleContainerTools/jib/tree/master/jib-gradle-plugin#reproducible-build-timestamps) by default.

Jib use UNIX rather than WINDOWS EPOCH, I'm curious why 1980 not 1970 at first time.

PR #34518, opened 2023-03-08
improve OpenTelemetry config

zeitlinger:
- configure OpenTelemetry via opentelemetry-sdk-extension-autoconfigure
- use opentelemetry-micrometer-1.5 bridge

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OpenTelemetryMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OpenTelemetryMetricsExportAutoConfiguration.java
@@ -1,0 +1,56 @@
+/*
+ * Copyright 2012-2022 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -1025,10 +1025,10 @@ bom {
 			]
 		}
 	}
-	library("OpenTelemetry", "1.19.0") {
+	library("OpenTelemetry", "1.23.1-alpha") {
 		group("io.opentelemetry") {
 			imports = [
-				"opentelemetry-bom"
+				"opentelemetry-bom-alpha"
```
jonatan-ivanov, line 1031: I'm not sure a stable version of Boot should depend on something that is unstable.
Also, right now upgrading OTel to anything higher than 1.20.1 will break baggage propagation in Micrometer Tracing.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OpenTelemetryMetricsExportAutoConfiguration.java
```diff
@@ -0,0 +1,56 @@
...
+		after = MetricsAutoConfiguration.class)
+@ConditionalOnBean({Clock.class})
+@ConditionalOnClass(OpenTelemetryMeterRegistry.class)
+@ConditionalOnEnabledMetricsExport("otlp")
+@Import(org.springframework.boot.actuate.autoconfigure.opentelemetry.OpenTelemetryAutoConfiguration.class)
+public class OpenTelemetryMetricsExportAutoConfiguration {
+
+	@Bean
+	@ConditionalOnMissingBean
+	public MeterRegistry openTelemetryMeterRegistry(OpenTelemetry openTelemetry, Clock clock) {
```
jonatan-ivanov, line 52: I think this should be auto-configured by OpenTelemetry I might be mistaken but I think they do. Also, Micrometer has an OTLP registry so if you want an OTLP output, you can use that (its auto-configured).

Issue #34405, opened 2023-02-28
Loading application.yml fails with NoSuchMethodError when using SnakeYAML 2.0


flashvayne:
When I upgrade snakeyaml from 1.33 to 2.0 version, Springboot Application run failed as below.
I have tried springboot 2.7.4 and 3.0.0, neither works.
Also I have tried JDK 8,11 and 17, none works.
+ snakeyaml dependency
```
<dependency>
    <groupId>org.yaml</groupId>
    <artifactId>snakeyaml</artifactId>
    <version>2.0</version>
</dependency>
```
+ Error log
```
11:28:31.187 [main] ERROR org.springframework.boot.SpringApplication - Application run failed
java.lang.NoSuchMethodError: org.yaml.snakeyaml.representer.Representer: method 'void <init>()' not found
    at org.springframework.boot.env.OriginTrackedYamlLoader.createYaml(OriginTrackedYamlLoader.java:76)
    at org.springframework.boot.env.OriginTrackedYamlLoader.createYaml(OriginTrackedYamlLoader.java:71)
    at org.springframework.beans.factory.config.YamlProcessor.process(YamlProcessor.java:164)
    at org.springframework.boot.env.OriginTrackedYamlLoader.load(OriginTrackedYamlLoader.java:84)
    at org.springframework.boot.env.YamlPropertySourceLoader.load(YamlPropertySourceLoader.java:50)
    at org.springframework.boot.context.config.StandardConfigDataLoader.load(StandardConfigDataLoader.java:54)
    at org.springframework.boot.context.config.StandardConfigDataLoader.load(StandardConfigDataLoader.java:36)
    at org.springframework.boot.context.config.ConfigDataLoaders.load(ConfigDataLoaders.java:107)
    at org.springframework.boot.context.config.ConfigDataImporter.load(ConfigDataImporter.java:128)
    at org.springframework.boot.context.config.ConfigDataImporter.resolveAndLoad(ConfigDataImporter.java:86)
    at org.springframework.boot.context.config.ConfigDataEnvironmentContributors.withProcessedImports(ConfigDataEnvironmentContributors.java:116)
    at org.springframework.boot.context.config.ConfigDataEnvironment.processInitial(ConfigDataEnvironment.java:240)
    at org.springframework.boot.context.config.ConfigDataEnvironment.processAndApply(ConfigDataEnvironment.java:227)
[...]

PR #34569, opened 2023-03-10
Replaces deprecated snakeyaml Representer constructor

kvmw:
Because of the recent CVEs in SnakeYAML, many projects need to bump snake-yaml version to 2.0 or higher.  But some deprecated codes in 1.33 have been removed in 2.0 which blocks the upgrade. This PR is about one of them which is used in Spring Boot 2.7.x source code. 

This is essentially the same change as [this commit](https://github.com/spring-projects/spring-boot/commit/bf5bd4f91c8b5617b589795b4db6a2371b26d5f5
) in 3.0 source but without dependency upgrade.


wilkinsona:
Thanks for the PR. Your fork appears to be a little bit out of date. This has already been fixed in #34405.

Issue #34324, opened 2023-02-21
Image builds with podman fail when image buildpacks are configured


Janrupf:
Following the documentation on how to configure podman for spring boot I ended up using the following configuration in gradle:
```kotlin
tasks.withType<org.springframework.boot.gradle.tasks.bundling.BootBuildImage> {
    environment.put("BP_NATIVE_IMAGE", "false")
    buildpacks.set(listOf(
        "gcr.io/paketo-buildpacks/ca-certificates:3.5.1",
        "gcr.io/paketo-buildpacks/bellsoft-liberica:9.10.0",
        "gcr.io/paketo-buildpacks/syft:1.23.0",
        "gcr.io/paketo-buildpacks/executable-jar:6.5.0",
        "gcr.io/paketo-buildpacks/spring-boot:5.20.0"
    ))
    val dockerHost = System.getenv("DOCKER_HOST")

    if (dockerHost != null) {
        docker {
            host.set(dockerHost)
            bindHostToBuilder.set(true)
        }
    }
}
```

This first appears to work, spring boot successfully communicates with the podman daemon and then pulls the buildpacks. However, when attempting to start the build the following error occurs:
```
> Task :bootBuildImage
Building image 'my.company/software:fdc1ac23'
 > Pulling builder image 'docker.io/paketobuildpacks/builder:base' ..................................................
 > Pulled builder image 'docker.io/paketobuildpacks/builder@sha256:11dcec699e2349838762e245c1c961d9e2c6e10d3e3617ec1e7bba19929caa11'
 > Pulling run image 'docker.io/paketobuildpacks/run:base-cnb' ..................................................
 > Pulled run image 'docker.io/paketobuildpacks/run@sha256:b9cbbd4e962acac787cc2bdba8b9c97996675ef57155418c0c5c9cb088261434'
 > Pulling buildpack image 'gcr.io/paketo-buildpacks/ca-certificates:3.5.1' ..................................................
 > Pulled buildpack image 'gcr.io/paketo-buildpacks/ca-certificates@sha256:ad50eae8fa996f0bb7c85f48135e81121fb1a5f8c9e98cf57703a623e07bd479'
 > Pulling buildpack image 'gcr.io/paketo-buildpacks/bellsoft-liberica:9.10.0' ..................................................
[...]

PR #34587, opened 2023-03-12
Polish Javadoc since tags for gh-34324

izeye:
This PR polishes Javadoc `@since` tags for gh-34324.

See gh-34324


wilkinsona:
Thanks very much, @izeye.

PR #34590, opened 2023-03-12
Include setter-only properties in metadata

### LINKED ISSUE #34309: @ConfigurationProperties on @Bean does not expose all properties
The Spring Configuration Annotation Processor does not seem to properly pick up setters on some objects.

Specifically:

```java
@ConfigurationProperties(prefix = "listener")
@Bean
SimpleMessageListenerContainer container(){
    return new SimpleMessageListenerContainer();
}
```

only discovers:

Property | Type |
-- | -- |
listener.acknowledge-mode | org.springframework.amqp.core.AcknowledgeMode |  
listener.auto-startup | java.lang.Boolean |  
listener.channel-transacted | java.lang.Boolean |  
listener.connection-factory | org.springframework.amqp.rabbit.connection.ConnectionFactory |  
listener.consumer-arguments | java.util.Map<java.lang.String,java.lang.Object> |  
listener.consumer-batch-enabled | java.lang.Boolean |  
listener.expose-listener-channel | java.lang.Boolean |  
listener.listener-id | java.lang.String |  
listener.phase | java.lang.Integer |  
listener.possible-authentication-failure-fatal | java.lang.Boolean |  
listener.queue-names | java.lang.String[]

but is missing properties like concurrentConsumerCount, which only has a setter, but no getter.

This is quite unintuitive, as `@ConfigurationProperties` on `@Bean` will set those properties.

Reproducer: https://github.com/ciis0/spring-boot-34309

criztovyl:
`@ConfigurationProperties` on a `@Bean` method allows to set properties based on setters, so those properties should also be included in metadata.

Fixes #34309.

It seems that not exposing setter-only properties is by design, some test actually test that; I have ~not~ adjusted them ~yet~, but intend to do that, still wanted I wanted to share my code early to get some feedback on this already, and not go to far before verifying this is actually wanted.

I started this mainly for collecting experience using this specific problem I have -- If there is already a alternative implementation in the works I have no problem stopping here. :)

I think this will expose a number of (leaf) properties that have no string representation (and thus are likely not useful for IDEs nor reports), I'll check if it is possible to exclude such properties in this scope, or if it is too big for this and needs to be done/discussed separately.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/PropertyDescriptorResolver.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/PropertyDescriptorResolver.java
@@ -19,6 +19,7 @@ package org.springframework.boot.configurationprocessor;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Optional;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
@@ -106,6 +107,18 @@ class PropertyDescriptorResolver {
 			register(candidates, new JavaBeanPropertyDescriptor(type, factoryMethod, getter, name, propertyType, field,
 					members.getPublicSetter(name, propertyType)));
 		});
+		// Also check based on setters, to mirror behavior of @ConfigurationProperties on
+		// @Bean
+		members.getPublicSetters().forEach((name, setters) -> {
+			VariableElement field = members.getFields().get(name);
+			// use Optional for fields that only have a setter
+			Optional<ExecutableElement> matchingGetter = Optional.ofNullable(members.getPublicGetters().get(name))
+				.map((g) -> findMatchingGetter(members, g, field));
+			ExecutableElement setter = findMatchingSetter(members, setters, field);
+			TypeMirror propertyType = setter.getParameters().get(0).asType();
+			register(candidates, new JavaBeanPropertyDescriptor(type, factoryMethod, matchingGetter.orElse(null), name,
+					propertyType, field, members.getPublicSetter(name, propertyType)));
+		});
 		// Then check for Lombok ones
 		members.getFields().forEach((name, field) -> {
 			TypeMirror propertyType = field.asType();
@@ -121,6 +134,14 @@ class PropertyDescriptorResolver {
 			VariableElement field) {
 		if (candidates.size() > 1 && field != null) {
 			return members.getMatchingGetter(candidates, field.asType());
+		}
+		return candidates.get(0);
+	}
+
+	private ExecutableElement findMatchingSetter(TypeElementMembers members, List<ExecutableElement> candidates,
+			VariableElement field) {
+		if (candidates.size() > 1 && field != null) {
+			return members.getMatchingSetter(candidates, field.asType());
 		}
 		return candidates.get(0);
 	}
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeElementMembers.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeElementMembers.java
@@ -113,7 +113,7 @@ class TypeElementMembers {
 		return getMatchingAccessor(candidates, type, ExecutableElement::getReturnType);
 	}
 
-	private ExecutableElement getMatchingSetter(List<ExecutableElement> candidates, TypeMirror type) {
+	ExecutableElement getMatchingSetter(List<ExecutableElement> candidates, TypeMirror type) {
 		return getMatchingAccessor(candidates, type, (candidate) -> candidate.getParameters().get(0).asType());
 	}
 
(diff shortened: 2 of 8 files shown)


spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/PropertyDescriptorResolver.java
```diff
@@ -106,6 +107,18 @@ Stream<PropertyDescriptor<?>> resolveJavaBeanProperties(TypeElement type, Execut
 			register(candidates, new JavaBeanPropertyDescriptor(type, factoryMethod, getter, name, propertyType, field,
 					members.getPublicSetter(name, propertyType)));
 		});
+		// Also check based on setters, to mirror behavior of @ConfigurationProperties on
+		// @Bean
```
criztovyl, line 111:
```suggestion
		// Also check based on setters
		// to mirror behavior of @ConfigurationProperties on @Bean
```

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/PropertyDescriptorResolver.java
```diff
@@ -125,6 +138,14 @@ private ExecutableElement findMatchingGetter(TypeElementMembers members, List<Ex
 		return candidates.get(0);
 	}
 
+	private ExecutableElement findMatchingSetter(TypeElementMembers members, List<ExecutableElement> candidates,
```
criztovyl, line 141: this is a a copy of findMatchingGetter, maybe a generalization for getters and setters could be introduced?

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeElementMembers.java
```diff
@@ -113,7 +113,7 @@ ExecutableElement getMatchingGetter(List<ExecutableElement> candidates, TypeMirr
 		return getMatchingAccessor(candidates, type, ExecutableElement::getReturnType);
 	}
 
-	private ExecutableElement getMatchingSetter(List<ExecutableElement> candidates, TypeMirror type) {
+	ExecutableElement getMatchingSetter(List<ExecutableElement> candidates, TypeMirror type) {
```
criztovyl, line 116: match visibility of getMatchingGetter

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/main/java/org/springframework/boot/configurationprocessor/TypeElementMembers.java
```diff
@@ -197,6 +197,10 @@ Map<String, List<ExecutableElement>> getPublicGetters() {
 		return Collections.unmodifiableMap(this.publicGetters);
 	}
 
+	Map<String, List<ExecutableElement>> getPublicSetters() {
+		return Collections.unmodifiableMap(this.publicSetters);
+	}
+
```
criztovyl, line 203: copy of getPublicGetters, but here a generalization most likely is not worth it.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/test/java/org/springframework/boot/configurationprocessor/ConfigurationMetadataAnnotationProcessorTests.java
```diff
@@ -112,8 +112,11 @@ void simpleProperties() {
 			.fromSource(SimpleProperties.class)
 			.withDescription("A simple flag.")
 			.withDeprecation(null, null));
+		assertThat(metadata).has(Metadata.withProperty("simple.number", Long.class)
+				.fromSource(SimpleProperties.class)
+				.withDescription("A setter-only property.")
+		);
```
criztovyl, line 118: I have renamed the property, a "counter", I think of a number of requests processed, most likely should not be manipulated from the outside, but a "number", I think of a concurrency limit, is legal and useful to manipulate.

spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/test/java/org/springframework/boot/configurationprocessor/JavaBeanPropertyDescriptorTests.java
```diff
@@ -110,17 +110,17 @@ void javaBeanSimplePropertyWithOnlyGetterShouldNotBeExposed() throws IOException
...
+			assertThat(property.getName()).isEqualTo("number");
 			assertThat(property.getSource()).isSameAs(property.getGetter());
 			assertThat(property.getGetter()).isNull();
-			assertThat(property.getSetter().getSimpleName()).hasToString("setCounter");
-			assertThat(property.isProperty(metadataEnv)).isFalse();
+			assertThat(property.getSetter().getSimpleName()).hasToString("setNumber");
+			assertThat(property.isProperty(metadataEnv)).isTrue();
 			assertThat(property.isNested(metadataEnv)).isFalse();
 		});
 	}
```
criztovyl, line 126: renamed property & inverted the check

philwebb:
I'm really sorry @criztovyl, but after discussing this today on the team call we no longer feel like generating meta-data for setters alone is the right approach. See https://github.com/spring-projects/spring-boot/issues/34309#issuecomment-1470796893 for more details.

I'm sorry about the wasted effort here. I should not have marked the issue as a bug and assigned it to a milestone in the first place.

criztovyl:
No worries, I was aware this effort might not pay off :)
I still learnt a thing or two, all good.

PR #34613, opened 2023-03-15
Avoid repeated class lookups in tests

dreis2211:
Hi 👋 

I've been profiling our test suites these days and noticed a very small, but probably avoidable, overhead in `RestDocsTestExecutionListener`

image

Feel free to decline if this is problematic in any scenario I'm currently not seeing. It's really minor.

Cheers,
Christoph


wilkinsona:
Thanks, @dreis2211. I don't foresee any problems with your proposed change. I have no idea why I wrote it like this originally. I guess I just thought that the CPU needed something to do…

snicoll:
Thanks again Christoph!

> Would you mind if I push another commit that does the same for MockWebServiceServerTestExecutionListener

I edited your commit to fix the check there.

PR #34619, opened 2023-03-16
Update filter method 

Java4ye:
Update filter method for adding isTraceEnabled condition to reduce calls to nanoTime method


mhalbritter:
What's the problem with `System.nanoTime`?

mhalbritter:
I get that, but why remove the call to `System.nanoTime()`? I mean I get that this variable is only used when logging on trace, but have you observed a performance problem with it?

Java4ye:
@mhalbritter 
I didn't do that, but I found it in the for loop.

wilkinsona:
Until something shows up as a performance bottleneck, I think we should leaves things as they are. It may even be that the `isTraceEnabled()` check is slower than the call to `System.nanoTime()`. Thanks anyway for the proposal.

PR #34633, opened 2023-03-18
fix CentralDirectoryFileHeader class omit Cloneable

SeasonPanPan:
The CentralDirectoryFileHeader class override Object.clone() method but does not implement Cloneable. 
I'm not sure if this was intentionally not done by the developer, or if it was a mistake.
pls check it.


wilkinsona:
I don't think there's any need for `CentralDirectoryFileHeader` to implement `Cloneable`. It's a package-private class and things are working fine as they are, including the single call to its `clone()` method. Thanks anyway for the proposal.

Issue #20402, opened 2020-03-05
Allow @DefaultValue to be used in fields


ttddyy:
Currently, `@DefaultValue` is limited to constructor arguments.
This prevents auto generating constructor with lombok.

If `@DefaultValue` can be assigned to fields, immutable properties class can be written like this:

```java
@Getter
@AllArgsConstructor
@ConfigurationProperties(prefix = "my")
@ConstructorBinding
public class MyProperties {

	@DefaultValue("Foo")
	private final String name;
}

snicoll:
I have some sympathy for that but enabling `@DefaultValue` on field opens up the possibility of the annotation to be used in non-lombok classes and I am not keen to do that.

Thanks for the suggestion anyway.

PR #34635, opened 2023-03-18
Allow DeprecatedConfigurationProperty on fields for Lombok

larsgrefer:
There is currently no elegant way to use `@DeprecatedConfigurationProperty` in order to provide additional deprecation metadata for a property when the `@ConfigurationProperties` class is built using Lombok.

This change allows `@DeprecatedConfigurationProperty` to be used directly on a field when using Lombok.


wilkinsona:
Thanks for the PR, @larsgrefer. I see you've targeted it at 2.7.x. I'm afraid that isn't an option for an enhancement such as this. It can only be included in a new minor or major release.

https://github.com/spring-projects/spring-boot/issues/20402 is related to this. It was a similar request but for `@DefaultValue`. IMO, a similar argument applies here. Also, give the Java 17 baseline and the availability of records in any release in which this enhancement could be added, I suspect that fewer and fewer people will want to use Lombok for configuration property classes.

Given the above, I don't think we should merge this. Flagging for team attention to see what everyone else thinks.

snicoll:
I agree with the reasoning here and I don't think we should merge this either. Thanks for the PR, in any case.

Issue #34638, opened 2023-03-19
Add 'bright' AnsiColor variants to Logback and Log4J ColorConverters


nightswimmings:
Currently 
`org.springframework.boot.logging.logback.ColorConverter` does not support "bright" colors despite `AnsiColor` does contain codes for all of them. Being a very old class, I have the feeling that the reason is simply nobody updated the available colors, but I might be wrong and there might be a rationale to disallow those

wilkinsona:
Thanks for the suggestion but we won't be making any enhancements to `AnsiColors` as it was [removed in 3.0](https://github.com/spring-projects/spring-boot/issues/32209).

philwebb:
Although we removed `AnsiColors`, we still have `AnsiColor` and we still support color codes in Logback and Log4J2. I think we could consider an update to the `ColorConverter` classes if someone wants to contribute them.

nightswimmings:
Maybe it's too unrelated, but IMHO it would not hurt to add 

<logger name="org.testcontainers" level="INFO" />
<logger name="com.github.dockerjava" level="WARN"/>

to org/springframework/boot/logging/logback/defaults.xml 

being that much of an ubiquitous library nowadays and considering there are other logger defaults :)

Abhijeetmishr:
@philwebb @wilkinsona Can I pick this up ?

wilkinsona:
@nightswimmings, that's really unrelated to this issue and, unfortunately, I don't think it will work as Testcontainers performs its logging before Spring Boot has a chance to initialise the logging system.

omlip:
Hi,
Is there any reason that White and black color are omit from the CodeConverter class ?

`@Plugin(name = "color", category = PatternConverter.CATEGORY)
@ConverterKeys({ "clr", "color" })
public final class ColorConverter extends LogEventPatternConverter {

	private static final Map<String, AnsiElement> ELEMENTS;

	static {
		Map<String, AnsiElement> ansiElements = new HashMap<>();
		ansiElements.put("faint", AnsiStyle.FAINT);
		ansiElements.put("red", AnsiColor.RED);
		ansiElements.put("green", AnsiColor.GREEN);
		ansiElements.put("yellow", AnsiColor.YELLOW);
		ansiElements.put("blue", AnsiColor.BLUE);
		ansiElements.put("magenta", AnsiColor.MAGENTA);
		ansiElements.put("cyan", AnsiColor.CYAN);
[...]

PR #34676, opened 2023-03-21
#34638 Added handling of bright ANSI colors in log4j and logback

krzyk:
For #34638.
Added handling of "bright" variants of colors for logback and log4j ColorConverter. Added also black and white (with bright variants) as those were also missing.


wilkinsona:
Thanks, @krzyk, but as discussed in the issue, we'd like to let @Abhijeetmishr work on this one.

PR #34697, opened 2023-03-21
Clarify servlet container compatibility

jgslima:
Fix documentation section on web container requirements for 2.7.x.

This is relevant for teams and companies that are planning upgrade to Spring Boot 3, upgrade the web container used, and so on.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/system-requirements.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/system-requirements.adoc
@@ -37,4 +37,4 @@ Spring Boot supports the following embedded servlet containers:
 | 4.0
 |===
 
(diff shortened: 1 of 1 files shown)


jgslima:
As a side note, this highlights an import aspect of environments that use a web container like Tomcat, to run .war applications (which is the case for many customers of the company I work for).

When upgrading to Spring Boot 3, Tomcat 10.1 must be upgraded (and not before) and also, **all** .war applications deployed in the same web container must be upgraded to Servlet API 5.0+.

All due to the `jakarta` namespace change.

wilkinsona:
I think we could simplify this a bit, by replacing the `+` with `or 4.0`:

> You can also deploy Spring Boot applications to any Servlet 3.1 or 4.0 compatible container

If we want to go into any more detail, we could then explain that Servet 5.0 broke backwards compatibility but I'm not sure it's necessary.

wilkinsona (approved):

snicoll:
@jgslima thank you for making your first Spring Boot contribution.

Issue #34678, opened 2023-03-21
Hints for including Liquibase changelogs in a native image are unnecessarily narrow


Ferioney:
Hi!

Currently I'm working on a migration application that uses Liquibase to Spring Boot Native (SB 3.0.4).
My Liquibase resources include the master file and a few sql files:

```
/resources/db/changelog/
    /tables/init.sql 
    db.changelog-master.yaml
```

The master file includes `/tables/init.sql` file.
When running application in native mode I have the error:
```
Caused by: java.io.FileNotFoundException: The file db/changelog/tables/init.sql was not found in the configured search path:
- Spring classpath
More locations can be added with the 'searchPath' parameter.
```

As I can see the master file was added to resources via hint: [LiquibaseAutoConfigurationRuntimeHints](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java#L178)
but other files were not included.

yep, I can add my custom hint with resources (like in [spring-aot-smoke-tests#LiquibaseRuntimeHints](https://github.com/spring-projects/spring-aot-smoke-tests/blob/main/boot/liquibase/src/main/java/com/example/liquibase/LiquibaseApplication.java#L22) ) but maybe you have a from-the-box solution which I didn't find.

Thanks!

wilkinsona:
For Flyway, we use the following pattern for resource inclusion:

```
hints.resources().registerPattern("db/migration/*");
```

By contrast, we use the following for Liquibase:

```
hints.resources().registerPattern("db/changelog/db.changelog-master.yaml");
```

For consistency and utility, we should align the later with the former and use `db/changelog/*`.

wilkinsona:
Closing in favor of #34729. Thanks for the PR, @Ferioney.

PR #34729, opened 2023-03-23
Hints for including Liquibase changelogs in a native image are unnecessarily narrow 

Ferioney:
gh-34678


snicoll:
@Ferioney thank you for making your first contribution to Spring Boot.

PR #34761, opened 2023-03-25
Remove source and target as maven.compiler.release is already set

arend-von-reinersdorff:
Follow-up on #34365.

When the compiler's release argument is set, arguments source and target are not used. No need to keep them around.

See https://docs.oracle.com/en/java/javase/17/docs/specs/man/javac.html#option-release
> Note: When using --release, you cannot also use the --source/-source or --target/-target options.

Maven actually allows setting both. But using both directly with javac would result in an error.

### DIFF
--- a/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
+++ b/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
@@ -21,8 +21,6 @@ publishing.publications.withType(MavenPublication) {
 			properties {
 				delegate."java.version"('17')
 				delegate."resource.delimiter"('@')
-				delegate."maven.compiler.source"('${java.version}')
-				delegate."maven.compiler.target"('${java.version}')
 				delegate."maven.compiler.release"('${java.version}')
 				delegate."project.build.sourceEncoding"('UTF-8')
 				delegate."project.reporting.outputEncoding"('UTF-8')


paulvi:
Indeed a073ef80e8d2ba732c10c1a79747767a3a2459a4 was not enough, 
removing 2 older option in 1f71fb5a8f990c97fd4b75906d40a6e3f395ea18 is correct

Note that with javac --release option
it is possible to to target even older releases 7,8 (while --release was introduced in JDK 9)

e.g. for JDK 19
```
  --release <release>
        Compile for the specified Java SE release. Supported releases: 7, 8, 9, 10, 11, 12, 13, 14, 15, 16, 17, 18, 19
```

I wonder it is possible to target Spring and Spring Boot to JDK 8 via this cross compilation?

But for sure simpler project can use this to target jdk 8 or jdk 11

paulvi:
And I am sure this switch to --release option should be refrected somewhere in docs
as in real world Spring Boot projects ther eis still mix of using
```
    <maven.compiler.source>${java.version}</maven.compiler.source>
    <maven.compiler.target>${java.version}</maven.compiler.target>
    ```
    
    or configuration section for maven-compiler-plugin.
    
    I wonder what exact error maven or javac give when there is mix of --target and --release options?

**Comment by khmarbaise** [comment]
There is also an explicit reference for JDK9+ https://maven.apache.org/plugins/maven-compiler-plugin/examples/set-compiler-release.html

**Comment by paulvi** [comment]
Maven docs however do not make it clear that there is big difference between `--source` `--target` vs `--release`
as now in Java it became common that APIs can appear and go away, e.g. Nashorn,
so it is imperative to compare against what APIs are actually avaiable for targeted release

https://stackoverflow.com/questions/43102787/what-is-the-release-flag-in-the-java-9-compiler

**Comment by paulvi** [comment]
I have tried Spring Boot 3.0 with added section like
```xml
			<plugin>
				<artifactId>maven-compiler-plugin</artifactId>
				<version>3.11.0</version>
				<configuration>
					<source>1.8</source>
					<target>1.8</target>
					<release>17</release>
				</configuration>
			</plugin>
```

In this case maven-compiler-plugin 3.11.0 just ignores source and target

```
[INFO] --- maven-compiler-plugin:3.11.0:compile (default-compile) @ ecbrates ---
[INFO] Changes detected - recompiling the module! :source
[INFO] Compiling 10 source files with javac [debug release 17] to target/classes
```

But when `<release>8</release>` the error message is misleading:  
`records are not supported in -source 8`

```
[ERROR] Failed to execute goal org.apache.maven.plugins:maven-compiler-plugin:3.11.0:compile (default-compile) on project ecbrates: Compilation failure
[ERROR] /Users/paulverest/Workspaces/Try/ecbrates/src/main/java/com/currencies/ecbrates/model/Quote.java:[7,8] records are not supported in -source 8
[ERROR]   (use -source 16 or higher to enable records)
```
That is tell about --source option not about --release
so of course configuration like
```xml
					<source>17</source>
					<target>17</target>
					<release>8</release>
```
will not make difference.

https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html page is also misleading
it says there is `Default value is: 1.8.` for 
[`<source>`](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#source)
But says nothing about default value for 
[`<release>`](https://maven.apache.org/plugins/maven-compiler-plugin/compile-mojo.html#release)
or that it will override `<source>` configuration.

---

**In short this PR will make `--release` work, though some developer will come through not so nice hints when running into errors.** But that is actually problem for maven-compiler-plugin to solve.

snicoll:
@arend-von-reinersdorff thank you for making your first contribution to Spring Boot.

PR #34763, opened 2023-03-25
Add Spring for Apache Pulsar

onobc:
Adds initial support for Spring for Apache Pulsar.

**Work remaining:**
- [x] docs

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-autoconfigure/build.gradle
@@ -179,6 +179,8 @@ dependencies {
 	optional("org.springframework.data:spring-data-redis")
 	optional("org.springframework.graphql:spring-graphql")
 	optional("org.springframework.hateoas:spring-hateoas")
+	optional("org.springframework.pulsar:spring-pulsar")
+	optional("org.springframework.pulsar:spring-pulsar-reactive")
 	optional("org.springframework.security:spring-security-acl")
 	optional("org.springframework.security:spring-security-config")
 	optional("org.springframework.security:spring-security-data") {
@@ -255,6 +257,7 @@ dependencies {
 	testImplementation("org.testcontainers:junit-jupiter")
 	testImplementation("org.testcontainers:mongodb")
 	testImplementation("org.testcontainers:neo4j")
+	testImplementation("org.testcontainers:pulsar")
 	testImplementation("org.testcontainers:testcontainers")
 	testImplementation("org.yaml:snakeyaml")
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
@@ -30,12 +30,11 @@ import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
-import org.springframework.boot.autoconfigure.pulsar.PulsarProperties.Producer.Cache;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.util.LambdaSafe;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.Import;
-import org.springframework.core.env.Environment;
 import org.springframework.pulsar.annotation.EnablePulsar;
 import org.springframework.pulsar.config.ConcurrentPulsarListenerContainerFactory;
 import org.springframework.pulsar.config.DefaultPulsarReaderContainerFactory;
@@ -66,7 +65,7 @@ import org.springframework.pulsar.reader.PulsarReaderContainerProperties;
  * @since 3.2.0
  */
 @AutoConfiguration
-@ConditionalOnClass({ PulsarClient.class, EnablePulsar.class })
+@ConditionalOnClass({ PulsarClient.class, PulsarTemplate.class })
 @Import(PulsarConfiguration.class)
 public class PulsarAutoConfiguration {
 
@@ -78,21 +77,31 @@ public class PulsarAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean(PulsarProducerFactory.class)
-	DefaultPulsarProducerFactory<Object> pulsarProducerFactory(Environment environment, PulsarClient pulsarClient,
-			TopicResolver topicResolver, ObjectProvider<ProducerBuilderCustomizer<?>> customizersProvider) {
-		String topicName = this.properties.getProducer().getTopicName();
+	@ConditionalOnProperty(name = "spring.pulsar.producer.cache.enabled", havingValue = "false")
+	DefaultPulsarProducerFactory<?> pulsarProducerFactory(PulsarClient pulsarClient, TopicResolver topicResolver,
+			ObjectProvider<ProducerBuilderCustomizer<?>> customizersProvider) {
+		return new DefaultPulsarProducerFactory<>(pulsarClient, this.properties.getProducer().getTopicName(),
+				lambdaSafeProducerBuilderCustomizers(customizersProvider), topicResolver);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(PulsarProducerFactory.class)
+	@ConditionalOnProperty(name = "spring.pulsar.producer.cache.enabled", havingValue = "true", matchIfMissing = true)
+	CachingPulsarProducerFactory<?> cachingPulsarProducerFactory(PulsarClient pulsarClient, TopicResolver topicResolver,
+			ObjectProvider<ProducerBuilderCustomizer<?>> customizersProvider) {
+		return new CachingPulsarProducerFactory<>(pulsarClient, this.properties.getProducer().getTopicName(),
+				lambdaSafeProducerBuilderCustomizers(customizersProvider), topicResolver,
+				this.properties.getProducer().getCache().getExpireAfterAccess(),
+				this.properties.getProducer().getCache().getMaximumSize(),
+				this.properties.getProducer().getCache().getInitialCapacity());
+	}
+
+	private List<ProducerBuilderCustomizer<Object>> lambdaSafeProducerBuilderCustomizers(
+			ObjectProvider<ProducerBuilderCustomizer<?>> customizersProvider) {
 		List<ProducerBuilderCustomizer<?>> customizers = new ArrayList<>();
 		customizers.add(PulsarPropertyMapper.producerBuilderCustomizer(this.properties));
 		customizers.addAll(customizersProvider.orderedStream().toList());
-		List<ProducerBuilderCustomizer<Object>> lambdaSafeCustomizers = List
-			.of((builder) -> applyProducerBuilderCustomizers(customizers, builder));
-		if (!environment.getProperty("spring.pulsar.producer.cache.enabled", Boolean.class, true)) {
-			return new DefaultPulsarProducerFactory<>(pulsarClient, topicName, lambdaSafeCustomizers, topicResolver);
-		}
-		Cache cacheProperties = this.properties.getProducer().getCache();
-		return new CachingPulsarProducerFactory<>(pulsarClient, topicName, lambdaSafeCustomizers, topicResolver,
-				cacheProperties.getExpireAfterAccess(), cacheProperties.getMaximumSize(),
-				cacheProperties.getInitialCapacity());
+		return List.of((builder) -> applyProducerBuilderCustomizers(customizers, builder));
 	}
 
 	@SuppressWarnings("unchecked")
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
@@ -1,0 +1,181 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.pulsar;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import org.apache.pulsar.client.api.ConsumerBuilder;
+import org.apache.pulsar.client.api.ProducerBuilder;
+import org.apache.pulsar.client.api.PulsarClient;
+import org.apache.pulsar.client.api.ReaderBuilder;
+import org.apache.pulsar.client.api.interceptor.ProducerInterceptor;
+
+import org.springframework.beans.factory.ObjectProvider;
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.autoconfigure.pulsar.PulsarProperties.Producer.Cache;
+import org.springframework.boot.util.LambdaSafe;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Import;
+import org.springframework.core.env.Environment;
+import org.springframework.pulsar.annotation.EnablePulsar;
+import org.springframework.pulsar.config.ConcurrentPulsarListenerContainerFactory;
+import org.springframework.pulsar.config.DefaultPulsarReaderContainerFactory;
+import org.springframework.pulsar.config.PulsarAnnotationSupportBeanNames;
+import org.springframework.pulsar.core.CachingPulsarProducerFactory;
+import org.springframework.pulsar.core.ConsumerBuilderCustomizer;
+import org.springframework.pulsar.core.DefaultPulsarConsumerFactory;
+import org.springframework.pulsar.core.DefaultPulsarProducerFactory;
+import org.springframework.pulsar.core.DefaultPulsarReaderFactory;
+import org.springframework.pulsar.core.ProducerBuilderCustomizer;
+import org.springframework.pulsar.core.PulsarConsumerFactory;
+import org.springframework.pulsar.core.PulsarProducerFactory;
+import org.springframework.pulsar.core.PulsarReaderFactory;
+import org.springframework.pulsar.core.PulsarTemplate;
+import org.springframework.pulsar.core.ReaderBuilderCustomizer;
+import org.springframework.pulsar.core.SchemaResolver;
+import org.springframework.pulsar.core.TopicResolver;
+import org.springframework.pulsar.listener.PulsarContainerProperties;
+import org.springframework.pulsar.reader.PulsarReaderContainerProperties;
+
+/**
+ * {@link EnableAutoConfiguration Auto-configuration} for Apache Pulsar.
+ *
+ * @author Chris Bono
+ * @author Soby Chacko
+ * @author Alexander Preuß
+ * @author Phillip Webb
+ * @since 3.2.0
+ */
+@AutoConfiguration
+@ConditionalOnClass({ PulsarClient.class, EnablePulsar.class })
+@Import(PulsarConfiguration.class)
+public class PulsarAutoConfiguration {
+
+	private PulsarProperties properties;
+
+	PulsarAutoConfiguration(PulsarProperties properties) {
+		this.properties = properties;
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(PulsarProducerFactory.class)
+	DefaultPulsarProducerFactory<Object> pulsarProducerFactory(Environment environment, PulsarClient pulsarClient,
+			TopicResolver topicResolver, ObjectProvider<ProducerBuilderCustomizer<?>> customizersProvider) {
+		String topicName = this.properties.getProducer().getTopicName();
+		List<ProducerBuilderCustomizer<?>> customizers = new ArrayList<>();
+		customizers.add(PulsarPropertyMapper.producerBuilderCustomizer(this.properties));
+		customizers.addAll(customizersProvider.orderedStream().toList());
+		List<ProducerBuilderCustomizer<Object>> lambdaSafeCustomizers = List
+			.of((builder) -> applyProducerBuilderCustomizers(customizers, builder));
+		if (!environment.getProperty("spring.pulsar.producer.cache.enabled", Boolean.class, true)) {
+			return new DefaultPulsarProducerFactory<>(pulsarClient, topicName, lambdaSafeCustomizers, topicResolver);
+		}
+		Cache cacheProperties = this.properties.getProducer().getCache();
+		return new CachingPulsarProducerFactory<>(pulsarClient, topicName, lambdaSafeCustomizers, topicResolver,
+				cacheProperties.getExpireAfterAccess(), cacheProperties.getMaximumSize(),
+				cacheProperties.getInitialCapacity());
+	}
+
+	@SuppressWarnings("unchecked")
+	private void applyProducerBuilderCustomizers(List<ProducerBuilderCustomizer<?>> customizers,
+			ProducerBuilder<?> builder) {
+		LambdaSafe.callbacks(ProducerBuilderCustomizer.class, customizers, builder)
+			.invoke((customizer) -> customizer.customize(builder));
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	PulsarTemplate<?> pulsarTemplate(PulsarProducerFactory<?> pulsarProducerFactory,
+			ObjectProvider<ProducerInterceptor> producerInterceptors, SchemaResolver schemaResolver,
+			TopicResolver topicResolver) {
+		return new PulsarTemplate<>(pulsarProducerFactory, producerInterceptors.orderedStream().toList(),
+				schemaResolver, topicResolver, this.properties.getTemplate().isObservationsEnabled());
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(PulsarConsumerFactory.class)
+	DefaultPulsarConsumerFactory<Object> pulsarConsumerFactory(PulsarClient pulsarClient,
+			ObjectProvider<ConsumerBuilderCustomizer<?>> customizersProvider) {
+		List<ConsumerBuilderCustomizer<?>> customizers = new ArrayList<>();
+		customizers.add(PulsarPropertyMapper.consumerBuilderCustomizer(this.properties));
+		customizers.addAll(customizersProvider.orderedStream().toList());
+		List<ConsumerBuilderCustomizer<Object>> lambdaSafeCustomizers = List
+			.of((builder) -> applyConsumerBuilderCustomizers(customizers, builder));
+		return new DefaultPulsarConsumerFactory<>(pulsarClient, lambdaSafeCustomizers);
+	}
+
+	@SuppressWarnings("unchecked")
+	private void applyConsumerBuilderCustomizers(List<ConsumerBuilderCustomizer<?>> customizers,
+			ConsumerBuilder<?> builder) {
+		LambdaSafe.callbacks(ConsumerBuilderCustomizer.class, customizers, builder)
+			.invoke((customizer) -> customizer.customize(builder));
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(name = "pulsarListenerContainerFactory")
+	ConcurrentPulsarListenerContainerFactory<Object> pulsarListenerContainerFactory(
+			PulsarConsumerFactory<Object> pulsarConsumerFactory, SchemaResolver schemaResolver,
+			TopicResolver topicResolver) {
+		PulsarContainerProperties containerProperties = new PulsarContainerProperties();
+		containerProperties.setSchemaResolver(schemaResolver);
+		containerProperties.setTopicResolver(topicResolver);
+		PulsarPropertyMapper.containerPropertiesCustomizer(this.properties).accept(containerProperties);
+		return new ConcurrentPulsarListenerContainerFactory<>(pulsarConsumerFactory, containerProperties);
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(PulsarReaderFactory.class)
+	DefaultPulsarReaderFactory<?> pulsarReaderFactory(PulsarClient pulsarClient,
+			ObjectProvider<ReaderBuilderCustomizer<?>> customizersProvider) {
+		List<ReaderBuilderCustomizer<?>> customizers = new ArrayList<>();
+		customizers.add(PulsarPropertyMapper.readerBuilderCustomizer(this.properties));
+		customizers.addAll(customizersProvider.orderedStream().toList());
+		List<ReaderBuilderCustomizer<Object>> lambdaSafeCustomizers = List
+			.of((builder) -> applyReaderBuilderCustomizers(customizers, builder));
+		return new DefaultPulsarReaderFactory<>(pulsarClient, lambdaSafeCustomizers);
+	}
+
+	@SuppressWarnings("unchecked")
+	private void applyReaderBuilderCustomizers(List<ReaderBuilderCustomizer<?>> customizers, ReaderBuilder<?> builder) {
+		LambdaSafe.callbacks(ReaderBuilderCustomizer.class, customizers, builder)
+			.invoke((customizer) -> customizer.customize(builder));
+	}
+
+	@Bean
+	@ConditionalOnMissingBean(name = "pulsarReaderContainerFactory")
+	DefaultPulsarReaderContainerFactory<?> pulsarReaderContainerFactory(PulsarReaderFactory<?> pulsarReaderFactory,
+			SchemaResolver schemaResolver) {
+		PulsarReaderContainerProperties readerContainerProperties = new PulsarReaderContainerProperties();
+		readerContainerProperties.setSchemaResolver(schemaResolver);
+		PulsarPropertyMapper.readerContainerPropertiesCustomizer(this.properties).accept(readerContainerProperties);
+		return new DefaultPulsarReaderContainerFactory<>(pulsarReaderFactory, readerContainerProperties);
+	}
+
+	@Configuration(proxyBeanMethods = false)
+	@EnablePulsar
+	@ConditionalOnMissingBean(name = { PulsarAnnotationSupportBeanNames.PULSAR_LISTENER_ANNOTATION_PROCESSOR_BEAN_NAME,
+			PulsarAnnotationSupportBeanNames.PULSAR_READER_ANNOTATION_PROCESSOR_BEAN_NAME })
+	static class EnablePulsarConfiguration {
+
+	}
+
+}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
@@ -1,0 +1,883 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.pulsar;
+
+import java.time.Duration;
+import java.util.ArrayList;
+import java.util.LinkedHashMap;
+import java.util.List;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+import org.apache.pulsar.client.api.CompressionType;
+import org.apache.pulsar.client.api.HashingScheme;
+import org.apache.pulsar.client.api.MessageRoutingMode;
+import org.apache.pulsar.client.api.ProducerAccessMode;
+import org.apache.pulsar.client.api.RegexSubscriptionMode;
+import org.apache.pulsar.client.api.SubscriptionInitialPosition;
+import org.apache.pulsar.client.api.SubscriptionMode;
+import org.apache.pulsar.client.api.SubscriptionType;
+import org.apache.pulsar.common.schema.SchemaType;
+
+import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.boot.context.properties.NestedConfigurationProperty;
+import org.springframework.util.Assert;
+
+/**
+ * Configuration properties Apache Pulsar.
+ *
+ * @author Chris Bono
+ * @author Phillip Webb
+ * @since 3.2.0
+ */
+@ConfigurationProperties("spring.pulsar")
+public class PulsarProperties {
+
+	private final Client client = new Client();
+
+	private final Admin admin = new Admin();
+
+	private final Defaults defaults = new Defaults();
+
+	private final Function function = new Function();
+
+	private final Producer producer = new Producer();
+
+	private final Consumer consumer = new Consumer();
+
+	private final Listener listener = new Listener();
+
+	private final Reader reader = new Reader();
+
+	private final Template template = new Template();
+
+	public Client getClient() {
+		return this.client;
+	}
+
+	public Admin getAdmin() {
+		return this.admin;
+	}
+
+	public Defaults getDefaults() {
+		return this.defaults;
+	}
+
+	public Producer getProducer() {
+		return this.producer;
+	}
+
+	public Consumer getConsumer() {
+		return this.consumer;
+	}
+
+	public Listener getListener() {
+		return this.listener;
+	}
+
+	public Reader getReader() {
+		return this.reader;
+	}
+
+	public Function getFunction() {
+		return this.function;
+	}
+
+	public Template getTemplate() {
+		return this.template;
+	}
+
+	public static class Client {
+
+		/**
+		 * Pulsar service URL in the format '(pulsar|pulsar+ssl)://host:port'.
+		 */
+		private String serviceUrl = "pulsar://localhost:6650";
+
+		/**
+		 * Client operation timeout.
+		 */
+		private Duration operationTimeout = Duration.ofSeconds(30);
+
+		/**
+		 * Client lookup timeout.
+		 */
+		private Duration lookupTimeout = Duration.ofMillis(-1); // FIXME
+
+		/**
+		 * Duration to wait for a connection to a broker to be established.
+		 */
+		private Duration connectionTimeout = Duration.ofSeconds(10);
+
+		/**
+		 * Authentication settings.
+		 */
+		private final Authentication authentication = new Authentication();
+
+		public String getServiceUrl() {
+			return this.serviceUrl;
+		}
+
+		public void setServiceUrl(String serviceUrl) {
+			this.serviceUrl = serviceUrl;
+		}
+
+		public Duration getOperationTimeout() {
+			return this.operationTimeout;
+		}
+
+		public void setOperationTimeout(Duration operationTimeout) {
+			this.operationTimeout = operationTimeout;
+		}
+
+		public Duration getLookupTimeout() {
+			return this.lookupTimeout;
+		}
+
+		public void setLookupTimeout(Duration lookupTimeout) {
+			this.lookupTimeout = lookupTimeout;
+		}
+
+		public Duration getConnectionTimeout() {
+			return this.connectionTimeout;
+		}
+
+		public void setConnectionTimeout(Duration connectionTimeout) {
+			this.connectionTimeout = connectionTimeout;
+		}
+
+		public Authentication getAuthentication() {
+			return this.authentication;
+		}
+
+	}
+
+	public static class Admin {
+
+		/**
+		 * Pulsar web URL for the admin endpoint in the format '(http|https)://host:port'.
+		 */
+		private String serviceUrl = "http://localhost:8080";
+
+		/**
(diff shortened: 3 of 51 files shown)


buildSrc/src/main/java/org/springframework/boot/build/classpath/CheckClasspathForProhibitedDependencies.java
```diff
@@ -79,6 +79,9 @@ private boolean prohibited(ModuleVersionIdentifier id) {
 		if (group.equals("javax.money")) {
 			return false;
 		}
+		if (group.equals("javax.xml.bind") && id.getName().equals("jaxb-api")) {
```
onobc, line 82: The Apache Pulsar client lib unfortunately brings this unfortunate baggage w/ it. The Pulsar admin client relies on this as it still codes to the old `javax.*` API. 

I have raised this https://github.com/apache/pulsar/issues/19856 w/ the Pulsar team and it will be address in the `3.0.0` release (they are going straight from `2.11.0 -> 3.0.0` so its not too far out on timeline but I'm not 100% certain the expected date). 

Would it be possible to augment the plugin to allow this dep. only in Pulsar (until 3.0.0 is out)?

spring-boot-project/spring-boot-autoconfigure/build.gradle
```diff
@@ -58,6 +58,11 @@ dependencies {
 	}
 	optional("org.apache.httpcomponents.client5:httpclient5")
 	optional("org.apache.kafka:kafka-streams")
+	optional("org.apache.pulsar:pulsar-client-reactive-producer-cache-caffeine") {
+		// spring-pulsar-reactive includes the pulsar-client-reactive-adapter with
```
onobc, line 62: The Reactive Pulsar client (`pulsar-client-reactive-adapter`) currently has a couple of prohibited deps that we filter out in Spring Pulsar (`spring-pulsar-reactive`). However, the `pulsar-client-reactive-producer-cache-caffeine` has a dep. on the `pulsar-client-reactive-adapter`. 

I am currently looking into splitting this out a bit in the client to avoid these annoying exclusions.

spring-boot-project/spring-boot-starters/spring-boot-starter-pulsar-reactive/build.gradle
```diff
@@ -0,0 +1,16 @@
...
+}
+
+description = "Starter for using Spring for Apache Pulsar Reactive"
+
+dependencies {
+	api(project(":spring-boot-project:spring-boot-starters:spring-boot-starter"))
+	api("io.projectreactor:reactor-core")
+	api("org.springframework.pulsar:spring-pulsar-reactive")
+	api("org.apache.pulsar:pulsar-client-reactive-producer-cache-caffeine") {
+		// spring-pulsar-reactive includes the pulsar-client-reactive-adapter with
```
onobc, line 12: Ditto https://github.com/spring-projects/spring-boot/pull/34763/files#r1148396121

spring-boot-project/spring-boot-starters/spring-boot-starter-pulsar/build.gradle
```diff
@@ -0,0 +1,11 @@
+plugins {
+	id "org.springframework.boot.starter"
+}
+
+description = "Starter for using Spring for Apache Pulsar"
+
+dependencies {
+	api(project(":spring-boot-project:spring-boot-starters:spring-boot-starter"))
+	api("com.github.ben-manes.caffeine:caffeine")
```
onobc, line 9: There is WIP to shade the fact that the cache mechanism is using caffeine. It is an internal detail and not intended to swapped out.

spring-boot-project/spring-boot-autoconfigure/build.gradle
```diff
@@ -58,6 +58,11 @@ dependencies {
 	}
 	optional("org.apache.httpcomponents.client5:httpclient5")
 	optional("org.apache.kafka:kafka-streams")
+	optional("org.apache.pulsar:pulsar-client-reactive-producer-cache-caffeine") {
```
onobc, line 61: This transitively brings in caffeine. There is WIP to shade the fact that the cache mechanism is using caffeine. It is an internal detail and not intended to swapped out.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
```diff
@@ -0,0 +1,178 @@
...
+				this.properties.getProducer().getCache().getMaximumSize(),
+				this.properties.getProducer().getCache().getInitialCapacity());
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public PulsarTemplate<?> pulsarTemplate(PulsarProducerFactory<?> pulsarProducerFactory,
+			ObjectProvider<ProducerInterceptor> interceptorsProvider, SchemaResolver schemaResolver,
+			TopicResolver topicResolver,
+			ObjectProvider<io.micrometer.observation.ObservationRegistry> observationRegistryProvider,
```
onobc, line 106: You may notice that we are "sneaking in" the reference to `ObservationRegistry`. I do understand why it is a prohibited import as it belongs under actuator auto-configure. However, the Pulsar template/listener use the registry directly (if provided) to record observations. 

Spring Kafka also uses the registry directly, but rather than constructor injection, it does a bean lookup in [afterSingletonsInstantiated()](https://github.com/spring-projects/spring-kafka/blob/4eacaa9dce252357c7ef4a7632430dc362a3fd8f/spring-kafka/src/main/java/org/springframework/kafka/core/KafkaTemplate.java#L446).  We will change Spring Pulsar to do the same thing shortly. Long term I want to explore instead using a filter or some other mechanism so that it does not have a direct dependency on the registry.

spring-boot-project/spring-boot-starters/spring-boot-starter-pulsar/build.gradle
onobc: This is done now.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
onobc: This has been fixed (no more observation registry in the auto-configuration.

buildSrc/src/main/java/org/springframework/boot/build/classpath/CheckClasspathForProhibitedDependencies.java
wilkinsona: I don't think we can do that. Even in 2.x we prohibit dependencies from a `javax.*` group in favor of the more modern `jakarta.*` equivalents that use the `javax` namespace for their packages. Even if we tolerated the `javax.xml.bind` API dependencies, I think we'd quickly hit problems with the runtime which uses the same coordinates (`org.glassfish.jaxb:jaxb-runtime`) for its `javax` and `jakarta` implementations.

onobc: > I think we'd quickly hit problems with the runtime which uses the same coordinates (org.glassfish.jaxb:jaxb-runtime) for its javax and jakarta implementations.

I don't see `org.glassfish.jaxb:jaxb-runtime` using any `javax` classes:
* https://mvnrepository.com/artifact/org.glassfish.jaxb/jaxb-bom/4.0.2
  * https://mvnrepository.com/artifact/org.glassfish.jaxb/jaxb-runtime/4.0.2
    * https://mvnrepository.com/artifact/org.glassfish.jaxb/jaxb-core/4.0.2

I don't think that will happen as the classes are "isolated" from one another by package name (`javax` / `jakarta`). In effect, the only thing still looking at the `javax` would be the older Pulsar client admin.

onobc: > the javax classes and the runtime for the jakarta classes on the classpath at the same time due to them have the same Maven coordinates.

The runtime has the same coordinates and only a version difference of that coordinates dictates which version it uses (javax/jakarta), is that correct? 

> The API without a runtime to go with it is of limited or perhaps even no use and I don't want users to have to try to straighten out that mess.

Agreed - that would not be good. 
Although, for the Pulsar client, it does shade the older runtime that it uses (it unfortunately did not shade the API). 
So users would not need to enable the older runtime - I think it would be invisible to them. 

I will setup an example app and see what happens.

wilkinsona: > The runtime has the same coordinates and only a version difference of that coordinates dictates which version it uses (javax/jakarta), is that correct?

Correct. 4.0.x is `jakarta.*`.

spring-boot-project/spring-boot-autoconfigure/build.gradle
onobc: https://github.com/apache/pulsar-client-reactive/pull/124

buildSrc/src/main/java/org/springframework/boot/build/classpath/CheckClasspathForProhibitedDependencies.java
onobc: Good news, the updates to the Pulsar client were merged and will make it into the `3.0.0-RC2` release. I will update the core to use that version and we should be rid of this violation.

onobc: We are now using Pulsar `3.0.0` and this is no longer an issue.

onobc:
I have also added support for the following:

- Pulsar connection details in auto-config
- Pulsar service connection from testcontainers
- Pulsar docker compose support + service connections from docker compose

These changes are in my fork (based off this same PR branch) in 3 separate commits, queued up and ready to PR once this base PR is merged:
- https://github.com/onobc/spring-boot/tree/cbono-add-pulsar-service-connection

vpavic:
Is this expected to target Spring Boot 3.2 as hinted in [this blog post](https://spring.io/blog/2023/03/24/spring-for-apache-pulsar-0-2-0-available-now)? I'm asking because this is still in 3.x milestone and work on 3.2 is well underway.

spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -1107,6 +1107,24 @@ bom {
 			]
 		}
 	}
+	library("Pulsar", "3.0.0") {
+		group("org.apache.pulsar") {
+			modules = [
+					"pulsar-client",
+					"pulsar-client-admin",
+					"pulsar-client-all"
```
wilkinsona, line 1115: I think this list of modules is incomplete. It should include every module that's released as a single "package".

spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -1107,6 +1107,24 @@ bom {
...
+			]
+		}
+	}
+	library("Pulsar Reactive", "0.3.0") {
+		group("org.apache.pulsar") {
+			modules = [
+					"pulsar-client-reactive-adapter",
+					"pulsar-client-reactive-producer-cache-caffeine",
+					"pulsar-client-reactive-producer-cache-caffeine-shaded"
```
wilkinsona, line 1124: I think this list of modules is incomplete. It should include every module that's released as a single "package".

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging.adoc
```diff
@@ -15,15 +15,12 @@ include::messaging/amqp.adoc[]
...
 include::messaging/spring-integration.adoc[]
 
 include::messaging/websockets.adoc[]
 
 include::messaging/whats-next.adoc[]
-
-
-
-
-
```
wilkinsona, line 29: This change seems unrelated and may adversely affect the docs' formatting.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
```diff
@@ -0,0 +1,144 @@
+[[messaging.pulsar]]
+== Apache Pulsar Support
+https://pulsar.apache.org/[Apache Pulsar] is supported by providing auto-configuration of the {spring-pulsar-docs}[Spring for Apache Pulsar] project.
+
+Spring Boot will auto-configure and register the classic (imperative) Spring Pulsar components when `org.springframework.pulsar:spring-pulsar` is on the classpath.
+It will do the same for the reactive components when both `org.springframework.pulsar:spring-pulsar-reactive` and Project Reactor are on the classpath.
```
wilkinsona, line 6: Doesn't `spring-pulsar-reactive` pull in Reactor? If so, I don't think there's any need to mention Reactor here.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
```diff
@@ -0,0 +1,144 @@
...
+
+There are `spring-boot-starter-pulsar` and `spring-boot-starter-pulsar-reactive` "`Starters`" for conveniently collecting the dependencies for imperative and reactive use, respectively.
+
+TIP: The Reactive support is built on top of the imperative support and therefore the reactive starter includes the imperative starter.
+As such, the imperative components are auto-configured and available in both reactive and imperative scenarios.
```
wilkinsona, line 11: What happens if you use the imperative components in a reactive app? Will the blocking cause a problem?

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
```diff
@@ -0,0 +1,144 @@
...
+[source,yaml,indent=0,subs="verbatim",role="secondary"]
+----
+spring:
+  pulsar:
+    client:
+      auth-plugin-class-name: org.apache.pulsar.client.impl.auth.oauth2.AuthenticationOAuth2
+      auth-params: "{\"privateKey\":\"file:///Users/some-key.json\",\"issuerUrl\":\"https://auth.server.cloud/", \"audience\":\"urn:sn:acme:dev:my-instance"}"
+----
+
+TIP: Using a map is the recommended approach as it is less error-prone and easier to read.
```
wilkinsona, line 51: I'd just document the map-based approach.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
```diff
@@ -0,0 +1,144 @@
...
+
+The `ReactivePulsarClient` adapts an instance of the previously described `PulsarClient`.
+Therefore, use the properties described previously (`spring.pulsar.client.*`) to configure the `ReactivePulsarClient`.
+
+[[messaging.pulsar.admin]]
+=== Connecting to Pulsar Administration
+Spring Pulsar's `PulsarAdministration` client is also auto-configured.
+
+By default, the application tries to connect to a local Pulsar instance at `\http://localhost:8080`.
+This can be adjusted by setting the `spring.pulsar.administration.service-url` property to a different value in the form `(http|https)://<host>:<port>`.
```
wilkinsona, line 67: Please use `configprop:spring.pulsar.administration.service-url[]` as this ensures that the property exists.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
```diff
@@ -0,0 +1,144 @@
...
+
+TIP: The Reactive support is built on top of the imperative support and therefore the reactive starter includes the imperative starter.
+As such, the imperative components are auto-configured and available in both reactive and imperative scenarios.
+
+[[messaging.pulsar.connecting]]
+=== Connecting to Pulsar
+When you use the Pulsar starter, Spring Boot will auto-configure and register a `PulsarClient` bean.
+
+By default, the application tries to connect to a local Pulsar instance at `pulsar://localhost:6650`.
+This can be adjusted by setting the `spring.pulsar.client.service-url` property to a different value.
```
wilkinsona, line 18: Please use `configprop:spring.pulsar.client.service-url[]` as this ensures that the property exists.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
```diff
@@ -0,0 +1,144 @@
...
+When the Apache Pulsar infrastructure is present, any bean can be annotated with `@PulsarListener` to create a listener endpoint.
+The following component creates a listener endpoint on the `someTopic` topic:
+
+include::code:MyBean[]
+
+Spring Boot auto-configuration provides all the components necessary for `PulsarListener`, such as the `PulsarListenerContainerFactory` and the consumer factory it uses to construct the underlying Pulsar consumers.
+You can configure these components by specifying any of the `spring.pulsar.listener.\*` and `spring.pulsar.consumer.*` prefixed application properties.
+
+[[messaging.pulsar.receiving-reactive]]
+=== Receiving a Message (Reactively)
```
wilkinsona, line 109: Remove the parentheses?

spring-boot-project/spring-boot-starters/spring-boot-starter-pulsar-reactive/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+}
+
+checkRuntimeClasspathForConflicts {
+	// pulsar bouncycastle ext includes duplicate bouncycastle classes
+	ignore { name -> name.startsWith("org/bouncycastle/") }
+
+	// spring-pulsar also shades/relocates caffeine to the same location
+	ignore { name -> name.startsWith("org/springframework/pulsar/shade/com/github/benmanes/caffeine/") }
+	ignore { name -> name.startsWith("org/springframework/pulsar/shade/com/google/errorprone/") }
+	ignore { name -> name.startsWith("org/springframework/pulsar/shade/org/checkerframework/") }
```
wilkinsona, line 20: Are these still needed? The latest snapshots don't seem to have any shaded classes.

spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-pulsar-reactive/build.gradle
```diff
@@ -0,0 +1,15 @@
+plugins {
+	id "java"
+	id "org.springframework.boot.conventions"
+}
+
+description = "Spring Boot Pulsar smoke test"
```
wilkinsona, line 6: Add "Reactive" to the description?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAnnotationDrivenConfiguration.java
```diff
@@ -0,0 +1,105 @@
...
+/**
+ * Configuration for Pulsar annotation-driven support.
+ *
+ * @author Soby Chacko
+ * @author Chris Bono
+ * @since 3.2
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnClass(EnablePulsar.class)
+public class PulsarAnnotationDrivenConfiguration {
```
wilkinsona, line 48: Package-private

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAnnotationDrivenConfiguration.java
```diff
@@ -0,0 +1,105 @@
...
+
+	private final PulsarProperties pulsarProperties;
+
+	public PulsarAnnotationDrivenConfiguration(PulsarProperties pulsarProperties) {
```
wilkinsona, line 52: Package-private

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/ConsumerConfigProperties.java
```diff
@@ -0,0 +1,540 @@
...
+import org.springframework.boot.context.properties.PropertyMapper;
+import org.springframework.pulsar.core.ConsumerBuilderCustomizer;
+
+/**
+ * Configuration properties used to specify Pulsar consumers.
+ *
+ * @author Chris Bono
+ * @since 3.2
+ */
+public class ConsumerConfigProperties {
```
wilkinsona, line 47: Could this be nested within `PulsarProperties`?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/ProducerConfigProperties.java
```diff
@@ -0,0 +1,406 @@
...
+import org.springframework.pulsar.core.ProducerBuilderCustomizer;
+import org.springframework.util.unit.DataSize;
+
+/**
+ * Configuration properties used to specify Pulsar producers.
+ *
+ * @author Chris Bono
+ * @since 3.2
+ */
+public class ProducerConfigProperties {
```
wilkinsona, line 42: Could this be nested within `PulsarProperties`?

spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -2133,6 +2133,64 @@
...
+      "type": "java.lang.Integer",
+      "description": "Maximum number of times that a message will be redelivered before being sent to the dead letter queue."
+    },
+    {
+      "name": "spring.pulsar.reactive.consumer.dead-letter-policy.retry-letter-topic",
+      "type": "java.lang.String",
+      "description": "Name of the retry topic where the failing messages will be sent."
+    },
+    {
+      "name": "spring.pulsar.reactive.sender.cache.enabled",
```
wilkinsona, line 2189: Where does this come from? The `Sender` class doesn't have a `Cache`.

wilkinsona (changes requested):
This for the PR, @onobc. In addition to the various individual comments that I've made for you consideration, I have the following more general comments that apply in multiple places:

- Please add Kotlin versions of code examples for the docs
- Copyright should be 2012-2023
- All `@since` versions should be 3.2.0
- Some configuration properties could use primitive types (`Integer` -> `int`, `Boolean` -> `boolean`)
- The property classes seem very flat. I think it would be good to group things together more. The four `subscription*` properties in `ConsumerConfigProperties` are one example of this.
- `@Bean` methods and constructors on auto-configuration should be package-private
- Please don't use javadoc `@link` and the like in config property descriptions
- We try to avoid binding properties directly to third-party types such as `org.apache.pulsar.client.api.DeadLetterPolicy`
- Enum properties with default values require additional metadata for those defaults as the annotation processor can't extract them automatically

onobc:
Thanks for the detailed review @wilkinsona . From a quick read, the requested changes seem light in nature and all make sense. I will get this updated over the next couple of days.

spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
onobc: It is hiding in the plethora of other properties ([here](https://github.com/spring-projects/spring-boot/pull/34763/files#diff-9ae54aecfff5dde08e706cda1fda7a45880dcee70f9e3f07f8ab460026d452f8R413) and [defined here](https://github.com/spring-projects/spring-boot/pull/34763/files#diff-9ae54aecfff5dde08e706cda1fda7a45880dcee70f9e3f07f8ab460026d452f8R226)).

Is it best practice to declare the nested props at the top of the containing class? I think so, and by moving the 
```java
private final Cache cache = new Cache();
```
to the top of `Sender` class would make this more readable. Wdyt?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/ConsumerConfigProperties.java
onobc: No, it is currently used by the SCSt Pulsar Binder config properties as well https://github.com/spring-cloud/spring-cloud-stream/blob/25af6a76b04598eb31bf2f84e59f45cee742b8ea/binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/main/java/org/springframework/cloud/stream/binder/pulsar/properties/PulsarBinderConfigurationProperties.java#L33-L46

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/ProducerConfigProperties.java
onobc: No, it is currently used by the SCSt Pulsar Binder config properties as well https://github.com/spring-cloud/spring-cloud-stream/blob/25af6a76b04598eb31bf2f84e59f45cee742b8ea/binders/pulsar-binder/spring-cloud-stream-binder-pulsar/src/main/java/org/springframework/cloud/stream/binder/pulsar/properties/PulsarBinderConfigurationProperties.java#L33-L46

spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -1107,6 +1107,96 @@ bom {
 			]
 		}
 	}
+	library("Pulsar", "3.0.0") {
```
onobc, line 1110: Pulsar does not do a good job of dependency hygiene wrt what is server, client, shared and there a TON of published modules. Also, there is not one single arbiter in the repo that says "Here is what I publish" - its scattered about the repo in several different places and I am not clear which is which.

I instead got the list using https://github.com/mthmulders/mcs.

```
mcs search org.apache.pulsar -l 200 | grep ":3.0.0" | sed 's/3.0.0.*/3.0.0/' | sort | pbcopy
```
and
```
mcs search org.apache.pulsar -l 200 | grep ":0.3.0" | sed 's/0.3.0.*/0.3.0/' | pbcopy
```

onobc:
@wilkinsona I believe I have addressed all of your points except `Please add Kotlin versions of code examples for the docs`. Would you be ok w/ me adding that in a subsequent smaller PR?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/ConsumerConfigProperties.java
wilkinsona: It shouldn't be doing that. From [the docs](https://docs.spring.io/spring-boot/docs/3.1.2/reference/htmlsingle/#features.external-config.typesafe-configuration-properties.java-bean-binding):

> The properties that map to `@ConfigurationProperties` classes available in Spring Boot, which are configured through properties files, YAML files, environment variables, and other mechanisms, are public API but the accessors (getters/setters) of the class itself are not meant to be used directly.

Please nest it and make the appropriate changes in SCSt Pulsar Binder.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/ProducerConfigProperties.java
wilkinsona: As above. Please nest it and make the appropriate changes in SCSt Pulsar Binder..

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/ConsumerConfigProperties.java
onobc: When the auto-config, binder, and core were in a single repo/project this was not an issue but I figured we would need to decouple it. I will do that and nest as you suggest.

spring-boot-project/spring-boot-starters/spring-boot-starter-pulsar-reactive/build.gradle
onobc: We do use the shaded version of the reactive cache https://github.com/spring-projects/spring-pulsar/blob/7f5ced49dc012164037a872933da5b954776437d/spring-pulsar-reactive/spring-pulsar-reactive.gradle#L13

Once I fix this https://github.com/apache/pulsar-client-reactive/issues/127 then we can remove these prefixes and they will not be considered duplicates.

spring-boot-project/spring-boot-dependencies/build.gradle
onobc: Yes, when you have to write a script to figure it out, its quite a lot of modules. I will file an issue and reach out to some committers but I am doubtful we will see any immediate changes in the SB 3.2 timeframe. I will probably end up submitting a PR for it (that would be faster - but have to find the time to do it).

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
```diff
@@ -0,0 +1,132 @@
...
+https://pulsar.apache.org/[Apache Pulsar] is supported by providing auto-configuration of the {spring-pulsar-docs}[Spring for Apache Pulsar] project.
+
+Spring Boot will auto-configure and register the classic (imperative) Spring Pulsar components when `org.springframework.pulsar:spring-pulsar` is on the classpath.
+It will do the same for the reactive components when `org.springframework.pulsar:spring-pulsar-reactive` is on the classpath.
+
+There are `spring-boot-starter-pulsar` and `spring-boot-starter-pulsar-reactive` "`Starters`" for conveniently collecting the dependencies for imperative and reactive use, respectively.
+
+TIP: The Reactive auto-configuration relies on the imperative auto-configuration and therefore the reactive starter includes the imperative starter.
+As such, the imperative components are auto-configured and available in both reactive and imperative scenarios.
+However, using the imperative components in a fully reactive app is not recommended as it could cause blocking issues.
```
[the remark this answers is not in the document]
onobc: Not as it is currently written. It would be quite the effort to do this. The reactive components leverage the reactive client which adapts the original client (which is imperative and auto-configured). I believe there are other technologies auto-configured in Boot whose reactive components leverage the imperative ones, no?

spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-pulsar-reactive/src/test/resources/logback-test.xml
```diff
@@ -0,0 +1,13 @@
...
+        <encoder>
+            <pattern>%d{HH:mm:ss.SSS} [%thread] %-5level %logger - %msg%n</pattern>
+        </encoder>
+    </appender>
+    <root level="WARN">
+        <appender-ref ref="STDOUT"/>
+    </root>
+    <logger name="org.testcontainers" level="WARN"/>
+    <logger name="com.github.dockerjava" level="WARN"/>
+	<logger name="org.springframework.pulsar" level="INFO"/>
```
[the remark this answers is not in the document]
onobc: Good catch - we can remove this now in that case.

onobc:
> Thanks for the latest updates, @onobc.

You're welcome @wilkinsona. I will make the updates requested shortly.  

> In addition to my specific comments, I've also spent some time reviewing the configuration properties again. Unfortunately, I'm quite concerned about the sheer number of them.
> 
> Looking at the generated JSON, there are 219 `spring.pulsar` properties. To put that into context, there are 1770 properties in total in `spring-boot-autoconfigure` with 12% of those being for Spring Pulsar. Put another way, adding Spring Pulsar support increases the properties in `spring-boot-autoconfigure` by 14%.
> 
> Maybe it's just the nature of the beast (there are 174 `spring.kafka.*` properties) but I do think it would be worth spending some time to see if the configuration model can be simplified. I think that would be of benefit both to us from a maintenance perspective and to users from an ease-of-use perspective.

I agree, there are many properties. I think it is the nature of the beast though. 
I also agree simplifying would be beneficial. However, I have attempted to do this but do not see a clear-cut path forward.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/AuthParameterUtils.java
```diff
@@ -0,0 +1,75 @@
...
+					Map.Entry::getValue));
+	}
+
+	private static Map<String, String> convertKebabCaseKeysToCamelCase(Map<String, String> params) {
+		return params.entrySet()
+			.stream()
+			.collect(Collectors.toMap((entry) -> convertKebabCaseToCamelCase(entry.getKey()), Map.Entry::getValue));
+	}
+
+	static String maybeConvertToEncodedParamString(Map<String, String> params) {
```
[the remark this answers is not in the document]
onobc: This simplified things nicely.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarClientBuilderConfigurer.java
```diff
@@ -0,0 +1,183 @@
...
+			.as(this::parseSocketAddress)
+			.to(clientBuilder::socks5ProxyAddress);
+		map.from(clientProperties::getSocks5ProxyUsername).to(clientBuilder::socks5ProxyUsername);
+		map.from(clientProperties::getSocks5ProxyPassword).to(clientBuilder::socks5ProxyPassword);
+	}
+
+	private void applyAuthentication(Client clientProperties, ClientBuilder clientBuilder) {
+		if (StringUtils.hasText(clientProperties.getAuthParams())
+				&& !CollectionUtils.isEmpty(clientProperties.getAuthentication())) {
+			throw new IllegalArgumentException(
```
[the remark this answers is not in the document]
onobc: TIL

onobc:
@wilkinsona I have covered all of the recent concerns other than:

- too many modules (this will require change in Pulsar)
- too many config props (I don't see a way to reduce this currently)
- don't expose imperative components in reactive app (big effort to do this)
- Add Kotlin examples to docs (would like to do in subsequent PR)

Thanks

philwebb:
I find the number of configuration properties quite a concern too. I wonder if we can just add the properties that we think will be popular and lean on `PulsarClientBuilderCustomizer` for the more exotic ones?

One thing we've learnt from previous features is it's easier to add properties later when they are needed than take them away because they aren't.

wilkinsona:
We discussed the properties in a team meeting yesterday (without Phil, he was on PTO) and came to the same conclusion. @snicoll reminded my that our Kafka support has got to where it is having evolved over many releases. This makes its large number of properties somewhat easier to justify. Our feeling is that we shouldn't jump straight into such a position with Pulsar and should instead provide plenty of room for things to evolve based on community feedback.

As Phil suggests, the room to evolve can be provided by using one or more customizer callbacks to configure things not covered by configuration properties. It then becomes a question of what should be a configuration property. We'd defer to you on that as we lack the Pulsar expertise to have an informed opinion. As a general rule of thumb, settings that change from environment to environment (dev, staging, prod and so on) are the strongest candidates for configuration properties. Next would be settings that are really commonly tuned. Ideally we'd see a 5-10x reduction in the number of properties that we offer to begin with.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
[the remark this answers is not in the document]
wilkinsona: > The reactive components leverage the reactive client which adapts the original client (which is imperative and auto-configured)

I wonder if the original, imperative client has to be a bean in the reactive case. Perhaps it does for ease of configuration property binding and the like. Not something to worry about for this PR, but probably worth revisiting down the road before 3.2 GAs.

>  there are other technologies auto-configured in Boot whose reactive components leverage the imperative ones, no?

We have some situations where there's a low-level bean that's shared between imperative and reactive (Elasticsearch client transport and the ElasticsearchClient and ReactiveElasticsearchClient that use it, for example) but I can't think of a situation where we have an imperative bean that we expect applications to use extensively that's then wrapped by a reactive bean. I may well be forgetting something though.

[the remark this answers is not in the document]
onobc: > I wonder if the original, imperative client has to be a bean in the reactive case. Perhaps it does for ease of configuration property binding and the like. Not something to worry about for this PR, but probably worth revisiting down the road before 3.2 GAs.

I'm happy to take a look and revisit in that timeframe.

onobc:
We took a first pass and were able to knock it down from ~ `196 -> 136` which is about a 30% reduction. Would that work?

We also verified that for all of the properties we are removing, the user has a way to customize them (we are good on that front).

**NOTE** This has not been committed yet - just wanted to run the numbers by the team while the tests are being updated etc..


### Data below 

**NOTE:** the `Admin` and `Client` have `8 + 12` and `30 + 14`, respectively. These are the `regular + ssl` props. I believe the SSL related ones may go away when we move to SSL bundles but I am not 100% sure so I left it like this for now so we can see the count w/ and w/o the SSL props.

| PulsarProperties | Before | After |
| ---------------- |------- | ----- |
| Admin	| 8 + 12 |	8 + 12 |
| Cache	| 3 | 3 |
| Client | 30 + 14 | 7 + 13 |
| Defaults | 1 | 1 |
| Function | 3 | 3 |
| Listener | 6 | 6 |
| Reader | 7 | 7 |
| Template | 1 | 1 |
| ConsumerConfigProps | 36 | 26 |
| ProducerConfigProps | 27 | 17 |
| **Total** | 122 | 79

| ReactivePulsarProperties | Before | After |
| ------------------------ |------- | ----- |
| Cache	| 3 | 3 |
| Consumer | 34 | 27 |
| Listener | 3 | 3 |
| Reader | 8 | 8 |
| Sender | 26 | 16 |
| **Total** | 74 | 57 |

| Original Total | New Total |
| ------------- | ---------- |
| 196 | 136 | 

Which is about a 30% reduction.

onobc:
@wilkinsona @philwebb config props reduced ~50% w/ the latest commit. 


### Data below 

**NOTE:** the `Admin` and `Client` have `8 + 12` and `30 + 14`, respectively. These are the `regular + ssl` props. I believe the SSL related ones may go away when we move to SSL bundles but I am not 100% sure so I left it like this for now so we can see the count w/ and w/o the SSL props.

| PulsarProperties | Before | After |
| ---------------- |------- | ----- |
| Admin	| 8 + 12 |	8 + 12 |
| Client | 30 + 14 | 7 + 13 |
| Defaults | 1 | 1 |
| Function | 3 | 3 |
| Listener | 6 | 6 |
| Reader | 7 | 6 |
| Template | 1 | 1 |
| ConsumerConfigProps | 36 | 15 |
| ProducerConfigProps | 27 | 12 |
| **Total** | 119 | 59

| ReactivePulsarProperties | Before | After |
| ------------------------ |------- | ----- |
| Consumer | 34 | 17 |
| Listener | 3 | 3 |
| Reader | 8 | 6 |
| Sender | 26 | 12 |
| **Total** | 71 | 38 |

| Original Total | New Total | % Reduction |
| ------------- | ---------- | ---- |
| 190 | 97 | 51.05% |

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/ClientBuilderSslConfigurer.java
```diff
@@ -0,0 +1,167 @@
...
+		SslType sslBundleType = SslType.forSslStoreBundle(sslBundle.getStores());
+		switch (sslBundleType) {
+			case JKS -> applyJksSslProperties(clientBuilderSslSettings, map, bundleName);
+			case PEM -> applyPemSslProperties(clientBuilderSslSettings, map, bundleName);
+			case UNSUPPORTED -> throw new IllegalArgumentException(
+					"Unsupported store bundle type %s".formatted(sslBundle.getStores().getClass().getName()));
+		}
+	}
+
+	private void applyJksSslProperties(ClientBuilderSslSettings<?> clientBuilderSslSettings, PropertyMapper map,
```
onobc, line 100: The Pulsar client builder is a bit hokey, in that it has a handful of TLS properties to set when using PEM and a handful of properties to set when using JKS. Then it takes those and under the hood handles the creation of the SSLContext. The SSLContext is not exposed to the builder.
As such I need more info that what is provided by the SSLStoreBundle (basically if I could access the Pem/JksStoreBundleDetails) I would be good). Until that time comes though, I have to dip back down into the SslProperties to get the info I need on the builder.

spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-pulsar/src/test/java/smoketest/pulsar/SamplePulsarApplicationPemBasedSslTests.java
```diff
@@ -0,0 +1,62 @@
...
+
+import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.boot.test.context.SpringBootTest;
+import org.springframework.test.context.DynamicPropertyRegistry;
+import org.springframework.test.context.DynamicPropertySource;
+
+import static org.assertj.core.api.Assertions.assertThat;
+
+@SpringBootTest(
+		properties = { "spring.pulsar.client.ssl.bundle=client", "spring.pulsar.administration.ssl.bundle=client",
```
onobc, line 36: I have added an SSL smoke test pair (JKS+PEM) for both imperative and reactive. 

I see that for data cassandra we always have reactive+ssl  and imperative+ssl smoke test - but none w/o SSL. I think having one w/o SSL is valuable though as not everyone will be using SSL.

Wdyt?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
```diff
@@ -78,21 +77,31 @@ public class PulsarAutoConfiguration {
...
+				this.properties.getProducer().getCache().getMaximumSize(),
+				this.properties.getProducer().getCache().getInitialCapacity());
+	}
+
+	private List<ProducerBuilderCustomizer<Object>> lambdaSafeProducerBuilderCustomizers(
+			ObjectProvider<ProducerBuilderCustomizer<?>> customizersProvider) {
 		List<ProducerBuilderCustomizer<?>> customizers = new ArrayList<>();
 		customizers.add(PulsarPropertyMapper.producerBuilderCustomizer(this.properties));
 		customizers.addAll(customizersProvider.orderedStream().toList());
-		List<ProducerBuilderCustomizer<Object>> lambdaSafeCustomizers = List
```
onobc: @philwebb I broke this out into 2 separate beans for consistency w/ the reactive side and to have more concrete bean return types.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfigurationTests.java
```diff
@@ -179,6 +216,49 @@ void injectsExpectedBeans() {
...
+						"spring.pulsar.producer.name=fromPropsCustomizer")
+				.withUserConfiguration(ProducerBuilderCustomizersConfig.class)
+				.run((context) -> {
+					DefaultPulsarProducerFactory<?> producerFactory = context
+						.getBean(DefaultPulsarProducerFactory.class);
+					List<ProducerBuilderCustomizer<Object>> customizersOnFactory = (List<ProducerBuilderCustomizer<Object>>) ReflectionTestUtils
+						.getField(producerFactory, "defaultConfigCustomizers");
+					ProducerBuilder<Object> builder = mock(ProducerBuilder.class);
+					customizersOnFactory.forEach((customizer) -> customizer.customize(builder));
+					// Order of customizers should be default props customizer
```
onobc, line 234: I added this "customizer execution order" test for each component. The only way I found to do so was via the customizers all executing the same method w/ different values and verify the order. 

I am looking at creating a `CompositeCustomizer` that takes in a default + list of user defined customizers. That way we could test that functionality once and then just make sure that the composite is invoked.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -535,7 +535,7 @@ public static class Consumer {
 		 * Dead letter policy to use.
 		 */
 		@NestedConfigurationProperty
-		private final DeadLetterPolicy deadLetterPolicy = new DeadLetterPolicy();
+		private DeadLetterPolicy deadLetterPolicy;
```
onobc, line 538: The `new DeadLetterPolicy()` has a zero for max retry count and the builder fails w/ that non-positive value. I have added a consumer path on the Integration test to make sure we pass our defeault props through an actual consumer builder as the mock does not do this validation.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
```diff
@@ -7,7 +7,9 @@ It will do the same for the reactive components when `org.springframework.pulsar
 
 There are `spring-boot-starter-pulsar` and `spring-boot-starter-pulsar-reactive` "`Starters`" for conveniently collecting the dependencies for imperative and reactive use, respectively.
 
-
+TIP: The Reactive support is built on top of the imperative support and therefore the reactive starter includes the imperative starter.
```
onobc, line 10: I thought it may be helpful to list this fact and tell them how to "hide the non-reactive" components.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
philwebb: I've added an extra assert in our own mapper code as well to catch problems early.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/pulsar.adoc
philwebb: I removed this initially because the reactive starter doesn't include the imperative starter (although the reactive starter does include imperative dependencies). I've removed it again for now as I think this kind of detail probably belongs in the Spring Pulsar docs rather than ours.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfigurationTests.java
philwebb: I've added a test utility to extract the common testing logic. I don't think we need a `CompositeCustomizer` in order to merge.

spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-pulsar/src/test/java/smoketest/pulsar/SamplePulsarApplicationPemBasedSslTests.java
philwebb: We've decided to leave SSL out for now, although we'll document how to set it up using a customizer.

spring-boot-project/spring-boot-starters/spring-boot-starter-pulsar-reactive/build.gradle
onobc: https://github.com/spring-projects/spring-boot/pull/37801

Issue #34713, opened 2023-03-22
Banner placeholders do not support default values


Saucistophe:
The `ResourceBanner` uses placeholders with the regular syntax `${variable.name}` syntax.

There are only a few variables supported, which is fine; and some of them, such as the application version, is not resolved when starting the project from IDE (as opposed from the jar).

I wanted to use default values, for instance `${application.version:dev}`; when I do that, the printed version becomes `dev` regardless of whether it's defined.

I reproduced the issue in unit tests in `ResourceBannerTests`, and I think it is due to the way we handle resolvers:

```java
for (PropertyResolver resolver : getPropertyResolvers(environment, sourceClass)) {
    banner = resolver.resolvePlaceholders(banner);
}
```

If the first resolver fails to find the value, it outputs the default value and it's final.

wilkinsona:
It's all yours, @krzyk. Thank you.

Saucistophe:
If that can help, here is the test I wrote:
```java
	@Test
	void renderDefaultValues() {
		Resource resource = new ByteArrayResource(
				"banner ${a}${spring-boot.formatted-version:default-value}".getBytes());
		String banner = printBanner(resource, "10.2", "2.0", null);
		assertThat(banner).startsWith("banner 1 (v10.2)");
	}
```

..Which yields:
```
java.lang.AssertionError: 
Expecting actual:
  "banner 1default-value
"
to start with:
  "banner 1 (v10.2)"
```

krzyk:
@Saucistophe I just tried to recreate it but, well, it works.
Your testcase has error, because you provide the "10.2" as the second argument (this is boot version) so if you change it to `null` it will pass.

wilkinsona:
Closing in favor of #34764. Thanks for the PR, @krzyk.

PR #34764, opened 2023-03-25
Banner placeholders use default values too soon

krzyk:
For #34713.
Fixed support for default value placeholders in banner.
I had to change the way they are handled - replace PropertyResolvers with PropertySources (and add a custom one/anonymous one for Environment).

Added two tests to make sure it really fixed the problem and didn't break in case of missing values.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ResourceBanner.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ResourceBanner.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


wilkinsona (approved):

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ResourceBanner.java
```diff
@@ -77,18 +78,25 @@ public void printBanner(Environment environment, Class<?> sourceClass, PrintStre
...
-		resolvers.add(getVersionResolver(sourceClass));
-		resolvers.add(getAnsiResolver());
-		resolvers.add(getTitleResolver(sourceClass));
-		return resolvers;
+		MutablePropertySources propertySources = new MutablePropertySources();
+		propertySources.addLast(getEnvironmentSource(environment));
+		propertySources.addLast(getTitleSource(sourceClass));
+		propertySources.addLast(getAnsiSource());
+		propertySources.addLast(getVersionSource(sourceClass));
+		return Collections.singletonList(new PropertySourcesPropertyResolver(propertySources));
```
edwardsre, line 86: @wilkinsona Using `singletonList` here changed the behavior of this method by making the returned list unmodifiable. Was this intentional?

For ResourceBanner subclasses, the change to use an unmodifiable list introduced a breaking change. We were able to work around it by creating a new list, adding items from this list, along with our own PropertyResolver(s).

philwebb: Good spot. I've fixed that in a707c5e83e479099f8dd71fb06806e143efdd119

PR #34769, opened 2023-03-26
Offer a way to customize the converters used by Spring Batch without extending DefaultBatchConfiguration

EvaristeGalois11:
Hi, I'm trying to use the new feature of Spring Batch 5 that let you set a job parameter of any kind of type.

The problem is that you have to provide a `Converter` from your custom parameter type to `String` because Spring Batch needs it, but the only way to register this new converter is extending `DefaultBatchConfiguration` and overriding the `getConverterService()` method. This however switches off the spring boot autoconfiguration for Spring Batch, which seems a bit extreme just to register a simple converter.

I created a small project to demonstrate my situation https://github.com/EvaristeGalois11/spring-batch-parameter

In this small project i rely on the spring boot autoconfiguration to initialize the database schema needed by Spring Batch.

The `WithoutConverterTest` test case fails because the converter isn't registered, but the `WithConverterTest` test case fails too because the spring boot autoconfiguration isn't activated and the tables aren't being created.

### Possible solution ###
My simple idea to resolve the problem is to provide a bean similar to `Jackson2ObjectMapperBuilderCustomizer`.
If the spring boot autoconfiguration finds some beans of this type, the `ConfigurableConversionService` returned from `DefaultBatchConfiguration:getConverterService()` is enriched accordingly. This will permit registering a new converter without disrupting the autoconfiguration execution.

Let me know what you think, bye!

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 the original author or authors.
+ * Copyright 2012-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -15,6 +15,8 @@
  */
 
 package org.springframework.boot.autoconfigure.batch;
+
+import java.util.List;
 
 import javax.sql.DataSource;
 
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the PR, @EvaristeGalois11. It looks good to me. What do you think, @fmbenhassine?

fmbenhassine:
LGTM as well. However, I don't understand why we need to autowire a list of customizers here: https://github.com/spring-projects/spring-boot/pull/34769/files#diff-956b655780b21aca86c5bd17fc65d604d5d4e08b0bb0492ddf14028570e83a10R125. Are we supposed to provide multiple customizers? As a Spring Boot user, I would expect to provide a single `BatchConversionServiceCustomizer` that enriches the base one with all my custom converters (ie a centralized customization rather than probably multiple scattered customizer beans around the application context). wdyt?

That said, if this is the Boot way of doing things, then I have no objection to use the suggested approach.

fmbenhassine:
I see. I will defer this decision to the Boot team to introduce this feature in a consistent way with how to customize things in other parts of Spring Boot.

wilkinsona (approved):

wilkinsona:
@EvaristeGalois11 Thanks very much for making your first contribution to Spring Boot.

PR #34771, opened 2023-03-27
Added annotation descriptions as a caption

faizak1:
The original ReadMe does not have a caption associated with the sample program shown, hence it might not be as beginner-friendly to those who are not familiar with spring annotations. I have added a caption/description to the sample program which describes the 3 spring annotations used within the example program.

### DIFF
--- a/README.adoc
+++ b/README.adoc
@@ -43,7 +43,11 @@ Here is a quick teaser of a complete Spring Boot application in Java:
 
 	}
 ----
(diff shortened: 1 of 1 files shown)


snicoll:
@faizak1 thanks for the proposal. I wonder what the current version missed in terms of additional information. There is a link above the code sample that links to https://docs.spring.io/spring-boot/docs/current-SNAPSHOT/reference/html/getting-started.html#getting-started.first-application.code.mvc-annotations. The explanations on these annotations shouldn't be duplicated ideally. 

There are also a bunch of other concepts that may confuse newcomers. For instance, if you don't have the required dependency, no web server will start.

What do you think?

Issue #34770, opened 2023-03-26
Kafka autoconfiguration fails to resolve spring.kafka.bootstrap-servers when using Testcontainers' KafkaContainer


jucosorin:
Hello.

I have a Spring Boot project that uses Testcontainers Kafka for integration tests. I use the following to add the Kafka container to the `@SpringBootTest`:
```java
@Container
private static final KafkaContainer kafkaContainer = new KafkaContainer(DockerImageName.parse(TESTCONTAINERS_KAFKA_IMAGE))
      .withEnv("KAFKA_AUTO_CREATE_TOPICS_ENABLE", "FALSE");
```

In order to inject the _**spring.kafka.bootstrap-servers**_ I use **_@DynamicPropertySource_** like:

```java
@DynamicPropertySource
static void setProperties(DynamicPropertyRegistry registry) {
  registry.add("spring.kafka.bootstrap-servers", kafkaContainer::getBootstrapServers);
}
```

Since upgrading to 3.1.0-M2, injecting the _**spring.kafka.bootstrap-servers**_ fails and the root exception is:

```
Caused by: java.lang.NumberFormatException: For input string: "//localhost:54456"
	at java.base/java.lang.NumberFormatException.forInputString(NumberFormatException.java:67)
	at java.base/java.lang.Integer.parseInt(Integer.java:654)
	at java.base/java.lang.Integer.parseInt(Integer.java:786)
	at org.springframework.boot.autoconfigure.kafka.PropertiesKafkaConnectionDetails.asNode(PropertiesKafkaConnectionDetails.java:72)
```
Debugging this I've tracked this down to **_org.springframework.boot.autoconfigure.kafka.PropertiesKafkaConnectionDetails_** and the following method:

```java
private Node asNode(String bootstrapNode) {
  int separatorIndex = bootstrapNode.indexOf(':');
  if (separatorIndex == -1) {
    return new Node(bootstrapNode, this.DEFAULT_PORT);
  }
    return new Node(bootstrapNode.substring(0, separatorIndex),
      Integer.parseInt(bootstrapNode.substring(separatorIndex + 1)));
}
```
This method expects a **_bootstrapNode_** string like **_"localhost:9092"_**, but `KafkaContainer::getBootstrapServers()` sends it **_"PLAINTEXT://localhost:9092"_**.

@wilkinsona I see that this class was introduced recently. If you agree that this is a bug, I could look into it and try to resolve it.

quaff:
[...]

PR #34774, opened 2023-03-27
Fix regression introduced by 042f0c8

quaff:
Closes GH-34770

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/PropertiesKafkaConnectionDetails.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/PropertiesKafkaConnectionDetails.java
@@ -24,6 +24,7 @@ import java.util.List;
  * @author Moritz Halbritter
  * @author Andy Wilkinson
  * @author Phillip Webb
+ * @author Yanming Zhou
  */
 class PropertiesKafkaConnectionDetails implements KafkaConnectionDetails {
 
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/PropertiesKafkaConnectionDetails.java
```diff
@@ -64,7 +65,7 @@ private List<Node> asNodes(List<String> bootstrapServers) {
 	}
 
 	private Node asNode(String bootstrapNode) {
-		int separatorIndex = bootstrapNode.indexOf(':');
+		int separatorIndex = bootstrapNode.lastIndexOf(':');
 		if (separatorIndex == -1) {
 			return new Node(bootstrapNode, this.DEFAULT_PORT);
 		}
```
jucosorin, line 71: For a bootstrapNode with a value of "PLAINTEXT://localhost:9092" this implementation will return a Node with a host of "PLAINTEXT://localhost" and port of "9092". I am not sure that the host part is ok since I guess it should be just "localhost".

quaff: I think it should include protocol part to restore server node from `Node` object, maybe the `host` should rename to a more meaningful name.

wilkinsona:
Thanks for the PR, but as indicated by the labels and comments on #34774, we're not yet ready to implement a fix for this.

PR #34775, opened 2023-03-27
Use current Neo4j version in Testcontainers-based examples.

meistermeier:
I just spotted an outdated Neo4j version in the docs.
From the commit message:
Version 4.2 went out of support May 16, 2022.
With the more general and supported tag 5, it is
a) up to date and
b) maintenance friendly until 6 arrives.


snicoll:
Thanks again, @meistermeier.

PR #34796, opened 2023-03-28
Small assertions improvements

scordio:
Inspired by b91f814e427ccd5b99f0f7d1fafaae77ad1fb318, I propose a slightly simpler way to write such assertions.


mhalbritter:
Thank you!

Issue #34827, opened 2023-03-31
Documentation's "getting-started" page creates spring-boot application in default package


Rammerhead:
To be more specific, I'm referring to [Section 4.3 in the "Getting Started" page](https://docs.spring.io/spring-boot/docs/current/reference/html/getting-started.html#getting-started.first-application.code) at docs.spring.io (which is currently making use of Spring 3.0.5. I have OpenJDK 17 and I build using Maven 3.9.1 on Linux Mint 20.3).

The code provided puts the class in the default package (as no other package gets created), and this ends up causing problems because of `@SpringBootApplication`. Ultimately, the application fails to start after executing `mvn spring-boot:run`.

The solution is to create a package in MyApplication.java. I would also recommend to move the file from  `src/main/java/MyApplication.java` to `src/main/java/com/example/myproject/MyApplication.java` to be more consistent with the tip provided in ["Developing with Spring Boot" documentation's "Structuring Your Code"](https://docs.spring.io/spring-boot/docs/current/reference/html/using.html#using.structuring-your-code) section.

The changed code would finally look like this:
```java
package com.example.myproject;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.web.bind.annotation.RequestMapping;
import org.springframework.web.bind.annotation.RestController;

@RestController
@SpringBootApplication
public class MyApplication {

    @RequestMapping("/")
    String home() {
        return "Hello World!";
    }

    public static void main(String[] args) {
        SpringApplication.run(MyApplication.class, args);
    }

}
```

snicoll:
Superseded by #34838

Issue #34513, opened 2023-03-08
Code sample in "Developing Your First Spring Boot Application" does not work


tehnodev:
[MyApplication.java](https://docs.spring.io/spring-boot/docs/current/reference/html/getting-started.html#getting-started.first-application.code) example code do not work without package declaration. spring-boot:run command shows
```
...
** WARNING ** : Your ApplicationContext is unlikely to start due to a @ComponentScan of the default package.
...
Caused by: java.lang.IllegalStateException: Could not evaluate condition on org.springframework.boot.autoconfigure.r2dbc.ConnectionFactoryConfigurations$PoolConfiguration due to io/r2dbc/spi/ValidationDepth not found. Make sure your own configuration does not rely on that class. This can also happen if you are @ComponentScanning a springframework package (e.g. if you put a @ComponentScan in the default package by mistake)
...
```

bclozel:
In which package is your main application located? The error message seems to indicate that it is using an invalid location.

Can you share a sample application showing the problem?

tehnodev:
No package. File is located as stated in documentation at src/main/java/MyApplication.java

tehnodev:
I suggest to change file location to src/main/java/sample/MyApplication.java and add "package sample;"  at the top of example file

petrovskimario:
Has this been picked up by anyone ? If not i can pick this up

wilkinsona:
Thanks for the offer, @petrovskimario. It's all yours.

The [code](https://github.com/spring-projects/spring-boot/blob/295702003ddb57bd5d6cfaf52a1c1426ced26249/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/gettingstarted/firstapplication/code/MyApplication.java) already has a package but it's deliberately removed by the [code chomping in our Asciidoctor backend](https://github.com/spring-io/spring-asciidoctor-backends#code-chomping). It should be possible to use `chomp_package_replacement` to configure the package for this particular code snippet. Please let us know if you have any questions.

wilkinsona:
You can set a document-wide attribute in `attributes.adoc`:

```
:chomp_package_replacement: example
```

But this will affect every code snippet in the documentation. I thought it should also be possible to set it on a per-snippet basis like this:

```
include::code:MyApplication[chomp_package_replacement=example]
```

But it doesn't seem to work. Should it have worked, @philwebb?

With the global attribute in place, the package always appears, even when the snippet isn't expanded:

```java
[...]

PR #34838, opened 2023-04-03
getting-started: move MyApplication.java to it's own package

Rammerhead:
In its current state, MyApplication.java does not have it's own package and exists in the default package which conflicts with SpringBootApplication. The changes here move the create com.example.myproject package and move MyApplication under it.
This also comes with creating a directory structure for the package and moving the code file under it. Fixes issue [#34827](https://github.com/spring-projects/spring-boot/issues/34827)

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/first-application.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/first-application.adoc
@@ -134,10 +134,31 @@ If you run `mvn dependency:tree` again, you see that there are now a number of a
 
 [[getting-started.first-application.code]]
 === Writing the Code
-To finish our application, we need to create a single Java file.
-By default, Maven compiles sources from `src/main/java`, so you need to create that directory structure and then add a file named `src/main/java/MyApplication.java` to contain the following code:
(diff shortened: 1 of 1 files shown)


snicoll:
@Rammerhead unfortunately, I don't think this is an obvious fix so you'll have to sign the CLEA if we decide to pursue this.

This PR is moving code back to the guide, which is something we're not keen on doing. If we want to make the package explicit, then we need to expand the tag used in the source code. Unfortunately, the current package is not something we'd use in the guide.

> In its current state, MyApplication.java does not have it's own package and exists in the default package

I don't think this is accurate. This is a sample code, not the complete file. It won't compile as it does not include a package statement. 

Can we take a step back and can you describe what you're trying to solve?

snicoll:
No worries and thanks for following-up.

izeye:
For your information, https://github.com/spring-projects/spring-boot/issues/34827 seems to be a duplicate of https://github.com/spring-projects/spring-boot/issues/34513, and specifically, the `src/main/java/MyApplication.java` part in [the section](https://docs.spring.io/spring-boot/docs/current/reference/html/getting-started.html#getting-started.first-application.code) seems to be a problem.

PR #34868, opened 2023-04-05
Fix banner location value hard coded

kdomo:
I think there is no reason to use hardcoded "baner.txt" instead of DEFAULT_BANNER_LOCATION


philwebb:
I agree. Good spot.

snicoll:
@kdomo thank you for making your first contribution to Spring Boot.

Issue #34871, opened 2023-04-06
ErrorController should produce problem details if spring.mvc.problemdetails.enabled=true


quaff:
Currently `ErrorController` will produce traditional error response even if problem details is enabled, nobody want to mix two different style of error handling.
```java
    @RequestMapping("/")
    void home(HttpServletResponse response) throws IOException {
        response.sendError(HttpServletResponse.SC_BAD_REQUEST, "bad.request");
    }
```
```java
	@Test
	void sendErrorShouldProduceProblemDetails() {
		ResponseEntity<Map<String, Object>> resp = testRestTemplate.exchange(RequestEntity.method(GET, "/").build(),
				new ParameterizedTypeReference<Map<String, Object>>() {
				});
		assertThat(resp.getBody()).containsKey("detail");
	}
```
```
Expecting actual:
  {"error"="Bad Request", "message"="bad.request", "path"="/", "status"=400, "timestamp"="2023-04-06T02:01:46.401+00:00"}
to contain key:
  "detail"
```
Here is minimal project [problemdetails.zip](https://github.com/spring-projects/spring-boot/files/11165661/problemdetails.zip)

snicoll:
Thanks for the report, this is a duplicate of #33885

Issue #33885, opened 2023-01-18
Inconsistent error response when problemdetails enabled


jonatan-ivanov:
Affects: Boot 3.0.1, Framework 6.0.3

With the current version (Boot 3.0.1, Framework 6.0.3), if Problem Details are enabled (`spring.mvc.problemdetails.enabled=true`) and if I throw out an error from an MVC controller method, sometimes I get a Problem Details response, sometimes I get the "old" (Boot 2.x, Framework 5.x) response format depending on the default error handler "knows" the exception or not.

```
@GetMapping("/")
public String trouble() {
	// throw new IllegalStateException("Noooooo!"); // "old" (5.x) response
	throw new ResponseStatusException(INTERNAL_SERVER_ERROR, "Noooooo!"); // -> problem details response
}
```

**When I throw `IllegalStateException`**
```
HTTP/1.1 500
Connection: close
Content-Type: application/json
Date: Wed, 18 Jan 2023 21:34:42 GMT
Transfer-Encoding: chunked

{
    "error": "Internal Server Error",
    "path": "/",
    "status": 500,
    "timestamp": "2023-01-18T21:34:42.764+00:00"
}
```

**When I throw `ResponseStatusException`**
```
HTTP/1.1 500
Connection: close
Content-Type: application/problem+json
Date: Wed, 18 Jan 2023 21:35:04 GMT
Transfer-Encoding: chunked

{
    "detail": "Noooooo!",
    "instance": "/",
    "status": 500,
    "title": "Internal Server Error",
    "type": "about:blank"
}
```

I think the expected behavior would be a consistent response format; `application/problem+json` if problem details enabled and the old (5.x) `application/json` format if not.

bclozel:
Superseded by https://github.com/spring-projects/spring-boot/issues/43850

PR #34872, opened 2023-04-06
Make BasicErrorController respect problem details

quaff:
Fix GH-34871


snicoll:
Thanks for the report but I agree with https://github.com/spring-projects/spring-boot/issues/33885#issuecomment-1396643234, and that requires a bit more design consideration on our end before proceeding.

Issue #8611, opened 2017-03-14
Refine logging packages


philwebb:
The current `org.springframework.boot.logging` package contains logging infrastructure and application listeners. It might be nicer to separate the concerns so that the logging system doesn't need to know about `context` concerns.

PR #34900, opened 2023-04-07
Short circuit @Value injection when using a source already covered by ConfigurationPropertySourcesPropertySource

### LINKED ISSUE #28687: Short circuit @Value injection when using a source already covered by ConfigurationPropertySourcesPropertySource
https://github.com/spring-projects/spring-boot/blob/aaad7727f0fff71b7614bbf10cf4712cd149c84b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/ConfigurationPropertySourcesPropertySource.java#L33

In Springboot application，ConfigurationPropertySourcesPropertySource is added to the environment and as first element, the ConfigurationPropertySourcesPropertySource acts as a facade over the existing property sources.

When handling `@Value` annotation, [org.springframework.core.env.PropertySourcesPropertyResolver#getProperty](https://github.com/spring-projects/spring-framework/blob/c705e32a887b55a746afb4c93cb4c6a031eae4c5/spring-core/src/main/java/org/springframework/core/env/PropertySourcesPropertyResolver.java#L78) will be called.
 If the property is found then ConfigurationPropertySourcesPropertySource will return the result. otherwise, PropertySourcesPropertyResolver continues to check all subsequent sources. This is a little inefficient for any source that has already been check via ConfigurationPropertySourcesPropertySource.

terminux:
Hi, I found that `PropertySourcesPropertyResolver` (a resolver that does not support short-circuiting) is used in these places:

* `LoggingSystemProperties`

https://github.com/spring-projects/spring-boot/blob/08ce0919cd7895a9551018c072985d014b8cd295/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggingSystemProperties.java#L165

* When the key is not a property name, such as when it is a SpEL expression:

https://github.com/spring-projects/spring-boot/blob/5817c8441d4a24ca0ad6efe70bec64d22875b652/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ErrorProperties.java#L35

In this case `ConfigurationPropertyName.of(key, true)` will return null, and then `defaultResolver` (inherited from `PropertySourcesPropertyResolver`) will be used for resolution.

https://github.com/spring-projects/spring-boot/blob/5817c8441d4a24ca0ad6efe70bec64d22875b652/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/ConfigurationPropertySourcesPropertyResolver.java#L87-L97

However the entire process property value will only be retrieved once, the current implementation is fine and we don't need to make any changes.

https://github.com/spring-projects/spring-boot/blob/5817c8441d4a24ca0ad6efe70bec64d22875b652/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/source/ConfigurationPropertySourcesPropertySource.java#L66-L68

* As @philwebb  [commented](https://github.com/spring-projects/spring-boot/issues/28687#issuecomment-990148089), Boot auto-configures `PropertySourcesPlaceholderConfigurer`, may need to wait spring-projects/spring-framework#30304 to fix

https://github.com/spring-projects/spring-boot/blob/5817c8441d4a24ca0ad6efe70bec64d22875b652/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/PropertyPlaceholderAutoConfiguration.java#L43

Closes gh-28687

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggingSystemProperties.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggingSystemProperties.java
@@ -21,11 +21,12 @@ import java.nio.charset.StandardCharsets;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
+import org.springframework.boot.context.properties.source.ConfigurationPropertySources;
 import org.springframework.boot.system.ApplicationPid;
 import org.springframework.core.env.ConfigurableEnvironment;
+import org.springframework.core.env.ConfigurablePropertyResolver;
 import org.springframework.core.env.Environment;
 import org.springframework.core.env.PropertyResolver;
-import org.springframework.core.env.PropertySourcesPropertyResolver;
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
@@ -217,8 +218,8 @@ public class LoggingSystemProperties {
 
 	private PropertyResolver getPropertyResolver() {
 		if (this.environment instanceof ConfigurableEnvironment configurableEnvironment) {
-			PropertySourcesPropertyResolver resolver = new PropertySourcesPropertyResolver(
-					configurableEnvironment.getPropertySources());
+			ConfigurablePropertyResolver resolver = ConfigurationPropertySources
+					.createPropertyResolver(configurableEnvironment.getPropertySources());
 			resolver.setConversionService(((ConfigurableEnvironment) this.environment).getConversionService());
 			resolver.setIgnoreUnresolvableNestedPlaceholders(true);
 			return resolver;
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/ConfigurationPropertySourcesPlaceholderConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/ConfigurationPropertySourcesPlaceholderConfigurer.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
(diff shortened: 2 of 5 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggingSystemProperties.java
```diff
@@ -20,11 +20,12 @@
 import java.nio.charset.StandardCharsets;
 import java.util.function.BiConsumer;
 
+import org.springframework.boot.context.properties.source.ConfigurationPropertySources;
```
terminux, line 23: The current `logging` package does not allow the content of the `context` package to be imported, I am not sure if I need to modify the rules or use other methods to create property resolver.

https://github.com/spring-projects/spring-boot/blob/2b5bd71eb46b9f7295e195b3db5b7fa27ce9b205/src/checkstyle/import-control.xml#L80

philwebb: That restriction was introduced in https://github.com/spring-projects/spring-boot/issues/8611. I'm not sure if we should relax it or not. I wonder if we can get away with using the `Environment` directly since it's already a `PropertyResolver`.

terminux: Thank you for your reply! If you use `Environment` directly, we may need to set its `ignoreUnresolvableNestedPlaceholders` property to `true` to ensure that it is consistent with the original behavior. Since the `Environment` is modified, this will also affect the external use of it. Maybe we should make a copy of `Environment`?

philwebb: Good point. I need to think about this one for a bit.

sparecycles:
huh.  I just came across some behavior that might be related to this: https://github.com/spring-projects/spring-boot/issues/49358

Issue #34854, opened 2023-04-05
ImportsContextCustomizer does not support AliasFor


lmartelli:
I've created [some test cases to demonstrate the problem](https://github.com/lmartelli/spring-boot/commit/ae39174a9abbda917032b925e70dab7fa1f82c4c) : 
the test that use a custom with `@AliasFor` on `@Import` fails (customizersForClassesWithDifferentAliasedImportsAreNotEqual), whereas those with a direct use of `@Import` or use through a custom annotation with a hard coded `@Import` pass.

snicoll:
Thanks for the report. `ContextCustomizerKey` does not use `MergedAnnoation` and friends so it does not support aliasing at all.

lmartelli:
I may have some time to work on a PR for this. Just to be sure, the requirement for equality is that 2 `ContextCustomizerKey` should be equal if and only if their `testClass` have the same annotations, ignoring a set of annotations (JavaLang, Kotlin, Junit and Spock). Am I correct ?

lmartelli:
There is this `DeterminableImports` thing to take into account also ...

wilkinsona:
Closing in favor of #34917.

PR #34917, opened 2023-04-08
ImportsContextCustomizer does not support AliasFor

lmartelli:
Fixes https://github.com/spring-projects/spring-boot/issues/34854

Use `MergedAnnotations` in order to support `@AliasFor`.

I used plain `Predicate` instead of `AnnotationFilter` because it is much simpler and generic.

### DIFF
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/ImportsContextCustomizer.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/ImportsContextCustomizer.java
@@ -20,7 +20,6 @@ import java.lang.reflect.Constructor;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
-import java.util.function.Predicate;
 import java.util.stream.Collectors;
 import java.util.stream.Stream;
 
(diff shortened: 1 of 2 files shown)


wilkinsona (changes requested):
Thanks for the proposal, @lmartelli. We try to minimise the scope of changes that we make, particularly in maintenance releases. With this in mind, these changes are too broad. For example, the move from `AnnotationFilter` to `Predicate` is not necessary to fix the problem. We also don't use `var` in Spring Boot's code. Would you like to rework your proposal so that it only makes the changes that are required to fix the problem and nothing more?

lmartelli:
> > Would you like to rework your proposal so that it only makes the changes that are required to fix the problem and nothing more?
> 
> Sure, I'll do that.

@wilkinsona But note that we still have to remove `ImportsContextCustomizer.AnnotationFilter` and use the one from `org.springframework.core.annotation.AnnotationFilter` instead since that's what `org.springframework.core.annotation.MergedAnnotations.Search#withAnnotationFilter()` wants.

lmartelli:
@wilkinsona is the new version OK ?

snicoll:
This still looks quite a large change to me and I am not sure the use of `stream` and `reduce` is something we want. The `MergedAnnotations` API you've used doesn't exist in `2.7.x` and we need to be able to rebase the change there. Can you please limit your change to support aliases and only that? We can refine the code further as a separate step.

philwebb:
Thanks very much @lmartelli, I've taken the tests from this commit and merged them into 2.7 along with a [minimal fix](https://github.com/spring-projects/spring-boot/commit/5d0640a35636368d149ea35b3e3c7017f1ddab5d). I've also taken the changes you made to `ImportsContextCustomizer` and applied them [with a little polish](https://github.com/spring-projects/spring-boot/commit/7c942679ad85f254acaa0318cf07753c65a21716) to `main` (see #36211).

Thanks again!

Issue #34787, opened 2023-03-28
Allow auto-configuring IdGenerator in OpenTelemetryAutoConfiguration


gnagy:
AWS XRay expects traceIds to be in a specific format. Otel SDK allows customizing id generation. Please consider extending `OpenTelemetryAutoConfiguration.otelSdkTracerProvider()` to support overriding this.

Something like:

```
// depends on io.opentelemetry.contrib:opentelemetry-aws-xray:1.23.0
SdkTracerProvider.builder().setIdGenerator(AwsXrayIdGenerator.getInstance())...
```

vpavic:
I agree that this is a bit of a pain, as it requires users to provide a configuration with a custom `SdkTracerProvider` and also manually configure several other dependent components.

I've opened #35041 to consider some improvements here.

quaff:
It's covered by https://github.com/spring-projects/spring-boot/pull/34945

jonatan-ivanov:
Closing this in favor of (probably) https://github.com/spring-projects/spring-boot/pull/34945.

gnagy:
Thanks @jonatan-ivanov , #34945 looks better 👍 

Do you mean X-Ray supporting standard headers as in not needing a custom ID generator, or standard header names?

If the former, AFAIK the reason for custom X-Ray id format is that they encode the timestamp in the trace ID so they can performantly drop too old data in a stateless way on ingress.

If the latter, their recommended approach is to process headers in the collector agent (aws-otel-collector), so no need to change apps for that.

jonatan-ivanov:
Kind of both: supporting the w3c standard would mean both the format and the header name.

Unfortunately the agent approach has many issues but I don't see too high chance of AWS supporting the w3c standard.

PR #34945, opened 2023-04-12
Allow to customize opentelemetry SdkTracerProvider

quaff:

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/SdkTracerProviderCustomizer.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/SdkTracerProviderCustomizer.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/SdkTracerProviderCustomizer.java
```diff
@@ -0,0 +1,37 @@
...
+
+/**
+ * Callback interface that can be used to customize auto-configured
+ * {@link SdkTracerProviderBuilder}.
+ *
+ * @author Yanming Zhou
+ * @since 3.1.0
+ */
+@FunctionalInterface
+public interface SdkTracerProviderCustomizer {
```
wilkinsona, line 29: I think this should be called `SdkTracerProviderBuilderCustomizer`. We can take care of that as part of merging it.

quaff: Actually I named it that at first, after some thoughts, I think `SdkTracerProviderCustomizer` is identical to `SdkTracerProviderBuilderCustomizer`, because `SdkTracerProvider` cannot constructed without `SdkTracerProviderBuilder`, and the only purpose of `SdkTracerProviderBuilder` is building `SdkTracerProvider`.

jonatan-ivanov:
Let me link things that belong together here:
- Merging this should also provide a solution for https://github.com/spring-projects/spring-boot/issues/34787 (so I closed that)
- It seems there is another PR that could be obviated by his one: https://github.com/spring-projects/spring-boot/pull/35041

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/SdkTracerProviderCustomizer.java
wilkinsona: I think we should aim for consistency with other similar interfaces that already exist. We have many that are named `…BuilderCustomizer` because the class that they are customised is a `…Builder`. The same applies here.

wilkinsona:
Thanks for the housekeeping, @jonatan-ivanov.

PR #34991, opened 2023-04-14
Improve documentation of tracing dependencies and configuration properties

RobertRad:
Document which property should contain the Zipkin URL.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/tracing.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/tracing.adoc
@@ -95,6 +95,7 @@ All tracer implementations need the `org.springframework.boot:spring-boot-starte
 
 * `io.micrometer:micrometer-tracing-bridge-brave` - which is needed to bridge the Micrometer Observation API to Brave.
 * `io.zipkin.reporter2:zipkin-reporter-brave` - which is needed to report traces to Zipkin.
+* Point the property `management.zipkin.tracing.endpoint` to the Zipkin server
 
 NOTE: If your project doesn't use Spring MVC or Spring WebFlux, the `io.zipkin.reporter2:zipkin-sender-urlconnection` dependency is needed, too.
 
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/tracing.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/tracing.adoc
@@ -72,38 +72,43 @@ As Micrometer Tracer supports multiple tracer implementations, there are multipl
 
 All tracer implementations need the `org.springframework.boot:spring-boot-starter-actuator` dependency.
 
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/tracing.adoc
```diff
@@ -95,6 +95,7 @@ All tracer implementations need the `org.springframework.boot:spring-boot-starte
 
 * `io.micrometer:micrometer-tracing-bridge-brave` - which is needed to bridge the Micrometer Observation API to Brave.
 * `io.zipkin.reporter2:zipkin-reporter-brave` - which is needed to report traces to Zipkin.
+* Point the property `management.zipkin.tracing.endpoint` to the Zipkin server
```
wilkinsona, line 98: Each item in the list is a dependency so I don't think this is quite the right place to draw attention to the `management.zipkin.tracing.endpoint` property. If we add something for the Zipkin property, we should also do the same for the `management.wavefront.uri` property.

I wonder if we'd be better restructuring `Tracer Implementations` to have `Zipkin` and `Wavefront` subsections. We could then mention the property in their respective sections and list the two dependency combinations that each supports as well.

wilkinsona (changes requested):

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/tracing.adoc
wilkinsona: Thanks for the changes, @RobertRad. Unfortunately, now that I've seen them, I've realised that what I proposed isn't quite right as Zipkin and Wavefront are not tracer implementations. They're tracing systems to which a tracer exports traces. I'll have give this one some more thought.

wilkinsona:
@RobertRad Thank you very much for making your first contribution to Spring Boot. I didn't want to waste more of your time so I went with an amended version of your first suggestion in the end. I've mentioned the configuration properties for Wavefront and Zipkin in each section where they're relevant. There's a bit of repetition but I think that's unavoidable without a much bigger overhaul of the documentation.

spring-boot-project/spring-boot-docs/src/docs/asciidoc/actuator/tracing.adoc
RobertRad: Thanks @wilkinsona 

I just wonder, how the `management.zipkin.tracing.endpoint` property is now documented?
I had a hard time getting tracing to work with Spring Boot 3 and needed to go through the Spring source code, to find out that property.
That was the reason, why I started this pull request.

wilkinsona: It's [listed in the appendix](https://docs.spring.io/spring-boot/docs/3.0.x/reference/htmlsingle/#application-properties.actuator.management.zipkin.tracing.endpoint). It will also be offered by auto-completion in your IDE if you have Spring Boot tooling enabled/installed.

PR #35015, opened 2023-04-17
Replace Mockito argument captors with assertArg

### LINKED ISSUE #34943: Replace Mockito argument captors with assertArg
See https://github.com/mockito/mockito/pull/2949 for details. We probably have a few captors we can replace [when we upgrade Mockito](https://github.com/spring-projects/spring-boot/issues/34998).

Ckram:
Replace Mockito argument captors with assertArg

Fixes #34943


mhalbritter:

wilkinsona:
I think there are some more places where we can use `assertArg`. I started polishing in https://github.com/wilkinsona/spring-boot/commit/33b8dfbf35ab196ea0d6a1d16405a63924e6a8c6 I stopped as I reached a place where https://github.com/mockito/mockito/issues/2982 would be useful. I think it'd be nice to move over everywhere we can without the enhancement though.

PR #35041, opened 2023-04-17
Add support for providing custom OTel IdGenerator

vpavic:
At present, `OpenTelemetryAutoConfiguration` offers no easy way of providing custom `IdGenerator`, which is necessary for some tracing systems (for example, AWS X-Ray). Consequently, user that need custom `IdGenerator` are forced to replace significant bits of `OpenTelemetryAutoConfiguration`.

This commit adds support for providing custom `IdGenerator` bean, that will be used in auto-configured `SdkTracerProvider`.

Closes gh-34787


jonatan-ivanov:
Thank you for the PR!
What do you think about this and https://github.com/spring-projects/spring-boot/pull/34945?
I feel that could supersede this one since it seems a little more general.

vpavic:
I wasn't aware of #34945 as it wasn't referenced from #34787 - I do agree it's more flexible and therefore generally usable so I'll close this one.

jonatan-ivanov:
Superseded by https://github.com/spring-projects/spring-boot/pull/34945

PR #35080, opened 2023-04-19
Add service connection from Testcontainers ActiveMQ

eddumelendez:


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/activemq/ActiveMQConnectionFactoryFactory.java
```diff
@@ -79,7 +81,7 @@ private <T extends ActiveMQConnectionFactory> T doCreateConnectionFactory(Class<
 
 	private <T extends ActiveMQConnectionFactory> T createConnectionFactoryInstance(Class<T> factoryClass)
 			throws InstantiationException, IllegalAccessException, InvocationTargetException, NoSuchMethodException {
-		String brokerUrl = determineBrokerUrl();
+		String brokerUrl = this.connectionDetails.getBrokerUrl();
```
wilkinsona, line 84: In addition to the broker URL, I think that `user` and `password` should come from the connection details as well, otherwise it won't be possible to use connected to an ActiveMQ service that requires authentication.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/activemq/PropertiesActiveMQConnectionDetails.java
```diff
@@ -0,0 +1,38 @@
...
+
+package org.springframework.boot.autoconfigure.jms.activemq;
+
+/**
+ * Adapts {@link ActiveMQProperties} to {@link ActiveMQConnectionDetails}.
+ *
+ * @author Eddú Meléndez
+ * @since 3.1.0
+ */
+public class PropertiesActiveMQConnectionDetails implements ActiveMQConnectionDetails {
```
wilkinsona, line 25: I think this could be package-private, and perhaps an inner-class of the `@Configuration` where it's defined.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/activemq/ConnectionDetailsConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+package org.springframework.boot.autoconfigure.jms.activemq;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+@Configuration(proxyBeanMethods = false)
+@EnableConfigurationProperties(ActiveMQProperties.class)
+class ConnectionDetailsConfiguration {
```
wilkinsona, line 26: Do we need a separate class here? Instead, could the bean be defined directly in the auto-configuration class that imports this class?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/activemq/ConnectionDetailsConfiguration.java
```diff
@@ -0,0 +1,34 @@
...
+
+	@Bean
+	@ConditionalOnMissingBean(ActiveMQConnectionDetails.class)
+	ActiveMQConnectionDetails activemqConnectionDetails(ActiveMQProperties properties) {
```
wilkinsona, line 30: Tiny nit, I'd name this method `activeMqConnectionDetails`.

wilkinsona (changes requested):
Thanks very much for the PR, @eddumelendez. You're quite prolific! I've left a few comments for your consideration when you have time.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/activemq/ConnectionDetailsConfiguration.java
eddumelendez: right, thanks! I totally mess up my mind :D

eddumelendez:
I've updated the PR addressing your comment. Thanks for the quick review, @wilkinsona !

spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/service/connection/activemq/ActiveMQContainerConnectionDetailsFactory.java
```diff
@@ -0,0 +1,79 @@
...
+ * from a {@link ServiceConnection @ServiceConnection}-annotated {@link GenericContainer}
+ * using the {@code "symptoma/activemq"} image.
+ *
+ * @author Eddú Meléndez
+ */
+class ActiveMQContainerConnectionDetailsFactory
+		extends ContainerConnectionDetailsFactory<ActiveMQConnectionDetails, Container<?>> {
+
+	ActiveMQContainerConnectionDetailsFactory() {
+		super("symptoma/activemq");
```
eddumelendez, line 40: I've updated this from `activemq` to `symptoma/activemq` because it's similar to the Testcontainers mechanism to make sure this works with this image or a similar one to this. The image besides running on arm too it provides good mechanism to set username and password

snicoll:
Thanks very much Eddú. I've polished the contribution to add support for Docker Compose as well.

PR #35084, opened 2023-04-20
Inject user-defined Attributes into SdkTracerProvider

quaff:
User-defined attributes is useful for tag at service level, for example Jaeger UI treat it as process tag beside span tag.

We can define a `SdkTracerProviderBuilderCustomizer` to customize Attributes, but it is tedious and you have to include `ResourceAttributes.SERVICE_NAME` again.

```java
	@Bean
	SdkTracerProviderBuilderCustomizer sdkTracerProviderBuilderCustomizer(Environment environment) {
		return builder -> {
			String applicationName = environment.getProperty("spring.application.name", "application");
			Attributes attributes = Attributes.of(ResourceAttributes.SERVICE_NAME, applicationName,
					AttributeKey.stringKey("java.version"), System.getProperty("java.version"));
			builder.setResource(Resource.create(attributes));
		};
	}

```
now you can do it like this
```java
	@Bean
	Attributes javaAttributes() {
		return Attributes.of(AttributeKey.stringKey("java.version"), System.getProperty("java.version"));
	}
```


wilkinsona (changes requested):
Thanks for the proposal. I don't think we can use an `Attributes` bean to affect the configuration of the `SdkTracerProvder` as it doesn't feel sufficiently precise. `Attributes` are used throughout the OTtel SDK and defining an `Attributes` bean doesn't make it clear where those attributes will be applied.

As an alternative, how about a default method on the new `SdkTracerProviderBuilderCustomizer` interface that is called with the `AttributesBuilder`?

quaff:
> SdkTracerProviderBuilderCustomizer

`SdkTracerProviderBuilder` doesn't expose method to obtain existed `Resource`, It means you have to inject `Environment` and set `ResourceAttributes.SERVICE_NAME`, which is done by spring boot already.

quaff:
Do you mean add default method to `SdkTracerProviderBuilderCustomizer` like this?
```java
@FunctionalInterface
public interface SdkTracerProviderBuilderCustomizer {

	/**
	 * Customize the given {@code builder}.
	 * @param builder the builder to customize
	 */
	void customize(SdkTracerProviderBuilder builder);

	/**
	 * Customize the given {@code builder}.
	 * @param builder the builder to customize
	 */
	default void customize(AttributesBuilder builder) {

	}
}
```
```java
		String applicationName = environment.getProperty("spring.application.name", DEFAULT_APPLICATION_NAME);
		AttributesBuilder attributesBuilder = Attributes.builder();
		attributesBuilder.put(ResourceAttributes.SERVICE_NAME, applicationName);
		customizers.orderedStream().forEach((customizer) -> customizer.customize(attributesBuilder));
```
```java
		@Bean
		SdkTracerProviderBuilderCustomizer sdkTracerProviderBuilderCustomizer() {
			return new SdkTracerProviderBuilderCustomizer() {

				@Override
				public void customize(SdkTracerProviderBuilder builder) {

				}

				@Override
				public void customize(AttributesBuilder builder) {
					builder.put(AttributeKey.stringKey("java.version"), System.getProperty("java.version"));
				}
			};
		}
```

wilkinsona:
Yes, that's what I meant. It's somewhat similar to what we've done in `org.springframework.boot.autoconfigure.elasticsearch.RestClientBuilderCustomizer` for example.

quaff:
> Yes, that's what I meant. It's somewhat similar to what we've done in `org.springframework.boot.autoconfigure.elasticsearch.RestClientBuilderCustomizer` for example.

Updated as your suggestion.

jonatan-ivanov:
@quaff Did you check the `ObservationFilter` component?
With it you can add `KeyValue`s that will translate to `Attributes`.
It has two advantages:
- It does not depend on OTel so it makes easier if you want to migrate
- You have a control to set attributes to metrics-only or both metrics and spans (or other things you can come up with)

I don't think it defeats the purpose of this customizer (e.g.: if you use OTel directly) but might help.

quaff:
> @quaff Did you check the `ObservationFilter` component? With it you can add `KeyValue`s that will translate to `Attributes`. It has two advantages:
> 
> * It does not depend on OTel so it makes easier if you want to migrate
> * You have a control to set attributes to metrics-only or both metrics and spans (or other things you can come up with)
> 
> I don't think it defeats the purpose of this customizer (e.g.: if you use OTel directly) but might help.

@jonatan-ivanov `ObservationFilter` works, can you give me a hint how to control to set attributes to metrics-only or both metrics and spans?

jonatan-ivanov:
By default the meter handler will use `.lowCardinalityKeyValue` and the tracing handler will use both low- and `.highCardinalityKeyValue`.
So:
- Both: add with `.lowCardinalityKeyValue`
- Spans-only: add with `.highCardinalityKeyValue`
- Metrics-only: does not exists by default, either you do both and remove tags using a `SpanFilter` or you override `DefaultTracingHandler`.

What is the use-case to make the information available only for metrics?

quaff:
> By default the meter handler will use `.lowCardinalityKeyValue` and the tracing handler will use both low- and `.highCardinalityKeyValue`. So:
> 
> * Both: add with `.lowCardinalityKeyValue`
> * Spans-only: add with `.highCardinalityKeyValue`
> * Metrics-only: does not exists by default, either you do both and remove tags using a `SpanFilter` or you override `DefaultTracingHandler`.
> 
> What is the use-case to make the information available only for metrics?

Thanks for your elaboration, I'd like to use `MeterRegistryCustomizer` to configure `commonTags` for metrics-only.
If I understand correctly, spans-only must be added by `.highCardinalityKeyValue` even though it's low cardinality like `java.version`?

jonatan-ivanov:
Yes, Micrometer does not know the cardinality of the tags so you can add low cardinality tags as `highCardinalityKeyValue` (and vice-versa but if you attach high cardinality tags as `lowCardinalityKeyValue`, your app and your metric backend will not like the high cardinality data).

quaff:
> @quaff Did you check the `ObservationFilter` component? With it you can add `KeyValue`s that will translate to `Attributes`. It has two advantages:
> 
> * It does not depend on OTel so it makes easier if you want to migrate
> * You have a control to set attributes to metrics-only or both metrics and spans (or other things you can come up with)
> 
> I don't think it defeats the purpose of this customizer (e.g.: if you use OTel directly) but might help.

@jonatan-ivanov tags added by `ObservationFilter` is span level not process level, not my desired behavior.

jonatan-ivanov:
I'm not sure I understand what you mean by span vs. process level tags.

quaff:
> I'm not sure I understand what you mean by span vs. process level tags.

You can check out the screenshot image at the begin of this thread, Jaeger UI will separate process level tags from span level tags, process level tags is like `commonTags` of `MeterRegistry`

jonatan-ivanov:
Do you happen to know if Jaeger behaves the same if you use the Zipkin format? If so, I _think_ you "just" need to take care about the naming of your tags and they should become "process" attributes. I don't know if this is an existing feature or not (should be) or what is the naming convention (if any).

amithkumarg:
> Do you happen to know if Jaeger behaves the same if you use the Zipkin format? If so, I _think_ you "just" need to take care about the naming of your tags and they should become "process" attributes. I don't know if this is an existing feature or not (should be) or what is the naming convention (if any).

This is where the mapping between OpenTelemetry & Jaeger is defined. 
https://www.jaegertracing.io/docs/1.45/architecture/#terminology
You can see the Jaeger Process tags comes out of OpenTelemetry Resource and Span tags maps from OpenTelemetry attributes.

quaff:
> Do you happen to know if Jaeger behaves the same if you use the Zipkin format? If so, I _think_ you "just" need to take care about the naming of your tags and they should become "process" attributes. I don't know if this is an existing feature or not (should be) or what is the naming convention (if any).

I don't think it related to naming, jaeger have tracer level tag since OpenTracing

https://github.com/jaegertracing/jaeger-client-java/blob/master/jaeger-core/src/main/java/io/jaegertracing/internal/JaegerTracer.java#L554

jonatan-ivanov:
@amithkumarg
>This is where the mapping between OpenTelemetry & Jaeger is defined.

Please check my question again, I was asking about the Zipkin format not OpenTelemetry. OTel is easier to handle for Jaeger since it has the concept of `Resource`s, Zipkin does not.

@quaff
>I don't think it related to naming, jaeger have tracer level tag since OpenTracing

That's fine, I'm asking about Zipkin since it does not have anything just tags on the spans. If Jaeger implemented handling this (I hope they did), you should be able to piggyback on this behavior.

quaff:
> @amithkumarg
> 
> > This is where the mapping between OpenTelemetry & Jaeger is defined.
> 
> Please check my question again, I was asking about the Zipkin format not OpenTelemetry. OTel is easier to handle for Jaeger since it has the concept of `Resource`s, Zipkin does not.
> 
> @quaff
> 
> > I don't think it related to naming, jaeger have tracer level tag since OpenTracing
> 
> That's fine, I'm asking about Zipkin since it does not have anything just tags on the spans. If Jaeger implemented handling this (I hope they did), you should be able to piggyback on this behavior.

Sorry I'm not familiar with Zipkin, but I can verify that Jaeger use `Resource` of `SdkTracerProvider` as process level tag, that's why I'm open this PR. I'm turning to customize `SdkTracerProvider`, just let you know that `ObservationFilter` doesn't satisfy.

PR #35107, opened 2023-04-22
Add service connection from Testcontainers Zipkin

eddumelendez:


dsyer:
It’s already supported in spring-boot-docker-compose, so I vote for this. Maybe you could add prometheus as well?

eddumelendez:
> or maybe prometheus isn't often used as a service endpoint - it scrapes apps

TBH, I'm not sure but given the [prometheus push configuration](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusProperties.java#L115-L125) is already possible in spring-boot it totally makes sense to add support for service connection. 

I would work on this later this week but if anyone interested on contributing before that happen that would be awesome

mhalbritter:
Thanks a lot!

PR #35113, opened 2023-04-22
Remove dependencies in same stream chain without storing in intermediate list

saurabh-rahate:
This PR is to improve the removal of dependencies in `CheckClasspathForUnnecessaryExclusions`.  

It enhances the code to use same stream chain to remove `dependencies` from `exclusions Set<String>` without storing/collecting in any intermediate list.

Also, `Set.removeAll(List)` call can be slow when the size of the argument is greater than or equal to the size of the set, and the set is a subclass of java.util.AbstractSet. In this case, List.contains() is called for each element in the set, which will perform a linear search.


wilkinsona:
Thanks for the suggestion, but I don't think this really improves the code. I find it slightly harder to read and it isn't performance critical.

Issue #32577, opened 2022-10-04
Migrate to common concourse-release-scripts


philwebb:
I think we can remove our own in favor of https://github.com/spring-io/concourse-release-scripts

wilkinsona:
https://github.com/wilkinsona/spring-boot/tree/gh-32577

We need an 0.4.0 release to pick up some enhancements:

- https://github.com/spring-io/concourse-release-scripts/issues/27
- https://github.com/spring-io/concourse-release-scripts/issues/28

PR #35118, opened 2023-04-23
Modifying code in order for maintaining stuff

tuanngocfun:


wilkinsona:
Thanks, @tuanngocfun, We're planning to [replace the scripts with a separately maintained and shared project](https://github.com/spring-projects/spring-boot/issues/32577). Until then, we'd prefer to leave them unchanged as we know that they work.

PR #35119, opened 2023-04-24
Add testcontainers as api dependencies for spring-boot-testcontainers

quaff:
It's quite reasonable IMO.


snicoll:
It's not unreasonable indeed but what do you expect to be able to do with only the API on the classpath?

wilkinsona:
`org.testcontainers:testcontainers` contains everything you need to use `GenericContainer` so it's useful for anything for anything for which Testcontainers doesn't have a specific module. The Redis support that we've got in `spring-boot-test-support` is one example.

quaff:
> It's not unreasonable indeed but what do you expect to be able to do with only the API on the classpath?

It's weird that if I want to use `GenericContainer` I must add testcontainers even if spring-boot-testcontainers present.

PR #35120, opened 2023-04-24
Containers managed by TestcontainersLifecycleBeanPostProcessor are closed twice

quaff:
Startable extends AutoCloseable will always be closed by context


snicoll:
Thanks again @quaff. Sorry, I missed the update you've made in the meantime as I was polishing locally.

PR #35125, opened 2023-04-24
Skipping replace DataSource if has JdbcConnectionDetails not created by DataSourceAutoConfiguration

quaff:
before this commit, `@AutoConfigureTestDatabase(replace = NONE)` is required for
```java
@DataJpaTest
@Testcontainers
//@AutoConfigureTestDatabase(replace = AutoConfigureTestDatabase.Replace.NONE)
public class ApplicationDataJpaTests {

	@Container
	@ServiceConnection
	static MySQLContainer<?> container = ...

```


wilkinsona:
Thanks for the suggestion. Unfortunately, I don't think that we can tell if the `DataSource` should be replaced depending on where the `JdbcConnectionDetails` have been defined. I would prefer that things are slightly more verbose but hopefully also more clear.

Issue #22851, opened 2020-08-10
Update transaction auto-configuration to backoff if a TransactionManager is defined


snicoll:
Spring Framework 5.2 has introduced a parent for the `PlatfromTransactionManager` that's also a parent for the newly introduced `ReactiveTransactionManager`. 

Our auto-configuration is checking the specific type at the moment. If we are auto-configuring a `PlatformTransactionManager` we check for that, and same for `ReactiveTransactionManager` for auto-configuring a reactive transaction manager.

The problem is that we can end up in a situation where we have two `TransactionManager` beans, a `PlatformTransactionManager` and a `ReactiveTransactionManager`. Considering that framework  expects a single bean of that type for unqualified `@Transactional` use, I think we should refine our auto-configuration to avoid that by default.

This affects neo4j in particular as they ship both the imperative and reactive style in the same driver and the presence of Reactor on the classpath is the only trigger (which could be very common and for unrelated reasons, like using `WebClient`).

See also https://github.com/spring-projects/spring-framework/issues/25532 and https://github.com/spring-projects/spring-boot/issues/22692

snicoll:
Spring Boot now backs off in the presence of a `TransactionManager` (vs. `PlatformTransactionManager`). This prevents that both an imperative and reactive transaction managers are auto-configured. See #22692 for a follow-up on the underlying issue.

PR #35261, opened 2023-05-04
Modify `@ConditionalOnMissingBean` of `transactionManager` in `JdbcTransactionManagerConfiguration`

davin111:
In `JdbcTransactionManagerConfiguration`, the `transactionManager` bean is registered if there is no bean of `TransactionManager` types. However, the autoconfiguration is responsible for jdbc transactionmanagers related with `PlatformTransactionManager`, not `ReactiveTransactionManager`. For now, with an existing `ReactiveTransactionManager`, this autoconfiguration doesn't register the expectable transactionmanager. I think the condition of the `transactionManager()` bean method should be modified.

I added a test related with it, and fixed one existing test because it didn't configure `DataSourceAutoConfiguration`, which made the test vacuously true.

* * *

After I made this PR, I checked the issue #22851 which is opposite. It said like below.
> Considering that framework expects a single bean of that type for unqualified `@Transactional` use, I think we should refine our auto-configuration to avoid that by default.

I also agreed with it, but I made this PR because when I develop a spring batch application with jdbc for `JobRepository` and r2dbc for main service and repository logic, I figured out this condition of the autoconfiguration prevents registering the required jdbc transactionmanager. Please give your opinion about this and I can close this PR or modify PR only for fixing the existing test.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceTransactionManagerAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceTransactionManagerAutoConfiguration.java
@@ -45,7 +45,7 @@ import org.springframework.transaction.TransactionManager;
  * @author Stephane Nicoll
  * @author Andy Wilkinson
  * @author Kazuki Shimizu
- * @author davin111
+ * @author Davin Byeon
  * @since 1.0.0
  */
 @AutoConfiguration(before = TransactionAutoConfiguration.class)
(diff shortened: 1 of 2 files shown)


snicoll:
I can confirm that the transaction auto-configuration is doing this on purpose and having more than one transaction managers is not something we'd expect. You shouldn't mix jdbc and R2DBC in the same app for a similar reason.

davin111:
@snicoll Do you think having more than one transaction managers should be avoided, even though `@Transactional` [supports the way to specify](https://github.com/spring-projects/spring-framework/blob/601de66c03ad20fda5f5ec7ab380bdd64095aac2/spring-tx/src/main/java/org/springframework/transaction/annotation/Transactional.java#L143) which transactionManager should be used? I think some apps can have both transactionManagers for some special purposes.

snicoll:
Yes, and you are right that some apps can have multiple transaction managers. These are meant for multi-ressources transaction management, such as a JMS resource and a JDBC resource. Our auto-configuration should promote the standard way of doing things, which is why this PR was declined.

PR #35263, opened 2023-05-04
Reuse rootProject's gradle.properties for buildSrc and allow to override

quaff:
Continue https://github.com/spring-projects/spring-boot/commit/695e58f418b46bc9e28a8c6bd88a641d874921ec

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -14,12 +14,14 @@ sourceCompatibility = 17
 targetCompatibility = 17
 
 ext {
-	def propertiesFile = new File(new File("$projectDir").parentFile, "gradle.properties")
-	propertiesFile.withInputStream {
+	new File("$projectDir/../gradle.properties").withInputStream {
 		def properties = new Properties()
 		properties.load(it)
-		set("kotlinVersion", properties["kotlinVersion"])
-		set("springFrameworkVersion", properties["springFrameworkVersion"])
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the suggestion, @quaff. I prefer the current approach of only copying over certain properties as I think it's less likely to lead to unexpected behavior. There are also some properties in the main `gradle.properties` that it doesn't really make sense to copy over. Filtering those out would be just as verbose if not more verbose than the current approach.

quaff:
> Thanks for the suggestion, @quaff. I prefer the current approach of only copying over certain properties as I think it's less likely to lead to unexpected behavior. There are also some properties in the main `gradle.properties` that it doesn't really make sense to copy over. Filtering those out would be just as verbose if not more verbose than the current approach.

@wilkinsona If you accept this change, I would open another PR.

before:
```java
	def propertiesFile = new File(new File("$projectDir").parentFile, "gradle.properties")
	propertiesFile.withInputStream {
```
after:
```java
	new File("$projectDir/../gradle.properties").withInputStream {
```

wilkinsona:
Personally, I prefer the use of `parentFile`. I'd get rid of the `propertiesFile` variable in https://github.com/spring-projects/spring-boot/commit/3a45d46583ae6a3a8c5fb31d9dfca52c02ff7514 though. Thanks for the idea.

Issue #35332, opened 2023-05-06
@ImportTestcontainers doesn't work


quaff:
According to https://docs.spring.io/spring-boot/docs/3.1.0-RC2/reference/htmlsingle/#features.testing.testcontainers
We could declare `@Container` as static field then use `@ImportTestcontainers`, but the following test fails, am I missed something?
```java
package com.example.demo;

import static org.assertj.core.api.Assertions.assertThat;

import javax.sql.DataSource;

import org.junit.jupiter.api.Test;
import org.springframework.beans.factory.annotation.Autowired;
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.boot.test.context.TestConfiguration;
import org.springframework.boot.testcontainers.context.ImportTestcontainers;
import org.testcontainers.containers.MySQLContainer;
import org.testcontainers.junit.jupiter.Container;

import com.zaxxer.hikari.HikariDataSource;

@SpringBootTest
public class ApplicationTests {

	@Autowired
	private DataSource dataSource;

	@Test
	void test() {
		assertThat(dataSource).isInstanceOfSatisfying(HikariDataSource.class,
				ds -> assertThat(ds.getJdbcUrl()).startsWith("jdbc:mysql://"));
	}

	@TestConfiguration
	@ImportTestcontainers(Containers.class)
	static class Config {

	}

	interface Containers {

		@Container
		MySQLContainer<?> mySQLContainer = new MySQLContainer<>("mysql:5.7");

	}

}
```
```groovy
plugins {
	id 'java'
	id 'org.springframework.boot' version '3.1.0-RC2'
	id 'io.spring.dependency-management' version '1.1.0'
}

group = 'com.example'
version = '0.0.1-SNAPSHOT'
sourceCompatibility = '17'

repositories {
	mavenCentral()
	maven {
		url "https://repo.spring.io/milestone"
	}
}

dependencies {
	implementation 'org.springframework.boot:spring-boot-starter-data-jpa'
	implementation 'org.springframework.boot:spring-boot-starter'
	testImplementation 'org.springframework.boot:spring-boot-starter-test'
	testImplementation 'org.springframework.boot:spring-boot-testcontainers'
	testImplementation 'org.testcontainers:junit-jupiter'
	testImplementation 'org.testcontainers:mysql'
	testRuntimeOnly 'com.h2database:h2'
	testRuntimeOnly 'com.mysql:mysql-connector-j'
}

tasks.named('test') {
[...]

PR #35333, opened 2023-05-06
Add missing @ServiceConnection in importing container declarations example

quaff:
Fix https://github.com/spring-projects/spring-boot/issues/35332


wilkinsona:
Thanks for the PR but I'm not sure that `@ServiceConnection` is missing here. As described in the text that comes before it, the code is intended to be an example of pure Testcontainers code.  The subsequent tip then explains that you can use `@ServiceConnection` or `@DynamicPropertySource`. You don't have to use `@ServiceConnection` so I don't think it really belongs in the earlier pure-Testcontainers example.

quaff:
> Thanks for the PR but I'm not sure that `@ServiceConnection` is missing here. As described in the text that comes before it, the code is intended to be an example of pure Testcontainers code. The subsequent tip then explains that you can use `@ServiceConnection` or `@DynamicPropertySource`. You don't have to use `@ServiceConnection` so I don't think it really belongs in the earlier pure-Testcontainers example.

The typo should be fixed at least, I've created another PR https://github.com/spring-projects/spring-boot/pull/35369

PR #34003, opened 2023-01-27
Add Spring Authorization Server support

sjohnr:
This PR introduces support for Spring Authorization Server. It includes:

* New module `spring-boot-starter-oauth2-authorization-server`
* Dependency management of `org.springframework.security:spring-security-oauth2-authorization-server`
* Support in `spring-boot-autoconfigure` for `org.springframework.security:spring-security-oauth2-authorization-server`

Overview
---

The auto-configuration is designed to closely match the [Getting Started](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/getting-started.html) guide in the reference manual. When `spring-security-oauth2-authorization-server` is detected on the classpath, the following components are optionally registered:

* `RegisteredClientRepository`
* `AuthorizationServerSettings`
* `SecurityFilterChain` for [protocol endpoints](https://docs.spring.io/spring-authorization-server/docs/current/reference/html/protocol-endpoints.html)
* `SecurityFilterChain` for user authentication with Form Login
* `com.nimbusds.jose.jwk.source.JWKSource<SecurityContext>` with a generated RSA key-pair
* `JwtDecoder` that uses the provided `JWKSource`

Because Spring Authorization Server is built on top of Spring Security, the order in which components are registered in collaboration with existing auto-configuration is important.

The main consideration is that `UserDetailsServiceAutoConfiguration` continue to be allowed to publish a `UserDetailsService` if necessary. However, a `JwtDecoder` must also be published afterwards. This is in-contrast with `OAuth2ResourceServerAutoConfiguration` which prevents a `UserDetailsService` from being published.

Configuration
---

The `OAuth2AuthorizationServerProperties` allow configuring `RegisteredClient`s and `AuthorizationServerSettings`. Here is a typical client configuration with default settings:

```yaml
spring:
  security:
    oauth2:
      authorizationserver:
        client:
          messaging-client:
            registration:
              client-id: messaging-client
              client-secret: "{noop}secret"
              client-authentication-methods:
                - client_secret_basic
              authorization-grant-types:
                - authorization_code
                - refresh_token
                - client_credentials
              redirect-uris:
                - http://127.0.0.1:8080/login/oauth2/code/messaging-client-oidc
                - http://127.0.0.1:8080/authorized
              scopes:
                - openid
                - profile
                - message.read
                - message.write
            require-authorization-consent: true
```

Here is a full configuration example:

```yaml
spring:
  security:
    oauth2:
      authorizationserver:
        issuer: https://provider.com
        endpoint:
[...]

PR #35368, opened 2023-05-09
Fix redirect to login page for token requests

sjohnr:
Fixes an issue where auto-configuration for Spring Authorization Server was overriding the default exception handling (`AuthenticationEntryPoint`) resulting in anonymous requests to the token endpoint being redirected to the Spring Security login page instead of returning `401 Unauthorized`.

Auto-configuration now registers a `defaultAuthenticationEntryPointFor` that is added to any other entry points already configured.

See gh-34003


mhalbritter:
Thanks!

sjohnr:
Thanks for the quick merge!

PR #35370, opened 2023-05-10
Fix configuring replica set via PropertiesMongoConnectionDetails

dreis2211:
Hi,

I was reviewing some of the new things in 3.1 and believe I found a typo leading to a bug when setting the `replicaSet` for MongoDB.

Cheers,
Christoph


mhalbritter:
Thank you!

Issue #33184, opened 2022-11-15
Native profile should configure execution in pluginManagement


alexandreroman:
When using Spring Boot Maven plugin 3.0.0-RC2 with a multi-module Maven project, the plugin fails to run at the parent level when packaging is set to `pom`.
It turns out that the `process-aot` target is trying to run, but since there is no source code the plugins fails:

```
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  1.321 s
[INFO] Finished at: 2022-11-15T11:39:28+01:00
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:3.0.0-RC2:process-aot (process-aot) on project wnisb3-otel-parent: Unable to find a suitable main class, please add a 'mainClass' property -> [Help 1]
[ERROR]
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR]
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/MojoExecutionException
```

As a workaround, I set `<skip>true</skip>` when the Maven project packaging is set to `pom`.

snicoll:
It's a bit strange to configure `process-aot` on the root. This should be only applied for the actual application. Can you share why you need to set it at the root?

alexandreroman:
I want to set a default buildpack builder for my multi-module project.
This builder is set at the root level in the `spring-boot-maven-plugin` section:

```
<build>
    <plugins>
        <plugin>
            <groupId>org.springframework.boot</groupId>
            <artifactId>spring-boot-maven-plugin</artifactId>
            <configuration>
                <image>
                    <builder>dashaun/java-native-builder-multiarch:7.37.0</builder>
                </image>
            </configuration>
        </plugin>
    </plugins>
</build>
```

wilkinsona:
I think you should configure that in plugin management so that it only affects modules that need to use Boot's Maven plugin:

```
<build>
    <pluginManagement>
        <plugins>
            <plugin>
                <groupId>org.springframework.boot</groupId>
                <artifactId>spring-boot-maven-plugin</artifactId>
                <configuration>
                    <image>
                        <builder>dashaun/java-native-builder-multiarch:7.37.0</builder>
[...]

PR #35377, opened 2023-05-10
Unlike other Spring Boot goals, process-aot and process-test-aot run on Maven reactor projects

ls-urs-keller:
When running in multi module maven projects, the aot phase should not run on modules with type `POM`

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/AbstractAotMojo.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/AbstractAotMojo.java
@@ -97,6 +97,10 @@ public abstract class AbstractAotMojo extends AbstractDependencyFilterMojo {
 
 	@Override
 	public void execute() throws MojoExecutionException, MojoFailureException {
+		if (this.project.getPackaging().equals("pom")) {
+			getLog().debug("process-*aot goals could not be applied to pom project.");
+			return;
+		}
 		if (this.skip) {
 			getLog().debug("Skipping AOT execution as per configuration");
 			return;


snicoll:
It won't, unless the `spring-boot-maven-plugin` is applied or you have configured an execution in the parent in `plugins` rather than `pluginsManagement`. 

I think this sort of change is actually hiding a build setup issue. See also https://github.com/spring-projects/spring-boot/issues/33184.

Can you please check your build and get back to us with a reason for this change?

ls-urs-keller:
> It won't, unless the `spring-boot-maven-plugin` is applied or you have configured an execution in the parent in `plugins` rather than `pluginsManagement`.

That's the case, but it is also a valid use case assuming a common parent project and N services all similarly configured inheriting dependencies and plugins from the parent.

We do this this way already for the repackage goal: https://github.com/spring-projects/spring-boot/blob/1770f065a5428f059b9659545e4ecbcdf7baa9bd/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/RepackageMojo.java#L198

Otherwise we would have to remove the spring-boot-plugin from `<build><plugins>...` and repeat it in every module.

The suggestion here is to align the behaviour `process-aot` with what's already a reality in `repackage` goal which also skips over reactor modules.

snicoll:
Are all your sub-modules spring boot applications?

> We do this this way already for the repackage goal: 

I am aware of that and I've explained above why I'd like to get a bit more detail about the setup.

ls-urs-keller:
> Are all your sub-modules spring boot applications?
> 
> > We do this this way already for the repackage goal:
> 
> I am aware of that and I've explained above why I'd like to get a bit more detail about the setup.

We also have library projects not spring boot, for those we add a repackage skip =true.
But those tend to be fewer.

Of course, we could do it the other way around adding a skip = false for those that need repackaging/aot-processing.

The process-aot goal behaves inconsistently with respect to the repackaging goal within the same plugin, maybe the solution is to remove the code from repackaging, but that would probably break things.

snicoll:
> Of course, we could do it the other way around adding a skip = false for those that need repackaging/aot-processing.

You could rather configure the execution for the plugin in your parent with whatever settings you need in `pluginManagement` and then add the plugin definition in the project that needs it. This is a better arrangement IMO as the spring boot plugin is only applied where it is necessary rather than applying it in the reactor for all modules and skipping things.

> The process-aot goal behaves inconsistently with respect to the repackaging goal within the same plugin,

`run` does not skip on pom projects. Doing always was an escape hatch for those who were applying a plugin globally in the reactor. It is easy enough for us to do this and help you, but I believe the problem can be avoided in the first place by applying the plugin where it is actually used.

ls-urs-keller:
>And your library projects would not have any reference to Spring Boot at all. This is a better arrangement IMO as the spring boot plugin is only applied where it is necessary rather than applying it in the reactor for all modules and skipping things.

We are using flatten plugins for this, since the libraries we publish need to have a flat pom anyway we configure it to remove the plugins 
But I guess that's opinionated.

> `run` does not skip on pom projects. Checking for pom always was an escape hatch for those who were applying a plugin globally in the reactor.

True, run doesn't skip.

wilkinsona:
@ls-urs-keller Thanks very much for making your first contribution to Spring Boot.

PR #35401, opened 2023-05-11
Spring Boot 2.7.x: Upgrade org.yaml/snakeyaml from 1.30 to 1.33

peterb-personio:


scottfrederick:
Thanks for the PR, but as mentioned in the [pull request template](https://raw.githubusercontent.com/spring-projects/spring-boot/master/.github/PULL_REQUEST_TEMPLATE.md) we have a semi-automated process for dependency upgrades that we prefer to use.

wilkinsona:
I think it's also worth noting that we won't upgrade to a new minor version of SnakeYAML in a maintenance version of Spring Boot. You can use the `snakeyaml.version` property to override the version if you wish.

PR #35424, opened 2023-05-15
Make CurrentTraceContext.Builder a bean

be-hase:
When I did Micrometer Tracing/Observation support on my product, I noticed a part that I want to fix on the spring boot side, so I'm throwing a PR.
( I was wondering if I should start an issue first, but since it is a trivial fix, I will throw a PR directly. Sorry.)

### BraveAutoConfiguration.java

I would like to put CurrentTraceContext.Builder in the bean and use it.

Why: 
Because [the framework(armeria)](https://armeria.dev/docs/advanced-zipkin) I am using, requires the use of `RequestContextCurrentTraceContext` instead of `ThereadLocalCurrentTraceContext`.
This modification will help users who are using something other than the `ThreadLocalCurrentTraceContext`.

For example, it is also useful when you want to use `StrictCurrentTraceContext` for debugging purposes.
https://github.com/openzipkin/brave/blob/f4a2c7f7d0b8c2725ffef99fb4a5ccca222b6492/brave/src/main/java/brave/propagation/StrictCurrentTraceContext.java

### ZipkinConfigurations.ReporterConfiguration

I want to use a `Reporter<Span>` that I created myself(bean configuration).

For some reason, the current `AsyncReporter<Span>` is used, so if I configure the bean by myself, two `Reporter<Span>`  are created.

## ToDo

- [ ] If there is no problem with the modification, add the unit test

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
@@ -64,6 +64,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.boot.context.properties.IncompatibleConfigurationException;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Configuration;
+import org.springframework.context.annotation.Scope;
 import org.springframework.core.Ordered;
 import org.springframework.core.annotation.Order;
 import org.springframework.core.env.Environment;
@@ -130,15 +131,22 @@ public class BraveAutoConfiguration {
 	}
 
 	@Bean
+	@Scope("prototype")
 	@ConditionalOnMissingBean
-	public CurrentTraceContext braveCurrentTraceContext(List<CurrentTraceContext.ScopeDecorator> scopeDecorators,
+	public CurrentTraceContext.Builder braveCurrentTraceContextBuilder() {
+		return ThreadLocalCurrentTraceContext.newBuilder();
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public CurrentTraceContext braveCurrentTraceContext(CurrentTraceContext.Builder currentTraceContextBuilder,
+			List<CurrentTraceContext.ScopeDecorator> scopeDecorators,
 			List<CurrentTraceContextCustomizer> currentTraceContextCustomizers) {
-		ThreadLocalCurrentTraceContext.Builder builder = ThreadLocalCurrentTraceContext.newBuilder();
-		scopeDecorators.forEach(builder::addScopeDecorator);
+		scopeDecorators.forEach(currentTraceContextBuilder::addScopeDecorator);
 		for (CurrentTraceContextCustomizer currentTraceContextCustomizer : currentTraceContextCustomizers) {
-			currentTraceContextCustomizer.customize(builder);
+			currentTraceContextCustomizer.customize(currentTraceContextBuilder);
 		}
-		return builder.build();
+		return currentTraceContextBuilder.build();
 	}
 
 	@Bean
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
@@ -129,7 +129,7 @@ class ZipkinConfigurations {
 		@Bean
 		@ConditionalOnMissingBean
 		@ConditionalOnBean(Sender.class)
-		Reporter<Span> spanReporter(Sender sender, BytesEncoder<Span> encoder) {
+		AsyncReporter<Span> spanReporter(Sender sender, BytesEncoder<Span> encoder) {
 			return AsyncReporter.builder(sender).build(encoder);
 		}
 
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
@@ -129,7 +129,7 @@ class ZipkinConfigurations {
 		@Bean
 		@ConditionalOnMissingBean
 		@ConditionalOnBean(Sender.class)
-		AsyncReporter<Span> spanReporter(Sender sender, BytesEncoder<Span> encoder) {
+		Reporter<Span> spanReporter(Sender sender, BytesEncoder<Span> encoder) {
 			return AsyncReporter.builder(sender).build(encoder);
 		}
 


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -129,7 +129,7 @@ static class ReporterConfiguration {
 		@Bean
 		@ConditionalOnMissingBean
 		@ConditionalOnBean(Sender.class)
-		AsyncReporter<Span> spanReporter(Sender sender, BytesEncoder<Span> encoder) {
+		Reporter<Span> spanReporter(Sender sender, BytesEncoder<Span> encoder) {
```
wilkinsona, line 132: I think this could be considered a bug and, therefore, it should be handled separately. Could you please split this out into a new PR, along with a test that verifies that the auto-configured `spanReporter` bean backs off when you define a custom `Reporter` bean?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
```diff
@@ -130,15 +131,22 @@ public brave.Tracer braveTracer(Tracing tracing) {
...
+	@ConditionalOnMissingBean
+	public CurrentTraceContext braveCurrentTraceContext(CurrentTraceContext.Builder currentTraceContextBuilder,
+			List<CurrentTraceContext.ScopeDecorator> scopeDecorators,
 			List<CurrentTraceContextCustomizer> currentTraceContextCustomizers) {
-		ThreadLocalCurrentTraceContext.Builder builder = ThreadLocalCurrentTraceContext.newBuilder();
-		scopeDecorators.forEach(builder::addScopeDecorator);
+		scopeDecorators.forEach(currentTraceContextBuilder::addScopeDecorator);
 		for (CurrentTraceContextCustomizer currentTraceContextCustomizer : currentTraceContextCustomizers) {
-			currentTraceContextCustomizer.customize(builder);
+			currentTraceContextCustomizer.customize(currentTraceContextBuilder);
```
wilkinsona, line 147: I think that any `CurrentTraceContextCustomizer` beans should be applied to the `braveCurrentTraceContextBuilder` bean rather than only applying them when creating the `CurrentTraceContext`. This will align with other areas where we define a builder bean such as `Jackson2ObjectMapperBuilder` and `RestTemplateBuilder`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
```diff
@@ -130,15 +131,22 @@ public brave.Tracer braveTracer(Tracing tracing) {
 	}
 
 	@Bean
+	@Scope("prototype")
 	@ConditionalOnMissingBean
-	public CurrentTraceContext braveCurrentTraceContext(List<CurrentTraceContext.ScopeDecorator> scopeDecorators,
+	public CurrentTraceContext.Builder braveCurrentTraceContextBuilder() {
```
wilkinsona, line 136: The tests should be updated to assert that this new bean is defined and that it backs off when a custom `CurrentTraceContext.Builder` bean is defined.

wilkinsona (changes requested):

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
be-hase: OK !

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
be-hase: Is this what you intend to do?

```java
@Bean
@Scope("prototype")
@ConditionalOnMissingBean
public CurrentTraceContext.Builder braveCurrentTraceContextBuilder(List<CurrentTraceContextCustomizer> currentTraceContextCustomizers) {
	CurrentTraceContext.Builder builder =ThreadLocalCurrentTraceContext.newBuilder();
	for (CurrentTraceContextCustomizer currentTraceContextCustomizer : currentTraceContextCustomizers) {
		currentTraceContextCustomizer.customize(builder);
	}
	return builder;
}
```

But if we do, we will have to write code to iterate and apply the CurrentTraceContextCustomizers each time you want to change the type of CurrentTraceContext. I think this is inconvenient.

```java
@Bean
@Scope("prototype")
public CurrentTraceContext.Builder braveCurrentTraceContextBuilder(List<CurrentTraceContextCustomizer> currentTraceContextCustomizers) {
	CurrentTraceContext.Builder builder = StrictCurrentTraceContext.newBuilder();
	for (CurrentTraceContextCustomizer currentTraceContextCustomizer : currentTraceContextCustomizers) {
		currentTraceContextCustomizer.customize(builder);
	}
	return builder;
}
```

wilkinsona: Yes, for consistency with how other builders and applying customisers to them, I think we need to do something like this.

Taking a step back, perhaps we would be better offering a configuration property that allows control of the type of builder that is auto-configured. Its value could be thread local, strict, or standard which would map to `ThreadLocalCurrentTraceContext.Builder`, `StrictCurrentTraceContext.Builder` and `CurrentTraceContext.Builder` respectively.

be-hase: Hmmm. How about putting `Supplier<CurrentTraceContext.Builder>` in the bean then?
This would make it easy to change the type of CurrentTraceContext.

```java
@Bean
@ConditionalOnMissingBean
public Supplier<CurrentTraceContext.Builder> braveCurrentTraceContextBuilderSupplier() {
	return ThreadLocalCurrentTraceContext::newBuilder;
}

@Bean
@Scope("prototype")
@ConditionalOnMissingBean
public CurrentTraceContext.Builder braveCurrentTraceContextBuilder(
		Supplier<CurrentTraceContext.Builder> builderSupplier,
		List<CurrentTraceContext.ScopeDecorator> scopeDecorators,
		List<CurrentTraceContextCustomizer> currentTraceContextCustomizers) {
	CurrentTraceContext.Builder builder = builderSupplier.get();
	scopeDecorators.forEach(builder::addScopeDecorator);
	for (CurrentTraceContextCustomizer currentTraceContextCustomizer : currentTraceContextCustomizers) {
		currentTraceContextCustomizer.customize(builder);
	}
	return builder;
}

@Bean
@ConditionalOnMissingBean
public CurrentTraceContext braveCurrentTraceContext(CurrentTraceContext.Builder currentTraceContextBuilder) {
	return currentTraceContextBuilder.build();
}
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
be-hase: submit: https://github.com/spring-projects/spring-boot/pull/35455

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
wilkinsona: Sorry, I don't think that's a good idea. I can't think of anywhere else where we have the indirection of a `Supplier` providing access to a builder. I don't think there's anything that's unique about this situation to justify it and it's likely to cause confusion.

be-hase: I think it is an idea that would make it easier to use different CurrentTraceContexts, but yes, Supplier may be confused.

Hmmm, ok, I'll close this PR, thanks.
Workaround by registering my own bean.

```kotlin
@Bean
fun braveCurrentTraceContext(
    scopeDecorators: List<CurrentTraceContext.ScopeDecorator>,
    currentTraceContextCustomizers: List<CurrentTraceContextCustomizer>,
): CurrentTraceContext {
    // See: https://armeria.dev/docs/advanced-zipkin
    val builder = RequestContextCurrentTraceContext.builder()
    scopeDecorators.forEach(Consumer { scopeDecorator: CurrentTraceContext.ScopeDecorator ->
        builder.addScopeDecorator(
            scopeDecorator
        )
    })
    for (currentTraceContextCustomizer in currentTraceContextCustomizers) {
        currentTraceContextCustomizer.customize(builder)
    }
    return builder.build()
}
```

PR #35447, opened 2023-05-16
Default for spring.flway.execute-in-transaction is not aligned with Flyway's default

TheHound:
FlywayProperties has executeInTransaction default as false, changing the org.flywaydb.core.api.configuration.Configuration#isExecuteInTransaction default. See also org.springframework.boot.autoconfigure.flyway.FlywayAutoConfiguration.FlywayConfiguration#configureProperties

Tested in 3.1.0-RC2.

Prior to change with no `spring.flyway` configuration set 
```
Migrating schema "public" to version "0.0.1.0 - init-tables [non-transactional]"
```


wilkinsona (changes requested):
Thanks for catching this, @TheHound. Could you please also update `defaultValuesAreConsistent` in `FlywayPropertiesTests` to assert that the defaults are aligned?

TheHound:
Updated the commit with extra line in `FlywayPropertiesTests#defaultValuesAreConsistent`

PR #35455, opened 2023-05-17
Defining a zipkin2.reporter.Reporter results in two reporter beans as the auto-configured AsyncReporter does not back off

be-hase:
from this PR: https://github.com/spring-projects/spring-boot/pull/35424

### ZipkinConfigurations.ReporterConfiguration

I want to use a `Reporter<Span>` that I created myself(bean configuration).

For some reason, the current `AsyncReporter<Span>` is used, so if I configure the bean by myself, two `Reporter<Span>`  are created.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurationsReporterConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurationsReporterConfigurationTests.java
@@ -56,10 +56,11 @@ class ZipkinConfigurationsReporterConfigurationTests {
 
 	@Test
 	void shouldBackOffOnCustomBeans() {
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurationsReporterConfigurationTests.java
```diff
@@ -56,10 +57,11 @@ void shouldNotSupplyReporterIfSenderIsMissing() {
 
 	@Test
 	void shouldBackOffOnCustomBeans() {
-		this.contextRunner.withUserConfiguration(CustomConfiguration.class).run((context) -> {
-			assertThat(context).hasBean("customReporter");
-			assertThat(context).hasSingleBean(Reporter.class);
-		});
+		this.contextRunner.withUserConfiguration(SenderConfiguration.class, CustomConfiguration.class)
```
be-hase, line 60: It seems that the test code was not correct because it did not include `SenderConfiguration.class`.

PR #35458, opened 2023-05-17
Remove try-catch or throws BeansException

cyeji:
postProcessBeanFactory() already throws BeansException, it seems that you need to clear the try-catch or clear the throws.


cyeji:
@wilkinsona Thank you for your answer. I have a question. Then already caught all the errors in the iteration, but is there a reason to throw them in the method? I think already caught an error, but I'm wondering why you're throwing an error toward the caller.

wilkinsona:
We're catching and swallowing `NoSuchBeanDefinitionException` as we want them to be ignored and for processing to continue. Any other `BeansException` should abort processing and reach the caller.

PR #35558, opened 2023-05-18
OpenTelemetryAutoConfiguration: allow to override SpanProcessor bean

mmeller-wikia:
Allow to customize only SpanProcessor bean without the need to throw away all `OpenTelemetryAutoConfiguration` value.
This is the only one bean in this class which is not annotated with `@ConditionalOnMissingBean`.


mmeller-wikia:
Thanks for clarification and quick response :)
I was hoping to use `BatchSpanProcessorBuilder.setMeterProvider()` to collect metrics from `BatchSpanProcessor.Worker`
See https://github.com/open-telemetry/opentelemetry-java/blob/0bad3c982ad05f54461e241f8a402536c6f9d4b9/sdk/trace/src/main/java/io/opentelemetry/sdk/trace/export/BatchSpanProcessor.java#L187-L219

I was able to do this by creating `@Primary` `SpanProcessor` bean and redeclaring `SdkTracerProvider` (using only one `SpanProcessor`). One downside is that 2 `SpanProcessor` beans are created

PR #35562, opened 2023-05-18
Remove references to Atomikos and Bitronix

dreis2211:
Hi,

given that both Atomikos and Bitronix support has been removed for a while now, I guess it's fair to remove the remaining references in the metadata.

Cheers,
Christoph


snicoll:
Actually, I am not sure we should. This serves as hints for those upgrading. Atomikos was removed in 3.0 so it's way too soon to remove it. I think we should rather review the error deprecation that we have and decide when to remove it.

dreis2211:
Ah fair. Bitronix was deprecated earlier. Atomikos might be a bit too early then. Whatever you decide here, I'm fine with it.

wilkinsona:
Thanks very much, @dreis2211.

PR #35576, opened 2023-05-21
Polish

izeye:
This PR polishes a bit.


snicoll:
Thanks again, Johnny. I've split the polish in two as I wanted to back port some of it.

PR #35595, opened 2023-05-22
EndpointRequest.toLinks() does not match when management.endpoints.web.base-path is '/'

### LINKED ISSUE #34834: EndpointRequest.toLinks() does not match when management.endpoints.web.base-path is '/'
I am using a different management port with `management.server.base-path: '/management'` and `management.endpoints.web.base-path: '/'`.

When setting up security as `.requestMatchers(EndpointRequest.toLinks()).permitAll()`, then performing  a request over `'/management/'`, returns 403 

Root cause:
`WebEndpointProperties.setBasePath` cleans `'/'` as `''`. 

Later on, upon first `'/'` call, `LinksRequestMatcher` sees the empty basepath and returns `EndpointRequest.EMPTY_MATCHER`, which always evaluates to false

somayaj:
Fix for bug fix #34834. The code now allows for no cleanup of the "/" path to "". 
The rest of the test cases stay as is. 
Added new tests to verify behaviour

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/WebEndpointProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/WebEndpointProperties.java
@@ -67,7 +67,7 @@ public class WebEndpointProperties {
 	}
 
 	private String cleanBasePath(String basePath) {
-		if (StringUtils.hasText(basePath) && basePath.endsWith("/")) {
+		if (StringUtils.hasText(basePath) && basePath.endsWith("/") && StringUtils.cleanPath(basePath).length() != 1) {
 			return basePath.substring(0, basePath.length() - 1);
 		}
 		return basePath;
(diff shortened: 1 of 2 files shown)


somayaj:
Hi @philwebb any eta on when the pr will be looked at?

wilkinsona:
@somayaj No ETA I'm afraid. We'll review it as soon as we have time to do so. Thanks for your patience in the meantime.

somayaj:
Thanks so much

wilkinsona:
Thanks for the PR, @somayaj. Unfortunately, I don't think we should have accepted your offer of a contribution to fix this one as it's deceptively complicated. The matching behavior needs to differ depending on whether or not a separate management port is in use as the use of a different port affects whether or not the links endpoint is available. If we merged these changes and a separate management port was not being used, permitting access to the links endpoint could accidentally permit access to something else. Sorry for the wasted effort here.

somayaj:
> Thanks for the PR, @somayaj. Unfortunately, I don't think we should have accepted your offer of a contribution to fix this one as it's deceptively complicated. The matching behavior needs to differ depending on whether or not a separate management port is in use as the use of a different port affects whether or not the links endpoint is available. If we merged these changes and a separate management port was not being used, permitting access to the links endpoint could accidentally permit access to something else. Sorry for the wasted effort here.

@wilkinsona sure, sounds good. Thanks for the opportunity.

Issue #35096, opened 2023-04-20
Add support for multiple context propagation types for tracing


marcingrzejszczak:
Currently we support only a single context propagation type (e.g. B3 or W3C).

We could consider allowing to pass in a list of propagation types so that you can at least read multiple formats and produce multiple formats out. 

Spring Cloud Sleuth has already that feature implemented.

mhalbritter:
Superseded by #35611.

PR #35611, opened 2023-05-24
Tracing only supports a single context propagation type

mhalbritter:
This change adds support for multiple context propagation types (e.g. B3 and W3C). It also adds support for a new propagation type, `B3_MULTI` (which is B3 with multiple headers instead of the single `b3` header).

There are now two new properties:

* `management.tracing.propagation.consume-types` which controls which trace formats the application will understand on incoming requests. It defaults to all known trace formats (W3C, B3 single, B3 multi).
* `management.tracing.propagation.produce-types` which controls which trace formats the application will add to outgoing requests. It defaults to W3C.

The `management.tracing.propagation.type` has been changed from `PropagationType` to `List<PropagationType>` and the default has been removed. If set, this will override both `management.tracing.propagation.consume-types` and `management.tracing.propagation.produce-types`.

The formats have an ordering, meaning that if one of the extractors matches, the one after it are no longer considered. The default ordering is:

1. W3C
2. B3 single
3. B3 multi

The `CompositePropagationFactory` is the main piece of code for Brave which knows how to extract and inject multiple formats. It's a copy from Sleuth, somewhat refactored.

The `CompositeTextMapPropagator` is the main piece of code for OTel. It's somewhat more complicated than the Brave pendant as we need to support the `BaggageTextMapPropagator` which must run always even if one of the extractors before it has already extracted the context.

I also removed a `@ConditionalOnMissingBean` for type `B3Propagator` and `W3CTraceContextPropagator`. This should be fine as it's not consistent anyway. In the `bagagge = true` case those `@ConditionalOnMissingBean` are not effective but the extractors are added unconditionally in `w3cTextMapPropagatorWithBaggage` and `b3BaggageTextMapPropagator`.

## Property examples

* To only use B3 single, set `management.tracing.propagation.type` to `B3`
* To only use B3 multi, set `management.tracing.propagation.type` to `B3_MULTI`
* To only use W3C multi, set `management.tracing.propagation.type` to `W3C`

* To produce B3 single and W3C and accept W3C, B3 single and multi, set `management.tracing.propagation.produce-types` to `B3,W3C` and `management.tracing.propagation.consume-types` to `B3,B3_MULTI,W3C`

See gh-35096

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
@@ -167,7 +167,8 @@ public class BraveAutoConfiguration {
 		@Bean
 		@ConditionalOnMissingBean
 		Factory propagationFactory(TracingProperties properties) {
-			return CompositePropagationFactory.create(properties.getPropagation().getType());
+			return CompositePropagationFactory.create(properties.getPropagation().getType(),
+					properties.getPropagation().getConsumeTypes());
 		}
 
 	}
@@ -187,7 +188,8 @@ public class BraveAutoConfiguration {
 		BaggagePropagation.FactoryBuilder propagationFactoryBuilder(
 				ObjectProvider<BaggagePropagationCustomizer> baggagePropagationCustomizers) {
 			Factory delegate = CompositePropagationFactory.create(BRAVE_BAGGAGE_MANAGER,
-					this.tracingProperties.getPropagation().getType());
+					this.tracingProperties.getPropagation().getType(),
+					this.tracingProperties.getPropagation().getConsumeTypes());
 			FactoryBuilder builder = BaggagePropagation.newFactoryBuilder(delegate);
 			baggagePropagationCustomizers.orderedStream().forEach((customizer) -> customizer.customize(builder));
 			return builder;
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
@@ -16,7 +16,6 @@
 
 package org.springframework.boot.actuate.autoconfigure.tracing;
 
-import java.util.Collections;
 import java.util.List;
 
 import brave.CurrentSpanCustomizer;
@@ -35,7 +34,6 @@ import brave.baggage.CorrelationScopeCustomizer;
 import brave.baggage.CorrelationScopeDecorator;
 import brave.context.slf4j.MDCScopeDecorator;
 import brave.handler.SpanHandler;
-import brave.propagation.B3Propagation;
 import brave.propagation.CurrentTraceContext;
 import brave.propagation.CurrentTraceContext.ScopeDecorator;
 import brave.propagation.CurrentTraceContextCustomizer;
@@ -48,7 +46,6 @@ import io.micrometer.tracing.brave.bridge.BravePropagator;
 import io.micrometer.tracing.brave.bridge.BraveSpanCustomizer;
 import io.micrometer.tracing.brave.bridge.BraveTracer;
 import io.micrometer.tracing.brave.bridge.CompositeSpanHandler;
-import io.micrometer.tracing.brave.bridge.W3CPropagation;
 import io.micrometer.tracing.exporter.SpanExportingPredicate;
 import io.micrometer.tracing.exporter.SpanFilter;
 import io.micrometer.tracing.exporter.SpanReporter;
@@ -169,12 +166,8 @@ public class BraveAutoConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
-		Factory propagationFactory(TracingProperties tracing) {
-			return switch (tracing.getPropagation().getType()) {
-				case B3 ->
-					B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.SINGLE_NO_PARENT).build();
-				case W3C -> new W3CPropagation();
-			};
+		Factory propagationFactory(TracingProperties properties) {
+			return CompositePropagationFactory.create(properties.getPropagation().getType());
 		}
 
 	}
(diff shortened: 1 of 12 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
```diff
@@ -252,4 +255,59 @@ ScopeDecorator correlationScopeDecorator(CorrelationScopeDecorator.Builder build
...
+
+		private PropagationFactoryFactory() {
+		}
+
+		/**
+		 * Creates a new B3 propagation factory.
+		 * @return B3 propagation factory.
+		 */
+		private static Factory b3() {
+			return B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.SINGLE_NO_PARENT).build();
```
jonatan-ivanov, line 271: Just a note for the docs: I would call it out that W3C and B3 single is used by default, if users want B3 multi instead (or both?), they need to create their own `CompositePropagationFactory` instance and `BaggagePropagation.FactoryBuilder` bean.

shakuzen: I wonder if it wouldn't be better for compatibility to check for B3 multi but check it last (so it only happens when other, higher precedent formats aren't present). Unless the concern is about the added cost of trying to extract b3 multi when no trace context is present. Perhaps then we could measure the cost. What do you think?

jonatan-ivanov: For compatibility, it would be definitely better. From cost perspective and trying to move away from B3 multi, it would not. I _assume_ the cost would be minimal since this should be just a map lookup.
If we do this should not we also support it on the output (i.e.: `B3_SINGE` and `B3_MULTI`)?

shakuzen: Yeah, probably. I think `B3` for single and `B3_MULTI` make sense.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
```diff
@@ -169,12 +173,11 @@ static class BraveNoBaggageConfiguration {
...
-		Factory propagationFactory(TracingProperties tracing) {
-			return switch (tracing.getPropagation().getType()) {
-				case B3 ->
-					B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.SINGLE_NO_PARENT).build();
-				case W3C -> new W3CPropagation();
-			};
+		Factory propagationFactory(TracingProperties properties) {
+			Iterable<Factory> injectorFactories = PropagationFactoryFactory
+				.factoriesFor(properties.getPropagation().getType());
+			Iterable<Factory> extractorFactories = PropagationFactoryFactory.factoriesFor(PropagationType.values());
```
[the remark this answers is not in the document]
mhalbritter: Not sure, I could make that more explicit. What order would be sensible?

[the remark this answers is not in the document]
mhalbritter: I'll make this more explicit. Our extraction order will be:

1. `W3C` (as this is the format we produce by default)
2. `B3` (single)
3. `B3_MULTI`

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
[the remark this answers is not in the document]
mhalbritter: FTR: https://github.com/spring-projects/spring-boot/issues/35674

mhalbritter:
It works for both Brave and Otel now, build is green too.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/TracingProperties.java
```diff
@@ -166,29 +168,54 @@ public void setFields(List<String> fields) {
...
+
+			PropagationType(int order) {
+				this.order = order;
+			}
+
+			/**
+			 * Returns the ordered values.
+			 * @return the ordered values
+			 */
+			static List<PropagationType> orderedValues() {
```
wilkinsona, line 216: Enums are already ordered. Could we change the order in which the values are declared instead?

mhalbritter: Sure! I preferred the explicit ordering here, but that's no strong opinion on my side.

jonatan-ivanov: I think I would go without having `order` on this enum too and add a line to the javadocs that the order these instances are defined matters.

jonatan-ivanov (approved):

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/TracingProperties.java
mhalbritter: Changed it.

shakuzen (approved):

jonatan-ivanov (approved):

mhalbritter:
When merging into 3.1.x, `shouldSupportJoinedSpansIfB3UsedAndBackendSupportsIt` fails. That's because it expects that span joining is supported when using B3. However, as we now extract all formats from the incoming request, `CompositePropagationFactory.supportsJoin` always returns `false`, as there is a `W3CPropagation` in the extractor list and `supportJoin` is implemented like this:

```java
Stream.concat(injectorFactories.stream(), extractorFactories.stream()).allMatch(Factory::supportsJoin);
```

Is that just an unfortunate consequence because we now apply all extractors, or is the span joining feature just for injecting and the `supportJoin` should be changed to only work on the `injectorFactories`?

@jonatan-ivanov As you implemented the test, maybe you have an idea?

jonatan-ivanov:
Yeah, if we consume all formats by default `supportsJoin` won't work, I completely missed this when we discussed this issue, I'm sorry.
I think letting the users define the consumed formats would work. I'm also thinking if we should set it if `supportsJoin=true`. That might be a little shady side effect though.

mhalbritter:
No worries Jonatan. I've now added properties to control which formats we produce, which we accept and the `management.tracing.propagation.type` controls both at the same time.

`management.tracing.propagation.type` no longer has a default. If set, it overrides `management.tracing.propagation.consume-types` and `management.tracing.propagation.produce-types`.

With that in place, we should have enough flexibility in our tracing setup. And users still have one property `management.tracing.propagation.type` if they aren't interested in configuring the low-level details.

jonatan-ivanov (approved):

wilkinsona:
> `management.tracing.propagation.consume-types` and `management.tracing.propagation.produce-types`

Alternative names for these properties to consider:

- `management.tracing.propagation.consume` and `management.tracing.propagation.produce`
- `management.tracing.propagation.types.consume` and `management.tracing.propagation.types.produce`

PR #35619, opened 2023-05-24
Auto-configured ExemplarSampler bean only backs off when a DefaultExemplarSampler is defined

johnnywiller:
Having Spring Actuator autoconfiguring a `DefaultExemplarSampler` makes it harder for the user to define its own bean of type `ExemplarSampler`, since `@ConditionalOnMissingBean` will still create a `DefaultExemplarSampler` causing two beans of type `ExemplarSampler` to be created.

This PR simply points to the interface instead.

Another option would be to explicitly add `@ConditionalOnMissingBean(ExemplarSampler.class)` but seems less correct.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
@@ -92,7 +92,7 @@ public class PrometheusMetricsExportAutoConfiguration {
 	@Bean
 	@ConditionalOnMissingBean
 	@ConditionalOnBean(SpanContextSupplier.class)
-	public DefaultExemplarSampler exemplarSampler(SpanContextSupplier spanContextSupplier) {
+	public ExemplarSampler exemplarSampler(SpanContextSupplier spanContextSupplier) {
 		return new DefaultExemplarSampler(spanContextSupplier);
 	}
 
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
@@ -90,9 +90,9 @@ public class PrometheusMetricsExportAutoConfiguration {
 	}
 
 	@Bean
-	@ConditionalOnMissingBean
+	@ConditionalOnMissingBean(ExemplarSampler.class)
 	@ConditionalOnBean(SpanContextSupplier.class)
-	public ExemplarSampler exemplarSampler(SpanContextSupplier spanContextSupplier) {
+	public DefaultExemplarSampler exemplarSampler(SpanContextSupplier spanContextSupplier) {
 		return new DefaultExemplarSampler(spanContextSupplier);
 	}
 
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
@@ -20,6 +20,7 @@ import io.micrometer.core.instrument.Clock;
 import io.micrometer.prometheus.PrometheusConfig;
 import io.micrometer.prometheus.PrometheusMeterRegistry;
 import io.prometheus.client.CollectorRegistry;
+import io.prometheus.client.exemplars.DefaultExemplarSampler;
 import io.prometheus.client.exemplars.ExemplarSampler;
 import io.prometheus.client.exemplars.tracer.common.SpanContextSupplier;
 import io.prometheus.client.exporter.BasicAuthHttpConnectionFactory;
@@ -45,6 +46,7 @@ import org.springframework.context.annotation.Import;
 import org.springframework.test.util.ReflectionTestUtils;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.Mockito.mock;
 
 /**
  * Tests for {@link PrometheusMetricsExportAutoConfiguration}.
@@ -62,6 +64,12 @@ class PrometheusMetricsExportAutoConfigurationTests {
 	@Test
 	void backsOffWithoutAClock() {
 		this.contextRunner.run((context) -> assertThat(context).doesNotHaveBean(PrometheusMeterRegistry.class));
+	}
+
+	@Test
+	void backsOfWhenExemplarSamplerIsPresent() {
+		this.contextRunner.withUserConfiguration(ExemplarsConfiguration.class)
+				.run((context) -> assertThat(context).doesNotHaveBean(DefaultExemplarSampler.class));
 	}
 
 	@Test
@@ -311,6 +319,10 @@ class PrometheusMetricsExportAutoConfigurationTests {
 			};
 		}
 
+		@Bean
+		ExemplarSampler exemplarSampler2() {
+			return mock(ExemplarSampler.class);
+		}
 	}
 
 }
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
@@ -69,7 +69,7 @@ class PrometheusMetricsExportAutoConfigurationTests {
 	@Test
 	void backsOfWhenExemplarSamplerIsPresent() {
 		this.contextRunner.withUserConfiguration(ExemplarsConfiguration.class)
-				.run((context) -> assertThat(context).doesNotHaveBean(DefaultExemplarSampler.class));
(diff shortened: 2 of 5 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfiguration.java
```diff
@@ -92,7 +92,7 @@ public CollectorRegistry collectorRegistry() {
 	@Bean
 	@ConditionalOnMissingBean
 	@ConditionalOnBean(SpanContextSupplier.class)
-	public DefaultExemplarSampler exemplarSampler(SpanContextSupplier spanContextSupplier) {
```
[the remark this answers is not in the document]
philwebb: When Spring Framework creates bean definitions from `@Configuration` classes it only has the method signature to go on. It won't actually invoke the method and get the real bean instance until much later.

For that reason, it's always best to use the most specific type in the method signature. That way Spring has as much information as possible.

wilkinsona (approved):

wilkinsona:
We should add a test for the backing-off behavior as part of merging this.

johnnywiller:
A few considerations on the backsOff test:

- Needed to split `ExemplarsConfiguration` and `SpanConfiguration`, otherwise the test `autoConfiguresExemplarSamplerIfSpanContextSupplierIsPresent` would be falsely asserting the bean defined on the test configuration (`.hasSingleBean(ExemplarSampler.class)`)
- `backsOfWhenExemplarSamplerIsPresent` is asserting on the absence of `DefaultExemplarSampler` since a bean of type `ExemplarSampler` is indeed present (from `ExemplarsConfiguration`)

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
```diff
@@ -313,4 +321,15 @@ public boolean isSampled() {
 
 	}
 
+	@Configuration(proxyBeanMethods = false)
+	@Import(BaseConfiguration.class)
+	static class ExemplarsConfiguration {
```
johnnywiller, line 326: Needed to split `ExemplarsConfiguration` and `SpanConfiguration`, otherwise the test `autoConfiguresExemplarSamplerIfSpanContextSupplierIsPresent` would be falsely asserting (with `.hasSingleBean(ExemplarSampler.class)`) the bean defined here

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
```diff
@@ -64,6 +66,12 @@ void backsOffWithoutAClock() {
 		this.contextRunner.run((context) -> assertThat(context).doesNotHaveBean(PrometheusMeterRegistry.class));
 	}
 
+	@Test
+	void backsOfWhenExemplarSamplerIsPresent() {
+		this.contextRunner.withUserConfiguration(ExemplarsConfiguration.class)
+				.run((context) -> assertThat(context).doesNotHaveBean(DefaultExemplarSampler.class));
```
johnnywiller, line 72: asserting on the absence of `DefaultExemplarSampler` since a bean of type `ExemplarSampler` is present from `ExemplarsConfiguration`

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
```diff
@@ -287,7 +295,7 @@ PrometheusScrapeEndpoint customEndpoint(CollectorRegistry collectorRegistry) {
 
 	@Configuration(proxyBeanMethods = false)
 	@Import(BaseConfiguration.class)
-	static class ExemplarsConfiguration {
+	static class SpanConfiguration {
```
jonatan-ivanov, line 298: Would it make sense calling this `DefaultExemplarSamplerConfiguration` and the other one `CustomExemplarSamplerConfiguration`?

jonatan-ivanov:
Out of curiosity (I think this is a good change that we should merge in): what is your use case for having a custom exemplar sampler?

johnnywiller:
> We should add a test for the backing-off behavior as part of merging this.



> Out of curiosity (I think this is a good change that we should merge in): what is your use case for having a custom exemplar sampler?
We are decorating the bean `DefaultExemplarSampler` to apply custom tagging to the spans. The custom tags will be something like `"exemplar": "sampled"` and serve to inform Otel Collector that it should not drop those spans.
We send metrics directly to Prometheus and Spans/Traces to Otel Collector, so if a metric contains exemplar, the referred span should not be dropped in otel collector.

The decorator looks like 

```
@Bean
    @ConditionalOnBean({SpanContextSupplier.class, Tracer.class})
    @ConditionalOnMissingBean
    DefaultExemplarSampler exemplarMarkingExemplarSampler(SpanContextSupplier spanContextSupplier,
                                                          Tracer tracer) {
        var defaultExemplarSampler = new DefaultExemplarSampler(spanContextSupplier);
        /*
         we are returning a DefaultExemplarSampler instead of ExemplarSampler
         interface because actuator autoconfiguration is doing ConditionalOnBean directly on the
         DefaultExemplarSampler. 
         The ExemplarMarkingExemplarSampler decorator should be changed
         to implement ExemplarSampler once Spring is updated
        */
        return new ExemplarMarkingExemplarSampler(spanContextSupplier, defaultExemplarSampler, tracer);
    }
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusMetricsExportAutoConfigurationTests.java
johnnywiller: Agree with you even if imo `DefaultExemplarSamplerConfiguration` expresses more what the configuration will be used for instead of what it actually configures, but the latter seems less important in this test context.
Agree also the use of terminology `Custom*` is harmonious with the style of this test class.
Changed

jonatan-ivanov:
I'm thinking, if prometheus sampler/supplier and/or Boot should provide this functionality.

I have another quick question: I did not check this so I'm not sure what can go wrong with it but wouldn't this be possible in the `SpanContextSupplier` where you have access to the `Span` anyways? So instead of having a `ExemplarMarkingExemplarSampler`, you would use the `DefaultSampler` but instead of the `SpanContextSupplier` that Boot creates (and that one backs off), you can implement a `ExemplarMarkingSpanContextSupplier` (doing the same just one level deeper, I guess more conveniently)?

johnnywiller:
> I'm thinking, if prometheus sampler/supplier and/or Boot should provide this functionality.

You mean, providing a extension point to add custom tags to the spans, or directly tag a span as exemplar? 
The latter seems a bit complicated as I think there is no standard tag for exemplars yet... And for the former, would be cool but not sure how to design such a thing.

> wouldn't this be possible in the SpanContextSupplier where you have access to the Span anyways?

We just want to tag spans when they are being used as exemplars on metrics, not for all spans.
I may be wrong but I believe `SpanContextSupplier` is a more generic bean that can be called whenever there is a need for getting a span, which may or may not be a exemplar situation.
As seem in `DefaultExemplarSampler#doSample`, spans are only exemplified when they meet the condition:
```
if ((previous == null || previous.getTimestampMs() == null
        || timestampMs - previous.getTimestampMs() > minRetentionIntervalMs)
        && spanContextSupplier.isSampled()) {
```
So to me, if feels more solid to add the tagging behaviour close to `ExemplarSampler`. Here's how we are doing it (partially omitted for brevity):

```
class ExemplarMarkingExemplarSampler extends DefaultExemplarSampler {
    private final ExemplarSampler delegate;
    private final Tracer tracer;

    @Override
    public Exemplar sample(double increment, Exemplar previous) {
        Exemplar exemplar = delegate.sample(increment, previous);
        if (exemplar != null) {
            markSpanAsExemplar();
        }
        return exemplar;
    }

    private void markSpanAsExemplar() {
        var span = tracer.currentSpan();
        if (span != null) {
            span.tag("exemplar", "sampled");
        }
    }
}
```

I forgot to mention in the use case description, the otel collector has a TAIL SAMPLING rule to avoid dropping spans with the `"exemplar"` tag, so the whole purpose is for tail sampling.

jonatan-ivanov:
>You mean, providing a extension point to add custom tags to the spans, or directly tag a span as exemplar?
The latter seems a bit complicated as I think there is no standard tag for exemplars yet... And for the former, would be cool but not sure how to design such a thing.

We can let you inject a `Supplier<Span>` span customizer so that you can modify your span when it is sampled.

>We just want to tag spans when they are being used as exemplars on metrics, not for all spans.
I may be wrong but I believe SpanContextSupplier is a more generic bean that can be called whenever there is a need for getting a span, which may or may not be a exemplar situation.

`SpanContextSupplier` should be only used when an exemplar is sampled, it is also in the "exemplars" package: `io.prometheus.client.exemplars.tracer.common`, why do you think it is used elsewhere and what is it used for if so (I can tell that in Micrometer and Boot it is only used to sample exemplars)?

>As seem in DefaultExemplarSampler#doSample, spans are only exemplified when they meet the condition:...

I'm not sure what is the relevance of this, `spanContextSupplier.getTraceId()` is called in this `if` body, you modify the span for the same condition so this works as it should be, doesn't it?

>So to me, if feels more solid to add the tagging behaviour close to ExemplarSampler. Here's how we are doing it (partially omitted for brevity):

The only thing I can think of right now is doing this in a getter (`getTraceId`) is not very elegant but getting the span twice (in the sampler) isn't either. I consider both kind of a hack until Prometheus will support this. One hack though is simpler to do because Boot backs off. Don't get me wrong, I still think you change is valid and we should merge it.

Let's continue this discussion here: https://github.com/prometheus/client_java/issues/843, if we can find a good general solution, I think we can add support for this in Micrometer and Boot.

snicoll:
@johnnywiller thank you for making your first contribution to Spring Boot.

PR #35620, opened 2023-05-24
Add DataSourceBuilderCustomizer

eduanb:
This PR introduces a DataSourceBuilderCustomizer. This allows customizing only parts of the DataSource without creating a full new DataSource. This is the JDBC equivalent to R2DBC's ConnectionFactoryOptionsBuilderCustomizer.

My real world use case is connecting to AWS Aurora using IAM. The AWS SDK can generate a password with the GenerateAuthenticationTokenRequest function.
https://sdk.amazonaws.com/java/api/latest/software/amazon/awssdk/services/rds/RdsUtilities.html#generateAuthenticationToken(software.amazon.awssdk.services.rds.model.GenerateAuthenticationTokenRequest


vpavic:
> My real world use case is **connecting to AWS RDS using IAM**. The AWS SDK can generate a password with the GenerateAuthenticationTokenRequest function.

Are you aware of [AWS JDBC Driver](https://github.com/awslabs/aws-advanced-jdbc-wrapper) which has support for exactly this use case?

PR #35631, opened 2023-05-25
refactor: Remove unnecessary local variables

birariro:
This PR is neither a bug nor a feature add-on
Removed unnecessary local variable 'portMappings'.


wilkinsona:
Thanks for the proposal, @birariro. The current code is more concise and we find it easier to read than the stream-based approach that you have proposed. Thanks anyway but we're going to decline this one.

PR #35637, opened 2023-05-25
refactor: Change loop code

birariro:
method was using stream and foreach together.
We changed foreach to stream

Also, using stream's toList() eliminates the need to use Collections.unmodifiableList().


philwebb:
Thanks, but I think I prefer the code as it is.

PR #35662, opened 2023-05-27
Move property notes up to external configuration section

twobiers:
This change moves the already present note about relaxed property binding which affects naming of environment variables upwards to a more appropriate position.
The reason for the change is that it's hard to find information about the relaxed binding if you're reading the externalized configuration paragraph. In my opinion, there's no reason to hide the note under a relatively unrelated paragraph and it should be mentioned straight at the same place where environment variables as a property source are being introduced. At least that's where I would search for such a note (e.g. If debugging why the binding doesn't work).


twobiers:
Oh yes, that makes sense. Did that aswell.

mhalbritter:
Thanks!

Issue #35664, opened 2023-05-28
'springProperty' not loaded after migration from 2.7 to 3.x


geminiKim:
### [Used Version]
3.1 (migrating from boot2.7 to 3.1)

### [Problem]
The `springProperty` of the existing logback setting does not load the variable correctly.
(and after the error below, the server will not run and will shut down.)

```xml
// logback-spring.xml
<?xml version="1.0" encoding="UTF-8"?>
<configuration>
    <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
    <springProperty scope="context" name="TEST_URL" source="test.url" />

    <include url="${TEST_URL}/logback/logback-remote.xml" />

    <appender name="STDOUT" class="ch.qos.logback.core.ConsoleAppender">
        <encoder>
            <pattern>%m%n</pattern>
            <charset>utf8</charset>
        </encoder>
    </appender>

    <root level="INFO">
        <appender-ref ref="STDOUT"/>
    </root>
</configuration>

```

I'm actually using a more complicated setup, but I've created an example where the problem occurs.
In an empty project, the above setup works fine with boot2.7, but not with 3.x.

The error message itself says it's a formatting issue, but we can see that the springProperty above doesn't work correctly, causing `TEST_URL_IS_UNDEFINED`.

I'm not sure if this is the right issue to raise in spring-boot. (maybe I should go to logback.)
However, I'm reporting it because I found it to be a backwards compatibility breaker in my actual project.

If this is not a bug and there is a correct way to use it, I would appreciate some guidance.
Thanks.

### [Stacktrace]
```
ERROR in ch.qos.logback.core.joran.action.IncludeAction - URL [TEST_URL_IS_UNDEFINED/logback/logback-remote.xml] is not well formed. java.net.MalformedURLException: no protocol: TEST_URL_IS_UNDEFINED/logback/logback-remote.xml
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.reportConfigurationErrorsIfNecessary(LogbackLoggingSystem.java:260)
	at org.springframework.boot.logging.logback.LogbackLoggingSystem.loadConfiguration(LogbackLoggingSystem.java:247)
	at org.springframework.boot.logging.AbstractLoggingSystem.initializeWithConventions(AbstractLoggingSystem.java:80)
	at org.springframework.boot.logging.AbstractLoggingSystem.initialize(AbstractLoggingSystem.java:60)
[...]

PR #35670, opened 2023-05-29
Fix issue with 'springProperty' not loaded from context scope when initializing with 'playSaxEvents' in logback

geminiKim:
related issue : https://github.com/spring-projects/spring-boot/issues/35664

I fixed an issue that worked in boot 2.7 but has been happening since boot 3.x.

I think this issue was caused by the deletion of this line in this [commit](https://github.com/spring-projects/spring-boot/commit/0bfa9cd7049b70134f3e352db70a55af8f31458e#diff-ebdfb5ded3ac62a268b27edb59cd910349ffdd305f65cb133ce844ea452d7e54L58).
(And while I'm not entirely certain, I think this might also be due to the effects of updating logback to version 1.4.x)

### [Causes and actions]
Looking at the `doConfigure(...)` function in `ch.qos.logback.core.joranGenericXMLConfigurator.java`, it works like this. 

```
        Model top = buildModelFromSaxEventList(recorder.getSaxEventList());
        if (top == null) {
            addError(ErrorCodes.EMPTY_MODEL_STACK);
            return;
        }
        sanityCheck(top);
        processModel(top);
```

In the current situation, the `processModel` is lagging behind, and `buildModelFromSaxEventList(...)` isn't loading the `springProperty`.

Hence, I've injected `springProperty` into the `context` using a `SpringPropertyAction`, which is a part of the flow executed by `playSaxEvents(...)`.

(Referring to the previously mentioned commit, it appears that initialization was done via `SpringPropertyAction` in the past.)

### Additional comments 

I haven't given much thought to this revision, as I'm unaware of the historical context.
However, this issue has prevented me from upgrading the current project to boot 3.x. 
It would be truly beneficial if this issue could be resolved.

I'm always using Spring-Boot well. Thank you for your efforts.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/SpringPropertyAction.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/SpringPropertyAction.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.logging.logback;
 
+import ch.qos.logback.core.joran.action.ActionUtil;
 import ch.qos.logback.core.joran.action.BaseModelAction;
 import ch.qos.logback.core.joran.spi.SaxEventInterpretationContext;
 import ch.qos.logback.core.model.Model;
@@ -46,6 +47,10 @@ class SpringPropertyAction extends BaseModelAction {
 		model.setSource(attributes.getValue(SOURCE_ATTRIBUTE));
 		model.setScope(attributes.getValue(SCOPE_ATTRIBUTE));
 		model.setDefaultValue(attributes.getValue(DEFAULT_VALUE_ATTRIBUTE));
+
+		if (ActionUtil.stringToScope(model.getScope()) == ActionUtil.Scope.CONTEXT) {
+			interpretationContext.getContext().putProperty(model.getName(), model.getSourceOrDefaultValue());
+		}
 		return model;
 	}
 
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/SpringPropertyModel.java
```diff
@@ -58,4 +58,11 @@ void setSource(String source) {
 		this.source = source;
 	}
 
+	String getSourceOrDefaultValue() {
+		if (this.source == null) {
+			return this.defaultValue;
+		}
+		return this.source;
+	}
```
geminiKim, line 66: I'm not certain if it's appropriate to find the default values here.
However, since the logic was handled within the object, I made sure this class took responsibility for it.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/SpringPropertyAction.java
```diff
@@ -46,6 +47,10 @@ protected Model buildCurrentModel(SaxEventInterpretationContext interpretationCo
 		model.setSource(attributes.getValue(SOURCE_ATTRIBUTE));
 		model.setScope(attributes.getValue(SCOPE_ATTRIBUTE));
 		model.setDefaultValue(attributes.getValue(DEFAULT_VALUE_ATTRIBUTE));
+
+		if (ActionUtil.stringToScope(model.getScope()) == ActionUtil.Scope.CONTEXT) {
+			interpretationContext.getContext().putProperty(model.getName(), model.getSourceOrDefaultValue());
```
geminiKim, line 52: I wanted to use an appropriate Util Class in `logback`, but I couldn't find it.
`ActionUtil, ModelUtil` were not applicable in this situation.

philwebb:
Thanks for the PR, but I don't think we can accept this change due to the reason documented in https://github.com/spring-projects/spring-boot/issues/35664#issuecomment-1571092176.

Issue #35680, opened 2023-05-31
Mongo connection ignores authenticationDatabase if database is set, but both are valid at the same time


mikebski:
I need a mongodb connection string like this:

`spring.data.mongodb.uri=mongodb://localhost:27017/itr-devel?authSource=admin`

I set these properties in the config expecting it to work:

```
spring.data.mongodb.host=localhost
spring.data.mongodb.port=27017
spring.data.mongodb.database=itr-devel
spring.data.mongodb.username=root
spring.data.mongodb.password=Welcome^1
spring.data.mongodb.authentication-database=admin
```

The authentication-database is ignored if the database value is set.

With the above properties, I should get this URI: `mongodb://localhost:27017/itr-devel?authSource=admin`

wilkinsona:
Duplicates #35567.

Issue #35567, opened 2023-05-19
MongoDB authentication to different DB than used no longer works in spring boot 3.1


bpfoster:
After upgrading from Spring Boot 3.0 to 3.1 today, MongoDB authentication via the autoconfiguration is broken.

If you configure both a database to use, and a separate database to authenticate against, [the authentication database is ignored](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/PropertiesMongoConnectionDetails.java#L61), and authentication is attempted against the to-be-used database.

In versions past, and the intended behavior is, that specifying the authentication database allows authenticating as the user to that DB and then switching over to the configured database for normal operations.

Perhaps the `PropertiesMongoConnectionDetails` should append the `authSource` [option](https://www.mongodb.com/docs/manual/reference/connection-string/#mongodb-urioption-urioption.authSource) to the connection string if `authenticationDatabse` is present?

ps-tunnelsoft:
@scottfrederick can you have a look at a similar issue of spring-boot-docker-compose MongoDB autoconfiguration? 

The StringBuilder in class [MongoDockerComposeConnectionDetailsFactory](https://github.com/spring-projects/spring-boot/blob/ce85290d01a4d425936c173cc63d565aeef8aabd/spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/service/connection/mongo/MongoDockerComposeConnectionDetailsFactory.java#LL63C4-L75C93)  is also missing the authSource option. But I am not sure, where to take this value from. 

MONGO_INITDB_ROOT_USERNAME and MONGO_INITDB_ROOT_PASSWORD are used to create a root user by default in admin database and not MONGO_INITDB_DATABASE.

scottfrederick:
@ps-tunnelsoft From reading the [documentation](https://hub.docker.com/_/mongo) for the official MongoDB image, I think you're right. 

> `MONGO_INITDB_ROOT_USERNAME`, `MONGO_INITDB_ROOT_PASSWORD`
These variables, used in conjunction, create a new user and set that user's password. This user is created in the `admin` authentication database

We might need to set the authentication database to the value `admin` if the username and password environment variables are configured. Can you create a new issue so we can investigate further?

PR #35681, opened 2023-05-31
Fixes https://github.com/spring-projects/spring-boot/issues/35680 - S…

mikebski:
…upport both database and authenticationDatabase options


wilkinsona:
Thanks for the PR, @mikebski. Unfortunately, as indicated by the assignment of #35567, @scottfrederick is already working on this one.

Issue #35628, opened 2023-05-25
Improve error message when accidentally specifying multiple hosts in spring.rabbitmq.host


francis-a:
When using YAML configuration with the Spring Boot AMQP starter in version 3.0.x it was possible to omit the `addresses` configuration property and instead use `host` and `port`.

For example:

```
spring:
  rabbitmq:
    host: "my-rmq-host.net"
    username: "host_username"
    password: "host_password"
    port: 5672
```

After upgrading to Spring Boot 3.1.0 this configuration option no longer works. The service fails to start with a `java.lang.ArrayIndexOutOfBoundsException`. 

```
Caused by: java.lang.ArrayIndexOutOfBoundsException: Index 1 out of bounds for length 1
	at org.springframework.boot.autoconfigure.amqp.PropertiesRabbitConnectionDetails.getAddresses(PropertiesRabbitConnectionDetails.java:57) ~[spring-boot-autoconfigure-3.1.0.jar:3.1.0]
	at org.springframework.boot.autoconfigure.amqp.RabbitConnectionDetails.getFirstAddress(RabbitConnectionDetails.java:71) ~[spring-boot-autoconfigure-3.1.0.jar:3.1.0]
	at org.springframework.boot.autoconfigure.amqp.RabbitConnectionFactoryBeanConfigurer.configure(RabbitConnectionFactoryBeanConfigurer.java:98) ~[spring-boot-autoconfigure-3.1.0.jar:3.1.0]
	at org.springframework.boot.autoconfigure.amqp.RabbitAutoConfiguration$RabbitConnectionFactoryCreator.rabbitConnectionFactory(RabbitAutoConfiguration.java:126) ~[spring-boot-autoconfigure-3.1.0.jar:3.1.0]
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method) ~[na:na]
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77) ~[na:na]
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43) ~[na:na]
	at java.base/java.lang.reflect.Method.invoke(Method.java:568) ~[na:na]
	at org.springframework.beans.factory.support.SimpleInstantiationStrategy.instantiate(SimpleInstantiationStrategy.java:139) ~[spring-beans-6.0.9.jar:6.0.9]
[...]

PR #35684, opened 2023-05-31
Implement validity check to spring.rabbitmq.host config property

rafaelrc7:
As discussed on issue #35628, at some point the host property accepted multiple comma-separated hosts. However, this was not intended, and for better clarification, it was decided to implement a clearer error message for this situation.

This PR adds an extra check in `org.springframework.boot.autoconfigure.amqp.RabbitProperties#determineAddresses` for commas in the host property, throwing an `InvalidConfigurationPropertyValueException` if any is found.


mhalbritter:

Issue #35179, opened 2023-04-27
spring-boot-maven-plugin spring-boot.run.directories not working


johnpoth:
Hi all,

It seems the **spring-boot.run.directories** property doesn't work and external JARs in the directory aren't found by the classloader :

```
Caused by: java.lang.NoClassDefFoundError: gattaca/character/Vincent
	at java.base/java.lang.ClassLoader.defineClass1(Native Method) ~[na:na]
	at java.base/java.lang.ClassLoader.defineClass(ClassLoader.java:1012) ~[na:na]
	at java.base/java.security.SecureClassLoader.defineClass(SecureClassLoader.java:150) ~[na:na]
	at java.base/jdk.internal.loader.BuiltinClassLoader.defineClass(BuiltinClassLoader.java:862) ~[na:na]
	at java.base/jdk.internal.loader.BuiltinClassLoader.findClassOnClassPathOrNull(BuiltinClassLoader.java:760) ~[na:na]
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClassOrNull(BuiltinClassLoader.java:681) ~[na:na]
	at java.base/jdk.internal.loader.BuiltinClassLoader.loadClass(BuiltinClassLoader.java:639) ~[na:na]
	at java.base/jdk.internal.loader.ClassLoaders$AppClassLoader.loadClass(ClassLoaders.java:188) ~[na:na]
	at java.base/java.lang.ClassLoader.loadClass(ClassLoader.java:520) ~[na:na]
```

When running the plugin in debug mode I do see that the directory is added to the classpath but the classloader doesn't seem to search for classes in it.

This was already reported on stackoverflow https://stackoverflow.com/questions/62389508/external-jar-class-not-found-while-run-time-in-spring-boot-maven but I can't used the mentioned workaround. Let me know if you need a reproducer.

Thanks!

snicoll:
The directory itself is added, that doesn't mean that JAR files are picked up. The documentation for the property states the following:

> Additional directories besides the classes directory that should be added to the classpath.

If you put a JAR file in the "classes" directory, it won't be picked up either.

snicoll:
What is misleading exactly? A directory and a file are two separate things and the Javadoc, as well as the name of the property is quite explicit about it being the former.

Standard Java requires you to use `*` if you want to load JAR files from directories.

johnpoth:
Hi @snicoll ! I can confirm pointing to **files** instead of **directories** works:

```
[...]

PR #35712, opened 2023-06-02
Rename directories property into additionalClasspathElements

rafaelrc7:
This PR is related to what was discussed in issue #35179. This code renames the ```spring-boot.run.directories``` property into ```spring-boot.run.additionalClasspathElements```.

Furthermore, it changes the code that appends the property's value to the classpath. Currently, the value goes through multiple conversions from ```String``` to ```File```, ```URI``` and ```URL```. This version only converts the string into a ```Path``` to treat non-absolute paths. However, it was commented that this may change as it is desirable to only append the string without any processing.


snicoll:
@rafaelrc7 thanks for the PR but, unfortunately, the change is more involved, please see https://github.com/spring-projects/spring-boot/issues/35179. for more details.

Issue #31658, opened 2022-07-08
Annotate EndpointCloudFoundryExtension with @AliasFor


mhalbritter:
Got this warning in my log:

```
o.s.c.annotation.AnnotationTypeMapping   : Support for convention-based annotation attribute overrides is deprecated and will be removed in Spring Framework 6.1. Please annotate the 'endpoint' attribute in @org.springframework.boot.actuate.autoconfigure.cloudfoundry.EndpointCloudFoundryExtension with an appropriate @AliasFor declaration.
```

I guess this is an oversight that it's not annotated with `@AliasFor`.

philwebb:
Interesting, I wonder if we might find a few more of those.

wilkinsona:
@Tristan971 Thanks, but that isn't related to this issue. Can you please open a Spring Framework issue for those warnings? There's little point in it recommending a change to a Hibernate Validator annotation over which you have no control.

wilkinsona:
Thanks!

bbulgarelli:
Can I do this one?

philwebb:
Yes please @bbulgarelli, I'll assign it to you.

philwebb:
Closing in favor of PR #35716. Thanks @bbulgarelli!

PR #35716, opened 2023-06-03
Processing of @EndpointCloudFoundryExtension logs a warnings as it does not use @AliasFor on its override of the endpoint attribute

bbulgarelli:
This PR is related to the issue #31658. I did what the warning suggested and added the @AliasFor annotation to the 'endpoint' attribute.


mhalbritter:

Issue #11325, opened 2017-12-12
Data store auto-configurations have different default connection behaviour


wilkinsona:
This was raised as part of a [discussion on Twitter](https://twitter.com/abelsromero/status/940537207881912321). The specific example given is Mongo which connects to `localhost` when it's on the classpath and `Couchbase` which doesn't connect to anything until your configure at least one bootstrap host.

There may well be good reasons for the apparent inconsistency, but it appears to be causing some confusion. It would be nice to align the behaviour across data stores if we can. If there are good reasons not to do that, it would be nice to see if we can somehow make the behaviour less surprising.

abelsromero:
The confusion for us came because of two reasons
*One*, we have been using Couchbase without issues, and when we migrated to MongoDB we had to exclude the autoconfigurer. That felt odd.

*Two*,  how the Autoconfiguration starts and logs the MongoClient behauviour...
1. if MongoClient is in the classpath and no configuration is present, it attempts to connect a localhost instance.
2. If connection fails, it does not propagate the error, so the context starts and the app is working fine.
3. However, the error is logged into the logs just after the "success"message. See attachment.
[mongoclient-springboot-autcoonfigurer-error.txt](https://github.com/spring-projects/spring-boot/files/1551476/mongoclient-springboot-autcoonfigurer-error.txt)

Note: whe use a logback logging configuration importing springs defaults
```
    <include resource="org/springframework/boot/logging/logback/defaults.xml"/>
```

Imho, all Autoconfigurators for components depend on some mid-complex configuration should check for properties. Simple things or when a default config is a common thing like Gson or Jackson are fine that they only check for Beans.


Btw, the reason why we have a custom connector and do not expose a MongoClient to the Spring context is because we have a multitennacy platform that manages several connections at once. We load the configurations from a config-service and then initialize all our MongoClients inside a custom bean. That bean is the exposed, but not the clients.

wilkinsona:
Here's the error from the attachment:

```
2017-12-12 12:58:38.732  INFO [multitenancy-server,,,] 9520 --- [           main] s.b.c.e.t.TomcatEmbeddedServletContainer : Tomcat started on port(s): 9040 (http)
2017-12-12 12:58:38.741  INFO [multitenancy-server,,,] 9520 --- [           main] c.e.e.m.MultitenancyServerApplication    : Started MultitenancyServerApplication in 12.765 seconds (JVM running for 13.878)
[...]

PR #35820, opened 2023-06-10
Add condition to Cassandra auto-configuration

jeffreye:
Update Cassandra auto-configuration so that beans in `CassandraAutoConfiguration` may be optionally created


snicoll:
@jeffreye thanks for the PR but we need to take a step back. Rather than offering what you think is a solution to the problem you're experiencing, we'd prefer a description of the problem first. I don't think the cassandra auto-config should be conditional on `spring.cassandra.keyspace-name`.

jeffreye:
1. We would like to use Cassandra driver with our own implementation instead of auto-configured bean. Without configuring spring properties, it will fail the startup currently or we have to exclude the autoconfiguration 
2. CouchbaseAutoConfiguration also has the similar conditions so that it won't create beans unless there is properties

If keyspace name is not a good choice, what is your recommendation?

jeffreye:
Thanks for the info. I'll vote +1 for the issue and hope it could bring more attention to the team.

snicoll:
Indeed, so let's close this one in favor of #11325. Thanks for the suggestion, in any case.

Issue #35694, opened 2023-06-01
unable to conntect to kafka after upgraded to spring boot 3.1.0


robtha:
After upgraded to spring boot 3.1.0 configuration of kafka server via configuration property _spring.kafka.properties.bootstrap.servers_ no longer working.

Always using default value _localhost:9092_ and therefore unable to connect.

Workaround to make it work is using configuration property _spring.kafka.bootstrapServers_ but I don't think this is the intended way for configuration.

wilkinsona:
> Workaround to make it work is using configuration property spring.kafka.bootstrapServers but I don't think this is the intended way for configuration.

This is the intended way.

> kafka server via configuration property `spring.kafka.properties.bootstrap.servers` no longer working

`spring.kafka.properties.*` configuration is for "Additional properties, common to producers and consumers, used to configure the client". In other words, it's for configuring Kafka settings for which Spring Boot does not have a specific property.

That said, even if it wasn't intended, this worked in previous versions. We should consider if we want to reinstate that behavior to make upgrades easier. It would require `PropertiesKafkaConnectionDetails` to consider entries in the `spring.kafka.properties` map.

robtha:
@wilkinsona thanks for clarifying

wilkinsona:
We've decided not to reinstate the previous behaviour as it wasn't intentional. We're going to tighten things up and fail-fast if one of Boot's properties is configured in the `spring.kafka.properties` map of additional properties.

stillya:
@wilkinsona Hello, I've created PR to close this issue. Could you take a look when you get a chance?

PR #35825, opened 2023-06-12
Add validation to fail-fast when a kafka property is configured in spring.kafka.properties when a spring.kafka property could have been used instead

stillya:
Closes gh-35694


snicoll:
@stillya thanks very much for the PR but I am not keen to add so much code for this. We either need to find a way to fail fast without checking each individual property or reconsider the merit of the issue.

PR #35872, opened 2023-06-13
Polish

izeye:
This PR polishes a bit.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/first-application.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/first-application.adoc
@@ -141,8 +141,9 @@ Open your favorite text editor and add the following:
 	plugins {
 		id 'java'
 		id 'org.springframework.boot' version '{spring-boot-version}'
-		id 'io.spring.dependency-management' version '1.1.0'
 	}
+
+	apply plugin: 'io.spring.dependency-management'
 
 	group = 'com.example'
 	version = '0.0.1-SNAPSHOT'
@@ -247,7 +248,7 @@ If you run `gradle dependencies` again, you see that there are now a number of a
 [[getting-started.first-application.code]]
 === Writing the Code
 To finish our application, we need to create a single Java file.
-By default, Maven and Gradle compiles sources from `src/main/java`, so you need to create that directory structure and then add a file named `src/main/java/MyApplication.java` to contain the following code:
+By default, Maven and Gradle compile sources from `src/main/java`, so you need to create that directory structure and then add a file named `src/main/java/MyApplication.java` to contain the following code:
 
 include::code:MyApplication[]
 
--- a/spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/lifecycle/TestcontainersLifecycleBeanPostProcessor.java
+++ b/spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/lifecycle/TestcontainersLifecycleBeanPostProcessor.java
@@ -56,7 +56,7 @@ class TestcontainersLifecycleBeanPostProcessor implements DestructionAwareBeanPo
 
 	private static final Log logger = LogFactory.getLog(TestcontainersLifecycleBeanPostProcessor.class);
 
-	private ConfigurableListableBeanFactory beanFactory;
+	private final ConfigurableListableBeanFactory beanFactory;
 
 	private volatile boolean containersInitialized = false;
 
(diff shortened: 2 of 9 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/first-application.adoc
```diff
@@ -141,9 +141,10 @@ Open your favorite text editor and add the following:
 	plugins {
 		id 'java'
 		id 'org.springframework.boot' version '{spring-boot-version}'
-		id 'io.spring.dependency-management' version '1.1.0'
 	}
 
+	apply plugin: 'io.spring.dependency-management'
```
jonatan-ivanov, line 146: Does this work without defining a `classpath` dependency in `buildscript`?
Even if it does, I might keep it as is (despite of I'm not a fan of using `id`) to keep it consistent with the [docs](https://docs.spring.io/dependency-management-plugin/docs/current/reference/html/#getting-started) of the plugin and the output of [initializr](https://start.spring.io/).

spring-boot-project/spring-boot-testcontainers/src/test/java/org/springframework/boot/testcontainers/service/connection/ServiceConnectionContextCustomizerTests.java
```diff
@@ -123,7 +123,7 @@ void equalsAndHashCode() {
 				List.of(new ContainerConnectionSource<>("test", this.origin, PostgreSQLContainer.class, "name",
 						annotation3, () -> container1)));
 		assertThat(t1.hashCode()).isEqualTo(t2.hashCode());
-		assertThat(t1).isEqualTo(t2).isNotEqualTo(t3);
+		assertThat(t1).isEqualTo(t2).isNotEqualTo(t3).isNotEqualTo(t3.hashCode());
```
jonatan-ivanov, line 126: `t1` is a `ServiceConnectionContextCustomizer` and `t3.hashCode()` is an `int`.
Did you want to do
```java
assertThat(t1.hashCode()).isEqualTo(t2.hashCode()).isNotEqualTo(t3.hashCode());
```
in the previous line?

spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/lifecycle/TestcontainersLifecycleBeanPostProcessor.java
```diff
@@ -56,7 +56,7 @@ class TestcontainersLifecycleBeanPostProcessor implements DestructionAwareBeanPo
 
 	private static final Log logger = LogFactory.getLog(TestcontainersLifecycleBeanPostProcessor.class);
 
-	private ConfigurableListableBeanFactory beanFactory;
+	private final ConfigurableListableBeanFactory beanFactory;
```
jonatan-ivanov, line 59: Why is this better as non-final?

jonatan-ivanov (changes requested):

spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/first-application.adoc
izeye: @jonatan-ivanov I thought it's a better way as we can rely on `org.springframework.boot` Gradle plugin managed version this way.

See https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/htmlsingle/#getting-started

spring-boot-project/spring-boot-testcontainers/src/test/java/org/springframework/boot/testcontainers/service/connection/ServiceConnectionContextCustomizerTests.java
izeye: @jonatan-ivanov Thanks! It's just a copy and paste error. I fixed it in 6b6d2a4.

spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/lifecycle/TestcontainersLifecycleBeanPostProcessor.java
izeye: @jonatan-ivanov I don't know its exact performance implication, but I thought expressing its immutability at least helps its readability.

jonatan-ivanov: I think I'm tired, I could swear that I read that `final` got removed here instead of added, it makes sense, sorry for the confusion.

jonatan-ivanov (approved):

jonatan-ivanov:
Thank you!

spring-boot-project/spring-boot-docs/src/docs/asciidoc/getting-started/first-application.adoc
mhalbritter: Interesting. The Gradle plugin documentation recommends something different than start.spring.io uses.

@wilkinsona What do you think what we should put in the docs?

wilkinsona: I think we should make this change so that Boot's documentation is consistent with itself at least. I can't recall why start.spring.io is different but I think that can be dealt with separately.

jonatan-ivanov: fyi: https://github.com/spring-io/initializr/issues/1428

PR #35890, opened 2023-06-14
Support for Base64 encoded p12 keystore using the ssl bundle configuration

erwinc1:
I've added support for Base64 encoded p12 formatted keystores. [This is my backing StackOverflow question](https://stackoverflow.com/questions/76464923/how-to-use-the-ssl-autoconfiguration-with-a-base64-encoded-pkcs12)

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ssl/jks/JksSslStoreBundle.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ssl/jks/JksSslStoreBundle.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.ssl.jks;
 
+import java.io.ByteArrayInputStream;
 import java.io.IOException;
 import java.io.InputStream;
 import java.net.URL;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ssl/jks/JksSslStoreBundle.java
```diff
@@ -109,14 +111,29 @@ private void loadHardwareKeyStore(KeyStore store, String location, char[] passwo
 	private void loadKeyStore(KeyStore store, String location, char[] password) {
 		Assert.state(StringUtils.hasText(location), () -> "Location must not be empty or null");
 		try {
-			URL url = ResourceUtils.getURL(location);
-			try (InputStream stream = url.openStream()) {
-				store.load(stream, password);
+			if (isBase64(location.getBytes())) {
```
martinvisser, line 114: Maybe instead of doing the decode twice, where you test first if you can decode it, you could do the else part in the catch block?

scottfrederick:
@erwinc1 Thanks very much for the suggestion. We agree this is something that we'd like Spring Boot to support. However, we'd like to take a different approach to the problem that would allow us to apply the same pattern to other places where file resources are currently supported in configuration properties instead of just making your suggested change to SSL bundle JKS keystores. I've opened #36033 to capture this.

PR #35914, opened 2023-06-15
SSL configuration overwrites other WebClient customization

fcappi:
Issue: Whenever you are configuring TLS on a webclient, any other custom configuration of HttpClient is lost and not applied to the final instance.

Example:
```java
  @Bean
  public WebClient webClient(
      ...
    return WebClient.builder().apply(clientSsl.fromBundle(sslBundle)).build();
  }

  @Bean
  public ReactorNettyHttpClientMapper httpClient() {
   // This is not applied to the WebClient instance generated by the builder on webclient()
    return client ->
        client.compress(true)
```

Check the code being changed, where we can see that mapper is being reassigned to only consider configurations from SslConfigurer. It should be just added to the map stream, so everything is applied


scottfrederick:
@fcappi Thanks for the suggestion. Can you describe the problem that this change is intended to solve? Ideally we'd have [tests](https://github.com/spring-projects/spring-boot/blob/3.1.x/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/function/client/AbstractClientHttpConnectorFactoryTests.java) to verify that the existing behavior is incorrect, and change the tests to match the new behavior.

fcappi:
Hey @scottfrederick , sorry for that. I created the PR before it was ready. Details added to the PR description. Tests are covering the expected behaviour.

philwebb:
Thanks very much for contributing @fcappi!

Issue #35791, opened 2023-06-08
Replace calls to verifyComplete() with expectComplete().verify(Duration) to avoid blocking indefinitely


wilkinsona:
We use `verifyComplete()` in several places. With the default `StepVerifier` timeout this will block indefinitely, potentially causing a test to hang. The "global" timeout can be configured by calling `StepVerifier.setDefaultTimeout(Duration)` but this is easy to forget and also has side-effects as it sets the timeout for all subsequent calls to `verifyComplete()`, even if they're in a different test within the same module. I think it would be more robust to use `expectComplete().verify(Duration)` everywhere instead.

We should also consider introducing an ArchUnit rule to prevent usage of `verifyComplete()` from being reintroduced.

PR #35915, opened 2023-06-15
Replace calls to verifyComplete() with expectComplete().verify(Duration) to avoid blocking indefinitely

ThomazPassarelli:
This pull request addresses the issue #35791, where it was required to replace `verifyComplete()` with `expectComplete().verify(Duration)` in test cases across multiple files. 

Changes made:

1. Replaced `verifyComplete()` with `expectComplete().verify(Duration.ofSeconds(5))` in all the necessary places. The decision to use `Duration.ofSeconds(5)` was influenced by its existing usage in another part of the project, ensuring consistency across the codebase.
2. Added `import java.time.Duration;` where it was not already imported.

I have successfully made changes to 23 files and built the project without any errors. All the tests ran successfully after the changes.


philwebb:
When we merge this we should also consider this comment from the original issue:

> We should also consider introducing an ArchUnit rule to prevent usage of verifyComplete() from being reintroduced.

wilkinsona:
I've got one locally.

We should also remove any calls to `StepVerifier.setDefaultTimeout` as they're unnecessary when using `expectedComplete().verify(Duration)` everywhere.

wilkinsona:
@ThomazPassarelli can you please [sign the CLA](https://cla.pivotal.io/sign/spring?repositoryId=spring-projects/spring-boot&pullRequestId=35915) when you have a moment?

wilkinsona:
Thanks very much, @ThomazPassarelli.

Issue #35861, opened 2023-06-13
Add SslAutoConfiguration to slice test annotations


eddumelendez:
Existing examples such as [SampleMongoApplicationReactiveSslTests](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-data-mongo/src/test/java/smoketest/data/mongo/SampleMongoApplicationReactiveSslTests.java) uses `@SpringBootTest` which will use `SslAutoConfiguration`. If `@DataMongo` is used then `@Import(SslAutoConfiguration.class)` should be declared. I think it would be great to add `SslAutoConfigure` [here](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.test.autoconfigure.data.mongo.AutoConfigureDataMongo.imports)

The same for the rest of them such `DataRedis`, `DataCouchbase`, `DataElasticSearch`.

I would like to know your opinion on this topic before raising any PR.

Thanks in advance

SoumyoNathTripathy:
// For MongoDB:
// If you are using the @SpringBootTest annotation, which relies on SslAutoConfiguration, and you are also using
// @DataMongo, you may need to explicitly import SslAutoConfiguration in your test class using @Import annotation.
// This ensures that SSL configuration is applied correctly for the MongoDB module.

scottfrederick:
Closing in favor of #36038

PR #35935, opened 2023-06-18
Create Add SslAutoConfiguration to slice test annotation.java

SoumyoNathTripathy:
In this example, the @Import(SslAutoConfiguration.class) annotation is added to import the SslAutoConfiguration class, which enables SSL configuration for MongoDB. Make sure to replace MyMongoConfiguration with your actual configuration class name.

Add SslAutoConfiguration to slice test annotations #35861 template


scottfrederick:
@SoumyoNathTripathy Thanks for the suggestion. The file you've added in this commit doesn't really help with the issue described in #35861, since that issue is meant to deal with tests using `@SpringBootTest` and slice tests such as `@DataMongo`.  We need to discuss how best to handle #35861 as a team, and will likely want a solution that addresses all supported test slices.

PR #35984, opened 2023-06-19
Refactor SpringBootBanner Class for Readability and Maintainability

dkswnkk:
Hi, this is a small thing, but I wanted to put a PR up for an improvement.

I made the printBanner method more concise and separated the logic for each action into separate methods so that each method better aligned with the SRP responsibilities.


wilkinsona:
Thanks for the proposal but I prefer the code as it is at the moment. At only 13 lines long, I don't think the method benefits much from being split up. Thanks anyway.

PR #35986, opened 2023-06-20
Remove redundant Files.delete() calls in tests

izeye:
This PR adds missing `Files.delete()` calls in the `PrivateKeyParserTests`.


scottfrederick:
@izeye The [cleanup call here](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/test/java/org/springframework/boot/buildpack/platform/docker/ssl/PrivateKeyParserTests.java#L57) should take care of deleting the temporary files. Are you seeing the files left around after tests are complete?

izeye:
@scottfrederick Ah, I just thought that they are missed as the other tests have a call to the `Files.delete()`. Does it make sense to remove the calls in the other tests then?

izeye:
@scottfrederick Thanks for the feedback! I updated this PR to remove them instead in both tests.

scottfrederick:
Thanks, as always @izeye.

PR #36032, opened 2023-06-22
Polish Kafka Properties Docs

garyrussell:
There are 4 client types; there were a couple of places where not all client types were mentioned (such as 'applies to producers, consumers and admins').


wilkinsona:

Issue #35252, opened 2023-05-03
@ServiceConnection on a @Bean method does not work in sliced tests


wilkinsona:
Sliced tests don't include `ServiceConnectionAutoConfiguration` so the `@ServiceConnection` annotations aren't honoured.

wilkinsona:
Closing in favor of #36037.

PR #36037, opened 2023-06-23
@ServiceConnection on a @Bean method does not work in sliced tests

eddumelendez:
Currently, in order to use `@ServiceConnection` in beans along with
slice test annotations `@ImportAutoConfiguration(ServiceConnectionAutoConfiguration.class)`
should be added to make it work. This commit register
`ServiceConnectionAutoConfiguration` in specific slice test annotations.

See gh-35252


spring-boot-project/spring-boot-test-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.test.autoconfigure.data.couchbase.AutoConfigureDataCouchbase.imports
```diff
@@ -1,5 +1,5 @@
 # AutoConfigureDataCouchbase auto-configuration imports
-
+org.springframework.boot.test.autoconfigure.service.connection.TestcontainersServiceConnectionAutoConfiguration
```
eddumelendez, line 2: Another idea I just had is having `META-INF/spring/org.springframework.boot.test.autoconfigure.data.cassandra.AutoConfigureDataCassandra.imports` and others in `spring-boot-testcontainers` module adding `org.springframework.boot.testcontainers.service.connection.ServiceConnectionAutoConfiguration`. I expect those will be merged when available and have the same result. Feedback is welcome

spring-boot-project/spring-boot-test-autoconfigure/src/test/java/org/springframework/boot/test/autoconfigure/data/redis/DataRedisTestWithServiceConnectionBeanIntegrationTests.java
```diff
@@ -70,7 +70,7 @@ void testRepository() {
 	static class ContainerConfig {
 
 		@Bean
-		@ServiceConnection
+		@ServiceConnection(name = "redis")
```
quaff, line 73: Should somewhere else be improved to avoid explicit `name`?

quaff: I've created https://github.com/spring-projects/spring-boot/pull/36064 to handle this.

wilkinsona:
Thanks for the PR, @eddumelendez.

These changes introduce a cycle between `spring-boot-testcontainers` and `spring-boot-test-autoconfigure`:

- the former has `.imports` files  in `src/main/resources` for each of the sliced test annotations defined by the latter 
- the latter has tests that make use of `@ServiceConnection` which is defined by the former

I'm not sure how best to break this cycle. A couple of options:

1. move all of the tests into `spring-boot-testcontainers` and remove the test dependency on `spring-boot-testcontainers` from `spring-boot-test-autoconfigure`
2. merge the `.imports` files into those already in `spring-boot-test-autoconfigure` but this would require support for an import being "optional" as `spring-boot-testcontainers` wouldn't necessarily be on the classpath when using `spring-boot-test-autoconfigure`

1 feels the slightly better of these two options. There may also be others that I've overlooked. I'll discuss things with the team.

wilkinsona:
Thanks very much for the PR, @eddumelendez. In the end, we went with the second of the two options above. An entry in a `.imports` file can now be prefixed with `optional:`. An optional import that is not present on the classpath will be ignored.

Issue #35940, opened 2023-06-19
JobLauncherApplicationRunner returns a success exit code even when no jobs have been run.


umbum:
When using `JobLauncherApplicationRunner`, if we pass an arbitrary `JobName`, the spring application terminates with a success exit code without failure.

If the job name is entered incorrectly by mistake, it could be mistaken for success even though no job was actually executed.

It would be good to add validation whether the job name is defined in application.

#### sol. 1
```java
public class JobLauncherApplicationRunner implements ApplicationRunner, Ordered, ApplicationEventPublisherAware {
    ...
	@Value("${spring.batch.job.fail-on-unknown-name:false}")
	private boolean failOnUnknownName;

	@PostConstruct
	public void validate() {
		if (this.jobs.size() > 1 && !StringUtils.hasText(this.jobName)) {
			throw new IllegalArgumentException("Job name must be specified in case of multiple jobs");
		}

		if (failOnUnknownName == true && this.jobs.stream().noneMatch(job -> StringUtils.equals(this.jobName, job.getName()))) {
			throw IllegalArgumentException("No such job name : [$jobName]")
		}
	}
```

#### sol. 2
```java
public class JobExecutionExitCodeGenerator implements ApplicationListener<JobExecutionEvent>, ExitCodeGenerator {
	@Override
	public int getExitCode() {
		if (this.executions.isEmpty()) {
			return 127;
		}

		for (JobExecution execution : this.executions) {
			if (execution.getStatus().ordinal() > 0) {
				return execution.getStatus().ordinal();
			}
		}
		return 0;
	}
```

philwebb:
Another option might be to change [this line](https://github.com/spring-projects/spring-boot/blob/16e993f3ca045128024c85dfc6a26d727793a105/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java#L191) to not always swallow `NoSuchJobException`.

philwebb:
@fmbenhassine Do you have any thoughts on this one?

fmbenhassine:
Hi @philwebb , yes, I would also follow your suggestion of not swallowing the exception.

itsAkshayDubey:
Hello @philwebb @fmbenhassine,

Raised PR #36039 to fix this, kindly have a look.

Thanks,
Akshay

scottfrederick:
Closing in favor of #36039

[...]

PR #36039, opened 2023-06-23
JobLauncherApplicationRunner returns a success exit code even when no jobs have been run

itsAkshayDubey:
JobLauncherApplicationRunner returns a success exit code even when no jobs have been run.
fix: Rethrowing JobExecutionException when NoSuchJobException is caught.
Closes: #35940

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
@@ -14,237 +14,233 @@
  * limitations under the License.
  */
 
-package org.springframework.boot.autoconfigure.batch;
+ package org.springframework.boot.autoconfigure.batch;
 
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.LinkedHashMap;
-import java.util.Map;
-import java.util.Properties;
-
-import org.apache.commons.logging.Log;
-import org.apache.commons.logging.LogFactory;
-
-import org.springframework.batch.core.BatchStatus;
-import org.springframework.batch.core.Job;
-import org.springframework.batch.core.JobExecution;
-import org.springframework.batch.core.JobExecutionException;
-import org.springframework.batch.core.JobParameter;
-import org.springframework.batch.core.JobParameters;
-import org.springframework.batch.core.JobParametersBuilder;
-import org.springframework.batch.core.JobParametersInvalidException;
-import org.springframework.batch.core.configuration.JobRegistry;
-import org.springframework.batch.core.converter.DefaultJobParametersConverter;
-import org.springframework.batch.core.converter.JobParametersConverter;
-import org.springframework.batch.core.explore.JobExplorer;
-import org.springframework.batch.core.launch.JobLauncher;
-import org.springframework.batch.core.launch.JobParametersNotFoundException;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
```diff
@@ -187,6 +187,7 @@ private void executeRegisteredJobs(JobParameters jobParameters) throws JobExecut
 				}
 				catch (NoSuchJobException ex) {
 					logger.debug(LogMessage.format("No job found in registry for job name: %s", jobName));
+					throw new JobExecutionException(ex.getMessage(), ex);
```
wilkinsona, line 190: `NoSuchJobException` is a `JobExecutionException` so I don't think there's any need to create a new exception. Instead, I think the `try` and `catch` could be removed completely so that the method just throws to original `NoSuchJobException`.

wilkinsona (changes requested):
Thanks for the PR, @itsAkshayDubey. I've left a comment for your consideration. It would also be good to update `JobLauncherApplicationRunnerTests` to add a test that asserts that an exception is thrown when an attempt is made to execute a job that doesn't exist.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
itsAkshayDubey: Fixed in [this](https://github.com/spring-projects/spring-boot/pull/36039/commits/c4d7bef0ae3c202a2765c2a3dbfd626a069ff8e4) commit.

itsAkshayDubey:
> Thanks for the PR, @itsAkshayDubey. I've left a comment for your consideration. It would also be good to update `JobLauncherApplicationRunnerTests` to add a test that asserts that an exception is thrown when an attempt is made to execute a job that doesn't exist.

Added in [this](https://github.com/spring-projects/spring-boot/pull/36039/commits/d0db639eee23c36f4f96d2db2f9b85ff8aa87296) commit

itsAkshayDubey:
Hello @wilkinsona @scottfrederick ,

Pl. have a look.

Thanks,
Akshay

scottfrederick:
Unfortunately, the suggestion of just removing the try/catch isn't the right fix for this issue. 

When this issue is fixed, [this test](https://github.com/spring-projects/spring-boot/blob/2.7.x/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java#L163-L174) should continue to pass as it is currently written, since the job named `discreteRegisteredJob` is a valid local job with this test setup. Removing the try/catch results in the exception being thrown from `executeRegisteredJobs` after the local job has been run successfully in `executeLocalJobs`.  

What we need is a way to detect that a given job in `spring.batch.job.names` was not run by [either `executeLocalJobs` or `executeRegisteredJobs`](https://github.com/spring-projects/spring-boot/blob/2.7.x/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java#L160-L161). A new test that sets the `spring.batch.job.names` property to include at least one job name that has not been configured will be required to verify the desired behavior. 

I'm going to close this PR and re-open the issue so we can think more about the solution. @itsAkshayDubey  If you'd like to take another look and suggest a fix that meets the requirements in a new PR, let us know in the issue and we can assign it to you.

Issue #35674, opened 2023-05-30
Change B3 extraction format to SINGLE


mhalbritter:
Right now, we're using `B3Propagation.Format.SINGLE_NO_PARENT`. We should change this to `SINGLE`.

See [this discussion](https://github.com/spring-projects/spring-boot/pull/35611#discussion_r1205290111).

KitBolourchi:
Hey @mhalbritter, I would be happy to look into this issue. Please assign it to me :)

scottfrederick:
Thanks @KitBolourchi, I've assigned the issue to you. We have not yet created the branches to start `3.2.x` development, so we won't be able to accept a contribution for a while.

KitBolourchi:
Hey @scottfrederick, no problem at all! Would you recommend I make the change and create a PR which can be merged when ready or wait until you are ready before raising the PR?

mhalbritter:
https://github.com/spring-projects/spring-boot/pull/35611/files will move some things around, so please hold off until this PR is merged, otherwise there will be quite some merge conflicts.

KitBolourchi:
Gotcha I understand, I will hold off until the release branch is ready!

mhalbritter:
#35611 is merged, you can now start :)

KitBolourchi:
Great, I will get started on this shortly :) Thanks

PR #36061, opened 2023-06-24
Change B3 extraction format to single

KitBolourchi:
Link to issue ticket: https://github.com/spring-projects/spring-boot/issues/35674

Change B3 propagation to B3Propagation.Format.SINGLE

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactory.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactory.java
@@ -124,7 +124,7 @@ class CompositePropagationFactory extends Propagation.Factory {
 		 * @return the B3 propagation factory
 		 */
 		private Propagation.Factory b3Single() {
-			return B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.SINGLE_NO_PARENT).build();
+			return B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.SINGLE).build();
 		}
 
 		/**
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactoryTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactoryTests.java
@@ -21,8 +21,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import brave.Span;
-import brave.Tracing;
 import brave.internal.propagation.StringPropagationAdapter;
 import brave.propagation.Propagation;
 import brave.propagation.TraceContext;
@@ -30,8 +28,6 @@ import brave.propagation.TraceContextOrSamplingFlags;
 import org.junit.jupiter.api.Nested;
 import org.junit.jupiter.api.Test;
 import org.mockito.Mockito;
-import org.springframework.boot.autoconfigure.AutoConfigurations;
-import org.springframework.boot.test.context.runner.ApplicationContextRunner;
 
 import static org.assertj.core.api.Assertions.assertThat;
 import static org.assertj.core.api.Assertions.entry;
@@ -71,9 +67,6 @@ class CompositePropagationFactoryTests {
 	@Nested
 	static class CompostePropagationTests {
 
-		private final ApplicationContextRunner contextRunner = new ApplicationContextRunner()
-				.withConfiguration(AutoConfigurations.of(CompositePropagationFactory.class));
-
 		@Test
 		void keys() {
 			CompositePropagationFactory factory = new CompositePropagationFactory(List.of(field("a")),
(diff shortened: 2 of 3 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactory.java
```diff
@@ -124,7 +124,7 @@ Propagation.Factory map(PropagationType type) {
 		 * @return the B3 propagation factory
 		 */
 		private Propagation.Factory b3Single() {
-			return B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.SINGLE_NO_PARENT).build();
+			return B3Propagation.newFactoryBuilder().injectFormat(B3Propagation.Format.SINGLE).build();
```
jonatan-ivanov, line 127: Should we add a test to verify this?
I think an extra verification step to one of our existing (integration?) tests is enough.

mhalbritter: Yep, i agree. Adding a test would be great.

KitBolourchi: Ah that's a good idea, my bad! I won't have access to my laptop until next week and I can re-review.

KitBolourchi: Hey both, I now have access to my laptop and am looking to review this. Quick question, should this test step be included in either of the existing unit tests in file CompositePropagationFactoryTests.java or should I create a new unit test?

jonatan-ivanov: You can try to test this in `CompositePropagationFactoryTests` since you have access to the package-private `create` method, you might be able to test this in a very simple way. I think (didn't try) the tricky part will be the verification since you might not have access to the propagation format to check what it is.

What you can also try if this won't work is testing the behavior or the propagator. Check `BraveAutoConfigurationTests` there you can see how to create spans (you will need at least two: a parent and a child). You can create a test injector and "inject the span context into a variable". Then you can check what was injected. Before your change the parent id should not be there after that it should. Does this make sense? Please let us know if you need help.

KitBolourchi: Hey @jonatan-ivanov thank you for the explanation, yes I think that makes sense. I will have a review and have a look into this and keep you updated :) have a lovely weekend.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactoryTests.java
```diff
@@ -86,6 +93,24 @@ void inject() {
 			assertThat(request).containsOnly(entry("a", "a-value"), entry("b", "b-value"));
 		}
 
+		@Test
+		void thisIsATest() {
```
jonatan-ivanov, line 97: Could you please name it based on what it does?
E.g.: `shouldUseB3SingleWithParentWhenPropagationTypeIsB3`

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactoryTests.java
```diff
@@ -86,6 +93,24 @@ void inject() {
...
+			this.contextRunner
+					.withPropertyValues("management.tracing.propagation.type=B3",
+							"management.tracing.brave.span-joining-supported=true")
```
jonatan-ivanov, line 100: Why does this need joint span support?
I don't think it does.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactoryTests.java
```diff
@@ -86,6 +93,24 @@ void inject() {
...
+					.run((context) -> {
+						CompositePropagationFactory factory = new CompositePropagationFactory(List.of(field("a"), field("b")),
+								List.of(field("c")));
+						Propagation<String> propagation = factory.get();
+						Tracing tracing = context.getBean(Tracing.class);
+						Span parentSpan = tracing.tracer().nextSpan();
+						Span childSpan = tracing.tracer().joinSpan(parentSpan.context());
```
jonatan-ivanov, line 107: I don't think you need `joinSpan` to test this.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactoryTests.java
```diff
@@ -86,6 +93,24 @@ void inject() {
...
+						Map<String, String> request = new HashMap<>();
+						propagation.injector(new CompositePropagationFactoryTests.MapSetter()).inject(parentSpan.context(), request);
```
jonatan-ivanov, line 109: This test does not contain any verification steps: it will ~never fail, thus it is not very useful. The `Map` should contain the B3 header (with the parent id).
Also, please check the docs of Micrometer Tracing how to create a child span (you need to open a scope).

jonatan-ivanov (changes requested):

jonatan-ivanov:
I rebased your changes on `main`, squashed your commits together and added a test: https://github.com/spring-projects/spring-boot/compare/main...jonatan-ivanov:spring-boot:b3-single
What do you think?

KitBolourchi:
> 

I see, yes this looks good. That makes a lot of sense and clearly outlines the mistakes I made. Would you propose I close this PR down and continue with your changes or I rectify and fix my tests for this PR?

jonatan-ivanov:
I don't think there is a need for another PR, you can `fetch` my branch and `reset` yours to my last commit (you will need to force `push`).

KitBolourchi:
> I don't think there is a need for another PR, you can `fetch` my branch and `reset` yours to my last commit (you will need to force `push`).

Should be all done now. @jonatan-ivanov Please check that all is aligned and as expected. I appreciate the help, I learned a lot with this PR :)

jonatan-ivanov (approved):

mhalbritter:
Thank you all!

PR #36063, opened 2023-06-25
Upgrade gradle to 8.1.1

quaff:
Adapt gradle API changes:
1. Gradle 8 does no longer allow `classifier` inside Jar task. `archiveClassifier` must be used instead.
2. Upgrade nohttp to fix `java.lang.NoSuchMethodError: 'void org.gradle.api.reporting.SingleFileReport.setDestination(org.gradle.api.provider.Provider)'`
3. `GradlePropertiesController::loadGradlePropertiesFrom` signature changed.


wilkinsona:
Thanks, @Quaff, but we prefer to handle upgrades that change Gradle's wrapper jar ourselves. This is based on the [recommendations of the Gradle team](https://blog.gradle.org/project-integrity):

> Prefer to update the wrapper yourself instead of merging a PR from an external contributor. Regenerate the `gradlew`, `gradle-wrapper.properties`, and `gradle-wrapper.jar` from a known good Gradle distribution. Invoking a local Gradle with `gradle wrapper` will not run the `gradle-wrapper.jar` of the project but it will configure the build.

I think we may need to wait before we can make this change as builds of our `main` branch currently [report a number of deprecations](https://ge.spring.io/s/ju3rdmbw3qxek/deprecations) for things that are "scheduled to be removed in Gradle 8.0" or that will "fail with an error in Gradle 8.0".

quaff:
@wilkinsona Currently this project cannot build by gradle 8, It's a bit inconvenient, I can revert the wrapper part and make some changes to allow it be built by both 7.6 and 8, I would open another PR if you agree.

wilkinsona:
Thanks, but we don't want to support building the project with multiple different versions of Gradle. You should use the wrapper or a local install that matches the wrapper's version.

wilkinsona:
Having said that, there may be some parts of this that still make sense generally across all supported branches. For example, the upgrade to nohttp 0.0.11. I've opened https://github.com/spring-projects/spring-boot/issues/36074. I'll look at some of the other changes too.

PR #36064, opened 2023-06-25
 Use beanName as default name of @ServiceConnection

quaff:
Before this commit, `ConnectionDetailsNotFoundException` will be raised if unnamed `@ServiceConnection` is annotated on `@Bean` method and `containerType` is not accepted by `ContainerConnectionSource`.

For example:
```java
	@Bean
	@ServiceConnection(name = "redis")
	GenericContainer<?> redis() {
		return new GenericContainer<>("redis").withExposedPorts(6379);
	}
```
After this commit, name of `@ServiceConnection` could be omitted.

### DIFF
--- a/spring-boot-project/spring-boot-testcontainers/build.gradle
+++ b/spring-boot-project/spring-boot-testcontainers/build.gradle
@@ -47,6 +47,7 @@ dependencies {
 	testImplementation("org.springframework:spring-core-test")
 	testImplementation("org.springframework:spring-r2dbc")
 	testImplementation("org.springframework.amqp:spring-rabbit")
+	testImplementation("org.springframework.data:spring-data-redis")
 	testImplementation("org.springframework.kafka:spring-kafka")
 	testImplementation("org.testcontainers:junit-jupiter")
 
--- a/spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/service/connection/ServiceConnectionAutoConfigurationRegistrar.java
+++ b/spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/service/connection/ServiceConnectionAutoConfigurationRegistrar.java
@@ -16,7 +16,9 @@
 
 package org.springframework.boot.testcontainers.service.connection;
 
+import java.util.HashMap;
 import java.util.LinkedHashSet;
+import java.util.Map;
 import java.util.Set;
 
 import org.testcontainers.containers.Container;
@@ -38,6 +40,7 @@ import org.springframework.core.type.AnnotationMetadata;
  * {@link ServiceConnectionAutoConfiguration}.
  *
  * @author Phillip Webb
+ * @author Yanming Zhou
  */
 class ServiceConnectionAutoConfigurationRegistrar implements ImportBeanDefinitionRegistrar {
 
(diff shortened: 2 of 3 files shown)


wilkinsona:
Thanks for the proposal, @quaff. @snicoll suggested this a few weeks ago but I wasn't in favor as it feels a bit too magical. It also won't work for containers with `/` in their name such as `openzipkin/zipkin`. Let's see what the rest of the team thinks.

quaff:
> Thanks for the proposal, @quaff. @snicoll suggested this a few weeks ago but I wasn't in favor as it feels a bit too magical. It also won't work for containers with `/` in their name such as `openzipkin/zipkin`. Let's see what the rest of the team thinks.

What about bean name as `openzipkin_zipkin` or `openzipkinZipkin`?

eddumelendez:
IIRC, this was introduced in order to make sure the ConnectionFactory to be used. This is due there is no specific Testcontainers implementation. Redis and Zipkin fall into this because both use GenericContainer but the issue comes with custom images ("my-custom-kv-store") which will always requires `@ServiceName(name = "redis")`. The existing approach is good, instead of making it optional and rely on the bean's name and increase cognitive complexity

quaff:
If people migrate
```java
@ServiceConnection
GenericContainer<?> redis = return new GenericContainer<>("redis").withExposedPorts(6379);
```
to
```java
@Bean
@ServiceConnection
GenericContainer<?> redis() {
	return new GenericContainer<>("redis").withExposedPorts(6379);
}
```
They will encounter unexpected `ConnectionDetailsNotFoundException `.

wilkinsona:
That's not because the field is called `redis` but because the `GenericContainer` instance is available, allowing us to call `getDockerImageName()` on it:

> By default `Container.getDockerImageName()` is used to obtain the name used to find connection details. If you are using a custom docker image, you can use the name attribute of `@ServiceConnection` to override it.

When you're using `@Bean`, the container instance isn't available without calling the method which we can't do as it would trigger eager initialization. As a result, all we have available is the return type of the `@Bean` method. We should improve [the documentation for this](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.testing.testcontainers.at-development-time) as it's not explained at the moment and I think it should be. I've opened https://github.com/spring-projects/spring-boot/issues/36071.

quaff:
> That's not because the field is called `redis` but because the `GenericContainer` instance is available, allowing us to call `getDockerImageName()` on it:
> 
> > By default `Container.getDockerImageName()` is used to obtain the name used to find connection details. If you are using a custom docker image, you can use the name attribute of `@ServiceConnection` to override it.
> 
> When you're using `@Bean`, the container instance isn't available without calling the method which we can't do as it would trigger eager initialization. As a result, all we have available is the return type of the `@Bean` method. We should improve [the documentation for this](https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.testing.testcontainers.at-development-time) as it's not explained at the moment and I think it should be. I've opened #36071.

I know that because I read sources before preparing this PR, but most people doesn't read sources even documents, they want intuitional programming experience.

quaff:
Using beanName as default name of `@ServiceConnection` is better choice, I've updated the commit.

wilkinsona:
We've discussed this today and decided that we don't want to make any changes here. We think there's too much magic in using the bean name. Thanks anyway for the suggestion, @quaff. We do think the docs should be improved though. https://github.com/spring-projects/spring-boot/issues/36071 is tracking that.

PR #36068, opened 2023-06-25
replace forloop with stream in enum

dkswnkk:
I tried changing the for-loop in the get method from enum to stream.

I wasn't going to submit it because readability can be subjective, but the 
- https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/EmbeddedDatabaseConnection.java#L153

I saw that this code, which does the same thing, is using a stream, so I thought it might be okay to submit, so I changed the two places where I was using the for-loop.


philwebb:
Thanks for the PR, but I don't think there's a massive improvement in readability so I'd prefer to keep the code as it is.

PR #36077, opened 2023-06-26
Dependency management for Selenium 4.8.x is incorrect

eddumelendez:


wilkinsona:
We'll have to double-check the bom's contents but we might be able to apply this to 3.1.x where we're using 4.8.3 which [has a bom](https://repo.maven.apache.org/maven2/org/seleniumhq/selenium/selenium-bom/4.8.3/selenium-bom-4.8.3.pom).

wilkinsona:
Checking the dependency management in 3.1.x has revealed that what we have at the moment is incorrect. We're managing `selenium-devtools-v108` which, in the 4.8.x line, only exists in `4.8.0` and `4.8.1`. We aren't managing `selenium-devtools-v111` which was introduced in `4.8.2`. Switching to the bom will correct this.

wilkinsona:

PR #36103, opened 2023-06-28
Change server log message to have "port" or "ports" rather than "ports(s)"

asashour:
There are some places in the logs, which says `ports(s)`.

I believe it would be better to say only `port` or `ports` according to the number of ports, if it is 1, then use the singular form, otherwise use the plural form.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyWebServer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyWebServer.java
@@ -168,8 +168,8 @@ public class JettyWebServer implements WebServer {
 					}
 				}
 				this.started = true;
-				logger.info("Jetty started on port" + (connectors.length == 1 ? "" : "s") + " "
-						+ getActualPortsDescription() + " with context path '" + getContextPath() + "'");
+				logger.info("Jetty started on " + getActualPortsDescription() + " with context path '"
+						+ getContextPath() + "'");
 			}
 			catch (WebServerException ex) {
 				stopSilently();
@@ -184,10 +184,18 @@ public class JettyWebServer implements WebServer {
 
 	private String getActualPortsDescription() {
 		StringBuilder ports = new StringBuilder();
-		for (Connector connector : this.server.getConnectors()) {
-			if (ports.length() != 0) {
+		Connector[] connectors = this.server.getConnectors();
+		ports.append("port");
+		if (connectors.length != 1) {
+			ports.append("s");
+		}
+		ports.append(" ");
+
+		for (int i = 0; i < connectors.length; i++) {
+			if (i != 0) {
 				ports.append(", ");
 			}
+			Connector connector = connectors[i];
 			ports.append(getLocalPort(connector)).append(getProtocols(connector));
 		}
 		return ports.toString();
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyWebServer.java
```diff
@@ -168,8 +168,8 @@ public void start() throws WebServerException {
 					}
 				}
 				this.started = true;
-				logger.info("Jetty started on port(s) " + getActualPortsDescription() + " with context path '"
-						+ getContextPath() + "'");
+				logger.info("Jetty started on port" + (connectors.length == 1 ? "" : "s") + " "
```
wilkinsona, line 171: I think it might be nice if the use of `connectors` was kept encapsulated in `getActualPortsDescription()`. Perhaps that method could be changed to return strings like "port 8080 (http)" or "ports 8080 (http), 8443 (https)"

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
```diff
@@ -105,7 +105,8 @@ public TomcatWebServer(Tomcat tomcat, boolean autoStart, Shutdown shutdown) {
 	}
 
 	private void initialize() throws WebServerException {
-		logger.info("Tomcat initialized with port(s): " + getPortsDescription(false));
+		logger.info("Tomcat initialized with port" + (this.tomcat.getService().findConnectors().length == 1 ? "" : "s")
```
wilkinsona, line 108: Similar to my comment about Jetty, it would be good to keep the use of the connectors encapsulated in `getPortsDescription`. An added benefit would be that they would only be found once then too.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/undertow/UndertowWebServer.java
```diff
@@ -315,7 +315,7 @@ private void notifyGracefulCallback(boolean success) {
 	}
 
 	protected String getStartLogMessage() {
-		return "Undertow started on port(s) " + getPortsDescription();
+		return "Undertow started on port" + (getActualPorts().size() == 1 ? "" : "s") + " " + getPortsDescription();
```
wilkinsona, line 318: Same comment again.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
```diff
@@ -105,7 +105,8 @@ public TomcatWebServer(Tomcat tomcat, boolean autoStart, Shutdown shutdown) {
...
+				+ ": " + getPortsDescription(false));
```
wilkinsona, line 109: Tomcat's the only case where we use a `:`. As part of this, it would be good to make that consistent. I'm inclined to remove it from Tomcat's log message rather than adding one to the other servers' messages.

wilkinsona (changes requested):
Thanks for the PR, @asashour. I've left a few comments for your consideration when you have a minute.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
asashour: Thanks for the feedback, kindly have another look.

wilkinsona:
Thanks very much, @asashour.

PR #36109, opened 2023-06-29
Auto-configured JacksonJsonpMapper is conditional on an ObjectMapper bean but does not use such a bean

### LINKED ISSUE #36107: JacksonJsonpMapper bean autoconfigure condition bug
# Spring Boot version
3.1.1

# Bug description
The autoconfigured [JacksonJsonpMapper bean](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchClientConfigurations.java#L52C22-L52C40) used to depend on a global `ObjectMapper` bean before SpringBoot version v3.0.3. 

After this [commit](https://github.com/spring-projects/spring-boot/commit/3af30b0a114be586e3cdfa0d49d5c5385753c511) introduced by SpringBoot version v3.0.3, the global `ObjectMapper` bean is not required when creating `JacksonJsonpMapper` bean.

However, the `JacksonJsonpMapper` bean creation condition `@ConditionalOnBean(ObjectMapper.class)` is not changed, resulting in `JacksonJsonpMapper` bean can be created only when ObjectMapper bean is present instead of the ObjectMapper class.

# Bug resolving solution
Change `@ConditionalOnBean(ObjectMapper.class)` to `@ConditionalOnClass(ObjectMapper.class)`

Pengfei-Lu:
# Spring Boot version
3.1.1

# Bug description
The autoconfigured [JacksonJsonpMapper bean](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchClientConfigurations.java#L52C22-L52C40) used to depend on a global `ObjectMapper` bean before SpringBoot version v3.0.3. 

After this [commit](https://github.com/spring-projects/spring-boot/commit/3af30b0a114be586e3cdfa0d49d5c5385753c511) introduced by SpringBoot version v3.0.3, the global `ObjectMapper` bean is not required when creating `JacksonJsonpMapper` bean.

However, the `JacksonJsonpMapper` bean creation condition `@ConditionalOnBean(ObjectMapper.class)` is not changed, resulting in `JacksonJsonpMapper` bean can be created only when ObjectMapper bean is present instead of the ObjectMapper class.

# Bug resolving solution
Change `@ConditionalOnBean(ObjectMapper.class)` to `@ConditionalOnClass(ObjectMapper.class)`

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchClientAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/elasticsearch/ElasticsearchClientAutoConfiguration.java
@@ -23,7 +23,6 @@ import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.autoconfigure.elasticsearch.ElasticsearchClientConfigurations.ElasticsearchClientConfiguration;
 import org.springframework.boot.autoconfigure.elasticsearch.ElasticsearchClientConfigurations.ElasticsearchTransportConfiguration;
-import org.springframework.boot.autoconfigure.jackson.JacksonAutoConfiguration;
 import org.springframework.boot.autoconfigure.jsonb.JsonbAutoConfiguration;
 import org.springframework.context.annotation.Import;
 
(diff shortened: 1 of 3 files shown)


Pengfei-Lu:
I don't know how to mock the "ObjectMapper class is not loaded" senario, thus deleting the two failed `withoutJsonbOrJacksonShouldDefineSimpleMapper` & `withJsonbShouldDefineJsonbMapper` tests cases.

Is there any better solution?

wilkinsona:
You should be able to use `contextRunner.withClassLoader(new FilteredClassLoader(ObjectMapper.class))`.

Pengfei-Lu:
> You should be able to use `contextRunner.withClassLoader(new FilteredClassLoader(ObjectMapper.class))`.

It works! Please review the code.

wilkinsona (changes requested):
Thanks for the updates, @Pengfei-Lu. I think `ElasticsearchClientAutoConfiguration` could also be updated as it no longer needs to be ordered after `JacksonAutoConfiguration`. Could you do that as well please?

Pengfei-Lu:
> Thanks for the updates, @Pengfei-Lu. I think `ElasticsearchClientAutoConfiguration` could also be updated as it no longer needs to be ordered after `JacksonAutoConfiguration`. Could you do that as well please?

Done. Please check.

Pengfei-Lu:
@wilkinsona If there is any update, please let me know, thanks!

wilkinsona (approved):

wilkinsona:
Thanks very much for the PR, @Pengfei-Lu.

PR #36116, opened 2023-06-29
Support Jetty in ClientHttpRequestFactories

poutsma:
This commit introduces support for the JettyClientHttpRequestFactory in ClientHttpRequestFactories.


poutsma:
This PR uses code that will be in Spring Framework 6.1-M2, hence the snapshot version change in `gradle.properties`.

wilkinsona:
Thanks very much, @poutsma.

PR #36157, opened 2023-07-02
Fix metadata for logging.include-application-name

izeye:
This PR fixes metadata for the `logging.include-application-name` property.


philwebb:
I don't know how I manage to mess up things like this so often, but I'm very grateful we have you around to find and fix them :)

PR #36167, opened 2023-07-03
Use Stream.toList()

izeye:
This PR polishes a bit by using `Stream.toList()`.


wilkinsona:
Thanks very much once again, @izeye.

Issue #33746, opened 2023-01-10
MockitoPostProcessor doesn't check FactoryBean.OBJECT_TYPE_ATTRIBUTE correctly


philwebb:
`MockitoPostProcessor.getExistingBeans` checks the `FactoryBean.OBJECT_TYPE_ATTRIBUTE` on bean definitions but it only does a `String` match.

Framework expects a [`Class` or `ResolvableType`](https://github.com/spring-projects/spring-framework/blob/8dbdfb0449f5223e102c090591d38ed956722147/spring-beans/src/main/java/org/springframework/beans/factory/support/AbstractBeanFactory.java#L1699)

bbulgarelli:
Can I do this one?

philwebb:
You're welcome to take a look at this one @bbulgarelli. I'll assign it to you.

wilkinsona:
Thanks for the progress update.

> When I run the tests, one is always failing. I can't figure it out.

I guess it's `canMockBeanProducedByFactoryBeanWithObjectTypeAttribute` that's failing. It sets the attribute using a `String` but your changes mean that only a `Class<?>` or `ResolvableType` will be considered.

> I am also not sure if I should remove the string match or not

Please keep the string match for now. There are some other projects [that set the attribute with a String value](https://github.com/spring-projects/spring-data-commons/blob/8aff63103230780e319a34a5b95dd67cd7c90092/src/main/java/org/springframework/data/repository/config/RepositoryRegistrationAotContribution.java#L231-L232) and we should give them some time to adapt.

It would be good to update the tests so that in addition to the existing test for a `String` value, we also test the object type attribute with `Class` and `ResolvableType` values as well.

wilkinsona:
How's it going now, @bbulgarelli?

bbulgarelli:
> How's it going now, @bbulgarelli?
Sorry for keeping you waiting. I have been very busy in the past few weeks. I have created a PR, and I believe it solves the issue.

wilkinsona:
Closing in favor of #36224. Thanks for the PR, @bbulgarelli.

PR #36224, opened 2023-07-05
MockitoPostProcessor doesn't check FactoryBean.OBJECT_TYPE_ATTRIBUTE correctly

bbulgarelli:
This PR relates to the issue #33746. I added conditionals that verify if `FactoryBean.OBJECT_TYPE_ATTRIBUTE` is an instance of `Class<?>` or `ResolvableType` and I also created two more tests that test the object type attribute with `Class` and `ResolvableType` values.


wilkinsona:
Thanks very much, @bbulgarelli.

Issue #34779, opened 2023-03-27
Deprecate DelegatingApplicationContextInitializer  and DelegatingApplicationListener in favour of registering each delegate programatically or in spring.factories


kaleev:
`DelegatingApplicationContextInitializer` and `DelegatingApplicationListener` read `context.initializer.classes` and `context.listener.classes` properties to register `ApplicationContextInitializer` and `ApplicationListener`. `additional-spring-configuration-metadata.json` should contain metadata for these properties in order to provide proper editor support for these properties in IDEs.
Spring Boot 3.0 still doesn't contain the metadata.

wilkinsona:
Thanks for the suggestion but I am not sure that those properties should be documented in the metadata as they have no effect unless you've configured your application to use `DelegatingApplicationContextInitializer` or `DelegatingApplicationListener`.

I'm wondering if we should deprecate those `Delegating…` classes. I'm not sure they offer much value and, unlike listing classes directly in `spring.factories`, they won't work in a native image without additional reflection hints.

wilkinsona:
We discussed this today and decided to deprecate `DelegatingApplicationContextInitializer` and `DelegatingApplicationListener`.

snicoll:
Forgot to close this as superseded by the PR #36251

PR #36251, opened 2023-07-05
Deprecate DelegatingApplicationContextInitializer and DelegatingApplicationListener in favour of registering each delegate programatically or in spring.factories

bbulgarelli:
This PR is related to the issue #34779. I did what the issue suggested and deprecated `DelegatingApplicationContextInitializer` and `DelegatingApplicationListener`.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/DelegatingApplicationContextInitializer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/DelegatingApplicationContextInitializer.java
@@ -39,6 +39,7 @@ import org.springframework.util.StringUtils;
  * @author Phillip Webb
  * @since 1.0.0
  */
+@Deprecated
 public class DelegatingApplicationContextInitializer
 		implements ApplicationContextInitializer<ConfigurableApplicationContext>, Ordered {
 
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks for the PR. We usually also add `@Deprecated` javadoc in the form described [here](https://github.com/spring-projects/spring-boot/wiki/Deprecations). We probably want something like "@deprecated since 3.2 for removal in 3.5 as property based initialization is not longer recommended".

wilkinsona:
We also use the `since` and `forRemoval` attributes on `@Deprecated`:

```
@Deprecated(since = "3.2.0", forRemoval = true)
```

bbulgarelli:
> @bbulgarelli can you please run the build before submitting a pull request? 
Sorry about that. I tried to run it this time, but I still got warnings related to the test classes, even though I already added the deprecation annotation to both test classes. I'm not sure what I'm doing wrong.


> We probably want something like `@deprecated since 3.2 for removal in 3.4 as property based initialization is not longer recommended`
OK. I added this to both the deprecated classes and the test classes.

snicoll:
Thanks @bbulgarelli, I've polished your contribution in [60df7e3](https://github.com/spring-projects/spring-boot/commit/60df7e3bce6873a21351e2164e835233c7e3159f).

izeye:
> We also use the `since` and `forRemoval` attributes on `@Deprecated`:
> 
> ```
> @Deprecated(since = "3.2.0", forRemoval = true)
> ```

It would be nice if it's also documented in [the "Deprecations" section](https://github.com/spring-projects/spring-boot/wiki/Deprecations) in the wiki.

snicoll:
Thanks @izeye, I've updated the wiki.

rishiraj88:
Nice idea, @izeye

Issue #34062, opened 2023-02-04
Tighten rules around profile naming


onobc:
If you set `spring.profiles.active` to a value that contains a `*` the app will fail to start. 

### Steps to reproduce

Spring Boot `3.0.2` simplest application from start.spring.io:
```java
@SpringBootApplication
public class WackyProfileApplication {
	public static void main(String[] args) {
		SpringApplication.run(WackyProfileApplication.class, args);
	}
}
```

Start it w/ `spring.profiles.active` set to `foo*bar` via any normal means such as: 
- sysprop
- envvar
- property in application.yml
- profile property in maven/gradle plugin

App fails to start and reports the following in the log:
```
com.example.wackyprofile.WackyprofileApplication
23:14:37.244 [main] ERROR org.springframework.boot.SpringApplication - Application run failed
java.lang.IllegalStateException: Location 'file:./application-foo*bar.yaml' must end with '*/'
	at org.springframework.util.Assert.state(Assert.java:97)
	at org.springframework.boot.context.config.LocationResourceLoader.validatePattern(LocationResourceLoader.java:134)
	at org.springframework.boot.context.config.LocationResourceLoader.getResources(LocationResourceLoader.java:95)
	at org.springframework.boot.context.config.StandardConfigDataLocationResolver.resolvePattern(StandardConfigDataLocationResolver.java:313)
	at org.springframework.boot.context.config.StandardConfigDataLocationResolver.resolve(StandardConfigDataLocationResolver.java:299)
	at org.springframework.boot.context.config.StandardConfigDataLocationResolver.resolve(StandardConfigDataLocationResolver.java:251)
	at org.springframework.boot.context.config.StandardConfigDataLocationResolver.resolveProfileSpecific(StandardConfigDataLocationResolver.java:150)
	at org.springframework.boot.context.config.ConfigDataLocationResolvers.lambda$resolve$2(ConfigDataLocationResolvers.java:107)
	at org.springframework.boot.context.config.ConfigDataLocationResolvers.resolve(ConfigDataLocationResolvers.java:113)
	at org.springframework.boot.context.config.ConfigDataLocationResolvers.resolve(ConfigDataLocationResolvers.java:106)
	at org.springframework.boot.context.config.ConfigDataLocationResolvers.resolve(ConfigDataLocationResolvers.java:94)
	at org.springframework.boot.context.config.ConfigDataImporter.resolve(ConfigDataImporter.java:105)
[...]

PR #36262, opened 2023-07-06
Checking for invalid characters in SpringBootContextLoader.setActiveProfiles

bbulgarelli:
This PR relates to #34062. I created a loop at the beginning of the method that goes through the list of profiles and checks if there are any of these invalid characters: `*` `&` `!` `|` `,`. If so, an exception is thrown informing which profile is invalid and which characters are prohibited. 
I also created the necessary tests to verify the method changes.


mhalbritter:
Hey, thanks for that. What about the profile names `(` and `)`? I think they should be prohibited, too.

mentallurg (commented):
This pull request has multiple problems.

1) This pull requests addresses a particular case only. What about ':', ';', '/', '\', '?'? I suggest completely different approach: Allow letters, digits and hyphen only. Use functions `Character.isAlphabetic()`, `Character.isDigit()`.

2) Only a single case of profile usage is addressed, the list of active profile. But there are also other cases, when profile names are used, for instance, in the `@Profile` annotation. The same validation should be used everywhere.

3) Documentation should be extended to describe what profile names will be considered valid.

PR #36278, opened 2023-07-07
smoke test for WebFlux gzip

sullis:
add smoke test for WebFlux "Accept-Encoding: gzip"


wilkinsona:
Thanks for the proposal but I don't think the additional tests are necessary. We already [have tests for compression with reactive web servers](https://github.com/spring-projects/spring-boot/blob/3b2551619f4ac74aac238cf3fa34501088240924/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/reactive/server/AbstractReactiveWebServerFactoryTests.java#L374-L430) in the `spring-boot` module.

PR #36288, opened 2023-07-10
Order ProblemDetailsExceptionHandler beans

mzeijen:
In this pull request I added an explicit zero order to the ` ProblemDetailsExceptionHandler` beans for WebMVC and Webflux.

Without the order, it means these beans automatically have the lowest precedence, and they will usually be executed last. This can be problematic when someone would like to create a “catch-all” `@ExceptionHandler` in a `@ControllerAdvice` bean. Meaning an exception handler that should handle any exception that was not handled by any other exception handler with.
Such an exception handler should have the absolute lowest precedence, and the `ProblemDetailsExceptionHandler` must have a higher precedence. However, because both now have the same precedence, it is not a guarantee that this will be the case. In my tests, my own catch-all exception handler would always start handling those exceptions that should be handled by the `ProblemDetailsExceptionHandler`.

The workaround is to create custom beans that implement `ResponseEntityExceptionHandler` and give those an order that has a higher precedence then my catch-all exception handler. However, I don’t think this should be necessary.

Because the `ResponseEntityExceptionHandler` has a defined list of exceptions it is going to handle, there should be no issue in it having a specific order by default. If it has the order `0` then anyone can easily create an exception handler that has a higher or lower precedence. 

I created a test which shows the issue, and the current workaround:

https://github.com/mzeijen/spring-boot-problem-support/blob/main/src/test/java/com/example/demo/ResponseEntityExceptionHandlerOrderingTest.java

This test has nested test classes which show the difference in behavior for both WebMVC and Webflux, when the default `ResponseEntityExceptionHandler` are used or when the workaround is applied.

In the pull request I added a test that verifies that the ordering takes effect.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
@@ -343,6 +343,7 @@ public class WebFluxAutoConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean(ResponseEntityExceptionHandler.class)
+		@Order(0)
 		ProblemDetailsExceptionHandler problemDetailsExceptionHandler() {
 			return new ProblemDetailsExceptionHandler();
 		}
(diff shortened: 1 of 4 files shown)


wilkinsona:
WDYT, @bclozel? This seems sensible to me but there's a small change that it will be a breaking change. As such, I'd target at 3.2.x.

snicoll:
Thank you @mzeijen!

PR #36422, opened 2023-07-17
Make HikariDataSource participate in checkpoint-restore

christophstrobl:
This PR allows to suspend and gracefully shut down the connection pool of a `HikariDataSource` within the container lifecycle phases.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/HikariLifecycle.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/HikariLifecycle.java
@@ -1,0 +1,221 @@
+/*
+ * Copyright 2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.jdbc;
+
+import java.lang.reflect.Field;
+import java.time.Duration;
+import java.util.concurrent.CompletableFuture;
+import java.util.concurrent.ExecutionException;
+import java.util.concurrent.ThreadPoolExecutor;
+import java.util.concurrent.TimeUnit;
+import java.util.concurrent.TimeoutException;
+import java.util.function.Supplier;
+
+import com.zaxxer.hikari.HikariDataSource;
+import com.zaxxer.hikari.HikariPoolMXBean;
+import com.zaxxer.hikari.pool.HikariPool;
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
+import org.springframework.aop.TargetSource;
+import org.springframework.aop.framework.Advised;
+import org.springframework.aop.support.AopUtils;
+import org.springframework.context.Lifecycle;
+import org.springframework.util.Assert;
+import org.springframework.util.ReflectionUtils;
+
+/**
+ * {@link Lifecycle} management for a {@link HikariDataSource} taking care of
+ * {@link Lifecycle#start() starting}/{@link Lifecycle#stop() stopping} the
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/HikariLifecycle.java
```diff
@@ -0,0 +1,221 @@
...
+		private Supplier<Boolean> hasOpenConnections;
+
+		LifecycleExecutor(HikariDataSource hikariDataSource) {
+
+			this.dataSource = getUltimateTargetObject(hikariDataSource);
+
+			if (hikariDataSource.getHikariPoolMXBean() instanceof HikariPool pool) {
+
+				Field closeConnectionExecutor = ReflectionUtils.findField(HikariPool.class, "closeConnectionExecutor");
+				if (closeConnectionExecutor != null) {
```
wilkinsona, line 115: If the field's not found something's gone really wrong. I think it would be better to fail hard.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/HikariLifecycle.java
```diff
@@ -0,0 +1,221 @@
...
+
+					ReflectionUtils.makeAccessible(closeConnectionExecutor);
+					Object field = ReflectionUtils.getField(closeConnectionExecutor, pool);
+					if (field instanceof ThreadPoolExecutor executor) {
```
wilkinsona, line 119: Won't this always be a `ThreadPoolExecutor`?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/HikariLifecycle.java
```diff
@@ -0,0 +1,221 @@
...
+						this.hasOpenConnections = () -> executor.getActiveCount() > 0;
+					}
+				}
+			}
+			if (this.hasOpenConnections == null) {
```
wilkinsona, line 124: Wondering about thread-safety here. Ideally `hasOpenConnections` would be `final`.

wilkinsona (changes requested):
Thanks for the PR, @christophstrobl. I'm still not sure about the reflection, but ignoring that for the time being I think there are a few refinements that could be made. WDYT?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/HikariLifecycle.java
christophstrobl: The idea was to be very defensive here, and fall back to checking _total connections_ if for whatever reason `HikariPool` does not meet the expectations.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/HikariLifecycle.java
christophstrobl: Same defensive strategy as with the field above. 
I looked up the history of `HikaryPool` and it seems that part did not change too much in the past. Will take care of this.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/HikariLifecycle.java
wilkinsona: If Hikari doesn't meet expectations, then I think we should just give up as we have no way of knowing that our approach is valid. We'll know, through tests, that it works with the version of Hikari in dependency management. If someone overrides that version and Hikari has changed incompatibility I think we'd be better to fail fast. That'll make the problem much easier to diagnose rather than dealing with a checkpoint failure and trying to figure out why it occurred.

christophstrobl:
@wilkinsona rebased and now failing hard when conditions are not met. I had to delay the reflective lookup to pass all tests. So initially there'll only be a field and type check, but the `HikariPool` instance check of `HikariPoolMXBean` will happen later. Also made sure to use `DataSourceUnwrapper` to get the target of a  proxy/delegate instance.

wilkinsona:
Thanks for the updates, @christophstrobl. We discussed this today and decided that we're happy with the reflection as it'll only come into effect when checkpoint-restore is being used. Would you like to take the PR out of draft status?

wilkinsona:
Thanks very much, @christophstrobl.

PR #36463, opened 2023-07-20
Harmonize use of Stream in ConfigDataLocationBindHandler

elevne:
Changed a little bit of code in ConfigDataLocationBindHandler.onSuccess method. I thought there's no reason to separate Stream and for logic, and that this way would be better to read.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataLocationBindHandler.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataLocationBindHandler.java
@@ -57,13 +57,10 @@ class ConfigDataLocationBindHandler extends AbstractBindHandler {
 			return list;
 		}
 		if (result instanceof ConfigDataLocation[] unfilteredLocations) {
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR @elevne but I believe the code is fine as it is and we prefer to do such stylistic code change ourselves.

snicoll:
@wilkinsona made me aware of the use of Stream in the same class and that this part of the code is a little inconsistent. That's different from the stylistic change I was mentioning early so I am reopening this one.

wilkinsona:
I'd like to reconsider this one as I don't think the change is purely stylistic. I like the fact that the proposal makes the code more consistent as it does everything using the stream rather than splitting things across a stream and a for-loop.

mhalbritter:
I like the changes. I find the existing code somewhat odd, as it starts out with a stream and then updates the array via a for loop. The new code is more consistent.

elevne:
Thank you for all your comments! I checked the whole body of method again, and thought it would be more consistent to change to streamify upper part as well.

snicoll:
@elevne thanks very much for making your first contribution to Spring Boot.

PR #36477, opened 2023-07-20
Documentation of spring.redis.url incorrectly states that it does not override spring.redis.user

izeye:
This PR updates Javadoc for the `RedisProperties.url` field to include username in override targets as it seems to override it.

[A Redis URI seems to be able to include other properties like database](https://github.com/lettuce-io/lettuce-core/wiki/Redis-URI-and-connection-details#uri-syntax), but they are not included in override targets. It seems to be inconsistent, but I'm not sure if it's intentional.


wilkinsona:
When merging forwards, the property becomes `spring.data.redis.url`.

PR #36507, opened 2023-07-23
Descriptions of started and ready time metrics contain time units but the unit may change when the metrics are exported

fzyzcjy:
When using this metric in Grafana/Prometheus, it is a bit confusing, because as the screenshot below shows, it is `application_started_time_seconds` (ends with "seconds"), and the value is really measured in seconds, but the hint text say it is "ms" not "s".

I guess this is because micrometer is smart enough to convert ms to s, since spring already tells micrometer it is ms.

image


snicoll:
@fzyzcjy good catch, thanks for making your first contribution to Spring Boot.

PR #36509, opened 2023-07-24
Harmonize Stream.collect() usage

quaff:
use `Stream.toList()` or `collect(Collectors.toUnmodifiable*())` if returned Collection should be immutable.

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/bom/bomr/StandardLibraryUpdateResolver.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/bom/bomr/StandardLibraryUpdateResolver.java
@@ -19,12 +19,10 @@ package org.springframework.boot.build.bom.bomr;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Collection;
-import java.util.Collections;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.SortedSet;
-import java.util.stream.Collectors;
 
 import org.apache.maven.artifact.versioning.DefaultArtifactVersion;
 import org.slf4j.Logger;
@@ -98,19 +96,14 @@ class StandardLibraryUpdateResolver implements LibraryUpdateResolver {
 						getLaterVersionsForModule(group.getId(), plugin, libraryVersion));
 			}
 		}
-		List<DependencyVersion> allVersions = moduleVersions.values()
+		return moduleVersions.values()
 			.stream()
 			.flatMap(SortedSet::stream)
 			.distinct()
 			.filter((dependencyVersion) -> isPermitted(dependencyVersion, library.getProhibitedVersions()))
-			.toList();
-		if (allVersions.isEmpty()) {
-			return Collections.emptyList();
-		}
-		return allVersions.stream()
-			.map((version) -> new VersionOption.ResolvedVersionOption(version,
+			.map((version) -> (VersionOption) new VersionOption.ResolvedVersionOption(version,
 					getMissingModules(moduleVersions, version)))
-			.collect(Collectors.toList());
+			.toList();
 	}
 
 	private boolean isPermitted(DependencyVersion dependencyVersion, List<ProhibitedVersion> prohibitedVersions) {
--- a/ci/images/releasescripts/src/test/java/io/spring/concourse/releasescripts/sonatype/SonatypeServiceTests.java
+++ b/ci/images/releasescripts/src/test/java/io/spring/concourse/releasescripts/sonatype/SonatypeServiceTests.java
@@ -21,7 +21,6 @@ import java.io.IOException;
 import java.nio.file.Files;
 import java.nio.file.Path;
 import java.util.Arrays;
-import java.util.HashSet;
 import java.util.Iterator;
 import java.util.Set;
 import java.util.stream.Collectors;
@@ -105,7 +104,7 @@ class SonatypeServiceTests {
 					.filter((artifact) -> !artifact.startsWith("build-info.json"))
 					.map((artifact) -> requestTo(
 							"/service/local/staging/deployByRepositoryId/example-6789/" + artifact.toString()))
-					.collect(Collectors.toCollection(HashSet::new));
+					.collect(Collectors.toSet());
 			AnyOfRequestMatcher uploadRequestsMatcher = anyOf(uploads);
 			assertThat(uploadRequestsMatcher.candidates).hasSize(150);
 			this.server.expect(ExpectedCount.times(150), uploadRequestsMatcher).andExpect(method(HttpMethod.PUT))
@@ -157,7 +156,7 @@ class SonatypeServiceTests {
 					.filter((artifact) -> !"build-info.json".equals(artifact.toString()))
 					.map((artifact) -> requestTo(
 							"/service/local/staging/deployByRepositoryId/example-6789/" + artifact.toString()))
-					.collect(Collectors.toCollection(HashSet::new));
+					.collect(Collectors.toSet());
 			AnyOfRequestMatcher uploadRequestsMatcher = anyOf(uploads);
 			assertThat(uploadRequestsMatcher.candidates).hasSize(150);
 			this.server.expect(ExpectedCount.times(150), uploadRequestsMatcher).andExpect(method(HttpMethod.PUT))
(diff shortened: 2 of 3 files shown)


philwebb:
We unfortunately can't do a global search/replace with these since `toList()` returns an unmodifiable list where as `toCollection(ArrayList::new)` does not. Looking at the changes I think at least https://github.com/spring-projects/spring-boot/pull/36509/files#diff-0a3440d22ed49ae2bf92694192d1a9de54c92e6c1b8cdcacd2ac7919c5f229bdR61 will fail.

It's certainly work us seeking more of these out, but we'll have to be quite careful about where we do it.

quaff:
> We unfortunately can't do a global search/replace with these since `toList()` returns an unmodifiable list where as `toCollection(ArrayList::new)` does not. Looking at the changes I think at least https://github.com/spring-projects/spring-boot/pull/36509/files#diff-0a3440d22ed49ae2bf92694192d1a9de54c92e6c1b8cdcacd2ac7919c5f229bdR61 will fail.
> 
> It's certainly work us seeking more of these out, but we'll have to be quite careful about where we do it.

`Collector.toList()` is using `ArrayList::new` internally, It will return an mutable list, am I missing something?

wilkinsona:
That behavior [isn't guaranteed](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toList--):

> There are no guarantees on the type, mutability, serializability, or thread-safety of the List returned; if more control over the returned List is required, use [toCollection(Supplier)](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toCollection-java.util.function.Supplier-).

As such, we can't rely upon it and need to use the Supplier variant where, for example, we require the result to be mutable.

quaff:
> That behavior [isn't guaranteed](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toList--):
> 
> > There are no guarantees on the type, mutability, serializability, or thread-safety of the List returned; if more control over the returned List is required, use [toCollection(Supplier)](https://docs.oracle.com/javase/8/docs/api/java/util/stream/Collectors.html#toCollection-java.util.function.Supplier-).
> 
> As such, we can't rely upon it and need to use the Supplier variant where, for example, we require the result to be mutable.

So we should keep mutable List as it is, and others refactor to `Collectors.toUnmodifiableList()` ensure immutability, WDYT?

wilkinsona:
In 3.0.x and later branches, `Stream.toList()` should be used the vast majority of the time. Similarly, in 2.7.x `Collectors.toList()` should be used the vast majority of the time. On any maintained branch where mutability is required, `Collectors.toList(ArrayList::new)` should be used.

quaff:
> In 3.0.x and later branches, `Stream.toList()` should be used the vast majority of the time. Similarly, in 2.7.x `Collectors.toList()` should be used the vast majority of the time. On any maintained branch where mutability is required, `Collectors.toList(ArrayList::new)` should be used.

I've updated this PR as you suggested @wilkinsona

ci/images/releasescripts/src/test/java/io/spring/concourse/releasescripts/sonatype/SonatypeServiceTests.java
```diff
@@ -105,7 +104,7 @@ void publishWithSuccessfulClose() throws IOException {
 					.filter((artifact) -> !artifact.startsWith("build-info.json"))
 					.map((artifact) -> requestTo(
 							"/service/local/staging/deployByRepositoryId/example-6789/" + artifact.toString()))
-					.collect(Collectors.toCollection(HashSet::new));
+					.collect(Collectors.toUnmodifiableSet());
```
wilkinsona, line 107: `Collectors.toSet()`

ci/images/releasescripts/src/test/java/io/spring/concourse/releasescripts/sonatype/SonatypeServiceTests.java
```diff
@@ -157,7 +156,7 @@ void publishWithCloseFailureDueToRuleViolations() throws IOException {
 					.filter((artifact) -> !"build-info.json".equals(artifact.toString()))
 					.map((artifact) -> requestTo(
 							"/service/local/staging/deployByRepositoryId/example-6789/" + artifact.toString()))
-					.collect(Collectors.toCollection(HashSet::new));
+					.collect(Collectors.toUnmodifiableSet());
```
wilkinsona, line 159: `Collectors.toSet()`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurations.java
```diff
@@ -51,7 +50,7 @@ protected Collection<Class<?>> sort(Collection<Class<?>> classes) {
 		List<String> sorted = SORTER.getInPriorityOrder(names);
 		return sorted.stream()
 			.map((className) -> ClassUtils.resolveClassName(className, null))
-			.collect(Collectors.toCollection(ArrayList::new));
+			.collect(Collectors.toUnmodifiableList());
```
wilkinsona, line 53: If we're going to change this, it should use `Stream`'s `toList()` rather than `collect`. That said, it's technically public API so there's a (very slight) chance of a breaking change and I'm not sure it's worth changing at all.

wilkinsona (changes requested):

buildSrc/src/main/java/org/springframework/boot/build/bom/bomr/StandardLibraryUpdateResolver.java
```diff
@@ -98,19 +97,14 @@ private List<VersionOption> determineResolvedVersionOptions(Library library) {
...
 			.filter((dependencyVersion) -> isPermitted(dependencyVersion, library.getProhibitedVersions()))
-			.toList();
-		if (allVersions.isEmpty()) {
-			return Collections.emptyList();
-		}
-		return allVersions.stream()
 			.map((version) -> new VersionOption.ResolvedVersionOption(version,
 					getMissingModules(moduleVersions, version)))
-			.collect(Collectors.toList());
+			.collect(Collectors.toUnmodifiableList());
```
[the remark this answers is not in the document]
quaff: `Stream.toList()` can't compile here.

ci/images/releasescripts/src/test/java/io/spring/concourse/releasescripts/sonatype/SonatypeServiceTests.java
quaff: We don't need mutable Set here, I think `toUnmodifiableSet()` is more appropriate

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurations.java
quaff: `Stream.toList()` can't compile here.

buildSrc/src/main/java/org/springframework/boot/build/bom/bomr/StandardLibraryUpdateResolver.java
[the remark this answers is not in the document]
wilkinsona: 🤔 

It should compile if the map step casts to `VersionOption`.

ci/images/releasescripts/src/test/java/io/spring/concourse/releasescripts/sonatype/SonatypeServiceTests.java
wilkinsona: We disagree. Please use `toSet()`.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/AutoConfigurations.java
quaff: PR updated, I reverted this change.

wilkinsona: Gah, generics are a pain. Please revert this change.

quaff: Already reverted.

PR #36538, opened 2023-07-24
Upgrade Spring AMQP to 3.0.7

garyrussell:


snicoll:
@garyrussell thanks but we don't need PR for dependency upgrades.

garyrussell:
I understand you don't normally, but I needed this for https://github.com/spring-projects/spring-boot/pull/36539 and didn't think you would accept a dependency upgrade as part of a PR.

snicoll:
Right. I've ran bomr to get this one done and I'll look at the PR shortly. Thanks!

PR #36564, opened 2023-07-25
Document that server.forward-headers-strategy property defaults to native when running on Kubernetes

MahatmaFatalError:
I found that my app on k8s has `server.forward-headers-strategy` NATIVE but the docs do not state so.
Hence I propose this quick fix.
Please confirm, I could not find the particular code line to control that behavior.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/webserver.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/webserver.adoc
@@ -443,7 +443,7 @@ You can register it as a servlet filter in your application by setting `server.f
 TIP: If you are using Tomcat and terminating SSL at the proxy, configprop:server.tomcat.redirect-context-root[] should be set to `false`.
 This allows the `X-Forwarded-Proto` header to be honored before any redirects are performed.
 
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/webserver.adoc
```diff
@@ -443,7 +443,7 @@ You can register it as a servlet filter in your application by setting `server.f
 TIP: If you are using Tomcat and terminating SSL at the proxy, configprop:server.tomcat.redirect-context-root[] should be set to `false`.
 This allows the `X-Forwarded-Proto` header to be honored before any redirects are performed.
 
-NOTE: If your application runs in Cloud Foundry or Heroku, the configprop:server.forward-headers-strategy[] property defaults to `NATIVE`.
+NOTE: If your application runs in Cloud Foundry, Heroku or K8s, the configprop:server.forward-headers-strategy[] property defaults to `NATIVE`.
```
wilkinsona, line 446: Thanks for the suggestion. I think it would be better to use Kubernetes here rather than the K8S abbreviation.

wilkinsona (changes requested):

wilkinsona (approved):
Looks good to me. Thanks for the update, @MahatmaFatalError. Can you please remove the draft status from the PR?

mhalbritter:

PR #36572, opened 2023-07-25
Tomcat warns about a missing +/- prefix when enabling multiple protocols through server.ssl.enabled-protocols

chicobento:
When configuring SSL with Tomcat, the following message is being logged in warning level by [SSLHostConfig](https://github.com/apache/tomcat/blob/a2615d18d686af9a17376017bad8dde22678f3a9/java/org/apache/tomcat/util/net/SSLHostConfig.java#L482):

`"The protocol [{0}] was added to the list of protocols on the SSLHostConfig named [{1}]. Check if a +/- prefix is missing."`

This is because Tomcat's [SSLHostConfig semantics](https://github.com/apache/tomcat/blob/a2615d18d686af9a17376017bad8dde22678f3a9/java/org/apache/tomcat/util/net/SSLHostConfig.java#L446) for protocol string is:
```
        // List of protocol names, separated by ",", "+" or "-".
        // Semantics is adding ("+") or removing ("-") from left
        // to right, starting with an empty protocol set.
        // Tokens are individual protocol names or "all" for a
        // default set of supported protocols.
        // Separator "," is only kept for compatibility and has the
        // same semantics as "+", except that it warns about a potentially
        // missing "+" or "-".
```


mhalbritter:

PR #36604, opened 2023-07-27
Reinstate use of the configprop macro to reference management.observations.enable.spring.security

quaff:
`configprop:` prefix is intentional here.


wilkinsona:
@snicoll just pointed out that our Asciidoctor extension that checks the validity of configuration properties doesn't render `configprop:management.observations.enable.spring.security[]` correctly. It appears as `management.observations.enable` in the docs, losing the fact that `spring.security` is being place in the `management.observations.enable` map. We'll need to fix the extension before this can be merged.

wilkinsona:
Thanks very much, @quaff.

Issue #35892, opened 2023-06-14
Implement infrastructure for virtual threads


mhalbritter:
We need at least:

* a new enum `TWENTY_ONE` in `JavaVersion`
* a `@ConditionalOnVirtualThreads` which can be used in out auto-configurations

mhalbritter:
We removed that condition. It's replaced by `@ConditionalOnThreading(Threading.VIRTUAL)`, see #36624.

PR #36607, opened 2023-07-27
Add Javadoc since for JavaVersion.TWENTY_ONE

izeye:
This PR adds a Javadoc `@since` tag for the `JavaVersion.TWENTY_ONE`.

See gh-35892


wilkinsona:
Nearer the release of Java 21, we'll add `TWENTY_ONE` to the enum in 2.7.x. With the final RC of Java 21 on 24 August, the same day as 2.7.15 is released, that'll probably happen in 2.7.16. We can add `@since` at that point. This will be before 3.2 GAs so I don't think there's much point in applying this change now. Thanks anyway.

I've also opened #36608 as many of the existing values are missing `@since` in their javadoc.

PR #36617, opened 2023-07-27
Enable debug option for Kotlin in Gradle Plugin tests

dreis2211:
Hi,

this PR enables `GradleRunner.withDebug(true)` for Kotlin tests as https://github.com/gradle/gradle/issues/6862 seems to be fixed by now. 

https://github.com/gradle/gradle/issues/14125 on the other hand still seems to be broken, so I updated the comment accordingly.

Cheers,
Christoph


wilkinsona:
Thanks very much, @dreis2211.

PR #36767, opened 2023-08-06
Artemis ConnectionFactory is not configured when CachingConnectionFactory is missing and enabled properties are false

eddumelendez:
Currently, if `CachingConnectionFactory` is excluded no ConnectionFactory is
created.


eddumelendez:
Issue #17483 fixed it in ActiveMQ

snicoll:
Good catch, thanks @eddumelendez!

PR #36778, opened 2023-08-08
Tune order of built-in Customizers

quaff:
Leave some space let application Customizers executed after built-in Customizers.
It's arguable to use `Ordered.LOWEST_PRECEDENCE - 10` instead of `0`, the advantage is keeping backward compatibility as possible.


wilkinsona:
Thanks for the proposal but I don't think we should do this. There's quite some risk of a breaking change and until someone raises a requirement to be able to override a particular customizer, I don't think that risk can be justified.

Issue #36909, opened 2023-08-10
BindableRuntimeHintsRegistrar does not generate hints for all nested types of a ConfigurationProperties type


snicoll:
We still need to figure it out but it looks like the nested class detection can be faulty. This can be reproduced by running `BindableRuntimeHintsRegistrar` against `RabbitProperties`, there is no metadata for `org.springframework.boot.autoconfigure.amqp.RabbitProperties$SimpleContainer`.

omercelikceng:
@snicoll. Hello, thank you very much for your development. A truly useful and important development. However, there are some points I don't understand. I'm a newbie at this and I'm trying to learn. Please help me understand.

Your development solves the problem of the classes we register using the BindableRuntimeHintsRegistrar class. However, when I examined the RabbitProperties class, it was not registered using the BindableRuntimeHintsRegistrar class. The annotations that enable the RabbitProperties class to be registered are @ConfigurationProperties and @EnableConfigurationProperties.

Then I read the spring documentation. I read that configuration classes are automatically saved in reflection hints.(@ConfigurationProperties) But nested configurations cannot be registered for configuration classes. In order to register nested classes, we need to add the @NestedConfigurationProperty annotation.

**Documentation Link :** https://docs.spring.io/spring-boot/docs/current/reference/html/native-image.html#native-hints 

**Documentation Photo :** 
![NestedProperty](https://github.com/spring-projects/spring-boot/assets/34758411/5c33e55c-4254-4202-9868-07a403e8dcad)

Did I understand correctly? You solved the problems in the classes we registered using the BindableRuntimeHintsRegistrar class. But you haven't fixed the issues for RabbitProperties. Am I thinking right? Could you please help me?

omercelikceng:
@snicoll I reviewed class ConfigurationPropertiesBeanFactoryInitializationAotProcessor. And I understood what you said. Thank you. But then isn't the document wrong?

The Spring documentation says: . _Nested configuration properties which are not inner classes, however, must be annotated with @NestedConfigurationProperty, otherwise they won't be detected and will not be bindable._

But with the update you made, nested configurations are automatically registered.

snicoll:
The documentation states the exact same thing.  Those are inner classes don’t they?

omercelikceng:
[...]

PR #36956, opened 2023-08-14
Fix issuer #36909

markswell:
Add simpleContainer to properties files as:
spring.rabbitmq.simpleContainer


philwebb:
Thanks for the PR but I don't think this is the fix we need for #36909. The `SimpleContainer` property is available via `spring.rabbitmq.listener.simple` and I don't think we need to add it directly to `RabbitProperties`.

PR #37013, opened 2023-08-16
Remove unnecessary null check

ramilS:


snicoll:
Thanks for the PR but I prefer avoid doing this until we have #10712 in place. We don't automate those checks at the moment and I prefer that to be the case before we start reviewing our code base consistently.

PR #37017, opened 2023-08-17
 Change log messages to use the singular or plural forms instead of "noun(s)"

asashour:
This changes the log to use the singular form or the plural one based on the number of endpoints.

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/bom/bomr/UpgradeDependencies.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/bom/bomr/UpgradeDependencies.java
@@ -160,8 +160,9 @@ public abstract class UpgradeDependencies extends DefaultTask {
 		if (!availableLabels.containsAll(issueLabels)) {
 			List<String> unknownLabels = new ArrayList<>(issueLabels);
 			unknownLabels.removeAll(availableLabels);
+			String suffix = (unknownLabels.size() == 1) ? "" : "s";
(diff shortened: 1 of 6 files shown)


snicoll:
@asashour thanks. This is the second PR of that nature from you (#36103). I think it would be better to consider the entire code base and do this consistently. Also, please take the review of the previous PRs into account.

Would you amend your PR to replace those usages consistently?

jonatan-ivanov:
I've missed https://github.com/spring-projects/spring-boot/pull/36103 but if I can share my two cents:
I think this is definitely better for humans to read but if someone wants to process these log entries and if the log messages vary based on some condition on the producer side (endpoint/endpoints), they might need to "reimplement" that logic on the consumer side. E.g.: let's say you want to match these lines with regex, after this change you need to write a slightly more complicated one.

asashour:
Thanks for feedback. I hope all cases are covered, as checkout on Windows fails, and a manual search/edit on Ubuntu was done.

It is true that parsing the log messages would be a little more difficult, but I think the use case of logs are meant more for humans, and having a better language overweight the effort for the less likely message processing.

wilkinsona:
We discussed this a few months ago and, just to capture our discussion while I can still remember it, I wanted to note that we did not agree that this change would make the log messages harder to parse. With these changes, the singular or plural will tell you if there's one or many values to come. Previously that was unknown and any parsing had to cope with both.

wilkinsona:
Thanks very much, @asashour.

Issue #37008, opened 2023-08-16
Logging config URLs with query parameters are not detected as XML


nikhilraj6692:
While upgrading to Spring boot 3.0.x, if consul contains below property:
```
logging:
  config: ${spring.cloud.consul.scheme}://${spring.cloud.consul.host}:${spring.cloud.consul.port}/v1/kv/${spring.cloud.consul.config.prefix}/logger.xml?raw=true&xml
```

it works till spring boot 3.0.4, since LogbackLoggingSystem class expects url to end with "xml"
![image](https://github.com/spring-projects/spring-boot/assets/49581523/68400d7e-4250-4fa3-98e3-e3e718b4ac82)

But, with versions above 3.0.4, this has changed and expects logging url to end with ".xml"
![image](https://github.com/spring-projects/spring-boot/assets/49581523/dfbff088-76ef-445d-addb-eed46f6a6bf5)

The problem is that the issue could have been found easily, if it would have printed in logs. But with consul config, nothing prints in logs when this error occurs. Please document this in spring boot 3.0.4+ changes

snicoll:
>  But with consul config, nothing prints in logs when this error occurs.

I'd argue you should be reporting the issue to Consul. There's nothing that we can do here I am afraid. Recent version of logback only offers XML-based configuration so we just adapted to that.

quaff:
@snicoll I think it's a regression, `/logger.xml?raw=true&xml` is xml.

philwebb:
I think I agree, but I think we should look at fixing the detection logic rather than documenting the limitation.

scottfrederick:
Closing in favor of #37039

PR #37039, opened 2023-08-21
Logging config URLs with query parameters are not detected as XML

quaff:
Fix GH-37008


quaff:
cc @philwebb

philwebb:
Thanks @quaff!

PR #37124, opened 2023-08-29
Tomcat Version updated to 10.1.13 due to CVE-2023-41080

gterminator:
Hi,
I know this is against the TOS.
Due to I don't know when the next Spring boot minor release is. Here is the fix for CVE-2023-41080.
Which is in fact only a version change.

regards
Max

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -7,6 +7,6 @@ org.gradle.jvmargs=-Xmx2g -Dfile.encoding=UTF-8
 kotlinVersion=1.9.10
 nativeBuildToolsVersion=0.9.24
 springFrameworkVersion=6.1.0-M4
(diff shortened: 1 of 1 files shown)


bclozel:
@gterminator next minors are already scheduled on [our milestone page](https://github.com/spring-projects/spring-boot/milestones) and [public release calendar](https://calendar.spring.io/). As usual, obvious dependency upgrades (which require no code change) will be applied with [our semi-automated process](https://github.com/spring-projects/spring-boot/wiki/Team-Practices#third-party-dependency-upgrades).

> I know this is against the TOS.

Then why submit a PR? Unless this PR contributes code changes, this is not making the upgrade easier for us.

PR #37167, opened 2023-09-01
Use project.getLayout().getBuildDirectory() instead of project.getBuildDir()

quaff:
`Project.getBuildDir()` is deprecated since Gradle 8.3, It will cause `Task :spring-boot-project:spring-boot-tools:spring-boot-gradle-plugin:compileJava FAILED`

See https://github.com/gradle/gradle/issues/20210


quaff:
I don't know why those test fails, It fails locally with main branch also.

wilkinsona:
While I'm happy to merge this, as I have said before you shouldn't be building the project with Gradle 8.3. Please use the wrapper.

PR #37189, opened 2023-09-05
Add virtual host support for Rabbit Stream

### LINKED ISSUE #37123: Add support for virtual host in RabbitProperties.Stream
The current version of [RabbitProperties.Stream](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java#L1197) currently does not allow to set a virtual host. It also does not consider the virtual host set on [RabbitProperties](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java#L88).

This is an issue in case a non-default virtual host (other than `/`) is used for streams.
Also the virtual host does not necessarily have to be the same as the virtual host that is set in [RabbitProperties](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java#L88). In our setup we use a dedicated virtual host for streams. For consistency reasons a fallback would make sense as implemented for the other properties.

The following changes are required to support this:
* Extend [RabbitProperties.Stream](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java#L1197) with a new property for the virtual host.
* Extend [RabbitStreamConfiguration](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitStreamConfiguration.java#L106) to map the property with a fallback to `properties::getVirtualHost`
```java
map.from(stream.getVirtualHost()).as(withFallback(properties::getVirtualHost)).whenNonNull().to(builder::virtualHost);
```

If you are fine with this change, I can try to create a pull request.

anessi:
Fixes #37123

I assume the documentation at https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/docs/asciidoc/application-properties.adoc?plain=1#L32 is generated automatically.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java
@@ -1207,6 +1207,12 @@ public class RabbitProperties {
 		private int port = DEFAULT_STREAM_PORT;
 
 		/**
+		 * Virtual host of a RabbitMQ instance with the Stream plugin enabled. When not
+		 * set, spring.rabbitmq.virtual-host is used.
+		 */
+		private String virtualHost;
+
+		/**
 		 * Login user to authenticate to the broker. When not set,
 		 * spring.rabbitmq.username is used.
 		 */
@@ -1237,6 +1243,14 @@ public class RabbitProperties {
 
 		public void setPort(int port) {
 			this.port = port;
+		}
+
+		public String getVirtualHost() {
+			return this.virtualHost;
+		}
+
+		public void setVirtualHost(String virtualHost) {
+			this.virtualHost = virtualHost;
 		}
 
 		public String getUsername() {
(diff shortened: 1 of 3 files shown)


mhalbritter:
Looks good, thank you. The only thing i'm not sure about is the automatic fallback to `spring.rabbit.virtual-host`. If you got a working setup right now (virtual host for rabbit, none for streams), then this will break after the merge, because stream will now use the same virtual host as rabbit does.

anessi:
Thanks for the feedback.

Yes, you're right that this changes the behavior slightly, but I think most people use the default virtual host `/` for everything. If someone considers separation by virtual host I would expect that they do it for everything, including streams. But that's opinionated or course 😄. Now they would have to set `spring.rabbit.stream.virtual-host` to `/` to revert back to the old values.

anessi:
fixed the formatting (line length) using `./gradlew format`

mhalbritter:
After discussion with the team, the fallback is fine. We should add a note to the release notes how to restore the old behavior.

mhalbritter:
Thank you very much @anessi !

Issue #36497, opened 2023-07-21
management.metrics.export.signalfx properties are incomplete


wilkinsona:
#33743 has revealed that the `management.metrics.export.signalfx` properties are incomplete. A property for `pushCumulativeHistogram` is missing.

wilkinsona:
In 3.0.x, `publishDeltaHistogram` is also missing.

wilkinsona:
Only one of delta or cumulative histograms may be published. If a `SignalFxConfig` implementation returns `true` from both `publishCumulativeHistogram` and `publishDeltaHistogram`, delta is used.

Rather than having two properties for this and relying on documentation for their mutual exclusivity, we could take advantage of this fact that we missed adding the properties and have an opportunity to consider both together by using a single property backed by an enum. In 2.7.x it would have two values (`NONE`, and `CUMULATIVE`). In 3.0.x we'd add `DELTA` to the values.

PR #37210, opened 2023-09-06
Add default value metadata for management.metrics.export.signalfx.published-histogram-type

izeye:
This PR adds default value metadata for `management.metrics.export.signalfx.published-histogram-type`.

See gh-36497


mhalbritter:
Thanks!

Issue #37221, opened 2023-09-07
Expose property to set strictServletCompliance on StandardServletMultipartResolver


ekko100120:
https://github.com/spring-projects/spring-boot/blob/d9af7cec141190fb32e20de6a64c2d3b03b644e4/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/servlet/MultipartAutoConfiguration.java#L74C47-L74C47

the Odata standard $batch request use `multipart/mixed;boundary=<uuid>`, but currently springboot `MultipartAutoConfiguration` autoconfigure not provide the configure properties to config strictServletCompliance field, which will result in `this.multipartResolver.isMultipart(request)` return true and the request mark as file upload request and result in bug
```java
	protected HttpServletRequest checkMultipart(HttpServletRequest request) throws MultipartException {
		if (this.multipartResolver != null && this.multipartResolver.isMultipart(request)) {
			if (WebUtils.getNativeRequest(request, MultipartHttpServletRequest.class) != null) {
				if (DispatcherType.REQUEST.equals(request.getDispatcherType())) {
					logger.trace("Request already resolved to MultipartHttpServletRequest, e.g. by MultipartFilter");
				}
			}
			else if (hasMultipartException(request)) {
				logger.debug("Multipart resolution previously failed for current request - " +
						"skipping re-resolution for undisturbed error rendering");
			}
			else {
				try {
					return this.multipartResolver.resolveMultipart(request);
				}
				catch (MultipartException ex) {
					if (request.getAttribute(WebUtils.ERROR_EXCEPTION_ATTRIBUTE) != null) {
						logger.debug("Multipart resolution failed for error dispatch", ex);
						// Keep processing error dispatch with regular request handle below
					}
					else {
						throw ex;
					}
				}
			}
		}
		// If not returned before: return original request.
		return request;
	}
```

philwebb:
Thanks for the suggestion. In the meantime you can add a `StandardServletMultipartResolver` bean to your own configuration:

```java
@Bean(name = DispatcherServlet.MULTIPART_RESOLVER_BEAN_NAME)
public StandardServletMultipartResolver multipartResolver(MultipartProperties multipartProperties) {
	StandardServletMultipartResolver multipartResolver = new StandardServletMultipartResolver();
	multipartResolver.setResolveLazily(multipartProperties.isResolveLazily());
	multipartResolver.setStrictServletCompliance(true);
[...]

PR #37242, opened 2023-09-08
Introduce configuration property "spring.servlet.multipart.strict-servlet-compliance"

quaff:
Closes GH-37221


mhalbritter:
Thank you!

PR #37260, opened 2023-09-08
Correct the description of spring.artemis.broker-url

mdeinum:
Fix description of brokerUrl property refer to url instead of port.


wilkinsona:
Thanks very much, @mdeinum.

PR #37272, opened 2023-09-08
Add pulsar starters into dependency management

eddumelendez:
Spring for Apache Pulsar was added introducing two new starters.
Those starters should be listed.


wilkinsona:
Good catch. Thanks very much, @eddumelendez.

Issue #36610, opened 2023-07-27
RabbitMQ smoke test has a invalid docker-compose


snicoll:
The `docker-compose` file for `spring-boot-smoke-test-amqp` is invalid, it's missing a `services` top-level element.

snicoll:
It's a bit worse than that. The smoke test doesn't have a test at all.

izeye:
This seems to be "type: task" rather than "type: bug" from a user perspective.

wilkinsona:
Thanks, @izeye. I've correct that and update [the changelog](https://github.com/spring-projects/spring-boot/releases/tag/v2.7.15).

PR #37394, opened 2023-09-14
Polish Javadoc for DockerImageNames.rabbit()

izeye:
This PR polishes Javadoc for `DockerImageNames.rabbit()` a bit.

See gh-36610


philwebb:
Thanks @izeye. I wonder if we should drop the `@since` tags since all this stuff is pretty internal.

wilkinsona:
I've opened https://github.com/spring-projects/spring-boot/issues/37403 for the removal of `@since`.

wilkinsona:
Thanks very much, @izeye.

PR #37434, opened 2023-09-15
Remove duplicate code in NettyWebServerFactoryCustomizer

shin-mallang:
Since the PropertyMapper's `alwaysApplyingWhenNonNull()` has already been called, the subsequent `whenNonNull()` is unnecessary.


mhalbritter:
Thank you!

Issue #33991, opened 2023-01-26
[Observability] Add local baggage and tagged baggage


marcingrzejszczak:
If we look at Sleuth & Boot tracing Baggage features comparison we're still missing two.

* Local Baggage - local baggage means baggage available within this JVM but one that does not get propagated over the wire
* Tagged Baggage - tagged baggage means baggage that automatically gets added as a tag

*Local Baggage* [Sleuth implementation ](https://github.com/spring-cloud/spring-cloud-sleuth/blob/3.1.x/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/brave/BraveBaggageConfiguration.java#L131-L174)

*Tagged Baggage* [Sleuth implementation](https://github.com/spring-cloud/spring-cloud-sleuth/blob/3.1.x/spring-cloud-sleuth-autoconfigure/src/main/java/org/springframework/cloud/sleuth/autoconfig/brave/BraveBaggageConfiguration.java#L228-L252)

michael-wirth:
Here's a little workaround for Spring Boot 3.1.x

OpenTelementry and Brave implementation behave differently.

### Brave
requires the local baggage to be registered:
```kotlin
    @Bean
    internal fun localBaggagePropagationCustomizer() =
        brave.baggage.BaggagePropagationCustomizer { customizer ->
           customizer.add(SingleBaggageField.local(BaggageField.create("localBaggage")))
        }
```

Set the local baggage in a WebFilter:
```kotlin
    override fun filter(exchange: ServerWebExchange, chain: WebFilterChain): Mono<Void> =
        chain.filter(exchange)
                        .doOnSubscribe { tracer.createBaggageInScope(localBaggage, "value") }
```

### OpenTelementry
doesn't require registering local baggage properties, but setting the baggage is more complicated:
- it is possible to set only 1 baggage. calling createBaggageInScope will cancel all previous baggages
- a new observation must be started with `Micrometer.observation`
```kotlin
    override fun filter(exchange: ServerWebExchange, chain: WebFilterChain): Mono<Void> {
        tracer.createBaggageInScope("localBaggage", "value")  // the value can also be set via "contextWrite", but not via "doOnSubscribe"
        return chain.filter(exchange)
            .tap(reactor.core.observability.micrometer.Micrometer.observation(registry))
    }
```

PR #37435, opened 2023-09-16
Add local correlation fields for Brave

adispezo:
Currently with the following config example it's not possible to use BaggageField.getByName, BaggageField.updateValue for example.

```yaml
management:
    tracing:
        baggage:
            correlation:
                fields: 
                    - some_local_field
                    - some_remote_field
           remote-fields:
               - some_remote_field
```

Only the "some_remote_field" can be retrieved and updated.

With the code proposed also the "local" correlation fields will be accessible.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
@@ -16,7 +16,6 @@
 
 package org.springframework.boot.actuate.autoconfigure.tracing;
 
-import java.util.ArrayList;
 import java.util.List;
 
 import brave.CurrentSpanCustomizer;
@@ -246,16 +245,25 @@ public class BraveAutoConfiguration {
 		}
 
 		@Bean
-		@Order(0)
+		@Order(1)
 		BaggagePropagationCustomizer localFieldsBaggagePropagationCustomizer() {
 			return (builder) -> {
-				final List<String> localFields = new ArrayList<>(
-						this.tracingProperties.getBaggage().getCorrelation().getFields());
-				localFields.removeAll(this.tracingProperties.getBaggage().getRemoteFields());
+				final List<String> localFields = this.tracingProperties.getBaggage().getLocalFields();
 				for (final String localFieldName : localFields) {
 					builder.add(BaggagePropagationConfig.SingleBaggageField.local(BaggageField.create(localFieldName)));
 				}
 			};
+		}
+
+		@Bean
+		@Order(2)
+		SpanHandler baggageTagSpanHandler() {
+			final List<String> tagFields = this.tracingProperties.getBaggage().getTagFields();
+
+			if (tagFields.isEmpty()) {
+				return SpanHandler.NOOP; // Brave ignores these
+			}
+			return new BaggageTagSpanHandler(tagFields.stream().map(BaggageField::create).toArray(BaggageField[]::new));
 		}
 
 		@Bean
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
@@ -16,6 +16,7 @@
 
 package org.springframework.boot.actuate.autoconfigure.tracing;
 
+import java.util.ArrayList;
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
```diff
@@ -244,6 +245,19 @@ BaggagePropagationCustomizer remoteFieldsBaggagePropagationCustomizer() {
 			};
 		}
 
+		@Bean
+		@Order(0)
```
mhalbritter, line 249: We already have a `Order(0)` `BaggagePropagationCustomizer`, i'd prefer `Order(1)`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
```diff
@@ -244,6 +245,19 @@ BaggagePropagationCustomizer remoteFieldsBaggagePropagationCustomizer() {
...
+		BaggagePropagationCustomizer localFieldsBaggagePropagationCustomizer() {
+			return (builder) -> {
+				final List<String> localFields = new ArrayList<>(
```
mhalbritter, line 252: I don't think it's a good idea to use the `getFields()` for both remote and local fields. In https://github.com/spring-projects/spring-boot/issues/33991 we want to add tagged fields. I'd prefer an explicit configuration property for that, as we have it for the remote fields.

mhalbritter (changes requested):

mhalbritter:
Hello,

thanks for the PR! I've added some comments.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
adispezo: I have changed the order.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java
adispezo: I have missed that conversation for some reason, but I have implemented both local and tagged fields. I also think it is to have the extra property fields.

mhalbritter:
Hey @adispezo , thanks for the changes! I'll get some feedback from the observability team on the local field change in Brave and then come back to you.

mhalbritter:
Looks good, thank you!

Note to the boot team: when merging, we need something like this for OpenTelemetry, too.

marcingrzejszczak:
There's no way to do it easily with OTel. What currently is being done is that all baggage gets extracted / injected (e.g.https://github.com/open-telemetry/opentelemetry-java/blob/main/api/all/src/main/java/io/opentelemetry/api/baggage/propagation/W3CBaggagePropagator.java#L50). We would need to wrap their baggage propagation mechanism into our one that would modify the current baggage not to contain local keys and then delegate to their code.

mhalbritter:
Hm :( In this case, we won't support local fields when using OpenTelemetry. We could issue a log WARN statement if local fields are configured and OTel is used

mhalbritter:
I have some polished changes and tagged fields for Otel here: https://github.com/mhalbritter/spring-boot/tree/pr/37435

mhalbritter:
Before merging this, https://github.com/spring-projects/spring-boot/issues/38724 needs to be considered.

mhalbritter:
Merged, thank you. To accommodate the changes from https://github.com/spring-projects/spring-boot/issues/38724, i had to refactor a bit.

o-shevchenko:
Hi
Do we already support tags `tag-fields` in Spring Boot 3.3.0? 
We've set `tag-fields`, but we still don't see tags sent to the OTEL collector.

mhalbritter:
Yes, this should work. If it doesn't, please open a new issue with a small reproducer. Thanks!

o-shevchenko:
Thanks @mhalbritter !
Should it be supported for all tracers or only Brave? I don't see it works with `io.micrometer:micrometer-tracing-bridge-otel`

mhalbritter:
Works with both tracers. [See the Otel configuration here](https://github.com/spring-projects/spring-boot/commit/25517350938f843f3077aa74d883d9525d1850af#diff-ffad1b43e8b173c888b93cf9becac2d848a524b9db2f5c30ba387e1b80a59e5cR145).

PR #37441, opened 2023-09-18
Use "spring.application.name" as fallback of "spring.kafka.consumer.group-id"

quaff:
Keep it consistent with `spring.kafka.streams.application-id`.

https://github.com/spring-projects/spring-boot/blob/4c4fdb35e9b7e4d83405355e75b26b2c20b5d380/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaStreamsAnnotationDrivenConfiguration.java#L63-L76

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java
@@ -37,6 +37,7 @@ import org.springframework.boot.context.properties.EnableConfigurationProperties
 import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Import;
+import org.springframework.core.env.Environment;
 import org.springframework.kafka.core.ConsumerFactory;
 import org.springframework.kafka.core.DefaultKafkaConsumerFactory;
 import org.springframework.kafka.core.DefaultKafkaProducerFactory;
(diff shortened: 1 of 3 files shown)


wilkinsona:
Seems sensible to me. Thanks, @quaff. WDYT, @garyrussell?

garyrussell:
I don't have any serious objections; but the consumer factory's `group.id` is just a default.

It's generally recommended to set the group id (either via the `id` or `groupId` property on each `@KafkaListener`).

It is not good practice to use the default (when multilple listeners are present) because multiple listeners would be placed in the same group, which would cause an unnecessary rebalance on all listeners when one is initiated on one of the listeners.

Another benefit of providing the `id` property is that you can manage the listener container's lifecycle via the `KafkaListenerEndpointRegistry`.

wilkinsona:
Thanks, @garyrussell. In light of that explanation, I'm less sure that this makes sense now. The `group-id` and `application-id` seem to have quite different roles and I'm not sure that `spring.application.name` is a good default for the group ID. I'm leaning towards declining now.

garyrussell:
Just to clarify further, in Kafka Streams, the `application.id` streams property is indeed used by `kafka-streams` to populate the consumer's `group.id`. 

There is no equivalent behavior for the regular consumer when not using streams.

quaff:
At least it has one advantage, we could ignore `@KafkaListener(groupId)` for `@SpringBootTest` tests.
According to my experience, many applications only contains one `@KafkaListener`, use `spring.application.name` as `groupId` is idiomatic.

philwebb:
We discussed this as a team today and given https://github.com/spring-projects/spring-boot/pull/37441#issuecomment-1723604951 we don't think that we should automatically set the group-id to the application name.

Thanks anyway for the PR @quaff!

Issue #37448, opened 2023-09-18
Rename spring.jms.listener.concurrency property to spring.jms.listener.min-concurrency


philwebb:
See discussion in #37180, specifically https://github.com/spring-projects/spring-boot/pull/37180#issuecomment-1704770720

We can call the `setConcurrentConsumers` and `setMaxConcurrentConsumers` methods on `DefaultMessageListenerContainer` directly rather than needing to create a string.

wilkinsona:
Thanks, @vpavic. I've corrected the issue's title.

The note about `DefaultMessageListenerContainer` is on me. I noticed the two separate setters during a Zoom discussion but didn't check that they're exposed on the factory. Thanks for #37451.

PR #37451, opened 2023-09-18
Rename JMS listener minimum concurrency property

vpavic:
This commit renames `spring.jms.listener.concurrency` property to `spring.jms.listener.min-concurrency` in order to better align it with `spring.jms.listener.max-concurrency`.

Closes gh-37448


mhalbritter:
Thanks very much!

PR #37491, opened 2023-09-20
Use more idiomatic Kotlin in example for "Map Health Indicators to Micrometer Metrics"

esperar:
> Please understand that I am not good at grammar because I am not an English speaker..

Hello, I am a 16-year-old developer who lives in Korea., I was looking at the spring docs part developed by Kotlin and saw MyHealth `Metrics ExportConfiguration`'s private fun getStatusCode(). It was using the if statement to return it, so I suggest using a `when` statement while using Kotlin!

----

### Before
```kt
private fun getStatusCode(health: HealthEndpoint): Int {
		val status = health.health().status
		if (Status.UP == status) {
			return 3
		}
		if (Status.OUT_OF_SERVICE == status) {
			return 2
		}
		if (Status.DOWN == status) {
			return 1
		}
		return 0
	}
```

### After
1.
```kt
private fun getStatusCode(health: HealthEndpoint): Int {
    	return when (health.health().status) {
        	Status.UP -> 3
        	Status.OUT_OF_SERVICE -> 2
        	Status.DOWN -> 1
        	else -> 0
    	}
}
```

**or**

2.
```kt
private fun getStatusCode(health: HealthEndpoint) = when (health.health().status) {
    Status.UP -> 3
    Status.OUT_OF_SERVICE -> 2
    Status.DOWN -> 1
    else -> 0
}
```

If you want to apply the second option, please leave a comment I'll change it.

Thank you :)


mhalbritter:

PR #37501, opened 2023-09-20
Disable validate connection on borrow for Oracle UCP datasource

sreekarareddy:
Removed setValidateConnectionOnBorrow method call while creating UCP datasource. This value should be false by default and can be overridden through datasource properties.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceConfiguration.java
@@ -180,7 +180,6 @@ abstract class DataSourceConfiguration {
 				throws SQLException {
 			PoolDataSourceImpl dataSource = createDataSource(connectionDetails, PoolDataSourceImpl.class,
 					properties.getClassLoader());
-			dataSource.setValidateConnectionOnBorrow(true);
 			if (StringUtils.hasText(properties.getName())) {
 				dataSource.setConnectionPoolName(properties.getName());
 			}


philwebb:
> This value should be false by default and can be overridden through datasource properties

Can you explain a bit more why the default value should be `false`? I'm concerned that if we change the default we will break back-compatibilty.

sreekarareddy:
The default value has always been false in UCP, this is a feature that should be explicitly set by the user.

UCP Doc: "setValidateConnectionOnBorrow(boolean): Specifies whether or not connections are validated when the connection is borrowed from the connection pool. The method enables validation for every connection that is borrowed from the pool. A value of false means no validation is performed. The default value is false."

link: https://docs.oracle.com/en/database/oracle/oracle-database/21/jjucp/validating-ucp-connections.html#GUID-211C9AFC-6ABA-4FEB-B10C-B0E2BEB526FD

The behaviour of UCP with Spring boot changes based on this property. When using a standalone ucp datasource, the default value is false and it should be the same when configured though spring boot properties as well.

wilkinsona:
Assuming that I have read the diff correctly, @snicoll added the `setValidateConnectionOnBorrow(true)` call as part of [polishing the Oracle UCP contribution](https://github.com/spring-projects/spring-boot/commit/75554ce236dead9ed5e55fd2408af97b7a8bc686). Stephane, can you remember the reasoning behind that change?

snicoll:
Ahem, I am afraid I do not. This looks strange.

jeandelavarene:
I agree that, by default, we should not turn on connection validation. It makes UCP do a roundtrip to the DB at every connection checkout which negatively impacts performance.

mhalbritter:
> Ahem, I am afraid I do not. This looks strange.

My guess is that @snicoll did this because all the other datasources do this by default, see the tests in `DataSourceAutoConfigurationTests`.

mhalbritter:

Issue #36347, opened 2023-07-11
When virtual threads are enabled, auto-configure Spring Pulsar to use them


wilkinsona:
Details are TBD (https://github.com/spring-projects/spring-pulsar/issues/414) but it's likely to be a configuration setting on `ConcurrentPulsarListenerContainerFactory`

vpavic:
Is this still blocked and/or should it be superseded by #37559?

wilkinsona:
Thanks for the nudge, @vpavic. This is no longer blocked. I don't think it's superseded by #37559 as we can implement this without it (and I would prefer to do so as I'm not yet sold on the complexity introduced by #37559).

PR #37559, opened 2023-09-24
Add Spring Pulsar container property customizers

onobc:
This commit adds the ability for users to register property customizers for the auto-configured Spring Pulsar listener containers.

The motivation is to ultimately give users the ability to set the `AsyncTaskExecutors` on the underlying listener containers to support Java virtual threads capability.

See #36347

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarContainerPropertiesCustomizer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarContainerPropertiesCustomizer.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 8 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarContainerPropertiesCustomizer.java
```diff
@@ -0,0 +1,36 @@
...
+import org.springframework.pulsar.listener.PulsarContainerProperties;
+
+/**
+ * The interface to customize a {@link PulsarContainerProperties}.
+ *
+ * @author Chris Bono
+ * @since 3.2.0
+ */
+@FunctionalInterface
+public interface PulsarContainerPropertiesCustomizer {
```
onobc, line 28: I was originally going to put these container props customizers in the framework. However, the container factories do not need to accept a customizer as they already accept the container props. Non-boot users will already have to create the container props in order to create the container factory. They will therefore just adjust the props accordingly - no need to also have a customizer. 

IOW - the framework does not need the concept of container props customizers.

Issue #36769, opened 2023-08-07
Update documentation to align with Mockito 5 using the inline mock maker by default


wilkinsona:
With thanks to @romainmoreau, we [need to](https://github.com/spring-projects/spring-boot/issues/34998#issuecomment-1666776423) update the docs following the upgrade to Mockito 5.3 as it dropped the `mockito-inline` module. Furthermore, the inline mock maker (now part of the code module) has been the default since 5.0. [This note](https://github.com/spring-projects/spring-boot/blob/9d42cff472c7d67972036d60ed2b1e64b41ee253/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/testing.adoc?plain=1#L302-L305) can be removed from the documentation.

wilkinsona:
Closing in favor of #37561.

PR #37561, opened 2023-09-24
 Update documentation to align with Mockito 5 using the inline mock maker by default

sushant1987:
With the upgrade to Mockito 5.3,  it dropped the mockito-inline module. Removed [this note](https://github.com/spring-projects/spring-boot/blob/9d42cff472c7d67972036d60ed2b1e64b41ee253/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/testing.adoc?plain=1#L302-L305) from the documentation.

Fixes https://github.com/spring-projects/spring-boot/issues/36769


mhalbritter:
Thank you very much!

PR #37574, opened 2023-09-25
Managed types for Neo4j are not used in Neo4j Data auto configuration

### LINKED ISSUE #37573: Managed types for Neo4j are not used in Neo4j Data auto configuration.
The auto configuration for imperative and reactive Spring Data variant still uses 

```java
Set<Class<?>> initialEntityClasses = new EntityScanner(applicationContext).scan(Node.class,
				RelationshipProperties.class);
```

as initial entity classes for the mapping context. This works fine in JVM mode and to some extent in native mode. It fails however with certain inheritance scenarios (See https://github.com/spring-projects/spring-data-neo4j/issues/2798).

The solution is to use populate the mapping context with the managed entities and I'll have a PR up in a moment.

michael-simons:
This fixes #37573.

I tested the fix locally in native image mode and added a test that reliable fails without using `Neo4jManagedTypes` proper. 
I don't know whether you run the integration tests in native mode, or put them somewhere else. If the latter is the case, please point me where I can move this or feel free to move it around yourself.

I suggest back porting this to 3.1 and 3.0 (I think 3.0 already managed SDN 7, including the dedicated managed types).

Thanks a lot!

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/neo4j/Neo4jDataAutoConfiguration.java
@@ -33,6 +33,7 @@ import org.springframework.boot.autoconfigure.transaction.TransactionManagerCust
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.ApplicationContext;
 import org.springframework.context.annotation.Bean;
+import org.springframework.data.neo4j.aot.Neo4jManagedTypes;
 import org.springframework.data.neo4j.core.DatabaseSelectionProvider;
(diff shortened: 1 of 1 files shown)


wilkinsona:
https://github.com/spring-projects/spring-aot-smoke-tests would be a good place for the new test. Having looked there just now, I've noticed that we don't have any coverage for SDN. Setting up a new project in the smoke tests is a little bit involved so we can take a look at that and copy over something that's equivalent to the test proposed here.

michael-simons:
That failed with no apparent connection for me towards my changes…

wilkinsona:
Sorry about that. GitHub Actions and Pulsar don't seem to get on. We're not sure why yet. It might be resource constraints on the runner in which case #37578 should help.

michael-simons:
I removed the additional tests in favour to have it in the smoke tests project where it makes a lot more sense https://github.com/spring-projects/spring-aot-smoke-tests/pull/190

Thanks for pointing me there.

PR #37606, opened 2023-09-27
Refer to ActiveMQ as ActiveMQ "Classic"

jbertram:
The Apache ActiveMQ project has adopted the label "Classic" for the 5.x broker in order to more easily compare and contrast it with the broker code-named Artemis.

The Spring docs should do the same to clarify ActiveMQ broker usage and configuration.


mhalbritter:
Thank you!

PR #37607, opened 2023-09-27
Default value of spring.jmx.registration-policy is not documented

### LINKED ISSUE #37596: Default value of spring.jmx.registration-policy is not documented
https://github.com/spring-projects/spring-boot/pull/34122 should have added one. When I merged the contribution I missed that the default's declared in the `hints` section rather than the `properties` section so the current additional metadata has no effect.

sushant1987:
Moves default value declaration of `spring.jmx.registration-policy` from `hints`  to `properties`
Closes #37596


wilkinsona:
Thanks, @sushant1987, but I already have this change locally from when I opened the issue earlier today.

Issue #34657, opened 2023-03-20
Allow auto-configuration to consume service connection information from sources other than configuration properties


wilkinsona:
We'd like our auto-configurations to be able to consume service connection information from sources other than configuration properties. When such connection details are available, they should take precedence over connection-related configuration properties while continuing to use other configuration properties that are not related to establishing a connection with a remote service. For example, connection details for an SQL database may be a JDBC url, a username, and a password. When such details are available, application properties for configuring things like a maximum connection pool size will still be used.

Using connection details rather than application properties allows the properties related to connecting to a service to be ignored as a single unit, avoiding problems where a configuration property has been set and needs to be unset to allow the service to be used correctly.

This feature is largely intended as a building block on top of which other more user-facing features will be built. For example, it will allow us to offer improved integration with Testcontainers, removing boilerplate for getting the auto-configuration to use a service in a `@Container`-annotated field in a test class.

PR #37619, opened 2023-09-28
IPv6 IP addresses cannot be used with RabbitMQ

jonasfugedi:
Failed to parse ipv6 addresses due to assumption of ipv4 addresses.

I think the regression was introduced in https://github.com/spring-projects/spring-boot/issues/34657

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/PropertiesRabbitConnectionDetails.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/PropertiesRabbitConnectionDetails.java
@@ -53,8 +53,10 @@ class PropertiesRabbitConnectionDetails implements RabbitConnectionDetails {
 	public List<Address> getAddresses() {
 		List<Address> addresses = new ArrayList<>();
 		for (String address : this.properties.determineAddresses().split(",")) {
-			String[] components = address.split(":");
-			addresses.add(new Address(components[0], Integer.parseInt(components[1])));
+			int portSeparatorPosition = address.lastIndexOf(':');
+			String host = address.substring(0, portSeparatorPosition);
+			String port = address.substring(portSeparatorPosition + 1);
+			addresses.add(new Address(host, Integer.parseInt(port)));
 		}
 		return addresses;
 	}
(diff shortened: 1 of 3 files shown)


scottfrederick:
@jonasfugedi Thanks very much for your first contribution to Spring Boot. I've merged it along with [a polishing commit](https://github.com/spring-projects/spring-boot/commit/fad43adf144129730733c0500e55a326f529e99d) for style and a small bit of additional testing.

mjustin:
> public List<Address> getAddresses() {
> 	List<Address> addresses = new ArrayList<>();
> 	for (String address : this.properties.determineAddresses().split(",")) {
> 		int portSeparatorIndex = address.lastIndexOf(':');
> 		String host = address.substring(0, portSeparatorIndex);
> 		String port = address.substring(portSeparatorIndex + 1);
> 		addresses.add(new Address(host, Integer.parseInt(port)));
> 	}
> 	return addresses;
> }

Is it possible for an IPv6 address to be specified without a port (i.e. use the default port)?  Wouldn't this code fail in that scenario, since it'd split on a colon that's part of the IP address, and not related to the port?

jonasfugedi:
> > ```
> > public List<Address> getAddresses() {
> > 	List<Address> addresses = new ArrayList<>();
> > 	for (String address : this.properties.determineAddresses().split(",")) {
> > 		int portSeparatorIndex = address.lastIndexOf(':');
> > 		String host = address.substring(0, portSeparatorIndex);
> > 		String port = address.substring(portSeparatorIndex + 1);
> > 		addresses.add(new Address(host, Integer.parseInt(port)));
> > 	}
> > 	return addresses;
> > }
> > ```
> 
> 
> Is it possible for an IPv6 address to be specified without a port (i.e. use the default port)?  Wouldn't this code fail in that scenario, since it'd split on a colon that's part of the IP address, and not related to the port?

Not if I read the code correct, the source of the addresses is

https://github.com/jonasfugedi/spring-boot/blob/bebca55a8ffee3b4cd4d099fb6be199655cf8c7d/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/RabbitProperties.java#L207-L225

Which ensures a port is always appended to the host. Previous behavior was retained, i.e. I kept the same assumptions as the code did before except I needed it to work when using ipv6 addresses.

PR #37655, opened 2023-10-03
Use AssertJ assertions consistently

izeye:
This PR changes to use AssertJ assertions consistently.

This PR also updates Checkstyle rule for it.


wilkinsona:
Thanks very much, @izeye.

PR #37726, opened 2023-10-05
Mention new RestClient class when suggesting @RegisterReflectionForBinding

SimonVerhoeven:
Spring boot 6.1 includes a new way to execute this class, which has the same behaviour as WebClient so is ideally also mentioned in the documentation.


mhalbritter:

PR #37737, opened 2023-10-05
Create RepositoryEventBeanValidationConfiguration

yangli136:
create bean validation against Repository create and save operations.

Validating domain objects before invoking Repository operations (create and save) should be a default feature.


philwebb:
Thanks for the pull request, but I'm not sure we should make this change in the way that you've suggested. We already have a `ValidationAutoConfiguration` and it feels a bit strange to have another `LocalValidatorFactoryBean` bean defined by configuration in the `org.springframework.boot.autoconfigure.data.rest` package.

I'm also not really sure that configuring the validator is something we should do out of the box. Perhaps you could open an issue describing the rationale for the change so we can discuss it a little more.

Issue #36264, opened 2023-07-06
`restTemplateBuilderConfigurer` doesn't need to be `@ConditionalOnMissingBean`


wilkinsona:
`restTemplateBuilderConfigurer ` defined by `RestTemplateAutoConfiguration` is `@ConditonalOnMissingBean` but that condition doesn't appear to be necessary. `RestTemplateBuilderConfigurer` is a `final` class and is immutable from the perspective of its public API. You could define a custom configurer bean but it would be in its default state and would do nothing when asked to configure the `RestTemplateBuilder`. You could just define a `RestTemplateBuilder` bean directly instead:

```
@Bean
public RestTemplateBuilder restTemplateBuilder() {
	 return new RestTemplateBuilder();
}
```

We should remove `@ConditonalOnMissingBean`. In the unlikely event of someone having defined their own `RestTemplateBuilderConfigurer` this would be a breaking change as two `RestTemplateBuilderConfigurer` beans would then be defined. As such, we should remove the condition in 3.2 at the earliest and mention the change in the release notes.

PR #37746, opened 2023-10-07
restTemplateBuilderConfigurer doesn't need to be @ConditionalOnMissingBean

techiesarav:
Fix the issue #36264


techiesarav:
Thanks @mhalbritter kind of missed the format though. Thanks for taking care of build failure

PR #37768, opened 2023-10-09
Add SSL store bundle which uses PEM files from a directory

mhalbritter:
This PR implements an SSL bundle which reads PEM encoded files from a directory. It uses a `CertificateMatcher` to find all certificate files, a `CertificateSelector` to select one certificate from all the found ones, and a `KeyLocator` to find the key belonging to the selected certificate.

I've implemented 3 different strategies to select a cerificate:

* The one with the longest lifetime (`MaximumNotAfterCertificateSelector`)
* The newest one, based on the not before field (`MaximumNotBeforeCertificateSelector`)
* The newest one, based on the file creation time (`NewestFileCertificateSelector`)

Those bundles be configured with the `spring.ssl.pemdir` config property.

New public API:

* class `PemDirectorySslStoreBundle` with nested interfaces `KeyLocator`, `CertificateMatcher` and `CertificateSelector`
* record `PemDirectorySslStoreDetails`


philwebb:
See https://github.com/philwebb/spring-boot/tree/gh-37768 for updated code.

PR #37808, opened 2023-10-10
Implement SSL bundle reload

mhalbritter:
Implements reloading of SSL bundles. Netty and Tomcat react to SSL bundle reload and hot-reload their SSL certificates (existing connections use the old key material, new connections use the new material).

Implemented a FileWatcher based on `java.nio` to watch for file changes.

Adds new properties to JKS and PEM bundles to control reload. If reload is enabled, the keystore / key files / certificate files are automatically watched and reload the SSL bundle if they change (which in turn will trigger reload of SSL in tomcat / netty).

I had to refactor the `PemContent` class a bit, because we now need to know at 2 places if the PEM SSL bundle configuration points is a URL or PEM content.


gradle.properties
```diff
@@ -7,6 +7,6 @@ org.gradle.jvmargs=-Xmx2g -Dfile.encoding=UTF-8
 kotlinVersion=1.9.10
 nativeBuildToolsVersion=0.9.27
 springFrameworkVersion=6.1.0-SNAPSHOT
-tomcatVersion=10.1.13
+tomcatVersion=10.1.14
```
mhalbritter, line 10: This breaks some tests, which I haven't fixed in this PR. I had to upgrade the Tomcat version to be able to call `protocol.addSslHostConfig(sslHostConfig, true);`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/FileWatcher.java
```diff
@@ -0,0 +1,302 @@
...
+					}
+				}
+				catch (InterruptedException ex) {
+					Thread.currentThread().interrupt();
+				}
+			}
+			logger.debug("Watch thread stopped");
+		}
+		catch (IOException ex) {
+			throw new UncheckedIOException(ex);
```
mhalbritter, line 163: If an exception is thrown, this thread dies and stays dead. Should we make this more robust and keep running in case of exceptions?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/SslAutoConfiguration.java
```diff
@@ -37,19 +36,28 @@
...
 
-	SslAutoConfiguration() {
+	private final SslProperties sslProperties;
+
+	SslAutoConfiguration(SslProperties sslProperties) {
+		this.sslProperties = sslProperties;
+	}
+
+	@Bean
+	FileWatcher fileWatcher() {
```
mhalbritter, line 46: The FileWatcher is lazy, it only creates its working thread when at least one file / directory is watched.

mhalbritter:
@philwebb did a pass over the code, I force pushed this PR with his changes. Besides the Tomcat upgrade which breaks some tests this is ready to merge.

PR #37813, opened 2023-10-11
Exclude unwanted transitive dependencies from Apache Pulsar

onobc:
Several of the Apache Pulsar libs are bringing along transitive dependencies (`com.sun.activation:javax.activation` and `javax.validation:validation-api`). This commit adds exclusions to the Spring Boot BOM for these problematic transitive dependencies.

**NOTE:** Currently these exclusions are handled in the Spring for Apache Pulsar framework for a couple of the Pulsar libs
- https://github.com/spring-projects/spring-pulsar/blob/5ff3d99b99673c67c59528ae02c02c2eae4cd837/spring-pulsar/spring-pulsar.gradle#L9-L12
- https://github.com/spring-projects/spring-pulsar/blob/5ff3d99b99673c67c59528ae02c02c2eae4cd837/spring-pulsar-reactive/spring-pulsar-reactive.gradle#L9-L16
However, this list is not exhaustive and it is possible for users to include other Pulsar libs which will not have the unwanted visitors excluded. Adding the exclusions here guarantees the unwanted visitors will never arrive. 

This commit adds exclusions for these problematic transitive dependencies here in the Spring Boot BOM.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1148,17 +1148,30 @@ bom {
 				"pulsar-client-1x",
 				"pulsar-client-2x-shaded",
 				"pulsar-client-admin-api",
-				"pulsar-client-admin-original",
-				"pulsar-client-admin",
-				"pulsar-client-all",
+				"pulsar-client-admin-original" {
+					exclude group: 'com.sun.activation', module: 'javax.activation'
+				},
+				"pulsar-client-admin" {
+					exclude group: 'com.sun.activation', module: 'javax.activation'
+					exclude group: 'javax.validation', module: 'validation-api'
+				},
+				"pulsar-client-all" {
+					exclude group: 'com.sun.activation', module: 'javax.activation'
+					exclude group: 'javax.validation', module: 'validation-api'
+				},
 				"pulsar-client-api",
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal. Unfortunately, I'm not in favor of doing this for a couple of reasons.

Excludes in the bom may or may not work depending on how it is consumed. For example, they do not work when the bom's used as a Gradle Platform. Rather than adding more exclusions, if we do anything we should remove those that are already declared.

Secondly, if Pulsar publishes a bom in the future – which I really hope they will given the number of modules that the project has – we'd lose these excludes when switching to it which would be an unwanted change in behavior.

Boot's build already prohibits dependencies like `javax.validation:validation-api` so I'm confident that they aren't appearing transitively anywhere in Boot's modules. Beyond this, I don't think we should be trying to improve Pulsar's dependency hygiene. Ideally, that's a job for the Pulsar project itself. Failing that it will have to be done by users so that the experience is consistent.

onobc:
Thanks @wilkinsona 

I was curious what would come back from this PR as I was not 100% sure the exclusion mechanism would do what we want in 100% cases. 

> Secondly, if Pulsar publishes a bom in the future –

Yep, I have a TODO to create an issue in Pulsar to add the BOM and will likely submit the code proposal myself. 

Thanks for the info and taking the time to explain all the details.

Issue #37812, opened 2023-10-10
Upgrade to Tomcat 10.1.15


philwebb:
Needed for https://github.com/spring-projects/spring-boot/pull/37808

philwebb:
Looks like we might need to wait for 10.1.15 with [this fix](https://github.com/apache/tomcat/commit/c674ec9442b563b51c2cb4958301f2adf98423de)

grgrzybek:
Just a reminded - you also need [this fix|https://github.com/apache/tomcat/commit/4729f7e16f6bcab28c61bf5f5bac546a45542c64] related to gzip compression.

@philwebb hi - I didn't find a ticket for Spring Boot 2.7.x which uses Tomcat 9.0.80 affected by [CVE-2023-44487](https://nvd.nist.gov/vuln/detail/CVE-2023-44487).
Mind that there's the same situation - version 9.0.81 fixes this CVE, but breaks `gzip` encoding. 9.0.82 should be available soon (under ASF voting now).

grgrzybek:
Tomcat 9.0.82 is already available. 10.1.15 should be in Maven Central soon I guess.

wilkinsona:
Thanks for the info, but it isn't really necessary. Please rest assured that we're well aware of the situation with CVE-2023-44487 and we will be upgrading our various web server dependencies once the remaining releases are available.

In the hope that it minimises any further noise for those watching the repository, I am going to lock this issue.

wilkinsona:
Superseded by https://github.com/spring-projects/spring-boot/issues/37901.

PR #37845, opened 2023-10-12
Update tomcat version fixes DOS vulnerability

sreekesh93:
This update fix the vulnerability found in tomcat https://security.snyk.io/vuln/SNYK-JAVA-ORGAPACHETOMCATEMBED-5953331

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -13,6 +13,6 @@ kotlinVersion=1.9.10
 mavenVersion=3.9.4
 nativeBuildToolsVersion=0.9.27
 springFrameworkVersion=6.1.0-SNAPSHOT
-tomcatVersion=10.1.13
+tomcatVersion=10.1.14
 
 kotlin.stdlib.default.dependency=false


bclozel:
We're very much aware of this, thank you. 10.1.14 is the wrong version to pick anyway as it shipped with several regressions, so our semi-automated process will pick 10.1.15 instead (see #37812).

As mentioned in our issue template:
> You DO NOT need to raise an issue for a managed dependency version upgrade as there's a semi-automatic process for checking managed dependencies for new versions before a release.

We will perform upgrades matching [our 3rd party upgrade policy](https://github.com/spring-projects/spring-boot/wiki/Supported-Versions#third-party-dependencies).
If this is an urgent matter for you, there is no need to wait for us releasing as you can use a [Gradle](https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/htmlsingle/#managing-dependencies.dependency-management-plugin.customizing) or [Maven](https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#using.parent-pom) build property to override the library version.
All [version properties are listed in the reference documentation appendix](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#appendix.dependency-versions.properties).

PR #37885, opened 2023-10-14
Add properties for configuring Jackson's EnumFeature and JsonNodeFeature

eddumelendez:
Both, `EnumFeature` and `JsonNodeFeature` implement `DataTypeFeature`,
which was recently added in spring-framework. This commits introduces
support to allow the configuration via properties.

See spring-projects/spring-framework#31380


wilkinsona:

Issue #24744, opened 2021-01-11
Switch to Apache EL implementation by default


TranceTip:
We have noticed that JSP EL expression evaluation has become significantly slower when non-existing attributes are accessed, starting with Spring Boot 2.3.0. This problem seems to affect all current versions of Spring Boot 2.3.x (2.3.0 to 2.3.7) and 2.4.x (2.4.0 and 2.4.1), but not 2.2.x.

Starting with Spring Boot 2.3.0, the following happens: For each non-existing attribute accessed during Spring EL expression evaluation on a JSP (for example, when using `${myProperty}` or `<c:if test="${myProperty}">...</c:if>` and no such attribute is defined in the view model), there are class loader calls to find the classes "`java.lang.<attribute>`", "`java.servlet.<attribute>`", "`java.servlet.http.<attribute>`" and "`java.servlet.jsp.<attribute>`".

This happens once per unique attribute on each request (or per JSP/tag context); accessing the same attribute again in the same context does not lead to additional class lookups. This significantly slows down the EL expression evaluation, since the 4 lookups for each attribute take a few milliseconds on an idle machine. When properties that do exist are accessed, there is no performance difference between the different Spring Boot versions. At least in our HTML frontend it is common that we only set model attributes in certain cases, and the JSPs/tags then check whether they are set, so accesses to non-existing attributes happen very frequently, which leads to noticeable performance problems.

Root cause seems to be that a different `ExpressionFactory` is used (`com.sun.el.ExpressionFactoryImpl` in 2.3.0+ instead of `org.apache.el.ExpressionFactoryImpl` in 2.2.x). The latter sets a certain context attribute to true (see below), but the former does not.

The `ScopedAttributeELResolver` contains this code:

```java
   boolean resolveClass = true;
    // Performance short-cut available when running on Tomcat
    if (AST_IDENTIFIER_KEY != null) {
        // Tomcat will set this key to Boolean.TRUE if the
        // identifier is a stand-alone identifier (i.e.
        // identifier) rather than part of an AstValue (i.e.
        // identifier.something). Imports do not need to be
        // checked if this is a stand-alone identifier
        Boolean value = (Boolean) context.getContext(AST_IDENTIFIER_KEY);
        if (value != null && value.booleanValue()) {
            resolveClass = false;
        }
    }
    // This might be the name of an imported class
    ImportHandler importHandler = context.getImportHandler();
    if (importHandler != null) {
        Class<?> clazz = null;
[...]

PR #37973, opened 2023-10-20
Remove tomcat-embed-el from spring-boot-starter-jetty

leleuj:
I don't understand why the `org.apache.tomcat.embed:tomcat-embed-el` dependency is defined with the `api` scope in the `spring-boot-starter-jetty` module: it perfectly builds without it.


wilkinsona:
Thanks for the suggestion but the dependency is required. It's there to ensure that apps using `spring-boot-starter-jetty` have an expression language implementation available to them. For consistency, [we use Tomcat's EL implementation everywhere](https://github.com/spring-projects/spring-boot/issues/24744).

leleuj:
OK. Thanks for the feedback.

PR #37975, opened 2023-10-20
Add JVM Checkpoint Restore documentation

sdeleuze:
My proposal is to put CRaC documentation in the efficient deployment section, since we are providing an initial support and given the conciseness of the documentation, a dedicated top level section looks like not the best outcome from my point of view.

I have not added a link to the very useful https://github.com/sdeleuze/spring-boot-crac-demo repository since that's a personal one, and I was not sure it was ok for you. But if you are, I think it would add quite a lot of added value to include it (feel free to update my PR accordingly).

I did not add additional warnings on Spring Boot side related to being careful on not leaking sensitive data since Spring Framework documentation is already pretty clear about it, but if you want add a warning on Spring Boot side, that's on option.


mhalbritter:
Thank you!

PR #37986, opened 2023-10-21
Fix links in Spring Pulsar section of ref guide

onobc:
* Spring for Apache Pulsar recently moved to Antora docs which in turn modified the base url slightly. This commmit adjusts for the move.


mhalbritter:
Thanks!

PR #37987, opened 2023-10-21
Fix links in Spring Kafka section of ref guide

onobc:
* Spring for Apache Kafka recently moved to Antora docs which in turn modified the base url slightly. This commmit adjusts for the move.

cc: @garyrussell


mhalbritter:
Thanks!

PR #38009, opened 2023-10-24
Remove unreachable URL

yeseong0412:
http://livereload.com/
is not connect server is down
http://schemas.xmlsoap.org/
There seems to be a problem with connection.

I think I need to check the link and edit or delete it.

Thank you.


wilkinsona:
Thanks for the proposal but this breaks the build as there are four files in the repository that reference the two removed URLs and `http://` links are prohibited by default. For future reference, please build your changes using `gradlew build` and fix any problems before submitting them.

PR #38101, opened 2023-10-29
Refactor: Extract "hibernate.physical_naming_strategy" 

dr94406:
In this way, other elements were extracted and used as constants, but hibernate.physical_naming_strategy was hardcoded and used. Change accordingly


philwebb:
Thanks for the pull-request, but I don't think we should extract these. Since they're only used in one place there is little benefit in moving them away from the code. In addition, your change has accidentally change the implicit naming strategy to use the wrong `hibernate.physical_naming_strategy` value.

Issue #33689, opened 2023-01-04
Auto-configure a JwtAuthenticationConverter


hannah23280:
Hi,
I learn the below way to customize the authority prefix and claim name. Would like to suggest provide configuration properties to allow developer to easily customize the JwtAuthenticationConverter, so that developer does not have to write these boilerplate codes

```java
@Bean
    public JwtAuthenticationConverter jwtAuthenticationConverter() {
        JwtGrantedAuthoritiesConverter grantedAuthoritiesConverter = new JwtGrantedAuthoritiesConverter();
        grantedAuthoritiesConverter.setAuthorityPrefix("");
        //grantedAuthoritiesConverter.setAuthoritiesClaimName("role");
        JwtAuthenticationConverter jwtAuthenticationConverter = new JwtAuthenticationConverter();
        jwtAuthenticationConverter.setJwtGrantedAuthoritiesConverter(grantedAuthoritiesConverter);
        return jwtAuthenticationConverter;
    }
```

wilkinsona:
Thanks for the suggestion. It looks like we could auto-configure a `JwtAuthenticationConverter` bean if there's no such existing bean in the context and a property has been set for the principal claim name, the authority prefix, or the authorities claim name.

mhalbritter:
Superseded by https://github.com/spring-projects/spring-boot/pull/38105.

PR #38105, opened 2023-10-30
Auto-configure a JwtAuthenticationConverter

ykardziyaka:
Adds support for auto-configuring `JwtAuthenticationConverter` (and `ReactiveJwtAuthenticationConverter`) with following properties:
- `spring.security.oauth2.resourceserver.jwt.authority-prefix`
- `spring.security.oauth2.resourceserver.jwt.authorities-claim-delimiter`
- `spring.security.oauth2.resourceserver.jwt.authorities-claim-name`
- `spring.security.oauth2.resourceserver.jwt.principal-claim-name`

Closes gh-33689

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/OAuth2ResourceServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/resource/OAuth2ResourceServerProperties.java
@@ -26,6 +26,7 @@ import java.util.List;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.source.InvalidConfigurationPropertyValueException;
 import org.springframework.core.io.Resource;
+import org.springframework.security.core.GrantedAuthority;
 import org.springframework.util.Assert;
 import org.springframework.util.StreamUtils;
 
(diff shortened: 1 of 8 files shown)


mhalbritter:
I got some amended changes here: https://github.com/mhalbritter/spring-boot/tree/pr/38105

Jojoooo1:
Are you planning to add nested authorities-claim-name ? For example, Keycloak is nesting the roles in realm_access.roles

wilkinsona:
@Jojoooo1 I don't think so, but I'm not sure that I understand what you're asking for. Please open a new issue describing how to configure a `JwtGrantedAuthoritiesConverter` that meets your needs. We can then consider how we could support such configuration using application properties.

Jojoooo1:
Sorry I was not very clear, I was wondering if in the future you are planning to allow nested claim name with the property spring.security.oauth2.resourceserver.jwt.authorities-claim-name. My use case was regarding Keycloak for example the jwt structure is as follows:
```json
"realm_access": {
    "roles": [
      "default-roles-api",
      "offline_access",
      "uma_authorization",
      "management_user"
    ]
  },
  "resource_access": {
    "account": {
      "roles": [
        "manage-account",
        "manage-account-links",
        "view-profile"
      ]
    }
  },
```

wilkinsona:
`spring.security.oauth2.resourceserver.jwt.authorities-claim-name` maps onto `org.springframework.security.oauth2.server.resource.authentication.JwtGrantedAuthoritiesConverter.setAuthoritiesClaimName(String)`. If you want that to support more than it currently does today, I believe a change to `JwtGrantedAuthoritiesConverter`, which is part of [Spring Security](https://github.com/spring-projects/spring-security), would be required. If this is something that you would like to see, please open an issue over there.

Issue #38155, opened 2023-10-31
Rename startDaemonAwaitThread() inside TomcatWebServer to match intent


jxblum:
This is just an observation and suggestion, but when reviewing the `TomcatWebServer` class ([source](https://github.com/spring-projects/spring-boot/blob/v3.2.0-RC1/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java)), I noticed the method name for "_create a blocking non-daemon to stop immediate shutdown_" was named `startDaemonAwaitThread()` (called from [here](https://github.com/spring-projects/spring-boot/blob/v3.2.0-RC1/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java#L135-L137)), when, in fact, the `Thread` that is created is a "non-daemon" `Thread` as was intended, [apparent inside the method](https://github.com/spring-projects/spring-boot/blob/v3.2.0-RC1/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java#L202), and as is necessary to prevent shutdown of the Tomcat Web Server JVM.

This method would probably be better named as `startNonDaemonAwaitThread()`, or maybe `startTomcatAwaitThread()`

itsayopapi:
@jxblum can I please work on this issue?

philwebb:
Thanks for the offer @itsayopapi, but I think it will be easier for us to handle this one ourselves rather than merging a pull-request.

jxblum:
Hi @itsayopapi - That is up to the Boot team and you. I am sure they would welcome help on this rather simple change if the Boot team feels the change is valid.

Thank you for your interest and help.

Cheers!

wilkinsona:
As Phil has said above, we'd prefer to handle this one ourselves.

PR #38162, opened 2023-11-01
Replace and Rename

itsayopapi:


wilkinsona:
Thanks, @itsayopapi, but [as Phil said](https://github.com/spring-projects/spring-boot/issues/38155#issuecomment-1788596642), we'd prefer to handle this one ourselves.

PR #38184, opened 2023-11-02
Deprecate OutputCaptureRule for removal

dreis2211:
Hi,

I know it's quite late in the 3.2.0 release process, but I wonder if we can deprecate `OutputCaptureRule` for removal still.

JUnit 5 has been released in 2017 if I remember correctly. It's also the recommended default for quite some time. I think it's time to deprecate JUnit 4 and related functionality.

Let me know what you think.
Cheers,
Christoph


dreis2211:
Found the relevant commit (from May 2019) where the docs were changed to document JUnit 5 as the recommended default. 1de6f6823850845f0fa6da0cf44c14d1aca31ea2

philwebb:
We discussed this today and decided that since the current code hasn't really caused any maintenance burden for us we should leave it there. We'd obviously prefer that folks migrate to JUnit 5, but we anticipate that a fair few might not have made the leap.

PR #38208, opened 2023-11-04
Add missing Category attribute to productionRuntimeClasspath configuration

belovaf:
Spring boot gradle plugin creates configuration `productionRuntimeClasspath`, but it doesn't add a very important attribute `Category` which leads to wrong variant selection for some libraries.

For example, [kotlin-stdlib-1.9.20.module](https://repo1.maven.org/maven2/org/jetbrains/kotlin/kotlin-stdlib/1.9.20/kotlin-stdlib-1.9.20.module) contains a `jvmSourcesElements` variant with `category=documentation`, which is erroneously selected during `productionRuntimeClasspath` resolution.


wilkinsona:
Thanks for the proposal but I don't think this is the right way to tackle this one. It would be better to copy the attributes over so that `productionRuntimeClasspath` automatically aligns with `runtimeClasspath`. The changes are also breaking the build as the tests have not been updated to account for the new attribute.

We'll take care of things using #38209 which appears to be the same problem. Thanks anyway for the PR.

Issue #38218, opened 2023-11-06
Opentelemetry service.name default should align with OpenTelemetry Semantic Convention


lenin-jaganathan:
https://github.com/spring-projects/spring-boot/blob/4bc63b579ac90e52614be7ce214cf9d874c63b34/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryAutoConfiguration.java#L54

Spring's Opentelemetry auto configuration seem to use `application` as a fall back value when the property `spring.application.name` is not set. As per [Opentelemetry semantic convention](https://opentelemetry.io/docs/specs/semconv/resource/#service), the default value when the app name is not available should be `unknown_service`. 

Note: Micrometer also defaults to the value `unknown_service`. But when used with spring's auto configuration the resource is set to a different value.

mhalbritter:
Superseded by #38219.

PR #38219, opened 2023-11-06
Use unknown_service as default application name for OpenTelemetry

lenin-jaganathan:
Fixes https://github.com/spring-projects/spring-boot/issues/38218

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryAutoConfiguration.java
@@ -51,7 +51,7 @@ public class OpenTelemetryAutoConfiguration {
 	/**
 	 * Default value for application name if {@code spring.application.name} is not set.
 	 */
-	private static final String DEFAULT_APPLICATION_NAME = "application";
+	private static final String DEFAULT_APPLICATION_NAME = "unknown_service";
 
 	private static final AttributeKey<String> ATTRIBUTE_KEY_SERVICE_NAME = AttributeKey.stringKey("service.name");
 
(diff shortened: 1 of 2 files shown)


mhalbritter:
Hey, thanks for the issue and the PR.

In the OpenTelemetry SDK, they use this as the default:

```java
  private static final Resource MANDATORY =
      create(Attributes.of(SERVICE_NAME, "unknown_service:java"));
```

I wonder if we should use our default application at all. What do you think of this change?

```java
	SdkTracerProvider otelSdkTracerProvider(Environment environment, ObjectProvider<SpanProcessor> spanProcessors,
			Sampler sampler) {
		String applicationName = environment.getProperty("spring.application.name");
		Resource springResource = Resource.create((applicationName != null) ? Attributes.of(ResourceAttributes.SERVICE_NAME, applicationName) : Attributes.empty());
// ...
```

That would only set the service name if the application name is set, otherwise it will use OTels SDK defaults.

lenin-jaganathan:
Micrometer is not using OTEL SDK's so metrics will be using `unknown_service` as default. If `service.name` is not set in Resource bean, do you think it will cause inconsistencies between traces and metrics?

mhalbritter:
Could be, that would be a question for the Micrometer team, I'll point them to that issue. Let's see what they say.

mhalbritter:
So, OTels default of `unknown_service:java` is wrong when the application is running as a native image, as the format of [`service.name` is `name:process.executable.name`](https://opentelemetry.io/docs/specs/semconv/resource/#service).

I lean towards making the Boot default `unknown_service`, which is correct when running in a native image and is also the same as in Micrometer.

The only thing I'm worried is that this is a breaking change if users haven't set `spring.application.name`. We have to decide in which version to ship this.

wilkinsona:
> The only thing I'm worried is that this is a breaking change if users haven't set spring.application.name. We have to decide in which version to ship this.

I think it should be 3.2 at the very earliest and probably wait until 3.3 given where we are in 3.2's schedule.

lenin-jaganathan:
I believe for metric users moving from 3.1 to 3.2 will be a breaking change. As per [this code](https://github.com/spring-projects/spring-boot/blob/3.1.x/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpPropertiesConfigAdapter.java#L54) and [this code](https://github.com/micrometer-metrics/micrometer/blob/main/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java#L85-L87), I believe if `service.name` was not explicitly set in properties or OTEL_SERVICE_NAME is not an environmental variable 3.1 will use `unknown_service` as default. OpenTelemetry Resources were made common (between metrics and tracing) only in 3.2.

For tracing yes, it might be a breaking change given that in 3.1 "application" was used a fallback value.

mhalbritter:
We need to adjust the name in `org.springframework.boot.actuate.autoconfigure.metrics.export.otlp.OtlpPropertiesConfigAdapter`, too.

mhalbritter:

Issue #35354, opened 2023-05-08
@AutoConfigureObservability has unintentional consequences on integration tests


mhalbritter:
By default, metrics and tracing are disabled in `@SpringBootTest` integration tests. While the `MeterRegistry` gets replaced by a `SimpleMeterRegistry` which collects the metrics in-memory, we replace the Micrometer Tracing `Tracer` with a noop implementation.

There are other beans which are `@ConditionalOnEnabledTracing`, and those don't get created when running in a integration test. If users happen to have code using these beans they observe that the application is running fine when starting `main`, but fail with a missing bean when running the `@SpringBootTest`. This is strange behavior and we should think about if we want to behave this way.

We initially designed it that way so that integration tests don't send metrics / traces to monitoring systems. Ideally we would replace such beans with in memory implementations, but I don't think this is feasible for all affected beans.

jonatan-ivanov:
I remember I wanted to mention this earlier but both `Tracer` implementations (Brave/OTel) in Micrometer Tracing are in-memory implementations. They will not send the data anywhere unless you configure an exporter. I think the solution for tracing that is similar to `SimpleMeterRegistry` would be configuring the Tracer but not configuring an exporter. Or configuring an in-memory (queue) exporter (though this is by definition a memory leak if no-one consumes the queue).

kiview:
There is another (accidental) side effect we observed through this:
If users just have some specific tests annotated with `@AutoConfigureObservability`, while the majority of test does not have it, different Spring Contexts need to be created. This adds the overhead of Spring Context creation, but potentially in case of Testcontainers usage, also additional container creations. 

What would be the negative impact of enabling `@AutoConfigureObservability`as the default?

mhalbritter:
## Zipkin and Brave

Only the `ZipkinSpanHandler` sends the spans. When we disable this bean in tests, nothing get sent anymore. All the other beans could stay in place.

Another option would be to replace `AsyncReporter<Span>` with `Reporter.NOOP` or `Reporter.CONSOLE`.

mhalbritter:
If integration tests are run, tracing by default is disabled. You can enable it again with `@AutoconfigureObservability`. When tracing is disabled, we don't create the following beans:

* `OtlpHttpSpanExporter`
* `WavefrontSpanHandler`
* `WavefrontSender`
* `WavefrontBraveSpanHandler`
* `WavefrontOtelSpanExporter`
* `ZipkinSpanHandler`
[...]

PR #38273, opened 2023-11-08
Re-add @ConditionalOnEnabledTracing on BraveAutoConfiguration

paulbakker:
`@ConditionalOnEnabledTracing` used to be on the `BraveAutoConfiguration`. See this revision: https://github.com/spring-projects/spring-boot/blob/411586347c8fae00373028bac51dfd16e69661c5/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BraveAutoConfiguration.java

It's unclear to me from the git history why it was removed, and I think it's incorrect.
This is causing an issue with (testing) upgrading to 3.2 currently.


scottfrederick:
This change was made as part of #35354, and mentioned in the `3.2.0-M1` [release notes](https://github.com/spring-projects/spring-boot/wiki/Spring-Boot-3.2.0-M1-Release-Notes#observability-in-tests). Do the suggestions in the release notes resolve your issues with testing?

paulbakker:
That explains! I had missed it in the release notes.

We're doing something like this in Spring Boot Netflix to have similar behavior in tests, so this is good, actually. We'll have to rework a few assumptions, but that's easy enough.

Thanks!

PR #38234, opened 2023-11-06
Improve and Update BatchAutoConfiguration

cachescrubber:
### Improve and Update BatchAutoConfiguration

This PR will update and improve `BatchAutoConfiguration`

- Add possibility to configure a custom `ExecutionContextSerializer`
- Remove the unnecessary `JobOperator` bean  definition
- Update `JobLauncherApplicationRunner` to use `JobParameters#getIdentifyingParameters`
- Add a **Restarting a stopped or failed Job** section to the batch howto.

PR #38327, opened 2023-11-11
Update JobLauncherApplicationRunner to use JobParameters#getIdentifyingParameters

cachescrubber:
### Update JobLauncherApplicationRunner to use JobParameters#getIdentifyingParameters

Target is main, since 3.2.x is not yet available. See #38234


mhalbritter:
Thank you!

PR #38328, opened 2023-11-11
Add possibility to configure a custom ExecutionContextSerializer in BatchAutoConfiguration

cachescrubber:
### Add possibility to configure a custom ExecutionContextSerializer in BatchAutoConfiguration

See #38234


cachescrubber:
I would like to auto configure a `Jackson2ExecutionContextStringSerializer` if enables by a use property.

```
	@Bean
	@ConditionalOnMissingBean(ExecutionContextSerializer.class)
	@ConditionalOnProperty(prefix = "spring.batch.jdbc", name = "use-jackson2-context-serializer", havingValue = "false")
	public Jackson2ExecutionContextStringSerializer executionContextStringSerializer() {
		return new Jackson2ExecutionContextStringSerializer();
	}

```

@fmbenhassine WDYT?

fmbenhassine:
There is no need for a property, the presence of a bean of type `ExecutionContextSerializer` is enough to be picked up by Boot, otherwise a `DefaultExecutionContextSerializer` should be used. The last PR update LGTM.

mhalbritter:
Thank you very much!

PR #38329, opened 2023-11-12
Rename method, field name of getInitializeSchema

LenKIM:
While learning Spring batch, I noticed some variable names and functions are strange.

The function name getInitializeSchema() is confusing, so you need to change it to mode().


rishiraj88:
Thanks, @pivotal-cla , for links. I'll read and sign those as well.

philwebb:
Thanks for the suggestion but this will break back-compatibility and I think I prefer the existing names anyway.

PR #38383, opened 2023-11-17
friendly reminder

shenjianeng:
friendly reminder


philwebb:
Thanks for the suggestion but I'm not sure we should do this since we don't know what log levels the user has already applied or if increasing them will add anything beneficial. I think we should keep help about debug logging in the reference guide.

Issue #16820, opened 2019-05-10
Allow the comment prefix used in data source initialisation scripts to be configured


szumszym:
Any line in the schema script with a sql comment and without semi-colon at the end of that line causes that the next sql statement is not executed but the rest of the script will be executed anyway.

Spring Boot v2.1.4.RELEASE

DOESNT' WORK:
`DROP TABLE abc IF EXISTS;`
 `# blablabla`
`DROP TABLE efg IF EXISTS;`

DOES WORK:
`DROP TABLE abc IF EXISTS;`
 `# blablabla;`    <-- semi-colon added
`DROP TABLE efg IF EXISTS;`

(Please check if the issue occurs also for data.sql script)

thackel:
I am pretty sure SQL comments are only
`-- line comment`
and
`/* block comment */`
but not starting with `#`.

So just change your DDL to use correct SQL comments and it will work.

wilkinsona:
Spring Boot uses `ResourceDatabasePopulator` which is part of Spring Framework. By default, it uses `--` for line comments and `/* … */` for block comments. AIUI, these are the standard SQL comment types. Some databases, MySQL for example, also support `#` for line comments. The comment prefix used by `ResourceDatabasePopulator` can be configured, but Spring Boot does not expose this for its data source initialization. We can consider adding such a configuration option but I would recommend stick with the standard SQL comment syntax.

philwebb:
I think it's better that we keep ourselves aligned with the defaults in Spring Framework. A configuration option for such an edge case feels a bit too much. Only allowing standard SQL comments will also help keep scripts portable if developers ever choose to move to a different server.

PR #38385, opened 2023-11-17
Document supported SQL comment prefixes

PENEKhun:
# Situation

I am attempting to initialize a database following the instructions in the [Spring Boot documentation](https://docs.spring.io/spring-boot/docs/3.1.0/reference/html/howto.html#howto.data-initialization.using-basic-sql-scripts) under the "Initialize a Database Using Basic SQL Scripts" section.

When initializing the database with SQL files using `sql.init.mode=always`, you can specify the SQL file for initialization in the `/resources` directory (e.g., `/resources/schema.sql`).

Below is the content of the `/resources/schema.sql` file for initializing the schema:

```sql
# 1. remove table start
drop table if exists firstTable;
drop table if exists secondTable;
# 1. remove table end

# 2. create table start
create table firstTable
(
    id                   bigint auto_increment
        primary key,
    test_field           bigint               not null
);

create table secondTable
(
    id                   bigint auto_increment
        primary key,
    test_field           bigint               not null
);
# 2. create table end
```

In `schema.sql`, '#' is used for inline comments. However, since the `DEFAULT_COMMENT_PREFIX` is '--', some SQL statements were not parsed correctly, leading to issues as shown in the following image:

Parsing Issue

As a result, the table creation for `firstTable`, as specified in `schema.sql`, was ignored:

Table Creation Ignored

To avoid any potential misunderstanding, I suggest adding a mention of this issue in the documentation.

# Changes

Added the following statement to [[howto.data-initialization.using-basic-sql-scripts]]

> NOTE: When including comments in your SQL file, it is advisable to use `--` for single-line comments and `/*` and `*/` for block comments. Be cautious when using other comment formats, as they may lead to the omission of certain SQL statements during parsing.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/data-initialization.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/howto/data-initialization.adoc
@@ -62,6 +62,8 @@ While we do not recommend using multiple data source initialization technologies
 This will defer data source initialization until after any `EntityManagerFactory` beans have been created and initialized.
 `schema.sql` can then be used to make additions to any schema creation performed by Hibernate and `data.sql` can be used to populate it.
 
+NOTE: When including comments in your SQL file, it is advisable to use `--` for single-line comments and `/++*++` and `++*++/` for block comments. Be cautious when using other comment formats, as they may lead to the omission of certain SQL statements during parsing.
+
 If you are using a <<howto#howto.data-initialization.migration-tool,Higher-level Database Migration Tool>>, like Flyway or Liquibase, you should use them alone to create and initialize the schema.
 Using the basic `schema.sql` and `data.sql` scripts alongside Flyway or Liquibase is not recommended and support will be removed in a future release.
 


mhalbritter:
Hello! Which SQL dialect are you using that `#` is a valid comment? I've only encountered SQL comments with `--` so far.

PENEKhun:
@mhalbritter 

mysql and mariadb support # comments.
Please refer to the following official references

- [MySQL :: MySQL 8.0 Reference Manual :: 9.7 Comments](https://dev.mysql.com/doc/refman/8.0/en/comments.html)
- [Comment Syntax - MariaDB Knowledge Base](https://mariadb.com/kb/en/comment-syntax/)

mhalbritter:
I wonder if we should make the comment settings configurable so that you can continue to use `#` for comments.

mhalbritter:
Ah, never mind, i found an old discussion about that: https://github.com/spring-projects/spring-boot/issues/16820

mhalbritter:

PR #38390, opened 2023-11-18
Replace Function<String, String> with UnaryOperator<String> in CorrelationIdFormatter

abdullah-jaffer:
- Removed unnecessary braces -> Had to remove this change due to checkStyle failures
- Javadoc grammatical fix
- Use UnaryOperator instead of Function
- Remove redundant default constant in annotation

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java
@@ -474,7 +474,7 @@ public class LogbackLoggingSystem extends AbstractLoggingSystem implements BeanF
 	/**
 	 * {@link LoggingSystemFactory} that returns {@link LogbackLoggingSystem} if possible.
 	 */
-	@Order(Ordered.LOWEST_PRECEDENCE)
+	@Order()
 	public static class Factory implements LoggingSystemFactory {
 
 		private static final boolean PRESENT = ClassUtils.isPresent("ch.qos.logback.classic.LoggerContext",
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks! The lambda style is intentional because we like the consistency between single and multiple argument versions. E.g. `(foo) -> baz`, `(foo, bar) -> baz`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java
```diff
@@ -474,7 +473,7 @@ private void withLoggingSuppressed(Runnable action) {
 	/**
 	 * {@link LoggingSystemFactory} that returns {@link LogbackLoggingSystem} if possible.
 	 */
-	@Order(Ordered.LOWEST_PRECEDENCE)
```
wilkinsona, line 477: There are [several other places where we use `@Order(Ordered.LOWEST_PRECEDENCE)`. If we're going to change one, we should change them all. Personally, I don't think we should change any of them as I prefer to see the order explicitly declared.

abdullah-jaffer: That makes sense for the sake of consistency, I removed it because IntelliJ usually doesn't recommend explicitly mentioning default values for annotations. Have added the constant back.

philwebb:
I wasn't too keen on the Javadoc grammatical fix either I'm afraid but the Function to UnaryOperator change is a nice change so I've merged that. Thanks for contributing @abdullah-jaffer!

PR #38486, opened 2023-11-21
Optimisations for jackson package

k-korotaev:
1. "Lazy initialization" for mixin registration
2. Optimization of exception catching in Serializer/Deserializer


philwebb:
Thanks for the suggestions, but I don't think we should make these changes. The `JsonMixinModule ` should not have state and should only be called once anyway (from `JacksonAutoConfiguration`). Directly throwing the `IOException` is also not likely to buy us much and it changes the current API.

Thanks anyway.

Issue #38506, opened 2023-11-22
spring.config.activate.on-cloud-platform should consider NONE to be semantically equivalent to null


Cameron-Meyer:
Apologies if this has been reported before, couldn't find any StackOverflow posts or GitHub issues.

https://github.com/spring-projects/spring-boot/blob/152acf8a59ac952340ab653c61b5dca9649dacf7/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/ConfigDataProperties.java#L128

When attempting to use `spring.config.activate.on-cloud-platform` in conjunction with https://docs.spring.io/spring-boot/docs/current/reference/html/features.html#features.external-config.files.multi-document, there is no mechanism available to simply state "If not running on a cloud platform" without forcibly setting the `NONE` value via `spring.main.cloud-platform` through some other mechanism.

This is undesirable for libraries providing configuration files with default values for scenarios when the consumers is running on a cloud platform and for when they aren't, as it forces one of the following scenarios:

- Force consumers to specify `spring.main.cloud-platform: NONE` when running locally (ex: through their application-{local-env-name}.yml)
- Abandon `spring.config.activate.on-cloud-platform` and multi-document config files, and conditionally load the properties through custom beans and manual CloudPlatform checks
-  Abandon `spring.config.activate.on-cloud-platform` and use a different, likely more unreliable, property condition, such as `spring.config.activate.on-profile: 'DEV | LOCAL | NO-PLATFORM'`

Proposed change:
```java
		private boolean isActive(CloudPlatform cloudPlatform) {
			return this.onCloudPlatform == null 
                                       || (this.onCloudPlatform == NONE && cloudPlatform == null) 
                                       || this.onCloudPlatform == cloudPlatform);
		}
```

wilkinsona:
We think this makes sense but we're unsure about when to make the change. It could be considered a bug, in which case 3.1.x would be an option if the risk isn't too great. WDYT, @philwebb?

philwebb:
I think we should consider this an enhancement.

The original intention of the `spring.config.activate.on-cloud-platform` property was to allow additional or override configuration for a specific cloud platform, I don't think we really considered `none` being used there.

The `CloudPlatform.NONE` enum was primary added to allow users to prevent our detection logic from running. From the javadoc:

> No Cloud platform. Useful when false-positives are detected.

In other words, if we're guessing wrong for whatever reason then set this to tell us.

[...]

PR #38510, opened 2023-11-23
Make spring.config.activate.on-cloud-platform=none match when the current cloud platform is null

quaff:
Fix GH-38506


wilkinsona:
Thanks very much, @quaff.

PR #38550, opened 2023-11-25
Exception replaced with Throwable

anavrotski:
Exception replaced with Throwable (same as in Builder class).


philwebb:
Thanks for the suggestion. I have a feeling this will break binary compatibility so we'll probably have to live with it until 3.3.

philwebb:
We'll merge this in 3.3. There is a risk that existing libraries might fail in which case they'll need to migrate to `down().withException()` if they want to support Spring Boot 3.2 and 3.3+

wilkinsona:
Thanks very much, @anavrotski.

PR #38570, opened 2023-11-27
Fix Observation Filter docs

aleksandrserbin:
`WebMvcMetricsFilter` no longer exists since 3.0 and should be replaced with `ServerHttpObservationFilter`. Documentation still references the old class name


mhalbritter:

PR #38576, opened 2023-11-27
Add support for multiple flyway instances

bekoenig:
Hi,

after my comment in another discussion to [provide support for auto-configuring multiple beans](https://github.com/spring-projects/spring-boot/issues/15732#issuecomment-1774167022) I would like to contribute my solution for flyway. This is an integrated version from an internal adaption which we use in my company. The implementation is fully compatible with the current default behaviour.

To use the new behaviour you have to define properties like `spring.flyway.instances.{name}.*`. The {name} is a custom definition for the flyway instance. Every property tree overrides than the global properties which are placed in `spring.flyway.*`. In general these are default-schema and location, but all other work too. Every instance can be disabled separately. Auto completion for properties works in the IDE too.

In order to bind configuration beans to a flyway instance a new annotation `@FlywayInstance({name})` exists, which binds a bean to the specific instance by the {name} defined in the properties. When no explicit binding exists, the bean will be the default for all instances. For `FlywayConnectionDetails`, `FlywayDataSource` and `FlywayMigrationStrategy` a specific bean will be preferred. For `FlywayConfigurationCustomizer` und `Callback` the default and the instance specific beans are joined by order. Only `JavaMigration` expects an explicit binding, to avoid configuration errors. A predefined `FlywayMigrationInitializer` will only be used in the default mode. In multi instance mode every flyway bean gets its own `FlywayMigrationInitializer`.

Most tests work and I added tests for the new feature. I can add documentation based on feedback. 

Greetings,
Ben

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java
@@ -16,28 +16,14 @@
 
 package org.springframework.boot.autoconfigure.flyway;
 
-import java.sql.DatabaseMetaData;
-import java.time.Duration;
-import java.util.Collection;
 import java.util.Collections;
 import java.util.HashSet;
-import java.util.List;
-import java.util.Map;
 import java.util.Set;
-import java.util.function.BiConsumer;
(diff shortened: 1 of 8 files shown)


wilkinsona:
Thanks, @bekoenig. It's interesting to see how you've tackled the problem for Flyway specifically. Unfortunately, as I said on #15732, we cannot merge. We don't want to add something that's Flyway-specific that hampers our ability to provide a broader solution in the future.

Issue #35815, opened 2023-06-09
Java system properties can not be applied to RestTemplate HttpClient connection in some cases


alex-kormukhin:
Spring boot 3.1.0 `RestTemplateBuilder` with apache HTTP client in classpath (`org.apache.httpcomponents.client5:httpclient5`) ignore java system HTTP proxy command line options like `-Dhttp.proxyHost=127.0.0.1 -Dhttp.proxyPort=8888`.

Spring boot before 3.1 respects such settings.

For simple `RestTemplate restTemplate = restTemplateBuilder.build()` spring boot 2.7 [instantiate](https://github.com/spring-projects/spring-boot/blob/v2.7.12/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/ClientHttpRequestFactorySupplier.java#L54) [HttpComponentsClientHttpRequestFactory](https://github.com/spring-projects/spring-framework/blob/v5.3.27/spring-web/src/main/java/org/springframework/http/client/HttpComponentsClientHttpRequestFactory.java#L79) via default constructor, which create HTTP client as `HttpClients.createSystem()`, i.e. with system proxy support.

![image](https://github.com/spring-projects/spring-boot/assets/99888299/d7c8ae39-cf87-4a0f-8f54-f8fb26f7c359)

Spring boot 3.0 (if readTimeout duration is not set) also create HttpComponentsClientHttpRequestFactory via default constructor and HTTP client as [`HttpClients.createSystem()`](https://github.com/apache/httpcomponents-client/blob/rel/v5.1.4/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClients.java#L65).

![image](https://github.com/spring-projects/spring-boot/assets/99888299/75f432ad-6939-48b2-b468-34a1adcda670)

But spring boot 3.1 create HTTP client as `HttpClientBuilder.create().setConnectionManager(connectionManager).build()`. Without [`useSystemProperties()`](https://github.com/apache/httpcomponents-client/blob/rel/v5.1.4/httpclient5/src/main/java/org/apache/hc/client5/http/impl/classic/HttpClients.java#L65) as `HttpClients.createSystem()` makes.

![image](https://github.com/spring-projects/spring-boot/assets/99888299/8d07178e-c7b6-4cb8-96bd-847f710a6ef4)

Suggestion: add useSystemProperties() to createHttpClient method like this
```
return HttpClientBuilder.create()
    .useSystemProperties()
[...]

PR #38591, opened 2023-11-28
System SSL certificates are not used by the Apache HTTP Client in a RestTemplate built with RestTemplateBuilder

tomfrenken:
With this change, the SSL certificates that are provided by the system are picked up by default when using the `RestTemplateBuilder`.

This aligns with the implementation of `new RestTemplate()` which is already doing this, and can be considered a follow-up to a [previous (partial) fix](https://github.com/spring-projects/spring-boot/commit/c22cbb0e2dfbc71ba19d83084af962f207c152fb) from 5 months ago relating to [this issue](https://github.com/spring-projects/spring-boot/issues/35815).

(For context, `RestTemplate` is using `SimpleClientHttpRequestFactory` by default, which appears to be picking up the SSL certificates from the environment)

In our Cloud Foundry environment, where the SSL certificates are provided by the system through [a build-pack](https://github.com/cloudfoundry/java-buildpack-security-provider/blob/8117fa264727de9688db6539158d445918f3a472/src/main/java/org/cloudfoundry/security/CloudFoundryContainerKeyManagerFactory.java#L72C34-L72C54), we encountered a problem with SSL certificates. The SSL certificates from the system work fine with `new RestTemplate()`, but not with `new RestTemplateBuilder.build()`. This seems unintentional, especially considering a previous (partial) fix that added some functionality (through `useSystemProperties()`) but missed the SSL properties.

As I am quite new to the Java & Spring Boot ecosystem I am not entirely sure how to best test this, but the issue we are facing seems to be a spring-boot issue to me, which can hopefully be fixed with my PR.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/ClientHttpRequestFactories.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/ClientHttpRequestFactories.java
@@ -31,7 +31,6 @@ import javax.net.ssl.SSLSocketFactory;
 import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
 
-import okhttp3.OkHttpClient;
 import org.apache.hc.client5.http.classic.HttpClient;
 import org.apache.hc.client5.http.impl.classic.HttpClientBuilder;
 import org.apache.hc.client5.http.impl.io.PoolingHttpClientConnectionManager;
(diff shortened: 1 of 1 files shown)


tomfrenken:
Probably worth mentioning that to reproduce this issue, you have to have Apache's HTTP Client in your classpath, otherwise RestTemplateBuilder won't pick up the changes of this PR. :D

mhalbritter:

PR #38599, opened 2023-11-29
Use switch statement where appropriate

jinsim:
Hi,
The current code uses multiple if statements in a while loop. 
Since the index is fixed and conditions are mutually exclusive, replacing the if statements with a switch statement could improve efficiency. 
This pull request suggests refactoring for cleaner code and potential performance gains.


jinsim:
@wilkinsona 
Thank you for your response. I appreciate your valid explanation. 
However, is there a specific reason for using repetitive if statements instead of an if-else statement? 
Since the index is fixed, it seems more appropriate to use if-else. 
If my understanding is incorrect, please let me know.
Thank you.

philwebb:
I don't think there's a lot of benefit either way, as Andy said the code isn't performance critical and we don't want to make a change just for the sake of it.

jinsim:
@philwebb 
Thank you for your prompt response.

PR #38622, opened 2023-12-01
upgrade logback dependency version to resolve CVE-2023-6378

ZheSun88:


andersthorbeck:
@bclozel So, just to be clear, there is no fast-track to upgrading the managed dependency versions even if there is a CVE which would be fixed by doing so? I ask because I cannot find it explicitly mentioned in the 3rd party upgrade policy you linked to.

bclozel:
> So, just to be clear, there is no fast-track to upgrading the managed dependency versions even if there is a CVE which would be fixed by doing so?

The fast-track option is to override the version in your project - this is the safest option that we officially recommend. If we were to release Spring Boot maintenance versions for every CVE out there 1) there would be a lot of those 2) we would always lack the context of the CVE and how that applies to apps and 3) we would never be as fast as the third party since we learn about those at the same time as everyone else.

Issue #36073, opened 2023-06-26
Upgrade to Jetty 12


wilkinsona:
Work in progress is available in https://github.com/wilkinsona/spring-boot/tree/jetty-12. Related Jetty issues with 12.0.0.beta2:

- https://github.com/eclipse/jetty.project/issues/9972 (inconvenient)
- https://github.com/eclipse/jetty.project/issues/9973 (blocker)
- https://github.com/eclipse/jetty.project/issues/9984 (blocker)

Sineaggi:
Jetty 12 has been released! https://github.com/eclipse/jetty.project/releases/tag/jetty-12.0.0

mikekuzak:
Hi, 

Does the spring-boot-starter for Spring Boot 3 work now with Jetty 12 ? 
I understand SP3 doesn't work with Jetty 11, because it's not supporting  EE10 (Servlet 6.0)

wilkinsona:
> Does the spring-boot-starter for Spring Boot 3 work now with Jetty 12 ?

No, not yet. There are a significant number of breaking API changes in Jetty 12 to which Spring Boot will have to adapt. We hope to support Jetty 12 in Spring Boot 3.2 and Spring Framework 6.1.

> I understand SP3 doesn't work with Jetty 11, because it's not supporting EE10 (Servlet 6.0)

Spring Boot 3 works with Jetty 11 but you have to downgrade the Servlet API to 5.0. https://github.com/spring-projects/spring-boot/issues/33044 contains various ways to do that with both Maven and Gradle.

joakime:
> > Does the spring-boot-starter for Spring Boot 3 work now with Jetty 12 ?
> 
> No, not yet. There are a significant number of breaking API changes in Jetty 12 to which Spring Boot will have to adapt. We hope to support Jetty 12 in Spring Boot 3.2 and Spring Framework 6.1.

Is there an active branch with this effort where I can participate in?

wilkinsona:
Thanks, @joakime. https://github.com/spring-projects/spring-framework/issues/30698 is the next piece in the puzzle. I don't know if there's an active branch for that work.

lzysuqianqiu:
Jetty 12.0.1 is GA now,any progress

wilkinsona:
Thanks, @lzysuqianqiu. As you can see from [the branch](https://github.com/wilkinsona/spring-boot/tree/jetty-12) that's linked in this issue's description, we are already aware of that.

[...]

PR #38632, opened 2023-12-01
Correct references to Jetty's HTTP2 server module

gavlyukovskiy:
Update documentation to use `org.eclipse.jetty.http2:jetty-http2-server` for Jetty HTTP2 server support.

All jetty dependencies were updated in https://github.com/spring-projects/spring-boot/issues/36073, this only updates [reference documentation](https://docs.spring.io/spring-boot/docs/3.2.0/reference/htmlsingle/#howto.webserver.configure-http2.jetty) and the exception message.


wilkinsona (changes requested):
Thanks for the PR, @gavlyukovskiy. Can you revert the code changes as they are unrelated to correcting the docs.

If we want to improve the diagnostics when a required dependency is missing, we may want to consider a specific exception type and a failure analyzer rather than a plain state assertion failure. For this side of things, please open an issue for consideration.

gavlyukovskiy:
@wilkinsona I've reverted changes to `JettyServletWebServerFactory`

mhalbritter:
Thank you!

PR #38702, opened 2023-12-07
Use idiomatic AssertJ assertions

quaff:
1. Replace assertThat(x.isEmpty()).isTrue() with assertThat(x).isEmpty() Search for : assertThat\((.+).isEmpty\(\)\).isTrue\(\) Replace with : assertThat($1).isEmpty()
Search for : assertThat\((.+).isEmpty\(\)\).isFalse\(\) Replace with : assertThat($1).isNotEmpty()

2. Replace assertThat(x.iterator().next()) with assertThat(x).element(0) Search for : assertThat\((.+).iterator\(\).next\(\)\) Replace with : assertThat($1).element(0)

3. Replace assertThat(x.get(i)). with assertThat(x).element(i). Search for : assertThat\((.+)\.get\((\d+)\)\)\.
Replace with : assertThat($1).element($2).

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader-classic/src/test/java/org/springframework/boot/loader/jar/StringSequenceTests.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader-classic/src/test/java/org/springframework/boot/loader/jar/StringSequenceTests.java
@@ -92,12 +92,12 @@ class StringSequenceTests {
 
 	@Test
 	void isEmptyWhenEmptyShouldReturnTrue() {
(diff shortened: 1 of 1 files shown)


quaff:
> It's more verbose and I don't think it makes the code easier to read.

It will report more meaningful error not just for readable consideration.

philwebb:
I like 1 but I don't think 2 or 3 give enough value given the number of files they change.

quaff:
OK, I've updated the commit.

mhalbritter:
Thank you!

PR #38703, opened 2023-12-07
Add property for maximum number of reactive sessions

JohnNiang:
By default, the maxinum number of sessions is [10,000](https://github.com/spring-projects/spring-framework/blob/302cdeeee67f9291be45147ffc5b543bfcb32d77/spring-web/src/main/java/org/springframework/web/server/session/InMemoryWebSessionStore.java#L51) while using Reactive server. Our users always receive errors that reach the maxinum session limit. See [here](https://github.com/spring-projects/spring-framework/blob/302cdeeee67f9291be45147ffc5b543bfcb32d77/spring-web/src/main/java/org/springframework/web/server/session/InMemoryWebSessionStore.java#L281) for source code.

This PR allows users to freely configure the maxinum number of sessions by setting property `server.reactive.session.max-sessions`.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -329,6 +329,11 @@ public class ServerProperties {
 			@DurationUnit(ChronoUnit.SECONDS)
 			private Duration timeout = Duration.ofMinutes(30);
 
+			/**
+			 * The maximum number of sessions that can be stored.
+			 */
+			private int maxSessions = 10_000;
+
 			@NestedConfigurationProperty
 			private final Cookie cookie = new Cookie();
 
@@ -338,6 +343,14 @@ public class ServerProperties {
 
 			public void setTimeout(Duration timeout) {
 				this.timeout = timeout;
+			}
+
+			public int getMaxSessions() {
+				return this.maxSessions;
+			}
+
+			public void setMaxSessions(int maxSessions) {
+				this.maxSessions = maxSessions;
 			}
 
 			public Cookie getCookie() {
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
@@ -332,7 +332,10 @@ public class WebFluxAutoConfiguration {
 		public WebSessionManager webSessionManager(ObjectProvider<WebSessionIdResolver> webSessionIdResolver) {
 			DefaultWebSessionManager webSessionManager = new DefaultWebSessionManager();
 			Duration timeout = this.serverProperties.getReactive().getSession().getTimeout();
-			webSessionManager.setSessionStore(new MaxIdleTimeInMemoryWebSessionStore(timeout));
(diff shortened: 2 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -329,6 +329,11 @@ public static class Session {
 			@DurationUnit(ChronoUnit.SECONDS)
 			private Duration timeout = Duration.ofMinutes(30);
 
+			/**
+			 * The maximum number of sessions that can be stored.
+			 */
+			private Integer maxSessions;
```
wilkinsona, line 335: This should be an `int` with a default value that matches that of `InMemoryWebSessionStore`. Providing a default improves the metadata that's used for auto-completion when editing application properties and YAML files. A test should be added to assert that the defaults are in sync.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
```diff
@@ -332,7 +332,12 @@ public LocaleContextResolver localeContextResolver() {
 		public WebSessionManager webSessionManager(ObjectProvider<WebSessionIdResolver> webSessionIdResolver) {
 			DefaultWebSessionManager webSessionManager = new DefaultWebSessionManager();
 			Duration timeout = this.serverProperties.getReactive().getSession().getTimeout();
-			webSessionManager.setSessionStore(new MaxIdleTimeInMemoryWebSessionStore(timeout));
+			var maxSessions = this.serverProperties.getReactive().getSession().getMaxSessions();
+			var sessionStore = new MaxIdleTimeInMemoryWebSessionStore(timeout);
```
wilkinsona, line 336: We don't use `var` in Spring Boot's code.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java
```diff
@@ -332,7 +332,12 @@ public LocaleContextResolver localeContextResolver() {
...
+			if (maxSessions != null) {
```
wilkinsona, line 337: This `null` check can be removed once the `Integer` is an `int`.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfigurationTests.java
```diff
@@ -620,6 +622,17 @@ void customSessionTimeoutConfigurationShouldBeApplied() {
 			})));
 	}
 
+	@Test
+	void customSessionMaxSessionsConfigurationShouldBeApplied() {
+		this.contextRunner.withPropertyValues("server.reactive.session.max-sessions:123").run((context) -> {
+			var sessionManager = context.getBean(WebSessionManager.class);
+			assertThat(sessionManager).isInstanceOf(DefaultWebSessionManager.class);
+			var sessionStore = ((DefaultWebSessionManager) sessionManager).getSessionStore();
```
wilkinsona, line 630: Don't use `var` please.

wilkinsona (commented):
Thanks for the PR, @JohnNiang. I've left a few comments for your consideration.

JohnNiang:
Hi @wilkinsona , thank you for code review. I have adjusted some of the code according to your feedback.

mhalbritter:
Thank you!

Issue #38522, opened 2023-11-23
Spring Boot 3.2 is not compatible with older versions of Liquibase


ebussieres:
Spring-boot 3.2.0 seems to have introduced a regression. I am using liquibase with a liquibase parameters in a sql script.

When setting the flag `test-rollback-on-update` to true, there's a checksum validation error when performing the rollback step. It seems to be related when you use a liquibase parameter in a sql script (see [here](https://github.com/ebussieres/spring-liquibase-bug/blob/master/src/main/resources/application.properties#L2) and [here](https://github.com/ebussieres/spring-liquibase-bug/blob/master/src/main/resources/db/changelog/testWithParams.sql))

There's a [demo ](https://github.com/ebussieres/spring-liquibase-bug) available here which demonstrate the regression. Just run the [test ](https://github.com/ebussieres/spring-liquibase-bug/blob/master/src/test/java/com/example/demo/DemoApplicationTests.java)to see the problem.

If you change the spring boot version to 3.1.6, eveything's working fine.

wilkinsona:
I don't think this is a Spring Boot regression. The same problem occurs with Spring Boot 3.1.6 if you use Liquibase 4.24.0 (the default version in Spring Boot 3.2) or 4.25.0 (the latest version at the time of writing). Please report the problem to the Liquibase team.

Unfortunately, you cannot downgrading Liquibase to 4.23.3 with Spring Boot 3.2 (the default version of Spring Boot 3.1.6) as it does not contain the `liquibase.UpdateSummaryOutputEnum` class [upon which Spring Boot now depends](https://github.com/spring-projects/spring-boot/pull/38274).

ebussieres:
I've opened an [issue](https://github.com/liquibase/liquibase/issues/5278) to the Liquibase team too

wilkinsona:
Thanks, @ebussieres. It looks like it's going to be fixed on the Liquibase side. I'll leave this one open for now though as I think we should consider making a change to `LiquibaseProperties` so that it doesn't directly rely upon the recently introduced Liquibase enums.

EvaristeGalois11:
> it does not contain the liquibase.UpdateSummaryOutputEnum class upon which Spring Boot now depends.

> we should consider making a change to LiquibaseProperties so that it doesn't directly rely upon the recently introduced Liquibase enums.

[...]

PR #38722, opened 2023-12-10
Set default values for showSummary and showSummaryOutput

izeye:
This PR changes to set default values for `showSummary` and `showSummaryOutput` in `LiquibaseProperties` to align with its configuration metadata.

See gh-38522


wilkinsona:
The `null` values are intentional. Making this change will break compatibility with earlier versions of Liquibase and cause `Liquibase423AutoConfigurationTests` to fail, effectively reverting #38522.

izeye:
@wilkinsona Ah, sorry, I missed the point, but I thought default values in configuration metadata are for explicit default values in configuration properties, not for implicit default values. Am I misunderstanding it?

wilkinsona:
They can be used for implicit default values, too, as is the case here. https://github.com/spring-projects/spring-boot/commit/3e4e59a8f03bc5ca5d7cad90e0e58e8bcf6a277d updated the tests to ensure that the metadata and the implicit defaults are in sync.

PR #38731, opened 2023-12-11
Fix efficient.adoc

pblanchardie:
Fix JarLauncher package: `org.springframework.boot.loader` (not in a `launch` subpackage)


mhalbritter:
Hello,

there are two implementations of `JarLauncher`. The new one is at `org.springframework.boot.loader.launch.JarLauncher`, the old one is at `org.springframework.boot.loader.JarLauncher`. The docs point to the new one, which is correct.

PR #38739, opened 2023-12-12
Cleanup java sources

quaff:
1. use `.isEmpty()` where feasible
2. remove unnecessary `static` modifier of inner record or interface type
3. remove unnecessary `final` modifier of private or static method
4. remove unnecessary `toString()`


philwebb:
Thanks very much @quaff. I've split up the commits and merged this into 3.1.x and 3.2.x.

PR #38740, opened 2023-12-12
Ban call of URLEncoder.encode/URLDecoder.decode(String,String)

quaff:
Use URLEncoder.encode/URLDecoder.decode(String,Charset) instead


philwebb:
Nice one! Thanks @quaff.

Issue #38736, opened 2023-12-11
Make JooqExceptionTranslator conditional 


MelleD:
I would like to set my own Exception Translator or ExecuteListener to translate exception. Currently the exception is always translated twice.
Can't the bean be set conditionally?  What is the use case for the Order to have 2 beans?

See `JooqAutoConfiguration`

```java
@Bean
@Order(0)
public DefaultExecuteListenerProvider jooqExceptionTranslatorExecuteListenerProvider() {
	return new DefaultExecuteListenerProvider(new JooqExceptionTranslator());
}
```

Proposal if only 1 provider is necessary
```java
@Bean
@ConditionalOnMissingBean
public DefaultExecuteListenerProvider jooqExceptionTranslatorExecuteListenerProvider() {
	return new DefaultExecuteListenerProvider(new JooqExceptionTranslator());
}
```

Quick Fix with order
```java
@Bean
@Order(0)
@CondationalOnProperty("spring.xyz.jooq.translator.enabled")
public DefaultExecuteListenerProvider jooqExceptionTranslatorExecuteListenerProvider() {
	return new DefaultExecuteListenerProvider(new JooqExceptionTranslator());
}
```

Or own interface for jOOQ Bean Translator

`public interface JooqExceptionTranslator extends ExecuteListener`

```java
@Bean
@Order(0)
public DefaultExecuteListenerProvider jooqExceptionTranslatorExecuteListenerProvider(JooqExceptionTranslator translator) {
	return new DefaultExecuteListenerProvider(translator);
}

@Bean
@ConditionalOnMissingBean
public JooqExceptionTranslator jooqExceptionTranslator() {
	return new DefaultJooqExceptionTranslator();
}
```

What do you think?

philwebb:
All `ExecuteListenerProvider` beans currently get passed to `org.jooq.impl.DefaultConfiguration.set(...)` so I don't think we can make `DefaultExecuteListenerProvider` conditional. That would help in your case, but not for folks that want a `ExecuteListenerProvider` bean as well as our `JooqExceptionTranslator`.

We'd probably also want to avoid `@CondationalOnProperty("spring.xyz.jooq.translator.enabled")` because as a general rule we try not to add too many `*.enabled` properties.

I think the third option isn't too bad and perhaps we could even change the existing `JooqExceptionTranslator` to make it easier to subclass.

[...]

PR #38762, opened 2023-12-13
Add conditional bean for jOOQ exception translator

MelleD:
Add conditional bean for jOOQ exception translator

#38736


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jooq/JooqExceptionTranslator.java
```diff
@@ -30,68 +30,17 @@
...
-		SQLDialect dialect = context.configuration().dialect();
-		if (dialect != null && dialect.thirdParty() != null) {
-			String dbName = dialect.thirdParty().springDbName();
-			if (dbName != null) {
-				return new SQLErrorCodeSQLExceptionTranslator(dbName);
-			}
-		}
-		return new SQLStateSQLExceptionTranslator();
-	}
+public interface JooqExceptionTranslator extends ExecuteListener {
```
mhalbritter, line 39: This change breaks backwards compatability. Unfortunately `JooqExceptionTranslator` is public and therefore we need to either deprecate it and then replace it. Or we leave it in place as it is and let users subclass it if they want to change something.

mhalbritter (changes requested):
Hello,

thanks for the PR. I've added a comment regarding breaking the public API. Please take a look.

MelleD:
> Hello,
> 
> thanks for the PR. I've added a comment regarding breaking the public API. Please take a look.

Yes I can change the name back. Do you have proposal for the other class names?

mhalbritter:
For what class are you looking for a name? I'd leave `JooqExceptionTranslator` as it is, create an auto-configuration bean method for it and use it in the `jooqExceptionTranslatorExecuteListenerProvider` method. No new classes are needed, or am I missing something?

mhalbritter:
Which new interface? I'd leave `JooqExceptionTranslator` the class it is, and add a bean method for it. If users want to customize the exception translation, create a subclass of `JooqExceptionTranslator`, override some methods and provide it as a bean. The auto-configuration will then pick up the user class and exception translation is customized. That would work, wouldn't it?

wilkinsona:
>  But looks not really nice to create subclass which using nothing from the parent class just to have the same type to provide a bean

There was a proposal in #38736 to "change the existing JooqExceptionTranslator to make it easier to subclass". I think that makes sense and would be less disruptive than the introduction of an interface.

MelleD:
> > But looks not really nice to create subclass which using nothing from the parent class just to have the same type to provide a bean
> 
> There was a proposal in #38736 to "change the existing JooqExceptionTranslator to make it easier to subclass". I think that makes sense and would be less disruptive than the introduction of an interface.

Yes third option is with interface. And also you can subclass it. That's directly what the proposal was and what I did.
For me this make a lot of sense.

wilkinsona:
I don't think we want a new interface. Instead, we'd prefer for `JooqExceptionTranslator` to remain a class and for it to be updated so that it's more useful as a super-class. This is what was proposed by @philwebb in #38762. For example, perhaps `getTranslator(ExecuteContext)` can be made `protected` so that it can be used by subclasses.

MelleD:
> I don't think we want a new interface.

But why?

wilkinsona:
It's an interface in your current proposal:

```java
public interface JooqExceptionTranslator extends ExecuteListener
```

MelleD:
> It's an interface in your current proposal:
> 
> ```java
> public interface JooqExceptionTranslator extends ExecuteListener
> ```

Yes I would like to rename the interface + rename the translator class back to JooqExceptionTranslator.
That’s it and you have the advantage of both. You can subclass the JooqExceptionTranslator or implement a own class.

philwebb:
> Yes I would like to rename the interface + rename the translator class back to JooqExceptionTranslator.
That’s it and you have the advantage of both. You can subclass the JooqExceptionTranslator or implement a own class

Although the interface is more flexible, in this instance I think it's best if we make only the smallest change possible required to support the feature. We've learnt from experience that it's harder to remove code than add it, so from our point of view it would be better not to introduce new public API before we need to.

MelleD:
> > Yes I would like to rename the interface + rename the translator class back to JooqExceptionTranslator.
> > That’s it and you have the advantage of both. You can subclass the JooqExceptionTranslator or implement a own class
> 
> Although the interface is more flexible, in this instance I think it's best if we make only the smallest change possible required to support the feature. We've learnt from experience that it's harder to remove code than add it, so from our point of view it would be better not to introduce new public API before we need to.

Yes I know, for our use case we don't need the `JooqExceptionTranslator` to subclass it. We just need the interface. So there is a use case. 

I honestly don't understand the problem at all.
I haven't heard any reason as to what exactly the problem is, other than personal opinion.

Although it is actually best practice to have an interface and an implementation.

What exactly is the problem?

philwebb:
> Although it is actually best practice to have an interface and an implementation.

There's some subtly when it comes to best practices for a general project and best practices for an open source project with many users a long support window. In a general project, the cost of introducing new API isn't that great since the the project is self contained and it's easy to change things without impacting others.

For a project such as ours, we need to be more intentional about how and when we introduce new API. We need to consider back compatibility issue and long-term maintenance concerns.

For example, we can't accept the pull-request as it stands because it changes `JooqExceptionTranslator` from a class to an interface. This could break existing users and will break binary back-compatibility. That means we have to leave `JooqExceptionTranslator` as a class. We could introduce a new interface and retrofit it, but then we have an interesting problem of naming. What should we call the new interface given that `JooqExceptionTranslator` is already taken.

If we zoom out a bit, there's a lot of advantages to not adding an interface. We don't break back-compatibility and we don't need to come up with any new class names or introduce new public API. Furthermore, we've not prevented ourselves from adding an interface later if we decide that one is really needed.

>  I haven't heard any reason as to what exactly the problem is, other than personal opinion.

This is true, but it's personal opinion that's been formed from a lot of experience managing this project. As a general philosophy, making the smallest change possible required to support a given use-case has historically served us well.

MelleD:
I understand the reasons, but I don't see it here because the interface provides exactly the same thing as the class. It is just as public as it currently is. There is no new public API, because there is already an interface --> ExecuteListener ;).
There is simply no difference except that it is more flexible and cleaner.
In addition, as already mentioned, for our use case it perfectly fit .We don't need the parent class. That's actually better for backwards compatibility, because if you deprecated the class and removed it, we don't even notice it because we don't have to inherit from it. One dependency less.

How I said:
I will revert the current class name (understand that this i breaking) and I will also find a name for the interface, but if there are already concrete ideas, you can simply change the PR as you think. So for me it's okay you can takeover the PR.

philwebb:
Thanks for the updates to the code @MelleD. Having looked at your changes I've changed my mind about the interface and I agree that it's the better approach. I've merged this into `main` and updated the interface name to `ExceptionTranslatorExecuteListener`. I've also taken the opportunity to deprecate `JooqExceptionTranslator` since we can now hide the implementation behind some statics on the interface.

Thanks for the PR!

PR #38771, opened 2023-12-14
Add timeOffset to CacheProperties.Redis

liudaac:
Add a new configuration option "timeOffset" in Redis to introduce a random time offset within the range [0, timeOffset) added to the specified timeToLive. This is aimed at effectively preventing cache avalanche issues at a specific time point in scenarios involving bulk cache refreshing.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CacheProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/CacheProperties.java
@@ -215,6 +215,11 @@ public class CacheProperties {
 		 * Entry expiration. By default the entries never expire.
 		 */
 		private Duration timeToLive;
+		
+		/**
+		 * Random time displacement to solve the cache avalanche issue.
+		 */
+		private Duration timeOffset;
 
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the suggestion but I wonder if this change is too broad. As I understand it, cache avalanche is only a concern when adding many entries at the same time such that they would all expire at the same time. The change proposed here will affect all entries that are added to Redis, not just those that are added in bulk.

What API calls do you typically use to perform bulk inserts? I wonder if this would be better handled in a way that the randomised TTL only applies to those calls.

wilkinsona:
> In my view, it seems more like an example, providing developers with a direction for customized development.

What you have proposed isn't an example. It will affect everyone using a Redis-based cache in a Spring Boot application.

> change the type of "redis" in CacheProperties from Redis to a new type like (Redis, Map<String, Redis>)

I don't think we can do that. It would be a breaking change that would affect everyone using a Redis-based cache yet it would only be of benefit to those performing bulk inserts.

I think it would be better to implement this when needed through your own `RedisCacheManagerBuilderCustomizer` bean rather than Spring Boot providing built-in support. The customizer can configure the TTL function as needed, both for the default cache configuration and for specific caches.

Thanks anyway for the proposal.

PR #38772, opened 2023-12-14
Add jsonb parser

xiaoma20082008:
Add a new JsonbParser based on jakarta.json


mhalbritter:
Hello,

thanks for the PR, but I don't think we need a 4th implementation of a JSON parser. In most of the Spring Boot applications Jackson should be on the classpath anyway, and if not, it will fall back to Basic (or GSON), which is totally fine. I don't think adding johnzon to the mix will improve things. Thanks anyway!

PR #38821, opened 2023-12-14
JSON starter without spring-web

michaldo:
spring-web uses JSON, but JSON does not use spring-web


philwebb:
The JSON starters make use of APIs in `spring-web` so we need to keep this dependency.

PR #38839, opened 2023-12-16
Pulsar authentication param properties cause IllegalStateException with Pulsar Client 3.1.0 

onobc:
The Pulsar authentication is currently configured using a map of auth parameters. However, this avenue is deprecated in Pulsar and instead needs to use an encoded auth parameters string. This commit simply takes the auth params map and converts them to the expected encoded json string of auth parameters.

Resolves https://github.com/spring-projects/spring-pulsar/issues/500

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
@@ -19,6 +19,7 @@ package org.springframework.boot.autoconfigure.pulsar;
 import java.time.Duration;
 import java.util.ArrayList;
 import java.util.Map;
+import java.util.TreeMap;
 import java.util.concurrent.TimeUnit;
 import java.util.function.BiConsumer;
 import java.util.function.Consumer;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
```diff
@@ -71,13 +73,23 @@ void customizeAdminBuilder(PulsarAdminBuilder adminBuilder, PulsarConnectionDeta
...
-		if (StringUtils.hasText(properties.getPluginClassName())) {
+		if (!StringUtils.hasText(properties.getPluginClassName())) {
+			return;
+		}
+		try {
+			// sort keys for testing and readability
+			Map<String, String> params = new TreeMap<>(properties.getParam());
+			String authParamString;
 			try {
-				authentication.accept(properties.getPluginClassName(), properties.getParam());
```
onobc, line 76: @philwebb it looks like going w/ the map of auth params was a short-lived avenue as the use of auth params map has been deprecated in favor of the auth params string. I think we should keep the properties using the map approach and simply json encode them to the expected auth param string (implementation detail).

onobc:
Thank you very much @philwebb

PR #38842, opened 2023-12-16
Consistently log default context paths for embedded servers

### LINKED ISSUE #36149: Make WebServers' started log messages more consistent
Working on #36103 showed that there are some inconsistencies in how we log the Servlet WebServer's context path:

- Jetty logs `'/'` by default
- Tomcat logs `''` by default
- Undertow logs nothing by default

There are also some similar inconsistencies on the reactive side where there is no (configurable) context path:

- Jetty logs `'/'`
- Tomcat logs `''`
- Undertow logs nothing
- Netty logs nothing

We should try to make things as consistent as possible, removing as many of the optional regex groups as possible from the tests added in #36103.

vicaba:
This PR is intended to fix #36149 with the aim to be consistent when the embedded servers start and log their context path as in `Jetty started on port 8080 (http/1.1) with context path '/'`.

My first approach is to output on every *start server* line the context path, which by default should be `/`. However, I've seen that some servers (underlying embedded servers) present some problems when I try to do so.

For example, for `undertow`, the only way to get the context path is during runtime (i.e., on receiving a request), context path can't be retrieved when building the object. If we want to set the context path, we need to modify more files. (there is a longer explanation regarding the context path in #36149 )
For `netty`, there is no concept of context path, the only reference to *path* is to the network address path.

So, for now, I'm only submitting changes for `tomcat`, as the default context path was *nothing*, I've changed it to `/`. This was already the case for `jetty`.

I am wondering if we should even log the context path and if it has any use at all.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java
@@ -256,7 +256,7 @@ public class TomcatWebServer implements WebServer {
 	}
 
 	String getStartedLogMessage() {
-		return "Tomcat started on " + getPortsDescription(true) + " with context path '" + getContextPath() + "'";
+		return "Tomcat started on " + getPortsDescription(true) + " with context path '" + getPathFromEmbeddedContext() + "'";
 	}
 
 	private void checkThatConnectorsHaveStarted() {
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, @vicaba. Unfortunately, this isn't really what we had in mind for #36149. As stated in that issue, the goal was to remove "as many of the optional regex groups as possible from the tests added in https://github.com/spring-projects/spring-boot/pull/36103". To do that, we need to make updates across all four web servers for both reactive and imperative apps. I'll push something shortly for #36194. This in any case for taking a look at this one.

PR #38884, opened 2023-12-20
otel tracing support for env vars

zeitlinger:
Fall back to otel env vars for tracing endpoint, headers, and resource attributes, in a similar way as micrometer does for metrics.

This gives users the option to use the same settings for tracing and metrics - and even the same as in other languages.

Why no spring properties?
- to make it more similar to metrics
- resource attributes and headers need to be parsed from a string - this approach doesn't require a converter.

Note: tests will be added later - need a hint how to test env vars.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/otlp/OtlpTracingConfigurations.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/otlp/OtlpTracingConfigurations.java
@@ -97,17 +97,17 @@ class OtlpTracingConfigurations {
 			return url;
 		}
 		Map<String, String> env = System.getenv();
-	            String endpoint = env.get("OTEL_EXPORTER_OTLP_TRACES_ENDPOINT");
-	            if (endpoint == null) {
-	                endpoint = env.get("OTEL_EXPORTER_OTLP_ENDPOINT");
-	            }
-	            if (endpoint == null) {
-	                endpoint = "http://localhost:4318/v1/traces";
-	            }
-	            else if (!endpoint.endsWith("/v1/traces")) {
-	                endpoint = endpoint + "/v1/traces";
-	            }
-	            return endpoint;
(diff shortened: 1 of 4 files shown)


zeitlinger:
I'm wondering if this logic should actually be in micrometer tracing, since it only depends on env vars.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/otlp/OtlpTracingConfigurations.java
```diff
@@ -76,15 +80,55 @@ static class Exporters {
...
 		}
 
 	}
 
+	private static String getUrl(OtlpTracingConnectionDetails connectionDetails) {
+		String url = connectionDetails.getUrl();
+		if (url != null) {
+			return url;
+		}
+		Map<String, String> env = System.getenv();
```
nkonev, line 99: Why don't use the existing Spring's environment ?
https://docs.spring.io/spring-framework/reference/core/beans/environment.html
> from a variety of sources: properties files, JVM system properties, system environment variables, JNDI, servlet context parameters, ad-hoc Properties objects, Map objects, and so on.

wilkinsona:
We discussed this today and our feeling is that if this logic belongs anywhere, it should go in Micrometer as it doesn't appear to be Spring Boot-specific. That said, if we wanted to use Spring's `Environment` rather than working directly with `System.getEnv()`, it could not be done in Micrometer. We'll discuss further soon, hopefully next week.

zeitlinger:
> We discussed this today and our feeling is that if this logic belongs anywhere, it should go in Micrometer as it doesn't appear to be Spring Boot-specific. 

I think this would be a good idea.

wilkinsona:
We discussed this again today and concluded that this does belong in Micrometer tracing as a companion to [the existing, similar logic for metrics](https://github.com/micrometer-metrics/micrometer/blob/771a494dba4b7fb164b0da195566d156918eccf5/implementations/micrometer-registry-otlp/src/main/java/io/micrometer/registry/otlp/OtlpConfig.java#L56-L60). Thanks anyway for the proposal.

PR #38900, opened 2023-12-21
Adding MIME Type Mapping for .mjs Extension in DefaultMimeMappings

dlemaignent:
### Description

This pull request addresses the issue where the server responds with a MIME type of "application/octet-stream" for JavaScript module scripts with the ".mjs" extension. The problem manifests as the following error:

```
Failed to load module script: Expected a JavaScript module script but the server responded with a MIME type of "application/octet-stream". Strict MIME type checking is enforced for module scripts per HTML spec.
```

I have tested this change in a local environment using the pdfjs-dist component with the following dependency:

```xml
<dependency>
    <groupId>org.webjars.npm</groupId>
    <artifactId>pdfjs-dist</artifactId>
    <version>4.0.269</version>
</dependency>
```
and using script tag like :

```html
<script type="module" th:src="@{/webjars/pdfjs-dist/build/pdf.mjs}"></script>
<script type="module" th:src="@{/webjars/pdfjs-dist/build/pdf.worker.mjs}"></script>
<script type="module" th:src="@{/webjars/pdfjs-dist/web/pdf_viewer.mjs}"></script>
```

The test results confirm that the proposed changes resolve the reported issue.

To resolve this issue, I propose adding a default MIME type mapping for the ".mjs" extension to "text/javascript" directly within the Spring Boot application. This eliminates the need for a separate class like `CustomMimeMapping` with `WebServerFactoryCustomizer` like :

```java
@Override
public void customize (ConfigurableServletWebServerFactory factory){
    MimeMappings mappings = new MimeMappings(MimeMappings.DEFAULT);
    mappings.add("mjs","text/javascript");
    factory.setMimeMappings(mappings);
}

Changes Made

* Added MIME type mapping for ".mjs" extension to "text/javascript" directly in DefaultMimeMappings.

Thanks

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/MimeMappings.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/MimeMappings.java
@@ -260,6 +260,7 @@ public sealed class MimeMappings implements Iterable<MimeMappings.Mapping> {
 			mappings.add("jpeg", "image/jpeg");
 			mappings.add("jpg", "image/jpeg");
 			mappings.add("js", "text/javascript");
+			mappings.add("mjs", "text/javascript");
 			mappings.add("json", "application/json");
 			mappings.add("otf", "font/otf");
 			mappings.add("pdf", "application/pdf");


philwebb:
Thanks very much for the suggestion and pull request. We've had several similar requests in the past for additional mime types but we intentionally keep our list aligned with [the defaults supported by Tomcat](https://github.com/apache/tomcat/blob/11fb662af9f8e26290be5c881b5a6837157894e7/java/org/apache/catalina/startup/MimeTypeMappings.properties). 

If you want to support additional types, you can do something like this in your application:

```java
@Configuration
public class MyMimeMapping implements WebServerFactoryCustomizer<ConfigurableServletWebServerFactory> {
    @Override
    public void customize(ConfigurableServletWebServerFactory factory) {
        MimeMappings mappings = new MimeMappings(MimeMappings.DEFAULT);
        mappings.add(""mjs", "text/javascript");
        factory.setMimeMappings(mappings);
    }
}```

Issue #38883, opened 2023-12-20
Document virtual threads limitations


PRIESt512:
Good afternoon.
I'd like to discuss the potential issue of enabling virtual threads at the platform level. I would update spring's documentation to use virtual threads. I would highlight this feature in red and warn other developers about possible problems. Virtual threads can be a very big problem if used incorrectly.
`spring.threads.virtual.enabled: true` - creates a strong temptation to install immediately and everywhere.
This property makes us think that the entire spring will work correctly with virtual threads. But this is definitely not the case. 

Let's say I want to use `spring-boot-starter-mail`... I start sending a message in a virtual thread. 
So what do I see? If I check the application with Flight Recorder I see a lot of `jdk.VirtualThreadPinned` events. If you look at the source code, this is an anti-pattern for virtual threads. There is `synchronized` inside and data is sent via the socket inside.
I sent messages 3 times and received 59 VirtualThreadPinned. Subject to triggering 20 ms, pinned time was from 100 to 900 ms...

Due to such limitations of virtual threads, you need to be extremely careful when using. Otherwise it can be fatal for the application.

Quote from Quarkus:

> Monopolization cases
> When running long computations, we do not allow the JVM to unmount and switch to another virtual thread until the virtual thread terminates. Indeed, the current scheduler does not support preempting tasks.
> 
> This monopolization can lead to the creation of new carrier threads to execute other virtual threads. Creating carrier threads results in creating platform threads. So, there is a memory cost associated with this creation.
> 
> Suppose you run in a constrained environment, such as containers. In that case, monopolization can quickly become a concern, as the high memory usage can lead to out-of-memory issues and container termination. The memory usage may be higher than with regular worker threads because of the inherent cost of the scheduling and virtual threads.

> Pinning cases
> The promise of "cheap blocking" might not always hold: a virtual thread might pin its carrier on certain occasions. The platform thread is blocked in this situation, precisely as it would have been in a typical blocking scenario.
> 
> According to [JEP 425](https://openjdk.org/jeps/425) this can happen in two situations:
> 
> when a virtual thread performs a blocking operation inside a synchronized block or method
> 
> when it executes a blocking operation inside a native method or a foreign function

[See more details here](https://github.com/quarkusio/quarkus/blob/main/docs/src/main/asciidoc/virtual-threads.adoc#monopolization-cases)

[...]

PR #38915, opened 2023-12-22
This pull requests closes issue number spring-projects#38883

bondress:


scottfrederick:
As [Phil stated](https://github.com/spring-projects/spring-boot/issues/38883#issuecomment-1865369181) in #38883, we don't want to go as far in was originally suggested in documenting virtual threads limitations, but would consider adding a link to Oracle documentation.

Issue #38902, opened 2023-12-21
Make it easier to add custom MimeMappings


philwebb:
#38900 reminded me that it's not easy to add additional mime types to the web server. It would be nice if this was possible using properties, or at the very least documented.

bondress:
I've been doing some research on this and I've found some interesting stuff. Can this be assigned to me, please?

scottfrederick:
Closing in favor of #39430

PR #38917, opened 2023-12-22
Closes issue #38902

bondress:


scottfrederick:
This appears to be a proposal to close #38883, instead of or in addition to #38902. As [Phil stated](#38883) in #38883, we don't want to go as far in was originally suggested in documenting virtual threads limitations, but would consider adding a link to Oracle documentation. #38902 suggests making a code enhancement, not just documentation. As this PR does not resolve either of these issues, I'm going to close it.

PR #38939, opened 2023-12-28
Test Gradle plugin against Gradle 8.6-rc-1

snicoll:


wilkinsona:
Thanks for the PR, @snicoll. We'll start testing against 8.6 in every build once it has been released but don't want to slow down every CI build in the meantime. Till then, the one off build from this PR is sufficient.

PR #39049, opened 2024-01-07
Upgrade to Brave 6 and Zipkin Reporter 3

codefromthecrypt:
This updates to Brave 6 and Zipkin Reporter 3 which are compatible with the classpath of Otel 1.35

Here are some new features for end-users, applicable to both Otel and Brave senders:
* Use protobuf by setting the property "management.zipkin.tracing.encoding" to "PROTO3"
* Integrate a dynamic zipkin endpoint source by providing a bean `HttpEndpointSupplier.Factory`
  * This will allow spring-cloud-loadbalancer or discovery integration like sleuth had, but also for otel
  * The input to this function is `ZipkinConnectionDetails.getSpanEndpoint()`
  * When absent `ZipkinConnectionDetails.getSpanEndpoint()` is called once like before.


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BravePropagationConfigurations.java
```diff
@@ -101,8 +101,18 @@ BaggagePropagation.FactoryBuilder propagationFactoryBuilder(
 		private Factory createThrowAwayFactory() {
 			return new Factory() {
 
+				/**
+				 * @deprecated end users and instrumentation should never call this, and instead
+				 * use {@link #get()}. This only remains to avoid rev-lock upgrading to Brave 6.
+				 */
+				@Deprecated
```
wilkinsona, line 108: There's no point in deprecating this as it's package-private API for which the javadoc is not published.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactory.java
```diff
@@ -70,10 +69,19 @@ public boolean requires128BitTraceId() {
 		return this.injectors.requires128BitTraceId() || this.extractors.requires128BitTraceId();
 	}
 
+	/**
+	 * @deprecated end users and instrumentation should never call this, and instead
+	 * use {@link #get()}. This only remains to avoid rev-lock upgrading to Brave 6.
+	 */
+	@Deprecated
```
wilkinsona, line 76: There's no point in deprecating this as it's package-private API for which the javadoc is not published.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/CompositePropagationFactoryTests.java
```diff
@@ -142,10 +141,19 @@ private DummyPropagation(String field) {
 			this.field = field;
 		}
 
+		/**
+		 * @deprecated end users and instrumentation should never call this, and instead
+		 * use {@link #get()}. This only remains to avoid rev-lock upgrading to Brave 6.
+		 */
+		@Deprecated
```
wilkinsona, line 148: There's no point in deprecating this as it's test code

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BravePropagationConfigurations.java
```diff
@@ -101,8 +101,18 @@ BaggagePropagation.FactoryBuilder propagationFactoryBuilder(
...
 				@Override
-				public <K> Propagation<K> create(KeyFactory<K> keyFactory) {
+				public <K> Propagation<K> create(Propagation.KeyFactory<K> ignored) {
+					throw new UnsupportedOperationException("As of Brave 5.12, call PropagationFactory.get()");
+				}
+
+				@Override
+				public Propagation<String> get() {
```
wilkinsona, line 115: My reading of https://github.com/openzipkin/brave/pull/1396 is that Brave 5.18 will require `get` to be implemented. Have I missed something? This change appears to be unrelated to upgrading to Brave 5.17.1.

I'm also not sure about throwing an `UnsupportedOperationException`. Is it not pushing a guaranteed failure to runtime when remove the method would have caught it at compile time instead? If the goal is to provide a smoother migration path, I wonder if `create` could delegate somehow to `get` instead.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/LocalBaggageFieldsTests.java
```diff
@@ -53,8 +53,18 @@ void empty() {
 	@SuppressWarnings("deprecation")
 	private static FactoryBuilder createBuilder() {
 		return BaggagePropagation.newFactoryBuilder(new Factory() {
+			/**
+			 * @deprecated end users and instrumentation should never call this, and instead
+			 * use {@link #get()}. This only remains to avoid rev-lock upgrading to Brave 6.
+			 */
+			@Deprecated
```
wilkinsona, line 60: There's no point in deprecating this as it's test code

wilkinsona (commented):
Thanks for the PR, @codefromthecrypt.

codefromthecrypt:
@wilkinsona I think we can get away from all these deprecated things, sorry about the distraction. I'll mark draft and triple check!

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BravePropagationConfigurations.java
codefromthecrypt: The issue is that `create` was formerly abstract and hasn't been called directly in a very long time.

In any case, I removed the special casing. I updated this to Brave 6, but it will also work without change with 5.18.

codefromthecrypt:
should be an easier review, now. I had to add the zipkin-reporter bom not because of code in this change, but because Brave 6 no longer manages zipkin-reporter (Brave 6 has no dependency on zipkin types). spring-boot-actuator-autoconfigure defines reporter types, so needs this for those to compile. When that code updates to Zipkin Reporter 3 is up to you, I'm fine helping soon, but if not this week I can't promise availability as this stuff isn't related to my dayjob anymore.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BravePropagationConfigurations.java
```diff
@@ -97,12 +96,10 @@ BaggagePropagation.FactoryBuilder propagationFactoryBuilder(
 			return builder;
 		}
 
-		@SuppressWarnings("deprecation")
 		private Factory createThrowAwayFactory() {
 			return new Factory() {
-
 				@Override
-				public <K> Propagation<K> create(KeyFactory<K> keyFactory) {
```
codefromthecrypt, line 105: the key is that create hasn't been called by Brave since 5.12 I think. It has been calling `get()` which defaulted to forward to create. `create()` is deprecated and will be removed in Brave 7, so not implementing that here allows less maintenance in the future.

codefromthecrypt:
actually update to zipkin-reporter v3 will need to wait until this is released, as it seems boot is also importing otel exporter for `OpenTelemetryConfiguration` (note brave doesn't actually use this type). Anyway, migration to v3 is changing packages, so anyone can do it similar to this https://github.com/openzipkin/zipkin-aws/pull/208

wilkinsona:
We can't upgrade to Brave 6.0 [until Micrometer has done](https://github.com/micrometer-metrics/tracing/pull/511) so.

wilkinsona (approved):

efemoney:
Soft bump 🙏🏾  hope everyone is having a lovely day!

codefromthecrypt:
@efemoney I think the current status is micrometer are awaiting https://github.com/open-telemetry/opentelemetry-java to release 1.35.0. Then it can be updated per notes [here](https://github.com/micrometer-metrics/tracing/pull/536) and released, then this one can progress after updating to new micrometer.

TL;DR; waiting on otel to release and I don't know the schedule of that

efemoney:
Thank you. I checked the micrometer PR but somehow missed that 🙌🏾

codefromthecrypt:
turns out https://github.com/micrometer-metrics/tracing/pull/536 only affected test deps as there was no main code change. So, there's no inter-dependency afaict except maybe if otel latest is needed here. Anyway I will sort this out next!

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/BravePropagationConfigurations.java
codefromthecrypt: since otel is already up with Brave 6 and Zipkin Reporter 3, let's not worry too much about this part. Feel free to unresolve if you disagree!

codefromthecrypt:
ok there's probably a lint or some test glitch, but I'm time out for the day.

Effectively, we don't want to depend on zipkin core (zipkin.Reporter) types anymore, so detecting Brave already configured if there is an `AsyncZipkinSpanHandler` rather wrapping and bringing back a zipkin dep with `AsyncReporter`

Also, almost everyone uses json and there's no built-in set of encoders in brave. So, I used optional injection of an encoder for both use cases brave and otel. The only known custom user of that encoder would be zipkin-gcp, and again lacking any built-in choice except json, optional is the best way.

Feedback welcome, but I think you can see where I'm going with this and it is a lot simpler to implement.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinAutoConfiguration.java
```diff
@@ -56,10 +51,4 @@ PropertiesZipkinConnectionDetails zipkinConnectionDetails(ZipkinProperties prope
 		return new PropertiesZipkinConnectionDetails(properties);
 	}
 
-	@Bean
-	@ConditionalOnMissingBean
-	public BytesEncoder<Span> spanBytesEncoder() {
```
codefromthecrypt, line 61: removing this from top-level because brave uses `BytesEncoder<MutableSpan>`, not `BytesEncoder<Span>` which pulls in zipkin core types. We can push this into the otel config if desired as it still requires the core zipkin jar, but let's be sure to not accidentally add that dep for normal brave.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -126,29 +129,16 @@ ZipkinWebClientSender webClientSender(ZipkinProperties properties,
...
-		@ConditionalOnMissingBean(Reporter.class)
-		@ConditionalOnBean(Sender.class)
-		AsyncReporter<Span> spanReporter(Sender sender, BytesEncoder<Span> encoder) {
-			return AsyncReporter.builder(sender).build(encoder);
-		}
-
-	}
-
-	@Configuration(proxyBeanMethods = false)
-	@ConditionalOnClass(ZipkinSpanHandler.class)
```
codefromthecrypt, line 141: ZipkinSpanHandler is a wrapper for those using core types via `Reporter<Span>`. It is more efficient and less deps to not force users through that, so removed that for `AsyncZipkinSpanHandler`

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -157,10 +147,12 @@ static class OpenTelemetryConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
-		@ConditionalOnBean(Sender.class)
+		@ConditionalOnBean(BytesMessageSender.class)
 		@ConditionalOnEnabledTracing
-		ZipkinSpanExporter zipkinSpanExporter(BytesEncoder<Span> encoder, Sender sender) {
-			return ZipkinSpanExporter.builder().setEncoder(encoder).setSender(sender).build();
+		ZipkinSpanExporter zipkinSpanExporter(BytesMessageSender sender, Optional<BytesEncoder<Span>> encoder) {
```
codefromthecrypt, line 152: we *could* swap out this `Optional<BytesEncoder<Span>> encoder` for a bean contained inside OpenTelemetryConfig, like what used to be top-level. When you have a dep on otel's zipkin exporter, you already have the classes needed for below. However, brave will not bring in these classes, and it wouldn't be great to force normal users to do that when they don't need them.

```java
@Bean
	@ConditionalOnMissingBean
	public BytesEncoder<Span> spanBytesEncoder() {
		return SpanBytesEncoder.JSON_V2;
	}
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/HttpSender.java
```diff
@@ -18,37 +18,37 @@
...
+	 */
+	private static final DataSize COMPRESSION_THRESHOLD = DataSize.ofKilobytes(1);
+
 	private volatile boolean closed;
 
-	@Override
-	public Encoding encoding() {
-		return Encoding.JSON;
+	HttpSender() {
+		super(Encoding.JSON);
```
codefromthecrypt, line 51: note: the built-in senders can only work with json, as both here and also the media type header assume that. This is fine for almost everyone.

codefromthecrypt:
encoder swapping works, but marking draft as finally have tests running locally. I need to look into these:

```
ZipkinAutoConfigurationIntegrationTests > zipkinsUseOfRestTemplateDoesNotCauseACycle() FAILED
    java.lang.AssertionError at ZipkinAutoConfigurationIntegrationTests.java:52

ZipkinAutoConfigurationIntegrationTests > zipkinsUseOfWebClientDoesNotCauseACycle() FAILED
    java.lang.AssertionError at ZipkinAutoConfigurationIntegrationTests.java:59

ZipkinAutoConfigurationTests > definesPropertiesBasedConnectionDetailsByDefault() FAILED
    java.lang.AssertionError at ZipkinAutoConfigurationTests.java:39

ZipkinAutoConfigurationTests > shouldUseCustomConnectionDetailsWhenDefined() FAILED
    java.lang.AssertionError at ZipkinAutoConfigurationTests.java:52
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -157,10 +155,42 @@ static class OpenTelemetryConfiguration {
 
 		@Bean
 		@ConditionalOnMissingBean
-		@ConditionalOnBean(Sender.class)
 		@ConditionalOnEnabledTracing
-		ZipkinSpanExporter zipkinSpanExporter(BytesEncoder<Span> encoder, Sender sender) {
-			return ZipkinSpanExporter.builder().setEncoder(encoder).setSender(sender).build();
+		BytesEncoder<Span> spanBytesEncoder() {
+			return SpanBytesEncoder.JSON_V2;
```
codefromthecrypt, line 160: allows use of stackdriver, like this https://github.com/openzipkin/zipkin-gcp/blob/10827206f1e4f6b65e261d936ccd1c45e8a42ef9/encoder-stackdriver-brave/src/main/java/zipkin2/reporter/stackdriver/brave/StackdriverV2Encoder.java#L25

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinAutoConfiguration.java
codefromthecrypt, line 61: `BytesEncoder<Span>` is only used by opentelemetry so moved it there

jonatan-ivanov:
Fyi: Micrometer Tracing 1.3.0-M1 has the Brave 6 (+ OTel 1.35) upgrade.

codefromthecrypt:
@jonatan-ivanov cool I will mark this one final today, just sorting out one glitch this morning.

codefromthecrypt:
almost done.. sorry I meant to finish yesterday, but 100pct will today.

codefromthecrypt:
ok I've updated the description and also want to thank @reta and @anuraaga who spent the last several days with me making sure the baseline library is high quality.

Particularly, we have a replacement for how to integrate eureka at a lower level (the sender) so that brave and otel can use it. Basically, eureka integration is pinned to sleuth (boot 2.x) until this is out.

I'm doing some final integration tests of micrometer tracing around this, and following that will remove the snapshot version of reporter for a real one.

codefromthecrypt:
if someone can click approve on the workflow, as far as I know this is done except getting rid of the snapshot dep. I'll do more integration testing meanwhile, but curious if the official CI is happy.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -59,15 +65,20 @@ static class SenderConfiguration {
 	static class UrlConnectionSenderConfiguration {
 
 		@Bean
-		@ConditionalOnMissingBean(Sender.class)
-		URLConnectionSender urlConnectionSender(ZipkinProperties properties,
-				ObjectProvider<ZipkinConnectionDetails> connectionDetailsProvider) {
+		@ConditionalOnMissingBean(BytesMessageSender.class)
+		URLConnectionSender urlConnectionSender(ZipkinProperties properties, Encoding encoding,
+				ObjectProvider<ZipkinConnectionDetails> connectionDetailsProvider,
+				ObjectProvider<HttpEndpointSupplier.Factory> endpointSupplierFactoryProvider) {
```
codefromthecrypt, line 71: this bean doesn't seem to be found in my integration tests, I'll have to dig deeper into what I did wrong here..

codefromthecrypt: nevermind it works.. I'll post back!

codefromthecrypt:
ok I moved to release version and here's the bit that works in a separate project and re-routes endpoint to loadbalancer like sleuth once did. cc @spencergibb @OlgaMaciaszek @marcingrzejszczak @jonatan-ivanov @shakuzen 

I'm sure you'd do it differently, but anyway here it is!
```java
package brave.example;

import java.net.URI;
import org.springframework.cloud.client.ServiceInstance;
import org.springframework.cloud.client.loadbalancer.LoadBalancerClient;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import zipkin2.reporter.HttpEndpointSupplier;
import zipkin2.reporter.HttpEndpointSuppliers;

public class ZipkinDiscoveryConfiguration {
  @Bean HttpEndpointSupplier.Factory loadbalancerEndpoints(LoadBalancerClient loadBalancerClient) {
    LoadBalancerHttpEndpointSupplier.Factory httpEndpointSupplierFactory =
        new LoadBalancerHttpEndpointSupplier.Factory(loadBalancerClient);
    // don't ask more than 30 seconds (just to show)
    return HttpEndpointSuppliers.newRateLimitedFactory(httpEndpointSupplierFactory, 30);
  }

  record LoadBalancerHttpEndpointSupplier(LoadBalancerClient loadBalancerClient, URI virtualURL)
      implements HttpEndpointSupplier {
    record Factory(LoadBalancerClient loadBalancerClient) implements HttpEndpointSupplier.Factory {

      @Override public HttpEndpointSupplier create(String endpoint) {
        return new LoadBalancerHttpEndpointSupplier(loadBalancerClient, URI.create(endpoint));
      }
    }

    @Override public String get() {
      // At least spring-cloud-netflix wants the actual hostname as a lookup value
      ServiceInstance instance = loadBalancerClient.choose(virtualURL.getHost());
      if (instance != null) {
        return instance.getUri() + virtualURL.getPath();
      }
      throw new IllegalArgumentException(
          virtualURL.getHost() + " is not an instance registered in Eureka");
    }

    @Override public void close() {
    }

    @Override public String toString() {
      return "LoadBalancer{" + virtualURL + "}";
    }
  }
}
```

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinWebClientSender.java
```diff
@@ -34,68 +33,27 @@
...
-
+	void postSpans(URI endpoint, HttpHeaders headers, byte[] body) {
+		this.webClient.post()
+			.uri(endpoint)
+			.headers((h) -> h.addAll(headers))
+			.bodyValue(body)
+			.retrieve()
+			.toBodilessEntity()
+			.timeout(this.timeout)
+			.block();
```
mhalbritter, line 56: I guess the contract of `BytesMessageSender` is that `send` blocks, right?

This `.block()` on the webclient will block one event loop thread. Maybe it would be better to schedule that on a threadpool?

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/DefaultEncodingConfiguration.java
```diff
@@ -0,0 +1,37 @@
...
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+
+/**
+ * Configures the bean {@linkplain ZipkinAutoConfiguration} would from properties.
+ */
+@Configuration(proxyBeanMethods = false)
+class DefaultEncodingConfiguration {
```
mhalbritter, line 29: This can be a `@TestConfiguration`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurationsBraveConfigurationTests.java
```diff
@@ -40,61 +46,126 @@
...
 
 	@Test
 	void shouldNotSupplySpanHandlerIfReporterIsMissing() {
-		this.contextRunner.run((context) -> assertThat(context).doesNotHaveBean(ZipkinSpanHandler.class));
+		this.contextRunner.run((context) -> assertThat(context).doesNotHaveBean(AsyncZipkinSpanHandler.class));
 	}
 
 	@Test
 	void shouldNotSupplyIfZipkinReporterBraveIsNotOnClasspath() {
+		// Note: Technically, Brave can work without zipkin-reporter. For example,
```
mhalbritter, line 67: Looking at `org.springframework.boot.actuate.autoconfigure.tracing.zipkin.ZipkinConfigurations.BraveConfiguration`, it provides `BytesEncoder<MutableSpan>` and `AsyncZipkinSpanHandler`, both are in the `zipkin-reporter` dependency. 

You mean removing the zipkin dependency from Brave could resolve https://github.com/micrometer-metrics/tracing/issues/509?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -111,42 +126,43 @@ private RestTemplateBuilder applyCustomizers(RestTemplateBuilder restTemplateBui
...
+	@ConditionalOnClass(AsyncZipkinSpanHandler.class)
+	static class BraveConfiguration {
 
 		@Bean
-		@ConditionalOnMissingBean(Reporter.class)
-		@ConditionalOnBean(Sender.class)
-		AsyncReporter<Span> spanReporter(Sender sender, BytesEncoder<Span> encoder) {
-			return AsyncReporter.builder(sender).build(encoder);
+		@ConditionalOnMissingBean(value = MutableSpan.class, parameterizedContainer = BytesEncoder.class)
+		@ConditionalOnEnabledTracing
```
mhalbritter, line 152: I'd remove that annotation from the bean method. The idea is only to disable the data-reporting components and leaving everything else untouched.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurations.java
```diff
@@ -155,12 +171,19 @@ ZipkinSpanHandler zipkinSpanHandler(Reporter<Span> spanReporter) {
 	@ConditionalOnClass(ZipkinSpanExporter.class)
 	static class OpenTelemetryConfiguration {
 
+		@Bean
+		@ConditionalOnMissingBean(value = Span.class, parameterizedContainer = BytesEncoder.class)
+		@ConditionalOnEnabledTracing
```
mhalbritter, line 176: I'd remove that annotation from the bean method. The idea is only to disable the data-reporting components and leaving everything else untouched.

mhalbritter:
Thanks a lot for the PR, there are some cool features in there! :)

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinWebClientSender.java
mhalbritter: We have this because the `RestTemplate` is in `spring-web`, which is not on the classpath when using webflux. That's why we added the webflux one.

Maybe we could implement one based on the JDK httpclient, then we can get rid of the webflux one which is misusing the event loop with the `.block` here.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurationsBraveConfigurationTests.java
codefromthecrypt: So, if things are same as when I wrote it, wavefront doesn't use any of the zipkin-reporter types, just brave's `brave.handler.SpanHandler`. Means its only dependency is brave's core jar.

Practically speaking, all others will need zipkin-reporter-brave, which depends on zipkin-reporter, but not zipkin's core jar (anymore) as BytesEncoder<MutableSpan> and AsyncZipkinSpanHandler do not need those types.

```xml
    <dependency>
      <groupId>${project.groupId}</groupId>
      <artifactId>zipkin-reporter</artifactId>
      <version>${project.version}</version>
      <!-- Senders don't use zipkin types. Excluding allows brave users to
           avoid them by default. -->
      <exclusions>
        <exclusion>
          <groupId>io.zipkin.zipkin2</groupId>
          <artifactId>zipkin</artifactId>
        </exclusion>
      </exclusions>
    </dependency>
```

That said, the otel exporter *does* use io.zipkin.zipkin2:zipkin, for AsyncReporter<zipkin2.Span> and correspondingly `BytesEncoder<MutableSpan>`

hope these notes help. I'm knackered from several very long days so will get back to more feedback tomorrow asia time!

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinAutoConfiguration.java
mhalbritter: I wonder if `BytesEncoder<Span>` will be useful outside of OpenTelemetry. It has been moved to `ZipkinConfigurations.OpenTelemetryConfiguration`, and I wonder if we should leave it in `ZipkinAutoConfiguration`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinWebClientSender.java
mhalbritter: I've opened #39545 for that.

mhalbritter: We still have to decide what to do with the webclient one. Calling `.block` is bad, removing the webclient based sender will break users who currently only have webflux on the classpath. Maybe using an executor where we run the webclient on?

What's your take on this, @wilkinsona ?

wilkinsona: We already call `block()` in some circumstances:

https://github.com/spring-projects/spring-boot/blob/2abc25d51cbb672ef25bdb8141bb5fef25088404/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinWebClientSender.java#L76

That means that it's not clear to me how much worse this change is making things. That said, dropping the `WebClient`-based sender in favour of one that uses `HttpClient` makes sense. I think we could accept this change as proposed and then #39545 would make the use of `block()` irrelevant anyway.

mhalbritter: Thanks. Sounds good.

codefromthecrypt: I agree lets leave it alone. the part of the code in question just moved from one place to another. The "which client" thing is the same concern as before and can be done in a follow-up (with folks who know better about pros/cons and ways out than me!)

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinConfigurationsBraveConfigurationTests.java
codefromthecrypt: this discussion is something that can be done in another PR. I'm glad the comment drew attention to it, but it isn't super critical as zipkin-reporter-brave jar is only a few 10kb

codefromthecrypt:
rebased with feedback from @mhalbritter accepted! 🤞 on green

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/zipkin/ZipkinAutoConfiguration.java
mhalbritter: Right now, this could break users. I don't really know how many users depend on `BytesEncoder<Span>` be available, but I would leave it in `ZipkinAutoConfiguration` instead of moving it to the OpenTelemetry configuration.

codefromthecrypt:
red looks unrelated unless I'm missing something

codefromthecrypt:
rebased. let me know what else is needed here!

codefromthecrypt:
same snapshot download error as before..

mhalbritter:
Don't worry. I'll take a look at the PR and merge it. It builds fine on my machine.

mhalbritter:
Thanks a lot for the PR!

codefromthecrypt:
thanks @mhalbritter .. one less thing on my nag list, but more importantly a big step forward for those using this stuff. cheers!

PR #39052, opened 2024-01-07
Fix Jetty ConnectionLimit configuration

### LINKED ISSUE #39050: Jetty ConnectionLimit feature is not working
Affects: 3.2.1

Jetty ConnectionLimit feature is introduced via 

`server.jetty.max-connections` in https://github.com/spring-projects/spring-boot/issues/35899

However, this ConnectionLimit is not working properly due to problem that introduced in this issue https://github.com/spring-projects/spring-boot/issues/30565

If Spring Boot decided to give first-class support for this, I think we need to find a proper solution.

As a shortcut from #30565, instead of supported prop, this works like a charm.   
```
@Bean
public JettyServerCustomizer connectorsConnectionLimit() {
    return server -> server.addBean(new ConnectionLimit(1, server.getConnectors()));
}
```

Thanks

okohub:
Fixes #39050


mhalbritter:
Thank you!

Issue #35926, opened 2023-06-16
Improve Elasticsearch ServiceConnection


eddumelendez:
Currently, [ElasticsearchContainerConnectionDetailsFactory](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-testcontainers/src/main/java/org/springframework/boot/testcontainers/service/connection/elasticsearch/ElasticsearchContainerConnectionDetailsFactory.java) only maps `nodes` and that works with Elasticsearch 7. However, Elasticsearch 8 has security enabled by default and additional configuration is required.

[This](https://github.com/eddumelendez/testcontainers-samples/blob/main/spring-boot-elasticsearch/src/test/java/com/example/springbootelasticsearch/BookRepositoryElasticsearch7Tests.java) is an example using Elasticsearch 7 and [this](https://github.com/eddumelendez/testcontainers-samples/blob/main/spring-boot-elasticsearch/src/test/java/com/example/springbootelasticsearch/BookRepositoryElasticsearch8Tests.java) one using Elasticsearch 8. As you can see, with Elasticsearch 8 the `SSLContext` is provided via a `customizer` so the connection through https is successful.

You can find a draft proposal [here](https://github.com/eddumelendez/spring-boot/commit/ba47358691a1e9a033461f061243a8042e039dc8). I tried to do it via ssl bundles by copying the `http_ca.crt` from the container to the test resources directory but I couldn't set it up correctly.

Divyansh1711:
I m beginner and i have no idea ...how to start contributing for project.. Anyone any guides for me??

wilkinsona:
Thanks for your interest in contributing. I'm not sure if we have any at the moment, but please keep an eye out for [unassigned issues labelled with ideal for contribution](https://github.com/spring-projects/spring-boot/issues?q=is%3Aopen+label%3A%22status%3A+ideal-for-contribution%22+no%3Aassignee) or, as your haven't contributed before, [first-timers only](https://github.com/spring-projects/spring-boot/issues?q=is%3Aopen+label%3A%22status%3A+first-timers-only%22+no%3Aassignee).

kevinwheeler:
I'll see if this ends up being over my head or not, but I'm starting to investigate this / work on it. Feedback is always welcome.

kevinwheeler:
Something came up that I'm going to have to work on for a few days before I can get back to this unfortunately.

kevinwheeler:
[...]

PR #39092, opened 2024-01-11
Added detecting protocol, username and password in ElasticsearchConta…

pioorg:
Improve Elasticsearch ServiceConnection #35926 

Currently only Elasticsearch 7.x works with `@ServiceConnection` without any additional (manual) configuration or disabling security and SSL.

The proposed changes make it possible to use also:
* version 8.x (where security and SSL are enabled by default)
* version 7.x with with security and SSL enabled.

It is achieved by:
* detecting if SSL is in place (therefore needing HTTPS)
* security is enabled (thus requiring username and password)


wilkinsona:
Thanks, @pioorg. Unfortunately, as indicated by it being assigned to someone, #35926 is already being worked on. There's also an existing prototype linked to in the issue's comments.

pioorg:
Hello @wilkinsona
I saw both prototypes mentioned in the issue and the issue is that (in my understanding), they won't work, because they assume security and HTTPS always come together. I also made a comment there.
I'm not interested in this PR being merged. What I'm after is having a working and correct solution.
Thank you!

PR #39107, opened 2024-01-11
Remove unreachable throw code

### LINKED ISSUE #39086: Polish SpringApplication to avoid dead code after calls to handleRunFailure
As reported in #38951, there's some dead code after calls to `handleRunFailure`. We think it might be better to have `handleRunFailure` return the exception to be thrown and then have the caller throw it.

BenchmarkingBuffalo:
Remove the dead code after calling 'handleRunFailure'. 

No adjustment of tests needed.
Fix #39086

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
@@ -95,7 +95,6 @@ import org.springframework.util.Assert;
 import org.springframework.util.ClassUtils;
 import org.springframework.util.CollectionUtils;
 import org.springframework.util.ObjectUtils;
-import org.springframework.util.ReflectionUtils;
 import org.springframework.util.StringUtils;
 import org.springframework.util.function.ThrowingConsumer;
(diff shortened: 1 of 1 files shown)


wilkinsona (changes requested):
Thanks for the PR. Unfortunately, this isn't the change that we want to make. As described in the issue, we want to change `handleRunFailure` to return the exception and to then have it thrown by the caller. This will eliminate the dead code while continuing to make it clear that the method won't continue to execute.

BenchmarkingBuffalo:
Hi @wilkinsona,
thanks for clarifying.
I will prepare something.

BenchmarkingBuffalo:
Hi @wilkinsona,
I changed the behavior to the one described in the issue.
I also did a small change to the structure of the exception-handling. 
This was originally due to a hint from sonar, but I actually like it. For me it is easier to understand looking at multiple catch-statements than a wrapped if-statement distinguishing between the exception types.
If I did not get it right or you would like me to change something in my solution, just hit me up.

philwebb:
Thanks very much @BenchmarkingBuffalo. I made one further tweak by pushing the `AbandonedRunException` logic into the `handleRunFailure` method.

PR #39124, opened 2024-01-15
Autoconfiguration for Kafka Cluster ID

### LINKED ISSUE #38825: Auto configuration for Kafka Cluster ID
Starting with version 3.1, the KafkaAdmin has this new API setClusterId which should be used to prevent attempting to fetch the cluster id from the broker, perhaps if the user does not have admin permissions. 

Can we please have application property as `spring.kafka.consumer.cluster-id` and `spring.kafka.producer.cluster-id ` to facilitate auto configuration of cluster id when user has ready only access or don't have admin access. 

It is important to note that cluster id is needed to enabled Micrometer Observability without which it is not working.

Following is the Method with Java Docs of `setClusterId` method of `KafkaAdmin` class present in Spring Kafka with` version 3.1`, the has this new API:

```
	/**
	 * Set the cluster id. Use this to prevent attempting to fetch the cluster id
	 * from the broker, perhaps if the user does not have admin permissions.
	 * @param clusterId the clusterId to set
	 * @since 3.1
	 */
	public void setClusterId(String clusterId) {
```

Wzy19930507:
Closes #38825


wilkinsona:
Unfortunately, this change isn't what was described in the issue. I also don't think we're ready to implement anything just yet as we have some design to figure out first. Thanks anyway for the PR.

PR #39125, opened 2024-01-15
Use the term "tags" in documentation consistently

### LINKED ISSUE #38761: Use the term "tags" in documentation consistently
In our documentation, sometimes we use "key-values" and sometimes we use "tags". The key value introduction in Micrometer had technical reasons, we should stick in our documentation with one term, and IMHO, that should be "tags".

Wzy19930507:
Closes #38761


philwebb:
Thanks very much @Wzy19930507

PR #39137, opened 2024-01-15
SslBundle implementations do not provide useful toString() results

### LINKED ISSUE #39057: SslBundle does miss an approriate toString() method
Hi,

we are using different bundles and the one who does configure them and the one who does operate the app in question are not always the same person.
We're logging the used bundle, however we only get this as toString():

`[org.springframework.boot.autoconfigure.ssl.PropertiesSslBundle@166e52ca]`

Tomcat e.g. does print something like this for the connector config:

` Connector [https-openssl-nio-8443], TLS virtual host [_default_], certificate type [UNDEFINED] configured from keystore [/home/app/.keystore] using alias [tomcat] with trust store [null]`

It would be nice if a better toString could be done which prints all important configured props of the bundle (without sensitive stuff like passwords etc.), but ciphers, options, aliases, key / trust store path etc. would be nice to have on the configured bundle.

amparab:
Fixes #39057

Change proposed :

I have added a toString() method in the implementations for SslBundle interface.
There were two implementations :
1. PropertiesSslBundle
2. WebServerSslBundle

This toString() method prints the following attributes of the respective class:
1. alias of key
2. protocol
3. ciphers
4. enabledProtocols

Attaching a screenshot of the line printed when object is logged :

image

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/PropertiesSslBundle.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/PropertiesSslBundle.java
@@ -146,8 +146,8 @@ public final class PropertiesSslBundle implements SslBundle {
 
 	@Override
 	public String toString() {
-		return "PropertiesSslBundle{" + "key=" + this.key + ", protocol='" + this.protocol + '\'' + ", ciphers="
-				+ Arrays.toString(this.options.getCiphers()) + ", enabled-protocols="
+		return "PropertiesSslBundle{" + "key-alias=" + this.key.getAlias() + ", protocol='" + this.protocol + '\''
+				+ ", ciphers=" + Arrays.toString(this.options.getCiphers()) + ", enabled-protocols="
 				+ Arrays.toString(this.options.getEnabledProtocols()) + '}';
 	}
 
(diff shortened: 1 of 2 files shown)


tkrah:
What about keystore type, keystore location, truststore type and location, those would also be beneficial to know, no?

amparab:
> What about keystore type, keystore location, truststore type and location, those would also be beneficial to know, no?

I have added keystore type and truststore type (if truststore is not null, otherwise it would throw NullPointerException)
Printing the keystore and truststore location is little tricky, because the keystore and truststore has already been loaded at this point and it's path is not available.

wilkinsona (commented):
Thanks very much for the PR, @amparab.

There's a utility in Spring Framework, `ToStringCreator`, that we like to use for this sort of thing. Here's one existing example in Spring Boot's code of how it's used:

https://github.com/spring-projects/spring-boot/blob/4fa47aa5015b0d7374d47005e01ab5163af7fd03/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/Bindable.java#L162-L170

Would you like to update your proposal to use it? Please don't worry if you don't have time, we can always take care of it as part of merging the change.

tkrah:
> > What about keystore type, keystore location, truststore type and location, those would also be beneficial to know, no?
> 
> I have added keystore type and truststore type (if truststore is not null, otherwise it would throw NullPointerException) Printing the keystore and truststore location is little tricky, because the keystore and truststore has already been loaded at this point and it's path is not available.

Could you transport that location info (add it to the code) just for that purpose maybe?

tkrah:
> > Could you transport that location info (add it to the code) just for that purpose maybe?
> 
> That's a much broader and more complex change, particularly as some values can either be a location from which something is loaded or something inlined directly.
> 
> As a first pass, for this PR let's focus on the things that are readily available. We can then perhaps consider a second pass that does something more complex, possibly through implementation of the `Origin` interface.

Sounds good :)

amparab:
> Thanks very much for the PR, @amparab.
> 
> There's a utility in Spring Framework, `ToStringCreator`, that we like to use for this sort of thing. Here's one existing example in Spring Boot's code of how it's used:
> 
> https://github.com/spring-projects/spring-boot/blob/4fa47aa5015b0d7374d47005e01ab5163af7fd03/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/Bindable.java#L162-L170
> 
> Would you like to update your proposal to use it? Please don't worry if you don't have time, we can always take care of it as part of merging the change.

Hello @wilkinsona. I have updated the PR :-) Please let me know if any other changes.

wilkinsona:
@amparab thank you very much for making your first contribution to Spring Boot

PR #39145, opened 2024-01-16
Remove entry for OCI starter as it is no longer maintained

OlgaMaciaszek:
The project has been archived and is no longer maintained.


philwebb:
Thanks @OlgaMaciaszek !

wilkinsona:
@philwebb shouldn't this have gone into 3.1.x and later?

philwebb:
I want sure with README content. I figured most folks would only look at it on main.

Issue #37426, opened 2023-09-15
Refactor PemPrivateKeyParser to use DerElement


mhalbritter:
The `PemPrivateKeyParser` uses a brute-force approach to find a suitable algorithm to parse a PEM-encoded key. This could be improved by using the `DerElement` class to read the ASN.1 and find out what type of key this is. This works for all PKCS#8 keys.

Wzy19930507:
@mhalbritter Hello, I see the oids are maintained in KnownOIDs, can I refer to it and convert oid into an algorithm to solve this problem?

such as 
```
AlgName algName = spring.boot.KnownOIDs.getAlgName(int[] RSA_ALGORITHM);
new PKCS8EncodedKeySpec(bytes, algorithmName);
```

philwebb:
Closing in favor of PR #39162. Thanks @Wzy19930507

PR #39162, opened 2024-01-17
Refactor PemPrivateKeyParser to use DerElement

Wzy19930507:
If oid resolution algorithms failed, back to brute force.

Closes https://github.com/spring-projects/spring-boot/issues/37426


philwebb:
Thanks very much @Wzy19930507!

PR #39170, opened 2024-01-17
Update CRaC support status link

sdeleuze:
See https://github.com/spring-projects/spring-lifecycle-smoke-tests/issues/51 related issue.


wilkinsona:
Thanks very much, @sdeleuze.

PR #39227, opened 2024-01-17
Add liquibase ui-service property

EvaristeGalois11:
Hi, I added to the autoconfiguration the property `ui-service` introduced in Liquibase 4.25.0.


mhalbritter:
Thank you!

PR #39241, opened 2024-01-18
Upgrade Java 17 version in CI image and .sdkmanrc

Siddharth1410:
Fixes issues https://github.com/spring-projects/spring-boot/issues/39221, https://github.com/spring-projects/spring-boot/issues/39218, and https://github.com/spring-projects/spring-boot/issues/39216


philwebb:
Thanks for PR but we prefer to handle these kind of updates directly ourselves. We have commits lined up for this one but we were waiting for sdkman to catch up with the releases.

PR #39248, opened 2024-01-19
Fixes Issue #28841 Adding more info about scanned packages for NoSuchBeanDefinitionException

### LINKED ISSUE #28841: Include information about the scanned packages in failure analysis for NoSuchBeanDefinitionException
One possible cause of #28832 is that the packages containing the components haven't been scanned. If it's possible to do so, I think it would be useful to include information about the scanning packages in the failure analysis for a NoSuchBeanDefinitionException.

amparab:
Fixes #28841

Aim : Include information about the scanned packages in the failure analysis for a `NoSuchBeanDefinitionException`

In this change I am getting the list of scanned packages from `ConfigurableListableBeanFactory beanFactory` in the `NoSuchBeanDefinitionFailureAnalyzer`
I am excluding `org.springframework` packages because that would result in a huge list.

PFA the results.

image

I am not sure if this is the best approach. 
Would be happy to get your suggestions and insights.

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/bom/BomExtension.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/bom/BomExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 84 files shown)


wilkinsona:
Thanks for looking at this, @amparab. Unfortunately, there may be many beans in the bean factory that weren't found through component scanning so your proposal here may list too many packages.

What we wanted to try and do is to include information about the packages listed in `@ComponentScan` (used directly or through another annotation like `@SpringBootApplication`) in the failure analysis. When I opened #28841 I wasn't sure that this was possible. Having looked more closely just now, I'm pretty sure that it is not. `@ComponentScan` is processed within  `org.springframework.context.annotation.ConfigurationClassParser` in a way that is an implementation detail. To implement #28841, we'd need a change in Framework that somehow exposed some information about the every package that has been scanned due to a `@ComponentScan` annotation.

Thanks anyway for looking at this one. Unfortunately, I think it's best if we close this one (and perhaps #28841 as well). Sorry for the wasted effort.

PR #39253, opened 2024-01-20
Apply @AutoConfigureJson to test slices

### LINKED ISSUE #38869: Use meta-annotations on @AutoConfigureWebMvc to compose it with @AutoConfigureJson
Currently `@AutoConfigureWebMvc` imports JSON auto-configuration directly. If we use `@AutoConfigureJson` then any user customizations to `org.springframework.boot.test.autoconfigure.json.AutoConfigureJson.imports` would be automatically picked up. We should review if the current arrangement is intentional and if we should switch.

Wzy19930507:
…figureJson

* AutoConfigureJsonTesters
* AutoConfigureWebClient
* AutoConfigureWebMvc

Closes #38869


mhalbritter:
Thanks!

PR #39255, opened 2024-01-20
Fix typo

criztovyl:
Fix a typo in the docs that uses "war" instead of "WAR"


wilkinsona:
Thanks for the proposal but I don't consider this to be a typo. I prefer to see `war` styled as `war` rather than `WAR` (and also `jar` rather than `JAR`).

PR #39259, opened 2024-01-21
Apply IntelliJ inspections to improve code quality

tobias-lippert:
I ran IntelliJ inspections on the Spring Boot codebase and found a couple of places worth applying the suggestions.
I've placed each of the different suggestions in a separate commit, so if you disagree with some of them, I can easily revert them.
The commit messages should indicate what's changed.


philwebb:
Thanks very much @tobias-lippert. I've merged almost all of these except the changes in the JSON code which we copied from another project.

PR #39265, opened 2024-01-22
TracingProperties exposes package-private PropagationType from public methods

vj-atlassian:
PropagationType enum is non-public outside of TracingProperties class which makes it unusable. I would request you to make it accessible so that we can access it to make customisation in tracing needs within our services.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/TracingProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/tracing/TracingProperties.java
@@ -251,7 +251,7 @@ public class TracingProperties {
 		/**
 		 * Supported propagation types. The declared order of the values matter.
 		 */
-		enum PropagationType {
+		public enum PropagationType {
 
 			/**
 			 * <a href="https://www.w3.org/TR/trace-context/">W3C</a> propagation.


wilkinsona:
We don't really consider the Java API of the configuration properties classes to be public API. Can you please describe what you're doing with `TracingProperties`?

vj-atlassian:
Sure. So we have our own customisation over Spring Boot provided tracing, where we support both W3C and B3 context propagation simultaneously. While Spring Boot doesn't support multiple tracing propagators but we do and have a code something like:

`for (String propType : propagationTypes) {
            switch (propType.toUpperCase()) {
                case "W3C" -> propagators.add(...);
                case "B3" -> propagators.add(...);
                default -> {}
            }
        }`

where propogationTypes refers to the spring boot property - `management.tracing.propagation.type`. Now, since the API is not available/public, I had to hardcode strings (W3C, B3) unnecessarily or the other option is duplicate the enum in my code (which didn't make sense IMO).

As per the Java API, I see that almost every other class is accessible publicly in `TracingProperties` but just the `TracingProperties.Propagation.PropagationType`.  Not sure if that was intentional though.

Let me know if that helps!

philwebb:
I think in this case it's an oversight that `PropagationType` isn't public. It's returned from the public `Propagation.getEffectiveConsumedTypes()` method so in my opinion either everything should be package-private or the enum should be public.

philwebb:
Flagging to see if others in the team agree.

wilkinsona:
Fine by me. The current part public, part package-private arrangement isn't ideal.

More broadly, I do wonder if we should change our policy about `@ConfigurationProperties` classes. While we document that they're not really public API they are, in Java terms, public API so it's almost inevitable that people will use them. We even [encourage](https://docs.spring.io/spring-boot/docs/3.2.2/reference/html/howto.html#howto.data-access.configure-two-datasources) some usage ourselves.

philwebb:
Thanks very much @vj-atlassian, this has now been merged into 3.1.x and forward.

PR #39274, opened 2024-01-23
feat: make finalName editable

tisonkun:
We recommend configuring finalName to change the artifact name but mark it as read-only. It's better to make it editable.


tisonkun:
Oops .. I forget I encountered this ever 🤣

PR #39276, opened 2024-01-23
Unify 'observation-enabled' property defaults

### LINKED ISSUE #38737: Unify 'observation-enabled' property defaults if possible
See #36451 and #38644 for examples.

Currently some technologies have an `observation-enabled` property which can be used to enable/disable observations. We should try to align the default value of this property for all of them.

There may be performance reasons why we currently don't enable all observations by default.

Wzy19930507:
change spring.pulsar.listener.observation-enabled and spring.pulsar.template.observations-enabled to false

Closes #38737


mhalbritter:
Thank you!

PR #39283, opened 2024-01-24
Log Trust Material source

### LINKED ISSUE #39057: SslBundle does miss an approriate toString() method

Wzy19930507:
log Trust Material source at `SslPropertiesBundleRegistrar.registerBundles` method.

If add Trust Material source at `SslBundle.toString`, need to pass Trust Material source to `JksSslStoreBundle`, `PemSslStoreBundle` and `WebServerSslStoreBundle`, feels like it violates the semantics of `SslBundle`. And `Tomcat` print Material source at `AbstractEndpoint.createSSLContext` method, so I implemented a similar method to solve this issue.

Closes #39057

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/SslPropertiesBundleRegistrar.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/SslPropertiesBundleRegistrar.java
@@ -25,6 +25,10 @@ import java.util.function.Function;
 import java.util.function.Supplier;
 import java.util.stream.Collectors;
 
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the suggestion but we really do want to do this through `toString()` rather than logging if possible. I've added the `pending-design-work` label to #39057 as we're not yet sure how to do that or even certain that it's possible in a way that we'll find acceptable.

Wzy19930507:
Thanks for review and explanation.

PR #39304, opened 2024-01-25
Add String to CronExpression conversion support

amiriahmad72:
Add String to CronExpression conversion support.

This converter use `CronExpression` `parse ()` method to convert String to CronExpression.

For example, it helps us to use type-safe `CronExpression` in `ConfigurationProperties` classes.


wilkinsona:
Thanks for the suggestion. Given that `CronExpression` is a Spring Framework type, I wonder if the should be added to Spring Framework rather than to Spring Boot. I'll discuss it with the Framework team.

wilkinsona:
@amiriahmad72 it would also be useful to know how you're using `CronExpression`. There's very little, if any, API in Spring Framework that uses it. For example, `CronTrigger` takes a `String` rather than a pre-parsed `CronExpression` instance. It's not yet clear to me how having such conversion support would be generally useful.

wilkinsona:
No feedback received. Closing. We can reconsider if there's a compelling case for this conversion support and it can't, for some reason, be implemented in Spring Framework.

PR #39308, opened 2024-01-25
Auto-configured ConcurrentPulsarListenerContainerFactory and PulsarConsumerFactory cannot be injected into injection points with specific generic type information

jonasgeiregat:
It is not possible to inject a specifically typed `PulsarConsumerFactory` nor `ConcurrentPulsarListenerContainerFactory` where the former is blocking creating custom typed consumers. 
@onobc you probably have an opinion on this.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
@@ -65,6 +65,7 @@ import org.springframework.pulsar.reader.PulsarReaderContainerProperties;
  * @author Soby Chacko
  * @author Alexander Preuß
  * @author Phillip Webb
+ * @author Jonas Geiregat
  * @since 3.2.0
  */
 @AutoConfiguration
@@ -131,7 +132,7 @@ public class PulsarAutoConfiguration {
 
 	@Bean
 	@ConditionalOnMissingBean(PulsarConsumerFactory.class)
-	DefaultPulsarConsumerFactory<Object> pulsarConsumerFactory(PulsarClient pulsarClient,
+	DefaultPulsarConsumerFactory<?> pulsarConsumerFactory(PulsarClient pulsarClient,
 			ObjectProvider<ConsumerBuilderCustomizer<?>> customizersProvider) {
 		List<ConsumerBuilderCustomizer<?>> customizers = new ArrayList<>();
 		customizers.add(this.propertiesMapper::customizeConsumerBuilder);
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
```diff
@@ -131,7 +132,7 @@ PulsarTemplate<?> pulsarTemplate(PulsarProducerFactory<?> pulsarProducerFactory,
 
 	@Bean
 	@ConditionalOnMissingBean(PulsarConsumerFactory.class)
-	DefaultPulsarConsumerFactory<Object> pulsarConsumerFactory(PulsarClient pulsarClient,
+	DefaultPulsarConsumerFactory<?> pulsarConsumerFactory(PulsarClient pulsarClient,
```
onobc, line 135: Might be a good idea to add a test (or adjust existing tests) for each of these scenarios in `org.springframework.boot.autoconfigure.pulsar.PulsarAutoConfigurationTests`.

There are tests that verify custom user beans are respected, but they inject in `PulsarConsumerFactory<String>` rather than attempt to use a specific type of these changed beans (DPCF, CPLCF). (e.g. `org.springframework.boot.autoconfigure.pulsar.PulsarAutoConfigurationTests.ConsumerFactoryTests#whenHasUserDefinedBeanDoesNotAutoConfigureBean`).

onobc (changes requested):
Thanks for the contribution @jonasgeiregat . 

These 2 beans were the only ones typed as `Object` so this brings things into a consistent state. I am not sure why we had them typed as such, but will see if @philwebb can refresh my memory. Either way, they should be consistent w/ the other bean types. 

My only suggestion is to add/adjust test for these 2 cases.

Disclaimer, I have a non-binding vote in Spring Boot project so will need to wait for one of the Boot team members to follow on w/ their review. 

Thanks

philwebb:
I'm not sure it was intentional. Despite many years of Java experience I still struggle with generics. Usually my default rule is wildcards on parameters and concrete types on return values so I probably just followed that.

> It is not possible to inject a specifically typed PulsarConsumerFactory

@jonasgeiregat It would be useful to see a sample or test that shows the failure with the current arrangement.

jonasgeiregat:
I have added a test that should prove the failure. Simply changing <?> back to  <Object> will trigger an injection error.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfiguration.java
jonasgeiregat, line 135: I also noticed a concrete impl. is in the signature as opposed to the interface `PulsarConsumerFactory` which is on its turn used in the `@ConditionalOnMissingBean`, is this deliberate?

wilkinsona: Yes. This provides as much type information as possible to the bean factory while allow the bean to back off when any `PulsarConsumerFactory` bean is defined.

mhalbritter:
Thank you!

PR #39313, opened 2024-01-26
Provide MvcRequestMatcher.Builder that correctly configures the servlet path

### LINKED ISSUE #37917: Provide PathPatternRequestMatcher.Builder that correctly configures the servlet path
If there is only one dispatcher servlet in the servlet context, then it would be nice to have a simpler way to construct `MvcRequestMatcher` instances.

As it is, an application that is constructing an `MvcRequestMatcher` must do:

```java
@Bean 
SecurityFilterChain appSecurity(HttpSecurity http, HandlerMappingIntrospector introspector) throws Exception {
    MvcRequestMatcher.Builder mvc = new MvcRequestMatcher.Builder(introspector);
    http
        .securityMatchers((security) -> security.requestMatchers(mvc.pattern("/controller/**")))
        /// ...
```

If they have configured `spring.mvc.servlet.path`, then they additionally need to provide that to the matcher as well like so:

```java
@Bean 
SecurityFilterChain appSecurity(HttpSecurity http, HandlerMappingIntrospector introspector) throws Exception {
    MvcRequestMatcher.Builder mvc = new MvcRequestMatcher.Builder(introspector).servletPath("/mvc");
    http
        .securityMatchers((security) -> security.requestMatchers(mvc.pattern("/controller/**")))
        /// ...
```

These two values, the `HandlerMappingIntrospector` and the servlet path configuration, Spring Boot already knows. Given that, it would be nice if a Boot application could instead do:

```java
@Bean 
SecurityFilterChain appSecurity(HttpSecurity http, MvcRequestMatcher.Builder mvc) throws Exception {
    http
        .securityMatchers((security) -> security.requestMatchers(mvc.pattern("/controller/**")))
        /// ...
```

I think it best to publish this bean in the event that there is only one `DispatcherServlet` servlet mapping.

The logic would be something like this (pseudocode follows):

```java
@Conditional(ExactlyOneDispatcherServlet.class)
@Bean 
MvcRequestMatcher.Builder mvcRequestMatcherBuilder(HandlerMappingIntrospector introspector, WebMvcProperties properties) {
    String servletPath = properties.getServlet().getPath();
    MvcRequestMatcher.Builder mvc = new MvcRequestMatcher.Builder(introspector)
    return ("/".equals(servletPath)) ? mvc : mvc.servletPath(servletPath);
}
```

where `ExactlyOneDispatcherServlet` checks the servlet configuration for multiple servlet mappings tied to servlets of type `DispatcherServlet`.

In Spring Security 6.2, the need for an `MvcRequestMatcher` when using `authorizeHttpRequests` [is greatly reduced](https://github.com/spring-projects/spring-security/issues/13568#issuecomment-1759913041). That said, it will still be needed for the foreseeable future when activating the default servlet or when using the `securityMatchers`, `csrf#ignoringRequestMatchers`, and `requiresSecure` DSLs.

Wzy19930507:
When only one `DispatcherServlet` servlet mapping, publish `MvcRequestMatcher.Builder` bean

Closes #37917


wilkinsona:
Unfortunately, I don't think we can accept a contribution for this one as dealing with `DispatcherServlet` is rather complicated and we need to be certain that we get it right when it relates to security. For example, one thing that we need to do is to use the `DispatcherServletPath`. I think we'll also need to spend some time thinking about how we can be certain that there's only a single `DispatcherServlet` and if there's any chance that one could be "hiding" behind a `ServletRegistrationBean`. Thanks anyway for the PR but I think it's better if the core team handle this one in collaboration with the Spring Security team. Sorry for the wasted effort here, I've added `pending-design-work` to #37917 to more accurately reflect its current status.

Issue #38641, opened 2023-12-04
Disable trace propagation if tracing is disabled


quaff:
According to https://github.com/spring-projects/spring-boot/issues/38625#issuecomment-1836120594, `management.tracing.enabled` is used to disable tracing only in tests currently, If we want to [dynamic disable tracing entirely](https://github.com/spring-projects/spring-boot/issues/38625#issuecomment-1838108363) we have to exclude `AutoConfiguration` by `spring.autoconfigure.exclude`, That's not handy especially for [profile combination](https://github.com/spring-projects/spring-boot/issues/27414).

philwebb:
We discussed this today and we'd like to make it so that `management.tracing.enabled` also disables the propagation of traces (probably by adding conditions to brave and otel beans).

mhalbritter:
I reworded the `management.tracing.enabled` property description in the documenation. That property will now disable trace exporting (sending traces to backends) and also trace propagation (e.g. sending trace and span ids via http for example).

PR #39333, opened 2024-01-30
Set LOG_CORRELATION_PATTERN only when correlation ID is expected

danielhcx:
Prior to this commit, if `logging.pattern.correlation` has been set, then the correlation ID is logged regardless of the flag `logging.expect-correlation-id`. This is different from what happens with the default correlation patterns.

This commit updates LoggingSystemProperties so that the system property `LOG_CORRELATION_PATTERN` is set when `logging.expect-correlation-id` is `true`. If it is `false` then the correlation ID won't be logged in any cases.

This also follows gh-38641 from the perspective of logging.


wilkinsona:
Thanks for the PR. This looks sensible to me but I'd like to double-check with the rest of the team and @jonatan-ivanov.

philwebb:
Sorry for the delay in reviewing this one. Looking again, I'm not sure that we should make this change. The `EXPECT_CORRELATION_ID_PROPERTY` is only supposed to change the default value of `CORRELATION_PATTERN`. If a user specifies their own `logging.pattern.correlation` property, then it's taken as a signal that they want correlation IDs.

If we change that logic, we could break existing application that want to set correlation IDs even if they aren't using Micrometer tracing.

@danielhcx Was there a specific use-case that caused you to submit this pull-request. If so, perhaps we can find a different way of solving it.

philwebb:
I'm going to close this one for now. If there's a specific need identified for why we need to make this change we can reopen the issue.

PR #39339, opened 2024-01-30
fix: HttpMessageConverters addDefaultConverters should only add if there is not exists one.

qxo:


wilkinsona:
Thanks for the proposal but this would be a breaking change as the existing converter is intentionally left in the list after its replacement. There's a comment to this effect in the test that you modified:

https://github.com/spring-projects/spring-boot/blob/10e750ab3bd3e711acb5cc6d406b4d9c1033519e/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/http/HttpMessageConvertersTests.java#L75

PR #39360, opened 2024-01-31
Remove System.out usage from Jetty GracefulShutdown

### LINKED ISSUE #39359: Usage of System.out in Jetty GracefulShutdown
When using graceful shutdown with Jetty, current requests are written with system.out instead of logger. Was it intentional? Here is the code piece:

org.springframework.boot.web.embedded.jetty.GracefulShutdown#awaitShutdown

```java
	private void awaitShutdown(GracefulShutdownCallback callback) {
		while (this.shuttingDown && this.activeRequests.get() > 0) {
			sleep(100);
		}
		System.out.println(this.activeRequests.get());
		this.shuttingDown = false;
		long activeRequests = this.activeRequests.get();
		if (activeRequests == 0) {
			logger.info("Graceful shutdown complete");
			callback.shutdownComplete(GracefulShutdownResult.IDLE);
		}
		else {
			logger.info(LogMessage.format("Graceful shutdown aborted with %d request(s) still active", activeRequests));
			callback.shutdownComplete(GracefulShutdownResult.REQUESTS_ACTIVE);
		}
	}
```

Looks like logger is already used in this method. I can fix it if ok.

okohub:
Fixes #39359

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/GracefulShutdown.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/GracefulShutdown.java
@@ -36,6 +36,7 @@ import org.springframework.util.ReflectionUtils;
  * Handles Jetty graceful shutdown.
  *
  * @author Andy Wilkinson
+ * @author Onur Kagan Ozcan
  */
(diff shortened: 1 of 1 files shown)


mhalbritter:
Thanks!

PR #39376, opened 2024-02-01
Make DataSourceBuilder be able to derive driverClassName from derived url

ronaldmik:
When using DataSourceBuilder to "deriveFrom" another DataSource, it is possible that the other Datasource has a getter for url, but not driverClassName. In this case, when the url is derived from the the provided DataSource, the DataSourceBuilder should try to derive the driverClassName from this derived url.

The use case is that when using Liquibase it is possible that a pooling DataSource is used that does not provide a getDriverClassName().  When using different credentials to run Liquibase with a DataSource, the user and password of LiquibaseProperties should be used and the DataSource should be "derivedFrom" the provided DataSource. This DataSource may provide a getUrl() but not a getDriverClassName(). In this case the driverClassName could also be "derived from" the "derived" url. This way, other custom (Pooling) DataSources may be used to run Liquibase using different credentials to connect to the 'same' DataSource.


philwebb:
Flagging to see if the team think we should consider this a bug and target the fix to 3.1 or an enhancement for 3.x

wilkinsona:
Feels like an enhancement to me. `@LiquibaseDataSource` could be used as a workaround in the meantime.

wilkinsona:
Thanks very much, @ronaldmik.

PR #39381, opened 2024-02-02
Add a check for HTTP(S) resource URLs without the '.xml' suffix

ashinnotfound:
Add a check for HTTP(S) resource URLs without the '.xml' suffix.

Currently, when loading logback configuration, the method of checking whether a URL refers to an XML file is as follows: 
```
if (url.getPath().endsWith(".xml") {
    // IS XML
} else {
    // IS NOT XML
}
```
However, some URLs' path referring to XML files do not end with ".xml". 

for example, 
`"http://${spring.cloud.nacos.config.server-addr}/nacos/v1/cs/configs?group=DEFAULT_GROUP&dataId=logback-common.xml"` refers to my logback configuration, but its path (`"/nacos/v1/cs/configs"`) doesn't end with ".xml", so it cannot be loaded correctly. 

Therefore, it is necessary to add an additional check.


wilkinsona:
Thanks for suggestion. Unfortunately, I don't think we should make an extra HTTP call when trying to load logging configuration. If you would like us to consider a different way of tackling this, please open an issue for consideration.

PR #39382, opened 2024-02-02
Polish

snicoll:
I am not 100% sure what we should do for `List.of` so let me know if you prefer to not have that.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/documentation/LoggersEndpointDocumentationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/documentation/LoggersEndpointDocumentationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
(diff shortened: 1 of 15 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/documentation/LoggersEndpointDocumentationTests.java
```diff
@@ -59,14 +59,10 @@ class LoggersEndpointDocumentationTests extends MockMvcEndpointDocumentationTest
...
-	private static final List<FieldDescriptor> groupLevelFields;
-
-	static {
-		groupLevelFields = Arrays
-			.asList(fieldWithPath("configuredLevel").description("Configured level of the logger group, if any.")
-				.type(JsonFieldType.STRING)
-				.optional(), fieldWithPath("members").description("Loggers that are part of this group"));
-	}
+	private static final List<FieldDescriptor> groupLevelFields = Arrays
+		.asList(fieldWithPath("configuredLevel").description("Configured level of the logger group, if any.")
```
wilkinsona, line 63: `List.of` here too perhaps?

wilkinsona (changes requested):
Thanks, @snicoll. `List.of` is all good. Our intention is that it's used "everywhere" other than for creating an empty list.

PR #39389, opened 2024-02-03
NoSuchMethod error when using the non-shaded Pulsar client and configuring authentications parameters

onobc:
This commit removes the use of the Pulsar ObjectMapperFactory when converting the authentication config props map to a JSON string. The Pulsar factory operates on a shaded returned value of Jackson ObjectMapper which may not exist when users are using the non-shaded version of the Pulsar client lib.

See https://github.com/spring-projects/spring-pulsar/issues/562


mhalbritter:
Thanks!

PR #39390, opened 2024-02-03
Match order between metadataLocation and reportLocation

sunshiningsoo:
Match Order between properties and constructor, 'metadataLocation' and 'reportLocation' for more reachable code


wilkinsona:
Thanks for the suggestion but this level of ordering isn't part of our coding conventions.

PR #39391, opened 2024-02-03
Add OAuth2-Client ConnectionDetails and Keycloak Docker-Compose support

### LINKED ISSUE #36777: ConnectionDetails for external providers in Spring Security modules
Part of my dev workflow involves using Testcontainers with Keycloak for integration testing against a local, preconfigured IdP/authorization server. It would be awesome if I could use the very same configuration with a running instance through spring-boot-testcontainers, but from what I can tell this isn't currently possible unless the respective configurators implement or expose a `ConnectionDetails` in some way. (I assume I could achieve something like it if I replace the Spring Security autoconfigurations with my own, but I'd prefer to stick with conventions as much as possible.)

The use cases that come to mind are:
 - OAuth2 client provider(s) - URLs defined under `spring.security.oauth2.client.provider`
 - SAML asserting party - Metadata URI defined under `spring.security.saml2.relyingparty.registration.<id>.assertingparty`
 - Probably a few other properties/modules I'm missing

Does this seem like a sensible use case for the ConnectionDetails abstraction? I've noticed the current implementations seem concerned with strictly persistent connections, so perhaps this does not fit the intent. But, since the provider autoconfigurations can lead to startup failure when the server is unreachable, the connections at least have some sort of persistent essence.

PhilKes:
Closes #36777
This PR includes:

* Adding the ConnectionDetails abstraction for OAuth2 `Registration` + `Provider`
* Docker-Compose Support for the [keycloak/keycloak](https://hub.docker.com/r/keycloak/keycloak) Image

Open Questions:
* Since the `Registration` has a lot fields that cant simply be extracted from any Keycloak env. variables, I added Docker service labels to set e.g. the `scope` of the client, or the client secret. Is this the best way to handle this?

  I also thought about not ignoring the `PropertiesOAuth2ClientConnectionDetails` if the Docker-Compose support already created an `OAuth2ClientConnectionDetails` Bean, but rather merge them together, so that the registration-id, provider are extracted from the Keycloak env. variables and the other settings (`scope`, `client-secret`, etc.) could be set in the `application.properties`. I am not sure which to prefer.
  
 When the open questions are answered, I would also add the TestContainer-Support for Keycloak.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientConnectionDetails.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/client/OAuth2ClientConnectionDetails.java
@@ -1,0 +1,246 @@
+/*
+ * Copyright 2012-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.security.oauth2.client;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Set;
+
+import org.springframework.boot.autoconfigure.service.connection.ConnectionDetails;
+
+/**
+ * Details required to establish connections to a OAuth2 authentication server.
+ *
+ * @author Philipp Kessler
+ * @since 3.3.0
+ */
+public interface OAuth2ClientConnectionDetails extends ConnectionDetails {
+
+	default Map<String, Registration> getRegistrations() {
+		return new HashMap<>();
+	}
+
+	default Map<String, Provider> getProviders() {
+		return new HashMap<>();
+	}
+
+	interface Registration {
+
+		/**
+		 * Reference to the OAuth 2.0 provider to use. May reference an element from the
+		 * 'provider' property or used one of the commonly used providers (google, github,
+		 * facebook, okta).
+		 * @return reference to the OAuth 2.0 provider to use.
+		 */
+		String getProvider();
+
+		/**
+		 * Client ID for the registration.
+		 * @return client ID for the registration.
+		 */
+		String getClientId();
+
+		/**
+		 * Client secret of the registration.
+		 * @return client secret of the registration.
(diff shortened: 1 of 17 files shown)


PhilKes:
Any opinions on this @philwebb , @mhalbritter? 😀

mhalbritter:
We're currently preparing the Spring Boot 3.3 release. This PR is planned as an enhancement for 3.4 and we'll look into it in detail when 3.3 is out.

wilkinsona:
Thanks for the PR, @PhilKes. Can you please describe how you identified the properties that should be part of `OAuth2ClientConnectionDetails`? My first impression is that there are too many of them but I may be overlooking something.

The four different labels that are being used to supply some of the values feels like a smell to me as it tells me that these settings (probably) aren't coming from the Docker Compose-managed service but from user configuration. Typically, we'd want the user's standard configuration to be reused (so that their use of Docker Compose matches as closely as possible to running their application "properly") and where test-specific configuration is needed, it can be provided through application properties rather than Docker Compose labels.

PhilKes:
> Thanks for the PR, @PhilKes. Can you please describe how you identified the properties that should be part of `OAuth2ClientConnectionDetails`? My first impression is that there are too many of them but I may be overlooking something.
> 
> The four different labels that are being used to supply some of the values feels like a smell to me as it tells me that these settings (probably) aren't coming from the Docker Compose-managed service but from user configuration. Typically, we'd want the user's standard configuration to be reused (so that their use of Docker Compose matches as closely as possible to running their application "properly") and where test-specific configuration is needed, it can be provided through application properties rather than Docker Compose labels.

Thanks for your remarks @wilkinsona 
I wasnt really sure what the minimum amount of properties were, since it's not as trivial as e.g. host + user + password, so I simply copied the attributes from `OAuth2ClientProperties` for a first draft.
Now I suppose the bare minimum for `OAuth2ConnectionDetails` would be:
- Registration:
  - clientId
  - clientSecret
  - redirectUri
  - provider 
- Provider:
  - issuerUri?
  
All other properties should then only be set via `OAuth2ClientProperties`?

About the labels, the `clientId` and `clientSecret` are set for a `Registration` which has a specified name. If the user defined a `Registration` in the `application.properties`, how do we match that `Registration` to the `Provider` in e.g. `KeycloakDockerComposeConnectionDetailsFactory`? In theory there could be more than 1 `Registration` or `Provider`, so how do we know which `Provider` and `Registation` name should be used in `KeycloakDockerComposeConnectionDetailsFactory`?

wilkinsona:
I think the breadth of the connection details is going to be quite hard to get right.

I'm not sure that it makes sense to be able to override a registration's provider when using Docker Compose or Testcontainers. If you're using one of the common providers, would we want to encourage overriding it to use a "test" provider? I'm not sure that we would as what you're testing would be quite far removed from what will happen at runtime.

I think I'll take this to #36777 as that's probably a better place to figure it out.

wilkinsona:
Thanks again for the PR, @PhilKes. Unfortunately, it's become clear from https://github.com/spring-projects/spring-boot/issues/36777 that I don't think we're ready to start implementing something and it looks like this may not be heading in the right direction. I'm going to close this one so that we can take a step back and figure out exactly what we want to do.

Issue #39388, opened 2024-02-03
Debug the Application documentation uses deprecated features


craigmit:
The documentation on how to debug here:  https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#run.examples.debug says to use:
```
<jvmArguments>
  -Xdebug -Xrunjdwp:transport=dt_socket,server=y,suspend=y,address=5005
</jvmArguments>
```
Since Java8, the -Xdebug support has been dropped.  Ref:  https://docs.oracle.com/javase/8/docs/technotes/tools/unix/java.html#BABHDABI

While it still works, the better way to do it is:
```
<jvmArguments>
  -agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005
</jvmArguments>
```
And via the command line:
```
$ mvn spring-boot:run -Dspring-boot.run.jvmArguments=-agentlib:jdwp=transport=dt_socket,server=y,suspend=n,address=*:5005
```
You also no longer need the quotes, so the "make sure to wrap that properly" comment can also be removed.

Cheers, and thanks for the awesome Spring Boot product!

scottfrederick:
Closing in favor of #39392.

PR #39392, opened 2024-02-03
Update the Debugging Documentation of the Spring Boot Maven Plugin 

lukasdo:
Replaced the old JVM Argument -Xdebug (which does nothing according to documentation) with newer ones as described in  #39388.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/running-your-application.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/running-your-application.adoc
@@ -40,7 +40,7 @@ Doing so lets you attach a debugger to your packaged application, as shown in th
 
 [source,shell,indent=0,subs="verbatim"]
 ----
-	$ java -Xdebug -Xrunjdwp:server=y,transport=dt_socket,address=8000,suspend=n \
+	$ java -agentlib:jdwp=server=y,transport=dt_socket,address=8000,suspend=n \
 	       -jar target/myapplication-0.0.1-SNAPSHOT.jar
 ----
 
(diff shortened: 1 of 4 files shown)


wanger26:
Should we update the test for [parseDebugFlags](https://github.com/spring-projects/spring-boot/blob/d4affd7f851c318decd7da1eee71e12ec9c0e6d3/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/test/java/org/springframework/boot/maven/RunArgumentsTests.java#L65-L71) as well as the [running-your-application](https://github.com/spring-projects/spring-boot/blob/c416c17fe85507e82ea172d43b54086a1a8d02a4/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/running-your-application.adoc#L38-L45) doc to reflect this change?

lukasdo:
I added another commit for the [running-your-application](https://github.com/spring-projects/spring-boot/blob/c416c17fe85507e82ea172d43b54086a1a8d02a4/spring-boot-project/spring-boot-docs/src/docs/asciidoc/using/running-your-application.adoc#L38-L45) documentation. 

An updated `parseDebugsFlags` test could look like this:
```	@Test
	void parseDebugFlag() {
		String[] args = parseArgs("-agentlib:jdwp=server=y,transport=dt_socket,address=8000,suspend=n");
		assertThat(args).hasSize(1);
		assertThat(args[0]).isEqualTo("-agentlib:jdwp=server=y,transport=dt_socket,address=8000,suspend=n");
	}
```

I am not really sure if that would make sense as it would be just testing if one argument is parsed correctly which is also tested in e.g. `parseWithExtraSpaces`. If desired I could update this test or delete it completely.

wanger26:
Excellent, thank you 💯  It might be worth updating the test to ensure that the parsing works as expected in cases changes are made in the parser in the future.

wanger26 (commented):
LGTM 🚀

mhalbritter:

Issue #39236, opened 2024-01-18
Webflux actuator endpoints respond with 500 when a parameter is missing


mhalbritter:
Take this endpoint:

```java
@Endpoint(id = "test")
@Component
public class TestEndpoint {
    @ReadOperation
    public Mono<String> sessionsForUsername(String username) {
        return Mono.just("test");
    }
}
```

When I execute `curl http://localhost:8080/actuator/test`, I get

```
HTTP/1.1 500 Internal Server Error
Content-Type: application/json
Content-Length: 139

{
    "timestamp": "2024-01-18T10:44:27.123+00:00",
    "path": "/actuator/test",
    "status": 500,
    "error": "Internal Server Error",
    "requestId": "3076ac2f-2"
}
```

When I remove the `Mono`, or I use WebMVC, i get a 400 Bad Request, like i would have expected it.

```
HTTP/1.1 400 
Content-Type: application/json
Transfer-Encoding: chunked
Date: Thu, 18 Jan 2024 10:45:36 GMT
Connection: close

{
    "timestamp": "2024-01-18T10:45:36.289+00:00",
    "status": 400,
    "error": "Bad Request",
    "path": "/actuator/test"
}
```

Reproducer: [repro.zip](https://github.com/spring-projects/spring-boot/files/13976998/repro.zip)

scottfrederick:
Closing in favor of #39397

PR #39397, opened 2024-02-05
Webflux actuator endpoints respond with 500 when a parameter is missing

kse-music:
Closes gh-39236


scottfrederick:
@kse-music After writing a test to reproduce the original problem, it looked like your fix would work. Unfortunately, after further testing the change you've made is not correct. The `getInvoker` method that was changed should return an `OperationInvoker`, but with this change it is actually invoking the invoker (with `Mono.fromCallable(() -> invoker.invoke(context))`) and returning the results of the invocation. In essence, the invocation is happening too early after this change, which causes existing tests to fail. I'm testing a different fix. 

Thanks anyway for the contribution.

PR #39399, opened 2024-02-05
Showcase and fix for actuator configprops endpoint serialization issu…

tomekrojek:
…e of nested records (and fix).

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
@@ -533,8 +533,7 @@ public class ConfigurationPropertiesReportEndpoint implements ApplicationContext
 			// that's what the metadata generator does. This filter is not used if there
 			// is JSON metadata for the property, so it's mainly for user-defined beans.
 			return (setter != null) || ClassUtils.getPackageName(parentType).equals(ClassUtils.getPackageName(type))
-					|| Map.class.isAssignableFrom(type) || Collection.class.isAssignableFrom(type)
-					|| parentType.isRecord();
+					|| Map.class.isAssignableFrom(type) || Collection.class.isAssignableFrom(type);
 		}
 
 		private AnnotatedMethod findSetter(BeanDescription beanDesc, BeanPropertyWriter writer) {
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
@@ -533,7 +533,8 @@ public class ConfigurationPropertiesReportEndpoint implements ApplicationContext
 			// that's what the metadata generator does. This filter is not used if there
 			// is JSON metadata for the property, so it's mainly for user-defined beans.
 			return (setter != null) || ClassUtils.getPackageName(parentType).equals(ClassUtils.getPackageName(type))
-					|| Map.class.isAssignableFrom(type) || Collection.class.isAssignableFrom(type);
+					|| Map.class.isAssignableFrom(type) || Collection.class.isAssignableFrom(type)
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
```diff
@@ -533,7 +533,8 @@ private boolean isReadable(BeanDescription beanDesc, BeanPropertyWriter writer)
 			// that's what the metadata generator does. This filter is not used if there
 			// is JSON metadata for the property, so it's mainly for user-defined beans.
 			return (setter != null) || ClassUtils.getPackageName(parentType).equals(ClassUtils.getPackageName(type))
-					|| Map.class.isAssignableFrom(type) || Collection.class.isAssignableFrom(type);
+					|| Map.class.isAssignableFrom(type) || Collection.class.isAssignableFrom(type)
+					|| parentType.isRecord();
```
wilkinsona, line 537: I'm not sure that this is the correct fix, unfortunately. From a nesting perspective, records and classes are the same and the existing comparison of the package names should be sufficient. I suspect that expanding it to whenever a parent is a record may cause problems for richer, non-scalar types that shouldn't be included but now will be.

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpointSerializationTests.java
```diff
@@ -648,4 +648,91 @@ public String convert(ComplexProperty source) {
...
+					.getBean(ConfigurationPropertiesReportEndpoint.class);
+			ConfigurationPropertiesDescriptor applicationProperties = endpoint.configurationProperties();
+			ConfigurationPropertiesBeanDescriptor nestedRecords = applicationProperties.getContexts()
+					.get(context.getId())
+					.getBeans()
+					.get("nestedRecords");
+			assertThat(nestedRecords.getPrefix()).isEqualTo("nested-records");
+
+			Map<String, Object> propertiesMap = nestedRecords.getProperties();
+			assertThat(propertiesMap).containsOnlyKeys("exposedNested1", "hiddenNested1");
```
wilkinsona, line 668: I'm a little confused by the naming here. Why is one `exposed` and one `hidden`? As far as I can tell, their visibility in the endpoint's output should be the same and the assertion seems to confirm this.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
tomekrojek: Thanks for looking so quickly into this, I'm not sure what to tell you as tbh not sure what this impl is trying to do in the first place - this fix was just aimed to make those fields visible in serialized json. Can you share some light on purpose of this class? What it is trying to filter?

spring-boot-project/spring-boot-actuator/src/test/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpointSerializationTests.java
tomekrojek: Sorry just copy-pasted test case from our code, we're using custom annotations (that I didn't copy) to hide some of the fields - naming remained but yeah probably now it should be changed.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/context/properties/ConfigurationPropertiesReportEndpoint.java
tomekrojek: I went through binding logic and understood that records are handled with matching constructor parameters - and as I was using multiple records constructors I was missing `@ConstructorBinding` annotation on one of them. It's working fine now - closing PR.

tomekrojek:
I went through binding logic and understood that records are handled with matching constructor parameters - and as I was using multiple records constructors I was missing `@ConstructorBinding` annotation on one of them. It's working fine now - closing PR.

PR #39403, opened 2024-02-05
SpringProfile arbiter fails without a Spring's environment

timgrohmann:
A issue has been raised and fixed in `apache/logging-log4j2`: https://github.com/apache/logging-log4j2/issues/1783

A detailed explanation of the proposed change may be found in this issue.

Basically, the null check for the environment _is not necessary_ in the builder as the `isCondition()` method already correctly checks for a null environment.


wilkinsona:
I don't think we can do that as it would introduce an (even stronger) cyclic dependency between the two projects.

timgrohmann:
Can we still integrate the upstream change into this project to fix the problem for Spring Boot users?

mhalbritter:
Yeah, I'll flag it to talk about it in our team meeting.

PR #39430, opened 2024-02-07
Add possibility for custom MimeMappings

BenchmarkingBuffalo:
Add a new property called 'mime-mappings' under the 'server' property. 
This is a key-value-map, which is added to the default MimeMappings.

If their is a need to adjust the documentation, just hit me up and I will add this.

See #38902


BenchmarkingBuffalo:
Sure thing, I rebased it on main and fixed the conflicts.
closes #38902

mhalbritter:
Thanks a lot!

Issue #37105, opened 2023-08-25
Support config property --> object field custom mapping 


mipo256:
The following [SO question](https://stackoverflow.com/questions/52960921/spring-boot-configurationproperties-change-property-key) illustrates the problem well.

Long story short, sometimes I want to have a property in my config file, which name is different from the object field. As far as I know, and seems it is still the case, there is no way to do it other than generating some custom setters, but that seems to be a workaround, not a solution. 

It would be great to introduce a new annotation (or another way, but the annotation seems to be the smoothest) to handle this case, like:

```
class MyS3Config {

     /**
      * This annotation of course does not exists, but this is just an example
      * I cannot have an identifier named 'public' in Java, so I have to do custom mapping
      */
     @ConfigurationPropertiesMapping(property = "public")
     private String publicBucket;
}
```

Thanks  in advance

quaff:
Could you explain why you would like to use a different key in config file?
There is another solution:
```yaml
public: ***
publicBucket: ${public}
```

Mikhail2048:
Sure.
Example you provided, as I pointed out specifically, is not a solution, it is a workaround to just make this work. What we have in reality is this: there is a `Buckets` nested condif class in `AWSConfig` in our domain. We have a naming conention for decades of buckets, like:

```
company.iam.buckets.<bucket_name>
```

and the config from the top level looks like this:

```
public class AWSConfig {

    static class Buckets {
    } 
}
```

So now image we have a bucket named public. We cannot use `public` as an identifier in java. The only option now is to name it like:

```
company.iam.buckets.publicBucket
```

but we already have a bunch of them named like: 
```
company.iam.buckets.local
company.iam.buckets.restricted
company.iam.buckets.data
company.iam.buckets.terraform
```
So we would break the naming convention. Or we need to change the naming convention just becasue of one bucket. It is clearly easier to just name the bucket in Java code as `publicBucket`, but retain the properties naming convention in config.

I hope it is clear now.

quaff:
> We cannot use `public` as an identifier in java. 

I got the point.

quaff:
Introducing a new annotation will make things more complicated, I think custom setter `setPublic(String value)` or placeholder `publicBucket: ${public}` is graceful enough, If the config class is maintained by yourself then adding setter is more appropriate, otherwise use placeholder instead.

Mikhail2048:
[...]

PR #39452, opened 2024-02-08
Support @Name with JavaBean-based configuration properties

BenchmarkingBuffalo:
Makes the Name annotation applicable on fields and takes this name when looking up the corresponding property.

See #37105


wilkinsona:
Thank you, @BenchmarkingBuffalo.

Issue #39324, opened 2024-01-29
Provide a customizer callback for the WebHttpHandlerBuilder that creates the auto-configured HttpHandler


mpapenbr:
According to the [documentation](https://docs.spring.io/spring-boot/docs/current/reference/html/web.html#web.reactive.webflux.web-filters) all `WebFilter` Beans are added to each exchange.

When using multiple `SecurityWebFilterChain`s that share some, but not all available `WebFilter` I found no way to declare such `WebFilter` as Beans. Instead I have to define them locally in the method creating the `SecurityWebFilterChain`.

In the Servlet-world there is a way via `FilterRegistrationBean` to prevent the addition of a Filter like this.
```
@Bean
public FilterRegistrationBean<OncePerRequestFilter> myFilterRegistration(
		@Qualifier("myCustomFilter") OncePerRequestFilter filter) {
  FilterRegistrationBean<OncePerRequestFilter> reg = new FilterRegistrationBean<>(filter);
  reg.setEnabled(false);
  return reg;
}
```
Please provide a similar functionality for `WebFilter`

wilkinsona:
Spring Boot has quite limited control over this. The gathering of all of the `WebFilter` beans from the application context is done by Spring Framework's `WebHttpHandlerBuilder` when Boot initialises it using the application context:

https://github.com/spring-projects/spring-boot/blob/df67dec561e1cd9b21c3110c6b417c7e7f18cacb/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/HttpHandlerAutoConfiguration.java#L62-L71

You could try defining your own `HttpHandler` bean which will cause the bean above to back off. One possibility would be to use the builder's `filters(Consumer<List<WebFilter>> consumer)` and remove from the list those filters found in the application context that you don't want to be included.

If the above works for you, we could consider a `WebHttpHandlerBuilder` customizer callback interface that would allow you to customize the auto-configured builder and avoid the need to define your own `HttpHandler`.

mpapenbr:
Thanks for the hint, using the `filters(Consumer<List<WebFilter>> consumer)` produces the wanted results. 
A customizer callback would be nice.

```
...
@Bean
Consumer<List<WebFilter>> globalFilterCustomizer() {
        // for demo purpose: I don't want any filters in the global chain that contain CustomFilter in the class name
[...]

PR #39467, opened 2024-02-08
Add customizer callback for WebHttpHandlerBuilder

BenchmarkingBuffalo:
Add a new interface for customizing the WebHttpHandlerBuilder before the HttpHandler is built from it.

For the test, there is also the alternative to use a customizer adding a filter and then take AssertJ and some InstanceOf checks to see, if the filter was added to the chain.
It works as well, but seemed a bit long in code to me.
If you think, this is a better test, hit me up and I will change it.

See gh-39324

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/HttpHandlerAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/HttpHandlerAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -41,6 +41,7 @@ import org.springframework.web.server.adapter.WebHttpHandlerBuilder;
  *
  * @author Brian Clozel
  * @author Stephane Nicoll
+ * @author Lasse Wulff
  * @since 2.0.0
  */
 @AutoConfiguration(after = { WebFluxAutoConfiguration.class })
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/HttpHandlerAutoConfiguration.java
```diff
@@ -60,8 +61,11 @@ public AnnotationConfig(ApplicationContext applicationContext) {
 		}
 
 		@Bean
-		public HttpHandler httpHandler(ObjectProvider<WebFluxProperties> propsProvider) {
-			HttpHandler httpHandler = WebHttpHandlerBuilder.applicationContext(this.applicationContext).build();
+		public HttpHandler httpHandler(ObjectProvider<WebFluxProperties> propsProvider,
+				ObjectProvider<WebHttpHandlerBuilderCustomizer> handlerBuilderCustomizers) {
+			WebHttpHandlerBuilder handlerBuilder = WebHttpHandlerBuilder.applicationContext(this.applicationContext);
+			handlerBuilderCustomizers.forEach((customizer) -> customizer.customize(handlerBuilder));
```
wilkinsona, line 67: Can you please use `orderedStream()` here so that the customisers will be applied in the order determined by their `@Order` annotation or `Ordered` implementation?

wilkinsona (changes requested):
Thanks very much for the PR, @BenchmarkingBuffalo.

In addition to my comment about ordering, it would be good to mention the new callback very briefly in the reference docs, perhaps in [this section](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#web.reactive.webflux.auto-configuration). The source for that section can be found [here](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-docs/src/docs/asciidoc/web/reactive.adoc#spring-webflux-auto-configuration).

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/HttpHandlerAutoConfiguration.java
BenchmarkingBuffalo: Hi @wilkinsona,
I updated the PR and added some documentation.

scottfrederick:
Thanks for another great contribution @BenchmarkingBuffalo!

Issue #39471, opened 2024-02-08
Property placeholders aren't resolved when configuration property binding creates a Map from a property value using a converter


JohannLangloys:
When trying to fill an object that implements `Map` but is valued by a single property via a converter, the `MapBinder` class does not use a `PlaceholderResolver` after retrieving the property value. It can be seen in the method `bindAggregate`.

This result in placeholder not being resolved in the property.

The usecase seems strange, but this is the solution I find to allow multiple way of filing a pseudo Map object from either sub properties or a legacy string format.

JohannLangloys:
Just to add some more information.

In the Binder class, when binding a "simple" property :
```java
private <T> Object bindProperty(Bindable<T> target, Context context, ConfigurationProperty property) {
    context.setConfigurationProperty(property);
    Object result = property.getValue();
    result = this.placeholdersResolver.resolvePlaceholders(result);
    result = context.getConverter().convert(result, target);
    return result;
}
```
We can see the use of the `placeholdersResolver`.

Meanwhile in the MapBinder class, when binding a property corresponding to a map :
```java
if (property != null && !hasDescendants) {
    getContext().setConfigurationProperty(property);
    return getContext().getConverter().convert(property.getValue(), target);
}
```

The property is converted but no call to the `placeholdersResolver` is done.

wanger26:
You can assign this one to me, hopefully will have a fix for it shortly.

scottfrederick:
Closing in favor of #39507

PR #39507, opened 2024-02-10
Property placeholders aren't resolved when configuration property binding creates a Map from a property value using a converter

wanger26:
### Change Summary
- Add call to placeholder resolver to ensure property placeholders are resolved for the MapperBinder similar to [Binder.java](https://github.com/wanger26/spring-boot/blob/f6090227313fd8077b00851acd31c1fc6e860974/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/Binder.java#L459-L465)

### Issue
- https://github.com/spring-projects/spring-boot/issues/39471


scottfrederick:
Thanks again @wanger26. I've merged your changes along with a small [commit](https://github.com/spring-projects/spring-boot/commit/f90f29dd12020d9643da87587242e7086902cba4) to address formatting.

Issue #39453, opened 2024-02-08
Gradle plugin allows the use of Gradle 7.4 but the documented and tested minimum is 7.5


mhalbritter:
I think we missed to adapt `org.springframework.boot.gradle.plugin.SpringBootPlugin#verifyGradleVersion`. It still reads:

```java
	private void verifyGradleVersion() {
		GradleVersion currentVersion = GradleVersion.current();
		if (currentVersion.compareTo(GradleVersion.version("7.4")) < 0) {
			throw new GradleException("Spring Boot plugin requires Gradle 7.x (7.4 or later). "
					+ "The current version is " + currentVersion);
		}
	}
```

_Originally posted by @mhalbritter in https://github.com/spring-projects/spring-boot/issues/31215#issuecomment-1933867203_

scottfrederick:
Closing in favor of #39508

PR #39508, opened 2024-02-10
Gradle plugin allows the use of Gradle 7.4 but the documented and tested minimum is 7.5

wanger26:
### Change Summary
- Enforce Gradle version to be at least of v7.5 as the documented and tested minimum Gradle version is 7.5

### Issue
- https://github.com/spring-projects/spring-boot/issues/39453


wilkinsona (changes requested):
Thanks for the PR. Can you please also update `SpringBootPluginIntegrationTests` to match?

wanger26:
> Thanks for the PR. Can you please also update `SpringBootPluginIntegrationTests` to match?

My apologies, completely missed that test. Updated now :sunglasses:

scottfrederick:
@wanger26 Thanks for the contribution!

PR #39530, opened 2024-02-13
Modify isMatch method in InMemoryAuditEventRepository

rlawngus0910:
I think it's possible to return true/false values ​​even without a match variable.
Also, I think readability can be improved with early return.


wilkinsona:
WDYT, @philwebb? The approach of reassigning `match` rather than returning early seems to have been introduced in https://github.com/spring-projects/spring-boot/commit/516df88ea042b0b6de64c33b85c8b0b0ed0fe760.

Issue #39565, opened 2024-02-14
Document configuration property default for the show-values properties


mhalbritter:
See https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#application-properties.actuator.management.endpoint.configprops.show-values for example. They all default to `NEVER`.

There are three of them: `management.endpoint.configprops.show-values`, `management.endpoint.env.show-values` and `management.endpoint.quartz.show-values`.

mhalbritter:
This can be done by adding an entry in `spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json`. You can take `management.prometheus.metrics.export.histogram-flavor` (the backing code is `org.springframework.boot.actuate.autoconfigure.metrics.export.prometheus.PrometheusProperties#histogramFlavor`) for a blueprint, it's an enum with a default, too.

scottfrederick:
Closing in favor of #39589.

PR #39589, opened 2024-02-15
Document default configuration property for show-value props

lukasdo:
Documenting the default config properties as described by in GH-39565.


mhalbritter:
Thanks!

PR #39590, opened 2024-02-15
Issue 39565

jakesmolka:
Addressing: #39565

Dear @mhalbritter and team, I saw the ideal-for-contribution label and just gave it a try. Honestly, since there are just some lines of changes, I'm not sure if that is really what was requested.

Question for my understanding: Is the "When to show unsanitized values." line from the ref doc page parsed from the javadoc comment in the source? This metadata.json file has a `description` field, but not for all entries.

Targeting `3.1.x` branch, because of the milestone.


mhalbritter:
> Question for my understanding: Is the "When to show unsanitized values." line from the ref doc page parsed from the javadoc comment in the source?

Yes, this is parsed from the JavaDoc.

PR #39591, opened 2024-02-15
Document configuration property default for the show-values propertie…

jakesmolka:
Addressing: #39565

Dear @mhalbritter and team, I saw the ideal-for-contribution label and just gave it a try. Honestly, since there are just some lines of changes, I'm not sure if that is really what was requested.

Question for my understanding: Is the "When to show unsanitized values." line from the ref doc page parsed from the javadoc comment in the source? This metadata.json file has a `description` field, but not for all entries.

Targeting `3.1.x` branch, because of the milestone.

(Made a mistake in PR #39590 and created a new one, sorry.)

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -31,6 +31,14 @@
       "defaultValue": true
     },
     {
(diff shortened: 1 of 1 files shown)


scottfrederick:
@jakesmolka Thanks for the contribution. PR #39589 was opened just before yours, so I'm afraid we'll need to defer to the other PR. Please keep an eye out for other opportunities to contribute.

mhalbritter:
> Question for my understanding: Is the "When to show unsanitized values." line from the ref doc page parsed from the javadoc comment in the source? 

Yeah, it's being parsed from the JavaDoc.

PR #39605, opened 2024-02-16
Fix json path assertion

snicoll:
This commit fixes the json path assertion of a test. The second argument of `jsonpath` is an `Object[]` in case the expression is using placeholders.

We've had several tests not doing anything in the core framework for the same reason and we've decided to deprecate that method signature to require a fully-formed expression instead.


mhalbritter:
Thanks Stéphane!

PR #39606, opened 2024-02-16
Change format of git.commit.time used by git-commit-id-maven-plugin from RFC 822 to ISO 8601

TheSnoozer:
Related to https://github.com/git-commit-id/git-commit-id-maven-plugin/issues/674.

This change is required to make the times produced by the git-commit-id-maven-plugin usable for Maven's [reproducible builds]( https://maven.apache.org/guides/mini/guide-reproducible-builds.html).

Timestamp for reproducible output archive entries must either formatted as ISO 8601 `yyyy-MM-dd'T'HH:mm:ssXXX` or as an int representing seconds since the [epoch](https://reproducible-builds.org/docs/source-date-epoch/">SOURCE_DATE_EPOCH)

Example usage might be
E.g.
```
   <properties>
     <project.build.outputTimestamp>${git.commit.time}</project.build.outputTimestamp>
   </properties>
```

### DIFF
--- a/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
+++ b/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
@@ -148,7 +148,7 @@ publishing.publications.withType(MavenPublication) {
 							}
 							configuration {
 								delegate.verbose('true')
-								delegate.dateFormat("yyyy-MM-dd'T'HH:mm:ssZ")
+								delegate.dateFormat("yyyy-MM-dd'T'HH:mm:ssXXX")
 								delegate.generateGitPropertiesFile('true')
 								delegate.generateGitPropertiesFilename('${project.build.outputDirectory}/git.properties')
 							}


michael-o (commented):
LGTM

jonatan-ivanov:
Isn't `yyyy-MM-dd'T'HHH:mm:ssZ` (2024-02-16T19:32:16Z) in fact conforms ISO 8601?

>If the time is in [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time), add a Z directly after the time without a space. Z is the zone designator for the zero UTC offset. "09:30 UTC" is therefore represented as "09:30Z" or "T0930Z". "14:45:15 UTC" would be "14:45:15Z" or "T144515Z".

([wiki: ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Coordinated_Universal_Time_(UTC)))

michael-o:
> Isn't `yyyy-MM-dd'T'HHH:mm:ssZ` (2024-02-16T19:32:16Z) in fact conforms ISO 8601?
> 
> > If the time is in [UTC](https://en.wikipedia.org/wiki/Coordinated_Universal_Time), add a Z directly after the time without a space. Z is the zone designator for the zero UTC offset. "09:30 UTC" is therefore represented as "09:30Z" or "T0930Z". "14:45:15 UTC" would be "14:45:15Z" or "T144515Z".
> 
> ([wiki: ISO 8601](https://en.wikipedia.org/wiki/ISO_8601#Coordinated_Universal_Time_(UTC)))
> 
> That's what the Maven docs are using too on the page [you linked above](https://maven.apache.org/guides/mini/guide-reproducible-builds.html):
> 
> ```
> <project.build.outputTimestamp>2023-01-01T00:00:00Z</project.build.outputTimestamp>
> ```

No, you are confusing to `yyyy-MM-dd'T'HHH:mm:ss'Z'` with `yyyy-MM-dd'T'HHH:mm:ssZ`.

jonatan-ivanov:
Oops, indeed, I meant `Z` in the output (UTC + ISO 8601), I missed that in the diff too. `XXX` does that if the time is in UTC but I'm not sure if that is guaranteed (if it isn't maybe that should be the concern of another issue). Btw, I think you also meant `HH` instead of `HHH`.

mhalbritter:
So, `git.commit.time=2024-02-19T10\:18\:10+0100` is bad, and `git.commit.time=2024-02-19T10\:18\:10+01\:00` is good? The latter can be parsed by `Instant.parse`, too

michael-o:
> So, `git.commit.time=2024-02-19T10:18:10+0100` is bad, and `git.commit.time=2024-02-19T10:18:10+01:00` is good? The latter can be parsed by `Instant.parse`, too.

Correct.

mhalbritter:
Note to myself: Check what the Gradle plugin does.

michael-o:
@mhalbritter Please change the title because ISO 8601 is not aware of any timezones, but offsets only.

mhalbritter:

TheSnoozer:
Great, thanks for getting this integrated :-)

wilkinsona:
This has been superseded by https://github.com/spring-projects/spring-boot/issues/40015 as we can now use the plugin's default.

PR #39610, opened 2024-02-17
Modify toCharArray loop of parameterId

rlawngus0910:
I think using charAt rather than toCharArray can reduce the cost of creating a char array. Also, I think the number of calls can be reduced by declaring the length of name as a variable.


bclozel:
Thanks for your contribution, but this code belongs to a documentation task in our build, so micro-optimizations aren't really worth it, especially if they make the code harder to read. This code isn't part of any code path for Spring Boot applications and this change isn't likely to speed up our build significantly.

Issue #39314, opened 2024-01-26
Configure suitable TaskExecutor for WebSocket when virtual threads are enabled


snicoll:
A recent issue in Spring Framework (https://github.com/spring-projects/spring-framework/issues/32081) made us aware that we overlooked websocket broker support for virtual threads. 

Now that `ChannelRegistration` allows a plain `TaskExecutor` to be specified via the new `executor` method, it would be nice to configure it when virtual threads are enabled.

BenchmarkingBuffalo:
HI @wilkinsona,
I can work on this, once spring 6.1.4 is out.
You can assign it to me, if you like.

mhalbritter:
Superseded by #39611.

PR #39611, opened 2024-02-17
Configure suitable TaskExecutor for WebSocket when virtual threads are enabled

BenchmarkingBuffalo:
For WebSocket support the preconfigured ThreadExecutor is set for ChannelRegistrations similar to the JpaRepositoriesAutoConfiguration.

See gh-39314

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/jpa/JpaRepositoriesAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/jpa/JpaRepositoriesAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/testcontainers/BitnamiImageNames.java
```diff
@@ -32,7 +32,7 @@ public final class BitnamiImageNames {
 
 	private static final String MARIADB_VERSION = "11.2.3";
 
-	private static final String MONGO_VERSION = "7.0.5";
```
mhalbritter, line 35: This doesn't look right.

BenchmarkingBuffalo: This was necessary to get the build green (see https://github.com/bitnami/containers/issues/61882)
The image seems to be fixed now, I will remove the commit.

mhalbritter:
Thanks!

Issue #39310, opened 2024-01-25
Using SSL Bundle for server configuration ignores other server options


shollander:
I am attempting to use an SSL Bundle to configure the certificate used by the embedded web server.

If I do something like this in my `application.yaml` file:
```
spring:
  ssl:
    bundle:
      jks:
        mybundle:
          keystore:
            location: /path/to/keystore
            password: secret
server:
  ssl:
    enabled: true
    enabled-protocols:
      - TLSv1.3
    bundle: mybundle    
```

It works to configure the server certificate, but the other options (such as `server.ssl.enabled-protocols`) are ignored.
With this configuration, even though I configure it to only use TLSv1.3, it still responds to TLSv1.2 and TLSv1.3.

If I configure the certificate directly on `server.ssl` like this:
```
spring:
  ssl:
    bundle:
      jks:
        mybundle:
          keystore:
            location=/path/to/keystore
            password=secret
server:
  ssl:
    enabled: true
    enabled-protocols:
      - TLSv1.3
    key-store: /path/to/keystore
    key-store-password: secret
```
then the `enabled-protocols` setting works and only TLSv1.3 is enabled.

wilkinsona:
Rather than mixing bundle-based configuration and directly configuring the server's SSL settings, you should only use one or the other. For example:

```yaml
spring:
  ssl:
    bundle:
      jks:
        mybundle:
          keystore:
            location: /path/to/keystore
            password: secret
          options:
            enabled-protocols:
              - TLSv1.3
server:
  ssl:
    enabled: true
    bundle: mybundle 
```

We should look at improving the documentation here or perhaps improving the runtime behavior. It could fail-fast when mixing configuration or perhaps the server-specific configuration could somehow override the bundle configuration. Both may have implications for backwards compatibility though.

philwebb:
We're going to add a `NOTE:` to the documentation

scottfrederick:
Closing in favor of #39616

PR #39616, opened 2024-02-18
Add note regarding mixing of bundle-based and direct SSL configuration

wanger26:
### Change Summary
- Clarify that mixing of bundle-based configuration and directly configuring the server's SSL settings is not supported and that you should only use one or the other.

### Issue
- https://github.com/spring-projects/spring-boot/issues/39310

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/ssl.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/asciidoc/features/ssl.adoc
@@ -43,7 +43,7 @@ When used to secure a client-side connection, a `truststore` is typically config
 
 See {spring-boot-autoconfigure-module-code}/ssl/JksSslBundleProperties.java[JksSslBundleProperties] for the full set of supported properties.
 
-
+NOTE: When using SSL with Java KeyStore Files, only the bundle-based configurations will be applied as mixing of bundle-based configurations and directly configuring the server's SSL settings is not supported.
 
 [[features.ssl.pem]]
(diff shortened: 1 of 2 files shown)


scottfrederick:
@wanger26 Thanks for the PR. 

The issue #39310 is specific to the configuration of SSL for embedded web servers. As such, a better place for the note would be in the section that is specific to that configuration [here](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto.webserver.configure-ssl). 

Also, the use of the `ssl.bundle` property can't be combined with either the discrete JKS or PEM property options under `server.ssl`. 

Would you mind making those changes in a new commit and updating the PR?

wanger26:
> @wanger26 Thanks for the PR.
> 
> The issue #39310 is specific to the configuration of SSL for embedded web servers. As such, a better place for the note would be in the section that is specific to that configuration [here](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto.webserver.configure-ssl).
> 
> Also, the use of the `ssl.bundle` property can't be combined with either the discrete JKS or PEM property options under `server.ssl`.
> 
> Would you mind making those changes in a new commit and updating the PR?

I updated the wording and the location. Good catch @scottfrederick! :100:. Please let me know what you think of the new location and wording.

scottfrederick:
Thanks again @wanger26. I moved the note down just a bit further in the file and removed a few words in eb4d6e7c15099a8ecd1b8ad4aed9a551e2189149.

Issue #39246, opened 2024-01-19
bootBuildImage runs forever when a custom image name is used and the tag contains an illegal character


ilozano2:
**Spring Boot Version: 3.2.1**

### Issue

Given a image reference `<hostname>/<library>/<artifact>:1.0.0+1.2.3` 
When the length of <library> or <artifact> is long, the method `ImageReference.of(...)` freezes

Tests: https://github.com/spring-projects/spring-boot/commit/30c4bd11e2460a28ee4d381c7b58fab837958439

Related: https://github.com/spring-projects/spring-boot/issues/23115 

### TL;DR

Internally, the Regex.PATH generated is `^[a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?(?:(?:[\/][a-z0-9]+(?:(?:(?:[._]|__|[-]*)[a-z0-9]+)+)?)+)?$` which will report a [Catastrophic Backtracking](https://www.regular-expressions.info/catastrophic.html) (you can see it in regex101 for example).

wilkinsona:
See also https://github.com/spring-projects/spring-boot/issues/35657.

wilkinsona:
Thanks, @ilozano2. Is the plan for https://github.com/pivotal-cf/vmware-app-analyzer/issues/240 to include a `+` character in the image's tag? AIUI, `+` isn't a permitted character in a tag and any fix for this issue would result in such an image reference being rejected.

ilozano2:
@wilkinsona In fact, it isn't. Docker images don't support that character in the TAG. 

We are following a different format. No rush here then, but it could be nice to fail without freezing.
Thanks!

scottfrederick:
Closing in favor of #39617

PR #39617, opened 2024-02-18
FIX - Catastrophic Backtracking for images with illegal char in tag

wanger26:
### Change Summary
- Simplify regular expression checking for validity of Image Reference to fix Catastrophic Backtracking for long custom images with illegal characters in tag

### Issue
- https://github.com/spring-projects/spring-boot/issues/39246

### DIFF
--- a/spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/core/Regex.java
+++ b/spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/core/Regex.java
@@ -50,7 +50,7 @@ final class Regex implements CharSequence {
 	private static final Regex PATH_COMPONENT;
 	static {
 		Regex segment = Regex.of("[a-z0-9]+");
-		Regex separator = Regex.group("[._]|__|[-]*");
+		Regex separator = Regex.group("[._-]{1,2}");
 		Regex separatedSegment = Regex.group(separator, segment).oneOrMoreTimes();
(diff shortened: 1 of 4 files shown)


scottfrederick:
Thanks @wanger26. This change looked good to me, but a test somewhere else in the project failed. A test added to `ImageReferenceTests` like this passes without the change in this PR (as it should) but fails after the change: 

```
	@Test
	void ofSimpleNameWithSingleCharacterSuffix() {
		ImageReference reference = ImageReference.of("ubuntu-a");
		assertThat(reference.getDomain()).isEqualTo("docker.io");
		assertThat(reference.getName()).isEqualTo("library/ubuntu-a");
		assertThat(reference.getTag()).isNull();
		assertThat(reference.getDigest()).isNull();
		assertThat(reference).hasToString("docker.io/library/ubuntu-a");
	}
```

A single character (lower-case alpha or numeric) after a `-` separator does not parse correctly. 

Note that the `RegEx`, `ImageReference` and `ImageReferenceTests` are all (nearly) duplicated in the project. The `RegEx` and `ImageReferenceTests` classes will need the same change in both places. 

Do you want to try to address this issue?

wanger26:
@scottfrederick I can address the build issue as well as make the recommended adjustments. I am going to steal that test case too if you don't mind to make sure we get it right :100:

scottfrederick:
Thanks very much @wanger26. The change has been merged along with a small polishing commit in c93acdafbd6de650f5afc7e2d33ff4689f7907c4.

Issue #39529, opened 2024-02-12
Document that task executor pool size properties are ignored when using virtual threads


peacemaker123456:
Hello dear team,

I noticed, that setting `spring.threads.virtual.enabled` to true, will always lead to the usage of `SimpleAsyncTaskExecutor`, therefore ignoring the config settings affecting the pool size. Likely you chose to do that because pooling with virtual threads makes no sense, given the low overhead of virtual thread creation. Yet it's not clear enough without looking at the source code `TaskExecutorConfiguration`. So maybe docs could be clearer here.

The only possibility to limit the amount of parallel tasks is therefore the `spring.task.execution.simple.concurrency-limit` property, which essentially behaves like a max pool size. The issue I see here, is that `ConcurrencyThrottleSupport`, which is being used under the hood, uses the synchronized blocks which leads to thread pinning and might cause several issues with virtual threads. Maybe switching to `ReentrantLock` would mitigate that?

philwebb:
Thanks @peacemaker123456. We'll use this issue for the documentation updates and I've opened https://github.com/spring-projects/spring-framework/issues/32251 to see if Framework want to make changes to `ConcurrencyThrottleSupport`

mhalbritter:
Superseded by #39629.

PR #39629, opened 2024-02-20
Document that task executor pool size properties are ignored when using virtual threads

wanger26:
### Change Summary
- Document that task executor pool size properties are ignored when using virtual threads

### Issue
- https://github.com/spring-projects/spring-boot/issues/39529


mhalbritter:
Thanks!

Issue #39635, opened 2024-02-20
Spring boot plugin issue using gradle cache


alexisgayte:
#39326 
Running Spring boot plugin with gradle 8.5 generate :

```
Configuration cache problems found in this build.
2 problems were found storing the configuration cache, 1 of which seems unique.

Task :bootJar of type org.springframework.boot.gradle.tasks.bundling.BootJar: execution of task ':bootJar' caused invocation of 'Task.project' in other task at execution time which is unsupported.
See https://docs.gradle.org/8.5/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution
Execution of task ':bootJar' caused invocation of 'Task.project' by task ':resolveMainClassName' at execution time which is unsupported.
```

```
org.gradle.api.InvalidUserCodeException: Execution of task ':bootJar' caused invocation of 'Task.project' by task ':resolveMainClassName' at execution time which is unsupported.
	at org.gradle.configurationcache.problems.DefaultProblemFactory$problem$1.exception(DefaultProblemFactory.kt:52)
	at org.gradle.configurationcache.initialization.DefaultConfigurationCacheProblemsListener.onTaskExecutionAccessProblem(ConfigurationCacheProblemsListener.kt:116)
	at org.gradle.configurationcache.initialization.DefaultConfigurationCacheProblemsListener.onProjectAccess(ConfigurationCacheProblemsListener.kt:62)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at java.base/jdk.internal.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:77)
	at java.base/jdk.internal.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.base/java.lang.reflect.Method.invoke(Method.java:568)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:36)
	at org.gradle.internal.dispatch.ReflectionDispatch.dispatch(ReflectionDispatch.java:24)
	at org.gradle.internal.event.DefaultListenerManager$ListenerDetails.dispatch(DefaultListenerManager.java:472)
	at org.gradle.internal.event.DefaultListenerManager$ListenerDetails.dispatch(DefaultListenerManager.java:454)
[...]

Issue #32980, opened 2022-11-02
Fix deprecation warnings due to tasks calling getProject() at execution time


wilkinsona:
Some of our custom tasks (`org.springframework.boot.build.autoconfigure.AutoConfigurationMetadata` is one) call `getProject()` at execution time. This results in a deprecation warning and means that our build will break with Gradle 9.0. While we won't be able to fix all such warnings in our build as some come from third-party plugins, we should update our own tasks.

0x100:
Hi! As far as I understood, the build should not break with Gradle 8.0. The deprecated class there is `org.gradle.api.internal.AbstractTask`, its removal should not affect the `getProject()` method.

wilkinsona:
The `getProject()` method isn't going away, but Gradle is going to prohibit/restrict calling it during task execution. That's what we want to address. To see the warnings you have to [enable the `STABLE_CONFIGURATION_CACHE` feature preview](https://docs.gradle.org/current/userguide/configuration_cache.html#config_cache:stable).

ThomazPassarelli:
Hi! I find this issue interesting. May I start working on it? Thanks in advance!

wilkinsona:
Thanks for the offer, @ThomazPassarelli. but I've already made a start on this (several months ago now) and just need to find some time to finish it off.

PR #39644, opened 2024-02-21
ResolveMainClassName calls getProject() at execution time, making it incompatible with Gradle's configuration cache

quaff:
>> A task must not use any Project objects at execution time. This includes calling Task.getProject() while the task is running.

See https://docs.gradle.org/7.0/userguide/configuration_cache.html#config_cache:requirements:use_project_during_execution

Fix GH-39635

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/plugin/ResolveMainClassName.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/plugin/ResolveMainClassName.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, @quaff. Can you please revert the changes within `buildSrc`? They're unrelated to #39635 as they'll only affect Boot's own build. Fixing that is a much bigger piece of work being tracked by #32980.

quaff:
> Thanks, @quaff. Can you please revert the changes within `buildSrc`? They're unrelated to #39635 as they'll only affect Boot's own build. Fixing that is a much bigger piece of work being tracked by #32980.

Done.

wilkinsona:
For merge with amendments as I'd like to see if we can test this.

wilkinsona:
We already appear to have tests for this as both `BootRunIntegrationTests` and `BootJarIntegrationTests` run with the configuration cache enabled and have tests that rely on resolution of the main class name. I've also tried and failed to reproduce the problem described in #39635 using a basic project generated from start.spring.io. As such, I don't think we can merge this as we don't know if the change is necessary and, more importantly, we don't know if it'll fix the reported problem. I'll leave it on hold for now until we make some progress on understanding the underlying problem.

wilkinsona:
Thanks again for the PR, @quaff, but I've decided to take a different approach here. Please see #40463.

Issue #39655, opened 2024-02-21
Spring Boot 3.2.2 new nested jar mechanism does not resolve resources containing space in the path


slovi:
Hi,
after we migrated from Spring Boot 3.1.8 to 3.2.2, our applications installed in path containing empty space suddenly stoped working.

The problem seems to be in new "nested jar" mechanism used to resource locating in the jar/war file.

I tried to prepare simple test which fails in 3.2.2:

  
```
    @BeforeAll
    static void setUpClass() {
        Handlers.register();
    }

    @Test
    void testNestedJarPathWithSpaceEncoded() throws Exception {

        var nestedJarUrl = new URL("jar:nested:/C:/Program%20Files/vynamic-view/incidentmanager.jar/!BOOT-INF/classes/!/static/app/rabclient-5.11.0.jar");
        var nestedJarConnection = nestedJarUrl.openConnection();
        assertThat(nestedJarConnection.getLastModified(), greaterThan(0L));
    }
```

This works fine in 3.1.8:

```
    @BeforeAll
    static void setUpClass() {
        JarFile.registerUrlProtocolHandler();
    }

    @Test
    void testNestedJarPathWithSpaceEncoded() throws Exception {

        var jarUrl = new URL("jar:file:/C:/Program%20Files/vynamic-view/incidentmanager.jar/!/BOOT-INF/classes/!/static/app/rabclient-5.11.0.jar");
        var jarConnection = jarUrl.openConnection();
        assertThat(jarConnection.getLastModified(), greaterThan(0L));
    }
```

I think that the problem is in implementation of the NestedUrlConnection class:

```
	private NestedLocation parseNestedLocation(URL url) throws MalformedURLException {
		try {
			return NestedLocation.parse(url.getPath());
		}
		catch (IllegalArgumentException ex) {
			throw new MalformedURLException(ex.getMessage());
		}
	}
```

Path should be decoded before passing it to NestedLocation.parse method. The URL passed to static method NestedLocation.fromUrl is also decoded before passed to NestedLocation.parse method.

So I would propose this change:

```
        private NestedLocation parseNestedLocation(URL url) throws MalformedURLException {
		try {
			return NestedLocation.parse(UrlDecoder.decode(url.getPath()));
		}
		catch (IllegalArgumentException ex) {
			throw new MalformedURLException(ex.getMessage());
		}
	}
```

Will create PR for that If you do not mind.

philwebb:
[...]

PR #39675, opened 2024-02-21
Decode URL content before passing it to NestedLocation.parse method.

slovi:
URL can contains empty spaced encoded as %20, so it should be decoded before passing it to NestedLocation. NestedLocation expects file system path which should not contain URL encoded values.

Closes gh-39655


philwebb:
Thanks very much @slovi! Perfect timing as I managed to squeeze this in just before tomorrows releases.

slovi:
That is great, thanks a lot! 👍

PR #39687, opened 2024-02-22
Add test coverage for ExampleController

LiangJyunKuo:
add testHome() test case under SampleActuatorCustomSecurityApplicationTests

## Before
截圖 2024-02-21 下午9 47 52

## After
截圖 2024-02-21 下午9 46 40

> Format calibration performed


wilkinsona:
Thanks for the proposal. From a security perspective, testInsecureApplicationPath has already checked that an insecure path is accessible so I don't think testing home adds anything. If we were to change anything here I think we could delete the  home method on the controller but I don't think it's really worth it. Thanks anyway.

PR #39688, opened 2024-02-22
foo test coverage for actuator ui somoke test

gdmrw:
## before
截屏2024-02-14 上午4 30 59
## after
FireShot Capture 045 - Coverage Report } SampleActuatorUiApplication -

> Format calibration performed


wilkinsona:
Thanks for the proposal. From a smoke test perspective, the other tests have already covered this scenario so I don't think testing `/foo` really adds anything. If we were to change anything here I think we could delete the `foo` method on the controller but I don't think it's really worth it. Thanks anyway.

Issue #39732, opened 2024-02-22
Make description of virtual threads and JVM exit easier to understand by removing "no-non" double-negative


wilkinsona:
Chatting with @snicoll, the current description of virtual threads and the impact on the JVM exiting is a little hard to understand. It currently reads as follows:

> One side effect of virtual threads is that these threads are daemon threads. A JVM will exit if there are no non-daemon threads.

We think the following may be easier to understand:

> One side effect of virtual threads is that they are daemon threads. A JVM will exit if all of its threads are daemon threads.

PR #39734, opened 2024-02-22
GH-39732 Update description of virtual threads

lukasdo:


wilkinsona:
Thanks for the PR but this is just a copy-paste of what's proposed in the issue. It's easier and quicker for us to apply this sort of change ourselves.

PR #39737, opened 2024-02-23
Make resolution algorithm of ConnectionDetailsFactory more explicit

snicoll:
This commit moves the resolution check for ConnectionDetailsFactory to a dedicated method to make it more clear that it is meant to verify that the implementation is resolved and can be loaded from the classpath.

The previous algorithm relied on a behavior of ResolvableType that only resolves the first level generics. Further improvements in Spring Framework 6.2 make this check invalid as some implementations use a Container that can hold a nested generic.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/service/connection/ConnectionDetailsFactories.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/service/connection/ConnectionDetailsFactories.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/service/connection/ConnectionDetailsFactories.java
```diff
@@ -122,13 +122,29 @@ record Registration<S, D extends ConnectionDetails>(Class<S> sourceType, Class<D
 		@SuppressWarnings("unchecked")
 		private static <S, D extends ConnectionDetails> Registration<S, D> get(ConnectionDetailsFactory<S, D> factory) {
 			ResolvableType type = ResolvableType.forClass(ConnectionDetailsFactory.class, factory.getClass());
-			if (!type.hasUnresolvableGenerics()) {
+			if (isResolved(type)) {
```
snicoll, line 125: Of course, I've just noticed after submitting that `type.resolveGenerics()` is called twice now. Let me know if you want me to refactor.

PR #39752, opened 2024-02-25
Enable netty leak detector

sullis:
Motivation:

detect Netty resource leaks using Netty's paranoid leak detector

Modifications:

- add Netty leak detector junit library
- enable junit extension on NettyReactiveWebServerFactoryTests

Result:

NettyReactiveWebServerFactoryTests will check for netty leaks

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1098,6 +1098,13 @@ bom {
 			]
 		}
 	}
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the proposal. I don't think Boot is the right place for this low-level leak detection and it's better done in the library, Reactor Netty, that's interacting with Netty directly. Reactor Netty [is already tested with leak detection enabled](https://github.com/reactor/reactor-netty/blob/559b5efaa538d3d1157b832cae271cd13129fb45/build.gradle#L263) so I think we're already covered here.

Issue #23988, opened 2020-11-01
Integrate async-profiler in spring boot actuators


JigarJoshi:
To add power to the spring boot actuator, I have been running a Web endpoint that allows developer & tools to profile the JVM and give meaningful representation back (flamegraphs to human and JFR export to tools). Using [async-profiler](https://github.com/jvm-profiling-tools/async-profiler) can add lots of value to Spring boot eco system.

wilkinsona:
Thanks for the suggestion. There's quite a lot to the async profiler, as illustrated by all the options that the profiler.sh script can take. Could you share a bit more about what you've done in your endpoint, what works well for you, etc?

JigarJoshi:
I have hand built their Java library and their linux specific binary and I load the native library and expose actuator endpoint that takes some parameters like event/duration/output format and then render the output.

For Spring boot to use it, first async-profiler has to be published to central maven repository. As a first step towards it I am working on https://github.com/jvm-profiling-tools/async-profiler/issues/93

I will be happy to add this once async-profiler is available in central maven repository.

wilkinsona:
Thanks. Unfortunately, publishing the profiler to Maven Central doesn't seem to be getting much traction. While that's the case, I don't think there's much point in keeping this issue open as we may never be able to act upon it. I'm going to close this one for now. Please let us know by commenting here if the situation changes and there's a clear picture of what would be involved in the proposal. We can then re-open the issue and take another look.

parttimenerd:
I'm going to work on publishing an async-profiler wrapper to maven central in the following days. This will build all async-profiler releases automatically (on the sapmachine.io build infrastructure, on all platforms supported by async-profiler).

wilkinsona:
Thanks, @parttimenerd.

@JigarJoshi, now that there's something on Central, would you like to expand a bit on "I load the native library and expose actuator endpoint that takes some parameters like event/duration/output format and then render the output"?

parttimenerd:
You could do something like done by Mirko Sertic in https://github.com/mirkosertic/flight-recorder-starter. I would recommend running async-profiler with the option `--jfrsync --alloc` anyways, so you could JFR related code. You could even use https://github.com/parttimenerd/jfrtofp to export JFR to the speedscope format to have a more interactive view.

parttimenerd:
I'm happy to help implementation the feature.

andris-rauda:
[...]

PR #39775, opened 2024-02-27
#23988 Integrate async-profiler in spring boot actuators

andris-rauda:
Adds an actuator Web endpoint for invoking async-profiler.

Currently, this actuator is "half baked" (it has a TODO for code duplicate and commented out @ConditionalOnClass annotation), yet it has already proven to be very useful "in field", when used as a custom actuator in our projects. Also, async-profiler has some built in MXBean capabilities, so, probably exposing them as a JMX endpoint would be appropriate.

This pull request is opened in hopes that #23988 could be reopened and someone from Spring Boot team could pick it up and elaborate.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/build.gradle
+++ b/spring-boot-project/spring-boot-actuator/build.gradle
@@ -33,6 +33,7 @@ dependencies {
 	optional("io.undertow:undertow-servlet")
 	optional("javax.cache:cache-api")
 	optional("jakarta.jms:jakarta.jms-api")
+	optional("tools.profiler:async-profiler:2.9")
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the PR, @andris-rauda.

Unfortunately, we still feel that there's too much complexity here and no clear consensus on how the async profiler would be used as an actuator endpoint. Until there's a clear consensus, we don't feel that we can provide built-in support for it in Spring Boot's actuator as it will either be overly complex or unlikely to meet a large of portion of our users' needs.

This feels to me like something that would be better maintained as an external add-on the Spring Boot where it can evolve more rapidly and react to community feedback without the backwards compatibility constraints that comes with being part of Spring Boot itself.

PR #39782, opened 2024-02-28
Replace string concatenation with StringBuilder in a loop

Cespito:
I am aware that this is a Test class, and therefore there is no significant performance justification, but in any case there is no good reason to use string concatenation instead of StringBuilder **in a loop**. It has better performance and is more memory-efficient.


philwebb:
Thanks for the suggestion but I find the existing code easier to read and I don't think we should make this change.

Issue #7562, opened 2016-12-03
Add support for enum default values in configuration processor


qerub:
It's quite common to use enums for configuration properties, but the configuration processor is currently unable to extract the default value for those (because of lack of support in `JavaCompilerFieldValuesParser`). This means that the generated configuration JSON specification is currently incomplete/incorrect wrt. to enum properties. Unless this functionality has been left out on purpose I can take a stab at adding it.

snicoll:
This is something that can be extracted from code and the IDE is supposed to do that for you. I am not keen to write such duplication in the metadata. 

Why are you asking for this? Which IDE do you use?

qerub:
I use IntelliJ IDEA but the wish for this came up outside the context of an IDE when trying to use the configuration meta-data to generate documentation describing all properties and their default values.

qerub:
For example, I was hoping to generate tables with default configuration using `jq`:

```shell
cat **/spring-boot-sample-property-validation/**/spring-configuration-metadata.json | \
jq -r '.properties[] | [.name, .description, .type, .defaultValue | tostring] | @csv' | \
csvfix ascii_table -h "Name, Description, Type, Default value"
```

```
+-------------+--------------+-------------------+---------------+ 
|    Name     | Description  |       Type        | Default value |
+-------------+--------------+-------------------+---------------+
| sample.host | Sample host. | java.lang.String  | null          |
| sample.port | Sample port. | java.lang.Integer | 8080          |
+-------------+--------------+-------------------+---------------+
```

wilkinsona:
It sounds to me like there's some confusion here between listing all possible values for a property with an enum type (which an IDE can do automatically) and listing the default value for such a property. 

In our own metadata I'd expect there to be a default value listed for `security.headers.hsts` but there is not:

```json
{
    "name": "security.headers.hsts",
    "type": "org.springframework.boot.autoconfigure.security.SecurityProperties$Headers$HSTS",
    "description": "HTTP Strict Transport Security (HSTS) mode (none, domain, all).",
    "sourceType": "org.springframework.boot.autoconfigure.security.SecurityProperties$Headers"
},
```

This means that the IDE (I checked in IntelliJ) doesn't display a default value. It does, however, list the three possible values as completion options having extracted them from the enum.

snicoll:
Wow, for some reason my brain read "list of values" instead of default value. Thank you @wilkinsona for pointing that mistake. 

[...]

PR #39797, opened 2024-02-29
Add configuration key "spring.data.web.pageable.serialization-mode" defaults to DIRECT

quaff:
see https://github.com/spring-projects/spring-data-commons/commit/5dd7b322b65652f90717c4f8cb930c5e471ad483

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -31,6 +31,7 @@ import org.springframework.data.web.PageableHandlerMethodArgumentResolver;
 import org.springframework.data.web.config.EnableSpringDataWebSupport;
 import org.springframework.data.web.config.PageableHandlerMethodArgumentResolverCustomizer;
 import org.springframework.data.web.config.SortHandlerMethodArgumentResolverCustomizer;
+import org.springframework.data.web.config.SpringDataWebSettings;
 import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
 
 /**
@@ -42,6 +43,7 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
  *
  * @author Andy Wilkinson
  * @author Vedran Pavic
+ * @author Yanming Zhou
  * @since 1.2.0
  */
 @AutoConfiguration(after = RepositoryRestMvcAutoConfiguration.class)
@@ -77,6 +79,12 @@ public class SpringDataWebAutoConfiguration {
 	@ConditionalOnMissingBean
 	public SortHandlerMethodArgumentResolverCustomizer sortCustomizer() {
 		return (resolver) -> resolver.setSortParameter(this.properties.getSort().getSortParameter());
+	}
+
+	@Bean
+	@ConditionalOnMissingBean
+	public SpringDataWebSettings springDataWebSettings() {
+		return new SpringDataWebSettings(this.properties.getPageable().getSerializationMode());
 	}
 
 }
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2019 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -17,11 +17,13 @@
 package org.springframework.boot.autoconfigure.data.web;
 
 import org.springframework.boot.context.properties.ConfigurationProperties;
+import org.springframework.data.web.config.EnableSpringDataWebSupport.PageSerializationMode;
 
 /**
  * Configuration properties for Spring Data Web.
  *
  * @author Vedran Pavic
+ * @author Yanming Zhou
  * @since 2.0.0
  */
 @ConfigurationProperties("spring.data.web")
@@ -81,6 +83,11 @@ public class SpringDataWebProperties {
 		 */
 		private int maxPageSize = 2000;
 
+		/**
+		 * Configures how to render spring data Pageable instances.
+		 */
+		private PageSerializationMode serializationMode = PageSerializationMode.DIRECT;
+
 		public String getPageParameter() {
 			return this.pageParameter;
 		}
@@ -135,6 +142,14 @@ public class SpringDataWebProperties {
 
 		public void setMaxPageSize(int maxPageSize) {
 			this.maxPageSize = maxPageSize;
+		}
+
+		public PageSerializationMode getSerializationMode() {
+			return this.serializationMode;
+		}
+
+		public void setSerializationMode(PageSerializationMode serializationMode) {
+			this.serializationMode = serializationMode;
 		}
 
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebAutoConfigurationTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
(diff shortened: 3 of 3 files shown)


quaff:
Please hold on until Spring Data Bom 2024.0.0 released.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebProperties.java
```diff
@@ -81,6 +83,19 @@ public static class Pageable {
...
+		/**
+		 * Configures how to render {@link org.springframework.data.domain.PageImpl}
+		 * instances. Defaults to {@link PageSerializationMode#DIRECT} for backward
+		 * compatibility reasons. Prefer explicitly setting this to
+		 * {@link PageSerializationMode#VIA_DTO}, or manually convert
+		 * {@link org.springframework.data.domain.PageImpl} instances before handing them
+		 * out of a controller method, either by manually calling
+		 * {@code new PagedModel<>(page)} or using Spring HATEOAS
+		 * {@link org.springframework.hateoas.PagedModel} abstraction.
+		 * @since 3.3
```
wilkinsona, line 95: We can't use javadoc tags in the comments for a configuration property. Please follow the style of the comments on the other fields in this class.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebProperties.java
```diff
@@ -81,6 +83,19 @@ public static class Pageable {
...
+		 */
+		private PageSerializationMode serializationMode = PageSerializationMode.DIRECT;
```
wilkinsona, line 97: The annotation processor the generates configuration property metadata can't handle enums due to a limitation of the annotation processing model. Please add an entry to `additional-spring-configuration-metadata.json` for this property's default value.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebAutoConfiguration.java
```diff
@@ -79,4 +82,10 @@ public SortHandlerMethodArgumentResolverCustomizer sortCustomizer() {
 		return (resolver) -> resolver.setSortParameter(this.properties.getSort().getSortParameter());
 	}
 
+	@Primary // override bean created by @EnableSpringDataWebSupport
```
wilkinsona, line 85: I think we should provide a better way to do this, perhaps requiring a change in Spring Data, as I'd prefer not to have multiple beans of the same type in the context.

wilkinsona (changes requested):

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebAutoConfiguration.java
quaff: I agree.

wilkinsona: @odrotbohm I have a vague recollection that we've already discussed this but I haven't been able to find it.

I guess would could use `@ConditionalOnProperty` to either auto-configure `@EnableSpringDataWebSupport` or `@EnableSpringDataWebSupport(pageSerializationMode = VIA_DTO)`. I'm a little hesitant as this approach won't scale as more attributes are added but it would work for now I think.

odrotbohm: We ultimately want to nudge people into using `VIA_DTO` at some point, as that's the better way of serializing `Page` instances by default. We kept the default at `DIRECT` for 2024.0.0 to not force applications into the new model, potentially causing them to break their APIs and thus not upgrading. That said, if the Boot team was fine with such a change, we could of course already flip the default, with the users being able to re-instantiate the old behavior by explicitly adding `@EnableSpringDataWebSupport(pageSerializationMode = DIRECT)` to their application.

quaff: `@ConditionalOnProperty` is not appropriate here, It's just a flag that should be stick to other pageable settings as possible, not like `spring.data.redis.client-type` which affect creation of other beans.
I'd prefer to not register `SpringDataWebSettings` bean if `pageSerializationMode` is default as `SpringDataWebAutoConfiguration` did, then the `@Primary` could be removed.

wilkinsona:
A labelled this as blocked but neglected to say why. We're waiting for the outcome of https://github.com/spring-projects/spring-data-commons/pull/3054.

odrotbohm:
It could be that I am missing something here, but the presence of a `SpringDataWebSettings` is an indication of a user-provided `@EnableSpringDataWebSupport` annotation. Shouldn't it be possible to simply guard the creation of the `SpringDataWebSettings` bean here with the presence of one of the `ImportBeanDefinitionRegistrars` the annotation registers and of course the lack of a `SDWS` bean definition itself?

It feels a bit weird to, if the annotation is configured, not forward its configuration default value into downstream configuration setup. That extending into having to write that latter code re-implementing a default that's already set and conveyed through the annotation attribute's value.

wilkinsona:
Thanks, @odrotbohm.

> Adding a configuration property to flip that switch might be nice, but at the same time, I wonder whether having to use the annotation isn't just enough?

As things stand, if you use the annotation, you'll lose our customization of paging and sorting.

> Shouldn't it be possible to simply guard the creation of the `SpringDataWebSettings` bean here with the presence of one of the ImportBeanDefinitionRegistrars the annotation registers.

We already use `@EnableSpringDataWebSupport` in our auto-configuration when we detect that the user hasn't already used it. Unfortunately, `SpringDataWebSettingsRegistar` then always registers a `SpringDataWebSettings` that we can't (easily) replace. We can work around this with two different classes (ignoring fuzzier matching of `direct` and `via-dto`:

```java
	@EnableSpringDataWebSupport
	@ConditionalOnProperty(name = "spring.data.web.pageable.serialization-mode", havingValue = "direct",
			matchIfMissing = true)
	class EnableSpringDataWebSupportConfiguration {

	}

	@EnableSpringDataWebSupport(pageSerializationMode = PageSerializationMode.VIA_DTO)
	@ConditionalOnProperty(name = "spring.data.web.pageable.serialization-mode", havingValue = "via-dto",
			matchIfMissing = false)
	class EnableCustomizedSpringDataWebSupportConfiguration {

	}
```

https://github.com/spring-projects/spring-data-commons/pull/3054 would remove the need for this little bit of complexity in Boot.

snicoll:
@odrotbohm could you please review the comment above when you get a chance? Thanks!

odrotbohm:
I'm sorry this took so long to get to. I've merged @quaff's contribution and ported it back into the 3.3.x branch. We now do not register a `SpringDataWebSettings` bean anymore, unless a non-default value for `pageSerializationMode` is set. Hope this eases the situation for Boot. Please let me know, if there's anything else we can do.

quaff:
`@Primary` removed now.

spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -1125,6 +1125,12 @@
         "level": "error"
       }
     },
+    {
+      "name": "spring.data.web.pageable.serialization-mode",
+      "type": "org.springframework.data.web.config.EnableSpringDataWebSupport.PageSerializationMode",
+      "description": "Configures how to render spring data Pageable instances.",
+      "defaultValue": "DIRECT"
+    },
```
wilkinsona, line 1133: We can remove this when merging thanks to https://github.com/spring-projects/spring-boot/issues/7562.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebAutoConfiguration.java
```diff
@@ -79,4 +81,9 @@ public SortHandlerMethodArgumentResolverCustomizer sortCustomizer() {
 		return (resolver) -> resolver.setSortParameter(this.properties.getSort().getSortParameter());
 	}
 
+	@Bean
+	public SpringDataWebSettings springDataWebSettings() {
```
wilkinsona, line 85: We should make this `@ConditionalOnMissingBean` and add a test for that behavior.

spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
quaff: done.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/web/SpringDataWebAutoConfiguration.java
quaff: done.

wilkinsona (approved):

wilkinsona:
Thank you, @quaff.

PR #39800, opened 2024-02-29
Support virtual threads for Jackson pool.

### LINKED ISSUE #39783: Configure a suitable Jackson object pool when virtual threads are enabled
The default pool uses `ThreadLocal`s, which aren't a great fit when using virtual threads.

See https://twitter.com/jkuipers/status/1762505881000808516

youssef3wi:
Closes #39783


mhalbritter:
Thanks for the PR. We've updated to Jackson 2.17 in the meantime, which [by default uses a lock free pool](https://github.com/spring-projects/spring-boot/issues/39783#issuecomment-2005123961), which makes this PR obsolete. Thanks again!

Issue #29193, opened 2021-12-31
Configure the auto-configured LocalContainerEntityManagerFactoryBean with any ManagedClassNameFilter beans


hannah23280:
Hi,

   I would like to request for `@EntityScan` to be able to provide the ability to scan for specific entities class, rather than based on package.  E.g `@EntityScan(classes={"com.aa.ClassA", "com.aa.ClassB"))` will only scan for these 2 classes and other classes in the same package will NOT be scan.  
Searching internet shows many people asking for such capabilities and end up  having to write some codes to circumvent this.  
Adding such capabilities made life easy for developers

hanct:
> Hi,
> 
> I would like to request for @EntityScan to be able to provide the ability to scan for specific entities class, rather than based on package. E.g @EntityScan(classes={"com.aa.ClassA", "com.aa.ClassB")) will only scan for these 2 classes and other classes in the same package will NOT be scan. Searching internet shows many people asking for such capabilities and end up having to write some codes to circumvent this. Adding such capabilities made life easy for developers

In addition, should also provide similar way to configure this in properties/yaml as well.
spring.jpa.entity-scan=com.aa.ClassA,com.aa.ClassB

wilkinsona:
Thanks for the suggestion.  Entity scanning is heavily package-based at the moment, particularly for JPA where Spring Framework's `LocalContainerEntityManagerFactoryBean ` and `DefaultPersistentUnitManager` is used. As such, for Spring Boot to be able to offer more control over the entities that are found by scanning, we'd need some changes in Spring Framework. If this is something that you'd like to pursue, please open a Spring Framework issue. If the Framework team add filtering support or similar we can then revisit this issue and consider exposing that new capability via `@EntityScan`.

In the meantime, assuming that you're using JPA, you may be able to achieve a custom property-based solution by providing your own `PersistentUnitManager` implementation.

wilkinsona:
@snicoll has [prototyped](https://github.com/snicoll/spring-framework/commit/add5d98f9d9a1331e3fb98efce4c9c0d8b8bfd39) something for this. We should take a look at it and see what we could now do with `@EntityScan`.

wilkinsona:
[...]

PR #39813, opened 2024-03-01
Configure JpaBaseConfiguration with custom ManagedClassNameFilter

quaff:
Fix GH-29193

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/JpaBaseConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/JpaBaseConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -46,6 +46,7 @@ import org.springframework.core.io.ResourceLoader;
 import org.springframework.orm.jpa.JpaTransactionManager;
 import org.springframework.orm.jpa.JpaVendorAdapter;
 import org.springframework.orm.jpa.LocalContainerEntityManagerFactoryBean;
+import org.springframework.orm.jpa.persistenceunit.ManagedClassNameFilter;
 import org.springframework.orm.jpa.persistenceunit.PersistenceManagedTypes;
 import org.springframework.orm.jpa.persistenceunit.PersistenceManagedTypesScanner;
 import org.springframework.orm.jpa.persistenceunit.PersistenceUnitManager;
@@ -69,6 +70,7 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
  * @author Andy Wilkinson
  * @author Kazuki Shimizu
  * @author Eddú Meléndez
+ * @author Yanming Zhou
  * @since 1.0.0
  */
 @Configuration(proxyBeanMethods = false)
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/JpaBaseConfiguration.java
```diff
@@ -195,9 +197,11 @@ static class PersistenceManagedTypesConfiguration {
 		@Bean
 		@Primary
 		@ConditionalOnMissingBean
-		static PersistenceManagedTypes persistenceManagedTypes(BeanFactory beanFactory, ResourceLoader resourceLoader) {
+		static PersistenceManagedTypes persistenceManagedTypes(BeanFactory beanFactory, ResourceLoader resourceLoader,
+				ObjectProvider<ManagedClassNameFilter> managedClassNameFilter) {
 			String[] packagesToScan = getPackagesToScan(beanFactory);
-			return new PersistenceManagedTypesScanner(resourceLoader).scan(packagesToScan);
+			return new PersistenceManagedTypesScanner(resourceLoader, managedClassNameFilter.getIfAvailable())
```
wilkinsona, line 203: I wonder if we should support multiple filters and create a composite. For consistency with exposing actuator endpoints, exclude would take precedence over include. In other words, a class name would have to be matched by every filter for it to be included.

quaff: > a class name would have to be matched by every filter for it to be included.

It's unintuitive.

wilkinsona: It's hard to continue a discussion as you haven't described why you think that it's unintuitive. As I said above, this is how the actuator endpoints behave and that seems to be serving us well. I think the consistency would be a good thing.

quaff: My apologize, as typical spring framework user, I’m expecting one bean available for injection, If multiple beans need to be composed, application should write a primary composite implementation they can take over the control, not hardcoded AND or OR. 
l said unintuitive because I don’t know If you actuator endpoints doing that way if you don’t mention it.

wilkinsona: @snicoll do you have an opinion here? Do you think there's likely to be a need for multiple filters that we compose automatically or will most users' needs be served by supporting only one. I think I'm now leaning towards only handling one in the auto-configuration.

snicoll: Sorry, I read that and mean to reply and I forgot about it. I honestly don't know but I prefer that the decision is in a central place. If users need to have multiple filters for whatever reason, I'd prefer them to do the gathering rather than us.

philwebb: Sounds like we should stick with a single bean for now. Perhaps some helper factory methods can be added `ManagedClassNameFilter` if composing filters turns out to be a common need? E.g. `ManagedClassNameFilter.allOf(f1,f2,f3)`.

PR #39819, opened 2024-03-01
spring.data.redis.cluster.nodes and spring.data.redis.sentinel.nodes do not handle IPv6 addresses correctly

tobi-laa:
When specifying colon-separated pairs of host and port with IPv6 addresses (i.e. `[::1]:6379`) for either `spring.data.redis.cluster.nodes` or `spring.data.redis.sentinel.nodes`, the parsing of these nodes/addresses was broken as the splitting of these strings was not performed explicitly on the last found colon. The behavior is fixed with this PR.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/PropertiesRedisConnectionDetails.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/PropertiesRedisConnectionDetails.java
@@ -113,9 +113,9 @@ class PropertiesRedisConnectionDetails implements RedisConnectionDetails {
 	}
 
 	private Node asNode(String node) {
-		var posLastColon = node.lastIndexOf(':');
-		var host = node.substring(0, posLastColon).replaceFirst("^\\[(.+)]$", "$1");
-		var port = Integer.parseInt(node.substring(posLastColon + 1));
+		int portSeparatorIndex = node.lastIndexOf(':');
+		String host = node.substring(0, portSeparatorIndex);
+		int port = Integer.parseInt(node.substring(portSeparatorIndex + 1));
 		return new Node(host, port);
 	}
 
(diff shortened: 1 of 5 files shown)


tobi-laa:
Fixed formatting.

tobi-laa:
Fixed Checkstyle violation. Sorry, should have fixed that from the start.

mhalbritter:
If you run `./gradlew build`, you get those errors on your local machine instead of waiting for the CI.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/PropertiesRedisConnectionDetails.java
```diff
@@ -113,8 +113,10 @@ public Cluster getCluster() {
 	}
 
 	private Node asNode(String node) {
-		String[] components = node.split(":");
-		return new Node(components[0], Integer.parseInt(components[1]));
+		var posLastColon = node.lastIndexOf(':');
+		var host = node.substring(0, posLastColon).replaceFirst("^\\[(.+)]$", "$1");
```
wilkinsona, line 117: I don't think the `[` and `]` should be removed. They can be useful in situations where the `:` characters in an address can cause problems. We should trust the user to have used them in the property's value intentionally.

I think we can just use code that's very similar to `PropertiesRabbitConnectionDetails`:

https://github.com/spring-projects/spring-boot/blob/9a4d0fb8b62aac303e9879f625cf4e7a4cf43cbc/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/amqp/PropertiesRabbitConnectionDetails.java#L56-L59

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/PropertiesRedisConnectionDetails.java
```diff
@@ -113,8 +113,10 @@ public Cluster getCluster() {
 	}
 
 	private Node asNode(String node) {
-		String[] components = node.split(":");
-		return new Node(components[0], Integer.parseInt(components[1]));
+		var posLastColon = node.lastIndexOf(':');
```
wilkinsona, line 116: We don't use `var` in Spring Boot's codebase.

wilkinsona (changes requested):

tobi-laa:
@wilkinsona Thank you for the review! I updated the PR accordingly.

wilkinsona:
@tobi-laa thanks very much for making your first contribution to Spring Boot.

PR #39837, opened 2024-03-05
When using Maven, configuring the spring-boot.excludes or spring-boot-includes user properties causes the build to fail with "Cannot find default setter"

mstahv:
The change makes excludes configurable via CLI.

For example this:

    mvn install -Dspring-boot.excludes=foo:bar,foo:baz

Would equal to following pom.xml configuration:

    <excludes>
        <exclude>
            <groupId>foo</groupId>
            <artifactId>bar</artifactId>
        </exclude>
        <exclude>
            <groupId>foo</groupId>
            <artifactId>baz</artifactId>
        </exclude>
    </excludes>

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/JarIntegrationTests.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/JarIntegrationTests.java
@@ -195,8 +195,23 @@ class JarIntegrationTests extends AbstractArchiveIntegrationTests {
 				.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-context")
 				.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-core")
 				.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-jcl")
-				.doesNotHaveEntryWithName("BOOT-INF/lib/servlet-api-2.5.jar");
+				.doesNotHaveEntryWithNameStartingWith("BOOT-INF/lib/jakarta.servlet-api-");
 		});
+	}
+
+	@TestTemplate
+	void whenAnEntryIsExcludedWithPropertyItDoesNotAppearInTheRepackagedJar(MavenBuild mavenBuild) {
+		mavenBuild.project("jar")
+			.systemProperty("spring-boot.excludes", "jakarta.servlet:jakarta.servlet-api")
+			.goals("install")
+			.execute((project) -> {
+				File repackaged = new File(project, "target/jar-0.0.1.BUILD-SNAPSHOT.jar");
+				assertThat(jar(repackaged)).hasEntryWithNameStartingWith("BOOT-INF/classes/")
+					.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-context")
+					.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-core")
+					.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-jcl")
+					.doesNotHaveEntryWithNameStartingWith("BOOT-INF/lib/jakarta.servlet-api-");
+			});
 	}
 
 	@TestTemplate
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/AbstractDependencyFilterMojo.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/AbstractDependencyFilterMojo.java
@@ -82,7 +82,8 @@ public abstract class AbstractDependencyFilterMojo extends AbstractMojo {
 	/**
 	 * Collection of artifact definitions to exclude. The {@link Exclude} element defines
 	 * mandatory {@code groupId} and {@code artifactId} properties and an optional
-	 * {@code classifier} property.
+	 * {@code classifier} property. If passing in excludes as a property the syntax is
+	 * <pre>-Dspring-boot.excludes=groupId1:artifactId1,groupId2:artifactId2:optional-qualifier</pre>
 	 * @since 1.1.0
 	 */
 	@Parameter(property = "spring-boot.excludes")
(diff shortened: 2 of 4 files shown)


wilkinsona:
I think this is a bug. We already expose the excludes as a property:

https://github.com/spring-projects/spring-boot/blob/2c4fb5baaa5af931730b610ecfee2645586ba0f1/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/AbstractDependencyFilterMojo.java#L88

It's [included in the documentation](https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#packaging.repackage-goal.parameter-details.excludes) as well but it doesn't work:

```
$ mvn clean package -Dspring-boot.excludes=com.example:example-module
…
[INFO] --- spring-boot:2.7.18:repackage (repackage) @ example ---
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time:  4.886 s
[INFO] Finished at: 2024-03-06T09:58:08Z
[INFO] ------------------------------------------------------------------------
[ERROR] Failed to execute goal org.springframework.boot:spring-boot-maven-plugin:2.7.18:repackage (repackage) on project test-spring-boot-2.7: Unable to parse configuration of mojo org.springframework.boot:spring-boot-maven-plugin:2.7.18:repackage for parameter #: Cannot find default setter in class org.springframework.boot.maven.Exclude -> [Help 1]
[ERROR] 
[ERROR] To see the full stack trace of the errors, re-run Maven with the -e switch.
[ERROR] Re-run Maven using the -X switch to enable full debug logging.
[ERROR] 
[ERROR] For more information about the errors and possible solutions, please read the following articles:
[ERROR] [Help 1] http://cwiki.apache.org/confluence/display/MAVEN/PluginConfigurationException
```

mstahv:
Now that it works (after this change) and I have the branch open in my NetBeans, should I also add an example of the property syntax to the doc 🤔 Or is it too obvious?

mstahv:
Additional note: the string syntax also doesn't in this change yet support classifier. Not sure if very often needed though 🤷‍♂️

wilkinsona:
Yes, please. An example in the docs would be good. I think we should also assert the format in the setter with something like this:

```
Assert.isTrue(parts.length == 2, "Exclude must be in the form groupId:artifactId");
```

Some tests would be very welcome too please.

wilkinsona:
> the string syntax also doesn't in this change yet support classifier

That's a good point. I think classifiers should be supported otherwise the property doesn't meet its goal as provided an alternative to XML-based configuration. My proposed assertion above would have to be updated to take this into account.

mstahv:
Yeah, I think at least some integration test at least. @snicoll just hinted me to the correct directory 👍

mstahv:
✅ Discussed changes in. Also fixed a related IT that had been false positive since Spring Boot 3.

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/JarIntegrationTests.java
```diff
@@ -195,10 +195,25 @@ void whenAnEntryIsExcludedItDoesNotAppearInTheRepackagedJar(MavenBuild mavenBuil
...
+		mavenBuild.project("jar")
+			.systemProperty("spring-boot.excludes", "jakarta.servlet:jakarta.servlet-api")
+			.goals("install")
+			.execute((project) -> {
+				File repackaged = new File(project, "target/jar-0.0.1.BUILD-SNAPSHOT.jar");
+				assertThat(jar(repackaged)).hasEntryWithNameStartingWith("BOOT-INF/classes/")
+					.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-context")
+					.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-core")
+					.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-jcl")
+					.doesNotHaveEntryWithNameStartingWith("BOOT-INF/lib/jakarta.servlet-api-");
```
therealaleko, line 213: Is the last "-" on purpose ?

therealaleko: Ah seems needed for classifier, sorry missed that.

mstahv:
@wilkinsona Was there still something missing from this PR. Would love to get forward with my prototype that would require this, but wouldn't like to work with own snapshots (testing would be deployments to could environments that'd build my apps with buildpacks...).

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/JarIntegrationTests.java
mstahv: More like making the assert as long as possible. The version number comes next ofter the last dash.

wilkinsona:
No, nothing's missing from the PR, we just have a lot on at the moment. We'll get to this as soon as we can. Thanks for your patience in the meantime.

spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/JarIntegrationTests.java
```diff
@@ -195,10 +195,25 @@ void whenAnEntryIsExcludedItDoesNotAppearInTheRepackagedJar(MavenBuild mavenBuil
 				.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-context")
 				.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-core")
 				.hasEntryWithNameStartingWith("BOOT-INF/lib/spring-jcl")
-				.doesNotHaveEntryWithName("BOOT-INF/lib/servlet-api-2.5.jar");
+				.doesNotHaveEntryWithNameStartingWith("BOOT-INF/lib/jakarta.servlet-api-");
```
wilkinsona, line 198: This change could result in the test passing when it should not. The exclude is targeting this dependency:

https://github.com/spring-projects/spring-boot/blob/d642d3554749d4a368f62b911f6d6d588fa2ffeb/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/projects/jar-exclude-entry/pom.xml#L43-L48

wilkinsona: I'll revert it when merging.

wilkinsona:
Thanks, @mstahv. While merging this, I realised that includes had the same problem. I've fixed those too in https://github.com/spring-projects/spring-boot/commit/baf5a7f13023d937a9d2bed973f297368a02a99d.

PR #39842, opened 2024-03-06
Refactor Unnecessary return code modification

taeseokyang:
It seems like removing the return and moving the bottom code would make the code much easier to understand.


taeseokyang:
sorry I thought wrong.. haha

Issue #28213, opened 2021-10-06
Remove support for Java's deprecated SecurityManager


wilkinsona:
Following [Framework's lead](https://github.com/spring-projects/spring-framework/issues/26901), we should remove support for `SecurityManager` and use of related classes such as `AccessController`.

PR #39856, opened 2024-03-08
JarUrlConnection.getPermission() can throw NullPointerException if jarFileConnection is null

yokotaso:
Thank you for developing great springframework and spring-boot ecosystem, I appreciate development.

if given resource path is not found and JavaSecurityManager is enabled, NullPointerException is thrown internally.
Not only it is simply meaningless exception,
but also it might be cause of performance regression.

StackTrace is like bellow. 

```
java.lang.NullPointerException: Cannot invoke "java.net.URLConnection.getPermission()" because "this.jarFileConnection" is null
        at org.springframework.boot.loader.net.protocol.jar.JarUrlConnection.getPermission(JarUrlConnection.java:175)
        at java.base/jdk.internal.loader.URLClassPath.check(URLClassPath.java:553)
        at java.base/jdk.internal.loader.URLClassPath$Loader.findResource(URLClassPath.java:612)
        at java.base/jdk.internal.loader.URLClassPath.findResource(URLClassPath.java:296)
        at java.base/java.net.URLClassLoader$2.run(URLClassLoader.java:629)
        at java.base/java.net.URLClassLoader$2.run(URLClassLoader.java:627)
        at java.base/java.security.AccessController.doPrivileged(AccessController.java:399)
        at java.base/java.net.URLClassLoader.findResource(URLClassLoader.java:626)
        at org.springframework.boot.loader.net.protocol.jar.JarUrlClassLoader.findResource(JarUrlClassLoader.java:70)
        at java.base/java.lang.ClassLoader.getResource(ClassLoader.java:1403)
        at java.base/java.net.URLClassLoader.getResourceAsStream(URLClassLoader.java:290)
        at java.base/java.lang.Class.getResourceAsStream(Class.java:2850)
```

I got above stacktrace with using debugger around `catch` statement in  `jdk.internal.loader.URLClassPath.Loader#findResource` and invoking `Exception#printStackTrace()`

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/net/protocol/jar/JarUrlConnection.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/net/protocol/jar/JarUrlConnection.java
@@ -172,7 +172,7 @@ final class JarUrlConnection extends java.net.JarURLConnection {
 
 	@Override
 	public Permission getPermission() throws IOException {
-		return this.jarFileConnection.getPermission();
+		return (this.jarFileConnection != null) ? this.jarFileConnection.getPermission() : null;
 	}
 
 	@Override


wilkinsona:
Thanks for the proposal but, following its deprecation in the JDK, we don't support running with the security manager enabled. Support was removed in https://github.com/spring-projects/spring-boot/issues/28213.

philwebb:
Looking again at this one I think we probably should merge it. Although we don't support running with the security manager the `getPermission()` method hasn't been deprecated and we do protect other call that delegate to `this.jarFileConnection` to make sure it's not null.

The only thing I'm not sure about is if we can return null. The class loader implementation used to throw an exception. We could do that or return `SecurityConstants.ALL_PERMISSION;`.

philwebb:
Throwing an exception looks like the best option.

yokotaso:
@philwebb Thank you for seeing my suggestion. Sorry for the delay in noticing.

according to javadoc in `java.net.URLConnection#getPermission`,
it seems ok that `java.net.URLConnection#getPermission` returns null

```
Returns a permission object representing the permission
necessary to make the connection represented by this
object. This method returns null if no permission is
required to make the connection.
```

philwebb:
Thanks again @yokotaso. This has now been merged.

PR #39859, opened 2024-03-08
Set all dispatch types for OncePerRequestFilter

fisco-unimatic:
As I said in a Stack Overflow question that I posted yesterday, [Spring SessionRepositoryFilter is not applied to forwarded requests](https://stackoverflow.com/questions/78124343/spring-sessionrepositoryfilter-is-not-applied-to-forwarded-requests). 
When there's an error, then the request is forwarded to _/error_, with a dispatch type of `ERROR`, but `SessionRepositoryFilter` is only applied to requests with dispatch type `REQUEST`. That means that a new session ID is created by tomcat, and the client uses this on subsequent requests, which are then rejected as unauthorized by the session repository.
`AbstractFilterRegistrationBean#determineDispatcherTypes` sets all dispatch types for filters that extend `org.springframework.web.filter.OncePerRequestFilter` in `spring-web`, but `SessionRepositoryFilter` extends `org.springframework.session.web.http.OncePerRequestFilter` in `spring-session-core`.
This change sets all dispatch types for filters extending either `OncePerRequestFilter` class.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/AbstractFilterRegistrationBean.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/AbstractFilterRegistrationBean.java
@@ -21,6 +21,7 @@ import java.util.Collection;
 import java.util.Collections;
 import java.util.EnumSet;
 import java.util.LinkedHashSet;
+import java.util.Optional;
 import java.util.Set;
 
 import jakarta.servlet.DispatcherType;
@@ -30,9 +31,7 @@ import jakarta.servlet.FilterRegistration.Dynamic;
 import jakarta.servlet.ServletContext;
 
 import org.springframework.util.Assert;
-import org.springframework.util.ClassUtils;
 import org.springframework.util.StringUtils;
-import org.springframework.web.filter.OncePerRequestFilter;
 
 /**
  * Abstract base {@link ServletContextInitializer} to register {@link Filter}s in a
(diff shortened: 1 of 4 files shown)


quaff:
It's very risky to change the behavior of `org.springframework.web.filter.OncePerRequestFilter`.

wilkinsona:
> It's very risky to change the behavior of org.springframework.web.filter.OncePerRequestFilter.

Unless I have missed something, this PR doesn't do that. Instead, it's treating `org.springframework.session.web.http.OncePerRequestFilter` in the same way as we already treat `org.springframework.web.filter.OncePerRequestFilter`.

That said, I'm still not sure that this is something that we should do. I would prefer that it's fixed in Spring Session by it using `org.springframework.web.filter.OncePerRequestFilter` instead of `org.springframework.session.web.http.OncePerRequestFilter`, although we may still need to do something in the meantime. I've raised this with the Spring Session team and we're waiting to hear back from them.

fisco-unimatic:
ok - I'll try to put together an example and then raise an issue

fisco-unimatic:
In case any ends up here looking for a solution to a similar problem, it turns out that our application was excluding `SessionAutoConfiguration`. If I remove that exclusion, then it imports `SessionRepositoryFilterConfiguration`, which sets the dispatch types.

If for some reason you wanted to exclude `SessionAutoConfiguration`, you could instead make your session configuration class extend `AbstractHttpSessionApplicationInitializer`. That would also set the dispatch types for you.

Issue #39860, opened 2024-03-08
Deprecate @MockBean and @SpyBean


philwebb:
With https://github.com/spring-projects/spring-framework/issues/29917 fixed we should consider deprecating our versions.

wanger26:
@philwebb do we want to target release 2.3.0 with this change? Also are we thinking of marking those for removal at this time?

philwebb:
We'll probably look at this in Spring Boot 3.4.0 which will be the first release to build on Spring Framework 6.2. I'm not sure yet about when well mark the existing annotations for removal.

wanger26:
Okay sounds good. I can take this one on. Do we want to take this opportunity to migrate our own usages of those two annotations over to the new spring framework ones?

philwebb:
Closing in favor of PR #39864 which we'll merge sometime after we create the 3.3.x branch. Thanks @wanger26!

OrangeDog:
What is the replacement for `@MockBean` on a type? For example, this was very useful:

```java
@TestConfiguration
@Import({ /* some config shared by tests */ })
@MockBean({ /* some beans that need to be mocked in shared tests */ })
public class SharedWebTestConfig {
    // ...
}

@WebMvcTest(SomeController.class)
@Import(SharedWebTestConfig.class)
public class SomeTetsts {
    // ...
}
```

Now the only solution is to make the mock beans manually?

```java
@TestConfiguration
@Import({ /* some config shared by tests */ })
public class SharedWebTestConfig {
    @Bean
    public BeanClass myBean() {
        return Mockito.mock(BeanClass.class);
    }
    // ...
}
```

wilkinsona:
I don't believe Framework has a direct equivalent although you don't have to resort to declaring the mock beans manually. You can use `@MockitoBean` on fields in a test class, its type hierarchy and, if it's a nested test class, its enclosing class hierarchy so you could move your mocks into a class that `SomeTests` extends. If you'd like to see `@MockitoBean` support something beyond this, please open a Framework issue.

sbrannen:
What Andy said is correct: Spring Framework does not support `@MockitoBean` at the type-level.

See also https://github.com/spring-projects/spring-framework/issues/29917#issuecomment-2428703548.

nmck257:
fyi @OrangeDog, other readers -- I've opened a standalone issue in Spring Framework for this topic:
https://github.com/spring-projects/spring-framework/issues/33925

PR #39864, opened 2024-03-10
Deprecate @MockBean and @SpyBean

wanger26:
### Change Summary
- Deprecate Deprecate @MockBean and @SpyBean in favor of Spring Framekworks 6.2.0 @MockitoBean and @MockitoSpy. See https://github.com/spring-projects/spring-framework/commit/e1bbdf09139dca7c21ec64e140bcc3bda463b2f6
- Migrate usages of @MockBean and @SpyBean to @MockitoBean an d @MockitoSpy

### Issue
- https://github.com/spring-projects/spring-boot/issues/39860

### DIFF
--- a/buildSrc/build.gradle
+++ b/buildSrc/build.gradle
@@ -22,7 +22,7 @@ new File(projectDir.parentFile, "gradle.properties").withInputStream {
 	}
 }
 versions["jackson"] = "2.15.3"
-versions["springFramework"] = "6.0.12"
+versions["springFramework"] = "6.2.0-SNAPSHOT"
 ext.set("versions", versions)
 if (versions.springFramework.contains("-")) {
 	repositories {
--- a/gradle.properties
+++ b/gradle.properties
@@ -12,7 +12,7 @@ junitJupiterVersion=5.10.2
 kotlinVersion=1.9.22
 mavenVersion=3.9.4
 nativeBuildToolsVersion=0.10.1
-springFrameworkVersion=6.1.4
+springFrameworkVersion=6.2.0-SNAPSHOT
 tomcatVersion=10.1.19
 
 kotlin.stdlib.default.dependency=false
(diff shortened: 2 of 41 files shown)


snicoll:
FTR, I am maintaining a branch of the changes require to upgrade to SF `6.2.0-SNAPSHOT` and some of the changes in this PR are already applied there, which I intend to submit once Spring Boot is ready to build against it.

wanger26:
Closing PR /draft based on comments above.

philwebb:
@wanger26 Can we keep this one? It's likely that @snicoll won't migrate the existing tests and the work you've done here looks valuable for us.

wanger26:
Yes of course, that is no problem 👍

snicoll:
Here is what's needed so far to get a green build with Spring Framework `6.2.0-SNAPSHOT`: https://github.com/snicoll/spring-boot/tree/compat-6.2.x

Regarding Mockito support, I had to remove the automatic intialization of mocks as the support in the core framework does that now and they were trumping each other. Given that the support is going to be deprecated, tests should still use the existing infrastructure, rather than moving to the new annotations. It's a different story for smoke tests & co though.

wanger26:
> Here is what's needed so far to get a green build with Spring Framework `6.2.0-SNAPSHOT`: https://github.com/snicoll/spring-boot/tree/compat-6.2.x
> 
> Regarding Mockito support, I had to remove the automatic intialization of mocks as the support in the core framework does that now and they were trumping each other. Given that the support is going to be deprecated, tests should still use the existing infrastructure, rather than moving to the new annotations. It's a different story for smoke tests & co though.

Should we leave all tests using the Mockito outside of smoke tests and co untouched, or should we only leave the unit tests for the Mockito support untouched?

wilkinsona:
Thank you, @wanger26.

PR #38631, opened 2023-12-01
Support user-defined ExecutionContextSerializer in BatchAutoConfiguation

benelog:
Since Spring Batch 5, it's possible to easily specify an ExecutionContextSerializer as shown below:

```java
@Configuration
@EnableBatchProcessing(executionContextSerializerRef = "myCustomSerializer")
public class MyJobConfigWithCustomSerializer {
    @Bean
    public ExecutionContextSerializer myCustomSerializer() {
        Jackson2ExecutionContextStringSerializer serializer = new Jackson2ExecutionContextStringSerializer();
        // customize serializer
        return serializer;
    }
}
```

However, when using BatchAutoConfiguration in Spring Boot, there isn’t a straightforward way to achieve this.

This pull request modifies the BatchAutoConfiguration to refer to an ExecutionContextSerializer registered in the ApplicationContext.

The behavior of the ExecutionContextSerializer has changed in Spring Batch 5.0. ( https://github.com/spring-projects/spring-batch/issues/4122 )
So, if this feature is available, I expect it to be more helpful for those who prefer the previous version of the policy to upgrade to the latest version.

I believe the changes in this PR represent a user-expected behavior. 
In Spring Batch, DefaultJobParametersConverter is used as a similar policy. 
The class is used by default, but if there is another implementation class that the user has registered in ApplicationContext, it is replaced.

PR #39911, opened 2024-03-12
Delegate getExecutionContextSerializer() to DefaultBatchConfiguration if not provided

izeye:
Like other delegations in `SpringBootBatchConfiguration`, it would be better to delegate `getExecutionContextSerializer()` to `DefaultBatchConfiguration` if a custom one is not provided rather than create one based on the current implementation of `DefaultBatchConfiguration.getExecutionContextSerializer()`.

See gh-38328


scottfrederick:
Good catch @izeye, thanks!

benelog:
With a similar intention, I revised it at https://github.com/spring-projects/spring-boot/pull/38631/files

```java
this.serializer = executionContextSerializer.getIfAvailable(super::getExecutionContextSerializer);
```

PR #39912, opened 2024-03-12
Re-enable Spring Pulsar interceptor tests

onobc:
The PulsarTemplate recently replaced its list of ProducerInterceptors with a list of ProducerBuilderCustomizers that customize the builder by adding each interceptor to the builder. The PulsarAutoConfigurationTests previosuly relied on the previous field. This commit adjusts the tests to instead use the Customizers testing utility to verify the interceptors.

cc: @wilkinsona

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfigurationTests.java
@@ -27,8 +27,6 @@ import org.apache.pulsar.client.api.PulsarClient;
 import org.apache.pulsar.client.api.ReaderBuilder;
 import org.apache.pulsar.client.api.interceptor.ProducerInterceptor;
 import org.apache.pulsar.common.schema.SchemaType;
-import org.assertj.core.api.InstanceOfAssertFactories;
-import org.junit.jupiter.api.Disabled;
 import org.junit.jupiter.api.Nested;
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarAutoConfigurationTests.java
```diff
@@ -285,24 +283,29 @@ void injectsExpectedBeans() {
...
+		<T> void whenHasUseDefinedProducerInterceptorInjectsBean() {
 			ProducerInterceptor interceptor = mock(ProducerInterceptor.class);
 			this.contextRunner.withBean("customProducerInterceptor", ProducerInterceptor.class, () -> interceptor)
-				.run((context) -> assertThat(context).getBean(PulsarTemplate.class)
-					.extracting("interceptorsCustomizers")
-					.asInstanceOf(InstanceOfAssertFactories.LIST)
-					.contains(interceptor));
+				.run((context) -> {
+					PulsarTemplate<?> pulsarTemplate = context.getBean(PulsarTemplate.class);
+					Customizers<ProducerBuilderCustomizer<T>, ProducerBuilder<T>> customizers = Customizers
```
onobc, line 291: @philwebb added this `Customizers` testing utility to ease the burden of verifying customizers. Basically because they are lambdas we have to execute them and see if they "customized" what we expected them to customizer (and in order). In this case, we know our producer builder customizers are going to call `ProducerBuilder.intercept` with the specified interceptor.

mhalbritter:
Thank you!

Issue #39839, opened 2024-03-05
Add public getUndertow method to UndertowWebServer


isaacirvin:
My team is currently evaluating the embedded Undertow within Spring Boot.  One concern that has come up which is that it does not appear as though Undertow exposes its internal metrics (active connections, active requests, etc.) via the MeterRegistry.  

Additionally, the [UndertowWebServer](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/undertow/UndertowWebServer.java) class does not expose the underlying Undertow server object in the same way that [TomcatWebServer](https://github.com/spring-projects/spring-boot/blob/2c4fb5baaa5af931730b610ecfee2645586ba0f1/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/tomcat/TomcatWebServer.java#L440) and [JettyWebServer](https://github.com/spring-projects/spring-boot/blob/2c4fb5baaa5af931730b610ecfee2645586ba0f1/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/embedded/jetty/JettyWebServer.java#L315) expose their underlying server objects.  

What we'd like to have is either Spring support for Undertow internal metrics from the [ConnectorStatistics](https://undertow.io/javadoc/2.1.x/io/undertow/server/ConnectorStatistics.html) (similar to the [TomcatMetricsBinder](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/tomcat/TomcatMetricsBinder.java) and [AbstractJettyMetricsBinder](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/web/jetty/AbstractJettyMetricsBinder.java)) or at least have the Undertow object exposed via the `UndertowWebServer` class so that we could expose those metrics ourselves.

wilkinsona:
[...]

Issue #3016, opened 2015-05-20
Make the structure of the Undertow support and the structure of the Tomcat and Jetty support consistent


wilkinsona:
Both `JettyEmbeddedServletContainerFactory` and `TomcatEmbeddedServletContainerFactory` have very few fields and the customisation required to apply the configuration in `ServerProperties.Jetty` and `ServerProperties.Tomcat` is applied using customisers. In comparison, `UndertowEmbeddedServletContainerFactory` contains numerous fields that duplicate those in `ServerProperties.Undertow` and it handles its own customisation.

I think this inconsistency should be removed, and I think that should happen by changing the Undertow support to align with the Jetty and Tomcat support. At the same time, we could consider moving the customisation logic out of `ServerProperties.Jetty` and `ServerProperties.Tomcat` into separate classes still in the `spring-boot-autoconfigure` module.

snicoll:
Paging @bclozel as I think we might have fixed this one in the meantime.

wilkinsona:
Closing as Undertow support has been removed.

PR #39916, opened 2024-03-12
Add public getUndertow method to UndertowWebServer

wanger26:
### Change Summary
- getUndertow() to UndertowWebServer will be null until the server has been started. Issue #3016 would enable getUndertow() to behave in a similar way to getTomcat() on TomcatWebServer.

### Issue
- https://github.com/spring-projects/spring-boot/issues/39839


mhalbritter:
Thank you!

PR #39948, opened 2024-03-15
Delay interaction with Lazy CqlSession bean until first required usage

christophstrobl:
Though the `CqlSession` provided by the `CassandraAutoConfiguration` can be lazy, the configuration for data-cassandra triggers early bean instantiation. 
This PR uses changed API in the data-cassandra project to make use of the intended lazy bean initialization and therefore prevents the application from failing already at startup when cassandra might not (yet) be ready.

Requires: Upgrade to `spring-data:2024.0.0-RC1`
Depends on: spring-projects/spring-data-cassandra#1485

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/cassandra/CassandraDataAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/cassandra/CassandraDataAutoConfiguration.java
@@ -32,6 +32,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.domain.EntityScanPackages;
 import org.springframework.boot.context.properties.bind.Binder;
 import org.springframework.context.annotation.Bean;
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/cassandra/CassandraDataAutoConfiguration.java
```diff
@@ -63,7 +65,7 @@ public class CassandraDataAutoConfiguration {
 
 	private final CqlSession session;
 
-	public CassandraDataAutoConfiguration(CqlSession session) {
+	public CassandraDataAutoConfiguration(@Lazy CqlSession session) {
```
wilkinsona, line 68: I wonder if we should inject and store an `ObjectProvider<CqlSession>` here rather than relying on the proxying that `@Lazy` requires.

christophstrobl: Relying on the proxy was the less invasive variant.

Issue #39913, opened 2024-03-12
Add standardized property to distinguish a group of applications


jonatan-ivanov:
A property (e.g.: `spring.application.namespace`) that distinguishes group of applications (e.g.: the company/org/team that owns these apps) would be beneficial to the users.

Right now `spring.application.name` needs to be unique for each app, multiple teams cannot use the same application name if they for example ship logs/metrics/traces/etc. to the same backend. With an additional "namespace" property, this would be much easier since `spring.application.name` only needs to be unique in the context of its "namespace".

This property could be used for
- Log correlation: right now `spring.application.name` is logged and it would be easier to search for application logs if there is also a "namespace" property.
- Metrics: the name and the "namespace" of the app can be aded as tags (dimensions) so that teams would be able to aggregate metrics across these dimensions, e.g.: what is the wrror rate of all of the apps I own?
- Tracing: similarly to log correlation, it helps to search for traces/spans.
- Discovery: finding the owner of an app can be easier using the "namespace" as well as finding the applications that are owned by a team.

mhalbritter:
> so that teams would be able to aggregate metrics across these dimensions, e.g.: what is the error rate of all of the apps I own?

I really like this idea.

csterwa:
In many of my conversations with development teams at different organizations, this is a common discussion topic. The ability to provide some indicator that a set of applications are part of a larger "business application" so that they can be viewed in metrics, portals, traces and more could be very helpful.

The term `namespace` may have be confused with the Kubernetes concept of Namespace. Maybe we need a few suggestions for the property name to choose from? I have a few here that may be candidates, as well:

* `spring.application.group`
* `spring.application.family`
* `spring.application.membership`
* `spring.application.classification`
* `spring.application.association`

philwebb:
We like the suggestion and so far prefer `spring.application.group` as the property name.

jonatan-ivanov:
FWIW, I like `group` and `family`, group sounds a bit better: it sounds more neutral/natural for this.

philwebb:
Closing in favor of PR #39957. Thanks @wanger26!

PR #39957, opened 2024-03-17
Add standardized property to distinguish a group of applications

wanger26:
### Change Summary 
- Add a standardized property to provide some indicator that a set of applications are part of a larger "business application" so that they can be viewed in metrics, portals, traces and more

### Issue 
- https://github.com/spring-projects/spring-boot/issues/39913

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpPropertiesConfigAdapter.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpPropertiesConfigAdapter.java
@@ -43,6 +43,11 @@ class OtlpPropertiesConfigAdapter extends StepRegistryPropertiesConfigAdapter<Ot
 	 */
 	private static final String DEFAULT_APPLICATION_NAME = "unknown_service";
 
+	/**
+	 * Default value for application group if {@code spring.application.group} is not set.
+	 */
+	private static final String DEFAULT_APPLICATION_GROUP = "unknown_group";
+
 	private final OpenTelemetryProperties openTelemetryProperties;
 
 	private final OtlpMetricsConnectionDetails connectionDetails;
@@ -79,11 +84,16 @@ class OtlpPropertiesConfigAdapter extends StepRegistryPropertiesConfigAdapter<Ot
 		Map<String, String> result = new HashMap<>((!CollectionUtils.isEmpty(resourceAttributes)) ? resourceAttributes
 				: get(OtlpProperties::getResourceAttributes, OtlpConfig.super::resourceAttributes));
 		result.computeIfAbsent("service.name", (key) -> getApplicationName());
+		result.computeIfAbsent("service.group", (key) -> getApplicationGroup());
 		return Collections.unmodifiableMap(result);
 	}
 
 	private String getApplicationName() {
 		return this.environment.getProperty("spring.application.name", DEFAULT_APPLICATION_NAME);
+	}
+
+	private String getApplicationGroup() {
+		return this.environment.getProperty("spring.application.group", DEFAULT_APPLICATION_GROUP);
 	}
 
 	@Override
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/actuator/tracing.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/actuator/tracing.adoc
@@ -88,9 +88,10 @@ logging:
   pattern:
     correlation: "[${spring.application.name:},%X{traceId:-},%X{spanId:-}] "
   include-application-name: false
+  include-application-group: false
 ----
 
-NOTE: In the example above, configprop:logging.include-application-name[] is set to `false` to avoid the application name being duplicated in the log messages (configprop:logging.pattern.correlation[] already contains it).
+NOTE: In the example above, configprop:logging.include-application-name[] and configprop:logging.include-application-group[] is set to `false` to avoid the application name being duplicated in the log messages (configprop:logging.pattern.correlation[] already contains it).
 It's also worth mentioning that configprop:logging.pattern.correlation[] contains a trailing space so that it is separated from the logger name that comes right after it by default.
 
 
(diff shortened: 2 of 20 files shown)


wilkinsona:
Thanks for the PR, @wanger26. You've marked it as a draft but it looks pretty complete to me. Did you have some other changes in mind that you wanted to make before it's ready for review?

wanger26:
Sorry for the delay on this PR @wilkinsona, I have completed all the changes I was hoping to make but was holding off on publishing the PR until I got a green build. I was hoping to pick this up at some point this week and get the build green.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpPropertiesConfigAdapter.java
```diff
@@ -43,6 +43,11 @@ class OtlpPropertiesConfigAdapter extends StepRegistryPropertiesConfigAdapter<Ot
 	 */
 	private static final String DEFAULT_APPLICATION_NAME = "unknown_service";
 
+	/**
+	 * Default value for application group if {@code spring.application.group} is not set.
+	 */
+	private static final String DEFAULT_APPLICATION_GROUP = "unknown_group";
```
quaff, line 49: I think `default` is better than `unknown` here.

wilkinsona: I disagree. `unknown_group` aligns with the existing `unknown_service` that's used when there's no application name.

wilkinsona: I'm afraid I still disagree. This feature isn't specific to Kubernetes so aligning with its default namespace does not make sense. Furthermore, it isn't a default group but an indication that we do not know the application's group. As such, I think `unknown_group` is more meaningful as it accurately reflects the situation that we're in.

Citymonstret: The service name is supposed to be unique per application. It not being specified means that it is indeed unknown, as there cannot be any sensible default value that actually makes sense. The groups are shared, so it not being specified is a pretty good indicator that it belongs to the (shared) default group, not the shared unknown group. `unknown_group` makes it seem like there is something wrong with the application configuration, whereas you have just opted to not specify an actual group.

wilkinsona: >  It not being specified means that it is indeed unknown, as there cannot be any sensible default value that actually makes sense.

That is the case here.

> The groups are shared, so it not being specified is a pretty good indicator that it belongs to the (shared) default group

There's no "the groups" here. As described in #39913, a group can be a company, org, or team, or pretty much whatever else makes sense for grouping a set of applications. There's no concept of a default group as we don't know that such a thing makes sense in any given deployment environment.

> `unknown_group` makes it seem like there is something wrong with the application configuration, whereas you have just opted to not specify an actual group.

"unknown" doesn't imply that something is wrong, all it means is that the group is unknown. I think that accurately reflects the situation as the group really is not known.

quaff: IMO Tag `service.group` shouldn't be added if it's `unknown_group`, It doesn't make sense from observation perspective but increase network packet size.

spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/actuator/tracing.adoc
```diff
@@ -88,9 +88,10 @@ logging:
   pattern:
     correlation: "[${spring.application.name:},%X{traceId:-},%X{spanId:-}] "
   include-application-name: false
+  include-application-group: false
 ----
 
-NOTE: In the example above, configprop:logging.include-application-name[] is set to `false` to avoid the application name being duplicated in the log messages (configprop:logging.pattern.correlation[] already contains it).
+NOTE: In the example above, configprop:logging.include-application-name[] and configprop:logging.include-application-group[] is set to `false` to avoid the application name being duplicated in the log messages (configprop:logging.pattern.correlation[] already contains it).
```
wilkinsona, line 94: `include-application-group` does not need to be set to `false` here as the preceding example of a custom correlation pattern doesn't include `spring.application.group`. The example shouldn't be changed as it matches Spring Cloud Sleuth's behavior. I think the change on this line should be reverted instead.

spring-boot-project/spring-boot/src/main/resources/org/springframework/boot/logging/logback/defaults.xml
```diff
@@ -6,15 +6,16 @@ Default logback configuration provided for import
 
 <included>
 	<conversionRule conversionWord="applicationName" converterClass="org.springframework.boot.logging.logback.ApplicationNameConverter" />
+	<conversionRule conversionWord="applicationGroup" converterClass="org.springframework.boot.logging.logback.ApplicationGroupConverter" />
```
wilkinsona, line 9: Minor nit: these rules are ordered based the conversion word so `applicationGroup` should go before `applicationName`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpPropertiesConfigAdapter.java
mhalbritter: I agree. We only set `service.name` to `unknown_application` because it's a [required attribute in otel](https://opentelemetry.io/docs/specs/semconv/resource/#service). This isn't the case for `service.group`, so we can just omit it, if it's not set.

mhalbritter:
Thanks! I've polished quite a bit in https://github.com/spring-projects/spring-boot/commit/de001f5af14233ce0541f9c35b7abafe1349cd42, e.g. removing the `service.group` from OTel if the group property isn't set, and I also removed the duplicate logging of thread name in both Logback and Log4j.

Issue #39748, opened 2024-02-24
Thread name prefix is not always set when using virtual threads


OathMeadow:
We noticed that the thread name was missing as a field in our logs logged by the rabbit listener after activating `VirtualThreads`.
We use Loki in Grafana where we have multiple fields, where the thread name is one of them. To be able to see which thread logs what is useful to identify issues.

We noticed that in the class `org.springframework.boot.autoconfigure.amqp.RabbitAnnotationDrivenConfiguration`, Spring Boot does not set a thread prefix to the `VirtualThreadTaskExecutor` class when `VirtualThreads` is configured. In fact, when not set, it defaults to `null`. Perhaps that is by design, but It seems Spring Boot sets a thread name prefix in some AutoConfigurations and some not.

May I suggest names something like:

```
	@Bean(name = "directRabbitListenerContainerFactoryConfigurer")
	@ConditionalOnMissingBean
	@ConditionalOnThreading(Threading.VIRTUAL)
	DirectRabbitListenerContainerFactoryConfigurer directRabbitListenerContainerFactoryConfigurerVirtualThreads() {
		DirectRabbitListenerContainerFactoryConfigurer configurer = directListenerConfigurer();
		configurer.setTaskExecutor(new VirtualThreadTaskExecutor("direct-rabbit-listener-"));
		return configurer;
	}

	@Bean(name = "simpleRabbitListenerContainerFactoryConfigurer")
	@ConditionalOnMissingBean
	@ConditionalOnThreading(Threading.VIRTUAL)
	SimpleRabbitListenerContainerFactoryConfigurer simpleRabbitListenerContainerFactoryConfigurerVirtualThreads() {
		SimpleRabbitListenerContainerFactoryConfigurer configurer = simpleListenerConfigurer();
		configurer.setTaskExecutor(new VirtualThreadTaskExecutor("simple-rabbit-listener-"));
		return configurer;
	}
```
Thank you

mhalbritter:
We set the names for Redis, Kafka and Undertow. We don't set them for Rabbit and for Pulsar. We should make that consistent. IMHO we should name threads in all cases.

mhalbritter:
Sure. I'll assign the issue to you. This is targeted for the 3.2.x branch, so please create your PR based on that. If you need help, please chime back in.

MazizEsa:
@mhalbritter . When is the 3.2.x going to be release. I will do this outside of my working time.

mhalbritter:
[...]

PR #39958, opened 2024-03-17
Thread name prefix is not always set when using virtual threads

MazizEsa:
Currently the thread names are missing for rabbit listener, pulsars' consumer and task executor.
This PR added the thread names for mentioned components following convention specified in other autoconfigs via RabbitAnnotationDrivenConfiguration and PulsarAutoconfiguration.

Tests are added as well to check the threads names are specified when threads are created (via java.util.concurrent.ThreadFactory#newThread) through the factory.

Request / bug ticket, https://github.com/spring-projects/spring-boot/issues/39748


MazizEsa:
cc. @mhalbritter @scottfrederick

mhalbritter:
Thanks @MazizEsa! I've polished your changes a bit in https://github.com/spring-projects/spring-boot/commit/09652cb2826aabf4aac56b15f99110417fa6c8e9.

MazizEsa:
I see. You guys prefer that way. Got it.

PR #39989, opened 2024-03-18
Upgrade to MongoDB 5.0

christophstrobl:
The MongoDB 5.0 Java Driver ships with previously deprecated API now removed (mongodb/mongo-java-driver#1268). To allow users to downgrade if needed, calls to those parts of the driver need to be done reflectively.

Related to: spring-projects/spring-data-mongodb#4663

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoReactiveAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoReactiveAutoConfiguration.java
@@ -129,10 +129,8 @@ public class MongoReactiveAutoConfiguration {
 			}
 		}
 
-		@SuppressWarnings("deprecation")
 		private boolean isCustomTransportConfiguration(MongoClientSettings settings) {
-			return settings != null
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks, @christophstrobl. Do you anticipate a strong need to downgrade to the 4.x driver? If at all possible, we'd prefer not to support two different major versions at the same time.

christophstrobl:
There's a few more places with removed API around indexing options, etc. and as far as I know the 5.0 release is one of the more aggressive ones in terms of breaking changes done by the MongoDB team so far. For data we intend to support both 5.0 and 4.11 for the 2024.0 train and remove backwards compatibility with 2024.1.

wilkinsona:
That sounds good to me. We can always restore backwards compatibility if we get enough feedback in that direction.

wilkinsona:
To be merged after M3 and once we're back on Data snapshots again.

christophstrobl:
@mauhiz IIRC the `com.mongodb.kotlin.client.coroutine.MongoClient` is a wrapper around the `com.mongodb.reactivestreams.client.MongoClient`. 
For data-mongodb the Kotlin variants cannot be used directly along with the `DatabaseFactory` implementations. There's an open [PR#4402](https://github.com/spring-projects/spring-data-mongodb/pull/4402) in the data project for that, but it would require to access mongodb driver internals reflectively.

wilkinsona:
Superseded by #40259.

Issue #40027, opened 2024-03-19
Replace activemq-client-jakarta with activemq-client


mattrpav:
ActiveMQ 6.x is jakarta-based, so using the activemq-client-jakarta transition package is no longer needed.

wilkinsona:
Closing in favor of #40028.

Issue #38404, opened 2023-11-21
Reinstate support for auto-configuring an embedded ActiveMQ broker


mattrpav:
ActiveMQ 6.0.1 has been released and includes Jakarta support on the broker-side.

Related:
#34011
#35048

philwebb:
Given that we're at the RC phase we've decided to wait until Spring Boot 3.3 to do the upgrade.

Grey1024:
I am a Java backend programmer from China, and I have lost my job. Now, I am looking for a coding job outside of China, and I don’t mind the salary. I am even willing to work for free. Additionally, I hope to improve my English communication skills. If you need my help, please contact me at [jeeun2023@outlook.com](mailto:jeeun2023@outlook.com).
I just want to find something to do; I love programming. If you have the patience to explain your work tasks to someone whose English is not good, I am willing to work for you without caring about payment.

mattrpav:
Revising to ActiveMQ v6.1.0

mattrpav:
@wilkinsona thanks for the update. I’ll work to get a PR up this week for the broker-side work.

mattrpav:
@snicoll thanks for picking it up! I lost my in-progress work in a git repo shuffle.

PR #40028, opened 2024-03-19
Use activemq-client dependency vs transitional activemq-client-jakarta

mattrpav:
Use the activemq-client dependency per: #40027 

Ahead of adding embedded broker support in as part of [#38404]

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
@@ -79,7 +79,7 @@ dependencies {
 	optional("jakarta.persistence:jakarta.persistence-api")
 	optional("jakarta.servlet:jakarta.servlet-api")
 	optional("javax.cache:cache-api")
-	optional("org.apache.activemq:activemq-client-jakarta")
+	optional("org.apache.activemq:activemq-client")
 	optional("org.apache.commons:commons-dbcp2") {
 		exclude group: "commons-logging", module: "commons-logging"
 	}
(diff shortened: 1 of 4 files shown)


mattrpav:
@wilkinsona for your review. This would be cleaner to have in ahead of [#38404].

NOTE: I'll be removing activemq-client-jakarta module in next ActiveMQ 6.x release. (The Maven relocation has been published.)

spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -21,7 +21,6 @@ bom {
 					"activemq-blueprint",
 					"activemq-broker",
 					"activemq-client",
-					"activemq-client-jakarta",
```
wilkinsona, line 24: This should remain so that users' builds aren't broken if they have a dependency on `activemq-client-jakarta`.

wilkinsona (changes requested):
Thanks very much, @mattrpav. It looks good to me other than I think we should leave `spring-boot-dependencies` as-is so that we continue to manage `activemq-client-jakarta` for now.

BTW, have you considered publishing a bom for ActiveMQ? We would certainly welcome it as it would save us from having to manually list ~all of ActiveMQ Classic's modules.

spring-boot-project/spring-boot-dependencies/build.gradle
mattrpav: Fixed

mattrpav:
@wilkinsona re: activemq-bom yep ;-) I've already got a PR for it (https://github.com/apache/activemq/pull/1178) that should land in the next ActiveMQ release.

wilkinsona (approved):

wilkinsona:
Thanks very much, @mattrpav.

adianand13:
thanks for the fix. I was about to raise the PR

PR #40029, opened 2024-03-19
Reinstate support for auto-configuring an embedded ActiveMQ broker

mattrpav:
#38404


spring-boot-project/spring-boot-docs/src/docs/asciidoc/messaging/jms.adoc
```diff
@@ -8,13 +8,13 @@ Spring Boot also auto-configures the necessary infrastructure to send and receiv
 
 
 [[messaging.jms.activemq]]
-=== ActiveMQ "Classic" Support
```
wilkinsona, line 11: We can certainly remove this quotes, but this should be done separately as it's unrelated to reinstating support for auto-configuring an embedded broker.

wilkinsona (changes requested):

mattrpav:
@wilkinsona Thanks for the ping. Yes, I plan to get this wrapped up. What is the timeframe to make it in the next Spring Boot release?

scottfrederick:
@mattrpav Since this would be considered a feature addition, we would only add it in a new minor release. The next scheduled minor release is 3.4 in November of this year. There will be several milestone releases and a release candidate leading up to the minor version release, which would give users a chance to try it out.

snicoll:
@mattrpav I've looked at what you've done and there's the question on whether we want to reinstate the behavior as it was before. To figure that out, I am happy to look at the various options and polish your proposal. Are you ok with that? We'd like to make sure you haven't wasted time on this.

snicoll:
> I have the WIP PR at about 80%

Can you push the code then?

> At the same time, improvements always sound compelling, what did you have in mind? Can these be done side-by-side with the existing capabilities and property flags?

I didn't mention improvements. It's figuring out how we get from what we have to this. Right now, a Spring Boot 3.x users of ActiveMQ with no configuration would get an infrastructure that connects to an existing broker on localhost using tcp. After this change, it would use an embedded broker in memory. This can be very surprising and we're wondering if that's what we should be doing.

snicoll:
@mattrpav is this PR ready for review now?

snicoll:
@mattrpav thanks for the proposal but it's incomplete and the build is failing. Also, we can't just add the broker in the starter as you suggest as it would be a breaking change. I am taking over.

Issue #39234, opened 2024-01-18
Validation not working when nested records are used for configuration properties


michalgebauer:
I tried both ways how to use nested records as configuration properties. In non of them validation works. I have in the same project configuration with just one record without nesting and that works.

Examples

```
@Validated
@ConfigurationProperties(prefix = "client")
public record ClientProperties(@NestedConfigurationProperty ServiceOne serviceOne) {

}
@Validated
record ServiceOne(@NotBlank String url, @Email String email) {
}
```

```
@Validated
@ConfigurationProperties(prefix = "client")
public record ClientProperties(ServiceOne serviceOne) {

@Validated
public record ServiceOne(@NotBlank String url, @Email String email) {
}

}

```

michalgebauer:
I'm using Spring Boot 3.2.1, all properties are actually null (not provided).
See my demo project I created just for the purpose of this issue:
[https://github.com/michalgebauer/nested-records-validation/blob/main/src/main/java/com/example/demo/DemoApplication.java](https://github.com/michalgebauer/nested-records-validation/blob/main/src/main/java/com/example/demo/DemoApplication.java)

I have similar problems with nested properties even if I use classes and not records.

wilkinsona:
Thanks. The behaviour you have described is to be expected. With no properties to bind to an instance of `ServiceOne`, no instance of it is created and, therefore, there's nothing to validate. If you want to prohibit this, you can annotate the record component with `@NotNull`:

```java
@Validated
@ConfigurationProperties(prefix = "client")
record ClientProperties(@NotNull ServiceOne serviceOne) {

	@Validated
	public record ServiceOne(@NotBlank String url, @Email String email) {
	}

}
```

With this in place, the app will fail to start due to a validation failure.

michalgebauer:
Thank you 🙌

PR #40035, opened 2024-03-20
BindValidationFailureAnalyzer uses wrong target

izeye:
This PR fixes target in `BindValidationFailureAnalyzer.getBindValidationExceptionDetails()` as it seems to be using `BindException` accidentally.

For example, the following output is included in https://github.com/spring-projects/spring-boot/issues/39234#issuecomment-1898725399.

```
Binding to target org.springframework.boot.context.properties.bind.BindException: Failed to bind properties under 'client' to com.example.demo.ClientProperties failed:
```


mhalbritter:
Thanks!

PR #40037, opened 2024-03-20
Add SslBundle support to MailSender

ruifigueira:
Currently, `MailProperties` doesn't support SslBundle configuration. There are scenarios where a client has a SMTP server with a certificate signed by a local CA and we don't have a simple way to configure MailSender to use a different SSL context, and `SslBundles` seems to be a great fit here.

This PR adds support to SSL configuration using a SslBundle.


ruifigueira:
I added some integration tests using a [Mailpit](https://mailpit.axllent.org) testcontainer, both in TLS/SSL mode and using `STARTTLS`.

mhalbritter:
Thanks a lot!

PR #40050, opened 2024-03-21
Upgrade to Lombok 1.18.32

anbusampath:
https://projectlombok.org/changelog -  Initial JDK22 support added.
https://github.com/projectlombok/lombok/issues/3630

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1098,7 +1098,7 @@ bom {
 			site("https://logback.qos.ch")
 		}
 	}
-	library("Lombok", "1.18.30") {
+	library("Lombok", "1.18.32") {
 		group("org.projectlombok") {
 			modules = [
 				"lombok"


bclozel:
Thanks, but there is no need to to provide PRs for dependency upgrades, especially if no code change is required.

As mentioned in our issue template:
> You DO NOT need to raise an issue for a managed dependency version upgrade as there's a semi-automatic process for checking managed dependencies for new versions before a release.

We will perform upgrades matching [our 3rd party upgrade policy](https://github.com/spring-projects/spring-boot/wiki/Supported-Versions#third-party-dependencies).
If this is an urgent matter for you, there is no need to wait for us releasing as you can use a [Gradle](https://docs.spring.io/spring-boot/docs/current/gradle-plugin/reference/htmlsingle/#managing-dependencies.dependency-management-plugin.customizing) or [Maven](https://docs.spring.io/spring-boot/docs/current/maven-plugin/reference/htmlsingle/#using.parent-pom) build property to override the library version.
All [version properties are listed in the reference documentation appendix](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#appendix.dependency-versions.properties).

wilkinsona:
Unfortunately, it narrowly missed today's releases. If you're using Lombok and Java 22, you should override the version as Brian has described above.

PR #40054, opened 2024-03-21
Fix forward-headers-strategy documentation regarding cloud defaults

ViliusS:
forward-headers-strategy=NATIVE is actually used on all cloud platforms: https://github.com/spring-projects/spring-boot/blob/3abf43bb11edd7fdc7f0b35d7773aa32bb291505/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudPlatform.java#L219-L221

This PR adjusts for that. I didn't know how to correctly specify what are exactly those supported Cloud Platforms, though. I'm linking it to API docs, hopefully that's OK.


philwebb:
Thanks very much @ViliusS!

PR #40073, opened 2024-03-22
spring-boot-dependencies should not manage org.flywaydb:flyway-community-db-support as it is not released as part of Flyway

eddumelendez:
`org.flywaydb:flyway-community-db-support` is not provided anymore as part of the flyway release. Now, it is an independent project and the version is not aligned with flyway-core. See https://github.com/flyway/flyway-community-db-support

Also, see https://github.com/flyway/flyway-community-db-support/issues/33


mhalbritter:
Thanks!

PR #40081, opened 2024-03-23
PropertiesLdapConnectionDetails should not be public

izeye:
This PR changes to make the `PropertiesLdapConnectionDetails` package-private as given that the other `Properties...ConnectionDetails` classes except the `PropertiesMongoConnectionDetails` are package-private, it appears to be accidental.

I didn't change the `PropertiesMongoConnectionDetails` as it's already been technically `public` on releases.


mhalbritter:
Thanks!

Issue #32054, opened 2022-08-10
Add @WebEndpointTest attribute to specify target frameworks


vpavic:
As discussed in https://github.com/spring-projects/spring-boot/pull/32046#pullrequestreview-1067421804, it would be useful if `@WebEndpointTest` had an attribute that would allow to specify target web frameworks that given test should be executed against.

wilkinsona:
Possible implementation: https://github.com/wilkinsona/spring-boot/tree/gh-32054.

PR #40083, opened 2024-03-23
Add Javadoc since to WebEndpointTest.Infrastructure

izeye:
This PR adds a Javadoc `@since` to the `WebEndpointTest.Infrastructure`.

See gh-32054


wilkinsona:
Thanks, @izeye, but as this is a class in `src/test/java` it doesn't need `@since` documentation.

PR #40085, opened 2024-03-23
Apply instanceof pattern matching

FelixDes:
This pull request applies the instanceof pattern matching across the codebase. This enhancement will improve code readability.


scottfrederick:
Thanks for the contribution @FelixDes.

PR #40086, opened 2024-03-24
Upgrade gradle from 7.6.4 to 8.7

AlexElin:
Upgrade gradle from 7.6.4 to the latest new major version - 8.7


wilkinsona:
Thanks, but there's quite a bit more to a Gradle upgrade as the version has an effect on the versions of Gradle that we test against. We'll tackle this ourselves when we're ready.

PR #40088, opened 2024-03-24
Add new ConditionalOnAnnotation for conditional bean loading

Zodiacxxy:
This commit introduces the ConditionalOnAnnotation, a custom conditional annotation that enables bean registration based on the presence of a specified annotation with RUNTIME retention. It ensures beans are only loaded when the annotated class or method is marked with a user-defined annotation, facilitating feature toggling and modular configuration.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/ConditionalOnAnnotation.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/condition/ConditionalOnAnnotation.java
@@ -1,0 +1,79 @@
+/*
+ * Copyright 2012-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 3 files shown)


Zodiacxxy:
Add new ConditionalOnAnnotation for conditional bean loading

This commit introduces the ConditionalOnAnnotation, a custom conditional annotation that enables bean registration based on the presence of a specified annotation with RUNTIME retention. It ensures beans are only loaded when the annotated class or method is marked with a user-defined annotation, facilitating feature toggling and modular configuration.

wilkinsona:
Thanks for the suggestion. Looking at the implementation, is this not the same as the functionality that's already provided by the `annotation` attribute of `@ConditionalOnBean`?

Zodiacxxy:
> Thanks for the suggestion. Looking at the implementation, is this not the same as the functionality that's already provided by the `annotation` attribute of `@ConditionalOnBean`?

yes, you are right. That @ConditionalOnBean is same with that and it works for me. Thanks for so quickly replay.

PR #40092, opened 2024-03-25
Do not use okhttp client

ddaaac:
In my project, Okhttp exists in the classpath. 
Therefore, when initializing RestTemplate, OkHttp is selected from ClientHttpRequestFactories.
However, starting from Spring 6.1, this client has been deprecated and _does not work_.

So I have to declare the requestFactory like code below.

```kotlin
        return RestTemplateBuilder()
            .requestFactory { _ -> SimpleClientHttpRequestFactory() }
```

But since Spring 6.2, OkHttpClient will be removed.
Therefore, there is no need to declare a requestFactory as the default client is chosen automatically.

So, it would be nice if there was no need to declare requestFactory even in spring 6.1.x version.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/ClientHttpRequestFactories.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/client/ClientHttpRequestFactories.java
@@ -32,6 +32,8 @@ import javax.net.ssl.TrustManager;
 import javax.net.ssl.X509TrustManager;
 
 import okhttp3.OkHttpClient;
+import org.apache.commons.logging.Log;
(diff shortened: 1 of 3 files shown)


wilkinsona:
> starting from Spring 6.1, this client has been deprecated and does not work.

Even though it's deprecated OkHttp should continue to work as it did before. If that's not the case it should be treated as a bug.

The deprecation period is intended to give people who are using OkHttp time to migrate to another client library. If we make the changes proposed here, it will break anyone who has not yet completed that migration, making the deprecation pointless.

Spring Boot's support for OkHttp will be removed in Spring Boot 3.4 alongside our upgrade to Spring Framework 6.2.

PR #40101, opened 2024-03-25
Fix extension hint pattern regex

zinzoddari:
Hi 🙋🏻‍♀️
I removed unnecessary escape characters from the regular expression.


mhalbritter:
Thanks for letting us know and for the PR. There are more redundant escapes in the codebase, we're fixing this with #40116.

PR #40111, opened 2024-03-26
Avoid use of bean overriding in tests

snicoll:
While working on https://github.com/spring-projects/spring-framework/issues/31288, I wanted to get a feeling of how disruptive the change was going to be.

I've noticed a number of legitimate use cases and some that simply look an oversight. It's not the full list but here's a few for a start.


mhalbritter:
Thanks Stéphane!

Issue #40039, opened 2024-03-20
batch.adoc: broken API links


commonquail:
The Spring Boot 3.2.3 reference documentation's section [_18.12.1 Specifying a Batch Data Source_](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#howto.batch.specifying-a-data-source), which lives in `howto/batch.adoc`, attempts to link to Spring Batch's Javadoc for `@EnableBatchProcessing` and `DefaultBatchConfiguration`. However, the generated URLs are

* <https://docs.spring.io/spring-batch/docs/5.1.1/org/springframework/batch/core/configuration/annotation/EnableBatchProcessing.html>
* <https://docs.spring.io/spring-batch/docs/5.1.1/org/springframework/batch/core/configuration/support/DefaultBatchConfiguration.html>

where instead the common prefix should have been <https://docs.spring.io/spring-batch/docs/5.1.1/api/>. The `/api/` segment appears to be missing from `asciidoc/attributes.adoc`'s `:spring-batch-api:` attribute.

scottfrederick:
The links are broken starting with 3.2.x. Marking this as `on-hold` for now as we don't want to introduce documentation changes until the Antorra migration is complete.

mhalbritter:
Antora migration complete.

nahidshahin:
In main branch I found /api/ is there but `/org/springframework/batch` is missing. Check:
https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-dependencies/build.gradle#L1824

So the broken URL was like: https://docs.spring.io/spring-batch/docs/5.1.1/api/core/configuration/annotation/EnableBatchProcessing.html

nahidshahin:
This PR https://github.com/spring-projects/spring-boot/pull/40141 fixed the issue in main branch, I tested locally. Thanks

PR #40141, opened 2024-03-31
Links to Spring Batch javadoc for EnableBatchProcessing and DefaultBatchConfiguration are broken

nahidshahin:
Fixed broken API links in batch.adoc

This is a fix for https://github.com/spring-projects/spring-boot/issues/40039

In main branch I found /api/ is there but `/org/springframework/batch` is missing. Check:
https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-dependencies/build.gradle#L1824

So the broken URL was like: https://docs.spring.io/spring-batch/docs/5.1.1/api/core/configuration/annotation/EnableBatchProcessing.html

This PR fixed the issue in main branch, I tested locally. Thanks

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1821,7 +1821,7 @@ bom {
 		links {
 			site("https://spring.io/projects/spring-batch")
 			github("https://github.com/spring-projects/spring-batch")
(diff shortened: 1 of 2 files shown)


scottfrederick:
This will need a different fix for 3.2.x, which doesn't have the build metadata for links.

wilkinsona:
Thanks very much, @nahidshahin.

PR #40147, opened 2024-04-02
Log with Current ClassLoader

gaosaroma:
In [koupleless](https://github.com/koupleless/koupleless), multiple spring boot applications run together in a single JVM, and these applications are isolated by different class loaders. 

In order to allow different spring boot applications to print logs according to their own configurations when application running, two issues need to be addressed:

1. Put the application's configuration into thread context, so that the application can read the log configuration from the current thread when it starts.
2. The spring boot applications use the LoggerContext corresponding to the current thread context's classloader.

I'm not sure if this is the best solution. If possible, I hope we can have more discussion.


philwebb:
Thanks for the suggestion but this kind of deployment isn't something that we want to support directly in Spring Boot. We've had difficulty in the past trying to manage multiple applications in a single JVM and we think that users that want to do that should consider an application server with war deployments.

Issue #40160, opened 2024-04-03
AnsiOutput.detectIfAnsiCapable broken on JDK22


kilink:
The `detectIfAnsiCapable` method in `AnsiOutput` does not work correctly on JDK22. It checks if System.console() returns null, but on JDK22 System.console() never returns null; there's a [new method on Console](https://docs.oracle.com/en%2Fjava%2Fjavase%2F22%2Fdocs%2Fapi%2F%2F/java.base/java/io/Console.html#isTerminal()) to detect if the instance is a terminal or not.

Some details [here](https://inside.java/2023/07/31/quality-heads-up/) about the change:

> System.console() now returns a Console object when the standard streams are redirected or connected to a virtual terminal. Prior to JDK 22, System.console() instead returned null for these cases. This change may impact code that checks the return from System.console() to test if the JVM is connected to a terminal. If required, the -Djdk.console=java.base flag will restore the old behavior where the console is only returned when it is connected to a terminal. Starting JDK 22, one could also use the new Console.isTerminal() method to test if the console is connected to a terminal.

The result is that on JDK22, color / ANSI output is turned on when it shouldn't be.

facewise:
@scottfrederick @kilink 
Hello. Can I pick up this issue?

philwebb:
@facewise Feel free to have a go, but we're not 100% sure how to fix it since we need to keep our build on Java 17. We might be able to use reflection to make the call to `Console.isTerminal()`.

It's also going to be pretty hard to test. We might be able to create a testcontainers test, or we might just have to accept manual testing for now.

facewise:
@philwebb 
Is it OK to add a testcontainers test in `spring-boot-tests/spring-boot-integration-tests`?

It seems great to make a new test similar to tests in `LoaderIntegrationTests`.

philwebb:
@facewise That sounds fine, we can always move it around when we merge the code if we find a better long-term home.

wilkinsona:
I think we may be able to test it by reverting most of https://github.com/spring-projects/spring-boot/commit/8efdc1e46dc4168774b9a683ea7140770ae30ebc.

facewise:
@philwebb 
I made a PR of it. Please check it out.

Should I accept the reversion @wilkinsona said?

scottfrederick:
Closing in favor of #40172

PR #40172, opened 2024-04-05
AnsiOutput.detectIfAnsiCapable broken on JDK22

facewise:
### Change Summary
Fix a method `detectIfAnsiCapable` in `AnsiOutput` using reflection invoking `Console.isTerminal()` which is newly added on JDK22.

As mentioned in #40160, starting JDK 22, `System.console()` never returns `null`.
Instead, `Console.isTerminal()` is provided so we can use it to detect if the console is a terminal.

Since Spring Boot still has to built on Java 17, check for the existence of `Console.isTerminal()` is added and invokes only if it exists.

### Related Issue
#40160


mhalbritter:

wilkinsona:
@mhalbritter did you try [this suggestion](https://github.com/spring-projects/spring-boot/issues/40160#issuecomment-2039109060) rather than adding the new smoke test?

mhalbritter:
Oh sorry, I missed that. Let me investigate.

mhalbritter:
@wilkinsona Indeed that worked :)

Issue #28174, opened 2021-10-02
Use java.util.List.of for creation of unmodifiable lists


wilkinsona:

mbhave:
Placing this `on-hold` because we've decided to approach this on a case-by-case basis for new usages. There are subtle differences between `List.of()` and the equivalent `Collections` method such as `contains(null)` throwing an NPE which aren't suitable for a blanket search and replace.

Abhishekverma2002:
Would love to contribute in java and algos. If you have any such set of work ,consider me in.

philwebb:
I think we can close this one and adapt to `List.of` in new code or when we touch existing classes. Doing an entire sweep at this point doesn't seem worth it.

PR #40181, opened 2024-04-06
Use List.of() for immutable field azureEnvVariables instead of Arrays.asList()

smc2315:
The Arrays.asList method creates a List based on variable arguments, which can lead to heap pollution.
azureEnvVariables is an immutable field, so it has been improved to utilize Immutable List.


quaff:
There are ubiquitous `Arrays.asList()`, I'll try replace them all and add checkstyle rule to guard that.

wilkinsona:
Thanks for the PR but we considered such a change in the past and [decided against it](https://github.com/spring-projects/spring-boot/issues/28174).

Issue #40177, opened 2024-04-05
spring.jpa.generate-ddl: true cause schema update even if hibernate options not set


nateha1984:
In Spring Boot 2, having the`spring.jpa.generate-ddl` property set to `true` did not perform any updates as long as the corresponding hibernate properties (`hibernate.hbm2ddl.auto`) were not explicitly set. When updating to Spring Boot 3, the same does not apply; the schema update does take place and the Hibernate sql statement is printed to the logs (`show-sql=true`). Here's a git repo where the issue can be demonstrated. Steps to duplicate are in the README.

Hibernate Java Docs also indicate that if no value is provided, the default for `hibernate.hbm2ddl.auto` is `none`

https://github.com/nateha1984/ubiquitous-spork

quaff:
https://docs.spring.io/spring-boot/docs/current/reference/html/data.html#data.sql.jpa-and-spring-data.creating-and-dropping

>> There is also a `spring.jpa.generate-ddl` flag, but it is not used if Hibernate auto-configuration is active, because the `ddl-auto` settings are more fine-grained.

quaff:
> https://docs.spring.io/spring-boot/docs/current/reference/html/data.html#data.sql.jpa-and-spring-data.creating-and-dropping
> 
> > > There is also a `spring.jpa.generate-ddl` flag, but it is not used if Hibernate auto-configuration is active, because the `ddl-auto` settings are more fine-grained.

I think the current behavior should be improved, I've created https://github.com/spring-projects/spring-boot/pull/40185

quaff:
> > https://docs.spring.io/spring-boot/docs/current/reference/html/data.html#data.sql.jpa-and-spring-data.creating-and-dropping
> > > > There is also a `spring.jpa.generate-ddl` flag, but it is not used if Hibernate auto-configuration is active, because the `ddl-auto` settings are more fine-grained.
> 
> I think the current behavior should be improved, I've created #40185

After further investigation, I realize the behavior is correct but the document is wrong.

quaff:
> In Spring Boot 2, having the`spring.jpa.generate-ddl` property set to `true` did not perform any updates as long as the corresponding hibernate properties (`hibernate.hbm2ddl.auto`) were not explicitly set.

@nateha1984 I think it's a bug of Spring Boot 2, and Spring Boot 3 fixed it.

nateha1984:
[...]

PR #40185, opened 2024-04-07
Correct document about spring.jpa.generate-ddl and add tests

quaff:
Closes GH-40177


scottfrederick:
Superseded by #40503. Thanks for helping to clarify this issue @quaff!

PR #40189, opened 2024-04-08
Add Spring Pulsar transaction support

onobc:
Adds auto-config for Spring for Apache Pulsar transactions.

Introduces new txn config props for template and listener under the `spring.pulsar.template.transaction` and
`spring.pulsar.listener.transaction` prefixes, respectively.

Disables transaction support by default (opt-in feature).

> [!NOTE]
> I still need to update/add docs for this but wanted to get this front-loaded as I will be out of pocket at DevNexus this week. 

**TODO**
- [ ] Update ref docs w/ info about transaction support

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
@@ -103,6 +103,14 @@ public class PulsarProperties {
 		return this.template;
 	}
 
+	/**
+	 * Whether transactions are enabled for either the template or the listener.
+	 * @return whether transactions are enabled for either the template or the listener
+	 */
+	public boolean isTransactionEnabled() {
+		return this.template.getTransaction().isEnabled() || this.listener.getTransaction().isEnabled();
+	}
+
 	public static class Client {
 
 		/**
@@ -763,6 +771,11 @@ public class PulsarProperties {
 		 */
 		private boolean observationEnabled;
 
+		/**
+		 * Transaction settings.
+		 */
+		private final Transaction transaction = new Transaction();
+
 		public SchemaType getSchemaType() {
 			return this.schemaType;
 		}
@@ -777,6 +790,10 @@ public class PulsarProperties {
 
 		public void setObservationEnabled(boolean observationEnabled) {
 			this.observationEnabled = observationEnabled;
+		}
+
+		public Transaction getTransaction() {
+			return this.transaction;
 		}
 
 	}
@@ -858,12 +875,65 @@ public class PulsarProperties {
 		 */
 		private boolean observationsEnabled;
 
+		/**
+		 * Transaction settings.
+		 */
+		private final Transaction transaction = new Transaction();
+
 		public boolean isObservationsEnabled() {
 			return this.observationsEnabled;
 		}
 
 		public void setObservationsEnabled(boolean observationsEnabled) {
 			this.observationsEnabled = observationsEnabled;
+		}
+
+		public Transaction getTransaction() {
+			return this.transaction;
+		}
+
+	}
+
+	public static class Transaction {
+
+		/**
+		 * Whether the component supports transactions.
+		 */
+		private boolean enabled;
+
+		/**
+		 * Whether the component requires transactions.
+		 */
+		private boolean required;
+
+		/**
+		 * Duration representing the transaction timeout - null to use default timeout of
+		 * the underlying transaction system, or none if timeouts are not supported.
+		 */
+		private Duration timeout;
+
+		public boolean isEnabled() {
+			return this.enabled;
+		}
+
+		public void setEnabled(boolean enabled) {
+			this.enabled = enabled;
+		}
+
+		public boolean isRequired() {
+			return this.required;
+		}
+
+		public void setRequired(boolean required) {
+			this.required = required;
+		}
+
+		public Duration getTimeout() {
+			return this.timeout;
+		}
+
+		public void setTimeout(Duration timeout) {
+			this.timeout = timeout;
 		}
 
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
@@ -34,6 +34,7 @@ import org.apache.pulsar.client.api.SubscriptionMode;
 import org.apache.pulsar.client.api.SubscriptionType;
 import org.apache.pulsar.common.schema.SchemaType;
 
+import org.springframework.beans.factory.InitializingBean;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
 import org.springframework.util.Assert;
@@ -47,7 +48,7 @@ import org.springframework.util.Assert;
  * @since 3.2.0
  */
 @ConfigurationProperties("spring.pulsar")
-public class PulsarProperties {
+public class PulsarProperties implements InitializingBean {
 
 	private final Client client = new Client();
 
@@ -103,11 +104,17 @@ public class PulsarProperties {
 		return this.template;
 	}
 
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		this.getTemplate().getTransaction().validate("spring.pulsar.template");
+		this.getListener().getTransaction().validate("spring.pulsar.listener");
+	}
+
 	/**
 	 * Whether transactions are enabled for either the template or the listener.
 	 * @return whether transactions are enabled for either the template or the listener
 	 */
-	public boolean isTransactionEnabled() {
+	boolean isTransactionEnabled() {
 		return this.template.getTransaction().isEnabled() || this.listener.getTransaction().isEnabled();
 	}
 
@@ -897,7 +904,7 @@ public class PulsarProperties {
 	public static class Transaction {
 
 		/**
-		 * Whether the component supports transactions.
+		 * Whether transactions are enabled for the component.
 		 */
 		private boolean enabled;
 
@@ -934,6 +941,11 @@ public class PulsarProperties {
 
 		public void setTimeout(Duration timeout) {
 			this.timeout = timeout;
+		}
+
+		void validate(String prefix) {
+			Assert.state((this.enabled || !this.required), "If transactions are required "
+					+ "they must also be enabled - consult your '%s.transaction' properties.".formatted(prefix));
 		}
 
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
@@ -34,9 +34,9 @@ import org.apache.pulsar.client.api.SubscriptionMode;
 import org.apache.pulsar.client.api.SubscriptionType;
 import org.apache.pulsar.common.schema.SchemaType;
 
-import org.springframework.beans.factory.InitializingBean;
 import org.springframework.boot.context.properties.ConfigurationProperties;
 import org.springframework.boot.context.properties.NestedConfigurationProperty;
+import org.springframework.boot.context.properties.source.InvalidConfigurationPropertyValueException;
 import org.springframework.util.Assert;
 
 /**
@@ -48,7 +48,7 @@ import org.springframework.util.Assert;
  * @since 3.2.0
  */
 @ConfigurationProperties("spring.pulsar")
-public class PulsarProperties implements InitializingBean {
+public class PulsarProperties {
 
 	private final Client client = new Client();
 
@@ -102,12 +102,6 @@ public class PulsarProperties implements InitializingBean {
 
 	public Template getTemplate() {
 		return this.template;
-	}
-
-	@Override
-	public void afterPropertiesSet() throws Exception {
-		this.getTemplate().getTransaction().validate("spring.pulsar.template");
-		this.getListener().getTransaction().validate("spring.pulsar.listener");
 	}
 
 	/**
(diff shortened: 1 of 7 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -103,6 +103,14 @@ public Template getTemplate() {
 		return this.template;
 	}
 
+	/**
+	 * Whether transactions are enabled for either the template or the listener.
+	 * @return whether transactions are enabled for either the template or the listener
+	 */
+	public boolean isTransactionEnabled() {
```
onobc, line 110: This is a convenience that is currently used by the `PulsarPropertiesMapper` to decide whether or not to enable the Pulsar client w/ transaction support. We do that if at least one of `spring.pulsar.template.transaction.enabled` or `spring.pulsar.listener.transaction.enabled` is `true`.  I was not sure if this is the best place to add that logic but it seemed better here than in the properties mapper.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesTests.java
```diff
@@ -390,8 +400,34 @@ class TemplateProperties {
...
+		}
+
+	}
+
+	@Nested
+	class TransactionProperties {
+
+		@ParameterizedTest
+		@MethodSource
+		void transactionsEnabledTests(boolean listenerEnablesTransactions, boolean templateEnablesTransactions,
```
onobc, line 420: I am on the fence here w/ the use of `@ParameterizedTest` w/ 4 items in the test arguments stream provider vs. having 4 individual (yet repetitive) tests methods (feels a bit ""six of one, half a dozen of the other"). 

On one hand, the 4 individual test methods makes it clear what is being tested. And to get that same clarity w/ this impl we have to add the test name as arg0 and then do `@ParameterizedTest(name = "{0}")`. 
On the other hand, this current impl reduces the 4 tests that are about 99% the same. 🤷🏻

onobc: Yeh, I think the value for doing this for 4 cases does not outweigh having 4 concise test methods (delegating to a single test impl method). I will adjust.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
```diff
@@ -64,6 +65,10 @@ void customizeClientBuilder(ClientBuilder clientBuilder, PulsarConnectionDetails
 		map.from(properties::getConnectionTimeout).to(timeoutProperty(clientBuilder::connectionTimeout));
 		map.from(properties::getOperationTimeout).to(timeoutProperty(clientBuilder::operationTimeout));
 		map.from(properties::getLookupTimeout).to(timeoutProperty(clientBuilder::lookupTimeout));
+		if (this.properties.getTemplate().getTransaction().isEnabled()
+				|| this.properties.getListener().getTransaction().isEnabled()) {
```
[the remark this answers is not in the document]
onobc: Ha, this usage was the reason behind creating the convenience method but I forgot to push the commit that adjusted its usage.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -866,6 +888,54 @@ public void setObservationsEnabled(boolean observationsEnabled) {
...
+
+		/**
+		 * Whether the component supports transactions.
+		 */
+		private boolean enabled;
+
+		/**
+		 * Whether the component requires transactions.
+		 */
+		private boolean required;
```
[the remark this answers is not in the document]
onobc: Yep, similar to `if equals then hashcode` this is `if required then enabled`. I will add validation to ensure these do not get out of order.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
```diff
@@ -157,6 +162,14 @@ <T> void customizeProducerBuilder(ProducerBuilder<T> producerBuilder) {
 		map.from(properties::getAccessMode).to(producerBuilder::accessMode);
 	}
 
+	<T> void customizeTemplate(PulsarTemplate<T> template) {
```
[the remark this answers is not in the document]
onobc: > Will bad things happen if we accidentally call customizeClientBuilder without customizeTemplate as transactions would then be partially configured?

In that case, no harm is done if you enable txn on the client but do not use them. However, If we instead called `customizeTemplate` but forgot to call `customizeClientBuilder` then the code would bark at runtime when attempting to create a transaction. 

Moving to a single customizeTransactions method would solve that but we would also have to add in the configuration for the PulsarContainerProperties which does something similar. I will see what that looks like and follow up either w/ that code change or my reasoning why not.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
onobc: Done

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesTests.java
onobc: Done

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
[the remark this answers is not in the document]
onobc: Done

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -866,6 +888,54 @@ public void setObservationsEnabled(boolean observationsEnabled) {
...
+		public Transaction getTransaction() {
+			return this.transaction;
+		}
+
+	}
+
+	public static class Transaction {
+
+		/**
+		 * Whether the component supports transactions.
```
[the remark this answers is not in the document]
onobc: Done

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -47,7 +48,7 @@
  * @since 3.2.0
  */
 @ConfigurationProperties("spring.pulsar")
-public class PulsarProperties {
+public class PulsarProperties implements InitializingBean {
```
wilkinsona, line 51: Can we do this differently, please? There are no other `@ConfigurationProperties` classes in Boot's codebase that implement `InitializingBean`. A common place for it is in the consuming auto-configuration.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -936,6 +943,11 @@ public void setTimeout(Duration timeout) {
 			this.timeout = timeout;
 		}
 
+		void validate(String prefix) {
+			Assert.state((this.enabled || !this.required), "If transactions are required "
```
wilkinsona, line 947: `InvalidConfigurationPropertyValueException` should be used to report this sort of problem.

wilkinsona (changes requested):

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
onobc: Absolutely. I got pulled away before I could add a self-review comment asking if this is the recommended approach for cross-field validation. I found it in the following (but was skeptical still):

- OAuth2ClientProperties
- DataSourceProperties
- OAuth2AuthorizationServerProperties

I will adjust.

wilkinsona: Sorry, I failed to find those when searching earlier. On balance I think I still prefer it being done in the auto-configuration, although I feel less strongly about it now.

onobc: No worries. I'm about to push a commit, I think this is much cleaner. It did have an eyebrow raising effect in its current incantation. :)

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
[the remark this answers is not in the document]
onobc: Looking into this further, it gets a bit complicated trying to configure the client inline w/ the template (as well as the listener). By the time the client gets to the template (or listener) it has already been built and the builder has already been customized. Also, the client is required by all the things and as such it will always have customizeClient called.  

However, if you feel strongly about this I can spend a bit more time trying a few things.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -943,9 +937,39 @@ public void setTimeout(Duration timeout) {
...
+		 */
+		protected abstract String propertyPath();
+
+	}
+
+	static class TemplateTransaction extends Transaction {
+
+		@Override
+		protected String propertyPath() {
+			return "spring.pulsar.template.transaction";
```
onobc, line 963: This avoids callers needing to know internals about where the property is mapped.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
onobc: Done

wilkinsona (changes requested):
I've [rebased, squashed, and polished](https://github.com/wilkinsona/spring-boot/tree/gh-40189) the changes. While doing so, I had a bit more time to think about the `.enabled` and `.required` properties that can be configured in a way that's invalid. I still don't like it, I'm afraid.

With the `.enabled` and `.required` properties, it feels like one configuration option has been split across two settings.


I wonder if we could have a single property backed by an enum instead. The enum's values, something like `DISABLED`, `SUPPORTS`, and `REQUIRES`, would map to the current properties like this:

| | `.enabled` | `.required` |
| ----- | ----------- | ----------- |
| **`DISABLED`** | `false` | `false` |
| **`SUPPORTS`** | `true` | `false` |
| **`REQUIRES`** | `true` | `true` |
| | `false` | `true` |

With no value for the invalid `.enabled=false` `.required=true` combination, we've hopefully made things easier to configure. I wonder if Spring Pulsar would also benefit from such a change as it too has `enabled` and `required` properties. If so, it could then also provide the `enum` that Boot exposes as a configuration property.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
wilkinsona: The `protected` method will be available to users who sub-class `PulsarProperties`. I would prefer that we keep the surface area of the API as small as possible so I've reworked this and it's now all encapsulated in the mapper.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
wilkinsona: As this was only used by the mapper, I've moved it there. Even with it being package-private, I think it reads better if the code's closer to where it's used.

onobc:
> properties that can be configured in a way that's invalid. I still don't like it, I'm afraid.

Not a problem. Thanks for the focus on this. This is the perfect time to get it right (before people start using it). 

> With the .enabled and .required properties, it feels like one configuration option has been split across two settings.

Agreed. 


> I wonder if we could have a single property backed by an enum instead. The enum's values, something like DISABLED, SUPPORTS, and REQUIRES, would map to the current properties like this:

I like this approach. 

> With the .enabled and .required properties, it feels like one configuration option has been split across two settings.

Absolutely.

Based on the fact that we will want the enum to exist in Spring Pulsar and it has already cut the RC1 release, are you ok if we get this in during RC phase (like early next week) ?

philwebb:
> Based on the fact that we will want the enum to exist in Spring Pulsar and it has already cut the RC1 release, are you ok if we get this in during RC phase (like early next week) ?

I don't think we should attempt to add this to Spring Boot 3.3 if we can't get something in before our RC1. Perhaps I can take a look today and add the enum on our side then just make a minor amendment post RC to migrate the enum. That will at least give folks a chance to try it before we GA.

onobc:
> Perhaps I can take a look today and add the enum on our side then just make a minor amendment post RC to migrate the enum. That will at least give folks a chance to try it before we GA.

That would be excellent @philwebb . If you can't get to it, let me know and I would be happy to add it to the Boot side as well. I am already in progress on adding it to the Spring Pulsar side (which we would make the minor amendment post RC to switch the enum that is used).

PR #40193, opened 2024-04-08
Enforce using `List.of(...)` instead of `Arrays.asList(...)`

quaff:
See https://github.com/spring-projects/spring-boot/pull/40181

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/JavaConventions.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/JavaConventions.java
@@ -16,7 +16,6 @@
 
 package org.springframework.boot.build;
 
-import java.util.Arrays;
 import java.util.Collections;
 import java.util.List;
 import java.util.Map;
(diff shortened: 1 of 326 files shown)


bclozel:
`List.of(...)` does not allow `null` elements and is immutable, unlike `Arrays.asList(...)`. Are you sure all those changes are in line with this? As for the case of heap pollution, I don't see where this happens in our codebase (we don't cast lists as far as I can see) nor how `List.of` prevents the same.

quaff:
> `List.of(...)` does not allow `null` elements and is immutable, unlike `Arrays.asList(...)`.

They are same except that `Arrays.asList(...)` does accept `null` and allow `set(int index, E element)`, that is rarely used and error prone we should avoid.

wilkinsona:
Thanks for the PR but we considered such a change in the past and [decided against it](https://github.com/spring-projects/spring-boot/issues/28174).

PR #40235, opened 2024-04-09
Loading of custom deny-all filter can cause a StackOverflowError when deploying to Tomcat with Log4j2 configured to use a single JVM-wide logger context

ppkarwasz:
This PR replaces the custom "deny all" filter used by `Log4J2LoggingSystem` with the standard [`DenyAllFilter`](https://logging.apache.org/log4j/2.x/javadoc/log4j-core/org/apache/logging/log4j/core/filter/DenyAllFilter), which is available since Log4j Core 2.13.0.

The purpose of this change is:

- to simplify the code base,
- to work around a pesky interaction between the custom Spring Boot filter and Tomcat's classloader (cf. apache/logging-log4j2#1430). If Log4j Core is configured to use a **single** logger context for the whole JVM, Spring Boot will add its custom filter implementation (contained in Spring Boot's classloader) to the global logging system. Since Tomcat's classloader performs logging calls, this will cause a `StackOverflowException` while attempting to resolve the filter class. Using a class from Log4j Core (which must be in the system classloader in this configuration) will work around the problem.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/Log4J2LoggingSystem.java
@@ -31,9 +31,7 @@ import java.util.logging.Handler;
 
 import org.apache.logging.log4j.Level;
 import org.apache.logging.log4j.LogManager;
-import org.apache.logging.log4j.Marker;
 import org.apache.logging.log4j.core.Filter;
-import org.apache.logging.log4j.core.LogEvent;
 import org.apache.logging.log4j.core.Logger;
 import org.apache.logging.log4j.core.LoggerContext;
 import org.apache.logging.log4j.core.config.AbstractConfiguration;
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks very much, @ppkarwasz.

ppkarwasz:
You're welcome, @wilkinsona.

There are still some small adjustments to the `Log4j2LoggingSystem` I would like to contribute in preparation for the upcoming Log4j Core 3.x release (e.g. a fix of apache/logging-log4j2#1799). Any chance we can make it in time for Spring Boot 3.3.0? Or should we rather target Spring Boot 3.4.0?

**Remark**: Personally I would prefer to keep the current `2.x` version of `log4j-to-slf4j` in Log4j 3.x instead of releasing a `3.0.0` for that artifact (see [this thread on `dev@logging`](https://lists.apache.org/thread/8w3y3sj9lmv857y7mtbf5b7tsyyfn5nq)). This would certainly ease the upgrade of the `log4j2.version` property to `3.0.0`, since it would not change anything in the default `spring-boot-starter-logging` artifact.

wilkinsona:
3.3.0-RC1 is scheduled for next week. Depending on the nature of the changes that means we may only have a week to make any further changes. If it's a bug, we have longer as we can make the fix post RC1 or even in a subsequent maintenance release.

PR #40236, opened 2024-04-09
As of Jetty12, Graceful shutdown is broken.

yokotaso:
Since Jetty12, StatisticHandler is no longer to check shutdown state. JettyServer failed to count active request count correctly. Insted of StatisticHandler, Jetty12 introduce GracefulHandler.

Spring-Boot uses StaticstaticHandler as a active request counter in Jetty11.
It was ok because StaticstaticHandler count active request count in acount of shutdown state internally.

https://github.com/jetty/jetty.project/blob/jetty-11.0.x/jetty-server/src/main/java/org/eclipse/jetty/server/handler/StatisticsHandler.java#L146-L168

But As of Jetty12 StaticstaticHandler doesn't see shutdown state.
https://github.com/jetty/jetty.project/blob/jetty-12.0.x/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/StatisticsHandler.java#L70-L74

We can use GracefulHandler instead of StaticsticHandler and it'll success to count
active request count on shutdown phase.
https://github.com/jetty/jetty.project/blob/jetty-12.0.x/jetty-core/jetty-server/src/main/java/org/eclipse/jetty/server/handler/GracefulHandler.java#L89-L116

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/context/ServletWebServerApplicationContext.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/context/ServletWebServerApplicationContext.java
@@ -170,12 +170,11 @@ public class ServletWebServerApplicationContext extends GenericWebApplicationCon
 	protected void doClose() {
 		if (isActive()) {
 			AvailabilityChangeEvent.publish(this, ReadinessState.REFUSING_TRAFFIC);
-			super.doClose();
-
-			WebServer webServer = this.webServer;
-			if (webServer != null) {
-				webServer.destroy();
-			}
+		}
+		super.doClose();
+		WebServer webServer = this.webServer;
+		if (webServer != null) {
+			webServer.destroy();
 		}
 	}
 
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/context/ServletWebServerApplicationContext.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/context/ServletWebServerApplicationContext.java
@@ -170,11 +170,12 @@ public class ServletWebServerApplicationContext extends GenericWebApplicationCon
 	protected void doClose() {
 		if (isActive()) {
 			AvailabilityChangeEvent.publish(this, ReadinessState.REFUSING_TRAFFIC);
-		}
-		super.doClose();
-		WebServer webServer = this.webServer;
-		if (webServer != null) {
-			webServer.destroy();
+			super.doClose();
+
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/context/ServletWebServerApplicationContext.java
```diff
@@ -170,11 +170,12 @@ protected void onRefresh() {
...
-		super.doClose();
-		WebServer webServer = this.webServer;
-		if (webServer != null) {
-			webServer.destroy();
+			super.doClose();
+
+			WebServer webServer = this.webServer;
+			if (webServer != null) {
+				webServer.destroy();
+			}
```
wilkinsona, line 178: This change will affect every embedded web server, not just Jetty. It either shouldn't be made at all, or if it's really needed, it should be made separately. Please revert it in this PR.

wilkinsona (changes requested):
Can you please explain how you identified the need for this change? We have integration tests, such as [this one](https://github.com/spring-projects/spring-boot/blob/1dd823b5495bc1720e35bbf2e7beca455bfc59a9/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/servlet/server/AbstractServletWebServerFactoryTests.java#L1183-L1202), that verify the graceful shutdown of all embedded web servers, including Jetty. Either that test is faulty and should be corrected or these changes are not needed.

yokotaso:
Sorry, it is my misunderstanding that Graceful shutdown is broken.
I've investegated cause of curious log happend in my production environment. and Log content is like bellow:

```
handleException /actuator/prometheus org.springframework.web.context.request.async.AsyncRequestNotUsableException: ServletOutputStream failed to write: java.util.concurrent.TimeoutException: Idle timeout expired: 30000/30000 ms
```
```
handleException /actuator/prometheus java.io.IOException: java.util.concurrent.TimeoutException: Idle timeout expired: 30000/30000 ms
```
In order to understand cause of log, 
I have read jetty and spring-boot codes, and found StatiticsHandler is no longer check state of Shutdown.
So I thought it is better to use GracefulHandler instead of StatisticHandler in GracefulShutdown.

wilkinsona:
Thanks for the additional details.

Spring Boot's graceful shutdown doesn't rely on the statistics handler being aware of the shutdown state. It only needs its count of the currently active requests. It may be that we could achieve the same with `GracefulHandler` but I'm not aware of anything that would mean that it will improve graceful shutdown with Jetty or indeed that it will fix the problem that you're seeing. As such, I'm going to decline this change as I don't think we should make it without a strong reason to do so.

PR #40384, opened 2024-04-17
Upgrade to Spring Integration v 6.3.0-M2

dev-yugantar:
Fix for #40226


wilkinsona:
Thanks for the PR but we need to upgrade to RC1 (that has not yet been released) rather than M2. We also prefer to handle these one-line dependency upgrades ourselves using some automated tooling.

PR #40484, opened 2024-04-23
Add a customizer for Lettuce's ClientOptions

SOOHYUN-LIM:
The current LettuceConnectionConfiguration requires too much code to be written when additional settings such as Cluster NodeFilter are needed. Therefore, to enhance scalability, I have added a ClientOptions Customizer.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/ClientOptionsBuilderCustomizer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/ClientOptionsBuilderCustomizer.java
@@ -1,0 +1,37 @@
+/*
+ * Copyright 2012-2023 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks for the proposal.

This would leave us with the existing `LettuceClientConfigurationBuilderCustomizer` that can customize an entire `LettuceClientConfigurationBuilder` and the new `ClientOptionsBuilderCustomizer` that can be used to customize the builder that creates the client configuration's `ClientOptions`. Having a second customizer that targets a subset of another customizer is unusual and I'm not 100% sure what we should do. We could:

1. Keep things as proposed with some additions to the javadoc of `LettuceClientConfigurationBuilderCustomizer` to draw attention to `ClientOptionsBuilderCustomizer` and vice versa
2. Maybe add a default method to the existing `LettuceClientConfigurationBuilderCustomizer` for customization of the client options. This may cause problems with the use of method references.
3. Leave things as they are but this would require duplicating quite a bit of code as, particularly when using SSL, creating the client options is quite involved


1 is probably the best option but I'd like to discuss with the team.

philwebb:
WE discussed this as a team and decided that option 1 is the best way to go. We'd like to rename `ClientOptionsBuilderCustomizer` to `LettuceClientOptionsBuilderCustomizer`. We can take care of that when we merge the PR.

wilkinsona:
@SOOHYUN-LIM Thanks very much for making your first contribution to Spring Boot.

PR #40490, opened 2024-04-23
Use SpringBootPlugin constants for task name

chaewss:
Update task names to use constants from SpringBootPlugin.


wilkinsona:
Thanks for the suggestion. Unfortunately, we don't want to make this change as it introduces a cycle between the `org.springframework.boot.gradle.plugin` and `org.springframework.boot.gradle.dsl` packages.

PR #40493, opened 2024-04-23
Pulsar auth parameters don't properly encode JSON values

### LINKED ISSUE #40492: PulsarPropertiesMapper.getAuthenticationParamsJson
@onobc 
[PulsarPropertiesMapper.getAuthenticationParamsJson](https://github.com/spring-projects/spring-boot/blob/6e37567a9d9932fdecd8ace18bb28a570e0827a9/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java#L135)  does not return a proper Json encoded in String
after this change https://github.com/spring-projects/spring-boot/commit/41ed4d6cf4c5e316f236c99ce59cf7ddfa48ee89 because values are not properly encoded.

onobc:
The values in the `spring.pulsar.client.authentication.param` config props map are not currently JSON encoded. For simple values this is fine. However, some custom auth modules may require more complex parameter values that may contain special characters that results in invalid JSON. This commmit encodes the parameter values using a very simple hand-rolled escape function.

Fixes #40492

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
@@ -137,12 +137,23 @@ final class PulsarPropertiesMapper {
 		try {
 			return sortedParams.entrySet()
 				.stream()
-				.map((entry) -> "\"%s\":\"%s\"".formatted(entry.getKey(), entry.getValue()))
+				.map((entry) -> "\"%s\":\"%s\"".formatted(entry.getKey(), escapeJson(entry.getValue())))
 				.collect(Collectors.joining(",", "{", "}"));
 		}
 		catch (Exception ex) {
 			throw new IllegalStateException("Could not convert auth parameters to encoded string", ex);
 		}
+	}
+
+	private String escapeJson(String raw) {
+		return raw.replace("\\", "\\\\")
+			.replace("\"", "\\\"")
+			.replace("/", "\\/")
+			.replace("\b", "\\b")
+			.replace("\t", "\\t")
+			.replace("\n", "\\n")
+			.replace("\f", "\\f")
+			.replace("\r", "\\r");
 	}
 
 	<T> void customizeProducerBuilder(ProducerBuilder<T> producerBuilder) {
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapperTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapperTests.java
@@ -81,8 +81,10 @@ class PulsarPropertiesMapperTests {
 	@Test
 	void customizeClientBuilderWhenHasAuthentication() throws UnsupportedAuthenticationException {
 		PulsarProperties properties = new PulsarProperties();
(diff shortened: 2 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
```diff
@@ -137,14 +137,25 @@ private String getAuthenticationParamsJson(Map<String, String> params) {
...
-				.map((entry) -> "\"%s\":\"%s\"".formatted(entry.getKey(), entry.getValue()))
+				.map((entry) -> "\"%s\":\"%s\"".formatted(entry.getKey(), escapeJson(entry.getValue())))
 				.collect(Collectors.joining(",", "{", "}"));
 		}
 		catch (Exception ex) {
 			throw new IllegalStateException("Could not convert auth parameters to encoded string", ex);
 		}
 	}
 
+	private String escapeJson(String raw) {
```
onobc, line 148: Because we do not directly bring in a JSON library (e.g. Gson or Jackson) we can not use the simple facilities that these libs have to "encode a JSON string". 

Here were the options:

## Use JSON lib
Add required dependency on Jackson or Gson and use their built-in methods

- ✅ solid encode impl maintained by JSON lib
- ❌ ❌ ❌ have to bring in direct dep on JSON lib and this forces it upon our users (not an option)
 
## Add feature to Spring Boot JsonParser
Add "encode a JSON string" to our `JsonParser` abstraction

- ✅ can be used in other areas of Spring Boot codebase
- ❌ created and maintained by us (more work)
- ❌ impl no more solid than if it were a private impl inside the mapper
- ❌ nowhere else in the Boot codebase needs this AFAIK
 
## Hand-roll basic encoder
Inline a simple encode JSON facility in the mapper.

- ❌ basic impl maintained by us
- ❌ ✅ not a perfect solution but will solve the majority of the cases
- ✅ ✅ no direct dep on JSON lib
- ✅ private impl localized to the mapper

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapperTests.java
```diff
@@ -81,8 +81,10 @@ void customizeClientBuilderWhenHasNoAuthentication() {
 	@Test
 	void customizeClientBuilderWhenHasAuthentication() throws UnsupportedAuthenticationException {
 		PulsarProperties properties = new PulsarProperties();
-		Map<String, String> params = Map.of("param", "name");
-		String authParamString = "{\"param\":\"name\"}";
+		Map<String, String> params = Map.of("simpleParam", "foo", "complexParam",
```
onobc, line 84: We are not exhausting all [possible replacements characters](https://github.com/spring-projects/spring-boot/pull/40493/files#diff-5bb3152ef3fa91e2d911e7c908848b632a31bfd8f298b76536cbbc56c25a1891R149) **but** are covering the majority. 

@philwebb should we add a bit more coverage here?

mhalbritter:
Thanks Chris!

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapperTests.java
mhalbritter: I'm not Phil, but I think this is fine :)

PR #40524, opened 2024-04-25
Remove unnecessary null check for expectedDeprecationMessages

chaewss:
Removed the unnecessary null check for `expectedDeprecationMessages` as it's already initialized as an ArrayList and cannot be null.


mhalbritter:

PR #40554, opened 2024-04-27
Add support for AstraDB

michaelsembwever:
autoconfigure properties for a CassandraVectorStore cql connection to AstraDB

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/astra/AstraConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/astra/AstraConfiguration.java
@@ -1,0 +1,49 @@
+/*
+ * Copyright 2024 - 2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.boot.autoconfigure.astra;
+
+import java.nio.file.Path;
+
+import com.datastax.oss.driver.api.core.CqlSession;
+
+import org.springframework.boot.autoconfigure.AutoConfiguration;
+import org.springframework.boot.autoconfigure.cassandra.CassandraAutoConfiguration;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.boot.autoconfigure.cassandra.CqlSessionBuilderCustomizer;
+import org.springframework.context.annotation.Bean;
+
+
+/**
+ * @author Cédrick Lunven
+ * @author Mick Semb Wever
+ * @since 3.2.6
+ */
+@AutoConfiguration(before = CassandraAutoConfiguration.class)
+@ConditionalOnClass({CqlSession.class})
+@EnableConfigurationProperties(AstraExtraSettings.class)
+public class AstraConfiguration {
+
+    @Bean
+    public CqlSessionBuilderCustomizer sessionBuilderCustomizer(AstraExtraSettings astraSettings) {
+        if (null != astraSettings.getSecureConnectBundle()) {
+            Path bundle = astraSettings.getSecureConnectBundle().toPath();
+            return builder -> builder.withCloudSecureConnectBundle(bundle);
+        }
+        return (cqlSessionBuilder) -> {/* noop */};
+    }
+
+}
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the proposal. Please note that as this would be a new feature it cannot be added to a 3.2.x maintenance release. The earliest that it could be added would now be Spring Boot 3.4 as we've just started the RC phase for 3.3 and the feature set is frozen.

If we want a configuration property for the cloud secure connect bundle then, rather than introducing new `spring.astra.*` properties for it, I think it would be better for it to sit alongside existing properties that are applied to the `CqlSessionBuilder` that we already auto-configure. For example, `spring.cassandra.keyspace-name` is applied to the `CqlSessionBuilder` using `withKeyspace`. Similarly, a new `spring.cassandra.cloud-secure-connect-bundle` property could be applied to the `CqlSessionBuilder` using `withCloudSecureConnectBundle`.

michaelsembwever:
> Thanks for the proposal. Please note that as this would be a new feature it cannot be added to a 3.2.x maintenance release. The earliest that it could be added would now be Spring Boot 3.4 as we've just started the RC phase for 3.3 and the feature set is frozen.

This is your decision, but it's a trivial addition… 🤷 

> If we want a configuration property for the cloud secure connect bundle then, rather than introducing new spring.astra.* properties for it, I think it would be better for it to sit alongside existing properties that are applied to the CqlSessionBuilder that we already auto-configure. For example, spring.cassandra.keyspace-name is applied to the CqlSessionBuilder using withKeyspace. Similarly, a new spring.cassandra.cloud-secure-connect-bundle property could be applied to the CqlSessionBuilder using withCloudSecureConnectBundle.

It can't be part of the existing Cassandra connections.  This only works with AstraDB.  AstraDB is CQL compatible with Cassandra, but not vice versa.

I've no objection to putting the classes into the `org.springframework.boot.autoconfigure.cassandra` package.
But a property `spring.cassandra.cloud-secure-connect-bundle` definitely implies the wrong thing.
What about ` spring.cassandra.astra.cloud-secure-connect-bundle` instead ?

wilkinsona:
> I've no objection to putting the classes into the `org.springframework.boot.autoconfigure.cassandra` package.
But a property `spring.cassandra.cloud-secure-connect-bundle` definitely implies the wrong thing.

Generally speaking, when naming and structuring properties, we try to take our lead from the names and structure of the API that the properties are configuring. In this case the Astra-specific `withCloudSecureConnectBundle` method's on `SessionBuilder` alongside several other general Cassandra methods. Why does the API of `SessionBuilder` not also imply the wrong thing? If there were several Astra properties, grouping them together beneath `spring.cassandra.astra` would be quite compelling. With only one, deviating from the structure of the API we're using is less appealing.

wilkinsona:
> Putting it alongside the other cassandra options would be very confusing (and likely create errors) for cassandra users.

Then why has the API of `SessionBuilder`/`CqlSessionBuilder` been modelled as it has? Is it not very confusing for the general purpose `CqlSessionBuilder` API to contain a method that's Astra specific? If not, what is it about Spring Boot that would make a similarly structured property name confusing?

What I'm trying to understand is why the structure of our properties should different from the structure of the API in the Cassandra Driver's Java API.

michaelsembwever:
The secureConnectBundle works as part of existing cql connections and drivers. 

This information does go into the `CqlSessionBuilder`, but it breaks a vanilla cql connection.

> What I'm trying to understand is why the structure of our properties should different from the structure of the API in the Cassandra Driver's Java API.

The java-driver was only recently donated to the ASF.  When it was at DataStax then proprietary features like this could be put directly into it.  Now it's at the ASF these features will be removed over time – and then you'll see a `AstraCqlSessionBuilder` (i'd envision this would eventuate as most of the `org.springframework.boot.autoconfigure.cassandra` package being copied to `org.springframework.boot.autoconfigure.astra` – and that explains my original preference to that latter package :) )

philwebb:
> Now it's at the ASF these features will be removed over time – and then you'll see a AstraCqlSessionBuilder.

Given this fact, I think we should hold off on adding direct support to Spring Boot at this time. It looks like the existing `org.springframework.boot.autoconfigure.cassandra.CqlSessionBuilderCustomizer` should suffice for now.

I think we should consider adding support when the `AstraCqlSessionBuilder` has been added.

Thanks anyway for the PR @michaelsembwever

michaelsembwever:
>  I think we should hold off on adding direct support to Spring Boot at this time.

That could be years away.

Take a look at this PR in action here:  https://github.com/datastax/ai-agent-java

With the application.properties it is very easy to switch from using Apache Cassandra to using AstraDB, just by adding a few properties.

https://github.com/datastax/ai-agent-java/blob/workshop-step-1/src/main/resources/application.properties 

This layout of property name spaces makes sense to me.

philwebb:
Thanks for the additional info @michaelsembwever.  We discussed this again but still feel like it's a little early to consider adding support directly in Spring Boot. It looks like most folks should be able to use the `CqlSessionBuilderCustomizer` with a relatively small amount of code to get AstraDB working.

We'd like to wait a while to see how things play out and how much demand there is.

michaelsembwever:
FTR, the aligned the suggested patch as closely to the feedback provided and to my understanding of how you would want it here: 
https://github.com/spring-projects/spring-boot/compare/main...michaelsembwever:spring-boot:mck/astradb-autoconfigure-0 

That puts it under the same package as cassandra, i.e. `org.springframework.boot.autoconfigure.cassandra`
 and puts the property under `spring.cassandra.astra` so that it's aligned and a subset of `spring.cassandra`.


>  It looks like most folks should be able to use the CqlSessionBuilderCustomizer with a relatively small amount of code to get AstraDB working.

From our experience spring-boot users have absolutely no idea where to start.  Spring boot autoconfigure and the existing cassandra classes in it are not in the slightest intuitive.  I'd be surprised if you're not getting the feedback from this demographic of users…

Issue #40010, opened 2024-03-18
Add customizers to R2DBC ProxyConnectionFactory / ConnectionFactoryDecorator


pauljohe-dnb:
Would it be possible to make it easy to apply customizers to R2DBC observability?

For example, consider this fuctional interface:

```java
@FunctionalInterface
public interface ProxyConnectionFactoryBuilderCustomizer {
    void customize(ConnectionFactory connectionFactory, ProxyConnectionFactory.Builder builder);
}
```

The current listener `org.springframework.boot.actuate.autoconfigure.r2dbc.R2dbcObservationAutoConfiguration` could be re-written:

```java
@Bean
ConnectionFactoryDecorator connectionFactoryDecorator(ObjectProvider<ProxyConnectionFactoryBuilderCustomizer> customizers) {
    return (connectionFactory) -> {
        var builder = ProxyConnectionFactory.builder(connectionFactory);
        customizers.forEach(customizer -> customizer.customize(connectionFactory, builder));
        return builder.build();
    };
}

@Bean
@ConditionalOnBean(ObservationRegistry.class)
ProxyConnectionFactoryBuilderCustomizer connectionObserver(R2dbcObservationProperties properties,
                                                           ObservationRegistry observationRegistry,
                                                           ObjectProvider<QueryObservationConvention> queryObservationConvention,
                                                           ObjectProvider<QueryParametersTagProvider> queryParametersTagProvider) {
    return (connectionFactory, builder) -> {
        HostAndPort hostAndPort = extractHostAndPort(connectionFactory);
        ObservationProxyExecutionListener listener = new ObservationProxyExecutionListener(observationRegistry,
                connectionFactory, hostAndPort.host(), hostAndPort.port());
        listener.setIncludeParameterValues(properties.isIncludeParameterValues());
        queryObservationConvention.ifAvailable(listener::setQueryObservationConvention);
        queryParametersTagProvider.ifAvailable(listener::setQueryParametersTagProvider);
        builder.listener(listener);
    };
}
```

With this pattern, it is very easy for other applications to apply their own listeners and customizations to the `ProxyConnectionFactory` initalization.

philwebb:
Thanks for the suggestion @pauljohe-dnb. We generally like the idea of a customizer, although having one that accepts two parameters isn't that common.
 
[...]

PR #40555, opened 2024-04-28
Add a customizer for ProxyConnectionFactory.Builder from r2dbc-proxy

ttddyy:
With upgrading to `r2dbc-proxy 1.1.5.RELEASE`(should be addressed by the dependency upgrade commits), this could be a change mentioned in https://github.com/spring-projects/spring-boot/issues/40010

NOTE:  
Now the `R2dbcObservationAutoConfiguration` is more generic to r2dbc-proxy. It could be renamed to `R2dbcProxyAutoConfiguration`.


mhalbritter:
Thanks Tadaya! I've split the auto-configurations in two: one for the r2dbc proxy, and left the observation bits in `R2dbcObservationAutoConfiguration`.

PR #40588, opened 2024-05-01
Document default value of management.zipkin.tracing.encoding

izeye:
This PR adds a configuration properties metadata entry for the default value of the `management.zipkin.tracing.encoding`.

See gh-39049


mhalbritter:
Thanks!

PR #40590, opened 2024-05-01
Remove reassignments for builders in GraphQL auto-configurations

izeye:
This PR removes reassignments for builders in GraphQL auto-configurations as it doesn't seem to be necessary.


philwebb:
Thanks @izeye!

PR #40604, opened 2024-05-02
Clarify devtools restart class loader

dsyer:
I read the section in the user guide and it helpfully suggested setting spring.devtools.restart.enabled=false to eliminate possible sources of failure. And that fixed it. But now I want to switch it back on, and I don't know how to craft the right configuration.

This change would help.


mhalbritter:
Thanks Dave!

PR #40611, opened 2024-05-02
Avoid using ClientHttpRequestFactory which may not on the classpath

snicoll:
This commit fixes a test that uses a different `ClientHttpRequestFactory` implementation for the sole purpose of checking that it's not overridden. Previously, this used the Jetty implementation but jetty is not on the classpath so that might fail if the factory attempts to manipulate the underlying client.

This broke this test when testing against Spring Framework 6.2 and the purpose of this PR is to fix the test and limit the amount of change in upgrading to 6.2


mhalbritter:
Thanks Stéphane!

PR #40617, opened 2024-05-03
Add How-To for ManagedClassNameFilter

snicoll:
This commit documents the support of ManagedClassNameFilter that was added in #39813.


scottfrederick:
Thanks @snicoll!

PR #40619, opened 2024-05-03
added http health indicator

rohitp-a:


wilkinsona:
Thanks for the proposal. This has already been considered and declined in https://github.com/spring-projects/spring-boot/pull/12972. I don't think anything has changed since then and we still recommend using a service-specific check. Thanks anyway for the PR.

Issue #38763, opened 2023-12-13
Add property to specify Docker Compose startup and shutdown flags


mhalbritter:
@dsyer ran into an issue where a Postgres update prevented the start of the Postgres container used for integration tests. Adding a `--force-recreate --renew-anon-volumes` would fix that, but as of now, there's no possibility to add additional flags to the `docker compose up` call.

habuma:
Similarly, there's no way (at least not that I see) to specify options for Docker Compose "down". For example, when I have a Docker Compose file that stands up a Chroma database in Docker, it also creates a volume. But when the app stops, only the container goes away, not the volume. `docker compose down --volumes` takes the volume down at the command line, but it would be useful to be able to have it remove the volume when the app stops.

wilkinsona:
https://github.com/spring-projects/spring-boot/issues/41086 is requesting the ability to set `-p/--project-name`.

osopromadze:
Hi @wilkinsona, can I work on this issue?

wilkinsona:
Thanks for the offer, @osopromadze, but we're not ready to start work on this one, hence the `status: pending-design-work` label.

mhalbritter:
Changes are here and ready to be merged as soon as we have the `3.3.x` branch: https://github.com/mhalbritter/spring-boot/tree/mh/38763-add-property-to-specify-docker-compose-startup-and-shutdown-flags

ozozgun:
Hello. Thanks for this great feature. Is there any documentation on this ?

bclozel:
@ozozgun it's called out [in the appendix here](https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.docker-compose.spring.docker.compose.arguments), but it's not mentioned [in the relevant section](https://docs.spring.io/spring-boot/reference/features/dev-services.html#features.dev-services.docker-compose.lifecycle). Feel free to create a new issue asking for more documentation for this particular feature.

wilkinsona:
We also seem to be missing some docs for `spring.docker.compose.arguments`. I've opened https://github.com/spring-projects/spring-boot/issues/48657.

PR #40634, opened 2024-05-07
added extra start up flags

rohitp-a:


wilkinsona:
I assume, judging from the branch name, that this is intended to address #38763. Unfortunately, it doesn't really do that as there's no way for a user to provide the flags that should be passed to `up`. Thanks for the PR, but I think we need to do some design work on #38763 before we can consider any code changes. For example, we need to consider how the user will specify the flags (application properties?) and whether we need to support flags only for `up` or also for one or more of `down`, `start` and `stop` as well.

PR #40695, opened 2024-05-11
Add support for org.testcontainers.kafka.KafkaContainer

eddumelendez:
Testcontainers 1.19.8 provides `org.testcontainers.kafka.KafkaContainer`, which relies on `apache/kafka` image.


wilkinsona:
Thanks, @eddumelendez. A few notes and things for us to think about prior to merging this (plenty of time as this is just too late for 3.3 unfortunately):

 - Should we rename our existing `KafkaContainerConnectionDetailsFactory` to `ConfluentKafkaContainerConnectionDetailsFactory`, and similar for the existing `KafkaContainerConnectionDetails`?
 - Should we rename `org.springframework.boot.testsupport.testcontainers.DockerImageNames.kafka()`?

In addition to these, we'll also need to do something about the docs. When we upgrade to Testcontainers 1.19.8, "Containers of type `KafkaContainer`" will become ambiguous as there will be two types named `KafkaContainer`. I've opened #40699 for that. We'll then need to make some further changes here so that it's clear that both `KafkaContainer` types are supported.

eddumelendez:
Hi @wilkinsona,

> Should we rename our existing KafkaContainerConnectionDetailsFactory to ConfluentKafkaContainerConnectionDetailsFactory, and similar for the existing KafkaContainerConnectionDetails?

We are planning to deprecate the existing one under `org.testcontainers.container.KafkaContainer` and provide a Confluent module with `org.testcontainers.confluent.KafkaContainer`. So, with those changes I think we can add a new `ConfluentKafkaContainerConnectionDetailsFactory` and `ConfluentKafkaContainerConnectionDetails` when available and deprecate existing Kafka* implementations in SB.

> Should we rename org.springframework.boot.testsupport.testcontainers.DockerImageNames.kafka()?

Yes.

wilkinsona:

Issue #38923, opened 2023-12-24
Flyway auto-configuration does not work with Flyway 10 when using GraalVM


UbiquitousBear:
I've first seen this when trying to build a native SB image 3.2.1 with `flyway-core:10.4.1`. Flyway Autoconfigure  `NativeImageResourceProviderCustomizer` ([code link](https://github.com/spring-projects/spring-boot/blob/7216d2b031be648763744a371b871fa7c320c722/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/NativeImageResourceProviderCustomizer.java#L33)) creates a custom `org.flywaydb.core.api.ResourceProvider` using a `org.flywaydb.core.internal.scanner.Scanner`.

The constructor for `Scanner` has changed between version 9 and 10, whereby the former accepted more arguments.

The code:

```
Scanner<JavaMigration> scanner = new Scanner<>(JavaMigration.class,
					Arrays.asList(configuration.getLocations()), configuration.getClassLoader(),
					configuration.getEncoding(), configuration.isDetectEncoding(), false, new ResourceNameCache(),
					new LocationScannerCache(), configuration.isFailOnMissingLocations());
```

should be replaced with
```
Scanner<JavaMigration> scanner = new Scanner<>(
     JavaMigration.class,                   // Class<I> implementedInterface
     false,                                            // boolean stream
     new ResourceNameCache(),    // ResourceNameCache resourceNameCache
     new LocationScannerCache(), // LocationScannerCache locationScannerCache
     configuration                              // Configuration configuration [FluentConfiguration provided will do]					
);
```

I can make the change for this in SB, however, how should Flyway 9 be handled?

goverdhan07:
@UbiquitousBear - I was wondering if we could introduce a condition wherein if the flyway version is >= 10 use the newer constructor and if version == 9 use the older constructor that accepts more arguments. let me know what you think?

MazizEsa:
Is it open for contribution? I am interested to work on this.

wilkinsona:
Thanks, @MazizEsa. It's all yours.

MazizEsa:
@wilkinsona which branch I should merge to?

MazizEsa:
Got it @scottfrederick . Thanks!

wilkinsona:
How's it going, @MazizEsa?

MazizEsa:
[...]

PR #40698, opened 2024-05-12
[WIP] Add compatibility for both flyway 9 and 10 for 38923

MazizEsa:
In summary added compatibility with flyway 10 and 9 by using reflection when initializing flyway scanner.
Add mechanism to check the dependency based on the mentioned flyway versions.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/NativeImageResourceProviderCustomizer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/NativeImageResourceProviderCustomizer.java
@@ -16,6 +16,8 @@
 
 package org.springframework.boot.autoconfigure.flyway;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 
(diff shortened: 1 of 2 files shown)


MazizEsa:
@wilkinsona can you have a look. This is a draft, WIP. 
Im wanna check first whether it's in the right direction.
Still need to add checks / conditions on dependency part to cater the different dependencies and package structure when importing (postgres and hsql).

wilkinsona:
Thanks, @MazizEsa. For the default case (Flyway 9), please don't use reflection. Instead, just directly call the Flyway API that we compile against. Only when Flyway 10's detected should reflection be used.

Also, if you can, please build your changes (`./gradlew build` or `./gradlew spring-boot-project:spring-boot-autoconfigure:build`) and fix any problems before pushing.

MazizEsa:
Sure. I'm working from the main branch. Hence flyway 10 is the default. Is
it fine? Or should I work from other branch?

wilkinsona:
As I [said a couple of weeks ago](https://github.com/spring-projects/spring-boot/issues/38923#issuecomment-2081392373), the change is only needed in the `3.2.x` branch. As you've noticed, we've already upgraded to Flyway 10 in the main branch so nothing's needed there. Please use `3.2.x` as the base branch for your changes.

MazizEsa:
Thanks @wilkinsona . My bad. I was old either branch was fine. Will make the changes soon.

PR #40763, opened 2024-05-14
Update StartupEndpoint.java

viniciusgiani:
This version keeps the code clean and readable, making it easier to maintain and understand.


philwebb:
There's a lot of formatting changes which go against our styling rules and I prefer the code as it is. Thanks anyway,

PR #40821, opened 2024-05-19
Add compatibility for both flyway 10 for 38923

MazizEsa:
In summary added compatibility with flyway 10 using reflection when initializing flyway scanner.
Add mechanism to check the dependency based on the mentioned flyway versions.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/NativeImageResourceProviderCustomizer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/NativeImageResourceProviderCustomizer.java
@@ -16,6 +16,8 @@
 
 package org.springframework.boot.autoconfigure.flyway;
 
+import java.lang.reflect.Constructor;
+import java.lang.reflect.InvocationTargetException;
 import java.util.Arrays;
 
 import org.flywaydb.core.api.configuration.FluentConfiguration;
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/NativeImageResourceProviderCustomizer.java
```diff
@@ -24,26 +26,74 @@
...
 	public void customize(FluentConfiguration configuration) {
 		if (configuration.getResourceProvider() == null) {
-			Scanner<JavaMigration> scanner = new Scanner<>(JavaMigration.class,
-					Arrays.asList(configuration.getLocations()), configuration.getClassLoader(),
-					configuration.getEncoding(), configuration.isDetectEncoding(), false, new ResourceNameCache(),
-					new LocationScannerCache(), configuration.isFailOnMissingLocations());
+			Constructor<?> scannerConstructor;
+			Scanner scanner;
+			try {
+				scannerConstructor = ClassUtils.forName("org.flywaydb.core.internal.scanner.Scanner", null)
```
wilkinsona, line 48: I think it would be better to avoid reflection in the default case when using Flyway 9.

One way to do that would be to catch the `NoSuchMethodError` and then fall back to reflection. That fallback could even just assume that it's Flyway 10 and assume that `Scanner` will take 5 constructor arguments.

wilkinsona (changes requested):
Thanks, @MazizEsa.

In addition to my comment about avoiding reflection with Flyway 9. `FlywayAutoConfiguration.FlywayAutoConfigurationRuntimeHints` also needs to be updated to allow a `Scanner` instance to be created reflectively when using Flyway 10. This reflection hint should only be registered when a class that only exists in Flyway 10 is on the classpath.

wilkinsona:
Thank you, @MazizEsa, and congratulations on making your first contribution to Spring Boot.

MazizEsa:
Thanks @wilkinsona for polishing up. My apologies towards the end, I got busier with my work, hence slower in the implementation.

Issue #40768, opened 2024-05-15
Upgrade CI to Docker 26.1.2


spring-builds:
https://download.docker.com/linux/static/stable/x86_64/docker-26.1.2.tgz

wilkinsona:
All being well, we won't need this as we'll complete our GitHub Actions migration with Thursday's 3.3.0 release.

PR #40847, opened 2024-05-21
Upgrade CI to Docker 26.1.2

PiyalAhmed:
Upgrade CI to Docker 26.1.2

Closes: https://github.com/spring-projects/spring-boot/issues/40768

### DIFF
--- a/ci/images/get-docker-url.sh
+++ b/ci/images/get-docker-url.sh
@@ -1,5 +1,5 @@
 #!/bin/bash
 set -e
 
-version="26.1.1"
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, @PiyalAhmed, but as described in the PR template we prefer to deal with one-line version upgrades such as this ourselves. Also, we may just close #40768 as we're moving away from container-based CI on Concourse to GitHub Actions.

PiyalAhmed:
I see, thank you for pointing this out!

PR #40856, opened 2024-05-21
Refactor: Use HTTP_LOCALHOST constant for local URL in DevToolsIntegrationTests

khmgobe:
- Introduced a constant HTTP_LOCALHOST to store the base URL.
- Updated all test methods to use the new constant, improving readability and maintainability.
- Ensured all instances of the local URL are consistent across the test methods.


wilkinsona:
Thanks for the proposal but I find the code easier to read as it's currently written. Given that changes are all in a single file, I don't think it really improves maintainability either. Thanks anyway.

PR #40873, opened 2024-05-22
Refactor: null check to use Objects.isNull for clarity

khmgobe:
Updated the convert method to use Objects.isNull instead of a direct null check for better readability and consistency.


wilkinsona:
Thanks for the proposal but I don't find the code more readable when using`Objects.isNull`. The method's javadoc states that "this method exists to be used as a `java.util.function.Predicate`, `filter(Objects::isNull)`" and I don't think we should be using it elsewhere. I'm also not keen on using `requireNonNullElse` either as I don't think it's an improvement over a ternary operator. Furthermore, if we were going to make this sort of stylistic change, it's the sort of thing that we'd want to discuss and agree on as a team of core maintainers and then apply consistently across the whole codebase.

PR #40899, opened 2024-05-24
Make maxConnection property of Reactor Netty configurable

haydin505:
I was unable to set maxConnection configuration to my application. This way I will be able to manipulate it. We were having issues during peak times due to low connection limits. Alternetively we can set `useGlobalResources` to `false` and set `connectionProviderSupplier`

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/reactor/netty/ReactorNettyConfigurations.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/reactor/netty/ReactorNettyConfigurations.java
@@ -16,6 +16,8 @@
 
 package org.springframework.boot.autoconfigure.reactor.netty;
 
+import reactor.netty.resources.ConnectionProvider;
+
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.context.annotation.Bean;
@@ -27,6 +29,7 @@ import org.springframework.http.client.ReactorResourceFactory;
  * auto-configuration class.
  *
  * @author Moritz Halbritter
+ * @author Huseyin Aydin
  * @since 2.7.9
  */
 public final class ReactorNettyConfigurations {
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks for the PR but I'm not sure that this is the right way to offer this configuration option. As proposed, this change will affect the global `HttpResources` which is contrary to the assertion in `ReactorResourceFactory` that prevents explicit configuration of resources when using the global resources. I think defining your own `ReactorResourceFactory` and disabling the use of global resources is the best option at the moment.

@violetagg @bclozel is there a way for us to limit the maximum number of connections without "polluting" the global resources?

violetagg:
> Thanks for the PR but I'm not sure that this is the right way to offer this configuration option. As proposed, this change will affect the global `HttpResources` which is contrary to the assertion in `ReactorResourceFactory` that prevents explicit configuration of resources when using the global resources. I think defining your own `ReactorResourceFactory` and disabling the use of global resources is the best option at the moment.
> 
> @violetagg @bclozel is there a way for us to limit the maximum number of connections without "polluting" the global resources?

IMO if one wants a configuration different than the global one, it is better to provide its own `ConnectionProvider` instead of changing the global one.

haydin505:
> Quote reply
> Reference in n

I've set `connectionProviderSupplier` but to use custom supplier we must set `useGlobalResources` to false. Also I'm not sure about `globalResourcesConsumer`, should we make a change [here](https://github.com/spring-projects/spring-framework/blob/main/spring-web/src/main/java/org/springframework/http/client/ReactorResourceFactory.java#L270) @violetagg?

wilkinsona:
Thanks, @violetagg.

@haydin505 Based on Violeta's feedback, we've decided that we don't want to add a property for this. Please disable global resources and use your own `ConnectionProvider` as you described above.

wilkinsona:
`globalResourcesConsumer` only applies when `useGlobalResources` is set to `true`. It's not applicable in a situation where global resources are disable and you're using a custom `ConnectionProvider`.

If you have further questions, particularly about Spring Framework which is a separate project, please follow up on Stack Overflow. As mentioned in [the guidelines for contributing](https://github.com/spring-projects/spring-boot/blob/main/CONTRIBUTING.adoc#using-github-issues), we prefer to use GitHub issues only for bugs and enhancements.

haydin505:
Actually, I've already applied corresponding enchancement to our project. I thought it would be cool to add this functionality out of the box with auto configuration. Thank you for your time and response @wilkinsona  🙏

PR #40918, opened 2024-05-28
Warn in the documentation that spring.profiles.group can only be used in non-profile-specific documents

vsanna:
## Why
Currently, [the official reference](https://docs.spring.io/spring-boot/reference/features/profiles.html
) says that `spring.profiles.active`, `spring.profiles.default` and `spring.profiles.include` can only be used in non profile-specific.
However, after some investigation, I believe `spring.profiles.group` also must be in non-profile-specific configuration files.

Lacking the warning made me confused a little bit. So I suppose it would be great if we have the similar section for `spring.profiles.group`.

## What
- I've added a warning section, that is similar to one used for `spring.profiles.include`, for `spring.profiles.group`
- I tweaked a little bit to align with the terminology "profile-specific" used in a headline.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/profiles.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/profiles.adoc
@@ -32,7 +32,7 @@ spring:
     default: "none"
 ----
 
-`spring.profiles.active` and `spring.profiles.default` can only be used in non-profile specific documents.
+`spring.profiles.active` and `spring.profiles.default` can only be used in non-profile-specific documents.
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/profiles.adoc
```diff
@@ -32,7 +32,7 @@ spring:
     default: "none"
 ----
 
-`spring.profiles.active` and `spring.profiles.default` can only be used in non-profile specific documents.
+`spring.profiles.active` and `spring.profiles.default` can only be used in non-profile-specific documents.
```
vsanna, line 35: this is a trivial point, so I don't have strong preferense or opinion. I will revert if I should

mhalbritter:

PR #40960, opened 2024-05-31
Update testing.adoc

AiJiangnan:
Update spying to spring.


philwebb:
Spying is the correct term here since we're talking about Mockito's mock and spy functionality.

PR #40963, opened 2024-05-31
Remove unnecessary null check

quaff:
`isUsingSources(attached, sources)` always return `false` if `attached` is `null`


mhalbritter:
Thanks!

Issue #36640, opened 2023-07-30
Hibernate binding parameters not showing in logs


bizwms:
I'm using Spring Boot version 3.1.1 and I've noticed that the Hibernate binding parameters are not showing in the logs. I've set the logging level for org.hibernate.type.descriptor.sql.BasicBinder to trace in my application.properties file, but it doesn't seem to have any effect.

Here is a snippet of my application.properties:

```
spring.jpa.show-sql=true
logging.level.org.hibernate.SQL=debug
logging.level.org.hibernate.type.descriptor.sql.BasicBinder=trace
logging.level.org.hibernate.type.EnumType=trace
spring.jpa.properties.hibernate.format_sql=true
```

I expect to see the binding parameters in the logs when executing SQL queries, but currently, I only see the SQL statements without the parameters.

Any help would be appreciated.

quaff:
Hibernate renamed `org.hibernate.type.descriptor.sql.BasicBinder` to `org.hibernate.orm.jdbc.bind`.

bizwms:
I was able to output the binding parameters. thank you.
application.properties:

`logging.level.org.hibernate.orm.jdbc.bind=trace`

bizwms:
I will close this issue.

PR #40964, opened 2024-05-31
Add `org.hibernate.orm.jdbc.bind` to `sql` logging group

quaff:
It's reasonable to treat prepared statement parameters as part of sql.

See https://github.com/spring-projects/spring-boot/issues/36640#issuecomment-1657394103


wilkinsona:
Thanks for the suggestion, but I don't think we should do this. If it were "reasonable to treat prepared statement parameters as part of sql", I think the Hibernate team would have included them in the `org.hibernate.SQL` logger that's already part of the `sql` logging group. I suspect that they're not included and have to be enabled separately as logging the actual parameters increases the risk of the logs including sensitive data. I think we should continue to align with that and require separate configuration to enable the logging of the actual parameter values.

quaff:
> I suspect that they're not included and have to be enabled separately as logging the actual parameters increases the risk of the logs including sensitive data.

Fair enough, I didn't notice this.

PR #40968, opened 2024-05-31
Include metadata for management.metrics.export.atlas.aggrConfig replacement

jjacobs44:
Upgrading some applications from boot 2.7 to 3.0+ I noticed this property has changed. I use OpenRewrite to usually catch such property renames and opened up this pull request with them to fix the missed rename and they redirected me here https://github.com/openrewrite/rewrite-spring/pull/517#issuecomment-2083266665

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -460,6 +460,14 @@
       }
     },
     {
+      "name": "management.metrics.export.atlas.aggrConfig",
+      "type": "java.lang.String",
+      "deprecation": {
+        "level": "error",
+        "replacement": "management.atlas.metrics.export.aggrConfig"
+      }
+    },
+    {
       "name": "management.metrics.export.atlas.batch-size",
       "type": "java.lang.Integer",
       "deprecation": {


philwebb:
I'm a little confused because looking at [`AtlasProperties`](https://github.com/spring-projects/spring-boot/blob/2.7.x/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/atlas/AtlasProperties.java) for 2.7.x we don't seem to offer a `aggrConfig` property.

@jjacobs44 Are you sure `aggrConfig` worked with 2.7? What did it actually configure on Atlas?

wilkinsona:
I'm confused too. In addition to Phil's point above, there doesn't appear to be an [`aggrConfig` property in current versions of Spring Boot](https://github.com/spring-projects/spring-boot/blob/fa131fa6a0dbab704597ea15e3d88d763ee3d027/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/atlas/AtlasProperties.java) or on [the Atlas side either](https://github.com/Netflix/spectator/blob/e87d0695302462b57e589dc3b590ee8b00c7808a/spectator-reg-atlas/src/main/java/com/netflix/spectator/atlas/AtlasConfig.java).

jjacobs44:
My apologies, this is a config specific to the company I work at that just happened to use the same prefix/scope. I'll close this out.

PR #40985, opened 2024-06-04
Polish code to use 'switch' instead of 'if', remove unessary unboxing and remote redundant cast

asashour:
This PR has the following changes:
- Use `switch` instead of `if`
- Remove unnecessary (un)boxing

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/InfinispanCacheConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/InfinispanCacheConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.boot.autoconfigure.cache;
 import java.io.IOException;
 import java.io.InputStream;
 import java.util.List;
+import java.util.Objects;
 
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/cache/InfinispanCacheConfiguration.java
```diff
@@ -81,10 +82,7 @@ private EmbeddedCacheManager createEmbeddedCacheManager(CacheProperties cachePro
 
 	private org.infinispan.configuration.cache.Configuration getDefaultCacheConfiguration(
 			ConfigurationBuilder defaultConfigurationBuilder) {
-		if (defaultConfigurationBuilder != null) {
-			return defaultConfigurationBuilder.build();
-		}
-		return new ConfigurationBuilder().build();
+		return Objects.requireNonNullElseGet(defaultConfigurationBuilder, ConfigurationBuilder::new).build();
```
mhalbritter, line 85: I think the old code is more readable.

asashour: Reverted, thanks.

PR #40986, opened 2024-06-04
Auto-configure SpringLiquibase with Liquibase Customizer

### LINKED ISSUE #38055: Auto-configure SpringLiquibase with any liquibase.integration.spring.Customizer bean
I would like to add lifecycle logging before, after, as well as on failed migrations.

To implement this, I planned to register a liquibase.changelog.visitor.ChangeExecListener on the Liquibase instance.

I found that the SpringLiquibase initialization bean fully encapsulates the migration.

So, to register an ChangeExecListener in the Liquibase instance, I only see the option to:

1. Inherit SpringLiquibase and override createLiquibase.
2. Copy the LiquibaseAutoConfiguration into our codebase and return the inherited SpringLiquibase in LiquibaseAutoConfiguration.LiquibaseConfiguration#createSpringLiquibase.

Versions used:
- Spring Boot: 3.1.4
- Liquibase: 4.22.0

Do you know of any other way to register custom logic, such as logging before, after, and on failed migrations?

asashour:
This PR considers a recently added Liquibase `Customizer`.

Fixes #38055

It includes an upgrade of `Liquibase` to `4.28.0`, which can be separately done.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
@@ -100,8 +100,7 @@ public class LiquibaseAutoConfiguration {
 		@Bean
 		public SpringLiquibase liquibase(ObjectProvider<DataSource> dataSource,
 				@LiquibaseDataSource ObjectProvider<DataSource> liquibaseDataSource, LiquibaseProperties properties,
-				ObjectProvider<Customizer<Liquibase>> customizer,
-				LiquibaseConnectionDetails connectionDetails) {
+				ObjectProvider<Customizer<Liquibase>> customizer, LiquibaseConnectionDetails connectionDetails) {
 			SpringLiquibase liquibase = createSpringLiquibase(liquibaseDataSource.getIfAvailable(),
 					dataSource.getIfUnique(), connectionDetails);
 			liquibase.setChangeLog(properties.getChangeLog());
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
```diff
@@ -125,6 +128,7 @@ public SpringLiquibase liquibase(ObjectProvider<DataSource> dataSource,
 			if (properties.getUiService() != null) {
 				liquibase.setUiService(UIServiceEnum.valueOf(properties.getUiService().name()));
 			}
+			liquibase.setCustomizer(customizer.getIfAvailable());
```
wilkinsona, line 131: This could be `customizer.ifAvailable(liquibase::setCustomizer)`. It would then only be set if the customizer's available which I think is slightly better than setting it to `null`.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfigurationTests.java
```diff
@@ -532,6 +535,14 @@ void shouldRegisterHints() {
 		assertThat(RuntimeHintsPredicates.resource().forResource("db/changelog/tables/init.sql")).accepts(hints);
 	}
 
+	@Test
+	void customizer() {
```
wilkinsona, line 539: We prefer longer names for test methods that describe what's being tested. Something like `whenCustomizerBeanIsDefinedThenItIsConfiguredOnSpringLiquibase`.

wilkinsona (changes requested):

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/liquibase/LiquibaseAutoConfiguration.java
asashour: Done, thanks.

asashour:
I am scratching my head: `./gradlew checkstyleMain checkstyleTest` locally passes, but not on the build server.

philwebb:
I'm not sure what's going on with CI, it may be unrelated to your code. Don't worry about it for now, we'll make sure things work before we merge.

mhalbritter:
I did the Liquibase update in #41194.

mhalbritter:
Thanks! I polished the changes a bit so that support for older Liquibase versions isn't broken.

PR #41051, opened 2024-06-11
Improve readability when listing three pillars of observability

mateusscheper:
Adding ":" to sentence to improve its readability.


mhalbritter:
Thanks!

PR #41053, opened 2024-06-11
Use Collection-based construction rather than separate construction and addAll()

PiyalAhmed:
Redundant Collection.addAll() call


philwebb:
Thanks very much @PiyalAhmed

PR #41090, opened 2024-06-12
Use chained append calls in StringBuilder for improved performance, and readability.

cmabdullah:
Use chained append calls in StringBuilder for improved memory usage, performance, and readability, and replace enhanced for loop with Java lambda expressions using .stream() API for better readability and functional programming.


philwebb:
Thanks for the suggestion but I prefer the code as it currently is.

cmabdullah:
Thank you for your response. I appreciate your feedback.

PR #41173, opened 2024-06-20
Update Kotlin DSL examples that configure the environment of bootBuildImage to be additive

sdeleuze:
Reading Gradle documentation and seeing other similar code samples, I don't think the `environment.get()` part is needed, and it looks not idiomatic.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/docs/antora/modules/gradle-plugin/examples/packaging/boot-build-image-env.gradle.kts
+++ b/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/docs/antora/modules/gradle-plugin/examples/packaging/boot-build-image-env.gradle.kts
@@ -7,7 +7,7 @@ plugins {
 
 // tag::env[]
 tasks.named<BootBuildImage>("bootBuildImage") {
-	environment.set(environment.get() + mapOf("BP_JVM_VERSION" to "17"))
+	environment.put("BP_JVM_VERSION", "17")
 }
 // end::env[]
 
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/docs/antora/modules/gradle-plugin/examples/packaging/boot-build-image-env.gradle.kts
```diff
@@ -7,7 +7,7 @@ plugins {
 
 // tag::env[]
 tasks.named<BootBuildImage>("bootBuildImage") {
-	environment.set(environment.get() + mapOf("BP_JVM_VERSION" to "17"))
+	environment.set(mapOf("BP_JVM_VERSION" to "17"))
```
wilkinsona, line 10: The intent is to add a `BP_JVM_VERSION` to any existing entries. Won't this replace any existing entries in the environment map such that it now only contains `BP_JVM_VERSION -> 17`?

wilkinsona: `environment` is a `MapProperty` and it has a `put(String, String)` method. I think that's what we should call. I'm not sure if there's some Kotlin DSL support for it, or if it should just be `environment.put("BP_JVM_VERSION", "17")`.

sdeleuze: Ah good point, I think I was confused by the fact the other samples in this page are also using `environment.set(mapOf("key" to "value"))`. I guess we should fix those as well.

I think `environment.put("BP_JVM_VERSION", "17")` is indeed what should be used. I have tested it successfully [here](https://github.com/sdeleuze/spring-boot-cds-demo/blob/fba078c6ca42db70b5f099bdc1f0f540589c4df9/build.gradle.kts#L29-L30). I don't think there is a better DSL than that.

Let me try to update other code samples accordingly.

sdeleuze: PR updated accordingly.

wilkinsona: Thanks.

mhalbritter:
Thanks Sébastien!

PR #41211, opened 2024-06-24
Simplify conditional expression

PiyalAhmed:
Simplify conditional expression


wilkinsona:
Thanks for the proposal but this doesn't look simpler to me. If anything, I find it slightly harder to read.

PR #41245, opened 2024-06-26
Update plexus-build-api as to not use plexus-utils 1.5.8

JaredHatfield:
While performing builds with the latest version of Spring Boot 3.3.1 I noticed that the build was pulling in an insecure dependency. Specifically plexus-utils 1.5.8 which is quite old:

https://mvnrepository.com/artifact/org.codehaus.plexus/plexus-utils/1.5.8

The issue here is that plexus-utils 1.5.8 has CVEs and build pipelines that block the use of vulnerable artifacts will result in this build failing.

I know that straight forward dependency upgrades aren't meant to be here but this looks like it may have fallen through the cracks.  

It took me a while to determine where this dependency was coming from and why it was so old. It turns out that org.sonatype.plexus:plexus-build-api:0.0.7 is where this is being pulled from as a transitive dependency:
https://github.com/spring-projects/spring-boot/blob/0bbaa77530c757c75574456ba519831aab8e6d38/spring-boot-project/spring-boot-parent/build.gradle#L162

Looking at this dependency it is on the latest version available, but the artifact was actually moved.

https://mvnrepository.com/artifact/org.sonatype.plexus/plexus-build-api

This artifact moved from org.sonatype.plexus to org.codehaus.plexus.  This old version appears to be widely used and hasn't been updated since 2011. In this new location the latest version is not as widely used but is from 2023.

https://mvnrepository.com/artifact/org.codehaus.plexus/plexus-build-api

The proposal here is to update this dependency to the latest version:

https://mvnrepository.com/artifact/org.codehaus.plexus/plexus-build-api/1.2.0

There is a workaround for this is in the build that uses the spring-boot-maven-plugin you can force the plexus-utils dependency to a newer version, but ideally this would be addressed by updating Spring Boot itself.

```xml
    <plugin>
        <dependencies>
            <dependency>
                <groupId>org.codehaus.plexus</groupId>
                <artifactId>plexus-utils</artifactId>
                <version>4.0.0</version>
            </dependency>
        </dependencies>
    </plugin>
```

### DIFF
--- a/spring-boot-project/spring-boot-parent/build.gradle
+++ b/spring-boot-project/spring-boot-parent/build.gradle
@@ -159,8 +159,8 @@ bom {
 			]
 		}
 	}
-	library("Plexus Build API", "0.0.7") {
-		group("org.sonatype.plexus") {
+	library("Plexus Build API", "1.2.0") {
+		group("org.codehaus.plexus") {
 			modules = [
 				"plexus-build-api"
 			]
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/build.gradle
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/build.gradle
@@ -14,7 +14,7 @@ configurations {
 
 dependencies {
 	compileOnly("org.apache.maven.plugin-tools:maven-plugin-annotations")
-	compileOnly("org.sonatype.plexus:plexus-build-api")
+	compileOnly("org.codehaus.plexus:plexus-build-api")
 	compileOnly("org.apache.maven:maven-core") {
 		exclude(group: "javax.annotation", module: "javax.annotation-api")
 		exclude(group: "javax.inject", module: "javax.inject")
@@ -72,7 +72,7 @@ dependencies {
 	mavenRepository(project(path: ":spring-boot-project:spring-boot-devtools", configuration: "mavenRepository"))
 	mavenRepository(project(path: ":spring-boot-project:spring-boot-docker-compose", configuration: "mavenRepository"))
 
-	runtimeOnly("org.sonatype.plexus:plexus-build-api")
+	runtimeOnly("org.codehaus.plexus:plexus-build-api")
 
 	versionProperties(project(path: ":spring-boot-project:spring-boot-dependencies", configuration: "effectiveBom"))
 }
--- a/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/build.gradle
+++ b/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/build.gradle
@@ -15,6 +15,7 @@ configurations {
 dependencies {
 	compileOnly("org.apache.maven.plugin-tools:maven-plugin-annotations")
 	compileOnly("org.codehaus.plexus:plexus-build-api")
+	compileOnly("org.codehaus.plexus:plexus-xml")
 	compileOnly("org.apache.maven:maven-core") {
 		exclude(group: "javax.annotation", module: "javax.annotation-api")
 		exclude(group: "javax.inject", module: "javax.inject")
@@ -73,6 +74,7 @@ dependencies {
 	mavenRepository(project(path: ":spring-boot-project:spring-boot-docker-compose", configuration: "mavenRepository"))
 
 	runtimeOnly("org.codehaus.plexus:plexus-build-api")
+	runtimeOnly("org.codehaus.plexus:plexus-xml")
 
 	versionProperties(project(path: ":spring-boot-project:spring-boot-dependencies", configuration: "effectiveBom"))
 }


wilkinsona (changes requested):
As proposed, this change will break the build. Can you please update `spring-boot-maven-plugin` so the it uses the new coordinates? We'll then be able to see if the Maven Plugin's tests continue to pass.

JaredHatfield:
Updated the references for the plugin as requested to fix the build.

JaredHatfield:
Looks like the build is failing. Given how large of a jump the version was in terms of years between releases this isn't surprising to me. When I get some time later today I'll look to see if I can resolve the build failures.

JaredHatfield:
Debugging plexus-utils more...

The version of plexus-build-api went from 0.0.7 to 1.2.0

https://github.com/codehaus-plexus/plexus-build-api/compare/plexus-build-api-0.0.7...plexus-build-api-1.2.0

The impacting change here is taking plexus-utils from 1.5.8 to version 4.0.0

https://github.com/codehaus-plexus/plexus-utils/compare/plexus-utils-1.5.8...plexus-utils-4.0.0

The failure from the build in the Spring Boot logs for this PR is as follows:

```
/home/runner/work/spring-boot/spring-boot/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/JavaCompilerPluginConfiguration.java:23: error: package org.codehaus.plexus.util.xml does not exist
import org.codehaus.plexus.util.xml.Xpp3Dom;
                                   ^
/home/runner/work/spring-boot/spring-boot/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/JavaCompilerPluginConfiguration.java:86: error: cannot find symbol
      private String getNodeValue(Xpp3Dom dom, String... childNames) {
                                  ^
  symbol:   class Xpp3Dom
  location: class JavaCompilerPluginConfiguration
/home/runner/work/spring-boot/spring-boot/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/JavaCompilerPluginConfiguration.java:72: error: cannot find symbol
                  if (pluginConfiguration instanceof Xpp3Dom dom) {
                                                     ^
  symbol:   class Xpp3Dom
  location: class JavaCompilerPluginConfiguration
/home/runner/work/spring-boot/spring-boot/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/JavaCompilerPluginConfiguration.java:87: error: cannot find symbol
            Xpp3Dom childNode = dom.getChild(childNames[0]);
            ^
  symbol:   class Xpp3Dom
  location: class JavaCompilerPluginConfiguration
4 errors
> Task :spring-b
```

Looking at the diff between those version it is confirmed that Xpp3Dom wsa removed.

Verifying this removal was introduced in the 4.0.0 version going back to the previous version of plexus-utils 3.5.1 the Xpp3Dom is still there. The relevant PR removing the class:

https://github.com/codehaus-plexus/plexus-utils/compare/plexus-utils-1.5.8...plexus-utils-3.5.1

Looking for alternatives here the plexus-build API version 1.0.0 does use plexus-util 3.5.0 but that is not ideal and better to use the most recent version.

https://repo1.maven.org/maven2/org/codehaus/plexus/plexus-build-api/1.0.0/plexus-build-api-1.0.0.pom

Looking into that change where the XML component was removed, the intent there was to split it out into a separate module.

https://github.com/codehaus-plexus/plexus-utils/commit/8bf874bb9563116bd6ecd4e697c59fd0662d0a2f

"Starting with version 4, XML classes (in `org.codehaus.plexus.util.xml` and `org.codehaus.plexus.util.xml.pull`) have been extracted to a separate `plexus-xml` 4: if you need them, just use this new artifact."

The good news here is that plexus-xml is included as a dependency on plexus-utils!  Unfortunately this is listed as an optional dependency.

https://repo1.maven.org/maven2/org/codehaus/plexus/plexus-utils/4.0.0/plexus-utils-4.0.0.pom

```xml
<dependency>
    <groupId>org.codehaus.plexus</groupId>
    <artifactId>plexus-xml</artifactId>
    <version>4.0.0</version>
    <optional>true</optional>
</dependency>
```

Therefore the solution here is to include this optional transitive dependency or remove the need for this code which is bigger of a change than I'm willing to make on my own.

I'm not entirely certain how to force this transitive optional dependency to be included in Gradle.

wilkinsona:
Thanks for digging in, @JaredHatfield.

Looking more closely, `plexus-utils` is a part of Maven's core. You can see this in the code above where `org.apache.maven.model.ConfigurationContainer.getConfiguration()` returns an object that is expected to be an `Xpp3Dom` instance. Looking at the `ConfigurationContainer` code, that's exactly what it is:

```
copy.configuration = new org.codehaus.plexus.util.xml.Xpp3Dom( (org.codehaus.plexus.util.xml.Xpp3Dom) this.configuration );
```

In light of this, I don't think we can upgrade plexus-utils to 4.x as it's a dependency of Maven itself. It uses 3.5.1 by default and we can see this in the dependencies of `spring-boot-maven-plugin` from Gradle's perspective where version conflicts are resolved by selecting the highest version. Unfortunately, Maven takes a more brittle approach and selects the version that's the nearest the root of the pom. This means that we end up with 1.5.8 from `plexus-build-api` rather than 3.5.1 that would come from elsewhere. It's this unwanted downgrading when Maven's building the dependency graph that we need to fix.

I've opened https://github.com/spring-projects/spring-boot/issues/41248 so that we can figure out what to do here. Thanks for bringing the problem to our attention.

PR #41251, opened 2024-06-26
@NestedConfigurationProperty doesn't work on records

acouvreur:
Add `ElementType.RECORD_COMPONENT` to `NestedConfigurationProperty` and implement `isMarkedAsNested` for `RecordParameterPropertyDescriptor`.

This will allow nested record to be properly harvested for their properties.

I've added a test case which fails to retrieve nested record groups property using the `@NestedConfigurationProperty` annotation.

This change adds the support for nested record to be properly extracted into groups when using `@NestedConfigurationProperty`.


acouvreur:
Fixed formatting

philwebb:
Thanks. I think it's an oversight that we didn't add support for this in 3.3 when we added record metadata support.

mhalbritter:
Thanks very much!

PR #41278, opened 2024-06-30
Publish an AuditEvent on logout

Chu3laMan:

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/security/AuthenticationAuditListener.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/security/AuthenticationAuditListener.java
@@ -24,6 +24,7 @@ import org.springframework.boot.actuate.audit.AuditEvent;
 import org.springframework.security.authentication.event.AbstractAuthenticationEvent;
 import org.springframework.security.authentication.event.AbstractAuthenticationFailureEvent;
 import org.springframework.security.authentication.event.AuthenticationSuccessEvent;
+import org.springframework.security.authentication.event.LogoutSuccessEvent;
 import org.springframework.security.web.authentication.switchuser.AuthenticationSwitchUserEvent;
 import org.springframework.util.ClassUtils;
 
@@ -51,6 +52,8 @@ public class AuthenticationAuditListener extends AbstractAuthenticationAuditList
 	 */
 	public static final String AUTHENTICATION_SWITCH = "AUTHENTICATION_SWITCH";
 
+	public static final String LOGOUT_SUCCESS = "LOGOUT_SUCCESS";
+
 	private static final String WEB_LISTENER_CHECK_CLASS = "org.springframework.security.web.authentication.switchuser.AuthenticationSwitchUserEvent";
 
 	private final WebAuditListener webListener = maybeCreateWebListener();
@@ -73,6 +76,9 @@ public class AuthenticationAuditListener extends AbstractAuthenticationAuditList
 		else if (event instanceof AuthenticationSuccessEvent successEvent) {
 			onAuthenticationSuccessEvent(successEvent);
 		}
+		else if(event instanceof LogoutSuccessEvent logoutSuccessEvent) {
+			onLogoutSuccessEvent(logoutSuccessEvent);
+		}
 	}
 
 	private void onAuthenticationFailureEvent(AbstractAuthenticationFailureEvent event) {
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/security/AuthenticationAuditListener.java
```diff
@@ -93,6 +99,18 @@ private void onAuthenticationSuccessEvent(AuthenticationSuccessEvent event) {
 		publish(new AuditEvent(event.getAuthentication().getName(), AUTHENTICATION_SUCCESS, data));
 	}
 
+	private void onLogoutSuccessEvent(LogoutSuccessEvent event) {
+		Map<String, Object> data = new LinkedHashMap<>();
+		if(event.getAuthentication() != null) {
```
wilkinsona, line 104: This check isn't needed as `getAuthentication()` should never return `null`.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/security/AuthenticationAuditListener.java
```diff
@@ -93,6 +99,18 @@ private void onAuthenticationSuccessEvent(AuthenticationSuccessEvent event) {
...
+			if(event.getAuthentication().getDetails() != null) {
+				data.put("details", event.getAuthentication().getDetails());
+			}
+			data.put("username", event.getAuthentication().getName());
```
wilkinsona, line 108: The username's already included in the event (it's its `principal`) so it doesn't needed to be included as a separate piece of data.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/security/AuthenticationAuditListener.java
```diff
@@ -51,6 +52,8 @@ public class AuthenticationAuditListener extends AbstractAuthenticationAuditList
 	 */
 	public static final String AUTHENTICATION_SWITCH = "AUTHENTICATION_SWITCH";
 
+	public static final String LOGOUT_SUCCESS = "LOGOUT_SUCCESS";
```
wilkinsona, line 55: Please add some javadoc here that describes the constant. It should also be marked with `@since 3.4.0`.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/security/AuthenticationAuditListener.java
```diff
@@ -73,6 +76,9 @@ else if (this.webListener != null && this.webListener.accepts(event)) {
 		else if (event instanceof AuthenticationSuccessEvent successEvent) {
 			onAuthenticationSuccessEvent(successEvent);
 		}
+		else if(event instanceof LogoutSuccessEvent logoutSuccessEvent) {
```
wilkinsona, line 79: Please check the formatting. You can run `./gradlew spring-boot-project:spring-boot-actuator:format` to correct it automatically.

wilkinsona (changes requested):
Thanks for the PR, @Chu3laMan. I've left a few comments for you to consider when you have a minute.

Chu3laMan:
> Thanks for the PR, @Chu3laMan. I've left a few comments for you to consider when you have a minute.

Thank you for your remarks, one question, don't we need to care about handling error during logout phase through the same method?

Thank you

Chu3laMan:
> Thanks for the PR, @Chu3laMan. I've left a few comments for you to consider when you have a minute.

Thank you for your remarks, do not we need to handle error during logout phase in the same function?

mhalbritter:
>  do not we need to handle error during logout phase in the same function?

I'm not sure I don't understand - what exactly do you mean?

mhalbritter:

Chu3laMan:
> > do not we need to handle error during logout phase in the same function?
> 
> I'm not sure I don't understand - what exactly do you mean?

I mean during the logout session, if an error occurred, don't we need to handle it via a custom event?

Chu3laMan:
> I merged the PR - thank you very much and congratulations on your first contribution 🎉!

I really appreciate it, Thank you @philwebb ,  @wilkinsona, and @mhalbritter

mhalbritter:
> > > do not we need to handle error during logout phase in the same function?
> > 
> > 
> > I'm not sure I don't understand - what exactly do you mean?
> 
> I mean during the logout session, if an error occurred, don't we need to handle it via a custom event?

This is all handled by Spring Security. This component in Boot only translates the Spring Security Events into audit events.

Issue #32806, opened 2022-10-19
Upgrade Java 8 version in CI image and .sdkmanrc


spring-builds:
https://github.com/bell-sw/Liberica/releases/download/8u352+8/bellsoft-jdk8u352+8-linux-amd64.tar.gz

PR #41326, opened 2024-07-05
Updated documentation at issue #32806

Aysha-py:


wilkinsona:
Given that this PR changes over 4000 files, I assume that it was opened in error.

PR #41333, opened 2024-07-05
Support Otlp Tracing's GRPC port from service connections

eddumelendez:
Otlp Tracing's exporter is configured using `Transport`. Current support for service connections read the mapped port for HTTP transport `4318`. This commits adds support to read port for GRPC transport `4317`.


wilkinsona:
Thanks for the PR, @eddumelendez. I'm not sure what the answer is, but this doesn't feel quite right to me. I don't think the connection details should have a separate method for the GRPC endpoint. Ideally, the existing `getUrl()` method would return the URL that's appropriate for the configured transport but I'm not sure how we could achieve that.

eddumelendez:
Hi @wilkinsona, yes, I agree. I think that's the only way I was able to achieve it. Looking forward for any hint to update the PR.

philwebb:
We discussed this today and we think we might be able to pass the `Transport` to a `OtlpTracingConnectionDetails.getUrl(...)` method. We can experiment with this before we merge the PR.

mhalbritter:
Thanks!

I've modified `OtlpTracingConnectionDetails` to add a `String getUrl(Transport transport)` method.

Issue #41336, opened 2024-07-07
Execution of Spring Batch with JobLauncherApplicationRunner and application exit codes


diqksrk:
`Spring Batch` is primarily executed by schedulers. It is typically run by schedulers like `Jenkins`, which determines whether a Jenkins job has failed or succeeded by reading the exit status of the application JVM.

However, the current code exits with a status of 0 even if the job fails, leading Jenkins to assume that the job completed successfully, which causes issues in operations.

Therefore, I believe there is a need to enhance the logic to transition the application's exit status according to the job's exit status.

The `CommandLineJobRunner` in Spring Batch illustrates these points well.
Please refer to the following link:
https://github.com/spring-projects/spring-batch/blob/main/spring-batch-core/src/main/java/org/springframework/batch/core/launch/support/CommandLineJobRunner.java

#### point 1
```
JobExecution jobExecution = launcher.run(job, jobParameters);
return exitCodeMapper.intValue(jobExecution.getExitStatus().getExitCode());
```
#### point 2
```
int result = command.start(jobPath, jobIdentifier, parameters, opts);
command.exit(result);
```

Lastly, please review the proposed changes.

wilkinsona:
@diqksrk Spring Boot has [general support for exit codes](https://docs.spring.io/spring-boot/reference/features/spring-application.html#features.spring-application.application-exit) that's intended to be used in situations like this. Please review the documentation and let us know if it meets your needs.

diqksrk:
I am aware of the general support for exit codes. I think I may have conveyed my intention incorrectly.

What I intended to say is that the application exit status in Spring Batch is determined based on the job's status. This is important when using it with a scheduler like Jenkins because the success of a Jenkins job depends on the application exit status.

Therefore, as I have proposed in the [PR](https://github.com/spring-projects/spring-boot/pull/41337/commits/6960f102a579579f5686bd3e8efa094c134f6fad), the application should throw an exception based on the job's status, so that the exit code is determined by the Spring context upon termination.

Here is the relevant code:
```
private void validateJobTerminateStatus() throws JobExecutionException {
    String exitCode = jobExplorer.getLastJobExecution(jobExplorer.getLastJobInstance(this.jobName))
		.getExitStatus().getExitCode();

     if (exitCodeMapper.intValue(exitCode) != JVM_EXITCODE_COMPLETED) {
   	  throw new JobExecutionException("Job Terminated with BatchStatus : " + exitCode);
     } 
}
```

[...]

PR #41337, opened 2024-07-07
add validate job status code

diqksrk:
see https://github.com/spring-projects/spring-boot/issues/41336

I have added logic to determine the application's state based on the job's status

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
@@ -39,6 +39,10 @@ import org.springframework.batch.core.converter.DefaultJobParametersConverter;
 import org.springframework.batch.core.converter.JobParametersConverter;
 import org.springframework.batch.core.explore.JobExplorer;
 import org.springframework.batch.core.launch.JobLauncher;
+import org.springframework.batch.core.launch.support.ExitCodeMapper;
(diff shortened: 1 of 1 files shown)


codespearhead:
Does this patch intent to solve https://github.com/spring-projects/spring-boot/issues/41336 ?

If so, can you [link this PR to that issue](https://docs.github.com/en/issues/tracking-your-work-with-issues/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), so that issue is closed automatically when this PR is merged?

To do that, you just need to add the exact word combination in the PR description:

fix https://github.com/spring-projects/spring-boot/issues/41336

scottfrederick:
Thanks for trying to help @codespearhead, but the steps you mention are not necessary. When we're ready, we will manually close the issue as superseded by the PR, and we will [ensure](https://github.com/spring-projects/spring-boot/wiki/Merging-Pull-Requests#polishing-prs) that the commit comments contain what's necessary to track back to the PR.

wilkinsona:
The changes proposed here shouldn't be necessary as `JobExecutionExitCodeGenerator` should produce an exit code based on the status of an jobs that have been executed:

https://github.com/spring-projects/spring-boot/blob/293739b9f0ff73eeea67657d1c8fa906ac5ba525/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobExecutionExitCodeGenerator.java#L41-L49

If you use `SpringApplication.exit`, as I linked to above, this exit code generator should be involved in deciding the JVM's exit code. I'm going to close this one as I don't think the changes are needed. If you're using `System.exit` and `SpringApplication.exit` and it isn't working as expected, a different change to the one proposed here will be needed to fix it. We can figure that out in #41336.

PR #41355, opened 2024-07-09
Spring Authorization Server now defaults multipleIssuersAllowed to false and it cannot be easily re-enabled

opcooc:


wilkinsona:
Thanks for the PR, @opcooc.

It looks like the default [changed late in the 1.3 development cycle](https://github.com/spring-projects/spring-authorization-server/issues/1611). Without this change, if a Boot users wants to flip the default back to its previous value, they would have to stop using the `spring.security.oauth2.authorizationserver.*` properties and define their own `AuthorizationServerSettings` bean. That feels rather clunky to me so I am tempted to treat this is a bug and to add a property to make it easier in Spring Boot 3.3.x.

@jgrandja, what's you take on this please? Does the above sound reasonable or is allowing multiple issuers sufficiently rare that would could introduce the property in Boot 3.4 and require an `AuthorizationServerSettings` bean in 3.3.x?

jgrandja:
@wilkinsona Yes, the default for `AuthorizationServerSettings.multipleIssuersAllowed` was reverted to `false` just before we released `1.3` because a default of `true` did not automatically make the authorization server multi-tenant capable as there are a a few custom configurations required by the application as indicated in the [How-to: Implement Multitenancy](https://docs.spring.io/spring-authorization-server/reference/guides/how-to-multitenancy.html) guide.

We should add a Boot property for `AuthorizationServerSettings.multipleIssuersAllowed` to be consistent with the existing properties available for the `AuthorizationServerSettings`. I think merging this PR in `3.3.x` is totally fine but I'll leave it up to you. And apologies, I should have raised a ticket for this so it made it into `3.3.0`.

wilkinsona:
Thanks, @jgrandja. In that case, we'll consider this to be a bug of omission and add the property in 3.3.x.

wilkinsona:
Thank you, @opcooc. I polished your changes as part of merging them. Please see https://github.com/spring-projects/spring-boot/commit/1a6760e21d1f95d3bf7169c1e062733cde66cff6 if you're interested.

opcooc:
> Polish "Add configuration property to allow multiple issuers"

Thank you for helping me modify

PR #41466, opened 2024-07-12
Update RepositoryMetricsAutoConfiguration.java

heowc:


wilkinsona:
Thanks for the suggestion but it's intentional that the `@Bean` method is `static` as it's defining a bean post-processor.

PR #41479, opened 2024-07-13
Fix cyclonedx caching

### LINKED ISSUE #41478: Provide a mechanism to opt out of including the CycloneDX SBOM in the uber jar or war
Since Spring Boot 3.3 the gradle plugin copies the bom generated by cyclonedx in the resources, s.t. the jars contain these files. CycloneDx generates a creation timestamp into the generated SBOM file and thus this file changes on each execution. All tasks having a input on the jar (such as `test`) but also the the jar generation task itself thus have different gradle cache keys each execution and therefore effectively the cache is disabled for all projects that have the `cyclonedx` plugin applied and configured. 

I am aware that one could argue that this should be fixed by `cyclonedx`. On the other hand I believe this timestamp is necessary for other tools (such as the dependency track) and I don't think they will disable it. 

One quick fix for all consumers is to add the following into their `build.gradle` (use the correct filename here): 

```groovy
    normalization {
        runtimeClasspath {
            ignore "META-INF/sbom/bom.xml"
        }
    }
```

aykborstelmann:
Ignore sboms in classpath during caching because they contain timestamps.

Partially fixes_ #41478 by adding input normalization for runtime classpaths. 
This does not yet fix caching for the jar and bootJar task.


wilkinsona:
Thanks for the suggestion, but as described in https://github.com/spring-projects/spring-boot/issues/41478#issuecomment-2226921312, I don't think we can do this.

PR #41551, opened 2024-07-17
Add support for grafana/otel-lgtm and LgtmStackContainer from Service Connection

eddumelendez:
Support `grafana/otel-lgtm` from Docker Compose and `LgtmStackContainer`
from Testcontainers' Grafana module as a Service Connection for Metrics
and Tracing with Opentelemetry.

The image provides Loki, Grafana, Tempo, Mimir, Prometheus and Opentelemetry Collector installed, which is nice for local development.

This PR use Testcontainers for Java 1.20.0, which contains the brand new `grafana` module. The new version works well with previous spring boot versions as you can see [here](https://github.com/eddumelendez/testcontainers-samples), so, would be great to port it to 3.2.x and 3.3.x branches as well.

Thanks!

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -2114,7 +2114,7 @@ bom {
 			releaseNotes("https://github.com/xerial/sqlite-jdbc/releases/tag/{version}")
 		}
 	}
-	library("Testcontainers", "1.19.8") {
+	library("Testcontainers", "1.20.0") {
 		group("org.testcontainers") {
 			imports = [
 				"testcontainers-bom"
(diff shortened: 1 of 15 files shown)


snicoll:
> This PR use Testcontainers for Java 1.20.0, which contains the brand new grafana module. The new version works well with previous spring boot versions as you can see [here](https://github.com/eddumelendez/testcontainers-samples), so, would be great to port it to 3.2.x and 3.3.x branches as well.

Thanks for the suggestion but our upgrade policy states we won't upgrade to a new feature release in a maintenance release of Spring Boot. I've upgraded to 1.20.0 on `main` in a separate issue.

snicoll:
Thanks again @eddumelendez

PR #41553, opened 2024-07-18
Set spring.elasticsearch.socket-keep-alive default to true

YMNNs:
Elasticsearch's first connection after a period of inactivity (about 30 min) causes a connection timeout. Set spring.elasticsearch.socket-keep-alive default to true can aviod this problem


wilkinsona:
Thanks for the suggestion. The Elasticsearch client does not have keep-alive enabled by default and we want to align with that. If you believe it should be enabled by default, please raise it with the client's maintainers.

Issue #40342, opened 2024-04-13
RFC: new property "spring.devtools.livereload.additional-paths" 


metaforte:
Thymeleaf allows using files from src folder during development.
 However live reload does not have a file system watcher for "additional paths" to auto reload based on those templates

Just as there is there is `spring.devtools.restart.additional-paths`, 

request new property `spring.devtools.livereload.additional-paths`

this should work in support of `thymeleaf.prefix`

````yaml
logging:
  config: classpath:log4j2-local.xml
spring:
  devtools:
    livereload:
      enabled: 'true'
      additional-paths:
        - file:src/main/resources/static/
        - file:src/main/resources/templates/
  thymeleaf:
    cache: 'false'
    prefix: file:src/main/resources/templates/
    check-template-location: 'true'

  web:
    resources:
      static-locations:
        - file:src/main/resources/static/
        - classpath:/static/
      cache:
        period: 0
````

mhalbritter:
I think this is already possible, but it's somewhat non-intuitive:

Add the additional paths to `spring.devtools.restart.additional-paths` but then exclude them from restarting the application through `spring.devtools.restart.additional-exclude`.

This should set the filewatcher to watch for changes, triggering the livereload, but don't restart the app as the location has been excluded from restart.

But I agree that there should be a property for that.

itsAkshayDubey:
Hi @mhalbritter ,

I would like to work on this enhancement, could you please assign me this issue so that I can submit my PR in favour of this issue.

Thanks,
Akshay

philwebb:
All yours @itsAkshayDubey

metaforte:
I used a work around and achieved this behavior 

see LiveReloadConfig.java in this repo https://github.com/ibhatech/world-facts/tree/dev

itsAkshayDubey:
Unable to access this repo. @metaforte

metaforte:
made it public now

itsAkshayDubey:
Hi @philwebb ,

Kindly have a look at PR #41566

Thanks,
Akshay

bclozel:
We decided to deprecate live reload entirely, see #48971

PR #41566, opened 2024-07-20
Add a property for configuring additional livereload paths

itsAkshayDubey:
This PR enhances the functionality related to issue #40342 by introducing configurable additional paths for live reload. This feature mirrors the existing capability for configuring additional paths for application restart.

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/DevToolsProperties.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/DevToolsProperties.java
@@ -30,6 +30,7 @@ import org.springframework.util.StringUtils;
  *
  * @author Phillip Webb
  * @author Stephane Nicoll
+ * @author Akshay Dubey
  * @since 1.3.0
  */
 @ConfigurationProperties(prefix = "spring.devtools")
@@ -198,6 +199,22 @@ public class DevToolsProperties {
 		 */
 		private int port = 35729;
 
+		/**
+		 * Additional paths to watch for changes.
+		 */
+		private List<File> additionalPaths = new ArrayList<>();
+
+		/**
+		 * Amount of time to wait between polling for classpath changes.
+		 */
+		private Duration pollInterval = Duration.ofSeconds(1);
+
+		/**
+		 * Amount of quiet time required without any classpath changes before a reload is
+		 * triggered.
+		 */
+		private Duration quietPeriod = Duration.ofMillis(400);
+
 		public boolean isEnabled() {
 			return this.enabled;
 		}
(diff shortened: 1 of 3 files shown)


mhalbritter:
Hey, thanks for the PR. I've looked at the changes and have some feedback:

You're creating a new `FileSystemWatcher` bean to watch for live reload changes. However, the restart functionality already has a `FileSystemWatcher` to watch for changes. It would be good to reuse that `FileSystemWatcher` so that it can trigger a restart (if restart paths change) or trigger a live reload (if reload paths change).

That would mean that `pollInterval` and `quietPeriod` can't be configured separately, but I don't think this is a big deal.

Do you have the time to implement this idea?

itsAkshayDubey:
Hi @mhalbritter , sure I'll submit my PR super soon.

philwebb:
Hi @itsAkshayDubey, any luck implementing the requested changes?

itsAkshayDubey:
On it, will submit my PR in next 2 days.

itsAkshayDubey:
Hi @mhalbritter @philwebb , kindly have a look.

mhalbritter:
Hello @itsAkshayDubey. Thanks for taking another go at it, but unfortunately, the changes aren't what we are looking for. Now we have two new `FileSystemWatcher`. In [that comment above](https://github.com/spring-projects/spring-boot/pull/41566#issuecomment-2257710960) I suggested that we instead use the already existing `FileSystemWatcher` bean which is already there to watch for restart changes and use it to also watch for livereload changes. That way we would only have one `FileSystemWatcher`.

Do you want to give it another go or should we abort here?

itsAkshayDubey:
Hi @mhalbritter,
Thank you for your input. I'll definitely address this. However, I would appreciate a bit of clarification to proceed effectively.
As I understand it, `RestartConfiguration` is a conditional bean based on the `spring.devtools.restart.enabled` property. To access the `FileSystemWatcher` from `RestartConfiguration`, we would need the bean to be available. In the current implementation, I am utilising `FileSystemWatcher` when the `RestartConfiguration` bean is present, or otherwise, creating a new instance.
Could you kindly advise on the best approach to consistently use the same `FileSystemWatcher` from the `RestartConfiguration` bean, regardless of its conditional nature?

mhalbritter:
The `FileSystemWatcher` is created in `org.springframework.boot.devtools.autoconfigure.LocalDevToolsAutoConfiguration.RestartConfiguration#newFileSystemWatcher`. Right now, it uses `org.springframework.boot.devtools.autoconfigure.DevToolsProperties.Restart#additionalPaths` to configure additional paths.

My idea was that we add the `org.springframework.boot.devtools.autoconfigure.DevToolsProperties.Livereload#additionalPaths` to the same `FileSystemWatcher`, and then check in in the listeners if the change should lead to a restart or a livereload. Then we only need one `FileSystemWatcher`.

I haven't looked into detail how we need to move the auto-configuration classes and conditions around, but I guess at least for the `FileSystemWatcher` we need a condition which is true if either `spring.devtools.restart.enabled=true` OR `spring.devtools.livereload.enabled=true`. This can be done with a `org.springframework.boot.autoconfigure.condition.AnyNestedCondition`.

30atm:
Hi @itsAkshayDubey have you had a chance to look into the requested code change?

mhalbritter:
Closing because of inactivity.

PR #41587, opened 2024-07-23
Update "Amazon Redshift" DataSource for DatabaseDriver.java

lamtrinhdev:
As the latest guide document at https://docs.aws.amazon.com/redshift/latest/mgmt/jdbc20-register-driver-class.html they are provide "Amazon Redshift" DataSource as "com.amazon.redshift.jdbc.DataSource" . I would like to update it into DatabaseDriver.java .


snicoll:
@lamtrinhdev thanks for the suggestion but `com.amazon.redshift.jdbc.DataSource` doesn't seem to be an XA `DataSource` and that's what that second argument is meant to convey. Have I missed something?

lamtrinhdev:
Dear @snicoll ,

You are right. Sorry for my mistake.

Thanks,
Lam

Issue #40682, opened 2024-05-09
Create an alternative to InvalidConfigurationPropertyValueException for cases where there isn't a single solution


philwebb:
See https://github.com/spring-projects/spring-boot/issues/33261#issuecomment-2075055929

We'd like a new exception type and failure analyzer for the example listed in the above comment.

stevbark:
Hello @philwebb  my name is Stephen, I am new to this project. 

Does this story still need to be done? If so I am interested in picking this story up.  

Is the best place to ask questions related to working on this project in stackoverflow.com tagged with spring-boot or the Gitter community mentioned in the readme? 

Thanks

philwebb:
Hi @stevbark,

This issue is still open and needs to be handled, however, we haven't yet branched the code so we won't be able to merge anything until that happens.

Questions relating to a specific issue can be asked on the the issue itself, for general help contributing Gitter is probably best. Questions on stackoverflow.com should be for general problems, I think questions specifically about contributing are probably going to be down-voted on stackoverflow.com.

We also have [this wiki page](https://github.com/spring-projects/spring-boot/wiki/Working-with-the-Code) which should help you get started with the basics of working with the code.

Thanks for getting involved!

stevbark:
Thanks for the warm welcome @philwebb,

When you say we havent branched the code, does this mean I create a new branch from the main branch that I work on? Or is this branched code something else entirely that I need to wait for?

Thanks

wilkinsona:
@stevbark There may be a delay in merging the changes until we've created the 3.3.x branch and main is building 3.4.0 snapshots but there's no need for you to wait. Please submit your changes based off the main branch and we'll take things from there.

AnneMayor:
Hi, I am going to open PR within this weekend if nobody handles this issue.
Thanks🙂

stevbark:
Hello, I got a notification last week saying that Lee Anne was "going to open PR within [last] weekend if nobody handles this issue." However, I dont see that message here. Should I continue to work on this story or has it been completed? Thanks

stevbark:
Sorry I had family matters to attend to.

wilkinsona:
Please carry on, @stevbark, we haven't received a contribution from anyone else. I'll assign the issue to you so that the ownership is clearer to others.

stevbark:
Thanks! 

[...]

PR #41597, opened 2024-07-23
Create an alternative to InvalidConfigurationPropertyValueException for cases where there isn't a single solution #40682

stevbark:
Created InvalidConfigurationPropertyValueInCombinationException class. Updated several classes to call the exception when necessary and updated some test cases.


stevbark:
Are you sure? I would really like to try this one again if you can give me a little more guidance.

wilkinsona:
Sorry, @stevbark. We don't have time to provide that level of guidance for #40682. Please look for issues labeled as `status: ideal-for-contribution` for that.

stevbark:
ok no problem, I figured I should check. Ill look into issues labeled status: ideal-for-contribution. Thanks!

PR #41606, opened 2024-07-24
Use reflection for deprecated gradle file permission API

### LINKED ISSUE #41599: Using Gradle's new file permission API is implemented in a way that prevents removal of the old API
Addressing issue
* https://github.com/spring-projects/spring-boot/issues/37878

resulted in changes in this code:
* https://github.com/spring-projects/spring-boot/blob/793aca60d2d6ad4ff1618bc19672038cd4755120/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootArchiveSupport.java#L153

The problem with the change is that it should call the fallback code (the old API methods, like `copySpec::getDirMode` and `copySpec::getFileMode`) also via reflection, because as it is now, we are unabled to remove them (they've been deprecated for a while now in Gradle).

We intend to do the removal in Gradle 9, which is in the near future and this would completely prevent us from doing so.

cms04:
This should close #41599


wilkinsona:
Thanks for the PR but, as described in #41599, we prefer to swap from method references to lambdas in situations like this.

Issue #41603, opened 2024-07-24
Add rule to prevent calls to Objects.requireNonNull()


wilkinsona:
From @mhalbritter in https://github.com/spring-io/spring-javaformat/issues/339:

> We should add a rule preventing calls to `Objects.requireNonNull()`, instead we should use `Assert.notNull()` from Spring Framework. `Objects.requireNonNull()` throws a `NullPointerException`, while `Assert.notNull()` throws a `IllegalArgumentException`.

Checkstyle can't do this accurately. We can do it in our own build instead using ArchUnit by adding a rule to the `ArchitectureCheck` task.

ivamly:
Hello!

I’m relatively new to open source contributions, but I’m confident that I can handle this task. Could I get started on this?

Thank you!

philwebb:
Thanks @ivamly, you might want to look at [this code](https://github.com/mapstruct/mapstruct/blob/main/processor/src/main/java/org/mapstruct/ap/internal/util/Collections.java) which is pretty similar to what we want to do.

ivamly:
Hello!

Thanks for giving me the chance to tackle this task. I’m a bit confused about the code snippet you shared. I was initially thinking we’d be adding a rule directly in ArchitectureCheck.java.

@philwebb, could you clarify how the provided code fits into the task? Any guidance would be super helpful!

Thanks again for the support!

wilkinsona:
@ivamly I suspect that was a stale clipboard entry and Phil meant to link to [this code](https://github.com/spring-projects/spring-boot/blob/e8391f121e2d3512c1b8461655d28d25ce13c605/buildSrc/src/main/java/org/springframework/boot/build/architecture/ArchitectureCheck.java#L198-L203). It'd also be good to add a test to [`ArchitectureCheckTests`](https://github.com/spring-projects/spring-boot/blob/main/buildSrc/src/test/java/org/springframework/boot/build/architecture/ArchitectureCheckTests.java) too please.

philwebb:
Yeah, sorry. Too many tabs open and I copied the wrong thing! I meant to link to https://github.com/spring-projects/spring-boot/blob/f1e98d0a73f842d7e163ed248315634d42893ce5/buildSrc/src/main/java/org/springframework/boot/build/architecture/ArchitectureCheck.java#L198-L203

ivamly:
Thanks for the clarification! I’ll check out the code and add the rule and tests.

wilkinsona:
[...]

PR #41611, opened 2024-07-25
Add rule to prevent calls to Objects.requireNonNull()

ivamly:
# Add Check for Objects.requireNonNull Usage and Corresponding Tests

This pull request for issue ticket #41603 introduces a new architecture check for ensuring that Objects.requireNonNull is not used directly. Instead, the Assert.notNull method from Spring Framework should be used. Additionally, this update includes new test cases and necessary files to validate this architecture rule.

### New Architecture Check:

Added a method to the ArchitectureCheck class to verify that Objects.requireNonNull is not used within the codebase.

### Test Cases:

Implemented tests to validate the new architecture check:
RequireNonNullUsage: Tests that the direct use of Objects.requireNonNull in the code results in a failure.
AssertNotNullUsage: Tests that using Assert.notNull is allowed and does not cause a failure.

### File Updates:

Created the necessary files and classes for testing:
RequireNonNullUsage: Contains a method using Objects.requireNonNull to trigger the architecture check.
AssertNotNullUsage: Contains a method using Assert.notNull to verify that it does not violate the rule.

These changes help enforce best practices for null-checking within the project and ensure that the code adheres to the preferred use of Spring utilities.


ivamly:
I've reviewed the contribution guidelines and considered adding Javadoc comments to ArchitectureCheck.java and ArchitectureCheckTests.java. However, I wasn't sure if it was necessary to include Javadoc for these classes, so I have refrained from adding them for now. If it's required, please let me know, and I will update the code accordingly.

ivamly:
I noticed from the build report that the build is failing due to missing newline characters at the end of my files. What should I do to resolve this issue? Do I need to make another commit, update the pull request, or take any other action?

Apologies if this seems like a basic question—I'm still new to this process. Thank you for your help!

wilkinsona:
Either we can take care of it as part of merging the changes, or you can push another commit to your `main` branch and it will the automatically update this PR.

ivamly:
I'll make a commit tonight.

wilkinsona:
It looks like there's a bug in ArchUnit that's causing it to report many false positives. It trips up on code that uses a method reference such as this:

```java
public void someCodeThatUsesAMethodReference() {
	List.of("1", "2", "3").forEach(System.out::println);
}
```

We'll have to see if there's a workaround or if there's a fix available.

ivamly:
Thank you for the update. That's interesting. Is there anything I can do? I’m happy to assist if needed.

wilkinsona:
It isn't a bug in ArchUnit, but an unfortunate side effort of the bytecode that's produced by a method reference. The example above looks like this:

```
  public void someCodeThatUsesAMethodReference();
    descriptor: ()V
    flags: (0x0001) ACC_PUBLIC
    Code:
      stack=3, locals=1, args_size=1
         0: ldc           #15                 // String 1
         2: ldc           #17                 // String 2
         4: ldc           #19                 // String 3
         6: invokestatic  #21                 // InterfaceMethod java/util/List.of:(Ljava/lang/Object;Ljava/lang/Object;Ljava/lang/Object;)Ljava/util/List;
         9: getstatic     #27                 // Field java/lang/System.out:Ljava/io/PrintStream;
        12: dup
        13: invokestatic  #33                 // Method java/util/Objects.requireNonNull:(Ljava/lang/Object;)Ljava/lang/Object;
        16: pop
        17: invokedynamic #39,  0             // InvokeDynamic #0:accept:(Ljava/io/PrintStream;)Ljava/util/function/Consumer;
        22: invokeinterface #43,  2           // InterfaceMethod java/util/List.forEach:(Ljava/util/function/Consumer;)V
        27: return
      LineNumberTable:
        line 30: 0
        line 31: 27
      LocalVariableTable:
        Start  Length  Slot  Name   Signature
            0      28     0  this   Lorg/springframework/boot/build/architecture/objects/noRequireNonNull/AssertNotNullUsage;
```

Note 13 where it's calling `java.lang.Objects.requireNonNull(Object)`.

>  Is there anything I can do? I’m happy to assist if needed.

Thanks for the offer.

Could you please update the PR so that it doesn't prohibit `Objects.requireNonNull(Object)`? We can still prohibit `Objects.requireNonNull(Object, String)` with `Assert.notNull(Object, String)` being the recommend replacement. We should also prohibit `Objects.requireNonNull(Object, Supplier)` with `Assert.notNull(Object, Supplier)` being the recommend replacement. These two will probably have to be separate rules as the `because` reason will be different.

wilkinsona:
Thanks very much, @ivamly.

Issue #7333, opened 2016-11-07
Lock contention in LaunchedURLClassLoader


dreis2211:
Hey,

we recently noticed lock contention coming from the LaunchedURLClassLoader in two of our applications, which boils down to the following in Classloader:

```java
   // This method is invoked by the virtual machine to load a class.
    private Class<?> loadClassInternal(String name)
        throws ClassNotFoundException
    {
        // For backward compatibility, explicitly lock on 'this' when
        // the current class loader is not parallel capable.
        if (parallelLockMap == null) {
            synchronized (this) {
                 return loadClass(name);
            }
        } else {
            return loadClass(name);
        }
    }
```

In case there is no parallelLockMap, it locks on the current instance of the classloader if it's not parallel capable. Which the LaunchedURLClassLoader unfortunately isn't, while its superclass URLClassLoader is. I wonder if there is a reason for that apart from the fact that this was introduced in 1.7 and might not be adopted yet. Setting the fast exception mechanism seems to go via a ThreadLocal variable already. Am I missing something that prohibits the use of the following?

```java
	static {
		ClassLoader.registerAsParallelCapable();
	}
```

I appreciate any feedback or clarification.

Cheers,
Christoph

wilkinsona:
Thanks for spotting this. LaunchedURLClassLoader was parallel capable, but it was lost in https://github.com/spring-projects/spring-boot/commit/87fe0b2adeef85c842c009bfeebac1c84af8a5d7#diff-dc50de280501378d899d1ca38e9877fc. We need to reinstate that logic.

wilkinsona:
Thanks to the simplified delegation model, we don't need to mess around with the lock provider stuff. Simply registering as parallel capable should be sufficient. Closing in favour of #7334.

PR #41665, opened 2024-08-01
Launcher's ClassLoader is no longer parallel capable

Rajin9601:
While debugging deadlock in our application, I found that spring-boot's LaunchedClassLoader is not registered as parallel capable class loader. 

Even though LaunchedClassLoader register itself as parallel capable, ClassLoader only accepts registerAsParallelCapable if the super class is also parallel capable.

https://github.com/openjdk/jdk/blob/f174bbd3baf351ae9248b70454b3bc5a89acd7c6/src/java.base/share/classes/java/lang/ClassLoader.java#L270-L284
```
        static boolean register(Class<? extends ClassLoader> c) {
            synchronized (loaderTypes) {
                if (loaderTypes.contains(c.getSuperclass())) {
                    // register the class loader as parallel capable
                    // if and only if all of its super classes are.
                    // Note: given current classloading sequence, if
                    // the immediate super class is parallel capable,
                    // all the super classes higher up must be too.
                    loaderTypes.add(c);
                    return true;
                } else {
                    return false;
                }
            }
```

Looking at this issue (https://github.com/spring-projects/spring-boot/issues/7333), it seems LaunchedClassLoader is parallel capable. So this PR registers LaunchedClassLoader's parent class, JarUrlClassLoader as parallel capable.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/net/protocol/jar/JarUrlClassLoader.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-loader/src/main/java/org/springframework/boot/loader/net/protocol/jar/JarUrlClassLoader.java
@@ -40,6 +40,10 @@ import org.springframework.boot.loader.launch.LaunchedClassLoader;
  */
 public abstract class JarUrlClassLoader extends URLClassLoader {
 
+	static {
+		ClassLoader.registerAsParallelCapable();
+	}
+
 	private final URL[] urls;
 
 	private final boolean hasJarUrls;


wilkinsona:
Thanks very much, @Rajin9601.

This looks like a regression in 3.2 to me where `JarUrlClassLoader` was introduced as the super-class of `LaunchedURLClassLoader`. In 3.2 with the classic loader or in 3.1, `LaunchedURLClasssLoader` extends `URLClassLoader` and will be parallel capable.

wilkinsona:
Thanks very much, @Rajin9601 .

PR #41718, opened 2024-08-08
Use activemq-bom

eddumelendez:
ActiveMQ 6.1.3 provides activemq-bom.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -14,37 +14,10 @@ bom {
 			issueLabels = ["type: dependency-upgrade"]
 		}
 	}
-	library("ActiveMQ", "6.1.2") {
+	library("ActiveMQ", "6.1.3") {
 		group("org.apache.activemq") {
-			modules = [
-					"activemq-amqp",
-					"activemq-blueprint",
-					"activemq-broker",
-					"activemq-client",
-					"activemq-console" {
-						exclude group: "commons-logging", module: "commons-logging"
-					},
-					"activemq-http",
-					"activemq-jaas",
-					"activemq-jdbc-store",
-					"activemq-jms-pool",
-					"activemq-kahadb-store",
-					"activemq-karaf",
-					"activemq-log4j-appender",
-					"activemq-mqtt",
-					"activemq-openwire-generator",
-					"activemq-openwire-legacy",
-					"activemq-osgi",
-					"activemq-pool",
-					"activemq-ra",
-					"activemq-run",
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -14,37 +14,10 @@ bom {
...
-	library("ActiveMQ", "6.1.2") {
+	library("ActiveMQ", "6.1.3") {
 		group("org.apache.activemq") {
-			modules = [
-					"activemq-amqp",
-					"activemq-blueprint",
-					"activemq-broker",
-					"activemq-client",
-					"activemq-console" {
-						exclude group: "commons-logging", module: "commons-logging"
```
snicoll, line 25: Did you check this is no longer necessary?

spring-boot-project/spring-boot-dependencies/build.gradle
```diff
@@ -14,37 +14,10 @@ bom {
...
-					"activemq-openwire-generator",
-					"activemq-openwire-legacy",
-					"activemq-osgi",
-					"activemq-pool",
-					"activemq-ra",
-					"activemq-run",
-					"activemq-runtime-config",
-					"activemq-shiro",
-					"activemq-spring" {
-						exclude group: "commons-logging", module: "commons-logging"
```
snicoll, line 44: Ditto

snicoll (commented):
Thanks for the PR @eddumelendez but there are two excludes there that are lost with the use of the BOM. Did you check that it isn't relevant anymore?

spring-boot-project/spring-boot-dependencies/build.gradle
eddumelendez: Hi @snicoll, those are still necessary. How should I proceed?

wilkinsona: We can use the bom and then also declare the modules that require exclusions. Here's an example for Micrometer:

```
	library("Micrometer", "1.9.17") {
		considerSnapshots()
		group("io.micrometer") {
			modules = [
				"micrometer-registry-stackdriver" {
					exclude group: "javax.annotation", module: "javax.annotation-api"
				}
			]
			imports = [
				"micrometer-bom"
			]
		}
	}
```

This uses `micrometer-bom` while also adding an exclusion to the `micrometer-registry-stackdriver` dependency that it manages.

eddumelendez: Thanks!

eddumelendez:
PR has been updated

wilkinsona:

PR #41744, opened 2024-08-08
Fix NullPointerException Risk in Dependency Extraction in BomPlugin

hyunmin0317:
## Summary

This pull request addresses a potential NullPointerException issue in the BomPlugin class. The original code invoked the text() method on the result of findChild() without null-checking, which could lead to exceptions if findChild() returns null.

## Changes Made

Introduced a new method extractText() to safely handle null values when extracting text.
Updated the usage of the text() method to use extractText() to prevent NullPointerException.

## Details

The changes involve modifying the BomPlugin class as follows:

```java
String groupId = extractText(findChild(dependency, "groupId"));
String artifactId = extractText(findChild(dependency, "artifactId"));
```

The extractText() method ensures that if findChild() returns null, it will not cause a NullPointerException, thus improving the robustness of the code.

## Reason for Change

This modification enhances the stability of the code by handling potential null values gracefully, ensuring that the application can handle missing child elements without crashing.

Thank you for reviewing this pull request!

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/bom/BomPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/bom/BomPlugin.java
@@ -140,10 +140,9 @@ public class BomPlugin implements Plugin<Project> {
 			Node dependencies = findChild(dependencyManagement, "dependencies");
 			if (dependencies != null) {
 				for (Node dependency : findChildren(dependencies, "dependency")) {
-					String groupId = findChild(dependency, "groupId").text();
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR but I don't think this change is necessary. There should always be `<groupId>` and `<artifactId>` nodes. If one or both of those nodes are missing then the bom that is being processed is an invalid state and we should fail hard as we currently do.

hyunmin0317:
Thank you very much for your thoughtful feedback.
I appreciate you taking the time to review the changes and share your insights.

PR #41851, opened 2024-08-14
Add support for Pulsar default tenant/namespace

onobc:
This commit allows Pulsar users to configure a default tenant and/or namespace to be used when producing or consuming messages to topic URLs that are not fully-qualified.

The following changes accomplish this:
- add `tenant` and `namespace` config props to `spring.pulsar.defaults`
- auto-configure a `PulsarTopicBuilder` bean populated w/ above props
- provide above topic builder to producer/consumer/reader factories (imperative and reactive)
- add tests for all of the above

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarConfiguration.java
@@ -182,9 +182,10 @@ class PulsarConfiguration {
 	@Bean
 	@Scope("prototype")
 	@ConditionalOnMissingBean
+	@ConditionalOnProperty(name = "spring.pulsar.defaults.topic.enabled", havingValue = "true", matchIfMissing = true)
 	PulsarTopicBuilder pulsarTopicBuilder() {
-		return new PulsarTopicBuilder(TopicDomain.persistent, this.properties.getDefaults().getTenant(),
-				this.properties.getDefaults().getNamespace());
+		return new PulsarTopicBuilder(TopicDomain.persistent, this.properties.getDefaults().getTopic().getTenant(),
+				this.properties.getDefaults().getTopic().getNamespace());
 	}
 
 }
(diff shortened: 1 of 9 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -250,13 +250,43 @@ public Authentication getAuthentication() {
...
+		 */
+		private String tenant;
+
+		/**
+		 * Default namespace to use when producing or consuming messages against a
+		 * non-fully-qualified topic URL. When not specified Pulsar uses a default
+		 * namespace of 'default'.
+		 */
+		private String namespace;
+
```
wilkinsona, line 266: As these properties both relate to topics, it would be good to put them both in a `topic` group so they'd become `spring.pulsar.defaults.topic.tenant` and `spring.pulsar.defaults.topic.namespace`. I think that still makes sense given that `typeMappings` is partially topic related. What do you think though, @cbono?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarConfiguration.java
```diff
@@ -176,4 +179,12 @@ PulsarFunctionAdministration pulsarFunctionAdministration(PulsarAdministration p
 				properties.isFailFast(), properties.isPropagateFailures(), properties.isPropagateStopFailures());
 	}
 
+	@Bean
+	@Scope("prototype")
```
wilkinsona, line 183: Why is this a prototype bean? Do the factories mutate the builder and need to do so independently?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
onobc: Great suggestion as this also gives way to adding an `enabled` flag to the `spring.pulsar.defaults.topic` to disable default topics if so desired.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarConfiguration.java
onobc: The factories only use the stateless `getFullyQualifiedName` method on the builder currently. However, this bean will also be used when users want to create `PulsarTopic` beans though. In that case, the builder has "name" and "numberOfPartitions" state that would vary during building.

The other alternative is to provide a "start building" method on the builder which would clone the builder. 

```java
public PulsarTopicBuilder forName(String name) {
    return this.cloneBuilder();
}
```

I saw other builders using prototype in SB so I went w/ that approach. I meant to call this point out during my self-review of the PR. 

Any preference?

wilkinsona: > However, this bean will also be used when users want to create PulsarTopic beans though

Thanks for the explanation. This sounds like a good reason for it to be a prototype.

PR #41876, opened 2024-08-15
Refactor BasicJsonParser to simplify JSON parsing logic

hyunmin0317:
This pull request refactors the `BasicJsonParser` class to simplify the JSON parsing logic.

The following changes are made:

- Consolidated `trimLeadingCharacter` and `trimTrailingCharacter` methods into a single `trimEdges` method to reduce code duplication and improve clarity.
- Extracted the number parsing logic into a new `parseNumber` method to separate concerns and enhance maintainability.

These changes aim to improve code readability and make future maintenance easier.
No functional changes are introduced.

Thank you for reviewing this pull request!

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/BasicJsonParser.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/BasicJsonParser.java
@@ -52,7 +52,7 @@ public class BasicJsonParser extends AbstractJsonParser {
 
 	private List<Object> parseListInternal(int nesting, String json) {
 		List<Object> list = new ArrayList<>();
-		json = trimLeadingCharacter(trimTrailingCharacter(json, ']'), '[').trim();
+		json = trimEdges(json, '[', ']').trim();
 		for (String value : tokenize(json)) {
 			list.add(parseInternal(nesting + 1, value));
 		}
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/BasicJsonParser.java
```diff
@@ -70,37 +70,37 @@ private Object parseInternal(int nesting, String json) {
...
 					"Expecting double-quotes around field names");
-			String key = trimLeadingCharacter(trimTrailingCharacter(values[0], '"'), '"');
+			String key = trimEdges(values[0], '"', '"');
 			Object value = parseInternal(nesting, values[1]);
 			map.put(key, value);
 		}
 		return map;
 	}
 
+	private Object parseNumber(String json) {
```
asashour, line 92: This method could be `static`.

wilkinsona: Thanks for trying to help, @asashour, but this suggestion doesn't align with our code style where, generally speaking, we only declare methods as static when they're used in a static context. `BasicJsonParser` has no instance variables so, in theory, every method other than those that are being overridden could be declared static but that just makes the code more verbose and adds some visual noise. It would be better leave this as it is.

hyunmin0317: Dear @asashour and @wilkinsona,

Thank you very much for taking the time to review my pull request.

I greatly appreciate your valuable feedback and suggestions!

mhalbritter:

PR #41904, opened 2024-08-16
updgraded couchbase dependency to 3.7.2

ukmastercom:
Upgraded couchbase client dependency to 3.7.2


snicoll:
Thanks @ukmastercom but, as noted in the PR template, we don't need PR for straightforward dependency upgrades like this one.

Issue #41905, opened 2024-08-16
Tests that set java.io.tmp dir may pollute java.io.File.TempDirectory.tmpdir with a directory that doesn't exist


wilkinsona:
I'm seeing this in `LogbackLoggingSystemTests` but it could happen in any test that sets `java.io.tmpdir`. The pollution happens like this:

1. A test's setup code sets`java.io.tmpdir` to a JUnit-provided temporary directory
2. Some code, a call to `Files.createTempFile` for example, causes `java.io.File.TempDirectory` to be loaded and its `tmpdir` `static` is set to the value of `java.io.tmpdir`
3. A test's cleanup code restores `java.io.tmpdir` to its original value and JUnit deletes its temporary directory
4. Some code tries to use `Files.createTempFile` and it fails because `java.io.File.TempDirectory.tmpdir` points to the directory that was deleted in step 3

Ideally, we'd avoid setting `java.io.tmpdir` in our tests. Failing that, we could ensure that `java.io.File.TempDirectory` has been initialized before we set it so that its `tmpdir` static doesn't get polluted with a custom directory that may only exist for the duration of a single test.

There's a similar problem with `java.nio.file.TempFileHelper.tmpdir` that's used by `java.nio.file.Files.createTempFile(String, String, FileAttribute<?>...)`

PR #41910, opened 2024-08-16
Skip redundant LoggerContext reset

mches:
Skips a redundant call to `LoggerContext#reset()`.

* Prior to Logback 1.5.7, calling `#stop()` on _any_ `LoggerContext` implicitly called `#reset()`. On the other hand, since Logback 1.5.7, only calling `#stop()` on a _started_ `LoggerContext` implicitly calls `#reset()`. Going forward, given the next version of Spring Boot will not be binary or source compatible with Logback 1.5.6 and earlier, it's only necessary to call `#reset()` on a _stopped_ `LoggerContext`.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystem.java
@@ -76,6 +76,7 @@ import org.springframework.util.StringUtils;
  * @author Dave Syer
  * @author Andy Wilkinson
  * @author Ben Hale
+ * @author Mark Chesney
  * @since 1.0.0
  */
 public class LogbackLoggingSystem extends AbstractLoggingSystem implements BeanFactoryInitializationAotProcessor {
@@ -210,7 +211,7 @@ public class LogbackLoggingSystem extends AbstractLoggingSystem implements BeanF
 			applySystemProperties(initializationContext.getEnvironment(), logFile);
 		}
 		LoggerContext loggerContext = getLoggerContext();
-		stopAndReset(loggerContext);
+		stopOrReset(loggerContext);
 		SpringBootJoranConfigurator configurator = new SpringBootJoranConfigurator(initializationContext);
 		configurator.setContext(loggerContext);
 		boolean configuredUsingAotGeneratedArtifacts = configurator.configureUsingAotGeneratedArtifacts();
@@ -223,7 +224,7 @@ public class LogbackLoggingSystem extends AbstractLoggingSystem implements BeanF
 	@Override
 	protected void loadDefaults(LoggingInitializationContext initializationContext, LogFile logFile) {
 		LoggerContext context = getLoggerContext();
-		stopAndReset(context);
+		stopOrReset(context);
 		withLoggingSuppressed(() -> {
 			boolean debug = Boolean.getBoolean("logback.debug");
 			if (debug) {
@@ -245,7 +246,7 @@ public class LogbackLoggingSystem extends AbstractLoggingSystem implements BeanF
 	protected void loadConfiguration(LoggingInitializationContext initializationContext, String location,
 			LogFile logFile) {
 		LoggerContext loggerContext = getLoggerContext();
-		stopAndReset(loggerContext);
+		stopOrReset(loggerContext);
 		withLoggingSuppressed(() -> {
 			if (initializationContext != null) {
 				applySystemProperties(initializationContext.getEnvironment(), logFile);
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot/src/test/java/org/springframework/boot/logging/logback/LogbackLoggingSystemTests.java
```diff
@@ -93,8 +94,9 @@
  * @author Scott Frederick
  * @author Jonatan Ivanov
  * @author Moritz Halbritter
+ * @author Mark Chesney
  */
-@ExtendWith(OutputCaptureExtension.class)
+@ExtendWith({ MockitoExtension.class, OutputCaptureExtension.class })
```
philwebb, line 99: Is this change accidental? It doesn't seem related to the other changes?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackConfigurator.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2024 the original author or authors.
```
philwebb, line 2: There doesn't seem to be any changes in this file so the copyright header should not be updated.

philwebb (commented):
Thanks for the PR. I've added a couple of issues to review.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/logging/logback/LogbackLoggingSystemTests.java
mches: This appeared in #41885, which was reverted.

Without this, I ran into [this issue](https://github.com/mockito/mockito/issues/1379#issuecomment-1967392526) in my environment, which failed the tests that use `Mockito.mock` within this class for being unable to delete `mockitoboot*.jar` from the JUnit temp directory. 

With the change, the test is less finicky on Windows.

That said, I can take it out if required.

philwebb:
> Going forward, given the next version of Spring Boot will not be binary or source compatible with Logback 1.5.6 and earlier

Is this the case? Is Logback 1.5.7 not binary compatible with Logback 1.5.6?

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackConfigurator.java
mches: This file was updated in #41885, which was reverted, but updated again without updating the years. So I feel it's accurate, but I'm happy to take it out if required.

mches:
Thanks for the review 🙇‍♂️

mches:
> > Going forward, given the next version of Spring Boot will not be binary or source compatible with Logback 1.5.6 and earlier
> 
> 
> 
> Is this the case? Is Logback 1.5.7 not binary compatible with Logback 1.5.6?

Yes, because the return type of `LoggerContext#getConfigurationLock()` changed from `Object` to `ReentrantLock`. It's source compatible, but not binary compatible.

snicoll:
So this is in addition of the [actual upgrade](https://github.com/spring-projects/spring-boot/issues/41887) to Logback then?

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/logging/logback/LogbackLoggingSystemTests.java
wilkinsona: I don't think it's needed. Regardless, it isn't related to this change so I think it should be taken out. https://github.com/spring-projects/spring-boot/issues/41905 has hopefully helped here.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackConfigurator.java
wilkinsona: I should have bumped the year as part of https://github.com/spring-projects/spring-boot/commit/dd2ed5f12d23b15ff0055e366ab6da8e95218596. That should now be done separately as it isn't related to this change so please take it out of this PR.

wilkinsona (changes requested):

mches:
> So this is in addition of the [actual upgrade](https://github.com/spring-projects/spring-boot/issues/41887) to Logback then?

Yes, I thought this made sense to do after reviewing dd2ed5f.

mches:
Thank you for the review. All feedback has been implemented.

wilkinsona:
I think we should leave this as-is. Arguably, calling both `stop()` and then `reset()` has been unnecessary for some time as in 1.5.6 and earlier versions a call to `stop()` will also call `reset()` but it hasn't caused any problems.

In 1.5.7, `stop()` will only call `reset()` if the context is started. A context is started in its constructor so, if it isn't started, it must have already been stopped and, therefore, must have already been reset. I think that means that we could, perhaps, do nothing at all if the context isn't started. It's hard to justify making a change as there doesn't appear to be any benefit yet there is some risk in making the change. I'm going to close this one, for now at least. We can reconsider in the future and perhaps do nothing at all if the context isn't started. Thanks anyway, @mches.

PR #41911, opened 2024-08-16
Disable Artemis tests on Java 23

### LINKED ISSUE #41909: Avoid using JRE.OTHER when disabling Artemis tests to work around ARTEMIS-4975
Apache Artemis is incompatible with Java 23: https://issues.apache.org/jira/browse/ARTEMIS-4975. We're seeing failures on `main` due to this. I'm not sure why we're not also seeing them on our maintenance branches. We'll need to work around this where necessary. I think we have two options:

1. Disable the Artemis tests on Java 23
2. Configure the relevant test task(s) to run with `-Djava.security.manager=allow`

I think I prefer 1, as 2 would mask problems with other dependencies.

cms04:
I completely disabled the following test on Java 23:
- `ArtemisAutoConfigurationTests`
- `ArtemisEmbeddedConfigurationFactoryTests`
- `ArtemisEnvironmentTests`
- `JmsHealthContributorAutoConfigurationTests`

Besides that I disabled the tests with Artemis context in the following test class:
- `JmsAutoConfigurationTests`

Because the `org.junit.jupiter.api.condition.JRE` does not contain a `JAVA_23` enum value yet, I used the `@EnabledForJreRange` annotation and not the `@DisabledOnJre` annotation.

This should close #41909

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/jms/JmsHealthContributorAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/jms/JmsHealthContributorAutoConfigurationTests.java
@@ -17,6 +17,8 @@
 package org.springframework.boot.actuate.autoconfigure.jms;
 
 import org.junit.jupiter.api.Test;
+import org.junit.jupiter.api.condition.EnabledForJreRange;
+import org.junit.jupiter.api.condition.JRE;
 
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks, @cms04, but this is broader than hopefully necessary. As far as we know, it's only `ArtemisAutoConfigurationTests` that's affected and we just need to adapt its condition know that we've upgraded to a version of JUnit Jupiter where `JRE.OTHER` no longer matches Java 23. If you're interested, please see https://github.com/spring-projects/spring-boot/issues/41909#issuecomment-2295972695 for a longer explanation of what was happening.

Issue #41914, opened 2024-08-18
Javadoc of slice test annotations should describe more accurately which components are considered


Narkissus:
Spring Boot version: 3.3.2

According to the javadoc on `org.springframework.boot.test.autoconfigure.webservices.server.WebServiceServerTest` annotation, it should configure the context in a way that webservice relevant beans are present - i.e. beans of type `Endpoint` and `EndpointInterceptor`. 
During my testing I have noticed that the `EndpointInterceptor` types are absent from the application context during tests while using `@WebServiceServerTest`.
The current workaround is to import the configuration class where the `EndpointInterceptor` beans are defined.

You can find a minimal reproducible example code at https://github.com/Narkissus/WebServiceServerTest-issue-example

snicoll:
@Narkissus thanks for the sample.

>  it should configure the context in a way that webservice relevant beans are present - i.e. beans of type Endpoint and EndpointInterceptor.

I think that the word "scan" is missing in the Javadoc of this slice test for some reason. The intent is that component scanning will only pick up beans annotated with `@Endpoint` or of type `EndpointInterceptor`. Your interceptors are configured in a configuration class, slice tests won't selectively pick up `@Bean` methods as you seem to expect.

We'll get the Javadoc fixed. I've submitted a [PR to your sample](https://github.com/Narkissus/WebServiceServerTest-issue-example/pull/1) that fixes the test and provide more context.

snicoll:
I've extended the scope of this issue as other slice tests suffer from the same problem.

PR #41930, opened 2024-08-19
Update Javadoc for WebServiceServerTest annotation gh-41914

KonarzewskiP:
Closes gh-41914

Update Javadoc according to the following suggestion 
image


wilkinsona:
Thanks, @KonarzewskiP, but as indicated by the issue's assignment, #41914 is already being worked on by @snicoll.

Issue #38754, opened 2023-12-12
SslBundle - Allow loading PEM encoded trusted certificates from multiple files


bgK:
At the moment, PEM `SslBundle`s can be instantiated through the following properties:
```yml
spring:
  ssl:
    bundle:
      pem:
        client:
          truststore:
            certificate: "classpath:client.crt"
```
Where `client.crt` can contain multiple certificates.

In some situations, multiple very different certificates need to be trusted. For instance:
* An external service changes the CA it uses to sign the certificate it exposes. Both the old one and the new one need to be trusted on the client side to avoid downtime.
* When exposing a service with mutual authentication enabled, and when the corporate CA is re-used for multiple systems, it can be somewhat easy to obtain a keypair for a given FQDN. Trusting the CA is not an option, the trust store must contain all the allowed certificates.

While concatenating all the trusted certificates in the same file is an option, it makes it quite hard to see at a glance which certificates are trusted, as they are PEM-encoded.
It would be nice to be able to use file names to identify the certificates:

```yml
spring:
  ssl:
    bundle:
      pem:
        client:
          truststore:
            certificates:
            - "classpath:allowed-client1.crt"
            - "classpath:allowed-client2.crt"
```

philwebb:
There might be some overlap here with #38242 where we are looking to support directory glob patterns. If the order of the certificates isn't important we might be able to support something like this:

```yaml
spring:
  ssl:
    bundle:
      pem:
        client:
          truststore:
            certificate: "/my/certs/allowed-*.crt"
            select: all
```

bgK:
I hadn't seen #38242. Indeed, it fills almost the same need. It's nice in that the properties are backwards compatible and makes it easy to have large numbers of certificates.

Here's a use case that I don't think is easily covered by #38242. A Spring Boot app deployed in Kubernetes. It calls an external service for which some instances are hosted in the same cluster, some instances are scaled out to an external provider. The internal instances are exposed using the Kubernetes CA, the other instances use some other CA. Kubernetes bind mounts the CA at a fixed location inside the pod `/var/run/secrets/kubernetes.io/serviceaccount/ca.crt`:

```yml
spring:
  ssl:
    bundle:
      pem:
        client:
          truststore:
            certificates:
            - "/var/run/secrets/kubernetes.io/serviceaccount/ca.crt"
            - "classpath:other-ca.crt"
```

IMO, me both #38754 and this are good enhancements.

dopsun:
[...]

PR #41932, opened 2024-08-19
Add support for optional trust certificates

elisabeteaprcd:
**Problem Statement**

The challenge is the dynamic updating and management of certificates in an application that supports mTLS and interacts with client services that are frequently added and removed.

Suppose your application needs to support mTLS with two well-known client services, `foo` and `bar`, each having a specific intermediate CA (Certificate Authority). Client services are dynamically installed/removed from the cluster. For improved resiliency/manageability, you do not want to change the deployment of your application whenever a new client service is installed in the cluster: the client CAs must be dynamically added/removed to your server truststore as their services installed/removed from the cluster.

**Proposed Solution**

This PR aims to address this issue by allowing certificates to be marked as optional. With this feature, the application can start and continue running regardless of the availability of these certificates.

**Current Status**

This pull request is currently a draft/proof of concept. It does not yet include tests or documentation and is intended to gather feedback. The implementation includes several TODOs, such as:

- Adjust the validation for cases that involve PEM content used directly.

An example of `application.yaml` with optionality would look like:

```
spring.ssl.bundle.pem.server.reload-on-update=true
spring.ssl.bundle.pem.server.truststore.certificate=optional:/var/run/secrets/foo/ca.crt

```

**Related Issue**

This PR is related to this [comment](https://github.com/spring-projects/spring-boot/issues/38754#issuecomment-2274269939) in the [issue #38754](https://github.com/spring-projects/spring-boot/issues/38754)

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/BundleContentProperty.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/BundleContentProperty.java
@@ -35,6 +35,8 @@ import org.springframework.util.StringUtils;
  */
 record BundleContentProperty(String name, String value) {
 
+	private static final String OPTIONAL_URL_PREFIX = "optional:";
+
 	/**
 	 * Return if the property value is PEM content.
 	 * @return if the value is PEM content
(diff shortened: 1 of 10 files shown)


scottfrederick:
@elisabeteaprcd Thanks for the suggestion. We will take a look at the changes and give some feedback. 

An alternative to these changes would be a boolean property at the top of the bundle structure such as `spring.ssl.bundle.pem.server.required: false` or `spring.ssl.bundle.pem.server.ignore-non-existing: true` that would provide the same "optional" behavior for all certificates in the bundle, without needing to introduce a new `optional:` keyword to the certificate property value. Would that also meet your needs?

chicobento:
Hi @scottfrederick, 

Im also working with @elisabeteaprcd  on this. While the scope of this change is to support optionality for the single item - which is what is currently supported by SB, the end goal is to have it working with multiple items as proposed in https://github.com/spring-projects/spring-boot/issues/38754#issuecomment-2274269939.

So, in the scenario of multiple CAs we should be able to define which ones are mandatory for the service to start and which ones are optional, and in this sense your single-property proposal would not work at the spring boot level, however it could be expressed at the k8s deployment level by making the volume mounts mandatory.

chicobento:
hi @scottfrederick , just a heads-up that this has been now updated with support for multiple/optional certificates (initial PR had just support for optional)

wilkinsona:
Is this PR ready for review? It's been in draft state for over a year now.

wilkinsona:
Closing due to lack of feedback.

PR #41934, opened 2024-08-19
Update MyUserHandler.java

ritzykey:
markdown error

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
@@ -957,9 +957,9 @@ For example, the configuration property `spring.main.log-startup-info` would be 
 
 Environment variables can also be used when binding to object lists.
 To bind to a `List`, the element number should be surrounded with underscores in the variable name.
-
 For example, the configuration property `my.service[0].other` would use an environment variable named `MY_SERVICE_0_OTHER`.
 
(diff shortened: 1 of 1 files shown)


snicoll:
@ritzykey what do you mean by "markdown error"? As far as I understand it, the purpose of `/**/` is to [replace irrelevant body with `...`](https://docs.spring.io/spring-boot/reference/web/reactive.html#web.reactive.webflux)

snicoll:
@ritzykey what browser and what link is that? This is what I have on Firefox for https://docs.spring.io/spring-boot/reference/web/reactive.html#web.reactive.webflux

image

wilkinsona:
Thanks, @ritzykey. The code that you have changed is for https://docs.spring.io/spring-boot/reference/web/reactive.html where the problem does not occur. The code for https://docs.spring.io/spring-boot/reference/web/servlet.html is 

https://github.com/spring-projects/spring-boot/blob/06c008832254b955b2c7318abe4678b7160700c6/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/web/servlet/springmvc/MyUserHandler.java#L24-L39

Would you like to update this class instead? The fix is to remove the newlines between `/**/` and the `return` statement so that everything's on one line as it is in the reactive variant:

https://github.com/spring-projects/spring-boot/blob/06c008832254b955b2c7318abe4678b7160700c6/spring-boot-project/spring-boot-docs/src/main/java/org/springframework/boot/docs/web/reactive/webflux/MyUserHandler.java#L26-L38

snicoll:
superseded by #41948

PR #41948, opened 2024-08-19
Harmonize code sample for MyUserHandler in reference documentation

ritzykey:
fix markdown problem

image


snicoll:
Thanks. I used the issue to apply the same to the Kotlin variant of the sample, see [8243c7b](https://github.com/spring-projects/spring-boot/commit/8243c7ba98254c327931520ce3ae1f86c669441d)

PR #41975, opened 2024-08-21
Use enhanced switch statement

PiyalAhmed:
Use enhanced switch statement


wilkinsona:
Thanks for the proposal but the source in spring-boot/spring-boot-project/spring-boot-tools/spring-boot-configuration-processor/src/json-shade is copied from elsewhere and we prefer to keep it also close to the original as we can.

PR #42019, opened 2024-08-24
Use List.copyOf() instead of Collections.unmodifiableList()

bazzi2548:
We can safely use List.copyOf() because we check if (providers) is null.

If use List.copyOf()
- Clean code
- Performance optimization
- Guarantees a truly immutable list.


philwebb:
I think this is probably safe, but we need to be a little careful using `List.copyOf`. Unlike `Collections.unmodifiableList(new ArrayList<>(...))` it will throw an exception if any of the elements are `null`.

philwebb:
Thanks @bazzi2548!

PR #42020, opened 2024-08-25
Using configureEach instead of all to avoid unnecessary configuration

quaff:
Resolving all configurations is usually bad practice, because it might trigger work that could be avoided, which will make the build slower.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -20,7 +20,7 @@ allprojects {
 		}
 	}
 
-	configurations.all {
+	configurations.configureEach {
 		resolutionStrategy.cacheChangingModulesFor 0, "minutes"
 	}
 }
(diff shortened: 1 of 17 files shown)


wilkinsona:
For tasks, using `configureEach` is recommended as it avoids eager creation of tasks that have been [lazily registered](https://docs.gradle.org/current/userguide/lazy_configuration.html). As far as I know, the same does not apply to configurations as there's no support for lazy registration. Furthermore, I don't think changing from `all` to `configureEach` has any bearing on whether or not the configuration is resolved and Gradle's own documentation uses `configurations.all` [in places](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/ResolutionStrategy.html). Given this, I'm going to close this one as I don't think it will bring any noticeable benefit. Thanks anyway.

PR #42049, opened 2024-08-28
Apply instanceof pattern matching

abc5259:
Refactored the code to use instanceof pattern matching, replacing the older style of instanceof checks and casts. This change improves code readability and maintainability by reducing boilerplate code.

### Changes
- Applied `instanceof` pattern matching in relevant sections of the code where it was previously missing.
- Removed unnecessary explicit casting.


philwebb:
Thanks @abc5259

PR #42052, opened 2024-08-28
Add support for configuring Pulsar client IO and listener threads

vpavic:
This commit adds configuration properties that allow users to configure number of IO threads and listener threads used by the Pulsar client.

---

From my experience with Pulsar, using these settings is often needed to achieve the optimal setup of the Pulsar client so ideally Spring Boot's auto-config would provide a way to do this using configuration properties.

/cc @onobc

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
@@ -44,6 +44,7 @@ import org.springframework.util.Assert;
  * @author Chris Bono
  * @author Phillip Webb
  * @author Swamy Mavuri
+ * @author Vedran Pavic
  * @since 3.2.0
  */
 @ConfigurationProperties("spring.pulsar")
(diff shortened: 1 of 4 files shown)


onobc:
> From my experience with Pulsar, using these settings is often needed to achieve the optimal setup of the Pulsar client so ideally Spring Boot's auto-config would provide a way to do this using configuration properties.

Thanks for the contribution @vpavic . When we brought the auto-configuration into Spring Boot we greatly reduced the number of configuration properties we supported. We wanted the community to let us know which properties were important. Thank you for doing that here. Let's see what the Boot team thinks.

philwebb:
It sounds like they'd be useful for a lot of apps so I'm in favor of adding them. I wonder if we should put them under a nested `Threads` class (similar to `server.tomcat.thread.<name>`)?

onobc:
> It sounds like they'd be useful for a lot of apps so I'm in favor of adding them. I wonder if we should put them under a nested `Threads` class (similar to `server.tomcat.thread.<name>`)?

I think grouping them under `thread` would be a good forward-thinking idea. The naming here is interesting....

- `ioThreads` -> `thread.num-io` **or** `thread.io` ?
- `listenerThreads` -> `thread.num-listener` **or** `thread.listener` ?

philwebb:
I would suggest something similar to [existing tomcat ones](https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.server.server.tomcat.threads.max). `threads.io` and `threads.listener` would be my suggestion.

vpavic:
I've updated the PR accordingly.

philwebb:
Thanks @vpavic! Merged into main.

PR #42056, opened 2024-08-29
Allow to enable the native profile via Maven property

obfischer:
In some cases it could be desired to enable the `native` profile directly from Maven, so that the caller of Maven does not need to call Maven with `-Pnative`. To enable this behavior, I added an activation section to the generated POM with the profile definition.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/how-to/pages/native-image/developing-your-first-application.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/how-to/pages/native-image/developing-your-first-application.adoc
@@ -210,6 +210,15 @@ With the `native` profile active, you can invoke the `native:compile` goal to tr
 $ mvn -Pnative native:compile
 ----
 
+The `native` profile can be activated by default by setting the Maven property `native` to `true` in the POM.
+
+[source,xml]
+----
+<properties>
+  <native>true</native>
+</properties>
+----
+
 The native image executable can be found in the `target` directory.
 
 
--- a/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
+++ b/spring-boot-project/spring-boot-starters/spring-boot-starter-parent/build.gradle
@@ -248,6 +248,12 @@ publishing.publications.withType(MavenPublication) {
 			profiles {
 				profile {
 					delegate.id("native")
+					activation {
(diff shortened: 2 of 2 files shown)


philwebb:
Seems sensible to me. Our resident Maven expert is away at the moment. I'll flag this one for team attention to see if anyone can foresee any problems.

Blackboy007are:
obfischer:improvement/allow-to-activate-native-from-within-the-POM

Blackboy007are:
obfischer:improvement/allow-to-activate-native-from-within-the-POM

snicoll:
>  so that the caller of Maven does not need to call Maven with -Pnative

Can you expand a bit more on that? We don't really want to expand what our parent does without a good reason of doing so. Also I am not sure it would work in child projects as setting the property would be too late for the profile to evaluate. Have you tested that?

Blackboy007are:
obfischer:improvement/allow-to-activate-native-from-within-the-POM

Blackboy007are:
bb7480279@gmail.com

Blackboy007are:
import org.springframework.boot.*;
import org.springframework.boot.autoconfigure.*;
import org.springframework.web.bind.annotation.*;

@RestController
@SpringBootApplication
public class Example {

	@RequestMapping("/")
	String home() {
		return "Hello World!";
	}

		SpringApplication.run(Example.class, args);
	}

}

Blackboy007are:
import org.springframework.boot.*;
import org.springframework.boot.autoconfigure.*;
import org.springframework.web.bind.annotation.*;

@RestController
@SpringBootApplication
public class Example {

	@RequestMapping("/")
	String home() {
		return "Hello World!";
	}

		SpringApplication.run(Example.class, args);
	}

}

philwebb:
@Blackboy007are Will you please stop spamming this issue

snicoll:
I am going to close this as what's been suggested does not work for me. If the feedback is provided and it turns out I was wrong we can reopen this one.

obfischer:
Hi, sorry, currently I am quite busy. I will come back with a more details description.

Blackboy007are:
42056

၂၀၂၄၊ စက် ၁၇၊ အင်္ဂါ နံနက် ၃:၁၇ တွင် Oliver B. Fischer <
***@***.***> က ရေးခဲ့သည်−

PR #42062, opened 2024-08-30
Add support for configuring Pulsar listener container concurrency

vpavic:
This commit adds configuration property that allows users to configure Pulsar message listener container concurrency.

---

In similar fashion to https://github.com/spring-projects/spring-boot/pull/42052, this is much needed to achieve optimal setup of message consumption, especially when dealing with partitioned topics.

/cc @onobc

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
@@ -812,6 +812,11 @@ public class PulsarProperties {
 		private SchemaType schemaType;
 
 		/**
+		 * Number of threads used by listener container.
+		 */
+		private Integer concurrency;
+
+		/**
 		 * Whether to record observations for when the Observations API is available and
 		 * the client supports it.
 		 */
@@ -823,6 +828,14 @@ public class PulsarProperties {
 
 		public void setSchemaType(SchemaType schemaType) {
 			this.schemaType = schemaType;
+		}
+
+		public Integer getConcurrency() {
+			return this.concurrency;
+		}
+
+		public void setConcurrency(Integer concurrency) {
+			this.concurrency = concurrency;
 		}
 
 		public boolean isObservationEnabled() {
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
@@ -39,6 +39,7 @@ import org.apache.pulsar.client.impl.AutoClusterFailover.AutoClusterFailoverBuil
 
 import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.boot.json.JsonWriter;
+import org.springframework.pulsar.config.ConcurrentPulsarListenerContainerFactory;
 import org.springframework.pulsar.core.PulsarTemplate;
 import org.springframework.pulsar.listener.PulsarContainerProperties;
 import org.springframework.pulsar.reader.PulsarReaderContainerProperties;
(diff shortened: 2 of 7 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -801,6 +802,11 @@ public static class Listener {
 		 */
 		private SchemaType schemaType;
 
+		/**
+		 * Number of threads used by listener container.
```
onobc, line 806: It may be helpful to elaborate on this one just a bit, maybe something like:
> Number of listener threads used by the container to process messages concurrently

Wdyt?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
```diff
@@ -195,6 +197,13 @@ private void customizePulsarContainerListenerProperties(PulsarContainerPropertie
 		map.from(properties::isObservationEnabled).to(containerProperties::setObservationEnabled);
 	}
 
+	<T> void customizeConcurrentPulsarListenerContainerFactory(
```
onobc, line 200: I just noticed that the listener containers on the framework side are inconsistent w/ the `concurrency` attribute between reactive and imperative. I created [this issue to simplify the latter](https://github.com/spring-projects/spring-pulsar/issues/820). 
If you wanted to handle that issue first it would then simplify this PR so that we would not need this new `customizeConcurrentPulsarListenerContainerFactory` method and could simply map the property in `customizePulsarContainerListenerProperties` instead.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -801,6 +802,11 @@ public static class Listener {
...
+		 */
+		private Integer concurrency;
```
onobc, line 808: Can you add this support to the Reactive side as well?

See https://github.com/spring-projects/spring-boot/blob/fa686bb593f2a3b85497546d3995c4050501ef61/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarReactivePropertiesMapper.java#L92

onobc (changes requested):
Thanks for the contribution @vpavic . I believe this is another property that makes sense to make available via config props. 

I have a couple of comments/suggestions to follow up on.

Thanks

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
vpavic: Sure, I can pick up that issue on Spring Pulsar side. So basically this PR is then blocker until https://github.com/spring-projects/spring-pulsar/issues/820 is resolved.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
vpavic: I've addressed this in the latest PR update.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
onobc: > Sure, I can pick up that issue on Spring Pulsar side. So basically this PR is then blocker until [spring-projects/spring-pulsar#820](https://github.com/spring-projects/spring-pulsar/issues/820) is resolved.

Thanks @vpavic . Yeh, I think we should do the spring-pulsar one first.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
onobc: >  tried to aligned property description with these

Yep, I saw those descriptions as well. I think there is room for improvement in the spring-kafka properties. This is really a nit and consistency is a good thing. As you said, let's see what Boot peepz think.

philwebb:
Thanks @vpavic!

vpavic:
@philwebb this was merged perhaps too soon as I didn't have chance to update this PR to benefit from the changes introduced in https://github.com/spring-projects/spring-pulsar/issues/820.

I've opened https://github.com/spring-projects/spring-boot/pull/42120 to address that separately.

PR #42067, opened 2024-08-30
Apply spring.pulsar.consumer.subscription.name to auto-configured Pulsar listener container

### LINKED ISSUE #42053: Clarify purpose of Pulsar consumer subscription related properties
Some while ago I opened these two issues against the Spring Pulsar project:

- https://github.com/spring-projects/spring-pulsar/issues/480
- https://github.com/spring-projects/spring-pulsar/issues/488

After setting up another project recently that uses Spring Pulsar for consuming messages, I've again tripped up on the same thing attempting to set Pulsar consumer subscription related settings using Spring Boot provided configuration properties.

Though Spring Pulsar reference manual now contains the following two tips:

> The `spring.pulsar.consumer.subscription.name` property is ignored and is instead generated when not specified on the annotation. 

> The `spring.pulsar.consumer.subscription-type` property is ignored and is instead taken from the value on the annotation. However, you can set the `subscriptionType = {}` on the annotation to instead use the property value as the default.

IMO it would be good to revisit these properties as their purpose is (at least to me) unclear at the moment.

/cc @onobc

onobc:
The subscription name config prop was not being set on the Pulsar listener container properties. This commit adds the subscription name to the Pulsar property mappers.

Resolves #42053

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
@@ -186,6 +186,7 @@ final class PulsarPropertiesMapper {
 		PulsarProperties.Consumer.Subscription properties = this.properties.getConsumer().getSubscription();
 		PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
 		map.from(properties::getType).to(containerProperties::setSubscriptionType);
+		map.from(properties::getName).to(containerProperties::setSubscriptionName);
 	}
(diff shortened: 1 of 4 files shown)


wilkinsona:
Should we be treating this as a bug, @onobc? The property and the API to apply it seem to be there in Boot 3.2. Should we apply the change to 3.2.x and later or only to main?

vpavic:
My understanding is that this depends on changes Chris did in https://github.com/spring-projects/spring-pulsar/pull/821 meaning this is a 3.4 target.

onobc:
@vpavic summary is correct. The property is there and still does configure the default consumer factory but the property is ignored when it comes to the message container (i.e. `@PulsarListener`). The Spring Pulsar docs do point this fact out. I consider this an improvement so the property is respected in all applicable areas.

PR #42074, opened 2024-08-31
Refactor keep-alive thread implementation in SpringApplication

yeong-hwan:
### Refactor keep-alive thread implementation in SpringApplication

- Replaced the inline thread implementation in startKeepAliveThread with a dedicated keepAlive method.
- Improved readability and maintainability of the code by separating the keep-alive logic into its own method.
- Enhanced interrupted status handling by restoring the interrupted state when an InterruptedException occurs.

This change simplifies the thread management logic and adheres to best practices for thread interruption handling.

**Advantages:**
- Interruption Propagation: The interrupted status is preserved, allowing higher-level code to handle the interruption appropriately.
- Thread Safety: This approach ensures that the thread's state is accurately reflected and can be managed properly throughout its lifecycle.

**Expected Disadvantages:**
- Deferred Handling: If the interrupt is not handled immediately, it might require more logic to determine the appropriate next steps.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplication.java
@@ -1670,20 +1670,23 @@ public class SpringApplication {
 		}
 
 		private void startKeepAliveThread() {
-			Thread thread = new Thread(() -> {
-				while (true) {
-					try {
-						Thread.sleep(Long.MAX_VALUE);
-					}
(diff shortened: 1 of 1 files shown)


philwebb:
Thanks for the suggestions, but I don't think these changes are necessary for this specific piece of code. It's purpose is only to keep the application alive and on first interruption it should exit.

I'd like a second opinion from @wilkinsona, but I think we probably leave things as they are.

wilkinsona:
On balance, I think this makes the code harder to read. Thanks anyway.

PR #42078, opened 2024-09-01
Fix incorrect command in Docker configuration for Colima

martinfrancois:
As the dot is currently included in the backticks in the docs, this suggests that it should be included in the commend that should be executed as well. However, this would be incorrect, as it would result in a path that would end with a dot as well.

I put the dot after the backticks to make it clear it doesn't belong to the command. As the dot was always present in other similar instances in the docs I kept it, but if it wasn't for consistency I would be inclined to remove it entirely to make it even more foolproof.


philwebb:
Thanks @martinfrancois, for even more clarity we could do something like this:

```adoc
The plugin can communicate with the Docker daemon provided by https://github.com/abiosoft/colima[Colima].
The `DOCKER_HOST` environment variable can be set by using the following command:

[source,shell,subs="verbatim,attributes"]
----
$ export DOCKER_HOST=$(docker context inspect colima -f '{{.Endpoints.docker.Host}}')
----
```

What do you think? Are you interested in updating your PR?

martinfrancois:
You're welcome, and thanks as well @philwebb, I really like the idea, makes it even clearer without introducing inconsistencies. I changed it as you suggested.

philwebb:
Thanks again @martinfrancois, this has now been merged.

Issue #4665, opened 2015-12-02
Please make "banner.location" and default value public available


marceloverdijk:
`SpringApplication` now uses the below to locate the banner and use the default banner:

```
String location = environment.getProperty("banner.location", "banner.txt");
```

Could both the property name and default value be moved to a public constant?
This is mainly for personal reasons in a custom `EnvironmentPostProcessor` where I check for existence for the banner.location value or the presence of the default banner. If not the `EnvironmentPostProcessor` defaults to company wide banner. 
Currently I have the property and default value hardcoded in my class but it would be nice to refer to public constant.

marceloverdijk:
Thx!

PR #42084, opened 2024-09-02
Remove unused constants from SpringApplication.java

mushroom528:
- Remove an unused constants in `SpringApplication`.


wilkinsona:
Thanks for the suggestion but these constants are public API. Removing them would undo https://github.com/spring-projects/spring-boot/issues/4665 which we do not want to happen.

PR #42091, opened 2024-09-02
Improve documentation in "Command-line Completion"

Alchemik:
Replaced deprecated Spring CLI options with the current ones, also improved section related to completion setup for zsh.


wilkinsona:
Thanks very much for making your first contribution to Spring Boot, @Alchemik.

PR #42095, opened 2024-09-03
Polish

arefbehboudi:
- Remove any unnecessary exception throws to streamline the code.
-  Use Lambda Expression

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/view/MustacheView.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/view/MustacheView.java
@@ -66,7 +66,7 @@ public class MustacheView extends AbstractTemplateView {
 	}
 
 	@Override
-	public boolean checkResource(Locale locale) throws Exception {
+	public boolean checkResource(Locale locale) {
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/view/MustacheView.java
```diff
@@ -66,7 +66,7 @@ public void setCharset(String charset) {
 	}
 
 	@Override
-	public boolean checkResource(Locale locale) throws Exception {
```
wilkinsona, line 69: This is a breaking change as the method is public API. Please revert.

wilkinsona (changes requested):

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/view/MustacheView.java
arefbehboudi: Thank you for the review. I’ve updated the code to address the points you mentioned. Please have another look when you have a chance.

wilkinsona (approved):

PR #42173, opened 2024-09-06
cosmetic change, remove 2 imports

muybien3433:
Hi, some cosmetic changes, by that I was able to cut some lines :)


wilkinsona:
Thanks for the suggestion but we prefer things as they are. See https://github.com/spring-projects/spring-boot/pull/38352.

PR #42180, opened 2024-09-08
Add Pulsar config props for startup policy

onobc:
This commit adds properties to configure the startup behavior for the Pulsar message containers  that back the `@PulsarListener`, `@ReactivePulsarListener`, and `@PulsarReader`.

This is to support the recently added [startup policy feature](https://github.com/spring-projects/spring-pulsar/commit/a518e110893772a86239a1252b27126c28286188) in Spring for Apache Pulsar.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
@@ -854,6 +854,7 @@ public class PulsarProperties {
 		public Startup getStartup() {
 			return this.startup;
 		}
+
 	}
 
 	public static class Reader {
@@ -932,6 +933,7 @@ public class PulsarProperties {
 		public Startup getStartup() {
 			return this.startup;
 		}
+
 	}
 
 	public static class Startup {
@@ -961,6 +963,7 @@ public class PulsarProperties {
 		public void setOnFailure(FailurePolicy onFailure) {
 			this.onFailure = onFailure;
 		}
+
 	}
 
 	public enum FailurePolicy {
@@ -973,6 +976,7 @@ public class PulsarProperties {
 
 		/** Retry startup asynchronously. */
 		RETRY;
+
 	}
 
 	public static class Template {
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
@@ -822,6 +822,11 @@ public class PulsarProperties {
 		 */
 		private boolean observationEnabled;
 
+		/**
+		 * Startup settings.
+		 */
+		private final Startup startup = new Startup();
+
 		public SchemaType getSchemaType() {
 			return this.schemaType;
 		}
@@ -846,6 +851,9 @@ public class PulsarProperties {
 			this.observationEnabled = observationEnabled;
 		}
 
+		public Startup getStartup() {
+			return this.startup;
+		}
 	}
 
 	public static class Reader {
@@ -875,6 +883,11 @@ public class PulsarProperties {
 		 * backlog of a topic.
 		 */
 		private boolean readCompacted;
+
+		/**
+		 * Startup settings.
+		 */
+		private final Startup startup = new Startup();
 
 		public String getName() {
 			return this.name;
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -916,6 +930,53 @@ public void setReadCompacted(boolean readCompacted) {
...
+		public Startup getStartup() {
+			return this.startup;
+		}
+
+	}
+
+	public static class Startup {
+
+		/**
+		 * The max time to wait for the container to start.
```
snicoll, line 942: Description of configuration properties do not start with `the`, `a`, etc. I suggest `Time to wait for the container to start.`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -916,6 +930,53 @@ public void setReadCompacted(boolean readCompacted) {
...
+		 */
+		private Duration timeout;
+
+		/**
+		 * The action to take when the container fails to start.
```
snicoll, line 947: Same. I suggest `Action to take when the container fails to start.`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -916,6 +930,53 @@ public void setReadCompacted(boolean readCompacted) {
...
+		}
+
+	}
+
+	public static class Startup {
+
+		/**
+		 * The max time to wait for the container to start.
+		 */
+		private Duration timeout;
```
snicoll, line 944: Is there a default for this property? If so I think we should use the opportunity to set the default so that it's documented, and very in a test that it's consistent in case it changes.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarProperties.java
```diff
@@ -916,6 +930,53 @@ public void setReadCompacted(boolean readCompacted) {
...
+
+		/**
+		 * The action to take when the container fails to start.
+		 */
+		private FailurePolicy onFailure;
```
snicoll, line 949: I understand the previous behavior was to continue and log an exception. There's no default value here so it's hard to see if that's still the case. If that's the case, this field should have a default value that matches the default behavior.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
```diff
@@ -214,6 +227,17 @@ void customizeReaderContainerProperties(PulsarReaderContainerProperties readerCo
 		PulsarProperties.Reader properties = this.properties.getReader();
 		PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
 		map.from(properties::getTopics).to(readerContainerProperties::setTopics);
+		customizeReaderStartupProperties(readerContainerProperties);
+	}
+
+	private void customizeReaderStartupProperties(PulsarReaderContainerProperties readerContainerProperties) {
```
snicoll, line 233: It's unfortunate that this is the exact same code as `customizeListenerStartupProperties`? Since the feature is shared, isn't there a way to avoid the code duplication?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarReactivePropertiesMapper.java
```diff
@@ -96,6 +98,16 @@ private void customizePulsarContainerListenerProperties(ReactivePulsarContainerP
 		PropertyMapper map = PropertyMapper.get().alwaysApplyingWhenNonNull();
 		map.from(properties::getSchemaType).to(containerProperties::setSchemaType);
 		map.from(properties::getConcurrency).to(containerProperties::setConcurrency);
+		customizeListenerStartupProperties(containerProperties);
+	}
+
+	private void customizeListenerStartupProperties(ReactivePulsarContainerProperties<?> containerProperties) {
```
snicoll, line 104: Here as well.

snicoll (changes requested):
Thanks for the PR. I've added some nit and I would have processed them myself but I'd like some confirmation about the default values and, potentially, a code change in Spring Pulsar to avoid the duplication.

Also, can you please build locally before pushing?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/pulsar/PulsarPropertiesMapper.java
onobc: The code to de-duplicate was more complicated than having this in multiple places.

onobc:
Thanks for the review @snicoll. I think we can close this in favor of https://github.com/spring-projects/spring-boot/pull/42182 . I apologize for not marking this as such or closing it myself before you spent time reviewing (my bad).

onobc:
> Also, can you please build locally before pushing?

Yes, I saw the red builds. I was building `./gradlew build -x check` to avoid testing everything. I now know that `-x check` does not run format or checkstyle checks (which makes sense because they are "checks"). I was just trying to avoid running all tests in all modules.

This `./gradlew build -x test -x intTest` seems to work.

onobc:
Closing in favor of https://github.com/spring-projects/spring-boot/pull/42182

PR #42182, opened 2024-09-08
Add Pulsar container factory customizers

onobc:
This commit adds the ability for users to customize the auto-configured Spring for Apache Pulsar message container factories. Each container factory holds a set of container properties that is a common target for users to configure. Allowing the customization of these properties prevents a rapid increase of configuration properties for secondary priority properties.

> [!NOTE]
> I originally submitted a [PR to add config props for the startup failure options and timeout](https://github.com/spring-projects/spring-boot/pull/42180). However, in most cases, that alone is not sufficient and users will want to create their own retry template custom to their environment/scenario. This lead for the need to either allow users to configure in retry templates for each container (which would require `@Qualifer` and got messy quick. By allowing customization of the container factory we allow ourselves to get out of the "I need to add another config property" business for properties that are not primary in nature. Configuring the startup failure policy is probably not a primary operation. If we agree w/ this then the other PR can be closed in favor of this proposal.


snicoll:
Thanks for the PR. Rather than 3 customizer interfaces, can you please look if we could replace that with a generic? See `org.springframework.boot.autoconfigure.cache.CacheManagerCustomizers` for an example of how that can be used.

onobc:
Closing in favor of https://github.com/spring-projects/spring-boot/pull/42180

onobc:
@snicoll thanks for pointing out the `CacheManagerCustomizers`. I have done the same in the latest commit. The cache manager would have an explosion of interfaces had you not introduced the customizers concept then. Thanks for that - it also cleaned up these 3 into 1.

spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/messaging/pulsar.adoc
```diff
@@ -150,11 +150,11 @@ include-code::MyBean[]
 Spring Boot auto-configuration provides all the components necessary for `PulsarListener`, such as the `PulsarListenerContainerFactory` and the consumer factory it uses to construct the underlying Pulsar consumers.
 You can configure these components by specifying any of the `spring.pulsar.listener.\*` and `spring.pulsar.consumer.*` prefixed application properties.
 
-If you need more control over the consumer factory configuration, consider registering one or more `ConsumerBuilderCustomizer` beans.
+If you need more control over the configuration of the consumer factory used by the container factory to create consumers, consider registering one or more `ConsumerBuilderCustomizer` beans.
 These customizers are applied to all consumers created by the factory, and therefore all `@PulsarListener` instances.
 You can also customize a single listener by setting the `consumerCustomizer` attribute of the `@PulsarListener` annotation.
 
-
+If you need more control over the actual container factory configuration, consider registering one or more `PulsarContainerFactoryCustomizer<ConcurrentPulsarListenerContainerFactoryCustomizer<?>>` beans.
```
onobc, line 157: I chose to describe how to customize only the 3 concrete listener container factories.
I left out the fact that they can actually customize anywhere up/down the `PulsarContainerFactory` hierarchy as the listener hierarchy is a bit noisy and I think it may confuse things.

spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/messaging/pulsar.adoc
```diff
@@ -165,13 +165,13 @@ The following component creates a reactive listener endpoint on the `someTopic`
 include-code::MyBean[]
 
 Spring Boot auto-configuration provides all the components necessary for `ReactivePulsarListener`, such as the `ReactivePulsarListenerContainerFactory` and the consumer factory it uses to construct the underlying reactive Pulsar consumers.
-You can configure these components by specifying any of the `spring.pulsar.listener.*` and `spring.pulsar.consumer.*` prefixed application properties.
+You can configure these components by specifying any of the `spring.pulsar.listener.\*` and `spring.pulsar.consumer.*` prefixed application properties.
```
snicoll, line 168: This is an unrelated change that can easily be missed and the fix is in a different branch than this PR. Please don't do this.

I've fixed that in https://github.com/spring-projects/spring-boot/issues/42200

spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/messaging/pulsar.adoc
snicoll: I don't have an opinion, except the mouthful name there is unfortunate (there's an extra customizer in the generic that I've removed).

PR #42186, opened 2024-09-09
Polish

PiyalAhmed:
Polish `spring-javadoc.css`.

**Changes made:**

- Used shorthand properties where suitable
- Removed redundant unit -px 
- Inserted white-spaces after the  colon for better readability


wilkinsona:
Thanks for the PR. Reviewing it made me realise that the file you've polished is no longer used so I've deleted it in https://github.com/spring-projects/spring-boot/issues/42187.

PR #42192, opened 2024-09-09
Polish

arefbehboudi:
- Removed @SuppressWarnings as it was no longer needed.
- Used pattern matching.
- Updated documentation.
- Replaced class with record.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
@@ -698,7 +698,6 @@ public final class DataSourceBuilder<T extends DataSource> {
 	 */
 	private static class SimpleDataSourceProperties extends MappedDataSourceProperties<SimpleDriverDataSource> {
 
-		@SuppressWarnings("unchecked")
 		SimpleDataSourceProperties() {
 			add(DataSourceProperty.URL, SimpleDriverDataSource::getUrl, SimpleDriverDataSource::setUrl);
 			add(DataSourceProperty.DRIVER_CLASS_NAME, Class.class, (dataSource) -> dataSource.getDriver().getClass(),
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
@@ -698,6 +698,7 @@ public final class DataSourceBuilder<T extends DataSource> {
 	 */
 	private static class SimpleDataSourceProperties extends MappedDataSourceProperties<SimpleDriverDataSource> {
 
+		@SuppressWarnings("unchecked")
 		SimpleDataSourceProperties() {
 			add(DataSourceProperty.URL, SimpleDriverDataSource::getUrl, SimpleDriverDataSource::setUrl);
 			add(DataSourceProperty.DRIVER_CLASS_NAME, Class.class, (dataSource) -> dataSource.getDriver().getClass(),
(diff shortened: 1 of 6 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/rsocket/context/RSocketPortInfoApplicationContextInitializer.java
```diff
@@ -51,49 +51,43 @@ public void initialize(ConfigurableApplicationContext applicationContext) {
 		applicationContext.addApplicationListener(new Listener(applicationContext));
 	}
 
-	private static class Listener implements ApplicationListener<RSocketServerInitializedEvent> {
+	private record Listener(ConfigurableApplicationContext applicationContext) implements ApplicationListener<RSocketServerInitializedEvent> {
```
wilkinsona, line 54: Making this a `record` adds an accessor for the application context which we don't want. Please revert.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/json/JsonValueWriter.java
```diff
@@ -106,8 +106,8 @@ else if (value instanceof Map<?, ?> map) {
 		else if (value instanceof Number) {
 			append(value.toString());
 		}
-		else if (value instanceof Boolean) {
-			append(Boolean.TRUE.equals(value) ? "true" : "false");
+		else if (value instanceof Boolean bool) {
+			append(bool ? "true" : "false");
```
wilkinsona, line 110: I think it would be better to use `toString()` here. It could then be combined with the handling of `Number`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
```diff
@@ -698,7 +698,6 @@ private void setDriverClass(ComboPooledDataSource dataSource, String driverClass
 	 */
 	private static class SimpleDataSourceProperties extends MappedDataSourceProperties<SimpleDriverDataSource> {
 
-		@SuppressWarnings("unchecked")
```
wilkinsona, line 701: Removing this results in a warnings with `javac`:

```
 warning: [unchecked] unchecked method invocation: method add in class MappedDataSourceProperties is applied to given types
                        add(DataSourceProperty.DRIVER_CLASS_NAME, Class.class, (dataSource) -> dataSource.getDriver().getClass(),
                           ^
  required: DataSourceProperty,Class<V>,Getter<T,V>,Setter<T,V>
  found:    DataSourceProperty,Class<Class>,Getter<SimpleDriverDataSource,Class>,Setter<SimpleDriverDataSource,Class>
  where V,T are type-variables:
    V extends Object declared in method <V>add(DataSourceProperty,Class<V>,Getter<T,V>,Setter<T,V>)
    T extends DataSource declared in class MappedDataSourceProperties
error: warnings found and -Werror specified
1 error
1 warning
```

Please revert.

wilkinsona (changes requested):

arefbehboudi:
Thanks for the review. I've updated the code based on your feedback. Please check it again when you can.

snicoll:
@arefbehboudi thanks for the contribution. It wasn't building as the formatting was no longer correct. FTR you could have seen that locally by building the module, something like:

```
./gradlew :spring-boot-project:spring-boot:check
```

I've polished things in https://github.com/spring-projects/spring-boot/commit/ece5c6fe9e613831d2f0b2dd67ee4f7aa3828c76

arefbehboudi:
@snicoll Thanks for pointing that out! I'll make sure to run the build locally using the provided command to catch any formatting issues in the future. Thanks again for the guidance!

PR #42268, opened 2024-09-11
Polish

arefbehboudi:
- Removed @SuppressWarnings as it was no longer needed.
- Used pattern matching.
- Replace the while loop with string repeat


mhalbritter:
Thanks!

arefbehboudi:
thank you @mhalbritter , Can I ask why my pull request build is getting errors?

mhalbritter:
On the CI, you mean? There's something wrong in the infrastructure I think.

arefbehboudi:
Thank you, so the issue is not with my pull request.

PR #42277, opened 2024-09-12
Add tests to ensure private constructor is not used for value object binding

quaff:
Sometimes we want to duplicate `@ConfigurationProperties` bean with different prefix, we should use the primary configuration properties as default of additional configuration properties instead of re-configuring them all, then we could inject that primary bean and copy its values as default via constructor, there are two ways to force constructor used for injection instead of binding, annotating it with `@Autowired` or marking it as `private`, and the former one is already covered by tests, this commit add test for latter one.

### DIFF
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/bind/DefaultBindConstructorProviderTests.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/bind/DefaultBindConstructorProviderTests.java
@@ -33,6 +33,7 @@ import static org.assertj.core.api.Assertions.assertThatIllegalStateException;
  *
  * @author Phillip Webb
  * @author Madhura Bhave
+ * @author Yanming Zhou
  */
 class DefaultBindConstructorProviderTests {
 
@@ -90,6 +91,12 @@ class DefaultBindConstructorProviderTests {
 		assertThatIllegalStateException()
 			.isThrownBy(() -> this.provider.getBindConstructor(TwoConstructorsWithBothConstructorBinding.class, false))
 			.withMessageContaining("has more than one @ConstructorBinding");
+	}
+
+	@Test
+	void getBindConstructorWhenIsTypeWithPrivateConstructorReturnsNull() {
+		Constructor<?> constructor = this.provider.getBindConstructor(TypeWithPrivateConstructor.class, false);
+		assertThat(constructor).isNull();
 	}
 
 	@Test
@@ -220,6 +227,13 @@ class DefaultBindConstructorProviderTests {
 
 		@ConstructorBinding
 		TwoConstructorsWithBothConstructorBinding(String name, int age) {
+		}
+
+	}
+
+	static final class TypeWithPrivateConstructor {
+
+		private TypeWithPrivateConstructor(Environment environment) {
 		}
 
 	}


spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java
```diff
@@ -1270,6 +1271,25 @@ void loadWhenBindingToJavaBeanWithConversionToCustomListImplementation() {
...
+	}
+
+	@Test
+	void privateConstructorShouldInjectBeanInsteadOfBindProperties() {
+		load(ServicePropertiesConfiguration.class, "primary.service.port=1234", "bar.service.host=127.0.0.1");
+		BarServiceProperties barServiceProperties = this.context.getBean(BarServiceProperties.class);
+		assertThat(barServiceProperties.getHost()).isEqualTo("127.0.0.1");
+		assertThat(barServiceProperties.getPort()).isEqualTo(1234);
+	}
+
```
wilkinsona, line 1292: This is similar to this test:

https://github.com/spring-projects/spring-boot/blob/2bfd784d7baef889fab89ab406f9fcb762b81c6d/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/bind/DefaultBindConstructorProviderTests.java#L95-L100

If we want to cover the non-member case as well, I think it should be done in `DefaultBindConstructorProviderTests` rather than here as it's `DefaultBindConstructorProvider` that contains the logic for ignoring private constructors:

https://github.com/spring-projects/spring-boot/blob/2bfd784d7baef889fab89ab406f9fcb762b81c6d/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/DefaultBindConstructorProvider.java#L188

wilkinsona (changes requested):

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java
```diff
@@ -1270,6 +1271,25 @@ void loadWhenBindingToJavaBeanWithConversionToCustomListImplementation() {
...
+	@Test
+	void autowiredConstructorShouldInjectBeanInsteadOfBindProperties() {
+		load(ServicePropertiesConfiguration.class, "primary.service.port=1234", "foo.service.host=127.0.0.1");
+		PrimaryServiceProperties primaryServiceProperties = this.context.getBean(PrimaryServiceProperties.class);
+		assertThat(primaryServiceProperties.getHost()).isEqualTo("localhost");
+		assertThat(primaryServiceProperties.getPort()).isEqualTo(1234);
+		FooServiceProperties fooServiceProperties = this.context.getBean(FooServiceProperties.class);
+		assertThat(fooServiceProperties.getHost()).isEqualTo("127.0.0.1");
+		assertThat(fooServiceProperties.getPort()).isEqualTo(1234);
+	}
```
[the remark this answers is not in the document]
quaff: I suggest keeping them as integration tests, it demonstrates the use case, WDYT?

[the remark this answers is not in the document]
wilkinsona: I would prefer not to have the duplication.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/context/properties/ConfigurationPropertiesTests.java
quaff: I’ll add unit test tomorrow.

quaff: done.

philwebb:
Thanks @quaff!

Issue #41668, opened 2024-08-01
Support @Scope @ConfigurationProperties beans


sfilipiak-inpost:
When using `@ConfigurationProperties` via either `@ConfigurationPropertiesScan` or `@EnableConfigurationProperties`, bean scope specified via `@Scope` is not respected.
On the other hand, when defining `@ConfigurationProperties` with additional `@Configuration`/`@Component`/`@Bean` annotation, specified scope is respected

Provided below is a sample application using spring boot 3.3.2

```java
package configbug;

import org.springframework.beans.factory.config.ConfigurableBeanFactory;
import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.boot.context.event.ApplicationReadyEvent;
import org.springframework.boot.context.properties.ConfigurationProperties;
import org.springframework.boot.context.properties.EnableConfigurationProperties;
import org.springframework.context.annotation.AnnotationConfigApplicationContext;
import org.springframework.context.annotation.Bean;
import org.springframework.context.annotation.Configuration;
import org.springframework.context.annotation.Scope;
import org.springframework.context.event.EventListener;
import org.springframework.stereotype.Component;

@SpringBootApplication
@EnableConfigurationProperties(ConfigBugApplication.MyProperties.class)
public class ConfigBugApplication {

    public static void main(String[] args) {
        SpringApplication.run(ConfigBugApplication.class, args);
    }

    @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
    @ConfigurationProperties("my.properties")
    public static class MyProperties {
        private String value;
    }

    @Component
    @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
    @ConfigurationProperties("my.other.properties")
    public static class MyOtherProperties {
        private String value;
    }

    @Configuration
    @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
    @ConfigurationProperties("my.yet.another.properties")
    public static class MyYetAnotherProperties {
        private String value;
    }

    @ConfigurationProperties("my.yet.yet.another.properties")
    public static class MyYetYetAnotherProperties {
        private String value;
    }

    @Bean
    @Scope(ConfigurableBeanFactory.SCOPE_PROTOTYPE)
    MyYetYetAnotherProperties myYetYetAnotherProperties() {
        return new MyYetYetAnotherProperties();
    }

    @EventListener
[...]

PR #42289, opened 2024-09-13
Add common definition annotations support for ConfigurationProperties beans

quaff:
Such as `@Lazy`, `@Primary` and `@Fallback` etc.


philwebb:
This looks quite similar to #41668 and seems sensible to so alongside it in 3.4

philwebb:
Thanks once again @quaff!

PR #42313, opened 2024-09-14
management.health.db.ignore-routing-datasources=true has no effect when an AbstractRoutingDataSource has been wrapped

nosan:
Currently, the logic inside  `DataSourceHealthContributorAutoConfiguration` auto-configuration is based solely on `instanceof` and does not account for situations where someone has added a proxy/wrapper around `AbstractRoutingDataSource` through `BeanPostProcessor` or other ways.


snicoll:
Thanks again @nosan!

Issue #42307, opened 2024-09-13
Add support for partitioned cookies


miskr-instructure:
Since the fix of https://github.com/spring-projects/spring-session/issues/2787 the `DefaultCookieSerializer` of spring session now supports the partitioned attribute. However **that attribute cannot be set via configuration**, only by providing a custom 
`CookieSerializer` Bean.

This application.yml does not work as one would intuitively expect:
```yaml
server:
  servlet:
    session:
      cookie:
        same-site: 'none'
        partitioned: 'true'  # <-- no effect
        secure: 'true'       # <-- also no effect?
```
... it seems it's only possible to set the new value by providing a custom `@Bean`:
```java
@Configuration
class CustomCookieSerializerConfig {
  @Bean
  CookieSerializer cookieSerializer() {
    var cookieSerializer = new DefaultCookieSerializer();
    cookieSerializer.setSameSite("None");
    cookieSerializer.setPartitioned(true);
    cookieSerializer.setUseSecureCookie(true);
    return cookieSerializer;
  }
}
```

The likely cause is missing implementation in [org.springframework.session.config.annotation.web.http.SpringHttpSessionConfiguration.createDefaultCookieSerializer()](https://github.com/spring-projects/spring-session/blob/3.4.0-M2/spring-session-core/src/main/java/org/springframework/session/config/annotation/web/http/SpringHttpSessionConfiguration.java#L180)

Regarding affected versions:
* "server.servlet.session.cookie.secure" doesn't work in spring boot 3.3.x (and probably never has based on git blame)
* the partitioned feature will be a new one in spring boot 3.4.x which is not released yet (but it'd be nice if it worked by the time it goes GA)

philwebb:
We don't currently have support for `partitioned`, but I'm surprised to see `secure` not working. Are you sure that's the case @miskr-instructure?

philwebb:
Closing in favor of PR #42316. Thanks @nosan!

PR #42316, opened 2024-09-15
Add support for partitioned cookies

nosan:
gh-42307


philwebb:
Thanks @nosan. I think we should also update `AbstractServletWebServerFactory.configureSessionCookie` and `WebSessionIdResolverAutoConfiguration.initializeCookie` as well to support the new property. If you have time to update your PR, that would be great. Otherwise we can take it on when we merge.

nosan:
Thanks for the feedback @philwebb.

I did update my PR but unfortunately, I have not found a strict way to update `AbstractServletWebServerFactory.configureSessionCookie` because `jakarta.servlet.SessionCookieConfig` does not have a method to set the `partitioned` property. 

The only way as I understand is:
```java
map.from(cookie::getPartitioned).whenTrue().toCall(() -> config.setAttribute(PARTITIONED_ATTRIBUTE, ""));
```

https://jakarta.ee/specifications/servlet/6.1/apidocs/jakarta.servlet/jakarta/servlet/http/cookie

bclozel:
I think this is the only way to support partitioned cookies, see Tomcat for reference:

https://github.com/apache/tomcat/blob/d6e5a088433af43a35293f9a2b19cff359efb79b/java/org/apache/catalina/core/ApplicationSessionCookieConfig.java#L224

nosan:
@bclozel 
However, this value could be overridden later through session cookie attributes.  

https://github.com/apache/tomcat/blob/d6e5a088433af43a35293f9a2b19cff359efb79b/java/org/apache/catalina/core/ApplicationSessionCookieConfig.java#L229

bclozel:
@nosan  you can raise this as a Tomcat issue if you think this is a bug.

nosan:
@bclozel 
For me, it does not sound like a bug.
As I understood you can set a value for partitioned through `Context.setUsePartitioned` and then if you specify
`sessionCookieConfig.setAttribute(PARTITIONED_ATTRIBUTE_NAME, Boolean.toString(partitioned)` this value will be overridden.

>   // NOTE: The priority order for session cookie configuration is:
        // 1. Context level configuration
        // 2. Values from SessionCookieConfig
        // 3. Defaults
 
 so, this line which I added in this PR should work:
 ```java
 	map.from(cookie::getPartitioned)
				.to(partitioned -> config.setAttribute(PARTITIONED_ATTRIBUTE_NAME, Boolean.toString(partitioned)));
```

bclozel:
That was as well my assessment in the first place.

mhalbritter:
Thanks @nosan !

PR #42343, opened 2024-09-18
Add statsd graceful shutdown support

popotec:
I have encountered an issue in my Spring Batch job where, intermittently, the error "Singleton bean creation not allowed while the singletons of this factory are in destruction" occurs with the StatsdMeterRegistry bean.

This issue seems to arise when the application unexpectedly receives a SIGTERM signal (such as when a node goes down in a cloud environment) and begins the shutdown process, some background tasks in the StatsdMeterRegistry are not properly terminated before the bean destruction phase. As a result, there appears to be an attempt to access the StatsdMeterRegistry after the ApplicationContext has already started to shut down.

To address this, I believe it would be beneficial to gracefully shut down the StatsdMeterRegistry before the bean is destroyed. A possible solution could be to modify the StatsdMetricsExportAutoConfiguration to ensure that StatsdMeterRegistry.close() is called during the shutdown phase, as shown below:

```java
    @PreDestroy
    public void destroy(StatsdMeterRegistry statsdMeterRegistry) {
        statsdMeterRegistry.close();
    }
```

This would help to properly terminate any ongoing background tasks and prevent the issue from occurring during application shutdown.

Thank you for considering this PR. I look forward to your feedback!

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdMetricsExportAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdMetricsExportAutoConfiguration.java
@@ -19,6 +19,7 @@ package org.springframework.boot.actuate.autoconfigure.metrics.export.statsd;
 import io.micrometer.core.instrument.Clock;
 import io.micrometer.statsd.StatsdConfig;
 import io.micrometer.statsd.StatsdMeterRegistry;
+import jakarta.annotation.PreDestroy;
 
 import org.springframework.boot.actuate.autoconfigure.metrics.CompositeMeterRegistryAutoConfiguration;
 import org.springframework.boot.actuate.autoconfigure.metrics.MetricsAutoConfiguration;
@@ -59,4 +60,8 @@ public class StatsdMetricsExportAutoConfiguration {
 		return new StatsdMeterRegistry(statsdConfig, clock);
 	}
 
+	@PreDestroy
+	public void destroy(StatsdMeterRegistry statsdMeterRegistry) {
+		statsdMeterRegistry.close();
+	}
 }


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/statsd/StatsdMetricsExportAutoConfiguration.java
```diff
@@ -59,4 +60,8 @@ public StatsdMeterRegistry statsdMeterRegistry(StatsdConfig statsdConfig, Clock
 		return new StatsdMeterRegistry(statsdConfig, clock);
 	}
 
+	@PreDestroy
+	public void destroy(StatsdMeterRegistry statsdMeterRegistry) {
+		statsdMeterRegistry.close();
+	}
 }
```
mdeinum, line 67: This won't work as `@PrreDestroy` should not have any arguments (See https://jakarta.ee/specifications/platform/9/apidocs/jakarta/annotation/predestroy). 

Finally as the `StatsdMeterRegistry` has a [`close`](https://www.javadoc.io/doc/io.micrometer/micrometer-registry-statsd/latest/io/micrometer/statsd/StatsdMeterRegistry.html#close()) method that should already be registered automatically as a pre-destroy method. For this see the [`@Bean`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/annotation/Bean.html#destroyMethod()). 

So the solution to the problem lies not in adding additional annotations but probably elsewhere.

wilkinsona: Thanks, @mdeinum. I was just going to comment with the same analysis.

@popotec can you please open an issue so that we can investigate further? Seeing the stack trace of the exception when bean creation is attempted during shutdown may be enough for us to figure out the cause.

wilkinsona:
Thanks for the PR, @popotec, but as [discussed](https://github.com/spring-projects/spring-boot/pull/42343#discussion_r1764455153) I don't think this is the right solution to your problem. Please do open an issue if you'd like us to investigate further.

PR #42402, opened 2024-09-20
Code cleanup

mjimani:


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ansi/AnsiOutput.java
```diff
@@ -169,6 +166,9 @@ private static boolean detectIfAnsiCapable() {
 					}
 				}
 			}
+			if (!consoleAvailable) {
+				return false;
+			}
```
wilkinsona, line 171: I don't think this is any cleaner. Please revert.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/availability/ApplicationAvailabilityBean.java
```diff
@@ -84,7 +84,7 @@ private <S extends AvailabilityState> Object getLogMessage(Class<S> type, Availa
 		StringBuilder message = new StringBuilder(
 				"Application availability state " + type.getSimpleName() + " changed");
 		message.append((lastChangeEvent != null) ? " from " + lastChangeEvent.getState() : "");
-		message.append(" to " + event.getState());
+		message.append(" to ").append(event.getState());
```
philwebb, line 87: I'm afraid I don't think we should do this one either. There's not going to be much of a performance impact.

philwebb:
Thanks for the suggestions, but given the review comments I don't think we should apply these changes.

Issue #42339, opened 2024-09-17
Improve SpringEnvironmentLookup exception to hint that you might need a file ending '-spring'


philwebb:
See #42336. We could improve the message or throw a specific exception and have a failure analyzer. 

```
2024-09-17T10:29:21.475587700Z main ERROR Resolver failed to lookup spring:spring.application.name java.lang.IllegalStateException: Unable to obtain Spring Environment from LoggerContext
	at org.springframework.util.Assert.state(Assert.java:76)
	at org.springframework.boot.logging.log4j2.SpringEnvironmentLookup.lookup(SpringEnvironmentLookup.java:46)
	at org.apache.logging.log4j.core.lookup.StrLookup.evaluate(StrLookup.java:108)
	at org.apache.logging.log4j.core.lookup.Interpolator.evaluate(Interpolator.java:197)
	at org.apache.logging.log4j.core.lookup.StrSubstitutor.resolveVariable(StrSubstitutor.java:1227)
```

nosan:
@philwebb 
The exception message could be something like this
```
	public String lookup(String key) {
		Assert.state(this.environment != null, """
				Unable to obtain Spring Environment from LoggerContext.
				Usually, it happens when log4j2.xml file is being used instead of log4j2-spring.xml
				""");
		return this.environment.getProperty(key);
	}
```

What do you think?

philwebb:
@nosan I wonder if we should make it clear that other extensions might also be used. The person reporting the issue was using yaml I think!

Perhaps something like:

Unable to obtain Spring Environment from LoggerContext. This can happen if your logback configuration filename does not end '-spring' (for example using 'log4j2.xml' instead of 'log4j2-spring.xml')

We should probably also keep the message on one line I think.

nosan:
```java
public String lookup(String key) {
		Assert.state(this.environment != null,
				"Unable to obtain Spring Environment from LoggerContext. "
						+ "This can happen if your log4j2 configuration filename does not end with '-spring' "
						+ "(for example using 'log4j2.xml' instead of 'log4j2-spring.xml')");
		return this.environment.getProperty(key);
	}
	
```
The error message will be	

> Unable to obtain Spring Environment from LoggerContext. This can happen if your log4j2 configuration filename does not end with '-spring' (for example using 'log4j2.xml' instead of 'log4j2-spring.xml')

philwebb:
> If it is ok, I can create PR 😄

ABSOLUTELY!! 👍

[...]

PR #42405, opened 2024-09-20
Improve SpringEnvironmentLookup exception to hint that the log4j2 file must have '-spring' suffix

nosan:
gh-42339


philwebb:
Nice one! Thanks @nosan!

Issue #40904, opened 2024-05-25
Prometheus Exemplars are missing from _count


jonatan-ivanov:
It seems that 3.3.0 introduced a regression during the migration of Prometheus 1.x: exemplars are missing from `_count` time series.

I think there are two bugs that are causing this:
1. It seems `PrometheusPropertiesConfigAdapter` does not respect the Micrometer defaults for "Prometheus Properties" ([these](https://prometheus.github.io/client_java/config/config/#exporter-properties)) when there are no user-defined custom properties set: https://github.com/spring-projects/spring-boot/blob/82d8222f56361079daae64eec4056fe4d7ce6bc9/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/prometheus/PrometheusPropertiesConfigAdapter.java#L65-L74
It seems Micrometer defaults (`PrometheusConfig.super.prometheusProperties()`) are only respected and merged with user-defined custom properties when they present, otherwise Micrometer defaults are ignored. This is a problem since Micrometer [sets some defaults](https://github.com/micrometer-metrics/micrometer/blob/23b6c43d9ce7a758dd5aa4620c776358f3a86039/implementations/micrometer-registry-prometheus/src/main/java/io/micrometer/prometheusmetrics/PrometheusConfig.java#L72-L76) which can be ignored or not ignored depending on the presence of user-defined properties.
2. Ignoring Micrometer defaults could have been worked around by manually setting them from user-properties (or by setting any custom Prometheus property) but another bug prevents them to take effect on the exporter. It seems `PrometheusOutputFormat` does not use these properties when it initializes the exporters: https://github.com/spring-projects/spring-boot/blob/82d8222f56361079daae64eec4056fe4d7ce6bc9/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/metrics/export/prometheus/PrometheusOutputFormat.java#L81
[...]

PR #42406, opened 2024-09-21
Add Javadoc since for PrometheusScrapeEndpoint(PrometheusRegistry, Properties)

izeye:
This PR adds a Javadoc `@since` tag for the `PrometheusScrapeEndpoint(PrometheusRegistry, Properties)`.

See gh-40904


philwebb:
Good spot! Thanks yet again @izeye!!

Issue #42175, opened 2024-09-06
Make OtlpMeterRegistry virtual thread aware


mhalbritter:
It's possible to pass a `ThreadFactory` into the `OtlpMeterRegistry` constructor. At the moment we don't do this, which then uses a platform thread based factory.

We could enable virtual threads by passing factory which creates virtual threads.

jonatan-ivanov:
I think I would ask the question why would one want to use virtual threads for metrics publishing (registry).

Publishing is happening on a single blocking thread (there is another one that does some non-blocking work). They do their job very infrequently (once per minute by default) so I think the registry would not really get any benefits from using virtual threads. But we might(?) get a (probably negligible) drawback: the two infrequent virtual threads would be mounted to the same platform thread pool where other (business logic) virtual threads are, kind of "competing" with them so the business logic might be ever so slightly more disrupted in that case (I'm not sure this is measurable though).

bclozel:
I think that one of the expected characteristics of a virtual threads setups is memory consumption: the app only needs platform threads and does not consume more memory with additional threads pools.

As for the drawback you've highlighted I think it's quite positive actually. The virtual threads scheduler is very efficient and competing there should not be a problem as long as work is mostly I/O based.

So, 👍 from my side for this change.

mhalbritter:
Superseded by https://github.com/spring-projects/spring-boot/pull/42407.

PR #42407, opened 2024-09-21
Make OtlpMeterRegistry virtual thread aware

mturbe:
Add support for virtual threads in OtlpMeterRegistry configuration per issue #42175.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpProperties.java
@@ -79,6 +79,11 @@ public class OtlpProperties extends StepRegistryProperties {
 	 */
 	private TimeUnit baseTimeUnit = TimeUnit.MILLISECONDS;
 
+	/**
+	 * Whether virtual threads should be used for publishing metrics.
+	 */
+	private boolean virtualThreadsEnabled = false;
+
 	public String getUrl() {
 		return this.url;
 	}
@@ -144,6 +149,14 @@ public class OtlpProperties extends StepRegistryProperties {
 
 	public void setBaseTimeUnit(TimeUnit baseTimeUnit) {
 		this.baseTimeUnit = baseTimeUnit;
+	}
+
+	public boolean isVirtualThreadsEnabled() {
+		return this.virtualThreadsEnabled;
+	}
+
+	public void setVirtualThreadsEnabled(boolean virtualThreadsEnabled) {
+		this.virtualThreadsEnabled = virtualThreadsEnabled;
 	}
 
 }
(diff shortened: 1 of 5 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpProperties.java
```diff
@@ -79,6 +79,11 @@ public class OtlpProperties extends StepRegistryProperties {
 	 */
 	private TimeUnit baseTimeUnit = TimeUnit.MILLISECONDS;
 
+	/**
+	 * Whether virtual threads should be used for publishing metrics.
+	 */
+	private boolean virtualThreadsEnabled = false;
```
mhalbritter, line 85: We don't need a new property here. We have a condition, named `ConditionalOnThreading` which can be used to detect if the user wants to enable virtual threads. There are already usages in the codebase for this annotation which you can use to see how it's used.

mhalbritter (changes requested):
Hello @mturbe, thanks for the PR. I've added a comment for your consideration.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsExportAutoConfigurationTests.java
```diff
@@ -76,6 +79,35 @@ void allowsCustomConfigToBeUsed() {
...
+			assertThat(properties.isVirtualThreadsEnabled()).isFalse();
+			OtlpMeterRegistry registry = context.getBean(OtlpMeterRegistry.class);
+			assertThat(registry).extracting("scheduledExecutorService")
+				.satisfies((executor) -> ScheduledExecutorServiceAssert.assertThat((ScheduledExecutorService) executor)
+					.usesPlatformThreads());
+		});
+	}
+
+	@Test
+	void allowsVirtualThreadsToBeUsed() {
```
mhalbritter, line 96: We need to guard this test to only run on Java 21 and up.

spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/assertj/ScheduledExecutorServiceAssert.java
```diff
@@ -0,0 +1,84 @@
...
+		isNotNull();
+		if (!producesVirtualThreads()) {
+			failWithMessage("Expected executor to use virtual threads, but it uses platform threads");
+		}
+		return this;
+	}
+
+	private boolean producesVirtualThreads() {
+		try {
+			return this.actual.schedule(() -> Thread.currentThread().isVirtual(), 0, TimeUnit.SECONDS).get();
```
mhalbritter, line 68: We can't use this API, as our codebase is on Java 17, and `isVirtual` has been (stable) introduced with Java 21.

mhalbritter (changes requested):

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpProperties.java
mturbe: Great, thanks for sharing that. I'll switch to `ConditionalOnThreading`.

mhalbritter:

PR #42410, opened 2024-09-22
Add default value for spring.data.web.pageable.serialization-mode to configuration metadata

izeye:
See gh-39797

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -1126,6 +1126,10 @@
       }
     },
     {
+      "name": "spring.data.web.pageable.serialization-mode",
+      "defaultValue": "direct"
+    },
+    {
       "name" : "spring.datasource.continue-on-error",
       "type" : "java.lang.Boolean",
       "deprecation" : {


wilkinsona:
This isn’t needed:

https://docs.spring.io/spring-boot/3.4/appendix/application-properties/index.html#application-properties.data.spring.data.web.pageable.serialization-mode

The annotation processor now collects default values from enum fields.

izeye:
@wilkinsona Ah, sorry. I missed it.

So are the existing configuration metadata for default values not necessary, for example, `spring.jackson.constructor-detector` as follows?

https://github.com/spring-projects/spring-boot/blob/c7c73be8593b7d871e3a34dea8e360dc0d5b7ccd/spring-boot-project/spring-boot-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json#L1614-L1617

wilkinsona:
Many of the defaults were removed in https://github.com/spring-projects/spring-boot/commit/788417df7f62c79fc5ab5acdc18a5519a28ca212 as part of https://github.com/spring-projects/spring-boot/issues/7562. It's entirely possible that we missed some but that doesn't appear to be the case with `spring.jackson.constructor-detector` as the code has no default value:

https://github.com/spring-projects/spring-boot/blob/d4df6f75d13699f5c750fbc49fee98299bbf91f1/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jackson/JacksonProperties.java#L103-L107

PR #42411, opened 2024-09-22
remove duplicated file name access and use StringBuilder to insert in WebServerPortFileWriter

choi-hyeseong:
The File.getName() method always slice the path to get the name. so, it's not cached
It doesn't have a huge impact, but it's not efficient, so i fixed it :)

Also, the original code was truncating the extension from the filename, adding the namespace, and then putting the extension back together again, but i modified the code to be cleaner by using StringBuilder to do the insertion.

This change passed the test (WebServerPortFileWriterTests.java)

* Remove duplicated File.getName() method
* using StringBuilder to insert namespace


philwebb:
Thanks very much @choi-hyeseong, I wasn't aware that calls to `getName()` aren't cached. I'm not totally sure about using a `StringBuilder` here, to me it makes things a little harder to read. I've changed things back to using Strings, but tried to improve the variable names a little in 60f615849d02bf64bab605bff32ec65e05fcaa57.

choi-hyeseong:
Thank you for considering my PR :)

Issue #42401, opened 2024-09-20
Upgrade to Maven Javadoc Plugin 3.10.1


ZIRAKrezovic:
See discussion at https://issues.apache.org/jira/browse/MJAVADOC-812

Spring Boot Parent has updated maven-javadoc-plugin as part of spring-boot 3.4.0-M3 (#42233)

wilkinsona:
Thanks for bringing this to our attention. I've started watching MJAVADOC-812 and hopefully it will be fixed before our next 3.4 release which is RC1 on 24 October. We can keep this open to remind us to do something (either upgrade or rollback).

snicoll:
The PMC vote for `3.10.1` just passed so we'll be able to upgrade.

PR #42425, opened 2024-09-23
Downgraded Maven Javadoc Plugin to v3.8.0 from v3.10.0

moinb:
Downgraded Maven Javadoc Plugin to v3.8.0 from v3.10.0 as per 
- issue reported #42401
- MJAVADOC-812 https://issues.apache.org/jira/browse/MJAVADOC-812

Revert of commit https://github.com/spring-projects/spring-boot/commit/53554278bcb3ba43f11d9829e9277a711d4e01f6

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1264,7 +1264,7 @@ bom {
 			]
 		}
 	}
-	library("Maven Javadoc Plugin", "3.10.0") {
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, but we don't want to do this yet, if at all. Also, as described in the PR template, we don't accept contributions for one-line dependency version changes. Hopefully we'll be able to upgrade to 3.10.1 before Spring Boot 3.4.0-RC1 is released in a month's time. If we need to roll back to 3.8, we'll do that in October shortly before the RC1 release.

PR #42427, opened 2024-09-24
Polishing KotlinConfigurationPropertiesTests to use the same naming style

quaff:


wilkinsona:
I don't think we should make this change as it's the use of `@Import` that's being tested so the `…Importer` naming is appropriate. Changing it would also create a mismatch between the Kotlin tests and the Java tests, both of which were added in https://github.com/spring-projects/spring-boot/commit/090802b11b25a16694cc50f39431b16c1e54fc72#diff-16b057899dd46d8bae51b7064f58725836a605aa25c77660f0070e838b6ddbfb. Thanks anyway.

quaff:
> I don't think we should make this change as it's the use of `@Import` that's being tested so the `…Importer` naming is appropriate. Changing it would also create a mismatch between the Kotlin tests and the Java tests, both of which were added in [090802b](https://github.com/spring-projects/spring-boot/commit/090802b11b25a16694cc50f39431b16c1e54fc72). Thanks anyway.

But other tests in this test class use `Enable*`.

Issue #27897, opened 2021-09-09
Allow to configure order of @EnableTransactionManagement or provide a sensible value


quaff:
Currently order of @EnableTransactionManagement is fixed `Ordered.LOWEST_PRECEDENCE`, It's impossible to provide an `@Aspect` bean intercepting after transaction started, for example:
```java
@Before("execution(public * *(..)) and @annotation(transactional)")
public void registerTransactionSynchronization(JoinPoint jp, Transactional transactional) {
	if (!transactional.readOnly())
		TransactionSynchronizationManager.registerSynchronization(this); // always failed
}
``` 
I hope spring boot could provide option `spring.tx.order` just like `spring.aop.proxy-target-class`, It's a bit difficult.
The easy way is change order to `Ordered.LOWEST_PRECEDENCE+1`, but it could break compatibility potentially.

wilkinsona:
`@EnableTransactionManagement` doesn't support using an expression for its `order` attribute as it's typed as an `int`. This means that there's no way for us to parameterise it. For such a low-level requirement, I think you're better defining the configuration yourself. Spring Boot's `EnableTransactionManagementConfiguration` is small and should be copied into your own application and modified to meet your needs.

PR #42450, opened 2024-09-26
Configure sensible order for metrics aspects

quaff:
See https://github.com/spring-projects/spring-framework/issues/33595

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAspectsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsAspectsAutoConfiguration.java
@@ -33,12 +33,15 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.context.annotation.Bean;
 import org.springframework.context.annotation.Conditional;
+import org.springframework.core.Ordered;
+import org.springframework.core.annotation.Order;
 
 /**
  * {@link EnableAutoConfiguration Auto-configuration} for Micrometer-based metrics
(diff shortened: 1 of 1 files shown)


quaff:
Not sure tests should be added and how, I see other configuration doesn't test `@Order` on `@Bean` method.

wilkinsona:
Thanks for the proposal. I'm not too keen on adding magic numbers for the aspects' ordering, particularly as it working correctly relies upon Framework using a particular ordering. Let's see how https://github.com/spring-projects/spring-framework/issues/33595 plays out before doing anything in Boot.

wilkinsona:
On second reading of the Framework issue, I think this is really subjective: some will argue that the current ordering is sensible will others will want the opposite. I can see both sides as it depends on whether or not you see the transaction processing as part of the method invocation. Given this, I'm going to close this one as I don't think hard-coding a different order is the right answer. Depending on what happens on the Framework side, we may want to make a different change, for example one that allows the ordering to be configurable.

quaff:
> On second reading of the Framework issue, I think this is really subjective: some will argue that the current ordering is sensible will others will want the opposite. I can see both sides as it depends on whether or not you see the transaction processing as part of the method invocation. Given this, I'm going to close this one as I don't think hard-coding a different order is the right answer. Depending on what happens on the Framework side, we may want to make a different change, for example one that allows the ordering to be configurable.

It's hard-coding the default order, application always can register their own Bean to override it.
Allowing configure the order is great, and please reconsider https://github.com/spring-projects/spring-boot/issues/27897

wilkinsona:
> It's hard-coding the default order, application always can register their own Bean to override it.

As I said, the best ordering is subjective. Changing it may break some users so I think it's better to leave it as it is until we have a better way of users configuring their preferred order.

> Allowing configure the order is great, and please reconsider https://github.com/spring-projects/spring-boot/issues/27897

We can't reconsider #27897 as the `order` attribute on `@EnableTransactionManagement` is still an `int`.

There's no point doing anything more here until we know what direction to head in. Let's slow down a bit and see what comes of https://github.com/spring-projects/spring-framework/issues/33595.

Issue #40772, opened 2024-05-16
Auto-configure the Postgres application_name when using Docker Compose


sdavids:
Spring Boot 3.2.5

# Logs

```shell
$ docker compose logs db -f
...
db-1  | 2024-05-16 10:18:14.746 UTC [37] LOG:  connection received: host=192.168.65.1 port=42395
db-1  | 2024-05-16 10:18:14.766 UTC [37] LOG:  connection authenticated: identity="test" method=scram-sha-256 (/var/lib/postgresql/data/pg_hba.conf:128)
db-1  | 2024-05-16 10:18:14.766 UTC [37] LOG:  connection authorized: user=test database=test
db-1  | 2024-05-16 10:18:14.769 UTC [37] LOG:  execute <unnamed>: SET extra_float_digits = 3
db-1  | 2024-05-16 10:18:14.770 UTC [37] LOG:  execute <unnamed>: SET application_name = 'PostgreSQL JDBC Driver'
```

# Setup

## application.yaml

```yaml
spring:
  application:
    name: application-name-test
```

## compose.yaml

```yaml
services:
  db:
    image: postgres
    restart: always
    ports:
      - '5432:5432'
    environment:
      POSTGRES_USER: test
      POSTGRES_PASSWORD: test
      POSTGRES_DB: test
    command: ["postgres", "-c", "log_statement=all", "-c", "log_connections=true"]
```

---

Currently one has to explicitly configure the application name via `org.springframework.boot.jdbc.parameters`:

```yaml
services:
  db:
    image: postgres
...
    labels:
      org.springframework.boot.jdbc.parameters: 'ApplicationName=application-name-test'
```

⇓

```shell
db-1  | 2024-05-16 10:21:53.323 UTC [39] LOG:  execute <unnamed>: SET application_name = 'application-name-test'
```

wilkinsona:
Thanks for the suggestion. I can see how this is useful in production when, presumably, you're configuring the application name on the JDBC URL. What's the benefit at development time when using Docker Compose? Do you perhaps have multiple apps sharing a compose file and the name makes each app's queries easier to distinguish?

snicoll:
Closing in favor of PR, thanks again @nosan!

PR #42460, opened 2024-09-26
Auto-configure the Postgres application_name when using Docker Compose

nosan:
gh-40772

I added `since 3.4.0` everywhere but maybe it could be merged earlier if everything is ok.

Thanks in advance


nosan:
Should I also configure ApplicationName in `PostgresR2dbcDockerComposeConnectionDetailsFactory`?

nosan:
> Should I also configure ApplicationName in `PostgresR2dbcDockerComposeConnectionDetailsFactory`?

Added for r2dbc as well.

spring-boot-project/spring-boot-docker-compose/src/dockerTest/java/org/springframework/boot/docker/compose/service/connection/postgres/PostgresJdbcDockerComposeConnectionDetailsFactoryIntegrationTests.java
```diff
@@ -57,6 +57,38 @@ void runWithBitnamiImageCreatesConnectionDetails(JdbcConnectionDetails connectio
 		assertConnectionDetails(connectionDetails);
 	}
 
+	@DockerComposeTest(composeFile = "postgres-compose-application-name-label.yaml", image = TestImage.POSTGRESQL,
```
snicoll, line 60: Those tests shouldn't need to be full integration tests. The only thing that we assert here is that the `ConnectionDetails` auto-configured bean has the correct state. I can see that we don't have such tests so far but I wonder if it wouldn't be useful to put something in place so that we don't need all the docker machinery.

spring-boot-project/spring-boot-docker-compose/src/dockerTest/java/org/springframework/boot/docker/compose/service/connection/postgres/PostgresR2dbcDockerComposeConnectionDetailsFactoryIntegrationTests.java
```diff
@@ -60,6 +61,35 @@ void runWithBitnamiImageCreatesConnectionDetails(R2dbcConnectionDetails connecti
 		assertConnectionDetails(connectionDetails);
 	}
 
+	@DockerComposeTest(composeFile = "postgres-compose-application-name-label.yaml", image = TestImage.POSTGRESQL,
```
snicoll, line 64: Ditto. I can see this also changes the test infrastructure quite a bit and I wonder if that's warranted for this change.

spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/service/connection/postgres/PostgresJdbcDockerComposeConnectionDetailsFactory.java
```diff
@@ -78,6 +80,42 @@ public String getJdbcUrl() {
...
+				this.environment = environment;
+			}
+
+			@Override
+			protected String getParameters(RunningService service) {
+				String parameters = super.getParameters(service);
+				return appendApplicationName(parameters);
+			}
+
+			private String appendApplicationName(String parameters) {
```
snicoll, line 98: This is quite involved. I wonder if there wouldn't be a way to simplify this. And if not, perhaps `getParameters` may need to evolve so that such a handling of the raw string is unnecessary.

snicoll (commented):
Thanks for the PR @nosan. I feel that the testing bits of it are too involved and we should rather attempt to test the logic in simpler tests, if possible. Let us know if you have time to explore that route.

spring-boot-project/spring-boot-docker-compose/src/dockerTest/java/org/springframework/boot/docker/compose/service/connection/postgres/PostgresJdbcDockerComposeConnectionDetailsFactoryIntegrationTests.java
wilkinsona: +1. For this to warrant being a Docker-based test, I think it should somehow check that setting the `application_name` has had the desired effect on the Postgres side of things.

I'd then prefer that we only have a single Docker-based test that checks this and that the precedence of the different sources of the application name is unit tested without involving Docker. Right now we're adding a lot to the build time and I don't think the extra test coverage warrants that increase.

wilkinsona (changes requested):

nosan:
Thank you for your feedback @snicoll and @wilkinsona 

1. I can add unit tests for `PostgresJdbcDockerComposeConnectionDetails` and `PostgresDbR2dbcDockerComposeConnectionDetails` without involving docker. To verify that all options are set correctly.
2. I can add a docker integration test for `PostgresJdbcDockerComposeConnectionDetailsFactory` and `PostgresR2dbcDockerComposeConnectionDetailsFactory` to verify the application name is correctly set.

```java
	@DockerComposeTest(composeFile = "postgres-compose.yaml", image = TestImage.POSTGRESQL,
			properties = "spring.application.name=my-app")
	void runCreatesConnectionDetailsWithAutoConfiguredApplicationName(JdbcConnectionDetails connectionDetails)
			throws ClassNotFoundException {
		assertThat(connectionDetails.getUsername()).isEqualTo("myuser");
		assertThat(connectionDetails.getPassword()).isEqualTo("secret");
		assertThat(connectionDetails.getJdbcUrl()).startsWith("jdbc:postgresql://").endsWith("?ApplicationName=my-app");
		assertThat(getApplicationName(connectionDetails)).isEqualTo("my-app");
	}

	private String getApplicationName(JdbcConnectionDetails connectionDetails) throws ClassNotFoundException {
		JdbcTemplate template = getJdbcTemplate(connectionDetails);
		return template.queryForObject("select current_setting('application_name')", String.class);
	}
```

and

```java
@DockerComposeTest(composeFile = "postgres-compose.yaml", image = TestImage.POSTGRESQL,
			properties = "spring.application.name=my-app")
	void runCreatesConnectionDetailsWithAutoConfiguredApplicationName(R2dbcConnectionDetails connectionDetails) {
		assertConnectionDetails(connectionDetails);
		assertThat(getApplicationName(connectionDetails)).isEqualTo("my-app");
	}

	private String getApplicationName(R2dbcConnectionDetails connectionDetails) {
		ConnectionFactoryOptions connectionFactoryOptions = connectionDetails.getConnectionFactoryOptions();
		return DatabaseClient.create(ConnectionFactories.get(connectionFactoryOptions))
			.sql("select current_setting('application_name')")
			.map((row, metadata) -> row.get(0))
			.first()
			.map(Objects::toString)
			.block(Duration.ofSeconds(30));
	}
```

Do you think it sounds good to you?

spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/service/connection/postgres/PostgresJdbcDockerComposeConnectionDetailsFactory.java
nosan: I will take a look at how it could be done better and easy to read. I have the same feeling that it looks quite messy.

nosan:
Also, Should I revert my `DockerComposeExtension` changes?  I can test the application name via docker-compose labels.

nosan:
PR has been updated.

- I reverted my `DockerComposeTestExtension`, and `JdbcUrlBuilder` changes.
- I added unit tests for both `ConnectionDetails` and removed docker-compose integration tests.
- I left only two docker-compose integration tests for checking that `application_name` had the desired effect on the Postgres side.

spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/service/connection/postgres/PostgresJdbcUrl.java
```diff
@@ -0,0 +1,103 @@
...
+
+import org.springframework.util.LinkedMultiValueMap;
+import org.springframework.util.MultiValueMap;
+
+/**
+ * Utility class to customize Postgres JDBC URL.
+ *
+ * @author Dmytro Nosan
+ */
+class PostgresJdbcUrl {
```
wilkinsona, line 29: We discussed this today and it feels bit overly complicated to us. We think the same could be achieved with something similar to the code that [disables encryption for SQL Server](https://github.com/spring-projects/spring-boot/blob/69162f4f9fdf373c418ebb89165f19b48ef5c4cf/spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/service/connection/sqlserver/SqlServerJdbcDockerComposeConnectionDetailsFactory.java#L62-L72). One addition that's probably needed here is to encode the application name as it's user-provided input and may contain characters that are restricted in a query string.

wilkinsona (changes requested):

spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/service/connection/postgres/PostgresJdbcUrl.java
nosan: Fixed

PR #42462, opened 2024-09-27
Centralize slash character in static final variable, improves clarity and maintainability.

cmabdullah:


wilkinsona:
Thanks for the proposal but, IMO, this doesn't improve the clarify of the code and I'm not sure about the maintainability either.

Issue #42466, opened 2024-09-27
Update link to the official Micrometer @Timed documentation


gheyle:
The current link of the official Micrometer @Timed annotation documentation is at: <https://docs.micrometer.io/micrometer/reference/concepts/timers.html#_the_timed_annotation>, while the current Spring Boot documentation at <https://github.com/spring-projects/spring-boot/blob/b82830c59605e0bc8f39d0df14dcc8774d6abc5a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/actuator/metrics.adoc?plain=1#L1129> points to https://docs.micrometer.io/micrometer/reference/concepts#_the_timed_annotation.

jeanbisutti:
@gheyle It seems an issue for the https://github.com/spring-projects/spring-boot repository?

snicoll:
Closing in favor of PR #42467

PR #42467, opened 2024-09-27
Fix links to Micrometer reference doc

ngocnhan-tran1996:
Closes https://github.com/spring-projects/spring-boot/issues/42466


snicoll:
Thanks again, @ngocnhan-tran1996.

PR #42472, opened 2024-09-29
Allow common messages to be specified for message sources

mmoayyed:
This change extends the configuration schema to support the specification of common messages, when defining a `messageSource` bean. Common messages are [described here](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/context/support/AbstractMessageSource.html#setCommonMessages(java.util.Properties)):

> Specify locale-independent common messages, with the message code as key and the full message String (may contain argument placeholders) as value.

Still in the process of testing this, and figured it would be good to post this early to get feedback.

Thank you!


mmoayyed:
Hello all, I see this PR has been "partially" reviewed. In concept, does this pass muster, and can it be included in the coming RC? Is the idea sound here?

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/MessageSourceProperties.java
```diff
@@ -42,6 +42,11 @@ public class MessageSourceProperties {
 	 */
 	private String basename = "messages";
 
+	/**
+	 * Comma-separated list of locale-independent common messages.
+	 */
+	private String commonMessages;
```
philwebb, line 48: I think we might be better off using a `List<Resource>` here, that way the `Binder` can take care of creating the resource instances.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/MessageSourceAutoConfiguration.java
```diff
@@ -81,6 +89,25 @@ public MessageSource messageSource(MessageSourceProperties properties) {
 		}
 		messageSource.setAlwaysUseMessageFormat(properties.isAlwaysUseMessageFormat());
 		messageSource.setUseCodeAsDefaultMessage(properties.isUseCodeAsDefaultMessage());
+
+		try {
+			if (StringUtils.hasText(properties.getCommonMessages())) {
+				PropertiesFactoryBean propertiesFactory = new PropertiesFactoryBean();
```
philwebb, line 95: We might be better off using `PropertiesLoaderUtils.fillProperties` here rather than `PropertiesFactoryBean`. Using a `FactoryBean` isn't ideal since we need to call `afterPropertiesSet` etc.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/context/MessageSourceAutoConfiguration.java
```diff
@@ -81,6 +89,25 @@ public MessageSource messageSource(MessageSourceProperties properties) {
...
+				ResourceLoader resourceLoader = new DefaultResourceLoader();
+				String[] commonMessages = StringUtils.commaDelimitedListToStringArray(
+						StringUtils.trimAllWhitespace(properties.getCommonMessages()));
+				List<Resource> commonResources = Arrays.stream(commonMessages)
+						.map(resourceLoader::getResource)
+						.toList();
+				propertiesFactory.setLocations(commonResources.toArray(Resource[]::new));
+				propertiesFactory.setSingleton(true);
+				propertiesFactory.setIgnoreResourceNotFound(true);
```
philwebb, line 104: I think we should fail hard if the resource is not found.

philwebb (commented):
Thanks for the PR @mmoayyed. I've left some comments for your consideration.

mmoayyed:
Thank you very much @philwebb . Updated PR based on your comments and resolved conflicts with upstream. Please resume.

mmoayyed:
Side note: I think it would be good to disable the Gradle configuration cache for the project explicitly (in case someone has it globally enabled). otherwise things like this show up:

```
* What went wrong:
Could not load the value of field `b` of `com.gradle.scan.plugin.internal.service.d` bean found in field `pluginServiceFactory` of `org.gradle.internal.enterprise.impl.DefaultGradleEnterprisePluginAdapter` bean found in Gradle runtime.
> Could not load the value of field `onAccess` of `org.gradle.internal.configuration.inputs.AccessTrackingEnvMap` bean found in field `env` of `io.spring.develocity.conventions.gradle.DevelocityConventionsPlugin$2` bean found in field `capturedArgs` of `java.lang.invoke.SerializedLambda` bean found in field `capturedArgs` of `java.lang.invoke.SerializedLambda` bean found in field `a` of `com.gradle.develocity.agent.gradle.internal.scan.BuildScanConfigurationInternal$1` bean found in field `a` of `com.gradle.scan.plugin.internal.service.a.e$a` bean found in field `a` of `com.gradle.scan.plugin.internal.service.a.g` bean found in field `m` of `com.gradle.scan.plugin.internal.service.a.e` bean found in field `m` of `com.gradle.scan.plugin.internal.service.n` bean found in field `a` of `com.gradle.scan.plugin.internal.service.PluginServiceFactory` bean found in field `b` of `com.gradle.scan.plugin.internal.service.d` bean found in field `pluginServiceFactory` of `org.gradle.internal.enterprise.impl.DefaultGradleEnterprisePluginAdapter` bean found in Gradle runtime.
```


...which can be bypassed with an explicit `--no-configuration-cache` parameter but of course are less than pleasant.

philwebb:
Thanks very much @mmoayyed, this has been merged to `main`.

PR #42508, opened 2024-10-03
Use builder-jammy-java-tiny when using Buildpacks

anthonydahanne:
* with dual arch builder-jammy-java-tiny
* update documentation and trusted builders too


mhalbritter:
@anthonydahanne i think you meant @mhalbritter :)

mhalbritter:
I've tested it with x64 and aarch64 and it works. Great job, thanks @anthonydahanne !

Issue #42500, opened 2024-10-02
Add property to specify Docker Compose flags


mhalbritter:
https://github.com/spring-projects/spring-boot/issues/38763 allows specifying the arguments passed to the start and stop Docker Compose commands.

However, there are some commands which needs to be passed to `docker compose` directly after and before any subcommand like `up`. `--project-name` is one of them.

We should add a new property to allow that. Maybe `spring.docker.compose.arguments`?

nosan:
I've prototyped some changes in my [branch](https://github.com/spring-projects/spring-boot/compare/main...nosan:spring-boot:gh-42500). If it is something that you are looking for, let me know and I will create PR.

mudabirhussain:
> #38763 allows specifying the arguments passed to the start and stop Docker Compose commands.
> 
> However, there are some commands which needs to be passed to `docker compose` directly after and before any subcommand like `up`. `--project-name` is one of them.
> 
> We should add a new property to allow that. Maybe `spring.docker.compose.arguments`?

Yes, arguments should be passed via props, but I suggest a slightly different approach, like spring.docker.compose.arguments.file, spring.docker.compose.arguments.project-name, and spring.docker.compose.arguments.env-file. Alternatively, we can simplify it by omitting 'arguments' and use: spring.docker.compose.project-name, spring.docker.compose.env-file, and spring.docker.compose.file.

What do you suggest?

mhalbritter:
> I've prototyped some changes in my [branch](https://github.com/spring-projects/spring-boot/compare/main...nosan:spring-boot:gh-42500). If it is something that you are thinking about, let me know and I will create PR.

Hey @nosan , thanks for working on that. Please create a PR, i'll try to get it into 3.4.0-RC1. Thanks!

nosan:
Hey @mhalbritter,
[PR](https://github.com/spring-projects/spring-boot/pull/42571) has been created.
Thanks

mhalbritter:
Superseded in favor of #42571.

PR #42571, opened 2024-10-09
Add property to specify Docker Compose flags

nosan:
#42500


mhalbritter:
Thanks @nosan !

nosan:
Thanks @mhalbritter
That was super fast ❤️

Issue #42576, opened 2024-10-09
Add spring.redis.lettuce.read-from property for configuring LettuceClientConfiguration's read from setting


thiagoteichmann:
Currently using Lettuce but noticed that config properties lacks readFrom property and require code change to implement it, would be possible to externalize this in a property so we can override the default?

I currently have a redis sentinel totally configured by properties but I cannot configure read from property the same way.

Something like:
```
spring.redis.lettuce.config.read-from=REPLICA_PREFERRED
```
or 
```
spring.redis.config.read-from=REPLICA_PREFERRED
````

Originally opened a ticket against spring data redis but was requested to open a ticket here.

**Comment by thiagoteichmann** [comment]
Hi @nosan  @wilkinsona , is there any update on this?

**Comment by nosan** [comment]
Hi @thiagoteichmann,

The issue is still in progress. Before moving forward, @wilkinsona reached out to the Lettuce team to request support for more complex ReadFrom types, such as `regex` and `subnet`.

I've created a PR: https://github.com/redis/lettuce/pull/3016. Once it's merged, I believe it would be possible to support this in Spring Boot as well.

**Comment by nosan** [comment]
However, please keep in mind that I'm just a contributor, so it would be best to wait for @wilkinsona's feedback.

**Comment by thiagoteichmann** [comment]
Sure, no worries, thanks for the update! 👍🏻

**Comment by tishun** [comment]
@thiagoteichmann are you using snake_case because it is a requirement, or could you also work with kebab-case, e.g.

```properties
spring.redis.lettuce.config.read-from=replica-preferred
```

It would help us consider how we should implement the solution in the end greatly if we know more about your use case.

wilkinsona:
@tishun Boot's relaxed binding can take handle various different property formats so, personally, I wouldn't worry about handling multiple formats in Lettuce. I'd just pick one and stick with it and leave Boot to deal with that complexity.

nosan:
@wilkinsona 
 I thought that relaxed bindings only applied to property names, not values. Have I misunderstood or missed something here?

wilkinsona:
Boot can bind enums and similar in a relaxed form as well (see `LenientObjectToEnumConverterFactory` for example). Although we won't be targetting an enum, I think we should do something here and leave Boot to bear the burden of relaxed binding rather than Lettuce having to deal with it too.

thiagoteichmann:
[...]

PR #42588, opened 2024-10-10
Add spring.redis.lettuce.read-from property

nosan:
#42576

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/redis/LettuceConnectionConfiguration.java
@@ -17,8 +17,10 @@
 package org.springframework.boot.autoconfigure.data.redis;
 
 import java.time.Duration;
+import java.util.regex.Pattern;
 
 import io.lettuce.core.ClientOptions;
+import io.lettuce.core.ReadFrom;
 import io.lettuce.core.RedisClient;
 import io.lettuce.core.SocketOptions;
 import io.lettuce.core.TimeoutOptions;
@@ -51,6 +53,7 @@ import org.springframework.data.redis.connection.lettuce.LettuceClientConfigurat
 import org.springframework.data.redis.connection.lettuce.LettuceClientConfiguration.LettuceClientConfigurationBuilder;
 import org.springframework.data.redis.connection.lettuce.LettuceConnectionFactory;
 import org.springframework.data.redis.connection.lettuce.LettucePoolingClientConfiguration;
+import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
 
 /**
@@ -163,6 +166,10 @@ class LettuceConnectionConfiguration extends RedisConnectionConfiguration {
 			if (lettuce.getShutdownTimeout() != null && !lettuce.getShutdownTimeout().isZero()) {
 				builder.shutdownTimeout(getProperties().getLettuce().getShutdownTimeout());
 			}
+			RedisProperties.Lettuce.ReadFrom readFrom = lettuce.getReadFrom();
+			if (readFrom.getType() != null) {
+				builder.readFrom(getReadFrom(readFrom));
+			}
 		}
 		if (StringUtils.hasText(getProperties().getClientName())) {
 			builder.clientName(getProperties().getClientName());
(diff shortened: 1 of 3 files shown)


wilkinsona:
Thanks, @nosan. Unfortunately, when I looked at this yesterday, I mistakenly thought that `ReadFrom` was an enum. and the changes proposed here demonstrate that's not the case.  Sorry for any wasted effort here but I'm not sure that I like having three properties (`read-from.type`, `read-from.cidr-notations`, and `read-from.pattern`). A lot of the time, only `type` will be needed and the other two are only needed when `type` has a particular value.

I wonder if we could still use a single property. The values that map directly to a `ReadFrom` constant are straightforward but we'd need some special handling for `subnet` and `regex`. Perhaps we could support something like `subnet:notationOne,notationTwo,notationThree` and `regex:pattern`? This support could perhaps be implemented as a `@ConfigurationPropertiesBinding` converter and the property could be typed as a `ReadFrom` instance.

I don't want to waste any more of your time (apologies again for the time possibly already wasted) so please don't do anything about the above until we've had a chance to discuss this as a team.

nosan:
Thanks, @wilkinsona.
No problem at all.

nosan:
```java

@Bean
@ConfigurationPropertiesBinding
StringToReadFromConverter stringToReadFromConverter() {
   return new StringToReadFromConverter();
}
	

static class StringToReadFromConverter implements GenericConverter {

		private static final Set<ConvertiblePair> CONVERTIBLE_TYPES;

		static {
			CONVERTIBLE_TYPES = Set.of(new ConvertiblePair(String.class, ReadFrom.class));
		}

		@Override
		public Set<ConvertiblePair> getConvertibleTypes() {
			return CONVERTIBLE_TYPES;
		}

		@Override
		public Object convert(Object source, TypeDescriptor sourceType, TypeDescriptor targetType) {
			if (source == null) {
				return null;
			}
			String value = source.toString();
			int index = value.indexOf(':');
			if (index != -1) {
				String type = value.substring(0, index);
				if (type.equalsIgnoreCase("subnet")) {
					return ReadFrom.subnet(StringUtils.commaDelimitedListToStringArray(value.substring(index + 1)));
				}
				if (type.equalsIgnoreCase("regex")) {
					return ReadFrom.regex(Pattern.compile(value.substring(index + 1)));
				}
			}
			return ReadFrom.valueOf(value);
		}

	}
	
```
in that case, it will be possible to use the following syntax:


```properties
spring.data.redis.lettuce.read-from=regex:192.*
spring.data.redis.lettuce.read-from=subnet:192.12.128.0/32,192.168.255.255/32
spring.data.redis.lettuce.read-from=any
```

nosan:
Is it ok to use `ReadFrom` field type from `io.lettuce.core` in `RedisProperties.Lettuce` properties? 

```java
/**
* Defines from which Redis nodes data is read.
*/
private ReadFrom readFrom;

```

nosan:
perhaps it would be beneficial to reach out to someone from the Lettuce team to explore the possibility of adding support for the `regex` and `subnet` types in the `ReadFrom.valueOf(...)` method

At the moment:

```java

 //ReadFrom valueOf(String name)
 
        if (name.equalsIgnoreCase("subnet")) {
            throw new IllegalArgumentException("subnet must be created via ReadFrom#subnet");
        }

        if (name.equalsIgnoreCase("regex")) {
            throw new IllegalArgumentException("regex must be created via ReadFrom#regex");
        }
        
 ```

wilkinsona:
> perhaps it would be beneficial to reach out to someone from the Lettuce team

I think that's a good idea. WDYT, @mp911de, about supporting `subnet:…` and `regex:…` or similar syntax with `ReadFrom.valueOf` to ease property-based configuration?

mp911de:
I like the idea a lot, especially flexibility introduced with `regex:` and `subnet:` variants is pretty exciting.

wilkinsona:
Thanks, @mp911de. I've opened https://github.com/redis/lettuce/issues/3013 for further consideration.

nosan:
@wilkinsona 
https://github.com/redis/lettuce/pull/3016 It's been merged and will be available in Lettuce **6.5.0.RELEASE**.
Meanwhile, I've updated the PR. Could I kindly ask you to review it once again, please?

I have the following questions:
- How can I correctly map kebab-case values? `ReadFrom.valueOf` supports case-insensitive names (e.g., `upstreampreferred`). I used `readFrom.replaceAll("-", "")`
- What values should I use for `regex` and `subnet` in `additional-spring-configuration-metadata.json`?

philwebb:
> How can I correctly map kebab-case values? ReadFrom.valueOf supports case-insensitive names (e.g., upstreampreferred). I used readFrom.replaceAll("-", "")

I think we can use the same logic as in `LenientObjectToEnumConverterFactory`. I've done that in https://github.com/spring-projects/spring-boot/commit/bea7704b09f2782e60a8490e4e266e9ed9b4db62

> What values should I use for regex and subnet in additional-spring-configuration-metadata.json?

I think the ones you've added are correct.

nosan:
Thank you, @philwebb 
I `cherry-pick` your commit 👍

wilkinsona:
I wonder if we should hold off on documenting `regex:` and `subnet:` in the metadata until we're using Lettuce 6.5 by default. The feature's still useful before then and if someone using Lettuce 6.4 wants to configure a regex or subnet read from, they can always use a customizer till Lettuce 6.5 is out and they've overridden the version or we've upgraded.

nosan:
> I wonder if we should hold off on documenting `regex:` and `subnet:` in the metadata until we're using Lettuce 6.5 by default. The feature's still useful before then and if someone using Lettuce 6.4 wants to configure a regex or subnet read from, they can always use a customizer till Lettuce 6.5 is out and they've overridden the version or we've upgraded.

It depends on when it gets merged :) If you think it's ready to be merged, I can remove subnet: and regex: from the metadata.

wilkinsona:
We discussed this today and decided that we're going to err on the side of caution and wait till Lettuce 6.5 is out before merging anything. Hopefully that'll be in the Boot 3.5 timeframe.

nosan:
[6.5.0.RELEASE](https://github.com/redis/lettuce/releases/tag/6.5.0.RELEASE) has been released. I bumped the version to 6.5.0 and uncommented tests for `regex` and `subnet`.

philwebb:
Thanks @nosan. We're going to wait for 3.5 to merge this one. The version bump is a bit risky post RC

nosan:
Thanks, @philwebb. 
Should I revert bump version commit, meanwhile?

philwebb:
No it's fine. We'll just deal with that nearer the time.

Issue #33992, opened 2023-01-26
Detect accidental misconfiguration of @JsonMixin


odrotbohm:
`@JsonMixin` exposes both a `value` and a `type` attribute, the latter being an alias of the former, which means that both attributes define an empty default value.

`JsonMixinModuleEntries`' `registerMixinClass(…)` inspects the attributes and only registers a mixin if the attribute is set. This means that a sole `@JsonMixin` annotation on a class will be found but dropped in that purpose. While this is a pretty obvious mistake, as one does not define which class to apply the mixin for, it can easily be overseen and we could actually detect that misconfiguration. It would be nice if `….registerMixinClass(…)` could reject an empty `@JsonMixin` annotation to avoid this oversight.

MatejNedic:
Hey, is this issue open for taking? :)

philwebb:
Closing in favor of PR #42592. Thanks @nosan!

PR #42592, opened 2024-10-10
Detect accidental misconfiguration of JsonMixin annotation

nosan:
#33992


wilkinsona:
Thanks very much, @nosan.

PR #42599, opened 2024-10-11
Fix Spring Data Cassandra Spring LDAP reference doc links

ngocnhan-tran1996:


mhalbritter:
Thanks @ngocnhan-tran1996 !

Issue #33434, opened 2022-12-01
@TestPropertySource not recognizing yaml files


DataWorm:
Spring Boot supports not only `application.properties` but also `application.yml` files for long time now and we are using it in all our projects for better readability. But when it comes to testing with annotations like `@TestPropertySource` then yaml files are not supported for whatever reason. If you try to specify a yml file as source, it does not complain but the test also won't be able to resolve the properties unless you are using the properties notation format and just store it with the `.yml` file extension. 

I wonder why this is not supported yet. Is there another annotation I haven't seen yet or is there any plan to let `TestPropertySource` automatically recognize the file extension and make use of the corresponding yaml file loader?

philwebb:
It's currently not supported because `@TestPropertySource` is a Spring Framework annotation and isn't aware of YAML. We've had a request to support it previously (#11921) which resulted in Framework issue that was ultimately closed. I wonder if we should try again for Framework 6.1.

@sbrannen any thought on this? Perhaps we could have a hook point to allow `@TestPropertySource` to deal with different file types?

sbrannen:
> @sbrannen any thought on this? Perhaps we could have a hook point to allow `@TestPropertySource` to deal with different file types?

Yes, I'd like to introduce a new `Class<? extends PropertySourceFactory> factory() default PropertySourceFactory.class;` attribute in `@TestPropertySource` analogous to the one already present in `@PropertySource` since Spring Framework 4.3.

@philwebb and @wilkinsona, what do you think about that?

sbrannen:
> Though, I'll need to investigate the feasibility of such a change.

At a first glance, the biggest challenge I foresee with supporting a custom `factory` per annotation is that `MergedContextConfiguration.getPropertySourceLocations()` contains all merged locations, meaning there would not be any easy way to determine which `factory` to use for which files.

One option would be to use the first (or last) configured factory to process all locations -- which would require that all locations configured via `@TestPropertySource` on a given test class be the same type of file format.

Another option would be to create some sort of map from file extension to factory -- though I'm not sure how we would want to support that mapping via annotation-based configuration.

Thoughts?

wilkinsona:
I like the idea of aligning with `@PropertySource` but hadn't considered the complications introduced by `MergedContextConfiguration`.

[...]

PR #42603, opened 2024-10-11
Add YamlPropertySourceFactory that can be used for loading YAML files through the @TestPropertySource annotation

nosan:
I have prepared two potential solutions for this #33434 enhancement.

The first approach offers a simple and intuitive way to use it:

```java
@SpringJUnitConfig
@TestPropertySource(locations = "classpath:test.yaml", factory = YamlPropertySourceFactory.class)
class SomeTest {

}
```

`YamlPropertySourceFactory` loads all properties from the specified YAML file using `YamlPropertySourceLoader` and returns a `CompositePropertySource` containing the loaded properties.

As a potential improvement, a dedicated annotation like `@TestYamlPropertySource` could be introduced to simplify the loading of YAML files in tests.

```java
@Target(ElementType.TYPE)
@Documented
@Inherited
@TestPropertySource(factory = YamlPropertySourceFactory.class)
@Repeatable(TestYamlPropertySources.class)
public @interface TestYamlPropertySource {

	@AliasFor(attribute = "value", annotation = TestPropertySource.class)
	String[] value() default {};

	@AliasFor(attribute = "locations", annotation = TestPropertySource.class)
	String[] locations() default {};

	@AliasFor(attribute = "inheritLocations", annotation = TestPropertySource.class)
	boolean inheritLocations() default true;

	@AliasFor(attribute = "properties", annotation = TestPropertySource.class)
	String[] properties() default {};

	@AliasFor(attribute = "inheritProperties", annotation = TestPropertySource.class)
	boolean inheritProperties() default true;

}

```
With a dedicated annotation, the following syntax could be supported:

```java
@SpringJUnitConfig
@TestYamlPropertySource({ "classpath:test.yaml", "classpath:test1.yaml" })
@TestYamlPropertySource(locations = "classpath:test2.yaml", properties = "key:value")
class TestYamlPropertySourceIntegrationTests {

}
```

---

An [alternative](https://github.com/nosan/spring-boot/commit/766b6b2b460d525ed2619b2515a20218a804a009) approach is to customize `PropertySourceDescriptor` in `SpringBootTestContextBootstrapper` by using the `PropertySourceLoaderPropertySourceFactory`, which leverages the existing `PropertySourceLoader`. With that approach, the following syntax would be possible:

```java


@SpringBootTest
@TestPropertySource(locations = "classpath:test.yaml")
class SomeTest {
	
}
```

But it has several disadvantages:

- It works only with the `@SpringBootTest` annotation. For `@SpringJUnitConfig`, the factory needs to be explicitly used.

- It does not support `EncodedResource` for **.properties** files because `PropertySourceLoader` only works with Resource objects. As a result, if someone is using `@TestPropertySource` with the encoding attribute, these changes could potentially impact them.
- Not obvious.

The first approach is much better and straightforward, and people would not have any problem using it.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
@@ -525,7 +525,7 @@ The lines immediately before and after the separator must not be same comment pr
 TIP: Multi-document property files are often used in conjunction with activation properties such as `spring.config.activate.on-profile`.
 See the xref:features/external-config.adoc#features.external-config.files.activation-properties[next section] for details.
 
-WARNING: Multi-document property files cannot be loaded by using the javadoc:org.springframework.context.annotation.PropertySource[format=annotation] or javadoc:org.springframework.test.context.TestPropertySource[format=annotation] annotations.
+WARNING: Multi-document property files cannot be loaded by using the javadoc:org.springframework.context.annotation.PropertySource[format=annotation], javadoc:org.springframework.test.context.TestPropertySource[format=annotation] or javadoc:org.springframework.boot.test.context.TestYamlPropertySource[format=annotation] annotations.
 
 
 
--- a/spring-boot-project/spring-boot-test/build.gradle
+++ b/spring-boot-project/spring-boot-test/build.gradle
@@ -58,5 +58,5 @@ dependencies {
 	testImplementation("org.testng:testng")
 
 	testRuntimeOnly("org.junit.vintage:junit-vintage-engine")
+	testRuntimeOnly("org.yaml:snakeyaml")
 }
-
--- a/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/TestYamlPropertySource.java
+++ b/spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/TestYamlPropertySource.java
@@ -1,0 +1,98 @@
+/*
+ * Copyright 2012-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.test.context;
+
+import java.lang.annotation.Documented;
+import java.lang.annotation.ElementType;
+import java.lang.annotation.Inherited;
+import java.lang.annotation.Repeatable;
+import java.lang.annotation.Retention;
+import java.lang.annotation.RetentionPolicy;
+import java.lang.annotation.Target;
+
+import org.springframework.core.annotation.AliasFor;
+import org.springframework.test.context.TestPropertySource;
+
+/**
+ * {@code @TestYamlPropertySource} is an annotation that can be applied to a test class to
+ * configure the locations of YAML files and inlined properties to be added to the
+ * Environment's set of PropertySources for an ApplicationContext for integration tests.
+ * <p>
+ * Provides a convenient alternative for
+ * {@code @TestPropertySource(locations = "...", factory = YamlPropertySourceFactory.class)}.
+ * <p>
+ * {@code @TestYamlPropertySource} should be considered as {@code @TestPropertySource} but
+ * for YAML files. It intentionally does not support multi-document YAML files to maintain
+ * consistency with the behavior of {@code @TestPropertySource}.
+ *
+ * @author Dmytro Nosan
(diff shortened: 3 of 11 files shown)


nosan:
The third [option](https://github.com/nosan/spring-boot/commit/35f366897b30ea945d61f06540eecf9c79861efe)
is similar to the first approach, but instead of supporting only **.yaml** files, it also
supports _multi-document_ **.properties**.

```java

@SpringJUnitConfig
@TestPropertySource(locations = { "classpath:test.yaml", "classpath:test1.yaml", "classpath:test.properties" },
		factory = PropertySourceLoaderPropertySourceFactory.class)
class PropertySourceLoaderPropertySourceFactoryIntegrationTests {

}

```

philwebb:
I like `YamlPropertySourceFactory` since it's quite obvious what it does. Once thing that makes me pause is that we don't/can't support `spring.config.activate.on-profile` for multi-document files. We also can't support `spring.config.import`. I'm not sure if this is really a problem, but folks using the annotation will need to understand the restrictions.

nosan:
> Once thing that makes me pause is that we don't/can't support spring.config.activate.on-profile for multi-document files.

I was trying to add this support, but to be honest, I have not found a way. 

> I'm not sure if this is really a problem

I think most people will use a dedicated YAML file to configure their test context, and I doubt anyone will try to use the `spring.config.activate.on-profile` feature


> annotation will need to understand the restrictions.

I completely agree, and it would be beneficial to include this information in the Javadoc and documentation.

philwebb:
Perhaps we should make it a hard requirement that only a single document is in the YAML and that no profile restrictions or imports are specified. We'll discuss this on a team call when can.

philwebb:
I'm wondering if basic YAML support should be something that's added to Spring Framework. Perhaps with some hook point so that we can warn if users accidentally point at YAML files that won't work. What do you think @sbrannen?

sbrannen:
> I'm wondering if basic YAML support should be something that's added to Spring Framework. 

We actually considered that back when we introduced `@TestPropertySource(factory = ...)` in Spring Framework 6.1.

I also implemented a custom [`YamlPropertySourceFactory`](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/test/java/org/springframework/test/context/env/YamlPropertySourceFactory.java) and an accompanying [`@YamlTestProperties`](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/test/java/org/springframework/test/context/env/YamlTestProperties.java) annotation as a proof of concept.

You can see it in action in [`YamlTestPropertySourceTests`](https://github.com/spring-projects/spring-framework/blob/main/spring-test/src/test/java/org/springframework/test/context/env/YamlTestPropertySourceTests.java).

The reason we didn't make that an official feature of the Spring TestContext Framework is that we doubted that it would meet the needs of enough developers.

The rationale was that Framework's YAML support in `YamlPropertiesFactoryBean` is very basic and that many developers would likely expect it to be as powerful as Boot's built-in YAML support. So, since we did not want to port Boot's YAML support to Framework and since the implementation is relatively straightforward, we decided it was best not to provide `YamlPropertySourceFactory` directly in Framework.

> Perhaps with some hook point so that we can warn if users accidentally point at YAML files that won't work.

What kind of hook point did you have in mind, and who would hook into that?

philwebb:
> What kind of hook point did you have in mind, and who would hook into that?

If you added some kind of YAML support, we could have a `spring.factories` hook to make sure that the YAML is sensible from a Spring Boot perspective. It seems like perhaps it's best just to keep `YamlPropertySourceFactory` entirely in Boot based on your previous comments.

nosan:
Honestly, the more I think about this PR, the more I want to close it.
This YAML support is so limited, and I’m not sure whether it’s worth including in Spring Boot at all.

If support for `spring.config.activate.on-profile`, `spring.config.import`, multi-document, etc., were possible,
a dedicated annotation might be considered for addition.


As a compromise, `YamlPropertySourceFactory` could be added, but having a dedicated annotation seems too much.


```java

public class YamlPropertySourceFactory implements PropertySourceFactory {

	private static final YamlPropertySourceLoader loader = new YamlPropertySourceLoader();

	@Override
	public PropertySource<?> createPropertySource(String name, EncodedResource encodedResource) throws IOException {
		Resource resource = encodedResource.getResource();
		String propertySourceName = getPropertySourceName(name, resource);
		List<PropertySource<?>> propertySources = loader.load(propertySourceName, resource);
		CompositePropertySource compositePropertySource = new CompositePropertySource(propertySourceName);
		propertySources.forEach(compositePropertySource::addFirstPropertySource);
		return compositePropertySource;
	}

	private static String getPropertySourceName(String name, Resource resource) {
		if (StringUtils.hasText(name)) {
			return name;
		}
		String description = resource.getDescription();
		if (StringUtils.hasText(description)) {
			return description;
		}
		return resource.getClass().getSimpleName() + "@" + System.identityHashCode(resource);
	}


}
```

And if the provided `YAML` is a multi-document such as:

```yaml
spring:
  bar: bar
  foo: baz
---
spring:
  foo: foo
```

The values from the later document replace those from the earlier one.

```properties
spring.bar=bar
spring.foo=foo
```

This behaviour will be similar to `ResourcePropertySource`, which does not support multi-document properties, causing the later values to overwrite the earlier ones.

The usage of this factory will not be verbose:

```java
@TestPropertySource(locations = { "test.yaml", "test1.yaml" }, factory = YamlPropertySourceFactory.class)
class YamlTestPropertySourceIntegrationTests { }
```

From my perspective, it is the best option.

PR #42606, opened 2024-10-13
Polish SpringApplication

gkdis6:
- Reuse the `webApplicationType` variable in both methods to avoid redundant calls.
- Simplify null handling for environmentType by adopting a consistent approach across both methods.
- Enhance overall readability and reduce code duplication.


mhalbritter:
Thank you very much @gkdis6 and congratulations on your first contribution :tada:!

PR #42610, opened 2024-10-14
Polish ConfigurationPropertiesBinder

quaff:
1. `ConfigurationPropertiesBinder.` is redundant
2. `type` is `@Nullable`


mhalbritter:
Thanks @quaff !

Issue #33870, opened 2023-01-18
 Kotlin Jackson Mapper not used when on classpath when serializing/deserializing @JdbcTypeCode(SqlTypes.JSON) using Spring Data JPA


BernhardPosselt:
After migrating to Spring Boot 3.0.1 and Hibernate 6, I swapped out the Hibernate Types JSON support for the new built in **@JdbcTypeCode(SqlTypes.JSON)**, however my Kotlin data classes failed to deserialize and I had to annotate everything with **@JsonCreator** and **@JsonProperty** to fix constructors.

I feel like this should default to the Jackson instance on the Spring context rather than the Jackson library itself, that is pulled in by Hibernate by default. It should be possible to define a custom mapper factory that can be configured via **hibernate.type.json_format_mapper**  and point Hibernate to that instance, see https://github.com/hibernate/hibernate-orm/blob/main/documentation/src/main/asciidoc/userguide/chapters/domain/basic_types.adoc#json-mapping

wilkinsona:
Thanks for the suggestion. If I have undertood it correctly, you should be able to do this yourself using a `HibernatePropertiesCustomizer` bean:

```java
@Bean
HibernatePropertiesCustomizer jsonFormatMapperCustomizer(ObjectMapper objectMapper) {
	return (properties) -> properties.put(AvailableSettings.JSON_FORMAT_MAPPER,
			new JacksonJsonFormatMapper(objectMapper));
}
```

Can you please try this and let us know if it allows you to remove the `@JsonCreator` and `@JsonProperty` annotations?

BernhardPosselt:
Thank you for getting back, can confirm that this does in fact work fine after removing the annotations.

davidfortin55555:
I have implemented the following code but it is nor working with springboot 3.0.1 and hibernate 6
@Bean
HibernatePropertiesCustomizer jsonFormatMapperCustomizer(ObjectMapper objectMapper) {
	return (properties) -> properties.put(AvailableSettings.JSON_FORMAT_MAPPER,
			new JacksonJsonFormatMapper(objectMapper));
}
Do you habe another proposal?

wilkinsona:
It should work so, without knowing exactly how it isn't working, it's difficult to propose something else. If you'd like some help, please post a question on Stack Overflow that includes a minimal, reproducible example of the problem.

wilkinsona:
While looking at #42676 and trying to find some inspiration for a property name, a couple of other things came to mind:

[...]

PR #42676, opened 2024-10-14
Auto-configure Hibernate with a JsonFormatMapper

nosan:
#33870

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaAutoConfiguration.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks, @nosan.

Before merging this, we should consider if we want a way to disable it (an application property?) as it has the potential to be a breaking change.

nosan:
Thanks, @wilkinsona 
I was also thinking about this, but I have not found a good name for it. Maybe `spring.jpa.hibernate.auto-configure.json-format-mapper?` ?

nosan:
I also used 
`properties.putIfAbsent` in case if someone already has already registered they own `JsonFormatMapper`

wilkinsona:
Thanks, @nosan. While looking for some inspiration for a property name, a couple have other things came to mind. I've noted them [in a comment on 33870](https://github.com/spring-projects/spring-boot/issues/33870#issuecomment-2422737522). I'll put this on hold for now until we've had a chance to figure those out.

nosan:
Thanks, @wilkinsona. That makes sense. 

What do you think about the following property format?

```
spring.jpa.hibernate.auto-configure.json-format-mapper=none
spring.jpa.hibernate.auto-configure.json-format-mapper=jackson
spring.jpa.hibernate.auto-configure.json-format-mapper=jsonb
```

By default, for example, `jackson`, as it is more popular.

Issue #42110, opened 2024-09-04
In the README, accessing 'Converting a Spring Boot JAR Application to a WAR' returns a 404 error.


ldlb9527:

philwebb:
Thanks. The guide has been removed so I think we can just drop that line.

PR #42691, opened 2024-10-15
Similar guide links related to not found guide links

IMWoo94:
Version : [Spring Boot 3.3.4 Reference Doc.](https://docs.spring.io/spring-boot/index.html)

Changed the Not Found link provided in the Reference Doc to a link covering the Build Plugin.

The reference Doc already provides a direct content to Jar and War. I think it would create better synergy if you provided a link to the Build Tool rather than providing a removed guide.

Reference Doc : https://docs.spring.io/spring-boot/how-to/deployment/traditional-deployment.html#howto.traditional-deployment.convert-existing-application

Not Found Link : [Getting Started Guide on Converting a jar to a war.]( https://spring.io/guides/gs/convert-jar-to-war/ )

Suggest Link : [Build Tool Plugins](https://docs.spring.io/spring-boot/build-tool-plugin/index.html)

스크린샷 2024-10-15 오후 8 41 11

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/how-to/pages/deployment/traditional-deployment.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/how-to/pages/deployment/traditional-deployment.adoc
@@ -74,7 +74,6 @@ This means that, in addition to being deployable to a servlet container, you can
 
 To convert an existing non-web Spring application to a Spring Boot application, replace the code that creates your `ApplicationContext` and replace it with calls to `SpringApplication` or `SpringApplicationBuilder`.
 Spring MVC web applications are generally amenable to first creating a deployable war application and then migrating it later to an executable war or jar.
-See the https://docs.spring.io/spring-boot/build-tool-plugin/index.html[build-tool-plugin].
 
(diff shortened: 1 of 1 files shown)


snicoll:
Thanks for the PR.

The guide has been dropped so this looks like an oversight, see https://github.com/spring-projects/spring-boot/issues/42110

IMWoo94:
Thank you for finding the issue

The current PR provides a new link, but can I delete it and rewrite it with the code that's been removed?

philwebb:
Thanks for the PR. I don't think we should reproduce the code from the old guide, but we should update the link when we merge this. It should be `xref:build-tool-plugin:index.adoc[]`

IMWoo94:
Update can be carried out in two ways.

1. "See the Getting Started Guide on Converting a jar to a war." 문구 삭제
2. Modify links using xref:build-tool-plugin:index.adoc[]

#42110 and the comments
Number 2 provides the same link using xref:build-tool-plugin:index.adoc[] at the top.
This is overlapping, and I think option 2 is ambiguous because it is located not far away.

I think option 1 is correct.
Which of the two options do you think is more correct?

philwebb:
Thanks @IMWoo94

PR #42700, opened 2024-10-15
Fix typos in actuator docs for "when_authorized"

t0r0X:
This PR fixes two typos in documentation of actuator health endpoints:
"Spring Boot > Reference > Production-ready Features > Endpoints"

https://docs.spring.io/spring-boot/reference/actuator/endpoints.html#actuator.endpoints.health

Wrong: `when-authorized`
Correct: `when_authorized`

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/actuator/endpoints.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/actuator/endpoints.adoc
@@ -550,7 +550,7 @@ The information exposed by the `health` endpoint depends on the configprop:manag
 | `never`
 | Details are never shown.
 
-| `when-authorized`
+| `when_authorized`
 | Details are shown only to authorized users.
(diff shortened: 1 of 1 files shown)


philwebb:
Thanks, but we prefer kebab case in properties and both of those should work in that form.

t0r0X:
I did know that kebab-case is preferred for properties names, does this apply also to properties values?
The PR is about the value of property `management.endpoint.health.show-details`, not the name. Its own documentation says:
```
Values:
    never:           ...
    when_authorized: ...
    always:          ...
```

Also please note that in this section https://docs.spring.io/spring-boot/reference/actuator/endpoints.html#actuator.endpoints.sanitization there are already two usages of the enum value, but with upper case: `show-values: WHEN_AUTHORIZED ` (backed by enum `org.springframework.boot.actuate.endpoint.Show`)

philwebb:
Yes, we generally use it for both. That's the format used in the configuration properties metadata file. Thanks for pointing out the issue with the "Sanitize Sensitive Values" section, I've fixed that in #42702

PR #42707, opened 2024-10-16
Update `DataSourceHealthIndicator`, `ConnectionFactoryHealthIndicator` to improve code readability and maintainability

tlsgmltjd:
Hello, I'm a 17-year-old developer living in South Korea. I'm looking at the actuator part of the Spring Boot project, and I found something that I think would be good to improve and I'm giving you a suggestion!

I found a string literal used repeatedly in the `DataSourceHealthIndicator` and `ConnectionFactoryHealthIndicator` classes and managed it as a constant.

I also modified the Indicator class and test code. 

```java
public static final String DATABASE = "database";

public static final String VALIDATION_QUERY = "validationQuery";
```

I would really appreciate it if you could check it.

> Please understand that my English is not good enough


wilkinsona:
Unfortunately, I don't think this improves the readability of the code. It's also debatable that it improves the maintainability as those values aren't something that we change. FWIW, I'm also not keen on using constants in tests as it increases the chance that a typo will go unnoticed. Thanks anyway for the proposal.

tlsgmltjd:
Thank you for your confirmation!

PR #42718, opened 2024-10-16
Fix typos and formatting errors in documentation

ngocnhan-tran1996:
I fixed typo document


philwebb:
Thanks. The `tag::code[]` comments should not be required, are you seeing problems with those samples?

wilkinsona:
As shown in the docs, the Kotlin snippets are empty at the moment: https://docs.spring.io/spring-boot/reference/io/jta.html#io.jta.mixing-xa-and-non-xa-connections

philwebb:
Thanks @wilkinsona. We might want to go the other way and update the Java samples to remove the code tag, that way imports can be shown in the sample.

ngocnhan-tran1996:
@philwebb 
I removed the code tag in Java sample and reverted the code tag in Kotlin sample in local. The result is the code snippets are empty, both Java and Kotlin

wilkinsona:
I think the includes need to be updated as well to remove `[tag=*]`.

ngocnhan-tran1996:
Thanks @wilkinsona 

It works. I will update PR

philwebb:
Thanks very much @ngocnhan-tran1996

PR #42751, opened 2024-10-17
Upgrade to Spring RESTDocs 3.0.2

santo-raj:
Implemented below issue 

https://github.com/spring-projects/spring-boot/issues/42745


wilkinsona:
Thanks for the PR, but, as per the PR template, we prefer to handle single-line version upgrades like this ourselves:

https://github.com/spring-projects/spring-boot/blob/4b87918b6d410cc3cb63b8a32a30e81ead77400e/.github/PULL_REQUEST_TEMPLATE.md?plain=1#L14-L19

Furthermore, 3.0.2 hasn't been released yet so these changes will not build cleanly.

PR #42770, opened 2024-10-17
Add GitHub Issue templates

nosan:
Screenshot 2024-10-17 at 18 01 50

If you think it's a good idea, it might be worth considering adding templates for PRs.

### DIFF
--- a/.github/ISSUE_TEMPLATE/issue.md
+++ b/.github/ISSUE_TEMPLATE/issue.md
@@ -1,5 +1,5 @@
 ---
-name: Issue
+name: General
 about: Bugs, enhancements, documentation, tasks.
 title: ''
 labels: ''


SECURITY.md
```diff
@@ -0,0 +1,5 @@
+# Reporting a Vulnerability
+
+Please, [open a draft security advisory](https://github.com/spring-projects/security-advisories/security/advisories/new) if you need to disclose and discuss a security issue in private with the Spring Boot team. Note that we only accept reports against [supported versions](https://spring.io/projects/spring-boot#support).
+
+For more details, check out our [security policy](https://spring.io/security-policy).
```
wilkinsona, line 5: I don't think we need this as we already inherit the policy from spring-projects/.github. We can just link to https://github.com/spring-projects/spring-boot/security/policy.

nosan: Added to `contact_links` but the issue page changed a bit.
Screenshot 2024-10-17 at 19 07 42

nosan:
@mhalbritter 
I reviewed the issue page at https://github.com/spring-io/start.spring.io/issues/new/choose, and I noticed it includes a `Report of a security vulnerability` option, but the repository itself does not have a `SECURITY.md`.

Is it possible to do something similar in the Spring Boot repository? If so, I could remove the link from `contact_links`

SECURITY.md
nosan: Report a security vulnerability option should be inherited from https://github.com/spring-projects/.github automatically.

bclozel:
We recently applied this in Spring Framework because we had several security reports made public as issues. I'm not sure we have the same problem in the Spring Boot repository at this point.

Also about the issue type, its title is "Issue" which renders as "Issue: Issue" in the GitHub UI. See https://github.com/spring-projects/spring-framework/issues/new?assignees=&labels=&projects=&template=issue.md&title=

nosan:
Thank you, @snicoll 

Screenshot 2024-10-22 at 12 38 45

PR #42798, opened 2024-10-19
Polish

ngocnhan-tran1996:
This PR includes
- Simplify ternary operator condition. I think the method can return direct result instead
- Use `String#isEmpty` instead `String#equals("")`
- Polish diamond operator usage


mhalbritter:
Thanks @ngocnhan-tran1996 !

Issue #42781, opened 2024-10-18
Add property to control export of OTLP logs


andrzej-fisglobal:
In Spring Boot 3.4.0-M3, `OtlpLoggingConfigurations.otlpHttpLogRecordExporter` is a created if `OtlpLoggingConnectionDetails` is created, the latter is a conditionally created if the property `management.otlp.logging.endpoint` is defined. It looks like the only option to disable this to not define the endpoint key.

However, it makes it a bit difficult if we have already defined the endpoint and want to disable it. 

Is it possible to define an `enabled` toggle in a similar manner that we have for e.g. `management.otlp.metrics.export`? In general, it would be nice to be able to disable signal exporters in a consistent way -- thus something similar to `OnEnabledTracingCondition` and `OnMetricsExportEnabledCondition`?

mhalbritter:
We have `management.*.metrics.export.enabled`, `management.*.tracing.export.enabled` and I think for consistency we should also add something like `management.otlp.logging.export.enabled`.

nosan:
You should be able to disable the `otlpHttpLogRecordExporter` by setting `management.otlp.logging.transport=none`, right?

nosan:
But this property does not exist in `additional-spring-configuration-metadata.json` 🥲

nosan:
Apologies @mhalbritter and @andrzej-fisglobal for the confusion. I overlooked the fact that this property maps to enum values. My apologies for the oversight.

mhalbritter:
Don't worry, no harm done :)

andrzej-fisglobal:
Thanks for your comment. That makes it indeed a bit easier, but still, it would be nice to have consistency among those -- what Moritz proposed.

mhalbritter:
Superseded in favor of #42813.

PR #42813, opened 2024-10-21
Add property to control export of OTLP logs

nosan:
#42781


andrzej-fisglobal:
Thanks!

nosan:
Thanks @mhalbritter 

the funny thing is that I initially put an annotation on `Exporters` and then for some reason moved to `Autoconfiguration`
By the way, thanks for the polishing.

Issue #41933, opened 2024-08-19
Improve documented logging property descriptions and default values


mnisius:
Hello, now that I'm able to set a custom default for `logging.pattern.console` i noticed that the documentation of the default value seems to be slightly incorrect. 

By Default Log messages look like this: 

```
2024-08-19T15:17:36.200+02:00  INFO 7876 --- [  restartedMain] d.c.operations.rest.RestServiceApp       : Starting RestServiceApp using Java 21.0.4 with PID 7876 (D:\playground\event-test\target\classes started by user in D:\playground\event-test)
2024-08-19T15:17:36.202+02:00  INFO 7876 --- [  restartedMain] d.c.operations.rest.RestServiceApp       : No active profile set, falling back to 1 default profile: "default"
```

The documentation says about `logging.pattern.console` : 

Appender pattern for output to the console. Supported only with the default Logback setup. 
Default: `%clr(%d{${LOG_DATEFORMAT_PATTERN:-yyyy-MM-dd'T'HH:mm:ss. SSSXXX}}){faint} %clr(${LOG_LEVEL_PATTERN:-%5p}) %clr(${PID:- }){magenta} %clr(---){faint} %clr([%15.15t]){faint} %clr(%-40.40logger{39}){cyan} %clr(:){faint} %m%n${LOG_EXCEPTION_CONVERSION_WORD:-%wEx}`

But ff I copy paste this value set this as the default the log looks like this:

```
-2024-08-19T15:51:14. 008+02:00 - INFO 17532 --- [  restartedMain] d.c.operations.rest.RestServiceApp       : Starting RestServiceApp using Java 21.0.4 with PID 17532 (D:\playground\event-test\target\classes started by v108745 in D:\playground\event-test)
--2024-08-19T15:51:14. 010+02:00 - INFO 17532 --- [  restartedMain] d.c.operations.rest.RestServiceApp       : No active profile set, falling back to 1 default profile: "default"
 ```

I after some tests I think this is the correct default:
`%clr(%d{${LOG_DATEFORMAT_PATTERN:yyyy-MM-dd'T'HH:mm:ss.SSSXXX}}){faint} %clr(${LOG_LEVEL_PATTERN:%5p}) %clr(${PID:- }){magenta} %clr(---){faint} %clr([%15.15t]){faint} %clr(%-40.40logger{39}){cyan} %clr(:){faint} %m%n${LOG_EXCEPTION_CONVERSION_WORD:%wEx}` 

[...]

PR #42848, opened 2024-10-23
Move default value descriptions to "description" in logging property metadata

izeye:
This PR moves default value descriptions to "description" properties in logging property metadata as it doesn't seem to be conventional.

Feel free to close this if it's intentional for some reason.

See gh-41933

### DIFF
--- a/spring-boot-project/spring-boot/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -175,16 +175,14 @@
     {
       "name": "logging.pattern.console",
       "type": "java.lang.String",
-      "description": "Appender pattern for output to the console.",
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks, @izeye, but this is intentional. The default value in the metadata is equivalent to the value to which a field is initialized in a `@ConfigurationProperties` class. As such, it shouldn't be descriptive. It should just be the actual default value. In this case, there isn't such a value so we have to add something to the description explaining the situation.

izeye:
@wilkinsona Thanks for the feedback!

> In this case, there isn't such a value so we have to add something to the description explaining the situation.

I'm a bit confused with this as it sounds like the description is added to the `description` property, not the `defaultValue` property.

You mean "the defaultValue explaining the situation", not "the description explaining the situation", right?

wilkinsona:
Oh dear, I somehow managed to read the diff backwards. Thanks for persevering with me. My comment above is, in fact, an explanation of why your change is needed 😬.

mhalbritter:
Thanks @izeye !

PR #42856, opened 2024-10-23
provide a solution to delete temporary directories on server shutdown

### LINKED ISSUE #9983: Tomcat temporary's directories are not deleted

CharlesLgn:
…context is closed #9983

The directory /tmp/tomcat-* are not deleted when stopping a Spring Boot application.
This PR provide a solution with backward compatibility

 - Add a parameter server.tmp-deletion-strategy with 3 possible value :
   - `DELETE_ON_EXIT` : The default strategy. when the context is shutdown, the created temporary directory is deleted. If the temporary directory is not empty, then it will not be deleted.
   - `RECURSIVE_DELETE` : When the context is shutdown, then created temporary directory is deleted with all its content.
   - `NOTHING` : When the context is shutdown, we keep all the created temporary directory. With this configuration, if you need to remove the directory, you need to do it manually or with a cron task.
 - Use this parameter when the Servlet needs to create a temporary directory

by default, the behaviour is the same. but if you want to delete everything, you need to add in your `application.properties` ;

```
server.tmp-deletion-strategy=RECURSIVE_DELETE
```

NB : this is a proposal to use Tomcat/Jetty/Undertow solutions to delete the tmp dirs after the server is stopped

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -40,6 +40,7 @@ import org.springframework.boot.web.server.Http2;
 import org.springframework.boot.web.server.MimeMappings;
 import org.springframework.boot.web.server.Shutdown;
 import org.springframework.boot.web.server.Ssl;
+import org.springframework.boot.web.server.TempDirectoryDeletionStrategy;
 import org.springframework.boot.web.servlet.server.Encoding;
 import org.springframework.boot.web.servlet.server.Jsp;
 import org.springframework.boot.web.servlet.server.Session;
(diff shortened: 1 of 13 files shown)


wilkinsona:
Thanks for your enthusiasm to contribute, @CharlesLgn, but I'm afraid we're not in a position to review and potentially accept a contribution for this. The core team needs to do some design work for #9983 first. I also note that you're still using a shutdown hook in this proposal. As already explained in https://github.com/spring-projects/spring-boot/pull/42852, that's one approach that we know we definitely cannot use as it will result in a race condition.

PR #42899, opened 2024-10-26
Refactor transformAnt from RepositoryTransformersExtension for clarity

tusharAgarwal2511:
Refactor transformAnt method from RepositoryTransformersExtension to improve readability 

- Inverted condition to return early if MARKER is not found
- Simplified logic by reducing indentation level


philwebb:
Something is wrong with your pull-request as it's showing more than just the change your suggesting. Regardless, these kinds of small stylistic changes are often a matter of personal preference and I think we're happy with this one as it is.

Thanks anyway.

Issue #42802, opened 2024-10-21
Improve document to elaborate why the slash in the map key will be removed


quaff:
https://github.com/spring-projects/spring-boot/blob/50d05192c828a299002ccf71020dbe0f7ea0499b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc?plain=1#L914-L915

It would be nice if more words here to show the motivation.

quaff:
Further more, if possible, I think it should raise error with meaningful message instead of removing characters silently.

wilkinsona:
We documented when and why the special bracket notation is needed in https://github.com/spring-projects/spring-boot/issues/13506. This was then refined in https://github.com/spring-projects/spring-boot/issues/23390. I don't think we should get into the why in the reference documentation. My feeling is that it's a level of detail that the vast majority of users don't need and it will make the documentation unnecessarily verbose. Let's see what the rest of the team thinks.

mhalbritter:
I agree with Andy. I think we should document the observable behavior, and not necessarily the why.

quaff:
> We documented when and why the special bracket notation is needed in #13506. This was then refined in #23390. I don't think we should get into the why in the reference documentation. My feeling is that it's a level of detail that the vast majority of users don't need and it will make the documentation unnecessarily verbose. Let's see what the rest of the team thinks.

How about adding a link to #23390?

bclozel:
With my vote, this makes the majority of the team. Sorry @quaff but the reference docs are not meant to track design decisions and tradeoffs but rather describe how things work.

quaff:
> Further more, if possible, I think it should raise error with meaningful message instead of removing characters silently.

@philwebb How about this proposal?

The current behavior is counter-intuitive, there are several issues that reporter is confused by `my.map./key3=value3` is mapped as `key3=value3`, exception should be thrown to state that last element of the `ConfigurationPropertyName` should be written as `ElementType.INDEXED` instead of `ElementType.NON_UNIFORM`, better than documenting that, not everyone read document before writing code.

Issue #41099, opened 2024-06-14
`:` characters are removed from `application.yaml` keys in mapped values when using `@ConfigurationProperties`


testersen:
With a configuration properties class like this:

```kt
@ConfigurationProperties(prefix = "something")
data class MyConfigurationProperties(val values: Map<String, String>)
```

I would expect to be able to provide the following `application.yaml`:

```yaml
something.values:
  "urn:some:specifier": hello world
  "urn:another:specifier": cool stuff
```

And then expect to receive `MyConfigurationProperties.values` as:

```kt
mapOf(
  "urn:some:specifier" to "hello world",
  "urn:another:specifier" to "cool stuff",
)
```

But the actual received values are:

```kt
mapOf(
  "urnsomespecifier" to "hello world",
  "urnanotherspecifier" to "cool stuff",
)
```

quaff:
It can be reproduced by java:
```java
package com.example.demo;

import java.util.Map;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.boot.context.properties.ConfigurationProperties;
import org.springframework.boot.context.properties.EnableConfigurationProperties;

import com.example.demo.DemoApplication.MyConfigurationProperties;

@SpringBootApplication
@EnableConfigurationProperties(MyConfigurationProperties.class)
public class DemoApplication {

	public static void main(String[] args) {
		try (var ctx = SpringApplication.run(DemoApplication.class, args)) {
			System.out.println(ctx.getBean(MyConfigurationProperties.class));
		}
	}

	@ConfigurationProperties(prefix = "something")
	public record MyConfigurationProperties(Map<String, String> values) {

	}
}

```

quaff:
Here is a workaround:
```yaml
something.values[urn:some:specifier]: hello world
something.values[urn:another:specifier]: cool stuff
```

wilkinsona:
Thanks, @quaff. That's not a workaround, it's the [documented](https://docs.spring.io/spring-boot/reference/features/external-config.html#features.external-config.typesafe-configuration-properties.relaxed-binding.maps) way to preserve keys as-is when binding to a map.

Issue #29582, opened 2022-01-27
UNIFORM


qxo:
```
@Configuration
@ConfigurationProperties("app")
public class AppConfig implements InitializingBean {

    private Map<String, String> mapping;

    public Map<String, String> getMapping() {
        return mapping;
    }

    public void setMapping(Map<String, String> mapping) {
        this.mapping = mapping;
    }

```
```
app:
  mapping:
    /AAA/*1: "/*AAAAAAAAAAA*/"
```
===> why "/AAA/*1" ==> AAA ?

mbhave:
It sounds like you're asking why binding a map key with special characters doesn't contain the actual value. Please see [this section](https://docs.spring.io/spring-boot/docs/2.6.x/reference/html/features.html#features.external-config.typesafe-configuration-properties.relaxed-binding.maps) of the docs to see what needs to be done when binding to a map with special characters in the key.

qxo:
> It sounds like you're asking why binding a map key with special characters doesn't contain the actual value. Please see [this section](https://docs.spring.io/spring-boot/docs/2.6.x/reference/html/features.html#features.external-config.typesafe-configuration-properties.relaxed-binding.maps) of the docs to see what needs to be done when binding to a map with special characters in the key.

Thanks, It's worked for me:)

```
app:
  mapping:
    "[/AAA/*1]": "/*AAAAAAAAAAA*/"
    "[/key2]": "value2"
```

Issue #24548, opened 2020-12-17
Configuration properties binding bug with * in property name


ztomic:
Hi,

I have some configuration properties class which has `Map<String, Integer>` field. Key in map can contain `*` for some ant path matching. If property source contains that keys, binder will bind them without `*` characters and with unexpected values also.

Here is minimal sample to reproduce this:
```java
static class DemoConfigurationProperties {

	private Map<String, Integer> properties = Map.of("foo-*", 1, "*-foo", 2, "*-foo-*", 3);

	public Map<String, Integer> getProperties() {
		return properties;
	}

	public void setProperties(Map<String, Integer> properties) {
		this.properties = properties;
	}
}

/**
 * Tis test fails with:
 * AssertionFailedError:
 * Expected :{foo-*=4, *-foo-*=6, *-foo=5}
 * Actual   :{foo-*=1, *-foo-*=3, *-foo=2, -foo-=6, foo-=6, -foo=6}
 */
@Test
public void testWithSameProperties() {
	DemoConfigurationProperties demoConfigurationProperties = new DemoConfigurationProperties();
	Assertions.assertEquals(Map.of("foo-*", 1, "*-foo", 2, "*-foo-*", 3), demoConfigurationProperties.getProperties());

	// now bind same properties
	StandardEnvironment environment = new StandardEnvironment();
	Map<String, Object> properties = Map.of("demo.properties.foo-*", 4, "demo.properties.*-foo", 5, "demo.properties.*-foo-*", 6);
	environment.getPropertySources().addFirst(new MapPropertySource("demo-map-source", properties));
	Binder.get(environment).bind("demo", Bindable.ofInstance(demoConfigurationProperties));

	Assertions.assertEquals(Map.of("foo-*", 4, "*-foo", 5, "*-foo-*", 6), demoConfigurationProperties.getProperties());
}
```

Note above that all properties which are bound from environment have same value `6` but in environment only one property has value `6`.

NOTE: I'm using Spring Boot 2.4.1

Regards,

ztomic:
On Spring Boot 2.2.10.RELEASE output is not same on each run. (I know it is an unsupported version, just wanted to make sure that my problem is not tied exclusively to the new version)

E.g.
```log
org.opentest4j.AssertionFailedError: 
Expected :{*-foo-*=6, *-foo=5, foo-*=4}
Actual   :{*-foo-*=3, *-foo=2, foo-*=1, -foo-=6, foo-=6, -foo=5}
```

[...]

PR #42904, opened 2024-10-28
Reject binding non-uniform property key as map key

quaff:
Given:
```yaml
my:
  map:
    "/key": "value"
```
---
Before this commit:
It's equivalent to
```yaml
my:
  map:
    "[key]": "value" # "[/key]" is expected
```
Such counter-intuitive behavior will confuse developers, there are several reported issues, incomplete list: GH-41099 GH-29582 GH-24548

---
After this commit:
```
***************************
APPLICATION FAILED TO START
***************************

Description:

Failed to bind properties under 'my.map' to java.util.Map<java.lang.String, java.lang.String>:

    Reason: java.lang.IllegalArgumentException: Please rewrite key '/key' to '[/key]' and surround it with quotes if YAML is using

Action:

Update your application's configuration

```
---

See GH-42802

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
@@ -912,7 +912,7 @@ TIP: We recommend that, when possible, properties are stored in lower-case kebab
 ==== Binding Maps
 
 When binding to `Map` properties you may need to use a special bracket notation so that the original `key` value is preserved.
-If the key is not surrounded by `[]`, any characters that are not alpha-numeric, `-` or `.` are removed.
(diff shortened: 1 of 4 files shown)


philwebb:
Thanks @quaff. I read your comment on the other issue but didn't have a chance to reply due to last minute work for the release. We'll discuss this one when we can.

philwebb:
Thanks for the PR @quaff. I agree that it would be nice to fix this, but I don't think we should put so much YAML knowledge into the `MapBinder`. I think we probably need to do something with `ConfigurationPropertyName` and possibly add a failure analyzer. I'm not too sure on the design details yet, that will have to be something we look at after 3.4 is our of the door.

philwebb:
I've opened #42984 so we don't forget about the issue.

quaff:
> but I don't think we should put so much YAML knowledge into the `MapBinder`

Just for friendly error message, it doesn't couple YAML and `MapBinder`.

PR #42910, opened 2024-10-28
X-Registry-Auth header sent to Docker Engine API contains field "authHeader"

### LINKED ISSUE #42905: X-Registry-Auth header sent to Docker Engine API contains field "authHeader"
When constructing the header value for `X-Registry-Auth` when talking to the Docker Engine API, `org.springframework.boot.buildpack.platform.docker.configuration.JsonEncodedDockerRegistryAuthentication#getAuthHeader` is called. This getter is backed by the field `authHeader`, which is filled from `org.springframework.boot.buildpack.platform.docker.configuration.JsonEncodedDockerRegistryAuthentication#createAuthHeader`.

This uses `SharedObjectMapper.get().writeValueAsBytes(this)`. However, the JSON from that serialization not only includes the necessary fields like `username` and `password`, but also the field `authHeader`, which is only used for caching the constructed header.

We should annotate the `authHeader` field with `@JsonIgnore` and verify in the tests (`DockerRegistryUserAuthenticationTests` and `DockerRegistryTokenAuthenticationTests`) that the header doesn't contain the `authHeader` field in the JSON.

What it looks like:

```json
{
  "authHeader" : null,
  "username" : "user",
  "password" : "secret",
  "email" : "docker@example.com",
  "serveraddress" : "https://docker.example.com"
}
```

What it should look like (note the removed `authHeader` field):

```json
{
  "username" : "user",
  "password" : "secret",
  "email" : "docker@example.com",
  "serveraddress" : "https://docker.example.com"
}
```

1328032567:
## Overview of Changes
1. Add annotation `@JsonIgnore` on variable `authHeader` to avoid auto-serialization.
2. Import package `com.fasterxml.jackson.annotation.JsonIgnore` to make annotation valid.
3. run command `./gradlew checkstyleMain` and `./gradlew format` to fix style error.

## Related Issues
Fixed #42905 

## Type of Change
- [X] Fix bug

## Impact
Now when class `DockerRegistryUserAuthentication` and `DockerRegistryTokenAuthentication` use method `createAuthHeader()` would't add `"authHeader" : null` field.

## Test
I've already test the class `DockerRegistryUserAuthentication` and `DockerRegistryTokenAuthentication` method `getAuthHeader()`, but only write rough test code not create a new branch to test. Do I need create a new branch?

## Effect of Test
image
Use decode tools to display the original string.

### Expected Result:

```
{
  "username" : "Ben",
  "password" : "123455",
  "serveraddress" : "docker@example.com",
  "email" : "https://docker.example.com"
}
```
it works as expected.

## Note
If anything I did worked not as expected, plz contact with me. Finally, looking forward to your reply.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/configuration/JsonEncodedDockerRegistryAuthentication.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/configuration/JsonEncodedDockerRegistryAuthentication.java
@@ -23,6 +23,7 @@ import com.fasterxml.jackson.core.JsonProcessingException;
 import org.springframework.boot.buildpack.platform.json.SharedObjectMapper;
 
 import com.fasterxml.jackson.annotation.JsonIgnore;
+
 /**
  * {@link DockerRegistryAuthentication} that uses a Base64 encoded auth header value based
  * on the JSON created from the instance.
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/configuration/JsonEncodedDockerRegistryAuthentication.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/configuration/JsonEncodedDockerRegistryAuthentication.java
@@ -18,11 +18,10 @@ package org.springframework.boot.buildpack.platform.docker.configuration;
 
 import java.util.Base64;
 
+import com.fasterxml.jackson.annotation.JsonIgnore;
 import com.fasterxml.jackson.core.JsonProcessingException;
 
 import org.springframework.boot.buildpack.platform.json.SharedObjectMapper;
-
-import com.fasterxml.jackson.annotation.JsonIgnore;
 
 /**
  * {@link DockerRegistryAuthentication} that uses a Base64 encoded auth header value based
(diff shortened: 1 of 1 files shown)


snicoll:
@1328032567 thank you for making your first contribution to Spring Boot. To validate the auth header would not be included, I've changed the assertion of the "full" sample to be strict. See [d4010d3](https://github.com/spring-projects/spring-boot/commit/d4010d3be041016efb360d7dc8f672b06be43787).

1328032567:
> @1328032567 thank you for making your first contribution to Spring Boot. To validate the auth header would not be included, I've changed the assertion of the "full" sample to be strict. See [d4010d3](https://github.com/spring-projects/spring-boot/commit/d4010d3be041016efb360d7dc8f672b06be43787).

Alright, though you did't merge my branch(what a pity) but also used my modification. And finally I understand the true meaning of `DockerRegistryUserAuthenticationTests` and `DockerRegistryTokenAuthenticationTests` 🤣. And you change the test assert more strict, is that necessary? 🤔and why need to do this, could you plz explain it for me? Appreciate so much🥺.

snicoll:
>  though you did't merge my branch(what a pity) but also used my modification.

I did merge it, see [351018e](https://github.com/spring-projects/spring-boot/commit/351018ea65ad549b7b11b0d41978e2be0461e2bd). If you're worried about this not being recognized as a contribution of yours, See the tags next to your name on this page. 

> the test assert more strict, is that necessary

Nothing was validating that the auth header is gone. By using a strict assertion (`true` instead of `false`) the test will fail if there is an extra attribute present.

1328032567:
> > though you did't merge my branch(what a pity) but also used my modification.
> 
> I did merge it, see [351018e](https://github.com/spring-projects/spring-boot/commit/351018ea65ad549b7b11b0d41978e2be0461e2bd). If you're worried about this not being recognized as a contribution of yours, See the tags next to your name on this page.
> 
> > the test assert more strict, is that necessary
> 
> Nothing was validating that the auth header is gone. By using a strict assertion (`true` instead of `false`) the test will fail if there is an extra attribute present.

oh, I try to look through the doc for `JSONAssert()` api, and figure out the `null` attribution would also be valid through flag `false`, so we need to modify the flag into `true`. Thanks for your explanation🤩, Im very grateful.
Finally, about contribution, thanks for merging my PR #42910, Im overjoyed🥳.

PR #42918, opened 2024-10-29
Link to Eclipse setup instructions

ahoehma:
Just for faster finding how to setup eclipse to contribute to spring boot. Also the current "see below" make not so much sense because there is no eclipse section on this page.


mhalbritter:
Thanks @ahoehma !

PR #42919, opened 2024-10-29
Avoid calling `Task.getProject()` from a task action at execution time

quaff:
Fix
```
> Task :spring-boot-project:spring-boot-tools:spring-boot-maven-plugin:formatHelpMojoSource
Invocation of Task.project at execution time has been deprecated. This will fail with an error in Gradle 9.0. Consult the upgrading guide for further information: https://docs.gradle.org/8.10.2/userguide/upgrading_version_7.html#task_project
```


wilkinsona:
Duplicates #32980.

Issue #42871, opened 2024-10-24
build-info doesn't support seconds since the epoch from project.build.outputTimestamp


ahoehma:
I'm trying to enable https://maven.apache.org/guides/mini/guide-reproducible-builds.html for my springboot project.

I'm using maven-git-version-extension. And found that springboot can't handle seconds,

```
Caused by: java.time.format.DateTimeParseException: Text '1729757074' could not be parsed at index 0
    at java.time.format.DateTimeFormatter.parseResolved0 (DateTimeFormatter.java:2052)
    at java.time.format.DateTimeFormatter.parse (DateTimeFormatter.java:1954)
    at java.time.Instant.parse (Instant.java:397)
    at org.springframework.boot.maven.BuildInfoMojo.getBuildTime (BuildInfoMojo.java:157)
    at org.springframework.boot.maven.BuildInfoMojo.getProjectDetails (BuildInfoMojo.java:131)
    at org.springframework.boot.maven.BuildInfoMojo.execute (BuildInfoMojo.java:114)
```

See for details https://github.com/qoomon/maven-git-versioning-extension/issues/340.

https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/main/java/org/springframework/boot/maven/BuildInfoMojo.java#L149

A alternative round of parsing for the epoch seconds would be an improvement.

wilkinsona:
We [document `time`](https://docs.spring.io/spring-boot/maven-plugin/build-info.html#build-info.build-info-goal.parameter-details.time) as "the value used for the build.time property in a form suitable for Instant#parse(CharSequence)" so, on the one hand this is working as designed. On the other hand, we [document](https://docs.spring.io/spring-boot/maven-plugin/build-info.html#build-info.build-info-goal.optional-parameters) that it defaults to `${project.build.outputTimestamp}` and the [spec](https://cwiki.apache.org/confluence/pages/viewpage.action?pageId=74682318#Reproducible/VerifiableBuilds-OutputArchiveEntriesTimestamp) isn't clear that ISO-8601 is the only valid format as it says "with an ISO 8601 formatted date+time" in one place but also shows an example that supports both ISO 8601 and seconds since the epoch. On balance, I think it's worth us adding a fallback to try seconds since the epoch if parsing as ISO 8601 fails.

ahoehma:
I would like to provide a fix for this? If this is okay for you? I already forked springboot.

[...]

PR #42922, opened 2024-10-29
build-info doesn't support seconds since the epoch from project.build.outputTimestamp

ahoehma:
Use MavenBuildOutputTimestamp inside BuildInfoMojo to create Instant from project.build.outputTimestamp where the value of this property is in just epoch seconds.

#42871


snicoll:
@ahoehma thanks and congrats for making you first contribution to Spring Boot.

Issue #42906, opened 2024-10-28
HikariCheckpointRestoreLifecycle - new connection opened immediately after stop()


deki:
Despite the open issue https://github.com/brettwooldridge/HikariCP/issues/2082 I had the feeling `HikariCheckpointRestoreLifecycle` would properly close all connections. It turned out though that a new connection is opened directly:

```
| 1730117809191 | 2024-10-28T12:16:49.191Z  INFO 8 --- [       Thread-0] o.s.boot.SpringApplication               : Started application in 7.355 seconds (process running for 7.877)                                                                                                                                                                                                                                                                                                                                                                                    |
| 1730117809192 | 2024-10-28T12:16:49.192Z DEBUG 8 --- [       Thread-0] o.s.b.a.ApplicationAvailabilityBean      : Application availability state LivenessState changed to CORRECT                                                                                                                                                                                                                                                                                                                                                                                     |
| 1730117809194 | 2024-10-28T12:16:49.194Z DEBUG 8 --- [       Thread-0] o.s.b.a.ApplicationAvailabilityBean      : Application availability state ReadinessState changed to ACCEPTING_TRAFFIC                                                                                                                                                                                                                                                                                                                                                                          |
| 1730117809194 | 2024-10-28T12:16:49.194Z  INFO 8 --- [       Thread-0] o.s.c.f.serverless.web.ServerlessMVC     : Application is started successfully.                                                                                                                                                                                                                                                                                                                                                                                                                |
| 1730117809198 | 2024-10-28T12:16:49.198Z DEBUG 8 --- [           main] o.s.c.support.DefaultLifecycleProcessor  : Stopping Spring-managed lifecycle beans before JVM checkpoint                                                                                                                                                                                                                                                                                                                                                                                       |
| 1730117809199 | 2024-10-28T12:16:49.199Z DEBUG 8 --- [           main] o.s.c.support.DefaultLifecycleProcessor  : Stopping beans in phase 2147483647                                                                                                                                                                                                                                                                                                                                                                                                                  |
| 1730117809199 | 2024-10-28T12:16:49.199Z DEBUG 8 --- [           main] o.s.c.support.DefaultLifecycleProcessor  : Bean 'applicationTaskExecutor' completed its stop procedure                                                                                                                                                                                                                                                                                                                                                                                         |
[...]

PR #42937, opened 2024-10-30
Improve logging for HikariCheckpointRestoreLifecycle

deki:
Adressing #42906, open to change it to an exception instead of the warning message as it won't work as expected without `allow-pool-suspension=true`.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/HikariCheckpointRestoreLifecycle.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/HikariCheckpointRestoreLifecycle.java
@@ -115,7 +115,7 @@ public class HikariCheckpointRestoreLifecycle implements Lifecycle {
 	private void closeConnections(Duration shutdownTimeout) {
 		logger.info("Evicting Hikari connections");
 		this.dataSource.getHikariPoolMXBean().softEvictConnections();
-		logger.debug("Waiting for Hikari connections to be closed");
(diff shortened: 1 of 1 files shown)


sdeleuze:
Will the warning be also printed when doing a regular shutdown of the application?

wilkinsona:
Yes, I believe so. If `org.crac.Resource` is on the classpath then `HikariCheckpointRestoreLifecycle` will be auto-configured. The log message will then be output irrespective of the type of shutdown that's being performed.

christophstrobl:
Adding the shutdown time is a nice addition, though I think it should be formatted with a reasonable time unit. 
Not so sure about the warning log message for suspension. Thought the requirement for pool suspension had been documented/is easy to find.

deki:
I don't see it in https://docs.spring.io/spring-boot/reference/packaging/checkpoint-restore.html, where could I have read about it?
I think as long as crac is on classpath, users will consider using checkpoints at some point and the warning message would point them in the right direction.

christophstrobl:
the link in the docs seem to be broken.
> ... resources such as socket, files and thread pools [on a limited scope](https://github.com/spring-projects/spring-lifecycle-smoke-tests/blob/main/STATUS.adoc).

...so the [info](https://github.com/spring-projects/spring-lifecycle-smoke-tests/tree/main/data/data-jpa#jvm-checkpoint-restore) is not as easy to be found as I had assumed.

wilkinsona:
https://github.com/spring-projects/spring-boot/issues/42938 should have fixed the link to the status page.

PR #42951, opened 2024-10-31
Use Map.computeIfAbsent where feasible

quaff:


philwebb:
Thanks for the PR, but I've had some issues in the past using `computeIfAbsent` with a `ConcurrentHashMap` so I'd rather not risk changing the code purely for cosmetic reasons.

Issue #42807, opened 2024-10-21
Support timeout property for GraphQL over SSE


rstoyanchev:
It is now possible to configure a timeout on `GraphQlSseHandler`s after https://github.com/spring-projects/spring-graphql/issues/1079. This can be exposed as a property in Boot with some default value.

philwebb:
Closing in favor of PR #42966. Thanks @nosan!

PR #42966, opened 2024-11-01
Support timeout property for GraphQL over SSE

nosan:
#42807


philwebb:
Might try and squeeze this into 3.4 since it's low risk

philwebb:
Reopening because I missed the comment about a default timeout

philwebb:
Actually, I think `null` is the best option for the default timeout.

Issue #42970, opened 2024-11-01
Spring Boot 3.4.0-RC1 Configuration Issue with Multiple Datasources and EntityManagers


rajadilipkolli:
I have a Spring Boot application that utilizes both Liquibase and Flyway for schema management, connecting to PostgreSQL and MySQL databases. I recently upgraded to Spring Boot 3.4.0-RC1 and it is working as expected with [my code](https://github.com/[rajadilipkolli-throwaway/multi-datasources](https://github.com/rajadilipkolli-throwaway/multi-datasources)). 

I have encountered below issue when attempting to simplify the configuration using the [new approach](https://docs.spring.io/spring-boot/3.4-SNAPSHOT/how-to/data-access.html#howto.data-access.configure-two-datasources).

```
Cannot resolve reference to bean 'jpaSharedEM_entityManagerFactory' while setting bean property 'entityManager'
```

**Steps to Reproduce** :

1. Download code from the branch [qualifier](https://github.com/rajadilipkolli-throwaway/multi-datasources/tree/qualifier)
2. Bring Postgresql and Mysql database from `docker-compose.yml`
3. Application should start succesfully.

Changes made to existing repository can be seen in this [PR](https://github.com/rajadilipkolli-throwaway/multi-datasources/pull/1/files)

 
**Expected Behavior**:

The application should start successfully using new approach and use Liquibase and Flyway to manage the schema for each database.

rajadilipkolli:
Hi @wilkinsona, Thanks for the fix.

I have updated my sample to use snapshot and the earlier reported issue is fixed. But when schema validation is performed. spring boot default `entityManagerFactory` is considering entities from other entityManagerFactory which is `cardHolderEntityManagerFactory` in the sample.

error
```
2024-11-01T18:04:32.471Z ERROR 2125 --- [boot-data-multipledatasources] [           main] j.LocalContainerEntityManagerFactoryBean : Failed to initialize JPA EntityManagerFactory: [PersistenceUnit: default] Unable to build Hibernate SessionFactory; nested exception is org.hibernate.tool.schema.spi.SchemaManagementException: Schema-validation: missing table [card_holder]
[...]

PR #42913, opened 2024-10-29
Add virtual thread option for `ThreadPoolTaskExecutorBuilder`/`ThreadPoolTaskSchedulerBuilder`

quaff:
See https://github.com/spring-projects/spring-framework/issues/33807

PR #42978, opened 2024-11-04
ContainerConnectionDetails should inherit qualifiers from Container bean

quaff:
See https://github.com/spring-projects/spring-boot/issues/42970#issuecomment-2453759784


wilkinsona:
Thanks, @quaff, but it feels like we're rushing [again](https://github.com/spring-projects/spring-boot/pull/42913). I don't think we can take this approach as it may not be desirable for the container beans to be qualified. It also doesn't consider the case where `@ServiceConnection` is used on a field. I've opened https://github.com/spring-projects/spring-boot/issues/42979 so that we can do some design work in due course.

Issue #42985, opened 2024-11-04
Logback logging system does not process URLs with paths not ending in .xml


philwebb:
See #42964. We're going to remove the XML check restriction.

philwebb:
Closing in favor of PR #42986. Thanks @nosan!

PR #42986, opened 2024-11-04
Logback logging system does not process URLs with paths not ending in .xml

nosan:
#42985

3.2.x target branch


philwebb:
That was quick! Thanks @nosan

nosan:
Thanks @philwebb

Issue #42957, opened 2024-10-31
Warning due to duplicate MockResolver extensions


wilkinsona:
```
[WARNING] Found duplicate and different resources in [org.springframework.boot:spring-boot-test:3.4.0-20241031.110032-485, org.springframework:spring-test:6.2.0-RC3]:   mockito-extensions/org.mockito.plugins.MockResolver
```

Following https://github.com/spring-projects/spring-framework/commit/d8a6423c0cbb0aa717fa5f2762187c6ce860bca0, we should remove Boot's custom `MockResolver` and rely on Framework's instead.

philwebb:
Closing in favor of PR #42988. Thanks @nosan!

PR #42988, opened 2024-11-04
Warning due to duplicate MockResolver extensions

nosan:
#42957

I hope I’ve understood this issue correctly, but if not, please feel free to close this PR.


snicoll:
Thanks @nosan. We can't remove the class itself as it is public API. Rather, we should stop registering by default as we used to be. I did that in https://github.com/spring-projects/spring-boot/commit/606709ab8b071a3331f367c5a4c72619c9a96fa2

nosan:
Thanks @snicoll 
I had the same feeling that I could not touch this public class. However, I eventually decided to drop it as there’s a minimal chance that it's being used outside of Spring Boot itself. If anyone is using it, they can switch to `SpringMockResolver` as an alternative.

Anyway, your solution is better and safer than mine. Thank you.

PR #43065, opened 2024-11-08
Replace scope strings with constants

quaff:
And replace `"session"` with `WebApplicationContext.SCOPE_SESSION`


wilkinsona:
Thanks, @quaff. As part of merging this, I switched from the constants on `BeanDefinition` to those on `ConfigurableBeanFactory` as the latter is what the javadoc of `@Scope` recommends.

Issue #42530, opened 2024-10-09
Add a capability to trigger a quartz job on-demand via the actuator endpoint


mranile-dev:
Currently, SB is having actuator endpoints to retrieve details of Quartz scheduled jobs and its triggers. This feature request is to extend the actuators capability to be able to do the following:

1. On-demand trigger of any Quartz job(s)
2. Ability to override or update the existing cron expression of the specific job(s)

philwebb:
We think that 1) is possible, and hopefully not too hard to implement. 2) is pretty hard because Quartz supports multiple trigger types.

nosan:
I've worked a little bit on this issue, especially the trigger job on demand part. I've prototyped some changes in my [branch](https://github.com/spring-projects/spring-boot/compare/main...nosan:spring-boot:42530).

To trigger a job on demand, overall does not require a lot of changes and can be implemented, but it requires some changes in `AbstractWebFluxEndpointHandlerMapping` and `AbstractWebMvcEndpointHandlerMapping` since both of them do not support `Map<String, Object>` as an endpoint parameter. (Consider as a bug? By the way, **Jersey** works fine)

To support `Map<String, Object>` as an endpoint parameter the following should be changed
in both classes:

`@RequestBody(required = false) Map<String, String> body` should be changed to `@RequestBody(required = false) Map<String, Object> body`, otherwise the following exception will be thrown:

```
[org.springframework.http.converter.HttpMessageNotReadableException: 
JSON parse error: Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)]
```

Regarding the ability to override or update the existing trigger of the specific job. I am
not sure whether it is worth supporting this at all. For example, I checked the `QuartzSchedulerMBean` interface and it does not have such functionality for JMX. Overall, it could be implemented, but it requires a quite significant piece of work.

mranile-dev:
Let's implement capability #1 for now if it's doable. It will be a good feature for remotely triggering the job on-demand without waiting on the next scheduled trigger. Thanks @nosan Cheers!

nosan:
> Overall, it could be implemented, but it requires quite a significant piece of work.

Given the limitation that `@Endpoint` cannot accept complex types in the body, it seems impossible to achieve. When writing my initial comment, I assumed `@Endpoint` supported such types, but it turns out it does not.  (https://github.com/spring-projects/spring-boot/issues/43209)


[...]

PR #43086, opened 2024-11-10
Add the ability to trigger a Quartz job on-demand through an Actuator endpoint

nosan:
Before this commit, triggering a Quartz job on demand was not possible. This commit introduces a new `@WriteOperation` endpoint at `/actuator/quartz/jobs/{groupName}/{jobName}/trigger`, allowing a job to be triggered by specifying the `{jobName}`, `{groupName}`, and an optional `JobDataMap`.

See gh-42530

```bash

$ curl 'http://localhost:8080/actuator/quartz/jobs/samples/jobOne/trigger' -i -X POST \
    -H 'Content-Type: application/json' \
    -d '{"jobData":{"key":"value","keyN":"valueN"}}'
    
```

```http

HTTP/1.1 200 OK
Content-Type: application/vnd.spring-boot.actuator.v3+json
Content-Length: 166

{
  "group" : "samples",
  "name" : "jobOne",
  "className" : "org.springframework.scheduling.quartz.DelegatingJob",
  "triggerTime" : "2024-11-10T09:44:05.425233Z"
}

```

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/quartz/QuartzEndpointWebExtension.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/quartz/QuartzEndpointWebExtension.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -27,6 +27,7 @@ import org.springframework.boot.actuate.endpoint.SecurityContext;
 import org.springframework.boot.actuate.endpoint.Show;
 import org.springframework.boot.actuate.endpoint.annotation.ReadOperation;
 import org.springframework.boot.actuate.endpoint.annotation.Selector;
+import org.springframework.boot.actuate.endpoint.annotation.WriteOperation;
 import org.springframework.boot.actuate.endpoint.web.WebEndpointResponse;
 import org.springframework.boot.actuate.endpoint.web.annotation.EndpointWebExtension;
 import org.springframework.boot.actuate.quartz.QuartzEndpoint.QuartzGroupsDescriptor;
@@ -77,6 +78,18 @@ public class QuartzEndpointWebExtension {
 		boolean showUnsanitized = this.showValues.isShown(securityContext, this.roles);
 		return handle(jobsOrTriggers, () -> this.delegate.quartzJob(group, name, showUnsanitized),
 				() -> this.delegate.quartzTrigger(group, name, showUnsanitized));
+	}
+
+	@WriteOperation
+	public WebEndpointResponse<Object> triggerQuartzJob(@Selector String jobs, @Selector String group,
+			@Selector String name, String state) throws SchedulerException {
+		if (!"jobs".equals(jobs)) {
+			return new WebEndpointResponse<>(WebEndpointResponse.STATUS_BAD_REQUEST);
+		}
+		if (!"running".equals(state)) {
+			return new WebEndpointResponse<>(WebEndpointResponse.STATUS_BAD_REQUEST);
+		}
+		return handleNull(this.delegate.triggerQuartzJob(group, name));
 	}
 
 	private <T> WebEndpointResponse<T> handle(String jobsOrTriggers, ResponseSupplier<T> jobAction,
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/docs/antora/modules/api/pages/rest/actuator/quartz.adoc
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/docs/antora/modules/api/pages/rest/actuator/quartz.adoc
@@ -156,6 +156,39 @@ The following table describes the structure of the response:
 include::partial$rest/actuator/quartz/job-details/response-fields.adoc[]
 
 
+[[quartz.trigger-job]]
+== Trigger Quartz Job On Demand
+
+To trigger a particular Quartz job, make a `POST` request to `/actuator/quartz/jobs/\{groupName}/\{jobName}`, as shown in the following curl-based example:
+
+include::partial$rest/actuator/quartz/trigger-job/curl-request.adoc[]
+
+The preceding example demonstrates how to trigger a job that belongs to the `samples` group and is named `jobOne`.
+
+The response will look similar to the following:
+
+include::partial$rest/actuator/quartz/trigger-job/http-response.adoc[]
+
+
+[[quartz.trigger-job.request-structure]]
+=== Request Structure
+
+The request specifies a desired `state` associated with a particular job.
+Sending an HTTP Request with a  `"state": "running"` body indicates that the job should be run now.
+The following table describes the structure of the request:
+
+[cols="2,1,3"]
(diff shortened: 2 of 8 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/endpoint/web/reactive/AbstractWebFluxEndpointHandlerMapping.java
```diff
@@ -298,7 +298,7 @@ protected interface LinksHandler {
 	@FunctionalInterface
 	protected interface ReactiveWebOperation {
 
-		Mono<ResponseEntity<Object>> handle(ServerWebExchange exchange, Map<String, String> body);
+		Mono<ResponseEntity<Object>> handle(ServerWebExchange exchange, Map<String, Object> body);
```
snicoll, line 301: Can you explain why you had to change this?

nosan: I added a comment regarding these changes in the https://github.com/spring-projects/spring-boot/issues/42530#issuecomment-2457605930

The Map<String, String> request body does not support complex values like Map or List, only basic JSON types.

For example, this JSON is not supported at the moment:

```json
{
  "jobData": {
    "key": "value"
  }
}

```
due to
```
[org.springframework.http.converter.HttpMessageNotReadableException: 
JSON parse error: Cannot deserialize value of type `java.lang.String` from Object value (token `JsonToken.START_OBJECT`)]
```

nosan: I meant that `@WebEndpointTest` with `JERSEY` infrastructure does support `Map<String, Object` as an endpoint parameter. 

If I revert my changes related to `@RequestBody Map<String, Object` the MVC and WebFlux will fail but Jersey will not. 

You can reproduce this issue by running this test `QuartzEndpointWebIntegrationTests.quartzTriggerJobWithCustomData`

Branch:  https://github.com/nosan/spring-boot/tree/42530-map-issue

snicoll: Thanks @nosan, I've created #43209 and reproduced the problem.

nosan:
Thanks, @snicoll
PR has been updated.

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/quartz/QuartzEndpointWebExtension.java
```diff
@@ -79,6 +80,18 @@ public WebEndpointResponse<Object> quartzJobOrTrigger(SecurityContext securityCo
 				() -> this.delegate.quartzTrigger(group, name, showUnsanitized));
 	}
 
+	@WriteOperation
+	public WebEndpointResponse<Object> triggerQuartzJob(@Selector String jobs, @Selector String group,
```
snicoll, line 84: We're not fan of this approach and we don't want the action to be part of the URL. We've brainstormed quite a bit and we believe something else is missing that I am going to tackle in #43226. 

Rather than this, we'd like an approach where a `POST` on the job detail could trigger a new execution if the payload contains `"running" : true`. Can you please review the PR in that direction? It doesn't need the related issue but we'll process them both once they're ready to be merged.

snicoll (changes requested):

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/quartz/QuartzEndpointWebExtension.java
nosan: Thanks, @snicoll.

Initially, I was thinking of including `action=TRIGGER|RUN|EXECUTE` as part of the request body.

I have one more question regarding `running:true`. What should be returned if someone provides `running:false`? Should it result in a bad request or a not-found response?

nosan: I was thinking about `action` because quartz supports `pause`, `resume` actions a well.

nosan: This feature might be requested in the future. For instance, if a job is broken or encountering issues, someone might want the ability to pause it.

wilkinsona: The idea is that you `POST` a body indicating the desired state. For a job, posting `running: true` indicates that the job should be run now. AFAIK, Quartz doesn't support cancellation of a running job so posting `running:false` would be a bad request.

AFAIK, pausing is supported by triggers not jobs so I'd expect that support to be added to a resource that represents a trigger. You'd post something like `"state": "paused"` to pause a trigger and `"state": "normal"` to resume it.

nosan: Apologies, I misunderstood the `running: false/true part`. You intend to include this information in a detailed response to inform users that the job can be triggered using the **POST** method `/actuator/quartz/jobs/{groupName}/{jobName}`.
Am I right?

nosan: > AFAIK, pausing is supported by triggers not jobs so I'd expect that support to be added to a resource that represents a trigger. You'd post something like "state": "paused" to pause a trigger and "state": "normal" to resume it.
 
 
 ```java
 
     /**
     * Pause the <code>{@link org.quartz.JobDetail}</code> with the given
     * key - by pausing all of its current <code>Trigger</code>s.
     * 
     * @see #resumeJob(JobKey)
     */
    void pauseJob(JobKey jobKey)
        throws SchedulerException;
        
            /**
     * Resume (un-pause) the <code>{@link org.quartz.JobDetail}</code> with
     * the given key.
     * 
     * <p>
     * If any of the <code>Job</code>'s<code>Trigger</code> s missed one
     * or more fire-times, then the <code>Trigger</code>'s misfire
     * instruction will be applied.
     * </p>
     * 
     * @see #pauseJob(JobKey)
     */
    void resumeJob(JobKey jobKey)
        throws SchedulerException;
        
 ```

nosan: To be honest, I do like `"state": "running" ` more than  `"running": "true" `

nosan:
Thank you, @wilkinsona and @snicoll 

The PR has been updated. I've implemented the `"state": "running"` approach, as I believe it will be easier to extend in the future. However, if you think `"running": true` is a better option, please let me know, and I’ll update the PR accordingly.

philwebb:
Thanks very much @nosan. I made a small polish commit to remove the tests that call the endpoint with different API versions. I think since, we're not returning different content, they're probably not necessary.

nosan:
Thank you, @philwebb

PR #43123, opened 2024-11-12
Update "Upgrading From" section to use "2.x"

izeye:
This PR updates "Upgrading From" section to use "2.x".


mhalbritter:
Thanks @izeye !

PR #43176, opened 2024-11-15
Tighten rules around profile naming

YangSiJun528:
I added the `validateProfiles()` method to enforce stricter rules for Spring profile names. 

Currently, only alphanumeric characters, hyphens (`-`), and underscores (`_`) are allowed. Please let me know if additional characters should be permitted - I'll update accordingly.

I've verified validation works in `application.properties` and `@ActiveProfiles`. 

While `@Profile` doesn't perform validation, this seems acceptable since invalid profiles will fail during activation. Perhaps we could explore adding warnings as a potential enhancement in the future.

Note: this change may break applications using non-conforming profile names.

Fixes gh-34062

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/StandardConfigDataLocationResolver.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/StandardConfigDataLocationResolver.java
@@ -166,18 +166,24 @@ public class StandardConfigDataLocationResolver
 	}
 
 	private void validateProfiles(Profiles profiles) {
-		Pattern validProfilePattern = Pattern.compile("^[a-zA-Z0-9_\\-]+$");
-		Assert.notNull(profiles, "Profiles must not be null");
 		for (String profile : profiles) {
-			Assert.notNull(profile, "Profile must not be null");
-			Assert.hasText(profile, "Profile must not be empty");
-			Assert.state(!profile.startsWith("-") && !profile.startsWith("_"),
-					() -> String.format("Invalid profile '%s': must not start with '-' or '_'", profile));
-			Assert.state(!profile.endsWith("-") && !profile.endsWith("_"),
-					() -> String.format("Invalid profile '%s': must not end with '-' or '_'", profile));
-			Assert.state(validProfilePattern.matcher(profile).matches(), () -> String
-				.format("Invalid profile '%s': must only contain alphanumeric characters, '-', or '_'", profile));
+			validateProfile(profile);
 		}
+	}
+
+	private void validateProfile(String profile) {
+		Assert.notNull(profile, "Profile must not be null");
+		Assert.hasText(profile, "Profile must not be empty");
+		Assert.state(!profile.startsWith("-") && !profile.startsWith("_"),
+				() -> String.format("Invalid profile '%s': must not start with '-' or '_'", profile));
+		Assert.state(!profile.endsWith("-") && !profile.endsWith("_"),
+				() -> String.format("Invalid profile '%s': must not end with '-' or '_'", profile));
+		profile.codePoints().forEach((codePoint) -> {
+			if (codePoint == '-' || codePoint == '_' || Character.isLetterOrDigit(codePoint)) {
+				return;
+			}
+			throw new IllegalStateException(String.format("Invalid profile '%s': must contain only letters or digits or '-' or '_'", profile));
+		});
 	}
 
 	private String getResourceLocation(ConfigDataLocationResolverContext context,
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/StandardConfigDataLocationResolver.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/StandardConfigDataLocationResolver.java
@@ -54,6 +54,7 @@ import org.springframework.util.StringUtils;
  * @author Madhura Bhave
  * @author Phillip Webb
  * @author Scott Frederick
+ * @author Sijun Yang
  * @since 2.4.0
  */
 public class StandardConfigDataLocationResolver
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/StandardConfigDataLocationResolver.java
```diff
@@ -163,6 +165,21 @@ private Set<StandardConfigDataReference> getProfileSpecificReferences(ConfigData
 		return references;
 	}
 
+	private void validateProfiles(Profiles profiles) {
+		Pattern validProfilePattern = Pattern.compile("^[a-zA-Z0-9_\\-]+$");
```
mhalbritter, line 169: This can be simplified to `Pattern validProfilePattern = Pattern.compile("^[\\w-]+$");`.

mhalbritter:
Hello!

This currently breaks: 

* `org.springframework.boot.test.context.SpringBootContextLoaderTests#activeProfileWithComma`
* `org.springframework.boot.test.context.SpringBootContextLoaderTests#testPropertyValuesShouldTakePrecedenceWhenInlinedPropertiesPresentAndProfilesActive`

I think we can remove `activeProfileWithComma` and use a profile without a comma in `testPropertyValuesShouldTakePrecedenceWhenInlinedPropertiesPresentAndProfilesActive`.

It also breaks:

* `org.springframework.boot.SpringApplicationTests#logsActiveProfilesWithoutProfileAndMultipleDefaults`: Here we can remove the comma.
* `org.springframework.boot.SpringApplicationTests#logsActiveProfilesWithMultipleProfiles`: Same here.

mhalbritter:
Flagging this for team meeting as I want to talk about if the positive list is a good approach. Right now, it's very western centric by allowing only A-Z and digits.

I wonder if we should approach it like this:

```java
private void validateProfiles(Profiles profiles) {
		for (String profile : profiles) {
			validateProfile(profile);
		}
	}

	private void validateProfile(String profile) {
		Assert.notNull(profile, "Profile must not be null");
		Assert.hasText(profile, "Profile must not be empty");
		Assert.state(!profile.startsWith("-") && !profile.startsWith("_"),
				() -> String.format("Invalid profile '%s': must not start with '-' or '_'", profile));
		Assert.state(!profile.endsWith("-") && !profile.endsWith("_"),
				() -> String.format("Invalid profile '%s': must not end with '-' or '_'", profile));
		for (int i = 0; i < profile.length(); i++) {
			char c = profile.charAt(i);
			if (c == '-' || c == '_') {
				continue;
			}
			if (Character.isLetterOrDigit(c)) {
				continue;
			}
			throw new IllegalStateException(String.format("Invalid profile '%s': must contain only letters or digits or '-' or '_'", profile));
		}
	}
```

This would also block `,`, `(` etc. but lets more non-western letters and digits through.

YangSiJun528:
I incorporated the feedback provided. @mhalbritter
- Fixed the broken tests
- Fixed to allow non-ascii letters

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/config/StandardConfigDataLocationResolver.java
```diff
@@ -163,6 +165,27 @@ private Set<StandardConfigDataReference> getProfileSpecificReferences(ConfigData
...
 	}
 
+	private void validateProfiles(Profiles profiles) {
+		for (String profile : profiles) {
+			validateProfile(profile);
+		}
+	}
+
+	private void validateProfile(String profile) {
+		Assert.notNull(profile, "Profile must not be null");
```
akashpawar6474, line 175: maybe we can combine null and empty checks in the single line with `Assert.state(!profile.isNullOrBlank())`

philwebb: Java has no such `isNullOrBlank` method, however, `Assert.hasText(...)` deals with `null` and empty strings.

YangSiJun528: I removed `Assert.notNull()` and updated error message in `Assert.hasText()`.

YangSiJun528:
To test empty string and `null` cases for active profile, I used `mock(Environment.class)` because `AbstractEnvironment#setActiveProfiles()` blocks direct testing. However, this approach is challenging because `Profiles` is used in many places.

Is there a better way to handle this testing?

I've included example tests in a demo branch (f4ed4d1490fdb085062aff200dba13a6250c57c0).

YangSiJun528:
@mhalbritter 
Gentle ping. 😉

mhalbritter:
@YangSiJun528 We need to wait until the `main` branch has been switched to `3.5.x`. Right now, it's on `3.4.x` and this PR is targeted for `3.5.x`. Hang in there, your contribution won't be lost :)

YangSiJun528:
I understood. Thank you for letting me know.

mhalbritter:
Thanks @YangSiJun528 !

leibf2099:
it is a edundant and confusing restriction，why can't there be brackets or other letters?

bclozel:
@leibf2099 the rationale is explained in the relevant issue https://github.com/spring-projects/spring-boot/issues/34062

The short version is: it can break the parsing of properties.

leibf2099:
The StandardConfigDataLocationResolver class or validateProfiles method should can be overridden. I want profiles to contain brackets, plus signs, equal signs, etc.

iparadiso:
We recently ran into this issue as we have apps that rely on `.` in a spinnaker stack that we translate into an active profile.  

IMO, these kinds of breakages should first be released with a warning log and then break in 3.6.  This approach has been used many times before for alerting that breaking changes are coming before disrupting users.

iparadiso:
@philwebb based on my comment above, is there any recommended back-port workarounds of overriding `StandardConfigDataLocationResolver ` to restore the previous behavior while we find an way to migrate users off of this pattern?

philwebb:
@iparadiso Given the feedback in #45947 this is clearly more of an issue than we anticipated it would be. I'm going to use that issue investigate options.

Issue #43173, opened 2024-11-15
Add TaskDecorator support for scheduled tasks


mhalbritter:
Spring Framework has added `TaskDecorator` support in https://github.com/spring-projects/spring-framework/issues/23755.

Right now, we auto-configure `TaskDecorator` for task execution, but not for task scheduling. I think we should also support `TaskDecorator` beans for task scheduling.

snicoll:
Closing in favor of PR #43190

PR #43190, opened 2024-11-15
Add TaskDecorator support for scheduled tasks

nosan:
gh-43173

I also marked the existing `ThreadPoolTaskSchedulerBuilder` constructor with the `@Deprecated` annotation, as I believe the **public** modifier was added unintentionally.

I set the version to `since 3.5.0`, but if possible, could this be merged into `3.4.0` instead? Of course, if you have time.


snicoll:
Thanks @nosan. We're post RC now so we won't merge additional enhancements.

snicoll:
This PR has been reviewed and is ready to go once we branch. @nosan please disregard the conflict, I changed something on `main` on purpose and rebased in my polish of this PR.

PR #43203, opened 2024-11-16
Classes are accidentally named "structure logging" instead of "structured logging"

izeye:
This PR replaces "structure logging" with "structured logging" as it seems to be a typo of "structured logging".


philwebb:
Nice spot. Thanks for finding this before GA!

PR #43204, opened 2024-11-16
add Local zipkin url to application.yml

NatnaelSisay:
I followed the tutorial page, and while it mentions running Zipkin, it doesn't explain how to connect to it using the local URL.


snicoll:
What tutorial page did you follow @NatnaelSisay? The property you've added seems to come from Spring Cloud Sleuth which is no longer supported and not part of this project.

PR #43214, opened 2024-11-19
Prohibit unnecessary value on `@EnumSource`

quaff:


philwebb:
Oh very nice! I didn't realize you could do that.

PR #43235, opened 2024-11-21
Replace TestObservationRegistryAssert.assertThat() with Assertions.assertThat()

izeye:
This PR replaces `TestObservationRegistryAssert.assertThat()` with `Assertions.assertThat()`.

See https://github.com/micrometer-metrics/micrometer/pull/5551


philwebb:
Thanks @izeye

Issue #43275, opened 2024-11-23
Add a configuration property to customize the Tomcat connector's max parameter count


rafaello7:
Tomcat has a limit of HTML form parameter count, which is 10000 by default. Currently the limit change is possible in Spring Boot only programmatically. On the other hand, the Undertow embedded server has a property [server.undertow.max-parameters](https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.server.server.undertow.max-parameters). Could it be possible to have such property for Tomcat also?

wilkinsona:
Thanks for the suggestion, @rafaello7.

It looks like the proposed property would map onto `org.apache.catalina.connector.Connector#setMaxParameterCount(int)`.  There's also `org.eclipse.jetty.ee10.servlet.ServletContextHandler#setMaxFormKeys(int)` for Jetty that looks like it serves a very similar purpose. We may want to add a property for that as well.

@violetagg is there a similar setting in Reactor Netty?

violetagg:
@wilkinsona Reactor Netty does not parse the request line, this is done by Spring Framework's parser. Also I can see that this configuration [maxParameterCount](https://tomcat.apache.org/tomcat-11.0-doc/config/http.html#Common_Attributes) is also relevant to multipart. Reactor Netty multipart implementation is not used by Spring Framework as it uses its own.

```

maxParameterCount - The maximum total number of request parameters (including uploaded files) obtained from the query
string and, for POST requests, the request body if the content type is application/x-www-form-
urlencoded or multipart/form-data. Requests that exceed this limit will be rejected. A value of less than 0 means no limit. If
not specified, a default of 1000 is used.
```

wilkinsona:
Thanks, Violeta. So we'd keep this as a web server-specific setting.

wilkinsona:
Closing in favor of #43286.

PR #43286, opened 2024-11-26
Add a configuration property to customize the Tomcat connector's max parameter count

quaff:
Closes GH-43275

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -73,6 +73,7 @@ import org.springframework.util.unit.DataSize;
  * @author Florian Storz
  * @author Michael Weidmann
  * @author Lasse Wulff
+ * @author Yanming Zhou
  * @since 1.0.0
  */
 @ConfigurationProperties(prefix = "server", ignoreUnknownFields = true)
@@ -513,13 +514,11 @@ public class ServerProperties {
 		 */
 		private DataSize maxHttpResponseHeaderSize = DataSize.ofKilobytes(8);
 
-		public DataSize getMaxHttpFormPostSize() {
-			return this.maxHttpFormPostSize;
-		}
-
-		public void setMaxHttpFormPostSize(DataSize maxHttpFormPostSize) {
-			this.maxHttpFormPostSize = maxHttpFormPostSize;
-		}
+		/**
+		 * Maximum number of parameters (GET plus POST) that will be automatically parsed
+		 * by the container. A value of less than 0 means no limit.
+		 */
+		private int maxParameterCount = 10000;
 
 		public Accesslog getAccesslog() {
 			return this.accesslog;
@@ -667,6 +666,22 @@ public class ServerProperties {
 
 		public void setMaxHttpResponseHeaderSize(DataSize maxHttpResponseHeaderSize) {
 			this.maxHttpResponseHeaderSize = maxHttpResponseHeaderSize;
+		}
+
+		public DataSize getMaxHttpFormPostSize() {
+			return this.maxHttpFormPostSize;
+		}
+
+		public void setMaxHttpFormPostSize(DataSize maxHttpFormPostSize) {
+			this.maxHttpFormPostSize = maxHttpFormPostSize;
+		}
+
+		public int getMaxParameterCount() {
+			return this.maxParameterCount;
+		}
+
+		public void setMaxParameterCount(int maxParameterCount) {
+			this.maxParameterCount = maxParameterCount;
 		}
 
 		/**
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -513,13 +514,11 @@ public static class Tomcat {
...
-		}
-
-		public void setMaxHttpFormPostSize(DataSize maxHttpFormPostSize) {
-			this.maxHttpFormPostSize = maxHttpFormPostSize;
-		}
+		/**
+		 * Maximum number of parameters (GET plus POST) that will be automatically parsed
+		 * by the container.
+		 */
+		private Integer maxParameterCount;
```
wilkinsona, line 521: To align with [embedded Tomcat's own default](https://github.com/apache/tomcat/blob/b42637028c26a6012da1e106c3b1aeff623aeace/java/org/apache/catalina/connector/Connector.java#L213), this should be an `int` with a value of `10000`. We can then add a test to ensure that our default and Tomcat's default are in sync.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -513,13 +514,11 @@ public static class Tomcat {
...
-		public DataSize getMaxHttpFormPostSize() {
-			return this.maxHttpFormPostSize;
-		}
-
-		public void setMaxHttpFormPostSize(DataSize maxHttpFormPostSize) {
-			this.maxHttpFormPostSize = maxHttpFormPostSize;
-		}
+		/**
+		 * Maximum number of parameters (GET plus POST) that will be automatically parsed
+		 * by the container.
```
wilkinsona, line 519: I think it's worth documenting that the limit can be disabled by specifying a value that is less than zero.

wilkinsona (changes requested):
Thanks for the PR, @quaff. I've left a couple of comments for your consideration.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/embedded/TomcatWebServerFactoryCustomizerTests.java
```diff
@@ -194,6 +197,14 @@ void customMaxHttpRequestHeaderSize() {
 			.isEqualTo(DataSize.ofMegabytes(10).toBytes()));
 	}
 
+	@ParameterizedTest
+	@ValueSource(ints = { -1, 0, 100 })
```
quaff, line 201: Is `@ParameterizedTest` required here to test negative value? @wilkinsona

wilkinsona: I don't think so. Given that all three values are passed through and then returned by `getMaxParameterCount()` without any having special treatment, I think testing with a single value would be sufficient.

quaff: OK, I reverted this part.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
```diff
@@ -513,13 +514,11 @@ public static class Tomcat {
...
-		}
-
-		public void setMaxHttpFormPostSize(DataSize maxHttpFormPostSize) {
-			this.maxHttpFormPostSize = maxHttpFormPostSize;
-		}
+		/**
+		 * Maximum number of parameters (GET plus POST) that will be automatically parsed
+		 * by the container. A value of less than 0 means no limit.
+		 */
+		private int maxParameterCount = 10000;
```
akashpawar6474, line 521: I’m curious—could you explain where and how this is used for conditional checks? I’d love to learn more about it.

wilkinsona: What do you mean by "conditional checks"? You can review the other changes in this PR to see how this property's value will be used.

philwebb: That's not our code, Tomcat would use the value set. The `factory.addConnectorCustomizers((connector) -> connector.setMaxParameterCount(maxParameterCount));` line passes the value along.

PR #43301, opened 2024-11-27
Document how to use structured logging with custom log configuration

quaff:
I try to enable structured logging by configuring
```yaml
logging:
  structured.format.console: ecs
  structured.format.file: ecs
```
It doesn't works because there is `log4j2.xml` in classpath, I think it's worthy to add warning to document.

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/logging.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/logging.adoc
@@ -450,6 +450,34 @@ Spring Boot supports structured logging and has support for the following JSON f
 
 To enable structured logging, set the property configprop:logging.structured.format.console[] (for console output) or configprop:logging.structured.format.file[] (for file output) to the id of the format you want to use.
 
+If you are using xref:#features.logging.custom-log-configuration[Custom Log Configuration], update your configuration to respect `CONSOLE_LOG_STRUCTURED_FORMAT` and `FILE_LOG_STRUCTURED_FORMAT` system properties.
+Take `CONSOLE_LOG_STRUCTURED_FORMAT` for example:
+[tabs]
+======
+Logback::
++
+[source,xml]
+----
+<!-- replace your encoder with StructuredLogEncoder -->
+<encoder class="org.springframework.boot.logging.logback.StructuredLogEncoder">
+	<format>${CONSOLE_LOG_STRUCTURED_FORMAT}</format>
+	<charset>${CONSOLE_LOG_CHARSET}</charset>
+</encoder>
+----
+Log4j2::
++
+[source,xml]
+----
+<!-- replace your PatternLayout with StructuredLogLayout -->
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/logging.adoc
```diff
@@ -450,6 +450,7 @@ Spring Boot supports structured logging and has support for the following JSON f
 
 To enable structured logging, set the property configprop:logging.structured.format.console[] (for console output) or configprop:logging.structured.format.file[] (for file output) to the id of the format you want to use.
 
+WARNING: xref:#features.logging.custom-log-configuration[Custom Log Configuration] take precedence over structured logging.
```
wilkinsona, line 453: Thanks, @quaff. I wonder if people might miss this? They could read the paragraph about and then stop, happy that they know which configuration property to set. I think it might be more noticeable if we changed the paragraph above instead to start with something like "To enable structured logging, ensure that you aren't using custom log configuration and set the property…". WDYT?

quaff: I agree.

quaff: Actually, "Structured Logging" and "Custom Log Configuration" are not mutually exclusive, I updated the document. @wilkinsona

spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/logging.adoc
```diff
@@ -450,6 +450,35 @@ Spring Boot supports structured logging and has support for the following JSON f
...
+	<format>${CONSOLE_LOG_STRUCTURED_FORMAT}</format>
+	<charset>${CONSOLE_LOG_CHARSET}</charset>
+</encoder>
+----
+Log4j2::
++
+[source,xml]
+----
+<!-- wrap your PatternLayout with <Select> and <DefaultArbiter> -->
+<Select>
```
mhalbritter, line 472: I think we can skip the whole select and arbiter stuff here. The user wants to have structured logging, no need for conditionals. The logback section isn't conditional, too.

quaff: FYI, my actual case requires conditional.

wilkinsona: I agree with @mhalbritter. We shouldn't over-complicate the documentation by showing the use of conditionals if they aren't always needed.

quaff: OK, I added links to default configurations for fine-grained control.

mhalbritter:
Thanks @quaff !

PR #43310, opened 2024-11-28
Fix wrong document

quaff:
The javadoc of `StructuredLoggingJsonMembersCustomizer` doesn't mention that, and I'm failed to reproduce it, I believe it cannot be loaded from `spring.factories`.


quaff:
@wilkinsona #43312 created.

PR #43313, opened 2024-11-28
Update javadoc of StructuredLoggingJsonMembersCustomizer to note that implementations can registered through spring.factories

quaff:
See https://github.com/spring-projects/spring-boot/pull/43310#issuecomment-2505434497

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/structured/StructuredLoggingJsonMembersCustomizer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/structured/StructuredLoggingJsonMembersCustomizer.java
@@ -27,7 +27,9 @@ import org.springframework.core.env.Environment;
  * customize {@link JsonWriter} {@link Members}.
  * <p>
  * An implementation may be provided using the {@code logging.structured.json.customizer}
- * property.
+ * property. Alternatively, implementations can be registered in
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/structured/StructuredLoggingJsonMembersCustomizer.java
```diff
@@ -24,7 +24,8 @@
  * customize {@link JsonWriter} {@link Members}.
  * <p>
  * An implementation may be provided using the {@code logging.structured.json.customizer}
- * property.
+ * property. Alternatively, implementations may be listed in a
+ * {@code META-INF/spring.factories} file.
```
wilkinsona, line 28: Thanks, @quaff. It'd be nice to align the wording with other classes like `DatabaseInitializerDetector`. Something like:

```
Alternatively, implementations can be registered in {@code META-INF/spring.factories} under the key {@code org.springframework.boot.logging.structured.StructuredLoggingJsonMembersCustomizer}.
```

quaff: done.

wilkinsona:
Thanks, @quaff!

PR #43314, opened 2024-11-28
Document that StructuredLoggingJsonMembersCustomizer implementations may optionally take constructor parameters

quaff:
Similar to https://github.com/spring-projects/spring-boot/blob/f9aedf5a43e0b7d3282833c89d1a3c6611f7963c/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/env/EnvironmentPostProcessor.java#L37-L46


wilkinsona:
Thanks, @quaff. I polished things a bit to match the javadoc of `org.springframework.boot.logging.structured.StructuredLogFormatter` in the same package.

Issue #43318, opened 2024-11-28
Upgrade log4j to 2.24.2 to resolve broken logger initialization in 2.24.1


YanivKunda:
Log4j have release 2.24.2 solving https://github.com/apache/logging-log4j2/issues/3143 -
which can cause `LogManager.getLogger()` to return null under certain conditions.

The release notes for 2.24.1 (currently in use in spring-boot 3.4.0) was retrofitted with a warning encouraging the upgrade.

Log4j 2.x release notes: https://logging.apache.org/log4j/2.x/release-notes.html#release-notes-2-24-2

wilkinsona:
Thanks but we don't need an issue for a dependency upgrade such as this. As explained in the issue template that was displayed when you opened this issue, [we have some semi-automated tooling that will handle it](https://github.com/spring-projects/spring-boot/blob/f3a161afdde5cfb7a815c83b9f67704e5fc0afb3/.github/ISSUE_TEMPLATE/issue.md?plain=1#L20).

YanivKunda:
Sorry, it was my mistake not reading the template all the way through...
Thanks for the fail-fast response!

PR #43319, opened 2024-11-28
upgraded log4j to 2.24.2

YanivKunda:
Upgrades log4j to 2.24.2 - 
resolving https://github.com/spring-projects/spring-boot/issues/43318


wilkinsona:
Thanks, but we don't accept pull requests for one-line dependency upgrades as we have a semi-automated process that handles it for us before each release.

PR #43322, opened 2024-11-29
Polish formatting of log configuration file

quaff:
1. Remove space before `/>` to keep consistency.
2. Capitalize `filters` for Log4j2.
3. Uppercase log level to keep consistency.


quaff:
No sure which style is preferred by Spring Boot, but mixing different styles is not good.

wilkinsona:
Thanks, @quaff.

> Remove space before /> to keep consistency.

The `--format` support in `xmllint` removes them so that feels like as good a reason as any to go in that direction.

> Unwrap <Filters> for Log4j2

I'm not sure we should do this. It's certainly more than a formatting change as it removes the use of the filters plugin. I think I'd prefer that this part of the changes is reverted.

quaff:
> I think I'd prefer that this part of the changes is reverted.

Done, and capitalize `filters`.

PR #43323, opened 2024-11-29
Fix link to proxyBeanMethods in @AutoConfiguration javadoc

scordio:
Based on how the `@AutoConfiguration` Javadoc is [currently rendered](https://javadoc.io/doc/org.springframework.boot/spring-boot-autoconfigure/latest/org/springframework/boot/autoconfigure/AutoConfiguration.html), I have the impression that the `@literal` should have been an `@link`.

image

If that isn't the case, feel free to close this PR 🙂


wilkinsona:
Thanks for noticing and reporting this, @scordio.

PR #43327, opened 2024-11-29
Accept progress on numbers >2GB

kgb-financial-com:
Docker `ProgressUpdateEvent` may report file sizes larger than 2GB, but `int` has been used as the target type on parsing the according json output. If we attempt to push an image with a total size of >2GB, this will result in a build failure (although the image will still be pushed correctly).

This Pull Request fixes that by changing the affected types from `int` to `long`.


kgb-financial-com:
Sorry, I used the wrong branch for this pull request. Will create another one with the correct source branch.

PR #43328, opened 2024-11-29
Accept progress on numbers >2GB

kgb-financial-com:
Docker ProgressUpdateEvent may report file sizes larger than 2GB, but int has been used as the target type on parsing the according json output. If we attempt to push an image with a total size of >2GB, this will result in a build failure (although the image will still be pushed correctly).

This Pull Request fixes that by changing the affected types from int to long.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/ProgressUpdateEvent.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-buildpack-platform/src/main/java/org/springframework/boot/buildpack/platform/docker/ProgressUpdateEvent.java
@@ -22,6 +22,7 @@ import com.fasterxml.jackson.annotation.JsonCreator;
  * An {@link UpdateEvent} that includes progress information.
  *
  * @author Phillip Webb
+ * @author Wolfgang Kronberg
  * @since 2.3.0
  */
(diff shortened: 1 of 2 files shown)


wilkinsona:
@kgb-financial-com Thanks for the PR. Unfortunately, this doesn't qualify as an obvious fix as it changes functionality. Please sign the CLA.

kgb-financial-com:
Thanks. Please give me a few days to get approval from our top brass.

kgb-financial-com:
I have removed the tag as obvious fix from the original pull request description, and signed the CLA. Please proceed. :)

wilkinsona:
Thanks for signing the CLA.

Technically speaking, this is a breaking change as the constructor of `ProgressDetail` and the two getter methods are public API. In reality, I think it's likely that the API isn't used publicly so I'd be comfortable applying the change in 3.5 but I'm slightly wary of applying it in 3.3.x and 3.4.x. We'll discuss it as a team to figure out the next steps.

philwebb:
Looking at our own code, we only have one place that we call those methods which we can replace with a new `asPercentage()` method. I've polished the contribution to do that, and to restore the `int` returns in a deprecated form.

I agree with @wilkinsona that this class is mostly internal so I think we can be aggressive about removing the deprecated methods in 3.5.

Thanks for the contribution @kgb-financial-com

PR #43330, opened 2024-11-29
Ensure DefaultErrorAttributes adds JSON serialization safe errors

### LINKED ISSUE #42013: DefaultErrorAttributes doesn't populate errors for MethodValidationResult
GH-39865 added support for `MethodValidationResult` to `DefaultErrorAttributes` but it filters the list errors published in the result and used for counting to instances of `ObjectError`.

With a method signature (in a RestController) of e.g. `public void method(@PathVariable @Pattern(regexp = "^a.*") String parameter)`, this results in
* `message = "Validation failed for method='public void <class>.method(java.lang.String)'. Error count: 0"`
* `errors = []`

This is due to the fact, that the `MethodValidationResult` generated by `MethodValidationAdapter` will contain instances of `ParameterValidationResult` that have resolvable errors of type `DefaultMessageSourceResolvable` (and not `ObjectError`).

Spring Boot 3.3.1
Spring Framework 6.1.10

YongGoose:
resolved #42013 

- Added `ErrorWrapper` to enable `DefaultErrorAttributes` to handle errors related to `MethodValidationResult`.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/error/ErrorWrapper.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/error/ErrorWrapper.java
@@ -1,0 +1,90 @@
+/*
+ * Copyright 2012-2024 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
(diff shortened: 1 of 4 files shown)


YongGoose:
@wilkinsona @philwebb 

If there are any issues with my approach or implementation, I would appreciate your feedback 😁

philwebb:
@YongGoose I think from a cursory glance this is what we had in mind, but we won't be able to do a detailed review until we look at merging the PR. We can't do that until we branch for Spring Boot 3.5 which is still a few weeks away.

YongGoose:
@philwebb 
Thank you for the detailed explanation!
I'll wait for the branching process for Spring Boot 3.5 and the subsequent review.

philwebb:
Thanks very much @YongGoose!

PR #43331, opened 2024-11-30
Fix management.logging.export.enabled property in OnEnabledLoggingExportConditionTests

izeye:
This PR fixes `management.logging.export.enabled` property in `OnEnabledLoggingExportConditionTests`.


mhalbritter:
Thanks @izeye !

PR #43341, opened 2024-12-02
Restore System property in Logging section of the reference documentation

quaff:
Before this commit, it's rendered as unexpected link with simple class name, it should be full qualified class name.


snicoll:
Good catch, thanks again @quaff.

PR #43343, opened 2024-12-02
Refactor Remove redundant null check for sorter

Jae-Young98:
I think the second Null check on the classifier field assignment is redundant because its value has already been processed in the previous part of the method. Therefore, I modified it by specifying the sorter that has already been processed.


mhalbritter:
Thanks @Jae-Young98 !

PR #43347, opened 2024-12-02
refactor: remove unused  method from ValidationErrors

minwoo1999:
It appears that the hasErrors method is not referenced anywhere in the codebase. Removing it helps to simplify the class and improve code maintainability.


wilkinsona:
Thanks for the suggestion but we can't just remove this method as it's part of the public API. While it isn't used within Spring Boot itself, it may be used in third-party code. We'll have to decide if we want to keep the method or if it should be deprecated for removal in a future release.

Issue #43345, opened 2024-12-02
Make ReactiveUserDetailsServiceAutoConfiguration conditional on reactive web app


mhalbritter:
Related to https://github.com/spring-projects/spring-boot/pull/43334.

mhalbritter:
It's already conditional, see `org.springframework.boot.autoconfigure.security.reactive.ReactiveUserDetailsServiceAutoConfiguration.RSocketEnabledOrReactiveWebApplication`.

PR #43350, opened 2024-12-02
Make ReactiveUserDetailsServiceAutoConfiguration conditional on reactive web app

BenchmarkingBuffalo:
The UserDetailsServiceAutoConfiguration is only to be active in a reactive web application

I tried to align my tests to this commit https://github.com/spring-projects/spring-boot/commit/ff9fde0ef4a116bf13f0c5860414b0b2798fbdda
But I am not sure, if that is now entirely correct.
If you have remarks, please hit me up :)

See gh-43345

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/reactive/ReactiveUserDetailsServiceAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/reactive/ReactiveUserDetailsServiceAutoConfiguration.java
@@ -31,6 +31,7 @@ import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingBean
 import org.springframework.boot.autoconfigure.condition.ConditionalOnMissingClass;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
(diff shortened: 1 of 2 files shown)


BenchmarkingBuffalo:
The problem I am facing in this PR is that an RSocket-Application is not automatically recognized as a reactive webapplication.
This leads me to the question whether it should be adjusted, hence an RSocket-Application is always a reactive webapplication or if the Condition for `ReactiveUserDetailsServiceAutoConfiguration` needs to be adjusted to work on webapps and non webapps.
What do you think @mhalbritter @philwebb ?

mhalbritter:
Hmm yeah, I think we can't do that then. I overlooked the fact that RSocket is also supported by Spring Security. @bclozel WDYT?

bclozel:
Sorry about the very late reply.
Indeed, Spring Security supports RSocket and we need this auto-configuration to be activated in that case as well.

I think we would probably need a custom `AnyNestedCondition` that would have both  `@ConditionalOnBean(RSocketServerBootstrap.class)` and `@ConditionalOnWebApplication(type = Type.REACTIVE)`. This auto-configuration class would also need to be `@AutoConfiguration(after = RSocketServerAutoConfiguration.class)`.

@BenchmarkingBuffalo are you still available for amending your PR with this proposed change? Thanks!

mhalbritter:
Huh, there's already a condition for "RSocket or Reactive web app". It has been added in 40ac5b4ae2faff0baf1888531218eedbf4c1aa22. I have overlooked that while opening https://github.com/spring-projects/spring-boot/issues/43345. Sorry!

PR #43360, opened 2024-12-03
Remove unused ON_LOCATION_NOT_FOUND_PROPERTY constant

minwoo1999:
Remove unused ON_LOCATION_NOT_FOUND_PROPERTY constant


philwebb:
This is a public constant and is used by Spring Cloud so we cannot remove it.

PR #43364, opened 2024-12-03
refactor Test

Felixqaq:
refactor the case to test


philwebb:
These files are not intended to be tests, they are resources used by `ArchitectureCheckTests`

Felixqaq:
Thank you! I understand that.

Issue #43312, opened 2024-11-28
StructuredLoggingJsonMembersCustomizer implementations declared in spring.factories with a generic type more specific than Object are not called


quaff:
```java
package com.example;

import org.springframework.boot.json.JsonWriter.Members;
import org.springframework.boot.logging.structured.StructuredLoggingJsonMembersCustomizer;

public class MyCustomizer implements StructuredLoggingJsonMembersCustomizer<String> {

	@Override
	public void customize(Members<String> members) {
		members.add("test", "value");

	}

}

```
```java
package com.example;

import static org.assertj.core.api.Assertions.assertThat;

import org.junit.jupiter.api.Test;
import org.junit.jupiter.api.extension.ExtendWith;
import org.springframework.boot.test.context.SpringBootTest;
import org.springframework.boot.test.system.CapturedOutput;
import org.springframework.boot.test.system.OutputCaptureExtension;

@SpringBootTest
@ExtendWith(OutputCaptureExtension.class)
public class MyCustomizerTests {

	@Test
	void test(CapturedOutput output) {
		assertThat(output).contains("{\"@timestamp\""); // structured logging is working
		assertThat(output).contains("\"test\":\"value\""); // MyCustomizer is working
	}
}

```
test failed with:
```
java.lang.AssertionError: 
Expecting actual:
  {"@timestamp":"2024-11-28T08:42:45.558754Z","log.level":"INFO","process.pid":71176,"process.thread.name":"main","log.logger":"com.example.MyCustomizerTests","message":"Starting MyCustomizerTests using Java 17.0.13 with PID 71176","ecs.version":"8.11"}
{"@timestamp":"2024-11-28T08:42:45.562533Z","log.level":"INFO","process.pid":71176,"process.thread.name":"main","log.logger":"com.example.MyCustomizerTests","message":"No active profile set, falling back to 1 default profile: \"default\"","ecs.version":"8.11"}
{"@timestamp":"2024-11-28T08:42:46.289631Z","log.level":"INFO","process.pid":71176,"process.thread.name":"main","log.logger":"com.example.MyCustomizerTests","message":"Started MyCustomizerTests in 0.939 seconds (process running for 1.622)","ecs.version":"8.11"}

to contain:
  ""test":"value"" 
	at com.example.MyCustomizerTests.test(MyCustomizerTests.java:18)
	at java.base/java.lang.reflect.Method.invoke(Method.java:569)
[...]

PR #43368, opened 2024-12-04
Add support for multiple StructuredLoggingJsonMembersCustomizers

quaff:
See https://github.com/spring-projects/spring-boot/issues/43312#issuecomment-2509058761


quaff:
Should `logging.structured.json.customizer` be deprecated and rename to `logging.structured.json.customizers`?

philwebb:
Thanks @quaff, I've kept both the `customizer` and `customizers` properties so that folks can pick the one that reads best for their situation.

philwebb:
Reopening because we discussed this today and we're going to deprecate the singular customizer version. I'll take care of that today.

PR #43424, opened 2024-12-05
spring-boot-maven-plugin sets imagePlatform even if it's empty

vonZeppelin:
A Docker image cannot be built with the spring-boot-maven-plugin:3.4.0 if Docker API version is less than 1.41. In particular, a Spring Boot project builds are failing in Bitbucket pipelines when no 'imagePlatform' option configured for the project with the following error
```
Docker API version must be at least 1.41 to support the 'imagePlatform' option, but current API version is 1.24
```

Using an empty `<imagePlatform/>` tag in the `pom.xml` doesn't help as Maven, most probably, assigns an empty string to the corresp. Java property.

The proposed fix is to check the config property for both null and empty string values.


mhalbritter:
Thanks @vonZeppelin !

johnasiano:
Why would someone be using an API version less than 1.41?

Issue #24170, opened 2020-11-16
SpringBoot actuator with different port causes FileNotFoundException: /dev/management_stdout


marcellodesales:
* We need to change the actuator endpoints to be on a different port number (To serve the Kubernetes probes)
  * We also need the logs from the Actuator to go to `stdout`
* Changing the value of `management.server.port: 8404` causes an exception as the server configures the new port to stream logs to `/dev/management_stdout` when it should be `/dev/stdout` (already configured for the server, as shown in the `Question` section below)

# SpringBoot Setup

```gradle
buildscript {
    repositories {
        mavenCentral()

        // For src/gradle/tests.gradle
        maven {
            url "https://plugins.gradle.org/m2/"
        }
    }
    dependencies {
        classpath("org.springframework.boot:spring-boot-gradle-plugin:2.3.5.RELEASE")
    }
}

dependencies {
    implementation('org.springframework.boot:spring-boot-starter-actuator')
}
```

# Changing SpringBoot Actuator

* Changed the port number as follows in `application.yaml`

```yaml
management:
  endpoints:
    web:
      exposure:
        include: "*"
  server:
    port: 8404
```

# Error shown during bootstrap

* The default logger is trying to write to `/dev/management_stdout`

```
2020-11-15T17:23:41 ERROR [app=parking-plus-service,prf=default][tid=,sid=,sxp=][cid=] 77926 --- 
  [  restartedMain] o.apache.catalina.valves.AccessLogValve  : 
        Failed to open access log file [/dev/management_stdout]

java.io.FileNotFoundException: /dev/management_stdout (Operation not permitted)
	at java.base/java.io.FileOutputStream.open0(Native Method)
	at java.base/java.io.FileOutputStream.open(FileOutputStream.java:291)
	at java.base/java.io.FileOutputStream.<init>(FileOutputStream.java:234)
	at org.apache.catalina.valves.AccessLogValve.open(AccessLogValve.java:651)
	at org.apache.catalina.valves.AccessLogValve.startInternal(AccessLogValve.java:685)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:183)
	at org.apache.catalina.core.StandardPipeline.startInternal(StandardPipeline.java:176)
	at org.apache.catalina.util.LifecycleBase.start(LifecycleBase.java:183)
	at org.apache.catalina.core.ContainerBase.startInternal(ContainerBase.java:933)
[...]

Issue #14948, opened 2018-10-24
Allow specifying a different management access log prefix


frzme:
Currently access logs created by the management server (when the management server is run on a different port) automatically get the prefix "management_" prepended (implemented in `org.springframework.boot.actuate.autoconfigure.web.servlet.ServletManagementChildContextConfiguration` ).
For configurations of access logs which write to stdout (see https://github.com/ImmobilienScout24/tomcat-stdout-accesslog/issues/2 for an example on how to configure that in tomcat) it would be useful to be able to drop this prefix.
There might also be people interested in configuring this static prefix.

Please expose the prefix as a property and thus make it configurable by end users.

philwebb:
For a workaround you could your write your own `WebServerFactoryCustomizer` ordered above ours that removes the default `AccessLogValve` from `getEngineValves` and replaces it with subclass that overrides `setPrefix` but doesn't call `super`.

skarzhevskyy:
workaround code that works for us
```
/**
 * Add this class to /META-INF/spring.factories under the {@code org.springframework.boot.actuate.autoconfigure.web.ManagementContextConfiguration}
 */
@ManagementContextConfiguration(value = ManagementContextType.CHILD)
public class MyAppManagementConfig {

    @Bean
    public WebServerFactoryCustomizer<TomcatServletWebServerFactory> managementServerustomizer() {
        return factory -> {
            AccessLogValve accessLogValve = findAccessLogValve(factory);
            if ((accessLogValve != null) && "management_stdout".equals(accessLogValve.getPrefix())) {
                accessLogValve.setPrefix("stdout");
            }
        };
    }

    private static AccessLogValve findAccessLogValve(TomcatServletWebServerFactory factory) {
        return (AccessLogValve) factory.getEngineValves().stream()
                .filter(AccessLogValve.class::isInstance)
                .findFirst().orElse(null);
    }
}
```

virgilsb:
Hi! Was there any progress with this feature?

Note for the workaround:
Since spring-boot 2.7, the ManagementConfig class needs to be specified in this file: `/META-INF/spring/org.springframework.boot.actuate.autoconfigure.web.ManagementContextConfiguration.imports`, according to the [documentation](https://docs.spring.io/spring-boot/docs/2.7.0/api/org/springframework/boot/actuate/autoconfigure/web/ManagementContextConfiguration.html).

akundrock:
[...]

PR #43434, opened 2024-12-06
Allow specifying a different management access log prefix

mpalourdio:
Hi,

This a WIP in order to try to tackle GH-14948, GH-24170 for example.
It's today very common to handle logs in stdout, regarding k8s and such 'cloud' platforms., ie

```yml
server:
 tomcat:
    accesslog:
      enabled: true
      directory: /dev
      prefix: stdout
      suffix:
 ```

It's also reassuring (for SEC) to make actuator available on another port. Wrong properties can easily leak and bad things may happen.

This is just a draft to see if this could be an answer in order to have the ability to disable the access logs prefix while using actuator on an other port. Most companies that put logs in stdout can easily filter logs with tools like [Splunk](https://www.splunk.com), so putting all those logs in the same output is not a problem at all.

The idea is to add a property to disable access logs prefix for actuator. Default is `true` to keep BC

`management.server.accesslog-prefix=false`

Is this the good direction?

Thanks

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/server/ManagementServerProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/server/ManagementServerProperties.java
@@ -57,6 +57,8 @@ public class ManagementServerProperties {
 	@NestedConfigurationProperty
 	private Ssl ssl;
 
+	private final Accesslog accesslog = new Accesslog();
+
 	/**
 	 * Returns the management port or {@code null} if the
 	 * {@link ServerProperties#getPort() server port} should be used.
@@ -115,6 +117,28 @@ public class ManagementServerProperties {
 			}
 		}
 		return candidate;
+	}
+
+	public Accesslog getAccesslog() {
+		return this.accesslog;
+	}
+
+	public static class Accesslog {
+
+		/**
+		 * Enable management access logs prefix customization
+		 * management.server.accesslog.prefix.
+		 */
+		private String prefix = "management_";
+
+		public String getPrefix() {
+			return this.prefix;
+		}
+
+		public void setPrefix(String prefix) {
+			this.prefix = prefix;
+		}
+
 	}
 
 }
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
@@ -47,6 +47,7 @@ import org.springframework.boot.autoconfigure.web.embedded.UndertowWebServerFact
 import org.springframework.boot.autoconfigure.web.servlet.ServletWebServerFactoryCustomizer;
 import org.springframework.boot.autoconfigure.web.servlet.TomcatServletWebServerFactoryCustomizer;
 import org.springframework.boot.autoconfigure.web.servlet.UndertowServletWebServerFactoryCustomizer;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
 import org.springframework.boot.web.embedded.jetty.JettyServletWebServerFactory;
 import org.springframework.boot.web.embedded.tomcat.TomcatServletWebServerFactory;
 import org.springframework.boot.web.embedded.undertow.UndertowServletWebServerFactory;
@@ -73,6 +74,7 @@ import org.springframework.util.StringUtils;
  */
 @ManagementContextConfiguration(value = ManagementContextType.CHILD, proxyBeanMethods = false)
 @ConditionalOnWebApplication(type = Type.SERVLET)
+@EnableConfigurationProperties(ManagementServerProperties.class)
 class ServletManagementChildContextConfiguration {
 
 	@Bean
@@ -83,20 +85,22 @@ class ServletManagementChildContextConfiguration {
 
 	@Bean
 	@ConditionalOnClass(name = "io.undertow.Undertow")
-	UndertowAccessLogCustomizer undertowManagementAccessLogCustomizer() {
-		return new UndertowAccessLogCustomizer();
+	UndertowAccessLogCustomizer undertowManagementAccessLogCustomizer(
+			ManagementServerProperties managementServerProperties) {
+		return new UndertowAccessLogCustomizer(managementServerProperties);
 	}
 
 	@Bean
 	@ConditionalOnClass(name = "org.apache.catalina.valves.AccessLogValve")
-	TomcatAccessLogCustomizer tomcatManagementAccessLogCustomizer() {
-		return new TomcatAccessLogCustomizer();
+	TomcatAccessLogCustomizer tomcatManagementAccessLogCustomizer(
+			ManagementServerProperties managementServerProperties) {
+		return new TomcatAccessLogCustomizer(managementServerProperties);
 	}
 
 	@Bean
 	@ConditionalOnClass(name = "org.eclipse.jetty.server.Server")
-	JettyAccessLogCustomizer jettyManagementAccessLogCustomizer() {
-		return new JettyAccessLogCustomizer();
+	JettyAccessLogCustomizer jettyManagementAccessLogCustomizer(ManagementServerProperties managementServerProperties) {
+		return new JettyAccessLogCustomizer(managementServerProperties);
 	}
 
 	@Configuration(proxyBeanMethods = false)
@@ -145,14 +149,18 @@ class ServletManagementChildContextConfiguration {
 
 	abstract static class AccessLogCustomizer implements Ordered {
 
-		private static final String MANAGEMENT_PREFIX = "management_";
+		protected final ManagementServerProperties managementServerProperties;
+
+		AccessLogCustomizer(ManagementServerProperties managementServerProperties) {
+			this.managementServerProperties = managementServerProperties;
+		}
 
 		protected String customizePrefix(String prefix) {
 			prefix = (prefix != null) ? prefix : "";
-			if (prefix.startsWith(MANAGEMENT_PREFIX)) {
+			if (prefix.startsWith(this.managementServerProperties.getAccesslog().getPrefix())) {
 				return prefix;
 			}
-			return MANAGEMENT_PREFIX + prefix;
+			return this.managementServerProperties.getAccesslog().getPrefix() + prefix;
 		}
 
 		@Override
@@ -165,12 +173,17 @@ class ServletManagementChildContextConfiguration {
 	static class TomcatAccessLogCustomizer extends AccessLogCustomizer
 			implements WebServerFactoryCustomizer<TomcatServletWebServerFactory> {
 
+		TomcatAccessLogCustomizer(ManagementServerProperties managementServerProperties) {
+			super(managementServerProperties);
+		}
+
 		@Override
 		public void customize(TomcatServletWebServerFactory factory) {
 			AccessLogValve accessLogValve = findAccessLogValve(factory);
 			if (accessLogValve == null) {
 				return;
 			}
+
 			accessLogValve.setPrefix(customizePrefix(accessLogValve.getPrefix()));
 		}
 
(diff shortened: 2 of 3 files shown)

### CODE REVIEW

File: spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/server/ManagementServerProperties.java [hunk 1]
```diff
@@ -54,6 +54,12 @@ public class ManagementServerProperties {
 	 */
 	private String basePath = "";
 
+	/**
+	 * Enable management access logs to be prefixed with management_
+	 * management.server.accesslog.prefix.
+	 */
+	private boolean accesslogPrefix = true;
```
mhalbritter, line 61: Instead of a boolean toggle for the prefix, I'd prefer a `String` variable defaulting to `management_`. This way, people can just set it to `""` to get rid of the prefix, but also allows them to customize it.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
```diff
@@ -165,13 +166,22 @@ public int getOrder() {
 	static class TomcatAccessLogCustomizer extends AccessLogCustomizer
 			implements WebServerFactoryCustomizer<TomcatServletWebServerFactory> {
 
+		private final ManagementServerProperties managementServerProperties;
+
+		TomcatAccessLogCustomizer(ManagementServerProperties managementServerProperties) {
```
mhalbritter, line 171: You only have changed it for Tomcat. This change also has to be applied to Undertow and Jetty.

mhalbritter (changes requested):
Hello @mpalourdio,

thanks for the PR. I've added some comments for your consideration.

mpalourdio:
@mhalbritter I have forced-push with your suggestion. I have also taken care of Undertow and Jetty

mpalourdio:
Just wondering if would not be better to create an inner class `AccessLog` in `ManagementServerProperties` so the property would be more in sync with `server.tomcat.accesslog.prefix`, and could handle further configuration properties. WDYT?

mhalbritter:
Yes, we should add an `Accesslog` inner class to `ManagementServerProperties` like you suggested.

There's already a `server.tomcat.accesslog.prefix` which is by default set to `access_log`. Right now, what happens if I set `management.server.accesslog-prefix=management_`, i'll end up with an management access log which is named `management_access_log` because `server.tomcat.accesslog.prefix` is prefixed with `management.server.accesslog-prefix`. I wonder if it would be clearer if `management.server.accesslog-prefix` would override `server.tomcat.accesslog.prefix` completely.

So to get a management access log called `management_access_log`, i'd have to set `management.server.accesslog-prefix=management_access_log`. That would work regardless of the value of the `server.tomcat.accesslog.prefix` property.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/server/ManagementServerProperties.java
```diff
@@ -54,6 +54,12 @@ public class ManagementServerProperties {
 	 */
 	private String basePath = "";
 
+	/**
+	 * Enable management access logs prefix customization
+	 * management.server.accesslog.prefix.
+	 */
+	private String accesslogPrefix = "management_";
```
mhalbritter, line 61: Please refactor to use an inner static class named `Accesslog`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
```diff
@@ -147,12 +149,12 @@ abstract static class AccessLogCustomizer implements Ordered {
 
 		private static final String MANAGEMENT_PREFIX = "management_";
 
-		protected String customizePrefix(String prefix) {
+		protected String customizePrefix(String prefix, String accessLogPrefix) {
```
mhalbritter, line 152: Instead of changing the signature and storing the `ManagementServerProperties` in every webserver implementation, you could add a `private final ManagementServerProperties managementServerProperties` to `AccessLogCustomizer` and use the value in `customizePrefix`.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
```diff
@@ -147,12 +149,12 @@ abstract static class AccessLogCustomizer implements Ordered {
...
 			prefix = (prefix != null) ? prefix : "";
 			if (prefix.startsWith(MANAGEMENT_PREFIX)) {
```
mhalbritter, line 154: This looks wrong to me. It should check against `accessLogPrefix`, not against a hardcoded "management_".

mhalbritter (changes requested):

mpalourdio:
> Yes, we should add an `Accesslog` inner class to `ManagementServerProperties` like you suggested.
> 
right, will give this a shot

> There's already a `server.tomcat.accesslog.prefix` which is by default set to `access_log`. Right now, what happens if I set `management.server.accesslog-prefix=management_`, i'll end up with an management access log which is named `management_access_log` because `server.tomcat.accesslog.prefix` is prefixed with `management.server.accesslog-prefix`. I wonder if it would be clearer if `management.server.accesslog-prefix` would override `server.tomcat.accesslog.prefix` completely.
> 
> So to get a management access log called `management_access_log`, i'd have to set `management.server.accesslog-prefix=management_access_log`. That would work regardless of the value of the `server.tomcat.accesslog.prefix` property.

That behavior seems counter-intuitive to me. If you have already tweaked  `server.tomcat.accesslog.prefix` for any reason, and then you add this new property, (so you remove  `server.tomcat.accesslog.prefix` because it has become useless), then  you decide to remove actuator from your project you would have to reconfigure the tomcat prefix etc.

IMO with this new property, it's good as is without any user intervention and does the only job it's made for.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
```diff
@@ -83,20 +83,22 @@ ServletManagementWebServerFactoryCustomizer servletManagementWebServerFactoryCus
 
 	@Bean
 	@ConditionalOnClass(name = "io.undertow.Undertow")
-	UndertowAccessLogCustomizer undertowManagementAccessLogCustomizer() {
-		return new UndertowAccessLogCustomizer();
+	UndertowAccessLogCustomizer undertowManagementAccessLogCustomizer(
```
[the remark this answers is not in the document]
mpalourdio: It works without for the moment. Is it Undertow specific ?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
mpalourdio: Was note really sure about doing this, but it can be changed easily

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
mpalourdio: `AccessLogCustomizer` is abstract, if I add a constructor to it, I will need to call `super(ManagementServerProperties)` in every implementation anyway. Is this what you want ?

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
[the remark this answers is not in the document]
mhalbritter: No, I guess something else imports the configuration properties, so that this isn't a problem. But generally speaking, every configuration (or bean) which uses configuration properties should enable them with `@EnableConfigurationProperties` to be on the safe side.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
mhalbritter: Yes.

mhalbritter:
Flagging this to see what the team thinks about this:

> There's already a `server.tomcat.accesslog.prefix` which is by default set to `access_log`. Right now, what happens if I set `management.server.accesslog-prefix=management_`, i'll end up with an management access log which is named `management_access_log` because `server.tomcat.accesslog.prefix` is prefixed with `management.server.accesslog-prefix`. I wonder if it would be clearer if `management.server.accesslog-prefix` would override `server.tomcat.accesslog.prefix` completely.

> So to get a management access log called `management_access_log`, i'd have to set `management.server.accesslog-prefix=management_access_log`. That would work regardless of the value of the `server.tomcat.accesslog.prefix` property.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/server/ManagementServerProperties.java
mpalourdio: done

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
mpalourdio: done

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/web/servlet/ServletManagementChildContextConfiguration.java
```diff
@@ -145,14 +149,18 @@ private String getContextPath(ManagementServerProperties managementServerPropert
...
 
-		private static final String MANAGEMENT_PREFIX = "management_";
+		protected final ManagementServerProperties managementServerProperties;
 
-		protected String customizePrefix(String prefix) {
+		AccessLogCustomizer(ManagementServerProperties managementServerProperties) {
+			this.managementServerProperties = managementServerProperties;
+		}
+
+		protected String customizePrefix(String prefix, String managementAccesslogPrefix) {
```
mhalbritter, line 158: You can drop this parameter and use `this.managementServerProperties.getAccesslog().getPrefix()` instead.

mpalourdio: Good catch, done

mhalbritter (approved):

mpalourdio:
Hi, i see this targets `3.5.x`. Will this land in further `3.3.next` and `3.4.next` ? It's BC and the code has not moved from `3.2.X` AFAIK. 
Thanks for feedback.

mhalbritter:
Hi - this is labeled as an enhancement, and enhancements are put in new versions only, not in the maintenance one. That means this feature will land in Spring Boot 3.5.0 the earliest.

mhalbritter:
Thanks @mpalourdio !

Issue #42972, opened 2024-11-01
Add JsonObjectDeserializer.nullSafeValue method that accepts a mapper Function


philwebb:
See #42956 for background. Rather than adding direct support for types we could add support for a general purpose mapper function.

PR #43441, opened 2024-12-08
Add test for nullSafeValue with mapper transformation

minwoo1999:
I’ve added a test case for the newly introduced nullSafeValue method that accepts a mapper function (introduced in #42972, see #42956 for background). This test ensures the method correctly transforms the value using the provided mapper function, validating its behavior. Thank you for considering and addressing the use case!

If there are any improvements or additional suggestions, I’d greatly appreciate your feedback.


philwebb:
Thanks very much @minwoo1999!

PR #43258, opened 2024-11-22
TestRestTemplate does not allow redirects to be customized

quaff:
Closes GH-27360

PR #43444, opened 2024-12-09
Expose accessor method for underlying ClientHttpRequestFactorySettings

quaff:
After this commit, we can customize ClientHttpRequestFactorySettings base on underlying one.


philwebb:
I'm not keen to add access for the settings. I've reopened #43258 so we can continue the discussion there.

PR #43450, opened 2024-12-09
Kafka dependency management does not include the kafka-server module

sobychacko:
Apache Kafka now ships a new module, kafka-server, since the 3.7.0 release. The `3.9.0` kafka-client introduced some breaking changes that require this dependecy for the `EmbeddedKafka` support in Spring for Apache Kafka. This commit adds this dependecny for Spring Boot based Spring Kafka projects.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1115,6 +1115,7 @@ bom {
 				"kafka-log4j-appender",
 				"kafka-metadata",
 				"kafka-raft",
+				"kafka-server",
 				"kafka-server-common",
 				"kafka-server-common" {
 					classifier = "test"


artembilan (approved):

philwebb:
@sobychacko @artembilan Do you consider this a bug (something for 3.3 or 3.4) or an enhancement (something to wait for 3.5).

artembilan:
It is hard to judge since it works with currently included Kafka client, but a new one (`3.9.0`), which we are trying to be compatible, doesn't work as is. That `kafka-server` has to be included.
Our goal is to not have next minor version for Spring Kafka, but rather keep the current one compatible with the latest client version and just aim for `4.0` next year, where Kafka Client `4.0` should appear as well.

It is not too hard to include that dependency into the target project, but might be very convenient just to override version and that's it.
At the same time, we include into Spring Boot deps management many other Kafka libs, so, since this one is missed for while (https://mvnrepository.com/artifact/org.apache.kafka/kafka-server), the fix might be applied for Spring Boot `3.3` as well.
That's where we have started to support Kafka client `3.7.x`.

philwebb:
I think it's probably safe to apply the fix to 3.3. It's a managed dependency, but not actually pulled in by default.

PR #43508, opened 2024-12-14
Add ThreadPoolTaskExecutorsEndpoint

lucky8987:
Add ThreadPoolTaskExecutorsEndpoint, It is mainly used to observe the information about the asynchronous task thread pool in the @Async annotation, such as the number of core threads, queue size, and number of tasks to be processed.


bclozel:
I don't think we should expose observability data on a different endpoint. This endpoint merely exports metrics on a different endpoint with lower capabilities since those are not multi-dimensional and cannot be filtered accordingly.

Micrometer instruments executors with "ExecutorServiceMetrics" and metrics will be exposed on the dedicated endpoint.

snicoll:
Agreed, @lucky8987 you should be using the metrics support for this.

PR #43520, opened 2024-12-16
Polish

arefbehboudi:


wilkinsona:
Thanks for the proposal but I'm not sure that this is an improvement. The constructor to which you've added the assertions is package-private and should never be called with `null` for any of its arguments. Have you seen a `NullPointerException` that suggests otherwise?

arefbehboudi:
You're absolutely right. These changes are unnecessary, and since the constructor is package-private and shouldn't be called with null arguments, there's no need for these checks. Thank you for pointing that out!

PR #43532, opened 2024-12-16
Add Pulsar 4.0.x smoke test

onobc:
Unfortunately, I missed the fact that the `3.3.3` version of Pulsar (which is the version in Boot `3.4.x` and Spring for Apache Pulsar `1.2.x`) is the last patch release of the the `3.3.x` line. The next LTS version is `4.0.x`.

I knew that `4.0.x` was released a few weeks prior to SP `1.2.0` and SB `3.4.0` but I missed the detail about it being the last patch release and the team decided not to jump a major version in a minor SB/SP version. Otherwise, I would have recommended that we update to `4.0.x`.

Hindsight is 20/20....

So, this PR adds a compatibility test for users that want to update to Pulsar 4.0.1 client in Spring Boot apps. I have done the same on the Spring for Apache Pulsar framework version 1.2.x and it all works well.

> [!NOTE]
> This is the same approach taken for [Tomcat 11](https://github.com/spring-projects/spring-boot/commit/37ae78ea73e043dbe8b605c586e3cc845a9d2240)

### DIFF
--- a/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-pulsar4/build.gradle
+++ b/spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-pulsar4/build.gradle
@@ -1,0 +1,27 @@
+plugins {
+	id "java"
+	id "org.springframework.boot.docker-test"
+}
+
+description = "Spring Boot Pulsar smoke test"
+
+configurations.all {
+	resolutionStrategy.eachDependency {
+		if (it.requested.group == 'org.apache.pulsar' &&
+				!(it.requested.name.startsWith('pulsar-client-reactive'))) {
+			it.useVersion '4.0.1'
+		}
+	}
+}
+
+dependencies {
+	dockerTestImplementation(project(":spring-boot-project:spring-boot-starters:spring-boot-starter-test"))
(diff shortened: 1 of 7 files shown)


spring-boot-tests/spring-boot-smoke-tests/spring-boot-smoke-test-pulsar4/build.gradle
```diff
@@ -0,0 +1,27 @@
...
+description = "Spring Boot Pulsar smoke test"
+
+configurations.all {
+	resolutionStrategy.eachDependency {
+		if (it.requested.group == 'org.apache.pulsar' &&
+				!(it.requested.name.startsWith('pulsar-client-reactive'))) {
+			it.useVersion '4.0.1'
+		}
+	}
+}
```
onobc, line 15: I really really do not like the fact that this is 99.98% the same as `spring-boot-smoke-test-pulsar`.  These lines are the **ONLY** difference between the modules. I wanted to keep this PR simple but if there is an approach already in place to respect DRY that would be a great polish commit on this.

philwebb: I think in this case the duplication is fine. The samples aren't published and we can drop this one in Boot 3.5 and just revert to having a single smoke test.

PR #43571, opened 2024-12-19
Hint about Testcontainers in DataSourceBeanCreationFailureAnalyzer

mstahv:
I'm a big fan of setting up the developent time datasource with Testcontainers. The problem I see people checking out and trying my examples/demos is that they are not (yet) familiar with the idea of starting the development time server from the srs/test/java.  

Similarly to which DataSourceBeanCreationFailureAnalyzer currently hints about embedded databases, I think in 2024 it should hint somehow about Testcontainers too. Probably just guiding them to seek an application class from the test sources is enough (like currently in this PR), but maybe it should mention e.g. `mvn spring-boot:test-run` 🤷‍♂️

For one of my recent demos I [added an app specific FailureAnalyzer](https://github.com/mstahv/sakila-spring-data-jpa-starter/blob/mysql/src/main/java/com/example/demo/TestcontainersNewbieAnalyzer.java), but I don't think that is a good approach.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceBeanCreationFailureAnalyzer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jdbc/DataSourceBeanCreationFailureAnalyzer.java
@@ -66,8 +66,8 @@ class DataSourceBeanCreationFailureAnalyzer extends AbstractFailureAnalyzer<Data
 		StringBuilder action = new StringBuilder();
 		action.append(String.format("Consider the following:%n"));
 		if (EmbeddedDatabaseConnection.NONE == cause.getConnection()) {
-			action.append(String
-					.format("\tIf you're using Testcontainers to set up your development datasource, consider starting the application from the test sources instead.%n"));
+			action.append(String.format(
+					"\tIf you're using Testcontainers to set up your development datasource, consider starting the application from the test sources instead.%n"));
 			action.append(String
 				.format("\tIf you want an embedded database (H2, HSQL or Derby), please put it on the classpath.%n"));
 		}


philwebb:
Thanks for the suggestion, but we're not keen to add this link because there are many other services as well as databases that could be started via Testcontainers. We also don't have a good way of detecting the test main class when running just the main application.

mstahv:
How are the "many other services" related to missing DataSource🤔 Would be indeed nice if this message could be hidden if already starting with test classpath, as then it is obviously not valid. For a moment I considered checking if some classes from JUnit/TestNG are available, but then though that also the existing hint about embedded databases is (most?) often invalid.

philwebb:
> How are the "many other services" related to missing DataSource

By other services, I mean any other Testcontainer that we have `ConnectionDetails` support for. E.g. Redis. i.e. the problem is larger than just datasources.

> For a moment I considered checking if some classes from JUnit/TestNG are available

That's probably not going to work for this situation because the IDE usually doesn't include test dependencies when running the main code. I don't think we have a good way of telling that there is another class with an alternative main method available under `src/test/java`.

That only leaves us with the option of always suggesting applications might want to be starting from the test sources, but we're really not keen on that.

mstahv:
Ah, so you would wish that the hint would be in some more generic place than in DataSource related analyzer👍 Don't have good ideas for that.

Yes, I know how IDEs regarding test/main dependencies, but checking for non-exisiting "org.junit.jupiter.api.Test" (or something similar like "SpringBootTest") works pretty well as the hint could be hidden if already runnign from the test sources (and the issue is then probably somewhere else).

```
 public class TestDemoApplication {

		if (checkTestClassAvailable()) {
			System.out.println("You have JUnit 5 on the classpath, running from test sources hint should be hidden.");
		}
		SpringApplication.from(DemoApplication::main).with(TestcontainersConfiguration.class).run(args);
	}

	private static boolean checkTestClassAvailable() {
		try {
			Class.forName("org.junit.jupiter.api.Test");
			return true;
		} catch (ClassNotFoundException e) {
			return false;
		}
	}

}
```

But sure, I agree the hint can't be made perfect, but I think it might help many newbies in the same way as the current embedded database hint.

PR #43586, opened 2024-12-20
FileWatcher used for SSL reload does not support symlinks

tmaciejewski:
This allows using symlinks for SSL certificate hot reloading.

In my company we have services on k8s with SSL certificates stored on a mounted volumes. Since they are symlinks, we cannot use hot-reloading, because the content is changed in the file under the symlink, not the symlink itself. This change make `FileWatcher` watch not only the symlink, but also the file it resolves to.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/FileWatcher.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/FileWatcher.java
@@ -86,12 +86,23 @@ class FileWatcher implements Closeable {
 					this.thread = new WatcherThread();
 					this.thread.start();
 				}
-				this.thread.register(new Registration(paths, action));
+				this.thread.register(new Registration(resolveSymlinks(paths), action));
 			}
 			catch (IOException ex) {
 				throw new UncheckedIOException("Failed to register paths for watching: " + paths, ex);
 			}
 		}
+	}
+
+	private Set<Path> resolveSymlinks(Set<Path> paths) throws IOException {
+		Set<Path> result = new HashSet<>();
+		for (Path path : paths) {
+			result.add(path);
+			if (Files.isSymbolicLink(path)) {
+				result.add(Files.readSymbolicLink(path));
+			}
+		}
+		return result;
 	}
 
 	@Override
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/FileWatcher.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/FileWatcher.java
@@ -97,10 +97,7 @@ class FileWatcher implements Closeable {
 	private Set<Path> resolveSymlinks(Set<Path> paths) throws IOException {
 		Set<Path> result = new HashSet<>();
 		for (Path path : paths) {
-			result.add(path);
-			if (Files.isSymbolicLink(path)) {
-				result.add(Files.readSymbolicLink(path));
-			}
+			result.add(Files.isSymbolicLink(path) ? Files.readSymbolicLink(path) : path);
 		}
 		return result;
 	}
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/ssl/FileWatcherTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/ssl/FileWatcherTests.java
@@ -95,6 +95,18 @@ class FileWatcherTests {
 	}
 
 	@Test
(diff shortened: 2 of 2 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/FileWatcher.java
```diff
@@ -86,14 +86,25 @@ void watch(Set<Path> paths, Runnable action) {
...
 			catch (IOException ex) {
 				throw new UncheckedIOException("Failed to register paths for watching: " + paths, ex);
 			}
 		}
 	}
 
+	private Set<Path> resolveSymlinks(Set<Path> paths) throws IOException {
+		Set<Path> result = new HashSet<>();
+		for (Path path : paths) {
+			result.add(path);
```
snicoll, line 100: What is the reason for adding the symlink to the files to watch. Shouldn't we only add the real file?

snicoll (commented):
Thanks for the PR. This looks good but I am wondering about something and added a question.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/ssl/FileWatcher.java
tmaciejewski: Thanks for the review. Yeah, you're right. I wanted to keep backward compatibility, but since the real file is watched, it would still be there if the link is changed to some other file, which is less intuitive. Changed it to watch only the real file.

snicoll:
@tmaciejewski thank you for making your first contribution to Spring Boot.

PetarDimitrovZH:
Hi Everyone!

Is this released? I have the same or similar issue. I use kubernetes cert manager in order to manage my certs. It stores those as sym links:

```
drwxr-xr-x 2 maniax maniax 4096 Feb 18 10:58 ..data
lrwxrwxrwx 1 maniax maniax   19 Feb 18 10:59 keystore.p12 -> ..data/keystore.p12
lrwxrwxrwx 1 maniax maniax   21 Feb 18 11:00 truststore.p12 -> ..data/truststore.p12
```


ssl config:


```yaml
spring:
  application:
    name: "shell"
  ssl:
    bundle:
      jks:
        shell:
          reload-on-update: false
          keystore:
            location: ${KEYSTORE:file:/opt/certs/keystore.p12}
            password: ${KEYSTORE_PW:changeit}
            type: "PKCS12"
          truststore:
            location: ${TRUSTSTORE:file:/opt/certs/truststore.p12}
            password: ${TRUSTSTORE_PW:changeit}
            type: "PKCS12"
```


But somehow spring cannot resolve them and I got this exception:

Caused by: java.lang.IllegalStateException: Unable to watch for reload on update
	at org.springframework.boot.autoconfigure.ssl.SslPropertiesBundleRegistrar.watchForUpdates(SslPropertiesBundleRegistrar.java:85) ~[spring-boot-autoconfigure-3.4.2.jar:3.4.2]
	at org.springframework.boot.autoconfigure.ssl.SslPropertiesBundleRegistrar.lambda$registerBundles$2(SslPropertiesBundleRegistrar.java:70) ~[spring-boot-autoconfigure-3.4.2.jar:3.4.2]
	... 74 common frames omitted
Caused by: java.io.UncheckedIOException: Failed to register paths for watching: [/opt/certs/keystore.p12, /opt/certs/truststore.p12]
	at org.springframework.boot.autoconfigure.ssl.FileWatcher.watch(FileWatcher.java:96) ~[spring-boot-autoconfigure-3.4.2.jar:3.4.2]
	at org.springframework.boot.autoconfigure.ssl.SslPropertiesBundleRegistrar.watchForUpdates(SslPropertiesBundleRegistrar.java:82) ~[spring-boot-autoconfigure-3.4.2.jar:3.4.2]
	... 75 common frames omitted
Caused by: java.io.IOException: '/home/maniax/projects/am-hub/..data/keystpre.p12' is neither a file nor a directory
	at org.springframework.boot.autoconfigure.ssl.FileWatcher$WatcherThread.register(FileWatcher.java:150) ~[spring-boot-autoconfigure-3.4.2.jar:3.4.2]
	at org.springframework.boot.autoconfigure.ssl.FileWatcher.watch(FileWatcher.java:93) ~[spring-boot-autoconfigure-3.4.2.jar:3.4.2]
	... 76 common frames omitted
	
------------------------------------------------------------------
	
**The problem is in the FileWatcher class witch resolves the file wrong and add the wrong absolute path to the resolved file. I my case int should prepend /opt/certs. The corect path is /opt/certs/..data/keystore.p12

image


The logs are from my local machine but in a kubernetes cluster I got the same result, just the path prefix is the path where the jar file executed.**


Can anyone help?

bclozel:
@PetarDimitrovZH See https://github.com/spring-projects/spring-boot/issues/43966 which will be released this week.

PetarDimitrovZH:
@bclozel  Thank you!

PR #43596, opened 2024-12-23
feat: spring-boot.build-image.* should not be read-only

mpalourdio:
Hi!

With Spring Boot 3.4+, it's now possible to build a docker image from an `amd64` host for an `arm64` platform. It works and it's awesome!

But while setting the parameter in the pom, or by command line, ie `mvn clean -Pnative spring-boot:build-image -Dspring-boot.build-image.imagePlatform=linux/arm64`, maven warns :

`Parameter 'imagePlatform' (user property 'spring-boot.build-image.imagePlatform') is read-only, must not be used in configuration`

IMO, the readonly flag should be removed here


snicoll:
> IMO, the readonly flag should be removed here

If it should be removed there, it should be removed pretty much in every parameter above where we're applying the exact same thing.

mpalourdio:
Indeed, i have remove it for all the `spring-boot.build-image.*` parameters. It's convenient to set them via CLI

PR #43603, opened 2024-12-24
DataSourceTransactionManagerAutoConfiguration should not enable DataSourceProperties as it does not use them

quaff:
`DataSourceProperties` bean is not used by `DataSourceTransactionManagerAutoConfiguration`.


philwebb:
Thanks @quaff. I think just to be safe, we'll wait until 3.5 to merge this one. I think the extra `@EnableConfigurationProperties` is harmless in the meantime.

wilkinsona:
I've changed this to an enhancement, primarily so that it will appear in the changelog.

PR #43614, opened 2024-12-26
Make application main class explicit in usage help of SpringApplicationAotProcessor

quaff:
`applicationName` is ambiguous here.


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplicationAotProcessor.java
```diff
@@ -46,12 +46,12 @@ public class SpringApplicationAotProcessor extends ContextAotProcessor {
 
 	/**
 	 * Create a new processor for the specified application and settings.
-	 * @param application the application main class
+	 * @param applicationMainClass the application main class
 	 * @param settings the general AOT processor settings
 	 * @param applicationArgs the arguments to provide to the main method
 	 */
-	public SpringApplicationAotProcessor(Class<?> application, Settings settings, String[] applicationArgs) {
```
snicoll, line 53: The current parameter name is fine, please revert

snicoll (changes requested):

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/SpringApplicationAotProcessor.java
quaff: done.

snicoll (approved):

PR #43619, opened 2024-12-27
Fix assert null checks with dynamic error messages

taewoocode:
This pull request introduces improvements to the assert error message format in the constructors of ReflectiveComponentsClientHttpRequestFactoryBuilder. By using dynamic error messages, it becomes easier to modify and improve the messages in the future.


snicoll:
@taewoocode thanks but the code is fine as it is.

PR #43601, opened 2024-12-24
Documentation still has references to 'layertools'

ngocnhan-tran1996:
Change `layertools` to `tools`
Closes https://github.com/spring-projects/spring-boot/issues/43565

<br>

Fix `@XmlRootElement` javadoc links

Reference
https://docs.spring.io/spring-boot/how-to/spring-mvc.html#howto.spring-mvc.write-xml-rest-service

PR #43622, opened 2024-12-27
Paragraph HTML tags are rendered as is in Maven Plugin reference documentation

ngocnhan-tran1996:
Reference

https://docs.spring.io/spring-boot/3.4-SNAPSHOT/maven-plugin/packaging.html#packaging.repackage-goal.parameter-details.executable


snicoll:
Thanks @ngocnhan-tran1996 but this is exactly the same as https://github.com/spring-projects/spring-boot/pull/43601#issuecomment-2560906916. Please do not submit PRs with unrelated changes. Polish on multiple files are ok but the `<p>` isn't a typo.

PR #43638, opened 2024-12-30
feat: Optimize DatabaseDriver fromJdbcUrl method with caching

minwoo1999:
This PR introduces a cache (URL_PREFIX_CACHE) to store URL prefixes for all DatabaseDriver enums. By precomputing prefixes during initialization, the fromJdbcUrl method avoids redundant calls to getUrlPrefixes, significantly improving performance for repeated lookups.

Caching was chosen to reduce computational overhead while keeping the code maintainable. Future updates to getUrlPrefixes automatically reflect in the cache, requiring no additional changes.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
@@ -19,7 +19,11 @@ package org.springframework.boot.jdbc;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.HashMap;
+import java.util.HashSet;
 import java.util.Locale;
+import java.util.Map;
+import java.util.Set;
 
 import org.springframework.util.Assert;
 import org.springframework.util.StringUtils;
@@ -228,6 +232,14 @@ public enum DatabaseDriver {
 
 	private final String validationQuery;
 
+	private static final Map<DatabaseDriver, Set<String>> URL_PREFIX_CACHE = new HashMap<>();
+
+	static {
+		for (DatabaseDriver driver : values()) {
+			URL_PREFIX_CACHE.put(driver, new HashSet<>(driver.getUrlPrefixes()));
+		}
+	}
+
 	DatabaseDriver(String productName, String driverClassName) {
 		this(productName, driverClassName, null);
 	}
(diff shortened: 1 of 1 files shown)


snicoll:
> By precomputing prefixes during initialization, the fromJdbcUrl method avoids redundant calls to getUrlPrefixes, significantly improving performance for repeated lookups.

What would be the reason for calling this method repeatedly?

minwoo1999:
> > By precomputing prefixes during initialization, the fromJdbcUrl method avoids redundant calls to getUrlPrefixes, significantly improving performance for repeated lookups.
> 
> What would be the reason for calling this method repeatedly?

In my opinion, the fromJdbcUrl method is frequently invoked in applications that dynamically manage database connections. This is especially true in multi-tenant systems or environments with multiple data sources, where determining the correct database driver for each connection is essential.

Therefore, by precomputing and caching URL prefixes in URL_PREFIX_CACHE, the method avoids recalculating them repeatedly, leading to significantly faster and more efficient driver resolution. This optimization is particularly advantageous in high-throughput, scalable environments, ensuring smoother performance when handling large numbers of database interactions.

philwebb:
I don't really see how this will improve the performance that much. Calling `getUrlPrefixes()` should be expensive, and if we did want to optimize it we could make the result a `static final` constant which would save the one call to `List.of(...)`.

@minwoo1999 Have you actually encountered performance issues? Have you benchmarked your changes?

minwoo1999:
> I don't really see how this will improve the performance that much. Calling `getUrlPrefixes()` shouldn't be expensive, and if we did want to optimize it we could make the result a `static final` constant which would save the one call to `List.of(...)`.
> 
> @minwoo1999 Have you actually encountered performance issues? Have you benchmarked your changes?

I haven’t encountered specific performance issues or conducted benchmarks related to this change. However, the fromJdbcUrl method is commonly used in scenarios where database connections need to be dynamically determined at runtime. For instance, in multi-tenant systems, each tenant might have its own database, and the application must resolve the correct driver for each request. Similarly, applications managing multiple data sources rely on this method to handle various JDBC URLs effectively.

In addition, runtime configuration changes, such as switching between development, testing, or production environments, often require dynamic connection updates. In sharded or distributed database setups, this method is also essential for routing queries to the appropriate database instance.

This optimization was made with these scenarios in mind, aiming to make the process more efficient and scalable. While no immediate performance bottleneck was observed, the goal was to proactively improve the method’s efficiency for systems with high database interaction or dynamic configurations.

philwebb:
Thanks for the PR, but without actually encountering performance issues or seeing benchmark results this feels like premature optimization to me.

Issue #21374, opened 2020-05-09
Duplicate HttpMessageConverter


brucelwl:
Duplicate StringHttpMessageConverter
Duplicate MappingJackson2HttpMessageConverter

![image](https://user-images.githubusercontent.com/13413715/81479022-88021480-9253-11ea-987d-d22eab5223d2.png)

bclozel:
Hi @brucelwl 

This behavior is intentional and is documented. See #1293, #2444, #20235 and #15027 for more information on that (and duplicate issues in some cases).

The bottom line is: they're not really duplicate because they might not have the same configuration and one or the other might be used depending on the case.

Thanks!

brucelwl:
But the configuration in ``org.springframework.http.converter.json.MappingJackson2HttpMessageConverter`` is exactly the same

![image](https://user-images.githubusercontent.com/13413715/82751224-5f981f80-9de8-11ea-83e0-34a2e0304a3c.png)

bclozel:
@brucelwl the `ObjectMapper` instance aren't the same. See #1293 as mentioned in my previous comment.

brucelwl:
@bclozel I have roughly understood what you mean, but what should I do if I don't want to use the default ``converter``? I think we should replace the ``converter`` instead of appending the converter before the default to avoid creating redundant objects. The media types to be supported can be customized by developers

In terms of ``StringHttpMessageConverter``, it is difficult to use the second object because their judgment conditions are the same,

hassanrehman01398:
I have the same question. I have two MappingJackson Convertor of same data type. The modules registered in the first convertor are 5 and in the other one are 2. When I am calling my rest template, the second convertor is been called resulting in error.

Failed to evaluate Jackson deserialization for type 
com.fasterxml.jackson.databind.exc.InvalidDefinitionException: Invalid type definition for type 
Argument #0 of constructor com.fasterxml.jackson.annotation.JsonCreator=@com.fasterxml.jackson.annotation.JsonCreator(mode=DEFAULT)} 
has no property name (and is not Injectable): can not use as property-based Creator

PR #43640, opened 2024-12-31
Fix default StringHttpMessageConverter

sonjih00n:
image
https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.server.server.servlet.encoding.charset

Since the default value of `server.servlet.encoding.charset` is UTF-8, StringHttpMessageConverter with UTF-8 encoding is automatically created by the code below.

https://github.com/spring-projects/spring-boot/blob/c1a733130873c712f162d98fac812dde4a9de05a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/http/HttpMessageConvertersAutoConfiguration.java#L79-L86

Therefore, HttpMessageConverters has two StringHttpMessageConverter as shown below.
image

So, StringHttpMessageConverter with ISO-8859-1 encoding is not used.
(Especially because the default value of `server.servlet.encoding.charset` is UTF-8)

Even if the developer sets ISO-8859-1 encoding according to the servlet default specification, There are two StringHttpMessageConverter using ISO-8859-1 encoding as shown below.
```properties
server.servlet.encoding.charset=ISO-8859-1
```
image
So, if there is an additional StringHttpMessageConverter, I would like to change it to not add the default StringHttpMessageConverter.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/http/HttpMessageConverters.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/http/HttpMessageConverters.java
@@ -26,6 +26,7 @@ import java.util.List;
 import java.util.Map;
 
 import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.converter.StringHttpMessageConverter;
 import org.springframework.http.converter.support.AllEncompassingFormHttpMessageConverter;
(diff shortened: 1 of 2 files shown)


bclozel:
Build is failing. Please run the build before submitting PRs.
I think this is a duplicate of #21374 anyway - Spring Framework configures the default instance and Spring Boot contributes an additional one. While this might not be useful for the server-side case, message converters can be used in clients as well.

I'm declining this change as a result.

Issue #37904, opened 2023-10-16
Align versions used in spring-boot-gradle-plugin


philwebb:
Looking at the tests in `/spring-boot-gradle-plugin/src/test/resources/org/springframework/boot/gradle/tasks/bundling` we have several that pull in old dependencies. We should probably try to align versions with the main build.

sharoha:
What is the version that we are targeting in these test dependencies?

scottfrederick:
@sharoha This issue is referring to Gradle build files used in tests for the Gradle plugin. Some of those test files use dependencies with hard-coded versions that aren't up-to-date, as in [this example](https://github.com/spring-projects/spring-boot/blob/2.7.x/spring-boot-project/spring-boot-tools/spring-boot-gradle-plugin/src/test/resources/org/springframework/boot/gradle/tasks/bundling/BootJarIntegrationTests-customLayers.gradle#L33-L37) using Apache Commons and Spring Framework dependencies. 

I don't think we would want to just update those versions to something more current, because they will quickly get out-of-date again. We should try to come up with a way to replace those hard-coded versions with versions supplied by Boot's build system.

PR #43648, opened 2025-01-02
Align versions used in spring-boot-gradle-plugin

JoshuaChen:
#37904 Align versions used in spring-boot-gradle-plugin


wilkinsona:
Thanks for the proposal. These changes seem to be quite a bit more broad than required which makes it harder to review. #37904 is focused on the `spring-boot-gradle-plugin` module and particular the tests in the `org.springframework.boot.gradle.tasks.bundling`. Would you like to update your PR to narrow its scope so that it can be reviewed more easily?

JoshuaChen:
@wilkinsona  hi~ Completed, please review again~

wilkinsona:
Thanks for the updates. Unfortunately, now that I can see the proposed changes more clearly, I'm not keen on the approach that has been taken. For example, I find rather distracting all the fields that have been added to AbstractBootArchiveIntegrationTests and all the properties that have been added to `gradle.properties`. I'm not yet sure what the answer is, but I think we'll need to find an alternative approach here. Thanks anyway.

PR #43659, opened 2025-01-04
Refactor: Use 'source' variable consistently in log formatting

minwoo1999:
This PR refactors the format method of the SimpleFormatter class to replace multiple calls to record.getLoggerName() with the source variable. This change improves code readability and ensures consistent usage of the logger name throughout the formatting process.


philwebb:
Thanks for the suggestion, but I prefer using distinct variable names so the format call is easier to follow. I've actually extracted more in commit 29e9ce17ac158f541208e326026d420baab77c8c

PR #43660, opened 2025-01-04
Added instructions for dash0 to documentation

CodingFabian:
Added instructions on how to use the OLTP Exporter for dash0

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/actuator/metrics.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/actuator/metrics.adoc
@@ -5,6 +5,7 @@ Spring Boot Actuator provides dependency management and auto-configuration for {
 
 - xref:actuator/metrics.adoc#actuator.metrics.export.appoptics[]
 - xref:actuator/metrics.adoc#actuator.metrics.export.atlas[]
+- xref:actuator/metrics.adoc#actuator.metrics.export.dash0[]
 - xref:actuator/metrics.adoc#actuator.metrics.export.datadog[]
 - xref:actuator/metrics.adoc#actuator.metrics.export.dynatrace[]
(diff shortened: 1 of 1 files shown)


philwebb:
Thanks for the suggestion, but I'm not sure that we should update our documentation with OpenTelemetry compatible implementations. Since it's a drop-in replacement, I think our existing OpenTelemetry documentation should be enough.

wilkinsona:
I've opened https://github.com/spring-projects/spring-boot/issues/43727 to update the docs to talk about OTLP rather than OpenTelemetry.

CodingFabian:
The alternative would be to create a dash0 metric registry and exporter. then it gets a place on the documentation page like all the others. right?

wilkinsona:
I can't speak for the Micrometer team but I think that's quite unlikely to happen when the existing OTLP registry serves the same purpose.

CodingFabian:
of course I am poking the bear here, but most of the listed vendors support OTLP too. So apparently having OTLP support doesn't rule out a custom implementation and listing

bclozel:
I guess other vendors ship proprietary data formats in addition to supporting OTLP and that explains the specific registry.
If I understand correctly, dash0 markets itself as fully OpenTelemetry compliant with no proprietary format. Why require a vendor-specific registry then?

If you're looking to promote dash0, I don't think adding a code snippet in our reference documentation is going to help. Right now searching for "dash0 Spring Boot" returns zero relevant result in my search engine.

I did notice [a recent spring-demo app in the dash0 org](https://github.com/dash0hq/spring-demo/tree/main) but it actually uses the `paketo-buildpacks/opentelemetry`, so the OpenTelemetry Java agent to instrument the application. This is not an invalid combination but [it's not the one we support officially and it won't produce the observations documented in the Spring projects](https://docs.spring.io/spring-boot/reference/actuator/observability.html#actuator.observability.opentelemetry). I would suggest promoting this sample and other content before considering this approach.

PR #43664, opened 2025-01-06
Remove duplicate logic in DefaultErrorViewResolver

sonjih00n:
Remove duplicates in DefaultErrorViewResolver constructor


snicoll:
Thanks for the PR but the constructor is like that on purpose. Your change makes two tests fail.

Going forward, please run the build before submitting.

Issue #32111, opened 2022-08-18
Revisit Dev Tools live reload capabilities


vpavic:
Currently, as explained in [LiveReload section](https://docs.spring.io/spring-boot/docs/current/reference/htmlsingle/#using.devtools.livereload) of the reference manual, Dev Tools support for live reload requires use of 3rd party browser extensions. The suggested extensions have been unmaintained for more than a few years now as evident by the last publishing dates and the activity in the appropriate repos:
- https://addons.mozilla.org/en-US/firefox/addon/livereload-web-extension/
- https://chrome.google.com/webstore/detail/livereload/jnihajbhpnppcggbcgedagnkighmdlei
- https://github.com/livereload/livereload-extensions
- https://github.com/twolfson/livereload-extensions

In addition to that, the linked [livereload.com](http://livereload.com/) site itself is down, which apparently isn't a transient issue as [wayback machine](https://web.archive.org/web/20220000000000*/http://livereload.com) was last able to index it on Jun 10th.

With this in mind Spring Boot should IMO revisit live reload capabilities of Dev Tools module. Additionally, some other technologies that I've work with recently offer live reload capabilities without requiring browser extensions at all. Some examples of that are [hugo](https://gohugo.io/commands/hugo_server/) and even [IntelliJ IDEA](https://www.jetbrains.com/help/idea/live-editing.html). Both of these work by appending the script within the served HTML, which contains some code that updates the page when needed. Something like this would both improve the developer experience and also remove the requirement to use (unmaintained) 3rd party browser plugins.

philwebb:
Thanks for the suggestion, I think we should revisit this at some point. There's also https://browsersync.io/docs/gulp which might help.

vpavic:
While the livereload browser extensions aren't maintained, the core [livereload/livereload-js](https://github.com/livereload/livereload-js) project is. I've tried pulling it in as a WebJar and including it in my templates and it just works. I can try putting together a PR that explores this approach, if you don't have anything against it.

philwebb:
That would be useful if you have time, but we're not likely to get to this until after 3.0 has been released so no rush on our side. Perhaps we might be able to bundle livereload-js ourselves so we don't need to use webjars.

vpavic:
> That would be useful if you have time, but we're not likely to get to this until after 3.0 has been released so no rush on our side.

Got it.

> Perhaps we might be able to bundle livereload-js ourselves so we don't need to use webjars.

[...]

Issue #29406, opened 2022-01-14
Include default Dev Tools properties in the reference documentation


vpavic:
At the moment Dev Tools are configuring the following properties:

https://github.com/spring-projects/spring-boot/blob/fe82098d7da53de0547e5cafb38bd170916377ad/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/env/DevToolsPropertyDefaultsPostProcessor.java#L63-L78

Some of those (`server.error.include-*`) impact HTTP responses (either HTML or JSON) returned to the user which I find to be a bit controversial. IMO, something like Dev Tools shouldn't have impact to an application in functional way, which changing responses sort of is. I'd rather see such information being logged instead of returned in responses.

The similar case could be made for enabling H2 console, which most of the time also requires configuring security in order to be accessible.

Additionally, it might be worth considering to also provide more information about which exact properties were enabled. At the moment, only this is logged on startup.

```
2022-01-14 22:22:17.289  INFO 186000 --- [  restartedMain] .e.DevToolsPropertyDefaultsPostProcessor : Devtools property defaults active! Set 'spring.devtools.add-properties' to 'false' to disable
```

Since the list might change over time, it might be a good idea to log exact properties that were set.

philwebb:
We discussed this issue today and we'd like to extend our documentation to list the properties that are set. We think we can do this by extracting them to a `.properties` file that we can also use to generate and `.adoc` file. We think this might be better than logging the properties since there's quite a few of them and it could get annoying.

We think that the `server.error.*` and `spring.h2.console.enabled` properties still make sense to have in a "developer only" setup. They're generally useful and we think they're worth setting by default.

PR #43697, opened 2025-01-06
Add support for LiveReload without browser extensions

vpavic:
This commit improves Dev Tools live reload capabilities by adding support for appending LiveReload.js script to rendered web pages.

See gh-32111

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LiveReloadServletConfiguration.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LiveReloadServletConfiguration.java
@@ -1,0 +1,55 @@
+/*
+ * Copyright 2012-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.devtools.autoconfigure;
+
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnWebApplication.Type;
+import org.springframework.boot.devtools.livereload.LiveReloadScriptInjectingFilter;
+import org.springframework.boot.devtools.restart.RestartScope;
+import org.springframework.context.annotation.Bean;
+import org.springframework.context.annotation.Configuration;
+import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistration;
+import org.springframework.web.servlet.config.annotation.ResourceHandlerRegistry;
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+/**
+ * Servlet-specific local LiveReload configuration.
+ *
+ * @author Vedran Pavic
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnWebApplication(type = Type.SERVLET)
+class LiveReloadServletConfiguration {
+
+	@Bean
+	@RestartScope
+	LiveReloadScriptInjectingFilter liveReloadScriptInjectingFilter(DevToolsProperties properties) {
+		return new LiveReloadScriptInjectingFilter(properties.getLivereload().getPort());
+	}
+
+	@Configuration(proxyBeanMethods = false)
+	static class LiveReloadResourcesConfiguration implements WebMvcConfigurer {
+
+		@Override
+		public void addResourceHandlers(ResourceHandlerRegistry registry) {
+			ResourceHandlerRegistration registration = registry.addResourceHandler("/livereload.js");
+			registration.addResourceLocations("classpath:/org/springframework/boot/devtools/livereload/");
+		}
+
+	}
+
+}
--- a/spring-boot-project/spring-boot-devtools/src/main/resources/org/springframework/boot/devtools/livereload/livereload.js
+++ b/spring-boot-project/spring-boot-devtools/src/main/resources/org/springframework/boot/devtools/livereload/livereload.js
@@ -1,27 +1,2459 @@
-(function() {
-var __customevents = {}, __protocol = {}, __connector = {}, __timer = {}, __options = {}, __reloader = {}, __livereload = {}, __less = {}, __startup = {};
+(function(){function r(e,n,t){function o(i,f){if(!n[i]){if(!e[i]){var c="function"==typeof require&&require;if(!f&&c)return c(i,!0);if(u)return u(i,!0);var a=new Error("Cannot find module '"+i+"'");throw a.code="MODULE_NOT_FOUND",a}var p=n[i]={exports:{}};e[i][0].call(p.exports,function(r){var n=e[i][1][r];return o(n||r)},p,p.exports,r,e,n,t)}return n[i].exports}for(var u="function"==typeof require&&require,i=0;i<t.length;i++)o(t[i]);return o}return r})()({1:[function(require,module,exports){
+module.exports = function (it) {
+  if (typeof it != 'function') throw TypeError(it + ' is not a function!');
+  return it;
+};
 
-// customevents
-var CustomEvents;
-CustomEvents = {
-  bind: function(element, eventName, handler) {
+},{}],2:[function(require,module,exports){
+// 22.1.3.31 Array.prototype[@@unscopables]
+var UNSCOPABLES = require('./_wks')('unscopables');
+var ArrayProto = Array.prototype;
+if (ArrayProto[UNSCOPABLES] == undefined) require('./_hide')(ArrayProto, UNSCOPABLES, {});
+module.exports = function (key) {
+  ArrayProto[UNSCOPABLES][key] = true;
+};
+
+},{"./_hide":27,"./_wks":81}],3:[function(require,module,exports){
+'use strict';
+var at = require('./_string-at')(true);
+
+ // `AdvanceStringIndex` abstract operation
+// https://tc39.github.io/ecma262/#sec-advancestringindex
+module.exports = function (S, index, unicode) {
+  return index + (unicode ? at(S, index).length : 1);
+};
+
+},{"./_string-at":68}],4:[function(require,module,exports){
+var isObject = require('./_is-object');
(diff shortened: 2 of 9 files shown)


spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/autoconfigure/LiveReloadServletConfiguration.java
```diff
@@ -0,0 +1,55 @@
...
+import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
+
+/**
+ * Servlet-specific local LiveReload configuration.
+ *
+ * @author Vedran Pavic
+ */
+@Configuration(proxyBeanMethods = false)
+@ConditionalOnWebApplication(type = Type.SERVLET)
+class LiveReloadServletConfiguration {
```
vpavic, line 36: This configuration contains Servlet-specific bits needed to set up everything. If something like this is a viable approach, I assume WebFlux-specific equivalent is needed as well (I've never used Dev Tools with WebFlux so I'm unaware if there are any limitations).

spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/livereload/LiveReloadScriptFilter.java
```diff
@@ -0,0 +1,53 @@
...
+ *
+ * @author Vedran Pavic
+ * @since 3.5.0
+ */
+public class LiveReloadScriptFilter extends OncePerRequestFilter {
+
+	private final String scriptSnippet;
+
+	public LiveReloadScriptFilter(int liveReloadPort) {
+		this.scriptSnippet = String.format("<script src=\"/livereload.js?port=%d\"></script>", liveReloadPort);
```
vpavic, line 40: Perhaps something to consider is embedding the script in what LiveReload.js readme refers to as _slightly smarter way_, see https://github.com/livereload/livereload-js?tab=readme-ov-file#using-livereloadjs.

spring-boot-project/spring-boot-devtools/src/main/resources/org/springframework/boot/devtools/livereload/livereload.js
vpavic, line 1: This updates the existing `livereload.js` to latest version from https://github.com/livereload/livereload-js.

spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/using/devtools.adoc
```diff
@@ -284,8 +284,6 @@ If you find such a problem, you need to request a fix with the original authors.
 == LiveReload
 
 The `spring-boot-devtools` module includes an embedded LiveReload server that can be used to trigger a browser refresh when a resource is changed.
-LiveReload browser extensions are freely available for Chrome, Firefox and Safari.
```
vpavic, line 287: Not sure whether it's needed to explicitly call out here that browser extensions are no longer needed and Spring Boot provides everything needed.

philwebb:
Unfortunately I think the filter might generate invalid HTML. If I'm reading the code correctly then it inserts the `<script>` tag after `</html>`. That might work for now, but according to [this stackoverflow.com post](https://stackoverflow.com/questions/3037725/is-it-wrong-to-place-the-script-tag-after-the-body-tag) browsers might ban it in the future. We also have the issue of this not working outside of Spring MVC due to the use of a `ResourceHandlerRegistration`.

I guess the filter could parse the HTML and try to insert the script in the correct location. The problem there is that we might need to parse all responses just to tell if we're getting HTML. On top of that, the WebFlux side is also going to be pretty hard to address.

Flagging for team attention in case anyone else has any bright ideas.

vpavic:
> Unfortunately I think the filter might generate invalid HTML. If I'm reading the code correctly then it inserts the `<script>` tag after `</html>`.

That's true. I didn't (yet) put much effort into making the filter implementation robust as I wanted to pick up some feedback on the high-level approach first.

Looking into what some other web frameworks do this shouldn't be too hard - for example Hugo injects the `<script>` tag into `<head>` - see [here](https://github.com/gohugoio/hugo/blob/master/transform/livereloadinject/livereloadinject.go). Apparently it used to inject it just before closing `</body>` tag but made the change in https://github.com/gohugoio/hugo/issues/6821.

> The problem there is that we might need to parse all responses just to tell if we're getting HTML.

The filter already takes response content type into consideration (requiring it to be `text/html` compatible) before doing anything with the response. If that condition is satisfied, only then we could parse the response.

philwebb:
> The filter already takes response content type into consideration (requiring it to be text/html compatible) before doing anything with the response. If that condition is satisfied, only then we could parse the response.

I think it can't determine the content type until it's called the filter chain. That means we probably need to capture content for all requests in case they're HTML and need the script tag injected.

vpavic:
Yes, but that (wrapping) is different than parsing every response - that can still only be done if the response indeed is `text/html` compatible.

I've experimented with my [PoC project](https://github.com/vpavic/poc-spring-boot-livereload) a bit and got something (again very simple, not robust) working using `ContentCachingResponseWrapper` that places LiveReload `script` tag inside `head` tag.

I've also pushed those changes here.

vpavic:
@philwebb could you share the reason this is marked as on-hold? The concerns you expressed in https://github.com/spring-projects/spring-boot/pull/43697#issuecomment-2635212601 are addressed, with the exception of WebFlux support. I haven't looked at that part yet but if that's the only thing blocking this from moving forward I'll try to find some time to do it.

philwebb:
Sorry for the delay in replying @vpavic, I've been at Devnexus and then focusing on another piece of work.

We have quite serious reservations about attempting to automatically insert the live reload scripts into applications using a filter. We think that cost of intercepting and buffering every request, along with the risk of rewriting HTML on the fly is ultimately too high.

We're now thinking that if we are to offer live-reload support, we should perhaps offer it as helpers that users can insert into their templates directly. Whilst this won't be quite as seamless an experience, we think it might be safer and more aligned with the way that other frameworks handle live reload.

I'm afraid we're going to decline this PR and leave #32111 open for a little while longer. This is a problem that we want to solve, but currently our focus needs to be on the Spring Boot 4.0 work.

I'm sorry for any wasted effort, and I hope we can eventually find a solution that makes everybody happy.

vpavic:
The approach taken in this PR is I believe equivalent to one Hugo and Jekyll - two of the most popular static site generators. If such approach is not problematic for them, could you maybe clarify a bit more what are these _quite serious reservations_?

> cost of intercepting and buffering every request

The implementation proposed here leverages `ContentCachingResponseWrapper` which is also used by Framework's `ShallowEtagHeaderFilter` - a filter that's actually used in production and quite frequently, unlike the filter proposed by this PR, that's used only in development time. If that aspect of implementation is what you're referring to as _buffering every request_, why is it such a concern here and not for `ShallowEtagHeaderFilter`?

Semi-related remark regarding Spring Boot's Livereload.js support: some static site generators, for example Jekyll, don't enable Livereload by default when development server is started - rather they require developers to explicitly enable it (using `jekyll serve --livereload` in Jekyll's case). Maybe Spring Boot could also consider taking such approach, especially since a lot of Spring Boot use cases don't include server-side rendered UIs? I mentioning that because such approach could perhaps alleviate some of the concerns you're having.

philwebb:
> could you maybe clarify a bit more what are these quite serious reservations

Our main concern is that we're changing the developed application behind the scenes, potentially without the user really being aware that it's going on. The filter needs to inject itself into the HTML, which means we either need to parse the HTML or use a regex solution as you've suggested in the PR. Either way, it feels a little risky and we think that having the user opt-in with the helper methods would be safer.

>  what you're referring to as buffering every request, why is it such a concern here and not for `ShallowEtagHeaderFilter`

It is a concern for `ShallowEtagHeaderFilter`, so much so that Framework spent significant time developing the resource abstraction to support cache busting. You can see an example in https://github.com/spring-projects/spring-framework/issues/33256#issuecomment-2243469954 where the `ShallowEtagHeaderFilter` is causing the user issues.

At least the `ShallowEtagHeaderFilter` is opt-in and has a method to disable it for a specific request if you want to. The live reload filter would be transparent, and could break an app when devtools is active that would otherwise run fine.

> explicitly enable it (using jekyll serve --livereload in Jekyll's case). Maybe Spring Boot could also consider taking such approach

Yeah, that's a nice idea. The filter would be a lot less concerning if users opt-in. I see it then being something similar to the helper utilities. We can also put a big warning on the property to warn folks not to use it if they're sending large amounts of data.

Another though I had was a more intelligent version of `ContentCachingResponseWrapper` that bails if it starts to cache too much data. That might be a good Framework feature.

vpavic:
Thanks for sharing additional details.

> Our main concern is that we're changing the developed application behind the scenes

I agree with this, but in many ways this is already the case with DevTools - I've actually expressed similar concerns earlier in https://github.com/spring-projects/spring-boot/issues/29406.

I remember that Framework issue you linked, though it's a bit extreme example - dynamically generated responses (JSON, HTML) will very rarely be large enough for this to be an issue in practice. And cache busting mechanism IMO targets responses that are of completely different nature (static).

Would you reconsider this PR for 4.0 if a separate change flips the default value of `spring.devtools.livereload.enabled` to `false`? I also like the idea to enhance `ContentCachingResponseWrapper` by allowing to set some kind of a threshold, I'll try to look into that in the meanwhile.

philwebb:
> dynamically generated responses (JSON, HTML) will very rarely be large enough for this to be an issue in practice.

Yes, they probably won't be, but the problem is we need to inspect/cache every request including those that aren't JSON/HTML. The threshold idea would certainly help with that.

> Would you reconsider this PR for 4.0 if a separate change flips the default value of spring.devtools.livereload.enabled to `false`.

I think if we could flip default and have the threshold change we'd be much more likely to accept the change. I can't make any promises for 4.0 just due to the amount of other work we need to get into that release.

vpavic:
> the problem is we need to inspect/cache every request including those that aren't JSON/HTML

Similar to what `LiveReloadScriptInjectingFilter#shouldInjectScript` does before attempting to inject the script element, filter could also do a similar check based on the requested content type and wrap the response only if HTML has been requested.

> 4k lines of unmaintained javascript is a liability for our project

In https://github.com/spring-projects/spring-boot/issues/32111#issuecomment-1247937735 I propsed to pull it in as a WebJar dependency (potentially shaded) but in the subsequent comment it was suggested to just include it in the project as the current version is.

vpavic:
I've pushed some updates here:

- this is now based on changes from #47387
- response wrapping is now performed only if `text/html` compatible content type has been requested
- LiveReload.js is now pulled in as a WebJar dependency - this is not intended to be a final solution (as mentioned before, perhaps shading it would be an option)

I plan to look into add cache size threshold of some sort to `ContentCachingResponseWrapper` as the next step.

GitHub doesn't render the diff anymore (I assume because PR has been closed for some while) so it can be seen here:

- https://github.com/vpavic/spring-boot/compare/livereload-default-off..gh-32111

Since most of the feedback discussed has been addressed, could you consider re-opening it?

PR #43739, opened 2025-01-09
Deprecate ConditionOutcome.inverse()

quaff:
1. It's not used anywhere.
2. The condition message is not applicable for reusing if match is reversed in practice.
3. The method is not doing much, we should call the constructor directly since it is public.
4. It's better to define it as instance method if it indeed required.


philwebb:
Thanks @quaff!

PR #43774, opened 2025-01-10
Prevent redis pool to initialize in unit test

hezean:
As https://github.com/spring-projects/spring-boot/commit/52d419a26aaebc290317c990b61c96532f60fc80 bumps spring-data bom to `2024.1.2-SNAPSHOT`, an unconditional [preparePool](https://github.com/apache/commons-pool/blob/e5c44f5184a55a58fef4a1efec8124d162a348bd/src/main/java/org/apache/commons/pool2/impl/GenericObjectPool.java#L978-L983) invocation is added to the start lifecycle of JedisConnectionFactory (see: https://github.com/spring-projects/spring-data-redis/commit/2bcdd36a368d68e5ecffc204294c890556017079). That is, with `min-idle` set to 1, the pool will attempt to create a socket over the dummy domain `foo`, failing the bean creation.

Related to:
- https://github.com/spring-projects/spring-boot/actions/runs/12710459852/job/35433379770?pr=43771#step:3:4925
- https://github.com/spring-projects/spring-boot/actions/runs/12716605691/job/35451295161?pr=43773#step:3:4735


snicoll:
@hezean thanks very much.

Issue #43431, opened 2024-12-06
Align TestRestTemplate default redirect settings to be the same as RestTemplate


philwebb:
The current situation is quite confusing. We use the defaults for the underlying library (usually follow redirects), unless you use HttpComponents. At that point we stop following redirects unless you create the `TestRestTemplate` with `HttpClientOption.ENABLE_REDIRECTS`.

thecooldrop:
Hello @philwebb if you would be available to explain what you mean by second sentence in a bit more detail I would like to start working on this issue.

philwebb:
Thanks for the offer @thecooldrop, but I think this one might be better handled by the team. We need to do some design work before and have some additional discussion before come up with the direction we want to take.

bananer:
Seems like the documentation also needs to be updated: https://docs.spring.io/spring-boot/reference/testing/test-utilities.html#testing.utilities.test-rest-template

> If you do use Apache’s HTTP client, some additional test-friendly features are enabled:
> - Redirects are not followed (so you can assert the response location).
> - Cookies are ignored (so the template is stateless).

wilkinsona:
Thanks, @bananer. I've opened https://github.com/spring-projects/spring-boot/issues/45842.

israelrios:
The javadoc in TestRestTemplate needs to be updated too:

> A TestRestTemplate can optionally carry Basic authentication headers. If Apache Http Client 4.3.2 or better is available (recommended) it will be used as the client, and by default configured to ignore cookies and **redirects**.

wilkinsona:
Thanks, @israelrios. I've opened #47474.

PR #43778, opened 2025-01-11
Change RestTemplateBuilder redirections at HttpComponentsClientHttpRequestFactory to FOLLOW_WHEN_POSSIBLE

panic08:
Our main problem is that `RestTemplate` has a default redirect status of `FOLLOW_WHEN_POSSIBLE`. When we build `RestTemplateBuilder` and want to use Apache HttpComponents as a basis, for some reason we check `httpClientOptions` for `HttpClientOption.ENABLE_REDIRECTS`, otherwise we set it to false, which is wrong.

The solution to this problem I saw was to set `builder.redirects(Redirects.FOLLOW_WHEN_POSSIBLE);` the default `if (httpClientOptions != null)`.

Closes: #43431

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/bom/CheckLinks.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/bom/CheckLinks.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2024 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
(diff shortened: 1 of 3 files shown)


snicoll:
@panic thanks for the PR but it contains unrelated changes. Looking at the actual change, fixing #43431 is going to require more work (testing, and reviewing the consistency).

PR #43789, opened 2025-01-13
Polish ProcessInfo

nosan:
1. Update the Copyright header 
2. Add check if a class is present before `Class.forName(...)`
3. Remove **public** modifier from the `VirtualThreadsInfo` constructor (align with other nested classes)


mhalbritter:
Thanks @nosan !

PR #43790, opened 2025-01-13
Docker Compose support for ClickHouse does not allow an empty password when ALLOW_EMPTY_PASSWORD=yes

hezean:
Adds `yes` as a positive value for `ALLOW_EMPTY_PASSWORD` in [bitnami/clickhouse](https://hub.docker.com/r/bitnami/clickhouse).

### DIFF
--- a/spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/service/connection/clickhouse/ClickHouseEnvironment.java
+++ b/spring-boot-project/spring-boot-docker-compose/src/main/java/org/springframework/boot/docker/compose/service/connection/clickhouse/ClickHouseEnvironment.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2024 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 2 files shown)


hezean:
Actually I'm wondering if simply using `continsKey` would make the code more consistent with others like `bitnami/mysql` and `bitnami/mariadb`.

> I'd prefer that we just consistently look for `ALLOW_EMPTY_PASSWORD` being set to any value and allow Bitnami to do the actual checking. That isolates Boot from any changes to what Bitnami supports and/or documents in the future. This isolation is important as Boot doesn't control the version of any Bitnami container that's used so we don't control when a user may try to use a container with different rules for the value of `ALLOW_EMPTY_PASSWORD`.

_Originally posted by @wilkinsona in https://github.com/spring-projects/spring-boot/issues/43771#issuecomment-2586852282_
            
But for the below existing test, we can either: delete it, or add a check to prevent breaking changes, like

```java
boolean allowEmpty = env.containsKey("ALLOW_EMPTY_PASSWORD") && !"false".equalsIgnoreCase(env.get("ALLOW_EMPTY_PASSWORD"))
```

https://github.com/spring-projects/spring-boot/blob/2015faef95d94a08325a73266f6c836b898b8f53/spring-boot-project/spring-boot-docker-compose/src/test/java/org/springframework/boot/docker/compose/service/connection/clickhouse/ClickHouseEnvironmentTests.java#L58-L63

wilkinsona:
> Actually I'm wondering if simply using continsKey would make the code more consistent with others like bitnami/mysql and bitnami/mariadb.

Yes, please. I think that's what we should do.

> But for the below existing test, we can either: delete it, or add a check to prevent breaking changes

I think we should just delete it for consistency with the other tests.

wilkinsona:
Thanks very much, @hezean.

Issue #31995, opened 2022-08-05
Resolve JDBC driver class name for AWS drivers


maciejwalkowiak:
Spring Boot could resolve JDBC driver class names for AWS variants of MySQL and PostgreSQL drivers:

- https://github.com/awslabs/aws-mysql-jdbc (production ready) - for `jdbc:mysql:aws:` prefix 
- https://github.com/awslabs/aws-postgresql-jdbc (in developer preview) - for `jdbc:postgresql:aws:` prefix

If you believe this change makes sense, I would be happy to contribute a PR.

wilkinsona:
Thanks for the suggestion, @maciejwalkowiak. This could be a nice addition, particularly for `aws-mysql-jdbc`. I think it may be better to wait for `jdbc:postgresql:aws` to stabilise before we add support for it. A PR that adds support for MySQL would be most welcome. Thanks.

maciejwalkowiak:
I noticed few days ago there is a new JDBC driver related project from AWS [in progress](https://github.com/awslabs/aws-postgresql-jdbc/pull/68/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5R8) (https://github.com/awslabs/aws-advanced-jdbc-wrapper - 404 for now) - not sure if it only replaces Postgres driver or as the name suggests will cover more database types. Perhaps @susanmdouglas-aws or @marksailes can shine some light? So that we don't integrate driver that gets deprecated in upcoming months.

wilkinsona:
I'd noticed that yesterday and got the same 404 but hadn't taken the time to investigate. Thanks for doing that, @maciejwalkowiak. I agree that we should hold off until the role of the new JDBC driver becomes clear.

susanmdouglas-aws:
Hi - Can I recommend you reach out to @adalevinAWS for more information?

adalevinAWS:
I'll update here once we have more information to share. I see you added the on-hold label for now, though. Thank you.

vpavic:
Release `1.0.0` of `software.amazon.jdbc:aws-advanced-jdbc-wrapper` (hosted on https://github.com/awslabs/aws-advanced-jdbc-wrapper) has apparently been published in early October.

With that in mind, maybe this could be revisited?

wilkinsona:
@adalevinAWS any updates on your side?

vpavic:
[...]

PR #43812, opened 2025-01-14
Add support for detecting AWS Advanced JDBC Wrapper to DatabaseDriver

vpavic:
This commit adds entry for AWS Advanced JDBC Wrapper to `DatabaseDriver` enum.

See gh-31995

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
@@ -19,6 +19,7 @@ package org.springframework.boot.jdbc;
 import java.util.Arrays;
 import java.util.Collection;
 import java.util.Collections;
+import java.util.List;
 import java.util.Locale;
 
 import org.springframework.util.Assert;
@@ -216,6 +217,19 @@ public enum DatabaseDriver {
 		@Override
 		protected Collection<String> getUrlPrefixes() {
 			return Arrays.asList("ch", "clickhouse");
+		}
+
+	},
+
+	/**
+	 * AWS Advanced JDBC Wrapper.
+	 * @since 3.5.0
+	 */
+	AWS_JDBC_WRAPPER(null, "software.amazon.jdbc.Driver") {
+
+		@Override
+		protected Collection<String> getUrlPrefixes() {
+			return List.of("aws-wrapper");
 		}
 
 	};
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchDataSourceScriptDatabaseInitializerTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchDataSourceScriptDatabaseInitializerTests.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2024 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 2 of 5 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
```diff
@@ -218,6 +219,19 @@ protected Collection<String> getUrlPrefixes() {
...
+
+	/**
+	 * AWS Advanced JDBC Wrapper.
+	 * @since 3.5.0
+	 */
+	AWS_JDBC_WRAPPER(null, "software.amazon.jdbc.Driver") {
+
+		@Override
+		protected Collection<String> getUrlPrefixes() {
+			return List.of("aws-wrapper");
```
vpavic, line 232: I've used `List#of` as it seems to be the simplest option, but note that implementations of `#getUrlPrefixes` in this enum are quite inconsistent - there are usages of `Arrays#asList` and `Collections#singleton`.

Perhaps this could be aligned in a separate issue to use `List#of` consistently (`first-timers-only` candidate?).

wilkinsona:
Thanks, @vpavic. In isolation, I wonder how useful this will be. Looking at how `DatabaseDriver` is used, being able to detect AWS's wrapper may not be enough in some cases. Take Flyway's auto-configuration for example. As you may remember as I think you contributed it, it uses `DatabaseDriver` to figure out which vendor-specific migrations to use. In that situation, I think we'd need to detect the driver that's being wrapped rather than the wrapper. I think the same applies to the [basic script-based initialization](https://docs.spring.io/spring-boot/how-to/data-initialization.html#howto.data-initialization.using-basic-sql-scripts) and the like as well.

vpavic:
Thanks for the feedback.

Motivation here is to simplify usage of AWS Advanced JDBC Wrapper by not having to specify `spring.datasource.driver-class-name` property. Depending on the environment (and the exact database used), sometimes we need to use plain PostgreSQL JDBC driver and other times AWS wrapper - it would be very useful to do that switch only by changing a single configuration property.

I understand your concerns, and have refreshed my memory a bit on support for vendor-specific Flyway migrations (yes, I contributed that quite some while ago :slightly_smiling_face:) and I believe your comment aims at this usage of `DatabaseDriver`:
https://github.com/spring-projects/spring-boot/blob/269c7612e000d0cc0a1fcd62ec21f7edc29f3327/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/flyway/FlywayAutoConfiguration.java#L391-L392

This indeed doesn't play nice with wrapper drivers but such use cases could be updated to rely on database product name instead. I can contribute that if you agree with that direction.

I can see that such approach is already used in `PlatformPlaceholderDatabaseDriverResolver`:
https://github.com/spring-projects/spring-boot/blob/269c7612e000d0cc0a1fcd62ec21f7edc29f3327/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/init/PlatformPlaceholderDatabaseDriverResolver.java#L139-L141

In any case, I think these concerns are not new since Testcontainers JDBC driver also is of similar nature as it wraps the underlying JDBC driver and requires inserting its prefix into JDBC URL.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DatabaseDriver.java
wilkinsona: They seem consistent to me, using `Collections#singleton` when there's a single prefix and `Arrays#asList` when there a multiple prefixes. I'll polish to use `Collections#singleton` when merging.

wilkinsona:
Thanks very much, @vpavic.

PR #43837, opened 2025-01-15
Change to List.of rather than Arrays.asList because it is a List whose values ​​do not change.

dukbong:
Change to List.of rather than Arrays.asList because it is a List whose values ​​do not change.


snicoll:
@dukbong thanks for the PR but we prefer to focus on non-cosmetic changes for external contributions. In this case there are many other use cases of `Arrays.asList` that we should revisit if we decided to change this line.

philwebb:
See also https://github.com/spring-projects/spring-boot/issues/28174#issuecomment-1204242023 as to why we didn't do a blanket conversion already.

PR #43838, opened 2025-01-15
Change to List.of rather than Arrays.asList because it is a List whose values ​​do not change.

dukbong:
Change to List.of rather than Arrays.asList because it is a List whose values ​​do not change.


wilkinsona:
Duplicates #43837.

PR #43876, opened 2025-01-18
Polish

arefbehboudi:
- Variables that were instantiated only once were made final.
- Methods whose return values were not used were changed to void.

### DIFF
--- a/buildSrc/src/main/java/org/springframework/boot/build/DeployedPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/DeployedPlugin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -38,7 +38,6 @@ public class DeployedPlugin implements Plugin<Project> {
 	public static final String GENERATE_POM_TASK_NAME = "generatePomFileForMavenPublication";
 
 	@Override
-	@SuppressWarnings("deprecation")
 	public void apply(Project project) {
 		project.getPlugins().apply(MavenPublishPlugin.class);
 		project.getPlugins().apply(MavenRepositoryPlugin.class);
--- a/buildSrc/src/main/java/org/springframework/boot/build/DeployedPlugin.java
+++ b/buildSrc/src/main/java/org/springframework/boot/build/DeployedPlugin.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2025 the original author or authors.
+ * Copyright 2012-2023 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


buildSrc/src/main/java/org/springframework/boot/build/context/properties/Asciidoc.java
```diff
@@ -29,15 +29,14 @@ class Asciidoc {
 		this.content = new StringBuilder();
 	}
 
-	Asciidoc appendWithHardLineBreaks(Object... items) {
+	void appendWithHardLineBreaks(Object... items) {
```
mhalbritter, line 32: I don't agree with this. Before it was a fluent API, now it's not anymore.

buildSrc/src/main/java/org/springframework/boot/build/context/properties/Asciidoc.java
```diff
@@ -29,15 +29,14 @@ class Asciidoc {
...
 		for (Object item : items) {
 			appendln("`+", item, "+` +");
 		}
-		return this;
 	}
 
-	Asciidoc appendln(Object... items) {
-		return append(items).newLine();
+	void appendln(Object... items) {
```
mhalbritter, line 38: Same here.

buildSrc/src/main/java/org/springframework/boot/build/context/properties/Asciidoc.java
```diff
@@ -47,8 +46,8 @@ Asciidoc append(Object... items) {
 		return this;
 	}
 
-	Asciidoc newLine() {
-		return append(System.lineSeparator());
+	void newLine() {
```
mhalbritter, line 49: And here.

mhalbritter (commented):
Thanks for the PR. I've added some comments.

arefbehboudi:
@mhalbritter Thank you for your feedback.
I have updated the pull request.

buildSrc/src/main/java/org/springframework/boot/build/DeployedPlugin.java
```diff
@@ -38,7 +38,6 @@ public class DeployedPlugin implements Plugin<Project> {
 	public static final String GENERATE_POM_TASK_NAME = "generatePomFileForMavenPublication";
 
 	@Override
-	@SuppressWarnings("deprecation")
```
wilkinsona, line 41: Why way this removed? The call to `isEnabled()` on line 47 is deprecated so I think the warning still needs to be suppressed.

arefbehboudi: The `AbstractTask` class is deprecated, but `isEnabled()` is not. 

Should we still suppress it?

wilkinsona: Yes, please, otherwise there'll be a compiler warning (in Eclipse at least).

buildSrc/src/main/java/org/springframework/boot/build/bom/Library.java
```diff
@@ -541,7 +541,7 @@ public String toString() {
 
 	public record Link(String rootName, Function<LibraryVersion, String> factory, List<String> packages) {
 
-		private static final Pattern PACKAGE_EXPAND = Pattern.compile("^(.*)\\[(.*)\\]$");
+		private static final Pattern PACKAGE_EXPAND = Pattern.compile("^(.*)\\[(.*)]$");
```
wilkinsona, line 544: We're still escaping the opening `[` but the closing `]` is no longer escaped. I think the parser may cope with that but I find the expression harder to read as you have to know that the parser treats the `]` as a plain `]` rather than closing a never-opened character set. Please revert.

wilkinsona (approved):

wilkinsona:
Thank you, @arefbehboudi.

Issue #43442, opened 2024-12-09
springboot3.4 start error: java.lang.ClassNotFoundException: com.google.gson.Strictness


liufengvling:
pom.xml
```
<?xml version="1.0" encoding="UTF-8"?>
<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
	xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 https://maven.apache.org/xsd/maven-4.0.0.xsd">
	<modelVersion>4.0.0</modelVersion>
	<parent>
		<groupId>org.springframework.boot</groupId>
		<artifactId>spring-boot-starter-parent</artifactId>
		<version>3.4.0</version>
		<relativePath/> <!-- lookup parent from repository -->
	</parent>
	<groupId>com.example</groupId>
	<artifactId>demo</artifactId>
	<version>0.0.1-SNAPSHOT</version>
	<name>demo</name>
	<description>Demo project for Spring Boot</description>

	<properties>
		<java.version>21</java.version>
	</properties>
	<dependencies>
		<dependency>
			<groupId>org.springframework.boot</groupId>
			<artifactId>spring-boot-starter-web</artifactId>
		</dependency>
	</dependencies>
</project>

```

start Application

```
@SpringBootApplication
public class DemoApplication {

	public static void main(String[] args) {
		SpringApplication.run(DemoApplication.class, args);
	}

}
```

ERROR:
```

Exception in thread "main" java.lang.IllegalStateException: java.lang.NoClassDefFoundError: com/google/gson/Strictness
	at org.springframework.boot.SpringApplication.handleRunFailure(SpringApplication.java:823)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:328)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1361)
	at org.springframework.boot.SpringApplication.run(SpringApplication.java:1350)
	at com.example.demo.DemoApplication.main(DemoApplication.java:10)
Caused by: java.lang.NoClassDefFoundError: com/google/gson/Strictness
	at java.base/java.lang.Class.getDeclaredFields0(Native Method)
	at java.base/java.lang.Class.privateGetDeclaredFields(Class.java:3473)
	at java.base/java.lang.Class.getDeclaredField(Class.java:2780)
[...]

PR #43877, opened 2025-01-18
Fix Javadoc since for GsonProperties.Strictness

izeye:
This PR fixes the Javadoc `@since` tag for the `GsonProperties.Strictness`.

See gh-43442


snicoll:
Thanks again, @izeye.

PR #43879, opened 2025-01-19
Polish

izeye:
This PR polishes a bit.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -1926,7 +1926,7 @@
     },
     {
       "name": "management.metrics.graphql.autotime.enabled",
-      "description": "Whether to automatically time web client requests.",
+      "description": "Whether to automatically time GraphQL requests.",
       "defaultValue": true,
       "deprecation": {
         "level": "error",
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
@@ -1926,7 +1926,7 @@
     },
     {
       "name": "management.metrics.graphql.autotime.enabled",
-      "description": "Whether to automatically time GraphQL requests.",
+      "description": "Whether to automatically time web client requests.",
       "defaultValue": true,
       "deprecation": {
         "level": "error",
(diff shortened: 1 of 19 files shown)


spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/dev-services.adoc
```diff
@@ -402,7 +402,7 @@ You can also add javadoc:org.springframework.test.context.DynamicPropertySource[
 
 When using devtools, you can annotate beans and bean methods with javadoc:org.springframework.boot.devtools.restart.RestartScope[format=annotation].
 Such beans won't be recreated when the devtools restart the application.
-This is especially useful for Testcontainer javadoc:org.testcontainers.containers.Container[] beans, as they keep their state despite the application restart.
+This is especially useful for Testcontainers javadoc:org.testcontainers.containers.Container[] beans, as they keep their state despite the application restart.
```
wilkinsona, line 405: I think this was intentionally singular. If it's to change, I think it should be changed to "container".

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/resources/META-INF/additional-spring-configuration-metadata.json
```diff
@@ -1926,7 +1926,7 @@
     },
     {
       "name": "management.metrics.graphql.autotime.enabled",
-      "description": "Whether to automatically time web client requests.",
+      "description": "Whether to automatically time GraphQL requests.",
```
wilkinsona, line 1929: I've opened https://github.com/spring-projects/spring-boot/issues/43901 for this as I think it warrants a separate issue that'll appear in the changelog.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/FilteringStatusListener.java
```diff
@@ -35,7 +35,7 @@ class FilteringStatusListener extends ContextAwareBase implements StatusListener
 
 	/**
 	 * Creates a new {@link FilteringStatusListener}.
-	 * @param delegate the {@link StatusListener} delegate to
+	 * @param delegate the {@link StatusListener} to delegate
```
wilkinsona, line 38: I think this should be "to delegate to".

wilkinsona (changes requested):

spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/dev-services.adoc
izeye: @wilkinsona How about "Testcontainers'" or just removing "Testcontainer"?

"container Container beans" doesn't sound right to me.

wilkinsona: Oops. Yeah, I think we can just remove it as we're already in a section about Testcontainers. I think "for Container beans" reads better and there's enough context to know that it's a Testcontainers container.

izeye:
@wilkinsona Thanks for the feedback!

I applied your suggestions in cbb44e1.

izeye:
Sorry, I forgot to revert the change on the description of the `management.metrics.graphql.autotime.enabled` property. I've done it in cadf0e0.

wilkinsona:
Thanks very much, @izeye!

PR #43917, opened 2025-01-22
Enforce using @ConfigurationProperties("xxx") instead of @ConfigurationProperties(prefix = "xxx")

quaff:


philwebb:
Thanks @quaff!

PR #43918, opened 2025-01-22
Polish usage of Collection in @ConfigurationProperties classes

quaff:
1. It's safe to use immutable Collection object as default value of field.
2. It's unnecessary to create new Collection object in setter method.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/MetricsProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2024 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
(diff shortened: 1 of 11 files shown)


wilkinsona:
Thanks for the PR but I don't think we should make this change.

When binding properties, the binder will first try to mutate an existing collection by clearing it and then adding the new contents:

https://github.com/spring-projects/spring-boot/blob/91da8c56a2bdcb54d0ae9b2e696d45a3d2819f96/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/CollectionBinder.java#L62-L63

This change stops making use of that capability. 

Perhaps more importantly, this may be a breaking change for consumers of the collections as they may no longer be mutable. While we document that the configuration properties classes are not officially public API, we know that people do use them and this change may break their code for reasons that are hard to justify.

quaff:
I see some List are initialized by `List.of()`, should them be wrapped with `new ArrayList<>()`?

wilkinsona:
I'm not sure that it's worth it as the breaking change reasoning above does not apply.

quaff:
> I'm not sure that it's worth it as the breaking change reasoning above does not apply.

I think it's worthy, I created https://github.com/spring-projects/spring-boot/pull/43933.

PR #43920, opened 2025-01-22
Remove unnecessary field initialization in @ConfigurationProperties class

quaff:
It is redundant to assign field to `null` or `false`.


wilkinsona:
Thanks for the PR, but if we're going to make a change like this it should be driven by a team decision to adjust our code style. The adjustment would result in a change in Spring Java Format's Checkstyle rules to add the [explicit initialization check](https://checkstyle.sourceforge.io/checks/coding/explicitinitialization.html). It would then be consistently enforced across the whole codebase.

PR #43922, opened 2025-01-22
Remove unused PrometheusProperties.Pushgateway

quaff:
It's safe to remove it since it's not used anywhere, feel free to close it if I'm wrong.


wilkinsona:
Thanks for raising this @quaff as it's led to me realising that our Promtheus Push Gateway support is in a bit of a mess. I've opened https://github.com/spring-projects/spring-boot/issues/43923 so that we can try to work things out. In the meantime, I don't think we should merge this as it may not be going in the right direction.

PR #43933, opened 2025-01-23
Polish Collection field initialization in @ConfigurationProperties classes

quaff:
1. avoid adding elements to immutable Collection, see https://github.com/spring-projects/spring-boot/blob/91da8c56a2bdcb54d0ae9b2e696d45a3d2819f96/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/CollectionBinder.java#L66-L68
2. use `List.of()` instead of `Arrays.asList()`


bclozel:
This was already declined by a team member. What are you expecting?

quaff:
> This was already declined by a team member. What are you expecting?

The changes by this PR is not same as @wilkinsona declined, please read the description then give your advices.

wilkinsona:
This is the change that was suggested as an alternative on #43918 and that I said was not worth it. Please don’t waste the team’s time by then opening a PR anyway.

PR #43934, opened 2025-01-23
Use isXxx() instead of getXxx() for boolean property

quaff:


philwebb:
Thanks @quaff, I've dropped the `@Deprecated` method in the hope that we can get away with the change since configuration properties aren't considered public API.

PR #43941, opened 2025-01-23
Align return type of VirtualThreadsInfo.getMounted() with the type of the mounted field

izeye:
This PR changes the return type for the `VirtualThreadsInfo.getMounted()` to `int` as it seems to have been `long` accidentally.


wilkinsona:
Well spotted, @izeye. Thank you.

PR #43946, opened 2025-01-23
Throwing specific exception(JSONException) instead of Exception

PiyalAhmed:
Changes made:
- Throwing specific exception(JSONException) instead of Exception
- Added `getDeprecationJsonObject` method for better readability


wilkinsona:
Thanks for the PR but I'm afraid that we prefer not to receive contributions of this nature. The changes are entirely internal and quite subjective and we feel that the time spent thoroughly reviewing and merging them is better spent elsewhere.

PR #43955, opened 2025-01-24
Warn that some Quartz database schema scripts must be modified before use

timotheeandres:
https://github.com/quartznet/quartznet/blob/b9574e915496487e9f031174a61efc2ba572433d/database/tables/tables_sqlServer.sql shows that the SQL Server script requires replacing `enter_db_name_here` manually, so we need to make a copy of the script in our project for it to work.

If this modification is not satisfactory, feel free to edit it; my goal is simply to inform the next person using SQL Server that there is no easy way of using this property.


wilkinsona:
Thanks for bringing this to our attention and for the PR, @timotheeandres. It looks like there's a similar problem with Sybase. I've reworked the warning to take this into account as part of merging this.

Thanks again for the PR and congratulations on making your first contribution to Spring Boot.

PR #43957, opened 2025-01-24
Remove the use of a FactoryBean when auto-configuring VirtualThreadMetrics

nosan:
See gh-43956

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
@@ -16,9 +16,8 @@
 
 package org.springframework.boot.actuate.autoconfigure.metrics;
 
-import java.io.Closeable;
-
 import io.micrometer.core.instrument.MeterRegistry;
+import io.micrometer.core.instrument.binder.MeterBinder;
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/JvmMetricsAutoConfiguration.java
```diff
@@ -53,10 +50,17 @@
...
 
 	private static final String VIRTUAL_THREAD_METRICS_CLASS = "io.micrometer.java21.instrument.binder.jdk.VirtualThreadMetrics";
 
+	private ClassLoader beanClassLoader;
+
+	@Override
+	public void setBeanClassLoader(ClassLoader classLoader) {
+		this.beanClassLoader = classLoader;
+	}
+
```
wilkinsona, line 63: I think we should simplify this further and use the class loader that loaded `JvmMetricsAutoConfiguration`.  Using the bean class loader theoretically allows `JvmMetricsAutoConfiguration` to be loaded by one class loader and `VirtualThreadMetrics` to be loaded by another. I don't think we need to support that arrangement as it wouldn't be supported if this were a straightforward `@Bean` method with `VirtualThreadMetrics` as its return type.

wilkinsona (commented):
Wow! That was quick, @nosan. Thanks very much. I've left one comment for your consideration when you have a minute.

mhalbritter:
Thanks @nosan !

Issue #43964, opened 2025-01-26
Developing Your First Spring Boot Application section asks the user to create the application in the root package


zouchengzhuo:
https://github.com/spring-projects/spring-boot/blob/v3.4.2/spring-boot-project/spring-boot-docs/src/docs/antora/modules/tutorial/pages/first-application/index.adoc

run a simple demo project step by step with the doc below,  a error occur: 

java.lang.ClassNotFoundException:org.springframework.dao.DataAccessException

The #1839 tell me that should not put main method in the default package. maybe we should update the doc?

snicoll:
Thanks for the report. Two things are going one here

1. We were supposed to have a [`com.example` package](https://docs.spring.io/spring-boot/docs/3.2.12/reference/html/getting-started.html#getting-started.first-application.code) but it looks like [it's gone in recent versions](https://docs.spring.io/spring-boot/3.3/tutorial/first-application/index.html#getting-started.first-application.code). Perhaps a regression of the move to Antora?
2. In all cases we instruct the users to create a source file in `src/main/java/MyApplication.java` which is wrong given the package statement.

bclozel:
This rings a bell indeed.

https://github.com/spring-projects/spring-boot/issues/34513

wilkinsona:
For point 1 above, there are a couple of problems:

 - A block-level `chomp_package_replacement` attribute is ignored
 - `include-code` does not transfer the `chomp_package_replacement` attribute to the blocks that it creates.

Both of these will have to be addressed in https://github.com/spring-io/asciidoctor-extensions. I'm working on that now.

wilkinsona:
This is now blocked on:

- https://github.com/spring-io/asciidoctor-extensions/pull/24
- https://github.com/spring-io/asciidoctor-extensions/pull/25

philwebb:
#43982 should have unblocked this issue

wilkinsona:
https://github.com/spring-projects/spring-boot/commit/2fe9de53a0a4e67719ef4f66e26cbe31490d47b2 has taken care of this as it's only the upgrade that was needed.

PR #43965, opened 2025-01-26
Correct the location of MyApplication.java in "Developing Your First Spring Boot Application"

Ru311:
Corrected the file path to create the java project from  src/main/java/MyApplication.java 
to src/main/java/com/example/MyApplication.java


mhalbritter:
Hello,

thanks for the PR. However, for #43964 there's more to be done. Besides changing the path (which this PR does) we also need to include a `package com.example` statement in the example code.

mhalbritter:
Thanks @Ru311 !

PR #43967, opened 2025-01-27
Display package in section Developing Your First Spring Boot Application

ngocnhan-tran1996:
Related https://github.com/spring-projects/spring-boot/issues/43964

`[chomp_package_replacement=com.example]` doesn't seem to work, we can set it on the document and the result:
image


wilkinsona:
Thanks for the PR, @ngocnhan-tran1996, but I don't think that's the right way to fix this one. It's moving the information about packaging chomping too far away from where it applies. I think we need to look at the extension instead and figure out why it's not working as it should.

PR #44013, opened 2025-01-29
Remove shouldNotRegisterStructuredLoggingJsonMembersCustomizerRuntimeHints()

izeye:
It seems to be a duplicate of `shouldNotRegisterStructuredLoggingJsonMembersCustomizerRuntimeHintsWhenCustomizerIsNotSet()` practically.


philwebb:
Thanks @izeye!

nosan:
Not exactly

**shouldNotRegisterStructuredLoggingJsonMembersCustomizerRuntimeHints** checks the following condition 

`return (properties != null) ? AotContribution.get(properties.allCustomizers()) : null` (when the properties variable is null)  

**shouldNotRegisterStructuredLoggingJsonMembersCustomizerRuntimeHintsWhenCustomizerIsNotSet**
checks when the properties variable is not null but customizers are empty.

philwebb:
Whoops, I guess I should revert this then?

nosan:
I think yes :)

izeye:
Sorry, I thought it's a duplicate based on their names.

I created https://github.com/spring-projects/spring-boot/pull/44014 to restore it with rename to reflect its intention.

Issue #44016, opened 2025-01-29
Unable to use native boot with Loki4jAppender


fkjaekel:
I found this [issue](https://github.com/spring-projects/spring-boot/issues/32839), but in my case the loki endpoint has authentication.

Error:
```
ERROR in ch.qos.logback.core.model.processor.ImplicitModelHandler - Could not find an appropriate class for property [auth]
```

Stacktrace:
```
13:11:52.116 [main] INFO org.springframework.boot.devtools.restart.RestartApplicationListener -- Restart disabled due to context in which it is running
13:11:52,146 |-WARN in ch.qos.logback.core.model.processor.AppenderModelHandler - Appender named [LOKI_PRODUCTION] not referenced. Skipping further processing.
Logging system failed to initialize using configuration from 'null'
java.lang.IllegalStateException: Logback configuration error detected: 
ERROR in ch.qos.logback.core.model.processor.ImplicitModelHandler - Could not find an appropriate class for property [auth]
        at org.springframework.boot.logging.logback.LogbackLoggingSystem.reportConfigurationErrorsIfNecessary(LogbackLoggingSystem.java:291)
        at org.springframework.boot.logging.logback.LogbackLoggingSystem.initializeFromAotGeneratedArtifactsIfPossible(LogbackLoggingSystem.java:222)
        at org.springframework.boot.logging.logback.LogbackLoggingSystem.initialize(LogbackLoggingSystem.java:196)
        at org.springframework.boot.context.logging.LoggingApplicationListener.initializeSystem(LoggingApplicationListener.java:332)
        at org.springframework.boot.context.logging.LoggingApplicationListener.initialize(LoggingApplicationListener.java:298)
        at org.springframework.boot.context.logging.LoggingApplicationListener.onApplicationEnvironmentPreparedEvent(LoggingApplicationListener.java:246)
        at org.springframework.boot.context.logging.LoggingApplicationListener.onApplicationEvent(LoggingApplicationListener.java:223)
        at org.springframework.context.event.SimpleApplicationEventMulticaster.doInvokeListener(SimpleApplicationEventMulticaster.java:185)
        at org.springframework.context.event.SimpleApplicationEventMulticaster.invokeListener(SimpleApplicationEventMulticaster.java:178)
        at org.springframework.context.event.SimpleApplicationEventMulticaster.multicastEvent(SimpleApplicationEventMulticaster.java:156)
[...]

PR #44021, opened 2025-01-30
Logback configuration that relies on inner-classes does not work in a native image

nosan:
Before this commit, the generated name for the inner class had the wrong format `<package>.<parent>.<child>` (canonical name). GraalVM expects **$** to separate the parent from the inner class.

This commit updates `SpringBootJoranConfigurator` to generate an appropriate format for a class name. Specifically, an inner class should be separated by a dollar sign, not a dot.

See gh-44016


nosan:
Oh, I apologize for the mistake. The target branch should be 3.3.x. :(

wilkinsona:
No problem at all as you've done more than enough already. I'll take care of it as part of merging the fix.

Issue #43780, opened 2025-01-12
Use consistent exception messages in Assert calls


philwebb:
Currently calls to `Assert` methods use an inconsistent mix of message text. For example, "Thing must not be null", "thing must not be null" or "'thing' must not be null".

I'd like to standardize around the Spring Framework recommendation of using "'thing' must not be null".

PR #44044, opened 2025-02-02
Use consistent exception messages in Assert calls

izeye:
This PR tries to apply https://github.com/spring-projects/spring-boot/issues/43780 to some more places where seem to be applicable.


mhalbritter:
Thanks @izeye !

PR #44071, opened 2025-02-04
Pull up setDisableMBeanRegistry to ConfigurableTomcatWebServerFactory

### LINKED ISSUE #44070: Pull up setDisableMBeanRegistry to ConfigurableTomcatWebServerFactory
The method can be pulled up and we can reduce customizer duplication.

panic08:
Since all inheritors implement `setDisableMBeanRegistry` we should put it in their interface

Fix #44070


panic08:
However, I have a reasonable question for me. You may notice that I removed @since in the above method in javadoc. Should we keep it? Ideally we have just moved it as a new method to the interface, but on the other hand the method was in the inheritors before.

philwebb:
Thanks for the pull-request @panic08. Having looked at this in more detail, we're going to need to do a little more design work and push this back to 4.0.x. The goal behind pulling up the method was to refactor `TomcatReactiveWebServerFactoryCustomizer` and `TomcatServletWebServerFactoryCustomizer`, but that turns out to be quite hard to do without breaking things. Thanks anyway for your efforts.

Issue #44101, opened 2025-02-04
WebServers are not destroyed on application failure in ReactiveWebServerApplicationContext


boolivar:
With #34955 feature (spring-boot >=3.2.0) tomcat is not destroyed on reactive webserver application failure.
Application reports an error on startup but process is not terminated.

Minimal application to reproduce: https://github.com/boolivar/spring-webserver-test (3 dependencies and trivial `TestApplication` class).
Issue is not reproduced on spring-boot v3.1.12 or if servlet web application is configured.

Here is the differences in context behaviors:
[ReactiveWebServerApplicationContext](https://github.com/spring-projects/spring-boot/blob/ddc45ea8ca6b7f4d73f3f2b5bcac66141a83a58f/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/context/ReactiveWebServerApplicationContext.java#L71) - webserver stopped
[ServletWebServerApplicationContext](https://github.com/spring-projects/spring-boot/blob/ddc45ea8ca6b7f4d73f3f2b5bcac66141a83a58f/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/servlet/context/ServletWebServerApplicationContext.java#L152) - webserver stopped and destroyed

wilkinsona:
Closing in favour of #44134.

Issue #34955, opened 2023-04-12
Allow WebServers to be stopped and restarted


tzolov:
On a plain web Spring Boot (3.0.5) application if one do `context.stop()` and then `context.start()` the context on a  I get an error:

```
Exception in thread "main" org.springframework.context.ApplicationContextException: Failed to start bean 'webServerStartStop'
        at org.springframework.context.support.DefaultLifecycleProcessor.doStart(DefaultLifecycleProcessor.java:181)
        at org.springframework.context.support.DefaultLifecycleProcessor$LifecycleGroup.start(DefaultLifecycleProcessor.java:356)
        at java.base/java.lang.Iterable.forEach(Iterable.java:75)
        at org.springframework.context.support.DefaultLifecycleProcessor.startBeans(DefaultLifecycleProcessor.java:155)
        at org.springframework.context.support.DefaultLifecycleProcessor.start(DefaultLifecycleProcessor.java:103)
        at org.springframework.context.support.AbstractApplicationContext.start(AbstractApplicationContext.java:1418)
        at com.example.crac.springwebcrac.SpringWebCracApplication.main(SpringWebCracApplication.java:35)
Caused by: java.lang.IllegalStateException: The host does not contain a Context
        at org.springframework.boot.web.embedded.tomcat.TomcatWebServer.findContext(TomcatWebServer.java:153)
        at org.springframework.boot.web.embedded.tomcat.TomcatWebServer.start(TomcatWebServer.java:232)
        at org.springframework.boot.web.servlet.context.WebServerStartStopLifecycle.start(WebServerStartStopLifecycle.java:44)
        at org.springframework.context.support.DefaultLifecycleProcessor.doStart(DefaultLifecycleProcessor.java:178)
```

To reproduce create a plain web boot application `SpringWebCracApplication` like this:
```java
	public static void main(String[] args) throws InterruptedException {
		ConfigurableApplicationContext context = SpringApplication.run(SpringWebCracApplication.class, args);
		context.stop();
		Thread.sleep(5000);
		context.start();
	}
```
It looks like the `tomcat.destroy()` , called on `context.stop()`, destroys the `Tomcat.Context` and later is not re-created on start.

Note that If we replace the `Tomcat` with a `Jetty` the application works fine and recovers after stop/start.

jhoeller:
[...]

PR #44134, opened 2025-02-05
WebServer is not destroyed when ReactiveWebServerApplicationContext refresh fails

nosan:
Prior to this commit, if `ReactiveWebServerApplicationContext` failed to refresh, only `WebServer.stop()` was called.

This commit additionally invokes `WebServer.destroy()`, aligning the behaviour with `ServletWebServerApplicationContext` when a refresh failure occurs.

gh-44101

I am also thinking about these changes as well:

https://github.com/spring-projects/spring-boot/compare/3.3.x...nosan:spring-boot:44100-1

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/context/ReactiveWebServerApplicationContext.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/context/ReactiveWebServerApplicationContext.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2022 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -66,9 +66,10 @@ public class ReactiveWebServerApplicationContext extends GenericReactiveWebAppli
 			super.refresh();
 		}
 		catch (RuntimeException ex) {
-			WebServerManager serverManager = this.serverManager;
-			if (serverManager != null) {
-				serverManager.getWebServer().stop();
+			WebServer webServer = getWebServer();
+			if (webServer != null) {
+				webServer.stop();
+				webServer.destroy();
 			}
 			throw ex;
 		}
@@ -147,6 +148,10 @@ public class ReactiveWebServerApplicationContext extends GenericReactiveWebAppli
 			AvailabilityChangeEvent.publish(this, ReadinessState.REFUSING_TRAFFIC);
 		}
 		super.doClose();
+		WebServer webServer = getWebServer();
+		if (webServer != null) {
+			webServer.destroy();
+		}
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/context/ReactiveWebServerApplicationContext.java
```diff
@@ -66,9 +66,15 @@ public final void refresh() throws BeansException, IllegalStateException {
...
 		}
 		catch (RuntimeException ex) {
-			WebServerManager serverManager = this.serverManager;
-			if (serverManager != null) {
-				serverManager.getWebServer().stop();
+			try {
+				WebServer webServer = getWebServer();
+				if (webServer != null) {
+					webServer.stop();
+					webServer.destroy();
```
boolivar, line 73: Doesn't it make sense to destroy webServer even if `stop()` throws an exception?

_Probably out of the scope for this fix, but it looks like Juergen was thinking about calling `destroy()` method in context of a bean lifecycle: https://github.com/spring-projects/spring-boot/issues/34955#issuecomment-1525344836_

Thank you for the quick action in regard #44101! 🙏

nosan: I don’t think so. If `stop()` throws an exception, it likely means that some resources, such as files, networks, etc. are still in use, which could cause `destroy()` to fail as well.

nosan:
An alternative approach is https://github.com/spring-projects/spring-boot/compare/3.3.x...nosan:spring-boot:44101-alternative, which always tries to stop and destroy `WebServer` in case of a refresh failure or during `ApplicationContext.close()`.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/context/ReactiveWebServerApplicationContext.java
```diff
@@ -66,9 +66,15 @@ public final void refresh() throws BeansException, IllegalStateException {
...
+				}
+			}
+			catch (Exception shutdownEx) {
+				ex.addSuppressed(shutdownEx);
```
wilkinsona, line 77: Let's consider this separately please. If we do it, it should be applied on the servlet side as well.

wilkinsona (changes requested):

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/reactive/context/ReactiveWebServerApplicationContext.java
nosan: Sure, I've reverted these changes.

wilkinsona: We think that adding the stop or destroy failure as a suppressed exception is a good idea. Would you like to open a PR for it or would you prefer that we take care of it?

nosan: [#44310](https://github.com/spring-projects/spring-boot/pull/44310)

wilkinsona: Thanks!

PR #44172, opened 2025-02-08
Mappings Endpoint support web-servlet router funtion

Puppy4C:
1. Mappings Endpoint support web-servlet router funtion (Previously, only web-reactive was supported).
2. Enrich the mapping info of router funtion (Refer to `ToStringVisitor`).


wilkinsona:
Thanks for the PR, @Puppy4C, but it looks like it contains two changes that should be considered separately.

Can you please update this proposal so that it only adds support for router functions on the servlet stack? The output that's produced should align as closely as possible with the current output for router functions on the reactive stack.

Once that change has been reviewed and hopefully merged we can then consider improvements to the mapping information in a separate PR.

Puppy4C:
Hi @wilkinsona , I have already removed another change.

wilkinsona:
Closed by https://github.com/spring-projects/spring-boot/commit/b2ba2a53fe43f9204aa8d50ae1a5e327b1c5ba61. Thanks very much @Puppy4C.

Issue #42330, opened 2024-09-16
Consider showing more information in StartupInfoLogger


philwebb:
See #42324 for background.

Since `StartupInfoLogger` was first introduced we've added more features that could be useful to log. The `application.name` and `application.version` could be useful as could the git SHA if it's available.

We need to be a little careful not to add too much information and not to change the defaults in a way that folks won't like.

philwebb:
We like the idea of at least logging the application name, if we can do it without adding too much noise.

Dunemaster:
My two cents:
 I would vote for logging application version as well. It is very useful when receiving logs from client om-premise installations. 
Version extraction can be implemented in several ways (maybe with a configurable provider). We implemented extracting version from the JPMS module version, this seems to be a reasonable default.

One of the stackoverflow answers show how to add a module version using maven https://stackoverflow.com/questions/48151160/does-jpms-support-module-version

The official docs:
https://docs.oracle.com/en/java/javase/17/docs/specs/man/jar.html (see --module-version)

And in app code the version can be retrieved like this

```
System.out.println("Hello, modular World!");
Module module = HelloModularWorld.class.getModule();
ModuleDescriptor descriptor = module.getDescriptor();
System.out.println(descriptor.rawVersion());
```

sonu4578:
@philwebb Hi Philip, I can create a PR for logging the application name and version. Could you please assign this issue to me?

wilkinsona:
Thanks for the offer, @sonu4578, but as indicated by the issue being labelled as "pending design work", we're not ready to accept contributions for this issue.

PR #44203, opened 2025-02-11
Add application name support to StartupInfoLogger

sonjih00n:
`spring.application.name` does not work with StartupInfoLogger.
Only mainApplicationClass name is logged. below is an example.

```yml
spring:
  application:
    name: HelloApp
```
```java
@SpringBootApplication
public class MyApplication {
   // something
  }
}
```

### Expected
```log
[INFO ][](StartupInfoLogger.java:53) Starting HelloApp using Java 17.0.6 with PID ...
```

### Actual
```log
[INFO ][](StartupInfoLogger.java:53) Starting MyApplication using Java 17.0.6 with PID ...
```


philwebb:
Thanks, this is open in #42330 however the issue is pending design work so we aren't able to accept contributions at the moment. The suggested fix in this PR could also break existing applications that rely on the application name from the `MANFIEST.MF`.

Thanks anyway for your efforts.

Issue #43864, opened 2025-01-16
Allow stacktrace processing when using structured logging


michaldo:
Structured logging significantly simplifies production setup. Unfortunately, missing equivalent of `ShortenedThrowableConverter` is disqualifying.

It is not just readability issue. Uncontrolled stacktraces can exceed log buffer, break parsing JSON and finally "hide" from search engine
Ref: https://stackoverflow.com/questions/79294615/shortenedthrowableconverter-in-spring-boot-3-4-structured-logging

mhalbritter:
For reference, `ShortenedThrowableConverter` is from `logstash-logback-encoder` and can be found [here](https://github.com/logfellow/logstash-logback-encoder/blob/main/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java).

philwebb:
Stacktrace processing was also discussed in #42486

philwebb:
Collating references for ideas:
 
* https://github.com/mp911de/logstash-gelf/blob/main/src/main/java/biz/paluch/logging/StackTraceFilter.java
* https://github.com/logfellow/logstash-logback-encoder/blob/main/src/main/java/net/logstash/logback/stacktrace/ShortenedThrowableConverter.java

michaldo:
Hello, I see that StandardStackTracePrinter is prepared to get me what I want

If I configure `logging.structured.json.stacktrace.printer: org.michaldo.MyPrinter`
and I implement

```
public class MyPrinter implements StackTracePrinter  {

    private final StandardStackTracePrinter standard;

    public MyPrinter(StandardStackTracePrinter standard) {
        List<Pattern> exclusionPatterns = Stream.of(
                "java.lang.reflect.Method",
                "org.apache.catalina",
                "org.springframework.aop",
                "org.springframework.security",
                "org.springframework.transaction",
                "org.springframework.web",
                "jdk.internal.reflect.DirectMethodHandleAccessor")
                .map(Pattern::compile)
                .toList();
        this.standard = standard.withFrameFilter((any, element) -> {
                    String classNameAndMethod = element.getClassName() + "." + element.getMethodName();
                    return exclusionPatterns
                            .stream()
                            .noneMatch(pattern -> pattern.matcher(classNameAndMethod).find());
        });
    }

    @Override
    public void printStackTrace(Throwable throwable, Appendable out) throws IOException {
        standard.printStackTrace(throwable, out);
    }
```
[...]

PR #44242, opened 2025-02-12
RuntimeHints are missing for StackTracePrinter

nosan:
See gh-43864


snicoll:
The feature hasn't been released yet so I've moved this to a task.

Issue #43481, opened 2024-12-11
Metrics and health do not include non-default candidate beans


mdaepp:
The documentation for Spring Boot 3.4 now recommends the `@Qualifier("xyz")` and `@Bean(defaultCandidate =  false)` approach to define multiple `DataSources`: https://docs.spring.io/spring-boot/3.4/how-to/data-access.html#howto.data-access.configure-two-datasources

At first glance, this approach is neat because the first `DataSource` is set up by auto-configuration as if there were no second `DataSource`. However, the new approach has some probably unwanted side-effects that users should be made aware of.

With the previous approach (see https://docs.spring.io/spring-boot/3.3/how-to/data-access.html#howto.data-access.configure-two-datasources), the additional `DataSources` would still participate in other auto-configurations such as `DataSourcePoolMetricsAutoConfiguration` and `DataSourceHealthContributorAutoConfiguration`. That is no longer the case with the new approach, because an additional `DataSource` with `defaultCandidate =  false` is neither present in `Map<String, DataSource>` nor is it available via `ObjectProvider<DataSource>`.

I leave it to the subject matter experts to decide what should be done:

1. Update documentation to mention the limitations of the new approach
2. Revert documentation to previous approach
3. Fix other auto-configurations that apply to a "list" of DataSources of some sort (such as `ObjectProvider<DataSource>`)—not sure if that is even possible or a good idea for that matter

quaff:
> Fix other auto-configurations that apply to a "list" of DataSources of some sort (such as `ObjectProvider<DataSource>`)

`ObjectProvider<DataSource>` filter out non-defaultCandidate beans, we could use `beanFactory.getBeanNamesForType(DataSource.class)` instead of injection of `Map<String, DataSource>`.

quaff:
> > Fix other auto-configurations that apply to a "list" of DataSources of some sort (such as `ObjectProvider<DataSource>`)
> 
> `ObjectProvider<DataSource>` filter out non-defaultCandidate beans, we could use `beanFactory.getBeansOfType(DataSource.class)` instead of injection of `Map<String, DataSource>`.

I'd like to prepare an PR if the team accept this solution.

philwebb:
I'm not a fan of using the `BeanFactory` directly, but I can't immediately think of a better option. @jhoeller Ant thoughts on the best way to inject something that will include non-defaultCandidate beans? Perhaps we need some new methods on `ObjectProvider` or a new annotation to signal intent?

[...]

PR #44253, opened 2025-02-13
Fix non-default HikariDataSource candidates are not included in metrics

quaff:
See https://github.com/spring-projects/spring-boot/issues/43481#issuecomment-2655527313


spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/jdbc/DataSourcePoolMetricsAutoConfigurationTests.java
```diff
@@ -82,7 +83,7 @@ void dataSourceInstrumentationCanBeDisabled() {
 	}
 
 	@Test
-	void allDataSourcesCanBeInstrumented() {
+	void allDataSourceCandidatesCanBeInstrumented() {
```
quaff, line 86: I renamed test names for accuracy, I'll revert this part if you object. @wilkinsona

wilkinsona: I'd prefer it without `Candidates`, for consistency with all the other tests for similar scenarios if nothing else.

wilkinsona (approved):

Issue #44250, opened 2025-02-13
Correction in Spring Security documentation: Missing conditional "if" in UserDetailsService auto-configuration explanation


Saaiiikrishna:
#### **Description of the Issue**  
In the Spring Security documentation, the following sentence appears incorrect due to a missing conditional "if":  

> "The auto-configuration of a UserDetailsService will also back off any of the following Spring Security modules is on the classpath:"  

This sentence is misleading as it suggests that the `UserDetailsService` auto-configuration will actively "back off" those modules, which is not the intended behavior.  

#### **Suggested Correction**  
The corrected sentence should be:  

> "The auto-configuration of a UserDetailsService will also back off if any of the following Spring Security modules is on the classpath:"  

#### **Why is this Correction Necessary?**  
- **Grammar & Clarity:** The missing "if" makes the sentence grammatically incorrect and harder to understand.  
- **Technical Accuracy:** Without "if," it implies that `UserDetailsService` actively disables other modules, while the correct behavior is that `UserDetailsService` auto-configuration is **skipped** when those modules are present.  

#### **Steps to Reproduce:**  
1. Visit the [[Spring Security documentation](https://docs.spring.io/spring-security/reference/index.html)](https://docs.spring.io/spring-security/reference/index.html) (or the specific file in the GitHub repository).  
2. Locate the section discussing `UserDetailsService` auto-configuration.  
3. Observe the incorrect sentence structure.  

#### **Proposed Fix:**  
Update the documentation to include the missing **"if"** for better clarity and correctness.  

Would appreciate your confirmation and update on this!

wilkinsona:
Thanks for spotting this. I assume you're referring to this phrase in [this section](https://docs.spring.io/spring-boot/reference/web/spring-security.html#web.security.spring-mvc) of Spring Boot's documentation:

> The auto-configuration of a [UserDetailsService](https://docs.spring.io/spring-security/site/docs/6.4.x/api/org/springframework/security/core/userdetails/UserDetailsService.html) will also back off any of the following Spring Security modules is on the classpath

I think we should add "when". This will align with the similar phrase in the [section on WebFlux security](https://docs.spring.io/spring-boot/reference/web/spring-security.html#web.security.spring-webflux) immediately after.

Saaiiikrishna:
Thank you for reviewing my issue and for pointing out the consistency with the WebFlux security section. Yes, I was referring to that phrase in the Spring Boot documentation.  

[...]

PR #44267, opened 2025-02-13
Correct typo in MVC security when explaining when UserDetailsService auto-configuration will back off

ngocnhan-tran1996:
Closes gh-44250


Saaiiikrishna (commented):
I think this is would fix the issue.

PR #44309, opened 2025-02-17
Handle absolute file path on Windows

nosan:
Fixes build on Windows (relates to https://github.com/spring-projects/spring-boot/commit/528b7e9ad919571bc04f3b4b2d0c557cf8ef00d6)


nosan:
I am not sure this is a bug, it should be a task, these changes have not been released yet as far as I know.

mhalbritter:
I was conflicted, and you're right :)

mhalbritter:
Thanks @nosan !

PR #44310, opened 2025-02-17
When web server application context refresh fails, the original failure is lost if stopping or destroying the web server throws an exception

nosan:


snicoll:
Thanks again, @nosan.

Issue #39600, opened 2024-02-16
Rename "micrometer.observations.*" configuration properties to "management.observations.*"


bclozel:
Introducing the `micrometer.observations.annotations.enabled` configuration property in that namespace was a mistake introduced in #39128 and we should move it to `management.observations.annotations.enabled`.

As part of this issue, we should try to support both and deprecate `micrometer.observations.annotations.enabled` in our metadata, for removal in a future Spring Boot version.

PR #44320, opened 2025-02-18
Remove deprecated micrometer.observations.annotations.enabled

izeye:
This PR removes the `micrometer.observations.annotations.enabled` property that has been deprecated since Spring Boot 3.2.3 in #39600.


snicoll:
Thanks Johnny, the condition could be further simplified. I've done that in [cacb3d4](https://github.com/spring-projects/spring-boot/commit/cacb3d4820ce1e3987d03fbb368926a18199db1e)

PR #44340, opened 2025-02-18
Don't use ArgFile when using JavaCompiler

nosan:
Fixes build on Windows

see https://github.com/spring-projects/spring-boot/pull/44305

ArgFile should not be used when compiling sources via `JavaCompiler`.  

>[CreateProcess error=206, The filename or extension is too long]

This happens only when running a process via `ProcessBuilder` but not `JavaCompiler`.


philwebb:
Thanks @nosan, I was trying to look at this but it's taken me all day to get a Windows box installed!

nosan:
I've installed VMware Fusion. Works perfectly :)

snicoll:
There's no need for an apology @nosan. That's part of the review process and it's always tricky with limited access to a windows box.

Issue #43118, opened 2024-11-12
Console logging incorrectly uses Charset.defaultCharset()


crmky:
[JEP 400: UTF-8 by Default](https://openjdk.org/jeps/400) was included in Java 18 and later version. Basically it changes the default `file.encoding` to `UTF-8` by default and introduced a few additional properties to identify system native encoding. However the console output charset is different:

> Those APIs include the ones listed above, **but not System.out and System.err, whose charset will be as specified by [Console.charset()](https://bugs.openjdk.java.net/browse/JDK-8264208)**.

However Spring Boot doesn't respect `Console.charset()`. It always use `org.springframework.boot.logging.LoggingSystemProperties.getDefaultCharset()` to set the default charset for console and file encoding. Depends on the implementation, it uses either hard-coded `UTF-8` or `Charset.defaultCharset()`.

Below is a simple Spring Boot application to reproduce the issue (The system locale needs to be different than English):

```java
    public static void main(String[] args) {
        SpringApplication.run(Application.class, args);
        System.out.println("你好");
        log.info("你好");
    }
```

Run it with Java 17:

```
你好
2024-11-12T10:16:10.992+08:00  INFO 19176 --- [test] [           main] test.Application           : 你好
```

Run it with Java 21 (my system locale is GBK):

```
你好
2024-11-12T10:16:28.478+08:00  INFO 524 --- [test] [           main] test.Application           : 浣犲ソ
```

The workaround is to add properties like below. But I think a better value should be provided by default.
```properties
logging.charset.console=${stdout.encoding}
```

mhalbritter:
The charset has been set to `Charset.defaultCharset()` in https://github.com/spring-projects/spring-boot/issues/27230.

Having read the JEP, I think we should try to use `System.console().charset()` for the console loggers.

One thing to note is that `System.console()` can return `null`, if the process doesn't have a console (or if it's running with the IntelliJ IDEA console?!)

Flagging this for team attention.

mhalbritter:
Superseded by https://github.com/spring-projects/spring-boot/pull/44353.

PR #44353, opened 2025-02-19
Use Console charset for console logging when available

nosan:
Prior to this commit, `LogbackLoggingSystemProperties` doesn't respect `Console.charset()`. It used `Charset.getDefaultCharset()` for logback and  `UTF-8` for log42j as defaults.

This commit changes the behaviour of `LogbackLoggingSystemProperties` to use
 `Console.charset()` when available. If no console is present, the default
 charset is used instead. These changes bring consistency across
 logging implementations.

See gh-43118

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggingSystemProperties.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggingSystemProperties.java
@@ -16,8 +16,8 @@
 
 package org.springframework.boot.logging;
 
+import java.io.Console;
 import java.nio.charset.Charset;
-import java.nio.charset.StandardCharsets;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
@@ -91,8 +91,12 @@ public class LoggingSystemProperties {
 		this.setter = (setter != null) ? setter : systemPropertySetter;
 	}
 
+	protected Console getConsole() {
+		return System.console();
+	}
+
 	protected Charset getDefaultCharset() {
-		return StandardCharsets.UTF_8;
+		return Charset.defaultCharset();
 	}
 
 	public final void apply() {
@@ -116,12 +120,11 @@ public class LoggingSystemProperties {
 	}
 
 	protected void apply(LogFile logFile, PropertyResolver resolver) {
-		String defaultCharsetName = getDefaultCharset().name();
 		setSystemProperty(LoggingSystemProperty.APPLICATION_NAME, resolver);
 		setSystemProperty(LoggingSystemProperty.APPLICATION_GROUP, resolver);
 		setSystemProperty(LoggingSystemProperty.PID, new ApplicationPid().toString());
-		setSystemProperty(LoggingSystemProperty.CONSOLE_CHARSET, resolver, defaultCharsetName);
-		setSystemProperty(LoggingSystemProperty.FILE_CHARSET, resolver, defaultCharsetName);
+		setSystemProperty(LoggingSystemProperty.CONSOLE_CHARSET, resolver, getConsoleCharset().name());
+		setSystemProperty(LoggingSystemProperty.FILE_CHARSET, resolver, getDefaultCharset().name());
 		setSystemProperty(LoggingSystemProperty.CONSOLE_THRESHOLD, resolver, this::thresholdMapper);
 		setSystemProperty(LoggingSystemProperty.FILE_THRESHOLD, resolver, this::thresholdMapper);
 		setSystemProperty(LoggingSystemProperty.EXCEPTION_CONVERSION_WORD, resolver);
@@ -135,6 +138,11 @@ public class LoggingSystemProperties {
 		if (logFile != null) {
 			logFile.applyToSystemProperties();
 		}
+	}
+
+	private Charset getConsoleCharset() {
+		Console console = getConsole();
+		return (console != null) ? console.charset() : getDefaultCharset();
 	}
 
 	private void setSystemProperty(LoggingSystemProperty property, PropertyResolver resolver) {
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystemProperties.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/logback/LogbackLoggingSystemProperties.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2024 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,7 @@
 
 package org.springframework.boot.logging.logback;
 
-import java.nio.charset.Charset;
+import java.io.Console;
 import java.util.function.BiConsumer;
 import java.util.function.Function;
 
@@ -71,8 +71,8 @@ public class LogbackLoggingSystemProperties extends LoggingSystemProperties {
 	}
 
 	@Override
-	protected Charset getDefaultCharset() {
-		return Charset.defaultCharset();
+	protected Console getConsole() {
+		return super.getConsole();
 	}
 
 	@Override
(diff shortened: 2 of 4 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/LoggingSystemProperties.java
```diff
@@ -91,8 +92,12 @@ public LoggingSystemProperties(Environment environment, Function<String, String>
 		this.setter = (setter != null) ? setter : systemPropertySetter;
 	}
 
+	protected Console getConsole() {
+		return System.console();
+	}
+
 	protected Charset getDefaultCharset() {
-		return StandardCharsets.UTF_8;
+		return Charset.defaultCharset();
```
nosan, line 100: This change will impact **log42j** `CONSOLE_CHARSET` and `FILE_CHARSET`.

mhalbritter: I think it's fine that the log4j console output is changed, but I'm not sure about the file encoding. I'd prefer if the file stays utf-8. #43118 talks about console logging only.

nosan: The current situation

|             | Log4j2 | Logback           |
|-------------|--------|-------------------|
| **Console** | UTF-8  | Charset.default() |
| **File**    | UTF-8  | Charset.default() |


The proposed PR aims to align the situation as follows:



|             | Log4j2                                 | Logback                                |
|-------------|----------------------------------------|----------------------------------------|
| **Console** | Console.charset() or Charset.default() | Console.charset() or Charset.default() |
| **File**    | Charset.default()                      | Charset.default()                      |


Alternatively, we could leave `protected Charset getDefaultCharset()` untouched, and this would result in


|             | Log4j2                     | Logback                                |
|-------------|----------------------------|----------------------------------------|
| **Console** | Console.charset() or UTF-8 | Console.charset() or Charset.default() |
| **File**    | UTF-8                      | Charset.default()                      |


What do you think?

mhalbritter: The last table. I have some changes here, WDYT? https://github.com/mhalbritter/spring-boot/tree/pr/44353

mhalbritter:
I'm actually surprised that the logback file is not written in UTF-8. I'll talk to the team about that.

mhalbritter:
I investigated that charset stuff a bit more, and I'll try to summarize my findings:

On Linux, with the `LANG` variable, you can control the value of `System.console().charset()`, and with `-Dfile.encoding` you can change `Charset.defaultCharset()` (however, this is undefined behavior on Java 18+). If you don't specify `-Dfile.encoding`, Java < 18 uses some environment dependent logic to find the default locale. This is UTF-8 on Linux and MacOS most of the time, and on Windows it's not. On Java 18+ this returns UTF-8 unless backward compatibility mode is enabled.

mhalbritter:
[Archive.zip](https://github.com/user-attachments/files/19010033/Archive.zip)

This small program prints some environment variables and uses Logback and Log4j without Spring Boot to print some UTF-8 chars.

That's the output:

# Java 17

```
LANG=en_US.utf8 app/build/distributions/app/bin/app     
Java version: 17.0.14
stdout.encoding: null
file.encoding: UTF-8
native.encoding: UTF-8
sun.stdout.encoding: UTF-8
sun.stderr.encoding: UTF-8
sun.jnu.encoding: UTF-8
Console charset: UTF-8
Default charset: UTF-8
System.out.println: 你好
UTF-8: 你好
Console charset: 你好
Default charset: 你好
11:13:29.981 [main] INFO  org.example.App -- Logback: 你好
Log4j: 你好
```

```
LANG=en_US.iso885915 app/build/distributions/app/bin/app
Java version: 17.0.14
stdout.encoding: null
file.encoding: ISO-8859-15
native.encoding: ISO-8859-15
sun.stdout.encoding: ISO-8859-15
sun.stderr.encoding: ISO-8859-15
sun.jnu.encoding: ISO-8859-15
Console charset: ISO-8859-15
Default charset: ISO-8859-15
System.out.println: ??
UTF-8: 你好
Console charset: ??
Default charset: ??
11:13:45.118 [main] INFO  org.example.App -- Logback: ??
Log4j: ??
```

# Java 23

```
LANG=en_US.utf8 app/build/distributions/app/bin/app     
Java version: 23.0.2
stdout.encoding: UTF-8
file.encoding: UTF-8
native.encoding: UTF-8
sun.stdout.encoding: null
sun.stderr.encoding: null
sun.jnu.encoding: UTF-8
Console charset: UTF-8
Default charset: UTF-8
System.out.println: 你好
UTF-8: 你好
Console charset: 你好
Default charset: 你好
11:13:13.565 [main] INFO  org.example.App -- Logback: 你好
Log4j: 你好
```

```
LANG=en_US.iso885915 app/build/distributions/app/bin/app
Java version: 23.0.2
stdout.encoding: ISO-8859-15
file.encoding: UTF-8
native.encoding: ISO-8859-15
sun.stdout.encoding: null
sun.stderr.encoding: null
sun.jnu.encoding: ISO-8859-15
Console charset: ISO-8859-15
Default charset: UTF-8
System.out.println: ??
UTF-8: 你好
Console charset: ??
Default charset: 你好
11:12:43.208 [main] INFO  org.example.App -- Logback: 你好
Log4j: 你好
```

# Logback

If `charset` is not set, `ch.qos.logback.core.encoder.LayoutWrappingEncoder.convertToBytes` uses `String.getBytes()` - which uses the default charset (Java < 18: depends on OS, Java >= 18: UTF-8)

# Log4j

If the `ConsoleAppender` doesn't have an explicit layout, it creates a `PatternLayout` using the default target charset (see logic in `org.apache.logging.log4j.core.appender.AbstractAppender.Builder.getOrCreateLayout(java.nio.charset.Charset)`).
The default target charset is found by reading the system properties `sun.stdout.encoding` and `sun.stderr.encoding`. If `null`, `Charset.defaultCharset()` is used.
When explicitly creating a `PatternLayout` and charset is not set, `UTF-8` is used (see constructor of `AbstractStringLayout`)

mhalbritter:
In the above output, every time `System.out.println:` differs from `Logback:` or from `Log4j:`, there's a bug where the output doesn't respect the users locale. That's the case for Java 23, because there `Charset.defaultCharset()` returns UTF-8 (isn't affected by the `LANG` variable) and both Logback and Log4j use it, but `System.out.println()` uses the `Console.charset()` charset, which is affected by the `LANG` variable.

mhalbritter:
Java 23 and Logback is broken, because `System.out.println()` uses `Console.charset()` and Logback is configured to use `Charset.defaultCharset()` (which returns UTF-8).

Java 17 and Log4j is broken, because `System.out.println` uses the value from `LANG` but Log4j is hardcoded to use UTF-8.

Java 23 and Log4j is broken, because `System.out.println()` uses `Console.charset()` but Log4j is hardcoded to use UTF-8.

mhalbritter:
So, to clean this up, I **think** (my brain is a bit exhausted from all the charsets) if we implement it like this it works on all systems:

|             | Log4j2                                 | Logback                                |
|-------------|----------------------------------------|----------------------------------------|
| **Console** | Console.charset() or Charset.default() | Console.charset() or Charset.default() |

nosan:
Thank you very much, @mhalbritter for such a detailed analysis!

> Which is what this PR does. But it also touches the default file encoding, which I
> a) either leave as it is or
> b) change to UTF-8

I think both options are good as they bring consistency between Log4j2 and Logback. I personally prefer to have:

|             | Log4j2                                 | Logback                                |
|-------------|----------------------------------------|----------------------------------------|
| **Console** | Console.charset() or Charset.default() | Console.charset() or Charset.default() |
| **File**    | UTF-8                                  | UTF-8                                  |

mhalbritter:
Thanks @nosan! I've decided to not touch the file encoding in this PR. I'm going to open a new issue for that.

nosan:
Thanks @mhalbritter

I believe this is a good compromise. On one hand, the bug is fixed; on the other, the encoding can be revisited in a major version. 
I really like it 👍

PR #44354, opened 2025-02-19
Introduce TestSliceTestContextBootstrapper for test slices

quaff:


spring-boot-project/spring-boot-test/src/main/java/org/springframework/boot/test/context/TestSliceTestContextBootstrapper.java
```diff
@@ -0,0 +1,72 @@
...
+			.resolve();
+		Assert.notNull(this.annotationType, "'%s' doesn't contain type parameter of '%s'"
+			.formatted(getClass().getName(), TestSliceTestContextBootstrapper.class.getName()));
+		try {
+			this.propertiesMethod = this.annotationType.getDeclaredMethod("properties");
+			Assert.state(this.propertiesMethod.getReturnType() == String[].class,
+					"Method '%s.properties()' should return String[]".formatted(this.annotationType.getName()));
+		}
+		catch (NoSuchMethodException ex) {
+			throw new RuntimeException(
```
quaff, line 54: I need advice for exception type and message.

nosan: You can simplify things a little bit

```java
public class TestSliceTestContextBootstrapper<A extends Annotation> extends SpringBootTestContextBootstrapper {

	private final Class<A> annotationType;


	@Override
	protected String[] getProperties(Class<?> testClass) {
		MergedAnnotation<A> annotation = MergedAnnotations.search(SearchStrategy.TYPE_HIERARCHY)
			.withEnclosingClasses(TestContextAnnotationUtils::searchEnclosingClass)
			.from(testClass)
			.get(this.annotationType);
		return annotation.getStringArray("properties");
	}

}
```

quaff: Yes, I have considered this way, but I think it's better to keep the logic in `TestContextAnnotationUtils`, WDYT?

nosan: > but I think it's better to keep the logic in TestContextAnnotationUtils, WDYT?

Honestly, I don't see any issues with using `MergedAnnotations.search(...)` instead of
`TestContextAnnotationUtils`. On one hand, it would be nice if the Spring Framework
could provide this functionality, but it does not make the situation significantly better.
Considering that it is used only once in a single place, IMO, it is too much to ask to add it to the Spring Framework.

Also, this PR is still in a _draft_ state, and there is no feedback from the Spring Boot team regarding
this enhancement, so maybe it makes sense to wait for feedback first and then ask the Spring Framework team to
add something.

quaff: Updated as you suggested, thank you @nosan

wilkinsona:
Thanks very much, @quaff.

PR #44355, opened 2025-02-19
Polish ClassPath

nosan:
Introduced `toString` to fix the following:

```java


	private void addClasspath(List<String> args) throws MojoExecutionException {
		try {
			ClassPath classpath = ClassPath.of(getClassPathUrls());
			if (getLog().isDebugEnabled()) {
				getLog().debug("Classpath for forked process: " + classpath);
			}
			args.addAll(classpath.args(true));
		}
		catch (Exception ex) {
			throw new MojoExecutionException("Could not build classpath", ex);
		}
	}
```


philwebb:
Gahhhh! I knew I was going to forget to add a `toString()` when I was refactoring! Thanks for catching that @nosan!

Issue #43851, opened 2025-01-16
Introduce spring.data.mongodb.protocol to allow the MongoDB protocol to be customized when not using spring.data.mongodb.uri


storyzero:
https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/PropertiesMongoConnectionDetails.java#L51

I can't generate a connection string with mongodb+srv:// because PropertiesMongoConnectionDetails is hardcoded to use mongodb://
MongoProperties should be able to specify the protocol, and PropertiesMongoConnectionDetails should generate the connection string using the specified protocol.

wilkinsona:
This is similar to https://github.com/spring-projects/spring-boot/issues/17215, but the proposal there was to inject the separate properties into the URI. This was rejected as it would be a breaking change that would prevent the URI being used to override any other settings. We can consider introducing a `protocol` property for those who prefer to use the separate properties. It would be ignored when `spring.data.mongodb.uri` is specified.

In the meantime, please note that you can specify `mongodb+srv` as the protocol but, to do so, you must use `spring.data.mongodb.uri` to provide all of the connection information. When configuring the URI you can reference other properties if you wish:

```
example.mongodb.username=alice
example.mongodb.password=secret
spring.data.mongodb.uri=mongodb+srv://${example.mongodb.username}:${example.mongodb.password}@mongo.example/testdb
```

storyzero:
@wilkinsona 
Thank you for your response. 
To elaborate more on my situation, I want to use spring-cloud-vault-config-databases to auto-configure spring.data.mongodb.username and password. For security reasons, it is mandatory. Similarly, using mongodb+srv is also mandatory. 

The example you provided seems worth trying, but the approach I was considering until just now was to manually configure everything, from the connection string to MongoTemplate, without any auto-configuration, using Vault.

However, it's a bit unfortunate that mongodb+srv can't be used unless it's through the URI.

mhalbritter:
Superseded by https://github.com/spring-projects/spring-boot/pull/44366.

PR #44366, opened 2025-02-19
Add spring.data.mongodb.protocol property

nosan:
See gh-43851


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoProperties.java
```diff
@@ -51,6 +51,11 @@ public class MongoProperties {
 	 */
 	public static final String DEFAULT_URI = "mongodb://localhost/test";
 
+	/**
+	 * Protocol to be used for the MongoDB connection. Ignored if 'uri' is set.
```
snicoll, line 55: I was just looking at this and laughed that it matches the polish I've made in 44384. What a coincidence! Nope, you forced push to adapt... You really didn't have to do it, thanks!

Issue #43963, opened 2025-01-25
Log4j2 not working with Spring Boot 3.4.2 together with BlockHound


micopiira:
When upgrading to Spring Boot 3.4.2 our logging stopped completely working. 

After narrowing down I figured the problem happens with Spring Boot 3.4.2 when using Log4j2 together with BlockHound.

Here is a minimum reproducer repo: https://github.com/micopiira/spring-boot-log4j2-issue

Spring Boot 3.4.1 with exact same Log4j2 version and BlockHound version works.

Spring Boot 3.4.1:

<img width="1283" alt="Image" src="https://github.com/user-attachments/assets/0cd4cda6-5c40-4851-905d-3157ea04c30e" />

Spring Boot 3.4.2:

<img width="726" alt="Image" src="https://github.com/user-attachments/assets/49aab7e0-6304-48e0-ad6a-ab0af2e97422" />

wilkinsona:
Thanks for the report and for the minimal. I assume that something has changed in Spring Boot 3.4.2 that is causing BlockHound to break Log4j2 in some way. Unfortunately, given that this works fine without BlockHound and the way in which BlockHound works, I don't think there's anything we can do about this in Spring Boot. Please raise a [BlockHound issue](https://github.com/reactor/BlockHound/issues) so that they can investigate further.

wilkinsona:
Re-opening. Thanks to @violetagg's [analysis](https://github.com/reactor/BlockHound/issues/469#issuecomment-2621694954), this has been tracked down to an assumption in `StatusConsoleListener#closeNonSystemStream` that is faulty when BlockHound is involved. It's only started causing a problem due to https://github.com/spring-projects/spring-boot/issues/43578.

nosan:
Thanks for the sample, @micopiira.

`StandardOutputIntegration.applyTo(...)` replaces the standard OUT and ERR streams with `PrintStreamDelegate`. During initialization, `Log4J2LoggingSystem` attempts to reset the `FallbackListener` stream (which was previously replaced by BlockHound). This results in `PrintStreamDelegate` being closed, which in turn delegates the close call to the standard OUT and ERR streams.

One of the options is to revert https://github.com/spring-projects/spring-boot/commit/b6b9237f2c0f8b3767c04d017bfd0a5e5778d18d and fix `SampleLog4j2StructuredLoggingApplicationTests`.
https://github.com/spring-projects/spring-boot/compare/main...nosan:spring-boot:43963



[...]

PR #44380, opened 2025-02-20
Console output may be lost when using Log4j2 with something that replaces System.out

nosan:
See gh-43963


nosan:
Thanks, @wilkinsona 

I did typo in commit message 🤦  (I used Log42j instead of **Log4j2**)

wilkinsona:
Sorry, I would have corrected that if I'd noticed it. It's no big deal though, we've managed plenty of those over the years so welcome to the club.

Yobyn:
Any idea when 3.4.3 will be available on maven?

spencergibb:
It is there already https://repo1.maven.org/maven2/org/springframework/boot/spring-boot-dependencies/3.4.3/

Issue #44382, opened 2025-02-20
Documentation bug: Inconsistency in Mongo connection setup


michaldo:
On one hand documentation shows how to configure database with uri
https://docs.spring.io/spring-boot/reference/data/nosql.html#data.nosql.mongodb.connecting

> as shown in the following example:

```
spring:
  data:
    mongodb:
      uri: "mongodb://user:secret@mongoserver1.example.com:27017,mongoserver2.example.com:23456/test"
```

On the other hand property reference denies uri configurability

https://docs.spring.io/spring-boot/appendix/application-properties/index.html#application-properties.data.spring.data.mongodb.port

>  spring.data.mongodb.port | Mongo server port. Cannot be set with URI.
spring.data.mongodb.uri | Mongo database URI. Overrides host, port, username, and password.
spring.data.mongodb.database | Database name. Overrides database in URI.
spring.data.mongodb.host | Mongo server host. Cannot be set with URI.
spring.data.mongodb.password | Login password of the mongo server. Cannot be set with URI.
spring.data.mongodb.replica-set-name | Required replica set name for the cluster. Cannot be set with URI.

Unable to understand, basing on documentation, what is allowed in URI and what overrides what

nosan:
If Mongo database URI **spring.data.mongodb.uri** property is specified it will be used as `ConnectionString`, otherwise `ConnectionString` will be built based on other properties such as `spring.data.mongodb.host`, etc.  

https://github.com/spring-projects/spring-boot/blob/109fd6f97da95ab78d20cbaddb2ca817e647cd1c/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/PropertiesMongoConnectionDetails.java#L55
 
The one exception is `spring.data.mongodb.database` property,  this property has higher precedence over `spring.data.mongodb.uri` property. 

https://github.com/spring-projects/spring-boot/blob/a8948790854796789a4087739ed7199a43a0300c/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/mongo/MongoReactiveDataAutoConfiguration.java#L88

https://github.com/spring-projects/spring-boot/blob/593fa1dc42177cc33b076643bd2ba6c66e84a3db/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/data/mongo/MongoDatabaseFactoryConfiguration.java#L49

[...]

PR #44384, opened 2025-02-20
Clarify which Mongo properties are ignored when URI property is set

nosan:
See [#44382](https://github.com/spring-projects/spring-boot/issues/44382)


quaff:
It apply to `RedisProperties` and others.

wilkinsona:
I’m not sure that it does as RedisProperties does not use them same “cannot be set” language that caused confusion here. Let’s focus on Mongo for now. We can consider a broader sweep separately.

PR #44385, opened 2025-02-20
Adapt Javadoc reference of JooqExceptionTranslator to use ExceptionTranslatorExecuteListener

ngocnhan-tran1996:
`JooqExceptionTranslator` will be removed in 3.5.0 and this PR update the reference javadoc link

Reference
https://docs.spring.io/spring-boot/3.5-SNAPSHOT/how-to/data-access.html#howto.data-access.configure-jooq-with-multiple-datasources


snicoll:
It has been removed already and deprecated since 3.3.x.

PR #44388, opened 2025-02-20
Use Javadoc attributes for Gradle reference guide

ngocnhan-tran1996:


snicoll:
Nice, thanks again @ngocnhan-tran1996.

PR #44398, opened 2025-02-21
StandardStackTracePrinter prints root first if configured otherwise

typat:
Makes the StandardStackTracePrinter root option respect the `logging.structured.json.stacktrace.root` property value

I'm assuming based on the documentation that `last` should be the default value, which this respects

```
 Use `last` to print the root item last (same as Java) or `first` to print the root item first.
```


mhalbritter:
Thanks @typat !

PR #44413, opened 2025-02-23
Samples for metadata annotation processers have invalid fold attribute

ngocnhan-tran1996:
When I expand snippet code

image

Reference:

https://docs.spring.io/spring-boot/3.4-SNAPSHOT/specification/configuration-metadata/annotation-processor.html


snicoll:
Thanks again @ngocnhan-tran1996.

PR #44465, opened 2025-02-27
Add support for auto-configuring SimpleMessageListenerContainer

vpavic:
This commit introduces a new `spring.jms.listener.container-type` configuration property which can be used to auto-configure JMS listener support backend by `SimpleMessageListenerContainer` instead of `DefaultMessageListenerContainer`.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/DefaultJmsListenerContainerFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/DefaultJmsListenerContainerFactoryConfigurer.java
@@ -20,18 +20,14 @@ import java.time.Duration;
 
 import io.micrometer.observation.ObservationRegistry;
 import jakarta.jms.ConnectionFactory;
-import jakarta.jms.ExceptionListener;
 
 import org.springframework.boot.autoconfigure.jms.JmsProperties.Listener.Session;
 import org.springframework.boot.context.properties.PropertyMapper;
 import org.springframework.jms.config.DefaultJmsListenerContainerFactory;
-import org.springframework.jms.support.converter.MessageConverter;
-import org.springframework.jms.support.destination.DestinationResolver;
 import org.springframework.transaction.jta.JtaTransactionManager;
-import org.springframework.util.Assert;
 
 /**
- * Configure {@link DefaultJmsListenerContainerFactory} with sensible defaults.
+ * Configures {@link DefaultJmsListenerContainerFactory} with sensible defaults.
  *
  * @author Stephane Nicoll
  * @author Eddú Meléndez
@@ -39,46 +35,10 @@ import org.springframework.util.Assert;
  * @author Lasse Wulff
  * @since 1.3.3
  */
-public final class DefaultJmsListenerContainerFactoryConfigurer {
-
-	private DestinationResolver destinationResolver;
-
-	private MessageConverter messageConverter;
-
-	private ExceptionListener exceptionListener;
+public final class DefaultJmsListenerContainerFactoryConfigurer
+		extends AbstractJmsListenerContainerFactoryConfigurer<DefaultJmsListenerContainerFactory> {
 
 	private JtaTransactionManager transactionManager;
-
-	private JmsProperties jmsProperties;
-
-	private ObservationRegistry observationRegistry;
-
-	/**
-	 * Set the {@link DestinationResolver} to use or {@code null} if no destination
-	 * resolver should be associated with the factory by default.
-	 * @param destinationResolver the {@link DestinationResolver}
-	 */
-	void setDestinationResolver(DestinationResolver destinationResolver) {
-		this.destinationResolver = destinationResolver;
-	}
-
-	/**
-	 * Set the {@link MessageConverter} to use or {@code null} if the out-of-the-box
-	 * converter should be used.
-	 * @param messageConverter the {@link MessageConverter}
-	 */
-	void setMessageConverter(MessageConverter messageConverter) {
-		this.messageConverter = messageConverter;
-	}
-
-	/**
-	 * Set the {@link ExceptionListener} to use or {@code null} if no exception listener
-	 * should be associated by default.
-	 * @param exceptionListener the {@link ExceptionListener}
-	 */
-	void setExceptionListener(ExceptionListener exceptionListener) {
-		this.exceptionListener = exceptionListener;
-	}
 
 	/**
 	 * Set the {@link JtaTransactionManager} to use or {@code null} if the JTA support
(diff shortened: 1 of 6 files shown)


vpavic:
Semantics of `SimpleJmsListenerContainer` are quite different from polling-based approach that `DefaultMessageListenerContainer` are a better fit for some use cases. Your comment sounds like use of `SimpleJmsListenerContainer` is discouraged and I don't see that being the case looking at [_Message Listener Containers_ section](https://docs.spring.io/spring-framework/reference/integration/jms/using.html#jms-mdp) of the Spring Framework reference docs that says this about `SimpleJmsListenerContainer`:

> This message listener container is the simpler of the two standard flavors. It creates a fixed number of JMS sessions and consumers at startup, registers the listener by using the standard JMS MessageConsumer.setMessageListener() method, and leaves it up the JMS provider to perform listener callbacks. This variant does not allow for dynamic adaption to runtime demands or for participation in externally managed transactions. Compatibility-wise, it stays very close to the spirit of the standalone JMS specification, but is generally not compatible with Jakarta EE’s JMS restrictions.

My specific use case is that I'm using SQS and [Amazon SQS Java Messaging Library](https://github.com/awslabs/amazon-sqs-java-messaging-lib) provided by AWS. This JMS implementation doesn't support transactions, but provides custom acknowledge mode that better fits SQS, and internally has its own prefetching/polling mechanism that aligns usage of SQS API with JMS spec. From my experience `SimpleJmsListenerContainer` is a better fit here since I don't want another layer of polling (and associated set of threads) and dynamic concurrency.

I'd like to avoid copying boilerplate configuration from project to project and use configuration property based approach to opt into using `SimpleJmsListenerContainer` while also retaining the ability to configure other message listener aspects using properties.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/jms/DefaultJmsListenerContainerFactoryConfigurer.java
```diff
@@ -89,50 +49,26 @@ void setTransactionManager(JtaTransactionManager transactionManager) {
...
 	/**
 	 * Set the {@link ObservationRegistry} to use.
 	 * @param observationRegistry the {@link ObservationRegistry}
 	 * @since 3.2.1
 	 */
+	@Override
 	public void setObservationRegistry(ObservationRegistry observationRegistry) {
-		this.observationRegistry = observationRegistry;
+		super.setObservationRegistry(observationRegistry);
 	}
```
vpavic, line 60: This override is here to preserve backwards compatibility (see #44466), as `AbstractJmsListenerContainerFactoryConfigurer` introduced in this PR reduces visibility to package private.

snicoll:
> Your comment sounds like use of SimpleMessageListenerContainer is discouraged 

No, it doesn't sound like that. If it was, the class would be deprecated. I wrote "invest", which means we haven't been touching the simple implementation and recommend folks to use the pooling approach. I think we can move forward but I would make the following changes to make the PR actionable:

* Remove the property that switches from one to another. If you want to use the simple implementation, then it's fair to ask folks to create a default. In particular, I don't like the update to configuration properties where half of the properties are now working with one, but not the other
* Update the documentation to explain the difference between the two implementations and why the Simple implementation would be a good fit. I don't want to offer a new option without a justification. Right now, the change is bringing more confusion to regular users.

We can reconsider based on the above, thanks!

snicoll:
Closing due to the lack of feedback.

vpavic:
Sorry for the late follow-up, I've been struggling to find time for open-source activities in the recent weeks.

In my previous comment I've already linked to [_Message Listener Containers_ section](https://docs.spring.io/spring-framework/reference/integration/jms/using.html#jms-mdp) of the Spring Framework reference docs that describes both message listener container implementations quite well, and coupled with documentation present in javadoc of both [`SimpleMessageListenerContainer`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/jms/listener/SimpleMessageListenerContainer.html) and [`DefaultMessageListenerContainer`](https://docs.spring.io/spring-framework/docs/current/javadoc-api/org/springframework/jms/listener/DefaultMessageListenerContainer.html) it provides substantial information for developers to decide which message listener container implementation fits their needs the best.

From there, it's clear that the two have quite different semantics - push-based in `SimpleMessageListenerContainer` vs pull-based in `DefaultMessageListenerContainer`. Additionally the docs say that `SimpleMessageListenerContainer` _stays very close to the spirit of the standalone JMS specification_ and javadoc that _Its main advantage is its low level of complexity and the minimum requirements on the JMS provider_. Depending on JMS provider and workload, these characteristics are actually often preferable (I already described specifics of my setup in previous comment).

My point is, from Spring Framework perspective both message listener container implementations are equally represented in the documentation that very clearly describes pros and cons of each. On the other hand, Spring Boot's auto-configuration just sets up `DefaultMessageListenerContainer` and does not mention the other option in any way. This to me leaves the impression of `SimpleMessageListenerContainer` being discouraged, a feeling that was reinforced by @snicoll mentioning in two separate comments that have choice will cause confusion - I can't see that being the case given how well both message listener container implementations are documented and how long they've actually existed. And having in mind the differences between the two, I don't think Spring Boot should support just one option.

To address the specific requests:

> Update the documentation to explain the difference between the two implementations and why the Simple implementation would be a good fit.

I'm not sure what exactly you had in mind, but I've already mentioned the wealth of  documentation on Framework side - shouldn't Spring Boot's reference documentation somewhere just link to that? I can try to add a sentence or two that do that.

> Remove the property that switches from one to another.

Without that property, what's the benefit? I'm currently doing this in my project:

```java
@Bean
SimpleJmsListenerContainerFactory jmsListenerContainerFactory(ConnectionFactory connectionFactory,
        BeanFactoryDestinationResolver destinationResolver) {
    SimpleJmsListenerContainerFactory factory = new SimpleJmsListenerContainerFactory();
    factory.setConnectionFactory(connectionFactory);
    factory.setDestinationResolver(destinationResolver);
    factory.setSessionAcknowledgeMode(SQSSession.UNORDERED_ACKNOWLEDGE);
    return factory;
}
```

Since this is a very frequent piece of boilerplate configuration that I would need to copy from project to project I was hoping to reduce that to:

```properties
spring.jms.listener.container-type=simple
spring.jms.listener.session.acknowledge-mode=100
```

Seems like a reasonable expectation to me. If I would need to declare `SimpleJmsListenerContainerFactory` bean to have the ability to customize it using properties that doesn't sound too attractive. Also, properties of similar nature exist elsewhere in auto-configuration, for example `spring.rabbitmq.listener.type` (very similar concern) and `spring.session.redis.repository-type` come to mind.

> I don't like the update to configuration properties where half of the properties are now working with one, but not the other

This I agree isn't ideal but isn't too bad IMO, as this pattern exists in some other auto-configurations and such properties already contain _Applies only to container type `default`_ in their description.

Could you please reconsider this PR based on the feedback provided here? I see that it was labeled `for: team-meeting` before being closed and I'm really curious to hear your take on this @philwebb.

snicoll:
> This to me leaves the impression of SimpleMessageListenerContainer being discouraged, a feeling that was reinforced by @snicoll mentioning in two separate comments that have choice will cause confusion

That's an interpretation on your end I am afraid. The choice will cause confusion. Just because you're versed in the topic doesn't mean that the average user is. And, from experience, I've seen quite a bit of people not knowing what to chose. Not providing auto-configuration for the Simple implementation is not oversight but a conscious decision.

> I'm not sure what exactly you had in mind, but I've already mentioned the wealth of documentation on Framework side 

This is a technical doc and one should put the pieces together to form an opinion. What you wrote in your comment would be a very good start in our reference doc, I think.

> Without that property, what's the benefit?

Two things. One, we don't want to confuse users with the option. By your own admission, you use case is quite specific and you'd like better support for it. That's fine, it doesn't mean though that it should lead to more configuration properties. Second, this PR shows that a number of properties are not applicable to the simple implementation. This is problematic and, for a non-mainstream use case, doesn't warrant the switch.

The benefit is that you can inject the builder and easily configure the connection factory you want.

> Could you please reconsider this PR based on the feedback provided here? I see that it was labeled for: team-meeting before being closed and I'm really curious to hear your take on this @philwebb.

I wish you gave me more credit. Phil was wondering about the differences between the two and we obviously discussed this before I closed it.

snicoll:
> I cannot agree on the confusion part you keep insisting on.

That is fine. I've witnessed the confusion time and time again but you don't have to trust me, I guess.

> What about Spring Boot offering spring.rabbitmq.listener.type, spring.session.redis.repository-type

I don't see the argument of providing some "type" property that we already provide. None of those match the conversation that we're having here.

All in all, I think this conversation has reached its course. If you want to amend your PR as I've suggested above, then please do so and I'll be happy to review it. Perhaps some more refinement will occur based on the code. And it is a first good step in what you're trying to achieve and that doesn't prevent us to keep refining it based on the community's feedback.

vpavic:
I've finally circled back to this and pushed the changes to align with expectations from https://github.com/spring-projects/spring-boot/pull/44465#issuecomment-2716988657.

The diff can be seen here:
- https://github.com/spring-projects/spring-boot/compare/main...vpavic:spring-boot:simple-jms-listener-container

snicoll:
Awesome @vpavic, thank you! If we can reopen this then let's do that. Or please create a separate PR.

vpavic:
I can't reopen it because I don't have the permission to do so. The changes are in the same branch that was used when PR was initially created, so if you reopen it I'm pretty sure everything will be available here. If that doesn't happen ping me here and I can open a separate PR.

snicoll:
I can't reopen the pull request either. We'll have to go with another one then :-/

vpavic:
Here it is:

- https://github.com/spring-projects/spring-boot/pull/47716

Issue #44419, opened 2025-02-24
Some JdbcTemplate configuration properties are missing


quaff:
Currently, there are three configuration properties under `spring.jdbc.template`:
- `fetch-size`
- `max-rows`
- `query-timeout`

there are more configuration properties not covered by `JdbcProperties`:
- `ignore-warnings`
- `skip-results-processing`
- `skip-undeclared-results`
- `results-map-case-insensitive`

Before starting to work on this, I would ask the team is it worthy to autoconfigure them by `JdbcProperties` or provide an `JdbcTemplateCustomizer` or both?

philwebb:
I personally lean towards adding the properties but not the customizer since replacing the `JdbcTemplate` bean entirely isn't that onerous. Flagging to see what others in the team think.

philwebb:
I wonder if an enum set might also work well for the `results` options? E.g. `spring.jdbc.template.results=skip-processing,skip-undeclared,case-insensitive-map`

quaff:
> I wonder if an enum set might also work well for the `results` options? E.g. `spring.jdbc.template.results=skip-processing,skip-undeclared,case-insensitive-map`

Is `callable-results` a better name? and defaults to `null` or add another option such as `default` or `case-sensitive-map`?

wilkinsona:
Just properties sounds good to me.

> and defaults to `null` or add another option such as `default` or `case-sensitive-map`

Along these lines, I'm not sure about a single property as it creates a problem with switching things off. For example, if someone's set `spring.jdbc.template.results=skip-undeclared` how do we go back to the defaults? A `default` value in the enum would help with that, but then what does a value of `default, skip-undeclared` mean?

I think three separate properties, each defaulting to `false`, might be better unfortunately.

quaff:
Superseded by GH-44470

PR #44470, opened 2025-02-28
Add configuration properties for JdbcTemplate's ignoreWarnings, skipResultsProcess, skipUndeclaredResults, and resultsMapCaseInsensitive

quaff:
- ignore-warnings
- skip-results-processing
- skip-undeclared-results
- results-map-case-insensitive

Fix GH-44419


wilkinsona:
Flagging for team attention to make sure everyone else is happy with the separate property approach.

snicoll:
Read https://github.com/spring-projects/spring-boot/issues/44419#issuecomment-2688767488 and I agree (unfortunately) that three properties are better.

snicoll:
@quaff since you're a prolific contributor, can you please review [the polish commit](https://github.com/spring-projects/spring-boot/commit/0e09a2d25f8ff21a6314a83bd7e0df9e5185b2a6). We use standard patterns for configuration properties, when in doubt please look other classes.

quaff:
I'm confused about the order of fields, it's lexicographic ordered, but not exactly.

snicoll:
It matches the order in `JdbcTemplate`.

PR #44477, opened 2025-02-28
Update LogbackLoggingSystemTests to use Console.charset() when available

nosan:
Relates to #44353


mhalbritter:
Thanks @nosan !

Issue #44399, opened 2025-02-21
Application initialization fails if a log4j2 override file is not present


rgoers:
Log4j2LoggingSystem contains the following method:
```
	protected void loadConfiguration(String location, LogFile logFile, List<String> overrides) {
		Assert.notNull(location, "'location' must not be null");
		try {
			List<Configuration> configurations = new ArrayList<>();
			LoggerContext context = getLoggerContext();
			configurations.add(load(location, context));
			for (String override : overrides) {
				configurations.add(load(override, context));
			}
			Configuration configuration = (configurations.size() > 1) ? createComposite(configurations)
					: configurations.iterator().next();
			context.start(configuration);
		}
		catch (Exception ex) {
			throw new IllegalStateException("Could not initialize Log4J2 logging from " + location, ex);
		}
	}
```
If an error occurs loading an override file the exception will percolate to the catch block and fail loading the logging configuration. The log4j-spring-boot module provided by Log4j had:
```
                        try {
                            final Configuration config =
                                    ConfigurationFactory.getInstance().getConfiguration(ctx, source);
                            if (config instanceof AbstractConfiguration) {
                                configs.add((AbstractConfiguration) config);
                            } else {
                                LOGGER.warn(
                                        "Configuration at {} cannot be combined in a CompositeConfiguration",
                                        sourceLocation);
                                return;
                            }
                        } catch (Exception ex) {
                            if (!first) {
                                LOGGER.warn(
                                        "Error accessing {}: {}. Ignoring override", sourceLocation, ex.getMessage());
                            } else {
                                throw ex;
                            }
                        }
```
So with log4j-spring-boot a warning message would be logged that the override file couldn't be loaded but configuration would still take place with the primary config file. 

This behavior in Log4j2 was intentional. All the services I work with specify a shared primary logging configuration and a per application override location such as

```
logging:
  label: ${spring.cloud.config.label:master}
  config: "https://spring-configuration-server.acme.org/some-service/default/${logging.label}/log4j2-json-dev.xml"
  log4j2:
    config:
      override: "https://spring-configuration-server.acme.org/some-service/default/${logging.label}/log4j2-org-some-service-dev.xml"
```
If the application doesn't need to override anything then it just uses the primary configuration. If we want to add an override we can just add it to Spring Cloud Config without having to change the application. 

In short, the configurations.add() call needs to be placed in its own try/catch block that logs a similar message to what log4j-spring-boot was doing and ignore the failure.

[...]

PR #44488, opened 2025-02-28
Support optional: prefix with logging.log4j2.config.override

nosan:
See gh-44399


nosan:
Converted to draft
See https://github.com/spring-projects/spring-boot/issues/44399#issuecomment-2690424688

snicoll:
Thanks again, Dmytro.

nosan:
Thanks, @snicoll 

I've noticed that you used:

```
[NOTE]
 ====
 ```
 
Instead of `NOTE:`
 
I wanted to ask if there is an issue with using just `NOTE:`? I am trying to understand the recommended approach for future PRs.

snicoll:
I wanted to make the note multi-lines. In retrospect I suppose it would have worked with `NOTE:` and I am not sure we have a standardization for this.

PR #44493, opened 2025-02-28
Remove default value for OtlpMetricsProperties.url

izeye:
Unlike Micrometer, Spring Boot's Micrometer integration doesn't support the `OTEL_EXPORTER_OTLP_ENDPOINT` and `OTEL_EXPORTER_OTLP_METRICS_ENDPOINT` environment variables due to its default value.

This PR removes the default value for the `OtlpMetricsProperties.url` property to try to align with Micrometer.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
@@ -175,6 +175,7 @@ dependencies {
 	testImplementation("org.springframework.restdocs:spring-restdocs-webtestclient")
 	testImplementation("org.springframework.security:spring-security-test")
 	testImplementation("org.yaml:snakeyaml")
+	testImplementation("uk.org.webcompere:system-stubs-jupiter:2.1.7")
 
 	testRuntimeOnly("jakarta.management.j2ee:jakarta.management.j2ee-api")
 	testRuntimeOnly("jakarta.transaction:jakarta.transaction-api")
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsProperties.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsProperties.java
@@ -39,7 +39,7 @@ public class OtlpMetricsProperties extends StepRegistryProperties {
 	/**
 	 * URI of the OTLP server.
 	 */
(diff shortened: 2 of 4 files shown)


mhalbritter:
This would also align with `OtlpLoggingProperties` and `OtlpTracingProperties`, which have no default for their `endpoint` property.

spring-boot-project/spring-boot-actuator-autoconfigure/build.gradle
```diff
@@ -175,6 +175,7 @@ dependencies {
 	testImplementation("org.springframework.restdocs:spring-restdocs-webtestclient")
 	testImplementation("org.springframework.security:spring-security-test")
 	testImplementation("org.yaml:snakeyaml")
+	testImplementation("uk.org.webcompere:system-stubs-jupiter:2.1.7")
```
mhalbritter, line 178: This library has been added for mocking environment variables. But on the [GitHub page](https://github.com/webcompere/system-stubs) it says:

> From JDK16 onwards, there are deeper restrictions on the ability to use reflection. [...] Consequently, this library now uses bytebuddy to enable the interception of calls for reading environment variables. This might interact with your chosen version of Mockito or other libraries.

I wanted to know the team's opinion if we want to have that library for tests or if we want to ditch the tests (`whenPropertiesUrlIsNotSetAndOtelExporterOtlpEndpointIsSetAdapterUrlUsesIt` and `whenPropertiesUrlIsNotSetAndOtelExporterOtlpMetricsEndpointIsSetAdapterUrlUsesIt`).

nosan: It seems like these tests verify the Micrometer `OtlpConfig` implementation rather than the `OtlpMetricsPropertiesConfigAdapter`.

I believe `OtlpMetricsPropertiesConfigAdapter` tests can be a little updated to ensure that
adapter delegates to `OtlpConfig.url()` when the URL is not set.

```java

@Test
void whenPropertiesUrlIsNotSetAdapterUrlReturnsDefault() {
	assertThat(this.properties.getUrl()).isNull();
	assertThat(createAdapter().url()).isEqualTo("http://localhost:4318/v1/metrics");
}

@Test
void whenPropertiesUrlIsNotSetAndUseOtlpConfigUrlIsFallback() {
	OtlpMetricsPropertiesConfigAdapter adapter = spy(createAdapter());
	when(adapter.get("management.otlp.metrics.export.url")).thenReturn("https://my-endpoint/v1/metrics");
	assertThat(adapter.url()).isEqualTo("https://my-endpoint/v1/metrics");
}

```

In that case, `uk.org.webcompere:system-stubs-jupiter:2.1.7`  is unnecessary.

mhalbritter:
Thanks @izeye !

PR #44394, opened 2025-02-21
Add Support for OTEL_SERVICE_NAME and OTEL_RESOURCE_ATTRIBUTES Environment Variables

nosan:
This update introduces support for OTEL-specific environment variables. The behavior now aligns with `OtlpMetricsPropertiesConfigAdapter`, which automatically handles this through `OtlpConfig.resourceAttributes`.

The attribute order for `OpenTelemetryResourceAttributes` is as follows:
-  `management.opentelemetry.resource-attributes`
-  `OTEL_RESOURCE_ATTRIBUTES` and `OTEL_SERVICE_NAME` environment variables
-  Spring Environment 'spring.application.name' and 'spring.application.group' always apply if absent.

See gh-43712

For consistency with `OtlpMetricsPropertiesConfigAdapter`, I do not merge attributes from `management.opentelemetry.resource-attributes` with attributes from environment variables.

https://github.com/nosan/spring-boot/blob/fdcc8d9d1f632f00dd71093d604c374b6d0a38d3/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsPropertiesConfigAdapter.java#L79-L87

PR #44494, opened 2025-02-28
Refine the handling of OpenTelemetry resource attributes

nosan:
Relates to #44394 

Initially, I considered proposing these changes: https://github.com/spring-projects/spring-boot/compare/main...nosan:spring-boot:gh-44394-1.
However, I eventually realized that it might be too much for this specific case.

This PR aims to eliminate duplications in `OpenTelemetryAutoConfiguration` and `OtlpMetricsPropertiesConfigAdapter`, consolidating everything into a single location.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsPropertiesConfigAdapter.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsPropertiesConfigAdapter.java
@@ -17,6 +17,7 @@
 package org.springframework.boot.actuate.autoconfigure.metrics.export.otlp;
 
 import java.util.Collections;
+import java.util.LinkedHashMap;
 import java.util.Map;
 import java.util.concurrent.TimeUnit;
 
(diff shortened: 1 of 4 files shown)


nosan:
Another option is to leave things as they are and simply add logic to ignore empty keys.

> The attribute key MUST be a non-null and non-empty string.
> Case sensitivity of keys is preserved. Keys that differ in casing are treated as distinct
> keys.

https://github.com/spring-projects/spring-boot/compare/main...nosan:spring-boot:44394

mhalbritter:
Thanks @nosan !

FyiurAmron:
@mhalbritter @nosan sorry for chiming in so late, but I found this in the current code:
```
	/**
	 * Default value for service name if {@code service.name} is not set.
	 */
	private static final String DEFAULT_SERVICE_NAME = "unknown_service";
```
&nbsp;- shouldn't it say (similarly what it said before)
```
	/**
	 * Default value for service name if neither {@code service.name} nor {@code spring.application.name} is set.
	 */
	private static final String DEFAULT_SERVICE_NAME = "unknown_service";
```
&nbsp;- because, based on the codes and later comments, both would be checked?
```
attributes.computeIfAbsent("service.name", (key) -> getApplicationName());
// ...
	private String getApplicationName() {
		return this.environment.getProperty("spring.application.name", DEFAULT_SERVICE_NAME);
	}
```

nosan:
I am not sure I understood your question. The current arrangement is:
- If the user specifies **service.name** through `management.opentelemetry.resource-attributes.service.name`, it will be used.
- If the user specifies the environment variable `OTEL_SERVICE_NAME`, it will be used.
- If neither the environment variable nor the property is specified, then `spring.application.name` will be used.
- If `spring.application.name` is not set, then `unknown_service` will be used.

nosan:
In simple terms, if `service.name` is not set by any of the above methods, `"unknown_service"` will be used.

FyiurAmron:
@nosan that seems like what I would expect. The current `DEFAULT_SERVICE_NAME` javadoc is lacking in that case, precisely because it lost the 2nd part of the OR (previously, it was the only part that it had). Yeah, I know it's a private const, but people debugging stuff read those docs :)

BTW you probably read the previous version of my comment, I've edited it to make it obvious what I was expecting here.

philwebb:
I've polished the javadoc in 22781fc6a0867e0f44d324bd73827add878fbfe3

PR #44497, opened 2025-02-28
Upgrade to Kotlin 2.1.10

dmitrysulman:
Upgrade to **Kotlin 2.1.10**:
- Upgrade `kotlinVersion` to `2.1.10`
- Remove `org.jetbrains.kotlin:kotlin-compiler-embeddable` from the classpath, see: https://kotl.in/gradle/internal-compiler-symbols
- Replace deprecated `kotlinOptions` with `compilerOptions`, see: https://kotl.in/u1r8ln
- Fix `KotlinConventions` of `ConventionsPlugin` to set Kotlin `apiVersion` and `languageVersion` to `2.1`
- Fix `PluginClasspathGradleBuild` to inject missed Kotlin K2 compiler related jars for running Kotlin Gradle plugin tests on Gradle 8.11+, see Gradle compatibility matrix for Kotlin: https://docs.gradle.org/current/userguide/compatibility.html#kotlin
- ~~Move all Kotlin examples of `spring-boot-doc` to `kotlin` subpackage to prevent `Redeclaration:` compile errors, see: https://youtrack.jetbrains.com/issue/KTLC-85/Forbid-Java-and-Kotlin-classes-with-the-same-FQ-name-in-sources. **This fix introduces most of the changes in the commit.**~~
- Exclude `javaSources` from the `KotlinCompile` task of `spring-boot-doc`, except for Java classes that are used by some Kotlin example classes, to avoid a redeclaration error with the Kotlin compiler.

See spring-projectsgh-38539


wilkinsona:
Thanks for the PR. Unfortunately, moving the doc examples to a `kotlin` subpackage will cause problems with docs. Instead, we prefer this `build.gradle` approach:

```
// To avoid a redeclaration error with Kotlin compiler
tasks.named('compileKotlin', KotlinCompilationTask.class) {
    javaSources.from = []
}
```

It hides the Java sources from Kotlin compilation.

dmitrysulman:
@wilkinsona thanks for the feedback.

> Unfortunately, moving the examples to a kotlin subpackage will cause problems with docs.

Yes, I wasn't sure about this approach. I changed `include-kotlin` property value in the `antora-asciidoc-attributes.properties`:
```
include-kotlin= ROOT:example$kotlin/org/springframework/boot/docs/kotlin
```
and was able to build the docs locally without a problem.

Anyway I'll fix the PR according to your suggestion.

spring-boot-project/spring-boot-docs/build.gradle
```diff
@@ -304,6 +306,26 @@ task runLoggingFormatExample(type: org.springframework.boot.build.docs.Applicati
 	normalizeTomcatPort()
 }
 
+// To avoid a redeclaration error with Kotlin compiler
+tasks.withType(KotlinCompile) {
+	javaSources.from = [
```
dmitrysulman, line 311: These are the Java classes that are used by some of the Kotlin examples. We have 14 such classes.

Another approach is to completely decouple the Java and Kotlin examples and add a Kotlin replacement for each of the below classes. But this must be done carefully since some of these classes are used directly in the `include-code` block of the docs, and having a Kotlin version of such classes is unwanted. To prevent Asciidoctor from picking them up, we need to place them in a package with a name that differs from the Java version.

snicoll:
For the record we’ve already done the work of upgrading to Kotlin2. I’d rather us taking this forward and then review whatever needs to be. 

See 4c6b49351c386907ddbfe97c3c4aa0e8242671ed

dmitrysulman:
@snicoll thanks for the information, didn't know about this. I'll be happy with any decision made by the Spring Boot team.

Let me add a couple of comments to the referenced commit.

1.
> As of Kotlin 2, it is now longer possible to have a Java type and a
> Kotlin type with the same name. As our code samples follow that
> unfortunate pattern, this commit makes sure that the Kotlin sample code
> does not depend on any of the Java counterpart and configure the kotlin
> compilation plugin to ignore Java sources.

As I mentioned [earlier](https://github.com/spring-projects/spring-boot/pull/44497#discussion_r1976389388) some of these Java code samples are used directly in the `include-code` block of the docs. For example [Nested](https://github.com/spring-projects/spring-boot/commit/4c6b49351c386907ddbfe97c3c4aa0e8242671ed#diff-947f2053bf57cf4e46e21448d2f42dd8a28015078479c433c6c0ca6d9103e7fe) class is used [here](https://docs.spring.io/spring-boot/reference/packaging/native-image/advanced-topics.html#packaging.native-image.advanced.nested-configuration-properties):
Screenshot 2025-03-02 at 14 44 17
Now in the `Kotlin` tab of this code block we will have an empty `Nested` class which is undesirable. A possible solution is to place such classes in a different package to prevent them from being picked up by Asciidoctor.


2.
> The minimum version of Gradle is 7.6.4. It bundles a version of Kotlin that
> cannot compile a Kotlin build script when spring-core, compiled with Kotlin
> 2.1, is on the classpath. Using Gradle 8.12 to run the DSL tests avoids
> the problem.

It’s still possible to compile Kotlin DSL build scripts using Gradle 7.6.4, even when Spring itself is compiled with Kotlin 2.0+. The issue arises due to the presence of multiple Kotlin JARs compiled with Kotlin 2.0+ on the build script classpath, specifically `kotlin-build-tools-api` and `kotlin-native-utils`. To resolve the issue, these JARs should be excluded from the build script classpath for Gradle versions that embed a Kotlin version prior to 2.0.

wilkinsona:
@dmitrysulman please don't use `4.0.x` as a base branch at this time. It's only intended for collaboration among the core team at this time and it's likely that it won't be in a position to accept external contributions for several weeks if not months.

dmitrysulman:
@wilkinsona I see, thanks.

dmitrysulman:
Closing this PR since it's not relevant.

PR #44499, opened 2025-03-01
Add support for OpenTelemetry's service.namespace

### LINKED ISSUE #44411: Use 'service.namespace' instead of 'service.group' for OpenTelemetry
As per the OpenTelemetry documentation (https://opentelemetry.io/docs/specs/semconv/attributes-registry/service), **service.group** does not exist. Instead, **service.namespace** should be used for this purpose.

**service.namespace:**
>  A string value having a meaning that helps to distinguish a group of services, for example the team name that owns a group of services. service.name is expected to be unique within the same namespace. If service.namespace is not specified in the Resource then service.name is expected to be unique for all services that have no explicit namespace defined (so the empty/unspecified namespace is simply one more valid namespace). Zero-length namespace string is assumed equal to unspecified namespace.

**service.group** is set up in both `OpenTelemetryAutoConfiguration` and `OtlpMetricsPropertiesConfigAdapter`

thecooldrop:
Add OTEL service attribute `service.namespace`. The value is set to same value as `service.group` currently. Since there is no specific environment variable to specify the `service.namespace` prescribed by OTEL, following chain is used to decide the value:

- management.opentelemetry.resource-attributes configuration property
- OTEL_RESOURCE_ATTRIBUTES environment variable
- If `service.group` is configured, then set it to same name
- Otherwise empty

Closes #44411

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryResourceAttributes.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryResourceAttributes.java
@@ -99,6 +99,7 @@ public final class OpenTelemetryResourceAttributes {
 		});
 		attributes.computeIfAbsent("service.name", (k) -> getApplicationName());
 		attributes.computeIfAbsent("service.group", (k) -> getApplicationGroup());
+		attributes.computeIfAbsent("service.namespace", (key) -> getServiceNamespace());
 		attributes.forEach(consumer);
 	}
 
@@ -106,9 +107,16 @@ public final class OpenTelemetryResourceAttributes {
 		return this.environment.getProperty("spring.application.name", DEFAULT_SERVICE_NAME);
 	}
 
+	@Deprecated(since = "3.5.0", forRemoval = true)
+	// See https://github.com/spring-projects/spring-boot/issues/44411 for potential
+	// information about deprecation of "service.group" attribute
 	private String getApplicationGroup() {
 		String applicationGroup = this.environment.getProperty("spring.application.group");
 		return (StringUtils.hasLength(applicationGroup)) ? applicationGroup : null;
+	}
+
+	private String getServiceNamespace() {
+		return this.environment.getProperty("spring.application.group");
 	}
 
 	/**
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryAutoConfigurationTests.java
@@ -107,6 +107,22 @@ class OpenTelemetryAutoConfigurationTests {
 	}
 
 	@Test
+	void shouldApplyServiceNamespaceIfApplicationGroupIsSet() {
+		this.runner.withPropertyValues("spring.application.group=my-group").run((context) -> {
+			Resource resource = context.getBean(Resource.class);
+			assertThat(resource.getAttributes().asMap()).containsEntry(AttributeKey.stringKey("service.namespace"), "my-group");
+		});
+	}
+
+	@Test
+	void shouldNOtApplyServiceNamespaceIfApplicationGroupIsNotSet() {
+		this.runner.run((context -> {
+			Resource resource = context.getBean(Resource.class);
+			assertThat(resource.getAttributes().asMap()).doesNotContainKey(AttributeKey.stringKey("service.namespace"));
+		}));
+	}
+
+	@Test
 	void shouldFallbackToDefaultApplicationNameIfSpringApplicationNameIsNotSet() {
 		this.runner.run((context) -> {
 			Resource resource = context.getBean(Resource.class);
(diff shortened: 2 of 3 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryAutoConfiguration.java
```diff
@@ -80,6 +80,9 @@ private Resource toResource(Environment environment, OpenTelemetryProperties pro
 			.asMap();
 		attributes.computeIfAbsent("service.name", (key) -> getApplicationName(environment));
 		attributes.computeIfAbsent("service.group", (key) -> getApplicationGroup(environment));
+		// See https://github.com/spring-projects/spring-boot/issues/44411 for potential
+		// information about deprecation of "service.group" attribute
```
nosan, line 84: I think that instead of reusing the existing `getApplicationGroup` method, a new method,
`getServiceNamespace`, can be introduced. This would make it possible to mark the
existing method with the `@Deprecated` annotation and include appropriate javadoc.

```java
private String getServiceNamespace(Environment environment) {
    return environment.getProperty("spring.application.group");
}
/**
 * @deprecated ...
 */
@Deprecated(since = "3.5.0", forRemoval = true)
private String getApplicationGroup(Environment environment) {
}
```

Using the `@Deprecated` annotation would make it easier to ensure that the
`service.group` property is deleted in future versions.


The same could be done in `OtlpMetricsPropertiesConfigAdapter`

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsPropertiesConfigAdapterTests.java
```diff
@@ -163,6 +163,17 @@ void shouldUseDefaultApplicationGroupIfApplicationGroupIsNotSet() {
 		assertThat(createAdapter().resourceAttributes()).doesNotContainKey("service.group");
 	}
 
+	@Test
+	void shouldUseServiceGroupForServiceNamespaceIfServiceGroupIsSet() {
```
nosan, line 167: It would be great to include a similar test in `OpenTelemetryAutoConfigurationTests` as well. Since `OpenTelemetryAutoConfigurationTests` already contains a test for checking `service.group`, you could potentially reuse that.

nosan (commented):
Thank you so much, @thecooldrop
I've left some comments for your consideration.

Please don't do anything about the above until Spring Boot team had a chance to review it.

I’ve opened https://github.com/spring-projects/spring-boot/pull/44494 to eliminate the duplications in `OpenTelemetryAutoConfiguration` and `OtlpMetricsPropertiesConfigAdapter`, consolidating everything in one place.

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsPropertiesConfigAdapterTests.java
thecooldrop: I was looking for that in exactly that class, but somehow I missed it in the mass. 

Thanks for pointing this out. I will add this.

spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryAutoConfiguration.java
mhalbritter: Thanks @nosan, i think that's a good idea.

mhalbritter:
Could you please also rebase your changes on top of `main`? There has been some changes to the OpenTelemetry variables handling,

spring-boot-project/spring-boot-actuator-autoconfigure/src/test/java/org/springframework/boot/actuate/autoconfigure/metrics/export/otlp/OtlpMetricsPropertiesConfigAdapterTests.java
thecooldrop: Given that entire logic for determining the attributes, which are included is now contained in `OpenTelemetryResourceAttributes`, I will place the tests only once in this class.

thecooldrop: There is a bit of a weird corner case. If `service.group` is populated via `OTEL_RESOURCE_ATTRIBUTES` env variable, and overrides `spring.application.group` configuration from `Environment`, then the question is if `service.namespace` should be populated from `OTEL_RESOURCE_ATTRIBUTES` as well, or from the `service.application.group`.

My expectation would be to populate it from `service.application.group`, as long as it is not listed in `OTEL_RESOURCE_ATTRIBUTES`

This choice is reflected in following test (note the similar, but different values for two entries):


```java
	@Test
	void otelResourceAttributesShouldTakePrecedenceOverSpringApplicationGroupName() {
		this.environmentVariables.put("OTEL_RESOURCE_ATTRIBUTES", "service.group=spring-boot");
		this.environment.setProperty("spring.application.group", "spring-boot-app");
		assertThat(getAttributes()).hasSize(3)
			.containsEntry("service.name", "unknown_service")
			.containsEntry("service.group", "spring-boot")
			.containsEntry("service.namespace", "spring-boot-app");
	}

```

thecooldrop:
> Could you please also rebase your changes on top of `main`? There has been some changes to the OpenTelemetry variables handling.

Done

PR #44505, opened 2025-03-03
Add Lifecycle/CRaC support for Cassandra

rvansa:
Adds a bean that oversees Cassandra's `CqlSession` through `Lifecycle`; upon `stop()` all connections are forcefully terminated by marking all nodes as forced down, and unmarking these on `start()`. This might not be perfect if there's an ongoing topology change, but for a quiescent state it should be sufficient.

In order to support checkpoint `onRefresh` which does not stop beans that haven't been started (see disussion in https://github.com/spring-projects/spring-framework/issues/34510#issuecomment-2690319525 ) this registers a CRaC resource that will handle the checkpoint. After properly started the resource is removed left to be garbage-collected; potential interference between lifecycle and CRaC resource methods invocation is handled.

I have been trying to mimic `HikariCheckpointRestoreLifecycle` and I have added some tests similar to those in `HikariDataSourceConfigurationTests`; I would do the same for `HikariCheckpointRestoreLifecycleTests` but I am not sure where these should land, as the `:spring-boot` project does not depend on `org.springframework.data:spring-data-cassandra`. Please advise.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/build.gradle
+++ b/spring-boot-project/spring-boot-autoconfigure/build.gradle
@@ -232,6 +232,8 @@ dependencies {
 	optional("org.thymeleaf.extras:thymeleaf-extras-springsecurity6")
 	optional("redis.clients:jedis")
 
+	compileOnly("org.crac:crac")
+
 	testImplementation(project(":spring-boot-project:spring-boot-tools:spring-boot-test-support"))
 	testImplementation(project(":spring-boot-project:spring-boot-test"))
 	testImplementation("ch.qos.logback:logback-classic")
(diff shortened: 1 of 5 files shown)


wilkinsona:
Thanks for the PR. If we accept the contribution, I would consider it to be a new feature so the base branch should be `main` rather than `3.4.x` please.

@sdeleuze do you think this is worth supporting? I am sceptical that it is for the checkpoint-restore case but if it may benefit CDS as well then it may be worth the ongoing maintenance cost. Perhaps we need to wait for https://github.com/spring-projects/spring-framework/issues/34510 to run its course before considering anything on the Boot side.

sdeleuze:
@wilkinsona I am wondering the same. If such improvement can benefit to both CRaC and CDS, maybe it would be worth to consider. In order to have more information to decide, it would be interested to know if other databases have the same issue to estimate the scope of such effort if we want to unlock those use cases consistently. @christophstrobl I think you did some related analysis, if that's the case, could you please share which databases exposing directly lifecycle oriented beans are impacted?

mp911de:
Looking at the change, while it seems possible to restart driver connections, there's a lot of reliance on driver internals. Having lifecycle methods in the driver itself would create a more sustainable approach that doesn't require us to know the driver's internals.

wilkinsona:
Thanks, @mp911de. I agree.

@rvansa please open a Cassandra issue to see if they'll consider some lifecycle methods that would be useful for both coordinated restore at checkpoint and CDS training runs. Please comment here with a link to that issue so that we can follow along. Should something be added to Cassandra, a Spring Boot issue or pull request to make use of the new capabilities would be welcome.

PR #44526, opened 2025-03-04
Fix javadoc link

ngocnhan-tran1996:
Reference
https://docs.spring.io/spring-boot/3.4-SNAPSHOT/gradle-plugin/running.html#running-your-application.passing-arguments


mhalbritter:
Thanks @ngocnhan-tran1996 !

PR #44629, opened 2025-03-06
Fix AliasX509ExtendedKeyManager to process both server and client aliases properly

ecnabogs:
SpringBoot provides a default implementation of the `org.springframework.boot.ssl.SslManagerBundle` interface (`DefaultSslManagerBundle`) that allows to specify a `SslBundleKey` (referencing the private key within the associated `KeyStore`) to be used for establishing an SSL connection. This feature, added in SpringBoot 3.1.0, is really appreciable as it simplifies a lot the selection of the private key to be used during SSL handshake.
However, the thing is that the underlying implementation (`AliasX509ExtendedKeyManager`) only considers the server-side of an SSL connection. If a client has to communicate with multiple partners through mTLS protocol and has a single KeyStore containing its private keys, it cannot select the appropriate key alias to be used with the related partner.

My proposal is to support this use case and to make  `AliasX509ExtendedKeyManager` working similarly for both server and client sides. That's the first point.

Then, looking more in details at the implementation, I noticed that the selected alias was returned unconditionally at the server side. I think it is not correct as SSL handshake is expected to reach an agreement between client and server, especially regarding the key algorithm to be used. Here, this algorithm is not taken into account meaning that although the key with the specified alias could exist in the underlying `KeyStore`, it could also be of the wrong type.
Thus, my implementation gets inspired by the JDK default implementation (`sun.security.ssl.SunX509KeyManagerImpl`) in which the alias selection is made by calling the getServerAliases(keyType, ...) or getClientAliases(keyType, ...) for restricting the choice to compatible keys only.

Finally, the `AliasX509ExtendedKeyManager` current implementation also only considers the transport-independent [SSLEngine](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/javax/net/ssl/SSLEngine.html) but not the `Socket`-specific one. To cover all use cases, the implementation should also apply the same alias selection logic to the `Socket`-specific methods. That's what I did actually and factorized this logic for all contexts.

### DIFF
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/ssl/AliasKeyManagerFactoryTests.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/ssl/AliasKeyManagerFactoryTests.java
@@ -23,32 +23,298 @@ import javax.net.ssl.KeyManagerFactory;
 import javax.net.ssl.X509ExtendedKeyManager;
 
 import org.junit.jupiter.api.Test;
+import org.mockito.ArgumentMatcher;
 
 import static org.assertj.core.api.Assertions.assertThat;
+import static org.mockito.ArgumentMatchers.any;
+import static org.mockito.ArgumentMatchers.argThat;
+import static org.mockito.ArgumentMatchers.eq;
 import static org.mockito.BDDMockito.given;
 import static org.mockito.Mockito.mock;
 
 /**
- * Tests for {@link AliasKeyManagerFactory}.
+ * Tests for {@link org.springframework.boot.ssl.AliasKeyManagerFactory}.
  *
  * @author Phillip Webb
+ * @author Stéphane Gobancé
  */
 class AliasKeyManagerFactoryTests {
 
 	@Test
-	void chooseEngineServerAliasReturnsAlias() throws Exception {
-		KeyManagerFactory delegate = mock(KeyManagerFactory.class);
-		given(delegate.getKeyManagers()).willReturn(new KeyManager[] { mock(X509ExtendedKeyManager.class) });
-		AliasKeyManagerFactory factory = new AliasKeyManagerFactory(delegate, "test-alias",
-				KeyManagerFactory.getDefaultAlgorithm());
+	void chooseEngineServerAliasReturnsTestAliasMatchingSupportedDelegateAliasAndAlgorithm() throws Exception {
+
+		final String delegateSupportedAlgorithm = "supported-alg";
+		final String[] delegateSupportedAlias = new String[] { "alias0", "alias1", "alias2" };
+		final String delegateChosenAlias = delegateSupportedAlias[0];
+		final String testAlias = delegateSupportedAlias[1];
+		final String testAlgorithm = delegateSupportedAlgorithm;
+		final String expectedAlias = testAlias;
+
+		KeyManagerFactory delegate = mockServerKeyManagerFactory(delegateSupportedAlgorithm, delegateSupportedAlias,
+				delegateChosenAlias);
+		AliasKeyManagerFactory factory = createAliasKeyManagerFactory(testAlias, delegate);
+		X509ExtendedKeyManager x509KeyManager = getX509ExtendedKeyManager(factory);
+
+		String chosenAlias = x509KeyManager.chooseEngineServerAlias(testAlgorithm, null, null);
+		assertThat(chosenAlias).isEqualTo(expectedAlias);
+	}
+
+	@Test
+	void chooseEngineServerAliasReturnsDelegateAliasWhenTestAliasIsUnknown() throws Exception {
+
+		final String delegateSupportedAlgorithm = "supported-alg";
+		final String[] delegateSupportedAlias = new String[] { "alias0", "alias1", "alias2" };
+		final String delegateChosenAlias = delegateSupportedAlias[0];
+		final String testAlias = "unknown-alias";
+		final String testAlgorithm = delegateSupportedAlgorithm;
+		final String expectedAlias = delegateChosenAlias;
+
+		KeyManagerFactory delegate = mockServerKeyManagerFactory(delegateSupportedAlgorithm, delegateSupportedAlias,
+				delegateChosenAlias);
+		AliasKeyManagerFactory factory = createAliasKeyManagerFactory(testAlias, delegate);
+		X509ExtendedKeyManager x509KeyManager = getX509ExtendedKeyManager(factory);
+
+		String chosenAlias = x509KeyManager.chooseEngineServerAlias(testAlgorithm, null, null);
+		assertThat(chosenAlias).isEqualTo(expectedAlias);
+	}
+
+	@Test
+	void chooseEngineServerAliasReturnsNullWhenAlgorithmDoesNotMatch() throws Exception {
+
+		final String delegateSupportedAlgorithm = "supported-alg";
+		final String[] delegateSupportedAlias = new String[] { "alias0", "alias1", "alias2" };
+		final String delegateChosenAlias = delegateSupportedAlias[1];
+		final String testAlias = delegateSupportedAlias[0];
+		final String testAlgorithm = "other-alg";
+		final String expectedAlias = null;
+
+		KeyManagerFactory delegate = mockServerKeyManagerFactory(delegateSupportedAlgorithm, delegateSupportedAlias,
+				delegateChosenAlias);
+		AliasKeyManagerFactory factory = createAliasKeyManagerFactory(testAlias, delegate);
+		X509ExtendedKeyManager x509KeyManager = getX509ExtendedKeyManager(factory);
+
+		String chosenAlias = x509KeyManager.chooseEngineServerAlias(testAlgorithm, null, null);
+		assertThat(chosenAlias).isEqualTo(expectedAlias);
+	}
+
+	@Test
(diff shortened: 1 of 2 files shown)


ecnabogs:
Hi,
Would it be possible to review this PR please ?
I published it a week ago and it seems that nobody went through it. There is no posts here while all other PRs, even the latest ones, are commented. The approving workflow has not even been approved. Did you or I miss something ?
Thanks

ecnabogs:
No problem Andy. I understand you are overwhelmed with lots of other stuff. I only wanted to be sure that my PR will be considered.
Thanks and have a good day.

ecnabogs:
I have just rebased my branch on the main with last changes. Waiting for the workflow to be re-rerun.

philwebb:
@ecnabogs Thanks, workflows should run automatically now the first one has been approved.

spring-boot-project/spring-boot/src/test/java/org/springframework/boot/ssl/AliasKeyManagerFactoryTests.java
```diff
@@ -23,32 +23,298 @@
...
+				delegateChosenAlias);
+		AliasKeyManagerFactory factory = createAliasKeyManagerFactory(testAlias, delegate);
+		X509ExtendedKeyManager x509KeyManager = getX509ExtendedKeyManager(factory);
+
+		String chosenAlias = x509KeyManager.chooseEngineServerAlias(testAlgorithm, null, null);
+		assertThat(chosenAlias).isEqualTo(expectedAlias);
+	}
+
+	@Test
+	void chooseEngineServerAliasReturnsDelegateAliasWhenTestAliasIsUnknown() throws Exception {
```
wilkinsona, line 63: I'm not sure that this is the behavior that we want. If the user's configured a particular alias, I would not expect another alias to be used. The same applies to the other `…WhenTestAliasIsUnknown` tests below but I may well be missing something and we don't need to have the same discussion multiple times.

ecnabogs: Since all `chooseXXXAlias(...)` methods in the current Spring Boot implementation, except `chooseEngineServerAlias(...)` one, are delegating their behavior to the internal delegate factory, I was assuming that it was a fallback choice. If we consider that we should return our configured alias (if available and compatible) or nothing, I can change the implementation and the tests. The JDK default implementation returns the first alias whose associated key's algorithm and issuer match the ones requested. Then, the key point now is to choose between the strategy selecting an exclusive configured key alias or the one selecting a preferred one and falling back to another one if unavailable.  I have no preference regarding this choice. Maybe it could also be parameterized, couldn't it

wilkinsona: Thus far, we'd only really considered the server and SSLEngine side of things and it, as you noted above, always returns the user's configured alias.

> Then, the key point now is to choose between the strategy selecting an exclusive configured key alias or the one selecting a preferred one and falling back to another one if unavailable

My preference is for the former. I think it could be considered a regression if we changed the behaviour to return an alias that isn't the user-configured alias.

>  I have no preference regarding this choice. Maybe it could also be parameterized, couldn't it ?

If there's demand for it, that's certainly something that we could consider. I could also imagine some users may want to be able to configure multiple aliases which can then be part of the negotiation to determine the preferred alias. I'd wait for some demand before adding that complexity too, though.

ecnabogs: You are right, I had not considered regression issues. I agree not to return delegate choice in this case. However, in the current implementation, key's algorithm and issuer are not taken into account while selecting the alias. If we keep returning the configured alias unconditionally, it might also be considered as a bug. If the configured key's algorithm and issuer do not match the passed parameters, we should return `null`, right ?

wilkinsona: > If the configured key's algorithm and issuer do not match the passed parameters, we should return null, right?

Yes, I think so. It's return too broadly now and it should be fine to narrow things down. AIUI, with the current implementation we may return an alias that can't actually be used which violates the contract of the `choose…` methods.

ecnabogs: Ok, fine. I will update the code and the tests accordingly.

ecnabogs: Fix done. Awaiting for approval to run the worklow.

wilkinsona:
Thanks for the updates, @ecnabogs.

Thinking about this some more, I'm not sure that applying the alias in the `chooseClientAlias` and `chooseEngineClientAlias` methods is the right thing to do. Previously, the alias only affected the server side. Applying it to the client side as well could be a breaking change, particularly if someone's sharing an SSL bundle between client and server components and they're used to the alias only affecting the server side.

Going back to your original description:

>  If a client has to communicate with multiple partners through mTLS protocol and has a single KeyStore containing its private keys, it cannot select the appropriate key alias to be used with the related partner.

I'm not sure that this is the case. For the client methods, we currently delegate to the `X509ExtendedKeyManager` implementation which will typically be an instance of `sun.security.ssl.SunX509KeyManagerImpl`. It will [return the first matching alias](https://github.com/openjdk/jdk/blob/4c3c2b32a1b9dea6f3e258cf44666932afc179e4/src/java.base/share/classes/sun/security/ssl/SunX509KeyManagerImpl.java#L222).

I think we should take a step back here and broaden what we're looking at. I'd like to see an example of attempting to use mTLS with SSL bundles failing due to the behavior of `AliasX509ExtendedKeyManager`. If that example was in the form of a unit test, that would be ideal as we could then use the test to ensure that whatever fix we end up making isn't regressed in the future.

ecnabogs:
> Thinking about this some more, I'm not sure that applying the alias in the `chooseClientAlias` and `chooseEngineClientAlias` methods is the right thing to do. Previously, the alias only affected the server side. Applying it to the client side as well could be a breaking change, particularly if someone's sharing an SSL bundle between client and server components and they're used to the alias only affecting the server side.

I see your point and I understand you do not want to break backward compatibility. But, by the way, I think the current implementation is partially correct since, for the server-side, the alias is only considered with the method that takes a `SSLEngine` as parameter, not the one that takes a `Socket` instead. The former is more likely to be used but the latter should be considered too for completeness. Then, backward compatibility might be kept by either supporting two aliases, one for the server side and an additional one for the client side. We might also support one or the other with, for instance, a discriminator enum value, specifying which side of the communication channel to represent. The default being for server side.

> I'm not sure that this is the case. For the client methods, we currently delegate to the `X509ExtendedKeyManager` implementation which will typically be an instance of `sun.security.ssl.SunX509KeyManagerImpl`. It will [return the first matching alias](https://github.com/openjdk/jdk/blob/4c3c2b32a1b9dea6f3e258cf44666932afc179e4/src/java.base/share/classes/sun/security/ssl/SunX509KeyManagerImpl.java#L222).

Yes, I know that the current implementation delegates to the JDK default implementation for the client side and that's why I was interested in using another implementation allowing me to take control over the client certificate choice. The thing is that "the first matching" criteria for choosing a client certificate is only based on the public key types and the list of certificate issuer authorities recognized by the peer. However, if I have multiple keys that satisfy these criteria, it is not possible to select which one to use. This scenario might occur while working with distinct partners and/or applications that have the same certificate issuer authority (within the same organization for instance).

> I think we should take a step back here and broaden what we're looking at. I'd like to see an example of attempting to use mTLS with SSL bundles failing due to the behavior of `AliasX509ExtendedKeyManager`. If that example was in the form of a unit test, that would be ideal as we could then use the test to ensure that whatever fix we end up making isn't regressed in the future.

We already have such an example in the unit tests I added. Actually, the method `chooseClientAliasReturnsTestAliasMatchingSupportedDelegateAliasAndAlgorithm` does exactly this. The delegate is asked for the keys with the given algorithms and issuers it can handle. The delegate is then mocked to return a compatible key whose alias does not correspond to the one we expect and finally the unit test checks our `AliasX509ExtendedKeyManager` overrides the delegate choice with our compatible aliased-key.

wilkinsona:
> We already have such an example in the unit tests I added. Actually, the method chooseClientAliasReturnsTestAliasMatchingSupportedDelegateAliasAndAlgorithm does exactly this

There are quite a lot of moving parts here and my concern is that a unit test may make incorrect assumptions about exactly how and when the key manager is being called. I'd like to see a test that tries to make one or more HTTPS requests using mTLS and fails due to the current behavior of Boot's AliasX509ExtendedKeyManager.

ecnabogs:
> There are quite a lot of moving parts here [...]

Not sure what moving parts you refer to.

> I'd like to see a test that tries to make one or more HTTPS requests using mTLS and fails due to the current behavior of Boot's AliasX509ExtendedKeyManager.

It is not an easy task as it would imply a lot of infrastructure code such as creating multiple SSL/TLS certificates, deploying at least two servers with their certificates, configured with their own keystore/truststore. It is really not something that I can easily develop right now. I do not even know how I could integrate and automate such a solution in a GitHub environment.
Do you have any idea how we could do this ? Especially, I think it would be important to be as close as possible as a real use case but working with self-signed certificates on localhost is not very representative. I'm a bit confused how to deal with that.

wilkinsona:
Closing due to lack of updates.

Issue #40174, opened 2024-04-05
Kafka consumer related auto-config takes effect even if we are overdefining it


notusedusername:
## Summary
Over-defining the `ConsumerFactory` instance in our configuration as stated in the `spring-kafka` [docs](https://docs.spring.io/spring-kafka/docs/2.8.11/reference/html/#serdes-mapping-types) won't take effect, but the default instance will be created and used instead. With a minor modification in the bean definition, the intended behavior can be achieved, but it isn't straightforward to do so. 

To work the way the documentation states some `spring-boot-autoconfigure` related changes are required, or if it is not possible, then the `spring-kafka` documentation should be modified accordingly.

## Details
I'm working with Spring Boot-based microservices connected by Kafka. The message values are in JSON format, in the serialized form of a common DTO, let's call it `Document`. 
- the Kafka connection is handled by `spring-kafka`
- there's a `JsonDeserializer` for the `Document`
- we are defining a `ConsumerFactory` and applying our custom deserializer to it

```java
@Bean
public JsonDeserializer<Document> deserializer() {
    return new DocumentMessageDeserializer();
}

@Bean
public ConsumerFactory<String, Document> consumerFactory(
        JsonDeserializer<Document> deserializer,
        KafkaBinderConfigurationProperties kafkaBinderConfigurationProperties) {
    return new DefaultKafkaConsumerFactory<>(
            kafkaBinderConfigurationProperties.mergedConsumerConfiguration(),
            new StringDeserializer(),
            deserializer);
}
```

### Expected behavior
Based on the `spring-kafka` [documentation](https://docs.spring.io/spring-kafka/docs/2.8.11/reference/html/#serdes-mapping-types) we should be able to override the default factory this way with our custom instance with the custom deserializer.

### What happens instead
*The service will get runtime errors about failed deserialization* instead, as it deserializes the message with the default into a `byte[]` and then tries to cast it into a `Document`.
```
java.lang.RuntimeException: java.lang.ClassCastException: class [B cannot be cast to class com.example.dto.Document ([B is in module java.base of loader 'bootstrap'; com.example.dto.Document is in unnamed module of loader org.springframework.boot.loader.LaunchedURLClassLoader @59e84876)
```

On the other hand, if the **property-based config is provided, then the service runs without issues.**
```yaml
spring:
  kafka:
    consumer:
      value-deserializer: com.example.serializers.DocumentMessageDeserializer
```

[...]

Issue #19221, opened 2019-12-04
KafkaAutoConfiguration should back off when a bean of type ConcurrentKafkaListenerContainerFactory is found


manderson23:
We have a Spring Boot app configured with `@EnableKafka` and multiple instances of `ConcurrentKafkaListenerContainerFactory` to consume from different topics that require different deserializers.

With this configuration application startup fails with message 
```
Error creating bean with name 'kafkaListenerContainerFactory' defined in class path resource [org/springframework/boot/autoconfigure/kafka/KafkaAnnotationDrivenConfiguration.class]: Unsatisfied dependency expressed through method 'kafkaListenerContainerFactory' parameter 1; nested exception is org.springframework.beans.factory.NoSuchBeanDefinitionException: No qualifying bean of type 'org.springframework.kafka.core.ConsumerFactory<java.lang.Object, java.lang.Object>' available: expected at least 1 bean which qualifies as autowire candidate. Dependency annotations: {}
```

`KafkaAutoConfiguration` attempts to create a `ConcurrentKafkaListenerContainerFactory ` bean due to the condition `@ConditionalOnMissingBean(name = "kafkaListenerContainerFactory")`. Our configuration does not have a bean with this name hence the error.

Should the condition not check a bean exists with type `ConcurrentKafkaListenerContainerFactory ` rather than a specific bean name? It would then backoff in a scenario like ours where we have multiple beans of this type defined?

wilkinsona:
Thanks for the issue report. Can you please share a minimal example that reproduces the failure? Looking at the auto-configuration and your description, I cannot see why there's no bean of type `ConsumerFactory<Object, Object>`. It should have been created by `KafkaAutoConfiguration`:

https://github.com/spring-projects/spring-boot/blob/2c9cf6511efd8453b39e1c31d66996e8da9b1e33/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/KafkaAutoConfiguration.java#L82-L86

manderson23:
Please see the test case at https://gist.github.com/manderson23/bed77497cd556f24bf77d758ae781455

If fails with the error message I mentioned. If I exclude `KafkaAutoConfiguration` the test runs.

wilkinsona:
[...]

PR #44638, opened 2025-03-07
Make ConcurrentKafkaListenerContainerFactoryConfigurer Generic

lipiridi:
### Summary
This PR updates ConcurrentKafkaListenerContainerFactoryConfigurer to be generic, allowing it to configure any container factory instead of being limited to <Object, Object>.

### Motivation

- Previously, the class was restricted to <Object, Object>, which prevented reuse for other container factories with different key/value types. Making it generic improves flexibility and reusability across various Kafka container configurations.
- This change also enables configuring programmatically created container factories using Kafka listener properties from application.yml, providing greater control and consistency in Kafka consumer setup.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/kafka/ConcurrentKafkaListenerContainerFactoryConfigurer.java
@@ -79,7 +79,7 @@ public class ConcurrentKafkaListenerContainerFactoryConfigurer<K, V> {
 	 * Set the {@link KafkaProperties} to use.
 	 * @param properties the properties
 	 */
-	void setKafkaProperties(KafkaProperties properties) {
+	public void setKafkaProperties(KafkaProperties properties) {
 		this.properties = properties;
 	}
 
@@ -87,7 +87,7 @@ public class ConcurrentKafkaListenerContainerFactoryConfigurer<K, V> {
 	 * Set the {@link BatchMessageConverter} to use.
 	 * @param batchMessageConverter the message converter
 	 */
-	void setBatchMessageConverter(BatchMessageConverter batchMessageConverter) {
+	public void setBatchMessageConverter(BatchMessageConverter batchMessageConverter) {
 		this.batchMessageConverter = batchMessageConverter;
 	}
 
@@ -95,7 +95,7 @@ public class ConcurrentKafkaListenerContainerFactoryConfigurer<K, V> {
 	 * Set the {@link RecordMessageConverter} to use.
 	 * @param recordMessageConverter the message converter
 	 */
-	void setRecordMessageConverter(RecordMessageConverter recordMessageConverter) {
+	public void setRecordMessageConverter(RecordMessageConverter recordMessageConverter) {
 		this.recordMessageConverter = recordMessageConverter;
 	}
 
@@ -103,7 +103,7 @@ public class ConcurrentKafkaListenerContainerFactoryConfigurer<K, V> {
 	 * Set the {@link RecordFilterStrategy} to use to filter incoming records.
 	 * @param recordFilterStrategy the record filter strategy
 	 */
-	void setRecordFilterStrategy(RecordFilterStrategy<? super K, ? super V> recordFilterStrategy) {
+	public void setRecordFilterStrategy(RecordFilterStrategy<? super K, ? super V> recordFilterStrategy) {
 		this.recordFilterStrategy = recordFilterStrategy;
 	}
 
@@ -111,7 +111,7 @@ public class ConcurrentKafkaListenerContainerFactoryConfigurer<K, V> {
 	 * Set the {@link KafkaTemplate} to use to send replies.
 	 * @param replyTemplate the reply template
 	 */
-	void setReplyTemplate(KafkaTemplate<?, ?> replyTemplate) {
+	public void setReplyTemplate(KafkaTemplate<?, ?> replyTemplate) {
 		this.replyTemplate = replyTemplate;
 	}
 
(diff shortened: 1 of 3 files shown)


wilkinsona:
When we're reviewing this we should keep in mind https://github.com/spring-projects/spring-boot/issues/40174 and https://github.com/spring-projects/spring-boot/issues/19221. Both issues are related to Kafka auto-configuration and generics.

snicoll:
Unfortunately, I am not sure that we should be making this change. The configurer is mostly for applying auto-configuration to custom container that are to be exposed for management of the `ApplicationContext`.  I also wonder if the additional complexity is worth it given the primary goal of this class.

This PR doesn't introduce any tests and isn't actionable as such. Before you amend that, I'd like to better understand what you actually mean in the two motivations. Can you provide examples of both and show how the current class is relevant?

lipiridi:
@snicoll thanks for your response!
We are currently creating container factories manually, but we want to configure them using `spring.kafka.listener` properties in a more convenient way, instead of writing boilerplate code.
Another risk is that the listener properties might be extended in the future, and we could miss these changes, failing to update the property injection code accordingly.

Here is the a real example of usage I'm trying to achieve:
```java
@Configuration
@RequiredArgsConstructor
public class KafkaConfiguration {

    private final KafkaProperties kafkaProperties;

    @Bean
    public ConcurrentKafkaListenerContainerFactory<String, GenericRecord> buildFactory() {
        var buildProperties = this.kafkaProperties.buildConsumerProperties(null);

        DefaultKafkaConsumerFactory<String, Object> consumerFactory =
                new DefaultKafkaConsumerFactory<>(buildProperties, new StringDeserializer(), new ErrorHandlingDeserializer<>(new KafkaAvroDeserializer()));

        ConcurrentKafkaListenerContainerFactory<String, GenericRecord> factory =
                new ConcurrentKafkaListenerContainerFactory<>();

        ConcurrentKafkaListenerContainerFactoryConfigurer<String, GenericRecord> configurer = new ConcurrentKafkaListenerContainerFactoryConfigurer();
        configurer.setKafkaProperties(this.kafkaProperties);
        configurer.configure(factory, consumerFactory);

        return factory;
    }
}
```

snicoll:
@lipiridi we won't make the setters public, for sure as this is meant to be injected in user's code and they can't mutate what's been configured for them. Rather, they should apply the configurer and adapt whatever code is necessary.

I had a look to your sample, and we'll take it to the team. This part of the API always was a bit strange to me, perhaps @artembilan has an opinion about this change?

artembilan:
I think this is wrong:
```
 ConcurrentKafkaListenerContainerFactoryConfigurer<String, GenericRecord> configurer = new ConcurrentKafkaListenerContainerFactoryConfigurer<>();
```
There is no need in creating this object manually, rather do this:
```
 @Bean
    public ConcurrentKafkaListenerContainerFactory<String, GenericRecord> buildFactory(ConcurrentKafkaListenerContainerFactoryConfigurer configurer) {
        var buildProperties = this.kafkaProperties.buildConsumerProperties(null);

        DefaultKafkaConsumerFactory<String, Object> consumerFactory =
                new DefaultKafkaConsumerFactory<>(buildProperties, new StringDeserializer(), new ErrorHandlingDeserializer<>(new KafkaAvroDeserializer()));

        ConcurrentKafkaListenerContainerFactory<String, GenericRecord> factory =
                new ConcurrentKafkaListenerContainerFactory<>();

        configurer.configure(factory, consumerFactory);

        return factory;
    }
```

I agree with @snicoll , that those generics for key/value on all the Apache Kafka API is painful and useless.

Tell us, please, why current `ConcurrentKafkaListenerContainerFactoryConfigurer` API is out of use for you?

lipiridi:
@artembilan Hello!
I cannot use it because the `configure` method expects `<Object, Object>` listener and consumer factories, whereas my factory is typified with GenericRecord as the value.
image

artembilan:
Hm. OK. My bad.
So, you have to cast it into expectations of that method.
However, let's see now how those generics on your container factory bean are useful afterwards?
Cannot it just be `<Object, Object>` as well?

lipiridi:
@artembilan the question is how is it supposed to cast the typified `StringDeserializer` to `Object`?
Anyway, I believe erasing types is not the way we want to write code.
image

I'm okay with not making all methods public, especially since I customize the configurer only by setting my own error handler, and the `setCommonErrorHandler` method is already public. I just thought it could be useful for many others.

snicoll:
It's unfortunate that the Kafka API is done this way, but I don't think this is the right place to fix the problem. Introducing the generics would complicate the code for no good reason. 

The current situation is far from ideal but I'd rather open this against the Kafka project.

artembilan:
You can do just like this instead:
```
	@Bean
	public ConcurrentKafkaListenerContainerFactory<Object, Object> buildFactory(
			ConcurrentKafkaListenerContainerFactoryConfigurer configurer) {
		
		var buildProperties = this.kafkaProperties.buildConsumerProperties(null);

		DefaultKafkaConsumerFactory<Object, Object> consumerFactory =
				new DefaultKafkaConsumerFactory<>(buildProperties, null, new JsonDeserializer<>());

		ConcurrentKafkaListenerContainerFactory<Object, Object> factory =
				new ConcurrentKafkaListenerContainerFactory<>();

		configurer.configure(factory, consumerFactory);

		return factory;
	}
```
That `StringDeserializer` is a default value for the `ConsumerConfig.VALUE_DESERIALIZER_CLASS_CONFIG`.

>  I believe erasing types is not the way we want to write code.

I understand your pain and really don't see reason in those generics on the `ListenerContainerFactory` and `ConsumerFactory`.
Same applies for the producer side. 
We really even don't need those generics in Kafka Client itself.
But the cat is out of the bag and we have to live with whatever is there so far.

Perhaps with the current `4.0.0` generation in the work we may reconsider these generics in Spring Kafka and remove them altogether to avoid confusion like this.
It would be huge effort for Spring Kafka itself.
Then this Spring Boot auto-configuration has to be adjusted as well.
And so on for all the projects using Spring Kafka.
I cannot predict all the impact such a breaking change would cause.
Plus right now I don't see an easy migration path from this generics-aware API.

PR #44660, opened 2025-03-11
Make Jetty version configurable via command line for testing new version or snapshot

olamy:
Related to discussion around this https://github.com/jetty/jetty.project/issues/12882

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1014,14 +1014,14 @@ bom {
 			releaseNotes("https://github.com/eclipse-ee4j/jersey/releases/tag/{version}")
 		}
 	}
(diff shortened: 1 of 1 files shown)


olamy:
@wilkinsona ping.

wilkinsona:
As already discussed in the Jetty issue, I don’t think we should do this. Our build doesn’t need it and we could accidentally break it or remove it as it appears to unneeded.

olamy:
Sadly this will not really help the jetty project to test new release with spring projects :(

wilkinsona:
We’re more than happy to help but I think there’s a better approach that will be more robust. Please see my latest comment in the Jetty issue.

olamy:
Sure but if the dsl change again in the future our usage of the init script will fail again without any way for us to notice it.

PR #44677, opened 2025-03-11
Polish OpenTelemetryResourceAttributes

nosan:
Update `OpenTelemetryResourceAttributes` to use `StringUtils.uriDecode(...)`, as it provides the same functionality as the custom `decode(...)` method.

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryResourceAttributes.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryResourceAttributes.java
@@ -16,7 +16,6 @@
 
 package org.springframework.boot.actuate.autoconfigure.opentelemetry;
 
-import java.io.ByteArrayOutputStream;
 import java.nio.charset.StandardCharsets;
 import java.util.Collections;
 import java.util.LinkedHashMap;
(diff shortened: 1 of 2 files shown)


mhalbritter:
Thanks @nosan !

nosan:
@mhalbritter 

I just found this core issue: https://github.com/spring-projects/spring-framework/issues/34570.

It actually does alter a Unicode character.

Consider the following test:

```java
@Test
void otelResourceAttributes() {
    this.environmentVariables.put("OTEL_RESOURCE_ATTRIBUTES", "key8=%20ś\u2019");
    assertThat(getAttributes()).containsEntry("key8", " ś’");
}
```

The custom `decode` method works fine, but `StringUtils.uriDecode` fails with an error:
```

Expecting map:
  {"key8"=" [", "service.name"="unknown_service"}
to contain entries:
  ["key8"=" ś’"]
but the following map entries had different values:
  ["key8"=" [" (expected: " ś’")]
java.lang.AssertionError: 
Expecting map:
  {"key8"=" [", "service.name"="unknown_service"}
to contain entries:
  ["key8"=" ś’"]
but the following map entries had different values:
  ["key8"=" [" (expected: " ś’")]
```

mhalbritter:
Hm. We should then stay on the custom decode method until this is fixed.

mhalbritter:
No problem. I've fixed that in 3bd75f6ce5ab9dd01e7ad2a7b2fe055fd5f404c5 and subscribed to https://github.com/spring-projects/spring-framework/issues/34570. When that's fixed in Framework, we can use their decode method.

PR #44714, opened 2025-03-14
Add support for omitting the SameSite attribute from the session cookie

filiphr:
The main reason for this pull request is to make it easier to disable setting the SameSite cookie in the Spring Session `CookieSerializer`. The default for the Spring Session is Lax. However, this does not work when using SAML with POST redirect, when the cookie is Lax, then the session cookie will not be sent when the redirect happens and thus the authentication will not work (see https://stackoverflow.com/questions/60068271/samesite-attribute-break-saml-flow/60096206#60096206 and https://www.linkedin.com/pulse/samesite-cookie-infinite-redirections-saml-digvijay-singh/).

The current workaround I have is to expose the following bean

```java
@Bean
public DefaultCookieSerializerCustomizer disableSameSite() {
    return cookieSerializer -> cookieSerializer.setSameSite(null);
}
```

i.e. use the serializer to disable the cookie.

If this PR is accepted it would make it easier to configure this without the need to expose a bean by just doing 

```properties
server.servlet.session.cookie.same-site=unset
```

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/Cookie.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/Cookie.java
@@ -146,9 +146,9 @@ public class Cookie {
 	public enum SameSite {
 
 		/**
-		 * Don't set the SameSite cookie attribute.
+		 * The SameSite cookie attribute will be omitted when creating the cookie.
 		 */
-		UNSET("Unset"),
+		OMITTED(null),
 
 		/**
 		 * Cookies are sent in both first-party and cross-origin requests.
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/Cookie.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/Cookie.java
@@ -146,6 +146,11 @@ public class Cookie {
 	public enum SameSite {
 
 		/**
+		 * Don't set the SameSite cookie attribute.
+		 */
+		UNSET("Unset"),
+
+		/**
 		 * Cookies are sent in both first-party and cross-origin requests.
 		 */
 		NONE("None"),
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/session/SessionAutoConfiguration.java
@@ -97,7 +97,9 @@ public class SessionAutoConfiguration {
 			map.from(cookie::getHttpOnly).to(cookieSerializer::setUseHttpOnlyCookie);
 			map.from(cookie::getSecure).to(cookieSerializer::setUseSecureCookie);
 			map.from(cookie::getMaxAge).asInt(Duration::getSeconds).to(cookieSerializer::setCookieMaxAge);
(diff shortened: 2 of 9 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/web/server/Cookie.java
```diff
@@ -145,6 +145,11 @@ public void setPartitioned(Boolean partitioned) {
 	 */
 	public enum SameSite {
 
+		/**
+		 * Don't set the SameSite cookie attribute.
+		 */
+		UNSET("Unset"),
```
wilkinsona, line 151: It may just be me, but `UNSET` makes me have to stop and think to understand what it does. I wonder if `OMITTED`, described as "The SameSite cookie attribute is omitted" would be clearer?

Also, could the attribute value be `null`? I wonder if that would simplify the logic for applying it.

filiphr: I remember now why I picked `UNSET`, it's because the enum `org.apache.tomcat.util.http.SameSiteCookies` has it. In any case `OMITTED` is more clear so I'll go ahead and use that

filiphr:
Thanks for the review @wilkinsona. I've changed the name and tried using `null` for the attribute value.

wilkinsona:
Thanks very much, @filiphr.

vpavic:
@filiphr are you sure it wasn't possible to solve the issue you're having using this combination of properties:

```properties
server.servlet.session.cookie.same-site=none
server.servlet.session.cookie.secure=true
```

I distinctly remember one of the teams in organization I'm working with having what looks like the same issue, and we were able to solve it this way.

It's been a while since I looked at SameSite at depth, but resources like [this one](https://web.dev/articles/samesite-cookies-explained) still mention `Lax` being the default if `SameSite` attribute is omitted. Any behavior different from that might be browser specific and only temporary - from that standpoint I'm not sure if adding `omitted` as an option is a wise thing.

Also see this StackOverflow answer: https://stackoverflow.com/a/72652458/3944191

vpavic:
From what I can see your issue description links to different SO question, but that's not so important as both SO questions we linked mention the same thing as [the article about SameSite](https://web.dev/articles/samesite-cookies-explained) I also linked to in previous comment:

> Cookies with SameSite=None must also specify Secure, meaning they require a secure context.

I personally wouldn't rely on unspecified behavior that might change in future without notice.

filiphr:
>From what I can see your issue description links to different SO question, but that's not so important as both SO questions we linked mention the same thing as [the article about SameSite](https://web.dev/articles/samesite-cookies-explained) I also linked to in previous comment:

You are right, I saw it incorrectly. 

>>Cookies with SameSite=None must also specify Secure, meaning they require a secure context.
>
>I personally wouldn't rely on unspecified behavior that might change in future without notice.

I am not sure that I understand. This is not about using `SameSite=None`.  This PR was created to make sure that we can make sure that no `SameSite` attribute is specified on the cookie.

Reading the article you shared (https://web.dev/articles/samesite-cookies-explained) again, it indeed looks like the combination of 

```
server.servlet.session.cookie.same-site=none
server.servlet.session.cookie.secure=true
```

should be enough. I'll need to test this for our scenario to validate this fully.

vpavic:
> This PR was created to make sure that we can make sure that no `SameSite` attribute is specified on the cookie.

If you read the article I shared, it says that _Cookies without a SameSite attribute are treated as SameSite=Lax_. From your description, sounds like you're relying on browser behavior that's different from that and is what I refer to as _unspecified behavior_ that might change in the future. There's also a sentence about that at the end of SO answer I linked.

Issue #44722, opened 2025-03-14
Rationalize Groovy Template configuration properties


philwebb:
We currently have

```java
@ConfigurationProperties("spring.groovy.template.configuration")
public GroovyMarkupConfigurer
```

There are duplicates and the `.configuration` could be removed.

PR #44733, opened 2025-03-17
Replace GroovyMarkupConfigurer with a properties class

kwondh5217:
gh-44722


wilkinsona:
Thanks for the PR, @kwondh5217. Unfortunately, these aren't the changes that we need as it doesn't address the duplicate properties or deprecate the old ones. I'll take care of things shortly.

PR #44746, opened 2025-03-18
Allow @Name to be used on record components

quaff:
It's not actually necessary, but it's better to align with `@DefaultValue`.

See https://github.com/spring-projects/spring-boot/pull/29010


wilkinsona:
Thanks for the PR, but I don't think we should do this. We made a mistake in #29010 as described in [this comment](https://github.com/spring-projects/spring-boot/pull/29010#issuecomment-1081654428) and aligning `@Name` with `@DefaultValue` would be a step in the wrong direction. I'm not sure we ever actually discussed removing `RECORD_COMPONENT` from `@DefaultValue`. I have labelled it for discussion now.

PR #44753, opened 2025-03-18
Update org.cyclonedx.bom version in docs to latest v2.2.0

abelsromero:
Silly update to docs.
If I understand the code correctly, cyclonedx's maven plugin is managed and updated automatically.
But SB does nothing for Gradle, so without any automatic bump, the docs still mentioned an old version.
I did some local tests and all seems to work fine. The only thing is that the default cyclonedx's version is 1.6 now, instead of 1.5.


abelsromero:
I assume this will need to be backported to other branches, let me know which ones.

wilkinsona:
Arguable, the docs are correct at the moment as they're aligned with [the version of the plugin that we test against](https://github.com/spring-projects/spring-boot/blob/39a18948bd5ece81bb2e6b18b2b321ad56689589/spring-boot-project/spring-boot-parent/build.gradle#L64). We can update the docs for 3.5 but we should update the tested version at the same time. Can you please update this PR to do that?

abelsromero:
Again, it's highly likely I am missing it. But I don't think there are tests for the JAR/WAR integration, there are plenty for the BOM actuator endpoint. But I suspect JAR test should be in [JarIntegrationTests](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/JarIntegrationTests.java) and [WarIntegrationTests](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/WarIntegrationTests.java), and I don't see anything. Plus the [commit](https://github.com/spring-projects/spring-boot/commit/b6c9914c0b7c8eb76500cabd46d9bfbbca140492) to add the feature does not include any.

wilkinsona:
I don't recall the details (perhaps @mhalbritter does), but a key thing is that with these changes we'll be compiling against the version that we're recommending in the docs.

mhalbritter:
We have some tests for the `Packager` itself, which is used by JAR/WAR:

https://github.com/spring-projects/spring-boot/commit/4047c00aa50e2d46156ac756e9ee2edd283e0bdf#diff-6fe88f48f1c151681b13735cb445a199fdf66a81ec8fda42143a49db5c077c9d

But yeah, we don't have any in `JarIntegrationTests` / `WarIntegrationTests`.

abelsromero:
I am not familiar enough with the tests (yet). But it seems something could be improved, let me have a look.

wilkinsona:
Thanks very much, @abelsromero.

PR #44787, opened 2025-03-19
Polish javadoc of SpringProfileAction

rainboyan:


snicoll:
Thanks again @rainboyan. I've fixed the formatting in https://github.com/spring-projects/spring-boot/commit/28f8e3812b94c9867ca6af7ff1cd203dde99bb2a. You can figure this out yourself by building before submitting. In this case, you could run:

```
$ ./gradlew :spring-boot-project:spring-boot:check
```

This would have failed telling you the formatting is wrong and to run `format`. You can do as as follows:

```
$ ./gradlew :spring-boot-project:spring-boot:format
```

rainboyan:
@snicoll Thank you for your suggestion, I will do it next time.

Issue #44793, opened 2025-03-20
Allow overriding JobParametersConverter for spring batch auto-configuration


pPanda-beta:
Since https://github.com/spring-projects/spring-batch/issues/4650#issuecomment-2405684917 is closed and there is an way to override the `JobParametersConverter` in `org.springframework.batch.core.configuration.support.DefaultBatchConfiguration` it would be very convenient if spring boot auto configuration also allows this.
Right now

`class BatchAutoConfiguration.SpringBootBatchConfiguration` does not override 

`protected JobParametersConverter getJobParametersConverter()`

https://github.com/spring-projects/spring-boot/blob/4a2c26eeec751b409f614b20e374e0f21bc0f595/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java#L99

And the only alternative workaround is to utilize the `BatchConversionServiceCustomizer` and install converters per type, while most parameter objects are data classes, and can easily be handled by json serialization. 
Explained in details [here](https://github.com/spring-projects/spring-batch/issues/4650#issuecomment-2737879843)

cc: @fmbenhassine

quaff:
What a coincidence! I just created PR GH-44799.

wilkinsona:
Closing in favor of #44799. Thanks for the PR, @quaff.

PR #44799, opened 2025-03-20
Auto-configure Spring Batch to use a JobsParametersConverter bean when available

quaff:
Fix GH-44793

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
@@ -22,6 +22,7 @@ import javax.sql.DataSource;
 
 import org.springframework.batch.core.configuration.annotation.EnableBatchProcessing;
 import org.springframework.batch.core.configuration.support.DefaultBatchConfiguration;
+import org.springframework.batch.core.converter.JobParametersConverter;
 import org.springframework.batch.core.explore.JobExplorer;
 import org.springframework.batch.core.launch.JobLauncher;
 import org.springframework.batch.core.repository.ExecutionContextSerializer;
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks, @quaff. I think it would be better to handle `jobParametersConverter` separately, particularly as it isn't configured using a property. Could you pull that out into a separate PR please?

quaff:
> Thanks, @quaff. I think it would be better to handle `jobParametersConverter` separately, particularly as it isn't configured using a property. Could you pull that out into a separate PR please?

Sure.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchProperties.java
```diff
@@ -83,6 +89,12 @@ public static class Jdbc {
 		 */
 		private String platform;
 
+		/**
+		 * The length of long string columns in database. Do not override this if you
+		 * haven't modified the schema.
+		 */
+		private Integer maxVarCharLength;
+
```
wilkinsona, line 97: I don't think we should offer this as a property for reasons similar tohttps://github.com/spring-projects/spring-boot/issues/44802#issuecomment-2739655041. We can reconsider should Batch ever provide first-class support for modifying its schema scripts.

wilkinsona (changes requested):
I'm not sure about the max varchar length either given how tightly coupled it is to the schema. Being able to configure `validateTransactionState` looks useful though. Can we just focus on that please in this PR?

wilkinsona:
Thanks again, @quaff.

pPanda-beta:
Very silly question @wilkinsona is this PR merged via a separate commit or rejected?

https://github.com/spring-projects/spring-boot/commit/4ff00612b143a8026e4df5897cd49216ed9d0779 seems like an identical commit

wilkinsona:
It was merged in https://github.com/spring-projects/spring-boot/commit/4ff00612b143a8026e4df5897cd49216ed9d0779. Unfortunately, limitations of GitHub's web UI cause it to appear unmerged.

PR #44803, opened 2025-03-20
Provide a configuration property for Spring Batch's validateTransactionState

quaff:


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchProperties.java
```diff
@@ -67,6 +68,11 @@ public static class Jdbc {
 		private static final String DEFAULT_SCHEMA_LOCATION = "classpath:org/springframework/"
 				+ "batch/core/schema-@@platform@@.sql";
 
+		/**
+		 * Whether to validate the transaction state.
+		 */
+		private Boolean validateTransactionState;
```
wilkinsona, line 74: This could be a `boolean` with a default value of `true`.

wilkinsona:
Thanks, @quaff.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchProperties.java
quaff: I think it should align with `tablePrefix` since it's not default to `BATCH_`, Spring Boot doesn't need to change if Spring Batch change its defaults, but many other AutoConfigurations doesn't work in such way.

wilkinsona: Our approach has changed over time. These days, we prefer to provide a default if we can as it's better for users when looking at a property and its documentation, particularly in an IDE that supports that through the metadata. It's very unlikely that Batch will change its default. If it does, there's now a test in place that will fail.

quaff: > we prefer to provide a default if we can as it's better for users when looking at a property and its documentation, particularly in an IDE that supports that through the metadata.

Then `tablePrefix` and `isolationLevelForCreate` should have default value.

Issue #44864, opened 2025-03-24
Drop /reference from docs build YAML


philwebb:
[This line](https://github.com/spring-projects/spring-boot/blob/5cb70372b8feba68780340afb098ba075891abf4/.github/workflows/deploy-docs.yml#L54) can be simplified now that https://github.com/spring-io/spring-doc-actions/issues/20 is fixed.

philwebb:
Closed in 8a533d7f940c1bd1266750b13689a2611665b9e6

PR #44865, opened 2025-03-24
drop /reference from httpdocs-path

reduan96:


wilkinsona:
Thanks, @reduan96, but #44864 is already assigned to someone.

PR #44866, opened 2025-03-25
Clarify the use of multiple profile expressions with "spring.config.activate.on-profile"

quaff:
See https://github.com/spring-cloud/spring-cloud-commons/issues/1416#issuecomment-2749903603

### DIFF
--- a/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
+++ b/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
@@ -573,7 +573,9 @@ The following activation properties are available:
 | Property | Note
 
 | `on-profile`
-| A profile expression that must match for the document to be active.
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the suggestion but I think this may cause confusion. `,` isn't really the equivalent of `|`. It's a separator for the expressions when configured as a single comma-separate value due to it being bound to a `String[]`. If it's configured as a list, no separation occurs. In other words, the following two snippets of YAML are not equivalent:

```yaml
spring.config.activate.on-profile:
    - a|b
```

```yaml
spring.config.activate.on-profile:
    - a,b
```

I think it would be better to update the table in [this section of the documentation](https://docs.spring.io/spring-boot/3.3/reference/features/external-config.html#features.external-config.files.activation-properties) to describe `on-profile` more accurately. Perhaps something like "profile expressions of which at least one must match for the document to be active".

quaff:
Updated, could someone review the wording?

PR #44901, opened 2025-03-26
Add spring.kafka.consumer.max-poll-interval configuration property 

quaff:
I think `max.poll.interval.ms` is more important than `max.poll.records` which has its own configuration property key.

Feel free to close if the team disagree.


wilkinsona:
Thanks, @quaff. Kafka defines both `max.poll.records` and `max.poll.interval.ms` as medium importance configuration options. Given that we have a property for one of them, I see no reason not to add the other.

mhalbritter:
Thanks @quaff !

PR #44917, opened 2025-03-27
Remove redundant @Order and add corresponding ArchRule

quaff:


wilkinsona:
Thanks for the PR but I prefer the code without these changes. Using `@Order` in this way makes it clear that the order is important and that it has been thought about.

PR #44927, opened 2025-03-28
Update gradle description in SendGrid module

anbusampath:
Change description from "Spring Boot New Project" to "Spring Boot SendGrid"


wilkinsona:
Thanks for spotting this, @anbusampath. Unfortunately, we can't accept contributions for the 4.0.x branch as we need to be able to rewrite history at this stage to keep it clean. That would mean that you'd lose credit for the change.

Issue #44929, opened 2025-03-28
Spring Data Cassandra tests are using the wrong property for customizing the repository type


snicoll:
It's using `spring.cassandra.repositories.type` instead of `spring.data.cassandra.repositories.type`.

krutip7:
Hi @snicoll, I'd like to pick this.

snicoll:
@krutip7 thanks but that's not just about renaming the property, we need also to make sure it's testing the right thing. Given that the property was wrong, so are the tests.

wilkinsona:
Closing in favor of #44942.

PR #44942, opened 2025-03-28
Spring Data Cassandra tests are using the wrong property for customizing the repository type

nosan:
The tests did not verify anything because `CityCassandraRepository` is located outside the `City` package, which includes the `CityRepository` and `ReactiveCityRepository` repositories only, and also because of an incorrect property name.

See https://github.com/spring-projects/spring-boot/issues/44929


wilkinsona:
Thanks very much, @nosan.

PR #44944, opened 2025-03-29
Add more tests to OpenTelemetryResourceAttributesTests

nosan:
Add more tests for decoding percent-encoded strings with non-ASCII characters and multi-byte sequences.


mhalbritter:
Thanks @nosan !

PR #44947, opened 2025-03-29
Replace mentions of deprecated MockBean annotation

aahlenst:
A few places were missed when deprecating MockBean.


mhalbritter:
Thanks @aahlenst !

PR #44952, opened 2025-03-30
Enable RestClientAutoConfiguration in reactive web applications when virtual threads are enabled

### LINKED ISSUE #44912: Enable RestClientAutoConfiguration in reactive web applications when virtual threads are enabled
Currently `RestClientAutoConfiguration` uses `NotReactiveWebApplicationCondition` so that it is only enabled for non-reactive web applications which helps prevent users from accidentally blocking the event loop thread.

This foot-gun protection is less useful when virtual threads are enabled, as we configure WebFlux with an async task executor which allows blocking calls to be made for any controller method that does not return reactive types. Since `RestClientAutoConfiguration` provides a lot of helpful functionality, we should relax the condition to a `NotReactiveWebApplicationOrVirtualThreadsEnabledCondition`

dmitrysulman:
This PR introduces `NotReactiveWebApplicationOrVirtualThreadsEnabledCondition` for use in `RestClientAutoConfiguration`.

Closes #44912

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/client/NotReactiveWebApplicationOrVirtualThreadsEnabledCondition.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/client/NotReactiveWebApplicationOrVirtualThreadsEnabledCondition.java
@@ -1,0 +1,47 @@
+/*
+ * Copyright 2012-2025 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      https://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.boot.autoconfigure.web.client;
+
+import org.springframework.boot.autoconfigure.condition.AnyNestedCondition;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnThreading;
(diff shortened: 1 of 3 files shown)


philwebb:
Thanks @dmitrysulman! I've merged this with one small amendment to try and also check that there's an `applicationTaskExecutor` bean. This should hopefully align with [this code in `WebFluxAutoConfiguration`](https://github.com/spring-projects/spring-boot/blob/f77f3c68b7b202d1a2e7e22bf2d0bbe48ae172b1/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/reactive/WebFluxAutoConfiguration.java#L197-L206)

blake-bauman:
I'm curious as to why only when Virtual Threads are enabled.  WebFlux can be configured with a `ThreadPoolTaskExecutor` for blocking execution using `BlockingExecutionConfigurer.setExecutor()`, which should be sufficient for a RestClient.

philwebb:
@blake-bauman We I looked at the PR I couldn't see an obvious way to tell if the `BlockingExecutionConfigurer` has been called. Looking again, we might be able to do that if `WebFluxConfigurationSupport.getBlockingExecutionConfigurer()` wasn't final. We could then return our own subclass and get access to `BlockingExecutionConfigurer.getExecutor()`. I'll open a Spring Framework issue to see if that's an option.

blake-bauman:
I think my main question is, what is the need to detect?  Even if an app has Virtual Threads enabled, I don't believe the execution is offloaded to a Virtual Thread if the Controller method return type is `Mono`, and they try to use a RestClient.  I believe that throws the same exception if they tried to use a RestClient on a non-Mono Controller method and they haven't configured the `BlockingExecutionConfigurer`.  So, I'm mainly wondering if there's a benefit to trying to detect, or if we should just let them use it and it'll fail later without the proper setup.

(For reference, this is for apps which, for one reason or another, can't yet update to Java 21 and therefore can't enable Virtual Threads)

philwebb:
@blake-bauman I'll need to discuss this with the team again to see and with some folks that have more knowledge of the reactive stack. I've flagged this one for our next meeting.

blake-bauman:
For now, we've just copy/pasted the entire `RestClientAutoConfiguration` internally without the conditions in order to unblock, so that we can get the Micrometer enrichments for RestClients.

PR #44962, opened 2025-04-01
Rename 'spring.http.client' properties to 'spring.http.clients'

### LINKED ISSUE #44958: Rename 'spring.http.client' properties to 'spring.http.client.settings'
Much like we have `properties.actuator.management.endpoints` for properties that change all endpoints, we should have probably really named `spring.http.client` as `spring.http.clients`.

EDIT: We've reconsidered and we'd like to use `spring.http.client.settings`

shivammm21:
This PR renames the properties from `spring.http.client` to `spring.http.clients` to follow the same pattern used elsewhere in Spring Boot (like `properties.actuator.management.endpoints`).

The change affects
- The main `HttpClientProperties` class configuration property
- All test files using these properties
- Documentation in rest-client.adoc

This is a breaking change for anyone already using the `spring.http.client.*` properties in the 3.4.0 milestone.

Fixes #44958


bclozel:
Thanks for this PR but the issue is already assigned so I assume that Phil has already started working on this. Also, renaming the property is not enough as additional metadata must be added to help IDEs and the properties migrator with the renaming.

PR #44964, opened 2025-04-01
Update 'for removal' versions > 3.5.0 to 4.0.0

### LINKED ISSUE #44959: Update "for removal" versions > 3.5.0 to 4.0.0
Since 3.5.x is the last 3.x generation we should update all our deprecation removal versions

shivammm21:
Fixes #44959
   
Since 3.5.x is the last 3x generation, this PR updates all our deprecation removal versions from >3.5.0 (like 3.6.0, 3.7.0, etc.) to 4.0.0. This ensures a consistent deprecation timeline with no scheduled removals in versions that won't exist.
   
The updates are only to JavaDoc comments, not to the actual code or @Deprecated annotations, which don't contain version-specific removal information.


wilkinsona:
Thanks for the PR, @shivammm21, but once again this issue is already assigned to someone else so we won't accept a contribution. To avoid wasted effort, I'd recommend commenting on an unassigned issue to indicate that you'd like to contribute. If it's an issue that's suitable for contribution, the issue can then be assigned to you and you can start working on it.

Issue #43618, opened 2024-12-26
AutoConfiguredRestClientSsl overwrites configuration from HttpClientProperties


kzander91:
This is similar to #36594 and #27360 but I'm not sure if a solution for those issues would also fix mine, so feel free to close this if you consider this to be a duplicate.

Issue: Configuration made through the `spring.http.client` properties is ignored if I use `AutoConfiguredRestClientSsl`. For example, configuration like
```yaml
spring:
  http:
    client:
      redirects: dont_follow
```
does not work if I also have this:
```java
@Bean
RestClientCustomizer clientCertConfigurer(RestClientSsl restClientSsl) {
    return builder -> builder.apply(restClientSsl.fromBundle("my-bundle"));
}
```

The reason for this is that `AutoConfiguredRestClientSsl` _replaces_ the request factory and thus discards any configuration made via the `spring.http.client` properties:
https://github.com/spring-projects/spring-boot/blob/183285258d7b5092224eab1600a2cfa8332682d9/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/client/AutoConfiguredRestClientSsl.java#L52-L56

For _this_ specific issue I can work around this by removing my `RestClientCustomizer` and use
```yaml
spring:
  http:
    client:
      redirects: dont_follow
      ssl:
        bundle: "my-bundle"
```
instead, but this only works in this static scenario, if I needed some more sophisticated logic to determine the SSL bundle to use at runtime, this wouldn't work.

wilkinsona:
I think this is working as documented:

https://github.com/spring-projects/spring-boot/blob/15b63faec51c52a306a074724781bc43137ae1ef/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/client/RestClientSsl.java#L38-L42

kzander91:
This note and also the corresponding sentences in the [reference docs](https://docs.spring.io/spring-boot/reference/io/rest-client.html#io.rest-client.restclient.ssl) don't really tell me that this affects the `spring.http.client` properties. Without studying the corresponding auto configuration code, how am I supposed to know that these properties work on their own instance of a `ClientHttpRequestFactory` that may then later be replaced if other configuration beans are defined?
Other features provided by Boot are additive, where I can use a mix of properties and customizer beans, but here for HTTP clients, I can't.

[...]

PR #44979, opened 2025-04-02
AutoConfiguredRestClientSsl should build upon, rather than overwrite, the configuration from HttpClientProperties

nosan:
See gh-43618


philwebb:
Thanks very much @nosan!

PR #44982, opened 2025-04-02
Polish RestClientAutoConfiguration and RestClientBuilderConfigurer

nosan:


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/client/RestClientBuilderConfigurer.java
```diff
@@ -32,21 +32,16 @@
...
+	private final ClientHttpRequestFactoryBuilder<?> requestFactoryBuilder;
 
-	private ClientHttpRequestFactorySettings requestFactorySettings;
+	private final ClientHttpRequestFactorySettings requestFactorySettings;
 
-	private List<RestClientCustomizer> customizers;
+	private final List<RestClientCustomizer> customizers;
 
-	void setRequestFactoryBuilder(ClientHttpRequestFactoryBuilder<?> requestFactoryBuilder) {
+	RestClientBuilderConfigurer(ClientHttpRequestFactoryBuilder<?> requestFactoryBuilder,
```
mhalbritter, line 41: This is an API breaking change. Before that, there has been an default public constructor. Now there's only a package-protected one.

nosan: Fixed it, apologies for completely overlooking this! 🤦

I'm not sure if anyone directly creates a `RestClientBuilderConfigurer` instance, as it wouldn't make sense given that all setters are package-protected. However, you're correct, this is indeed an API breaking change. My mistake!

mhalbritter: No need to apologize, that was a subtle one! :)

mhalbritter:
Thanks @nosan !

PR #45008, opened 2025-04-04
Lourw/perf logging

lourw:


lourw:
So sorry can we close this, was supposed to PR this onto a fork

PR #45011, opened 2025-04-05
TypeUtils does not handle generics with identical names in different positions

nosan:
This commit improves the `TypeUtils` class to handle generic parameters that share similar names but are positioned differently.


philwebb:
Thanks @nosan !

PR #45013, opened 2025-04-06
Validate that sslInfo is not null in SslHealthIndicator constructor

geniusYoo:
The SslHealthIndicator constructor previously did not validate if the provided SslInfo instance was null. This could potentially lead to a NullPointerException later when the doHealthCheck method is invoked if the SslInfo bean was not properly configured or available.

This commit adds an Assert.notNull check for the sslInfo parameter to ensure fail-fast behavior during application startup or bean creation. This aligns with the common practice in other Spring Boot components and health indicators where essential dependencies are validated in the constructor.

### DIFF
--- a/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/ssl/SslHealthIndicator.java
+++ b/spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/ssl/SslHealthIndicator.java
@@ -35,7 +35,7 @@ import org.springframework.util.Assert;
  * {@link Status#OUT_OF_SERVICE} when a certificate is invalid.
  *
  * @author Jonatan Ivanov
- * @author geniuus
+ * @author Young Jae You
  * @since 3.4.0
  */
 public class SslHealthIndicator extends AbstractHealthIndicator {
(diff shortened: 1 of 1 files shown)


spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/ssl/SslHealthIndicator.java
```diff
@@ -34,13 +34,16 @@
  * {@link Status#OUT_OF_SERVICE} when a certificate is invalid.
  *
  * @author Jonatan Ivanov
+ * @author geniuus
```
mhalbritter, line 37: This author isn't valid. It has to be "<Name> <Surname>".

mhalbritter (changes requested):

spring-boot-project/spring-boot-actuator/src/main/java/org/springframework/boot/actuate/ssl/SslHealthIndicator.java
geniusYoo: Hi @mhalbritter,

Thanks for the feedback! I've updated the `@author` tag to use my full name (`Young Jae You`) as you requested.

I also added the missing import statement for `Assert` that was causing the build failure in the previous commit.

Could you please take another look? Let me know if further changes are needed. Thanks!

mhalbritter:
Thanks @geniusYoo !

PR #45018, opened 2025-04-07
MongoDB's dependency management is missing Kotlin coroutine driver modules

nenros:
Currently to keep same version of coroutine driver we need to make nasty trick

```kotlin 
val mongodbVersion: String? = dependencyManagement.importedProperties["mongodb.version"]

dependencies {
    implementation("org.mongodb:mongodb-driver-kotlin-coroutine:${mongodbVersion}")
    implementation("org.mongodb:bson-kotlin:${mongodbVersion}")
}
```

It would be much easier to keep that version in spring bom.

### DIFF
--- a/spring-boot-project/spring-boot-dependencies/build.gradle
+++ b/spring-boot-project/spring-boot-dependencies/build.gradle
@@ -1570,10 +1570,12 @@ bom {
 			modules = [
 				"bson",
 				"bson-record-codec",
+				"bson-kotlin",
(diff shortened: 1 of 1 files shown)


snicoll:
@nenros thanks very much for making your first contribution to Spring Boot. FTR we order dependencies alphabetically for easier management, I've polished that in https://github.com/spring-projects/spring-boot/commit/6d3686451b561a63d8d08d31773b085006896142

Issue #44990, opened 2025-04-03
Show the use of token properties in authorization server clients configuration example


OrangeDog:
Spring Authorization Server allows defining per-client token settings.
https://github.com/spring-projects/spring-authorization-server/issues/1385#issuecomment-1772439977

However, Boot's `OAuth2AuthorizationServerProperties` does not expose this. I'd like to see something like this:

```yaml
spring.security.oauth2.authorizationserver.client:
  my-client:
    client-id: 41fd9212-dc6f-4ced-a7a8-e6431a3f49da
    client-secret: '{noop}secret'
    tokens:
      authorization-code-time-to-live: 5m
      access-token-time-to-live: 10m
      access-token-format: reference
      reuse-refresh-tokens: false
      refresh-token-time-to-live: 28d
```

Possibly an ability to set a default for all clients too.

wilkinsona:
As far as I can tell, we already have the per-client properties:

https://github.com/spring-projects/spring-boot/blob/05a9521fde5af5675a798bd3f3fc84253907d238/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/security/oauth2/server/servlet/OAuth2AuthorizationServerProperties.java#L459-L488

The metadata and documentation for them is lacking at the moment. https://github.com/spring-projects/spring-boot/issues/9945 will help with that in time.

OrangeDog:
Ah I see, that's good news.
For some reason, my IDE (IDEA Ultimate) knows about the other client properties, but not `token`.

wilkinsona:
I would report that to Jetbrains. As I understand it, they do not use the metadata but instead introspect the types to which the properties are bound. As such, I'd expect IDEA ultimate to be able to offer auto-complete for the token properties.

wilkinsona:
We can improve the documentation by adding the token properties to the example in [this existing section](https://docs.spring.io/spring-boot/reference/web/spring-security.html#web.security.oauth2.authorization-server).

wilkinsona:
#45019 has addressed this but only on `main`. We can use this issue to back port the changes to `3.3.x` and `3.4.x`.

PR #45019, opened 2025-04-07
Show the use of token properties in authorization server clients configuration example

ngocnhan-tran1996:
Issue gh-44990


philwebb:
Thanks very much @ngocnhan-tran1996 !

PR #45020, opened 2025-04-07
Polish AutoConfigurationMetadataLoader

quaff:
Align to other final classes, the protected modifier is meaningless.


mhalbritter:
Thanks @quaff !

Issue #44853, opened 2025-03-24
Revisit excludes on javax.xml.bind:jaxb-api


snicoll:
There are a number of excludes that look outdated, in particular for jOOQ.

DimaVilda:
hi there!
yeah, jooq has jakarta dependency indeed so I can simplify and remove excludes and check other cases as well. Might be a good first task in this project for me btw. Can I be assigned to it @wilkinsona ?

snicoll:
@DimaVilda thanks for the offer. Yes, please go ahead and let us know how it goes.

snicoll:
@DimaVilda how is it going?

DimaVilda:
> [@DimaVilda](https://github.com/DimaVilda) how is it going?

hi @snicoll ! Was busy on my curr project with releases and all this things around this whole week. I'll start do it this weekend/Monday max.
For the future to avoid misunderstandings: do you have some ETA for issues when smbd is assigned? Ofc it depends on your release date, task severity and other factor but would be cool to know some times frames.

snicoll:
> do you have some ETA for issues when smbd is assigned? 

Not really. When someone asks to be assigned, I kind of expect they're asking because they can work on it. If you don't have time, I'd recommend not doing that. It really depends on the issue but for small things like this we prefer to get them out of the way than having the issue opened for a while.

DimaVilda:
> > do you have some ETA for issues when smbd is assigned?
> 
> Not really. When someone asks to be assigned, I kind of expect they're asking because they can work on it. If you don't have time, I'd recommend not doing that. It really depends on the issue but for small things like this we prefer to get them out of the way than having the issue opened for a while.

makes sense, ok, as I told I can do it this weekend or Monday max so keep in touch

snicoll:
Closing in favor of PR.

PR #45025, opened 2025-04-07
#44853 - revisit excludes on javax.xml.bind:jaxb-api

DimaVilda:
Reviewed and removed (in case needed) outdated jaxb-api 'excludes' from dependencies.
Note that for "org.liquibase:liquibase-core" we still depend on jaxb-api so I left it untouched!


snicoll:
@DimaVilda thanks for this. You need to sign-off your commit for DCO to pass. You can do so by amending your commit to include a [Sign-off clause](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for).

DimaVilda:
> @DimaVilda thanks for this. You need to sign-off your commit for DCO to pass. You can do so by amending your commit to include a [Sign-off clause](https://stackoverflow.com/questions/1962094/what-is-the-sign-off-feature-in-git-for).

I see, should be updated now.

philwebb:
Thanks very much @DimaVilda!

PR #45032, opened 2025-04-08
Escape the asterisk in spring-application.adoc

nevenc:
Asterisk needs to be escaped not to be interpreted as bold formatting.
Only the first asterisk needs to be escaped, second one is not interpreted as bold formatting.

Here's the error:
image

Notice all the bold text in the second paragraph, and missing '*' asterisk in the name of the file, e.g. `application.*`

Here is how the error shows in adoc:
image

Here's how it looks fixed in adoc:
image


mhalbritter:
Thanks @nevenc !

PR #45156, opened 2025-04-11
Polish ternary operator

quaff:


philwebb:
Thanks, but we generally negate the test in ternary expressions.

philwebb:
It turns out I didn't quite get the reason we prefer the negation first. After some discussion with @snicoll I've seen the light! We're really trying to enforce `(predicate) ? main : fallback` and that's why we mostly end up using `!= null`. I'll merge this one after all.

PR #45161, opened 2025-04-11
Add more tests for BootstrapExecutorConfiguration

nosan:
Add tests for manually registered singleton and when the `boostrapExecutor` alias is defined.


mhalbritter:
Thanks @nosan !

Issue #45150, opened 2025-04-10
Migrate from AntPathRequestMatcher to PathPatternRequestMatcher


philwebb:
The following classes currently use the deprecated `AntPathRequestMatcher` and should be migrated to `PathPatternRequestMatcher`:

* `org.springframework.boot.devtools.autoconfigure.RemoteDevtoolsSecurityConfiguration`
* `org.springframework.boot.autoconfigure.security.servlet.StaticResourceRequest`
* `org.springframework.boot.autoconfigure.security.servlet.PathRequest`
* `org.springframework.boot.actuate.autoconfigure.security.servlet.AntPathRequestMatcherProvider`
* `org.springframework.boot.actuate.autoconfigure.security.servlet.EndpointRequest`
* `org.springframework.boot.actuate.autoconfigure.cloudfoundry.servlet.CloudFoundryActuatorAutoConfiguration`
* `smoketest.actuator.customsecurity.SecurityConfiguration`
* `org.springframework.boot.actuate.autoconfigure.security.servlet.ManagementWebSecurityAutoConfigurationTests`

philwebb:
Closing in favor of PR #45163. Thanks @ngocnhan-tran1996!

PR #45163, opened 2025-04-11
Migrate from AntPathRequestMatcher to PathPatternRequestMatcher

ngocnhan-tran1996:
Issue: gh-45150


philwebb:
Thanks very much @ngocnhan-tran1996!

PR #45165, opened 2025-04-11
Inconsistency in code style and make source code easier to understand

iggzq:
In the `ApplicationProperties.java`, the `lazyInitialization` parameter explicitly declares its default value as `false`. However, the `keepAlive`, `allowCircularReferences`, and `allowBeanDefinitionOverriding` parameters do not have their default values of `false` explicitly declared, leading to an inconsistency in style. 

Before `Spring Boot 3.4`, a similar inconsistency existed, where in the `SpringApplication.java` file, `lazyInitialization` was explicitly set to default to false, whereas `keepAlive`, `allowCircularReferences`, and `allowBeanDefinitionOverriding` did not have their default values of `false` explicitly stated. 

In my opinion, explicitly declaring the default values for boolean objects makes them easier to understand.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ApplicationProperties.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/ApplicationProperties.java
@@ -37,13 +37,13 @@ class ApplicationProperties {
 	 * Whether bean definition overriding, by registering a definition with the same name
 	 * as an existing definition, is allowed.
 	 */
-	private boolean allowBeanDefinitionOverriding;
+	private boolean allowBeanDefinitionOverriding = false;
 
 	/**
 	 * Whether to allow circular references between beans and automatically try to resolve
 	 * them.
 	 */
-	private boolean allowCircularReferences;
+	private boolean allowCircularReferences = false;
 
 	/**
 	 * Mode used to display the banner when the application runs.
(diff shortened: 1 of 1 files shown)


nosan:
There are many similar inconsistencies throughout the codebase due to numerous people working on it. In my opinion, minor polishing changes like these don't significantly improve overall consistency—they only affect `ApplicationProperties.java`, while other classes remain untouched and likely contain the same or similar issues.

I am wondering if this could be addressed by using a Checkstyle rule such as [ExplicitInitialization](https://checkstyle.sourceforge.io/checks/coding/explicitinitialization.html) to enforce avoiding explicit initialization of primitives with their default values.

```xml
<module name="ExplicitInitialization">
	<property name="onlyObjectReferences" value="false"/>
	<property name="tokens" value="VARIABLE_DEF"/>
</module>
```

philwebb:
I'm not totally sure about a blanket rule, we do tend to set values in `@ConfigurationProperties` explicitly but generally not anywhere else. For now I've done a quick sweep of the code and removed the booleans that I think make sense. See 2143d702f21ef814d07b8f910cb452383a2fd64b

philwebb:
I've opened https://github.com/spring-projects/spring-boot/issues/45168 for the checkstyle rule. Thanks for raising the PR @iggzq, but since we want to go the other way I'm going to decline this one in favor of commit  2143d702f21ef814d07b8f910cb452383a2fd64b

PR #45206, opened 2025-04-16
Add config-view actuator endpoint to expose runtime config with sources

siripuramGanesh:
This PR adds a new Actuator endpoint `/actuator/config-view` that exposes a runtime view of all configuration properties, their current values (with sensitive data masked), and their sources.

Features:
- Lists config properties from all EnumerablePropertySources
- Shows property name, value, and source
- Masks sensitive values (e.g., passwords, tokens)
- Provides diagnostics for debugging config issues

Example output:
[
  { "key": "server.port", "value": "8080", "source": "application.properties" },
  { "key": "spring.datasource.password", "value": "****", "source": "env" }
]

Use case: Developers and operators can quickly debug which config values are in effect, especially when using layered configs like profiles, env vars, and config servers.


wilkinsona:
Thanks for the proposal, but this functionality is already provided by the existing [`env` endpoint](https://docs.spring.io/spring-boot/api/rest/actuator/env.html).

Issue #42782, opened 2024-10-18
Improve MessageSourceMessageInterpolator when using MessageSource.setUseCodeAsDefaultMessage(true) and Bean Validation attributes


nosan:
When I was doing research for #42773, I came across two potential improvements for
`MessageSourceMessageInterpolator`:

- The logic around `MessageSource.setUseCodeAsDefaultMessage(true)` #28930 could be
  improved a little bit:

```java

//Current test in MessageSourceMessageInterpolatorTests

@Test
void interpolateWhenParametersAreUnknownUsingCodeAsDefaultShouldLeaveThemUnchanged() {
	this.messageSource.setUseCodeAsDefaultMessage(true);
	this.messageSource.addMessage("top", Locale.getDefault(), "{child}+{child}");
	assertThat(this.interpolator.interpolate("{foo}{top}{bar}", this.context))
			.isEqualTo("{foo}{child}+{child}{bar}");
}

// Changed a little bit
@Test
void interpolateWhenParametersAreUnknownUsingCodeAsDefaultShouldLeaveThemUnchanged() {
	this.messageSource.setUseCodeAsDefaultMessage(true);
	this.messageSource.addMessage("top", Locale.getDefault(), "{child}+{child}");
	this.messageSource.addMessage("foo", Locale.getDefault(), "foo");
	this.messageSource.addMessage("bar", Locale.getDefault(), "bar");
	// Actually, I expected the result to be "foo{child}+{child}bar" 
	// since I explicitly specified these values in the MessageSource.
	// However, the test fails! 
	assertThat(this.interpolator.interpolate("{foo}{top}{bar}", this.context))
			.isEqualTo("foo{child}+{child}bar");
}
```

If `MessageSource` contains the searched parameter, it should be used, even if the value is identical to the parameter itself.

The solution for this is quite straightforward

```java 
private static final String DEFAULT_MESSAGE = MessageSourceMessageInterpolator.class.getName();

private String replaceParameter(String parameter, Locale locale, Set<String> visitedParameters) {
	parameter = replaceParameters(parameter, locale, visitedParameters);
	String value = this.messageSource.getMessage(parameter, null, DEFAULT_MESSAGE, locale);
	if (value == null || value.equals(DEFAULT_MESSAGE)) {
		return null;
	}
	return replaceParameters(value, locale, visitedParameters);
}

```

- The message lookup for EL expressions like `${validatedValue}` and for Bean Validation
[...]

PR #45212, opened 2025-04-16
MessageSourceMessageInterpolator does not replace a parameter when the message matches its code

nosan:
See https://github.com/spring-projects/spring-boot/issues/42782


wilkinsona:
As far as I can tell, this is also a problem when `useCodeAsDefaultMessage` is `false` and I've convinced myself that we should treat it as a bug.

PR #45240, opened 2025-04-20
Restore visibility for DockerComposeProperties.get()

izeye:
This PR reverts the visibility change for the `DockerComposeProperties.get()` that has been made in #45014  as it doesn't seem to be necessary.


philwebb:
Thanks @izeye, I totally missed that change.

Issue #29918, opened 2022-02-21
PortInUseException is not thrown if system language is not english


mhalbritter:
System language: german

The test `Jetty10ServletWebServerFactoryTests#portClashOfPrimaryConnectorResultsInPortInUseException` only works if system language is set to english (`export LANG=en`). But this should be a problem with production code, too.

The code in `org.springframework.boot.web.server.PortInUseException#ifPortBindingException` checks if the error message contains the words "in use", which is not the case on non-english systems. On a german system the message is "java.net.BindException: Die Adresse wird bereits verwendet" for example.

One can workaround by setting `LANG=en java -jar ...` when starting the application, then the error message will be english and `PortInUseException` is thrown.

viktorardelean:
@mhalbritter I have checkout out the 2.5.x branch and I cannot see the portClashOfPrimaryConnectorResultsInPortInUseException test method in the org.springframework.boot.web.embedded.jetty.Jetty10ServletWebServerFactoryTests class.

wilkinsona:
It is inherited from a super-class: https://github.com/spring-projects/spring-boot/blob/1872af056ec059db2316db40ef19e2c1801af208/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/web/servlet/server/AbstractServletWebServerFactoryTests.java#L916

viktorardelean:
@wilkinsona Thanks for pointing it out! The problem here is that BindException contains just the error message - which is different for every language. Do you have any thoughts on what would be a good fix for this?
Should we just remove the if condition: `if (bindException.getMessage().toLowerCase().contains("in use"))` and accept all `BindExceptions `in the `ifPortBindingException `method?

mhalbritter:
The error message comes from the operating system, and depends on the `LANG` environment variable. We will have to find a better way to determine if there's a port clash going on than parsing the (localized) message. How we do that, I don't know at the moment.

somayaj:
Can we try based on the OS type -- lsof, netstat use for this?

mhalbritter:
I don't think that's worth it.

somayaj:
Well them reading the OS error codes, 98 (EADDRINUSE) or 10048 (WSAEADDRINUSE) ..somewhere in the error stack would do it?

wilkinsona:
As far as I know, those error codes do not appear anywhere in the stack trace of the `BindException`. If you have evidence to the contrary, please do share it.

somayaj:
try {
[...]

PR #45245, opened 2025-04-21
PortInUseFailureAnalyzer should attempt to detect the bounding PID in case of a port conflict

yonatang:
## Description
Right now, if the port that Spring attempts to bind if already in use, Spring Boot will provide you a general notification saying the port is bound, without any other details:
```
Description:

Web server failed to start. Port 8080 was already in use.

Action:

Identify and stop the process that's listening on port 8080 or configure this application to listen on another port.
```

It could be very convenient if the analyzer will make a best-effort attempt to detect the bounding PID and provide it as part of the failure details. Something like this:
```
Description:

Web server failed to start. Port 8080 was already in use by java (PID: 76781).

Action:

Stop the process java (PID: 76781) that's listening on port 8080 or configure this application to listen on another port.
```

That would be particularly useful when dealing with development environment with multiple services running at the same time, or a headless process that is still running after an IDE crash. 

## Implementation details

Java has no way to directly access the details of other processes, but since this is a best-effort attempt, we can try to invoke an OS-specific command to fetch the information. If the attempt to get it failed or timed out, we can fall back to the old generic message; either way to app failed to start, so no harm done.

## Assistance required
I don't have a windows environment, I would need help to check it works well in that OS. Running the PortInUseFailureAnalyzerTests unit test on a Windows machine should be suffice, it will invoke the necessary OS specific command.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/diagnostics/analyzer/PortInUseFailureAnalyzer.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/diagnostics/analyzer/PortInUseFailureAnalyzer.java
@@ -16,24 +16,170 @@
 
 package org.springframework.boot.diagnostics.analyzer;
 
+import java.io.BufferedReader;
+import java.io.IOException;
+import java.io.InputStreamReader;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+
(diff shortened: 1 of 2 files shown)


philwebb:
Thanks for the PR, but I personally think the complexity to benefit ratio is little high for this one. Flagging to see what the rest of the team think.

wilkinsona:
I'm not sure that it's worth investing much more in this failure analysis while it [only works in an English locale](https://github.com/spring-projects/spring-boot/issues/29918).

philwebb:
Thanks @wilkinsona, given that additional problem I'm pretty convinced that this isn't something we should add. Thanks anyway for the PR @yonatang!

PR #45250, opened 2025-04-22
feat(logging): add support for JSON structured logging via spring.logging.json.enabled

siripuramGanesh:
Problem

Spring Boot does not currently offer native support for enabling JSON-formatted structured logging via a configuration property. Developers often have to manually configure logback.xml or custom appenders to integrate with observability tools like ELK, Loki, or CloudWatch.

This creates unnecessary boilerplate and increases setup complexity, especially for developers who prefer convention over configuration.

Solution

This PR introduces a simple toggle to enable structured JSON logging using:

spring.logging.json.enabled=true

When enabled, it configures the root logger to use LogstashEncoder from the logstash-logback-encoder library, outputting structured logs in JSON format.

Key Highlights

New config binding class: JsonLoggingProperties

New auto-configuration: JsonLoggingAutoConfiguration (conditionally activated)

Integration with logstash-logback-encoder

Added spring.logging.json.enabled config flag

Registered auto-config in AutoConfiguration.imports

Included unit test to verify encoder is applied

No impact on existing users (opt-in feature only)

Example Usage

spring: logging: json: enabled: true

Once enabled, logs are output in structured format like:

{ "@timestamp": "2025-04-22T12:00:00.000+00:00", "level": "INFO", "message": "Application started", "logger_name": "com.example.Main", "thread_name": "main" }

Motivation

Simplifies developer setup for structured logging

Aligns with Spring Boot’s philosophy of zero-XML and config-based behavior

Improves compatibility with logging and observability pipelines

Let me know if anything needs improvement. Thanks for reviewing!

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/logging/JsonLoggingAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/logging/JsonLoggingAutoConfiguration.java
@@ -1,0 +1,35 @@
+package org.springframework.boot.autoconfigure.logging;
+
+import jakarta.annotation.PostConstruct;
+import net.logstash.logback.encoder.LogstashEncoder;
+import org.slf4j.LoggerFactory;
+import org.springframework.boot.autoconfigure.condition.ConditionalOnProperty;
+import org.springframework.boot.context.properties.EnableConfigurationProperties;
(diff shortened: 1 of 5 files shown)


wilkinsona:
This is already supported. You can set `logging.structured.format.console` (for console logging) or `logging.structured.format.file` (for file logging) as [described in the documentation](https://docs.spring.io/spring-boot/reference/features/logging.html#features.logging.structured). Updating XML is only necessary if you're using custom logging configuration. Thanks anyway for the PR.

PR #45251, opened 2025-04-22
Fail fast when both management 'base path' and 'endpoint mapping' are set to '/'

### LINKED ISSUE #43885: Fail fast if management base path and an endpoint mapping are both '/' on the main server port
Impossible for me to say if this is a bug or a feature/documented change, hence the ticket. Facts:
* the config below worked fine before the patch from #35426 (e.g. SB actuator 3.2.1) - actually, it's been used for literal ages, since early 2.x versions
* after going to any version with #35426 fix, (e.g. SB actuator 3.2.2) - it suddenly breaks. The RestController is no longer accessible at all, app acts like it "disappeared", even though it is still listed in the mappings

test snippets:

```kotlin
@RestController
@RequestMapping("/api")
class TestController {

    @PostMapping("endpoint")
    fun endpoint(): HttpEntity<*> {
        return ResponseEntity.EMPTY
    }

}
```

```yaml
management:
  endpoints.web:
    base-path: "/"
    path-mapping:
      health: "/"
```

results for 3.2.1:
```
GET -> 405
POST -> 200
PUT -> 405
```
results for 3.2.2:
```
GET -> 404
POST -> 405
PUT -> 405
```

Further details:

* The config shown doesn't throw any exception and never did.
* I also didn't notice any kind of warnings emitted with it (went through logs with DEBUG log level)
* Changing `health` from `/` to e.g. `/foo` "fixes" the behaviour (i.e. the endpoints start to work again)
* Changing `base-path` to anything else than `/` (e.g. `/foo`) in this case triggers: ` Ambiguous mapping. Cannot map 'Actuator root web endpoint' method Actuator root web endpoint to {GET [/foo], produces [application/vnd.spring-boot.actuator.v3+json || application/vnd.spring-boot.actuator.v2+json || application/json]}: There is already 'Actuator web endpoint 'health'' bean method`

My take on it is:

* If this config (i.e. both `base-path` and at least one of the mappings equal to `/`) is deemed invalid, it should throw an exception during bean init for this particular case, like the case with non-`/` base and `/` endpoint. However, the valid use case for this was hiding the "root" actuator endpoint AFAIK.
* If this config is deemed valid, the RestController should still work when that config is used.
---

With all honesty, I'd say this is a bug, mainly due to backwards-compatibility reasons. The code in question did work in previous patches in the same major/minor release (AFAIK both 3.2 and 3.1 are affected in this way), and a patch shouldn't introduce breaking changes that ain't backwards-compatible even if they are just fixing some abuses or corner cases (and the code shown isn't even really one of those I'd argue). If it would introduced purely with a minor increase and clearly documented, only the lack of exception thrown would be a problem here.

YongGoose:
fixes #43885

### DIFF
--- a/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/servlet/WebMvcEndpointManagementContextConfiguration.java
+++ b/spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/endpoint/web/servlet/WebMvcEndpointManagementContextConfiguration.java
@@ -69,6 +69,7 @@ import org.springframework.web.servlet.config.annotation.WebMvcConfigurer;
  *
  * @author Andy Wilkinson
  * @author Phillip Webb
+ * @author Yongjun Hong
  * @since 2.0.0
  */
 @ManagementContextConfiguration(proxyBeanMethods = false)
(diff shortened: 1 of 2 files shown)


philwebb:
Thanks very much @YongGoose. I've tweaked things a little to relocate the code to `WebEndpointAutoConfiguration` so that it also applies to Jersey.

PR #45253, opened 2025-04-22
Support system properties for Docker registries in Maven plugin

ryanjbaxter:
I would like to be able to configure Docker registry credentials in my global Maven settings.xml file. In order to do that we can introduce system properties for both the publish registry and builder registry.


philwebb:
Thanks for the PR @ryanjbaxter, but think with #45269 merged the extra properties are hopefully no longer necessary. I've tested things locally and no longer get the "Sign-in enforcement is enabled" message so hopefully the helper support will be enough.

ryanjbaxter:
Thanks @philwebb.  I tried 3.5.0-RC1 and it seemed to work without these properties set.  Question though, does that mean these properties will eventually be removed in a future release?

wilkinsona:
This PR wasn't merged so the properties were never added.

ryanjbaxter:
Sorry I mean the configuration properties for `DockerRegistry`s (publisher and builder) that can be configured in the POM XML.

wilkinsona:
We don't plan to remove them. They remain an alternative when a credential helper's not being used. If the properties are configured in your pom they will take precedence over any credential helper that may be configured.

ryanjbaxter:
Wouldn't this PR still be useful the case where the credential helper is not being used so the properties can easily be externalized to the global maven settings.xml file?

wilkinsona:
We're trying to strike a balance here and we think that what we have now achieves that as we expect most users to have a credential helper configured. For those that don't, we believe the existing support is sufficient – the properties can still be externalized by using the already-documented approach. The extra properties proposed here add more complexity than we feel is justified for something that should now be a niche requirement.

ryanjbaxter:
OK thanks for the clarification @wilkinsona!  We can always revisit this if we find there is a bigger demand for it.

Issue #44924, opened 2025-03-27
Rename spring.mvc.view.prefix and spring.mvc.view.suffix as they are specific to JSPs


wilkinsona:
The prefix suggests they apply MVC-wide, however they're used to configure `InternalResourceViewResolver` which is largely specific to JSPs.

wilkinsona:
I'm not sure what the property name should be. `InternalResourceViewResolver` is only largely and not entirely JSP-specific. The Web Method Security smoke test (among others) demonstrates this as it has some plain HTML files in `src/main/webapp` that are registered as views using `ViewControllerRegistry`.

wilkinsona:
We discussed this today and decided to leave the property names as they are.

PR #45254, opened 2025-04-22
Rename MVC view prefix/suffix properties to be JSP specific

big-cir:
Overview
---
- Updated WebMvcProperties to include a nested Jsp class for hierarchical binding.
- Registered a jspViewResolver bean when spring.mvc.jsp.prefix and spring.mvc.jsp.suffix are set.
- Added tests to verify property binding for JSP-specific prefix and suffix.

Related
---
ISSUE: #44924 

Note
---
I’m aware that the issue is currently labeled with `status: pending-design-work`, which suggests that the final design direction might still be under consideration. This implementation reflects my current understanding of the issue and is intended to serve as a starting point for discussion.

If this approach doesn't align with the intended design direction, please feel free to close the PR or share feedback — I’m happy to revise or take a different direction as needed. Thank you!


big-cir:
> Thanks for your efforts here @big-cir, but issue #44924 is labeled as `pending-design-work` which means we aren't ready to accept contributions for it.

Thank you for letting me know! I'll keep an eye on the issue and check back once the design work is complete. Have a great day!

PR #45267, opened 2025-04-23
Configuration properties changelog generator misses items that are removed without prior deprecation

### LINKED ISSUE #40075: Configuration properties changelog generator misses items that are removed without prior deprecation
Sometimes we have to remove a configuration property without it being deprecated first. An example of this is when we upgraded to Flyway 10. They removed a handful of settings without any replacement so our properties had to be updated to match. These changes were reflected in the metadata in [this commit](https://github.com/spring-projects/spring-boot/commit/62f922369d2714847f8e0f559e282f62d8eac63d) but the generator does not include them in the changelog.

yybmion:
## Issue
Closes #40075 

The configuration properties changelog generator misses items that are removed without prior deprecation. When a property is removed without being deprecated first (like during the Flyway 10 upgrade), these changes are reflected in the metadata but not included in the generated changelog.

## Solution
Modified the `computeDifferences` method in the `Changelog` class to detect properties that only exist in the new metadata with error-level deprecation and properly mark them as DELETED in the changelog. This ensures that properties that are added to a new version with error-level deprecation (indicating immediate removal) are correctly included in the "Removed in X.X" section of the changelog.

Specifically, added logic to check for properties in the new metadata that
1. Don't exist in the old metadata
2. Have an error-level deprecation
3. Should be classified as DELETED rather than being ignored

## Tests
The existing test `diffContainsDifferencesBetweenLeftAndRightInputs` already verifies this functionality by testing that properties with error-level deprecation are properly classified as DELETED. 

To make the test cover this fix, we enhanced the `sample-2.0.json` test data to include an additional property (`test.removed.directly`) that demonstrates a property removed without prior deprecation, marked with error-level deprecation and a reason "Removed in Upgrade 10".

After fix, this test confirms that both regular removed properties and those marked with error-level deprecation are correctly included in the changelog.

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-configuration-metadata-changelog-generator/src/main/java/org/springframework/boot/configurationmetadata/changelog/Changelog.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-configuration-metadata-changelog-generator/src/main/java/org/springframework/boot/configurationmetadata/changelog/Changelog.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2012-2023 the original author or authors.
+ * Copyright 2012-2025 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -21,6 +21,7 @@ import java.util.List;
 
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataProperty;
 import org.springframework.boot.configurationmetadata.ConfigurationMetadataRepository;
+import org.springframework.boot.configurationmetadata.Deprecation;
 
 /**
  * A changelog containing differences computed from two repositories of configuration
@@ -32,6 +33,7 @@ import org.springframework.boot.configurationmetadata.ConfigurationMetadataRepos
  * @author Stephane Nicoll
  * @author Andy Wilkinson
  * @author Phillip Webb
+ * @author Yoobin Yoon
  */
 record Changelog(String oldVersionNumber, String newVersionNumber, List<Difference> differences) {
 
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot-tools/spring-boot-configuration-metadata-changelog-generator/src/main/java/org/springframework/boot/configurationmetadata/changelog/Changelog.java
```diff
@@ -48,14 +50,24 @@ static List<Difference> computeDifferences(ConfigurationMetadataRepository oldMe
...
-			if (difference != null) {
-				differences.add(difference);
+			if (newProperty == null) {
+				differences.add(new Difference(DifferenceType.DELETED, oldProperty, null));
+			}
+			else {
+				Difference difference = Difference.compute(oldProperty, newProperty);
+				if (difference != null) {
+					differences.add(difference);
+				}
```
wilkinsona, line 60: This change isn't necessary as `Difference.compute` already handles the case where `newProperty` is `null`. Additionally, it checks that the old property isn't deprecated at error level before reporting a difference, something that's lost here.

I'll revert this part of the changes and add a test for this scenario (removing metadata for a property that's deprecated at error level) as part of merging this.

wilkinsona:
Thanks very much, @yybmion.

spring-boot-project/spring-boot-tools/spring-boot-configuration-metadata-changelog-generator/src/main/java/org/springframework/boot/configurationmetadata/changelog/Changelog.java
yybmion: I appreciate your explanation about why that part of the change wasn't necessary and how the existing code already handles the case.

yybmion:
Thank you @wilkinsona for reviewing and merging the PR!

Issue #18816, opened 2019-10-30
@ConfigurationProperties should have an option to fail fast for unresolvable properties


anno1985:
When using placeholders in `application.yaml` (property files/externalised configuration) together with the `@ConfigurationProperties` annotation, there should be an option to have Spring **_fail fast at startup_** when a defined property is not found.

**Example:**
```yaml
example.key: ${MY_ENV_VAR}
```

```java
@ConfigurationProperties(prefix="example")
public class AcmeProperties {
  public String key;
  // Getters, setters, constructors omitted... 
}
```

**Precondition:**
**No** environment variable `MY_ENV_VAR` defined.

**Current behaviour:**
`key` above is populated with the verbatim String `${MY_ENV_VAR}`

**Expected/desired behaviour:**
An exception is thrown while the application is starting up.

**Potential cause:**
Hardcoded flag in https://github.com/spring-projects/spring-boot/blob/v2.2.0.RELEASE/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/context/properties/bind/PropertySourcesPlaceholdersResolver.java#L51

**Further information:**
https://stackoverflow.com/q/58622189/2018047

**NB:**
Validating the String with `@Validated` after (failed) resolution is not the same as checking if resolution fails. Also, when no fail-fast is active, it might potentially be better to mirror `System.getenv("MY_ENV_VAR")`'s [behaviour](https://docs.oracle.com/javase/7/docs/api/java/lang/System.html#getenv(java.lang.String)), and return `null` instead of the actual placeholder, `${MY_ENV_VAR}`, verbatim.

philwebb:
I believe we intentionally set `ignoreUnresolvablePlaceholders` to `true` to keep back compatibility with previous releases. I agree it would be nice to have a fail fast option.

philwebb:
See also #8693

philwebb:
@mbhave also tracked down some internal discussion we had about this last time around:

From @wilkinsona
> Is it by design that the 2.0 binder doesn’t ignore placeholder resolution failures? Someone on Gitter is complaining and I can’t remember if it’s intentional or not. Here’s a little app that starts fine with 1.5.x but fails with 2.0.x:

```java
package com.example.demo;

import org.springframework.boot.SpringApplication;
import org.springframework.boot.autoconfigure.SpringBootApplication;
import org.springframework.boot.context.properties.ConfigurationProperties;
[...]

PR #45341, opened 2025-04-30
Add a property to fail fast for unresolvable placeholders in @ConfigurationProperties

nosan:
Introduce the `spring.configurationproperties.ignore-unresolvable-placeholders` property to control whether an exception should be thrown when a property placeholder `${...}` cannot be resolved.

See https://github.com/spring-projects/spring-boot/issues/18816


snicoll:
Thanks for the PR but I don't think we should focus on this property alone. We might want to customize the property replacement altogether and do it consistently (i.e. for the whole app). I am going to continue brainstorming on the original issue and your insights are very much appreciated.

nosan:
Thanks, @snicoll 

Indeed, this PR was focused solely on `@ConfigurationProperties`, as it addressed the original issue and I didn't want to include more than necessary. I agree that it's better to have a broader option to control the entire application rather than just a part of it.

PR #45044, opened 2025-04-09
Ci spring boot

xiongemi:

PR #44811, opened 2025-03-20
Kotlin plugin staging

lourw:

PR #45344, opened 2025-05-01
Nx cloud testing 3

lourw:


bclozel:
Similar to https://github.com/spring-projects/spring-boot/pull/45044

philwebb:
And #45008 and #44811.

@lourw could you please try and be more careful before submitting pull-requests.

PR #45509, opened 2025-05-13
org/springframework/boot/logging/log4j2/log4j2.xml should not expose a FILE_LOG_PATTERN property as it only configures the console appender

quaff:


ig-raiderler:
Impressive project structure! The architecture really demonstrates solid design principles.

igs-raiderler:
Very clean code with good test coverage. This is a great example of best practices.

philwebb:
I don't think we've documented it, but I guess it's possible someone might be (including that configuration](https://logging.apache.org/log4j/2.x/manual/configuration.html#xinclude) and using the property. I think the risk is small, but just to be safe lets only do this in 4.0.x

snicoll:
Merged in https://github.com/spring-projects/spring-boot/commit/ee90996e5f840773dd7bac8433a39f29c5c8b3fb

PR #45521, opened 2025-05-13
Polish MetricsAutoConfiguration

quaff:
Align `meterRegistryPostProcessor` with `meterRegistryCloser`, the `@Bean` method shouldn't be public if the return type is not public.


snicoll:
Closed by https://github.com/spring-projects/spring-boot/commit/c25e330b2e0d5396fb55a44dbef774ace3e7510c

Issue #42396, opened 2024-09-20
Duplicate meter binding when context contains multiple registries, none are primary, and one or more is a composite


wilkinsona:
See #42310 for discussion and background.

When there are multiple `MeterRegistry` beans and none of them is marked `@Primary`, auto-configuration creates a `@Primary` composite meter registry that contains every `MeterRegistry` bean. In such a situation, `MeterRegistryPostProcessor` will bind and `MeterBinder` beans to every `CompositeMeterRegistry` in the context. This works as intended when the auto-configured composite is the only `CompositeMeterRegistry`. However, when there's a user-defined composite as well as the auto-configured composite, it results in meters being bound twice to the user-defined composite: once directly when the user-defined composite is post-processed and once when the auto-configured composite cascades the binding down to the meter registries that it contains.

When the context contains an auto-configured composite, we should ignore every meter registry other than the auto-configured composite.

Issue #42310, opened 2024-09-13
Make sure MeterBinder and CompositeMeterRegistry registration behavior is consistent


jonatan-ivanov:
First of all, maybe this is intentional and Boot should do what it does today but maybe the behavior of registering `MeterBinder` and `CompositeMeterRegistry` have some inconsistencies (maybe `MeterFilter` too).

In  the latest GA version of Boot (`3.3.3`), if I create a custom registry bean, let's say a `SimpleMeterRegistry` in addition to what Boot creates, (let's say `PrometheusMeterRegistry`), binders are bound once, and the registry that will be injected if I ask for a `MeterRegistry` is `AutoConfiguredCompositeMeterRegistry` (what binders bound to).
```java
@Bean
SimpleMeterRegistry simpleMeterRegistry() {
    return new SimpleMeterRegistry();
}

@Bean
MeterBinder debuggingHelperBinder() {
	return registry -> {
		System.out.println("Binder bound to " + registry);
		registry.counter("test").increment();
	};
}
```
This behavior looks expected to me.

If I also create a `CompositeMeterRegistry` on top of the previous scenario (so I will have 1. Composite, 2. Simple, 3. Prometheus) Boot will create an `AutoConfiguredCompositeMeterRegistry` where all three registries are registered (expected) but it will also bind every binders to two registries: to `AutoConfiguredCompositeMeterRegistry` (expected) and to the `CompositeMeterRegistry` that I created (unexpected?).

```java
@Bean
SimpleMeterRegistry simpleMeterRegistry() {
	return new SimpleMeterRegistry();
}

@Bean
CompositeMeterRegistry compositeMR() {
	CompositeMeterRegistry compositeMeterRegistry = new CompositeMeterRegistry();
	compositeMeterRegistry.add(new SimpleMeterRegistry());
	return compositeMeterRegistry;
}

@Bean
MeterBinder debuggingHelperBinder() {
	return registry -> {
		System.out.println("Binder bound to " + registry);
		registry.counter("test").increment();
	};
}
```
This will result in the following structure:
```
AutoConfiguredCompositeMeterRegistry
    PrometheusMeterRegistry
    SimpleMeterRegistry
    CompositeMeterRegistry
        SimpleMeterRegistry
```

Should there be two composites or should Boot back off and use the user-created composite to register other registries? If so, should binders be bound to all the composites?

wilkinsona:
[...]

PR #45536, opened 2025-05-14
Exclude AutoConfiguredCompositeMeterRegistry from global registry

quaff:
I think it's not necessary to add Spring Boot's auto-configured one to global registry, feel free to close if I'm wrong.


wilkinsona:
We spent quite a bit of time analysing this in https://github.com/spring-projects/spring-boot/issues/42310 and made some changes in https://github.com/spring-projects/spring-boot/issues/42396. Without a concrete example of a problem caused by the current arrangement, I think we should leave things as they are. Thanks anyway.

Issue #44849, opened 2025-03-23
Document the way that primary Kotlin constructors are used when binding


vominhtri231:
I have a spring boot - Kotlin project, which I am upgrading from Spring 5, Java 8 to Spring 6, Java 21

`build.gradle.kts`

```
plugins {
    kotlin("jvm") version "2.1.20"
    id("org.springframework.boot") version "3.4.4"
    id("org.jetbrains.kotlin.plugin.spring") version "2.1.20"
    id("io.spring.dependency-management") version "1.1.7"
}

dependencies {
    implementation("org.springframework.boot:spring-boot-starter")
    implementation(kotlin("reflect"))
}
```

`MyProperties.kt`

```
@ConfigurationProperties(prefix = "my.test")
@Component
class MyProperties {
    var customs: List<CustomA> = listOf()
    var bases: List<Base> = listOf()

    open class Base(
        var type: String = ""
    )

    class CustomA(
        var y: String = "",
        var z: String = "",
    ) : Base()
}
```

Run the application by `bootRun` task, with these env variables:
```
MY_TEST_BASES_0_TYPE=value4;MY_TEST_CUSTOMS_0_TYPE=value3;MY_TEST_CUSTOMS_0_Y=value1;MY_TEST_CUSTOMS_0_Z=value2
```

I would expect that the `myProperties.customs[0].type` would return `value3`, but it would return an empty string instead.

The problem only happens if the `kotlin("reflect")` dependency exists and does not occur in Spring 5, Java 8.

It looks like the base class's properties are not bound at all. If I modify the `MyProperties` class as this, the `type` property will bind correctly

```
open class Base(
   open var type: String = ""
)

class CustomA(
   override var type: String = "",
   var y: String = "",
   var z: String = "",
) : Base()
```

philwebb:
Could you please provide a complete sample application that works with earlier version and fails with the later one. Please either attach a zip file or provide a link to a GitHub project.

vominhtri231:
Here is the Github project that indicates the problem: https://github.com/vominhtri231/PropertyNotMap

quaff:
I confirm it's a regression.

The `Binder` in v2.7.8 instantiate `CustomA` with default constructor parameters, then bind properties via setters (include setters from super classes).
The `Binder` in v3.4.4 instantiate `CustomA` with provided configuration properties, no setter invoked later, I think invoking setters in super classes should fix it.

quaff:
[...]

PR #45554, opened 2025-05-15
Remove unused imports in sample

quaff:


snicoll:
See #44849

snicoll:
Good catch @quaff - I've created https://github.com/spring-projects/spring-boot/issues/45555 to see what we can do to prevent that going forward.

Issue #45599, opened 2025-05-19
Micrometer "enable" annotations property does not cover observed aspect


ggsurrel:
I am working on a large project with some AOP in it, and evaluating pointcuts can be a bit lenghty when starting the application. So I would like to get rid of the unused ones.

In order to do this, it would be nice to have a way to disable `ObservedAspectConfiguration`: the class is package protected so I cannot exclude it from the autoconfiguration, and there is no `ConditionalOnProperty` annotation on it either.

bclozel:
It looks like we missed the `ObservedAspect` when we took care of #39128. I think it's fair to assume that `management.observations.annotations.enabled=false` should do this already.

nosan:
@ggsurrel 
As a temporary solution, you can register your own `BeanDefinitionRegistryPostProcessor` to exclude `ObservedAspect`:

```java
    @Bean
    static BeanDefinitionRegistryPostProcessor excludeObservedAspect() {
        return (registry) -> {
            if (registry.containsBeanDefinition("observedAspect")) {
                registry.removeBeanDefinition("observedAspect");
            }
        };
    }

```

jidonghao001:
good idea

PR #45601, opened 2025-05-19
Micrometer "enable" annotations property does not cover observed aspect

nosan:
See #45599 

When merging this into **main**, `ObservationAnnotationsEnabledCondition` should be removed and replaced with `@ConditionalOnBooleanProperty("management.observations.annotations.enabled")`, and the test `allowsObservedAspectToBeEnabledWithDeprecatedProperty` should also be removed.

Target branch is **3.3.x**


nosan:
@bclozel 

 I've created this PR based on `3.3.x` because `MetricsAspectsAutoConfiguration` already contains this logic in `3.3.x`. Shouldn't it be merged into `3.3.x`?

PR #45624, opened 2025-05-20
Make heapdump endpoint restricted by default

lhotari:
In Spring Boot, all actuator endpoints other than `shutdown` are "unrestricted" by default.
For misconfigured Spring Boot applications, it would improve security when the `heapdump` actuator endpoint is restricted by default.


philwebb:
Thanks! Very timely given we were just discussing [your post](https://bsky.app/profile/lhotari.bsky.social/post/3lpmbdrbzd22a) about this :)

philwebb:
Thanks very much @lhotari !

Issue #45623, opened 2025-05-20
Enforce that @ConfigurationPropertiesBinding bean methods are static


wilkinsona:
See #45621 for why this is necessary. Depending on the outcome of #45622, we may also want to enforce that they're have the infrastructure role.

wilkinsona:
Closing in favor of #45640.

PR #45640, opened 2025-05-22
Enforce that `@ConfigurationPropertiesBinding` bean methods are static

ngocnhan-tran1996:
Issues gh-45623


quaff:
You should update `ArchitectureRules` also.

PR #45642, opened 2025-05-23
Reuse properties from field to create PropertiesKafkaConnectionDetails

quaff:
AutoConfigurations such as `RabbitAutoConfiguration` and `CassandraAutoConfiguration` will reuse properties injected by the constructor.


wilkinsona:
Thanks, @quaff.

Issue #45636, opened 2025-05-21
SAML2 autoconfiguration is not imported by @WebMvcTest


gbaso:
I have an application configured with both OIDC and SAML2 authentication. When creating a test slice with `@WebMvcTest` the spring security configuration for OIDC is imported, while the configuration for SAML2 is not, therefore the tests don't work as expected.

The [AutoConfigureMockMvc import list](https://github.com/spring-projects/spring-boot/blob/main/spring-boot-project/spring-boot-test-autoconfigure/src/main/resources/META-INF/spring/org.springframework.boot.test.autoconfigure.web.servlet.AutoConfigureMockMvc.imports) has `org.springframework.boot.autoconfigure.security.oauth2.client.servlet.OAuth2ClientAutoConfiguration`, but `org.springframework.boot.autoconfigure.security.saml2.Saml2RelyingPartyAutoConfiguration` is missing.

minimal sample: https://github.com/gbaso/web-mvc-test-saml2

snicoll:
Closing in favor of PR #45650. Thanks @nosan!

PR #45650, opened 2025-05-23
SAML2 autoconfiguration is not imported by @WebMvcTest

nosan:
See #45636


wilkinsona:
Thanks for the PR. At the risk of being a pain, could you please move the OAuth2-related changes out into a separate PR?

PR #45684, opened 2025-05-26
Example of using prometheus-metrics-exporter-pushgateway has wrong artifactId

juliojgd:
MAven coordinates for io.prometheus:prometheus-metrics-exporter-pushgateway were duplicating the groupId inside the artifactId.


snicoll:
Good catch, thanks @juliojgd

PR #45733, opened 2025-05-31
Assign distinct order values to Management customizers to prevent con…

### LINKED ISSUE #45728: ManagementWebServerFactoryCustomizer and ManagementErrorPageCustomizer should not have the same order
This seems to have worked thus far by accident. They're both at order 0 and one is clearing the error pages while the other is registering an error page. 

Flagging for team-attention as I am not sure where to fix this.

naveenprasanth2:
Closes #45728

Currently, both `ManagementWebServerFactoryCustomizer` and `ManagementErrorPageCustomizer` return the same order value, which can lead to non-deterministic behavior during application startup.

This pull request assigns distinct order values using the `Ordered` interface:
- `ManagementWebServerFactoryCustomizer`: returns `Ordered.HIGHEST_PRECEDENCE + 10`
- `ManagementErrorPageCustomizer`: returns `Ordered.HIGHEST_PRECEDENCE + 11`

This change ensures a deterministic execution order for the management-related web server customizers.


bclozel:
Thanks for the proposal but the issue is already assigned

Issue #44971, opened 2025-04-01
Rename spring.codec properties to spring.http.codecs


wilkinsona:
The configuration is specific to HTTP (both client and server) so the properties should move beneath `spring.http`. `codec` is changed to `codecs` to align with `spring.mvc.converters` / `spring.http.converters` on the non-reactive side that serves a similar purpose.

wilkinsona:
Closed by https://github.com/spring-projects/spring-boot/commit/f039c73db7f5e387701e3a6fd46fb99cb9a630a6.

PR #45743, opened 2025-06-02
Deprecation replacement for spring.codec.* properties has a typo

Peksa:
The replacement properties for the now deprecated configuration properties `spring.codec.*` have a slight typo in them:
`spring.http.codec.log-request-details`
`spring.http.codec.max-in-memory-size`

The correct replacement properties should be:
`spring.http.codecs.log-request-details`
`spring.http.codecs.max-in-memory-size`

(note the missing **s** in codec**s**)

I noticed this when upgrading to Spring Boot 3.5 and saw the warning in IntelliJ:
image

image

See #44971


snicoll:
Good catch, thanks @Peksa!

PR #45795, opened 2025-06-05
Update auditevents.adoc

tdinev:
Adapt date in explanation to snippet content


wilkinsona:
Thanks for the proposal but this will be wrong again the next time the documentation is built as the snippet's content is dynamic. Instead, `filteredAuditEvents()` in `AuditEventsEndpointDocumentationTests` should be updated to use a hard-coded timestamp as `allAuditEvents()` already does. 

Would you like to rework your PR to make this change to the test instead?

tdinev:
Certainly. I apologize; I made the change without even consulting the snippet.

tdinev:
Sorry, but I cannot find the referenced `curl-request.adoc`. Could you point me towards its location?

PR #45801, opened 2025-06-06
fix(docs): Fix broken links to testcontainers-junit-jupiter

maringuu:
The updated links were semi-broken. They did not point to where they shall point, but redirected to the docs startpage.

The build fails due to the following, so please dont start the CI yet :)
I am struggeling to find out where the version is picked up from.
Grepping for the other testcontainer url definitions did lead me to the [anthora.yml](spring-boot-project/spring-boot-docs/src/docs/antora/antora.yml) which apparently picks the versions up from the version catalog.
However I struggle to find the catalog.

I'd be happy if a maintainer could point me to the right place or fixup the commit directly.
Thanks!


wilkinsona:
Thanks for bringing this to our attention, @maringuu. The problem is broader than just the JUnit Jupiter extension as it affects classes in every module other than the core `testcontainers` module. This will require quite a different fix to the one proposed here so I have opened #45802. Thanks anyway for the PR.

maringuu:
> This will require quite a different fix to the one proposed here so I have opened https://github.com/spring-projects/spring-boot/issues/45802

Thank you! While its not probable that I find the time soon I'll probably get back to this issue if it isn't closed yet :)

Issue #45846, opened 2025-06-06
Setting spring.reactor.context-propagation has no effect when lazy initialization is enabled


Akaame:
Setting `spring.main.lazy-initialization: true` effectively disables the `spring.reactor.context-propagation: AUTO`, as the `ReactorAutoConfiguration`[^1] is never instantiated during the application startup. Turning on eager initialization alleviates this problem, which leads me to believe that this particular setup was overlooked when the `AutoConfiguration` class what added. Hence, I created this issue, even though the text reads more like a question.

We are currently doing the following

```java
@AutoConfiguration
@ConditionalOnClass(Hooks.class)
@EnableConfigurationProperties(ReactorProperties.class)
@Lazy(false)
public class EagerReactorAutoConfiguration {
  EagerReactorAutoConfiguration(ReactorProperties properties) {
    if (properties.getContextPropagation() == ReactorProperties.ContextPropagationMode.AUTO) {
      Hooks.enableAutomaticContextPropagation();
    }
  }
}
```

This is not elegant at all and I am not super sure if there is a way of forcing eager auto configuration for such cases (or if this should be rather fixed at Spring Boot level).

Cheers!

[^1]: [ReactorAutoConfiguration](https://github.com/spring-projects/spring-boot/blob/139436ea23cbefbe315aaf17de9d4c50483eb8b4/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/reactor/ReactorAutoConfiguration.java#L35)

wilkinsona:
You're right, this is a bug. Until it's fixed, you can use a filter to exclude the `ReactorAutoConfiguration` bean from lazy init:

```
@Bean
LazyInitializationExcludeFilter eagerReactorAutoConfiguration() {
	return LazyInitializationExcludeFilter.forBeanTypes(ReactorAutoConfiguration.class);
}
```

Akaame:
That worked out perfectly for our setup. Thanks for the quick answer @wilkinsona 🍺

philwebb:
We discussed this today and we've decided to add the `LazyInitializationExcludeFilter` directly in our auto-configuration. For 4.x we'll try and change the static calls. I've opened #46014 to deal with that.

PR #45853, opened 2025-06-08
Setting spring.reactor.context-propagation has no effect when lazy initialization is enabled 

nosan:
See https://github.com/spring-projects/spring-boot/issues/45846

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/reactor/ReactorAutoConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/reactor/ReactorAutoConfiguration.java
@@ -18,10 +18,12 @@ package org.springframework.boot.autoconfigure.reactor;
 
 import reactor.core.publisher.Hooks;
 
+import org.springframework.boot.LazyInitializationExcludeFilter;
 import org.springframework.boot.autoconfigure.AutoConfiguration;
 import org.springframework.boot.autoconfigure.EnableAutoConfiguration;
 import org.springframework.boot.autoconfigure.condition.ConditionalOnClass;
 import org.springframework.boot.context.properties.EnableConfigurationProperties;
+import org.springframework.context.annotation.Bean;
 
 /**
  * {@link EnableAutoConfiguration Auto-configuration} for Reactor.
@@ -38,6 +40,11 @@ public class ReactorAutoConfiguration {
 		if (properties.getContextPropagation() == ReactorProperties.ContextPropagationMode.AUTO) {
 			Hooks.enableAutomaticContextPropagation();
 		}
+	}
+
+	@Bean
+	static LazyInitializationExcludeFilter reactorAutoConfigurationLazyInitializationExcludeFilter() {
+		return LazyInitializationExcludeFilter.forBeanTypes(ReactorAutoConfiguration.class);
 	}
 
 }
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/reactor/ReactorAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/reactor/ReactorAutoConfigurationTests.java
@@ -28,6 +28,7 @@ import reactor.core.publisher.Hooks;
 import reactor.core.publisher.Mono;
 import reactor.util.context.Context;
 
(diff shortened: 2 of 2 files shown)


nosan:
I was initially thinking of the following approach:

```java

@AutoConfiguration
@ConditionalOnClass(Hooks.class)
@ConditionalOnProperty(name = "spring.reactor.context-propagation", havingValue = "auto")
@EnableConfigurationProperties(ReactorProperties.class)
public class ReactorAutoConfiguration {

	ReactorAutoConfiguration() {
		Hooks.enableAutomaticContextPropagation();
	}

	@Bean
	static LazyInitializationExcludeFilter reactorLazyInitializationExcludeFilter() {
		return LazyInitializationExcludeFilter.forBeanTypes(ReactorAutoConfiguration.class);
	}

}

```

However, `ReactorProperties` will stop being registered, which makes this approach somewhat risky to merge.


As a good alternative, maybe something like this:

```java

@AutoConfiguration
@ConditionalOnClass(Hooks.class)
@EnableConfigurationProperties(ReactorProperties.class)
public class ReactorAutoConfiguration {

	@ConditionalOnProperty(name = "spring.reactor.context-propagation", havingValue = "auto")
	static class ReactorContextPropagationConfiguration {

		ReactorContextPropagationConfiguration() {
			Hooks.enableAutomaticContextPropagation();
		}

		@Bean
		static LazyInitializationExcludeFilter reactorContextPropagationLazyInitExcludeFilter() {
			return LazyInitializationExcludeFilter.forBeanTypes(ReactorContextPropagationConfiguration.class);
		}

	}

}

```

wilkinsona:
Arguably, we shouldn't be relying on a side-effect of the constructor being called to configure the hooks and I'd prefer to use a specific callback mechanism instead. For example, implementing `SmartInitializingSingleton` and configuring the hooks in `afterSingletonsInstantiated` fixes the problem without the need for an exclude filter.

Flagging for team attention as I'm not 100% sure that `SmartInitializingSingleton` is the ideal callback here. @snicoll in particular may have an alternative suggestion.

nosan:
Thanks, @wilkinsona 


`SmartInitializingSingleton` is a decent option. However, it has one downside: `enableAutomaticContextPropagation` will not be visible to other `SmartInitializingSingleton` beans, whereas the constructor-based approach does not have such an issue. The following test demonstrates the issue:

```java

	@Test
	void propagationShouldBeAppliedToSmartInitializingSingletonBeans() {
		this.contextRunner.withPropertyValues("spring.reactor.context-propagation=AUTO")
			.withInitializer(
					(context) -> context.addBeanFactoryPostProcessor(new LazyInitializationBeanFactoryPostProcessor()))
			.withBean(VerySmartSingleton.class)
			.run((context) -> assertThat(context.getBean(VerySmartSingleton.class).getValue()).isEqualTo("updated"));
	}

	static class VerySmartSingleton implements SmartInitializingSingleton {

		private final AtomicReference<String> threadLocalValue = new AtomicReference<>();

		String getValue() {
			return this.threadLocalValue.get();
		}

		@Override
		public void afterSingletonsInstantiated() {
			Mono.just("test")
				.doOnNext((element) -> this.threadLocalValue.set(THREADLOCAL_VALUE.get()))
				.contextWrite(Context.of(THREADLOCAL_KEY, "updated"))
				.block();
		}

	}

```

wilkinsona:
Indeed. That's why it may not be ideal. There are still ordering concerns with the constructor-based approach as it leaves a window between a bean using Reactor during its standard creation/initialization and `ReactorAutoConfiguration` being created. However, using `SmartInitializingSingleton` widens the window.

nosan:
>  There are still ordering concerns with the constructor-based approach as it leaves a window between a bean using Reactor during its standard creation/initialization and ReactorAutoConfiguration being created. 

How about that one? 

```java

@AutoConfiguration
@ConditionalOnClass(Hooks.class)
@EnableConfigurationProperties(ReactorProperties.class)
public class ReactorAutoConfiguration {

	@ConditionalOnProperty(name = "spring.reactor.context-propagation", havingValue = "auto")
	static class ReactorContextPropagationConfiguration {

		@Bean
		static BeanFactoryInitializer<ListableBeanFactory> reactorContextPropagationBeanFactoryInitializer() {
			return (beanFactory) -> Hooks.enableAutomaticContextPropagation();
		}

	}

}
	
```

mhalbritter:
Oh, sorry @nosan, i missed your latest update and have implemented the same solution in https://github.com/spring-projects/spring-boot/commit/e2571a41bf2215cddcef9cfe0cccc42a70cec885.

nosan:
No worries, @mhalbritter 😃

PR #45855, opened 2025-06-08
Improve readability of CloudFoundryVcapEnvironmentPostProcessor by removing constants

wonyongg:
This pull request extracts the hardcoded "vcap" string into a named constant 
(`VCAP_PROPERTY_SOURCE_NAME`) within the `CloudFoundryVcapEnvironmentPostProcessor` class.

- Improves readability and consistency
- Makes the purpose of the property source name clearer
- Aligns with common practices for well-known property source identifiers

This is a minor internal refactoring and does not affect runtime behaviour.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudFoundryVcapEnvironmentPostProcessor.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/cloud/CloudFoundryVcapEnvironmentPostProcessor.java
@@ -91,10 +91,6 @@ import org.springframework.util.StringUtils;
  */
 public class CloudFoundryVcapEnvironmentPostProcessor implements EnvironmentPostProcessor, Ordered {
 
-	private static final String VCAP_APPLICATION = "VCAP_APPLICATION";
-
-	private static final String VCAP_SERVICES = "VCAP_SERVICES";
-
 	private final Log logger;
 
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the proposal.

The use of constants in this class is certainly a bit inconsistent at the moment. For example, we have  `VCAP_APPLICATION` and `VCAP_SERVICES` constants that are only used once but with the values repeated in string literal error messages. Then, as you've shown here, we have `"vcap"` that's used twice without a constant.

Flagging for team attention to see which direction we should go in:

1. Use constants consistently throughout the class
2. Remove the use of constants

wonyongg:
@wilkinsona 

Thanks for the thoughtful feedback. I'm happy to follow whatever direction the team decides.
Please feel free to let me know if you'd like me to update the PR accordingly.

mhalbritter:
I'd remove both existing constants from `CloudFoundryVcapEnvironmentPostProcessor`, i don't feel they make the code clearer.

wilkinsona:
@wonyongg could you please update your PR to remove the existing constants rather than introducing a new one?

wonyongg:
@wilkinsona 
Thanks for the clarification! I've updated the PR to remove all constants and replace them with string literals, as suggested.

Please have a look when you have time—I'd really appreciate your review. Thank you!

mhalbritter:
Thanks @wonyongg !

PR #45879, opened 2025-06-11
Tomcat and Jetty metrics Auto-configurations should be enabled only for servlet-based web application

quaff:
It's possible that Tomcat's `Manager.class` or Jetty's `Server.class` is present in reactive web applications.


bclozel:
@quaff These metrics seem to apply to both reactive and servlet applications. Can you explain why they don't make sense for reactive apps?

quaff:
> These metrics seem to apply to both reactive and servlet applications.

With my understanding, both Tomcat and Jetty are servlet container, reactive applications will not start them, feel free to close it if I'm wrong.

PR #45880, opened 2025-06-11
Polish usage of '@ConditionalOnWebApplication' for consistency

quaff:


philwebb:
It's unfortunately hard to come up with a general rules for these that we can enforce. In some cases (like this one), I like the proposed change. In others, I prefer the extra context of the enclosing class (e.g. `WebClient.Builder`).

We should probably just deal with most of these as we find them. I was going to close this issue and suggest we do that, but I do like the update. I think we might just leave this one for a little while until the restructure work lands because we don't want to cause more rebase pain.

Thanks for the PR @quaff.

wilkinsona:
Thanks, @quaff.

Issue #45875, opened 2025-06-10
Update javadoc of test slice annotations to suggest MockitoBean rather than MockBean


wilkinsona:

wilkinsona:
Closing in favor of #45887. Thanks for the PR, @ngocnhan-tran1996.

PR #45887, opened 2025-06-11
Update javadoc of test slice annotations to suggest MockitoBean rather than MockBean

ngocnhan-tran1996:
Issues: gh-45875


mhalbritter:
Thanks @ngocnhan-tran1996 !

PR #45888, opened 2025-06-11
Align default value of maxParameterCount with Tomcat (1000)

wonyongg:
This PR updates the default value of `maxParameterCount` in `ServerProperties.Tomcat` from `10000` to `1000` to align with the default value used in Tomcat.

See: [Tomcat source](https://github.com/apache/tomcat/blob/cd2cb43579161cdd17cdebf49d4a4b828a45cccb/webapps/docs/config/http.xml#L181)

If the current value was set to 10000 intentionally, please feel free to close this PR. Just wanted to flag the discrepancy in case it was unintentional.

If the current value is indeed intentional and you feel it would be helpful to clarify this in the JavaDoc, I’d be happy to submit a follow-up PR for that as well.

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/web/ServerProperties.java
@@ -517,7 +517,7 @@ public class ServerProperties {
 		 * Maximum number of parameters (GET plus POST) that will be automatically parsed
 		 * by the container. A value of less than 0 means no limit.
 		 */
-		private int maxParameterCount = 10000;
+		private int maxParameterCount = 1000;
 
 		/**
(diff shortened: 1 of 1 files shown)


wilkinsona:
You've linked to the documented for Tomcat 11 where the default was lowered. Tomcat 10.1's default [is 10000](https://github.com/apache/tomcat/blob/efcfb95dfdaf4b36233968537efc3e917c38bfa8/webapps/docs/config/http.xml#L182). We also have [a test](https://github.com/spring-projects/spring-boot/blob/5b759c750b4d21acf09120e6c34769cc4fd956a8/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/web/ServerPropertiesTests.java#L390-L394) that ensures that Tomcat's default and Boot's property default are aligned.

wonyongg:
@wilkinsona 
Thanks for the explanation! Makes sense now 🙏

Issue #45881, opened 2025-06-11
Expose new Tomcat properties maxPartCount and maxPartHeaderSize


bsanchezb:
Recent versions of Tomcat [11.0.8](https://tomcat.apache.org/tomcat-11.0-doc/changelog.html) and [10.1.42](https://tomcat.apache.org/tomcat-10.1-doc/changelog.html) introduced new system properties, namely [maxPartCount](https://github.com/apache/tomcat/commit/e34fe96ef8ee782b0e56b64358e8dc57cbe336a6#diff-cf7c488ef6191a648674047b1664a37570cc0282cdf90a4cc45e7c8837d2067dR184) and [maxPartHeaderSize](https://github.com/apache/tomcat/commit/e34fe96ef8ee782b0e56b64358e8dc57cbe336a6#diff-cf7c488ef6191a648674047b1664a37570cc0282cdf90a4cc45e7c8837d2067dR192).

Please provide the properties for configuration though an application.properties file.

wilkinsona:
Thanks for pointing these out.

Until the properties have been added, you can configure these settings using a `TomcatConnectorCustomizer`:

```java
@Bean
TomcatConnectorCustomizer connectorCustomizer() {
	return (connector) -> {
		connector.setMaxPartCount(10);
		connector.setMaxPartHeaderSize(1024);
	};
}
```

123Haynes:
@wilkinsona Please note that the new, enforced default values of these properties are a breaking change for some applications.  
Before there was no limit on Partcount and Headersize.  
It broke 2 of my more complex apps when I updated to Tomcat 10.1.42.  
Unfortunately without any error messages from tomcat.  
I think this should be at least mentioned in the release notes 😃

wilkinsona:
@123Haynes Apologies for the breakage. I believe that the Tomcat team are reviewing the new defaults.

123Haynes:
@wilkinsona oh no need to apologise 😄   
It's not a big deal once you know about it. I just wanted to give a headsup because this could break a lot of apps.  
So when Spring Boot pulls in Tomcat 10.1.42 this should be mentioned in the release notes so people can adjust the values as necessary :)  
Or if you want to restore the old behaviour, spring boot could set the default values to `-1` (no limits) until the next major release 😄

wilkinsona:
[...]

PR #45890, opened 2025-06-11
Add support for configuring Tomcat connector maxPartCount and maxPartHeaderSize

kwondh5217:
This PR addresses gh-45881.

- Expose `server.tomcat.max-part-count` (default 10)  
- Expose `server.tomcat.max-part-header-size` (default 512B)  
- Bind in `TomcatWebServerFactoryCustomizer`  
- Add tests for custom and default values


wilkinsona:
Thanks for the PR, @kwondh5217, but we've decided that we need to make this change in 3.3.x and the proposed changes do not apply cleanly to that branch. There are also a number of test failures that need to be addressed. Thanks again for the PR, but we think it's better if we take things from here.

Issue #32883, opened 2022-10-26
Make auto-configurations final and remove public members


mhalbritter:
Our auto-configurations are public, but users should reference only the names of these classes. The contained methods, inner classes etc. are not part of the public API (even if they are public by accident or technical reasons).

Make sure that all contained methods and inner classes are not public.

If a `@Bean` method or inner class is public, deprecate it so that users get a warning that they should not use that method. After some time we will remove the deprecation and drop the public modifier, removing it from public exposed API.

This will leave us with public auto-configuration classes which do not expose any inner classes or methods as public API.

wilkinsona:
I think we should aim to do this in a first milestone to give us much time as possible for feedback.

philwebb:
We can also add package-private constructors to stop folks subclassing them.

wilkinsona:
Beginnings of an Arch Unit rule to police method visibility:

```java
private ArchRule autoConfigurationClassesShouldHaveNoPublicMethods() {
	return ArchRuleDefinition.methods()
		.that()
		.areDeclaredInClassesThat()
		.areAnnotatedWith("org.springframework.boot.autoconfigure.AutoConfiguration")
		.should()
		.notBePublic()
		.allowEmptyShould(true);
}
```

Alternatively, if we want to deprecate them first:

```java
private ArchRule publicMethodsOnAutoConfigurationClassesShouldBeDeprecated() {
	return ArchRuleDefinition.methods()
		.that()
		.arePublic()
		.and()
		.areDeclaredInClassesThat()
		.areAnnotatedWith("org.springframework.boot.autoconfigure.AutoConfiguration")
		.should()
		.beAnnotatedWith(Deprecated.class)
		.allowEmptyShould(true);
}
```

There are 167 violations in `spring-boot-autoconfigure`:

```
Architecture Violation [Priority: MEDIUM] - Rule 'methods that are public and are declared in classes that are annotated with @AutoConfiguration should be annotated with @Deprecated' was violated (167 times):
Method <org.springframework.boot.autoconfigure.admin.SpringApplicationAdminJmxAutoConfiguration.springApplicationAdminRegistrar(org.springframework.beans.factory.ObjectProvider, org.springframework.core.env.Environment)> is not annotated with @Deprecated in (SpringApplicationAdminJmxAutoConfiguration.java:61)
[...]

PR #45891, opened 2025-06-12
Prevent @Bean method being public if it should not be

quaff:
Fix the violation that classes are exposed outside their defined visibility scope.


wilkinsona:
This is close to being part of https://github.com/spring-projects/spring-boot/issues/32883 although we want to go further and make all `@Bean` methods package-private. We're holding off on doing that at least until the restructuring work is complete to avoid lots of conflicts on merge or rebase. Unfortunately, I don't think we can use this PR as it'll either create lots of conflicts if we merge it now, or have to be almost completely redone once the restructuring has landed. Thanks anyway.

Issue #45842, opened 2025-06-06
Update documentation to reflect changes in TestRestTemplate's default redirect behavior


wilkinsona:
See https://github.com/spring-projects/spring-boot/issues/43431#issuecomment-2948712765

furmak331:
@wilkinsona Please check the PR #45964 . 
I tried to change the documentation . 
Please let me know if i have done correctly and if there are more changes needed in the documentation .

PR #45964, opened 2025-06-14
Update documentation to reflect changes in TestRestTemplate's default redirect behavior

furmak331:
fix issue:  #45842
I went though the previous PR and changed the documentation . 

I have removed :
Redirects are not followed  (so you can assert the response location).
I added :
By default, `TestRestTemplate` follows redirects in the same way as `RestTemplate`, regardless of which HTTP client implementation is used.

please feel free to suggest any changes .


wilkinsona:
Thanks for the proposal, @furmak331, but I don't think we need to mention that the redirect behaviour's the same. I also don't think it makes sense to keep a list of test-friendly features when there's only one.

Thanks again for the PR but I'm going to close this one as I think it would be better for a member of the core team to update the documentation. The alternative is for us to describe exactly how the changed documentation should read which isn't an efficient use of our time and probably isn't a satisfying way to contribute either.

furmak331:
Thanks for the feedback @wilkinsona  . I completely understand your perspective, having a core team member handle this makes more sense.
I'm still learning the best ways to contribute effectively to the project, so this kind of guidance is really valuable. I'll close this PR and look forward to finding more impactful ways to contribute in the future.

PR #45968, opened 2025-06-15
Improve properties bind performance by add enumerable propertySource

tanruian:
Update `ManagementContextAutoConfiguration`. use an EnumerablePropertySource to provide better property binding performance

Upgrade spring-boot to 3.5.0, an issue where a property source binds to java objects slowly, for deep nested collection (consists of 4 depth), around 1 minutes. Number of scans of property with exponential growth.
Because the ManagementContextAutoConfiguration automatically adds an un enumerable property source to the environment，resulting in the inability to confirm whether the property source contains bindable property, requiring guessing and traversing property one by one based on the index 0-12

#45970


philwebb:
We might also be able to implement `OriginLookup` with `isImmutable()` returning `true`.

philwebb:
Thanks very much @tanruian. I've made some changes for #45994 that should make these performance improvements less necessary. I still think they're worth fixing and I consider this a bug, but given it's less critical now the binder functionality has been restored, I'm only going to apply it to 3.5.x.

Thanks again for the PR!

PR #45976, opened 2025-06-16
 DataSouceBuilder can fail with a NPE when the driver is null

chanbinme:
This PR fixes a potential `NullPointerException` in `SimpleDataSourceProperties` when mapping the driver class name property.

When a `SimpleDriverDataSource` is created without explicitly setting a driver, calling `dataSource.getDriver().getClass()` throws NPE because `getDriver()` returns `null`.

Added null check to safely handle the case when no driver is configured, returning `null` instead of throwing NPE.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/jdbc/DataSourceBuilder.java
@@ -508,6 +508,9 @@ public final class DataSourceBuilder<T extends DataSource> {
 		}
 
 		private String convertToString(V value) {
+			if (value == null) {
+				return null;
+			}
 			if (String.class.equals(this.type)) {
 				return (String) value;
 			}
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR, @chanbinme. Could you please add a test that fails due to an NPE without the proposed change? If I've understood the problem correctly, I think that will require deriving a `DataSourceBuilder` from an existing `SimpleDriverDataSource` instance.

chanbinme:
@wilkinsona 

Thank you for the feedback! You're absolutely right about adding a test to demonstrate the NPE issue.

I'll add a test case following the existing naming convention, something like:
`buildWhenDerivedFromSimpleDriverDataSourceWithDriverNotSetSucceeds()`

While writing the test, I discovered an additional NPE issue in the `convertToString` method where null values were being passed, causing `NullPointerException` when trying to call `.getName()` on a null Class object. I've addressed this as well by adding proper null checks.

chanbinme:
@wilkinsona

Thank you for the feedback! You're absolutely right about adding a test to demonstrate the NPE issue.

I'll add a test case following the existing naming convention, something like:
`buildWhenDerivedFromSimpleDriverDataSourceWithDriverNotSetSucceeds()`

While writing the test, I discovered an additional NPE issue in the `convertToString` method where null values were being passed, causing `NullPointerException` when trying to call `.getName()` on a null Class object. I've addressed this as well by adding proper null checks.

snicoll:
Thanks @chanbinme

chanbinme:
@snicoll 
Thank you for the guidance and for merging this PR! It was a great learning experience working on this issue. Looking forward to contributing more to Spring Boot in the future. 😊

Issue #46013, opened 2025-06-18
spring.netty.leak-detection has no effect when lazy initialization is enabled


mhalbritter:
Like https://github.com/spring-projects/spring-boot/issues/45846, when enabling lazy init, Netty's `ResourceLeakDetector` won't be enabled. It's set up in the constructor of `NettyAutoConfiguration`, which won't run when lazy init is enabled.

nosan:
I've made some changes to my branch https://github.com/spring-projects/spring-boot/compare/3.4.x...nosan:spring-boot:46013

mhalbritter:
Nice. Could you please create a PR for that? Thanks!

mhalbritter:
Superseded by https://github.com/spring-projects/spring-boot/pull/46164.

PR #46164, opened 2025-06-23
Setting spring.netty.leak-detection has no effect when lazy initialization is enabled

nosan:
See #46013


mhalbritter:
Thanks @nosan !

nosan:
Thanks @mhalbritter

PR #46166, opened 2025-06-23
Skip registering all OTLP tracing related beans if tracing is disabled

quaff:
Before this commit, `OtlpTracingProperties` and `OtlpTracingConnectionDetails` are registered even if tracing is disabled.


quaff:
It applies to other tracing `AutoConfiguration`s such as `ZipkinAutoConfiguration`, should I create separate PR or append commits in this PR?

mhalbritter:
It's intended that `OtlpTracingProperties` and `OtlpTracingConnectionDetails` are registered. The `@ConditionalOnEnabledTracing` is intended to shut off tracing export (it should really be named `@ConditionalOnEnabledTracingExport` or something like that), not all tracing related beans.

Issue #46061, opened 2025-06-20
Review upgrade to Elasticsearch 9


snicoll:
The upgrade made it so that `RestClient` isn't automatically on the classpath.  However `ElasticsearchClientAutoConfiguration` is guarded by the fact that such a bean has been configured. We'd need to see if there is an alternative option that doesn't need the extra library.

Removing `org.elasticsearch.client:elasticsearch-rest-client` from `spring-boot-testcontainers` can reproduce that issue.

We also moved our test altogether to ES9 server. Perhaps we still want to test against ES8, although it needs additional settings to work, see https://github.com/elastic/elasticsearch-ruby/issues/2665

wilkinsona:
[`Rest5Client`](https://www.elastic.co/docs/reference/elasticsearch/clients/java/transport/rest5-client/?v=EF5DFA09) appears to be the replacement. It's part of `elasticsearch-java` so we already have it on the classpath. We should remove `org.elasticsearch.client:elasticsearch-rest-client` from `spring-boot-starter-data-elasticsearch` as part of switching to the new client.

wilkinsona:
> We also moved our test altogether to ES9 server

Tests that use `TestImage.container(ElasticsearchContainer.class)` are still using Elasticsearch 7.

wilkinsona:
I have [something](https://github.com/wilkinsona/spring-boot/tree/rest5client) for this, but I'd like to wait and see if there are any [improvements to the configurability of the new client](https://github.com/elastic/elasticsearch-java/issues/1018#issuecomment-3032850004) before proceeding.

wilkinsona:
Thanks to @l-trotta, the client is now more configurable and it should be easier for us to preserve its defaults while also offering the customisation that we had before.

wilkinsona:
I've updated [my branch](https://github.com/wilkinsona/spring-boot/tree/rest5client) to use the new callbacks and it's significantly simpler. We now need to wait for Spring Data Elasticsearch to upgrade to the latest patch release of the client where the new callbacks have been added.

sothawo:
@wilkinsona I updated Spring Data Elasticsearch this morning to use EL 9.0.4, so the dependency for you should be there. I still need to adjust my code to use/provide these callbacks to Spring Data Elasticsearch users, but that should not block you. Anyway I think I'll do this this weekend.

wilkinsona:
[...]

PR #46167, opened 2025-06-23
Allow APIKey to be used as authentication method for Elasticsearch

l-trotta:
APIKey is a popular authentication method for elasticsearch, available both for local installations and for cloud instances; in fact, it's the [recommended way to setup](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/8.18/connecting.html) the Elasticsearch Java client!

This PR adds `spring.elasticsearch.apikey` as one of the elasticsearch properties, so that users will be able to setup a connection simply with:
```
spring.elasticsearch.uris=${ES_SERVER_URL}
spring.elasticsearch.apikey=${ES_APIKEY}
```
If this gets merged I will also take care of the documentation of affected repos ([spring-ai](https://github.com/spring-projects/spring-ai), [spring-data-elasticsearch](https://github.com/spring-projects/spring-data-elasticsearch)) to add and describe the new property.

Disclaimer: I am one of the maintainer of https://github.com/elastic/elasticsearch-java

### DIFF
--- a/module/spring-boot-elasticsearch/src/main/java/org/springframework/boot/elasticsearch/autoconfigure/ElasticsearchConnectionDetails.java
+++ b/module/spring-boot-elasticsearch/src/main/java/org/springframework/boot/elasticsearch/autoconfigure/ElasticsearchConnectionDetails.java
@@ -58,6 +58,14 @@ public interface ElasticsearchConnectionDetails extends ConnectionDetails {
 	}
 
 	/**
+	 * APIKey for authentication with Elasticsearch.
+	 * @return APIKey for authentication with Elasticsearch or {@code null}
+	 */
+	default @Nullable String getAPIKey() {
+		return null;
+	}
+
+	/**
 	 * Prefix added to the path of every request sent to Elasticsearch.
 	 * @return prefix added to the path of every request sent to Elasticsearch or
 	 * {@code null}
--- a/module/spring-boot-elasticsearch/src/main/java/org/springframework/boot/elasticsearch/autoconfigure/ElasticsearchProperties.java
+++ b/module/spring-boot-elasticsearch/src/main/java/org/springframework/boot/elasticsearch/autoconfigure/ElasticsearchProperties.java
@@ -48,6 +48,10 @@ public class ElasticsearchProperties {
 	 * Password for authentication with Elasticsearch.
 	 */
 	private @Nullable String password;
+	/**
+	 * APIKey for authentication with Elasticsearch.
+	 */
+	private @Nullable String APIKey;
 
 	/**
 	 * Connection timeout used when communicating with Elasticsearch.
@@ -93,6 +97,14 @@ public class ElasticsearchProperties {
 
 	public void setPassword(@Nullable String password) {
 		this.password = password;
+	}
+
+	public @Nullable String getAPIKey() {
+		return this.APIKey;
+	}
+
+	public void setAPIKey(@Nullable String APIKey) {
+		this.APIKey = APIKey;
 	}
 
 	public Duration getConnectionTimeout() {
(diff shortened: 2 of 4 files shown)


snicoll:
Thanks for the PR

> it's the [recommended way to setup](https://www.elastic.co/guide/en/elasticsearch/client/java-api-client/8.18/connecting.html) the Elasticsearch Java client!

I wasn't aware of that. How does it work with Testcontainers and Docker compose then? Both of them retrieve a password from the environment. I can see you have updated `ElasticsearchConnectionDetails` but none of their two implementations.

l-trotta:
Creating the API key has to be done through either an [API call](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-security-create-api-key) or in the [dedicated Kibana menu](https://www.elastic.co/docs/deploy-manage/api-keys/elasticsearch-api-keys), this is because you can decide to create one with certain privileges, or with an expiration date. So the default for both implementations of `ElasticsearchConnectionDetails` would be null, because the API key cannot be retreived from the environment, but only created after the elasticsearch instance is running.

But it works the exact same for cloud, testcontainers or docker, it's just an additional step unlike username and password credentials.

ezimuel:
@l-trotta and @snicoll maybe we can simplify the Elasticsearch testing using [start-local](https://github.com/elastic/start-local). This script installs Elasticsearch with a `.env` file containing all the information for the connection, including the API key (`ES_LOCAL_API_KEY`). If you only need Elasticsearch and not Kibana you can use the [--esonly](https://github.com/elastic/start-local) option.

eddumelendez:
Hi, what about providing and env var in Elasticsearch image which allows to configure the API Key? That would benefit Docker Compose and Testcontainers implementations. Also, if this is the recommended way then the image could provide a default way to enable this.

On the Testcontainers side we can add a custom implementation to generate the API Key but I am more in favor to provide the same experience for Docker Compose and Testcontainers users by just enabling the API key with a env var.

bclozel:
@eddumelendez Looking at what @l-trotta pointed out, it seems that one of the [main driver behind API keys is to have fine-grained roles per index](https://www.elastic.co/docs/api/doc/elasticsearch/operation/operation-security-create-api-key), since you need to provide those during API key generation. I guess that for a given user, you want to use different keys for specific usage patterns and restrict roles accordingly.
Is this something that testcontainers is already supporting in some way, or is it something that developers should write as part of their tests?

eddumelendez:
Hi @bclozel, currently, developers must extend and override `ElasticsearchContainer`, implement `containerIsStarted` as part of the Testcontainers Lifecycle and make a similar call to [start-local](https://github.com/elastic/start-local/blob/main/start-local.sh#L316) using a HTTP library.

eddumelendez:
I've uploaded an [example](https://github.com/eddumelendez/testcontainers-samples/blob/main/spring-boot-elasticsearch/src/test/java/com/example/springbootelasticsearch/BookRepositoryElasticsearch8ApiKeyTests.java#L35-L55) using Testcontainers and Elasticsearch using API Key.

wilkinsona:
I've placed this on hold for now as I'd like to tackle [the migration to `Rest5Client`](https://github.com/spring-projects/spring-boot/issues/46061) before we add any more configuration options.

ezimuel:
@wilkinsona do you have any update on the migration to Rest5Client? The PR seems to be merged. Thanks.

wilkinsona:
Yes, that work's complete now. Next step would be to revisit the proposal here on top of that new base.

ezimuel:
Thanks @wilkinsona

snicoll:
@ezimuel I don't think you mean to close this PR, isn't it? That happens when you create it from main and you force push which would close the PR. It's shame as we'd have liked to keep the conversation history...

l-trotta:
@snicoll I'm working on it! will push the new code ASAP and reopen it

wilkinsona:
Thanks for the updates, @l-trotta. What do you think should happen if the user configures both an API key and a username and password? Should we:

- Configure the client with only the API key?
- Configure the client with only the username and password?
- Configure the client with both?
- Fail?
- Something else?

l-trotta:
Hey @wilkinsona! Currently the java client builder helper (`ElasticsearchClient.of()`) fails with `java.lang.IllegalArgumentException: Cannot set both username/password and API key`, but there are no checks for the extended builder, which Spring uses. Should we apply the same logic here so that it's coherent? Sending both Basic and ApiKey header to the server won't result in an error, the first valid one will be used.

wilkinsona:
Thanks, @l-trotta.

Once a property has been set in Spring Boot, it can be a bit harder than we'd like to then unset it (https://github.com/spring-projects/spring-boot/issues/24133). If we made the API key and the username and password mutually exclusive this could make it hard for a user to override one with the other. For this reason, I think we should leave things as they are and allow both to be configured. We can always refine this in the future if it proves to be problematic.

l-trotta:
good to know @wilkinsona! everything's ready then

wilkinsona:
Thanks very much, @l-trotta.

PR #46208, opened 2025-06-25
Add FONTS location to StaticResourceLocation

shahabkondri:
### Motivation
`StaticResourceLocation` handles common static assets (`CSS`, `JS`, `IMAGES`, …) but omits web-fonts.  
As a result, `PathRequest.toStaticResources()` does **not** match `/fonts/**`, forcing users to add custom ant matchers.  
Adding a dedicated `FONTS` location ("/fonts/**") removes that gap.

### Implementation
* New enum constant `FONTS("/fonts/**")` in `StaticResourceLocation`.
_No breaking changes; existing constants keep the same semantics._

### Tests
* Unit test verifies that `StaticResourceLocation ` now matches a `/fonts/font.woff2` request.


wilkinsona:
We discussed this again today. The issue is noteworthy as we want to call out that this change will affect access to `/fonts/**` for anyone using `atCommonLocations()` on `StaticResourceRequestMatcher`.

Issue #46209, opened 2025-06-25
Migrate SpringBeanContainer to org.springframework.orm.jpa.hibernate package for Hibernate ORM 7.0


lesiak:
In https://github.com/spring-projects/spring-framework/issues/35111 SpringBeanContainer changed location from `org.springframework.orm.hibernate5.SpringBeanContainer` to `org.springframework.orm.jpa.hibernate.SpringBeanContainer`

HibernateJpaConfiguration needs to be updated accordingly.

wilkinsona:
Superseded by #46210.

@lesiak there's no need to open an issue if you intend to open a PR as well.

PR #46210, opened 2025-06-25
Migrate SpringBeanContainer to org.springframework.orm.jpa.hibernate package for Hibernate ORM 7.0

lesiak:
Fixes https://github.com/spring-projects/spring-boot/issues/46209

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaConfiguration.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/orm/jpa/HibernateJpaConfiguration.java
@@ -55,7 +55,7 @@ import org.springframework.context.annotation.Configuration;
 import org.springframework.context.annotation.ImportRuntimeHints;
 import org.springframework.jdbc.support.SQLExceptionTranslator;
 import org.springframework.jndi.JndiLocatorDelegate;
-import org.springframework.orm.hibernate5.SpringBeanContainer;
+import org.springframework.orm.jpa.hibernate.SpringBeanContainer;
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thanks for the PR, but there's more that needs to be done to adjust to the changes to Framework's Hibernate support. There's also a problem with persistence exception translation that we're investigating at the moment with the Framework team.

lesiak:
@wilkinsona I completely understand that more changes may be required to support Hibernate 7, but currently `spring-boot-autoconfigure` module does not even compile due to upstream change in https://github.com/spring-projects/spring-framework/issues/35111.

bclozel:
@lesiak the relevant changes are already on the main branch. I see now that this PR was submitted against 4.0.x.

lesiak:
Indeed, it seems https://github.com/spring-projects/spring-boot/commit/19818fb844c66cd4fc0e2e45581335e22aba1ce2 was added with identical change after rejecting this PR.

bclozel:
I don't think so.

lesiak:
I guess we both have better things to do than comparing timestamps (but I'm pretty sure about the order of events)

Hope you'll merge this change to 4.x soon.

wilkinsona:
@lesiak FWIW, I hadn't seen this PR when I began working on the changes in response to a heads-up from the Framework team. I only noticed it once I had identified the problem with persistence exception translation that needed to be fixed on the Framework side. By that point, it was apparent that the change proposed here wasn't sufficient which is when I made [the comment above](https://github.com/spring-projects/spring-boot/pull/46210#issuecomment-3006052293). 

> Hope you'll merge this change to 4.x soon.

There's no need to do that. The `4.0.x` branch was temporary and has now been deleted. `main` is producing `4.0.0-SNAPSHOT` builds and contains the necessary changes already.

lesiak:
@wilkinsona Thanks for the clarification

Issue #46026, opened 2025-06-18
Only enable tests that require an English locale when the locale is English


LinkDownZ:
[					(project) -> assertThat(buildLog(project)).contains("invalid flag: --invalid-compiler-arg"));](https://github.com/spring-projects/spring-boot/blob/77877b4c69a03cd92dab3ba33609142475d52b20/spring-boot-project/spring-boot-tools/spring-boot-maven-plugin/src/intTest/java/org/springframework/boot/maven/AotTests.java#L132) "Is it possible to remove the 'invalid flag:' log? The judgment may be incorrect in non-English locales."

wilkinsona:
I think we have some other tests that also assume an English locale, [such as those that expect a `PortInUseException` to be thrown](https://github.com/spring-projects/spring-boot/issues/29918).

How do you deal with this @bclozel, @mhalbritter, and @snicoll?

bclozel:
I'm using an English locale...

mhalbritter:
I've ran our `main` branch with a `de_DE` locale, and got this build scan:

https://ge.spring.io/s/6ewqquhhxvln4

When looking at [the failures](https://ge.spring.io/s/6ewqquhhxvln4/tests/overview?outcome=FAILED):

* 2x times this problem
* 6x #29918
* 1x Elasticsearch which expected "Connection refused" but got the German equivalent

wilkinsona:
We discussed this today and decided that we're going to introduce `@EnabledOnLocale` in `spring-boot-test-support` and annotate the affected tests so that they only run when the locale's language is English – perhaps something like `@EnabledOnLocale(language = "en")`.

snicoll:
Closing in favor of PR #46211

PR #46211, opened 2025-06-25
Only enable tests that require an English locale when the locale is English

nosan:
#46026

### DIFF
--- a/spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/junit/EnabledOnLocaleCondition.java
+++ b/spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/junit/EnabledOnLocaleCondition.java
@@ -1,0 +1,50 @@
+/*
+ * Copyright 2012-present the original author or authors.
+ *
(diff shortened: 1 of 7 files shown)


spring-boot-project/spring-boot-tools/spring-boot-test-support/src/main/java/org/springframework/boot/testsupport/junit/EnabledOnLocaleCondition.java
```diff
@@ -0,0 +1,50 @@
...
+	@Override
+	public ConditionEvaluationResult evaluateExecutionCondition(ExtensionContext context) {
+		return AnnotationSupport.findAnnotation(context.getElement(), EnabledOnLocale.class)
+			.map(this::evaluate)
+			.orElseGet(() -> ConditionEvaluationResult.enabled("No @EnabledOnLocale annotation found"));
+	}
+
+	private ConditionEvaluationResult evaluate(EnabledOnLocale annotation) {
+		Locale locale = Locale.getDefault();
+		String language = locale.getLanguage();
```
nosan, line 43: As an alternative, `System.getProperty("user.language")` can be used, but as far as I remember, `Locale.getDefault()` returns the same value.

PR #46216, opened 2025-06-26
Migrate to Spring Batch 6

fmbenhassine:
This PR upgrades Spring Boot 4 to Spring Batch 6.

Most of the changes are documented here: https://github.com/spring-projects/spring-batch/wiki/Spring-Batch-6.0-Migration-Guide

### DIFF
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchConversionServiceCustomizer.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchConversionServiceCustomizer.java
@@ -16,14 +16,12 @@
 
 package org.springframework.boot.autoconfigure.batch;
 
-import org.springframework.batch.core.configuration.support.JdbcDefaultBatchConfiguration;
 import org.springframework.core.convert.support.ConfigurableConversionService;
 
 /**
  * Callback interface that can be implemented by beans wishing to customize the
- * {@link ConfigurableConversionService} that is
- * {@link JdbcDefaultBatchConfiguration#getConversionService provided by
- * DefaultBatchConfiguration} while retaining its default auto-configuration.
+ * {@link ConfigurableConversionService} that is used by the batch infrastructure
+ * while retaining its default auto-configuration.
  *
  * @author Claudio Nave
  * @since 3.1.0
--- a/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
@@ -209,7 +209,6 @@ public class JobLauncherApplicationRunner
 		}
 	}
 
-	@SuppressWarnings("deprecated")
 	private JobParameters getNextJobParameters(Job job, JobParameters jobParameters) {
 		if (this.jobRepository != null && this.jobRepository.getJobInstance(job.getName(), jobParameters) != null) {
 			return getNextJobParametersForExisting(job, jobParameters);
@@ -223,7 +222,6 @@ public class JobLauncherApplicationRunner
 		return merge(nextParameters, jobParameters);
 	}
 
-	@SuppressWarnings("deprecated")
 	private JobParameters getNextJobParametersForExisting(Job job, JobParameters jobParameters) {
 		JobExecution lastExecution = this.jobRepository.getLastJobExecution(job.getName(), jobParameters);
 		if (isStoppedOrFailed(lastExecution) && job.isRestartable()) {
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java
@@ -85,6 +85,7 @@ import org.springframework.jdbc.core.JdbcTemplate;
 import org.springframework.jdbc.datasource.DataSourceTransactionManager;
 import org.springframework.jdbc.datasource.init.DatabasePopulator;
 import org.springframework.orm.jpa.JpaTransactionManager;
+import org.springframework.test.util.AopTestUtils;
 import org.springframework.transaction.PlatformTransactionManager;
 import org.springframework.transaction.annotation.Isolation;
 
@@ -107,7 +108,6 @@ import static org.mockito.Mockito.mock;
  * @author Yanming Zhou
  */
 @ExtendWith(OutputCaptureExtension.class)
-@SuppressWarnings("removal")
 class BatchAutoConfigurationTests {
 
 	private final ApplicationContextRunner contextRunner = new ApplicationContextRunner().withConfiguration(
@@ -355,10 +355,8 @@ class BatchAutoConfigurationTests {
 				assertThat(batchTaskExecutor).isInstanceOf(AsyncTaskExecutor.class);
 				assertThat(context.getBean(SpringBootBatchConfiguration.class).getTaskExecutor())
 					.isEqualTo(batchTaskExecutor);
-				// the following assertion should pass as taskExecutor is inherited from
-				// TaskExecutorJobLauncher
-				// assertThat(context.getBean(JobOperator.class)).hasFieldOrPropertyWithValue("taskExecutor",
-				// batchTaskExecutor);
+				JobOperator jobOperator = AopTestUtils.getTargetObject(context.getBean(JobOperator.class));
+				assertThat(jobOperator).hasFieldOrPropertyWithValue("taskExecutor", batchTaskExecutor);
 			});
 	}
 
--- a/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunnerTests.java
+++ b/spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunnerTests.java
@@ -203,7 +203,6 @@ class JobLauncherApplicationRunnerTests {
 			this.runner = new JobLauncherApplicationRunner(jobOperator, jobRepository);
 		}
 
-		@SuppressWarnings("removal")
 		List<JobInstance> jobInstances() {
 			return this.jobRepository.getJobInstances("job", 0, 100);
 		}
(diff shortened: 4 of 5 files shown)


spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
```diff
@@ -172,6 +172,7 @@ protected ExecutionContextSerializer getExecutionContextSerializer() {
 					: super.getExecutionContextSerializer();
 		}
 
+		@SuppressWarnings("removal")
```
snicoll, line 175: I am not sure what to make of this. The method is deprecated but the class isn't. If we don't want this to be configurable going forward, I'd rather remove the override (and the `ObjectProvider`)

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchConversionServiceCustomizer.java
```diff
@@ -16,13 +16,13 @@
...
 
-import org.springframework.batch.core.configuration.support.DefaultBatchConfiguration;
+import org.springframework.batch.core.configuration.support.JdbcDefaultBatchConfiguration;
 import org.springframework.core.convert.support.ConfigurableConversionService;
 
 /**
  * Callback interface that can be implemented by beans wishing to customize the
  * {@link ConfigurableConversionService} that is
- * {@link DefaultBatchConfiguration#getConversionService provided by
+ * {@link JdbcDefaultBatchConfiguration#getConversionService provided by
```
snicoll: I think we should relax the javadoc a bit. It looks weird that we refer to that method. Something like

> Callback interface that can be implemented by beans wishing to customize the ConfigurableConversionService that is use by the batch infrastructure while retaining its default auto-configuration.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
```diff
@@ -204,28 +199,31 @@ private void executeRegisteredJobs(JobParameters jobParameters) throws JobExecut
...
+		JobExecution execution = this.jobOperator.start(job, parameters);
 		if (this.publisher != null) {
 			this.publisher.publishEvent(new JobExecutionEvent(execution));
 		}
 	}
 
+	@SuppressWarnings("deprecated")
 	private JobParameters getNextJobParameters(Job job, JobParameters jobParameters) {
-		if (this.jobRepository != null && this.jobRepository.isJobInstanceExists(job.getName(), jobParameters)) {
+		if (this.jobRepository != null && this.jobRepository.getJobInstance(job.getName(), jobParameters) != null) {
```
snicoll, line 213: I think something should be adapted in Spring Batch here. `JobRepository` isn't deprecated but `JobExplorer` is. The former extends the latter so any method that is not specified in the child is flagged as deprecated.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
```diff
@@ -204,28 +199,31 @@ private void executeRegisteredJobs(JobParameters jobParameters) throws JobExecut
...
 			return jobParameters;
 		}
-		JobParameters nextParameters = new JobParametersBuilder(jobParameters, this.jobExplorer)
+		JobParameters nextParameters = new JobParametersBuilder(jobParameters, this.jobRepository)
 			.getNextJobParameters(job)
 			.toJobParameters();
 		return merge(nextParameters, jobParameters);
 	}
 
+	@SuppressWarnings("deprecated")
```
snicoll: Not sure I see what this suppresses.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java
```diff
@@ -114,6 +107,7 @@
  * @author Yanming Zhou
  */
 @ExtendWith(OutputCaptureExtension.class)
+@SuppressWarnings("removal")
```
snicoll: What's the plan to deal with those deprecations?

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java
```diff
@@ -376,8 +355,10 @@ void testBatchTaskExecutor() {
 				assertThat(batchTaskExecutor).isInstanceOf(AsyncTaskExecutor.class);
 				assertThat(context.getBean(SpringBootBatchConfiguration.class).getTaskExecutor())
 					.isEqualTo(batchTaskExecutor);
-				assertThat(context.getBean(JobLauncher.class)).hasFieldOrPropertyWithValue("taskExecutor",
-						batchTaskExecutor);
+				// the following assertion should pass as taskExecutor is inherited from
+				// TaskExecutorJobLauncher
+				// assertThat(context.getBean(JobOperator.class)).hasFieldOrPropertyWithValue("taskExecutor",
+				// batchTaskExecutor);
```
snicoll: Please don't comment code like that.

The assertion does not pass because `context.getBean(JobOperator.class)` returns a JDK Proxy. And that proxy does not have such field. We should figure out why it returns a proxy and it didn't before.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunnerTests.java
```diff
@@ -226,6 +227,7 @@ SimpleJobBuilder configureJob() {
 	}
 
 	@EnableBatchProcessing
+	@EnableJdbcJobRepository
```
snicoll, line 229: Did you consider adding an attribute to `@EnableBatchProcessing` Looking at the code sample there, it looks odd to me and the second annotation does not make it very obvious it's Spring Batch specific.

snicoll (changes requested):
Thanks for the PR. There are a number of things we need to fix before going forward. I've added comments.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunnerTests.java
snicoll: Now I understand it's a bit similar to Spring Session, we can ignore this. I also missed "Job" from the annotation name.

snicoll: Let's ignore this for now. I understood after the intention and missed the term `Job` for some reason.

snicoll:
There a number of things that should happen in Spring Batch before we can make progress here, notably a status on `JobParametersConverter` (https://github.com/spring-projects/spring-batch/issues/4894) and methods that are deprecated while they aren't (due to inheritance and the lack of an override in the interface that isn't deprecated).

fmbenhassine:
Thanks for the review.

> notably a status on `JobParametersConverter` ([spring-projects/spring-batch#4894](https://github.com/spring-projects/spring-batch/issues/4894))

The deprecation was reverted.

> methods that are deprecated while they aren't (due to inheritance and the lack of an override in the interface that isn't deprecated).

This was fixed in Spring Batch.

I will add some inline comments as well.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchConversionServiceCustomizer.java
fmbenhassine: Done

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
fmbenhassine: Spring Batch was updated. I have removed the `@SuppressWarning("removed")`

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunner.java
fmbenhassine: This was meant to suppress the warning about the call of the deprecated method `jobRepository.getLastJobExecution`. But this is now fixed.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java
fmbenhassine: Fixed in Spring Batch.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfigurationTests.java
fmbenhassine: I updated the assertion.

spring-boot-project/spring-boot-autoconfigure/src/test/java/org/springframework/boot/autoconfigure/batch/JobLauncherApplicationRunnerTests.java
fmbenhassine: ok no problem.

spring-boot-project/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/batch/BatchAutoConfiguration.java
fmbenhassine: This could still be configurable until the method is removed in Spring Batch 6.2. Maybe this configuration option should be documented as deprecated in Spring Boot docs? Wdyt?

snicoll: Ah I see it now. It's a support for the method that you've deprecated. I think you might have missed another deprecation in `JobOperatorFactoryBean#setJobParametersConverter`

PR #43886, opened 2025-01-21
Add spring.validation.method.adapt-constraint-violations property

quaff:
Closes GH-43881

PR #46220, opened 2025-06-27
Move "*.enabled" keys from metadata to dedicated `@ConfigurationProperties` bean

quaff:
IIRC, this is the recommended way.


snicoll:
> IIRC, this is the recommended way.

What makes you think that?

philwebb:
I don't think we really have a single way to deal with enabled properties, and I'm not really sure that we should. I think there's at least one or two that only in the metadata because we don't have a `@ConfigurationProperties` class for them.

All things considered, I think we should leave things as they are for now.

Thanks anyway for the PR.

quaff:
> > IIRC, this is the recommended way.
> 
> What makes you think that?


See https://github.com/spring-projects/spring-boot/pull/43886#discussion_r1923395581

@wilkinsona prefer creating new `@ConfigurationProperties` bean over hand-written metadata, this PR reuse existing `@ConfigurationProperties` bean.

wilkinsona:
That was a preference in a specific situation rather than a rule. Thanks again for the PR but, generally speaking, I think it's better to leave the introduction of stylistic changes that cut across the whole codebase to the core team.

Issue #41227, opened 2024-06-25
Explain difference between OTel agent and Micrometer instrumentations


marcingrzejszczak:
The suggested approach is to use Micrometer Observation bridge since all Spring projects are using that for instrumentation. One can then use Micrometer OTLP Meter Registry to be OTLP compatible. For Tracing, Micrometer Tracing is suggested with either Brave ([once this projects is released in GA](https://github.com/openzipkin-contrib/zipkin-otel)) or OTel bridge, in order to be OTLP compliant.

Usage of Agent is discouraged due to various restrictions:

- [OpenTelemetry Java Instrumentation](https://github.com/open-telemetry/opentelemetry-java-instrumentation/) project contains multiple `alpha` JARs (e.g. check the [release notes](https://github.com/open-telemetry/opentelemetry-java-instrumentation/releases/tag/v2.5.0))
- You can't create a native image when using the agent
- Spring Boot OTel configuration is maintained by the Spring Boot team, other auto configurations may not be compatible with Spring Boot supported solutions
- Agent instrumentation is buggy (e.g. [issue](https://github.com/spring-projects/spring-boot/issues/33921))

bclozel:
Thanks Marcin for this suggestion. Rather than trying to compare technologies (we don't usually do that in our reference docs), it would make sense to clear up some points that could confuse developers:

* If your application is using the `Observation` API, I guess produced metrics and traces won't be exported if you're using the OTel agent?
* If you are using the micrometer bridge, the observation metadata described in Spring reference docs apply (see [Spring for GraphQL](https://docs.spring.io/spring-graphql/reference/observability.html) or [Spring Framework](https://docs.spring.io/spring-framework/reference/integration/observability.html)). If you're using the OTel agent, [its semantics apply instead](https://github.com/open-telemetry/opentelemetry-java-instrumentation/blob/main/docs/supported-libraries.md#libraries--frameworks)
* The Spring team does not provide support for the OTel agent in general, you should reach out to the OTel community

Spring projects could also consider echoing this by saying that the metrics/traces described in their reference docs only apply if Micrometer is used, not if a third party instrumentation agent is used.

famaridon:
Hello,

It's currently quite challenging to understand how to configure Spring Boot with OpenTelemetry, as there are three different solutions available:

1. **OpenTelemetry JavaAgent**
2. **OpenTelemetry Spring Boot Starter**
3. **Spring Boot Configuration (Micrometer OTel Bridge)**

[...]

PR #46230, opened 2025-06-30
Allow reusing registered global OpenTelemetry

quaff:
Spring Boot instrumentation will be included when applications are running with OpenTelemetry Java Agent.

### DIFF
--- a/spring-boot-project/spring-boot-opentelemetry/src/main/java/org/springframework/boot/opentelemetry/autoconfigure/OpenTelemetryProperties.java
+++ b/spring-boot-project/spring-boot-opentelemetry/src/main/java/org/springframework/boot/opentelemetry/autoconfigure/OpenTelemetryProperties.java
@@ -25,15 +25,29 @@ import org.springframework.boot.context.properties.ConfigurationProperties;
  * Configuration properties for OpenTelemetry.
  *
  * @author Moritz Halbritter
+ * @author Yanming Zhou
  * @since 4.0.0
  */
 @ConfigurationProperties("management.opentelemetry")
 public class OpenTelemetryProperties {
 
 	/**
+	 * Whether to reuse registered global OpenTelemetry.
+	 */
+	private boolean reuseRegisteredGlobal;
+
+	/**
 	 * Resource attributes.
 	 */
 	private Map<String, String> resourceAttributes = new HashMap<>();
+
+	public boolean isReuseRegisteredGlobal() {
+		return this.reuseRegisteredGlobal;
+	}
+
+	public void setReuseRegisteredGlobal(boolean reuseRegisteredGlobal) {
+		this.reuseRegisteredGlobal = reuseRegisteredGlobal;
+	}
 
 	public Map<String, String> getResourceAttributes() {
 		return this.resourceAttributes;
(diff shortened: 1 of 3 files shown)


spring-boot-project/spring-boot-actuator-autoconfigure/src/main/java/org/springframework/boot/actuate/autoconfigure/opentelemetry/OpenTelemetryAutoConfiguration.java
```diff
@@ -29,23 +30,37 @@
...
  * @since 3.2.0
  */
 @AutoConfiguration
 @ConditionalOnClass(OpenTelemetrySdk.class)
 @EnableConfigurationProperties(OpenTelemetryProperties.class)
 public class OpenTelemetryAutoConfiguration {
 
+	@Bean
+	@ConditionalOnMissingBean(OpenTelemetry.class)
+	@Conditional(OnRegisteredGlobalOpenTelemetryCondition.class)
```
quaff: Not sure reuse it by default or a boolean configuration property should be introduced.

bclozel:
I don't think this is a sane goal. It is already complicated for Spring Boot users to figure out the available options.

We have tried to clarify things in https://github.com/spring-projects/spring-boot/issues/41227 (even if the reverse is not true) but if I understand correctly, this change will cause duplicate metrics and will disrupt OTel Java agent users.

quaff:
> this change will cause duplicate metrics and will disrupt OTel Java agent users.

I think it's necessary to introduce a flag to allow users reuse it.

EDIT: I've updated the PR to introduce configuration key `management.opentelemetry.reuse-registered-global`.

wilkinsona:
Adding a flag doesn't really address Brian's point. We don't want users to be able to opt into having duplicate metrics. I'm going to close this one as the downsides outweigh any potential benefit. If we've misunderstood the implications of the change then please let us know and we can take another look.

quaff:
> Adding a flag doesn't really address Brian's point. We don't want users to be able to opt into having duplicate metrics. I'm going to close this one as the downsides outweigh any potential benefit. If we've misunderstood the implications of the change then please let us know and we can take another look.

FYI, I'm using OpenTelemetry Java Agent instrument traces only, and didn't configure `management.otlp.tracing.endpoint` to export spans at Spring Boot side, spans collected by Spring Boot will not be included without this change.
Of course we could reuse global OpenTelemetry in our own applications, but it's good if Spring Boot autoconfigure that, maybe I should add another condition to check OpenTelemetry Java Agent is not used for metrics (`OTEL_METRICS_EXPORTER=none`)?

quaff:
Another option, we could reuse global registered OpenTelemetry only for `Tracer` if `management.otlp.tracing.endpoint` doesn't configured:

https://github.com/spring-projects/spring-boot/blob/5db3cacde833e9e2652a61f36570c017aad4a789/spring-boot-project/spring-boot-tracing/src/main/java/org/springframework/boot/tracing/autoconfigure/OpenTelemetryTracingAutoConfiguration.java#L142-L146

to

```java
	@Bean
	@ConditionalOnMissingBean
	Tracer otelTracer(OpenTelemetry openTelemetry, Optional<OtlpTracingConnectionDetails> otlpTracingConnectionDetails) {
		OpenTelemetry otel = openTelemetry;
		if (GlobalOpenTelemetry.get() != OpenTelemetry.noop() && otlpTracingConnectionDetails.isEmpty()) {
			otel = GlobalOpenTelemetry.get();
		}
		return otel.getTracer("org.springframework.boot", SpringBootVersion.getVersion());
	}
```

PR #46238, opened 2025-07-01
Upgrade to Kotlin 2.2.0

sdeleuze:
This commit leverages Kotlin 2.2 which is expected to be the new baseline for Spring Boot 4.0, so API and language versions are upgraded to `KOTLIN_2_2` as well.

It also enables the `-Xannotation-default-target=param-property` compiler flag, mainly for test purpose, in order to avoid the related warnings and to use what will likely be the default in an upcoming Kotlin version short term. For more details see https://youtrack.jetbrains.com/issue/KT-73255.

`KotlinPlatformJvmPlugin` was a class from an old KMP plugin, deprecated for a long time and now removed. So its usage in `PluginClasspathGradleBuild` has been removed as well. @wilkinsona I **think** I don't see side effects after removing the offending line but please double check as I am not able to run some tests locally.


wilkinsona:
Thanks, @sdeleuze. Would you mind taking a look at https://github.com/spring-projects/spring-boot/blob/5db3cacde833e9e2652a61f36570c017aad4a789/spring-boot-project/spring-boot-docs/src/docs/antora/modules/reference/pages/features/kotlin.adoc?plain=1#L15-L28 in the context of Kotlin 2.2 please? We can update the 2.1.x to 2.2.x for the minimum requirements, but I'm also wondering if there any other changes to be made there.

sdeleuze:
Sure I will have a look and refine the documentation accordingly.

sdeleuze:
Documentation update pushed.

wilkinsona:
Thanks very much, @sdeleuze.

PR #46239, opened 2025-07-02
Remove unnecessary constructor to align with other AutoConfigurations

quaff:


philwebb:
Those are actually intentional and we plan to adopt them more widely (but we're not ready to do that yet so please don't submit a PR :) )

The idea is that the class is public so that it can be excluded, but the constructor is package private to stop folks using or extending them directly because we consider auto-configuration classes to not be public API.

quaff:
@philwebb Is it possible to generate package private constructor by annotation processor?

philwebb:
I think we'd rather just keep it simple and add them manually.

wilkinsona:
I thought we'd decided on making them final rather than adding package-private constructors, hence the update to the title of https://github.com/spring-projects/spring-boot/issues/32883.

philwebb:
I can't remember where we landed. Using `final` would stop subclassing, using a package private constructor would stop folks delegating. Perhaps it's a step too far.

PR #46240, opened 2025-07-02
Fix dependencies

quaff:
Both `IntegrationDataSourceScriptDatabaseInitializer` and `QuartzDataSourceScriptDatabaseInitializer` extend `DataSourceScriptDatabaseInitializer` from `spring-boot-jdbc` which should be required dependency.

### DIFF
--- a/spring-boot-project/spring-boot-integration/build.gradle
+++ b/spring-boot-project/spring-boot-integration/build.gradle
@@ -29,11 +29,10 @@ dependencies {
 	api(project(":spring-boot-project:spring-boot"))
 	api("org.springframework.integration:spring-integration-core")
 
-	implementation(project(":spring-boot-project:spring-boot-sql"))
+	implementation(project(":spring-boot-project:spring-boot-jdbc"))
 
 	optional(project(":spring-boot-project:spring-boot-actuator-autoconfigure"))
 	optional(project(":spring-boot-project:spring-boot-autoconfigure"))
-	optional(project(":spring-boot-project:spring-boot-jdbc"))
 	optional(project(":spring-boot-project:spring-boot-metrics"))
 	optional(project(":spring-boot-project:spring-boot-rsocket"))
 	optional("org.springframework.integration:spring-integration-jdbc")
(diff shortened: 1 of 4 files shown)


wilkinsona:
Thanks for spotting the potential problems here, @quaff, but the proposed changes are moving things in the wrong direction.

The auto-configuration of `IntegrationDataSourceScriptDatabaseInitializer` will already back off without `spring-boot-jdbc`:

https://github.com/spring-projects/spring-boot/blob/5db3cacde833e9e2652a61f36570c017aad4a789/spring-boot-project/spring-boot-integration/src/main/java/org/springframework/boot/integration/autoconfigure/IntegrationAutoConfiguration.java#L267

Given this, the implementation dependency on `spring-boot-sql` in unnecessary. We're also missing a test. I've opened https://github.com/spring-projects/spring-boot/issues/46244.

The auto-configuration of `QuartzDataSourceScriptDatabaseInitializer` does not back off without `spring-boot-jdbc` and it should. I've opened #46243 for this.

quaff:
@wilkinsona What about `BatchDataSourceScriptDatabaseInitializer` and `JdbcSessionDataSourceScriptDatabaseInitializer`, should `spring-boot-jdbc` be optional dependency also?

wilkinsona:
No, I don't think so. Boot's auto-configuration for Spring Batch requires JDBC as it always sets up the JDBC-based job store. As such, a required dependency on `spring-boot-jdbc` is OK there. Similarly, `spring-boot-session-jdbc` is a separate module that requires JDBC so a required dependency on `spring-boot-jdbc` is OK there too.

quaff:
> Spring Batch requires JDBC as it always sets up the JDBC-based job store.

Spring Batch supports to set up the MongoDB-based job store now, see https://github.com/spring-projects/spring-batch/issues/877

wilkinsona:
We're aware of that but it doesn't affect the accuracy of what I said about Boot's auto-configuration for Batch. As and when we add auto-configuration for the MongoDB-based store, this situation will change.

PR #46252, opened 2025-07-02
Polish: Remove unnecessary semicolon

PiyalAhmed:
Remove unnecessary semicolons from .gradle files.
Remove unnecessary 'def' modifier and make method static


wilkinsona:
Thank you, @PiyalAhmed.

PR #46253, opened 2025-07-02
Reuse parsed expression in SpelValueExpressionResolver

doumdoum:
So far expression is parsed everytime the value is resolved.
My proposal is to reuse the parsed expression.
Since the value resolver is used only for a predefined set of expressions the map should contain only a few items.

### DIFF
--- a/spring-boot-project/spring-boot-observation/src/main/java/org/springframework/boot/observation/autoconfigure/SpelValueExpressionResolver.java
+++ b/spring-boot-project/spring-boot-observation/src/main/java/org/springframework/boot/observation/autoconfigure/SpelValueExpressionResolver.java
@@ -16,10 +16,12 @@
 
 package org.springframework.boot.observation.autoconfigure;
 
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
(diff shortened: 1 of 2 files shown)


wilkinsona:
> Since the value resolver is used only for a predefined set of expressions

I don't think we can be certain of that as the `ValueExpressionResolver` is exposed as a bean. It could then be (mis)used in such a way that the cache becomes very large. Using Spring Framework's  `ConcurrentReferenceHashMap` would avoid this potential problem.

> So far expression is parsed everytime the value is resolved. My proposal is to reuse the parsed expression.

I wonder if this may be premature optimisation as we haven't seen any issues with `@SpanTag` where the SpEL-based resolver's been used for some time now. Do you have performance data showing the expression resolution adding an unacceptably large overhead?

doumdoum:
@wilkinsona, i did some rudimentary performance tests and the gain is marginal (few micro-seconds).
Maybe not worth it.

wilkinsona:
It's certainly not clear that it's worth it. Let's leave things as they are for now. We can reconsider in the future if it becomes apparent that we need to strike a different balance between memory usage and CPU usage. Thanks anyway.

PR #46256, opened 2025-07-03
Use ThreadLocal.remove() instead of set(null)

Dockerel:
Use ThreadLocal.remove() instead of set(null) to prevent memory leak

- Replaced usages of `ThreadLocal.set(null)` with `ThreadLocal.remove()` to properly clear thread-local values.

Previously, `ThreadLocal.set(null)` was used to clear the ThreadLocal variable. However, this only sets the value to null and leaves an entry in the thread’s local variable map, which can lead to memory leaks. Instead, by using `ThreadLocal.remove()`, we can properly remove the ThreadLocal variable from the thread's local variable map, preventing memory leaks.

I have reviewed the codebase, and this was the only place where set(null) needed to be replaced with remove().

Let me know if you have any concerns or suggestions for further improvement.
Thank you for your time and consideration!

### DIFF
--- a/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/SpringApplicationTests.java
+++ b/spring-boot-project/spring-boot/src/test/java/org/springframework/boot/SpringApplicationTests.java
@@ -159,6 +159,7 @@ import static org.mockito.Mockito.spy;
  * @author Tadaya Tsuyukubo
  * @author Yanming Zhou
  * @author Sijun Yang
+ * @author Giheon Do
  */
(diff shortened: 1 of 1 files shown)


wilkinsona:
Thank you, @Dockerel.

FWIW, I don't think there's much risk from a memory leak here. The `ThreadLocal` in question is only used in a single test and will only ever be accessed on the main thread. That said, it's semantically correct to call `remove()` rather than `set(null)` so it was a worthwhile change regardless. Thank again.

zessjls:
good

PR #46289, opened 2025-07-04
Optimize resource lookup in DevTools restart

move-hoon:
# Optimize resource lookup in DevTools restart
This pull request significantly improves the performance of resource lookups within the DevTools restart mechanism, particularly benefiting projects with a large number of files and source directories.

# Motivation and Problem
The existing `ClassLoaderFilesResourcePatternResolver` and its underlying `ClassLoaderFiles` implementation suffer from performance degradation in large-scale projects. Key methods such as `getFile()`, `size()`, `isDeleted()`, and `getAdditionalResources()` rely on nested loops that iterate through all source directories and their respective files. This results in a time complexity of O(n*m) for many critical operations, where 'n' is the number of source directories and 'm' is the number of files. As a project grows, these repeated iterations can introduce noticeable delays during application restarts.

# Solution
This PR addresses the performance bottleneck by refactoring `ClassLoaderFiles` to use a pre-computed, flattened map for direct lookups, and updating `ClassLoaderFilesResourcePatternResolver` to leverage this new, efficient structure.

## 1. `ClassLoaderFiles` Refactoring:
- **Introduced a Flattened Map**: 
A new `Map<String, ClassLoaderFile> filesByName` field has been added. This map is populated once at construction time and serves as a unified, central cache for all files across all source directories.
- **O(1) Complexity for Core Methods**:
    - `getFile(name)` now performs a direct `get()` on the `filesByName` map, reducing its complexity from O(n) to O(1).
    - `size()` now directly returns `filesByName.size()`, reducing its complexity from O(n) to O(1).
- **Data Consistency**:
The `addFile()` and `removeAll()` methods have been updated to ensure that both the original `sourceDirectories` map and the new `filesByName` map are kept in sync, guaranteeing data integrity.
- **New Efficient Iterator**:
 A new `getFileEntries()` method provides a way to iterate over all files in a single loop, which is used by the resolver.

## 2. `ClassLoaderFilesResourcePatternResolver` Optimization:
- **Optimized `isDeleted()`**: 
This method has been refactored to eliminate the nested loop structure (from O(n*m) to O(m)). It now iterates over the single, unified collection provided by `classLoaderFiles.getFileEntries()`, significantly reducing the complexity of the operation while preserving the original exception handling behavior.
- **Optimized `getAdditionalResources()`**:
Similarly, this method now iterates over the single, unified collection provided by `classLoaderFiles.getFileEntries()`, eliminating the outer loop and improving readability and performance.

# Benefits
**Significant Performance Improvement:**
Reduces the time complexity of critical file lookup operations. Direct lookups like `getFile` are improved from O(n) to O(1), while iterative searches like `isDeleted` are improved from O(n*m) to O(m), leading to faster application restarts in DevTools.

**Enhanced Code Quality**: 
Eliminates nested loops, which simplifies the code, improves readability, and makes the logic easier to maintain.

**Minimal Memory Overhead**:
The optimization introduces a map of references, not data duplicates. The trade-off of a negligible increase in memory usage for a substantial performance gain is highly favorable.

This change maintains full backward compatibility with the existing API while providing a much-needed performance boost for developers working on large applications.

### DIFF
--- a/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/classloader/ClassLoaderFiles.java
+++ b/spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/classloader/ClassLoaderFiles.java
@@ -44,10 +44,17 @@ public class ClassLoaderFiles implements ClassLoaderFileRepository, Serializable
 	private final Map<String, SourceDirectory> sourceDirectories;
 
 	/**
+	 * A flattened map of all files from all source directories for fast, O(1) lookups.
+	 * The key is the file's relative path, and the value is the ClassLoaderFile.
+	 */
+	private final Map<String, ClassLoaderFile> filesByName;
+
+	/**
 	 * Create a new {@link ClassLoaderFiles} instance.
 	 */
 	public ClassLoaderFiles() {
 		this.sourceDirectories = new LinkedHashMap<>();
+		this.filesByName = new LinkedHashMap<>();
 	}
 
 	/**
@@ -57,6 +64,7 @@ public class ClassLoaderFiles implements ClassLoaderFileRepository, Serializable
 	public ClassLoaderFiles(ClassLoaderFiles classLoaderFiles) {
 		Assert.notNull(classLoaderFiles, "'classLoaderFiles' must not be null");
 		this.sourceDirectories = new LinkedHashMap<>(classLoaderFiles.sourceDirectories);
+		this.filesByName = new LinkedHashMap<>(classLoaderFiles.filesByName);
 	}
 
 	/**
@@ -94,12 +102,14 @@ public class ClassLoaderFiles implements ClassLoaderFileRepository, Serializable
 		Assert.notNull(file, "'file' must not be null");
 		removeAll(name);
 		getOrCreateSourceDirectory(sourceDirectory).add(name, file);
+		this.filesByName.put(name, file);
 	}
 
 	private void removeAll(String name) {
 		for (SourceDirectory sourceDirectory : this.sourceDirectories.values()) {
 			sourceDirectory.remove(name);
 		}
+		this.filesByName.remove(name);
 	}
 
 	/**
@@ -125,22 +135,21 @@ public class ClassLoaderFiles implements ClassLoaderFileRepository, Serializable
 	 * @return the size of the collection
 	 */
 	public int size() {
-		int size = 0;
-		for (SourceDirectory sourceDirectory : this.sourceDirectories.values()) {
-			size += sourceDirectory.getFiles().size();
-		}
-		return size;
+		return this.filesByName.size();
 	}
 
 	@Override
 	public ClassLoaderFile getFile(String name) {
-		for (SourceDirectory sourceDirectory : this.sourceDirectories.values()) {
-			ClassLoaderFile file = sourceDirectory.get(name);
-			if (file != null) {
-				return file;
-			}
-		}
-		return null;
+		return this.filesByName.get(name);
+	}
+
+	/**
+	 * Returns a set of all file entries across all source directories for efficient
+	 * iteration.
+	 * @return a set of all file entries
+	 */
+	public Set<Entry<String, ClassLoaderFile>> getFileEntries() {
+		return Collections.unmodifiableSet(this.filesByName.entrySet());
 	}
 
 	/**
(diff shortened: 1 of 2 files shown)


wilkinsona:
Thanks for the PR, @hoonyworld. Can you please sign your commit as [described in the check failure](https://github.com/spring-projects/spring-boot/pull/46289/checks?check_run_id=45361842471).

move-hoon:
@wilkinsona Thanks for the confirmation. It seems I missed the DCO-related part. I'll take care of it right away as guided.

spring-boot-project/spring-boot-devtools/src/main/java/org/springframework/boot/devtools/restart/classloader/ClassLoaderFiles.java
```diff
@@ -43,11 +43,18 @@ public class ClassLoaderFiles implements ClassLoaderFileRepository, Serializable
 
 	private final Map<String, SourceDirectory> sourceDirectories;
 
+	/**
+	 * A flattened map of all files from all source directories for fast, O(1) lookups.
+	 * The key is the file's relative path, and the value is the ClassLoaderFile.
+	 */
+	private final Map<String, ClassLoaderFile> filesByName;
```
mhalbritter, line 50: This tracks files by name. However,

```java
public void addFile(String sourceDirectory, String name, ClassLoaderFile file) {
```

takes a `sourceDirectory` and a `name`. What happens if I add two files, which are in different directories, but have the same name?

move-hoon: Thank you for the excellent discussion on this topic. I've been following along, and I have a question to make sure I understand correctly.

It appears that the root cause of the file collision problem is the removeAll(name) call, which was part of the original implementation. This would mean that even before the introduction of the filesByName map for performance, the logic would still overwrite a file from one directory when adding a file with the same name from another.

So, my understanding is that the reviewer is pointing out a long-standing, fundamental flaw rather than a new bug introduced in this pull request.

Could you confirm if this understanding is correct? Any clarification would be greatly appreciated if I'm on the wrong track.

hoonyworld: ## The Problem
```java
files.addFile("dir1", "MyClass.class", file1);
files.addFile("dir2", "MyClass.class", file2);  // file1 gets removed!
```
- The removeAll(name) call removes all existing files with the same name before adding the new one.

## Proposed Solution: Composite Key Strategy
### Core Idea: Use (sourceDirectory, fileName) combination as a unique key

```java
// Current
private final Map<String, ClassLoaderFile> filesByName;

// Improved  
private final Map<FileKey, ClassLoaderFile> filesByKey;
private final Map<String, List<FileKey>> fileNameIndex;  // For fast name-based lookup
```

### FileKey class:
```java
public static class FileKey {
    private final String sourceDirectory;
    private final String fileName;
    // equals, hashCode implementation
}
```

## Benefits
1. **Complete Separation**: Files with same names from different directories can coexist
2. **O(1) Performance**: Direct key access for fast retrieval
3. **Flexible Access Patterns**:
- getFile(name) - returns first found file
- getFile(sourceDirectory, name) - returns specific location file
- getAllFilesByName(name) - returns all files with same name
4. **Backward Compatibility**: Existing API preserved

This approach ensures accurate file management in Spring Boot DevTools class reloading scenarios without file conflicts.

mhalbritter: Oh, you're right! I missed the `removeAll` call in the `void addFile(String sourceDirectory, String name, ClassLoaderFile file)` method. Then your approach with the map is the correct one. Thanks for pointing that out.

mhalbritter:
Thanks @move-hoon !

move-hoon:
Hi @mhalbritter

Thank you for the review and guidance. Since this is my first contribution to Spring Boot, I’d like to briefly confirm my understanding:

1. The filesByName map for O(1) lookups has been merged (commit c4a4b63) and will be included in 4.0.0-M3.
2. The behavior where files with the same relative name from different directories overwrite each other (removeAll(name))  remains as before, meaning this PR focused on performance rather than changing that design.

Could you confirm if this is correct? Thanks again for your time and support! 🙏

mhalbritter:
Hello @move-hoon , congrats on your first contribution to Spring Boot! Keep them coming!

1. Yes.
2. Yes.

PR #46317, opened 2025-07-07
Remove unnecessary method overrides

PiyalAhmed:
Remove methods that are identical to the super methods


snicoll:
Thanks but these are fine. Them being identical to their parent is no reason to remove them as they make the contract of the class more explicit and any change in the parent would have an impact on the child.

Issue #46212, opened 2025-06-26
Auto-configured JCacheMetrics cannot be customized


wilkinsona:
See https://github.com/spring-projects/spring-boot/issues/46017 for background.

Micrometer is going to provide a setting that allows the JCache removals to be [registered as a function counter](https://github.com/micrometer-metrics/micrometer/pull/6471). As far as I can tell, that option won't be usable in Boot as the auto-configured `JCacheCacheMeterBinderProvider` isn't `@ConditionalOnMissingBean` so there's no easy way to define your own provider with a custom `JCacheMetrics` that uses a function counter for removals.

A couple of options:

1. Make the auto-configured `JCacheCacheMeterBinderProvider` `@ConditionalOnMissingBean`
2. Provide a configuration property, something like `management.metrics.cache.jcache.register-removals-as-function-counter`, to control Micrometer's new setting.

We could do either or both of these.

hakimrabet:
I've submitted a PR. [spring-projects/spring-boot#46355](https://github.com/spring-projects/spring-boot/pull/46355). It makes JCacheCacheMeterBinderProvider overridable and add a configuration to register removals as a function counter. I'm not sure if it's exactly what you're looking for, but i'm happy to adjust it based on feedback.

wilkinsona:
We don't know what we're looking for at the moment, hence the pending-design-work label. That design work will have to be done by a member of the core team before we can proceed.

jonatan-ivanov:
fyi: https://github.com/micrometer-metrics/micrometer/pull/6471 is released
Also, latest bits of Micrometer should prevent registering different Prometheus types with the same name: https://github.com/micrometer-metrics/micrometer/pull/6470 (scrape should not fail but time series will be missing).

wilkinsona:
You can now choose to have removals registered as a function counter by defining your own bean:

```java
@Bean
JCacheCacheMeterBinderProvider customMeterBinderProvider() {
    return new JCacheCacheMeterBinderProvider(true);
}
```

PR #46355, opened 2025-07-08
allow JCacheMetrics to be customized and support function counter

hakimrabet:
this is bug fix for issue:  #46212 , for autoconfiguration of jCache

[https://github.com/spring-projects/spring-boot/issues/46212]()


wilkinsona:
#46212 is labelled as pending design work. That means that we're not ready to make changes to the code. Thanks anyway for the PR.

PR #46372, opened 2025-07-09
Avoid using deprecated Log4j exception converter constructors

vy:
Log4j `2.25.0` deprecated `ThrowablePatternConverter` ctors, where they are employed in Spring Boot's `[Extended]WhitespaceThrowablePatternConverter` – see apache/logging-log4j2#3809. This PR amends these classes to avoid using deprecated ctors.

### DIFF
--- a/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/ExtendedWhitespaceThrowablePatternConverter.java
+++ b/spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/ExtendedWhitespaceThrowablePatternConverter.java
@@ -44,7 +44,7 @@ public final class ExtendedWhitespaceThrowablePatternConverter extends LogEventP
 	private ExtendedWhitespaceThrowablePatternConverter(Configuration configuration, String[] options) {
 		super("WhitespaceExtendedThrowable", "throwable");
 		this.delegate = ExtendedThrowablePatternConverter.newInstance(configuration, options);
-		this.separator = WhitespaceThrowablePatternConverter.readSeparatorOption(options);
+		this.separator = ((ExtendedThrowablePatternConverter) this.delegate).getOptions().getSeparator();
 	}
 
 	@Override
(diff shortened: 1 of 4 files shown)


spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/ExtendedWhitespaceThrowablePatternConverter.java
```diff
@@ -34,24 +35,32 @@
  */
 @Plugin(name = "ExtendedWhitespaceThrowablePatternConverter", category = PatternConverter.CATEGORY)
 @ConverterKeys({ "xwEx", "xwThrowable", "xwException" })
-public final class ExtendedWhitespaceThrowablePatternConverter extends ThrowablePatternConverter {
+public final class ExtendedWhitespaceThrowablePatternConverter extends LogEventPatternConverter {
```
ppkarwasz, line 38: _Nit-pick_: Since `ExtendedWhitespaceThrowablePatternConverter` and `WhitespaceThrowablePatternConverter` differ only in the specific `LogEventPatternConverter` they delegate to, the implementation can be simplified by:

* Having one implemented in terms of the other, or
* Taking advantage of the fact that pattern converters don’t need to directly implement `LogEventPatternConverter`, but only provide a `public static LogEventPatternConverter newInstance()` method, you could have both plugins return the same shared private implementation of `LogEventPatternConverter`, reducing duplication.

ppkarwasz (commented):
Looks good to me, although the implementation could be further simplified.

vy:
@snicoll, please let us know how you want to proceed with. I've tried to keep the changes minimal, though I can implement the improvements suggested by @ppkarwasz, if it is of your liking.

spring-boot-project/spring-boot/src/main/java/org/springframework/boot/logging/log4j2/WhitespaceThrowablePatternConverter.java
```diff
@@ -32,21 +34,43 @@
...
+
+	private final LogEventPatternConverter delegate;
+
+	private final String separator;
 
 	private WhitespaceThrowablePatternConverter(Configuration configuration, String[] options) {
-		super("WhitespaceThrowable", "throwable", options, configuration);
+		super("WhitespaceThrowable", "throwable");
+		this.delegate = ExtendedThrowablePatternConverter.newInstance(configuration, options);
+		this.separator = readSeparatorOption(options);
```
wilkinsona: I'd like to avoid the lightweight parsing of the options that's proposed below in `readSeparatorOption`. Could this not use the separator from the delegate instead?

```
this.separator = this.delegate.getOptions().getSeparator();
```

vy: Given the problem was originated from relying too much on Log4j internals, I intentionally did not want to use the separator from the delegate. Changed as you requested in 71f8f5e3c0.

wilkinsona:
Unfortunately, we need to revert this for now due to https://github.com/apache/logging-log4j2/issues/3893 as warnings caused by the undefined property break our verification tests.

snicoll:
This was re-applied as part of https://github.com/spring-projects/spring-boot/commit/b7695200a90523167a39da5654e1ece55c4dc42f

Issue #45348, opened 2025-04-30
Auto-configure virtual threads on JdkClientHttpRequestFactory when enabled


drub0y:
I have seen the conversation in spring-projects/spring-framework#34393, but I wanted to re-open the conversation because I disagree with the reasoning for why the issue was closed as won't fix and would like to make a new case to help people avoid the issue I see my teams running into constantly.

To recap, when a developer sets `spring.threads.virtual.enabled` to true, they have what I think is a reasonable expectation that the Spring framework will do it's best to make choices to configure anything that it _can_ configure to use virtual threading for them. It does this in a lot of cases and, one of those cases, is to configure the default `SimpleAsyncTaskExecutor` to use virtual threads.

In the case of `JDKClientHttpRequestFactory` it will detect that no executor has been set on the HttpClient and create a brand new, dedicated `SimpleAsyncTaskExecutor`, but it doesn't bother consulting the `spring.threads.virtual.enabled` setting to see if it should be configured to use virtual threads. 

IMHO, this level of indirection that `JDKClientHttpRequestFactory` provides is Spring's chance to "help" meet the expectations of developers and, today, it doesn't.

NOTE: if the developer "brings their own" `HttpClient` and it happens to have an executor already set, the logic in `JDKClientHttpRequestFactory` already won't override it, so making changes here should be non-breaking. It's really just about updating the implementation to configure the dedicated `SimpleAsyncTaskExecutor` with virtual threads when `spring.threads.virtual.enabled` is `true`.

drub0y:
Totally understand the `RestClient.create()` case, but I think it's fair to say if you drop down to that level of API you should know you're bypassing all manner of injection/autoconfiguration... obviously wouldn't hurt to note that in documentation of course.

While we do use `RestClient.Builder` today, we're still the ones having to directly instantiate and configure the `HttpClient` and wrap it with the `JdkHttpClientHttpRequestFactory` and assign it via `RestClient.Builder`'s `requestExecutor(...)` and that's where all the `new`'ing comes in that obviously bypasses Spring configuration and leads to this unfortunate situation. 

[...]

PR #46404, opened 2025-07-11
Ensure RestClient uses Virtual Thread Executor when spring.threads.virtual.enabled=true and JdkClientHttpRequestFactoryBuilder

smartandhandsome:
Previously, even when setting spring.threads.virtual.enabled=true, HttpClient in RestClient continued to create platform threads. This meant that enabling virtual threads via configuration did not actually switch the executor as users would expect.

**What has changed**
- The configuration for the ClientHttpRequestFactoryBuilder bean has been split based on the threading mode.
- When Threading.VIRTUAL is active and the builder is a JdkClientHttpRequestFactoryBuilder, the builder is explicitly configured to use a virtual thread executor.
- For platform threads, the previous behavior is preserved.


philwebb:
Flagging for team attention to see if we consider this a bug or enhancement. I lean towards enhancement.

smartandhandsome:
While checking, I noticed that there is an existing related issue. I’ll attach it here so it can be referenced.

https://github.com/spring-projects/spring-boot/issues/45348

philwebb:
Thanks @smartandhandsome!

smartandhandsome:
@philwebb

Thank you so much for adding me as an author and supporting my contribution to Spring Boot! It is such an honor for me as a Spring user. Thank you again!

hannah23280:
@smartandhandsome   
Thanks for the contribution.  But not sure as of now, April 2026, is this enhancement already implemented in the current version of Spring ?

bclozel:
@hannah23280 yes it is. You can see in which version this was released on this page. In this case, it is available as of Spring Boot 4.0

PR #46408, opened 2025-07-12
Fix BootZipCopyAction.Processor.getDirMode(FileCopyDetails)

izeye:
This PR fixes `BootZipCopyAction.Processor.getDirMode(FileCopyDetails)` as it seems to reference a wrong one accidentally.

See gh-46193


wilkinsona:
Thanks very much, @izeye. I've said it before, but I think it's worth repeating: I have no idea how you spot these things but I am very grateful that you do.

PR #46467, opened 2025-07-18
Replace String.format with concatenation in CorrelationIdFormatter

Kguswo:
Replace `String.format()` with string concatenation for simple text joining in `CorrelationIdFormatter`.

**Changes:**
- Use `"[" + ... + "] "` instead of `String.format("[%s] ", ...)` in constructor

**Motivation:**
- Improves performance for simple string concatenation (no actual formatting needed)
- String concatenation is more appropriate when no format specifiers are used
- Reduces unnecessary overhead of format string parsing

This change maintains identical functionality while using a more efficient approach for basic string joining.

### DIFF
--- a/core/spring-boot/src/main/java/org/springframework/boot/logging/CorrelationIdFormatter.java
+++ b/core/spring-boot/src/main/java/org/springframework/boot/logging/CorrelationIdFormatter.java
@@ -76,7 +76,7 @@ public final class CorrelationIdFormatter {
 
 	private CorrelationIdFormatter(List<Part> parts) {
 		this.parts = parts;
-		this.blank = String.format("[%s] ", parts.stream().map(Part::blank).collect(Collectors.joining(" ")));
+		this.blank = "[" + parts.stream().map(Part::blank).collect(Collectors.joining(" ")) + "] ";
 	}
 
 	/**


core/spring-boot/src/main/java/org/springframework/boot/logging/CorrelationIdFormatter.java
```diff
@@ -76,7 +76,7 @@ public final class CorrelationIdFormatter {
 
 	private CorrelationIdFormatter(List<Part> parts) {
 		this.parts = parts;
-		this.blank = String.format("[%s] ", parts.stream().map(Part::blank).collect(Collectors.joining(" ")));
+		this.blank = "[" + parts.stream().map(Part::blank).collect(Collectors.joining(" ")) + "] ";
```
kzander91, line 79: Why don't we at this point just use the `Collectors.joining()` [variant with a prefix and suffix](https://docs.oracle.com/en/java/javase/17/docs/api/java.base/java/util/stream/Collectors.html#joining(java.lang.CharSequence,java.lang.CharSequence,java.lang.CharSequence))?
```suggestion
		this.blank = parts.stream().map(Part::blank).collect(Collectors.joining(" ", "[", "] "));
```

philwebb:
@OrangeDoro I'm sorry but we're a small team and we aren't able to respond to requests such as yours.

philwebb:
Thanks for the PR and the additional suggestion @kzander91. As this code is called pretty infrequently I don't think we have any performance issues. The other changes are pretty subjective and as such I'd prefer to leave things as they are.

PR #46505, opened 2025-07-22
Polish

izeye:
This PR polishes a bit.

### DIFF
--- a/documentation/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
+++ b/documentation/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
@@ -1173,7 +1173,7 @@ Beans used for property conversion are requested very early during the applicati
 Typically, any dependency that you require may not be fully initialized at creation time.
 ====
 
-TIP: You may want to rename your custom javadoc:org.springframework.core.convert.ConversionService[] if it is not required for configuration keys coercion and only relies on custom converters qualified with javadoc:org.springframework.boot.context.properties.ConfigurationPropertiesBinding[format=annotation].
+TIP: You may want to rename your custom javadoc:org.springframework.core.convert.ConversionService[] if it is not required for configuration keys coercion and only rely on custom converters qualified with javadoc:org.springframework.boot.context.properties.ConfigurationPropertiesBinding[format=annotation].
 When qualifying a `@Bean` method with `@ConfigurationPropertiesBinding`, the method should be `static` to avoid "`bean is not eligible for getting processed by all BeanPostProcessors`" warnings.
 
 
--- a/documentation/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
+++ b/documentation/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
@@ -1173,7 +1173,7 @@ Beans used for property conversion are requested very early during the applicati
 Typically, any dependency that you require may not be fully initialized at creation time.
 ====
 
-TIP: You may want to rename your custom javadoc:org.springframework.core.convert.ConversionService[] if it is not required for configuration keys coercion and only rely on custom converters qualified with javadoc:org.springframework.boot.context.properties.ConfigurationPropertiesBinding[format=annotation].
+TIP: You may want to rename your custom javadoc:org.springframework.core.convert.ConversionService[] if it is not required for configuration keys coercion and only relies on custom converters qualified with javadoc:org.springframework.boot.context.properties.ConfigurationPropertiesBinding[format=annotation].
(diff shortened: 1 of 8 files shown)


documentation/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
```diff
@@ -1173,7 +1173,7 @@ Beans used for property conversion are requested very early during the applicati
 Typically, any dependency that you require may not be fully initialized at creation time.
 ====
 
-TIP: You may want to rename your custom javadoc:org.springframework.core.convert.ConversionService[] if it is not required for configuration keys coercion and only rely on custom converters qualified with javadoc:org.springframework.boot.context.properties.ConfigurationPropertiesBinding[format=annotation].
+TIP: You may want to rename your custom javadoc:org.springframework.core.convert.ConversionService[] if it is not required for configuration keys coercion and only relies on custom converters qualified with javadoc:org.springframework.boot.context.properties.ConfigurationPropertiesBinding[format=annotation].
```
wilkinsona: I think "rely" is correct here.

module/spring-boot-tracing/src/test/java/org/springframework/boot/tracing/autoconfigure/zipkin/ZipkinConfigurationsOpenTelemetryConfigurationTests.java
```diff
@@ -44,11 +44,12 @@ class ZipkinConfigurationsOpenTelemetryConfigurationTests {
 
 	@Test
 	void backsOffWithoutEncoding() {
-		new ApplicationContextRunner().withUserConfiguration(OpenTelemetryConfiguration.class).run((context) -> {
-			assertThat(context).hasNotFailed();
-			assertThat(context).doesNotHaveBean(ZipkinSpanExporter.class);
-			assertThat(context).doesNotHaveBean(BytesEncoder.class);
-		});
+		new ApplicationContextRunner().withConfiguration(AutoConfigurations.of(OpenTelemetryConfiguration.class))
```
wilkinsona: `OpenTelemetryConfiguration` isn't an `@AutoConfiguration` so this is fine as it was.

documentation/spring-boot-docs/src/docs/antora/modules/reference/pages/features/external-config.adoc
izeye: @wilkinsona Thanks for the feedback!

I reverted it in 32c9bbe.

module/spring-boot-tracing/src/test/java/org/springframework/boot/tracing/autoconfigure/zipkin/ZipkinConfigurationsOpenTelemetryConfigurationTests.java
izeye: @wilkinsona I don't know exactly what it does behind the scenes, but I saw the following code in the same class, so I thought that a part of the auto-configuration also uses `AutoConfigurations.of()`:

https://github.com/spring-projects/spring-boot/blob/4babe2e59d03971925f8f52f13495ee348a34a04/module/spring-boot-tracing/src/test/java/org/springframework/boot/tracing/autoconfigure/zipkin/ZipkinConfigurationsOpenTelemetryConfigurationTests.java#L42-L43

There are more occurrences like this, for example, as follows:

https://github.com/spring-projects/spring-boot/blob/4babe2e59d03971925f8f52f13495ee348a34a04/module/spring-boot-pulsar/src/test/java/org/springframework/boot/pulsar/autoconfigure/PulsarConfigurationTests.java#L73-L75

I also thought that "User" in the `withUserConfiguration()` doesn't sound right for a part of the auto-configuration.

Are you suggesting that the mixed uses are okay, or that they need to be used with the `withUserConfiguration()` if it's not an `@AutoConfiguration`?

wilkinsona: `AutoConfigurations.of` does two things:

1. It sorts all of the passed-in configurations based on their before and after attributes
2. It orders the configurations so that they're processed before any user configurations

In this case, the configuration isn't an auto-configuration so there are no before and after attributes to consider and there's only of them so there's nothing to sort. There's also only one configuration class in the test so the order in which they're processed doesn't matter either.

We're pretty inconsistent about this across our tests. What was there before is fine and what you've proposed is also fine. Given this, I'd prefer to leave things as they were.

izeye: @wilkinsona Thanks for the explanation!

I reverted it in 00e2d1b.

Issue #46510, opened 2025-07-23
Upgrade to Groovy 4.0.28


jamesfredley:
Upgrade to Groovy 4.0.28

https://github.com/apache/groovy/releases/tag/GROOVY_4_0_28

https://groovy-lang.org/changelogs/changelog-4.0.28.html

jamesfredley:
https://github.com/spring-projects/spring-boot/pull/46511

snicoll:
Please review the issue template before creating an issue. We don't need an issue for dependency upgrades:

```
- Managed Dependency Upgrade
You DO NOT need to raise an issue for a managed dependency version upgrade as there's a semi-automatic process for checking managed dependencies for new versions before a release. BUT pull requests for upgrades that are more involved than just a version property change are still most welcome.
```

Also, you should not open an issue if you intend to create a PR immediately:

```
- With an Immediate Pull Request
An issue will be closed as a duplicate of the immediate pull request, so you don't have to raise an issue if you plan to create a pull request immediately.
```

jamesfredley:
@snicoll Ok, thank you.  I was trying to be helpful and follow your prior issue/PRs for Groovy version updates.  

Thank you for running with this on:
https://github.com/spring-projects/spring-boot/issues/46512
https://github.com/spring-projects/spring-boot/issues/46514
https://github.com/spring-projects/spring-boot/issues/46516

snicoll:
No worries

>  your prior issue/PRs for Groovy version updates.

Pull Request are [declined](https://github.com/spring-projects/spring-boot/pulls?q=is%3Apr+%22upgrade+to+groovy%22+is%3Aclosed) unless they aren't a simple version bump (code change is required). As for the issues, these are created automatically by a tool we're managing.

PR #46511, opened 2025-07-23
Upgrade to Groovy 4.0.28

jamesfredley:
Upgrade to Groovy 4.0.28
Closes gh-46510


snicoll:
Thanks for the PR but, as described in the issue template, we do not need PRs for simple dependency upgrades.

PR #46520, opened 2025-07-23
Remove unused SHA-1 hash from UNPACK markers

### LINKED ISSUE #46183: Redundant hash calculation for requiresUnpack archive entries
In `BootZipCopyAction` and `AbstractJarWriter` SHA1 hash is calculated for stored entries requiring unpack and set as entry comment:

```java
archiveEntry.setComment("UNPACK:" + FileUtils.sha1Hash(details.getFile()));
```

However the hash isn't used anywhere, just the marker prefix `UNPACK:` is checked.

Dug through history and first introduction of `UNPACK:` + hash seems to have been in https://github.com/spring-projects/spring-boot/commit/f30b962ff9327d3264d519853992aff972cdbe5a

At that point the hash was extracted and used for the filename:

```java
AsciiBytes hash = data.getComment().substring(UNPACK_MARKER.length());
...
File file = new File(getTempUnpackFolder(), hash.toString() + "-" + name);
```

But then later in https://github.com/spring-projects/spring-boot/commit/7e718cda265b85ad6337a0aa711906c7e1097e34 this got removed from output filename:

```patch
-		AsciiBytes hash = data.getComment().substring(UNPACK_MARKER.length());
-		File file = new File(getTempUnpackFolder(), hash.toString() + "-" + name);
+		File file = new File(getTempUnpackFolder(), name);
```

So now the hash is still being calculated and set for the entry but I don't see it being used anywhere.
Could it just be left as `UNPACK:` without any hash?

The hashing reads the file completely in memory due to usage of `DigestInputStream.readAllBytes()` so in the extreme case the file is fully read 3 times: once by `CrcAndSize` calculation, once for the sha1 hash and once for actual copying to the `ZipArchiveOutputStream`

academey:
This PR removes the unused SHA-1 hash calculation from UNPACK markers in the JAR/WAR packaging process.

**Current behavior:**
When creating executable JARs/WARs, Spring Boot calculates a SHA-1 hash for entries that require unpacking and appends it to the entry comment as `UNPACK:<hash>`. However, this hash is never used - only the `UNPACK:` prefix is checked.

**Changes made:**
- Removed SHA-1 hash calculation from `StoredEntryPreparator` in both `BootZipCopyAction` and `AbstractJarWriter`
- Changed the UNPACK comment from `UNPACK:<sha1-hash>` to simply `UNPACK:`
- Updated tests to match the new format

**Performance improvement:**
This eliminates unnecessary file reads during packaging. Previously, files could be read up to 3 times (CRC calculation, SHA-1 hash, and actual copying). Removing the hash calculation reduces I/O operations and memory usage, especially beneficial for large libraries.

Fixes #46183

### DIFF
--- a/build-plugin/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootZipCopyAction.java
+++ b/build-plugin/spring-boot-gradle-plugin/src/main/java/org/springframework/boot/gradle/tasks/bundling/BootZipCopyAction.java
@@ -22,13 +22,10 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.OutputStream;
 import java.io.OutputStreamWriter;
-import java.security.MessageDigest;
-import java.security.NoSuchAlgorithmException;
 import java.time.OffsetDateTime;
 import java.time.ZoneOffset;
 import java.util.Collection;
 import java.util.HashMap;
-import java.util.HexFormat;
 import java.util.LinkedHashMap;
 import java.util.LinkedHashSet;
 import java.util.List;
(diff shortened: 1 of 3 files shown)


wilkinsona:
We should probably update the runtime side of things as well:

https://github.com/spring-projects/spring-boot/blob/376ad327591e9c5ec428966d8de2b84e85ce4216/loader/spring-boot-loader/src/main/java/org/springframework/boot/loader/launch/JarFileArchive.java#L97-L99

https://github.com/spring-projects/spring-boot/blob/376ad327591e9c5ec428966d8de2b84e85ce4216/loader/spring-boot-loader-classic/src/main/java/org/springframework/boot/loader/archive/JarFileArchive.java#L112-L114

The `startsWith` check could be tightened up to `equals()`. We may even want to change the marker from `UNPACK:` to just `UNPACK`.

@academey please don't make any changes to the PR just yet, I'd like to discuss with the team first.

wilkinsona:
We discussed this today and want to use `UNPACK` for the marker and then do an `equals` rather than `startsWith` check.

wilkinsona:
Thanks very much for the PR, @academey.

PR #46527, opened 2025-07-24
Use "2012-present" for copyright headers

izeye:
This PR changes to use "2012-present" for copyright headers as they seem to have been introduced accidentally in d49b37a6b5b53de13df3ebcb325a5625d41f4eb3.


snicoll:
Wondering if we can't make the build fail if the format isn't the one we expect.

izeye:
> Wondering if we can't make the build fail if the format isn't the one we expect.

@snicoll I tried it in c260e4b.

